PR #950, opened 2018-01-16
Support the boolean array type in DDR code

ChengJin01:
The change is to ensure the boolean array type is
correctly displayed in DDR output as we already
changed the code in verifier to differentiate the
boolean arrays from the byte arrays.


ChengJin01:
DDR code reviewer:  @pshipton 

FYI: @DanHeidinga

PR #943, opened 2018-01-16
Migrate testSCCMLTests2 to openj9/test

sophia-guo:
Signed-off-by: Sophia Guo <sophiag@ca.ibm.com>

PR #951, opened 2018-01-16
Update commandline java -version related tests match pattern

sophia-guo:
Update commandline java -version tests match pattern from "version"
without regex to "(java|openjdk) version" with regex


smlambert:
Will run extended tests from #943

PR #952, opened 2018-01-16
Revert Co-operative copy of long object

amicic:


charliegracie:
This is backing a change out so the copyright check does not matter.  I am not going to wait for the travis CI build to finish as the change was made by applying the original path in reverse order.  This will fix the current compile issues with OpenJ9

PR #960, opened 2018-01-17
Co-operative copy of large objects in CS

amicic:
For Concurrent Scavenger, use copyOrWait() API to ensure that object is
fully copied before exposing the new verison of the object. The API is
no-op in Standard Scavenger.

Introduce a varion of getArrayLayout() that will return layout info
based on preserved info in ForwardedHeader, instead of looking into the
(destination) object header itself, which is not available with
co-operative copying, since the destination object header contains copy
progress info, while copy is still in progress.


amicic:
Dependent on https://github.com/eclipse/omr/pull/2238.

amicic:
jenkins test sanity depends eclipse/omr#2238

PR #962, opened 2018-01-17
Fix Initial Memory minimum size for non-flat configuration

dmitripivkine:
This problem was exposed by Concurrent Scavenger. There is a case where
region size can be set larger then default. The Initial Memory size for
non-flat configuration can not be smaller then three regions (one for
Tenure and one for each half of Nursery). If -Xms value is smaller then
minimum it is silently corrected.
Also if Initial Memory size is smaller then eight regions the default
split 75% for Tenure / 25% for Nursery can not be established
(considering minimum two regions for Nursery is 25%). So Nursery size
has to be set to two regions and reminder goes to Tenure. 

Signed-off-by: dmitripivkine <Dmitri_Pivkine@ca.ibm.com>


dmitripivkine:
Jenkins test sanity

dmitripivkine:
I believe OOM in pLinux is unrelated. Restarting.

dmitripivkine:
Jenkins test sanity pLinux

PR #965, opened 2018-01-17
Add nest API to java.lang.Class

taliamccormick:
Adds/updates the following nest-related API methods to java.lang.Class
* Class<?> getNestHost()
* Class<?>[] getNestMembers()
* boolean isNestmateOf(Class<?> c)

Changes related to 'JEP 181: Support Nestmate Access Control in the
JVM'.

API spec here: http://cr.openjdk.java.net/~dholmes/8188075/webrev/raw_files/new/java/lang/Class.html


taliamccormick:
@tajila @DanHeidinga

jcl/src/java.base/share/classes/java/lang/Class.java
```diff
@@ -4291,8 +4292,17 @@ ConstantPool getConstantPool() {
  * @return		true if class is a nestmate of this class; false otherwise.
  *
  */
-public boolean isInSameNest(Class<?> that) {
+public boolean isNestmateOf(Class<?> that) {
```
keithc-ca: The new name isn't consistent with the (old) behavior.

runtime/jcl/common/java_lang_Class.cpp
```diff
@@ -1785,5 +1785,69 @@ Java_java_lang_Class_getNestHostImpl(JNIEnv *env, jobject recv)
...
+	vmFuncs->internalEnterVMFromJNI(currentThread);
+
+	J9Class *clazz = J9VM_J9CLASS_FROM_HEAPCLASS(currentThread, J9_JNI_UNWRAP_REFERENCE(recv));
+	J9Class *nestHost = clazz->nestHost;
+	J9ROMClass *romHostClass = nestHost->romClass;
+	nestMemberCount = romHostClass->nestMemberCount;
+
+	/*  Grab the Class<?> class for result object size */
+	jlClass = J9VMJAVALANGCLASS(vm);
+	arrayClass = fetchArrayClass(currentThread, jlClass);
```
keithc-ca: As this is C++ code, jlClass and arrayClass can also be _declared_ here.

jcl/src/java.base/share/classes/java/lang/Class.java
taliamccormick: From what I understand, this method should return true if the two classes have the same nest host (ie. are in the same nest). In both cases, it would compare the nest host of the two classes and return true if they have the same nest host and false otherwise. 

The spec gives:
```
public boolean isNestmateOf​(Class<?> c)

Determines if the given Class is a nestmate of the object represented by this Class. Two classes are nestmates if they have the same nest host.

Parameters:
    c - the class to check
Returns:
    true if this class and c are valid members of the same nest; and false otherwise.
```
What is the inconsistency between the name and the behaviour?

keithc-ca: Consider classes A and B nested within C: The way I originally read the new name I would expect `A.class.isNestmateOf(B.class)` to return false - I think the old name was clearer, but I understand you have to respect the specification.

runtime/jcl/common/java_lang_Class.cpp
```diff
@@ -1785,5 +1785,66 @@ Java_java_lang_Class_getNestHostImpl(JNIEnv *env, jobject recv)
...
+		J9ClassLoader *classLoader = clazz->classLoader;
+
+		for (i = 0; i < nestMemberCount; i++) {
+			J9Class *nestMember = NULL;
+			J9UTF8 *nestMemberName = NNSRP_GET(nestMembers[i],J9UTF8*);
+
+			PUSH_OBJECT_IN_SPECIAL_FRAME(currentThread, resultObject);
+			nestMember = vmFuncs->internalFindClassUTF8(currentThread, J9UTF8_DATA(nestMemberName), J9UTF8_LENGTH(nestMemberName), classLoader, J9_FINDCLASS_FLAG_THROW_ON_FAIL);
+			resultObject = POP_OBJECT_IN_SPECIAL_FRAME(currentThread);
+			if (NULL != nestMember) {
```
tajila: From the spec it seems like If you cant find a nestMember is a LinkageError expected. I think `internalFindClassUTF8` may set it to `ClassNotFoundException` which may not be right in this case.  

The spec also mentions that each nestMember must be validated to ensure that it has the correct nestHost.

Also, once an error is found, it might be better to stop at that point, rather then to continue looping. We should probably test this out with the OpenJDK build to see how they handle failures in this case.

jcl/src/java.base/share/classes/java/lang/Class.java
```diff
@@ -4291,8 +4292,17 @@ ConstantPool getConstantPool() {
...
 	return this.getNestHost() == that.getNestHost();
 }
+
+/**
+ * Answers the host class of the receiver's nest.
+ *
+ * @return		the host class of the receiver.
```
tajila: the comment seems like it is intended for `getNestHost`

runtime/jcl/common/java_lang_Class.cpp
taliamccormick: Thanks!

It looks like I could use `vmfuncs->vmFuncs->setCurrentExceptionUTF(currentThread, <nls message code> , <parameters to the nls message code>);`

The only examples of `setCurrentExceptionUTF` and a `LinkageError` that I'm seeing in the codebase all use `J9VMCONSTANTPOOL_JAVALANGLINKAGEERROR` - which I'm assuming should be related to constant pool issues. However some use cases of it (eg. `runtime/jcl/common/java_dyn_methodhandle.c/Java_java_lang_invoke_PrimitiveHandle_lookupField`, `runtime/vm/resolvesupport.cpp/resolveVirtualMethodRefInto`) don't look like they are dealing with constant pool issues.

Would I add a new NLS message into `runtime/nls/j9vm/j9vm.nls`, use the existing `J9VMCONSTANTPOOL_JAVALANGLINKAGEERROR`, or keep looking?

--

Also - I'm wondering what happens if a nest member doesn't claim the nest host, in this case. It should load+link without issue but should it raise a LinkageError or not? Adding a check for this could go in the same place - leaving a note here so I remember to check

jcl/src/java.base/share/classes/java/lang/Class.java
```diff
@@ -4286,6 +4286,18 @@ ConstantPool getConstantPool() {
  */
 private native Class<?> getNestHostImpl();
 
+/**
+ * Answers the nest member classes of the receiver's nest host.
+ * The list of nest members in the classfile is permitted to contain
+ * duplicates,or to explicitly include the nest host. It is not required
```
keithc-ca: nit: Please add a space after the comma.

keithc-ca: This documentation might be better attached to the public method.

runtime/jcl/common/java_lang_Class.cpp
```diff
@@ -1785,5 +1820,72 @@ Java_java_lang_Class_getNestHostImpl(JNIEnv *env, jobject recv)
...
+	/* If host class claims nest members, they should be placed in second index onwards */
+	if (0 != nestMemberCount) {
+		J9SRP *nestMembers = J9ROMCLASS_NESTMEMBERS(romHostClass);
+		U_16 i = 0;
+		/* Classes in nest are in same runtime package & therefore have same classloader */
+		J9ClassLoader *classLoader = clazz->classLoader;
+
+		for (i = 0; i < nestMemberCount; i++) {
+			J9Class *nestMember = NULL;
+			J9UTF8 *nestMemberName = NNSRP_GET(nestMembers[i],J9UTF8*);
```
keithc-ca: Coding style: space after comma, before asterisk (and in cast above).

runtime/jcl/common/java_lang_Class.cpp
```diff
@@ -1785,5 +1820,72 @@ Java_java_lang_Class_getNestHostImpl(JNIEnv *env, jobject recv)
...
+				goto _throwException;
+			} else {
+				J9JAVAARRAYOFOBJECT_STORE(currentThread, resultObject, i + 1, J9VM_J9CLASS_TO_HEAPCLASS(nestMember));
+			}
+		}
+	}
+
+	result = vmFuncs->j9jni_createLocalRef(env, resultObject);
+
+_throwException:
```
keithc-ca: This name of this label is misleading: jumping here doesn't throw an an exception. Perhaps just 'done' or 'exit'.

runtime/jcl/common/java_lang_Class.cpp
```diff
@@ -1785,5 +1820,72 @@ Java_java_lang_Class_getNestHostImpl(JNIEnv *env, jobject recv)
...
+		for (i = 0; i < nestMemberCount; i++) {
+			J9Class *nestMember = NULL;
+			J9UTF8 *nestMemberName = NNSRP_GET(nestMembers[i],J9UTF8*);
+
+			PUSH_OBJECT_IN_SPECIAL_FRAME(currentThread, resultObject);
+			nestMember = vmFuncs->internalFindClassUTF8(currentThread, J9UTF8_DATA(nestMemberName), J9UTF8_LENGTH(nestMemberName), classLoader, J9_FINDCLASS_FLAG_THROW_ON_FAIL);
+			resultObject = POP_OBJECT_IN_SPECIAL_FRAME(currentThread);
+
+			if (NULL == nestMember) {
+				goto _throwException;
```
keithc-ca: Should an exception be thrown in this case?

runtime/nls/j9cl/j9jcl.nls
```diff
@@ -422,3 +422,16 @@ J9NLS_JCL_HOSTCLASS_ANONCLASS_DIFFERENT_PACKAGES.explanation=NOTAG
...
+J9NLS_JCL_NEST_MEMBER_CLAIMS_DIFFERENT_NEST_HOST.sample_input_1=Foobar
+J9NLS_JCL_NEST_MEMBER_CLAIMS_DIFFERENT_NEST_HOST.sample_input_2=6
+J9NLS_JCL_NEST_MEMBER_CLAIMS_DIFFERENT_NEST_HOST.sample_input_3=Baz
+J9NLS_JCL_NEST_MEMBER_CLAIMS_DIFFERENT_NEST_HOST.sample_input_4=3
+J9NLS_JCL_NEST_MEMBER_CLAIMS_DIFFERENT_NEST_HOST.sample_input_3=Foobar
+J9NLS_JCL_NEST_MEMBER_CLAIMS_DIFFERENT_NEST_HOST.sample_input_4=6
+J9NLS_JCL_NEST_MEMBER_CLAIMS_DIFFERENT_NEST_HOST.explanation=Nest host claims a nest member but the nest member claims a different nest host.
+J9NLS_JCL_NEST_MEMBER_CLAIMS_DIFFERENT_NEST_HOST.system_action=The JVM will throw an IncompatibleClassChangeError exception
+J9NLS_JCL_NEST_MEMBER_CLAIMS_DIFFERENT_NEST_HOST.user_response=Contact the provider of the classfile for a corrected version.
+# END NON-TRANSLATABLE
```
keithc-ca: Please add a newline at the end of this file.

runtime/jcl/common/java_lang_Class.cpp
taliamccormick: I believe so. This is based on an updated spec draft that says

>Each listed nest member must be validated by checking its own declared nest host. Any exceptions that occur as part of this process will be thrown.

(http://cr.openjdk.java.net/%7Edholmes/8188075/webrev/raw_files/new/java/lang/Class.html#getNestMembers()).

When the case `NULL == nestMember` is true, then something went wrong with loading the nest member ie.
```
nestMember = vmFuncs->internalFindClassUTF8(currentThread, J9UTF8_DATA(nestMemberName), J9UTF8_LENGTH(nestMemberName), classLoader, J9_FINDCLASS_FLAG_THROW_ON_FAIL);
```
failed. In this case, `internalFindClassUTF8` will set an exception (eg. NoClassDefFound)

keithc-ca:
jenkins test sanity plinux,zlinux

keithc-ca:
Please squash in preparation for merging. Thanks.

PR #909, opened 2018-01-10
Fix verbose log output of freePhysicalMemory

harryyu1994:
Made the following changes in Control/CompilationThread.cpp:
- Change _cachedFreePhysicalMemoryB from int64_t to uint64_t because all the fields in J9MemoryInfo are of this type
- Remove availSwap from physical memory calculation
- Remove the test for LONG_MAX because in practice we could easily have physical memory larger than that (2GB)
- Change %lld to %llu
- Print physical memory report only if it is complete
- Change t=%u to t=%llu since getElapsedTime() returns a 64-bit value

Issue: #424
Signed-off-by: Harry Yu <harryyu1994@gmail.com>

PR #970, opened 2018-01-18
Fix display of jit memory statistics

### LINKED ISSUE #424: Fix verbose log output of freePhysicalMemory on 31-bit z/OS
Currently on 31-bit z/OS the verbose log printout of `-Xjit:verbose={compile*}` says that `freePhysicalMemory=1125899906842624 KB` which is definitely false. There seems to be either a bug in the printout of the wrong value is returned by on of the APIs. If we are unable to retrieve this value we should not be printing it at all rather than misleading the user with an incorrect value.

pdbain-ibm:
Remove code that sometimes unreasonably capped the amount of memory returned.

Fixed output format specifiers to match actual argument sizes.

Fixes #424

### DIFF
--- a/runtime/compiler/trj9/control/CompilationThread.cpp
+++ b/runtime/compiler/trj9/control/CompilationThread.cpp
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2000, 2017 IBM Corp. and others
+ * Copyright (c) 2000, 2018 IBM Corp. and others
  *
  * This program and the accompanying materials are made available under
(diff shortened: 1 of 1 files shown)


mpirvu:
Similar fixes (and more) are waiting to be integrated in PR #909

pshipton:
@pdbain-ibm are you ok to close this since the fixes are being covered by #909?

pdbain-ibm:
Okay.  Please close.

pdbain-ibm:
Covered by other pull request.

PR #971, opened 2018-01-18
Rename new encodedUTF8Length overload to fix DDR failures

fjeremic:
DDR does not seem to like overloaded functions with parameter types.
Rename the newly added encodedUTF8Length function to fix this problem
until the new DDR solution is rolled in.


fjeremic:
Jenkins test sanity plinux

PR #978, opened 2018-01-18
Fix timing hole releasing safe point VM access

gacholio:
When releasing safe point VM access, the vmThreadListMutex must be
released after resetting the VM safePointState in order to prevent
newly-created threads from incorrectly setting the halted at safe point
bit.

The vmThreadListMutex is held for the entirety of the safe point, and it
is also required while creating a new thread, so new threads will not be
created during the safe point.  Releasing the vmThreadListMutex early
allows the thread creation code to proceed, where it examines the
safePointState and incorrectly set the halt bit which will never be
cleared, causing VM access acquisition to block forever.

### DIFF
--- a/runtime/vm/VMAccess.cpp
+++ b/runtime/vm/VMAccess.cpp
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 1991, 2017 IBM Corp. and others
+ * Copyright (c) 1991, 2018 IBM Corp. and others
  *
  * This program and the accompanying materials are made available under
  * the terms of the Eclipse Public License 2.0 which accompanies this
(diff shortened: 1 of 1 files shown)


DanHeidinga:
Jenkins test sanity plinux

DanHeidinga:
It looks like the PR build passed:
```
'Build finished. 1013 tests run, 351 skipped, 0 failed.'
```
But I'll give it one more chance to actually become a checkmark...

Jenkins test sanity plinux

pshipton:
> It looks like the PR build passed

The thing to note is that there should be ~21152 tests passed in the PR build. The previous test run aborted early and didn't complete the test run. Its unfortunate that message shows up, I've opened #981.

PR #982, opened 2018-01-19
Initial draft of release process for OpenJ9

DanHeidinga:

### DIFF
--- a/doc/processes/release_process.md
+++ b/doc/processes/release_process.md
@@ -29,7 +29,7 @@ The OpenJ9 JVM is composed of two code repositories:
 
 OpenJ9 needs to be combined with code from OpenJDK to create a Java SDK.
 The OpenJDK code is managed in separate github repos referred to as the 
-"OpenJDK extensions for OpenJ9 for JDK X".  Currently there are two 
+"Extensions for OpenJDKx for OpenJ9".  Currently there are two 
 extensions repos:
 
 * JDK8 [https://github.com/ibmruntimes/openj9-openjdk-jdk8]()
@@ -48,6 +48,8 @@ ensure that releases can be rebuilt or branched later.
 binaries at the correct levels have been created. 
 * Clearly define the supported platforms
 * Should not regress performance from release to release.
+* Use Github releases to identify releases and link to the relevant
+data.
 
 
 ## Release cadence
@@ -64,7 +66,7 @@ functional and stress, and 3rd party applications that should pass for
 every release.
 
 ## Platform support
-Define the platforms that are supported by this release.  The set of 
+Define the platforms that are supported by a release.  The set of 
 platforms will depend on the availability of hardware at OpenJ9 and 
 AdoptOpenJDK that can be used to validate the platform meets the quality 
 bar.  The code base may support more platforms than can be validated by 
@@ -72,28 +74,39 @@ the quality.  Such platforms will not be officially included in the release.
 
 ## Performance
 OpenJ9 should avoid regressing performance from release to release.
-This is harder to validate in the open project given the limitted 
-hardware resources available.  Work is being done at AdoptOpenJDK to 
-enable performance testing.
+Each release should document the set of benchmarks and performance measures
+that will be tracked for the release.  These measures should include the
+acceptable noise margins that will be used when tracking the measure.
+Regressions may be unavoidable at times due to changes required to be compliant
+with the Java specifications, to close security holes, or to address the
+need to update / modify the architecture.  Some regressions may simply be 
+found too late in a release to address them.  
+The project may decide to trade performance - whether throughput, startup 
+or fooprint - for another important measure. 
+Any regressions should be documented as part of the release plan.
+
+Note, detecting regressions will be difficult in the open project 
+given the limited hardware resources available.  Work is being done at 
+AdoptOpenJDK to enable performance testing.
 
 # Release Process
 
 1. Create an issue to track the release and record the information required
 above, including the SHAs and tags.  This information will mostly be duplicated
 with the Eclipse release plan.
-1. Ask committers - via Slack and mailing list -  to refrain from committing any
-further changes until an "all clear" signal is broadcast.  Disable the promotion
-from the `openj9-omr` `master` branch to the `openj9` branch.
 1. Pick a candidate OpenJ9 level to mark as the initial release candidate.  This
 should be a level that has successfully passed the nightly testing builds at 
-AdoptOpenJDK.
-1. Tag the `openj9` & `openj9-omr` repos at the specified level with a tag of the
-following form: `openj9-#releaseNumber#RC#candidatenumber#`.  For the `0.8` 
-release, this would result in a `openj9-0.8RC1` tag.
-1. Tag the Extensions repos at the levels that passed the AdoptOpenJDK builds with
-the same tag.
-1. Sound the "all-clear". The OpenJ9 `master` branch and `openj9-omr` promotion 
-can be reopened for pull requests at this point.
+both OpenJ9 and AdoptOpenJDK. Given these builds may target a range of SHAs, a
+candidate SHA should be taken from within that range.
+1. Branch the `openj9` & `openj9-omr` repos at the specified level.  Immediately
+tag the newly created branch with a tag of the following form: 
+`openj9-#releaseNumber#RC#candidatenumber#`.  For the `0.8` release, this would 
+result in a `openj9-0.8RC1` tag.  These branches are not intended as development
+branches and no continued support will be done on them.  They are merely points
+in time.
+1. The Extensions repos should have been branched for each of the releases.
+Update the Extensions branches to pull the tagged levels from the `openj9` 
+& `openj9-omr` release branches.
 1. Rebuild the tagged levels and ensure they pass the quality bar defined by 
 AdoptOpenJDK.
 1. Provide a window of time (a week?) for any stakeholders to highlight any 
@@ -104,6 +117,9 @@ determination can be made to either:
 	changes that have gone in between the initial RC tag and now are safe.
 1. Retag the `RCx` level as `openj9-#releaseNumber#`.  For the `0.8` release this 
 will be `openj9-0.8`.
+1. Create the [https://help.github.com/articles/creating-releases/](github release)
+corresponding to the tagged level.  The release should link to the Eclipse Release 
+document, the release issue, and the AdoptOpenJDK download links.
 
 The `java -version` should now show the tagged level.
 
--- a/doc/processes/release_process.md
+++ b/doc/processes/release_process.md
@@ -98,7 +98,8 @@ with the Eclipse release plan.
 should be a level that has successfully passed the nightly testing builds at 
 both OpenJ9 and AdoptOpenJDK. Given these builds may target a range of SHAs, a
 candidate SHA should be taken from within that range.
-1. Branch the `openj9` & `openj9-omr` repos at the specified level.  Immediately
+1. Branch the `openj9` & `openj9-omr` repos at the specified level.  Branch names
+should be of the form: `openj9-#releaseNumber#`.  Immediately
 tag the newly created branch with a tag of the following form: 
 `openj9-#releaseNumber#RC#candidatenumber#`.  For the `0.8` release, this would 
 result in a `openj9-0.8RC1` tag.  These branches are not intended as development
@@ -112,14 +113,16 @@ AdoptOpenJDK.
 1. Provide a window of time (a week?) for any stakeholders to highlight any 
 stopship issues with the release candidate build.  If any are found, a 
 determination can be made to either:
-	* Branch from the the `RCx` tag and apply a targetted fix, or
-	* Create the next release candidate build from the master stream if the 
+	* Apply a targetted fix to the release branch and retag, or
+	* Rebase the release branch on the master branch if the 
 	changes that have gone in between the initial RC tag and now are safe.
 1. Retag the `RCx` level as `openj9-#releaseNumber#`.  For the `0.8` release this 
 will be `openj9-0.8`.
 1. Create the [https://help.github.com/articles/creating-releases/](github release)
 corresponding to the tagged level.  The release should link to the Eclipse Release 
 document, the release issue, and the AdoptOpenJDK download links.
+1. Open an Eclipse Bugzilla requesting the branch be marked `protected` to prevent 
+commits after the release is complete.
 
 The `java -version` should now show the tagged level.
 
--- a/doc/processes/release_process.md
+++ b/doc/processes/release_process.md
@@ -1,0 +1,109 @@
+<!--
+Copyright (c) 2018, 2018 IBM Corp. and others
+
+This program and the accompanying materials are made available under
+the terms of the Eclipse Public License 2.0 which accompanies this
+distribution and is available at https://www.eclipse.org/legal/epl-2.0/
+or the Apache License, Version 2.0 which accompanies this distribution and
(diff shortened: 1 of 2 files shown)


doc/processes/release_process.md
```diff
@@ -0,0 +1,109 @@
...
+
+## Platform support
+Define the platforms that are supported by this release.  The set of 
+platforms will depend on the availability of hardware at OpenJ9 and 
+AdoptOpenJDK that can be used to validate the platform meets the quality 
+bar.  The code base may support more platforms than can be validated by 
+the quality.  Such platforms will not be officially included in the release.
+
+## Performance
+OpenJ9 should avoid regressing performance from release to release.
```
andrewcraik: This is a good goal to have. There are times where a regression may not be identified late in the process or that the project has taken a decision to pay some performance for a new feature/capability. We need a way to document these decisions and account for them in testing. Sometimes things get slower before they get faster.

We also need to keep in mind performance is a multi-aspect thing which can have competing demands - the project may choose to trade start-up for throughput, throughput for start-up, and then there are the myriad of memory usage and compile-time metrics.

At the very least we need to have tolerances for noise and other such things. Probably an area we need to flesh out in more details before we can say we have a fully defined 'plan'.

doc/processes/release_process.md
```diff
@@ -0,0 +1,109 @@
...
+hardware resources available.  Work is being done at AdoptOpenJDK to 
+enable performance testing.
+
+# Release Process
+
+1. Create an issue to track the release and record the information required
+above, including the SHAs and tags.  This information will mostly be duplicated
+with the Eclipse release plan.
+1. Ask committers - via Slack and mailing list -  to refrain from committing any
+further changes until an "all clear" signal is broadcast.  Disable the promotion
```
andrewcraik: How are we going to make sure the freezes are of short duration? The danger with freezing the codebase is that PR queues grow and the community's development efforts stall while we release. Since we can branch (as the process described below details) why do we need a freeze when history can't change?

doc/processes/release_process.md
```diff
@@ -0,0 +1,109 @@
...
+bar.  The code base may support more platforms than can be validated by 
+the quality.  Such platforms will not be officially included in the release.
+
+## Performance
+OpenJ9 should avoid regressing performance from release to release.
+This is harder to validate in the open project given the limitted 
+hardware resources available.  Work is being done at AdoptOpenJDK to 
+enable performance testing.
+
+# Release Process
```
fjeremic: I sent an email about this to the mailing list but only one response by @mstoodle was recieved. Quoting it here for visibility:

> Has anyone proposed the idea of having "GitHub releases" (https://github.com/eclipse/openj9/releases) which correspond to official Eclipse releases? i.e. I would love to have a technical overview of exactly what went into a release. From a developers point of view the things developers may care about is:
> 
> - Release Notes
> - Notable Issues fixed
> - What commit of all the repositories that go into Eclipse OpenJ9 a particular release is based off
> 	- For example if I am looking for whether a particular Pull Request for an Issue fix that is relevant to me is in a particular release I would need to know the commit SHAs
> 
> A very good example of this can be seen on the .NET Core project:
> https://github.com/dotnet/coreclr/releases
> 
> Each release is tagged to a particular git commit SHA (you can click on it on the left). It has versioning numbers. GitHub conveniently diffs the git commits between this version and the last automatically. Each commit corresponding to a version is git tagged (https://github.com/dotnet/coreclr/tags). All tags have release notes which can point to a Markdown document right inside the repository (https://github.com/dotnet/core/blob/master/release-notes/2.0/2.0.3.md). These Markdown documents detail all release notes including linking to the official announcement and release (this can be the Eclipse release). It lists important information such as security considerations, notable issues fixed, etc. 
> 
> This can also be a great place to list Acknowledgements to all the contributors in this release so it recognizes the developers working on OpenJ9 giving contributors another thing to be proud about. And of course we can link directly to the AdoptOpenJDK build which corresponds to this particular release, as well as docker images, etc.
> 
> In essence the GitHub release would have more technical information and link to the official Eclipse release which can contain more highlevel content non-developers may care about.

Since we are already hosted on GitHub why not take advantage of features it provides for release management out of the box as users will expect to see such information on the repository releases page (https://github.com/eclipse/openj9/releases).

doc/processes/release_process.md
```diff
@@ -0,0 +1,109 @@
...
+1. Tag the `openj9` & `openj9-omr` repos at the specified level with a tag of the
+following form: `openj9-#releaseNumber#RC#candidatenumber#`.  For the `0.8` 
+release, this would result in a `openj9-0.8RC1` tag.
+1. Tag the Extensions repos at the levels that passed the AdoptOpenJDK builds with
+the same tag.
+1. Sound the "all-clear". The OpenJ9 `master` branch and `openj9-omr` promotion 
+can be reopened for pull requests at this point.
+1. Rebuild the tagged levels and ensure they pass the quality bar defined by 
+AdoptOpenJDK.
+1. Provide a window of time (a week?) for any stakeholders to highlight any
```
mstoodle: I think a week is too short. I think we should be requesting feedback on milestone builds and splitting a release branch early.  Maybe the 2nd monthly milestone build? Even that represents only a month for everyone to run their tests, decide which ones are likely to be OpenJ9's "fault", to report them as issues with all relevant details, and then for us to diagnose and fix them.

I think we should branch early so that stakeholders have more time to report problems before we actually release it, without A) being sabotaged by any ongoing development work (which would otherwise encourage stakeholders to test *late* in our release cycle and encourage us not to fix our stakeholder's problems in the current release), or B) requiring everyone to step up in a very short period of time (which will encourage people not to test, will give us a false sense of quality, and ultimately drives users away).

I think the "early" release branch is also important so that the project can continue to operate even while a release is happening (something @andrewcraik mentioned). If someone wants to deliver something into the release in that last month, then we need to put extra vigiliance against it and the developer pays the cost to put it into two releases (git is pretty helpful here, actually). That additional delivery cost encourages developers to contribute features early rather than late (or to hold off to the next cycle), which is probably a good pressure from a project quality standpoint.

I think the "continuous" delivery model is an excellent aspiration for this project, and I support it as a direction. But I don't want our goals for continuous delivery to become a bar that blocks users from engaging with us because we set a high bar for their engagement. OpenJDK sets a lower bar here (I'm not throwing stones at all) in that they fork their repositories from the "mainline" pretty early which gives stakeholders a long time to soak and report problems.

If some stakeholders can afford to test us all the time, we welcome them with open arms, but let's not make it table stakes, at least not right out of the gate.

doc/processes/release_process.md
```diff
@@ -0,0 +1,109 @@
...
+should be a level that has successfully passed the nightly testing builds at 
+AdoptOpenJDK.
+1. Tag the `openj9` & `openj9-omr` repos at the specified level with a tag of the
+following form: `openj9-#releaseNumber#RC#candidatenumber#`.  For the `0.8` 
+release, this would result in a `openj9-0.8RC1` tag.
+1. Tag the Extensions repos at the levels that passed the AdoptOpenJDK builds with
+the same tag.
+1. Sound the "all-clear". The OpenJ9 `master` branch and `openj9-omr` promotion 
+can be reopened for pull requests at this point.
+1. Rebuild the tagged levels and ensure they pass the quality bar defined by
```
mstoodle: I suggest jenkins jobs get created to do this part. Then they can be faithfully repeated whenever needed (particularly if we could be making changes on branches, which I know we don't want to do regularly, but I really doubt we'll be able to completely avoid it).

doc/processes/release_process.md
mstoodle: I also see little value in freezing the code just to create some branches (or tags).

I despise code freezes.

doc/processes/release_process.md
```diff
@@ -0,0 +1,109 @@
...
+* JDK8 [https://github.com/ibmruntimes/openj9-openjdk-jdk8]()
+* JDK9 [https://github.com/ibmruntimes/openj9-openjdk-jdk9]()
+
+OpenJ9 binaries are built by the AdoptOpenJDK community.  
+
+An OpenJ9 release should:
+
+* Use a single code base to support the various JDK-levels.
+* Track the upstream OpenJDK releases.  
+* Use a consistent set of tags across OpenJ9 & extensions repos to
```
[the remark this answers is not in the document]
mstoodle: Ok, i can see one value of the tag moving along the branch: if we commit something that's not actually very good, we don't have to move the tag to the tip of the branch without having to revert on the branch. I still think it's a good idea to branch though.

doc/processes/release_process.md
DanHeidinga: Thanks for the feedback.   I agree parts of this are aspirational and that there may be times when a regression is necessary - either late discovery of the cause or due to necessary bug fix.

Starting from a goal of "no performance regressions", each release should document a set of measures that can the project will track and ensure don't regress.  If the measures do regress, then the reason should be called out in release issue.

Sound reasonable?

andrewcraik: I think that sounds fine though I think 'no performance regressions' does need to include a noise margin - variations in benchmarks mean that on some benchmarks regressions less than a certain % are not considered significant. Given the difficulty of running perf testing as well as finding and fixing issues found from perf testing having the wording here such that we admit and accept (as policy) that there are tolerances for regressions and that those tolerances are often non-zero. Setting tolerances etc is a test specific thing, but I think codifying that we expect them to exist and that setting them is part of tests being added to release criteria and that they are subject to periodic review would seem the most prudent course to me based on past experience with perf measurement.

doc/processes/release_process.md
DanHeidinga: I too despise code freezes.  My thinking here was there are a number of projects and repos involved and having a short window where the world isn't changing under our feet may make it easier to get through the process, especially if something goes wrong.

Today, it's 4 repos that need to be coordinated and it may not be immediately clear which SHAs in each repo should be used for the release.  As far as I am aware, there is no guarantee that the Adopt nightly builds have run on the exact same commit for all elements in the matrix of JDK releases x platforms.

Having a window where nothing goes in makes it possible to ensure that Adopt is building the same code on all releases x platforms to give the initial set of SHAs

doc/processes/release_process.md
DanHeidinga: I think this is a great idea.

doc/processes/release_process.md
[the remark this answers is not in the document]
DanHeidinga: @mstoodle Is your suggestion to branch at the targeted point and then lay the `openj9-#releaseNumber#RC#candidatenumber#` tag on the branch?  Or are you expecting to just keep building out of the branch?  

I think a tag is a minimum requirement to be able to differentiate between different release candidates or builds from that branch.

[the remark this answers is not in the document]
mstoodle: Fair point. I wasn't thinking far enough ahead. I still like the idea of creating a branch so you can just grab the latest version for that release without having to know which tag you're after. I don't think I like moving tags.

DanHeidinga:
Pushed an update that addresses the following:
```
Remove code freeze
Use tagged branches
Update the performance section
Call out github releases
```

DanHeidinga:
@mstoodle @andrewcraik @fjeremic Any further comments?

doc/processes/release_process.md
```diff
@@ -0,0 +1,125 @@
...
+in time.
+1. The Extensions repos should have been branched for each of the releases.
+Update the Extensions branches to pull the tagged levels from the `openj9` 
+& `openj9-omr` release branches.
+1. Rebuild the tagged levels and ensure they pass the quality bar defined by 
+AdoptOpenJDK.
+1. Provide a window of time (a week?) for any stakeholders to highlight any 
+stopship issues with the release candidate build.  If any are found, a 
+determination can be made to either:
+	* Branch from the the `RCx` tag and apply a targetted fix, or
```
AdamBrousseau: Do we have to branch from the RCx branch or can we simply apply another change to the release branch and lay tag RC(x+1)

doc/processes/release_process.md
```diff
@@ -0,0 +1,125 @@
...
+	* Create the next release candidate build from the master stream if the 
+	changes that have gone in between the initial RC tag and now are safe.
+1. Retag the `RCx` level as `openj9-#releaseNumber#`.  For the `0.8` release this 
+will be `openj9-0.8`.
```
AdamBrousseau: Will the RC tags be removed once the final tag is laid?

doc/processes/release_process.md
```diff
@@ -0,0 +1,125 @@
...
+1. Create the [https://help.github.com/articles/creating-releases/](github release)
+corresponding to the tagged level.  The release should link to the Eclipse Release 
+document, the release issue, and the AdoptOpenJDK download links.
+
+The `java -version` should now show the tagged level.
```
AdamBrousseau: Do we have a mechanism that will adjust the `java -version` or are you simply stating that the SHAs will match the release tags?

doc/processes/release_process.md
```diff
@@ -0,0 +1,125 @@
...
+determination can be made to either:
+	* Branch from the the `RCx` tag and apply a targetted fix, or
+	* Create the next release candidate build from the master stream if the 
+	changes that have gone in between the initial RC tag and now are safe.
+1. Retag the `RCx` level as `openj9-#releaseNumber#`.  For the `0.8` release this 
+will be `openj9-0.8`.
+1. Create the [https://help.github.com/articles/creating-releases/](github release)
+corresponding to the tagged level.  The release should link to the Eclipse Release 
+document, the release issue, and the AdoptOpenJDK download links.
+
```
AdamBrousseau: Will we want to protect the release branches on Github? We will need to open an Eclipse Bugzilla if we want to do this.
Also, we could enable the option to block commits to the release branches after they are released

doc/processes/release_process.md
```diff
@@ -0,0 +1,125 @@
...
+1. Pick a candidate OpenJ9 level to mark as the initial release candidate.  This
+should be a level that has successfully passed the nightly testing builds at 
+both OpenJ9 and AdoptOpenJDK. Given these builds may target a range of SHAs, a
+candidate SHA should be taken from within that range.
+1. Branch the `openj9` & `openj9-omr` repos at the specified level.  Immediately
+tag the newly created branch with a tag of the following form: 
+`openj9-#releaseNumber#RC#candidatenumber#`.  For the `0.8` release, this would 
+result in a `openj9-0.8RC1` tag.  These branches are not intended as development
+branches and no continued support will be done on them.  They are merely points
+in time.
```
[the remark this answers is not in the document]
DanHeidinga: The branch name should be the same as the release tag, minus the `RC#candidatenumber#` part.

doc/processes/release_process.md
```diff
@@ -0,0 +1,125 @@
...
+1. The Extensions repos should have been branched for each of the releases.
```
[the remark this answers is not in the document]
DanHeidinga: My understanding from Simon / Andrew is that the Extensions repos will be branched to match the OpenJDK release branches.  So we'll tag their existing branches.

doc/processes/release_process.md
DanHeidinga: I don't see a benefit to removing the RC tags.  They may be helpful if we ever need to do a retrospective on any release issues.

doc/processes/release_process.md
DanHeidinga: Great idea.

doc/processes/release_process.md
andrewcraik: I like the revisions to this section - my concerns have been addressed - thank!

DanHeidinga:
I think this is ready to merge.   Can one of the reviewers with commit rights merge this?

PR #983, opened 2018-01-19
Protect orphan ROM class table with mutex

gacholio:
Ensure that all accesses to orphan ROM class tables in class loaders are
protected by the classTableMutex.

### DIFF
--- a/runtime/bcutil/defineclass.c
+++ b/runtime/bcutil/defineclass.c
@@ -115,7 +115,7 @@ internalDefineClass(
 	}
 
 	if (!isAnonFlagSet) {
-		/* See if there's already an orphan romClass available */
+		/* See if there's already an orphan romClass available - still own classTableMutex at this point */
 		if (NULL != classLoader->romClassOrphansHashTable) {
 			orphanROMClass = romClassHashTableFind(classLoader->romClassOrphansHashTable, className, classNameLength);
 			if (NULL != orphanROMClass) {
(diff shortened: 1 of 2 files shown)


gacholio:
Jenkins test sanity plinux

runtime/bcutil/defineclass.c
```diff
@@ -186,11 +189,19 @@ internalDefineClass(
...
 			}
 		} else if (NULL != orphanROMClass) {
 			/* ramClass creation succeeded - the orphanROMClass is no longer an orphan */
 			Trc_BCU_romClassOrphansHashTableDelete(vmThread, classNameLength, className, orphanROMClass);
-			romClassHashTableDelete(classLoader->romClassOrphansHashTable, orphanROMClass);
+			/* All access to the orphan table must be done while holding classTableMutex.
+			 * Ensure the entry is still in the table before removing it.
+			 */
+			omrthread_monitor_enter(vm->classTableMutex);
+			if (romClassHashTableFind(classLoader->romClassOrphansHashTable, className, classNameLength) == orphanROMClass) {
```
DanHeidinga: Should there be a tracepoint for the case where the orphan wasn't in the orphan's table?  It would give us something to trigger on if this unlikely (impossible) situation occurred.... though writing that makes me think assert rather than tracepoint

gacholio: Tracepoint added.  Assert makes me a little leary as it's possible to disable the parallel class loading table via a command-line option.

DanHeidinga:
The travis.ci build ensures this compiles and PR build passed before the tracepoint change.  I'm comfortable merging as is.

Issue #859, opened 2017-12-20
Access denied ("java.lang.RuntimePermission" "accessClassInPackage.sun.nio.cs.ext")


pshipton:
@aguibert mentioned a problem running Liberty on OpenJ9 Java 9

```
>Stack Dump = java.lang.ExceptionInInitializerError
>    at java.lang.J9VMInternals.ensureError(java.base@9-internal/J9VMInternals.java:191)
>    at java.lang.J9VMInternals.recordInitializationFailure(java.base@9-internal/J9VMInternals.java:180)
>    at java.nio.charset.Charset.lookupExtendedCharset(java.base@9-internal/Charset.java:456)
>    at java.nio.charset.Charset.lookup2(java.base@9-internal/Charset.java:486)
>    at java.nio.charset.Charset.lookup(java.base@9-internal/Charset.java:474)
>    at java.nio.charset.Charset.forName(java.base@9-internal/Charset.java:538)
>    at com.sun.org.apache.xml.internal.serializer.Encodings$EncodingInfos.findCharsetNameFor(java.xml@9-internal/Encodings.java:402)
>    at com.sun.org.apache.xml.internal.serializer.Encodings$EncodingInfos.findCharsetNameFor(java.xml@9-internal/Encodings.java:438)
>    at com.sun.org.apache.xml.internal.serializer.Encodings$EncodingInfos.loadEncodingInfo(java.xml@9-internal/Encodings.java:466)
>    at com.sun.org.apache.xml.internal.serializer.Encodings$EncodingInfos.<init>(java.xml@9-internal/Encodings.java:324)
>    at com.sun.org.apache.xml.internal.serializer.Encodings$EncodingInfos.<init>(java.xml@9-internal/Encodings.java:312)
>    at com.sun.org.apache.xml.internal.serializer.Encodings.<clinit>(java.xml@9-internal/Encodings.java:580)
>    at com.sun.org.apache.xml.internal.serializer.ToStream.<init>(java.xml@9-internal/ToStream.java:135)
>    at com.sun.org.apache.xml.internal.serializer.ToXMLStream.<init>(java.xml@9-internal/ToXMLStream.java:65)
>    at com.sun.org.apache.xalan.internal.xsltc.runtime.output.TransletOutputHandlerFactory.getSerializationHandler(java.xml@9-internal/TransletOutputHandlerFactory.java:163)
[...]

PR #984, opened 2018-01-19
Coalesce module lookups in createRamClass

tajila:
Coalesce module lookups in createRamClass

Combine module lookups in createRamClass.c. Move up the lookup for
module with findModuleForPackage() earlier in the code so that the
module info can be used to do access checking for superclasses and
interfaces (this will come in another PR).

Reviewer: @pshipton

### DIFF
--- a/runtime/vm/createramclass.cpp
+++ b/runtime/vm/createramclass.cpp
@@ -152,7 +152,7 @@ static J9Class* internalCreateRAMClassDone(J9VMThread *vmThread, J9ClassLoader *
 	J9UTF8 *className, J9CreateRAMClassState *state);
 static J9Class* internalCreateRAMClassFromROMClassImpl(J9VMThread *vmThread, J9ClassLoader *classLoader, J9ROMClass *romClass, UDATA options, J9Class *elementClass,
 	J9ROMMethod **methodRemapArray, IDATA entryIndex, I_32 locationType, J9Class *classBeingRedefined, UDATA packageID, J9Class *superclass, J9CreateRAMClassState *state,
-	J9ClassLoader* hostClassLoader, J9Class *hostClass);
+	J9ClassLoader* hostClassLoader, J9Class *hostClass, J9Module *module);
 static J9MemorySegment* internalAllocateRAMClass(J9JavaVM *javaVM, J9ClassLoader *classLoader, RAMClassAllocationRequest *allocationRequests, UDATA allocationRequestCount);
 static I_32 interfaceDepthCompare(const void *a, const void *b);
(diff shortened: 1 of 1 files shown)


runtime/vm/createramclass.cpp
```diff
@@ -2826,6 +2777,47 @@ internalCreateRAMClassFromROMClass(J9VMThread *vmThread, J9ClassLoader *classLoa
 	/* To prevent deadlock, release the classTableMutex before loading the classes required for the new class. */
 	omrthread_monitor_exit(javaVM->classTableMutex);
 
+	pkgNameLength = packageNameLength(romClass);
```
pshipton: This should be declared and initialized in the block where it is used.

pshipton:
jenkins test sanity plinux

JasonFengJ9:
A changeset I am working on to address #859 is using `ramclass->module` to determine if package access check for superclasses and interfaces should proceed or not.

tajila:
@JasonFengJ9 We might be working on a similar issue. We are currently failing JCKs where a class is incorrectly loading a superclass.  
  
Here is the existing code:
```
			/* ensure that the superclass is visible */
			if (!isROMClassUnsafe) {
				if ((superclass->romClass->modifiers & J9_JAVA_PUBLIC) != J9_JAVA_PUBLIC) {
					if (packageID != superclass->packageID) {
						Trc_VM_CreateRAMClassFromROMClass_superclassNotVisible(vmThread, superclass, superclass->classLoader, classLoader);
						setCurrentExceptionForBadClass(vmThread, superclassName, J9VMCONSTANTPOOL_JAVALANGILLEGALACCESSERROR);
						return FALSE;
					}
				}
			}
```
To really make sure the class is visible we need to to module access checks. 

This PR is refactoring the code to make it easier to do access checks over there. Is that the change you were intending to make?

JasonFengJ9:
@tajila The issue #859 seems due to unnecessary package access check when resolving the classes loading from same module. To make a complete solution, I have to address the classes being resolved, super classes and interface classes loading. When dealing with super/interface classes, I have similar need for `module` access, so your PR is going to help mine though some rebasing work is required.
It appears the JCK failure you are working on was caused by an incorrect super class loading (i.e., shouldn't allowed but proceed), so the issues we are working on need different PR solution but both need `ramclass->module`. It is nice to see you submit this common framework PR first.

tajila:
@pshipton I'm investigating the errors, will post again once its ready

tajila:
@pshipton the changes are ready for another sanity test

pshipton:
sanity testing was done internal to IBM

PR #994, opened 2018-01-22
Allow opt level downgrades during grace period

mpirvu:
This commit changes the existing heuristics for when the optimization level
of compilations is changed from 'warm' to 'cold'. Previously, such downgrades
were not allowed during the "grace period" (first second) to preserve behavior
of short running benchmarks. Recent experiments showed that we could further
improve startup of applications if we lift that restriction.
The new heuristic can be disabled with a new option:
  -Xjit:dontDowngradeToColdDuringGracePeriod

### DIFF
--- a/runtime/compiler/trj9/control/CompilationThread.cpp
+++ b/runtime/compiler/trj9/control/CompilationThread.cpp
@@ -474,17 +474,12 @@ bool TR::CompilationInfo::shouldDowngradeCompReq(TR_MethodToBeCompiled *entry)
          }
       else
          {
-         // Is it grace-period?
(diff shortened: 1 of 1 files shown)


mpirvu:
Dependent on eclipse/omr#2245

fjeremic:
@vijaysun-omr assigning this one to you as you are the committer on the OMR dependent PR.

fjeremic:
Jenkins test sanity depends eclipse/omr#2245

mpirvu:
jenkins test sanity

mpirvu:
@vijaysun-omr The dependent OMR change has been integrated, so this PR is ready to be merged.

PR #998, opened 2018-01-22
Acceleration for x86 compressed String hashCode with vectorization

msqb:
For obtaining the hashCode of a compressed String, each iteration of the SSE loop moves 4 bytes from memory to the low doubleword of the XMM register. The stride shift amount is set to 0 for handling bytes. These packed 1 byte integers are then zero extended with instruction pmovzxbd and the same procedure to obtain the hashCode as with decompressed Strings is used.

A vector conversion example has been added to show how this procedure works.

### DIFF
--- a/runtime/compiler/x/codegen/J9TreeEvaluator.cpp
+++ b/runtime/compiler/x/codegen/J9TreeEvaluator.cpp
@@ -11765,14 +11765,63 @@ static bool inlineObjectHashCode(TR::Node *node, bool isIndirect)
 
 // Convert serial String.hashCode computation into vectorization copy and implement with SSE instruction
 //
+// Conversion process example:
+//
+//    str[8] = example string representing 8 characters (compressed or decompressed)
+//
+//    The serial method for creating the hash:
+//          hash = 0, offset = 0, count = 8
+//          for (int i = offset; i < offset+count; ++i) {
+//                hash = (hash << 5) - hash + str[i];
+//          }
+//
+//    Note that ((hash << 5) - hash) is equivalent to hash * 31
+//
+//    Expanding out the for loop:
+//          hash = ((((((((0*31+str[0])*31+str[1])*31+str[2])*31+str[3])*31+str[4])*31+str[5])*31+str[6])*31+str[7])
+//
+//    Simplified:
+//          hash =        (31^7)*str[0] + (31^6)*str[1] + (31^5)*str[2] + (31^4)*str[3]
+//                      + (31^3)*str[4] + (31^2)*str[5] + (31^1)*str[6] + (31^0)*str[7]
+//
+//    Rearranged:
+//          hash =        (31^7)*str[0] + (31^3)*str[4]
+//                      + (31^6)*str[1] + (31^2)*str[5]
+//                      + (31^5)*str[2] + (31^1)*str[6]
+//                      + (31^4)*str[3] + (31^0)*str[7]
+//
+//    Factor out [31^3, 31^2, 31^1, 31^0]:
+//          hash =        31^3*((31^4)*str[0] + str[4])           Vector[0]
+//                      + 31^2*((31^4)*str[1] + str[5])           Vector[1]
+//                      + 31^1*((31^4)*str[2] + str[6])           Vector[2]
+//                      + 31^0*((31^4)*str[3] + str[7])           Vector[3]
+//
+//    Keep factoring out any 31^4 if possible (this example has no such case). If the string was 12 characters long then:
+//          31^3*((31^8)*str[0] + (31^4)*str[4] + (31^0)*str[8]) would become 31^3*(31^4((31^4)*str[0] + str[4]) + (31^0)*str[8])
+//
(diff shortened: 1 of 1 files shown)


msqb:
Requires upstream OMR pull request [Adds x86 instruction PMOVZXBD](https://github.com/eclipse/omr/pull/2246). @andrewcraik

andrewcraik:
Some comment on the performance evaluation of this change is warranted since this is a performance optimization/improvement - I believe @msqb is working on distilling his results to be posted here shortly.

runtime/compiler/trj9/x/codegen/J9TreeEvaluator.cpp
```diff
@@ -11940,32 +11990,51 @@ inlineStringHashCode(
...
 
       generateLabelInstruction(LABEL, node, SSELabel, cg);
-      generateRegMemInstruction(MOVSDRegMem,
-                                node,
-                                xmm2,
-                                generateX86MemoryReference(valueReg, indexReg, 1, TR::Compiler->om.contiguousArrayHeaderSizeInBytes(), cg),
-                                cg);
+
+      // move 8 bytes (4 2-byte characters) to xmm2. Stride shift = 1 (each element in valueReg is 2 bytes. Used with indexReg for correct traversal)
+      if(!isCompressed)
```
fjeremic: This if/else has both paths. Would it be possible to turn this into a positive query and avoid the negation?

msqb: @fjeremic consulting OMR's compiler/z/codegen's inlineStringHashCode as a reference, it had the same type of negation if/else structure so I used that format. However if you'd like I can change it to become a positive query.

https://github.com/eclipse/omr/blob/ec2c3bf81b14c6ee01f5e03e8a35a730682ebb6a/compiler/z/codegen/OMRTreeEvaluator.cpp#L17252

andrewcraik: I think avoiding the negation makes sense for new code - there are certainly dark areas in the existing code worthy of clean up but if we make things a quantum better every time a change is made the overall quality of everything moves up over time.

msqb: sure, I'll make the change

fjeremic: > However if you'd like I can change it to become a positive query.

Definitely fix up both while you're in this area of code. Thanks!

msqb:
@fjeremic replaced negations with positive query

msqb:
Performance analysis was conducted using a microbenchmark using strings of lengths 3, 12, and 48 characters long composed of English lowercase letters. For each of these length variations, String's hashCode() method was repeatedly called on 10,000 unique strings with steps taken to prevent cached hashcodes from being used as part of the measurement.

Previously when enabling string compression, String's hashcode was constructed serially. Comparing the current serial string hashCode() implementation for compressed strings to the vectorized implementation in this changeset, all three string lengths showed a 60% performance improvement when running OpenJ9 OpenJDK9 with string compression enabled. OpenJ9 OpenJDK8 showed a similar performance improvement with string compression enabled.

The performance of hashCode() with string compression disabled remained unchanged for both OpenJ9 OpenJDK9 and OpenJ9 OpenJDK8.

OpenJ9 OpenJDK9 and OpenJ9 OpenJDK8 used for this analysis were compiled from:
OpenJ9   - d7cc8fd
OMR       - 9bc3259

andrewcraik:
@msqb this seems to have been lost in the churn - could you please rebase forward and I can then launch the appropriate sanity builds to try and get this merged?

andrewcraik:
@msqb could you please rebase this forward so we can get it tested?

andrewcraik:
@0dvictor could you let me know your thoughts on this change?

0dvictor:
The code looks good to me. It is a straightforward extension of current hash code implementation of uncompressed strings.
FYI, https://github.com/eclipse/openj9/pull/1647 has a better implementation then current one, and it supports both compressed and uncompressed strings.

andrewcraik:
@0dvictor so is your suggestion that this should be closed?

andrewcraik:
Jenkins test sanity

andrewcraik:
Jenkins test sanity xlinux jdk8,jdk9,jdk10

andrewcraik:

msqb:
I've rebased forward. Thanks

andrewcraik:

andrewcraik:

andrewcraik:
@msqb so we've had a bit of build instability - can you rebase forward so we can finally try to clear the x86 sanity tests to get this in?

msqb:
@andrewcraik I've rebased forward. Thanks

andrewcraik:

andrewcraik:
x86 Linux is completely clean failures on windows look to be related to #2129

PR #1002, opened 2018-01-23
Improve path length for String equalsIgnoreCase

yanluo7:
- Compare the last character first as a shortcut to tell two Strings are not equal.
- When string compression is active, only need to convert both chars to
  uppercase and compare once, as all characters are within the ASCII char set.
- When string compression is not active, if either character is not in ASCII
  char set, we have to convert to upper and lower case and compare twice.

Signed-off-by: Yan Luo <yan_luo@ca.ibm.com>

### DIFF
--- a/jcl/src/java.base/share/classes/java/lang/String.java
+++ b/jcl/src/java.base/share/classes/java/lang/String.java
@@ -2364,6 +2364,11 @@ public final class String implements Serializable, Comparable<String>, CharSeque
 			return false;
 		}
 
+		// Zero length strings are equal
+		if (0 == s1len) {
+			return true;
+		}
+
 		int o1 = 0;
 		int o2 = 0;
 
@@ -2379,7 +2384,20 @@ public final class String implements Serializable, Comparable<String>, CharSeque
 		/*[ENDIF]*/
 
 		if (enableCompression && (null == compressionFlag || (s1.count | s2.count) >= 0)) {
-			while (o1 < end) {
+			// Compare the last chars.
+			// In order to tell 2 chars are different:
+			// Under string compression, the compressible char set obeys 1-1 mapping for upper/lower case,
+			// converting to lower cases then compare should be sufficient.
+			byte byteAtO1Last = helpers.getByteFromArrayByIndex(s1Value, s1len-1);
+			byte byteAtO2Last = helpers.getByteFromArrayByIndex(s2Value, s1len-1);
+
+			if (byteAtO1Last != byteAtO2Last
+					&& toUpperCase(helpers.byteToCharUnsigned(byteAtO1Last)) != toUpperCase(helpers.byteToCharUnsigned(byteAtO2Last))) {
+				return false;
+			}
+
+
+			while (o1 < end-1) {
 				byte byteAtO1 = helpers.getByteFromArrayByIndex(s1Value, o1++);
 				byte byteAtO2 = helpers.getByteFromArrayByIndex(s2Value, o2++);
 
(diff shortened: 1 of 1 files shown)


yanluo7:
Performance implications:
Instrumentation has shown for calls to `String.equalsIgnoreCase`, among cases where the returned results are not equal, vast majority of them are that the last two chars are also not equal. On open liberty ACMEAIR benchmark, the percentage is 80%, while for SPECj, the percentage is 90%. So we create a shortcut here and compare last char directly, as we do for String equals case.

In general, to compare two code points while ignoring cases, we need to convert them both to upper case and compare, and convert both to lower cases and compare; only when both comparisons return false (i.e. chars are different) can we tell the two characters are indeed not equal. The reason seems to be certain language includes a 3rd case in addition to the upper and lower cases, and their conversion rule creates a non-unique mapping between lower-upper case (one lower case can have more than one "upper" cases). But if we know the comparison involves an ASCII char where the case conversion is unique one-to-one mapping and both cases fall in ASCII char set as well, only one comparison is necessary. 

Using microbenchmark hunblebench, we measure the performance of equal case (meaning we need to compare every char in the String) to improve by 2%, the performance of not equal case (meaning comparing only the last char vs looping the char array to find the first char that are different) to improve by 30%.

jcl/src/java.base/share/classes/java/lang/String.java
```diff
@@ -2379,7 +2384,20 @@ public boolean equalsIgnoreCase(String string) {
 		/*[ENDIF]*/
 
 		if (enableCompression && (null == compressionFlag || (s1.count | s2.count) >= 0)) {
-			while (o1 < end) {
+			// Compare the last chars.
+			// In order to tell 2 chars are different:
+			// Under string compression, the compresible char set obeys 1-1 mapping for upper/lower case,
```
pshipton: typo - `compresible`

jcl/src/java.base/share/classes/java/lang/String.java
```diff
@@ -2388,17 +2406,31 @@ public boolean equalsIgnoreCase(String string) {
...
+			// Compare the last chars.
+			// In order to tell 2 chars are different:
+			// If at least one char is ASCII, converting to upper cases then compare should be sufficient.
+			// If both chars are not in ASCII char set, need to convert to lower case and compare as well.
+			char charAtO1Last = s1.charAtInternal(s1len-1, s1Value);
+			char charAtO2Last = s2.charAtInternal(s1len-1, s2Value);
+
+			if (charAtO1Last != charAtO2Last 
+					&& toUpperCase(charAtO1Last) != toUpperCase(charAtO2Last)
+					&& (charAtO1Last < 128 || charAtO2Last < 128 || Character.toLowerCase(charAtO1Last) != Character.toLowerCase(charAtO2Last))) {
```
pshipton: Since compressed characters (the first 256) don't need a lower case check, this check can at least use 256.

jcl/src/java.base/share/classes/java/lang/String.java
```diff
@@ -2388,17 +2406,31 @@ public boolean equalsIgnoreCase(String string) {
...
+			}
+
+			while (o1 < end-1) {
 				char charAtO1 = s1.charAtInternal(o1++, s1Value);
 				char charAtO2 = s2.charAtInternal(o2++, s2Value);
 
-				if (charAtO1 != charAtO2 && toUpperCase(charAtO1) != toUpperCase(charAtO2) && toLowerCase(charAtO1) != toLowerCase(charAtO2)) {
+				if (charAtO1 != charAtO2 
+						&& toUpperCase(charAtO1) != toUpperCase(charAtO2) 
+						&& (charAtO1 < 128 || charAtO2 < 128 || Character.toLowerCase(charAtO1) != Character.toLowerCase(charAtO2))) {
```
pshipton: Similarly to the above, this can use 256

jcl/src/java.base/share/classes/java/lang/String.java
pshipton: According to my testing, 304 would also work.

runtime/gc_glue_java/ObjectModel.hpp
```diff
@@ -1,6 +1,6 @@
 
 /*******************************************************************************
```
pshipton: There are some extra classes added to this PR now.

jcl/src/java.base/share/classes/java/lang/String.java
yanluo7: Fixed.

runtime/gc_glue_java/ObjectModel.hpp
yanluo7: Fixed the extraneous files.

jcl/src/java.base/share/classes/java/lang/String.java
yanluo7: I'm fine with extending the check to 304 (and possibly further). But if the chars between 256 and 304 are not often used, I prefer to leave at 256 since it's more clear and 304 seems a little random.

jcl/src/java.base/share/classes/java/lang/String.java
yanluo7: Fixed.

jcl/src/java.base/share/classes/java/lang/String.java
pshipton: I guess it depends what language you are using and what characters you need. I won't push for 304. I wanted to double check the first 256 and I extended my test to see how far it could go before finding a problem, 304 was the first character that needed the toLowerCase() check.

jcl/src/java.base/share/classes/java/lang/String.java
```diff
@@ -2388,17 +2406,31 @@ public boolean equalsIgnoreCase(String string) {
...
+			// Compare the last chars.
+			// In order to tell 2 chars are different:
+			// If at least one char is ASCII, converting to upper cases then compare should be sufficient.
+			// If both chars are not in ASCII char set, need to convert to lower case and compare as well.
+			char charAtO1Last = s1.charAtInternal(s1len-1, s1Value);
+			char charAtO2Last = s2.charAtInternal(s1len-1, s2Value);
+
+			if (charAtO1Last != charAtO2Last 
+					&& toUpperCase(charAtO1Last) != toUpperCase(charAtO2Last)
+					&& (charAtO1Last < 256 || charAtO2Last < 256 || Character.toLowerCase(charAtO1Last) != Character.toLowerCase(charAtO2Last))) {
```
pshipton: I think this test is wrong. Both characters need to be < 256.

yanluo7: Great catch. You're right, the first upper-lower case mappings that're non-unique happens to step on ASCII character (i), at char (304) as you pointed out. So this check needs to be more strict to limit both char to ASCII, which guarantees 1-1 mapping.

yanluo7:
jam.29 personal build on Linux x86 level promotion sanity promotionhttp://vmfarm.ottawa.ibm.com/build_info.php?build_id=376870

pshipton:
> jvm.29 personal build level promotion regression sanity.

Generally we try to keep IBM internal links out of the public Issues and PRs

Issue #189, opened 2017-09-29
java.base exports com.ibm.cuda and com.ibm.gpu


pshipton:
The java.base module exports packages com.ibm.cuda and com.ibm.gpu which is violation of the java 9 specification. These packages need to be moved and exported from the existing module stubs openj9.cuda and openj9.gpu

PR #1008, opened 2018-01-24
Move com.ibm.le.conditionhandling outside java.base

theresa-m:
- Move com.ibm.le.conditionhandling outside java.base to openj9.zosconditionhandling

Reviewer: @JasonFengJ9


runtime/vm/gphandle.c
```diff
@@ -347,7 +347,14 @@ throwConditionException(J9VMThread *vmThread, struct ConditionExceptionConstruct
 
 	cls = (*env)->FindClass(env, "com/ibm/le/conditionhandling/ConditionException");
 	if (NULL == cls) {
-		goto cleanup;
+		/* Since this class belongs to a module other than java.base, load another type of
+		 * exception to provide more information about what went wrong.
+		 */
+		cls = (*env)->FindClass(env, "java/lang/ClassNotFoundException");
```
DanHeidinga: I'm not convinced using CNFE is a good alternative to ConditionException. Code that expects to catch ConditionException will fail as CNFE won't match the `catch` clause.

Also, the `cls` is used to lookup an `<init>` method with the signature `"(Ljava/lang/String;J[BLjava/lang/String;IIIIIJ)V"` which I'd be surprised to find on CNFE.

DanHeidinga: We may need to find a way to have the `openj9.zosconditionhandling` register this class with the VM so we can use it from native code.

jcl/src/java.base/share/classes/module-info.java.extra
```diff
@@ -40,7 +39,7 @@ module java.base {
   uses com.ibm.gpu.spi.GPUAssist.Provider;
   exports com.ibm.gpu.spi to openj9.gpu;
   // TODO remove the following after they move to their new modules
-  exports com.ibm.cuda;
-  exports com.ibm.gpu;
+  exports com.ibm.cuda to openj9.cuda;
```
pshipton: These packages should be moved to their respective modules. Adding these qualified exports is going to hide these packages that should be exported unconditionally. I suggest dealing with cuda and gpu is better left to another PR.

pshipton: i.e. #189

runtime/vm/gphandle.c
pshipton: CNFE seems reasonable when ConditionException is not available, as the problem would be that the openj9.zosconditionhandling module is missing from the runtime. However it should include a message like "com.ibm.le.conditionhandling.ConditionException not found, include the openj9.zosconditionhandling module". 

I don't think code that catches ConditionException would even be able to load in the absence of ConditionException. I tried a simple testcase and it throws a ClassNotFoundException when a caught exception is missing, although the exception was never thrown.

```
public class TestMissingException {
	
	public static void test() throws Throwable {
	}

		try {
			test();
		} catch (TestException e) {
}
}
```

DanHeidinga: I don't have a problem with throwing a CNFE if its due to not finding the `ConditionException` class.  If that's the case, the original `FindClass()` JNI call should leave a pending exception along with its null return value.

Bottom line: We don't need to explicitly write this code as there will already be a pending exception.

theresa-m: okay. I'll remove it.

jcl/src/java.base/share/classes/module-info.java.extra
theresa-m: okay.

JasonFengJ9:
The commit comment (also in PR description) is not accurate.
Pls change it to "Move `com.ibm.le.conditionhandling` outside `java.base` to `openj9.zosconditionhandling`" cause packages exported without `qualified` are accessible to every module including `unnamed modules`.
Other than that, looks good to me.

PR #1010, opened 2018-01-24
Add OpenJDK Git SHA to Version String

adamfarley:
Currently, we add the OpenJDK SHA to the version string by
manually appending the SHA to the -version output. 

This code allows us to populate the -version output with 
the OpenJDK SHA, *and* remove the non-closed extensions 
currently used for that purpose.

One step closer to no extensions at all.

depends on
ibmruntimes/openj9-openjdk-jdk8#34
ibmruntimes/openj9-openjdk-jdk9#111

Fixes: #1011

Signed-off-by: Adam Farley <adam.farley@uk.ibm.com>

### DIFF
--- a/runtime/jcl/common/jclcinit.c
+++ b/runtime/jcl/common/jclcinit.c
@@ -41,6 +41,9 @@
 #include "omrversionstrings.h"
 #include "j9modron.h"
 #include "omr.h"
+#if defined(OPENJ9_BUILD)
+#include "openj9_version_info.h"
+#endif
 #include "vendor_version.h"
 
 /* The vm version which must match the JCL.
@@ -146,6 +149,12 @@ jint computeFullVersionString(J9JavaVM* vm)
 	#define OMR_INFO ""
 #endif /* J9VM_GC_MODRON_GC */
 
+#if defined(OPENJDK_TAG) && defined(OPENJDK_SHA)
+	#define OPENJDK_INFO "\nJCL      - " OPENJDK_SHA " based on " OPENJDK_TAG
+#else
+	#define OPENJDK_INFO ""
+#endif /* OPENJDK_TAG && OPENJDK_SHA */
+
 #if defined(VENDOR_SHORT_NAME) && defined(VENDOR_SHA)
 	#define VENDOR_INFO "\n" VENDOR_SHORT_NAME "      - " VENDOR_SHA
 #else
@@ -153,7 +162,7 @@ jint computeFullVersionString(J9JavaVM* vm)
 #endif /* VENDOR_SHORT_NAME && VENDOR_SHA */
 
 	if (BUFFER_SIZE <= j9str_printf(PORTLIB, fullversion, BUFFER_SIZE + 1,
-			"JRE %s IBM J9 %s %s %s" MEM_INFO "%s" JIT_INFO J9VM_VERSION_STRING OMR_INFO VENDOR_INFO,
+			"JRE %s IBM J9 %s %s %s" MEM_INFO "%s" JIT_INFO J9VM_VERSION_STRING OMR_INFO VENDOR_INFO OPENJDK_INFO,
 			j2se_version_info,
 			EsVersionString,
 			(NULL != osname ? osname : " "),
@@ -166,7 +175,7 @@ jint computeFullVersionString(J9JavaVM* vm)
 	}
 
 	if (BUFFER_SIZE <= j9str_printf(PORTLIB, vminfo, BUFFER_SIZE + 1,
-			"JRE %s %s %s" MEM_INFO "%s" JIT_INFO J9VM_VERSION_STRING OMR_INFO VENDOR_INFO,
+			"JRE %s %s %s" MEM_INFO "%s" JIT_INFO J9VM_VERSION_STRING OMR_INFO VENDOR_INFO OPENJDK_INFO,
 			j2se_version_info,
 			(NULL != osname ? osname : " "),
 			osarch,
--- a/runtime/jcl/cl_se10/module.xml
+++ b/runtime/jcl/cl_se10/module.xml
@@ -105,6 +105,7 @@
 			<makefilestub data="UMA_ENABLE_ALL_WARNINGS=1"/>
 			<makefilestub data="UMA_TREAT_WARNINGS_AS_ERRORS=1"/>
 			<makefilestub data="ifeq ($(OPENJ9_BUILD),true)"/>
+			<makefilestub data="jclcinit$(UMA_DOT_O) : CFLAGS += -DOPENJ9_BUILD"/>
 			<makefilestub data="vm_scar$(UMA_DOT_O) : CFLAGS += -DOPENJ9_BUILD"/>
 			<makefilestub data="endif"/>
 		</makefilestubs>
--- a/runtime/jcl/cl_se7_basic/module.xml
+++ b/runtime/jcl/cl_se7_basic/module.xml
@@ -105,6 +105,7 @@
 			<makefilestub data="UMA_ENABLE_ALL_WARNINGS=1"/>
 			<makefilestub data="UMA_TREAT_WARNINGS_AS_ERRORS=1"/>
 			<makefilestub data="ifeq ($(OPENJ9_BUILD),true)"/>
+			<makefilestub data="jclcinit$(UMA_DOT_O) : CFLAGS += -DOPENJ9_BUILD"/>
 			<makefilestub data="vm_scar$(UMA_DOT_O) : CFLAGS += -DOPENJ9_BUILD"/>
 			<makefilestub data="endif"/>
 		</makefilestubs>
--- a/runtime/jcl/cl_se9/module.xml
+++ b/runtime/jcl/cl_se9/module.xml
@@ -105,6 +105,7 @@
 			<makefilestub data="UMA_ENABLE_ALL_WARNINGS=1"/>
 			<makefilestub data="UMA_TREAT_WARNINGS_AS_ERRORS=1"/>
 			<makefilestub data="ifeq ($(OPENJ9_BUILD),true)"/>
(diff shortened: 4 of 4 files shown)


runtime/jcl/common/jclcinit.c
```diff
@@ -42,6 +42,7 @@
 #include "j9modron.h"
 #include "omr.h"
 #include "vendor_version.h"
+#include "openj9_version_info.h"
```
keithc-ca: This will break our (old) b148 builds because openj9_version_info.h doesn't exist in that world.

pshipton: To fix it, see the following example in runtime/util/module.xml (i.e. add something similar to `runtime/jcl/cl_se*/module.xml` where `*` is 7_basic, 9, 10) which allows use of `#if defined(OPENJ9_BUILD)`

```
			<!--Special compilation for openj9_version.c with extra version flags. -->
			<makefilestub data="ifeq ($(OPENJ9_BUILD),true)"/>
			<makefilestub data="openj9_version$(UMA_DOT_O) : CFLAGS += -DOPENJ9_BUILD"/>
			<makefilestub data="endif"/>
```

and

pshipton: Also, since we allow vendor_version.h to override the default values, the include for vendor_version.h should come after openj9_version_info.h

pshipton:
This change would result in the system property java.vm.info containing the `OpenJDK <sha> based on <tag>` twice.

https://github.com/ibmruntimes/openj9-openjdk-jdk9/blob/30a5439fc0b01586b2a8c419edb6dff46e6c42e4/jdk/src/java.base/share/classes/java/lang/VersionProps.java.template#L201

pshipton:
I see, the linked code is what you are trying to remove. Please prepare the Pull Requests to remove this code so it can be delivered at the same time as this OpenJ9 change.

pshipton:
Upon further consideration, this isn't the correct way to fix appending the OpenJDK information. VENDOR_INFO and VENDOR_SHA are intended to be used to add information about an additional repo, such as the IBM internal repo that is used for IBM builds. Re-using these variables to add OpenJDK information won't allow IBM to add this repo information.

You can use `#if defined(OPENJ9_BUILD)` to append a new line for OpenJDK using the OPENJDK_SHA and OPENJDK_TAG defines. Please remove the definitions of VENDOR_SHORT_NAME and VENDOR_SHA from the extension repos.

pshipton:
At the same time, we might as well add the OPENJDK information to the javacore. See the vendor example https://github.com/eclipse/openj9/blob/1d6c78422de14f37afbb9f4d2ba4a13491749463/runtime/rasdump/javadump.cpp#L999

adamfarley:
All done.

Javadump has the SHA, the includes are conditional, and the VENDOR variable use has been removed.

I can start getting people to merge the extensions once this has been reviewed, approved, and merged.

runtime/jcl/common/jclcinit.c
```diff
@@ -146,14 +149,20 @@ jint computeFullVersionString(J9JavaVM* vm)
 	#define OMR_INFO ""
 #endif /* J9VM_GC_MODRON_GC */
 
+#if defined(OPENJDK_TAG) && defined(OPENJDK_SHA)
+	#define OPENJDK_INFO "\nOpenJDK  - " OPENJDK_SHA " based on " OPENJDK_TAG
```
pshipton: From an IBM perspective, I don't think the "OpenJDK" should be hard coded, but passed in as another define. For IBM internal builds it should be configurable since I expect they would use "Oracle" instead of "OpenJDK".

Also, based on the -version discussion on Slack today I'm not sure we need or want to include the OPENJDK_TAG along with the SHA, as this piece of information should be elsewhere in the -version output.

runtime/jcl/common/jclcinit.c
```diff
@@ -146,14 +149,20 @@ jint computeFullVersionString(J9JavaVM* vm)
...
+
 #if defined(VENDOR_SHORT_NAME) && defined(VENDOR_SHA)
 	#define VENDOR_INFO "\n" VENDOR_SHORT_NAME "      - " VENDOR_SHA
 #else
 	#define VENDOR_INFO ""
 #endif /* VENDOR_SHORT_NAME && VENDOR_SHA */
 
 	if (BUFFER_SIZE <= j9str_printf(PORTLIB, fullversion, BUFFER_SIZE + 1,
-			"JRE %s IBM J9 %s %s %s" MEM_INFO "%s" JIT_INFO J9VM_VERSION_STRING OMR_INFO VENDOR_INFO,
+			"JRE %s IBM J9 %s %s %s" MEM_INFO "%s" JIT_INFO J9VM_VERSION_STRING OMR_INFO OPENJDK_INFO VENDOR_INFO,
```
pshipton: Please keep the same order and put the OPENJDK_INFO after the VENDOR_INFO

pshipton:
Please include links to the (public) extension changes here. Once this is ready I can merge them all at the same time.

runtime/jcl/common/jclcinit.c
adamfarley: Sure. 1 moment.

adamfarley:
https://github.com/ibmruntimes/openj9-openjdk-jdk8/pull/34
https://github.com/ibmruntimes/openj9-openjdk-jdk9/pull/111

runtime/jcl/common/jclcinit.c
```diff
@@ -41,6 +41,9 @@
 #include "omrversionstrings.h"
 #include "j9modron.h"
 #include "omr.h"
+#if defined(OPENJ9_BUILD)
```
pshipton: The required change to set OPENJ9_BUILD isn't added to the module.xml's so I don't think it will ever be defined.

runtime/rasdump/javadump.cpp
```diff
@@ -48,6 +48,9 @@
 #include "omr.h"
 #include "omrutilbase.h"
 #include "j9version.h"
+#if defined(OPENJ9_BUILD)
```
pshipton: The required change to set OPENJ9_BUILD isn't added to the module.xml so I don't think it will ever be defined.

runtime/jcl/common/jclcinit.c
adamfarley: I will freely admit that I'm not clear on how or why that variable is only set on some of our builds and not others. Any advice?

runtime/rasdump/javadump.cpp
```diff
@@ -994,6 +997,11 @@ JavaCoreDumpWriter::writeEnvironmentSection(void)
 	_OutputStream.writeCharacters(_VirtualMachine->memoryManagerFunctions->omrgc_get_version(_VirtualMachine->omrVM));
 	_OutputStream.writeCharacters("\n");
 
+#if defined(OPENJDK_SHA) && defined(OPENJDK_TAG)
+	/* Write the OpenJDK Git SHA and build tag */
+	_OutputStream.writeCharacters("1CIOPENJDK " OPENJDK_SHA " based on " OPENJDK_TAG "\n");
```
pshipton: Similarly to the other comment about hard coding OpenJDK, I don't think it should be hard coded here. We could pick a more generic code, like "JDKEXT", or something else.

pshipton: Another suggestion is "JCL"

runtime/jcl/common/jclcinit.c
pshipton: I'd be ok with hard coding "JCL".

adamfarley:
The javacore additions have been removed, and a new issue will be raised to handle them.

runtime/jcl/common/jclcinit.c
pshipton: To fix it, see the following example copied from runtime/jcl/cl_se7_basic/module.xml which allows use of `#if defined(OPENJ9_BUILD)`. Add another line for jclcinit. All of jcl_se7_base, cl_se9, cl_se10 need to be changed .

```
			<makefilestub data="ifeq ($(OPENJ9_BUILD),true)"/>
			<makefilestub data="vm_scar$(UMA_DOT_O) : CFLAGS += -DOPENJ9_BUILD"/>
			<makefilestub data="endif"/>
```

pshipton: You'll need to do the similar thing in runtime/rasdump/module.xml for javadump.cpp

adamfarley: Ahh, I see. Took me a minute to figure out vm_scar was a file. Doh!

adamfarley: Done. I've also added notes to the javadump issue.

pshipton:
This change is not working yet. One problem is the extensions are not yet properly defining OPENJDK_SHA and OPENJDK_TAG, but there seems to be other problem as well since the JCL line isn't showing up in the output at all. I'm not sure why.

adamfarley:
I think it's because I didn't update the master branches with the first set of extensions changes before creating the branches for the second set. My mistake. Again. :(

pshipton:
Ok makes sense. Actually the -version in the travis build contains the JCL line we expect, the only problem being the missing OPENJDK_TAG

openjdk version "9-internal"
OpenJDK Runtime Environment (build 9-internal+0-adhoc.travis.openj9-openjdk-jdk9)
Eclipse OpenJ9 VM (build 2.9, JRE 9 Linux amd64-64 Compressed References 20180125_000000 (JIT enabled, AOT enabled)
OpenJ9   - ad5411b2
OMR      - 4badb2f4
JCL      - 5314f37d based on 
OpenJDK  - 5314f37d based on )

adamfarley:
Ok, I think it looks good new. Are we good to merge?

pshipton:
Waiting for the extensions changes to be squashed

adamfarley:
Done.

Issue #786, opened 2017-12-08
Remove java.logging from j9jcl's java.management classes


adamfarley:
If we can remove the "requires java.logging" from the j9jcl java.management's module-info.java.extra file (along with the need for it), then we won't need the Extensions changes to GenModuleInfoSource.java anymore.

Additionally:

I explored the possibility of contributing the GenModuleInfoSource.java changes to the OpenJDK community. They responded by reiterating that the code is intended to be used only for platform-specific module-info.java changes, and that they agreed at the file's creation that module dependencies should not change between platforms. So this functionality should not be here. This leads me to conclude that is is OpenJ9 that should change, and Dan Heidinga also thinks this.

DanHeidinga:
Links to the OpenJDK discussion about @adamfarley's proposed patch.  Please note these are direct links to OJDK mailing lists:
http://mail.openjdk.java.net/pipermail/jigsaw-dev/2017-December/013403.html
http://mail.openjdk.java.net/pipermail/core-libs-dev/2017-December/050364.html

Given the discussion from OJDK that they've deliberately avoided adding this capability, OJ9 should look at other ways to access "java.logging".

Maybe using Class.forName or equivalent to look up the classes at runtime?  More investigation required.

DanHeidinga:
Marking this as medium priority as it reduces the # of patches required by OpenJ9.

DanHeidinga:
@theresa-m Can you take a look at this?

adamfarley:
Thanks @theresa-m :)

PR #1019, opened 2018-01-25
Remove java.logging requires from java.management

theresa-m:
Use reflection instead of module-info export.

Fixes: #786

Reviewers: @tajila @DanHeidinga

### DIFF
--- a/jcl/src/java.management/share/classes/com/ibm/java/lang/management/internal/LoggingMXBeanImpl.java
+++ b/jcl/src/java.management/share/classes/com/ibm/java/lang/management/internal/LoggingMXBeanImpl.java
@@ -1,6 +1,6 @@
 /*[INCLUDE-IF Sidecar17]*/
 /*******************************************************************************
- * Copyright (c) 2005, 2017 IBM Corp. and others
+ * Copyright (c) 2005, 2018 IBM Corp. and others
  *
  * This program and the accompanying materials are made available under
  * the terms of the Eclipse Public License 2.0 which accompanies this
@@ -23,14 +23,25 @@
 package com.ibm.java.lang.management.internal;
 
 import java.lang.management.PlatformLoggingMXBean;
+/*[IF Sidecar19-SE]*/
+/*[IF Sidecar19-SE-B165]*/
+import java.lang.Module;
+import java.lang.ModuleLayer;
+/*[ELSE]
+import java.lang.reflect.Module;
+import java.lang.reflect.Layer;
+/*[ENDIF]*/
+/*[ENDIF]*/
+import java.lang.reflect.InvocationTargetException;
+import java.lang.reflect.Method;
 import java.util.ArrayList;
 import java.util.Enumeration;
 import java.util.List;
+/*[IF !Sidecar19-SE]
+import java.util.logging.LoggingMXBean;
 import java.util.logging.Level;
 import java.util.logging.LogManager;
 import java.util.logging.Logger;
-/*[IF !Sidecar19-SE]
-import java.util.logging.LoggingMXBean;
 /*[ENDIF]*/
 
 import javax.management.ObjectName;
@@ -49,8 +60,53 @@ public final class LoggingMXBeanImpl
 /*[ENDIF]*/
 			PlatformLoggingMXBean {
 
-	private static final LoggingMXBeanImpl instance = new LoggingMXBeanImpl();
+	private static final LoggingMXBeanImpl instance;
 
+/*[IF Sidecar19-SE]*/
+	private static final Method logManager_getLogManager;
+	private static final Method logManager_getLogger;
+	private static final Method logManager_getLoggerNames;
+	private static final Method logger_getLevel;
+	private static final Method logger_getParent;
+	private static final Method logger_getName;
+	private static final Method logger_setLevel;
+	private static final Method level_getName;
+	private static final Method level_parse;
+	private static final String logManager_LOGGING_MXBEAN_NAME;
+/*[ENDIF]*/	
+
+	static {
+/*[IF Sidecar19-SE]*/
+		try {
+
+/*[IF Sidecar19-SE-B165]*/
+			Module java_logging = ModuleLayer.boot().findModule("java.logging").get(); //$NON-NLS-1$
+/*[ELSE]
+			Module java_logging = Layer.boot().findModule("java.logging").get(); //$NON-NLS-1$
+/*[ENDIF]*/			
+			Class<?> logManagerClass = Class.forName(java_logging, "java.util.logging.LogManager"); //$NON-NLS-1$
+			Class<?> loggerClass = Class.forName(java_logging, "java.util.logging.Logger"); //$NON-NLS-1$
+			Class<?> levelClass = Class.forName(java_logging, "java.util.logging.Level"); //$NON-NLS-1$
+			logManager_getLogManager = logManagerClass.getMethod("getLogManager"); //$NON-NLS-1$
+			logManager_getLogger = logManagerClass.getMethod("getLogger", String.class); //$NON-NLS-1$
+			logManager_getLoggerNames = logManagerClass.getMethod("getLoggerNames"); //$NON-NLS-1$
+			logger_getLevel = loggerClass.getMethod("getLevel"); //$NON-NLS-1$
+			logger_getParent = loggerClass.getMethod("getParent"); //$NON-NLS-1$
+			logger_getName = loggerClass.getMethod("getName"); //$NON-NLS-1$
+			logger_setLevel = loggerClass.getMethod("setLevel", levelClass); //$NON-NLS-1$
+			level_getName = levelClass.getMethod("getName"); //$NON-NLS-1$
+			level_parse = levelClass.getMethod("parse", String.class); //$NON-NLS-1$
+			logManager_LOGGING_MXBEAN_NAME = logManagerClass.getField("LOGGING_MXBEAN_NAME").get(null).toString(); //$NON-NLS-1$
+
+		} catch (Exception e) {
+			throw handleError(e);
+		}
+/*[ENDIF]*/		
+		
+		/* Initialize class */
+		instance = new LoggingMXBeanImpl();
+	}
+	
 	/**
 	 * the object name
 	 */
@@ -62,7 +118,11 @@ public final class LoggingMXBeanImpl
 	 */
 	private LoggingMXBeanImpl() {
 		super(PlatformLoggingMXBean.class, true);
+/*[IF Sidecar19-SE]*/
+		objectName = ManagementUtils.createObjectName(logManager_LOGGING_MXBEAN_NAME);
+/*[ELSE]		
 		objectName = ManagementUtils.createObjectName(LogManager.LOGGING_MXBEAN_NAME);
+/*[ENDIF]*/
 	}
 
 	/**
@@ -88,20 +148,40 @@ public final class LoggingMXBeanImpl
 	@Override
 	public String getLoggerLevel(String loggerName) {
 		String result = null;
-		Logger logger = LogManager.getLogManager().getLogger(loggerName);
 
-		if (logger != null) {
-			// The named Logger exists. Now attempt to obtain its log level.
-			Level level = logger.getLevel();
-			if (level != null) {
-				result = level.getName();
-			} else {
-				// A null return from getLevel() means that the Logger
-				// is inheriting its log level from an ancestor. Return an
-				// empty string to the caller.
-				result = ""; //$NON-NLS-1$
+/*[IF Sidecar19-SE]*/
+		try {
+			Object logger = getLoggerFromName(loggerName);
+/*[ELSE]	
+			Logger logger = LogManager.getLogManager().getLogger(loggerName);
+/*[ENDIF]*/			
+			
+			if (logger != null) {
+				// The named Logger exists. Now attempt to obtain its log level.
+/*[IF Sidecar19-SE]*/
+				Object level = logger_getLevel.invoke(logger);
+/*[ELSE]					
+				Level level = logger.getLevel();
+/*[ENDIF]*/					
+				if (level != null) {
+/*[IF Sidecar19-SE]*/
+					result = (String)level_getName.invoke(level);
+/*[ELSE]	
+					result = level.getName();
+/*[ENDIF]*/	
+				} else {
+					// A null return from getLevel() means that the Logger
+					// is inheriting its log level from an ancestor. Return an
+					// empty string to the caller.
+					result = ""; //$NON-NLS-1$
+				}
 			}
+/*[IF Sidecar19-SE]*/
+		} catch (Exception e) {
+			throw handleError(e);
 		}
+/*[ENDIF]*/
+		
 		return result;
 	}
 
@@ -112,11 +192,21 @@ public final class LoggingMXBeanImpl
 	public List<String> getLoggerNames() {
 		// By default, return an empty list to caller
 		List<String> result = new ArrayList<>();
-		Enumeration<String> enumeration = LogManager.getLogManager().getLoggerNames();
+		Enumeration<?> enumeration;
 
+/*[IF Sidecar19-SE]*/
+		try {
+			Object logManagerInstance = logManager_getLogManager.invoke(null);
+			enumeration = (Enumeration<?>) logManager_getLoggerNames.invoke(logManagerInstance);
+		} catch (Exception e) {
+			throw handleError(e);
+		}
+/*[ELSE]	
+		enumeration = LogManager.getLogManager().getLoggerNames();
+/*[ENDIF]*/	
 		if (enumeration != null) {
 			while (enumeration.hasMoreElements()) {
-				result.add(enumeration.nextElement());
+				result.add((String)enumeration.nextElement());
 			}
 		}
 
@@ -129,20 +219,38 @@ public final class LoggingMXBeanImpl
 	@Override
 	public String getParentLoggerName(String loggerName) {
 		String result = null;
-		Logger logger = LogManager.getLogManager().getLogger(loggerName);
 
-		if (logger != null) {
-			// The named Logger exists. Now attempt to obtain its parent.
-			Logger parent = logger.getParent();
-			if (parent != null) {
-				// There is a parent
-				result = parent.getName();
-			} else {
-				// logger must be the root Logger
-				result = ""; //$NON-NLS-1$
+/*[IF Sidecar19-SE]*/
+		try {
+			Object logger = getLoggerFromName(loggerName);
+/*[ELSE]		
+			Logger logger = LogManager.getLogManager().getLogger(loggerName);
+/*[ENDIF]*/
+			if (logger != null) {
+				// The named Logger exists. Now attempt to obtain its parent.
+/*[IF Sidecar19-SE]*/
+				Object parent = logger_getParent.invoke(logger);
+/*[ELSE]	
+				Logger parent = logger.getParent();
+/*[ENDIF]*/
+				if (parent != null) {
+					// There is a parent
+/*[IF Sidecar19-SE]*/
+					result = (String)logger_getName.invoke(parent);
+/*[ELSE]	
+					result = parent.getName();
+/*[ENDIF]*/
+				} else {
+					// logger must be the root Logger
+					result = ""; //$NON-NLS-1$
+				}
 			}
+/*[IF Sidecar19-SE]*/
+		} catch (Exception e) {
+			throw handleError(e);
 		}
-
+/*[ENDIF]*/
+		
 		return result;
 	}
 
@@ -151,15 +259,34 @@ public final class LoggingMXBeanImpl
 	 */
 	@Override
 	public void setLoggerLevel(String loggerName, String levelName) {
+/*[IF Sidecar19-SE]*/
+		final Object logger;
+		try {
+			logger = getLoggerFromName(loggerName);
+		} catch (Exception e) {
+			throw handleError(e);
+		}
+/*[ELSE]	
 		final Logger logger = LogManager.getLogManager().getLogger(loggerName);
-
+/*[ENDIF]*/	
+		
 		if (logger != null) {
 			// The named Logger exists. Now attempt to set its level. The
 			// below attempt to parse a Level from the supplied levelName
 			// will throw an IllegalArgumentException if levelName is not
 			// a valid level name.
+/*[IF Sidecar19-SE]*/
+			try {
+				Object newLevel = level_parse.invoke(null, levelName);
+				logger_setLevel.invoke(logger, newLevel);
+			} catch (Exception e) {
+				throw handleError(e);
+			}
+/*[ELSE]	
 			Level newLevel = Level.parse(levelName);
 			logger.setLevel(newLevel);
+/*[ENDIF]*/
+			
 		} else {
 			// Named Logger does not exist.
 			/*[MSG "K05E7", "Unable to find Logger with name {0}."]*/
@@ -167,4 +294,34 @@ public final class LoggingMXBeanImpl
 		}
 	}
 
+/*[IF Sidecar19-SE]*/
+	private static Object getLoggerFromName(String loggerName) throws Exception {
+		Object logManagerInstance = logManager_getLogManager.invoke(null);
+		return logManager_getLogger.invoke(logManagerInstance, loggerName);
+	}
+	
+	/* Handle error thrown by method invoke as internal error */
+	private static InternalError handleError(Exception cause) {
+		handleInvocationTargetException(cause);
+		
+		InternalError err = new InternalError(cause.toString());
+		err.initCause(cause);
+		throw err;
+	}
+	
+	/* invoke throws InvocationTargetException if the method it is invoking
+	 * throws an error. This method unwraps that error for this class to maintain
+	 * its specification.
+	 */
+	private static void handleInvocationTargetException(Exception cause) {
+		if (cause instanceof InvocationTargetException) {
+			Throwable wrappedCause = cause.getCause();
+			if (wrappedCause instanceof Error) {
+				throw (Error)wrappedCause;
+			} else if (wrappedCause instanceof RuntimeException) {
+				throw (RuntimeException)wrappedCause;
+			}
+		}
+	}
+/*[ENDIF]*/
 }
--- a/jcl/src/java.management/share/classes/java/lang/management/DefaultPlatformMBeanProvider.java
+++ b/jcl/src/java.management/share/classes/java/lang/management/DefaultPlatformMBeanProvider.java
@@ -1,6 +1,6 @@
 /*[INCLUDE-IF Sidecar19-SE]*/
 /*******************************************************************************
- * Copyright (c) 2016, 2017 IBM Corp. and others
+ * Copyright (c) 2016, 2018 IBM Corp. and others
  *
  * This program and the accompanying materials are made available under
  * the terms of the Eclipse Public License 2.0 which accompanies this
@@ -21,6 +21,12 @@
  * SPDX-License-Identifier: EPL-2.0 OR Apache-2.0 OR GPL-2.0 WITH Classpath-exception-2.0 OR LicenseRef-GPL-2.0 WITH Assembly-exception
  *******************************************************************************/
 package java.lang.management;
+
+/*[IF Sidecar19-SE-B165]*/
+import java.lang.ModuleLayer;
+/*[ELSE]
+import java.lang.reflect.Layer;
+/*[ENDIF]*/
 
 import java.util.ArrayList;
 import java.util.Collections;
@@ -61,9 +67,19 @@ final class DefaultPlatformMBeanProvider extends sun.management.spi.PlatformMBea
 			.addInterface(java.lang.management.OperatingSystemMXBean.class)
 			.register(allComponents);
 
-		ComponentBuilder.create(LoggingMXBeanImpl.getInstance())
-			.addInterface(java.lang.management.PlatformLoggingMXBean.class)
-			.register(allComponents);
+		/* LoggingMXBeanImpl depends on java.logging. If java.logging is not 
+		 * available exclude this component.
+		 */
+/*[IF Sidecar19-SE-B165]*/
+		if (ModuleLayer.boot().findModule("java.logging").isPresent()) //$NON-NLS-1$
+/*[ELSE]
+		if (Layer.boot().findModule("java.logging").isPresent()) //$NON-NLS-1$
+/*[ENDIF]*/
+		{
+			ComponentBuilder.create(LoggingMXBeanImpl.getInstance())
+				.addInterface(java.lang.management.PlatformLoggingMXBean.class)
+				.register(allComponents);
+		}
 
 		ComponentBuilder.create(RuntimeMXBeanImpl.getInstance())
 			.addInterface(java.lang.management.RuntimeMXBean.class)
--- a/jcl/j9modules.txt
+++ b/jcl/j9modules.txt
@@ -32,6 +32,7 @@ openj9.sharedclasses
 openj9.traceformat
 jdk.attach
 openj9.jvm
+java.management
 
 # requires openj9.cuda
 openj9.gpu
@@ -39,7 +40,6 @@ openj9.gpu
 # requires java.logging
 openj9.dtfj
 openj9.dtfjview
-java.management
 
 # requires java.management
 jdk.management
--- a/jcl/src/java.management/share/classes/module-info.java.extra
+++ b/jcl/src/java.management/share/classes/module-info.java.extra
@@ -1,6 +1,6 @@
 /*[INCLUDE-IF Sidecar19-SE]*/
 /*******************************************************************************
- * Copyright (c) 2016, 2017 IBM Corp. and others
+ * Copyright (c) 2016, 2018 IBM Corp. and others
  *
  * This program and the accompanying materials are made available under
  * the terms of the Eclipse Public License 2.0 which accompanies this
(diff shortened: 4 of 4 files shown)


jcl/src/java.management/share/classes/com/ibm/java/lang/management/internal/LoggingMXBeanImpl.java
```diff
@@ -151,20 +206,61 @@ public String getParentLoggerName(String loggerName) {
...
+	/* invoke throws InvocationTargetException if the method it is invoking
+	 *  throws an error. This method unwraps that error for this class to maintain
+	 *  its specification.
+	 */
+	private static void handleInvocationTargetException(Exception cause) {
+		if (cause instanceof InvocationTargetException) {
+			Exception wrappedCause = (Exception)cause.getCause();
+			
+			if (wrappedCause instanceof SecurityException) {
+				throw new SecurityException(wrappedCause.getMessage());
```
DanHeidinga: Why throw a new SecurityException when you have one in hand?  Can't you just do `throw wrappedCause;`?
Same for the NPE and IAE cases.

jcl/src/java.management/share/classes/com/ibm/java/lang/management/internal/LoggingMXBeanImpl.java
```diff
@@ -151,20 +206,61 @@ public String getParentLoggerName(String loggerName) {
...
+	private static InternalError newInternalInvokeError(Exception cause){
+		handleInvocationTargetException(cause);
+		
+		InternalError err = new InternalError(cause.toString());
+		err.initCause(cause);
+		return err;
+	}
+	
+	/* invoke throws InvocationTargetException if the method it is invoking
+	 *  throws an error. This method unwraps that error for this class to maintain
```
DanHeidinga: nitpick: spacing between the starts of the lines here is off

jcl/src/java.management/share/classes/com/ibm/java/lang/management/internal/LoggingMXBeanImpl.java
```diff
@@ -151,20 +206,61 @@ public String getParentLoggerName(String loggerName) {
...
+		return lm_getLogger.invoke(logManagerInstance, loggerName);
+	}
+	
+	/* Handle error thrown by method invoke as internal error */
+	private static InternalError newInternalInvokeError(Exception cause){
+		handleInvocationTargetException(cause);
+		
+		InternalError err = new InternalError(cause.toString());
+		err.initCause(cause);
+		return err;
```
DanHeidinga: this should `throw err`, not return to match the behaviour of `handleInvocationTargetException()` and to ensure the caller can't forgot to throw the Exception

jcl/src/java.management/share/classes/com/ibm/java/lang/management/internal/LoggingMXBeanImpl.java
```diff
@@ -49,8 +48,47 @@
...
+	private static final Method lo_getParent;
+	private static final Method lo_getName;
+	private static final Method lo_setLevel;
+	private static final Method le_getName;
+	private static final Method le_parse;
+	private static final String lm_LOGGING_MXBEAN_NAME;
+	
+	static {
+		try {
+			logManagerClass = Class.forName("java.util.logging.LogManager");
```
DanHeidinga: Shouldn't this be using the `Class.forName(Module, String)` api instead?  The `java.logging` module can be looked up using `ModuleLayer.boot().findModule("java.logging")`?

jcl/src/java.management/share/classes/com/ibm/java/lang/management/internal/LoggingMXBeanImpl.java
```diff
@@ -88,20 +126,26 @@ public static LoggingMXBeanImpl getInstance() {
...
+					result = (String)le_getName.invoke(level);
+				} else {
+					// A null return from getLevel() means that the Logger
+					// is inheriting its log level from an ancestor. Return an
+					// empty string to the caller.
+					result = ""; //$NON-NLS-1$
+				}
 			}
+		} catch (Exception e) {
+			newInternalInvokeError(e);
```
DanHeidinga: Should be `throw newInternalInvokeError(e)` as it makes it obvious from reading this code that execution can't continue past this point without needing to know the implementation of `newInternalInvokeError(e)`

jcl/src/java.management/share/classes/com/ibm/java/lang/management/internal/LoggingMXBeanImpl.java
```diff
@@ -112,7 +156,16 @@ public String getLoggerLevel(String loggerName) {
...
 		List<String> result = new ArrayList<>();
-		Enumeration<String> enumeration = LogManager.getLogManager().getLoggerNames();
+		Enumeration<String> enumeration;
+		Object logManagerInstance;
+		
+		try {
+			logManagerInstance = lm_getLogManager.invoke(null);
+			enumeration = (Enumeration<String>) lm_getLoggerNames.invoke(logManagerInstance);
+		} catch (Exception e) {
+			newInternalInvokeError(e);
```
DanHeidinga: Same here - `throw newInternalInvokeError(e)`

jcl/src/java.management/share/classes/com/ibm/java/lang/management/internal/LoggingMXBeanImpl.java
```diff
@@ -129,18 +182,23 @@ public String getLoggerLevel(String loggerName) {
...
+				if (parent != null) {
+					// There is a parent
+					result = (String)lo_getName.invoke(parent);
+				} else {
+					// logger must be the root Logger
+					result = ""; //$NON-NLS-1$
+				}
 			}
+		} catch (Exception e) {
+			newInternalInvokeError(e);
```
DanHeidinga: `throw newInternalInvokeError(e)`

jcl/src/java.management/share/classes/com/ibm/java/lang/management/internal/LoggingMXBeanImpl.java
```diff
@@ -151,20 +209,62 @@ public String getParentLoggerName(String loggerName) {
 	 */
 	@Override
 	public void setLoggerLevel(String loggerName, String levelName) {
-		final Logger logger = LogManager.getLogManager().getLogger(loggerName);
-
+		final Object logger;
+		try {
+			logger = getLoggerFromName(loggerName);
+		} catch (Exception e) {
+			newInternalInvokeError(e);
```
DanHeidinga: throw....

jcl/src/java.management/share/classes/com/ibm/java/lang/management/internal/LoggingMXBeanImpl.java
```diff
@@ -151,20 +209,62 @@ public String getParentLoggerName(String loggerName) {
...
 			// below attempt to parse a Level from the supplied levelName
 			// will throw an IllegalArgumentException if levelName is not
 			// a valid level name.
-			Level newLevel = Level.parse(levelName);
-			logger.setLevel(newLevel);
+			try {
+				Object newLevel = le_parse.invoke(null, levelName);
+				lo_setLevel.invoke(logger, newLevel);
+			} catch (Exception e) {
+				newInternalInvokeError(e);
```
DanHeidinga: throw...

jcl/src/java.management/share/classes/com/ibm/java/lang/management/internal/LoggingMXBeanImpl.java
```diff
@@ -151,20 +209,62 @@ public String getParentLoggerName(String loggerName) {
...
 	}
-
+	
+	private static Object getLoggerFromName(String loggerName) throws Exception {
+		Object logManagerInstance = lm_getLogManager.invoke(null);
+		return lm_getLogger.invoke(logManagerInstance, loggerName);
+	}
+	
+	/* Handle error thrown by method invoke as internal error */
+	private static void newInternalInvokeError(Exception cause) {
```
DanHeidinga: This should claim to return `InternalError` as it makes it easier to write the obviously correct code.

jcl/src/java.management/share/classes/com/ibm/java/lang/management/internal/LoggingMXBeanImpl.java
```diff
@@ -49,8 +48,47 @@
...
+			levelClass = Class.forName("java.util.logging.Level");
+			lm_getLogManager = logManagerClass.getMethod("getLogManager");
+			lm_getLogger = logManagerClass.getMethod("getLogger", String.class);
+			lm_getLoggerNames = logManagerClass.getMethod("getLoggerNames");
+			lo_getLevel = loggerClass.getMethod("getLevel");
+			lo_getParent = loggerClass.getMethod("getParent");
+			lo_getName = loggerClass.getMethod("getName");
+			lo_setLevel = loggerClass.getMethod("setLevel", levelClass);
+			le_getName = levelClass.getMethod("getName");
+			le_parse = levelClass.getMethod("parse", String.class);
```
DanHeidinga: lm = LogManager.
lo = logger?
le = level?

Can more explicit names be used?  It will be easier on readers when trying to figure out what class is being used.

jcl/src/java.management/share/classes/com/ibm/java/lang/management/internal/LoggingMXBeanImpl.java
```diff
@@ -112,7 +178,15 @@ public String getLoggerLevel(String loggerName) {
 	public List<String> getLoggerNames() {
 		// By default, return an empty list to caller
 		List<String> result = new ArrayList<>();
-		Enumeration<String> enumeration = LogManager.getLogManager().getLoggerNames();
+		Enumeration<String> enumeration;
+		Object logManagerInstance;
```
DanHeidinga: Nitpick: this can be declared inside the try block

jcl/src/java.management/share/classes/com/ibm/java/lang/management/internal/LoggingMXBeanImpl.java
```diff
@@ -49,8 +57,60 @@
 /*[ENDIF]*/
 			PlatformLoggingMXBean {
 
-	private static final LoggingMXBeanImpl instance = new LoggingMXBeanImpl();
+	private static final LoggingMXBeanImpl instance;
+
+/*[IF Sidecar19-SE]
+	private static final Module java_logging;
```
pshipton: Does this code need to be commented out? As much as possible the code should not be in comments to help avoid missing something when the code is changed. This applies to the initialization code as well.

pshipton: @keithc-ca you probably have an opinion on this since you try to keep your workspace clean

theresa-m: Okay thanks. I've seen the directives commenting done a few ways so I wasn't sure what is preferred.

jcl/src/java.management/share/classes/com/ibm/java/lang/management/internal/LoggingMXBeanImpl.java
```diff
@@ -151,20 +230,61 @@ public String getParentLoggerName(String loggerName) {
...
+	
+	/* invoke throws InvocationTargetException if the method it is invoking
+	 * throws an error. This method unwraps that error for this class to maintain
+	 * its specification.
+	 */
+	private static void handleInvocationTargetException(Exception cause) {
+		if (cause instanceof InvocationTargetException) {
+			Exception wrappedCause = (Exception)cause.getCause();
+			
+			if (wrappedCause instanceof SecurityException) {
```
pshipton: This could be more generic and check for RuntimeException and Error, as any subclasses of these can be re-thrown. Might avoid a problem in the future if the specification changes to add another RuntimeException.

jcl/src/java.management/share/classes/com/ibm/java/lang/management/internal/LoggingMXBeanImpl.java
```diff
@@ -49,8 +57,60 @@
...
+			logger_getLevel = loggerClass.getMethod("getLevel");
+			logger_getParent = loggerClass.getMethod("getParent");
+			logger_getName = loggerClass.getMethod("getName");
+			logger_setLevel = loggerClass.getMethod("setLevel", levelClass);
+			level_getName = levelClass.getMethod("getName");
+			level_parse = levelClass.getMethod("parse", String.class);
+			logManager_LOGGING_MXBEAN_NAME = logManagerClass.getField("LOGGING_MXBEAN_NAME").get(null).toString();
 
+		} catch (Exception e) {
+			throw newInternalInvokeError(e);
```
pshipton: The throw isn't required, as newInternalInvokeError() doesn't ever return but throws itself. Applies to all places that use newInternalInvokeError. This also makes "newInternalInvokeError" a strange name for this helper.

jcl/src/java.management/share/classes/com/ibm/java/lang/management/internal/LoggingMXBeanImpl.java
```diff
@@ -49,8 +57,60 @@
...
 
-	private static final LoggingMXBeanImpl instance = new LoggingMXBeanImpl();
+	private static final LoggingMXBeanImpl instance;
+
+/*[IF Sidecar19-SE]
+	private static final Module java_logging;
+/*[ENDIF]*/
+	private static final Class<?> logManagerClass;
+	private static final Class<?> loggerClass;
+	private static final Class<?> levelClass;
```
keithc-ca: These three class references are only used in the static initializer: please make them local variables there.

jcl/src/java.management/share/classes/com/ibm/java/lang/management/internal/LoggingMXBeanImpl.java
DanHeidinga: By adding the throw, it's immediately obvious that execution cannot continue past this point.  I strongly believe in keeping the `throw` there.

jcl/src/java.management/share/classes/com/ibm/java/lang/management/internal/LoggingMXBeanImpl.java
```diff
@@ -49,8 +57,60 @@
...
+	private static final Method level_parse;
+	private static final String logManager_LOGGING_MXBEAN_NAME;
+	
+	static {
+		try {
+/*[IF Sidecar19-SE]
+/*[IF Sidecar19-SE-B165]
+			java_logging = ModuleLayer.boot().findModule("java.logging").get();
+/*[ELSE]
+			java_logging = Layer.boot().findModule("java.logging").get();
```
keithc-ca: This can fail if the (optional) java.logging module is not available, but we still need to provide an implementation of `PlatformLoggingMXBean`.

jcl/src/java.management/share/classes/com/ibm/java/lang/management/internal/LoggingMXBeanImpl.java
```diff
@@ -151,20 +230,61 @@ public String getParentLoggerName(String loggerName) {
...
+	 */
+	private static void handleInvocationTargetException(Exception cause) {
+		if (cause instanceof InvocationTargetException) {
+			Exception wrappedCause = (Exception)cause.getCause();
+			
+			if (wrappedCause instanceof SecurityException) {
+				throw (SecurityException)wrappedCause;
+			} else if (wrappedCause instanceof NullPointerException) {
+				throw (NullPointerException)wrappedCause;
+			} else if (wrappedCause instanceof IllegalArgumentException) {
```
keithc-ca: Are these the ***only*** exceptions that the target method might throw?

theresa-m: currently yes - I'm making this more general according to Peter's comment in case that changes

jcl/src/java.management/share/classes/com/ibm/java/lang/management/internal/LoggingMXBeanImpl.java
theresa-m: The reason I did it like this is because the constructor of this class is dependent on `logManager_LOGGING_MXBEAN_NAME` being set properly. I could set `objectName` conditionally on `java.logging` being found.

pshipton:
I jlink'ed a new runtime containing only java.management on OpenJDK, tried to run `ManagementFactory.getPlatformMXBean(PlatformLoggingMXBean.class)` and got the following exception.

```
Exception in thread "main" java.lang.IllegalArgumentException: java.lang.management.PlatformLoggingMXBean is not a platform management interface
        at java.management/java.lang.management.ManagementFactory$PlatformMBeanFinder.findSingleton(ManagementFactory.java:1000)
        at java.management/java.lang.management.ManagementFactory.getPlatformMXBean(ManagementFactory.java:678)
        at Test1a.main(Test1a.java:23)
```

The same test on OpenJ9 with this PR results in the following:

```
Exception in thread "main" java.lang.InternalError: java.util.NoSuchElementException: No value present
        at com.ibm.java.lang.management.internal.LoggingMXBeanImpl.newInternalInvokeError(java.management@9-internal/LoggingMXBeanImpl.java:243)
        at com.ibm.java.lang.management.internal.LoggingMXBeanImpl.<clinit>(java.management@9-internal/LoggingMXBeanImpl.java:82)
        at java.lang.management.DefaultPlatformMBeanProvider.<clinit>(java.management@9-internal/DefaultPlatformMBeanProvider.java:63)
        at java.lang.management.ManagementFactory$PlatformMBeanFinder.lambda$static$0(java.management@9-internal/ManagementFactory.java:920)
        at java.lang.management.ManagementFactory$PlatformMBeanFinder$$Lambda$5.00000000C4351760.run(Unknown Source)
        at java.security.AccessController.doPrivileged(java.base@9-internal/AccessController.java:833)
        at java.lang.management.ManagementFactory$PlatformMBeanFinder.<clinit>(java.management@9-internal/ManagementFactory.java:915)
        at java.lang.management.ManagementFactory.getPlatformMXBean(java.management@9-internal/ManagementFactory.java:678)
        at Test1a.main(Test1a.java:23)
Caused by: java.util.NoSuchElementException: No value present
        at java.util.Optional.get(java.base@9-internal/Optional.java:151)
        at com.ibm.java.lang.management.internal.LoggingMXBeanImpl.<clinit>(java.management@9-internal/LoggingMXBeanImpl.java:65)
        ... 7 more
```

jcl/src/java.management/share/classes/com/ibm/java/lang/management/internal/LoggingMXBeanImpl.java
pshipton: Then the helper could return the Throwable instead of throwing it.

DanHeidinga: The helper deliberately throws it to catch places were the caller has forgotten to write the throws.  It makes the code clear that the exception will be thrown and covers the case were the caller forgets the `throws`.

Its a pattern used heavily in the JSR 292 code.

jcl/src/java.management/share/classes/com/ibm/java/lang/management/internal/LoggingMXBeanImpl.java
keithc-ca: My preference is that they are only single line comments except in cases where a Java-aware editor would be confused (e.g. two imports or declarations of the same name).

pshipton: My guidelines are to comment out only when the code cannot compile. The master code isn't expected to work. If something needs to be commented, keep the latest code (i,e, Java 9) uncommented.

theresa-m:
I updated the error thrown in the static block to have a more meaningful message. The output is pretty confusing as to what is going on. Does this look okay?

```
Exception in thread "main" java.lang.InternalError: There was an issue accessing classes from java.logging. The module may not exist.
	at com.ibm.java.lang.management.internal.LoggingMXBeanImpl.handleError(java.management@9-internal/LoggingMXBeanImpl.java:241)
	at com.ibm.java.lang.management.internal.LoggingMXBeanImpl.<clinit>(java.management@9-internal/LoggingMXBeanImpl.java:77)
	at java.lang.management.DefaultPlatformMBeanProvider.<clinit>(java.management@9-internal/DefaultPlatformMBeanProvider.java:63)
	at java.lang.management.ManagementFactory$PlatformMBeanFinder.lambda$static$0(java.management@9-internal/ManagementFactory.java:920)
	at java.lang.management.ManagementFactory$PlatformMBeanFinder$$Lambda$5.000000008C31A040.run(Unknown Source)
	at java.security.AccessController.doPrivileged(java.base@9-internal/AccessController.java:833)
	at java.lang.management.ManagementFactory$PlatformMBeanFinder.<clinit>(java.management@9-internal/ManagementFactory.java:915)
	at java.lang.management.ManagementFactory.getPlatformMXBean(java.management@9-internal/ManagementFactory.java:678)
	at Test.main(Test.java:6)
Caused by: java.util.NoSuchElementException: No value present
	at java.util.Optional.get(java.base@9-internal/Optional.java:151)
	at com.ibm.java.lang.management.internal.LoggingMXBeanImpl.<clinit>(java.management@9-internal/LoggingMXBeanImpl.java:61)
	... 7 more
```

pshipton:
Better to update java.lang.management.ManagementFactory to duplicate the OpenJDK behavior, and throw IllegalArgumentException. We'll have to test all the ways of creating the PlatformLoggingMXBean, and ensure the OpenJ9 behavior is consistent with OpenJDK.

theresa-m:
It looks like we aren't using the OpenJ9 implementation of ManagementFactory in Java 9.

pshipton:
Right, forgot about that. com.ibm.java.lang.management.internal.ManagementUtils needs to be updated to avoid adding the interface when logging isn't loaded, and then ManagementFactory will behave the same.

pshipton:
ok, that code is also for Java 8. See java.lang.management.DefaultPlatformMBeanProvider

theresa-m:
OpenJ9 error with `ManagementFactory.getPlatformMXBean(PlatformLoggingMXBean.class)`:
```
Exception in thread "main" java.lang.IllegalArgumentException: java.lang.management.PlatformLoggingMXBean is not a platform management interface
	at java.lang.management.ManagementFactory$PlatformMBeanFinder.findSingleton(java.management@9-internal/ManagementFactory.java:1000)
	at java.lang.management.ManagementFactory.getPlatformMXBean(java.management@9-internal/ManagementFactory.java:678)
	at Test_1.main(Test_1.java:11)
```

OpenJ9 error with `ManagementFactory.newPlatformMXBeanProxy` for `java.util.logging:type=Logging`:
```
Exception in thread "main" java.lang.IllegalArgumentException: javax.management.InstanceNotFoundException: java.util.logging:type=Logging
	at java.lang.management.ManagementFactory.newPlatformMXBeanProxy(java.management@9-internal/ManagementFactory.java:617)
	at Test_3.main(Test_3.java:16)
Caused by: javax.management.InstanceNotFoundException: java.util.logging:type=Logging
	at com.sun.jmx.interceptor.DefaultMBeanServerInterceptor.getMBean(java.management@9-internal/DefaultMBeanServerInterceptor.java:1083)
	at com.sun.jmx.interceptor.DefaultMBeanServerInterceptor.isInstanceOf(java.management@9-internal/DefaultMBeanServerInterceptor.java:1379)
	at com.sun.jmx.mbeanserver.JmxMBeanServer.isInstanceOf(java.management@9-internal/JmxMBeanServer.java:1091)
	at java.lang.management.ManagementFactory.isInstanceOf(java.management@9-internal/ManagementFactory.java:645)
	at java.lang.management.ManagementFactory.newPlatformMXBeanProxy(java.management@9-internal/ManagementFactory.java:605)
	... 1 more
```

theresa-m:
Hotspot error with ManagementFactory.newPlatformMXBeanProxy for java.util.logging:type=Logging:
```
Exception in thread "main" java.lang.IllegalArgumentException: javax.management.InstanceNotFoundException: java.util.logging:type=Logging
	at java.management/java.lang.management.ManagementFactory.newPlatformMXBeanProxy(ManagementFactory.java:617)
	at Test_3.main(Test_3.java:16)
Caused by: javax.management.InstanceNotFoundException: java.util.logging:type=Logging
	at java.management/com.sun.jmx.interceptor.DefaultMBeanServerInterceptor.getMBean(DefaultMBeanServerInterceptor.java:1083)
	at java.management/com.sun.jmx.interceptor.DefaultMBeanServerInterceptor.isInstanceOf(DefaultMBeanServerInterceptor.java:1379)
	at java.management/com.sun.jmx.mbeanserver.JmxMBeanServer.isInstanceOf(JmxMBeanServer.java:1091)
	at java.management/java.lang.management.ManagementFactory.isInstanceOf(ManagementFactory.java:645)
	at java.management/java.lang.management.ManagementFactory.newPlatformMXBeanProxy(ManagementFactory.java:605)
	... 1 more
```

jcl/src/java.management/share/classes/com/ibm/java/lang/management/internal/LoggingMXBeanImpl.java
```diff
@@ -49,8 +57,54 @@
...
+	private static final Method logger_setLevel;
+	private static final Method level_getName;
+	private static final Method level_parse;
+	private static final String logManager_LOGGING_MXBEAN_NAME;
+	
+	static {
+		try {
+/*[IF Sidecar19-SE]*/
+/*[IF Sidecar19-SE-B165]*/
+			Module java_logging = ModuleLayer.boot().findModule("java.logging").get();
```
pshipton: Don't forget to tag the non-translatable strings with //$NON-NLS-1$

jcl/src/java.management/share/classes/java/lang/management/DefaultPlatformMBeanProvider.java
```diff
@@ -61,9 +67,21 @@
...
+/*[IF Sidecar19-SE-B165]*/
+			ModuleLayer.boot().findModule("java.logging").get();
+/*[ELSE]
+			Layer.boot().findModule("java.logging").get();
+/*[ENDIF]*/
+		
+			ComponentBuilder.create(LoggingMXBeanImpl.getInstance())
+				.addInterface(java.lang.management.PlatformLoggingMXBean.class)
+				.register(allComponents);
+		} catch (Exception e) {
```
pshipton: What is the exception that occurs when logging is not available? Can this catch something more specific than Exception?

pshipton: Seems this code could test if logging is available using isPresent​() rather than causing an exception by calling get().

theresa-m:
Added NON-NLS tags and used `isPresent` for java.logging load check.

pshipton:
jenkins test sanity plinux

pshipton:
jenkins test extended plinux

pshipton:
@keithc-ca any remaining concerns?

pshipton:
Sanity and extended test has passed on the current commit.

theresa-m:
^ didn't change anything here just restarting the travis build

jcl/src/java.management/share/classes/java/lang/management/DefaultPlatformMBeanProvider.java
```diff
@@ -61,9 +67,18 @@
...
-		ComponentBuilder.create(LoggingMXBeanImpl.getInstance())
-			.addInterface(java.lang.management.PlatformLoggingMXBean.class)
-			.register(allComponents);
+		/* LoggingMXBeanImpl depends on java.logging. If java.logging is not 
+		 * available exclude this component 
+		 */
+/*[IF Sidecar19-SE-B165]*/
+		if (ModuleLayer.boot().findModule("java.logging").isPresent()) { //$NON-NLS-1$
+/*[ELSE]
+		if (Layer.boot().findModule("java.logging").isPresent()) { //$NON-NLS-1$
```
keithc-ca: I'd like to see the opening `{` on lines 74 and 76 moved outside the preprocessor conditional.

keithc-ca:
It might be worthwhile updating jcl/j9modules.txt to reflect the fact that java.management no longer requires java.logging.

jcl/src/java.management/share/classes/com/ibm/java/lang/management/internal/LoggingMXBeanImpl.java
```diff
@@ -88,20 +142,26 @@ public static LoggingMXBeanImpl getInstance() {
...
-			// The named Logger exists. Now attempt to obtain its log level.
-			Level level = logger.getLevel();
-			if (level != null) {
-				result = level.getName();
-			} else {
-				// A null return from getLevel() means that the Logger
-				// is inheriting its log level from an ancestor. Return an
-				// empty string to the caller.
-				result = ""; //$NON-NLS-1$
+
```
keithc-ca: Perhaps we should keep the old implementation for Java 8?

jcl/src/java.management/share/classes/com/ibm/java/lang/management/internal/LoggingMXBeanImpl.java
```diff
@@ -112,7 +172,14 @@ public String getLoggerLevel(String loggerName) {
 	public List<String> getLoggerNames() {
 		// By default, return an empty list to caller
 		List<String> result = new ArrayList<>();
-		Enumeration<String> enumeration = LogManager.getLogManager().getLoggerNames();
+		Enumeration<String> enumeration;
```
keithc-ca: Can you change the type to `Enumeration<?>` to avoid 'Unchecked cast' warning?
(a cast will be needed to the value returned by `enumeration.nextElement()`).

jcl/src/java.management/share/classes/com/ibm/java/lang/management/internal/LoggingMXBeanImpl.java
```diff
@@ -151,20 +223,58 @@ public String getParentLoggerName(String loggerName) {
...
+	 * throws an error. This method unwraps that error for this class to maintain
+	 * its specification.
+	 */
+	private static void handleInvocationTargetException(Exception cause) {
+		if (cause instanceof InvocationTargetException) {
+			Throwable wrappedCause = cause.getCause();
+			if (wrappedCause instanceof Error) {
+				throw (Error)wrappedCause;
+			} else {
+				throw (RuntimeException)wrappedCause;
```
keithc-ca: I think this should be guarded by `wrappedCause instanceof  RuntimeException`.

jcl/src/java.management/share/classes/com/ibm/java/lang/management/internal/LoggingMXBeanImpl.java
pshipton: It would be a little less risky to keep Java 8 unmodified, as there is some overhead that will occur when optimizing the reflection.

theresa-m: Okay. I'll work on that.

jcl/src/java.management/share/classes/com/ibm/java/lang/management/internal/LoggingMXBeanImpl.java
```diff
@@ -142,25 +148,39 @@ public static LoggingMXBeanImpl getInstance() {
 	@Override
 	public String getLoggerLevel(String loggerName) {
```
theresa-m: I haven't finished / tested this out with the java 8 code back in but does this look too confusing to read? I'm not sure if its really worth the effort to not just have the whole method for java 8 and java 9 one after another.

theresa-m: This method I mean

pshipton: Works for me. This is typically how we have done it in the past to avoid duplicating code and to avoid needing multiple updates when making changes, which risks that something will be missed.

theresa-m: Just pushed the rest of the changes to include the old code for Java 8

jcl/src/java.management/share/classes/com/ibm/java/lang/management/internal/LoggingMXBeanImpl.java
```diff
@@ -151,20 +259,69 @@ public String getParentLoggerName(String loggerName) {
...
+		handleInvocationTargetException(cause);
+		
+		InternalError err = new InternalError(cause.toString());
+		err.initCause(cause);
+		throw err;
+	}
+	
+	/* invoke throws InvocationTargetException if the method it is invoking
+	 * throws an error. This method unwraps that error for this class to maintain
+	 * its specification.x
```
keithc-ca: Please remove the trailing `x`.

jcl/src/java.management/share/classes/java/lang/management/DefaultPlatformMBeanProvider.java
```diff
@@ -61,9 +67,19 @@
 			.addInterface(java.lang.management.OperatingSystemMXBean.class)
 			.register(allComponents);
 
-		ComponentBuilder.create(LoggingMXBeanImpl.getInstance())
-			.addInterface(java.lang.management.PlatformLoggingMXBean.class)
-			.register(allComponents);
+		/* LoggingMXBeanImpl depends on java.logging. If java.logging is not 
+		 * available exclude this component
```
keithc-ca: Please finish the sentence with a period (`.`).

theresa-m:
Fixed the commenting issues.

pshipton:
jenkins test all

pshipton:
All tests have passed and the failure is related to #1114

PR #1022, opened 2018-01-25
Fix compile failures on windows

tajila:
Fix compile failures on windows

### DIFF
--- a/runtime/vm/createramclass.cpp
+++ b/runtime/vm/createramclass.cpp
@@ -2804,10 +2804,10 @@ retry:
 					}
 
 					packageUTF = J9UTF8_DATA(J9ROMCLASS_CLASSNAME(packageRomClass));
-					pkgNameLength = packageNameLength(packageRomClass);
+					pkgNameLength = (U_32) packageNameLength(packageRomClass);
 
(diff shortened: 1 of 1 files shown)


DanHeidinga:
Jenkins test sanity plinux

runtime/vm/createramclass.cpp
```diff
@@ -2804,10 +2804,10 @@ internalCreateRAMClassFromROMClass(J9VMThread *vmThread, J9ClassLoader *classLoa
 					}
 
 					packageUTF = J9UTF8_DATA(J9ROMCLASS_CLASSNAME(packageRomClass));
-					pkgNameLength = packageNameLength(packageRomClass);
+					pkgNameLength = (U_32) packageNameLength(packageRomClass);
```
keithc-ca: Changing the declared type to `UDATA` would remove the need for these two changes.

DanHeidinga: I started down the same route but `findModuleForPackage` takes a U_32, and then truncates it to a U_16.  Fixing that is outside the scope of unblocking the windows builds.

keithc-ca: I was only suggesting the one-line change based on the other uses of `packageNameLength` I could find, but I agree this is close enough.

PR #1028, opened 2018-01-26
Ensure slot snapshot in Read Barrier

amicic:
Enusure that snapshot is really taken of the object slot, which is
generally volatile (can be mutated by another thread), since an
aggressive compiler can decide to re-read the slot instead of using the
snapshot variable.


charliegracie:
Jenkins test sanity plinux

Issue #968, opened 2018-01-17
OOM running pull request testing on plinux


pshipton:
We've seen this a couple of times today. 

https://ci.eclipse.org/openj9/job/PullRequest-Sanity-linux_ppc-64_cmprssptrs_le/328/console

```
15:40:02 package org.openj9.resources.annotationPackage
15:40:02 PASSED: test_elementFormDefault
15:40:02 PASSED: test_isAnnotationPresent
15:40:02 
15:40:02 ===============================================
15:40:02     JCL_TEST_Test-Annotation-Package
15:40:02     Tests run: 2, Failures: 0, Skips: 0
15:40:02 ===============================================
15:40:02 
15:40:02 org.openj9.test.java.lang.
15:40:03 JVMDUMP039I Processing dump event "systhrow", detail "java/lang/OutOfMemoryError" at 2018/01/17 20:40:03 - please wait.
15:40:03 JVMDUMP032I JVM requested System dump using '/home/jenkins/jenkins-agent/workspace/PullRequest-Sanity-linux_ppc-64_cmprssptrs_le/openj9/test/TestConfig/core.20180117.204003.20846.0001.dmp' in response to an event
```
It also occurred in testing for #962. 
https://ci.eclipse.org/openj9/job/PullRequest-Sanity-linux_ppc-64_cmprssptrs_le/326/console

Both of these occurred on ub16-ppcle-3

It didn't reoccur on restart, which occurred on ub16-ppcle-2
https://ci.eclipse.org/openj9/job/PullRequest-Sanity-linux_ppc-64_cmprssptrs_le/329/console

pshipton:
Happened in the last nightly build https://ci.eclipse.org/openj9/view/Test/job/Test-Sanity-JDK9-linux_ppc-64_cmprssptrs_le/49/console 
on ub16-ppcle-4

AdamBrousseau:
Also [315](https://ci.eclipse.org/openj9/job/PullRequest-Sanity-linux_ppc-64_cmprssptrs_le/315/consoleFull) on ub16-ppcle-3

dmitripivkine:
I have seen this before. This looks like an attempt of allocation of reference array with 2GB elements (intentional OOM). Test itself seems passed even with OOM thrown

pshipton:
Looking at the console output, in this case the OOM seems to have terminated the VM.

pshipton:
A working test should look like the following, and continue to run

```
org.openj9.test.java.lang.
[...]

PR #1029, opened 2018-01-26
Change some tracepoints from level 2 to 3, disabled by default

### LINKED ISSUE #1016: Should Trc_VM_sidecarInvokeReflectMethod_Entry/Exit be a level2 tracepoint
The Trc_VM_sidecarInvokeReflectMethod_Entry/Exit traces are level2 tracepoints which means they are on by default. There are cases where these tracepoints can account for up to 75% percent of the snaptrace output. 

These tracepoints do not provide any information other than the currentThread and the fact that the tracepoint was triggered. 

Should these tracepoints be level2?

pshipton:
As the tracepoints are not typically useful for problem diagnosis they
may as well be disabled by default to avoid the tracepoint overhead.

Fixes #1016


DanHeidinga:
Jenkins test sanity plinux

Kicking a PR build to ensure that none of the tests depend on these tracepoints firing.

DanHeidinga:
Re-launching the PR build as it was hit by #968 

Jenkins test sanity plinux

PR #1031, opened 2018-01-26
Copy jar inside the test

sophia-guo:
[ci skip]

Remove jar the duplicate utilsBk.jar in the build.xml


sophia-guo:
@llxia

Issue #1015, opened 2018-01-24
Java 9 missing packages in conf/security/java.security


pdbain-ibm:
Note that  com.ibm.oti. at al. became unrestricted in Java 9.

The package.access for Java 9 is:
package.access=sun.misc.,\
               sun.reflect.

The corresponding entry in Java 8  jre/lib/security/java.security  is:
package.access=sun.,\
               com.sun.xml.internal.,\
               com.sun.imageio.,\
               com.sun.istack.internal.,\
               com.sun.jmx.,\
               com.sun.media.sound.,\
               com.sun.naming.internal.,\
               com.sun.proxy.,\
               com.sun.corba.se.,\
               com.sun.org.apache.bcel.internal.,\
               com.sun.org.apache.regexp.internal.,\
               com.sun.org.apache.xerces.internal.,\
               com.sun.org.apache.xpath.internal.,\
               com.sun.org.apache.xalan.internal.extensions.,\
               com.sun.org.apache.xalan.internal.lib.,\
               com.sun.org.apache.xalan.internal.res.,\
               com.sun.org.apache.xalan.internal.templates.,\
               com.sun.org.apache.xalan.internal.utils.,\
               com.sun.org.apache.xalan.internal.xslt.,\
               com.sun.org.apache.xalan.internal.xsltc.cmdline.,\
               com.sun.org.apache.xalan.internal.xsltc.compiler.,\
               com.sun.org.apache.xalan.internal.xsltc.trax.,\
               com.sun.org.apache.xalan.internal.xsltc.util.,\
               com.sun.org.apache.xml.internal.res.,\
               com.sun.org.apache.xml.internal.resolver.helpers.,\
               com.sun.org.apache.xml.internal.resolver.readers.,\
               com.sun.org.apache.xml.internal.security.,\
               com.sun.org.apache.xml.internal.serializer.utils.,\
               com.sun.org.apache.xml.internal.utils.,\
               com.sun.org.glassfish.,\
               com.oracle.xmlns.internal.,\
               com.oracle.webservices.internal.,\
               oracle.jrockit.jfr.,\
               org.jcp.xml.dsig.internal.,\
               jdk.internal.,\
               jdk.nashorn.internal.,\
               jdk.nashorn.tools.,\
               **com.ibm.oti.,\**
               com.sun.activation.registries.

This blocks https://github.com/eclipse/openj9/issues/849.

pshipton:
[...]

PR #1033, opened 2018-01-26
Add permission checking to shared cache utilities

pshipton:
In Java 9 and later, com.ibm.oti. is not a restricted package. The
shared cache utilities need to check permissions to avoid untrusted
callers when the security manager is enabled.

Fixes: #1015

### DIFF
--- a/jcl/src/openj9.sharedclasses/share/classes/com/ibm/oti/shared/SharedClassUtilities.java
+++ b/jcl/src/openj9.sharedclasses/share/classes/com/ibm/oti/shared/SharedClassUtilities.java
@@ -2,7 +2,7 @@
 package com.ibm.oti.shared;
 
 /*******************************************************************************
- * Copyright (c) 2010, 2017 IBM Corp. and others
+ * Copyright (c) 2010, 2018 IBM Corp. and others
  *
  * This program and the accompanying materials are made available under
  * the terms of the Eclipse Public License 2.0 which accompanies this
@@ -24,6 +24,7 @@ package com.ibm.oti.shared;
  *******************************************************************************/
 
 import java.util.*;
+import com.ibm.oti.shared.SharedClassesNamedPermission.SharedPermissions;
 
 /**
  * SharedClassUtilities provides APIs to get information about all shared class caches in a directory and
@@ -107,8 +108,16 @@ public class SharedClassUtilities {
 	 * 					If shared classes is disabled for this JVM (that is -Xshareclasses:none is present).
 	 * @throws		IllegalArgumentException
 	 * 					If <code>flags</code> is not a valid value.
+	 * @throws		SecurityException
+	 * 					If a security manager is enabled and the calling thread does not
+	 * 					have SharedClassesNamedPermission("getSharedCacheInfo")
 	 */
 	static public List<SharedClassCacheInfo> getSharedCacheInfo(String cacheDir, int flags, boolean useCommandLineValues) {
+		SecurityManager sm = System.getSecurityManager();
+		if (sm != null) {
+			sm.checkPermission(SharedPermissions.getSharedCacheInfo);
+		}
+		
 		int retVal;
 		/*[MSG "K0553", "parameter {0} has invalid value"]*/
 		if (flags != NO_FLAGS) {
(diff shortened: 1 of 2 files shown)


pshipton:
Example SecurityException calling a shared classes utility from Java 9. Note by default this SecurityException doesn't occur on Java 8 because you get `Access denied ("java.lang.RuntimePermission" "accessClassInPackage.com.ibm.oti.shared")` instead.

```
		System.setSecurityManager(new SecurityManager());
		SharedClassUtilities.getSharedCacheInfo(".", 0, true);
```

```
Exception in thread "main" java.security.AccessControlException: Access denied ("com.ibm.oti.shared.SharedClassesPermission" "getSharedCacheInfo")
        at java.security.AccessController.throwACE(java.base@9-internal/AccessController.java:176)
        at java.security.AccessController.checkPermissionHelper(java.base@9-internal/AccessController.java:236)
        at java.security.AccessController.checkPermission(java.base@9-internal/AccessController.java:371)
        at java.lang.SecurityManager.checkPermission(java.base@9-internal/SecurityManager.java:558)
        at com.ibm.oti.shared.SharedClassUtilities.getSharedCacheInfo(openj9.sharedclasses@9-internal/SharedClassUtilities.java:117)
        at Test1a.main(Test1a.java:20)
```

pshipton:
@hangshao0 fyi

hangshao0:
Probably it is better to rename the new class to SharedClassUtilPermission to avoid confusion. We already have a class called SharedClassPermission.

pshipton:
> Probably it is better to rename the new class to SharedClassUtilPermission to avoid confusion.

Ha. That was my original thought as well, although I actually used SharedClassesUtilPermission, but then I changed it because I figured the name should be more generic. In the future the named permissions might not all be related to the utilities. Although the names are similar, the constructors take different parameters. SharedClassPermission takes a ClassLoader, while SharedClassesPermission takes a String. We can extend SharedClassesPermission in the future to add addition names as needed. I'm still open to changing the name, but I don't think the name should contain "Util". Perhaps SharedCacheNamedPermission?

pshipton:
Oops, I meant SharedClassesNamedPermission. "SharedClasses" to be consistent because that is the name of the command line option and module. Not "SharedClass" because the permissions are not necessarily related to loading classes.

hangshao0:
SharedClassesNamedPermission is better. I got confused about "SharedClassesPermission" until I noticed the extra "es" after 10 seconds...

pshipton:
I've updated the class to SharedClassesNamedPermission.

DanHeidinga:
@pshipton Is there a documentation update that goes with this so users know to add the permission?

pshipton:
@DanHeidinga Thanks for the reminder. Since there isn't any doc for OpenJ9 yet, I've created a tracking item internal to IBM to update the docs. Otherwise, the javadoc for the APIs is updated as part of this PR.

pshipton:
jenkins test sanity zlinux
jenkins test extended zlinux

pshipton:
The extended test failure is caused by a know issue #1086 and unrelated to this change.

jcl/src/openj9.sharedclasses/share/classes/com/ibm/oti/shared/SharedClassesNamedPermission.java
```diff
@@ -0,0 +1,90 @@
...
+public final class SharedClassesNamedPermission extends BasicPermission {
+	
+	private static final long serialVersionUID = -4800623387760880313L;
+
+	/*
+	 * Helper class to avoid creating permission objects unless needed at runtime.
+	 */
+	static final class SharedPermissions {
+		public static SharedClassesNamedPermission getSharedCacheInfo = new SharedClassesNamedPermission("getSharedCacheInfo"); //$NON-NLS-1$
+		public static SharedClassesNamedPermission destroySharedCache = new SharedClassesNamedPermission("destroySharedCache"); //$NON-NLS-1$
```
keithc-ca: Shouldn't these be `final`?

jcl/src/openj9.sharedclasses/share/classes/com/ibm/oti/shared/SharedClassUtilities.java
```diff
@@ -107,8 +108,16 @@
 	 * 					If shared classes is disabled for this JVM (that is -Xshareclasses:none is present).
 	 * @throws		IllegalArgumentException
 	 * 					If <code>flags</code> is not a valid value.
+	 * @throws		SecurityException
+	 * 					If a security manager is enabled and the calling thread does not
+	 * 					have SharedClassesUtilPermission("getSharedCacheInfo")
```
keithc-ca: The permission class is `SharedClassesNamedPermission`, right? Please update the javadocs (here and in `destroySharedCache`).

jcl/src/openj9.sharedclasses/share/classes/com/ibm/oti/shared/SharedClassesNamedPermission.java
pshipton: Yup, thanks.

pshipton:
@keithc-ca commit is updated as per your earlier comments.

keithc-ca:
I don't think the updates warrant another set of jenkins builds.

PR #1044, opened 2018-01-29
Use OMR implementation of omrfile_chown()

hangshao0:
1. Use OMR version of file_chown().
2. Add a test case for setting groupID of shared cache control file.

### DIFF
--- a/test/cmdLineTests/shareClassTests/SCCMLTests/ShareClassesCMLTests-6.xml
+++ b/test/cmdLineTests/shareClassTests/SCCMLTests/ShareClassesCMLTests-6.xml
@@ -24,7 +24,7 @@
 
 <!DOCTYPE suite SYSTEM "cmdlinetester.dtd">
 
-<!-- Test 200-a to Test 206-aa: 81 tests -->
+<!-- Test 200-a to Test 207 cleanup: 83 tests -->
 
 <suite id="Shared Classes CommandLineOptionTests Suite">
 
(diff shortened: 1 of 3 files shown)


test/cmdLineTests/shareClassTests/SCCMLTests/ShareClassesCMLTests-6.xml
```diff
@@ -911,6 +911,29 @@
 		<output type="failure" caseSensitive="no" regex="no">corrupt</output>
 	</test>
 	
+	<test id="Test 207: Create a non-persistent cache" timeout="600" runPath=".">
+		<command>$JAVA_EXE$ -Xtrace:print=j9prt.700 $currentMode$,nonpersistent -version</command>
+		<output type="success" caseSensitive="yes" regex="yes" javaUtilPattern="yes">(java|openjdk) version</output>
```
DanHeidinga: Shouldn't this check for the tracepoint output as well as a success condition?

hangshao0: The tracepoint output (ControlFileFDWithWriteLock (Info: could not chown file.)) is a failure condition. chown() failure is something that should not happen.

DanHeidinga:
Jenkins test sanity plinux

test/cmdLineTests/shareClassTests/SCCMLTests/ShareClassesCMLTests-6.xml
pshipton: Rather than adding yet another test, could we just modify an existing test? i.e. Test 26 in test/cmdLineTests/shareClassTests/SCCMLTests/ShareClassesCMLTests-1.xml creates a nonpersistent cache.

pshipton: There is another one here test/cmdLineTests/shareClassTests/SCCMLTests/testSCCMLSnapshot.xml
Test 2s

hangshao0: testSCCMLSnapshot.xml does not run on windows.

pshipton: right, I forgot about that. lgtm then.

hangshao0:
Minor change:
regex should be "yes" when checking for tracepoint j9prt.700

DanHeidinga:
Jenkins test sanity plinux

PR #1045, opened 2018-01-29
Use host class loader for anon class constraint checks

gacholio:
When performing class loading constraint checks, use the host class
loader rather than the internal anon loader (of which there is only
one).


DanHeidinga:
Jenkins test sanity plinux

Issue #966, opened 2018-01-17
Accelerate uses of copyCharsIntoUTF8Helper


fjeremic:
While profiling a webserver type workload `copyCharsIntoUTF8Helper` is the top VM function in the profile at steady state throughput. It is a clear winner in terms of bang for the buck when it comes to making VM APIs faster.

There seems to be one direct use of `copyStringToUTF8Helper` without going through the function interface in `getStringUTFRegion`. It looks as follows:

https://github.com/eclipse/openj9/blob/29b1bc6770ef92e162b923945627c206b4dfba6c/runtime/vm/jnimisc.cpp#L806-L826

From the profile disassembly we do not seem to inline the call to `copyCharsIntoUTF8Helper`. Note that the `stringFlags` argument is `J9_STR_NONE` and `utf8Length` is `UDATA_MAX`. This means that the flag checking and remaining bytes checking we do for each character is unneeded. If we were to inline this function all of this would have been folded away by the compiler. This is a simple optimization.

In fact after doing this I don't see too much reason for this API to exist as we already have `copyStringToUTF8Helper` which more or less does the exact same thing. We should perhaps unify the two APIs.

PR #1047, opened 2018-01-29
Accelerate String copies to UTF8

fjeremic:
- Return UTF8 length from copyStringToUTF8Helper
- Provide an I_8 version of encodeUTF8Char
- Remove utf8Len parameter from copyStringToUTF8Helper
- Add prependStrLength parameter to copyStringToUTF8WithMemAlloc
- Use stack allocated buffer for copyStringToUTF8WithMemAlloc if possible
- Replace uses of copyStringToUTF8Helper with copyStringToUTF8WithMemAlloc
- Introduce outgoing utf8Length parameter to copyStringToUTF8WithMemAlloc
- Deprecate null termination parameter in copyStringToUTF8Helper
- Deprecate copyFromStringIntoUTF8
- Force null termination on all calls to copyStringToUTF8Helper
- Deprecate copyStringToUTF8
- Deprecate copyCharsIntoUTF8Helper

Closes: #966

### DIFF
--- a/runtime/j9vm/j7vmi.c
+++ b/runtime/j9vm/j7vmi.c
@@ -2527,6 +2527,7 @@ jvmDefineClassHelper(JNIEnv *env, jobject classLoaderObject,
 	J9ClassLoader *classLoader = NULL;
 	UDATA retried = FALSE;
 	UDATA utf8Length = 0;
+	char utf8NameStackBuffer[J9VM_PACKAGE_NAME_BUFFER_LENGTH];
 	U_8 *utf8Name = NULL;
 	J9Class *clazz = NULL;
 	jclass result = NULL;
@@ -2553,7 +2554,7 @@ jvmDefineClassHelper(JNIEnv *env, jobject classLoaderObject,
 	/* Allocate and initialize a UTF8 copy of the Unicode class-name */
 	if (className != NULL) {
 		/* For now, use malloc, but this should use stack allocation for short class-names */
-		utf8Name = vmFuncs->copyStringToUTF8WithMemAlloc(currentThread, J9_JNI_UNWRAP_REFERENCE(className), J9_STR_XLAT, "", utf8Name, 0, &utf8Length);
+		utf8Name = vmFuncs->copyStringToUTF8WithMemAlloc(currentThread, J9_JNI_UNWRAP_REFERENCE(className), J9_STR_XLAT, "", utf8NameStackBuffer, J9VM_PACKAGE_NAME_BUFFER_LENGTH, &utf8Length);
 		if (NULL == utf8Name) {
 			vmFuncs->setNativeOutOfMemoryError(currentThread, 0, 0);
 			goto done;
@@ -2643,7 +2644,10 @@ done:
 
 	vmFuncs->internalReleaseVMAccess(currentThread);
 
-	j9mem_free_memory(utf8Name);
+	if (utf8NameStackBuffer != utf8Name) {
+		j9mem_free_memory(utf8Name);
+	}
+
 	return result;
 }
 
--- a/runtime/j9vm/j7vmi.c
+++ b/runtime/j9vm/j7vmi.c
@@ -2551,9 +2551,7 @@ jvmDefineClassHelper(JNIEnv *env, jobject classLoaderObject,
 
 	vmFuncs->internalEnterVMFromJNI(currentThread);
 
-	/* Allocate and initialize a UTF8 copy of the Unicode class-name */
 	if (NULL != className) {
-		/* For now, use malloc, but this should use stack allocation for short class-names */
 		utf8Name = vmFuncs->copyStringToUTF8WithMemAlloc(currentThread, J9_JNI_UNWRAP_REFERENCE(className), J9_STR_NULL_TERMINATE_RESULT | J9_STR_XLAT, "", 0, utf8NameStackBuffer, J9VM_PACKAGE_NAME_BUFFER_LENGTH, &utf8Length);
 		if (NULL == utf8Name) {
 			vmFuncs->setNativeOutOfMemoryError(currentThread, 0, 0);
--- a/runtime/j9vm/j7vmi.c
+++ b/runtime/j9vm/j7vmi.c
@@ -2554,7 +2554,7 @@ jvmDefineClassHelper(JNIEnv *env, jobject classLoaderObject,
 	/* Allocate and initialize a UTF8 copy of the Unicode class-name */
 	if (className != NULL) {
 		/* For now, use malloc, but this should use stack allocation for short class-names */
-		utf8Name = vmFuncs->copyStringToUTF8WithMemAlloc(currentThread, J9_JNI_UNWRAP_REFERENCE(className), J9_STR_XLAT, "", utf8NameStackBuffer, J9VM_PACKAGE_NAME_BUFFER_LENGTH, &utf8Length);
+		utf8Name = vmFuncs->copyStringToUTF8WithMemAlloc(currentThread, J9_JNI_UNWRAP_REFERENCE(className), J9_STR_XLAT, "", 0, utf8NameStackBuffer, J9VM_PACKAGE_NAME_BUFFER_LENGTH, &utf8Length);
 		if (NULL == utf8Name) {
 			vmFuncs->setNativeOutOfMemoryError(currentThread, 0, 0);
 			goto done;
--- a/runtime/j9vm/j7vmi.c
+++ b/runtime/j9vm/j7vmi.c
@@ -2551,19 +2551,11 @@ jvmDefineClassHelper(JNIEnv *env, jobject classLoaderObject,
 	vmFuncs->internalEnterVMFromJNI(currentThread);
 
 	/* Allocate and initialize a UTF8 copy of the Unicode class-name */
-	if (className == NULL) {
-		utf8Length = 0;
-		utf8Name = NULL;
-	} else {
-		utf8Length = vmFuncs->getStringUTF8Length(currentThread, J9_JNI_UNWRAP_REFERENCE(className));
+	if (className != NULL) {
 		/* For now, use malloc, but this should use stack allocation for short class-names */
-		utf8Name = j9mem_allocate_memory(utf8Length+1, J9MEM_CATEGORY_CLASSES);
-		if (utf8Name == NULL) {
+		utf8Name = vmFuncs->copyStringToUTF8WithMemAlloc(currentThread, J9_JNI_UNWRAP_REFERENCE(className), J9_STR_XLAT, "", utf8Name, 0, &utf8Length);
+		if (NULL == utf8Name) {
 			vmFuncs->setNativeOutOfMemoryError(currentThread, 0, 0);
-			goto done;
-		}
-		if (UDATA_MAX == vmFuncs->copyStringToUTF8Helper(currentThread, J9_JNI_UNWRAP_REFERENCE(className), J9_STR_XLAT, utf8Name, utf8Length + 1)) {
-			vmFuncs->setCurrentException(currentThread, J9VMCONSTANTPOOL_JAVALANGINTERNALERROR, NULL);
 			goto done;
 		}
 	}
--- a/runtime/j9vm/j7vmi.c
+++ b/runtime/j9vm/j7vmi.c
@@ -2554,7 +2554,7 @@ jvmDefineClassHelper(JNIEnv *env, jobject classLoaderObject,
 	/* Allocate and initialize a UTF8 copy of the Unicode class-name */
 	if (NULL != className) {
 		/* For now, use malloc, but this should use stack allocation for short class-names */
-		utf8Name = vmFuncs->copyStringToUTF8WithMemAlloc(currentThread, J9_JNI_UNWRAP_REFERENCE(className), J9_STR_XLAT, "", 0, utf8NameStackBuffer, J9VM_PACKAGE_NAME_BUFFER_LENGTH, &utf8Length);
+		utf8Name = vmFuncs->copyStringToUTF8WithMemAlloc(currentThread, J9_JNI_UNWRAP_REFERENCE(className), J9_STR_NULL_TERMINATE_RESULT | J9_STR_XLAT, "", 0, utf8NameStackBuffer, J9VM_PACKAGE_NAME_BUFFER_LENGTH, &utf8Length);
 		if (NULL == utf8Name) {
 			vmFuncs->setNativeOutOfMemoryError(currentThread, 0, 0);
 			goto done;
--- a/runtime/j9vm/j7vmi.c
+++ b/runtime/j9vm/j7vmi.c
@@ -2552,7 +2552,7 @@ jvmDefineClassHelper(JNIEnv *env, jobject classLoaderObject,
 	vmFuncs->internalEnterVMFromJNI(currentThread);
 
 	/* Allocate and initialize a UTF8 copy of the Unicode class-name */
-	if (className != NULL) {
+	if (NULL != className) {
 		/* For now, use malloc, but this should use stack allocation for short class-names */
 		utf8Name = vmFuncs->copyStringToUTF8WithMemAlloc(currentThread, J9_JNI_UNWRAP_REFERENCE(className), J9_STR_XLAT, "", 0, utf8NameStackBuffer, J9VM_PACKAGE_NAME_BUFFER_LENGTH, &utf8Length);
 		if (NULL == utf8Name) {
--- a/runtime/j9vm/j7vmi.c
+++ b/runtime/j9vm/j7vmi.c
@@ -2562,9 +2562,7 @@ jvmDefineClassHelper(JNIEnv *env, jobject classLoaderObject,
 			vmFuncs->setNativeOutOfMemoryError(currentThread, 0, 0);
 			goto done;
 		}
-		if (UDATA_MAX == vmFuncs->copyStringToUTF8Helper(
-			currentThread, J9_JNI_UNWRAP_REFERENCE(className), TRUE, J9_STR_XLAT, utf8Name, utf8Length + 1)
-		) {
+		if (UDATA_MAX == vmFuncs->copyStringToUTF8Helper(currentThread, J9_JNI_UNWRAP_REFERENCE(className), J9_STR_XLAT, utf8Name, utf8Length + 1)) {
 			vmFuncs->setCurrentException(currentThread, J9VMCONSTANTPOOL_JAVALANGINTERNALERROR, NULL);
 			goto done;
 		}
--- a/runtime/j9vm/j7vmi.c
+++ b/runtime/j9vm/j7vmi.c
@@ -2644,7 +2644,7 @@ done:
 
 	vmFuncs->internalReleaseVMAccess(currentThread);
 
-	if (utf8NameStackBuffer != utf8Name) {
+	if ((U_8*)utf8NameStackBuffer != utf8Name) {
 		j9mem_free_memory(utf8Name);
 	}
 
--- a/runtime/j9vm/java9vmi.c
+++ b/runtime/j9vm/java9vmi.c
@@ -409,7 +409,10 @@ trcModulesCreationPackage(J9VMThread * currentThread, J9Module * fromModule, j9o
 		if (moduleNameBuf != moduleNameUTF) {
 			j9mem_free_memory(moduleNameUTF);
 		}
+	} else {
+		vmFuncs->setNativeOutOfMemoryError(currentThread, 0, 0);
 	}
+
 #if J9VM_JAVA9_BUILD < 156
 	if (packageNameBuf != packageNameUTF) {
 		j9mem_free_memory(packageNameUTF);
--- a/runtime/j9vm/java9vmi.c
+++ b/runtime/j9vm/java9vmi.c
@@ -381,11 +381,11 @@ trcModulesCreationPackage(J9VMThread * currentThread, J9Module * fromModule, j9o
 	J9InternalVMFunctions const * const vmFuncs = currentThread->javaVM->internalVMFunctions;
 	char moduleNameBuf[J9VM_PACKAGE_NAME_BUFFER_LENGTH];
 	char *moduleNameUTF = vmFuncs->copyStringToUTF8WithMemAlloc(
-		currentThread, fromModule->moduleName, J9_STR_NONE, "", moduleNameBuf, J9VM_PACKAGE_NAME_BUFFER_LENGTH);
+		currentThread, fromModule->moduleName, J9_STR_NONE, "", moduleNameBuf, J9VM_PACKAGE_NAME_BUFFER_LENGTH, NULL);
 #if J9VM_JAVA9_BUILD < 156
 	char packageNameBuf[J9VM_PACKAGE_NAME_BUFFER_LENGTH];
 	char *packageNameUTF = vmFuncs->copyStringToUTF8WithMemAlloc(
-		currentThread, package, J9_STR_NONE, "", packageNameBuf, J9VM_PACKAGE_NAME_BUFFER_LENGTH);
+		currentThread, package, J9_STR_NONE, "", packageNameBuf, J9VM_PACKAGE_NAME_BUFFER_LENGTH, NULL);
 #endif /* J9VM_JAVA9_BUILD < 156 */
 
 	if (NULL != moduleNameUTF) {
@@ -627,14 +627,14 @@ trcModulesAddModuleExportsToAll(J9VMThread * currentThread, J9Module * fromModul
 	J9InternalVMFunctions const * const vmFuncs = currentThread->javaVM->internalVMFunctions;
 	char fromModuleNameBuf[J9VM_PACKAGE_NAME_BUFFER_LENGTH];
 	char *fromModuleNameUTF = vmFuncs->copyStringToUTF8WithMemAlloc(
-		currentThread, fromModule->moduleName, J9_STR_NONE, "", fromModuleNameBuf, J9VM_PACKAGE_NAME_BUFFER_LENGTH);
+		currentThread, fromModule->moduleName, J9_STR_NONE, "", fromModuleNameBuf, J9VM_PACKAGE_NAME_BUFFER_LENGTH, NULL);
 	if (NULL != fromModuleNameUTF) {
 #if J9VM_JAVA9_BUILD >= 156
 		Trc_MODULE_add_module_exports_to_all(currentThread, package, fromModuleNameUTF);
 #else
 		char packageNameBuf[J9VM_PACKAGE_NAME_BUFFER_LENGTH];
 		char *packageNameUTF = vmFuncs->copyStringToUTF8WithMemAlloc(
-			currentThread, J9_JNI_UNWRAP_REFERENCE(package), J9_STR_NONE, "", packageNameBuf, J9VM_PACKAGE_NAME_BUFFER_LENGTH);
+			currentThread, J9_JNI_UNWRAP_REFERENCE(package), J9_STR_NONE, "", packageNameBuf, J9VM_PACKAGE_NAME_BUFFER_LENGTH, NULL);
 		if (NULL != packageNameUTF) {
 			Trc_MODULE_add_module_exports_to_all(currentThread, packageNameUTF, fromModuleNameUTF);
 			if (packageNameBuf != packageNameUTF) {
@@ -682,14 +682,14 @@ trcModulesAddModuleExportsToAllUnnamed(J9VMThread * currentThread, J9Module * fr
 	J9InternalVMFunctions const * const vmFuncs = currentThread->javaVM->internalVMFunctions;
 	char fromModuleNameBuf[J9VM_PACKAGE_NAME_BUFFER_LENGTH];
 	char *fromModuleNameUTF = vmFuncs->copyStringToUTF8WithMemAlloc(
-		currentThread, fromModule->moduleName, J9_STR_NONE, "", fromModuleNameBuf, J9VM_PACKAGE_NAME_BUFFER_LENGTH);
+		currentThread, fromModule->moduleName, J9_STR_NONE, "", fromModuleNameBuf, J9VM_PACKAGE_NAME_BUFFER_LENGTH, NULL);
 	if (NULL != fromModuleNameUTF) {
 #if J9VM_JAVA9_BUILD >= 156
 		Trc_MODULE_add_module_exports_to_all_unnamed(currentThread, package, fromModuleNameUTF);
 #else
 		char packageNameBuf[J9VM_PACKAGE_NAME_BUFFER_LENGTH];
 		char *packageNameUTF = vmFuncs->copyStringToUTF8WithMemAlloc(
-			currentThread, J9_JNI_UNWRAP_REFERENCE(package), J9_STR_NONE, "", packageNameBuf, J9VM_PACKAGE_NAME_BUFFER_LENGTH);
+			currentThread, J9_JNI_UNWRAP_REFERENCE(package), J9_STR_NONE, "", packageNameBuf, J9VM_PACKAGE_NAME_BUFFER_LENGTH, NULL);
 		if (NULL != packageNameUTF) {
 			Trc_MODULE_add_module_exports_to_all_unnamed(currentThread, packageNameUTF, fromModuleNameUTF);
 			if (packageNameBuf != packageNameUTF) {
@@ -778,16 +778,16 @@ trcModulesAddModuleExports(J9VMThread *currentThread, J9Module *fromModule, jstr
 	char fromModuleNameBuf[J9VM_PACKAGE_NAME_BUFFER_LENGTH];
 	char toModuleNameBuf[J9VM_PACKAGE_NAME_BUFFER_LENGTH];
 	char *fromModuleNameUTF = vmFuncs->copyStringToUTF8WithMemAlloc(
-		currentThread, fromModule->moduleName, J9_STR_NONE, "", fromModuleNameBuf, J9VM_PACKAGE_NAME_BUFFER_LENGTH);
+		currentThread, fromModule->moduleName, J9_STR_NONE, "", fromModuleNameBuf, J9VM_PACKAGE_NAME_BUFFER_LENGTH, NULL);
 	char *toModuleNameUTF = vmFuncs->copyStringToUTF8WithMemAlloc(
-		currentThread, toModule->moduleName, J9_STR_NONE, "", toModuleNameBuf, J9VM_PACKAGE_NAME_BUFFER_LENGTH);
+		currentThread, toModule->moduleName, J9_STR_NONE, "", toModuleNameBuf, J9VM_PACKAGE_NAME_BUFFER_LENGTH, NULL);
 	if ((NULL != fromModuleNameUTF) && (NULL != toModuleNameUTF)) {
 #if J9VM_JAVA9_BUILD >= 156
 		Trc_MODULE_add_module_exports(currentThread, package, fromModuleNameUTF, toModuleNameUTF);
 #else
 		char packageNameBuf[J9VM_PACKAGE_NAME_BUFFER_LENGTH];
 		char *packageNameUTF = vmFuncs->copyStringToUTF8WithMemAlloc(
-			currentThread, J9_JNI_UNWRAP_REFERENCE(package), J9_STR_NONE, "", packageNameBuf, J9VM_PACKAGE_NAME_BUFFER_LENGTH);
+			currentThread, J9_JNI_UNWRAP_REFERENCE(package), J9_STR_NONE, "", packageNameBuf, J9VM_PACKAGE_NAME_BUFFER_LENGTH, NULL);
 		if (NULL != packageNameUTF) {
 			Trc_MODULE_add_module_exports(currentThread, packageNameUTF, fromModuleNameUTF, toModuleNameUTF);
 			if (packageNameBuf != packageNameUTF) {
@@ -977,7 +977,7 @@ JVM_DefineModule(JNIEnv * env, jobject module, jstring version, jstring location
 					char *nameUTF = buf;
 
 					nameUTF = vmFuncs->copyStringToUTF8WithMemAlloc(
-						currentThread, moduleName, J9_STR_NONE, "", buf, J9VM_PACKAGE_NAME_BUFFER_LENGTH);
+						currentThread, moduleName, J9_STR_NONE, "", buf, J9VM_PACKAGE_NAME_BUFFER_LENGTH, NULL);
 					if (NULL == nameUTF) {
 						success = FALSE;
 						goto nativeOOM;
@@ -1154,12 +1154,12 @@ trcModulesAddReadsModule(J9VMThread *currentThread, jobject toModule, J9Module *
 	char fromModuleNameBuf[J9VM_PACKAGE_NAME_BUFFER_LENGTH];
 	char toModuleNameBuf[J9VM_PACKAGE_NAME_BUFFER_LENGTH];
 	char *fromModuleNameUTF = vmFuncs->copyStringToUTF8WithMemAlloc(
-		currentThread, j9FromMod->moduleName, J9_STR_NONE, "", fromModuleNameBuf, J9VM_PACKAGE_NAME_BUFFER_LENGTH);
+		currentThread, j9FromMod->moduleName, J9_STR_NONE, "", fromModuleNameBuf, J9VM_PACKAGE_NAME_BUFFER_LENGTH, NULL);
 	char *toModuleNameUTF = NULL;
 
 	if (NULL != toModule) {
 		toModuleNameUTF = vmFuncs->copyStringToUTF8WithMemAlloc(
-			currentThread, j9ToMod->moduleName, J9_STR_NONE, "", toModuleNameBuf, J9VM_PACKAGE_NAME_BUFFER_LENGTH);
+			currentThread, j9ToMod->moduleName, J9_STR_NONE, "", toModuleNameBuf, J9VM_PACKAGE_NAME_BUFFER_LENGTH, NULL);
 	} else {
 #define LOOSE_MODULE   "loose "
 		PORT_ACCESS_FROM_VMC(currentThread);
@@ -1321,7 +1321,7 @@ JVM_IsExportedToModule(JNIEnv * env, jobject fromModule, jstring package, jobjec
 		char buf[J9VM_PACKAGE_NAME_BUFFER_LENGTH];
 		char *nameUTF = buf;
 		nameUTF = vmFuncs->copyStringToUTF8WithMemAlloc(
-			currentThread, pkgObject, J9_STR_NONE, "", buf, J9VM_PACKAGE_NAME_BUFFER_LENGTH);
+			currentThread, pkgObject, J9_STR_NONE, "", buf, J9VM_PACKAGE_NAME_BUFFER_LENGTH, NULL);
 		if (NULL == nameUTF) {
 			vmFuncs->setNativeOutOfMemoryError(currentThread, 0, 0);
 			goto done;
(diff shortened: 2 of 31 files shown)


fjeremic:
## Performance evaluation of this change

I measured steady state performance running Liberty DayTrader7 and noticed we have quite a lot of ticks in the libj9vm.so module related to UTF8 string copies. The profile looked as follows:

```
Ticks	Pct_Total	%    	Symbol/Functions                             	
16813	0.53     	13.83	Fast_java_lang_J9VMInternals_identityHashCode	
13222	0.42     	10.88	Fast_java_lang_J9VMInternals_primitiveClone  	
10060	0.32     	8.28 	objectMonitorExit                            	
9456 	0.30     	7.78 	classHashFn                                  	
8949 	0.28     	7.36 	copyCharsIntoUTF8Helper                      	
7936 	0.25     	6.53 	bytecodeLoop                                 	
7574 	0.24     	6.23 	classHashEqualFn                             	
6373 	0.20     	5.24 	objectMonitorEnterNonBlocking                	
5559 	0.18     	4.57 	getStringUTF8Length                          	
5061 	0.16     	4.16 	instanceOfOrCheckCast                        	
3213 	0.10     	2.64 	getIntField                                  	
2825 	0.09     	2.32 	setArrayRegion                               	
2775 	0.09     	2.28 	hashTableFind                                	
2565 	0.08     	2.11 	getInterfacesHelper                          	
2255 	0.07     	1.86 	getArrayRegion                               	
1444 	0.05     	1.19 	internalFindClassInModule                    	
1284 	0.04     	1.06 	walkFrame                                    	
1281 	0.04     	1.05 	walkStackFrames                              	
1114 	0.04     	0.92 	Fast_java_lang_System_currentTimeMillis      	
1049 	0.03     	0.86 	Fast_java_lang_reflect_Array_newArrayImpl    	
856  	0.03     	0.70 	getStringUTFRegion                           	
850  	0.03     	0.70 	getStringLength                              	
```

The `getStringUTF8Length` can be accelerated when String compression is enabled. Furthermore there seemed to be a lot of ticks in `copyCharsIntoUTF8Helper` presumably coming from the `getStringUTFRegion` JNI interface. The disassembly for this function showed that there are many cold parts in the function we never execute, predominantly related to checking of `J9_STR_XLAT` and null termination. The compiler does not version these outside the loop. The main beneficial change here is to manually version these out of the tight for loop to allow for a very fast sequence.

We further simplify APIs at this point, avoid unnecessary computations by introducing additional parameters, and add buffering where possible. There is still room for improvement when it comes to String compression which will be handled in a separate PR.

The profile following the change shows a reduction in overall time spent in the refactored and improved functions:

```
Ticks	Pct_Total	%    	Symbol/Functions                             	
16762	0.53     	14.24	Fast_java_lang_J9VMInternals_identityHashCode	
13563	0.43     	11.53	Fast_java_lang_J9VMInternals_primitiveClone  	
10703	0.34     	9.10 	objectMonitorExit                            	
9179 	0.29     	7.80 	classHashFn                                  	
7482 	0.24     	6.36 	classHashEqualFn                             	
6327 	0.20     	5.38 	objectMonitorEnterNonBlocking                	
8647 	0.19     	5.05 	copyStringToUTF8Helper                       	
5701 	0.18     	4.84 	getStringUTF8Length                          	
5007 	0.16     	4.25 	instanceOfOrCheckCast                        	
3970 	0.13     	3.37 	bytecodeLoop                                 	
3330 	0.11     	2.83 	getIntField                                  	
2811 	0.09     	2.39 	getInterfacesHelper                          	
2793 	0.09     	2.37 	hashTableFind                                	
2530 	0.08     	2.15 	setArrayRegion                               	
2321 	0.07     	1.97 	getArrayRegion                               	
1400 	0.04     	1.19 	internalFindClassInModule                    	
1221 	0.04     	1.04 	walkFrame                                    	
1199 	0.04     	1.02 	walkStackFrames                              	
1132 	0.04     	0.96 	Fast_java_lang_System_currentTimeMillis      	
1030 	0.03     	0.88 	Fast_java_lang_reflect_Array_newArrayImpl    	
821  	0.03     	0.70 	getStringUTFRegion                           	
740  	0.02     	0.63 	getStringUTFLength                           	
```

fjeremic:
I'll have to do copyright updates it seems, but I'll wait for technical reviews first and bundle the changes at once. I kindly request a review from anyone knowledgeable in this area of code.

runtime/vm/stringhelpers.cpp
```diff
@@ -258,131 +258,67 @@ compareStringToUTF8(J9VMThread *vmThread, j9object_t string, UDATA translateDots
...
-{
-	Assert_VM_notNull(string);
-
-	UDATA unicodeLength = J9VMJAVALANGSTRING_LENGTH(vmThread, string);
-	j9object_t unicodeBytes = J9VMJAVALANGSTRING_VALUE(vmThread, string);
-
-	return copyCharsIntoUTF8Helper(vmThread, IS_STRING_COMPRESSED(vmThread, string), nullTermination, stringFlags, unicodeBytes, 0, unicodeLength, utf8Data, utf8Length);
-}
+			for (UDATA i = 0; i < unicodeLength; i++) {
+				UDATA encodedLength = VM_VMHelpers::encodedUTF8Length(J9JAVAARRAYOFCHAR_LOAD(vmThread, unicodeBytes, i));
```
JasonFengJ9: Can `stringFlags` check along with potential `. --> /` replacement be performed within last `for` loop to avoid duplicated call of `J9JAVAARRAYOFCHAR_LOAD()`?

runtime/vm/stringhelpers.cpp
```diff
@@ -258,131 +258,67 @@ compareStringToUTF8(J9VMThread *vmThread, j9object_t string, UDATA translateDots
 }
 
 #if !defined (J9VM_OUT_OF_PROCESS)
-
 /**
- * Copy certain number of Unicode characters from an offset into a Unicode character array to a UTF8 data buffer.
+ * Copy a Unicode String to a UTF8 data buffer with NULL termination. This function makes an assumption that the
+ * caller has guaranteed the buffer is of the appropriate size to fit the encoded UTF8 string.
```
JasonFengJ9: Pls specify the buffer size according to incoming string which can guaranteed it fits the encoded string.

runtime/vm/stringhelpers.cpp
```diff
@@ -396,41 +332,39 @@ copyFromStringIntoUTF8(J9VMThread * vmThread, j9object_t string, char * dest)
...
 		}
-		if (UDATA_MAX == copyStringToUTF8Helper(
-			vmThread, string, TRUE, stringFlags, (U_8 *)(strUTF + prependStrLen), length - prependStrLen)
-		) {
-			if (buffer != strUTF) {
-				j9mem_free_memory(strUTF);
-			}
-			return NULL;
+
+		UDATA computedUtf8Length = copyStringToUTF8Helper(vmThread, string, stringFlags, (U_8 *)(strUTF + prependStrLength));
```
JasonFengJ9: Defining a variable not at the beginning of the code block will cause compilation failure in certain platforms (zOS).

runtime/vm/stringhelpers.cpp
```diff
@@ -396,41 +332,39 @@ copyFromStringIntoUTF8(J9VMThread * vmThread, j9object_t string, char * dest)
...
+copyStringToUTF8WithMemAlloc(J9VMThread *vmThread, j9object_t string, UDATA stringFlags, const char *prependStr, UDATA prependStrLength, char *buffer, UDATA bufferLength, UDATA *utf8Length)
 {
 	Assert_VM_notNull(prependStr);
 	Assert_VM_notNull(string);
 
 	char *strUTF = NULL;
-	UDATA stringLen = getStringUTF8Length(vmThread, string);
-	UDATA prependStrLen = strlen(prependStr);
-	UDATA length = stringLen + prependStrLen + 1;
+	UDATA stringLen = J9VMJAVALANGSTRING_LENGTH(vmThread, string) * 3;
```
JasonFengJ9: Since this assume `3 bytes` encoding for every `char`, pls add a comment to warn this API user that allocating incoming buffer accordingly to avoid another `j9mem_allocate_memory` in this method.

runtime/vm/stringhelpers.cpp
```diff
@@ -396,41 +332,39 @@ copyFromStringIntoUTF8(J9VMThread * vmThread, j9object_t string, char * dest)
  * @param[in] stringFlags the flag to determine performing '.' --> '/'
  * @param[in] prependStr the string to be prepended before the string object to be copied
  * 				it can't be NULL but can be an empty string ""
+ * @param[in] prependStrLength The length of prependStr as computed by strlen.
  * @param[in] buffer the buffer for the string
  * @param[in] bufferLength the buffer length, not expected to larger than 64K
+ * @param[out] utf8Length If not NULL returns the computed length (in bytes) of the copied UTF8 string in the buffer excluding the NULL terminator.
  *
  * @return a char pointer to the string (with NULL termination)
  */
```
JasonFengJ9: Pls remove these function comments, they are in head file `vm_api.h`, not duplicate them here. (asking you do the cleanup since you touched it :)

runtime/vm/stringhelpers.cpp
fjeremic: I'm not sure what you mean by this. The `stringFlags` check is performed before the last for loop. We cannot avoid `J9JAVAARRAYOFCHAR_LOAD` here because we only replace the first byte of each individual code point, so we need to traverse the array by code points not by bytes.

runtime/vm/stringhelpers.cpp
fjeremic: Doesn't this comment say exactly this? We give responsibility to the caller to provide us a buffer for the string which is guaranteed to fit any encoding of the specified length.

runtime/vm/stringhelpers.cpp
fjeremic: > Defining a variable not at the beginning of the code block will cause compilation failure in certain platforms (zOS).

This is a statement valid for source files compiled with a C compiler. This is a C++ file. There are other examples of declarations in the middle of a scoped block in the same file here:

https://github.com/eclipse/openj9/blob/b8d557446f0eb50dca4ade4b0ff5002f30193d52/runtime/vm/stringhelpers.cpp#L176-L180

runtime/vm/stringhelpers.cpp
JasonFengJ9: Current code:
````
		for (i = unicodeOffset; i < (unicodeOffset + unicodeLength); i++) {
			result = VM_VMHelpers::encodeUTF8CharN(J9JAVAARRAYOFCHAR_LOAD(vmThread, unicodeBytes, i), data, (U_32)remaining);
			if (0 == result) {
				return UDATA_MAX;
			} else {
				if ((stringFlags & J9_STR_XLAT) && ('.' == *data)) {
					*data = '/';
				}
				remaining -= result;
				data += result;
			}
		}
````

This Pull Request:
````
		for (UDATA i = 0; i < unicodeLength; i++) {
			data += VM_VMHelpers::encodeUTF8Char(J9JAVAARRAYOFCHAR_LOAD(vmThread, unicodeBytes, i), data);
		}

		if (stringFlags & J9_STR_XLAT) {
			data = utf8Data;

			for (UDATA i = 0; i < unicodeLength; i++) {
				UDATA encodedLength = VM_VMHelpers::encodedUTF8Length(J9JAVAARRAYOFCHAR_LOAD(vmThread, unicodeBytes, i));

				if ('.' == *data) {
					*data = '/';
				}

				data += encodedLength;
			}
		}
````
Why introduce an extra loop (and duplicated code) to handle `J9_STR_XLAT` which might hurt startup performance? 
Does it intend to avoid comparison of `J9_STR_XLAT` inside of first loop and improve the application runtime?
If so, I would suggest to make two version of `copyCharsIntoUTF8HelperXXX`, one for `J9_STR_XLAT` check during startup, the other without `J9_STR_XLAT` for application runtime, and they could share a helper invoking `VM_VMHelpers::encodeUTF8Char`.
Does it make sense?

File: runtime/vm/stringhelpers.cpp (hunk 5 above)
Comment (fjeremic) [reply runtime/vm/stringhelpers.cpp]: Fixed in 84f8304

File: runtime/vm/stringhelpers.cpp (hunk 4 above)
Comment (fjeremic) [reply runtime/vm/stringhelpers.cpp]: Fixed in 84f8304

File: runtime/vm/stringhelpers.cpp (hunk 1 above)
Comment (fjeremic) [reply runtime/vm/stringhelpers.cpp]: As per my performance evaluation comment (https://github.com/eclipse/openj9/pull/1047#issuecomment-361365752)

> The compiler does not version these outside the loop. The main beneficial change here is to manually version these out of the tight for loop to allow for a very fast sequence.

I dislike the idea of introducing another `copyCharsIntoUTF8HelperXXX` just to handle the case of `J9_STR_XLAT`. If we go down that route should we not specialize every single function which takes in any sort of flag since we mostly pass flags as constants? At what point do you make the cutoff then?

How about this best of both worlds:

```
if ((stringFlags & J9_STR_XLAT) == 0) {
    for (UDATA i = 0; i < unicodeLength; i++) {
        data += VM_VMHelpers::encodeUTF8Char(J9JAVAARRAYOFCHAR_LOAD(vmThread, unicodeBytes, i), data);
    }
} else {
    for (UDATA i = 0; i < unicodeLength; i++) {
        UDATA encodedLength = VM_VMHelpers::encodedUTF8Length(J9JAVAARRAYOFCHAR_LOAD(vmThread, unicodeBytes, i));

        if ('.' == *data) {
            *data = '/';
        }

        data += encodedLength;
    }
}
```

This eliminates the extra `J9JAVAARRAYOFCHAR_LOAD` in the second for loop and we still perform fast in both cases. Thoughts?

Comment (JasonFengJ9) [reply runtime/vm/stringhelpers.cpp]: This works. Some comments about this arrangement will ensure that no _optimization_ to remove this optimization.

File: runtime/vm/stringhelpers.cpp (hunk 3 above)
Comment (JasonFengJ9) [reply runtime/vm/stringhelpers.cpp]: Agreed though the example is not accurate in which the variables were defined at beginning of a code block `{}`.
For the sake of consistency and avoiding confusion, I suggest moving the variable definition just four lines above.

Comment (fjeremic) [reply runtime/vm/stringhelpers.cpp]: Fixed in bced3dc

File: runtime/vm/stringhelpers.cpp (hunk 1 above)
Comment (fjeremic) [reply runtime/vm/stringhelpers.cpp]: Fixed in bced3dc

File: runtime/j9vm/j7vmi.c [hunk 6]
```diff
@@ -2559,23 +2560,13 @@ jvmDefineClassHelper(JNIEnv *env, jobject classLoaderObject,
 	vmFuncs->internalEnterVMFromJNI(currentThread);
 
 	/* Allocate and initialize a UTF8 copy of the Unicode class-name */
-	if (className == NULL) {
-		utf8Length = 0;
-		utf8Name = NULL;
-	} else {
-		utf8Length = vmFuncs->getStringUTF8Length(currentThread, J9_JNI_UNWRAP_REFERENCE(className));
+	if (className != NULL) {
```
JasonFengJ9: Pls put rvalue on the left for inequality comparison.

runtime/jcl/common/java_dyn_methodhandle.c
```diff
@@ -924,15 +924,10 @@ allocateJ9UTF8(JNIEnv *env, jstring name)
 
 	PORT_ACCESS_FROM_ENV(env);
 
-	utf8 = j9mem_allocate_memory(sizeof(U_16) + length, J9MEM_CATEGORY_VM_JCL);
+	utf8 = j9mem_allocate_memory(sizeof(U_16) + length + 1, J9MEM_CATEGORY_VM_JCL);
```
JasonFengJ9: This is an extra byte for `NULL` not required by `J9UTF8`. Can we avoid this waste via a parameter to `copyStringToUTF8Helper` like before? I understand this might complicate the manual versioning inside the method but that should not allow an additional unused byte for each `allocateJ9UTF8` call.

runtime/jcl/common/java_lang_Class.cpp
```diff
@@ -901,25 +901,27 @@ Java_java_lang_Class_getMethodImpl(JNIEnv *env, jobject recv, jobject name, jobj
 
 			J9JNINameAndSignature nameAndSig;
 			char nameBuffer[J9VM_PACKAGE_NAME_BUFFER_LENGTH];
+			UDATA nameBufferLength;
```
JasonFengJ9: Uninitialized variable.

runtime/jcl/common/java_lang_Class.cpp
```diff
@@ -901,25 +901,27 @@ Java_java_lang_Class_getMethodImpl(JNIEnv *env, jobject recv, jobject name, jobj
...
 			char signatureBuffer[J9VM_PACKAGE_NAME_BUFFER_LENGTH];
+			UDATA signatureLength;
```
JasonFengJ9: Same here.

runtime/jcl/common/jcldefine.c
```diff
@@ -88,23 +89,14 @@ defineClassCommon(JNIEnv *env, jobject classLoaderObject,
 	}
 
 	/* Allocate and initialize a UTF8 copy of the Unicode class-name */
-	if (className == NULL) {
-		utf8Length = 0;
-		utf8Name = NULL;
-	} else {
-		utf8Length = vmFuncs->getStringUTF8Length(currentThread, J9_JNI_UNWRAP_REFERENCE(className));
+	if (className != NULL) {
```
JasonFengJ9: rvalue on the left pls.

runtime/vm/jnimisc.cpp
```diff
@@ -792,11 +792,7 @@ getStringUTFChars(JNIEnv *env, jstring string, jboolean *isCopy)
 		gpCheckSetNativeOutOfMemoryError(currentThread, 0, 0);
 	} else {
 		JAVA_OFFLOAD_SWITCH_ON_WITH_REASON_IF_LIMIT_EXCEEDED(currentThread, J9_JNI_OFFLOAD_SWITCH_GET_STRING_UTF_CHARS, utfLength);
-		if (UDATA_MAX != copyStringToUTF8Helper(currentThread, stringObject, TRUE, J9_STR_NONE, utfChars, utfLength)) {
-			if (NULL != isCopy) {
-				*isCopy = JNI_TRUE;
-			}
-		}
+		copyStringToUTF8Helper(currentThread, stringObject, J9_STR_NONE, utfChars);
```
JasonFengJ9: Setting `isCopy` code is missing.

runtime/vm/stringhelpers.cpp
```diff
@@ -258,179 +258,95 @@ compareStringToUTF8(J9VMThread *vmThread, j9object_t string, UDATA translateDots
...
+ * @param[in] string a string object to be copied, it can't be NULL
  * @param[in] stringFlags the flag to determine performing '.' --> '/'
- * @param[in] unicodeBytes a Unicode character array
- * @param[in] unicodeOffset an offset into the Unicode character array
- * @param[in] unicodeLength the number of Unicode characters to be copied
  * @param[in] utf8Data a utf8 data buffer
- * @param[in] utf8Length the size of the utf8 data buffer
  *
- * @return UDATA_MAX if a failure occurred, otherwise the number of utf8 data copied excluding null termination
+ * @return The computed length (in bytes) of the copied UTF8 string in the buffer excluding the NULL terminator.
```
JasonFengJ9: Pls remove the comment above, it is only required in the header file `vm_api.h`.

JasonFengJ9: Pls rebase for next commit.

runtime/j9vm/j7vmi.c
fjeremic: Fixed in 1baa27d

runtime/jcl/common/java_lang_Class.cpp
fjeremic: Fixed in 1baa27d

runtime/jcl/common/java_lang_Class.cpp
fjeremic: Fixed in 1baa27d

runtime/jcl/common/jcldefine.c
fjeremic: Fixed in 1baa27d

runtime/vm/jnimisc.cpp
fjeremic: Fixed in 1baa27d. I wonder why no tests failed due to this bug which was introduced by me. Thank you very much for catching this! I think we need more testing here.

fjeremic: Opened https://github.com/eclipse/openj9/issues/1102 to track this.

runtime/vm/stringhelpers.cpp
fjeremic: Fixed in 1baa27d

runtime/j9vm/java9vmi.c
```diff
@@ -1697,23 +1698,18 @@ JVM_GetModuleByPackageName(JNIEnv *env, jobject classLoader, jstring packageName
...
-		J9UTF8_SET_LENGTH(packageUTF8, (U_16) length);
-		if (NULL != strchr((const char*)J9UTF8_DATA(packageUTF8), '.')) {
+		
+		if (NULL != strchr((const char*)packageName, '.')) {
 			vmFuncs->setCurrentExceptionUTF(currentThread, J9VMCONSTANTPOOL_JAVALANGILLEGALARGUMENTEXCEPTION, "package name contains '.' instead of '/'");
 		} else {
-			if (vmFuncs->isAnyClassLoadedFromPackage(thisVMClassLoader, J9UTF8_DATA(packageUTF8), J9UTF8_LENGTH(packageUTF8))) {
-				vmModule = vmFuncs->findModuleForPackageUTF8(currentThread, thisVMClassLoader, packageUTF8);
+			if (vmFuncs->isAnyClassLoadedFromPackage(thisVMClassLoader, packageName, packageLength)) {
+				vmModule = vmFuncs->findModuleForPackage(currentThread, thisVMClassLoader, packageName, (U_32)packageLength);
```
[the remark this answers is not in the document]
fjeremic: The APIs seem orthogonal to me. For example why does `isAnyClassLoadedFromPackage` accept a `U_8*` and `UDATA` to represent a UTF8 string and a length and `findModuleForPackageUTF8` needs a `J9UTF8*`? Why aren't they all synchronized on one set of APIs. Otherwise we get into these situations where we need to mix and match.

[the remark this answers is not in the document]
fjeremic: Fixed in d05ffd5

runtime/jcl/common/java_dyn_methodhandle.c
fjeremic: Fixed in 4cc4617

fjeremic:
@JasonFengJ9 all review comments have been addressed. I'll rebase after you have approved or added more reviews as it makes it much easier to track which issues have been addressed.

runtime/j9vm/java9vmi.c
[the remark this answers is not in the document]
JasonFengJ9: `J9UTF8` structure is a `U_8*` string (without `null` terminator) and a length at beginning of the structure. `J9UTF8` requires `J9UTF8_DATA`, `J9UTF8_SET_LENGTH` and `J9UTF8_LENGTH` for accessing its fields. There are cases that the APIs expect the data and length without going through the getter/setter. In all these scenarios, the data are not ended with `null` and the end are determined by `LENGTH`. 
Hope this makes sense.

runtime/jvmti/jvmtiThread.c
```diff
@@ -442,22 +442,15 @@ jvmtiGetThreadInfo(jvmtiEnv* env,
...
-				/* Be sure to allocate at least one byte for the nul termination */
-
-				name = j9mem_allocate_memory(threadNameLen, J9MEM_CATEGORY_JVMTI_ALLOCATE);
-				if (name == NULL) {
-					/* failed to allocate memory, so release VMTthread and exit */
-					rc = JVMTI_ERROR_OUT_OF_MEMORY;
-					goto done;
-				}
 
 				if (threadName == NULL) {
```
JasonFengJ9: Nitpick: pls change this existing code to put rvalue on the left.

JasonFengJ9:
Thanks @fjeremic, it looks good to me except couple of minor things (as per comments supplied).
Note: this PR need blessings from @DanHeidinga or @pshipton before being merged.

runtime/jvmti/jvmtiThread.c
fjeremic: Fixed in 738e980

fjeremic:
> Thanks @fjeremic, it looks good to me except couple of minor things (as per comments supplied).

Awesome! I think I've fixed the last request. I've rebased all the changes now on the latest master branch. It would be wonderful if this PR could get a "Review Approval" from you if you are satisfied with the changes so it is documented for the future. Thanks!

runtime/j9vm/java9vmi.c
```diff
@@ -381,11 +381,11 @@ trcModulesCreationPackage(J9VMThread * currentThread, J9Module * fromModule, j9o
 	J9InternalVMFunctions const * const vmFuncs = currentThread->javaVM->internalVMFunctions;
 	char moduleNameBuf[J9VM_PACKAGE_NAME_BUFFER_LENGTH];
 	char *moduleNameUTF = vmFuncs->copyStringToUTF8WithMemAlloc(
-		currentThread, fromModule->moduleName, J9_STR_NONE, "", moduleNameBuf, J9VM_PACKAGE_NAME_BUFFER_LENGTH);
+		currentThread, fromModule->moduleName, J9_STR_NONE, "", 0, moduleNameBuf, J9VM_PACKAGE_NAME_BUFFER_LENGTH, NULL);
```
JasonFengJ9: `setNativeOutOfMemoryError(currentThread, 0, 0)` need to be called if this returns `null`. (I know it missed such check at first place)

fjeremic: Fixed in 76f520f

runtime/j9vm/java9vmi.c
```diff
@@ -381,11 +381,11 @@ trcModulesCreationPackage(J9VMThread * currentThread, J9Module * fromModule, j9o
...
 #if J9VM_JAVA9_BUILD < 156
 	char packageNameBuf[J9VM_PACKAGE_NAME_BUFFER_LENGTH];
 	char *packageNameUTF = vmFuncs->copyStringToUTF8WithMemAlloc(
```
JasonFengJ9: `null` (native OOM) check for `packageNameUTF` was missing as well. There are quite a few similar cases in this file as well.
How about let API itself set `setNativeOutOfMemoryError` when `j9mem_allocate_memory` failed?
P.S. rebasing seems bring up more questions :)

fjeremic: @JasonFengJ9 this seems like pre-existing bugs in the code. I think we are starting to scope creep in this PR. This PR is targeted towards performance improvements of String related APIs. May I suggest we focus only on those? We can open up issues for related code in the area of these changes which need additional fixes. Does that sound reasonable?

I'm afraid there will be more and more places to fix pre-existing functional issues not related to this _performance_ change. I suggest we leave these for a subsequent PR.

runtime/util/cphelp.c
```diff
@@ -149,15 +149,13 @@ getModuleJRTURL(J9VMThread *currentThread, J9ClassLoader *classLoader, J9Module
...
+			jrtURL = j9mem_allocate_memory(sizeof(jrtURL->length) + jrtURLLen + 1, OMRMEM_CATEGORY_VM);
 			if (NULL == jrtURL) {
 				goto _exit;
 			}
 			strncpy((char *)J9UTF8_DATA(jrtURL), JRT_URL_PROTOCOL, sizeof(JRT_URL_PROTOCOL) - 1);
 			nameLocation = (char *)J9UTF8_DATA(jrtURL) + sizeof(JRT_URL_PROTOCOL) - 1;
-			if (UDATA_MAX == vmFuncs->copyStringToUTF8Helper(currentThread, module->moduleName, FALSE, J9_STR_NONE, (U_8*)nameLocation, jrtURLLen - sizeof(JRT_URL_PROTOCOL) + 1)) {
-				goto _exit;
-			}
+			vmFuncs->copyStringToUTF8Helper(currentThread, module->moduleName, J9_STR_NONE, (U_8*)nameLocation, TRUE);
```
JasonFengJ9: I understand that this API makes an assumption "that the caller has guaranteed the buffer is of the appropriate size to fit the encoded UTF8 string." However it might not always obvious to caller what's the appropriate size. In case there is a mis-calculation, the API doesn't provide a safe net and a buffer overflow could happen.
I feel this API need have some kind of error detection mechanism in place just like before the caller passes the buffer size and the method returns error if it is not sufficient.

fjeremic: Fixed in f1d4863

fjeremic: > I feel this API need have some kind of error detection mechanism in place just like before the caller passes the buffer size and the method returns error if it is not sufficient.

By the way any such detection will involve a test within the tight for loop which is critical for performance. I don't see at this moment how this can be done for free. The major purpose of this PR was to make that loop as tight as possible since it does occur in steady state during various benchmark runs. We need to be cautious of the performance overhead vs. debugging checks.

andrewcraik: The usual middle ground in the JIT would be some kind of assert that is not enabled in production builds, but easily enabled in development builds for early fault detection. Perhaps that might be a compromise here?

fjeremic: > The usual middle ground in the JIT would be some kind of assert that is not enabled in production builds, but easily enabled in development builds for early fault detection. Perhaps that might be a compromise here?

Sure we can do that. @JasonFengJ9 do you agree? If so could you provide me with a debug flag I can use that is off in production builds?

fjeremic: Alternatively there seems to be only three uses of `copyStringToUTF8Helper` in the VM now. We can easily get rid of them all and just deprecate the entire API. This is also on the table.

JasonFengJ9: @fjeremic VM has low cost check like `Assert_VM_true` which runs in production as well, and it doesn't seem it can help performance in this case. Considering the buffer size depends on the incoming string at runtime, performing a memory copy without boundary check at production builds could introduce a fatal error.
Maybe it is better to remove these API references.

runtime/vm/stringhelpers.cpp
```diff
@@ -258,179 +258,96 @@ compareStringToUTF8(J9VMThread *vmThread, j9object_t string, UDATA translateDots
...
+	UDATA length = stringLen + prependStrLength;
+
+	if ((stringFlags & J9_STR_NULL_TERMINATE_RESULT) != 0) {
+		++length;
+	}
 
 	PORT_ACCESS_FROM_VMC(vmThread);
 
 	if (length > bufferLength) {
 		strUTF = (char *)j9mem_allocate_memory(length, OMRMEM_CATEGORY_VM);
```
JasonFengJ9: How often do we expect this allocation happens? i.e., the caller doesn't allocate enough memory and this method actually allocates maximum buffer according to the string length.

runtime/vm/stringhelpers.cpp
```diff
@@ -258,179 +258,96 @@ compareStringToUTF8(J9VMThread *vmThread, j9object_t string, UDATA translateDots
...
- */
-UDATA
-copyStringToUTF8(J9VMThread *vmThread, j9object_t string, UDATA stringFlags, U_8 *utf8Data, UDATA utf8Length)
-{
-	if (UDATA_MAX == copyStringToUTF8Helper(vmThread, string, TRUE, stringFlags, utf8Data, utf8Length)) {
-		return 1;
+		Assert_VM_true(utf8DataLength >= returnLength + 1);
 	} else {
-		return 0;
+		Assert_VM_true(utf8DataLength >= returnLength);
```
JasonFengJ9: I don't believe it should be fatal when incoming buffer is not sufficient which is recoverable.

fjeremic: Actually it is not. This API still assumes the caller has ensured the buffer is large enough. If this is not the case the prior loop will have overrun the buffer and clobbered arbitrary data after the buffer. This assert is a result from a previous review where it was mentioned there needs to be a safeguard. This assert is the said safeguard which makes sure if a buffer overflow did occur we will assert right away to let the user know.

JasonFengJ9: Right, that's what I meant fatal. `Assert_VM_true` is going to effectively kill JVM when the buffer is not large enough. Previous implementation can recover from this scenario via allocating larger buffer.
It appears to me this is not a robust way for string copying.

runtime/vm/stringhelpers.cpp
fjeremic: There is only a few places which do pass a `NULL` value for `buffer`. If you look at the rest of the commit all the other places previously allocated `string.length * 3` bytes anyway which is what this API now does. Further to that many of the places that allocated the 3x bytes have been modified to now use a stack buffer so we will actually have less allocations.

I think a lot of this was done in 30111e3.

runtime/vm/stringhelpers.cpp
fjeremic: > Previous implementation can recover from this scenario via allocating larger buffer.

This is not true. Here is the previous implementation in the repository right now:

https://github.com/eclipse/openj9/blob/15c192e26849a745204f73456a33a765e5b5ec3f/runtime/vm/stringhelpers.cpp#L277-L322

There are no allocations in this function.

fjeremic:
I've rebased the change on the latest master commit.

runtime/vm/stringhelpers.cpp
fjeremic: Perhaps we can rename to `copyCharsIntoUTF8Unsafe` and further emphasize the caller is responsible for ensuring a properly sized buffer?

fjeremic: I would also like to move this function to the VMHelpers.hpp and make it inlinable as we can squeeze extra performance if the compiler is able to inline it.

JasonFengJ9: Comparing with existing implementation, this PR introduces two new behaviors, one is that a buffer overflow copying could happen which is not desirable, the other is that JVM stops right away while existing code sets an internal error and allows a grace existing.

fjeremic: > this PR introduces two new behaviors, one is that a buffer overflow copying could happen which is not desirable

Hence the assert so the caller can fix their arguments so this does not occur. This would be a development time bug that will always be caught. This assert will never fire off in production and if it does it indicates a very serious bug which needs to be fixed with the highest priority anyway. The major point of this change is to eliminate the checks in the tight for loop validating at each iteration that a buffer overflow did not occur. There were many locations that used to call this API which did not even check the return value and just "assumed things will work". If anything the assert makes the world a better place by forcing the callers hand and making sure correct arguments are passed through.

> the other is that JVM stops right away while existing code sets an internal error and allows a graceful exit.

Where do you see the internal error and graceful exit happening in the above snippet? All I see is a return code being returned and you have no guarantee the caller will ever check it. Again these are development time issues for people working on the VM and using this API incorrectly.

JasonFengJ9: > This assert will never fire off in production and if it does it indicates a very serious bug which needs to be fixed with the highest priority anyway

If there is no buffer overflow, it might not be serious or dangerous as it could be. 

You are right about some examples out of ~20 current usages of `copyStringToUTF8Helper` that didn't check the return value correctly, which doesn't mean it is correct. Some other usages have following pattern:
```
		if (UDATA_MAX == vmFuncs->copyStringToUTF8Helper(
			currentThread, J9_JNI_UNWRAP_REFERENCE(className), TRUE, J9_STR_XLAT, utf8Name, utf8Length + 1)
		) {
			vmFuncs->setCurrentException(currentThread, J9VMCONSTANTPOOL_JAVALANGINTERNALERROR, NULL);
			goto done;
		}
```

I think it is possible to use `Assert_VM_true` (kill JVM right way) instead of setting an internal error. However my feeling is that we should avoid potential buffer overflow. The performance gain over security risk doesn't sound right.

JasonFengJ9: Not sure if the usage case within `jnimisc.cpp` will be the only one in future.
Can we introduce a less performing but safe version `copyCharsIntoUTF8Helper` and use it in scenarios like `jnimisc.cpp`?

fjeremic: > Can we introduce a less performing but safe version of copyCharsIntoUTF8Helper and use it in scenarios like jnimisc.cpp?

This will be even more code duplication then just manually inlining the current implementation in that spot and getting rid of all the junk we don't need. This way no "unsafe" API will exist at all. I cannot forecast whether or not there will be more use cases as the none in jnimisc.cpp in the future. I propose we do not make such assumptions and let that future change introduce whatever safe helper is needed.

fjeremic: Incidentally the use in jnimisc.cpp is the primary one introducing ticks in the profile as you can see this JNI API in the list of methods quoted in https://github.com/eclipse/openj9/pull/1047#issuecomment-361365752, so using a "safe" API in this location is counter intuitive to the purpose of this PR (to speed up performance).

JasonFengJ9: I am fine to postpone the API creation when it is needed later.

JasonFengJ9: From code history, this JNI API assumes the buffer is large enough and didn't check the return value before. I guess your PR doesn't make it worse hence I am not going to object it.

runtime/vm/stringhelpers.cpp
```diff
@@ -258,179 +258,96 @@ compareStringToUTF8(J9VMThread *vmThread, j9object_t string, UDATA translateDots
...
 	if (length > bufferLength) {
 		strUTF = (char *)j9mem_allocate_memory(length, OMRMEM_CATEGORY_VM);
 	} else {
 		strUTF = buffer;
-		length = bufferLength;
 	}
 	if (NULL != strUTF) {
-		if (0 < prependStrLen) {
-			memcpy(strUTF, prependStr, prependStrLen);
+		UDATA computedUtf8Length;
```
JasonFengJ9: Variable initialization is needed.

runtime/j9vm/j7vmi.c
```diff
@@ -2559,23 +2560,13 @@ jvmDefineClassHelper(JNIEnv *env, jobject classLoaderObject,
 	vmFuncs->internalEnterVMFromJNI(currentThread);
 
 	/* Allocate and initialize a UTF8 copy of the Unicode class-name */
-	if (className == NULL) {
-		utf8Length = 0;
-		utf8Name = NULL;
-	} else {
-		utf8Length = vmFuncs->getStringUTF8Length(currentThread, J9_JNI_UNWRAP_REFERENCE(className));
+	if (NULL != className) {
 		/* For now, use malloc, but this should use stack allocation for short class-names */
```
JasonFengJ9: This comment can be removed now.

runtime/vm/stringhelpers.cpp
fjeremic: Should we still deprecate `copyCharsIntoUTF8Helper` or rename to `copyCharsIntoUTF8Unsafe` or leave it as is? I'm fine with whichever given there are only two uses of it.

JasonFengJ9: I am leaning forward to leave it as is, i.e., kill the JVM and throw assertion failure when a buffer overflow occurs. This is a new behavior comparing with previous releases. So ensure run as much tests as possible.

runtime/vm/stringhelpers.cpp
fjeremic: Fixed in 8722de7

runtime/j9vm/j7vmi.c
fjeremic: Fixed in 8722de7

fjeremic:
@DanHeidinga @pshipton just making sure this didn't run past your radar.

runtime/vm/stringhelpers.cpp
```diff
@@ -258,179 +258,96 @@ compareStringToUTF8(J9VMThread *vmThread, j9object_t string, UDATA translateDots
...
 	Assert_VM_notNull(string);
 
 	char *strUTF = NULL;
-	UDATA stringLen = getStringUTF8Length(vmThread, string);
-	UDATA prependStrLen = strlen(prependStr);
-	UDATA length = stringLen + prependStrLen + 1;
+	UDATA stringLen = J9VMJAVALANGSTRING_LENGTH(vmThread, string) * 3;
+	UDATA length = stringLen + prependStrLength;
+
+	if ((stringFlags & J9_STR_NULL_TERMINATE_RESULT) != 0) {
```
pshipton: Its preferred to use J9_ARE_ALL_BITS_SET(stringFlags, J9_STR_NULL_TERMINATE_RESULT)

runtime/vm/stringhelpers.cpp
```diff
@@ -258,179 +258,96 @@ compareStringToUTF8(J9VMThread *vmThread, j9object_t string, UDATA translateDots
...
 		}
-		if (UDATA_MAX == copyStringToUTF8Helper(
-			vmThread, string, TRUE, stringFlags, (U_8 *)(strUTF + prependStrLen), length - prependStrLen)
-		) {
-			if (buffer != strUTF) {
-				j9mem_free_memory(strUTF);
-			}
-			return NULL;
+
+		computedUtf8Length = copyStringToUTF8Helper(vmThread, string, stringFlags, (U_8 *)(strUTF + prependStrLength), length);
```
pshipton: Instead of length this should pass stringLen or bufferLength - prependStrLength, whichever is smaller.

runtime/vm/vmthread.c
```diff
@@ -1964,18 +1964,10 @@ startJavaThreadInternal(J9VMThread * currentThread, UDATA privateFlags, UDATA os
...
-			memset(threadName, 0, unicodeLength);
-			if (UDATA_MAX == copyStringToUTF8Helper(
-				currentThread, unicodeChars, TRUE, J9_STR_NONE, (U_8 *)threadName, unicodeLength)
-			) {
-				j9mem_free_memory(threadName);
-				threadName = NULL;
-			}
+
+		if (NULL != unicodeChars) {
+			threadName = copyStringToUTF8WithMemAlloc(currentThread, unicodeChars, J9_STR_NULL_TERMINATE_RESULT, "", 0, threadName, 0, NULL);
```
pshipton: Should pass NULL instead of threadName.

runtime/jvmti/jvmtiThread.c
```diff
@@ -433,31 +433,24 @@ jvmtiGetThreadInfo(jvmtiEnv* env,
...
-				}
 
-				if (threadName == NULL) {
+				if (NULL == threadName) {
+					name = j9mem_allocate_memory(1, J9MEM_CATEGORY_JVMTI_ALLOCATE);
 					name[0] = '\0';
 				} else {
-					if (UDATA_MAX == vm->internalVMFunctions->copyStringToUTF8Helper(currentThread, threadName, TRUE, J9_STR_NONE, (U_8*)name, threadNameLen)) {
-						rc = JVMTI_ERROR_INTERNAL;
+					name = vm->internalVMFunctions->copyStringToUTF8WithMemAlloc(currentThread, threadName, J9_STR_NULL_TERMINATE_RESULT, "", 0, name, 0, NULL);
```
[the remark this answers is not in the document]
fjeremic: Fixed in a544cba.

runtime/jvmti/jvmtiThreadGroup.c
```diff
@@ -103,16 +101,11 @@ jvmtiGetThreadGroupInfo(jvmtiEnv* env,
 			threadGroupObject = *((j9object_t*) group);
 
 			groupName = J9VMJAVALANGTHREADGROUP_NAME(currentThread, threadGroupObject);
-			/* Allocate the maximum possible bytes per UTF8 character (3 bytes), + 1 for '\0' at the end of the string */
-			nameLen = J9VMJAVALANGSTRING_LENGTH(currentThread, groupName) * 3 + 1;
-			name = j9mem_allocate_memory(nameLen, J9MEM_CATEGORY_JVMTI_ALLOCATE);
-			if (name == NULL) {
+			name = vm->internalVMFunctions->copyStringToUTF8WithMemAlloc(currentThread, groupName, J9_STR_NULL_TERMINATE_RESULT, "", 0, name, 0, NULL);
```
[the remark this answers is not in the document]
fjeremic: Fixed in a544cba.

runtime/vm/stringhelpers.cpp
fjeremic: Fixed in a544cba. Good catch!

fjeremic:
Fixed all noted concerns and rebased to latest master.

runtime/jvmti/jvmtiThreadGroup.c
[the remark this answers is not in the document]
fjeremic: Actuall I'll have to revert this guy due to the continuous build failure:

pshipton:
@fjeremic if you want to squash first, then I'll run some PR testing and merge.

pshipton:
Although I just realized the travis build failure looks related
`22:23:32.125 0x110ce00    j9vm.224    *   ** ASSERTION FAILED ** at stringhelpers.cpp:311: ((utf8DataLength >= returnLength + 1))`

pshipton:
Now the travis build is failing with `15:23:09.859 0x1ec4e00    j9mm.107    *   ** ASSERTION FAILED ** at StringTable.cpp:1048: ((false && (0 < consumed)))`

fjeremic:
Taking a look at this now. Will report back once fixed. This must be a regression introduced recently.

fjeremic:
Jenkins test all

fjeremic:
It's a little odd. Java 8 seems to work fine, but Java 9 is busted. This is definitely a regression as it used to work before. The backtrace indicates the following 

```
#0  0x00007ffff633172a in getUnicodeLength ()
#1  0x00007ffff6333a5f in j9gc_allocStringWithSharedCharData ()
#2  0x00007ffffd653ea5 in resolveStringRef ()
#3  0x00007ffffd5c8b50 in VM_BytecodeInterpreter::run(J9VMThread*) ()
#4  0x00007ffffd5a912b in bytecodeLoop ()
#5  0x00007ffffd666f02 in c_cInterpreter ()
#6  0x00007ffffd617def in sendClinit ()
#7  0x00007ffffd5ca742 in initializeImpl ()
#8  0x00007ffffd5cbd6c in classInitStateMachine ()
#9  0x00007ffffd6549e8 in resolveStaticMethodRefInto ()
#10 0x00007ffffd654b12 in resolveStaticMethodRef ()
#11 0x00007ffffd5b6e08 in VM_BytecodeInterpreter::run(J9VMThread*) ()
```

And it consistently fails in the same place. Debugging this now.

fjeremic:
Ok issues have been addressed in f7200cf. I've made several improvements and removed a bunch of duplicate code.

Jenkins test all

fjeremic:
Jenkins line endings check

fjeremic:
@pshipton we seem to be good to go here. All builds are passing. Any other concerns?

pshipton:
lgtm

pshipton:
Sorry, forgot about the squashing.

fjeremic:
> Sorry, forgot about the squashing.

Ah, no worries. There is nothing really wrong with the commits. All of them are fully documented. I was just worried about polluting the codebase with so many.

PR #1048, opened 2018-01-29
Silent correction for -Xmns and -Xmos is value is too small

dmitripivkine:
- Initial Nursery size -Xmns can not be smaller then two regions.
Adjust it silently if it is a case.
- Initial Tenure size -Xmos can not be smaller then one region.
Adjust it silently if it is a case.
- In case if Initial Tenure size is smaller then six regions default
split Tenure/Nursery 75%/25% can not be set. So set Initial Nursery size
to minimum two regions and assign reminder to Tenure.
- "-Xmn<value>" option is treated as "-Xmns<value> -Xmnx<value>". Modify
error message if it is too small to prevent "-Xmn too large for -Xmn"
message, replace it to " -Xmn too small, must be at least 1 Mbytes"

Signed-off-by: dmitripivkine <Dmitri_Pivkine@ca.ibm.com>


dmitripivkine:
jenkins test sanity

PR #1051, opened 2018-01-29
Updates ROM nest members loading

taliamccormick:
Earlier implementation accidentally kept nest members classref during
ROM class loading, rather than nest members utf8 name.


tajila:
Looks good, @pshipton can you please take a look

PR #1057, opened 2018-01-30
Use checkVisibility for superClass access checking

tajila:
Use checkVisibility for superClass access checking

In ramClass creation when loading super classes and interfaces use
checkvisibility to ensure that the current class has access to the super
class or interface. In java9 we can not rely on the PUBLIC specifier to
know if a class has access.

### DIFF
--- a/runtime/vm/createramclass.cpp
+++ b/runtime/vm/createramclass.cpp
@@ -1509,10 +1509,15 @@ loadSuperClassAndInterfaces(J9VMThread *vmThread, J9ClassLoader *classLoader, J9
 	UDATA packageID, BOOLEAN hotswapping, UDATA classPreloadFlags, J9Class **superclassOut, J9Module *module)
 {
 	J9JavaVM *vm = vmThread->javaVM;
+	J9InternalVMFunctions *vmFuncs = vm->internalVMFunctions;
 	const BOOLEAN isROMClassUnsafe = (J9ROMCLASS_IS_UNSAFE(romClass) != 0);
 	J9UTF8 *className = J9ROMCLASS_CLASSNAME(romClass);
 	J9UTF8 *superclassName = NULL;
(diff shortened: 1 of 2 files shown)


DanHeidinga:
@tajila Does this change the behaviour for java 8?  (I haven't check the PR yet)

tajila:
@DanHeidinga No it doesn't, checkVisibility detects the Java level.

runtime/vm/createramclass.cpp
```diff
@@ -1509,10 +1509,15 @@ loadSuperClassAndInterfaces(J9VMThread *vmThread, J9ClassLoader *classLoader, J9
...
 	J9JavaVM *vm = vmThread->javaVM;
+	J9InternalVMFunctions *vmFuncs = vm->internalVMFunctions;
 	const BOOLEAN isROMClassUnsafe = (J9ROMCLASS_IS_UNSAFE(romClass) != 0);
 	J9UTF8 *className = J9ROMCLASS_CLASSNAME(romClass);
 	J9UTF8 *superclassName = NULL;
 	J9Class *superclass = NULL;
+	J9Class thisClass = {0};
+	thisClass.module = module;
+	thisClass.packageID = packageID;
+	thisClass.romClass = romClass;
```
DanHeidinga: I'm not a fan of this approach.  The on-stack `thisClass`'s address will show up in tracepoint data and be confusing to anyone trying to debug this.  Also, the first thing `checkVisibility` does it try to get the current J9Class which happens to work as the obsolete flag isn't set.

I prefer adding a new helper method to visible.c that handles the lookup using just the module/packageID/romClass and having the existing method call into that one.

We've historical done this by marker the new helper as VMINLINE so the existing code can share a single implementation without incurring the overhead of a second call

pshipton:
Note this change caused #1086

PR #1059, opened 2018-01-30
Move nestmates host loading to join superclass loading

taliamccormick:
Original nest host loading sometimes failed due to shared vTableScratch
for methods in internalCreateRAMClassFromROMClassImpl

This commit moves nest host loading so that it occurs earlier - at the
same time as loading of superclass and interfaces - and therefore avoids
this issue.

### DIFF
--- a/runtime/vm/createramclass.cpp
+++ b/runtime/vm/createramclass.cpp
@@ -142,9 +142,10 @@ static UDATA checkPackageAccess(J9VMThread *vmThread, J9Class *foundClass, UDATA
 static void setCurrentExceptionForBadClass(J9VMThread *vmThread, J9UTF8 *badClassName, UDATA exceptionIndex);
 static BOOLEAN verifyClassLoadingStack(J9VMThread *vmThread, J9ClassLoader *classLoader, J9ROMClass *romClass);
 static void popFromClassLoadingStack(J9VMThread *vmThread);
-static VMINLINE BOOLEAN loadSuperClassAndInterfaces(J9VMThread *vmThread, J9ClassLoader *classLoader, J9ROMClass *romClass, J9Class *elementClass, UDATA packageID, BOOLEAN hotswapping, UDATA classPreloadFlags, J9Class **superclassOut, J9Module *module);
 #if defined(J9VM_OPT_VALHALLA_NESTMATES)
-static J9Class *loadNestHost(J9VMThread *vmThread, J9ClassLoader *classLoader, J9UTF8 *nestHostName, UDATA classPreloadFlags);
+static VMINLINE BOOLEAN loadSuperClassAndInterfaces(J9VMThread *vmThread, J9ClassLoader *classLoader, J9ROMClass *romClass, J9Class *elementClass, UDATA packageID, BOOLEAN hotswapping, UDATA classPreloadFlags, J9Class **superclassOut, J9Module *module, J9Class **nestHostOut);
+#else /* defined(J9VM_OPT_VALHALLA_NESTMATES) */
+static VMINLINE BOOLEAN loadSuperClassAndInterfaces(J9VMThread *vmThread, J9ClassLoader *classLoader, J9ROMClass *romClass, J9Class *elementClass, UDATA packageID, BOOLEAN hotswapping, UDATA classPreloadFlags, J9Class **superclassOut, J9Module *module);
 #endif /* defined(J9VM_OPT_VALHALLA_NESTMATES) */
 static J9Class* internalCreateRAMClassDropAndReturn(J9VMThread *vmThread, J9ROMClass *romClass, J9CreateRAMClassState *state);
 static J9Class* internalCreateRAMClassDoneNoMutex(J9VMThread *vmThread, J9ROMClass *romClass, UDATA options, J9CreateRAMClassState *state);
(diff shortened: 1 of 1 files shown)


runtime/vm/createramclass.cpp
```diff
@@ -1506,7 +1507,11 @@ popFromClassLoadingStack(J9VMThread *vmThread)
  */
 static VMINLINE BOOLEAN
 loadSuperClassAndInterfaces(J9VMThread *vmThread, J9ClassLoader *classLoader, J9ROMClass *romClass, J9Class *elementClass,
-	UDATA packageID, BOOLEAN hotswapping, UDATA classPreloadFlags, J9Class **superclassOut, J9Module *module)
+	UDATA packageID, BOOLEAN hotswapping, UDATA classPreloadFlags, J9Class **superclassOut, J9Module *module
+#if defined(J9VM_OPT_VALHALLA_NESTMATES)
+	, J9Class **nestHostOut
```
taliamccormick: I'm not overjoyed about splitting the function declaration across three lines like this. Would it be better to rewrite from
```
loadSuperClassAndInterfaces(J9VMThread *vmThread, J9ClassLoader *classLoader, J9ROMClass *romClass, J9Class *elementClass,
UDATA packageID, BOOLEAN hotswapping, UDATA classPreloadFlags, J9Class **superclassOut, J9Module *module)
```
to
```
loadSuperClassAndInterfaces(J9VMThread *vmThread, J9ClassLoader *classLoader, J9ROMClass *romClass, J9Class *elementClass,
UDATA packageID, BOOLEAN hotswapping, UDATA classPreloadFlags, J9Class **superclassOut, J9Module *module , J9Class **nestHostOut)
#else /* defined(J9VM_OPT_VALHALLA_NESTMATES) */
loadSuperClassAndInterfacesAndNestHost(J9VMThread *vmThread, J9ClassLoader *classLoader, J9ROMClass *romClass, J9Class *elementClass,
UDATA packageID, BOOLEAN hotswapping, UDATA classPreloadFlags, J9Class **superclassOut, J9Module *module)
#endif /* defined(J9VM_OPT_VALHALLA_NESTMATES) */
```
instead of
```
static VMINLINE BOOLEAN
loadSuperClassAndInterfaces(J9VMThread *vmThread, J9ClassLoader *classLoader, J9ROMClass *romClass, J9Class *elementClass,
	UDATA packageID, BOOLEAN hotswapping, UDATA classPreloadFlags, J9Class **superclassOut, J9Module *module
	, J9Class **nestHostOut
#endif /* defined(J9VM_OPT_VALHALLA_NESTMATES) */
	)
```
or is there an alternate approach?

runtime/vm/createramclass.cpp
```diff
@@ -1604,19 +1609,25 @@ loadSuperClassAndInterfaces(J9VMThread *vmThread, J9ClassLoader *classLoader, J9
...
 	}
+#if defined(J9VM_OPT_VALHALLA_NESTMATES)
+	{
+		J9UTF8 *nestHostName = J9ROMCLASS_NESTHOSTNAME(romClass);
 
-	return TRUE;
-}
+		/* Only loads nest host if one is named*/
+		if (NULL != nestHostName) {
+			*nestHostOut = internalFindClassUTF8(vmThread, J9UTF8_DATA(nestHostName), J9UTF8_LENGTH(nestHostName), classLoader, classPreloadFlags);
```
tajila: If the nest host is not visible to the class being loaded does Oracle?

If so we will need to do a visibility check here before assigning to nestHostOut

runtime/vm/createramclass.cpp
```diff
@@ -2871,6 +2861,17 @@ internalCreateRAMClassFromROMClass(J9VMThread *vmThread, J9ClassLoader *classLoa
 		result->classFlags |= J9ClassIsAnonymous;
 	}
 
+
+#if defined(J9VM_OPT_VALHALLA_NESTMATES)
+	if (NULL != result) {
```
tajila: You can split the check above `if ((NULL != result) && (0 != (J9_FINDCLASS_FLAG_ANON & options))) {` and add this code there rather than NULL checking twice

runtime/vm/createramclass.cpp
tajila: Actually, looking at the code, there are some places where we have duplicated function headers doing the same here would be consistent.

taliamccormick: For what it's worth, my default eclipse auto-indent handles the first case correctly but mishandles the second case by adding an extra indent for all lines following. I'm leaning towards the first option again because of that convenience.

taliamccormick: Alternate option: Call a `loadNestHost` function in the same section of code as `loadSuperClassAndInterfaces`.
- Should still occur early enough to avoid the shared vscratch table problem.
- Does not involve changing function header
- Keeps the function name accurate (ie. `loadSuperClassAndInterfaces` does not indicate that the nest is being loaded)
- Does create some code duplication

taliamccormick:
Nestmates spec changed so that nest host resolution occurs during the access check, rather than during class loading. The nest host loading code will therefore need to move a second time.

I'm closing this PR to avoid having two subsequent nest host loading PRs.

PR #1060, opened 2018-01-30
Update test README.md

renfeiw:
- updated test tag part
- updated test target part

### DIFF
--- a/test/README.md
+++ b/test/README.md
@@ -49,8 +49,8 @@ make test
     ```
     JAVA_BIN=<path to JDK bin directory that you wish to test>
     SPEC=[linux_x86-64|linux_x86-64_cmprssptrs|...] platform_on_which_to_test
-    JAVA_VERSION=[SE80|SE90] (SE90 default value)
-    JAVA_IMPL=openj9 (openj9 default value)
+    JAVA_VERSION=[SE80|SE90|SE100|Panama|Valhalla] (SE90 default value)
+    JAVA_IMPL=[openj9|hotspot] (openj9 default value)
     ```
 
   * list of dependent jars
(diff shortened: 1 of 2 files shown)


renfeiw:
reviewers: @smlambert @llxia

test/README.md
```diff
@@ -94,6 +94,16 @@ make test
...
+    - A test can be tagged with elements to describe its attribute:
+      - level:   [sanity|extended] (extended default value)
+      - group:   [functional|openjdk|external|perf|jck|system] (required 
+                 to provide one group per test.)
+      - impl:    [openj9|hotspot] (filter test based on exported JAVA_IMPL 
+                 value; openj9 is the default value)
+      - subset:  [SE80|SE90] (filter test based on exported JAVA_VERSION 
+                 value; a test can be tagged with SE80 and SE90 at the 
+                 same time; default to all subsets)
+
```
llxia: I would say something like `a test can be tagged with multiple subsets`

test/README.md
```diff
@@ -94,6 +94,16 @@ make test
     examples with your own code, update the build.xml and playlist.xml
     files to match your new Test class names.
 
+    - A test can be tagged with elements to describe its attribute:
```
llxia: A test can be tagged with `following` elements:

llxia:
Overall looks good to me. :+1:

test/README.md
```diff
@@ -50,7 +50,7 @@ make test
     JAVA_BIN=<path to JDK bin directory that you wish to test>
     SPEC=[linux_x86-64|linux_x86-64_cmprssptrs|...] platform_on_which_to_test
     JAVA_VERSION=[SE80|SE90] (SE90 default value)
```
smlambert: We already have the need for SE100 (at Adopt), so I would add it in the README

test/README.md
```diff
@@ -92,7 +92,18 @@ make test
...
-    files to match your new Test class names.
+    files to match your new Test class names. The playlist.xml format 
+    is defined in TestConfig/playlist.xsd.
+
+    - A test can be tagged with following elements:
+      - level:   [sanity|extended] (extended default value)
+      - group:   [functional|openjdk|external|perf|jck|system] (required 
+                 to provide one group per test.)
+      - impl:    [openj9|hotspot] (filter test based on exported JAVA_IMPL 
+                 value; openj9 is the default value)
```
smlambert: I had thought that the lack of a tag for <impl> would mean that the test can be run against both openj9 or hotspot implementations.  Presumably, if openj9 is the default, it means that we have to tag all systemtests and openjdk regression tests at Adopt with both hotspot and openj9?

test/README.md
```diff
@@ -92,7 +92,18 @@ make test
...
+      - subset:  [SE80|SE90] (filter test based on exported JAVA_VERSION
```
smlambert: SE100 too

test/README.md
renfeiw: Yes, it's both. I made a mistake, updating the readme now.

renfeiw:
In addition to SE100, I also added Panama and Valhalla in readme.

renfeiw:
I added Panama and Valhalla, because the two projects also live in the open. I can remove them from readme, let me know.

smlambert:
We will leave them in the doc

PR #1061, opened 2018-01-30
Comment out the report function to avoid causing lock failure

fengxue-IS:
The reporter function is causing intermittent hangs during testing,
Removing this until hanging issue is fixed


gacholio:
Not bothering with sanity test as this code has no effect.  Travis (compile) build is sufficient.

fengxue-IS:
fixed, sorry about that

Issue #858, opened 2017-12-20
Get Test_AccessController:test_doPrivileged tests running at least for Java 8


pshipton:
The following tests are excluded for both Java 8 & 9 without any explanation. For starters, they need to be evaluated for Java 8 to determine why they are failing, and then fixed. Once Java 8 is fixed we can determine the proper action for Java 9.

- org.openj9.test.java.security.Test_AccessController:test_doPrivilegedWithCombiner4
- org.openj9.test.java.security.Test_AccessController:test_doPrivileged_createAccessControlContext

pdbain-ibm:
I'll take this.

pdbain-ibm:
Several issues:
1) Test creates some custom URLClassLoaders, which are unable to load framework classes such as the logger classes.  We fix this by checking the package name in the classloader and using the system classloader for framework classes.

2) Certain tests, particularly test_doPrivileged_createAccessControlContext() run fine when the test
is run standalone by adding a main() method but fail when run under TestNG.  I believe that this may be due to reflection by TestNG.

pdbain-ibm:
We get a NoClassDefFound on AppNoCreateACCPriv$1, a PrivilegedAction internal class.
In fact, loading any inner class in AppNoCreateACCPriv throws a NCDF exception.
Test also fails with Oracle JDK and Java 8: not a regression.

pdbain-ibm:
1) NCDF caused by a security exception when TestNG tries to write to the log file.  Updating the policy file.
2) Custom classloaders fail to load classes in the test framework: update them to handle those packages specially.

pdbain-ibm:
Pull request https://github.com/eclipse/openj9/pull/1076 opened.
Depends on other forthcoming changes before merging.

PR #1076, opened 2018-01-31
Fix Test_AccessController

pdbain-ibm:
Add special cases in custom classloaders, via a common superclass, to load test
harness classes.

Add a new security policy which allows the test harness to read and write the
results directory while the security manager is active.

Improve test progress reporting.

Delete obsolete comments, extract redundant strings to a constant.

Fixes issue #858.

### DIFF
--- a/test/Java8andUp/playlist.xml
+++ b/test/Java8andUp/playlist.xml
@@ -714,7 +714,6 @@
 	JCL_TEST_Java-Lang-Invoke,\
 	JCL_TEST_Java-Lang-Ref,\
 	JCL_TEST_Java-Lang-Reflect,\
-	JCL_TEST_Java-Security,\
 	JCL_TEST_Java-Math,\
 	JCL_TEST_Test-RunLast \
 	-groups $(TEST_GROUP) \
@@ -758,7 +757,6 @@
 	JCL_TEST_Java-Lang-Invoke,\
 	JCL_TEST_Java-Lang-Ref,\
 	JCL_TEST_Java-Lang-Reflect,\
-	JCL_TEST_Java-Security,\
 	JCL_TEST_Java-Math,\
 	JCL_TEST_Test-RunLast \
 	-groups $(TEST_GROUP) \
@@ -830,6 +828,78 @@
 		</groups>
 		<subsets>
 			<subset>SE80</subset>
+		</subsets>
+	</test>
+<!--
+	Java 8 specific test case.
+	Java 9 and above require VM options not available in Java 8.
+-->
+	<test>
+		<testCaseName>JCL_TEST_Java-Security_SE80</testCaseName>
+		<variations>
+			<variation>NoOptions</variation>
+		</variations>
+		<command>$(MKTREE) $(REPORTDIR); \
+	cd $(REPORTDIR); \
+	$(JAVA_COMMAND) $(JVM_OPTIONS) -verbose:stacktrace \
+	-Djava.security.policy=$(Q)$(TEST_RESROOT)$(D)testacc.policy$(Q) \
+	-Xbootclasspath/a:$(TEST_RESROOT)$(D)TestResources.jar \
+	-cp $(Q)$(RESOURCES_DIR)$(P)$(TESTNG)$(P)$(TEST_RESROOT)$(D)GeneralTest.jar$(P)$(TEST_RESROOT)$(D)TestResources.jar$(P)$(LIB_DIR)$(D)asm-all.jar$(Q) \
+	org.testng.TestNG -d $(REPORTDIR) $(Q)$(TEST_RESROOT)$(D)testng.xml$(Q) \
+	-testnames JCL_TEST_Java-Security \
+	-groups $(TEST_GROUP) \
+	-excludegroups $(DEFAULT_EXCLUDE); \
+	$(TEST_STATUS)</command>
+		<platformRequirements>^vm.hrt</platformRequirements>
+		<levels>
+			<level>sanity</level>
+		</levels>
+		<groups>
+			<group>functional</group>
+		</groups>
+		<subsets>
+			<subset>SE80</subset>
+		</subsets>
+	</test>
+
+<!--
+	Java 9 and above.
+-->
+	<test>
+		<testCaseName>JCL_TEST_Java-Security</testCaseName>
+		<variations>
+			<variation>NoOptions</variation>
+		</variations>
+		<command>$(MKTREE) $(REPORTDIR); \
+	cd $(REPORTDIR); \
+	$(JAVA_COMMAND) $(JVM_OPTIONS) -verbose:stacktrace \
+	--add-modules java.se.ee,openj9.sharedclasses \
+	--add-exports java.base/com.ibm.tools.attach.target=ALL-UNNAMED \
+	--add-exports java.base/com.ibm.oti.vm=ALL-UNNAMED \
+	--add-exports java.base/com.ibm.oti.util=ALL-UNNAMED \
+	--add-exports java.base/jdk.internal.vm.annotation=ALL-UNNAMED \
+	--add-exports java.base/jdk.internal.misc=ALL-UNNAMED \
+	--add-exports java.base/com.ibm.jit.crypto=ALL-UNNAMED \
+	--add-exports java.base/com.ibm.oti.reflect=ALL-UNNAMED \
+	--add-opens=java.base/java.lang=ALL-UNNAMED \
+	--add-opens=java.base/java.security=ALL-UNNAMED \
+	-Djava.security.policy=$(Q)$(TEST_RESROOT)$(D)testacc.policy$(Q) \
+	-Xbootclasspath/a:$(TEST_RESROOT)$(D)TestResources.jar \
+	-cp $(Q)$(RESOURCES_DIR)$(P)$(TESTNG)$(P)$(TEST_RESROOT)$(D)GeneralTest.jar$(P)$(TEST_RESROOT)$(D)TestResources.jar$(P)$(LIB_DIR)$(D)asm-all.jar$(Q) \
+	org.testng.TestNG -d $(REPORTDIR) $(Q)$(TEST_RESROOT)$(D)testng.xml$(Q) \
+	-testnames JCL_TEST_Java-Security \
+	-groups $(TEST_GROUP) \
+	-excludegroups $(DEFAULT_EXCLUDE); \
+	$(TEST_STATUS)</command>
+		<platformRequirements>^vm.hrt</platformRequirements>
+		<levels>
+			<level>sanity</level>
+		</levels>
+		<groups>
+			<group>functional</group>
+		</groups>
+		<subsets>
+			<subset>SE90</subset>
 		</subsets>
 	</test>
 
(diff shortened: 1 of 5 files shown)


pdbain-ibm:
Other changes expected for Test_AccessControlContext.

pdbain-ibm:
Copyright failure is in the Java security policy file.

test/Java8andUp/playlist.xml
```diff
@@ -922,6 +922,69 @@
 		</subsets>
 	</test>
 
+	<test>
+		<featureIds>
+			<featureId>Java access control</featureId>
+		</featureIds>
+		<testCaseName>JCL_TEST_Java-Security_SE90</testCaseName>
```
smlambert: The final make target that gets generated takes testCaseName + Subset + NumberOfVariation (count starting from 0).  If you call the test this, the target name will be JCL_TEST_Java-Security_SE90_SE90_0  is that what you intend?  Same comment applies to testCaseName J9VMInternals_Test_SE80 below... J9VMInternals_Test_SE80_SE80_0 and J9VMInternals_Test_SE80_SE80_1,

pdbain-ibm: Shelley

In re your comment:

> the target name will be JCL_TEST_Java-Security_SE90_SE90_0

Could you please clarify?

Embedding the Java version appears to be not unusual in the playlist, cf. analogous existing entry <testCaseName>JCL_Test_SE90</testCaseName> (from which I extracted the new test cases), et al.

Looking at a recent internal build, I see:

make -f /bluebird/builds/bld_377668/jvmtest/test/SE90/TestConfig/wrapper.mk _JCL_Test_SE90_0 TEST_ROOT=...

and the generated makefile SE90/Java8andUp/autoGen.mk contains:

JCL_Test_SE90_0: TEST_RESROOT=$(JVM_TEST_ROOT)$(D)Java8andUp
JCL_Test_SE90_0: JVM_OPTIONS=$(RESERVED_OPTIONS) $(EXTRA_OPTIONS)
JCL_Test_SE90_0: TEST_GROUP=level.sanity
JCL_Test_SE90_0:
    @$(MKTREE) $(REPORTDIR);
    @echo "" | tee -a $(Q)$(TESTOUTPUT)$(D)TestTargetResult$(Q);
    @echo "===============================================" | tee -a $(Q)$(TESTOUTPUT)$(D)TestTargetResult$(Q);
    @echo "Running test $@ ..." | tee -a $(Q)$(TESTOUTPUT)$(D)TestTargetResult$(Q);
...



All that said, this is still WIP (and likely has other problems).
I'm trying to merge the two cases into one, so this may be moot.

@llxia @TianyuZuo FYI.

llxia: Just to clarify, we do not need to add SE90 or SE80 in test name. However, the test name has to be unique. One case I have seen people use SE80 and SE90 is that if they have one test that uses different commands for SE80 and SE90, people use testXXX_SE80 and testXXX_SE90 as test name.  In your case, since the test will only run in SE90, there is no need to have SE90 in the test name.

pdbain-ibm: Correction: this test runs in both Java 8 and Java 9. As it happens, I believe I can use the same command line for both, so the point is now moot.  Personal build in progress.

pdbain-ibm: ... or not:
  java.lang.AssertionError: unexpected: java.lang.reflect.InaccessibleObjectException: Unable to make field java.security.ProtectionDomain[] java.security.AccessControlContext.context accessible: module java.base does not "opens java.security" to unnamed module @6579a4e

Java 8 passes fine.

Looks like I need separate test cases with "--add-opens" for Java 9 and above.

pdbain-ibm: @smlambert @llxia please review.  
As note above, copyright failure is in the Java security policy file.

 Thanks.

test/Java8andUp/playlist.xml
```diff
@@ -921,6 +919,84 @@
...
 		</subsets>
 	</test>
+<!--
+	Java 8 specific test case.
+	Java 9 and above require VM options not available in Java 8.
+-->
+	<test>
+		<featureIds>
+			<featureId>Java access control</featureId>
+		</featureIds>
```
llxia: Please remove <featureIds>.
We are in the process of removing <featureIds> in all playlist.

test/Java8andUp/src/org/openj9/test/java/security/Test_AccessControlContext.java
```diff
@@ -1,7 +1,7 @@
 package org.openj9.test.java.security;
 
 /*******************************************************************************
- * Copyright (c) 1998, 2017 IBM Corp. and others
+ * Copyright (c) 1998, 2018 IBM Corp. and others
```
llxia: Did we modify this file? Why we need to update the copyright?

test/Java8andUp/src/org/openj9/test/java/security/Test_AccessController.java
```diff
@@ -494,17 +493,6 @@ public PermissionCollection getPermissions(CodeSource cs) {
 				Object ret = AccessController.doPrivileged(new PrivilegedAction<Object>() {
 					public Object run() {
 						try {
-							/* This tests a use scenario with following stacktrace from a standalone testcase
```
pshipton: Why is this removed?

test/Java8andUp/src/org/openj9/test/java/security/Test_AccessController.java
```diff
@@ -517,28 +505,6 @@ public Object run() {
 				if (null == ret) {
 					Assert.fail("FAIL: TEST FAILED");
 				}
-				/*
```
pshipton: Why is this removed?

test/Java8andUp/src/org/openj9/test/java/security/Test_AccessController.java
pdbain-ibm: I took out the stack traces because they clutter the code and don't add much value.
They also may not be accurate still.

test/Java8andUp/src/org/openj9/test/java/security/Test_AccessController.java
pdbain-ibm: Ibid.

test/Java8andUp/src/org/openj9/test/java/security/Test_AccessController.java
pshipton: They add value when understanding what the tests are doing, and when having to debug a failure. Please leave them in.

pdbain-ibm: Done.

test/Java8andUp/src/org/openj9/test/java/security/Test_AccessController.java
```diff
@@ -353,17 +352,18 @@ public void test_doPrivilegedFrameStackWalking() {
...
 			String result = (String)java.security.AccessController.doPrivileged(pa);
 			Assert.fail("FAIL: security exception NOT thrown, get system property " + PROP_JAVA_HOME + " : " + result);
 		} catch (java.security.AccessControlException e) {
-			logger.debug("PASS: security exception expected.");
+			logger.debug(PASSED_SECURITY_EXCEPTION_EXPECTED);
 		}
 
-		// loop 20 time here such that the reflection object are optimized and bytecodes are created in a new class
-		// (after calling invoke() about 15 times), which introduces a slightly different reflection frame
+		/*
```
pshipton: Just FYI there is nothing wrong with using // comments in Java code

pdbain-ibm:
From the C/C++ coding standards (emphasis mine):

   Comments that span more than one line should use the Java style as
   follows:
   /* This is a long comment
       * that spans many lines
       * and this is how it
       * should look.
       */

(style quibble: should the "/*' not be on its own line?)

Also, from the "J9 Team Process" page:

   Java Coding Standards
   The coding standards for Java generally follow the C conventions with a
   few exceptions like braces to be more consistent with Java style.

---------------------------------------------------------------------
Peter Bain
Software Designer, IBM Runtime Technologies
IBM Ottawa Software Lab
3755 Riverside Drive, Ottawa, Ontario, K1V 1B8
613-356-5452
pdbain@ca.ibm.com

test/Java8andUp/src/org/openj9/test/java/security/Test_AccessControlContext.java
pdbain-ibm: That was a mistake.  Change reverted.

test/Java8andUp/playlist.xml
pdbain-ibm: Done.

pdbain-ibm:
Do the reviewers have any further comments?
Thank you

pshipton:
13:09:04 The following files were modified and have incorrect copyrights
[Pipeline] echo
13:09:04 test/Java8andUp/testacc.policy

test/Java8andUp/playlist.xml
```diff
@@ -830,6 +828,78 @@
...
+<!--
+	Java 8 specific test case.
+	Java 9 and above require VM options not available in Java 8.
+-->
+	<test>
+		<testCaseName>JCL_TEST_Java-Security_SE80</testCaseName>
+		<variations>
+			<variation>NoOptions</variation>
+		</variations>
+		<command>$(MKDIR) -p $(REPORTDIR); \
```
llxia: We can use $(MKTREE) instead of $(MKDIR). Same for the test below.

pdbain-ibm:
In re @pshipton 's comment: none of the other policy files have copyrights.  Is that deliberate?

pshipton:
An oversight, not adding copyrights to the other policy files. Somebody could volunteer to fix them all, or they can be fixed as needed in order to pass the copyright check.

test/Java8andUp/testacc.policy
```diff
@@ -0,0 +1,26 @@
+//  Copyright (c) 2016, 2018 IBM Corp. and others
```
pshipton: The start year should not be 2016 for a new file.

pdbain-ibm: Fixed.

pdbain-ibm:
@smlambert ready for your review.

test/Java8andUp/testacc.policy
```diff
@@ -0,0 +1,26 @@
+//  Copyright (c) 2018, 2018 IBM Corp. and others
```
smlambert: You are really taking Peter Shipton's comment about its okay to use // for comments seriously here for this 20 line block of commented code.  Out of curiosity, are policy files limited to using that style of comment or can /* */ also be used?

test/Java8andUp/playlist.xml
```diff
@@ -830,6 +828,78 @@
 			<subset>SE80</subset>
 		</subsets>
 	</test>
+<!--
```
smlambert: Thought we are currently limited to not support comments in the playlist files (blocked by and awaiting an infra issue to have XML::Parser installed on test machines).  Good if its not causing issues, though not sure the comments adds any information of value, as its pretty clear by the test name and the list of options that they are different.  I would likely get rid of the comments from the playlist.

smlambert:
Jenkins test sanity plinux

smlambert:
Please update the branch to resolve the conflict (latest_exclude_SE90.txt).

pdbain-ibm:
Done.

Issue #54, opened 2017-09-18
Investigate JVM signal handling functions


mgaudet:
The implementations of all 3 Signal JVM functions:

* `JVM_RegisterSignal`
* `JVM_FindSignal`
* `JVM_RaiseSignal`

Are questionable. All three need to be validated and completed.

DanHeidinga:
The implementations of all 3 Signal JVM functions:
* JVM_RegisterSignal
* JVM_FindSignal
* JVM_RaiseSignal

Are questionable.  All three need to be validated and completed.  My vague understanding is that J9 relies on patches to the JCL code to work around the need for these 3 functions.

mgaudet:
I'll re-title this as "Investigate JVM signal handling", as this came from me looking at https://github.com/eclipse/openj9/issues/40

mstoodle:
I marked as "enhancement" but if "bug" becomes more appropriate we should update it :)

pshipton:
@babsingh is this resolved by #283?

babsingh:
@pshipton no. more changes will be required to complete this. currently, the following scenario is not covered.

```
public class SigIntTest extends Thread {

        private static Object monitor = new Object();

        public static void main(String args[]) {
                Runtime.getRuntime().addShutdownHook(new SigIntTest());
                System.out.println("Press CTRL-C to exit - Sends SIGINT signal");

                synchronized (monitor) {
                        try {
                                monitor.wait();
                        } catch (InterruptedException e) {
                                System.out.println("Main thread interrupted");
                        }
                        System.out.println("Terminating main thread correctly");
                }
        }

        public void run() {
                System.out.println("OK ... exiting");
                synchronized (monitor) {
                        monitor.notify();
                }
        }

}
```

**Current behavior:**
```
Press CTRL-C to exit
^C
```

**Expected behavior:**
```
Press CTRL-C to exit
^COK... exiting
Terminating main thread correctly
```

DanHeidinga:
Current approach is https://github.com/eclipse/omr/issues/2332

pshipton:
The initial changes are delivered to enable signal handling on Windows, Linux, and AIX. There will be more internal changes but the functionality should be working as of now.

Linux/AIX eclipse/omr#2494 #1603 #1642
Windows #1816 not the full functionality but enough for CTRL-C to work

babsingh:
Adding @ibm-rtvs

pshipton:
[...]

PR #1079, opened 2018-01-31
Support JVM_*Signal functions 

babsingh:
- Fix/Update JVM_*Signal functions in jvm.c.
- Added code to initialize/destroy synchronization variables for VM
Signal Dispatch Thread.
- Added code to start and stop VM Signal Dispatch Thread.
- Added implementation of VM Signal Dispatch Thread.
- Added data structures to support VM Signal Dispatch Thread.
- Added function prototypes for VM Signal Dispatch Thread.

### DIFF
--- a/runtime/j9vm/jvm.c
+++ b/runtime/j9vm/jvm.c
@@ -324,6 +324,7 @@ static BOOLEAN librariesLoaded(void);
 #endif /* defined(J9ZTPF) */
 
 #define J9_SIG_ERR -1
+#define J9_SIG_IGNORED 1
 
 #define J9_PRE_DEFINED_HANDLER_CHECK 2
 #define J9_OLDHANDLER_SAME_AS_NEWHANDLER 2
@@ -333,9 +334,9 @@ static BOOLEAN librariesLoaded(void);
 /* Chosen based upon signal names listed in signalMap below. */
 #define J9_SIGNAME_BUFFER_LENGTH 16
 
-static void dummySignalHandler(jint sigNum);
-static BOOLEAN isSignalSpecial(jint sigNum);
+static void defaultSignalHandler(jint sigNum);
 static BOOLEAN isSignalUsedByVM(jint sigNum);
+static BOOLEAN isSignalIgnored(jint sigNum);
 
 typedef struct {
 	const char *signalName;
@@ -4408,33 +4409,6 @@ JVM_NativePath(char* path)
 }
 
 /**
- * Checks if a signal is special.
- *
- * @param sigNum Integer value of the signal
- *
- * @returns TRUE if the signal is special
- *          FALSE if the signal is not special
- */
-static BOOLEAN
-isSignalSpecial(jint sigNum)
-{
-	return
-#if defined(SIGHUP)
-		(SIGHUP == sigNum) ||
-#endif /* defined(SIGHUP) */
-#if defined(SIGINT)
-		(SIGINT == sigNum) ||
-#endif /* defined(SIGINT) */
-#if defined(SIGQUIT)
-		(SIGQUIT == sigNum) ||
-#endif /* defined(SIGQUIT) */
-#if defined(SIGTERM)
-		(SIGTERM == sigNum) ||
-#endif /* defined(SIGTERM) */
-		FALSE;
-}
-
-/**
  * Check if a signal is used by the VM.
  *
  * @param sigNum Integer value of the signal
@@ -4455,7 +4429,37 @@ isSignalUsedByVM(jint sigNum)
 #if defined(SIGSEGV)
 		(SIGSEGV == sigNum) ||
 #endif /* defined(SIGSEGV) */
+#if defined(SIGQUIT)
+		(SIGQUIT == sigNum) ||
+#endif /* defined(SIGQUIT) */
 		FALSE;
+}
+
+/**
+ * Check if a signal is ignored.
+ *
+ * @param sigNum Integer value of the signal
+ *
+ * @returns TRUE if the signal is ignored
+ *          FALSE if the signal is not ignored
+ */
+static BOOLEAN
+isSignalIgnored(jint sigNum) {
+	BOOLEAN signalIgnored = FALSE;
+
+#if !defined(WIN32)
+	struct sigaction oldSignalAction = {{0}};
+	OMRSIG_SIGACTION(sigNum, NULL, &oldSignalAction);
+	void *oldHandler = (void *)oldSignalAction.sa_sigaction;
+	if (NULL == oldHandler) {
+		oldHandler = (void *)oldSignalAction.sa_handler;
+	}
+	if (oldHandler == (void *)SIG_IGN) {
+		signalIgnored = TRUE;
+	}
+#endif /* !defined(WIN32) */
+
+	return signalIgnored;
 }
 
 /**
@@ -4474,11 +4478,18 @@ JVM_RaiseSignal(jint sigNum)
 {
 	jboolean rc = JNI_FALSE;
 	J9JavaVM *javaVM = (J9JavaVM *)BFUjavaVM;
+	BOOLEAN isShutdownSignal = javaVM->internalVMFunctions->isSignalUsedForShutdown(sigNum);
 
 	Trc_SC_RaiseSignal_Entry(sigNum);
 
-	if (J9_ARE_ALL_BITS_SET(javaVM->sigFlags, J9_SIG_XRS)
-			&& isSignalSpecial(sigNum)) {
+	if ((J9_ARE_ANY_BITS_SET(javaVM->sigFlags, J9_SIG_XRS | J9_SIG_XRS_ASYNC)
+			&& isShutdownSignal)
+#if defined(SIGQUIT)
+			|| (SIGQUIT == sigNum)
+#endif /* defined(SIGQUIT) */
+	) {
+		/* Ignore signal */
+	} else if (isShutdownSignal && isSignalIgnored(sigNum)) {
 		/* Ignore signal */
 	} else {
 		raise(sigNum);
@@ -4491,13 +4502,12 @@ JVM_RaiseSignal(jint sigNum)
 }
 
 /**
- * This is a stub for the pre-defined handler. The pre-defined
- * handler is supposed to be used in JVM_RegisterSignal when
- * the special value of J9_PRE_DEFINED_HANDLER_CHECK (2) is specified
- * in the handler. It hasn't been implemented since its functionality
- * is not known.
- *
- * TODO: Implement the pre-defined handler.
+ * The default handler is supposed to be used in JVM_RegisterSignal
+ * when the special value of J9_PRE_DEFINED_HANDLER_CHECK (2) is specified
+ * as the handler. The default handler is supposed to invoke the handler
+ * that is registered with the signal. The handler is invoked by a call to
+ * Signal.dispatch method. Signal -> jdk.internal.misc.Signal in Java 9 and
+ * onwards, and sun.misc.Signal in Java 8.
  *
  * @param sigNum Integer value of the signal to be sent to the
  *               calling process or thread
@@ -4505,8 +4515,17 @@ JVM_RaiseSignal(jint sigNum)
  * @returns void
  */
 static void
-dummySignalHandler(int sigNum) {
+defaultSignalHandler(jint sigNum) {
+	J9JavaVM *javaVM = (J9JavaVM *)BFUjavaVM;
+	javaVM->internalVMFunctions->recordAndNotifySignalReceived(javaVM, sigNum);
 
+#if defined(WIN32)
+	/* Handlers get reset on Windows. So, we have to register the handler
+	 * again once a signal is raised. If we use J9_PRE_DEFINED_HANDLER_CHECK,
+	 * then defaultSignalHandler will be set as the handler by JVM_RegisterSignal.
+	 */
+	JVM_RegisterSignal(sigNum, (void *)J9_PRE_DEFINED_HANDLER_CHECK);
+#endif /* defined(WIN32) */
 }
 
 /**
@@ -4517,7 +4536,7 @@ dummySignalHandler(int sigNum) {
  * and SIGTERM are also ignored; thus, we don't register a signal
  * handler for these signals. If handler has the special value of
  * J9_PRE_DEFINED_HANDLER_CHECK (2), then handler is changed to
- * dummySignalHandler before it is registered. If the old handler
+ * defaultSignalHandler before it is registered. If the old handler
  * is same as the new handler, then a special value,
  * J9_OLDHANDLER_SAME_AS_NEWHANDLER (2) is returned.
  *
@@ -4533,6 +4552,7 @@ JVM_RegisterSignal(jint sigNum, void* handler)
 {
 	J9JavaVM *javaVM = (J9JavaVM *)BFUjavaVM;
 	void *oldHandler = (void *)J9_SIG_ERR;
+	BOOLEAN isShutdownSignal = javaVM->internalVMFunctions->isSignalUsedForShutdown(sigNum);
 
 #if !defined(WIN32)
 	struct sigaction newSignalAction;
@@ -4548,30 +4568,35 @@ JVM_RegisterSignal(jint sigNum, void* handler)
 		/* Don't allow user to register a native handler since
 		 * the signal is already used by the VM.
 		 */
-	} else if (J9_ARE_NO_BITS_SET(javaVM->sigFlags, J9_SIG_XRS)
-			&& isSignalSpecial(sigNum)) {
+		return (void *)J9_SIG_ERR;
+	} else if (isShutdownSignal && J9_ARE_ANY_BITS_SET(javaVM->sigFlags, J9_SIG_XRS | J9_SIG_XRS_ASYNC)) {
 		/* Don't allow user to register a native handler since
 		 * the signal is already used by the VM.
 		 */
+		return (void *)J9_SIG_ERR;
+	} else if (isShutdownSignal && isSignalIgnored(sigNum)) {
+		return (void *)J9_SIG_IGNORED;
 	} else {
-		/* Register the signal */
+		/* Register the signal. */
 #if defined(WIN32)
 		if ((void *)J9_PRE_DEFINED_HANDLER_CHECK == handler) {
-			handler = (void *)dummySignalHandler;
+			handler = (void *)defaultSignalHandler;
 		}
 		oldHandler = OMRSIG_SIGNAL(sigNum, handler);
 #else /* defined(WIN32) */
-		sigemptyset(&newSignalAction.sa_mask);
-#if !defined(J9ZTPF)
-		newSignalAction.sa_flags = SA_RESTART;
-#else /* !defined(J9ZTPF) */
-		newSignalAction.sa_flags = 0;
-#endif /* !defined(J9ZTPF) */
+		/* Don't block any signals. */
+		if (0 != sigemptyset(&newSignalAction.sa_mask)) {
+			return (void *)J9_SIG_ERR;
+		}
+
+		newSignalAction.sa_flags = SA_RESTART | SA_NODEFER;
+
 		if ((void *)J9_PRE_DEFINED_HANDLER_CHECK == handler) {
-			newSignalAction.sa_handler = dummySignalHandler;
+			newSignalAction.sa_handler = (void (*)(int))defaultSignalHandler;
 		} else {
 			newSignalAction.sa_handler = (void (*)(int))handler;
 		}
+
 		OMRSIG_SIGACTION(sigNum, &newSignalAction, &oldSignalAction);
 #endif /* defined(WIN32) */
 	}
--- a/runtime/oti/j9nonbuilder.h
+++ b/runtime/oti/j9nonbuilder.h
@@ -4785,6 +4785,8 @@ typedef struct J9InternalVMFunctions {
 #if defined(J9VM_RAS_EYECATCHERS)
 	void (*rasSetServiceLevel)(struct J9JavaVM *vm, const char *runtimeVersion);
 #endif /* J9VM_RAS_EYECATCHERS */
+	BOOLEAN (*recordAndNotifySignalReceived)(struct J9JavaVM * vm, UDATA sigNum);
+	BOOLEAN (*isSignalUsedForShutdown)(UDATA sigNum);
 } J9InternalVMFunctions;
 
 
@@ -5071,6 +5073,18 @@ typedef struct J9VMRuntimeStateListener {
 #define J9VM_RUNTIME_STATE_LISTENER_STOP 2
 #define J9VM_RUNTIME_STATE_LISTENER_ABORT 3
 #define J9VM_RUNTIME_STATE_LISTENER_TERMINATED 4
+
+/* @ddr_namespace: map_to_type=J9VMSignalDispatchInfo */
+
+struct J9JavaVM; /* Forward struct declaration */
+struct J9VMThread; /* Forward struct declaration */
+typedef struct J9VMSignalDispatchInfo {
+	U_32 signalDispatchState;
+	struct J9VMThread *signalDispatchVMThread;
+	omrthread_monitor_t monitor;
+	UDATA *unhandledSignals;
+	j9sem_t osSemaphore;
+} J9VMSignalDispatchInfo;
 
 /* @ddr_namespace: map_to_type=J9JavaVM */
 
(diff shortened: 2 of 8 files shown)


babsingh:
fyi: @pshipton @DanHeidinga @andrew-m-leonard

pshipton:
@babsingh Does this fully resolve #54?

runtime/j9vm/jvm.c
```diff
@@ -4412,7 +4414,7 @@ JVM_NativePath(char* path)
 }
 
 /**
- * Checks if a signal is special.
+ * Checks if a signal is special (used for shutdown).
```
pshipton: Please rename the function to better correspond to what is does.

runtime/j9vm/jvm.c
```diff
@@ -4478,11 +4507,18 @@ JVM_RaiseSignal(jint sigNum)
...
+	BOOLEAN isSpecialSignal = isSignalSpecial(sigNum);
 
 	Trc_SC_RaiseSignal_Entry(sigNum);
 
-	if (J9_ARE_ALL_BITS_SET(javaVM->sigFlags, J9_SIG_XRS)
-			&& isSignalSpecial(sigNum)) {
+	if (J9_ARE_ANY_BITS_SET(javaVM->sigFlags, J9_SIG_XRS | J9_SIG_XRS_ASYNC)
+			&& (isSpecialSignal
+#if defined(SIGQUIT)
+			|| (SIGQUIT == sigNum)
```
pshipton: Should this check isSignalUsedByVM()?

runtime/j9vm/jvm.c
```diff
@@ -4549,27 +4590,30 @@ JVM_RegisterSignal(jint sigNum, void* handler)
...
-			handler = (void *)dummySignalHandler;
+			handler = (void *)defaultSignalHandler;
 		}
 		oldHandler = OMRSIG_SIGNAL(sigNum, handler);
 #else /* defined(WIN32) */
-		sigemptyset(&newSignalAction.sa_mask);
+		sigfillset(&newSignalAction.sa_mask);
 #if !defined(J9ZTPF)
-		newSignalAction.sa_flags = SA_RESTART;
+		newSignalAction.sa_flags = SA_RESTART | SA_SIGINFO;
```
pshipton: Not sure why SA_SIGINFO is being set. This means to use sa_sigaction (which takes 2 args) which this code is not doing.

runtime/j9vm/jvm.c
```diff
@@ -4495,22 +4531,26 @@ JVM_RaiseSignal(jint sigNum)
...
- * is not known.
+ * The default handler is supposed to be used in JVM_RegisterSignal
+ * when the special value of J9_PRE_DEFINED_HANDLER_CHECK (2) is specified
+ * as the handler. The default handler is supposed to invoke the handler
+ * that is registered with the signal. The handler is invoked by a call to
+ * Signal.dispatch method. Signal -> jdk.internal.misc.Signal in Java 9 and
+ * onwards, and sun.misc.Signal in Java 8.
  *
- * TODO: Implement the pre-defined handler.
+ * TODO: Is an exclusive signal dispatch thread required?
```
pshipton: Right, this is going to need another thread in order to call into Java.

runtime/j9vm/jvm.c
```diff
@@ -4549,27 +4590,30 @@ JVM_RegisterSignal(jint sigNum, void* handler)
...
+		newSignalAction.sa_flags = SA_RESTART | SA_SIGINFO;
 #else /* !defined(J9ZTPF) */
 		newSignalAction.sa_flags = 0;
 #endif /* !defined(J9ZTPF) */
 		if ((void *)J9_PRE_DEFINED_HANDLER_CHECK == handler) {
-			newSignalAction.sa_handler = dummySignalHandler;
+			newSignalAction.sa_handler = (void *)defaultSignalHandler;
 		} else {
 			newSignalAction.sa_handler = (void (*)(int))handler;
 		}
```
pshipton: I wonder if omrsig_set_async_signal_handler() can be used instead of OMRSIG_SIGACTION(), as this will take care of creating another thread.

runtime/j9vm/jvm.c
babsingh: SIGHUP, SIGINT and SIGTERM are used for shutdown. so, renamed function to `isSignalUsedForShutdown`.

runtime/j9vm/jvm.c
babsingh: `omrsig_set_async_signal_handler` doesn't work with all the signals. it only works with these signals: `QUIT`, `ABRT`, `TERM`, `RECONFIG` and `XFSZ`.

pshipton: It could be fixed to handle all the shutdown signals.

runtime/j9vm/jvm.c
babsingh: I was trying to determine the functionality of `defaultSignalHandler`. So, I tried existing handlers such as `omrsig_handler` and `structuredSignalHandlerVM`, which had three input arguments. `JVM_RegisterSignal` would register a `handler` for a specific signal. When `JVM_RaiseSignal` is called for a signal, then the `defaultSignalHandler` is invoked. I found that the job of `defaultSignalHandler` is to invoke the registered handler for the signal. `defaultSignalHandler` invokes the registered handler for the signal by calling `Signal.dispatch`. I will remove `SA_SIGINFO` since `defaultSignalHandler` only takes one input argument. but, things still work with `SA_SIGINFO`.

runtime/j9vm/jvm.c
babsingh: According to https://github.com/ibmruntimes/openj9-openjdk-jdk9/blob/openj9/jdk/test/sun/misc/SunMiscSignalTest.java, 

`SIGHUP`, `SIGINT`, `SIGTERM` can't be registered, invoked or raised if `-Xrs` is present.
`SIGQUIT` can't be registered, invoked or raised at all times. 

Updated the definition of JVM_RaiseSignal to reflect this behavior.

babsingh:
@pshipton Does this fully resolve #54? No.

I am using https://github.com/ibmruntimes/openj9-openjdk-jdk9/blob/openj9/jdk/test/sun/misc/SunMiscSignalTest.java to verify the functionality of `JVM_*Signal` functions.

58 tests are run and 2 failures are seen with the new changes. Without the new changes, 25+ failures are seen in OpenJ9. 27 failures are seen in IBM J9.

`isDaemonThread` testcase is failing since the thread that runs the `handler` is not a daemon thread.

`testSunMisc("INT", YES, YES, YES, YES)` testcase is failing with a NPE. The registered handler ends up as NULL after `testRaiseNoConsumer` testcase is run. If we don't run `testRaiseNoConsumer` testcase, `testSunMisc("INT", YES, YES, YES, YES)` testcase passes.

`Ctrl + C` [`SIGINT`] testcase doesn't work properly. `java.lang.Terminator.setup` registers a handler for shutdown signals: `SIGINT`, `SIGHUP` and `SIGTERM` using `sun.misc.Signal.handle` (`JVM_RegisterSignal`). The handler calls `java.lang.Shutdown.exit`. When `Ctrl + C` is issued, the testcase executes the handler which calls `Shutdown.exit`. While executing the handler, core dump is generated. Neither the shutdown hooks are executed nor the JVM exits cleanly. Is this happening because we are not using another thread to call the handler? Is this happening because `Shutdown.exit` is called randomly? Do we need to do something before invoke the handler for `Shutdown.exit`?

pshipton:
> Is this happening because we are not using another thread to call the handler?

I expect so. Another thread is required, and it needs to be attached (i.e. see shutDownHookWrapper()).

runtime/j9vm/jvm.c
babsingh: `JVM_RegisterSignal` needs to return the old handler while registering a new handler. `omrsig_set_async_signal_handler` will register a new handler to a signal but will not unregister the old handler. So, we would need to store the current handler for all signals, and unregister the current handler when a new handler is registered. There will also be an issue if a single handler is registered with multiple signals. Multiple calls to `omrsig_set_async_signal_handler` for the same handler will override the previous setting.

andrew-m-leonard:
@babsingh great detective work :-)

DanHeidinga:
> I expect so. Another thread is required, and it needs to be attached (i.e. see shutDownHookWrapper()).

Can we find a way to repurpose an existing helper thread?  Every new thread adds to the JVM's footprint...

babsingh:
Can we find a way to repurpose an existing helper thread? Every new thread adds to the JVM's footprint...

By `repurpose`, we want an existing thread to handle its current responsibilities + take on signal dispatch responsibilities. It's possible. `VM Runtime State Listener` (vmruntimestate.c) may be a good candidate. This optimization should be done after a standalone thread implementation is delivered. If this optimization doesn't work, we will have a working (standalone thread) solution. If the optimization works, we should keep the standalone thread solution in a disabled state if throughput issues arise in the future.

babsingh:
Currently, I am stuck with an issue related to `Ctrl + C`. `omrthread_monitor_t` is used to enforce synchronization in the signal handler code. Issuing `Ctrl + C` results in a seg fault in `omthread_monitor_enter` code since `self` ends being NULL ... `omrthread_t self = MACRO_SELF()`. If I execute `kill -2 PID` (send a signal to the process group) == `Ctrl + C` (SIGINT == 2), seg fault happens again. If I execute `kill -2 TID` (send a signal to a specific thread) or execute `sun.misc.Signal.raise(SIGINT)` (raise the signal in the caller), then the signal handler executes properly: runs shutdown hooks and VM exits properly. I thought `Ctrl + C` may be flooding the process with SIGINT signals by raising a SIGINT signal in each thread. This doesn't seem to be the case since the seg fault happens in the first invocation of the signal handler after `Ctrl + C` is issued.

Does anyone know why `Ctrl + C` or `kill -2 PID` causes `omrthread_t self = MACRO_SELF()` to be NULL? Does OMR run some cleanup code? @charliegracie @DanHeidinga @gacholio @pshipton

babsingh:
@DanHeidinga @pshipton 
>Can we find a way to repurpose an existing helper thread? Every new thread adds to the JVM's footprint...

another possibility: In OMR code, we have a `Signal Reporter` thread (`asynchSignalReporterThread` in `omrsignal.c`, which supports few signals: `QUIT`, `ABRT`, `TERM` and `XFSZ`. we can extend it to support all the signals that the JVM needs to support. Since this thread is in OMR, we won't be able to attach it to a `J9VMThread`; this will prevent us to call Java code. We will have to investigate if `Signal Reporter` thread can be exposed to the JVM, and if we can attach it to a J9VMThread for calling Java code.

babsingh:
Jenkins test sanity depends eclipse/omr#2289

babsingh:
> Currently, I am stuck with an issue related to Ctrl + C ...

The above issue has been resolved.

DanHeidinga:
> The above issue has been resolved.

Can you provide some context on how it was resolved?  If there's ever further problems in this area having something we can look back at may make it easier to debug.  If we don't put the info in the PR / code / etc, then future us (or others) will miss out on it.

As I mentioned in an other PR today, let's leave breadcrumbs along the way to make it easier to retrace the path if necessary.

babsingh:
> Currently, I am stuck with an issue related to Ctrl + C ...

The above issue has been resolved. `Ctrl + C` / `SIGINT` signal was received by the `JLI_Launch` thread. `JLI_Launch` thread has no awareness of the OMR environment. When the `defaultSignalHandler` is invoked from the `JLI_Launch` thread, then `omrthread_monitor_enter/exit` functions started to segfault since `MACRO_SELF()` returns `NULL`. I rewrote the code to use a mutex and a condition variable directly without relying upon `monitor` code. This resolved this issue.

babsingh:
Notes:
* I tried using `OPENJ9_BUILD` flag to enable this feature only in OpenJ9 builds. But, `OPENJ9_BUILD` flag isn't set consistently across all OpenJ9 source code. This is causing compilation issues while using `OPENJ9_BUILD` flag.
* DDR is causing compilation issues in internal builds. `NSIG` macro is defined in `<signal.h>`. For `volatile UDATA unhandledSignals[NSIG + 1]` in `j9nonbuilder.h`, DDR can't find a definition of `NSIG` while creating `source.zip`. A solution would be to dynamically allocate memory ... `UDATA *unhandledSignals` ... `omrmem_allocate_memory(sizeof(UDATA)*(NSIG + 1))` ... during run-time.

JasonFengJ9:
FYI: I am working on a PR to move `OPENJ9_BUILD` to the makefile top level.

babsingh:
I am trying to expose `J9OSMutex` and `J9OSCond` to OpenJ9. Currently, I accomplish this by including `thrdsup.h` directly in OpenJ9. This works fine on Linux. But, it causes the below mentioned compilation issues on Windows and AIX. The source files mentioned below are not changed.

We would need to neatly expose `J9OSMutex` and `J9OSCond` to OpenJ9. I have started related work in https://github.com/eclipse/omr/pull/2301.

AIX:
```
"FileStream.cpp", line 45.1: 1540-2963 (S) The member "void open64(const char *)" does not match any member declared in its containing class.
gmake[2]: *** [FileStream.o] Error 1
gmake[2]: *** Waiting for unfinished jobs....
"TextFileStream.cpp", line 51.1: 1540-2963 (S) The member "void open64(const char *, bool)" does not match any member declared in its containing class.
gmake[2]: *** [TextFileStream.o] Error 1
```

Windows: 
```
c:\temp\bld_378349\bld_win_x86-64\ddr\bcutil\ClassFileOracle.hpp(780) : error C2332: 'struct' : missing tag name
c:\temp\bld_378349\bld_win_x86-64\ddr\bcutil\ClassFileOracle.hpp(780) : error C2011: 'ClassFileOracle::<unnamed-tag>' : 'enum' type redefinition
c:\temp\bld_378349\bld_win_x86-64\ddr\bcutil\ClassFileOracle.hpp(941) : see declaration of 'ClassFileOracle::<unnamed-tag>'
c:\temp\bld_378349\bld_win_x86-64\ddr\bcutil\ClassFileOracle.hpp(780) : error C2226: syntax error : unexpected type 'ClassFileOracle::<unnamed-tag>'
c:\temp\bld_378349\bld_win_x86-64\ddr\bcutil\ClassFileOracle.hpp(780) : error C2332: 'struct' : missing tag name
c:\temp\bld_378349\bld_win_x86-64\ddr\bcutil\ClassFileOracle.hpp(780) : error C2011: 'ClassFileOracle::<unnamed-tag>' : 'enum' type redefinition
c:\temp\bld_378349\bld_win_x86-64\ddr\bcutil\ClassFileOracle.hpp(941) : see declaration of 'ClassFileOracle::<unnamed-tag>'
c:\temp\bld_378349\bld_win_x86-64\ddr\bcutil\ClassFileOracle.hpp(780) : error C2143: syntax error : missing ';' before '!='
c:\temp\bld_378349\bld_win_x86-64\ddr\bcutil\ClassFileOracle.hpp(780) : error C2143: syntax error : missing ';' before '!='
c:\temp\bld_378349\bld_win_x86-64\ddr\bcutil\ClassFileOracle.hpp(780) : error C2332: 'struct' : missing tag name
```

pshipton:
I'm guessing you are including thrdsup.h in some other include which is used by many files? You could include it just in the file where you need it. The compiler code includes it in a few files.

babsingh:
Jenkins test sanity depends https://github.com/eclipse/omr/pull/2301

babsingh:
Status:
1) Squashed everything into a single commit
2) Commented `OPENJ9_BUILD` ifdefs until `OPENJ9_BUILD` flag is globally defined.
3) Planning to replace `omrthread_monitor_t` with `j9sem_t` to notify the "Signal Dispatch Thread" in the signal handler since blocking operations are not allowed in the signal handler, and `sem_post` operation is considered async-signal-safe (http://man7.org/linux/man-pages/man7/signal-safety.7.html).
4) `runStaticMethod` is used to call `sun.misc.Signal.dispatch(final int number)` from native code. On AIX, `runStaticMethod` doesn't propagate the input argument to the `dispatch` method. `number` (input argument) is always `0` on AIX which results in a NPE. On Windows and Linux, `runStaticMethod` functions properly.

babsingh:
In OMR `j9sem.c`, it says `@deprecated Semaphores are no longer supported`. There are references to `j9sem_t` in OpenJ9 but they are not used. @pshipton @DanHeidinga should I use the OMR semaphore implementation?

Notes:
- On zOS, semaphore implementation uses a pthread mutex and condvar. On zOS, we can get System V implementation of a semaphore but it is not signal safe.
- On OSX, `dispatch_semaphore_t` is used. This is a Grand Central Dispatch semaphore, which is not signal safe. Posix semaphores are built over Mach semaphores. Mach semaphores will be a better option to achieve signal safety on OSX. Reference - https://github.com/adrienverge/openfortivpn/issues/105.

pshipton:
Discussed with Babneet to design a new OMR API, and request early approval of the same, so the main signal handler can be in OMR and use the same mechanisms as the current omrsignal.c implementation.

babsingh:
@pshipton @DanHeidinga thoughts?

**I studied the signal API in OMR**: `omrsig_set_async_signal_handler`. In OMR, they maintain a list of `J9*AsyncHandlerRecord` a.k.a `asyncHandlerList`.

Whenever `omrsig_set_async_signal_handler` is invoked, we look for a record associated to the specified handler (`omrsig_handler_fn`) in `asyncHandlerList`.

If a `J9*AsyncHandlerRecord` exists in `asyncHandlerList`, we update `flags` to associate the record with a signal. If no record exists, then we add a new record for the handler in `asyncHandlerList` and set `flags` appropriately. On Unix, we register handlers with the OS using `sigaction`. On Windows, we don't register signals with the OS using `signal`; it is assumed signals will only be sent using `Ctrl + ...` so `SetConsoleCtrlHandler` is used to set an handler (`consoleCtrlHandler`).

When an async signal is received, `masterASynchSignalHandler` is invoked by the signal on Linux/Unix platfroms. This function increments signal counts and wakes up the `asynchSignalReporter` thread. `asynchSignalReporter` thread decrements the signal count, finds all records in `asyncHandlerList` for the raised signal and invokes the `handler` in each record.

On Windows, handlers are not registered with the OS, and there is no thread to invoke handlers. When a signal is raised, the OS invokes `consoleCtrlHandler` which was set using `SetConsoleCtrlHandler`. `consoleCtrlHandler` finds all records in asyncHandlerList for the raised signal and invokes the handler in each record.

**What needs to be done in OMR?**

1) Currently, only four signals are supported on Unix/Linux platforms: `SIGXFSZ`, `SIGQUIT`, `SIGABRT` and `SIGTERM`. We would need to extend support to all Posix signals. AIX and zOS may have some system-specific signals, which would require support.

2) On Windows, we need to support the following three signals: SIGINT, SIGTERM and SIGABRT. Currently, we don't register handlers with the OS using the `signal` system call so system calls such as `raise` and `sun.misc.Signal.raise` won't work. We would have add a new thread to process raised signals; similar to Linux. Currently, only `Ctrl + BREAK | C | CLOSE` events will be processed on Windows. Also, we would need a mechanism to keep track of signal counts.

3) Each bit in `J9*AsyncHandlerRecord->flags` represents a signal. 32 bits can only represent 32 signals. On some platforms, there are more than 32 signals. If we use `uint64_t` to represent flags, then we will be able to represent 64 signals. There will be an issue if there are more than 64 signals on a system. `NSIG` macro is available in `<signal.h>` on all platforms except zOS. `NSIG` tells us how many signals are there on a system. This can be used to create an array version of flags.

4) We would need a new API:
   a) Function to unset a signal flag from all `J9*AsyncHandlerRecord->flags`. By invoking this function before calling `omrsig_set_async_signal_handler`, we can assure that one handler is associated to a signal.
   b) `omrsig_set_async_signal_handler` will call sigaction or signal to register a `{handler, signal}` tuple with the OS. `sigaction` and `signal` will return the old signal handler. We would need a variant of `omrsig_set_async_signal_handler` which returns the value returned by `sigaction` and `signal`.
   c) We would also need a variant of `omrsig_set_async_signal_handler` which would register `masterASynchSignalHandler` as a handler for a signal.

5) There are four implementations of `omrsig_set_async_signal_handler`: `unix`, `win32`, `win64amd` and `ztpf`. For any new API, four variants would need to be added to support each platform specification.

6) OMR signal API is part of the port library so it can't access internal macros in the thread library. For implementing the Windows thread in `omrsignal.c`, we won't have access to the synchronization macros in the thread library for using a semaphore, mutex etc. So, we would have to write those macros from scratch.

7) ... More changes will be required as we implement the above barebone specification ...

**Should we do this?**

The OMR signal API is inconsistently implemented between Windows and Unix. Extending the OMR signal API to add signal support in the JVM will take a lot of time. I can't guarantee if things will work at the end if we take the OMR approach. It will be easier and time efficient to just add a cross platform sempahore implementation in OpenJ9 and use the current signal changes in the JVM.

```
-- omrsig_set_async_signal_handler(...,  omrsig_handler_fn handler, void *handler_arg, uint32_t flags)
---- registerMasterHandlers
------ registerSignalHandlerWithOS //call to sigaction
```

```
struct {
	uint32_t portLibSignalNo;
	int	unixSignalNo;
} signalMap[] =	{
	{OMRPORT_SIG_FLAG_SIGSEGV, SIGSEGV},
	{OMRPORT_SIG_FLAG_SIGILL,  SIGILL},
	{OMRPORT_SIG_FLAG_SIGFPE,  SIGFPE},
	{OMRPORT_SIG_FLAG_SIGQUIT, SIGIQUIT},
	{OMRPORT_SIG_FLAG_SIGABRT, SIGABRT},
	{OMRPORT_SIG_FLAG_SIGTERM, SIGTERM}
};
```

```
/* Store the previous signal handlers, we need to restore them when we're done */
static struct {
	struct sigaction action;
	uint32_t restore;
} oldActions[MAX_UNIX_SIGNAL_TYPES];
```

```
typedef struct J9*AsyncHandlerRecord {
	OMRPortLibrary *portLib;
	omrsig_handler_fn handler;
	void *handler_arg;
	uint32_t flags;
	struct J9*AsyncHandlerRecord *next;
} J9*AsyncHandlerRecord;
```

pshipton:
4a seems like it should be an internal OMR detail of calling 4b or 4c. Does it need to be an explicit API?

@charliegracie fyi

pshipton:
It will benefit OMR to have a better signal handling story. ATM its specific to obsolete IBM Java requirements I believe. If OpenJ9 can't use the existing OMR signal implementation, it may not be much use to any other project.

There is also a benefit to having a single implementation to maintain (i.e. platform semaphore support), instead of multiple copies of things.

babsingh:
True, 4a can be an internal detail. Also, 4a, 4b and 4c can be a single function with some control mechanism. For further discussion, I have created an OMR issue: https://github.com/eclipse/omr/issues/2332.

babsingh:
@gacholio I am trying to execute the `jdk.internal.misc.Signal.dispatch` using `callin.cpp::runStaticMethod`. The method returns `void` and takes one input argument.

```
public final class Signal {
    ...
    private static void dispatch(final int number) {
```

On Windows and Linux, the input argument gets passed on properly while running `Signal.dispatch` using `runStaticMethod`. On AIX, the input argument is not passed properly to the Java method; it is always 0 inside `Signal.dispatch`.

The code to execute `jdk.internal.misc.Signal.dispatch` using `callin.cpp::runStaticMethod` is here (`signalDispatch` method): https://github.com/eclipse/openj9/pull/1079/files#diff-abe0cd1587b6a77eb44fb6920607a2c6R50.

I have verified that the input argument is set properly in `currentThread->sp = arguments[i]` (inside `callin.cpp`). After this, a call to `c_cInterpreter` is done. I couldn't debug/trace the call to `c_cInterpreter`. Can you help with this issue?

babsingh:
**Update:** I have resolved all issues. Consistent behavior is seen on AIX, Windows and Linux.
Currently, mutexes and condvars are used. Special measures have been taken to avoid hangs during usage of mutexes and condvar. 

I haven't received any response or feedback for the OMR proposal. In my opinion, the OMR extension/approach is inefficient so I am against implementing it. Extending OMR signal API doesn't guarantee usage in other projects.

OMR project no longer supports a cross platform semaphore API. Neither will OMR ever support a cross platform semaphore API because we can use such an API without attaching to the thread library. So, we should add a cross-platform semaphore API into OpenJ9, and replace the usage of mutex/condvar in the signal handler.

babsingh:
@pshipton @DanHeidinga Update:
1. Changed the implementation to use `j9sem_t (sem_post)` inside the signal handler (`recordAndNotifySignalReceived`). `j9sem_t` provides access to `sem_post` equivalent on Windows, Linux and Unix. On zOS, `sem_post` isn't available so `j9sem_t` mimics a semaphore using `omrthread_monitor_t`.

pshipton:
Its my understanding we are pursuing https://github.com/eclipse/omr/issues/2332 rather than this solution.

PR #1090, opened 2018-02-01
Disable testDefaultDisclaimMemory on zlinux31

llxia:
[skip ci]


fjeremic:
It would be good for the future to describe why we are disabling such tests. Was there an issue with the test? Does it not run on the platform? Is it stale? Some more information would be nice in the commit itself for other readers (like myself) to understand the context of the change and why it was necessary.

smlambert:
Extracted from details in internal RTC item 136837: 
According documentation running Java on zLinux31 with heap larger then -Xmx768m is unreliable.

Issue #995, opened 2018-01-22
org.openj9.test.java.lang.Test_ClassLoader tests are excluded for Java 8


pshipton:
The org.openj9.test.java.lang.Test_ClassLoader tests (openj9/test/Java8andUp/src_80/org/openj9/test/java/lang/Test_ClassLoader.java) are excluded for Java 8. The tests need to be fixed if necessary to work property on Java 8, and included.

pdbain-ibm:
Per Dan's request, I'll look at this.

pdbain-ibm:
The tests are already running in Java 8 and 9.  Both are passing:
PASSED: test_classloaders
PASSED: test_latestUserDefinedLoader

===============================================
    JCL_TEST_Test-Annotation-ClassLoader
    Tests run: 2, Failures: 0, Skips: 0
===============================================

There is an exclusion in TestConfig/resources/excludes/latest_exclude_SE90.txt, but it doesn't seem to be excluding anything.

pdbain-ibm:
Further to above comment: there are actually TWO "Test_ClassLoader" test classes: one in the org.openj9.test.java.lang package and the second in org.openj9.test.annotationClassLoader.
The latter is running and passing.  The former exists in the OpenJ9 repository but the testng and playlist files are in an internal repo.

Since the test implementation is already open source, I am moving the playlist and testng entries to 
the open source repo.

The test passes when  run manually but fails in personal builds:
```
FAILED: test_getResource
java.lang.AssertionError: Unable to find resource
	at org.testng.AssertJUnit.fail(AssertJUnit.java:59) from sun.misc.Launcher$AppClassLoader@f4341f65(file:/tmp/bld_378145/openj9/jvmtest/TestConfig/lib/testng.jar)
	at org.testng.AssertJUnit.assertTrue(AssertJUnit.java:24) from sun.misc.Launcher$AppClassLoader@f4341f65(file:/tmp/bld_378145/openj9/jvmtest/TestConfig/lib/testng.jar)
	at org.openj9.test.java.lang.Test_ClassLoader.test_getResource(Test_ClassLoader.java:72) from sun.misc.Launcher$AppClassLoader@f4341f65(file:/tmp/bld_378145/openj9/jvmtest/IBM_Internal/Java8andUp_IBM_Internal/GeneralTest.jar)
```

 I'm investigating.

pdbain-ibm:
The IBM_Internal copy of the test fails, but the open-source version works fine.
[...]

PR #1092, opened 2018-02-01
Remove exclusion for org.openj9.test.java.lang.Test_ClassLoader

pdbain-ibm:
This exclusion has no effect.  The purportedly excluded test is in fact running and passing.

Fixes https://github.com/eclipse/openj9/issues/995

### DIFF
--- a/test/Java8andUp/playlist.xml
+++ b/test/Java8andUp/playlist.xml
@@ -705,6 +705,7 @@
 	JCL_TEST_Test-Annotation,\
 	JCL_TEST_Test-Annotation-Identity,\
 	JCL_TEST_Test-Annotation-ClassLoader,\
+	JCL_TEST_Java-Lang_ClassLoader,\
 	JCL_TEST_Test-Annotation-Package,\
 	JCL_TEST_Test-Defects,\
 	JCL_TEST_Test-UnsafeFence,\
@@ -748,6 +749,7 @@
 	JCL_TEST_Test-Annotation,\
 	JCL_TEST_Test-Annotation-Identity,\
 	JCL_TEST_Test-Annotation-ClassLoader,\
+	JCL_TEST_Java-Lang_ClassLoader,\
 	JCL_TEST_Test-Annotation-Package,\
 	JCL_TEST_Test-Defects,\
 	JCL_TEST_Test-UnsafeFence,\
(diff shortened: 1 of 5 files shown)


smlambert:
@llxia - prior to merging this change which apparently does nothing, let's figure out why.  

@pdbain-ibm - it would be more useful for you to provide a link to a real build, than a copy paste of the output of some unknown build (as you did in issue #995).

smlambert:
I also see that there are 2 Test_ClassLoader classes:
https://github.com/eclipse/openj9/blob/master/test/Java8andUp/src_80/org/openj9/test/java/lang/Test_ClassLoader.java

https://github.com/eclipse/openj9/blob/master/test/Java8andUp/src/org/openj9/test/annotationClassLoader/Test_ClassLoader.java

I imagine the one that is running in builds at present in the test suite called JCL_TEST_Test-Annotation-ClassLoader, is the latter, which is not the one that is currently excluded.

smlambert:
There appears to be no test suite that contains the former Test_ClassLoader, at least where one would expect to find it:
https://github.com/eclipse/openj9/blob/master/test/Java8andUp/testng.xml

smlambert:
Jenkins test sanity plinux

pdbain-ibm:
This test refers to a testcase (playlist and testng.xml) from an internal IBM repo.  

There is a copy of the test program org.openj9.test.java.lang.Test_ClassLoader in both the 
internal repo and OpenJ9.  They are largely identical: some minor functional changes (which I will check), plus changes to copyrights and comments.  

The playlist and testng file entries relating to this look suitable for inclusion in the open source repo.

@llxia do you know why this wasn't moved along with all the other tests? Unless you have an objection, I suggest that I migrate this test completely to the open source repo.

llxia:
@TianyuZuo any insight on this?

TianyuZuo:
This file was initially moved into internal because it has extra dependencies which may not be valid to public. Since most of the invalid parts have been migrated to closed repo, this test may have become valid to public. 

The duplicate file in OpenJ9 is because some one migrated it out but didn't delete the internal one. And the exclude file is effective as it is not running anywhere. If this test needs to be run, please delete its exclusion from OpenJ9 repo and internal repo. 

Upon enabling this test, please add this test into `JCL_Test_SE90` and `JCL_Test_SE80` since it is part of the JCL test bucket. They do not need to be stand alone test cases. Also please make sure its dependencies have been migrated as well.

pdbain-ibm:
@TianyuZuo Updated per your request.  Thanks!
When the changes to the internal repo have fully propagated I will retest
and will have the pull request ready tomorrow.

pdbain-ibm:
Personal build passed.  @smlambert Ready for review.

smlambert:
Jenkins test sanity plinux

PR #1093, opened 2018-02-01
Emit read barrier before dereference

liqunl:
Read barrier is needed before dereferencing an object at compile time
when in concurrentScavenge, this is because object can be moved in
parallel with java and JIT compilation threads.

### DIFF
--- a/runtime/compiler/trj9/env/VMJ9.cpp
+++ b/runtime/compiler/trj9/env/VMJ9.cpp
@@ -1280,6 +1280,12 @@ TR_J9VMBase::getReferenceFieldAtAddress(uintptrj_t fieldAddress)
    {
    TR_ASSERT(haveAccess(), "Must haveAccess in getReferenceFieldAtAddress");
    TR::Compilation* comp = TR::comp();
+#if defined(OMR_GC_CONCURRENT_SCAVENGER)
+   // Emit read barrier
+   if (TR::Compiler->om.shouldGenerateReadBarriersForFieldLoads())
+      vmThread()->javaVM->javaVM->memoryManagerFunctions->J9ReadBarrier(vmThread(), (fj9object_t *)fieldAddress);
+#endif
+
 #if defined(J9VM_GC_COMPRESSED_POINTERS)
    uintptrj_t compressedResult = *(uint32_t*)fieldAddress;
    return (compressedResult << TR::Compiler->om.compressedReferenceShift()) + TR::Compiler->vm.heapBaseAddress();
@@ -1292,7 +1298,7 @@ uintptrj_t
 TR_J9VMBase::getReferenceFieldAt(uintptrj_t objectPointer, uintptrj_t fieldOffset)
    {
    TR_ASSERT(haveAccess(), "Must haveAccess in getReferenceFieldAt");
-   return getReferenceFieldAtAddress(objectPointer + sizeof(J9Object) + fieldOffset);
+   return (uintptrj_t)J9OBJECT_OBJECT_LOAD(vmThread(), objectPointer, J9_OBJECT_HEADER_SIZE + fieldOffset);
    }
 
 uintptrj_t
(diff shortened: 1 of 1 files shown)


liqunl:
@fjeremic @amicic Can you review my change?

runtime/compiler/trj9/env/VMJ9.cpp
```diff
@@ -1280,6 +1280,12 @@ TR_J9VMBase::getReferenceFieldAtAddress(uintptrj_t fieldAddress)
    {
    TR_ASSERT(haveAccess(), "Must haveAccess in getReferenceFieldAtAddress");
    TR::Compilation* comp = TR::comp();
+   // Emit read barrier
```
fjeremic: Can we push the comment into the `#if` so it doesn't dangle if the file is preprocessed without defining `OMR_GC_CONCURRENT_SCAVENGER`?

liqunl: Done

runtime/compiler/trj9/env/VMJ9.cpp
```diff
@@ -1292,7 +1298,11 @@ uintptrj_t
 TR_J9VMBase::getReferenceFieldAt(uintptrj_t objectPointer, uintptrj_t fieldOffset)
    {
    TR_ASSERT(haveAccess(), "Must haveAccess in getReferenceFieldAt");
-   return getReferenceFieldAtAddress(objectPointer + sizeof(J9Object) + fieldOffset);
+   if (TR::Compiler->om.shouldGenerateReadBarriersForFieldLoads())
+      return (uintptrj_t)vmThread()->javaVM->javaVM->memoryManagerFunctions->j9gc_objaccess_mixedObjectReadObject(vmThread(),
+         (J9Object*)objectPointer, J9_OBJECT_HEADER_SIZE + fieldOffset, false);
+   else
```
dsouzai: Not sure if it matters for performance, but if it does, you can guard the if/else with J9VM_GC_COMPRESSED_POINTERS, ie:

```
-   return getReferenceFieldAtAddress(objectPointer + sizeof(J9Object) + fieldOffset);
+   if (TR::Compiler->om.shouldGenerateReadBarriersForFieldLoads())
+      return (uintptrj_t)vmThread()->javaVM->javaVM->memoryManagerFunctions->j9gc_objaccess_mixedObjectReadObject(vmThread(),
+         (J9Object*)objectPointer, J9_OBJECT_HEADER_SIZE + fieldOffset, false);
+   else
+#endif
+      return getReferenceFieldAtAddress(objectPointer + sizeof(J9Object) + fieldOffset);
```

runtime/compiler/trj9/env/VMJ9.cpp
```diff
@@ -1292,7 +1298,11 @@ uintptrj_t
 TR_J9VMBase::getReferenceFieldAt(uintptrj_t objectPointer, uintptrj_t fieldOffset)
    {
    TR_ASSERT(haveAccess(), "Must haveAccess in getReferenceFieldAt");
-   return getReferenceFieldAtAddress(objectPointer + sizeof(J9Object) + fieldOffset);
+   if (TR::Compiler->om.shouldGenerateReadBarriersForFieldLoads())
```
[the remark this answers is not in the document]
amicic: After the change getReferenceFieldAt and getVolatileReferenceFieldAt will be effectively equivalent, and one should go away.

runtime/compiler/trj9/env/VMJ9.cpp
liqunl: Do you mean by guarding the if/else with `OMR_GC_CONCURRENT_SCAVENGER`? I don't think there is a performance issue because `getReferenceFieldAt` is not called very often. Plus it is used at compile time, it should at least not affect the run time performance and I highly doubt it will affect the compilation time.

dsouzai: >Do you mean by guarding the if/else with OMR_GC_CONCURRENT_SCAVENGER? 

Yeah sorry, typo on my part. 

> I don't think there is a performance issue because getReferenceFieldAt is not called very often.

Ah then I suppose it doesn't matter.

runtime/compiler/trj9/env/VMJ9.cpp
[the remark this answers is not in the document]
liqunl: Right, my bad. I was modifying `TR_J9VMBase::getReferenceElement` then I was side tracked to `getReferenceFieldAtAddress`. Working on the change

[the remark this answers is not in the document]
amicic: Many methods in this class duplicates the knowledge of object layout (compressed ref or not, contiguous arrays or not etc) and dynamics (barrier needed or not) , that is effectively part of core VM (of course, the class name is TR_J9VMBase). For many of these things there is already API or macro that exists elsewhere in core VM (like j9accessbarrier.h) that are aware of all these variables. Unless there is a performance aspect to optimize specific case, or if API happen not to exist, you should just pass the call to those core VM APIs/macros.

For original purpose of this change (adding awareness of read barrier) you need to treat getReferenceFieldAt (just pass it to  J9OBJECT_OBJECT_LOAD) and getReferenceElement  (just pass it to J9JAVAARRAYOFOBJECT_LOAD).

But there are others that outside of scope of this change, that would be wise to simplify, like getInt32Element (just pass it to J9JAVAARRAYOFINT_LOAD), the one about array length, but some others too.

One case that of non existent macro is J9OBJECT__PRE_OBJECT_LOAD, but it should really exist., and then used in getReferenceFieldAtAddress. Instead of waiting for one to be created,  its ok to just call J9ReadBarrier (as you already did).

Interestingly, there seems to be nothing in this class that does a plain 'store' (there are some compare-and-swaps, though), so there is no need for any extra write barrier awareness (no surprise, otherwise, our concurrent marking GCs would already have experienced problems).

And hopefully the last thing, perhaps it's ok to keep both  getReferenceFieldAt,  getVolatileReferenceFieldAt, they do differ in the value of the last argument (hard coded false vs IS_VOLATILE()).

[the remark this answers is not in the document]
fjeremic: > For many of these things there is already API or macro that exists elsewhere in core VM (like j9accessbarrier.h) that are aware of all these variables. Unless there is a performance aspect to optimize specific case, or if API happens not to exist, you should just pass the call to those core VM APIs/macros.

I definitely agree! Most JIT developers have not stepped out of the JIT boundary so we are mostly ignorant when it comes to all the APIs the VM supports. I saw a bunch of stuff related to strings that we are re-inventing in the JIT instead of calling well defined VM APIs. We need to do a better job moving forward. Reviews like this are a great resource for us to learn about existing APIs so thanks for this!

liqunl:
@fjeremic @amicic I've updated the change and the new commit has passed a pbuild. Tested the change with a debug counter on DT3, the modified APIs were not called, so there is very little/no performance impact.

runtime/compiler/trj9/env/VMJ9.h
```diff
@@ -505,8 +505,7 @@ class TR_J9VMBase : public TR_FrontEnd
    uint32_t                   getAllocationSize(TR::StaticSymbol *classSym, TR_OpaqueClassBlock * clazz);
 
    virtual TR_OpaqueClassBlock *getObjectClass(uintptrj_t objectPointer);
-   virtual uintptrj_t           getReferenceFieldAt(uintptrj_t objectPointer, uintptrj_t offsetFromHeader);
-   virtual uintptrj_t           getVolatileReferenceFieldAt(uintptrj_t objectPointer, uintptrj_t offsetFromHeader);
+   virtual uintptrj_t           getReferenceFieldAt(uintptrj_t objectPointer, uintptrj_t offsetFromHeader, bool isVolatile = false);
    virtual uintptrj_t           getReferenceFieldAtAddress(uintptrj_t fieldAddress);
```
amicic: In my last comment I said (and still feel), it's cleaner to keep both getReferenceFieldAt and getVolatileReferenceFieldAt separate and thus avoid and extra argument and 'if (isVolatile)' check in getReferenceFieldAt.

But I'm fine with this, too - it's more up to JIT reviewers what prefer more.

fjeremic: I would tend to agree with @amicic on this one, especially because `isVolatile` is an optional argument. It may be best to keep the APIs separate so as to show the exact intent of the user. At the very least I'd like to see `isVolatile` not optional so it makes the consumer of the API really think about what they're doing.

liqunl: Thanks for the suggestion, I've reverted the change to getVolatileReferenceFieldAt

fjeremic:
Jenkins test sanity

Issue #1020, opened 2018-01-25
OpenJ9 version information contains "build 2.9"


pshipton:
The java.vm.version system property contains 2.9, which doesn't make much sense for OpenJ9. The "2.9" also ends up in the system property java.fullversion. The -version output prints "build ${java.vm.version}"

pshipton:
All the VM/JIT/GC shared libraries also use a 29 suffix.

pshipton:
Note that changing this is not a trivial effort (but not major either). There is testing that expects the 2.9, or 29 shared library suffix which all needs to be fixed, although it should mostly use a variable.

DanHeidinga:
@JasonFengJ9 Can you take a look a this?

JasonFengJ9:
Working on it ...

JasonFengJ9:
Discussed with @pshipton, the share library suffix are not going to be changed for now before a conclusion is reached about the new suffix via an issue.

pshipton:
#1095 has been merged

PR #1095, opened 2018-02-01
Remove JVM version 2.9

JasonFengJ9:
Remove JVM version `2.9`

Added a vendor specific define

This PR meets the minimum requirement to remove JVM version `2.9` from system properties, and replace it with `OpenJ9 JVM_VERSION_STRING`.
There are other internal references to JVM version such as `kca_offsets_generator.cpp` and `share classes`, which use this version as their support levels. Will address that in a separated PR instead.

Related to #1020, #1096

Related PRs:
https://github.com/ibmruntimes/openj9-openjdk-jdk9/pull/121
https://github.com/ibmruntimes/openj9-openjdk-jdk8/pull/36

Reviewer @pshipton 
FYI @DanHeidinga 

### DIFF
--- a/runtime/cfdumper/main.c
+++ b/runtime/cfdumper/main.c
@@ -1855,7 +1855,7 @@ static void dumpHelpText( J9PortLibrary *portLib, int argc, char **argv)
 	PORT_ACCESS_FROM_PORT(portLib);
 
 	vmDetailString(portLib, detailString, 1024);
-	j9file_printf( PORTLIB, J9PORT_TTY_OUT, "\nOpenJ9 Java(TM) Class File Reader, Version " EsVersionString);
+	j9file_printf( PORTLIB, J9PORT_TTY_OUT, "\nOpenJ9 Java(TM) Class File Reader, Version " J9JVM_VERSION_STRING);
 	j9file_printf( PORTLIB, J9PORT_TTY_OUT, "\n" J9_COPYRIGHT_STRING);
 	j9file_printf( PORTLIB, J9PORT_TTY_OUT, "\nTarget: %s\n", detailString);
 	j9file_printf( PORTLIB, J9PORT_TTY_OUT, "\nJava and all Java-based marks and logos are trademarks or registered");
--- a/runtime/include/j9cfg.h.ftl
+++ b/runtime/include/j9cfg.h.ftl
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 1998, 2017 IBM Corp. and others
+ * Copyright (c) 1998, 2018 IBM Corp. and others
  *
  * This program and the accompanying materials are made available under
  * the terms of the Eclipse Public License 2.0 which accompanies this
@@ -33,6 +33,8 @@ extern "C" {
 #endif
 
 #include "omrcfg.h"
+#include "openj9_version_info.h"
+#include "vendor_version.h"
 
 #define J9_COPYRIGHT_STRING "(c) Copyright 1991, ${uma.year} IBM Corp. and others."
 
@@ -42,7 +44,6 @@ extern "C" {
 #undef EsVersionMinor
 #define EsVersionMinor ${uma.buildinfo.version.minor}0
 
-#define EsVersionString "${uma.buildinfo.version.major}.${uma.buildinfo.version.minor}"
 #define EsExtraVersionString ""
 
 /*  Note: The following defines record flags used to build VM.  */
--- a/runtime/include/vendor_version.h
+++ b/runtime/include/vendor_version.h
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2017, 2017 IBM Corp. and others
+ * Copyright (c) 2017, 2018 IBM Corp. and others
  *
  * This program and the accompanying materials are made available under
  * the terms of the Eclipse Public License 2.0 which accompanies this
@@ -20,7 +20,11 @@
  * SPDX-License-Identifier: EPL-2.0 OR Apache-2.0 OR GPL-2.0 WITH Classpath-exception-2.0 OR LicenseRef-GPL-2.0 WITH Assembly-exception
  *******************************************************************************/
 
-/* Example usage for inclusion of a vendor name and repository sha.  These values
+/* This file provides a means to supply vendor specific version info such as 
+ * short name, SHA, and version string.
+ * These vendor version info can be defined either in this file or other places.
+ * 
+ * Example usage for inclusion of a vendor name and repository sha.  These values
  * will be inserted into the java.fullversion and java.vm.info system properties
  * and in a generated javacore file.
  *
@@ -31,4 +35,8 @@
  *
  * #define VENDOR_SHORT_NAME "ABC"
  * #define VENDOR_SHA "1a2b3c4"
+ *
+ * Example usage for inclusion of a vendor version string.
+ * This value will be stored in the system property java.vm.version.
+ * #define J9JVM_VERSION_STRING "0.8.1"
  */
--- a/runtime/gc_check/CheckCycle.cpp
+++ b/runtime/gc_check/CheckCycle.cpp
@@ -1,6 +1,6 @@
 
 /*******************************************************************************
- * Copyright (c) 1991, 2014 IBM Corp. and others
+ * Copyright (c) 1991, 2018 IBM Corp. and others
  *
  * This program and the accompanying materials are made available under
  * the terms of the Eclipse Public License 2.0 which accompanies this
@@ -80,7 +80,7 @@ GC_CheckCycle::printHelp(J9PortLibrary *portLib)
 {
 	PORT_ACCESS_FROM_PORT(portLib);
 
-	j9tty_printf(PORTLIB, "gcchk for J9, Version " EsVersionString "\n");
+	j9tty_printf(PORTLIB, "gcchk for J9, Version " J9JVM_VERSION_STRING "\n");
 	j9tty_printf(PORTLIB, J9_COPYRIGHT_STRING "\n\n");
 	j9tty_printf(PORTLIB, "Usage: -Xcheck:gc[:scanOption,...][:verifyOption,...][:miscOption,...]\n");
 	j9tty_printf(PORTLIB, "scan options (default is all):\n");
--- a/runtime/jnichk/jnicheck.c
+++ b/runtime/jnichk/jnicheck.c
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 1991, 2017 IBM Corp. and others
+ * Copyright (c) 1991, 2018 IBM Corp. and others
  *
  * This program and the accompanying materials are made available under
  * the terms of the Eclipse Public License 2.0 which accompanies this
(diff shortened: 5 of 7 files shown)


runtime/include/vendor_version.h
```diff
@@ -32,3 +32,10 @@
  * #define VENDOR_SHORT_NAME "ABC"
  * #define VENDOR_SHA "1a2b3c4"
  */
+
+#ifndef vendor_version_h
+#define vendor_version_h
+
+#define VENDOR_VERSION_STRING "0.8.0-internal"
```
pshipton: This change seems unnecessary since the same value is found in j9cfg.h.ftl

runtime/include/vendor_version.h
```diff
@@ -32,3 +32,10 @@
  * #define VENDOR_SHORT_NAME "ABC"
  * #define VENDOR_SHA "1a2b3c4"
  */
+
```
pshipton: Please document VENDOR_VERSION_STRING in the comments above.

pshipton:
jenkins test sanity zlinux
jenkins test extended zlinux

runtime/include/vendor_version.h
JasonFengJ9: Removed it to following existing pattern, use the value in `j9cfg.h.ftl`.

runtime/include/vendor_version.h
JasonFengJ9: Updated.

pshipton:
jenkins test sanity zlinux
jenkins test extended zlinux

runtime/include/j9cfg.h.ftl
```diff
@@ -42,7 +43,12 @@ extern "C" {
 #undef EsVersionMinor
 #define EsVersionMinor ${uma.buildinfo.version.minor}0
 
-#define EsVersionString "${uma.buildinfo.version.major}.${uma.buildinfo.version.minor}"
+#ifdef VENDOR_VERSION_STRING
+#define EsVersionString VENDOR_VERSION_STRING
+#else
+#define EsVersionString "0.8.0-internal"
```
DanHeidinga: I don't think we want to hardcode this value or we'll need to continual update it for each release and each milestone.  Is there a way we can generate it based on the branch name / sha / tag?

ie: For the 0.8 release, we'll create a `openj9-0.8` branch, and then apply tags like `openj9-0.8RC1` to that branch.  It would be nice if this version string used those tags for the release branches and maybe `openj9-master-sha` (?) for the master branch.

runtime/include/vendor_version.h
```diff
@@ -20,7 +20,9 @@
  * SPDX-License-Identifier: EPL-2.0 OR Apache-2.0 OR GPL-2.0 WITH Classpath-exception-2.0 OR LicenseRef-GPL-2.0 WITH Assembly-exception
  *******************************************************************************/
 
-/* Example usage for inclusion of a vendor name and repository sha.  These values
+/* Vendor specific version info which is for OpenJ9 by default.
```
pshipton: This sentence needs to be clarified.

pshipton:
Noting that both sanity and extended testing has passed with the change to java.vm.version.

runtime/include/j9cfg.h.ftl
JasonFengJ9: Agreed. Working a way to grab `openj9` branch/tag/sha info before invoking target `j9vm-build`, and pass such info `$(VENDOR_VERSION_STRING)` to the target via `VENDOR_VERSION_STRING=$(VENDOR_VERSION_STRING)`.

runtime/include/vendor_version.h
```diff
@@ -20,7 +20,11 @@
  * SPDX-License-Identifier: EPL-2.0 OR Apache-2.0 OR GPL-2.0 WITH Classpath-exception-2.0 OR LicenseRef-GPL-2.0 WITH Assembly-exception
  *******************************************************************************/
 
-/* Example usage for inclusion of a vendor name and repository sha.  These values
+/* This file provides a mean to provide vendor specific version info such as
```
DanHeidinga: `This file provides a means to set vendor specific...`

runtime/include/j9cfg.h.ftl
```diff
@@ -42,7 +43,13 @@ extern "C" {
 #undef EsVersionMinor
 #define EsVersionMinor ${uma.buildinfo.version.minor}0
 
-#define EsVersionString "${uma.buildinfo.version.major}.${uma.buildinfo.version.minor}"
+#undef EsVersionString
+#ifdef VENDOR_VERSION_STRING
+#define EsVersionString VENDOR_VERSION_STRING
+#else
+#define EsVersionString "0.8.0-internal.j9cfg"
```
DanHeidinga: I'm not sure I understand the `0.8.0-internal.j9cfg` value.  Does that get replaced by some other process?

JasonFengJ9: Sorry, my intermediate change was pushed to this branch, should be updated soon.

runtime/include/vendor_version.h
JasonFengJ9: Yeah, I noticed that as well, will update it.

runtime/include/j9cfg.h.ftl
JasonFengJ9: Pls have another look.

JasonFengJ9:
The travis CI build failure is due to missing PR companies ibmruntimes/openj9-openjdk-jdk9#121 and
ibmruntimes/openj9-openjdk-jdk8#36.

runtime/include/j9cfg.h.ftl
JasonFengJ9: Updated PR to use `OpenJ9 tag` or `branch/sha` when `tag` is not available.
Pls review again.

runtime/include/j9cfg.h.ftl
```diff
@@ -33,6 +33,10 @@ extern "C" {
 #endif
 
 #include "omrcfg.h"
+#include "vendor_version.h"
+#ifndef JVM_VERSION_STRING
```
pshipton: This doesn't seem extensible, if more defines are added in the future or something else needs to be overridden. I'm thinking vendor_version.h should just be included after openj9_version_info.h so it can override the values.

JasonFengJ9: `JVM_VERSION_STRING` is not defined within `OpenJ9 vendor_version.h` but it is defined in vendor specific (such as IBM) `vendor_version.h`.
So if `JVM_VERSION_STRING` is not defined, `openj9_version_info.h` will be included to supply `OpenJ9` specific defines. Here checking `JVM_VERSION_STRING` is defined is to avoid checking `OPENJ9_BUILD` which requires adding the flag in `j9cfg.h` references.  
In the future if more defines are added, they are need to be provided within `vendor_version.h` (for vendor) or `OpenJ9 vendor_version.h` (for `OpenJ9`).
Will update the comments.

pshipton: I don't understand the argument. Imagine if there are two variables which are defined in openj9_version_info.h. How would you write the #if? i.e. #if '!defined(A) && !defined(B)` or '!defined(A) || !defined(B)`. What happens if vendor_version.h only needs to define one of them?

JasonFengJ9: Ideally the includes in `j9cfg.h` should look like following:
```
#if defined(OPENJ9_BUILD)
#include "../util/openj9_version_info.h"
#endif 
```
However currently `OPENJ9_BUILD` flag is enabled for individual `c/cpp` files via `module.xml` and not for head files. It can be changed to move the flag definition to the top makefile level. I feel it should be done in a separated PR. 
At mean time, this PR proposes an intermediate fix to remove the check for `JVM_VERSION_STRING` and include `openj9_version_info.h` anyway. Since this head file is only generated for `OpenJ9`, so vendor specific builds (such as IBM SDK) have to add such a head file but empty.

@DanHeidinga @pshipton does this make sense?

pshipton: It makes sense to me to define `OPENJ9_BUILD` at the top level for all files, instead of for individual files. Still vendor_version.h needs to be included after openj9_version_info.h

JasonFengJ9: Updated PR to include `vendor_version.h` after `openj9_version_info.h`.

runtime/include/j9cfg.h.ftl
```diff
@@ -33,6 +33,8 @@ extern "C" {
 #endif
 
 #include "omrcfg.h"
+#include "../util/openj9_version_info.h"
```
DanHeidinga: Why is this header file in util/ and not in either /include or /oti?

runtime/include/vendor_version.h
```diff
@@ -20,7 +20,11 @@
  * SPDX-License-Identifier: EPL-2.0 OR Apache-2.0 OR GPL-2.0 WITH Classpath-exception-2.0 OR LicenseRef-GPL-2.0 WITH Assembly-exception
  *******************************************************************************/
 
-/* Example usage for inclusion of a vendor name and repository sha.  These values
+/* This file provides a mean to supply vendor specific version info such as
```
DanHeidinga: nitpick: `mean` -> `means`

runtime/include/j9cfg.h.ftl
DanHeidinga: We typically change the include path rather than using relative paths to get to header files.

I think moving the header is probably a better choice.

runtime/cfdumper/main.c
```diff
@@ -1855,7 +1855,7 @@ static void dumpHelpText( J9PortLibrary *portLib, int argc, char **argv)
 	PORT_ACCESS_FROM_PORT(portLib);
 
 	vmDetailString(portLib, detailString, 1024);
-	j9file_printf( PORTLIB, J9PORT_TTY_OUT, "\nOpenJ9 Java(TM) Class File Reader, Version " EsVersionString);
+	j9file_printf( PORTLIB, J9PORT_TTY_OUT, "\nOpenJ9 Java(TM) Class File Reader, Version " JVM_VERSION_STRING);
```
DanHeidinga: Should we put this into a `OPENJ9_` namespace?  I'm concerned this may conflict with a definition from OpenJDK

JasonFengJ9: How about `J9JVM_VERSION_STRING`? hopefully it applies to both `OpenJ9` and internal builds.

runtime/include/j9cfg.h.ftl
JasonFengJ9: Moved generated `openj9_version_info.h` from folder `util` to `include`.

runtime/include/vendor_version.h
JasonFengJ9: Fixed.

runtime/include/vendor_version.h
```diff
@@ -31,4 +35,8 @@
  *
  * #define VENDOR_SHORT_NAME "ABC"
  * #define VENDOR_SHA "1a2b3c4"
+ *
+ * Example usage for inclusion of a vendor version string.
+ * This value will be inserted into the system property java.vm.version.
```
pshipton: I don't think the word `inserted` is appropriate. java.vm.version is set to this value.

JasonFengJ9: Fixed. Pls have another look.

runtime/include/vendor_version.h
```diff
@@ -31,4 +35,8 @@
  *
  * #define VENDOR_SHORT_NAME "ABC"
  * #define VENDOR_SHA "1a2b3c4"
+ *
+ * Example usage for inclusion of a vendor version string.
+ * This value will be set to the system property java.vm.version.
```
keithc-ca: I think this should say 'will be *used* to set the system property'.

keithc-ca: Or: This value will be stored in the system property java.vm.version.

JasonFengJ9: Changed to `This value will be stored to` to avoid confusion.

DanHeidinga:
The travis build is failing:
```
In file included from ../../../oti/j9.h:29:0,
                 from ../../../gc_glue_java/objectdescription.h:26,
                 from omrgcstartup.cpp:26:
../../../include/j9cfg.h:36:33: fatal error: openj9_version_info.h: No such file or directory
 #include "openj9_version_info.h"
```

JasonFengJ9:
@DanHeidinga the travis failure was due to the dependence of ibmruntimes/openj9-openjdk-jdk9#121 & ibmruntimes/openj9-openjdk-jdk8#36 which generate openj9_version_info.h files for Java 8/9 respectively. A personal build passed successfully.

DanHeidinga:
Thanks @JasonFengJ9 - my bad.

pshipton:
jenkins test sanity plinux jdk9 depends ibmruntimes/openj9-openjdk-jdk9#121

PR #1098, opened 2018-02-02
Use decompressedArrayCopy in StringBuilder/StringBuffer to remove redundant bound checks

yanluo7:
- Use String.decompressedArrayCopy when array bound safety is already ensured
- Use String.getCharsNoBoundChecks when array bound safety is already ensured

Signed-off-by: Yan Luo <yan_luo@ca.ibm.com>

### DIFF
--- a/jcl/src/java.base/share/classes/java/lang/StringBuffer.java
+++ b/jcl/src/java.base/share/classes/java/lang/StringBuffer.java
@@ -227,7 +227,7 @@ public StringBuffer (String string) {
 			value = new char[newLength];
 			/*[ENDIF]*/
 
-			string.getChars(0, stringLength, value, 0);
+			string.getCharsNoBoundChecks(0, stringLength, value, 0);
 
 			/*[IF !Sidecar19-SE]*/
 			capacity = newLength;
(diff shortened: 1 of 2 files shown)


yanluo7:
Performance implication:

This is extension to work done in https://github.com/eclipse/openj9/pull/753, which uses `String.decompressedArrayCopy` and `String.getCharsNoBoundChecks` for `StringBuilder.append(String)`. There are other such opportunities in `StringBuilder`/`StringBuffer` (both Java classes are near identical only that `StringBuffer` methods are `synchronized`) when we know the copy does not exceed array bounds -- sufficiently large array buffer has already been allocated. `String.decompressedArrayCopy` and `System.arraycopy` generate the same JIT code, only difference is that the latter generates additional array bound checks. We measured humble bench performance for APIs `append`, `delete`, `insert`, `replace`, `move`, `setLength`, `trimToSize`. They show an average of 3-5% improvement under humble bench.

yanluo7:
Launched JVM 29 personal build level sanity, regression, promotion: Ottawa build_id=377887

yanluo7:
@pshipton Could you review this?

pshipton:
jenkins test sanity plinux

pshipton:
jenkins test extended plinux

pshipton:
JDK9 sanity tests failed due to #968, retrying

jenkins test sanity plinux jdk9

pshipton:
and again

jenkins test sanity all jdk9

pshipton:
Finally got a passing sanity run for jdk9 on plinux. However as backup I did a jdk9 sanity on zlinux and it failed. It seems to have passed all the tests but failed the job. Perhaps its #1114 @AdamBrousseau?

I'm going to go ahead and merge since no testing failed.

AdamBrousseau:
Yes, failed because #1114

PR #1100, opened 2018-02-02
Fix unresolved symbol error loading JIT on Linux IA32

0xdaryl:
`compiler/x/i386/J9CodeGenerator.cpp` is missing an include (`il/Node_inlines.hpp`)
which is preventing the JIT from loading on IA32 Linux due to unresolved
symbols.


fjeremic:
No need for Jenkins builds here as this is an x86 only change. I will merge if the Travis-CI build passes.

Issue #1086, opened 2018-02-01
StaticAccessChecks-IllegalAccessDeny fails due to java/lang/IllegalAccessError: java.lang.Object


pshipton:
The failure is seen in the nightly extended test build i.e. https://ci.eclipse.org/openj9/job/Test-Extended-JDK9-linux_ppc-64_cmprssptrs_le/55/console

```
02:58:51 Running test StaticAccessChecks-IllegalAccessDeny_0 ...
02:58:51 ===============================================
02:58:51 test with NoOptions
02:58:51 { "/home/jenkins/jenkins-agent/workspace/Test-Extended-JDK9-linux_ppc-64_cmprssptrs_le/build/linux-ppc64le-normal-server-release/images/jdk/bin/java"  -Xcompressedrefs  --illegal-access=deny \
02:58:51 -cp "/home/jenkins/jenkins-agent/workspace/Test-Extended-JDK9-linux_ppc-64_cmprssptrs_le/openj9/test/TestConfig/scripts/testKitGen/../../../../jvmtest/TestConfig/resources:/home/jenkins/jenkins-agent/workspace/Test-Extended-JDK9-linux_ppc-64_cmprssptrs_le/openj9/test/TestConfig/scripts/testKitGen/../../../../jvmtest/TestConfig/lib/testng.jar:/home/jenkins/jenkins-agent/workspace/Test-Extended-JDK9-linux_ppc-64_cmprssptrs_le/openj9/test/TestConfig/scripts/testKitGen/../../../../jvmtest/TestConfig/lib/jcommander.jar:/home/jenkins/jenkins-agent/workspace/Test-Extended-JDK9-linux_ppc-64_cmprssptrs_le/openj9/test/TestConfig/scripts/testKitGen/../../../../jvmtest/Java9andUp/GeneralTest.jar:/home/jenkins/jenkins-agent/workspace/Test-Extended-JDK9-linux_ppc-64_cmprssptrs_le/openj9/test/TestConfig/scripts/testKitGen/../../../../jvmtest/TestConfig/lib/asm-all.jar" \
[...]

PR #1103, opened 2018-02-05
Ensure primitive types are assigned to java.base

tajila:
Ensure primitive types are assigned to java.base

Java primitive types are currently loaded under unnamedmodules. This
patch fixes this and correctly loads them under java.base.

Fixes: #1086

### DIFF
--- a/runtime/vm/createramclass.cpp
+++ b/runtime/vm/createramclass.cpp
@@ -2700,6 +2700,10 @@ fail:
 #if defined(J9VM_THR_LOCK_NURSERY) && defined(J9VM_THR_LOCK_NURSERY_FAT_ARRAYS)
 				ramArrayClass->lockOffset = (UDATA)TMP_OFFSETOF_J9INDEXABLEOBJECT_MONITOR;
 #endif
+			} else if (J9ROMCLASS_IS_PRIMITIVE_TYPE(ramClass->romClass)) {
+				ramClass->module = module;
+			} else {
+				Assert_VM_unreachable();
 			}
 		}
 	}
@@ -2785,7 +2789,14 @@ retry:
 
 	if (J2SE_VERSION(javaVM) >= J2SE_19) {
 		if (NULL == classBeingRedefined) {
-			if (J9_ARE_ALL_BITS_SET(javaVM->runtimeFlags, J9_RUNTIME_JAVA_BASE_MODULE_CREATED)) {
+			if (J9ROMCLASS_IS_ARRAY(romClass)) {
+				/* At this point the elementClass has been loaded. No
+				 * need to check if java.base was created and no need to
+				 * do a lookup. Just assign the arrayClass the same module
+				 * which was used for the elementClass.
+				 */
+				module = elementClass->module;
+			} else if (J9_ARE_ALL_BITS_SET(javaVM->runtimeFlags, J9_RUNTIME_JAVA_BASE_MODULE_CREATED)) {
 				/* VM does not create J9Module for unnamed modules except the unnamed module for bootloader.
 				 * Therefore J9Class.module should be NULL for classes loaded from classpath by non-bootloaders.
 				 * The call to findModuleForPackage() should correctly set J9Class.module as it would return
@@ -2800,25 +2811,19 @@ retry:
 					|| ((LOAD_LOCATION_PATCH_PATH == locationType)
 					|| (LOAD_LOCATION_MODULE == locationType))
 				) {
-					U_8* packageUTF = NULL;
-					J9ROMClass *packageRomClass = romClass;
-					U_32 pkgNameLength = 0;
-
-					if (J9ROMCLASS_IS_ARRAY(romClass)) {
-						packageRomClass = elementClass->romClass;
-					}
-
-					packageUTF = J9UTF8_DATA(J9ROMCLASS_CLASSNAME(packageRomClass));
-					pkgNameLength = (U_32) packageNameLength(packageRomClass);
+					U_32 pkgNameLength = (U_32) packageNameLength(romClass);
 
 					omrthread_monitor_enter(javaVM->classLoaderModuleAndLocationMutex);
-					module = findModuleForPackage(vmThread, classLoader, packageUTF, pkgNameLength);
+					module = findModuleForPackage(vmThread, classLoader, J9UTF8_DATA(className), pkgNameLength);
 					omrthread_monitor_exit(javaVM->classLoaderModuleAndLocationMutex);
 				} else {
 					module = javaVM->unamedModuleForSystemLoader;
 				}
 			} else {
-				if ((LOAD_LOCATION_PATCH_PATH == locationType) || (LOAD_LOCATION_MODULE == locationType)) {
+				if (J9ROMCLASS_IS_PRIMITIVE_TYPE(romClass)
+					|| (LOAD_LOCATION_PATCH_PATH == locationType)
+					|| (LOAD_LOCATION_MODULE == locationType)
+				) {
 					module = javaVM->javaBaseModule;
 				} else {
 					module = javaVM->unamedModuleForSystemLoader;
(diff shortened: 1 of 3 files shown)


runtime/vm/createramclass.cpp
```diff
@@ -2700,6 +2700,9 @@ internalCreateRAMClassFromROMClassImpl(J9VMThread *vmThread, J9ClassLoader *clas
 #if defined(J9VM_THR_LOCK_NURSERY) && defined(J9VM_THR_LOCK_NURSERY_FAT_ARRAYS)
 				ramArrayClass->lockOffset = (UDATA)TMP_OFFSETOF_J9INDEXABLEOBJECT_MONITOR;
 #endif
+			} else {
```
DanHeidinga: Can you write this as:

```
} else if (J9ROMCLASS_IS_PRIMITIVE_TYPE(ramClass->romClass)) {
....
} else {
  Assert_VM_unreachable();
}
```

runtime/vm/createramclass.cpp
```diff
@@ -2785,7 +2788,9 @@ internalCreateRAMClassFromROMClass(J9VMThread *vmThread, J9ClassLoader *classLoa
 
 	if (J2SE_VERSION(javaVM) >= J2SE_19) {
 		if (NULL == classBeingRedefined) {
-			if (J9_ARE_ALL_BITS_SET(javaVM->runtimeFlags, J9_RUNTIME_JAVA_BASE_MODULE_CREATED)) {
+			if (J9ROMCLASS_IS_ARRAY(romClass)) {
```
DanHeidinga: Why is this check before the check if java.base is created?

runtime/vm/createramclass.cpp
```diff
@@ -2804,10 +2809,6 @@ internalCreateRAMClassFromROMClass(J9VMThread *vmThread, J9ClassLoader *classLoa
 					J9ROMClass *packageRomClass = romClass;
 					U_32 pkgNameLength = 0;
 
-					if (J9ROMCLASS_IS_ARRAY(romClass)) {
```
DanHeidinga: Also not sure why this was deleted.

runtime/vm/createramclass.cpp
tajila: Element classes are always loaded before array classes, so at this it doesn't matter if java.base is created or not because we will take the module from the element class.

runtime/vm/createramclass.cpp
tajila: We no longer need the array check here because array classes are simply assigned the same module as their element class.

pshipton:
jenkins test sanity plinux

pshipton:
jenkins test extended plinux jdk9

runtime/vm/romclasses.c
```diff
@@ -247,7 +247,7 @@ internalCreateBaseTypePrimitiveAndArrayClasses(J9VMThread *currentThread)
 				NULL, /* protectionDomain */
 				NULL, /* methodRemapArray */
 				J9_CP_INDEX_NONE, /* entryIndex */
-				LOAD_LOCATION_UNKNOWN, /* locationType */
+				LOAD_LOCATION_MODULE, /* locationType */
```
pshipton: This doesn't seem right for Java 8.

tajila: would `LOAD_LOCATION_CLASSPATH` be correct for java8?

DanHeidinga: I don't think so as that would imply that the classfile was loaded from the class path.  Thinking about it further, I'm not certain `LOAD_LOCATION_MODULE` is correct for 9 as primitives are generated by the VM.

pshipton:
Using the current commit, note that JDK8 santity testing passed, JDK9 santity testing failed due to known issue #968. JDK9 extended testing passed.

runtime/vm/romclasses.c
pshipton: Good point.

tajila: The other option is to keep it as `LOAD_LOCATION_UNKNOWN` and add a condition in createramclass to add primitives to java.base instead of the unnamedModule

tajila:
@DanHeidinga @pshipton changes are ready for another look

runtime/vm/createramclass.cpp
DanHeidinga: Can you add a comment to that effect to the code?  Makes it obvious to the next reader and leaves a breadcrumb for yourself / others.

runtime/vm/createramclass.cpp
```diff
@@ -2818,7 +2820,10 @@ internalCreateRAMClassFromROMClass(J9VMThread *vmThread, J9ClassLoader *classLoa
 					module = javaVM->unamedModuleForSystemLoader;
 				}
 			} else {
-				if ((LOAD_LOCATION_PATCH_PATH == locationType) || (LOAD_LOCATION_MODULE == locationType)) {
+				if (J9ROMCLASS_IS_PRIMITIVE_TYPE(romClass)
```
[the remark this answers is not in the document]
tajila: This particular code pre-emtively assigns the J9Class to the java.base J9Module. At this point it is not constructed, but when java.base is created it will be completed. Since the J9Class's already point to it they wont have to be updated.

Same applies for the bootloader unnamed module

runtime/vm/createramclass.cpp
tajila: I added a comment

runtime/vm/createramclass.cpp
```diff
@@ -2801,24 +2812,21 @@ internalCreateRAMClassFromROMClass(J9VMThread *vmThread, J9ClassLoader *classLoa
...
-					J9ROMClass *packageRomClass = romClass;
 					U_32 pkgNameLength = 0;
 
-					if (J9ROMCLASS_IS_ARRAY(romClass)) {
-						packageRomClass = elementClass->romClass;
-					}
-
-					packageUTF = J9UTF8_DATA(J9ROMCLASS_CLASSNAME(packageRomClass));
-					pkgNameLength = (U_32) packageNameLength(packageRomClass);
+					pkgNameLength = (U_32) packageNameLength(romClass);
```
tajila: I had forgotten to remove some code that was used for the array case. I fixed it to make it more clear.

runtime/vm/createramclass.cpp
```diff
@@ -2800,25 +2811,19 @@ internalCreateRAMClassFromROMClass(J9VMThread *vmThread, J9ClassLoader *classLoa
...
-					J9ROMClass *packageRomClass = romClass;
-					U_32 pkgNameLength = 0;
-
-					if (J9ROMCLASS_IS_ARRAY(romClass)) {
-						packageRomClass = elementClass->romClass;
-					}
-
-					packageUTF = J9UTF8_DATA(J9ROMCLASS_CLASSNAME(packageRomClass));
-					pkgNameLength = (U_32) packageNameLength(packageRomClass);
+					U_32 pkgNameLength = (U_32) packageNameLength(romClass);
```
tajila: I forgot  to remove some of the code that as used in the array case. I have fixed it to make it more clear.

runtime/vm/createramclass.cpp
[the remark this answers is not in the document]
DanHeidinga: Thanks for clarifying.  I had to check the code that initializes `javaBaseModule` as I thought it was done latter.

This looks fine to me.

DanHeidinga:
Jenkins test all plinux

Need to run both sanity and extended as the test change enables an extended test.

DanHeidinga:
Trying this again:

Jenkins test sanity plinux
Jenkins test extended plinux

pshipton:
Jenkins test sanity plinux

right now it doesn't work to try and run both sanity and extended from the same comment

pshipton:
Jenkins test sanity plinux

pshipton:
Apparently it also doesn't work to put a comment after the request to run tests

PR #1106, opened 2018-02-05
Fix hash table collision bug in Class.getMethods()

pdbain-ibm:
MethodInfo.equals() did not distinguish methods with the same name, causing
them to be dropped in the case of a hash table collision.

Add a specific test case.

Fix the equals() method.


pdbain-ibm:
FYI @llxia @smlambert @pshipton 
Tested in personal build (sanity level, Linux X86).

pshipton:
jenkins test sanity plinux
jenkins test extended plinux

pshipton:
jenkins test sanity plinux

pshipton:
JDK9 extended test failure is a known problem #1110.

PR #1118, opened 2018-02-06
Guard scratch memory initialization in a try/catch

dsouzai:


dsouzai:
WIP because I've yet to run a pbuild.

@mpirvu could you review the code?

runtime/compiler/trj9/control/CompilationThread.cpp
```diff
@@ -6136,58 +6136,38 @@ TR::CompilationInfoPerThreadBase::generatePerfToolEntry()
...
+ * @brief TR::CompilationInfoPerThreadBase::preCompilationTasks
+ * @param vmThread Pointer to the current J9VMThread
+ * @param javaVM Pointer to the J9JavaVM
+ * @param entry Pointer to the method to be compiled entry
+ * @param method Pointer to the J9Method
+ * @param aotCachedMethod Pointer to a pointer to the AOT code to be relocated (if it exists)
+ * @param trMemory Reference to the TR_Memory associated with this compilation
+ * @param canDoRelocatableCompile Reference to a bool; indicates if the current compilation can be AOT compiled
+ * @param eligibleForRelocatableCompile Reference to a bool; Indicates if the current compilation is eligible for AOT
+ * @param reloRuntime Pointer to a TR_RelocationRuntime; NULL if JVM does not have AOT support
```
mpirvu: I would like to specify which parameters are for input and which are for output. For instance `canDoRelocatableCompile` is an output parameter set the function itself.
Input parameters should be marked 'const' as much as possible.

runtime/compiler/trj9/control/CompilationThread.cpp
```diff
@@ -6136,58 +6136,38 @@ TR::CompilationInfoPerThreadBase::generatePerfToolEntry()
...
-   void *startPC = 0;
-   J9Method *method = entry->getMethodDetails().getMethod();
-
-   vmThread->jitMethodToBeCompiled = method;
-
+/**
+ * @brief TR::CompilationInfoPerThreadBase::preCompilationTasks
+ * @param vmThread Pointer to the current J9VMThread
+ * @param javaVM Pointer to the J9JavaVM
+ * @param entry Pointer to the method to be compiled entry
```
mpirvu: Typo here: `Pointer to the method to be compiled entry`

runtime/compiler/trj9/control/CompilationThread.cpp
```diff
@@ -6403,97 +6377,41 @@ TR::CompilationInfoPerThreadBase::compile(J9VMThread * vmThread,
...
-      }
-
-
+   return vm;
+   }
+
+/**
+ * @brief TR::CompilationInfoPerThreadBase::postCompilationTasks
+ * @param vmThread Pointer to the current J9VMThread
+ * @param entry Pointer to the method to be compiled entry
```
mpirvu: Again, this doesn't sound right: `Pointer to the method to be compiled entry`

runtime/compiler/trj9/control/CompilationThread.cpp
```diff
@@ -6403,97 +6377,41 @@ TR::CompilationInfoPerThreadBase::compile(J9VMThread * vmThread,
...
+ * @param method Pointer to the J9Method
+ * @param aotCachedMethod Pointer to the AOT code to be relocated (if it exists)
+ * @param metaData Pointer to the TR_MethodMetaData created on a successful compile
+ * @param compParam Reference to the CompileParameters associated with the current compilation
+ * @param canDoRelocatableCompile Indicates if the current compilation could be AOT compiled
+ * @param eligibleForRelocatableCompile Indicates if the current compilation was eligible for AOT
+ * @param reloRuntime Pointer to a TR_RelocationRuntime; NULL if JVM does not have AOT support
+ * @param scratchMemoryAvailable Indicates if the scratch memory infrastructre for the current compilation can be used
```
mpirvu: typo for infrastructure

runtime/compiler/trj9/control/CompilationThread.cpp
```diff
@@ -6403,97 +6377,41 @@ TR::CompilationInfoPerThreadBase::compile(J9VMThread * vmThread,
...
+TR::CompilationInfoPerThreadBase::postCompilationTasks(J9VMThread * vmThread,
+                                                       TR_MethodToBeCompiled *entry,
+                                                       J9Method *method,
+                                                       const void *aotCachedMethod,
+                                                       TR_MethodMetaData *metaData,
+                                                       CompileParameters &compParam,
+                                                       bool canDoRelocatableCompile,
+                                                       bool eligibleForRelocatableCompile,
+                                                       TR_RelocationRuntime *reloRuntime,
+                                                       bool scratchMemoryAvailable)
```
mpirvu: I saw that `scratchMemoryAvailable` is not used in this function. Where or how could it be used?

runtime/compiler/trj9/control/CompilationThread.cpp
```diff
@@ -6403,97 +6377,41 @@ TR::CompilationInfoPerThreadBase::compile(J9VMThread * vmThread,
...
+ * @param reloRuntime Pointer to a TR_RelocationRuntime; NULL if JVM does not have AOT support
+ * @param scratchMemoryAvailable Indicates if the scratch memory infrastructre for the current compilation can be used
+ * @return StartPC of the compiled body, NULL if the compilation failed
+ *
+ * This method is used to perform a set of tasks needed after attempting a compilation. It is important to note
+ * that the scratch memory infrastructure (TR_Memory, TR::Region, TR::SegmentAllocator) may not exist within this
+ * method as this method is called in both the try and catch block in TR::CompilationInfoPerThreadBase::compile.
+ */
+void *
+TR::CompilationInfoPerThreadBase::postCompilationTasks(J9VMThread * vmThread,
```
mpirvu: Can this method throw? If so at what line? Since this method is executed in the catch block as well it's better if we can ensure that it can never throw.

runtime/compiler/trj9/control/CompilationThread.hpp
```diff
@@ -82,35 +82,33 @@ struct CompileParameters
...
          J9VMThread * vmThread,
          TR_OptimizationPlan * optimizationPlan,
-         TR::SegmentAllocator &scratchSegmentProvider,
-         TR::Region &dispatchRegion,
-         TR_Memory &trMemory,
          const TR::CompileIlGenRequest &ilGenRequest
       ) :
       _compilationInfo(compilationInfo),
-      _vm(vm),
+      _vm(NULL),
```
mpirvu: Partial construction of objects is error prone. The CompileParam object is mostly used to pass a bunch of parameters from the main `compile()` routine to `wrappedCompile()` which is static rather than a member function.
It is also used to pass the compiler object back. However, for this purpose we could use another parameter (pointer to TR::Compilation) or we could use `TR::CompilationInfoPerThreadBase *that` object because we do `     that->setCompilation(compiler);`
My suggestion would be to eliminate `_compiler` from `CompileParameters` and create the` CompilerParameters` object close to where it's needed (just before calling `wrappedCompile`). Then change `postCompilationTasks` so that it does not need `CompileParameters`

runtime/compiler/trj9/control/HookedByTheJit.cpp
```diff
@@ -4775,21 +4775,27 @@ void JitShutdown(J9JITConfig * jitConfig)
...
-      TR::Options::getScratchSpaceLimit(),
-      segmentAllocator,
-      rawAllocator
-      );
-   TR::Region dispatchRegion(regionSegmentProvider, rawAllocator);
-   TR_Memory trMemory(*compInfo->persistentMemory(), dispatchRegion);
-   if (options && options->getOption(TR_DumpFinalMethodNamesAndCounts))
-      compInfo->getPersistentInfo()->getPersistentCHTable()->dumpMethodCounts(vm, trMemory);
-   }
+   try
```
mpirvu: We create all these allocators just to print method counts under an option. I would say protect the entire logic with `if (options && options->getOption(TR_DumpFinalMethodNamesAndCounts))`.

runtime/compiler/trj9/control/CompilationThread.cpp
dsouzai: I don't think anything explicitly throws right now. Any exception that could be thrown would likely be a `std::bad_alloc()` thrown because of some memory allocated via the `TR_Memory` or `TR::Region` associated with the `comp` object; however, that would require the `comp` object to exist which could only happen in the try block (since it is set to NULL in the catch block). The other possibility is if `acquireVMaccessIfNeeded` is called, which can throw `TR::CompilationInterrupted` or `J9::JITShutdown`, though I don't think that's called at any point in `postCompilationTasks`. 

So I suppose it's technically possible for something to throw, but no one should since there is a problem with a throw within `postCompilationTasks`, namely:
```
   // Re-acquire the compilation monitor so that we can update the J9Method.
   //
   _compInfo.debugPrint(vmThread, "\tacquiring compilation monitor\n");
   _compInfo.acquireCompMonitor(vmThread);
   _compInfo.debugPrint(vmThread, "+CM\n");
   if (_onSeparateThread)
      {
      // Acquire the queue slot monitor now
      //
      _compInfo.debugPrint(vmThread, "\tre-acquiring queue-slot monitor\n");
      entry->acquireSlotMonitor(vmThread);
      _compInfo.debugPrint(vmThread, "+AM-", entry);
      }
```
which could end up running twice.

I'll update the comment at the top to state that an exception should not escape `post CompilationTasks`.

runtime/compiler/trj9/control/CompilationThread.cpp
dsouzai: I intended this as a means of determining if one can use the `TR_Memory` and `TR::Region` associated with the comp object to do any sort of temporary memory allocation. However, I realize that a non-NULL `comp` object implies that the the scratch memory infrastructure is available so this bool is unnecessary. Additionally, no one should be using any scratch memory since it might result in a `std::bad_alloc` thrown and, as stated in the previous comment, an exception should not escape `postCompilationTasks`.

dsouzai:
Ran a pbuild and everything looks good.

@mpirvu I updated the commits with the requested changes, could you re-review.

runtime/compiler/trj9/control/CompilationThread.cpp
```diff
@@ -6750,14 +6665,219 @@ TR::CompilationInfoPerThreadBase::compile(J9VMThread * vmThread,
...
+            {
+            TR_VerboseLog::writeLineLocked(TR_Vlog_FAILURE,"<EARLY TRANSLATION FAILURE: JIT Shutdown signaled>");
+            }
+         catch (const std::bad_alloc &e)
+            {
+            TR_VerboseLog::writeLineLocked(TR_Vlog_FAILURE,"<EARLY TRANSLATION FAILURE: out of scratch memory>");
+            }
+         catch (const std::exception &e)
+            {
+            TR_VerboseLog::writeLineLocked(TR_Vlog_FAILURE,"<EARLY TRANSLATION FAILURE: compilation aborted>");
```
mpirvu: What kind of exceptions can happen here? EARLY_TRANSLATION_FAILURE was a message appearing when we couldn't allocate the "compilee" (the resolved method that needed to be compiled which now allocated in preCompilationTasks). Since this 'try' section goes around the entire compilation process it may catch more than just early failures.

mpirvu:
As discussed, let's add some documentation about all the existent try/catch blocks, what their purpose is (what exceptions we are trying to catch) and how they nest.

dsouzai:
@mpirvu Made the suggest changes and added documentation as well. Could you please review both.

mpirvu:
Jenkins test sanity

PR #1119, opened 2018-02-06
Clean up queries related to SIMDStringCaseConv

cathyzhyi:
Clean up the queries to check supports of SIMDStringCaseConv
to make it easier to extend to platforms other than Z.

### DIFF
--- a/runtime/compiler/trj9/codegen/J9CodeGenerator.hpp
+++ b/runtime/compiler/trj9/codegen/J9CodeGenerator.hpp
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2000, 2017 IBM Corp. and others
+ * Copyright (c) 2000, 2018 IBM Corp. and others
  *
  * This program and the accompanying materials are made available under
  * the terms of the Eclipse Public License 2.0 which accompanies this
@@ -369,6 +369,15 @@ public:
    // Determines whether high-resolution timer can be used to implement java/lang/System.currentTimeMillis()
    bool getSupportsMaxPrecisionMilliTime() {return _j9Flags.testAny(SupportsMaxPrecisionMilliTime);}
    void setSupportsMaxPrecisionMilliTime() {_j9Flags.set(SupportsMaxPrecisionMilliTime);}
+   
+   /** \brief
+    *    Determines whether the code generator supports inlining of java/lang/String.toUpperCase() and toLowerCase()
+    */
+   bool getSupportsInlineStringCaseConversion() { return _j9Flags.testAny(SupportsInlineStringCaseConversion);}
+   /** \brief
+    *    The code generator supports inlining of java/lang/String.toUpperCase() and toLowerCase()
+    */
+   void setSupportsInlineStringCaseConversion() { return _j9Flags.set(SupportsInlineStringCaseConversion);}
 
    /**
     * \brief
@@ -381,8 +390,9 @@ private:
 
    enum // Flags
       {
-      HasFixedFrameC_CallingConvention = 0x00000001,
-      SupportsMaxPrecisionMilliTime    = 0x00000002
+      HasFixedFrameC_CallingConvention    = 0x00000001,
+      SupportsMaxPrecisionMilliTime       = 0x00000002,
+      SupportsInlineStringCaseConversion  = 0x00000004 /*! codegen inlining of Java string case conversion */ 
       };
 
    flags32_t _j9Flags;
--- a/runtime/compiler/trj9/env/VMJ9.cpp
+++ b/runtime/compiler/trj9/env/VMJ9.cpp
@@ -3110,7 +3110,7 @@ bool TR_J9VMBase::supressInliningRecognizedInitialCallee(TR_CallSite* callsite, 
          case TR::java_lang_String_toLowerHWOptimizedDecompressed:
          case TR::java_lang_String_toUpperHWOptimized:
          case TR::java_lang_String_toLowerHWOptimized:
-            if(comp->cg()->getSupportsVectorRegisters() && !comp->getOption(TR_DisableSIMDStringCaseConv))
+            if(comp->cg()->getSupportsInlineStringCaseConversion())
                {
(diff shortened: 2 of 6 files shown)


runtime/compiler/trj9/z/codegen/J9CodeGenerator.cpp
```diff
@@ -104,6 +104,9 @@ J9::Z::CodeGenerator::CodeGenerator() :
    if (!comp->getOption(TR_FullSpeedDebug))
       cg->setSupportsDirectJNICalls();
 
+   if (cg->getSupportsVectorRegisters())
+      cg->setSupportsSIMDStringCaseConv();
```
fjeremic: Did we check that `setSupportsVectorRegisters` is called before this point? It seems to be called in `TR::S390PrivateLinkage::S390PrivateLinkage`. We need to verify this is called before this point.

runtime/compiler/trj9/z/codegen/J9TreeEvaluator.cpp
```diff
@@ -12477,7 +12477,7 @@ J9::Z::CodeGenerator::inlineDirectCall(
       }
 #endif
 
-   if (!comp->getOption(TR_DisableSIMDStringCaseConv) && cg->getSupportsVectorRegisters())
+   if (cg->getSupportsSIMDStringCaseConv() && !comp->getOption(TR_DisableSIMDStringCaseConv))
```
fjeremic: Can we get rid of the `!comp->getOption(TR_DisableSIMDStringCaseConv)` check? This check should be done in `CodeGenerator` constructor where we call `setSupportsSIMDStringCaseConv`.

runtime/compiler/trj9/ilgen/Walker.cpp
```diff
@@ -5215,32 +5215,27 @@ break
          }
       }
 
-   if (!comp()->getOption(TR_DisableSIMDStringCaseConv))
+   bool doSIMDStringCaseConv = cg()->getSupportsSIMDStringCaseConv() && !comp()->getOption(TR_DisableSIMDStringCaseConv);
```
fjeremic: Similarly to below comment about `!comp()->getOption(TR_DisableSIMDStringCaseConv)`.

runtime/compiler/trj9/ilgen/IlGenerator.cpp
```diff
@@ -1484,10 +1484,9 @@ TR_J9ByteCodeIlGenerator::genHWOptimizedStrProcessingAvailable()
    if (constToLoad == -1)
       {
       bool disable = comp()->getOption(TR_DisableSIMDStringCaseConv);
-      bool platformSupported = TR::Compiler->target.cpu.isZ();
-      bool vecInstrAvailable = cg()->getSupportsVectorRegisters();
+      bool platformSupported = cg()->getSupportsSIMDStringCaseConv();
```
fjeremic: This variable has exactly one occurrence. Can we inline its use? Similarly to below we can get rid of the `!disable` check and fold that variable away.

runtime/compiler/trj9/env/VMJ9.cpp
```diff
@@ -3110,7 +3110,7 @@ bool TR_J9VMBase::supressInliningRecognizedInitialCallee(TR_CallSite* callsite,
          case TR::java_lang_String_toLowerHWOptimizedDecompressed:
          case TR::java_lang_String_toUpperHWOptimized:
          case TR::java_lang_String_toLowerHWOptimized:
-            if(comp->cg()->getSupportsVectorRegisters() && !comp->getOption(TR_DisableSIMDStringCaseConv))
+            if(comp->cg()->getSupportsSIMDStringCaseConv() && !comp->getOption(TR_DisableSIMDStringCaseConv))
```
fjeremic: Similar to below.

runtime/compiler/trj9/codegen/J9CodeGenerator.hpp
```diff
@@ -382,7 +386,8 @@ class OMR_EXTENSIBLE CodeGenerator : public OMR::CodeGeneratorConnector
    enum // Flags
       {
       HasFixedFrameC_CallingConvention = 0x00000001,
-      SupportsMaxPrecisionMilliTime    = 0x00000002
+      SupportsMaxPrecisionMilliTime    = 0x00000002,
+      SupportsSIMDStringCaseConv       = 0x00000004
```
fjeremic: Can we rename this to something more generic. You don't have to use SIMD here. You can also "vectorize" the operation in general purpose registers. Could we rename to perhaps `SupportsInlineStringCaseConversion` and add a Doxygen comment on what exactly this flag does?

We should then appropriately rename the getter and setter a few lines above.

runtime/compiler/trj9/codegen/J9CodeGenerator.hpp
```diff
@@ -370,6 +370,10 @@ class OMR_EXTENSIBLE CodeGenerator : public OMR::CodeGeneratorConnector
    bool getSupportsMaxPrecisionMilliTime() {return _j9Flags.testAny(SupportsMaxPrecisionMilliTime);}
    void setSupportsMaxPrecisionMilliTime() {_j9Flags.set(SupportsMaxPrecisionMilliTime);}
 
+   // Determines whether the platform supports vectorized version of java/lang/String.toUpperCase() and toLowerCase()
+   bool getSupportsSIMDStringCaseConv() { return _j9Flags.testAny(SupportsSIMDStringCaseConv);}
+   void setSupportsSIMDStringCaseConv() { return _j9Flags.set(SupportsSIMDStringCaseConv);}
```
fjeremic: Can we add proper Doxygen comments for this so it gets parsed properly?

cathyzhyi:
@fjeremic Thanks for the comments. I've changed accordingly.

fjeremic:
> @fjeremic Thanks for the comments. I've changed accordingly.

Thanks! Code looks much simpler and easier to understand now.

Jenkins test sanity

Issue #1112, opened 2018-02-06
Test failure not reported


pshipton:
Extended testing is failing due to #1110, however I don't see anything in the final report that indicates what failed. 

For example https://ci.eclipse.org/openj9/job/PullRequest-Extended-JDK8-linux_390-64_cmprssptrs-OpenJ9/3/consoleFull

```
19:48:07 [Test] 1..84
19:48:07 [Test] ok 1 - jit_hw_0
19:48:07 [Test] ok 2 - jit_hw_1
19:48:07 [Test] ok 3 - jit_hw_2
19:48:07 [Test] ok 4 - cmdLineTester_vmRuntimeState_0
19:48:07 [Test] ok 5 - cmdLineTester_lockWordAlignment_Object_Standard_SE80_0
19:48:07 [Test] ok 6 - cmdLineTester_LazyClassLoading_0
19:48:07 [Test] ok 7 - cmdLineTest_stackSizeInfoTest_0
19:48:07 [Test] ok 8 - cmdLineTester_verbosetest_0
19:48:07 [Test] ok 9 - cmdLineTester_verbosetest_3
19:48:07 [Test] ok 10 - cmdLineTester_verbosetest_6
19:48:07 [Test] ok 11 - cmdLineTester_javaAssertions_0
19:48:07 [Test] ok 12 - cmdLineTester_softmxCmdOpt_2
19:48:07 [Test] ok 13 - cmdLineTester_PageAlignDirectMemory_0
19:48:07 [Test] ok 14 - cmdLineTester_getCallerClassTests_SE80_0
19:48:07 [Test] ok 15 - testSCCMLTests2_1
19:48:07 [Test] ok 16 - testSCCMLTests3_1
19:48:07 [Test] ok 17 - testSCCMLTests4_1
19:48:07 [Test] ok 18 - testSCCMLTests5_1
19:48:07 [Test] ok 19 - testSCCMLTests6_1
19:48:07 [Test] ok 20 - testSCCMLSnapshot_1
19:48:07 [Test] ok 21 - testSCCMLSoftmx_1
19:48:07 [Test] ok 22 - testSCCMLAotMethodOperation_1
19:48:07 [Test] ok 23 - TestGCClassWithStaticRetransformInBalanced_SE80_0
19:48:07 [Test] ok 24 - ThreadRegressionTests_4
19:48:07 [Test] ok 25 - jsr292Test_SE80_0
19:48:07 [Test] ok 26 - jsr292Test_SE80_1
[...]

PR #1120, opened 2018-02-06
Support different diagnostic level in tap file.

sophia-guo:
Support different diagnostic level in tap file. User can decide what
diagnostic information is needed. Values for DIAGNOSTICLEVEL are 'failure', 'all', 'nodetails'.
Also removed unused top level make variables

[ci skip]

### DIFF
--- a/test/README.md
+++ b/test/README.md
@@ -281,7 +281,21 @@ disabled.spec.<spec> (e.g. disabled.spec.linux_x86-64)
     or were skipped, along with other information like execution time and
     error messages, exceptions and logs from the individual test methods.
 
+  * TAP result files
+
+    TAP files are created. Depending on the requirement there are three
+    different diagnostic levels.
+    
+		* export DIAGNOSTICLEVEL=failure
+			Default level. Log all detailed failure information if test fails
+		* export DIAGNOSTICLEVEL=all
+			Log all detailed information no matter test failures or succeed
+		* export DIAGNOSTICLEVEL=nodetails
+			No need to log any detailed information. Top level TAP test result
+			summary is enough 
+			
   * in the cloud (WIP)
+  
 
 7. Attach a debugger:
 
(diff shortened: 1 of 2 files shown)


smlambert:
What are the other possible values that one may set DIAGNOSTICLEVEL at?  
Is there a related update to the README?  
Is this to address #1112 ?

sophia-guo:
No, this doesn't address #1112. Will update README.

smlambert:
Thank you for the update, I now believe I know what this change is for.  :)

For background to others, the default setting of "failure" allows us to put a little extra detail into the otherwise extremely terse TAP (Test Anything Protocol - https://testanything.org/) output (its default is a 'no details' approach).  We grab the details of the test failure (stack trace and/or console output) and present it as part of the TAP result.  

We decided to use TAP as the "lowest common denominator / LCD" format for test output, because we run several different types of tests using different test frameworks, some of which do not use a standard format for their output.  By generating TAP output, we enforce an LCD so that we can present the results of all these varied tests and test frameworks under one common approach.

Your change allows a user to switch from the default setting of 'failure' to one of the other 2 choices by  exporting DIAGNOSTICLEVEL, if they want more or less detail produced.  This may be desirable, if someone is running tests locally and suddenly have 100s of failures, and want to rerun, but not scroll through reams of failure details (especially if they are all failing for the same reason and developer is trial'n'erroring some changes for a fix).

test/README.md
```diff
@@ -281,7 +281,21 @@ disabled.spec.<spec> (e.g. disabled.spec.linux_x86-64)
     or were skipped, along with other information like execution time and
     error messages, exceptions and logs from the individual test methods.
 
+  * tap result files
```
smlambert: Please change to TAP (Test Anything Protocol) result files  (and change any subsequent use of tap to TAP as its an acronym).

sophia-guo:
@smlambert Thanks! Updated tap to TAP.

PR #1122, opened 2018-02-07
Update acquireVMAccess to release JNI Access before halt check

fengxue-IS:
A deadlock could be caused if thread holding JNI Critical access is
halted during acquireVMAccess by HALTED_AT_SafePoint without
releasing JNI access and then blocked by exclusiveVMAccess_HALT
which has higher priority. In this case, the thread requesting exclusive
access will be locked waiting for JNI Critical access to be released

### DIFF
--- a/runtime/vm/VMAccess.cpp
+++ b/runtime/vm/VMAccess.cpp
@@ -301,31 +301,31 @@ void   internalAcquireVMAccessNoMutexWithMask(J9VMThread * vmThread, UDATA haltM
 	}
 	Assert_VM_mustNotHaveVMAccess(vmThread);
 
-	if((vmThread->publicFlags & (J9_PUBLIC_FLAGS_JNI_CRITICAL_ACCESS | J9_PUBLIC_FLAGS_HALT_THREAD_EXCLUSIVE)) == (J9_PUBLIC_FLAGS_JNI_CRITICAL_ACCESS | J9_PUBLIC_FLAGS_HALT_THREAD_EXCLUSIVE)) {
+	if(J9_ARE_ALL_BITS_SET(vmThread->publicFlags, J9_PUBLIC_FLAGS_JNI_CRITICAL_ACCESS)) {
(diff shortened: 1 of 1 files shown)


fengxue-IS:
800 Grinders all passed

runtime/vm/VMAccess.cpp
```diff
@@ -301,31 +301,33 @@ void   internalAcquireVMAccessNoMutexWithMask(J9VMThread * vmThread, UDATA haltM
 	}
 	Assert_VM_mustNotHaveVMAccess(vmThread);
 
-	if((vmThread->publicFlags & (J9_PUBLIC_FLAGS_JNI_CRITICAL_ACCESS | J9_PUBLIC_FLAGS_HALT_THREAD_EXCLUSIVE)) == (J9_PUBLIC_FLAGS_JNI_CRITICAL_ACCESS | J9_PUBLIC_FLAGS_HALT_THREAD_EXCLUSIVE)) {
+	if((vmThread->publicFlags & J9_PUBLIC_FLAGS_JNI_CRITICAL_ACCESS) == J9_PUBLIC_FLAGS_JNI_CRITICAL_ACCESS) {
 		/* In a critical region and about to block acquiring VM access. */
 		U_64 timeNow;
```
DanHeidinga: This is a c++ file - can you move the definition of timeNow to its first use?

gacholio: Also, since this line is being changed, it would be good to use the J9_ARE_ANY_BITS_SET macro (or ALL_BITS if you want to be pedantic).

gacholio: re: Dan's comment, I believe he means to declare and initialize the variable at the same time.

DanHeidinga:
@fengxue-IS Great work on this.  Can you update the description to explain the root cause?

fengxue-IS:
Updated description, fixed definition and bit check as suggested

gacholio:
jenkins test sanity plinux

PR #1137, opened 2018-02-08
Enable final field reporter, check class init status before reporting

fengxue-IS:
The class initStatus check is added since final fields are allowed
to be modified during initialization

### DIFF
--- a/runtime/oti/VMHelpers.hpp
+++ b/runtime/oti/VMHelpers.hpp
@@ -1373,8 +1373,11 @@ exit:
 	static VMINLINE void
 	reportFinalFieldModified(J9VMThread* currentThread, J9Class* fieldClass)
 	{
-#if 0
-		if (J9_ARE_NO_BITS_SET(fieldClass->classFlags, J9ClassHasIllegalFinalFieldModifications)) {
+		/** Only report modifications after class initalization
+		 *  Since final field write is allowed during class init process,
+		 *  JIT will not start to trust final field values until the class has completed initialization
+		 */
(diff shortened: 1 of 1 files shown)


gacholio:
jenkins test sanity plinux

gacholio:
Please add a brief description to the commit (and PR).

runtime/oti/VMHelpers.hpp
```diff
@@ -1373,8 +1373,7 @@ class VM_VMHelpers
 	static VMINLINE void
 	reportFinalFieldModified(J9VMThread* currentThread, J9Class* fieldClass)
 	{
-#if 0
-		if (J9_ARE_NO_BITS_SET(fieldClass->classFlags, J9ClassHasIllegalFinalFieldModifications)) {
+		if (J9_ARE_NO_BITS_SET(fieldClass->classFlags, J9ClassHasIllegalFinalFieldModifications) && (J9ClassInitSucceeded == fieldClass->initializeStatus)) {
```
DanHeidinga: Please add a comment to describe why this is only reported after the class is initialized.  It makes it obvious that to the next person reading / debugging this.

gacholio:
Tracepoint is problematic, since this in a header file, in no particular module.

gacholio:
I suppose we should get the JIT to emit the tracepoint.

runtime/oti/VMHelpers.hpp
fengxue-IS: added

Issue #1114, opened 2018-02-06
Archive test step fails intermittently when run in parallel with testing


AdamBrousseau:
During the PR builds, once the compile is complete, we archive the SDK in parallel with the testing. Sometimes, the SDK fails to archive ([eg](https://ci.eclipse.org/openj9/blue/organizations/jenkins/PullRequest-Sanity-JDK9-linux_ppc-64_cmprssptrs_le-OpenJ9/detail/PullRequest-Sanity-JDK9-linux_ppc-64_cmprssptrs_le-OpenJ9/22/pipeline)). We shouldn't fail a PR build if this step fails (currently), however, this step needs to pass for a nightly build or an OMR Acceptance build. We could try/catch the archive and make failing optional, but I think its probably safe enough, and cleaner, to retry an archive rather than make optional whether it is critical it passes or not.

We could add a retry(3) block around the [archive step](https://github.com/eclipse/openj9/blob/20aeaf1fffb3bb330600afe484959717276b5bfa/buildenv/jenkins/build#L220).

pshipton:
Ideally we should fix the underlying cause so the archive doesn't fail. Do you know why the archive fails? Disk space, job timeout, machine overloaded? If its out of disk space then would retrying help?

AdamBrousseau:
Failed during tar of the test directory. Probably shouldn't be taring a directory while test stuff is being written to it. Before the pipelines were committed, we only archived the sdk in a PR. Now that the code is shared, we archive the test material in a PR unnecessarily. I can think of three options off the top of my head
1. Break this out into 2 separate archive steps
1. Wrap the test archive in some kind of if block
1. Put an exclude in the tar to ignore certain dir(s) where we write test results.

AdamBrousseau:
More options to add to my previous list

4. Don't archive in PR builds at all
   - I don't know who how often these are used

5. Stop archiving in parallel with testing
   - Typical times to archive:
      - JDK8/zLinux: 4.5 Min
      - JDK9/zLinux: 9 min
      - JDK8/pLinux: <1 min
      - JDK9/pLinux: 2 min

pshipton:
I've been downloading a few PR builds lately. 

6. It doesn't seem to fail very often. Archive in parallel with testing, but if it fails print a message to explain that it failed in case somebody ever looks, and don't fail the testing.

AdamBrousseau:
(Implemented # 2)

PR #1141, opened 2018-02-08
Enable OPENJ9_BUILD at makefile top level

JasonFengJ9:
Enable `OPENJ9_BUILD` at `makefile` top level

Added `vendor_version.h`;
Removed `openj9_version.h`;
Removed individual `OPENJ9_BUILD` defines; 
Removed IBM vendor defines;
Refactored `openj9_version.c` and `openj9_version.h` to avoid multiple similar header files.

Reviewer @pshipton 
FYI: @DanHeidinga 


pshipton:
A couple files contain the following, but since #1095 j9cfg.h.ftl does not. They should all be consistent.

```
#if defined(OPENJ9_BUILD)
/*The file collects version information in OpenJ9 build */
#include "openj9_version_info.h"
#endif /* OPENJ9_BUILD */
```

JasonFengJ9:
Added `OPENJ9_BUILD` decoration for `#include "openj9_version_info.h"`.
Pls have another look.

pshipton:
I wonder if we should remove the `#ifdef OPENJ9_BUILD` for all the `#include "openj9_version_info.h"`. Technically it is only there "temporarily" to support IBM builds, and shouldn't be used at all. 
I'd like to get @DanHeidinga opinions on this PR.

runtime/include/j9cfg.h.ftl
```diff
@@ -33,7 +33,9 @@ extern "C" {
 #endif
 
 #include "omrcfg.h"
+#if defined(OPENJ9_BUILD)
```
DanHeidinga: if `openj9_version_info.h` is expected to always exist, then we shouldn't be guarding its inclusion with the `OPENJ9_BUILD` define.

I think this should be included unconditionally.

JasonFengJ9: Working on a PR to include `openj9_version_info.h` within `vendor_version.h` which is included unconditionally.

JasonFengJ9: PR has been updated, pls have another look.

pshipton:
jenkins test all plinux

pshipton:
Extended - JDK8 testing passed, the job failed due to #1114

PR #1142, opened 2018-02-08
Implement -XjniAcc for all platforms

fjeremic:
- Deprecate current Power code
- Move logic setting various FastJNI flags to a common location

### DIFF
--- a/runtime/compiler/trj9/env/j9method.cpp
+++ b/runtime/compiler/trj9/env/j9method.cpp
@@ -24,6 +24,7 @@
 
 #include <stddef.h>
 #include "bcnames.h"
+#include "fastJNI.h"
 #include "j9.h"
 #include "j9cfg.h"
 #include "jitprotos.h"
@@ -46,6 +47,7 @@
 #include "il/DataTypes.hpp"
 #include "il/Node.hpp"
 #include "il/Node_inlines.hpp"
+#include "infra/SimpleRegex.hpp"
 #include "optimizer/Inliner.hpp"
 #include "optimizer/PreExistence.hpp"
 #include "runtime/MethodMetaData.h"
@@ -2218,6 +2220,17 @@ TR_ResolvedJ9Method::TR_ResolvedJ9Method(TR_OpaqueMethodBlock * aMethod, TR_Fron
       // a non-NULL target address is returned for both JNI natives and non-JNI natives with FastJNI replacements, NULL otherwise
       // properties will be non-zero IFF the target address points to a FastJNI replacement
       _jniTargetAddress = j9fe->getJ9JITConfig()->javaVM->internalVMFunctions->jniNativeMethodProperties(j9fe->vmThread(), _ramMethod, &_jniProperties);
+
+	  // Check for user specified FastJNI override
+	  if (TR::Options::getJniAccelerator() != NULL && TR::SimpleRegex::match(TR::Options::getJniAccelerator(), signature(trMemory)))
+         {
+		 _jniProperties |= 
+             J9_FAST_JNI_RETAIN_VM_ACCESS | 
+             J9_FAST_JNI_NOT_GC_POINT |
+             J9_FAST_NO_NATIVE_METHOD_FRAME |
+             J9_FAST_JNI_NO_EXCEPTION_THROW |
+             J9_FAST_JNI_NO_SPECIAL_TEAR_DOWN;
+         }
       }
    else
       {
--- a/runtime/compiler/trj9/p/codegen/PPCJNILinkage.cpp
+++ b/runtime/compiler/trj9/p/codegen/PPCJNILinkage.cpp
@@ -108,7 +108,6 @@ TR::Register *TR::PPCJNILinkage::buildDirectDispatch(TR::Node *callNode)
    bool wrapRefs;
    bool passReceiver;
    bool passThread;
-   bool jniAccelerated;
    uintptrj_t targetAddress;
 
    bool crc32m1 = (callSymbol->getRecognizedMethod() == TR::java_util_zip_CRC32_update);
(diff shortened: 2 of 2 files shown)


runtime/compiler/trj9/env/j9method.cpp
```diff
@@ -2218,6 +2220,17 @@ TR_ResolvedJ9Method::TR_ResolvedJ9Method(TR_OpaqueMethodBlock * aMethod, TR_Fron
       // a non-NULL target address is returned for both JNI natives and non-JNI natives with FastJNI replacements, NULL otherwise
       // properties will be non-zero IFF the target address points to a FastJNI replacement
       _jniTargetAddress = j9fe->getJ9JITConfig()->javaVM->internalVMFunctions->jniNativeMethodProperties(j9fe->vmThread(), _ramMethod, &_jniProperties);
+
+	  // Check for user specified FastJNI override
+	  if (TR::Options::getJniAccelerator() != NULL && TR::SimpleRegex::match(TR::Options::getJniAccelerator(), signature(trMemory)))
```
fjeremic: @dsouzai what is the easiest way for me to use stack allocated memory here? The `signature` call here will allocate on the heap at the moment I believe but the value is only used temporarily within this scope. It should be eligible for stack allocation.

runtime/compiler/trj9/p/codegen/PPCJNILinkage.cpp
```diff
@@ -168,16 +166,10 @@ TR::Register *TR::PPCJNILinkage::buildDirectDispatch(TR::Node *callNode)
...
       targetAddress = (uintptrj_t)callSymbol->getMethodAddress();
       }
 
-   if (!isGPUHelper && (TR::Options::getJniAccelerator() != NULL))
-      {
-      jniAccelerated = TR::SimpleRegex::match(TR::Options::getJniAccelerator(), resolvedMethod->signature(comp()->trMemory()));
-      }
-
-   if (!isGPUHelper && (callSymbol->isPureFunction() || resolvedMethodSymbol->canDirectNativeCall() || jniAccelerated || specialCaseJNI))
+   if (!isGPUHelper && (callSymbol->isPureFunction() || resolvedMethodSymbol->canDirectNativeCall() || specialCaseJNI))
```
fjeremic: @zl-wang there are 3 checks for `wrapRefs`, `passReceiver`, `passThread`. Are these not meant to be controlled by FastJNI?

runtime/compiler/trj9/env/j9method.cpp
dsouzai: There's two ways to do stack allocation. One is to use `signature(trMemory, TR_AllocationKind::stackAlloc)` which will allocate memory in the current stack region. However, if the current stack region at this point is the outermost stack region, then that memory won't get freed until the TR_Memory object goes out of scope.

The other way is to create a new stack region by doing
```
{
...
TR::StackMemoryRegion stackMemoryRegion(*trMemory);
... signature(trMemory, TR_AllocationKind::stackAlloc);
}
```
which will definitely free up the memory when it goes out of scope, but the downside is having to construct the `TR::StackMemoryRegion` object which also involves allocating a 4K segment of memory on the C stack which might result in a stack overflow (though this is very unlikely).

fjeremic: In your expert opinion which one should I use here?

runtime/compiler/trj9/p/codegen/PPCJNILinkage.cpp
zl-wang: @fjeremic  VM-side recognized a set of FastJNI methods.  One of the checks reflects that control. But there are other cases VM doesn't know, e.g., Math.log() etc.   jniAcc is one of the cases as well.

runtime/compiler/trj9/env/j9method.cpp
dsouzai: It's somewhat subjective, but I would take the first approach. If the current stack region isn't the outermost region, then the memory will get released when that region goes out of scope; if it is, then it's no worse than what's happening now. I can see the benefit of creating a new stack region namely that since it allocates 4K on the C Stack a small allocation like would be relatively quick, but initializing a new region for a one time small allocation feels like overkill (again this is subjective). I'm ok with either approach though.

I'll have to think about what we can do to facilitate small temporary allocations.

fjeremic: Fixed in https://github.com/eclipse/openj9/pull/1142/commits/0bf38d59883b7e9428d6bf624a2e5c1a87302cc0. Thanks for your suggestions!

fjeremic:
This should be ready for review. @zl-wang could I ask you to be the committer for this PR given your experience in this area and as we also have Power changes here as well?

Jenkins test sanity
Jenkins test extended

fjeremic:
Jenkins test sanity

zl-wang:
sure, i will watch over this.

fjeremic:
> Do we know the proportion of ResolvedJ9Method(s) eventually setting up its _fullSignature field? If the answer is "majority", I would suggest not using stackAlloc for perf reasons.

I don't think this matters at all really. The call to `signature` is guarded by a check for `TR::Options::getJniAccelerator() != NULL` which is `NULL` by default unless the user specifies the `-XjniAcc` option which is unsafe anyway and should only be used for very extreme cases or for testing. I think we are dwelling into micro optimization territory here.

dsouzai:
In the current compiler memory manager implementation, using stackAlloc or heapAlloc takes the same amount of time; creating a new stack region has initialization costs but when using an existing stack region, it is equivalent to heapAlloc in terms of perf. This is because when doing heapAlloc, we're actually allocating from a region that exists for the lifetime of the compilation. If we're using stackAlloc from an existing region, then the cost of destroying that region already exists.

zl-wang:
@fjeremic @dsouzai  I was not referring to the speed of memory alloc/free, but memory consumption.  only if it is heapAlloc, the signature is retained in the symbol.  if majority symbols need signature anyway, you are spending 2x of memory for signature in addition to the cost of retrieving the signature.

fjeremic:
> @fjeremic @dsouzai I was not referring to the speed of memory alloc/free, but memory consumption. only if it is heapAlloc, the signature is retained in the symbol. if majority symbols need signature anyway, you are spending 2x of memory for signature in addition to the cost of retrieving the signature.

Fixed in https://github.com/eclipse/openj9/pull/1142/commits/8c9c6a03bdda2f14b7980c498197991e09d0ce8a

Jenkins test all

zl-wang:
ppc64le sanity failure was a false alarm due to #1114

PR #1145, opened 2018-02-09
Scan RS concurrently

amicic:
This is just a preparation change for (it's not dependent on):
eclipse/omr#2302

- move the pre-processing pass for Unfinalized from RS Scavneger
RootScanner wrapper to main roots wrapper. This is cleaner, and more
importantly needed for Concurrent Scavenger to avoid calling this
pre-processing pass twice, since CS will have two RS root scan
passes

- report TGC scan roots stats for both STW pauses of a CS cycle, by
hooking into to local-end instead of cycle-end event

### DIFF
--- a/runtime/gc_glue_java/ScavengerRootScanner.cpp
+++ b/runtime/gc_glue_java/ScavengerRootScanner.cpp
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2015, 2017 IBM Corp. and others
+ * Copyright (c) 2015, 2018 IBM Corp. and others
  *
  * This program and the accompanying materials are made available under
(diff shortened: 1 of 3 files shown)


dmitripivkine:
Jenkins test sanity depends eclipse/omr#2302

amicic:
Since Travis check was failing due to missing dependent changes, I removed parts of this change that were dependent on https://github.com/eclipse/omr/pull/2302. Will put them later as a separate PR.

Now, this is a stand alone change - @dmitripivkine feel free to merge at any point.

https://github.com/eclipse/omr/pull/2302 is still dependent on this change, and I will keep WIP prefix in its subject, until this one is merged.

amicic:
Jenkins test sanity

PR #1148, opened 2018-02-09
Prevent register spilling inside internal control flow

klangman:
We were not preventing the register assigner from assigning
R0 in the evaluator that inlines Atomic operations like
AtomicInteger.getAndDecrement(). Some virtual registers
are used in addi and stwcx which can't use R0 forcing
the register assigner to spill inside the internal control
flow generated by the evaluator if R0 was assigned.


klangman:
@zl-wang Can you review this change?

zl-wang:

zl-wang:

PR #1154, opened 2018-02-12
Add Export for JVM_VERSION

Mesbah-Alam:
Signed-off-by: Mesbah <Mesbah_Alam@ca.ibm.com>


Mesbah-Alam:
Related Issue: https://github.com/AdoptOpenJDK/openjdk-tests/issues/251

Mesbah-Alam:
@sophia-guo .. Could you please review the PR?

test/TestConfig/settings.mk
```diff
@@ -58,6 +58,12 @@ else
 export JAVA_VERSION:=$(JAVA_VERSION)
 endif
 
+ifndef JVM_VERSION
```
sophia-guo: Please do not hard code `JVM_VERSION:=openjdk8`.  Talked with @llxia  ifndef JVM_VERSION we should construct value by JAVA_VERSION and JAVA_IMPL.
Extra code would be like following:
```
ifndef JAVA_IMPL
export JAVA_IMPL:=openj9
endif

ifeq ($(JAVA_VERSION), SE80)
	JDK_VERSION = openjdk8
else
	ifeq ($(JAVA_VERSION), SE90)
		JDK_VERSION = openjdk9
	else
		ifeq ($(JAVA_VERSION), SE100)
			JDK_VERSION = openjdk10
		endif
	endif
endif

ifneq (, $(findstring openj9, $(JAVA_IMPL)))
	JVM_VERSION = $(JDK_VERSION)-openj9
else
	JVM_VERSION = $(JDK_VERSION)
endif

```

test/TestConfig/settings.mk
```diff
@@ -58,6 +58,12 @@ else
...
+export JVM_VERSION:=openjdk8
+else
+export JVM_VERSION:=$(JVM_VERSION)
```
sophia-guo: I suspect we do not need this else export.

test/TestConfig/settings.mk
Mesbah-Alam: Can you explain why we need to "construct" JVM_VERSION using JAVA_IMPL and and JAVA_VERSION? I am just trying to understand :)) 

The way I have it in this PR seems to be working perfectly. For example, in the job below, the JVM_VERSION provided while launching the Jenkins job is 'openjdk8-openj9', and it gets passed down into the actual third party application test build script as expected:  e.g.: https://ci.adoptopenjdk.net/view/work%20in%20progress/job/test_personal/223/console

sophia-guo: test_personal are using Jenkinsfile_personal, which export JVM_VERSION explicitly. JenkinsfileBase doesn't do the export JVM_VERSION explicitly. That is if you run a build with JAVA_VERSION=SE90,  JVM_VERSION will be set to openjdk8. This can be solved by export JVM_VERSION explicitly in JenkinsfileBase file.

However export JVM_VERSION is not a must. Say running locally and user set JAVA_VERSION=SE90, hard code  JVM_VERSION=openjdk8 would cause a failure.

llxia:
Background information: 
JVM_VERSION is a variable that is needed in multiple test components (currently regression test and container test) in AdoptOpenJDK. Setting JVM_VERSION in a common place, so that test component can use it directly.

test/TestConfig/settings.mk
Mesbah-Alam: Thanks for the explanation. Makes sense to me now! I've made the change in the PR. 

Validated the change with personal build: https://ci.adoptopenjdk.net/view/work%20in%20progress/job/test_personal/230/console

test/TestConfig/settings.mk
```diff
@@ -58,6 +58,28 @@ else
 export JAVA_VERSION:=$(JAVA_VERSION)
 endif
 
+ifndef JAVA_IMPL
```
sophia-guo: I would prefer there is `ifndef JVM_VERSION` block. That is saying user can also define his own JVM_VERSION.

sophia-guo:
Could you please also add echo JAVA_IMPL and JVM_VERSION ins setup_% target? Thanks.

test/TestConfig/settings.mk
Mesbah-Alam: Done!

Mesbah-Alam:
> Could you please also add echo JAVA_IMPL and JVM_VERSION ins setup_% target? Thanks.

- Done.

PR #1157, opened 2018-02-12
Guard pointer dereference with VM Access

dsouzai:


dsouzai:
WIP as I haven't run a pbuild yet.

@andrewcraik could you review / verify?

dsouzai:
pBuild looks good, removing WIP.

runtime/compiler/trj9/optimizer/J9TransformUtil.cpp
```diff
@@ -830,26 +831,28 @@ J9::TransformUtil::transformIndirectLoad(TR::Compilation *comp, TR::Node *node)
...
-                     case TR::Int64:
-                        node = TR::Node::lconst(node, *(int64_t*)(structure + node->getSymbolReference()->getOffset()));
-                        break;
-                     default:
-                        TR_ASSERT(0, "Unexpected native field type %s", node->getDataType().toString());
-                        break;
+                     if (secondDereference)
+                        structure = (uint8_t*)dereferenceStructPointerChain(structure, firstDereference, node->getFirstChild(), comp);
+
+                     if (verifyFieldAccess(structure, node->getSymbolReference(), comp)) switch (loadType)
```
mstoodle: would you clean up this switch statement while you're in here, please?

dsouzai:
@mstoodle cleaned up the switch statement :)

jdmpapin:
All loads between `baseObject` and `node` should result in pointers to "unrepresentable" native structs, because they were skipped [here](https://github.com/eclipse/openj9/blob/9c6d60f70ee45aa785d31583ae5bad73078ca17c/runtime/compiler/trj9/optimizer/J9TransformUtil.cpp#L691-L696). Since those aren't moveable,

1. there are no moveable objects in the chain when `baseObject` is a class; and
2. if `baseObject` is an object, it's the only one, so only `firstDereference` should need VM access.

There might be an issue when performing `firstDereference` though, where `fieldAddress` could be stale if the base object were moved before acquiring `digPointerOutOfObject`.  It looks like `firstDereference` should always be address-typed as well.

dsouzai:
After discussing with Devin, we agreed that VM Access isn't needed before the calls to `dereferenceStructPointerChain` since every dereference in the chain (with the potential exception of `objectBase`) is safe. However, as Devin noted, we need to move the calculation of the `fieldAddress` to a scope guarded with VMAccess.

@jdmpapin @andrewcraik Could you guys please review.

dsouzai:
@andrewcraik Could you please review.

andrewcraik:
looks good
Jenkins test sanity

dsouzai:
```
Setting status of 0fb3a007dd16e3da86e5692548fc968c11b89c69 to FAILURE with url https://ci.eclipse.org/openj9/job/PullRequest-Sanity-JDK9-linux_ppc-64_cmprssptrs_le-OpenJ9/86/ and message: 'Build finished. 21963 tests run, 1992 skipped, 0 failed.'
Using context: Sanity - JDK9 - linux_ppc-64_cmprssptrs_le
ERROR: script returned exit code 1
Finished: FAILURE
```

0 jobs failed but the build still failed; not sure what to make of that :S.

pshipton:
Its a known problem https://github.com/eclipse/openj9/issues/1114

dsouzai:
@andrewcraik Could you please merge this, or perhaps delegate this to another committer?

andrewcraik:
Just missed it in my scan - merging now.

PR #1158, opened 2018-02-12
Updates method handles to use newer java.lang.Class API

taliamccormick:
Earlier draft for nestmates changes to java.lang.Class defined new
method:
> public boolean isNestmateOf(Class<?> c)

Newer draft for nestmates changes to java.lang.Class defines new method:
> public boolean isInSameNest(Class<?> c)

This commit updates method handles to make use of the newer
java.lang.Class method definition.


tajila:
looks good, @pshipton please review

DanHeidinga:
Not running a pull request build as the Nest Mates tests aren't run in the builds (yet).

PR #1162, opened 2018-02-12
Add generic test target

renfeiw:
Generic target includes all subtests with different variations.

### DIFF
--- a/test/TestConfig/scripts/testKitGen/makeGenTool/mkgen.pl
+++ b/test/TestConfig/scripts/testKitGen/makeGenTool/mkgen.pl
@@ -303,6 +303,7 @@ sub writeTargets {
 	}
 	foreach my $test ( @{ $result->{'tests'} } ) {
 		my $count     = 0;
+		my @subtests = ();
 		foreach my $var ( @{ $test->{'variation'} } ) {
 			my $jvmoptions = ' ' . $var . ' ';
 			$jvmoptions =~ s/\ NoOptions\ //x;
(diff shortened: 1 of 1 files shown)


renfeiw:
reviewers: @llxia @TianyuZuo

llxia:
Jenkins test sanity

smlambert:
To be clear about what this PR does, in a playlist.xml file, when there is more than 1 variation defined (the test should be run in more than 1 mode or set of commandline options), this PR creates a test target that will run all of them.

For example, if my playlist contained something like this:
		<testCaseName>jit_hw</testCaseName>
		<variations>
			<variation>-Xdump:java -Xgcpolicy:gencon -Xjit:count=0,optLevel=hot</variation>
			<variation>-Xdump:java -Xgcpolicy:optavgpause -Xjit:count=0</variation>
			<variation>-Xdump:java -Xjit:enableHCR,enableOSR,count=0</variation>
		</variations> ...

And I wanted to run this test, I would run jit_hw_0 (which would execute the test with the gencon option, and jit_hw__1, which would execute it with the optavgpause option, and jit_hw_2 would use the enableHCR option). Now if I run jit_hw, it will run the test 3x, first with gencon, then with optavgpause option and finally with enableHCR.

Correct?

renfeiw:
Thanks @smlambert for clarifying the content of this PR.

renfeiw:
@fjeremic thanks, updated commit message and PR comment.

smlambert:
Jenkins test sanity plinux

smlambert:
JDK8 sanity test failure is a separate issue being worked, will proceed with merging this PR.

PR #1163, opened 2018-02-12
Add stub classes to support Java 11 builds

### LINKED ISSUE #1153: JDK next:  Compile fails due to missing class java.lang.invoke.MethodHandleNatives
Trying to compile the latest openjdk with openj9 results in the following error:

```Creating support/modules_libs/java.base/libjsig.so from J9 sources

ERROR: Build failed for target 'all' in configuration 'linux-x86_64-normal-server-release' (exit code 2) 

=== Output from failing command(s) repeated here ===
* For target jdk_modules_java.base__the.java.base_batch:
/root/openJDK10-J9-mirror-update-test/src/java.base/share/classes/java/lang/invoke/BootstrapMethodInvoker.java:34: error: cannot access MethodHandleNatives
import static java.lang.invoke.MethodHandleNatives.*;
                              ^
  bad source file: /root/openJDK10-J9-mirror-update-test/build/linux-x86_64-normal-server-release/support/j9jcl_sources/java.base/share/classes/java/lang/invoke/MethodHandleNatives.java
    file does not contain class java.lang.invoke.MethodHandleNatives
    Please remove or make sure it appears in the correct subdirectory of the sourcepath.
1 error

* All command lines available in /root/openJDK10-J9-mirror-update-test/build/linux-x86_64-normal-server-release/make-support/failure-logs.
=== End of repeated output ===```

This is because there is a new class src/java.base/share/classes/java/lang/invoke/BootstrapMethodInvoker.java that imports java.lang.invoke.MethodHandleNatives.* which openj9 doesn't implement (there is an empty file with just the copyright statement)

### PR DESCRIPTION
**Pull request by JasonFengJ9**
Add stub classes to support `Java 11` builds

* Added pConfig JAVA11
* Added stub class `java.lang.invoke.BootstrapMethodInvoker`;
* Added stub class `java.lang.invoke.MethodHandleNatives`;
* Added stub method `java.lang.invoke.MemberName.isFinal()`.

Closes #1153 

With this PR, `java -version` output is as following:
```
Exception in thread "(unnamed thread)" java/lang/ExceptionInInitializerError
java/lang/NullPointerException
```
This is same as raw build result discovered earlier which appears due to `NULL helpers` within `java.lang.String`
```  
private static final com.ibm.jit.JITHelpers helpers = com.ibm.jit.JITHelpers.getHelpers();
```
This `NPE` is still under investigation but not related to the compilation error reported in the issue specified.

Reviewer @pshipton 
FYI: @DanHeidinga 

### DIFF
--- a/jcl/src/java.base/share/classes/java/lang/invoke/BootstrapMethodInvoker.java
+++ b/jcl/src/java.base/share/classes/java/lang/invoke/BootstrapMethodInvoker.java
@@ -1,7 +1,7 @@
-/*[INCLUDE-IF Sidecar18-SE-OpenJ9|Sidecar19-SE-B174]*/
+/*[INCLUDE-IF Java11]*/
 
 /*******************************************************************************
- * Copyright (c) 2017, 2018 IBM Corp. and others
+ * Copyright (c) 2018, 2018 IBM Corp. and others
  *
  * This program and the accompanying materials are made available under
  * the terms of the Eclipse Public License 2.0 which accompanies this
@@ -21,19 +21,15 @@
  *
  * SPDX-License-Identifier: EPL-2.0 OR Apache-2.0 OR GPL-2.0 WITH Classpath-exception-2.0 OR LicenseRef-GPL-2.0 WITH Assembly-exception
  *******************************************************************************/
-/*[IF Java11]*/
-
 package java.lang.invoke;
 
 /*
  * Stub class for compilation
  */
 
-class MethodHandleNatives {
+class BootstrapMethodInvoker {
 
-	static LinkageError mapLookupExceptionToError(ReflectiveOperationException roe) {
+	static <T> T invoke(Class<T> clz1, MethodHandle mh, String str, Object obj1, Object obj2, Class<?> clz2) {
 		throw OpenJDKCompileStub.OpenJDKCompileStubThrowError();
 	}
 }
-
-/*[ENDIF]*/
--- a/jcl/src/java.base/share/classes/java/lang/invoke/MemberName.java
+++ b/jcl/src/java.base/share/classes/java/lang/invoke/MemberName.java
@@ -1,7 +1,7 @@
 /*[INCLUDE-IF Sidecar18-SE-OpenJ9]*/
 
 /*******************************************************************************
- * Copyright (c) 2017, 2017 IBM Corp. and others
+ * Copyright (c) 2017, 2018 IBM Corp. and others
  *
  * This program and the accompanying materials are made available under
  * the terms of the Eclipse Public License 2.0 which accompanies this
@@ -70,5 +70,10 @@ final class MemberName {
 	public Class<?> getDeclaringClass() {
 		throw OpenJDKCompileStub.OpenJDKCompileStubThrowError();
 	}
+	/*[IF Java11]*/
+	public boolean isFinal() {
+		throw OpenJDKCompileStub.OpenJDKCompileStubThrowError();
+	}
+	/*[ENDIF]*/
 	/*[ENDIF]*/
 }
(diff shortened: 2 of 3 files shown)

### CODE REVIEW

File: jcl/src/java.base/share/classes/java/lang/invoke/BootstrapMethodInvoker.java [hunk 1]
```diff
@@ -0,0 +1,35 @@
+/*[INCLUDE-IF Java10]*/
+
+/*******************************************************************************
+ * Copyright (c) 2017, 2018 IBM Corp. and others
```
pshipton: A new file should not have a start date of 2017.

jcl/src/java.base/share/classes/java/lang/invoke/BootstrapMethodInvoker.java
```diff
@@ -0,0 +1,35 @@
+/*[INCLUDE-IF Java10]*/
```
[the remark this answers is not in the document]
pshipton: One argument is that Java 10 is only supported for 6 months, and will be replaced by Java 11. In the future the tags will be folded so only Java 11 is supported and its not important to distinguish between 10 and 11. 

In response:
1) We don't know what will happen in the future. Its not hard to add another configuration for Java 11, that will likely be needed anyway in order to have both Java 10 and 11 builds working at the same time.
2) Making the changes Java 11 specific ensure that Java 10 cannot be broken by the changes (i.e. was Java 10 tested to ensure these changes didn't affect it?)

jcl/src/java.base/share/classes/java/lang/invoke/MemberName.java
```diff
@@ -70,5 +70,8 @@ String getMethodDescriptor() {
 	public Class<?> getDeclaringClass() {
 		throw OpenJDKCompileStub.OpenJDKCompileStubThrowError();
 	}
+	public boolean isFinal() {
```
DanHeidinga: Should this be wrapped with `Java11` preprocessor tags?

jcl/src/java.base/share/classes/java/lang/invoke/MethodHandleNatives.java
```diff
@@ -21,3 +21,19 @@
  *
  * SPDX-License-Identifier: EPL-2.0 OR Apache-2.0 OR GPL-2.0 WITH Classpath-exception-2.0 OR LicenseRef-GPL-2.0 WITH Assembly-exception
  *******************************************************************************/
+/*[IF Java10]*/
```
DanHeidinga: `10` -> `11`

jcl/src/java.base/share/classes/java/lang/invoke/BootstrapMethodInvoker.java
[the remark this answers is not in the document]
JasonFengJ9: Agreed. It is easier to remove `Java 10` than add `Java 11` later.

jcl/src/java.base/share/classes/java/lang/invoke/BootstrapMethodInvoker.java
JasonFengJ9: Fixed.

jcl/src/java.base/share/classes/java/lang/invoke/MemberName.java
JasonFengJ9: Added `Java11` flag.

jcl/src/java.base/share/classes/java/lang/invoke/MethodHandleNatives.java
JasonFengJ9: Fixed.

pshipton:
jenkins compile plinux jdk9

pshipton:
Travis build failed as it exceeded the time limit.

PR #1167, opened 2018-02-13
Check CP consistency during method resolve

fengxue-IS:
Add new CP tags to include methodref info from classfile
(CFR_CONSTANT_Methodref / CFR_CONSTANT_InterfaceMethodref)
- updated test code from asm-5.0 to asm-6.0 and use updated 
   function `visitMethodInsn`
- remove invalid testcases
- update testcase to expect IncompatibleClassChangeError and
   class redefine failures


fengxue-IS:
Reviewer: @DanHeidinga 
Test Reviewer: @TianyuZuo

fengxue-IS:
@TianyuZuo 
Updated commit message to provide more information

Two SharedCPEntryInvokerTest are removed because the content they are testing is no longer valid in the new spec, they used to test that cp entry shared by `invokeinterface/invokespecial/invokestatic`  pointing to non-interface method will raise exception during `invokeinterface` while allowing `invokespecial/invokestatic` to succeed
After adding the CP consistency check, all three invoker should fail check if the CP entry is interface methodref (which is already covered by other testcases) or cause static verifier to fail if the CP entry is methodref (also covered by the modified testcases)

@DanHeidinga One thing we should consider is if this check should also be used in Java 8, where the behavior is not specified in the spec. (This doesn't affect classfile generated by javac, only incorrect classfile from third party tools)

TianyuZuo:
Thanks for updating the commit message and providing those info.

runtime/bcutil/ConstantPoolMap.cpp
```diff
@@ -392,7 +401,7 @@ ConstantPoolMap::findVarHandleMethodRefs()
 	U_16 *varHandleMethodTable = NULL;
 
 	for (U_16 i = 1; i < _romConstantPoolCount; i++) {
-		if  (J9CPTYPE_INSTANCE_METHOD == _romConstantPoolTypes[i]) {
+		if  (J9CPTYPE_INSTANCE_METHOD == _romConstantPoolTypes[i] || J9CPTYPE_INTERFACE_INSTANCE_METHOD == _romConstantPoolTypes[i]) {
```
DanHeidinga: formatting:
`((J9CPTYPE_INSTANCE_METHOD == _romConstantPoolTypes[i]) || (J9CPTYPE_INTERFACE_INSTANCE_METHOD == _romConstantPoolTypes[i]))`

runtime/bcutil/ConstantPoolMap.cpp
```diff
@@ -455,11 +464,13 @@ ConstantPoolMap::constantPoolDo(ConstantPoolVisitor *visitor)
 			case J9CPTYPE_ANNOTATION_UTF8:
 				visitor->visitString(cfrCPIndex);
 				break;
-			case J9CPTYPE_FIELD: /* fall through */
+			case J9CPTYPE_FIELD:  /* fall through */
```
DanHeidinga: accidental change?

runtime/jcl/common/jclcinit.c
```diff
@@ -196,7 +196,7 @@ static jint initializeStaticMethod(J9JavaVM* vm, UDATA offset)
 	J9ROMClass * jclROMClass = jclConstantPool->ramClass->romClass;
 	U_32 * cpShapeDescription = J9ROMCLASS_CPSHAPEDESCRIPTION(jclROMClass);
 
-	if (J9CPTYPE_STATIC_METHOD == J9_CP_TYPE(cpShapeDescription, offset)) {
+	if (J9CPTYPE_STATIC_METHOD == J9_CP_TYPE(cpShapeDescription, offset) || J9CPTYPE_INTERFACE_STATIC_METHOD == J9_CP_TYPE(cpShapeDescription, offset)) {
```
DanHeidinga: Formatting:
`((J9CPTYPE_STATIC_METHOD == J9_CP_TYPE(cpShapeDescription, offset)) || (J9CPTYPE_INTERFACE_STATIC_METHOD == J9_CP_TYPE(cpShapeDescription, offset)))`

runtime/jcl/common/jclcinit.c
```diff
@@ -305,7 +305,7 @@ jint initializeKnownClasses(J9JavaVM* vm, U_32 runtimeFlags)
 					return JNI_ERR;
 				}
 			}
-		} else if (J9CPTYPE_INSTANCE_METHOD == J9_CP_TYPE(cpShapeDescription, i)) {
+		} else if (J9CPTYPE_INSTANCE_METHOD == J9_CP_TYPE(cpShapeDescription, i) || J9CPTYPE_INTERFACE_INSTANCE_METHOD == J9_CP_TYPE(cpShapeDescription, i)) {
```
DanHeidinga: Both `J9CPTYPE_INTERFACE_INSTANCE_METHOD` & `J9CPTYPE_INTERFACE_STATIC_METHOD` should never happen for the VM's constantpool.  We should have both of these treated as an error if they ever occurred.

For background, the VM generates a constantpool of data - references to class and {static, instance}x{fields, methods} and should never generate this kind of incorrect data.

runtime/vm/resolvesupport.cpp
```diff
@@ -451,6 +451,26 @@ resolveStaticMethodRefInto(J9VMThread *vmStruct, J9ConstantPool *ramCP, UDATA cp
 	} else {
 		cpClass = J9_CLASS_FROM_CP(ramCP);
 		lookupOptions |= J9_LOOK_CLCONSTRAINTS;
+
+		if (cpClass != NULL && cpClass->romClass != NULL) {
```
DanHeidinga: formatting: `((cpClass != NULL) && (cpClass->romClass != NULL))`

runtime/vm/resolvesupport.cpp
```diff
@@ -451,6 +451,26 @@ resolveStaticMethodRefInto(J9VMThread *vmStruct, J9ConstantPool *ramCP, UDATA cp
...
+			if (isResolvedClassAnInterface) {
+				if ((J9CPTYPE_INTERFACE_STATIC_METHOD != J9_CP_TYPE(J9ROMCLASS_CPSHAPEDESCRIPTION(cpClass->romClass), cpIndex)) &&
+					(J9CPTYPE_INTERFACE_INSTANCE_METHOD != J9_CP_TYPE(J9ROMCLASS_CPSHAPEDESCRIPTION(cpClass->romClass), cpIndex)) &&
+					(J9CPTYPE_INTERFACE_METHOD != J9_CP_TYPE(J9ROMCLASS_CPSHAPEDESCRIPTION(cpClass->romClass), cpIndex))) {
```
DanHeidinga: formatting - put the closing `) {` on their own line indented to the same level as the `if`

runtime/oti/j9nonbuilder.h
```diff
@@ -2084,6 +2084,8 @@ typedef struct J9ConstantPool {
 #define J9CPTYPE_METHOD_TYPE  13
 #define J9CPTYPE_METHODHANDLE  14
 #define J9CPTYPE_ANNOTATION_UTF8  15
+#define J9CPTYPE_INTERFACE_STATIC_METHOD 16
```
DanHeidinga: These need a comment to describe when they will be used and why they need to be different than the existing static / instance method cp types.

Both of these tags will always result in an error at resolve - we should call that out as well.

runtime/jvmti/jvmtiClass.c
```diff
@@ -1779,9 +1779,13 @@ jvmtiGetConstantPool_translateCP(J9PortLibrary *privatePortLibrary, jvmtiGcp_tra
 			case J9CPTYPE_INSTANCE_METHOD:
 			case J9CPTYPE_STATIC_METHOD:
 			case J9CPTYPE_INTERFACE_METHOD:
+			case J9CPTYPE_INTERFACE_INSTANCE_METHOD:
+			case J9CPTYPE_INTERFACE_STATIC_METHOD:
 
 				rc = jvmtiGetConstantPool_addReference(translation, cpIndex,
-								       (U_8) ((cpItemType == J9CPTYPE_INTERFACE_METHOD) ? CFR_CONSTANT_InterfaceMethodref : CFR_CONSTANT_Methodref),
+								       (U_8) (((cpItemType == J9CPTYPE_INTERFACE_METHOD) ||
```
DanHeidinga: This is too much code for an inline if statement.  Move it a local variable.

runtime/vm/resolvesupport.cpp
DanHeidinga: Pull `J9_CP_TYPE(J9ROMCLASS_CPSHAPEDESCRIPTION(cpClass->romClass), cpIndex)` into a local variable.

test/Jsr335_interfaceStaticMethod/src/org/openj9/test/jsr335/interfaceStaticMethod/InvokeStaticTest.java
```diff
@@ -69,13 +69,13 @@ public void test_B_bar() {
...
-	// test invokestatic A.bar() should fail with NoSuchMethodError public
+	// test invokestatic A.bar() should fail with IncompatibleClassChangeError
 	@Test
 	public void test_A_bar() {
 		try {
 			String s = org.openj9.test.jsr335.interfaceStaticMethod.GenInvokeStatic.test_A_bar();
-			Assert.fail("noSuchMethodErrorThrown not thrown");
-		} catch (NoSuchMethodError e) {
+			Assert.fail("IncompatibleClassChangeError not thrown");
+		} catch (java.lang.IncompatibleClassChangeError e) {
```
DanHeidinga: `java.lang` doesn't need to be included here

runtime/vm/resolvesupport.cpp
```diff
@@ -1102,6 +1122,29 @@ resolveSpecialMethodRefInto(J9VMThread *vmStruct, J9ConstantPool *ramCP, UDATA c
 	if (jitFlags) {
 		lookupOptions |= J9_LOOK_NO_THROW;
 	}
+
+	if (currentClass != NULL) {
+		if (resolvedClass->romClass != NULL && currentClass->romClass != NULL) {
```
DanHeidinga: Can you clean up the formatting in here as suggested above for all the functions and then I'll take another look.

runtime/oti/j9nonbuilder.h
fengxue-IS: I am not sure what you mean by both tags will always result in error, they are checked at resolve time, which gives `IncompatibleClassChangeError` if the check failed. Otherwise the resolve should succeed. did I miss some check?

runtime/jcl/common/jclcinit.c
fengxue-IS: `J9CPTYPE_INTERFACE_INSTANCE_METHOD & J9CPTYPE_INTERFACE_STATIC_METHOD` are a subset of `J9CPTYPE_INSTANCE_METHOD & J9CPTYPE_STATIC_METHOD` where it have a classfile tag of `CFR_CONSTANT_InterfaceMethodref`, not sure why they are incorrect.

FYI, the new types are added to carry extra data (since we remap the classfile methodrefs) for resolve checks. They are not meant as incorrect.

runtime/oti/j9nonbuilder.h
DanHeidinga: Sorry, you're right.  This is for the case were there is an `invokestatic` targeting an `InterfaceMethodRef` which requires that the resolved class is an interface.  Do I have that right now?

The comment added here should explain **why** these new types are needed.

runtime/jcl/common/jclcinit.c
DanHeidinga: Sorry, I was wrong about the purpose of this change.  Leaving it as is should be fine.

fengxue-IS:
@DanHeidinga I fixed the code formatting issues you pointed out, rebased on current master (xa sanity passed), can I get a re-review? Thanks

runtime/jcl/common/jclcinit.c
DanHeidinga: This has the same formatting concern as above.  Missing `()` on the individual clauses and would be better as 3 lines.

runtime/jcl/common/jclcinit.c
```diff
@@ -331,7 +333,7 @@ jint initializeKnownClasses(J9JavaVM* vm, U_32 runtimeFlags)
 					}
 				}
 			}
-		} else if (J9CPTYPE_STATIC_METHOD == J9_CP_TYPE(cpShapeDescription, i)) {
+		} else if (J9CPTYPE_STATIC_METHOD == J9_CP_TYPE(cpShapeDescription, i) || J9CPTYPE_INTERFACE_STATIC_METHOD == J9_CP_TYPE(cpShapeDescription, i)) {
```
DanHeidinga: Same formatting.

runtime/jcl/common/sun_reflect_ConstantPool.c
```diff
@@ -260,7 +262,7 @@ getMethodAt(JNIEnv *env, jobject constantPoolOop, jint cpIndex, UDATA resolveFla
 
 		if (NULL != methodID) {
 			if (NULL != jlClass) {
-				returnValue = (*env)->ToReflectedMethod(env, jlClass, methodID, J9CPTYPE_STATIC_METHOD == cpType);
+				returnValue = (*env)->ToReflectedMethod(env, jlClass, methodID, (J9CPTYPE_STATIC_METHOD == cpType || J9CPTYPE_INTERFACE_STATIC_METHOD == cpType));
```
DanHeidinga: This is a good candidate for moving into a local:
```
const BOOLEAN isStatic = ((J9CPTYPE_STATIC_METHOD == cpType) || (J9CPTYPE_INTERFACE_STATIC_METHOD == cpType));
```

runtime/jvmti/jvmtiClass.c
```diff
@@ -1784,9 +1785,20 @@ jvmtiGetConstantPool_translateCP(J9PortLibrary *privatePortLibrary, jvmtiGcp_tra
 			case J9CPTYPE_INSTANCE_METHOD:
 			case J9CPTYPE_STATIC_METHOD:
 			case J9CPTYPE_INTERFACE_METHOD:
-
+			case J9CPTYPE_INTERFACE_INSTANCE_METHOD:
+			case J9CPTYPE_INTERFACE_STATIC_METHOD:
+
+				if (((cpItemType == J9CPTYPE_INTERFACE_METHOD) ||
+					(cpItemType == J9CPTYPE_INTERFACE_STATIC_METHOD) ||
+					(cpItemType == J9CPTYPE_INTERFACE_STATIC_METHOD)))
```
DanHeidinga: Should one of these be INSTANCE instead of STATIC twice?

runtime/jvmti/jvmtiClass.c
```diff
@@ -1784,9 +1785,20 @@ jvmtiGetConstantPool_translateCP(J9PortLibrary *privatePortLibrary, jvmtiGcp_tra
...
+				{
```
DanHeidinga: formatting: `) {`

runtime/vm/resolvesupport.cpp
```diff
@@ -455,6 +455,26 @@ resolveStaticMethodRefInto(J9VMThread *vmStruct, J9ConstantPool *ramCP, UDATA cp
 	} else {
 		cpClass = J9_CLASS_FROM_CP(ramCP);
 		lookupOptions |= J9_LOOK_CLCONSTRAINTS;
+
+		if ( (cpClass != NULL) && (cpClass->romClass != NULL) ) {
```
DanHeidinga: formatting: remove the extra spaces between brackets.
```
if ((cpClass != NULL) && (cpClass->romClass != NULL)) {
```

runtime/vm/resolvesupport.cpp
```diff
@@ -455,6 +455,26 @@ resolveStaticMethodRefInto(J9VMThread *vmStruct, J9ConstantPool *ramCP, UDATA cp
...
+			if (isResolvedClassAnInterface) {
+				if ((J9CPTYPE_INTERFACE_STATIC_METHOD != J9_CP_TYPE(J9ROMCLASS_CPSHAPEDESCRIPTION(cpClass->romClass), cpIndex)) &&
```
DanHeidinga: Please pull `J9_CP_TYPE(J9ROMCLASS_CPSHAPEDESCRIPTION(cpClass->romClass), cpIndex)` into a local so the code is easier to read.

runtime/vm/resolvesupport.cpp
```diff
@@ -1237,7 +1280,7 @@ resolveVirtualMethodRefInto(J9VMThread *vmStruct, J9ConstantPool *ramCP, UDATA c
 		 */
 		if ((NULL != cpShapeDescription)
 		&& (resolvedClass == J9VMJAVALANGINVOKEMETHODHANDLE(vm))
-		&& (J9CPTYPE_INSTANCE_METHOD != J9_CP_TYPE(cpShapeDescription, cpIndex))
+		&& (J9CPTYPE_INSTANCE_METHOD != J9_CP_TYPE(cpShapeDescription, cpIndex) && J9CPTYPE_INTERFACE_INSTANCE_METHOD != J9_CP_TYPE(cpShapeDescription, cpIndex))
```
DanHeidinga: Can you add the extra brackets around the subclauses?

runtime/jvmti/jvmtiClass.c
```diff
@@ -1784,9 +1785,20 @@ jvmtiGetConstantPool_translateCP(J9PortLibrary *privatePortLibrary, jvmtiGcp_tra
 			case J9CPTYPE_INSTANCE_METHOD:
 			case J9CPTYPE_STATIC_METHOD:
 			case J9CPTYPE_INTERFACE_METHOD:
-
+			case J9CPTYPE_INTERFACE_INSTANCE_METHOD:
+			case J9CPTYPE_INTERFACE_STATIC_METHOD:
```
DanHeidinga: Can you define `U_8 cpType;` in this `case` block?  We can add `{` & `}` to possible to declare here.

 It keeps the scope of the variable small.

runtime/oti/j9nonbuilder.h
```diff
@@ -2125,6 +2125,13 @@ typedef struct J9ConstantPool {
 #define J9CPTYPE_METHODHANDLE  14
 #define J9CPTYPE_ANNOTATION_UTF8  15
 
+/* 
+ * INTERFACE_STATIC and INTERFACE_INSTANT cpType are used to track the classfile tag of a methodref [CFR_CONSTANT_InterfaceMethodref] 
+ * This will be used to check for cp consistency during resolve time
```
DanHeidinga: `We need them to support the correct initializer for the constantpool due to the way we pre-initialize the cp entries`

Can you add a comment like ^^^^ to the comment here?  It makes it clear why we don't just use `J9CPTYPE_INTERFACE`

fengxue-IS:
moved `U_8 cpType` to switch case scope, added comment
I ll fix the NLS message in another PR

DanHeidinga:
Jenkins test sanity plinux

DanHeidinga:
Pushed a commit with some formatting changes so we can get this built and merged

PR #1176, opened 2018-02-14
Fix comment for getProcessCpuTimeImpl() to specify 1 ns units

### LINKED ISSUE #707: return value comment for getProcessCpuTimeImpl is out of date
In openj9/jcl/src/com.ibm.management/share/classes/com/ibm/lang/management/internal/ExtendedOperatingSystemMXBeanImpl.java

the return value comment for getProcessCpuTimeImpl() says that it uses units of 100ns, but that is no longer the case.

At about line 308:
`	 @return process cpu ime in 100 ns units`

In my understanding, getProcessCpuTimeImpl() now returns a value in 1ns units by default, and the value uses 100ns units only if system property com.ibm.lang.management.OperatingSystemMXBean.isCpuTime100ns is set to true.

pshipton:
[ci skip]

The comment for
ExtendedOperatingSystemMXBeanImpl.getProcessCpuTimeImpl() wasn't updated
to specify the correct units.

Fixes: #707


pshipton:
Fixed the footer

PR #1178, opened 2018-02-15
Remove some obsolete jcl preprocessor configurations

pshipton:
Removed
  SIDECAR19-SE-B165
  SIDECAR19-SE-B169
  SIDECAR19-SE-B174
  SIDECAR19-SE-B175

Depends on ibmruntimes/openj9-openjdk-jdk9#127

### DIFF
--- a/jcl/src/java.base/share/classes/java/lang/invoke/VarHandleGuards.java
+++ b/jcl/src/java.base/share/classes/java/lang/invoke/VarHandleGuards.java
@@ -1,7 +1,7 @@
-/*[INCLUDE-IF Sidecar19-SE-B174]*/
+/*[INCLUDE-IF Sidecar19-SE]*/
 
 /*******************************************************************************
- * Copyright (c) 2017, 2017 IBM Corp. and others
+ * Copyright (c) 2017, 2018 IBM Corp. and others
  *
  * This program and the accompanying materials are made available under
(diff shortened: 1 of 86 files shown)


pshipton:
jenkins test sanity plinux jdk9 depends ibmruntimes/openj9-openjdk-jdk9#127

pshipton:
The PR testing had passed, but I'll run it again since I rebased.

jenkins test sanity plinux jdk9 depends ibmruntimes/openj9-openjdk-jdk9#127

jcl/src/java.base/share/classes/java/lang/invoke/VarHandleGuards.java
```diff
@@ -1,7 +1,7 @@
-/*[INCLUDE-IF Sidecar19-SE-B174]*/
+/*[INCLUDE-IF Sidecar19-SE]*/
```
DanHeidinga: Shouldn't this be `Sidecar19-SE-OpenJ9`?  Here and below

pshipton: Technically yes, but using Sidecar19-SE doesn't break anything. I expect Sidecar19-SE-OpenJ9 to be replaced with Sidecar19-SE once SIDECAR19-SE-B148 is deleted, so using Sidecar19-SE now saves some changes later.

PR #1182, opened 2018-02-15
Check for arithmetic overflow in JNI string and array functions

pdbain-ibm:
Verify that start + size is a valid non-negative 32-bit number.


pshipton:

PR #1183, opened 2018-02-15
Add methods inherited from superinterfaces to iTables

gacholio:
Modify iTables to contain every method in both the local interface and
any extended interfaces so that the resolved interface class in the RAM
constant pool for interface methods represents the interface that the
receiver class must implement, rather than the class that declared the
resolved interface method.


DanHeidinga:
Jenkins test sanity plinux

PR #1185, opened 2018-02-15
Remove useless stackwalk macro

keithc-ca:
The macro `J9SW_JIT_RECOMPILATION_RESOLVE_OFFSET_TO_SAVED_RECEIVER` is #undef'd on all platforms - remove those #undef's and all references to that name and the related `_DEFINED` ddr generated constant.


keithc-ca:
@JamesKingdon, do you know the answer to Graham's question?

JamesKingdon:
The macro isn't something I'm familiar with, but I had a quick search and I couldn't find anywhere that it's set for arm. Most of the uses appear to be for DDR, which worryingly doesn't seem to have anything for arm.

Not sure of the significance, but method recompilation is working on arm.

keithc-ca:
@gacholio Given that James reports method recompilation is working on arm, I hope you'll agree this should be safe to merge.

PR #1189, opened 2018-02-15
Variable cm should be set after the NULL check of sharedClassConfig

hangshao0:
Currently we set cm using sharedClassConfig->sharedClassCache, then we
check if sharedClassConfig is NULL, which does not make sense. 
cm should be set after the NULL check.


pshipton:
Was there any issue reported for this problem, should it be flagged as a bug?

hangshao0:
Just notice this when looking at the code. No issue reported, it is not a bug.

pshipton:

PR #1191, opened 2018-02-15
Improve GC pause times by removing assumptions in one pass

klangman:
Use a mark and sweep approach to removing runtime assumptions
from the Runtime Assumption Table (RAT) in places where we are
doing assumption clean up during a GC pause time. This will
significantly reduce the GC pause times in cases were we have
a large number of assumptions and the GC is unloading classes.

Signed-off-by: Kevin Langman langman@ca.ibm.com

### DIFF
--- a/runtime/compiler/trj9/env/RuntimeAssumptionTable.hpp
+++ b/runtime/compiler/trj9/env/RuntimeAssumptionTable.hpp
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2000, 2016 IBM Corp. and others
+ * Copyright (c) 2000, 2018 IBM Corp. and others
  *
  * This program and the accompanying materials are made available under
  * the terms of the Eclipse Public License 2.0 which accompanies this
@@ -68,6 +68,7 @@ char const * const runtimeAssumptionKindNames[LastAssumptionKind] =
 struct TR_RatHT
    {
    OMR::RuntimeAssumption ** _htSpineArray;
+   uint32_t              * _markedforDetachCount;
    size_t                  _spineArraySize;
    };
 
@@ -111,15 +112,21 @@ class TR_RuntimeAssumptionTable
    void incReclaimedAssumptionCount(int32_t tableId) { reclaimedAssumptionCount[tableId]++; }
    void detachFromRAT(OMR::RuntimeAssumption *assumption);
 
+   void markAssumptionsAndDetach(void *reclaimedMetaData, bool reclaimPrePrologueAssumptions = false);
+   void reclaimMarkedAssumptionsFromRAT();
+
    int32_t countRatAssumptions();
 
    private:
    friend class OMR::RuntimeAssumption;
    void addAssumption(OMR::RuntimeAssumption *a, TR_RuntimeAssumptionKind kind, TR_FrontEnd *fe, OMR::RuntimeAssumption **sentinel);
    int32_t reclaimAssumptions(void *md, OMR::RuntimeAssumption **hashTable, OMR::RuntimeAssumption **possiblyRelevantHashTable);
+   void markForDetachFromRAT(OMR::RuntimeAssumption *assumption);
 
    // My table of hash tables; init() allocate memory and populate it
    TR_RatHT _tables[LastAssumptionKind];
+   bool     _detachPending[LastAssumptionKind]; // Marks tables that have assumptions waiting to be removed
+   uint32_t _marked;                            // Counts the number of assumptions waiting to be removed
    int32_t assumptionCount[LastAssumptionKind]; // this never gets decremented
    int32_t reclaimedAssumptionCount[LastAssumptionKind];
    };
--- a/runtime/compiler/trj9/runtime/ClassUnloadAssumption.cpp
+++ b/runtime/compiler/trj9/runtime/ClassUnloadAssumption.cpp
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2000, 2017 IBM Corp. and others
+ * Copyright (c) 2000, 2018 IBM Corp. and others
  *
  * This program and the accompanying materials are made available under
  * the terms of the Eclipse Public License 2.0 which accompanies this
@@ -154,10 +154,14 @@ TR_PersistentClassInfo::removeUnloadedSubClasses()
        _tables[i]._spineArraySize = sizes[i];
        size_t storageSize = sizeof(OMR::RuntimeAssumption*)*_tables[i]._spineArraySize;
        _tables[i]._htSpineArray = (OMR::RuntimeAssumption**)TR_PersistentMemory::jitPersistentAlloc(storageSize);
-       if (!_tables[i]._htSpineArray)
+       _tables[i]._markedforDetachCount = (uint32_t*)TR_PersistentMemory::jitPersistentAlloc(sizeof(uint32_t)*_tables[i]._spineArraySize);
+       if (!_tables[i]._htSpineArray || !_tables[i]._markedforDetachCount)
           return false;
        memset(_tables[i]._htSpineArray, 0, storageSize);
+       memset(_tables[i]._markedforDetachCount, 0, sizeof(uint32_t)*_tables[i]._spineArraySize);
        }
+    _marked=0;
+    memset(_detachPending, 0, sizeof(bool)*LastAssumptionKind);
     return true;
     }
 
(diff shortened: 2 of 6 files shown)


andrewcraik:
@klangman your pull request failed the IP validation - it looks like the signature on your commit is missing angle brackets around the email address (<>). Could you please fix this?

runtime/compiler/trj9/runtime/ClassUnloadAssumption.cpp
```diff
@@ -351,6 +354,149 @@ TR_RuntimeAssumptionTable::detachFromRAT(OMR::RuntimeAssumption *assumption)
...
+            numAssumptionsNotReclaimed++; // RAS
+            }
+         }
+
+      if (notReclaimedList != sentry) // some entries were not reclaimed
+         {
+         // Must attach the notReclaimedList to the sentinel
+         sentry->setNextAssumptionForSameJittedBody(notReclaimedList);
+
+#ifdef PROD_WITH_ASSUMES
```
andrewcraik: I think this may want to be #if defined(PROD_WITH_ASSUMES) || defined(DEBUG) since we want this code in either case. Perhaps this should also be available under an option or env var for prod builds to help with bug finding later?

runtime/compiler/trj9/runtime/ClassUnloadAssumption.cpp
```diff
@@ -351,6 +354,149 @@ TR_RuntimeAssumptionTable::detachFromRAT(OMR::RuntimeAssumption *assumption)
...
+            {
+            cursor = sentry->getNextAssumptionForSameJittedBody();
+            while (cursor != sentry)
+               {
+               TR_RuntimeAssumptionKind kind = cursor->getAssumptionKind();
+               if (kind != RuntimeAssumptionOnClassRedefinitionPIC && 
+                   kind != RuntimeAssumptionOnClassRedefinitionUPIC && 
+                   kind != RuntimeAssumptionOnClassRedefinitionNOP)
+                  {
+                  fprintf(stderr, "%d assumptions were left after metadata %p assumption reclaiming\n", numAssumptionsNotReclaimed, metaData);
```
andrewcraik: fprintf to stderr is not very elegant for debug code - this should either use the tracing framework or the information integrated into the assert message.

klangman: This part of the code is a direct copy of the existing code in TR_RuntimeAssumptionTable::reclaimAssumptions() which could be made redundant in the future if we decide that this mark@sweep methods is appropriate for the compile abort cases (and I believe it is).  I left this code as is so that it's behaviour (and output) is identical to what it was before.

andrewcraik: @mpirvu is there a better way that fprintf that should be used here? I'm just not a fan of hardcoded prints to stderr because they defeat so much of the infrastructure around gathering diagnostics and output. If this is the 'right' way then I'll relent.

runtime/compiler/trj9/env/RuntimeAssumptionTable.hpp
```diff
@@ -111,15 +112,21 @@ class TR_RuntimeAssumptionTable
...
    private:
    friend class OMR::RuntimeAssumption;
    void addAssumption(OMR::RuntimeAssumption *a, TR_RuntimeAssumptionKind kind, TR_FrontEnd *fe, OMR::RuntimeAssumption **sentinel);
    int32_t reclaimAssumptions(void *md, OMR::RuntimeAssumption **hashTable, OMR::RuntimeAssumption **possiblyRelevantHashTable);
+   void markForDetachFromRAT(OMR::RuntimeAssumption *assumption);
 
    // My table of hash tables; init() allocate memory and populate it
    TR_RatHT _tables[LastAssumptionKind];
+   bool     _detachPending[LastAssumptionKind]; // Marks tables that have assumptions waiting to be removed
+   uint32_t marked;                             // Counts the number of assumptions waiting to be removed
```
mpirvu: `marked` should have an underscore in front

runtime/compiler/trj9/runtime/ClassUnloadAssumption.cpp
```diff
@@ -351,6 +354,149 @@ TR_RuntimeAssumptionTable::detachFromRAT(OMR::RuntimeAssumption *assumption)
    TR_ASSERT(cursor, "Must find my assumption in rat\n");
    }
 
+/** 
+ * Mark an assumption for future detach and recliaming from the RAT
```
mpirvu: typo `recliaming`

runtime/compiler/trj9/runtime/ClassUnloadAssumption.cpp
```diff
@@ -351,6 +354,149 @@ TR_RuntimeAssumptionTable::detachFromRAT(OMR::RuntimeAssumption *assumption)
...
+void TR_RuntimeAssumptionTable::markForDetachFromRAT(OMR::RuntimeAssumption *assumption)
+   {
+   TR_RatHT *hashTable = findAssumptionHashTable(assumption->getAssumptionKind());
+   _detachPending[assumption->getAssumptionKind()] = true;
+   hashTable->_markedforDetachCount[(assumption->hashCode() % hashTable->_spineArraySize)]++;
+   assumption->markForDetach();
+   }
+
+/**
+ * Traverse the entire RAT detaching and recliaming all marked assumptions.
```
mpirvu: type `recliaming`

runtime/compiler/trj9/env/RuntimeAssumptionTable.hpp
```diff
@@ -111,15 +112,21 @@ class TR_RuntimeAssumptionTable
...
 
    private:
    friend class OMR::RuntimeAssumption;
    void addAssumption(OMR::RuntimeAssumption *a, TR_RuntimeAssumptionKind kind, TR_FrontEnd *fe, OMR::RuntimeAssumption **sentinel);
    int32_t reclaimAssumptions(void *md, OMR::RuntimeAssumption **hashTable, OMR::RuntimeAssumption **possiblyRelevantHashTable);
+   void markForDetachFromRAT(OMR::RuntimeAssumption *assumption);
 
    // My table of hash tables; init() allocate memory and populate it
    TR_RatHT _tables[LastAssumptionKind];
+   bool     _detachPending[LastAssumptionKind]; // Marks tables that have assumptions waiting to be removed
```
mpirvu: Where is `detachPending` initialized?

runtime/compiler/trj9/runtime/ClassUnloadAssumption.cpp
```diff
@@ -351,6 +354,149 @@ TR_RuntimeAssumptionTable::detachFromRAT(OMR::RuntimeAssumption *assumption)
...
+         _detachPending[kind] = false;
+         }
+      }
+   assumptionTableMutex->exit();
+   TR_ASSERT( marked == reclaimed, "Did not reclaime all the marked assumptions!" );
+   marked=0;
+   }
+
+/**
+ * Mark and detach all of the assumptions in the metadata's circular linked list
```
mpirvu: typo `detach`

runtime/compiler/trj9/runtime/ClassUnloadAssumption.cpp
```diff
@@ -351,6 +354,149 @@ TR_RuntimeAssumptionTable::detachFromRAT(OMR::RuntimeAssumption *assumption)
...
+ * @param md The metadata for which to mark & detach assumptions from.
+ * @param reclaimPrePrologueAssumptions Are all assumptions free to be removed (ClassUnloading vs. recompiled)
+ * This only detaches from the metadata's linked list, not a detech from the RAT,
```
mpirvu: typo `detech`

runtime/compiler/trj9/runtime/ClassUnloadAssumption.cpp
```diff
@@ -351,6 +354,149 @@ TR_RuntimeAssumptionTable::detachFromRAT(OMR::RuntimeAssumption *assumption)
...
+   assumptionTableMutex->enter();
+   if (sentry)
+      {
+      int32_t numAssumptionsNotReclaimed = 0; // RAS
+      TR_ASSERT(sentry->getAssumptionKind() == RuntimeAssumptionSentinel, "First assumption must be the sentinel\n");
+      OMR::RuntimeAssumption *notReclaimedList = sentry;
+      for ( cursor=sentry->getNextAssumptionForSameJittedBody(); cursor != sentry; cursor=next)
+         {
+         next = cursor->getNextAssumptionForSameJittedBody();
+         if (!metaData || cursor->isAssumingMethod(metaData, reclaimPrePrologueAssumptions))
```
mpirvu: `metaData` cannot be null here, otherwise we would have crashed when doing `metaData->runtimeAssumptionList`
In the original case metaData could have been null when a method failed compilation and the metadat was not created yet.

runtime/compiler/trj9/runtime/HookHelpers.cpp
```diff
@@ -260,10 +272,12 @@ void jitReleaseCodeCollectMetaData(J9JITConfig *jitConfig, J9VMThread *vmThread,
       cleanUpJitExceptionHandlerCache(vmThread,metaData);
       cleanUpJitArtifactSearchCache(vmThread,metaData);
 
-      // 3rd parameter to reclaimAssumptions determines whether we reclaim preprologue assumptions.
+      // Mark the runtime assumptions so thay can be deleted at the end of the GC cycle.
+      // The 3rd parameter to markAssumptionsAndDetach determines whether we mark preprologue assumptions.
```
mpirvu: Is it the third or the fourth parameter?

runtime/compiler/trj9/runtime/ClassUnloadAssumption.cpp
mpirvu: I am in favor of removing the entire section because we no longer build/test PROD_WITH_ASSUMES

mpirvu:
As I mentioned with another occasion, I would like a doc file describing the design of the algorithm.

runtime/compiler/trj9/runtime/ClassUnloadAssumption.cpp
andrewcraik: I'd prefer to keep it with both PROD_WITH_ASSUMES and DEBUG since it is a useful validation that the code is running right... Some people do manually build prod with assumes still I believe.

mpirvu: ok, then the fprintf could become a TR_ASSERT

runtime/compiler/trj9/env/RuntimeAssumptionTable.hpp
klangman: Nice catch!

runtime/compiler/trj9/runtime/HookHelpers.cpp
klangman: It's the 3rd. 1st = jitConfig, 2nd = metaData, 3rd = (NULL == faintCacheBlock)

runtime/compiler/trj9/runtime/ClassUnloadAssumption.cpp
klangman: I decided to remove the entire #define section and just add an assert into the main loop. Seems easier and cleaner.

mpirvu:
Jenkins test all

PR #1195, opened 2018-02-16
Variable cm should be set after the NULL check of sharedClassConfig

DanHeidinga:
Address more instances of pull request #1189 (use of sharedClassConfig
before the null check) and move variable declarations to first use
when possible.  This is a minimal change to only address
variables close to uses of sharedClassConfig before the null check.


DanHeidinga:
FYI @hangshao0 @pshipton

pshipton:

PR #1201, opened 2018-02-16
Updates valhalla nestmates test to match spec update

taliamccormick:
Valhalla nestmates introduces the concept of a 'nest' of classes which share access between eachothers private members. This nest is defined by a 'nest host' which claims a list of 'nest members' and by the nest members which link back to the nest host.

An earlier spec draft loaded the nest host during class loading. A newer spec puts off nest host loading and verification until required for an access check.  Exceptions and errors that arise within class loading and access checking have therefore changed and the tests must change in order to remain up-to-date.

Current spec draft here: cr.openjdk.java.net/~dlsmith/nestmates.html

### DIFF
--- a/test/README.md
+++ b/test/README.md
@@ -67,6 +67,12 @@ make test
     To get the above dependent jars, please run `make getdependency`.
     getdependency target will get called when running make compile
     or test target.
+    
+    Valhalla tests currently make use of the nestmates capabilities
+    which are available starting in asm version 7. In order to run
+    the Valhalla nestmates on a local machine, you must manually get
+    the asm-7.0.jar and put it in TestConfig/lib/. Avoid using asm 7
+    in other tests as this is a temporary solution.
 
 2. Compile tests:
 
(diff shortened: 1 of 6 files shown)


taliamccormick:
@tajila

taliamccormick:
@tajila @llxia

test/README.md
```diff
@@ -67,6 +67,12 @@ make test
     To get the above dependent jars, please run `make getdependency`.
     getdependency target will get called when running make compile
     or test target.
+    
+    Valhalla tests currently make use of the nestmates capabilities
+    which are available starting in asm version 7. In order to run
+    the Valhalla nestmates on a local machine, you must manually get
+    the asm 7 jar. Avoid using asm 7 in other tests as this is a
```
llxia: how about a little more detail? `you must manually get the asm 7 jar and put it under TestConfig/lib.`

DanHeidinga:
Do these tests pass when run against OpenJDK Project Valhalla builds?

taliamccormick:
I ran it against one from January 10 and all fifteen passed. 

They do not yet pass when run against ours - I'm still moving nest host loading from class loading to the visibility check, so the tests are slightly ahead.

taliamccormick:
Changes to move nest host loading from class loading to the visibility check are here: https://github.com/eclipse/openj9/pull/1224
When compiled & run with those changes enabled, the newer nestmates tests pass.

tajila:
@llxia can you please review this

tajila:
@pshipton  can you take a look

PR #1206, opened 2018-02-18
Make unsafePut64 alignment safe for ARM

JamesKingdon:
unsafePut64 used the strd instruction which raises SIGBUS if the
destination is unaligned when running on a 64 bit OS. To prevent
this the strd was replaced by a pair of str instructions.

Fixes https://github.com/eclipse/openj9/issues/1204

### DIFF
--- a/runtime/vm/xr32/unsafeHelper.s
+++ b/runtime/vm/xr32/unsafeHelper.s
@@ -1,4 +1,4 @@
-@ Copyright (c) 1991, 2017 IBM Corp. and others
+@ Copyright (c) 1991, 2018 IBM Corp. and others
 @
 @ This program and the accompanying materials are made available under
 @ the terms of the Eclipse Public License 2.0 which accompanies this
@@ -24,7 +24,8 @@
 	.global	unsafePut64
 	.type	unsafePut64, %function
 unsafePut64:
-	strd	r2, [r0]
+	str	r2, [r0]
+	str	r3, [r0, #4]
 	bx	lr
 	.size	unsafePut64, .-unsafePut64
 @ Prototype: void unsafePut32(I_32 *address, I_32 value);


runtime/vm/xr32/unsafeHelper.s
```diff
@@ -24,7 +24,8 @@
 	.global	unsafePut64
 	.type	unsafePut64, %function
 unsafePut64:
-	strd	r2, [r0]
+	str	r2, [r0]
+	str	r3, [r0, #4]
```
andrewcraik: I assume the semantics of this API ensures that word tearing can't be an issue from the use of a two stage write? Just wondering about a prempt of the thread between these two operations.

JamesKingdon: I would love to know what the semantics of anything from the unsafe APIs actually are. However, in this particular case, it can be no worse than the original form as ARM doesn't provide atomic writes for non-aligned addresses.

andrewcraik: Ok so I think there is a lurking issue here. The unsafePut64 seems to be a primitive used in the VM to implement unsafe 64bit puts (both reference and primitive). Since this is a 32bit platform this should be safe - more care might be required for 64bit depending on the write operations supported.

JamesKingdon: I wonder if it would be worth specialising the call into aligned and non-aligned variants. It's not worth a test for alignment, but if we knew at the caller that the target was always aligned we could exploit the (marginally) faster encoding. But I think we're talking about saving 1 cycle, so probably not worth the effort.
For sure we'll need to review/rewrite all the assembly helpers when we move to 64bit.

andrewcraik: +1 that it is probably not worth the complexity in this case if the performance difference is 1 cycle and we are on a 32 bit platform so don't have to worry about bit shearing in a 64bit reference.

DanHeidinga:
ARM doc explicitly stating http://infocenter.arm.com/help/index.jsp?topic=/com.arm.doc.faqs/ka15414.html the `strd` instruction doesn't support unaligned access.

I'm not aware of any pattern other than the two `str` instructions that will work with unaligned memory.

DanHeidinga:
As we don't have ARM machines in our CI, I'll merge this without a build.

Issue #1138, opened 2018-02-08
Support "jenkins test all" as a synonym for "jenkins test sanity" and "jenkins test extended"


fjeremic:
Looking at [1] we seem to support launching:

> Current supported PR builds are "Compile", "Compile & Sanity", or "Compile & Extended"

It would be wonderful if we could also support "Compile & Sanity & Extended" via "jenkins test all", or any other flavor of testing such as "functional" which I'm told exists. Our guideline at the moment is to launch both sanity and extended testing for every PR pending ample machine resources. This would simplify launching such builds via a single command.

[1] https://github.com/eclipse/openj9/blob/6fa061ffc2730e3d0b3de1c247581b93b478feae/buildenv/jenkins/README.md

fjeremic:
@AdamBrousseau FYI, is this something within the realm of possibilities?

fjeremic:
It also seems you cannot post:

> Jenkins test sanity
> Jenkins test extended

in the same comment to trigger both builds. They have to be posted in separate comments which creates unnecessary spam. It would definitely be wonderful if we could implement this.

AdamBrousseau:
`Jenkins test all` should work now.
Ex regex
`.*\bjenkins\s+test\s+(all|sanity)\b\s*($|\n|depends\s+.*|(all|([a-z]+,)*zlinux(,[a-z]+)*)\s*($|\n|depends\s+.*|all|(jdk[0-9]+,)*jdk8(,jdk[0-9]+)*)(\s+depends.*)?)`

Have yet to figure out how to fix the multiline problem though. I think if I simply add `.*` to the end it starts to get too hungry. Will have to spend a few minutes to figure it out.

fjeremic:
> Jenkins test all should work now.

Awesome! Thank you very much!

> Have yet to figure out how to fix the multiline problem though. I think if I simply add .* to the end it starts to get too hungry. Will have to spend a few minutes to figure it out.

Sounds like we need some tests here? Perhaps write a few "sample comments" with various combinations and have passing / failing criteria?

fjeremic:
Just tried "Jenkins test all" in https://github.com/eclipse/openj9/pull/875 and it works as envisioned! @AdamBrousseau could I ask you to update https://github.com/eclipse/openj9/buildenv/jenkins/README.md with the new option documentation and behavior? In the PR you can link against closing out this issue.

AdamBrousseau:
I think this was resolved in #1208

PR #1208, opened 2018-02-19
Support 'all' test levels in PR builds

AdamBrousseau:
- PR regex has been updated to support 'all'
- Using 'all' will launch sanity and extended builds
- Can extend to more test levels when they are added

[skip ci]

Issue #1138

### DIFF
--- a/buildenv/jenkins/README.md
+++ b/buildenv/jenkins/README.md
@@ -26,13 +26,17 @@ This folder contains Jenkins pipeline scripts that are used in the OpenJ9 Jenkin
 
 ### Triggering Pull Request Builds from Github
 
-- Current supported PR builds are "Compile", "Compile & Sanity", or "Compile & Extended"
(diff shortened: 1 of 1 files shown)


buildenv/jenkins/README.md
```diff
@@ -41,10 +41,13 @@ This folder contains Jenkins pipeline scripts that are used in the OpenJ9 Jenkin
 - Request a Sanity build on zLinux all java versions
```
fjeremic: We should also change line 29:

>  - Current supported PR builds are "Compile", "Compile & Sanity", or "Compile & Extended"

AdamBrousseau: That statement is still true. I think we need another sentence that specifically calls out this feature.

AdamBrousseau:
Commit amended @fjeremic

PR #1211, opened 2018-02-19
Disable legacy system initialization code

### LINKED ISSUE #1210: System.java refers to non-existent class com.ibm.utils.SystemIntialization
static void completeInitialization() {
...
	try {
		/*[PR 167854] - Please note the incorrect name of the class is correct - somewhere the spelling was wrong and we just need to use the incorrect spelling */
		/*[IF Sidecar19-SE]*/
		systemInitialization = Class.forName("com.ibm.utils.SystemIntialization"); //$NON-NLS-1$
		/*[ELSE]*/
		systemInitialization = Class.forName("com.ibm.misc.SystemIntialization"); //$NON-NLS-1$
		/*[ENDIF]*/
	} catch (ClassNotFoundException e) {
		// Assume this is a raw configuration and suppress the exception
	} catch (Exception e) {
		throw new InternalError(e.toString());
	} 	

This class does not exist in the OpenJ9 Java 8 or Java 9 builds.
We do a futile class lookup and throw an unnecessary exception.

pdbain-ibm:
Remove references to "com.ibm.*.SystemIntialization" (sic) and assorted
variables in OpenJ9 builds. These classes are not included therein.

Add comments to relevant JPP directives for clarity.

Fixes #1210 

### DIFF
--- a/jcl/src/java.base/share/classes/java/lang/System.java
+++ b/jcl/src/java.base/share/classes/java/lang/System.java
@@ -2,7 +2,7 @@
 package java.lang;
 
 /*******************************************************************************
- * Copyright (c) 1998, 2017 IBM Corp. and others
+ * Copyright (c) 1998, 2018 IBM Corp. and others
  *
  * This program and the accompanying materials are made available under
  * the terms of the Eclipse Public License 2.0 which accompanies this
@@ -203,6 +203,7 @@ native static void startSNMPAgent();
 	
 static void completeInitialization() {
 	/*[IF !Sidecar19-SE_RAWPLUSJ9]*/	
+	/*[IF !Sidecar18-SE-OpenJ9]*/
 	Class<?> systemInitialization = null;
 	Method hook;
 	try {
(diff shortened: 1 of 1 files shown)


jcl/src/java.base/share/classes/java/lang/System.java
```diff
@@ -226,18 +227,20 @@ static void completeInitialization() {
...
+	/*[ENDIF]*/ //Sidecar18-SE-OpenJ9|Sidecar19-SE
+	/*[ENDIF] */ //!Sidecar19-SE_RAWPLUSJ9
 		
 	/*[PR 102344] call Terminator.setup() after Thread init */
 	Terminator.setup();
 	
-	/*[IF !Sidecar19-SE_RAWPLUSJ9]*/
+	/*[IF !Sidecar19-SE_RAWPLUSJ9&!Sidecar18-SE-OpenJ9]*/
 	try {
 		/*[PR 111936] Give Hursley hooks into JCL startup */
```
pshipton: If you are willing, please remove this line, and the same copy above, since these comments aren't appropriate (meaningless) for OpenJ9.

pdbain-ibm: No problem.  Glad to do it.

pshipton:
jenkins test sanity plinux

PR #1217, opened 2018-02-20
Remove remaining Valhalla MVT code

eric98zhang:
Sections relating to MVT Method Handles are removed.

The _isClassValueCapable field of ClassFileOracle is removed since value
classes in L-World are denoted using the ACC_VALUE_TYPE instead of the
@ValueCapableClass annotation.


eric98zhang:
@tajila

pshipton:
CI testing passed. Skipping PR testing as it doesn't cover Valhalla code.

Issue #378, opened 2017-10-18
Enable OMR's DDR support for OpenJ9


DanHeidinga:
The OMR project has been implementing Direct Dump Reader (DDR) support on top of libdwarf.  

OpenJ9 needs to enable DDR support and ensure that the functionality in the OMR project is sufficient to provide the rich debug & introspection capabilities J9 has historically supported.

OpenJ9 issues are difficult to debug without this support.

OMR links
----------
https://github.com/eclipse/omr/blob/f4394252018f04074cb375cbbd7f1f00f41a106c/doc/ddr.md

https://github.com/eclipse/omr/blob/f4394252018f04074cb375cbbd7f1f00f41a106c/doc/ddr/DDR.md

https://github.com/eclipse/omr/pull/1222

DanHeidinga:
This task is to drive the enablement of DDR across the OpenJ9 platforms.

hangshao0:
I have an impression DDR is short for "Direct Dump Reader" :-)

deesebas:
Yeah I think it's Direct Dump Reader too.  And [this ](https://github.com/eclipse/omr/blob/f4394252018f04074cb375cbbd7f1f00f41a106c/doc/ddr.md)page title calls it "DDR: The Data Dump Reader"! :)

keithc-ca:
https://github.com/eclipse/openj9/blob/master/runtime/ddr/ddr.readme says:
```
DDR (Direct Dump Reader)
```

DanHeidinga:
Updated the description to use the correct term.

keithc-ca:
#432 allows blob stubs to compile cleanly if DDR is enabled (but generated blob tables are not available).

keithc-ca:
#461 updates the docker image descriptions in anticipation of the requirement for libdwarf-dev.

keithc-ca:
Generated DDR code is incomplete.

The hand-written code to support DDR should be able to make references to historical constants and features so that debugging core files produced by older JVMs is possible: the fact that #495 was needed is a sign that this is broken.

The set of types, etc. should include everything mentioned in `DDR_Artifacts/supersets/29/superset.ibuild.dat`.

The change for #495 should be reverted, catching `NoSuchFieldError` when that constant is not available at runtime.

DanHeidinga:
@keithc-ca I know from talking with you (and watching the PRs) that you've made a lot of progress on this.  Can you provide a brief update on state of DDR?  What's done and what work still remains?

[...]

PR #1221, opened 2018-02-20
Relocate DDR test code

keithc-ca:
Simplifies exclusion from main builds.

Related eclipse launch configurations are adjusted.

Issue: #378


keithc-ca:
Copyright failures are for *.launch files which *could* have embedded notices, but those would disappear the next time eclipse made any changes.

pshipton:
@llxia or @smlambert is this change ok with you?

smlambert:
I am not clear on the purpose of this PR, are you intending these tests to be run with the other tests?  

But these are not in test dir (where they'd automatically be picked up as part of testing), rather in debugtools dir, what is the story for how and when these tests will be run?  

As part of our test migration, we have been converting JUnit tests to testng, using the testng logger for any System.out.println statements, renaming test packages that were previously called com.ibm.test.xxxx to org.openj9.test.xxxx, playlist.xml and testng.xml files created and so forth.

keithc-ca:
I didn't change any of the test code, it has simply moved from the main source folder to the 'testsrc' folder of the DDR_VM project. I don't know the original intent of the tests so I don't know whether it makes sense to migrate them to the main 'test' folder.

llxia:
I am not aware these tests. Do we run them at all?

keithc-ca:
I don't think they are run automatically. Would people prefer that we simply remove these tests instead of just relocating them?

smlambert:
Do you consider them valuable for testing the DDR feature you have been reworking?

If so, I think they should be migrated to the test dir (as done with other tests) into a new project for all of the DDR tests you wish to include to test the DDR feature.

keithc-ca:
I haven't really spent any time trying to figure out what those tests are about. They depend upon org.junit and things like that don't belong in j9ddr.jar. I just needed to avoid trying to build them along with the classes that do belong in j9ddr.jar and avoid the need to have junit.jar on the classpath. I could have explicitly excluded them where they were, but relocating them seemed like the better solution.

smlambert:
Here is the plan to move forward on this:  
1) approve and merge this change, to unblock the remaining DDR dev work
2) assess these DDR tests and determine if they should be retained (issue #1250)
3) if we should keep these tests, migrate them to the test dir (convert to testng, use logger, etc)

smlambert:
As these tests will not be executed, will forego a PR build.

PR #1222, opened 2018-02-20
Remove useless #undef directives

keithc-ca:
The only definitions of EsVersionMajor and EsVersionMinor are in this header file: there's no value in the `#undef` immediately preceding them.


DanHeidinga:
@keithc-ca Does OMR define similar defines?

keithc-ca:
Those macros are not defined or used in OMR, nor could I find anything similar in OMR.

pshipton:

PR #1223, opened 2018-02-20
Export access to DDR for b165+

keithc-ca:
The existing code which reflectively exports access to DDR was
written prior to the refactoring which moved java.lang.reflect.Module
to java.lang.Module.

This code doesn't have the advantage of being preprocessed and so
can't make distinctions at that time. Instead, this version attempts
to locate Module in the new package first and then falls back to
looking in the old package.

Issue: #378

### DIFF
--- a/jcl/src/openj9.dtfj/share/classes/com/ibm/dtfj/image/j9/ImageFactory.java
+++ b/jcl/src/openj9.dtfj/share/classes/com/ibm/dtfj/image/j9/ImageFactory.java
@@ -1,6 +1,6 @@
 /*[INCLUDE-IF Sidecar18-SE]*/
 /*******************************************************************************
- * Copyright (c) 2004, 2017 IBM Corp. and others
+ * Copyright (c) 2004, 2018 IBM Corp. and others
  *
  * This program and the accompanying materials are made available under
  * the terms of the Eclipse Public License 2.0 which accompanies this
@@ -34,6 +34,7 @@ import java.net.URLClassLoader;
 import java.util.ArrayList;
 import java.util.Iterator;
 import java.util.List;
+import java.util.Objects;
 import java.util.logging.Level;
 import java.util.logging.Logger;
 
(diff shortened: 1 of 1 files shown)


jcl/src/openj9.dtfj/share/classes/com/ibm/dtfj/image/j9/ImageFactory.java
```diff
@@ -52,39 +53,53 @@
...
+			Class<?> moduleClass;
+
+			try {
+				moduleClass = Class.forName("java.lang.Module"); //$NON-NLS-1$
+			} catch (ClassNotFoundException e) {
+				// fallback to name before b165
+				moduleClass = Class.forName("java.lang.reflect.Module"); //$NON-NLS-1$
+			}
+
+			Method exportToAll = modulesClass.getDeclaredMethod("addExportsToAllUnnamed", //$NON-NLS-1$
```
pshipton: I think the original variable name was more appropriate, why change it?

jcl/src/openj9.dtfj/share/classes/com/ibm/dtfj/image/j9/ImageFactory.java
```diff
@@ -93,56 +108,57 @@
...
 	public ImageFactory() {
-		//when the vm shuts down we need to make sure that all our extracted files are deleted
+		// when the vm shuts down we need to make sure that all our extracted files are deleted
 	}
 
+	@Override
 	public Image[] getImagesFromArchive(File archive, boolean extract) throws IOException {
-		if (archive != null && !FileManager.fileExists(archive)) {
-			throw new FileNotFoundException("Archive '" + archive.getAbsolutePath() + "' not found.");
+		Objects.requireNonNull(archive);
```
pshipton: Should change the javadoc to mention NPE.

jcl/src/openj9.dtfj/share/classes/com/ibm/dtfj/image/j9/ImageFactory.java
```diff
@@ -172,79 +188,81 @@ public ImageFactory() {
 	 * @return an instance of Image (null if no image can be constructed from the given file)
 	 * @throws IOException
 	 */
+	@Override
 	public Image getImage(File imageFile) throws IOException {
+		Objects.requireNonNull(imageFile);
```
pshipton: javadoc should mention NPE

jcl/src/openj9.dtfj/share/classes/com/ibm/dtfj/image/j9/ImageFactory.java
```diff
@@ -309,13 +329,15 @@ private File getTempDirParent() {
 	 * This is left as a separate implementation to provide a route by which legacy behaviour
 	 * can be invoked first as all other Image creation mechanisms will try DDR first.
 	 */
+	@Override
 	public Image getImage(File imageFile, File metadata) throws IOException {
+		Objects.requireNonNull(imageFile);
```
pshipton: javadoc should mention NPE

jcl/src/openj9.dtfj/share/classes/com/ibm/dtfj/image/j9/ImageFactory.java
keithc-ca: While the javadoc (for the interface) does not explicitly say null is *not* permitted, it is my understanding that one should assume null is not acceptable unless the documentation explicitly says it *is* permitted.

The old version threw NPEs under the same circumstances, it's just clearer now.

jcl/src/openj9.dtfj/share/classes/com/ibm/dtfj/image/j9/ImageFactory.java
keithc-ca: I suggest the new name is just as descriptive and uses much less screen space making it easier to read.

jcl/src/openj9.dtfj/share/classes/com/ibm/dtfj/image/j9/ImageFactory.java
pshipton: I still think the javadoc should be updated, but I won't hold up this PR for it.

pshipton:
CI testing has passed. PR testing doesn't cover these changes.

PR #1224, opened 2018-02-20
[Nest mates] Moves nestmates host loading to visibility check

taliamccormick:
Earlier nest mates spec draft loaded a class's nest host during class loading. More recent nest mates spec puts off host loading until required for an access check. This commit moves the nest host loading and verification to the visibility check.

The spec draft for checking visibility states:
> To determine whether a class or interface C belongs to the same nest as
D (that is, whether C and D are nestmates), the following steps are
performed:
> 1. If C and D are the same class or interface, they belong to the same nest.
> 2. Otherwise, the nest host of D, H, is determined. If an exception is thrown, the nest membership test fails for the same reason.
> 3. Otherwise, the nest host of C, H', is determined. If an exception is thrown, the nest membership test fails for the same reason.
> 4. Otherwise, C and D belong to the same nest if H and H' are the same class or interface.
> 
> To determine the nest host of a class M, the following steps are performed:
> 
> 1. If M lacks a NestHost attribute (4.7.28), M is its own nest host.
> 2. Otherwise, where i is the host_class_index item of M's NestHost attribute, the symbolic reference at index i of M's run-time constant pool is resolved to a class or interface H (5.4.3.1). Any of the exceptions pertaining to class or interface resolution can be thrown.
> 3. If resolution of H succeeds, but H is not declared in the same
run-time package as M, an IncompatibleClassChangeError is thrown.
> 4. Otherwise, if H lacks a NestMembers attribute (4.7.29), or if, where M has name N, there exists no entry in the classes array of the NestMembers attribute of H that refers to a class or interface with name N, an IncompatibleClassChangeError is thrown.
> 5. Otherwise, H is the nest host of M.

As taken from here: http://cr.openjdk.java.net/~dlsmith/nestmates.html


taliamccormick:
@tajila

runtime/vm/visible.c
```diff
@@ -142,3 +155,55 @@ checkVisibility(J9VMThread *currentThread, J9Class* sourceClass, J9Class* destCl
...
+					break;
+				}
+			}
+			if (!verified) {
+				result = J9_VISIBILITY_NEST_MEMBER_NOT_CLAIMED_ERROR;
+			}
+		}
+
+		/* If any problem occurred in nest host verification, mask 0th bit */
+		if (J9_VISIBILITY_ALLOWED != result) {
```
tajila: do you still need this if you are returning the `J9_VISIBILITY_NEST*` error code?

runtime/nls/j9vm/j9vm.nls
```diff
@@ -1782,3 +1782,14 @@ J9NLS_VM_LOADING_CONSTRAINT_OVERRIDE_VIOLATION.explanation=A class loading const
 J9NLS_VM_LOADING_CONSTRAINT_OVERRIDE_VIOLATION.system_action=A java/lang/LinkageError will be thrown
 J9NLS_VM_LOADING_CONSTRAINT_OVERRIDE_VIOLATION.user_response=Examine the use of ClassLoaders to determine why a class of the same name is loaded in two different loaders
 # END NON-TRANSLATABLE
+
+# First argument is class name length
+# Second argument is class name
+# Third argument is nest host name length
+# Fourth argument is nest host name
+J9NLS_VM_NEST_HOST_FAILED_TO_LOAD=Class %2$.*1$s must be able to load its nest host %4$.*3$s.
```
tajila: do these messages matchup with the ones OpenJDK is using?

runtime/vm/visible.c
taliamccormick: I need some way of identifying which class - destClass or sourceClass - causes the nest verification error. I thought that the best way to do so was flag the 0 bit. A quick overview of the other options that I saw and my reasoning is below. Am open to other ideas/suggestions - having a non-functional non-null pointer may not be a good idea

---

Options:
1. Separate error codes for nest host errors in destClass & sourceClass
2. Changing function headers to pass a boolean/flag/other parameter to indicate nest host errors in either destClass or sourceClass
3. Flagging the 0th bit on the address
4. Putting a non-address value on the nest host (eg. -1)
5. Making error messages non-dependent on which classes have issues

---

Issues:
1. On a conceptual level, nest loading errors in either class would raise the same type of error, but  nothing explicitly wrong with it (other than longer if-statements).
2. Creates excessive changes to existing code.
3. Equivalent to this method, but would require searching for a class if later changes to the code want to access the nest host.
4. Risks that later changes to the code try to use the address to access nest host without masking the 0 bit; may cause confusion later.
5. Need to know which class has nest loading issue, in order to raise accurate errors.

runtime/nls/j9vm/j9vm.nls
taliamccormick: Some of them have different messages. The ones in which I add/directly use existing NLS messages, I will update to correspond with the ones OpenJDK is using.

What should I do for the cases in which existing error messages are different? By which I mean, the cases I did not add/change, eg. NoClassDefFound for testing the class loader case

---

I ran my updated set of test cases (currently found here: https://github.com/eclipse/openj9/pull/1201) against both. For each test that was supposed to raise an error/exception, I printed: test case name + error/exception + cause of error/exception.

Relevant test cases for this PR are:
* testDifferentClassLoadersWithAccess
* testDifferentPackagesWithAccess
* testNestMemberNotVerifiedWithAccess
* testNesthostNotClaimedWithAccess

---

Ours:
> testDifferentClassLoadersWithAccess             java.lang.NoClassDefFoundError: FieldAccess$Inner               java.lang.ClassNotFoundException

> testDifferentPackagesWithAccess         java.lang.IllegalAccessError: Class FieldAccess illegally accessing "package private" member of class pkg/FieldAccess$Inner             null

> testNestMemberNotVerifiedWithAccess             java.lang.VerifyError: Class FieldAccess$Inner must be claimed by its nest host .               null

> testNesthostNotClaimedWithAccess                java.lang.IllegalAccessError: Class FieldAccess illegally accessing "private" member of class FieldAccess$Inner         null

OpenJDK:
> testDifferentClassLoadersWithAccess             java.lang.NoClassDefFoundError: FieldAccess$Inner               java.lang.ClassNotFoundException: FieldAccess$Inner

> testDifferentPackagesWithAccess         java.lang.IllegalAccessError: failed to access class pkg.FieldAccess$Inner from class FieldAccess               null

> testNestMemberNotVerifiedWithAccess             java.lang.IllegalAccessError: Type FieldAccess$Inner is not a nest member of FieldAccess: current type is not listed as a nest member           null

> testNesthostNotClaimedWithAccess                java.lang.IllegalAccessError: tried to access field FieldAccess$Inner.f from class FieldAccess          null

---

On a side note, how close do/should the parser errors be? I've copied the same output for those cases below and they are similar but not exactly the same. 

This:
> testMultipleNestMembersAttributes               java.lang.ClassFormatError: JVMCFRE143 Multiple nest mates attributes; class=MultipleNestMembersAttributes, offset=166          null

> testMultipleNestHostAttributes          java.lang.ClassFormatError: JVMCFRE143 Multiple nest mates attributes; class=MultipleNestHostAttributes, offset=282             null

> testMultipleNestAttributes              java.lang.ClassFormatError: JVMCFRE143 Multiple nest mates attributes; class=MultipleNestAttributes, offset=188         null

OpenJDK:
> testMultipleNestMembersAttributes               java.lang.ClassFormatError: Multiple NestMembers attributes in class file MultipleNestMembersAttributes         null

> testMultipleNestHostAttributes          java.lang.ClassFormatError: Multiple NestHost attributes in class file MultipleNestHostAttributes               null

> testMultipleNestAttributes              java.lang.ClassFormatError: Conflicting NestHost and NestMembers attributes in class file MultipleNestAttributes                null

runtime/vm/visible.c
taliamccormick: As discussed yesterday, new/current approach is:
* Create two sets of visibility error codes; one set for nest visibility/verification issues in the source class and one set for nest visibility/verification issues in the destination class
* Shift the value of these visibility error codes such that they can be separated from the non-nest related error codes with a single comparison and that the source class codes and destination class codes can be separated from eachother via a single bitwise comparison.

eg.

Visibility check result | Value | Hex value | Last eight bits
--- | --- | --- | ---
J9_VISIBILITY_ALLOWED | 1 | 1 | 0000 0001
J9_VISIBILITY_NON_MODULE_ACCESS_ERROR | 0 | 0 | 0
J9_VISIBILITY_MODULE_READ_ACCESS_ERROR | -1 | FFFF FFFF | 1111 1111
J9_VISIBILITY_MODULE_PACKAGE_EXPORT_ERROR | -2 | FFFF FFFE | 1111 1110 
J9_VISIBILITY_NEST_HOST_LOADING_FAILURE_ERROR_SOURCE_CLASS | -17 | FFFF FFFF | 1110 1111
J9_VISIBILITY_NEST_HOST_LOADING_FAILURE_ERROR_TARGET_CLASS | -18 | FFFF FFEF | 1110 1110
J9_VISIBILITY_NEST_HOST_DIFFERENT_PACKAGE_ERROR_SOURCE_CLASS | -19 |  FFFF FFED | 1110 1101
J9_VISIBILITY_NEST_HOST_DIFFERENT_PACKAGE_ERROR_TARGET_CLASS | -20 | FFFF FFEC | 1110 1100
J9_VISIBILITY_NEST_MEMBER_NOT_CLAIMED_ERROR_SOURCE_CLASS | -21 | FFFF FFEB | 1110 1011
J9_VISIBILITY_NEST_MEMBER_NOT_CLAIMED_ERROR_TARGET_CLASS | -22 | FFFF FFEA | 1110 1010

Then we can detect a nest-related error via checking the range or a single bitwise comparison on the 5th- 8th last bits
```
if ((J9_VISIBILITY_NEST_HOST_LOADING_FAILURE_ERROR_SOURCE_CLASS >= errorCode)
    && (J9_VISIBILITY_NEST_MEMBER_NOT_CLAIMED_ERROR_TARGET_CLASS <= errorCode)) { ... }
```
```
#define J9_VISIBILITY_NEST_CODE 240 // 0000 ... 0000 1111 0000
#define J9_VISIBILITY_NEST_MASK 224 // 0000 ... 0000 1110 0000
if (J9_VISIBILITY_NEST_CODE == (J9_VISIBILITY_NEST_MASK & errorCode)) { ... }
```
And we can also use the parity bit to detect if the source class or destination class is causing the error/exception
```
if (0 == errorCode % 2) { ... /* target class */ }
if (1 == errorCode % 2) { ... /* destination class */ }
```

runtime/vm/lookupmethod.c
```diff
@@ -1031,10 +1031,81 @@ illegalAccessMessage(J9VMThread *currentThread, IDATA badMemberModifier, J9Class
...
 	Trc_VM_illegalAccessMessage_Entry(currentThread, J9UTF8_LENGTH(senderClassNameUTF), J9UTF8_DATA(senderClassNameUTF),
 			J9UTF8_LENGTH(targetClassNameUTF), J9UTF8_DATA(targetClassNameUTF), badMemberModifier);
+
+#if defined(J9VM_OPT_VALHALLA_NESTMATES)
+	/* Values of error codes for nestmates visibility checks on source class is
+	 * odd and on destination/target class is even
+	 */
+	if ((J9_VISIBILITY_NEST_HOST_LOADING_FAILURE_ERROR_SOURCE_CLASS <= errorType)
+			&& (J9_VISIBILITY_NEST_MEMBER_NOT_CLAIMED_ERROR_DEST_CLASS >= errorType)) {
+		if (0 == errorType % 2) {
```
tajila: a bit check might be less expensive then a modulo

runtime/vm/lookupmethod.c
```diff
@@ -1031,10 +1031,81 @@ illegalAccessMessage(J9VMThread *currentThread, IDATA badMemberModifier, J9Class
...
+			unverifiedNestMemberClass = targetClass;
+		} else if (1 == ((UDATA)targetClass->nestHost & 1)) {
```
tajila: do you still need to check this? if errorType is an odd value is the error always sender?

runtime/vm/visible.c
```diff
@@ -142,3 +155,55 @@ checkVisibility(J9VMThread *currentThread, J9Class* sourceClass, J9Class* destCl
...
+	J9UTF8 *nestHostName = J9ROMCLASS_NESTHOSTNAME(romClass);
+
+	/* If no nest host is named, class is own nest host */
+	if (NULL == nestHostName) {
+		nestHost = clazz;
+	} else {
+		UDATA classLoadingFlags = J9_FINDCLASS_FLAG_THROW_ON_FAIL;
+		nestHost = vmThread->javaVM->internalVMFunctions->internalFindClassUTF8(vmThread, J9UTF8_DATA(nestHostName), J9UTF8_LENGTH(nestHostName), clazz->classLoader, classLoadingFlags);
+
+		/* Nest host must successfully loaded by the same classloader in the same package & verify the nest member */
```
tajila: Nest host must <b>be</b> successfully loaded by

runtime/vm/visible.c
```diff
@@ -142,3 +155,55 @@ checkVisibility(J9VMThread *currentThread, J9Class* sourceClass, J9Class* destCl
...
+			}
+			if (!verified) {
+				result = J9_VISIBILITY_NEST_MEMBER_NOT_CLAIMED_ERROR_SOURCE_CLASS;
+			}
+		}
+
+		/* If any problem occurred in nest host verification, error code was set
+		 * assuming that the nest host is being loaded for the source class; if
+		 * not, add a -1 offset to change to dest class error code */
+		if (TRUE != isSourceClass) {
```
tajila: Will this decrement result even if it is `J9_VISIBILITY_ALLOWED`

taliamccormick: Missed that - thanks

runtime/vm/visible.c
```diff
@@ -142,3 +155,55 @@ checkVisibility(J9VMThread *currentThread, J9Class* sourceClass, J9Class* destCl
...
+			}
+			if (!verified) {
+				result = J9_VISIBILITY_NEST_MEMBER_NOT_CLAIMED_ERROR_SOURCE_CLASS;
+			}
+		}
+
+		/* If any problem occurred in nest host verification, error code was set
+		 * assuming that the nest host is being loaded for the source class; if
+		 * not, add a -1 offset to change to dest class error code */
+		if ((J9_VISIBILITY_ALLOWED == result) && (TRUE != isSourceClass)) {
```
tajila: should this be `J9_VISIBILITY_ALLOWED != result`

runtime/vm/lookupmethod.c
```diff
@@ -1031,10 +1031,84 @@ illegalAccessMessage(J9VMThread *currentThread, IDATA badMemberModifier, J9Class
...
+	 * 		J9_VISIBILITY_MODULE_READ_ACCESS_ERROR
+	 * 		J9_VISIBILITY_MODULE_PACKAGE_EXPORT_ERROR
+	 */
+	if (NULL != unverifiedNestMemberClass) {
+		J9ROMClass romClass = unverifiedNestMemberClass->romClass;
+		J9UTF8 *nestMemberNameUTF = J9ROMCLASS_CLASSNAME(romClass);
+		J9UTF8 *nestHostNameUTF = J9ROMCLASS_NESTHOSTNAME(romClass);
+
+		if ((J9_VISIBILITY_NEST_HOST_LOADING_FAILURE_ERROR_SOURCE_CLASS == errorType)
+				|| (J9_VISIBILITY_NEST_HOST_LOADING_FAILURE_ERROR_DEST_CLASS == errorType)
```
tajila: it looks like you have an extra tab here.

same below

runtime/vm/resolvesupport.cpp
```diff
@@ -678,8 +678,8 @@ resolveStaticFieldRefInto(J9VMThread *vmStruct, J9Method *method, J9ConstantPool
 						} else {
 							errorMsg = illegalAccessMessage(vmStruct, -1, classFromCP, targetClass, checkResult);
 						}
-						setCurrentExceptionUTF(vmStruct, J9VMCONSTANTPOOL_JAVALANGILLEGALACCESSERROR, errorMsg);
-						j9mem_free_memory(errorMsg);
+					setCurrentExceptionUTF(vmStruct, J9VMCONSTANTPOOL_JAVALANGILLEGALACCESSERROR, errorMsg);
```
tajila: this should also be tabbed

runtime/vm/lookupmethod.c
```diff
@@ -1031,10 +1031,84 @@ illegalAccessMessage(J9VMThread *currentThread, IDATA badMemberModifier, J9Class
...
 	
 	PORT_ACCESS_FROM_VMC(currentThread);
 	Trc_VM_illegalAccessMessage_Entry(currentThread, J9UTF8_LENGTH(senderClassNameUTF), J9UTF8_DATA(senderClassNameUTF),
 			J9UTF8_LENGTH(targetClassNameUTF), J9UTF8_DATA(targetClassNameUTF), badMemberModifier);
+
+#if defined(J9VM_OPT_VALHALLA_NESTMATES)
+	/* Values of error codes for nestmates visibility checks on source class is
+	 * odd and on destination/target class is even
+	 */
+	if ((J9_VISIBILITY_NEST_HOST_LOADING_FAILURE_ERROR_SOURCE_CLASS <= errorType)
```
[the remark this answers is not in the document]
DanHeidinga: This approach feels very hacky.  It will be easy for someone to add a new error message that doesn't follow this pattern without realizing it.

A switch statement will be more explicit and less fragile to future changes:

```
switch(errorType) {
case J9_VISIBILITY_NEST_HOST_LOADING_FAILURE_ERROR_SOURCE_CLASS:
case J9_VISIBILITY_NEST_HOST_DIFFERENT_PACKAGE_ERROR_SOURCE_CLASS:
case J9_VISIBILITY_NEST_MEMBER_NOT_CLAIMED_ERROR_SOURCE_CLASS:
    unverifiedNestMemberClass = targetClass;
    break;

case J9_VISIBILITY_NEST_HOST_LOADING_FAILURE_ERROR_DEST_CLASS
case J9_VISIBILITY_NEST_HOST_DIFFERENT_PACKAGE_ERROR_DEST_CLASS
case J9_VISIBILITY_NEST_MEMBER_NOT_CLAIMED_ERROR_DEST_CLASS
    unverifiedNestMemberClass = senderClass;
    break;
default:
   // Assert unreachable here is good when possible.  In this case, would still need to handle other error cases.
```

runtime/vm/lookupmethod.c
```diff
@@ -1031,10 +1031,84 @@ illegalAccessMessage(J9VMThread *currentThread, IDATA badMemberModifier, J9Class
...
+	 * 		J9_VISIBILITY_NON_MODULE_ACCESS_ERROR
+	 * 		J9_VISIBILITY_MODULE_READ_ACCESS_ERROR
+	 * 		J9_VISIBILITY_MODULE_PACKAGE_EXPORT_ERROR
+	 */
+	if (NULL != unverifiedNestMemberClass) {
+		J9ROMClass *romClass = unverifiedNestMemberClass->romClass;
+		J9UTF8 *nestMemberNameUTF = J9ROMCLASS_CLASSNAME(romClass);
+		J9UTF8 *nestHostNameUTF = J9ROMCLASS_NESTHOSTNAME(romClass);
+
+		if ((J9_VISIBILITY_NEST_HOST_LOADING_FAILURE_ERROR_SOURCE_CLASS == errorType)
```
DanHeidinga: This is another opportunity to use a switch.  It may or may not be cleaner here.

runtime/vm/visible.c
```diff
@@ -29,6 +29,10 @@
 #include "util_api.h"
 #include "vm_internal.h"
 
+#if defined(J9VM_OPT_VALHALLA_NESTMATES)
+static VMINLINE loadAndVerifyNestHost(J9VMThread *vmThread, J9Class *clazz, BOOLEAN isSourceClass);
```
DanHeidinga: What's the return type for this function?

runtime/vm/visible.c
```diff
@@ -142,3 +155,55 @@ checkVisibility(J9VMThread *currentThread, J9Class* sourceClass, J9Class* destCl
 
 	return result;
 }
+
+#if defined(J9VM_OPT_VALHALLA_NESTMATES)
+static VMINLINE UDATA loadAndVerifyNestHost(J9VMThread *vmThread, J9Class *clazz, BOOLEAN isSourceClass) {
```
DanHeidinga: This should only be called once per class in non-failing cases as the `nestHost` will either be itself or filled in by the first call.  I don't think it needs to be VMINLINEd.

For classes without a nest attribute, we can set the `nestHost` to themselves in createramclass.cpp and never call this function.  (TODO: confirm we do that already)

runtime/vm/visible.c
```diff
@@ -142,3 +155,55 @@ checkVisibility(J9VMThread *currentThread, J9Class* sourceClass, J9Class* destCl
...
+	UDATA result = J9_VISIBILITY_ALLOWED;
+	J9ROMClass *romClass = clazz->romClass;
+	J9UTF8 *nestHostName = J9ROMCLASS_NESTHOSTNAME(romClass);
+
+	/* If no nest host is named, class is own nest host */
+	if (NULL == nestHostName) {
+		nestHost = clazz;
+	} else {
+		UDATA classLoadingFlags = J9_FINDCLASS_FLAG_THROW_ON_FAIL;
+		nestHost = vmThread->javaVM->internalVMFunctions->internalFindClassUTF8(vmThread, J9UTF8_DATA(nestHostName), J9UTF8_LENGTH(nestHostName), clazz->classLoader, classLoadingFlags);
```
DanHeidinga: This doesn't need to call through the `internalVMFunctions` table as we're in the VM.  It can call `internalFindClassUTF8` directly

runtime/vm/visible.c
```diff
@@ -142,3 +155,55 @@ checkVisibility(J9VMThread *currentThread, J9Class* sourceClass, J9Class* destCl
...
+			}
+			if (!verified) {
+				result = J9_VISIBILITY_NEST_MEMBER_NOT_CLAIMED_ERROR_SOURCE_CLASS;
+			}
+		}
+
+		/* If any problem occurred in nest host verification, error code was set
+		 * assuming that the nest host is being loaded for the source class; if
+		 * not, add a -1 offset to change to dest class error code */
+		if ((J9_VISIBILITY_ALLOWED != result) && (TRUE != isSourceClass)) {
```
DanHeidinga: Not a fan of this approach.  It assumes no one will ever renumber the constants and if they do, it will result in subtly wrong error messages.

I would rather see a new out parameter to indicate source / dest or a more obvious way of mapping between the two.  A table maybe?

runtime/vm/lookupmethod.c
```diff
@@ -193,6 +193,7 @@ processMethod(J9VMThread * currentThread, UDATA lookupOptions, J9Method * method
 					} else {
 							*exception = J9VMCONSTANTPOOL_JAVALANGILLEGALACCESSERROR;
 							*exceptionClass = methodClass;
+							*errorType = J9_VISIBILITY_NON_MODULE_ACCESS_ERROR;
```
DanHeidinga: This is a really good find.  Since this is a bug fix, can it be split into a separate PR?

That PR can address the other exit points from this method and ensure they set errorType appropriately as well.

runtime/vm/lookupmethod.c
```diff
@@ -1035,6 +1036,69 @@ illegalAccessMessage(J9VMThread *currentThread, IDATA badMemberModifier, J9Class
...
+	 * 		J9_VISIBILITY_NEST_HOST_DIFFERENT_PACKAGE_ERROR
+	 * 		J9_VISIBILITY_NEST_MEMBER_NOT_CLAIMED_ERROR
+	 * Otherwise, it is one of:
+	 * 		J9_VISIBILITY_NON_MODULE_ACCESS_ERROR
+	 * 		J9_VISIBILITY_MODULE_READ_ACCESS_ERROR
+	 * 		J9_VISIBILITY_MODULE_PACKAGE_EXPORT_ERROR
+	 */
+	if ((J9_VISIBILITY_NEST_HOST_LOADING_FAILURE_ERROR == errorType)
+			|| (J9_VISIBILITY_NEST_HOST_DIFFERENT_PACKAGE_ERROR == errorType)
+			|| (J9_VISIBILITY_NEST_MEMBER_NOT_CLAIMED_ERROR == errorType)) {
```
DanHeidinga: formatting nitpick: the closing `) {` should be on its own line at the same indent level as the `if`.  Makes it much easier to visually track the start / end of the if block.

runtime/vm/lookupmethod.c
```diff
@@ -1035,6 +1036,69 @@ illegalAccessMessage(J9VMThread *currentThread, IDATA badMemberModifier, J9Class
...
+		J9Class *unverifiedNestMemberClass = NULL;
+		J9ROMClass *romClass = NULL;
+		J9UTF8 *nestMemberNameUTF = NULL;
+		J9UTF8 *nestHostNameUTF = NULL;
+
+		if (NULL == senderClass->nestHost) {
```
DanHeidinga: This needs a comment to describe why we can check in this order and know if the issue was with the sender or target.

runtime/vm/lookupmethod.c
```diff
@@ -1035,6 +1036,69 @@ illegalAccessMessage(J9VMThread *currentThread, IDATA badMemberModifier, J9Class
...
+
+		bufLen = j9str_printf(PORTLIB, NULL, 0, errorMsg,
+				J9UTF8_LENGTH(nestMemberNameUTF),
+				J9UTF8_DATA(nestMemberNameUTF),
+				J9UTF8_LENGTH(nestHostNameUTF),
+				J9UTF8_DATA(nestHostNameUTF));
+
+		if (bufLen > 0) {
+			buf = j9mem_allocate_memory(bufLen, OMRMEM_CATEGORY_VM);
+			if (NULL != buf) {
```
DanHeidinga: Since it uses a goto anyway, code can be written as:
```
if (NULL == buf) { 
    goto allocationFailures;
}
j9str_printf(PORTLIB, buf, bufLen, errorMsg,
	J9UTF8_LENGTH(nestMemberNameUTF),
	J9UTF8_DATA(nestMemberNameUTF),
	J9UTF8_LENGTH(nestHostNameUTF),
	J9UTF8_DATA(nestHostNameUTF));
```

DanHeidinga: This isn't something that needs to be changed.  Just another way to write this code.

runtime/vm/visible.c
```diff
@@ -100,12 +105,21 @@ checkVisibility(J9VMThread *currentThread, J9Class* sourceClass, J9Class* destCl
 			}
 		} else if (modifiers & J9AccPrivate) {
 			/* Private */
-			if ((sourceClass != destClass)
+			if (sourceClass != destClass) {
 #if defined(J9VM_OPT_VALHALLA_NESTMATES)
-					&& (sourceClass->nestHost != destClass->nestHost)
+				/* loadAndVerifyNestHost returns an error if setting nest host field fails */
+				if (NULL == destClass->nestHost) {
+					result = loadAndVerifyNestHost(currentThread, destClass);
```
DanHeidinga: if we add a `goto`, we can avoid the repeated `(J9_VISIBILITY_ALLOWED == result)` check in the common path.

Something like:
```
if (NULL == destClass->nestHost) {
	result = loadAndVerifyNestHost(currentThread, destClass);
	if ((J9_VISIBILITY_ALLOWED != result) {
		goto _exit;
	}
}
if (NULL == sourceClass->nestHost) {
....
}
```

runtime/vm/visible.c
```diff
@@ -142,3 +156,51 @@ checkVisibility(J9VMThread *currentThread, J9Class* sourceClass, J9Class* destCl
...
+	J9Class *nestHost = NULL;
+	UDATA result = J9_VISIBILITY_ALLOWED;
+	J9ROMClass *romClass = clazz->romClass;
+	J9UTF8 *nestHostName = J9ROMCLASS_NESTHOSTNAME(romClass);
+
+	/* If no nest host is named, class is own nest host */
+	if (NULL == nestHostName) {
+		nestHost = clazz;
+	} else {
+		UDATA classLoadingFlags = J9_FINDCLASS_FLAG_THROW_ON_FAIL;
```
DanHeidinga: TODO: check the code to see if we need to worry about this at JIT compile time.  If it's a compile time resolve, we hopefully aren't doing access already.

runtime/vm/visible.c
```diff
@@ -142,3 +156,51 @@ checkVisibility(J9VMThread *currentThread, J9Class* sourceClass, J9Class* destCl
...
+
+			for (i = 0; i < nestMemberCount; i++) {
+				J9UTF8 *nestMemberName = NNSRP_GET(nestMembers[i], J9UTF8*);
+				if (J9UTF8_EQUALS(className, nestMemberName)) {
+					verified = TRUE;
+					break;
+				}
+			}
+			if (!verified) {
+				result = J9_VISIBILITY_NEST_MEMBER_NOT_CLAIMED_ERROR;
```
DanHeidinga: This can avoid the `verified` boolean if `result = J9_VISIBILITY_NEST_MEMBER_NOT_CLAIMED_ERROR;` is set before the loop and then set to `J9_VISIBILITY_ALLOWED` if we find the nest member.

runtime/vm/visible.c
DanHeidinga: This is called by `checkVisibility`, which is called by `resolveClassRef` during jit compile time resolves.  

This code needs to check if this is a compile time resolve and if it is, replace `J9_FINDCLASS_FLAG_THROW_ON_FAIL` with `J9_FINDCLASS_FLAG_EXISTING_ONLY`.

DanHeidinga: `checkVisiblity` <- call by resolvefield.cpp code.  Needs to also respect `J9_RESOLVE_FLAG_NO_THROW_ON_FAIL` to indicate not to throw.

runtime/vm/lookupmethod.c
taliamccormick: Created a separate PR here: https://github.com/eclipse/openj9/pull/1383
Other exit points look like they handle it correctly; this was the odd one out.

runtime/vm/visible.c
DanHeidinga:
```
if ((resolveFlags & J9_RESOLVE_FLAG_NO_THROW_ON_FAIL)
|| (resolveFlags & COMPILE_TIME)
|| .... AOT....
|| ... J9_FINDCLASS_FLAG_EXISTING_ONLY ...
) {
classLoadingFlags = J9_FINDCLASS_FLAG_EXISTING_ONLY;
}
```

Or make the callers pass in `BOOLEAN canRunJavaCode` and use that to pick the `EXISTING_ONLY` or not flags.  Some callers may already know this.

runtime/vm/lookupmethod.c
```diff
@@ -1035,6 +1035,79 @@ illegalAccessMessage(J9VMThread *currentThread, IDATA badMemberModifier, J9Class
...
+	 * 		J9_VISIBILITY_NEST_MEMBER_NOT_CLAIMED_ERROR
+	 * Otherwise, it is one of:
+	 * 		J9_VISIBILITY_NON_MODULE_ACCESS_ERROR
+	 * 		J9_VISIBILITY_MODULE_READ_ACCESS_ERROR
+	 * 		J9_VISIBILITY_MODULE_PACKAGE_EXPORT_ERROR
+	 */
+	if ((J9_VISIBILITY_NEST_HOST_LOADING_FAILURE_ERROR == errorType)
+			|| (J9_VISIBILITY_NEST_HOST_DIFFERENT_PACKAGE_ERROR == errorType)
+			|| (J9_VISIBILITY_NEST_MEMBER_NOT_CLAIMED_ERROR == errorType)
+		) {
```
tajila: needs to be shifted to the left by one tab

runtime/vm/visible.c
```diff
@@ -141,12 +146,34 @@ checkVisibility(J9VMThread *currentThread, J9Class* sourceClass, J9Class* destCl
...
+				 *
+				 * Note that the specification asserts an order for nest host
+				 * loading during access checking. The accessing class's nest
+				 * host is loaded first and the class being accessed loads its
+				 * nest host after.
+				 */
+				if (NULL == sourceClass->nestHost) {
+					result = loadAndVerifyNestHost(currentThread, sourceClass, lookupOptions);
+				}
+				if (J9_VISIBILITY_ALLOWED != result) {
```
tajila: why isn't this in the same scope has the `(NULL == sourceClass->nestHost)` check above

runtime/vm/visible.c
```diff
@@ -183,3 +213,52 @@ checkVisibility(J9VMThread *currentThread, J9Class* sourceClass, J9Class* destCl
 
 	return result;
 }
+
+#if defined(J9VM_OPT_VALHALLA_NESTMATES)
+static UDATA loadAndVerifyNestHost(J9VMThread *vmThread, J9Class *clazz, UDATA options) {
```
tajila: the `{` at the end should be on a separate line

runtime/vm/visible.c
```diff
@@ -183,3 +213,52 @@ checkVisibility(J9VMThread *currentThread, J9Class* sourceClass, J9Class* destCl
...
+	UDATA result = J9_VISIBILITY_ALLOWED;
+	J9ROMClass *romClass = clazz->romClass;
+	J9UTF8 *nestHostName = J9ROMCLASS_NESTHOSTNAME(romClass);
+
+	/* If no nest host is named, class is own nest host */
+	if (NULL == nestHostName) {
+		nestHost = clazz;
+	} else {
+		UDATA classLoadingFlags = 0;
+		if ((options & J9_LOOK_NO_THROW ) != J9_LOOK_NO_THROW) {
```
tajila: use J9_ARE_NO_BITS_SET(...) macro for clarity

runtime/nls/j9vm/j9vm.nls
```diff
@@ -1782,3 +1782,14 @@ J9NLS_VM_LOADING_CONSTRAINT_OVERRIDE_VIOLATION.explanation=A class loading const
...
+# START NON-TRANSLATABLE
+J9NLS_VM_NEST_HOST_FAILED_TO_LOAD.explanation=Class and nest host class were loaded with different class loaders.
+J9NLS_VM_NEST_HOST_FAILED_TO_LOAD.system_action=The JVM will throw a java/lang/VerifyError
```
DanHeidinga: `java/lang/VerifyError` -> `?????` This is an IllegalAccessError now isn't it?  The check doesn't occur during class load anymore...

runtime/vm/visible.c
```diff
@@ -141,12 +146,35 @@ checkVisibility(J9VMThread *currentThread, J9Class* sourceClass, J9Class* destCl
...
+					result = loadAndVerifyNestHost(currentThread, sourceClass, lookupOptions);
+					if (J9_VISIBILITY_ALLOWED != result) {
+						goto _exit;
+					}
+				}
+				if (NULL == destClass->nestHost) {
+					result = loadAndVerifyNestHost(currentThread, destClass, lookupOptions);
+					if (J9_VISIBILITY_ALLOWED != result) {
+						goto _exit;
+					}
```
DanHeidinga: Same for destClass here.  It also needs to be reloaded.

runtime/vm/visible.c
```diff
@@ -158,12 +158,16 @@ checkVisibility(J9VMThread *currentThread, J9Class* sourceClass, J9Class* destCl
 				 */
 				if (NULL == sourceClass->nestHost) {
 					result = loadAndVerifyNestHost(currentThread, sourceClass, lookupOptions);
+					sourceClass = J9_CURRENT_CLASS(sourceClass);
```
DanHeidinga: Small nitpick - these only need to be reloaded if the visibility succeeds so this could move after the `if`

runtime/vm/visible.c
```diff
@@ -141,12 +146,39 @@ checkVisibility(J9VMThread *currentThread, J9Class* sourceClass, J9Class* destCl
...
+				 * loading during access checking. The accessing class's nest
+				 * host is loaded first and the class being accessed loads its
+				 * nest host after.
+				 */
+				if (NULL == sourceClass->nestHost) {
+					result = loadAndVerifyNestHost(currentThread, sourceClass, lookupOptions);
+					if (J9_VISIBILITY_ALLOWED != result) {
+						goto _exit;
+						sourceClass = J9_CURRENT_CLASS(sourceClass);
+						destClass = J9_CURRENT_CLASS(destClass);
```
DanHeidinga: Shouldn't these be after the `if` statement?  Otherwise they're dead code following the `goto`

DanHeidinga:
The code in `checkVisibility` is missing some tracepoints for the module and nestmate failing cases.  A separate PR can be used to add the tracepoints.

DanHeidinga:

DanHeidinga:
Recording that the build passed: https://ci.eclipse.org/openj9/job/PullRequest-Sanity-JDK9-linux_ppc-64_cmprssptrs_le-OpenJ9/202/

@taliamccormick Can you squash into logical sets of commits and we can merge this?

taliamccormick:
@DanHeidinga Squashed

Issue #1225, opened 2018-02-20
Write test case for JMX Agent


pdbain-ibm:
This is to test issue #1196 .
Test:
1) default agent
2) replacement agent specified via system property javax.management.builder.initial
3) bogus value of javax.management.builder.initial

PR #1226, opened 2018-02-20
Enable System.startSNMPAgent in OpenJ9 Java 8

### LINKED ISSUE #1196: Error cause not printed when JMX Agent fails to start
When appropriate command line options are set, a JMX Agent is started when the JVM launches. If the JMX Agent fails to start for any reason, the JVM also fails to start.

Here's an example command, where the JMX Agent fails because the specified class in the `javax.management.builder.initial` property does not exist:

`> java -Djavax.management.builder.initial=foo -Dcom.sun.management.jmxremote -version`

With Oracle/HotSpot the resulting error looks like this:

```
javax.management.JMRuntimeException: Failed to load MBeanServerBuilder class foo: java.lang.ClassNotFoundException: foo
        at javax.management.MBeanServerFactory.checkMBeanServerBuilder(MBeanServerFactory.java:503)
        at javax.management.MBeanServerFactory.getNewMBeanServerBuilder(MBeanServerFactory.java:539)
        at javax.management.MBeanServerFactory.newMBeanServer(MBeanServerFactory.java:316)
        at javax.management.MBeanServerFactory.createMBeanServer(MBeanServerFactory.java:231)
        at javax.management.MBeanServerFactory.createMBeanServer(MBeanServerFactory.java:192)
        at java.lang.management.ManagementFactory.getPlatformMBeanServer(ManagementFactory.java:469)
        at sun.management.jmxremote.ConnectorBootstrap.startLocalConnectorServer(ConnectorBootstrap.java:533)
        at sun.management.Agent.startLocalManagementAgent(Agent.java:138)
        at sun.management.Agent.startAgent(Agent.java:260)
        at sun.management.Agent.startAgent(Agent.java:447)
Caused by: java.lang.ClassNotFoundException: foo
        at java.net.URLClassLoader.findClass(URLClassLoader.java:381)
        at java.lang.ClassLoader.loadClass(ClassLoader.java:424)
        at sun.misc.Launcher$AppClassLoader.loadClass(Launcher.java:335)
        at java.lang.ClassLoader.loadClass(ClassLoader.java:357)
        at javax.management.MBeanServerFactory.loadBuilderClass(MBeanServerFactory.java:446)
        at javax.management.MBeanServerFactory.checkMBeanServerBuilder(MBeanServerFactory.java:488)
        ... 9 more
Exception thrown by the agent : javax.management.JMRuntimeException: Failed to load MBeanServerBuilder class foo: java.lang.ClassNotFoundException: foo
```
With OpenJ9 the same error looks like this:

```
Exception in thread "main" java/lang/InternalError: java.lang.reflect.InvocationTargetException
        at java/lang/System.completeInitialization (System.java:186)
        at java/lang/Thread.<init> (Thread.java:157)
```

As you can see, the root cause of the problem is swallowed in an InvocationTargetException, which is then wrapped in an InternalError. It's not even clear that the JMX Agent is involved.

The problem seems to be that OpenJ9 constructs the InternalError like this:
```
} catch (Exception e) {
	throw new InternalError(e.toString());
}
```
which means that the root cause of the InvocationTargetException (i.e. `e.getCause()`) is lost.

I tried changing the code to this:
```
} catch (Exception e) {
	throw new InternalError(e);
}
```
with the expectation that the JVM would print a chained stack trace with all the causes printed recursively (as described in the API doc for `Throwable.printStackTrace()`). But that isn't what happens - I still only see the single stack for the InternalError.

I also tried this testcase:

```
public class CauseTest {
	public static void main (String args[]) {
		try {
			throw new RuntimeException("Low level problem");
		} catch (Exception e) {
			throw new InternalError(e);
		}
	}
}
```
and it behaves as expected - i.e. I get a chained stack trace:

```
Exception in thread "main" java.lang.InternalError: java.lang.RuntimeException: Low level problem
        at CauseTest.main(CauseTest.java:6)
Caused by: java.lang.RuntimeException: Low level problem
        at CauseTest.main(CauseTest.java:4)
```
So there seems to be something different about how the InternalError is printed in the JMX Agent scenario.

pdbain-ibm:
Enable the same code as Java 9 for launching mbean server.

Fixes #1196 

### DIFF
--- a/jcl/src/java.base/share/classes/java/lang/Thread.java
+++ b/jcl/src/java.base/share/classes/java/lang/Thread.java
@@ -2,7 +2,7 @@
 package java.lang;
 
 /*******************************************************************************
- * Copyright (c) 1998, 2017 IBM Corp. and others
+ * Copyright (c) 1998, 2018 IBM Corp. and others
  *
  * This program and the accompanying materials are made available under
  * the terms of the Eclipse Public License 2.0 which accompanies this
@@ -184,8 +184,10 @@ void completeInitialization() {
 	contextClassLoader = ClassLoader.getSystemClassLoader();
 	/*[IF Sidecar19-SE]*/
 	jdk.internal.misc.VM.initLevel(4);
+	/*[ENDIF]*/ // Sidecar19-SE
+	/*[IF Sidecar19-SE|Sidecar18-SE-OpenJ9]*/
 	System.startSNMPAgent();
-	/*[ENDIF]*/
+	/*[ENDIF]*/ // Sidecar19-SE|Sidecar18-SE-OpenJ9
 }
 
 /**
--- a/runtime/jcl/uma/se8_exports.xml
+++ b/runtime/jcl/uma/se8_exports.xml
@@ -1,5 +1,5 @@
 <!-- 
-	Copyright (c) 2014, 2017 IBM Corp. and others
+	Copyright (c) 2014, 2018 IBM Corp. and others
 	
 	This program and the accompanying materials are made available under
 	the terms of the Eclipse Public License 2.0 which accompanies this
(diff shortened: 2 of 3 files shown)


pdbain-ibm:
Ready for review.

pshipton:
FYI #1225 

jenkins test all plinux jdk8

PR #1228, opened 2018-02-20
JIT runtime helper for final field notification

fengxue-IS:
- Implemented JIT runtime helper for final field notification
- Added placeholder function of JIT final field callback
- Removed null check for jitConfig function pointer

### DIFF
--- a/runtime/codert_vm/pnathelp.m4
+++ b/runtime/codert_vm/pnathelp.m4
@@ -1,4 +1,4 @@
-dnl Copyright (c) 2017, 2017 IBM Corp. and others
+dnl Copyright (c) 2017, 2018 IBM Corp. and others
 dnl
 dnl This program and the accompanying materials are made available under
 dnl the terms of the Eclipse Public License 2.0 which accompanies this
@@ -326,6 +326,7 @@ SLOW_PATH_ONLY_HELPER_NO_RETURN_VALUE(jitReportMethodExit,2)
 DUAL_MODE_HELPER_NO_RETURN_VALUE(jitTypeCheckArrayStore,2)
 DUAL_MODE_HELPER_NO_RETURN_VALUE(jitTypeCheckArrayStoreWithNullCheck,2)
 FAST_PATH_ONLY_HELPER(jitObjectHashCode,1)
+SLOW_PATH_ONLY_HELPER_NO_RETURN_VALUE(jitReportFinalFieldModified,1)
 
 dnl Trap handlers
 
--- a/runtime/codert_vm/xnathelp.m4
+++ b/runtime/codert_vm/xnathelp.m4
@@ -1,4 +1,4 @@
-dnl Copyright (c) 2017, 2017 IBM Corp. and others
+dnl Copyright (c) 2017, 2018 IBM Corp. and others
 dnl
 dnl This program and the accompanying materials are made available under
 dnl the terms of the Eclipse Public License 2.0 which accompanies this
@@ -380,6 +380,7 @@ SLOW_PATH_ONLY_HELPER_NO_RETURN_VALUE(jitReportMethodExit,2)
 DUAL_MODE_HELPER_NO_RETURN_VALUE(jitTypeCheckArrayStore,2)
 DUAL_MODE_HELPER_NO_RETURN_VALUE(jitTypeCheckArrayStoreWithNullCheck,2)
 FAST_PATH_ONLY_HELPER(jitObjectHashCode,1)
+SLOW_PATH_ONLY_HELPER_NO_RETURN_VALUE(jitReportFinalFieldModified,1)
 
 dnl Trap handlers
 
--- a/runtime/codert_vm/znathelp.m4
+++ b/runtime/codert_vm/znathelp.m4
@@ -1,4 +1,4 @@
-dnl Copyright (c) 2017, 2017 IBM Corp. and others
+dnl Copyright (c) 2017, 2018 IBM Corp. and others
 dnl
 dnl This program and the accompanying materials are made available under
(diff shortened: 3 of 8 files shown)


fjeremic:
> Please name this consistently with the other helpers. It appears slow path only should probably be old_slow_name.

Further to that is there a reason why we cannot use a fast helper here?

gacholio:
Should be a simple matter to add the JIT helper to the m4 files now.

gacholio:
Fast helpers cannot be GC points, which this certainly is (it acquires exclusive).

It could be turned into a dual-mode helper, so if we don't need to invalidate anything, it would be a bit faster.  No change is required in the JIT to change this later on.

The theory is that these illegal final field writes will be extremely infrequent.

gacholio:
For the m4 helpers, you want to use SLOW_PATH_ONLY_HELPER_NO_EXCEPTION_NO_RETURN_VALUE and change the C code to return void rather than void* (see old_slow_jitCallCFunction).

gacholio:
I also suggest adding an empty function in the JIT to implement the callback, then you can remove the NULL check of the function from the helper.

gacholio:
I'm wrong about the no exception - this is a possible decompile point, so use SLOW_PATH_ONLY_HELPER_NO_RETURN_VALUE and continue returning the void*.  You can further optimize the restore of the resolve frame by adding ", true, false" to the parameter list, indicating that it should check for frame pop, but there is no possibility of a java exception needing to be thrown.

runtime/codert_vm/pnathelp.m4
```diff
@@ -326,6 +326,7 @@ SLOW_PATH_ONLY_HELPER_NO_RETURN_VALUE(jitReportMethodExit,2)
 DUAL_MODE_HELPER_NO_RETURN_VALUE(jitTypeCheckArrayStore,2)
 DUAL_MODE_HELPER_NO_RETURN_VALUE(jitTypeCheckArrayStoreWithNullCheck,2)
 FAST_PATH_ONLY_HELPER(jitObjectHashCode,1)
+SLOW_PATH_ONLY_HELPER_NO_RETURN_VALUE(jitReportFinalFieldModified,0)
```
gacholio: Should be 1 rather than 0.

runtime/codert_vm/xnathelp.m4
```diff
@@ -380,6 +380,7 @@ SLOW_PATH_ONLY_HELPER_NO_RETURN_VALUE(jitReportMethodExit,2)
 DUAL_MODE_HELPER_NO_RETURN_VALUE(jitTypeCheckArrayStore,2)
 DUAL_MODE_HELPER_NO_RETURN_VALUE(jitTypeCheckArrayStoreWithNullCheck,2)
 FAST_PATH_ONLY_HELPER(jitObjectHashCode,1)
+SLOW_PATH_ONLY_HELPER_NO_RETURN_VALUE(jitReportFinalFieldModified,0)
```
gacholio: 1 again.

runtime/codert_vm/znathelp.m4
```diff
@@ -326,6 +326,7 @@ SLOW_PATH_ONLY_HELPER_NO_RETURN_VALUE(jitReportMethodExit,2)
 DUAL_MODE_HELPER_NO_RETURN_VALUE(jitTypeCheckArrayStore,2)
 DUAL_MODE_HELPER_NO_RETURN_VALUE(jitTypeCheckArrayStoreWithNullCheck,2)
 FAST_PATH_ONLY_HELPER(jitObjectHashCode,1)
+SLOW_PATH_ONLY_HELPER_NO_RETURN_VALUE(jitReportFinalFieldModified,0)
```
gacholio: 1 again.

runtime/compiler/trj9/runtime/codertinit.cpp
```diff
@@ -511,6 +512,7 @@ void codert_init_helpers_and_targets(J9JITConfig * jitConfig, char isSMP)
 #endif
    jitConfig->jitDiscardPendingCompilationsOfNatives = jitDiscardPendingCompilationsOfNatives;
    jitConfig->jitMethodBreakpointed = jitMethodBreakpointed;
+   jitConfig->jitIllegalFinalFieldModification = jitIllegalFinalFieldModification;
```
gacholio: This is not the file to fill in the function pointer - it should be somwhere in the JIT.  Ask Andrew.

fengxue-IS: this seems to be the only place that jitConfig function pointers are set? it is not part of JIT?

gacholio: Look in rossa.cpp for:

   // set up the entryPoints for compiling and for support of the
   // java.lang.Compiler class.

gacholio: And I suppose you could move the dummy function in there for simplicity - the JIT implementors can move it later if need be.  Make the function declaration static.

gacholio:
Once these changes are made, also remove the WIP from the title and add a description to both the commit and PR (might be best to start doing this from the beginning), as this completes our side of the work for now.

runtime/compiler/trj9/control/rossa.cpp
```diff
@@ -582,6 +582,12 @@ jitExclusiveVMShutdownPending(J9VMThread * vmThread)
    #endif
    }
 
+// Placeholder for JIT callback of final field notification
+extern "C" void
```
gacholio: I believe this should be declared static.

gacholio:
One last nitpick change, and remove the WIP from the title.

fengxue-IS:
fixed, testing with personal build

PR #1229, opened 2018-02-20
Update the format of method specification

hangshao0:
The method specification is used by option
"-Xshareclasses:invalidateAotMethods="
Current method specification classname[.methodname][(parameters)]
implies we can parameters without a method name, which is not correct.
The format should be changed to classname[.methodname[(parameters)]]


DanHeidinga:
@hangshao0 Are you aware of any tests for this option?  I'm wondering if either the extended or sanity tests buckets cover this.

hangshao0:
No. We do not have any test for this. These two messages are only in the text output of java -Xshareclasses:findAotMethods=help and -Xshareclasses:help.

hangshao0:
Fixed in all j9shr_XX.nls translations

pshipton:

PR #1230, opened 2018-02-20
Adjust DDR namespace annotations

keithc-ca:
* declarations in stackwalk.h grouped to reduce number of annotations required
* also in generated constant pool
* added VMCPTool option '-verbose' (default off)

Issue: #378


DanHeidinga:
@gacholio Can you review?  In particular the stackwalk.h changes

gacholio:
Two things:

1) What is this change for?

2) I really object to changesets that are 90% blank line additions and reordering - makes it very hard to find what to review.

The stackwalk.h change itself seems functionally correct.

keithc-ca:
The new OMR-based DDR generation tool looks for annotations of the form `@ddr_namespace: map_to_type=XXX` for guidance about what namespace the constants belong in. The old annotations were not consistent with the legacy tooling.

gacholio:
jenkins test sanity plinux

pshipton:
> hoping this eventually leads to no constant namespacing at all

I expect we'd have to make a choice to discard the namespaces, check if there are any collisions, and fix up all the DDR code to match.

gacholio:
It's a non-trivial change, to be sure.

keithc-ca:
There shouldn't be any collisions, but adjusting all the references would be tedious.

PR #1231, opened 2018-02-20
Fix ddr plugin manager

keithc-ca:
Not all Entry objects have an associated File and so, can't all contribute to the classpath.

Issue: #378

### DIFF
--- a/jcl/src/openj9.dtfj/share/classes/com/ibm/java/diagnostics/utils/plugins/impl/PluginManagerImpl.java
+++ b/jcl/src/openj9.dtfj/share/classes/com/ibm/java/diagnostics/utils/plugins/impl/PluginManagerImpl.java
@@ -1,6 +1,6 @@
 /*[INCLUDE-IF Sidecar18-SE]*/
 /*******************************************************************************
- * Copyright (c) 2012, 2017 IBM Corp. and others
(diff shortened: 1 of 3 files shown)


jcl/src/openj9.dtfj/share/classes/com/ibm/java/diagnostics/utils/plugins/impl/PluginManagerImpl.java
```diff
@@ -55,140 +55,157 @@
...
+			for (String part : parts) {
+				String path = stripQuotesFromPath(part);
+				File file = new File(path);
 				try {
-					String path = stripQuotesFromPath(parts[i]);
-					File file = new File(path);
 					pluginSearchPath.add(file);
 				} catch (Exception e) {
-					logger.warning("Failed to create a URI or URL from " + parts[i]);
+					logger.warning("Failed to create a URI or URL from " + path); //$NON-NLS-1$
```
pshipton: At least the type of exception should be logged

pshipton:
lgtm. CI testing has passed. I assume you'll want to squash before I merge.

pshipton:
Skip the ci testing since it passed before the squash.

PR #1232, opened 2018-02-20
WIP: Static read barrier

amicic:
Introduce API (external MemoryManagament and VM access barrier macros)
for read barrier in class static slots. Implementation is just slightly
different from the heap slot, due to slot being always non-compressed
reference. 

Also unifying the macros for reads (both static and non-static), to
resemble more to store macros.

Unrelated to barriers, removing J9 and using OMR constants for
Remembered Set.


amicic:
Current attempt for array read barrier is non-optimal, because effective address is in theory evaluated twice:

#define J9JAVAARRAYOFOBJECT_LOAD(vmThread, array, index) \
	(J9OBJECT__PRE_OBJECT_LOAD_ADDRESS(vmThread, array, J9JAVAARRAY_EA(vmThread, array, index, fj9object_t)), \
	 (j9object_t)J9_CONVERT_POINTER_FROM_TOKEN__(vmThread, *J9JAVAARRAY_EA(vmThread, array, index, fj9object_t)) \
	)

In practice, it would only matter for configuration that both have 1) dis-contiguous arrays and 2) read barrier enabled, and there is no such ATM.

This would be better, but apparently not supported by ANSI C, but only by specific compilers like GNU (https://gcc.gnu.org/onlinedocs/gcc/Statement-Exprs.html):

#define J9JAVAARRAYOFOBJECT_LOAD(vmThread, array, index) \
	({fj9object_t *loadAddress = J9JAVAARRAY_EA(vmThread, array, index, fj9object_t); \
	 J9OBJECT__PRE_OBJECT_LOAD_ADDRESS(vmThread, array, loadAddress); \
	 (j9object_t)J9_CONVERT_POINTER_FROM_TOKEN__(vmThread, *loadAddress);} \
	)

Note that this load is more complex than similar store macro, because this macro is supposed to have a 'return' value which a simple block with curly braces (and without the outer parenthesis) does not have.

@charliegracie, @dmitripivkine  or anybody else with an opinion or alternative idea?

amicic:
First approach is not even correct, since array and index could be in general case complex expressions and would resolved twice, potentially into different values. For example if index is 'i++' expression, the real index would be incorrect in the second invocation of J9JAVAARRAY_EA.

I'm trying now with the second approach (with temporary loadAddress), but use an inline method instead of that non-fully compatible macro. There are also problems with that approach since it uncovers some new inter-dependencies  between files,

amicic:
Jenkins test sanity

amicic:
Jenkins test sanity jdk8

amicic:

amicic:
I did not do any perf testing for specs that do not require read barrier. While statics accesses are affected they are much less frequent (than object accesses). For object access, while the code has been quite shuffled, I believed no extra runtime penalty is introduced - in fact it should stay same (it would fail on checking javaVM->gcReadBarrierType prior to an attempt to call any MM helper), and back then when I originally introduced object read barriers, we did not see measurable difference.

So, I'd proceed with the merge and follow up with perf test after it,

PR #1235, opened 2018-02-20
Add initial implementation of defaultvalue bytecode

eric98zhang:
Add initial implementation for the Valhalla L-World defaultvalue
bytecode.

This implementation allocates the new value class instance on the heap
and places a reference on the stack, in the same way the "new" bytecode
is implemented. Value class specific optimizations are intended for
subsequent commits.


eric98zhang:
@tajila

runtime/bcutil/ClassFileOracle.cpp
```diff
@@ -1734,6 +1734,13 @@ ClassFileOracle::walkMethodCodeAttributeCode(U_16 methodIndex)
 			}
 			break;
 		}
+#if defined(J9VM_OPT_VALHALLA_VALUE_TYPES)
+		case CFR_BC_defaultvalue:
+			cpIndex = PARAM_U16();
+			addBytecodeFixupEntry(entry++, codeIndex + 1, cpIndex, ConstantPoolMap::NEW);
```
eric98zhang: Instead of using `ConstantPoolMap::NEW`, should I define a `ConstantPoolMap::DEFAULT_VALUE = SPLIT1`?

runtime/vm/BytecodeInterpreter.hpp
```diff
@@ -7885,8 +7885,74 @@ done:;
...
+		j9object_t newObject = NULL;
 		VM_BytecodeAction rc = EXECUTE_BYTECODE;
-		// TODO: Implement bytecode
+		U_16 index = *(U_16*)(_pc + 1);
+		J9ConstantPool *ramConstantPool = J9_CP_FROM_METHOD(_literals);
+		J9RAMClassRef *ramCPEntry = ((J9RAMClassRef*)ramConstantPool) + index;
+		J9Class* volatile resolvedClass = ramCPEntry->value;
+
+		if (NULL != resolvedClass) {
+
```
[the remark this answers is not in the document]
tajila: in fact I think the only difference between new at this point is the `(0 == (resolvedClass->romClass->modifiers & J9AccValueType)` check

runtime/bcutil/ClassFileOracle.cpp
tajila: yes

runtime/vm/BytecodeInterpreter.hpp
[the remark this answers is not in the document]
eric98zhang: To make this change, I added in a new constant in `j9consts.h` for use in `resolveClassRef()` in `resolvesupport.cpp` when throwing the ICCE.

tajila:

tajila:
@babsingh please take a look

babsingh:
@eric98zhang 
- Can you post a link to `defaultvalue` bytecode's documentation/specification?
- Can you post a test plan? How are you going to test these changes?

eric98zhang:
Here are the most up to date L-World links I could find.

Latest L-World proposal: http://cr.openjdk.java.net/~acorn/LWorldValueTypesFeb13.pdf  
Latest rough draft JVMS: http://cr.openjdk.java.net/~fparain/L-world/L-World-JVMS-4b.pdf  

For a better description of the bytecodes, there is also this older MVT document: http://cr.openjdk.java.net/~jrose/values/shady-values.html

and also #1390

eric98zhang:
**TEST PLAN**

**1. Positive case**

Class `A` is a value type with some fields. `defaultvalue` is used with `A`. A new value type instance should be created. The fields of the instance are checked using `getfield` to ensure they are set to the correct default values. 

**2. `defaultvalue` on a non-value type**

 Class `B` is not a value type and does not have the `ACC_VALUE_TYPE` access flag. `defaultvalue` used with `B` should throw `IncompatibleClassChangeError`.

Note: ASM will be necessary to generate the `defaultvalue` bytecode and class `A`.

Testing is limited only to these `defaultvalue` specific cases, and does not include cases for class resolutions errors, trying to instantiate interfaces or abstract classes, etc.

runtime/util/jbcmap.c
```diff
@@ -247,7 +247,7 @@ const unsigned char sunJavaInstructionSizeTable[] = {
 0x00 /* JBunimplemented = 222 = 0xDE */,
 0x00 /* JBunimplemented = 223 = 0xDF */,
 #if defined(J9VM_OPT_VALHALLA_VALUE_TYPES)
-0x01 /* JBdefaultvalue = 224 = 0xE0 */ ,
+0x03 /* JBdefaultvalue = 224 = 0xE0 */ ,
```
babsingh: Referring to `page 36` in http://cr.openjdk.java.net/%7Efparain/L-world/L-World-JVMS-4b.pdf, `defaultvalue
 = 203 (0xcb)`. You need to update `jbcmap.c`.

From - `0x03 /* JBdefaultvalue = 224 = 0xE0 */ ,`

To - `0x03 /* JBdefaultvalue = 203 = 0xCB */ ,`

runtime/vm/resolvesupport.cpp
```diff
@@ -394,6 +394,15 @@ resolveClassRef(J9VMThread *vmStruct, J9ConstantPool *ramCP, UDATA cpIndex, UDAT
 		}
 	}
 
+#if defined(J9VM_OPT_VALHALLA_VALUE_TYPES)
+	if (J9_ARE_ANY_BITS_SET(resolveFlags, J9_RESOLVE_FLAG_CHECK_VALUE_CLASS)) {
```
babsingh: Use `J9_ARE_ALL_BITS_SET` instead of `J9_ARE_ANY_BITS_SET`.
`((bits) == ((value) & (bits)))` reinforces the flag check better than `(0 != ((value) & (bits)))`.

runtime/vm/resolvesupport.cpp
```diff
@@ -394,6 +394,15 @@ resolveClassRef(J9VMThread *vmStruct, J9ConstantPool *ramCP, UDATA cpIndex, UDAT
...
+		if (J9_ARE_NO_BITS_SET(accessModifiers, J9AccValueType)) {
```
babsingh: This can be combined with the previous `if` statement using `&&`.

runtime/vm/resolvesupport.cpp
```diff
@@ -394,6 +394,15 @@ resolveClassRef(J9VMThread *vmStruct, J9ConstantPool *ramCP, UDATA cpIndex, UDAT
...
+			setCurrentException(vmStruct, J9VMCONSTANTPOOL_JAVALANGINCOMPATIBLECLASSCHANGEERROR, NULL); // TODO: add detail message
```
babsingh: `TODO` comment is hidden; put it on a new line, and use `/* ... */` instead of `//`. According to the coding standards, `//` is not preferred for a comment.

runtime/vm/BytecodeInterpreter.hpp
```diff
@@ -7885,8 +7885,65 @@ done:;
...
+		j9object_t newObject = NULL;
 		VM_BytecodeAction rc = EXECUTE_BYTECODE;
-		// TODO: Implement bytecode
+		U_16 index = *(U_16*)(_pc + 1);
+		J9ConstantPool *ramConstantPool = J9_CP_FROM_METHOD(_literals);
+		J9RAMClassRef *ramCPEntry = ((J9RAMClassRef*)ramConstantPool) + index;
+		J9Class* volatile resolvedClass = ramCPEntry->value;
+
+		if ((NULL != resolvedClass) && (resolvedClass->romClass->modifiers & J9AccValueType)) {
+
```
babsingh: unneeded new line.

runtime/vm/BytecodeInterpreter.hpp
```diff
@@ -7885,8 +7885,65 @@ done:;
...
+retry:
+		j9object_t newObject = NULL;
 		VM_BytecodeAction rc = EXECUTE_BYTECODE;
-		// TODO: Implement bytecode
+		U_16 index = *(U_16*)(_pc + 1);
+		J9ConstantPool *ramConstantPool = J9_CP_FROM_METHOD(_literals);
+		J9RAMClassRef *ramCPEntry = ((J9RAMClassRef*)ramConstantPool) + index;
+		J9Class* volatile resolvedClass = ramCPEntry->value;
+
+		if ((NULL != resolvedClass) && (resolvedClass->romClass->modifiers & J9AccValueType)) {
```
babsingh: Use a `j9comp.h` macro for `(resolvedClass->romClass->modifiers & J9AccValueType)`.

runtime/vm/BytecodeInterpreter.hpp
```diff
@@ -7885,8 +7885,65 @@ done:;
...
+				rc = GOTO_ASYNC_CHECK;
+				goto done;
+			} else if (VM_VMHelpers::exceptionPending(_currentThread)) {
+				rc = GOTO_THROW_CURRENT_EXCEPTION;
+				goto done;
+			}
+			goto retry;
+		}
+
+		/* Unresolved */
```
babsingh: Add more detail to this comment. What is unresolved, and why we need to resolve it?

runtime/vm/BytecodeInterpreter.hpp
```diff
@@ -7885,8 +7885,65 @@ done:;
...
 	{
+retry:
+		j9object_t newObject = NULL;
 		VM_BytecodeAction rc = EXECUTE_BYTECODE;
-		// TODO: Implement bytecode
+		U_16 index = *(U_16*)(_pc + 1);
+		J9ConstantPool *ramConstantPool = J9_CP_FROM_METHOD(_literals);
+		J9RAMClassRef *ramCPEntry = ((J9RAMClassRef*)ramConstantPool) + index;
+		J9Class* volatile resolvedClass = ramCPEntry->value;
+
```
babsingh: Add a new a local variable `BOOLEAN doInitializeClass = FALSE;` [Remove redundant code]

runtime/vm/BytecodeInterpreter.hpp
```diff
@@ -7885,8 +7885,65 @@ done:;
...
+					VMStructHasBeenUpdated(REGISTER_ARGS);
+					if (J9_UNEXPECTED(NULL == instance)) {
+						rc = THROW_HEAP_OOM;
+						goto done;
+					}
+				}
+				newObject = instance;
+				goto done;
+			}
+			/* Class requires initialization */
```
babsingh: Set `doInitializeClass = TRUE;` here. Remove lines `7913-7925`. [Remove redundant code]

runtime/vm/BytecodeInterpreter.hpp
```diff
@@ -7885,8 +7885,65 @@ done:;
...
+				rc = GOTO_THROW_CURRENT_EXCEPTION;
+				goto done;
+			}
+			goto retry;
+		}
+
+		/* Unresolved */
+		buildGenericSpecialStackFrame(REGISTER_ARGS, 0);
+		updateVMStruct(REGISTER_ARGS);
+		resolveClassRef(_currentThread, ramConstantPool, index, J9_RESOLVE_FLAG_RUNTIME_RESOLVE | J9_RESOLVE_FLAG_INIT_CLASS | J9_RESOLVE_FLAG_INSTANTIABLE | J9_RESOLVE_FLAG_CHECK_VALUE_CLASS);
```
babsingh:
```
if (doInitializeClass) {
    initializeClass(_currentThread, resolvedClass);
} else {
    resolveClassRef(_currentThread, ramConstantPool, index, J9_RESOLVE_FLAG_RUNTIME_RESOLVE | J9_RESOLVE_FLAG_INIT_CLASS | J9_RESOLVE_FLAG_INSTANTIABLE | J9_RESOLVE_FLAG_CHECK_VALUE_CLASS);
}
```
You will be able to remove redundant code related to building/restoring stack frames and error checking.

runtime/vm/BytecodeInterpreter.hpp
```diff
@@ -7885,8 +7885,65 @@ done:;
 	VMINLINE VM_BytecodeAction
 	defaultvalue(REGISTER_ARGS_LIST)
 	{
+retry:
+		j9object_t newObject = NULL;
 		VM_BytecodeAction rc = EXECUTE_BYTECODE;
-		// TODO: Implement bytecode
+		U_16 index = *(U_16*)(_pc + 1);
+		J9ConstantPool *ramConstantPool = J9_CP_FROM_METHOD(_literals);
+		J9RAMClassRef *ramCPEntry = ((J9RAMClassRef*)ramConstantPool) + index;
```
babsingh: Can we use `const` for `index`, `ramConstantPool` and `ramCPEntry`? http://duramecho.com/ComputerInformation/WhyHowCppConst.html

babsingh:
Can you verify if the following `Linking Exceptions` (**LE**) and `Run-time Exception` (**RE**) covered `defaultvalue`'s implementation?
- [**LE**] During resolution of the symbolic reference to the class, array, or interface type, 
any of the exceptions documented in §5.4.3.1 can be thrown. 
- [**LE**] Otherwise, if the symbolic reference to the class, array, or interface type resolves 
to an interface or is an abstract class, defaultvalue throws an InstantiationError.
- [**LE**] Otherwise, if the symbolic reference to the class, array, or interface type resolves to an object class, defaultvalue throws an IncompatibleClassChangeError.
- [**RE**] Otherwise, if execution of this defaultvalue instruction causes initialization of the referenced class, defaultvalue may throw an Error as detailed in JLS §15.9.4.

Refer to page `36-37` in http://cr.openjdk.java.net/%7Efparain/L-world/L-World-JVMS-4b.pdf.

runtime/util/jbcmap.c
eric98zhang: As with withfield:

It looks like there are currently some internal bytecodes using 203 and 204 at the moment (iloadw and lloadw according bcnames.c). I think we decided to go with a hack in ClassFileOracle.cpp:1392 to remporarily remap 204 to 226 and 203 to 224 until a final LWVT spec is released.

eric98zhang:
I spoke with @tajila about the linking and run-time exceptions. No checks have been added to the static verifier yet so testing for most of the exceptions are TODO. I think we wanted to first get the core functionality down.

For now, there is a new run time check that covers
> **[LE]** Otherwise, if the symbolic reference to the class, array, or interface type resolves to an object class, defaultvalue throws an IncompatibleClassChangeError.
in the same way `newLogic()` checks for interfaces and abstract classes.

babsingh:
lgtm. please verify that your local testing passes. for future reference, your local tests need to be accessible. although your local tests aren't in TestNG format, you can create a `WIP` pull request for your test code.
- Create a new directory under https://github.com/eclipse/openj9/tree/master/test for your tests
- Add your test code and a `README` (compile and run commands) in the new directory

babsingh:
Senior Reviewer: @DanHeidinga

eric98zhang:
@babsingh I have added tests in #1619 to address:
>- **[LE]** Otherwise, if the symbolic reference to the class, array, or interface type resolves
to an interface or is an abstract class, defaultvalue throws an InstantiationError.
>- **[LE]** Otherwise, if the symbolic reference to the class, array, or interface type resolves to an object class, defaultvalue throws an IncompatibleClassChangeError.

I believe the other two are covered as well (since they aren't unique to `defaultvalue` and I have used the pre-existing mechanisms for class resolution and initialization), but I did not write any tests for them.

runtime/bcutil/ConstantPoolMap.hpp
```diff
@@ -348,6 +348,9 @@ class ConstantPoolMap
 		PUT_FIELD = SPLIT1,
 		GET_FIELD = SPLIT1,
 		NEW = SPLIT1,
+#if defined(J9VM_OPT_VALHALLA_VALUE_TYPES)
+		DEFAULT_VALUE = SPLIT1,
```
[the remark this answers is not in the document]
DanHeidinga: Sorry @eric98zhang I missed your response.  

Both `new` and `defaultvalue` are able to point to the same constantpool entry.  Unless there's an update to the JVMS to indicate otherwise, the verifier won't track which bytecode uses which cp entry - only that the data is consistent.  In the case of `new` and `defaultvalue`, both  bytecodes use the CONSTANT_Class cpEntries so there's nothing else for the verifier to check.

This is the kind of issue that needs to be handled at resolution or in the bytecodes.  In OpenJ9, each cpEntry is two slots and a bytecode like `new` relies on the values in those slots to determine whether its resolved or not.  

Once the cpEntry is resolved, all users share that resolved status so we need to either not resolve a cpEntry if its used by both new and defaultvalue (slow), or find another way to encode the data so that both bytecodes have unique resolution states, or to add a check in the bytecode to validate the resolved value is of the right kind (class vs value).

OpenJ9's design has typically been to aim for finding the right data encoding so that all the cost is at resolution time, not execution time.  

@tajila pointed out that `new` already has a check though for whether the resolved J9Class is `abstract` or `interface` so adding a new case to this for `value` seems reasonable and no more expensive then current code.

runtime/util/jbcmap.c
eric98zhang: there are several other places where the opcodes will need to be remapped (`bcnames.h`, `bcnames.c`, `cfr.h`, `bcsizes`, etc.). Should I leave the same comments in all of those locations?

Issue #1105, opened 2018-02-05
AOT processor feature validation is too strict


mpirvu:
When using AOT code loaded from the shared class cache the JIT compiler must make sure that the processor is currently running on is compatible in terms of features with the processor used at the time the AOT code was generated. For instance, if the AOT code exploits features like SSE4 or Hardware Lock Elision we cannot allow the code to run on processors that do not have these features. 

To guarantee this compatibility a function called `checkAOTHeaderFlags` ensures that the features of the processors used during AOT compilation and during AOT load match. The problem is that `checkAOTHeaderFlags` looks for a perfect match in terms of features, i.e. the AOT loads are prohibited even when the processor used to generate the AOT body has less features than the processor on which an AOT load is attempted.

This aspect has implications on AOT usage in the cloud. Assuming we would like to embed a pre-populated shared class cache (with AOT) in a docker image, that AOT code cannot run on any processor that is not identical to the processor used for the creation of the image and AOT code. This is too strong of a requirement.

This issues proposes to relax the processor feature compatibility check:

1. AOT loads should be permitted when the current processor has more features than the processor used to generate the AOT body
2. Even for a feature mismatch AOT loads should be permitted if the JIT does not exploit that feature (this requires knowledge of hardware exploits by the JIT)

harryyu1994:
@mpirvu Hi Marius, I'm wondering where in the code the AOTHeader(specifically its feature flag) gets stored in the shared class cache. Also, how is the currently feature flag created? Does it contain features that it's not using? We should have such a feature flag that it only contains features that it has exploited.

harryyu1994:
@mpirvu Hi Marius, I'm wondering where in the code or how the AOT header's feature flag stored to shared class cache is generated. Does that feature flag only contain features that's needed? We should make it only contain needed features as the minimum requirement to the current processor.

mpirvu:
@harryyu1994 
The AOTHeader is stored when the first AOT compilation happens, in `compilationEnd()`.
```
                  if (static_cast<TR_JitPrivateConfig *>(jitConfig->privateConfig)->aotValidHeader == TR_yes)
                     {
                     safeToStore = true;
                     }
                  else if (static_cast<TR_JitPrivateConfig *>(jitConfig->privateConfig)->aotValidHeader == TR_maybe)
                     {
                     // If validation has been performed, then a header already existed
                     // or one was already been created in this JVM
[...]

PR #1237, opened 2018-02-21
Relax AOT processor feature compatibility check

harryyu1994:
- Currently AOT compatibility check looks for a perfect match in terms of processor models and processor features
- This fix will allow AOT loads to be permitted when the features of the processor are the same or a super set and when the processor models are compatible

Issue: #1105

### DIFF
--- a/runtime/compiler/trj9/env/J9CPU.hpp
+++ b/runtime/compiler/trj9/env/J9CPU.hpp
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2000, 2016 IBM Corp. and others
+ * Copyright (c) 2000, 2018 IBM Corp. and others
  *
  * This program and the accompanying materials are made available under
  * the terms of the Eclipse Public License 2.0 which accompanies this
@@ -35,7 +35,6 @@ namespace J9 { typedef CPU CPUConnector; }
 #include "env/OMRCPU.hpp"
 #include "j9port.h"
 
-
 namespace J9
 {
 class CPU : public OMR::CPUConnector
@@ -44,11 +43,7 @@ protected:
    CPU() : OMR::CPUConnector() {}
 
 public:
-   bool getPPCSupportsVMX();
-   bool getPPCSupportsVSX();
-   bool getPPCSupportsAES();
-   bool getPPCSupportsTM();
-   bool getPPCSupportsLM();
+
    J9ProcessorDesc *TO_PORTLIB_getJ9ProcessorDesc();
    };
 }
--- a/runtime/compiler/trj9/control/rossa.cpp
+++ b/runtime/compiler/trj9/control/rossa.cpp
@@ -1082,7 +1082,7 @@ onLoadInternal(
       return -1;
 
 #if defined(TR_TARGET_X86)
-   TR_J9VM::initializeX86ProcessorVendorId(jitConfig);
+   TR_J9VM::initializeX86ProcessorInfo(jitConfig);
 #endif
 
    if (!TR_J9VMBase::createGlobalFrontEnd(jitConfig, TR::CompilationInfo::get()))
--- a/runtime/compiler/trj9/env/J9CPU.cpp
+++ b/runtime/compiler/trj9/env/J9CPU.cpp
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2000, 2016 IBM Corp. and others
+ * Copyright (c) 2000, 2018 IBM Corp. and others
  *
  * This program and the accompanying materials are made available under
  * the terms of the Eclipse Public License 2.0 which accompanies this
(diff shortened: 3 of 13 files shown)


harryyu1994:
@andrewcraik Hi Andrew, could you take a look at this changeset?
For x, can we eliminate the check for processorSignature altogether and only check the X86ProcessorFeatureFlags for compatibility?
We permit AOT load only if the current processor’s X86ProcessorFeatureFlags are the same or a super set of the X86ProcessorFeatureFlags in the code cache.

This change is Part 1 of Marius' proposed solution for issue 1105.

andrewcraik:
So I'm not a fan of the #ifdef usage in this change - it is expanding the use of platform specific macros rather than reducing them which I feel should be a goal. I don't see why we need different fields on the header object for the flags - we could have 3 generic fields and how they are interpreted is based on platform. I am not a fan of the platform specific methods being put in the top level API - would be nice to delegate or have a single function which tests the platform at runtime and picks the right interpretation of the flags and checks them.

The TODO comment left in one of your functions should at least have a description of what is required. Some kind of tracing around the level checks is necessary so we can see why a check passed or failed.

harryyu1994:
Hi Andrew @andrewcraik , thanks for your feedbacks! I've made the corresponding changes: removed the #ifdef's and platform specific methods in the top level API, and all platforms now share 3 generic processor feature flag fields.

Question regarding X86: would the comparison of x86 processor feature flags alone be sufficient to determine compatibility or do we still need to check the processor signatures? (i.e. if x86_processor_A contains all the features x86_processor_B has, can we conclude that A is compatible with B without looking into which type of processor A and B is?)

harryyu1994:
Marius is suggesting a better design, no need to review for now @andrewcraik .

andrewcraik:
@mpirvu thanks for talking with @harryyu1994 - a summary of the plan would be great whenever you get the chance (if the pull request doesn't come first). The contribution in this area is really appreciated and I'm sure with a few tweaks to the design we can get something that achieves the goal of helping to relax the AOT validation constraints while also not needing quite so many ifdefs

andrewcraik:
@harryyu1994 in response to your question about processor features - testing the features should be sufficient for functional compatibility. Sometimes the code generator may query the model to select instruction sequences that will run better on the given architecture. For AOT functional compatibility is probably the main thing we need so the flags should be what we need to ensure that is satisfied - the recompilation can help take care of any poorly performing choices.

harryyu1994:
Design summary: there is a common CPU class in trj9/env/J9CPU.hpp, then there is a CPU class in each of the 3 platforms that inherits from the common CPU class. I found this place to be most suitable for our code. I created a `getProcessorFeatureFlags` and `isCompatible` method in CPU class and each platform has its own implementation.`TR::Compiler->target.cpu.isCompatible((TR_Processor)hdrInCache->processorSignature, hdrInCache->processorFeatureFlags)` gives you compatibility and `TR::Compiler->target.cpu.getProcessorFeatureFlags()` gives you featureFlag of the corresponding platform. The current commit is a rough draft of what it looks like, is this design okay with you guys? @mpirvu @andrewcraik Thanks!

harryyu1994:
Also in the common CPU class, there are a couple of PPC specific methods that I think should be moved to the CPU class for Power.

mpirvu:
Agree with moving those PPC specific queries.
Regarding you implementation, is there any reason you have virtual methods for `getProcessorFeatureFlags()` and `isCompatible()`? Isn't better just to define them in their platform specific files?

harryyu1994:
I added the pure virtual methods as a way to force new derived classes to define these methods.

andrewcraik:
It would be better to use extensible inheritance for the flag checking, IMO, so the inheritance model matches that of the code generators. In that model it is a style of static inheritance (a bit like CRTP) and there would be no virtual methods. The base class would just implement a method that generates a hard fatal assert and implementers replace the implementation. The virtual keyword will increase the size of objects and require virtual dispatch which can be unfortunate.

runtime/compiler/trj9/env/J9CPU.hpp
```diff
@@ -49,6 +61,10 @@ class CPU : public OMR::CPUConnector
    bool getPPCSupportsAES();
    bool getPPCSupportsTM();
    bool getPPCSupportsLM();
+
+   virtual TR_ProcessorFeatureFlags getProcessorFeatureFlags() = 0;
```
mpirvu: As discussed, please remove these virtual functions.

harryyu1994:
Now the implementation follows extensible inheritance, `TR_ProcessorFeatureFlags`, previously defined in the common J9CPU.hpp, has been moved to the platform specific J9CPU.hpp's. This allows us to have different `TR_ProcessorFeatureFlags` size for different platforms.

mpirvu:
@harryyu1994 Please do some manual verification that the code works as expected. You need two machines that have processors with different feature sets. Generate the shared class cache on one, copy the SCC to the other and look to see what happens to the AOT loads. Then repeat the experiment switching the machine roles.

mpirvu:
Jenkins test all

harryyu1994:
There is a bug with the code, I will look into it and let you know @mpirvu when it's ready for running "Jenkins test all".

harryyu1994:
Hi Andrew @andrewcraik, would it be okay if I remove/move these functions from the common TR_J9VMBase class to its platform specific CPU class?(Maybe they should be dealt with in another issue/pull request.) I'm seeing a lot of platform specific methods that should not be in the top level api and could be put into the CPU class. 
```
const char *
TR_J9VMBase::getX86ProcessorVendorId()
   {
   TR_ASSERT(x86VendorIDInitialized, "X86 Vendor ID has not yet been initialized");
   return x86VendorID;
   }

uint32_t
TR_J9VMBase::getX86ProcessorSignature()
   {
   TR_ASSERT(_jitConfig, "jitConfig is not defined!");
   TR_ASSERT(_jitConfig->javaVM, "jitConfig->javaVM is not defined!");
   return queryX86TargetCPUID(_jitConfig->javaVM)->_processorSignature;
   }

uint32_t
TR_J9VMBase::getX86ProcessorFeatureFlags()
   {
   TR_ASSERT(_jitConfig, "jitConfig is not defined!");
   TR_ASSERT(_jitConfig->javaVM, "jitConfig->javaVM is not defined!");
   return queryX86TargetCPUID(_jitConfig->javaVM)->_featureFlags;
   }

uint32_t
TR_J9VMBase::getX86ProcessorFeatureFlags2()
   {
   TR_ASSERT(_jitConfig, "jitConfig is not defined!");
   TR_ASSERT(_jitConfig->javaVM, "jitConfig->javaVM is not defined!");
   return queryX86TargetCPUID(_jitConfig->javaVM)->_featureFlags2;
   }

uint32_t
TR_J9VMBase::getX86ProcessorFeatureFlags8()
   {
   TR_ASSERT(_jitConfig, "jitConfig is not defined!");
   TR_ASSERT(_jitConfig->javaVM, "jitConfig->javaVM is not defined!");
   return queryX86TargetCPUID(_jitConfig->javaVM)->_featureFlags8;
   }

bool
TR_J9VMBase::testOSForSSESupport()
   {
   return true; // VM guarantees SSE/SSE2 are available
   }

bool
TR_J9VMBase::getX86SupportsSSE4_1()
   {
   uint32_t flags = getX86ProcessorFeatureFlags2();
   if ((flags & 0x00080000) != 0x00080000)
      return false;
   return true;
   }

bool
TR_J9VMBase::getX86SupportsHLE()
   {
   uint32_t flags8 = getX86ProcessorFeatureFlags8();
   if ((flags8 & TR_HLE) != 0x00000000)
      return true;
   else return false;
   }

bool
TR_J9VMBase::getX86SupportsPOPCNT()
   {
   uint32_t flags2 = getX86ProcessorFeatureFlags2();
   if ((flags2 & TR_POPCNT) != 0x00000000)
      return true;
   else return false;
   }
```

harryyu1994:
```
uint32_t
CPU::getX86ProcessorFeatureFlags(TR::Compilation *comp)
   {
   return comp->fej9()->getX86ProcessorFeatureFlags();
   }

uint32_t
CPU::getX86ProcessorFeatureFlags2(TR::Compilation *comp)
   {
   return comp->fej9()->getX86ProcessorFeatureFlags2();
   }
```

Another question: since I'm seeing these functions have a `TR::Compilation` parameter passed in, are the X86featureFlags always fixed per run or we can get different featureFlags for different compilations in a single run? I'm planning to store the x86featureFlags values in its CPU class so that we don't need to call `queryX86TargetCPUID` every time we want the featureFlag values. But if there's a chance we get different featureFlag values from queryX86TargetCPUID each time we call, then the storing probably won't work.

harryyu1994:
Hi Marius, @mpirvu, I did some manual verifications on two x86 linux machines with different featureFlags (one is superset of the other). The code works as expect. The code is now ready for running the extended jenkins tests.

mpirvu:
Jenkins test all

harryyu1994:
```
09:43:47 /home/jenkins/jenkins-agent/workspace/PullRequest-Sanity-JDK8-linux_ppc-64_cmprssptrs_le-OpenJ9/build/linux-ppc64-normal-server-release/vm/compiler/../compiler/trj9/env/ProcessorDetection.cpp: In static member function ‘static void TR_J9VMBase::initializeX86ProcessorInfo(J9JITConfig*)’:
09:43:47 /home/jenkins/jenkins-agent/workspace/PullRequest-Sanity-JDK8-linux_ppc-64_cmprssptrs_le-OpenJ9/build/linux-ppc64-normal-server-release/vm/compiler/../compiler/trj9/env/ProcessorDetection.cpp:586:29: error: ‘class TR::CPU’ has no member named ‘setX86ProcessorFeatureFlags’
09:43:47     TR::Compiler->target.cpu.setX86ProcessorFeatureFlags(queryX86TargetCPUID(jitConfig->javaVM)->_featureFlags);
```
In the code, there is a macro guard for this, but it looks like the above ppc64 build has TR_TARGET_X86 set
```
#if defined(TR_TARGET_X86)
   TR_J9VM::initializeX86ProcessorInfo(jitConfig);
#endif
```

harryyu1994:
Never mind I know what went wrong.

harryyu1994:
Ready for Jenkins test all now

mpirvu:
Jenkins test all

mpirvu:
@harryyu1994   I did a final review and it occurred to me that we may need to bump `TR_AOTHeaderMinorVersion` to 1 because we changed the structure of the header (at least on Z and P).
While you are at it can you please delete the `#if defined(TR_TARGET_X86)` from `TR_J9VMBase::initializeX86ProcessorInfo(J9JITConfig *jitConfig)`. That function is already protected by `#if defined(TR_TARGET_X86)` at the outer level.

harryyu1994:
Removing that `#if defined(TR_TARGET_X86)` causes the build to fail on other platforms, since `TR_J9VMBase::initializeX86ProcessorInfo` is not guarded by any ifdefs so it gets compiled on every platform and the function `TR::Compiler->target.cpu.setX86ProcessorFeatureFlags(queryX86TargetCPUID(jitConfig->javaVM)->_featureFlags);` is only defined in X86.  We can either put a guard on `TR_J9VMBase::initializeX86ProcessorInfo(J9JITConfig *jitConfig)` or move this X86 specific function out of TR_J9VMBase. It should be easy to move because it's only called once in one location.

mpirvu:
I am ok with leaving the #ifdef in place for now. Restructuring should be part of another item.

harryyu1994:
This change is ready to be "Jenkins test all" and merged.

mpirvu:
Jenkins test sanity

harryyu1994:
Created an issue (#1612) for the restructuring/refactoring.

PR #1238, opened 2018-02-21
Return if a classloader never loaded a class

liqunl:
Add an early return to save time in the unloading hook since no
assumption will be registered for a classloader that has never loaded
a class.


andrewcraik:
Jenkins test all

andrewcraik:
Looks good to me - @mpirvu FYI

mpirvu:
Looks good to me to too. 
The inference  "classLoader->classSegments == NULL"  ==> "no classes have been loaded by this classloader" totally makes sense to me, though it may be prudent to check with the VM team if you haven't done it already.

liqunl:
The check is provided by the gc team.

liqunl:
I have jobs in my pbuild unfinished, will remove WIP once those jobs passed.

mpirvu:
@liqunl I'll merge this PR once you remove the WIP prefix

PR #1240, opened 2018-02-21
Remove #if defined in j9nonbuilder.h to fix DDR compile error

eric98zhang:
Resolves DDR compile error introduced in:
9a88388c32389458b98adb3bfa26ab882f59cf53


eric98zhang:
@tajila

PR #1242, opened 2018-02-22
Don't report final field modification at jit compile time

DanHeidinga:
JIT compile time resolves shouldn't report the final field modifications.
Reporting can cause deadlocks between the safepoint exclusive and
ClassUnloadMutex.

Better to fail the compile resolve and act as though unresolved.


gacholio:
jenkins test sanity plinux

Issue #577, opened 2017-11-14
Add ThreadMXBean.dumpAllThreads(..., int maxDepth) and getThreadInfo(..., int maxDepth) APIs


pshipton:
ThreadMXBean.dumpAllThreads(..., int maxDepth) and getThreadInfo(..., int maxDepth) need to be implemented for Java 18.3 / 10. Added in +31.

pdbain-ibm:
Per discussion with pshipton, will investigate

pdbain-ibm:
The code is implemented and manually tested.

Working on the testcase, but the test repository will likely need some rework.  Discussing this
with @llxia and @smlambert.

pdbain-ibm:
I have a testcase written.  Blocked waiting for https://github.com/eclipse/openj9/pull/1247.

pshipton:
Fixed by #1471

PR #1245, opened 2018-02-22
Move common ThreadMXBean testcase code to TestUtilities

pdbain-ibm:
There were two nearly identical copies of this code in Java8andUp. Further, there will be 
new tests in Java10andUp which need this code.  Migrate this to the shared TestUtilities
so it can be used by multiple projects.

Extract the differences into methods implemented by subclasses.  Note that
ThreadMXBeanTestCaseCommon.java is essentially ThreadMXBeanTestCase.java with
the differences encapsulated in an abstract method.  

Also fix warnings due to missing attributes.


pdbain-ibm:
@llxia and @smlambert ready for review.
This is a prerequisite for testing #577.

llxia:
Jenkins test sanity

pdbain-ibm:
Sanity test has passed.

PR #1246, opened 2018-02-22
Fix trees around jProfiling value call trees

r30shah:
Before lowering JProfling value calls, scan the trees around to make
sure it won't orphan a compressedrefs anchor node.

### DIFF
--- a/runtime/compiler/trj9/optimizer/JProfilingValue.cpp
+++ b/runtime/compiler/trj9/optimizer/JProfilingValue.cpp
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2017, 2017 IBM Corp. and others
+ * Copyright (c) 2017, 2018 IBM Corp. and others
  *
  * This program and the accompanying materials are made available under
  * the terms of the Eclipse Public License 2.0 which accompanies this
@@ -264,10 +264,10 @@ void
 TR_JProfilingValue::lowerCalls()
    {
    TR::TreeTop *cursor = comp()->getStartTree();
+   TR_BitVector *backwardAnalyzedAddressNodesToCheck = new (comp()->trStackMemory()) TR_BitVector();
    while (cursor)
       {
       TR::Node * node = cursor->getNode();
-
       if (node->isProfilingCode() &&
           node->getOpCodeValue() == TR::treetop &&
           node->getFirstChild()->getOpCode().isCall() &&
@@ -275,12 +275,35 @@ TR_JProfilingValue::lowerCalls()
            node->getFirstChild()->getSymbolReference()->getReferenceNumber() == TR_jProfile64BitValue) &&
           node->getFirstChild()->getNumChildren() == 3)
          {
-         TR::TreeTop *nextTree = cursor->getNextTreeTop();
-         TR::Node *child = node->getFirstChild();
+         // Backward Analysis in the extended basic block to get list of address nodes 
+         for (TR::TreeTop *iter = cursor->getPrevTreeTop();
+            iter && (iter->getNode()->getOpCodeValue() != TR::BBStart || iter->getNode()->getBlock()->isExtensionOfPreviousBlock());
+            iter = iter->getPrevTreeTop())
+            {
+            TR::Node *currentTreeTopNode = iter->getNode();
+            if (currentTreeTopNode->getNumChildren() >= 1 && currentTreeTopNode->getFirstChild()->getType() == TR::Address)
+               backwardAnalyzedAddressNodesToCheck->set(currentTreeTopNode->getFirstChild()->getGlobalIndex());
+            }
+         // Forward walk to check for compressedref anchors of any evaluated address nodes identified above
+         for (TR::TreeTop *iter = cursor->getNextTreeTop();
+            iter && (iter->getNode()->getOpCodeValue() != TR::BBStart || iter->getNode()->getBlock()->isExtensionOfPreviousBlock());
+            iter = iter->getNextTreeTop())
+            {
+            TR::Node *currentTreeTopNode = iter->getNode();
+            if (currentTreeTopNode->getOpCodeValue() == TR::compressedRefs
+               && backwardAnalyzedAddressNodesToCheck->isSet(currentTreeTopNode->getFirstChild()->getGlobalIndex()))
+               {
+               dumpOptDetails(comp(), "%s Moving treetop node n%dn above the profiling call to avoid uncommoning\n",
+                  optDetailString(), iter->getNode()->getGlobalIndex());
+               iter->unlink(false);
+               cursor->insertBefore(iter);
+               }
+            }
 
+         backwardAnalyzedAddressNodesToCheck->empty();
+         TR::Node *child = node->getFirstChild();
          dumpOptDetails(comp(), "%s Replacing profiling placeholder n%dn with value profiling trees\n",
             optDetailString(), child->getGlobalIndex());
-
          // Extract the arguments and add the profiling trees
          TR::Node *value = child->getFirstChild();
          TR_AbstractHashTableProfilerInfo *table = (TR_AbstractHashTableProfilerInfo*) child->getSecondChild()->getAddress();
@@ -288,10 +311,8 @@ TR_JProfilingValue::lowerCalls()
 
          // Remove the original trees and continue from the tree after the profiling
          TR::TransformUtil::removeTree(comp(), cursor);
-         cursor = nextTree;
          }
-      else
-         cursor = cursor->getNextTreeTop();
+      cursor = cursor->getNextTreeTop();
       }
    }
 


r30shah:
@andrewcraik Can you please review this fix?

runtime/compiler/trj9/optimizer/JProfilingValue.cpp
```diff
@@ -280,15 +280,42 @@ TR_JProfilingValue::lowerCalls()
...
+         if (prevTTNode->getNumChildren() >=1 && prevTTNode->getFirstChild()->getType() == TR::Address)
+            {
+            for (TR::TreeTop *itr = nextTT; itr && 
+               (itr->getNode()->getOpCodeValue() != TR::BBStart || itr->getNode()->getBlock()->isExtensionOfPreviousBlock());
+               itr = itr->getNextTreeTop())
+               {
+               if (itr->getNode()->getOpCodeValue() == TR::compressedRefs
+                  && itr->getNode()->getFirstChild() == prevTTNode->getFirstChild())
+                  {
+                  traceMsg(comp(), "Moving treetop node n%dn above the profiling call to avoid uncommoning\n", itr->getNode()->getGlobalIndex());
```
fjeremic: Should this be guarded by `if (trace())`?

runtime/compiler/trj9/optimizer/JProfilingValue.cpp
```diff
@@ -280,15 +280,42 @@ TR_JProfilingValue::lowerCalls()
...
+                  TR::TreeTop *itrPrev = itr->getPrevTreeTop();
+                  TR::TreeTop *itrNext = itr->getNextTreeTop();
+                  prevTT->setNextTreeTop(itr);
+                  cursor->setPrevTreeTop(itr);
+                  itr->setPrevTreeTop(prevTT);
+                  itr->setNextTreeTop(cursor);
+                  itrPrev->setNextTreeTop(itrNext);
+                  itrNext->setPrevTreeTop(itrPrev);
```
fjeremic: Are we able to use one of the following APIs to simplify this code greatly?

https://github.com/eclipse/omr/blob/f698070a815a92a8f1da548b6de042ef73bcd9e7/compiler/il/OMRTreeTop_inlines.hpp#L80-L103

r30shah: Those simplified inlines should be used here. Let me make changes. Thanks for the suggestion...

runtime/compiler/trj9/optimizer/JProfilingValue.cpp
```diff
@@ -280,15 +280,42 @@ TR_JProfilingValue::lowerCalls()
...
             optDetailString(), child->getGlobalIndex());
-
+         TR::TreeTop *prevTT = cursor->getPrevTreeTop();
+         TR::TreeTop *nextTT = cursor->getNextTreeTop();
+         TR::Node *prevTTNode = prevTT->getNode();
          // Extract the arguments and add the profiling trees
          TR::Node *value = child->getFirstChild();
          TR_AbstractHashTableProfilerInfo *table = (TR_AbstractHashTableProfilerInfo*) child->getSecondChild()->getAddress();
+         // While lowering value profiling calls, Scan the previous trees with the next trees to make sure by lowering the value profiler calls,
+         // We are not splitting blocks which orphans compressedref anchor nodes.
```
andrewcraik: I'm not clear on why prevTT is the right node to look at - the node we care about is the first child of the call that is going to be replaced correct? The treetop for the call is cursor not cursor->getPrevTreeTop()

r30shah:
As per chat with @andrewcraik, from the point of value profiling tree, we need to scan both backward and forward to make sure we have not orphaned compressedref anchor node which would be more concrete solution here. Working on making those changes. Will update on same PR.

andrewcraik:
Marked WIP to show the change is being revised per last comment from @r30shah

andrewcraik:
I think the current implementation uses too many maps. If you change the walk order to walk backwards first to record the nodes evaluated before the call and then walk forward to find any compressed refs anchors that could be adversely affected by the block split then you should need only 1 bit vector. You could likely make the check incremental for the block - always just adding to the backwards walk bit vector and re-running the forward walk which would save further time.

r30shah:
Thanks for the suggestion @andrewcraik . Just made the changes that uses only one `BitVector` . With this change we will move `compressedrefs anchor node` just before `jProfilingValue` call node not right after address node, which I think should be fine.

runtime/compiler/trj9/optimizer/JProfilingValue.cpp
```diff
@@ -264,34 +264,56 @@ void
 TR_JProfilingValue::lowerCalls()
    {
    TR::TreeTop *cursor = comp()->getStartTree();
+   int32_t numberOfNodes = comp()->getFlowGraph()->getNextNodeNumber();
+   TR_BitVector *backwardAnalyzedListOfAddressNodes = new (comp()->trStackMemory())TR_BitVector(numberOfNodes, trMemory(), stackAlloc);
```
fjeremic: Isn't the number of nodes going to grow in the loop below if you end up uncommoning a node? @andrewcraik will this break things on a subsequent iteration?

andrewcraik: This should be allocated growable and should just grow as needed to save space rather than zeroing everything. If it is growable and not presized the increase in nodes won't hurt.

r30shah:
Made the changes to use stack allocated grow-able `TR_BitVector` without prefixing size. Should be good for review.

runtime/compiler/trj9/optimizer/JProfilingValue.cpp
```diff
@@ -264,34 +264,56 @@ void
 TR_JProfilingValue::lowerCalls()
    {
    TR::TreeTop *cursor = comp()->getStartTree();
+   int32_t numberOfNodes = comp()->getFlowGraph()->getNextNodeNumber();
+   TR_BitVector *backwardAnalyzedListOfAddressNodes = new (comp()->trStackMemory())TR_BitVector();
```
andrewcraik: please add a space between the argument to new and the TR_BitVector for code style.

andrewcraik: Can we please get a more helpful variable name for the BitVector? It should describe what the bit vector holds, not how it was generated. Some suggestions could include: evaluatedAddressNodes, addressNodesToCheck, ....

runtime/compiler/trj9/optimizer/JProfilingValue.cpp
```diff
@@ -264,34 +264,56 @@ void
...
-
       if (node->isProfilingCode() &&
           node->getOpCodeValue() == TR::treetop &&
           node->getFirstChild()->getOpCode().isCall() &&
           (node->getFirstChild()->getSymbolReference()->getReferenceNumber() == TR_jProfile32BitValue ||
            node->getFirstChild()->getSymbolReference()->getReferenceNumber() == TR_jProfile64BitValue) &&
           node->getFirstChild()->getNumChildren() == 3)
          {
-         TR::TreeTop *nextTree = cursor->getNextTreeTop();
+         // Backward Analysis in extended BBs to get list of address nodes
```
andrewcraik: in the extended basic block this should read

runtime/compiler/trj9/optimizer/JProfilingValue.cpp
```diff
@@ -264,34 +264,56 @@ void
...
           node->getFirstChild()->getOpCode().isCall() &&
           (node->getFirstChild()->getSymbolReference()->getReferenceNumber() == TR_jProfile32BitValue ||
            node->getFirstChild()->getSymbolReference()->getReferenceNumber() == TR_jProfile64BitValue) &&
           node->getFirstChild()->getNumChildren() == 3)
          {
-         TR::TreeTop *nextTree = cursor->getNextTreeTop();
+         // Backward Analysis in extended BBs to get list of address nodes 
+         for (TR::TreeTop *iter = cursor->getPrevTreeTop();
+            iter && (iter->getNode()->getOpCodeValue() != TR::BBStart || iter->getNode()->getBlock()->isExtensionOfPreviousBlock());
+            iter=iter->getPrevTreeTop())
```
andrewcraik: spaces around the =

runtime/compiler/trj9/optimizer/JProfilingValue.cpp
```diff
@@ -264,34 +264,56 @@ void
...
+            {
+            TR::Node *currentTreeTopNode = iter->getNode();
+            if (currentTreeTopNode->getNumChildren() >= 1 && currentTreeTopNode->getFirstChild()->getType() == TR::Address)
+               backwardAnalyzedListOfAddressNodes->set(currentTreeTopNode->getFirstChild()->getGlobalIndex());
+            }
+         // Forward Analysis in extended BBs to get a list of compressedref anchor nodes
```
andrewcraik: something like forward walk to check for compressedref anchors of any evaluated address nodes identified above

runtime/compiler/trj9/optimizer/JProfilingValue.cpp
```diff
@@ -264,34 +264,56 @@ void
...
+            iter=iter->getPrevTreeTop())
+            {
+            TR::Node *currentTreeTopNode = iter->getNode();
+            if (currentTreeTopNode->getNumChildren() >= 1 && currentTreeTopNode->getFirstChild()->getType() == TR::Address)
+               backwardAnalyzedListOfAddressNodes->set(currentTreeTopNode->getFirstChild()->getGlobalIndex());
+            }
+         // Forward Analysis in extended BBs to get a list of compressedref anchor nodes 
+         for (TR::TreeTop *iter = cursor->getNextTreeTop();
+            iter && (iter->getNode()->getOpCodeValue() != TR::BBStart || iter->getNode()->getBlock()->isExtensionOfPreviousBlock());
+            iter=iter->getNextTreeTop())
```
andrewcraik: spaces around =

runtime/compiler/trj9/optimizer/JProfilingValue.cpp
```diff
@@ -264,34 +264,56 @@ void
...
+            {
+            TR::Node *currentTreeTopNode = iter->getNode();
+            if (currentTreeTopNode->getOpCodeValue() == TR::compressedRefs &&
+               backwardAnalyzedListOfAddressNodes->isSet(currentTreeTopNode->getFirstChild()->getGlobalIndex()))
```
andrewcraik: it might be easier to read this if the && is moved to the start of the isSet check like is done for many long conditions in the optimizer

runtime/compiler/trj9/optimizer/JProfilingValue.cpp
```diff
@@ -264,34 +264,56 @@ void
...
+               {
+               iter->unlink(false);
+               cursor->insertBefore(iter);
+               dumpOptDetails(comp(), "%s Moving treetop node n%dn above the profiling call to avoid uncommoning\n",
```
andrewcraik: This should happen before the unlink and insert incase of a crash

andrewcraik:
Change looks reasonable - given the nature of the change can you please provide some compile-time measurements to show the implementation does not significantly affect compile-time? I would expect a small increase, but it should be reasonable. Main thing is we don't want a major regression - Spark HiBench or another suite which does a lot of profiled compilations would be ideal. Measuring the compile-time increase using the verbose log to get the compilation time for the profiled compiles before and after would be ideal.

runtime/compiler/trj9/optimizer/JProfilingValue.cpp
```diff
@@ -264,34 +264,56 @@ void
 TR_JProfilingValue::lowerCalls()
    {
    TR::TreeTop *cursor = comp()->getStartTree();
+   int32_t numberOfNodes = comp()->getFlowGraph()->getNextNodeNumber();
```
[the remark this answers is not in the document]
r30shah: Yes should be removed

r30shah:
@andrewcraik Made the suggested changes in last commit. Going to run Spark today With and without this changes enabling jProfiling in compilations to measure the overhead this has introduced.

r30shah:
Finished `Spark/HiBench` runs to get a summary of compile time performance enabling jProfiling with and without this fix. I had to omit `micro.terasort` and `websearch.pagerank` to get fair comparisons as these two were failing without fix.
In summary, I do not see a significant difference in the total compilation time in all micro benchmarks. While in very-hot compilations where this change was expect us to slow down, it was `6%` slower on `Geo Mean`. Although looking into the % standard deviation in these times which ranges from `15%` to `70%` in various micro benchmarks, I think it is safe to say that this is well in the noise range.
If changes are OK, this can be merged after rebasing the branch and sanity testing

r30shah:
@andrewcraik  Not sure what is happening with travis-ci (Can not open the job log) but personal sanity tests passes with this change. Is it possible to restart travis-ci build? I am removing WIP as this is ready to merge.

andrewcraik:
@r30shah I think normally a force push will restart it. Once the CI is green I'll launch sanity here before merging.

r30shah:
@andrewcraik  Seems like CI build is passed. Can you launch sanity build?

andrewcraik:
Jenkins test sanity

r30shah:
Seems like we needed to rebase to today to get it passes. 
```
../makelib/targets.mk:301: recipe for target 'SchedulingDelegate.o' failed
``` 
Looks like missing this commit: https://github.com/eclipse/openj9/commit/65120285da31e9006748b68c6fd1739113f046a1

fjeremic:
Jenkins test sanity

andrewcraik:
@r30shah something is still not right - thoughts? might be related to issues reported with the OMR acceptance build on the OpenJ9 slack?

r30shah:
@andrewcraik  Yes, Just checked the slack channel. It seems like the issue with OMR acceptance.

andrewcraik:
Jenkins test sanity

PR #1249, opened 2018-02-22
Include libdwarf-dev in docker images

keithc-ca:
To support use of ddrgen from OMR and consistency between Java 8 and Java 9.

Issue: #378


pshipton:
Skipping PR testing since it doesn't exercise these changes.

PR #1258, opened 2018-02-23
Change fastIdentityHashCode() to use unsigned int to long conversion

DanHeidinga:
J9VMInternals.fastIdentityHashCode(Object) incorrectly sign-extended the
32bit J9Class pointer when reading it from an object header in a
compressedrefs VM.

This can can cause crashes with i.e. InaccessibleReadAddress=FFFFFFFF999609D8
in Java_com_ibm_jit_JITHelpers_getBackfillOffsetFromJ9Class64.

The int value needs to be assigned to the long without sign-extension.
Code now uses Integer.toUnsignedLong(int) to work around the issue.

JShell example showing this is valid:
```
jshell> int ptr = 0x99960900
ptr ==> -1718220544

jshell> long lptr = Integer.toUnsignedLong(ptr)
lptr ==> 2576746752

jshell> Long.toHexString(lptr)
$3 ==> "99960900"

jshell> long castptr = ptr
castptr ==> -1718220544

jshell> Long.toHexString(castptr)
$5 ==> "ffffffff99960900"
```

### DIFF
--- a/jcl/src/java.base/share/classes/java/lang/J9VMInternals.java
+++ b/jcl/src/java.base/share/classes/java/lang/J9VMInternals.java
@@ -30,7 +30,7 @@ import com.ibm.oti.util.Msg;
 
 
 /*******************************************************************************
- * Copyright (c) 1998, 2017 IBM Corp. and others
+ * Copyright (c) 1998, 2018 IBM Corp. and others
  *
  * This program and the accompanying materials are made available under
  * the terms of the Eclipse Public License 2.0 which accompanies this
@@ -439,7 +439,7 @@ final class J9VMInternals {
 				}
 			}
 		} else {
-			long ptr = (com.ibm.oti.vm.VM.FJ9OBJECT_SIZE == 4) ?  h.getIntFromObject(anObject, 0L) : h.getLongFromObject(anObject, 0L);
+			long ptr = (com.ibm.oti.vm.VM.FJ9OBJECT_SIZE == 4) ? Integer.toUnsignedLong(h.getIntFromObject(anObject, 0L)) : h.getLongFromObject(anObject, 0L);
 			if ((ptr & com.ibm.oti.vm.VM.OBJECT_HEADER_HAS_BEEN_MOVED_IN_CLASS) != 0) {
 				if (!h.isArray(anObject)) {
 					long j9class = ptr & com.ibm.oti.vm.VM.J9_JAVA_CLASS_MASK;
--- a/runtime/compiler/trj9/codegen/J9RecognizedMethodsEnum.hpp
+++ b/runtime/compiler/trj9/codegen/J9RecognizedMethodsEnum.hpp
@@ -427,6 +427,7 @@
    java_lang_Integer_rotateLeft,
    java_lang_Integer_rotateRight,
    java_lang_Integer_valueOf,
+   java_lang_Integer_toUnsignedLong,
    java_lang_Long_bitCount,
    java_lang_Long_lowestOneBit,
    java_lang_Long_highestOneBit,
--- a/runtime/compiler/trj9/env/j9method.cpp
+++ b/runtime/compiler/trj9/env/j9method.cpp
@@ -3359,6 +3359,7 @@ TR_ResolvedJ9Method::TR_ResolvedJ9Method(TR_OpaqueMethodBlock * aMethod, TR_Fron
       {x(TR::java_lang_Integer_rotateRight,             "rotateRight",           "(II)I")},
       {x(TR::java_lang_Integer_valueOf,                 "valueOf",               "(I)Ljava/lang/Integer;")},
       {  TR::java_lang_Integer_init,              6,    "<init>", (int16_t)-1,    "*"},
+      {x(TR::java_lang_Integer_toUnsignedLong,          "toUnsignedLong",         "(I)J")},
       {  TR::unknownMethod}
       };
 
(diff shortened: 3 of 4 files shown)


andrewcraik:
Before merging this we need to fast path the Integer.toUnsignedLong in the JIT for performance otherwise we are likely to have a lot of regressions since we don't seem to have acceleration for it right now.

liqunl:
Can you change the call to `Integer.toUnsignedLong` to a bitwise operation? i.e. use
```
long ptr = (com.ibm.oti.vm.VM.FJ9OBJECT_SIZE == 4) ? ((long)h.getIntFromObject(anObject, 0L)) & 0xffffffffL : h.getLongFromObject(anObject, 0L);
```
Then we don't need any change in the JIT.

fjeremic:
> Can you change the call to Integer.toUnsignedLong to a bitwise operation?

I disagree. We should take this opportunity and just accelerate `Integer.toUnsignedLong` so all future and existing code will benefit from this.

andrewcraik:
I think it should be as simple as adding a transformer in ilgen somewhere to replace calls to Integer.toUnsignedLong with iu2l nodes

fjeremic:
@DanHeidinga let us know if you need any help crafting the JIT commit. We'll be happy to assist.

liqunl:
We can't transform it in ilgen since we need a HCR guard. I'm ok with keeping the call by force it to inline.

andrewcraik:
I think we can white list replacing it for the JITHelper since that is an internal implementation detail would be my take.

liqunl:
Personally I would prefer using bitwise operation over creating a JITHelper and fast pathing the JITHelper. I'm not sure if we should add Integer.toUnsignedLong to the alwaysWorthInlining list since it is a very small method, we should have always inlined it already. But I agree that adding it doesn't hurt.

DanHeidinga:
@fjeremic Thanks for the pointers on how to mark toUnsignedLong as alwaysWorthInlining.  Is there a -Xjit:verbose option to print the inlines?  I'd like to confirm that the method is being inlined.

Minimal test case to force compilation of J9VMInternals.fastIdentityHashCode:
```
package java.lang;
public class Test {

        int i = 0;
        for (int j = 0; j < 100000; j++) {
                i += fastHash(new Object());
        }
        System.out.println(i);

}

public static int fastHash(Object o) {
        return J9VMInternals.fastIdentityHashCode(o);
}
}

```

Running as:
```
$ java -Xbootclasspath/a:dan -Xjit:verbose java.lang.Test
```

And I'm seeing `toUnsignedLong` being compiled which I wouldn't expect (maybe?) if it was always inlined:
```
+ (cold) java/lang/J9VMInternals.fastIdentityHashCode(Ljava/lang/Object;)I @ 00007F5273D80C20-00007F5273D80DAB OrdinaryMethod - Q_SZ=2 Q_SZI=2 QW=10 j9m=00000000006BD730 bcsz=135 GCR compThread=0 CpuLoad=1%(0%avg) JvmCpu=0%
+ (warm) java/lang/Integer.toUnsignedLong(I)J @ 00007F5273D80DF8-00007F5273D80E17 OrdinaryMethod - Q_SZ=1 Q_SZI=1 QW=4 j9m=00000000006E9BC8 bcsz=7 compThread=0 CpuLoad=1%(0%avg) JvmCpu=0%
+ (warm) java/lang/Test.fastHash(Ljava/lang/Object;)I @ 00007F5273D80E60-00007F5273D80F7A OrdinaryMethod - Q_SZ=0 Q_SZI=0 QW=2 j9m=000000000078F288 bcsz=5 GCR compThread=0 CpuLoad=1%(0%avg) JvmCpu=0%818586157
```

fjeremic:
> Is there a -Xjit:verbose option to print the inlines?

Not at the moment. The easiest way to tell is to log `java/lang/Test.fastHash` and check the inlining table. You can achieve this via `-Xjit:{java/lang/Test.fastHash*}(tracefull,log=fastHash.log)`. In the generated `fastHash.log` search for "Post Optimization Trees". You should see an inlining table with caller callee relationship. In there hopefully you will see `fastIdentityHashCode`.

fjeremic:
> Is there a -Xjit:verbose option to print the inlines?

Actually @DanHeidinga this is a cool idea. @mpirvu / @dsouzai do you guys think we could do something like this? i.e. in the verbose log print an indented tree of inlining done for each compiled method. This can be guarded via a `verbose={...}` option?

fjeremic:
I'll wait for a confirmation that the inlining did happen, though I'm fairly confident it will have and given it's the weekend let's kick off preemptive builds.

Jenkins test all

mpirvu:
@DanHeidinga For verbose inlining you could try something like
`-Xjit:verbose={compilePerformance},verbose={inlining},vlog=vlog.txt`

DanHeidinga:
Attached the jit log as I'm not seeing the inlining table in the Post Optimization Trees section.  Likely an error on my end.

[fastHash.log.18460.45150.20180224.040550.18460.txt](https://github.com/eclipse/openj9/files/1754038/fastHash.log.18460.45150.20180224.040550.18460.txt)

DanHeidinga:
> I think we can white list replacing it for the JITHelper since that is an internal implementation detail would be my take.

Anything further required to address this comment?

mpirvu:
Inlining table from the jit log above
```
Post Optimization Trees: for java/lang/Test.fastHash(Ljava/lang/Object;)I

Call Stack Info
CalleeIndex CallerIndex ByteCodeIndex CalleeMethod
       0         -1           1       java/lang/J9VMInternals.fastIdentityHashCode(Ljava/lang/Object;)I
       1          0          78       java/lang/Integer.toUnsignedLong(I)J
```

DanHeidinga:
Thanks @mpirvu.  I was looking for something called inlining table.

JamesKingdon:
Just a footnote that this problem may also surface as an unexpected NPE when the memory access is in jitted code, and even as a crash in getJitRegisterMap as the faulting address is not associated with a stackmap.

PR #1260, opened 2018-02-23
Read Barrier on collected reference dereference

dsouzai:
The compiler must use read barriers when dereferencing collected references.

Depends on https://github.com/eclipse/omr/pull/2344

### DIFF
--- a/runtime/compiler/trj9/env/VMJ9.cpp
+++ b/runtime/compiler/trj9/env/VMJ9.cpp
@@ -1269,10 +1269,15 @@ TR_OpaqueClassBlock *
 TR_J9VMBase::getObjectClass(uintptrj_t objectPointer)
    {
    TR_ASSERT(haveAccess(), "Must haveAccess in getObjectClass");
-   if (TR::Compiler->om.generateCompressedObjectHeaders())
-      return (TR_OpaqueClassBlock*)((uintptrj_t)*(uint32_t*)(objectPointer + TR::Compiler->om.offsetOfObjectVftField()) & TR::Compiler->om.maskOfObjectVftField());
-   else
-      return (TR_OpaqueClassBlock*)(*(uintptrj_t*)(objectPointer + TR::Compiler->om.offsetOfObjectVftField()) & TR::Compiler->om.maskOfObjectVftField());
+   J9Class *j9class = J9OBJECT_CLAZZ(vmThread(), objectPointer);
+   return convertClassPtrToClassOffset(j9class);
+   }
+
+uintptrj_t
+TR_J9VMBase::getStaticReferenceFieldAtAddress(uintptrj_t fieldAddress)
+   {
+   TR_ASSERT(haveAccess(), "Must haveAccess in getStaticReferenceFieldAtAddress");
+   return (uintptrj_t)J9STATIC_OBJECT_LOAD(vmThread(), NULL, fieldAddress);
    }
 
 uintptrj_t
@@ -5218,8 +5223,8 @@ TR_J9VMBase::getStringFieldByName(TR::Compilation * comp, TR::SymbolReference * 
 
    TR_ASSERT(!stringRef->isUnresolved(), "don't handle unresolved constant strings yet");
 
-   j9object_t string = *(j9object_t*)stringRef->getSymbol()->castToStaticSymbol()->getStaticAddress();
-
+   uintptrj_t stringStaticAddr = (uintptrj_t)stringRef->getSymbol()->castToStaticSymbol()->getStaticAddress();
+   j9object_t string = (j9object_t)getStaticReferenceFieldAtAddress(stringStaticAddr);
 
    TR::Symbol::RecognizedField   field = fieldRef->getSymbol()->getRecognizedField();
 
@@ -5257,9 +5262,9 @@ TR_J9VMBase::getFieldOffset(TR::Compilation * comp, TR::SymbolReference* classRe
    {
    TR_ResolvedMethod* method = classRef->getOwningMethod(comp);
    TR::StaticSymbol* classSym = classRef->getSymbol()->castToStaticSymbol();
-   j9object_t classString = *(j9object_t*)classSym->getStaticAddress();
+   j9object_t classString = (j9object_t)getStaticReferenceFieldAtAddress((uintptrj_t)classSym->getStaticAddress());
    TR::StaticSymbol* fieldSym = fieldRef->getSymbol()->castToStaticSymbol();
-   j9object_t fieldString = *(j9object_t*)fieldSym->getStaticAddress();
+   j9object_t fieldString = (j9object_t)getStaticReferenceFieldAtAddress((uintptrj_t)fieldSym->getStaticAddress());
 
    int32_t len, res;
    len = (int32_t)jitConfig->javaVM->internalVMFunctions->getStringUTF8Length(vmThread(), classString);
@@ -8500,22 +8505,6 @@ TR_J9VM::getClassClassPointer(TR_OpaqueClassBlock *objectClassPointer)
 
    j9class = (J9Class *)((uintptrj_t)j9class & TR::Compiler->om.maskOfObjectVftField());
 
-   return convertClassPtrToClassOffset(j9class);
-   }
-
-
-
-TR_OpaqueClassBlock *
-TR_J9VMBase::getClassFromStatic(void *p)
-   {
-   J9Class *j9class;
-
-   if (TR::Compiler->om.generateCompressedObjectHeaders())
-      j9class = (J9Class *) *((uint32_t *) ((uintptrj_t) *((void ***)p) + (uintptrj_t) TR::Compiler->om.offsetOfObjectVftField()));
-   else
-      j9class = (J9Class *)(*((J9Class **) ((uintptrj_t) *((void ***)p) + (uintptrj_t) TR::Compiler->om.offsetOfObjectVftField())));
-
-   j9class = (J9Class *)((uintptrj_t)j9class & TR::Compiler->om.maskOfObjectVftField());
    return convertClassPtrToClassOffset(j9class);
    }
 
--- a/runtime/compiler/trj9/env/VMJ9.h
+++ b/runtime/compiler/trj9/env/VMJ9.h
@@ -504,12 +504,13 @@ public:
 
    uint32_t                   getAllocationSize(TR::StaticSymbol *classSym, TR_OpaqueClassBlock * clazz);
 
-   virtual TR_OpaqueClassBlock *getObjectClass(uintptrj_t objectPointer);
-   virtual uintptrj_t           getReferenceFieldAt(uintptrj_t objectPointer, uintptrj_t offsetFromHeader);
-   virtual uintptrj_t           getVolatileReferenceFieldAt(uintptrj_t objectPointer, uintptrj_t offsetFromHeader);
-   virtual uintptrj_t           getReferenceFieldAtAddress(uintptrj_t fieldAddress);
-   virtual uintptrj_t           getReferenceFieldAtAddress(void *fieldAddress){ return getReferenceFieldAtAddress((uintptrj_t)fieldAddress); }
-   int32_t                      getInt32FieldAt(uintptrj_t objectPointer, uintptrj_t fieldOffset);
+   TR_OpaqueClassBlock *getObjectClass(uintptrj_t objectPointer);
+   uintptrj_t           getReferenceFieldAt(uintptrj_t objectPointer, uintptrj_t offsetFromHeader);
+   uintptrj_t           getVolatileReferenceFieldAt(uintptrj_t objectPointer, uintptrj_t offsetFromHeader);
+   uintptrj_t           getReferenceFieldAtAddress(uintptrj_t fieldAddress);
+   uintptrj_t           getReferenceFieldAtAddress(void *fieldAddress){ return getReferenceFieldAtAddress((uintptrj_t)fieldAddress); }
+   uintptrj_t           getStaticReferenceFieldAtAddress(uintptrj_t fieldAddress);
+   int32_t              getInt32FieldAt(uintptrj_t objectPointer, uintptrj_t fieldOffset);
 
    int32_t getInt32Field(uintptrj_t objectPointer, char *fieldName)
       {
@@ -977,8 +978,6 @@ public:
 
    virtual bool acquireClassTableMutex();
    virtual void releaseClassTableMutex(bool);
-
-   virtual TR_OpaqueClassBlock *getClassFromStatic(void *p);
 
    // --------------------------------------------------------------------------
    // Object model
--- a/runtime/compiler/trj9/il/J9SymbolReference.cpp
+++ b/runtime/compiler/trj9/il/J9SymbolReference.cpp
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2000, 2016 IBM Corp. and others
+ * Copyright (c) 2000, 2018 IBM Corp. and others
  *
  * This program and the accompanying materials are made available under
  * the terms of the Eclipse Public License 2.0 which accompanies this
@@ -258,35 +258,38 @@ SymbolReference::getTypeSignature(int32_t & len, TR_AllocationKind allocKind, bo
                   !comp->getOption(TR_AOT) &&
                    (type == TR::Address)))
                  {
-                 void * p = symbol->getStaticAddress();
                  TR::VMAccessCriticalSection vmAccessCriticalSection(comp->fej9(),
                                                                       TR::VMAccessCriticalSection::tryToAcquireVMAccess,
                                                                       comp);
-                 if ((*(void **)p != 0) &&
-                     vmAccessCriticalSection.hasVMAccess())
+                 if (vmAccessCriticalSection.hasVMAccess())
                     {
-                    TR_OpaqueClassBlock *classOfObject = comp->fej9()->getClassFromStatic(p);
-                    const char * s = TR::Compiler->cls.classNameChars(comp, classOfObject, len);
-                    if (s && (s[0] != '['))
+                    uintptrj_t objectStaticAddress = (uintptrj_t)symbol->getStaticAddress();
+                    uintptrj_t objectRef = comp->fej9()->getStaticReferenceFieldAtAddress(objectStaticAddress);
+                    if (objectRef != 0)
                        {
-                       s = classNameToSignature(s, len, comp);
-                       }
-                    else
-                       {
-                       int32_t numParens = 0;
-                       while (s && (s[0] == '[') && (s[1] == 'L'))
+                       TR_OpaqueClassBlock *classOfObject = comp->fej9()->getObjectClass(objectRef);
+                       const char * s = TR::Compiler->cls.classNameChars(comp, classOfObject, len);
+                       if (s && (s[0] != '['))
                           {
-                          numParens++;
-                          classOfObject = comp->fe()->getComponentClassFromArrayClass(classOfObject);
-                          s = TR::Compiler->cls.classNameChars(comp, classOfObject, len);
-                           }
-                       s = classNameToSignature(s, len, comp);
-                       s = prependNumParensToSig(s, len, numParens);
-                       }
+                          s = classNameToSignature(s, len, comp);
+                          }
+                       else
+                          {
+                          int32_t numParens = 0;
+                          while (s && (s[0] == '[') && (s[1] == 'L'))
+                             {
+                             numParens++;
+                             classOfObject = comp->fe()->getComponentClassFromArrayClass(classOfObject);
+                             s = TR::Compiler->cls.classNameChars(comp, classOfObject, len);
+                              }
+                          s = classNameToSignature(s, len, comp);
+                          s = prependNumParensToSig(s, len, numParens);
+                          }
 
-                    if (isFixed)
-                       *isFixed = true;
-                    return s;
+                       if (isFixed)
+                          *isFixed = true;
+                       return s;
+                       }
                     }
                  }
                }
--- a/runtime/compiler/trj9/optimizer/J9TransformUtil.cpp
+++ b/runtime/compiler/trj9/optimizer/J9TransformUtil.cpp
@@ -477,7 +477,8 @@ J9::TransformUtil::transformStringIndexOfCall(TR::Compilation * comp, TR::Node *
    if (!transformStringIndexCriticalSection.hasVMAccess())
       return callNode;
 
-   uintptrj_t string = *(uintptrj_t*)needle->getSymbol()->castToStaticSymbol()->getStaticAddress();
+   uintptrj_t stringStaticAddr = (uintptrj_t)needle->getSymbol()->castToStaticSymbol()->getStaticAddress();
+   uintptrj_t string = comp->fej9()->getStaticReferenceFieldAtAddress(stringStaticAddr);
    int32_t len = fej9->getStringLength(string);
    if (len < 3)
       return callNode;
@@ -805,7 +806,12 @@ J9::TransformUtil::transformIndirectLoad(TR::Compilation *comp, TR::Node *node)
                      {
                      // First dereference with VM access to dig a pointer out of an object
                      TR::VMAccessCriticalSection digPointerOutOfObject(comp->fej9());
-                     uintptrj_t fieldAddress = (*baseObjectRefLocation) + firstDereference->getSymbolReference()->getOffset();
+
+                     uintptrj_t baseObjectRef = baseObject->getSymbol()->isStatic() ?
+                                                comp->fej9()->getStaticReferenceFieldAtAddress((uintptrj_t)baseObjectRefLocation) :
+                                                *baseObjectRefLocation;
+                     uintptrj_t fieldAddress = baseObjectRef + firstDereference->getSymbolReference()->getOffset();
+
                      switch (firstDereference->getDataType())
                         {
                         case TR::Int32:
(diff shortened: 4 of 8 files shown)


dsouzai:
@fjeremic @amicic @jdmpapin @andrewcraik could you please review.

dsouzai:
Do we have any concerns about compile performance? Ie, previously what was a simple dereference is now a function call that likely won't be inlined; this is probably ok when running in Concurrent Scavenge mode, but since the majority of the platforms don't even have the option of Concurrent Scavenge, is there a better approach? We could just use `J9STATIC_OBJECT_LOAD` instead of going through the FrontEnd, but that's still more work on non CS supported platforms.

fjeremic:
> Do we have any concerns about compile performance?

Best thing to do is measure on someth8ing like DayTrader7 with the change and without and validate. I don't think these sporadic indirections will affect things but let's let the data speak for itself. I can help schedule runs once this is reviewed.

runtime/compiler/trj9/env/VMJ9.cpp
```diff
@@ -8508,12 +8544,13 @@ TR_J9VM::getClassClassPointer(TR_OpaqueClassBlock *objectClassPointer)
 TR_OpaqueClassBlock *
 TR_J9VMBase::getClassFromStatic(void *p)
    {
+   TR_ASSERT(haveAccess(), "Must haveAccess in getClassFromStatic");
    J9Class *j9class;
 
    if (TR::Compiler->om.generateCompressedObjectHeaders())
-      j9class = (J9Class *) *((uint32_t *) ((uintptrj_t) *((void ***)p) + (uintptrj_t) TR::Compiler->om.offsetOfObjectVftField()));
+      j9class = (J9Class *)(*((uint32_t *) ((uintptrj_t)p + (uintptrj_t)TR::Compiler->om.offsetOfObjectVftField())));
```
fjeremic: @amicic do we have VM APIs we could be using for something like this as well?

dsouzai:
>j9accessbarrier.h is your friend. I think you are looking for J9OBJECT_U32_LOAD.

The full code is 

```
   if (TR::Compiler->om.generateCompressedObjectHeaders())
      j9class = (J9Class *)(*((uint32_t *) ((uintptrj_t)p + (uintptrj_t)TR::Compiler->om.offsetOfObjectVftField())));
   else
      j9class = (J9Class *)(*((J9Class **) ((uintptrj_t)p + (uintptrj_t)TR::Compiler->om.offsetOfObjectVftField())));
```
where `p` is a pointer to a static. Is there no API to get the j9class from a static?

amicic:
@dsouzai Ok, I better understand what you wanted. There is J9OBJECT_CLAZZ(vmThread, object) macro. Static is just a heap object, right?

runtime/compiler/trj9/env/VMJ9.cpp
```diff
@@ -1275,15 +1275,49 @@ TR_J9VMBase::getObjectClass(uintptrj_t objectPointer)
...
+TR_J9VMBase::getStaticReferenceFieldAtAddress(TR::Compilation* comp, uintptrj_t fieldAddress, TR::SymbolReference *symRef)
+   {
+   TR_ASSERT(haveAccess(), "Must haveAccess in getStaticReferenceFieldAtAddress");
+#if defined(OMR_GC_CONCURRENT_SCAVENGER)
+   // Emit read barrier
+   if (TR::Compiler->om.shouldGenerateReadBarriersForFieldLoads())
+      {
+      J9MemoryManagerFunctions * mmf = vmThread()->javaVM->memoryManagerFunctions;
+      //return (uintptrj_t)J9STATIC_OBJECT_LOAD(vmThread(), clazz, fieldAddress);
+      return (uintptrj_t)mmf->j9gc_objaccess_staticReadObject(vmThread(),
```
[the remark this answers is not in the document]
dsouzai: I was curious about this, do we not need to specify the `isVolatile` flag? The macro doesn't need it but `j9gc_objaccess_staticReadObject` does.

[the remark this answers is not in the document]
dsouzai: Also, this code is currently guarded with `#if defined(OMR_GC_CONCURRENT_SCAVENGER)` so non CS supported platforms don't run this code.

dsouzai:
This is unrelated to readbarriers, but I dislike this bit of code

```
TR_OpaqueClassBlock *
TR_J9VM::getClassClassPointer(TR_OpaqueClassBlock *objectClassPointer)
   {
   TR::VMAccessCriticalSection getClassClassPointer(this);
   J9Class *j9class;
   j9class = TR::Compiler->cls.convertClassOffsetToClassPtr(objectClassPointer);

   if (TR::Compiler->cls.classesOnHeap())
      {
      void *javaLangClass = J9VM_J9CLASS_TO_HEAPCLASS(TR::Compiler->cls.convertClassOffsetToClassPtr(objectClassPointer));

      // j9class points to the J9Class corresponding to java/lang/Object
      if (TR::Compiler->om.generateCompressedObjectHeaders())
         j9class = (J9Class *) *((uint32_t *) ((uintptrj_t) javaLangClass + (uintptrj_t) TR::Compiler->om.offsetOfObjectVftField()));
      else
         j9class = (J9Class *)(*((J9Class **) ((uintptrj_t) javaLangClass + (uintptrj_t) TR::Compiler->om.offsetOfObjectVftField())));
      }
   else
      j9class = (J9Class *)(*((J9Class **) ((uintptrj_t)j9class + (uintptrj_t) TR::Compiler->om.offsetOfObjectVftField())));

   j9class = (J9Class *)((uintptrj_t)j9class & TR::Compiler->om.maskOfObjectVftField());

   return convertClassPtrToClassOffset(j9class);
   }
```

I'd like to replace it with 

```
TR_OpaqueClassBlock *
TR_J9VM::getClassClassPointer(TR_OpaqueClassBlock *objectClassPointer)
   {
   TR::VMAccessCriticalSection getClassClassPointer(this);

   void *javaLangClass = J9VM_J9CLASS_TO_HEAPCLASS(TR::Compiler->cls.convertClassOffsetToClassPtr(objectClassPointer));
  
   return convertClassPtrToClassOffset(J9OBJECT_CLAZZ(vmThread(), javaLangClass));
   }
```

However, I'm not sure if it's ok to do so. `TR::Compiler->cls.classesOnHeap()` always returns true. Will this change in the future? Also, if, for the sake of argument, `TR::Compiler->cls.classesOnHeap()` returned false, is `j9class = (J9Class *)(*((J9Class **) ((uintptrj_t)j9class + (uintptrj_t) TR::Compiler->om.offsetOfObjectVftField())));` even valid? @amicic @0xdaryl

amicic:
@dsouzai The later else clause does not seem right to me (trying to interpret J9Class same as a class object on heap). 

More importantly, If I'm reading this right, this effectively return J9Class *  pointer for java/lang/Class. If so, there might be better ways to do it (like using find-known-classes API)....

If so, the question if class on heap is always holds would be irrelevant.

dsouzai:
>More importantly, If I'm reading this right, this effectively return J9Class * pointer for java/lang/Class. If so, there might be better ways to do it (like using find-known-classes API).... and the question if class on heap always holds would be irrelevant.

Hm I see, ok I'll do that in a different PR then.

runtime/compiler/trj9/env/VMJ9.cpp
```diff
@@ -1269,21 +1269,47 @@ TR_OpaqueClassBlock *
...
+      {
+      return (uintptrj_t)J9STATIC_OBJECT_LOAD(vmThread(), symRef->getOwningMethod(comp)->containingClass(), fieldAddress);
+      }
    else
-      return (TR_OpaqueClassBlock*)(*(uintptrj_t*)(objectPointer + TR::Compiler->om.offsetOfObjectVftField()) & TR::Compiler->om.maskOfObjectVftField());
+#endif
+      {
+      return *(uintptrj_t *)fieldAddress;
+      }
+   }
```
amicic: I'd simplify this even further and only call J9STATIC_OBJECT_LOAD

runtime/compiler/trj9/env/VMJ9.cpp
```diff
@@ -1269,21 +1269,47 @@ TR_OpaqueClassBlock *
...
+TR_J9VMBase::getReferenceAtAddress(uintptrj_t address)
+   {
+   TR_ASSERT(haveAccess(), "Must haveAccess in getReferenceFieldAtAddress");
+#if defined(OMR_GC_CONCURRENT_SCAVENGER)
+   // Emit read barrier
+   if (TR::Compiler->om.shouldGenerateReadBarriersForFieldLoads())
+      vmThread()->javaVM->memoryManagerFunctions->J9ReadBarrier(vmThread(), (fj9object_t *)address);
+#endif
+
+   return *(uintptrj_t*)address;
```
amicic: same thing here, just call J9OBJECT_OBJECT_LOAD (which will do the barrier and the de-reference)

runtime/compiler/trj9/env/VMJ9.cpp
```diff
@@ -1283,12 +1280,7 @@ TR_J9VMBase::getStaticReferenceFieldAtAddress(TR::Compilation* comp, uintptrj_t
    // Emit read barrier
    if (TR::Compiler->om.shouldGenerateReadBarriersForFieldLoads())
       {
-      J9MemoryManagerFunctions * mmf = vmThread()->javaVM->memoryManagerFunctions;
-      //return (uintptrj_t)J9STATIC_OBJECT_LOAD(vmThread(), symRef->getOwningMethod(comp)->containingClass(), fieldAddress);
-      return (uintptrj_t)mmf->j9gc_objaccess_staticReadObject(vmThread(),
-                                                              (J9Class *)symRef->getOwningMethod(comp)->containingClass(),
-                                                              (j9object_t *)fieldAddress,
-                                                              (UDATA)symRef->getSymbol()->isVolatile());
+      return (uintptrj_t)J9STATIC_OBJECT_LOAD(vmThread(), symRef->getOwningMethod(comp)->containingClass(), fieldAddress);
```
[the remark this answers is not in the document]
jdmpapin: Separately, did you mean to update this function in the first commit of the series?

runtime/compiler/trj9/env/VMJ9.cpp
amicic: J9OBJECT_OBJECT_LOAD would do de-compressing of the address, so that above comment applies to getReferenceFieldAtAddress. I'm a bit confused with helper....

runtime/compiler/trj9/optimizer/J9TransformUtil.cpp
```diff
@@ -1135,7 +1147,16 @@ bool
...
+   void *baseAddress;
+   if (baseExpression->getOpCode().hasSymbolReference() && baseExpression->getSymbol()->isStatic())
+      {
+      baseAddress = (void *)comp->fej9()->getStaticReferenceFieldAtAddress(comp, (uintptrj_t)baseReferenceLocation, baseExpression->getSymbolReference());
+      }
+   else
+      {
+      baseAddress = (void *)comp->fej9()->getReferenceAtAddress((uintptrj_t)baseReferenceLocation);
+      }
+   bool result = TR::TransformUtil::transformIndirectLoadChainImpl(comp, node, baseExpression, baseAddress, removedNode);
```
jdmpapin: Depending on the caller, `*baseReferenceLocation` may be the address of e.g. a `J9Class` rather than an object. Is it safe to run the read barrier in that case?

runtime/compiler/trj9/optimizer/J9ValuePropagation.cpp
```diff
@@ -406,7 +406,8 @@ J9::ValuePropagation::constrainRecognizedMethod(TR::Node *node)
                                                         TR::VMAccessCriticalSection::tryToAcquireVMAccess);
             if (!getStringlength.hasVMAccess())
                break;
-            len = comp()->fej9()->getStringLength(*stringLocation);
+            uintptrj_t stringObject = comp()->fej9()->getStaticReferenceFieldAtAddress(comp(), (uintptrj_t)stringLocation, receiverChild->getSymbolReference());
```
jdmpapin: Here `receiverChild->getSymbolReference()` might not be the right symbol reference - e.g. it could be an auto.

If `receiverChildConstraint` is a `VPKnownObject` instead of a `VPConstString`, then there might not even be a static symbol reference that corresponds to the string

runtime/compiler/trj9/il/J9SymbolReference.cpp
```diff
@@ -258,14 +258,16 @@ SymbolReference::getTypeSignature(int32_t & len, TR_AllocationKind allocKind, bo
...
                    (type == TR::Address)))
                  {
-                 void * p = symbol->getStaticAddress();
+                 uintptrj_t objectStaticAddress = (uintptrj_t)symbol->getStaticAddress();
                  TR::VMAccessCriticalSection vmAccessCriticalSection(comp->fej9(),
                                                                       TR::VMAccessCriticalSection::tryToAcquireVMAccess,
                                                                       comp);
-                 if ((*(void **)p != 0) &&
+
+                 uintptrj_t objectRef = comp->fej9()->getStaticReferenceFieldAtAddress(comp, objectStaticAddress, (TR::SymbolReference *)this);
```
jdmpapin: VM access might not be held here IIUC

runtime/compiler/trj9/env/VMJ9.cpp
jdmpapin: It looks like `J9ReadBarrier` is used for possibly compressed references, as they occur within objects, but this function is used for uncompressed references only

amicic: Ok, I realized that J9OBJECT_OBJECT_LOAD takes an offset (as opposed to an absolute address of the field), so it could not be used to implement getReferenceFieldAtAddress. 

But I'm still confused by this helper. So it returns potentially a compressed value (fj9object_t). It's typically not how VM and GC code works, where locals and method arguments (for heap object addresses) are in decompressed form. Only slots in heap would really be in compressed form. But if this works for you, ok...

Still, there is concern for the input argument. By the analogy, I'm guessing it would also be in compressed form, but it's not what J9ReadBarrier expects as an address - it expects uncompressed form.

dsouzai: It looks like in all places where I use this helper, previously it was just a simple dereference. @jdmpapin and I had a discussion about it, and it seems like that's ok because they're JNI refs, which are always decompressed. Therefore, the use of `J9ReadBarrier` here is definitely incorrect. 

With that said however, we still need a way of dereferencing a pointer to an uncompressed object ref. After looking through the implementation of `J9STATIC_OBJECT_LOAD`, it looks like the `J9Class` parameter is never used. This brings up two questions:

1. Can we pass in NULL for the J9Class parameter?
2. Can we use J9STATIC_OBJECT_LOAD to dereference a pointer to an uncompressed object ref?

runtime/compiler/trj9/env/VMJ9.cpp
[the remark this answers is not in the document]
dsouzai: >It seems surprising that the containingClass() would be relevant to the read barrier. Is it used for anything? If not, is there any need for the distinction between this method and getReferenceAtAddress()? i.e. is there special treatment in the read barrier for static fields, as opposed to any other references that are guaranteed not to be compressed?

I suppose Alek's answer to the question asked above will answer this; given that the J9Class isn't used, it seems like there shouldn't be a distinction between the two.

> Separately, did you mean to update this function in the first commit of the series?

Ah damn, yes I did >.<

runtime/compiler/trj9/optimizer/J9TransformUtil.cpp
dsouzai: I could be wrong, but because this is an address that does not point to the heap (and not a 64-bit integer who's value might be an address into the heap), performing a read barrier shouldn't cause any problems. However, it's probably best confirmed by @amicic .

runtime/compiler/trj9/il/J9SymbolReference.cpp
dsouzai: Yeah you're right. The reason I left it as is was because:

1. If we failed to acquire vm access, the next condition in the `if` will fail, so the object getting moved twice doesn't matter.
2. if we acquired vm access, the code executes as expected.

However, now that I think of it, it might not be ok to execute the read barrier without VM Access, so it's probably for the best to switch the order of the conditions.

runtime/compiler/trj9/optimizer/J9ValuePropagation.cpp
dsouzai: Hm I see; well since the symref is really only used to get the class, if we don't need to provide a class to the read barrier api, this will become a non-issue.

runtime/compiler/trj9/il/J9SymbolReference.cpp
jdmpapin: > it might not be ok to execute the read barrier without VM Access

Yeah, I'm not sure whether that's safe either, but I think `getStaticReferenceFieldAtAddress` wants to require VM access

runtime/compiler/trj9/env/VMJ9.cpp
amicic: 1) you can as far as Concurrent Scavenger read barrier is concerned. It would work now, but not necessarily in future.
2) It would work, but definitely it would be abuse of API. We can create a new one (which would indeed do the same thing).

You don't have to introduce the barrier for JNI slots, yet. VM knows about all of them and they are treated as hard roots in CS (will be updated during root scanning in first STW phase), or if they are overwritten meanwhile with another value, that value is guarantied to be already up to date (like to would be a case with any other hard root slot).

However, this is not optimal behavior. Some JNI slots could be weak refs and should not be treated as hard roots (we leak this memory in CS, till next global GC.) This is just a temporary solution in CS, before we identify all JNI slots and update the reads with appropriate API - we actually have a lot of them using raw reads even in core VM.

So, anything related to JNI slots should be a separate change, and should wait for a new API to be introduced. What we need sooner is to identify accesses to class statics.

dsouzai: > you can as far as Concurrent Scavenger read barrier is concerned. It would work now, but not necessarily in future.

For clarification, would the `J9class` parameter be the class of the object who's static field we're dereferencing, or is it the class of the static object?

As @jdmpapin pointed out in a comment below:
>Here receiverChild->getSymbolReference() might not be the right symbol reference - e.g. it could be an auto.
>
>If receiverChildConstraint is a VPKnownObject instead of a VPConstString, then there might not even be a static symbol reference that corresponds to the string

The symbol reference is used to get the class by getting the owning method. This might be ok if the symref refers to an auto, but in the second case, if we don't have a symref, we won't be able to get the `J9Class`. What is the potential benefit of knowing what the `J9Class` is?

amicic: > For clarification, would the J9class parameter be the class of the object who's static field we're dereferencing, or is it the class of the static object?

It's the former.

It's hard to speculate what it could be used in future. 

If you don't have an easy way to find the class pointer, passing null should be ok.

jdmpapin: > What we need sooner is to identify accesses to class statics

Are string literals counted as class statics for this purpose? Or are those hard roots as well?

> would the J9class parameter be the class of the object who's static field we're dereferencing, or is it the class of the static object? [...] The symbol reference is used to get the class by getting the owning method.

I don't believe the class of the symref's owning method is necessarily either of these. It should be the class containing the access, e.g. consider these classes:

    class Foo {
        public static final Object BAR = [...];
    }

    class Baz {
        public Object quux() {
            return Foo.BAR;
        }
    }

If we're looking at `aload Foo.BAR` feeding the return, the symref corresponds to the symbolic reference `Foo.BAR` in the constant pool of `Baz`. Its owning method should be `Baz.quux`

amicic: > Are string literals (edit: and call site table entries, further edit: and method type table entries) counted as class statics for this purpose? Or are those hard roots as well?

Short answer is, we just need those that access J9Class::ramStatics. That's the minimum required for Concurrent Scavenger to work. I'm not completely sure why others are not needed (my claim is based on looking on Scavenger code that does class specific scanning in hard roots), but I believe all others are in Tenure space already (perhaps because they rarely or never die or get modified) and not subject to movement during CS.

Long answer is, we need all, because eventually concurrent copying logic can be part of other GC policies where any object anywhere in the heap will be subject to movement

dsouzai:
Thanks for all the review and information; I'm going to change my commits appropriately and request a review again once I'm done.

dsouzai:
I updated the commit to only deal with static fields. Could you guys please re-review.

amicic:
Looks good to me, but of course the tricky part is if you covered them all and vice versa if those you  covered are really static fields, since for at least one of them it's not so obvious.

runtime/compiler/trj9/env/VMJ9.h
```diff
@@ -504,12 +504,13 @@ class TR_J9VMBase : public TR_FrontEnd
...
-   virtual uintptrj_t           getVolatileReferenceFieldAt(uintptrj_t objectPointer, uintptrj_t offsetFromHeader);
-   virtual uintptrj_t           getReferenceFieldAtAddress(uintptrj_t fieldAddress);
-   virtual uintptrj_t           getReferenceFieldAtAddress(void *fieldAddress){ return getReferenceFieldAtAddress((uintptrj_t)fieldAddress); }
-   int32_t                      getInt32FieldAt(uintptrj_t objectPointer, uintptrj_t fieldOffset);
+   TR_OpaqueClassBlock *getObjectClass(uintptrj_t objectPointer);
+   uintptrj_t           getReferenceFieldAt(uintptrj_t objectPointer, uintptrj_t offsetFromHeader);
+   uintptrj_t           getVolatileReferenceFieldAt(uintptrj_t objectPointer, uintptrj_t offsetFromHeader);
+   uintptrj_t           getReferenceFieldAtAddress(uintptrj_t fieldAddress);
+   uintptrj_t           getReferenceFieldAtAddress(void *fieldAddress){ return getReferenceFieldAtAddress((uintptrj_t)fieldAddress); }
+   uintptrj_t           getStaticReferenceFieldAtAddress(uintptrj_t fieldAddress, TR_OpaqueClassBlock *clazz = NULL);
```
andrewcraik: This API worries me - the VM API seems to do nothing with the clazz it is passed and we are omitting sending the class through because it is hard to find/we don't always have it. Do we really need to pass the class through? If it is a future potential use case where we might need it why not keep the API simpler until we do need it? As it is the class clutters the code and is not being checked so is giving the false sense of security that it is being generated and that it is even right. Unused code rots and confuses things so wouldn't it be cleaner to just drop the pretense for now?

fjeremic: 100% agree. This is the exact same conclusion as in https://github.com/eclipse/openj9/pull/1047#discussion_r171374224. We should try not to speculate future APIs needed to do something. Simplify things now and let this future person introduce the parameter if they actually need to use it. This future use may not ever materialize and we'll end up paying the technical debt for a long time.

amicic: I'm on a fence. The code lived so long with treating the static reference as a raw pointer, and this is a step forward. On the other side, I do indeed see no obvious (other that some debug/redundancy check) scenario where the class pointer is really relevant. 

@charliegracie, @DanHeidinga  What do you think:
1) about static field access API:
  J9STATIC_OBJECT_LOAD/STORE(thread, clazz, address),
where address is an absolute address of the static.
Is/could class pointer be relevant?
2) if not relevant, how about simplifying it (or providing just another generic API without object/class ?
3) about keeping API and providing NULL, where class pointer is not trivial to provide?

runtime/compiler/trj9/il/J9SymbolReference.cpp
```diff
@@ -258,35 +258,38 @@ SymbolReference::getTypeSignature(int32_t & len, TR_AllocationKind allocKind, bo
...
+                    uintptrj_t objectRef = comp->fej9()->getStaticReferenceFieldAtAddress(objectStaticAddress);
+                    if (objectRef != 0)
                        {
-                       s = classNameToSignature(s, len, comp);
-                       }
-                    else
-                       {
-                       int32_t numParens = 0;
-                       while (s && (s[0] == '[') && (s[1] == 'L'))
+                       TR_OpaqueClassBlock *classOfObject = comp->fej9()->getClassFromStatic((void *)objectRef);
```
jdmpapin: Should this be `getClassFromStatic((void*)objectStaticAddress)`?

dsouzai: No; previously getClassFromStatic would dereference the object location. However, because the call to getClassFromStatic requires checking if the dereference is NULl, I changed getClassFromStatic to take in the dereferenced object pointer instead (https://github.com/eclipse/openj9/pull/1260/files#diff-cba9fe3d00eac2598c34d768e9173980R8516)

jdmpapin: In that case `getClassFromStatic` might now be redundant with `getObjectClass`

dsouzai: Hm I think you're right. I guess I can get rid of `getClassFromStatic` and replace the contents of `getObjectClass` with `J9OBJECT_CLAZZ`

dsouzai:
The OMR dependent changes have been approved. @fjeremic @jdmpapin could you guys review these changes?

andrewcraik:
Jenkins test sanity

dsouzai:
I'm pretty sure the OpenJ9 Builds are going to be broken until https://github.com/eclipse/omr/pull/2344 is merged.

PR #1262, opened 2018-02-23
[Nest mates] Adds JVMTI nestmates tests for class redefinition

taliamccormick:
Nestmates spec disallows for changes in nest mates
attributes during class redefinition.

See: https://github.com/eclipse/openj9/issues/1165

### DIFF
--- a/test/cmdLineTests/jvmtitests/src/com/ibm/jvmti/tests/nestMatesRedefinition/ClassGenerator.java
+++ b/test/cmdLineTests/jvmtitests/src/com/ibm/jvmti/tests/nestMatesRedefinition/ClassGenerator.java
@@ -1,0 +1,119 @@
+/*******************************************************************************
+ * Copyright (c) 2018, 2018 IBM Corp. and others
+ *
+ * This program and the accompanying materials are made available under
+ * the terms of the Eclipse Public License 2.0 which accompanies this
+ * distribution and is available at https://www.eclipse.org/legal/epl-2.0/
+ * or the Apache License, Version 2.0 which accompanies this distribution and
+ * is available at https://www.apache.org/licenses/LICENSE-2.0.
+ *
+ * This Source Code may also be made available under the following
+ * Secondary Licenses when the conditions for such availability set
+ * forth in the Eclipse Public License, v. 2.0 are satisfied: GNU
+ * General Public License, version 2 with the GNU Classpath
+ * Exception [1] and GNU General Public License, version 2 with the
+ * OpenJDK Assembly Exception [2].
+ *
+ * [1] https://www.gnu.org/software/classpath/license.html
+ * [2] http://openjdk.java.net/legal/assembly-exception.html
+ *
+ * SPDX-License-Identifier: EPL-2.0 OR Apache-2.0 OR GPL-2.0 WITH Classpath-exception-2.0 OR LicenseRef-GPL-2.0 WITH Assembly-exception
+ *******************************************************************************/
+
+package com.ibm.jvmti.tests.nestMatesRedefinition;
+
+import org.objectweb.asm.*;
+import org.objectweb.asm.ClassWriter;
+import org.objectweb.asm.FieldVisitor;
+import org.objectweb.asm.MethodVisitor;
+import org.objectweb.asm.Opcodes;
+
+public class ClassGenerator implements Opcodes {
+	
+	public static byte[] generateClass(String name, RedefinitionPhase phase, Attribute[] attributes) {
+		ClassWriter classWriter = new ClassWriter(ClassWriter.COMPUTE_MAXS | ClassWriter.COMPUTE_FRAMES);
+		classWriter.visit(V1_8, ACC_PUBLIC | ACC_SUPER, name, null, "java/lang/Object", null);
+		
+		switch(phase) {
+		case ORIGINAL:
+			/* Original has the source attribute */
+			classWriter.visitSource(name + ".java", null);	
+			break;
+		case REDEFINE:
+			/* REDEFINE has no source attribute to ensure the VM can't skip the redefine */
+			break;
+		default:
+			throw new IllegalArgumentException("Phase must be set correctly");
+		}
+		
+		/* Generate base constructor which just calls super.<init>() */
+		MethodVisitor methodVisitor = classWriter.visitMethod(ACC_PUBLIC, "<init>", "()V", null, null);
+		methodVisitor.visitCode();
+		methodVisitor.visitVarInsn(ALOAD, 0);
+		methodVisitor.visitMethodInsn(INVOKESPECIAL, "java/lang/Object", "<init>", "()V", false);
+		methodVisitor.visitInsn(RETURN);
+		methodVisitor.visitMaxs(1, 1);
+		methodVisitor.visitEnd();
+
+		/* Generate passed in attributes if there are any */
+		if (null != attributes) {
+			for (Attribute attr : attributes) {
+				classWriter.visitAttribute(attr);
+			}
+		}
+
+		classWriter.visitEnd();
+		return classWriter.toByteArray();
+	}
+
+	static byte[] testClassNoAttribute () throws Exception {
+		return generateClass("TestClass", RedefinitionPhase.ORIGINAL, null);
+	}
+	
+	static byte[] testClassNoAttributeForceRedefinition () throws Exception {
+		return generateClass("TestClass", RedefinitionPhase.REDEFINE, null);
+	}
+
+	static byte[] testClassWithNestMembersAttribute () throws Exception {
+		String[] nestMembers = {"A", "B", "C"};
+		NestMembersAttribute attr = new NestMembersAttribute(nestMembers);
+		return generateClass("TestClass", RedefinitionPhase.ORIGINAL, new Attribute[]{ attr });
+	}
+	
+	static byte[] testClassWithNestMembersAttributeForceRedefinition () throws Exception {
+		String[] nestMembers = {"A", "B", "C"};
+		NestMembersAttribute attr = new NestMembersAttribute(nestMembers);
+		return generateClass("TestClass", RedefinitionPhase.REDEFINE, new Attribute[]{ attr });
+	}
+
+	static byte[] nestMemberNoAttributeForceRedefinition () throws Exception {
+		return generateClass("TestClass", RedefinitionPhase.REDEFINE, null);
+	}
+
+	static byte[] testClassWithNestHostAttribute () throws Exception {
+		NestHostAttribute attr = new NestHostAttribute("host");
+		return generateClass("TestClass", RedefinitionPhase.ORIGINAL, new Attribute[]{ attr });
+	}
+	
+	static byte[] testClassWithNestHostAttributeForceRedefinition () throws Exception {
+		NestHostAttribute attr = new NestHostAttribute("host");
+		return generateClass("TestClass", RedefinitionPhase.REDEFINE, new Attribute[]{ attr });
+	}
+
+	static byte[] testClassWithAlteredNestMembersAttributeData () throws Exception {
+		String[] nestMembers = {"D", "E", "F"};
+		NestMembersAttribute attr = new NestMembersAttribute(nestMembers);
+		return generateClass("TestClass", RedefinitionPhase.ORIGINAL, new Attribute[]{ attr });
+	}
+
+	static byte[] testClassWithAlteredNestMembersAttributeLength () throws Exception {
+		String[] nestMembers = {"A", "B"};
+		NestMembersAttribute attr = new NestMembersAttribute(nestMembers);
+		return generateClass("TestClass", RedefinitionPhase.ORIGINAL, new Attribute[]{ attr });
+	}
+
+	static byte[] testClassWithAlteredNestHostAttribute () throws Exception {
+		NestHostAttribute attr = new NestHostAttribute("AlteredHostName");
+		return generateClass("TestClass", RedefinitionPhase.ORIGINAL, new Attribute[]{ attr });
+	}
+}
--- a/test/cmdLineTests/jvmtitests/src/com/ibm/jvmti/tests/nestMatesRedefinition/NestHostAttribute.java
+++ b/test/cmdLineTests/jvmtitests/src/com/ibm/jvmti/tests/nestMatesRedefinition/NestHostAttribute.java
@@ -1,0 +1,78 @@
+/*******************************************************************************
+ * Copyright (c) 2018, 2018 IBM Corp. and others
+ *
+ * This program and the accompanying materials are made available under
+ * the terms of the Eclipse Public License 2.0 which accompanies this
+ * distribution and is available at https://www.eclipse.org/legal/epl-2.0/
+ * or the Apache License, Version 2.0 which accompanies this distribution and
+ * is available at https://www.apache.org/licenses/LICENSE-2.0.
+ *
+ * This Source Code may also be made available under the following
+ * Secondary Licenses when the conditions for such availability set
+ * forth in the Eclipse Public License, v. 2.0 are satisfied: GNU
+ * General Public License, version 2 with the GNU Classpath
+ * Exception [1] and GNU General Public License, version 2 with the
+ * OpenJDK Assembly Exception [2].
+ *
+ * [1] https://www.gnu.org/software/classpath/license.html
+ * [2] http://openjdk.java.net/legal/assembly-exception.html
+ *
+ * SPDX-License-Identifier: EPL-2.0 OR Apache-2.0 OR GPL-2.0 WITH Classpath-exception-2.0 OR LicenseRef-GPL-2.0 WITH Assembly-exception
+ *******************************************************************************/
+
+package com.ibm.jvmti.tests.nestMatesRedefinition;
+
+import org.objectweb.asm.Attribute;
+import org.objectweb.asm.ByteVector;
+import org.objectweb.asm.ClassWriter;
+import org.objectweb.asm.Label;
+
+/*
+ * Nestmates attributes are a future java feature and, as such, the ASM6 API
+ * does not provide built-in methods for them. However, ASM6 allows the
+ * subclassing and use of its Attribute class. This can be leveraged to create
+ * a NestHostAttribute in order to write nest host attribute data.
+ * 
+ * Documentation on the ASM6 attribute class can be found here:
+ * http://asm.ow2.org/asm60/javadoc/user/org/objectweb/asm/Attribute.html
+ * Nestmates utilities are expected to be available in ASM7.
+ * 
+ * Nest hosts attributes have the format:
+ *		NestHost_attribute { 
+ *			u2 attribute_name_index; 
+ *			u4 attribute_length;
+ *			u2 host_class_index;
+ *		}
+ *
+ * Note that the spec explicitly disallows having more than one nestmates
+ * attribute (nest host attribute or nest members attribute) but that use
+ * of this workaround permits defining multiple nest attributes for a
+ * class.
+ */
+final class NestHostAttribute extends Attribute {
+	private String nestHost;
+
+	public Label[] getLabels(){
+		return null;
+	}
+	
+	public NestHostAttribute(String nestTop){
+		super("NestHost");
+		this.nestHost = nestTop;
+	}
+	
+	public boolean isCodeAttribute(){
+		return false;
+	}
+
+	public boolean isUnknown(){
+		return true;
+	}
+	
+	public ByteVector write(ClassWriter cw, byte[] code, int len, int maxStack, int maxLocals){
+		int topclass_index = cw.newClass(nestHost);		
+		ByteVector b = new ByteVector();
+		b.putShort(topclass_index);
+		return b;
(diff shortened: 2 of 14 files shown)


taliamccormick:
@tajila @DanHeidinga @smlambert 

Opening WIP PR to request discussion on how to handle conditional-includes of nestmates jvmti tests.

Quick summary:
- Nestmates are a future feature and their tests should not be run against java8/java9 builds
- Nestmates attributes changes are disallowed during class redefinition; relevant test cases would fall under the jvmti tests
- Build flags are enabled within the C/C++ code of the jvmti tests, but the JCL preprocessor does not handle the flags within the Java code of the jvmti tests
---
Current ideas:
1. Add an extra native method `isNestmatesEnabled` and add a check to the start of every test to see if nestmates are enabled.

2. Wrap build flags around C/C++ code, compile the java code, but exclude tests from being run within the excludes files

3. Handle class redefinition via the java.lang.instrument and java agent utilities and put it in the Valhalla test project

test/cmdLineTests/jvmtitests/src/com/ibm/jvmti/tests/nestMatesRedefinition/nmr001.c
```diff
@@ -0,0 +1,81 @@
...
+		return JNI_FALSE;
+	}
+	
+	(*jni_env)->GetByteArrayRegion(jni_env, classBytes, 0, classBytesSize, class_bytes); 
+
+	classdef.class_bytes = (unsigned char *) class_bytes;
+	classdef.class_byte_count = classBytesSize;
+	classdef.klass = originalClass;
+	
+    err = (*jvmti_env)->RedefineClasses(jvmti_env, 1, &classdef);
```
tajila: fix tabs

test/cmdLineTests/jvmtitests/src/com/ibm/jvmti/tests/nestMatesRedefinition/nmr001.c
```diff
@@ -0,0 +1,81 @@
...
+    (*jvmti_env)->Deallocate(jvmti_env, (unsigned char *) class_bytes);
+    if (err != JVMTI_ERROR_NONE) {
+    	error(env, err, "RedefineClasses failed");
+    	return JNI_FALSE;
+    }
+		
+	return JNI_TRUE;
```
tajila: can you return the error code instead, need to verify that it is failing the correct way

tajila:
@smlambert and @DanHeidinga Any objections to committing new jvmti tests that are excluded from test jobs until the nestmates spec is turned on?

test/cmdLineTests/jvmtitests/src/com/ibm/jvmti/tests/nestMatesRedefinition/nmr001.c
taliamccormick: Will do.

It looks like the other jvmti test native functions only ever return `JNI_TRUE`/`JNI_FALSE` (jboolean) or `JNI_ERR`/`JNI_OK` (jint) - even in the other redefine classes tests, but print the JVMTI. Is there a likely reason that the other test cases don't usually bother checking the type of jvmti error?

tajila: Depends what the test is trying to do. In some cases all that matters is that the function succeeds or fails. In other cases the test is looking for a certain type of error. See mt001.java

test/cmdLineTests/jvmtitests/src/com/ibm/jvmti/tests/nestMatesRedefinition/nmr001.c
```diff
@@ -0,0 +1,81 @@
...
+	jvmtiClassDefinition classdef;
+	jvmtiError err; 
+	
+	err = (*jvmti_env)->Allocate(jvmti_env, classBytesSize, (unsigned char **) &class_bytes);
+	if (err != JVMTI_ERROR_NONE) {
+		error(env, err, "Unable to allocate temp buffer for the class file");
+		return JNI_FALSE;
+	}
+	
+	(*jni_env)->GetByteArrayRegion(jni_env, classBytes, 0, classBytesSize, class_bytes);
```
DanHeidinga: This should have an exception check afterwards before calling other jni or jvmti calls.

test/cmdLineTests/jvmtitests/src/com/ibm/jvmti/tests/nestMatesRedefinition/nmr001.java
```diff
@@ -0,0 +1,226 @@
...
+	 * 
+	 * @return	true is redefinition successful; false otherwise
+	 */
+	public static native boolean redefineClass(Class name, int classBytesSize, byte[] classBytes);
+
+	/* 
+	 * Valid class definitions have the same nestmates attributes as the original
+	 * definition - either no nestmates attribute, a nest host attribute with the 
+	 * same nest host, or a nest members attribute with the same nest members in 
+	 * the same order.
```
DanHeidinga: Can you add an `@return` to the javadoc?  Here and for the other methods.

test/cmdLineTests/jvmtitests/src/com/ibm/jvmti/tests/nestMatesRedefinition/nmr001.java
```diff
@@ -0,0 +1,226 @@
...
+	 */
+	public boolean testRedefinitionNoNestAttributes() {
+		try {
+			boolean successfulRedefinition = false;
+			CustomClassLoader classloader = new CustomClassLoader();
+			byte[] bytes = ClassGenerator.nestHostNoAttribute();
+			Class<?> clazz = classloader.getClass("NestHost", bytes);
+			Object instance = clazz.getDeclaredConstructor().newInstance();
+			successfulRedefinition = redefineClass(clazz, bytes.length, bytes);
```
DanHeidinga: Is this redefining the class to the same bytes?

taliamccormick: Yes - I assume that redefinition is valid because it does not change method signatures, nest attributes, or any other components which are disallowed from being changed.
Would using the same class bytes be expected to cause an issue (eg. redefinition notices that there is no change & does not redefine) that I did not see?
Should I remove this test case, since it is likely caught by the other class redefinition test cases, none of which have nestmates attributes?

test/cmdLineTests/jvmtitests/src/com/ibm/jvmti/tests/nestMatesRedefinition/ClassGenerator.java
```diff
@@ -0,0 +1,197 @@
...
+import org.objectweb.asm.ClassWriter;
+import org.objectweb.asm.FieldVisitor;
+import org.objectweb.asm.MethodVisitor;
+import org.objectweb.asm.Opcodes;
+
+
+public class ClassGenerator implements Opcodes {
+
+	public static byte[] nestHostNoAttribute () throws Exception {
+		ClassWriter classWriter = new ClassWriter(0);
```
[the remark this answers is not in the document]
taliamccormick: **Relevant to later ClassGenerator format comments

When creating the class generator, I used the ASMifier asm utility to read in a compiled .class and produce the asm code to generate it. I then manually added nest host and nestmates attributes and neatened the code (indenting, removing unused variables). However, I had not changed constructor arguments, the location of variable definitions, or scopes. Will correct these now.

test/cmdLineTests/jvmtitests/src/com/ibm/jvmti/tests/nestMatesRedefinition/nmr001.java
DanHeidinga: A clever JVM could treat this as a no-op and skip the redefinition.  We don't today but if we're clever enough, we might in the future.

A more robust test would modify the bytecodes to force the redefinition.

test/cmdLineTests/jvmtitests/src/com/ibm/jvmti/tests/nestMatesRedefinition/NestMembersAttribute.java
```diff
@@ -0,0 +1,90 @@
...
+
+	public ByteVector write(ClassWriter cw,
+			byte[] code, 	int len,
+			int maxStack, 	int maxLocals){
+
+		ByteVector b = new ByteVector();
+		
+		b.putShort(nestMembers.length);
+		
+		int nestMemberIndex;
```
DanHeidinga: this can be declared inside the loop.

taliamccormick:
Related VM changes here: https://github.com/eclipse/openj9/pull/1412

@tajila @DanHeidinga @smlambert Am now ready for review

test/cmdLineTests/jvmtitests/src/com/ibm/jvmti/tests/nestMatesRedefinition/ClassGenerator.java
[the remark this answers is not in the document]
taliamccormick: When compiling locally, the compiler didn't recognize `COMPUTE_FRAMES` and `COMPUTE_MAXS` flags - left it as-is.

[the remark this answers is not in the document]
DanHeidinga: Did you reference them with a static import or by `ClassWriter.COMPUTE_FRAMES`?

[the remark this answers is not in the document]
taliamccormick: Updating with ClassWriter.COMPUTE_MAXS now.

Choosing COMPUTE_MAXS over COMPUTE_FRAMES because the classwriter constructor only uses one of them and it checks for the COMPUTE_FRAMES flag first and the class generator code here uses `methodVisitor.visitMaxs` (flag passed from classwriter constructor to methodvisitor constructor)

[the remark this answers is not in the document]
DanHeidinga: The two values can be `|` together and both passed in.  I'm pretty sure you'll find that pattern throughout the test code.

DanHeidinga:
TODO: pre-emptively add a `test/cmdLineTests/jvmtitests/jvmtitests_excludesSE10.xml` to exclude this in jdk10

taliamccormick:
I added the `test/cmdLineTests/jvmtitests/jvmtitests_excludesSE10.xml` in a new commit - should I keep it on this PR or make a separate one?

test/cmdLineTests/jvmtitests/src/com/ibm/jvmti/tests/nestMatesRedefinition/nmr001.java
```diff
@@ -0,0 +1,294 @@
...
+	 */
+	public boolean testRedefinitionWithNestMembersAttribute() {
+		int successfulRedefinition = 0;
+		try {
+			CustomClassLoader classloader = new CustomClassLoader();
+			byte[] bytes = ClassGenerator.nestMemberWithNestHostAttribute();
+			Class<?> clazz = classloader.getClass("NestMembers", bytes);
+			Object instance = clazz.getDeclaredConstructor().newInstance();
+			bytes = ClassGenerator.nestMemberNoAttributeForceRedefinition();
+			successfulRedefinition = redefineClass(clazz, bytes.length, bytes);
```
DanHeidinga: Shouldn't this redefinition fail?  It removes the NestHost attribute.

test/cmdLineTests/jvmtitests/src/com/ibm/jvmti/tests/nestMatesRedefinition/ClassGenerator.java
```diff
@@ -0,0 +1,370 @@
...
+		methodVisitor.visitMethodInsn(INVOKESPECIAL, "java/lang/Object", "<init>", "()V", false);
+		methodVisitor.visitInsn(RETURN);
+		methodVisitor.visitMaxs(1, 1);
+		methodVisitor.visitEnd();
+
+		classWriter.visitEnd();
+		return classWriter.toByteArray();
+	}
+	
+	public static byte[] nestHostNoAttributeForceRedefinition () throws Exception {
```
DanHeidinga: It looks like the same basic class with a few modifications repeated over and over.  It could be written as:
```

public enum RedefinitionPhase { ORIGIINAL, REDEFINE };

public class ClassGenerator implements Opcodes {

	public static byte[] generateClass(String name, RedefinitionPhase phase, Attribute[] attributes) {
		ClassWriter classWriter = new ClassWriter(ClassWriter.COMPUTE_MAXS | ClassWriter.COMPUTE_FRAMES);

		classWriter.visit(V1_8, ACC_PUBLIC | ACC_SUPER, name, null, "java/lang/Object", null);
		switch(phase) {
		case ORIGINAL:
			/* Original has the source attribute */
			classWriter.visitSource(name + ".java", null);	
			break;
		case REDEFINE:
			/* REDEFINE doesn't have the source attribute to 
			 * ensure the VM can't skip the redefine 
			 */
			break;
		default:
			throw IllegalArgumentException("phase needs to be set correctly");
		}
		
		/* Generate base constructor which just calls super.<init>() */
		MethodVisitor methodVisitor = classWriter.visitMethod(ACC_PUBLIC, "<init>", "()V", null, null);
		methodVisitor.visitCode();
		methodVisitor.visitVarInsn(ALOAD, 0);
		methodVisitor.visitMethodInsn(INVOKESPECIAL, "java/lang/Object", "<init>", "()V", false);
		methodVisitor.visitInsn(RETURN);
		methodVisitor.visitMaxs(1, 1);
		methodVisitor.visitEnd();

		/* Generate passed in attributes if there are any */
		for (Attribute attr : attributes {
			classWriter.visitAttribute(attr);
		}

		classWriter.visitEnd();
		return classWriter.toByteArray();

	}

	/* What is the purpose of these two classes?  Why is the host with no nest being redefined */
	public static byte[] nestHostNoAttribute () throws Exception {
		return generateClass("NestHost", Phase.ORIGINAL, null);
	}
	
	public static byte[] nestHostNoAttributeForceRedefinition () throws Exception {
		return generateClass("NestHost", Phase.REDEFINE, null);
	}


	public static byte[] nestHostWithNestMembersAttribute () throws Exception {
		String[] nestMembers = {"A", "B", "C"};
		NestMembersAttribute attr = new NestMembersAttribute(nestMembers);

		return generateClass("NestHost", Phase.ORIGINAL, new Attribute[]{ attr });

	}
	
	public static byte[] nestHostWithNestMembersAttributeForceRedefinition () throws Exception {
		String[] nestMembers = {"A", "B", "C"};
		NestMembersAttribute attr = new NestMembersAttribute(nestMembers);

		return generateClass("NestHost", Phase.REDEFINE, new Attribute[]{ attr });
	}
...
```

Or something similar.  This makes it clear that its a small set of modifications to the base class and makes for less code to maintain.

taliamccormick: Will do; thanks.

test/cmdLineTests/jvmtitests/src/com/ibm/jvmti/tests/nestMatesRedefinition/NestHostAttribute.java
```diff
@@ -69,10 +69,7 @@ public boolean isUnknown(){
 		return true;
 	}
 	
-	public ByteVector write(ClassWriter cw,
-			byte[] code, 	int len,
-			int maxStack, 	int maxLocals){
-
+	public ByteVector write(ClassWriter cw, byte[] code, 	int len, int maxStack, 	int maxLocals){
```
DanHeidinga: formatting - can you remove the extra spaces between some of the arguments?

test/cmdLineTests/jvmtitests/src/com/ibm/jvmti/tests/nestMatesRedefinition/NestMembersAttribute.java
```diff
@@ -71,12 +71,8 @@ public boolean isUnknown(){
 		return true;
 	}
 
-	public ByteVector write(ClassWriter cw,
-			byte[] code, 	int len,
-			int maxStack, 	int maxLocals){
-
+	public ByteVector write(ClassWriter cw, byte[] code, 	int len, int maxStack, 	int maxLocals){
```
DanHeidinga: formatting - can you remove the extra spaces?

test/cmdLineTests/jvmtitests/src/com/ibm/jvmti/tests/nestMatesRedefinition/nmr001.java
```diff
@@ -102,7 +82,7 @@ public boolean testRedefinitionWithNestMembersAttribute() {
 			byte[] bytes = ClassGenerator.nestMemberWithNestHostAttribute();
 			Class<?> clazz = classloader.getClass("NestMembers", bytes);
 			Object instance = clazz.getDeclaredConstructor().newInstance();
-			bytes = ClassGenerator.nestMemberNoAttributeForceRedefinition();
+			bytes = ClassGenerator.nestMemberWithNestHostAttributeForceRedefinition();
```
DanHeidinga: Was the test failing?  If it wasn't, then we have a bug else where and keeping the equivalent of the original test is very important.

DanHeidinga:
@taliamccormick Can you squash this and then I'll run a build on it before merging?

Can you also provide some evidence that the tests pass?

taliamccormick:
Squashed & rebased.

For initial testing, I had manually recompiled a VM with jvmti changes, compiled the tests (both java & native), and ran them against eachother. They all initially passed.

Re-running them today caused one test case to fail - I'm recompiling both the tests and a build and stepping through the test case until it works. The prior set can be run as:
```
cd /team/tdmccormick/projects/20170509-NestMates/openj9/test/TestConfig
"/team/tdmccormick/projects/20170509-NestMates/JVMTI_TESTING/build_03/bin/java"  -Xcompressedrefs -agentlib:jvmtitest=test:nmr001 -cp "/team/tdmccormick/projects/20170509-NestMates/JVMTI_TESTING/openj9/test/TestConfig/scripts/testKitGen/../../../../jvmtest/cmdLineTests/jvmtitests/jvmtitest.jar:/team/tdmccormick/projects/20170509-NestMates/asm-6.0.jar" com.ibm.jvmti.tests.util.TestRunner
```
I'll confirm the test cases once I finish recompiling & re-running them.

DanHeidinga:
@taliamccormick Can you take a look at the 3 commits again?  Maybe something was off when squashing as the 3rd commit has "Requested PR changes; will squash prior to merge" as the comment.

taliamccormick:
@DanHeidinga I think it may have been a fluke. After recompiling the java test code and running it against the same build, the test cases passed

I'm still recompiling the native code to confirm the overall change set is fine but I don't expect any issues. The third commit "Requested PR changes; will squash prior to merge" has changes to the java code and to the playlist.xml but not to the native code which is the only remaining unconfirmed part

DanHeidinga:

DanHeidinga:
Running builds on a single platform / single JDK level as nestmates aren't enabled yet.  One should be sufficient to ensure we haven't broken anything

taliamccormick:
All nine tests pass on a VM compiled from last week with nestmates enabled!

Being run as:
>  "/team/tdmccormick/projects/20170509-NestMates/JVMTI_TESTING/build_05_redo3/bin//java"  -agentlib:jvmtitest=test:nmr001 -cp "/team/tdmccormick/projects/20170509-NestMates/JVMTI_TESTING/openj9_05/test/TestConfig/scripts/testKitGen/../../../../jvmtest/cmdLineTests/jvmtitests/jvmtitest.jar:/team/tdmccormick/projects/20170509-NestMates/asm-6.0.jar"   com.ibm.jvmti.tests.util.TestRunner

produces
```
*** Testing [1/9]:      testRedefinitionAddingNestHostAttribute
*** Test took 43 milliseconds
OK

*** Testing [2/9]:      testRedefinitionAddingNestMembersAttribute
*** Test took 2 milliseconds
OK

*** Testing [3/9]:      testRedefinitionAlteringNestHostAttribute
*** Test took 0 milliseconds
OK

*** Testing [4/9]:      testRedefinitionAlteringNestMembersContent
*** Test took 1 milliseconds
OK

*** Testing [5/9]:      testRedefinitionAlteringNestMembersLength
*** Test took 1 milliseconds
OK

*** Testing [6/9]:      testRedefinitionRemovingNestHostAttribute
*** Test took 1 milliseconds
OK

*** Testing [7/9]:      testRedefinitionRemovingNestMembersAttribute
*** Test took 1 milliseconds
OK

*** Testing [8/9]:      testRedefinitionWithNestHostAttribute
*** Test took 6 milliseconds
OK

*** Testing [9/9]:      testRedefinitionWithNestMembersAttribute
*** Test took 7 milliseconds
OK
```

Issue #1266, opened 2018-02-26
Changes to Unsafe.java have regressed JDK10 & JDK11 builds


andrew-m-leonard:
It seems the J9 jcl now has a new version of jdk/internal/misc/Unsafe.java which is unfortunately not compatible on JDK10, with the following build errors:
=== Output from failing command(s) repeated here ===
* For target jdk_modules_java.base__the.java.base_batch:
/home/admin/mydocker/ccm/openj9-openjdk/openj9-openjdk/src/java.base/share/classes/java/util/concurrent/ConcurrentHashMap.java:3310: error: method objectFieldOffset in class Unsafe cannot be applied to given types;
                = U.objectFieldOffset(TreeBin.class, "lockState");
                   ^
  required: Field
  found: Class<TreeBin>,String
  reason: actual and formal argument lists differ in length
/home/admin/mydocker/ccm/openj9-openjdk/openj9-openjdk/src/java.base/share/classes/java/util/concurrent/ConcurrentHashMap.java:6376: error: method objectFieldOffset in class Unsafe cannot be applied to given types;
        SIZECTL = U.objectFieldOffset
                   ^
  required: Field
  found: Class<ConcurrentHashMap>,String
  reason: actual and formal argument lists differ in length
   ... (rest of output omitted)

* All command lines available in /home/admin/mydocker/ccm/openj9-openjdk/openj9-openjdk/build/linux-x86_64-normal-server-release/make-support/failure-logs.
=== End of repeated output ===

See build log: https://hyc-rt-openjdk-jenkins.swg-devops.com/job/build_ibmsdk/266/console

andrew-m-leonard:
@DanHeidinga Same problem in JDK11 as well:
https://hyc-rt-openjdk-jenkins.swg-devops.com/job/build_ibmsdk/267/console

DanHeidinga:
@theresa-m Can you look at this?

DanHeidinga:
I suspect we're missing the `objectFieldOffset(Class, String)` implementation

theresa-m:
Yes. Looking into this.

PR #1284, opened 2018-02-27
Add new objectFieldOffset for Java 10 and up

theresa-m:
Fixes: #1266

### DIFF
--- a/jcl/src/java.base/share/classes/jdk/internal/misc/Unsafe.java
+++ b/jcl/src/java.base/share/classes/jdk/internal/misc/Unsafe.java
@@ -851,6 +851,19 @@ public final class Unsafe {
 	 * @throws IllegalArgumentException if field is static
 	 */
 	private native long objectFieldOffset0(Field field);
+	
+/*[IF Java10]*/
+	/*
+	 * Returns byte offset to field.
+	 * 
+	 * @param class with desired field
+	 * @param string name of desired field
+	 * @return offset to start of class or interface
+	 * 
+	 * @throws IllegalArgumentException if field is static
+	 */
+	private native long objectFieldOffset1(Class<?> c, String fieldName);
+/*[ENDIF]*/
 
 	/* 
 	 * Returns byte offset to start of static class or interface.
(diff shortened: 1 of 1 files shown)


jcl/src/java.base/share/classes/jdk/internal/misc/Unsafe.java
```diff
@@ -851,6 +851,19 @@ private native void copySwapMemory0(Object srcObj, long srcOffset, Object destOb
...
+	/*
+	 * Returns byte offset to field.
+	 * 
+	 * @param class with desired field
+	 * @param string name of desired field
+	 * @return offset to start of class or interface
+	 * 
+	 * @throws IllegalArgumentException if field is static
+	 */
+	private native long objectFieldOffset1(Class c, String fieldName);
```
DanHeidinga: `Class` -> `Class<?>`

jcl/src/java.base/share/classes/jdk/internal/misc/Unsafe.java
```diff
@@ -1354,6 +1367,24 @@ public long objectFieldOffset(Field field) {
...
+	 * Returns byte offset to field.
+	 * 
+	 * @param class with desired field
+	 * @param string name of desired field
+	 * @return offset to start of class or interface
+	 * 
+	 * @throws NullPointerException if field parameter is null
+	 * @throws IllegalArgumentException if field is static
+	 */
+	public long objectFieldOffset(Class c, String fieldName) {
```
DanHeidinga: `Class` -> `Class<?>`: Avoid raw types.

jcl/src/java.base/share/classes/jdk/internal/misc/Unsafe.java
DanHeidinga: Where do I find the implementation of `objectFieldOffset1`?  Is that in a separate PR or is this intended to be a stub?

theresa-m: It was written in by @JasonFengJ9 in sun_misc_Unsafe.cpp for initial java 10 support. Thanks Jason :)

DanHeidinga:
Jenkins test sanity plinux

andrew-m-leonard:
@DanHeidinga Is this ready to merge?

DanHeidinga:
@andrew-m-leonard Merged.  Thanks for the poke on this.

PR #1289, opened 2018-02-27
Fix a crash in findClassInfo on PPC

klangman:
Due to PPCs weak memory coherency it's possible for a crash
to occur in findClassInfo() if the caller has not obtained
the class table lock. This fix changes call to findClassInfo
into calls to findClassInfoAfterLocking in places where
the class table lock is not held.

### DIFF
--- a/runtime/compiler/trj9/env/PersistentCHTable.cpp
+++ b/runtime/compiler/trj9/env/PersistentCHTable.cpp
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2000, 2017 IBM Corp. and others
+ * Copyright (c) 2000, 2018 IBM Corp. and others
  *
  * This program and the accompanying materials are made available under
  * the terms of the Eclipse Public License 2.0 which accompanies this
@@ -290,7 +290,10 @@ TR_PersistentCHTable::findSingleJittedImplementer(
    return resolvedMethod;
    }
 
-
+/**
+ * Find persistent JIT class information for a given class.
+ * The caller must insure that class table lock is obtained
+ */
 TR_PersistentClassInfo *
 TR_PersistentCHTable::findClassInfo(TR_OpaqueClassBlock * classId)
    {
@@ -301,7 +304,10 @@ TR_PersistentCHTable::findClassInfo(TR_OpaqueClassBlock * classId)
    return cl;
    }
 
-
+/**
+ * Find persistent JIT class information for a given class.
+ * The class table lock is used to synchronize use of this method
+ */
 TR_PersistentClassInfo *
 TR_PersistentCHTable::findClassInfoAfterLocking(
       TR_OpaqueClassBlock *classId,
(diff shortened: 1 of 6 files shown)


runtime/compiler/trj9/env/j9method.cpp
```diff
@@ -4605,7 +4605,7 @@ TR_ResolvedJ9Method::setRecognizedMethodInfo(TR::RecognizedMethod rm)
       TR_OpaqueClassBlock *clazz = fej9()->getClassOfMethod(getPersistentIdentifier());
       J9JITConfig * jitConfig = fej9()->getJ9JITConfig();
       TR::CompilationInfo * compInfo = TR::CompilationInfo::get(jitConfig);
-      TR_PersistentClassInfo *clazzInfo = compInfo->getPersistentInfo()->getPersistentCHTable()->findClassInfo(clazz);
+      TR_PersistentClassInfo *clazzInfo = compInfo->getPersistentInfo()->getPersistentCHTable()->findClassInfoAfterLocking(clazz, fej9());
```
mpirvu: How does this code work when CHTable does not exist or someone uses disableCHOpts? I am aware that the new code is no worse than the old one.

runtime/compiler/trj9/env/PersistentCHTable.cpp
```diff
@@ -301,30 +304,46 @@ TR_PersistentCHTable::findClassInfo(TR_OpaqueClassBlock * classId)
...
+      TR_OpaqueClassBlock *classId,
+      TR_FrontEnd *fe,
+      bool returnClassInfoForAOT)
+   {
+   TR_J9VMBase *fej9 = (TR_J9VMBase *)(fe);
+   if (fej9->isAOT_DEPRECATED_DO_NOT_USE() && !returnClassInfoForAOT) // for AOT do not use the class hierarchy
       return NULL;
 
-   if (comp->getOption(TR_DisableCHOpts))
+   if (TR::comp()->getOption(TR_DisableCHOpts))
```
mpirvu: If this code is executed on a non-compilation thread (or when comp object does not yet exist) it will crash

mpirvu:
After looking in more depth at the places where changes are proposes I reached the following conclusions:
1) TR_AnnotationBase::scanForKnownAnnotationsAndRecord: 
This code is used in CompilationThread.cpp when there is no compilation object. We could move the call to a later point when we have the comp object. However, the code scans for allocations for a particular class, so even if we say `-Xjit:{method}(disableCHOpts)`, another method from the same class may induce the scanning of annotations (it's done once per class). I claim that it's the code that tries to take advantage of annotations that needs to check disableCHOpts per method
2) isObsoleteClass
I would say that this doesn't have to do with CHTable optimizations, but rather with correctness. If a class is obsolete we need to detect it even if I don't want to use CHOpts for a particular method. So in this case we don't need check against disableCHOpts, but we need a check that the CHTable exists
3) setRecognizedMethod
In the end the code here tries to get the obsoleteClass information so same reasoning as above applies.
4) TR_StringPeepholes::classRedefined(TR_OpaqueClassBlock *clazz)
Same reasoning as above


In conclusion, what I am proposing is:
1) Keep the `findClassInfoAfterLocking(..fe..)` function, but remove any checks to TR_DisableCHOpts
Use this function in all 4 places IsObsoleteClass() setRecognizedMethod()  TR_StringPeepholes::classRedefined() and scanForKnownAnnotationsAndRecord() (the last one already has a check for TR::Options::getCmdLineOptions()->getOption(TR_DisableCHOpts) and compInfo->getPersistentInfo()->getPersistentCHTable()
2) Remove `findClassInfoAfterLocking(..comp..)`
2) In setRecognizedMethod add a check that the CHTable exists
3) In  `findClassInfoAfterLocking(..comp..)` keep the checks against AOT and TR_DisableCHOpts

runtime/compiler/trj9/env/j9method.cpp
```diff
@@ -4605,7 +4605,11 @@ TR_ResolvedJ9Method::setRecognizedMethodInfo(TR::RecognizedMethod rm)
       TR_OpaqueClassBlock *clazz = fej9()->getClassOfMethod(getPersistentIdentifier());
       J9JITConfig * jitConfig = fej9()->getJ9JITConfig();
       TR::CompilationInfo * compInfo = TR::CompilationInfo::get(jitConfig);
-      TR_PersistentClassInfo *clazzInfo = compInfo->getPersistentInfo()->getPersistentCHTable()->findClassInfo(clazz);
+      TR_PersistentClassInfo *clazzInfo = NULL;
+      if (compInfo->getPersistentInfo() && compInfo->getPersistentInfo()->getPersistentCHTable())
```
mpirvu: No need for this extra check. persistentInfo always exists or will not start the JVM. It's just too essential.

mpirvu:
Jenkins test all

mpirvu:
The Travis CI build continues to fail on me, every time for a different reason. Last time it was:
`The job exceeded the maximum time limit for jobs, and has been terminated`
I'll try one more time tomorrow.

PR #1291, opened 2018-02-27
Remove J9 RemSet const

amicic:
Remove J9 specific Remembered Set constants, and reuse OMR ones, since
they have equvivalent values.


dmitripivkine:
Jenkins test sanity

charliegracie:
Does DDR use these J9 constants?

amicic:
DDR isn't part of openJ9 yet, but anyhow... I did not see problems with some local builds. Just looked at the code I see that, for example, iteration within a puddle is done using base/current pointers that are part of the puddle, hence not relying on the max puddle size constant.

However, there is comment before the constants, which I was not sure how relevant is:

Unused constants, kept here in case the JCL compiles use them

pshipton:
I didn't find any reference to these constants in DDR code.

PR #1292, opened 2018-02-27
Heed @FrameIteratorSkip in Class, AccessController

jdmpapin:
Particularly in Class.getStackClasses, AccessController.getAccSnapshot,
and AccessController.getCallerPD. Now when these methods encounter calls
to doPrivileged, the choice of the privileged frame will be insensitive
to whether or not calling MethodHandles were compiled.

When doPrivileged is called through a MethodHandle, this prevents stack
walks from terminating before reaching the injected SecurityFrame.

MethodHandle thunk frames are also now filtered out in getStackClasses,
so that they don't affect the results of SecurityManager.classDepth and
SecurityManager.classLoaderDepth.


pshipton:

pshipton:
Sanity test passed, I won't bother waiting for the ci testing.

PR #1295, opened 2018-02-27
Identify unsafe stubs in javadocs

theresa-m:

### DIFF
--- a/runtime/jcl/common/sun_misc_Unsafe.cpp
+++ b/runtime/jcl/common/sun_misc_Unsafe.cpp
@@ -159,6 +159,8 @@ Java_sun_misc_Unsafe_pageSize(JNIEnv *env, jobject receiver)
 jint JNICALL
 Java_sun_misc_Unsafe_getLoadAverage(JNIEnv *env, jobject receiver, jdoubleArray loadavg, jint nelems)
 {
+	/* stub implementation */
+	assert(!"Java_sun_misc_Unsafe_getLoadAverage is unimplemented");
 	return -1;
 }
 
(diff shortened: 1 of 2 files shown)


runtime/jcl/common/sun_misc_Unsafe.cpp
```diff
@@ -159,6 +159,8 @@ Java_sun_misc_Unsafe_pageSize(JNIEnv *env, jobject receiver)
 jint JNICALL
 Java_sun_misc_Unsafe_getLoadAverage(JNIEnv *env, jobject receiver, jdoubleArray loadavg, jint nelems)
 {
+	/* stub implementation */
+	assert(!"Java_sun_misc_Unsafe_getLoadAverage is unimplemented");
```
theresa-m: I'm assuming this assert was forgotten at some point?

pshipton: Do you mean did we forget to implement this? OpenJDK uses the unsafe API to implement OperatingSystemMXBean.getSystemLoadAverage(). OpenJ9 has a separate implementation of OperatingSystemMXBean, and implements this without calling unsafe, using the OpenJ9 native getSystemLoadAverageImpl instead. Nobody has ever complained about the unsafe API not working, but it should be possible to implement using the existing OpenJ9 code as a basis, for a single element array. The parameters aren't exactly the same between the two implementations.

pshipton:

runtime/jcl/common/sun_misc_Unsafe.cpp
pshipton: Guess I misunderstood the original comment, but there you go.

pshipton:
jenkins test all plinux jdk9

pshipton:

PR #1296, opened 2018-02-27
Unify JVM_SocketAvailable() return values

JasonFengJ9:
Unify `JVM_SocketAvailable()` return values

Remove most of https://github.com/eclipse/openj9/pull/714 such that `JVM_SocketAvailable()` return `0` for failure, and `1` for success consistently;
Keep the code receiving `shape=sun_openjdk` but map it to `J2SE_SHAPE_SUN` as well.

Reviewer @pshipton
FYI: @danheidinga


runtime/j9vm/jvm.c
```diff
@@ -1321,15 +1321,12 @@ typedef struct VersionSetting {
...
- *    the "shape" value specified within classlib.properties such that both IBM and OpenJ9 Java 8 can work properly.
- *    This is required because IBM Java 8 is applied with a JCL patch which expects different return values from JVM_SocketAvailable().
- *    Newly added shape value and related code can be removed after IBM JCL patch is removed.
+ * Note: shapes "sun" and "sun_openjdk" here are for accommodation of such values within Java 8 classlib.properties.
+ *       Both shapes are mapped to J2SE_SHAPE_SUN to indicate a Java 8 SDK.
  */
 VersionSetting SHAPE_SETTINGS[] = {
 		{"sun", J2SE_SHAPE_SUN},
-		{"sun_openjdk", J2SE_SHAPE_SUN_OPENJDK},
+		{"sun_openjdk", J2SE_SHAPE_SUN},
```
pshipton: The shape could be called J2SE_SHAPE_OPENJDK as this seems more appropriate today. 

We could modify /openj9-openjdk-jdk8/closed/OpenJ9.gmk to use `shape=sun_openjdk` which will obsolete the `shape=sun` value. Then the only reason to keep `sun` is to support swapping VMs between older and newer builds. After a few months it could be deleted.

JasonFengJ9: Changed to `J2SE_SHAPE_OPENJDK`, also created https://github.com/ibmruntimes/openj9-openjdk-jdk8/pull/54 against `OpenJ9 Java 8` extension repo.
However there are some issues with personal builds and can't verify the PR(s) yet.

pshipton:

PR #1298, opened 2018-02-27
Fix Clang error

babsingh:
Clang compilation issues an error when we initialize `sigaction` variables
using `{{0}}`. `sigaction` variables need to be initialized using `memset`
in order to fix the Clang compilation error.

`error : braces around scalar initializer [-Werror,-Wbraced-scalar-init]`

### DIFF
--- a/runtime/j9vm/jvm.c
+++ b/runtime/j9vm/jvm.c
@@ -4542,8 +4542,11 @@ JVM_RegisterSignal(jint sigNum, void* handler)
 	void *oldHandler = (void *)J9_SIG_ERR;
 
 #if !defined(WIN32)
-	struct sigaction newSignalAction = {{0}};
-	struct sigaction oldSignalAction = {{0}};
+	struct sigaction newSignalAction;
+	struct sigaction oldSignalAction;
+
+	memset(&newSignalAction, 0, sizeof(struct sigaction));
+	memset(&oldSignalAction, 0, sizeof(struct sigaction));
 #endif /* !defined(WIN32) */
 
 	Trc_SC_RegisterSignal();


DanHeidinga:
@babsingh Can you add the warning here?  I'm not clear on why `{{0}}` would result in a warning in this case.

babsingh:
I don't have the specific warning. But, the following OMR PR can be used as a reference: https://github.com/eclipse/omr/pull/2198. I have requested the author of the OMR changes for the warning in https://github.com/eclipse/omr/pull/2190.

babsingh:
from @amitin (posted in https://github.com/eclipse/omr/pull/2190),
If I restore for example this:
`struct sigaction act = {{0}};`
I can see the following as an error message issued by Clang:
`error : braces around scalar initializer [-Werror,-Wbraced-scalar-init]`

pshipton:
Does it work with `{0}` instead of `{{0}}`?

babsingh:
i don't have access to the Clang compiler which issues this error. the shape of sigaction struct may be different across operating systems. so, it is best to initialize it with `memset`.

DanHeidinga:
I believe Peter's suggestion is correct - to use `{ 0 }`.

DanHeidinga:
Thanks for digging up that link @babsingh.  Given the info there, I agree with the using `memset`.  Sorry for go around on this.

babsingh:
Reverted back to using `memset`.

DanHeidinga:

babsingh:
`wget https://sourceforge.net/projects/freemarker/files/freemarker/2.3.8/freemarker-2.3.8.tar.gz/download -O freemarker.tgz`

The above command failed abruptly in the Travis CI build.

DanHeidinga:
Merging despite travis.ci failure as it's due to missing freemarker.jar caused by sourceforge being down.

PR #1300, opened 2018-02-28
Allow tests targets selection

vsebe:
Allow users to choose what tests to run.
Reformat Jenkins files to check for test targets build parameter.
By default it runs sanity and extended tests.
Remove hard-coded values for the downstream jobs name.

### DIFF
--- a/buildenv/jenkins/common/pipeline-functions
+++ b/buildenv/jenkins/common/pipeline-functions
@@ -89,4 +89,33 @@ def build_with_one_upstream(JOB_NAME, UPSTREAM_JOB_NAME, UPSTREAM_JOB_NUMBER, VA
         return JOB
     }
 }
+
+def workflow() {
+    def jobs = [:]
+
+    // compile the source and build the SDK
+    BUILD_JOB_NAME = "Build-JDK${SDK_VERSION}-${SPEC}"
+    jobs["build"] = build(BUILD_JOB_NAME, OPENJDK_REPO, OPENJDK_BRANCH, SHAS['OPENJDK'], OPENJ9_REPO, OPENJ9_BRANCH, SHAS['OPENJ9'], OMR_REPO, OMR_BRANCH, SHAS['OMR'], params.VARIABLE_FILE)
+
+    levels = TESTS_TARGETS.split(",")
+    levels.each { level ->
+        switch (level.trim().toLowerCase()) {
+            case "_sanity":
+                TEST_JOB_NAME = "Test-Sanity-JDK${SDK_VERSION}-${SPEC}"
+                break
+            case "_extended":
+                TEST_JOB_NAME = "Test-Extended-JDK${SDK_VERSION}-${SPEC}"
+                break
+            default:
+                echo "error: Unknown test target: ${level}"
+                error()
+        }
+
+        // run tests against the SDK build in the upstream job: Build-JDK${SDK_VERSION}-${SPEC}
+        jobs["${level}"] = build_with_one_upstream(TEST_JOB_NAME, BUILD_JOB_NAME, BUILD_JOB.getNumber(), params.VARIABLE_FILE)
+    }
+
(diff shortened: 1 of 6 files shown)


vsebe:
@AdamBrousseau please review

vsebe:
https://ci.eclipse.org/openj9/view/Sandbox/job/vsebe_Pipeline-Build-Test-JDK9-linux_390-64_cmprssptrs/65/console

buildenv/jenkins/common/variables-functions
```diff
@@ -207,6 +216,17 @@ def set_test_variables() {
     TEST_DEPENDENCIES_JOB_NAME = VARIABLES.test_dependencies_job_name
 }
 
+/*
+* Set TESTS_TARGETS, indicating the level of testing. 
+*/
+def set_test_targets() {
+    TESTS_TARGETS = params.TESTS_TARGETS
+    if ((TESTS_TARGETS == null) || (TESTS_TARGETS == '')) {
+        // set default TESTS_TARGETS for pipeline job (run all tests)
```
smlambert: presume by looking at default value, that parameter accepts a comma-separated list of test targets, any restrictions?

vsebe: Yes, it is restricted to _sanity and _extended for now.
We'll add a generic pipeline to be able to run any tests.

vsebe:
@AdamBrousseau ready for review

buildenv/jenkins/common/pipeline-functions
```diff
@@ -89,4 +89,33 @@ def build_with_one_upstream(JOB_NAME, UPSTREAM_JOB_NAME, UPSTREAM_JOB_NUMBER, VA
...
+    levels.each { level ->
+        switch (level.trim().toLowerCase()) {
+            case "_sanity":
+                TEST_JOB_NAME = "Test-Sanity-JDK${SDK_VERSION}-${SPEC}"
+                break
+            case "_extended":
+                TEST_JOB_NAME = "Test-Extended-JDK${SDK_VERSION}-${SPEC}"
+                break
+            default:
+                echo "error: Unknown test target: ${level}"
```
smlambert: Currently, it will not work to pass other targets to internal builds because of this default: handling.

vsebe: Correct (throws an error).
In order to work with other targets we need Jenkins job[s] and pipeline/groovy scripts that are not available at this moment. As agreed today, we will discuss and address this at a later time.

buildenv/jenkins/common/variables-functions
```diff
@@ -207,6 +216,17 @@ def set_test_variables() {
     TEST_DEPENDENCIES_JOB_NAME = VARIABLES.test_dependencies_job_name
 }
 
+/*
+* Set TESTS_TARGETS, indicating the level of testing. 
+*/
+def set_test_targets() {
+    TESTS_TARGETS = params.TESTS_TARGETS
+    if (!TESTS_TARGETS) {
```
AdamBrousseau: `TEST_TARGETS` will either be `null` (no param defined), blank (param defined but no value), or will have a value. I would like to see a test case for each to prove this `if` condition works as expected.
It is different than how we do it in [variables-functions](https://github.com/eclipse/openj9/blob/master/buildenv/jenkins/common/variables-functions).

Especially since I got burned by my change in #1320

vsebe: In groovy, both null and the empty string evaluate to false
if ((TESTS_TARGETS == null) || ((TESTS_TARGETS == '')) {...} thus the contraption (!TESTS_TARGETS){...}

AdamBrousseau: Right. I made the mistake of wrapping quotes around `null` in my change. I think this is correct.

vsebe:
@pshipton ready for review

PR #1307, opened 2018-02-28
Split zOS system linkage call dependencies

NigelYiboYu:
Split the zOS system linkage call dependencies by attaching the
pre-dependencies to the BASR call instruction, and attaching the
post-dependencies to the function call descriptor instruction.

The C XPLINK return will need a NOP after the BASR for entry to valid
instructions. If they are separated by spills, returning from C code can
cause exceptions. Having a post-dependency group after the NOP ensures
that no such spill can happen.

This pull request also contains  miscellaneous zOS system linkage code simplifications

    1. Simplify NOP and call descriptor generation for zOS system linkage.

    2. Use the insetPad() API to insert NOP after the BASR call in system
    linkage.

    3. Remove the oneliner createCallDescriptor() function.  Do the same thing inline.

    4. genCallNOPAndDescriptor() does not have to be virtual. Remove the virtual
    keyword.

### DIFF
--- a/runtime/compiler/trj9/z/codegen/J9S390SystemLinkage.cpp
+++ b/runtime/compiler/trj9/z/codegen/J9S390SystemLinkage.cpp
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2000, 2016 IBM Corp. and others
+ * Copyright (c) 2000, 2018 IBM Corp. and others
  *
  * This program and the accompanying materials are made available under
  * the terms of the Eclipse Public License 2.0 which accompanies this
@@ -395,7 +395,25 @@ TR::J9S390zOSSystemLinkage::generateInstructionsForCall(TR::Node * callNode, TR:
          }
       }
 
-   gcPoint = generateRRInstruction(codeGen, TR::InstOpCode::BASR, callNode, systemReturnAddressRegister, systemEntryPointRegister, deps);
+   /**
+    * NOP padding is needed because returning from XPLINK functions skips the XPLink eyecatcher and
+    * always return to a point that's 2 or 4 bytes after the return address.
+    *
+    * In 64 bit XPLINK, the caller returns with a 'branch relative on condition' instruction with a 2 byte offset:
+    *
+    *   0x47F07002                    B        2(,r7)
+    *
+    * In 31-bit XPLINK, this offset is 4-byte.
+    *
+    * As a result of this, JIT'ed code that does XPLINK calls needs 2 or 4-byte NOP paddings to ensure entry to valid instruction.
+    *
+    * The BASR and NOP padding must stick together and can't have reverse spills in the middle.
+    * Hence, splitting the dependencies to avoid spill instructions.
+    */
+   TR::RegisterDependencyConditions* callPreDeps = new (self()->trHeapMemory()) TR::RegisterDependencyConditions(deps->getPreConditions(), NULL, deps->getAddCursorForPre(), 0, codeGen);
+   TR::RegisterDependencyConditions* callPostDeps = new (self()->trHeapMemory()) TR::RegisterDependencyConditions(NULL, deps->getPostConditions(), 0, deps->getAddCursorForPost(), codeGen);
+
+   gcPoint = generateRRInstruction(codeGen, TR::InstOpCode::BASR, callNode, systemReturnAddressRegister, systemEntryPointRegister, callPreDeps);
    if (isJNIGCPoint)
          gcPoint->setNeedsGCMap(0x00000000);
 
@@ -403,15 +421,18 @@ TR::J9S390zOSSystemLinkage::generateInstructionsForCall(TR::Node * callNode, TR:
 
    generateS390LabelInstruction(codeGen, TR::InstOpCode::LABEL, callNode, returnFromJNICallLabel);
 
+   TR::Instruction * cursor = NULL;
    if (TR::Compiler->target.is64Bit())
       {
       //In 64 bit XPLINK, the caller returns at RetAddr+2, so add 2 byte nop
-      TR::Instruction * cursor = new (trHeapMemory()) TR::S390NOPInstruction(TR::InstOpCode::NOP, 2, callNode, codeGen);
+      cursor = new (trHeapMemory()) TR::S390NOPInstruction(TR::InstOpCode::NOP, 2, callNode, codeGen);
       }
    else
       {
-      genCallNOPAndDescriptor(NULL, callNode, callNode, callType);
+      cursor = genCallNOPAndDescriptor(NULL, callNode, callNode, callType);
       }
+
+   cursor->setDependencyConditions(callPostDeps);
 
    if (cg()->supportsJITFreeSystemStackPointer())
       {
--- a/runtime/compiler/trj9/z/codegen/J9S390SystemLinkage.cpp
+++ b/runtime/compiler/trj9/z/codegen/J9S390SystemLinkage.cpp
@@ -296,7 +296,6 @@ TR::J9S390zOSSystemLinkage::generateInstructionsForCall(TR::Node * callNode, TR:
    TR::Register * systemEnvironmentRegister = javaStackRegister;
    TR::Register * javaLitPoolRegister = privateLinkage->getLitPoolRealRegister();
    bool passLitPoolReg = false;
-   TR_XPLinkCallTypes callType;  // for XPLink calls
 
    if (codeGen->isLiteralPoolOnDemandOn())
       {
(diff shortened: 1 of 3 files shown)


NigelYiboYu:
@fjeremic  Tested and ready for review. 
Travis CI seems to be failing with gzip problems.

runtime/compiler/trj9/z/codegen/J9S390SystemLinkage.cpp
```diff
@@ -395,24 +395,35 @@ TR::J9S390zOSSystemLinkage::generateInstructionsForCall(TR::Node * callNode, TR:
          }
       }
 
-   gcPoint = generateRRInstruction(codeGen, TR::InstOpCode::BASR, callNode, systemReturnAddressRegister, systemEntryPointRegister, deps);
+   // NOP padding is needed because return from C function will skip the XPLink eyecatcher and
+   // always return to a point that's 4 bytes after the return address.
```
fjeremic: This is incorrect. It is 2 bytes after the call instruction `BASR` (which is itself 2 bytes). The NOP we generate is 2 bytes in length. Can you fix this comment?

runtime/compiler/trj9/z/codegen/J9S390SystemLinkage.cpp
```diff
@@ -395,24 +395,35 @@ TR::J9S390zOSSystemLinkage::generateInstructionsForCall(TR::Node * callNode, TR:
...
    callType = TR_XPLinkCallType_BASR;
 
    generateS390LabelInstruction(codeGen, TR::InstOpCode::LABEL, callNode, returnFromJNICallLabel);
 
+   TR::Instruction * cursor = NULL;
    if (TR::Compiler->target.is64Bit())
       {
       //In 64 bit XPLINK, the caller returns at RetAddr+2, so add 2 byte nop
-      TR::Instruction * cursor = new (trHeapMemory()) TR::S390NOPInstruction(TR::InstOpCode::NOP, 2, callNode, codeGen);
+      cursor = new (trHeapMemory()) TR::S390NOPInstruction(TR::InstOpCode::NOP, 2, callNode, codeGen);
```
fjeremic: Can you use the `OMR::Z::CodeGenerator::insertPad` API here instead?

runtime/compiler/trj9/z/codegen/J9S390SystemLinkage.cpp
NigelYiboYu: Currently, the JIT generates this for the JNI: 
```
 0d 76                       BASR    GPR7,GPR6  
 47 00 00 51              NOP
```

The BASR is 2 bytes. And the XPLINK return address in GPR7 should be pointing at the NOP, which is really a 4-byte `branch on condition` instruction that does not jump.

As an example, this is what xlc produces when I compile C++ with XPLINK option:
```
   207 15650ZOS V2.3 z/OS XL C++                                     SimpleCall.cpp: Java_com_jni_c...   02/20/18 12:28:59            5
    208
    209  OFFSET OBJECT CODE        LINE#  FILE#    P S E U D O   A S S E M B L Y   L I S T I N G
    210
    211  000124  987F  480C        000029 |                   LM       r7,r15,2060(r4)
    212  000128  4140  4080        000029 |                   LA       r4,128(,r4)
    213  00012C  47F0  7004        000029 |                   B        4(,r7)
```

The `B 4(,7)` instruction is `0x47f07004`. It's `branch on condition` instruction with an unconditional jump mask 0xF. The offset is 4, in bytes.

runtime/compiler/trj9/z/codegen/J9S390SystemLinkage.cpp
NigelYiboYu: OK. I'll try insertPad API.

runtime/compiler/trj9/z/codegen/J9S390SystemLinkage.cpp
fjeremic: But a few lines below we generate a 2 byte NOP:

```
//In 64 bit XPLINK, the caller returns at RetAddr+2, so add 2 byte nop
cursor = new (trHeapMemory()) TR::S390NOPInstruction(TR::InstOpCode::NOP, 2, callNode, codeGen);
```

https://github.com/eclipse/openj9/pull/1307/files#diff-bd656156fc34aa9cbcf33d25dba26a0eR417

So one of these comments has to be incorrect. Looking at the documentation for XPLINK [1] it is indeed 4 bytes in length. We have `0x470XYYYY` where `X` represents the call type which should be 0 as we're generating a `BASR` and `YYYY` should technically point to the entry point but this is not an XPLINK frame but a private JIT linkage frame so it too should be 0.

NigelYiboYu: 64bit XPLINK uses 2-byte NOP; and 32-bit XPLINK uses 4-byte NOP.
See `TR::J9S390zOSSystemLinkage::genCallNOPAndDescriptor()`

fjeremic: > 64bit XPLINK uses 2-byte NOP; and 31-bit XPLINK uses 4-byte NOP.

So if it XPLINK 64-bit uses a 2 byte NOP and we branch "to a point that's 4 bytes after the return address" wouldn't we branch in the middle of an instruction?

NigelYiboYu: Sorry, I wasn't clear about this.

The `BR 4(,7)` example I showed you was 31-bit C++ epilogue in XPLINK.
64-bit C++ XPLINK return has a 2-byte offset:

```
    193  0000F4                    Start of Epilog
    194  0000F4  E370  4818  0004  000029 |                   LG       r7,2072(,r4)
    195  0000FA  4140  4100        000029 |                   LA       r4,256(,r4)
    196  0000FE  47F0  7002        000029 |                   B        2(,r7)

```


So in our JIT'ed system call, the padding should be 2-byte for 64bit and 4-byte for 31-bit.

NigelYiboYu:
Re-testing my changes. Marking this as WIP.

runtime/compiler/trj9/z/codegen/J9S390SystemLinkage.cpp
NigelYiboYu: I refactored and simplified the BASR padding function in system linkage class to use `insertPad()` API. I'm currently running tests.

runtime/compiler/trj9/z/codegen/J9S390SystemLinkage.cpp
fjeremic: Ok, we need to properly document this as it is non-obvious. Could you amend the comment with some of the details regarding 31-bit vs. 64-bit XPLINK and what is expected as well as what the NOP is really supposed to represent (call type, call descriptor, etc.)?

r30shah: That is true.  @NigelYiboYu  Padding is 2 Bytes for 64-Bit zOS and 4 Bytes for 31-bit zOS.
https://github.com/eclipse/openj9/blob/26f8c7edb1fe37253d001243c0ce14473334d07f/runtime/compiler/trj9/z/codegen/J9S390CHelperLinkage.cpp#L348

NigelYiboYu: Updated comments in `SystemLinkage` and `CHelperLinkage` classes.

NigelYiboYu:
@fjeremic New changes were tested. Results look ok.

runtime/compiler/trj9/z/codegen/J9S390SystemLinkage.cpp
```diff
@@ -505,64 +516,38 @@ TR::J9S390zOSSystemLinkage::addFECustomizedReturnRegDependency(int64_t killMask,
...
-      {
-      //In 64 bit XPLINK, the caller returns at RetAddr+2, so add 2 byte nop
-      cursor = nop = new (trHeapMemory()) TR::S390NOPInstruction(TR::InstOpCode::NOP, 2, node, cursor, codeGen);
-      }
-   else
+   //In 64 bit XPLINK, the caller returns at RetAddr+2, so add 2 byte nop
+   //In 31 bit XPLINK, the caller returns at RetAddr+4, so add 4 byte nop with last two bytes
+   // as signed offset in doublewords at or preceding NOP
+   uint32_t padSize = TR::Compiler->target.is64Bit() ? 2 : 4;
+   cursor = nop = new (trHeapMemory()) TR::S390NOPInstruction(TR::InstOpCode::NOP, padSize, node, codeGen);
```
fjeremic: Can we use the `insertPad` API here? Also spaces between `//` and the comment.

NigelYiboYu:
Re-tested my changes after using `insertPad()`. Test passed.  @fjeremic

fjeremic:
Jenkins test all

fjeremic:
Failures seem infrastructure issues cloning the repos. The Linux s390 passed though which is all that should matter for this change.

PR #1308, opened 2018-02-28
Defer enabling dump agents in RAS

ChengJin01:
The change is to defer enabling agents in pushDumpFacade so as
to ensure an unexpected crash in initRasDumpGlobalStorage()
only leads to an normal OS dump before RAS is ready for dump
generation.

### DIFF
--- a/runtime/rasdump/dmpsup.c
+++ b/runtime/rasdump/dmpsup.c
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 1991, 2017 IBM Corp. and others
+ * Copyright (c) 1991, 2018 IBM Corp. and others
  *
  * This program and the accompanying materials are made available under
  * the terms of the Eclipse Public License 2.0 which accompanies this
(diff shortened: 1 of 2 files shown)


ChengJin01:
Senior reviewer: @keithc-ca 
Peer reviewer: @babsingh

babsingh:
Changes lgtm.

Minor note: `initRasDumpGlobalStorage` doesn't return an error code. `initRasDumpGlobalStorage` can fail. I think we need to verify if initialization code within `initRasDumpGlobalStorage` was successful before proceeding with the remaining initialization in `J9VMDllMain`. These changes can be ignored since `initRasDumpGlobalStorage` was not changed.

keithc-ca:
Ignoring the failure in https://ci.eclipse.org/openj9/job/PullRequest-Sanity-JDK9-linux_ppc-64_cmprssptrs_le-OpenJ9/94/ which is a known issue (https://github.com/eclipse/openj9/issues/1114).

PR #1319, opened 2018-03-01
Remove check for J2SE_18 assuming Java 8 is the lowest level supported

JasonFengJ9:
Remove check for `J2SE_18` assuming `Java 8` is the lowest level supported

This assumes that `OpenJ9` builds `Java 8` and upward Java levels.
Note: some `J2SE_1[5|6|7]` references are still around to provide back compatible support.

Reviewer @pshipton
FYI: @danheidinga

### DIFF
--- a/runtime/vm/BytecodeInterpreter.hpp
+++ b/runtime/vm/BytecodeInterpreter.hpp
@@ -2929,11 +2929,7 @@ done:
 			walkState->userData1 = (void*)2; /* skip newInstanceImpl and newInstance */
 			walkState->userData2 = NULL;
 			walkState->userData3 = (void *) FALSE;
-			walkState->frameWalkFunction = cInterpGetStackClassIterator;
-
-			if ((J2SE_VERSION(_vm) & J2SE_VERSION_MASK) >= J2SE_18) {
-				walkState->frameWalkFunction = cInterpGetStackClassJEP176Iterator;
-			}
+			walkState->frameWalkFunction = cInterpGetStackClassJEP176Iterator;
 
 			walkState->skipCount = 0;
 			walkState->walkThread = _currentThread;
@@ -3920,11 +3916,7 @@ done:
 		walkState->userData1 = (void*)(UDATA)(depth + 1); /* Skip the local INL frame */
 		walkState->userData2 = NULL;
 		walkState->userData3 = (void *) FALSE;
-		walkState->frameWalkFunction = cInterpGetStackClassIterator;
-
-		if ((J2SE_VERSION(_vm) & J2SE_VERSION_MASK) >= J2SE_18) {
-			walkState->frameWalkFunction = cInterpGetStackClassJEP176Iterator;
-		}
+		walkState->frameWalkFunction = cInterpGetStackClassJEP176Iterator;
 
 		walkState->skipCount = 0;
 		walkState->walkThread = _currentThread;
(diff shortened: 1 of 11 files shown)


runtime/vm/jvminit.c
```diff
@@ -2547,9 +2547,7 @@ modifyDllLoadTable(J9JavaVM * vm, J9Pool* loadTable, J9VMInitArgs* j9vm_args)
 			jvmti = TRUE;
 		} else
 #if defined(J9VM_OPT_SIDECAR)
-		if ((J2SE_VERSION(vm) & J2SE_VERSION_MASK) >= J2SE_16) {
-			jvmti = TRUE;
-		}
+		jvmti = TRUE;
```
JasonFengJ9: Note: this code can be further simplified if `J9VM_OPT_SIDECAR`is always enabled which is the case for all current specs but leave that to another PR instead.

runtime/vm/BytecodeInterpreter.hpp
```diff
@@ -4281,11 +4273,7 @@ done:;
 		walkState->userData1 = (void*)(UDATA)(depth + 1); /* Skip the local INL frame */
 		walkState->userData2 = NULL;
 		walkState->userData3 = (void *) FALSE;
-		walkState->frameWalkFunction = cInterpGetStackClassIterator;
-
-		if ((J2SE_VERSION(_vm) & J2SE_VERSION_MASK) >= J2SE_18) {
-			walkState->frameWalkFunction = cInterpGetStackClassJEP176Iterator;
-		}
+		walkState->frameWalkFunction = cInterpGetStackClassJEP176Iterator;
```
pshipton: Seems the function cInterpGetStackClassIterator can be deleted now.

runtime/vm/jvminit.c
pshipton: Is there some reason not to just delete all this code and replace it with `jvmti = TRUE;`? Seems an easy change.

JasonFengJ9: As per previous note, I preferred to leave further simplification to another PR for the flag `J9VM_OPT_SIDECAR`.
However just noticed that original code already relies on `J9VM_OPT_SIDECAR` is on, other syntax error occurs. 

```
#if defined(J9VM_OPT_JVMTI)
		if ((strstr(testString, VMOPT_AGENTLIB_COLON)==testString) || (strstr(testString, VMOPT_AGENTPATH_COLON)==testString)) {
			ADD_FLAG_AT_INDEX( ARG_REQUIRES_LIBRARY, i );
			jvmti = TRUE;
		} else   ---> following code won't exist if J9VM_OPT_SIDECAR is not enabled.
#if defined(J9VM_OPT_SIDECAR)
		if ((J2SE_VERSION(vm) & J2SE_VERSION_MASK) >= J2SE_16) {
			jvmti = TRUE;
		}
#endif
#endif
```

Will remove `jvmti` completely instead.

runtime/vm/BytecodeInterpreter.hpp
JasonFengJ9: Deleted. Pls have another look.

pshipton:
Copyright check failed

JasonFengJ9:
Copyright has been fixed. It passes now.

pshipton:
jenkins test sanity plinux

PR #1320, opened 2018-03-01
Do not archive test material in pr builds

### LINKED ISSUE #1114: Archive test step fails intermittently when run in parallel with testing

AdamBrousseau:
Because we were archiving in parallel to testing,
archiving the test material would often fail in PR
builds becasue we are trying to write to the test
folder at the same time we are trying to tar it.
This PR will change the behaviour so we do not
bother archiving the test material in PR builds. Since we
compile & test on the same machine in PRs. In non-pr
builds we do compile on one machine and then test on
another.

Fixes #1114

### DIFF
--- a/buildenv/jenkins/common/build
+++ b/buildenv/jenkins/common/build
@@ -163,9 +163,15 @@ def build() {
 def archive() {
     stage('Archive') {
         timestamps {
+            // Do not archive test material in PR jobs, only "Build" jobs
+            // https://github.com/eclipse/openj9/issues/1114
+            // ghprbPullId is the PullRequest ID which only shows up in Pull Requests
+            if ((params.ghprbPullId == "") || (params.ghprbPullId == "null")) {
+                sh "tar -zcvf ${WORKSPACE}/${TEST_PREFIX}`git -C openj9 rev-parse --short HEAD`${TEST_SUFFIX} openj9/test/"
+                archiveArtifacts artifacts: "**/${TEST_PREFIX}*${TEST_SUFFIX}", fingerprint: true, onlyIfSuccessful: true
+            }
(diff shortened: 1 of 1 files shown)


AdamBrousseau:

AdamBrousseau:

AdamBrousseau:
I don't think the Jenkins build is picking up my change because the pipeline files are always loaded from the eclipse repo and not from the PR/merge.

AdamBrousseau:
jenkins compile plinux jdk9

AdamBrousseau:
Ok, compile PR is running loaded out of my branch so it should be an actual test of the change

AdamBrousseau:
Ok, I'm satisfied, my compile build did not archive the test material

smlambert:
Thank you for this PR @AdamBrousseau

PR #1326, opened 2018-03-02
Add required dependencies on generate_j9vm

keithc-ca:
If make attempts to build some modules (e.g.  `jclse9_`) before `generate_j9vm` compiles will fail due to a missing `generated.h`. Adding these dependencies prevents those problems.

### DIFF
--- a/runtime/jcl/cl_se9_before_b165/module.xml
+++ b/runtime/jcl/cl_se9_before_b165/module.xml
@@ -20,7 +20,6 @@
 
 	SPDX-License-Identifier: EPL-2.0 OR Apache-2.0 OR GPL-2.0 WITH Classpath-exception-2.0 OR LicenseRef-GPL-2.0 WITH Assembly-exception
 -->
-
 <module xmlns:xi="http://www.w3.org/2001/XInclude">
 
 	<xi:include href="../uma/sun_misc_Unsafe_exports.xml"></xi:include>
@@ -68,6 +67,9 @@
 			<option name="dllDescription" data="Java SE 9"/>
 		</options>
 		<phase>core j2se</phase>
+		<dependencies>
+			<dependency name="generate_j9vm"/>
+		</dependencies>
 		
 		<exports>
 			<group name="se6_vm-side_lifecycle"/>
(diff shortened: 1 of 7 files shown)


keithc-ca:
The problem was introduced by #1009. Because UMA doesn't create generated.h up front, we need dependencies to the new target to ensure that file is created early enough.

runtime/jcl/cl_se9_before_b165/module.xml
```diff
@@ -1,172 +1,174 @@
...
-			<library name="j9pool" type="external"/>
-			<library name="j9thr"/>
-			<library name="j9vmi"/>
-			<library name="socket" type="macro"/>
-			<library name="sunvmi"/>
-		</libraries>
-	</artifact>
-</module>
\ No newline at end of file
+<?xml version="1.0" encoding="UTF-8"?>
```
DanHeidinga: Why does this file show up as 100% changed?

runtime/tests/jni/module.xml
```diff
@@ -1,389 +1,390 @@
...
-		</makefilestubs>
-		<libraries>
-			<library name="j9thr"/>
-			<library name="j9util"/>
-			<library name="j9utilcore"/>
-			<library name="jvm"/>
-		</libraries>
-	</artifact>
-</module>
+<?xml version="1.0" encoding="UTF-8"?>
```
DanHeidinga: Why does this show up as 100% changed?

keithc-ca: Those two files had the wrong line-ends and were caught by the automated check. The real changes are in the first commit.

keithc-ca: Visit https://github.com/eclipse/openj9/pull/1326/files?w=1 to see the changes, ignoring whitespaces.

pshipton:
I'm wondering if some of these dependencies on generated.h should be removed. I don't think the jcl or the jni tests actually need it. I'm not sure why these modules are including jvm.h

keithc-ca:
I hadn't thought of removing those dependencies instead of declaring them. I'll do some more testing to be sure that's a workable plan.

pshipton:
To be more specific, they might need jvm.h for some reason (that I don't know), but they don't appear to need generated.h which is included by jvm.h

keithc-ca:
Closing this in favour of https://github.com/eclipse/openj9/pull/1335.

Issue #1180, opened 2018-02-15
xLinux builds of openjdk8-openj9 are failing due to openj9 build failure


ben-walsh:
See https://github.com/ibmruntimes/openj9-openjdk-jdk8/issues/44.
Specifically https://github.com/ibmruntimes/openj9-openjdk-jdk8/issues/44#issuecomment-365925203.

NB : Builds have been failing like this since 6th February.

keithc-ca:
This is a long-standing issue with JIT when compiled with a pedantic compiler. It normally doesn't use the 'warnings-are-errors' option.

sxa555:
The build machine that is being used has gcc5 on it as well as 4.8. It would appear that there is an issue in the build process whereby parts of the build process will pick up the `CC`/`CXX` environment variables to override the defaults but some parts won't. This sounds like something that should be fixed as the `Tools summary` section of the log suggests it's using the `gcc-4.8` when the variable is set, but later on in the OMR configure step it's still picking up the default `gcc`

pshipton:
Based on the last comment, I'm going to re-tag this as a bug build.

OpenJ9 supports gcc 4.8 atm. 4.8.5 is the version used for ci testing and in the OpenJ9 nightly builds. Other 4.x versions may work but since they are not regularly tested, problems could be introduced.

sxa555:
Reference to the CI: https://ci.adoptopenjdk.net/job/openjdk10_build_x86-64_linux/

andrew-m-leonard:
@pshipton Peter this will block releasing JDK8, since we cannot build it at AdoptOpenJDK... Adopt are trying to pass the necessary autoconf "configure" env options through to openj9/omr, but they seem to be being ignored... can we look at that issue to workaround the lack of 5 support?
We probably ought to tag this as JDK8 as well?

pshipton:
It doesn't seem the problem is specific to jdk8 even though it affects jdk8. We can tag it as high priority.

DanHeidinga:
@sxa555 Can I get access to the Adopt machine to do some manual testing?  Is there a procedure at Adopt I can follow to get a temporary login / password?

smlambert:
@DanHeidinga - you raise an issue at the openjdk-infrastructure repo to request temporary access (example: https://github.com/AdoptOpenJDK/openjdk-infrastructure/issues/203)

DanHeidinga:
[...]

PR #1330, opened 2018-03-02
Allow for overriding build tools via the command line or environment

ymanton:
These patches allow for building OpenJ9 with alternative build tools by overriding `CC`, `CXX`, and the other standard GNU make implicit variables, as well as OpenJ9-specific variables.

Note: OMR still has issues and will require build tools known to autoconf to be in your path, and overriding `CC` and `CXX` with an absolute path will break OpenJDK's `configure`, so use just the tool name, e.g. `CC=gcc-4.8` and make sure it's in your path.

Fixes: #1180


DanHeidinga:
@ymanton Any suggestions for reviewers for the JIT makefiles?  Anyone particularly knowledgable in that area who can look would be appreciated.

ymanton:
@JamesKingdon, you've poked around the JIT makefiles recently for ARM. Can you take a look?

@fjeremic may have also poked around those parts when he re-organized the compiler sub-tree.

runtime/makelib/mkconstants.mk.ftl
```diff
@@ -33,27 +33,70 @@ ifndef UMA_TARGET_PATH
...
-<#if uma.spec.tools.ar.needed>AR=${uma.spec.tools.ar.name}<#else># AR not used</#if>
-<#if uma.spec.tools.ranlib.needed>RANLIB=${uma.spec.tools.ranlib.name}<#else># RANLIB not used</#if>
-<#if uma.spec.tools.dll_ld.needed>DLL_LD=${uma.spec.tools.dll_ld.name}<#else># DLL_LD not used</#if>
-<#if uma.spec.tools.cxx_dll_ld.needed>CXX_DLL_LD=${uma.spec.tools.cxx_dll_ld.name}<#else># CXX_DLL_LD not used</#if>
-<#if uma.spec.tools.exe_ld.needed>EXE_LD=${uma.spec.tools.exe_ld.name}<#else># EXE_LD not used</#if>
-<#if uma.spec.tools.cxx_exe_ld.needed>CXX_EXE_LD=${uma.spec.tools.cxx_exe_ld.name}<#else># CXX_EXE_LD not used</#if>
-<#if uma.spec.tools.dll_ld.needed>UMA_DLL_LD=$(if $(UMA_IS_C_PLUS_PLUS),$(CXX_DLL_LD),$(DLL_LD))</#if>
-<#if uma.spec.tools.exe_ld.needed>UMA_EXE_LD=$(if $(UMA_IS_C_PLUS_PLUS),$(CXX_EXE_LD),$(EXE_LD))</#if>
+<#if uma.spec.tools.interp_gcc.needed>
+# Order of preference: INTERP_GCC if set, CC if set and not default, otherwise uma.spec.tools.interp_gcc.name
```
DanHeidinga: Calling this out for whoever reviews this - the order here needs some careful thought as I'm not sure if the CC before uma.spec.tools.interp_gcc.name is correct

JamesKingdon:
Looks good to me, although I surely can't tell from looking at it if it's going to work or not. It's clearly a major improvement, so let's put it in and fix any problems as they arise. Thanks for tackling this painful task.

runtime/makelib/mkconstants.mk.ftl
DanHeidinga: Extensions repos may be hacking around this differently - see https://github.com/ibmruntimes/openj9-openjdk-jdk8/blob/84153c75da79342522533371b19330181ab4ff4f/closed/OpenJ9.gmk#L227

ymanton: My thinking here is that if a typical user sets `CC` the expectation will be that that everything is built with the requested compiler. If we don't allow `CC` to set `INTERP_GCC` they'll have to dig around to discover what's going on and override it themselves, which is an annoyance I wouldn't want to inflict.

On the flip side anyone who's aware of `INTERP_GCC` can still set it themselves if they override `CC`, although it just occurred to me that they'd have to know what the hardcoded value was, so perhaps another var like `INTERP_GCC_IGNORE_CC` to control this behaviour makes sense, that way you can set your own value or keep the hardcoded value without knowing/caring what it is.

DanHeidinga:
I've tested this PR on an Ubuntu 16 machine with `CC` and `CXX` set and was able to successfully build an OpenJ9 JDK.

```
export CC=gcc-4.8
export CXX=g++-4.8
```

DanHeidinga:
Jenkins test sanity

ymanton:
I decided to overrule myself on the `INTERP_GCC` bit because there's no guarantee that `CC` and `INTERP_GCC` are interchangeable; for example if `CC` is XLC or MSVC and the user sets it to an alternative XLC or MSVC it wouldn't make sense to set `INTERP_GCC` to either of those and would break the build.

In the interest of giving the user an idea of what's going on I've changed that part to print a notice if `CC` has been overridden but `INTERP_GCC` has not; it looks like the following and pauses for 5 seconds (to make it easier to spot):

```
****************
*
* CC=/usr/bin/gcc-4.8 (overridden), note that this build will also invoke another compiler that can be overridden: INTERP_GCC=gcc
*
****************
```

There are also `MINGW_CXX` and `PPC_GCC_CXX`, which are similar to `INTERP_GCC` which I missed earlier; these are handled same as above now, except they look for `CXX` being overridden.

runtime/makelib/mkconstants.mk.ftl
```diff
@@ -33,27 +33,137 @@ ifndef UMA_TARGET_PATH
...
+ifneq (default,$(origin CC))
+  ifndef INTERP_GCC
+    # If the user has overridden CC we'll want to let them know that INTERP_GCC exists.
+    ifndef PRINT_ONCE_INTERP_GCC
+      $(info ****************)
+      $(info *)
+      $(info * CC=$(CC) (overridden), note that this build will also invoke another compiler that can be overridden: INTERP_GCC=${uma.spec.tools.interp_gcc.name})
+      $(info *)
+      $(info ****************)
+      $(shell sleep 5)
```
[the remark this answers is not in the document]
ymanton: Sleep removed.

DanHeidinga:
Jenkins test sanity

DanHeidinga:
The JDK9 builds all failed due to freemarker.jar issues (see fix in pr #1406).

Relaunching just the 9 builds

DanHeidinga:
Jenkins test sanity jdk9

DanHeidinga:
Jenkins test sanity all jdk9

PR #1332, opened 2018-03-02
Updates final field access check

taliamccormick:
Final fields in java must be set by the class in which it is declared. Final instance variables must be set in every constructor of the class they are declared in (<init>) and final static variables by the static initializer of the class they are declared in (<clinit>).

Private fields/methods of a class can currently only be accessed by the class in which they are declared. Earlier access checks therefore assumed that checking if a field ref has private access to a final field was sufficient.

With the advent of nestmates (see JEP 181), private fields/methods of a class can be accessed by its fellow 'nest members' classes. The existing access check would therefore allow two classes within the same 'nest' to see eachothers' final fields within their own constructors.

This commit changes the access check from:
```
if ( (has private access) && (is <init> or <clinit>) )
```
to be
```
if ( (is same class) && (is <init> or <clinit>) )
```

Resolves: https://github.com/eclipse/openj9/issues/1331

### DIFF
--- a/runtime/vm/resolvesupport.cpp
+++ b/runtime/vm/resolvesupport.cpp
@@ -36,23 +36,24 @@
 #define MAX_STACK_SLOTS 255
 
 static void checkForDecompile(J9VMThread *currentThread, J9ROMMethodRef *romMethodRef, UDATA jitFlags);
-static bool finalFieldSetAllowed(J9VMThread *currentThread, bool isStatic, J9Method *method, J9Class *fieldClass, J9ROMFieldShape *field, UDATA jitFlags);
+static bool finalFieldSetAllowed(J9VMThread *currentThread, bool isStatic, J9Method *method, J9Class *fieldClass, J9Class *callerClass, J9ROMFieldShape *field, UDATA jitFlags);
 
 
 /**
 * @brief In class files with version 53 or later, setting of final fields is only allowed from initializer methods.
-* Note that this is called only after verifying that the calling class and declaring class are identical.
+* Note that this is called only after verifying that the calling class and declaring class share private access.
 *
 * @param currentThread the current J9VMThread
 * @param isStatic TRUE for static fields, FALSE for instance
 * @param method the J9Method performing the resolve, NULL for no access check
 * @param fieldClass the J9Class which declares the field
+* @param callerClass the J9Class which tries to access the field
 * @param field J9ROMFieldShape of the final field
 * @param jitFlags 0 for a runtime resolve, non-zero for JIT compilation or AOT resolves
 * @return true if access is legal, false if not
 */
 static bool
-finalFieldSetAllowed(J9VMThread *currentThread, bool isStatic, J9Method *method, J9Class *fieldClass, J9ROMFieldShape *field, UDATA jitFlags)
+finalFieldSetAllowed(J9VMThread *currentThread, bool isStatic, J9Method *method, J9Class *fieldClass, J9Class *callerClass, J9ROMFieldShape *field, UDATA jitFlags)
 {
 	bool legal = true;
 	/* NULL method means do not do the access check */
@@ -61,7 +62,10 @@ finalFieldSetAllowed(J9VMThread *currentThread, bool isStatic, J9Method *method,
 		if (romClass->majorVersion >= 53) {
 			J9ROMMethod *romMethod = J9_ROM_METHOD_FROM_RAM_METHOD(method);
 			J9UTF8 *name = J9ROMMETHOD_NAME(romMethod);
-			if (('<' != J9UTF8_DATA(name)[0]) || (J9UTF8_LENGTH(name) != strlen(isStatic ? "<clinit>" : "<init>"))) {
+			if (('<' != J9UTF8_DATA(name)[0])
+			|| (J9UTF8_LENGTH(name) != strlen(isStatic ? "<clinit>" : "<init>"))
+			|| ((fieldClass != callerClass) && (!J9ROMCLASS_IS_UNSAFE(callerClass->romClass)))
+			) {
 				if (0 == jitFlags) {
 					setIllegalAccessErrorFinalFieldSet(currentThread, isStatic, romClass, field, romMethod);
 				}
@@ -683,7 +687,7 @@ illegalAccess:
 					}
 					goto done;
 				}
-				if (!finalFieldSetAllowed(vmStruct, true, method, definingClass, field, jitFlags)) {
+				if (!finalFieldSetAllowed(vmStruct, true, method, definingClass, classFromCP, field, jitFlags)) {
 					staticAddress = NULL;
 					goto done;
 				} else { /* finalFieldSetAllowed */
@@ -888,7 +892,7 @@ illegalAccess:
 					}
 					goto done;
 				}
-				if (!finalFieldSetAllowed(vmStruct, false, method, definingClass, field, jitFlags)) {
+				if (!finalFieldSetAllowed(vmStruct, false, method, definingClass, classFromCP, field, jitFlags)) {
 					fieldOffset = -1;
 					goto done;
 				}


runtime/vm/resolvesupport.cpp
```diff
@@ -870,8 +869,7 @@ resolveInstanceFieldRefInto(J9VMThread *vmStruct, J9Method *method, J9ConstantPo
 			}
 
 			if ((resolveFlags & J9_RESOLVE_FLAG_FIELD_SETTER) != 0 && (modifiers & J9_JAVA_FINAL) != 0) {
-				checkResult = checkVisibility(vmStruct, classFromCP, definingClass, J9_JAVA_PRIVATE, resolveFlags);
-				if (checkResult < J9_VISIBILITY_ALLOWED) {
+				if ((classFromCP == definingClass) || (J9ROMCLASS_IS_UNSAFE(classFromCP->romClass) != 0)) {
```
DanHeidinga: This looks like it flips the check from saying: `If checkVisibility says no, fail` to `if checkVisible would have said yes, fail`.

How has this been tested?

taliamccormick: Accidentally pushed the wrong commit; fixed now.

I've been testing manually.

runtime/vm/resolvesupport.cpp
```diff
@@ -601,6 +601,8 @@ resolveStaticFieldRefInto(J9VMThread *vmStruct, J9Method *method, J9ConstantPool
 		UDATA modifiers = 0;
 		J9Class *localClassAndFlags = NULL;
 		UDATA initStatus = 0;
+		J9Class *currentClassFromCP = NULL;
```
tajila: can you define this in the scope that it's used in

taliamccormick: The compiler complains because of the 'illegalAccess' `goto` which is shortly after it

```
resolvesupport.cpp: In function 'void* resolveStaticFieldRefInto(J9VMThread*, J9Method*, J9ConstantPool*, UDATA, UDATA, J9ROMFieldShape**, J9RAMStaticFieldRef*)':
resolvesupport.cpp:675: error: jump to label 'illegalAccess'
resolvesupport.cpp:618: error:   from here
resolvesupport.cpp:669: error:   crosses initialization of 'J9Class* currentDefiningClass'
resolvesupport.cpp:668: error:   crosses initialization of 'J9Class* currentClassFromCP'
```

Should I take out the variable definitions and rewrite the check as:
```
if ((J9_CURRENT_CLASS(classFromCP) !=  J9_CURRENT_CLASS(definingClass)) && (J9ROMCLASS_IS_UNSAFE(v->romClass) == 0)) { ... }
```

DanHeidinga: That's odd.  Can you work around this by introducing a scope inside the if statement that defines and uses the locals?

taliamccormick: I restructured my changes so that the 'is this the same class' check occurs once within `finalFieldSetAllowed(...)` instead of within `resolveInstanceFieldRefInto(...)` and `resolveStaticFieldRefInto(...)`. 

The way that the VM handles illegal access to final fields is slightly different than the way it handles other illegal access (eg. non module access). Since the VM checks that a final field is accessed within a constructor occurs in `finalFieldSetAllowed`, I think it makes more sense to check at that stage if it is within the right constructor. 

I pushed the new changes, but am still in the process of retesting them

runtime/vm/resolvesupport.cpp
```diff
@@ -52,7 +52,7 @@ static bool finalFieldSetAllowed(J9VMThread *currentThread, bool isStatic, J9Met
 * @return true if access is legal, false if not
 */
 static bool
-finalFieldSetAllowed(J9VMThread *currentThread, bool isStatic, J9Method *method, J9Class *fieldClass, J9ROMFieldShape *field, UDATA jitFlags)
+finalFieldSetAllowed(J9VMThread *currentThread, bool isStatic, J9Method *method, J9Class *fieldClass, J9Class *methodClass, J9ROMFieldShape *field, UDATA jitFlags)
```
DanHeidinga: `methodClass` needs to be added to the `@param` section of the function description

DanHeidinga: `methodClass` -> `callerClass` or `setterClass` or .... It might be clearer to call this something other than `methodClass`.  

I had to do a double take on the `(fieldClass != methodClass)` check below purely due to the naming.

runtime/vm/resolvesupport.cpp
```diff
@@ -61,7 +61,9 @@ finalFieldSetAllowed(J9VMThread *currentThread, bool isStatic, J9Method *method,
 		if (romClass->majorVersion >= 53) {
 			J9ROMMethod *romMethod = J9_ROM_METHOD_FROM_RAM_METHOD(method);
 			J9UTF8 *name = J9ROMMETHOD_NAME(romMethod);
-			if (('<' != J9UTF8_DATA(name)[0]) || (J9UTF8_LENGTH(name) != strlen(isStatic ? "<clinit>" : "<init>"))) {
+			if (('<' != J9UTF8_DATA(name)[0])
+					|| (J9UTF8_LENGTH(name) != strlen(isStatic ? "<clinit>" : "<init>"))
+					|| ((fieldClass != methodClass) && (J9ROMCLASS_IS_UNSAFE(methodClass->romClass)))) {
```
DanHeidinga: Won't the logic here treat `(fieldClass != methodClass) == true` and `J9ROMCLASS_IS_UNSAFE(methodClass->romClass) == true` as an error case?

Shouldn't that be `J9ROMCLASS_IS_UNSAFE(methodClass->romClass) == 0`?

taliamccormick: From what I understand, the existing visibility checks called for final field access prior to this use are:
```
checkResult = checkVisibility(vmStruct, classFromCP, definingClass, J9_JAVA_PRIVATE, resolveFlags);
if (checkResult < J9_VISIBILITY_ALLOWED) { ... }
```
And the access checking within vm/visible.c uses the order:
```
UDATA result = J9_VISIBILITY_ALLOWED;
...
if (J9ROMCLASS_IS_UNSAFE(sourceClass->romClass) == 0) {
...
if (modifiers & J9AccPrivate) {
	/* Private */
	if ((sourceClass != destClass)
			&& (sourceClass->nestHost != destClass->nestHost)
#endif /* defined(J9VM_OPT_VALHALLA_NESTMATES) */
	) {
		result = J9_VISIBILITY_NON_MODULE_ACCESS_ERROR;
	}
...
return result;
```
I therefore assumed that if the class is unsafe or the class is itself, then we have access to private fields. So then if `(J9ROMCLASS_IS_UNSAFE(methodClass->romClass) == 0)` holds, then even with `fieldClass != methodClass` being true, we still have private field access.

If it does not hold hold (ie. `(J9ROMCLASS_IS_UNSAFE(methodClass->romClass) != 0)` - will correct formatting), that is when we require the classes to be the same (ie. `fieldClass != methodClass` must not be true). If the class is not unsafe and the final field is being set by a different class, then the illegal access error setting a final field should be set.

--

I'm still trying to get a test case to run correctly - should I close the PR and re-open it once I have done so?

runtime/vm/resolvesupport.cpp
```diff
@@ -61,7 +62,9 @@ finalFieldSetAllowed(J9VMThread *currentThread, bool isStatic, J9Method *method,
 		if (romClass->majorVersion >= 53) {
 			J9ROMMethod *romMethod = J9_ROM_METHOD_FROM_RAM_METHOD(method);
 			J9UTF8 *name = J9ROMMETHOD_NAME(romMethod);
-			if (('<' != J9UTF8_DATA(name)[0]) || (J9UTF8_LENGTH(name) != strlen(isStatic ? "<clinit>" : "<init>"))) {
+			if (('<' != J9UTF8_DATA(name)[0])
+					|| (J9UTF8_LENGTH(name) != strlen(isStatic ? "<clinit>" : "<init>"))
+					|| ((fieldClass != callerClass) && (J9ROMCLASS_IS_UNSAFE(callerClass->romClass)))) {
```
DanHeidinga: This needs to be `!J9ROMCLASS_IS_UNSAFE(callerClass->romClass)`.

The check should be saying "if you are different classes and not unsafe, then fail".

DanHeidinga: A formatting nitpick:
```
if (('<' != J9UTF8_DATA(name)[0])
|| (J9UTF8_LENGTH(name) != strlen(isStatic ? "<clinit>" : "<init>"))
|| ((fieldClass != callerClass) && (J9ROMCLASS_IS_UNSAFE(callerClass->romClass)))
) {
```
This makes it easier to visually see the clauses and where they end.

DanHeidinga:
Jenkins test sanity plinux

Issue #1207, opened 2018-02-18
ARM High cpu hang in floatSanityTests


JamesKingdon:
Attempting to run the openJ9 tests on arm I see a high cpu hang during

```
===============================================
Running test floatSanityTests_0 ...
===============================================
test with NoOptions
{ "/home/jkingdon/sdks/jdk-unsafePut64/bin/java"    \
-cp "/home/jkingdon/tests/openj9/openj9/test/TestConfig/scripts/testKitGen/../../../../jvmtest/TestConfig/resources:/home/jkingdon/tests/openj9/openj9/test/TestConfig/scripts/testKitGen/../../../../jvmtest/TestConfig/lib/testng.jar:/home/jkingdon/tests/openj9/openj9/test/TestConfig/scripts/testKitGen/../../../../jvmtest/TestConfig/lib/jcommander.jar:/home/jkingdon/tests/openj9/openj9/test/TestConfig/scripts/testKitGen/../../../../jvmtest/Java8andUp/GeneralTest.jar" \
org.testng.TestNG -d "/home/jkingdon/tests/openj9/openj9/test/TestConfig/scripts/testKitGen/../../../TestConfig/test_output_1518958271761/floatSanityTests_0" "/home/jkingdon/tests/openj9/openj9/test/TestConfig/scripts/testKitGen/../../../../jvmtest/Java8andUp/testng.xml" \
-testnames floatSanityTests \
-groups level.sanity \
-excludegroups d.*.linux_arm,d.*.arch.arm,d.*.os.linux,d.*.bits.32,d.*.generic-all; \
if [ $? -eq 0 ] ; then echo "floatSanityTests_0""_PASSED"; else echo "floatSanityTests_0""_FAILED"; fi; } 2>&1 | tee -a "/home/jkingdon/tests/openj9/openj9/test/TestConfig/scripts/testKitGen/../../../TestConfig/test_output_1518958271761/TestTargetResult"
[XmlSuite] [WARN] 'parallel' value 'false' is deprecated, default value will be used instead: 'none'.
[IncludeExcludeTestAnnotationTransformer] [INFO] exclude file is /home/jkingdon/tests/openj9/openj9/test/TestConfig/../../jvmtest/TestConfig/resources/excludes/latest_exclude_SE90.txt
...
... TestNG 6.10 by C?dric Beust (cedric@beust.com)
...

[TestNG] Running:
[...]

PR #1339, opened 2018-03-03
Use native cast for int to float on ARM

JamesKingdon:
Some int to float conversions hang when running on armhf. The problem appears
to be with gcc-linaro-4.9.4-2017.01 which generates incorrect code for the
while loop in fltconv_indexLeadingOne32. Then entire code path can be avoided
by enabling native cast in helperCConvertIntegerToFloat, which should also
improve performance.

Fixes: #1207 

### DIFF
--- a/runtime/util/fltconv.c
+++ b/runtime/util/fltconv.c
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 1991, 2017 IBM Corp. and others
+ * Copyright (c) 1991, 2018 IBM Corp. and others
  *
  * This program and the accompanying materials are made available under
  * the terms of the Eclipse Public License 2.0 which accompanies this
@@ -277,8 +277,11 @@ jfloat
 helperCConvertIntegerToFloat(I_32 src)
 {
 	jfloat tmpDst;
-	
-#if defined USE_NATIVE_CAST
+
+#if defined(USE_NATIVE_CAST) || defined(ARM)
+	/* Enabling native cast for ARM avoids a compiler bug in
+	 * gcc-linaro-4.9.4-2017.01
+	 */
 	tmpDst = (jfloat)src;
 #else
 	{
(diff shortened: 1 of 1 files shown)


JamesKingdon:
fixes #1207

runtime/util/fltconv.c
```diff
@@ -277,8 +277,10 @@ jfloat
 helperCConvertIntegerToFloat(I_32 src)
 {
 	jfloat tmpDst;
-	
-#if defined USE_NATIVE_CAST
+
+#if defined(USE_NATIVE_CAST) || defined(ARM)
+	// Enabling native cast for ARM avoids a compiler bug in
```
DanHeidinga: We avoid `//` comments in C code due to issues with some compilers.  The C code should still be compliant with `c89` due to certain supported platforms.

JamesKingdon: ok, will do.

JamesKingdon:
@DanHeidinga Can we get this merged now? It's one of the main problems that will show up if we enable armhf in the builds.

runtime/util/fltconv.c
```diff
@@ -424,6 +427,10 @@ fltconv_indexLeadingOne32(U_32 u32val)
 		leading = 7;
 		mask = 0x00000080;
 	}
+	// WARNING This loop will busy hang when compiled with
+	// gcc-linaro-4.9.4-2017.01 for armhf if the match should happen on
+	// the first iteration. The loop exit condition is moved to the end of
+	// the loop and doesn't execute before the first shift of mask.
```
fjeremic: This comment still needs `/* */`

JamesKingdon: Well, bother. Will fix. Thanks for catching.

JamesKingdon:
Thanks Peter. Build: https://github.com/JamesKingdon/openj9-openjdk-jdk10/releases/tag/20180429170711-71d04615
I've started an openJ9 _functional run with it.

PR #1344, opened 2018-03-05
Refactor check visiblity code

tajila:
Refactor check visiblity code

Split out the modularity access checks from the check visibility code so it can be called separately. Modify the tracepoints to output romClasses and J9Module info. J9Module also knows about the classLoader so we can determine runtime types from trace output. 

### DIFF
--- a/runtime/util/module.tdf
+++ b/runtime/util/module.tdf
@@ -1,4 +1,4 @@
-// Copyright (c) 2017, 2017 IBM Corp. and others
+// Copyright (c) 2017, 2018 IBM Corp. and others
 //
 // This program and the accompanying materials are made available under
 // the terms of the Eclipse Public License 2.0 which accompanies this
@@ -41,3 +41,4 @@ TraceEvent=Trc_MODULE_add_module_package Overhead=1 Level=5 Test Template="JVM_A
 TraceEvent=Trc_MODULE_setting_package Overhead=1 Level=5 Test Template="internally set package/class: %.*s, loader: %s, module: %s"
 
 TraceException=Trc_VM_checkVisibility_failed_with_errortype Overhead=1 Level=1 Template="checkVisibility from %p (%.*s) to %p (%.*s) failed, error type: %s"
+TraceException=Trc_VM_checkVisibility_failed_with_errortype_romclass Overhead=1 Level=1 Template="checkVisibility from romClass=%p (%.*s) j9mod=%p to romClass=%p (%.*s) j9mod=%p failed, packagedef'n errcode=%zu, error type: %s"
--- a/runtime/vm/createramclass.cpp
+++ b/runtime/vm/createramclass.cpp
@@ -1496,15 +1496,10 @@ loadSuperClassAndInterfaces(J9VMThread *vmThread, J9ClassLoader *classLoader, J9
 	UDATA packageID, BOOLEAN hotswapping, UDATA classPreloadFlags, J9Class **superclassOut, J9Module *module)
 {
 	J9JavaVM *vm = vmThread->javaVM;
-	J9InternalVMFunctions *vmFuncs = vm->internalVMFunctions;
 	const BOOLEAN isROMClassUnsafe = (J9ROMCLASS_IS_UNSAFE(romClass) != 0);
 	J9UTF8 *className = J9ROMCLASS_CLASSNAME(romClass);
 	J9UTF8 *superclassName = NULL;
 	J9Class *superclass = NULL;
-	J9Class thisClass = {0};
-	thisClass.module = module;
-	thisClass.packageID = packageID;
-	thisClass.romClass = romClass;
 
 	superclassName = J9ROMCLASS_SUPERCLASSNAME(romClass);
 	if (superclassName == NULL) {
@@ -1550,7 +1545,18 @@ loadSuperClassAndInterfaces(J9VMThread *vmThread, J9ClassLoader *classLoader, J9
 
 			/* ensure that the superclass is visible */
 			if (!isROMClassUnsafe) {
-				if (J9_VISIBILITY_ALLOWED != vmFuncs->checkVisibility(vmThread, &thisClass, superclass, superclass->romClass->modifiers, 0)) {
+				/*
+				 * Failure occurs if
+				 * 1) The superClass class not public and does not belong to
+				 * the same package as the class being loaded.
+				 * 2) The superClass class is public but the class being loaded
+				 * belongs to a module that doesn't have access to the module that
+				 * owns the superClass class.
+				 */
+				bool superClassIsPublic = J9_ARE_ALL_BITS_SET(superclass->romClass->modifiers, J9_JAVA_PUBLIC);
+				if ((!superClassIsPublic && (packageID != superclass->packageID))
+				|| (superClassIsPublic && !checkModuleAccess(vmThread, vm, romClass, module, superclass->romClass, superclass->module, superclass->packageID, 0))
+				) {
 					Trc_VM_CreateRAMClassFromROMClass_superclassNotVisible(vmThread, superclass, superclass->classLoader, classLoader);
 					setCurrentExceptionForBadClass(vmThread, superclassName, J9VMCONSTANTPOOL_JAVALANGILLEGALACCESSERROR);
 					return FALSE;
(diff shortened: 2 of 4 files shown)


runtime/vm/visible.c
```diff
@@ -142,3 +170,6 @@ checkVisibility(J9VMThread *currentThread, J9Class* sourceClass, J9Class* destCl
 
 	return result;
 }
+
```
DanHeidinga: nitpick: extra blank lines

runtime/vm/visible.c
```diff
@@ -29,6 +29,62 @@
...
+ * @param[in] destModule the module of the dest class
+ * @param[in] destPackageID packageID of the dest class
+ * @param[in] lookupOptions J9_LOOK* options
+ *
+ * @return 	J9_VISIBILITY_ALLOWED if the access is allowed,
+ * 			J9_VISIBILITY_MODULE_READ_ACCESS_ERROR if module read access error occurred,
+ * 			J9_VISIBILITY_MODULE_PACKAGE_EXPORT_ERROR if module package access error
+ */
+VMINLINE IDATA
+checkModuleAccess (J9VMThread *currentThread, J9JavaVM* vm, J9ROMClass* srcRomClass, J9Module* srcModule, J9ROMClass* destRomClass, J9Module* destModule, UDATA destPackageID, UDATA lookupOptions)
```
DanHeidinga: nipick: extra space between function name and opening `(`

runtime/vm/visible.c
```diff
@@ -29,6 +29,62 @@
...
+			const U_8* packageName = NULL;
+			UDATA packageNameLength = 0;
+			J9PackageIDTableEntry entry = {0};
+
+			entry.taggedROMClass = destPackageID;
+			packageName = getPackageName(&entry, &packageNameLength);
+
+			omrthread_monitor_enter(vm->classLoaderModuleAndLocationMutex);
+			if (!isPackageExportedToModuleWithName(currentThread, destModule, (U_8*) packageName, (U_16) packageNameLength, srcModule, J9_IS_J9MODULE_UNNAMED(vm, srcModule), &rc)) {
+				Trc_VM_checkVisibility_failed_with_errortypeRomClass(currentThread,
```
DanHeidinga: Should this tracepoint include the `rc` value?  Otherwise, it should be named something like `ignored`.

runtime/vm/visible.c
```diff
@@ -29,6 +29,62 @@
...
+ */
+VMINLINE IDATA
+checkModuleAccess (J9VMThread *currentThread, J9JavaVM* vm, J9ROMClass* srcRomClass, J9Module* srcModule, J9ROMClass* destRomClass, J9Module* destModule, UDATA destPackageID, UDATA lookupOptions)
+{
+	UDATA result = J9_VISIBILITY_ALLOWED;
+	if (srcModule != destModule) {
+		UDATA rc = ERRCODE_GENERAL_FAILURE;
+		if (!J9_ARE_ALL_BITS_SET(lookupOptions, J9_LOOK_REFLECT_CALL)) {
+			if (!isAllowedReadAccessToModule(currentThread, srcModule, destModule, &rc)) {
+				Trc_VM_checkVisibility_failed_with_errortypeRomClass(currentThread,
```
DanHeidinga: Should this tracepoint report the `rc` value?

runtime/vm/visible.c
```diff
@@ -29,6 +29,62 @@
 #include "util_api.h"
 #include "vm_internal.h"
 
+
+
+/**
+ * Check module access from srcModule to destModule
```
DanHeidinga: Can you expand on the checks that occur?  The reflective vs not difference,  unnamed modules, etc?  A good comment here can save significant debugging time later.

runtime/util/module.tdf
```diff
@@ -41,3 +41,4 @@ TraceEvent=Trc_MODULE_add_module_package Overhead=1 Level=5 Test Template="JVM_A
 TraceEvent=Trc_MODULE_setting_package Overhead=1 Level=5 Test Template="internally set package/class: %.*s, loader: %s, module: %s"
 
 TraceException=Trc_VM_checkVisibility_failed_with_errortype Overhead=1 Level=1 Template="checkVisibility from %p (%.*s) to %p (%.*s) failed, error type: %s"
+TraceException=Trc_VM_checkVisibility_failed_with_errortypeRomClass Overhead=1 Level=1 Template="checkVisibility from romClass=%p (%.*s) j9mod=%p to romClass=%p (%.*s) j9mod=%p failed, error type: %s"
```
DanHeidinga: `Trc_VM_checkVisibility_failed_with_errortypeRomClass` -> `Trc_VM_checkVisibility_failed_with_errortype_romclass`

runtime/vm/createramclass.cpp
```diff
@@ -1550,7 +1545,20 @@ loadSuperClassAndInterfaces(J9VMThread *vmThread, J9ClassLoader *classLoader, J9
...
+				 * the same package as the class being loaded.
+				 * 2) The superClass class is public but the class being loaded
+				 * belongs to a module that doesn't have access to the module that
+				 * owns the superClass class.
+				 */
+				bool superClassIsPublic = J9_ARE_ALL_BITS_SET(superclass->romClass->modifiers, J9_JAVA_PUBLIC);
+				if ((!superClassIsPublic && (packageID != superclass->packageID))
+					|| (superClassIsPublic
+						&& !checkModuleAccess(vmThread, vm, romClass, module, superclass->romClass, superclass->module, superclass->packageID, 0)
+					)
```
DanHeidinga: This brace shouldn't be on it's own line.

Try something like:
```
if ((!superClassIsPublic && (packageID != superclass->packageID))
|| (superClassIsPublic && !checkModuleAccess(vmThread, vm, romClass, module, superclass->romClass, superclass->module, superclass->packageID, 0))
) {
```

DanHeidinga: Same below

runtime/vm/visible.c
tajila: the RC value is not necessary as the string last arg) reports the type of failure

runtime/vm/visible.c
tajila: by `RC` do you mean `result`?

`result` is not passed in to the tracepoint

tajila:
@DanHeidinga Ive updated the patch. I had a question about your `RC` comment, not sure if it is still visible.

By `RC` I assume you meant `result` ?

The result is currently not part of the tracepoint. The last arg in the tracepoint is a string that describes the failure type, so adding the `result` just duplicates the info.

DanHeidinga:
There's an `rc` variable passed to `isPackageExportedToModuleWithName` and to `isAllowedReadAccessToModule` which is never read.  It may provide useful debugging information about why the access check failed, especially if due to environment factors (nativeOOM, etc)

tajila:
Thanks, I've updated the PR.

runtime/vm/visible.c
```diff
@@ -29,6 +29,70 @@
 #include "util_api.h"
 #include "vm_internal.h"
 
+
+
+/**
```
DanHeidinga: Thanks for expanding the comment.  You've got all the necessary info here.  I think it would be easier to read if formatted slightly differently:
```
* Check module access from srcModule to destModule. 
*
* The algorithm for validating module access is:
* 1) check to see if the source module `reads` the dest module. 
* 2) check to see if the dest module exports the package (that 
* dest class belongs to) to the source module. 
*
* For reflective calls, the rules are slightly different as all reflect
* reflect accesses implicitly have read access.  Set the 
*  J9_LOOK_REFLECT_CALL flag in the lookup options for reflective
* checks.
* 
* The unnamedModules export all the packages they own and have read access to all modules
* they require access to. 
```

DanHeidinga: Apart from this minor formatting change, this looks good.

DanHeidinga:
Jenkins test sanity plinux

DanHeidinga:
@tajila The tracepoint isn't valid
```
16:29:02 TDFParser.cpp:733 ERROR : %l is not valid modifier. (parameter 8 in mutated format string checkVisibility from romClass=%p (%.*s) j9mod=%p to romClass=%p (%.*s) j9mod=%p failed, packagedef'n errcode=%lu, error type: %s). util/module.tdf:44
16:29:02 Failed to parse trace line TraceException=Trc_VM_checkVisibility_failed_with_errortype_romclass Overhead=1 Level=1 Template="checkVisibility from romClass=%p (%.*s) j9mod=%p to romClass=%p (%.*s) j9mod=%p failed, packagedef'n errcode=%lu, error type: %s"
```

tajila:
@DanHeidinga fixed the trace issue

DanHeidinga:
I have to dig further - is `%zu` the correct modifier for a UDATA sized variable?  If you have a link that shows its the right thing, this will be even faster :)

tajila:
Officially, <a href="http://www.cplusplus.com/reference/cstdio/printf/">`%z` is for size_t</a>. However, the convention in other trace files is to use `%zu` for `UDATA`.  ie `Trc_VM_growJavaStack_TooSmall`

DanHeidinga:
Jenkins test sanity plinux

PR #1345, opened 2018-03-05
Add shared class CML test cases

hangshao0:
1. Add a CML test case that classes are being stored/found in shared
cache when using class path. 
2. Add a CML test case classes are being stored/found in shared cache
when using module path. 


hangshao0:
reviewer: @llxia

hangshao0:
I made a change a few days ago in my personal fork that accidentally broke finding classes in the shared cache for the boot class loader. I am surprised that change was able to pass all the sanity test. So I am adding these test cases here.

llxia:
Merge the change as the failure in travis ci build is not related.
In the future, please use [ci skip] for the changes that does not need travis ci build. Thanks.

PR #1335, opened 2018-03-02
Reduce use of j9vm/jvm.h

keithc-ca:
* don't include jvm.h where it's not needed
* add module dependency where it is needed
Supersedes #1326.

PR #1350, opened 2018-03-06
Fix another dependency on generated.h

pshipton:
Caused by #1009. Other dependencies fixed by #1335


pshipton:
I don't believe this change will be tested by PR testing, but just in case...

jenkins compile plinux jdk8

keithc-ca:
Looks good. Please squash before I merge.

Issue #660, opened 2017-11-22
Add JDK8 PR builds to repos OpenJ9, OpenJ9-OMR and OpenJ9-OpenJDK-JDK8 and nightly builds


AdamBrousseau:
OpenJ9 with [openjdk-jdk8](https://github.com/ibmruntimes/openj9-openjdk-jdk8) is now available.

Builds we need to add to [Jenkins](https://ci.eclipse.org/openj9) for jdk8 (as per the jdk9 list we currently have)
#### Nightlies
- [x] Build zLinux [8](https://ci.eclipse.org/openj9/view/Build/job/Build-JDK8-linux_390-64_cmprssptrs/) [![Build Status](https://ci.eclipse.org/openj9/job/Build-JDK8-linux_390-64_cmprssptrs/badge/icon)](https://ci.eclipse.org/openj9/job/Build-JDK8-linux_390-64_cmprssptrs) & [9](https://ci.eclipse.org/openj9/view/Build/job/Build-JDK9-linux_390-64_cmprssptrs/) [![Build Status](https://ci.eclipse.org/openj9/job/Build-JDK9-linux_390-64_cmprssptrs/badge/icon)](https://ci.eclipse.org/openj9/job/Build-JDK9-linux_390-64_cmprssptrs)
- [x] Build pLinux [8](https://ci.eclipse.org/openj9/view/Build/job/Build-JDK8-linux_ppc-64_cmprssptrs_le/) [![Build Status](https://ci.eclipse.org/openj9/job/Build-JDK8-linux_ppc-64_cmprssptrs_le/badge/icon)](https://ci.eclipse.org/openj9/job/Build-JDK8-linux_ppc-64_cmprssptrs_le) & [9](https://ci.eclipse.org/openj9/view/Build/job/Build-JDK9-linux_ppc-64_cmprssptrs_le/) [![Build Status](https://ci.eclipse.org/openj9/job/Build-JDK9-linux_ppc-64_cmprssptrs_le/badge/icon)](https://ci.eclipse.org/openj9/job/Build-JDK9-linux_ppc-64_cmprssptrs_le)
[...]

Issue #827, opened 2017-12-14
PR builds: Also support "depends" on a repo/branch


AdamBrousseau:
Currently we can launch PR builds with dependent changes that are in an open PR at omr, openj9-omr or openj9-openjdk-jdk9. There is a use case to launch a PR build against a branch instead of an open PR. For example, when the dependent or breaking PR has been merged in OMR and can't promote through OMR-Acceptance.

More info on dependent changes
https://github.com/eclipse/openj9/blob/master/buildenv/jenkins/README.md#dependent-changes

If we can temporarily survive without this feature, it would be nice to finish off #660, #851 first. Which will commit all the build code in.

DanHeidinga:
We've survived this long without it so I view as an enhancement.  Delaying until those PRs are merged is fine as long we are able to close those out in a reasonable time frame

DanHeidinga:
Bumping the priority on this as we're seeing more OpenJ9 PRs with dependent OMR changes that can't be built as the OMR change has already been merged.

PR #1354, opened 2018-03-06
Support PullRequest builds from other repos

AdamBrousseau:
- Determine if build is a PR
- Automatically checkout PR and any dependent PRs
- Support dependent changes from other repos

[ci skip]

Issue #660

#### TODO
- [x] Update README
- [x] Open Eclipse [Bugzilla](https://bugs.eclipse.org/bugs/show_bug.cgi?id=532195) to request Jenkins Webhook be added to openj9-omr repo

### DIFF
--- a/buildenv/jenkins/README.md
+++ b/buildenv/jenkins/README.md
@@ -56,14 +56,16 @@ This folder contains Jenkins pipeline scripts that are used in the OpenJ9 Jenkin
     - `Jenkins test all`
     - `Jenkins test all all all`
 
-You can also request a "Compile & Sanity" build from the extensions repos or openj9-omr
+You can also request a Pull Request build from the extensions repos or openj9-omr
+- [openj9-omr](https://github.com/eclipse/openj9-omr)
 - [openj9-openjdk-jdk8](https://github.com/ibmruntimes/openj9-openjdk-jdk8)
 - [openj9-openjdk-jdk9](https://github.com/ibmruntimes/openj9-openjdk-jdk9)
-- [openj9-omr](https://github.com/eclipse/openj9-omr)
+
+###### Note: When specifying a dependent change in an OpenJDK extensions repo, you can only build the SDK version that matches the repo where the dependent change lives. Eg. You cannot build JDK8 with a PR in openj9-openjdk-jdk9.
 
 ##### Dependent Changes
 
-- If you have dependent change(s) in either eclipse/omr, eclipse/openj9-omr, or ibmruntimes/openj9-openjdk-jdk9, you can build & test with all needed changes
+- If you have dependent change(s) in either eclipse/omr, eclipse/openj9-omr, or ibmruntimes/openj9-openjdk-jdk\*, you can build & test with all needed changes
 - Request a build by including the PR ref in your trigger comment
 - Ex. Dependent change in OMR Pull Request `#123`
     - `Jenkins test sanity depends eclipse/omr#123`
@@ -72,11 +74,9 @@ You can also request a "Compile & Sanity" build from the extensions repos or ope
 - Ex. Dependent change in OpenJDK Pull Request `#789`
     - `Jenkins test sanity depends ibmruntimes/openj9-openjdk-jdk9#789`
 - Ex. Dependent changes in OMR and OpenJDK
-    - `Jenkins test sanity depends eclipse/omr#123 ibmruntimes/openj9-openjdk-jdk9#789`
+    - `Jenkins test sanity all jdk9 depends eclipse/omr#123 ibmruntimes/openj9-openjdk-jdk9#789`
 - Ex. If you have a dependent change and only want one platform, depends comes last
-    - `Jenkins test sanity zlinux depends eclipse/omr#123`
-
-**Note:** Dependent changes can only be requested from an OpenJ9 Pull Request
+    - `Jenkins test sanity zlinux all depends eclipse/omr#123`
 
 ##### Other Pull Requests builds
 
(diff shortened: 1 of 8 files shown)


AdamBrousseau:
adam test sanity plinux

AdamBrousseau:
adam test sanity plinux jdk9

AdamBrousseau:
adam test sanity plinux jdk9

buildenv/jenkins/common/build
```diff
@@ -50,102 +50,154 @@ def get_source() {
...
+            omr_bool = true
+            break
+        case "eclipse/openj9":
+            openj9_bool = true
+            break
+        case ~/ibmruntimes(.*)/:
+            openjdk_bool = true
+            break
+        default:
+            error "Unable to determine sourece repo for PullRequest"
```
DanHeidinga: `sourece` -> `source`

AdamBrousseau:
adam test sanity plinux jdk9

AdamBrousseau:
adam test sanity plinux jdk9

AdamBrousseau:
adam test sanity plinux jdk9

AdamBrousseau:
adam test sanity plinux jdk9

AdamBrousseau:
adam test sanity plinux jdk9

AdamBrousseau:
adam test sanity plinux jdk9

AdamBrousseau:
adam test sanity plinux jdk9 depends eclipse/omr#2357 ibmruntimes/openjdk-jdk8#58

AdamBrousseau:
adam test sanity plinux jdk9 depends eclipse/omr#2357 ibmruntimes/openj9-openjdk-jdk8#58

AdamBrousseau:
[Regular OpenJ9 PR](https://ci.eclipse.org/openj9/view/Sandbox/job/tmp_PullRequest-Sanity-JDK9-linux_ppc-64_cmprssptrs_le-OpenJ9/19/console)
[OpenJ9 PR with 2 dependent changes](https://ci.eclipse.org/openj9/view/Sandbox/job/tmp_PullRequest-Sanity-JDK9-linux_ppc-64_cmprssptrs_le-OpenJ9/21/console)
[OpenJDK8 PR with dependent 2 changes](https://ci.eclipse.org/openj9/view/Sandbox/job/tmp_PullRequest-Sanity-JDK9-linux_ppc-64_cmprssptrs_le-OpenJ9/23/console)

Do we want to protect against people specifying dependent changes from the incorrect version of an openjdk-extensions repo?
eg. `jenkins test sanity all jdk9 depends ibmruntimes/openj9-openjdk-jdk8#123`

pshipton:
> Do we want to protect against people specifying dependent changes from the incorrect version of an openjdk-extensions repo?

I think so. Better than ignoring the dependent change and running the build anyway. How would it work? Perhaps running a job that immediately fails with an explanation of the problem.

AdamBrousseau:
adam test sanity plinux jdk9 depends eclipse/omr#2357 ibmruntimes/openj9-openjdk-jdk8#58

AdamBrousseau:
adam test sanity plinux jdk9 depends eclipse/omr#2357 ibmruntimes/openj9-openjdk-jdk8#58

AdamBrousseau:
adam test sanity plinux jdk9 depends eclipse/omr#2357 ibmruntimes/openj9-openjdk-jdk9#108

AdamBrousseau:
[Dependent JDK change with correct SDK version](https://ci.eclipse.org/openj9/view/Sandbox/job/tmp_PullRequest-Sanity-JDK9-linux_ppc-64_cmprssptrs_le-OpenJ9/24/console)
[Dependent JDK change with wrong SDK version](https://ci.eclipse.org/openj9/view/Sandbox/job/tmp_PullRequest-Sanity-JDK9-linux_ppc-64_cmprssptrs_le-OpenJ9/25/console)

AdamBrousseau:
@vsebe for review

vsebe:
desk review

AdamBrousseau:
updated

AdamBrousseau:
adam test sanity plinux jdk9 depends eclipse/omr#2357

AdamBrousseau:
recent adjustment to switch-case still works
https://ci.eclipse.org/openj9/view/Sandbox/job/tmp_PullRequest-Sanity-JDK9-linux_ppc-64_cmprssptrs_le-OpenJ9/26/console

AdamBrousseau:
updated

vsebe:
LGTM

AdamBrousseau:
Ready for committer review @pshipton or @DanHeidinga

keithc-ca:
I'm not convinced this is a good direction to go. Rather than adjust the infrastructure to manage pull requests that must be coordinated across multiple repositories, I think we should aim to have pull requests that can be merged sequentially (first omr then openj9, or vice-versa). It may require a little more thought and work on the part of the author of the PRs, but I am confident that those PRs can always be constructed so there is an order they can be merged without build breakage at any intermediate state.

It can also get very complicated: consider a PR that is connected to openj9-openjdk changes. Currently this PR will only let one specify dependencies on one extension repo. If changes are required for jdk8, it's likely it also affects jdk9 (and jdk10, etc. - something not supported by the current proposal): where do we draw the line? I say that line should be at zero dependencies.

AdamBrousseau:
@keithc-ca I agree that not having/needing dependent changes is a good idea. However, this PR is not about adding support for dependent changes, it is simply extending PR builds to other repos, which in turn extends the dependent change support. Also, there is an open issue (#827) to extend the "depends" support to also work with branches rather than just open PRs.

AdamBrousseau:
If we want to discuss the validity of having dependent changes can we open a separate issue where the team leads can discuss? I don't think that conversation should be a blocker to this PR.
@keithc-ca @DanHeidinga

keithc-ca:
On the contrary, I don't think we should be complicating things (like merging this PR would) until we have agreed that we want to consider PRs that come with dependencies.

AdamBrousseau:
Keep in mind, we already support having pull requests with dependent changes. The purpose of this PR is extending PR Build functionality to the extensions repos. The only added functionality is the protection against cross-version dependent changes, which you could argue should be in a different PR as we actually have this vulnerability today, without these changes.

keithc-ca:
I still don't think we should be doing this. If (and that a big *if*) we want to consider dependent PRs involving the extension repos, then it needs to allow specification of *all* the extension repos, not at most one.

AdamBrousseau:
https://github.com/eclipse/openj9/blob/master/buildenv/jenkins/README.md#dependent-changes

pshipton:
There are times when having dependent changes is just easier. Perhaps they could be committed in multiple steps to avoid the dependencies, but why would we want to do that dance when the changes can be committed (very closely) together. Seems like a waste of developer, committer, and machine time to avoid a very small timing hole. Dependent omr changes can be problematic because of the omr acceptance build, but there is no such issue for committing dependencies between openj9 and the extensions repos. 

With this PR, builds for all the extensions repos can be done, it just requires multiple commands. This could be improved in the future, if there is nothing better to do, but its not a requirement that needs to block merging this.

DanHeidinga:
@pshipton I agree with you this is the right approach for now but have to take issue with:

> but why would we want to do that dance when the changes can be committed (very closely) together. Seems like a waste of developer, committer, and machine time to avoid a very small timing hole. 

That small timing hole leaves the repos in a broken state.  It stops tools like `git bisect` from running.  It leaves other developers with potentially broken repos.

Dependent changes externalize the cost of delivering a change from the dev making the change to everyone consuming the project.  It's not a great trade off.

Unfortunately, there are some dependent changes that we can't avoid.  We need this as a tool for those (rare) cases.  

Committers should continue to encourage all developers to stage their changes so we don't need this kind of tool.

buildenv/jenkins/README.md
```diff
@@ -72,11 +74,9 @@ You can also request a "Compile & Sanity" build from the extensions repos or ope
 - Ex. Dependent change in OpenJDK Pull Request `#789`
     - `Jenkins test sanity depends ibmruntimes/openj9-openjdk-jdk9#789`
```
DanHeidinga: Isn't this line missing `all jdk9` as well?

buildenv/jenkins/README.md
```diff
@@ -72,11 +74,9 @@ You can also request a "Compile & Sanity" build from the extensions repos or ope
...
 - Ex. Dependent changes in OMR and OpenJDK
-    - `Jenkins test sanity depends eclipse/omr#123 ibmruntimes/openj9-openjdk-jdk9#789`
+    - `Jenkins test sanity all jdk9 depends eclipse/omr#123 ibmruntimes/openj9-openjdk-jdk9#789`
```
DanHeidinga: This fixes one of these - aren't other commands incorrect as well?

pshipton:
@DanHeidinga agreed. Good point, the timing hole I mentioned can sometimes be avoided by committing the changes in the correct order, even if they are within a few seconds of each other.

keithc-ca:
I hadn't thought of the use cases where it simply makes life easier for a committer - those make sense. Ultimately, a committer has the responsibility to ensure that merges go smoothly and if this helps, I'm fine with it.

PR #1360, opened 2018-03-06
Fix potential null element in AccessControlContext.context

JasonFengJ9:
Fix potential `null` element in `AccessControlContext.context`

Added `null` check for the `ProtectionDomain` object before adding it into an array and pass to an `AccessControlContext` constructor.

Reviewer @pshipton 
FYI: @DanHeidinga 

### DIFF
--- a/jcl/src/java.base/share/classes/java/lang/J9VMInternals.java
+++ b/jcl/src/java.base/share/classes/java/lang/J9VMInternals.java
@@ -333,7 +333,9 @@ final class J9VMInternals {
 	private static void checkPackageAccess(final Class clazz, ProtectionDomain pd) {
 		final SecurityManager sm = System.getSecurityManager();
 		if (sm != null) {
-			AccessController.doPrivileged(new PrivilegedAction() {
+			ProtectionDomain[] pdArray = (pd == null) ? new ProtectionDomain[]{} : new ProtectionDomain[]{pd};
+			AccessController.doPrivileged(new PrivilegedAction<Object>() {
+				@Override
 				public Object run() {
 					String packageName = clazz.getPackageName();
 					if (packageName != null) {
@@ -346,7 +348,7 @@ final class J9VMInternals {
 					}					
 					return null;
 				}
-			}, new AccessControlContext(new ProtectionDomain[]{pd}));
+			}, new AccessControlContext(pdArray));
 		}
 	}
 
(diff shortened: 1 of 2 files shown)


jcl/src/java.base/share/classes/java/lang/J9VMInternals.java
```diff
@@ -333,7 +333,9 @@ private static void threadCleanup(Thread thread) {
 	private static void checkPackageAccess(final Class clazz, ProtectionDomain pd) {
 		final SecurityManager sm = System.getSecurityManager();
 		if (sm != null) {
-			AccessController.doPrivileged(new PrivilegedAction() {
+			ProtectionDomain[] pdArray = (pd == null) ? new ProtectionDomain[]{} : new ProtectionDomain[]{pd};
```
pshipton: I don't think creation of an empty ProtectionDomain[] is required, null should be sufficient.

JasonFengJ9: There are constructors not checking `null` for incoming `ProtectionDomain` array. 
Will modify them instead.

JasonFengJ9: Actually incoming `ProtectionDomain` array can't be `null`.
https://docs.oracle.com/javase/9/docs/api/java/security/AccessControlContext.html#AccessControlContext-java.security.ProtectionDomain:A-
```
public AccessControlContext​(ProtectionDomain[] context)

Create an AccessControlContext with the given array of ProtectionDomains. Context must not be null. Duplicate domains will be removed from the context.

Parameters:
    context - the ProtectionDomains associated with this context. The non-duplicate domains are copied from the array. Subsequent changes to the array will not affect this AccessControlContext.
Throws:
    NullPointerException - if context is null 
```

pshipton: AccessControlContext​(ProtectionDomain[]) isn't being called, an internal constructor is being called and it seems to handle null for the ProtectionDomain[]. There is already code which sets AccessControlContext.context = null

pshipton: Sorry, I commented on the wrong part of the change. I meant this comment for the callers of AccessControlContext(AccessController, ProtectionDomain[], int)

JasonFengJ9: Updated PR accordingly, verified manually that `pConfig` still compiles.
Pls have another look.

pshipton:

PR #1366, opened 2018-03-06
Vectorize string case conversion on x86

cathyzhyi:
This changeset vectorizes string case conversion on x86. It also
implements the common code that calculating the array address
and string length in tree level so that the result can be shared
among code generators on all platforms.

### DIFF
--- a/runtime/compiler/trj9/optimizer/J9RecognizedCallTransformer.cpp
+++ b/runtime/compiler/trj9/optimizer/J9RecognizedCallTransformer.cpp
@@ -42,6 +42,98 @@
 #include "codegen/CodeGenerator.hpp"                      // for CodeGenerator
 #include "optimizer/TransformUtil.hpp"
 
+/*
+ * \brief transform the arguments of case conversion methods to make the call node friendly to code generator
+ *
+ * \details
+ *    The original case conversion method call has 2 children: the src string obj and the dst string obj which will be 
+ *    the converted string obj.
+ *    After the transformation, the 2 children are replaced with 3 new children: the src string array address, the dst string 
+ *    array address and the string length. The code generator can then focus on converting the actual content of the array and
+ *    doesn't need to calculate the various offsets on different platforms.
+ *
+ *    Before the transformation:
+ *    icall
+ *       =>aload src string obj
+ *       =>aload dst string obj
+ *
+ *    After the transformation:
+ *    icall
+ *       =>aladd
+ *          =>aloadi String.value 
+ *             =>aload src string obj
+ *          lonst arrayheaderoffset  
+ *       =>aladd
+ *          =>aloadi String.value 
+ *             =>aload dst string obj
+ *          lonst arrayheaderoffset  
+ *       =>aloadi String.count
+ *          =>aload src string obj
+ */
+void J9::RecognizedCallTransformer::processCaseConversion(TR::Node* node)
+   {
+   TR::Node *srcStringObject = node->getFirstChild();
+   TR::Node *dstStringObject = node->getSecondChild();
+
+   /* Offset of java/lang/String.value */
+   uint32_t stringValueFieldOffset = 0;
+   char *valueFieldString = NULL;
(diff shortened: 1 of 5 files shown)


andrewcraik:
@cathyzhyi looks like you were affected by some recent build issues - could you rebase and push to restart the travis ci build?

runtime/compiler/trj9/optimizer/J9RecognizedCallTransformer.cpp
```diff
@@ -60,6 +151,14 @@ bool J9::RecognizedCallTransformer::isInlineable(TR::TreeTop* treetop)
...
       case TR::java_lang_Math_min_L:
          return TR::Compiler->target.cpu.isX86() && !comp()->getOption(TR_DisableMaxMinOptimization);
+
+      case TR::java_lang_String_toLowerHWOptimized:
+      case TR::java_lang_String_toLowerHWOptimizedDecompressed:
+      case TR::java_lang_String_toLowerHWOptimizedCompressed:
+      case TR::java_lang_String_toUpperHWOptimized:
+      case TR::java_lang_String_toUpperHWOptimizedDecompressed:
+      case TR::java_lang_String_toUpperHWOptimizedCompressed:
+         return comp()->cg()->getSupportsInlineStringCaseConversion() && TR::Compiler->target.cpu.isX86();
```
fjeremic: Why do we need a `isX86()` check in a platform agnostic optimization? Shouldn't `getSupportsInlineStringCaseConversion()` cover us already as all code generators should be setting this flag if they want this inrinsic inlining?

fjeremic: I suppose Z would not currently support this sort of transformation via Recognized Call Transformer as we pattern match the call at direct evaluator time. Ok we can leave the x86 check in place for now. I'll open up a task on my end to start using this new infrastructure and get rid of the platfrom specific `isX86()` check.

runtime/compiler/trj9/optimizer/J9RecognizedCallTransformer.cpp
```diff
@@ -42,6 +42,97 @@
...
+   {
+   TR::Node *srcStringObject = node->getFirstChild();
+   TR::Node *dstStringObject = node->getSecondChild();
+
+   /* Offset of java/lang/String.value */
+   uint32_t stringValueFieldOffset = 0;
+   char *valueFieldString = NULL;
+   TR_OpaqueClassBlock *stringClass = comp()->fej9()->getClassFromSignature("Ljava/lang/String;", strlen("Ljava/lang/String;"), comp()->getCurrentMethod(), true);
+
+   if (comp()->fej9()->getInstanceFieldOffset(stringClass, "value", "[B") != ~0)
```
andrewcraik: wouldn't != -1 be more clear

cathyzhyi: yea, but the return type of getInstanceFieldOffset is unsigned int so we have to use ~0

cathyzhyi:
this pull request is dependent on https://github.com/eclipse/omr/pull/2358

cathyzhyi:
@andrewcraik @0dvictor Could you guys help to review this change? thanks!

0dvictor:
LGTM.

andrewcraik:
@cathyzhyi can you comment on the performance testing undertaken?

andrewcraik:
Jenkins test sanity

cathyzhyi:
The performance for micro benchmark is improved by 100%~300% depending on the string length. Tested on major benchmarks and shows no regression.

andrewcraik:
Jenkins test sanity

andrewcraik:
Jenkins test sanity zlinux jdk10

cathyzhyi:
@andrewcraik Hey Andrew, Seems there is a failure unrelated to my change. Is there a way to restart that test?

andrewcraik:
Jenkins test sanity zlinux jdk10

Issue #1377, opened 2018-03-07
FAILED: testAllocateMemoryInCategory value for Unknown did not change


pdbain-ibm:
Regression in this build:https://ci.eclipse.org/openj9/view/Test/job/Test-Extended-JDK8-linux_390-64_cmprssptrs/91/ 
Failed Tests: 2 (+2)
j9vm.test.memory.MemoryAllocator.testAllocateMemoryInAnUnknownCategory
j9vm.test.memory.MemoryAllocator.testAllocateMemoryInCategory

Exception java.lang.AssertionError

Message: value for Unknown did not change from 6,528 bytes / 32 allocations

Stacktrace:

at org.testng.AssertJUnit.fail(AssertJUnit.java:59)
at org.testng.AssertJUnit.assertTrue(AssertJUnit.java:24)
at org.testng.AssertJUnit.assertFalse(AssertJUnit.java:41)
at j9vm.test.memory.MemoryAllocator.verifyCategoryValueChanged(MemoryAllocator.java:174)
at j9vm.test.memory.MemoryAllocator.testAllocateMemoryInCategory(MemoryAllocator.java:130)

No obvious changes in OpenJ9.

llxia:
#1365 is not fix for this issue. #1365 is fix for #1355 to fix LIBPATH on AIX.

The above failure is not test issue and it was introduced on Mar 6th. The test failed on all platforms.
https://ci.eclipse.org/openj9/view/Test/job/Test-Extended-JDK8-linux_390-64_cmprssptrs/91/console
https://ci.eclipse.org/openj9/view/Test/job/Test-Extended-JDK8-linux_ppc-64_cmprssptrs_le/90/console
https://ci.eclipse.org/openj9/view/Test/job/Test-Extended-JDK9-linux_390-64_cmprssptrs/88/console
https://ci.eclipse.org/openj9/view/Test/job/Test-Extended-JDK9-linux_ppc-64_cmprssptrs_le/84/console

pdbain-ibm:
Will investigate.

pdbain-ibm:
The 1000 new allocations are being ascribed to trace:
Old javacore:
2MEMUSER       +--VM: 2,365,679,968 bytes / 2546 allocations
2MEMUSER       |  | 
3MEMUSER       |  +--Classes: 9,832,096 bytes / 336 allocations
2MEMUSER       |  | 
[...]

PR #1378, opened 2018-03-07
Fix sidecarExit

babsingh:
`java.lang.Shutdown.exit` takes an `int` as the input argument.
`runStaticMethod` expects ints to be stored in the low-memory half of the
stack slot on 64-bit machines. A negative consequence of not following
this convention: input argument is always zero inside `Shutdown.exit` on
big-endian platforms.

### DIFF
--- a/runtime/vm/jniinv.c
+++ b/runtime/vm/jniinv.c
@@ -1257,14 +1257,20 @@ jint JNICALL ResetJavaVM(JavaVM *vm)
 /* run the shutdown method in java.lang.Shutdown */
 void 
 sidecarExit(J9VMThread* shutdownThread) {
-	UDATA args[] = {130};	/* The exit code */
-	J9NameAndSignature nas;
+	/* static void java.lang.Shutdown.exit(int status)
+	 *
+	 * The exit code is 130. Store ints in the low-memory half
+	 * of the stack slot on 64-bit platforms. Otherwise, input
+	 * argument will always be zero on big-endian platforms.
+	 */
+	I_32 args[] = {130};
+	J9NameAndSignature nas = {0};
 
-	nas.name = (J9UTF8*)&j9_exit;
-	nas.signature = (J9UTF8*)&j9_int_void;
+	nas.name = (J9UTF8 *)&j9_exit;
+	nas.signature = (J9UTF8 *)&j9_int_void;
 
 	enterVMFromJNI(shutdownThread);
-	runStaticMethod(shutdownThread, (U_8*)"java/lang/Shutdown", &nas, 1, args);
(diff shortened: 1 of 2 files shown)


runtime/vm/jniinv.c
```diff
@@ -1257,8 +1257,16 @@ jint JNICALL ResetJavaVM(JavaVM *vm)
...
 void 
 sidecarExit(J9VMThread* shutdownThread) {
-	UDATA args[] = {130};	/* The exit code */
-	J9NameAndSignature nas;
+	UDATA args[] = {0};
+	J9NameAndSignature nas = {0};
+
+	/* static void java.lang.Shutdown.exit(int status)
+	 *
+	 * The exit code is 130. Store ints in the low-memory half
```
pshipton: The comment doesn't match the code. If we have 2 slots X & Y (in that order), the original code stored into X for little endian, and Y for big endian. The new code always stores into X. Also instead of AIX I expect you mean big endian platforms. Perhaps args should be declared as I_32 and could continue to be initialized when declared, then cast to UDATA for the call.

babsingh: fixed.

runtime/vm/jniinv.c
```diff
@@ -1257,14 +1257,20 @@ jint JNICALL ResetJavaVM(JavaVM *vm)
 /* run the shutdown method in java.lang.Shutdown */
 void 
 sidecarExit(J9VMThread* shutdownThread) {
-	UDATA args[] = {130};	/* The exit code */
-	J9NameAndSignature nas;
+	/* static void java.lang.Shutdown.exit(int status)
+	 *
+	 * The exit code is 130. Store ints in the low-memory half
```
pshipton: I still don't think the comment matches the code. The original code was storing into the low-memory half of a 64-bit value. The new code is always storing into the first 32-bit, which would be the upper-memory half of a 64-bit value on big endian.

pshipton: Actually its my interpretation of "low-memory" that is the problem. Technically the comment is correct.

pshipton:
jenkins test all

babsingh:
For the four failing jobs, the following tests are failing:
```
    j9vm.test.memory.MemoryAllocator.testAllocateMemoryInCategory
    j9vm.test.memory.MemoryAllocator.testAllocateMemoryInAnUnknownCategory
```
All failures have a common error message:
```
value for Unknown did not change from 6,528 bytes / 32 allocations
```
I think these failures are unrelated to my changes.

pshipton:
The test failures are a known issue #1377

PR #1380, opened 2018-03-07
Add Jenkins files for Windows x86 build and test pipelines

vsebe:
Add build and tests pipelines for JDK 8 and 9.
[ci skip]

By default the stage runs in the build workspace. Is not necessary to specify it.
On Windows, tar fails due to the fact that `WORKSPACE` is not converted to cygpath (it's a windows path C:\Users\...).

### DIFF
--- a/buildenv/jenkins/common/build
+++ b/buildenv/jenkins/common/build
@@ -149,13 +149,13 @@ def checkout_pullrequest() {
 def build() {
     stage('Compile') {
         timestamps {
-            sh "bash ./configure --with-freemarker-jar=$FREEMARKER --with-boot-jdk=$BOOT_JDK $EXTRA_CONFIGURE_OPTIONS"
+            sh "bash ./configure --with-freemarker-jar='$FREEMARKER' --with-boot-jdk='$BOOT_JDK' $EXTRA_CONFIGURE_OPTIONS"
             sh "make all"
         }
     }
     stage('Java Version') {
         timestamps {
-            sh "build/$RELEASE/images/${(SDK_VERSION == '8') ? 'j2sdk-image' : 'jdk'}/bin/java -version"
+            sh "build/$RELEASE/images/$JDK_FOLDER/bin/java -version"
         }
     }
 }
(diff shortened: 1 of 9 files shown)


buildenv/jenkins/common/build
```diff
@@ -149,13 +149,13 @@ def checkout_pullrequest() {
 def build() {
     stage('Compile') {
         timestamps {
-            sh "bash ./configure --with-freemarker-jar=$FREEMARKER --with-boot-jdk=$BOOT_JDK $EXTRA_CONFIGURE_OPTIONS"
+            sh "bash ./configure --with-freemarker-jar=$FREEMARKER --with-boot-jdk='$BOOT_JDK' $EXTRA_CONFIGURE_OPTIONS"
```
jdekonin: Why quotes around one and not required for any of the others?

vsebe: The boot-jdk might contain spaces if installed in the default location "C:/Program Files".
For consistency, I added quotes to freemarker too.

vsebe:
@jdekonin ready for review

vsebe:
@pshipton please review

buildenv/jenkins/common/build
```diff
@@ -167,10 +167,10 @@ def archive() {
             // https://github.com/eclipse/openj9/issues/1114
             // ghprbPullId is the PullRequest ID which only shows up in Pull Requests
             if (!params.ghprbPullId) {
-                sh "tar -zcvf ${WORKSPACE}/${TEST_PREFIX}`git -C openj9 rev-parse --short HEAD`${TEST_SUFFIX} openj9/test/"
```
pshipton: Please explain why `${WORKSPACE}` is removed. common/test still uses it, does it need to be removed there as well?

vsebe: By default the stage runs in the build workspace. Is not necessary to specify it.
On Windows itar fails due to the fact that `<WORKSPACE>` is not converted to cygpath it's a windows path C:\Users\...
In common/test is not necessary to remove it as java on cygwin uses windows paths not cygpaths.

PR #1383, opened 2018-03-08
Refactor access check for processing methods

taliamccormick:
When looking up a method, various access checks occur. The existing access checks on protected virtual methods produce a visibility non module access error. When producing an error message, one case assumes that the visibility error is a non module access error. Due to the current code path, this has been successful. This commit explicitly passes the type of the visibility error from the check to the creation of the error message.

In particular, it is necessary with the addition of nestmates access checks - in which case a nestmate related access failure can occur. This PR split off from a nestmates loading PR, here: https://github.com/eclipse/openj9/pull/1224

### DIFF
--- a/runtime/vm/lookupmethod.c
+++ b/runtime/vm/lookupmethod.c
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 1991, 2017 IBM Corp. and others
+ * Copyright (c) 1991, 2018 IBM Corp. and others
  *
  * This program and the accompanying materials are made available under
  * the terms of the Eclipse Public License 2.0 which accompanies this
@@ -193,6 +193,7 @@ processMethod(J9VMThread * currentThread, UDATA lookupOptions, J9Method * method
 					} else {
 							*exception = J9VMCONSTANTPOOL_JAVALANGILLEGALACCESSERROR;
 							*exceptionClass = methodClass;
+							*errorType = J9_VISIBILITY_NON_MODULE_ACCESS_ERROR;
 							return NULL;
 					}
 				}
@@ -217,6 +218,7 @@ processMethod(J9VMThread * currentThread, UDATA lookupOptions, J9Method * method
 	{
 		*exception = J9VMCONSTANTPOOL_JAVALANGINCOMPATIBLECLASSCHANGEERROR;
 		*exceptionClass = methodClass;
+		*errorType = J9_VISIBILITY_NON_MODULE_ACCESS_ERROR;
 		return NULL;
 	}
 
@@ -244,6 +246,7 @@ processMethod(J9VMThread * currentThread, UDATA lookupOptions, J9Method * method
 				if (j9bcv_checkClassLoadingConstraintsForSignature(currentThread, cl1, cl2, lookupSig, methodSig) != 0) {
 					*exception = J9VMCONSTANTPOOL_JAVALANGLINKAGEERROR; /* was VerifyError; but Sun throws Linkage */
 					*exceptionClass = methodClass;
+					*errorType = J9_VISIBILITY_NON_MODULE_ACCESS_ERROR;
 					Trc_VM_processMethod_ClassLoaderConstraintFailure(currentThread, method, cl1, cl2);
 					return NULL;
 				}
@@ -424,6 +427,7 @@ javaResolveInterfaceMethods(J9VMThread *currentThread, J9Class *targetClass, J9R
 						J9Method **newArray = j9mem_allocate_memory(sizeof(J9Method*) * newArrayLength, OMRMEM_CATEGORY_VM);
 						if (NULL == newArray) {
 							data->exception = J9VMCONSTANTPOOL_JAVALANGOUTOFMEMORYERROR;
+							data->errorType = J9_VISIBILITY_NON_MODULE_ACCESS_ERROR;
 							return NULL;
 						}
 						/* Only memset the upper (new) portion of the array */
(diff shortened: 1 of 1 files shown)


taliamccormick:
@DanHeidinga @tajila

runtime/vm/lookupmethod.c
```diff
@@ -193,6 +193,7 @@ processMethod(J9VMThread * currentThread, UDATA lookupOptions, J9Method * method
 					} else {
 							*exception = J9VMCONSTANTPOOL_JAVALANGILLEGALACCESSERROR;
 							*exceptionClass = methodClass;
+							*errorType = J9_VISIBILITY_NON_MODULE_ACCESS_ERROR;
```
DanHeidinga: Does line 219 need `errorType` set as well?

DanHeidinga: Also line ~248 as well?

DanHeidinga: The error handling at the end of `javaResolveInterfaceMethods` should also set the errorType in `J9InterfaceResolveData` struct

DanHeidinga: in `javaLookupMethodImpl`, near the comment `/* Reset exception state to initial values and move up the hierarchy */`, does the errorType need to be reset as well?

taliamccormick: I think that the value of errorType is only used here for java.lang.IllegalAccessError and java.lang.IllegalAccessException, so it does not need to be set for java.lang.IncompatibleClassChangeError/java.lang.LinkageError/etc.

---

Line 219:
* `processMethod` checks if the method that is found matches the method that we want (looking for static & find static; or looking for virtual and find non-static)
* If it does not match, we set a java.lang.IncompatibleClassChangeError and the exception class
* If `processMethod` is called by `javaResolveInterfaceMethods` then we do not check for java.lang.IncompatibleClassChangeError within our function, but pass the exception value up to the caller of `javaResolveInterfaceMethods`. The only caller of `javaResolveInterfaceMethods` is `javaLookupMethodImpl`, which already handles it correctly.
* If `processMethod` is called by javaLookupMethodImpl (lookupmethod.c:691), the errorType value is set & passed along correctly

---

Line 248:
* `processMethod` checks the class loading constraints for the method signature and sets a java.lang.LinkageError if the constraints are not met
* We eventually check if the exception is java.lang.LinkageError in `javaLookupMethodImpl` and set a class loading constraint error - at which point, we do not use the value of errorType

---

Error handling at end of `javaResolveInterfaceMethods (lines 490 - 550):
* I have now added a line to reset to `data->errorType = J9_VISIBILITY_ALLOWED` but am not sure it is correct.
* The value of errorType seems to only ever be used with visibility errors (eg. `J9_VISIBILITY_NON_MODULE_ACCESS_ERROR`, `J9_VISIBILITY_MODULE_READ_ACCESS_ERROR`) so we would want the default value of `J9_VISIBILITY_ALLOWED`
* Whenever a visibility error is set in the `exception` field, a new value is set for the `errorType`
* Actual value of `J9_VISIBILITY_ALLOWED` is 1, rather than 0, which may not be the default value if the `errorType` field is ever used to store non-visibility error code

---

`javaLookupMethodImpl` near `/* Reset exception state to initial values and move up the hierarchy */` (line 698):
* Same as prior case - I have now added the line `data->errorType = J9_VISIBILITY_NON_MODULE_ACCESS_ERROR` since that is the initial default value within `javaLookupMethodImpl`

DanHeidinga: To avoid everyone reading this code from having to do this analysis, we should either add a comment to either of these locations to indicate why we don't need to set `errorType` or set it to a reasonable default error condition.

Setting it to `J9_VISIBILITY_ALLOWED` when also setting an exception will lead to bugs in the future when someone checks the `errorType` first.  Both fields should have the same meaning - failure or success.  Mixing the two will lead to problems.

taliamccormick: The default value of `errorType` generally seems to be `J9_VISIBILITY_NON_MODULE_ACCESS_ERROR`, rather than `J9_VISIBILITY_ALLOWED` then. Could I cross-check an example to make sure we're on the same page?

--

eg, in `javaResolveInterfaceMethods`, we check if there is a method conflict causing an AbstractMethodError. 

Right now, we set:
```
data->exception = J9VMCONSTANTPOOL_JAVALANGINCOMPATIBLECLASSCHANGEERROR;
```
(javaResolveInterfaceMethods, vm.lookupmethod.c:509)
but ignore the value of errorType. Via the code path it currently follows, the value of `errorType` should be `J9_VISIBILITY_NON_MODULE_ACCESS_ERROR` but the value of `errorType` is not used.

However, changes added by this PR allow the value of `errorType` to be values which are not `J9_VISIBILITY_NON_MODULE_ACCESS_ERROR`. In this particular case, we would not want to produce a visibility error/exception and thus do not need a specific visibility error code. Since this is an exception-case (AbstractMethodError caused by method conflict), the exception code and errorType code should both indicate an error/exception state. Thus, picking a non-`J9_VISIBILITY_ALLOWED` for `errorType` should be sufficient. Given that the earlier default is `J9_VISIBILITY_NON_MODULE_ACCESS_ERROR`, we should therefore set the `errorType` value to be `J9_VISIBILITY_NON_MODULE_ACCESS_ERROR`.

--

Am I understanding this correctly?

DanHeidinga: Sounds right.  

By ensuring both `exception` and `errorType` are always set, we make it blatantly obvious to future readers of the code what they need to do when working with this code.  And ensures that code (bogusly) checking if `errorType == J9_VISIBILITY_ALLOWED` will never execute.

taliamccormick: That makes sense to me; thank you. Updated now.

DanHeidinga:
Jenkins test sanity plinux

taliamccormick:
I squashed, rebased, and repushed

taliamccormick:
I squashed, rebased, and repushed the commits

DanHeidinga:
Jenkins test sanity plinux

DanHeidinga:

Issue #1395, opened 2018-03-08
Jenkins: parse JOB_NAME to set SPEC and SDK_VERSION variables


AdamBrousseau:
Currently, we have Pipeline files which are essentially copy/pasted several times per SPEC and SDK_VERSION. It has been proposed that we just have a "generic" file that takes parameters for these 2 values. The issue with using parameters is they can be changed "on-the-fly" and since we have different jobs for each SPEC and SDK_VERSION, we do not want these to change from build to build.

Since we have a standard naming convention for our job names, another idea would be to write a function that sets the SPEC and SDK_VERSION based on the Jenkins JOB_NAME. I have prototyped this feature and it appears to work in the pipeline build. I wanted to open an issue to discuss this approach more. What would the limitations of this be? I can see difficulty arising when trying to debug a build. For example, I currently have a debug copy of a build and the SPEC and SDK in the job name doesn't always match the build I'm testing (I often change the JenkinsFile that is loaded). The other thing is we would need to maintain 2 more variables in the variables file which would be arrays of SPECS and SDK_VERSIONS. We may be able reuse these new variable arrays in the Build-Test-All pipeline?

Note: There is also hardcoded TEST_TARGET in the PR pipelines but I think we should ignore that for now because the test story may be changing.

AdamBrousseau:
Prototype: https://github.com/eclipse/openj9/pull/1396

AdamBrousseau:
fyi @jdekonin @vsebe

AdamBrousseau:
Going to close this for now as I do not plan to use it. Can be revisited later.

PR #1396, opened 2018-03-08
WIP: [skip ci] parse build name to set spec and sdk_version

AdamBrousseau:
Issue #1395

### DIFF
--- a/buildenv/jenkins/variables/defaults.yml
+++ b/buildenv/jenkins/variables/defaults.yml
@@ -29,6 +29,8 @@ openjdk_branch: 'openj9'
 #========================================#
 # Miscellaneous settings
 #========================================#
+specs: 'linux_ppc-64_cmprssptrs_le,linux_390-64_cmprssptrs,aix_ppc-64_cmprssptrs'
+sdks: '8,9'
 jdk_image_dir:
   8: 'j2sdk-image'
   9: 'jdk'
--- a/buildenv/jenkins/common/variables-functions
+++ b/buildenv/jenkins/common/variables-functions
@@ -232,6 +232,34 @@ def set_slack_channel() {
     SLACK_CHANNEL = VARIABLES.slack_channel
 }
 
+def parse_job_name() {
+    SPEC = ''
+    SDK_VERSION = ''
+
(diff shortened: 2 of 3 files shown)


buildenv/jenkins/variables/defaults.yml
```diff
@@ -29,6 +29,8 @@ openjdk_branch: 'openj9'
 #========================================#
 # Miscellaneous settings
 #========================================#
+specs: 'linux_ppc-64_cmprssptrs_le,linux_390-64_cmprssptrs,aix_ppc-64_cmprssptrs'
```
jdekonin: I like the idea, but I wonder if might we ever hit a situation where a spec won't be performed for a release?

buildenv/jenkins/variables/defaults.yml
```diff
@@ -29,6 +29,8 @@ openjdk_branch: 'openj9'
...
+sdks: '8,9'
 jdk_image_dir:
```
jdekonin: I like the idea, but I wonder if might we ever hit a situation where a spec won't be performed for a release?

AdamBrousseau: If I understand correctly, an imaginary example: no aix for 8 combination. I think that's fine, we wouldn't have a Jenkins job to do such a thing. We still would maintain the amount of Jenkins jobs we have now, but instead of having a file per job we may only have 3/4 files to cover all the job types. I think what you are thinking of we would have to worry about in the Pipeline-Build-Test-All file.

DanHeidinga:
Does this affect build launch times?

AdamBrousseau:
>Does this affect build launch times?

No. The time to loop through (at most) all the specs and (at most) all the sdk_versions is negligible.

AdamBrousseau:
adam test sanity

AdamBrousseau:
adam test sanity

jdekonin:
adam test sanity

AdamBrousseau:
adam test sanity

AdamBrousseau:
Going to close this for now as I do not plan to use it. Can be revisited later.

PR #1397, opened 2018-03-08
Fixed Inline Array Allocation

R-Santhir:
When inlining array allocations, several checks are performed to make sure that
the allocation is valid and can be done inline. Some of the checks performed
try to determine if the size or number of elements of the array that is being
allocated fall under a max value. If they do not, the VM handles the allocation
out-of-line.

An issue with the aformentioned checks was that the size or number of elements,
were being stored in registers that could have most significant bits that were
contaminated by  prior instructions. Thus when the checks were performed,
although the size of the array or number of elements was within the range of an
inline array allocation, it would seem as if they were not and an out-of-line
allocation would occur, affecting the performance.

In order to fix this, the bits that could be contaminated are cleared as to not
affect any following calculations.

### DIFF
--- a/runtime/compiler/trj9/p/codegen/J9TreeEvaluator.cpp
+++ b/runtime/compiler/trj9/p/codegen/J9TreeEvaluator.cpp
@@ -5575,8 +5575,8 @@ static void genHeapAlloc(TR::Node *node, TR::Instruction *&iCursor, TR_OpaqueCla
                uintptrj_t mask;
                if (TR::Compiler->target.is64Bit())
                   {
-                  mask = CONSTANT64(0xFFFFFFFFFFFFFFFF) << (27 - leadingZeroes(maxSafeSize));
-                  iCursor = generateTrg1Src1Imm2Instruction(cg, TR::InstOpCode::rldicr_r, node, temp2Reg, enumReg, condReg, 0, mask, iCursor);
+                  mask = 0xFFFFFFFF << (27 - leadingZeroes(maxSafeSize));
+                  iCursor = generateTrg1Src1Imm2Instruction(cg, TR::InstOpCode::rlwinm_r, node, temp2Reg, enumReg, condReg, 0, mask, iCursor);
                   }
                else
                   {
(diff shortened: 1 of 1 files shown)


R-Santhir:
@gita-omr @zl-wang

runtime/compiler/trj9/p/codegen/J9TreeEvaluator.cpp
```diff
@@ -5575,8 +5575,8 @@ static void genHeapAlloc(TR::Node *node, TR::Instruction *&iCursor, TR_OpaqueCla
                uintptrj_t mask;
                if (TR::Compiler->target.is64Bit())
                   {
-                  mask = CONSTANT64(0xFFFFFFFFFFFFFFFF) << (27 - leadingZeroes(maxSafeSize));
-                  iCursor = generateTrg1Src1Imm2Instruction(cg, TR::InstOpCode::rldicr_r, node, temp2Reg, enumReg, condReg, 0, mask, iCursor);
+                  mask = 0xFFFFFFFF << (27 - leadingZeroes(maxSafeSize));
+                  iCursor = generateTrg1Src1Imm2Instruction(cg, TR::InstOpCode::rlwinm_r, node, temp2Reg, enumReg, condReg, 0, mask, iCursor);
```
zl-wang: embedding the comment:
Minor: rlwinm essentially consumes a 32bit mask. you have a 64-bit mask there in 64bit mode. fortunately for rlwinm, binary encoding takes only the lower portion. it just works out.

R-Santhir: All method definitions define the mask as a 64bit value:
```
omr/compiler/p/codegen/GenerateInstructions.hpp:

234 TR::Instruction *generateTrg1Src1Imm2Instruction(
  1                    TR::CodeGenerator      *cg,
  2                    TR::InstOpCode::Mnemonic   op,
  3                    TR::Node        *n,
  4                    TR::Register    *treg,
  5                    TR::Register    *s1reg,
  6                    int32_t         imm1,
  7                    int64_t         imm2,
  8                    TR::Instruction *preced = 0);
  9
 10 TR::Instruction *generateTrg1Src1Imm2Instruction(
 11                    TR::CodeGenerator      *cg,
 12                    TR::InstOpCode::Mnemonic   op,
 13                    TR::Node        *n,
 14                    TR::Register    *treg,
 15                    TR::Register    *s1reg,
 16                    TR::Register    *cr0reg,
 17                    int32_t         imm1,
 18                    int64_t         imm2,
 19                    TR::Instruction *preced = 0);
```

There are also other cases in the code base where rlwinm is used with a 64bit mask. How should I proceed?

runtime/compiler/trj9/p/codegen/J9TreeEvaluator.cpp
```diff
@@ -5735,36 +5735,24 @@ static void genHeapAlloc(TR::Node *node, TR::Instruction *&iCursor, TR_OpaqueCla
...
-                  // of 4 (for the loop of stwu's)
-                  iCursor = generateTrg1Src1ImmInstruction(cg, TR::InstOpCode::addi, node, dataSizeReg, enumReg, 3, iCursor);
-                  if (TR::Compiler->target.is64Bit())
+                  else  //elementSize == 1
+                     iCursor = generateTrg1Src1ImmInstruction(cg, TR::InstOpCode::addi, node, dataSizeReg, enumReg, 3, iCursor);
+
+                  if (TR::Compiler->target.is64Bit() && elementSize != 1)
                      iCursor = generateTrg1Src1Imm2Instruction(cg, TR::InstOpCode::rldicr, node, dataSizeReg, dataSizeReg, 0, int64_t(-4), iCursor);
                   else
                      iCursor = generateTrg1Src1Imm2Instruction(cg, TR::InstOpCode::rlwinm, node, dataSizeReg, dataSizeReg, 0, -4, iCursor);
```
zl-wang: This existing (2nd) rounding of size looked unnecessary to me, although I knew fixing this is not part of the goal/mandate of this PR.  It has no bad side-effect when round is bigger than 4, and that is true anyway.

runtime/compiler/trj9/p/codegen/J9TreeEvaluator.cpp
zl-wang: it is acceptable as it is, as i indicated above.

zl-wang: initially i had a concern about what really ended up in "mask" variable by the expression: (0xFFFFFFFF << shift-amount).

Issue #1367, opened 2018-03-06
Upgrade TestNG to latest release (v6.14)


llxia:
We are using TestNG 6.10, which has a known issue - TestNG exits with a zero when there are configuration failures. When this happens, TKG cannot catch the failure. Therefore, the TAP results that is generated will not have the failure.

Link to TestNG issue:
https://github.com/cbeust/testng/issues/1517 
https://groups.google.com/forum/#!msg/testng-users/WlX5H5B8fwo/m_GR0svPAQAJ

This issue is fixed in TestNG 6.13. We should upgrade TestNG to latest release (v6.14).
http://central.maven.org/maven2/org/testng/testng/

llxia:
@TianyuZuo , could you take care of this? Thanks.

TianyuZuo:
will update

smlambert:
This upgrade will address/fix #1112

PR #1398, opened 2018-03-08
update TestNG to 6.14.2 version to fix false positive test result

TianyuZuo:
* TestNG before 6.14 treats configuration failure as pass
* this has been fixed since TestNG 6.14
* update TestNG version to 6.14.2 to fix this bug in product build
* separate JLM test targets to avoid OOM failure
* re-enable tests which are fixed and passing
* update exlucde file which failed on TestNG 6.14.2
* exclude reattachAfterExit from win since it is not supported on win

fixed:#1367
fixed:#1112
Issue:#1454

Signed-off-by: Tianyu Zuo <tianyu@ca.ibm.com>

### DIFF
--- a/test/JLM_Tests/playlist.xml
+++ b/test/JLM_Tests/playlist.xml
@@ -34,7 +34,15 @@
 	-XX:SharedCacheHardLimit=16m -Xscmx1m -Xshareclasses:name=testJLM,reset \
 	-cp $(Q)$(RESOURCES_DIR)$(P)$(TESTNG)$(P)$(TEST_RESROOT)$(D)jlm_tests.jar$(Q) \
 	org.testng.TestNG -d $(REPORTDIR) $(Q)$(TEST_RESROOT)$(D)testng.xml$(Q) \
-	-testnames JLM_Tests_interface \
+	-testnames TestCompilationMXBean,\
+	TestGarbageCollectorMXBean,\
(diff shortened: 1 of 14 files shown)


llxia:
Please also update TestNG jar version in readme. Thanks.
Please skip ci in your next commit msg, as this change cannot be verified by SDK compilation.

TianyuZuo:
Hi @llxia  @smlambert , this PR is ready for review

llxia:
Could we add description for why we have to separate JLM_Tests_interface, remove gpu exclude, and exclude on win? Thanks.

TianyuZuo:
they were in the commit message, updated the description of this PR.

smlambert:
Jenkins test sanity all

smlambert:
Thanks @llxia - I forgot that this is how the getDependency currently is called at OpenJ9 (could likely rethink this approach, and consider same approach as for freemarker.jar (try to fetch it, and only if fails, fall back to pre-fetched version).

Will merge this change (which has been verified in internal personal builds) and launch a getdependency job.

TianyuZuo:
Thanks for launching the getdependency job Shelley

smlambert:
This was a big PR to pull together (given all the internal dependencies also), thanks for sticking with it @TianyuZuo !

PR #1361, opened 2018-03-06
Update sdk and freemarker locations for zlinux machines

jdekonin:
* use downloadable adoptopenjdk sdks as bootjdk
* relocate freemarker.jar to $HOME
* [skip ci]

Signed-off-by: Joe deKoning <joe_dekoning@ca.ibm.com>

PR #1399, opened 2018-03-08
Enable use of ddrgen from OMR

keithc-ca:
* add new UMA flag 'opt_useOmrDdr'
* add rule to make *.i from *.m4 for ddrgen
* adjust modules and configure OMR accordingly
* omit work based on legacy ddr generation tools
* run omr/getmacros to discover constants
* run omr/ddrgen to produce blob and superset

Depends on https://github.com/eclipse/omr/pull/2364.

### DIFF
--- a/runtime/ddr/module.xml
+++ b/runtime/ddr/module.xml
@@ -1,37 +1,38 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <!--
-   Copyright (c) 2010, 2017 IBM Corp. and others
+Copyright (c) 2010, 2018 IBM Corp. and others
 
-   This program and the accompanying materials are made available under
-   the terms of the Eclipse Public License 2.0 which accompanies this
-   distribution and is available at https://www.eclipse.org/legal/epl-2.0/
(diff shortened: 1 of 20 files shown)


runtime/ddr/module.xml
```diff
@@ -1,351 +1,429 @@
...
+			<dependency name="j9jvmti"/>
+			<dependency name="j9prt"/>
+			<dependency name="j9shr"/>
+			<dependency name="j9thr"/>
+			<dependency name="j9trc"/>
+			<dependency name="j9vm"/>
+			<!--
+			We actually only need one of the following, but
+			the version we're building doesn't appear to be
+			availabue to UMA, so we'll require them all.
```
pshipton: typo availabue

pshipton:
@dmitripivkine fyi

keithc-ca:
Fixed type and squashed.

pshipton:
jenkins compile

pshipton:
The JDK9 failures are related to finding freemarker (fixed by #1361), and the ci build passed. The JDK8 compile on aix isn't finished yet, but the OpenJ9 part is. I believe this is good to merge.

Issue #1341, opened 2018-03-05
Make freemarker.jar downloads robust to sourceforge outages


DanHeidinga:
While we cache freemarker.jar on the CI machines, we should also update the makefiles to pull from https://ci.adoptopenjdk.net/userContent/freemarker-2.3.8.tar.gz if sourceforge returns an error.

This will make builds more robust and prevent outages like the recent ones due to sourceforge being down.

DanHeidinga:
See https://github.com/AdoptOpenJDK/openjdk-build/blob/ee35cfa2535d06a3b7250beada4678744db11b37/sbin/common-functions.sh#L65

Adopt has already handled this in their scripts and we've handled it by ensuring freemarker is available in the machine config.

jdekonin:
OpenJ9s travis build pulls directly from sourceforge still.  Is this worth the simple change of pulling from https://ci.eclipse.org/openj9/userContent/freemarker.jar?

PR #1400, opened 2018-03-08
Use a stored copy of freemarker.jar for the travis build

jdekonin:
* reduces travis failure due to external dependency
* #1341

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -64,8 +64,10 @@ before_script:
   # - ccache -s -z
   # Exit immediately if any unexpected error occurs.
   - set -e
-  - wget https://sourceforge.net/projects/freemarker/files/freemarker/2.3.8/freemarker-2.3.8.tar.gz/download -O freemarker.tgz
-  - tar -xzf freemarker.tgz freemarker-2.3.8/lib/freemarker.jar --strip=2
+  - if [ ! `wget https://ci.eclipse.org/openj9/userContent/freemarker-2.3.8.jar -O freemarker.jar` ]; then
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -63,8 +63,7 @@ before_script:
   # - ccache -s -z
   # Exit immediately if any unexpected error occurs.
   - set -e
-  - wget https://sourceforge.net/projects/freemarker/files/freemarker/2.3.8/freemarker-2.3.8.tar.gz/download -O freemarker.tgz
-  - tar -xzf freemarker.tgz freemarker-2.3.8/lib/freemarker.jar --strip=2
+  - wget https://ci.eclipse.org/openj9/userContent/freemarker.jar
```
fjeremic: Any reason to remove the version information from the file name?

jdekonin: Not really, and I was just thinking of putting that back in to make it obvious.  You're comment seals it.

DanHeidinga:
@jdekonin Can you have this pull from the Eclipse CI and fallback to sourceforge if the CI version isn't available?

jdekonin:
This commit verifies failover if freemarker.jar is not available from the eclipse userContent location; note it passes configure, the fail is in compile because my fork is apparently not up2date.
https://travis-ci.org/eclipse/openj9/builds/351489405
Note I have freemarker-2.3.9.jar which doesn't exist in userContent.
Next commit will revert the version number only.  Sorry for all the 'push' spam' to test.

DanHeidinga:
Tested manually by using the wrong link for the first wget, the fallback works fine.

Issue #1401, opened 2018-03-09
OpenJ9 throws ClassFormatError for old class files


knn-k:
The latest nightly build (March 8) fails to handle old Java class files, throwing ClassFormatError as shown below:
```
$ jdk-9+181/bin/java -version
openjdk version "9-internal"
OpenJDK Runtime Environment (build 9-internal+0-adhoc.jenkins.openjdk)
Eclipse OpenJ9 VM (build master-969cd999, JRE 9 Linux amd64-64 Compressed References 20180308_132 (JIT enabled, AOT enabled)
OpenJ9   - 969cd999
OMR      - d23f8d33
JCL      - 76f76bdb3 based on jdk-9+181)
$ jdk-9+181/bin/java -cp ~/tmp/embed.zip CaffeineMarkEmbeddedApp
Exception in thread "main" java.lang.ClassFormatError: JVMCFRE077 interfaces must be abstract; class=BenchmarkAtom, offset=253
        at java.lang.ClassLoader.defineClass(java.base@9-internal/ClassLoader.java:445)
        at java.security.SecureClassLoader.defineClass(java.base@9-internal/SecureClassLoader.java:174)
        at jdk.internal.loader.BuiltinClassLoader.defineClass(java.base@9-internal/BuiltinClassLoader.java:1098)
        at jdk.internal.loader.BuiltinClassLoader.findClassOnClassPathOrNull(java.base@9-internal/BuiltinClassLoader.java:886)
        at jdk.internal.loader.BuiltinClassLoader.loadClassOrNull(java.base@9-internal/BuiltinClassLoader.java:793)
        at jdk.internal.loader.BuiltinClassLoader.loadClass(java.base@9-internal/BuiltinClassLoader.java:751)
        at jdk.internal.loader.ClassLoaders$AppClassLoader.loadClass(java.base@9-internal/ClassLoaders.java:185)
        at java.lang.ClassLoader.loadClass(java.base@9-internal/ClassLoader.java:1026)
        at CaffeineMarkEmbeddedApp.main(CaffeineMarkEmbeddedApp.java)
```
I used the ancient CaffeineMark from 1997.

March 6 nightly build does not have this problem.

knn-k:
I guess it is caused by the change in https://github.com/eclipse/openj9/pull/1359

DanHeidinga:
@knn-k Is there a newer version of CaffeineMark available?

knn-k:
[...]

PR #1359, opened 2018-03-06
Interfaces must have acc_interface & acc_abstract set

DanHeidinga:
OpenJ9 used to fix the missing abstract bit unless -Xfuture was specified.
Stop fixing it and just mark it as a ClassFormatError.

Fixes: #1352

Signed-off-by: Dan Heidinga <daniel_heidinga@ca.ibm.com>

PR #1402, opened 2018-03-09
Automatically fix interface modifiers for pre-java6 classfiles

DanHeidinga:
Javac used to generate interfaces that didn't have the acc_abstract bit
set.  We used to only do this check under -Xfuture but that led to
behaviour differences.  Now, just fix pre-java6 interfaces and error
on newer classfiles.

Fixes: #1401


pshipton:
The ci testing passed. I won't bother with PR testing as it obviously doesn't contain older class files, otherwise this problem would have been noticed when #1359 was delivered.

PR #1411, opened 2018-03-09
Update nestmates fields during class redefinition

taliamccormick:
With the introduction of JEP 181, classes belong to a 'nest' - a group of classes that share access between private fields. In order to define a class's nest, class files gain two nestmates attributes: 'NestHost' and 'NestMembers'. A class determines its nest by providing the name of its 'NestHost'. In return, the 'NestHost' confirms a list of 'NestMembers'. During access checking, the NestHost field of two classes is compared in order to determine whether they have the same nest host, and thus the same nest.

During class redefinition, the nest host must find any nest members that have already claimed it and update them. Any nest member class which points to a nest host will leave its NestHost field blank during class loading and set it if/when necessary in a visibility check - these classes therefore do not need to update themselves during redefinition.

Related issue here: https://github.com/eclipse/openj9/issues/1165

### DIFF
--- a/runtime/jvmti/jvmtiClass.c
+++ b/runtime/jvmti/jvmtiClass.c
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 1991, 2017 IBM Corp. and others
+ * Copyright (c) 1991, 2018 IBM Corp. and others
  *
  * This program and the accompanying materials are made available under
  * the terms of the Eclipse Public License 2.0 which accompanies this
@@ -1102,6 +1102,11 @@ redefineClassesCommon(jvmtiEnv* env,
 
 				/* Indicate that a redefine has occurred */
 				vm->hotSwapCount += 1;
+
+#if defined(J9VM_OPT_VALHALLA_NESTMATES)
+				/* Update nests with redefined nest tops */
+				fixNestMembers(currentThread, classPairs);
+#endif /* defined(J9VM_OPT_VALHALLA_NESTMATES) */
 
 #ifdef J9VM_INTERP_NATIVE_SUPPORT
 				/* Notify the JIT about redefined classes */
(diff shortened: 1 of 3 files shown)


runtime/jvmti/jvmtiClass.c
```diff
@@ -1107,6 +1107,11 @@ redefineClassesCommon(jvmtiEnv* env,
 				/* Notify the JIT about redefined classes */
 				jitClassRedefineEvent(currentThread, &jitEventData, extensionsEnabled);
 #endif
+
+#if defined(J9VM_OPT_VALHALLA_NESTMATES)
+				/* Update nests with redefined nest tops */
+				fixNestMembers(currentThread, classPairs);
+#endif /* defined(J9VM_OPT_VALHALLA_NESTMATES) */
```
gacholio: This code should be before notifying the JIT.

runtime/util/hshelp.c
```diff
@@ -2149,6 +2149,54 @@ flushClassLoaderReflectCache(J9VMThread * currentThread, J9HashTable * classPair
...
+{
+	J9HashTableState hashTableState = {0};
+	J9JVMTIClassPair *classPair = hashTableStartDo(classPairs, &hashTableState);
+	J9InternalVMFunctions *vmFuncs = currentThread->javaVM->internalVMFunctions;
+
+	while (NULL != classPair) {
+		J9Class *originalRAMClass = classPair->originalRAMClass;
+		J9Class *replacementRAMClass = classPair->replacementClass.ramClass;
+		J9ClassLoader *classLoader = originalRAMClass->classLoader;
+		J9SRP *nestMembers = J9ROMCLASS_NESTMEMBERS(originalRAMClass->romClass);
```
gacholio: classLoader and nestMembers should be inside the NULL checks below.

gacholio:
Please squash and I'll launch the sanity builds.

gacholio:
jenkins test sanity plinux

PR #1412, opened 2018-03-09
Disallow nestmate changes during class redefinition

taliamccormick:
With the introduction of JEP 181, classes belong to a 'nest' - a group of classes that share access between private fields. In order to define a class's nest, class files gain two nestmates attributes: 'NestHost' and 'NestMembers'. When redefining a class, there are restrictions on the type of redefinition allowed. In particular, classes can not change their nestmates attributes. This commit adds verification to prevent these nestmate attribute changes from occuring.

Related issue here: https://github.com/eclipse/openj9/issues/1165

It corresponds to jvmti tests here: https://github.com/eclipse/openj9/pull/1262

### DIFF
--- a/runtime/include/jvmti.h
+++ b/runtime/include/jvmti.h
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 1991, 2017 IBM Corp. and others
+ * Copyright (c) 1991, 2018 IBM Corp. and others
  *
  * This program and the accompanying materials are made available under
  * the terms of the Eclipse Public License 2.0 which accompanies this
@@ -340,6 +340,9 @@ typedef enum jvmtiError {
 	JVMTI_ERROR_NAMES_DONT_MATCH = 69,
 	JVMTI_ERROR_UNSUPPORTED_REDEFINITION_CLASS_MODIFIERS_CHANGED = 70,
 	JVMTI_ERROR_UNSUPPORTED_REDEFINITION_METHOD_MODIFIERS_CHANGED = 71,
+#if defined(J9VM_OPT_VALHALLA_NESTMATES)
+	JVMTI_ERROR_UNSUPPORTED_REDEFINITION_CLASS_ATTRIBUTE_CHANGED = 72,
+#endif /* defined(J9VM_OPT_VALHALLA_NESTMATES) */
 	JVMTI_ERROR_UNMODIFIABLE_CLASS = 79,
 	JVMTI_ERROR_UNMODIFIABLE_MODULE = 80,
 	JVMTI_ERROR_NOT_AVAILABLE = 98,
--- a/runtime/util/hshelp.c
+++ b/runtime/util/hshelp.c
@@ -3046,6 +3046,40 @@ verifyClassesAreCompatible(J9VMThread * currentThread, jint class_count, J9JVMTI
 			return rc;
 		}
 
+#if defined(J9VM_OPT_VALHALLA_NESTMATES)
+		/* Verify that the nestmates attributes - nest host & nest members - are the same */
+		{
+			J9UTF8 *originalNestHostName = J9ROMCLASS_NESTHOSTNAME(originalROMClass);
+			J9UTF8 *replacementNestHostName = J9ROMCLASS_NESTHOSTNAME(replacementROMClass);
+			U_16 nestMemberCount = originalROMClass->nestMemberCount;
+
+			/* Nest hosts must both be null or must both contain the same string */
+			if ((NULL != originalNestHostName) || (NULL != replacementNestHostName) {
+				if ((NULL == originalNestHostName) || (NULL == replacementNestHostName) {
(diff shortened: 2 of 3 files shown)


runtime/util/hshelp.c
```diff
@@ -3046,6 +3046,33 @@ verifyClassesAreCompatible(J9VMThread * currentThread, jint class_count, J9JVMTI
 			return rc;
 		}
 
+#if defined(J9VM_OPT_VALHALLA_NESTMATES)
+		/* Verify that the nestmates attributes - nest host & nest members - are the same */
+
+		if (J9ROMCLASS_NESTHOSTNAME(originalROMClass) != J9ROMCLASS_NESTHOSTNAME(replacementROMClass)) {
```
gacholio: Should this be a UTF comparison rather than identity?

runtime/include/jvmti.h
```diff
@@ -340,6 +340,9 @@ typedef enum jvmtiError {
 	JVMTI_ERROR_NAMES_DONT_MATCH = 69,
 	JVMTI_ERROR_UNSUPPORTED_REDEFINITION_CLASS_MODIFIERS_CHANGED = 70,
 	JVMTI_ERROR_UNSUPPORTED_REDEFINITION_METHOD_MODIFIERS_CHANGED = 71,
+#if defined(J9VM_OPT_VALHALLA_NESTMATES)
+	JVMTI_ERROR_UNSUPPORTED_REDEFINITION_CLASS_ATTRIBUTE_CHANGED = 72,
```
gacholio: Is there a new JVMTI spec that adds this error?

runtime/util/hshelp.c
```diff
@@ -3046,6 +3046,33 @@ verifyClassesAreCompatible(J9VMThread * currentThread, jint class_count, J9JVMTI
...
+			} else if (0 != nestMemberCount) {
+				J9SRP *originalNestMembers = J9ROMCLASS_NESTMEMBERS(originalROMClass);
+				J9SRP *replacementNestMembers = J9ROMCLASS_NESTMEMBERS(replacementROMClass);
+				U_16 i = 0;
+
+				for (i = 0; i < nestMemberCount; i++) {
+					J9UTF8 *originalNestMemberName = NNSRP_GET(originalNestMembers[i], J9UTF8*);
+					J9UTF8 *replacementNestMemberName = NNSRP_GET(replacementNestMembers[i], J9UTF8*);
+					if (!J9UTF8_EQUALS(originalNestMemberName, replacementNestMemberName)) {
+						return JVMTI_ERROR_UNSUPPORTED_REDEFINITION_CLASS_ATTRIBUTE_CHANGED;
```
gacholio: This loop enforces ordering of nest members - is reodering not allowed?

gacholio: We do the same thing for implemented interfaces (which may also not be correct).

runtime/include/jvmti.h
taliamccormick: Yes - mailing list link is here: http://mail.openjdk.java.net/pipermail/valhalla-spec-experts/2018-February/000541.html

In the web revision, it adds:
```
 JVMTI_ERROR_UNSUPPORTED_REDEFINITION_CLASS_ATTRIBUTE_CHANGED (72)
    A new class version has unsupported differences in class attributes.
```
(found under 'Function Specific Agent Errors' here: http://cr.openjdk.java.net/~dholmes/8010319/specs/JVM-TI/diff.html)

runtime/util/hshelp.c
taliamccormick: Changing the NestHost or NestMembers class attributes is disallowed in the current spec draft.

In the version that I've been working from, it writes:

```
Unless the canUnrestrictedlyRedefineClasses capability is present the following redefinitions are restricted:
    changing the schema (the fields)
    changing the hierarchy (superclasses, interfaces)
    deleting a method
    changing class modifiers
    changing method modifiers
    changing the NestHost or NestMembers class attributes
```
(found under 'RedefineClasses Command (18)' here: http://cr.openjdk.java.net/~dholmes/8010319/specs/JDWP/diff.html)

My interpretation of this is that re-ordering the nest members counts as modifying them

runtime/util/hshelp.c
gacholio: I suppose some NULL checking would need to be done as well.

runtime/util/hshelp.c
gacholio: OK

runtime/include/jvmti.h
gacholio: OK

runtime/util/hshelp.c
taliamccormick: Missed that - thanks

taliamccormick: Missed that - thanks. Corrected now.

gacholio:
Looks good, please squash and I'll launch the sanity builds.

gacholio:
jenkins test sanity plinux

PR #1417, opened 2018-03-11
Add missing dependencies on generate_redirector

keithc-ca:
* required to produce generated.i for ddr, otherwise problems like this occur:
```
make  -C ../redirector -f redirector_jvm_b150.mk ddrgen
make[1]: *** No rule to make target 'generated.c', needed by 'generated.i'.  Stop.
makefile:78: recipe for target 'ddrgen_redirector_jvm_b150' failed
```

* all redirectors should have the same requirements


DanHeidinga:
Jenkins test sanity plinux

PR #1424, opened 2018-03-12
Use Adopt-OpenJ9 on ppcle systems for bootjdk

jdekonin:

### DIFF
--- a/buildenv/jenkins/variables/defaults.yml
+++ b/buildenv/jenkins/variables/defaults.yml
@@ -40,7 +40,7 @@ slack_channel: '#jenkins'
 linux_ppc-64_cmprssptrs_le:
   boot_jdk:
     8: '/usr/lib/jvm/java-7-openjdk-ppc64el'
-    9: '/home/jenkins/openj9/ibm-java-ppc64le-80'
+    9: '/usr/lib/jvm/adoptojdk-java-ppc64le-80'
   release:
     8: 'linux-ppc64-normal-server-release'
     9: 'linux-ppc64le-normal-server-release'
@@ -53,7 +53,7 @@ linux_ppc-64_cmprssptrs_le:
 #========================================#
 linux_390-64_cmprssptrs:
   boot_jdk:
-    8: '/usr/lib/jvm/adoptojdk-java-s390x-80'
+    8: '/usr/lib/jvm/ibm-java-s390x-70'
     9: '/usr/lib/jvm/adoptojdk-java-s390x-80'
   release:
     8: 'linux-s390x-normal-server-release'


jdekonin:
I have manually verified openj9 ci machines have the sdk present on the system, but have not verified compilation differences between openjdk and ibm sdks.

Currently compiles on ppcle are ~20 mins for JDK8 and ~15 mins for JDK9.  If the change to using what AdoptOpenJDK is using as a default negatively impacts these time, there will need to be a review.

DanHeidinga:
@jdekonin Are we able to use OpenJ9 as the bootJDK?  I would rather that we dogfood our own VM instead of using OJDK.

keithc-ca:
The flip side of 'dogfooding' is that we might unknowingly have a dependency on OpenJ9 such that a build would fail using OpenJDK as the boot JDK.

DanHeidinga:
Agreed @keithc-ca - that's a real risk.  Given Adopt is using OJDK as the bootJDK, I think we'll detect that issue fairly early.

I suggested to Joe that we get an extension mechanism in place so Adopt can continue to use OJDK and we can override that with OJ9 on platforms where it works today and roll that out to all platforms as issues are addressed.

DanHeidinga:
My understanding is that on Linux s390, the only OJDK build is interpreter only which will severely affect build times if we're unable to use OJ9.

keithc-ca:
Yes, I think OpenJDK for s390 is interpreter only.

We'll also mitigate the risk if we have at least one build that uses OpenJDK as the boot JDK.

DanHeidinga:
@jdekonin A path forward is to change the OpenJ9 machines to use an OpenJ9-based binary from Adopt to build OpenJ9.  We can start by hardcoding it similar to the existing SDKs and when we commonize the infrastructure scripts, we can make the bootjdk configurable so we can use OJ9 and Adopt can choose which JDK to use.

jdekonin:
WIP because I still need to add to all the machines.

zLinux SDKs are already using Adopt-OpenJ9 as are AIX even though its not obvious from the directory name.  I'll work to add as an extra install in the AdoptOpenJDK/infrastructure scripts separately.

jdekonin:
Machines are configured now.  The last push I did was a rebase only.

DanHeidinga:
@jdekonin can you confirm its valid to use 8 as the bootjdk for 8?

jdekonin:
http://hg.openjdk.java.net/jdk8/jdk8/file/1773f1fd0fac/README-builds.html
```
Building JDK 8 requires use of a version of JDK 7 that is at Update 7 or newer. 
JDK 8 developers should not use JDK 8 as the boot JDK, to ensure that JDK 8 dependencies 
are not introduced into the parts of the system that are built with JDK 7.
```
So I guess I'll drop the bootjdks for 8 to use openjdk7.

jdekonin:
@DanHeidinga, next action is yours.

DanHeidinga:
Jenkins test sanity plinux zlinux

DanHeidinga:
Jenkins test sanity plinux

DanHeidinga:
Jenkins test sanity plinux zlinux

DanHeidinga:
I'm having trouble launching a PR build for this as the jenkins pr builder has started respecting the `skip ci` tag.  Maybe a setting change on Jenkins?

DanHeidinga:
@jdekonin as ridiculous as this is, can you amend your commit to remove the skip ci text so I can launch the PR build?

DanHeidinga:
Jenkins test sanity plinux zlinux

DanHeidinga:
Jenkins test sanity plinux zlinux

DanHeidinga:
Jenkins test sanity plinux

DanHeidinga:
Jenkins test sanity zlinux

Issue #1271, opened 2018-02-26
ConstantDynamic: Building ROM and RAM Classes


tajila:
<h1>Marking CPEntries</h1>

Marking which CP entries are used by `ldc` will have to be updated to account for Condy entries.

`computeConstantMapAndSizes` will also have to be updated to include Condy CP entries. 

<h1>Add new CP entry type</h1>
Add new CP description types to BCT_J9DescriptionCPType and J9DescriptionBits (j9nonbuilder.h) for Condy. This will be used to detect the CP type in the bytecode interpreter during ldc (ldc_2 and ldc2_w) bytecodes.  
 
Update ConstantPoolWriter (RomClassWriter.cpp) to add `visitCondyType`. This will write out CP entries for Condy in the romConstantPool. The first CP entry slot is a SRP to the NameAndSignature. The second slot is the bootstrap Method index or’ed with the J9DescriptionCPType.

<table>
<tr><td>Slot1: SRP to NAS</td></tr>
<tr>
<td>Slot2: Index of BSM (shifted up by 4) | Cptype</td>
</tr>
</table>

This  matches the RomStringRef type
```
typedef struct J9ROMStringRef {
	J9SRP utf8Data;
	U_32 cpType;
} J9ROMStringRef;
```
The first 4bits of second slot in the romCPEntry must contain the cpType. This is checked by the ldc bytecode to determine what type of resolution to perform for the cp entry.

Update `constantPoolDo` in ConstantPoolMap for the new cp entry.

<h1> RamClass layout </h1>
The ramConstantPool entry for Condy is:
<table>
<tr><td>Slot1: (j9object_t) Constant</td></tr>
<tr>
<td>Slot2: (j9object_t) exception</td>
</tr>
</table>

This is done to match existing RamStringRefs. By convention the first slot is NULL checked to see if the ref requires resolution. For this reason the Condy constant must be the first slot.

```
typedef struct J9RAMStringRef {
	j9object_t stringObject;
	UDATA unused;
} J9RAMStringRef;
```

DanHeidinga:
> Add new CP description types to BCT_J9DescriptionCPType and J9DescriptionBits (j9nonbuilder.h) for Indy and Condy.

As you pointed out on the whiteboard, moving Indy CP entries back into the constantpool will increase the romclass size slightly in the common case of invokedynamics not sharing cp entries.

This design can do the Condy side and leave the Indy representation as is.

DanHeidinga:
[...]

PR #1426, opened 2018-03-12
Write CONSTANT_DYNAMIC entry to ROMClass

fengxue-IS:
- mark ConstantDynamic entry in constantpool
- add visit function to ROMClass Writer
- add new CP type for ConstantDynamic
- fixed DDR mapping of new CP type

Close: #1271

### DIFF
--- a/runtime/bcutil/ClassFileOracle.cpp
+++ b/runtime/bcutil/ClassFileOracle.cpp
@@ -30,8 +30,6 @@
 #include "ROMClassVerbosePhase.hpp"
 
 
-#include "bcnames.h"
-#include "cfreader.h"
 #include "j9port.h"
 #include "jbcmap.h"
 #include "ut_j9bcu.h"
@@ -1525,12 +1523,17 @@ ClassFileOracle::walkMethodCodeAttributeCode(U_16 methodIndex)
 
 		case CFR_BC_ldc2_w:
 			cpIndex = PARAM_U16();
-			markConstantAsUsedByLDC2W(cpIndex);
 			addBytecodeFixupEntry(entry++, codeIndex + 1, cpIndex, ConstantPoolMap::LDC2W);
+
 			if (isConstantLong(cpIndex)) {
 				code[codeIndex + 0] = JBldc2lw;
+				markConstantAsUsedByLDC2W(cpIndex);
 			} else if (isConstantDouble(cpIndex)) {
 				code[codeIndex + 0] = JBldc2dw;
+				markConstantAsUsedByLDC2W(cpIndex);
+			} else if (isConstantDynamic(cpIndex)) {
+				code[codeIndex + 0] = constantDynamicType(cpIndex);
+				markConstantDynamicAsReferenced(cpIndex);
 			} else {
 				Trc_BCU_Assert_ShouldNeverHappen();
 			}
@@ -2324,6 +2327,15 @@ ClassFileOracle::markConstantNameAndTypeAsReferenced(U_16 cpIndex)
 }
 
 void
+ClassFileOracle::markConstantDynamicAsReferenced(U_16 cpIndex)
+{
+	markNameAndDescriptorAsReferenced(_classFile->constantPool[cpIndex].slot2);
+	if (0 != cpIndex) { /* Never, never, never mark constantPool[0] referenced */
+		_constantPoolMap->markConstantAsReferenced(cpIndex);
+	}
+}
+
+void
 ClassFileOracle::markConstantAsUsedByAnnotation(U_16 cpIndex)
 {
 	UDATA cpTag = getCPTag(cpIndex);
@@ -2498,6 +2510,9 @@ ClassFileOracle::markConstantBasedOnCpType(U_16 cpIndex, bool assertNotDoubleOrL
 			break;
 		}
 		_constantPoolMap->markConstantAsReferencedDoubleSlot(cpIndex);
+		break;
+	case CFR_CONSTANT_Dynamic:
+		markConstantDynamicAsReferenced(cpIndex);
 		break;
 	default:
 		Trc_BCU_Assert_ShouldNeverHappen();
--- a/runtime/bcutil/ClassFileOracle.hpp
+++ b/runtime/bcutil/ClassFileOracle.hpp
@@ -32,6 +32,7 @@
 #include "cfr.h"
 #include "cfreader.h"
 #include "ut_j9bcu.h"
+#include "bcnames.h"
 
 #include "BuildResult.hpp"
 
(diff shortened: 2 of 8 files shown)


fengxue-IS:
Reviewer: @tajila

tajila:

runtime/bcutil/ClassFileOracle.cpp
```diff
@@ -2482,6 +2494,13 @@ ClassFileOracle::markConstantBasedOnCpType(U_16 cpIndex, bool assertNotDoubleOrL
 		}
 		_constantPoolMap->markConstantAsReferencedDoubleSlot(cpIndex);
 		break;
+	case CFR_CONSTANT_Dynamic:
+		if (assertNotDoubleOrLong) {
+			markConstantDynamicAsReferenced(cpIndex);
```
DanHeidinga: Constant_Dynamic can create a long or a double.  Can you explain why the check is necessary here?

runtime/bcutil/ClassFileOracle.hpp
```diff
@@ -921,6 +923,16 @@ class NameAndTypeIterator
...
 	bool isInnerClass() const { return _isInnerClass; }
 
+	U_8 constantDynamicType(U_16 cpIndex) const
+	{
+		J9CfrConstantPoolInfo* nas = &_classFile->constantPool[_classFile->constantPool[cpIndex].slot2];
+		if ('D' == nas->bytes[nas->slot1 - 1]) {
+			return JBldc2dw;
+		} else {
+			return JBldc2lw;
+		}
```
DanHeidinga: What about ConstantDynamics that don't return long or double?

runtime/bcutil/ROMClassWriter.cpp
```diff
@@ -508,6 +508,12 @@ class ROMClassWriter::ConstantPoolWriter : public ConstantPoolMap::ConstantPoolV
 		_cursor->writeU32((cfrKind << BCT_J9DescriptionCpTypeShift) | BCT_J9DescriptionCpTypeMethodHandle, Cursor::GENERIC);
 	}
 
+	void visitConstantDynamic(U_16 cfrCPIndex, U_16 bsmIndex)
+	{
```
DanHeidinga: Can you add an `/* assumes format: { ......} */` comment to this to match all the other visit* methods?

runtime/bcutil/ConstantPoolMap.cpp
```diff
@@ -105,7 +105,7 @@ ConstantPoolMap::computeConstantPoolMapAndSizes()
 		J9CPTYPE_UNUSED, /* 14 */
 		J9CPTYPE_METHODHANDLE, /* CFR_CONSTANT_MethodHandle */
 		J9CPTYPE_METHOD_TYPE, /* CFR_CONSTANT_MethodType */
-		J9CPTYPE_UNUSED, /* 17 */
+		J9CPTYPE_CONSTANT_DYNAMIC, /* CFR_CONSTANT_Dynamic */
```
DanHeidinga: Should this be 17 or 18?  If it's meant to be 18 it needs an UNUSED to fill the 17th slot

runtime/oti/j9nonbuilder.h
```diff
@@ -2124,6 +2126,7 @@ typedef struct J9ConstantPool {
 #define J9CPTYPE_METHOD_TYPE  13
 #define J9CPTYPE_METHODHANDLE  14
 #define J9CPTYPE_ANNOTATION_UTF8  15
+#define J9CPTYPE_CONSTANT_DYNAMIC 18
```
DanHeidinga: 17 or 18?  The numbering here and above should agree

runtime/bcutil/ClassFileOracle.cpp
fengxue-IS: for constantDynamic, this marking function should only be called from `ldc` and `ldc_w` which is not allowed to return a long or double. 
I ll double check with the static verify functionality if this check will ever be hit

runtime/bcutil/ConstantPoolMap.cpp
fengxue-IS: Constant Dynamic is 17

runtime/oti/j9nonbuilder.h
fengxue-IS: J9CPTYPE is our ROM class tag which doesn't match classfile CP due to optimizations (split entries for methodref and others)

runtime/bcutil/ClassFileOracle.hpp
fengxue-IS: this is a helper for `ldc2_w` bytecode marking which requires a special bytecode entry (jit optimization)
For ConstantDynamics that are `ldc` or `ldc_w` they will not require a special bytecode entry
(ie. `addBytecodeFixupEntry(entry++, codeIndex + 1, cpIndex, ConstantPoolMap::REFERENCED);` should cover that case)

runtime/oti/j9nonbuilder.h
fengxue-IS: Tag 16, 17 are used for the sharedCPEntry consistency check.
I'll move this definition down so the these appear in order

runtime/bcutil/ClassFileOracle.hpp
DanHeidinga: I need more detail on this.  Not currently following.

runtime/oti/j9nonbuilder.h
DanHeidinga: It would be better to move `J9CPTYPE_INTERFACE_INSTANCE_METHOD` and keep `J9CPTYPE_CONSTANT_DYNAMIC` as 17.  Its clearer when we can be consistent between our internal numbers and the ones in the spec

runtime/bcutil/ClassFileOracle.cpp
fengxue-IS: confirmed with @taliamccormick that static verify would catch the case when `ldc` or `ldc_w` has a double or long return type. I ll remove this check then

runtime/bcutil/ClassFileOracle.hpp
fengxue-IS: The VM have special bytecode (`JBldc2dw / JBldc2lw`) for loading with long and double in the interpreter depending wether it is a long or double (I believe it was for jit optimization).

this function is used to help determine the expected return type from `ldc2_w` instruction and return the correct fixup bytecode entry.

In the case that ConstantDynamic used in `ldc / ldc_w`, there is no special bytecode for different types, so only a fixup entry is required.
ie. `ClassFileOracle.cpp - line 1455 && 1506 > CFR_BC_ldc and CFR_BC_ldc_w`

runtime/bcutil/ROMClassWriter.cpp
fengxue-IS: added

runtime/oti/j9nonbuilder.h
fengxue-IS: I'd like to keep `J9CPTYPE_INTERFACE_STATIC_METHOD & J9CPTYPE_INTERFACE_INSTANCE_METHOD` continuous as they were used together.
I'll move ConstantDynamic to 17 and static/instance to 18,19 respectively, would that be okay?

DanHeidinga: Fine by me.  Don't forget to update the DDR / cfdump code when moving the constants

fengxue-IS:
In DDR dumpConstantPool(), ConstantDynamic is represented by `c` (lowercase)

runtime/bcutil/ClassFileOracle.hpp
```diff
@@ -921,6 +923,17 @@ class NameAndTypeIterator
 	bool annotationRefersDoubleSlotEntry() const { return _annotationRefersDoubleSlotEntry; }
 	bool isInnerClass() const { return _isInnerClass; }
 
+	U_8 constantDynamicType(U_16 cpIndex) const
+	{
+		J9CfrConstantPoolInfo* nas = &_classFile->constantPool[_classFile->constantPool[cpIndex].slot2];
+		J9CfrConstantPoolInfo* signature = &_classFile->constantPool[nas->slot2];
+		if ('D' == signature->bytes[signature->slot1 - 1]) {
+			return JBldc2dw;
+		} else {
```
DanHeidinga: If this is only intended to be used by `ldc2*` bytecodes, then please make this explicitly check the two cases and assert otherwise.

fengxue-IS: fixed

DanHeidinga:
Jenkins test sanity plinux

DanHeidinga:
I should have checked the failed travis build before merging this.  It caught the compile issue.  Odd that the other platforms didn't

Issue #1353, opened 2018-03-06
Open source Apache Tomcat 8.5.24 builds but tests fail


chandrams:
Executed Tomcat 8.5.24 tests, on both Openjdk8 and Openjdk9 builds below are the observations:

**With Openjdk8 and Openjdk9 ** - tests passed
**With Openjdk8-openj9 and Openjdk9-openj9 ** - 2 tests failed
```
test:
   [concat] Testsuites with skipped tests:
   [concat] TEST-org.apache.catalina.connector.TestMaxConnections.NIO2.txt
   [concat] TEST-org.apache.catalina.connector.TestRequest.NIO.txt
   [concat] TEST-org.apache.catalina.connector.TestRequest.NIO2.txt
   [concat] TEST-org.apache.catalina.core.TestDefaultInstanceManager.NIO.txt
   [concat] TEST-org.apache.catalina.core.TestDefaultInstanceManager.NIO2.txt
   [concat] TEST-org.apache.catalina.loader.TestWebappClassLoaderThreadLocalMemoryLeak.NIO.txt
   [concat] TEST-org.apache.catalina.loader.TestWebappClassLoaderThreadLocalMemoryLeak.NIO2.txt
   [concat] TEST-org.apache.coyote.http11.upgrade.TestUpgradeInternalHandler.NIO.txt
   [concat] TEST-org.apache.el.parser.TestELParser.NIO.txt
   [concat] TEST-org.apache.el.parser.TestELParser.NIO2.txt
   [concat] TEST-org.apache.tomcat.util.net.TestSsl.NIO.txt
   [concat] TEST-org.apache.tomcat.util.net.openssl.TestOpenSSLConf.NIO.txt
   [concat] TEST-org.apache.tomcat.util.net.openssl.TestOpenSSLConf.NIO2.txt
   [concat] TEST-org.apache.tomcat.util.net.openssl.ciphers.TestOpenSSLCipherConfigurationParser.NIO.txt
   [concat] TEST-org.apache.tomcat.util.net.openssl.ciphers.TestOpenSSLCipherConfigurationParser.NIO2.txt
   [concat] TEST-org.apache.tomcat.websocket.TestConnectionLimit.NIO.txt
   [concat] TEST-org.apache.tomcat.websocket.TestConnectionLimit.NIO2.txt
   [concat] TEST-org.apache.tomcat.websocket.TestWsWebSocketContainerWithProxy.NIO.txt
[...]

PR #1427, opened 2018-03-12
Avoid instantiating ClassLoader objects when throwing SecurityException

JasonFengJ9:
Avoid instantiating ClassLoader objects when throwing SecurityException

Removed flag `initDone` and related set & check code;
Added a static helper method `checkSecurityPermission` to check if callers have permission required;
Refactored `ClassLoader` constructors to invoke `checkSecurityPermission`.

closes: #1353

Reviewer @DanHeidinga 
FYI @paulcheeseman

### DIFF
--- a/jcl/src/java.base/share/classes/java/lang/ClassLoader.java
+++ b/jcl/src/java.base/share/classes/java/lang/ClassLoader.java
@@ -97,8 +97,6 @@ public abstract class ClassLoader {
 	private long vmRef;
 	ClassLoader parent;
 	
-	private boolean initDone;
-	
 	/*[IF Panama]*/
 	private static String[]	usr_paths = new String[1];
 	/*[ENDIF]*/
@@ -312,10 +310,23 @@ public abstract class ClassLoader {
  *					allow the creation of new ClassLoaders.
  */
 protected ClassLoader() {
-	this(applicationClassLoader);
+	this(checkSecurityPermission(), null, applicationClassLoader);
 }
 
-/*[IF Sidecar19-SE]*/
+/**
+ * This is a static helper method to perform security check earlier such that current ClassLoader object 
+ * can't be resurrected when there is a SecurityException thrown.
+ *
+ * @return Void a unused reference passed to the Constructor
+ */
+private static Void checkSecurityPermission() {
+	SecurityManager security = System.getSecurityManager();
+	if (security != null) {
+		security.checkCreateClassLoader();
+	}
+	return null;
+}
+
 /**
  * Constructs a new instance of this class with the given
  * class loader as its parent.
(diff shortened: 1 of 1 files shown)


jcl/src/java.base/share/classes/java/lang/ClassLoader.java
```diff
@@ -352,44 +360,35 @@ protected ClassLoader(ClassLoader parentLoader) {
  *@since 9
  */
 protected ClassLoader(String classLoaderName, ClassLoader parentLoader) {
+	this(checkSecurityPermission(), classLoaderName, parentLoader);
+}
+
+private ClassLoader(Void staticMethodHolder, ClassLoader parentLoader) {
```
DanHeidinga: Is this constructor necessary?  Won't this cause the security check to be run multiple times?

ie: 
```
protected ClassLoader(ClassLoader parentLoader) {
   this(checkSecurityPermission(), parentLoader);
}
private ClassLoader(Void staticMethodHolder, ClassLoader parentLoader) {
	this(checkSecurityPermission(), null, parentLoader);
}
private ClassLoader(Void staticMethodHolder,
 /*[IF Sidecar19-SE]*/
 	String classLoaderName,
 /*[ENDIF]*/
 	ClassLoader parentLoader) {
```

If it's only used by this class, can we pass `null` for the classloader name and bypass it?

JasonFengJ9: Good catch. Initial thought was to avoid introducing `Java 9` decoration into constructors for `Java 8` as well.
Fixed to remove the constructor in question, and decorate `null` for the classloader name with flag `Sidecar19-SE`. Verified that `pConfig` still compiles.
Please review again.

jcl/src/java.base/share/classes/java/lang/ClassLoader.java
```diff
@@ -352,44 +368,31 @@ protected ClassLoader(ClassLoader parentLoader) {
  *@since 9
  */
 protected ClassLoader(String classLoaderName, ClassLoader parentLoader) {
+	this(checkSecurityPermission(), classLoaderName, parentLoader);
+}
+/*[ENDIF]*/
+
+private ClassLoader(Void staticMethodHolder,
+/*[IF Sidecar19-SE]*/
+	String classLoaderName,
```
DanHeidinga: Since this is a private constructor, can we remove the preprocessor tests and just ignore the `classLoaderName` in JDK8?  It would remove the preprocessor checks in the other constructors as well.

JasonFengJ9: Agreed, passing `null` to the private `constructor` can avoid `JPP` decoration since `classLoaderName` already is ignored by `Java 8` via existing `JPP`, now the code looks cleaner, thanks. Verified that `pConfig` still compiles. 
Please have another look.

JasonFengJ9:
@DanHeidinga I have updated the comment regarding `checkSecurityPermission`, please have another look.

DanHeidinga:
Jenkins test sanity plinux

Issue #1419, opened 2018-03-12
omrvmem.c asserts with hugepages in Docker


JamesKingdon:
I noticed that when I had 1G hugepages configured in the host OS, running the JVM in Docker (in this case as part of the build) failed with an assert:

01:51:38.216 0x23c8a00 omrport.0      *   ** ASSERTION FAILED ** at ../../omr/port/linux/omrvmem.c:694: ((params->startAddress <= params->endAddress))

This is likely being caused by some sort of configuration problem with Docker (the Docker container was created before the hugepages were configured in the host OS), but it would be nice if we could handle the situation more gracefully.

JamesKingdon:
I started hitting this outside of Docker with a prototype build that can use multiple hugepage sizes. It's possible there is a failure path that isn't being handled correctly when shmget calls fail to return memory (or possibly my very hacky prototype has just broken something and it's a complete coincidence that it falls over at the same assert)

andrewcraik:
@mpirvu this looks familiar?

mpirvu:
Yes, Harry Yu is looking at a similar issue.
@haryyu1994

PR #1436, opened 2018-03-13
Fix code cache allocation issue with hugepages

harryyu1994:
- disable the use of J9PORT_VMEM_ADDRESS_HINT for large pages
- fix chooseCacheStartAddress for large pages
- extend search range by increasing SAFE_DISTANCE_REPOSITORY_JITLIBRARY from 24MB to 64MB

Closes: #1419

### DIFF
--- a/runtime/compiler/trj9/runtime/J9CodeCacheManager.cpp
+++ b/runtime/compiler/trj9/runtime/J9CodeCacheManager.cpp
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2000, 2017 IBM Corp. and others
+ * Copyright (c) 2000, 2018 IBM Corp. and others
  *
  * This program and the accompanying materials are made available under
  * the terms of the Eclipse Public License 2.0 which accompanies this
@@ -304,8 +304,12 @@ J9::CodeCacheManager::allocateCodeCacheSegment(size_t segmentSize,
 
    if (preferredStartAddress)
       {
+   // J9PORT_VMEM_ADDRESS_HINT is only supported for default pages and linux
 #if defined(LINUX)
-      vmemParams.options |= J9PORT_VMEM_ADDRESS_HINT;
+      if (largeCodePageSize == 0)
+         {
+         vmemParams.options |= J9PORT_VMEM_ADDRESS_HINT;
+         }
 #endif
       vmemParams.options |= J9PORT_VMEM_STRICT_ADDRESS; // if we cannot allocate a block whose start is between preferredStartAddress and endAddress, return NULL
       vmemParams.startAddress = preferredStartAddress;
@@ -344,29 +348,31 @@ J9::CodeCacheManager::allocateCodeCacheSegment(size_t segmentSize,
                                                                       segmentType,
                                                                       &vmemParams);
 
+   uintptr_t someJitLibraryAddress = getSomeJitLibraryAddress();
+   // isInRange() checks whether the allocated codeCacheSegment baseAddress is in range of JitLibrary to avoid trampoline
    if (codeCacheSegment &&
        (vmemParams.options & J9PORT_VMEM_ADDRESS_HINT) &&
-       !isInRangeToAvoidTrampoline((uintptr_t)(codeCacheSegment->baseAddress)))
+       !isInRange((uintptr_t)(codeCacheSegment->baseAddress), someJitLibraryAddress, MAX_DISTANCE_NEAR_JITLIBRARY_TO_AVOID_TRAMPOLINE))
       {
       // allocated code cache is not in range to avoid trampoline 
       // try with full address range that avoids trampoline
       // free old segment
       javaVM->internalVMFunctions->freeMemorySegment(javaVM, codeCacheSegment, 1);
-      uintptr_t someJitLibraryAddress = getSomeJitLibraryAddress();
       size_t alignment = 2 * 1024 * 1024;
-      // if jit library address is greater than UPPER_BOUND_DISTANCE, then we search in the addresses before the jit library address
-      // if jit library address is less than UPPER_BOUND_DISTANCE, which means we do not have the full UPPER_BOUND_DISTANCE(2GB-24MB) range before the jit library
+
+      // if jit library address is greater than MAX_DISTANCE, then we search in the addresses before the jit library address
+      // if jit library address is less than MAX_DISTANCE, which means we do not have the full MAX_DISTANCE(2GB-24MB) range before the jit library
       // in this case we search in the addresses after the jit library address
-      if (someJitLibraryAddress > UPPER_BOUND_DISTANCE_NEAR_JITLIBRARY_TO_AVOID_TRAMPOLINE)
+      if (someJitLibraryAddress > MAX_DISTANCE_NEAR_JITLIBRARY_TO_AVOID_TRAMPOLINE)
          {
-         // align the startAddress to 2MB page boundary
-         vmemParams.startAddress = (void *)align((uint8_t *)(someJitLibraryAddress - UPPER_BOUND_DISTANCE_NEAR_JITLIBRARY_TO_AVOID_TRAMPOLINE), alignment - 1);
+         // align the startAddress to page boundary
+         vmemParams.startAddress = (void *)align((uint8_t *)(someJitLibraryAddress - MAX_DISTANCE_NEAR_JITLIBRARY_TO_AVOID_TRAMPOLINE), alignment - 1);
          vmemParams.endAddress = preferredStartAddress;
          }
       else
          {
          vmemParams.startAddress = (void *)align((uint8_t *)(someJitLibraryAddress + SAFE_DISTANCE_REPOSITORY_JITLIBRARY), alignment -1);
-         vmemParams.endAddress = (void *)(someJitLibraryAddress + UPPER_BOUND_DISTANCE_NEAR_JITLIBRARY_TO_AVOID_TRAMPOLINE);
+         vmemParams.endAddress = (void *)(someJitLibraryAddress + MAX_DISTANCE_NEAR_JITLIBRARY_TO_AVOID_TRAMPOLINE);
          }
       // unset STRICT_ADDRESS and ADDRESS_HINT 
       vmemParams.options &= ~(J9PORT_VMEM_STRICT_ADDRESS);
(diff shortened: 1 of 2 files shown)


andrewcraik:
@mpirvu this looks like one in your area of expertise. FYI @liqunl I know you had seen this at some point.

harryyu1994:
when pageSize is >= 1GB, there is a chance that the `preferredStartAddress` is already outside of the `MAX_DISTANCE_NEAR_JITLIBRARY_TO_AVOID_TRAMPOLINE` range. The current code does not consider this case.

I'm proposing the solution that when `preferredStartAddress` is already outside the range, we do not reject the returned allocated address from port library because we simply can't do better.

runtime/compiler/trj9/runtime/J9CodeCacheManager.cpp
```diff
@@ -344,23 +344,29 @@ J9::CodeCacheManager::allocateCodeCacheSegment(size_t segmentSize,
...
 
+   uintptr_t someJitLibraryAddress = getSomeJitLibraryAddress();
+   // first isInRange() checks whether the allocated segment startAddress is in range to avoid trampoline
+   // second isInRange() checks if the preferredStartAddress is in range to avoid trampoline in the first place, for large pages (>1GB) it's possible they won't be in range
+   // if they are not, we will accept what we get 
    if (codeCacheSegment &&
        (vmemParams.options & J9PORT_VMEM_ADDRESS_HINT) &&
-       !isInRangeToAvoidTrampoline((uintptr_t)(codeCacheSegment->baseAddress)))
+       !isInRange((uintptr_t)(codeCacheSegment->baseAddress), someJitLibraryAddress, MAX_DISTANCE_NEAR_JITLIBRARY_TO_AVOID_TRAMPOLINE) &&
+       isInRange((uintptr_t)(preferredStartAddress), someJitLibraryAddress, MAX_DISTANCE_NEAR_JITLIBRARY_TO_AVOID_TRAMPOLINE))
```
mpirvu: If I remember correctly the ADDRESS_HINT is ignored for large pages because the allocation mechanism is not based on mmap, but on shmget. If that's the case I would simplify the code by not setting `vmemParams.options |= J9PORT_VMEM_ADDRESS_HINT;` In this case and then eliminating this second `isInRange`

runtime/compiler/trj9/runtime/J9CodeCacheManager.cpp
```diff
@@ -526,7 +532,7 @@ J9::CodeCacheManager::chooseCacheStartAddress(size_t repositorySize)
          if (someFunctionPointer > largeCodePageSize * 2)
             {
             // round down to the nearest GB first
-            startAddress = (void *)(((size_t)someFunctionPointer) & ~(largeCodePageSize-1));
+            startAddress = (void *)(((size_t)someFunctionPointer - SAFE_DISTANCE_REPOSITORY_JITLIBRARY) & ~(largeCodePageSize-1));
```
mpirvu: I think that size_t needs to be uintptr_t because it's a pointer.

mpirvu:
I would strongly advise for some local testing because the Jenkins sanity tests we do are not going to exercise this new code. I suggest talking to James Kingdon for help on enabling 1GB pages on a linux machine.

andrewcraik:
I would really like some logging (maybe to the vlog) to warn that trampolines will be required if we can't do better and that the reason was an unfortunate placement in terms of layout. The trampolines could have a surprising negative affect on performance and making sure we record that in a visible way I think will be important for perf investigations.

harryyu1994:
logging can be found here in J9CodeCacheManager.cpp: @andrewcraik 

```
         char * verboseLogString = "The code cache repository was allocated between addresses %p and %p";
         if (preferredStartAddress && isInRange((uintptr_t)(codeCacheSegment->baseAddress), someJitLibraryAddress, MAX_DISTANCE_NEAR_JITLIBRARY_TO_AVOID_TRAMPOLINE))
            {
            verboseLogString = "The code cache repository was allocated between addresses %p and %p to be near the VM/JIT modules to avoid trampolines";
            }
```

mpirvu:
Build fails with:
```
/home/travis/build/eclipse/openj9-openjdk-jdk9/build/linux-x86_64-normal-server-release/vm/compiler/../compiler/trj9/runtime/J9CodeCacheManager.cpp:547:81: error: invalid conversion from ‘uintptr_t {aka long unsigned int}’ to ‘uint8_t* {aka unsigned char*}’ [-fpermissive]
             startAddress = (void *) align((uintptr_t)startAddress, alignment - 1);
```
We should leave the cast as it was before: `uint8_t *`

mpirvu:
Jenkins test all

JamesKingdon:
I have 1G pages enabled on my laptop so I can do some testing. It looks like the issue is primarily around codecache? So I guess running as much 'stuff' as possible with 1G enabled for at least the codecache is what's wanted?

harryyu1994:
The issue was with code cache allocation, I think as long as we allocate code caches on 1G page machine with no issue we are fine.

JamesKingdon:
Just looking over the changes after spending some time getting the idea of how the old code worked. If I'm reading this right, it disables any attempt to place the codecache near the jit library when using either size of hugepages. I don't think we need to be that conservative - it's usually possible to get 2m pages in the right region, and often with 1G pages:
```jdk/bin/java -Xlp:codecache:pagesize=2m -version
#CODECACHE:  The code cache repository was allocated between addresses 00007F5A6C600000 and 00007F5AD9600000 to be near the VM/JIT modules to avoid trampolines

jdk/bin/java -Xlp:codecache:pagesize=1g -version
#CODECACHE:  The code cache repository was allocated between addresses 00007FA500000000 and 00007FA501800000 to be near the VM/JIT modules to avoid trampolines```

I think we should still try and place the codecache in the favourable window, but just make sure we handle cases where that's not possible in a sensible fashion (details TBD and doubtless subject to controversy!)

**Comment by mpirvu** [comment]
@JamesKingdon The code still tries to allocate the code cache repository within 2GB of the JIt dll. The messages you posted above is a testament to that (assuming large pages were enabled on the system)

**Comment by JamesKingdon** [comment]
Those messages are from my prototype, to demonstrate that it is possible for hugepages to be located within range. Doesn't the following change from this PR disable at least some of the code-path (the bit that actually succeeds in my tests) for attempting to allocate near the lib:
```
+      if (largeCodePageSize == 0)
+         {
+         vmemParams.options |= J9PORT_VMEM_ADDRESS_HINT;
+         }
```

**Comment by mpirvu** [comment]
The code above is going to work only for small (4K) pages which are allocated with mmap. We pass a HINT option to mmap so that if mmap cannot allocate at the address we want it will give something reasonably close. We then test to see if the address returned by mmap is within 2GB of the code cache.
For large pages the VM allocates memory using shmget which has no HINT functionality. The strategy there is to repeatedly try allocating within a given region (incrementing the desired address by 2MB each time).

**Comment by harryyu1994** [comment]
I do think that this change would disable hugepages from trying out the full 2GB near JIT library range because it no longer gets into that if statement.

**Comment by harryyu1994** [comment]
Before this change, we try the full 2GB range for both default pages and large pages in the 2nd attempt, with this change we took out the full range for large pages, but maybe we shouldn't

**Comment by harryyu1994** [comment]
I actually think that for non-default pages, we should just go with the full 2GB range in our first attempt, because when pages are larger, the searching won't be as slow as for default pages?

**Comment by harryyu1994** [comment]
@mpirvu Hi Marius, what are your thoughts on this?

**Comment by harryyu1994** [comment]
or in `J9::CodeCacheManager::chooseCacheStartAddress`, when we pick the `startAddress`, should we move a bit further back than just `startAddress = (void *)(((uint8_t *)someFunctionPointer) - safeDistance);`, I feel like this is the closest `startAddress` that can be allocated, which is only 1 point and not a range. Giving 1 point instead of a range is only beneficial for default pages with the HINT flag. For non-default pages, I think we should provide with a larger range in the first attempt.

**Comment by harryyu1994** [comment]
I think the extended testcase failed because of a testcase change, rebasing to latest

**Comment by JamesKingdon** [comment]
For hugepages, it may be worth making a first attempt without a specified address and seeing if that is in range by default, as that seemed to happen fairly often in some of my testing. Failing that, I think what we want to ask for is the address range that is +/- 2G (minus epsilon) of the known function address. If the worst case search time is an issue as you mentioned then we might need to choose a larger probe interval than we currently use, at the risk of not finding a space that we would find with the current interval.
We need to catch the case where we still don't succeed and either fall back to huge pages at any address, or try non-huge pages in the window (with the need to then potentially fall back again to non-huge at any address)

**Comment by mpirvu** [comment]
I agree that for large pages we may want to extend our search range. Currently we have a range of 24MB and some of this may be taken by part of the JIT dll. 2GB/2MB=1000 possible points, so maybe we don't want to go that far.

If using large pages close enough to the JIT dll is not possible I would prefer the usage of large pages to the detriment of helper trampolines, the reasoning being that the user asked for large pages, so let's give her/him that, while the avoidance of helper trampolines is just some internal optimization that the user is not aware.

**Comment by harryyu1994** [comment]
Hi Marius, @mpirvu, I added something that will extend the search range for large pages. Please take a look. I wonder what happens after the first attempt for large pages fails? Do we want to just clear the STRICT flag and allocate anywhere, or should we try a 2nd attempt with the full 2GB search range before we clear it in the 3rd attempt. Right now, we don't have that full 2GB range search for large pages.

**Comment by mpirvu** [comment]
I don't think your code will work for 1GB pages. 20 search points of 1GB each means 20GB which outside of the 2GB range. With 1GB pages we only have one shot at it (maybe 2 if we find space after the JIT dll).

**Comment by harryyu1994** [comment]
Hi James @JamesKingdon could you please verify to see if the assertion still appears when allocating code cache with 1GB pages enabled? If it doesn't then this change is ready to be "Jenkins test all" and merged.

**Comment by mpirvu** [comment]
Jenkins test sanity

**Comment by JamesKingdon** [comment]
I ran some tests with 1G hugepages, both inside and outside of docker and didn't see any asserts. Unfortunately I don't have a reliable repro for the original failure as it was showing up while building OpenJDK, and I can't easily shoe-horn the test build into that scenario.

**Comment by mpirvu** [comment]
Jenkins test sanity aix jdk9

**Comment by harryyu1994** [comment]
The error doesn't seem like it's related to this code change: 
```
16:38:00 Note: /home/u0020236/workspace/PullRequest-Sanity-JDK9-aix_ppc-64_cmprssptrs-OpenJ9/corba/src/java.corba/share/classes/com/sun/tools/corba/se/idl/som/idlemit/MetaPragma.java uses or overrides a deprecated API.
```

mpirvu:
Harry, could you please rebase you code and maybe that will eliminate that AIX -JDK9 error.

harryyu1994:
Hi Marius, it has been rebased, let's give the sanity tests another shot.

mpirvu:
Jenkins test sanity jdk9

mpirvu:
Jenkins test sanity

PR #1442, opened 2018-03-14
Remove redundant value profiling calls

r30shah:
Before we split blocks with emitting profiling trees for value profiling, scan it to remove redundant profiling value calls.

### DIFF
--- a/runtime/compiler/trj9/optimizer/JProfilingValue.cpp
+++ b/runtime/compiler/trj9/optimizer/JProfilingValue.cpp
@@ -181,6 +181,8 @@ TR_JProfilingValue::perform()
       return 0;
       }
 
+   // Scan and remove duplicate value profiling calls before lowering calls
+   removeRedundantProfilingValueCalls(); 
    // Lower all existing calls
    lowerCalls();
 
@@ -203,6 +205,44 @@ TR_JProfilingValue::perform()
    return 1;
    }
 
+void
+TR_JProfilingValue::removeRedundantProfilingValueCalls()
+   {
+   TR::TreeTop *cursor = comp()->getStartTree();
+   TR_BitVector *alreadyProfiledValues = new (comp()->trStackMemory()) TR_BitVector();
+   while (cursor)
+      {
+      TR::Node * node = cursor->getNode();
+      if (node->isProfilingCode() &&
+         node->getOpCodeValue() == TR::treetop &&
+         node->getFirstChild()->getOpCode().isCall() &&
+         (node->getFirstChild()->getSymbolReference()->getReferenceNumber() == TR_jProfile32BitValue ||
+         node->getFirstChild()->getSymbolReference()->getReferenceNumber() == TR_jProfile64BitValue) &&
+         node->getFirstChild()->getNumChildren() == 3)
+         {
+         TR::Node *value = node->getFirstChild()->getFirstChild();
+         
+         if (alreadyProfiledValues->isSet(value->getGlobalIndex()) &&
+               performTransformation(comp(), "%s Removing profiling treetop, node n%dn is already profiled\n",
+                  optDetailString(), value->getGlobalIndex()))
+            /* Found that second and third child were having more than one ref counts. 
+            TR_ASSERT_FATAL(node->getFirstChild()->getSecondChild()->getReferenceCount() == 1 &&
+               node->getFirstChild()->getThirdChild()->getReferenceCount() == 1,
+               "Second and Third Child of the value calls should be referenced only once");
+            */
+            TR::TransformUtil::removeTree(comp(), cursor);
+         else
+            alreadyProfiledValues->set(value->getGlobalIndex());
+         
+         }
+      // Emptying a bit vector after scanning whole extended basic blocks will keep number of bits set in bit vector low.
+      else if (cursor->getNode()->getOpCodeValue() == TR::BBStart && !cursor->getNode()->getBlock()->isExtensionOfPreviousBlock())
+         {
+         alreadyProfiledValues->empty();
+         }
+      cursor = cursor->getNextTreeTop();
+      }    
+   }
 void
 TR_JProfilingValue::performOnNode(TR::Node *node, TR::TreeTop *tt, TR::NodeChecklist *checklist)
    {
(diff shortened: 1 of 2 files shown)


runtime/compiler/trj9/optimizer/JProfilingValue.cpp
```diff
@@ -204,6 +206,47 @@ TR_JProfilingValue::perform()
...
+            TR::TransformUtil::removeTree(comp(), cursor);
+            }
+         else
+            {
+            alreadyProfiledValues->set(value->getGlobalIndex());
+            }
+         }
+      // TODO: Following condition will empty down the BitVectors for next Basic Blocks limiting this to  extended basic blocks only
+      // Although it is a safety measure, do we really need to do that?
+      else if (cursor->getNode()->getOpCodeValue() == TR::BBStart && !cursor->getNode()->getBlock()->isExtensionOfPreviousBlock())
```
andrewcraik: A node cannot appear outside of its extended basic block, but the emptying will keep the number of bits set in the bitvector low so perhaps that is a justification. It may also make debugging easier.

runtime/compiler/trj9/optimizer/JProfilingValue.cpp
```diff
@@ -204,6 +206,47 @@ TR_JProfilingValue::perform()
...
+         {
+         TR::Node *value = node->getFirstChild()->getFirstChild();
+         if (alreadyProfiledValues->isSet(value->getGlobalIndex()))
+            {
+            /* Found that second and third child were having more than one ref counts. 
+            TR_ASSERT_FATAL(node->getFirstChild()->getSecondChild()->getReferenceCount() == 1 &&
+               node->getFirstChild()->getThirdChild()->getReferenceCount() == 1,
+               "Second and Third Child of the value calls should be referenced only once");
+            */
+            dumpOptDetails(comp(), "%s Removing profiling treetop, node n%dn is already profiled before\n",
```
andrewcraik: please change already profiled before to already profiled

runtime/compiler/trj9/optimizer/JProfilingValue.cpp
```diff
@@ -182,6 +182,8 @@ TR_JProfilingValue::perform()
       }
 
    // Lower all existing calls
+   // TODO: Try an experimental pass here before lowering calls which will scan all the values we are proifling and tree top redundant
```
andrewcraik: Should this TODO still be here?

r30shah:
@andrewcraik  Here is the initial opt pass in JProfilingValue optimization before we add profiling trees lowering place holder jProfilingValue calls to find profiling sites that profiles already profiled values. Right now it only looks for the global index of nodes to find duplicate. We would like to expand it to cover more cases but wanted some comments from you. 
I had TR_ASSERT_FATAL to check ref count of the second and third child of place holder call. I found that in case where we are profiling same value at multiple places, will common out these nodes and increase the ref count.

andrewcraik:
copyright on trj9/optimizer/JProfilingValue.hpp  needs to be updated to 2017, 2018

runtime/compiler/trj9/optimizer/JProfilingValue.cpp
```diff
@@ -228,7 +228,7 @@ TR_JProfilingValue::removeRedundantProfilingValueCalls()
                node->getFirstChild()->getThirdChild()->getReferenceCount() == 1,
                "Second and Third Child of the value calls should be referenced only once");
             */
-            dumpOptDetails(comp(), "%s Removing profiling treetop, node n%dn is already profiled before\n",
+            dumpOptDetails(comp(), "%s Removing profiling treetop, node n%dn is already profiled\n",
```
andrewcraik: this should be a perform transformation since it isn't wrong to profile the same value multiple times.

r30shah:
Local Tests passes. @andrewcraik  Can you please review these changes?

andrewcraik:
Jenkins test sanity

andrewcraik:
Changes look reasonable waiting on sanity.

r30shah:
Rebased the changes. Going to run local sanity test again to verify. @andrewcraik  Can you start Jenkins Sanity tests for this PR?

andrewcraik:
Jenkins test sanity

runtime/compiler/trj9/optimizer/JProfilingValue.cpp
```diff
@@ -204,6 +206,46 @@ TR_JProfilingValue::perform()
...
+         {
+         TR::Node *value = node->getFirstChild()->getFirstChild();
+         if (alreadyProfiledValues->isSet(value->getGlobalIndex()))
+            {
+            /* Found that second and third child were having more than one ref counts. 
+            TR_ASSERT_FATAL(node->getFirstChild()->getSecondChild()->getReferenceCount() == 1 &&
+               node->getFirstChild()->getThirdChild()->getReferenceCount() == 1,
+               "Second and Third Child of the value calls should be referenced only once");
+            */
+            performTransformation(comp(), "%s Removing profiling treetop, node n%dn is already profiled\n",
```
andrewcraik: this needs to be in an if so that limiting transformations works since removing the tree is optional.

andrewcraik:
@r30shah in my final review I noticed the perform transformation is not in an if - please make sure the opt respects whether perform transformation says true (make the transformation) or false (skip the transformation) - thanks!

runtime/compiler/trj9/optimizer/JProfilingValue.cpp
```diff
@@ -204,6 +206,44 @@ TR_JProfilingValue::perform()
...
+      if (node->isProfilingCode() &&
+         node->getOpCodeValue() == TR::treetop &&
+         node->getFirstChild()->getOpCode().isCall() &&
+         (node->getFirstChild()->getSymbolReference()->getReferenceNumber() == TR_jProfile32BitValue ||
+         node->getFirstChild()->getSymbolReference()->getReferenceNumber() == TR_jProfile64BitValue) &&
+         node->getFirstChild()->getNumChildren() == 3)
+         {
+         TR::Node *value = node->getFirstChild()->getFirstChild();
+         
+         if (alreadyProfiledValues->isSet(value->getGlobalIndex()) &&
```
r30shah: @andrewcraik  Thanks for the review. Thanks for giving some background regarding to usage of `performTransformation`. I added it in if where I check the bitVector to see if value is already profiled and if performTransformation returns `true`.

r30shah:
Waiting for personal tests to finish. Will update as soon as they finish.

r30shah:
Local Tests passes. @andrewcraik  If you are OK with the chnages, this is good to merge.

andrewcraik:
Change looks good - Jenkins test sanity

PR #1443, opened 2018-03-14
Replace error prone "$(ECHO) -e" in makefile

renfeiw:
- We use "$(ECHO) -e" to print out new line, but "-e" confuses make.
- Use $(ECHO) $(Q)$(Q) to print new line instead.

[ci skip]


renfeiw:
reviewers: @smlambert  @llxia

llxia:

PR #1452, opened 2018-03-15
Reduce size of generated *.i files

keithc-ca:
Before this change (and the related OMR change), the total size of all *.i file was about 1.3GB; now the total is about 233MB. By avoiding writes of over 1GB to the file system, and avoiding later reads by getmacros of that same 1GB, build performance is improved.

### DIFF
--- a/runtime/makelib/targets.mk.ftl
+++ b/runtime/makelib/targets.mk.ftl
@@ -343,11 +343,14 @@ LIBCDEFS := $(wildcard /ztpf/$(UMA_ZTPF_ROOT)/base/lib/libCDEFSFORASM.so)
 </#if>
 
 # ddrgen preprocessor targets
+# We use sed in these scripts to retain only those lines that will be interesting
+# to omr/ddr/tools/getmacros; this reduces the total footprint of all *.i files by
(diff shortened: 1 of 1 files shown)


runtime/makelib/targets.mk.ftl
```diff
@@ -344,10 +344,10 @@ LIBCDEFS := $(wildcard /ztpf/$(UMA_ZTPF_ROOT)/base/lib/libCDEFSFORASM.so)
 
 # ddrgen preprocessor targets
 %.i : %.c
-	$(CC) $(CFLAGS) -E $< -o $@
+	$(CC) $(CFLAGS) -E $< | sed -n -e '/^@/p' > $@
```
DanHeidinga: Please add a comment in here describing what the sed command is doing and why it improves the size.  This helps all future readers of the code.

keithc-ca:
Rebased with added comment.

DanHeidinga:

DanHeidinga:
Jenkins compile

DanHeidinga:
That's a bit more testing than I meant - should have limited that build to compiles on either 8 or 9 but not both.

PR #1460, opened 2018-03-16
Bringup Java 11 b04 & raw builds

JasonFengJ9:
Bringup `Java 11 b04` & raw builds

1. Added call to `JCL` native `Java_java_lang_System_initProperties` to ensure `platform encoding` initialized for raw builds `OPENJ9-RAWBUILD`;
2. Modified `JVM_InitProperties` to just return incoming properties for `Java 11` `Java_java_lang_System_initProperties`, otherwise `assert(syspropsList != NULL)` will be triggered whenever `JVM_InitProperties` is called because `syspropsListRef` is defined within `J9JavaVM` but never initialized. I intended to leave the cleanup work to another PR.
3. Added call to `JCL` native `JNU_InitializeEncoding` which is going to exported by a `JCL` patch in `Java 11` extension repo.

Now `Java 11 b04` raw build `-version` output is as following:
```
openjdk version "11-internal" 2018-09-18
OpenJDK Runtime Environment (build 11-internal+0-adhoc.jason.ottj9build)
IBM J9 VM (build 2.9, JRE 11 Linux amd64-64 Compressed References 20180316_381412 (JIT enabled, AOT enabled)
OpenJ9   - 8a5cbe5
OMR      - c96856a
IBM      - 1d267ef)
```
A successful raw build depends on #1460, #1465 and #1466

Reviewer @DanHeidinga 

### DIFF
--- a/jcl/jpp_configuration.xml
+++ b/jcl/jpp_configuration.xml
@@ -243,6 +243,33 @@
 		<parameter name="msg:outputdir" value="java.base/share/classes/com/ibm/oti/util"/>
 		<parameter name="jxerules:outputdir" value="java/lang"/>
 	</configuration>
+
+	<configuration
+		  label="OPENJ9-RAWBUILD"
+		  outputpath="OPENJ9-RAWBUILD/src"
+		  flags="OpenJ9-RawBuild"
+		  dependencies="JAVA11"
+		  jdkcompliance="1.8">
+		<classpathentry kind="src" path="src/java.base/share/classes"/>
+		<classpathentry kind="src" path="src/java.management/share/classes"/>
+		<classpathentry kind="src" path="src/jdk.attach/share/classes"/>
+		<classpathentry kind="src" path="src/openj9.jvm/share/classes"/>
+		<classpathentry kind="src" path="src/jdk.management/share/classes"/>
+		<classpathentry kind="src" path="src/openj9.cuda/share/classes"/>
+		<classpathentry kind="src" path="src/openj9.dataaccess/share/classes"/>
+		<classpathentry kind="src" path="src/openj9.dtfj/share/classes"/>
+		<classpathentry kind="src" path="src/openj9.dtfjview/share/classes"/>
+		<classpathentry kind="src" path="src/openj9.gpu/share/classes"/>
+		<classpathentry kind="src" path="src/openj9.sharedclasses/share/classes"/>
+		<classpathentry kind="src" path="src/openj9.traceformat/share/classes"/>
+		<classpathentry kind="src" path="src/openj9.zosconditionhandling/share/classes"/>
+		<classpathentry kind="lib" path="/binaries/common/ibm/ibmjzos.jar"/>
+		<classpathentry kind="lib" path="/binaries/vm/third/rt-compressed.sunJava10.jar"/>
+		<source path="src"/>
+		<parameter name="macro:define" value="com.ibm.oti.vm.library.version=29"/>
+		<parameter name="msg:outputdir" value="java.base/share/classes/com/ibm/oti/util"/>
+		<parameter name="jxerules:outputdir" value="java/lang"/>
+	</configuration>
 	
 	<configuration
 		  label="PANAMA"
--- a/jcl/src/java.base/share/classes/java/lang/System.java
+++ b/jcl/src/java.base/share/classes/java/lang/System.java
@@ -351,12 +351,27 @@ private static final int PlatformEncoding = 1;
 private static final int FileEncoding = 2;
 private static final int OSEncoding = 3;
 
+/*[IF OpenJ9-RawBuild]*/
+	/* This is a JCL native required only by OpenJ9 raw build.
+	 * OpenJ9 raw build is a combination of OpenJ9 and OpenJDK binaries without JCL patches within extension repo.
+	 * Currently OpenJ9 depends on a JCL patch to initialize platform encoding which is not available to raw build.
+	 * A workaround for raw build is to invoke this JCL native which initializes platform encoding.
+	 * This workaround can be removed if that JCL patch is not required.
+	 */
+private static native Properties initProperties(Properties props);
+/*[ENDIF] OpenJ9-RawBuild */
+
 /**
  * If systemProperties is unset, then create a new one based on the values 
  * provided by the virtual machine.
  */
 @SuppressWarnings("nls")
 private static void ensureProperties() {
+/*[IF OpenJ9-RawBuild]*/
+	// invoke JCL native to initialize platform encoding
+	initProperties(new Properties());
+/*[ENDIF] OpenJ9-RawBuild */
+	
 	systemProperties = new Properties();
 
 	if (osEncoding != null)
--- a/runtime/j9vm/j7vmi.c
+++ b/runtime/j9vm/j7vmi.c
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2002, 2017 IBM Corp. and others
+ * Copyright (c) 2002, 2018 IBM Corp. and others
  *
  * This program and the accompanying materials are made available under
  * the terms of the Eclipse Public License 2.0 which accompanies this
@@ -1525,6 +1525,12 @@ JVM_IHashCode(JNIEnv *env, jobject obj)
 jobject JNICALL
 JVM_InitProperties(JNIEnv* env, jobject properties)
 {
+	/* This JVM method is invoked by JCL native Java_java_lang_System_initProperties
+	 * only for initialization of platform encoding.
+	 * This is only required by Java 11 raw builds.
+	 * This method is not invoked by other Java levels.
+	 */
+#if !defined(J9VM_JCL_SE11)
 	J9JavaVM* vm = ((J9VMThread*)env)->javaVM;
 	jobject syspropsList = (jobject)vm->syspropsListRef;
 	jclass propertiesClass = (*env)->GetObjectClass(env, properties);
@@ -1541,6 +1547,7 @@ JVM_InitProperties(JNIEnv* env, jobject properties)
 		jobject value = (*env)->GetObjectArrayElement(env, syspropsList, index++);
 		(*env)->CallObjectMethod(env, properties, setPropertyMID, key, value);
 	}
+#endif /* J9VM_JCL_SE11 */
 	return properties;
 }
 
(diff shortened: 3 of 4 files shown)


jcl/src/java.base/share/classes/java/lang/System.java
```diff
@@ -351,12 +351,22 @@ private static void arraycopy(Object[] A1, int offset1, Object[] A2, int offset2
...
+
 /**
  * If systemProperties is unset, then create a new one based on the values 
  * provided by the virtual machine.
  */
 @SuppressWarnings("nls")
 private static void ensureProperties() {
+/*[IF Java11]*/
+	// invoke JCL native to initialize platform encoding
+	initProperties(new Properties());
```
DanHeidinga: Doesn't the next line setting `systemProperties = new Properties();` throw away the ones that have been initialized?

runtime/j9vm/java9vmi.c
```diff
@@ -1777,3 +1777,11 @@ JVM_GetNanoTimeAdjustment(JNIEnv *env, jclass clazz, jlong offsetSeconds)
 	
 	return result;
 }
+
+#if J9VM_JCL_SE11
+void JNICALL
+JVM_BeforeHalt()
+{
+	/* To be implemented via https://github.com/eclipse/openj9/issues/1459 */
```
DanHeidinga: There should be an Assert or something here if this is called.

DanHeidinga:
@fjeremic Can you take a look at the String.java change and validate the string compression aspect?

jcl/src/java.base/share/classes/java/lang/System.java
DanHeidinga: Also looking at the PR for the JDK 11 extensions repo to understand how this fits together.

runtime/j9vm/java9vmi.c
JasonFengJ9: Actually this is being called indeed. I feel the implementation can be postponed in another issue to allow `Java 11 b04` build proceed first.

jcl/src/java.base/share/classes/java/lang/System.java
JasonFengJ9: The purpose of this invoking of JCL native is to initialize `platform encoding` only and the `Properties` set is `OpenJDK` specific and can't be used by `OpenJ9`.

DanHeidinga:
@JasonFengJ9 If you break this into a couple of PRs, we can get parts of it merged while sorting out the right way to handle `initializeProperities`.

The JVM_Halt function can be a separate PR and merged quickly.  The String function can also be a separate PR.

JasonFengJ9:
Created separated PRs: https://github.com/eclipse/openj9/pull/1465 and https://github.com/eclipse/openj9/pull/1466.

JasonFengJ9:
@DanHeidinga this PR has been updated to deal with `platform coding` initialization only. Please review again.

jcl/src/java.base/share/classes/java/lang/System.java
```diff
@@ -351,12 +351,22 @@ private static void arraycopy(Object[] A1, int offset1, Object[] A2, int offset2
 private static final int FileEncoding = 2;
 private static final int OSEncoding = 3;
 
+/*[IF OpenJ9-RawBuild]*/
+// JCL native
```
DanHeidinga: Can you describe here why this is only needed for the raw builds?  It will save someone from having to figure out if we can delete it in the future.

jcl/src/java.base/share/classes/java/lang/System.java
```diff
@@ -351,12 +351,22 @@ private static void arraycopy(Object[] A1, int offset1, Object[] A2, int offset2
...
+/*[ENDIF] OpenJ9-RawBuild */
+
 /**
  * If systemProperties is unset, then create a new one based on the values 
  * provided by the virtual machine.
  */
 @SuppressWarnings("nls")
 private static void ensureProperties() {
+/*[IF OpenJ9-RawBuild]*/
+	// invoke JCL native to initialize platform encoding
```
DanHeidinga: Same here - describe why is is required only for raw builds.

runtime/j9vm/j7vmi.c
```diff
@@ -1525,6 +1525,12 @@ JVM_IHashCode(JNIEnv *env, jobject obj)
 jobject JNICALL
 JVM_InitProperties(JNIEnv* env, jobject properties)
 {
+	/* This JVM method is invoked by JCL native Java_java_lang_System_initProperties
+	 * only for initialization of platform encoding.
+	 * This is only required by Java 11 raw builds.
+	 * This method is not invoked by other Java levels.
```
DanHeidinga: If this isn't invoked by other java levels, then why is most of the code only included in older levels?

runtime/jcl/common/system.c
```diff
@@ -385,6 +385,17 @@ jstring getEncoding(JNIEnv *env, jint encodingType)
 
 		case 1: 		/* platform encoding */
 			encoding = getPlatformFileEncoding(env, property, sizeof(property), encodingType);
+			{
+				UDATA handle = 0;
+				PORT_ACCESS_FROM_ENV(env);
+				if (0 == j9sl_open_shared_library("java", &handle, J9PORT_SLOPEN_DECORATE)) {
+					void (*nativeFuncAddrJNU)(JNIEnv *env, const char *str) = NULL;
+					if (0 == j9sl_lookup_name(handle, "JNU_InitializeEncoding", (UDATA*) &nativeFuncAddrJNU, "VLL")) {
```
DanHeidinga: `"VLL"` doesn't match the `JNIEnv *env, const char *str` declaration.  Is this correct?

jcl/src/java.base/share/classes/java/lang/System.java
JasonFengJ9: Added comments describing the reason to add/invoke this `JCL` native for `OpenJ9` raw build and the condition to remove this workaround.

jcl/src/java.base/share/classes/java/lang/System.java
JasonFengJ9: I think the comment above covers this case as well.

runtime/j9vm/j7vmi.c
JasonFengJ9: The code within this method can be removed, for example, `syspropsListRef` is defined within `J9JavaVM` but never initialized hence `assert(syspropsList != NULL)` will fire whenever `JVM_InitProperties` is called. I intended to leave the cleanup work to another PR as per this PR description (sorry for not making that obvious).

runtime/jcl/common/system.c
JasonFengJ9: As per `omrsl_lookup_name` API doc, `V` is for `void` (return type in this case), `L` is for `object / pointer (32 or 64, depending on platform)` (parameter types in this case). Some existing code use same signature such as following code within `stdinit.c`.
```
jstring (JNICALL *nativeFuncAddr)(JNIEnv *env, const char *str) = NULL;
j9sl_lookup_name(javaLibHandle->handle, "JNU_NewStringPlatform", (UDATA*)&nativeFuncAddr, "LLL");
```
i.e., `LLL` for `jstring (JNIEnv *env, const char *str)`.

Additional note: this signature parameter seems only used at `AIX J9OS_I5` platform (Window only uses it for calculation of padding length). It is not straightforward to do a quick verification.

JasonFengJ9:
Somehow I included a file not in my PR, will remove it.

JasonFengJ9:
Fixed. Please review at your convenience.

jcl/jpp_configuration.xml
```diff
@@ -243,6 +243,33 @@
 		<parameter name="msg:outputdir" value="java.base/share/classes/com/ibm/oti/util"/>
 		<parameter name="jxerules:outputdir" value="java/lang"/>
 	</configuration>
+
+	<configuration
+		  label="OPENJ9-RAWBUILD"
```
pshipton: This should include a version, such as OPENJ9-RAWBUILD11. Similarly the flag. Another naming suggestion: JAVA11-RAW.

jcl/src/java.base/share/classes/java/lang/System.java
```diff
@@ -351,12 +351,27 @@ private static void arraycopy(Object[] A1, int offset1, Object[] A2, int offset2
...
 private static final int OSEncoding = 3;
 
+/*[IF OpenJ9-RawBuild]*/
+	/* This is a JCL native required only by OpenJ9 raw build.
+	 * OpenJ9 raw build is a combination of OpenJ9 and OpenJDK binaries without JCL patches within extension repo.
+	 * Currently OpenJ9 depends on a JCL patch to initialize platform encoding which is not available to raw build.
+	 * A workaround for raw build is to invoke this JCL native which initializes platform encoding.
+	 * This workaround can be removed if that JCL patch is not required.
+	 */
+private static native Properties initProperties(Properties props);
```
pshipton: Please investigate calling this native even for non-raw builds. We'll have to compile java.base/share/native/javalib/System.c instead of stubbing it out. What problems does this cause?

jcl/jpp_configuration.xml
JasonFengJ9: This is really first time that a raw build need its own flag to workaround a `JCL` path not available to raw builds. I don't expect there are many more similar cases in the future. In additional raw builds intend to discover issues at earlier stage before normal builds can be performed. When a `JCL` level has a routine build, there is no need to perform raw builds for that level any more such as `Java 9`.
In worst scenario there is a need for another raw build `pConfig`, this generic raw build flag can be renamed first then add another one later.

jcl/src/java.base/share/classes/java/lang/System.java
JasonFengJ9: As per Dan's comments earlier, calling this native performs unnecessary work which is not desirable for non-raw builds. Instead we choose to provide a `JCL` patch to export a previously internal API `InitializeEncoding` (please refer `Java 11` extension repo PR). 
We hope to contribute this patch back to `OpenJDK` eventually.

runtime/jcl/common/system.c
DanHeidinga: Thanks for checking @JasonFengJ9.  I was expecting a "P" or similar character for pointers.

pshipton:
> Any further comments

Nope.

DanHeidinga:
Jenkins test sanity plinux

runtime/jcl/common/system.c
JasonFengJ9: As per `omrsl_lookup_name` API doc, `P` is for `pointer-width platform data. (in J9 terms an intptr_t)`. I followed the existing code and chose `L` in this case. Probably need run a test on an `AIX J9OS_I5` machine to determine if there is any difference between `L` and `P` for the sigatures.

JasonFengJ9: Apparently there is no `AIX i` series machines available in the farm.

Issue #1439, opened 2018-03-14
Latest JDK11 builds fail due to missing String.valueOfCodePoint(int) method


andrew-m-leonard:
OpenJDK11 has added a new method to String that our J9 impl needs to implement.
Build from repo https://github.ibm.com/runtimes/openj9-openjdk branch openj9-jdk to reproduce ...
```
=== Output from failing command(s) repeated here ===
* For target jdk_modules_java.base__the.java.base_batch:
/home/admin/mydocker/ccm/openj9-openjdk/src/java.base/share/classes/java/lang/Character.java:7597: error: cannot find symbol
        return String.valueOfCodePoint(codePoint);
                     ^
  symbol:   method valueOfCodePoint(int)
  location: class String
Note: Some input files use or override a deprecated API.
Note: Recompile with -Xlint:deprecation for details.
Note: Some input files use unchecked or unsafe operations.
Note: Recompile with -Xlint:unchecked for details.
1 error

* All command lines available in /home/admin/mydocker/ccm/openj9-openjdk/build/linux-x86_64-normal-server-release/make-support/failure-logs.
=== End of repeated output ===
```

DanHeidinga:
@JasonFengJ9 FYI

JasonFengJ9:
Looking at it.

PR #1466, opened 2018-03-16
Bringup Java 11 b04 - implement j.l.String.valueOfCodePoint()

JasonFengJ9:
Bringup `Java 11 b04` - implement `j.l.String.valueOfCodePoint()`

Implemented `j.l.String.valueOfCodePoint()`, manually verified that the output of `String.valueOfCodePoint(codePoint).codePointAt(0)` equals `codePoint` for all Unicode code point.

closes: #1439

Reviewer @DanHeidinga @fjeremic 

### DIFF
--- a/jcl/src/java.base/share/classes/java/lang/String.java
+++ b/jcl/src/java.base/share/classes/java/lang/String.java
@@ -5354,4 +5354,41 @@ written authorization of the copyright holder.
 		return value;
 	}
 /*[ENDIF] Java10 */	
+
+/*[IF Java11]*/
+	/**
+	 * Returns a string object containing the character (Unicode code point)
+	 * specified.
+	 * 
+	 * @param codePoint
+	 *            a Unicode code point.
+	 * @return a string containing the character (Unicode code point) supplied.
+	 * @throws IllegalArgumentException
+	 *             if the codePoint is not a valid Unicode code point.
+	 * @since 11
+	 */
+	static String valueOfCodePoint(int codePoint) {
+		String string;
+		if ((codePoint < Character.MIN_CODE_POINT) || (codePoint > Character.MAX_CODE_POINT)) {
+			/*[MSG "K0800", "Invalid Unicode code point - {0}"]*/
+			throw new IllegalArgumentException(com.ibm.oti.util.Msg.getString("K0800", Integer.toString(codePoint)));	//$NON-NLS-1$
+		} else if (codePoint <= 255) {
+			if (enableCompression) {
+				string = new String(compressedAsciiTable[codePoint], LATIN1);
+			} else {
+				string = new String(decompressedAsciiTable[codePoint], UTF16);
+			}
+		} else if (codePoint < Character.MIN_SUPPLEMENTARY_CODE_POINT) {
+			byte[] buffer = new byte[2];
+			helpers.putCharInArrayByIndex(buffer, 0, (char) codePoint);
+			string = new String(buffer, UTF16);
+		} else {
+			byte[] buffer = new byte[4];
+			helpers.putCharInArrayByIndex(buffer, 0, Character.highSurrogate(codePoint));
+			helpers.putCharInArrayByIndex(buffer, 1, Character.lowSurrogate(codePoint));
+			string = new String(buffer, UTF16);
+		}
+		return string;
+	}
+/*[ENDIF] Java11 */
 }
--- a/test/Java11andUp/build.xml
+++ b/test/Java11andUp/build.xml
@@ -1,0 +1,90 @@
+<?xml version="1.0"?>
+
+<!--
+  Copyright (c) 2018, 2018 IBM Corp. and others
+
+  This program and the accompanying materials are made available under
+  the terms of the Eclipse Public License 2.0 which accompanies this
+  distribution and is available at https://www.eclipse.org/legal/epl-2.0/
+  or the Apache License, Version 2.0 which accompanies this distribution and
+  is available at https://www.apache.org/licenses/LICENSE-2.0.
+
+  This Source Code may also be made available under the following
+  Secondary Licenses when the conditions for such availability set
+  forth in the Eclipse Public License, v. 2.0 are satisfied: GNU
+  General Public License, version 2 with the GNU Classpath
+  Exception [1] and GNU General Public License, version 2 with the
+  OpenJDK Assembly Exception [2].
+
+  [1] https://www.gnu.org/software/classpath/license.html
+  [2] http://openjdk.java.net/legal/assembly-exception.html
+
+  SPDX-License-Identifier: EPL-2.0 OR Apache-2.0 OR GPL-2.0 WITH Classpath-exception-2.0 OR LicenseRef-GPL-2.0 WITH Assembly-exception
+-->
+
+<project name="Java11AndUp" default="build" basedir=".">
+	<taskdef resource='net/sf/antcontrib/antlib.xml'/>
+	<description>
+		Tests for Java 11 and up
+	</description>
+	
+	<!-- set global properties for this build -->
+	<property name="DEST" value="${BUILD_ROOT}/Java11andUp" />
+	<property name="BUILDINFO_FSROOT" value="${BUILD_ROOT}" />
+		
+	<!--Properties for this particular build-->
+	<property name="src" location="./src"/>
+	<property name="build" location="./bin"/>
+	<property name="transformerListener" location="../Utils/src"/>
+
+	<target name="init">
+		<mkdir dir="${DEST}" />
+		<mkdir dir="${build}"/>
+	</target>
+	
+	<target name="compile" depends="init" description="Using ${JAVA_VERSION} java compile the source " >
+		<echo>Ant version is ${ant.version}</echo>
+		<echo>============COMPILER SETTINGS============</echo>
+		<echo>===fork:				yes</echo>
+		<echo>===executable:		${compiler.javac}</echo>
+		<echo>===debug:				on</echo>
+		<echo>===destdir:			${DEST}</echo>
+
+		<javac srcdir="${src}" destdir="${build}" debug="true" fork="true" executable="${compiler.javac}" includeAntRuntime="false" encoding="ISO-8859-1">
+			<src path="${src}"/>
+			<src path="${transformerListener}" />
+			<compilerarg line="--add-exports java.base/jdk.internal.reflect=ALL-UNNAMED --add-exports java.base/com.ibm.tools.attach.target=ALL-UNNAMED" />
+			<classpath>
+				<pathelement location="../TestConfig/lib/testng.jar"/>
+				<pathelement location="../TestConfig/lib/jcommander.jar"/>
+			</classpath>
+		</javac>
+	</target>
+
+	<target name="dist" depends="compile" description="generate the distribution" >
+		<mkdir dir="${DEST}"/>
+		<jar jarfile="${DEST}/GeneralTest.jar" filesonly="true">
+			<fileset dir="${build}"/>
+			<fileset dir="${src}/../" includes="*.properties,*.xml"/>
+		</jar>
+		<copy todir="${DEST}">
+			<fileset dir="${src}/../" includes="*.xml" />
+			<fileset dir="${src}/../" includes="*.mk" />
+		</copy>
+	</target>
+	
+	<target name="build" >
+		<if>
+			<not>
+				<matches string="${JAVA_VERSION}" pattern="^SE(8|9|10)0$$" />
+			</not>
+			<then>
+				<antcall target="clean" inheritall="true" />
+			</then>
+		</if>
+	</target>
+
+	<target name="clean" depends="dist" description="clean up" >
+		<delete dir="${build}"/>
+	</target>
+</project>
--- a/test/Java11andUp/src/org/openj9/test/java_lang/Test_String.java
+++ b/test/Java11andUp/src/org/openj9/test/java_lang/Test_String.java
@@ -51,12 +51,14 @@ public class Test_String {
 		// Verify that IllegalArgumentException is thrown for Character.MIN_CODE_POINT - 1.
 		try {
 			method.invoke(null, Character.MIN_CODE_POINT - 1);
+			Assert.fail("Failed to detect invalid Unicode point - Character.MIN_CODE_POINT - 1");
 		} catch (IllegalArgumentException | InvocationTargetException e) {
 			// Expected exception
 		}
 		// Verify that IllegalArgumentException is thrown for Character.MAX_CODE_POINT + 1.
 		try {
 			method.invoke(null, Character.MAX_CODE_POINT + 1);
+			Assert.fail("Failed to detect invalid Unicode point - Character.MAX_CODE_POINT + 1");
 		} catch (IllegalArgumentException | InvocationTargetException e) {
 			// Expected exception
 		}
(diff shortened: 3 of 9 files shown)


DanHeidinga:
Change looks reasonable to me.  @fjeremic Any comments on the string compression part?

pshipton:
Rather than manual verification, please add the automated test.

Did you manually test with string compression enabled?

JasonFengJ9:
Manually verified with both command line options `-XX:+CompactStrings` and `-XX:-CompactStrings`.

Will work on automated tests tomorrow.

jcl/src/java.base/share/classes/java/lang/String.java
```diff
@@ -5354,4 +5354,41 @@ public IntStream codePoints() {
...
+	static String valueOfCodePoint(int codePoint) {
+		String string;
+		if ((codePoint < Character.MIN_CODE_POINT) || (codePoint > Character.MAX_CODE_POINT)) {
+			/*[MSG "K0800", "Invalid Unicode code point"]*/
+			throw new IllegalArgumentException(com.ibm.oti.util.Msg.getString("K0800"));	//$NON-NLS-1$
+		} else if (codePoint <= 255) {
+			if (enableCompression) {
+				string = new String(compressedAsciiTable[codePoint], 0, 1, true);
+			} else {
+				string = new String(decompressedAsciiTable[codePoint], 0, 1, false);
```
fjeremic: Would it be possible to use:

https://github.com/JasonFengJ9/openj9/blob/c864cc795b78b3b1c573e8bc5684c7f9f413c6af/jcl/src/java.base/share/classes/java/lang/String.java#L402

Constructor here instead? The one currently being used will have to perform many of the checks again. Similarly for all other assignments to `string` in this method.

jcl/src/java.base/share/classes/java/lang/String.java
```diff
@@ -5354,4 +5354,41 @@ public IntStream codePoints() {
...
+	 * @return a string containing the character (Unicode code point) supplied.
+	 * @throws IllegalArgumentException
+	 *             if the codePoint is not a valid Unicode code point.
+	 * @since 11
+	 */
+	static String valueOfCodePoint(int codePoint) {
+		String string;
+		if ((codePoint < Character.MIN_CODE_POINT) || (codePoint > Character.MAX_CODE_POINT)) {
+			/*[MSG "K0800", "Invalid Unicode code point"]*/
+			throw new IllegalArgumentException(com.ibm.oti.util.Msg.getString("K0800"));	//$NON-NLS-1$
```
fjeremic: Would it be possible to put case as the last `else` case as it occurs least often. The JIT will have to do less work when block shuffling and if it every got it wrong we wouldn't pay the performance penalty.

DanHeidinga: Wouldn't the next check for `<= 255` need to expand then?  It would also need to validate codePoint > MIN_CODE_POINT.

jcl/src/java.base/share/classes/java/lang/String.java
JasonFengJ9: `String(byteArray, coder)` works as well, will update the PR accordingly.

jcl/src/java.base/share/classes/java/lang/String.java
JasonFengJ9: Will following checking help JIT performance?
```
if ((codePoint < Character.MIN_CODE_POINT)  {
  // error handling 
} else if (codePoint <= 255) {
} else if (codePoint < Character.MIN_SUPPLEMENTARY_CODE_POINT) {
} else if (codePoint <= Character.MAX_CODE_POINT) {
} else {
  // error handling
}
```
This saves the less often check `(codePoint > Character.MAX_CODE_POINT)` to the last.

JasonFengJ9:
@fjeremic the PR is ready for another review with following updates:
1.  Uses constructor `String(byteArray, coder)`;
2. Modified the order to save less often check `(codePoint > Character.MAX_CODE_POINT)` last;
3. Added tests. Note: this test suite is for `Java 11` hence will run when `Java 11` are introduced to the build jobs. Manually verified that it passes against `Java 11` raw builds with modified command line options.

@llxia please arrange a test team review.

Additional note: three project files failed Copyright check however this kind of files don't have copyright header either such as `Java[10|9|8]andUp`.

jcl/src/java.base/share/classes/java/lang/String.java
```diff
@@ -5354,4 +5354,45 @@ public IntStream codePoints() {
...
+	 * @throws IllegalArgumentException
+	 *             if the codePoint is not a valid Unicode code point.
+	 * @since 11
+	 */
+	static String valueOfCodePoint(int codePoint) {
+		String string;
+		// Save less often check (codePoint > Character.MAX_CODE_POINT) last
+		if (codePoint < Character.MIN_CODE_POINT) {
+			/*[MSG "K0800", "Invalid Unicode code point"]*/
+			throw new IllegalArgumentException(com.ibm.oti.util.Msg.getString("K0800"));	//$NON-NLS-1$
```
DanHeidinga: Can you include the string or the the code point in the exception?  It will make it much easier to debug one of these issues.

test/Java11andUp/src/org/openj9/test/java_lang/Test_String.java
```diff
@@ -0,0 +1,73 @@
...
+		try {
+			Method method = String.class.getDeclaredMethod("valueOfCodePoint", int.class);
+			method.setAccessible(true);
+			// Verify that for each valid Unicode code point i,
+			// i equals to String.valueOfCodePoint(i).codePointAt(0).
+			for (int i = Character.MIN_CODE_POINT; i <= Character.MAX_CODE_POINT; i++) {
+				try {
+					String str = (String) method.invoke(null, i);
+					Assert.assertEquals(i, str.codePointAt(0), "Unicode code point mismatch at " + i);
+				} catch (IllegalAccessException | IllegalArgumentException | InvocationTargetException e) {
```
DanHeidinga: Test cases shouldn't catch unexpected exceptions.  Let the test code throw them and the test harness will (usually) print the relevant stacktrace.

llxia:
@renfeiw , could you review the test change? Thanks.

jcl/src/java.base/share/classes/java/lang/String.java
```diff
@@ -5354,4 +5354,45 @@ public IntStream codePoints() {
...
+	 *            a Unicode code point.
+	 * @return a string containing the character (Unicode code point) supplied.
+	 * @throws IllegalArgumentException
+	 *             if the codePoint is not a valid Unicode code point.
+	 * @since 11
+	 */
+	static String valueOfCodePoint(int codePoint) {
+		String string;
+		// Save less often check (codePoint > Character.MAX_CODE_POINT) last
+		if (codePoint < Character.MIN_CODE_POINT) {
```
fjeremic: I would vouch to say this is also an uncommon path and should be moved with the `else` path. Do we agree?

DanHeidinga: We need to check that before checking `} else if (codePoint <= 255) {` or we need to move it into the check for `<= 255`.  Either way, its got to be first.

fjeremic: Sure, but merging the two checks has the benefit of smaller footprint and less code to jump over to get to the `0 <= codePoint <= 255` case.

renfeiw:
To enable SE110, we need to make an update in testKitGen. (https://github.com/eclipse/openj9/pull/1495)

jcl/src/java.base/share/classes/java/lang/String.java
DanHeidinga: I strongly dislike the duplication of the second form.  If we can't optimize the first form reasonably then we're going to have problems with most user code.

JasonFengJ9: It seems you are suggesting code like following:
```
if (codePoint >= Character.MIN_CODE_POINT && codePoint <= 255) {
  ...
} else if (codePoint >= Character.MIN_CODE_POINT && codePoint < Character.MIN_SUPPLEMENTARY_CODE_POINT) {
  ...
} else if (codePoint >= Character.MIN_CODE_POINT && codePoint <= Character.MAX_CODE_POINT) {
  ...
} else {
  error handling
}
```
This introduces extra code and confusion (later reader need lots of explanation about the rational of this code logic). I don't think any performance gain here worth the potential issue later. 
Now I feel this code should go back to original structure, i.e.,
```
if ((codePoint < Character.MIN_CODE_POINT) || (codePoint > Character.MAX_CODE_POINT)) {
  error handling
} else if (codePoint <= 255) {
  ...
} else if (codePoint < Character.MIN_SUPPLEMENTARY_CODE_POINT) {
  ...
} else {
  ...
}
```
It is clear and easy to understand. Can we modify it to improve performance later if this becomes a hot spot?

fjeremic: Yes, I hadn't thought about the extra checks introduced in my original comment. I'm fine either keeping as is or reverting back to your latest suggestion @JasonFengJ9. Apologies for the churn!

jcl/src/java.base/share/classes/java/lang/String.java
JasonFengJ9: Yup, now an exception looks like:
```
FAILED: testValueOfCodePoint
java.lang.AssertionError: Unicode code point mismatch at 0 expected [1] but found [0]
	at org.testng.Assert.fail(Assert.java:96)
	at org.testng.Assert.failNotEquals(Assert.java:776)
	at org.testng.Assert.assertEqualsImpl(Assert.java:137)
	at org.testng.Assert.assertEquals(Assert.java:118)
	at org.testng.Assert.assertEquals(Assert.java:652)
	at org.openj9.test.java_lang.Test_String.testValueOfCodePoint(Test_String.java:49)
```

test/Java11andUp/src/org/openj9/test/java_lang/Test_String.java
JasonFengJ9: Fixed, removed catching of unexpected exceptions.

jcl/src/java.base/share/classes/java/lang/String.java
JasonFengJ9: No problem, updated PR as latest suggestion above.

test/Java11andUp/build.xml
```diff
@@ -0,0 +1,88 @@
...
+		</jar>
+		<copy todir="${DEST}">
+			<fileset dir="${src}/../" includes="*.xml" />
+			<fileset dir="${src}/../" includes="*.mk" />
+		</copy>
+	</target>
+	
+	<target name="build" >
+		<if>
+			<equals arg1="${JAVA_VERSION}" arg2="SE110"/>
```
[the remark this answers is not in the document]
JasonFengJ9: Agreed, changed the condition to `not ^SE(8|9|10)0$$`.

JasonFengJ9:
The PR has been updated to address review comments.
Though `Java 11` hasn't been built and run in any test job yet, I manually verified that `Java11andUp` test project can be built, run and detect problem as expected.
Please review again.

jcl/src/java.base/share/classes/java/lang/String.java
```diff
@@ -5354,4 +5354,42 @@ public IntStream codePoints() {
...
+	 * @param codePoint
+	 *            a Unicode code point.
+	 * @return a string containing the character (Unicode code point) supplied.
+	 * @throws IllegalArgumentException
+	 *             if the codePoint is not a valid Unicode code point.
+	 * @since 11
+	 */
+	static String valueOfCodePoint(int codePoint) {
+		String string;
+		// Save less often check (codePoint > Character.MAX_CODE_POINT) last
```
DanHeidinga: Comment is no longer accurate

test/Java11andUp/src/org/openj9/test/java_lang/Test_String.java
```diff
@@ -0,0 +1,60 @@
...
+		// i equals to String.valueOfCodePoint(i).codePointAt(0).
+		for (int i = Character.MIN_CODE_POINT; i <= Character.MAX_CODE_POINT; i++) {
+			String str = (String) method.invoke(null, i);
+			Assert.assertEquals(i, str.codePointAt(0), "Unicode code point mismatch at " + i);
+		}
+		// Verify that IllegalArgumentException is thrown for
+		// Character.MIN_CODE_POINT - 1 & Character.MAX_CODE_POINT + 1
+		try {
+			method.invoke(null, Character.MIN_CODE_POINT - 1);
+			method.invoke(null, Character.MAX_CODE_POINT + 1);
```
DanHeidinga: The second method.invoke will never run as the first show throw.  We need this to be two tests or to have it written like:
```
try {
method.invoke(null, Character.MIN_CODE_POINT - 1);
fail()
} catch (IllegalArgumentException | InvocationTargetException e) {} 
try {
method.invoke(null, Character.MIN_CODE_POINT - 2);
fail()
} catch (IllegalArgumentException | InvocationTargetException e) {} 
```

jcl/src/java.base/share/classes/java/lang/String.java
JasonFengJ9: The comment in question has been removed.

test/Java11andUp/src/org/openj9/test/java_lang/Test_String.java
JasonFengJ9: Updated the test with suggested format above.
Please have another look.

test/Java11andUp/src/org/openj9/test/java_lang/Test_String.java
```diff
@@ -0,0 +1,64 @@
...
+		method.setAccessible(true);
+		// Verify that for each valid Unicode code point i,
+		// i equals to String.valueOfCodePoint(i).codePointAt(0).
+		for (int i = Character.MIN_CODE_POINT; i <= Character.MAX_CODE_POINT; i++) {
+			String str = (String) method.invoke(null, i);
+			Assert.assertEquals(i, str.codePointAt(0), "Unicode code point mismatch at " + i);
+		}
+		// Verify that IllegalArgumentException is thrown for Character.MIN_CODE_POINT - 1.
+		try {
+			method.invoke(null, Character.MIN_CODE_POINT - 1);
```
DanHeidinga: This needs an explicit `Assert.fail()` call or the method could succeed and pass the test when we know it should throw the exception.

test/Java11andUp/src/org/openj9/test/java_lang/Test_String.java
```diff
@@ -0,0 +1,64 @@
...
+		} catch (IllegalArgumentException | InvocationTargetException e) {
+			// Expected exception
+		}
+		// Verify that IllegalArgumentException is thrown for Character.MAX_CODE_POINT + 1.
+		try {
+			method.invoke(null, Character.MAX_CODE_POINT + 1);
```
DanHeidinga: This needs an explicit `Assert.fail()` call or the method could succeed and pass the test when we know it should throw the exception.

test/Java11andUp/src/org/openj9/test/java_lang/Test_String.java
JasonFengJ9: Agreed, added `Assert.fail(message)`.

test/Java11andUp/src/org/openj9/test/java_lang/Test_String.java
JasonFengJ9: Fixed as well. Please have another look, thanks.

DanHeidinga:
Jenkins test sanity plinux

DanHeidinga:
Running one builds worth of sanity to ensure that this is correctly ignored by older Java levels.

PR #1472, opened 2018-03-19
Add/correct advice for INCLUDE & LIB settings

keithc-ca:
Add '...\DIA SDK\include' to INCLUDE as required by OMR when DDR is enabled so dia2.h can be found.


pshipton:
@sxa555 fyi

PR #1474, opened 2018-03-19
Correct path to *.pdb files for shared libraries

keithc-ca:
The DDR build flow uses the generated `*_pdb` macros (e.g. j9vm_pdb) to find the debug information on Windows. This makes those macros reflect reality (where the files are actually created), including the stream (29) where appropriate.


pshipton:
This only affects Windows, which doesn't have any PR testing yet.

PR #1476, opened 2018-03-19
Check return value from query_auxv before using it

DanHeidinga:
There's a crash on PPC due to assuming query_auxv never fails.
The return value needs to be checked for -1 or 0 before using it
to map to something.

fixes: #1476

### DIFF
--- a/runtime/port/unix/j9sysinfo.c
+++ b/runtime/port/unix/j9sysinfo.c
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 1991, 2017 IBM Corp. and others
+ * Copyright (c) 1991, 2018 IBM Corp. and others
  *
  * This program and the accompanying materials are made available under
  * the terms of the Eclipse Public License 2.0 which accompanies this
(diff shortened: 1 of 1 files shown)


gacholio:
jenkins test sanity plinux

DanHeidinga:
For debugging auxv issue, I found this useful:
```
 $ LD_SHOW_AUXV=1 sleep 1
```
as it prints the AUXV output. See http://man7.org/linux/man-pages/man3/getauxval.3.html for details

DanHeidinga:
Updated the commit as the original just check the same variable over and over.

Behaviour is now:

We can't check the AT_HWCAP queries as they may actually return
0 or -1 as the field being looked up is a bit field. Assuming
they succeed is the best we can do given the API.

Allow the AT_BASE_PLATFORM call to fail and default to UNKNOWN
if the AT_PLATFORM has already succeeded.

DanHeidinga:
jenkins test sanity plinux

runtime/port/unix/j9sysinfo.c
```diff
@@ -479,20 +479,34 @@ getLinuxPPCDescription(struct J9PortLibrary *portLibrary, J9ProcessorDesc *desc)
...
 		desc->physicalProcessor = PROCESSOR_PPC_UNKNOWN;
+		desc->features[0] = 0;
+		desc->features[1] = 0;
 		return -1;
 	}
 
 	/* Linux PPC processor */
 	platform = (char *) query_auxv(AT_PLATFORM);
+	if ((NULL == platform) || (((char *) -1) == platform)) {
+		goto _error;
```
zl-wang: This goto is grotesque.  Why not initialize outside the info and avoid it all together?

DanHeidinga: :)  goto is used this way throughout the VM code base to simplify error handling.  

If you prefer the code be written another way, please open a PR.

PR #1481, opened 2018-03-19
OpenJ9 Java 10 set java.vendor.version

### LINKED ISSUE #842: OpenJ9 should consider setting java.vendor.version for Java 10
As per http://openjdk.java.net/jeps/322 OpenJ9 should consider setting java.vendor.version for Java 10 and later. TDB what it gets set to.

JasonFengJ9:
`OpenJ9 Java 10` set `java.vendor.version`

This new system property for `Java 10` and onward is set to an empty string for now.

Did a `Java 10` build, and got following system property output:
```
java.vendor.version =
```

closes #842 

Reviewer @pshipton 
FYI @DanHeidinga  

### DIFF
--- a/runtime/vm/vmprops.c
+++ b/runtime/vm/vmprops.c
@@ -772,6 +772,13 @@ initializeSystemProperties(J9JavaVM * vm)
 		goto fail;
 	}
 
+	if (J2SE_V10 <= j2seVersion) {
+		rc = addSystemProperty(vm, "java.vendor.version", "", 0);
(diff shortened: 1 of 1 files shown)


pshipton:
Is this the same as an OpenJDK builds, or should it just be undefined?

JasonFengJ9:
`OpenJDK` build doesn't have this property since it is source code only project and vendor neutral.

pshipton:
I don't know any reason to define it as an empty String. If we're not going to set a value, it should be undefined similarly to OpenJDK builds.

PR #1486, opened 2018-03-20
Correct name of JIT shared library

keithc-ca:
This corrects the situation that leads to a warning from the linker:
```
j9jit29.exp : warning LNK4070: /OUT:J9JIT28.dll directive in .EXP differs from output filename '...\j9jit29.dll'; ignoring directive
```


keithc-ca:
Copyright notice added.

ymanton:
Merged. Thanks for the fix.

PR #1490, opened 2018-03-21
Remove deprecated arraycopy statistics

0dvictor:
_copyStats has been long deprecated and its related code
has been removed a while ago. Removing last the leftover
piece.


andrewcraik:
Jenkins test sanity

andrewcraik:
@0dvictor just wanting to confirm that no other platforms are using this code correct?

0dvictor:
The code is guarded by `TR_HOST_X86` and hence it should only be used by X86.

andrewcraik:
failed sanity is a known issue unrelated to this change - everything else is good so merging.

PR #1494, opened 2018-03-21
Enable Windows stack protection (/GS flag)

theresa-m:
Performance summary:
- Throughput: no regression
- Compilation: 2% more CPU spent
- Startup: no regression
- Footprint: no regression

### DIFF
--- a/runtime/makelib/targets.mk.ftl
+++ b/runtime/makelib/targets.mk.ftl
@@ -531,6 +531,14 @@ cnathelp$(UMA_DOT_O):cnathelp.cpp
 	$(CXX) $(CXXFLAGS) -mstackrealign -c $<
 </#if>
 
+# special handling for cnathelp.cpp to disable buffer security check in windows for JIT helpers
+<#if uma.spec.flags.build_newCompiler.enabled && uma.spec.type.windows>	
+cnathelp$(UMA_DOT_O):cnathelp.cpp
+	$(CXX) $(CXXFLAGS) /GS- -c $<
+SharedService$(UMA_DOT_O):SharedService.c
+	$(CC) $(CFLAGS) /GS- -c $<
+</#if>
+
 <#if uma.spec.processor.ppc>
 ifndef USE_PPC_GCC
 # special handling BytecodeInterpreter.cpp and DebugBytecodeInterpreter.cpp
--- a/runtime/makelib/targets.mk.win.inc.ftl
+++ b/runtime/makelib/targets.mk.win.inc.ftl
@@ -1,5 +1,5 @@
 <#-- 
-	Copyright (c) 1998, 2017 IBM Corp. and others
+	Copyright (c) 1998, 2018 IBM Corp. and others
 	
 	This program and the accompanying materials are made available under
 	the terms of the Eclipse Public License 2.0 which accompanies this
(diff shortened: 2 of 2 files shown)


runtime/codert_vm/cnathelp.cpp
```diff
@@ -69,7 +69,7 @@ buildJITResolveFrameWithPC(J9VMThread *currentThread, UDATA flags, UDATA parmCou
 	currentThread->jitException = NULL;
 	resolveFrame->specialFrameFlags = flags;
 #if defined(J9SW_JIT_HELPERS_PASS_PARAMETERS_ON_STACK)
-	resolveFrame->parmCount = parmCount;
+	resolveFrame->parmCount = 0;
```
fjeremic: Is this change intentional? If so then the `#if` and `#else` paths are identical and can be merged.

DanHeidinga: Why this change?  Is it required by the /GS flag?  

Changing this indicates that the resolveFrame doesn't have any parameters, which means we won't attempt to walk them / update them / etc.

`J9SW_JIT_HELPERS_PASS_PARAMETERS_ON_STACK` is like a x86-32 thing.  Does /GS apply to both 32/64?

DanHeidinga:
@theresa-m you did a bunch of startup / throughput / footprint analysis on this change.  Can you summarize that here?

andrewcraik:
Does this change affect all components of J9 or just the VM/GC? If it includes the JIT measurements of the compile-time impact (amount of time it takes the JIT to compile code) would also be good to see.

runtime/codert_vm/cnathelp.cpp
theresa-m: The change was attempting to address this concern:
>Some of the JIT helpers (c code) run on the java stack now in c interpreter.  We'd need to keep the flag disabled for those helpers.  The file to disable the option for is cnathelp.cpp

/GS applies to all platforms according to its documentation.

theresa-m:
Performance summary (so far):
- Throughput: no regression
- Compilation: 2% more CPU spent
- Startup: no regression
- Footprint: no regression

@andrewcraik I assume that compile time number is what you are referring to? @mpirvu was okay with this number.

runtime/codert_vm/cnathelp.cpp
DanHeidinga: That should result in makefile changes to disable the /GS flag for those files.

andrewcraik:
@theresa-m yes FYI @vijaysun-omr due to the CPU increase

runtime/codert_vm/cnathelp.cpp
theresa-m: ah.. that makes more sense. thanks!

vijaysun-omr:
Yes I am okay with the CPU increase reported.

runtime/codert_vm/cnathelp.cpp
theresa-m: Made changes to correct this in the last push.

charliegracie:
I see a similar PR agains the OMR project.  Do these changes have to be co-ordinated in anyway?

theresa-m:
@charliegracie it won't break anything to have one in and not the other

pshipton:
@DanHeidinga any further concerns?

DanHeidinga:
Merging without a build as we don't have PR builds on windows yet.

PR #1496, opened 2018-03-21
Use full sha in the pipeline scripts to avoid ambiguousness

vsebe:
Return 40 characters sha instead of short sha to avoid any ambiguousness
when checking out from a sha (a short sha could match multiple shas,
thus throw an error).

[ci skip]


vsebe:
@jdekonin please review, thank you!

vsebe:
@pshipton or @DanHeidinga please review, thank you!

AdamBrousseau:
Has 7 actually been proven to be a problem with any of the repos we use (reproducible test case)? Seems like a big jump to go from 7 to 40. Also, if we want the whole SHA we could have just done `cut -f1`
The downfall is the build descriptions now aren't as pretty
image

vsebe:
There were 9 characters.
The build description can be adjusted by truncating the shas to a shorter lenght.

PR #1501, opened 2018-03-21
[Nestmates] Updates java.lang.Class API for new nestmates spec draft (class loading) depends on #1224

taliamccormick:
An earlier nestmates spec draft loaded a class's nest host at the same time that the class was loaded. A later spec draft puts off nest host loading until necessary for an access check. The existing java.lang.Class native methods for nestmates assumes that the nest host has already been loaded; this PR allows the native methods for nestmates to load/verify the nest host.

This PR is a follow-up to https://github.com/eclipse/openj9/pull/1224, which moves nest host loading to the visibility check.

### DIFF
--- a/runtime/jcl/common/java_lang_Class.cpp
+++ b/runtime/jcl/common/java_lang_Class.cpp
@@ -60,6 +60,12 @@ static UDATA isPrivilegedFrameIterator(J9VMThread * currentThread, J9StackWalkSt
 static UDATA isPrivilegedFrameIteratorGetAccSnapshot(J9VMThread * currentThread, J9StackWalkState * walkState);
 static UDATA frameIteratorGetAccSnapshotHelper(J9VMThread * currentThread, J9StackWalkState * walkState, j9object_t acc, j9object_t perm);
 static j9object_t storePDobjectsHelper(J9VMThread* vmThread, J9Class* arrayClass, J9StackWalkState* walkState, j9object_t contextObject, U_32 arraySize, UDATA framesWalked, I_32 startPos, BOOLEAN dupCallerPD);
+#if defined(J9VM_OPT_VALHALLA_NESTMATES)
+static void setNestmatesVerificationError(J9VMThread *vmThread, const char *nlsTemplate, UDATA exceptionNumber, J9UTF8 *nestMemberName, J9UTF8 *nestHostName);
+static J9Class *loadAndCheckNestHost(J9VMThread *vmThread, J9Class *clazz, BOOLEAN canThrow);
+jobject JNICALL Java_java_lang_Class_getNestHostImpl(JNIEnv *env, jobject recv);
+jobject JNICALL Java_java_lang_Class_getNestMembersImpl(JNIEnv *env, jobject recv);
+#endif /* defined(J9VM_OPT_VALHALLA_NESTMATES) */
 
 jobject JNICALL 
 Java_java_lang_Class_getDeclaredAnnotationsData(JNIEnv *env, jobject jlClass)
@@ -1787,21 +1793,14 @@ storePDobjectsHelper(J9VMThread* vmThread, J9Class* arrayClass, J9StackWalkState
 
 #if defined(J9VM_OPT_VALHALLA_NESTMATES)
 static void
-setNestmatesIncompatibleClassChangeError(J9VMThread *vmThread, J9Class *nestMember, J9Class *nestHost)
+setNestmatesVerificationError(J9VMThread *vmThread, const char *nlsTemplate, UDATA exceptionNumber, J9UTF8 *nestMemberName, J9UTF8 *nestHostName)
 {
 	J9JavaVM *vm = vmThread->javaVM;
 	J9InternalVMFunctions *vmFuncs = vm->internalVMFunctions;
+	char *msg = NULL;
+
 	PORT_ACCESS_FROM_VMC(vmThread);
 
-	J9UTF8 *nestMemberName = J9ROMCLASS_CLASSNAME(nestMember->romClass);
-	J9UTF8 *nestHostName = J9ROMCLASS_CLASSNAME(nestHost->romClass);
-
-	const char *nlsTemplate = j9nls_lookup_message(
-			J9NLS_DO_NOT_PRINT_MESSAGE_TAG | J9NLS_DO_NOT_APPEND_NEWLINE,
-			J9NLS_JCL_NEST_MEMBER_CLAIMS_DIFFERENT_NEST_HOST,
-			"Nest member %2$.*1$s in %4$.*3$s declares a different nest host of %6$.*5$s");
-
-	char *msg = NULL;
 	if (NULL != nlsTemplate) {
 		UDATA msgLen = j9str_printf(PORTLIB, NULL, 0, nlsTemplate,
 				J9UTF8_LENGTH(nestMemberName), J9UTF8_DATA(nestMemberName),
@@ -1814,8 +1813,125 @@ setNestmatesIncompatibleClassChangeError(J9VMThread *vmThread, J9Class *nestMemb
 				J9UTF8_LENGTH(nestMemberName), J9UTF8_DATA(nestMemberName));
 	}
 
-	vmFuncs->setCurrentExceptionUTF(vmThread, J9VMCONSTANTPOOL_JAVALANGINCOMPATIBLECLASSCHANGEERROR, msg);
+	vmFuncs->setCurrentExceptionUTF(vmThread, exceptionNumber, msg);
 	j9mem_free_memory(msg);
+}
+
+/**
+ * Loads, sets, verifies, & returns the claimed nest host of a class
+ *
+ * If the claimed nest host of a class is successfully loaded and verified,
+ * then the nest host field of the class is set and the nest host is returned.
+ * Otherwise, the nest host field is not set and the class itself is returned.
+ * If canThrow is true and loading/verifying the nest host fails, then an
+ * exception will be set.
+ *
+ * @param vmthread[in] the current vm thread
+ * @param clazz[in,out] class who's nest host is being acquired
+ * @param canThrow[in] true if a verification error/exception can be set & thrown
+ *
+ * @returns the nest host claimed by the class or the class itself
+ */
+static J9Class *
+loadAndCheckNestHost(J9VMThread *vmThread, J9Class *clazz, BOOLEAN canThrow) {
+	J9Class *nestHost = clazz->nestHost;
+
+	if (NULL == nestHost) {
+		BOOLEAN verified = FALSE;
+		J9ROMClass *romClass = clazz->romClass;
+		J9UTF8 *nestHostName = J9ROMCLASS_NESTHOSTNAME(romClass);
+
+		/* If no nest host is named, class is own nest host */
+		if (NULL == nestHostName) {
+			nestHost = clazz;
+			verified = TRUE;
+		} else {
+			J9InternalVMFunctions *vmFuncs = vmThread->javaVM->internalVMFunctions;
+			J9UTF8 *className = J9ROMCLASS_CLASSNAME(romClass);
+			UDATA classLoadingFlags = 0;
+
+			PORT_ACCESS_FROM_VMC(vmThread);
+
+			if (canThrow) {
+				classLoadingFlags = J9_FINDCLASS_FLAG_THROW_ON_FAIL;
+			} else {
+				classLoadingFlags = J9_FINDCLASS_FLAG_EXISTING_ONLY;
+			}
+			nestHost = vmFuncs->internalFindClassUTF8(vmThread, J9UTF8_DATA(nestHostName), J9UTF8_LENGTH(nestHostName), clazz->classLoader, classLoadingFlags);
+
+			/* If this can throw, then we must check that the nest host is successfully
+			 * loaded by the same classloader in the same package & verifies the nest
+			 * member. Upon successful verification of the nest host class, the nest
+			 * host field is set.
+			 * If this can not throw, then verification occurs in order to determine
+			 * whether a nest host or 'this' is returned.
+			 */
+			if (NULL == nestHost) {
+				if (canThrow) {
+					const char *nlsTemplate = j9nls_lookup_message(
+							J9NLS_DO_NOT_PRINT_MESSAGE_TAG | J9NLS_DO_NOT_APPEND_NEWLINE,
+							J9NLS_JCL_NESTMATES_CLASS_FAILED_TO_LOAD,
+							NULL);
+					setNestmatesVerificationError(vmThread,
+							nlsTemplate,
+							J9VMCONSTANTPOOL_JAVALANGINCOMPATIBLECLASSCHANGEERROR,
+							className,
+							nestHostName);
+				}
+			} else if (clazz->packageID != nestHost->packageID) {
+				if (canThrow) {
+					const char *nlsTemplate = j9nls_lookup_message(
+							J9NLS_DO_NOT_PRINT_MESSAGE_TAG | J9NLS_DO_NOT_APPEND_NEWLINE,
+							J9NLS_JCL_NEST_HOST_HAS_DIFFERENT_PACKAGE,
+							NULL);
+					setNestmatesVerificationError(vmThread,
+							nlsTemplate,
+							J9VMCONSTANTPOOL_JAVALANGINCOMPATIBLECLASSCHANGEERROR,
+							className,
+							nestHostName);
+				}
+			} else {
+				/* The nest host must have a nestmembers attribute that claims this class. */
+				J9SRP *nestMembers = J9ROMCLASS_NESTMEMBERS(nestHost->romClass);
+				U_16 nestMemberCount = nestHost->romClass->nestMemberCount;
+				U_16 i = 0;
+				J9UTF8 *nestMemberName = NULL;
+
+				for (i = 0; i < nestMemberCount; i++) {
+					nestMemberName = NNSRP_GET(nestMembers[i], J9UTF8*);
+					if (J9UTF8_EQUALS(className, nestMemberName)) {
+						verified = TRUE;
+					}
+				}
+
+				if (verified) {
+					clazz->nestHost = nestHost;
+				} else if (canThrow) {
+					const char *nlsTemplate = j9nls_lookup_message(
+							J9NLS_DO_NOT_PRINT_MESSAGE_TAG | J9NLS_DO_NOT_APPEND_NEWLINE,
+							J9NLS_JCL_NEST_MEMBER_NOT_CLAIMED_BY_NEST_HOST,
+							NULL);
+					setNestmatesVerificationError(vmThread,
+							nlsTemplate,
+							J9VMCONSTANTPOOL_JAVALANGINCOMPATIBLECLASSCHANGEERROR,
+							className,
+							nestHostName);
+				}
+			}
+		}
+
+		/* If a nest host is successfully verified, then setting the nest host
+		 * field prevents the same verification from recurring. If it can not
+		 * be loaded and verified, then the class itself is returned.
+		 */
+		if (verified) {
+			clazz->nestHost = nestHost;
+		} else {
+			nestHost = clazz;
+		}
+	}
+
+	return nestHost;
 }
 
 jobject JNICALL
@@ -1827,6 +1943,20 @@ Java_java_lang_Class_getNestHostImpl(JNIEnv *env, jobject recv)
 
 	J9Class *clazz = J9VM_J9CLASS_FROM_HEAPCLASS(currentThread, J9_JNI_UNWRAP_REFERENCE(recv));
 	J9Class *nestHost = clazz->nestHost;
+
+	/* Nest host field is set post-verification, during an access check or during
+	 * a call to relevant java.lang.Class API method via the loadAndCheckNestHost
+	 * function. If loadAndCheckNestHost fails to return a valid nest host, then
+	 * getNestHostImpl returns this.
+	 */
+	if (NULL == nestHost) {
+		/* If loadAndCheckNestHost can not successfully load & verify a class's
+		 * nest host, then it will return the class - as required by the
+		 * getNestHost spec.
+		 */
+		nestHost = loadAndCheckNestHost(currentThread, clazz, FALSE);
+	}
+
 	j9object_t resultObject = J9VM_J9CLASS_TO_HEAPCLASS(nestHost);
 	jobject result = vmFuncs->j9jni_createLocalRef(env, resultObject);
 
@@ -1849,16 +1979,36 @@ Java_java_lang_Class_getNestMembersImpl(JNIEnv *env, jobject recv)
 	j9object_t resultObject = NULL;
 	jobject result = NULL;
 
-	vmFuncs->internalEnterVMFromJNI(currentThread);
+	/* The Class<?> class will be used for result object size */
+	J9Class *jlClass = NULL;
+	J9Class *arrayClass = NULL;
 
 	J9Class *clazz = J9VM_J9CLASS_FROM_HEAPCLASS(currentThread, J9_JNI_UNWRAP_REFERENCE(recv));
 	J9Class *nestHost = clazz->nestHost;
-	J9ROMClass *romHostClass = nestHost->romClass;
-	U_16 nestMemberCount = romHostClass->nestMemberCount;
+	J9ROMClass *romHostClass = NULL;
+	U_16 nestMemberCount = 0;
+
+	PORT_ACCESS_FROM_VMC(currentThread);
+
+	vmFuncs->internalEnterVMFromJNI(currentThread);
+
+	if (NULL == nestHost) {
+		/* If loadAndCheckNestHost can not successfully load & verify a class's
+		 * nest host and canThrow is true, then it will return the class and
+		 * set an exception.
+		 */
+		nestHost = loadAndCheckNestHost(currentThread, clazz, TRUE);
+	}
+	if (NULL != currentThread->currentException) {
+		goto _done;
+	}
+
+	romHostClass = nestHost->romClass;
+	nestMemberCount = romHostClass->nestMemberCount;
 
 	/*  Grab the Class<?> class for result object size */
-	J9Class *jlClass = J9VMJAVALANGCLASS(vm);
-	J9Class *arrayClass = fetchArrayClass(currentThread, jlClass);
+	jlClass = J9VMJAVALANGCLASS(vm);
+	arrayClass = fetchArrayClass(currentThread, jlClass);
 	if (NULL != currentThread->currentException) {
 		goto _done;
 	}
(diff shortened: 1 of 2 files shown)


taliamccormick:
@tajila

runtime/jcl/common/java_lang_Class.cpp
```diff
@@ -1827,6 +1970,20 @@ Java_java_lang_Class_getNestHostImpl(JNIEnv *env, jobject recv)
...
 	J9Class *nestHost = clazz->nestHost;
+
+	/*
+	 * Nest host field is set post-verification, during an access check or during
+	 * a call to relevant java.lang.Class API method via the loadAndCheckNestHost
+	 * function. If loadAndCheckNestHost fails to return a valid nest host, then
+	 * getNestHostImpl returns this.
+	 */
+	if (NULL == nestHost) {
+		nestHost = loadAndCheckNestHost(currentThread, clazz, FALSE);
```
tajila: does OpenJDK load classes when getNestHost\getNestMembers is called?
If it does, what happens if there is an error verifying nest membership?

tajila:
@taliamccormick can you put `depends on: #1224` in the commit message so the committers know the other PR must be merged first

runtime/jcl/common/java_lang_Class.cpp
taliamccormick: I ran `java -verbose:class` against an older OpenJDK nestmates build and against this for the valid test case:
```
public class nestmates_api_tests_delayed_loading {
	public class NestMember {}
	public class NestHost {}
	
	public static void main (String[] args) {
		System.out.println("Nest: " + Arrays.toString(new nestmates_api_tests_delayed_loading().new NestHost().getClass().getClasses()));
	}
}
```

OpenJDK seems to load classes; output produced at the end was:
```
[0.237s][info][class,load] nestmates_api_tests_delayed_loading source: file:/team/tdmccormick/projects/20170509-NestMates/
[0.237s][info][class,load] java.lang.NamedPackage source: jrt:/java.base
[0.238s][info][class,load] java.lang.PublicMethods$MethodList source: jrt:/java.base
[0.238s][info][class,load] java.lang.PublicMethods$Key source: jrt:/java.base
[0.238s][info][class,load] java.lang.Void source: jrt:/java.base
[0.240s][info][class,load] nestmates_api_tests_delayed_loading$NestHost source: file:/team/tdmccormick/projects/20170509-NestMates/
[0.242s][info][class,load] nestmates_api_tests_delayed_loading$NestMember source: file:/team/tdmccormick/projects/20170509-NestMates/
Nest: [class nestmates_api_tests_delayed_loading$NestHost, class nestmates_api_tests_delayed_loading$NestMember, class nestmates_api_tests_delayed_loading]
[0.243s][info][class,load] java.lang.Shutdown source: jrt:/java.base
[0.244s][info][class,load] java.lang.Shutdown$Lock source: jrt:/java.base
```

This was similar; output produced at the end was:
```
class load: nestmates_api_tests_delayed_loading from: file:/team/tdmccormick/projects/20170509-NestMates/
class load: jdk/internal/reflect/UnsafeFieldAccessorFactory from: jrt:/java.base
class load: jdk/internal/reflect/FieldAccessor from: jrt:/java.base
class load: jdk/internal/reflect/FieldAccessorImpl from: jrt:/java.base
class load: jdk/internal/reflect/UnsafeFieldAccessorImpl from: jrt:/java.base
class load: jdk/internal/reflect/UnsafeObjectFieldAccessorImpl from: jrt:/java.base
class load: java/lang/Class$ReflectCache from: nestmates_api_patch_3.jar
class load: nestmates_api_tests_delayed_loading$NestHost from: file:/team/tdmccormick/projects/20170509-NestMates/
class load: nestmates_api_tests_delayed_loading$NestMember from: file:/team/tdmccormick/projects/20170509-NestMates/
Nest: [class nestmates_api_tests_delayed_loading$NestHost, class nestmates_api_tests_delayed_loading$NestMember, class nestmates_api_tests_delayed_loading]
class load: java/util/IdentityHashMap$KeySet from: jrt:/java.base
class load: java/util/IdentityHashMap$IdentityHashMapIterator from: jrt:/java.base
class load: java/util/IdentityHashMap$KeyIterator from: jrt:/java.base
class load: com/ibm/tools/attach/target/AttachHandler$1 from: jrt:/java.base
class load: java/util/Vector$Itr from: jrt:/java.base
class load: java/lang/Thread$State from: jrt:/java.base
```
(the `nestmates_api_patch_3.jar` is my patch for testing)

taliamccormick: **QUESTION: Should I change our error message so we don't exactly match OpenJDK?**

Running the same test case but invalid nestmates attributes (one of the listed nest members does not claim the nest host) raises an error for both OpenJDK and this.

OpenJDK:
```
/team/tdmccormick/jvms/Nestmates_Build_15.01.2018/jdk/bin/java  nestmates_api_tests_delayed_loading
Exception in thread "main" java.lang.IncompatibleClassChangeError: Nest member nestmates_api_tests_delayed_loading in nestmates_api_tests_delayed_loading$NestHost declares a different nest host of nestmates_api_tests_delayed_loading
        at java.base/java.lang.Class.getNestMembers0(Native Method)
        at java.base/java.lang.Class.getNestMembers(Class.java:3922)
        at nestmates_api_tests_delayed_loading.main(Unknown Source)
```

This:
```

Exception in thread "main" java.lang.IncompatibleClassChangeError: Nest member nestmates_api_tests_delayed_loading in nestmates_api_tests_delayed_loading$NestHost declares a different nest host of nestmates_api_tests_delayed_loading
        at java.lang.Class.getNestMembersImpl(java.base@9-internal/Native Method)
        at java.lang.Class.getNestMembers(java.base@9-internal/Class.java:4126)
        at nestmates_api_tests_delayed_loading.main(Unknown Source)
```

taliamccormick: I get two copies of the same IncompatibleClassChangeError (ICCE) from calling `getNestMembers` and none from `getNestHost`. This behaviour is expected. `getNestMembers` fails because the nest host iterates over its list of nest members and checks that they reciprocally verify eachother; it raises an ICCE when it finds that a nest member does not verify the nest host. `getNestHost` passes because the nest member class considers itself to be its own nest host and thus does no further checking.

--

The test case ran:
```
for (int x = 0; x < 2; x++) {
	try {
		System.out.println("Nest host: " + nestMemberClass.getNestHost());
	} catch (IncompatibleClassChangeError e) {
		System.out.println(e + " caused by " + e.getCause());
	}
}		
for (int x = 0; x < 2; x++) {
	try {
		System.out.println("Nest: " + Arrays.toString(nestHostClass.getNestMembers()));
	} catch (IncompatibleClassChangeError e) {
		System.out.println(e + " caused by " + e.getCause());
	}
}
```
and it produced
```
Nest host: class pkg.NestMember
Nest host: class pkg.NestMember
java.lang.IncompatibleClassChangeError: Nest member pkg/NestMember in pkg/NestMember$NestHost declares a different nest host of pkg/NestMember caused by null
java.lang.IncompatibleClassChangeError: Nest member pkg/NestMember in pkg/NestMember$NestHost declares a different nest host of pkg/NestMember caused by null
```

taliamccormick: Triggering an IncompatibleClassChangeError from `getNestHost` (ie. the nest member class claims a nest host which does not reciprocally verify it) is not possible. If any verification error occurs, then the class itself is returned. 

--

The test case ran:
```
for (int x = 0; x < 2; x++) {
	try {
		System.out.println("Nest host: " + nestMemberClass.getNestHost());
	} catch (IncompatibleClassChangeError e) {
		System.out.println(e + " caused by " + e.getCause());
	}
}
```
and it produced
```
Nest host: class pkg.NestMember
Nest host: class pkg.NestMember
```
despite claiming a different class as its nest host. This lines up with the current spec draft: 
>  Returns:
    the nest host of this class, or this if we cannot obtain a valid nest host 

(http://cr.openjdk.java.net/%7Edholmes/8010319/specs/java.lang/java/lang/Class.html)

and also matches the behaviour of a nestmates openjdk build.

runtime/jcl/common/java_lang_Class.cpp
```diff
@@ -1818,6 +1824,144 @@ setNestmatesIncompatibleClassChangeError(J9VMThread *vmThread, J9Class *nestMemb
...
+			UDATA classLoadingFlags = 0;
+			if (TRUE == canThrow) {
+				classLoadingFlags = J9_FINDCLASS_FLAG_THROW_ON_FAIL;
+			} else {
+				classLoadingFlags = J9_FINDCLASS_FLAG_EXISTING_ONLY;
+			}
+			nestHost = vmFuncs->internalFindClassUTF8(vmThread, J9UTF8_DATA(nestHostName), J9UTF8_LENGTH(nestHostName), clazz->classLoader, classLoadingFlags);
+
+			/*
+			 * If this can throw, then we must check that the nest host is successfully
```
babsingh: fix comment:
`/* If this can throw, then we must check that the nest host is successfully`

runtime/jcl/common/java_lang_Class.cpp
```diff
@@ -1888,8 +2061,22 @@ Java_java_lang_Class_getNestMembersImpl(JNIEnv *env, jobject recv)
...
+				 * a NoClassDefFoundError
+				 */
 				goto _done;
-			} else if (nestMember->nestHost != nestHost) {
+			} else if (NULL == nestMember->nestHost) {
+				/* If loadAndCheckNestHost fails to set the nest member's nest
+				 * host (due to loading or verification error), then it sets an
+				 * IncompatibleClassChangeError
+				 */
+				loadAndCheckNestHost(currentThread, nestMember, TRUE);
```
babsingh: use `loadAndCheckNestHost` consistently.

runtime/nls/j9cl/j9jcl.nls
```diff
@@ -436,3 +436,35 @@ J9NLS_JCL_NEST_MEMBER_CLAIMS_DIFFERENT_NEST_HOST.system_action=The JVM will thro
...
 # END NON-TRANSLATABLE
 
+# First argument is class name length
+# Second argument is class name
+# Third argument is nest host name length
+# Fourth argument is nest host name
+J9NLS_JCL_NESTMATES_CLASS_FAILED_TO_LOAD=Nest member %2$.*1$s must be able to load to be loaded by nest host %4$.*3$s
+# START NON-TRANSLATABLE
+J9NLS_JCL_NESTMATES_CLASS_FAILED_TO_LOAD.explanation=Nest member class and nest host class were loaded with different class loaders.
+J9NLS_JCL_NESTMATES_CLASS_FAILED_TO_LOAD.system_action=The JVM will throw a java/lang/VerifyError
```
babsingh: aren't we throwing an `ICCE` instead of a `VerifyError`? `ICCE` - Incompatible Class Change Error.

runtime/nls/j9cl/j9jcl.nls
```diff
@@ -436,3 +436,35 @@ J9NLS_JCL_NEST_MEMBER_CLAIMS_DIFFERENT_NEST_HOST.system_action=The JVM will thro
...
+# END NON-TRANSLATABLE
+
+# First argument is class name length
+# Second argument is class name
+# Third argument is nest host name length
+# Fourth argument is nest host name
+J9NLS_JCL_NEST_HOST_HAS_DIFFERENT_PACKAGE=Nest member class %2$.*1$s must be in the same package as nest host class %4$.*3$s
+# START NON-TRANSLATABLE
+J9NLS_JCL_NEST_HOST_HAS_DIFFERENT_PACKAGE.explanation=Nest member class and nest host class were loaded in different packages
+J9NLS_JCL_NEST_HOST_HAS_DIFFERENT_PACKAGE.system_action=The JVM will throw a java/lang/VerifyError
```
babsingh: aren't we throwing an `ICCE` instead of a `VerifyError`?

runtime/nls/j9cl/j9jcl.nls
```diff
@@ -436,3 +436,35 @@ J9NLS_JCL_NEST_MEMBER_CLAIMS_DIFFERENT_NEST_HOST.system_action=The JVM will thro
...
+# END NON-TRANSLATABLE
+
+# First argument is class name length
+# Second argument is class name
+# Third argument is nest host name length
+# Fourth argument is nest host name
+J9NLS_JCL_NEST_MEMBER_NOT_CLAIMED_BY_NEST_HOST=Class %2$.*1$s must be claimed by its nest host %4$.*3$s
+# START NON-TRANSLATABLE
+J9NLS_JCL_NEST_MEMBER_NOT_CLAIMED_BY_NEST_HOST.explanation=Class was not claimed by its nest host class.
+J9NLS_JCL_NEST_MEMBER_NOT_CLAIMED_BY_NEST_HOST.system_action=The JVM will throw a java/lang/VerifyError
```
babsingh: aren't we throwing an `ICCE` instead of a `VerifyError`?

runtime/jcl/common/java_lang_Class.cpp
```diff
@@ -1818,6 +1824,144 @@ setNestmatesIncompatibleClassChangeError(J9VMThread *vmThread, J9Class *nestMemb
...
+ * @returns the nest host claimed by the class or the class itself
+ */
+static J9Class *
+loadAndCheckNestHost(J9VMThread *vmThread, J9Class *clazz, BOOLEAN canThrow) {
+	J9Class *nestHost = clazz->nestHost;
+
+	if (NULL == nestHost) {
+		J9JavaVM *vm = vmThread->javaVM;
+		J9InternalVMFunctions *vmFuncs = vm->internalVMFunctions;
+		PORT_ACCESS_FROM_VMC(vmThread);
```
babsingh: this won't be needed if `setICCEForNestMates` equivalent is added.

runtime/nls/j9cl/j9jcl.nls
taliamccormick: Woops, thanks

runtime/jcl/common/java_lang_Class.cpp
taliamccormick: What is the inconsistency here? I use `loadAndCheckNestHost` on each nest member class in order to ensure that their nest host is loaded, prior to checking it against `this`'s nest host. The returned value does not matter in this case

--

This current setup of `loadAndCheckNestHost` is intentional. It allows the same `loadAndCheckNestHost` function to be used for `getNestHost`, `getNestMembers`, and `isNestMateOf` where `getNestMembers` is the only java.lang.Class nestmates method that can set an ICCE (see: http://cr.openjdk.java.net/~dholmes/8010319/specs/java.lang/java/lang/Class.html)

In particular, it allows for the case where:
- Class NestHostA claims nest member list 'NestMemberC'
- Class NestHostB claims nest member list 'NestMemberC'
- Class NestMemberC claims nest host class 'NestHostB'

Then we have the following possible situations
- Calling `NestHostA.isNestMateOf(NestMemberC)` should return false without throwing.
- Calling `NestMemberC .getNestHost()` should return `NestHostB` without throwing.
- Calling `NestHostA.getNestMembers()` should throw ICCE
regardless of whether `isNestMateOf` or `getNestHost` have been called.

Thus, the setup for `loadAndCheckNestHost` will return a valid nest host if possible and return the passed class if no valid nest host is set. The return value is used by `getNestHost` and `isNestmateOf` but is ignored by `getNestMembers`. The set nestmate field is ignored by `getNestHost` and `isNestmateOf` and used by `getNestMembers`.

runtime/jcl/common/java_lang_Class.cpp
```diff
@@ -1849,16 +2007,31 @@ Java_java_lang_Class_getNestMembersImpl(JNIEnv *env, jobject recv)
...
 
 	J9Class *clazz = J9VM_J9CLASS_FROM_HEAPCLASS(currentThread, J9_JNI_UNWRAP_REFERENCE(recv));
 	J9Class *nestHost = clazz->nestHost;
-	J9ROMClass *romHostClass = nestHost->romClass;
-	U_16 nestMemberCount = romHostClass->nestMemberCount;
+	J9ROMClass *romHostClass = NULL;
+	U_16 nestMemberCount = 0;
+
+	if (NULL == nestHost) {
+		loadAndCheckNestHost(currentThread, clazz, TRUE);
```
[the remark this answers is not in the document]
taliamccormick: Have now updated
```
if (NULL == nestHost) {
	loadAndCheckNestHost(currentThread, clazz, TRUE);
	nestHost = clazz->nestHost;
}
if (NULL == nestHost) {
	goto _done;
}
```
to
```
if (NULL == nestHost) {
	nestHost = loadAndCheckNestHost(currentThread, clazz, TRUE);
}
if (NULL != currentThread->currentException) {
	goto _done;
}
```

runtime/jcl/common/java_lang_Class.cpp
```diff
@@ -1818,6 +1824,144 @@ setNestmatesIncompatibleClassChangeError(J9VMThread *vmThread, J9Class *nestMemb
...
+		/*
+		 * If a nest host is successfully verified, then setting the nest host
+		 * field prevents the same verification from recurring.
+		 * If a nest host can not be successfully verified, then a LinkageError
+		 * is set (if permitted) and null is returned as the nest host.
+		 */
+		if (TRUE == verified) {
+			clazz->nestHost = nestHost;
+		} else {
+			if (TRUE == canThrow) {
```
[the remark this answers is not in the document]
taliamccormick: See discussion/comments above re setICCEForNestMates

runtime/jcl/common/java_lang_Class.cpp
```diff
@@ -1818,6 +1824,144 @@ setNestmatesIncompatibleClassChangeError(J9VMThread *vmThread, J9Class *nestMemb
...
+		J9JavaVM *vm = vmThread->javaVM;
+		J9InternalVMFunctions *vmFuncs = vm->internalVMFunctions;
+		PORT_ACCESS_FROM_VMC(vmThread);
+
+		J9ROMClass *romClass = clazz->romClass;
+		J9UTF8 *className = J9ROMCLASS_CLASSNAME(clazz->romClass);
+		J9UTF8 *nestHostName = J9ROMCLASS_NESTHOSTNAME(romClass);
+		BOOLEAN verified = FALSE;
+
+		char *nlsMsg = NULL;
```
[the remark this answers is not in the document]
taliamccormick: Declaring `nlsMsg` early decreases code duplication & makes the flow of code more clear.

There are multiple error cases (nest host fails to load; nest host in different runtime package; nest host fails to verify) which set the same exception. Declaring the same `nlsMsg` variable and using it for all cases allows the ICCE to only be set in one place. It also makes it more clear at the return point that an exception has been set.
```
if (nest host verification success) {
    set and return valid nest host
} else (nest host verification failure) {
    if (exception should be set) set exception
    return this
}
```

runtime/jcl/common/java_lang_Class.cpp
```diff
@@ -1818,6 +1824,144 @@ setNestmatesIncompatibleClassChangeError(J9VMThread *vmThread, J9Class *nestMemb
...
+
+			/*
+			 * If this can throw, then we must check that the nest host is successfully
+			 * loaded by the same classloader in the same package & verifies the nest
+			 * member. Upon successful verification of the nest host class, the nest
+			 * host field is set.
+			 * If this can not throw, then verification occurs in order to determine
+			 * whether a nest host or 'this' is returned.
+			 */
+			if (NULL == nestHost) {
```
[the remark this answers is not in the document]
taliamccormick: The NLS codes (eg. `J9NLS_JCL_NESTMATES_CLASS_FAILED_TO_LOAD`) can not be easily passed around like that, due to the way that NLS error messages are looked up.

Since `j9nls_lookup_message` returns a `constant char*`, we can not declare nlsTemplate at the start of the function and set its value within an 'if' clause.

--

The current handling of NLS messages is:
- NLS .h header files are generated via `make -f buildtools.mk nls`
- The header files created contain two values for each message - a MODULE value and an ID value. They also define a macro for each message that turns `msg` into `msg_MODULE, msg_ID`
eg. looking in nls/j9jclnls.h gives
```
#define J9NLS_JCL_NEST_MEMBER_CLAIMS_DIFFERENT_NEST_HOST__MODULE 0x4a39434c
#define J9NLS_JCL_NEST_MEMBER_CLAIMS_DIFFERENT_NEST_HOST__ID 42
#define J9NLS_JCL_NEST_MEMBER_CLAIMS_DIFFERENT_NEST_HOST J9NLS_JCL_NEST_MEMBER_CLAIMS_DIFFERENT_NEST_HOST__MODULE, J9NLS_JCL_NEST_MEMBER_CLAIMS_DIFFERENT_NEST_HOST__ID
```
- When compiling the VM, the pre-processor replaces each `msg` value with a pair of values
- `j9nls_lookup_message` is a wrapper for `nls_lookup_message` (from OMR). The `j9nls_lookup_message` macro takes three arguments. The omr version takes five - the port library and four values that it receives from `j9nls_lookup_message` (since the macro is expanded prior to passing it).
- Both the module and id values must be passed to `j9nls_lookup_message` in order to succeed. This causes a compilation failure, since `j9nls_lookup_message` receives too many arguments. Passing only one causes `nls_lookup_message` to fail because it does not receive enough arguments.

This can be worked around by passing both the `msg_MODULE` and `msg_ID` values into the `nls_lookup_message` or `omrnls_lookup_message` OMR macros, but I have been told to avoid doing so - that it's preferable to use the j9 versions.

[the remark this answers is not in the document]
babsingh: i have updated my previous comment. added an example. you don't need to access `moduleName` and `messageNum` explicitly.

why do you need to do this?
`Since j9nls_lookup_message returns a constant char*, we can not declare nlsTemplate at the start of the function and set its value within an 'if' clause.`

runtime/jcl/common/java_lang_Class.cpp
babsingh: > What is the inconsistency here?
inconsistency is related to the usage of `loadAndCheckNestHost`. at some places, the `return value` is used. at others, `clazz->nestHost` is used. there needs to be a comment in order to clarify why the `return value` is ignored.

runtime/jcl/common/java_lang_Class.cpp
```diff
@@ -1814,10 +1813,129 @@ setNestmatesIncompatibleClassChangeError(J9VMThread *vmThread, J9Class *nestMemb
...
+
+	if (NULL == nestHost) {
+		BOOLEAN verified = FALSE;
+
+		/* If no nest host is named, class is own nest host */
+		if (NULL == nestHostName) {
+			nestHost = clazz;
+			verified = TRUE;
+		} else {
+			J9JavaVM *vm = vmThread->javaVM;
```
babsingh: `J9JavaVM *vm` is only used once below. we don't need this local variable.

runtime/nls/j9cl/j9jcl.nls
```diff
@@ -436,3 +436,35 @@ J9NLS_JCL_NEST_MEMBER_CLAIMS_DIFFERENT_NEST_HOST.system_action=The JVM will thro
...
 # END NON-TRANSLATABLE
 
+# First argument is class name length
+# Second argument is class name
+# Third argument is nest host name length
+# Fourth argument is nest host name
+J9NLS_JCL_NESTMATES_CLASS_FAILED_TO_LOAD=Nest member %2$.*1$s must be able to load to be loaded by nest host %4$.*3$s
+# START NON-TRANSLATABLE
+J9NLS_JCL_NESTMATES_CLASS_FAILED_TO_LOAD.explanation=Nest member class and nest host class were loaded with different class loaders.
+J9NLS_JCL_NESTMATES_CLASS_FAILED_TO_LOAD.system_action=The JVM will throw a ICCE
```
babsingh: `ICCE` -> `IncompatibleClassChangeError` (newbies mayn't understand ICCE abbreviation)

runtime/jcl/common/java_lang_Class.cpp
```diff
@@ -1827,6 +1945,24 @@ Java_java_lang_Class_getNestHostImpl(JNIEnv *env, jobject recv)
...
+	 * function. If loadAndCheckNestHost fails to return a valid nest host, then
+	 * getNestHostImpl returns this.
+	 */
+	if (NULL == nestHost) {
+		/* If loadAndCheckNestHost can not successfully load & verify a class's
+		 * nest host, then it will return the class - as required by the
+		 * getNestHost spec.
+		 */
+		nestHost = loadAndCheckNestHost(currentThread, clazz, FALSE);
+		if (NULL == nestHost) {
```
babsingh: if `loadAndCheckNestHost` is going to return `clazz` on unsuccessful load and verify, then we don't need this `if` block.

taliamccormick:
Made PR changes per Babneet's review:

- Remove definition of local variable J9JavaVM *vm
- Expand ICCE to IncompatibleClassChangeError in error message
- Remove unnecessary null check

runtime/jcl/common/java_lang_Class.cpp
```diff
@@ -1827,6 +1944,22 @@ Java_java_lang_Class_getNestHostImpl(JNIEnv *env, jobject recv)
...
+	 * function. If loadAndCheckNestHost fails to return a valid nest host, then
+	 * getNestHostImpl returns this.
+	 */
+	if (NULL == nestHost) {
+		/* If loadAndCheckNestHost can not successfully load & verify a class's
+		 * nest host, then it will return the class - as required by the
+		 * getNestHost spec.
+		 */
+		nestHost = loadAndCheckNestHost(currentThread, clazz, FALSE);
+		nestHost = clazz;
```
babsingh: this was supposed to removed with the `if` block.

runtime/jcl/common/java_lang_Class.cpp
```diff
@@ -1827,6 +1944,22 @@ Java_java_lang_Class_getNestHostImpl(JNIEnv *env, jobject recv)
 
 	J9Class *clazz = J9VM_J9CLASS_FROM_HEAPCLASS(currentThread, J9_JNI_UNWRAP_REFERENCE(recv));
 	J9Class *nestHost = clazz->nestHost;
+
+	/*
+	 * Nest host field is set post-verification, during an access check or during
```
babsingh: fix comment: `/* Nest host field ...`

runtime/jcl/common/java_lang_Class.cpp
```diff
@@ -1814,10 +1813,128 @@ setNestmatesIncompatibleClassChangeError(J9VMThread *vmThread, J9Class *nestMemb
...
+							nlsTemplate,
+							J9VMCONSTANTPOOL_JAVALANGINCOMPATIBLECLASSCHANGEERROR,
+							className,
+							nestHostName);
+				}
+			}
+		}
+
+		/*
+		 * If a nest host is successfully verified, then setting the nest host
```
babsingh: Fix comment: `/* If a nest host ...`

taliamccormick:
@babsingh Updated per review; squashed; and retested.

babsingh:
Senior Reviewer: @DanHeidinga

taliamccormick:
> Do you have a link to the spec for the methods being added?

No external java.lang.Class API changes are added/created by this PR. Earlier spec drafts loaded a class's nest host during class loading - therefore, by the time a nestmate-related java.lang.Class method was called, each class's nest host would be loaded already. A newer spec draft postpones the nest host loading until an access check requires it. This PR therefore allows the native java.lang.Class methods to load nest host when necessary.

The PR that initiated this change is here: https://github.com/eclipse/openj9/pull/1224
And the corresponding spec draft is here: http://cr.openjdk.java.net/~dlsmith/nestmates.html

Some of the exception handling has also been changed. A second matching PR to this ones handles `SecurityExceptions` which may be thrown by the java.lang.Class nestmates API methods and lists more up-to-date details on exception handling.
That PR is here: https://github.com/eclipse/openj9/pull/1531
And the corresponding spec draft is here: http://cr.openjdk.java.net/%7Edholmes/8010319/specs/java.lang/java/lang/Class.html

taliamccormick:
Added PR changes per review; will squash prior to merge

- Adds a javadoc style description to the function
setNestmatesVerificationError
- Breaks out of verification loop loop early once class is verified
- Removes duplicate setting of nestHost field
- Moves Class<?> unwrapping to occur after acquiring VM access
- Moves variable declaration to limit live range (permitted in cpp
files)

Note that the variable declarations can not be moved very far because `goto`s are used so early on - compilation will fail because variables have to be nested inside another scope or defined before they are used. eg.
```
../common/java_lang_Class.cpp: In function '_jobject* Java_java_lang_Class_getNestMembersImpl(JNIEnv*, _jobject*)':
../common/java_lang_Class.cpp:2080: error: jump to label '_done'
../common/java_lang_Class.cpp:2008: error:   from here
../common/java_lang_Class.cpp:2012: error:   crosses initialization of 'U_16 nestMemberCount'
../common/java_lang_Class.cpp:2011: error:   crosses initialization of 'J9ROMClass* romHostClass'
make[1]: *** [java_lang_Class.o] Error 1
make[1]: Leaving directory `/team/tdmccormick/projects/20170509-NestMates/VM_Nestmates_MoveLinking_2/lib/amd64/compressedrefs/jcl/cl_se9_before_b165'
make: *** [jclse9_] Error 2
```

runtime/jcl/common/java_lang_Class.cpp
```diff
@@ -1846,19 +1983,36 @@ Java_java_lang_Class_getNestMembersImpl(JNIEnv *env, jobject recv)
...
+			goto _done;
+		}
+	}
+	romHostClass = nestHost->romClass;
+	nestMemberCount = romHostClass->nestMemberCount;
 
 	/*  Grab the Class<?> class for result object size */
-	J9Class *jlClass = J9VMJAVALANGCLASS(vm);
-	J9Class *arrayClass = fetchArrayClass(currentThread, jlClass);
+	jlClass = J9VMJAVALANGCLASS(vm);
```
tajila: this can be `J9VMJAVALANGCLASS_OR_NULL`

tajila:
@DanHeidinga This PR can be closed

PR #1505, opened 2018-03-22
Fix a PPC TLE issue, a missing NULLCHK causing a crash in a transaction

klangman:
When replacing a monitor-enter with a transaction we did not check if
the monitor-enter is under a Null-Check (i.e. a synchronized block) and
therefore we did not build the transaction with a preceding Null-Check.
This could allow for a SEGV during the transaction if the variable
being locked is null.
On PPC a SEGV within a transaction will result in a core dump rather
then a transaction rollback. For this reason we must have an explicit
Null-Check before the transaction starts. On platforms were a SEGV in
a transaction would result in a rollback, it still might be better to
Null-Check before the transaction to avoid a series of SEGV's during
the transaction attempts. At the moment TLE is only enabled on Z, so
this change is being done in case TLE is re-enabled on PPC in the
future.

### DIFF
--- a/runtime/compiler/trj9/optimizer/MonitorElimination.cpp
+++ b/runtime/compiler/trj9/optimizer/MonitorElimination.cpp
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2000, 2017 IBM Corp. and others
+ * Copyright (c) 2000, 2018 IBM Corp. and others
  *
  * This program and the accompanying materials are made available under
  * the terms of the Eclipse Public License 2.0 which accompanies this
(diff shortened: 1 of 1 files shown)


klangman:
@vijaysun-omr Did you say you were going to review this PR for me?

vijaysun-omr:
Looks okay to me.

runtime/compiler/trj9/optimizer/MonitorElimination.cpp
```diff
@@ -1464,6 +1464,26 @@ void TR::MonitorElimination::transformMonitorsIntoTMRegions()
       TR::CFG *cfg = comp()->getFlowGraph();
       cfg->setStructure(NULL);        // voiding out structure
 
+      if( monitor->getTreeTopNode()->getOpCodeValue() == TR::NULLCHK)
```
fjeremic: Parenthesis formatting fixup needed here and at the below `TR_ASSERT` so we're consistent with the rest of the codebase.

runtime/compiler/trj9/optimizer/MonitorElimination.cpp
```diff
@@ -1464,6 +1464,26 @@ void TR::MonitorElimination::transformMonitorsIntoTMRegions()
...
+         // Rip out the NULLCHK from the monent tree and put it on it's own TT above the tstart
+         debugTrace(tracer(),"monent TT node is NULLCHK!\n\tMoving NULLCHK to a new TT");
+         TR::Node *nullChkNode = monitor->getTreeTopNode();
+         TR::Node *monentNode = nullChkNode->getFirstChild();
+         TR::Node *passThroughNode = TR::Node::create(nullChkNode, TR::PassThrough, 1, monentNode->getFirstChild());
+         TR::TreeTop *nullchkTree = TR::TreeTop::create(comp(), nullChkNode);
+         nullChkNode->setFirst(passThroughNode);
+         TR::TreeTop *prevTree = monitor->getMonitorTree()->getPrevTreeTop();
+         prevTree->join(nullchkTree);
+         nullchkTree->join(monitor->getMonitorTree());
```
fjeremic: These 3 lines should be simplified to:

```
monitor->getMonitorTree()->insertBefore(nullchkTree)
```

klangman: Thanks Filip. I'll make those changes!

fjeremic:
Jenkins test all

PR #1508, opened 2018-03-22
Add Direct Byte Buffer native memory allocation to javacores

adamfarley:
Right now we need an extension to do this.

This change removes the need for that extension.

Signed-off-by: Adam Farley <adam.farley@uk.ibm.com>


runtime/rasdump/javadump.cpp
```diff
@@ -1565,6 +1565,28 @@ outerMemCategoryCallBack (U_32 categoryCode, const char * categoryName, UDATA li
...
+		if (strcmp(categoryName,"sun.misc.Unsafe") == 0) {
+		
+			/* First, we retrieve the amount of native memory allocated to DBBs from the static variables inside the Bits Java class. */
+			J9VMThread* vmThread = _Context->onThread;
+			J9JavaVM *vm = vmThread->javaVM;
+			J9Class *bitsClass = J9VMJAVANIOBITS_OR_NULL(vm);
+			if (bitsClass != NULL) { 
+				J9Class *countAtomicLongClass = J9VMJAVANIOBITS_COUNT(vm, bitsClass);
+				J9Class *reservedMemoryAtomicLongClass = J9VMJAVANIOBITS_RESERVEDMEMORY(vm, bitsClass);
+				UDATA liveBytesForDBBs = (UDATA)J9VMJAVAUTILCONCURRENTATOMICATOMICLONG_VALUE(vm, reservedMemoryAtomicLongClass);
```
pshipton: Please check reservedMemoryAtomicLongClass and countAtomicLongClass for null before using them.

runtime/rasdump/javadump.cpp
```diff
@@ -1565,6 +1565,28 @@ outerMemCategoryCallBack (U_32 categoryCode, const char * categoryName, UDATA li
...
 
+		/* Data on native memory used for DBBs is stored in java class java.nio.Bits. */
+		if (strcmp(categoryName,"sun.misc.Unsafe") == 0) {
+		
+			/* First, we retrieve the amount of native memory allocated to DBBs from the static variables inside the Bits Java class. */
+			J9VMThread* vmThread = _Context->onThread;
+			J9JavaVM *vm = vmThread->javaVM;
+			J9Class *bitsClass = J9VMJAVANIOBITS_OR_NULL(vm);
+			if (bitsClass != NULL) { 
+				J9Class *countAtomicLongClass = J9VMJAVANIOBITS_COUNT(vm, bitsClass);
```
pshipton: These are j9object_t and not `J9Class *`, they object instances and not classes.

adamfarley:
Worked more on this today. A little trial-and-error later, and the build now compiles. It provides accurate information in the javacore, and can run side-by-side with the existing extension, so we can compare the results.

I wrote some code to allocate DBBs, and I know for a fact that my new code is giving me accurate data on that.

E.g. My java code allocates 3 lots of 200byte DBBs, and the new java dump code I added reflects this accurately.

Unfortunately the VM allocates some extra memory whenever you allocate a DBB, and Bits.java doesn't know about that, so I still have to find a way to add this extra memory usage onto the total.

See this line in Unsafe_mem.c for what I mean:

```
size = size + offsetof(J9UnsafeMemoryBlock, data); /* leave space for the linked list pointers */
```

adamfarley:
It turns out OMR changes the amount we are reserving in a way that changes based on the precise quantity of memory reserved by individual memory transactions.

Since all Bits has is the overall total for all transactions, we can't know what the amount we're actually reserving is. I've put together a set of arguments for modifying Bits, and I'm waiting to hear back from the OpenJDK community.

genie-openj9:

PR #1516, opened 2018-03-23
Move JAVA_VERSION and JCL_VERSION check in each test project build.xml

llxia:
each test project should have logic to check whether or not its test
material can be compiled against current sdk. Removed the check in
top level build_test.xml and added check in each test project build.xml.

[ci skip]


llxia:
Reviewers: @smlambert @sophia-guo

smlambert:
@sophia-guo - I won't merge until you review/comment.  Thanks.

llxia:
We still call `check_sdk` target in internal build. More investigation is needed if we want to remove this target.

sophia-guo:
Internal personal build has set xa platform is a must, so this can be deleted. Agreed do it in a separate PR, which is not a high priority as tooling will need to update too.

Issue #1429, opened 2018-03-13
Adjust default max heap size when running in container


ashu-mehra:
In non-container scenarios physical memory is shared between multiple applications, therefore it makes sense for JVM to use only a reasonable amount of physical memory for heap, so as to not strangle other applications.
However in container environments, the scenario is different. Prominent use case of containers is to run a single application (along with some agents). If its a JVM based application, then JVM is going to be the single largest consumer of memory in the container. This means JVM can take leeway and allocate large fraction of container memory as heap.

Current setting for max heap size for Java 9 are as follows:

Physical memory (P) | Default max heap size (H)
--- | ---
< 1G | P/2
1G - 2G | 512 MB 
| > 2G | P/4 

For Java 8, it is as follows:

Physical memory (P) | Default max heap size (H)
--- | ---
< 1G | P/2
| > 1G | 512 MB 

With 1/4th (in Java 9) or 512 MB (in Java 8) limit, JVM is clearly not utilizing memory that it as at its disposal.
I believe we should change the default settings when we are running in container. We can discuss out what the default should be, but I guess we can set it as high as 75%.

fyi - @charliegracie @dmitripivkine @dinogun @DanHeidinga 

Edit 1: updated heap settings for java 8 which is different from java 9.

mpirvu:
If containers are not limited in the memory they can use then having 10 containers each with a single JVM in them is no different than running 10 JVMs natively.
For a memory limited container I can see the value of increasing the default Xmx value though we have to take into the account the fact that the native JVM footprint could easily be over 100MB (classes, code cache, data cache, JIT persistent memory, JIT scratch memory for compilations, memory for GC data structures, etc).

ashu-mehra:
I agree. We should increase the default Xmx value only if the container is started with some memory limit. If no memory limit is specified, default Xmx need not be changed.

ashu-mehra:
I have got two implementations for the changes required for fixing this and #1428.
The changes are currently present in my repo for review.

In both cases, I need to update sysinfo_cgroup_get_memlimit() to return -1 if the cgroup memory limit is not specified. This will allow us to check if cgroup imposes any memory limit or not.

Plan1:
https://github.com/ashu-mehra/omr/commit/ef6d7c8cf10adb03e18574e82a7378cae25b6d9e
https://github.com/ashu-mehra/openj9/commit/c01b532c50dc82c389e82b8f4d307fa4229bb495

Plan2:
[...]

PR #1517, opened 2018-03-23
Update default max heap size when running in container

ashu-mehra:
Increase default maximum heap size when the JVM is running in a
container with memory limit set:
- if (physical_memory < 1G) defaultMaxHeap = 50% of physical_memory
- if (1G < physical_memory < 2G) defaultMaxHeap = physical_memory - 512M
- if (physical_memory > 2G) defaultMaxHeap = 75% of physical_memory

Closes: #1429 

### DIFF
--- a/runtime/gc_base/GCExtensions.cpp
+++ b/runtime/gc_base/GCExtensions.cpp
@@ -1,6 +1,6 @@
 
 /*******************************************************************************
- * Copyright (c) 1991, 2017 IBM Corp. and others
+ * Copyright (c) 1991, 2018 IBM Corp. and others
  *
  * This program and the accompanying materials are made available under
  * the terms of the Eclipse Public License 2.0 which accompanies this
@@ -84,32 +84,6 @@ MM_GCExtensions::initialize(MM_EnvironmentBase *env)
 	if (!MM_GCExtensionsBase::initialize(env)) {
 		goto failed;
 	}
-
-#if defined(OMR_ENV_DATA64)
-	if (J2SE_VERSION((J9JavaVM *)getOmrVM()->_language_vm) >= J2SE_19) {
-		uint64_t physicalMemory = 0;
-		uint64_t memoryLimit = 0;
-		uint64_t usableMemory = 0;
-		/* Initial physicalMemory as per system call. */
-		physicalMemory = j9sysinfo_get_physical_memory();
-		if (OMRPORT_LIMIT_LIMITED == j9sysinfo_get_limit(OMRPORT_RESOURCE_ADDRESS_SPACE, &memoryLimit)) {
-			/* there is a limit on the memory we can use so take the minimum of this usable amount and the physical memory */
-			usableMemory = OMR_MIN(memoryLimit, physicalMemory);
-		} else {
-			/* if there is no memory limit being imposed on us, we will use physical memory as our max */
-			usableMemory = physicalMemory;
-		}
-
-		/* extend java default max memory to 25% of physical RAM */
-		memoryMax = OMR_MAX(memoryMax, usableMemory / 4);
-		/* limit maxheapsize up to MAXIMUM_HEAP_SIZE_RECOMMENDED_FOR_3BIT_SHIFT_COMPRESSEDREFS, then can set 3bit compressedrefs as the default */
-		memoryMax = OMR_MIN(memoryMax, MAXIMUM_HEAP_SIZE_RECOMMENDED_FOR_3BIT_SHIFT_COMPRESSEDREFS);
-
-		/* Initialize Xmx, Xmdx */
-		memoryMax = MM_Math::roundToFloor(heapAlignment, (uintptr_t)memoryMax);
-		maxSizeDefaultMemorySpace = memoryMax;
-	}
-#endif /* OMR_ENV_DATA64 */
 
 #if defined(J9VM_GC_REALTIME)
 #if defined(J9VM_GC_HYBRID_ARRAYLETS)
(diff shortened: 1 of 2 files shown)


ashu-mehra:
This PR depends on eclipse/omr#2388.

runtime/gc_base/GCExtensions.cpp
```diff
@@ -252,3 +226,35 @@ MM_GCExtensions::updateIdentityHashDataForSaltIndex(UDATA index)
...
+	if (J2SE_VERSION((J9JavaVM *)getOmrVM()->_language_vm) >= J2SE_19) {
+		/* extend java default max memory to 25% of physical RAM */
+		memoryMax = OMR_MAX(memoryMax, usableMemory / 4);
+	}
+
+	/* limit maxheapsize up to MAXIMUM_HEAP_SIZE_RECOMMENDED_FOR_3BIT_SHIFT_COMPRESSEDREFS, then can set 3bit compressedrefs as the default */
+	memoryMax = OMR_MIN(memoryMax, MAXIMUM_HEAP_SIZE_RECOMMENDED_FOR_3BIT_SHIFT_COMPRESSEDREFS);
+#endif /* OMR_ENV_DATA64 */
+
+	memoryMax = MM_Math::roundToFloor(heapAlignment, (uintptr_t)memoryMax);
```
dmitripivkine: memoryMax is uintptr_t already, cast is not necessary

runtime/gc_base/GCExtensions.cpp
```diff
@@ -252,3 +226,35 @@ MM_GCExtensions::updateIdentityHashDataForSaltIndex(UDATA index)
...
+	OMRPORT_ACCESS_FROM_OMRPORT(env->getPortLibrary());
+
+	MM_GCExtensionsBase::computeDefaultMaxHeap(env);
+
+	usableMemory = omrsysinfo_get_addressable_physical_memory();
+
+	rc = omrsysinfo_cgroup_get_memlimit(&cgroupMemory);
+	if (0 == rc) {
+		/* If running in a cgroup with memory limit, extend java default max memory to 75% of cgroup limit */
+		memoryMax = (uintptr_t)OMR_MAX(cgroupMemory/2, cgroupMemory - J9_NATIVE_FOOTPRINT_EXCLUDING_HEAP);
```
dmitripivkine: missed spaces around /

runtime/oti/j9.h
```diff
@@ -273,6 +273,8 @@ static const struct { \
 #define MAXIMUM_HEAP_SIZE_RECOMMENDED_FOR_COMPRESSEDREFS	((U_64)57 * 1024 * 1024 * 1024)
 #define MAXIMUM_HEAP_SIZE_RECOMMENDED_FOR_3BIT_SHIFT_COMPRESSEDREFS	((U_64)25 * 1024 * 1024 * 1024)
 
+#define J9_NATIVE_FOOTPRINT_EXCLUDING_HEAP ((U_64)512*1024*1024)
```
dmitripivkine: missed spaces around stars

ashu-mehra:
@dmitripivkine I have updated the commit to address comments.

dmitripivkine:
Jenkins test sanity pLinux depends eclipse/omr#2388

dmitripivkine:
Jenkins test sanity pLinux depends eclipse/omr#2388

runtime/oti/j9.h
```diff
@@ -273,6 +273,8 @@ static const struct { \
 #define MAXIMUM_HEAP_SIZE_RECOMMENDED_FOR_COMPRESSEDREFS	((U_64)57 * 1024 * 1024 * 1024)
 #define MAXIMUM_HEAP_SIZE_RECOMMENDED_FOR_3BIT_SHIFT_COMPRESSEDREFS	((U_64)25 * 1024 * 1024 * 1024)
 
+#define J9_NATIVE_FOOTPRINT_EXCLUDING_HEAP ((U_64)512 * 1024 * 1024)
```
DanHeidinga: Given this is only used in GCExtensions.cpp, shouldn't it be defined there?
Where does this value come from?  It needs a comment describing why this value is valid at a minimum

runtime/gc_base/GCExtensions.cpp
```diff
@@ -252,3 +226,35 @@ MM_GCExtensions::updateIdentityHashDataForSaltIndex(UDATA index)
 {
 	getJavaVM()->identityHashData->hashSaltTable[index] = (U_32)convertValueToHash(getJavaVM(), getJavaVM()->identityHashData->hashSaltTable[index]);
 }
+
+void
+MM_GCExtensions::computeDefaultMaxHeap(MM_EnvironmentBase *env)
+{
+	int32_t rc = 0;
```
DanHeidinga: The variables can be defined at first use since this is a cpp file.  Keeping variable life times short makes it easier to reason about the code.

runtime/gc_base/GCExtensions.cpp
```diff
@@ -85,32 +85,6 @@ MM_GCExtensions::initialize(MM_EnvironmentBase *env)
...
-		}
-
-		/* extend java default max memory to 25% of physical RAM */
-		memoryMax = OMR_MAX(memoryMax, usableMemory / 4);
-		/* limit maxheapsize up to MAXIMUM_HEAP_SIZE_RECOMMENDED_FOR_3BIT_SHIFT_COMPRESSEDREFS, then can set 3bit compressedrefs as the default */
-		memoryMax = OMR_MIN(memoryMax, MAXIMUM_HEAP_SIZE_RECOMMENDED_FOR_3BIT_SHIFT_COMPRESSEDREFS);
-
-		/* Initialize Xmx, Xmdx */
-		memoryMax = MM_Math::roundToFloor(heapAlignment, (uintptr_t)memoryMax);
-		maxSizeDefaultMemorySpace = memoryMax;
```
DanHeidinga: Was this line lost in the code refactoring?

runtime/oti/j9.h
ashu-mehra: I can move this to GCExtensions.cpp.
This value is suggested by @mpirvu in his comment https://github.com/eclipse/openj9/issues/1429#issuecomment-375045690. I guess this is just a rough estimate, isn't it @mpirvu ? Not sure if there is a better way to estimate this value.

runtime/gc_base/GCExtensions.cpp
ashu-mehra: Yup, thanks for catching it.

runtime/oti/j9.h
DanHeidinga: The rationale from that comment should be included in the code so we can figure out where the magic numbers came from in the future.

ashu-mehra: @DanHeidinga I updated the comment as:
```
                /* If running in a cgroup with memory limit > 1G, reserve at-least 512 MB for JVM's internal requirements
                 */
#define J9_NATIVE_FOOTPRINT_EXCLUDING_HEAP ((U_64)512 * 1024 * 1024)
```
Is that better?

DanHeidinga: Should the #define include `cgroup` or `container` in its name?  

I'd prefer describing the purpose of this define before tying it to cgroups & heap sizes so that the validity of the constant is clear.  Maybe something like:
```C
/* A very rough estimate of the minimum amount of memory required by the JVM excluding
 * the heap.  The size is a conservative estimate, attempting to leave room the JVM's internal
 * requirements given that one compilation thread can use up to 256M.
 */
#define J9_NATIVE_FOOTPRINT_EXCLUDING_HEAP ((U_64)512 * 1024 * 1024)
```

The rest of your comment is useful info and should be included in `computeDefaultMaxHeap`.

runtime/gc_base/GCExtensions.cpp
```diff
@@ -85,32 +85,7 @@ MM_GCExtensions::initialize(MM_EnvironmentBase *env)
...
-	}
-#endif /* OMR_ENV_DATA64 */
-
+	maxSizeDefaultMemorySpace = memoryMax;
```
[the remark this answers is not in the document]
ashu-mehra: Aah! I now realize `maxSizeDefaultMemorySpace = memoryMax;` is not really required.
`GCExtensions::initialize` calls super class `MM_GCExtensionsBase::initialize` which calls virtual method `computeDefaultMaxHeap` (which would be MM_GCExtensions's version in our case) and then sets `maxSizeDefaultMemorySpace = memoryMax;`.

So this statement is not actually required here at all.
Sorry for the confusion. Should have clarified it earlier itself.

PR #1525, opened 2018-03-23
Add new CP types to DDR

fengxue-IS:
- add J9CPTYPE_INTERFACE_STATIC_Method, J9CPTYPE_INTERFACE_INSTANCE_METHOD
  to DDR cpShapeDescription mapping


DanHeidinga:
@fengxue-IS This is mapping the two new interface_{static, special} to the same chars as the existing {static, special}, correct?

fengxue-IS:
yes, the new type are only subset of the old type, so they map to the same chars

DanHeidinga:
Jenkins test sanity, extended plinux

DanHeidinga:
enkins test sanity plinux

DanHeidinga:
Jenkins test sanity plinux

DanHeidinga:
Jenkins test extended plinux

fengxue-IS:
@DanHeidinga FYI I think the test that uses DDR J9BCUtil is only in Windows IA32

DanHeidinga:
> FYI I think the test that uses DDR J9BCUtil is only in Windows IA32

I know we found the problem on Win32 but I was hoping the tests we also present on other platforms.  As we don't have windows h/w yet, this is good to merge.

PR #1529, opened 2018-03-23
WIP: Fold constant displacements into vector instructions

NigelYiboYu:
Vector store and load instructions can take a 12-bit unsigned
displacement to reference memory locations. This can be folded into
various vector load and store instructions, such as VSTRL, VLRL, VSTEB,
and etc.

In doing so, the JIT generated code can potentially save a 'load address'
instruction.


NigelYiboYu:
@joransiu I'm currently testing this change. Should be OK to start reviewing.

runtime/compiler/trj9/z/codegen/J9TreeEvaluator.cpp
```diff
@@ -20927,6 +20932,81 @@ J9::Z::TreeEvaluator::pdstoreVectorEvaluatorHelper(TR::Node *node, TR::CodeGener
...
+ *
+ * The LA instruction can be eliminated if the constant is folded into the store instruction, resulting in:
+ *
+ * \verbatim
+ * VSTRL   VR0,X'8'(GPR1),0
+ * \endverbatim
+ *
+ */
+TR::Register *
+J9::Z::TreeEvaluator::foldAddressDisplacementIntoVectorLoadStore(TR::Node* addressNode,
```
fjeremic: Why is this function needed at all? There already exists logic in `generateMemoryReference` API to handle folding of the displacement constant into the memref. In the very first deleted lines in this PR:

```
TR::Register* addressReg = cg->evaluate(addressNode);
TR::MemoryReference* sourceMR = generateS390MemoryReference(addressReg, NULL, 0, cg);
```

Instead of doing this simply constructing the memory reference by passing in the `addressNode` will result in no `LA` instruction being generated. See the following API:

https://github.com/eclipse/omr/blob/225a9d93e5e202446df9dc368b04e567c48cbe34/compiler/z/codegen/OMRMemoryReference.hpp#L467

This is how memory references are created for all generic loads without materializing the memory reference. If code does not exist to handle this particular case then logic needs to be added in the MemoryReference constructor where all the other folding logic is.

NigelYiboYu: Changed the way we create memory reference to use the API you pointed out.
This change depends on an OMR change: https://github.com/eclipse/omr/pull/2391

NigelYiboYu:
This PR depends on OMR PR: https://github.com/eclipse/omr/pull/2391

PR #1531, opened 2018-03-23
[Nest mates] Updates java.lang.Class API for new nestmates spec draft

taliamccormick:
Earlier spec draft for java.lang.Class nestmates API methods 'getNestHost' and 'getNestMembers' threw no exceptions and LinkageError respectively. A newer spec draft allows both methods to throw a SecurityException. This commit adds the necessary check for this to occur.

Relevant portions of the newer spec draft include:
```
    public Class<?> getNestHost()
    ...
    Throws:
        SecurityException - If the returned class is not the current class, and if a security manager,
 s, is present and the caller's class loader is not the same as or an ancestor of the class loader
for the returned class and invocation of s.checkPackageAccess() denies access to the package of the 
returned class
    ...
    public Class<?>[] getNestMembers()
    ...
    Throws:
    ...
        SecurityException - If any returned class is not the current class, and if a security manager,
s, is present and the caller's class loader is not the same as or an ancestor of the class loader
for that returned class and invocation of s.checkPackageAccess() denies access to the package of that 
returned class
```
(http://cr.openjdk.java.net/%7Edholmes/8010319/specs/java.lang/java/lang/Class.html)

Together with https://github.com/eclipse/openj9/pull/1501, resolves https://github.com/eclipse/openj9/issues/1457. There are no dependencies between the two PRs. 

### DIFF
--- a/jcl/src/java.base/share/classes/java/lang/Class.java
+++ b/jcl/src/java.base/share/classes/java/lang/Class.java
@@ -247,8 +247,8 @@ public final class Class<T> implements java.io.Serializable, GenericDeclaration,
 	
 /*[IF Valhalla-NestMates]*/
 	/* Store the results of getNestHostImpl() and getNestMembersImpl() respectively */
-	private  Class<?> nestHost;
-	private  Class<?>[] nestMembers;
+	private Class<?> nestHost;
+	private Class<?>[] nestMembers;
 /*[ENDIF] Valhalla-NestMates*/
 	
 /**
@@ -4389,11 +4389,10 @@ private native Class<?>[] getNestMembersImpl();
 /**
  * Answers the host class of the receiver's nest.
  * 
- * @throws 		SecurityException if a returned class is not the current class, a
- * 				security manager, s, is present, the caller's class loader is not
- * 				the same or an ancestor of that returned class, and s.checkPackageAccess()
- * 				denies access
- * @return		the host class of the receiver.
+ * @throws SecurityException if nestHost is not same as the current class, a security manager
+ *	is present, the classloader of the caller is not the same or an ancestor of nestHost
+ * 	class, and checkPackageAccess() denies access
+ * @return the host class of the receiver.
  */
 @CallerSensitive
 public Class<?> getNestHost() throws SecurityException {
@@ -4414,7 +4413,7 @@ public Class<?> getNestHost() throws SecurityException {
 			ClassLoader nestHostClassLoader = nestHost.getClassLoader();			
 			if (!doesClassLoaderDescendFrom(nestHostClassLoader, callerClassLoader)) {
 				String nestHostPackageName = nestHost.getPackageName();
-				if (nestHostPackageName != null) {
+				if ((nestHostPackageName != null) && (nestHostPackageName != "")) {
 					securityManager.checkPackageAccess(nestHostPackageName);
 				}
 			}
@@ -4426,8 +4425,8 @@ public Class<?> getNestHost() throws SecurityException {
 /**
  * Returns true if the class passed has the same nest top as this class.
  * 
- * @param		that		The class to compare
- * @return		true if class is a nestmate of this class; false otherwise.
+ * @param that The class to compare
+ * @return true if class is a nestmate of this class; false otherwise.
  *
  */
 public boolean isNestmateOf(Class<?> that) {
@@ -4445,13 +4444,12 @@ public boolean isNestmateOf(Class<?> that) {
 /**
  * Answers the nest member classes of the receiver's nest host.
  *
- * @throws		SecurityException if a SecurityManager is present and package access is not allowed
- * @throws		LinkageError if there is any problem loading or validating a nest member or the nest host
- * @throws 		SecurityException if a returned class is not the current class, a
- * 				security manager, s, is present, the caller's class loader is not
- * 				the same or an ancestor of that returned class, and s.checkPackageAccess()
- * 				denies access
- * @return		the host class of the receiver.
+ * @throws SecurityException if a SecurityManager is present and package access is not allowed
+ * @throws LinkageError if there is any problem loading or validating a nest member or the nest host
+ * @throws SecurityException if a returned class is not the current class, a security manager is enabled,
+ *	the caller's class loader is not the same or an ancestor of that returned class, and the
+ * 	checkPackageAccess() denies access
+ * @return the host class of the receiver.
  */
 @CallerSensitive
 public Class<?>[] getNestMembers() throws LinkageError, SecurityException {
@@ -4465,7 +4463,7 @@ public Class<?>[] getNestMembers() throws LinkageError, SecurityException {
 		ClassLoader callerClassLoader = ClassLoader.getCallerClassLoader();
 		if (!doesClassLoaderDescendFrom(nestMemberClassLoader, callerClassLoader)) {
 			String nestMemberPackageName = this.getPackageName();
-			if (nestMemberPackageName != null) {
+			if ((nestMemberPackageName != null) && (nestMemberPackageName != "")) {
 				securityManager.checkPackageAccess(nestMemberPackageName);
 			}
 		}
--- a/jcl/src/java.base/share/classes/java/lang/Class.java
+++ b/jcl/src/java.base/share/classes/java/lang/Class.java
@@ -4409,9 +4409,7 @@ public Class<?> getNestHost() throws SecurityException {
 	 */
 	if (nestHost != this) {
 		SecurityManager securityManager = System.getSecurityManager();
-		if (nestHost == null) {
-			nestHost = this;
-		} else if (securityManager != null) {
+		if (securityManager != null) {
 			ClassLoader callerClassLoader = ClassLoader.getCallerClassLoader();
 			ClassLoader nestHostClassLoader = nestHost.getClassLoader();			
 			if (!doesClassLoaderDescendFrom(nestHostClassLoader, callerClassLoader)) {
@@ -4460,15 +4458,6 @@ public Class<?>[] getNestMembers() throws LinkageError, SecurityException {
 	if (nestMembers == null) {
 		nestMembers = getNestMembersImpl();
 	}
-	/* The specification requires that if:
-	 *    - a returned class is not the current class
-	 *    - a security manager, s, is present
-	 *    - the caller's class loader is not the same or an ancestor of the returned class
-	 *    - s.checkPackageAccess() disallows access to the package of the returned class
-	 * then throw a SecurityException. Since all classes in the same nest
-	 * must be in the same runtime package and thus the same class loader,
-	 * it is sufficient to check classloader ancestry once.
-	 */
 	SecurityManager securityManager = System.getSecurityManager();
 	if (securityManager != null) {
 		/* All classes in a nest must be in the same runtime package and therefore same classloader */
--- a/jcl/src/java.base/share/classes/java/lang/Class.java
+++ b/jcl/src/java.base/share/classes/java/lang/Class.java
@@ -52,9 +52,9 @@ import sun.reflect.generics.scope.ClassScope;
 import sun.reflect.annotation.AnnotationType;
 import java.util.Arrays;
 import com.ibm.oti.vm.VM;
-/*[IF Valhalla-NestMates]*/
+/*[IF Java11]*/
 import static com.ibm.oti.util.Util.doesClassLoaderDescendFrom;
-/*[ENDIF] Valhalla-NestMates*/
+/*[ENDIF] Java11*/
 
 /*[IF Sidecar19-SE]
 import jdk.internal.misc.Unsafe;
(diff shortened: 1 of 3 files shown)


jcl/src/java.base/share/classes/java/lang/Class.java
```diff
@@ -4296,36 +4296,108 @@ ConstantPool getConstantPool() {
...
+			if (!doesClassLoaderDescendFrom(nestHostClassLoader, callerClassLoader)) {
+				String nestHostPackageName = nestHost.getPackageName();
+				if (nestHostPackageName != null) {
+					securityManager.checkPackageAccess(nestHostPackageName);
+				}
+			}
+		}
+	}
+
+	return nestHost;
```
tajila: calling getNestHost is kinda expensive, we should cache in a private local

jcl/src/java.base/share/classes/java/lang/Class.java
```diff
@@ -4296,36 +4296,108 @@ ConstantPool getConstantPool() {
...
- * duplicates, or to explicitly include the nest host. It is not required
- * that an implementation of this method removes these duplicates.
  * 
  * @param		that		The class to compare
  * @return		true if class is a nestmate of this class; false otherwise.
  *
  */
 public boolean isNestmateOf(Class<?> that) {
-	return this.getNestHost() == that.getNestHost();
+	return this.getNestHostImpl() == that.getNestHostImpl();
```
tajila: @DanHeidinga 
I find it odd that `isNestmateOf` can't throw a security exception but `getNestHost` does.
From the spec `Two classes are nestmates if they have the same nest host .`

If a caller doesn't have permission query the nesthost, should it be allowed to test whether the nest host is equivalent to another classes nesthost?

jcl/src/java.base/share/classes/java/lang/Class.java
```diff
@@ -4296,36 +4296,108 @@ ConstantPool getConstantPool() {
...
+		 *    - the caller's class loader is not the same or an ancestor of the returned class
+		 *    - s.checkPackageAccess() disallows access to the package of the returned class
+		 * then throw a SecurityException. Since all classes in the same nest
+		 * must be in the same runtime package and thus the same class loader,
+		 * it is sufficient to check classloader ancestry once.
+		 */
+		SecurityManager securityManager = System.getSecurityManager();
+		if (null != securityManager) {
+			/* All classes in a nest must be in the same runtime package & therefore same classloader */
+			ClassLoader nestHostClassLoader = nestMembers[0].getClassLoader();
```
tajila: can use `this.getClassLoader`

tajila: since it will be the same

jcl/src/java.base/share/classes/java/lang/Class.java
```diff
@@ -4296,36 +4296,108 @@ ConstantPool getConstantPool() {
...
+			ClassLoader callerClassLoader = ClassLoader.getCallerClassLoader();
+			if (!doesClassLoaderDescendFrom(nestHostClassLoader, callerClassLoader)) {
+				for (Class<?> nm : nestMembers) {
```
tajila: dont need to loop since they will all have the same package, do `this.getPackageName()`

jcl/src/java.base/share/classes/java/lang/Class.java
```diff
@@ -4296,36 +4296,108 @@ ConstantPool getConstantPool() {
...
+	 *    - the returned class is not the current class
+	 *    - a security manager is present
+	 *    - the caller's class loader is not the same or an ancestor of the returned class
+	 *    - s.checkPackageAccess() disallows access to the package of the returned class
+	 * then throw a SecurityException.
+	 */
+	if (nestHost != this) {
+		SecurityManager securityManager = System.getSecurityManager();
+		if (nestHost == null) {
+			nestHost = this;
```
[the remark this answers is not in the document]
taliamccormick: We should never be able to reach the case where `nestHost == null` is true. Either the nest host is correctly loaded, verified, and returned by `getNestHostImpl` or the class itself is returned by `getNestHostImpl`. I had assumed that there was a possibility - either existing or with future spec changes - of `getNestHostImpl` failing without throwing and thought it was safer to check if nestHost is null. Would it be better to remove it?

taliamccormick:
Comment got lost when I pushed new changes, so rewriting it here:

Both of the native methods `getNestHostImpl` and `getNestMembersImpl` should never return null. `getNestHostImpl` should either return a valid verified nest host or itself. `getNestMembersImpl` should return a valid nest members list or throw.

I included null checks for both methods within the java code. While neither should be able to return null, I assumed there's a possibility - either existing or within future spec changes - of one of them failing without throwing and that it is safer to check if a return value is null, prior to checking/accessing. 

Should I remove the null check?

jcl/src/java.base/share/classes/java/lang/Class.java
```diff
@@ -4301,31 +4306,92 @@ ConstantPool getConstantPool() {
...
-	return this.getNestHost() == that.getNestHost();
+	Class<?> thisNestHost = this.nestHost;
+	if (null == thisNestHost) {
+		thisNestHost = this.getNestHostImpl();
+	}
+	Class<?> thatNestHost = that.nestHost;
+	if (null == thatNestHost) {
+		thatNestHost = that.getNestHostImpl();
+	}
+	return thisNestHost == thatNestHost;
```
babsingh: brackets - `(thisNestHost == thatNestHost)`

jcl/src/java.base/share/classes/java/lang/Class.java
```diff
@@ -4301,31 +4306,92 @@ ConstantPool getConstantPool() {
  *
  * @return		the host class of the receiver.
  */
-public Class<?> getNestHost() {
-	return getNestHostImpl();
+@CallerSensitive
+public Class<?> getNestHost() throws SecurityException {
+	if (null == nestHost) {
```
babsingh: use one convention: `(nestHost == null)` or `(null == nestHost)` everywhere.

jcl/src/java.base/share/classes/java/lang/Class.java
```diff
@@ -4301,31 +4306,92 @@ ConstantPool getConstantPool() {
...
+
+		/* The specification requires that if:
+		 *    - the returned class is not the current class
+		 *    - a security manager is present
+		 *    - the caller's class loader is not the same or an ancestor of the returned class
+		 *    - s.checkPackageAccess() disallows access to the package of the returned class
+		 * then throw a SecurityException.
+		 */
+		if (nestHost != this) {
+			SecurityManager securityManager = System.getSecurityManager();
```
babsingh: shouldn't the security check be always done `(null != nestHost)`? first call to `getNestHost` will set the `nestHost` and perform the security check. for subsequent `getNestHost` calls, `nestHost` will not be `null` and no security check will be performed.

jcl/src/java.base/share/classes/java/lang/Class.java
```diff
@@ -4301,31 +4306,92 @@ ConstantPool getConstantPool() {
  *
  * @return		the host class of the receiver.
```
babsingh: `@throws SecurityException` is missing in the `getNestHostImpl's` description.

jcl/src/java.base/share/classes/java/lang/invoke/MethodHandles.java
```diff
@@ -1483,7 +1462,7 @@ void checkSecurity(Class<?> definingClass, Class<?> referenceClass, int modifier
 					VMLangAccess vma = getVMLangAccess();
 
 					ClassLoader referenceClassLoader = referenceClass.getClassLoader();
-					if (isWeakenedLookup() || !doesClassLoaderDescendFrom(referenceClassLoader, accessClass.getClassLoader())) {
+					if (isWeakenedLookup() || !com.ibm.oti.util.Util.doesClassLoaderDescendFrom(referenceClassLoader, accessClass.getClassLoader())) {
```
babsingh: you can - `import static com.ibm.oti.util.Util.doesClassLoaderDescendFrom;`. then, you can directly call `doesClassLoaderDescendFrom`.

jcl/src/java.base/share/classes/java/lang/invoke/MethodHandles.java
```diff
@@ -1495,7 +1474,7 @@ void checkSecurity(Class<?> definingClass, Class<?> referenceClass, int modifier
 
 						/* third check */
 						if (definingClass.getClassLoader() != referenceClassLoader) {
-							if (!doesClassLoaderDescendFrom(definingClass.getClassLoader(), accessClass.getClassLoader())) {
+							if (!com.ibm.oti.util.Util.doesClassLoaderDescendFrom(definingClass.getClassLoader(), accessClass.getClassLoader())) {
```
babsingh: you can - `import static com.ibm.oti.util.Util.doesClassLoaderDescendFrom;`. then, you can directly call `doesClassLoaderDescendFrom`.

jcl/src/java.base/share/classes/java/lang/Class.java
```diff
@@ -4301,31 +4306,92 @@ ConstantPool getConstantPool() {
...
+		 * then throw a SecurityException.
+		 */
+		if (nestHost != this) {
+			SecurityManager securityManager = System.getSecurityManager();
+			if (nestHost == null) {
+				nestHost = this;
+			} else if (securityManager != null) {
+				ClassLoader callerClassLoader = ClassLoader.getCallerClassLoader();
+				ClassLoader nestHostClassLoader = nestHost.getClassLoader();			
+				if (!com.ibm.oti.util.Util.doesClassLoaderDescendFrom(nestHostClassLoader, callerClassLoader)) {
```
babsingh: you can - `import static com.ibm.oti.util.Util.doesClassLoaderDescendFrom;`. then, you can directly call `doesClassLoaderDescendFrom`.

jcl/src/java.base/share/classes/java/lang/Class.java
```diff
@@ -4301,31 +4306,92 @@ ConstantPool getConstantPool() {
...
+		 * then throw a SecurityException. Since all classes in the same nest
+		 * must be in the same runtime package and thus the same class loader,
+		 * it is sufficient to check classloader ancestry once.
+		 */
+		SecurityManager securityManager = System.getSecurityManager();
+		if (null != securityManager) {
+			/* All classes in a nest must be in the same runtime package & therefore same classloader */
+			ClassLoader nestMemberClassLoader = this.getClassLoader();
+			ClassLoader callerClassLoader = ClassLoader.getCallerClassLoader();
+			if (!com.ibm.oti.util.Util.doesClassLoaderDescendFrom(nestMemberClassLoader, callerClassLoader)) {
```
babsingh: you can - `import static com.ibm.oti.util.Util.doesClassLoaderDescendFrom;`. then, you can directly call `doesClassLoaderDescendFrom`.

jcl/src/java.base/share/classes/java/lang/Class.java
```diff
@@ -4301,31 +4306,92 @@ ConstantPool getConstantPool() {
...
+		/* The specification requires that if:
+		 *    - the returned class is not the current class
+		 *    - a security manager is present
+		 *    - the caller's class loader is not the same or an ancestor of the returned class
+		 *    - s.checkPackageAccess() disallows access to the package of the returned class
+		 * then throw a SecurityException.
+		 */
+		if (nestHost != this) {
+			SecurityManager securityManager = System.getSecurityManager();
+			if (nestHost == null) {
```
babsingh: will `getNestHostImpl` return `null` for `nestHost`? if not, then this `if` statement isn't needed.

jcl/src/java.base/share/classes/java/lang/Class.java
```diff
@@ -4301,31 +4306,92 @@ ConstantPool getConstantPool() {
...
  * @return		the host class of the receiver.
  */
-public Class<?>[] getNestMembers() {
-	return getNestMembersImpl();
+@CallerSensitive
+public Class<?>[] getNestMembers() throws LinkageError, SecurityException {
+	if (null == nestMembers) {
+		nestMembers = getNestMembersImpl();
+	}
+	if (null != nestMembers) {
```
babsingh: will `getNestMembersImpl` return a `null`? if not, then this `if` statement is not needed.

taliamccormick:
Made PR changes per Babneet's review:
- Removing unnecessary null-check in `getNestHost()` and `getNestMembers`
- Re-ordering null checks
- Importing `com.ibm.oti.util.Util.doesClassLoaderDescendFrom` to allow direct access of `doesClassLoaderDescendFrom`
- Adding `@throws SecurityException` to description of `getNestHost()` and `getNestMembers`

jcl/src/java.base/share/classes/java/lang/Class.java
taliamccormick: Missed that - thanks

taliamccormick:
@babsingh Updated per review; squashed; and retested.

babsingh:
Senior Reviewer: @DanHeidinga

DanHeidinga:

DanHeidinga:

DanHeidinga:
Jenkins compile xlinux jdk11

DanHeidinga:

Issue #1272, opened 2018-02-26
ConstantDynamic: Resolving Constant_Dynamic entries


tajila:
<h1>LDC bytecode</h1>
The ldc bytecode will have to be modified to detect Condy constants. If the constant or exception doesn’t exist in the ramConstantPool  attempt to resolve the constant. If the exception exists throw the exception (Note: if LDC of Condy causes an exception, the same exception must be thrown every time). If the constant exists return the constant.

<h1>LDC2lw bytecode</h1>
The ldc2_w byte will check the cp type via something like `J9_CP_TYPE(cpShapeDescription, cpIndex)`. If it is not a Condy entry normal execution will occur. If it is, romConstantPool entry will be NULL check for the Condy constant, then for the exception. If both are NULL then resolution will occur. When the constant is resolved the long or double will be extracted from the object with something like `J9VMJAVALANGLONG_VALUE` and then placed on the stack.

<h1>Resolving native side</h1>
Create a new helper in resovleSupport to resolve Condy CPentries. This can share a common helper with `resolveInvokeDynamic` as much of the functionality will be the same. 

In the case of Condy, the helper will need to fetch the `nameAndSig` and `bsmIndex` from the romCP entry. Otherwise, the rest of the calculations required to determine the start `bsmData` are the same as with Indy

The helper will call `sendResolveConstantDynamic` (new function) to retrieve and run the BSM for the LDC. After the result will be cached in the ramConstantPool it is Condy.

In the Condy case the helper will check if an exception was thrown upon returning from `sendResolveConsntantDynamic`. If an exception exists,  it will be stored in the exception slot in the ramConstantPool entry.

`sendResolveConstantDynamic` will be similar to `sendResolveInvokeDynamic`, it will perform a java call-in to `MethodHandle.resolveConstantDynamic()`

<h1>Resolving Java side</h1>
A new function  `MethodHandle.resolveConstantDynamic()` will need to be added, this can share a common helper with `MethodHandle.resolveInvokeDynamic` as much of the code is the same.   A boolean can be added the helper args to identify whether it is resolving for Condy or Indy.   

[...]

PR #1533, opened 2018-03-23
Implement resolve code for ConstantDynamic entries

fengxue-IS:
- added ConstantDynamic case to ldc, ldc_w, ldc2_w bytecode
- added native resolve helper for ConstantDynamic
- new MethodHandle resolve method for Java side
- implement mutex on ramslot to ensure single resolve is allowed

Close: #1272


fengxue-IS:
depend: #1426

DanHeidinga:
Once  #1426 is merged (should be later today when the builds finish), can you rebase this to remove the changes from  #1426?

fengxue-IS:
sure, I ll update the PR after #1426 is merged

fengxue-IS:
branch rebase on master

runtime/vm/j9vm.tdf
```diff
@@ -561,6 +561,9 @@ TraceExit=Trc_VM_jitFillOSRBuffer_Exit Overhead=1 Level=3 Template="jitFillOSRBu
 TraceEntry=Trc_VM_sendResolveMethodHandle_Entry Overhead=1 Level=2 Template="sendResolveMethodHandle"
 TraceExit=Trc_VM_sendResolveMethodHandle_Exit Overhead=1 Level=2 Template="sendResolveMethodHandle"
 
+TraceEntry=Trc_VM_sendResolveConstantDynamic_Entry Overhead=1 Level=2 Template="sendResolveConstantDynamic"
```
DanHeidinga: Tracepoints can only be added to the end of the file

runtime/vm/vmthinit.c
```diff
@@ -80,6 +80,8 @@ UDATA initializeVMThreading(J9JavaVM *vm)
 
 		omrthread_monitor_init_with_name(&vm->vmRuntimeStateListener.runtimeStateListenerMutex, 0, "VM state notification mutex") ||
 
+		omrthread_monitor_init_with_name(&vm->constantDynamicMutex, 0, "Lock for Constant Dynamic resolve store") ||
```
DanHeidinga: This is a single lock shared by all threads and all constantdynamics.  We should be able to resolve without a lock

runtime/vm/resolvesupport.cpp
```diff
@@ -1839,6 +1839,93 @@ resolveMethodHandleRef(J9VMThread *vmThread, J9ConstantPool *ramCP, UDATA cpInde
...
+	J9SRP *callSiteData = (J9SRP *) J9ROMCLASS_CALLSITEDATA(romClass);
+	U_16 *bsmIndices = (U_16 *) (callSiteData + romClass->callSiteCount);
+	U_16 *bsmData = bsmIndices + romClass->callSiteCount;
+	J9ROMNameAndSignature* nameAndSig = SRP_PTR_GET(&romConstantRef->nameAndSignature, J9ROMNameAndSignature*);
+	U_32 bsmIndex = romConstantRef->bsmIndexAndCpType >> J9DescriptionCpTypeShift;
+	J9MemoryManagerFunctions *gcFuncs = vmThread->javaVM->memoryManagerFunctions;
+	UDATA resolved = 0;
+
+	/* Check if already resolved */
+	omrthread_monitor_enter(vmThread->javaVM->constantDynamicMutex);
```
DanHeidinga: A memory barrier should be sufficient here to see if this has been resolved already or not.

runtime/vm/resolvesupport.cpp
```diff
@@ -1839,6 +1839,93 @@ resolveMethodHandleRef(J9VMThread *vmThread, J9ConstantPool *ramCP, UDATA cpInde
...
+		/* Walk bsmData */
+		for (U_32 i = 0; i < bsmIndex; i++) {
+			bsmData += bsmData[1] + 2;
+		}
+
+		sendResolveConstantDynamic(vmThread, ramCP, cpIndex, nameAndSig, bsmData);
+		value = (j9object_t)vmThread->returnValue;
+
+		/* lock write access to constantDynamic ram slots */
+		omrthread_monitor_enter(vmThread->javaVM->constantDynamicMutex);
```
DanHeidinga: Similar concern here.  We need to spend some time thinking about the ordering, memory barriers, and CAS to handle setting and reading these fields without a lock

runtime/vm/j9vm.tdf
fengxue-IS: fixed

jcl/src/java.base/share/classes/java/lang/invoke/MethodHandle.java
```diff
@@ -828,6 +849,150 @@ MethodHandle insertArguments(MethodHandle equivalent, MethodHandle unboxingHandl
...
+				case 6:
+					cpEntry = cp.getDoubleAt(index);
+					break;
+				case 13:
+					cpEntry = getCPMethodTypeAt(internalRamClass, index);
+					break;
+				case 14:
+					cpEntry = getCPMethodHandleAt(internalRamClass, index);
+					break;
+				default:
```
tajila: also need a `case 17:` for constant dynamic. Right now this will be a 
```
//TODO
break;
```

until https://github.com/eclipse/openj9/pull/1649 is complete

`resolveInvokeDynamic` will have to be updated as well

jcl/src/java.base/share/classes/java/lang/invoke/MethodHandle.java
```diff
@@ -828,6 +849,150 @@ MethodHandle insertArguments(MethodHandle equivalent, MethodHandle unboxingHandl
...
+			case 6:
+				result = (Object) bsm.invoke(staticArgs[0], staticArgs[1], staticArgs[2], staticArgs[3], staticArgs[4], staticArgs[5]);
+				break;
+			case 7:
+				result = (Object) bsm.invoke(staticArgs[0], staticArgs[1], staticArgs[2], staticArgs[3], staticArgs[4], staticArgs[5], staticArgs[6]);
+				break;
+			default:
+				result = (Object) bsm.invokeWithArguments(staticArgs);
+				break;
+			}
```
tajila: need to do the identity check here

runtime/vm/resolvesupport.cpp
```diff
@@ -1839,6 +1839,102 @@ resolveMethodHandleRef(J9VMThread *vmThread, J9ConstantPool *ramCP, UDATA cpInde
...
+j9object_t
+resolveConstantDynamic(J9VMThread *vmThread, J9ConstantPool *ramCP, UDATA cpIndex, UDATA resolveFlags)
+{
+	J9RAMConstantDynamicRef *ramCPEntry = (J9RAMConstantDynamicRef*)ramCP + cpIndex;
+	j9object_t value = ramCPEntry->value;
+
+	J9Class *ramClass = ramCP->ramClass;
+	J9ROMClass *romClass = ramClass->romClass;
+	J9ROMConstantDynamicRef *romConstantRef = (J9ROMConstantDynamicRef*)(J9_ROM_CP_FROM_CP(ramCP) + cpIndex);
+	J9SRP *callSiteData = (J9SRP *) J9ROMCLASS_CALLSITEDATA(romClass);
```
tajila: you can declare callSiteData, bsmIndices, bsmData, bsmIndex and nameAndSig later since they are only being used in the inner scope

runtime/vm/resolvesupport.cpp
```diff
@@ -1839,6 +1839,102 @@ resolveMethodHandleRef(J9VMThread *vmThread, J9ConstantPool *ramCP, UDATA cpInde
...
+	U_16 *bsmIndices = (U_16 *) (callSiteData + romClass->callSiteCount);
+	U_16 *bsmData = bsmIndices + romClass->callSiteCount;
+	U_32 bsmIndex = romConstantRef->bsmIndexAndCpType >> J9DescriptionCpTypeShift;
+	J9ROMNameAndSignature* nameAndSig = SRP_PTR_GET(&romConstantRef->nameAndSignature, J9ROMNameAndSignature*);
+
+	J9MemoryManagerFunctions *gcFuncs = vmThread->javaVM->memoryManagerFunctions;
```
tajila: you can also declare that later

runtime/vm/resolvesupport.cpp
```diff
@@ -1839,6 +1839,102 @@ resolveMethodHandleRef(J9VMThread *vmThread, J9ConstantPool *ramCP, UDATA cpInde
...
+	J9Class *throwable = internalFindKnownClass(vmThread, J9VMCONSTANTPOOL_JAVALANGTHROWABLE, J9_FINDKNOWNCLASS_FLAG_INITIALIZE);
```
tajila: j.l.Throwable should be loaded  before any exception object is created. So by the time you need it you should be able to get it directly from the JCl constant pool.

runtime/vm/resolvesupport.cpp
tajila: same with romClass and romConstantRef

runtime/vm/resolvesupport.cpp
```diff
@@ -1839,6 +1839,102 @@ resolveMethodHandleRef(J9VMThread *vmThread, J9ConstantPool *ramCP, UDATA cpInde
...
+	UDATA resolved = 0;
```
tajila: you can make this a BOOLEAN since you are only using it for binary checks

runtime/vm/resolvesupport.cpp
```diff
@@ -1839,6 +1839,102 @@ resolveMethodHandleRef(J9VMThread *vmThread, J9ConstantPool *ramCP, UDATA cpInde
...
+retry:
+	/* Check if already resolved */
+	if (NULL != value) {
+		resolved = 1;
+	} else if (NULL != ramCPEntry->exception) {
+		if (isSameOrSuperClassOf(throwable, J9VM_J9CLASS_FROM_HEAPCLASS(vmThread, ramCPEntry->exception))) {
+			/* Resolved with exception */
+			VM_VMHelpers::setExceptionPending(vmThread, ramCPEntry->exception);
+			resolved = 1;
+		} else if (ramCPEntry->exception == vmThread->javaVM->voidReflectClass->classObject) {
```
tajila: you are using javaVM multiple times, you can cache a copy in a local variable

runtime/vm/resolvesupport.cpp
```diff
@@ -1839,6 +1839,102 @@ resolveMethodHandleRef(J9VMThread *vmThread, J9ConstantPool *ramCP, UDATA cpInde
...
+
+		/* Walk bsmData */
+		for (U_32 i = 0; i < bsmIndex; i++) {
+			bsmData += bsmData[1] + 2;
+		}
+
+		/* Invoke BootStrap method*/
+		sendResolveConstantDynamic(vmThread, ramCP, cpIndex, nameAndSig, bsmData);
+		value = (j9object_t)vmThread->returnValue;
+
```
tajila: better to move the `(ramCPEntry->exception != vmThread->threadObject)` check here, no point even running the rest of the code if someone else has resolved it

probably need read barrier before the check

runtime/vm/resolvesupport.cpp
```diff
@@ -1839,6 +1839,102 @@ resolveMethodHandleRef(J9VMThread *vmThread, J9ConstantPool *ramCP, UDATA cpInde
...
+		if (isSameOrSuperClassOf(throwable, J9VM_J9CLASS_FROM_HEAPCLASS(vmThread, ramCPEntry->exception))) {
+			/* Resolved with exception */
+			VM_VMHelpers::setExceptionPending(vmThread, ramCPEntry->exception);
+			resolved = 1;
+		} else if (ramCPEntry->exception == vmThread->javaVM->voidReflectClass->classObject) {
+			/* Resolved with null reference */
+			resolved = 1;
+		} else if (ramCPEntry->exception != vmThread->threadObject) {
+			/* Another thread is currently trying to resolve the entry */
+			// yield(vmThread);
```
tajila: VM_AtomicSupport::yieldCPU();

runtime/vm/resolvesupport.cpp
```diff
@@ -1839,6 +1839,102 @@ resolveMethodHandleRef(J9VMThread *vmThread, J9ConstantPool *ramCP, UDATA cpInde
...
+				setHeapOutOfMemoryError(vmThread);
+				exceptionObject = vmThread->outOfMemoryError;
+			} else if (ramCPEntry->exception != vmThread->threadObject) {
+				goto retry;
+			} else {
+				ramCPEntry->value = value;
+			}
+		}
+
+		if (ramCPEntry->exception == vmThread->threadObject) {
```
tajila: in order to reach this point, doesn't this check always have to be true?

runtime/vm/resolvesupport.cpp
```diff
@@ -1839,6 +1839,102 @@ resolveMethodHandleRef(J9VMThread *vmThread, J9ConstantPool *ramCP, UDATA cpInde
...
+			ramCPEntry->exception = exceptionObject;
```
tajila: might need a write barrier before this. I'm not sure how much reordering the compiler will do, but if it is possible to reoder this before the write to `ramCPEntry->value`, then you definitely need a write barrier.

runtime/vm/resolvesupport.cpp
tajila: Actually, yield my not be sufficient to ensure that the resolving thread gets precedence.

You can add a monitor (ie. condyMonitor) to each j9vmthread. Then the thread can wait on `((J9VMThread *)J9VMJAVALANGTHREAD_THREADREF(ramCPEntry->exception))->condyMonitor`.

And the code that writes value can notify condyMonitor

runtime/vm/resolvesupport.cpp
fengxue-IS: in the case of nested condy, the inner most call could resolve before reaching here

runtime/vm/resolvesupport.cpp
fengxue-IS: fixed

jcl/src/java.base/share/classes/java/lang/invoke/MethodHandle.java
fengxue-IS: fixed

runtime/vm/resolvesupport.cpp
```diff
@@ -1839,6 +1839,101 @@ resolveMethodHandleRef(J9VMThread *vmThread, J9ConstantPool *ramCP, UDATA cpInde
...
+		if (isSameOrSuperClassOf(throwable, J9VM_J9CLASS_FROM_HEAPCLASS(vmThread, ramCPEntry->exception))) {
+			/* Resolved with exception */
+			VM_VMHelpers::setExceptionPending(vmThread, ramCPEntry->exception);
+			resolved = TRUE;
+		} else if (ramCPEntry->exception == vm->voidReflectClass->classObject) {
+			/* Resolved with null reference */
+			resolved = TRUE;
+		} else if (ramCPEntry->exception != vmThread->threadObject) {
+			/* Another thread is currently trying to resolve the entry */
+			VM_AtomicSupport::yieldCPU();
```
fengxue-IS: @DanHeidinga @gacholio In the case that another thread is currently resolving the entry, would thread yield be sufficient to wait until the resolving thread finish?

gacholio: No, yield is essentially a no-op - it guarantees nothing.

fengxue-IS: sry, I meant if yield is used here, would it be sufficient (as in suspending the thread for short period) so it will not be consistently spinning in the `goto retry;` loop as other thread is resolving the condy entry. (so we don't have this thread wasting resources while not doing any actual work, a mutex would be ideal but too expensive?)

DanHeidinga:
@fengxue-IS Tobi, GAC, and I sketched out a possible resolve algorithm.  We should sit down together and go over it to get your input and then update this PR.

tajila:
We might want to think about making a common helper for `MethodHandle.resolve[Constant|Invoke]Dynamic` and `sendResolve[Invoke|Constant}Dynamic` in a follow on PR. Rest lgtm.

@gacholio please take a look

DanHeidinga:
@gacholio Can you review this?

gacholio:
I'd like to see @DanHeidinga approve first, and the conflicts resolved.

fengxue-IS:
Rebased on master branch, conflicts resolved. All the review comments have been addressed.
Tests pending

gacholio:
I'll review this today or tomorrow.  @DanHeidinga can approve when he gets a chance.

runtime/vm/resolvesupport.cpp
```diff
@@ -1847,6 +1847,102 @@ resolveMethodHandleRef(J9VMThread *vmThread, J9ConstantPool *ramCP, UDATA cpInde
 
 }
 
+j9object_t
+resolveConstantDynamic(J9VMThread *vmThread, J9ConstantPool *ramCP, UDATA cpIndex, UDATA resolveFlags)
+{
+	J9RAMConstantDynamicRef *ramCPEntry = (J9RAMConstantDynamicRef*)ramCP + cpIndex;
+	j9object_t value = ramCPEntry->value;
+	J9JavaVM *vm = vmThread->javaVM;
+	BOOLEAN resolved = FALSE;
```
gacholio: This is a cpp file, so please use bool.

runtime/vm/resolvesupport.cpp
```diff
@@ -1847,6 +1847,102 @@ resolveMethodHandleRef(J9VMThread *vmThread, J9ConstantPool *ramCP, UDATA cpInde
...
+		if (isSameOrSuperClassOf(throwable, J9VM_J9CLASS_FROM_HEAPCLASS(vmThread, ramCPEntry->exception))) {
+			/* Resolved with exception */
+			VM_VMHelpers::setExceptionPending(vmThread, ramCPEntry->exception);
+			resolved = TRUE;
+		} else if (ramCPEntry->exception == vm->voidReflectClass->classObject) {
+			/* Resolved with null reference */
+			resolved = TRUE;
+		} else if (ramCPEntry->exception != vmThread->threadObject) {
+			/* Another thread is currently trying to resolve the entry */
+			omrthread_monitor_wait(vm->constantDynamicMutex);
```
gacholio: I see no indication the monitor is ever entered or exitted.

runtime/vm/callin.cpp
```diff
@@ -941,6 +941,51 @@ sendForGenericInvoke(J9VMThread *currentThread, j9object_t methodHandle, j9objec
...
+		j9object_t nameString = mmFuncs->j9gc_createJavaLangString(currentThread, J9UTF8_DATA(nameUTF), J9UTF8_LENGTH(nameUTF), 0);
+		if (NULL != nameString) {
+			J9UTF8 *sigUTF = J9ROMNAMEANDSIGNATURE_SIGNATURE(nameAndSig);
+			PUSH_OBJECT_IN_SPECIAL_FRAME(currentThread, nameString);
+			j9object_t sigString = mmFuncs->j9gc_createJavaLangString(currentThread, J9UTF8_DATA(sigUTF), J9UTF8_LENGTH(sigUTF), 0);
+			nameString = POP_OBJECT_IN_SPECIAL_FRAME(currentThread);
+			if (NULL != sigString) {
+				/* Run the method */
+				/* skip one slot because we are passing a long */
+				currentThread->sp -= 2;
```
gacholio: Pleaase move this after the long comment so it's beside the store.

fengxue-IS:
added `omrthread_monitor_enter/exit()` around `wait` and `notify_all` operation
fixed the comment position as suggested and changed `BOOLEAN` to `bool`

runtime/vm/resolvesupport.cpp
```diff
@@ -1847,6 +1847,106 @@ resolveMethodHandleRef(J9VMThread *vmThread, J9ConstantPool *ramCP, UDATA cpInde
...
+				exceptionObject = vmThread->outOfMemoryError;
+			} else {
+				ramCPEntry->value = value;
+			}
+		}
+
+		ramCPEntry->exception = exceptionObject;
+		omrthread_monitor_enter(vm->constantDynamicMutex);
+		omrthread_monitor_notify_all(vm->constantDynamicMutex);
+		omrthread_monitor_exit(vm->constantDynamicMutex);
```
gacholio: This is not a legal use of wait/notify.  Let's discuss tomorrow.

runtime/vm/resolvesupport.cpp
```diff
@@ -1847,6 +1847,106 @@ resolveMethodHandleRef(J9VMThread *vmThread, J9ConstantPool *ramCP, UDATA cpInde
...
+			VM_VMHelpers::setExceptionPending(vmThread, ramCPEntry->exception);
+			resolved = true;
+		} else if (ramCPEntry->exception == vm->voidReflectClass->classObject) {
+			/* Resolved with null reference */
+			resolved = true;
+		} else if (ramCPEntry->exception != vmThread->threadObject) {
+			/* Another thread is currently trying to resolve the entry */
+			omrthread_monitor_enter(vm->constantDynamicMutex);
+			omrthread_monitor_wait(vm->constantDynamicMutex);
+			omrthread_monitor_exit(vm->constantDynamicMutex);
```
gacholio: Wait should not be used bare like this - correct use is to wait for a condition to become true.

fengxue-IS:
Update the monitor entrance and exit around `omrthread_monitor_wait` to make sure that code is waiting on a looping check.
added fast path in resolve code to optimize performance.
fixed an error with return value not properly set.

runtime/vm/resolvesupport.cpp
gacholio: It would be better to stash the two written values in locals, and update the ramCPEntry while holding the monitor.

fengxue-IS: I know it is a better practice to update the ramCPEntry within the monitor, but not sure how much overhead this would add in a multi-thread case when parallel resolve takes place

gacholio: None at all - it's the same code (modulo a few local variable accesses), just shuffled around.

fengxue-IS: I've made the change to update the cp entry values inside the monitor scope, we can change this in the future if this becomes a performance issue

runtime/vm/resolvesupport.cpp
```diff
@@ -1847,6 +1847,113 @@ resolveMethodHandleRef(J9VMThread *vmThread, J9ConstantPool *ramCP, UDATA cpInde
...
+		j9object_t exceptionObject = NULL;
+		if (NULL != vmThread->currentException) {
+			/* Allocate the exception object in tenure */
+			exceptionObject = gcFuncs->j9gc_objaccess_asConstantPoolObject(
+											vmThread,
+											vmThread->currentException,
+											J9_GC_ALLOCATE_OBJECT_TENURED | J9_GC_ALLOCATE_OBJECT_NON_INSTRUMENTABLE);
+			if (NULL == exceptionObject) {
+				setHeapOutOfMemoryError(vmThread);
+				exceptionObject = vmThread->outOfMemoryError;
```
gacholio: I believe this should be vmThread->currentException.

runtime/vm/resolvesupport.cpp
```diff
@@ -1847,6 +1847,113 @@ resolveMethodHandleRef(J9VMThread *vmThread, J9ConstantPool *ramCP, UDATA cpInde
...
+		/* Slow path resolve */
+		omrthread_monitor_enter(vm->constantDynamicMutex);
+retry:
+		/* Check if already resolved */
+		if (NULL != ramCPEntry->value) {
+			value = ramCPEntry->value;
+			resolved = true;
+		} else if (NULL != ramCPEntry->exception) {
+			J9Class *throwable = J9VMJAVALANGTHROWABLE_OR_NULL(vm);
+			if (isSameOrSuperClassOf(throwable, J9VM_J9CLASS_FROM_HEAPCLASS(vmThread, ramCPEntry->exception))) {
```
gacholio: J9VM_J9CLASS_... works on class objects - if exception is NULL, this will crash, and if it's a thread object, the results are undefined at best.  At the very least, the threadObject check needs to be first.

runtime/vm/resolvesupport.cpp
```diff
@@ -1847,6 +1847,113 @@ resolveMethodHandleRef(J9VMThread *vmThread, J9ConstantPool *ramCP, UDATA cpInde
...
+				/* Resolved with exception */
+				VM_VMHelpers::setExceptionPending(vmThread, ramCPEntry->exception);
+				resolved = true;
+			} else if (ramCPEntry->exception == vm->voidReflectClass->classObject) {
+				/* Resolved with null reference */
+				resolved = true;
```
gacholio: What happens when the BSM returns the void class object?

gacholio: Nevermind, this is exception not value.

runtime/vm/resolvesupport.cpp
```diff
@@ -1851,6 +1851,122 @@ resolveMethodHandleRef(J9VMThread *vmThread, J9ConstantPool *ramCP, UDATA cpInde
...
+			} else if (isSameOrSuperClassOf(throwable, clazz)) {
+				/* Resolved with exception */
+				VM_VMHelpers::setExceptionPending(vmThread, ramCPEntry->exception);
+				resolved = true;
+			}
+		}
+	}
+
+	if (resolved) {
+		return value;
```
gacholio: Multiple returns are discouraged.  Please use if (!resolved) instead of if/return/else.

runtime/vm/resolvesupport.cpp
```diff
@@ -1851,6 +1851,122 @@ resolveMethodHandleRef(J9VMThread *vmThread, J9ConstantPool *ramCP, UDATA cpInde
...
+	} else {
+		/* Slow path, attempt to acquire resolve mutex or retry, or wait */
+		omrthread_monitor_enter(vm->constantDynamicMutex);
+		if (NULL == ramCPEntry->value && NULL == ramCPEntry->exception) {
```
gacholio: After entering the monitor, you should do all of the checks from before the monitor enter again, like:

enter monitor
checks
if (!resolved) {
}
exit monitor
...fall through to return...

runtime/vm/resolvesupport.cpp
fengxue-IS: fixed

runtime/vm/resolvesupport.cpp
fengxue-IS: inside the monitor it have a simplified block of checks that determine if there have been any update to the cpEntry and redirects with `goto retry;`

gacholio: I'm not convinced it's correct, and there's no point trying to optimize the slow path instead of keeping the code the same as outside for understandability.

runtime/vm/resolvesupport.cpp
```diff
@@ -1851,6 +1851,129 @@ resolveMethodHandleRef(J9VMThread *vmThread, J9ConstantPool *ramCP, UDATA cpInde
...
+					omrthread_monitor_exit(vm->constantDynamicMutex);
+					// internalAcquireVMAccess(vmThread);
+					goto retry;
+				}
+			}
+		}
+		ramCPEntry->exception = vmThread->threadObject;
+		omrthread_monitor_exit(vm->constantDynamicMutex);
+
+		if (!resolved) {
```
gacholio: Can resolved be false at this point?

runtime/vm/resolvesupport.cpp
```diff
@@ -1851,6 +1851,129 @@ resolveMethodHandleRef(J9VMThread *vmThread, J9ConstantPool *ramCP, UDATA cpInde
...
+					// internalReleaseVMAccess(vmThread);
+					omrthread_monitor_wait(vm->constantDynamicMutex);
+					omrthread_monitor_exit(vm->constantDynamicMutex);
+					// internalAcquireVMAccess(vmThread);
+					goto retry;
+				}
+			}
+		}
+		ramCPEntry->exception = vmThread->threadObject;
+		omrthread_monitor_exit(vm->constantDynamicMutex);
```
gacholio: Don't exit here - just do it at the end in both the resolved and unresolved paths.

runtime/vm/resolvesupport.cpp
```diff
@@ -1851,6 +1851,129 @@ resolveMethodHandleRef(J9VMThread *vmThread, J9ConstantPool *ramCP, UDATA cpInde
...
+					/* Another thread is currently resolving this entry */
+					// internalReleaseVMAccess(vmThread);
+					omrthread_monitor_wait(vm->constantDynamicMutex);
+					omrthread_monitor_exit(vm->constantDynamicMutex);
+					// internalAcquireVMAccess(vmThread);
+					goto retry;
+				}
+			}
+		}
+		ramCPEntry->exception = vmThread->threadObject;
```
gacholio: You only want to do this in the unresolved case.

runtime/vm/resolvesupport.cpp
```diff
@@ -1851,6 +1851,129 @@ resolveMethodHandleRef(J9VMThread *vmThread, J9ConstantPool *ramCP, UDATA cpInde
...
+			}
+		}
+	}
+
+	if (!resolved) {
+		/* Slow path, attempt to acquire resolve mutex or retry, or wait */
+		omrthread_monitor_enter(vm->constantDynamicMutex);
+		if (NULL != ramCPEntry->value) {
+			value = ramCPEntry->value;
+			resolved = true;
```
gacholio: These might be better done with a retry bool instead of resolved - detect the cases that means you don't need to resolve, and go back to retry to get the value/set the exception.

runtime/vm/resolvesupport.cpp
fengxue-IS: yes, if this is the first time resolving or nested resolves

gacholio: Right, misssed the NULL check on exception.

runtime/vm/resolvesupport.cpp
fengxue-IS: not sure what you meant, this is the end of all checks, anything after this point would be the actual resolution

runtime/vm/resolvesupport.cpp
fengxue-IS: updated

runtime/vm/resolvesupport.cpp
```diff
@@ -1851,6 +1851,131 @@ resolveMethodHandleRef(J9VMThread *vmThread, J9ConstantPool *ramCP, UDATA cpInde
...
+					/* Resolved with null reference */
+					retry = true;
+				} else if (isSameOrSuperClassOf(throwable, clazz)) {
+					/* Resolved with exception */
+					retry = true;
+				} else if (ramCPEntry->exception != vmThread->threadObject) {
+					/* Another thread is currently resolving this entry */
+					// internalReleaseVMAccess(vmThread);
+					omrthread_monitor_wait(vm->constantDynamicMutex);
+					omrthread_monitor_exit(vm->constantDynamicMutex);
```
gacholio: This will end up being exitting twice (below in the retry path).  Might be better to just exit and goto inline instead.

runtime/vm/resolvesupport.cpp
```diff
@@ -1851,6 +1851,131 @@ resolveMethodHandleRef(J9VMThread *vmThread, J9ConstantPool *ramCP, UDATA cpInde
...
+				J9Class *clazz = J9OBJECT_CLAZZ(vmThread, ramCPEntry->exception);
+				if (ramCPEntry->exception == vm->voidReflectClass->classObject) {
+					/* Resolved with null reference */
+					retry = true;
+				} else if (isSameOrSuperClassOf(throwable, clazz)) {
+					/* Resolved with exception */
+					retry = true;
+				} else if (ramCPEntry->exception != vmThread->threadObject) {
+					/* Another thread is currently resolving this entry */
+					// internalReleaseVMAccess(vmThread);
```
gacholio: This is definitely required.

runtime/vm/resolvesupport.cpp
fengxue-IS: I believe this exit is something I forgot to remove from last update, there are multiple retry paths, inlining all would make the code look bad. I ll remove this exit

runtime/vm/resolvesupport.cpp
fengxue-IS: uncommented

runtime/vm/resolvesupport.cpp
gacholio: You can't remove this one - you must exit the monitor before reacquiring VM access.

fengxue-IS: Right, I see what you meant
Corrected

runtime/vm/resolvesupport.cpp
```diff
@@ -1851,6 +1851,131 @@ resolveMethodHandleRef(J9VMThread *vmThread, J9ConstantPool *ramCP, UDATA cpInde
...
+					/* Another thread is currently resolving this entry */
+					internalReleaseVMAccess(vmThread);
+					omrthread_monitor_wait(vm->constantDynamicMutex);
+					omrthread_monitor_exit(vm->constantDynamicMutex);
+					internalAcquireVMAccess(vmThread);
+					goto RETRY;
+				}
+			}
+		}
+		if (retry) {
```
gacholio: Given that there's one goto, may as well use all gotos and elminate the new bool (and won't have to capitalize the label).  Don't forget the monitor exit in the other two cases.

runtime/vm/resolvesupport.cpp
```diff
@@ -1851,6 +1851,131 @@ resolveMethodHandleRef(J9VMThread *vmThread, J9ConstantPool *ramCP, UDATA cpInde
...
+
+		/* Check if entry resolved by nested constantDynamic calls */
+		if (ramCPEntry->exception != vmThread->threadObject) {
+			goto RETRY;
+		}
+
+		j9object_t exceptionObject = NULL;
+		if (NULL != vmThread->currentException) {
+			/* Allocate the exception object in tenure */
+			exceptionObject = gcFuncs->j9gc_objaccess_asConstantPoolObject(
```
gacholio: Dan points out that this is illegal.  The returned constant must be in the constant pool, not a tenured copy (which will not be identical).  The GC must first be updated to walk the constant pools in all policies, perhaps with an optimization tagging classes as possibly containing untenured CP objects.

runtime/vm/resolvesupport.cpp
```diff
@@ -1851,6 +1851,131 @@ resolveMethodHandleRef(J9VMThread *vmThread, J9ConstantPool *ramCP, UDATA cpInde
...
+	bool resolved = false;
+
+RETRY:
+	/* Fast path resolution for previously resolved entries */
+	if (NULL != ramCPEntry->value) {
+		value = ramCPEntry->value;
+		resolved = true;
+	} else {
+		/* check if already resolved to NULL or exception */
+		if (NULL != ramCPEntry->exception) {
```
gacholio: Please cache this in a local rather than reading it multiple times.

runtime/vm/resolvesupport.cpp
```diff
@@ -1851,6 +1851,131 @@ resolveMethodHandleRef(J9VMThread *vmThread, J9ConstantPool *ramCP, UDATA cpInde
...
+resolveConstantDynamic(J9VMThread *vmThread, J9ConstantPool *ramCP, UDATA cpIndex, UDATA resolveFlags)
+{
+	J9RAMConstantDynamicRef *ramCPEntry = (J9RAMConstantDynamicRef*)ramCP + cpIndex;
+	j9object_t value = NULL;
+	J9JavaVM *vm = vmThread->javaVM;
+	bool resolved = false;
+
+RETRY:
+	/* Fast path resolution for previously resolved entries */
+	if (NULL != ramCPEntry->value) {
```
gacholio: This is the expected case, so could read into value, and NULL check the local.  Also means value could be declared and initialized directly here.

runtime/vm/resolvesupport.cpp
```diff
@@ -1851,6 +1851,131 @@ resolveMethodHandleRef(J9VMThread *vmThread, J9ConstantPool *ramCP, UDATA cpInde
...
+
+	if (!resolved) {
+		/* Slow path, attempt to acquire resolve mutex or retry, or wait */
+		omrthread_monitor_enter(vm->constantDynamicMutex);
+		bool retry = false;
+		if (NULL != ramCPEntry->value) {
+			retry = true;
+		} else {
+			/* check if already resolved to NULL or exception */
+			if (NULL != ramCPEntry->exception) {
```
gacholio: Same multiple read comment as above.

fengxue-IS:
Updated the resolve check to use inline goto and local variables for value and exception, pending gc allocation until discussed with GC team

fengxue-IS:
fixed a typo in the code, updated resolved value/exception storing to use the original result object returned from the bootstrap method. the GC constant pool slot iterator will be updated in a separate PR

fengxue-IS:
@DanHeidinga can I get a re-review on this? The GC constant pool iterator change will be addressed in  #2457

runtime/oti/j9nonbuilder.h
```diff
@@ -252,6 +252,10 @@
 #define J9DescriptionCpTypeShift 0x4
 #define J9DescriptionImmediate 0x1
 
+/* Rom CP Flag for constant dynamic with primitive type */
+#define J9DescriptionCpPrimitiveType 1 << 20
```
DanHeidinga: Please add brackets to this.  Also, why 20?

runtime/oti/j9nonbuilder.h
```diff
@@ -252,6 +252,10 @@
...
+#define J9DescriptionCpBsmIndexMask 0xFFFF
```
DanHeidinga: Is this constant correct if the top bit is used for primitives?

runtime/oti/vmconstantpool.xml
```diff
@@ -339,6 +339,7 @@
 	<staticmethodref class="java/lang/invoke/MethodType" name="vmResolveFromMethodDescriptorString" signature="(Ljava/lang/String;Ljava/lang/ClassLoader;Ljava/lang/Class;)Ljava/lang/invoke/MethodType;" flags="opt_methodHandle"/>
 	<staticmethodref class="java/lang/invoke/MethodHandle" name="sendResolveMethodHandle" signature="(ILjava/lang/Class;Ljava/lang/Class;Ljava/lang/String;Ljava/lang/String;Ljava/lang/ClassLoader;)Ljava/lang/invoke/MethodHandle;" flags="opt_methodHandle"/> 
 	<staticmethodref class="java/lang/invoke/MethodHandle" name="resolveInvokeDynamic" signature="(JLjava/lang/String;Ljava/lang/String;J)Ljava/lang/invoke/MethodHandle;" flags="opt_methodHandle"/> 
+	<staticmethodref class="java/lang/invoke/MethodHandle" name="resolveConstantDynamic" signature="(JLjava/lang/String;Ljava/lang/String;J)Ljava/lang/Object;" flag="opt_methodHandle"/>
```
DanHeidinga: That should include a JDK11 or similar reference as this function doesn't apply to JDK 10 & under.

runtime/vm/BytecodeInterpreter.hpp
```diff
@@ -4568,11 +4568,39 @@ done:;
...
 	{
 		U_16 index = *(U_16*)(_pc + 1);
-		J9ROMConstantRef *romConstant = (J9ROMConstantRef*)romConstantPoolItem(_literals, index);
-		_pc += 3;
-		_sp -= 2;
-		((U_32*)_sp)[0] = romConstant->slot1;
-		((U_32*)_sp)[1] = romConstant->slot2;
+
+		if (J9CPTYPE_CONSTANT_DYNAMIC == J9_CP_TYPE(J9ROMCLASS_CPSHAPEDESCRIPTION(J9_CLASS_FROM_METHOD(_literals)->romClass), index)) {
+			J9ConstantPool *ramConstantPool = J9_CP_FROM_METHOD(_literals);
```
DanHeidinga: Move this outside the if so it's not fetched twice.

runtime/vm/BytecodeInterpreter.hpp
```diff
@@ -4568,11 +4568,39 @@ done:;
...
+				if (VM_VMHelpers::exceptionPending(_currentThread)) {
+					return GOTO_THROW_CURRENT_EXCEPTION;
+				}
+			}
+
+			_pc += 3;
+			_sp -= 2;
+			U_64 value = J9VMJAVALANGLONG_VALUE(_currentThread, resolvedValue);
+			*(U_64*)_sp = value;
+
```
DanHeidinga: unnecessary blank line

runtime/vm/BytecodeInterpreter.hpp
```diff
@@ -4568,11 +4568,39 @@ done:;
...
+			if (NULL == resolvedValue) {
+				buildGenericSpecialStackFrame(REGISTER_ARGS, 0);
+				updateVMStruct(REGISTER_ARGS);
+
+				resolvedValue = resolveConstantDynamic(_currentThread, ramConstantPool, index, J9_RESOLVE_FLAG_RUNTIME_RESOLVE);
+		
+				VMStructHasBeenUpdated(REGISTER_ARGS);
+				restoreGenericSpecialStackFrame(REGISTER_ARGS);
+
```
DanHeidinga: Please add the immediate async check - see other resolves for an example

runtime/vm/BytecodeInterpreter.hpp
```diff
@@ -4568,11 +4568,39 @@ done:;
...
+				if (VM_VMHelpers::exceptionPending(_currentThread)) {
+					return GOTO_THROW_CURRENT_EXCEPTION;
```
DanHeidinga: Only return from a single point

runtime/vm/BytecodeInterpreter.hpp
```diff
@@ -6032,6 +6060,12 @@ done:;
 			case J9DescriptionCpTypeMethodHandle:
 				resolveMethodHandleRef(_currentThread, ramConstantPool, index, J9_RESOLVE_FLAG_RUNTIME_RESOLVE | J9_RESOLVE_FLAG_NO_CLASS_INIT);
 				break;
+			case J9DescriptionCpTypeConstantDynamic:
+				resolveConstantDynamic(_currentThread, ramConstantPool, index, J9_RESOLVE_FLAG_RUNTIME_RESOLVE);
+				if (((J9RAMConstantDynamicRef*)ramCPEntry)->exception == _vm->voidReflectClass->classObject) {
```
DanHeidinga: This needs a comment to describe why it's correct

runtime/vm/BytecodeInterpreter.hpp
```diff
@@ -6047,10 +6081,18 @@ done:;
...
 			goto retry;
 		}
+resolved:
 		_pc += (1 + parmSize);
 		_sp -= 1;
 		if (J9DescriptionCpTypeClass == romCPEntry->cpType) {
 			value = J9VM_J9CLASS_TO_HEAPCLASS((J9Class*)value);
+		} else if ((J9DescriptionCpTypeConstantDynamic == romCPEntry->cpType & J9DescriptionCpTypeMask)
+			&& (J9_ARE_ALL_BITS_SET(romCPEntry->cpType, J9DescriptionCpPrimitiveType))
+		){
```
DanHeidinga: nitpick `){` -> `) {`

runtime/vm/BytecodeInterpreter.hpp
```diff
@@ -6047,10 +6081,18 @@ done:;
...
+			/* Constant Dynamic ROM CP entry uses J9DescriptionCpPrimitiveType flag to
+			 * indicate primitive return type which requires unboxing before returning the value
+			 */
+			value = (j9object_t)(UDATA)J9VMJAVALANGINTEGER_VALUE(_currentThread, value);
```
DanHeidinga: What about the other java.lang.* primitive classes like Char, Short, etc?  The shapes could be different.

runtime/vm/BytecodeInterpreter.hpp
DanHeidinga: We should do the Void.class check first and let the unresolved case go around the retry loop again.
Saves a call to `resolveConstantDynamic` on every condy of null

runtime/vm/callin.cpp
```diff
@@ -941,6 +941,53 @@ sendForGenericInvoke(J9VMThread *currentThread, j9object_t methodHandle, j9objec
 	Trc_VM_sendForGenericInvoke_Exit(currentThread);
 }
 
+void JNICALL
+sendResolveConstantDynamic(J9VMThread *currentThread, J9ConstantPool *ramCP, UDATA cpIndex, J9ROMNameAndSignature *nameAndSig, U_16 *bsmData)
+{
+	Trc_VM_sendResolveConstantDynamic_Entry(currentThread);
```
DanHeidinga: please include the ramCP, cpIndex, nas & bsmData in the entry tracepoint

runtime/vm/resolvesupport.cpp
```diff
@@ -1869,6 +1869,115 @@ resolveMethodHandleRef(J9VMThread *vmThread, J9ConstantPool *ramCP, UDATA cpInde
...
+	j9object_t cpException = NULL;
+	bool resolved = false;
+
+retry:
+	value = ramCPEntry->value;
+	/* Fast path resolution for previously resolved entries */
+	if (NULL != value) {
+		resolved = true;
+	} else {
+		cpException = ramCPEntry->exception;
```
DanHeidinga: move the declaration of cpException into the two scopes that use it.

runtime/vm/resolvesupport.cpp
```diff
@@ -1869,6 +1869,115 @@ resolveMethodHandleRef(J9VMThread *vmThread, J9ConstantPool *ramCP, UDATA cpInde
...
+					/* Another thread is currently resolving this entry */
+					internalReleaseVMAccess(vmThread);
+					omrthread_monitor_wait(vm->constantDynamicMutex);
+					omrthread_monitor_exit(vm->constantDynamicMutex);
+					internalAcquireVMAccess(vmThread);
+					goto retry;
+				}
+			}
+		}
+
```
DanHeidinga: add comment to indicate this thread is the one doing the resolve

runtime/vm/resolvesupport.cpp
```diff
@@ -1869,6 +1869,115 @@ resolveMethodHandleRef(J9VMThread *vmThread, J9ConstantPool *ramCP, UDATA cpInde
...
+		omrthread_monitor_exit(vm->constantDynamicMutex);
+
+		/* Enter if not previously resolved */
+		J9Class *ramClass = ramCP->ramClass;
+		J9ROMClass *romClass = ramClass->romClass;
+		J9ROMConstantDynamicRef *romConstantRef = (J9ROMConstantDynamicRef*)(J9_ROM_CP_FROM_CP(ramCP) + cpIndex);
+		J9SRP *callSiteData = (J9SRP *) J9ROMCLASS_CALLSITEDATA(romClass);
+		U_16 *bsmIndices = (U_16 *) (callSiteData + romClass->callSiteCount);
+		U_16 *bsmData = bsmIndices + romClass->callSiteCount;
+		U_32 bsmIndex = (romConstantRef->bsmIndexAndCpType >> J9DescriptionCpTypeShift) & J9DescriptionCpBsmIndexMask;
```
DanHeidinga: bsmIndex needs to take the `J9DescriptionCpPrimitiveType` into account

runtime/vm/resolvesupport.cpp
```diff
@@ -1869,6 +1869,115 @@ resolveMethodHandleRef(J9VMThread *vmThread, J9ConstantPool *ramCP, UDATA cpInde
...
+		J9ROMNameAndSignature* nameAndSig = SRP_PTR_GET(&romConstantRef->nameAndSignature, J9ROMNameAndSignature*);
+
+		/* Walk bsmData */
```
DanHeidinga: update the comment to describe this in more detail

runtime/vm/resolvesupport.cpp
```diff
@@ -1869,6 +1869,115 @@ resolveMethodHandleRef(J9VMThread *vmThread, J9ConstantPool *ramCP, UDATA cpInde
...
+		for (U_32 i = 0; i < bsmIndex; i++) {
+			bsmData += bsmData[1] + 2;
```
DanHeidinga: Please add brackets around the RHS

runtime/vm/resolvesupport.cpp
```diff
@@ -1869,6 +1869,115 @@ resolveMethodHandleRef(J9VMThread *vmThread, J9ConstantPool *ramCP, UDATA cpInde
...
+
+		/* Check if entry resolved by nested constantDynamic calls */
+		if (ramCPEntry->exception != vmThread->threadObject) {
+			goto retry;
+		}
+
+		j9object_t exceptionObject = NULL;
+		if (NULL != vmThread->currentException) {
+			/* Allocate the exception object in tenure */
+			exceptionObject = vmThread->currentException;
```
DanHeidinga: The comment and code are inconsistent.

runtime/oti/j9nonbuilder.h
fengxue-IS: the slot is formatted as
12 bit unused | 16 bit bsmIndex | 4 bit Cp descriptor
20 is the least significant bit unused

runtime/oti/j9nonbuilder.h
fengxue-IS: this is mask is suppose to get rid of the primitive bit.
`bsmIndex = (bsmIndexAndCpType >> J9DescriptionCpTypeShift) & J9DescriptionCpBsmIndexMask`
so we only get the U_16 value for the bsmIndex

runtime/bcutil/ConstantPoolMap.cpp
```diff
@@ -503,7 +503,20 @@ ConstantPoolMap::constantPoolDo(ConstantPoolVisitor *visitor)
...
 				break;
 			case J9CPTYPE_CONSTANT_DYNAMIC:
-				visitor->visitConstantDynamic(slot1, slot2);
+				/* Check if the return type of constant dynamic entry is a primitive type
+				 * Set the J9DescriptionCpPrimitiveType flag so interpreter know to unbox
+				 * the resolved object before returning it.
+				 */
+				{
+					char fieldDescriptor = (char)*_classFileOracle->getUTF8Data(getCPSlot2(cfrCPIndex));
+					if (fieldDescriptor == 'B' || fieldDescriptor == 'C' || fieldDescriptor == 'F' ||
```
DanHeidinga: Can this be written as a switch statement?  I think it would be clearer than the repeated ifs

runtime/oti/j9nonbuilder.h
DanHeidinga: Good info to put into the code

jcl/src/java.base/share/classes/java/lang/invoke/MethodHandle.java
```diff
@@ -828,6 +849,154 @@ MethodHandle insertArguments(MethodHandle equivalent, MethodHandle unboxingHandl
...
+			VMLangAccess access = VM.getVMLangAccess();
+			Object internalRamClass = access.createInternalRamClass(j9class);
+			Class<?> classObject = null;
+			if (JITHELPERS.is32Bit()) {
+				classObject = JITHELPERS.getClassFromJ9Class32((int)j9class);
+			} else {
+				classObject = JITHELPERS.getClassFromJ9Class64(j9class);
+			}
+
+			Objects.requireNonNull(classObject);
```
DanHeidinga: Please add a message to this throwable to indicate the j9class, name, etc so that issues can more easily be debugged

jcl/src/java.base/share/classes/java/lang/invoke/MethodHandle.java
```diff
@@ -828,6 +849,154 @@ MethodHandle insertArguments(MethodHandle equivalent, MethodHandle unboxingHandl
...
+			default:
+				result = (Object) bsm.invokeWithArguments(staticArgs);
+				break;
+			}
+
+			// result validation
+			result = MethodHandles.identity(typeClass).invoke(result);
+		} catch(Throwable e) {
+
+			/*[IF Sidecar19-SE]*/
```
DanHeidinga: none of this code is Java 9 specific - condy is 11+

jcl/src/java.base/share/classes/java/lang/invoke/MethodHandle.java
```diff
@@ -828,6 +849,154 @@ MethodHandle insertArguments(MethodHandle equivalent, MethodHandle unboxingHandl
...
+					cpEntry = cp.getDoubleAt(index);
+					break;
+				case 13:
+					cpEntry = getCPMethodTypeAt(internalRamClass, index);
+					break;
+				case 14:
+					cpEntry = getCPMethodHandleAt(internalRamClass, index);
+					break;
+				case 17:
+					// TODO - cpEntry = getCPConstantDynamicAt(internalRamClass, index);
```
DanHeidinga: please throw an InternalError for this case if we're not ready to handle it

jcl/src/java.base/share/classes/java/lang/invoke/MethodHandle.java
```diff
@@ -788,6 +789,26 @@ MethodHandle insertArguments(MethodHandle equivalent, MethodHandle unboxingHandl
...
+		}
+
+		char[] signature = new char[length];
+		fieldDescriptor.getChars(0, length, signature, 0);
+
+		MethodType.parseIntoClass(signature, 0, classList, classLoader, fieldDescriptor);
+
+		if (classList.isEmpty()) {
+			throw new NoClassDefFoundError(fieldDescriptor);
+		}
```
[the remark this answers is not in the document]
fengxue-IS: since parseIntoClass will only parse one valid descriptor and append that to the list or throw an exception, I don't think this check is even needed.

I'll remove the check

jcl/src/java.base/share/classes/java/lang/invoke/MethodHandle.java
fengxue-IS: updated

jcl/src/java.base/share/classes/java/lang/invoke/MethodHandle.java
```diff
@@ -925,6 +1094,9 @@ private static final MethodHandle resolveInvokeDynamic(long j9class, String name
 				case 14:
 					cpEntry = getCPMethodHandleAt(internalRamClass, index);
 					break;
+				case 17:
+					// TODO - cpEntry = getCPConstantDynamicAt(internalRamClass, index);
```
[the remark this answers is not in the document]
fengxue-IS: added

runtime/vm/BytecodeInterpreter.hpp
fengxue-IS: async check added

runtime/vm/resolvesupport.cpp
fengxue-IS: comment added

jcl/src/java.base/share/classes/java/lang/invoke/MethodHandle.java
fengxue-IS: added ExternalMessage in the catch block

runtime/bcutil/ConstantPoolMap.cpp
```diff
@@ -508,13 +508,35 @@ ConstantPoolMap::constantPoolDo(ConstantPoolVisitor *visitor)
...
-					char fieldDescriptor = (char)*_classFileOracle->getUTF8Data(getCPSlot2(cfrCPIndex));
-					if (fieldDescriptor == 'B' || fieldDescriptor == 'C' || fieldDescriptor == 'F' ||
-						fieldDescriptor == 'I' || fieldDescriptor == 'S' || fieldDescriptor == 'Z')
-					{
-						visitor->visitConstantDynamic(slot1, slot2, J9DescriptionCpPrimitiveType);
-					} else {
-						visitor->visitConstantDynamic(slot1, slot2, 0);
+					char fieldDescriptor = (char)*_classFileOracle->getUTF8Data(getCPSlot2(slot2));
+					switch (fieldDescriptor) {
+						case 'B':
```
DanHeidinga: nitpick: don't indent the case labels.  Write as:
```c
switch(fieldDescriptor) {
case 'B':
      visitor-......
}
This saves over indenting the code

File: runtime/vm/BytecodeInterpreter.hpp [hunk 62]
```diff
@@ -4591,25 +4593,37 @@ done:;
...
 			_pc += 3;
 			_sp -= 2;
-			U_64 value = J9VMJAVALANGLONG_VALUE(_currentThread, resolvedValue);
-			*(U_64*)_sp = value;
 
+			/* Constant Dynamic ROM CP entry uses J9DescriptionReturnType* flag to indicate
+			 * different primitive return type that require unboxing before returning the value
+			 */
+			switch (((J9ROMConstantDynamicRef *)romConstant)->bsmIndexAndCpType >> J9DescriptionReturnTypeShift) {
+				case J9DescriptionReturnTypeDouble:
```
DanHeidinga: same nitpick about indenting case labels.

runtime/vm/BytecodeInterpreter.hpp
```diff
@@ -6094,13 +6110,30 @@ done:;
...
-		){
-			/* Constant Dynamic ROM CP entry uses J9DescriptionCpPrimitiveType flag to
-			 * indicate primitive return type which requires unboxing before returning the value
+		} else if (J9DescriptionCpTypeConstantDynamic == (romCPEntry->cpType & J9DescriptionCpTypeMask)) {
+			/* Constant Dynamic ROM CP entry uses J9DescriptionReturnType* flag to indicate
+			 * different primitive return type that require unboxing before returning the value
 			 */
-			value = (j9object_t)(UDATA)J9VMJAVALANGINTEGER_VALUE(_currentThread, value);
+			switch (romCPEntry->cpType >> J9DescriptionReturnTypeShift) {
+				case J9DescriptionReturnTypeBoolean:
```
DanHeidinga: indenting

runtime/vm/resolvesupport.cpp
```diff
@@ -1998,7 +2000,7 @@ resolveConstantDynamic(J9VMThread *vmThread, J9ConstantPool *ramCP, UDATA cpInde
 
 		j9object_t exceptionObject = NULL;
 		if (NULL != vmThread->currentException) {
-			/* Allocate the exception object in tenure */
+			/* Set exceptionObject to current exception */
```
DanHeidinga: nitpick: comment says exactly what the code does.  It can be deleted.

fengxue-IS:
fixed indentation and code comment.

DanHeidinga:
> Can you squash this and then I'll run the builds on it?

Never mind - the 3 commits are already reasonably organized.

DanHeidinga:

DanHeidinga:

DanHeidinga:

DanHeidinga:
Relaunching the test builds to CI issues over the weekend

DanHeidinga:

gacholio:
Extended testing is failing for silly reasons, so ignoring.

DanHeidinga:
Cancelling the JDK11 build as it fails due to missing test natives.  Merging as is as the JDK8 build passes and this will unblock more of the JDK11 work

PR #1534, opened 2018-03-23
remove JCK_ROOT from passing into ant

TianyuZuo:
* we do not need to pass in JCK_ROOT since it needs be
  exported as an environment variable if user needs this
  variable in ant script.

Signed-off-by: Tianyu Zuo <tianyu@ca.ibm.com>


TianyuZuo:
reivewer: @llxia 
reviewer: @smlambert 

https://github.com/AdoptOpenJDK/openjdk-tests/pull/353 is depending on this PR

smlambert:
Change does not warrant PR testing.

PR #1540, opened 2018-03-26
Cache java/lang/Class related queries results

cathyzhyi:
In a recent performance investigation getInterfaces, getDeclaringClass
and getEnclosingClass are found to be calling native and helper
methods too many times. This changeset cache the results for those
APIs to reduce the number of times calling native or helper methods.

### DIFF
--- a/jcl/src/java.base/share/classes/java/lang/Class.java
+++ b/jcl/src/java.base/share/classes/java/lang/Class.java
@@ -214,6 +214,24 @@ public final class Class<T> implements java.io.Serializable, GenericDeclaration,
 	private static boolean reflectCacheDebug;
 	private static boolean reflectCacheAppOnly = true;
 
+	/*
+	 * This {@code ClassReflectNullPlaceHolder} class is created to indicate the cached class value is
+	 * initialized to null rather than the default value null ;e.g. {@code cachedDeclaringClass}
+	 * and {@code cachedEnclosingClass}. The reason default value has to be null is that
+	 * j.l.Class instances are created by the VM only rather than Java, and
+	 * any instance field with non-null default values have to be set by VM natives.
+	 */
+	private static final class ClassReflectNullPlaceHolder {}
+
+	private transient Class<?>[] cachedInterfaces;
+	private static long cachedInterfacesOffset = -1;
+
+	private transient Class<?> cachedDeclaringClass;
+	private static long cachedDeclaringClassOffset = -1;
+
+	private transient Class<?> cachedEnclosingClass;
+	private static long cachedEnclosingClassOffset = -1;
+
 	private static Annotation[] EMPTY_ANNOTATION_ARRAY = new Annotation[0];
 	
 	static MethodHandles.Lookup implLookup;
@@ -282,6 +300,27 @@ private void checkNonSunProxyMemberAccess(SecurityManager security, ClassLoader 
 			security.checkPackageAccess(packageName);
 		}
 	}
+}
+
+private long getFieldOffset(String fieldName) {
+	try {
+		Field field = Class.class.getDeclaredField(fieldName);
+		return getUnsafe().objectFieldOffset(field);
+	} catch (NoSuchFieldException e) {
+		throw newInternalError(e);
+	}
+}
+
+/**
+ * This helper method atomically writes the given {@code fieldValue} to the
+ * field specified by the {@code fieldOffset}
+ */
+private void writeFieldValue(long fieldOffset, Object fieldValue) {
+	/*[IF Sidecar19-SE]*/
+	getUnsafe().putObjectRelease(this, fieldOffset, fieldValue);
+	/*[ELSE]*/
+	getUnsafe().putOrderedObject(this, fieldOffset, fieldValue);
+	/*[ENDIF]*/
 }
 
 private static void forNameAccessCheck(final SecurityManager sm, final Class<?> callerClass, final Class<?> foundClass) {
@@ -1158,13 +1197,28 @@ private native Method[] getDeclaredMethodsImpl();
 /**
  * Answers the class which declared the class represented
  * by the receiver. This will return null if the receiver
- * is a member of another class.
+ * is not a member of another class.
  *
  * @return		the declaring class of the receiver.
  */
 @CallerSensitive
 public Class<?> getDeclaringClass() {
-	Class<?> declaringClass = getDeclaringClassImpl();
+	if (cachedDeclaringClassOffset == -1) {
+		cachedDeclaringClassOffset = getFieldOffset("cachedDeclaringClass"); //$NON-NLS-1$
+	}
+	if (cachedDeclaringClass == null) {
+		Class<?> localDeclaringClass = getDeclaringClassImpl();
+		if (localDeclaringClass == null) {
+			localDeclaringClass = ClassReflectNullPlaceHolder.class;
+		}
+		writeFieldValue(cachedDeclaringClassOffset, localDeclaringClass);
+	}
+
+	/**
+	 * ClassReflectNullPlaceHolder.class means the value of cachedDeclaringClass is null
+	 * @see ClassReflectNullPlaceHolder.class
+	 */
+	Class<?> declaringClass = cachedDeclaringClass == ClassReflectNullPlaceHolder.class ? null : cachedDeclaringClass;
 	if (declaringClass == null) {
 		return declaringClass;
 	}
@@ -1303,9 +1357,15 @@ private native Field[] getFieldsImpl();
  * @return		{@code Class<?>[]}
  *					the interfaces the receiver claims to implement.
  */
-public Class<?>[] getInterfaces()
-{
-	return J9VMInternals.getInterfaces(this);
+public Class<?>[] getInterfaces() {
+	if (cachedInterfacesOffset == -1) {
+		cachedInterfacesOffset = getFieldOffset("cachedInterfaces"); //$NON-NLS-1$
+	}
+	if (cachedInterfaces == null) {
+		writeFieldValue(cachedInterfacesOffset, J9VMInternals.getInterfaces(this));
+	}
+	Class<?>[] newInterfaces = cachedInterfaces.length == 0 ? cachedInterfaces: cachedInterfaces.clone();
+	return newInterfaces;
 }
 
 /**
@@ -3185,7 +3245,21 @@ private native Class<?> getEnclosingObjectClass();
 public Class<?> getEnclosingClass() throws SecurityException {
 	Class<?> enclosingClass = getDeclaringClass();
 	if (enclosingClass == null) {
-		enclosingClass = getEnclosingObjectClass();
+		if (cachedEnclosingClassOffset == -1) {
+			cachedEnclosingClassOffset = getFieldOffset("cachedEnclosingClass"); //$NON-NLS-1$
+		}
+		if (cachedEnclosingClass == null) {
+			Class<?> localEnclosingClass = getEnclosingObjectClass();
+			if (localEnclosingClass == null){
+				localEnclosingClass = ClassReflectNullPlaceHolder.class;
+			}
+			writeFieldValue(cachedEnclosingClassOffset, localEnclosingClass);
+		}
+		/**
+		 * ClassReflectNullPlaceHolder.class means the value of cachedEnclosingClass is null
+		 * @see ClassReflectNullPlaceHolder.class
+		 */
+		enclosingClass = cachedEnclosingClass == ClassReflectNullPlaceHolder.class ? null: cachedEnclosingClass;
 	}
 	if (enclosingClass != null) {
 		SecurityManager security = System.getSecurityManager();


jcl/src/java.base/share/classes/java/lang/Class.java
```diff
@@ -1303,11 +1315,24 @@ public Field getField(String name) throws NoSuchFieldException, SecurityExceptio
...
+		if (cachedInterfacesOffset == -1) {
+			Field cachedInterfacesField;
+			try {
+				cachedInterfacesField = Class.class.getDeclaredField("cachedInterfaces"); //$NON-NLS-1$
+				cachedInterfacesOffset = getUnsafe().objectFieldOffset(cachedInterfacesField);
+			} catch (NoSuchFieldException e) {
+				throw newInternalError(e);
+			}
+		}
+		getUnsafe().putOrderedObject(this, cachedInterfacesOffset, J9VMInternals.getInterfaces(this));
```
pshipton: Note the unsafe API is renamed to putObjectRelease() in Java 9 and you need to use the Java preprocessor to generate the correct code depending on the Java version. See the other examples of calling putOrderedObject() in this file.

jcl/src/java.base/share/classes/java/lang/Class.java
```diff
@@ -1162,12 +1162,22 @@ public Method getDeclaredMethod(String name, Class<?>... parameterTypes) throws
...
+
 @CallerSensitive
 public Class<?> getDeclaringClass() {
-	Class<?> declaringClass = getDeclaringClassImpl();
+	Class<?> declaringClass = null;
+	if (cachedDeclaringClass == null){
+		declaringClass = getDeclaringClassImpl();
+		cachedDeclaringClass = declaringClass;
+	}
+	else declaringClass = cachedDeclaringClass;
```
pshipton: Please use `{ }` although its only a single statement.

jcl/src/java.base/share/classes/java/lang/Class.java
```diff
@@ -3177,12 +3202,17 @@ public Method getEnclosingMethod() throws SecurityException {
...
+private Class<?> cachedEnclosingClass;
+
 @CallerSensitive
 public Class<?> getEnclosingClass() throws SecurityException {
 	Class<?> enclosingClass = getDeclaringClass();
 	if (enclosingClass == null) {
-		enclosingClass = getEnclosingObjectClass();
+		if (cachedEnclosingClass == null){
+			enclosingClass = getEnclosingObjectClass();
+		} else enclosingClass = cachedEnclosingClass;
```
JasonFengJ9: It seems `cachedEnclosingClass` is not set a value and will stay `null`.

jcl/src/java.base/share/classes/java/lang/Class.java
JasonFengJ9: Can this line be moved within the block `if (cachedInterfacesOffset == -1) {` to avoid setting `null` to `cachedInterfaces` for each call to `getInterfaces()` in case there is no interfaces specified in the receiver class?

pshipton: no interfaces returns an empty array and not null.

jcl/src/java.base/share/classes/java/lang/Class.java
```diff
@@ -1303,11 +1315,24 @@ public Field getField(String name) throws NoSuchFieldException, SecurityExceptio
...
+			try {
+				cachedInterfacesField = Class.class.getDeclaredField("cachedInterfaces"); //$NON-NLS-1$
+				cachedInterfacesOffset = getUnsafe().objectFieldOffset(cachedInterfacesField);
+			} catch (NoSuchFieldException e) {
+				throw newInternalError(e);
+			}
+		}
+		getUnsafe().putOrderedObject(this, cachedInterfacesOffset, J9VMInternals.getInterfaces(this));
+	}
+	Class<?>[] newInterfaces = cachedInterfaces ? cachedInterfaces.clone(): null;
```
JasonFengJ9: If `cachedInterfaces` can't be `null`, this check is redundant.

jcl/src/java.base/share/classes/java/lang/Class.java
```diff
@@ -1162,12 +1162,22 @@ public Method getDeclaredMethod(String name, Class<?>... parameterTypes) throws
  *
  * @return		the declaring class of the receiver.
  */
+private transient Class<?> cachedDeclaringClass;
```
DanHeidinga: Declaring the fields here disconnects the javadoc from the method it describes.  The fields should likely be declared at the top of the file with other variables.

jcl/src/java.base/share/classes/java/lang/Class.java
```diff
@@ -3177,12 +3202,17 @@ public Method getEnclosingMethod() throws SecurityException {
  * @see #isLocalClass()
  * @see #isMemberClass()
  */
+private Class<?> cachedEnclosingClass;
```
DanHeidinga: This again splits the javadoc from the described method.

DanHeidinga: Also, shouldn't this be `transient` as well?

jcl/src/java.base/share/classes/java/lang/Class.java
```diff
@@ -3177,12 +3202,17 @@ public Method getEnclosingMethod() throws SecurityException {
...
+
 @CallerSensitive
 public Class<?> getEnclosingClass() throws SecurityException {
 	Class<?> enclosingClass = getDeclaringClass();
 	if (enclosingClass == null) {
-		enclosingClass = getEnclosingObjectClass();
+		if (cachedEnclosingClass == null){
```
DanHeidinga: nitpick: `){` -> `) {`

jcl/src/java.base/share/classes/java/lang/Class.java
DanHeidinga: A possible optimization - if `cachedInterfaces.length == 0`, we can skip the clone and return the original as it can't be modified.

jcl/src/java.base/share/classes/java/lang/Class.java
```diff
@@ -1303,11 +1315,24 @@ public Field getField(String name) throws NoSuchFieldException, SecurityExceptio
  * @return		Class<?>[]
  *					the interfaces the receiver claims to implement.
  */
-public Class<?>[] getInterfaces()
-{
-	return J9VMInternals.getInterfaces(this);
+public Class<?>[] getInterfaces(){
+	if (cachedInterfaces == null){
```
DanHeidinga: formatting nitpick: `){`  -> `) {`

jcl/src/java.base/share/classes/java/lang/Class.java
```diff
@@ -1303,11 +1315,24 @@ public Field getField(String name) throws NoSuchFieldException, SecurityExceptio
...
+		if (cachedInterfacesOffset == -1) {
+			Field cachedInterfacesField;
```
DanHeidinga: The field is only used inside the `try` block.  It can be declared inside.

cathyzhyi: Thanks for the comments! Modified accordingly.

jcl/src/java.base/share/classes/java/lang/Class.java
```diff
@@ -1164,10 +1168,29 @@ public Method getDeclaredMethod(String name, Class<?>... parameterTypes) throws
...
+				cachedDeclaringClassOffset = getUnsafe().objectFieldOffset(cachedDeclaringClassField);
+			} catch (NoSuchFieldException e) {
+				throw newInternalError(e);
+			}
+		}
+		/*[IF Sidecar19-SE]
+		getUnsafe().putObjectRelease(this, cachedDeclaringClassOffset, getDeclaringClassImpl());
+		/*[ELSE]*/
+		getUnsafe().putOrderedObject(this, cachedDeclaringClassOffset, getDeclaringClassImpl());
+		/*[ENDIF]*/
```
JasonFengJ9: This `getUnsafe().put...` code can be moved within `if (cachedDeclaringClassOffset == -1) {` to avoid being invoked repeatedly for those classes with `null declaringClass`.

jcl/src/java.base/share/classes/java/lang/Class.java
```diff
@@ -1164,10 +1168,29 @@ public Method getDeclaredMethod(String name, Class<?>... parameterTypes) throws
  */
 @CallerSensitive
 public Class<?> getDeclaringClass() {
-	Class<?> declaringClass = getDeclaringClassImpl();
+	Class<?> declaringClass = null;
```
JasonFengJ9: Initialization to `null` is not necessary.

jcl/src/java.base/share/classes/java/lang/Class.java
```diff
@@ -1164,10 +1168,29 @@ public Method getDeclaredMethod(String name, Class<?>... parameterTypes) throws
...
+	if (cachedDeclaringClass == null) {
+		if (cachedDeclaringClassOffset == -1) {
+			try {
+				Field cachedDeclaringClassField;
+				cachedDeclaringClassField = Class.class.getDeclaredField("cachedDeclaringClass"); //$NON-NLS-1$
```
JasonFengJ9: The variable definition and assignment can be in one line.

jcl/src/java.base/share/classes/java/lang/Class.java
```diff
@@ -1303,11 +1328,28 @@ public Field getField(String name) throws NoSuchFieldException, SecurityExceptio
...
  */
-public Class<?>[] getInterfaces()
-{
-	return J9VMInternals.getInterfaces(this);
+public Class<?>[] getInterfaces(){
+	if (cachedInterfaces == null) {
+		if (cachedInterfacesOffset == -1) {
+			try {
+				Field cachedInterfacesField;
+				cachedInterfacesField = Class.class.getDeclaredField("cachedInterfaces"); //$NON-NLS-1$
```
JasonFengJ9: Similarly these two lines can be combined to one.

jcl/src/java.base/share/classes/java/lang/Class.java
```diff
@@ -1303,11 +1328,28 @@ public Field getField(String name) throws NoSuchFieldException, SecurityExceptio
...
+				cachedInterfacesOffset = getUnsafe().objectFieldOffset(cachedInterfacesField);
+			} catch (NoSuchFieldException e) {
+				throw newInternalError(e);
+			}
+		}
+	/*[IF Sidecar19-SE]
+	getUnsafe().putObjectRelease(this, cachedInterfacesOffset, J9VMInternals.getInterfaces(this));
+	/*[ELSE]*/
+	getUnsafe().putOrderedObject(this, cachedInterfacesOffset, J9VMInternals.getInterfaces(this));
+	/*[ENDIF]*/
```
JasonFengJ9: Please add an indent for `getUnsafe().put...` above.

jcl/src/java.base/share/classes/java/lang/Class.java
cathyzhyi: could synchronization problem happens here that another thread initializes cachedDeclaringClassOffset but not setting cachedDeclaringClass yet and the current thread end up getting a stale value cachedDeclaringClass?

JasonFengJ9: The declaring class offset value can be saved to a local variable after `getUnsafe().objectFieldOffset`, and is only stored into `cachedDeclaringClassOffset` after `cachedDeclaringClass` is set.

pshipton: Multi-threaded, you can't rely on the order which variables are set unless barriers are used.

JasonFengJ9: I think here `getUnsafe().put...` ensures `cachedDeclaringClass` set atomically. Regardless `cachedDeclaringClassOffset` is set or not, `cachedDeclaringClass` value won't be stale. In worst scenarios, `cachedDeclaringClass` will be set more than once which is acceptable.
Any thing I missed?

pshipton: No, looking again the suggestion is good since the unsafe call provides the required barrier.

jcl/src/java.base/share/classes/java/lang/Class.java
```diff
@@ -1164,10 +1168,27 @@ public Method getDeclaredMethod(String name, Class<?>... parameterTypes) throws
...
+			}
+		}
+		/*[IF Sidecar19-SE]
+		getUnsafe().putObjectRelease(this, cachedDeclaringClassOffset, getDeclaringClassImpl());
+		/*[ELSE]*/
+		getUnsafe().putOrderedObject(this, cachedDeclaringClassOffset, getDeclaringClassImpl());
+		/*[ENDIF]*/
+	}
+
+	Class<?> declaringClass = cachedDeclaringClass;
```
pshipton: Please follow the other existing examples in this class, and declare and assign the local before checking for null. Then assign the local variable to the value inside the if null block. The unsafe call ensures writes won't be reordered but doesn't guarantee reads. This applied to all three sets of changes.

```
Object localVar = cachedValue;
if (localVar == null) {
   localVar = <computed value>;
   unsafe.putObjectRelease(this, offset, localVar);
}
<use localVar>
```

jcl/src/java.base/share/classes/java/lang/Class.java
```diff
@@ -1155,6 +1155,10 @@ public Method getDeclaredMethod(String name, Class<?>... parameterTypes) throws
  */
 private native Method[] getDeclaredMethodsImpl();
 
+
+private transient Class<?> cachedDeclaringClass;
```
pshipton: Please move all the field declarations to the top of the file with the other field declarations.

jcl/src/java.base/share/classes/java/lang/Class.java
```diff
@@ -1164,10 +1173,29 @@ public Method getDeclaredMethod(String name, Class<?>... parameterTypes) throws
...
+			try {
+				Field cachedDeclaringClassField = Class.class.getDeclaredField("cachedDeclaringClass"); //$NON-NLS-1$
+				localCachedDeclaringClassOffset = getUnsafe().objectFieldOffset(cachedDeclaringClassField);
+			} catch (NoSuchFieldException e) {
+				throw newInternalError(e);
+			}
+		/*[IF Sidecar19-SE]
+		getUnsafe().putObjectRelease(this, cachedDeclaringClassOffset, getDeclaringClassImpl());
+		/*[ELSE]*/
+		getUnsafe().putOrderedObject(this, cachedDeclaringClassOffset, getDeclaringClassImpl());
```
JasonFengJ9: `cachedDeclaringClassOffset` should be `localCachedDeclaringClassOffset` instead.
Also pls add an indent to align the code in same code block.

jcl/src/java.base/share/classes/java/lang/Class.java
```diff
@@ -1164,10 +1173,29 @@ public Method getDeclaredMethod(String name, Class<?>... parameterTypes) throws
  */
 @CallerSensitive
 public Class<?> getDeclaringClass() {
-	Class<?> declaringClass = getDeclaringClassImpl();
+	if (cachedDeclaringClass == null) {
+		if (cachedDeclaringClassOffset == -1) {
```
JasonFengJ9: These two checks can be combined into one now.

jcl/src/java.base/share/classes/java/lang/Class.java
```diff
@@ -1164,10 +1173,29 @@ public Method getDeclaredMethod(String name, Class<?>... parameterTypes) throws
...
+		/*[ENDIF]*/
+		cachedDeclaringClassOffset = localCachedDeclaringClassOffset;
+		}
+	}
+
+	Class<?> declaringClass = cachedDeclaringClass;
 	if (declaringClass == null) {
 		return declaringClass;
 	}
+
```
JasonFengJ9: Nitpick: unnecessary space change.

jcl/src/java.base/share/classes/java/lang/Class.java
```diff
@@ -1176,7 +1204,7 @@ public Method getDeclaredMethod(String name, Class<?>... parameterTypes) throws
 		}
 		return declaringClass;
 	}
-	
+
```
JasonFengJ9: Same here.

jcl/src/java.base/share/classes/java/lang/Class.java
```diff
@@ -1303,11 +1331,27 @@ public Field getField(String name) throws NoSuchFieldException, SecurityExceptio
...
+				cachedInterfacesOffset = getUnsafe().objectFieldOffset(cachedInterfacesField);
+			} catch (NoSuchFieldException e) {
+				throw newInternalError(e);
+			}
+		}
+      /*[IF Sidecar19-SE]
+        getUnsafe().putObjectRelease(this, cachedInterfacesOffset, J9VMInternals.getInterfaces(this));
+      /*[ELSE]*/
+      getUnsafe().putOrderedObject(this, cachedInterfacesOffset, J9VMInternals.getInterfaces(this));
+      /*[ENDIF]*/
```
JasonFengJ9: Pls use tab instead of space for indents here.

jcl/src/java.base/share/classes/java/lang/Class.java
```diff
@@ -1303,11 +1331,27 @@ public Field getField(String name) throws NoSuchFieldException, SecurityExceptio
...
+   }
+	Class<?>[] newInterfaces = cachedInterfaces.length != 0 ? cachedInterfaces.clone(): cachedInterfaces;
+	return newInterfaces; 
 }
 
+
```
JasonFengJ9: Nitpick: unnecessary space change.

jcl/src/java.base/share/classes/java/lang/Class.java
```diff
@@ -3177,12 +3221,31 @@ public Method getEnclosingMethod() throws SecurityException {
  * @see #isLocalClass()
  * @see #isMemberClass()
  */
+
```
JasonFengJ9: Same here.

jcl/src/java.base/share/classes/java/lang/Class.java
```diff
@@ -3177,12 +3221,31 @@ public Method getEnclosingMethod() throws SecurityException {
...
+				try {
+					Field cachedEnclosingClassField = Class.class.getDeclaredField("cachedEnclosingClass"); //$NON-NLS-1$
+					localCachedEnclosingClassOffset = getUnsafe().objectFieldOffset(cachedEnclosingClassField);
+				} catch (NoSuchFieldException e) {
+					throw newInternalError(e);
+				}
+			/*[IF Sidecar19-SE]
+			getUnsafe().putObjectRelease(this, cachedEnclosingClassOffset, getEnclosingObjectClass());
+			/*[ELSE]*/
+			getUnsafe().putOrderedObject(this, cachedEnclosingClassOffset, getEnclosingObjectClass());
```
JasonFengJ9: Similarly it should be `localCachedEnclosingClassOffset` instead of `cachedEnclosingClassOffset`.
Also, pls add indents to align the code in same block.

jcl/src/java.base/share/classes/java/lang/Class.java
```diff
@@ -3177,12 +3221,31 @@ public Method getEnclosingMethod() throws SecurityException {
...
  * @see #isMemberClass()
  */
+
 @CallerSensitive
 public Class<?> getEnclosingClass() throws SecurityException {
 	Class<?> enclosingClass = getDeclaringClass();
 	if (enclosingClass == null) {
-		enclosingClass = getEnclosingObjectClass();
+		if (cachedEnclosingClass == null) {
+			if (cachedEnclosingClassOffset == -1) {
```
JasonFengJ9: These two checks can be combined into one line.

jcl/src/java.base/share/classes/java/lang/Class.java
```diff
@@ -3177,12 +3221,31 @@ public Method getEnclosingMethod() throws SecurityException {
...
+			getUnsafe().putObjectRelease(this, cachedEnclosingClassOffset, getEnclosingObjectClass());
+			/*[ELSE]*/
+			getUnsafe().putOrderedObject(this, cachedEnclosingClassOffset, getEnclosingObjectClass());
+			/*[ENDIF]*/
+			cachedEnclosingClassOffset = localCachedEnclosingClassOffset;
+			}
+		} 
+		enclosingClass = cachedEnclosingClass;
 	}
+
```
JasonFengJ9: Nitpick: unnecessary space change.

jcl/src/java.base/share/classes/java/lang/Class.java
```diff
@@ -214,6 +214,15 @@
 	private static boolean reflectCacheDebug;
 	private static boolean reflectCacheAppOnly = true;
 
+	private transient Class<?>[] cachedInterfaces;
```
JasonFengJ9: Nitpick: there is an extra space at end of this variable declaration.

jcl/src/java.base/share/classes/java/lang/Class.java
```diff
@@ -1164,7 +1173,23 @@ public Method getDeclaredMethod(String name, Class<?>... parameterTypes) throws
...
-	Class<?> declaringClass = getDeclaringClassImpl();
+	if (cachedDeclaringClassOffset == -1) {
+		long localCachedDeclaringClassOffset;
+		try {
+			Field cachedDeclaringClassField = Class.class.getDeclaredField("cachedDeclaringClass"); //$NON-NLS-1$
+			localCachedDeclaringClassOffset = getUnsafe().objectFieldOffset(cachedDeclaringClassField);
+		} catch (NoSuchFieldException e) {
+			throw newInternalError(e);
+		}
+		/*[IF Sidecar19-SE]
```
JasonFengJ9: Pls add `*/` at end of this JPP decoration (and following similar cases) such that following `Java 9` implementation can be checked by editor compiler.

jcl/src/java.base/share/classes/java/lang/Class.java
```diff
@@ -1164,7 +1173,23 @@ public Method getDeclaredMethod(String name, Class<?>... parameterTypes) throws
  */
 @CallerSensitive
 public Class<?> getDeclaringClass() {
```
JasonFengJ9: There is an existing JavaDoc error for this API.
"This will return null if the receiver is **NOT** a member of another class."
The NOT was missing.
Pls fix it.

jcl/src/java.base/share/classes/java/lang/Class.java
```diff
@@ -284,6 +293,26 @@ private void checkNonSunProxyMemberAccess(SecurityManager security, ClassLoader
 	}
 }
 
+/*
+ * This helper method atomically writes the given \p fieldvalue to the
```
JasonFengJ9: How about `<code></code>`?

jcl/src/java.base/share/classes/java/lang/Class.java
```diff
@@ -284,6 +293,26 @@ private void checkNonSunProxyMemberAccess(SecurityManager security, ClassLoader
 	}
 }
 
+/*
+ * This helper method atomically writes the given \p fieldvalue to the 
+ * field specified by the \p fieldname
+ */
+private long getFieldOffsetAndCacheFieldValue(String fieldname, Object fieldvalue){
```
JasonFengJ9: Please capitalize the first letter of each internal word so the parameters look like `fieldName` & `fieldValue`.

jcl/src/java.base/share/classes/java/lang/Class.java
```diff
@@ -284,6 +293,26 @@ private void checkNonSunProxyMemberAccess(SecurityManager security, ClassLoader
...
+	long fieldOffset;
+	try {
+		Field field = Class.class.getDeclaredField(fieldName); //$NON-NLS-1$
```
JasonFengJ9: `//$NON-NLS-1$` is not needed here. This field doesn't match the incoming parameter, pls verify the code generated by JPP still compiles.

jcl/src/java.base/share/classes/java/lang/Class.java
```diff
@@ -284,6 +293,26 @@ private void checkNonSunProxyMemberAccess(SecurityManager security, ClassLoader
 	}
 }
 
+/**
+ * This helper method atomically writes the given <code>fieldValue</code>to the 
+ * field specified by the <code>fieldName</code>
+ */
+private long getFieldOffsetAndCachefieldValue(String fieldName, Object fieldValue){
```
JasonFengJ9: Just noticed that this should be `getFieldOffsetAndCacheFieldValue`, please ensure the generated `pConfig` code still compiles.

jcl/src/java.base/share/classes/java/lang/Class.java
```diff
@@ -1158,13 +1187,17 @@ public Method getDeclaredMethod(String name, Class<?>... parameterTypes) throws
...
- * is a member of another class.
+ * is not a member of another class.
  *
  * @return		the declaring class of the receiver.
  */
 @CallerSensitive
 public Class<?> getDeclaringClass() {
-	Class<?> declaringClass = getDeclaringClassImpl();
+	if (cachedDeclaringClassOffset == -1) {
+		cachedDeclaringClassOffset = getFieldOffsetAndCacheFieldValue("cachedDeclaringClass", getDeclaringClassImpl());
```
pshipton: As I mentioned before, the fieldValue should be assigned to the local variable when its created. putObjectRelease() ensures the order of stores, but doesn't say anything about the order of reads.

JasonFengJ9: The `fieldOffset` is a local variable within the helper method `getFieldOffsetAndCacheFieldValue`, which is assigned to the static variable `cachedClassOffset` after `putObjectRelease` calls.

pshipton: I was referring to the fieldValue, not the fieldOffset.

JasonFengJ9: The earlier discussion was about `fieldOffset` so I thought you meant it in the comment above as well. 
Regarding `fieldValue`, it is a local (stack allocated) parameter variable which is assigned to the target field value before it is passed to `getUnsafe().putObjectRelease()`. The actual field value variable `cachedClass` is not used during this setting process.

pshipton: > which is assigned to the target field value before it is passed to getUnsafe().putObjectRelease()

I don't see this in the code.

cathyzhyi: > the fieldValue should be assigned to the local variable when its created

Is passing fieldValue as an argument to the `getFieldOffsetAndCacheFieldValue` equivalent to assigning it to a local variable?

pshipton: Its not enough to assign it to some local variable or parameter in a helper, the local variable actually has to be used in the implementations. declaringClass, newInterfaces  (or introduce another local) and enclosingClass  should all be assigned the computed value when the static variable is null.
i.e. something like the following
```
Class<?> declaringClass = cachedDeclaringClass;
if (cachedDeclaringClassOffset == -1) {
		declaringClass = getDeclaringClassImpl();
		cachedDeclaringClassOffset = getFieldOffsetAndCacheFieldValue("cachedDeclaringClass", declaringClass);//$NON-NLS-1$
	}

	. . .
```

JasonFengJ9: Pete concerned that JIT might depend on such local variables to ensure the program order honoured for `cachedClassValues`, i.e., in following PR code snippet:
```
	if (cachedDeclaringClassOffset == -1) {
		cachedDeclaringClassOffset = getFieldOffsetAndCacheFieldValue("cachedDeclaringClass", getDeclaringClassImpl());//$NON-NLS-1$
	}
	Class<?> declaringClass = cachedDeclaringClass;
```
Can JIT ensure `Class<?> declaringClass = cachedDeclaringClass;` occurs after `getFieldOffsetAndCacheFieldValue("cachedDeclaringClass", getDeclaringClassImpl());` since both code affect common memory, i.e., variable `cachedDeclaringClass`? Usually this is not a problem however `getFieldOffsetAndCacheFieldValue()` modify the variable via `unsafe` hence the question is "Does JIT recognize this memory `cachedDeclaringClass` modified by `unsafe`?".

@cathyzhyi , can you check this from JIT side? At mean time, please run the benchmark program with this PR for sanity check.

andrewcraik: @JasonFengJ9 the JIT will preserve semantic ordering even with unsafe - it would be incorrect not. That is if the local depends on the value written by the unsafe store being read the JIT will preserve that ordering. If we didn't do that a lot of things would break.

pshipton: ok thanks, then I retract my concern about using a local for the fieldValue

cathyzhyi:
@pshipton  @JasonFengJ9 I got some failures on my personal builds and seems it's because of the ``cachedInterfaces`` field is changed to NULL after being cached already. Do you guys know any reason that could happen from VM side?

JasonFengJ9:
There was an overlook about the initialization of `fieldOffset` and `cachedClass`, another change is in progress.

cathyzhyi:
Changed the code according to discussion with Jason but there is still the issue when the declaringClass or enclosingClass is null, the ``writeFieldValue`` would be called repeatedly, which is similar to the original code. We need to compare the cached field value to another unique value. @JasonFengJ9 @pshipton  Any suggestions about that?

pshipton:
What was wrong with the previous version of the code?

JasonFengJ9:
@cathyzhyi the `cachedClass` can be initialized to a dummy class such as 
```
private static final class Dummy { Dummy() {} }
cachedClass = new Dummy();
```

JasonFengJ9:
Note: `getDeclaredClasses` returns an array of length 0 if the class declares no classes or interfaces as members, or if this Class object represents a primitive type, an array class, or void.

pshipton:
> static fieldOffset can't be used to determine if the instance cacheClass has been set or not

why?

pshipton:
> cachedClass = new Dummy();

Could just use
```
static final Object dummy = new Object();
...
cachedClass = dummy;

**Comment by JasonFengJ9** [comment]
static `fieldOffset` will be set by first attempt retrieving of `cachedClass` instance, afterwards, another `cachedClass` instance is still `null`, and was thought already initialized since static `fieldOffset` is shared by all class instances though this is first time to get this instance of `cachedClass`.

**Comment by cathyzhyi** [comment]
> static final Object dummy = new Object();
...
cachedClass = dummy;

The cachedClass needs to be Class type though. I changed it to use 
```
private static final class Dummy { }
 ... 
cachedClass = Dummy.class
```

File: jcl/src/java.base/share/classes/java/lang/Class.java [hunk 35]
```diff
@@ -214,6 +214,22 @@
 	private static boolean reflectCacheDebug;
 	private static boolean reflectCacheAppOnly = true;
 
+	/*
+	 * This <code>Dummy</code> class is created to tell whether the cached
+	 * class value is initialized or not; e.g. <code>cachedDeclaringClass</code> and <code>cachedEnclosingClass</code> 
+	 *
+	 * */
+	private static final class Dummy {}
```
JasonFengJ9: Technically it is possible to do
```
private static final Object dummy = new Object();
Class<?> clz = (Class<?>)dummy;
```
later do a comparison against object `dummy`.
I am fine with current `Dummy` class approach unless Pete says otherwise.

jcl/src/java.base/share/classes/java/lang/Class.java
```diff
@@ -1158,13 +1197,20 @@ public Method getDeclaredMethod(String name, Class<?>... parameterTypes) throws
...
  * @return		the declaring class of the receiver.
  */
 @CallerSensitive
 public Class<?> getDeclaringClass() {
-	Class<?> declaringClass = getDeclaringClassImpl();
+	if (cachedDeclaringClassOffset == -1){
+		cachedDeclaringClassOffset = getFieldOffset("cachedDeclaringClass");
+	}
+	if (cachedDeclaringClass == Dummy.class) {
+		writeFieldValue(cachedDeclaringClassOffset, getDeclaringClassImpl());//$NON-NLS-1$
```
JasonFengJ9: `//$NON-NLS-1$` should be with `cachedDeclaringClass`, pls add a space before it.

jcl/src/java.base/share/classes/java/lang/Class.java
```diff
@@ -1303,9 +1349,15 @@ public Field getField(String name) throws NoSuchFieldException, SecurityExceptio
...
  */
-public Class<?>[] getInterfaces()
-{
-	return J9VMInternals.getInterfaces(this);
+public Class<?>[] getInterfaces(){
+	if (cachedInterfacesOffset == -1){
+		cachedInterfacesOffset = getFieldOffset("cachedInterfaces");
+	}
+	if (cachedInterfaces == null) {
+		writeFieldValue(cachedInterfacesOffset, J9VMInternals.getInterfaces(this));//$NON-NLS-1$
```
JasonFengJ9: `//$NON-NLS-1$` should be with `cachedInterfaces `, pls add a space before it.

jcl/src/java.base/share/classes/java/lang/Class.java
```diff
@@ -3181,7 +3233,13 @@ public Method getEnclosingMethod() throws SecurityException {
 public Class<?> getEnclosingClass() throws SecurityException {
 	Class<?> enclosingClass = getDeclaringClass();
 	if (enclosingClass == null) {
-		enclosingClass = getEnclosingObjectClass();
+		if (cachedEnclosingClassOffset == -1){
+			cachedEnclosingClassOffset = getFieldOffset("cachedEnclosingClass");
+		}
+		if (cachedEnclosingClass == Dummy.class) {
+			writeFieldValue(cachedEnclosingClassOffset, getEnclosingObjectClass());//$NON-NLS-1$
```
JasonFengJ9: `//$NON-NLS-1$` should be with `cachedEnclosingClass `, pls add a space before it.

jcl/src/java.base/share/classes/java/lang/Class.java
```diff
@@ -284,6 +300,29 @@ private void checkNonSunProxyMemberAccess(SecurityManager security, ClassLoader
 	}
 }
 
+private long getFieldOffset(String fieldName){
```
JasonFengJ9: Pls add a space between `)` and `{`.

jcl/src/java.base/share/classes/java/lang/Class.java
```diff
@@ -284,6 +300,29 @@ private void checkNonSunProxyMemberAccess(SecurityManager security, ClassLoader
...
+		throw newInternalError(e);
+	}
+	return fieldOffset;
+}
+
+/**
+ * This helper method atomically writes the given <code>fieldValue</code>to the
+ * field specified by the <code>fieldOffset</code>
+ */
+private void writeFieldValue(long fieldOffset, Object fieldValue){
```
JasonFengJ9: Pls add a space between `)` and `{`.

jcl/src/java.base/share/classes/java/lang/Class.java
```diff
@@ -214,6 +214,22 @@
 	private static boolean reflectCacheDebug;
 	private static boolean reflectCacheAppOnly = true;
 
+	/*
+	 * This <code>Dummy</code> class is created to tell whether the cached
+	 * class value is initialized or not; e.g. <code>cachedDeclaringClass</code> and <code>cachedEnclosingClass</code> 
+	 *
+	 * */
```
JasonFengJ9: Pls change this to `*/`.

jcl/src/java.base/share/classes/java/lang/Class.java
```diff
@@ -1158,13 +1197,20 @@ public Method getDeclaredMethod(String name, Class<?>... parameterTypes) throws
...
  * by the receiver. This will return null if the receiver
- * is a member of another class.
+ * is not a member of another class.
  *
  * @return		the declaring class of the receiver.
  */
 @CallerSensitive
 public Class<?> getDeclaringClass() {
-	Class<?> declaringClass = getDeclaringClassImpl();
+	if (cachedDeclaringClassOffset == -1){
```
JasonFengJ9: Pls add a space between `)` and `{`.

jcl/src/java.base/share/classes/java/lang/Class.java
```diff
@@ -1303,9 +1349,15 @@ public Field getField(String name) throws NoSuchFieldException, SecurityExceptio
  * @return		Class<?>[]
  *					the interfaces the receiver claims to implement.
  */
-public Class<?>[] getInterfaces()
-{
-	return J9VMInternals.getInterfaces(this);
+public Class<?>[] getInterfaces(){
+	if (cachedInterfacesOffset == -1){
```
JasonFengJ9: Pls add a space between `)` and `{`.

jcl/src/java.base/share/classes/java/lang/Class.java
```diff
@@ -3181,7 +3233,13 @@ public Method getEnclosingMethod() throws SecurityException {
 public Class<?> getEnclosingClass() throws SecurityException {
 	Class<?> enclosingClass = getDeclaringClass();
 	if (enclosingClass == null) {
-		enclosingClass = getEnclosingObjectClass();
+		if (cachedEnclosingClassOffset == -1){
```
JasonFengJ9: Pls add a space between `)` and `{`.

JasonFengJ9:
@cathyzhyi regarding the issue discovered that the `cachedClass` in a deserialized class object is `null`, I would suggest to add a `readObject(ObjectInputStream s)` method to re-construct the default value of `cachedClass`.

cathyzhyi:
Java class objects are initialized by VM and the default cachedClass field value needs special handling in ``createramclass.cpp``. @JasonFengJ9 has kindly agreed to help with the implementation.

JasonFengJ9:
It still can be done in Java (Thanks to Pete's hint).
The default `cacheClass` is `null`, it can be set to `Dummy.class` after first real call to get the actual class.
@cathyzhyi pls give a try.

jcl/src/java.base/share/classes/java/lang/Class.java
```diff
@@ -1158,13 +1196,21 @@ public Method getDeclaredMethod(String name, Class<?>... parameterTypes) throws
...
-	Class<?> declaringClass = getDeclaringClassImpl();
+	if (cachedDeclaringClassOffset == -1) {
+		cachedDeclaringClassOffset = getFieldOffset("cachedDeclaringClass"); //$NON-NLS-1$
+	}
+	if (cachedDeclaringClass == null) {
+		Class<?> localDeclaringClass = getDeclaringClassImpl();
+		writeFieldValue(cachedDeclaringClassOffset, localDeclaringClass != null ? localDeclaringClass: Dummy.class);
+	}
+
+	Class<?> declaringClass = cachedDeclaringClass != Dummy.class ? cachedDeclaringClass : null;
```
JasonFengJ9: A comment is needed to explain why we do this kind of variable initialization. 
In addition, pls re-run benchmark tests against this PR.

jcl/src/java.base/share/classes/java/lang/Class.java
```diff
@@ -214,6 +214,24 @@
 	private static boolean reflectCacheDebug;
 	private static boolean reflectCacheAppOnly = true;
 
+	/*
+	 * This {@code Dummy} class is created to indicate the cached class value is
+	 * initialized to null ;e.g. {@code cachedDeclaringClass} and {@code cachedEnclosingClass}
```
JasonFengJ9: The reason for not able to set `cachedClass` default value to `Dummy.class` is that `j.l.Class` can't be instantiated in `Java`, its instances are created by the VM only, and any instance field with non-null default values have to be set by VM natives.

jcl/src/java.base/share/classes/java/lang/Class.java
```diff
@@ -214,6 +214,24 @@
 	private static boolean reflectCacheDebug;
 	private static boolean reflectCacheAppOnly = true;
 
+	/*
+	 * This {@code Dummy} class is created to indicate the cached class value is
+	 * initialized to null rather than the default value null ;e.g. {@code cachedDeclaringClass}
+	 * and {@code cachedEnclosingClass}. The reason default value has to be null is that
+	 * j.l.Class instances are created by the VM only rather than Java, an
```
pshipton: an should be and

pshipton:
@guangjxu can you please help @cathyzhyi with a footprint analysis to determine if the footprint increase to java.lang.Class will be a problem.

BeverlyXu:
Discussed it with Yi. She will use the PerfNext infra with the help from Mansoor

cathyzhyi:
Tested on major benchmarks and show no obvious footprint increase.

pshipton:
Is this ready to be merged? If so please remove the WIP from the title, and add a comment so I notice.

cathyzhyi:
@pshipton Yes, it's ready to be merged. Thanks!

jcl/src/java.base/share/classes/java/lang/Class.java
```diff
@@ -284,6 +302,29 @@ private void checkNonSunProxyMemberAccess(SecurityManager security, ClassLoader
 	}
 }
 
+private long getFieldOffset(String fieldName) {
+	long fieldOffset;
+	try {
+		Field field = Class.class.getDeclaredField(fieldName);
+		fieldOffset = getUnsafe().objectFieldOffset(field);
```
DanHeidinga: This can be `return getUnsafe().objectFieldOffset(field);`.  The local var isn't required

jcl/src/java.base/share/classes/java/lang/Class.java
```diff
@@ -1158,13 +1199,25 @@ public Method getDeclaredMethod(String name, Class<?>... parameterTypes) throws
...
  */
 @CallerSensitive
 public Class<?> getDeclaringClass() {
-	Class<?> declaringClass = getDeclaringClassImpl();
+	if (cachedDeclaringClassOffset == -1) {
+		cachedDeclaringClassOffset = getFieldOffset("cachedDeclaringClass"); //$NON-NLS-1$
+	}
+	if (cachedDeclaringClass == null) {
+		Class<?> localDeclaringClass = getDeclaringClassImpl();
+		writeFieldValue(cachedDeclaringClassOffset, localDeclaringClass == null ? Dummy.class : localDeclaringClass);
```
DanHeidinga: While concise, this pattern is harder to set a breakpoint on.
Can this be written as:
```
if (localDeclaringClass == null) {
  localDeclaringClass = Dummy.class;
}
writeFieldValue(cachedDeclaringClassOffset, localDeclaringClass);
```
That's much easier to breakpoint on the case where we're writing null

jcl/src/java.base/share/classes/java/lang/Class.java
```diff
@@ -3181,7 +3241,20 @@ public Method getEnclosingMethod() throws SecurityException {
 public Class<?> getEnclosingClass() throws SecurityException {
 	Class<?> enclosingClass = getDeclaringClass();
 	if (enclosingClass == null) {
-		enclosingClass = getEnclosingObjectClass();
+		if (cachedEnclosingClassOffset == -1) {
+			cachedEnclosingClassOffset = getFieldOffset("cachedEnclosingClass"); //$NON-NLS-1$
+		}
+		if (cachedEnclosingClass == null) {
+			Class<?> localEnclosingClass = getEnclosingObjectClass();
+			if (localEnclosingClass == null)
```
DanHeidinga: Can you add the `{ }` to this if statement?

cathyzhyi: added `{ }` as suggested. thanks!

DanHeidinga:

DanHeidinga:
@cathyzhyi I looked at the sanity tests and don't see anything that would currently cover the cache part of this PR.  Could you add some tests (or adapt existing tests) to fetch the values a second time?  For both the normal and `ClassReflectNullPlaceHolder` cases?

Existing tests in https://github.com/eclipse/openj9/blob/f66506e7a528340d176260c3af47c63cf66979ec/test/functional/Java8andUp/src_80/org/openj9/test/java/lang/Test_Class.java may be a good starting point.

I'm OK with this being merged once a separate PR has been opened for the tests

DanHeidinga:
> I'm OK with this being merged once a separate PR has been opened for the tests

@cathyzhyi any update on opening a PR with tests for this change?  I see that test PR being created as a blocker to getting this merged.

cathyzhyi:
@DanHeidinga Hi Daniel, a test PR has been created here. https://github.com/eclipse/openj9/pull/2025

DanHeidinga:

DanHeidinga:
Thanks @cathyzhyi I've launched a new build on will merge when it passes

PR #1549, opened 2018-03-27
Use the right constant pool for primitive/array

liqunl:
A null pointer as the constant pool in the method meta data will cause
a crash in gc during stack walk. The constant pool for a
newInstanceImplThunk will be the constant pool of the class to be
instantiated. If a primitive or an array is asked to be instantiated
through Class.newInstance, an exception is expected instead of a crash.
Since primitive and array classes don't have constant pool, this change
uses the constant pool of java/lang/Class for their newInstanceImplThunk
to fix the crash.

### DIFF
--- a/runtime/compiler/trj9/env/j9method.cpp
+++ b/runtime/compiler/trj9/env/j9method.cpp
@@ -5656,10 +5656,18 @@ TR_ResolvedJ9Method::allocateException(uint32_t numBytes, TR::Compilation *comp)
       {
       TR_ASSERT(_j9classForNewInstance, "Must have the class for the newInstance");
       //J9Class *clazz = (J9Class*) ((intptrj_t)_ramMethod->extra & ~J9_STARTPC_NOT_TRANSLATED);
-      cpool = J9_CP_FROM_CLASS(_j9classForNewInstance);
+
+      // Primitives and arrays don't have constant pool, use the constant pool of java/lang/Class
+      if (TR::Compiler->cls.isPrimitiveClass(comp, (TR_OpaqueClassBlock*)_j9classForNewInstance) ||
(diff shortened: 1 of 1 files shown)


liqunl:
@andrewcraik Waiting for the build to finish, will unWIP it once all jobs finish.

DanHeidinga:
Class.newInstance should throw when a primitive or array is passed as the class.  See the check in 
```
static VMINLINE bool
	classCanBeInstantiated(J9Class *j9clazz)
	{
		J9ROMClass *romClass = j9clazz->romClass;
		return !J9ROMCLASS_IS_ABSTRACT(romClass) && !J9ROMCLASS_IS_INTERFACE(romClass) && !J9ROMCLASS_IS_PRIMITIVE_TYPE(romClass) && !J9ROMCLASS_IS_ARRAY(romClass);
	}
```
which limits out abstract classes, interfaces, primitives and arrays.

DanHeidinga:
Why are we trying to compile a newInstanceThunk for a case that should always throw?

liqunl:
@DanHeidinga Sorry, I'm not familiar with the mechanism of how we decide to request a compilation of newInstanceImplThunk. From the core dump, the thunk for boolean has been compiled. Based on an old internal documentation, it looks like it's possible that the JIT is asked to compile a newInstanceImplThunk for unsuitable classes. In that case, the JIT will compile the bytecode of java/lang/Class.newInstancePrototype, which always throws InstantiationException.

DanHeidinga:
Thanks for clarifying @liqunl.  I'm surprised user code calls this enough for primitives / arrays for it to get compiled.

liqunl:
Personal build has passed.

andrewcraik:
Jenkins test sanity

PR #1551, opened 2018-03-27
Mark security sensitive Class.forName() as privileged

JasonFengJ9:
Marked following method calls as privileged to avoid throwing `java.security.AccessControlException: Access denied
("java.lang.RuntimePermission" "getClassLoader")`:
```
Class.forName(java_logging, "java.util.logging.LogManager")
Class.forName(java_logging, "java.util.logging.Logger")
Class.forName(java_logging, "java.util.logging.Level")
```

Manually verified that this PR fixed following error:
```
Caused by: java.security.AccessControlException: Access denied ("java.lang.RuntimePermission" "getClassLoader")
	at java.security.AccessControlContext.checkPermissionWithCache(java.base@9.0.4-internal/AccessControlContext.java:645)
	at java.security.AccessController.checkPermissionHelper(java.base@9.0.4-internal/AccessController.java:243)
	at java.security.AccessController.checkPermission(java.base@9.0.4-internal/AccessController.java:385)
	at java.lang.SecurityManager.checkPermission(java.base@9.0.4-internal/SecurityManager.java:558)
	at java.lang.Class.forName(java.base@9.0.4-internal/Class.java:427)
	at com.ibm.java.lang.management.internal.LoggingMXBeanImpl.<clinit>(java.management@9.0.4-internal/LoggingMXBeanImpl.java:76)

```

Reviewer @DanHeidinga 

### DIFF
--- a/jcl/src/java.management/share/classes/com/ibm/java/lang/management/internal/LoggingMXBeanImpl.java
+++ b/jcl/src/java.management/share/classes/com/ibm/java/lang/management/internal/LoggingMXBeanImpl.java
@@ -34,6 +34,8 @@ import java.lang.reflect.Layer;
 /*[ENDIF]*/
 import java.lang.reflect.InvocationTargetException;
 import java.lang.reflect.Method;
+import java.security.AccessController;
+import java.security.PrivilegedAction;
 import java.util.ArrayList;
 import java.util.Enumeration;
 import java.util.List;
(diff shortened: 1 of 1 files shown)


jcl/src/java.management/share/classes/com/ibm/java/lang/management/internal/LoggingMXBeanImpl.java
```diff
@@ -83,10 +85,25 @@
 			Module java_logging = ModuleLayer.boot().findModule("java.logging").get(); //$NON-NLS-1$
 /*[ELSE]
 			Module java_logging = Layer.boot().findModule("java.logging").get(); //$NON-NLS-1$
-/*[ENDIF]*/			
-			Class<?> logManagerClass = Class.forName(java_logging, "java.util.logging.LogManager"); //$NON-NLS-1$
-			Class<?> loggerClass = Class.forName(java_logging, "java.util.logging.Logger"); //$NON-NLS-1$
-			Class<?> levelClass = Class.forName(java_logging, "java.util.logging.Level"); //$NON-NLS-1$
+/*[ENDIF]*/
+			Class<?> logManagerClass = AccessController.doPrivileged(new PrivilegedAction<Class<?>>() {
```
DanHeidinga: Can these be written as lambdas?  Or backed by a single class?

pshipton: Please use a helper to avoid creating 3 new classes.

pshipton: I hadn't seen Dan's comment before I added mine.

JasonFengJ9: Updated the PR to invoke `PrivilegedAction` with `lambda` expression. In addition, stored three `Class.forName()` into an array to avoid creating three new classes as per Pete's comment. Please review again.

pshipton:
jenkins test all plinux jdk9

pshipton:
The sanity failure is caused by the commit comment containing the phrase "Exception:", which ended up in an environment variable which is shown in the javacore being checked, and caused a command line tester failure `Failure condition was found: [Output match: Exception:]`. I don't expect this will affect future testing so it can be ignored.

Issue #1524, opened 2018-03-23
jdk_lang Test7157574 segfault


pdbain-ibm:
```
    $JAVA_HOME/jre/bin/../../bin/java \
        -Dtest.class.path.prefix=$JVMTEST/openjdk_regression/work/classes/3/java/lang/invoke/7157574:$TEST_HOME/openjdk_regression/openjdk-jdk/jdk/test/java/lang/invoke/7157574 \
        -Dtest.src=$TEST_HOME/openjdk_regression/openjdk-jdk/jdk/test/java/lang/invoke/7157574 \
        -Dtest.src.path=$TEST_HOME/openjdk_regression/openjdk-jdk/jdk/test/java/lang/invoke/7157574 \
        -Dtest.classes=$JVMTEST/openjdk_regression/work/classes/3/java/lang/invoke/7157574 \
        -Dtest.class.path=$JVMTEST/openjdk_regression/work/classes/3/java/lang/invoke/7157574 \
        -Dtest.vm.opts='-ea -esa -Xmx512m -Xcompressedrefs' \
        -Dtest.tool.vm.opts='-J-ea -J-esa -J-Xmx512m -J-Xcompressedrefs' \
        -Dtest.compiler.opts= \
        -Dtest.java.opts= \
        -Dtest.jdk=$JAVA_HOME/jre/bin/../.. \
        -Dcompile.jdk=$JAVA_HOME/jre/bin/../.. \
        -Dtest.timeout.factor=4.0 \
        -classpath $JVMTEST/openjdk_regression/work/classes/3/java/lang/invoke/7157574:$TEST_HOME/openjdk_regression/openjdk-jdk/jdk/test/java/lang/invoke/7157574:$JAVA_HOME/jre/bin/../../lib/tools.jar:$JVMTEST/openjdk_regression/jtreg/lib/javatest.jar:$JVMTEST/openjdk_regression/jtreg/lib/jtreg.jar \
        Test7157574

MethodHandle(Sub)void
MethodHandle(Sub)void
MethodHandle(Sub)int
MethodHandle(Sub,int)void
MethodHandle(Sub)int
Unhandled exception
Type=Segmentation error vmState=0x00000000
J9Generic_Signal_Number=00000004 Signal_Number=0000000b Error_Value=00000000 Signal_Code=00000001
Handler1=00007F8BA958C110 Handler2=00007F8BA875B7C0 InaccessibleAddress=0000000000000000
[...]

PR #1554, opened 2018-03-27
Set InterfaceHandle correctly for methods inherited from Object

pdbain-ibm:
When calling lookup().getVirtual() on an interface for a method in
java.lang.Object, don't check the itables for the method.  Instead, set the
methodIndex to -1 and return the concrete method.

Fixes https://github.com/eclipse/openj9/issues/1524

### DIFF
--- a/runtime/jcl/common/java_dyn_methodhandle.c
+++ b/runtime/jcl/common/java_dyn_methodhandle.c
@@ -95,9 +95,32 @@ lookupInterfaceMethod(J9VMThread *currentThread, J9Class *lookupClass, J9UTF8 *n
 			vmFuncs->setCurrentExceptionNLS(currentThread, J9VMCONSTANTPOOL_JAVALANGINCOMPATIBLECLASSCHANGEERROR, J9NLS_JCL_PRIVATE_INTERFACE_REQUIRES_INVOKESPECIAL);
 			method = NULL;
 		} else {
-			*methodIndex = getITableIndexForMethod(method, lookupClass);
-			if (*methodIndex == -1) {
-				method = NULL;
+			if (J9_ARE_ANY_BITS_SET(J9_CLASS_FROM_METHOD(method)->romClass->modifiers, J9_JAVA_INTERFACE)) {
+				*methodIndex = getITableIndexForMethod(method, lookupClass);
+				if (-1  == *methodIndex) {
+					PORT_ACCESS_FROM_VMC(currentThread);
+					J9Class *clazz = J9_CLASS_FROM_METHOD(method);
+					J9UTF8 *className = J9ROMCLASS_CLASSNAME(clazz->romClass);
+					J9ROMMethod *romMethod = J9_ROM_METHOD_FROM_RAM_METHOD(method);
+					J9UTF8 *methodName = J9ROMMETHOD_NAME(romMethod);
+					J9UTF8 *sig = J9ROMMETHOD_SIGNATURE(romMethod);
+					size_t nameAndSigLength = J9UTF8_LENGTH(className)
+						+ J9UTF8_LENGTH(methodName) + J9UTF8_LENGTH(sig)
+						+ 4 /* period, parentheses, and terminating null */;
+					char *msg = j9mem_allocate_memory(nameAndSigLength, OMRMEM_CATEGORY_VM);
+					if (NULL != msg) {
+						j9str_printf(PORTLIB, msg, nameAndSigLength, "%.*s.%.*s(%.*s)",
+								J9UTF8_LENGTH(className), J9UTF8_DATA(className),
+								J9UTF8_LENGTH(methodName), J9UTF8_DATA(methodName),
+								J9UTF8_LENGTH(sig), J9UTF8_DATA(sig));
+						vmFuncs->setCurrentExceptionUTF(currentThread,
+								J9VMCONSTANTPOOL_JAVALANGNOSUCHMETHODERROR, msg);
+						j9mem_free_memory(msg);
+					} else {
+						vmFuncs->setNativeOutOfMemoryError(currentThread, 0, 0);
+					}
+					method = NULL;
+				}
 			}
 		}
 	}
--- a/runtime/oti/util_api.h
+++ b/runtime/oti/util_api.h
@@ -1888,6 +1888,8 @@ getITableIndexForMethod(J9Method * method, J9Class *targetInterface);
  * Returns the first ROM method following the argument.
  * If this is called on the last ROM method in a ROM class
  * it will return an undefined value.
+ * The defining class of method must be an interface; do not use this
+ * for methods inherited from java.lang.Object.
  *
  * @param[in] romMethod - the current ROM method
  * @return - the ROM method following the current one


runtime/jcl/common/java_dyn_methodhandle.c
```diff
@@ -95,9 +95,8 @@ lookupInterfaceMethod(J9VMThread *currentThread, J9Class *lookupClass, J9UTF8 *n
 			vmFuncs->setCurrentExceptionNLS(currentThread, J9VMCONSTANTPOOL_JAVALANGINCOMPATIBLECLASSCHANGEERROR, J9NLS_JCL_PRIVATE_INTERFACE_REQUIRES_INVOKESPECIAL);
 			method = NULL;
 		} else {
-			*methodIndex = getITableIndexForMethod(method, lookupClass);
-			if (*methodIndex == -1) {
-				method = NULL;
+			if (J9_ARE_ANY_BITS_SET(J9_CLASS_FROM_METHOD(method)->romClass->modifiers, J9_JAVA_INTERFACE)) {
+				*methodIndex = getITableIndexForMethod(method, lookupClass);
```
DanHeidinga: We should still be setting method to null if getITableIndexForMethod() returned -1.

DanHeidinga: Also, is there a path through this code (or the caller) that allows returning with a null J9Method without having set an exception?

pdbain-ibm: Updated to return null and throw an exception.

lookupImpl() throws an exception when it returns null.
getMethodForSpecialSend() asserts notNull on the result

pdbain-ibm:
@DanHeidinga any other comments?

runtime/jcl/common/java_dyn_methodhandle.c
```diff
@@ -95,9 +95,12 @@ lookupInterfaceMethod(J9VMThread *currentThread, J9Class *lookupClass, J9UTF8 *n
...
+				if (-1  == *methodIndex) {
+					method = NULL;
+					vmFuncs->setCurrentException(currentThread, J9VMCONSTANTPOOL_JAVALANGNOSUCHMETHODERROR, NULL);
```
DanHeidinga: Can you sent the method name & signature as the message for this exception?

runtime/jcl/common/java_dyn_methodhandle.c
```diff
@@ -95,9 +96,22 @@ lookupInterfaceMethod(J9VMThread *currentThread, J9Class *lookupClass, J9UTF8 *n
 			vmFuncs->setCurrentExceptionNLS(currentThread, J9VMCONSTANTPOOL_JAVALANGINCOMPATIBLECLASSCHANGEERROR, J9NLS_JCL_PRIVATE_INTERFACE_REQUIRES_INVOKESPECIAL);
 			method = NULL;
 		} else {
-			*methodIndex = getITableIndexForMethod(method, lookupClass);
-			if (*methodIndex == -1) {
-				method = NULL;
+			if (J9_ARE_ANY_BITS_SET(J9_CLASS_FROM_METHOD(method)->romClass->modifiers, J9_JAVA_INTERFACE)) {
+				*methodIndex = getITableIndexForMethod(method, lookupClass);
+				if (-1  == *methodIndex) {
+					char msg[64]; /* long enough for most name and signature combinations */
```
DanHeidinga: Let's write the code to allocate a buffer if this assumption is invalid.  Not uncommon for ~40+ chars to be consumed by the package name.  We've got examples of that pattern in exceptionsupport.c (I think)

pdbain-ibm:
Dan
No problem. Since this situation is uncommon, I will allocate and free buffer unconditionally.
Testing now...

runtime/jcl/common/java_dyn_methodhandle.c
```diff
@@ -95,9 +96,28 @@ lookupInterfaceMethod(J9VMThread *currentThread, J9Class *lookupClass, J9UTF8 *n
...
+					J9Class *clazz = J9_CLASS_FROM_METHOD(method);
+					J9UTF8 *className = J9ROMCLASS_CLASSNAME(clazz->romClass);
+					J9ROMMethod *romMethod = J9_ROM_METHOD_FROM_RAM_METHOD(method);
+					J9UTF8 *methodName = J9ROMMETHOD_NAME(romMethod);
+					J9UTF8 *sig = J9ROMMETHOD_SIGNATURE(romMethod);
+					size_t nameAndSigLength = J9UTF8_LENGTH(className)
+													+ J9UTF8_LENGTH(methodName) + J9UTF8_LENGTH(sig)
+													+ 4 /* period, parentheses, and terminating null */;
+					char *msg = (char *)j9mem_allocate_memory(nameAndSigLength, OMRMEM_CATEGORY_VM);
+					Assert_JCL_notNull(msg);
```
DanHeidinga: We don't assert on the return of `j9mem_allocate_memory`.  The code can either throw the NSME or it can throw a NativeOOM but crashing the VM isn't the right choice.

runtime/jcl/common/java_dyn_methodhandle.c
```diff
@@ -95,9 +96,28 @@ lookupInterfaceMethod(J9VMThread *currentThread, J9Class *lookupClass, J9UTF8 *n
...
+				if (-1  == *methodIndex) {
+					J9Class *clazz = J9_CLASS_FROM_METHOD(method);
+					J9UTF8 *className = J9ROMCLASS_CLASSNAME(clazz->romClass);
+					J9ROMMethod *romMethod = J9_ROM_METHOD_FROM_RAM_METHOD(method);
+					J9UTF8 *methodName = J9ROMMETHOD_NAME(romMethod);
+					J9UTF8 *sig = J9ROMMETHOD_SIGNATURE(romMethod);
+					size_t nameAndSigLength = J9UTF8_LENGTH(className)
+													+ J9UTF8_LENGTH(methodName) + J9UTF8_LENGTH(sig)
+													+ 4 /* period, parentheses, and terminating null */;
+					char *msg = (char *)j9mem_allocate_memory(nameAndSigLength, OMRMEM_CATEGORY_VM);
```
DanHeidinga: The `(char *)` cast isn't required for the return value of `j9mem_allocate_memory()` in C code.

runtime/jcl/common/java_dyn_methodhandle.c
```diff
@@ -78,6 +78,7 @@ static UDATA
 lookupInterfaceMethod(J9VMThread *currentThread, J9Class *lookupClass, J9UTF8 *name, J9UTF8 *signature, UDATA *methodIndex)
 {
 	J9Method *method;
+	PORT_ACCESS_FROM_VMC(currentThread);
```
DanHeidinga: This can be moved to the smallest enclosing the scope - the one using `j9mem_allocate_memory`

runtime/jcl/common/java_dyn_methodhandle.c
```diff
@@ -95,9 +96,28 @@ lookupInterfaceMethod(J9VMThread *currentThread, J9Class *lookupClass, J9UTF8 *n
...
-				method = NULL;
+			if (J9_ARE_ANY_BITS_SET(J9_CLASS_FROM_METHOD(method)->romClass->modifiers, J9_JAVA_INTERFACE)) {
+				*methodIndex = getITableIndexForMethod(method, lookupClass);
+				if (-1  == *methodIndex) {
+					J9Class *clazz = J9_CLASS_FROM_METHOD(method);
+					J9UTF8 *className = J9ROMCLASS_CLASSNAME(clazz->romClass);
+					J9ROMMethod *romMethod = J9_ROM_METHOD_FROM_RAM_METHOD(method);
+					J9UTF8 *methodName = J9ROMMETHOD_NAME(romMethod);
+					J9UTF8 *sig = J9ROMMETHOD_SIGNATURE(romMethod);
+					size_t nameAndSigLength = J9UTF8_LENGTH(className)
```
DanHeidinga: Can you adjust the formatting on this?  The lines seem excessively tabbed over

Something like this is more readable
```
					size_t nameAndSigLength = J9UTF8_LENGTH(className)
							+ J9UTF8_LENGTH(methodName) + J9UTF8_LENGTH(sig)
							+ 4 /* period, parentheses, and terminating null */;
```

pdbain-ibm:
Thank you. Formatting, declarations, and allocation failure updated accordingly.

runtime/jcl/common/java_dyn_methodhandle.c
```diff
@@ -95,9 +95,32 @@ lookupInterfaceMethod(J9VMThread *currentThread, J9Class *lookupClass, J9UTF8 *n
...
+						+ J9UTF8_LENGTH(methodName) + J9UTF8_LENGTH(sig)
+						+ 4 /* period, parentheses, and terminating null */;
+					char *msg = j9mem_allocate_memory(nameAndSigLength, OMRMEM_CATEGORY_VM);
+					if (NULL != msg) {
+						j9str_printf(PORTLIB, msg, nameAndSigLength, "%s.%s(%s)",
```
DanHeidinga: The format string isn't correct.  It needs to be `%.*s.%.*s(%.*s)` to handle the non-null terminated UTF8s.

pdbain-ibm:
Fixed format string.

DanHeidinga:
Jenkins test sanity plinux

DanHeidinga:
@pdbain-ibm Can you create another PR to add your simple test case from #1524 to the JSR 292 test bucket?

pdbain-ibm:
https://github.com/eclipse/openj9/issues/1593

PR #1558, opened 2018-03-27
Avoid evaluation of VSI instruction memory references

NigelYiboYu:
VSI instructions have 12-bit displacements and a base register. There is
not need for a separate LA/LAY instruction if a memory reference lacks
index register.

If the index register is absent, codegen should avoid evaluating address
nodes, and delay the index register/large displacement handling to the
generateVSIInstruction() API call, which will explicitly call
separateIndexRegister() API for such matter.


NigelYiboYu:
Depends on OMR PR https://github.com/eclipse/omr/pull/2391

NigelYiboYu:
@fjeremic  new PR for the memory reference folding change.
Depends on OMR PR: https://github.com/eclipse/omr/pull/2391

fjeremic:
Jenkins test extended depends eclipse/omr#2391

PR #1559, opened 2018-03-27
WIP: Atomic-free JNI work

gacholio:
- re-add assert when acquiring exclusive
- support retaining VM access during exclusive
- disable OSR safe point if retaining VM access for atomic-free


pshipton:
It looks fine, but I'll run a minimal sanity just in case.

pshipton:
I think technically the PR title shouldn't contain WIP since this means don't merge it yet.

PR #1570, opened 2018-03-28
Fix the java doc contents in openJ9 class files to work with JDK file make/CompileJavaModules.gmk

archanaNogriya:
Fix the java doc contents in openJ9 class files to work with JDK file make/CompileJavaModules.gmk without making any patch changes.

Fixes : https://github.com/eclipse/openj9/issues/1569

Explanation:
Build has been failing with these 20 files.
**Basically these errors are related to 2 issues and the fixes is required to work is detailed here**

- code text in the generated HTML
Text within {@code ...} will be automatically HTML-escaped however <code> is not which is why above most of the files failed during compile.
{@code ...} is more concise: easier to read (and type).

The {@code} is equivalent to <code> except that it uses allows you to use angle brackets < and > instead of HTML-entities &lt; and &gt.
For example, {@code List<String>} is equivalent to  <code>List&lt;String&gt;</code>.

See also the documentation for {@code}.
https://docs.oracle.com/javase/1.5.0/docs/tooldocs/windows/javadoc.html#%7B@code%7D

also

https://blogs.oracle.com/darcy/javadoc-tip:-code-and-literal

- Invalid syntax which has not been supported by HTML5
example : align or summary in is not supported in HTML5, 
another example.
also there were few error such as take not closed or parameter was not given.

These 20 files I have fixed and I have created a pull request.
After these fixes, My personal local build is successfully passed. 

Signed-off-by: Archana Nogriya <archana.nogriya@uk.ibm.com>

### DIFF
--- a/jcl/src/java.base/share/classes/java/lang/Class.java
+++ b/jcl/src/java.base/share/classes/java/lang/Class.java
@@ -1300,7 +1300,7 @@ private native Field[] getFieldsImpl();
  * specified in the receiver classes <code>implements</code>
  * declaration
  *
- * @return		Class<?>[]
+ * @return		{@code Class<?>[]}
  *					the interfaces the receiver claims to implement.
  */
 public Class<?>[] getInterfaces()
@@ -1314,7 +1314,7 @@ public Class<?>[] getInterfaces()
  *
  * @param		name String
  *					the name of the method
- * @param		parameterTypes Class<?>[]
+ * @param		parameterTypes {@code Class<?>[]}
  *					the types of the arguments.
  * @return		Method
  *					the method described by the arguments.
@@ -2820,8 +2820,7 @@ public boolean isAnnotationPresent(Class<? extends Annotation> annotation) {
 
 /**
  * Cast this Class to a subclass of the specified Class. 
- * 
- * @param cls the Class to cast to
+ * @param <U> cls the Class to cast to
  * @return this Class, cast to a subclass of the specified Class
  * 
  * @throws ClassCastException if this Class is not the same or a subclass
--- a/jcl/src/java.base/share/classes/java/lang/String.java
+++ b/jcl/src/java.base/share/classes/java/lang/String.java
@@ -138,7 +138,7 @@ public final class String implements Serializable, Comparable<String>, CharSeque
 		 *			  an Object to compare
 		 * @param o2
 		 *			  an Object to compare
-		 * @return an int < 0 if object1 is less than object2, 0 if they are equal, and > 0 if object1 is greater
+		 * @return an {@code int < 0} if object1 is less than object2, 0 if they are equal, and {@code > 0} if object1 is greater
 		 *
 		 * @exception ClassCastException
 		 *					when objects are not the correct type
@@ -505,7 +505,7 @@ public final class String implements Serializable, Comparable<String>, CharSeque
 	 *			  the number of bytes to convert
 	 *
 	 * @throws IndexOutOfBoundsException
-	 *				when <code>length < 0, start < 0</code> or <code>start + length > data.length</code>
+	 *				when {@code length < 0, start < 0} or {@code start + length > data.length}
 	 * @throws NullPointerException
 	 *				when data is null
 	 *
@@ -585,7 +585,7 @@ public final class String implements Serializable, Comparable<String>, CharSeque
 	 *			  the number of bytes to convert
 	 *
 	 * @throws IndexOutOfBoundsException
-	 *				when <code>length < 0, start < 0</code> or <code>start + length > data.length</code>
+	 *				when {@code length < 0, start < 0} or {@code start + length > data.length}
 	 * @throws NullPointerException
 	 *				when data is null
 	 *
@@ -670,7 +670,7 @@ public final class String implements Serializable, Comparable<String>, CharSeque
 	 *			  the encoding
 	 *
 	 * @throws IndexOutOfBoundsException
-	 *				when <code>length < 0, start < 0</code> or <code>start + length > data.length</code>
+	 *				when {@code length < 0, start < 0} or {@code start + length > data.length}
 	 * @throws UnsupportedEncodingException
 	 *				when encoding is not supported
 	 * @throws NullPointerException
@@ -912,7 +912,7 @@ public final class String implements Serializable, Comparable<String>, CharSeque
 	 *			  the number of characters to use
 	 *
 	 * @throws IndexOutOfBoundsException
-	 *				when <code>length < 0, start < 0</code> or <code>start + length > data.length</code>
+	 *				when {@code length < 0, start < 0} or {@code start + length > data.length}
 	 * @throws NullPointerException
 	 *				when data is null
 	 */
@@ -1927,7 +1927,7 @@ public final class String implements Serializable, Comparable<String>, CharSeque
 	 * @return the character at the index
 	 *
 	 * @throws IndexOutOfBoundsException
-	 *				when <code>index < 0</code> or <code>index >= length()</code>
+	 *				when {@code index < 0} or {@code index >= length()}
 	 */
 	/*[IF]*/
 	// TODO : Is it better to throw the AIOOB here implicitly and catch and rethrow?
@@ -2071,7 +2071,7 @@ public final class String implements Serializable, Comparable<String>, CharSeque
 	 *
 	 * @param string
 	 *			  a String
-	 * @return an int < 0 if this String is less than the specified String, 0 if they are equal, and > 0 if this String is greater
+	 * @return an {@code int < 0} if this String is less than the specified String, 0 if they are equal, and {@code > 0} if this String is greater
 	 */
 	public int compareToIgnoreCase(String string) {
 		String s1 = this;
@@ -2225,7 +2225,7 @@ public final class String implements Serializable, Comparable<String>, CharSeque
 	 * @return the new String
 	 *
 	 * @throws IndexOutOfBoundsException
-	 *				if <code>length < 0, start < 0</code> or <code>start + length > data.length</code>
+	 *				if {@code length < 0, start < 0} or {@code start + length > data.length}
 	 * @throws NullPointerException
 	 *				if data is null
 	 */
@@ -2505,7 +2505,7 @@ public final class String implements Serializable, Comparable<String>, CharSeque
 	 * @throws NullPointerException
 	 *				when data is null
 	 * @throws IndexOutOfBoundsException
-	 *				when <code>start < 0, end > length(), index < 0, end - start > data.length - index</code>
+	 *				when {@code start < 0, end > length(), index < 0, end - start > data.length - index}
 	 *
 	 * @deprecated Use getBytes() or getBytes(String)
 	 */
@@ -2607,8 +2607,8 @@ public final class String implements Serializable, Comparable<String>, CharSeque
 	 *			  the starting offset in the character array
 	 *
 	 * @throws IndexOutOfBoundsException
-	 *				when <code>start < 0, end > length(),
-	 * 				start > end, index < 0, end - start > buffer.length - index</code>
+	 *				when {@code start < 0, end > length(),
+	 * 				start > end, index < 0, end - start > buffer.length - index}
 	 * @throws NullPointerException
 	 *				when buffer is null
 	 */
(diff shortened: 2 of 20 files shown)


archanaNogriya:
@DanHeidinga @pshipton

archanaNogriya:
@DanHeidinga , 
I have made copyright fix adding 2018, still it fails for these files
17:28:42 The following files were modified and have incorrect copyrights
[Pipeline] echo
17:28:42 jcl/src/java.base/share/classes/java/lang/invoke/ConstantCallSite.java
[Pipeline] echo
17:28:42 jcl/src/java.base/share/classes/java/lang/invoke/MethodType.java
[Pipeline] echo
17:28:42 jcl/src/java.management/share/classes/java/lang/management/ClassLoadingMXBean.java
[Pipeline] echo
17:28:42 jcl/src/java.management/share/classes/java/lang/management/CompilationMXBean.java
[Pipeline] echo
17:28:42 jcl/src/java.management/share/classes/java/lang/management/GarbageCollectorMXBean.java
[Pipeline] echo
17:28:42 jcl/src/java.management/share/classes/java/lang/management/LockInfo.java
[Pipeline] echo
17:28:42 jcl/src/java.management/share/classes/java/lang/management/MemoryMXBean.java
[Pipeline] echo
17:28:42 jcl/src/java.management/share/classes/java/lang/management/MemoryManagerMXBean.java
[Pipeline] echo
17:28:42 jcl/src/java.management/share/classes/java/lang/management/MemoryPoolMXBean.java
[Pipeline] echo
17:28:42 jcl/src/java.management/share/classes/java/lang/management/OperatingSystemMXBean.java
[Pipeline] echo
17:28:42 jcl/src/java.management/share/classes/java/lang/management/PlatformLoggingMXBean.java
[Pipeline] echo
17:28:42 jcl/src/java.management/share/classes/java/lang/management/RuntimeMXBean.java
[Pipeline] echo
17:28:42 jcl/src/java.management/share/classes/java/lang/management/ThreadInfo.java

not sure what else I should be doing.

keithc-ca:
If there are already two dates in the copyright notice, you should only ensure the second is the current year, not add a third. For example, in `ConstantCallSite.java` we want
```
Copyright (c) 2011, 2018 IBM Corp. and others
```
not
```
Copyright (c) 2011, 2011, 2018 IBM Corp. and others
```

jcl/src/java.base/share/classes/java/lang/Class.java
```diff
@@ -2820,7 +2820,7 @@ public boolean isAnnotationPresent(Class<? extends Annotation> annotation) {
 
 /**
  * Cast this Class to a subclass of the specified Class. 
- * 
+ * @param <U> the type to cast this class object to
```
pshipton: There shouldn't be two different @param statements for the single parameter.

jcl/src/java.base/share/classes/java/lang/String.java
```diff
@@ -138,7 +138,7 @@ static void checkOffset(int offset, int length) {
 		 *			  an Object to compare
 		 * @param o2
 		 *			  an Object to compare
-		 * @return an int < 0 if object1 is less than object2, 0 if they are equal, and > 0 if object1 is greater
+		 * @return an int is negative 0 if object1 is less than object2, 0 if they are equal, and greater than 0 if object1 is greater
```
pshipton: "is negative" should be "less than". From "and greater than" remove "and"

jcl/src/java.base/share/classes/java/lang/StringBuffer.java
```diff
@@ -316,8 +316,8 @@ public synchronized StringBuffer append (char[] chars) {
  * @param		length	the number of characters
  * @return		this StringBuffer
  *
- * @exception	IndexOutOfBoundsException when <code>length < 0, start < 0</code> or
- *				<code>start + length > chars.length</code>
+ * @exception	IndexOutOfBoundsException
+ *				If {@code length} is negative, {@code start} is negative or ({@code start} + {@code start}) is grater {@code chars.length}
```
pshipton: `{@code start} + {@code start}` should be `{@code start + length}`

jcl/src/java.base/share/classes/java/lang/StringBuffer.java
```diff
@@ -774,8 +774,8 @@ int capacityInternal() {
  * @param 		index	the zero-based index in this StringBuffer
  * @return		the character at the index
  *
- * @exception	IndexOutOfBoundsException when <code>index < 0</code> or
- *				<code>index >= length()</code>
+ * @exception	IndexOutOfBoundsException
+ *              If {@code index} is negative or {@code index} is grater or equal length()
```
pshipton: typo in grater - greater

jcl/src/java.base/share/classes/java/lang/String.java
archanaNogriya: Thanks @pshipton , I thought I have replaced this with {@code} .. let me see why my changes not reflected.

jcl/src/java.base/share/classes/java/lang/StringBuffer.java
```diff
@@ -803,8 +803,8 @@ public synchronized char charAt(int index) {
  * @param		end	the offset one past the last character
  * @return		this StringBuffer
  *
- * @exception	StringIndexOutOfBoundsException when <code>start < 0, start > end</code> or
- *				<code>end > length()</code>
+ * @exception	StringIndexOutOfBoundsException
+ *				If {@code start} is negative, {@code start} is grater {@code end} or {@code end} is grater length()
```
pshipton: grater - greater than (two places)

jcl/src/java.base/share/classes/java/lang/StringBuffer.java
```diff
@@ -974,8 +974,10 @@ private void ensureCapacityImpl(int min) {
  * @param		buffer	the destination character array
  * @param		index	the starting offset in the character array
  *
- * @exception	IndexOutOfBoundsException when <code>start < 0, end > length(),
- *				start > end, index < 0, end - start > buffer.length - index</code>
+ * @exception	IndexOutOfBoundsException
+ *              If {@code start} is negative, {@code end} is greater length(),
```
pshipton: greater - greater than

jcl/src/java.base/share/classes/java/lang/invoke/MethodHandles.java
```diff
@@ -3705,7 +3705,7 @@ private static boolean counterPredicate(int endValue, int counter) {
 	}
 	
 	/**
-	 * Produce a loop handle that iterates over a range of values produced by an Iterator<T>
+	 * Produce a loop handle that iterates over a range of values produced by an Iterator object
```
DanHeidinga: We want to keep the `<T>` generic.  Changing it to `Iterator object` changes the meaning.  Maybe wrap it in `{@code Iterator<T> }`

jcl/src/java.management/share/classes/java/lang/management/LockInfo.java
```diff
@@ -145,7 +146,7 @@ public static LockInfo from(CompositeData compositeData) {
 	 * The string will hold both the name of the lock object's class and it's
 	 * identity hash code expressed as an unsigned hexadecimal. i.e.<br>
 	 * <p>
-	 * {@link #getClassName()}&nbsp;+&nbsp;&commat;&nbsp;+&nbsp;Integer.toHexString({@link #getIdentityHashCode()})
+	 * {@link #getClassName()} &nbsp;+&nbsp;&nbsp;+&nbsp;Integer.toHexString({@link #getIdentityHashCode()})
 	 * </p>
```
DanHeidinga: The `</p>` was deleted above.  Was that accidental or was this one missed?

jcl/src/java.base/share/classes/java/lang/String.java
archanaNogriya: @pshipton Looks like you are seeing bit old code changes. Let me see why it is not refreshed with latest.

jcl/src/java.base/share/classes/java/lang/invoke/MethodHandles.java
archanaNogriya: Hi Dan, yes I have done all those changes.. some how it is not seen or refresh those latest changes here .. wired .. I am checking  on it. you will see latest one .. these are bit old changes then I made everything with {@code}

pshipton:
Apparently the commit is out of date. Please update it and make a comment when its ready for review.

pshipton:
Also, please improve the content of the commit comment. It should have some context, and mention the changes are to javadoc comments.

jcl/src/java.base/share/classes/java/lang/Class.java
archanaNogriya: Agree I will fix that. thanks

jcl/src/java.base/share/classes/java/lang/StringBuffer.java
archanaNogriya: Some how I missed this line.. actually this should be fixed with this line,
when {@code index < 0} or {@code index >= length()}

jcl/src/java.base/share/classes/java/lang/StringBuffer.java
archanaNogriya: Some how I missed this line.. actually this should be fixed with this line,
when {@code}

jcl/src/java.management/share/classes/java/lang/management/LockInfo.java
archanaNogriya: In these cases build has been complaining  about close <p> tag unnecessary, while getting to understand why it should be a problem then found HTML never required the <p> tag to be closed—it was always optional. You may close your HTML tags so it looks like well-formed XHTML, but that isn't necessary. XHTML is more strict than HTML. and when I removed build was fine.

archanaNogriya:
@pshipton @DanHeidinga , this commit is ready to review now, Code is uptodate now with all my changes.

archanaNogriya:
@andrew-m-leonard

jcl/src/java.base/share/classes/java/lang/StringBuffer.java
```diff
@@ -774,8 +774,8 @@ int capacityInternal() {
  * @param 		index	the zero-based index in this StringBuffer
  * @return		the character at the index
  *
- * @exception	IndexOutOfBoundsException when <code>index < 0</code> or
- *				<code>index >= length()</code>
+ * @exception	IndexOutOfBoundsException
+ *              If {@code index<0} or {@code index >= length()}
```
pshipton: Please add spaces for `index < 0`

jcl/src/java.management/share/classes/java/lang/management/ClassLoadingMXBean.java
```diff
@@ -30,18 +30,17 @@
...
+ * Accessing this <code>MXBean</code> can be done in one of three ways.
  * <ol>
- * <li>Invoking the static ManagementFactory.getClassLoadingMXBean() method.
- * </li>
+ * <li>Invoking the static ManagementFactory.getClassLoadingMXBean() method.</li>
  * <li>Using a javax.management.MBeanServerConnection.</li>
  * <li>Obtaining a proxy MXBean from the static
  * ManagementFactory.newPlatformMXBeanProxy(MBeanServerConnection connection,
- * String mxbeanName, Class <T>mxbeanInterface) method, passing in
+ * String mxbeanName, {@code Class<T> mxbeanInterface())} method, passing in
```
pshipton: Seems `@code` should be around the entire API and not just the single parameter

jcl/src/java.management/share/classes/java/lang/management/LockInfo.java
```diff
@@ -145,7 +146,7 @@ public static LockInfo from(CompositeData compositeData) {
 	 * The string will hold both the name of the lock object's class and it's
 	 * identity hash code expressed as an unsigned hexadecimal. i.e.<br>
 	 * <p>
-	 * {@link #getClassName()}&nbsp;+&nbsp;&commat;&nbsp;+&nbsp;Integer.toHexString({@link #getIdentityHashCode()})
+	 * {@link #getClassName()} &nbsp;+&nbsp;&nbsp;+&nbsp;Integer.toHexString({@link #getIdentityHashCode()})
```
pshipton: Was it intentional to remove the `&comma`?

pshipton:
Since this is ready to review, please remove the WIP from the title.

archanaNogriya:
@pshipton : I was on vacation for last week, Let me make those changes as you suggested. thanks

jcl/src/java.management/share/classes/java/lang/management/LockInfo.java
archanaNogriya: No, have not, let me try this one replacing. thanks

archanaNogriya: @keithc-ca  thanks it worked :)

PR #1571, opened 2018-03-29
Remove BinaryCommutativeAnalyser from ARM build

knn-k:
BinaryCommutativeAnalyser has been linked in the JIT for ARM, but
its functions are called from nowhere.  This commit removes the
class from the ARM link target.


knn-k:
This change makes the JIT shared library for ARM smaller by about 2KB with no harm.

DanHeidinga:
FYI @JamesKingdon

JamesKingdon:
Thanks Konno-san, good find. I think we should probably have a comment block in omr/compiler/arm/codegen/BinaryCommutativeAnalyser.cpp calling out that it isn't currently being used. Keeping the file seems like a good idea, at least for now, in case we find we can benefit from it somewhere when we start looking into the performance issues.

knn-k:
I agree with adding a comment block in BinaryCommutativeAnalyser.cpp for ARM.
Is there any file with such comment that is kept in the repository but unused?

JamesKingdon:
I doubt it, but I have no issue with being the first.

knn-k:
The Travis CI build of this PR failed due to "502 Bad Gateway" at sourceforge.net, which was a temporary trouble, I believe.  How can I kick it again?

JamesKingdon:
I think that a committers can trigger builds from the comments. Otherwise I suspect you have to push another change to trigger a new build.

DanHeidinga:
I've restarted the travis build

mstoodle:
@knn-k sorry for the long gap on this one. Would you please rebase if you still think this is an appropriate change?

knn-k:
Rebased to the latest level.

knn-k:
@mstoodle This one is ready to merge three weeks ago.

PR #1573, opened 2018-03-29
Read Barrier for CompareAndSwapObject

amicic:
Compare-And-Swap on object reference (like one used in sun/misc/Unsafe
but also any other used internally by JVM), while primarily is a store
operation, it is also an implicit read (it reads the existing value to
be compared with a provided compare value, before the store itself),
hence needs a read barrier.

This change introduces the pre-read barrier in all VM/GC
CompareAndSwapObject helpers, that is used in Evacuating read barrier
triggered by the read value, used in Concurrent Scavenger.

An interesting dilemma is which out of two pre operations (read and
store) is to be executed first. Currently in OpenJ9, we do not have any
collector that have both of them active at the same time, so the order
ATM is not important.

However, in theory (and it actually may happen later in OpenJ9), they
could both be active. An example, for Gencon GC policy, is the same read
barrier on reference value (Concurrent Scavenger) and
Snap-Shot-At-The-Beginning store barrier (for concurrent global
marking). Since read barrier can modify the value, and SATB cares about
the value being overwritten, read barrier should go first to bring it up
to date, before SATB evaluates it. On the other side, from SATB
perspective, we do not care about the very original value of the field
(even before read barrier modifies it), since it is an object being
evacuated/collected and is not part of the Snapshot (it is the copy of
that object that we want to be preserved as part of the Snapshot).

### DIFF
--- a/runtime/gc_base/ObjectAccessBarrier.cpp
+++ b/runtime/gc_base/ObjectAccessBarrier.cpp
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 1991, 2017 IBM Corp. and others
+ * Copyright (c) 1991, 2018 IBM Corp. and others
  *
  * This program and the accompanying materials are made available under
  * the terms of the Eclipse Public License 2.0 which accompanies this
@@ -1501,6 +1501,7 @@ MM_ObjectAccessBarrier::compareAndSwapObject(J9VMThread *vmThread, J9Object *des
 	fj9object_t *actualDestAddress;
 	fj9object_t compareValue = convertTokenFromPointer(compareObject);
 	fj9object_t swapValue = convertTokenFromPointer(swapObject);
+	bool result = false;
 
 	/* TODO: To make this API more consistent, it should probably be split into separate
 	 * indexable and non-indexable versions. Currently, when called on an indexable object,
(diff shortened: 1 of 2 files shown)


runtime/gc_base/ObjectAccessBarrier.cpp
```diff
@@ -1513,25 +1514,26 @@ MM_ObjectAccessBarrier::compareAndSwapObject(J9VMThread *vmThread, J9Object *des
...
-		postObjectStore(vmThread, destObject, actualDestAddress, swapObject, true);
+	if (preObjectRead(vmThread, destObject, actualDestAddress)) {
+		/* Note: This is a bit of a special case -- we call preObjectStore even though the store
+		 * may not actually occur. This is safe and correct for Metronome.
+		 */
+		preObjectStore(vmThread, destObject, actualDestAddress, swapObject, true);
+		protectIfVolatileBefore(vmThread, true, false, false);
+
+		/* One of these conditions will evaluate to true at compile time based on whether fields are 32 or 64 bits */
+		if (sizeof(fj9object_t) == sizeof(UDATA)) {
```
charliegracie: Could this just become a compile time check with a #error in the else?

```
#if (sizeof(fj9object) == sizeof(UDATA))
....
```

Not sure if sizeof gets evaluated at the right time though.

amicic: I did not change this (it only got indented), but if we are to change, I believe it the cleanest is to relay on 
#if defined (OMR_GC_COMPRESSED_POINTERS)
and then there is no need to error clause at all

amicic: Replaced the check with J9VM_GC_COMPRESSED_POINTERS (and swapped the order of two lines)

charliegracie:
@dmitripivkine you can merge at your leisure.

Issue #1552, opened 2018-03-27
Unreleased memory allocations from JIT at shutdown


dmitripivkine:
There is a number of messages about unreleased malloc'ed memory segments from JIT in run with -Xcheck:memory. Would you please check that non of them might be a reason for memory leak?

There are some of them:
MethodMetaData.c:149 (mapTable in TR_StackMapTable())
CompilationThread.cpp:758 (_activeThreadName in CompilationInfoPerThread())
CompilationThread.cpp:759 (_suspendedThreadName in CompilationInfoPerThread())
rossa.cpp:1264 (codeCacheManager in initializePersistentMemory())

Also there are a few allocations from RawAllocator.hpp:71 which I did not track to the consumer.

An ultimate goal is cleanup code to eliminate these messages. Currently it prevents to run tests with -Xcheck:memory with JIT enabled modes.

dmitripivkine:
@smlambert FYI

mpirvu:
The JIT is not doing a thorough job at deallocating everything at shutdown on the account that the process is going away anyway and deallocating every single piece individually may increase shutdown time.
These 3 should be fine because there is a very limited number of them
```
CompilationThread.cpp:758 (_activeThreadName in CompilationInfoPerThread())
CompilationThread.cpp:759 (_suspendedThreadName in CompilationInfoPerThread())
rossa.cpp:1264 (codeCacheManager in initializePersistentMemory())
```
The one in MethodMetaData.c could potentially lead to a memory leak as far as I ca tell, so we'll need to investigate it.

mpirvu:
Related to `MethodMetaData.c`, the code allocates a cache that should speed up stack walking. Not all methods will allocate this cache (there is some heuristic involved because of footprint concerns).
This cache is attached to the jittedBodyInfo and the JIT tries to deallocate it at least for recompilations (I have to make sure this is done for class unloading as well).
At shutdown inevitably there will be Java methods active that may have some cache attached to them. I can see in the code that we had an attempt of freeing all those caches by scanning all the active methods, but that code is commented out with this message
```
/*  Commented out because of rare failure during shutdown: 90700: JIT_Sanity.TestSCCMLTests1.Mode110 Crash during freeJITconfig()
    Deallocating malloced memory is not going to very useful even for zOS
```

harryyu1994:
[...]

PR #1577, opened 2018-03-29
Fix potential memory leak issue with mapTable

harryyu1994:
- the jittedBodyInfo contains a pointer to mapTable(mapTable is a cache used for speeding up stack walking)
- when bodyInfo was freed in J9::CodeCache::addFreeBlock() during class unloading  we should also free mapTable but we did not
- used freeFastWalkCache() to free its mapTable

Issue: #1552


mpirvu:
Jenkins test all

PR #1579, opened 2018-03-29
WIP Check type_index of runtime visible type annotations

pdbain-ibm:
If a type annotation's type_index does not refer to a CONSTANT_UTF8_info
structure, force it to zero to prevent constant pool entries being incorrectly
marked as used by annotations.

FYI @fengxue-IS 

### DIFF
--- a/runtime/bcutil/ClassFileOracle.cpp
+++ b/runtime/bcutil/ClassFileOracle.cpp
@@ -746,15 +746,21 @@ void
 ClassFileOracle::walkTypeAnnotations(U_16 annotationsCount, J9CfrTypeAnnotation *typeAnnotations) {
 	for (U_16 typeAnnotationIndex = 0; typeAnnotationIndex < annotationsCount; ++ typeAnnotationIndex) {
 		J9CfrAnnotation *annotation = &(typeAnnotations[typeAnnotationIndex].annotation);
-		markConstantAsUsedByAnnotation(annotation->typeIndex);
-		const U_16 elementValuePairsCount = annotation->numberOfElementValuePairs;
-		for (U_16 elementValuePairsIndex = 0; (elementValuePairsIndex < elementValuePairsCount) && (OK == _buildResult); ++elementValuePairsIndex) {
-			markConstantAsUsedByAnnotation(annotation->elementValuePairs[elementValuePairsIndex].elementNameIndex);
-			walkAnnotationElement(annotation->elementValuePairs[elementValuePairsIndex].value);
+		/* type_index in an annotation must refer to a CONSTANT_UTF8_info structure. */
+		if (getCPTag(annotation->typeIndex) == CFR_CONSTANT_Utf8) {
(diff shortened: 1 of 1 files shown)


pshipton:
jenkins test all plinux

pdbain-ibm:
Indentation amended.

DanHeidinga:
For future reference, see https://docs.oracle.com/javase/specs/jvms/se8/html/jvms-4.html#jvms-4.7.16 for the specification of the type_index field.

DanHeidinga:
Doesn't `walkAnnotations()` have the same issue?  Shouldn't it be validating the `type_index` is a UTF8 as well?

Where is the code that detects the error for `type_index = 0`?  Does this prevent reading subsequent annotations from the `annotations[]` (ie: prevent reading this annotation or all)?

pdbain-ibm:
We realized that 0 is a valid constant pool entry. Closing this while we develop another solution.

pdbain-ibm:
> Where is the code that detects the error for type_index = 0? 

That is handled by the annotation parser (see below).

> Shouldn't it be validating the type_index is a UTF8 as well?

Yes.  Our new algorithm handles all annotations.

> Does this prevent reading subsequent annotations from the annotations[] (ie: prevent reading this annotation or all)?

Yes. I submit that this behaviour is reasonable as the classfile is not well formed, so there are no guarantees.

I hacked a class with a bogus type annotation on a formal parameter and tried to get the parameter type annotations:
`
$ java TestAnnotations AnnClass
method: m
annotation: @VisibleAnn()
annotation: @VisibleAnn()
Exception in thread "main" java.lang.annotation.AnnotationFormatError: Could not parse bytes for type annotations
	at sun.reflect.annotation.TypeAnnotationParser.parseTargetInfo(TypeAnnotationParser.java:489)
	at sun.reflect.annotation.TypeAnnotationParser.parseTypeAnnotation(TypeAnnotationParser.java:397)
	at sun.reflect.annotation.TypeAnnotationParser.parseTypeAnnotations(TypeAnnotationParser.java:341)
	at sun.reflect.annotation.TypeAnnotationParser.buildAnnotatedTypes(TypeAnnotationParser.java:111)
	at java.lang.reflect.Executable.getAnnotatedParameterTypes(Executable.java:688)
	at TestAnnotations.main(TestAnnotations.java:20)
[1579.zip](https://github.com/eclipse/openj9/files/1871629/1579.zip)
`

DanHeidinga:
Do OpenJ9 and the RI have the same behaviour for the test case?

pdbain-ibm:
Mostly, though not exactly: they throw the same exception (java.lang.annotation.AnnotationFormatError), but with a different cause:

> Exception in thread "main" java.lang.annotation.AnnotationFormatError: java.lang.IllegalArgumentException: Wrong type at constant pool index

The handling of malformed type annotations is not specified in the documentation or compliance tests.

PR #1580, opened 2018-03-29
Fix Metadata Reclamation Bug

dsouzai:
When the metadata is reclaimed (due to a recompilation of a method), the
previous and next metadatas in the linked list are are updated in
updateExceptionTableLinkedList. However, there's a corner case that
wasn't handled. If the metadata being reclaimed is the first in the
list, then the pointer to it (either off the classloader or the class
(if anonymous)) was not updated to point to the new stub metadata. This
PR rectifies this.

### DIFF
--- a/runtime/compiler/trj9/runtime/HookHelpers.cpp
+++ b/runtime/compiler/trj9/runtime/HookHelpers.cpp
@@ -141,15 +141,36 @@ namespace  {
    
    // We need to update the nextMethod and prevMethod pointers of the J9JITExceptionTable that
    // point to the old J9JITExceptionTable to now point to the new (stub) J9JITExceptionTable.
-   inline void updateExceptionTableLinkedList(J9JITExceptionTable *stubMetadata)
+   inline void updateExceptionTableLinkedList(J9VMThread *vmThread, J9JITExceptionTable *oldMetadata, J9JITExceptionTable *newMetadata)
       {
-      J9JITExceptionTable *prev = stubMetadata->prevMethod;
-      J9JITExceptionTable *next = stubMetadata->nextMethod;
-      
+      // newMetaData's prevMethod and nextMethod should be the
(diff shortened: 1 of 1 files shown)


dsouzai:
WIP because I have yet to run tests.

@mpirvu could you review / verify for correctness.

runtime/compiler/trj9/runtime/HookHelpers.cpp
```diff
@@ -141,15 +141,37 @@ namespace  {
...
+   inline void updateExceptionTableLinkedList(J9VMThread *vmThread, J9JITExceptionTable *oldMetadata, J9JITExceptionTable *newMetadata)
       {
-      J9JITExceptionTable *prev = stubMetadata->prevMethod;
-      J9JITExceptionTable *next = stubMetadata->nextMethod;
+      // newMetaData's prevMethod and nextMethod should be the
+      // same as oldMetaData's due to the memcpy
+      J9JITExceptionTable *prev = oldMetadata->prevMethod;
+      J9JITExceptionTable *next = oldMetadata->nextMethod;
       
+      J9Class *j9class = J9_CLASS_FROM_METHOD(oldMetadata->ramMethod);
```
mpirvu: I would move these 2 statement in the `else` part, closer to where they are used.

dsouzai:
@mpirvu this should be ok to merge after the usual jenkins testing.

dsouzai:
@mpirvu I talked to Adam about the infra job failure on linux 390, and he said to relaunch it; I think the command is `Jenkins test extended zlinux jdk8`

mpirvu:

mpirvu:
The failure for JDK8 linux_390 is due to infrastructure (not a real test failure). Also, Irwin tells me that his personal build had no failure on linux_390. Based on this I am going to merge this PR.

PR #1589, opened 2018-03-31
Relocatable Debug Counters

dsouzai:
Depends on https://github.com/eclipse/omr/pull/2408

I tested this with a simple java program:
```
import java.lang.*;

public class SimpleAddTest {

   private static int iter;
   private static int count;

   public static void addNum(int num) {
         count += num;
   }

   public static void main (String[] args) {

      iter = 100;
      count = 0;
      
      int num = 1;

      int i = 0;
      while (i < iter) {
         addNum(num);
	 i++;
      }

      System.out.println(count);
   }
}
```

This just does an add 100 times and prints out the total (which is 100). I added a dynamic debug counter in simplifier called "CounterTest".

**JIT**
```
$> java '-Xjit:optLevel=cold,count=0,limit={SimpleAddTest.addNum*},debugCounters={CounterTest},verbose={compilePerformance},vlog=vlog' -Xshareclasses:name=debugCounterTest SimpleAddTest
100

== Dynamic debug counters ==
  0: CounterTest |          100 |   __   0 __

 (cold) Compiling SimpleAddTest.addNum(I)V  OrdinaryMethod j9m=0000000002449B58 t=150 compThread=0 memLimit=262144 KB freePhysicalMemory=5228 MB
+ (cold) SimpleAddTest.addNum(I)V @ 00007F0826600034-00007F0826600065 OrdinaryMethod - Q_SZ=0 Q_SZI=0 QW=2 j9m=0000000002449B58 bcsz=9 sync time=695us mem=[region=704 system=16384]KB compThread=0 CpuLoad=73%(18%avg) JvmCpu=0%
```
 
**AOT Compile run**
```
$> java '-Xaot:forceAOT,optLevel=cold,count=0,limit={SimpleAddTest.addNum*},debugCounters={CounterTest},verbose={compilePerformance},vlog=vlog' -Xshareclasses:name=debugCounterTest SimpleAddTest
100

== Dynamic debug counters ==
  0: CounterTest |          100 |   __   0 __

 (AOT cold) Compiling SimpleAddTest.addNum(I)V  OrdinaryMethod j9m=00000000025C4658 t=275 compThread=0 memLimit=262144 KB freePhysicalMemory=5227 MB
+ (AOT cold) SimpleAddTest.addNum(I)V @ 00007F7398A00034-00007F7398A00076 OrdinaryMethod - Q_SZ=0 Q_SZI=0 QW=2 j9m=00000000025C4658 bcsz=9 sync time=791us mem=[region=1408 system=16384]KB compThread=0 CpuLoad=73%(18%avg) JvmCpu=0%
```

**AOT Load run**
```
$> java '-Xaot:forceAOT,optLevel=cold,count=0,limit={SimpleAddTest.addNum*},debugCounters={CounterTest},verbose={compilePerformance},vlog=vlog' -Xshareclasses:name=debugCounterTest SimpleAddTest
100

== Dynamic debug counters ==
  0: CounterTest |          100 |   __   0 __

 (cold) Compiling SimpleAddTest.addNum(I)V  OrdinaryMethod j9m=0000000001D5A658 t=179 compThread=0 memLimit=262144 KB freePhysicalMemory=5227 MB
+ (AOT load) SimpleAddTest.addNum(I)V @ 00007F1D7A400034-00007F1D7A400076 Q_SZ=0 Q_SZI=0 QW=1 j9m=0000000001D5A658 bcsz=9 time=153us compThread=0
```

There's still a bug somewhere though, since if I enable all debug counters, the number printed out by the java code is wrong (though the counters seem ok :P)

**JIT**
```
$> java '-Xjit:optLevel=cold,count=0,limit={SimpleAddTest.addNum*},debugCounters={*},verbose={compilePerformance},vlog=vlog' -Xshareclasses:name=debugCounterTest SimpleAddTest
100

== Dynamic debug counters ==
  0: CounterTest                                                |          100 |   __   0 __
  4: blocks/coldCompiles/warmBlocks/=2                          |          100 |    100.00% |   100.00% |   100.00% |  __   4 __
  5: cg.blocks                                                  |          100 |   __   5 __
  7: cg.epilogues:no-preservedRegStoreBytesSaved                |          100 |    100.00% |  __   7 __
  8: cg.prologues                                               |          100 |   __   8 __
  9: cg.prologues:#frameBytes                                   |          800 |      8.00  |  __   9 __
 10: cg.prologues:#instructionBytes                             |        10800 |    108.00  |  __  10 __
 11: cg.prologues:#peakBytes                                    |          800 |      8.00  |  __  11 __
 12: cg.prologues:inline                                        |          100 |    100.00% |  __  12 __
 13: cg.prologues:no-preservedRegStoreBytesSaved                |          100 |    100.00% |  __  13 __
 21: warmBlocks/byJittedBody/(SimpleAddTest.addNum(I)V)/cold/=2 |          100 |    100.00% |   100.00% |   100.00% |   100.00% |  __  21 __
```

**AOT Compile run**
```
$> java '-Xaot:forceAOT,optLevel=cold,count=0,limit={SimpleAddTest.addNum*},debugCounters={*},verbose={compilePerformance},vlog=vlog' -Xshareclasses:name=debugCounterTest SimpleAddTest
-2030234688

== Dynamic debug counters ==
  0: CounterTest                                                |          100 |   __   0 __
  4: blocks/coldCompiles/warmBlocks/=2                          |          100 |    100.00% |   100.00% |   100.00% |  __   4 __
  5: cg.blocks                                                  |          100 |   __   5 __
  7: cg.epilogues:no-preservedRegStoreBytesSaved                |          100 |    100.00% |  __   7 __
  8: cg.prologues                                               |          100 |   __   8 __
  9: cg.prologues:#frameBytes                                   |          800 |      8.00  |  __   9 __
 10: cg.prologues:#instructionBytes                             |        10400 |    104.00  |  __  10 __
 11: cg.prologues:#peakBytes                                    |          800 |      8.00  |  __  11 __
 12: cg.prologues:inline                                        |          100 |    100.00% |  __  12 __
 13: cg.prologues:no-preservedRegStoreBytesSaved                |          100 |    100.00% |  __  13 __
 21: warmBlocks/byJittedBody/(SimpleAddTest.addNum(I)V)/cold/=2 |          100 |    100.00% |   100.00% |   100.00% |   100.00% |  __  21 __
```

**AOT Load run**
```
$> java '-Xaot:forceAOT,optLevel=cold,count=0,limit={SimpleAddTest.addNum*},debugCounters={*},verbose={compilePerformance},vlog=vlog' -Xshareclasses:name=debugCounterTest SimpleAddTest
1972250816

== Dynamic debug counters ==
  0: CounterTest                                                |          100 |   __   0 __
  4: blocks/coldCompiles/warmBlocks/=2                          |          100 |    100.00% |   100.00% |   100.00% |  __   4 __
  5: cg.blocks                                                  |          100 |   __   5 __
  7: cg.epilogues:no-preservedRegStoreBytesSaved                |          100 |    100.00% |  __   7 __
  8: cg.prologues                                               |          100 |   __   8 __
  9: cg.prologues:#frameBytes                                   |          800 |      8.00  |  __   9 __
 10: cg.prologues:#instructionBytes                             |        10400 |    104.00  |  __  10 __
 11: cg.prologues:#peakBytes                                    |          800 |      8.00  |  __  11 __
 12: cg.prologues:inline                                        |          100 |    100.00% |  __  12 __
 13: cg.prologues:no-preservedRegStoreBytesSaved                |          100 |    100.00% |  __  13 __
 18: warmBlocks/byJittedBody/(SimpleAddTest.addNum(I)V)/cold/=2 |          100 |    100.00% |   100.00% |   100.00% |   100.00% |  __  18 __
```

Credit to @lmaisons for the idea of using the metadata to store the debug counter names.

### DIFF
--- a/runtime/compiler/env/J9SharedCache.hpp
+++ b/runtime/compiler/env/J9SharedCache.hpp
@@ -69,6 +69,9 @@ public:
 
    UDATA *rememberClass(J9Class *clazz, bool create=true);
 
+   UDATA rememberDebugCounterName(const char *name);
+   const char *getDebugCounterName(UDATA offset);
+
    bool classMatchesCachedVersion(J9Class *clazz, UDATA *chainData=NULL);
    bool classMatchesCachedVersion(TR_OpaqueClassBlock *classPtr, UDATA *chainData=NULL)
       {
--- a/runtime/compiler/p/codegen/J9AheadOfTimeCompile.cpp
+++ b/runtime/compiler/p/codegen/J9AheadOfTimeCompile.cpp
@@ -38,6 +38,7 @@
 #include "il/Node_inlines.hpp"
 #include "il/SymbolReference.hpp"
 #include "il/symbol/LabelSymbol.hpp"
+#include "ras/DebugCounter.hpp"
 
 J9::Power::AheadOfTimeCompile::AheadOfTimeCompile(TR::CodeGenerator *cg) :
          J9::AheadOfTimeCompile(_relocationTargetTypeToHeaderSizeMap, cg->comp()),
@@ -801,6 +802,35 @@ uint8_t *J9::Power::AheadOfTimeCompile::initializeAOTRelocationHeader(TR::Iterat
          }
          break;
 
+      case TR_DebugCounter:
+         {
+         TR::DebugCounterBase *counter = (TR::DebugCounterBase *) relocation->getTargetAddress();
+         if (!counter || !counter->getReloData() || !counter->getName())
+            comp->failCompilation<TR::CompilationException>("Failed to generate debug counter relo data");
+
+         TR::DebugCounterReloData *counterReloData = counter->getReloData();
+
+         uintptrj_t offset = (uintptrj_t)fej9->sharedCache()->rememberDebugCounterName(counter->getName());
+
+         uint8_t flags = (uint8_t)counterReloData->_seqKind;
+         TR_ASSERT((flags & RELOCATION_CROSS_PLATFORM_FLAGS_MASK) == 0,  "reloFlags bits overlap cross-platform flags bits\n");
+         *flagsCursor |= (flags & RELOCATION_RELOC_FLAGS_MASK);
+
+         *(uintptrj_t *)cursor = (uintptrj_t)counterReloData->_callerIndex;
+         cursor += SIZEPOINTER;
+         *(uintptrj_t *)cursor = (uintptrj_t)counterReloData->_bytecodeIndex;
+         cursor += SIZEPOINTER;
+         *(uintptrj_t *)cursor = offset;
+         cursor += SIZEPOINTER;
+         *(uintptrj_t *)cursor = (uintptrj_t)counterReloData->_delta;
+         cursor += SIZEPOINTER;
+         *(uintptrj_t *)cursor = (uintptrj_t)counterReloData->_fidelity;
+         cursor += SIZEPOINTER;
+         *(uintptrj_t *)cursor = (uintptrj_t)counterReloData->_staticDelta;
+         cursor += SIZEPOINTER;
+         }
+         break;
+
       }
    return cursor;
    }
@@ -867,6 +897,7 @@ uint32_t J9::Power::AheadOfTimeCompile::_relocationTargetTypeToHeaderSizeMap[TR_
    0,                                        // TR_NativeMethodAbsolute                = 56,
    0,                                        // TR_NativeMethodRelative                = 57,
    32,                                       // TR_ArbitraryClassAddress               = 58,
+   56,                                       // TR_DebugCounter                        = 59
    };
 #else
 uint32_t J9::Power::AheadOfTimeCompile::_relocationTargetTypeToHeaderSizeMap[TR_NumExternalRelocationKinds] =
@@ -930,6 +961,7 @@ uint32_t J9::Power::AheadOfTimeCompile::_relocationTargetTypeToHeaderSizeMap[TR_
    0,                                        // TR_NativeMethodAbsolute                = 56,
    0,                                        // TR_NativeMethodRelative                = 57,
    16,                                       // TR_ArbitraryClassAddress               = 58,
+   28                                        // TR_DebugCounter                        = 59
    };
 
 #endif
--- a/runtime/compiler/arm/codegen/J9AheadOfTimeCompile.cpp
+++ b/runtime/compiler/arm/codegen/J9AheadOfTimeCompile.cpp
@@ -651,6 +651,35 @@ uint8_t *J9::ARM::AheadOfTimeCompile::initializeAOTRelocationHeader(TR::Iterated
          }
          break;
 
+      case TR_DebugCounter:
+         {
+         TR::DebugCounterBase *counter = (TR::DebugCounterBase *) relocation->getTargetAddress();
+         if (!counter || !counter->getReloData() || !counter->getName())
+            comp->failCompilation<TR::CompilationException>("Failed to generate debug counter relo data");
+
+         TR::DebugCounterReloData *counterReloData = counter->getReloData();
+
+         uintptrj_t offset = (uintptrj_t)fej9->sharedCache()->rememberDebugCounterName(counter->getName());
+
+         uint8_t flags = (uint8_t)counterReloData->_seqKind;
+         TR_ASSERT((flags & RELOCATION_CROSS_PLATFORM_FLAGS_MASK) == 0,  "reloFlags bits overlap cross-platform flags bits\n");
+         *flagsCursor |= (flags & RELOCATION_RELOC_FLAGS_MASK);
+
+         *(uintptrj_t *)cursor = (uintptrj_t)counterReloData->_callerIndex;
+         cursor += SIZEPOINTER;
+         *(uintptrj_t *)cursor = (uintptrj_t)counterReloData->_bytecodeIndex;
+         cursor += SIZEPOINTER;
+         *(uintptrj_t *)cursor = offset;
+         cursor += SIZEPOINTER;
+         *(uintptrj_t *)cursor = (uintptrj_t)counterReloData->_delta;
+         cursor += SIZEPOINTER;
+         *(uintptrj_t *)cursor = (uintptrj_t)counterReloData->_fidelity;
+         cursor += SIZEPOINTER;
+         *(uintptrj_t *)cursor = (uintptrj_t)counterReloData->_staticDelta;
+         cursor += SIZEPOINTER;
+         }
+         break;
+
       }
       return cursor;
    }
@@ -717,6 +746,7 @@ uint32_t J9::ARM::AheadOfTimeCompile::_relocationTargetTypeToHeaderSizeMap[TR_Nu
    0,                                        // TR_NativeMethodAbsolute                = 56,
    0,                                        // TR_NativeMethodRelative                = 57,
    16,                                       // TR_ArbitraryClassAddress               = 58,
+   28                                        // TR_DebugCounter                        = 59
    };
 
 
--- a/runtime/compiler/codegen/J9AheadOfTimeCompile.cpp
+++ b/runtime/compiler/codegen/J9AheadOfTimeCompile.cpp
@@ -937,6 +937,41 @@ J9::AheadOfTimeCompile::dumpRelocationData()
                   }
                }
             break;
+         case TR_DebugCounter:
+            cursor ++;
+            if (is64BitTarget)
+               {
+               cursor += 4;     // padding
+               ep1 = cursor;    // inlinedSiteIndex
+               ep2 = cursor+8;  // bcIndex
+               ep3 = cursor+16; // offsetOfNameString
+               ep4 = cursor+24; // delta
+               ep5 = cursor+40; // staticDelta
+               cursor += 48;
+               self()->traceRelocationOffsets(cursor, offsetSize, endOfCurrentRecord, orderedPair);
+               if (isVerbose)
+                  {
+                  traceMsg(self()->comp(), "\n Debug Counter: Inlined site index = %d, bcIndex = %d, offsetOfNameString = %p, delta = %d, staticDelta = %d",
+                                   *(int64_t *)ep1, *(int32_t *)ep2, *(UDATA *)ep3, *(int32_t *)ep4, *(int32_t *)ep5);
+                  }
+               }
+            else
+               {
+               ep1 = cursor;    // inlinedSiteIndex
+               ep2 = cursor+4;  // bcIndex
+               ep3 = cursor+8;  // offsetOfNameString
+               ep4 = cursor+12; // delta
+               ep5 = cursor+20; // staticDelta
+               cursor += 24;
+               self()->traceRelocationOffsets(cursor, offsetSize, endOfCurrentRecord, orderedPair);
+               if (isVerbose)
+                  {
+                  traceMsg(self()->comp(), "\n Debug Counter: Inlined site index = %d, bcIndex = %d, offsetOfNameString = %p, delta = %d, staticDelta = %d",
+                                   *(int32_t *)ep1, *(int32_t *)ep2, *(UDATA *)ep3, *(int32_t *)ep4, *(int32_t *)ep5);
+                  }
+               }
+            break;
+
          default:
             traceMsg(self()->comp(), "Unknown Relocation type = %d\n", kind);
             TR_ASSERT(false, "should be unreachable");
--- a/runtime/compiler/env/J9SharedCache.cpp
+++ b/runtime/compiler/env/J9SharedCache.cpp
@@ -526,6 +526,39 @@ TR_J9SharedCache::rememberClass(J9Class *clazz, bool create)
    return chainData;
    }
 
+UDATA
+TR_J9SharedCache::rememberDebugCounterName(const char *name)
+   {
+   TR_J9VMBase *fej9 = (TR_J9VMBase *)(fe());
+   J9VMThread *vmThread = fej9->getCurrentVMThread();
+
+   J9SharedDataDescriptor dataDescriptor;
+   dataDescriptor.address = (U_8*)name;
+   dataDescriptor.length  = (strlen(name) + 1); // +1 for the \0 terminator
+   dataDescriptor.type    = J9SHR_DATA_TYPE_JITHINT;
+   dataDescriptor.flags   = J9SHRDATA_NOT_INDEXED;
+
+   const U_8 *data = sharedCacheConfig()->storeSharedData(vmThread,
+                                        (const char*)NULL,
+                                        0,
+                                        &dataDescriptor);
+
+   UDATA offset = data ? (UDATA)offsetInSharedCacheFromPointer((void *)data) : (UDATA)-1;
+
+   //printf("\nrememberDebugCounterName: Tried to store %s (%p), data=%p, offset=%p\n", name, name, data, offset);
+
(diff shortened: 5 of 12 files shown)


dsouzai:
@mstoodle Could you take a look at this and https://github.com/eclipse/omr/pull/2408

@fjeremic @ymanton I'm gonna your help to get this going on P and Z.

fyi @andrewcraik @jdmpapin 


Eventually I'll try to figure out reason for the java code not functioning as expected, but I figure it's good to get the initial code reviewed.

dsouzai:
@mstoodle Could you review the design
@andrewcraik Could you review the x component
@ymanton Could you review the p component
@fjeremic Could you review the z component

dsouzai:
The problem of the JIT code getting messed up when running with all debug counters on X is due to a limitation of the JIT; with AOT, the JIT changes `inc` instructions to two instructions (reg mov + inc). Some of the cg.prologue counters are generated after RA which results in unallocated registers as seen in this log:

```
0x7f95ab89f042 0000000e [0x7f95aa835f30] 48 b8 80 94 91 bb 95 7f 00 00      mov GPR_0096, 0x00007f95bb919480    # MOV8RegImm64
0x7f95ab89f04c 00000018 [0x7f95aa835030] ff 04 05 00 00 00 00               inc dword ptr [1*GPR_0096]          # INC4Mem
0x7f95ab89f053 0000001f [0x7f95aa836080] 48 b8 10 97 91 bb 95 7f 00 00      mov GPR_0112, 0x00007f95bb919710    # MOV8RegImm64
0x7f95ab89f05d 00000029 [0x7f95aa8351f0] ff 04 05 00 00 00 00               inc dword ptr [1*GPR_0112]          # INC4Mem
0x7f95ab89f064 00000030 [0x7f95aa835290] 48 83 ec 08                        sub rsp, 0x00000008 # SUB8RegImms
0x7f95ab89f068 00000034 [0x7f95aa8361d0] 48 b8 10 99 91 bb 95 7f 00 00      mov GPR_0128, 0x00007f95bb919910    # MOV8RegImm64
0x7f95ab89f072 0000003e [0x7f95aa8354d0] ff 04 05 00 00 00 00               inc dword ptr [1*GPR_0128]          # INC4Mem
0x7f95ab89f079 00000045 [0x7f95aa836320] 48 b8 a0 9a 91 bb 95 7f 00 00      mov GPR_0144, 0x00007f95bb919aa0    # MOV8RegImm64
0x7f95ab89f083 0000004f [0x7f95aa835690] ff 04 05 00 00 00 00               inc dword ptr [1*GPR_0144]          # INC4Mem
0x7f95ab89f08a 00000056 [0x7f95aa836470] 48 b8 a0 9c 91 bb 95 7f 00 00      mov GPR_0160, 0x00007f95bb919ca0    # MOV8RegImm64
0x7f95ab89f094 00000060 [0x7f95aa835850] ff 04 05 00 00 00 00               inc dword ptr [1*GPR_0160]          # INC4Mem
0x7f95ab89f09b 00000067 [0x7f95aa834830]                                    vfpSave                     # AdjustFramePtr
0x7f95ab89f09b 00000067 [0x7f95aa8365c0] 48 b8 a0 9e 91 bb 95 7f 00 00      mov GPR_0176, 0x00007f95bb919ea0    # MOV8RegImm64
0x7f95ab89f0a5 00000071 [0x7f95aa835a10] ff 04 05 00 00 00 00               inc dword ptr [1*GPR_0176]          # INC4Mem
0x7f95ab89f0ac 00000078 [0x7f95aa831b90]                                    assocRegs                       # ASSOCREGS
0x7f95ab89f0ac 00000078 [0x7f95aa831930]                                    Label L0016:                        # LABEL
```
As such, until this underlying problem is fixed, the cg.prologue debug counters won't work.

fjeremic:
> The problem of the JIT code getting messed up when running with all debug counters on X is due to a limitation of the JIT; with AOT, the JIT changes inc instructions to two instructions (reg mov + inc). Some of the cg.prologue counters are generated after RA which results in unallocated registers as seen in this log:

This entire problem is fixed by implementing https://github.com/eclipse/omr/issues/397 on x86 and Power. We had the exact same problem on Z and hence why we redesigned the codegen part of the debug counter infrastructure.

fjeremic:
> The problem of the JIT code getting messed up when running with all debug counters on X is due to a limitation of the JIT; with AOT, the JIT changes inc instructions to two instructions (reg mov + inc). Some of the cg.prologue counters are generated after RA which results in unallocated registers as seen in this log:

This entire problem is fixed by implementing https://github.com/eclipse/omr/issues/397 on x86 and Power. We had the exact same problem on Z and hence why we redesigned the codegen part of the debug counter infrastructure. @andrewcraik / @0dvictor FYI.

runtime/compiler/trj9/compile/J9Compilation.hpp
```diff
@@ -284,6 +284,15 @@ class OMR_EXTENSIBLE Compilation : public OMR::CompilationConnector
...
    bool supportsQuadOptimization();
 
+   // Debug Counter Relocation
+   typedef std::pair<const char *, const char *> DebugCounterNameEntry;
+   typedef TR::typed_allocator<DebugCounterNameEntry, TR::Allocator> DebugCounterNameMapAllocator;
+   typedef std::map<const char *, const char *, std::less<const char *>, DebugCounterNameMapAllocator> DebugCounterNameMap;
+   void mapKeyToCounterName(const char *key, const char *counterName);
+   const char *getCounterName(const char *key);
+   uint32_t calculateSizeOfCounterNames(uint32_t &numDebugCounterNames);
+   DebugCounterNameMap &getDebugCounterNameMap() { return _debugCounterNameMap; }
```
fjeremic: These new APIs should have Doxygen documentation, unless this is an extensible class and they've been already documented in OMR. Similarly for the new field added as well.

runtime/compiler/trj9/codegen/J9AheadOfTimeCompile.cpp
```diff
@@ -936,6 +936,41 @@ J9::AheadOfTimeCompile::dumpRelocationData()
...
+         case TR_DebugCounter:
+            cursor ++;
+            if (is64BitTarget)
+               {
+               cursor += 4;     // padding
+               ep1 = cursor;    // inlinedSiteIndex
+               ep2 = cursor+8;  // bcIndex
+               ep3 = cursor+16; // keyOfNameString
+               ep4 = cursor+24; // delta
+               ep5 = cursor+40; // staticDelta
```
fjeremic: Can we not use `offsetof(TR_RelocationRecordDebugCounterBinaryTemplate, _staticDelta)` here and in nearby places? This seems not very robust to changes.

dsouzai: > This seems not very robust to changes.
Yeah, the entire file follows this paradigm. At some point when I get time I want to clean up the AOT store-to-scc code (the load-from-scc code is already properly refactored). For now though, I'll do what you suggest with this bit of code.

dsouzai: Heh I'm not easily able to change the code to use `offsetof(TR_RelocationRecordDebugCounterBinaryTemplate, _staticDelta)`; I'd have to do something like `cursor+offsetof(TR_RelocationRecordDebugCounterBinaryTemplate, _staticDelta)-sizeof(TR_RelocationRecordBinaryTemplate)` which is uglier and lacks robustness in a diff way (since changing the inheritance graph would cause problems again). I'll leave it as is for now.

runtime/compiler/trj9/runtime/RelocationRuntime.cpp
```diff
@@ -657,6 +657,34 @@ TR_RelocationRuntime::relocateMethodMetaData(UDATA codeRelocationAmount, UDATA d
...
+         {
+         const char *key = (const char *)(*(uintptrj_t *)cursor);
+         cursor += sizeof(uintptrj_t);
+
+         uint32_t size = *(uint32_t *)cursor;
+         cursor += sizeof(uint32_t);
+
+         if (!_comp->getCounterName(key))
+            {
+            // Need to allocate persistent memory since the metadata can be reclaimed
```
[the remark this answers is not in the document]
dsouzai: > Is it reclaimed or just disclaimed?

The metadata that gets allocated in the data cache gets reclaimed, ie, freed back into the data cache for reuse of something else. 

> the entire shared classes cache should be mapped into the address space, so I would have guessed you could just use pointers to the names stored in the meta data.
>
> In that case, it would be better to store the references as offsets from the beginning of the shared classes cache, so then you can easily convert the offset in the relocation data into a pointer (to a spot in the meta data holding the name) by just adding the cache address (we do that in a bunch of other places e.g. class chains).

At AOT compile, when we store all the relocation records to the buffer, the metadata hasn't be created yet (I believe), so the relocation record can't know the offset at which the string will be stored. That's why I used the string ptr as both the key and the value during the AOT compile step.

However, what you're suggesting does work at AOT load time. since I can calculate the offset into the memory mapped shared cache from the metadata that gets memcpy'd into data cache allocated memory. That way I don't need to allocate the strings into persistent memory.

All that said, I just realized that there's a problem with the current implementation. `TR::DebugCounterGroup::findCounter` finds the counter from a hashtable that's hashed via a `const char *`. For counters that are created during a compile where the name is dynamically generated, my current implementation is fine. However, for counters who's names are known at compile time (ie stored in the BSO section of the dll), my implementation breaks down; this is because if the counter is enabled for multiple methods, the same string will get stored in multiple metadatas, and each of those will result in a new counter created at relo time (because they will have a different `const char *` ptr as the key in the counter hash). I don't think it's a massive functional problem, the accumulation will just not be truly accumulative and instead will result in the developer having to add up stuff manually which is annoying. 

I'm gonna have to think about how to get around this issue without dealing directly with the SCC; I'm reluctant to deal with the SCC directly because, as far as I could tell, the only way to store data to it was to hash it with a key, and generating a unique key involved (at least looking at how classes in the class chain are stored) taking the pointer and converting it to an ASCII string. If there's an easier way to deal with the SCC I'm more than happy to look into it.

dsouzai:
@mstoodle Made the requested changes. 

I also changed the code to store the debug counter name strings directly to the SCC instead of the per method metadata. Unfortunately, this only partially solves the problem I mentioned in https://github.com/eclipse/openj9/pull/1589#discussion_r182260682 . Now if there's a counter name that sits in the BSO section of the jit dll, instead of seeing that counter for each AOT'd method + all JIT'd methods, there will be one counter for all AOT loaded methods and one counter for all JIT'd methods. Actually, if in the AOT load run we do an AOT compile, then on the subsequent run, there will be three counters instead of 1. There really isn't a good way to solve this issue, except creating a mapping between the string in the BSO and a copy that can be persisted...

runtime/compiler/trj9/env/J9SharedCache.cpp
```diff
@@ -455,7 +455,8 @@ TR_J9SharedCache::rememberClass(J9Class *clazz, bool create)
       return NULL;
       }
 
-   char key[17]; // longest possible key length is way less than 16 digits
+   const uint32_t bufferLength = 17; // longest possible key length is way less than 16 digits
+   char key[bufferLength];
```
dsouzai: Hm should I change this back? I had modified this line when I thought I had to change the number of arguments to `createClassKey` but as it turns out, I didn't need to. I guess I forgot to revert this.

dsouzai:
@mstoodle Made all the requested changes.

dsouzai:
Verified my changes on P and Z.

dsouzai:
Signed-off-by Check failed because of `'` in my name :)

AdamBrousseau:
Jenkins signed off by check

dsouzai:
@mstoodle Does everything look ok to you now?

andrewcraik:
@mstoodle could you please reply to @dsouzai ? this is some great work which would be good to get in...

dsouzai:
I'll rebase and re-test this once the OMR changes get to openj9-omr

mstoodle:
@andrewcraik I had been paying more attention to the dependent omr changes, but since I've now merged those, will now focus more on this one...Thanks for the ping.

dsouzai:
@mstoodle rebased changes look good (from a Travis CI pov). Do you mind now taking a look and seeing if everything from the previous reviews looks good.

runtime/compiler/p/codegen/J9AheadOfTimeCompile.cpp
```diff
@@ -801,6 +802,35 @@ uint8_t *J9::Power::AheadOfTimeCompile::initializeAOTRelocationHeader(TR::Iterat
...
 
+      case TR_DebugCounter:
+         {
+         TR::DebugCounterBase *counter = (TR::DebugCounterBase *) relocation->getTargetAddress();
+         if (!counter || !counter->getReloData() || !counter->getName())
+            comp->failCompilation<TR::CompilationException>("Failed to generate debug counter relo data");
+
+         TR::DebugCounterReloData *counterReloData = counter->getReloData();
+
+         UDATA offset = fej9->sharedCache()->rememberDebugCounterName(counter->getName());
```
mstoodle: I would prefer this to be a JIT data type rather than UDATA, just for consistency. I know it doesn't architecturally matter any more now that we're in the same repo, but I think the consistency is still helpful until we get around to unifying all our types.

dsouzai:
@mstoodle made the requested changes

dsouzai:
@mstoodle Does everything look ok to you?

runtime/compiler/env/J9SharedCache.hpp
```diff
@@ -69,6 +69,9 @@ class TR_J9SharedCache : public TR_SharedCache
 
    UDATA *rememberClass(J9Class *clazz, bool create=true);
 
+   UDATA rememberDebugCounterName(const char *name);
```
mstoodle: it would have been more correct to use uintptrj_t here rather than UDATA and do the cast inside these functions, but I'll let it go as something easily cleaned up later if it the typing inconsistency becomes any kind of issue.

mstoodle:
jenkins test sanity

dsouzai:
> Duplicating the code 3 times is in character for these files, sadly, though taking advantage of a function implemented in the base class might have been better.

I think one of my goals is to try and clean up the AOT store side infrastructure; hopefully I manage to get to that this year :)

mstoodle:
jenkins test sanity windows

mstoodle:
jenkins test sanity win

dsouzai:
Windows build still failing with `NPT ERROR: Cannot open library` :(

dsouzai:
@mstoodle Is it worth trying a sanity build for windows again?

mstoodle:
ok, i'll try one more time :) . It would be nice to have a clean run, but if this doesn't fail I am pretty confident we can merge this PR. You're convinced by "pretty confident" right? ;)

jenkins test sanity win

mstoodle:
the "intermittent" problem only affects the JDK8 runs, I believe, but it's hard to believe your changes are dependent on JDK10 vs. JDK8...

dsouzai:
Heh,  yeah, considering that changes to the actual compiler were all done in OMR that's been fine.

As an aside, only 2 of the 3 win jobs kicked off; not sure why, but it looks like it won't be a clean PR either way :P

andrewcraik:
be ware of #2129 NPT Error failures on windows - they happen pretty much every time on every PR build at the moment.

mstoodle:
jenkins sanity test win32

mstoodle:
I don't believe the JDK8 win or win32 failures are due to this PR. Merging.

PR #1594, opened 2018-04-03
Check type_index of runtime visible type annotations

pdbain-ibm:
If a type annotation's type_index does not refer to a CONSTANT_UTF8_info
structure, don't mark it as being used.  When writing the ROM class, ensure
that constant pool entries marked as used by annotations are UTF8.

FYI @fengxue-IS 

### DIFF
--- a/runtime/bcutil/ClassFileOracle.cpp
+++ b/runtime/bcutil/ClassFileOracle.cpp
@@ -746,11 +746,28 @@ void
 ClassFileOracle::walkTypeAnnotations(U_16 annotationsCount, J9CfrTypeAnnotation *typeAnnotations) {
 	for (U_16 typeAnnotationIndex = 0; typeAnnotationIndex < annotationsCount; ++ typeAnnotationIndex) {
 		J9CfrAnnotation *annotation = &(typeAnnotations[typeAnnotationIndex].annotation);
-		markConstantAsUsedByAnnotation(annotation->typeIndex);
-		const U_16 elementValuePairsCount = annotation->numberOfElementValuePairs;
-		for (U_16 elementValuePairsIndex = 0; (elementValuePairsIndex < elementValuePairsCount) && (OK == _buildResult); ++elementValuePairsIndex) {
-			markConstantAsUsedByAnnotation(annotation->elementValuePairs[elementValuePairsIndex].elementNameIndex);
-			walkAnnotationElement(annotation->elementValuePairs[elementValuePairsIndex].value);
+		/* type_index in an annotation must refer to a CONSTANT_UTF8_info structure. */
+		if (getCPTag(annotation->typeIndex) == CFR_CONSTANT_Utf8) {
+			markConstantAsUsedByAnnotation(annotation->typeIndex);
+			const U_16 elementValuePairsCount = annotation->numberOfElementValuePairs;
+			for (U_16 elementValuePairsIndex = 0;
+					(elementValuePairsIndex < elementValuePairsCount) && (OK == _buildResult); ++elementValuePairsIndex) {
+				markConstantAsUsedByAnnotation(annotation->elementValuePairs[elementValuePairsIndex].elementNameIndex);
+				walkAnnotationElement(annotation->elementValuePairs[elementValuePairsIndex].value);
+			}
+		} else {
+			/*
+			 * UTF-8 entries and method entries use the same set of marking labels when
+			 * preparing to build the ROM class, but
+			 * the label value meanings differ depending on the type of the entry.
+			 * Thus we cannot mark a non-UTF-8 entry with a label used for UTF-8 as that label
+			 * value will be (mis)interpreted according to the type of the entry.
+			 *
+			 * In this case, force the typeIndex to a null value.
+			 * This will cause the parser to throw
+			 * an error if the the VM or application tries to retrieve the annotation.
+			 */
+			annotation->typeIndex = 0;
 		}
 	}
 }


pdbain-ibm:
Please see https://github.com/eclipse/openj9/pull/1579 for comments on the issue and previous solution.

pdbain-ibm:
@fengxue-IS @DanHeidinga @gacholio would you kindly review this?

pdbain-ibm:
@pshipton would you please launch PR builds?

runtime/bcutil/ROMClassWriter.cpp
```diff
@@ -180,7 +180,7 @@ class ROMClassWriter::AnnotationWriter :
 
 	void visitAnnotation(U_16 typeIndex, U_16 elementValuePairCount)
 	{
-		_cursor->writeBigEndianU16(_constantPoolMap->getROMClassCPIndexForReference(typeIndex), Cursor::GENERIC);
+		_cursor->writeBigEndianU16(_constantPoolMap->getROMClassCPIndexForAnnotationWithCheck(typeIndex), Cursor::GENERIC);
```
DanHeidinga: Doesn't this still return 0 as the cpIndex for annotations with invalid type_index? I thought that was why the previous PR was closed:
`We realized that 0 is a valid constant pool entry. Closing this while we develop another solution.`

pdbain-ibm: The prior solution set the classfile constant pool index to zero (a valid entry).

This change set leaves the classfile constant pool index as-is but does not mark it as used by an annotation.  When we create the ROM constant pool, we check the validity of the  classfile index and force the ROM constant pool index to zero (i.e. not a valid index).   @fengxue-IS please would you please verify my logic?

fengxue-IS: The index zero for classfile CP is the first entry read, where as the zero index in ROM CP is a reserved null entry (which should trigger a runtime exception when attempting to read from it), in the previous PR, we set the classfile cp to zero, so this line of code with attempt to get an value the may be initialized (ie. classfile CP zero entry is not referenced)

runtime/bcutil/ConstantPoolMap.hpp
```diff
@@ -210,6 +210,13 @@ class ConstantPoolMap
 
 	U_16 getROMClassCPIndexForReference(U_16 cfrCPIndex)  const { return getROMClassCPIndex(cfrCPIndex); }
 	U_16 getROMClassCPIndexForAnnotation(U_16 cfrCPIndex) const { return getROMClassCPIndex(cfrCPIndex); }
+	U_16 getROMClassCPIndexForAnnotationWithCheck(U_16 cfrCPIndex) const { 
+		if ((getCPTag(cfrCPIndex) == CFR_CONSTANT_Utf8) && isMarked(cfrCPIndex, ANNOTATION)) {
+			 return getROMClassCPIndex(cfrCPIndex); 
+		} else {
+			return 0;
+		}
+	}
```
gacholio: Multiple returns are discouraged, especially with code this simple.  How about initializing a return value to 0, and setting it on success?

runtime/bcutil/ROMClassWriter.cpp
DanHeidinga: The zero index in constantpool means null, even in a classfile.  See JVMS 4.1
> The `constant_pool` table is indexed from 1 to `constant_pool_count-1`

DanHeidinga:
This is basically the same as #1579 but less clear.  Please reopen #1579 and we'll proceed with that item.

DanHeidinga:
I suspect that's due to having closed the other pr and then creating a new one with the same branch name.

Here's a link to the patch for the original PR: https://patch-diff.githubusercontent.com/raw/eclipse/openj9/pull/1579.patch

runtime/bcutil/ROMClassWriter.cpp
pdbain-ibm: There is no defined mapping from classfile offset 0, as it is never used, so we felt it was safer to leave a valid classfile index and detect the error when we do the mapping.

runtime/bcutil/ConstantPoolMap.hpp
pdbain-ibm: Done.  Thanks.

pdbain-ibm:
> I suspect that's due to having closed the other pr and then creating a new one with the same branch name.
I created the new  pull request because I couldn't reopen the old one.

runtime/bcutil/ROMClassWriter.cpp
DanHeidinga: index 0 is used as a null value.  We're safe to use it here.

pdbain-ibm:
Reverted to the previous changeset.

runtime/bcutil/ROMClassWriter.cpp
pdbain-ibm: In that case, I reverted to the previous changeset.

DanHeidinga:
Peter and I had an at desk discussion about this PR and I now understand the cause better.  He'll update the description and add a comment to the code describing why we need to swizzle to cpIndex 0 rather than using the existing cpIndex.  Then this will be ready to merge

DanHeidinga:
See also the comments in some of the marking code for method refs / field refs which indicates the same problem with a TODO
ie:
```
ClassFileOracle::markMethodRefAsReferenced(U_16 cpIndex)
{
	Trc_BCU_Assert_True((CFR_CONSTANT_Methodref == _classFile->constantPool[cpIndex].tag) || (CFR_CONSTANT_InterfaceMethodref == _classFile->constantPool[cpIndex].tag));

	// TODO markConstantAsReferenced(cpIndex); -- this breaks the iteration strategy in ROMClassBuilder
```

DanHeidinga:

Issue #1192, opened 2018-02-15
Remove -Xdiagnosticscollector option


pshipton:
-Xdiagnosticscollector just prints a warning. 

`JVMJ9VM138W The -Xdiagnosticscollector option is not supported by this JVM.`

As it is obsolete it should be removed.

PR #1595, opened 2018-04-03
WIP: Remove -Xdiagnosticscollector

stevewallin:
This patch removes the -Xdiagnosticscollector option as described in issue #1192

Fixes: #1192

Signed-off-by: Steve Wallin <stevewallin@gmail.com>


DanHeidinga:
For the NLS files, we try to keep the keys stable which means we can't delete old messages.  See the guidance at the top of the NLS files for how to handle this:
```
# New messages MUST be added at the end of this file.
```

stevewallin:
@DanHeidinga will fix the keys and conflict on Friday - had this in my cache for a while and was cleaning up :)

DanHeidinga:
The copyright dates should also be updated to 2018 as well so the copyright check will pass

PR #1600, opened 2018-04-03
Improve zero init sequence while iterating zero slots bit vector

yanluo7:


yanluo7:
Performance implication:

For small objects allocations that require zero initialization, we iterate a bit vector to indicate which 4-byte slots within the object need to be zero'd. Currently x86 uses 4-byte store instructions for each iterated slot in the bit vector. This change optimizes that by using a look-ahead scheme to greedily group nearby slots and use a wider 4, 8, or 16-byte store to save instructions. Running a server benchmark, we measured close to 1% in throughput improvement.

yanluo7:
Ottawa promotion regression build (xa xi): 
build_id=383081

fjeremic:
Removed a comment referencing internal builds which is not relevant to the open source OpenJ9 project.

andrewcraik:
Jenkins test sanity

PR #1601, opened 2018-04-03
WIP: Add jProfilingRecompilation test in loop

r30shah:
When jProfiling in compilation is enabled, add a test after aynchcheck node
in loop to calculate the frequency and trip up recompilation of the method
if reached certain threshold.

### DIFF
--- a/buildenv/jenkins/common/build
+++ b/buildenv/jenkins/common/build
@@ -219,7 +219,7 @@ def build() {
     stage('Compile') {
         timestamps {
             sh "bash ./configure --with-freemarker-jar='$FREEMARKER' --with-boot-jdk='$BOOT_JDK' $EXTRA_CONFIGURE_OPTIONS"
-            sh "make all"
+            sh "make $EXTRA_MAKE_OPTIONS all"
         }
     }
     stage('Java Version') {
--- a/buildenv/jenkins/common/variables-functions
+++ b/buildenv/jenkins/common/variables-functions
@@ -230,6 +230,7 @@ def set_build_variables() {
     BOOT_JDK = get_value(VARIABLES."${SPEC}".boot_jdk, SDK_VERSION)
     FREEMARKER = VARIABLES."${SPEC}".freemarker
     EXTRA_CONFIGURE_OPTIONS = get_value(VARIABLES."${SPEC}".extra_configure_options, SDK_VERSION)
+    EXTRA_MAKE_OPTIONS = get_value(VARIABLES."${SPEC}".extra_make_options, SDK_VERSION)
     OPENJDK_REFERENCE_REPO = VARIABLES."${SPEC}".openjdk_reference_repo
     set_release()
     set_jdk_folder()
--- a/buildenv/jenkins/jobs/builds/Build-JDK9-linux_x86-64_cmprssptrs_cmake
+++ b/buildenv/jenkins/jobs/builds/Build-JDK9-linux_x86-64_cmprssptrs_cmake
@@ -1,0 +1,40 @@
+/*******************************************************************************
+ * Copyright (c) 2018, 2018 IBM Corp. and others
+ *
+ * This program and the accompanying materials are made available under
+ * the terms of the Eclipse Public License 2.0 which accompanies this
+ * distribution and is available at https://www.eclipse.org/legal/epl-2.0/
+ * or the Apache License, Version 2.0 which accompanies this distribution and
+ * is available at https://www.apache.org/licenses/LICENSE-2.0.
+ *
+ * This Source Code may also be made available under the following
+ * Secondary Licenses when the conditions for such availability set
+ * forth in the Eclipse Public License, v. 2.0 are satisfied: GNU
+ * General Public License, version 2 with the GNU Classpath
+ * Exception [1] and GNU General Public License, version 2 with the
+ * OpenJDK Assembly Exception [2].
+ *
+ * [1] https://www.gnu.org/software/classpath/license.html
+ * [2] http://openjdk.java.net/legal/assembly-exception.html
+ *
+ * SPDX-License-Identifier: EPL-2.0 OR Apache-2.0 OR GPL-2.0 WITH Classpath-exception-2.0 OR LicenseRef-GPL-2.0 WITH Assembly-exception
+ *******************************************************************************/
+
+SDK_VERSION = '9'
+SPEC = 'linux_x86-64_cmprssptrs_cmake'
+
+node('master') {
+    timestamps {
+        retry(5) {
+            checkout scm
+        }
+        variableFile = load 'buildenv/jenkins/common/variables-functions'
+        variableFile.set_job_variables('build')
+        buildfile = load 'buildenv/jenkins/common/build'
+        cleanWs()
+    }
+}
+
+node("${NODE}") {
+    buildfile.build_all()
+}
--- a/buildenv/jenkins/jobs/pipelines/Pipeline-OMR-Acceptance
+++ b/buildenv/jenkins/jobs/pipelines/Pipeline-OMR-Acceptance
@@ -80,6 +80,16 @@ timeout(time: 8, unit: 'HOURS') {
                  }
             }
 
+            // Note: we are doing this as a special case rather than include it in SPECS,
+            // because at the moment cmake builds are only supported on java 9.
+            // Also we dont want to bother running tests at the moment.
+            def CMAKE_SDK_VERSION = '9'
+            def SHAS = [:]
+            SHAS['OPENJ9'] = ALL_SHAS['OPENJ9']
+            SHAS['OMR'] = ALL_SHAS['OMR']
+            SHAS['OPENJDK'] = buildFile.get_sha(OPENJDK_REPOS[CMAKE_SDK_VERSION].repo, OPENJDK_REPOS[CMAKE_SDK_VERSION].branch)
+            JOBS["Build-TestJDK${CMAKE_SDK_VERSION}-linux_x86-64_cmprssptrs_cmake"] = { buildFile.workflow(CMAKE_SDK_VERSION as int, linux_x86-64_cmprssptrs_cmake, SHAS, OPENJDK_REPOS[CMAKE_SDK_VERSION].repo, OPENJDK_REPOS[CMAKE_SDK_VERSION].branch, OPENJ9_REPO, OPENJ9_BRANCH, OMR_REPO, OMR_BRANCH, 'none') }
+
             cleanWs()
         }
     }
--- a/buildenv/jenkins/jobs/pipelines/Pipeline-OMR-Acceptance
+++ b/buildenv/jenkins/jobs/pipelines/Pipeline-OMR-Acceptance
@@ -88,7 +88,7 @@ timeout(time: 8, unit: 'HOURS') {
             SHAS['OPENJ9'] = ALL_SHAS['OPENJ9']
             SHAS['OMR'] = ALL_SHAS['OMR']
             SHAS['OPENJDK'] = buildFile.get_sha(OPENJDK_REPOS[CMAKE_SDK_VERSION].repo, OPENJDK_REPOS[CMAKE_SDK_VERSION].branch)
-            JOBS["Build-TestJDK${CMAKE_SDK_VERSION}-linux_x86-64_cmprssptrs_cmake"] = { buildFile.workflow(CMAKE_SDK_VERSION as int, linux_x86-64_cmprssptrs_cmake, SHAS, OPENJDK_REPOS[CMAKE_SDK_VERSION].repo, OPENJDK_REPOS[CMAKE_SDK_VERSION].branch, OPENJ9_REPO, OPENJ9_BRANCH, OMR_REPO, OMR_BRANCH, 'none') }
+            JOBS["Build-TestJDK${CMAKE_SDK_VERSION}-linux_x86-64_cmprssptrs_cmake"] = { buildFile.workflow(CMAKE_SDK_VERSION as int, 'linux_x86-64_cmprssptrs_cmake', SHAS, OPENJDK_REPOS[CMAKE_SDK_VERSION].repo, OPENJDK_REPOS[CMAKE_SDK_VERSION].branch, OPENJ9_REPO, OPENJ9_BRANCH, OMR_REPO, OMR_BRANCH, 'none') }
 
             cleanWs()
         }
--- a/buildenv/jenkins/variables/defaults.yml
+++ b/buildenv/jenkins/variables/defaults.yml
@@ -158,6 +158,37 @@ linux_x86-64_cmprssptrs:
       9: 'hw.arch.x86 && sw.os.ubuntu.16'
       10: 'hw.arch.x86 && sw.os.ubuntu.16'
 #========================================#
+# Linux x86 64bits Compressed Pointers /w CMake
+#========================================#
+linux_x86-64_cmprssptrs_cmake:
+  boot_jdk:
+    8: '/usr/lib/jvm/java-7-openjdk-amd64'
+    9: '/usr/lib/jvm/adoptojdk-java-80'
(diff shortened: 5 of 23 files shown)


r30shah:
@andrewcraik  This is the prototype for adding a recompilation test in the loop. I am running this optimization in late near jProfilingValue (just before GRA) with using bytecode info of `asynccheck` node to get the original block frequency counter info. 
Marking this as `WIP` as need your input on the design as well as we need to tune the recompilation threshold.

runtime/compiler/trj9/optimizer/JProfilingRecompLoopTest.cpp
```diff
@@ -0,0 +1,214 @@
...
+
+   while (cursor)
+      {
+      if (cursor->getNode()->getOpCodeValue() == TR::asynccheck && cursor->getEnclosingBlock()->getStructureOf()->getContainingLoop() != NULL)
+         {
+         int32_t loopDepth = 1;
+         TR::Block *block = cursor->getEnclosingBlock();
+         TR_RegionStructure *region = block->getParentStructureIfExists(comp()->getFlowGraph());
+         if (region)
+            region->calculateFrequencyOfExecution(&loopDepth);
```
andrewcraik: This doesn't look right and we shouldn't be backsolving a depth from the frequency this thing computes. You should walk the region structures out to NULL counting the loops.

runtime/compiler/trj9/optimizer/JProfilingRecompLoopTest.cpp
```diff
@@ -0,0 +1,214 @@
...
+         traceMsg(comp(), "JProfiling has not been enabled, skip JProfilingRecompLoopTest\n");
+      return 0;
+      }
+   RecompilationTestLocations testLocations(RecompilationTestLocationAllocator(comp()->trMemory()->currentStackRegion()));
+   TR::TreeTop *cursor = comp()->getStartTree();
+   TR::CFG *cfg = comp()->getFlowGraph();
+
+   while (cursor)
+      {
+      if (cursor->getNode()->getOpCodeValue() == TR::asynccheck && cursor->getEnclosingBlock()->getStructureOf()->getContainingLoop() != NULL)
```
andrewcraik: getEnclosing block is going to walk the trees - you should track the block as you walk the trees to save having to find the block everytime. The asynccheck my not be in the loop directly - you could have another subgraph node so every asynccheck needs to compute its loop nesting depth.

runtime/compiler/trj9/optimizer/JProfilingRecompLoopTest.cpp
```diff
@@ -0,0 +1,214 @@
...
+      cfg->addEdge(TR::CFGEdge::createEdge(callRecompileBlock, newSplittedBlock, comp->trMemory()));
+      cfg->addEdge(TR::CFGEdge::createEdge(originalBlock, callRecompileBlock, comp->trMemory()));
+      if (trace())
+         traceMsg(comp,"\t\t Newly created recompilation Test : n%dn in block_%d\n/\t\tRecompilation Call in block_%d\n",
+            callNode->getGlobalIndex(), originalBlock->getNumber(), callRecompileBlock->getNumber());
+      }
+   }
+
+TR::Node *
+TR_JProfilingRecompLoopTest::generateBlockFrequecnyCalculationNode(TR::Compilation *comp, TR::Node *node, TR_BlockFrequencyInfo *bfi, int32_t blockNumber)
```
andrewcraik: There is code duplication here between this opt and the where we would insert the recomp tree at the start of the method - can you please make sure to not duplicate this code? We need to common the code out and parameterize the tree generation.

runtime/compiler/trj9/optimizer/JProfilingRecompLoopTest.cpp
```diff
@@ -0,0 +1,214 @@
...
+   {
+   TR_BitVector **componentCounters = bfi->getCounterDerivativeInfo();
+   TR::Node *root = NULL;
+   if (blockNumber > -1 && (componentCounters[blockNumber * 2]
+      && (((uintptr_t)componentCounters[blockNumber * 2]) & 0x1 == 1)
```
andrewcraik: This internal knowledge of how the counters are stored should be hidden in the profiling infrastructure or in the persistent data structures - we shouldn't be hardcoding the storage format here in case it changes later.

runtime/compiler/trj9/optimizer/JProfilingRecompLoopTest.cpp
```diff
@@ -0,0 +1,214 @@
...
+
+void 
+TR_JProfilingRecompLoopTest::addRecompilationTests(TR::Compilation *comp, RecompilationTestLocations &testLocations)
+   {
+   TR_PersistentProfileInfo *profileInfo = comp->getRecompilationInfo()->findOrCreateProfileInfo();
+   TR_BlockFrequencyInfo *bfi = TR_BlockFrequencyInfo::get(profileInfo);
+   TR::CFG *cfg = comp->getFlowGraph();
+   cfg->setStructure(NULL);
+
+   for (auto testLocation = testLocations.begin(); testLocation != testLocations.end(); ++testLocation)
```
andrewcraik: you should should not run end() on every iteration of the loop. A more efficient style is
for (auto testLocItr = testLocations.begin(), testLocEnd = testLocations.end(); testLocItr != testLocEnd; ++testLocItr)

runtime/compiler/trj9/optimizer/JProfilingRecompLoopTest.cpp
```diff
@@ -0,0 +1,214 @@
...
+      {
+      TR::TreeTop *asyncCheckTreeTop = testLocation->first;
+      TR::Node *node = asyncCheckTreeTop->getNode();
+      int32_t depth = testLocation->second;
+      if (trace())
+         traceMsg(comp, "block_%d, n%dn, depth = %d\n",asyncCheckTreeTop->getEnclosingBlock()->getNumber(), asyncCheckTreeTop->getNode()->getGlobalIndex(), depth);
+      TR_ByteCodeInfo bci = asyncCheckTreeTop->getNode()->getByteCodeInfo();
+      performTransformation(comp, "%s Add recompilation test after asyncCheck node n%dn\n", optDetailString(), node->getGlobalIndex());
```
andrewcraik: this needs to be in an if - performTransformation returns true if you should do the transformation and false otherwise. If it is false you shouldn't perform the transformation.

runtime/compiler/trj9/optimizer/JProfilingRecompLoopTest.cpp
```diff
@@ -0,0 +1,214 @@
...
+      TR::Node *root = generateBlockFrequecnyCalculationNode(comp, node, bfi, bfi->getOriginalBlockNumberToGetLiveFrequency(bci,comp));
+      
+      TR::Block *originalBlock = asyncCheckTreeTop->getEnclosingBlock();
+      TR::Block *newSplittedBlock = originalBlock->split(asyncCheckTreeTop->getNextTreeTop(), cfg, true, true);
```
andrewcraik: this variable needs a better name

runtime/compiler/trj9/optimizer/JProfilingRecompLoopTest.cpp
```diff
@@ -0,0 +1,214 @@
...
+      TR::TreeTop *asyncCheckTreeTop = testLocation->first;
+      TR::Node *node = asyncCheckTreeTop->getNode();
+      int32_t depth = testLocation->second;
+      if (trace())
+         traceMsg(comp, "block_%d, n%dn, depth = %d\n",asyncCheckTreeTop->getEnclosingBlock()->getNumber(), asyncCheckTreeTop->getNode()->getGlobalIndex(), depth);
+      TR_ByteCodeInfo bci = asyncCheckTreeTop->getNode()->getByteCodeInfo();
+      performTransformation(comp, "%s Add recompilation test after asyncCheck node n%dn\n", optDetailString(), node->getGlobalIndex());
+      TR::Node *root = generateBlockFrequecnyCalculationNode(comp, node, bfi, bfi->getOriginalBlockNumberToGetLiveFrequency(bci,comp));
+      
+      TR::Block *originalBlock = asyncCheckTreeTop->getEnclosingBlock();
```
andrewcraik: An ascii block diagram should be shown above to show what is happening - take a look at the JProfiling lowering code, VirtualGuardHeadMerger, VirtualGuardTailSplitter etc. The diagram should have the blocks labelled with the local variable names for the pointers pointing to the blocks.

runtime/compiler/trj9/optimizer/JProfilingRecompLoopTest.cpp
```diff
@@ -0,0 +1,214 @@
...
+      TR::Node *callNode = TR::Node::createWithSymRef(node, TR::icall, 2, comp->getSymRefTab()->findOrCreateRuntimeHelper(TR_jitRetranslateCallerWithPrep, false, false, true));
+      callNode->setAndIncChild(0, TR::Node::createWithSymRef(node, TR::loadaddr, 0, comp->getSymRefTab()->findOrCreateStartPCSymbolRef()));
+      callNode->setAndIncChild(1, TR::Node::createWithSymRef(node, TR::loadaddr, 0, comp->getSymRefTab()->findOrCreateCompiledMethodSymbolRef()));
+      TR::TreeTop *callTree = TR::TreeTop::create(comp, TR::Node::create(node, TR::treetop, 1, callNode));
+      callTree->getNode()->setIsProfilingCode();
+      callRecompileBlock->append(callTree);
+      cfg->addNode(callRecompileBlock);
+      
+      TR::Node *loadBaseThreshold = TR::Node::createWithSymRef(node, TR::iload, 0, comp->getSymRefTab()->createKnownStaticDataSymbolRef(&recompileThreshold, TR::Int32));
+      TR::Node *adjustThreshold = TR::Node::create(node, TR::ishl, 2, loadBaseThreshold, TR::Node::iconst(node, (depth-1)*3));
```
andrewcraik: Since the depth is a compile constant atm this could just be a constant no? we only need to do the shift at runtime if it is going to be dynamic?

runtime/compiler/trj9/optimizer/JProfilingRecompLoopTest.cpp
```diff
@@ -0,0 +1,214 @@
...
+   TR_BitVector **componentCounters = bfi->getCounterDerivativeInfo();
+   TR::Node *root = NULL;
+   if (blockNumber > -1 && (componentCounters[blockNumber * 2]
+      && (((uintptr_t)componentCounters[blockNumber * 2]) & 0x1 == 1)
+      || !componentCounters[blockNumber * 2]->isEmpty()))
+      {
+      TR::Node *addRoot = NULL;
+      if (((uintptr_t)componentCounters[blockNumber * 2]) & 0x1 == 1)
+         {
+         TR::SymbolReference *symRef = comp->getSymRefTab()->createKnownStaticDataSymbolRef(bfi->getFrequencyForBlock(((uintptr_t)componentCounters[blockNumber * 2]) >> 1), TR::Int32);
```
andrewcraik: the logic for the static address should be commoned out since several people need that.

runtime/compiler/trj9/runtime/J9Profiler.cpp
```diff
@@ -1932,6 +1932,25 @@ TR_BlockFrequencyInfo::getRawCount(TR::ResolvedMethodSymbol *resolvedMethod, TR_
...
+   {
+   int32_t callerIndex = bci.getCallerIndex();
+   TR::ResolvedMethodSymbol *resolvedMethod = callerIndex < 0 ? comp->getMethodSymbol() : comp->getInlinedResolvedMethodSymbol(callerIndex);
+   int32_t byteCodeToSearch = resolvedMethod->getProfilingByteCodeIndex(bci.getByteCodeIndex());
+   TR_ByteCodeInfo searchBCI = bci;
+   searchBCI.setByteCodeIndex(byteCodeToSearch); 
+   for (auto i=0; i < _numBlocks; ++i)
+      {
+      if (_callSiteInfo->hasSameBytecodeInfo(_blocks[i], searchBCI, comp)
+         || (_blocks[i].getCallerIndex() == searchBCI.getCallerIndex() && _blocks[i].getByteCodeIndex() == searchBCI.getByteCodeIndex()))
```
andrewcraik: so this is a general API so the bci may not match a block start so this needs consider the range of offsets a given block covers. I am still checking to make sure the callerindex is safe across recompilation uses. The name should not include the word frequency since profiling related code frequency refers to the normalized frequency between 6 and 10000 - this should use the term raw count or similar to avoid confusion.

This needs some Javadoc of some kind as well.

runtime/compiler/trj9/optimizer/JProfilingRecompLoopTest.cpp
```diff
@@ -0,0 +1,214 @@
...
+#include "il/Node_inlines.hpp"
+#include "infra/Checklist.hpp"             // for TR::NodeChecklist
+#include "ras/DebugCounter.hpp"
+#include "control/Recompilation.hpp"              // for TR_Recompilation, etc
+#include "control/RecompilationInfo.hpp"              // for TR_Recompilation, etc
+#include "codegen/CodeGenerator.hpp"
+#include "optimizer/TransformUtil.hpp"            // for TransformUtil
+
+
+int32_t TR_JProfilingRecompLoopTest::recompileThreshold = 4096;
```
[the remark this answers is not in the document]
fjeremic: And a comment describing what this value actually represents.

runtime/compiler/trj9/optimizer/JProfilingRecompLoopTest.cpp
```diff
@@ -0,0 +1,205 @@
...
+   while (cursor)
+      {
+      if (cursor->getNode()->getOpCodeValue() == TR::BBStart)
+         {
+         currentBlock = cursor->getNode()->getBlock();
+         }
+      else if (cursor->getNode()->getOpCodeValue() == TR::asynccheck)
+         {
+         TR_ASSERT_FATAL(currentBlock!=NULL,"We should have encountered BBStart before and should have the enclosing block");
+         if (currentBlock->getStructureOf()->getContainingLoop() != NULL)
```
r30shah: @andrewcraik  Regarding to this comment https://github.com/eclipse/openj9/pull/1601#discussion_r178960155
I was checking the function which is called, [getContainingLoop()](https://github.com/eclipse/omr/blob/1c979dbe7b3251bf0cad1f5dd375083ae862cefb/compiler/optimizer/Structure.cpp#L106), Looks like it walks the parent region structure till encountering `NULL` to get the natural loop if found otherwise returns NULL which is what is checked here.

runtime/compiler/trj9/optimizer/JProfilingRecompLoopTest.cpp
```diff
@@ -52,6 +52,21 @@ int32_t TR_JProfilingRecompLoopTest::getLoopNestingDepth(TR::Compilation *comp,
    return nestingDepth;
    }
 
+/**
+ * A utility function that iterates through the TR::list of byte code info and checks if passed bytecode info exists in list
+ * by checking bytecode index and caller index of the bytecode info
+ */ 
+bool
+TR_JProfilingRecompLoopTest::isByteCodeInfoInCurrentTestLocationList(TR_ByteCodeInfo &bci, TR::list<TR_ByteCodeInfo, TR::Region&> addedLocationBCIList)
```
andrewcraik: you should pass that list by reference rather than value to save copying the list.

fjeremic:
@dnakamura did you mean to add commits to this PR?

dnakamura:
I did not add them. No clue how they got there. Bad rebase?

r30shah:
@dnakamura @fjeremic Seems like a bad rebase. I rebased this yesterday and pushed it. Let me clean this up.

r30shah:
Closing this as we have merged these changes with different PR.

PR #1602, opened 2018-04-03
Run explicit new initialization pass after local VP to recognize more field initialization

yanluo7:

### DIFF
--- a/runtime/compiler/trj9/optimizer/J9Optimizer.cpp
+++ b/runtime/compiler/trj9/optimizer/J9Optimizer.cpp
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2000, 2017 IBM Corp. and others
+ * Copyright (c) 2000, 2018 IBM Corp. and others
  *
  * This program and the accompanying materials are made available under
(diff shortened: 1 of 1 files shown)


yanluo7:
- During the early stage of the optimization, after object allocation the newly allocated object is typically temporarily held in an auto; this means object field stores in the constructor are represented as store into a temp instead of the new object directly, this prevents Explicit New Initialization to recognize the indirect store as a valid field initialization. Later on in the optimization passes, local VP will remove the temp and replace it with a direct reference to "new" node. So we add a pass of explicit new initialization after local VP to allow these field stores to be recognized.

yanluo7:
Performance implication:

This change allows Explicit New Initialization to recognize more object fields to be marked as already initialized in the constructor, hence reducing the amount of uninitialized fields that need to be initialized. This change, along with another change in VP that remove redundant null checks for certain objects, showed about 0.3-0.4% throughput improvement for a server benchmark.

andrewcraik:
Jenkins test sanity

andrewcraik:
Jenkins test sanity

andrewcraik:
Jenkins test sanity

PR #1604, opened 2018-04-03
DDR gccheck Reporter messages mismatch

dmitripivkine:
There were number of messages removed from the table mistakenly. This
creates a mismatch between message code as index with table content. I
prefer do not renumber messages itself to keep them the same as an older
versions. Instead I added a number of dummy messages to the table to
restore proper alignment.


dmitripivkine:
Jenkins test sanity pLinux

Issue #1612, opened 2018-04-04
Refactor code in ProcessorDetection.cpp and VMJ9.h


harryyu1994:
During the development of issue 1105, I found that there are a lot of platform specific APIs and some potentially redundant code in the top level TR_J9VMBase class that could be refactored and simplified. 

Example: these are x86 platform specific methods in the top level API, that could be easily moved to their x86 platform specific J9CPU class.
```
const char *
TR_J9VMBase::getX86ProcessorVendorId()
   {
   TR_ASSERT(x86VendorIDInitialized, "X86 Vendor ID has not yet been initialized");
   return x86VendorID;
   }

uint32_t
TR_J9VMBase::getX86ProcessorSignature()
   {
   TR_ASSERT(_jitConfig, "jitConfig is not defined!");
   TR_ASSERT(_jitConfig->javaVM, "jitConfig->javaVM is not defined!");
   return queryX86TargetCPUID(_jitConfig->javaVM)->_processorSignature;
   }

uint32_t
TR_J9VMBase::getX86ProcessorFeatureFlags()
   {
   TR_ASSERT(_jitConfig, "jitConfig is not defined!");
   TR_ASSERT(_jitConfig->javaVM, "jitConfig->javaVM is not defined!");
   return queryX86TargetCPUID(_jitConfig->javaVM)->_featureFlags;
   }

bool
TR_J9VMBase::getX86SupportsHLE()
   {
   uint32_t flags8 = getX86ProcessorFeatureFlags8();
   if ((flags8 & TR_HLE) != 0x00000000)
      return true;
   else return false;
   }

```

PR #1613, opened 2018-04-04
Clean up x86 platform specific code in TR_J9VMBase

harryyu1994:
- Move the x86 platform specific code out of the TR_J9VMBase top level api and to x/J9CPU class
- Remove redundant code in J9CPU and TR_J9VMBase
- Rewrite and simplify some of the APIs inside x/J9CPU
- Depends on https://github.com/eclipse/omr/pull/2425

Issue: #1612 

### DIFF
--- a/runtime/compiler/env/J9CPU.hpp
+++ b/runtime/compiler/env/J9CPU.hpp
@@ -34,6 +34,7 @@ namespace J9 { typedef CPU CPUConnector; }
 
 #include "env/OMRCPU.hpp"
 #include "j9port.h"
+#include "infra/Assert.hpp"                         // for TR_ASSERT
 
 namespace J9
 {
@@ -45,6 +46,9 @@ protected:
 public:
 
    J9ProcessorDesc *TO_PORTLIB_getJ9ProcessorDesc();
+
+   const char *getProcessorVendorId() { TR_ASSERT(false, "Vendor ID not defined for this platform!"); return NULL; }
+   uint32_t getProcessorSignature() { TR_ASSERT(false, "Processor Signature not defined for this platform!"); return 0; }
    };
 }
 
--- a/runtime/compiler/x/env/J9CPU.cpp
+++ b/runtime/compiler/x/env/J9CPU.cpp
@@ -24,6 +24,12 @@
 #include "env/CompilerEnv.hpp"
 #include "env/CPU.hpp"
 #include "env/VMJ9.h"
+#include "x/runtime/X86Runtime.hpp"
+#include "env/JitConfig.hpp"
+
+// This is a workaround to avoid J9_PROJECT_SPECIFIC macros in x/env/OMRCPU.cpp
+// Without this definition, we get an undefined symbol of JITConfig::instance() at runtime
+TR::JitConfig * TR::JitConfig::instance() { return NULL; }
 
 namespace J9
 {
@@ -31,64 +37,45 @@ namespace J9
 namespace X86
 {
 
+TR_X86CPUIDBuffer *
+CPU::queryX86TargetCPUID()
+   {
+   static TR_X86CPUIDBuffer buf = { {'U','n','k','n','o','w','n','B','r','a','n','d'} };
+   jitGetCPUID(&buf);
+   return &buf;
+   }
+
 const char *
-CPU::getX86ProcessorVendorId(TR::Compilation *comp)
+CPU::getProcessorVendorId()
    {
-   return comp->fej9()->getX86ProcessorVendorId();
+   return self()->getX86ProcessorVendorId();
    }
 
 uint32_t
-CPU::getX86ProcessorSignature(TR::Compilation *comp)
+CPU::getProcessorSignature()
    {
-   return comp->fej9()->getX86ProcessorSignature();
-   }
-
-uint32_t
-CPU::getX86ProcessorFeatureFlags(TR::Compilation *comp)
-   {
-   return comp->fej9()->getX86ProcessorFeatureFlags();
-   }
-
-uint32_t
-CPU::getX86ProcessorFeatureFlags2(TR::Compilation *comp)
-   {
-   return comp->fej9()->getX86ProcessorFeatureFlags2();
-   }
-
-uint32_t
-CPU::getX86ProcessorFeatureFlags8(TR::Compilation *comp)
-   {
-   return comp->fej9()->getX86ProcessorFeatureFlags8();
+   return self()->getX86ProcessorSignature();
    }
 
 bool
-CPU::testOSForSSESupport(TR::Compilation *comp)
+CPU::testOSForSSESupport()
    {
-   return comp->fej9()->testOSForSSESupport();
+   return true; // VM guarantees SSE/SSE2 are available
    }
 
-void
-CPU::setX86ProcessorFeatureFlags(uint32_t flags)
+bool
+CPU::hasPopulationCountInstruction()
    {
-   _featureFlags = flags;
-   }
-
-void
-CPU::setX86ProcessorFeatureFlags2(uint32_t flags2)
-   {
-   _featureFlags2 = flags2;
-   }
-
-void
-CPU::setX86ProcessorFeatureFlags8(uint32_t flags8)
-   {
-   _featureFlags8 = flags8;
+   if ((self()->getX86ProcessorFeatureFlags2() & TR_POPCNT) != 0x00000000)
+      return true;
+   else
+      return false;
    }
 
 TR_ProcessorFeatureFlags
 CPU::getProcessorFeatureFlags()
    {
-   TR_ProcessorFeatureFlags processorFeatureFlags = { {_featureFlags, _featureFlags2, _featureFlags} };
+   TR_ProcessorFeatureFlags processorFeatureFlags = { {self()->getX86ProcessorFeatureFlags(), self()->getX86ProcessorFeatureFlags2(), self()->getX86ProcessorFeatureFlags8()} };
    return processorFeatureFlags;
    }
 
--- a/runtime/compiler/x/env/J9CPU.hpp
+++ b/runtime/compiler/x/env/J9CPU.hpp
@@ -35,6 +35,7 @@ namespace J9 { typedef J9::X86::CPU CPUConnector; }
 #endif
 
 #include "compiler/env/J9CPU.hpp"
+#include "env/ProcessorInfo.hpp"
 
 namespace TR { class Compilation; }
 
@@ -62,33 +63,21 @@ class CPU : public J9::CPU
 protected:
 
    CPU() :
-         J9::CPU(),
-         _featureFlags(0),
-         _featureFlags2(0),
-         _featureFlags8(0)
-      {}
+         J9::CPU()
+      {
+      }
 
 public:
 
-   const char *getX86ProcessorVendorId(TR::Compilation *comp);
-   uint32_t getX86ProcessorSignature(TR::Compilation *comp);
-   uint32_t getX86ProcessorFeatureFlags(TR::Compilation *comp);
-   uint32_t getX86ProcessorFeatureFlags2(TR::Compilation *comp);
-   uint32_t getX86ProcessorFeatureFlags8(TR::Compilation *comp);
+   TR_X86CPUIDBuffer *queryX86TargetCPUID();
+   const char * getProcessorVendorId();
+   uint32_t getProcessorSignature();
+
+   bool testOSForSSESupport();
+   bool hasPopulationCountInstruction();
 
    TR_ProcessorFeatureFlags getProcessorFeatureFlags();
    bool isCompatible(TR_Processor processorSignature, TR_ProcessorFeatureFlags processorFeatureFlags);
-
-   bool testOSForSSESupport(TR::Compilation *comp);
-
-   void setX86ProcessorFeatureFlags(uint32_t flags);
-   void setX86ProcessorFeatureFlags2(uint32_t flags2);
-   void setX86ProcessorFeatureFlags8(uint32_t flags8);
-
-private:
-   uint32_t _featureFlags;
-   uint32_t _featureFlags2;
-   uint32_t _featureFlags8;
 
    };
 
--- a/runtime/compiler/control/rossa.cpp
+++ b/runtime/compiler/control/rossa.cpp
@@ -1098,10 +1098,6 @@ onLoadInternal(
    if (!TR::CompilationInfo::createCompilationInfo(jitConfig))
       return -1;
 
-#if defined(TR_TARGET_X86)
-   TR_J9VM::initializeX86ProcessorInfo(jitConfig);
-#endif
-
    if (!TR_J9VMBase::createGlobalFrontEnd(jitConfig, TR::CompilationInfo::get()))
       return -1;
 
--- a/runtime/compiler/env/ProcessorDetection.cpp
+++ b/runtime/compiler/env/ProcessorDetection.cpp
@@ -574,108 +574,6 @@ TR_J9VMBase::getPPCSupportsVSXRegisters()
 
 // -----------------------------------------------------------------------------
 
-extern TR_X86CPUIDBuffer *queryX86TargetCPUID(void * javaVM);
-
-char TR_J9VMBase::x86VendorID[] = {'U','n','k','n','o','w','n','B','r','a','n','d','\0'};
-bool TR_J9VMBase::x86VendorIDInitialized = false;
-
-void
-TR_J9VMBase::initializeX86ProcessorInfo(J9JITConfig *jitConfig)
-   {
-   TR_ASSERT(jitConfig, "jitConfig is not defined!");
-   TR_ASSERT(jitConfig->javaVM, "jitConfig->javaVM is not defined!");
-
-   // Terminate the vendor ID with NULL before returning.
-   //
-   strncpy(x86VendorID, queryX86TargetCPUID(jitConfig->javaVM)->_vendorId, 12);
-   x86VendorID[12] = '\0';
-   x86VendorIDInitialized = true;
-
-   // Initialize processorFeatureFlags
-   //
-#if defined(TR_TARGET_X86)
-   TR::Compiler->target.cpu.setX86ProcessorFeatureFlags(queryX86TargetCPUID(jitConfig->javaVM)->_featureFlags);
-   TR::Compiler->target.cpu.setX86ProcessorFeatureFlags2(queryX86TargetCPUID(jitConfig->javaVM)->_featureFlags2);
-   TR::Compiler->target.cpu.setX86ProcessorFeatureFlags8(queryX86TargetCPUID(jitConfig->javaVM)->_featureFlags8);
-#endif // TR_TARGET_X86
-   }
-
-
-const char *
-TR_J9VMBase::getX86ProcessorVendorId()
-   {
-   TR_ASSERT(x86VendorIDInitialized, "X86 Vendor ID has not yet been initialized");
-   return x86VendorID;
-   }
-
-uint32_t
-TR_J9VMBase::getX86ProcessorSignature()
-   {
-   TR_ASSERT(_jitConfig, "jitConfig is not defined!");
-   TR_ASSERT(_jitConfig->javaVM, "jitConfig->javaVM is not defined!");
-   return queryX86TargetCPUID(_jitConfig->javaVM)->_processorSignature;
-   }
-
-uint32_t
-TR_J9VMBase::getX86ProcessorFeatureFlags()
-   {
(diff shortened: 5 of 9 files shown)


harryyu1994:
The OpenJ9 side change is dependent on the OMR side change therefore will come back to this once the OMR side is merged.

harryyu1994:
Travis CI build won't pass until the OMR side change is merged.

andrewcraik:
If this is a coordinated change we should launch sanity with reference to the dependent change to get the change checked before either of the parts are merged.

harryyu1994:
@andrewcraik Hi Andrew, so could you launch one for me here? The pull request on the OMR side is this one: https://github.com/eclipse/omr/pull/2425

andrewcraik:
Jenkins test sanity depends eclipse/omr#2425

andrewcraik:
FYI @0dvictor

runtime/compiler/trj9/x/env/J9CPU.cpp
```diff
@@ -31,64 +31,52 @@ namespace J9
...
-uint32_t
-CPU::getX86ProcessorFeatureFlags(TR::Compilation *comp)
-   {
-   return comp->fej9()->getX86ProcessorFeatureFlags();
-   }
-
-uint32_t
-CPU::getX86ProcessorFeatureFlags2(TR::Compilation *comp)
+bool
+CPU::testOSForSSESupport()
```
0dvictor: Why not remove this function completely?

runtime/compiler/trj9/x/env/J9CPU.hpp
```diff
@@ -62,34 +63,23 @@ class CPU : public J9::CPU
...
-   const char *getX86ProcessorVendorId(TR::Compilation *comp);
-   uint32_t getX86ProcessorSignature(TR::Compilation *comp);
-   uint32_t getX86ProcessorFeatureFlags(TR::Compilation *comp);
-   uint32_t getX86ProcessorFeatureFlags2(TR::Compilation *comp);
-   uint32_t getX86ProcessorFeatureFlags8(TR::Compilation *comp);
+   bool testOSForSSESupport();
+   bool getX86SupportsSSE4_1();
+   bool getX86SupportsHLE();
+   bool getX86SupportsPOPCNT();
+
```
0dvictor: getX86SupportsSSE4_1(), getX86SupportsHLE(), getX86SupportsPOPCNT() are latency functions. They are duplicate to these feature checks. https://github.com/eclipse/omr/blob/b3e2f861e48894b42751e943e197606d685d50a4/compiler/x/codegen/OMRCodeGenerator.hpp#L169
It may be a good time to get ride of these completely.

harryyu1994: Hi Victor, thanks for the review. For getX86SupportsHLE(), it is easy to remove and replace because the HLE feature checks only happen in x86 specific files. But for getX86SupportsSS4_1() and getX86SupportsPOPCNT(), the checks show up in cross-platform files, so if we don't declare them for all platforms or put Macro guards around them, we will get compile errors on non X86 platforms. What do you think we should do with those two?

harryyu1994: i.e. the feature check of getX86SupportsSSE4_1() shows up in env/VMJ9.cpp `if ((TR::Compiler->target.cpu.isX86() && comp->cg()->getX86ProcessorInfo().supportsSSE4_1())`

runtime/compiler/trj9/x/env/J9CPU.cpp
harryyu1994: For this function:
there is a OMR::X86::CPU::testOSForSSESupport(), which returns false.
and a J9::X86::CPU::testOSForSSESupport(), which returns true.
There are two calls of this function in x/codegen/OMRCodeGenerator.cpp. If we remove the J9 version, they will return false because OMR version will get called. Do you want the J9 version to be removed? and change the OMR version to return true maybe? I'm not understanding why we want this removed.

0dvictor: I see, I guess we have to wait until OMR sorts out what to do with ```testOSForSSESupport()``` first. Let's keep it as is for now.

harryyu1994:
@0dvictor Hi Victor, would you kindly review this change again? (and maybe also the OMR side change as well https://github.com/eclipse/omr/pull/2425 since they are related). I've done some manual verification, seems to be working.

harryyu1994:
Hi Andrew @andrewcraik could you please launch a jenkins test sanity on this? Also could you help me merge the two changes once the tests pass? (OMR side and OpenJ9 side change depend on each other, will likely break the build for a moment)

runtime/compiler/trj9/env/J9CPU.hpp
```diff
@@ -45,6 +46,14 @@ class CPU : public OMR::CPUConnector
 public:
 
    J9ProcessorDesc *TO_PORTLIB_getJ9ProcessorDesc();
+
+   // The following methods should not be called in the base class here
+   // They are just here so that these x86 specific methods have a declaration/definition during compilation
+   // x86 specific functions:
+   bool getX86SupportsSSE4_1() { TR_ASSERT(false, "Should not be called!"); return false; }
+   const char *getX86ProcessorVendorId() { TR_ASSERT(false, "Should not be called!"); return NULL; }
+   uint32_t getX86ProcessorSignature() { TR_ASSERT(false, "Should not be called!"); return 0; }
```
mstoodle: is there a reason these are going in J9CPU and not pushed all the way down to OMRCPU in the OMR project?

runtime/compiler/trj9/x/env/J9CPU.cpp
```diff
@@ -24,71 +24,85 @@
...
    {
-   return comp->fej9()->getX86ProcessorSignature();
+   static char buf[13];
+
+   // Terminate the vendor ID with NULL before returning.
+   //
+   strncpy(buf, self()->queryX86TargetCPUID()->_vendorId, 12);
+   buf[12] = '\0';
+
+   return buf;
```
mstoodle: Isn't this code a duplicate of what's in OMR? Why is it needed here?

runtime/compiler/trj9/x/env/J9CPU.cpp
```diff
@@ -24,71 +24,85 @@
...
    }
 
 uint32_t
-CPU::getX86ProcessorFeatureFlags(TR::Compilation *comp)
+CPU::getX86ProcessorSignature()
    {
-   return comp->fej9()->getX86ProcessorFeatureFlags();
+   return self()->queryX86TargetCPUID()->_processorSignature;
```
mstoodle: same "duplicate of OMR" comment here

runtime/compiler/trj9/x/env/J9CPU.cpp
```diff
@@ -24,71 +24,85 @@
...
    }
 
 uint32_t
-CPU::getX86ProcessorFeatureFlags2(TR::Compilation *comp)
+CPU::getX86ProcessorFeatureFlags()
    {
-   return comp->fej9()->getX86ProcessorFeatureFlags2();
+   return self()->queryX86TargetCPUID()->_featureFlags;
```
mstoodle: same "duplicate of OMR" comment here

runtime/compiler/trj9/x/env/J9CPU.cpp
```diff
@@ -24,71 +24,85 @@
...
    }
 
 uint32_t
-CPU::getX86ProcessorFeatureFlags8(TR::Compilation *comp)
+CPU::getX86ProcessorFeatureFlags2()
    {
-   return comp->fej9()->getX86ProcessorFeatureFlags8();
+   return self()->queryX86TargetCPUID()->_featureFlags2;
```
mstoodle: same "duplicate of OMR" comment here

runtime/compiler/trj9/x/env/J9CPU.cpp
```diff
@@ -24,71 +24,85 @@
...
    }
 
-bool
-CPU::testOSForSSESupport(TR::Compilation *comp)
+uint32_t
+CPU::getX86ProcessorFeatureFlags8()
    {
-   return comp->fej9()->testOSForSSESupport();
+   return self()->queryX86TargetCPUID()->_featureFlags8;
```
mstoodle: same "duplicate of OMR" comment here

runtime/compiler/trj9/x/env/J9CPU.cpp
```diff
@@ -24,71 +24,85 @@
...
    }
 
-void
-CPU::setX86ProcessorFeatureFlags(uint32_t flags)
+bool
+CPU::testOSForSSESupport()
    {
-   _featureFlags = flags;
+   return true; // VM guarantees SSE/SSE2 are available
```
mstoodle: Does this mean that the VM will fail to start if SSE/SSE2 are not supported on the current processor?

harryyu1994: This code change was ported over from 
```
-bool
-TR_J9VMBase::testOSForSSESupport()
-   {
-   return true; // VM guarantees SSE/SSE2 are available
-   }
```
It will have the same behaviour as before. I don't have an answer for this question :(

runtime/compiler/trj9/env/J9CPU.hpp
harryyu1994: These 3 functions contain call sites that are exposed to all platforms, which means for non-X platforms, these functions won't have a declaration/definition due to its x/J9CPU class not being there. So this is just a workaround.

runtime/compiler/trj9/x/env/J9CPU.cpp
harryyu1994: ^ that was my guess what the problem was, I could be completely wrong

harryyu1994: Just did another set of tests, so "my duplicates" was not able to get rid of this error `[ERR] JVMJ9VM011W Unable to load j9jit29: /tmp/bld_386111/j2sdk-image/jre/lib/amd64/compressedrefs/libj9jit29.so: undefined symbol: _ZN2TR9JitConfig8instanceEv`.   `_ZN2TR9JitConfig8instanceEv` is something in OMRCPU's `queryX86TargetCPUID()`.

harryyu1994: I will get rid of the duplicates, these did not resolve the problem.

harryyu1994: I think removing the J9 specific macro in OMRCPU's queryX86TargetCPUID() is causing this undefined symbol problem (_ZN2TR9JitConfig8instanceEv).

harryyu1994:
@mstoodle Hi Mark, this PR is also ready for review. (and probably need a `Jenkins test sanity depends eclipse/omr#2425`). Thanks!

runtime/compiler/env/J9CPU.hpp
```diff
@@ -45,6 +46,14 @@ class CPU : public OMR::CPUConnector
 public:
 
    J9ProcessorDesc *TO_PORTLIB_getJ9ProcessorDesc();
+
+   // The following methods should not be called in the base class here
+   // They are just here so that these x86 specific methods have a declaration/definition during compilation
+   // x86 specific functions:
+   bool getX86SupportsSSE4_1() { TR_ASSERT(false, "Should not be called!"); return false; }
```
andrewcraik: I'm not sure why these are on the top level class - they should only be on x86 and we should only call them on an x86 codegenerator so not sure why they need to be here? The asserts message should be more helpful IMO.

harryyu1994: There's a call to this function in VMJ9.cpp, which is why I put an empty definition there in the top level class. Won't compile without it.
```
case TR::java_lang_String_hashCodeImplCompressed:
            if (!TR::Compiler->om.canGenerateArraylets()){
               if ((TR::Compiler->target.cpu.isX86() && TR::Compiler->target.cpu.getX86SupportsSSE4_1()) ||
                   (TR::Compiler->target.cpu.isZ() && comp->cg()->getSupportsVectorRegisters()))
                  {
                  dontInlineRecognizedMethod = true;
                  }
            }
```

runtime/compiler/x/env/J9CPU.cpp
```diff
@@ -24,71 +24,65 @@
...
 
-void
-CPU::setX86ProcessorFeatureFlags2(uint32_t flags2)
+bool
+CPU::getX86SupportsSSE4_1()
    {
-   _featureFlags2 = flags2;
+   if ((self()->getX86ProcessorFeatureFlags2() & TR_SSE4_1) != 0x00000000)
+      return true;
+   else return false;
```
mstoodle: not the correct indentation

runtime/compiler/x/env/J9CPU.cpp
```diff
@@ -24,71 +24,65 @@
...
 
-void
-CPU::setX86ProcessorFeatureFlags8(uint32_t flags8)
+bool
+CPU::hasPopulationCountInstruction()
    {
-   _featureFlags8 = flags8;
+   if ((self()->getX86ProcessorFeatureFlags2() & TR_POPCNT) != 0x00000000)
+      return true;
+   else return false;
```
mstoodle: indentation

runtime/compiler/env/J9CPU.hpp
mstoodle: Presumably that first line is the X86 equivalent of `getSupportsVectorRegisters()` which seems to be a Z-specific query unused on X86 or POWER. If that's right, that whole condition could be reduced to `if (comp->cg()->getSupportsVectorRegisters())` ? Hm, there are some optimizer uses of that query in both OpenJ9 and OMR that would need to be checked for validity and possibly augmented to conserve validity, but I think that would enable these X86-specific queries to be sunk into x86-specific code and therefore the queries themselves to sink into J9::X86::CPU which, I agree, would be a better state to be in.

What do you think @andrewcraik @harryyu1994 ?

I agree the assert messages should be more suggestive of the likely problem "X86 specific query called on non-X86 platform" or something like that, but that may be unnecessary if the queries can be moved out of the generic class.

andrewcraik: so the hashCodeImplCompressed shouldn't be based on vector queries - it should be based on whether the CG supports the accelerated implementation - that should probably be a query on the J9 code generators.

In general I agree and support Mark's view that queries in the optimizer shouldn't be talking about hardware features, but should be asking about codegen capabilities like supportVectorRegisters or maybe it should be getVectorRegisterWidth and a value greater than 0 tells you the width (all currently assume 128bits, but there are plans to change that in the not too distant future to better exploit AVX/SSE).

I think that if this refactoring is happening making sure these hardware details are not referenced outside the CG and making sure the CG has the right capability queries makes a lot of sense and prevents the kind of pollution of the CG API that is otherwise necessary as you have demonstrated with this current iteration of the clean-up.

harryyu1994: So is it true for x86 that `if (TR::Compiler->target.cpu.getX86SupportsSSE4_1())` then `comp->cg()->getSupportsVectorRegisters` is true and otherwise false? 

If not the case, which J9 CG query should I use to replace `TR::Compiler->target.cpu.getX86SupportsSSE4_1()` here. Do we have an existing CG query for "whether the CG supports the accelerated implementation"?
@andrewcraik

runtime/compiler/env/J9CPU.hpp
```diff
@@ -45,6 +46,15 @@ class CPU : public OMR::CPUConnector
 public:
 
    J9ProcessorDesc *TO_PORTLIB_getJ9ProcessorDesc();
+
+   // The following method isn't supposed to be on the top level class
+   // However, there exists places where this x86 specific query was called in common code
+   // Its empty definition is here to avoid compilation failures.
+   bool getX86SupportsSSE4_1() { TR_ASSERT(false, "X86 specific query called on non-X86 platform!"); return false; }
```
harryyu1994: @mstoodle ; Hi Mark, I'm still not sure how to get rid of this guy `getX86SupportsSSE4_1()` on the top level class. Can we just leave this workaround as it is and let the X codegen/optimizer people figure out what to do with it? if not, I'm gonna need help with this one lol.

0dvictor: With current code, it is impossible to get rid of this guy; however the place where it is queried is such a hack and we should clean it up. It is straightforward to clean it up but it is beyond the scope of this pull request. Let me create a pull request to remove its caller so that you can safely remove this guy once merged.

0dvictor: Once https://github.com/eclipse/openj9/pull/2719 is merged, `getX86SupportsSSE4_1()` can be removed.

runtime/compiler/x/env/J9CPU.cpp
```diff
@@ -24,71 +24,67 @@
...
+CPU::testOSForSSESupport()
    {
-   _featureFlags = flags;
+   return true; // VM guarantees SSE/SSE2 are available
    }
 
-void
-CPU::setX86ProcessorFeatureFlags2(uint32_t flags2)
+bool
+CPU::getX86SupportsSSE4_1()
```
0dvictor: This can also be removed once #2719 merged.

runtime/compiler/x/env/J9CPU.cpp
```diff
@@ -24,71 +24,67 @@
...
 
 bool
-CPU::testOSForSSESupport(TR::Compilation *comp)
-   {
-   return comp->fej9()->testOSForSSESupport();
-   }
-
-void
-CPU::setX86ProcessorFeatureFlags(uint32_t flags)
+CPU::testOSForSSESupport()
```
0dvictor: This can be removed if https://github.com/eclipse/omr/pull/2822 is merged before this PR gets merged.

andrewcraik:
@harryyu1994 and @0dvictor what's the status on this one now following the review comments and questions, rebasing, etc?

runtime/compiler/x/env/J9CPU.hpp
```diff
@@ -62,34 +63,23 @@ class CPU : public J9::CPU
...
-   uint32_t getX86ProcessorSignature(TR::Compilation *comp);
-   uint32_t getX86ProcessorFeatureFlags(TR::Compilation *comp);
-   uint32_t getX86ProcessorFeatureFlags2(TR::Compilation *comp);
-   uint32_t getX86ProcessorFeatureFlags8(TR::Compilation *comp);
+   TR_X86CPUIDBuffer *queryX86TargetCPUID();
+   const char * getProcessorVendorId();
+   uint32_t getProcessorSignature();
+
+   bool testOSForSSESupport();
+   bool getX86SupportsSSE4_1();
```
0dvictor: This can be removed as #2719 has been merged.

mstoodle:
Jenkins test sanity depends eclipse/omr#2425

harryyu1994:
Ready for review.

0dvictor:
LGTM.

mstoodle:
Jenkins test sanity depends eclipse/omr#2425

mstoodle:
Jenkins test sanity all jdk8,jdk11 depends eclipse/omr#2425

mstoodle:
jenkins test sanity x86 jdk8,jdk11 depends eclipse/omr#2425

mstoodle:
jenkins test sanity xlinux,win,win32 jdk8,jdk11 depends eclipse/omr#2425

mstoodle:
some outstanding problems not related to this PR...will try again later

harryyu1994:
Should I rebase both changes to latest?

mstoodle:
@harryyu1994 that would be great, yes please :) .

mstoodle:
Awesome! Thanks @harryyu1994 !

jenkins test sanity x86 jdk8,jdk11 depends eclipse/omr#2425

mstoodle:
Not sure this will work due to other issues, but let's see what happens:

jenkins test sanity xlinux,win,win32,osx jdk8,jdk11 depends eclipse/omr#2425

0xdaryl:
Jenkins test sanity xlinux,win,win32,osx jdk8,jdk11 depends eclipse/omr#2425

0xdaryl:
Travis failure is expected due to OMR dependency.

mstoodle:
OpenJ9 has asked that we hold off on this co-dependent merge until their release is closer to closed away (even though they've already split the release branch). Anyway, that request translated to "please hold off until Monday". Since this PR isn't urgent, I'm accepting the request and will revisit this PR on Monday.

mstoodle:
Merging now along with dependent OMR PR

PR #1614, opened 2018-04-04
Fix an issue with instanceof and multi-dimensional arrays

klangman:
When using instanceof for a multi-dimension array of Objects
(i.e. Object[][]) the JIT would fail to verify the number of
dimensions and only insure that the object is non-null and is
not an array of primitives. This fix will cause the JIT to
emit a proper set of tests and fall back to a helper routine
after the fast path tests. This change only effects Power and
ZSeries as this common code facility is not used on X86.


andrewcraik:
change looks reasonable - Jenkins test sanity

andrewcraik:
Jenkins test sanity aix jdk9

andrewcraik:
last run failed for some mysterious reason that looks unrelated to this change - try again - Jenkins test sanity aix jdk9

PR #1617, opened 2018-04-05
Improve start-up time when not using AOT

mpirvu:
There are two changes implemented by this commit:
1) Reduce the invocation counts for methods belonging to bootstrap classes
2) Downgrade to cold compilations performed during start-up phase

The motivation for change (1) is to get out of interpreter sooner (the
invocation count is the number of times an interpreted method needs to
be invoked to be JIT compiled). The main problem is that a lower
invocation count reduces the number of interpreter profiling samples
which in turn may affect throughput. One way around this issue is to restrict
this change to methods belonging to bootstrap classes that are executed a
lot during start-up but to a lesser degree during stready state.

The motivation for change (2) is that, with a lower invocation count more
methods are going to be compiled and the extra compilation overhead may
affect start-up time. Thus, we need to reduce the compilation overhead
by using lower optimization levels (i.e. "cold"). Methods compiled at
"cold" will be upgraded through the GCR mechanism if they are executed
enough times.

There are two new options to control change (1)
-Xjit:bcountForBootstrapMethods=<nnn>   // for methods with loops
-Xjit:countForBootstrapMethods=<nnn>    // for methods without loops
and one new option to control change (2)
-Xjit:queueSizeThresholdToDowngradeOptLevelDuringStartup=<nnn>

Performance experiments on x86 showed good startup improvements for
Liberty Application Server (6-7%) and no throughput regressions for a
few Liberty based benchmarks.

In this commit the feature will be disabled to allow for more performance
testing on other platforms.

### DIFF
--- a/runtime/compiler/trj9/control/HookedByTheJit.cpp
+++ b/runtime/compiler/trj9/control/HookedByTheJit.cpp
@@ -564,7 +564,11 @@ static void jitHookInitializeSendTarget(J9HookInterface * * hook, UDATA eventNum
          } // if (TR::Options::sharedClassCache())
       if (count == -1) // count didn't change yet
          {
-         count = getCount(romMethod, optionsJIT, optionsAOT);
+         if (!TR::Options::getCountsAreProvidedByUser() &&
+            fe->isClassLibraryMethod((TR_OpaqueMethodBlock *)method))
+            count = J9ROMMETHOD_HAS_BACKWARDS_BRANCHES(romMethod) ? TR::Options::getCountForLoopyBootstrapMethods() : TR::Options::getCountForLooplessBootstrapMethods();
+         if (count == -1)
+            count = getCount(romMethod, optionsJIT, optionsAOT);
 
          // If add-spreading-invocation mode is enabled, add a hash value (hashed from method name) to count
          if (optionsAOT->getOption(TR_EnableCompilationSpreading) || optionsJIT->getOption(TR_EnableCompilationSpreading))
@@ -578,7 +582,6 @@ static void jitHookInitializeSendTarget(J9HookInterface * * hook, UDATA eventNum
                count = count * TR::Options::_countPercentageForEarlyCompilation / 100;
             // TODO: disable this mechanism after an hour or so to conserve idle time
             }
-
          }
       }
 
--- a/runtime/compiler/trj9/control/J9Options.cpp
+++ b/runtime/compiler/trj9/control/J9Options.cpp
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2000, 2017 IBM Corp. and others
+ * Copyright (c) 2000, 2018 IBM Corp. and others
  *
  * This program and the accompanying materials are made available under
  * the terms of the Eclipse Public License 2.0 which accompanies this
@@ -85,6 +85,9 @@ int32_t J9::Options::_smallMethodBytecodeSizeThresholdForCold = -1; // -1 means 
 
 int32_t J9::Options::_countForMethodsCompiledDuringStartup = 10;
 
+int32_t J9::Options::_countForLoopyBootstrapMethods = -1; // -1 means feature disabled
+int32_t J9::Options::_countForLooplessBootstrapMethods = -1; // -1 means feature disabled
+
 TR::SimpleRegex *J9::Options::_jniAccelerator = NULL;
 
 int32_t J9::Options::_classLoadingPhaseInterval = 500; // ms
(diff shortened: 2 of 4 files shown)


mpirvu:
@ymanton Younes, could you please review this change? Thanks.

runtime/compiler/trj9/control/J9Options.cpp
```diff
@@ -85,6 +85,9 @@ int32_t J9::Options::_smallMethodBytecodeSizeThresholdForCold = -1; // -1 means
 
 int32_t J9::Options::_countForMethodsCompiledDuringStartup = 10;
 
+int32_t J9::Options::_countForLoopyBootstrapMethods = 3000; // 3000 means disable the feature by default
+int32_t J9::Options::_countForLooplessBootstrapMethods = 3000; // means disable the feature by default
```
ymanton: Why does 3000 mean disabled? I'm assuming it's because the default count and bcount are also 3000. The symbolic constants for those values are `TR_DEFAULT_INITIAL_COUNT` and `TR_DEFAULT_INITIAL_BCOUNT` but even then it looks like the actual initial count we go with is determined at run time based on other options (e.g. quickstart), so if the goal is to keep these changes disabled for now by keeping the behaviour the same this might not be sufficient. It seems like these counts shouldn't be statically initialized, they should be initialized with whatever values we decide on for the default counts.

runtime/compiler/trj9/control/HookedByTheJit.cpp
```diff
@@ -564,7 +564,11 @@ static void jitHookInitializeSendTarget(J9HookInterface * * hook, UDATA eventNum
          } // if (TR::Options::sharedClassCache())
       if (count == -1) // count didn't change yet
          {
-         count = getCount(romMethod, optionsJIT, optionsAOT);
+         if (!TR::Options::getCountsAreProvidedByUser() &&
+            fe->isClassLibraryMethod((TR_OpaqueMethodBlock *)method))
```
ymanton: The description for this change says we want to target "methods belonging to bootstrap classes that are executed a lot during start-up but to a lesser degree during steady state" but the check here is `isClassLibraryMethod()` which is basically every method from JCL and J9VM classes (regardless of when and how often it's executed) isn't it? Is there some other logic elsewhere that actually limits the behaviour to what was desired?

runtime/compiler/trj9/control/J9Options.cpp
mpirvu: Yeah, I was loose here because I am about to change them to 1000/250 anyway. I'll do better.

runtime/compiler/trj9/control/HookedByTheJit.cpp
mpirvu: What I meant in the description was that in general methods belonging to bootstrap classes are executed a lot during start-up/JVM-bootstrap by their nature. The counts are set when the classes/methods are loaded, so there is no way to dynamically determine how often they are going to be executed at that point.

runtime/compiler/trj9/control/J9Options.cpp
mpirvu: I changed the code to set the initial value of `_countForLoopyBootstrapMethods` and  `_countForLooplessBootstrapMethods` to -1. This will enable the old code path to be taken.

runtime/compiler/trj9/control/HookedByTheJit.cpp
ymanton: Is there any concern that this could negatively affect programs that make heavy use of the JCL long after startup? E.g. programs that use `java/{util/math/io}` or something of that nature in their primary task? Since the functionality will be disabled until more performance measurements are done I don't think we need an answer to this now, but finding/measuring important programs that go against this assumption is probably a worthwhile data point.

ymanton:
Jenkins test all

runtime/compiler/trj9/control/HookedByTheJit.cpp
mpirvu: My first attempt was actually using this kind of filtering thinking that throughput in Daytrader/AcmeAir would be affected. To my surprise it wasn't (actually I got a small improvement which may or may not be noise). The goal of this commit is to have a build that I could use for more benchmarks and on other platforms using the automation provided by the perf monitoring infrastructure.

ymanton:
Re: the test failures, they all look infra-related. At least one of sanity/extended passed for each combination of platform and JDK and I didn't actually intend to run extended so I think that's sufficient to merge this.

ymanton:
I'll wait for you to remove the WIP tag if/when you're ready to proceed.

mpirvu:
ok. I removed the WIP tag as there is nothing else to do for this commit.

Issue #1273, opened 2018-02-26
ConstantDynamic: Add support for balanced GC policy


tajila:
Update the `GC_ConstantPoolObjectSlotIterator::nextSlot()` method to recognize Constant dynamic CP entries

dmitripivkine:
I guess this item would be empty as far as Balanced does CP scan already. So probably changes (if any) will be required on level of CP Iterator

tajila:
@dmitripivkine Does balanced scan in all cases?

If we added a bit to the Class object or J9Class saying there were no non-tenured objects in the constantpool would it still have to scan?

> will be required on level of CP Iterator
Is the CP Iterator common for all policies?

We are introducing two new CP types, the iterator(s) will have to know about them

dmitripivkine:
Yes, Balanced does scan CP in all cases. I believe keep this item for now. We will double check after implementation.
I believe CP Iterator is common for all GC policies

dmitripivkine:
Actually there are two iterators related to CP walk:
- GC_ConstantPoolObjectSlotIterator iterates heap objects in CP
- GC_ConstantPoolClassSlotIterator iterates classes in CP

PR #1620, opened 2018-04-05
Add Constant Dynamic support in GC

fengxue-IS:
Update GC_ConstantPoolObjectSlotIterator to recognize constantDynamic

Close: #1273

### DIFF
--- a/runtime/gc_structs/ConstantPoolObjectSlotIterator.cpp
+++ b/runtime/gc_structs/ConstantPoolObjectSlotIterator.cpp
@@ -1,6 +1,6 @@
 
 /*******************************************************************************
- * Copyright (c) 1991, 2014 IBM Corp. and others
+ * Copyright (c) 1991, 2018 IBM Corp. and others
  *
  * This program and the accompanying materials are made available under
  * the terms of the Eclipse Public License 2.0 which accompanies this
@@ -43,6 +43,7 @@ GC_ConstantPoolObjectSlotIterator::nextSlot()
 {
 	U_32 slotType;
 	j9object_t *slotPtr;
+	j9object_t *result = NULL;
 
 	while(_cpEntryCount) {
 		if(0 == _cpDescriptionIndex) {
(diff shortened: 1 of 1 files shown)


fengxue-IS:
depends: #1426

runtime/gc_structs/ConstantPoolObjectSlotIterator.cpp
```diff
@@ -73,6 +73,13 @@ GC_ConstantPoolObjectSlotIterator::nextSlot()
 		if(slotType == J9CPTYPE_METHODHANDLE) {
 			return &(((J9RAMMethodHandleRef *) slotPtr)->methodHandle);
 		}
+		if(slotType == J9CPTYPE_CONSTANT_DYNAMIC) {
+			if(((J9RAMConstantDynamicRef *) slotPtr)->value != NULL) {
```
dmitripivkine: Would you please put NULL to the left side:
if(NULL != ((J9RAMConstantDynamicRef *) slotPtr)->value) {

runtime/gc_structs/ConstantPoolObjectSlotIterator.cpp
```diff
@@ -73,6 +73,13 @@ GC_ConstantPoolObjectSlotIterator::nextSlot()
 		if(slotType == J9CPTYPE_METHODHANDLE) {
 			return &(((J9RAMMethodHandleRef *) slotPtr)->methodHandle);
 		}
+		if(slotType == J9CPTYPE_CONSTANT_DYNAMIC) {
```
dmitripivkine: Would you mind to replace all if's to the switch? I believe it will work faster and look cleaner

dmitripivkine:
This PR can be tested and submitted when PRs it depends at are ready and in.

DanHeidinga:
Does this change actually depend on #1533?  Or is possible to merge it before the resolve code is there as the GC should just see both slots as null until we resolve one.

fengxue-IS:
@DanHeidinga this doesn't depend on the code change from #1533, I added that as a dependency since this will only be used after #1533 has been merged
fixed compile issue, testing

DanHeidinga:
@dmitripivkine This has changed slightly since you last reviewed it.  Do you want to give it another look?

dmitripivkine:
jenkins compile pLinux jdk8

dmitripivkine:
Jenkins test sanity aix jdk9

PR #1622, opened 2018-04-05
Adopt OMR's abs evaluators on X86

0dvictor:
OMR has full support for [i/l/f/d]abs evaluators,
adopting them


andrewcraik:
change looks good - Jenkins test sanity

PR #1628, opened 2018-04-06
Do not report references to object type from Class objects

pdbain-ibm:
Per JVMTI specification for FollowReferences(), do not report the class of
Class objects.

### DIFF
--- a/runtime/jvmti/jvmtiHeap.c
+++ b/runtime/jvmti/jvmtiHeap.c
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 1991, 2017 IBM Corp. and others
+ * Copyright (c) 1991, 2018 IBM Corp. and others
  *
  * This program and the accompanying materials are made available under
  * the terms of the Eclipse Public License 2.0 which accompanies this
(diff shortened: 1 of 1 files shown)


pdbain-ibm:
@DanHeidinga or @gacholio would you kindly do an initial review while I regression test the change?

DanHeidinga:
@pdbain-ibm Has the expected behaviour changed between different JDK releases?  Which version of the JVMTI spec describes this?

pdbain-ibm:
I am referring to Version 1.2 of the JVMTI specification, so I believe this behaviour was intended from the beginning, though never tested.  We now have a test which checks the kinds of references reported.

runtime/jvmti/jvmtiHeap.c
```diff
@@ -1137,7 +1137,20 @@ mapEventType(J9JVMTIHeapData * data, IDATA type, jint index, j9object_t referrer
             event->type = J9JVMTI_HEAP_EVENT_OBJECT;
```
gacholio: While you're here, please fix this line to use tabs to indent.

gacholio:
Change looks reasonable to me.  I'll do a final review once the testing has completed.

runtime/jvmti/jvmtiHeap.c
pdbain-ibm: Done.

pdbain-ibm:
Ready for merging.  Please run Jenkins test builds.  Thank you.

DanHeidinga:
Jenkins test sanity plinux

PR #1636, opened 2018-04-09
Prefetching, checkcasts and profiled compile changes

IBMJimmyk:
HashMap.resize and HashMapHashIterator.nextNode were added to
canSkipCheckCasts since they don't need check casts.

Node threshold inside ProfileGenerator was changed to 90000. This will
promote more profiled compilations and was shown to improve the throughput
of an internal appserver benchmark.

addPrefetch was moved from OMR to OpenJ9 and renamed to
insertPrefetchIfNecessary.

It was also modified to work on Itr.next in ArrayList and the prefetching
stride distance was also changed. The new value was found through testing
and displayed an improvement in throughput in the previously mentioned
internal appserver benchmark.

Also during the move, an bug in insertPrefetchIfNecessary where
compressedrefs decompression was not occuring was fixed.

iwrtbarEvaluator was modified to now use insertPrefetchIfNecessary instead
of addPrefetch.

### DIFF
--- a/runtime/compiler/il/symbol/J9MethodSymbol.cpp
+++ b/runtime/compiler/il/symbol/J9MethodSymbol.cpp
@@ -353,6 +353,8 @@ J9::MethodSymbol::safeToSkipDivChecks()
 static TR::RecognizedMethod canSkipCheckCasts[] =
    {
    TR::java_util_ArrayList_ensureCapacity,
+   TR::java_util_HashMap_resize,
+   TR::java_util_HashMapHashIterator_nextNode,
    TR::java_util_Hashtable_clone,
    TR::java_util_Hashtable_putAll,
    TR::unknownMethod
--- a/runtime/compiler/optimizer/ProfileGenerator.cpp
+++ b/runtime/compiler/optimizer/ProfileGenerator.cpp
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2000, 2017 IBM Corp. and others
+ * Copyright (c) 2000, 2018 IBM Corp. and others
  *
  * This program and the accompanying materials are made available under
  * the terms of the Eclipse Public License 2.0 which accompanies this
@@ -58,7 +58,7 @@
 #include "optimizer/TransformUtil.hpp"         // for calculateElementAddress
 
 #define OPT_DETAILS "O^O PROFILE GENERATOR: "
-#define MAX_NUMBER_OF_ALLOWED_NODES 30000
+#define MAX_NUMBER_OF_ALLOWED_NODES 90000
 
 #ifdef DEBUG
 // Define range of asyncs to become split points, for debugging.
--- a/runtime/compiler/p/codegen/J9CodeGenerator.cpp
+++ b/runtime/compiler/p/codegen/J9CodeGenerator.cpp
@@ -372,3 +372,368 @@ J9::Power::CodeGenerator::enableAESInHardwareTransformations()
       return false;
    }
 
+void
+J9::Power::CodeGenerator::insertPrefetchIfNecessary(TR::Node *node, TR::Register *targetRegister)
+   {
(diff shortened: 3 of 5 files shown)


andrewcraik:
This is clearly a performance item - a performance justification will be required in terms of benchmarks run. I suspect forcing the inlining of CHM.get may not be the right choice in many situations due to code growth for minimal benefit. Benchmark measurements on DayTrader3 and HiBench would be the minimum need to justify such a change - PPC numbers are good, but x86 numbers would also be nice.

JITProfiling is going to be replaced soon by the work @r30shah is doing to get JProfiling enabled so the change to the profiling node limit may not be in play for that long and you may want to verify the newer technology will work.

@vijaysun-omr FYI.

andrewcraik:
As part of the perf runs a measure of the compile-time increase as well the throughput would be a good idea since the forced inlining may have a measurable impact on compile-time.

IBMJimmyk:
I updated my changes again.

I removed java_util_concurrent_ConcurrentHashMap_get as a recognized method and from being force inlined as it was causing a throughput regression in an internal benchmark.

Previous MAX_NUMBER_OF_ALLOWED_NODES inside ProfileGenerator was changed from 30000 to 3000000 which was an 100x increase. This turned out to be excessive. I tuned it back to 90000 and was still able to see the expected throughput increase in an internal appserver benchmark.

addPrefetch code was being used by the omr half of this change set (https://github.com/eclipse/omr/pull/2431) but that code was supposed to be in openj9 so I also moved it here under the name insertPrefetchIfNecessary. The one prexisting location in openj9 that called addPrefetch was updated to use insertPrefetchIfNecessary.

IBMJimmyk:
The following numbers are based on 6 baseline iterations and 6 new build iterations that are interleaved.

Performance on daytrader3 on broadwell, redhat 7.3, kernel version 3.10.0-514
Relative throughput new build vs baseline, the higher the score, the better the performance:

Benchmark | Relative score | Standard deviation
-- | -- | --
daytrader3 | -0.23% | 0.54%

Benchmark | Relative score | Standard deviation
-- | -- | --
Pagerank | 0.85% | 1.63%
Terasort | -3.48% | 7.42%
Wordcount | 3.43% | 4.61%
Sort | -0.44% | 1.06%
Sleep | 0.06% | 0.11%

The performance numbers are inside the noise margin so there's no negative impact on these benchmarks.

andrewcraik:
@IBMJimmyk can you address the signed off by check failure?

Of the changes I am fine with those in the optimizer - the skip checkcast and the profiling node count increase.

Could you correct the pull request message and the commit message to reflect the changes you are currently proposing.

In future, it would be nicer to see these changes separated into separate PRs to simplify review, but we can continue here for now. We need a codegen review @zl-wang or @gita-omr could one of you oblige given your expertise in the area?

IBMJimmyk:
I've already fixed the commit message and just fixed the pull request message to reflect the current changes. I'm not really sure what caused the sign off check failure. I used the same sign off I always have for commits but it is getting rejected for some reason. The error message just says "The following commit appears to have an incorrect sign-off" so it doesn't really say what's wrong.

andrewcraik:
There may be some infra issues about the sign-off @AdamBrousseau ?

AdamBrousseau:
The check looks for a blank second last line in each commit. There is an open question as to whether or not this is correct/required. See ibmruntimes/openj9-openjdk-jdk9#58

IBMJimmyk:
Adding the blank line seems to have worked. The example good commit in CONTRIBUTING.md should probably be updated at some point since it does not have a blank second last line so it would fail the check.

andrewcraik:
Jenkins test sanity

IBMJimmyk:
@andrewcraik It looks like there were problems running the tests. Is it possible to try launching them again?

andrewcraik:
Jenkins test sanity

IBMJimmyk:
@andrewcraik There are Windows failures but they look unrelated to my change. There's one failure each for "NPT ERROR: Cannot open library".

PR #1640, opened 2018-04-09
Fix memory reference of vector packed decimal

NigelYiboYu:
Fix a crash in the memory reference binary generation phase.

Pass the packed decimal load/store nodes to the
generateS390MemoryReference(TR::Node*,TR::CodeGenerator*,bool) API
call so that the resulting memory reference has the node's
symbol reference. Without a symbol reference, the codegen can crash in
MemoryReference::calcDisplacement().
This applies to mainline code that is performance critical.

For BCDCHK evaluator's OOL codgen path, the aladd address node is not
under a PD load/store node and, therefore, can not use the API mentioned
above. This address node is evaluated to a register, which is used to
construct a memory reference.

### DIFF
--- a/runtime/compiler/trj9/z/codegen/J9TreeEvaluator.cpp
+++ b/runtime/compiler/trj9/z/codegen/J9TreeEvaluator.cpp
@@ -15688,26 +15688,19 @@ J9::Z::TreeEvaluator::BCDCHKEvaluatorImpl(TR::Node * node,
 
    if(isResultPD)
       {
+      TR::Register* srcBaseReg = cg->evaluate(secondChild);
+      TR::MemoryReference* srcMR = generateS390MemoryReference(srcBaseReg, 0, cg);
+      int32_t resultSize = TR::DataType::packedDecimalPrecisionToByteLength(pdopNode->getDecimalPrecision());
+
       if(isUseVector)
          {
          TR_ASSERT(bcdOpResultReg && (bcdOpResultReg->getKind() == TR_VRF || bcdOpResultReg->getKind() == TR_FPR),
                    "Vector register expected\n");
 
-         int32_t resultSize = TR::DataType::packedDecimalPrecisionToByteLength(pdopNode->getDecimalPrecision());
-
-         // No need to evalute the address node (secondChild).
-         // generateVSIInstruction() API will call separateIndexRegister() to separate the index
-         // register by emitting an LA instruction. If there's a need for large displacement adjustment,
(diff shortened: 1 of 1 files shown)


NigelYiboYu:
@fjeremic Test passed and ready for review.

runtime/compiler/trj9/z/codegen/J9TreeEvaluator.cpp
```diff
@@ -15688,26 +15688,19 @@ J9::Z::TreeEvaluator::BCDCHKEvaluatorImpl(TR::Node * node,
 
    if(isResultPD)
       {
+      TR::Register* srcBaseReg = cg->evaluate(secondChild);
+      TR::MemoryReference* srcMR = generateS390MemoryReference(srcBaseReg, 0, cg);
```
fjeremic: Can we fuse these two lines to be:

```
generateS390MemoryReference(secondChild, cg, false);
```

NigelYiboYu: This `secondChild` is the second child of BCDCHK and is an `aladd` node. It's not under a load or store node. Hence, it can not use the `generateS390MemoryReference(TR::Node*,TR::CodeGenerator*,bool)` API.  This particular API seems to be designed for load/store nodes with SymbolReferences. It calls a `MemoryReference` constructor that only sets its own symbolReference field if the root node has the `hasSymbolRef` property. `aladd` node has no such property.

 If we do `generateS390MemoryReference(secondChild, cg, false)`, it'll likely crash in the `MemoryReference::calcDisplacement()` method due to NULL SymbolReference pointer.

NigelYiboYu: Besides, this two lines are used to generate packed decimal data copy back in the BCDCHK OOL path. It's not on the performance critical path so whether the address is folded or not makes little difference in terms of performance.

fjeremic: Ok, in that case I'd like to see another PR in OMR documenting the semantics of `generateS390MemoryReference(TR::Node*,TR::CodeGenerator*,bool)` and what it should be used for as this is a subtlety that is non-obvious.

fjeremic:
Jenkins test zlinux zos

fjeremic:
Jenkins test zlinux

fjeremic:
Jenkins test extended zlinux

PR #1603, opened 2018-04-03
Add functions for JVM signal support

babsingh:
Two static functions added in `jvminit.c`:
1) `signalDispatch`: Invoke `jdk.internal.misc.Signal.dispatch(int number)`
in Java 9 and onwards. Invoke `sun.misc.Signal.dispatch(int number)` in
Java 8.
2) `predefinedHandlerWrapper`: When a signal is raised, this handler will
be invoked by the `asynchSignalReporterThread` in `omrsignal.c` if it is
registered using `omrsig_set_*async_signal_handler` with a specific
signal.

One function introduced in `vm_api.h` and defined in `jvminit.c`:
1) `registerPredefinedHandler`: Registers `predefinedHandlerWrapper` using
`omrsig_set_*async_signal_handler` for a specified signal. Old OS handler is 
passed via an input reference variable.

`registerPredefinedHandler` is also added to `J9InternalFunctions` since it
will be invoked from `jvm.c::JVM_RegisterSignal`.

`freeJavaVM` invokes `j9sig_set_async_signal_handler(predefinedHandlerWrapper...` 
to unregister the pre-defined signal handler during shutdown.

Signed-off-by: Babneet Singh <sbabneet@ca.ibm.com>

PR #1642, opened 2018-04-09
Update JVM_*Signal functions

babsingh:
1) Add a macro to reflect if a signal is ignored (`J9_SIG_IGNORED`).

2) Rename `isSignalSpecial` to `isSignalReservedByJVM`, and `isSignalSpecial`
to `isSignalUsedForShutdown`.

3) Add function `isSignalIgnored` to check if a signal is ignored.

4) Move `SIGQUIT` check from `isSignalUsedForShutdown` to
`isSignalReservedByJVM`. Add more signals to `isSignalReservedByJVM` using
IBM signal documentation.

5) Update `JVM_RaiseSignal`: Ignore shutdown signals if `-Xrs` is present or
if a shutdown signal is ignored. Ignore `SIGQUIT` if `-Xrs` is present.

6) Update `JVM_RegisterSignal`: Add better comments. Return immediately
for error cases. Check for ignored shutdown signals. Use `sigemptyset` to
avoid blocking any signals. Set `SA_NODEFER` in `sigaction.sa_flags`.
`SA_NODEFER` doesn't prevent the signal from being received from within
its own signal handler.

7) Fix `JVM_RegisterSignal` and `JVM_RaiseSignal` w.r.t -Xrs cmdline option.

8) Add `SIGRECONFIG` to `jvm.c::signalMap`.

9) Added new trace points for `JVM_RegisterSignal` in `j9scar.tdf`.

10) Removed function `dummySignalHandler` in `jvm.c`.

11) In `JVM_RegisterSignal`, `registerPredefinedHandler` is invoked to
register the pre-defined signal handler. Link `JVM_RegisterSignal`
function with OMR Signal API.

### DIFF
--- a/runtime/j9vm/j9scar.tdf
+++ b/runtime/j9vm/j9scar.tdf
@@ -1,4 +1,4 @@
-// Copyright (c) 2006, 2017 IBM Corp. and others
+// Copyright (c) 2006, 2018 IBM Corp. and others
 //
 // This program and the accompanying materials are made available under
 // the terms of the Eclipse Public License 2.0 which accompanies this
@@ -334,3 +334,7 @@ TraceExit=Trc_SC_GCNoCompact_Exit Overhead=1 Level=1 Template="JVM_GCNoCompact()
 TraceEntry=Trc_SC_LoadSystemLibrary_Entry NoEnv Overhead=1 Level=1 Template="JVM_LoadSystemLibrary(name=%s)"
 TraceExit=Trc_SC_LoadSystemLibrary_Exit NoEnv Overhead=1 Level=1 Template="JVM_LoadSystemLibrary -- return 0x%zx"
 TraceEvent=Trc_SC_LoadSystemLibrary_LoadFailed NoEnv Overhead=1 Level=1 Template="JVM_LoadSystemLibrary(name=%s) failed to load"
+
+TraceEntry=Trc_SC_RegisterSignal_Entry Overhead=1 Level=1 Template="JVM_RegisterSignal(signal=%d, handler=%p)"
+TraceExit=Trc_SC_RegisterSignal_Exit Overhead=1 Level=1 Template="JVM_RegisterSignal -- return old OS handler = %p"
+TraceException=Trc_SC_RegisterSignal_FailedToRegisterHandler Overhead=1 Level=1 Template="Failed to register handler: OS signal value = %d, new signal handler = %p, old signal handler = %p"
--- a/runtime/j9vm/jvm.c
+++ b/runtime/j9vm/jvm.c
@@ -324,6 +324,7 @@ static BOOLEAN librariesLoaded(void);
 #endif /* defined(J9ZTPF) */
 
 #define J9_SIG_ERR -1
+#define J9_SIG_IGNORED 1
 
 #define J9_PRE_DEFINED_HANDLER_CHECK 2
 #define J9_OLDHANDLER_SAME_AS_NEWHANDLER 2
@@ -334,8 +335,9 @@ static BOOLEAN librariesLoaded(void);
 #define J9_SIGNAME_BUFFER_LENGTH 16
 
 static void dummySignalHandler(jint sigNum);
-static BOOLEAN isSignalSpecial(jint sigNum);
-static BOOLEAN isSignalUsedByVM(jint sigNum);
+static BOOLEAN isSignalUsedForShutdown(jint sigNum);
+static BOOLEAN isSignalReservedByJVM(jint sigNum);
+static BOOLEAN isSignalIgnoredByOS(jint sigNum);
 
 typedef struct {
 	const char *signalName;
@@ -484,6 +486,9 @@ static const J9SignalMapping signalMap[] = {
 #if defined(SIGXFSZ)
 	J9_SIGNAL_MAP_ENTRY("XFSZ", SIGXFSZ),
 #endif /* defined(SIGXFSZ) */
+#if defined(SIGRECONFIG)
+	J9_SIGNAL_MAP_ENTRY("RECONFIG", SIGRECONFIG),
+#endif /* defined(SIGRECONFIG) */
 	{NULL, J9_SIG_ERR}
 };
 
@@ -4408,15 +4413,18 @@ JVM_NativePath(char* path)
 }
 
 /**
- * Checks if a signal is special.
+ * Check if a signal is used for shutdown. SIGHUP,
+ * SIGINT and SIGTERM are characterized as shutdown signals.
+ * Terminator.setup() registers handlers for these signals
+ * during startup.
  *
  * @param sigNum Integer value of the signal
  *
- * @returns TRUE if the signal is special
- *          FALSE if the signal is not special
+ * @returns TRUE if the signal is used for shutdown
+ *          FALSE if the signal is not used for shutdown
  */
 static BOOLEAN
-isSignalSpecial(jint sigNum)
+isSignalUsedForShutdown(jint sigNum)
 {
 	return
 #if defined(SIGHUP)
@@ -4425,9 +4433,6 @@ isSignalSpecial(jint sigNum)
 #if defined(SIGINT)
 		(SIGINT == sigNum) ||
 #endif /* defined(SIGINT) */
-#if defined(SIGQUIT)
-		(SIGQUIT == sigNum) ||
-#endif /* defined(SIGQUIT) */
 #if defined(SIGTERM)
 		(SIGTERM == sigNum) ||
 #endif /* defined(SIGTERM) */
@@ -4435,15 +4440,29 @@ isSignalSpecial(jint sigNum)
 }
 
 /**
- * Check if a signal is used by the VM.
+ * Check if a signal is reserved by the VM. This function is invoked from
+ * JVM_*Signal functions. This function only includes signals which
+ * are listed in jvm.c::signalMap. If the signal is not listed in
+ * jvm.c::signalMap, then a sun.misc.Signal instance can't be created
+ * for a signal value. The native JVM_*Signal functions can't be invoked
+ * in the absence of a sun.misc.Signal instance.
+ *
+ * Synchronous signals are handled by the VM via omrsig_protect: SIGFPE,
+ * SIGILL, SIGSEGV, SIGBUS and SIGTRAP. On Windows, SIGBREAK is reserved
+ * by the VM to trigger dumps. On Unix/Linux platforms, SIGQUIT is
+ * reserved by the VM to trigger dumps. SIGABRT is reserved by the JVM
+ * for abnormal termination. SIGCHLD is reserved for internal control.
+ * On zOS, SIGUSR1 is reserved by the JVM. SIGRECONFIG is reserved to
+ * detect any change in the number of CPUs, processing capacity, or
+ * physical memory.
  *
  * @param sigNum Integer value of the signal
  *
- * @returns TRUE if the signal is used by the VM
- *          FALSE if the signal is not used by the VM
+ * @returns TRUE if the signal is reserved by the VM
+ *          FALSE if the signal is not reserved by the VM
  */
 static BOOLEAN
-isSignalUsedByVM(jint sigNum)
+isSignalReservedByJVM(jint sigNum)
 {
 	return
 #if defined(SIGFPE)
@@ -4455,13 +4474,77 @@ isSignalUsedByVM(jint sigNum)
 #if defined(SIGSEGV)
 		(SIGSEGV == sigNum) ||
 #endif /* defined(SIGSEGV) */
+#if defined(SIGBUS)
+		(SIGBUS == sigNum) ||
+#endif /* defined(SIGBUS) */
+#if defined(SIGTRAP)
+		(SIGTRAP == sigNum) ||
+#endif /* defined(SIGTRAP) */
+#if defined(SIGQUIT)
+		(SIGQUIT == sigNum) ||
+#endif /* defined(SIGQUIT) */
+#if defined(SIGBREAK)
+		(SIGBREAK == sigNum) ||
+#endif /* defined(SIGBREAK) */
+#if defined(SIGABRT)
+		(SIGABRT == sigNum) ||
+#endif /* defined(SIGABRT) */
+#if defined(SIGCHLD)
+		(SIGCHLD == sigNum) ||
+#endif /* defined(SIGCHLD) */
+#if defined(SIGRECONFIG)
+		(SIGRECONFIG == sigNum) ||
+#endif /* defined(SIGRECONFIG) */
+#if defined(J9ZOS390) && defined(SIGUSR1)
+		(SIGUSR1 == sigNum) ||
+#endif /* defined(J9ZOS390) && defined(SIGUSR1) */
 		FALSE;
 }
 
 /**
- * Send a signal to the calling process or thread. If "-Xrs"
- * commandline option is specified, then signals such as SIGQUIT,
- * SIGHUP, SIGINT and SIGTERM are ignored.
+ * Check if a signal is ignored by using sigaction.
+ *
+ * @param sigNum Integer value of the signal
+ *
+ * @returns TRUE if the signal is ignored
+ *          FALSE if the signal is not ignored
+ */
+static BOOLEAN
+isSignalIgnoredByOS(jint sigNum) {
+	BOOLEAN signalIgnored = FALSE;
+
+#if !defined(WIN32)
+	void *oldHandler = NULL;
+
+	struct sigaction oldSignalAction;
+	memset(&oldSignalAction, 0, sizeof(struct sigaction));
+
+	OMRSIG_SIGACTION(sigNum, NULL, &oldSignalAction);
+	oldHandler = (void *)oldSignalAction.sa_sigaction;
+	if (NULL == oldHandler) {
+		oldHandler = (void *)oldSignalAction.sa_handler;
+	}
+
+	if (oldHandler == (void *)SIG_IGN) {
+		signalIgnored = TRUE;
+	}
+#endif /* !defined(WIN32) */
+
+	return signalIgnored;
+}
+
+/**
+ * Raise a signal to the calling process or thread.
+ *
+ * isSignalReservedByJVM function lists the signals reserved by the JVM.
+ * Do not raise a signal if it is reserved by the JVM and not registered
+ * via JVM_RegisterSignal.
+ *
+ * isSignalUsedForShutdown lists all signals characterized as shutdown signals.
+ * Do not raise a shutdown signal if the -Xrs or the -Xrs:async cmdnline option
+ * is specified. If a shutdown signal is ignored by the OS, do not raise that
+ * shutdown signal. Only raise a shutdown signal if the previous two conditions
+ * are false.
  *
  * @param sigNum Integer value of the signal to be sent to the
  *               calling process or thread
@@ -4474,12 +4557,20 @@ JVM_RaiseSignal(jint sigNum)
 {
 	jboolean rc = JNI_FALSE;
 	J9JavaVM *javaVM = (J9JavaVM *)BFUjavaVM;
+	BOOLEAN isShutdownSignal = isSignalUsedForShutdown(sigNum);
 
 	Trc_SC_RaiseSignal_Entry(sigNum);
 
-	if (J9_ARE_ALL_BITS_SET(javaVM->sigFlags, J9_SIG_XRS_SYNC)
-			&& isSignalSpecial(sigNum)) {
-		/* Ignore signal */
+	if (isSignalReservedByJVM(sigNum)) {
+		/* Don't raise a signal if it is reserved by the JVM, and not
+		 * registered via JVM_RegisterSignal.
+		 */
+	} else if (J9_ARE_ALL_BITS_SET(javaVM->sigFlags, J9_SIG_XRS_ASYNC) && isShutdownSignal) {
+		/* Ignore shutdown signals if -Xrs or -Xrs:async is specified.
+		 * If -Xrs:sync is specified, then raise shutdown signals.
+		 */
+	} else if (isShutdownSignal && isSignalIgnoredByOS(sigNum)) {
+		/* Ignore shutdown signal if it is ignored. */
 	} else {
 		raise(sigNum);
 		rc = JNI_TRUE;
@@ -4491,34 +4582,24 @@ JVM_RaiseSignal(jint sigNum)
 }
 
 /**
- * This is a stub for the pre-defined handler. The pre-defined
- * handler is supposed to be used in JVM_RegisterSignal when
- * the special value of J9_PRE_DEFINED_HANDLER_CHECK (2) is specified
- * in the handler. It hasn't been implemented since its functionality
- * is not known.
+ * Register a Java SignalHandler for a signal.
  *
- * TODO: Implement the pre-defined handler.
+ * isSignalReservedByJVM function lists the signals reserved by the JVM.
+ * This function shouldn't override handlers for signals that are reserved
+ * by the JVM. A user is allowed to register a native handler for the
+ * reserved signals if -Xrs cmdline option is specified. This function
+ * shouldn't register a Java SignalHandler for the reserved signals.
  *
- * @param sigNum Integer value of the signal to be sent to the
- *               calling process or thread
+ * isSignalUsedForShutdown lists all signals characterized as shutdown signals.
+ * This function shouldn't register a handler for shutdown signals if
+ * the -Xrs or the -Xrs:async cmdnline option is specified. If a shutdown
+ * signal is ignored by the OS, this function shouldn't register a handler
+ * for that shutdown signal. Only register a handler for shutdown signals if
+ * the previous two conditions are false.
  *
- * @returns void
- */
-static void
-dummySignalHandler(int sigNum) {
-
-}
-
-/**
- * Register a signal handler for a signal. Signals such as SIGFPE,
- * SIGILL and SIGSEGV are used by the VM; so, we don't register
- * a signal handler for these signals. If "-Xrs" commandline option
- * is specified, then signals such as SIGQUIT, SIGHUP, SIGINT
- * and SIGTERM are also ignored; thus, we don't register a signal
- * handler for these signals. If handler has the special value of
- * J9_PRE_DEFINED_HANDLER_CHECK (2), then handler is changed to
- * dummySignalHandler before it is registered. If the old handler
- * is same as the new handler, then a special value,
+ * If handler has the special value of J9_PRE_DEFINED_HANDLER_CHECK (2),
+ * then the predefinedHandlerWrapper is registered with asynchSignalReporterThread
+ * in OMR. If the old handler is same as the new handler, then a special value,
  * J9_OLDHANDLER_SAME_AS_NEWHANDLER (2) is returned.
  *
  * @param sigNum Integer value of the signal to be sent to the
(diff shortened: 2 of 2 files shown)


runtime/j9vm/jvm.c
```diff
@@ -4474,11 +4504,18 @@ JVM_RaiseSignal(jint sigNum)
...
-	if (J9_ARE_ALL_BITS_SET(javaVM->sigFlags, J9_SIG_XRS)
-			&& isSignalSpecial(sigNum)) {
+	if ((J9_ARE_ANY_BITS_SET(javaVM->sigFlags, J9_SIG_XRS | J9_SIG_XRS_ASYNC)
+			&& isShutdownSignal)
+#if defined(SIGQUIT)
+			|| (SIGQUIT == sigNum)
+#endif /* defined(SIGQUIT) */
+	) {
+		/* Ignore signal */
+	} else if (isShutdownSignal && isSignalIgnored(sigNum)) {
```
pshipton: Why only check for shutdown signals?

runtime/j9vm/jvm.c
```diff
@@ -4474,11 +4504,18 @@ JVM_RaiseSignal(jint sigNum)
 {
 	jboolean rc = JNI_FALSE;
 	J9JavaVM *javaVM = (J9JavaVM *)BFUjavaVM;
+	BOOLEAN isShutdownSignal = isSignalUsedForShutdown(sigNum);
 
 	Trc_SC_RaiseSignal_Entry(sigNum);
 
-	if (J9_ARE_ALL_BITS_SET(javaVM->sigFlags, J9_SIG_XRS)
-			&& isSignalSpecial(sigNum)) {
+	if ((J9_ARE_ANY_BITS_SET(javaVM->sigFlags, J9_SIG_XRS | J9_SIG_XRS_ASYNC)
```
pshipton: -Xrs:sync is supposed to have the JVM continue to handle SIG_TERM, which is one of the shutdown signals.

babsingh: `-Xrs:sync` seems to be an OpenJ9 only option. should the bit-check above exclude the `J9_SIG_XRS_ASYNC` flag?

runtime/j9vm/jvm.c
babsingh: this requirement was part of JCL team's specification

runtime/j9vm/jvm.c
babsingh: it isn't `SIGTERM`. `SIGQUIT` should be allowed with `-Xrs:sync`. `JVM_RegisterSignal` shouldn't prevent the JVM to register a handler for `SIGQUIT` via `stdinit.c::standardInit -> J9SigQuitStartup`. `JVM_RegisterSignal` will only prevent a user to register a signal via `sun.misc.Signal.register`.

pshipton: https://www.ibm.com/support/knowledgecenter/en/SSYKE2_8.0.0/com.ibm.java.lnx.80.doc/diag/appendixes/cmdline/Xrs.html mentions both SIGQUIT and SIGTERM

babsingh: i account for `-Xrs`, `-Xrs:sync` and `-Xrs:async` in `JVM_RegisterSignal` and `JVM_RaiseSignal`.

runtime/j9vm/jvm.c
```diff
@@ -4474,12 +4504,28 @@ JVM_RaiseSignal(jint sigNum)
...
 
-	if (J9_ARE_ALL_BITS_SET(javaVM->sigFlags, J9_SIG_XRS)
-			&& isSignalSpecial(sigNum)) {
-		/* Ignore signal */
+	if (isSignalUsedByVM(sigNum)) {
+		/* JVM_RaiseSignal shouldn't raise synchronous signals such as SIGFPE,
+		 * SIGILL and SIGSEGV. Based upon the spec, it should also not raise SIGQUIT.
+		 */
+	} else if (J9_ARE_ALL_BITS_SET(javaVM->sigFlags, J9_SIG_XRS_ASYNC) && isShutdownSignal) {
+		/* Ignore all shutdown signals if -Xrs or -Xrs:async is specified. */
```
pshipton: None of the code is checking J9_SIG_XRS any more, which doesn't match the comments.

pshipton: Hmm, just noticed that J9_SIG_XRS and J9_SIG_XRS_SYNC have the same value.

pshipton: I wonder if https://www.ibm.com/support/knowledgecenter/en/SSYKE2_8.0.0/com.ibm.java.lnx.80.doc/diag/appendixes/cmdline/Xrs.html meant to document -Xrs:async and not -Xrs:sync.

pshipton: Note there is an explicit J9_SIG_NO_SIG_QUIT as well.
```
		if (optionValue && 0 == strcmp(optionValue, "sync")) {
			vm->sigFlags |= J9_SIG_XRS_SYNC;
			sigOptions |= J9PORT_SIG_OPTIONS_REDUCED_SIGNALS_SYNCHRONOUS;
		} else if (optionValue && 0 == strcmp(optionValue, "async")) {
			vm->sigFlags |= (J9_SIG_XRS_ASYNC | J9_SIG_NO_SIG_QUIT);
			sigOptions |= J9PORT_SIG_OPTIONS_REDUCED_SIGNALS_ASYNCHRONOUS;
		} else {
			vm->sigFlags |= (J9_SIG_XRS_SYNC | J9_SIG_XRS_ASYNC | J9_SIG_NO_SIG_QUIT);
			sigOptions |= (J9PORT_SIG_OPTIONS_REDUCED_SIGNALS_SYNCHRONOUS | J9PORT_SIG_OPTIONS_REDUCED_SIGNALS_ASYNCHRONOUS);
		}
```

pshipton: As an aside, I think we should change the one piece of JIT code that uses J9_SIG_XRS and then delete this duplicate flag.

babsingh: according to `SunMiscSignalTest`, `JVM_RegisterSignal` and `JVM_RaiseSignal` should never register or raise `SIGQUIT`. JVM should manage `SIGQUIT` separately via `J9SigQuitStartup` and `J9SigQuitShutdown`.

I will open a pull request for removal of the duplicate flag.

babsingh:
@pshipton The below documentation lists the signals, which are used by the JVM. We shouldn't allow a user to register a handler for a signal via `JVM_RegsiterSignal` if the signal is used by the JVM. I have updated `isSignalReservedByJVM` to ignore the following signals: `SIGFPE`, `SIGILL`, `SIGSEGV`, `SIGBUS`,  `SIGTRAP`, `SIGQUIT`, `SIGBREAK`, `SIGABRT` and `SIGCHLD`. 

We don't have to worry about signals between `SIGRTMIN` (32) and `SIGRTMAX` (64). [Signal.init](https://github.com/ibmruntimes/openj9-openjdk-jdk9/blob/openj9/jdk/src/java.base/share/classes/jdk/internal/misc/Signal.java#L138) fails with `Support missing for ...: Unknown signal: ...` for signals between `SIGRTMIN` (32) and `SIGRTMAX` (64). So, `JVM_*Signal` functions will never need to handle these signals.

Only three shutdown signals remain: `SIGINT`, `SIGTERM` and `SIGHUP`. `JVM_RegisterSignal` is used to register a handler for these signals via [Terminator.setup()](https://github.com/ibmruntimes/openj9-openjdk-jdk9/blob/openj9/jdk/src/java.base/unix/classes/java/lang/Terminator.java#L47). `JVM_RegisterSignal` and `JVM_RaiseSignal` ignore these signals if `-Xrs` or `-Xrs:async` is specified. If `-Xrs:sync` is specified, then shutdown signals aren't ignored.

Pull request has been updated. All signals are ignored or handled appropriately based upon the below documentation.

IBM Signal Documentation:
1) [Windows](https://www.ibm.com/support/knowledgecenter/SSYKE2_8.0.0/com.ibm.java.win.80.doc/user/sighand.html)
2) [Linux](https://www.ibm.com/support/knowledgecenter/SSYKE2_8.0.0/com.ibm.java.lnx.80.doc/user/sighand.html)

pshipton:
> So, JVM_*Signal functions will never need to handle these signals [between SIGRTMIN (32) and SIGRTMAX (64)].

Code can change (and without notice to/from OpenJ9). It wouldn't be hard to add the restriction for the signal the VM is using, which may be useful in the future.

Please check the other platforms as well, [AIX](https://www.ibm.com/support/knowledgecenter/SSYKE2_8.0.0/com.ibm.java.aix.80.doc/user/sighand.html), [z/OS](https://www.ibm.com/support/knowledgecenter/SSYKE2_8.0.0/com.ibm.java.zos.80.doc/user/sighand.html), there is at least one additional signal listed SIGRECONFIG .

babsingh:
On AIX and zOS, `SIGRECONFIG` is ignored. On zOS, `SIGUSR1` is ignored. `SIGRTMIN` and `SIGRTMAX` also need to be ignored. I have updated `isSignalReservedByJVM` to ignore these signals.

On AIX, there is a reserved signal without a name: `No Name (40)`. This signal isn't included in `jvm.c::signalMap`. So, an `Signal` instance won't be created for this signal. `JVM_RegisterSignal` and `JVM_FindSignal` can't be invoked for this signal.

`isSignalReservedByJVM` only needs to ignore signals which are part of `jvm.c::signalMap`. If a signal is not included in `jvm.c::signalMap`, then `JVM_RegisterSignal` and `JVM_FindSignal` can't be invoked for that signal.

pshipton:
In that case, why put signals which aren't affected by -Xrs in the signalMap: SIGCHLD, SIGRTMIN, SIGRTMAX, SIGCHLD ?

runtime/j9vm/jvm.c
```diff
@@ -4544,34 +4633,42 @@ JVM_RegisterSignal(jint sigNum, void* handler)
...
 
-	if (isSignalUsedByVM(sigNum)) {
-		/* Don't allow user to register a native handler since
-		 * the signal is already used by the VM.
-		 */
-	} else if (J9_ARE_NO_BITS_SET(javaVM->sigFlags, J9_SIG_XRS_SYNC)
-			&& isSignalSpecial(sigNum)) {
-		/* Don't allow user to register a native handler since
-		 * the signal is already used by the VM.
+	if (isSignalReservedByJVM(sigNum)) {
```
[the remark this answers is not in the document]
babsingh: `JVM_RegisterSignal` registers a Java `SignalHandler` with a signal. When the reserved signals are unreserved by specifying the `-Xrs` option, the spec says that the user can register their own `native handler`. `JVM_RegisterSignal` doesn't allow a user to register a `_native_handler`; it registers a Java `SignalHandler`. So, we shouldn't let a user register Java `SignalHandlers` via `JVM_RegisterSignal` with the reserved signals ... when the reserved signals are unreserved by specifying `-Xrs` option. The users should use a native approach to register a `native handler` for the reserved signals. @pshipton Do you agree with my interpretation?

babsingh:
> In that case, why put signals which aren't affected by -Xrs in the signalMap: SIGCHLD, SIGRTMIN, SIGRTMAX, SIGCHLD ?

Since there is no spec for `sun.misc.Signal`, we can take either approach. We can either include or exclude `SIGCHLD`, `SIGRTMIN`, `SIGRTMAX` and `SIGCHLD` from `jvm.c:signalMap`. If we include these signals in `jvm.c:signalMap`, we should be able to successfully create a `Signal` instance for these signals. If we exclude these signals, then [Signal.init](https://github.com/ibmruntimes/openj9-openjdk-jdk9/blob/openj9/jdk/src/java.base/share/classes/jdk/internal/misc/Signal.java#L138) will throw an IAE. Inclusion shows that these signals are supported by the JVM. But, APIs such as `Signal.handle` (calls `JVM_RegisterSignal`) and `Signal.raise` (calls `JVM_RaiseSignal`) will not work for signals, which are reserved by the JVM.

runtime/j9vm/jvm.c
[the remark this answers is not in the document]
pshipton: Makes sense. How about adding this information to the JVM_RegisterSignal doc.

The whole implementation seems overly complicated for something that seems to only need to handle 3 signals: HUP, INT, TERM.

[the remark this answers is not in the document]
babsingh: We need to support more than four signals. Referring to [SunMiscSignalTest](https://github.ibm.com/runtimes/openj9-openjdk-jdk9/blob/ibm_sdk/jdk/test/sun/misc/SunMiscSignalTest.java#L101). You should see a 2d array. The fist column is signal name, the second column determines if that signal is supported, the third column determines if we can register a handler via `JVM_RegisterSignal`, the fourth column determines if that signal can be raised via `JVM_RaiseSignal` and the fifth column determines if the associated handler should be invoked.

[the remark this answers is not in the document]
pshipton: This implementation doesn't seem to match the test suite, which I believe shows that (for example) SIGBUS can be registered, raised, and the java handler invoked (unless running with -Xrs).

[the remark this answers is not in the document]
babsingh: It won't. List of reserved signals will vary between Hotspot and OpenJ9. In OpenJ9, `SIGBUS` is reserved by the JVM. It is handled as a synchronous signal via `omrsig_protect`.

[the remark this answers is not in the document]
pshipton: Since OpenJ9 is supporting synchronous signals differently then we should consider simplifying this implementation to the minimum required to support the asynchronous signals.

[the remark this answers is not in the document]
babsingh: Updated `JVM_RegisterSignal` and `JVM_RaiseSignal` doc to match the current implementation.

Currently, `JVM_*Signal` functions only support asynchronous signals. Synchronous signals are ignored or an error is returned for them (`...: Signal already used by VM or OS`). OpenJ9's error handling won't match Hotspot's error handling if we exclude synchronous signals. There won't be much simplification; only the list of reserved signals will reduce in `isSignalReservedByJVM()` and we may get rid of the five synchronous signals in `jvm.c::signalMap`.

babsingh:
I thought pushing code changes incrementally will be easier and smoother. The cross-project (J9/OMR) signal support implementation is highly vulnerable to failure. I will open a uber pull request when I have a fully functioning (J9/OMR) solution for signal support.

babsingh:
Based on Dan's recommendation, I will proceed with incremental development.

babsingh:
This pull request has been updated to accommodate new OMR and OpenJ9 changes.

This pull request depends on eclipse/omr#2494 and https://github.com/eclipse/openj9/pull/1603.

runtime/j9vm/j9scar.tdf
```diff
@@ -249,7 +249,7 @@ TraceExit=Trc_SC_Recv_Exit NoEnv Overhead=1 Level=1 Template="JVM_Recv -- return
 TraceEntry=Trc_SC_RecvFrom_Entry NoEnv Overhead=1 Level=1 Template="JVM_RecvFrom(descriptor=%d, buffer=%p, length=%d, flags=%d, fromAddr=%p, fromLength=%p)"
 TraceExit=Trc_SC_RecvFrom_Exit NoEnv Overhead=1 Level=1 Template="JVM_RecvFrom -- return %d, *length=%d"
 
-TraceEvent=Trc_SC_RegisterSignal NoEnv Overhead=1 Level=1 Template="JVM_RegisterSignal (not implemented -- returning -1)"
+TraceEvent=Trc_SC_RegisterSignal NoEnv Overhead=1 Level=1 Template="JVM_RegisterSignal(signal=%d, handler=%p)"
```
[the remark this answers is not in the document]
babsingh: Added a new TraceEntry for JVM_RegisterSignal

runtime/j9vm/j9scar.tdf
```diff
@@ -334,3 +334,7 @@ TraceExit=Trc_SC_GCNoCompact_Exit Overhead=1 Level=1 Template="JVM_GCNoCompact()
 TraceEntry=Trc_SC_LoadSystemLibrary_Entry NoEnv Overhead=1 Level=1 Template="JVM_LoadSystemLibrary(name=%s)"
 TraceExit=Trc_SC_LoadSystemLibrary_Exit NoEnv Overhead=1 Level=1 Template="JVM_LoadSystemLibrary -- return 0x%zx"
 TraceEvent=Trc_SC_LoadSystemLibrary_LoadFailed NoEnv Overhead=1 Level=1 Template="JVM_LoadSystemLibrary(name=%s) failed to load"
+
+TraceEvent=Trc_SC_RegisterSignal_FailedToRegisterHandler NoEnv Overhead=1 Level=1 Template="Failed to register pre-defined handler: OS signal value = %d, new signal handler = %p, old signal handler = %p"
+TraceExit=Trc_SC_RegisterSignal_Exit NoEnv Overhead=1 Level=1 Template="JVM_RegisterSignal -- return old OS handler = %p"
```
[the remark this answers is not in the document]
babsingh: Updated - the new TraceEntry for JVM_RegisterSignal matches the TraceExit for JVM_RegisterSignal

runtime/j9vm/j9scar.tdf
```diff
@@ -334,3 +334,7 @@ TraceExit=Trc_SC_GCNoCompact_Exit Overhead=1 Level=1 Template="JVM_GCNoCompact()
 TraceEntry=Trc_SC_LoadSystemLibrary_Entry NoEnv Overhead=1 Level=1 Template="JVM_LoadSystemLibrary(name=%s)"
 TraceExit=Trc_SC_LoadSystemLibrary_Exit NoEnv Overhead=1 Level=1 Template="JVM_LoadSystemLibrary -- return 0x%zx"
 TraceEvent=Trc_SC_LoadSystemLibrary_LoadFailed NoEnv Overhead=1 Level=1 Template="JVM_LoadSystemLibrary(name=%s) failed to load"
+
+TraceEvent=Trc_SC_RegisterSignal_FailedToRegisterHandler NoEnv Overhead=1 Level=1 Template="Failed to register pre-defined handler: OS signal value = %d, new signal handler = %p, old signal handler = %p"
```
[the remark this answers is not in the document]
babsingh: It will be executed for platforms where OMR support is missing. So, we should stick with TraceEvent.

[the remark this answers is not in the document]
pshipton: It still indicates a problem, on those platforms the signal handling won't work. We'll be fixing these platforms so they all work in the end. I still think it should be TraceException.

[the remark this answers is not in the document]
babsingh: fixed - changed to TraceException

babsingh:
@pshipton The required OMR changes for this PR will be merged soon. Would you like me to squash the commits in this PR?

pshipton:
@babsingh please go ahead

pshipton:
@babsingh #1603 has been merged. Is this ready for final review and merge?

babsingh:
@pshipton after https://github.com/eclipse/openj9/pull/1603 was merged, I have again rebased and pushed changes to this PR.

runtime/j9vm/j9scar.tdf
```diff
@@ -334,3 +334,7 @@ TraceExit=Trc_SC_GCNoCompact_Exit Overhead=1 Level=1 Template="JVM_GCNoCompact()
 TraceEntry=Trc_SC_LoadSystemLibrary_Entry NoEnv Overhead=1 Level=1 Template="JVM_LoadSystemLibrary(name=%s)"
 TraceExit=Trc_SC_LoadSystemLibrary_Exit NoEnv Overhead=1 Level=1 Template="JVM_LoadSystemLibrary -- return 0x%zx"
 TraceEvent=Trc_SC_LoadSystemLibrary_LoadFailed NoEnv Overhead=1 Level=1 Template="JVM_LoadSystemLibrary(name=%s) failed to load"
+
+TraceEntry=Trc_SC_RegisterSignal_Entry NoEnv Overhead=1 Level=1 Template="JVM_RegisterSignal(signal=%d, handler=%p)"
```
DanHeidinga: `NoEnv` tracepoints have to attempt to lookup the current J9VMThread to store into the thread's tracebuffer.  Since every invocation of this function will need to do lookup twice (once for enter/exit), can you make these take the env?

runtime/j9vm/jvm.c
```diff
@@ -4542,54 +4616,64 @@ JVM_RegisterSignal(jint sigNum, void* handler)
...
+		/* Return error if signal is reserved by the JVM. */
+		return (void *)J9_SIG_ERR;
```
DanHeidinga: This return is missing a TraceExit

runtime/j9vm/jvm.c
```diff
@@ -4542,54 +4616,64 @@ JVM_RegisterSignal(jint sigNum, void* handler)
...
+	} else if (J9_ARE_ANY_BITS_SET(javaVM->sigFlags, J9_SIG_XRS_ASYNC) && isShutdownSignal) {
+		/* Don't register a handler for shutdown signals if -Xrs or -Xrs:async is specified.
+		 * If -Xrs:sync is specified, then register a handler for shutdown signals.
 		 */
+		return (void *)J9_SIG_ERR;
```
DanHeidinga: This return is missing a TraceExit

runtime/j9vm/jvm.c
```diff
@@ -4542,54 +4616,64 @@ JVM_RegisterSignal(jint sigNum, void* handler)
...
+	} else if (isShutdownSignal && isSignalIgnoredByOS(sigNum)) {
+		/* Ignore shutdown signal if it is ignored. */
+		return (void *)J9_SIG_IGNORED;
```
DanHeidinga: This return is missing a TraceExit

runtime/j9vm/jvm.c
```diff
@@ -4542,54 +4616,64 @@ JVM_RegisterSignal(jint sigNum, void* handler)
...
 		} else {
-			newSignalAction.sa_handler = (void (*)(int))handler;
-		}
-		OMRSIG_SIGACTION(sigNum, &newSignalAction, &oldSignalAction);
+#if defined(WIN32)
+			oldHandler = (void *)OMRSIG_SIGNAL(sigNum, handler);
+#else /* defined(WIN32) */
+			/* Don't block any signals. */
+			if (0 != sigemptyset(&newSignalAction.sa_mask)) {
+				return (void *)J9_SIG_ERR;
```
DanHeidinga: This return is missing a TraceExit

runtime/j9vm/j9scar.tdf
babsingh: `JVM_RegisterSignal` doesn't get access to `JNIEnv *env`. Used `javaVM->mainThread` as `env`. Is `javaVM->mainThread` a satisfactory alternative for `env`?

runtime/j9vm/jvm.c
babsingh: added an `exit` label. `oldHandler` is initialized to `J9_SIG_ERR` so we can `goto exit` and return `oldHandler`.

runtime/j9vm/jvm.c
babsingh: added an `exit` label. `oldHandler` is initialized to `J9_SIG_ERR` so we can `goto exit` and return `oldHandler`.

runtime/j9vm/jvm.c
babsingh: added an `exit` label. `oldHandler` is changed to `J9_SIG_IGNORED`. then, we `goto exit` and return `oldHandler`.

runtime/j9vm/jvm.c
babsingh: fixed. added an `exit` label. `oldHandler` is initialized to `J9_SIG_ERR` so we can `goto exit` and return `oldHandler`.

runtime/j9vm/j9scar.tdf
babsingh: `javaVM->mainThread` won't be a satisfactory alternative for `env`. `JVM_RegisterSignal` may not always be invoked from `javaVM->mainThread`. in such scenarios, `env` will be set incorrectly. so, we have to go with the `NoEnv` approach for `JVM_RegisterSignal` tracepoints since `JVM_RegisterSignal` doesn't get access to `JNIEnv *env`. @DanHeidinga am i correct?

pshipton: You should be able to use `javaVM->internalVMFunctions->currentVMThread(javaVM)`

babsingh: updated tracepoints to take `env` as input.

runtime/j9vm/jvm.c
```diff
@@ -4425,25 +4433,32 @@ isSignalSpecial(jint sigNum)
...
+ *
+ * Synchronous signals are handled by the VM via omrsig_protect: SIGFPE,
+ * SIGILL, SIGSEGV, SIGBUS and SIGTRAP. On Windows, SIGBREAK is reserved
+ * by the VM to trigger dumps. On Unix/Linux platforms, SIGQUIT is
+ * reserved by the VM to trigger dumps. SIGABRT is reserved by the JVM
+ * for abnormal termination. SIGCHLD is reserved for internal control.
+ * On zOS, SIGUSR1 is reserved by the JVM. SIGRECONFIG is reserved to
+ * detect any change in the number of CPUs, processing capacity, or
+ * physical memory. This function only needs to ignore signals included
+ * in jvm.c::signalMap.
```
DanHeidinga: > This function only needs to ignore signals included in jvm.c::signalMap.

@babsingh I'm not sure what this comment means.  Does ignore mean return false?

babsingh: i will update the comment. i meant to say the following:

*if the `signal` is listed in `jvm.c::signalMap`, then only `isSignalReservedByJVM` needs to determine if the `signal` is used/reserved by the JVM. if the signal is not listed in `jvm.c::signalMap`, then a `sun.misc.Signal` instance can't be created for a `signal` value. The native `JVM_*Signal` functions can't be invoked in the absence of a `sun.misc.Signal` instance.*

babsingh: comment has been updated.

DanHeidinga: Thanks.  That's much clearer

runtime/j9vm/jvm.c
```diff
@@ -4474,12 +4557,20 @@ JVM_RaiseSignal(jint sigNum)
 {
 	jboolean rc = JNI_FALSE;
 	J9JavaVM *javaVM = (J9JavaVM *)BFUjavaVM;
+	BOOLEAN isShutdownSignal = isSignalUsedForShutdown(sigNum);
```
pshipton: Is isSignalUsedForShutdown() misnamed? Should it be called isAsyncSignal() or similar?

pshipton:
Just a minor comment about the proper name for isSignalUsedForShutdown(), which can be updated afterwards if necessary. I'll start PR testing.

jenkins test all all jdk8,jdk10

runtime/j9vm/jvm.c
babsingh: this function is supposed to determine if a signal is used by the JVM for shutdown. Referring to [Terminator.java Unix](https://github.com/ibmruntimes/openj9-openjdk-jdk9/blob/openj9/jdk/src/java.base/unix/classes/java/lang/Terminator.java#L59) and [Terminator.java Windows](https://github.com/ibmruntimes/openj9-openjdk-jdk9/blob/openj9/jdk/src/java.base/windows/classes/java/lang/Terminator.java#L59), a shutdown handler is registered for the following signals: `SIGINT`, `SIGTERM` and `SIGHUP`. this function checks if a signal is either `SIGINT`, `SIGTERM` or `SIGHUP`. this function is needed because the spec provided by the JCL team puts special constraints on these shutdown signals.

PR #1643, opened 2018-04-09
Back-off using OMR's iabs and labs evaluators

0dvictor:
OMR's iabs and labs evaluators currently do not produce
the value that Java Specs requires on certain cornor case,
backing it off to using J9's implementation.


andrewcraik:
Jenkins test sanity

andrewcraik:
looks like we had some infra issues going to try again - Jenkins test sanity

PR #1652, opened 2018-04-10
Fix return value warning treaded as error

fengxue-IS:


fengxue-IS:
@DanHeidinga found this issue when testing with condy bytecode resolve, not sure how it passed the sanity testing though

DanHeidinga:
Original PR that introduced the breakage is: https://github.com/eclipse/openj9/pull/1426

DanHeidinga:
Will check on this and merge once the build, including travis, has passed.

andrewcraik:
Note this appeared in an OMR acceptance build so once merged firing an OMR acceptance may be an idea just to make sure changes flow in.

gacholio:
I'm merging this preemptively, since it's obviously correct and the code out there fails compiles all over the place, so it can't possibly make things worse.

PR #1654, opened 2018-04-10
Preserved original pointer in getCustomFFIType

fjeremic:
The getIntFromLayout and getStructFFIType APIs take a reference to the
array returned by copyStringToUTF8WithMemAlloc and they modify the
pointer value. Subsequent calls to j9mem_free_memory will fail due to
metadata not being present.

We preserve the original array pointer and pass a temporary to the
aforementioned APIs to avoid these problems.

### DIFF
--- a/runtime/vm/FFITypeHelpers.hpp
+++ b/runtime/vm/FFITypeHelpers.hpp
@@ -112,17 +112,20 @@ public:
 
 		*typeFFI = NULL;
 		UDATA structSize = UDATA_MAX;
-		char bufOnStackLocal[J9VM_LAYOUT_STRING_ON_STACK_LIMIT];
-		char *layout = copyStringToUTF8WithMemAlloc(_currentThread, layoutStringObject, J9_STR_NULL_TERMINATE_RESULT, "", 0, bufOnStackLocal, J9VM_LAYOUT_STRING_ON_STACK_LIMIT, NULL);
+		char layoutBuffer[J9VM_LAYOUT_STRING_ON_STACK_LIMIT];
+		char* layout = copyStringToUTF8WithMemAlloc(_currentThread, layoutStringObject, J9_STR_NULL_TERMINATE_RESULT, "", 0, layoutBuffer, sizeof(layoutBuffer), NULL);
+
+		/* Subsequent calls will modify the contents of this pointer, so preserve the original pointer for j9mem_free_memory */
+		char* layoutTemp = layout;
 
 		if (NULL == layout) {
 			goto doneGetCustomFFIType;
 		}
 
-		structSize = getIntFromLayout(&layout);
-		*typeFFI = getStructFFIType(&layout, false);
+		structSize = getIntFromLayout(&layoutTemp);
+		*typeFFI = getStructFFIType(&layoutTemp, false);
 doneGetCustomFFIType:
-		if (layout != bufOnStackLocal) {
+		if (layout != layoutBuffer) {
 			j9mem_free_memory(layout);
 		}
 		return structSize;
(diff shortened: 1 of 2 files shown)


fjeremic:
@pshipton could I please get a review?

pshipton:
Can you please also fix the problem @dmitripivkine pointed out in OutOfLineINL_java_lang_invoke_NativeMethodHandle_initJ9NativeCalloutDataRef(). After `args = (ffi_type **)j9mem_allocate_memory(sizeof(ffi_type *) * typeCount, OMRMEM_CATEGORY_VM);` the memory should be memset() to zero because if an error occurs before all the entries in the array are initialized, the error handling code at freeAllMemoryThenExit will attempt to free all the pointers, some of which will not be initialized yet.

@tajila fyi

runtime/vm/FFITypeHelpers.hpp
```diff
@@ -112,17 +112,20 @@ class FFITypeHelpers
 
 		*typeFFI = NULL;
 		UDATA structSize = UDATA_MAX;
-		char bufOnStackLocal[J9VM_LAYOUT_STRING_ON_STACK_LIMIT];
-		char *layout = copyStringToUTF8WithMemAlloc(_currentThread, layoutStringObject, J9_STR_NULL_TERMINATE_RESULT, "", 0, bufOnStackLocal, J9VM_LAYOUT_STRING_ON_STACK_LIMIT, NULL);
+		char layoutBuffer[J9VM_LAYOUT_STRING_ON_STACK_LIMIT];
+		char* layout = copyStringToUTF8WithMemAlloc(_currentThread, layoutStringObject, J9_STR_NULL_TERMINATE_RESULT, "", 0, layoutBuffer, sizeof(layoutBuffer), NULL);
+
+		/* Subsequent calls will modify the contents of this pointer, so preserve the original pointer for j9mem_free_memory */
+		char* layoutTemp = layout;
```
pshipton: Technically this could be moved after the null check.

fjeremic: >Technically this could be moved after the null check.
@pshipton this is not possible because of the `goto`, otherwise we get the following error:

>FFITypeHelpers.hpp: In member function 'UDATA FFITypeHelpers::getCustomFFIType(ffi_type**, j9object_t)':
>FFITypeHelpers.hpp:127:1: error: jump to label 'doneGetCustomFFIType' [-fpermissive]
>FFITypeHelpers.hpp:119:9: error:   from here [-fpermissive]
>FFITypeHelpers.hpp:123:9: error:   crosses initialization of 'char* layoutTemp'

pshipton: ok

pshipton:
Alternatively to memset'ing the entire allocation, the logic could be changed to identify how many entries have already been initialized and only free those.

fjeremic:
Fixed in f3a8bee.

> Alternatively to memset'ing the entire allocation, the logic could be changed to identify how many entries have already been initialized and only free those.

 I will delegate this to @tajila as I am totally unfamiliar with this area of the code. But this does sound like a better approach as I'm guessing errors in this area are rare.

runtime/vm/OutOfLineINL_java_lang_invoke_NativeMethodHandle.cpp
```diff
@@ -66,6 +66,12 @@ OutOfLineINL_java_lang_invoke_NativeMethodHandle_initJ9NativeCalloutDataRef(J9VM
 	 */
 	U_32 typeCount = J9INDEXABLEOBJECT_SIZE(currentThread, argTypesObject) + 1;
 	args = (ffi_type **)j9mem_allocate_memory(sizeof(ffi_type *) * typeCount, OMRMEM_CATEGORY_VM);
+
+	/* Zero out the memory because if an error occurs before all the entries in this array are initialized then the error
+	 * handling code at freeAllMemoryThenExit will attempt to free all the pointers, some of which will not be initialized.
+	 */
+	memset(args, 0, sizeof(ffi_type *) * typeCount);
```
pshipton: should happen after the null check

fjeremic: Right. What a silly mistake. Fixed in 736be5e along with the copyright date.

pshipton:
PR builds won't test these changes since they are `#ifdef J9VM_OPT_PANAMA`

PR #1655, opened 2018-04-11
Add isAccessUserOnly0 for FileSystem

theresa-m:
libmanagement natives were replaced by a stub in openj9. isAccessUserOnly0 is required for the security enabled SVT JLM tests to pass on Java 8.

Depends on: https://github.com/eclipse/omr/pull/2536

Fixes: #1520

Test PR: #1656

Signed-off-by: Theresa Mammarella <Theresa.T.Mammarella@ibm.com>

PR #1656, opened 2018-04-11
Tests for isAccessUserOnly0 native

theresa-m:
- includes tests for isAccessUserOnly0 in Java8 and Java9
- expanded test infra to have src_90_current and src_90_latest files for different JCL versions in Java 9

Relies on: #1655 

### DIFF
--- a/test/functional/TestUtilities/src/org/openj9/test/jdk/internal/agent/Test_FileSystem_Base.java
+++ b/test/functional/TestUtilities/src/org/openj9/test/jdk/internal/agent/Test_FileSystem_Base.java
@@ -1,0 +1,71 @@
+package org.openj9.test.jdk.internal.agent;
+
+/*******************************************************************************
+ * Copyright (c) 2018, 2018 IBM Corp. and others
+ *
+ * This program and the accompanying materials are made available under
+ * the terms of the Eclipse Public License 2.0 which accompanies this
+ * distribution and is available at https://www.eclipse.org/legal/epl-2.0/
+ * or the Apache License, Version 2.0 which accompanies this distribution and
+ * is available at https://www.apache.org/licenses/LICENSE-2.0.
+ *
+ * This Source Code may also be made available under the following
+ * Secondary Licenses when the conditions for such availability set
+ * forth in the Eclipse Public License, v. 2.0 are satisfied: GNU
+ * General Public License, version 2 with the GNU Classpath
+ * Exception [1] and GNU General Public License, version 2 with the
+ * OpenJDK Assembly Exception [2].
+ *
+ * [1] https://www.gnu.org/software/classpath/license.html
+ * [2] http://openjdk.java.net/legal/assembly-exception.html
+ *
+ * SPDX-License-Identifier: EPL-2.0 OR Apache-2.0 OR GPL-2.0 WITH Classpath-exception-2.0 OR LicenseRef-GPL-2.0 WITH Assembly-exception
+ *******************************************************************************/
+
+import org.testng.annotations.Test;
+import org.testng.Assert;
+import org.testng.log4testng.Logger;
+import java.io.File;
+
+@Test(groups = { "level.sanity" })
+public abstract class Test_FileSystem_Base {
+	protected Logger logger = Logger.getLogger(getClass());
+
+	protected File testSetup() {
+		File file = null;
+		
+		/* method is not supported on windows */
+		if (isWindows()) {
+			logger.debug("skipping  test on Windows");
+		} else {
+		
+			try {
+				file = File.createTempFile("test", null);
+				file.deleteOnExit();
+			} catch (Exception e) {
+				logger.debug("Exception thrown in creating file.");
+				Assert.fail();
+			}
+		}
+		
+		return file;
+	}
+	
+	protected boolean isWindows() {
+		String osName = System.getProperty("os.name");
+		if ((null != osName) && osName.startsWith("Windows")) {
+			return true;
+		}
+		return false;
+	}
+	
+	protected void resetPermissions(File file) {
+		file.setReadable(false, false);
+		file.setWritable(false, false);
+	}
+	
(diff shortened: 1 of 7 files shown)


test/Java8andUp/src_80/org/openj9/test/jdk/internal/agent/Test_FileSystem.java
```diff
@@ -0,0 +1,130 @@
...
+	 */
+	public static void test_fileHasUserAccessOnly() {
+		/* method is not supported on windows */
+		if (isWindows()) {
+			logger.debug("skipping  test_fileHasUserAccessOnly on Windows");
+			return;
+		}
+		
+		FileSystem fs = FileSystem.open();
+		File file = new File("test-file-name");
```
theresa-m: @smlambert concerns with this file created for testing? I've made sure to delete it at the end of each test.

smlambert: Creating a file is not an issue (as you clean it up after).  There are some other test changes that we will ask for on this PR, I've added @llxia for a review.

test/Java8andUp/playlist.xml
```diff
@@ -1843,5 +1843,46 @@
 			<subset>SE90</subset>
 		</subsets>
 	</test>
+	
+	<test>
+		<testCaseName>FileSystem-isAccessUserOnlyTests_SE90</testCaseName>
```
smlambert: For tests that should apply to Java9 and Up, its preferred to call the Java8 tests _SE80, and then drop the suffix, so the same test can be used for 9, 10, 11 and so on.

smlambert:
I have not done a diff between your Test_FileSystem.java code for src_80, src_90_latest, and src_90_current, but seems like a good opportunity to reuse some code (see [examples of reuse](https://github.com/eclipse/openj9/tree/master/test/TestUtilities)).

Another question, perhaps for @DanHeidinga, is it necessary to add tests for src_90_current, does this refer to b148 class libraries, and when can we stop adding tests to continue to support that JCL version)?

DanHeidinga:
@theresa-m Have you seen some of the createTempFile api calls like: https://docs.oracle.com/javase/9/docs/api/java/io/File.html#createTempFile-java.lang.String-java.lang.String-java.io.File-  

Using those calls avoids having to name the file and ensures that the tests can be run in parallel as the file(s) will be unique.

The code explicitly deleting the file is good and to cover all the cases, please also call https://docs.oracle.com/javase/9/docs/api/java/io/File.html#deleteOnExit-- on the files.

DanHeidinga:
> Another question, perhaps for @DanHeidinga, is it necessary to add tests for src_90_current, does this refer to b148 class libraries, and when can we stop adding tests to continue to support that JCL version)?

We should not be adding new tests to support b148.  That's an IBM-ism

theresa-m:
Yes reusing code for this will be great. The files are exactly the same except for the package name of the class they are testing. Thanks for the link @smlambert

theresa-m:
The addition of the current/latest folders for Java 9 was mainly because of internal test compile failures. The FileSystem class being tested changed locations between jcl versions so my test would not work on both jcls. Are there some better ways this could work?

test/Java8andUp/build.xml
```diff
@@ -149,28 +151,59 @@
...
+							<classpath>
+								<pathelement location="../TestConfig/lib/asm-all.jar" />
+								<pathelement location="../TestConfig/lib/testng.jar" />
+								<pathelement location="../TestConfig/lib/jcommander.jar" />
+								<pathelement location="../TestConfig/lib/commons-cli.jar" />
+								<pathelement location="../TestConfig/lib/javassist.jar" />
+							</classpath>
+						</javac>
+					</else>
+				</if>
```
llxia: Instead of using if else, could we use variables to avoid code duplication? The differences between if and else are the following:
 - `${src_90_current}` and `${src_90_latest}`
 -  `--add-exports java.management/sun.management=ALL-UNNAMED` and `--add-exports jdk.management.agent/jdk.internal.agent=ALL-UNNAMED`

theresa-m: good idea. I just pushed this update

llxia:
Jenkins test sanity plinux jdk9,jdk10

theresa-m:
Marking as WIP so this does not get merged before #1655

theresa-m:
@llxia the dependencies for this pr have been merged

theresa-m:
rebasing as this is a while behind.

pshipton:
Jenkins test sanity plinux jdk9,jdk10

pshipton:

theresa-m:
Jdk10 test failure is fixed pending #1942

pshipton:
@llxia @smlambert ready for another look

test/functional/TestUtilities/src/org/openj9/test/jdk/internal/agent/Test_FileSystem_Base.java
```diff
@@ -0,0 +1,71 @@
...
+	}
+	
+	protected void resetPermissions(File file) {
+		file.setReadable(false, false);
+		file.setWritable(false, false);
+	}
+	
+	protected void resolveTest(boolean actualResult, boolean expectedResult) {
+		logger.debug("Test passed: " + (expectedResult == actualResult));
+		Assert.assertEquals(expectedResult, actualResult);
```
llxia: Do we need `logger.debug("Test passed: " + (expectedResult == actualResult))`? Does `Assert.assertEquals(expectedResult, actualResult)` provide the same information?

theresa-m: its not very helpful. I will take that out.

theresa-m:
@pshipton this is ready to go

PR #1673, opened 2018-04-12
Fix shared cache modlevel for Java 10 and 11

### LINKED ISSUE #1288: SharedClassCacheInfo is missing constants for Java 10 & 11
SharedClassCacheInfo is missing constants for Java 10 & 11, as is runtime/include/ibmjvmti.h

These constants are present in runtime/oti/shchelp.h

Before adding the constants we should consider if we really need to have new levels for each Java version, as there will be a new Java version every 6 months. Perhaps only the LTS levels should change the mod level.

@hangshao0 fyi

hangshao0:
1. Remove J2SE_15, J2SE_16, J2SE_17
2. Change J2SE_18 to 0x0800, J2SE_19 to 0x0900, J2SE_V10 to 0x0A00
J2SE_V11 to 0x0B00
3. Starting from Java 10, shared cache modlevel will be set to java
version number
4. Re-enable tests in testSCCMLTests2

Fixes #1288

### DIFF
--- a/runtime/oti/j2sever.h
+++ b/runtime/oti/j2sever.h
@@ -24,20 +24,21 @@
 
 /**
  * Constants for supported J2SE versions.
- * note: J2SE_15 needed for shared classes cache introspection but not supported by JVM.
  */
 /*
  * Note: J2SE_LATEST is the highest Java version supported by VM for a JCL level.
  *       This allows JVM operates with latest version when neither classlib.properties
  *       nor release file presents.
  */
-#define J2SE_15   0x1500
-#define J2SE_16   0x1600
-#define J2SE_17   0x1700
-#define J2SE_18   0x1800
-#define J2SE_19   0x1900
-#define J2SE_V10  0x1A00            /* This refers Java 10 */
-#define J2SE_V11  0x1B00            /* This refers Java 11 */
+#define J2SE_18   0x0800
+#define J2SE_19   0x0900
+#define J2SE_V10  0x0A00            /* This refers Java 10 */
+#define J2SE_V11  0x0B00            /* This refers Java 11 */
+/* Shared class cache is using JAVA_SPEC_VERSION_FROM_J2SE(j2seVersion) to get the Java version.
+ * So bits 9 to 16 of the J2SE constant should match the java version number.
+ */
+
+
 #if JAVA_SPEC_VERSION == 8
 	#define J2SE_LATEST  J2SE_18
 #elif JAVA_SPEC_VERSION == 9
@@ -84,6 +85,7 @@
 #define J2SE_SHAPE_RAW	 		0x90000
 #define J2SE_SHAPE_MASK 		0xF0000
 #define J2SE_SHAPE_SHIFT		16
+#define J2SE_JAVA_SPEC_VERSION_SHIFT 8
 
 /**
  * Masks and constants for describing J2SE shapes.
@@ -111,6 +113,11 @@
  * Macro to extract J2SE shape given a JNIEnv.
  */ 
 #define J2SE_SHAPE_FROM_ENV(env) J2SE_SHAPE(((J9VMThread*)env)->javaVM)
+
+/**
+ * Macro to extract java spec version from J2SE version.
+ */
+#define JAVA_SPEC_VERSION_FROM_J2SE(j2seVersion) 	((j2seVersion) >> J2SE_JAVA_SPEC_VERSION_SHIFT)
 
 #endif     /* j2sever_h */
 
--- a/runtime/shared_common/CacheLifecycleManager.cpp
+++ b/runtime/shared_common/CacheLifecycleManager.cpp
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2001, 2017 IBM Corp. and others
+ * Copyright (c) 2001, 2018 IBM Corp. and others
  *
  * This program and the accompanying materials are made available under
  * the terms of the Eclipse Public License 2.0 which accompanies this
@@ -237,7 +237,6 @@ printSharedCache(void* element, void* param)
 	SH_OSCache_Info* currentItem = (SH_OSCache_Info*) element;
 	if (((state->printCompatibleCache == true) && (currentItem->isCompatible))
 			|| ((state->printIncompatibleCache == true) && (!currentItem->isCompatible))) {
-		char jclLevelStr[10];
 		char addrmodeStr[10];
 
 		PORT_ACCESS_FROM_JAVAVM(state->vm);
@@ -286,9 +285,10 @@ printSharedCache(void* element, void* param)
 			j9tty_printf(PORTLIB, "\nIncompatible shared caches\n");
 			state->printIncompatibleHeader = 2;
 		}
-
 		j9tty_printf(PORTLIB, "%-16s\t", currentItem->name);
-		getStringForShcModlevel(PORTLIB, currentItem->versionData.modlevel, (char*)&jclLevelStr);
+		char jclLevelStr[10];
+		memset(jclLevelStr, 0, sizeof(jclLevelStr));
+		getStringForShcModlevel(PORTLIB, currentItem->versionData.modlevel, (char*)&jclLevelStr, sizeof(jclLevelStr));
 		getStringForShcAddrmode(PORTLIB, currentItem->versionData.addrmode, (char*)&addrmodeStr);
 		j9tty_printf(PORTLIB, "%s %s  ", jclLevelStr, addrmodeStr);
 		if (J9PORT_SHR_CACHE_TYPE_PERSISTENT == currentItem->versionData.cacheType) {
--- a/runtime/shared_common/OSCache.cpp
+++ b/runtime/shared_common/OSCache.cpp
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2001, 2015 IBM Corp. and others
+ * Copyright (c) 2001, 2018 IBM Corp. and others
  *
  * This program and the accompanying materials are made available under
  * the terms of the Eclipse Public License 2.0 which accompanies this
@@ -34,6 +34,7 @@
 #include "j2sever.h"
 #include "shrinit.h"
 #include "j9version.h"
+#include "shchelp.h"
 
 #include "OSCachesysv.hpp"
 #include "OSCachemmap.hpp"
@@ -69,14 +70,8 @@ SH_OSCache::getCacheVersionAndGen(J9PortLibrary* portlib, J9JavaVM* vm, char* bu
 	if (generation <= J9SH_GENERATION_07) {
 		J9SH_GET_VERSION_G07ANDLOWER_STRING(PORTLIB, versionStr, J9SH_VERSION(versionData->esVersionMajor, versionData->esVersionMinor), versionData->modlevel, versionData->addrmode);
 	} else {
-		J9PortShcVersion currentVersionData;
 		U_64 curVMVersion = 0;
 		U_64 oldVMVersion = 0;
-		
-		/* setCurrentCacheVersion is passed 'J2SE_17' because we don't care
-		 * about this field. Only the major, and minor numbers are needed for the current JVM
-		 */
-		setCurrentCacheVersion(vm, J2SE_17, &currentVersionData);
 		
 		/* CMVC 163957: 
 		 * There is overlap Java 6 cache file generations and Java 7. For example:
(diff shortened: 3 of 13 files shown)


llxia:
Jenkins test sanity aix jdk10

hangshao0:
Reviewers: @JasonFengJ9 , @pshipton

runtime/oti/j2sever.h
```diff
@@ -24,20 +24,16 @@
...
-#define J2SE_15   0x1500
-#define J2SE_16   0x1600
-#define J2SE_17   0x1700
-#define J2SE_18   0x1800
-#define J2SE_19   0x1900
-#define J2SE_V10  0x1A00            /* This refers Java 10 */
-#define J2SE_V11  0x1B00            /* This refers Java 11 */
+#define J2SE_18   0x0800
+#define J2SE_19   0x0900
+#define J2SE_V10  0x0A00            /* This refers Java 10 */
```
pshipton: Please add a comment to explain the constant values are used by the shared cache and the values of the constants must match the version of Java.

runtime/oti/j2sever.h
```diff
@@ -84,6 +80,7 @@
 #define J2SE_SHAPE_RAW	 		0x90000
 #define J2SE_SHAPE_MASK 		0xF0000
 #define J2SE_SHAPE_SHIFT		16
+#define J2SE_JAVA_SEPC_VERSIO_SHIFT 8
```
pshipton: There are a couple of typos in this name

runtime/util_core/j9shchelp.c
```diff
@@ -141,80 +141,67 @@ getShcModlevelForJCL(uintptr_t j2seVersion)
...
+ * Return the jave spec version for a given modification level
  * @param [in] modlevel  The modification level
  * 
  * @return the j2se version or 0 if one cannot be found
  */
 uint32_t
 getJCLForShcModlevel(uintptr_t modlevel)
 {
-	uint32_t j2seVersion = 0;
+	uint32_t javeVersion = 0;
```
pshipton: typo in this name

runtime/util_core/j9shchelp.c
```diff
@@ -141,80 +141,67 @@ getShcModlevelForJCL(uintptr_t j2seVersion)
...
-		j2seVersion = J2SE_V10;
+		javeVersion = 9;
 		break;
-	case J9SH_MODLEVEL_JAVA11 :
-		j2seVersion = J2SE_V11;
+	default:
+		if (modlevel > 9) {
+			/* J9SH_MODLEVEL_JAVA9 is 5. Does not have modlevel that is 6,7,8,9 */
+			javeVersion = modlevel;
+		}
```
pshipton: seems there should be an assert for the else case

runtime/util_core/j9shchelp.c
```diff
@@ -258,14 +246,13 @@ getStringForShcModlevel(J9PortLibrary* portlib, uint32_t modlevel, char* buffer)
...
-	case J9SH_MODLEVEL_JAVA10 :
-		strcpy(buffer, "Java10");
-		break;
-	case J9SH_MODLEVEL_JAVA11 :
-		strcpy(buffer, "Java11");
-		break;
 	default :
-		strcpy(buffer, "Unknown");
+		if (modlevel > 9) {
+			j9str_printf(portlib, buffer, 6, "%s%u", "Java", modlevel);
```
pshipton: Better to change the function to take the size as a parameter, otherwise this is a potential buffer overflow. (Plus I think it needs to be 7 to include the null).

runtime/util_core/j9shchelp.c
```diff
@@ -141,80 +141,67 @@ getShcModlevelForJCL(uintptr_t j2seVersion)
...
-		j2seVersion = J2SE_19;
-		break;
-	case J9SH_MODLEVEL_JAVA10 :
-		j2seVersion = J2SE_V10;
+		javeVersion = 9;
 		break;
-	case J9SH_MODLEVEL_JAVA11 :
-		j2seVersion = J2SE_V11;
+	default:
+		if (modlevel > 9) {
```
pshipton: Even though its the same thing, I think its a little clearer to use `>= 10` since from 10 its different.

runtime/util_core/j9shchelp.c
```diff
@@ -258,14 +246,13 @@ getStringForShcModlevel(J9PortLibrary* portlib, uint32_t modlevel, char* buffer)
...
 		break;
-	case J9SH_MODLEVEL_JAVA10 :
-		strcpy(buffer, "Java10");
-		break;
-	case J9SH_MODLEVEL_JAVA11 :
-		strcpy(buffer, "Java11");
-		break;
 	default :
-		strcpy(buffer, "Unknown");
+		if (modlevel > 9) {
```
pshipton: Even though its the same thing, I think its a little clearer to use `>= 10` since from 10 its different.

runtime/util_core/j9shchelp.c
hangshao0: This function is used when iterating/destroying the shared caches. I guess it is fine to not assert in this cases

pshipton: Not sure what you mean. If modlevel 6 - 9 should never occur, its fine to assert if they do.

runtime/oti/j2sever.h
pshipton: Is this requires many changes, better done in a separate commit.

JasonFengJ9: Yeah, that's true, never mind for this PR then.

runtime/util_core/j9shchelp.c
pshipton: Oh, you mean because jdk10 already uses 6, but is being converted to 10? Good point, how about adding a case for 6 to handle the transition.

hangshao0: Yes, the current JDK10 is already using 6. 
Also it is possible that someone put a fake cache file whose name is C280M7F0A32P_Cache1_G35 (modLevel is 7), VM would exit on this assertion failure and generate core files. User might not want to see this when destroying the caches (either using CML or SharedClassUtilities API). 

BTW, the cache name length needs to be increased. We currently have only 1 digit for modlevel in the cache name. We need two digits now. I will fix this.

runtime/util_core/j9shchelp.c
hangshao0: Passed in a buffer size parameter. memset this buffer to 0 before passing it to this function.

runtime/util_core/j9shchelp.c
hangshao0: Fixed

runtime/util_core/j9shchelp.c
hangshao0: Fixed

runtime/oti/j2sever.h
hangshao0: Fixed

runtime/shared_common/CacheLifecycleManager.cpp
```diff
@@ -286,9 +285,10 @@ printSharedCache(void* element, void* param)
 			j9tty_printf(PORTLIB, "\nIncompatible shared caches\n");
 			state->printIncompatibleHeader = 2;
 		}
-
 		j9tty_printf(PORTLIB, "%-16s\t", currentItem->name);
-		getStringForShcModlevel(PORTLIB, currentItem->versionData.modlevel, (char*)&jclLevelStr);
+		char jclLevelStr[10];
+		memset(jclLevelStr, 0, 10);
+		getStringForShcModlevel(PORTLIB, currentItem->versionData.modlevel, (char*)&jclLevelStr, 10);
```
pshipton: Please use sizeof(jclLevelStr) instead of hard coding 10.

runtime/shared_common/OSCache.cpp
```diff
@@ -91,9 +86,11 @@ SH_OSCache::getCacheVersionAndGen(J9PortLibrary* portlib, J9JavaVM* vm, char* bu
 		 oldVMVersion = SH_OSCache::getCacheVersionToU64(2, 60);
 		 curVMVersion = SH_OSCache::getCacheVersionToU64(versionData->esVersionMajor, versionData->esVersionMinor);
 		 
-		if ( curVMVersion >=  oldVMVersion) {
+		if ( curVMVersion >= oldVMVersion) {
 			if (generation <= J9SH_GENERATION_29) {
 				J9SH_GET_VERSION_G07TO29_STRING(PORTLIB, versionStr, J9SH_VERSION(versionData->esVersionMajor, versionData->esVersionMinor), versionData->modlevel, versionData->addrmode);
+			} else if (versionData->modlevel < 10){
```
pshipton: Missing space before `{`

runtime/util_core/j9shchelp.c
```diff
@@ -141,80 +165,67 @@ getShcModlevelForJCL(uintptr_t j2seVersion)
...
 	case J9SH_MODLEVEL_JAVA9 :
-		j2seVersion = J2SE_19;
-		break;
-	case J9SH_MODLEVEL_JAVA10 :
-		j2seVersion = J2SE_V10;
+		javaVersion = 9;
 		break;
-	case J9SH_MODLEVEL_JAVA11 :
-		j2seVersion = J2SE_V11;
+	default:
```
pshipton: I'm thinking there should be a case for `6` to handle obsoleted Java 10 cache names that may already exist.

runtime/util_core/j9shchelp.c
```diff
@@ -129,6 +129,30 @@ getGenerationFromName(const char* cacheNameWithVGen)
...
+ *
+ * @param [in] cacheNameWithVGen  the cache name
+ *
+ * @return the modeLevel number or 0 if an error occurred
+ */
+uintptr_t
+getModLevelFromName(const char* cacheNameWithVGen)
+{
+	char* cursor = (char*)cacheNameWithVGen;
+	uintptr_t modLevel;
```
pshipton: Please init variables when declared.

hangshao0:
Change updated. All comments addressed.

runtime/oti/j2sever.h
```diff
@@ -112,5 +114,10 @@
  */ 
 #define J2SE_SHAPE_FROM_ENV(env) J2SE_SHAPE(((J9VMThread*)env)->javaVM)
 
+/**
+ * Macro to extract java spec version from J2SE version.
+ */
+#define JAVA_SPEC_VERSION_FROM_J2SE(j2seVersion) 	(j2seVersion>> J2SE_JAVA_SPEC_VERSION_SHIFT)
```
pshipton: Missed it before, j2seVersion should be (j2seVersion) when used in the macro, with a space afterward

hangshao0: Fixed

pshipton:
jenkins test all plinux jdk8

hangshao0:
If I manually run the it, it will pass. 

```
OpenJ9-JDK10-PR1673-linux_ppc-64_cmprssptrs_le-201816042317/jdk/bin/java -Xshareclasses:name=ShareClassesUtilities,reset -version              
JVMSHRC023E Cache does not exist
openjdk version "10-internal" 2018-03-20
OpenJDK Runtime Environment (build 10-internal+0-adhoc.jenkins.PullRequest-Extended-JDK10-linuxppc-64cmprssptrsle-OpenJ9)
Eclipse OpenJ9 VM (build HEAD-a0c660b, JRE 10 Linux ppc64le-64 Compressed References 20180416_11 (JIT enabled, AOT enabled)
OpenJ9   - a0c660b
OMR      - a1e32cc
JCL      - a91ffe0 based on jdk-10+46)

OpenJ9-JDK10-PR1673-linux_ppc-64_cmprssptrs_le-201816042317/jdk/bin/java -cp /bluebird/builds/bld_384189/jvmtest/test/SE90/VM_Test/VM_Test.jar -Xshareclasses:name=ShareClassesCMLTests,utilities com.ibm.j9.sharedCacheAPI.tests.SharedUtilsTest
iterateSharedCache passed
destroySharedCache passed
```

I guess this failure is caused by an existing java10 cache left from previous tests.

hangshao0:
```
20:39:33 [Test]          [ERR] Cache name      	level         cache-type      feature         OS shmid       OS semid       last detach time
20:39:33 [Test]          [ERR] 
20:39:33 [Test]          [ERR] Compatible shared caches
20:39:33 [Test]          [ERR] ShareClassesUtilities	Java10 64-bit  persistent      cr                                            Mon Apr 16 16:35:23 2018
20:39:33 [Test]          [ERR] ShareClassesUtilities	Java10 64-bit  persistent      cr                                            Tue Apr 17 00:09:09 2018
```
There are 2 java 10 caches named ShareClassesUtilities on this machine. One created on today and one created yesterday. I believe the one created yesterday (has modLevel 6) is causing the test to fail. We can change the test to run -Xshareclasses:destroyAll on clean up.

pshipton:
The point was to handle old Java 10 caches so they can be used.

hangshao0:
All the other test failure on listAllCaches is caused by strrchr() in getModLevelFromName(). 
It is possible that cacheName has "M" in it (e.g. C290**M**10F1A64P_C**M**LTests_G35). So modLevel should the the number after the first "M" in the cache file name name rather last "M"
so strchr() instead of strrchr() should be used.

hangshao0:
Updates: 1. Allow modLevel 6 for SharedUtilsTest. 2. use strchr() instead of strrchr() in getModLevelFromName()

pshipton:
jenkins test all plinux

pshipton:
All the testing passed except jdk9 sanity which is still running (https://ci.eclipse.org/openj9/job/PullRequest-Sanity-JDK9-linux_ppc-64_cmprssptrs_le-OpenJ9/237/).

Please fix the merge conflicts.

hangshao0:
https://ci.eclipse.org/openj9/job/PullRequest-Sanity-JDK9-linux_ppc-64_cmprssptrs_le-OpenJ9/237/ also passed. 
Conflicts resolved.

runtime/shared_common/CacheLifecycleManager.cpp
```diff
@@ -286,9 +285,10 @@ printSharedCache(void* element, void* param)
 			j9tty_printf(PORTLIB, "\nIncompatible shared caches\n");
 			state->printIncompatibleHeader = 2;
 		}
-
 		j9tty_printf(PORTLIB, "%-16s\t", currentItem->name);
-		getStringForShcModlevel(PORTLIB, currentItem->versionData.modlevel, (char*)&jclLevelStr);
+		char jclLevelStr[10];
+		memset(jclLevelStr, 0, sizeof(jclLevelStr));
+		getStringForShcModlevel(PORTLIB, currentItem->versionData.modlevel, (char*)&jclLevelStr, sizeof(jclLevelStr));
```
keithc-ca: I'm surprise no compilers have complained about taking the address of an array. The name of an array yields (an r-value) which is the address of the first element of that array - you can't (normally) take the address of an r-value.

Instead of `(char*)&jclLevelStr` it should be just `jclLevelStr`. @hangshao0 Can you put together a new pull request that fixes that sort of problem in at least this file? Thanks.

hangshao0: New PR here: https://github.com/eclipse/openj9/pull/1719

PR #1675, opened 2018-04-12
Separate address node of BCDCHK and fix data type

NigelYiboYu:
BCDCHK is a node that can GC-and-Return. As a result of this,
LocalCSE does not common BCDCHK's address node with any address
node after the BCDCHK.

When the BCDCHK produces a PackedDecimal type, it'll be followed by a
pdstorei node, which has the same address node as the BCDCHK node.
If this BCDCHK and it's following pdstorei share the same
address node, then no address node after the pdstorei could be commoned
with the pdstorei's address node because BCDCHK prevents this from
happening. Failing to common such address nodes can result in unnecessary
load-after-store instruction sequence (e.g. VLRL after VSTRL). This is
redundant and inefficient.

Separating BCDCHK's and pdstorei's address nodes makes the aforementioned
address node commoning possible and opens up opportunities for packed
decimal copy propagations.

Also, fix the symbol reference data type for the pdstorei node of
I2PD and PD-shift-left. The stored type should be TR::PackedDecimal
instead of TR::Int8.

### DIFF
--- a/runtime/compiler/trj9/optimizer/DataAccessAccelerator.cpp
+++ b/runtime/compiler/trj9/optimizer/DataAccessAccelerator.cpp
@@ -1218,9 +1218,54 @@ bool TR_DataAccessAccelerator::generateI2PD(TR::TreeTop* treeTop, TR::Node* call
       TR::Node * i2pdNode = TR::Node::create((isI2PD)?TR::i2pd:TR::l2pd, 1, intNode);
       i2pdNode->setDecimalPrecision(precision);
 
-      //create pdstore
-      //TODO: use pdstore or ipdstore? the result is saved inside register.
-      TR::Node * arrayAddressNode = constructAddressNode(callNode, pdNode, offsetNode);
+      /**
+       * Create separate address nodes for BCDCHK and pdstorei because BCDCHK can GC-and-Return like a call.
+       *
+       * Having separate address nodes also allows AddrNode2 and AddrNode3 commoning, which then makes
+       * copy propagations possible.
+       *
+       * AddrNode1 could still be commoned with address nodes before the BCDCHK. Hence, the need for
+       * UncommonBCDCHKAddressNode codegen pass. AddrNode1 is special in that it has to be rematerialized and used
+       * at the end of the BCDCHK OOL path's GC point. No commoning of this node should happen.
+       *
+       * Example:
+       *
+       * BCDCHK
+       *   pdshlOverflow <prec=9 (len=5) adj=0 round=0>
+       *     ....
+       *   aladd (internalPtr sharedMemory )                                    AddrNode1
+       *     ==>newarray
+       *     lconst 8 (highWordZero X!=0 X>=0 )
+       *     ....
+       * pdstorei  <array-shadow>[#490  Shadow] <prec=9 (len=5)>
+       *   aladd (internalPtr sharedMemory )                                    AddrNode2
+       *     ==>newarray
+       *     ==>lconst 8
+       *   ==>pdshlOverflow <prec=9 (len=5)
+       * zdsleStorei  <array-shadow>[#492  Shadow]  <prec=9 (len=9)>
+       *   ....
+       *   zd2zdsle <prec=9 (len=9)>
+       *     pd2zd <prec=9 (len=9)>
+       *       pdloadi <prec=9 (len=5) adj=0 round=0>
+       *         aladd (internalPtr sharedMemory )                              AddrNode3
+       *           ==>newarray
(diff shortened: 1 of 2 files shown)


NigelYiboYu:
Marking this WIP until my tests are completed.

runtime/compiler/trj9/optimizer/DataAccessAccelerator.cpp
```diff
@@ -1218,9 +1218,54 @@ bool TR_DataAccessAccelerator::generateI2PD(TR::TreeTop* treeTop, TR::Node* call
...
+       * See BCDCHKEvaluatorImpl() for BCDCHK tree structure and intended use of its children.
+       *
+       * 'outOfLineCopyBackAddr' and 'storeAddressNode' correspond to AddrNode1 and AddrNode2, respectively. They
+       * are created as separate nodes so that LocalCSE is able to common up AddrNode2 and AddrNode3. If
+       * AddrNode1 and AddrNode2 were the same node, the LocalCSE would not consider AddrNode1 an alternative replacement
+       * of AddrNode3 because the BCDCHK's symbol canGCAndReturn().
+       *
+       * With AddrNode2 and AddrNode3 commoned up, the LocalCSE is able to copy propagate pdshlOverflow to the the pd2zd
+       * tree and replace its pdloadi.
+       */
```
fjeremic: Excellent documentation! Future readers, take note 😃

runtime/compiler/trj9/z/codegen/J9TreeEvaluator.cpp
```diff
@@ -15809,12 +15809,34 @@ J9::Z::TreeEvaluator::BCDCHKEvaluator(TR::Node * node, TR::CodeGenerator * cg)
                break;
                }
 
-            default: TR_ASSERT_FATAL(0, "BCDCHKEvaluator: Could not find recognized method for variable precision DAA operation.\n");
+            default:
+               {
+               /**
+                *
```
fjeremic: Extra line here. Can we remove?

runtime/compiler/trj9/z/codegen/J9TreeEvaluator.cpp
```diff
@@ -15809,12 +15809,34 @@ J9::Z::TreeEvaluator::BCDCHKEvaluator(TR::Node * node, TR::CodeGenerator * cg)
...
          }
 
-      default: TR_ASSERT_FATAL(0, "Operation Code not supported by BCDCHKEvaluator.\n");
+      default:
+         {
+         /**
+          * Unrecognized opCodes under BCDCHK should come from optimizations such as local CSE and tree simplifications.
+          * They should be commoned nodes that's evaluated previously. Skip these nodes.
+          */
+         TR_ASSERT_FATAL(resultReg != NULL, "BCDCHKEvaluator: BCDCHK has an unevaluated non-PD node %p (non-PD op code %s) \n",
```
fjeremic: Shouldn't it be the simplifier's job to remove the `BCDCHK` as well if it folds the operation underneath? This seems like special casing a situation that should never occur. The `BCDCHK` IL has a tight specification that it's first child is an operation that may raise a hardware exception. If a `BCDCHK` reaches codegen and this is not the case then the tree verifier (currently does not run in production builds) should have asserted that this is invalid IL.

runtime/compiler/trj9/z/codegen/J9TreeEvaluator.cpp
NigelYiboYu: Removed the extra line.

runtime/compiler/trj9/z/codegen/J9TreeEvaluator.cpp
NigelYiboYu: In theory, yes we can make the simplifier remove such `BCDCHK` nodes. There's one catch: the call parameter nodes under the `BCDCHK` has to be anchored somewhere and get evaluated. This is especially important for the single-reference-unevaluated call parameter nodes.

Shall we open an issue for the simplifier?

fjeremic: > Shall we open an issue for the simplifier?

@NigelYiboYu yes I think so. Please reference this assert in the item you open. This assertion should happen at the tree verification level, i.e. check first child of `BCDCHK` and make sure it is a BCD operation.

fjeremic:
Jenkins test all

fjeremic:
AIX build has an infrastructure failure and the Travis build was blocked by #1711 at the time of the latest commit. Everything else looks good for a merge.

PR #1676, opened 2018-04-12
Run redundant goto elimination before codegen

jdmpapin:
After global dead store elimination and dead trees elimination, there
may be blocks containing nothing but goto. This commit gives redundant
goto elimination a chance to look at any such blocks detected by the
preceding pass(es) of dead trees elimination.


jdmpapin:
This is marked WIP because it depends on eclipse/omr#2444.

andrewcraik:
Jenkins test sanity

jdmpapin:
I ran Daytrader3 with and without this change, 13 times each, and for each execution took the total of the time spent compiling on each compilation thread as shown by `TR_PrintCompTime`. There was no significant difference in the time spent compiling. The sample means differed by 0.3% with standard deviations of over 1%.

andrewcraik:
Jenkins test sanity

andrewcraik:
Jenkins test sanity

andrewcraik:
single failure was infrastructure and with all the other platforms passing I am good to merge this.

PR #1677, opened 2018-04-12
Run local CSE to aid idiom recognition at warm

jdmpapin:
An extra pass of local common subexpression elimination (CSE) before
idiom recognition allows the latter to recognize arraycmp in
String.equals at warm.

### DIFF
--- a/runtime/compiler/trj9/optimizer/J9Optimizer.cpp
+++ b/runtime/compiler/trj9/optimizer/J9Optimizer.cpp
@@ -670,6 +670,7 @@ static const OptimizationStrategy cheapWarmStrategyOpts[] =
    { OMR::loopReduction                                                              },
    { OMR::blockShuffling                                                             },
 #endif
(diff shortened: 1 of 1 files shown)


andrewcraik:
Jenkins test sanity

andrewcraik:
@jdmpapin could you comment on any evaluation you have done on the compile-time impact of this change?

jdmpapin:
I ran Daytrader3 with and without this change, 20 times each, and for each execution took the total of the time spent compiling on each compilation thread as shown by `TR_PrintCompTime`. There was no significant difference in the time spent compiling. The sample means differed by 0.15% with standard deviations of over 1%.

andrewcraik:
Jenkins test sanity

andrewcraik:

andrewcraik:
Jenkins test sanity win jdk9

andrewcraik:
Java 9 is not in the gating set due to #2020, the same code would have run with 8 and 10. Change is easy to verify and quite safe so merging.

PR #1678, opened 2018-04-12
Remove unnecessary loads in deduplicateStrings

jdmpapin:
Currently the preferred backing array reference is loaded twice: once as
an integer, and again as a reference. Because of the differing types,
the JIT can't common these loads. This commit adds JITHelpers.acmplt so
that deduplicateStrings can use an address comparison instead of an
integer comparison, which allows us to use the backing array references
that have already been loaded by the caller. The JIT compiler will
replace the call to JITHelpers.acmplt with an acmplt IL node during
Unsafe fast pathing.

Also skip the ((s1.count ^ s2.count) >= 0) test unless

    enableCompression && null != compressionFlag

Otherwise counts are always nonnegative. In particular, when compression
is disabled, this test can be eliminated entirely.


DanHeidinga:
@jdmpapin Did you consider creating a `JITHelpers.getAddress(Object)` helper?  This would give you the same ability to compare where objects were in memory (tenure or not) without needing the RefWrapper while still being as correct with respect to the GC moving objects as the current code.

Or even better from an avoiding potential future bugs, a `JITHelpers.compareAddresses(Object, Object)` method?

andrewcraik:
I agree that we should try to find a way to avoid allocating when running this code in the interpreter to avoid hurting startup performance. Will wait to hear thoughts from @jdmpapin.

jdmpapin:
> Please describe the purpose behind this change and the benchmarks that have been run to validate it.

The purpose is to speed up `String.equals` in its most expensive case: when the strings are equal but have not yet been deduplicated. This is expected to happen for instance when a hash lookup succeeds given a fresh string as the key. This change produces a speedup on a microbenchmark meant to exaggerate the fixed costs in this case. On Daytrader the effect is too small to be observed by itself.

It looks like I may be changing this code to avoid allocation, so I'll provide more details after I've made the necessary adjustments and re-measured.

> Did you consider creating a JITHelpers.getAddress(Object) helper?

I don't think we allow conversion from GC-managed references into integer addresses during optimization, since there's nothing to stop the addresses from getting stale.

> Or even better from an avoiding potential future bugs, a JITHelpers.compareAddresses(Object, Object) method?

This is exactly the `JITHelpers.acmplt` method whose non-JIT implementation we're discussing :smile:

> This also adds an allocation to the String.equals() path

I wasn't worried about this because it will only allocate when we're interpreting `deduplicateStrings`. However, I see that Andrew is concerned about it as well. Here are some ways we might be able to get rid of the allocation:

1. Make `acmplt` a native method. There's JNI overhead, but maybe fast JNI would be good?
2. Keep a thread-local `RefWrapper` so that no thread ever allocates it more than once. I believe this would replace the allocation with something like a hash lookup, which may not be a win.

Thoughts?

jdmpapin:
After speaking with Andrew and Dan, I'm going to change the new `JITHelpers.acmplt` so that it is declared `native`, and implement it as an out of line INL. The JIT compiler will still recognize it and replace it with a single comparison operation.

jdmpapin:
Updated to fix a copyright header, to avoid allocating as previously mentioned, and to make `deduplicateStrings` take the values as parameters, since the caller already has them.

While comparing the effect of this change I included in both builds an extra pass of local CSE before idiom recognition at warm, as in #1677. This allows `String.equals` to produce an `arraycmp` which in this case helps to exaggerate the cost of `deduplicateStrings`.

Daytrader3 still shows no change.

Here is a  [microbenchmark](https://github.com/eclipse/openj9/files/1926662/EqualButNotDeduplicated.java.txt) that repeatedly compares strings that are equal but not deduplicated. It uses `Unsafe` to reset the `value` references and "forget" the deduplication, so as to avoid GC. I mean for the change to benefit warm, so I ran with `-Xjit:initialOptLevel=warm,inhibitRecompilation` to prevent upgrades. On strings of length 9 (which are currently processed fastest by `arraycmp`), I see an improvement of 25%.

jcl/src/java.base/share/classes/com/ibm/jit/JITHelpers.java
```diff
@@ -353,6 +353,8 @@ public char byteToCharUnsigned(byte b) {
 		return (char) ((char) b & (char) 0x00ff);
 	}
 
+	public native boolean acmplt(Object lhs, Object rhs);
```
DanHeidinga: Can you javadoc the method?

jcl/src/java.base/share/classes/java/lang/String.java
```diff
@@ -2307,37 +2307,21 @@ public boolean equals(Object object) {
 		return false;
 	}
 
-	private static final void deduplicateStrings(String s1, String s2) {
-		if ((s1.count ^ s2.count) >= 0) {
+	private static final void deduplicateStrings(String s1, Object value1, String s2, Object value2) {
```
DanHeidinga: Should `value1` and `value2` be of type `char[]`?

Also please javadoc this method as well.  Its something we haven't always been good about but should improve that when making changes to method signatures.

jcl/src/java.base/share/classes/java/lang/String.java
```diff
@@ -2307,37 +2307,21 @@ public boolean equals(Object object) {
...
+		if (!enableCompression || null == compressionFlag || (s1.count ^ s2.count) >= 0) {
```
DanHeidinga: Is this meant to be checking whether the string is compressed and only dedup'ing non-compressed strings?

The check for compression in `isCompressed()` is:
```
if (enableCompression && (null == compressionFlag || count >= 0)) {
```
which should be this negated:
```
if (!enableCompression || (null != compressionFlag && count < 0)) {
```

Did I miss something?

jdmpapin: This test ensures that we only deduplicate strings that are both compressed or both uncompressed.

When one string is compressed and the other isn't, we can't deduplicate because doing so would require updating both `value` and `count`, and other threads could see an inconsistent state.

If `!enableCompression`, then both strings are always uncompressed. If OTOH `enableCompression` but `null == compressionFlag`, then both strings must be compressed. So it's only necessary to check the compression bits when `enableCompression && null != compressionFlag`.

jcl/src/java.base/share/classes/java/lang/String.java
jdmpapin: To give them a more specific type, they would have to be `char[]` or `byte[]` depending on the version using `/*[IF Sidecar19-SE]*/`, `/*[ELSE]*/`. Since the body only uses them at the type `Object` anyway, I wanted to avoid unnecessary noise from the preprocessing directives. However I don't feel super strongly about it and I can make the signature conditional if that's what you prefer.

I'll add the javadoc.

jdmpapin:
Updated to add javadoc for `acmplt` and `deduplicateStrings`

jcl/src/java.base/share/classes/java/lang/String.java
```diff
@@ -2307,37 +2307,43 @@ public boolean equals(Object object) {
...
+	 *			  the first string
+	 * @param value1
+	 *			  {@code s1.value}
+	 * @param s2
+	 *			  the second string
+	 * @param value2
+	 *			  {@code s2.value}
+	 */
+	private static final void deduplicateStrings(String s1, Object value1, String s2, Object value2) {
+		if (!enableCompression || null == compressionFlag || (s1.count ^ s2.count) >= 0) {
```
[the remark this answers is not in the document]
jdmpapin: OK, I've added a comment here

runtime/compiler/trj9/optimizer/UnsafeFastPath.cpp
```diff
@@ -143,6 +143,45 @@ int32_t TR_UnsafeFastPath::perform()
...
 
+         if (symbol->getRecognizedMethod() == TR::com_ibm_jit_JITHelpers_acmplt)
+            {
+            if (!performTransformation(comp(),
+                  "%s Found unsafe/JITHelpers calls, turning node "
+                  "[" POINTER_PRINTF_FORMAT "] "
+                  "into an address less-than comparison\n",
+                  optDetailString(),
+                  node))
+               continue;
```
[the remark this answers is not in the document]
jdmpapin: I agree that's a good practice, but it's atypical style in the JIT compiler. However, the multiline condition makes it less strange to add the braces in this particular case, and I've done so.

jcl/src/java.base/share/classes/com/ibm/jit/JITHelpers.java
```diff
@@ -353,6 +353,23 @@ public char byteToCharUnsigned(byte b) {
...
+	 * Determine whether {@code lhs} is at a lower address than {@code rhs}.
+	 *
+	 * Because objects can be moved by the garbage collector, the ordering of
+	 * the addresses of distinct objects is not stable over time. As such, the
+	 * result of this comparison should be used for heuristic purposes only.
+	 *
+	 * A null reference is considered less than any non-null reference.
+	 *
+	 * @param lhs
+	 *			  The left hand side of the comparison
```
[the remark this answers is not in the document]
jdmpapin: I would also prefer that formatting, but it's inconsistent with every existing `@param` in both `String.java` and `JITHelpers.java` with the sole exception of `String#split(String, int)`.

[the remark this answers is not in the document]
DanHeidinga: I would rather we start using a reasonable format for javadoc.  It'll mean less work to fix the formatting at some future point.

jcl/src/java.base/share/classes/java/lang/String.java
```diff
@@ -2307,37 +2307,55 @@ public boolean equals(Object object) {
...
+	 *
+	 * @param s1
+	 *			  the first string
+	 * @param value1
+	 *			  {@code s1.value}
+	 * @param s2
+	 *			  the second string
+	 * @param value2
+	 *			  {@code s2.value}
+	 */
```
DanHeidinga: Same comment about formatting here.  The javadoc is inconsistently formatted throughout the OpenJ9 class libraries.  New javadoc should adhere to the formats that are commonly used.  Over time this will hopefully lead to cleaner formatting.

jcl/src/java.base/share/classes/java/lang/String.java
DanHeidinga: good point

jcl/src/java.base/share/classes/com/ibm/jit/JITHelpers.java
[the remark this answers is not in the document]
jdmpapin: Updated to reformat `@param`s in the new javadoc comments

PR #1681, opened 2018-04-13
Vectorized core logic of String.indexOf() on X86

0dvictor:
Core logic of String.indexOf() is separated into an
internal helper function, which is implemented with
SIMD instructions on X86.

### DIFF
--- a/jcl/src/java.base/share/classes/com/ibm/jit/JITHelpers.java
+++ b/jcl/src/java.base/share/classes/com/ibm/jit/JITHelpers.java
@@ -474,6 +474,24 @@ public final class JITHelpers {
 		}
 	}
 
+	public int findElementFromArray(Object array, byte ch, int offset, int length) {
+		for (int i = offset; i < length; i++) {
+			if(getByteFromArrayByIndex(array, i) == ch) {
+				return i;
+			}
+		}
+		return -1;
+	}
+
+	public int findElementFromArray(Object array, char ch, int offset, int length) {
+		for (int i = offset; i < length; i++) {
+			if(getCharFromArrayByIndex(array, i) == ch) {
+				return i;
+			}
+		}
+		return -1;
+	}
+
 	/*
 	 * Constants for optimizedClone
 	 */
--- a/jcl/src/java.base/share/classes/java/lang/String.java
+++ b/jcl/src/java.base/share/classes/java/lang/String.java
@@ -1964,20 +1964,10 @@ public final class String implements Serializable, Comparable<String>, CharSeque
 				// Check if the String is compressed
 				if (enableCompression && (null == compressionFlag || coder == LATIN1)) {
 					if (c <= 255) {
-						byte b = (byte) c;
-
-						for (int i = start; i < len; ++i) {
-							if (helpers.getByteFromArrayByIndex(array, i) == b) {
-								return i;
-							}
-						}
+						return helpers.findElementFromArray(array, (byte)c, start, len);
 					}
 				} else {
-					for (int i = start; i < len; ++i) {
-						if (helpers.getCharFromArrayByIndex(array, i) == c) {
-							return i;
-						}
-					}
+					return helpers.findElementFromArray(array, (char)c, start, len);
 				}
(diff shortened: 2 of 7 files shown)


jcl/src/java.base/share/classes/com/ibm/jit/JITHelpers.java
```diff
@@ -134,6 +134,30 @@ private static int classIsInterfaceFlag() {
...
+	{
+		for (int i = offset; i < length; i++)
+		{
+			if(getCharFromArrayByIndex(array, i) == ch)
+			{
+				return i;
+			}
+		}
+		return -1;
+	}
```
fjeremic: Would it be possible to have consistent coding style with the rest of the file?

0dvictor: For sure. Updated.

fjeremic:
@0dvictor I think something bad happen with a rebase here. There are a ton of foreign commits.

0dvictor:
Thank you for the reminder. A rebase went bad. I've fixed it.

jcl/src/java.base/share/classes/java/lang/String.java
```diff
@@ -5958,20 +5958,10 @@ public int indexOf(int c, int start) {
...
 				if (enableCompression && (null == compressionFlag || count >= 0)) {
 					if (c <= 255) {
-						byte b = (byte) c;
-
-						for (int i = start; i < len; ++i) {
-							if (helpers.getByteFromArrayByIndex(array, i) == b) {
-								return i;
-							}
-						}
+						return helpers.findElementFromArray(array, (byte)c, start, len);
```
fjeremic: There are two implementations of `public int indexOf(int c, int start)` in this file. One for Java 8 and one for Java 9+. Could we make the same change for the Java 9+ version?

0dvictor: Sorry I didn't realized that the two functions now are separated. I've updated both of them.

andrewcraik:
@0dvictor the change from @msqb has been merged - is this ready to go?

0dvictor:
I've rebased the change. I'm re-running all the test to satisfy myself that the change is still good after rebasing. Will remove WIP as soon as the tests clear.

0dvictor:
The code has been re-tested and therefore is ready for merging.

runtime/compiler/env/VMJ9.cpp
```diff
@@ -3244,6 +3244,7 @@ int TR_J9VMBase::checkInlineableTarget (TR_CallTarget* target, TR_CallSite* call
    // Don't inline methods that are going to be reduced in ilgen or UnsafeFastPath
    switch (rm)
       {
+      case TR::com_ibm_jit_JITHelpers_findElementFromArray:
```
andrewcraik: so this isn't reduced by ilgen or UnsafeFastPath so the comment above now seems to be misleading...

what happens on platforms that don't support this call in the codegen? I think there needs to be a query so that if this isn't supported in the codegen we add it to the always inline list. The danger being until other codegens support this helper they will see a perf degradation...

0dvictor: Good point, I've moved this into the CG specific query `suppressInliningOfRecognizedMethod()`, so that each CG can opt-in/opt-out this optimization.

andrewcraik: Can we make sure that if the inlining isn't suppressed we put it into the list of always worth inlining to avoid regressions on the platforms which don't initially opt in?

0dvictor: In theory the Inliner should make better decision whether to inline `findElementFromArray`, but I see your point about affecting the existing code as little as possible to minimize surprises.
I've update that `findElementFromArray` is always inlined if CG does not request suppressing inlining of it.

andrewcraik:
Jenkins test sanity

andrewcraik:
win64 suffered a GIT failure, and win failed because of #2129. Others are clean. Change looks good - merging.

PR #1682, opened 2018-04-13
Fix devirtualization issues with TR_RelocationRecord/IlGeneratorMethodDetails

mpirvu:
Compiling OpenJ9 with a more recent compiler (e.g. gcc 5.x) can expose some
long existing bugs in two parts of the code:
(1) Allocating a TR_RelocationRecord and morphing it into a specialized
relocation record
(2) Copying an IlGeneratorMethodDetails object in
TR_MethodToBeCompiled::initialize()

In both cases the code constructs an object of a derived class on top of an
object of a base class using `placement new`. Of course, for this to have any
chance of success, the size of the two objects, base and derived, must be the
same and we make sure that indeed that's the case.
However, the code suffers from a subtle problem: the C++ compiler sees an
object of a base class being instantiated and it's free to devirtualize calls
off such an object, leading to the wrong code to be executed.

For clarification, let's use a generic example:
```
class B {
  public:
    B(param); // constructor
    virtual void foo();  
};
class D : public B {
  public:
    virtual void foo();
};
B::B(param) {
  if (param==...)
    new (this) D(); // constructing 'D' on top of existing object of type 'B'
}
---- main code ---
B b(param);
b.foo();  // A virtual call may call D::foo()
          // With devirtualization B::foo() is called always
```

The pattern used here is that of an object factory: based on some input we
want to create an object of a specific derived type, but return a pointer to
the base class so that it can be handled through virtual functions.
The difference from a traditional object factory is that we use placement
new to construct on top of an existing object (rather than allocating a brand
new object).
For the relocation record case mentioned above this approach has the advantage
of allocating a single TR_RelocationRecord on the stack and reusing that 
storage for every record that the code needs to process.
For the ILGeneratorMethodDetails case this approach is somewhat forced by our
decision to embed such an object into TR_MethodToBeCompiled.

The simplest fix for the existing bugs is to use the object factory pattern
and allowing it to construct at a designated address, but always use the value
returned by the factory when calling virtual functions. 
The example above could be modified along the following lines:
```
class B {
  ...
  // My object factory
  static B *create(B *storage, param) {
     if (param == ..) 
        return new (storage) D();
    ...
  }
};  

B storage; // this is where the new object is going to be created
B *b = B::create(storage, param); // create a specific object
b->foo();
```

### DIFF
--- a/runtime/compiler/trj9/control/MethodToBeCompiled.cpp
+++ b/runtime/compiler/trj9/control/MethodToBeCompiled.cpp
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2000, 2016 IBM Corp. and others
+ * Copyright (c) 2000, 2018 IBM Corp. and others
  *
  * This program and the accompanying materials are made available under
  * the terms of the Eclipse Public License 2.0 which accompanies this
@@ -50,7 +50,7 @@ TR_MethodToBeCompiled *TR_MethodToBeCompiled::allocate(J9JITConfig *jitConfig)
 
 void TR_MethodToBeCompiled::initialize(TR::IlGeneratorMethodDetails & details, void *oldStartPC, CompilationPriority p, TR_OptimizationPlan *optimizationPlan)
    {
-   _methodDetails = details;
+   _methodDetails = TR::IlGeneratorMethodDetails::clone(_methodDetailsStorage, details);
    _optimizationPlan = optimizationPlan;
    _next = NULL;
    _oldStartPC = oldStartPC;
--- a/runtime/compiler/trj9/ilgen/IlGeneratorMethodDetails.hpp
+++ b/runtime/compiler/trj9/ilgen/IlGeneratorMethodDetails.hpp
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2000, 2016 IBM Corp. and others
+ * Copyright (c) 2000, 2018 IBM Corp. and others
  *
  * This program and the accompanying materials are made available under
  * the terms of the Eclipse Public License 2.0 which accompanies this
@@ -50,8 +50,8 @@ public:
    IlGeneratorMethodDetails(const TR::IlGeneratorMethodDetails & other) :
       J9::IlGeneratorMethodDetailsConnector(other) {}
 
-   IlGeneratorMethodDetails & operator=(const TR::IlGeneratorMethodDetails & other);
    };
+
 }
 
 
--- a/runtime/compiler/trj9/ilgen/J9IlGeneratorMethodDetails.cpp
+++ b/runtime/compiler/trj9/ilgen/J9IlGeneratorMethodDetails.cpp
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2000, 2016 IBM Corp. and others
+ * Copyright (c) 2000, 2018 IBM Corp. and others
  *
  * This program and the accompanying materials are made available under
  * the terms of the Eclipse Public License 2.0 which accompanies this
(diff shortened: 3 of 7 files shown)


mpirvu:
@mstoodle Mark, could you please review the first commit related to relocation records? There will be another commit (under the same PR) for IlGeneratorMethodDetails. Thanks

runtime/compiler/trj9/runtime/RelocationRecord.hpp
```diff
@@ -156,15 +156,17 @@ class TR_RelocationRecord
...
       void * operator new (size_t s, TR_RelocationRecord *p)   { return p; }
 
       virtual void print(TR_RelocationRuntime *reloRuntime);
       virtual char *name() { return "TR_RelocationRecord"; }
 
-      void decode(TR_RelocationTarget *reloTarget);
+     
+      static TR_RelocationRecord *create(TR_RelocationRecord *storage, TR_RelocationRuntime *reloRuntime, TR_RelocationTarget *reloTarget, TR_RelocationRecordBinaryTemplate *recordPointer)
+         {
+         return decode(storage, reloRuntime, reloTarget, recordPointer);
```
mstoodle: Why not just fold decode() into create and keep create() out of the header? Nothing else calls decode, I don't think, so the extra call doesn't seem valuable to me.

runtime/compiler/trj9/runtime/RelocationRecord.cpp
```diff
@@ -293,162 +296,172 @@ TR_RelocationRecordGroup::handleRelocation(TR_RelocationRuntime *reloRuntime,
...
-         new (this) TR_RelocationRecordEmitClass();
+         reloRecord = new (storage) TR_RelocationRecordEmitClass();
          break;
       default:
          // TODO: error condition
          printf("Unexpected relo record: %d\n", reloType);fflush(stdout);
          TR_ASSERT(0, "Unexpected relocation record type found");
          exit(0);
       }
+      reloRecord->_reloRuntime = reloRuntime;
```
mstoodle: It's more code to change, but adding these fields to the constructors for all the relocation classes (which pass them into the base class constructor) to initialize would be a lot cleaner. Then you don't need to add the "setBinaryRecord()" function which isn't really something RelocationRecord should be allowing (especially as a public function).

runtime/compiler/trj9/runtime/RelocationRecord.hpp
```diff
@@ -216,6 +219,9 @@ class TR_RelocationRecord
       TR_RelocationRecordBinaryTemplate *_record;
 
       TR_RelocationRecordPrivateData _privateData;
+   private:
+      static TR_RelocationRecord * decode(TR_RelocationRecord *storage, TR_RelocationRuntime *reloRuntime, TR_RelocationTarget *reloTarget, TR_RelocationRecordBinaryTemplate *record);
+      static uint8_t getReloType(TR_RelocationTarget *reloTarget, TR_RelocationRecordBinaryTemplate *rec);
```
mstoodle: getReloType() is probably better implemented as a nonvirtual instance function on the TR_RelocationRecordBinaryTemplate struct. That's somewhat out of character for those structs, but I like that better than adding this function here.

mpirvu: getReloType() is used in a static function, so it has to be static as well (not an instance function). I could move it to TR_RelocationRecordBinaryTemplate, but still as a static function.

mpirvu:
@mstoodle I have addressed the suggestions from the review.

mstoodle:
Looks good for this first change. Thanks, @mpirvu !

runtime/compiler/trj9/control/MethodToBeCompiled.cpp
```diff
@@ -50,7 +50,7 @@ TR_MethodToBeCompiled *TR_MethodToBeCompiled::allocate(J9JITConfig *jitConfig)
 
 void TR_MethodToBeCompiled::initialize(TR::IlGeneratorMethodDetails & details, void *oldStartPC, CompilationPriority p, TR_OptimizationPlan *optimizationPlan)
    {
-   _methodDetails = details;
+   _methodDetails = J9::IlGeneratorMethodDetails::create(_methodDetailsStorage, details);
```
mstoodle: This should not referenced J9::IlGeneratorMethodDetails directly. Only TR::IlGeneratorMethodDetails should be referenced outside of its class hierarchy.

You'll need to expose a static create() call in the TR:: version that directs down to the J9 static implementation.

runtime/compiler/trj9/ilgen/IlGeneratorMethodDetails.hpp
```diff
@@ -50,8 +50,8 @@ class OMR_EXTENSIBLE IlGeneratorMethodDetails : public J9::IlGeneratorMethodDeta
    IlGeneratorMethodDetails(const TR::IlGeneratorMethodDetails & other) :
       J9::IlGeneratorMethodDetailsConnector(other) {}
 
-   IlGeneratorMethodDetails & operator=(const TR::IlGeneratorMethodDetails & other);
```
mstoodle: if you want that create function, you'll need to declare it here

runtime/compiler/trj9/ilgen/J9IlGeneratorMethodDetails.cpp
```diff
@@ -35,47 +35,35 @@
...
-namespace TR
-{
-
-IlGeneratorMethodDetails & IlGeneratorMethodDetails::operator=(const TR::IlGeneratorMethodDetails & other)
-   {
-   return self()->J9::IlGeneratorMethodDetailsConnector::operator=(other);
-   }
-
-}
-
```
mstoodle: ...and you'll need to implement create() here to call J9::IlGeneratorMethodDetails::create()

runtime/compiler/trj9/control/MethodToBeCompiled.cpp
dsouzai: I thought that for code within the `trj9` folder it was ok to refer use `J9::`, or is it because `IlGeneratorMethodDetails` is extensible that we have to use `TR::`. But then I see examples in CompilationThread.cpp of `J9::NewInstanceThunkDetails` instead of `TR::NewInstanceThunkDetails`.

runtime/compiler/trj9/ilgen/J9IlGeneratorMethodDetails.cpp
mpirvu: That's exactly what I did, but I hit a strange compilation error in trj9/env/j9method.cpp when using the other create function:
`TR::IlGeneratorMethodDetails & details = TR::IlGeneratorMethodDetails::create(storage, this);`
It's like the compiler cannot distinguish between
`inline static TR::IlGeneratorMethodDetails & create(TR::IlGeneratorMethodDetails & target, TR_ResolvedMethod *method);`
and
` static IlGeneratorMethodDetails *create(TR::IlGeneratorMethodDetails &storage, const TR::IlGeneratorMethodDetails & other);`
I thought that `TR_ResolvedMethod *method` was considered different than `TR::IlGeneratorMethodDetails & other`
Changing the name of the second create function to something else solves the problem.

mpirvu: I think I am still going to change the name of function I added to 'clone' because it better reflects what it does: it takes an object and it allocates an identical object at the specified address.

mpirvu:
Jenkins test all

mpirvu:
Off the 6 builds that failed, 2 are connection issues, 2 are disk space issues and for two (JDK8 lnx 390/ppc) I see java.io.FileNotFoundException.

mpirvu:
I verified manually on a Ubuntu 16.04 box (with gcc 5.4.0) that, before this fix, DLT methods were seen as OrdinaryMethods; after the fix they are correctly identified as MethodInProgress.

mpirvu:
@mstoodle Mark, the code is ready to be merged, unless you would like more modifications. Thanks

runtime/compiler/trj9/ilgen/J9IlGeneratorMethodDetails.cpp
```diff
@@ -35,12 +35,14 @@
...
 
+
 namespace TR
 {
 
-IlGeneratorMethodDetails & IlGeneratorMethodDetails::operator=(const TR::IlGeneratorMethodDetails & other)
+IlGeneratorMethodDetails *
+IlGeneratorMethodDetails::clone(TR::IlGeneratorMethodDetails &storage, const TR::IlGeneratorMethodDetails & other)
    {
-   return self()->J9::IlGeneratorMethodDetailsConnector::operator=(other);
```
mstoodle: Just curious why you put this particular forwarding function into the cpp file, which means little chance it will ever be inlined...?

runtime/compiler/trj9/ilgen/J9IlGeneratorMethodDetails.hpp
```diff
@@ -74,10 +74,10 @@ class OMR_EXTENSIBLE IlGeneratorMethodDetails : public OMR::IlGeneratorMethodDet
 
    IlGeneratorMethodDetails(const TR::IlGeneratorMethodDetails & other);
 
-   TR::IlGeneratorMethodDetails & operator=(const TR::IlGeneratorMethodDetails & other);
-
    static TR::IlGeneratorMethodDetails & create(TR::IlGeneratorMethodDetails & target, TR_ResolvedMethod *method);
```
mstoodle: there is no longer any create() implementation, is there? this declaration should be removed

mpirvu:
I removed the clone() method from TR::IlGeneratorMethodDetails so that it can call directly the implementation from J9::IlGeneratorMethodDetails through inheritance.

mstoodle:
Great set of changes; thanks @mpirvu for fixing these long standing obstacles!

JamesKingdon:
Time to warm up gcc 7 and see if we get any througput differences?

mpirvu:
Throughput no, but maybe, maybe some startup time improvements from lower compilation time due to better devirtualization.

Issue #1203, opened 2018-02-16
TestUTF8 fails with assertion Unsafe_copySwapMemory0 is unimplemented


JamesKingdon:
openjdk-tests/openjdk_regression/openjdk-jdk/jdk/test/sun/nio/cs/TestUTF8.java

fails with an assertion:

```
#3  0xf6f169b2 in __assert_fail_base (fmt=0x1 <error: Cannot access memory at address 0x1>, 
    assertion=0xf54360ec "!\"Java_jdk_internal_misc_Unsafe_copySwapMemory0 is unimplemented\"", 
    assertion@entry=0x2 <error: Cannot access memory at address 0x2>, file=file@entry=0xd20de470 "\001", line=857, 
    line@entry=4143841408, 
    function=function@entry=0xf5435f28 <Java_jdk_internal_misc_Unsafe_copySwapMemory0::__PRETTY_FUNCTION__> "void Java_jdk_internal_misc_Unsafe_copySwapMemory0(JNIEnv*, jobject, jobject, jlong, jobject, jlong, jlong, jlong)") at assert.c:92
#4  0xf6f16a4a in __GI___assert_fail (assertion=0x2 <error: Cannot access memory at address 0x2>, 
    file=0xd20de470 "\001", line=4143841408, 
    function=0xf5435f28 <Java_jdk_internal_misc_Unsafe_copySwapMemory0::__PRETTY_FUNCTION__> "void Java_jdk_internal_misc_Unsafe_copySwapMemory0(JNIEnv*, jobject, jobject, jlong, jobject, jlong, jlong, jlong)") at assert.c:101
#5  0xf541a7b0 in Java_jdk_internal_misc_Unsafe_copySwapMemory0 ()
   from /home/jkingdon/sdks/jdk/lib/default/libjclse9_29.so
#6  0xf62d4768 in ffi_call_VFP () from /home/jkingdon/sdks/jdk/lib/default/libffi29.so
#7  0xf62d3490 in ffi_call () from /home/jkingdon/sdks/jdk/lib/default/libffi29.so
#8  0xf6397560 in bytecodeLoop () from /home/jkingdon/sdks/jdk/lib/default/libj9vm29.so
#9  0xf63e3d28 in cInterpreter () from /home/jkingdon/sdks/jdk/lib/default/libj9vm29.so
```
[...]

PR #1683, opened 2018-04-13
Implement copySwapMemory native

theresa-m:
- implement native using general copyMemory method
- correct copySwapMemory description in jdk.internal.misc.Unsafe
- make Unsafe.java copySwapMemoryChecks easier to read
- add new UnsafeTests for copySwapMemory

Fixes: #1203

### DIFF
--- a/runtime/jcl/common/sun_misc_Unsafe.cpp
+++ b/runtime/jcl/common/sun_misc_Unsafe.cpp
@@ -469,6 +469,88 @@ determineCommonAlignment(UDATA sourceOffset, UDATA destOffset, UDATA size)
 	return alignment;
 }
 
+/**
+ * Copy actualSize bytes from source object to destination.
+ * Helper method for copyMemory and copyMemoryByte.
+ *
+ * @param currentThread
+ * @param sourceObject object to copy from
+ * @param sourceOffset location in srcObject to start copy
+ * @param destObject object to copy into
+ * @param destOffset location in destObject to start copy
+ * @param actualSize the number of bytes to be copied
+ * @param logElementSize common alignment of offsets and actualSize
+ * @param sourceIndex index into source array
+ * @param destIndex index into destination array
+ * @param elementCount elements to copy in array
+ */
+static VMINLINE void
+copyMemorySub(J9VMThread* currentThread, j9object_t sourceObject, UDATA sourceOffset, j9object_t destObject,
+		UDATA destOffset, UDATA actualSize, UDATA logElementSize, UDATA sourceIndex, UDATA destIndex,
+		UDATA elementCount)
+{
+	if (NULL == sourceObject) {
+		if (NULL == destObject) {
+			UDATA sourceEnd = sourceOffset + actualSize;
+			if ((destOffset > sourceOffset) && (destOffset < sourceEnd)) {
+				alignedBackwardsMemcpy(currentThread, (void*)(destOffset + actualSize), (void*)sourceEnd, actualSize, logElementSize);
+			} else {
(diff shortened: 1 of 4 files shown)


runtime/jcl/common/sun_misc_Unsafe.cpp
```diff
@@ -871,12 +934,139 @@ Java_jdk_internal_misc_Unsafe_registerNatives(JNIEnv *env, jclass clazz)
...
+			if (!J9CLASS_IS_ARRAY(clazz)) {
+				goto illegal;
+			}
+			if (!J9ROMCLASS_IS_PRIMITIVE_TYPE(((J9ArrayClass*)clazz)->componentType->romClass)) {
+				goto illegal;
+			}
+		}
+
+		jboolean overlapIsUnaligned = memOverlapIsUnaligned(sourceObject, sourceOffset, destObject, destOffset, actualCopySize);
+		jboolean overlapIsNone = memOverlapIsNone(sourceObject, sourceOffset, destObject, destOffset, actualCopySize);
```
DanHeidinga: Does this need to do both calls?  Can it be adjusted to only call `memOverlapIsUnaligned()` if `! memOverlapIsNone()`?

runtime/jcl/common/sun_misc_Unsafe.cpp
```diff
@@ -871,12 +934,139 @@ Java_jdk_internal_misc_Unsafe_registerNatives(JNIEnv *env, jclass clazz)
...
+		 * and the overlap is unaligned. This will prevent any errors during swapping.
+		 */
+		if (overlapIsUnaligned) {
+			copyMemory(currentThread, sourceObject, sourceOffset, destObject, destOffset, actualCopySize);
+		}
+
+		/* use temporary byte for swap if memory is overlapping */
+		UDATA tempOffset = 0;
+		if (! overlapIsNone) {
+			tempOffset = (UDATA)unsafeAllocateMemory(currentThread, (UDATA)1, FALSE);
```
DanHeidinga: Why malloc this?  Could a stack allocated value (local variable) be used instead?

DanHeidinga:
@theresa-m Do the tests pass on OpenJDK?

theresa-m:
The tests pass in OpenJDK.

theresa-m:
@j9build retest copyright

runtime/jcl/common/sun_misc_Unsafe.cpp
```diff
@@ -871,12 +934,118 @@ Java_jdk_internal_misc_Unsafe_registerNatives(JNIEnv *env, jclass clazz)
...
+ * @param sourceObject
+ * @param sourceOffset memory address
+ * @param destObject
+ * @param destOffset memory address
+ * @param actualCopySize size of memory after address
+ * @return true for no memory overlap, otherwise false
+ */
+jboolean
+ memOverlapIsNone(j9object_t sourceObject, UDATA sourceOffset, j9object_t destObject, UDATA destOffset, UDATA actualCopySize) {
+	if (sourceObject == NULL && destObject == NULL) {
```
DanHeidinga: A couple of nitpicks:
* `(sourceObject == NULL && destObject == NULL)` -> `((sourceObject == NULL) && (destObject == NULL))`
Extra brackets means no one has to worry about order of operations.  One less thing for readers of the code to have to remember.

* better to have a single return point.  It always setting a single breakpoint to catch all returns from the function when debugging.

theresa-m: I've applied these changes

DanHeidinga:

DanHeidinga:
Jenkins test sanity zlinux jdk10

Issue #685, opened 2017-11-25
ARM Crash when running on 64 bit OS due to unaligned IProfiler data


JamesKingdon:
When running on a 64 bit OS the JVM will crash with a gpf on a non-aligned address:

```
1XHEXCPCODE    J9Generic_Signal_Number: 00000008
1XHEXCPCODE    Signal_Number: 00000007
1XHEXCPCODE    Error_Value: 00000000
1XHEXCPCODE    Signal_Code: 00000001
1XHEXCPCODE    Handler1: F6421D38
1XHEXCPCODE    Handler2: F6364D4C
1XHEXCPCODE    InaccessibleAddress: F65B2015
```
The crash can be avoided by using -Xjit:disableInterpreterProfiling

JamesKingdon:
I have a patch in my forks which avoids the crash by using memcpy to set the data. This sounds bad, but I think the calls get inlined and we end up with near optimal code. I'm experimenting with inlined assembly to see if how that would look and if there's any benefit to be had.

JamesKingdon:
The inline assembly looks clean and avoids concerns that a newer compiler might go one step further and optimise successive memcpy calls to a store multiple after inlining.

PR #1691, opened 2018-04-14
Prevent unaligned errors from interpreterProfiling

JamesKingdon:
The interpreter profiling buffer frequently writes to unaligned addresses.
This is mostly ok on ARM, except when gcc transforms multiple writes into a
load or store multiple instruction, which generates sigbus when running under
a 64 bit OS. Unfortunately there doesn't seem to be an option to prevent gcc
from doing this, so I've added inline assembler with explicit individual
store instructions in the one known problem location.


JamesKingdon:
Fixed up the bad indentation.

JamesKingdon:
Fixes #685

runtime/vm/BytecodeInterpreter.hpp
```diff
@@ -294,11 +294,28 @@ class INTERPRETER_CLASS
 	{
 		U_8 *profilingCursor = startProfilingRecord(REGISTER_ARGS, sizeof(J9Class*) + 2*sizeof(J9Method*));
 		if (NULL != profilingCursor) {
+#ifdef J9ARM
```
gacholio: Please use "#if defined(J9VM_ARCH_ARM)"

runtime/vm/BytecodeInterpreter.hpp
```diff
@@ -294,11 +294,28 @@ class INTERPRETER_CLASS
...
+				"str %[clazz],  [%[cursor], #0]\n\t"
+				"str %[method], [%[cursor], #4]\n\t"
+				"str %[target], [%[cursor], #8]\n\t"
+				:            // No outputs
+				:[cursor] "r" (profilingCursor),
+				 [clazz]  "r" (clazz),
+				 [method] "r" (callingMethod),
+				 [target] "r" (targetMethod)
+			 );
+#else
```
gacholio: Should be commented.

runtime/vm/BytecodeInterpreter.hpp
```diff
@@ -294,11 +294,28 @@ class INTERPRETER_CLASS
...
 			*(J9Class**)profilingCursor = clazz;
 			profilingCursor += sizeof(J9Class*);
 			*(J9Method**)profilingCursor = callingMethod;
 			profilingCursor += sizeof(J9Method*);
 			*(J9Method**)profilingCursor = targetMethod;
+#endif
```
gacholio: Should be commented.

gacholio:
This code is only correct for 32-bit.  Should it maybe be gated with an "#if defined(J9VM_ENV_DATA64)" which errors out?  Presumably the C code produce bad results on 64-bit as well.

runtime/vm/BytecodeInterpreter.hpp
JamesKingdon: Just checking, you'd like a comment associated with the #else and #endif statements to connect them with the corresponding #ifdef? Or you'd like more comments on the individual inline assembly statements?

gacholio: Yeah, the "standard" here is a comment to match up the preprocessor directives (there are many examples in the file, I'm sure).

gacholio:
I suppose it would be simple enough to just include the 64-bit code (I think the only difference is in the offsets of the stores).

runtime/vm/BytecodeInterpreter.hpp
```diff
@@ -294,11 +294,28 @@ class INTERPRETER_CLASS
 	{
 		U_8 *profilingCursor = startProfilingRecord(REGISTER_ARGS, sizeof(J9Class*) + 2*sizeof(J9Method*));
 		if (NULL != profilingCursor) {
+#ifdef J9ARM
+			// ARMv8 does not support store multiple to unaligned addresses, and unfortunately GCC
+			// generates a store multiple for the normal version of these three assignments.
+			// To avoid the store multiple instructions, we can replace the assignments with
+			// inline assembly code.
+			asm volatile(
```
gacholio: Does this need to be volatile?  Not sure what reordering opportunities would actually occur here.

JamesKingdon: My take is that if I'm putting in inline assembly I want it emitted where I put it, so I use volatile by default. Are there disadvantages to using the keyword?

gacholio: It prevents any reordering across the block, by compiler or processor.

Also here, the cursor appears to not be updated after the asm block.

JamesKingdon: Indeed, the cursor is not used again so does not need updating.

gacholio: I'm not sure volatile would prevent the block itself from being moved, subject to all the usual dependency restrictions.  Certainly there is no compiler reordering within the block, but there could be processor reordering (which would be fine in this case anyway).

gacholio: Right - the cursor has already been updated and stored back before we fill in the data.

JamesKingdon:
Comments addressed.

I don't think the issue will impact the 64 bit JVM as the store multiple instruction has been dropped from the instruction set and the replacement is capable of writing to unaligned addresses. I added `&& !defined(J9VM_ENV_DATA64)` to the guard for the inlined code so that it will only apply to the 32 bit case. (In all likelihood the 64 bit JVM will use J)VM_ARCH_AARCH64 anyway, so the extra test is unnecessary, but we can clean that up later when such things have been finalised.)

runtime/vm/BytecodeInterpreter.hpp
```diff
@@ -294,11 +294,29 @@ class INTERPRETER_CLASS
...
+				"str %[clazz],  [%[cursor], #0]\n\t"
+				"str %[method], [%[cursor], #4]\n\t"
+				"str %[target], [%[cursor], #8]\n\t"
+				:            // No outputs
+				:[cursor] "r" (profilingCursor),
+				 [clazz]  "r" (clazz),
+				 [method] "r" (callingMethod),
+				 [target] "r" (targetMethod)
+			 );
+#else /* J9VM_ARCH_ARM */
```
gacholio: Last nitpick - comments no longer match now that the 64-bit exclude has been added.

gacholio:
Sorry for the delay - I'm not getting emails about updates here (likely because of my draconian notification settings).

JamesKingdon:
@gacholio comments updated

Issue #1205, opened 2018-02-17
ARM sigbus from initialiseAOTRelocationHeader


JamesKingdon:
J9::ARM::AheadOfTimeCompile::initialiseAOTRelocationHeader(TR::IteratedExternalRelocation*) uses stm to access unaligned memory causing sigbus and a core file when running on 64bit OS. This causes a failure in test/VM_Test

`=> 0xf5d8ee78 <_ZN2J93ARM18AheadOfTimeCompile29initialiseAOTRelocationHeaderEPN2TR26IteratedExternalRelocationE+1504>:	stmib	r4, {r1, r3}`

```
(gdb) info reg
r0             0xf0f0f6e8	4042323688
r1             0xffffffff	4294967295
r2             0xed13f102	3977507074
r3             0xee008df0	3993013744
r4             0xed13f0f6	3977507062
r5             0xf0c05608	4039136776
r6             0xf6399c78	4130970744
r7             0xefc00000	4022337536
r8             0xefc076c0	4022367936
r9             0xf11abcd8	4045061336
r10            0xefcdc810	4023240720
r11            0x3	3
r12            0x0	0
sp             0xf0ee3818	0xf0ee3818
lr             0x40	64
pc             0xf5d8ee78	0xf5d8ee78 <J9::ARM::AheadOfTimeCompile::initialiseAOTRelocationHeader(TR::IteratedExternalRelocation*)+1504>
cpsr           0x800f0010	-2146500592

```

JamesKingdon:
Problem doesn't show up when JIT is built with debug, which makes things a little more challenging.

knn-k:
The stmib instruction seems to be used in the following part of J9AheadOfTimeCompile.cpp:

```      case TR_ConstantPool:
      case TR_ConstantPoolOrderedPair:
      case TR_Trampolines:
      case TR_Thunks:
         {
         *(uintptrj_t *)cursor = (uintptrj_t)relocation->getTargetAddress2(); // inlined site index
         cursor += SIZEPOINTER;

         *(uintptrj_t *)cursor = (uintptrj_t)relocation->getTargetAddress(); // constantPool
         cursor += SIZEPOINTER;
         break;
         }
```

JamesKingdon:
Awesome, thanks! How did you find it?
[...]

PR #1692, opened 2018-04-15
WIP: Prevent unaligned errors from AOT relocation

JamesKingdon:
initialiseAOTRelocationHeader frequently writes to unaligned addresses.
This is mostly ok on ARM, except when gcc transforms multiple writes into a
load or store multiple instruction, which generates sigbus when running under
a 64 bit OS. Unfortunately there doesn't seem to be an option to prevent gcc
from doing this, so I've added inline assembler with explicit individual
store instructions to all the cases where the problem can happen.

### DIFF
--- a/runtime/compiler/trj9/arm/codegen/J9AheadOfTimeCompile.cpp
+++ b/runtime/compiler/trj9/arm/codegen/J9AheadOfTimeCompile.cpp
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2000, 2017 IBM Corp. and others
+ * Copyright (c) 2000, 2018 IBM Corp. and others
  *
  * This program and the accompanying materials are made available under
  * the terms of the Eclipse Public License 2.0 which accompanies this
@@ -127,6 +127,9 @@ uint8_t *J9::ARM::AheadOfTimeCompile::initialiseAOTRelocationHeader(TR::Iterated
    *flagsCursor = modifier;
    uint32_t *wordAfterHeader = (uint32_t*)cursor;
 
+   // gcc 4.9.4 will combine multiple writes to *cursor into a stm instruction
+   // which will cause sigbus when running under a 64 bit OS. The probablem can
+   // be avoided by using inline assembly.
    switch (relocation->getTargetKind())
       {
       case TR_HelperAddress:
@@ -160,13 +163,21 @@ uint8_t *J9::ARM::AheadOfTimeCompile::initialiseAOTRelocationHeader(TR::Iterated
          uintptr_t inlinedSiteIndex = findCorrectInlinedSiteIndex(tempSR->getOwningMethod(comp)->constantPool(), comp, recordInfo->data2);
          uint8_t flags = (uint8_t) recordInfo->data3;//Sequence ID
          //TODO
-         *(uintptrj_t *)cursor = inlinedSiteIndex;
-         cursor += SIZEPOINTER;
+
          // next word is the address of the constant pool to
          // which the index refers
-         *(uintptrj_t *)cursor =
-            (uintptr_t)tempSR->getOwningMethod(TR::comp())->constantPool();
-         cursor += SIZEPOINTER;
+         uint8_t constantPool = (uintptr_t)tempSR->getOwningMethod(TR::comp())->constantPool();
+
+         asm volatile(
+            "str %[inlinedSiteIndex], [%[cursor], #0]\n\t"
+            "str %[constantPool], [%[cursor], #4]\n\t"
+            :            // No outputs
+            :[cursor]            "r" (cursor),
+             [inlinedSiteIndex]  "r" (inlinedSiteIndex),
+             [constantPool]      "r" (constantPool)
+         );
+         cursor += 2*SIZEPOINTER;
+
          }
          break;
       case TR_JNIStaticTargetAddress:
@@ -178,18 +189,25 @@ uint8_t *J9::ARM::AheadOfTimeCompile::initialiseAOTRelocationHeader(TR::Iterated
          {
          TR::SymbolReference *tempSR = (TR::SymbolReference *)relocation->getTargetAddress();
          uintptr_t inlinedSiteIndex = (uintptr_t)relocation->getTargetAddress2();
-
          inlinedSiteIndex = findCorrectInlinedSiteIndex(tempSR->getOwningMethod(comp)->constantPool(), comp, inlinedSiteIndex);
 
-         *(uintptrj_t *)cursor = inlinedSiteIndex;  // inlinedSiteIndex
-         cursor += SIZEPOINTER;
-
-         *(uintptrj_t *)cursor = (uintptrj_t)tempSR->getOwningMethod(comp)->constantPool(); // constantPool
-         cursor += SIZEPOINTER;
+         uintptrj_t constantPool = (uintptrj_t)tempSR->getOwningMethod(comp)->constantPool();
 
          uintptrj_t cpIndex=(uintptrj_t)tempSR->getCPIndex();
-         *(uintptrj_t *)cursor =cpIndex;// cpIndex
-         cursor += SIZEPOINTER;
+
+         asm volatile(
+            "str %[inlinedSiteIndex],  [%[cursor], #0]\n\t"
+            "str %[constantPool], [%[cursor], #4]\n\t"
+            "str %[cpIndex], [%[cursor], #8]\n\t"
+            :            // No outputs
+            :[cursor]            "r" (cursor),
+             [inlinedSiteIndex]  "r" (inlinedSiteIndex),
+             [constantPool]      "r" (constantPool),
+             [cpIndex]           "r" (cpIndex)
+         );
+
+         cursor += 3*SIZEPOINTER;
+
          }
          break;
       case TR_ClassAddress:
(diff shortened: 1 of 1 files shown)


JamesKingdon:
Fixes #1205

mstoodle:
@JamesKingdon did you consider allocating a RelocationTarget object to encapsulate these sequences as a baby-ish step towards migrating the AOT code generation support with the AOT runtime support?

JamesKingdon:
I guess "what's a RelocationTarget object?" probably answers your question...

mstoodle:
Are there particular relocation records that are problematic? Maybe we should just adjust the alignment on some fields since I would actually have expected most fields to be properly aligned...


For educational purposes: the AOT relocation runtime was re-architected a few years ago to better reflect the activities during AOT loads. One part of that was the creation of a hierarchy of classes known as the `RelocationTarget`, which provides a set of services to read/write values from the code cache that are used by other classes in the runtime (e.g. `RelocationRecord`, which has subclasses to deal with each type of relocation record).

The original premise was that RelocationTarget would provide symmetric services for reading/writing values to the code cache so that the code generators and the runtime could reliably interpret/create relocatable code cache constructs, but somehow we never got around to converting the code generator implementions to use the newer framework.

What I was suggesting above would actually be a "new" use for RelocationTarget in that I suggesting you use it to read/write values from the relocation data (as opposed to the code cache). For most platforms, that's probably overkill but I think it would be better than injecting inlined assembler sequences all over the place.

JamesKingdon:
> Are there particular relocation records that are problematic?

The original debugging work (that produced the memcpy mitigation) happened a few months ago so I don't remember all the details As best I can recollect, after iteratively fixing about three of them I got fed up and just did every case that unconditionally writes >1 consecutive words. There are lots of questions left unanswered, such as; 
. how do the records get unaligned as all I ever see are 32 bit writes, and
. why don't we get problems when the records are read?

RelocationTarget sounds interesting, but given the lack of resources and the fairly dramatic problems still outstanding with the ARM build I'd be keen to get something that works in now and revisit for elegance and style later.

mstoodle:
> There are lots of questions left unanswered, such as;
>
>    * how do the records get unaligned as all I ever see are 32 bit writes, and
>    * why don't we get problems when the records are read?

Sorry to say it, but that sound suspiciously like not getting to the root cause to me...

JamesKingdon:
True, but there are **many** rabbit holes...

mstoodle:
Sure, but that's not an excuse for wallpapering over mouldy insulation just to flip the house (if you'll pardon the flippant home renovation metaphor :) ).

The "hey did you think about using RelocationTarget" suggestion I could take back as some notional nudge toward "elegance and style" that may not be warranted at this stage in the game.

But adding 15 very similar inline assembler snippets into a file without any other assembler snippets to address a problem you can't explain feels closer to my metaphor (with apologies). And any new relocation records (@dsouzai is in the process of adding a new one) will also need to insert these inline assembler snippets, so it's already going to be an expanding bandaid.

I am not a fan of solving a problem before the root cause has been found, just in case this "rabbit hole" is deeper than you realize.

mstoodle:
Hm, maybe I missed an important detail here.

The scenario that you're running 32-bit ARM code on a 64-bit ARM system with a 64-bit OS? In that case, I can see how it gets messed up since the 32-bit header will always mean the next part of the relocation record will not be 64-bit aligned. Assuming I've understood that properly, then another way to address the problem would be to change the #ifdef that adds the alignment word after the header on 64-bit platforms. In the runtime, for example, look at the definition of TR_RelocationRecordBinaryTemplate in runtime/RelocationRecord.cpp 👍 

```
struct TR_RelocationRecordBinaryTemplate
   {
   uint16_t _size;
   uint8_t _type;
   uint8_t _flags;

#if defined(TR_HOST_64BIT)
   uint32_t _extra;
#endif
   };
```

Maybe you need something extra there on the TR_HOST_64BIT to detect ARM32 on 64-bit OS? At least until we can move up to a GCC level that resolves the stm problem it creates?

JamesKingdon:
I see it a little differently. The existing C code works and has been substantially tested on 32 bit OS. The failure occurs specifically on 64 bit OS because of gcc generating an inappropriate instruction. Preventing that instruction from being used with inlined assembly fixes the problem and has also passed testing - that's what I spent most of my Sunday doing.
Digging deeper with a veiw to what, exactly? Changing the alignment that has been working and continues to work? That's a far riskier proposition that I don't currently have time for.
I accept the criticism of the 15 similar bits of assembly. By the time I got to about number 11 I was wishing I'd created a wrapper function for them all, but there are different numbers of words that need to be written (from 2 to 5), so you'd need to do something clever with variadics to handle it efficiently (I don't want a loop, the code isn't that ugly).

JamesKingdon:
Ok, so we cross posted. No, I don't believe it needs to be 64 bit aligned, 32 bit should be good enough - we're writing 32 bit words.

JamesKingdon:
So is it your assertion that the data should always be at least 32 bit aligned? If so, then perhaps we really do have something to go looking for. I was assuming that the data was at least potentially less aligned in the same way as the interpreter profiling data.

mstoodle:
That's where I was going with it, yes. The only place we access with less than 32-bit width is in the very first word, and that's accessed directly and consistently as a 16-bit value (via a 16-bit pointer type) and two 8 bit values (via 8-bit pointer type).

JamesKingdon:
Ah, ok. That is a bit different then. If it's not supposed to ever be unaligned and it somehow is, there's likely to be perf to be had from fixing it. @dsouzai Do you know where the buffer comes from? Wondering if it's wrong from the very start.

dsouzai:
> Do you know where the buffer comes from

https://github.com/eclipse/openj9/blob/1d55e83f9a15fbe8362b60330a112303ec3f3cd2/runtime/compiler/trj9/x/codegen/J9AheadOfTimeCompile.cpp#L65-L90

each codegen does this I believe; this where the buffer gets allocated; the size of each header is read from an array of sizes + some more space for each of the offsets in each relo record (the size depending on whether or not to use wide offsets :P)

mstoodle:
it's the offsets....they default to 16b offsets unless the method is bigger than 64KB of code. I bet we don't add an extra offset if there's an odd number of offsets :( .

JamesKingdon:
I don't have a feel for how many of these things are in play, or what unit the padding would need to be applied to. Is an extra two bytes to get things aligned a reasonable trade-off?

dsouzai:
> Is an extra two bytes to get things aligned a reasonable trade-off?

I think different people have different views about "wasted" bytes in the SCC. I can't imagine 2 bytes per relo-record with an odd-numbered set of offsets would add up to all that much.

That said, implementing that code right is probably going to be tricky. It has to be exact on both the write-to-SCC and load-from-SCC code paths.

JamesKingdon:
Well, at least we have achieved a better understanding of the problem. Personally, after staring at it so long, I'm ok with the inline assembler. It's obvious what it's doing and not actually that much more lines of code than the original C.

JamesKingdon:
There's a spurious place-holder comment that needs to be removed.

mstoodle:
Fortunately, there's only one place where realignment would be needed on the "runtime" side. And (in principle, at least) it could be done on a per-platform basis (meaning only arm32 codegen would need to add the extra 2 bytes per relocation record).

It should be a relatively simple change (famous last words). The external iterated relocations already have a notion of how many 16-bit offsets there are. That number needs to get rounded up when we calculate the size of the record, and the loop that lays down 16-bit offsets also needs to conditionally add a zero offset at the end. I *think* that should be confined to J9::ARM::AheadOfTimeCompile::processRelocations() but that might not be all there is to it.

On the runtime side, there's a "nextBinaryRecord()" query that adds the size of the relo record that could be augmented to ask the relocation target (with TR_ARM32RelocationTarget returning "true") if it needs to align relocation record size on 32-bits. Along with that change, the offset processing needs to iterate until it hits the size of the relocation record (i.e. the not-necessarily-32-bit aligned end of the relocation record), not nextBinaryRecord() as it does now.

JamesKingdon:
Ok, I can take a crack at that sometime. It looks like it's not going to be an issue on 64bit as the ldm and stm instructions were dropped from the instruction set.

Thanks for digging into what needs to be done!

JamesKingdon:
I seemed to have suffered from over-confidence in my previous comment. I can't find anything that matches your description, and for now the whole thing remains entirely opaque.

dsouzai:
Heh took a while, but it happens here:
https://github.com/eclipse/omr/blob/76d9c20d159dddb21f979e02589a85a0b44fff15/compiler/codegen/OMRCodeGenPhase.cpp#L166

which calls this
https://github.com/eclipse/openj9/blob/e2420d40b885fcd033c3040493d6f79fdbfd002d/runtime/compiler/trj9/codegen/J9CodeGenerator.cpp#L2467

and here's where the offsets get placed:
https://github.com/eclipse/openj9/blob/e2420d40b885fcd033c3040493d6f79fdbfd002d/runtime/compiler/trj9/codegen/J9CodeGenerator.cpp#L2711-L2716

https://github.com/eclipse/omr/blob/76d9c20d159dddb21f979e02589a85a0b44fff15/compiler/codegen/Relocation.cpp#L248-L259

https://github.com/eclipse/omr/blob/76d9c20d159dddb21f979e02589a85a0b44fff15/compiler/codegen/Relocation.cpp#L489-L503

dsouzai:
This is where it happens at load time:

https://github.com/eclipse/openj9/blob/e2420d40b885fcd033c3040493d6f79fdbfd002d/runtime/compiler/trj9/runtime/RelocationRecord.cpp#L602

JamesKingdon:
Many thanks Irwin

mstoodle:
i don't think you actually need to modify the offsets that are being inserted. What you need to do is "round up" the size of the relocation record itself.

For the record, I was trying to describe: https://github.com/eclipse/openj9/blob/master/runtime/compiler/trj9/arm/codegen/J9AheadOfTimeCompile.cpp#L89-L98 on the codegen side. If `getSizeOfRelocationData()` returned a rounded up to 32-bits size on the ARM platform, I believe you would always have relocation records starting on 32-bit boundaries.

You'll need to be careful that you only write the offsets that are actually needed, and record the right size into the relocation record (to only include the offsets) so that the runtime will only process actual offsets (and not treat the padding like an offset).


On the runtime side, I was proposing you modify how this code (https://github.com/eclipse/openj9/blob/master/runtime/compiler/trj9/runtime/RelocationRecord.cpp#L259-L270) operates by adjusting the return value from `nextBinaryRecord(reloTarget)`: https://github.com/eclipse/openj9/blob/master/runtime/compiler/trj9/runtime/RelocationRecord.cpp#L259-L270

Note how it only uses the size of the relocation record; I proposed to add a new RelocationTarget query to ask if the target needs 32-bit aligned relocation records, which could be used to add 2 bytes to the return value if necessary.

Does that make sense?

JamesKingdon:
Thanks Mark, I'll try and take another look at this.

mstoodle:
@JamesKingdon it's been a while on this one. What would you like to do with it? Is there value to keep this WIP PR alive (it's not the oldest by any stretch :) ) or should we just close and wait the newer better version to materialize?

JamesKingdon:
I won't be likely to work on this again until winter, so have no objections to it being closed.

andrewcraik:
Discussion on Aug 27 suggests this is abandoned and major AOT work is being done by at @dsouzai and @jdmpapin so likely this should be addressed? I'm going to close this since it is very stale at this point - a new PR is likely the best way to address this if it still needs to be looked at.

JamesKingdon:
Sounds good, and much better than my hacky fix :)

PR #1693, opened 2018-04-15
X86 CG inlines isAssignableFrom for abstract class

0dvictor:
java.lang.Class.isAssignableFrom() for abstract class is now inlined by CG on X86.


andrewcraik:
Jenkins test sanity

PR #1695, opened 2018-04-16
fix the testng filename

llxia:
- In Jsr292, the filename should be testng_SE80.xml (not testngSE80.xml)
according to variables.mk and playlist.xml settings

- In OpenJ9_Jsr_292_API, there is no need to have java version specific
testng.xml

[ci skip]


llxia:
Searched all dirs under test and https://github.com/eclipse/openj9/blob/master/test/OpenJ9_Jsr_292_API/testngSE90.xml is the only one.

It will not break the build as it is directly referenced in playlist.xml `$(Q)$(TEST_RESROOT)$(D)testngSE90.xml$(Q)`. However, we should change it to follow the same naming convention.

Will update the changeset once extended tests are done.

llxia:
On the second thought, I donot see the need to have Java version specific testng.xml for OpenJ9_Jsr_292_API test. We only have one testng file in the project. I am going to remove SE90 in the filename.

llxia:
Jenkins test extended plinux jdk9,jdk8

llxia:

Issue #1659, opened 2018-04-11
cmdLineTester_getCallerClassTests_SE90_0 failed in Test-Extended-JDK10-linux_ppc-64_cmprssptrs_le


llxia:
cmdLineTester_getCallerClassTests_SE90_0 failed in Test-Extended-JDK10-linux_ppc-64_cmprssptrs_le

https://ci.eclipse.org/openj9/job/Test-Extended-JDK10-linux_ppc-64_cmprssptrs_le/7/

getCallerClassTests_SE100.xml is missing.

```
cmdLineTester_getCallerClassTests_SE90_0 Start Time: Wed Apr 11 07:00:51 2018 Epoch Time (ms): 1523430051232
test with NoOptions
{ "/home/jenkins/jenkins-agent/workspace/Test-Extended-JDK10-linux_ppc-64_cmprssptrs_le/jdk/bin/java"  -Xcompressedrefs  -Xdump -DRESJAR="/home/jenkins/jenkins-agent/workspace/Test-Extended-JDK10-linux_ppc-64_cmprssptrs_le/openj9/test/TestConfig/scripts/testKitGen/../../../../jvmtest/cmdline_options_testresources/cmdlinetestresources.jar" -DEXE='"/home/jenkins/jenkins-agent/workspace/Test-Extended-JDK10-linux_ppc-64_cmprssptrs_le/jdk/bin/java"  -Xcompressedrefs  -Xdump --add-exports=java.base/com.ibm.oti.vm=ALL-UNNAMED --add-exports java.base/jdk.internal.reflect=ALL-UNNAMED' -jar "/home/jenkins/jenkins-agent/workspace/Test-Extended-JDK10-linux_ppc-64_cmprssptrs_le/openj9/test/TestConfig/scripts/testKitGen/../../../../jvmtest/cmdline_options_tester/cmdlinetester.jar" -config "/home/jenkins/jenkins-agent/workspace/Test-Extended-JDK10-linux_ppc-64_cmprssptrs_le/openj9/test/TestConfig/scripts/testKitGen/../../../../jvmtest/cmdLineTests/getCallerClassTests/getCallerClassTests_SE100.xml" -explainExcludes -nonZeroExitWhenError; \
[...]

PR #1696, opened 2018-04-16
Introduce a "VersionCheck" class for getting JDK major version

DanHeidinga:
I've seen a number of tests lately that need which JDK they're being run on. Here's a utility class that implements a cleaner version of the check.

It defaults to JDK8 when the Runtime.Version class isn't present, which was added in JDK9.  For JDK9+, it returns the `Runtime.Version.major()` value.

### DIFF
--- a/test/TestUtilities/src/org/openj9/test/util/VersionCheck.java
+++ b/test/TestUtilities/src/org/openj9/test/util/VersionCheck.java
@@ -1,0 +1,68 @@
+/*******************************************************************************
+ * Copyright (c) 2018, 2018 IBM Corp. and others
+ *
+ * This program and the accompanying materials are made available under
+ * the terms of the Eclipse Public License 2.0 which accompanies this
+ * distribution and is available at https://www.eclipse.org/legal/epl-2.0/
+ * or the Apache License, Version 2.0 which accompanies this distribution and
+ * is available at https://www.apache.org/licenses/LICENSE-2.0.
+ *
+ * This Source Code may also be made available under the following
+ * Secondary Licenses when the conditions for such availability set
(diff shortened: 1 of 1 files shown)


pdbain-ibm:
Just what I need for https://github.com/eclipse/openj9/issues/1659.

test/TestUtilities/src/org/openj9/test/util/VersionCheck.java
```diff
@@ -0,0 +1,68 @@
...
+	static {
+		Object versionInstanceTemp = null;
+		Method majorMethodTemp = null;
+		try { 
+			Method versionMethod = Runtime.class.getDeclaredMethod("version");
+			versionInstanceTemp = versionMethod.invoke(null);
+			majorMethodTemp = versionInstanceTemp.getClass().getDeclaredMethod("major");
+		} catch(NoSuchMethodException e) {
+			// Expected for Java 8
+		} catch (SecurityException | IllegalAccessException | InvocationTargetException e) {
```
fjeremic: Inconsistent formatting on `catch (...` vs. `catch(...`. Cool utility!

smlambert:
Code looks good (ignoring minor spacing inconsistencies).

No tests yet use this VersionCheck.major, so will forego PR testing, unless you want to insert the use of it into the tests referred to in #1659 as part of this PR (to catch 2 birds in one hand, a way to verify this code, and to fix that issue).

pdbain-ibm:
I suggest merging this on spec. Testing of my ad-hoc implementation on Java 8 & 9 is in progress.
If that checks out, we'll give this code a shot with Java 10.

DanHeidinga:
Fixed the spacing inconsistency

Issue #1428, opened 2018-03-13
Add support for specifying heap limit in terms of percentage of total memory


ashu-mehra:
OpenJDK has added options to control heap memory allocations based on percentage of total memory. Details can be found in this [bug report](https://bugs.openjdk.java.net/browse/JDK-8186315). 
The newly added options (`-XX:MaxRAMPercentage -XX:MinRAMPercentage and -XX:InitialRAMPercentage`) are pretty useful for Java applications running in container. Specifying these limits as percentage of total memory, instead of statically setting using -Xmx/-Xms, makes it easy to auto configure the heap settings if container memory is changed.
It would be good for OpenJ9 to support these options as well.

fyi - @charliegracie @dmitripivkine @DanHeidinga @dinogun

ashu-mehra:
I can understand `-XX:MaxRAMPercentage` would specify maximum percentage of RAM that java heap can occupy. Similarly, `-XX:InitialRAMPercentage` would be percentage of RAM that java heap starts with.
But I can't make out what `-XX:MinRAMPercentage` option is for? Does it mean minimum percentage of RAM memory that java heap should occupy? I can't understand why would someone want to specify that. I don't think OpenJ9 has anything like that.
@dmitripivkine your thoughts?

dmitripivkine:
Sounds like -XX:MinRAMPercentage is alias of -XX:InitialRAMPercentage

ashu-mehra:
> Sounds like -XX:MinRAMPercentage is alias of -XX:InitialRAMPercentage

Doesn't look like they are same.
If I just specify `-XX:MinRAMPercentage`, I don't see any change in initial heap size.
With `-XX:InitialRAMPercentage`, initial heap size gets updated as per the value given.
In the examples below, I am running jdk10 in a container with 2G memory limit.

```
# ./bin/java -XX:+UnlockDiagnosticVMOptions -XX:+PrintFlagsFinal -version | grep InitialHeapSize
   size_t InitialHeapSize                          = 33554432                                 {product} {ergonomic}
openjdk version "10" 2018-03-20
OpenJDK Runtime Environment 18.3 (build 10+46)
OpenJDK 64-Bit Server VM 18.3 (build 10+46, mixed mode)

# ./bin/java -XX:MinRAMPercentage=10 -XX:+UnlockDiagnosticVMOptions -XX:+PrintFlagsFinal -version | grep InitialHeapSize
[...]

PR #1699, opened 2018-04-16
Specify heap limits as percentage of total memory

ashu-mehra:
Add options for specifying heap limits in terms of percentage of
total memory available to the JVM.
Following options are added:
1. -XX:MaxRAMPercentage=N where 0 <= N <= 100.0
This option specifies maximum heap size. If -Xmx is specified, this
option is ignored.
2. -XX:InitialRAMPercentage=N where 0 <= N <= 100.0
This option specifies initial heap size. If -Xms is specified, this
option is ignored.

Closes: #1428

### DIFF
--- a/runtime/gc_base/GCExtensions.cpp
+++ b/runtime/gc_base/GCExtensions.cpp
@@ -256,10 +256,8 @@ MM_GCExtensions::computeDefaultMaxHeap(MM_EnvironmentBase *env)
 
 #if defined(OMR_ENV_DATA64)
 	if (J2SE_VERSION((J9JavaVM *)getOmrVM()->_language_vm) >= J2SE_19) {
-		uint64_t usableMemory = omrsysinfo_get_addressable_physical_memory();
-
-		/* extend java default max memory to 25% of physical RAM */
-		memoryMax = OMR_MAX(memoryMax, usableMemory / 4);
+		/* extend java default max memory to 25% of usable RAM */
+		memoryMax = OMR_MAX(memoryMax, usablePhysicalMemory / 4);
 	}
 
 	/* limit maxheapsize up to MAXIMUM_HEAP_SIZE_RECOMMENDED_FOR_3BIT_SHIFT_COMPRESSEDREFS, then can set 3bit compressedrefs as the default */
--- a/runtime/gc_base/GCExtensions.hpp
+++ b/runtime/gc_base/GCExtensions.hpp
@@ -183,6 +183,9 @@ public:
 	MM_IdleGCManager* idleGCManager; /**< Manager which registers for VM Runtime State notification & manages free heap on notification */
 #endif
 
+	double maxRAMPercent; /**< Value of -XX:MaxRAMPercentage specified by the user */
+	double initialRAMPercent; /**< Value of -XX:InitialRAMPercentage specified by the user */
+
 protected:
 private:
 protected:
@@ -299,6 +302,8 @@ public:
 #if defined(J9VM_GC_IDLE_HEAP_MANAGER)
 		, idleGCManager(NULL)
 #endif
+		, maxRAMPercent(0.0) /* this would get overwritten by user specified value */
+		, initialRAMPercent(0.0) /* this would get overwritten by user specified value */
 	{
 		_typeId = __FUNCTION__;
 	}
--- a/runtime/gc_modron_startup/mminit.cpp
+++ b/runtime/gc_modron_startup/mminit.cpp
@@ -144,6 +144,9 @@ static void hookValidatorVMThreadCrash(J9HookInterface * * hookInterface, UDATA 
 static bool gcInitializeVMHooks(MM_GCExtensionsBase *extensions);
 static void gcCleanupVMHooks(MM_GCExtensionsBase *extensions);
 
+static const char * displayXmxOrMaxRAMPercentage(IDATA* memoryParameters);
+static const char * displayXmsOrInitialRAMPercentage(IDATA* memoryParameters);
+
 /**
  * Initialize the threads mutator information (RS pointers, reference list pointers etc) for GC/MM purposes.
  *
@@ -304,7 +307,7 @@ gcCleanupHeapStructures(J9JavaVM * vm)
  * Initialized passive and active heap components
  */
 IDATA
-j9gc_initialize_heap(J9JavaVM *vm, UDATA heapBytesRequested)
+j9gc_initialize_heap(J9JavaVM *vm, IDATA *memoryParameterTable, UDATA heapBytesRequested)
 {
 	MM_GCExtensions *extensions = MM_GCExtensions::getExtensions(vm);
 	MM_EnvironmentBase env(vm->omrVM);
@@ -448,7 +451,7 @@ j9gc_initialize_heap(J9JavaVM *vm, UDATA heapBytesRequested)
 	globalCollector = extensions->configuration->createGlobalCollector(&env);
 	if (NULL == globalCollector) {
 		if(MM_GCExtensionsBase::HEAP_INITIALIZATION_FAILURE_REASON_METRONOME_DOES_NOT_SUPPORT_4BIT_SHIFT == extensions->heapInitializationFailureReason) {
-			j9nls_printf(PORTLIB, J9NLS_ERROR, J9NLS_GC_OPTION_OVERFLOW, "-Xmx");
+			j9nls_printf(PORTLIB, J9NLS_ERROR, J9NLS_GC_OPTION_OVERFLOW, displayXmxOrMaxRAMPercentage(memoryParameterTable));
 		}
 		loadInfo->fatalErrorStr = (char *)j9nls_lookup_message(J9NLS_DO_NOT_PRINT_MESSAGE_TAG | J9NLS_DO_NOT_APPEND_NEWLINE, J9NLS_GC_FAILED_TO_INSTANTIATE_GLOBAL_GARBAGE_COLLECTOR, "Failed to instantiate global garbage collector.");
 		goto error_no_memory;
@@ -902,6 +905,44 @@ setConfigurationSpecificMemoryParameters(J9JavaVM *javaVM, IDATA* memoryParamete
  *
  * Determine which memory parameter to display to the user in the event of an error.
  * @param memoryParameters array of parameter values
+ * @return pointer to "-Xmx" or "-XX:MaxRAMPercentage"
+ */
+static const char *
+displayXmxOrMaxRAMPercentage(IDATA* memoryParameters)
+{
+	if ((-1 != memoryParameters[opt_maxRAMPercent])
+		&& (memoryParameters[opt_Xmx] == memoryParameters[opt_maxRAMPercent])
+	) {
+		return "-Xmx (as set by -XX:MaxRAMPercentage)";
+	} else {
+		return "-Xmx";
+	}
+}
+
+/**
+ * Verify memory parameters.
+ *
+ * Determine which memory parameter to display to the user in the event of an error.
+ * @param memoryParameters array of parameter values
+ * @return pointer to "-Xms" or "-XX:InitialRAMPercentage"
+ */
+static const char *
+displayXmsOrInitialRAMPercentage(IDATA* memoryParameters)
+{
+	if ((-1 != memoryParameters[opt_initialRAMPercent])
+		&& (memoryParameters[opt_Xms] == memoryParameters[opt_initialRAMPercent])
+	) {
+		return "-Xms (as set by -XX:InitialRAMPercentage)";
+	} else {
+		return "-Xms";
+	}
+}
+
+/**
+ * Verify memory parameters.
+ *
+ * Determine which memory parameter to display to the user in the event of an error.
+ * @param memoryParameters array of parameter values
  * @return pointer to "-Xmn" or "-Xmns"
  */
 const char *
@@ -1034,7 +1075,7 @@ gcInitializeXmxXmdxVerification(J9JavaVM *javaVM, IDATA* memoryParameters, bool 
 #endif /* defined(J9ZOS39064) */
 
 	if (extensions->memoryMax > (extensions->heapCeiling - J9GC_COMPRESSED_POINTER_NULL_REGION_SIZE)) {
-		j9nls_printf(PORTLIB, J9NLS_ERROR, J9NLS_GC_OPTION_OVERFLOW, "-Xmx");
+		j9nls_printf(PORTLIB, J9NLS_ERROR, J9NLS_GC_OPTION_OVERFLOW, displayXmxOrMaxRAMPercentage(memoryParameters));
 		return JNI_ERR;
 	}
 #endif /* defined (J9VM_GC_COMPRESSED_POINTERS) */
@@ -1042,12 +1083,12 @@ gcInitializeXmxXmdxVerification(J9JavaVM *javaVM, IDATA* memoryParameters, bool 
 	/* Verify Xmx is too small */
 	if (extensions->memoryMax < minimumSizeValue) {
 		if (NULL == memoryOption1) {
-			memoryOption1 = "-Xmx";
+			memoryOption1 = displayXmxOrMaxRAMPercentage(memoryParameters);
 			memoryOption2 = NULL;
 			goto _subSpaceTooSmallForValue;
 		} else {
 			if (opt_XmxSet) {
-				subSpaceTooLargeOption = "-Xmx";
+				subSpaceTooLargeOption = displayXmxOrMaxRAMPercentage(memoryParameters);
 				goto _subSpaceTooLarge;
 			}
 			goto _subSpaceTooLargeForHeap;
@@ -1071,7 +1112,7 @@ gcInitializeXmxXmdxVerification(J9JavaVM *javaVM, IDATA* memoryParameters, bool 
 			memoryOption1 = "-Xmdx";
 			memoryOption2 = NULL;
 			if (opt_XmxSet) {
-				subSpaceTooLargeOption = "-Xmx";
+				subSpaceTooLargeOption = displayXmxOrMaxRAMPercentage(memoryParameters);
 				goto _subSpaceTooLarge;
 			}
 			goto _subSpaceTooLargeForHeap;
@@ -1180,8 +1221,8 @@ independentMemoryParameterVerification(J9JavaVM *javaVM, IDATA* memoryParameters
 	 * Can not verify minimum values until Xmns/Xmos have been verified.
 	 */
 	if (opt_XmxSet) {
-		maximumXmdxValueParameter = "-Xmx";
-		maximumXmsValueParameter = "-Xmx";
+		maximumXmdxValueParameter = displayXmxOrMaxRAMPercentage(memoryParameters);
+		maximumXmsValueParameter = displayXmxOrMaxRAMPercentage(memoryParameters);
 	}
 
 	if (opt_XmdxSet) {
@@ -1210,7 +1251,7 @@ independentMemoryParameterVerification(J9JavaVM *javaVM, IDATA* memoryParameters
 		}
 
 		if (extensions->initialMemorySize > maximumXmsValue) {
-			memoryOption = "-Xms";
+			memoryOption = displayXmsOrInitialRAMPercentage(memoryParameters);
 			if (maximumXmsValueParameter) {
 				subSpaceTooLargeOption = maximumXmsValueParameter;
 				goto _subSpaceTooLarge;
@@ -1220,7 +1261,7 @@ independentMemoryParameterVerification(J9JavaVM *javaVM, IDATA* memoryParameters
 
 		/* When comparing to maximum boundary use the Xms value */
 		maximumXmsValue = extensions->initialMemorySize;
-		maximumXmsValueParameter = "-Xms";
+		maximumXmsValueParameter = displayXmsOrInitialRAMPercentage(memoryParameters);
 	}
 
 	/* verify -Xsoftmx is set between -Xms and -Xmx */
@@ -1399,11 +1440,11 @@ independentMemoryParameterVerification(J9JavaVM *javaVM, IDATA* memoryParameters
 
 	if (opt_XmsSet && (extensions->initialMemorySize < XmsMinimumValue)) {
 		if (NULL == memoryOption) {
-			memoryOption = "-Xms";
+			memoryOption = displayXmsOrInitialRAMPercentage(memoryParameters);
 			minimumSizeValue = XmsMinimumValue; /* display min size */
 			goto _subSpaceTooSmallForValue;
 		} else {
-			subSpaceTooLargeOption = "-Xms";
+			subSpaceTooLargeOption = displayXmsOrInitialRAMPercentage(memoryParameters);
 			goto _subSpaceTooLarge;
 		}
 	}
@@ -1543,7 +1584,7 @@ combinationMemoryParameterVerification(J9JavaVM *javaVM, IDATA* memoryParameters
 		assume0(extensions->initialMemorySize <= extensions->maxSizeDefaultMemorySpace);
 
 		maximumXmsValue = extensions->initialMemorySize;
-		maximumXmsValueParameter = "-Xms";
+		maximumXmsValueParameter = displayXmsOrInitialRAMPercentage(memoryParameters); 
 		break;
 
 	case XMDX:
@@ -1567,7 +1608,7 @@ combinationMemoryParameterVerification(J9JavaVM *javaVM, IDATA* memoryParameters
 		}
 
 		maximumXmsValue = extensions->initialMemorySize;
-		maximumXmsValueParameter = "-Xms";
+		maximumXmsValueParameter = displayXmsOrInitialRAMPercentage(memoryParameters);
 		break;
 
 	case NONE:
@@ -1578,7 +1619,7 @@ combinationMemoryParameterVerification(J9JavaVM *javaVM, IDATA* memoryParameters
 
 		maximumXmsValue = extensions->memoryMax;
 		if (opt_XmxSet) {
-			maximumXmsValueParameter = "-Xmx";
+			maximumXmsValueParameter = displayXmxOrMaxRAMPercentage(memoryParameters);
 		}
 		break;
 
@@ -1622,7 +1663,7 @@ combinationMemoryParameterVerification(J9JavaVM *javaVM, IDATA* memoryParameters
 			 */
 			if (extensions->oldSpaceSize != extensions->initialMemorySize) {
 				memoryOption = displayXmoOrXmos(memoryParameters);
-				memoryOption2 = "-Xms";
+				memoryOption2 = displayXmsOrInitialRAMPercentage(memoryParameters);
 				goto _subSpaceNotEqualError;
 			}
 			break;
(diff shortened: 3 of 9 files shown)


pshipton:
The description should contain -XX:MaxRAMPercentage instead of -XX:MaxRAMPercent, same for -XX:InitialRAMPercentage

ashu-mehra:
> The description should contain -XX:MaxRAMPercentage instead of -XX:MaxRAMPercent, same for -XX:InitialRAMPercentage

Updated the comment, as well as the commit message.

dmitripivkine:
Discussion question: Do we need full (signed) double parsing for this particular case? If  we do it unsigned would be no underflow case to handle.

runtime/gc_modron_startup/mmparse.cpp
```diff
@@ -1443,6 +1477,31 @@ gcParseCommandLineAndInitializeWithValues(J9JavaVM *vm, IDATA *memoryParameters)
...
+				goto _error;
+			} else {
+				/* Use -XX:MaxRAMPercentage only if -Xmx is not specified */
+				if (memoryParameters[opt_Xmx] == -1) {
+					OMRPORT_ACCESS_FROM_J9PORT(vm->portLibrary);
+					uint64_t usableMemory = omrsysinfo_get_addressable_physical_memory();
+					extensions->memoryMax = (usableMemory / 100) * maxRAMPercent;
+
+					/* Hack table to appear that -Xmx was specified */
+					memoryParameters[opt_Xmx] = index;
```
dmitripivkine: I don't like this solution. I believe the option should be parsed and stored in gc extensions. MemoryMax and initialMemorySize should be adjusted in mminit. I would try to avoid any decisions in parse module

runtime/gc_modron_startup/mmparse.cpp
```diff
@@ -1500,6 +1559,30 @@ gcParseCommandLineAndInitializeWithValues(J9JavaVM *vm, IDATA *memoryParameters)
...
+				goto _error;
+			} else {
+				/* Use -XX:InitialRAMPercentage only if -Xms is not specified */
+				if (memoryParameters[opt_Xms] == -1) {
+					OMRPORT_ACCESS_FROM_J9PORT(vm->portLibrary);
+					uint64_t usableMemory = omrsysinfo_get_addressable_physical_memory();
+					extensions->initialMemorySize = (usableMemory / 100) * initialRAMPercent;
+
+					/* Hack table to appear that -Xms was specified */
+					memoryParameters[opt_Xms] = index;
```
dmitripivkine: the same as comment for MemoryMax

ashu-mehra:
> Do we need full (signed) double parsing for this particular case? If we do it unsigned would be no underflow case to handle. If we need to have general double parsing I am ok with it

I have updated the code to avoid signed parsing.

@dmitripivkine @DanHeidinga can you please review the changes.

ashu-mehra:
I have removed WIP from the subject line.

runtime/gc_modron_startup/mminit.cpp
```diff
@@ -2830,6 +2833,24 @@ gcInitializeDefaults(J9JavaVM* vm)
 	return J9VMDLLMAIN_FAILED;
 }
 
+static void
+adjustXmxXmsUsingRAMPercent(MM_GCExtensions *extensions, IDATA *memoryParameters)
+{
+	OMRPORT_ACCESS_FROM_OMRVM(extensions->getOmrVM());
+	uint64_t usableMemory = omrsysinfo_get_addressable_physical_memory();
```
dmitripivkine: This function is called already in MM_GCExtensions::computeDefaultMaxHeap. In my opinion you should integrate handling of maxRAMPercent and initialRAMPercent in that code. Please not modify memoryParameters[opt_Xms] and memoryParameters[opt_Xmx]. Let them keep original command line option values if any.

ashu-mehra: > In my opinion you should integrate handling of maxRAMPercent and initialRAMPercent in that code.

But computeDefaultMaxHeap() is called even before we parse gc options in gcParseCommandLineAndInitializeWithValues().

> Please not modify memoryParameters[opt_Xms] and memoryParameters[opt_Xmx]. 

Reason for modifying `memoryParameters[opt_Xms]` is the code in `gcInitializeCalculatedValues()` which sets `initialMemorySize` to default value if `memoryParameters[opt_Xms]` is not set.
So If I set `initialMemorySize` using initialRAMPercentage but don't update `memoryParameters[opt_Xms]`, it will get overwritten to default value in `gcInitializeCalculatedValues()`.

dmitripivkine: Right, I see. Let me think about it a little bit.

runtime/gc_modron_startup/mmparse.cpp
```diff
@@ -1573,6 +1604,30 @@ gcParseCommandLineAndInitializeWithValues(J9JavaVM *vm, IDATA *memoryParameters)
...
 	}
 
+	result = option_set_to_opt_double(vm, OPT_XXMAXRAMPERCENT, &index, EXACT_MEMORY_MATCH, &extensions->maxRAMPercent);
+	if (OPTION_OK != result) {
+		goto _error;
+	}
+	memoryParameters[opt_maxRAMPercent] = index;
+	if (memoryParameters[opt_maxRAMPercent] != -1) {
+		if (extensions->maxRAMPercent > 100) {
+			result = OPTION_OVERFLOW;
```
dmitripivkine: Can we reuse J9NLS_GC_OPTIONS_PERCENT_OUT_OF_RANGE here? The only issue I see is current output format is too few digits after decimal mark:
J9NLS_GC_OPTIONS_PERCENT_OUT_OF_RANGE=%1$s value must be between %2$.2f and %3$.2f (inclusive)

runtime/gc_base/GCExtensions.cpp
```diff
@@ -253,14 +253,12 @@ MM_GCExtensions::computeDefaultMaxHeap(MM_EnvironmentBase *env)
...
-
-		/* extend java default max memory to 25% of physical RAM */
-		memoryMax = OMR_MAX(memoryMax, usableMemory / 4);
+		/* extend java default max memory to 25% of usable RAM */
+		memoryMax = (uintptr_t)OMR_MAX(memoryMax, usablePhysicalMemory / 4);
 	}
 
 	/* limit maxheapsize up to MAXIMUM_HEAP_SIZE_RECOMMENDED_FOR_3BIT_SHIFT_COMPRESSEDREFS, then can set 3bit compressedrefs as the default */
-	memoryMax = OMR_MIN(memoryMax, MAXIMUM_HEAP_SIZE_RECOMMENDED_FOR_3BIT_SHIFT_COMPRESSEDREFS);
+	memoryMax = (uintptr_t)OMR_MIN(memoryMax, MAXIMUM_HEAP_SIZE_RECOMMENDED_FOR_3BIT_SHIFT_COMPRESSEDREFS);
```
dmitripivkine: Why we need this cast?

ashu-mehra: right, that's not required. I will remove this cast.

ashu-mehra:
Not related to this change but I think is a bug in `MM_GCExtensions::computeDefaultMaxHeap()`. On 32-bit JVM running on 64-bit system, cgroup memory limit can go well beyond 32-bit data type, in which case following code will suffer from data loss due to narrow conversion:
```
        if (0 == rc) {
                /* If running in a cgroup with memory limit > 1G, reserve at-least 512M for JVM's internal requirements
                 */
                memoryMax = (uintptr_t)OMR_MAX((int64_t)(cgroupMemory / 2), (int64_t)(cgroupMemory - OPENJ9_IN_CGROUP_NATIVE_FOOTPRINT_EXCLUDING_HEAP));
                memoryMax = (uintptr_t)OMR_MIN(memoryMax, (cgroupMemory / 4) * 3);
        }
```
I will do some test to verify this scenario.
I think we need to apply address space limit on cgroup memory as well to make sure 32-bit JVM works fine with large cgroup memory.

dmitripivkine:
@charliegracie FYI

ashu-mehra:
Updated the commit as suggested.

runtime/gc_modron_startup/mminit.cpp
```diff
@@ -2719,6 +2719,17 @@ gcInitializeDefaults(J9JavaVM* vm)
 		goto error;
 	}
 
+	if ((-1 == memoryParameterTable[opt_Xms]) && (-1 != memoryParameterTable[opt_initialRAMPercent])) {
+		extensions->initialMemorySize = (uintptr_t)(((double)extensions->usablePhysicalMemory / 100) * extensions->initialRAMPercent);
```
dmitripivkine: Would be better if "100" will be in FP representation as well "100." ?

runtime/gc_modron_startup/mminit.cpp
```diff
@@ -2719,6 +2719,17 @@ gcInitializeDefaults(J9JavaVM* vm)
...
+		/* Update memory parameter table to appear that -Xms was specified */
+		memoryParameterTable[opt_Xms] = memoryParameterTable[opt_initialRAMPercent];
+	}
+	if ((-1 == memoryParameterTable[opt_Xmx]) && (-1 != memoryParameterTable[opt_maxRAMPercent])) {
+		extensions->memoryMax = (uintptr_t)(((double)extensions->usablePhysicalMemory / 100) * extensions->maxRAMPercent);
```
dmitripivkine: same as above

runtime/gc_modron_startup/mmparse.cpp
```diff
@@ -1573,6 +1604,30 @@ gcParseCommandLineAndInitializeWithValues(J9JavaVM *vm, IDATA *memoryParameters)
 		memoryParameters[opt_Xmox] = memoryParameters[opt_Xmo];
 	}
 
+	result = option_set_to_opt_double(vm, OPT_XXMAXRAMPERCENT, &index, EXACT_MEMORY_MATCH, &extensions->maxRAMPercent);
+	if (OPTION_OK != result) {
+		goto _error;
+	}
+	memoryParameters[opt_maxRAMPercent] = index;
+	if (memoryParameters[opt_maxRAMPercent] != -1) {
+		if (extensions->maxRAMPercent > 100) {
```
dmitripivkine: please use 100.0

runtime/gc_modron_startup/mmparse.cpp
```diff
@@ -1573,6 +1604,30 @@ gcParseCommandLineAndInitializeWithValues(J9JavaVM *vm, IDATA *memoryParameters)
...
+		}
+	}
+
+	result = option_set_to_opt_double(vm, OPT_XXINITIALRAMPERCENT, &index, EXACT_MEMORY_MATCH, &extensions->initialRAMPercent);
+	if (OPTION_OK != result) {
+		goto _error;
+	}
+	memoryParameters[opt_initialRAMPercent] = index;
+	if (memoryParameters[opt_initialRAMPercent] != -1) {
+		if (extensions->initialRAMPercent > 100) {
```
dmitripivkine: please use 100.0

runtime/util/vmargs.c
```diff
@@ -346,6 +346,8 @@ optionValueOperations(J9PortLibrary *portLibrary, J9VMInitArgs* j9vm_args, IDATA
 	UDATA value, oldValue, done;
 	IDATA errorFound = 0;
 	UDATA mapType;
+	double dvalue = 0;
```
dmitripivkine: Should it be FP 0.0 ?

runtime/oti/jvminit.h
```diff
@@ -128,6 +129,7 @@ extern "C" {
 #define GET_INTEGER_VALUE(element, optname, result) vm->internalVMFunctions->optionValueOperations(vm->portLibrary, vm->vmArgsArray, element, GET_INT_VALUE, (char **)&optname, 0, 0, 0, &result)
 #define GET_PERCENT_VALUE(element, optname, result) vm->internalVMFunctions->optionValueOperations(vm->portLibrary, vm->vmArgsArray, element, GET_PRC_VALUE,(char **) &optname, 0, 0, 0, &result)
 #define GET_OPTION_OPTION(element, delimChar, delimChar2, resultPtr) vm->internalVMFunctions->optionValueOperations(vm->portLibrary, vm->vmArgsArray, element, GET_OPTION_OPT, resultPtr, 0, delimChar, delimChar2, NULL)
+#define GET_DOUBLE_VALUE(element, optname, result) vm->internalVMFunctions->optionValueOperations(vm->portLibrary, vm->vmArgsArray, element, GET_DBL_VALUE, (char **) &optname, 0, 0, 0, &result)
```
dmitripivkine: There is a good practice to have extra pair if brackets around every of input parameters for MACRO. It would prevent potential wrong order of operations. For example for this particular line it would like as:
\#define GET_DOUBLE_VALUE(element, optname, result) vm->internalVMFunctions->optionValueOperations(vm->portLibrary, vm->vmArgsArray, (element), GET_DBL_VALUE, (char **) &(optname), 0, 0, 0, &(result))
However for consistency all this definitions should be changed. Leaving final decision for Dan

runtime/util/vmargs.c
```diff
@@ -506,10 +508,16 @@ optionValueOperations(J9PortLibrary *portLibrary, J9VMInitArgs* j9vm_args, IDATA
...
 				break;
 
+			case GET_DBL_VALUE:
 			case GET_MEM_VALUE :
 			case GET_INT_VALUE :
 			case GET_PRC_VALUE :
-				*((UDATA*)(reserved)) = 0;				/* Initialize the return value */
+				/* Initialize the return value */
+				if (GET_DBL_VALUE == action) {
+					*(double *)reserved = 0;
```
dmitripivkine: at least should be 0.0
In general:
1)This piece looks ugly.
2)Also (unrelated to this PR): there are many calls of this function with NULL for "reserved".
So this code assumed that for actions listed above "reserved" must be not NULL otherwise it crashed.
Would it be better to add assertion here as well?

runtime/util_core/j9argscan.c
```diff
@@ -103,6 +105,39 @@ void scan_failed(J9PortLibrary * portLibrary, const char* module, const char *sc
...
 
+/* Scan the next unsigned number off of the argument string.
+ * Store the result in *result
+ * Answer 0 on success
+ */
+uintptr_t scan_double(char **scan_start, double *result)
+{
+	char *endPtr = NULL;
+
+	*result = 0;
```
dmitripivkine: should be 0.0

runtime/util_core/j9argscan.c
```diff
@@ -103,6 +105,39 @@ void scan_failed(J9PortLibrary * portLibrary, const char* module, const char *sc
...
+
+	/* Check for '-' at the beginning to handle negative numbers,
+	 * as strtod() does not fail for negative numbers.
+	 */
+	if (*scan_start == '-') {
+		return 1;
+	}
+	*result = strtod(*scan_start, &endPtr);
+	if (ERANGE == errno) {
+		if (0 == *result) {
```
dmitripivkine: should be 0.0

runtime/util_core/j9argscan.c
```diff
@@ -103,6 +105,39 @@ void scan_failed(J9PortLibrary * portLibrary, const char* module, const char *sc
...
+			/* underflow - value is so small that it cannot be represented as double.
+			 * treat it as zero.
+			*/
+			return 0;
+		}
+		/* overflow */
+		return 2;
+	} else if ((0 == *result) && (endPtr == *scan_start)) {
```
dmitripivkine: should be 0.0

runtime/util_core/j9argscan.c
```diff
@@ -103,6 +105,39 @@ void scan_failed(J9PortLibrary * portLibrary, const char* module, const char *sc
...
+	char *endPtr = NULL;
+
+	*result = 0;
+
+	/* Check for '-' at the beginning to handle negative numbers,
+	 * as strtod() does not fail for negative numbers.
+	 */
+	if (*scan_start == '-') {
+		return 1;
+	}
```
dmitripivkine: If we pretend we scan doubles should negative part be handled? I know that for this case we want positive double to prevent check and introducing new NLS message for negative values. However in my opinion this is kind of misleading.
Would it be better if:
- scanning would support signed doubles 
- check at parsing would be 0<=value<=100
- error message would say "out of range" instead of "overflow"

ashu-mehra: I agree with your comment, but I am not sure if we should check 0<=value<=100 during option parsing, as some other option may allow double value greater than 100.
I feel `optionValueOperations` should ideally be refactored a bit with at least following changes:
1) allow callers to explicitly pass the range of acceptable values, if there are any. Then the parser can check the value against this range. With this change, we wouldn't have to hardcode the range during parsing.
2) introduce a structure that aggregates all the parameters being passed to `optionValueOperations`.
3) break it up into smaller functions wherever possible to make it easier to read and comprehend.
It is something not related to take up in this PR. I can give this a shot later if time permits.

dmitripivkine: Sorry, I was not clear. When I mentioned 0<=value<=100 I told about GC options parsing in mmparse.cpp. Keep optionValueOperations parse double generally (any valid value). Later on, for example in in mmparse.cpp in this case, you can check for valid range for particular option and print a NLS message in the case of mismatch. I would not change optionValueOperations with your's 1-2-3

dmitripivkine: Seems we put update simultaneously, so I did not see your comment.
Yes, parse a general signed double. For each particular option check a range and print NLS message in case it is out of.

ashu-mehra:
I have updated the commit to display out of range message in mmparse.cpp for negative values of the options.
Please review.

runtime/gc_base/GCExtensions.hpp
```diff
@@ -183,6 +183,9 @@ class MM_GCExtensions : public MM_GCExtensionsBase {
 	MM_IdleGCManager* idleGCManager; /**< Manager which registers for VM Runtime State notification & manages free heap on notification */
 #endif
 
+	double maxRAMPercent; /**< Value of -XX:MaxRAMPercentage specified by the user */
+	double initialRAMPercent; /**< Value of -XX:InitialRAMPercentage specified by the user */
+
```
dmitripivkine: Please add an initialization of this variables in constructor

runtime/gc_modron_startup/mmparse.cpp
```diff
@@ -1573,6 +1604,30 @@ gcParseCommandLineAndInitializeWithValues(J9JavaVM *vm, IDATA *memoryParameters)
...
 	}
 
+	result = option_set_to_opt_double(vm, OPT_XXMAXRAMPERCENT, &index, EXACT_MEMORY_MATCH, &extensions->maxRAMPercent);
+	if (OPTION_OK != result) {
+		goto _error;
+	}
+	memoryParameters[opt_maxRAMPercent] = index;
+	if (memoryParameters[opt_maxRAMPercent] != -1) {
+		if ((extensions->maxRAMPercent < 0.0) || (extensions->maxRAMPercent > 100.0)) {
+			j9nls_printf(PORTLIB, J9NLS_ERROR, J9NLS_GC_OPTIONS_DOUBLE_OUT_OF_RANGE, "-XX:MaxRAMPercentage", 0.0, 100.0);
```
dmitripivkine: Can we reuse J9NLS_GC_OPTIONS_PERCENT_OUT_OF_RANGE ? Currently there is no need to have general "Double out-of-range" message.

runtime/oti/jvminit.h
dmitripivkine: Please put extra pair of brackets around each of input parameters for all four definitions

runtime/util/vmargs.c
```diff
@@ -506,10 +508,16 @@ optionValueOperations(J9PortLibrary *portLibrary, J9VMInitArgs* j9vm_args, IDATA
 				done = TRUE;
 				break;
 
+			case GET_DBL_VALUE:
```
dmitripivkine: Adding GET_DBL_VALUE option here makes code look ugly. Would you please try to implement independent case for GET_DBL_VALUE. Duplicated code like scanStart setup can be moved to separate function(s) and be called in both places. I believe it would be easier to read.

runtime/util_core/j9argscan.c
```diff
@@ -103,6 +105,31 @@ void scan_failed(J9PortLibrary * portLibrary, const char* module, const char *sc
...
+ * Store the result in *result
+ * Answer 0 on success
+ */
+uintptr_t scan_double(char **scan_start, double *result)
+{
+	char *endPtr = NULL;
+
+	*result = strtod(*scan_start, &endPtr);
+	if (ERANGE == errno) {
+		if (0.0 == *result) {
```
dmitripivkine: You can not do it according spec: http://www.cplusplus.com/reference/cstdlib/strtod/
in case of ERANGE if return is not equal +/- HUGE_VAL this is underflow. The output value in this case is "value whose magnitude is no greater than the smallest normalized positive number" (not necessarily 0.0). If you want to handle it as zero *result should be set 0.0 before return with rc 0.

runtime/gc_modron_startup/mminit.cpp
```diff
@@ -2719,6 +2719,17 @@ gcInitializeDefaults(J9JavaVM* vm)
 		goto error;
 	}
 
+	if ((-1 == memoryParameterTable[opt_Xms]) && (-1 != memoryParameterTable[opt_initialRAMPercent])) {
+		extensions->initialMemorySize = (uintptr_t)(((double)extensions->usablePhysicalMemory / 100.0) * extensions->initialRAMPercent);
+		/* Update memory parameter table to appear that -Xms was specified */
+		memoryParameterTable[opt_Xms] = memoryParameterTable[opt_initialRAMPercent];
```
dmitripivkine: This means that in case of error related to out-of-range case (<1m for example) output message would mention -Xms option. This fact should be documented at least.

runtime/gc_modron_startup/mminit.cpp
```diff
@@ -2719,6 +2719,17 @@ gcInitializeDefaults(J9JavaVM* vm)
...
+	}
+	if ((-1 == memoryParameterTable[opt_Xmx]) && (-1 != memoryParameterTable[opt_maxRAMPercent])) {
+		extensions->memoryMax = (uintptr_t)(((double)extensions->usablePhysicalMemory / 100.0) * extensions->maxRAMPercent);
+		/* Update memory parameter table to appear that -Xmx was specified */
+		memoryParameterTable[opt_Xmx] = memoryParameterTable[opt_maxRAMPercent];
```
dmitripivkine: This means that in case of error related to out-of-range case output message would mention -Xmx option. This fact should be documented at least.

runtime/util_core/j9argscan.c
ashu-mehra: As per man page of strtod:
*If the correct value would cause overflow, plus or minus HUGE_VAL (HUGE_VALF, HUGE_VALL) is returned (according to the sign of the value), and ERANGE is stored in errno. **If the correct value would cause underflow, zero is returned and ERANGE is stored in errno.** *
Based on that I consider it as underflow if `*return` is 0.0.

runtime/gc_modron_startup/mminit.cpp
ashu-mehra: How about updating the message to say " -Xmx (as set by -XX:InitialRAMPercentage)"  when -XX:MaxRAMPercentage is being used to set max heap size. Some examples of updated messages:
* Option too large: -Xmx (as set by -XX:InitialRAMPercentage) is too large
* -Xmx (as set by -XX:InitialRAMPercentage) too small, must be at least 1 Mbytes

I can add a function that returns "-Xmx" or "-Xmx (as set by -XX:InitialRAMPercentage) " based on the option being used:

```
const char *
displayXmsOrInitialRAMPercentage(IDATA* memoryParameters)
{
        if ((-1 == memoryParameters[opt_initialRAMPercent])
                && (memoryParameters[opt_Xms] == memoryParameters[opt_initialRAMPercent])
        ) {
                return "-Xms (as set by -XX:InitialRAMPercentage)";
        } else {
                return "-Xms";
        }
}
```
and we can use this function wherever we are currently displaying "-Xmx".
Does that look fine?

dmitripivkine: yes, something like this

runtime/util_core/j9argscan.c
dmitripivkine: From: http://pubs.opengroup.org/onlinepubs/007904975/functions/strtod.html
The Open Group Base Specifications Issue 6 IEEE Std 1003.1, 2004 Edition

"If the correct value would cause an underflow, a value whose magnitude is no greater than the smallest normalized positive number in the return type shall be returned and errno set to [ERANGE]."

ashu-mehra:
@dmitripivkine I updated the code as suggested. Can you please review it.

runtime/util/vmargs.c
```diff
@@ -629,6 +639,19 @@ optionValueOperations(J9PortLibrary *portLibrary, J9VMInitArgs* j9vm_args, IDATA
...
+getOptionValue(J9VMInitArgs* j9vm_args, IDATA element, const char *optionName)
+{
+	char *optionValue = NULL;
+	const char *option = optionName;
+
+	if (HAS_MAPPING(j9vm_args, element)) {
+		option = MAPPING_MAPNAME(j9vm_args, element);
+	}
+
+	return j9vm_args->actualVMArgs->options[element].optionString + strlen(option);
```
dmitripivkine: Is check for (NULL == option) missed there? I believe strlen(NULL) will crash, is not it?
getOptionValue() can return NULL in this case and caller might check it to return OPTION_ERROR

ashu-mehra: I added check for `(NULL == option)`.
The current users of `getOptionValue()` don't need to check for NULL return value as we already ensure that `optionName` is not NULL.

runtime/util/vmargs.c
```diff
@@ -506,22 +510,28 @@ optionValueOperations(J9PortLibrary *portLibrary, J9VMInitArgs* j9vm_args, IDATA
...
 			case GET_PRC_VALUE :
-				*((UDATA*)(reserved)) = 0;				/* Initialize the return value */
-				if ( HAS_MAPPING(j9vm_args, element) ) {
-					/* If we have a mapping, then we ignore the option name passed in. Instead we want the one that corresponds to the actual cmd line */
-					scanStart = &(j9vm_args->actualVMArgs->options[element].optionString[ strlen(MAPPING_MAPNAME(j9vm_args, element)) ]);
-				} else {
-					if (*valuesBuffer) {
-						scanStart = &(j9vm_args->actualVMArgs->options[element].optionString[ strlen(*valuesBuffer) ]);
-					} else {
-						return OPTION_ERROR;
```
dmitripivkine: For reviewers: this "else" is not necessary because of check lines 363-364, it is removed in new code

dmitripivkine:
@DanHeidinga Would you please do final review?

dmitripivkine:
Also this feature should be documented.

dmitripivkine:
Note: Currently Java 8 does not take the amount of RAM to consideration to calculate -Xmx default. It is applicable for Java 9 and higher. The -XX:MaxRAMPercentage option however is applicable for Java 8 so it introduces a way to set -Xmx as a percentage of machine's RAM.

runtime/util/vmargs.c
```diff
@@ -629,6 +639,23 @@ optionValueOperations(J9PortLibrary *portLibrary, J9VMInitArgs* j9vm_args, IDATA
 	return OPTION_OK;
 }
 
+static char * 
+getOptionValue(J9VMInitArgs* j9vm_args, IDATA element, const char *optionName)
```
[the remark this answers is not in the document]
ashu-mehra: In the absence of a better name, lets go with `getStartOfOptionValue`.

DanHeidinga:

DanHeidinga:

DanHeidinga:
Looks like the CI server is having issues.  Once we can build this it can be merged

PR #1700, opened 2018-04-16
Optimized Unsafe getAndAdd and getAndSet on X86

0dvictor:
Unsafe's methods getAndAdd and getAndSet are recognized and
transformed into OMR's corresponding helpers, allowing being
inlined by the CodeGen.

### DIFF
--- a/runtime/compiler/codegen/J9RecognizedMethodsEnum.hpp
+++ b/runtime/compiler/codegen/J9RecognizedMethodsEnum.hpp
@@ -371,6 +371,10 @@
    sun_misc_Unsafe_staticFieldBase,
    sun_misc_Unsafe_staticFieldOffset,
    sun_misc_Unsafe_objectFieldOffset,
+   sun_misc_Unsafe_getAndAddInt,
+   sun_misc_Unsafe_getAndSetInt,
+   sun_misc_Unsafe_getAndAddLong,
+   sun_misc_Unsafe_getAndSetLong,
 
    sun_misc_Unsafe_putBooleanOrdered_jlObjectJZ_V,
    sun_misc_Unsafe_putByteOrdered_jlObjectJB_V,
--- a/runtime/compiler/env/j9method.cpp
+++ b/runtime/compiler/env/j9method.cpp
@@ -3150,6 +3150,10 @@ TR_ResolvedJ9Method::TR_ResolvedJ9Method(TR_OpaqueMethodBlock * aMethod, TR_Fron
       {x(TR::sun_misc_Unsafe_staticFieldBase,               "staticFieldBase",   "(Ljava/lang/reflect/Field;)Ljava/lang/Object")},
       {x(TR::sun_misc_Unsafe_staticFieldOffset,             "staticFieldOffset", "(Ljava/lang/reflect/Field;)J")},
       {x(TR::sun_misc_Unsafe_objectFieldOffset,             "objectFieldOffset", "(Ljava/lang/reflect/Field;)J")},
+      {x(TR::sun_misc_Unsafe_getAndAddInt,                  "getAndAddInt",      "(Ljava/lang/Object;JI)I")},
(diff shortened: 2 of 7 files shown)


0dvictor:
This PR requires https://github.com/eclipse/omr/pull/2636 and https://github.com/eclipse/openj9/pull/2083

andrewcraik:
@0dvictor depedent changes ahve merged could you rebase and un-WIP once this is ready?

andrewcraik:
@0dvictor any update on this one?

0dvictor:
I'm still in the process of re-testing the changes to ensure it is still functional. Will remove the WIP as soon as the changes have been re-verified.

0dvictor:
The change has been re-verified and is ready for merging.

andrewcraik:

andrewcraik:
Windows failure due to #2129

andrewcraik:
@0dvictor couple of small things: 1) since we have a pass for handling unsafe why did you choose to put these methods in the RecognizedCallTransformer rather than the unsafe fast pathing? 2) there are the disableUnsafe and disableUnsafeFastPath options - can we make sure your changes respect these since they have proven to be great debugging tools when things have gone wrong 3) some of the code for the diamonds in here looks a lot like what is in UnsafeFastPath but doesn't handle the generation of SPINECHKs for arraylets - would moving your code to UnsafeFastPath let you reuse some of this arraylet support so we don't have to disable this stuff under arraylets - such a disable is not very desirable...

0dvictor:
Thank-you for the valuable opinions.
1) I believe these methods fit RecognizedCallTransformer better as they are (i) recognized methods that IL Gen transforms from a call to a sets of trees (ii) they are unconditionally transformed. All methods handled by UnsafeFastPath only conditionally transforms selected call sites based on the callers. Although this change use its caller information to OPTIMIZE generated trees, it is still a generic transformation that performs on every related Unsafe methods;
2) Good suggestion, I've updated the code and it now respects disableUnsafe; however, it does not response to disableUnsafeFastPath as it is conceptional not a "UnsafeFastPath". I believe this is sufficient for debugging;
3) I don't see the any diamond in UnsafeFastPath. UnsafeFastPath completely avoided the need of such a diamond based on the information of its call stack. I acknowledge the diamond code shares great similarity with the two diamond generation code in InlinerTempForJ9; however, InlinerTempForJ9's diamond code is so spaghetti that they could not even consolidate the cases of CompareAndSwap and get/put; therefore it need be refactored. In fact this changes are designed with the mind of replacing the spaghetti in InlinerTempForJ9, I've tested several methods include CompareAndSwap. I chose not commit them in the same Pull Request to ease the review and make it less difficult to isolate problems should anything goes wrong in the future;
3.1) For arraylet support: as mentioned earlier, there are currently two optimizations that transfers an Unsafe call: UnsafeFastPath for selected ones based on call stack information, and InlinerTempForJ9 for arbitrary ones. This proposed changes are designed to transform arbitrary Unsafe calls and hence mimics InlinerTempForJ9. Therefore it does not handle arraylets, neither does InlinerTempForJ9. I fully understand the importance of arraylets, and hence I will add such supports in subsequent Pull Requests. I prefer to keep this first change as a simple and clear implementation to replace Unsafe related code in InlinerTempForJ9; extra feature supports will be in subsequent Pull Requests.

andrewcraik:
Thank you for the consideration of the feedback - if there is a follow-on pull request to support arraylets then I think this is a reasonable change to make - I agree the current code isn't well factored - it would be nice to clean it up at some point as part of the future pull request.

andrewcraik:
Jenkins test sanity

0dvictor:
There were Travis CI failures caused by out-of-dated OMR; hence I've rebased and re-launched Travis CI. All checks have passed now, could you re-launch sanity tests? @andrewcraik

andrewcraik:
Jenkins test sanity

0dvictor:
Windows failures due to #2129

0dvictor:
Conflicts have been resolved. The changes have been rebased and re-verified.

andrewcraik:
Jenkins test sanity

PR #1705, opened 2018-04-17
Support DDR on Windows

keithc-ca:
Enable DDR for Windows platforms

* adjust timing of ddr processing
* export a symbol from j9ddr_misc
* fix dependency on ddrgen[.exe]
* expand blacklist and type overrides for Windows
* separate compiler-related typeoverrides
* when generating pointer classes, signal failure on any error, but don't stop at the first error

Clean up forward type references

* consolidate forward declarations
* add missing/required forward declarations


debugtools/DDR_VM/src/com/ibm/j9ddr/tools/PointerGenerator.java
```diff
@@ -876,12 +884,23 @@ private String getPointerType(String type) {
...
+		try {
+			String returnType = getArrayType(fieldDescriptor.getType());
+			if (returnType.equals("EnumPointer")) {
+				writeEnumEAMethod(writer, returnType, structure, fieldDescriptor);
+			} else {
+				writeEAMethod(writer, returnType, structure, fieldDescriptor);
+			}
+		} catch (Exception e) {
+			errorCount += 1;
+			if (errorCount < 100) {
```
pshipton: Please add a comment to explain the purpose of throwing up to 100 errors.

runtime/ddr/blacklist
```diff
@@ -27,6 +27,44 @@ file:<built-in>
 # zlib defines the type 'Byte' which conflicts with java.lang.Byte.
 file:*/zlib/zconf.h
 
-# Ignore types that are troublesome or not required. 
+# Ignore types that are troublesome or not required.
 type:__gnu*
 type:*__va_list*
+type:*vc_attributes*
+type:*<*
+type:tag*
```
pshipton: what is there is a type starting with tag we don't want to ignore?

runtime/ddr/blacklist
```diff
@@ -27,6 +27,44 @@ file:<built-in>
...
+# Ignore one (or all) of sets of types whose names differ only in case
+# (so we don't try to create Java source files whose names differ only
+# in case which isn't possible with case-insensitive file-systems).
+
+type:ADDRINFOW
+type:addrinfoW
+type:BYTE
+type:Byte
+type:byte
+type:classTableEntry
```
pshipton: should fix this one instead of excluding as this type may be useful for debugging

pshipton: I think I misunderstood. There is still a structure included with the same name, just not with this case.

pshipton: Maybe it doesn't matter given this structure is a union of different pointer types.

runtime/ddr/blacklist
```diff
@@ -27,6 +27,44 @@ file:<built-in>
...
+type:IN6_ADDR
+type:IN_ADDR
+type:SIZE_T
+type:SOCKADDR*
+type:UINT
+type:uInt
+type:ULONG
+type:uLong
+type:UserDataWrapper
+type:userDataWrapper
```
pshipton: Do both versions need to be excluded? We should keep one for debugging purposes.

runtime/ddr/run_omrddrgen.mk.ftl
```diff
@@ -27,10 +27,8 @@ include $(TOP_DIR)makelib/uma_macros.mk
 
 ifeq (8,$(VERSION_MAJOR))
   DDR_JCL_MODULE := jclse7b_
-else ifneq (,$(filter 9 10 11,$(VERSION_MAJOR)))
-  DDR_JCL_MODULE := jclse$(VERSION_MAJOR)_
 else
-  $(error Unsupported version: '$(VERSION_MAJOR)')
+  DDR_JCL_MODULE := jclse$(VERSION_MAJOR)_
```
pshipton: This may not work in the future, but I suppose it can be fixed when it breaks. We only made 9, 10, 11 versions because something changed which required a new version. This may not always be the case.

debugtools/DDR_VM/src/com/ibm/j9ddr/tools/PointerGenerator.java
keithc-ca: Done.

runtime/ddr/blacklist
keithc-ca: I think the majority of these are due to this pattern:
```
typedef struct tagSample { /*...*/ } Sample;
```
We will have the information for `Sample`, we don't need to have `tagSample` as well.

If we find we need some type, say `tagBar`, we can either
* blacklist tagA*, tagC*, tagD*, etc.
* expand the pattern language, or
* add the ability to white-list types

pshipton: The problem is that when we find we need some type, its already too late because the core file cannot necessarily be recreated.

runtime/ddr/blacklist
keithc-ca: Neither ClassTableEntry nor ClassTableEntry is currently used in DDR code.
* `ClassTableEntry` is defined in TgcInterRegionRememberedSetDemographics.cpp
* `classTableEntry` is defined in KeyHashTable.c

We could fix both, but I propose that be done in a separate PR.

pshipton: The structures don't need to be used in DDR code to be useful. When debugging, the structures are used to dump data found in the core file. Without the structures the data needs to be manually interpreted which would be much more painful and error prone. I'm ok to fix the problem in a separate PR, but suggest it be done before this one is merged.

runtime/ddr/blacklist
keithc-ca: I'll fix this along with classTableEntry in that separate PR.

runtime/ddr/blacklist
keithc-ca: Agreed.

runtime/ddr/blacklist
keithc-ca: I'll change the  struct tag from `userDataWrapper` to `UserDataWrapper` so we don't have to blacklist either type.

keithc-ca:
Addressed all review comments and rebased on current master branch.

keithc-ca:
Fixed caught exception type in PointerGenerator so it compiles with older versions of javac.

pshipton:
jenkins compile zlinux jdk8

PR #1706, opened 2018-04-17
Use the Runtime.Version API to get Java version

pdbain-ibm:
Update the VmArgumentTests to create the test jar filename using the Version
API where possible.

### DIFF
--- a/test/TestConfig/resources/excludes/latest_exclude_SE100.txt
+++ b/test/TestConfig/resources/excludes/latest_exclude_SE100.txt
@@ -32,5 +32,3 @@ org.openj9.test.attachAPI.TestAttachAPI:test_agntld02																	238 generi
 org.openj9.test.attachAPI.TestAttachAPI:test_agntld03																	238 generic-all
 org.openj9.test.attachAPI.TestSunAttachClasses:testAgentLoading															238 generic-all
 org.openj9.test.vmArguments.VmArgumentTests:testCrNocr																	244 generic-all
-org.openj9.test.vmArguments.VmArgumentTests:testExecutableJarArguments													1659 generic-all
-org.openj9.test.vmArguments.VmArgumentTests:testJarArguments															1659 generic-all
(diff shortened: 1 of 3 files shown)


smlambert:
Jenkins test extended plinux jdk10

test/Java8andUp/src/org/openj9/test/vmArguments/VmArgumentTests.java
```diff
@@ -145,9 +145,10 @@
 						"-Dsun.java.command",
 						"-Dsun.java.launcher"
 		};	
-		JAVA_VERSION=System.getProperty("java.version");
-		isJava8 = JAVA_VERSION.startsWith("1.8.0");
-		vmargsJarFilename = isJava8? "vmargs_SE80.jar": "vmargs_SE90.jar";
+		int javaMajorVersion = VersionCheck.major();
+		isJava8 = (8 == javaMajorVersion);
+		String javaMajorVersionString = "SE"+Integer.toString(javaMajorVersion)+"0";
+		vmargsJarFilename = "vmargs_"+javaMajorVersionString+".jar";
```
smlambert: I have recently understood that in a related PR it was mentioned that you do not need to add a version string to the jar file at all, as we only run one version or another, never at the same time.  Therefore, this could be simplified to simply :  vmargsJarFilename = "vmargs.jar";  

Please update the PR to do that.

pdbain-ibm:
The Java version suffix was added by https://github.com/eclipse/openj9/pull/149.  I will try without the version to ensure it is no longer necessary.  

I will squash the commits once this settles.

> eliminate the need for version check, and dependency on TestUtilities as it is not needed.

The Java version is still required for other tests which are sensitive to some peculiarities of the Java 8 default arguments.

smlambert:
Jenkins test extended plinux jdk10

pdbain-ibm:
@smlambert or @llxia I believe the compilation problems are fixed.  Would you kindly launch another JDK10 extended build?

llxia:
Jenkins test extended plinux jdk10

pdbain-ibm:
Jenkins jdk10 job passed.  The CI build failed for unrelated reasons.

PR #1712, opened 2018-04-18
Rename invocation of JNU_InitializeEncoding()

andrew-m-leonard:
to match OpenJDK(11) merged exported name InitializeEncoding().

Signed-off-by: Andrew Leonard <andrew_m_leonard@uk.ibm.com>


andrew-m-leonard:
OpenJDK community said to be consistent with other exports (eg.canonicalize) we simply add JNIEXPORT to the current method, rather than add a new JNU_... method...

DanHeidinga:
Change looks good.  As the JDK11 extensions repo isn't in the open yet, will rely on the travis build to validate the compile is good before merging (even though the change looks safe).

pshipton:
jenkins compile plinux

pshipton:
I didn't mean to run all the jdk level, I've stopped jdk8, 9 and left 10 running

PR #1716, opened 2018-04-18
Refactor Valhalla nestmates tests

taliamccormick:
- Revert from asm6 to asm7
- Refactor class generator to reduce code duplication

Precursor to https://github.com/eclipse/openj9/pull/1703 - external PR build does not allow use of asm7.

### DIFF
--- a/test/functional/Valhalla/src/org/openj9/test/valhalla/ClassGenerator.java
+++ b/test/functional/Valhalla/src/org/openj9/test/valhalla/ClassGenerator.java
@@ -24,137 +24,72 @@ package org.openj9.test.valhalla;
 import org.objectweb.asm.*;
 
 public class ClassGenerator implements Opcodes {
+	
+	public static byte[] generateClass(String name, Attribute[] attributes) {
+		ClassWriter classWriter = new ClassWriter(0);
+		classWriter.visit(55, ACC_PUBLIC | ACC_SUPER, name, null, "java/lang/Object", null);
+		
+		/* Generate base constructor which just calls super.<init>() */
+		MethodVisitor methodVisitor = classWriter.visitMethod(ACC_PUBLIC, "<init>", "()V", null, null);
+		methodVisitor.visitCode();
+		methodVisitor.visitVarInsn(ALOAD, 0);
+		methodVisitor.visitMethodInsn(INVOKESPECIAL, "java/lang/Object", "<init>", "()V", false);
+		methodVisitor.visitInsn(RETURN);
+		methodVisitor.visitMaxs(1, 1);
+		methodVisitor.visitEnd();
+
+		/* Generate passed in attributes if there are any */
+		if (null != attributes) {
+			for (Attribute attr : attributes) {
+				classWriter.visitAttribute(attr);
+			}
+		}
+
+		classWriter.visitEnd();
+		return classWriter.toByteArray();
+	}
 
 	public static byte[] MultipleNestMembersAttributesdump () throws Exception {
-		ClassWriter cw = new ClassWriter(0);
-		MethodVisitor mv;
-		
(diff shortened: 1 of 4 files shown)


taliamccormick:
Requesting reviewers @tajila @smlambert @llxia

test/functional/Valhalla/src/org/openj9/test/valhalla/ClassGenerator.java
```diff
@@ -24,128 +24,60 @@
...
 
 public class ClassGenerator implements Opcodes {
-
-	public static byte[] MultipleNestMembersAttributesdump () throws Exception {
-		ClassWriter cw = new ClassWriter(0);
-		MethodVisitor mv;
+	
+	public static byte[] generateClass(String name, Attribute[] attributes) {
+		ClassWriter classWriter = new ClassWriter(0);
+		classWriter.visit(V1_8, ACC_PUBLIC | ACC_SUPER, name, null, "java/lang/Object", null);
```
tajila: Will generating a class with classfileMajorVersion 52 trigger verification failures if nestmates attributes are detected?

taliamccormick:
> Will generating a class with classfileMajorVersion 52 trigger verification failures if nestmates attributes are detected?

It does not & (as of current spec draft), should not.

--

Working with this spec draft (http://cr.openjdk.java.net/~dlsmith/nestmates.html), my understanding is:
- If nestmates attributes are read on an earlier VM, then they are ignored
- If an access check requires nestmates functionality (eg. one class accesses another class's private field, both are in the same nest) but the VM is an earlier version without nestmates features, then the access check should fail
- We do not explicitly check if unrecognized attributes are valid for that version - we simply skip over them

Since nestmates code is within build flags, if nestmates are not enabled - which should occur if the VM is expecting classfileMajorVersion 52 or less -- then the class file reader will skip over the nestmates attributes and not trigger a verification failure. Verification works in our particular case for classfileMajorVersion 52 because we enable it on a VM which accepts classfileMajorVersion 52. 

Implementation details are in `readAttributes(...)` (runtime/bcutil/cfreader.c:130) - default unrecognized option stores the attribute but does not check version.

--

Details on recognizing new attributes in my copy of the JVM spec, version 8, section `4.7.1 Defining and Naming New Attributes` gives
> 4.7.1 Defining and Naming New Attributes

> Compilers are permitted to define and emit  class files containing new attributes in   the   
attributes   tables   of   class   file   structures, field_info   structures, method_info  structures,  and  Code  attributes  (4.7.3).  Java  Virtual  Machine implementations are permitted to recognize and use new attributes found in these attributes  tables.  However,  any  attribute  not  defined  as  part  of  the  
class  file specification must not affect the semantics of the class file. Java Virtual Machine implementations are required to silently ignore attributes they do not recognize.

> For  instance,  defining  a  new  attribute  to  support  vendor-specific  debugging  is permitted. Because Java Virtual Machine implementations are required to ignore attributes they do not recognize, class files intended for that particular Java Virtual Machine  implementation  will  be  usable  by  other  implementations  even  if  those implementations cannot make use of the additional debugging information that the class  files contain.

> Java Virtual Machine implementations are specifically prohibited from throwing an exception or otherwise refusing to use class files simply because of the presence of some new attribute. Of course, tools operating on class files may not run correctly if given class files that do not contain all the attributes they require.

smlambert:
Looks like we are good to go here.  I can merge this, or do we want to wait until #1751 is in, and then test the PR build as well?

taliamccormick:
I would prefer to merge this now - https://github.com/eclipse/openj9/pull/1751 depends on it, but this doesn't depend on https://github.com/eclipse/openj9/pull/1751. I have also tested the PR build after rebasing on this and it seems fine.

I'll squash the review commit & rebase now.

PR #1717, opened 2018-04-19
Ignore class instantiation error at stage preparing class

### LINKED ISSUE #1627: Class.getDeclaredFields throws exception after invoking Class.newInstance
Accordingly to Javadoc of Class.getDeclaredFields, "...an array of Field objects reflecting all the fields declared by the class or interface represented by this Class object" must be returned. But using the following program declared fields of class B can not be retrieved:

```
public class A {
   static {
      String x = null;
      x.getClass();
   }
}

public class B extends A {
   public int x;
}

public class C {
   public static void main(String[] args) throws NoSuchMethodException, NoSuchFieldException {
      try {
         B.class.newInstance();
      } catch (ExceptionInInitializerError e) {
      } catch (Exception e) {
      } finally {
         for (Field f : B.class.getDeclaredFields()) {
            System.out.println(f);
         }
      }
   }
}
```

**Affected versions:**

openjdk version "1.8.0_162"
OpenJDK Runtime Environment (build 1.8.0_162-b12)
Eclipse OpenJ9 VM (build openj9-0.8.0, JRE 1.8.0 Linux amd64-64 Compressed References 20180315_120 (JIT enabled, AOT enabled)
OpenJ9   - e24e8aa9
OMR      - 3e8296b4
JCL      - ee1e77df1d based on jdk8u162-b12)

Steps to reproduce:

	1. Compile above program
	2. Run it using openj9 java

**Current result:**

Exception in thread "main" java.lang.NoClassDefFoundError: samples.newInstance.B (initialization failure)
	at java.lang.J9VMInternals.initializationAlreadyFailed(J9VMInternals.java:96)
	at java.lang.J9VMInternals.prepareClassImpl(Native Method)
	at java.lang.J9VMInternals.prepare(J9VMInternals.java:300)
	at java.lang.Class.getDeclaredFields(Class.java:830)
	at samples.newInstance.C.main(C.java:27)
Caused by: java.lang.NullPointerException
	at samples.newInstance.A.<clinit>(A.java:7)
	at java.lang.J9VMInternals.newInstanceImpl(Native Method)
	at java.lang.Class.newInstance(Class.java:1773)
	at samples.newInstance.C.main(C.java:11)

**Expected result:**

public int x

JasonFengJ9:
Ignore class instantiation error at stage preparing class

Allow `j.l.Class` instance even though the class representing by this `j.l.Class` instance can't be instantiated.
Manually verified that JVM with this PR matches RI behaviour as per issue #1627 

Close #1627 

Reviewer: @DanHeidinga 

### DIFF
--- a/runtime/vm/ClassInitialization.cpp
+++ b/runtime/vm/ClassInitialization.cpp
@@ -464,7 +464,14 @@ doVerify:
 				break;
 			}
 			case J9ClassInitFailed: {
-				sendInitializationAlreadyFailed(currentThread, clazz, 0, 0, 0);
+				/* J9ClassInitFailed can only be set when unlockedStatus is J9ClassInitNotInitialized, and desiredState is J9_CLASS_INIT_INITIALIZED.
+				 * J9ClassInitFailed can be ignored if desiredState is not J9_CLASS_INIT_INITIALIZED, i.e., J9_CLASS_INIT_VERIFIED or J9_CLASS_INIT_PREPARED.
+				 */
+				if (desiredState < J9_CLASS_INIT_INITIALIZED) {
+					Trc_VM_classInitStateMachine_desiredStateReached(currentThread);
+				} else {
+					sendInitializationAlreadyFailed(currentThread, clazz, 0, 0, 0);
+				}
 				goto done;
 			}
 			case J9ClassInitNotInitialized: {


DanHeidinga:
@JasonFengJ9 I would have expected a change (or several) to `java.lang.Class#getDeclaredMethods` and related methods.  Basically anywhere that calls `J9VMInternals.prepare(this);` should allow Reflective access to classes that can't be initialized.  

Changing the ClassInitialization state machine will have effects throughout the rest of the JVM that may be unintended.

JasonFengJ9:
Current `J9VMInternals.prepare(this)` throws the exception caught during all three stages: verifying, preparing and initializing. The correct (or RI) behvaviour is that it should only throw error/exception during stages of verifying and preparing, the exception during initializing should not cause `J9VMInternals.prepare(this)` re-throwing that exception. In Java code, there is no straightforward way to differentiate the exception from verifying/preparing or  initializing.

In native, `clazz->initializeStatus` goes through `J9ClassInitUnverified --> J9ClassInitUnprepared --> J9ClassInitNotInitialized --> J9ClassInitSucceeded` in a successful loading. 
When an exception occurred, 
```
J9ClassInitUnverified --> J9ClassInitUnverified (i.e., the state stays same)
J9ClassInitUnprepared --> J9ClassInitUnprepared (i.e., the state stays same)
J9ClassInitNotInitialized --> J9ClassInitFailed
J9ClassInitFailed --> J9ClassInitFailed (i.e., the state stays same)
```
in current code `clazz->initializeStatus` is changed from `J9ClassInitNotInitialized` to `J9ClassInitFailed` when an exception occurred which is the only state changing in case of error, other states stay same when an exception occurred.
Note: the recursive call to `classInitStateMachine` for `superClass` and `superInferfaces` follows this rule as well.

To summarize, state `J9ClassInitFailed` means an error during initializing but not verifying or preparing hence `J9VMInternals.prepare(this)` should not throw the exception when current state is `J9ClassInitFailed`.
This PR ignores state `J9ClassInitFailed` when target state is J9_CLASS_INIT_VERIFIED or J9_CLASS_INIT_PREPARED, i.e., performing verifying or preparing.

DanHeidinga:

PR #1719, opened 2018-04-19
isCacheFileName() should be able to recognize old cache name.

hangshao0:
1. getModLevelFromName() returns 0 if there is no
J9SH_MODLEVEL_PREFIX_CHAR in the cache file name.
2. Fix caller of getStringForShcModlevel() and getStringForShcAddrmode()
to pass the name to array.


runtime/util_core/j9shchelp.c
```diff
@@ -148,6 +150,9 @@ getModLevelFromName(const char* cacheNameWithVGen)
 		if (0 != scan_udata(&cursor, &modLevel)) {
 			modLevel = 0;
 		}
+	} else {
+		/* old cache name that does not have J9SH_MODLEVEL_PREFIX_CHAR */
+		modLevel = (uintptr_t)-1;
```
pshipton: getModLevelFromName() is also used in SH_OSCache::removeCacheVersionAndGen() where -1 isn't handled specially, although it older caches shouldn't be a factor for snapshots of cache stats. Maybe return 1 instead so the `< 10` checks will just work?

hangshao0: I will change it to return 0 if it's an old cache that does not have modLevel, -1 on error.

hangshao0: getModLevelFromName() is changed to return intptr_t.

pshipton:
jenkins test sanity plinux jdk8,jdk10

PR #1722, opened 2018-04-19
Do not fail the test pipeline build when clean workspace fails

vsebe:

### DIFF
--- a/buildenv/jenkins/common/test
+++ b/buildenv/jenkins/common/test
@@ -118,6 +118,14 @@ def test_all() {
 def test_all_with_fetch() {
     fetch_artifacts()
     test_all()
-    cleanWs()
+
+    try {
+        cleanWs()
+    } catch(e) {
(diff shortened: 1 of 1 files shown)


vsebe:
@jdekonin please review, thx!

buildenv/jenkins/common/test
```diff
@@ -118,6 +118,14 @@ def test_all() {
 def test_all_with_fetch() {
     fetch_artifacts()
     test_all()
-    cleanWs()
+
+    try {
+        cleanWs()
+    } catch(e) {
+        echo e.toString()
+        if (SPEC.startsWith == "win_") {
```
jdekonin: Can you verify that this syntax will work?  It shouldn't be ```if(SPEC.startsWith("win_"))```?

vsebe: good catch, fixed

PR #1725, opened 2018-04-19
Fix type clashes in the global namespace

keithc-ca:
DDR expects types in the global namespace to have distinct (case-insensitive) names, otherwise there are problems on Windows (due to its case-insensitive file-system).

This removes the clashes over `UserDataWrapper/userDataWrapper` and `ClassTableEntry/classTableEntry`. The names in KeyHashTable.c would clash with TgcInterRegionRememberedSetDemographics.cpp if `classTableEntry` were simply renamed `ClassTableEntry`.


pshipton:

PR #1730, opened 2018-04-20
Fix order of exceptions for method sends

tajila:
Fix order of exceptions for method sends

Resolution must occur before checking the receiver for NPE

Reviewer: @gacholio

### DIFF
--- a/runtime/vm/BytecodeInterpreter.hpp
+++ b/runtime/vm/BytecodeInterpreter.hpp
@@ -6358,8 +6358,15 @@ done:
 		UDATA volatile methodIndexAndArgCount = ramMethodRef->methodIndexAndArgCount;
 		UDATA methodIndex = methodIndexAndArgCount >> 8;
 		j9object_t receiver = ((j9object_t*)_sp)[methodIndexAndArgCount & 0xFF];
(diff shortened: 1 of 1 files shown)


tajila:
The spec states:
> The Java Virtual Machine instructions anewarray, checkcast, getfield, getstatic, instanceof, invokedynamic, invokeinterface, invokespecial, invokestatic, invokevirtual, ldc, ldc_w, multianewarray, new, putfield, and putstatic make symbolic references to the run-time constant pool. Execution of any of these instructions requires resolution of its symbolic reference.

> Resolution is the process of dynamically determining concrete values from symbolic references in the run-time constant pool. 

This means that we need to resolve the methodRed before throwing NPE on a NULL receiver.

We will need a similar change for invokeSpecial. InvokeInterface does the right thing

gacholio:
In fact, the best thing to do is probably to just re-run the current bytecode after the resolution - there's no need to have two copies of the code.

tajila:
@gacholio I removed `fromBytecode` for the NULL receiver check. I still kept it for the `profileInvokeReceiver` so we dont call it twice when resolving.

gacholio:

PR #1732, opened 2018-04-20
Get segment size before erase

dsouzai:
In J9::SystemSegmentProvider::release, when a large segment is freed, it
doesn't get added to the free list, but rather, is immedately released
to the OS. However, the code would first free this memory, and then read
the size, which can cause crashes or risks reading invalid data. This
commit fixes this issue. This is only an issue because when a
TR::MemorySegment is added to the _segments set in
J9::SystemSegmentProvider::allocateNewSegment, what gets returned is not
a copy of the TR::MemorySegment, but a reference to the element in the
set; therefore, when the element is erased from the set, all references
to it become invalid.


dsouzai:
@mpirvu could you review.

mpirvu:
Jenkins test all

PR #1733, opened 2018-04-20
Add timeout to pipeline builds

vsebe:
Set timeout for build pipeline builds to 6 hours and to 8 hours 
for test pipeline builds. 

Issue: https://github.com/eclipse/openj9/issues/974

[ci skip]

### DIFF
--- a/buildenv/jenkins/common/build
+++ b/buildenv/jenkins/common/build
@@ -229,21 +229,23 @@ def add_node_to_description() {
 def build_all() {
 
     // Typically called by Build jobs and Compile only PRs
-
-    add_node_to_description()
-    get_source()
-    build()
-    archive()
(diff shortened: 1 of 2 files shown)


vsebe:
@AdamBrousseau please review, thank you!

vsebe:
Agreed, the stage is not necessary. 
In build_all(), those 3 steps would likely reveal slave connectivity issues (the build would fail immediately) or issues with the server itself so the timeout is unnecessary.

vsebe:
@pshipton please review, thank you!

pshipton:
Merging so we have some timeouts. Why are the timeouts so large, no build should take that long. I'd think 3 and 4 would be sufficient.

Issue #1043, opened 2018-01-29
Ability to pass WHITELIST to copyright checker


AdamBrousseau:
Not all files in a repo will have the standard copyright. We should have the ability to pass a whitelist of filenames so that the check does not fail when these files are modified.

Since we are using the copyright checker across multiple repositories, I was thinking we could automatically look for a `.copyright_whitelist` file in the project root directory.

PR #1734, opened 2018-04-20
Make spelling of "initialize" consistent

fjeremic:
This will make searching easier as one does not have to check for
alternate spellings.

### DIFF
--- a/runtime/port/win32/j9sock.c
+++ b/runtime/port/win32/j9sock.c
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 1991, 2015 IBM Corp. and others
+ * Copyright (c) 1991, 2018 IBM Corp. and others
  *
  * This program and the accompanying materials are made available under
  * the terms of the Eclipse Public License 2.0 which accompanies this
--- a/debugtools/DDR_Artifacts/specs/29/aix_ppc
+++ b/debugtools/DDR_Artifacts/specs/29/aix_ppc
@@ -11786,7 +11786,7 @@ F|_reduceStoreContentionDisabled|_reduceStoreContentionDisabled|bool|bool
 F|_debugData|_debugData|class ClassDebugDataProvider*|class ClassDebugDataProvider*
 F|_commonCCInfo|_commonCCInfo|struct J9ShrCompositeCacheCommonInfo*|struct J9ShrCompositeCacheCommonInfo*
 F|_newHdrPtr|_newHdrPtr|class SH_CompositeCacheImpl$SH_SharedCacheHeaderInit*|class SH_CompositeCacheImpl::SH_SharedCacheHeaderInit*
-S|SH_CompositeCacheImpl$SH_SharedCacheHeaderInit|SH_CompositeCacheImpl$SH_SharedCacheHeaderInitPointer|SH_OSCache$SH_OSCacheInitialiser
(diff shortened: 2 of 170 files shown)


fjeremic:
This PR depends on https://github.com/eclipse/omr/pull/2490.

fjeremic:
Jenkins test sanity plinux depends eclipse/omr#2490

fjeremic:
The copyright failures all seem to be in the debugtools/DDR_Artifacts directories. These files do not even have copyright dates however, so this seems to be a tooling issue. @AdamBrousseau FYI. Perhaps the Copyright Check tool should validate that a copyright header exists in the first place before marking failures?

AdamBrousseau:
I suppose we could ignore **Modified** files if they don't have any copyright, but we probably still want to fail if an **Added** file doesn't have a copyright. There is an open issue (#1043) to have a whitelist ability added to the tool, which would also solve this problem. Until the tool is rock solid, its mostly just a flag to point out "potential" issues. It's up to the committer to determine if the failure is a problem. Although I've seen the wrong decision made in the past.

fjeremic:
@vijaysun-omr given that you're the committer for https://github.com/eclipse/omr/pull/2490 this one also needs a review. Both need to be merged together.

0xdaryl:
@fjeremic , would you mind rebasing please?

fjeremic:
> @fjeremic , would you mind rebasing please?

@0xdaryl rebased boht OpenJ9 and OMR chnages.

Jenkins test sanity

fjeremic:
> @fjeremic , would you mind rebasing please?

@0xdaryl rebased boht OpenJ9 and OMR chnages.

Jenkins test sanity depends eclipse/omr#2490

keithc-ca:
Is this worth the churn given the ability to use regular expressions (e.g. `initiali[sz]e`)?

fjeremic:
> Is this worth the churn given the ability to use regular expressions (e.g. initiali[sz]e)?

Perhaps a discussion to be had in https://github.com/eclipse/omr/issues/2486?

keithc-ca:
FYI, none of the files below DDR_Artifacts need be changed: they're not consumed by current builds.
It may make sense to simply remove them (in another PR).

fjeremic:
> FYI, none of the files below DDR_Artifacts need be changed: they're not consumed by current builds.
> It may make sense to simply remove them (in another PR).

I suppose if that's the case no point amending commits since we are removing it anyway. I'll leave it as is. When this is merged I'll open a PR to remove all the DDR_Artifacts.

0xdaryl:
Jenkins test sanity depends eclipse/omr#2490

0xdaryl:
Now that infra issues seem to be resolved, I re-ran the sanity tests just to be sure the 6 failures weren't hiding anything.

runtime/port/win32/j9sock.c
```diff
@@ -158,7 +158,7 @@ findError (int32_t errorCode)
 {
 	switch (errorCode)
 	{
-		case WSANOTINITIALISED:
+		case WSANOTINITIALIZED:
```
0xdaryl: `WSANOTINITIALISED ` is actually a Windows socket error code.

https://msdn.microsoft.com/en-us/library/windows/desktop/ms740668(v=vs.85).aspx

fjeremic:
Jenkins test sanity depends eclipse/omr#2490

runtime/port/win32/j9sock.c
fjeremic: Fixed in 82cbf35. Launched a new build now. Thanks for catching this!

fjeremic:
Windows failures are due to https://github.com/eclipse/openj9/issues/2129. @0xdaryl this one looks good to review again.

PR #1740, opened 2018-04-23
Accept the subtype of Throwable in tryFinally

ChengJin01:
The change is to address the issue with
the subtype of Throwable so as to accept
it as the 1st parameter of finallyHandle.


ChengJin01:
Reviewer: @DanHeidinga

DanHeidinga:

PR #1742, opened 2018-04-23
 Prevent invalid variable values in MethodType

gacholio:
If an exception occurs during deserialization, instance of MethodType
can be left in an invalid state which may cause crashes if they are
used.  Detect these exceptions and reset the MethodType to something
valid.


DanHeidinga:
Jenkins test sanity plinux

PR #1744, opened 2018-04-23
Expand DDR support for bitfields

keithc-ca:
* increase maximum bitfield width to 64 bits
* add support for 8-bit, 16-bit, and 64-bit base types
* fix generation of bitfields that follow non-bitfields
* stop using the deprecated visitMethodInsn() method
* stop blacklisting types with bitfields that are now supported


pshipton:
Can we enable or write any testing to validate this?

keithc-ca:
I tried some manual testing and realized that jdmpview doesn't have a _field formatter_ for bit fields and that the changes made in this PR don't handle signed bit fields properly.

DanHeidinga:
@keithc-ca Is this still relevant or can it be closed?

keithc-ca:
I haven't looked at this in a while, but I think we should still support bit fields where the base type is not 32 bits, e.g. short, long.

genie-openj9:

keithc-ca:
This hasn't been a priority, but I'd like to address this eventually.

PR #1745, opened 2018-04-23
Fix Dump ConfigurationUnavailableException issue and dsn name for z/OS

chandrams:
1) As the dump settings are changed one after the other, at times we get a Dump configuration cannot be changed as another dump change is in progress, introducing a delay fixes the issue. 
2) Fixed systemDumpToFile in Dump API  to replace file with dsn for z/OS

### DIFF
--- a/test/functional/JLM_Tests/src/org/openj9/test/java/lang/management/RemoteProcess.java
+++ b/test/functional/JLM_Tests/src/org/openj9/test/java/lang/management/RemoteProcess.java
@@ -21,11 +21,15 @@
  *******************************************************************************/
 
 package org.openj9.test.java.lang.management;
-import java.io.File;
+
+import org.testng.Assert;
+import org.testng.log4testng.Logger;
 import java.io.UnsupportedEncodingException;
+import org.openj9.test.management.ProcessLocking;
 
 public class RemoteProcess {
-
+	public static final Logger logger = Logger.getLogger(RemoteProcess.class);
+	private static ProcessLocking lock;
 	/**
 	 * Start the RemoteProcess with jmx enabled, for example:
 	 * -Dcom.sun.management.jmxremote.port=9999
@@ -33,28 +37,27 @@ public class RemoteProcess {
 	 * -Dcom.sun.management.jmxremote.ssl=false
 	 */
 	public static void main(String[] args) {
-		System.out.println("=========RemoteProcess Starts!=========");
-		for (int i = 0; i < 60; i++) {
-			try {
-				int[] a = new int[1024 * 1024];
-				new String("hello").getBytes("Unsupported");
-			} catch (UnsupportedEncodingException e) {
-				/* Expected exception from getBytes() hence ignored */
-			}
-
-			try {
-				String currentDir = System.getProperty("user.dir");
-				File file = new File(currentDir + File.separator + "catch.txt");
-				if (!file.exists()) {
-					if (!file.createNewFile()) {
-						System.out.println("Creating of catch.txt failed!");
-					}
-				}
-				Thread.sleep(1000);
-			} catch (Exception e) {
-				e.printStackTrace();
-			}
+		logger.info("=========RemoteProcess Starts!=========");
+		String lockFile = System.getProperty("java.lock.file");
+		Assert.assertNotNull(lockFile, "lockFile from -Djava.lock.file is null ");
+		lock = new ProcessLocking(lockFile);
+		lock.notifyEvent("child started");
+		lock.waitForEvent("dump settings done");
+		try {
+			int[] a = new int[1024 * 1024];
+			new String("hello").getBytes("Unsupported");
+			Assert.fail("Expected Exception - UnsupportedEncodingException did not occur!");
+		} catch (UnsupportedEncodingException e) {
+			/* Expected exception from getBytes() hence ignored */
 		}
-		System.out.println("RemoteProcess Stops!");
-	}
+
+		try {
+			Thread.sleep(1000);
+		} catch (InterruptedException e) {
+			Assert.fail("Exception occurred while sleepping" + e.getMessage());
+		}
+		lock.notifyEvent("events occurred");
+		lock.waitForEvent("closed JMX connection");
+		logger.info("========RemoteProcess Complete=========");
+        }
 }
--- a/test/functional/JLM_Tests/src/org/openj9/test/java/lang/management/TestOpenJ9DiagnosticsMXBean.java
+++ b/test/functional/JLM_Tests/src/org/openj9/test/java/lang/management/TestOpenJ9DiagnosticsMXBean.java
@@ -33,11 +33,11 @@ import java.io.BufferedReader;
 import java.io.InputStreamReader;
 import java.io.FileReader;
 import java.io.UnsupportedEncodingException;
+import java.io.FileNotFoundException;
 import java.lang.management.ManagementFactory;
 import java.net.MalformedURLException;
 import java.util.ArrayList;
 import java.util.List;
-
 import javax.management.JMX;
 import javax.management.MBeanServer;
 import javax.management.MBeanServerConnection;
@@ -49,6 +49,7 @@ import javax.management.remote.JMXServiceURL;
 
 import com.ibm.lang.management.*;
 import openj9.lang.management.*;
+import org.openj9.test.management.ProcessLocking;
 
 /**
  * Class for testing the APIs that are provided as part of OpenJ9DiagnosticsMXBean
@@ -59,13 +60,15 @@ import openj9.lang.management.*;
 @Test(groups = { "level.extended" })
 public class TestOpenJ9DiagnosticsMXBean {
 	private static Logger logger = Logger.getLogger(TestOpenJ9DiagnosticsMXBean.class);
+	private static String os = System.getProperty("os.name");
 	private static Process remoteServer;
-	private static Watchdog watchdog;
 	private ObjectName mxbeanName = null;
 	private OpenJ9DiagnosticsMXBean diagBean = null;
 	private OpenJ9DiagnosticsMXBean diagBeanRemote = null;
 	List<String> initialDumpOptions = new ArrayList<String>();
 	private JMXConnector connector = null;
+	private static ProcessLocking lock;
+	private static String tmpFileName;
 
 	@BeforeClass
 	public void setUp() throws Exception {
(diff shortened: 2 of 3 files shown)


pdbain-ibm:
In re 1.) above, I suggest using a mechanism other than a fixed timeout, such as checking for process termination or presence of a file.  Fixed timeouts at best slow down the test unnecessarily, and at worst fail intermittently.

keithc-ca:
The changes look OK to me, but I'd like Shelley to have a look too.

test/functional/JLM_Tests/src/org/openj9/test/java/lang/management/TestOpenJ9DiagnosticsMXBean.java
```diff
@@ -264,6 +265,8 @@ private void testLocal_setDumpOptions() {
 		}		
 
 		try {
+			/* Introduced a sleep to avoid Dump configuration unavailable issue */
+			Thread.sleep(2000);
```
DanHeidinga: As @pdbain-ibm indicated:

> In re 1.) above, I suggest using a mechanism other than a fixed timeout, such as checking for process termination or presence of a file. Fixed timeouts at best slow down the test unnecessarily, and at worst fail intermittently.

I've concerned about the fragility of adding Sleeps to tests to make them pass.  Is there something else we can trigger off?

DanHeidinga: A Sleep is never a "fix".

The PR description indicates:
> As the dump settings are changed one after the other, at times we get a Dump configuration cannot be changed as another dump change is in progress, introducing a delay fixes the issue.

The root question is what else is modifying the dump configuration at this point?

If the exception is expected (!!), then code using the api should be written to loop until the options have been successfully set
```
while(true) {
  try {
    diagBean.setDumpOptions("java:events=catch,filter=java/io/UnsupportedEncodingException,range=1..1,file=" + dumpFilePath);
    break;
  } catch(??? Exception e) {
    // Eat exception and try setting options again
  }
}
```

I hope that's not the expected API usage

keithc-ca: Unfortunately, I expect that's (close to) the expected usage pattern. I think a (small?) delay is appropriate if `setDumpOptions` throws `DumpConfigurationUnavailableException`.

DanHeidinga: The test code should write that kind of loop than rather than rely on that sleep being sufficient on it's own.

An even better solution may be to reorganize this test to check for success on the first dump before attempting the second.  The creation of the file from the first check could be used a trigger to indicate that its safe to modify the config.

chandrams: I tried both options instead of sleep, the first option worked tried with 50 runs but the second option failed.  

option 1
while(true) {
  try {
    diagBean.setDumpOptions("java:events=catch,filter=java/io/UnsupportedEncodingException,range=1..1,file=" + dumpFilePath);
    break;
  } catch(ConfigurationUnavailableException e) {
    // Eat exception and try setting options again
  }
}

option 2

To check for success on the first dump before attempting the second. The creation of the file from the first check could be used a trigger to indicate that its safe to modify the config.

smlambert: Definitely preferred over brittle timeout (have seen too many times the where timeout values get set, then nudged/increased, because tests are at mercy of machine config, etc).  

In the scenario (however rare) that the config is never available, I guess this while loop hangs the test, but we can/should put upper execution time limits on test suites at the framework level for that case.

test/functional/JLM_Tests/src/org/openj9/test/java/lang/management/TestOpenJ9DiagnosticsMXBean.java
```diff
@@ -406,8 +409,12 @@ private void triggerDump(OpenJ9DiagnosticsMXBean diagBean) {
 			try {
 				diagBean.triggerDump(agent);
 				if (!(agent.equals("stack"))) {
-					found = findAndDeleteFile(".", files[index]);
-					Assert.assertTrue(found, files[index] + " not found");
+					if ("z/OS".equalsIgnoreCase(os) && "system".equals(agent)) {
```
smlambert: You are calling this 3x in the code, may be useful as a helper method isZOSAgent() or something, especially if going to be used frequently.  (Also, will getSystemProperty("os.name") always return z/OS?  If that return value changes for whatever reason, you only change it in one spot.

test/functional/JLM_Tests/src/org/openj9/test/java/lang/management/TestOpenJ9DiagnosticsMXBean.java
pshipton: I don't expect there is a problem to put a short sleep inside the loop used for option 1, in order to give the JVM time to be ready without burning cpu. You could then either count the iterations and fail after some number, or put an upper bound on the execution time as Shelly suggests.

test/functional/JLM_Tests/src/org/openj9/test/java/lang/management/TestOpenJ9DiagnosticsMXBean.java
chandrams: os.name System property value I think might be using uname -I internally to fetch the OS name. Atleast, the machines I tested on returned z/OS.

pshipton: https://github.com/eclipse/omr/blob/884959f423d0e969ffec7340b695e4ed13daabf3/port/unix/omrsysinfo.c#L520
I expect it will always return "z/OS". Possibly there would be another similar OS in the future, although that seems unlikely as well. Nothing wrong with using common code anyway to avoid potential bugs from having one of the checks incorrect, or when modifications are made in the future.

pdbain-ibm:
You can steal code from Java8andUp/src/org/openj9/test/attachAPI/TestAttachAPIEnabling.java to test if you are running on z/OS.  That is, until we have a utility function for that, right @smlambert :-) ?

chandrams:
Thank you all for the review comments, I have made the changes as suggested. Please review again.

test/functional/JLM_Tests/src/org/openj9/test/java/lang/management/TestOpenJ9DiagnosticsMXBean.java
```diff
@@ -263,10 +264,30 @@ private void testLocal_setDumpOptions() {
...
+                                Assert.fail("Unexpected exception thrown: " + e.getMessage());
+			} catch (InterruptedException e) {
+				e.printStackTrace();
+				Assert.fail("Exception occurred while sleeping thread: " + e.getMessage());
+			}
+			if (counter == 20) {
+				break;
+			} 
+			counter++;
+		}
```
DanHeidinga: We don't want to do the sleep the first time if we can avoid it.  The sleep is intended to be a fallback only if required.

Something like this:
```java
while(true) {
	try { 
		diagBean.setDumpOptions("java:events=catch,filter=java/io/UnsupportedEncodingException,range=1..1,file=" + dumpFilePath);
		break;
	} catch (ConfigurationUnavailableException e) {
		// Ignore exception and try setting options again
		try {
			Thread.sleep(100);
		} catch (InterruptedException e) {
			// Do nothing - the outer loop will retry if required
		}
		// Only attempt this 20 times before failing the test
		if (counter == 20) {
			e.printStackTrace();
			Assert.fail("Unable to update the dump options: " + e.getMessage());
		} 
		counter++;
	} catch (InvalidOptionException e) {
		e.printStackTrace();
		Assert.fail("Unexpected exception thrown: " + e.getMessage());
	}
}
```

The `} catch (InvalidOptionException e) {` block is better handled by letting the test throw the InvalidOperationException and having TestNG handle it.  Letting the test system deal with unexpected exceptions means you get consistent output from failed tests rather than System.out.printlns

test/functional/JLM_Tests/src/org/openj9/test/java/lang/management/TestOpenJ9DiagnosticsMXBean.java
```diff
@@ -263,10 +264,30 @@ private void testLocal_setDumpOptions() {
 			e.printStackTrace();
```
DanHeidinga: The same here - let this code throw and TestNG can handle the exception

test/functional/JLM_Tests/src/org/openj9/test/java/lang/management/TestOpenJ9DiagnosticsMXBean.java
chandrams: @DanHeidinga Thanks, I have made the changes as suggested. I have also included passing of mode to the remote process and re-used ProcessLocking test class to synchronize between both the processes. Please review again.

test/functional/JLM_Tests/src/org/openj9/test/java/lang/management/TestOpenJ9DiagnosticsMXBean.java
chandrams: Thanks, I have modified the test to throw the exception.

test/functional/JLM_Tests/src/org/openj9/test/java/lang/management/RemoteProcess.java
```diff
@@ -21,40 +21,48 @@
...
-				Thread.sleep(1000);
-			} catch (Exception e) {
-				e.printStackTrace();
-			}
-		}
-		System.out.println("RemoteProcess Stops!");
-	}
+                try {
+                        int[] a = new int[1024 * 1024];
+                        new String("hello").getBytes("Unsupported");
```
DanHeidinga: Should be an Assert.fail() here as the exception is expected.

DanHeidinga:
Thanks @chandrams.  I'm happy with the handling of the sleep / delay now and the exception-related cleanup is good.

Can you also check the tabs vs spaces?  There appears to be a mix of the two in some of the files (should be tabs)

Will defer to @smlambert and co for the final review.

test/functional/JLM_Tests/src/org/openj9/test/java/lang/management/TestOpenJ9DiagnosticsMXBean.java
```diff
@@ -249,32 +242,43 @@ private void testLocal_triggerClassicHeapDump() {
...
+				}
+				counter++;
+			}
+		}
 
 		try {
-			dumpFileName = "javacore_unsupported.txt";
-			dumpFilePath = dir + File.separator + dumpFileName;
-			diagBean.setDumpOptions("java:events=catch,filter=java/io/UnsupportedEncodingException,range=1..1,file=" + dumpFilePath);
 			new String("hello").getBytes("Unsupported");
```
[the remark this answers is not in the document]
chandrams: Thanks Dan. I have fixed this.

chandrams:
Replaced the spaces with the tabs, hope I have got it right this time :)

pdbain-ibm:
Any progress on this?

chandrams:
@keithc-ca @smlambert Can you please review the changes. Thanks.

smlambert:
hi @chandrams - was this the test, testOpenJ9DiagnosticsMXBean_1, that you would like to try and grind, using the new 'grinder' job (to help reproduce intermittent failures?  

apologies, I meant to reach out to you yesterday to give it a spin

chandrams:
@smlambert 
Hi Shelley,
Yes, wanted to submit this test in a grinder, using my git hub branch that has the test fixes to check if there are no intermittent failures

chandrams:
@pshipton  Is that failure seen with the latest test case fix, I was wondering why it didn't print the below message after 20 retries (i.e. timeout of 2 seconds).
Assert.fail("Unable to update the dump options: " + e.getMessage());

pshipton:
The failure was seen in the nightly builds with the latest merged code. This PR is not yet merged.

chandrams:
@keithc-ca @smlambert  Can you please review this. Thanks.

test/functional/JLM_Tests/src/org/openj9/test/java/lang/management/RemoteProcess.java
```diff
@@ -21,40 +21,43 @@
...
-		for (int i = 0; i < 60; i++) {
-			try {
-				int[] a = new int[1024 * 1024];
-				new String("hello").getBytes("Unsupported");
-			} catch (UnsupportedEncodingException e) {
-				/* Expected exception from getBytes() hence ignored */
-			}
+		logger.info("=========RemoteProcess Starts!=========");
+		String lockFile = System.getProperty("java.lock.file");
+		Assert.assertTrue(null != lockFile, "lockFile from -Djava.lock.file is null ");
```
keithc-ca: This should use `assetNotNull`.

test/functional/JLM_Tests/src/org/openj9/test/java/lang/management/TestOpenJ9DiagnosticsMXBean.java
```diff
@@ -520,80 +473,75 @@ private static boolean findAndDeleteFile(String dirPath, String fileName) {
...
-		logger.info("Start Remote Server!");
+	private static void startRemoteServer() throws IOException {
+		logger.info("Starting Remote Server to Monitor!");
+		char fs = File.separatorChar;
+		/* Set up a randomly named file and pass this to child process. Parent keeps checking
+		* for its existence on the file-system. When this goes missing, it implies the child
+		* process is up and running.
+		*/
+		Random randomGenerator = new Random();
+		String tmpName = System.getProperty("java.io.tmpdir") + fs + randomGenerator.nextInt(10000) + ".lock";
```
keithc-ca: A better approach would be to use `File.createTempFile()` instead of hoping the random number won't relate to an existing lock file.

chandrams: Thanks Keith, made the changes.

keithc-ca:
I'd like to hear final thoughts from @smlambert before this is merged.

test/functional/JLM_Tests/src/org/openj9/test/java/lang/management/TestOpenJ9DiagnosticsMXBean.java
```diff
@@ -407,8 +373,12 @@ private void triggerDump(OpenJ9DiagnosticsMXBean diagBean) {
 			try {
 				diagBean.triggerDump(agent);
 				if (!(agent.equals("stack"))) {
-					found = findAndDeleteFile(".", files[index]);
-					Assert.assertTrue(found, files[index] + " not found");
+					if (isZOSSystemAgent(agent)) {
+						/* checking for dataset presence will be included later */
```
smlambert: Is there an open issue to track this later todo task?

test/functional/JLM_Tests/src/org/openj9/test/java/lang/management/TestOpenJ9DiagnosticsMXBean.java
```diff
@@ -520,80 +478,75 @@ private static boolean findAndDeleteFile(String dirPath, String fileName) {
...
+		/* The lock file that the parent and child processes use for event based synchronization. */
+		processArgs.add("-Djava.lock.file=" + tmpFileName);
+		processArgs.add(RemoteProcess.class.getName());
+		ProcessBuilder builder = new ProcessBuilder(processArgs);
+		logger.info(builder.command());
+		lock = new ProcessLocking(tmpFileName);
+		remoteServer = builder.start();
+
+		lock.waitForEvent("child started");
+		logger.info("Staring remote server finished");
```
smlambert: Staring or starting?

PR #1748, opened 2018-04-23
Fail if we cannot determine NODE

### LINKED ISSUE #1746: Jenkins: Pipeline should fail if we cannot find the right node labels to run on
During the [set_node](https://github.com/eclipse/openj9/blob/master/buildenv/jenkins/common/variables-functions#L196) function, if the value returned to `NODE` is blank, we should fail, otherwise the build will run on any node. This should never happen in production, but protecting against it can make debug easier when adding new stuff.

AdamBrousseau:
- Throw error if we cannot find a 'label' entry in
  vairables file that matches the job_type, SPEC,
  and SDK_VERSION passed.

[skip ci]

Fixes #1746


AdamBrousseau:
jenkins compile zlinux jdk8

AdamBrousseau:
jenkins compile zlinux jdk8

PR #1751, opened 2018-04-24
Add Valhalla Nestmates PR Build

taliamccormick:
This commit adds an automated Valhalla Nestmates PR build which compiles the VM with the J9VM_OPT_VALHALLA_NESTMATES build flag enabled and runs the Valhalla tests against it.

Test playlists will be updated in a subsequent commit to include the other sanity tests. Test config will be updated in a subsequent commit to update the Valhalla nestmates test options.

### DIFF
--- a/buildenv/jenkins/common/variables-functions
+++ b/buildenv/jenkins/common/variables-functions
@@ -245,7 +245,7 @@ def set_test_variables() {
     set_release()
     set_jdk_folder()
     
-    if ( SPEC ==~ /.*valhalla_nestmates/) {
+    if ( SPEC ==~ /.*valhalla_nestmates.*/) {
         JAVA_VERSION = "Valhalla"
     } else {
         JAVA_VERSION = "SE" + "${SDK_VERSION}" + "0"
--- a/buildenv/jenkins/common/variables-functions
+++ b/buildenv/jenkins/common/variables-functions
@@ -246,7 +246,7 @@ def set_test_variables() {
     set_release()
     set_jdk_folder()
     
-    if ( SPEC ==~ /.*valhalla_nestmates.*/) {
+    if ( SPEC ==~ /.*valhalla.*/) {
         JAVA_VERSION = "Valhalla"
     } else {
         JAVA_VERSION = "SE" + "${SDK_VERSION}" + "0"
--- a/buildenv/jenkins/common/variables-functions
+++ b/buildenv/jenkins/common/variables-functions
@@ -240,6 +240,7 @@ def set_build_variables() {
 
 /*
 * Sets variables for a test job on a given platform.
+* Note that JAVA_BIN is set separately within buildenv/jenkins/common/test
 */
 def set_test_variables() {
     set_release()
--- a/buildenv/jenkins/common/variables-functions
+++ b/buildenv/jenkins/common/variables-functions
@@ -244,7 +244,12 @@ def set_build_variables() {
 def set_test_variables() {
     set_release()
     set_jdk_folder()
-    JAVA_VERSION = "SE" + "${SDK_VERSION}" + "0"
+    
+    if ( SPEC ==~ /.*valhalla_nestmates/) {
+        JAVA_VERSION = "Valhalla"
+    } else {
+        JAVA_VERSION = "SE" + "${SDK_VERSION}" + "0"
+    }
     TEST_DEPENDENCIES_JOB_NAME = VARIABLES.test_dependencies_job_name
 }
 
--- a/buildenv/jenkins/jobs/pull-requests/PullRequest-Sanity-JDK9-linux_ppc-64_cmprssptrs_le_valhalla_nestmates
+++ b/buildenv/jenkins/jobs/pull-requests/PullRequest-Sanity-JDK9-linux_ppc-64_cmprssptrs_le_valhalla_nestmates
@@ -1,0 +1,51 @@
+/*******************************************************************************
+ * Copyright (c) 2018, 2018 IBM Corp. and others
+ *
+ * This program and the accompanying materials are made available under
+ * the terms of the Eclipse Public License 2.0 which accompanies this
+ * distribution and is available at https://www.eclipse.org/legal/epl-2.0/
+ * or the Apache License, Version 2.0 which accompanies this distribution and
+ * is available at https://www.apache.org/licenses/LICENSE-2.0.
+ *
+ * This Source Code may also be made available under the following
+ * Secondary Licenses when the conditions for such availability set
+ * forth in the Eclipse Public License, v. 2.0 are satisfied: GNU
+ * General Public License, version 2 with the GNU Classpath
+ * Exception [1] and GNU General Public License, version 2 with the
+ * OpenJDK Assembly Exception [2].
+ *
+ * [1] https://www.gnu.org/software/classpath/license.html
+ * [2] http://openjdk.java.net/legal/assembly-exception.html
+ *
+ * SPDX-License-Identifier: EPL-2.0 OR Apache-2.0 OR GPL-2.0 WITH Classpath-exception-2.0 OR LicenseRef-GPL-2.0 WITH Assembly-exception
+ *******************************************************************************/
+
+SDK_VERSION = '10'
+SPEC = 'linux_ppc-64_cmprssptrs_le_valhalla_nestmates'
+TEST_TARGET = '_sanity'
+
+node('master') {
+    echo "creating node1"
+    checkout scm
+    variableFile = load 'buildenv/jenkins/common/variables-functions'
+    variableFile.set_job_variables('pullRequest')
+    cleanWs()
+}
+
+node("${NODE}") {
+    echo "creating node23"
+    checkout scm
+    buildfile = load 'buildenv/jenkins/common/build'
+    testfile = load 'buildenv/jenkins/common/test'
+
+    // Build
+    buildfile.build_pr()
+
+    // Test
+    archive_test = [:]
+    archive_test['Archive'] = { buildfile.archive() }
+    archive_test['Test'] = { testfile.test_all() }
+    parallel archive_test
+    
+    buildfile.git_clean()
+}
--- a/buildspecs/linux_ppc-64_cmprssptrs_le_valhalla_nestmates.spec
+++ b/buildspecs/linux_ppc-64_cmprssptrs_le_valhalla_nestmates.spec
@@ -1,0 +1,283 @@
+<?xml version="1.0" encoding="UTF-8"?>
+
+<!--
+  Copyright (c) 2018, 2018 IBM Corp. and others
+ 
+  This program and the accompanying materials are made available under
+  the terms of the Eclipse Public License 2.0 which accompanies this
+  distribution and is available at https://www.eclipse.org/legal/epl-2.0/
+  or the Apache License, Version 2.0 which accompanies this distribution and
+  is available at https://www.apache.org/licenses/LICENSE-2.0.
+ 
+  This Source Code may also be made available under the following
+  Secondary Licenses when the conditions for such availability set
+  forth in the Eclipse Public License, v. 2.0 are satisfied: GNU
+  General Public License, version 2 with the GNU Classpath
+  Exception [1] and GNU General Public License, version 2 with the
+  OpenJDK Assembly Exception [2].
+ 
+  [1] https://www.gnu.org/software/classpath/license.html
+  [2] http://openjdk.java.net/legal/assembly-exception.html
+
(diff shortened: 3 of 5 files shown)


taliamccormick:
Requesting review from @tajila @AdamBrousseau @smlambert

taliamccormick:
This requires https://github.com/eclipse/openj9/pull/1716 to handle the asm7 dependency

Tested here: https://ci.eclipse.org/openj9/job/PullRequest-Sanity-JDKValhalla-linux_ppc-64_cmprssptrs_le_valhalla_nestmates-OpenJ9/21/

Note that the Valhalla tests currently skipped because they were set up on linux x86 64 and this build is on ppc 64 le. Will update test in subsequent commit
> Skipped due to jvm options ( -Xcompressedrefs ) and/or platform requirements (os.linux,arch.x86,bits.64) => NestAttributeTest_0_SKIPPED

buildenv/jenkins/common/variables-functions
```diff
@@ -244,7 +244,12 @@ def set_build_variables() {
 def set_test_variables() {
     set_release()
     set_jdk_folder()
-    JAVA_VERSION = "SE" + "${SDK_VERSION}" + "0"
+    
+    if ( SPEC ==~ /.*valhalla_nestmates/) {
+        JAVA_VERSION = "Valhalla"
+    } else {
+        JAVA_VERSION = "SE" + "${SDK_VERSION}" + "0"
+    }
```
smlambert: Do you need to set this variable here (change this file at all?), would you not just set this variable where you set valhalla_nestmates specs?

buildenv/jenkins/jobs/pull-requests/PullRequest-Sanity-JDK9-linux_ppc-64_cmprssptrs_le_valhalla_nestmates
```diff
@@ -0,0 +1,51 @@
...
+ *
+ * [1] https://www.gnu.org/software/classpath/license.html
+ * [2] http://openjdk.java.net/legal/assembly-exception.html
+ *
+ * SPDX-License-Identifier: EPL-2.0 OR Apache-2.0 OR GPL-2.0 WITH Classpath-exception-2.0 OR LicenseRef-GPL-2.0 WITH Assembly-exception
+ *******************************************************************************/
+
+SDK_VERSION = '10'
+SPEC = 'linux_ppc-64_cmprssptrs_le_valhalla_nestmates'
+TEST_TARGET = '_sanity'
```
smlambert: JAVA_VERSION = Valhalla  
if you set it here, you do not need to change variables-functions, correct?

buildenv/jenkins/jobs/pull-requests/PullRequest-Sanity-JDK9-linux_ppc-64_cmprssptrs_le_valhalla_nestmates
```diff
@@ -0,0 +1,51 @@
...
+    echo "creating node23"
+    checkout scm
+    buildfile = load 'buildenv/jenkins/common/build'
+    testfile = load 'buildenv/jenkins/common/test'
+
+    // Build
+    buildfile.build_pr()
+
+    // Test
+    archive_test = [:]
```
smlambert: what do these 3 calls to archive_test do?  frustrating to have to find where those functions are defined and what each call does.

buildspecs/linux_ppc-64_cmprssptrs_le_valhalla_nestmates.spec
```diff
@@ -0,0 +1,283 @@
...
+<spec xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xmlns="http://www.ibm.com/j9/builder/spec" xsi:schemaLocation="http://www.ibm.com/j9/builder/spec spec-v1.xsd" id="linux_ppc-64_cmprssptrs_le_valhalla_nestmates">
+	<name>Linux PPC LE 64bit Compressed Pointers Valhalla Nestmates</name>
+	<asmBuilderName>LinuxPPC64LE</asmBuilderName>
+	<cpuArchitecture>ppc</cpuArchitecture>
+	<os>linux</os>
+	<defaultJCL>Sidecar</defaultJCL>
+	<defaultSizes>desktop (256M + big OS stack)</defaultSizes>
+	<priority>150</priority>
+	<owners>
+		<owner>Talia.McCormick@ibm.com</owner>
```
smlambert: do we still use these owners tag?

buildspecs/linux_ppc-64_cmprssptrs_le_valhalla_nestmates.spec
```diff
@@ -0,0 +1,283 @@
...
+	<features>
+		<feature id="cmprssptrs"/>
+		<feature id="combogc"/>
+		<feature id="core"/>
+		<feature id="dbgext"/>
+		<feature id="gpu"/>
+		<feature id="se"/>
+		<feature id="se60_26"/>
+		<feature id="se7"/>
+		<feature id="se70_27"/>
```
smlambert: are all of these features still valid/used?  
why se 6.26, 7, 7.27 and not 8/9/10?

buildspecs/linux_ppc-64_cmprssptrs_le_valhalla_nestmates.spec
```diff
@@ -0,0 +1,283 @@
...
+	</features>
+	<source>
+		<project id="com.ibm.jvmti.tests"/>
```
smlambert: is this PR valid without jvmtitest.so / native tests compiled?

buildenv/jenkins/jobs/pull-requests/PullRequest-Sanity-JDK9-linux_ppc-64_cmprssptrs_le_valhalla_nestmates
```diff
@@ -0,0 +1,51 @@
...
+ * Exception [1] and GNU General Public License, version 2 with the
+ * OpenJDK Assembly Exception [2].
+ *
+ * [1] https://www.gnu.org/software/classpath/license.html
+ * [2] http://openjdk.java.net/legal/assembly-exception.html
+ *
+ * SPDX-License-Identifier: EPL-2.0 OR Apache-2.0 OR GPL-2.0 WITH Classpath-exception-2.0 OR LicenseRef-GPL-2.0 WITH Assembly-exception
+ *******************************************************************************/
+
+SDK_VERSION = '10'
```
llxia: is this for JDK9 or 10? The file name and SDK version are inconsistent.

buildenv/jenkins/jobs/pull-requests/PullRequest-Sanity-JDK9-linux_ppc-64_cmprssptrs_le_valhalla_nestmates
```diff
@@ -0,0 +1,51 @@
...
+node('master') {
+    echo "creating node1"
+    checkout scm
+    variableFile = load 'buildenv/jenkins/common/variables-functions'
+    variableFile.set_job_variables('pullRequest')
+    cleanWs()
+}
+
+node("${NODE}") {
+    echo "creating node23"
```
llxia: should this be removed?

taliamccormick: Woops, missed that - thank you

taliamccormick:
@smlambert 

> Do you need to set this variable here (change this file at all?), would you not just set this variable where you set valhalla_nestmates specs?

> JAVA_VERSION = Valhalla
if you set it here, you do not need to change variables-functions, correct?

I am trying to minimize script duplication and keep this pull request build similar to the others so that it's less likely to be broken during future refactoring. If I set it at the same time as `SDK_VERSION `, `SPEC`, and `TEST_TARGET `, then it will be reset by the call to `set_job_variables` - so I would have to either duplicate or change portions of `set_job_variables`. I will try setting it immediately after the call to `set_job_variables` and update here if that works.
Your approach should scale better, given that I've been told we are also adding a Valhalla LWorld PR build, Panama PR build, and a Valhalla nestmates non-PR build/test job.

taliamccormick:
@smlambert 

> do we still use these owners tag?

> are all of these features still valid/used?
why se 6.26, 7, 7.27 and not 8/9/10?

I think that the owners tag is still used. I imitated the exisiting ppcle spec for the format/content and changed what was necessary to enable the nestmates pr build. The last time a spec was added (https://github.com/eclipse/openj9/commit/5f6f3814664c4a6f44c8e2138560cf24ba1b35ec#diff-ecf00c5c9b02bcdf4d8c6f300c21ec43), it included the owners tag. Same reason for features 6.26, 7, 7.27 - was not sure what they are for so I kept them to avoid breaking anything.

> is this PR valid without jvmtitest.so / native tests compiled?

It should be. As of right now, the jvmti tests (and non Valhalla specific sanity tests) are not being run against the PR build. In a second separate PR I will update the playlist.xml files so that the jvmti tests (and non Valhalla specific sanity tests) are run against the PR build.

taliamccormick:
@llxia 

> is this for JDK9 or 10? The file name and SDK version are inconsistent.

Nestmates is for JDK 11. Since the openj9-openjdk-jdk11 repo is not yet out, I'm temporarily using JDK 10 which requires the SDK version to be 10. File name was accidental, will correct

buildenv/jenkins/common/variables-functions
AdamBrousseau: I think this is the correct spot as this is the only spot we set JAVA_VERSION (which is only used for test purposes). If we set it in the top level PR file then we would have to set it in every top level file we end up adding.

AdamBrousseau:
The correct/only way to add a spec is to copy the most similar spec and change only what you need. Unfortunately trying to trim down a spec is a rabbit hole. The hope is that the specs will be obsoleted by cmake.

buildenv/jenkins/common/variables-functions
smlambert: Right, so stick with 1.  

I'd like to see a pass of simplification on the pipeline scripts separate from this PR.

The set_test_variables() name is confusing as the method only sets JAVA_VERSION (RELEASE and SDK_DIR are not directly used by testing, must be mapped to the variables that tests expect... somewhere else).

I had an expectation to set the few environment variables needed for test in one spot, rather than scattered around the different pipeline functions.   In that way, the pipeline code maps more closely to the standalone run on command line story, making it more readable/understandable/maintainable by developers who have run testing locally. We should have the goal of less levels of indirection.

I would suggest creating a naming standard for specs that gets followed, so this change doesn't become a giant if else, else else statement as it gets scaled up.  You could change the regex to match on any spec with valhalla in it to set version Valhalla, and not have to keep growing the if statement.

taliamccormick: I'll change the regex to check that it matches any spec whose name contains 'valhalla' instead of any check which ends in `valhalla_nestmates'

It looks like the other test variables are set within `buildenv/jenkins/common/test` and that the value of `JAVA_BIN` has to be set there. This is because `set_job_variables()` is called on the master machine and the `JAVA_BIN` value refers to a location on the node machine. The `JAVA_BIN` location is on the node machine because the tests run on the node. Since the `JAVA_BIN` is set as
> JAVA_BIN = "${WORKSPACE}/${BUILD_RELEASE_IMAGES}${JDK_FOLDER}${JRE_FOLDER}/bin"

and the value of `WORKSPACE` is different between the two machines, it can't be set while running `set_job_variables` on the master one.

I could try copying/moving the test variable definitions from the `variables-functions` script so that they run on the node machine and can be set together. However, this will cause some duplication since some variables also need to be set on the master machine

taliamccormick: So within this PR, I will add another commit with a comment in `set_test_variables()` which indicates/clarifies that `JAVA_BIN` is set elsewhere and then open an issue for further discussion?

smlambert:
This looks ready to be merged, which I can do.

There is still one piece that should be changed/completed, which would be the trigger used to launch the PR testing.  Right now it is "Jenkins test experimental plinux jdkvalhalla"  but should drop the experimental key word (as there is no level called experimental) and be able to run either sanity or extended levels.  I believe this change is not done in a repo, but in some configuration view of Jenkins.

smlambert:
thanks @taliamccormick and @AdamBrousseau for your work on this 
and updating the trigger to "Jenkins test sanity plinux jdk11-nestmates", etc.

smlambert:
Will merge this now, understanding that we will update this to jdk11 once that repo is available (in a week or two).

PR #1752, opened 2018-04-24
Limit the invokestatic/special CP consistency check to JDK 9

DanHeidinga:
The checks, added in PR #1167, are required for JDK 9.  We've had
reports of applications failing due to these ICCEs when running on
JDK 8.

This is a stop-gap solution to allow the JDK8 applications to continue
to run successfully while we get to the root cause on JDK8 as the
CP types look correct at first glance.

### DIFF
--- a/runtime/vm/resolvesupport.cpp
+++ b/runtime/vm/resolvesupport.cpp
@@ -456,23 +456,28 @@ tryAgain:
 		cpClass = J9_CLASS_FROM_CP(ramCP);
 		lookupOptions |= J9_LOOK_CLCONSTRAINTS;
 
-		if ((cpClass != NULL) && (cpClass->romClass != NULL)) {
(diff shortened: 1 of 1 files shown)


DanHeidinga:
FYI - I've also removed the check from the MethodHandle case in `resolveVirtualMethodRefInto` as the check is redundant based on the later check for exactly CP_TYPE_METHODHANDLE

pshipton:
jenkins test all aix jdk8,jdk9

DanHeidinga:
Looking at PR #1167, I may need to make test changes as well.  Looking at the tests now.

DanHeidinga:
Or maybe not as the extended and sanity builds were clean for both JDK8 & 9

DanHeidinga:
Turns out the tests for this are disabled and have been since before OpenJ9's initial contribution.

PR #1760, opened 2018-04-25
Fixed SystemToDumpFile method in com.ibm.jvm.Dump to use dsn instead of file for z/OS

chandrams:
SystemToDumpFile method in com.ibm.jvm.Dump uses file sub-option on z/OS which is unrecognized, replace this with dsn

### DIFF
--- a/jcl/src/openj9.jvm/share/classes/com/ibm/jvm/Dump.java
+++ b/jcl/src/openj9.jvm/share/classes/com/ibm/jvm/Dump.java
@@ -99,6 +99,7 @@ package com.ibm.jvm;
  * configuration.
  */
 public class Dump {
+	private static final String SystemRequestPrefix = "z/OS".equalsIgnoreCase(System.getProperty("os.name")) ? "system:dsn=" : "system:file=";
   
 	/**
 	 * Trigger a java dump. A java dump is in a human-readable format, and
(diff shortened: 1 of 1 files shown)


chandrams:
@keithc-ca Please review. Thanks.

keithc-ca:
There are four places where `file=` should be replaced by `dsn=`; please compute the appropriate value and capture it in a static field that can be used in those four places.

jcl/src/openj9.jvm/share/classes/com/ibm/jvm/Dump.java
```diff
@@ -366,13 +366,17 @@ public static String heapDumpToFile() {
 	 */
     public static String systemDumpToFile(String fileNamePattern) throws InvalidDumpOptionException {
     	String request = null;
+	String os = System.getProperty("os.name");
```
keithc-ca: Be sure to use indentation that matches the surrounding code.

chandrams:
Which four places? I thought we need it only for systemDumpToFile() as only system dumps are written to data sets.

keithc-ca:
Perhaps you're right; looking again at `runtime/rasdump/dmpagent.c`, other places where 'dsn=' is expected don't seem to be accessible via Dump.java.

Please do:
* fix the formatting
* only compute the option string once
* only query OS if necessary
* invert the equals and switch to equalsIgnoreCase

chandrams:
Thanks for your suggestions @keithc-ca.  Can you please take a look again.

keithc-ca:
jenkins test sanity plinux

PR #1770, opened 2018-04-25
Add copyValue method to ObjectAccessBarrier

eric98zhang:
Add new method to MM_ObjectAccessBarrier to copy fields of a value class
instance into another value class instance.

This method differs from the cloneObject method in that the target
instances can be flattened within other objects, arrays,  or values, and is
intended to be used with bytecodes such as putfield, getfield, etc.

Since values may be flattened, it may not have an object header. For
this reason, the method must be given the value's J9Class explicitly.
Flattening also means values may not be 8-byte aligned. In
nocompressedrefs, this means copyValue cannot simply iterate through
fields 8-bytes at a time with the description bits the way cloneObject
does, as mixedObjectReadObject and mixedObjectStoreObject would be
performing unaligned reads and writes.

### DIFF
--- a/runtime/gc_base/ObjectAccessBarrier.cpp
+++ b/runtime/gc_base/ObjectAccessBarrier.cpp
@@ -1389,7 +1389,7 @@ MM_ObjectAccessBarrier::cloneObject(J9VMThread *vmThread, J9Object *srcObject, J
 		UDATA hashcodeOffset = _extensions->mixedObjectModel.getHashcodeOffset(destObject);
 		if (hashcodeOffset <= limit) {
 			I_32 *hashcodePointer = (I_32*)((U_8*)destObject + hashcodeOffset);
-			*hashcodePointer = hashCode;		
+			*hashcodePointer = hashCode;
 		}
 	}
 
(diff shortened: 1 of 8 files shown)


eric98zhang:
@tajila

runtime/gc_base/ObjectAccessBarrier.cpp
```diff
@@ -1441,6 +1441,75 @@ MM_ObjectAccessBarrier::cloneIndexableObject(J9VMThread *vmThread, J9IndexableOb
...
+ *
+ * @oaram valueClass The value class.
+ * @param srcObject The object containing the value class instance fields being copied.
+ * @param srcOffset The offset of the value class instance fields in srcObject.
+ * @param destValue The object containing the value class instance fields being copied to.
+ * @param destOffset The offset of the value class instance fields in destObject.
+ */
+void
+MM_ObjectAccessBarrier::copyValue(
+							J9VMThread *vmThread, J9Class *valueClass,
```
charliegracie: Can you put all of the parameters on one line please?  This will better match the rest of the code in this file.

runtime/gc_base/ObjectAccessBarrier.cpp
```diff
@@ -1441,6 +1441,75 @@ MM_ObjectAccessBarrier::cloneIndexableObject(J9VMThread *vmThread, J9IndexableOb
...
+							J9Object *srcObject, UDATA srcOffset,
+							J9Object *destObject, UDATA destOffset
+) {
+	/* subtract size of lockword */
+	const UDATA fieldsSize = _extensions->mixedObjectModel.getSizeInBytesWithoutHeader(valueClass) - sizeof(j9objectmonitor_t);
```
charliegracie: Is it a guarantee that there is a lockword?

runtime/gc_base/ObjectAccessBarrier.cpp
```diff
@@ -1441,6 +1441,75 @@ MM_ObjectAccessBarrier::cloneIndexableObject(J9VMThread *vmThread, J9IndexableOb
...
+	if (srcOffset & (sizeof(fj9object_t) - 1) || destOffset & (sizeof(fj9object_t) - 1)) {
+		/* Check whether the source or destination is not sizeof(ftj9object_t)-aligned.
```
charliegracie: This will skip all of the object access read/write barriers.  I am not sure if that is correct in this context.  You may need to just have the code in the else statement to be correct.  Is your statement about no object references true for compressed references?

runtime/gc_base/accessBarrier.cpp
```diff
@@ -618,6 +618,18 @@ j9gc_objaccess_cloneIndexableObject(J9VMThread *vmThread, J9IndexableObject *src
 	return barrier->cloneIndexableObject(vmThread, srcObject, destObject);
 }
 
+#if defined(J9VM_OPT_VALHALLA_VALUE_TYPES)
+void
+j9gc_objaccess_copyValue(
+		J9VMThread *vmThread, J9Class *valueClass,
```
charliegracie: Similar to above can you move these to one line.

tajila:

PR #1771, opened 2018-04-25
Add flattenable value type field support to getfield

eric98zhang:
Add support for flattenable value fields to getfield. When used with a
flattenable field, getfield creates a new instance of the value on the
heap and copies the fields of the flattened value to the new instance.

This operation requires the J9Class of the field, which is not available
from the J9RAMFieldRef. This version uses internalFindClassUTF8 to
locate the J9Class on every use of getfield with a flattened field. A
subsequent version could, in the parent J9Class, keep an array of
J9Classes of flattenable fields, and then use 16 bits of the
J9RAMFieldRef's valueOffset as an index to the correct J9Class.

### DIFF
--- a/runtime/vm/BytecodeInterpreter.hpp
+++ b/runtime/vm/BytecodeInterpreter.hpp
@@ -6216,6 +6216,33 @@ retry:
 
 				{
 					UDATA const newValueOffset = valueOffset + J9_OBJECT_HEADER_SIZE;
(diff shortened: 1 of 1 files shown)


eric98zhang:
Note: depends on #1770

eric98zhang:
@tajila

tajila:

PR #1776, opened 2018-04-26
Remove unnecessary heap tail padding

dmitripivkine:
-Special tail padding for object heap for s390 platforms is not
necessary any more.
- Tail padding is not used any more - remove related code 


dmitripivkine:
jenkins test sanity

dmitripivkine:
I believe that failures are not related to change itself (test infrastructure and network problems)

charliegracie:
@fjeremic can you quickly review the JIT side.  It is just removing an unused function but I would still like to have a JIT developer OK the change :)

This code stopped being used a release or 2 ago but I before the components were all in the same repo it was difficult to do the final cleanup.  This is just completing that cleanup.

pshipton:
@SueChaplain this change removes the requirement to use another page of memory on z, which is mentioned in the User Guide for zLinux and z/OS. References to using another page or subtracting 16 bytes can be removed. 

One place where its mentioned is https://www.ibm.com/support/knowledgecenter/SSYKE2_8.0.0/com.ibm.java.vm.80.doc/docs/j9_configure_large_page.html

For IBM the changes occurs in SR5 FP16.

SueChaplain:
@pshipton I thought that requirement was removed with SR5. That is what the topic you mention suggests (Releases before service refresh 5 only)

pshipton:
@dmitripivkine is this just cleanup of code that was already disabled?

dmitripivkine:
Yes, this is removing of dead code

Issue #1669, opened 2018-04-12
Create a Large Heap (noncompressedrefs) build


DanHeidinga:
OpenJ9 supports both compressedrefs and non-compressedrefs packages.  Currently, the project only builds and tests the compressedrefs package.

This is a problem for anyone wanting a heap larger than ~57 gig and limits adoption of OpenJ9 for projects that need large heaps.

The long term solution is outlined in https://github.com/eclipse/openj9/issues/643 but we need a more immediate way to meet this need.

Proposal
---
* Create a "Large Heap" OpenJ9 package for the non-compressedrefs mode to support users that need > ~57 gig heaps. 
* Enable building this package on all platforms but make it optional.
* Due to machine limitations, enable this package for compile and test:
  * OpenJ9: pLinux or Windows (depending on machine availability)
  * Adopt: Enable on xLinux
* Support this for JDK8 and newer releases (JDK10+)

Next steps:
 * [x] Remove any hardcoded references to compressedrefs from OpenJ9 and Extensions makefiles
 * [x] Update redirector to generate reasonable messages pointing to the other package when using `-Xcompressedrefs`/`-Xnoncompressedrefs`
* [x] Update build pipelines here to build the non-compressed package #1884
* [x] Add it to the OMR promotion criteria #1886
* [x] Determine what level of testing here for the non-compressed package
* [x] Work with Adopt to get this package built / tested there as well

Related: https://github.com/eclipse/openj9/issues/479 https://github.com/ibmruntimes/openj9-openjdk-jdk9/issues/20

I'm sure I've missed details here so committers please feel free to edit this to add details I've missed.

DanHeidinga:
> this build just "works that way" and so the actual heap size isn't important for testing?

@mstoodle The testing should focus on validating the behaviour of the runtime in this mode not on the size of the heap actually allocated.  The heap size limitation of compressed is due to the shifting of pointers.  No shifting in non-compressedrefs = no limit besides size of pointers / hw capabilities.

We'll have a committer do a one-off test of the build on a machine with sufficient memory to validate the ability to allocate large heaps and then rely on the way the underlying tech works.

mstoodle:
Classic white box testing reasoning, got it :) .

I am not disagreeing, because the code certainly does work that way (at least at this point in time :) ) so doing much more than what you've, uh, dictated, probably doesn't provide additional testing value from a technical standpoint.

[...]

PR #1777, opened 2018-04-26
Update redirector messages for X(no)compressedrefs

theresa-m:
[skip ci]

- step 2 to cerate a noncompressedrefs build
- if build type does not exist direct user to compile with correct config option

Partial fix: #1669

### DIFF
--- a/runtime/redirector/redirector.c
+++ b/runtime/redirector/redirector.c
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2001, 2017 IBM Corp. and others
+ * Copyright (c) 2001, 2018 IBM Corp. and others
  *
  * This program and the accompanying materials are made available under
  * the terms of the Eclipse Public License 2.0 which accompanies this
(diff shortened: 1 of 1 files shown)


theresa-m:
Resulting comment when build type does not exist:

> Selected VM [default] by option -Xnocompressedrefs does not exist.
> Compile with configuration option --enable-noncompressedrefs to generate a default build.

and

> Selected VM [compressedrefs] by option -Xcompressedrefs does not exist.
> Compile without configuration option --with-noncompressedrefs to generate a compressedrefs build.

pshipton:
I think we should call it a noncompressedrefs build instead of a default build.

theresa-m:
updated to 
> Compile with configuration option --with-noncompressedrefs to generate a noncompressedrefs build.

pshipton:
Actually non-compressedrefs with the dash would read better.

theresa-m:
yes. added the dash.

pshipton:
jenkins test all plinux jdk8

pdbain-ibm:
> I think we should call it a noncompressedrefs build instead of a default build.

since compressedrefs is actually the default :-), right?

Is it "-nocompressedrefs" (cf. https://github.com/eclipse/openj9/pull/1777#issuecomment-384674994) or "noncompressedrefs" (modulo the hyphen)?

pshipton:
There is a VM option -Xnocompressedrefs, but this PR is printing a message which now includes "non-compressedrefs", which is what I was referring to.

>Selected VM [default] by option -Xnocompressedrefs does not exist.
Compile with configuration option --enable-noncompressedrefs to generate a non-compressedrefs build.

PR #1783, opened 2018-04-26
Modify shutdown sequence to fix CS issue

ebadyano:
Fixing two issues with Concurrent Scavenger:
1. Modified shutdown sequence to do Scavenger tear down later in the
sequence; after all of the mutator threads are done. 
2. Added a new case in flushGCCaches to clear a GC Slave thread copy
cache when starting up with many GC threads and GC slave thread is
brought up during active scavenger cycle but doesn't do any actual work
and so is actually acting like a mutator thread until the next cycle.

Signed-off-by: Evgenia Badyanova <evgeniab@ca.ibm.com>

### DIFF
--- a/runtime/gc/gctable.c
+++ b/runtime/gc/gctable.c
@@ -54,7 +54,6 @@ J9MemoryManagerFunctions MemoryManagerFunctions = {
 	gcShutdownHeapManagement,
 	initializeMutatorModelJava,
 	cleanupMutatorModelJava,
-	internalFreeMemorySpace,
 #if defined(J9VM_GC_FINALIZATION)
 	j9gc_finalizer_startup,
 	j9gc_finalizer_shutdown,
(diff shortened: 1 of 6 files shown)


ebadyano:
personal build: http://vmfarm.ottawa.ibm.com/build_info.php?build_id=385447
Grinder with CS with the options that were triggering an assert before the changes:http://vmfarm.ottawa.ibm.com/build_info.php?build_id=385445

charliegracie:
@amicic if we are going to remove the `_DELETE` hook we should also remove the `_NEW` one since I do not see any users of it either.  I am a little worried that they are not used and were accidentally lost in OMR refactoring.

charliegracie:
Has this been run with -Xcheck:memory to make sure we are not leaking any memory on shutdown or causing any memory corruption?  Looking back through history this was added to cleanup shutdown issues.

ebadyano:
@charliegracie I guess J9HOOK_MM_PRIVATE_HEAP_DELETE has private in its name, so nobody should be using it externally either?
the personal build runs a few sanity tests with -Xcheck:memory and it caught a few issues for me with my initial changes.

amicic:
There is also HEAP_RESIZE, which has Verbose GC registered to it (indeed, we have a VGC stanza for it). I checked two previous major IBM Java Releases and things are same: RESIZE is used and NEW/DELETE are not. I guess, there were mostly there for symmetry. I'm ok with deleting the two unused ones - but I'll @charliegracie give the final vote.

charliegracie:
Merging this change as cleaning up the hooks should be completed in a separate change.

PR #1786, opened 2018-04-27
Remove count field in String and adopt OpenJDK StringBuffer / StringBuilder

fjeremic:
- Remove count field in String and use coder for compression

    This change removes the count field from the String class in Java 9
    onward. The String class in Java 8 is left unmodified. The count
    field represents the length of the string in the underlying array.
    Instead of using a separate field to represent the length we size the
    underlying array to be the same length as the string.

    By doing this we no longer have the ability to share the underlying
    array with StringBuffer / StringBuilder objects (unless they are of
    the correct size). Performance measurements have indicated this is a
    worth while trade off.

    Moreover, the elimination of the count field with it eliminates our
    ability to use the sign bit of the count field to represent String
    compression. As we still need to do this moving forward, and given
    that Java 10 introduces a coder field we simply adopt to start using
    this coder field to represent String compression instead.

    To get to a performant state however our StringBuffer / StringBuilder
    implementations also need to change. By introducing the coder field
    we are more easily able to adapt to using OpenJDK's version of these
    classes. This change also adopts the use of OpenJDK's StringBuffer /
    StringBuilder for performance in Java 10 onward.

    There were several steps made to make this change work. The original
    commits were squashed ahead of time as development went on to ease the
    pain of rebasing upstream changes to the files modified. The changes
    made are outlined as follows:

    _Separate String, StringBuffer, and StringBuilder preprocessor paths_

    Separate class definitions into contiguous chunks rather than
    interleaved within the code. This makes it much simpler to remove the
    count field for Java 9 and keep the Java 8 implementation intact.

    _Temporarily disable JIT toUpper/toLower acceleration_

    These algorithms need to be adjusted to extract the length from the
    right place. In fact I think these APIs need to be redesigned so these
    implementation details are processed in the Java code. The helpers
    should already receive the array and length as arguments.

- Adopt OpenJDK StringBuffer / StringBuilder

    _Fold away Sidecar19-SE path in StringBuffer / StringBuilder_

    This pre-processed path will no longer be used as we adopt OpenJDK's
    version of these classes instead.

    _Implement missing String APIs_

    After adopting OpenJDK's StringBuffer / StringBuilder we need to add
    some APIs to String and modify some existing APIs which access package
    protected methods in these classes. We make use of the existing helpers
    in StringLatin1 and StringUTF16 to avoid code duplication.

    _Remove calls to missing StringBuffer / StringBuilder APIs_

    Some APIs have now been deprecated. Avoid use of deprecated APIs in
    Class and System classes.

-  Fix toLowerCase combining above in Lithuanian

    The toLowerCase API is not converting Lithuanian characters correctly.
    The Unicode standard has updated the combining above code points to a
    "Canonical Combining Above" class [1] where as the table used to be
    just "Above" [3]. There were new code points introduced in [1] which
    need to be properly updated in our implementations.

    In addition the tests for lower casing Lithuanian are refactored and
    partitioned into Java 8 and Java 9 tests since each Java version
    supports a different level of the Unicode standard. The Java 8 test
    has been updated in accordance to the updates to toLowerCase and
    the Java 9 tests have been updated in accordance to [2].

    [1] https://www.unicode.org/Public/6.2.0/ucd/extracted/DerivedCombiningClass.txt
    [2] https://www.unicode.org/Public/8.0.0/ucd/extracted/DerivedCombiningClass.txt
    [3] https://www.unicode.org/Public/4.0-Update/extracted/DerivedCombiningClass-4.0.0.txt


cathyzhyi:
The JIT portion to disable String case conversion looks good to me.

jcl/src/java.base/share/classes/java/lang/String.java
```diff
@@ -3816,14 +6873,20 @@ private static int binarySearchRange(char[] data, char c) {
 		return mid - (c < value ? 1 : 0);
 	}
 
-	private static char[] startCombiningAbove = { '\u0300', '\u033D', '\u0346', '\u034A', '\u0350', '\u0357', '\u0363', '\u0483', '\u0592', '\u0597',
```
pshipton: Please add a comment that this data is encoded from the Canonical_Combining_Class=Above table in https://www.unicode.org/Public/6.2.0/ucd/extracted/DerivedCombiningClass.txt

jcl/src/java.base/share/classes/java/lang/String.java
```diff
@@ -857,47 +667,3830 @@ public String(char[] data) {
...
+				return codePoint - ('a' - 'A');
+			} else {
+				return codePoint;
+			}
+		} else {
+			return Character.toUpperCase(codePoint);
+		}
+	}
+
+// Some of the data below originated from the Unicode Character Database file
```
pshipton: This comment and the Unicode license below is not needed here since Java 9 doesn't encode the unicode data in String. This notice is in the Java 8 copy of String where the unicode data is encoded.

jcl/src/java.base/share/classes/java/lang/String.java
```diff
@@ -857,47 +667,3830 @@ public String(char[] data) {
...
+	/**
+	 * Converts the specified object to its string representation. If the object is null answer the string {@code "null"}, otherwise use
+	 * {@code toString()} to get the string representation.
+	 *
+	 * @param value
+	 *			  the object
+	 * @return the object converted to a string
+	 */
+	public static String valueOf(Object value) {
+		return value != null ? value.toString() : "null"
```
pshipton: The indenting should match the `/*[IF`

pshipton: Please qualify what this matches, and also the matching `/*[ENDIF]`

jcl/src/java.base/share/classes/java/lang/System.java
```diff
@@ -417,9 +417,6 @@ private static void ensureProperties() {
 	sun.misc.Version.init();
 	/*[ENDIF] Sidecar19-SE */
 
-	StringBuffer.initFromSystemProperties(systemProperties);
```
pshipton: Should this be removed from Java 8?

runtime/gc_base/StringTable.cpp
```diff
@@ -642,48 +642,37 @@ j9gc_createJavaLangString(J9VMThread *vmThread, U_8 *data, UDATA length, UDATA s
...
-						goto nomem;
-					} else {
-						j9object_t flag = J9AllocateObject(vmThread, flagClass, allocateFlags);
+				J9VMJAVALANGSTRING_SET_COUNT(vmThread, result, (I_32)unicodeLength);
+			}
+		} else {
+			if (J2SE_VERSION(vm) >= J2SE_19) {
+				J9VMJAVALANGSTRING_SET_CODER(vmThread, result, 1);
+			} else {
+				J9VMJAVALANGSTRING_SET_COUNT(vmThread, result, (I_32)unicodeLength);
```
pshipton: The original code had a case which set the high bit for uncompressed strings.

test/functional/Java8andUp/src_90/org/openj9/test/java/lang/Test_String_ToLowerCase_Lithuanian.java
```diff
@@ -0,0 +1,144 @@
...
+import java.util.Locale;
+import java.util.Properties;
+
+@Test(groups = { "level.sanity" })
+public class Test_String_ToLowerCase_Lithuanian {
+	/**
+	 * @tests java.lang.String#toLowerCase(java.util.Locale) using Unicode 8.0.0 standard which is supported in Java 9
+	 */
+	@Test
+	public void test_toLowerCase_lithuanian() {
```
pshipton: The test is the same between Java 8 and later. Can we have a different helper that populates the accents array but share the test code?

pshipton: Or just use the same helper but test the java version at runtime.

pshipton:
In the PR/commit comment please use different terms other than SIDECAR, such as "class". I don't think SIDECAR is used appropriately. This term is not well understood, and its use is confusing.

Since some of the changes were squashed, it seems some of the commit comment which refers to stages of development should be as well, such as the iterations to remove the count field.

Ali-Ni:
Trying to compile with these changes applied now gives the following error:
```
/jdk11/build/linux-x86_64-normal-server-release/vm/compiler/../compiler/trj9/x/codegen/J9TreeEvaluator.cpp: In static member function 'static TR::Register* J9::X86::TreeEvaluator::VMifArrayCmpEvaluator(TR::Node*, TR::CodeGenerator*)':
/jdk11/build/linux-x86_64-normal-server-release/vm/compiler/../compiler/trj9/x/codegen/J9TreeEvaluator.cpp:4978:4: error: 'SSE2IfArraycmpEvaluator' is not a member of 'TR::TreeEvaluator'
    TR::TreeEvaluator::SSE2IfArraycmpEvaluator(node->getFirstChild(), branchLabel, branchOnEqaul, cg);
```

Ali-Ni:
Yes, please ignore my last comment, it compiles fine - thanks for the help.

test/functional/Java8andUp/src_90/org/openj9/test/java/lang/Test_String_ToLowerCase_Lithuanian.java
smlambert: yes, I may have directed you without knowing enough about the difference between versions...  In this case, since the API is not different, seems you could in fact just check what version inside the helper (https://github.com/eclipse/openj9/blob/master/test/functional/TestUtilities/src/org/openj9/test/util/VersionCheck.java).

jcl/src/java.base/share/classes/java/lang/String.java
fjeremic: Fixed in 45bb385

jcl/src/java.base/share/classes/java/lang/String.java
```diff
@@ -3866,7 +6929,11 @@ private static boolean isCombiningAbove(int codePoint) {
 			int index = binarySearchRange(startCombiningAbove, (char) codePoint);
 
 			return index >= 0 && endCombiningAbove[index] >= codePoint;
-		} else if ((codePoint >= 0x1D185 && codePoint <= 0x1D189) || (codePoint >= 0x1D1AA && codePoint <= 0x1D1AD)) {
```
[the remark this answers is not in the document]
fjeremic: Fixed in 45bb385

jcl/src/java.base/share/classes/java/lang/String.java
fjeremic: Fixed in bba0fb8

jcl/src/java.base/share/classes/java/lang/String.java
fjeremic: Fixed in bba0fb8

fjeremic: Fixed in bba0fb8

jcl/src/java.base/share/classes/java/lang/System.java
fjeremic: Good catch! Fixed in 013c63c

runtime/gc_base/StringTable.cpp
fjeremic: Refactored and fixed in b506e23

test/functional/Java8andUp/src_90/org/openj9/test/java/lang/Test_String_ToLowerCase_Lithuanian.java
fjeremic: I agree. I've consolidated the tests. Fixed in 0bee8a2

fjeremic:
I'll still need to rebase and fix any of the failing tests as noted in the above comments. This will be my first task tomorrow morning.

pshipton:
Reran some testing on the latest. The remaining problem is with the DDR code, which needs to be updated for the new String shape.

The particular failure is with !dumpramclasslinear, but anything that reads Strings needs to be updated. Not sure if there is more than one place.
```
 [OUT] === Section Start: J9CPTYPE_STRING (16 bytes)                                             ===
 [OUT] Problem running command:
 [OUT] No field named count with signature I in java/lang/String
```

fjeremic:
> The particular failure is with !dumpramclasslinear, but anything that reads Strings needs to be updated. Not sure if there is more than one place.

Thanks @pshipton. I remember doing this now when we first implemented compact strings. I've patched the appropriate places in 1257183

jenkins test all plinux jdk8,jdk10

fjeremic:
> In the PR/commit comment please use different terms other than SIDECAR, such as "class". I don't think SIDECAR is used appropriately. This term is not well understood, and its use is confusing.
> 
> Since some of the changes were squashed, it seems some of the commit comment which refers to stages of development should be as well, such as the iterations to remove the count field.

I will address this as part of my final squash. I'm delaying until then so I can more easily manage the various commits and reviews in case I need to back something out.

pshipton:
> I've patched the appropriate places in 1257183

I'm willing to accept it, but just to point it out, I believe these changes will break DDR support for core files created by VMs which pre-date merging this change. Now since the OpenJ9 0.8.0 release doesn't have DDR support, these will only be nightly builds. Using DDR from the same VM will work, just not using the latest DDR.

pshipton:
Please fix the copyright problems, after we see the results of the tests.

12:29:21 debugtools/DDR_VM/src/com/ibm/j9ddr/vm29/j9/AlgorithmVersion.java
12:29:23 debugtools/DDR_VM/src/com/ibm/j9ddr/vm29/pointer/helper/J9ObjectHelper.java
12:29:24 runtime/ddr/algorithm_versions.c

fjeremic:
> Please fix the copyright problems, after we see the results of the tests.

Yes, noted. Delaying this as a force push would cancel the currently running builds. Already have commit ready. Just waiting on an all clear from the builds.

keithc-ca:
@pshipton Doesn't the new 'algorithm version' make the proper distinction between old and new core files?

fjeremic:
Jenkins test all

pshipton:
> Doesn't the new 'algorithm version' make the proper distinction between old and new core files?

I'm not sure what 'algorithm version' you are referring to. Unless there is code to explicitly handle the older core files, they won't work.

keithc-ca:
Reading the change to debugtools/DDR_VM/src/com/ibm/j9ddr/vm29/j9/AlgorithmVersion.java, I thought this was an addition instead of a removal:
```
-   public static final String STRING_COMPRESSION_VERSION = "STRING_COMPRESSION_VERSION";
```
Perhaps it (and related code) should stay and that version incremented so we can know how to interpret various core files?

pshipton:
I believe the STRING_COMPRESSION_VERSION code is obsolete for OpenJ9 (jvm.29).

We could add support for the older String format (no coder field), but it doesn't seem worth carrying that code forward forever considering
- it only affects Java 9 and forward. 9 was never supported and 10 won't be supported until the 0.9.0 release
- the last OpenJ9 release 0.8.0 doesn't support DDR

keithc-ca:
This looks good to me. I am a little concerned with the duplication that we will have in String.java, but extracting that commonality can wait for another PR.

fjeremic:
> In the PR/commit comment please use different terms other than SIDECAR, such as "class". I don't think SIDECAR is used appropriately. This term is not well understood, and its use is confusing.
> 
> Since some of the changes were squashed, it seems some of the commit comment which refers to stages of development should be as well, such as the iterations to remove the count field.

Rebased and fixed up all the comments. There are no code changes since the last builds are launched. @pshipton I think we're ready for a final review and testing.

pshipton:
jenkins compile xlinux jdk8,jdk10

pshipton:
The commit comment could use some clarity about removal of the count field, as this is only done for Java 9 and later.

OpenJDK StringBuffer / StringBuilder are used for Java 9 forward, not 10.

"Add se9_before_b165 for coder" no longer seems relevant. 

The builds which is running on the rebased content are recorded, so feel free up push fixes.
https://ci.eclipse.org/openj9/job/PullRequest-Compile-JDK8-linux_ppc-64_cmprssptrs_le-OpenJ9/44/
https://ci.eclipse.org/openj9/job/PullRequest-Compile-JDK10-linux_x86-64_cmprssptrs-OpenJ9/2/

pshipton:
Is there a plan to re-enable JIT toUpper/toLower acceleration? Or at least an Issue to reflect that something should be done.

fjeremic:
> The commit comment could use some clarity about removal of the count field, as this is only done for Java 9 and later.
> 
> OpenJDK StringBuffer / StringBuilder are used for Java 9 forward, not 10.

Fixed in the commit and the PR description:

>This change removes the count field from the String class in Java 9
>onward. The String class in Java 8 is left unmodified. The count
>field represents the length of the string in the underlying array.
>Instead of using a separate field to represent the length we size the
>underlying array to be the same length as the string.

> "Add se9_before_b165 for coder" no longer seems relevant.

Removed from the commit and PR description.

> Is there a plan to re-enable JIT toUpper/toLower acceleration? Or at least an Issue to reflect that something should be done.

I'll be opening up an issue for this momentarily. This will be my immediate next task (to re-enable support).

pshipton:
> Please update PR comment as well.

My bad, I see it now.

fjeremic:
> "Add se9_before_b165 for coder" no longer seems relevant.
> 
>         I believe there's still a use of that in this PR. Maybe in the vmconstantpool?
> 
> Correct, I didn't make the connection.
> 
> Please update PR comment as well.

I'm slightly confused. Would you like me to re-instate the comment regarding `se9_before_b165` or leave things as is?

pshipton:
> I'm slightly confused. Would you like me to re-instate the comment regarding se9_before_b165 or leave things as is?

Its fine the way it is. The change is still visible in the code, I don't think the comment added much value.

PR #1790, opened 2018-04-27
Update Valhalla nestmates spec flags

taliamccormick:
- Turn off module_ddr flag


taliamccormick:
@pshipton

PR #1793, opened 2018-04-27
Atomic-free JNI work

gacholio:
- Remove unnecessary enter/exit VM from critical wrappers
- Fix illegal use of VM access in enterCriticalRegion


gacholio:
Every change is either an addition or removal of code under my ifdef, so there's no point running any sanity builds.

PR #1794, opened 2018-04-28
Avoid catching multiple exceptions

keithc-ca:
Avoid catching multiple exceptions: DDR is also compiled with '-source 1.6'; cannot catch multiple exceptions before Java 7.


pshipton:
fyi the PR/commit subject should describe the fix and not the problem.

keithc-ca:
Commit message, etc. updated.

PR #1797, opened 2018-04-30
Increase default max heap only if memory subsystem is enabled

ashu-mehra:
JVM should use higher percentage of memory as default max heap when
running in a cgroup only if -XX:+UseContainerSupport is specified.
Note that -XX:+UseContainerSupport enables JVM to use cgroup's memory
subsystem limits.
However, a missing check for memory subsystem in
MM_GCExtensions::computeDefaultMaxHeap() causes this change to be
enabled by default.
Fix is to add code for checking memory subsystem is enabled before using
higher percentage of cgroup memory limit as default max heap.

### DIFF
--- a/runtime/gc_base/GCExtensions.cpp
+++ b/runtime/gc_base/GCExtensions.cpp
@@ -235,21 +235,24 @@ MM_GCExtensions::computeDefaultMaxHeap(MM_EnvironmentBase *env)
 
 	MM_GCExtensionsBase::computeDefaultMaxHeap(env);
 
-	int32_t rc = omrsysinfo_cgroup_get_memlimit(&cgroupMemory);
-
+	if (OMR_CGROUP_SUBSYSTEM_MEMORY == omrsysinfo_cgroup_are_subsystems_enabled(OMR_CGROUP_SUBSYSTEM_MEMORY)) {
 /* A very rough estimate of the minimum amount of memory required by the JVM excluding
  * the heap. The size is a conservative estimate, attempting to leave room for the JVM's
  * internal requirements given that one compilation thread can use up to 256M.
  */
 #define OPENJ9_IN_CGROUP_NATIVE_FOOTPRINT_EXCLUDING_HEAP ((U_64)512 * 1024 * 1024)
-	if (0 == rc) {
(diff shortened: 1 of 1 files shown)


ashu-mehra:
I realized the change done in #1517 has become a default behavior.
Ideally it should be protected by `-XX:+UseContainerSupport` option.

```
$ docker run -m2g --rm -it adoptopenjdk/openjdk8-openj9:nightly java -verbose:sizes -version | grep Xmx
WARNING: Your kernel does not support swap limit capabilities or the cgroup is not mounted. Memory limited without swap.
  -Xmx1536M       memory maximum
```

Note that Xmx gets set to 1.5 G even though I didn't specify `-XX:+UseContainerSupport` option.
To address this, I added a check for memory subsystem is enabled in `MM_GCExtensions::computeDefaultMaxHeap()`.
We could have also added check for `-XX:+UseContainerSupport` but I feel depending on memory subsystem is more appropriate.

charliegracie:
I would expect that the JVM uses less memory by default in a container.  This has been a significant feature for OpenJ9 compared to other JVMs.  We should not be increasing the default Xmx when running in a container.

Can you provide more details on the rationale behind #1517?

ashu-mehra:
@charliegracie #1429 details the reason for the change made in #1517.
Please correct me if I am wrong but as per my understanding memory advantage of OpenJ9 is more because of conservative approach towards heap expansion, which wouldn't change even if default Xmx is increased in container. Isn't that the case?

ashu-mehra:
> The important part I guess is that we do this iif the container is configured with a max. Is there a default max for docker containers? Can you distinguish between the default value and a value set via the command line / at build time?

There is no default max for docker containers. If the user doesn't specify memory limit for the docker container, it can use as much as it wants, just like any other process running on the host.
The change done in #1517 kicks into action only if the user is running docker container with memory limit. Eg if user runs `docker run <container>` (i.e. no memory limit) then Xmx would not be increased, but if user runs `docker run -m2g <container>` (i.e. 2G memory limit) then Xmx would be set to 75% of 2G.

DanHeidinga:
@charliegracie This looks like an improvement over the current code which should have included this check from the beginning.  As this is GC code, I'll defer to you or @dmitripivkine but I this should be merged.

PR #1801, opened 2018-04-30
Check module visibility in method handles

pdbain-ibm:
Ensure target classes, paramater types, and return types are visible.

### DIFF
--- a/jcl/src/java.base/share/classes/java/lang/invoke/MethodHandles.java
+++ b/jcl/src/java.base/share/classes/java/lang/invoke/MethodHandles.java
@@ -47,23 +47,25 @@ import com.ibm.oti.lang.ArgumentHelper;
 import com.ibm.oti.vm.VM;
 import com.ibm.oti.vm.VMLangAccess;
 
-/*[IF Sidecar19-SE]
+/*[IF Sidecar19-SE]*/
 import java.util.AbstractList;
 import java.util.Iterator;
 import java.util.ArrayList;
 import java.nio.ByteOrder;
 import jdk.internal.reflect.CallerSensitive;
 import java.lang.invoke.VarHandle.AccessMode;
-/*[IF Sidecar19-SE-OpenJ9]
+/*[IF Sidecar19-SE-OpenJ9]*/
 import java.lang.Module;
 import jdk.internal.misc.SharedSecrets;
 import jdk.internal.misc.JavaLangAccess;
 import java.security.ProtectionDomain;
 import jdk.internal.org.objectweb.asm.ClassReader;
+/*[ELSE] Sidecar19-SE-OpenJ9
+import java.lang.reflect.Module;
 /*[ENDIF] Sidecar19-SE-OpenJ9*/
-/*[ELSE]*/
+/*[ELSE] Sidecar19-SE*/
 import sun.reflect.CallerSensitive;
-/*[ENDIF]*/
+/*[ENDIF] Sidecar19-SE*/
 
 /**
  * Factory class for creating and adapting MethodHandles.
@@ -388,10 +390,28 @@ public class MethodHandles {
 		 * 				 to avoid extra type checking, the parameter is {@link MethodHandle}. If we need to
 		 * 				 handle {@link VarHandle} in the future, the type can be {@link Object}.
 		 * @throws IllegalAccessException If the member is not accessible.
+		 * @throws IllegalAccessError If a handle argument or return type is not accessible.
 		 */
 		private void checkAccess(Class<?> targetClass, String name, int memberModifiers, MethodHandle handle) throws IllegalAccessException {
 			checkClassAccess(targetClass);
 
+			/*[IF Sidecar19-SE]*/
+			if (null != handle) {
+				MethodType type = handle.type();
+				Module accessModule = accessClass.getModule();
+
+				try {
+					checkClassModuleVisibility(accessMode, accessModule, type.returnType);
+					for (Class<?> c: type.arguments) {
+						checkClassModuleVisibility(accessMode, accessModule, c);
+					}
+				} catch (IllegalAccessException exc) {
+					IllegalAccessError err = new IllegalAccessError(exc.getMessage());
+					err.initCause(exc);
+					throw err;
+				}
+			}
+			/*[ENDIF]*/
 			if (Modifier.isPublic(memberModifiers)) {
 				/* checkClassAccess already determined that we have more than "no access" (public access) */
 				return;
@@ -482,6 +502,9 @@ public class MethodHandles {
 		 * @throws IllegalAccessException If the {@link Class} is not accessible from {@link #accessClass}. 
 		 */
 		private void checkClassAccess(Class<?> targetClass) throws IllegalAccessException {
+			/*[IF Sidecar19-SE]*/
+			checkClassModuleVisibility(accessMode, accessClass.getModule(), targetClass);
+			/*[ENDIF]*/
 			if (NO_ACCESS != accessMode) {
 				/* target class should always be accessible to the lookup class when they are the same class */
 				if (accessClass == targetClass) {
@@ -625,6 +648,23 @@ public class MethodHandles {
 			return handle;
 		}
 		
+		/*[IF Sidecar19-SE]*/
+		/**
+		 * Return a MethodHandle to a virtual method.  The method will be looked up in the first argument 
+		 * (aka receiver) prior to dispatch.  The type of the MethodHandle will be that of the method
+		 * with the receiver type prepended.
+		 * 
+		 * @param clazz - the class defining the method
+		 * @param methodName - the name of the method
+		 * @param type - the type of the method
+		 * @return a MethodHandle that will do virtual dispatch on the first argument
+		 * @throws IllegalAccessException - if method is static or access is refused or clazz's package is not exported to the current module 
+		 * @throws IllegalAccessError - if the type's packages are not exported to the current module 
+		 * @throws NullPointerException - if clazz, methodName or type is null
+		 * @throws NoSuchMethodException - if clazz has no virtual method named methodName with signature matching type
+		 * @throws SecurityException - if any installed SecurityManager denies access to the method
+		 */
+		 /*[ELSE]
 		/**
 		 * Return a MethodHandle to a virtual method.  The method will be looked up in the first argument 
 		 * (aka receiver) prior to dispatch.  The type of the MethodHandle will be that of the method
@@ -639,6 +679,7 @@ public class MethodHandles {
 		 * @throws NoSuchMethodException - if clazz has no virtual method named methodName with signature matching type
 		 * @throws SecurityException - if any installed SecurityManager denies access to the method
 		 */
+		/*[ENDIF]*/
 		public MethodHandle findVirtual(Class<?> clazz, String methodName, MethodType type) throws IllegalAccessException, NoSuchMethodException {
 			nullCheck(clazz, methodName, type);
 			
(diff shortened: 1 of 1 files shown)


pdbain-ibm:
FYI @DanHeidinga @pshipton

jcl/src/java.base/share/classes/java/lang/invoke/MethodHandles.java
```diff
@@ -47,7 +47,7 @@
 import com.ibm.oti.vm.VM;
 import com.ibm.oti.vm.VMLangAccess;
 
-/*[IF Sidecar19-SE]
+/*[IF Sidecar19-SE]*/
```
DanHeidinga: why this change?

jcl/src/java.base/share/classes/java/lang/invoke/MethodHandles.java
```diff
@@ -634,14 +636,26 @@ public MethodHandle findStatic(Class<?> clazz, String methodName, MethodType typ
 		 * @param methodName - the name of the method
 		 * @param type - the type of the method
 		 * @return a MethodHandle that will do virtual dispatch on the first argument
-		 * @throws IllegalAccessException - if method is static or access is refused
+		 * @throws IllegalAccessException - if method is static or access is refused or clazz's package is not exported to the current module
```
DanHeidinga: This javadoc will be generated for both 8 & 9+.  Module should only be mentioned for 9

jcl/src/java.base/share/classes/java/lang/invoke/MethodHandles.java
```diff
@@ -634,14 +636,26 @@ public MethodHandle findStatic(Class<?> clazz, String methodName, MethodType typ
 		 * @param methodName - the name of the method
 		 * @param type - the type of the method
 		 * @return a MethodHandle that will do virtual dispatch on the first argument
-		 * @throws IllegalAccessException - if method is static or access is refused
+		 * @throws IllegalAccessException - if method is static or access is refused or clazz's package is not exported to the current module 
+		 * @throws IllegalAccessError - if the type's packages are not exported to the current module
```
DanHeidinga: module mentioned in 8

DanHeidinga:
The PR only adds this check to `findVirtual`.  What about the other `find{static,special,bind,setter,getter}` methods?  Do they need the module access checks as well?

jcl/src/java.base/share/classes/java/lang/invoke/MethodHandles.java
```diff
@@ -634,14 +636,26 @@ public MethodHandle findStatic(Class<?> clazz, String methodName, MethodType typ
...
+		 * @throws IllegalAccessError - if the type's packages are not exported to the current module 
 		 * @throws NullPointerException - if clazz, methodName or type is null
 		 * @throws NoSuchMethodException - if clazz has no virtual method named methodName with signature matching type
 		 * @throws SecurityException - if any installed SecurityManager denies access to the method
 		 */
 		public MethodHandle findVirtual(Class<?> clazz, String methodName, MethodType type) throws IllegalAccessException, NoSuchMethodException {
 			nullCheck(clazz, methodName, type);
 			
+			/*[IF Sidecar19-SE]*/
+			if (INTERNAL_PRIVILEGED != accessMode) {
```
DanHeidinga: shouldn't this be added to `checkAccess()`?

pdbain-ibm:
> What about the other find{static,special,bind,setter,getter} methods?

Possibly. We don't have tests for these yet.

jcl/src/java.base/share/classes/java/lang/invoke/MethodHandles.java
pdbain-ibm: Will fix.

jcl/src/java.base/share/classes/java/lang/invoke/MethodHandles.java
DanHeidinga: Doesn't that just move the syntax error from 9 to 8?

pdbain-ibm: No: this change eliminates most of the syntax errors in the unprocessed code.

jcl/src/java.base/share/classes/java/lang/invoke/MethodHandles.java
```diff
@@ -392,6 +394,17 @@ void checkAccess(Class<?> clazz) throws IllegalAccessException {
 		private void checkAccess(Class<?> targetClass, String name, int memberModifiers, MethodHandle handle) throws IllegalAccessException {
 			checkClassAccess(targetClass);
 
+			/*[IF Sidecar19-SE]*/
+			if (null != handle) {
+				MethodType type = handle.type();
+				Module accessModule = accessClass.getModule();
+				checkClassModuleVisibility(accessModule, targetClass, false);
+				checkClassModuleVisibility(accessModule, type.returnType, true);
```
DanHeidinga: There are two calls to `checkClassModuleVisibility`, one setting `throwError` to false while the other says true.  Which is intended?

jcl/src/java.base/share/classes/java/lang/invoke/MethodHandles.java
```diff
@@ -681,6 +717,32 @@ public MethodHandle findVirtual(Class<?> clazz, String methodName, MethodType ty
...
+		static void checkClassModuleVisibility(Module accessModule, Class<?> targetClass, boolean throwError) 
+				throws IllegalAccessError, IllegalAccessException {
+			Module targetModule = targetClass.getModule();
+			if (!Objects.equals(accessModule, targetModule)
+					&& !(targetModule.isExported(targetClass.getPackageName(), accessModule)
+							&& accessModule.canRead(targetModule))) {
+				/*[MSG "K0664", "{0} is not in an exported package"]*/
+				String message = Msg.getString("K0664", targetClass.getPackageName()); //$NON-NLS-1$
+				if (throwError) {
+					throw new IllegalAccessError(message);
```
DanHeidinga: likely better to only throw one exception and let the code that needs the Error wrap the Exception.

jcl/src/java.base/share/classes/com/ibm/oti/util/ExternalMessages-MasterIndex.properties
```diff
@@ -1160,6 +1160,7 @@ K0660="Internal error while obtaining private constructor."
 K0661="Internal error while obtaining GcInfo instance."
 K0662="maxDepth must not be negative."
 K0663="Invalid or unsupported dump agent option, cannot be triggered."
+K0664="{0} is not in an exported package"
```
DanHeidinga: Is there no existing message that can be reused?

jcl/src/java.base/share/classes/java/lang/invoke/MethodHandles.java
```diff
@@ -392,6 +394,17 @@ void checkAccess(Class<?> clazz) throws IllegalAccessException {
 		private void checkAccess(Class<?> targetClass, String name, int memberModifiers, MethodHandle handle) throws IllegalAccessException {
 			checkClassAccess(targetClass);
 
+			/*[IF Sidecar19-SE]*/
+			if (null != handle) {
+				MethodType type = handle.type();
+				Module accessModule = accessClass.getModule();
+				checkClassModuleVisibility(accessModule, targetClass, false);
+				checkClassModuleVisibility(accessModule, type.returnType, true);
+				for (Class<?> c: type.parameterArray()) {
```
DanHeidinga: `type.parameterArray()` allocates a new array - since this code is in the j.l.i package and we know it won't modify the array, better to use `type.arguments` directly

jcl/src/java.base/share/classes/java/lang/invoke/MethodHandles.java
```diff
@@ -681,6 +717,32 @@ public MethodHandle findVirtual(Class<?> clazz, String methodName, MethodType ty
...
+		/*[IF Sidecar19-SE]*/
+		/**
+		 * Check if targetClass is in a package visible from the accessModule
+		 * @param accessModule module of the referring class
+		 * @param targetClass Class which the referring class is accessing
+		 * @param throwError Select the throwable thrown if the targetClass is not visible
+		 * @throws IllegalAccessError thrown if throwError is true
+		 * @throws IllegalAccessException thrown if throwError is false
+		 */
+		static void checkClassModuleVisibility(Module accessModule, Class<?> targetClass, boolean throwError)
```
DanHeidinga: Please put the throw clauses on the same line.

jcl/src/java.base/share/classes/java/lang/invoke/MethodHandles.java
```diff
@@ -681,6 +717,32 @@ public MethodHandle findVirtual(Class<?> clazz, String methodName, MethodType ty
...
+		 * @throws IllegalAccessError thrown if throwError is true
+		 * @throws IllegalAccessException thrown if throwError is false
+		 */
+		static void checkClassModuleVisibility(Module accessModule, Class<?> targetClass, boolean throwError) 
+				throws IllegalAccessError, IllegalAccessException {
+			Module targetModule = targetClass.getModule();
+			if (!Objects.equals(accessModule, targetModule)
+					&& !(targetModule.isExported(targetClass.getPackageName(), accessModule)
+							&& accessModule.canRead(targetModule))) {
+				/*[MSG "K0664", "{0} is not in an exported package"]*/
```
DanHeidinga: Can you add the two modules to this message?  It will be easier for a user to resolve an access failure if they get the module info in the exception.

jcl/src/java.base/share/classes/java/lang/invoke/MethodHandles.java
```diff
@@ -681,6 +717,32 @@ public MethodHandle findVirtual(Class<?> clazz, String methodName, MethodType ty
...
+		 * @param accessModule module of the referring class
+		 * @param targetClass Class which the referring class is accessing
+		 * @param throwError Select the throwable thrown if the targetClass is not visible
+		 * @throws IllegalAccessError thrown if throwError is true
+		 * @throws IllegalAccessException thrown if throwError is false
+		 */
+		static void checkClassModuleVisibility(Module accessModule, Class<?> targetClass, boolean throwError) 
+				throws IllegalAccessError, IllegalAccessException {
+			Module targetModule = targetClass.getModule();
+			if (!Objects.equals(accessModule, targetModule)
```
DanHeidinga: Can you rewrite this as positive conditions?

jcl/src/java.base/share/classes/java/lang/invoke/MethodHandles.java
pdbain-ibm: The throwable depends on the context.  If the target class is not visible, we need to throw an IllegalAccessException. If the parameter or return types are not visible, we need to throw a IllegalAccessError.

jcl/src/java.base/share/classes/com/ibm/oti/util/ExternalMessages-MasterIndex.properties
pdbain-ibm: Yes.  Fixed.

jcl/src/java.base/share/classes/java/lang/invoke/MethodHandles.java
pdbain-ibm: Okay.

pdbain-ibm:
Passes designer tests.  Will run a larger regression suite tomorrow.

pdbain-ibm:
JCKL tests pass.  @DanHeidinga or @pshipton could you take another look at the code?

jcl/src/java.base/share/classes/java/lang/invoke/MethodHandles.java
```diff
@@ -625,6 +658,23 @@ public MethodHandle findStatic(Class<?> clazz, String methodName, MethodType typ
...
+		 * Return a MethodHandle to a virtual method.  The method will be looked up in the first argument 
+		 * (aka receiver) prior to dispatch.  The type of the MethodHandle will be that of the method
+		 * with the receiver type prepended.
+		 * 
+		 * @param clazz - the class defining the method
+		 * @param methodName - the name of the method
+		 * @param type - the type of the method
+		 * @return a MethodHandle that will do virtual dispatch on the first argument
+		 * @throws IllegalAccessException - if method is static or access is refused or clazz's package is not exported to the current module 
+		 * @throws IllegalAccessError - if the type's packages are not exported to the current module
```
DanHeidinga: Has the RI thrown IA**Error** from findVirtual?

The convention is usually that the VM throws IllegalAccessError and reflective style operations throw Exceptions.  We're using the same code to do double duty as the VM is calling into Java for these lookups.

I would expect exception / error conversion to happen at the boundaries.

jcl/src/java.base/share/classes/java/lang/invoke/MethodHandles.java
```diff
@@ -512,10 +533,22 @@ private void checkClassAccess(Class<?> targetClass) throws IllegalAccessExceptio
...
 				}
 			}
-			
+
 			/*[MSG "K0587", "'{0}' no access to: '{1}'"]*/
 			throw new IllegalAccessException(com.ibm.oti.util.Msg.getString("K0587", accessClass.getName(), targetClass.getName()));  //$NON-NLS-1$
 		}
+
+		void checkLookupVisibility(Class<?> targetClass) throws IllegalAccessException, IllegalAccessError {
+			if (this.equals(PUBLIC_LOOKUP)) {
```
DanHeidinga: Rather than checking is identically PUBLIC_LOOKUP, we should check for the `UNCONDITIONAL` mode access bit - see https://docs.oracle.com/javase/9/docs/api/java/lang/invoke/MethodHandles.html#publicLookup--

jcl/src/java.base/share/classes/java/lang/invoke/MethodHandles.java
```diff
@@ -681,6 +732,25 @@ public MethodHandle findVirtual(Class<?> clazz, String methodName, MethodType ty
...
 		}
+
+		/*[IF Sidecar19-SE]*/
+		/**
+		 * Check if targetClass is in a package visible from the accessModule
+		 * @param accessModule module of the referring class
+		 * @param targetClass Class which the referring class is accessing
+		 * @throws IllegalAccessException if the targetClass is not visible
+		 */
+		static void checkClassModuleVisibility(Module accessModule, Class<?> targetClass) throws IllegalAccessError, IllegalAccessException {
```
DanHeidinga: This probably also needs to take the `UNCONDITIONAL` access bit into account (ie: public lookup)

pdbain-ibm:
The current change causes a regression in another test.  I am investigating.

pdbain-ibm:
Regression was due to stale base code.  
@DanHeidinga or @pshipton would you kindly review and run a PR build?

jcl/src/java.base/share/classes/java/lang/invoke/MethodHandles.java
pdbain-ibm: > Has the RI thrown IAError from findVirtual?

Yes.  This matches the behaviour of the reference implementation.

pdbain-ibm:
@DanHeidinga or @pshipton do you have any comments on this PR?  If not, could you
please run a sanity build?

Thank you.

DanHeidinga:

PR #1805, opened 2018-04-30
Fix error message to use the correct stack map frame

fengxue-IS:
- added new flag in J9BytecodeVerificationData to track stackMapFrame
- avoid printing internal generated StackMap Table to output


fengxue-IS:
Reviewer: @ChengJin01 
FYI: @DanHeidinga

fengxue-IS:
updated commit message

ChengJin01:
The code change looks good to me. Please add comments to make it clear to reviewers.

fengxue-IS:
comments added

DanHeidinga:
Is there a test case or something that demonstrates the problem?

fengxue-IS:
This fixes an error during fallback verification (for classfile <= ver50) where the stackMapFrame is pointing incorrectly to the classfile data rather then the internally generated stackMapFrame.
(During fallback verification, we internally recompute the stackMapFrame to avoid reporting error on mismatching stackMap from bytecode changes)

DanHeidinga:
Jenkins test sanity aix jdk8

PR #1810, opened 2018-04-30
Atomic-free JNI work

gacholio:
- ensure vm state at every init stage
- optimize JNI critical enter and exit

### DIFF
--- a/runtime/vm/jvminit.c
+++ b/runtime/vm/jvminit.c
@@ -480,6 +480,7 @@ void OMRNORETURN exitJavaVM(J9VMThread * vmThread, IDATA rc)
 		PORT_ACCESS_FROM_JAVAVM(vm);
 
 #if defined(J9VM_INTERP_ATOMIC_FREE_JNI)
+		/* exitJavaVM is always called from a JNI context */
 		enterVMFromJNI(vmThread);
 		releaseVMAccess(vmThread);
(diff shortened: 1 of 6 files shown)


runtime/vm/jvminit.c
```diff
@@ -2960,6 +2961,22 @@ processVMArgsFromFirstToLast(J9JavaVM * vm)
...
+#if defined(J9VM_INTERP_ATOMIC_FREE_JNI)
+		if (mainThread->inNative) {
+			enterVMFromJNI(mainThread);
+			releaseVMAccess(mainThread);
+		} else
+#endif /* J9VM_INTERP_ATOMIC_FREE_JNI */
+		if (J9_ARE_ANY_BITS_SET(mainThread->publicFlags, J9_PUBLIC_FLAGS_VM_ACCESS)) {
+			releaseVMAccess(mainThread);
+		}
+	}
```
DanHeidinga: Does this have any effect on startup performance?

gacholio: It runs about 10 times, so I can't imagine it does.  I suspect it's mostly no-op anyway, if the stage leaves itself in the correct state on exit.

DanHeidinga: As far as we're aware, there was only a single stage that didn't obey this rule.  Sounds like it'll be pretty close to a nop

DanHeidinga:

PR #1812, opened 2018-05-01
Add missing dependency for 64-bit Windows

keithc-ca:
Without this dependency, the Windows resource compiler might see the annotations added by the getmacros script resulting in numerous complaints about illegal '@' characters. The build doesn't fail completely, because it seems the win_64 module is built a second time, after the annotations have been removed.


keithc-ca:
jenkins line endings check

pshipton:
jenkins compile plinux

pshipton:
Forgot to qualify the compile, compiling all 8,9,10,11 is overkill so I'm stopping all except 8.

PR #1816, opened 2018-05-01
Add temporary Ctrl+C hack for Windows

babsingh:
A temporary hack for Ctrl+C (SIGINT) has been added on Windows. This hack will
run shutdown hooks if an OpenJ9 Java process receives SIGINT on Windows.
This hack will be removed once the required OMR API has been
implemented. The following OMR issue needs to be closed before removing this hack:
https://github.com/eclipse/omr/issues/2332.

### DIFF
--- a/runtime/vm/jvminit.c
+++ b/runtime/vm/jvminit.c
@@ -5443,8 +5443,12 @@ protectedInitializeJavaVM(J9PortLibrary* portLibrary, void * userData)
 	 * 			This applies to Windows only for now. Since we don't have Hursley JCLs for this, we will need to provide our own support for all other cases/platforms.
 	 */
 	if (J9_ARE_NO_BITS_SET(vm->sigFlags, J9_SIG_XRS_ASYNC)) {
-		/* only register the handler if -Xrs is not present */
-		if (0 != j9sig_set_async_signal_handler(shutDownHookWrapper, vm, J9PORT_SIG_FLAG_SIGTERM)) {
+		/* Only register the handler if -Xrs is not present. This is a temporary hack,
+		 * which will be removed once the required OMR signal API has been implemented.
+		 * The following OMR issue needs to be closed before removing this hack:
+		 * https://github.com/eclipse/omr/issues/2332.
+		 */
+		if (0 != j9sig_set_async_signal_handler(shutDownHookWrapper, vm, J9PORT_SIG_FLAG_SIGTERM | J9PORT_SIG_FLAG_SIGINT)) {
 			goto error;
 		}
 	}


babsingh:
Testing this change. Will post a comment once testing is complete.

babsingh:
@pshipton I have tested that Ctrl+C works on Windows with these changes.

pshipton:
@babsingh when do you expect the Windows port library changes to be working?

runtime/vm/jvminit.c
```diff
@@ -5443,8 +5443,10 @@ protectedInitializeJavaVM(J9PortLibrary* portLibrary, void * userData)
 	 * 			This applies to Windows only for now. Since we don't have Hursley JCLs for this, we will need to provide our own support for all other cases/platforms.
 	 */
 	if (J9_ARE_NO_BITS_SET(vm->sigFlags, J9_SIG_XRS_ASYNC)) {
-		/* only register the handler if -Xrs is not present */
-		if (0 != j9sig_set_async_signal_handler(shutDownHookWrapper, vm, J9PORT_SIG_FLAG_SIGTERM)) {
+		/* Only register the handler if -Xrs is not present. This is a temporary hack,
+		 * which will be removed once the required OMR signal API has been implemented.
+		 */
+		if (0 != j9sig_set_async_signal_handler(shutDownHookWrapper, vm, J9PORT_SIG_FLAG_SIGTERM | J9PORT_SIG_FLAG_SIGINT)) {
```
pshipton: Not sure if we should accept this "hack", but if so it needs to be #ifdef'ed so it only affects OpenJ9 builds. Also the comment should be clear about which part should be removed later.

DanHeidinga: Is there an issue # for the OMR signal API implementation?  That will make it easier to know which issue needs to be closed to complete this.

babsingh: everything should be removed later. even code for `J9PORT_SIG_FLAG_SIGTERM` is a hack. the `shutDownHookWrapper` registered here gets overridden when [Terminator.java](https://github.ibm.com/runtimes/openj9-openjdk-jdk9/blob/ibm_sdk/jdk/src/java.base/windows/classes/java/lang/Terminator.java#L61) registers handlers for `SIGINT` and `SIGTERM` using `sun.misc.Signal.register`. this code shouldn't influence IBM J9's behavior.

babsingh:
> when do you expect the Windows port library changes to be working?

it will approximately take a month to port changes to Windows. i still have to address few more issues on Unix. i will be opening another pull request for new Unix changes by the end of this week. Signals should work without the new Unix changes.

runtime/vm/jvminit.c
babsingh: > Is there an issue # for the OMR signal API implementation?

Referred to OMR issue eclipse/omr#2332 in the code and pull request.

pshipton:
PR testing for Windows is not available yet.

PR #1820, opened 2018-05-02
Remove OpenJ9 implementation of sun.reflect.CallerSensitive

pshipton:
OpenJ9 does not need to re-implement this annotation class which is
provided by OpenJDK.


pshipton:

Issue #1375, opened 2018-03-07
Test cannot directly copy and run cmd that is printed in console


llxia:
Running test in build, we will get console output as following:
```
===============================================
Running test memoryCategories_0 ...
===============================================
test with NoOptions
{ mkdir -p "/tmp/bld_380527/memoryCategories_0"; \
		cd "/tmp/bld_380527/memoryCategories_0"; \
		export LIBPATH="/bluebird/builds/bld_380527/sdk/xa6480/jre/bin/../lib/default:/bluebird/builds/bld_380527/sdk/xa6480/jre/bin/../lib/amd64/default:/bluebird/builds/bld_380527/sdk/xa6480/jre/bin/j9vm:"; \
		"/bluebird/builds/bld_380527/sdk/xa6480/jre/bin/java"  -Xnocompressedrefs  \
	-cp "/bluebird/builds/bld_380527/jvmtest/test/SE80/TestConfig/resources:/bluebird/builds/bld_380527/jvmtest/test/SE80/TestConfig/lib/testng.jar:/bluebird/builds/bld_380527/jvmtest/test/SE80/TestConfig/lib/jcommander.jar:/bluebird/builds/bld_380527/jvmtest/test/SE80/Java8andUp/GeneralTest.jar" \
	org.testng.TestNG -d "/tmp/bld_380527/memoryCategories_0" "/bluebird/builds/bld_380527/jvmtest/test/SE80/Java8andUp/testng.xml" \
	-testnames memoryCategories \
	-groups level.extended \
	-excludegroups d.*.linux_x86-64,d.*.arch.x86,d.*.os.linux,d.*.bits.64,d.*.generic-all; \
	if [ $? -eq 0 ] ; then echo "memoryCategories_0""_PASSED"; else perl "-I/bluebird/builds/bld_380527/../../javatest/HEAD_379871/test/lib/perl" -mResultStore::Uploader -e "ResultStore::Uploader::upload('.',380527,532827023,'vmfarm.ottawa.ibm.com:31','results-532827023')"; echo; echo "memoryCategories_0""_FAILED"; fi; } 2>&1 | tee -a "/tmp/bld_380527/TestTargetResult"
[IncludeExcludeTestAnnotationTransformer] [INFO] exclude file is /bluebird/builds/bld_380527/jvmtest/test/SE80/TestConfig/resources/excludes/current_exclude_SE80.txt
...
... TestNG 6.10 by C?dric Beust (cedric@beust.com)
...
```
[...]

PR #1821, opened 2018-05-02
Fix for issue 1375 (exclude file not found)

smlambert:
Create tmp exclude file if EXCLUDE_FILE env var null or empty
Issue #1375 
[ci skip]
Signed-off-by: smlambert <slambert@gmail.com>

### DIFF
--- a/test/Utils/src/org/openj9/test/util/IncludeExcludeTestAnnotationTransformer.java
+++ b/test/Utils/src/org/openj9/test/util/IncludeExcludeTestAnnotationTransformer.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2001, 2017 IBM Corp. and others
+ * Copyright (c) 2001, 2018 IBM Corp. and others
  *
  * This program and the accompanying materials are made available under
  * the terms of the Eclipse Public License 2.0 which accompanies this
--- a/test/Utils/src/org/openj9/test/util/IncludeExcludeTestAnnotationTransformer.java
+++ b/test/Utils/src/org/openj9/test/util/IncludeExcludeTestAnnotationTransformer.java
@@ -43,6 +43,11 @@ public class IncludeExcludeTestAnnotationTransformer implements IAnnotationTrans
 		String excludeFile = System.getenv("EXCLUDE_FILE");
 		logger.info("exclude file is " + excludeFile);
 		try {
+			if (excludeFile == null || excludeFile.length() == 0) {
+            	File temp = File.createTempFile("exclude", ".txt");
+            	temp.deleteOnExit();
+           		excludeFile = temp.getAbsolutePath();
+         	}
 			FileReader fileReader = new FileReader(excludeFile);
 			BufferedReader bufferedReader = new BufferedReader(fileReader);
 			while ((line = bufferedReader.readLine()) != null) {
--- a/test/Utils/src/org/openj9/test/util/IncludeExcludeTestAnnotationTransformer.java
+++ b/test/Utils/src/org/openj9/test/util/IncludeExcludeTestAnnotationTransformer.java
@@ -44,10 +44,10 @@ public class IncludeExcludeTestAnnotationTransformer implements IAnnotationTrans
 		logger.info("exclude file is " + excludeFile);
 		try {
 			if (excludeFile == null || excludeFile.length() == 0) {
-            	File temp = File.createTempFile("exclude", ".txt");
-            	temp.deleteOnExit();
-           		excludeFile = temp.getAbsolutePath();
-         	}
+				File temp = File.createTempFile("exclude", ".txt");
+				temp.deleteOnExit();
+				excludeFile = temp.getAbsolutePath();
+			}
 			FileReader fileReader = new FileReader(excludeFile);
 			BufferedReader bufferedReader = new BufferedReader(fileReader);
 			while ((line = bufferedReader.readLine()) != null) {
--- a/test/Utils/src/org/openj9/test/util/IncludeExcludeTestAnnotationTransformer.java
+++ b/test/Utils/src/org/openj9/test/util/IncludeExcludeTestAnnotationTransformer.java
@@ -42,12 +42,8 @@ public class IncludeExcludeTestAnnotationTransformer implements IAnnotationTrans
 		String line = null;
 		String excludeFile = System.getenv("EXCLUDE_FILE");
 		logger.info("exclude file is " + excludeFile);
+		if (excludeFile != null) {
 		try {
-			if (excludeFile == null || excludeFile.length() == 0) {
-				File temp = File.createTempFile("exclude", ".txt");
-				temp.deleteOnExit();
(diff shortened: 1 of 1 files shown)


smlambert:
Test of whether one can run the direct testng command (defined in a `<command>` element of playlist.xml file) and not get a "Cannot instantiate... IncludeExcludeTestAnnotationTransformer..." error was tested manually.

Will run sanity test to ensure the parse of an actual exclude file still works in automated builds as expected.

smlambert:
Jenkins test sanity jdk8 plinux

smlambert:
Jenkins test sanity jdk8 plinux

test/Utils/src/org/openj9/test/util/IncludeExcludeTestAnnotationTransformer.java
```diff
@@ -43,6 +43,11 @@
 		String excludeFile = System.getenv("EXCLUDE_FILE");
 		logger.info("exclude file is " + excludeFile);
 		try {
+			if (excludeFile == null || excludeFile.length() == 0) {
+				File temp = File.createTempFile("exclude", ".txt");
+				temp.deleteOnExit();
+				excludeFile = temp.getAbsolutePath();
+			}
```
llxia: Do we have to create an empty file? Is it possible to skip read file code completely if exclude file does not exist?
```
if (excludeFile != null || excludeFile.length() > 0) {
   // read file
  try {
      FileReader fileReader = new FileReader(excludeFile);
      ...
}
```

llxia:
Talked with @sophia-guo , we think the best way to fix this is to print export in cmd from TestKitGen as we do not want inconsistency between running with make and just cmd.

When TestKitGen generate cmd, it should be something like:

```
mkdir -p "/tmp/bld_380527/memoryCategories_0"; \
		cd "/tmp/bld_380527/memoryCategories_0"; \
		export LIBPATH="/bluebird/builds/bld_380527/sdk/xa6480/jre/bin/../lib/default:/bluebird/builds/bld_380527/sdk/xa6480/jre/bin/../lib/amd64/default:/bluebird/builds/bld_380527/sdk/xa6480/jre/bin/j9vm:"; \
                export EXCLUDE_FILE=/bluebird/builds/bld_380527/jvmtest/test/SE80/TestConfig/resources/excludes/latest_exclude_SE80.txt; \
		"/bluebird/builds/bld_380527/sdk/xa6480/jre/bin/java"  -Xnocompressedrefs  \
	-cp "/bluebird/builds/bld_380527/jvmtest/test/SE80/TestConfig/resources:/bluebird/builds/bld_380527/jvmtest/test/SE80/TestConfig/lib/testng.jar:/bluebird/builds/bld_380527/jvmtest/test/SE80/TestConfig/lib/jcommander.jar:/bluebird/builds/bld_380527/jvmtest/test/SE80/Java8andUp/GeneralTest.jar" \
	org.testng.TestNG -d "/tmp/bld_380527/memoryCategories_0" "/bluebird/builds/bld_380527/jvmtest/test/SE80/Java8andUp/testng.xml" \
	-testnames memoryCategories \
	-groups level.extended \
	-excludegroups d.*.linux_x86-64,d.*.arch.x86,d.*.os.linux,d.*.bits.64,d.*.generic-all;
```

Users can set EXCLUDE_FILE to their own file if needed.

test/Utils/src/org/openj9/test/util/IncludeExcludeTestAnnotationTransformer.java
smlambert: It does not work to skip the code completely here (I tried).  I will look to see where the testng framework loads the transformer, to see if we have an option there to take action of ignoring the transformation step entirely.

sophia-guo: This is expected as error 'Cannot instantiate... IncludeExcludeTestAnnotationTransformer' was not triggered by FileReader. The null file has been caught by "FileNotFoundException".

sophia-guo: Actually I tried with  if (excludeFile!=null) { } it works. The error 'Cannot instantiate...' is caused by runtime FileReader(null), which can not be caught by exception.  If directly trying FileReader(null), you will get compile error.  FileNotFoundException only works for filename or directory is wrong, which doesn't include this null situation.

smlambert: Yes, that was my original approach, but when I 'tested' locally did not seem to work...  Perhaps a late night attempt that I did not do properly.  I have tried again, and pushed the updated change.

smlambert:
Verified locally can run the command directly:

"/team/smlamber/fix1375/jdk-9.0.4+12/bin/java"  -Xcompressedrefs -XX:+UseCompressedOops -Xjit -Xgcpolicy:gencon -Xfastresolve0  -Xss1536k \
> -cp "/team/smlamber/fix1375/openj9/test/TestConfig/scripts/testKitGen/../../../../jvmtest/TestConfig/resources:/team/smlamber/fix1375/openj9/test/TestConfig/scripts/testKitGen/../../../../jvmtest/TestConfig/lib/testng.jar:/team/smlamber/fix1375/openj9/test/TestConfig/scripts/testKitGen/../../../../jvmtest/TestConfig/lib/jcommander.jar:/team/smlamber/fix1375/openj9/test/TestConfig/scripts/testKitGen/../../../../jvmtest/functional/Java8andUp/GeneralTest.jar" \
> org.testng.TestNG -d "/team/smlamber/fix1375/openj9/test/TestConfig/scripts/testKitGen/../../../TestConfig/test_output_15253536317521/regressionFastresolve_mode150_0" "/team/smlamber/fix1375/openj9/test/TestConfig/scripts/testKitGen/../../../../jvmtest/functional/Java8andUp/testng.xml" -testnames regression \
> -groups level.sanity \
> -excludegroups d.*.linux_x86-64_cmprssptrs,d.*.arch.x86,d.*.os.linux,d.*.bits.64,d.*.generic-all; 
[IncludeExcludeTestAnnotationTransformer] [INFO] exclude file is null
...
... TestNG 6.14.2 by C?dric Beust (cedric@beust.com)
...

PASSED: testFastresolveWithCustomClassLoader
PASSED: testFastresolveWithDefaultClassLoader

    regression
    Tests run: 2, Failures: 0, Skips: 0

llxia:

llxia:
fyi @renfeiw for the TKG change discussion above.

sophia-guo:
@llxia On second thought  'TKG change' may not be needed. The inconsistency between running with make and just cmd only happens when doing 're-run'.  The purpose of 're-run' is to re-run the specific testcases, which should not be excluded( the existing exclude file is for the whole test set). Personally I think this change is enough. 
If people really want to exclude the subset tests under the specific testcases they can create their own exclude file and export it, which we need to update README.

llxia:
Sure, let's merge this change and get feedback from the users.

PR #1823, opened 2018-05-02
Atomic-free JNI work

gacholio:
In JVMTI, do not enter the VM multiple times.  Entering the VM allows
the GC to run, which corrupts any object pointers fetched between
successive enters.


DanHeidinga:
I'm not going to run a build on this as it's 99% under ifdef and the travis.ci compile covers the removed line.

PR #1824, opened 2018-05-02
Remove validateArbitraryClass calls in J9ValuePropagation.cpp

mstoodle:
There are two calls to validateArbitraryClass() in J9ValuePropagation
that are unnecessary.

The first passes the result from getComponentClassFromArrayClass()
which itself calls validateArbitraryClass() before returning a result.
The J9ValuePropagation code only needs to check to see if the returned
class is NULL.

The second uses the result from getSuperClass() which also calls
validateArbitraryClass() before returning a result. In this case,
the NULL return value is already handled by the code around this
call, so the entire clause to "bail" if validateArbitraryClass()
fails can simply be removed.


andrewcraik:
Jenkins test sanity

PR #1825, opened 2018-05-02
Remove unused validateInterfaceImplementation from inliner

mstoodle:
Vestigial call related to early AOT processing can be removed.


mstoodle:
Note that I see no dependency between the OMR and OpenJ9 changes....they aren't called in either project, so whoever goes first will either fail to extend or fail to establish the base implementation.

PR #1826, opened 2018-05-02
Add docker build support for Ubuntu 18.04 LTS

keithc-ca:
* Same set of docker files as for Ubuntu 16.04.
* All tested with a build from scratch.
* Note: Java 9 and Java 10 require use of the `--disable-warnings-as-errors` configure option.


keithc-ca:
To clarify, the warnings seen in Java 9 and 10 seem all to be about the use of the (deprecated) function `readdir64_r`:
```
jdk09/jdk/src/java.base/unix/native/libjava/TimeZone_md.c:150:5: warning: 'readdir64_r' is deprecated (declared at /usr/include/dirent.h:201) [-Wdeprecated-declarations]
```

PR #1827, opened 2018-05-03
WIP: Inline byte-reversing functions on X86

0dvictor:
X86 CodeGen now properly inlines following three functions:
java.lang.Long.reverseBytes()
java.lang.Integer.reverseBytes()
java.lang.Short.reverseBytes()

### DIFF
--- a/runtime/compiler/optimizer/J9Simplifier.cpp
+++ b/runtime/compiler/optimizer/J9Simplifier.cpp
@@ -219,27 +219,6 @@ J9::Simplifier::unaryCancelOutWithChild(TR::Node *node, TR::Node *firstChild, TR
    return grandChild;
    }
 
-bool
-J9::Simplifier::isRecognizedPowMethod(TR::Node *node)
-   {
-   TR::Symbol *symbol = node->getSymbol();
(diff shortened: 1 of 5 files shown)


andrewcraik:
@0dvictor  why is this WIP and when do you plan to make this ready for merge? There is a copyright check failure you should investigate.

0dvictor:
These changes are founding from some internal benchmarks, I'll rebase, retest and double check if they are update-to-date then remove the WIP.

runtime/compiler/optimizer/J9Simplifier.cpp
```diff
@@ -266,89 +245,167 @@ J9::Simplifier::foldAbs(TR::Node *node)
    return node;
    }
 
-TR::Node *
-J9::Simplifier::simplifyiCallMethods(TR::Node * node, TR::Block * block)
+template<typename T> struct BSWAP
```
andrewcraik: Why the template - can't we handle this case in a more traditional manner? Template handling in different compiles can be highly variable so I'd like to keep away from them if there isn't a good reason for them.

PR #1830, opened 2018-05-03
Add instructions for registering msdia*.dll libraries

keithc-ca:


keithc-ca:
Advise explicitly registering the two specific libraries to avoid errors with version for ARM in Visual Studio 2013.

sxa555:
@keithc-ca What are the implications of "registering" in this way (This is a noob question as I've not heard if it before) - what sort of errors show up with an unregistered DLL?

keithc-ca:
There should be no errors registering these DLLs this way - no matter whether they were previously registered or not. Microsoft claims (see [1] & [2]) that they are registered during installation of Visual Studio, but that has not been my experience. If that claim were true, this would do nothing; if false it corrects the error.

[1] https://msdn.microsoft.com/en-us/library/t6tay6cz(v=vs.100).aspx
[2] https://docs.microsoft.com/en-us/previous-versions/visualstudio/visual-studio-2013/t6tay6cz(v=vs.120)

sxa555:
I was mostly interested in understanding the symptoms that are visible when you don't register - if we make a note of them in this issue we'll be able to search if we see it again :-)

keithc-ca:
Sorry, I guess I missed that part of your question.

If the required DLL is not registered, we'll see build errors from `ddrgen` that include this text:
```
CoCreateInstance failed with HRESULT = 80040154
```

Issue #1662, opened 2018-04-11
Add new Windows machines for pull requests and build pipelines


jdekonin:
Four new Windows 2012R2 machines have been created and in need configuring with the ansible scripts from AdoptOpenJDK.

1. oj9-win2012r2-1; 169.55.170.69
1. oj9-win2012r2-2; 169.55.150.148
1. oj9-win2012r2-3; 169.55.170.74
1. oj9-win2012r2-4; 169.55.150.158

Administrator login passwords available under private communication.

Each one is 8 CPU, 8GB RAM, Windows 2012R2, 100Gb boot disk, 25 Gb data disk, 100 Mbs network link.

- [x] PR for JenkinsFiles #1778 
- [x] Ansible machines (1/4 done)
- [x] Connect to Jenkins (1/4 done)
- [x] Setup Build jobs
- [x] Setup test jobs
- [x] Setup Pipeline jobs
- [x] Setup PR jobs
- [x] README

jdekonin:
@sxa555 for installation and @AdamBrousseau for adding to jenkins and incorporating into build and pull requests.

DanHeidinga:
Many thanks to IBM for providing Windows machines for OpenJ9's CI pipeline!

sxa555:
Memo to self after a couple of discussions this evening:
- Windows OpenJ9 machines are set up manually based on the AdoptOpenJDK playbooks
  - We will also need the Debugging parts of the SDK installed as per https://github.com/AdoptOpenJDK/openjdk-build/issues/243
- There is no ansible tower (AWX) for OpenJ9
- The jenkins slave cannot be started directly via JNLP since jenkins is behind a proxy

sxa555:
@DanHeidinga I can't seem to assign this work item to me (I'm the only person around and will attempt to begin attacking it tomorrow more - today has been an information consolidation phase) - can you either (1) Add me to whatever team lets me do stuff in this repo (preferred!) or (2) Assign this item to me

DanHeidinga:
One of the annoyances here is that issues can only be assigned to openj9 committers.  We've worked around this by having contributors just comment that they are working on a particular item

sxa555:
Oh hadn't realised that even you wouldn't be able to. Grrr ... Maybe I'll go through the signup process just so I can be assigned stuff. Or use it as an excuse not to work on issues here ;-)

sxa555:
[...]

PR #1832, opened 2018-05-03
JVM Signal Handling Support [Part 2]

babsingh:
5 commits added to extend/improve JVM signal handling support.

1) **Add new function registerOSHandler**

    registerOSHandler allows the JVM to register a signal handler function
    with the OS via omrsig_register_os_handler. OMR signal source code to
    restore the original OS handler during shutdown won't work properly if
    we directly register a handler with the OS using OMRSIG_SIGACTION or
    OMRSIG_SIGNAL. By using omrsig_register_os_handler, we make sure that
    the original OS handler is properly cached and restored by OMR.

    registerOSHandler is added to J9InternalFunctions since it will be
    invoked from "jvm.c::JVM_RegisterSignal".

2) **Fix comments and formatting**

3) **Use registerOSHandler in JVM_RegisterSignal**

    Use registerOSHandler in JVM_RegisterSignal instead of directly
    registering a handler using OMRSIG_SIGACTION or OMRSIG_SIGNAL. This will
    fix a potential OMR bug related to caching and restoring original OS
    handlers during shutdown.

4) **Fix usage of omrsig_set_async_signal_handler**

    omrsig_set_async_signal_handler's signature has been changed to return
    an int32_t since it sets return value to OMRPORT_SIG_ERROR (-1) in some
    cases.

    OpenJ9 source code has been amended to accept an int32_t return value
    from omrsig_set_async_signal_handler.

5) **Fix code to remove predefinedHandlerWrapper**

    Since j9sig_set_single_async_signal_handler was used to register
    predefinedHandlerWrapper, j9sig_set_single_async_signal_handler must be
    used to remove/free predefinedHandlerWrapper.

### DIFF
--- a/runtime/vm/jvminit.c
+++ b/runtime/vm/jvminit.c
@@ -6231,13 +6231,14 @@ shutDownHookWrapper(struct J9PortLibrary* portLibrary, U_32 gpType, void* gpInfo
  * Invoke jdk.internal.misc.Signal.dispatch(int number) in Java 9 and
  * onwards. Invoke sun.misc.Signal.dispatch(int number) in Java 8.
  *
- * @param vmThread pointer to a J9VMThread
- * @param signal integer value of the signal
+ * @param[in] vmThread pointer to a J9VMThread
+ * @param[in] signal integer value of the signal
  *
  * @return void
  */
 static void
-signalDispatch(J9VMThread *vmThread, I_32 signal) {
+signalDispatch(J9VMThread *vmThread, I_32 signal)
+{
 	J9JavaVM *vm = vmThread->javaVM;
 	J9NameAndSignature nas = {0};
 	I_32 args[] = {signal};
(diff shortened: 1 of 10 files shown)


babsingh:
Jenkins test sanity depends eclipse/omr#2524

babsingh:
@pshipton these changes can be reviewed.

pshipton:
I didn't see any concerns.

runtime/vm/jvminit.c
```diff
@@ -6323,7 +6325,8 @@ predefinedHandlerWrapper(struct J9PortLibrary *portLibrary, U_32 gpType, void *g
 }
 
 IDATA
-registerPredefinedHandler(J9JavaVM *vm, U_32 signal, void **oldOSHandler) {
+registerPredefinedHandler(J9JavaVM *vm, U_32 signal, void **oldOSHandler)
```
DanHeidinga: Do we need two copies of this code?  Could this one call `registerOSHandler` with `predefinedHandlerWrapper` as the `newOSHandler`?  Or can we push knowledge of which handler to pass to the caller?

babsingh: `registerPredefinedHandler` sets OS handler as `omrsignal.c::masterASynchSignalHandler` [not exposed externally], and adds `predefinedHandlerWrapper` to the `asyncHandlerList` list. `predefinedHandlerWrapper` is not an OS handler.

`registerOSHandler` registers the user-provided OS handler with specified signal.

we can merge `registerOSHandler` and `registerPredefinedHandler`. use `newOSHandler` as `J9_PRE_DEFINED_HANDLER_CHECK(2)` to specify the `predefinedHandlerWrapper`. should i proceed with this?

DanHeidinga: You're right.  The two functions are subtly different.  Better to leave them as two different calls.

DanHeidinga:
Once the OMR side is merged and a build passes, this should be good to go.

babsingh:
OMR changes have been merged: eclipse/omr#2524. They should be in the OpenJ9 pipeline by the end of the day.

pshipton:
jenkins test all xlinux,win jdk10

pshipton:
jenkins test all aix jdk8

pshipton:
The Windows failures are timeouts due to not having any Windows machines to test on right now.

pshipton:
I'm thinking Windows machines will be available shortly, #1662. There was one but it got rebooted and didn't restart properly.

pshipton:
Do you want to do any squashing before this is merged?

pshipton:
Windows fails because there are no machines.

PR #1855, opened 2018-05-04
fix contribution guide

Bananeweizen:
Contributors don't need to become foundation members.


pshipton:
Please add a Signed-off-by footer to the commit.

pshipton:
Also you can add [ci skip] to the new commit message to avoid the unnecessary travis build.

PR #1868, opened 2018-05-08
Add valgrind support

Varun-garg:
#### Changes

With these changes Openj9 runs cleanly on valgrind using interpretor (`-Djava.compiler=NONE` or `-Xint`. This goes alongside: https://github.com/eclipse/omr/pull/1759

1. Added `opt_valgrindSupport(J9VM_OPT_VALGRIND_SUPPORT)` option.
2. When J9VM_OPT_VALGRIND_SUPPORT is enabled inline allocation is skipped to fall back to omr.
3. Fixed 2 valgrind warnings in runtime/rasdump/dmpsup.c and runtime/vm/vmthread.c

#### Questions:

1. How can I set option in OpenJ9 and OMR from OpenJDK without altering openJDK code? I did some changes [here](https://github.com/ibmruntimes/openj9-openjdk-jdk9/compare/openj9...Varun-garg:openj9). Do I have to request for these changes to OpenJDK community?

2. I want to know [default initialization](https://docs.oracle.com/javase/specs/jls/se7/html/jls-4.html#jls-4.12.5) is happening and if it is happening during runtime. If it is happening during runtime, Valgrind could know that and user would be wanned of it. Example
```java
class Z {
    static int peek() { return j; }
    static int i = peek();
    static int j = 1;
}
class Test {
    public static void main(String[] args) {
        System.out.println(Z.i);
    }
}
````
This gives output as `0` instead of `1`.  User can be warned of this if valgrind is informed.

Signed-off-by: Varun Garg <varun.10@live.com>

### DIFF
--- a/runtime/vm/vmthread.c
+++ b/runtime/vm/vmthread.c
@@ -1392,6 +1392,7 @@ allocateJavaStack(J9JavaVM * vm, UDATA stackSize, J9JavaStack * previousStack)
 		stack->size = stackSize;
 		stack->previous = previousStack;
 		stack->firstReferenceFrame = 0;
+		memset((void*) ( (uintptr_t) stack + sizeof(J9JavaStack)), 0, mallocSize - sizeof(J9JavaStack));
 
 
 		/* If this is a profiling VM, or verbose:stack is enabled, paint the stack with a distinctive pattern so we can
--- a/buildspecs/core.feature
+++ b/buildspecs/core.feature
@@ -131,6 +131,7 @@ SPDX-License-Identifier: EPL-2.0 OR Apache-2.0 OR GPL-2.0 WITH Classpath-excepti
 		<flag id="opt_phpSupport" value="false"/>
 		<flag id="opt_romImageSupport" value="true"/>
 		<flag id="opt_useOmrDdr" value="false"/>
+		<flag id="opt_valgrindSupport" value="false"/>
 		<flag id="opt_veeSupport" value="false"/>
 		<flag id="opt_vmLocalStorage" value="true"/>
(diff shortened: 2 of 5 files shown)

### CODE REVIEW

**Comment by DanHeidinga** [comment]
The default initialization is something that should be detected by static analysis.  

The bytecodes for the static initializer (`<clinit>`) show:
```
$ javap -c -v Z.class
...
  static {};
    descriptor: ()V
    flags: (0x0008) ACC_STATIC
    Code:
      stack=1, locals=0, args_size=0
         0: invokestatic  #3                  // Method peek:()I
         3: putstatic     #4                  // Field i:I
         6: iconst_1
         7: putstatic     #2                  // Field j:I
        10: return
```

We don't need to track whether a field has been default initialized vs will be initialized at runtime in the VM.  In OpenJ9, static fields are memset to 0 as part of allocating the class meta data (J9Class).  The VM only tracks the initialization status of the class as a whole, not the individual static fields.

File: runtime/vm/vmthread.c [hunk 1]
```diff
@@ -1392,6 +1392,7 @@ allocateJavaStack(J9JavaVM * vm, UDATA stackSize, J9JavaStack * previousStack)
 		stack->size = stackSize;
 		stack->previous = previousStack;
 		stack->firstReferenceFrame = 0;
+		memset((void*) ( (uintptr_t) stack + sizeof(J9JavaStack)), 0, mallocSize - sizeof(J9JavaStack));
```
DanHeidinga: The bytecode verifier already validates that bytecodes can't access uninitialized memory on the Java stack.  memseting the stack is an unnecessary runtime overhead.

genie-openj9:

PR #1870, opened 2018-05-08
Use 'usablePhysicalMemory' when calculating default max heap

ashu-mehra:
MM_GCExtensions::computeDefaultMaxHeap() currently uses cgroup memory
limit for computing heap limit when running in container.
However, this does not take into account any address space limit.
This can be fixed if we use MM_GCExtensionsBase::usablePhysicalMemory
which is set using sysinfo_get_addressable_physical_memory() that takes
into account cgroup memory limit (when running in container) as well as
address space limit.

### DIFF
--- a/runtime/gc_base/GCExtensions.cpp
+++ b/runtime/gc_base/GCExtensions.cpp
@@ -231,27 +231,21 @@ void
 MM_GCExtensions::computeDefaultMaxHeap(MM_EnvironmentBase *env)
 {
 	OMRPORT_ACCESS_FROM_OMRPORT(env->getPortLibrary());
-	uint64_t cgroupMemory = 0;
 
 	MM_GCExtensionsBase::computeDefaultMaxHeap(env);
 
 	if (OMR_CGROUP_SUBSYSTEM_MEMORY == omrsysinfo_cgroup_are_subsystems_enabled(OMR_CGROUP_SUBSYSTEM_MEMORY)) {
-/* A very rough estimate of the minimum amount of memory required by the JVM excluding
- * the heap. The size is a conservative estimate, attempting to leave room for the JVM's
- * internal requirements given that one compilation thread can use up to 256M.
- */
-#define OPENJ9_IN_CGROUP_NATIVE_FOOTPRINT_EXCLUDING_HEAP ((U_64)512 * 1024 * 1024)
-
-		int32_t rc = omrsysinfo_cgroup_get_memlimit(&cgroupMemory);
-		if (0 == rc) {
+		if (omrsysinfo_cgroup_is_memlimit_set()) {
 			/* If running in a cgroup with memory limit > 1G, reserve at-least 512M for JVM's internal requirements
 			 * like JIT compilation etc, and extend default max heap memory to at-most 75% of cgroup limit.
+			 * The value reserved for JVM's internal requirements excludes heap. This value is a conservative
+			 * estimate of the JVM's internal requirements, given that one compilation thread can use up to 256M.
 			 */
-			memoryMax = (uintptr_t)OMR_MAX((int64_t)(cgroupMemory / 2), (int64_t)(cgroupMemory - OPENJ9_IN_CGROUP_NATIVE_FOOTPRINT_EXCLUDING_HEAP));
-			memoryMax = (uintptr_t)OMR_MIN(memoryMax, (cgroupMemory / 4) * 3);
-		}
-
+#define OPENJ9_IN_CGROUP_NATIVE_FOOTPRINT_EXCLUDING_HEAP ((U_64)512 * 1024 * 1024)
+			memoryMax = (uintptr_t)OMR_MAX((int64_t)(usablePhysicalMemory / 2), (int64_t)(usablePhysicalMemory - OPENJ9_IN_CGROUP_NATIVE_FOOTPRINT_EXCLUDING_HEAP));
+			memoryMax = (uintptr_t)OMR_MIN(memoryMax, (usablePhysicalMemory / 4) * 3);
 #undef OPENJ9_IN_CGROUP_NATIVE_FOOTPRINT_EXCLUDING_HEAP
+		}
 	}
 
 #if defined(OMR_ENV_DATA64)


runtime/gc_base/GCExtensions.cpp
```diff
@@ -247,19 +247,17 @@ MM_GCExtensions::computeDefaultMaxHeap(MM_EnvironmentBase *env)
 			/* If running in a cgroup with memory limit > 1G, reserve at-least 512M for JVM's internal requirements
 			 * like JIT compilation etc, and extend default max heap memory to at-most 75% of cgroup limit.
 			 */
-			memoryMax = (uintptr_t)OMR_MAX((int64_t)(cgroupMemory / 2), (int64_t)(cgroupMemory - OPENJ9_IN_CGROUP_NATIVE_FOOTPRINT_EXCLUDING_HEAP));
-			memoryMax = (uintptr_t)OMR_MIN(memoryMax, (cgroupMemory / 4) * 3);
+			memoryMax = (uintptr_t)OMR_MAX((int64_t)(usablePhysicalMemory / 2), (int64_t)(cgroupMemory - OPENJ9_IN_CGROUP_NATIVE_FOOTPRINT_EXCLUDING_HEAP));
```
dmitripivkine: should cgroupMemory be changed to usablePhysicalMemory?

runtime/gc_base/GCExtensions.cpp
```diff
@@ -241,25 +240,23 @@ MM_GCExtensions::computeDefaultMaxHeap(MM_EnvironmentBase *env)
  * internal requirements given that one compilation thread can use up to 256M.
  */
 #define OPENJ9_IN_CGROUP_NATIVE_FOOTPRINT_EXCLUDING_HEAP ((U_64)512 * 1024 * 1024)
-
+		uint64_t cgroupMemory = 0;
```
dmitripivkine: There is no usage of cgroupMemory in this code. Why do we need a call of omrsysinfo_cgroup_get_memlimit() here?

ashu-mehra: call to `omrsysinfo_cgroup_get_memlimit()` would tell us if there is any memory limit imposed by cgroup in which case we set the heap to large size.

dmitripivkine: ok, I see. RC is still be in use

runtime/gc_base/GCExtensions.cpp
```diff
@@ -241,25 +240,23 @@ MM_GCExtensions::computeDefaultMaxHeap(MM_EnvironmentBase *env)
...
 		int32_t rc = omrsysinfo_cgroup_get_memlimit(&cgroupMemory);
 		if (0 == rc) {
 			/* If running in a cgroup with memory limit > 1G, reserve at-least 512M for JVM's internal requirements
 			 * like JIT compilation etc, and extend default max heap memory to at-most 75% of cgroup limit.
 			 */
-			memoryMax = (uintptr_t)OMR_MAX((int64_t)(cgroupMemory / 2), (int64_t)(cgroupMemory - OPENJ9_IN_CGROUP_NATIVE_FOOTPRINT_EXCLUDING_HEAP));
-			memoryMax = (uintptr_t)OMR_MIN(memoryMax, (cgroupMemory / 4) * 3);
+			memoryMax = (uintptr_t)OMR_MAX((int64_t)(usablePhysicalMemory / 2), (int64_t)(usablePhysicalMemory - OPENJ9_IN_CGROUP_NATIVE_FOOTPRINT_EXCLUDING_HEAP));
+			memoryMax = (uintptr_t)OMR_MIN(memoryMax, (usablePhysicalMemory / 4) * 3);
```
dmitripivkine: if we can not remove omrsysinfo_cgroup_get_memlimit() call so I think would be better DO NOT change cgroupMemory to usablePhysicalMemory. In case of using if cgroupMemory it is easy to read - requested and used directly.
Changing it to usablePhysicalMemory means we assume that usablePhysicalMemory must be equal cgroupMemory if rc is 0. (Is this statement is correct? Is there any expanded logic in sysinfo_get_addressable_physical_memory() you want to apply?!). If these two values are equal I would prefer keep code as is (with exception to reduce the scope for cgroupMemory variable)

ashu-mehra: Use of `cgroupMemory` for calculating `memoryMax` is incorrect because it does not take into account address space limit, if any. Value returned by `sysinfo_get_addressable_physical_memory` takes into account cgroup memory (when use of cgroup is enabled) as well as the address space limit.

> Changing it to usablePhysicalMemory means we assume that usablePhysicalMemory must be equal cgroupMemory if rc is 0

This is correct only if there is no address space limit. If there is a limit on address space then `usablePhysicalMemory` would be different than `cgroupMemory` but `rc` would still be 0.

dmitripivkine: Well, it is even worse... if omrsysinfo_cgroup_get_memlimit() does not return proper value looks like problem should be solved on level of Port library. We should resolve this mess first

ashu-mehra: I would say `omrsysinfo_cgroup_get_memlimit` is expected to return the actual memory limit imposed by the cgroup. Correct API to get physical memory is `sysinfo_get_addressable_physical_memory`.
We end up using `omrsysinfo_cgroup_get_memlimit` only to check if there is any cgroup memory limit. I think it would have been more clear if there had been an API like `omrsysinfo_cgroup_is_memlimit_set`. But as of now such an API is not present, so we rely on `omrsysinfo_cgroup_get_memlimit`.

As of now there are no direct users of `omrsysinfo_cgroup_get_memlimit`, but there is use-case for providing cgroup informatoin in javacore in which case I expect this API would be called directly.

dmitripivkine: @charliegracie Do you have suggestion how we can improve this code?

dmitripivkine: Would you please add comment explaining why usablePhysicalMemory should be used instead of cgroupMemory?

runtime/gc_base/GCExtensions.cpp
```diff
@@ -231,35 +231,34 @@ void
...
 
 	if (OMR_CGROUP_SUBSYSTEM_MEMORY == omrsysinfo_cgroup_are_subsystems_enabled(OMR_CGROUP_SUBSYSTEM_MEMORY)) {
-/* A very rough estimate of the minimum amount of memory required by the JVM excluding
- * the heap. The size is a conservative estimate, attempting to leave room for the JVM's
- * internal requirements given that one compilation thread can use up to 256M.
- */
-#define OPENJ9_IN_CGROUP_NATIVE_FOOTPRINT_EXCLUDING_HEAP ((U_64)512 * 1024 * 1024)
-
+		uint64_t cgroupMemory = 0;
 		int32_t rc = omrsysinfo_cgroup_get_memlimit(&cgroupMemory);
```
charliegracie: Why are we calling this API since we are not using the value?

ashu-mehra: We actually need to determine if cgroup memory limit is set or not. We currently don't have any API for that but we can achieve that using `omrsysinfo_cgroup_get_memlimit`. We are only interested in its return value.

charliegracie: Hmmmmm... since we want to know this shouldn't we add an API for that?  The API you are calling seems like overkill and not very useful.  Maybe we should delete this API and create one that just gives you the information you are looking for.

ashu-mehra: well, actually an API to determine if cgroup memory limit is set or not would end up doing the same set of steps that `omrsysinfo_cgroup_get_memlimit` is doing. Only difference would be it wouldn't return the memory limit, but just return true or false.

ashu-mehra: Probably I shouldn't have added `omrsysinfo_cgroup_get_memlimit()` as an API, but some internal function (since it is used by `sysinfo_get_physical_memory`) and exposed it only when required.
Currently we don't have any other use for this API, but it could be used for https://github.com/eclipse/openj9/issues/1932.
If you feel strongly about removing this API, then I can do that, convert it into an internal function and a new API for determining if cgroup memory limit is set or not.

charliegracie: I think it would be clearer if there was an API to query if a cgroup memory limit is set.

ashu-mehra: Created https://github.com/eclipse/omr/pull/2430 for adding the API to query if a cgroup memory limit is set.

ashu-mehra:
@charliegracie I have updated the code to use new API `omrsysinfo_cgroup_is_memlimit_set`. Can you please review.

PR #1871, opened 2018-05-08
Fix inlining of Unsafe.monitorEnter/Exit()

klangman:
The JIT would use DirectToJNI for Unsafe.monitorEnter
and Unsafe.monitorExit. This would result in loadaddr
childern for the call. When inlining these calls we
use the 2nd child of the call as the monent/monexit
child, resulting an a stack reference being used as
if it was a object reference. This fix will convert
the loadaddr back to a aload when DirectToJNI was used


andrewcraik:
Jenkins test sanity

andrewcraik:
Jenkins test sanity

andrewcraik:
Failure on windows is an infrastructure issue - the change looks fine so I am going to merge this.

Issue #1588, opened 2018-03-30
java.lang.IllegalArgumentException: Constant pool index out of bounds, reading Annotations


smlambert:
Issue originally raised at AdoptOpenJDK, but seems like an openj9 issue, so raising it here (originally sser raised issue from Ingo60):

When there are a lot of custom annotations on a class, and later you want to read the annotations from the class file, the following happens (Open J9 JDK9, JDK10, linux x64):

java.lang.IllegalArgumentException: Constant pool index out of bounds
at sun.reflect.annotation.AnnotationParser.parseAnnotations(java.base@10-adoptopenjdk/AnnotationParser.java:77)
at java.lang.Class.getAnnotationCache(java.base@10-adoptopenjdk/Class.java:2646)
at java.lang.Class.getAnnotation(java.base@10-adoptopenjdk/Class.java:2294)

I'd provide an example if needed (this would be quite a bit of work though, since the annotations are compiler generated), however, I'm not even sure this is the right place to post this.

This used to work in Open j9 jdk8.

Ingo60:
Original Poster here.

Clarification: the annotation parser is apparently wrong in all Open J9 versions I tested:

    OpenJDK8-OPENJ9_x64_Linux_jdk8u162-b12_openj9-0.8.0
    OpenJDK9-OPENJ9_x64_Linux_jdk-9.181
    OpenJDK10-OPENJ9_x64_Linux_20182903

It doesn't matter whether the class with the annotation was compiled with an Open J9 compiler or any other compiler. Whether the exception is raised depends solely on the JVM the annotation parsing is running under: if it's Open J9, it raises the exception, elsewise it works.

Ingo60:
Here's code to try:

https://github.com/Frege/frege/blob/master/frege/runtime/Meta.java
https://github.com/Frege/frege/blob/master/tests/comp/Annotations.java

Here's a sample session (the `j9` is a link to the Open J9 java binary):

	ingo@freguntu:~/dev/test$ j9 -version
	openjdk version "1.8.0_162"
	OpenJDK Runtime Environment (build 1.8.0_162-b12)
	Eclipse OpenJ9 VM (build openj9-0.8.0, JRE 1.8.0 Linux amd64-64 Compressed References 20180315_120 (JIT enabled, AOT enabled)
	OpenJ9   - e24e8aa9
	OMR      - 3e8296b4
	JCL      - ee1e77df1d based on jdk8u162-b12)
       
[...]

PR #1880, opened 2018-05-09
Fix result check on readAnnotations(...)

fengxue-IS:
check if the return value is BCT_ERR_OUT_OF_ROM and return the
error to increase buffer size rather than marking annotation as
malformed.

Close: #1588 

### DIFF
--- a/runtime/bcutil/cfreader.c
+++ b/runtime/bcutil/cfreader.c
@@ -427,7 +427,11 @@ readAttributes(J9CfrClassFile * classfile, J9CfrAttribute *** pAttributes, U_32 
 			}
 
 			result = readAnnotations(classfile, annotations->annotations, annotations->numberOfAnnotations, data, dataEnd, segment, segmentEnd, &index, &freePointer, flags);
-			if ((BCT_ERR_NO_ERROR != result) || (index != end)) {
+
+			if (BCT_ERR_OUT_OF_ROM == result) {
+				/* Return out of memory error code to allocate larger buffer for classfile */
+				return result;
+			} else if ((BCT_ERR_NO_ERROR != result) || (index != end)) {
 				U_32 cursor = 0;
 				Trc_BCU_MalformedAnnotation(address);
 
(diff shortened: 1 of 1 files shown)


runtime/bcutil/cfreader.c
```diff
@@ -528,7 +530,9 @@ readAttributes(J9CfrClassFile * classfile, J9CfrAttribute *** pAttributes, U_32
 
 				result = readAnnotations(classfile, parameterAnnotations->annotations,	parameterAnnotations->numberOfAnnotations, data, dataEnd,
 						segment, segmentEnd, &index, &freePointer, flags);
-				if (result != 0) {
+				if (BCT_ERR_OUT_OF_ROM == result) {
+					return result;
+				} else if (result != 0) {
```
DanHeidinga: Can you change this to `(BCT_ERR_NO_ERROR != result)`?  Using the named error code is clearer and allows changing the value in the future

DanHeidinga: Is this missing a `index != end` check?

DanHeidinga:
Does the test in #1588 pass with this change?

runtime/bcutil/cfreader.c
fengxue-IS: This check is inside a loop, so `index != end` isn't required as it will break out and be checked in more detail, see line 543 below.
I will update the 0 to `BCT_ERR_NO_ERROR`, thanks for pointing out

fengxue-IS:
The test case mentioned in #1588 passes with this change added.

runtime/bcutil/cfreader.c
DanHeidinga: Agreed.

fengxue-IS:
updated the commit with comments, moved the check outside of the loop to be more clear/consistent, also added this check for `readTypeAnnotation(...)` as this also calls `ALLOC_CAST` and `readAnnotations(...)`

DanHeidinga:

PR #1883, opened 2018-05-10
Limit variable lifetimes to smallest enclosing scope

DanHeidinga:
Minor cleanup to move variable definitions to smallest enclosing scopes.


gacholio:

PR #1887, opened 2018-05-10
Remove DDR interactive 'limit' command

keithc-ca:
It relied on API [Thread.stop(Throwable)] that has long since been deprecated (and marked for removal starting with Java 9); there's no practical replacement implementation strategy.


pshipton:
jenkins compile xlinux jdk8

PR #1891, opened 2018-05-10
Set OMR_CROSS_CONFIGURE properly

keithc-ca:
OpenJ9 builds are normally configured on a host compatible with the target, but can be configured on one host and built on another in which case CALLED_BY_SOURCE_ZIP is 'yes' and it is (only then) appropriate to set OMR_CROSS_CONFIGURE=yes when configuring OMR.

We don't normally want this because it causes the environment variable CC to be ignored.

### DIFF
--- a/runtime/gc_glue_java/configure_includes/configure_common.mk.ftl
+++ b/runtime/gc_glue_java/configure_includes/configure_common.mk.ftl
@@ -20,12 +20,8 @@
 # SPDX-License-Identifier: EPL-2.0 OR Apache-2.0 OR GPL-2.0 WITH Classpath-exception-2.0 OR LicenseRef-GPL-2.0 WITH Assembly-exception
 ###############################################################################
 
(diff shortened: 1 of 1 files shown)


pshipton:
jenkins compile all jdk8

pshipton:
Enabling cmake enables CALLED_BY_SOURCE_ZIP. Did you mention that you plan to fix this?
https://github.com/ibmruntimes/openj9-openjdk-jdk9/blob/d2ad4f4e061055d789c0ca579d88e34cb37cbc3e/closed/OpenJ9.gmk#L380
https://github.com/ibmruntimes/openj9-openjdk-jdk10/blob/24b9d1f27ce3b97a292a5c80418d4e69ea4afd75/closed/OpenJ9.gmk#L373

pshipton:
lgtm, I forgot to approve before merging

PR #1892, opened 2018-05-10
j.l.Class.getMethods() should select more specific method

JasonFengJ9:
`j.l.Class.getMethods()` should select more specific method

As per Java spec:
For methods with same signature (name, parameter types) and return type, only the most specific method should be selected. 
Method `N` is more specific than `M` if:
`N` is declared by a `class` and `M` is declared by an `interface`; or
`N` and `M` are both declared by either `classes` or `interfaces` and `N's` declaring type is the same as or a subtype of `M's` declaring type.
Select most specific method from a set of methods with same signature using algorithm above.
Added tests to verify this behaviour.

Reviewer: @DanHeidinga 

### DIFF
--- a/jcl/src/java.base/share/classes/java/lang/Class.java
+++ b/jcl/src/java.base/share/classes/java/lang/Class.java
@@ -1566,9 +1566,26 @@ private HashMap<MethodInfo, MethodInfo> getMethodSet(
 			/* if we are here, this is the target class, so return static and virtual methods */
 			boolean scanInterfaces = false;
 			for (Method m: methods) {
+				Class<?> mDeclaringClass = m.getDeclaringClass();
 				MethodInfo mi = new MethodInfo(m);
-				myMethods.put(mi, mi);
-				if (m.getDeclaringClass().isInterface()) {
+				MethodInfo prevMI = myMethods.put(mi, mi);
+				if (prevMI != null) {
+					/* As per Java spec:
+					 * For methods with same signature (name, parameter types) and return type,
+					 * only the most specific method should be selected. 
+					 * Method N is more specific than M if:
+					 * N is declared by a class and M is declared by an interface; or
+					 * N and M are both declared by either classes or interfaces and N's
+					 * declaring type is the same as or a subtype of M's declaring type.
+					 */
+					Class<?> prevMIDeclaringClass = prevMI.me.getDeclaringClass();
+					if ((mDeclaringClass.isInterface() && !prevMIDeclaringClass.isInterface())
+						|| (mDeclaringClass.isAssignableFrom(prevMIDeclaringClass))
+					) {
+						myMethods.put(prevMI, prevMI);
+					}
+				}
+				if (mDeclaringClass.isInterface()) {
 					scanInterfaces = true;
 					/* The vTable may contain one declaration of an interface method with multiple declarations. */
 					if (null == methodFilter) {
@@ -1589,7 +1606,7 @@ private HashMap<MethodInfo, MethodInfo> getMethodSet(
 					continue;
 				}
 				MethodInfo mi = new MethodInfo(m);
-				myMethods.put(mi, mi);				
+				myMethods.put(mi, mi);
 			}
 			addInterfaceMethods(infoCache, null, myMethods);
 		}
--- a/test/functional/Java8andUp/src_80/org/openj9/test/java/lang/Test_Class.java
+++ b/test/functional/Java8andUp/src_80/org/openj9/test/java/lang/Test_Class.java
@@ -377,6 +377,72 @@ public class Test_Class {
 		java.security.AccessController.doPrivileged(action);
 	}
 
+	static class ParentClass {
+		public static void methodPublicStatic() {}
+		public void methodPublicInstance() {}
+		static void methodPkgPrivateStatic() {}
+		void methodPkgPrivateInstance() {}
+		private static void methodPrivateStatic() {}
+		private void methodPrivateInstance() {}
+	}
+
+	static class ChildClass extends ParentClass {
+		public static void methodPublicStatic() {}
+		public void methodPublicInstance() {}
+		static void methodPkgPrivateStatic() {}
+		void methodPkgPrivateInstance() {}
+		private static void methodPrivateStatic() {}
+		private void methodPrivateInstance() {}
+	}
+
+	static interface SuperInterface {
+		void methodPublicInterface(); 
+		public static void methodPublicStatic() {};
+	}
+
+	static interface SubInterface extends SuperInterface {
+		void methodPublicInterface(); 
+		public static void methodPublicStatic() {};
+		default void defaultMethodPublicInterface() {}
+	}
+
+	static class ClassImplInterface implements SubInterface {
+		public void methodPublicInterface() {}
+		public static void methodPublicStatic() {};
+		public void defaultMethodPublicInterface() {}
+	}
+	
+	@Test
+	public void test_getMethods_subtest3() {
+		final Method[] methods = ChildClass.class.getMethods();
+		for (Method method : methods) {
+			final Class<?> declaringClass = method.getDeclaringClass();
+			if (declaringClass.equals(ParentClass.class)) {
+				Assert.fail("Should not return Method " + method.getName() + " declared in class " + declaringClass.getName());
+			}
+		}
+	}
+	@Test
+	public void test_getMethods_subtest4() {
+		final Method[] methods = SubInterface.class.getMethods();
+		for (Method method : methods) {
+			final Class<?> declaringClass = method.getDeclaringClass();
+			if (declaringClass.equals(SuperInterface.class)) {
+				Assert.fail("Should not return Method " + method.getName() + " declared in class " + declaringClass.getName());
+			}
+		}
+	}
+	@Test
+	public void test_getMethods_subtest5() {
+		final Method[] methods = ClassImplInterface.class.getMethods();
+		for (Method method : methods) {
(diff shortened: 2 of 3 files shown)


test/functional/Java8andUp/src_80/org/openj9/test/java/lang/Test_Class.java
```diff
@@ -377,6 +377,70 @@ public Object run() {
...
+			if (declaringClassName.indexOf("ParentClass") != -1) {
+				Assert.fail("Should not return Method " + method.getName() + " declared in class " + declaringClassName);
+			}
+		}
+	}
+	@Test
+	public void test_getMethods_subtest4() {
+		final Method[] methods = SubInterface.class.getMethods();
+		for (Method method : methods) {
+			String declaringClassName = method.getDeclaringClass().getName();
```
DanHeidinga: Why not compare to the actual java.lang.Class rather than its name?  ie:
`if (method.getDeclaringClass().equals(SuperInterface.class)) { ...`

test/functional/Java8andUp/src_80/org/openj9/test/java/lang/Test_Class.java
```diff
@@ -377,6 +377,70 @@ public Object run() {
...
+		public void methodPublicInterface() {}
+		public static void methodPublicStatic() {};
+	}
+	
+	@Test
+	public void test_getMethods_subtest3() {
+		Method[] methods = ChildClass.class.getMethods();
+		for (Method method : methods) {
+			String declaringClassName = method.getDeclaringClass().getName();
+			if (declaringClassName.indexOf("ParentClass") != -1) {
```
DanHeidinga: Wouldn't comparing the actual class object be better?  See suggestion below

test/functional/Java8andUp/src_80/org/openj9/test/java/lang/Test_Class.java
```diff
@@ -377,6 +377,70 @@ public Object run() {
...
+				Assert.fail("Should not return Method " + method.getName() + " declared in class " + declaringClassName);
+			}
+		}
+	}
+	@Test
+	public void test_getMethods_subtest5() {
+		final Method[] methods = ClassImplInterface.class.getMethods();
+		for (Method method : methods) {
+			String declaringClassName = method.getDeclaringClass().getName(); 
+			if ((declaringClassName.indexOf("SuperInterface") != -1) || (declaringClassName.indexOf("SubInterface") != -1)) {
```
DanHeidinga: String match

test/functional/Java8andUp/src_80/org/openj9/test/java/lang/Test_Class.java
```diff
@@ -560,7 +624,7 @@ public void test_getClasses() {
 		if (classes.length != 15) {
 			for (int i = 0; i < classes.length; i++)
 				logger.debug("classes[" + i + "]: " + classes[i]);
-			Assert.fail("Incorrect class array returned");
+			Assert.fail("Incorrect class array returned: expected 15 but returned " + classes.length);
```
DanHeidinga: Awesome change!  Makes it much easier to understand the error

test/functional/Java8andUp/src_90/org/openj9/test/java/lang/Test_Class.java
```diff
@@ -353,6 +353,70 @@ public Object run() {
...
+static class ClassImplInterface implements SubInterface {
+	public void methodPublicInterface() {}
+	public static void methodPublicStatic() {};
+}
+
+@Test
+public void test_getMethods_subtest3() {
+	Method[] methods = ChildClass.class.getMethods();
+	for (Method method : methods) {
+		String declaringClassName = method.getDeclaringClass().getName();
```
DanHeidinga: same comments as the Java 8 version

jcl/src/java.base/share/classes/java/lang/Class.java
```diff
@@ -1567,7 +1567,24 @@ Method getMethodHelper(
...
+				if ( prevMI != null) {
+					/* As per Java spec:
+					 * For methods with same signature (name, parameter types) and return type,
+					 * only the most specific method should be selected. 
+					 * Method N is more specific than M if:
+					 * N is declared by a class and M is declared by an interface; or
+					 * N and M are both declared by either classes or interfaces and N's
+					 * declaring type is the same as or a subtype of M's declaring type.
+					 */
+					Class<?> miDeclaringClz = mi.me.getDeclaringClass();
```
DanHeidinga: Can we call this `mDeclaringClass` and  fetch it after the `mi` is created?  Line 1588 immediately fetches the same value so looking it up once would be better.

test/functional/Java8andUp/src_80/org/openj9/test/java/lang/Test_Class.java
```diff
@@ -377,6 +377,70 @@ public Object run() {
...
+		private void methodPrivateInstance() {}
+	}
+
+	static interface SuperInterface {
+		void methodPublicInterface(); 
+		public static void methodPublicStatic() {};
+	}
+
+	static interface SubInterface extends SuperInterface {
+		void methodPublicInterface();
```
DanHeidinga: Can you add a default method to the interface?  Maybe two - one overridden by the class and one not.

jcl/src/java.base/share/classes/java/lang/Class.java
```diff
@@ -1567,7 +1567,24 @@ Method getMethodHelper(
...
+					Class<?> prevMIDeclaringClz = prevMI.me.getDeclaringClass();
+					if ((miDeclaringClz.isInterface() && !prevMIDeclaringClz.isInterface())
+						|| (miDeclaringClz.isAssignableFrom(prevMIDeclaringClz))
+					) {
```
DanHeidinga: Does the logic below to handle interface methods need similar validation before methods are added to `myMethods`?  line ~1603

jcl/src/java.base/share/classes/java/lang/Class.java
```diff
@@ -1567,7 +1567,24 @@ Method getMethodHelper(
 			boolean scanInterfaces = false;
 			for (Method m: methods) {
 				MethodInfo mi = new MethodInfo(m);
-				myMethods.put(mi, mi);
+				MethodInfo prevMI = myMethods.put(mi, mi);
+				if ( prevMI != null) {
```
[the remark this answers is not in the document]
JasonFengJ9: Fixed.

test/functional/Java8andUp/src_80/org/openj9/test/java/lang/Test_Class.java
JasonFengJ9: Agreed, comparing with `j.l.Class`.

test/functional/Java8andUp/src_80/org/openj9/test/java/lang/Test_Class.java
JasonFengJ9: Fixed.

test/functional/Java8andUp/src_80/org/openj9/test/java/lang/Test_Class.java
JasonFengJ9: Changed to compare with `j.l.Class`.

test/functional/Java8andUp/src_90/org/openj9/test/java/lang/Test_Class.java
JasonFengJ9: Fixed for `Java 9` version as well.

jcl/src/java.base/share/classes/java/lang/Class.java
JasonFengJ9: Agreed, fetching `mDeclaringClass` once and use it two cases.

test/functional/Java8andUp/src_80/org/openj9/test/java/lang/Test_Class.java
JasonFengJ9: Added a `default method` overridden by the `class`, this test intends to verify that more specific methods (i.e., child class methods overriding parent methods) are returned by `j.l.Class.getMethods()`. 
Current tests treat any method declared by parent class/interface as a failure since all parent methods are overridden. Adding a parent method not overridden by the child class makes the verification less straightforward, and doesn't add more value to this PR either.
`Test_Class` already has `test_getMethods_subtest1()` and `test_getMethods_subtest2()` to verify the methods returned from `java.lang.Class#getMethods()`.

jcl/src/java.base/share/classes/java/lang/Class.java
JasonFengJ9: The methods in line `~1603` are from `getDeclaredMethods()` which doesn't include any parent interface methods. Hence the logic mentioned is not required.

JasonFengJ9: @DanHeidinga PR has been updated, please have another look.

DanHeidinga:
Jenkins test sanity xlinux jdk8,jdk9

DanHeidinga:
Jenkins test sanity xlinux jdk8, jdk9

DanHeidinga:

DanHeidinga:

pshipton:
The eclipe si appears to be down. https://ci.eclipse.org/

pshipton:
jenkins test all xlinux jdk8,jdk10

DanHeidinga:
jenkins test extended plinux jdk10

PR #1894, opened 2018-05-10
Change printf tracing statements in escape analysis to traceMsg's

mstoodle:
Several random printf's in escape analysis are turned on by
traceEscapeAnalysis. Change them to match the other tracing messages
in this opt (traceMsg by default with a commented out printf in
case someone wants/needs stdout).


0xdaryl:
Jenkins test sanity

PR #1899, opened 2018-05-11
Remove unused debug code from ClassUnloadAssumption.cpp

mstoodle:
Attention was drawn to this code by a warning that a return code
from fscanf wasn't being checked; was mildly shocked to learn about
an fscanf() call in ClassUnloadAssumption.cpp . After consulting
with some other knowledgeable folks, it seems this worrisome code
is unlikely to have been used since its inception, so let's just
remove it.


mpirvu:
Since this change is only about removing inactive code, comprehensive tests are not required, hence merging the code.

PR #1900, opened 2018-05-11
Turn libffi into a static link library

gacholio:
The version of libffi used in OpenJ9 is only linked by the VM library,
so make libffi a static link library instead of dynamic to avoid issues
where another version of libffi has already been loaded into the
process.

### DIFF
--- a/runtime/libffi/module.xml
+++ b/runtime/libffi/module.xml
@@ -1,6 +1,6 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <!--
-Copyright (c) 2014, 2017 IBM Corp. and others
+Copyright (c) 2014, 2018 IBM Corp. and others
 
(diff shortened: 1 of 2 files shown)


pshipton:
jenkins compile all jdk8

DanHeidinga:
Marking this as WIP due to extra testing required - particularly on zOS.

gacholio:
Tested as much as possible with the current infra issues.  Z/OS not tested at all, but there's no reason to think it's any different than the rest.

gacholio:
There is an awful and fragile workaround here which will hopefully only live for a few days.  The open builds scripts need to stop referring to the ffi shared library in any way - at that point I will remove the workaround.

gacholio:
jenkins test sanity all jdk8

PR #1902, opened 2018-05-11
Update Build_Instructions_V8.md

ChengJin01:
Update the description as to building a 32-bit OpenJ9 version on Windows

### DIFF
--- a/buildenv/Build_Instructions_V8.md
+++ b/buildenv/Build_Instructions_V8.md
@@ -490,7 +490,6 @@ Eclipse OpenJ9 VM (build master-9329b7b9, JRE 1.8.0 Windows 7 amd64-64-Bit Compr
 OpenJ9   - 9329b7b9
 OMR      - 884959f4
 JCL      - 7f27c537a8 based on jdk8u172-b11)
-
 ```
 2) Win 32-bit
 ```
@@ -500,7 +499,6 @@ Eclipse OpenJ9 VM (build master-9f924a1a, JRE 1.8.0 Windows 7 x86-32-Bit 2018051
 OpenJ9   - 9f924a1a
 OMR      - e5db96ba
 JCL      - 7f27c537a8 based on jdk8u172-b11)
-
 ```
 
 :ledger: *Congratulations!* :tada:
--- a/buildenv/Build_Instructions_V8.md
+++ b/buildenv/Build_Instructions_V8.md
@@ -403,7 +403,7 @@ bash configure --disable-ccache \
 ```
 
 :pencil: Modify the paths for freemarker and freetype if you manually downloaded and unpacked these dependencies into different directories.
-
+:ledger:
 2) Win 32-bit
 [1] Modify the script to support the 32-bit setting
 ```
@@ -441,7 +441,7 @@ openj9-openjdk-jdk8\jdk\make\closed\autoconf\generated-configure.sh
         else
 ```
 
-[3] run configure script as follows:
+[3] run the configure script as follows:
 ```
 bash configure --disable-ccache --with-boot-jdk=/cygdrive/c/temp/jdk7 --with-freemarker-jar=/cygdrive/c/temp/freemarker.jar --with-freetype-include=/cygdrive/c/temp/freetype/include --with-freetype-lib=/cygdrive/c/temp/freetype/lib32  --with-conf-name=windows-x86-normal-release  --target=x86 --with-target-bits=32 --with-noncompressedrefs
 ```
@@ -457,6 +457,7 @@ Two Java builds are produced: a full developer kit (jdk) and a runtime environme
 1) Win 64b-bit
 - **build/windows-x86_64-normal-server-release/images/j2sdk-image**
 - **build/windows-x86_64-normal-server-release/images/j2re-image**
+
 2) Win 32-bit
 - **build/windows-x86-normal-release/images/j2sdk-image**
 - **build/windows-x86-normal-release/images/j2re-image**
--- a/buildenv/Build_Instructions_V8.md
+++ b/buildenv/Build_Instructions_V8.md
@@ -299,7 +299,7 @@ OpenJDK  - 27f5b8f based on jdk8u152-b03)
 
 ## Windows
 :ledger:
-The following instructions guide you through the process of building a Windows 64-bit OpenJDK V8 binary that contains Eclipse OpenJ9. This process can be used to build binaries for Windows 7, 8, and 10.
+The following instructions guide you through the process of building a Windows 32/64-bit OpenJDK V8 binary that contains Eclipse OpenJ9. This process can be used to build binaries for Windows 7, 8, and 10.
 
 ### 1. Prepare your system
 :ledger:
@@ -313,6 +313,12 @@ OpenJ9 uses the mingw/GCC compiler during the build process. In the `Archive` ca
 - [Microsoft Visual Studio 2010 Service Pack 1](https://support.microsoft.com/en-us/help/983509/description-of-visual-studio-2010-service-pack-1)
 - [Freemarker V2.3.8](https://sourceforge.net/projects/freemarker/files/freemarker/2.3.8/freemarker-2.3.8.tar.gz/download)
 - [Freetype2 V2.3 or newer](https://www.freetype.org/)
+
+Note:
+For building a 32-bit version, please install 32-bit Cygwin and all required packages must be installed via Cygwin as follows:
+1. GNU make 3.81 (unncessary if already installed)
+2. gcc-g++ (unncessary if already installed)
+3. mingw64-i686-gcc-g++
 
 Update your `LIB` and `INCLUDE` environment variables to provide a path to the Windows debugging tools with the following commands:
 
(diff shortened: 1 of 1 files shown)


DanHeidinga:
@ChengJin01 The ip-validation check is failing. Can you add a signed-off by clause to your commit?

pshipton:
Please add a signed-off-by to your commit in order to pass the ip-validation.

buildenv/Build_Instructions_V8.md
```diff
@@ -386,6 +404,48 @@ bash configure --disable-ccache \
...
+```
+openj9-openjdk-jdk8\jdk\make\closed\autoconf\generated-configure.sh
+  if test "x$OPENJ9_CPU" = xx86-64; then
+    if test "x$OPENJDK_BUILD_OS" = xlinux; then
+      OPENJ9_PLATFORM_CODE=xa64
+    elif test "x$OPENJDK_BUILD_OS" = xwindows; then
+      OPENJ9_PLATFORM_CODE=wa64
+      if test "x$OPENJ9_LIBS_SUBDIR" = xdefault; then
+        OPENJ9_BUILDSPEC="win_x86-64"
+      if test "x$OPENJDK_TARGET_CPU_BITS" = x32; then  <----- for 32bit setting
```
DanHeidinga: Can you open a PR for each of the extensions repos with this change?  ie: https://github.com/ibmruntimes/openj9-openjdk-jdk8 https://github.com/ibmruntimes/openj9-openjdk-jdk9 etc

buildenv/Build_Instructions_V8.md
```diff
@@ -386,6 +404,48 @@ bash configure --disable-ccache \
...
+        # Now check for other permutations
+        ifeq ($(JVM_VARIANT_SERVER), true)
+	  $(PRINTF) "-j9vm KNOWN\n">>$(@)  <-------------
+		$(PRINTF) "-client IGNORE\n">>$(@)
+		$(PRINTF) "-server IGNORE\n">>$(@)
+          ifeq ($(JVM_VARIANT_MINIMAL1), true)
+	    $(PRINTF) "-minimal KNOWN\n">>$(@)
+          endif
+        else
+```
```
DanHeidinga: This change should be made at the Extensions projects.  Please open a PR for this change on 8,9,10, etc

buildenv/Build_Instructions_V8.md
```diff
@@ -386,6 +404,48 @@ bash configure --disable-ccache \
...
+
+[3] run configure script as follows:
+```
+bash configure --disable-ccache --with-boot-jdk=/cygdrive/c/temp/jdk7 --with-freemarker-jar=/cygdrive/c/temp/freemarker.jar --with-freetype-include=/cygdrive/c/temp/freetype/include --with-freetype-lib=/cygdrive/c/temp/freetype/lib32  --with-conf-name=windows-x86-normal-release  --target=x86 --with-target-bits=32 --with-noncompressedrefs
```
DanHeidinga: Are all of these required?  I would have thought `-with-target-bits=32` would make `--with-noncompressedrefs` redundant

buildenv/Build_Instructions_V8.md
```diff
@@ -314,6 +314,12 @@ OpenJ9 uses the mingw/GCC compiler during the build process. In the `Archive` ca
 - [Freemarker V2.3.8](https://sourceforge.net/projects/freemarker/files/freemarker/2.3.8/freemarker-2.3.8.tar.gz/download)
 - [Freetype2 V2.3 or newer](https://www.freetype.org/)
 
+Note:
+For building a 32-bit version, please install 32-bit Cygwin and all required packages must be installed via Cygwin as follows:
```
pshipton: Is this really necessary? If cygwin is installed as described above, is there anything missing which is required for 32-bit? These instructions imply that mingw64-i686-gcc-g++ is missing, but isn't this part of the `Devel` category? If not, what category is it in?

buildenv/Build_Instructions_V8.md
pshipton: Note for Java 8 64-bit https://github.com/ibmruntimes/openj9-openjdk-jdk8/blob/openj9/jdk/src/solaris/bin/amd64/jvm.cfg is modified. Is https://github.com/ibmruntimes/openj9-openjdk-jdk8/blob/openj9/jdk/src/solaris/bin/i586/jvm.cfg the right place for this change to support 32-bit?

buildenv/Build_Instructions_V8.md
ChengJin01: Will do

buildenv/Build_Instructions_V8.md
pshipton: We'd like to be able to install a single cygwin installation which is suitable for building for 32-bit and 64-bit.

ChengJin01:
Signed-off already added.

buildenv/Build_Instructions_V8.md
ChengJin01: Will do

buildenv/Build_Instructions_V8.md
ChengJin01: 32 bit Cygwin might be unnecessary but mingw64-i686-gcc-g++ might not be installed automatically by default.
I think it is reasonable for users to install mingw64-i686-gcc-g++ if compilation failure prompts them to do so

buildenv/Build_Instructions_V8.md
ChengJin01: I need to double-check whether "--with-noncompressedrefs" is mandatory but the rest of options are required for 32-bit.

buildenv/Build_Instructions_V8.md
pshipton: The instructions say `Install all packages in the `Devel` category, which includes `mingw`, a minimalist subset of GNU tools for Microsoft Windows.`. I'm hoping the `Devel` category includes mingw64-i686-gcc-g++, we'll have to check. Otherwise we need specific instructions for how to find it in the cygwin installer.

buildenv/Build_Instructions_V8.md
ChengJin01: Is https://github.com/ibmruntimes/openj9-openjdk-jdk8/blob/openj9/jdk/src/solaris/bin/i586/jvm.cfg the right place for this change to support 32-bit?

----> I don't think it would work because jvm.cfg on 32-bit is only generated ./jdk/make/CopyFiles.gmk instead of modifying the existing jvm.cfg.

The script in ./jdk/make/CopyFiles.gmk shows:  if the target is 64-bit, then it automatically copise the jvm.cfg to the destination. That's why we don't care on win 64-bit as all setting of jvm.cfg on 64-bit are already updated by us.

buildenv/Build_Instructions_V8.md
ChengJin01: "--with-noncompressedrefs" must be specified otherwise it will up with a bunch of compilation failures such as " error C2371: 'size_t' : redefinition; different basic type", etc.

buildenv/Build_Instructions_V8.md
pshipton: Ok. We'll want j9vm to be the only KNOWN place

buildenv/Build_Instructions_V8.md
pshipton: After everything else is done, we should change the code so --with-noncompressedrefs is automatic when -with-target-bits=32 is specified. So in the end this should not be required.

pshipton: and also check if we can simplify anything else so not all of `-with-conf-name=windows-x86-normal-release  --target=x86 --with-target-bits=32` is required. Although this is the last priority.

buildenv/Build_Instructions_V8.md
pshipton: Making this change in all the extensions repos should be the first priority, along with setting j9vm KNOWN

buildenv/Build_Instructions_V8.md
ChengJin01: -with-conf-name=windows-x86-normal-release <-----  you need to specify a name for your build; otherwise, the name will end up being 64-bit name automatically picked up by build. e.g windows-x86_64-normal-server-release

Both --target=x86  both --with-target-bits=32 are the most important options to set up all 32-bit related configuration unless we modify all the scripts related to unify these option.

I used to try "--openjdk-target=i586-pc-cygwin" to cover everything but it ended up with failure to generate a certain .exe file during compilation.

buildenv/Build_Instructions_V8.md
ChengJin01: Already double-checked that mingw64-i686-gcc-g++ is included in Devel category.

pshipton: In that case there shouldn't need to be any special note about cygwin for building 32-bit.

ChengJin01: Agreed.

ChengJin01:
I create a new pull request from eclipse as this one (created from my repo) keeps failing in ip-check.

ChengJin01:
https://github.com/eclipse/openj9/pull/1904 has been created to replace this PR.

PR #1903, opened 2018-05-11
Fix stackMapFrame index init value and increment

fengxue-IS:
- Set initial index to 0 instead of -1
- Move counter increment to after decoding each frame


gacholio:
jenkins test sanity jdk8 plinux

gacholio:

PR #1904, opened 2018-05-11
Add steps to compile 32-bit Java8 on Windows

ChengJin01:
The change is to add building steps for
32-bit Java 8 version on Windows platform.

Depends on ibmruntimes/openj9-openjdk-jdk8#84

### DIFF
--- a/buildenv/Build_Instructions_V8.md
+++ b/buildenv/Build_Instructions_V8.md
@@ -348,11 +348,22 @@ tar -xzf freemarker.tgz freemarker-2.3.8/lib/freemarker.jar --strip=2
 tar --one-top-level=/cygdrive/c/temp/freetype --strip-components=1 -xzf freetype-2.5.3.tar.gz
 ```
 
+Note:
+In order to build Windows 32-bit, Please enure the freetype version supports win 32-bit application (e.g. freetype-2.4.7 which includes the "win32" folder)
+
 - To build the Freetype dynamic and static libraries, open the Visual Studio Command Prompt (VS2010) (see C:\ProgramData\Microsoft\Windows\Start Menu\Programs\Microsoft Visual Studio 2010\Visual Studio Tools) and run:
+1) Win 64-bit
 ```
 cd c:\temp
 msbuild.exe C:/temp/freetype/builds/windows/vc2010/freetype.vcxproj /p:PlatformToolset=v100 /p:Configuration="Release Multithreaded" /p:Platform=x64 /p:ConfigurationType=DynamicLibrary /p:TargetName=freetype /p:OutDir="C:/temp/freetype/lib64/" /p:IntDir="C:/temp/freetype/obj64/" > freetype.log
 msbuild.exe C:/temp/freetype/builds/windows/vc2010/freetype.vcxproj /p:PlatformToolset=v100 /p:Configuration="Release Multithreaded" /p:Platform=x64 /p:ConfigurationType=StaticLibrary /p:TargetName=freetype /p:OutDir="C:/temp/freetype/lib64/" /p:IntDir="C:/temp/freetype/obj64/" >> freetype.log
+```
+
+2) Win 32-bit
+```
+cd c:\temp
+msbuild.exe C:/temp/freetype/builds/win32/vc2010/freetype.vcxproj /p:PlatformToolset=v100 /p:Configuration="Release Multithreaded" /p:PlatformTarget=x86 /p:ConfigurationType=DynamicLibrary /p:TargetName=freetype /p:OutDir="C:/temp/freetype/lib32/" /p:IntDir="C:/temp/freetype/obj32/" > freetype.log
+msbuild.exe C:/temp/freetype/builds/win32/vc2010/freetype.vcxproj /p:PlatformToolset=v100 /p:Configuration="Release Multithreaded" /p:PlatformTarget=x86 /p:ConfigurationType=StaticLibrary /p:TargetName=freetype /p:OutDir="C:/temp/freetype/lib32/" /p:IntDir="C:/temp/freetype/obj32/" >> freetype.log
 ```
 :pencil: Check the `freetype.log` for errors.
 
@@ -376,6 +387,7 @@ bash get_source.sh
 ### 3. Configure
 :ledger:
 When you have all the source files that you need, run the configure script, which detects how to build in the current build environment.
+1) Win 64-bit
 ```
 bash configure --disable-ccache \
                --with-boot-jdk=/cygdrive/c/temp/jdk7 \
@@ -384,6 +396,15 @@ bash configure --disable-ccache \
                --with-freetype-lib=/cygdrive/c/temp/freetype/lib64
 ```
 
+2) Win 32-bit
+```
+bash configure --disable-ccache \
+               --with-boot-jdk=/cygdrive/c/temp/jdk7 \
+               --with-freemarker-jar=/cygdrive/c/temp/freemarker.jar \
+               --with-freetype-include=/cygdrive/c/temp/freetype/include \
+               --with-freetype-lib=/cygdrive/c/temp/freetype/lib32  \
+               --target=x86 --with-target-bits=32
+```
 :pencil: Modify the paths for freemarker and freetype if you manually downloaded and unpacked these dependencies into different directories.
 
 ### 4. build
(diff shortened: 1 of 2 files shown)


ChengJin01:
Reviewer: @DanHeidinga , @pshipton 

This PR is created to replace https://github.com/eclipse/openj9/pull/1902

buildenv/Build_Instructions_V8.md
```diff
@@ -348,12 +348,23 @@ tar -xzf freemarker.tgz freemarker-2.3.8/lib/freemarker.jar --strip=2
 tar --one-top-level=/cygdrive/c/temp/freetype --strip-components=1 -xzf freetype-2.5.3.tar.gz
 ```
 
+Note:
+Please enure the freetype version supports win 32-bit application (e.g. freetype-2.4.7 which includes the "win32" folder)
```
pshipton: Please add something like "In order to build Windows 32-bit, please ensure ..."

buildenv/Build_Instructions_V8.md
```diff
@@ -348,12 +348,23 @@ tar -xzf freemarker.tgz freemarker-2.3.8/lib/freemarker.jar --strip=2
...
+
 - To build the Freetype dynamic and static libraries, open the Visual Studio Command Prompt (VS2010) (see C:\ProgramData\Microsoft\Windows\Start Menu\Programs\Microsoft Visual Studio 2010\Visual Studio Tools) and run:
+1) Win 64b-bit
```
pshipton: Looks like an unnecessary 'b' in '64b'

buildenv/Build_Instructions_V8.md
```diff
@@ -376,6 +387,7 @@ bash get_source.sh
 ### 3. Configure
 :ledger:
 When you have all the source files that you need, run the configure script, which detects how to build in the current build environment.
+1) Win 64b-bit
```
pshipton: Looks like an unnecessary 'b' in '64b'

buildenv/Build_Instructions_V8.md
```diff
@@ -394,9 +415,14 @@ make all
 ```
 
 Two Java builds are produced: a full developer kit (jdk) and a runtime environment (jre)
+1) Win 64b-bit
```
pshipton: Looks like an unnecessary 'b' in '64b'

buildenv/Build_Instructions_V8.md
```diff
@@ -411,12 +437,24 @@ Run:
 
 Here is some sample output:
 
+1) Win 64b-bit
```
pshipton: Looks like an unnecessary 'b' in '64b'

buildenv/Build_Instructions_V8.md
```diff
@@ -384,6 +396,15 @@ bash configure --disable-ccache \
                --with-freetype-lib=/cygdrive/c/temp/freetype/lib64
 ```
 
+2) Win 32-bit
+```
+bash configure --disable-ccache \
+			   --with-boot-jdk=/cygdrive/c/temp/jdk7 \
```
pshipton: Compared to 64-bit, these lines are all indented differently. Perhaps because of different tab width. Use spaces.

buildenv/Build_Instructions_V8.md
```diff
@@ -411,12 +437,24 @@ Run:
 
 Here is some sample output:
```
pshipton: The `cd build/windows-x86_64-normal-server-release/images/j2sdk-image` line above is only appropriate for 64-bit

runtime/makelib/targets.mk.ftl
```diff
@@ -128,6 +128,7 @@ all: $(TARGETS)
 # Add OMR include paths
 #  Note assumptions about the directory layout of the OMR project.
 UMA_INCLUDES+=$(UMA_PATH_TO_ROOT)nls
+UMA_INCLUDES+=$(UMA_PATH_TO_ROOT)include
```
pshipton: The original PR #1902 didn't contain this change. It doesn't seem this change would be needed to add 32-bit support. Please confirm.

buildenv/Build_Instructions_V8.md
ChengJin01: Already added.

buildenv/Build_Instructions_V8.md
ChengJin01: Fixed.

buildenv/Build_Instructions_V8.md
ChengJin01: Fixed.

buildenv/Build_Instructions_V8.md
ChengJin01: Fixed.

buildenv/Build_Instructions_V8.md
ChengJin01: Fixed.

buildenv/Build_Instructions_V8.md
ChengJin01: Replaced with spaces.

buildenv/Build_Instructions_V8.md
ChengJin01: Already updated this section

runtime/makelib/targets.mk.ftl
ChengJin01: This is part of mk/config script changes (no need to create a separate PR in the same repo). This is required by the Resource Compiler (RC) on Windows 32-bit to locate the requested header files at ./include

pshipton:
lgtm. This shouldn't be merged before the required extensions changes.

runtime/makelib/targets.mk.ftl
pshipton: Is there an associated omr PR? There is no omr 'include' directory.

runtime/makelib/targets.mk.ftl
```diff
@@ -125,6 +125,11 @@ all: $(TARGETS)
 </#if>
 <#include "targets.mk.vendor.inc.ftl">
 
+# Add runtime include path for resource compiler on Windows.
+<#if uma.spec.type.windows>
+UMA_INCLUDES+=$(UMA_PATH_TO_ROOT)include
```
pshipton: This is a scary change, as it changes the include path for everything. Its possible it could introduce weird side affects. I expect most things already #include the include directory, but something may not. It also seems like a hack, as modules that need the include directory should specify that. Can the problem be fixed in a more specific way?

ChengJin01: Agreed and already updated the option specific to RC on Windows.

runtime/makelib/targets.mk.win.inc.ftl
```diff
@@ -100,6 +100,9 @@ endif
 
 UMA_RC_OPTIONS+=$(UMA_C_INCLUDES)
 
+# Add runtime include path for resource compiler on Windows.
+UMA_RC_OPTIONS:=$(addprefix $(UMA_C_INCLUDE_PREFIX),$(UMA_PATH_TO_ROOT)include) $(UMA_RC_OPTIONS)
```
pshipton: What specifically is causing the problem? This isn't required for IBM builds that create Windows 32-bit builds. What is different about OpenJ9 builds?

pshipton: What is an example directory where the problem occurs? Can you explain why all the OpenJ9 .rc files compile in an IBM build? Perhaps the module.xml needs to be updated in some OpenJ9 specific directory?

ChengJin01: Investigation shows the problem (failure to locate openj9/runtime/include) only occurs on Cygwin 32-bit version  as follows:
win.rc(4) : fatal error RC1015: cannot open include file 'j9cfg.h'.
make[3]: *** [../makelib/targets.mk:384: win.res] Error 1
make[2]: *** [makefile:128: win] Error 2
make[1]: *** [OpenJ9.gmk:321: build-j9] Error 2
It gets compiled only if the path to  openj9/runtime/include is specified in $(UMA_C_INCLUDES).

It works good without any RC-related issue on Cygwin 64-bit version. So there is no need to change any .mk file here as we use Cygwin 64-bit version for compilation.

ChengJin01: The root cause is the include folder wasn't specified the runtime/win/module.xml (32bit) as follows:
<includes>
		<include path="j9include"/>
</includes>

With the update, the build got compiled as expected.

pshipton:
https://github.com/ibmruntimes/openj9-openjdk-jdk8/pull/84 is now merged. Should be testing Windows but not available in PR builds yet. Validate another platform.

jenkins compile xlinux jdk8

pshipton:
jenkins compile win jdk10

pshipton:
Thought I saw an available Windows machine, but I guess they aren't available for builds yet.

PR #1905, opened 2018-05-12
Change Windows ref_repo to Win style path

AdamBrousseau:
- Git clone step looks for Windows style path

Issue #1873

[skip ci]

Signed-off-by: Adam Brousseau <adam.brousseau88@gmail.com>

PR #1873, opened 2018-05-09
WIP: Add reference repository to OpenJDK clone

AdamBrousseau:
- Speeds up OpenJDK clone
- Allows for full cleanWs at the end of all builds
- Won't fail if ref repo does not exist
- Ref repo stored in ~/openjdk_cache
- Ref repo is a bare clone of extensions repos

[skip ci]

Fixes #1727

Signed-off-by: Adam Brousseau <adam.brousseau88@gmail.com>

PR #1906, opened 2018-05-12
 cleanWs after load jenkinsfiles

AdamBrousseau:
- If the OpenJ9 repo is in the Workspace then
  the git clone will not try to use the reference
  repo. Instead it will simply try and fetch
  the OpenJDK extensions repo from Github. If we
  remove the OpenJ9 repo from the workspace then
  the OpenJDK clone will use the reference repo.
- Only affects PR builds that run testing
  because we only load the OpenJ9 repo in those
  builds.

[skip ci]

Issue #1873

Depends #1905


AdamBrousseau:
Jenkins test sanity all jdk8

AdamBrousseau:
Jenkins test sanity win jdk10

PR #1907, opened 2018-05-14
Cmake simplifications

keithc-ca:
Use OPENJ9_ENABLE_CMAKE directly, instead of ENABLE_CMAKE: buildtools.mk includes spec.gmk and thus has direct access to OPENJ9_ENABLE_CMAKE. Cmake is now consistently selected if $(OPENJ9_ENABLE_CMAKE)=true and not based on whether `ENABLE_CMAKE` is defined.

GNUmakefile should not be used with cmake and so doesn't need conditional code based on ENABLE_CMAKE.

Add missing dependencies for ddr.

### DIFF
--- a/runtime/GNUmakefile
+++ b/runtime/GNUmakefile
@@ -29,9 +29,7 @@
 clean: clean_phase_omr
 
 phase_omr:
-ifneq ($(ENABLE_CMAKE),true)
 	$(MAKE) $(MFLAGS) -C omr
-endif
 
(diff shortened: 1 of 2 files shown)


pshipton:
@dnakamura are you ok with this change?

dnakamura:
Yup. LGTM

pshipton:
I forgot to run a PR build, but my understanding is this has been tested manually so I don't expect any problems.

PR #1910, opened 2018-05-14
Skip jar sha check on windows for now

sophia-guo:
Skip jar sha check on windows for now. The jar sha generated on windows
using command is different from the ones other platforms. If we do need
support it we can add all corresponding SHAs later.

[ci skip]


smlambert:
verified in winNewSh branch

PR #1911, opened 2018-05-14
Remove libffi workaround

gacholio:
Now that the build scripts no longer refer to the shared library,
stoping faking it up.

Follow up from #1900


pshipton:
jenkins compile all jdk8

PR #1913, opened 2018-05-14
Add implementation of String.repeat(count)

### LINKED ISSUE #1660: jdk11 add String.repeat(count)
http://cr.openjdk.java.net/~iris/se/11/build/latest/diffsFrom10%2b45/java/lang/String.html

Ali-Ni:
Add implementation of java.lang.String.repeat(count).
This is only for Java 11.

Closes #1660
    
### DIFF
--- a/jcl/src/java.base/share/classes/java/lang/String.java
+++ b/jcl/src/java.base/share/classes/java/lang/String.java
@@ -3847,7 +3847,49 @@ public final class String implements Serializable, Comparable<String>, CharSeque
 		}
 		return string;
 	}
-	/*[ENDIF] Java11*/
+
+	/**
+	 * Returns a string whose value is the concatenation of this string repeated
+	 * count times. 
+	 * 
+	 * @param count
+	 *            a positive integer indicating the number of times to be repeated
+	 * @return a string whose value is the concatenation of this string repeated count times
+	 * @throws IllegalArgumentException
+	 *             if the count is negative
+	 * @since 11
+	 */
+	public String repeat(int count) {
+		if (count < 0) {
+			throw new IllegalArgumentException();
+		} else if (count == 0 || isEmpty()) {
+			return "";
+		} else if (count == 1) {
+			return this;
+		}
+
+		int length = lengthInternal();
+		int repeatlen = length * count;
+
+		if (enableCompression && (null == compressionFlag || coder == LATIN1)) {
+			byte[] buffer = new byte[repeatlen];
+
+			for (int i = 0; i < count; i++) {
+				compressedArrayCopy(value, 0, buffer, i * length, length);				
+			}
+
+			return new String(buffer, LATIN1);
+		} else {
+			byte[] buffer = new byte[repeatlen * 2];
+
+			for (int i = 0; i < count; i++) {
+				decompressedArrayCopy(value, 0, buffer, i * length, length);				
+			}
+				
+			return new String(buffer, UTF16);
+		}
+	}
+	/*[ENDIF] Java11 */
 
 /*[ELSE] Sidecar19-SE*/
 	// DO NOT CHANGE OR MOVE THIS LINE
--- a/test/functional/Java11andUp/playlist.xml
+++ b/test/functional/Java11andUp/playlist.xml
@@ -24,32 +24,9 @@
 
 <playlist xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:noNamespaceSchemaLocation="../../TestConfig/playlist.xsd">
 	<test>
-		<testCaseName>StringvalueOfCodePoint_JDK11_CompactStrings_Enabled</testCaseName>
+		<testCaseName>String_CompactStrings</testCaseName>
 		<variations>
 			<variation>-XX:+CompactStrings</variation>
-		</variations>
-		<command>$(JAVA_COMMAND) $(JVM_OPTIONS) \
-			-cp $(Q)$(RESOURCES_DIR)$(P)$(TESTNG)$(P)$(TEST_RESROOT)$(D)GeneralTest.jar$(Q) \
-			org.testng.TestNG -d $(REPORTDIR) $(Q)$(TEST_RESROOT)$(D)testng.xml$(Q) -testnames Test_String \
-			-groups $(TEST_GROUP) \
-			-excludegroups $(DEFAULT_EXCLUDE); \
-			$(TEST_STATUS)
-		</command>
-		<levels>
-			<level>sanity</level>
-		</levels>
-		<groups>
-			<group>functional</group>
-		</groups>
-		<subsets>
-			<subset>SE110</subset>
-		</subsets>
-	</test>
-
-	<test>
-		<testCaseName>StringvalueOfCodePoint_JDK11_CompactStrings_Disabled</testCaseName>
-		<variations>
-			<variation>-XX:-CompactStrings</variation>
 		</variations>
 		<command>$(JAVA_COMMAND) $(JVM_OPTIONS) \
 			-cp $(Q)$(RESOURCES_DIR)$(P)$(TESTNG)$(P)$(TEST_RESROOT)$(D)GeneralTest.jar$(Q) \
--- a/test/functional/Java11andUp/src/org/openj9/test/java_lang/Test_String.java
+++ b/test/functional/Java11andUp/src/org/openj9/test/java_lang/Test_String.java
@@ -63,4 +63,41 @@ public class Test_String {
 			// Expected exception
 		}
 	}
+
+	/*
+	 * Test Java 11 method String.repeat(count)
+	 */
+	@Test(groups = { "level.sanity" })
+	public void testRepeat() {
+		String empty = "";
+		String latin1 = "abc123";
+		String nonLatin1 = "abc\u0153";
+		String str;
+
+		// Verify that correct strings are produced under normal conditions
+		Assert.assertTrue(latin1.repeat(3).equals("abc123abc123abc123"), "Repeat failed to produce correct string (LATIN1)");
+		Assert.assertTrue(nonLatin1.repeat(3).equals("abc\u0153abc\u0153abc\u0153"), "Repeat failed to produce correct string (non-LATIN1)");
+
+		Assert.assertTrue(empty.repeat(0) == empty, "Failed to return identical empty string - empty base (0)");
+		Assert.assertTrue(empty.repeat(2) == empty, "Failed to return identical empty string - empty base (2)");
+		Assert.assertTrue(latin1.repeat(0) == empty, "Failed to return identical empty string - nonempty base (0)");
+
+		// Verify that IllegalArgumentException is thrown for count < 0
+		try {
+			str = empty.repeat(-1);
+			Assert.fail("Failed to detect invalid count -1");
+		} catch (IllegalArgumentException e) {
+			// Expected exception
+		}
+
+		try {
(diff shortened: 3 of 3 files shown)


Ali-Ni:
This depends on #1786. fyi @pshipton. Tests will be added in a new pull request once the #1786 changes are merged.

jcl/src/java.base/share/classes/java/lang/String.java
```diff
@@ -857,47 +667,3874 @@ public String(char[] data) {
...
+	/**
+	 * Converts the specified object to its string representation. If the object is null answer the string {@code "null"}, otherwise use
+	 * {@code toString()} to get the string representation.
+	 *
+	 * @param value
+	 *			  the object
+	 * @return the object converted to a string
+	 */
+	public static String valueOf(Object value) {
+		return value != null ? value.toString() : "null"
```
pshipton: To include only for Java 11, put `/*[IF Java11]*/` before the method, and `/*[ENDIF]*/` afterwards. You can find other examples of this in this file. These are java preprocessor commands, this source is preprocessed to the correct level before being used for a Java version.

pshipton: Ha, I see its already included in a Java 11 block. Missed that at first.

pshipton:
Ideally the tests should be added in this same PR so they can be run as part of the testing done before this change is merged.

jcl/src/java.base/share/classes/java/lang/String.java
pshipton: Please create an appropriate description.

babsingh:
fyi - @tajila

tajila:
@Ali-Ni can you please rebase this PR, its hard to see what are your changes and what are other peoples commits

Ali-Ni:
Should I squash the implementation and tests into one commit or leave as is?

jcl/src/java.base/share/classes/java/lang/String.java
```diff
@@ -3827,7 +3827,51 @@ static String valueOfCodePoint(int codePoint) {
 		}
 		return string;
 	}
-	/*[ENDIF] Java11*/
+
+
```
fjeremic: Unnecessary double new line can be removed (similarly at the end of the function).

jcl/src/java.base/share/classes/java/lang/String.java
```diff
@@ -3827,7 +3827,51 @@ static String valueOfCodePoint(int codePoint) {
...
+			throw new IllegalArgumentException();
+		}
+		if (count == 1) {
+			return this;
+		}
+
+		int length = this.lengthInternal();
+		int repeatlen = length * count;
+
+		if (enableCompression && (null == compressionFlag || this.coder == LATIN1)) {
```
fjeremic: For consistency with all the other code in String.java I suggest we avoid the `this.` prefix for fields unless there is ambiguity.

jcl/src/java.base/share/classes/java/lang/String.java
```diff
@@ -3827,7 +3827,51 @@ static String valueOfCodePoint(int codePoint) {
...
+	 * 
+	 * @param count
+	 *            a positive integer indicating the number of times to be repeated
+	 * @return a string whose value is the concatenation of this string repeated count times
+	 * @throws IllegalArgumentException
+	 *             if the count is negative
+	 * @since 11
+	 */
+	public String repeat(int count) {
+		if (count < 0) {
```
tajila: what happens if the count is zero

test/functional/Java11andUp/src/org/openj9/test/java_lang/Test_String.java
```diff
@@ -63,4 +63,48 @@ public void testValueOfCodePoint() throws IllegalAccessException, IllegalArgumen
...
+			str = n.repeat(3);
+			Assert.fail("Repeat failed to throw NP exception");
+		} catch (NullPointerException e) {
+			// Expected exception
+		}
+
+		// Verify that a reference to the instance that called the function is returned for count == 1
+		Assert.assertTrue(populated.repeat(1) == populated, "Should be identical on count 1");
+	}
 }
```
pshipton: Please add a new line after }

jcl/src/java.base/share/classes/java/lang/String.java
Ali-Ni: If the count is zero then a byte array of length 0 is passed into the String constructor and an empty String is returned as desired. Should this be done more explicitly?

tajila: Ya, I think it would be beneficial to provide a faster path for `count ==0` as well as `this.equals("")` just like you have done for `count==1`

jcl/src/java.base/share/classes/java/lang/String.java
```diff
@@ -3827,7 +3827,49 @@ static String valueOfCodePoint(int codePoint) {
...
+		} else if (count == 0 || isEmpty()) {
+			return new String();
+		} else if (count == 1) {
+			return this;
+		}
+
+		int length = lengthInternal();
+		int repeatlen = length * count;
+
+		if (enableCompression && (null == compressionFlag || this.coder == LATIN1)) {
```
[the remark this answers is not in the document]
fjeremic: Yes please: https://github.com/eclipse/openj9/pull/1913#discussion_r189081984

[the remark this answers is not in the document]
Ali-Ni: [Done](https://github.com/eclipse/openj9/pull/1913/files#diff-6c803ce7e383eaf8ebe1e6771f5e5fe9R3854)

tajila:
@pshipton please review

jcl/src/java.base/share/classes/java/lang/String.java
```diff
@@ -3847,7 +3847,49 @@ static String valueOfCodePoint(int codePoint) {
...
+	 * @return a string whose value is the concatenation of this string repeated count times
+	 * @throws IllegalArgumentException
+	 *             if the count is negative
+	 * @since 11
+	 */
+	public String repeat(int count) {
+		if (count < 0) {
+			throw new IllegalArgumentException();
+		} else if (count == 0 || isEmpty()) {
+			return new String();
```
pshipton: Should this be a new String every time, or return "", which is the identical empty string every time.

test/functional/Java11andUp/src/org/openj9/test/java_lang/Test_String.java
```diff
@@ -63,4 +63,48 @@ public void testValueOfCodePoint() throws IllegalAccessException, IllegalArgumen
...
+		try {
+			str = populated.repeat(-3);
+			Assert.fail("Failed to detect invalid count -3");
+		} catch (IllegalArgumentException e) {
+			// Expected exception
+		}
+		
+		// Verify that NP exception is thrown as expected
+		try {
+			str = n.repeat(3);
```
pshipton: Its not harmful but this test isn't testing repeat().

test/functional/Java11andUp/src/org/openj9/test/java_lang/Test_String.java
```diff
@@ -63,4 +63,48 @@ public void testValueOfCodePoint() throws IllegalAccessException, IllegalArgumen
...
+			Assert.fail("Repeat failed to throw NP exception");
+		} catch (NullPointerException e) {
+			// Expected exception
+		}
+
+		// Verify that a reference to the instance that called the function is returned for count == 1
+		Assert.assertTrue(populated.repeat(1) == populated, "Should be identical on count 1");
```
pshipton: Please add a test for a String containing non-LATIN1 character(s).

jcl/src/java.base/share/classes/java/lang/String.java
pshipton: After determining the correct answer, add the case to the test suite.

pshipton:
@llxia any concerns with the tests?

test/functional/Java11andUp/src/org/openj9/test/java_lang/Test_String.java
```diff
@@ -63,4 +63,41 @@ public void testValueOfCodePoint() throws IllegalAccessException, IllegalArgumen
...
 		}
 	}
-}
\ No newline at end of file
+
+	/*
+	 * Test Java 11 method String.repeat(count)
+	 */
+	@Test(groups = { "level.sanity" })
+	public void testRepeat() throws IllegalAccessException, IllegalArgumentException, InvocationTargetException, NoSuchMethodException, SecurityException {
```
pshipton: Seems like this was copied from somewhere, not sure any of these exceptions need to be declared for this test.

test/functional/Java11andUp/playlist.xml
```diff
@@ -47,7 +47,7 @@
 	</test>
 
 	<test>
-		<testCaseName>StringvalueOfCodePoint_JDK11_CompactStrings_Disabled</testCaseName>
+		<testCaseName>String_JDK11_CompactStrings_Disabled</testCaseName>
```
llxia: There is no need to separate the test into two `<test>` if the only difference is the variation. We can use one `<test>` and two `<variation>`.

```
<test>
   <testCaseName>String_CompactStrings</testCaseName>
   <variations>
      <variation>-XX:+CompactStrings</variation>
      <variation>-XX:-CompactStrings</variation>
   </variations>
   ...
```
Also, please do not have JDK number in the test name, in case we need to run this test in JDK11+ in the future.

llxia:
Copyright check and Line Ending check failed due to git fetch issue. You can just push the change again to re-trigger the builds.

test/functional/Java11andUp/playlist.xml
```diff
@@ -24,31 +24,9 @@
 
 <playlist xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:noNamespaceSchemaLocation="../../TestConfig/playlist.xsd">
 	<test>
-		<testCaseName>StringvalueOfCodePoint_JDK11_CompactStrings_Enabled</testCaseName>
+		<testCaseName>String_CompactStrings</testCaseName>
```
pshipton: This test name is no longer appropriate since the test includes both compact and non-compact. I know Lan didn't like JDK11 in the name, but perhaps String_Java11andUp is appropriate. @llxia

fjeremic: I agree. IMO we should run the entire JCL_TEST_Java-Lang with compact strings enabled explicitly and disabled explicitly. Or at the very least run all the String, StringBuffer, StringBuilder, Integer, Long, and StringCoding tests in both variations.

pshipton: Makes sense, although if we run the entire JCL_TEST_Java-Lang a second time that may significantly increase the sanity build time.

pshipton: @Ali-Ni the summary of all this is that if you remove `<variation>-XX:-CompactStrings</variation>` this will be ready to merge.

Ali-Ni: Done

pshipton: > IMO we should run the entire JCL_TEST_Java-Lang with compact strings enabled explicitly and disabled explicitly

Created #2029

pshipton:
We don't have PR testing for Java 11 yet. The travis build succeeded, which seems sufficient.

Issue #1849, opened 2018-05-04
Different results presented by Class.getResourceAsStream method


felipepontes-research:
Class.getResourceAsStream method presents different results as presented by Oracle HotSpot JVM when executing the following program:

```
package p;
public enum A {
    X, Y
}

package p;
public class B {
    public static void main(String[] args) {
        System.out.println(A[].class.getResourceAsStream(""));
    }
}
```

**Affected versions:**

Eclipse OpenJ9

openjdk version "1.8.0_162"
OpenJDK Runtime Environment (build 1.8.0_162-b12)
Eclipse OpenJ9 VM (build openj9-0.8.0, JRE 1.8.0 Linux amd64-64 Compressed References 20180315_120 (JIT enabled, AOT enabled)
OpenJ9 - e24e8aa9
OMR - 3e8296b4
JCL - ee1e77df1d based on jdk8u162-b12)

Oracle HotSpot

Picked up _JAVA_OPTIONS:   -Dawt.useSystemAAFontSettings=gasp
java version "1.8.0_151"
Java(TM) SE Runtime Environment (build 1.8.0_151-b12)
Java HotSpot(TM) 64-Bit Server VM (build 25.151-b12, mixed mode)

**Steps to reproduce:**

	1. Compile above program
	2. Run it using Eclipse OpenJ9 java
	3. Run it using Oracle HotSpot java

**Current result:**

Eclipse OpenJ9: null
Oracle HotSpot: java.io.ByteArrayInputStream@...

**Expected result:**

Eclipse OpenJ9 JVM presenting same resources as Oracle JVM.

DanHeidinga:
We'll need to check the spec on this one.  Its not immediately clear what the right behaviour is

DanHeidinga:
@theresa-m Can you take a look at this?

theresa-m:
Confirmed this issue was fixed with #1909. I've added an additional test for this case so its covered under `getResourceAsStream` as well as `getResource` #1916

PR #1916, opened 2018-05-15
Additional test for Class.getResourceAsStream

theresa-m:
Test with object as an array

Tests: #1849


pshipton:
jenkins test sanity plinux jdk8,jdk10

PR #1920, opened 2018-05-16
Add access checks for MethodHandle.invoke parameter and return types

JasonFengJ9:
Add access checks for `MethodHandle.invoke(Exact)` parameter and return types

Though methods `j.l.i.MethodHandle.invoke(Exact)` are public and there is no need for access check, the parameter and return classes require access check by access class.
Throw `IllegalAccessError` instead of `IllegalAccessException` when an application attempts (not refectively) to access or modify a field, or to invoke a method that it doesn't have access to. This is to satisfy Java spec requirements.

Reviewer: @DanHeidinga 

### DIFF
--- a/jcl/src/java.base/share/classes/java/lang/invoke/MethodHandles.java
+++ b/jcl/src/java.base/share/classes/java/lang/invoke/MethodHandles.java
@@ -779,13 +779,9 @@ public class MethodHandles {
 		}
 
 		/*
-		 * Check for methods MethodHandle.invokeExact, MethodHandle.invoke,
-		 * or MethodHandles.varHandleInvoker.  
-		 * Access checks to these methods are not required as these methods are public and therefore
-		 * accessible to everyone.
-		 * However the access checks to the parameter and return classes are required.
+		 * Check access to the parameter and return classes within incoming MethodType
 		 */
-		MethodHandle handleForMHInvokeMethods(Class<?> clazz, String methodName, MethodType type) throws IllegalAccessException {
+		private void accessCheckArgRetTypes(MethodType type) throws IllegalAccessException {
 			if (INTERNAL_PRIVILEGED != accessMode) {
 				for (Class<?> para : type.arguments) {
 					if (!para.isPrimitive()) {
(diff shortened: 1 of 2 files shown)


jcl/src/java.base/share/classes/java/lang/invoke/MethodHandles.java
```diff
@@ -781,10 +781,22 @@ static MethodHandle adaptInterfaceLookupsOfObjectMethodsIfRequired(MethodHandle
...
 		 * Check for methods MethodHandle.invokeExact, MethodHandle.invoke,
 		 * or MethodHandles.varHandleInvoker.  
-		 * Access checks are not required as these methods are public and therefore
+		 * Access checks to these methods are not required as these methods are public and therefore
 		 * accessible to everyone.
+		 * However the access checks to the parameter and return classes are required.
 		 */
-		static MethodHandle handleForMHInvokeMethods(Class<?> clazz, String methodName, MethodType type) {
+		MethodHandle handleForMHInvokeMethods(Class<?> clazz, String methodName, MethodType type) throws IllegalAccessException {
+			if (INTERNAL_PRIVILEGED != accessMode) {
```
DanHeidinga: We're likely doing double access checks as this check is being done even if that the method being looked up isn't MH.invoke/invokeExact, or VarHandle.

Can this be modified to only do the check if the MH is for invoke/invokeExact or VarHandle invoker?

JasonFengJ9: Added a helper method to perform `parameter/return` class access check for `invoke/invokeExact` and `VarHandle invoker`.
Verified manually that `pConfig` still compiles.
@DanHeidinga please have another look.

jcl/src/java.base/share/classes/java/lang/invoke/MethodHandles.java
```diff
@@ -809,6 +817,7 @@ MethodHandle handleForMHInvokeMethods(Class<?> clazz, String methodName, MethodT
 			 * the resulting method handle should be equivalent to one produced by varHandleInvoker().
 			 */
 			if (VarHandle.class.isAssignableFrom(clazz)) {
+				accessCheckArgRetTypes(type);
 				for (AccessMode m : AccessMode.values()) {
 					if (m.methodName().equals(methodName)) {
```
DanHeidinga: Check should be here - in the success case - not at the top of the if scope

jcl/src/java.base/share/classes/java/lang/invoke/MethodHandles.java
```diff
@@ -778,17 +778,37 @@ static MethodHandle adaptInterfaceLookupsOfObjectMethodsIfRequired(MethodHandle
 			return handle;
 		}
 
+		/*
+		 * Check access to the parameter and return classes within incoming MethodType
+		 */
+		void accessCheckArgRetTypes(MethodType type) throws IllegalAccessException {
```
DanHeidinga: Can this be `private` as well?

jcl/src/java.base/share/classes/java/lang/invoke/MethodHandles.java
JasonFengJ9: Yeah, that's right. Fixed, thanks Dan.

jcl/src/java.base/share/classes/java/lang/invoke/MethodHandles.java
JasonFengJ9: Sure thing, updated.

DanHeidinga:

PR #1921, opened 2018-05-16
Update Linux makefile to allow compilation with GCC 7.3

babsingh:
GCC versions greater than 5 default to GNU11 but OpenJ9 uses
GNU89 inline semantics. -fgnu89-inline option is appended to CFLAGS
to allow compilation with GCC versions greater than 5 and GNU89 inline
semantics. Reference - https://gcc.gnu.org/gcc-5/porting_to.html.

JDK11 Linux x86 DockerFile no longer needs to initialize CFLAGS with
-fgnu89-inline since the Linux makefile now accounts for -fgnu89-inline
option if the GCC version is greater than 5.

### DIFF
--- a/runtime/makelib/targets.mk.linux.inc.ftl
+++ b/runtime/makelib/targets.mk.linux.inc.ftl
@@ -1,5 +1,5 @@
 <#-- 
-	Copyright (c) 1998, 2017 IBM Corp. and others
+	Copyright (c) 1998, 2018 IBM Corp. and others
 	
 	This program and the accompanying materials are made available under
 	the terms of the Eclipse Public License 2.0 which accompanies this
@@ -152,6 +152,16 @@ ifdef USE_PPC_GCC
   PPC_GCC_CXXFLAGS+=-O3 -fno-strict-aliasing
 endif
 </#if>
+
+<#-- GCC versions greater than 5 default to GNU11 but OpenJ9 uses
+GNU89 inline semantics. -fgnu89-inline option is appended to CFLAGS
+to allow compilation with GCC versions greater than 5 and GNU89 inline
+semantics. Reference - https://gcc.gnu.org/gcc-5/porting_to.html.
+-->
+GCCVERSIONGTEQ5 := $(shell expr `gcc -dumpversion | cut -f1 -d.` \>= 5)
+ifeq "$(GCCVERSIONGTEQ5)" "1"
+    CFLAGS+=-fgnu89-inline
+endif
 
 <#if !uma.spec.processor.ppc>
 CXXFLAGS+=-fno-exceptions -fno-rtti -fno-threadsafe-statics
--- a/buildenv/docker/jdk11/x86_64/ubuntu16/Dockerfile
+++ b/buildenv/docker/jdk11/x86_64/ubuntu16/Dockerfile
@@ -93,7 +93,4 @@ RUN cd /root \
 ENV JAVA_HOME="/root/bootjdk10"
 ENV PATH="${JAVA_HOME}/bin:${PATH}"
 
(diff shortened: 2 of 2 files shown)


babsingh:
Compilation passes on xLinux, pLinux and zLinux with gcc-4.8. Compilation passes on xLinux with gcc-7.3. Compile failures related to OpenJDK source code seen when compiling on pLinux with gcc-7.3. No zLinux machine available to compile with gcc-7.3.

babsingh:
This pull request solves the `error: inline function $INLINED_FUNCTION_NAME declared but never defined` when using GCC 7.3. Reference: https://github.com/eclipse/openj9/issues/1684. This error was first seen in multiple places within `trj9/runtime/MethodMetaData.hMethodMetaData.h` (**JIT source**). GNU89 inline semantics are used within OpenJ9 but GCC 5+ defaults to GNU11 semantics. To allow compilation with GCC 5+ and GNU89 inline semantics, `-fgnu89-inline` is appended to CFLAGS.

For using this approach, I would like to request approval from JIT members: @mpirvu, @fjeremic, @andrewcraik since JIT source code is impacted by the compile error.

babsingh:
fyi - @pshipton

runtime/makelib/targets.mk.linux.inc.ftl
```diff
@@ -153,6 +153,16 @@ ifdef USE_PPC_GCC
 endif
 </#if>
 
+<#-- GCC versions greater than 5 default to GNU11 but OpenJ9 uses
+GNU89 inline semantics. -fgnu89-inline option is appended to CFLAGS
+to allow compilation with GCC versions greater than 5 and GNU89 inline
+semantics. Reference - https://gcc.gnu.org/gcc-5/porting_to.html.
+-->
+GCCVERSIONGTEQ5 := $(shell expr `gcc -dumpversion | cut -f1 -d.` \>= 5)
```
DanHeidinga: What happens when gcc 10 ships?  (Not a blocker for merging this now but something to think about while we're looking at this or we'll have to rediscover it later)

babsingh: 1. `gcc -dumpversion | cut -f1 -d.` will evaluate to `10`
2. `expr 10 \>= 5` will evaluate to `1`
3. `GCCVERSIONGTEQ5` = `1`
4. `if GCCVERSIONGTEQ5 == 1` then `CFLAGS+=-fgnu89-inline`

DanHeidinga: I misread the `cut` command.

DanHeidinga:
Jenkins test sanity xlinux

runtime/makelib/targets.mk.linux.inc.ftl
keithc-ca: We should not be invoking `gcc`, but instead checking if we're using the gcc toolchain and only then invoking `$(CC) -dumpversion`. A system has at most one tool called 'gcc', but the user may have configured their build to use a different version; the test should be based on the selected compiler.

keithc-ca: I'll put together a PR to address my previous comment.

DanHeidinga: Good catch.  Thanks @keithc-ca

babsingh: Thanks @keithc-ca

keithc-ca: https://github.com/eclipse/openj9/pull/2205 proposes a fix for this.

PR #1922, opened 2018-05-16
Add new OMR platform define for Windows

youngar:


runtime/makelib/targets.mk.win.inc.ftl
```diff
@@ -135,22 +134,35 @@ UMA_EXEFLAGS+=/INCREMENTAL:NO /NOLOGO /LARGEADDRESSAWARE
...
 <#if uma.spec.processor.amd64>
-  <#if uma.spec.flags.build_newCompiler.enabled>
-    CFLAGS+=-D_AMD64_=1 -DWIN64 -D_WIN64 -DWIN32 -D_WIN32 -DJ9HAMMER
-    CXXFLAGS+=-D_AMD64_=1 -DWIN64 -D_WIN64 -DWIN32 -D_WIN32 -DJ9HAMMER
-  <#else>
-    CFLAGS+=-D_AMD64_=1 -DWIN64 -D_WIN64 -DWIN32 -D_WIN32 -DJ9HAMMER /FIPRE64PRA.H -Wp64
-    CXXFLAGS+=-D_AMD64_=1 -DWIN64 -D_WIN64 -DWIN32 -D_WIN32 -DJ9HAMMER /FIPRE64PRA.H -Wp64
+  CFLAGS+=-D_AMD64_=1 -DWIN64 -D_WIN64 -DJ9HAMMER
+  CXXFLAGS+=-D_AMD64_=1 -DWIN64 -D_WIN64 -DJ9HAMMER
+  <#if !uma.spec.flags.build_newCompiler.enabled>
```
pshipton: You can just delete the build_newCompiler check and options, its obsolete and no longer supported.

youngar: It looks like it is enabled in our specs: 
```xml
buildspecs/j9.flags:159:        <flag id="build_newCompiler">
buildspecs/win_x86-64.spec:141:         <flag id="build_newCompiler" value="true"/>
buildspecs/win_x86-64_cmprssptrs.spec:137:              <flag id="build_newCompiler" value="true"/>
buildspecs/win_x86.spec:140:            <flag id="build_newCompiler" value="true"/>
```
I will delete this flag, but should I delete the compile options?

pshipton: Right sorry. We need the options, but we don't need to check the flag which is always true and can be removed.

youngar: Right, makes sense. Thanks for clarifying!

youngar: Woops I misread the code last night.  We *did* want to delete the compile options.  Fixed.

youngar:
Passed an IBM internal Java build on xa, wi, and wa platforms.

PR #1925, opened 2018-05-17
Clean up attach API code

pdbain-ibm:
Fix visibility of methods to avoid synthetic methods.  Remove autoboxing. Add
try-with-resources to close files.  Add javadoc for public classes and methods.
Remove spurious object creation.  Remove all warning suppression.

### DIFF
--- a/jcl/src/jdk.attach/share/classes/com/ibm/tools/attach/attacher/OpenJ9VirtualMachine.java
+++ b/jcl/src/jdk.attach/share/classes/com/ibm/tools/attach/attacher/OpenJ9VirtualMachine.java
@@ -2,7 +2,7 @@
 package com.ibm.tools.attach.attacher;
 
 /*******************************************************************************
- * Copyright (c) 2009, 2017 IBM Corp. and others
+ * Copyright (c) 2009, 2018 IBM Corp. and others
  *
  * This program and the accompanying materials are made available under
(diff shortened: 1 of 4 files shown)


jcl/src/jdk.attach/share/classes/com/ibm/tools/attach/attacher/OpenJ9VirtualMachine.java
```diff
@@ -353,15 +350,13 @@ private static boolean parseResponse(String response) throws IOException,
...
-	@SuppressWarnings("boxing")
 	private static Integer getStatusValue(String response) {
 		Pattern rvPattern = Pattern.compile("(-?\\d+)\\s*$");  //$NON-NLS-1$
 		Matcher rvMatcher = rvPattern.matcher(response);
 		if (rvMatcher.find()) {
 			String status = rvMatcher.group(1);
 			try {
-				int statusValue = Integer.parseInt(status);
-				return statusValue;
+				return Integer.valueOf(Integer.parseInt(status));
```
pshipton: You can use Integer.getInteger(String)

pdbain-ibm: Right! Thanks.  Fixed.

pshipton:
jenkins test all plinux jdk8,jdk10

PR #1927, opened 2018-05-17
Enable native jvmtitests

### LINKED ISSUE #1918: Enable native tests jvmtitest
#1901 Move jvmtitests native code to runtime/tests to build automatically

As native code is built jvmtitest can be enabled. Some other native tests can also be enabled.

sophia-guo:
[ci skip]

close #1918 


llxia:

llxia:
failure is due to https://github.com/eclipse/openj9/issues/1787
relaunch PR build

llxia:

test/functional/cmdLineTests/jvmtitests/build.xml
```diff
@@ -63,19 +64,27 @@
...
 					<else>
-						<property name="addExports" value="--add-exports java.base/jdk.internal.module=ALL-UNNAMED --add-exports java.base/com.ibm.tools.attach.target=ALL-UNNAMED --add-exports java.base/com.ibm.tools.attach=ALL-UNNAMED" />
+						<javac srcdir="${src}" destdir="${build}" debug="true" fork="true" executable="${compiler.javac}" includeAntRuntime="false" encoding="ISO-8859-1">
+							<src path="${src}"/>
+							<src path="${src_current}"/>
+							<compilerarg line="--add-exports java.base/jdk.internal.module=ALL-UNNAMED --add-exports java.base/com.ibm.tools.attach.target=ALL-UNNAMED --add-exports java.base/com.ibm.tools.attach=ALL-UNNAMED" />
+							<classpath>
+								<pathelement location="${asm.jar}" />
+							</classpath>
+						</javac>
```
llxia: Is it possible to use variable to reduce code duplication?

sophia-guo: Yes,  I can add variable  'addExports' and update as <compilerarg line='${addExports}' />. Though it will not reduce  duplicate lines as those are  attributes of javac task.

sophia-guo:
I could do like following, but I don't see much difference.
```				<property name="addExports" value="--add-exports java.base/jdk.internal.module=ALL-UNNAMED --add-exports java.base/com.ibm.tools.attach.target=ALL-UNNAMED --add-exports java.base/com.ibm.tools.attach=ALL-UNNAMED" />
				<if>
					<equals arg1="${JCL_VERSION}" arg2="latest"/>
					<then>
						<property name="extraSrc" value="${src_latest}" />
						<property name="excludeFile" value="**/modularityTests/**" />
					</then>
					<else>
						<property name="extraSrc" value="${src_current}" />
						<property name="excludeFile" value="" />
					</else>
				</if>
				<javac srcdir="${src}" destdir="${build}" debug="true" fork="true" executable="${compiler.javac}" includeAntRuntime="false" encoding="ISO-8859-1">
					<src path="${src}"/>
					<src path="${extraSrc}"/>
					<exclude name="${excludeFile}" />
					<compilerarg line="${addExports}" />
					<classpath>
						<pathelement location="${asm.jar}" />
					</classpath>
				</javac>
```

llxia:
The point to reduce code duplication is to make code easy to maintain. For example, if we need to change the value for `add-exports` or add classpath, we do not need to change two places.

sophia-guo:
yes, understood. According to our status, if 'addExports' need to change most of time are it's different for latest and current, which means we do need two copies.

llxia:
`addExports` value is the same for latest and current right now. We can deal with it if it changes in the future.

sophia-guo:
updated.

llxia:

Issue #1764, opened 2018-04-25
Missed check for GS support in Port Library for zLinux


dmitripivkine:
We might miss check for availability of Guarded Storage support in Port Library for zLinux. This might be a reason for crash if OS does not support GS. Could we double check?

@ebadyano @fjeremic Is it a real issue? Should it be addressed before we release CS for zLinux?

dmitripivkine:
@joransiu FYI

fjeremic:
> Is it a real issue?

It does certainly seem to be an issue. I cannot locate the code which checks the `auxv` bits on Linux on z. I was certain we had this, but all I can find so far is the z/OS CVT bit checking.

joransiu:
> It does certainly seem to be an issue. I cannot locate the code which checks the auxv bits on Linux on z. I was certain we had this, but all I can find so far is the z/OS CVT bit checking.

Yeah, I'm utterly confused here too.  Other features like Vector Facility would have OS dependencies as well, and I did not find queries for auxv and/or cpu features in `/proc/cpuinfo`.

fjeremic:
It seems we do check `/proc/cpuinfo` for the vector support from here:

https://github.com/eclipse/openj9/blob/78560086e6911b7d0998aeb8118018a8afc22772/runtime/compiler/trj9/z/env/J9CPU.cpp#L401-L406

which calls:

https://github.com/eclipse/openj9/blob/78560086e6911b7d0998aeb8118018a8afc22772/runtime/compiler/trj9/z/env/J9CPU.cpp#L401-L406

We don't seem to have anything like this for Guarded Storage. In fact we probably shouldn't. We need to eliminate this kind of dodgy string checking code and just rely on the result of `auxv`, probably even before we check STFLE bits.

SueChaplain:
Do we need to add anything to the 0.9.0 release notes for this please? What is the status w.r.t support on RHEL7.5 and Ubuntu 18.04.

ebadyano:
https://www-03preprod.ibm.com/support/knowledgecenter/SSYKE2_8.0.0_test/com.ibm.java.vm.80.doc/docs/xgc.html RHEL 7.5 supports Guarded Storage, not sure about Ubuntu @joransiu Joran, do you know if it's supported on Ubuntu?

If -Xgc:concurrentScavenge is used on hardware that supports guarded storage with OS that doesn't, the symptom will be **stderr Fatal error: Failed to initialize thread; please disable Concurrent Scavenge.**

[...]

PR #1929, opened 2018-05-17
Adding missed OS check for GS support in Port Library for zLinux

ebadyano:
Use aux vector to check Guarded Storage support for Linux OS on s390
hardware.

Fixes: #1764

Signed-off-by: Evgenia Badyanova <evgeniab@ca.ibm.com>

### DIFF
--- a/runtime/oti/j9port.h
+++ b/runtime/oti/j9port.h
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 1991, 2017 IBM Corp. and others
+ * Copyright (c) 1991, 2018 IBM Corp. and others
  *
  * This program and the accompanying materials are made available under
  * the terms of the Eclipse Public License 2.0 which accompanies this
@@ -507,6 +507,11 @@ typedef struct J9ProcessorDesc {
 
 /* STFLE bit 57 - Message-security-assist-extension-5 facility */
 #define J9PORT_S390_FEATURE_MSA_EXTENSION_5 57
+
+/* zLinux features
+ * Auxiliary Vector Hardware Capability (AT_HWCAP) features for Linux on z.
+ */
+#define J9PORT_HWCAP_S390_GS 0x4000
 
 /* x86 features
  * INTEL INSTRUCTION SET REFERENCE, A-M
--- a/runtime/port/unix/j9sysinfo.c
+++ b/runtime/port/unix/j9sysinfo.c
@@ -66,10 +66,10 @@
 #include "j9csrsi.h"
 #endif /* defined(J9ZOS390) */
 
-#if defined(LINUXPPC)
+#if (defined(LINUXPPC) || (defined(S390) && defined(LINUX)))
 #include "auxv.h"
 #include <strings.h>
-#endif /* defined(LINUXPPC) */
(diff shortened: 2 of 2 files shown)


ebadyano:
@fjeremic I couldn't find Daniel Hong to add him here..

ebadyano:
Personal build link: http://vmfarm.ottawa.ibm.com/build_info.php?build_id=386810

ebadyano:
depends on https://github.com/eclipse/omr/pull/2556

ebadyano:
fix for: https://github.com/eclipse/openj9/issues/1764

runtime/oti/j9port.h
```diff
@@ -508,6 +508,11 @@ typedef struct J9ProcessorDesc {
 /* STFLE bit 57 - Message-security-assist-extension-5 facility */
 #define J9PORT_S390_FEATURE_MSA_EXTENSION_5 57
 
+/* zLinux features
+ * Auxiliary Vector Hardware Capability (AT_HWCAP) features for zLinux.
+ */
+#define J9PORT_HWCAP_S390_GS 16384
```
fjeremic: Could we use a hex 0x4000 for this value instead? It is a lot more clear IMO. Also "zLinux" is a deprecated term. We should try to use "Linux on Z" instead.

runtime/port/unix/j9sysinfo.c
```diff
@@ -66,10 +66,10 @@
 #include "j9csrsi.h"
 #endif /* defined(J9ZOS390) */
 
-#if defined(LINUXPPC)
+#if (defined(LINUXPPC) || (defined(S390) && !defined(J9ZOS390)))
```
fjeremic: Are we able to turn this negative statement into a positive one using `defined(LINUX)`? i.e.

```
if (defined(LINUXPPC) || (defined(S390) && defined(LINUX)))
```

dmitripivkine: what do we have set for ZTPF? It is "LINUX" for sure... Does it have "S390" set as well?

fjeremic: @pelemie could you confirm what we have to ZTPF?

ebadyano: how would auxv work on ZTPF? Could it be the right thing to do to run auxv the check?

ebadyano: I changed it to if (defined(LINUXPPC) || (defined(S390) && defined(LINUX))). If it is true for ZTPF would that be an issue or something we actually want to be true for ZTPF?

runtime/oti/j9port.h
ebadyano: fixed the comment and changed the value to hex.

runtime/port/unix/j9sysinfo.c
fjeremic: > I changed it to if (defined(LINUXPPC) || (defined(S390) && defined(LINUX))). If it is true for ZTPF would that be an issue or something we actually want to be true for ZTPF?

I'm not really sure. ZTPF is not a supported platform in OpenJ9 so there is no way for us to test either way. I think what you have is fine.

fjeremic:
Jenkins test sanity zlinux

Issue #1931, opened 2018-05-17
Duplication in String.java for different Java versions


fjeremic:
As pointed out by @keithc-ca in https://github.com/eclipse/openj9/pull/1786#issuecomment-389915950 once #1786 gets merged there will be quite a bit of duplication in String.java for Java 8 and other Java versions. We effectively split the file into two with the preprocessor macros.

Are there any alternatives to doing this? Would it be more clean to partition these ifdefs into separate files? How would this work? Would this be more intuitive than what we have today?

Any suggestions or comments are welcome!

keithc-ca:
I don't think there's a _best_ answer, but I think we can improve the situation considerably by organizing String.java like this:
```
public final class String ... {
    // common features
/*[IF Sidecar19-SE]*/
    // Java 9+ implementations
/*[ELSE]*/
    // Java 8 implementations
/*[ENDIF]*/
}
```
There will still be a lot of duplication (e.g. javadoc for API with more than one implementation), but I think it's a step in the right direction. We can make further improvements later if warranted.

PR #1936, opened 2018-05-17
Throw IAE when last char in replacement string is '$' or '\'

JasonFengJ9:
Throw `IAE` when last char in replacement string is `$` or `\`

The last character of `String substitute` in `j.l.String.replaceAll(regex, substitute)` can't be `$` which is treated as references to captured subsequences, or `\` which is used to escape literal characters.
`$` requires a group index to follow, and `\` requires a character to be escaped.
Added test to verify this behaviour.

Reviewer: @pshipton 
FYI: @DanHeidinga 

### DIFF
--- a/jcl/src/java.base/share/classes/java/lang/String.java
+++ b/jcl/src/java.base/share/classes/java/lang/String.java
@@ -1,5 +1,4 @@
-/*[INCLUDE-IF Sidecar16]*/
-
+/*[INCLUDE-IF Sidecar18-SE]*/
 package java.lang;
 
 /*******************************************************************************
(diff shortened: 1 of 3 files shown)


pshipton:
I think the tests should be in java.lang.String and not java.lang.Class

JasonFengJ9:
Moved the tests to `Test_String` (somehow that `Test_Class` stuck in mind), thanks for catching that.
Rebased and made change to both `j.l.String.replaceAll()`.

jcl/src/java.base/share/classes/java/lang/String.java
```diff
@@ -7301,6 +7318,21 @@ public boolean matches(String expr) {
 		return Pattern.matches(expr, this);
 	}
 
+	/*
```
pshipton: I think we can have one copy of common code, i.e. #1931. Please move the helper so its shared between both versions.

JasonFengJ9: Shared the helper method with minimum modification to existing duplication. 
Will open separated PR to move other shared field/methods.
Please review.

pshipton:
jenkins test sanity xlinux jdk8,jdk10

PR #1939, opened 2018-05-18
Pass through tool selections to OMR

keithc-ca:
* respect tool choices made at configuration time.
* fix Linux 390 default for CXX


DanHeidinga:

DanHeidinga:

DanHeidinga:

pshipton:
Was asking Keith why we set a default for CXX and not CC and now I believe he's going to make additional changes.

keithc-ca:
Updated to supply a default for CC (gcc) and use g++ as the default for CXX.
For the xlc toolchain on Power, don't assume that CC is a C++ compiler.

DanHeidinga:
@pshipton I'm still OK with this change.  Do you any further comments?

pshipton:
lgtm

jenkins compile xlinux,plinux,zlinux jdk8

pshipton:
PR tested passed, but got an error afterward, which I assume is related to eclipse ci being down.

PR #1942, opened 2018-05-18
Add support for JNI_VERSION_10

theresa-m:

@pshipton


runtime/vm/jnimisc.cpp
```diff
@@ -411,7 +411,9 @@ getVersion(JNIEnv *env)
 	J9VMThread *currentThread = (J9VMThread*)env;
 	J9JavaVM *vm = currentThread->javaVM;
 
-	if (J2SE_VERSION(vm) >= J2SE_19) {
+	if (J2SE_VERSION(vm) >= J2SE_V10) {
+		return JNI_VERSION_10;
+	} else if (J2SE_VERSION(vm) >= J2SE_19) {
```
DanHeidinga: Should this be == now that there's a new constant for 10?

theresa-m: the result should be the same. I will add the == in.

theresa-m: > The only concern I have is that the code allows using newer JNI constants with older vms.
@DanHeidinga do you mean for versions below 1.8?

DanHeidinga:
> > The only concern I have is that the code allows using newer JNI constants with older vms.

The existing supported / valid version checks allow JDK 8 to use `JNI_VERSION_9`.  This change extends that to `JNI_VERSION_10` as well.  

Not something that needs to be addressed in this PR but its a possible future cleanup task

DanHeidinga:
Jenkins test sanity plinux jdk10

pshipton:
>  its a possible future cleanup task

Agreed. I don't think its that easy to fix entirely because not all the code doing the checking knows which version is running. Although I think a subset of the cases could be fixed if we used BFUjavaVM in JVM_IsSupportedJNIVersion()

pshipton:
The best place to fix is jniVersionIsValid(), and most places have a J9JavaVM that could be passed along. Its problematic for the initial call in J9_CreateJavaVM(), but we could allow all versions from there.

Created #1944 for follow up.

Issue #999, opened 2018-01-22
The nightly build (jdk-9+181-20 on January 22 2018) has different behaviors from OpenJDK


tianxiaogu:
[C0.class (via Google Drive)](https://drive.google.com/open?id=1CdaiP02IUO32hKKbilQIQiFhm2hRg97d) triggers different behaviors on OpenJDK and OpenJ9.

The nightly build (jdk-9+181-20 on January 22 2018) throws an exception in default mode,
where OpenJDK (head) exits normally.

This bug may relate to #997. I am not sure whether this is a bug of OpenJDK or OpenJ9.

OpenJ9-JDK9
```
$ ./jdk-9+181/bin/java C0
Exception in thread "main" java.lang.NegativeArraySizeException
	at C0.main(Unknown Source)
$ ./jdk-9+181/bin/java -Xcomp C0
Exception in thread "main" java.lang.NegativeArraySizeException
	at C0.main(Unknown Source)
$ ./jdk-9+181/bin/java -version
openjdk version "9-internal"
OpenJDK Runtime Environment (build 9-internal+0-adhoc.jenkins.openjdk)
Eclipse OpenJ9 VM (build 2.9, JRE 9 Linux amd64-64 Compressed References 20180122_95 (JIT enabled, AOT enabled)
OpenJ9   - d38352e
OMR      - cfd2f17
OpenJDK  - be5903d based on jdk-9+181)
```

OpenJDK
```
$ ./openjdk/jdk/build/linux-x86_64-normal-server-release/jdk/bin/java C0
$ echo $?
0
$ ./openjdk/jdk/build/linux-x86_64-normal-server-release/jdk/bin/java --version
openjdk 10-internal 2018-03-20
OpenJDK Runtime Environment (build 10-internal+0-adhoc.t.jdk)
OpenJDK 64-Bit Server VM (build 10-internal+0-adhoc.t.jdk, mixed mode)
```

Oracle JDK 9.0.4
```
$ ./jdk-9.0.4/bin/java C0
$ echo $?
0
$ ./jdk-9.0.4/bin/java -version
java version "9.0.4"
Java(TM) SE Runtime Environment (build 9.0.4+11)
Java HotSpot(TM) 64-Bit Server VM (build 9.0.4+11, mixed mode)
```

pshipton:
What happens when you run `./jdk-9+181/bin/java -Xint C0`?

tianxiaogu:
[...]

PR #1943, opened 2018-05-18
WIP: Add bytecodes for returning byte,char,short and bool

fjeremic:
Add infrastructure for new bytecodes defined by the the VM as part of
#1780. Emit TR IL representing these new bytecodes.


fjeremic:
This PR is in support of #999 however it is untested. We need to forge some unit tests for all of these new bytecodes and add them to the continuous testing hence why this PR is marked as WIP. @dnakamura / @DanHeidinga FYI.

I'd like to request a review from @jdmpapin / @andrewcraik in the mean while.

runtime/compiler/trj9/ilgen/Walker.cpp
```diff
@@ -7037,7 +7053,7 @@ TR_J9ByteCodeIlGenerator::genMultiANewArray(int32_t dims)
 //----------------------------------------------
 
 int32_t
-TR_J9ByteCodeIlGenerator::genReturn(TR::ILOpCodes nodeop, bool monitorExit)
+TR_J9ByteCodeIlGenerator::genReturn(TR::ILOpCodes nodeop, TR::ILOpCodes truncationConversion, TR::ILOpCodes extensionConversion, bool isZConversion, bool monitorExit)
```
andrewcraik: Rather than polluting the genReturn call with flags why can't we just use the bytecode with a switch statement in the genReturn method to set these conversion opcodes? It will make the method much more self contained... We are trying to avoid Boolean flags on APIs as much as possible.

fjeremic: I completely agree. The updated API indeed looks much better. Fixed in 8d8a7f4

fjeremic:
@dnakamura the JIT portion is now complete. I suppose we can merge this as it is not enabled by default but whatever PR ends up enabling this support I'd like to see some unit tests checked in along with it to make sure these paths are regularly exercised. Given your VM expertise and deeper understanding of #999 do you think it is reasonable if I delegate this task to you?

@andrewcraik could you please review and be the committer for this change?

andrewcraik:
Jenkins test sanity

andrewcraik:
change looks fine - just waiting for sanity to complete. Should be fine to merge ahead of VM changes as long as the bytecode enums will line up.

DanHeidinga:
JVM infrastructure for the truncating return bytecodes has been merged already.  The pending VM change is to start generating the new return bytecodes: https://github.com/eclipse/openj9/pull/1831

fjeremic:
Jenkins test sanity

fjeremic:
Windows failures are unrelated infrastructure issues. @andrewcraik I think we should be good to go here.

PR #1947, opened 2018-05-18
Implement default behaviour for ClassLoader.findResource

pdbain-ibm:
Some subclasses of ClassLoader do not implement
findResource(moduleName, resName).
Implement the default behaviour (return result of
findResource(String resName) ).

### DIFF
--- a/jcl/src/java.base/share/classes/java/lang/ClassLoader.java
+++ b/jcl/src/java.base/share/classes/java/lang/ClassLoader.java
@@ -2285,7 +2285,8 @@ ServicesCatalog getServicesCatalog() {
 /**
  * Answers an URL which can be used to access the resource
  * described by resName, using the class loader's resource lookup
- * algorithm. The default behavior is just to return null.
+ * algorithm. By default, return null, unless moduleName is null,
+ * in which case return findResource(resName).
  * This should be implemented by a ClassLoader.
  *
  * @return		URL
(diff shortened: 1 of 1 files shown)


pshipton:
lgtm

pshipton:
jenkins test all plinux jdk9,jdk10

pshipton:
The eclipse ci is down, will need to run PR tests later

pshipton:
jenkins test all plinux jdk9,jdk10

pshipton:
Please fix the copyright and then this is good to go.

pdbain-ibm:
Copyright failures were due to unrelated files which were stale.
Do we need this (undocumented) method? I cannot find any users.
```
InputStream getResourceAsStream(String moduleName, String name) throws IOException
{
	return null;
}
```

pshipton:
Agreed ClassLoader.getResourceAsStream(String, String) doesn't seem to be used.

pdbain-ibm:
Per https://github.com/eclipse/openj9/pull/1947#issuecomment-390979609
https://github.com/eclipse/openj9/pull/1953 removes getResourceAsStream(String, String)

pdbain-ibm:
@DanHeidinga per our previous conversation, I have a test for findResource(null, "some.resource")
(see above).  For completeness, we should have a test for findResource("some.module", "some.resource"), but that will take a while longer.  Do you want the partial test now or the whole thing later?

PR #1948, opened 2018-05-18
GAM teardown

amicic:
Remove GlobalAllocationManager teardown from
ConfigurationIncrementalGenerational teardown. Rely on base
Configuration teardown to do it (at a later point, after Regions through
MemorySubspace have been torn down, which rely on GAM and more
specifically AllocationContext's which GAM owns).

Besides ATM being redudant, it will soon be even incorrect staying where
it is now, once https://github.com/eclipse/omr/pull/2564 is merged,
which will make MemorySpace be torn down later, after
ConfigurationIncrementalGenerational specific teardown, hence reversing
the order of MemorySpace and GAM teardown.


charliegracie:

amicic:
Yes, it's a standalone change. I, however, just changed the copyright year, so it'll take a bit for checks to complete again.

Issue #1117, opened 2018-02-06
Remove J9_PROJECT_SPECIFIC guards from compiler


0xdaryl:
Remove the `J9_PROJECT_SPECIFIC` guards around code in OpenJ9 as they are not necessary.

14 uses found:

```
compiler/trj9/optimizer/InlinerTempForJ9.cpp:#ifdef J9_PROJECT_SPECIFIC
```

tmcdermo:
I'd like to tackle this as a means for getting my feet wet in this project (and open source as a whole).  

Quick question though: When you say these guards are not necessary, does that mean it always evaluates to true?  I see the 14 uses mentioned but most are in #ifdef statements combined with #else.  Is it safe to assume the #else statements never execute and can be removed with the all of these?

fjeremic:
> Quick question though: When you say these guards are not necessary, does that mean it always evaluates to true?

That's right. This is because the macro is defined directly in the makefiles for the entire project:

https://github.com/eclipse/openj9/blob/26f8c7edb1fe37253d001243c0ce14473334d07f/runtime/compiler/trj9/build/toolcfg/common.mk#L51

PR #1952, opened 2018-05-22
Remove J9_PROJECT_SPECIFIC guards

tmcdermo:
Remove J9_PROJECT_SPECIFIC guards

This patch removes all 14 J9_PROJECT_SPECIFIC guards from
InlinerTempForJ9.cpp.

Closes: #1117
Signed-off-by: Thomas McDermott <mcdermot@oregonstate.edu>


andrewcraik:
Jenkins test sanity

fjeremic:
Sadly our builds seem to be having a bad day. I suspect it's related to all the infrastructure issues with disk space over at Eclipse CI. I checked all the failing builds and it's all related to infrastructure. I would say this PR should be good to go.

andrewcraik:
I agree the change looks fine - I'd just like to see some sanity builds before the official merge; welcome @tmcdermo!

andrewcraik:
so the two failures are infrastructure, the test has passed on win_x86-64_comprssptrs for Java 8 and the same code will run across all platforms and for all JVM versions so this should be good to merge. Change is also easy to visually verify so I will merge this now.

PR #1953, opened 2018-05-22
Remove unused getResourceAsStream(String, String)

pdbain-ibm:
ClassLoader.getResourceAsStream(String, String) has package scope but is not
documented and is not used.


DanHeidinga:

PR #1959, opened 2018-05-22
Revert "Adding missed OS check for GS support in Port Library for zLinux"

dmitripivkine:
Reverts eclipse/openj9#1929


fjeremic:
@dmitripivkine I understand the need for this PR but there is no need for the churn. Let's just wait for @ebadyano to fix up https://github.com/eclipse/omr/pull/2556 which is a simple two line change.

PR #1962, opened 2018-05-22
Remove invalid assertion check

LinHu2016:
The committed size of Usage could be decreasing from later
	MemoryPoolMXBean.getXXXUsage() call, if GC happens between 
	the calls and GC decides to contrast the memory pool.


llxia:
- we suspect there are other similar issue in the test, so we added comment and commented out the assertion for future reference.
- merge the change without testing as the PR only commented out the assertion.

PR #1965, opened 2018-05-23
Read Barrier for CompareAndSwapObject

fjeremic:
Compare-And-Swap on object reference, while primarily is a store
operation, it is also an implicit read (it reads the existing value to
be compared with a provided compare value, before the store itself),
hence needs a read barrier.

This change is an extension of #1573 with JIT support.

### DIFF
--- a/runtime/compiler/trj9/z/codegen/J9TreeEvaluator.cpp
+++ b/runtime/compiler/trj9/z/codegen/J9TreeEvaluator.cpp
@@ -9756,6 +9756,7 @@ static TR::Register *VMinlineCompareAndSwap(
       // Compare-And-Swap on object reference, while primarily is a store operation, it is also an implicit read (it 
       // reads the existing value to be compared with a provided compare value, before the store itself), hence needs
       // a read barrier
+      generateS390IEInstruction(cg, TR::InstOpCode::NIAI, 1, 0, node);
       generateRXYInstruction(cg, guardedLoadMnemonic, node, tempReadBarrier, generateS390MemoryReference(*casMemRef, 0, cg));
 
       cg->stopUsingRegister(tempReadBarrier);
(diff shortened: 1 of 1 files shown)


fjeremic:
@amicic, @dsouzai, @joransiu could I please get a review?

Note there are also compare and swap instructions generated in `inlineAtomicOps` and `inlineDoubleWordCAS`. The former only deals with primitives so it does not need a read barrier. The latter does have a path for objects but the code is completely dead (the associated Java recognized methods are deprecated and do not exist) and I will remove it in a subsequent PR I'm working on right now.

fjeremic:
> Btw, I bet, in real life, very often just a few instructions before this one you'll have a similar one, loading the value to oldVReg (and again very likely, no async check in-between), which would make this one redundant. But I guess, it's hard to prove it.

I suppose we can write a peephole pass in the future to validate this and remove the redundant read barrier. This is a correctness issue though while a peephole pass would be a performance issue. I'll leave things as is for now.

> On the other side, the extra cost of this guarded read should not be that high (data would be in L1 already), anyway.

And the guarded load would not trigger in that case either so yes the load will be as fast as it can be.

Thanks for the review!

fjeremic:
> The latter does have a path for objects but the code is completely dead (the associated Java recognized methods are deprecated and do not exist) and I will remove it in a subsequent PR I'm working on right now.

https://github.com/eclipse/openj9/pull/1966

fjeremic:
> We should consider adding an NIAI in front of the guarded load, to ensure that the cache line is loaded in the write exclusive state for CAS.

This is a good suggestion. Added in c451080. @joransiu could you also be a committer for this PR please?

joransiu:
Jenkins test sanity

fjeremic:
Just a heads up that the Jenkins builds will almost certanly fail because we are waiting for a new level of OMR to be promoted. We may have to re-launch these once this has happened. I can luanch the new builds on your behalf as I'm tracking https://github.com/eclipse/openj9-omr/tree/openj9 like a hawk!

fjeremic:
Jenkins test sanity

fjeremic:
Windows failures are unrelated infrastructure issues. @joransiu I think we should be good to go here.

PR #1967, opened 2018-05-23
Use the OpenJ9 SHA to distinguish different builds for SCC.

### LINKED ISSUE #1645: -Xshareclasses unique id unsupported by OpenJ9 builds
The -Xshareclasses unique build id generation is specific to the IBM internal build process and needs to be fixed to support the OpenJ9 build process.

-Xshareclasses uses a unique build id to identify which JVM created the shared cache. If a JVM with a different build id attempts to attach to the cache, the cache is destroyed (when possible) and recreated. This mechanism helps avoid stale entries in the cache, and incompatibilities that can occur in the rom class representation as the implementation is changed.

hangshao0:
1. Add a helper to turn the OpenJ9 SHA string (J9VM_VERSION_STRING)
 to a number.
2. Set header->buildID to OpenJ9 SHA.

Fixes #1645

### DIFF
--- a/runtime/oti/j9argscan.h
+++ b/runtime/oti/j9argscan.h
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2001, 2015 IBM Corp. and others
+ * Copyright (c) 2001, 2018 IBM Corp. and others
  *
  * This program and the accompanying materials are made available under
  * the terms of the Eclipse Public License 2.0 which accompanies this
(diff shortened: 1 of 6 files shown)


hangshao0:
Reviewer: @pshipton

runtime/util_core/j9shchelp.c
```diff
@@ -373,3 +373,13 @@ isCacheFileName(J9PortLibrary* portlib, const char* nameToTest, uintptr_t expect
...
 	return 0;	
 }
+
+/* Return the OpenJ9 SHA */
+uint64_t
+getOpenJ9Sha()
+{
+	uint64_t sha = 0;
+	char *str = J9VM_VERSION_STRING;
+	scan_hex_u64(&str, &sha);
```
pshipton: There should be checks (or a test) to ensure getting the sha is working. If scan_hex_u64() returns the number of bits, the code could check there are at least 28 bits and the sha value is non-zero, otherwise assert.

How does j9version.h get #included in j9shchelp.c, I don't see it included by any .h file.

pshipton: and note we don't want j9version.h included by a .h file as this isn't good for incremental builds.

hangshao0:
Updates: 
1. Change scan_hex_caseflag_u64() to return the number of bits of the hexadecimal number
2. Assert if SHA is less than 28 bits or it is 0
3. Move getOpenJ9Sha() to /util/shchelp_j9.c (we are using Assert_VMUtil_ShouldNeverHappen() which is only visible under /util)
4. include j9version.h in shchelp_j9.c

hangshao0:
The CI failure is due to

> The job exceeded the maximum time limit for jobs, and has been terminated.

pshipton:

PR #1980, opened 2018-05-24
Fix for getting hottest profiled method

liqunl:
This commit has the following changes:
- Fix the problem that getMethodsList returns a mix of methods and
  classes
- Use isSameMethod instead of '==' to test the equality of two
  TR_ResolvedMethods in getMethodsList
- Use classOfMethod to get the target class

### DIFF
--- a/runtime/compiler/optimizer/J9Inliner.cpp
+++ b/runtime/compiler/optimizer/J9Inliner.cpp
@@ -924,6 +924,10 @@ void TR_ProfileableCallSite::findSingleProfiledMethod(ListIterator<TR_ExtraAddre
       {
       return;
       }
+   TR_OpaqueClassBlock* callSiteClass = _receiverClass ? _receiverClass : getClassFromMethod();
+   if (!callSiteClass)
+      return;
+
    // first let's do sanity test on all profiled targets
    if (comp()->trace(OMR::inlining))
       traceMsg(comp(), "No decisive class profiling info for the virtual method, we'll try to see if more than one class uses the same method implementation.\n");
@@ -945,7 +949,7 @@ void TR_ProfileableCallSite::findSingleProfiledMethod(ListIterator<TR_ExtraAddre
          {
          TR_SharedCache *sharedCache = fej9->sharedCache();
          if (!sharedCache->canRememberClass(clazz) ||
-             !sharedCache->canRememberClass(_initialCalleeMethod->classOfMethod()))
+             !sharedCache->canRememberClass(callSiteClass))
             {
             classValuesAreSane = false;
             break;
@@ -961,7 +965,7 @@ void TR_ProfileableCallSite::findSingleProfiledMethod(ListIterator<TR_ExtraAddre
 
    TR_ScratchList<TR_AddressInfo::ProfiledMethod> methodsList(comp()->trMemory());
    // this API doesn't do a sort
-   valueInfo->getMethodsList(comp(), _callerResolvedMethod, calleeClass(), _vftSlot, &methodsList);
+   valueInfo->getMethodsList(comp(), _callerResolvedMethod, callSiteClass, _vftSlot, &methodsList);
 
    int numMethods = methodsList.getSize();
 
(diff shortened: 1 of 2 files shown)


liqunl:
@andrewcraik Can you review my changes?

liqunl:
Changes have passed the personal build.

runtime/compiler/trj9/optimizer/J9Inliner.cpp
```diff
@@ -973,16 +977,13 @@ void TR_ProfileableCallSite::findSingleProfiledMethod(ListIterator<TR_ExtraAddre
    TR_AddressInfo::ProfiledMethod *profiledMethodInfo;
    TR_AddressInfo::ProfiledMethod *bestMethodInfo = methodValuesIt.getFirst();
 
-   TR_AddressInfo::ProfiledMethod* firstBestMethodInfo = NULL;
```
andrewcraik: why is this removed?

runtime/compiler/trj9/optimizer/J9Inliner.cpp
```diff
@@ -998,52 +999,10 @@ void TR_ProfileableCallSite::findSingleProfiledMethod(ListIterator<TR_ExtraAddre
             static const char* userMinProfiledCallFreq = feGetEnv("TR_MinProfiledCallFrequency");
             static const float minProfiledCallFrequency = userMinProfiledCallFreq ? atof (userMinProfiledCallFreq) : MIN_PROFILED_CALL_FREQUENCY;
 
-            if (methodProbability >= minProfiledCallFrequency ||
-                  (firstBestMethodInfo && methodProbability >= SECOND_BEST_MIN_CALL_FREQUENCY))
```
andrewcraik: Why is the functionality around SECOND_BEST_MIN_CALL_FREQUENCY removed?

liqunl:
@AdamBrousseau One job failed in fetching remote repo, is there a way to restart the job?

AdamBrousseau:

andrewcraik:
So there is a functional difference between between the new implementation and the old - namely if PrexArgInfo proves we can't call a target we won't inline it. In the original implementation we would inline it to later rip it out during VP or a similar opt. I don't know that we want to reject this change because of that change in behaviour, but it will require full perf validation across DayTrader, HiBench etc because rejecting the candidate could allow something else to inline leading to different (hopefully better) performance. @liqunl can you do some perf validation and let us know if you see any problems as a result of the behviour change @cathyzhyi has highlighted?

liqunl:
Perf numbers, the following result is based on 8 iterations of interleaved runs of baseline and the fix build.

Performance on daytrader3 on broadwell, redhat 7.3, kernel version 3.10.0-514
Relative throughput fix build vs baseline, the higher the score, the better the performance

| Benchmark | Relative score| Standard deviation|
| daytrader3  | +0.13%  |0.45%

| Benchmark | Relative score| Standard deviation|
| Pagerank  | +1.10%  |1.5%
| Terasort | -5% |13%
|Wordcount|+0.3%|1.56%|
|Sort|-1%|3%|

The performance difference between the fix build and the baseline is within the noise margin for above benchmarks, thus the fix build is on par with the baseline.

andrewcraik:
@cathyzhyi can I get an opinion on this new iteration - looks good to me but would like to hear your thoughts before making a final call.

cathyzhyi:
LGTM

andrewcraik:
Jenkins test sanity

andrewcraik:
@liqunl I think a rebase might be needed due to the build failures...

liqunl:
Looking at the failures.

liqunl:
The failures are caused by https://github.com/eclipse/openj9/issues/2073 and has been fixed by https://github.com/eclipse/openj9/pull/2080. The commit has been rebased too include the fix.

andrewcraik:
Jenkins test sanity

liqunl:
On a second look at the test output, I got confused about the failure. @AdamBrousseau Any idea what the actual failure is? It looks like all tests passed except the result uploading part.

AdamBrousseau:
Looks like test failures
https://ci.eclipse.org/openj9/job/PullRequest-Sanity-JDK8-win_x86-64_cmprssptrs-OpenJ9/36/tapResults/
eg.
```
Running test cmdLineTester_jvmtitests_extended_3 ...
cmdLineTester_jvmtitests_extended_3 Start Time: Wed Jun  6 12:08:27 2018 Epoch Time (ms): 1528304907512
test with Mode154

cmdLineTester_jvmtitests_extended_3_FAILED
```

liqunl:
The failed tests indicate it's related to debug mode. After a bit of googling, I found that there're other people seeing `NPT ERROR: Cannot open library` when they turned on debug mode in java. I don't think my change can cause this error. Any ideas? @gacholio

liqunl:
Discussed with @gacholio This looks like a machine issue where the agent lib was missing or had wrong permission. The windows were out of space according to https://github.com/eclipse/openj9/issues/2107, it's very likely they're related. @andrewcraik @AdamBrousseau Can we restart Sanity -JDK8 -win_x86-64_comprssptrs test?

andrewcraik:
Jenkins test sanity win jdk8

andrewcraik:
ok the failures in the JDK8 build are being seen in the other PR builds and the change is not JVMTI specific. The failure modes are identical and the change is fully cross platform so I am going to merge this now while the JVMTI issue is investigated as all evidence says it is unrelated.

PR #1981, opened 2018-05-24
Generate all DDR pointer types

### LINKED ISSUE #1973: MM_BasePointer.java not generated by PointerGenerator
MM_BasePointer.java cannot be generated by PointerGenerator due to change made on line 278:
`structure.getConstants().size() > 1` was changed to `structure.getConstants().size() != 0` in by commit 70c18d5a81bb53b12276bf8716aec31843aaaddd

keithc-ca:
Whether a generated pointer class is needed is independent of how
many fields or constants it might declare. We could analyze all
the types and generate only those with fields, or that are used
by some other type, but not all references can be expressed in
the superset.dat file. In particular, the format of that file
permits a type to have at most one superclass, and there are C++
classes that employ multiple inheritance (e.g. SH_CompositeCacheImpl
inherits from both SH_CompositeCache and AbstractMemoryPermission).
As a result, a few extra classes will be generated, but they will be
small (they have no fields else they would been generated previously).
On zLinux, a few is about 140.

Minor cleanup:
* remove empty 'omitList' and its uses
* simplify generation of base class name

Fixes #1973

### DIFF
--- a/debugtools/DDR_VM/src/com/ibm/j9ddr/tools/PointerGenerator.java
+++ b/debugtools/DDR_VM/src/com/ibm/j9ddr/tools/PointerGenerator.java
@@ -88,7 +88,7 @@ public class PointerGenerator {
 	private static final String BEGIN_USER_CODE = "[BEGIN USER CODE]";
 	private static final String END_USER_CODE = "[END USER CODE]";
 
-	final Map<String, String> opts = new HashMap<String, String>();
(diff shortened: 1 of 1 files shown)


pshipton:
jenkins compile plinux,zlinux,aix,win jdk10

pshipton:
jenkins test all xlinux jdk8

PR #1982, opened 2018-05-24
Link using xlc_r because using xlC_r results in libibmc++.so.1 not found

keithc-ca:


DanHeidinga:

fjeremic:
@DanHeidinga would love to see this merged soon as it is causing hiatus on our Linux Power builds. Thank you very much!

PR #1983, opened 2018-05-24
Unfinalized backout for CS

amicic:
Align the way how the unfinalized list is updated in backout for
Concurrent Scavenger. It's same as for the finalizable list.

Effectively, 'backout' in CS does not move objects back where there are;
at leaves them where there are, but the only thing it backs out is the
potential self forwarding pointer. Overall there are 3 cases:

- object successfully moved: get the new location via f/w pointer and
connect the new version of the object to the unfinalized list (same as
if there was no aborted Scavenge)
- object did not move, and was never attempted to move: connect the
(original) object
- object did not move, but was not successfully moved: remove the
self-forwarding bit, and connect the object.

An unrelated change: #include GlobalCollector instead of Collector. This
is just a pre-req change for a future OMR change.


dmitripivkine:

Issue #1950, opened 2018-05-21
Reflective method invocation may work incorrectly after class redefinition 


Lerm:
Virtual method call through reflection API may work incorrectly after base class is redefined by Java agent.
Here is the simplified description test case for the problem – we have child class (Bar) which overrides methods of its base class (Foo), then base class is redefined using Instrumentation API (Java Agent):
```java
public class Foo {
  private String v;
  public void setValue(String v) { this.v = v;  }
  public String getValue() { return v; }
}
public class Bar extends Foo {
  private String v;
  public void setValue(String v) { this.v = v; }
  public String getValue() { return v; }
}
...
// Redefine class Foo here (Intstrumentation.redefineClasses)
...
Foo foo = new Bar();
foo.setValue("Bar");
System.out.println("Direct getValue() = " + foo.getValue());
System.out.println("Reflection getValue() = " + Foo.class.getMethod("getValue").invoke(foo));
```

Expected result – same value ("Bar") is returned from both calls.
Actual result – direct call returns correct value, while call via reflection returns null.

Once Foo class is redefined (without actual changes to its bytecode), then attempt to access overridden methods of Bar class using reflection-based methods for Foo class gives wrong results – call will be dispatched to the Foo implementation instead of Bar implementation (i.e. virtual invocation is incorrectly dispatched to the super class). At the same time, direct virtual call (without using reflection) is dispatched correctly.

The problem is not reproduced under following conditions:
- If Foo.class.getMethod('getValue') call is performed before class redefinition
- If interpreter is used (-Xint)

This problem was observed on Linux x86-64 IBM Java 8u161 and is reproduced on the latest OpenJ9 jdk10 build from adoptopenjdk:
```
openjdk version "10.0.1-adoptopenjdk" 2018-04-17
OpenJDK Runtime Environment (build 10.0.1-adoptopenjdk+0-adhoc.jenkins.openjdk) Eclipse OpenJ9 VM (build master-25d93508, JRE 10 Linux amd64-64-Bit Compressed References 20180518_54 (JIT enabled, AOT enabled)
OpenJ9   - 25d93508
OMR      - 15290506
JCL      - dee0290582 based on jdk-10.0.1+10)
```

Attached is the full test case to reproduce the problem (unzip, set path to JDK in setenv.sh, run build.sh, then run.sh): [agent-test.zip](https://github.com/eclipse/openj9/files/2021943/agent-test.zip)

DanHeidinga:
[...]

PR #1989, opened 2018-05-24
Always re-initialize JNI method IDs after class redefinition

gacholio:
In fast HCR, newly-created JNI method IDs were being created incorrectly
(vTableIndex of 0), resulting in JNI invocations beaing treated as
non-virtual.
 
Fixes: #1950


DanHeidinga:

PR #1993, opened 2018-05-25
Increase timeout on OMR build form 6 to 8 hours

AdamBrousseau:
- We had a build during the day that was almost complete
  but timed out after 6 hours. It timed out becasue it
  spent alot of time waiting for a machine. Increasing
  to 8 hours will likely be sufficient for the time being.


pshipton:
I don't think 8 hours is enough, although we can check again after #2020. The build https://ci.eclipse.org/openj9/view/Pipelines/job/Pipeline-OMR-Acceptance/492/ has one platform to go which is part way done, but only 15 min to complete before the timeout kills it.

pshipton:
Made it with a few minutes to spare, but I don't think we should be cutting it that close.

PR #1995, opened 2018-05-25
Open modules export all packages just like unnamed modules

DanHeidinga:
Update the `isPackageExportedToModuleHelper` function to treat open
modules exactly like unnamed ones.  All packages in such modules
should be treated as exported.

issue #1789

### DIFF
--- a/runtime/util/modularityHelper.c
+++ b/runtime/util/modularityHelper.c
@@ -69,7 +69,7 @@ isAllowedReadAccessToModule(J9VMThread *currentThread, J9Module *fromModule, J9M
 	BOOLEAN canRead = FALSE;
 	if ((fromModule == toModule) /* modules are the same, no need to check */
 		|| (toModule == vm->javaBaseModule) /* java.base is implicitly read by every module, no need to check */
-		|| (J9_IS_J9MODULE_UNNAMED(vm, fromModule)) /* unamed Modules can read all other modules */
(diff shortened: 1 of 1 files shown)


DanHeidinga:
Marking as "WIP" as I'm having trouble re-building the petclinic modularity demo to validate this change.

DanHeidinga:
I haven't been able to confirm if this fixes petclinic yet but have confirmation that other open module tests are passing.  This is at least a positive step forward

JasonFengJ9:
The change looks good.

Just noticed that `visible.c:checkModuleAccess()` can check `unnamed | open` modules which export all packages, hence skip the call to `isPackageExportedToModuleWithName()`. I think this can be done in a separated PR though.

DanHeidinga:
I'd prefer to keep that logic in `isPackageExportedToModuleWithName()` so that there's only once place to modify it.

gacholio:
jenkins test sanity plinux jdk9

JasonFengJ9:
`isPackageExportedToModuleWithName` invokes `isPackageExportedToModuleHelper(currentThread, fromModule, getPackageDefinitionWithName(currentThread, fromModule, packageName, len, errCode), toModule, toUnnamed)` in which `getPackageDefinitionWithName` might not be a cheap call for `j9mem_allocate_memory`, `hashPackageTableAtWithUTF8Name`, etc.

Agreed that more thinking is needed to avoid spreading out access checking code.

PR #1997, opened 2018-05-25
Fix a debug build failing due to incorrect reference count

r30shah:
In building transaction regions, when we have a monent under nullchk, we rip out nullchk to new tree and replace it's first child with newly created PassThrough node. We should increase reference count of PassThrough node when we set it as first child of NullChk node. Incorrect reference count was causing asserts to be thrown.


r30shah:
@andrewcraik While working on fixing disabling register pressure simulation, ran into different issue with debug with assert build. This one fixes that. Would you please review this?

fjeremic:
Jenkins test sanity

fjeremic:
Jenkins test sanity win,zlinux

PR #2004, opened 2018-05-28
Fix JVM_RegisterSignal

### LINKED ISSUE #1984: Signals not working properly for Apache Kafka
How to reproduce the issue?

**Instructions:** https://kafka.apache.org/quickstart

- Step 1: Download the code
`tar -xzf kafka_2.11-1.1.0.tgz; cd kafka_2.11-1.1.0`

- Step 2: Start the server in a new shell or in background (`&`)
`bin/zookeeper-server-start.sh config/zookeeper.properties`

- Step 3: Start the Kafka server in a new shell or in background (`&`)
`bin/kafka-server-start.sh config/server.properties`

**Observations:** Kafka overwrites the default signal handlers for `SIGTERM`, `SIGHUP` and `SIGINT`: [related Kafka source code](https://github.com/apache/kafka/blob/ed8b0315a6c3705b2a163ce3ab4723234779264f/core/src/main/scala/kafka/Kafka.scala#L58).

Kafka also adds a [shutdown hook](https://github.com/apache/kafka/blob/ed8b0315a6c3705b2a163ce3ab4723234779264f/core/src/main/scala/kafka/Kafka.scala#L58).

**Issue:** When a signal is raised (`SIGTERM`, `SIGHUP` and `SIGINT`), the newly assigned signal handler is invoked in an infinite loop. Need to investigate this behavior further.

babsingh:
1. Add j9sig* macro for the newly added omrsig_is_master_signal_handler

	omrsig_is_master_signal_handler is a newly added OMR function. So,
	j9sig_is_master_signal_handler macro is added to invoke
	omrsig_is_master_signal_handler from J9.

2. Fix JVM_RegisterSignal

	JVM_RegisterSignal should return J9_USE_OLD_JAVA_SIGNAL_HANDLER if old
	OS handler is a master signal handler. If old OS handler is a master
	signal handler, then a Java signal handler was previously registered
	with the signal. sun.misc.Signal.handle(...) or
	jdk.internal.misc.Signal.handle(...) will return the old Java signal
	handler if JVM_RegisterSignal returns J9_USE_OLD_JAVA_SIGNAL_HANDLER.
	Otherwise, an instance of NativeHandler is returned with the
	oldHandler's address stored in NativeHandler.handler. 

	Java 8 - NativeHandler is sun.misc.NativeSignalHandler
	Java 9 - NativeHandler is jdk.internal.misc.Signal.NativeHandler

Fixes #1984

### DIFF
--- a/runtime/j9vm/jvm.c
+++ b/runtime/j9vm/jvm.c
@@ -327,7 +327,7 @@ static BOOLEAN librariesLoaded(void);
 #define J9_SIG_IGNORED 1
 
 #define J9_PRE_DEFINED_HANDLER_CHECK 2
-#define J9_OLDHANDLER_SAME_AS_NEWHANDLER 2
+#define J9_USE_OLD_JAVA_SIGNAL_HANDLER 2
 
 #define J9_SIG_PREFIX "SIG"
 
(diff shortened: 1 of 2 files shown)


babsingh:
This pull request depends on https://github.com/eclipse/omr/pull/2598

runtime/vm/jvminit.c
```diff
@@ -6360,6 +6360,13 @@ registerOSHandler(J9JavaVM *vm, U_32 signal, void *newOSHandler, void **oldOSHan
 	return rc;
 }
 
+BOOLEAN
+isMasterSignalHandler(J9JavaVM *vm, void *osHandler)
+{
+	PORT_ACCESS_FROM_JAVAVM(vm);
+	return j9sig_is_master_signal_handler(osHandler);
```
pshipton: This is just a port library call, does it really need to be an internal function?

pshipton: Guess the same question applies to registerPredefinedHandler() and registerOSHandler().

babsingh: just verified that i can directly call j9sig_* functions from jvm.c. will either remove the wrappers or move the wrappers in jvm.c.

babsingh: I removed isMasterSignalHandler.

I can't move registerPredefinedHandler since predefinedHandlerWrapper and signalDispatch can't be moved to jvm.c.

we should keep registerOSHandler since it will help in uniformly registering signal handlers within the JVM. it may help in resolving [Unclear use of OMRSIG_SIGNAL / OMRSIG_SIGACTION)](https://github.com/eclipse/openj9/issues/1889) (OpenJ9 issue).

babsingh:
eclipse/omr#2598 was merged 10 minutes ago; expecting OMR changes be in OpenJ9 pipeline soon.

pshipton:
yes, https://ci.eclipse.org/openj9/view/Pipelines/job/Pipeline-OMR-Acceptance/489/ will need to promote.

pshipton:
The OMR change has been accepted. I don't expect any testing will cover this change, a simple compile should be sufficient.

jenkins compile plinux jdk8

PR #2006, opened 2018-05-28
Correctly handle special cases of interface invocation

gacholio:
- When the resolved method is an Object method, the invokeinterface
bytecode was not ensuring that the receiver class implements the
targetted interface.

- Allow invokeinterface to target private interface methods.

- Allow JNI method IDs to target private interface methods.

- Allow InterfaceMethodHandle adapt private interface methods to a
DirectHandle in Java 11.  Full MethodHandle support for private
interface methods will likely require more work in the future.

- This does not fix the JIT for the private interface method case - that
requires further design work.


gacholio:
Closing for now while I do a bit more analysis on the JIT implications.

gacholio:
Turns out this breaks the JIT, so this will need to be rolled into the private virtual/interface wotk for JDK11.

PR #2012, opened 2018-05-29
Fix register dependency in x86 tstart evaluator

doubletea:
There is an explicit register dependency on eax when using the Intel RTM
transactional memory instructions. Currently, eax may become clobbered
by the RA as a result of adding debug counters, for example. This change
adds the appropriate register dependency condition on eax to fix this.

Signed-off-by: John Xu <xujohn@ca.ibm.com>


fjeremic:
> Closing this PR for now. Upon further investigation it seems to be related to debug counter problems on x86. @0xdaryl

@doubletea sounds like you need to go back to your old item and implement https://github.com/eclipse/openj9/pull/1589#issuecomment-379758756.

Issue #1977, opened 2018-05-24
jdk11 add String.strip(), stripLeading(), stripTrailing()


pshipton:
http://cr.openjdk.java.net/~iris/se/11/build/latest/

DanHeidinga:
@theresa-m Can you take a look at this?

PR #2015, opened 2018-05-29
 Added String.strip, stripLeading, stripTrailing

theresa-m:
Add new Java 11 methods

Fixes: #1977

### DIFF
--- a/jcl/src/java.base/share/classes/java/lang/String.java
+++ b/jcl/src/java.base/share/classes/java/lang/String.java
@@ -2524,7 +2524,69 @@ public final class String implements Serializable, Comparable<String>, CharSeque
 	public boolean startsWith(String prefix, int start) {
 		return regionMatches(start, prefix, 0, prefix.lengthInternal());
 	}
+	
+/*[IF Java11]*/
+	/**
+	 * Strip leading and trailing white space from a string.
+	 * 
+	 * @return a substring of the original containing no leading
+	 * or trailing white space
+	 * 
+	 * @since 11
+	 */
+	public String strip() {
+		String result;
+		
+		if (enableCompression && (null == compressionFlag || coder == LATIN1)) {
+			result = StringLatin1.strip(value);
+		} else {
+			result = StringUTF16.strip(value);
+		}
+		
+		return (result == null) ? this : result;
+	}
+	
+	/**
+	 * Strip leading white space from a string.
+	 * 
+	 * @return a substring of the original containing no leading
+	 * white space
+	 * 
+	 * @since 11
+	 */
+	public String stripLeading() {
+		String result;
+		
+		if (enableCompression && (null == compressionFlag || coder == LATIN1)) {
+			result = StringLatin1.stripLeading(value);
+		} else {
+			result = StringUTF16.stripLeading(value);
+		}
 
+		return (result == null) ? this : result;
+	}
+	
+	/**
+	 * Strip trailing white space from a string.
+	 * 
+	 * @return a substring of the original containing no trailing
+	 * white space
+	 * 
+	 * @since 11
+	 */
+	public String stripTrailing() {
+		String result;
+		
+		if (enableCompression && (null == compressionFlag || coder == LATIN1)) {
+			result = StringLatin1.stripTrailing(value);
+		} else {
+			result = StringUTF16.stripTrailing(value);
+		}
+		
+		return (result == null) ? this : result;
+	}
+/*[ENDIF]*/
+	
 	/**
 	 * Copies a range of characters into a new String.
 	 *
--- a/test/functional/Java11andUp/src/org/openj9/test/java_lang/Test_String.java
+++ b/test/functional/Java11andUp/src/org/openj9/test/java_lang/Test_String.java
@@ -34,6 +34,14 @@ import org.testng.log4testng.Logger;
  */
 public class Test_String {
 	public static Logger logger = Logger.getLogger(Test_String.class);
+	
+	/*
+	 * Useful variables for testing
+	 */
+	String empty = "";
+	String allWhiteSpace = "   ";
+	String latin1 = "abc123";
+	String nonLatin1 = "abc\u0153";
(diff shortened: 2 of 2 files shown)


jcl/src/java.base/share/classes/java/lang/String.java
```diff
@@ -2524,7 +2524,101 @@ public boolean startsWith(String prefix) {
 	public boolean startsWith(String prefix, int start) {
 		return regionMatches(start, prefix, 0, prefix.lengthInternal());
 	}
+	
+/*[IF Java11]*/
+	/**
+	 * Strip leading and trailing white spaces from a string.
+	 * 
+	 * @return a substring of the original containing no leading
+	 * or trailing white spaces.
```
pshipton: I think "spaces" should be "space" in all the places its used here.

test/functional/Java11andUp/src/org/openj9/test/java_lang/Test_String.java
```diff
@@ -63,4 +63,78 @@ public void testValueOfCodePoint() throws IllegalAccessException, IllegalArgumen
 			// Expected exception
 		}
 	}
+	
+	/*
+	 * Variables for strip leading/trailing white space tests
+	 */
+	String empty = "";
+	String noWhite = "test";
```
pshipton: Please add a test for a String containing a non-LATIN1 character.

jcl/src/java.base/share/classes/java/lang/String.java
```diff
@@ -2524,7 +2524,101 @@ public boolean startsWith(String prefix) {
...
+			if (stripTrailing) {
+				while ((end >= start) && Character.isWhitespace(charAtInternal(end))) {
+					end--;
+				}
+			}
+
+			if (start == 0 && end == last) {
+				return this;
+			} else {
+				return new String(value, start, end - start + 1, false);
```
pshipton: If the result is empty this can return "" instead of creating a new String object.

test/functional/Java11andUp/src/org/openj9/test/java_lang/Test_String.java
```diff
@@ -63,4 +63,78 @@ public void testValueOfCodePoint() throws IllegalAccessException, IllegalArgumen
...
+		// empty string parameter should return empty string
+		Assert.assertEquals(empty.strip(), empty, 
+				"strip: Striped empty string did not return an empty string.");
+		
+		// pass string with all white space
+		Assert.assertEquals(allWhite.strip(), empty, 
+				"strip: Striped all white space string did not return an empty string.");
+		
+		// pass in string without white space
+		Assert.assertEquals(noWhite.strip(), noWhite,
```
pshipton: This, and the other equivalent tests, should check for identity.

test/functional/Java11andUp/src/org/openj9/test/java_lang/Test_String.java
```diff
@@ -63,4 +63,78 @@ public void testValueOfCodePoint() throws IllegalAccessException, IllegalArgumen
...
+		// pass in string without white space
+		Assert.assertEquals(noWhite.stripTrailing(), noWhite, 
+				"stripTrailing: Striped string with no leading/trailing white space did not return original string.");
+
+		// pass in string with white space on either side
+		Assert.assertEquals((allWhite + noWhite + allWhite).stripTrailing(), allWhite + noWhite, 
+				"stripTrailing: Value of striped string with leading/trailing white space was unexpected.");
+
+	}
 }
```
pshipton: Please end the file with a newline.

jcl/src/java.base/share/classes/java/lang/String.java
```diff
@@ -2524,7 +2524,101 @@ public boolean startsWith(String prefix) {
...
 		return regionMatches(start, prefix, 0, prefix.lengthInternal());
 	}
+	
+/*[IF Java11]*/
+	/**
+	 * Strip leading and trailing white spaces from a string.
+	 * 
+	 * @return a substring of the original containing no leading
+	 * or trailing white spaces.
+	 */
```
pshipton: Please add `@since 11` to all the new API methods.

jcl/src/java.base/share/classes/java/lang/String.java
```diff
@@ -2524,7 +2524,101 @@ public boolean startsWith(String prefix) {
...
+		return stripSub(false, true);
+	}
+	
+	/**
+	 * Strip leading and/or trailing white spaces from a string
+	 * 
+	 * @param set to true if leading white spaces should be removed,
+	 * otherwise leading spaces are ignored.
+	 * @param set to true if trailing white spaces should be removed,
+	 * otherwise trailing spaces are ignored.
```
fjeremic: I think we need names after the `@param` clause. Also need to remove the ending period as per [1].

[1] http://www.oracle.com/technetwork/java/javase/documentation/index-137868.html#@param

theresa-m:
ready for another review. I've incorporated the StringLatin1 and StringUTF16 strip methods.

jcl/src/java.base/share/classes/java/lang/String.java
```diff
@@ -2524,7 +2524,90 @@ public boolean startsWith(String prefix) {
...
+		String result;
+		
+		if (enableCompression && (null == compressionFlag || coder == LATIN1)) {
+			result = StringLatin1.strip(value);
+		} else {
+			result = StringUTF16.strip(value);
+		}
+		
+		if (result == null) {
+			if (0 == lengthInternal()) {
```
fjeremic: As far as I can tell the StringLatin1 and StringUTF16 `strip` methods already check for this, so there is no need for a double check. Similarly for other methods.

theresa-m: good point. fixed that and retested.

pshipton:
Merging without running any testing since we don't have Java 11 PR testing yet.

PR #2018, opened 2018-05-30
Included validation for the presence of system dump data set on zOS

chandrams:
Included validation for the presence of system dump data set on zOS when the fileNamePattern is specified


chandrams:
@keithc-ca @smlambert - Can you please review this. Thanks.

test/functional/JLM_Tests/src/org/openj9/test/java/lang/management/TestOpenJ9DiagnosticsMXBean.java
```diff
@@ -69,6 +71,7 @@
 	private JMXConnector connector = null;
 	private static ProcessLocking lock;
 	private static String tmpFileName;
+	private static boolean zOS = "z/OS".equalsIgnoreCase(os) ? true : false;
```
keithc-ca: The field should be `final`.

Why the use of the ternary operator? Why not just `"z/OS".equalsIgnoreCase(os)`?

test/functional/JLM_Tests/src/org/openj9/test/java/lang/management/TestOpenJ9DiagnosticsMXBean.java
```diff
@@ -414,15 +420,25 @@ private void triggerDumpToFile(OpenJ9DiagnosticsMXBean diagBean, String test) th
...
 				File dumpFile = new File(fileName);
-
 				if (!(agent.equals("stack"))) {
 					if (isZOSSystemAgent(agent)) {
-						/* checking for dataset presence will be included later */
+						String actual_dsn = "";
+						logger.info("Dataset name returned by triggerDumpToFile = " + fileName);
+						if (fileName.endsWith("X&DS")) {
+							logger.info("Replacing X&DS with X001 in system dump dataset name");
+							actual_dsn = fileName.replace("X&DS", "X001");
```
keithc-ca: This should replace the suffix not just any occurrence of `X&DS`.
The first argument to `replace` is a regular expression which might misinterpret `&`.

test/functional/JLM_Tests/src/org/openj9/test/java/lang/management/TestOpenJ9DiagnosticsMXBean.java
```diff
@@ -435,6 +451,7 @@ private void triggerDumpToFile(OpenJ9DiagnosticsMXBean diagBean, String test) th
 		}
 	}
 
+
```
keithc-ca: Why do we need more blank lines?

test/functional/JLM_Tests/src/org/openj9/test/java/lang/management/TestOpenJ9DiagnosticsMXBean.java
```diff
@@ -488,14 +505,37 @@ private static boolean findAndDeleteFile(String dirPath, String fileName) throws
...
+			found = (boolean) zfile_exists.invoke(null, quoted_dsn);
+			if (found) {
+				Method zfile_remove = zfileClass.getMethod("remove", String.class);
+				zfile_remove.invoke(null, quoted_dsn);
+			}
+		}
+		return found;
+	}
+
+        /**
```
keithc-ca: Please restore the indentation of this line.

test/functional/JLM_Tests/src/org/openj9/test/java/lang/management/TestOpenJ9DiagnosticsMXBean.java
```diff
@@ -219,7 +222,9 @@ private void testLocal_triggerDump() throws FileNotFoundException, IOException {
 	 *
 	 */	
 	@Test
-	private void testLocal_triggerDumpToFile() throws InvalidOptionException, FileNotFoundException, IOException {
+	private void testLocal_triggerDumpToFile() throws InvalidOptionException, FileNotFoundException, IOException, ClassNotFoundException, NoSuchMethodException,
+										IllegalAccessException, InvocationTargetException {
```
keithc-ca: The list of exceptions is rather long: is there value above just saying `throws Exception`?

test/functional/JLM_Tests/src/org/openj9/test/java/lang/management/TestOpenJ9DiagnosticsMXBean.java
```diff
@@ -488,14 +505,37 @@ private static boolean findAndDeleteFile(String dirPath, String fileName) throws
...
 	/**
+	 * Function to check if the specified system dump dataset is present
+	 *
+	 * @param actual_dsn System dump data set name
+	 * @return a boolean indicating if the system dump data set was found or not.
+	 */
+	private static boolean findAndDeleteDataset(String actual_dsn) throws ClassNotFoundException, NoSuchMethodException, IllegalAccessException, InvocationTargetException {
+		boolean found = false;
+		if (zOS) {
+			Class<?> zfileClass = Class.forName("com.ibm.jzos.ZFile");
```
keithc-ca: Won't this just (always) fail with ClassNotFoundException? As far as I'm aware, we don't yet have builds which include the jzos support.

test/functional/JLM_Tests/src/org/openj9/test/java/lang/management/TestOpenJ9DiagnosticsMXBean.java
chandrams: I don't see the dumps getting created are always mult-ipart or being suffixed with X&DS, hence included a check to detect if fileName ends with "X&DS" and if so replace with "X001"

http://vmfarm.ottawa.ibm.com/job_output.php?id=547308220
[TestOpenJ9DiagnosticsMXBean] [INFO] Dataset name returned by triggerDumpToFile = J9BUILD.MYJVM.J9BUILD1.D180530.T030804
[TestOpenJ9DiagnosticsMXBean] [INFO] Dataset name does not contain X&DS

http://vmfarm.ottawa.ibm.com/job_output.php?id=547307586
[TestOpenJ9DiagnosticsMXBean] [INFO] Dataset name returned by triggerDumpToFile = J9BUILD.MYJVM.J9BUILD5.D180530.T040442.X&DS
[TestOpenJ9DiagnosticsMXBean] [INFO] Replacing X&DS with X001 in system dump dataset name

keithc-ca: My point is that the condition uses `endsWith()` so the adjustment should only replace an occurrence of that string at the *end*.

chandrams:
@keithc-ca - Thanks for clarifying, I have made the changes as suggested. Please review again.

test/functional/JLM_Tests/src/org/openj9/test/java/lang/management/TestOpenJ9DiagnosticsMXBean.java
chandrams: I don't see a zOS openjdk8-openj9 build on adopt as of now, will it be made available?

pshipton: Not any time soon. Perhaps in the future, still tbd.

chandrams: Thanks Peter.  Another option I can think of, is to copy the dataset to unix file system and verify using the File API instead of jzos APIs.  Would this be a better solution?

pshipton: Not sure there is any concern. IBM Java 8 builds will have com.ibm.jzos.ZFile. The test could be excluded for any other z/OS build until such time as com.ibm.jzos.ZFile is available. An alternative is to print a warning when the class cannot be found, and pass the core dump portion of the testing.

genie-openj9:

PR #2026, opened 2018-05-30
Catch failures in cleanWs stage of the job

jdekonin:
* [skip ci]
* occasionally there are failure to cleanup the workspace due to ??
* adding the try/catch causes the job to pass if only failing in cleanWs
    as well as displaying 'ps' output at the time of failing to cleanWs
* remove windows specific handling in the catch for the test/cleanWs case
    as 'ps -ef' works for windows via jnlp or ssh jenkins connection

### DIFF
--- a/buildenv/jenkins/common/build
+++ b/buildenv/jenkins/common/build
@@ -259,7 +259,13 @@ def build_all() {
                 archive()
             }
         } finally {
-            cleanWs()
+            try {
+		cleanWs()
+	    } catch(e) {
+		echo e.toString()
+		echo e.printStackTrace()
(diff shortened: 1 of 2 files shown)


buildenv/jenkins/common/build
```diff
@@ -259,7 +259,12 @@ def build_all() {
                 archive()
             }
         } finally {
-            cleanWs()
+            try {
+		cleanWs()
+	    } catch(e) {
+		echo e.toString()
```
pshipton: Might be more useful to print the entire stack trace rather than just the exception message

PR #2035, opened 2018-05-31
Implement and enable fast JNI VM access enter and exit

NigelYiboYu:
Implement and enable atomic-free sequence for JNI VM access release
and acquire on z platforms. Use this to replace the atomic
compare and swap instruction.


NigelYiboYu:
@gacholio @fjeremic 
This is the z codegen change for the atomic-free JNI feature. It's asymmetric on the zLinux (just like x86 linux) and symmetric on zOS. zOS will remain so until a proper flush function can be found.

Shall I enabled them by default?

NigelYiboYu:
Marking this pull request as WIP until it's thoroughly tested.

runtime/compiler/z/codegen/J9S390PrivateLinkage.hpp
```diff
@@ -165,6 +165,15 @@ class J9S390JNILinkage : public TR::S390PrivateLinkage
...
 
+#ifdef J9VM_INTERP_ATOMIC_FREE_JNI
+   void releaseVMAccessMaskAtomicFree(TR::Node * callNode,
+                                      TR::Register * methodMetaDataVirtualRegister,
+                                      TR::Register * tempReg1);
+
+   void acquireVMAccessMaskAtomicFree(TR::Node * callNode,
+                                      TR::Register * methodMetaDataVirtualRegister,
+                                      TR::Register * tempReg1);
+#endif /* J9VM_INTERP_ATOMIC_FREE_JNI */
```
fjeremic: Could we please document the newly added APIs?

runtime/compiler/z/codegen/J9S390PrivateLinkage.cpp
```diff
@@ -3051,6 +3051,97 @@ void TR::J9S390JNILinkage::acquireVMAccessMask(TR::Node * callNode, TR::Register
...
+                         generateS390MemoryReference(methodMetaDataVirtualRegister, offsetof(J9VMThread, inNative), cg));
+
+#if !defined(J9VM_INTERP_ATOMIC_FREE_JNI_USES_FLUSH)
+   // The asymmetric algorithm does not need this serialization.
+   TR::InstOpCode::S390BranchCondition cond = TR::InstOpCode::COND_MASK14;
+   if (!cg->getS390ProcessorInfo()->supportsArch(TR_S390ProcessorInfo::TR_z196))
+      cond = TR::InstOpCode::COND_MASK15;
+
+   TR::RealRegister *gpr0 = cg->machine()->getS390RealRegister(TR::RealRegister::GPR0);
+   new (cg->trHeapMemory()) TR::S390RegInstruction(TR::InstOpCode::BCR, callNode, cond, gpr0, cg);
```
fjeremic: Can we write a new `generate*` helper for this one?

runtime/compiler/z/codegen/J9S390PrivateLinkage.cpp
```diff
@@ -3051,6 +3051,97 @@ void TR::J9S390JNILinkage::acquireVMAccessMask(TR::Node * callNode, TR::Register
...
+#endif
+
+   // Compare vmthread public flag with J9_PUBLIC_FLAGS_VM_ACCESS
+   generateRXInstruction(cg, TR::InstOpCode::getLoadOpCode(), callNode, tempReg1,
+                         generateS390MemoryReference(methodMetaDataVirtualRegister, fej9->thisThreadGetPublicFlagsOffset(), cg));
+
+   TR::LabelSymbol * longReleaseSnippetLabel = generateLabelSymbol(cg);
+   TR::LabelSymbol * longReleaseRestartLabel = generateLabelSymbol(cg);
+
+   TR::InstOpCode::Mnemonic cmpImmAndJumpOp = TR::Compiler->target.is64Bit() ? TR::InstOpCode::CGIJ : TR::InstOpCode::CIJ;
```
fjeremic: Do we already have a helper for this in `TR::InstOpCode`? If not let's add one.

runtime/compiler/z/codegen/J9S390PrivateLinkage.cpp
```diff
@@ -3051,6 +3051,97 @@ void TR::J9S390JNILinkage::acquireVMAccessMask(TR::Node * callNode, TR::Register
...
+TR::J9S390JNILinkage::acquireVMAccessMaskAtomicFree(TR::Node * callNode,
+                                                    TR::Register * methodMetaDataVirtualRegister,
+                                                    TR::Register * tempReg1)
+   {
+   TR_J9VMBase *fej9 = (TR_J9VMBase *)fe();
+   TR::CodeGenerator* cg = self()->cg();
+   TR::Compilation* comp = self()->comp();
+
+   // Zero vmthread->inNative, which is a UDATA field
+   uint32_t pointerSize = TR::Compiler->target.is64Bit() ? 7 : 3;
```
fjeremic: I believe there is a helper for this either in `ObjectModel` or the base `CodeGenerator` class (for getting size of pointer). We should use that rather than magic constants.

runtime/compiler/z/codegen/J9S390PrivateLinkage.cpp
```diff
@@ -3051,6 +3051,97 @@ void TR::J9S390JNILinkage::acquireVMAccessMask(TR::Node * callNode, TR::Register
...
+                                                    TR::Register * tempReg1)
+   {
+   TR_J9VMBase *fej9 = (TR_J9VMBase *)fe();
+   TR::CodeGenerator* cg = self()->cg();
+   TR::Compilation* comp = self()->comp();
+
+   // Store 1 into vmthread->inNative
+   generateRIInstruction(cg, TR::InstOpCode::getLoadHalfWordImmOpCode(), callNode, tempReg1, 1);
+   generateRXInstruction(cg, TR::InstOpCode::getStoreOpCode(), callNode, tempReg1,
+                         generateS390MemoryReference(methodMetaDataVirtualRegister, offsetof(J9VMThread, inNative), cg));
```
fjeremic: Can this be done with an `MVI`?

gacholio:
I haven't looked at all the code, but if the VM access flag is out of the immediate range, I'm asserting on PPC (and soon on X86).  I'd rather see an error than a silent increase in the size of the codegen for direct JNI.

https://github.com/eclipse/openj9/pull/2039

gacholio:
I'll enable the build spec flags internally (like X and P) once this has been committed.  You guys can decide if you want ZOS enabled with the barrier or not - I suspect it'll still be faster than the atomic solution.

gacholio:
Please remove the buildspec changes from this PR.

runtime/compiler/z/codegen/J9S390PrivateLinkage.cpp
NigelYiboYu: There's no such API for cmpImmAndJump.. Will add one.

fjeremic:
> I haven't looked at all the code, but if the VM access flag is out of the immediate range, I'm asserting on PPC (and soon on X86). I'd rather see an error than a silent increase in the size of the codegen for direct JNI.

https://github.com/eclipse/openj9/pull/2039#discussion_r192297564

@NigelYiboYu let's do the same on Z.

runtime/compiler/z/codegen/J9S390PrivateLinkage.cpp
NigelYiboYu: Fixed. I should've searched a bit more carefully. We already have a helper for this.

runtime/compiler/z/codegen/J9S390PrivateLinkage.cpp
NigelYiboYu: Changed to `sizeof(UDATA)-1`.

fjeremic: I think what you want is `TR::Compiler->om.sizeofReferenceAddress() - 1`.

runtime/compiler/z/codegen/J9S390PrivateLinkage.cpp
NigelYiboYu: Changed to use `MVGHI` or `MVHI` depending on bitness (via the `getMoveHalfWordImmOpCode` query).

runtime/compiler/z/codegen/J9S390PrivateLinkage.hpp
NigelYiboYu: Done

NigelYiboYu:
Updated this pull request and it's now dependent on OMR PR https://github.com/eclipse/omr/pull/2620/files

runtime/compiler/z/codegen/J9S390PrivateLinkage.cpp
NigelYiboYu: OK

NigelYiboYu:
This change should be ready for Linux-on-z.

fjeremic:
Jenkins test all zlinux,zos

fjeremic:
Actually this is guarded by an ifdef that is not even activated so one job passing should be enough.

PR #2038, opened 2018-05-31
Fix JIT/AOT status in Java version string

fengxue-IS:
- define the default value of JIT/AOT status before checking JITConfig


pshipton:
The problem was that -verison showed `(JIT abled, AOT abled)` when running with -Xint or equivalent.

pshipton:
jenkins compile plinux jdk8

PR #2039, opened 2018-05-31
Atomic-free JNI work

gacholio:
Fix releaseVMAccess on PPC to properly use compare instead of and.
Assert if the VM access bit doesn't fit into an immediate to prevent a
constant change from unknowingly bloating the direct JNI path.

Also add the immediate value assertion on AMD64.

### DIFF
--- a/runtime/compiler/p/codegen/PPCJNILinkage.cpp
+++ b/runtime/compiler/p/codegen/PPCJNILinkage.cpp
@@ -728,18 +728,11 @@ void TR::PPCJNILinkage::releaseVMAccessAtomicFree(TR::Node* callNode, TR::Regist
    generateTrg1ImmInstruction(cg(), TR::InstOpCode::li, callNode, tempReg1, 1);
    generateMemSrc1Instruction(cg(), TR::InstOpCode::Op_st, callNode, new (trHeapMemory()) TR::MemoryReference(metaReg, (int32_t)offsetof(struct J9VMThread, inNative), TR::Compiler->om.sizeofReferenceAddress(), cg()), tempReg1);
 #if !defined(J9VM_INTERP_ATOMIC_FREE_JNI_USES_FLUSH)
-	generateInstruction(cg(), TR::InstOpCode::sync, callNode); // Necessary?
+	generateInstruction(cg(), TR::InstOpCode::sync, callNode);
 #endif /* !J9VM_INTERP_ATOMIC_FREE_JNI_USES_FLUSH */
    generateTrg1MemInstruction(cg(), TR::InstOpCode::Op_load, callNode, tempReg1, new (trHeapMemory()) TR::MemoryReference(metaReg, fej9->thisThreadGetPublicFlagsOffset(), TR::Compiler->om.sizeofReferenceAddress(), cg()));
-   if (J9_PUBLIC_FLAGS_VM_ACCESS >= LOWER_IMMED && J9_PUBLIC_FLAGS_VM_ACCESS <= UPPER_IMMED)
-      generateTrg1Src1ImmInstruction(cg(), TR::InstOpCode::andi_r, callNode, tempReg2, tempReg1, cr0Reg, J9_PUBLIC_FLAGS_VM_ACCESS);
-   else
-      {
-      loadConstant(cg(), callNode, J9_PUBLIC_FLAGS_VM_ACCESS, tempReg2);
-      // FIXME: Apparently I'm the first one to ever use a Trg1Src2 record-form instruction...
-      // ctor for Trg1Src2 + condReg only has the preced form, if you don't pass a preceding instruction the following instruction will end up as the first instruction after the prologue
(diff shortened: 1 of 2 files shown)


gacholio:
I've examined the 32-bit and 64-bit code, and they are correct for both entering and exitting the VM.

runtime/compiler/p/codegen/PPCJNILinkage.cpp
```diff
@@ -728,18 +728,11 @@ void TR::PPCJNILinkage::releaseVMAccessAtomicFree(TR::Node* callNode, TR::Regist
...
-   if (J9_PUBLIC_FLAGS_VM_ACCESS >= LOWER_IMMED && J9_PUBLIC_FLAGS_VM_ACCESS <= UPPER_IMMED)
-      generateTrg1Src1ImmInstruction(cg(), TR::InstOpCode::andi_r, callNode, tempReg2, tempReg1, cr0Reg, J9_PUBLIC_FLAGS_VM_ACCESS);
-   else
-      {
-      loadConstant(cg(), callNode, J9_PUBLIC_FLAGS_VM_ACCESS, tempReg2);
-      // FIXME: Apparently I'm the first one to ever use a Trg1Src2 record-form instruction...
-      // ctor for Trg1Src2 + condReg only has the preced form, if you don't pass a preceding instruction the following instruction will end up as the first instruction after the prologue
-      generateTrg1Src2Instruction(cg(), TR::InstOpCode::and_r, callNode, tempReg2, tempReg1, tempReg2, cr0Reg, /* FIXME */ cg()->getAppendInstruction());
-      }
+   TR_ASSERT(J9_PUBLIC_FLAGS_VM_ACCESS >= LOWER_IMMED && J9_PUBLIC_FLAGS_VM_ACCESS <= UPPER_IMMED, "VM access bit must be immediate");
```
fjeremic: This assert will not fire unless we are in a debug build. If you want something that runs in production use `TR_ASSERT_FATAL`.

gacholio: Which is preferred here do you think?  If the value gets out of range, we'll encode a bogus instruction and likely crash, at which point a debug build would show the problem.  On the other hand, these two asserts aren't likely to have any effect on codegen performance.

fjeremic: I would think we want a `TR_ASSERT_FATAL` here. I'd rather us fail the compilation right away with a full backtrace than almost surely crash at runtime, or yet worse get undefined behavior. We'll make the same change on Z.

gacholio: Sounds good - I'll make the change (and add it to X where I'm already assuming it, but not asserting it).

gacholio:
@fjeremic Ready for re-review.  If you're comfortable merging this, I'll remove Julian.

fjeremic:
Jenkins sanity xlinux, plinux, aix, win

fjeremic:
Jenkins test sanity xlinux, plinux, aix, win

fjeremic:
Jenkins test sanity xlinux,plinux,aix,win

fjeremic:
Windows builds failed due to infrastructure issues on the machine:

```
```

runtime/compiler/p/codegen/PPCJNILinkage.cpp
```diff
@@ -728,18 +728,11 @@ void TR::PPCJNILinkage::releaseVMAccessAtomicFree(TR::Node* callNode, TR::Regist
...
    generateTrg1MemInstruction(cg(), TR::InstOpCode::Op_load, callNode, tempReg1, new (trHeapMemory()) TR::MemoryReference(metaReg, fej9->thisThreadGetPublicFlagsOffset(), TR::Compiler->om.sizeofReferenceAddress(), cg()));
-   if (J9_PUBLIC_FLAGS_VM_ACCESS >= LOWER_IMMED && J9_PUBLIC_FLAGS_VM_ACCESS <= UPPER_IMMED)
-      generateTrg1Src1ImmInstruction(cg(), TR::InstOpCode::andi_r, callNode, tempReg2, tempReg1, cr0Reg, J9_PUBLIC_FLAGS_VM_ACCESS);
-   else
-      {
-      loadConstant(cg(), callNode, J9_PUBLIC_FLAGS_VM_ACCESS, tempReg2);
-      // FIXME: Apparently I'm the first one to ever use a Trg1Src2 record-form instruction...
-      // ctor for Trg1Src2 + condReg only has the preced form, if you don't pass a preceding instruction the following instruction will end up as the first instruction after the prologue
-      generateTrg1Src2Instruction(cg(), TR::InstOpCode::and_r, callNode, tempReg2, tempReg1, tempReg2, cr0Reg, /* FIXME */ cg()->getAppendInstruction());
-      }
```
zl-wang: Although not applicable anymore,  andis_r instruction can be used for the case where J9_PUBLIC_FLAGS_VM_ACCESS falls in the upper 16-bit.

gacholio: On the off chance we ever feel the need to change the bit value, we can re-assess.  It didn't seem like a good idea to maintain a bunch of dead code in these functions.  I think every platform will be able to represent 0x20 as an immediate today.

PR #2044, opened 2018-05-31
Build JNI IFA switching helpers as C calls

NigelYiboYu:
The JNI IFA switching on and off helpers use transition assembly sequences
that saves and restores all GPRs, all FPRs and volatile VRFs. This is slow
and unnecessary.

This commit changes builds IFA calls as C function calls, which saves only
the assigned registers.


NigelYiboYu:
@fjeremic  JNI IFA glue code removal commit here. 
Marking this as WIP until more testing is done.

runtime/compiler/build/files/target/z.mk
```diff
@@ -66,6 +66,7 @@ JIT_PRODUCT_SOURCE_FILES+=\
     compiler/z/codegen/J9MemoryReference.cpp \
     compiler/z/codegen/J9S390CHelperLinkage.cpp \
     compiler/z/codegen/J9S390PrivateLinkage.cpp \
+    compiler/z/codegen/J9S390JNILinkage.cpp \
```
fjeremic: I think we'll need to modify the CMake builds as well.

runtime/compiler/runtime/Runtime.cpp
```diff
@@ -616,8 +616,8 @@ JIT_HELPER(_nativeStaticHelper);
 JIT_HELPER(jitLookupInterfaceMethod);
 JIT_HELPER(jitMethodIsNative);
 JIT_HELPER(jitMethodIsSync);
-JIT_HELPER(jitPreJNICallOffloadCheck);
-JIT_HELPER(jitPostJNICallOffloadCheck);
+JIT_HELPER(fast_jitPreJNICallOffloadCheck);
+JIT_HELPER(fast_jitPostJNICallOffloadCheck);
```
fjeremic: Is there a reason for deviation from the naming convention?

runtime/compiler/trj9/z/codegen/J9S390JNILinkage.hpp
```diff
@@ -0,0 +1,70 @@
+/*******************************************************************************
+ * Copyright (c) 2000, 2018 IBM Corp. and others
```
fjeremic: Copyright needs to be updated to be 2018, 2018. Similarly for the other file.

runtime/compiler/trj9/z/codegen/J9S390JNILinkage.hpp
```diff
@@ -0,0 +1,70 @@
...
+
+#include "trj9/z/codegen/J9S390PrivateLinkage.hpp"
+
+namespace TR { class CodeGenerator; }
+namespace TR { class RegisterDependencyConditions; }
+namespace TR { class Snippet; }
+
+namespace TR {
+
+class J9S390JNILinkage : public TR::S390PrivateLinkage
```
fjeremic: We need to Doxygen document this entire class.

runtime/compiler/build/files/target/z.mk
NigelYiboYu: I don't see CMake lists contain these `cpp` file names. It operates on directories.

runtime/compiler/trj9/z/codegen/J9S390JNILinkage.hpp
NigelYiboYu: Fixed

runtime/compiler/runtime/Runtime.cpp
NigelYiboYu: I changed this to use the same name as the VM helper name for consistency. Didn't realize that the JIT helpers have a naming convention. I'll change it back.

NigelYiboYu: I'm keeping the `fast_` in the names because these two lines are effectively `extern c fast_*` and referring to the VM helpers.

NigelYiboYu:
The changes are tested OK. Ready for reviews.

NigelYiboYu:
To illustrate the differences before and after this change, here's the trace files from a JNI benchmark where the same native function is invoked in a loop for millions of times. 

Full trace files:
[JNI-after.txt](https://github.com/eclipse/openj9/files/2094528/JNI-after.txt)
[JNI-before.txt](https://github.com/eclipse/openj9/files/2094529/JNI-before.txt)


Annotated excerpt for them:
[annotated JNI excerpt after the change.txt](https://github.com/eclipse/openj9/files/2094521/annotated.JNI.excerpt.after.the.change.txt)
[annotated JNI excerpt before the change.txt](https://github.com/eclipse/openj9/files/2094522/annotated.JNI.excerpt.before.the.change.txt)

runtime/compiler/z/codegen/J9S390CHelperLinkage.cpp
```diff
@@ -439,3 +459,537 @@ TR::Register * TR::S390CHelperLinkage::buildDirectDispatch(TR::Node * callNode,
...
+                               getS390RealRegister(TR::RealRegister::HPR15),
+                               generateS390MemoryReference(systemStackPointer, hprSaveBias, cg()));
+
+      saveFPRsOrVRFs(callNode, systemStackPointer, vmThreadRegister);
+      }
+   }
+#endif
+
+TR::Register *
+TR::S390CHelperLinkage::buildIndirectDispatch(TR::Node * callNode,
```
fjeremic: This function seems built very specifically for JNI IFA. Does it even belong here? There are also a handful of booleans below which can be folded away. I'm questioning the nature of the existence of this API. CHelper functions should all be statically addressable. Why exactly do we need indirect dispatch for them?

runtime/compiler/z/codegen/J9S390CHelperLinkage.cpp
```diff
@@ -439,3 +459,537 @@ TR::Register * TR::S390CHelperLinkage::buildDirectDispatch(TR::Node * callNode,
...
+                                                     offsetof(J9VMThread, javaVM), cg()));
+
+
+   // get jitConfig from JavaJVM
+   cursor = generateRXInstruction(cg(), TR::InstOpCode::getLoadOpCode(), callNode, nativeMethodEntryAddressReg,
+                         generateS390MemoryReference(nativeMethodEntryAddressReg,
+                                                     offsetof(J9JavaVM, jitConfig), cg()));
+
+
+   // get J9TR_JitConfig_fast_jitPreJNICallOffloadCheck from jit config
```
fjeremic: Isn't this a compile time constant? Why do we need these 3 indirections? i.e. how is this any different from `fast_jitInstanceOf`? Are we able to extract the address statically at compile time via:

```
callNode->getSymbolReference()->getSymbol()->castToMethodSymbol()->getMethodAddress();
```

runtime/compiler/z/codegen/J9S390JNILinkage.cpp
```diff
@@ -0,0 +1,927 @@
...
+   jniCallDataSnippet->setPC(constJNICallOutFrameType);
+   jniCallDataSnippet->setLiterals(0);
+   jniCallDataSnippet->setJitStackFrameFlags(0);
+
+   generateSS1Instruction(cg(), TR::InstOpCode::MVC, callNode, 3*(TR::Compiler->om.sizeofReferenceAddress()) - 1,
+                          generateS390MemoryReference(methodMetaDataVirtualRegister, fej9->thisThreadGetJavaPCOffset(), codeGen),
+                          generateS390MemoryReference(jniCallDataSnippet->getBaseRegister(), jniCallDataSnippet->getPCOffset(), codeGen));
+
+   // store out jsp
+   generateRXInstruction(codeGen, TR::InstOpCode::getStoreOpCode(), callNode, tempReg,
```
fjeremic: Can we rename `codeGen` to `cg` to stay consistent with the rest of the codebase?

runtime/compiler/z/codegen/J9S390CHelperLinkage.cpp
NigelYiboYu: Yes, they are compile time constants. The indirections are copied from the old glue code. 

I had difficulty building direct `BRASL` calls to `fast_jitPre(post)JNICallOffloadCheck`. VM changes might be needed for us to jump directly to `fast_jitPostJNICallOffloadCheck`.

runtime/compiler/z/codegen/J9S390CHelperLinkage.cpp
NigelYiboYu: #1 Admittedly this API is very customized towards the JNI IFA calls because IFA is the only consumer of this. Some of the booleans (`isFastPathOnly` and even the `#if defined(J9ZOS390)` for instance) can be folded away. I kept them there in case future users may need them. The IFA calls are zOS only but I figured that we can build indirect helper calls for other helpers on Linux.

#2 Also, the c function `fast_jitPost(pre)JNICallOffloadCheck` can't be addressed statically as per current design. Direct invocation will result in VM assert failures. I suspect all function pointers in the JITConfig structure are not directly invoke-able. By `directly invoke-able`, I mean BRASL jump to the `fast_jitPre(post)JNICallOffloadCheck` function pointers directly.

I didn't use `callNode->getSymbolReference()->getSymbol()->castToMethodSymbol()->getMethodAddress();` Will try and see how this is different from direct function pointer approach.

#3 One JNI IFA specific aspect is the register saving/restoring such as `saveRegistersForIFAHelper()`. This logic used to be baked inside the glue code is needed if CEEHDLR is enabled. CEE handlers can crash if sys SP and the vmthread registers are not saved on the native side before invoking the native. This has to be baked into the IFA call builder.

In conclusion:
#1 is simple to do. I can fold away the booleans.
For #2 we have to see why the C functions are not directly invoke-able using function pointers.
#3 has to be in the builder function, whether it be a CHelper function or a function in a different class.

runtime/compiler/z/codegen/J9S390JNILinkage.cpp
```diff
@@ -0,0 +1,927 @@
...
+   TR::Instruction * cursor = NULL;
+
+   int32_t stackAdjust = (-5 * (int32_t)sizeof(intptrj_t));
+
+   cursor = generateRXYInstruction(codeGen, TR::InstOpCode::LAY, callNode, tempReg,
+                                   generateS390MemoryReference(javaStackPointerRealRegister, stackAdjust, codeGen), cursor);
+
+   // set up Java Thread
+   intptrj_t constJNICallOutFrameType = fej9->constJNICallOutFrameType();
+   TR_ASSERT( constJNICallOutFrameType < MAX_IMMEDIATE_VAL, "OMR::Z::Linkage::setupJNICallOutFrame constJNICallOutFrameType is too big for MVHI");
```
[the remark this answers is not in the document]
NigelYiboYu: Yeah, this is from the old code and should be deleted.

runtime/compiler/z/codegen/J9S390JNILinkage.cpp
```diff
@@ -0,0 +1,927 @@
...
+  *   8/16 |-----|
+  *        |     |      <-- constJNICallOutFrameFlags()
+  *   4/8   -----
+  *        |     |      <-- ramMethod for the native method
+  *         -----       <-- stack pointer
+  *
+  * \endverbatim
+  */
+void
+TR::J9S390JNILinkage::setupJNICallOutFrame(TR::Node * callNode,
```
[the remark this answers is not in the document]
NigelYiboYu: Will cleanup the `S390JNICallDataSnippet` and consolidate the code.

NigelYiboYu:
@fjeremic This PR is ready.

runtime/compiler/z/codegen/J9S390CHelperLinkage.cpp
```diff
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2000, 2018 IBM Corp. and others
+ * Copyright (c) 2017, 2018 IBM Corp. and others
```
fjeremic: Just leaving a note that this is a valid change as the file was introduced in 2017 originally, not 2000.

runtime/compiler/z/codegen/J9S390JNILinkage.cpp
```diff
@@ -0,0 +1,964 @@
...
+                                          bool isJNICallOutFrame,
+                                          bool isPreJNIHelper)
+   {
+   TR_RuntimeHelper helperIndex = isPreJNIHelper ? TR_S390jitPreJNICallOffloadCheck : TR_S390jitPostJNICallOffloadCheck;
+   const char* symName  = isPreJNIHelper ? "fast_jitPreJNICallOffloadCheck" : "fast_jitPostJNICallOffloadCheck";
+   void* targetAddr     = isPreJNIHelper ? (void*)fast_jitPreJNICallOffloadCheck : (void*)fast_jitPostJNICallOffloadCheck;
+
+   TR::SymbolReference * offloadSymRef = cg()->symRefTab()->findOrCreateRuntimeHelper(helperIndex, false, false, false);
+   offloadSymRef->getSymbol()->setName(symName);
+   offloadSymRef->getSymbol()->castToMethodSymbol()->setMethodAddress(targetAddr);
```
fjeremic: I don't think we need this code as this is already handled when the helper is created here:

https://github.com/eclipse/omr/blob/70ff3c7d2a7fa2f85f380dadf23e5acda65a45bc/compiler/compile/OMRSymbolReferenceTable.cpp#L714-L735

Are we able to remove this and simply pass in the `TR_RuntimeHelper` index instead of `bool isPreJNIHelper`? We can also remove the above two `extern "C"` definitions and simplify this function greatly. In fact it may become so small as to not even warrant an existence?

NigelYiboYu:
After refactoring the `J9S390JNILinkage` and `J9S390CHelperLinkage` classes, I still need to introduce two things to `CHelper`: 
1. the stack adjustment for JNI IFA only, and 

2 the loading of an environment.  I believe the environment should be part of the CHelper because the XPLINK has it as an `optional` thing in GPR5 which is probably not used often. In the case of JNI IFA helper call, the environment is needed for proper functioning.

runtime/compiler/z/codegen/J9S390CHelperLinkage.cpp
```diff
@@ -364,6 +373,26 @@ TR::Register * TR::S390CHelperLinkage::buildDirectDispatch(TR::Node * callNode,
...
+                                         javaStackPointerRegister, static_cast<int32_t>(environment));
+         }
+      else
+         {
+         int32_t envHi = static_cast<int32_t>((environment >> 32) & 0x00000000FFFFFFFFull);
+         int32_t envLo = static_cast<int32_t>(environment & 0x00000000FFFFFFFFull);
+         cursor = generateRILInstruction(cg(), TR::InstOpCode::IIHF, callNode, javaStackPointerRegister, envHi);
+         cursor = generateRILInstruction(cg(), TR::InstOpCode::IILF, callNode, javaStackPointerRegister, envLo);
+         }
+      }
```
fjeremic: This whole sequence should simply use `genLoadAddressConstant` and pass the `environment` constant through. The function will figure out the best sequence to generate.

runtime/compiler/z/codegen/J9S390JNILinkage.cpp
```diff
@@ -0,0 +1,1032 @@
...
+   TR::Instruction* end =generateS390LabelInstruction(cg(), TR::InstOpCode::LABEL, callNode, tempLabel);
+   end->setEndInternalControlFlow();
+   end->setDependencyConditions(postDeps);
+   }
+
+extern "C" void fast_jitPreJNICallOffloadCheck();
+extern "C" void fast_jitPostJNICallOffloadCheck();
+
+void
+TR::J9S390JNILinkage::callIFAOffloadCheck(bool isPreJNIHelper, bool isJNICallOutFrame)
```
fjeremic: We should fold this function away. I don't think it's needed. The environment is already present via `TR_S390CEnvironmentAddress` so there is no need to deal with the function pointers etc. which renders this function an effective three liner:

```
TR::SymbolReference * offloadSymRef = cg()->symRefTab()->findOrCreateRuntimeHelper(helperIndex, false, false, false);
TR::Node* helperCallNode = TR::Node::createWithSymRef(TR::call, 0, offloadSymRef);
cHelperLinkage->buildDirectDispatch(helperCallNode, NULL, NULL, true,
                                    isJNICallOutFrame,
                                    reinterpret_cast<intptrj_t>(TR_S390CEnvironmentAddress));
```

I'm not sure it has a need for existence.

runtime/compiler/z/codegen/J9S390CHelperLinkage.cpp
r30shah: @NigelYiboYu Loading environment register is necessary for calling a C function from JIT code. I omitted that part from Linkage as Assembly listing for `fast_jitInstanceOf` was using it as volatile and we overwrite the register so anyway loaded value was not used. I checked other dual mode helper, Some of them returns the function pointer of the slow path helper function which belongs to same library and in that scenario environment register is needed to get the function pointer which is currently taken care by assembly glue in `znathelp.m4`.  That is the only use I could find currently. \
Correct change here would be respecting XPLINK linkage and also pass environment register for instanceOf as well. Means before calling fastPath helper, environment register should be loaded.

runtime/compiler/z/codegen/J9S390JNILinkage.cpp
NigelYiboYu: Removed the `extern` statements and used `TR_RuntimeHelper` indices instead.

runtime/compiler/z/codegen/J9S390JNILinkage.cpp
NigelYiboYu: Changed to `cg()`

NigelYiboYu:
@gacholio Hello GAC, this JNI IFA helper call change involves changes on the VM `beforeJNI()` and `afterJNI()` calls. The idea is to change the IFA switching functions to accept a second parameter `J9Method*` and use this parameter instead of reading a `J9Method*` off the hidden JNI callout frame. 

The JIT generated code does not assume changing java stack pointers. Having a callout frame and the helpers reading from it means the JIT'ed code has to expose this frame prior to IFA offload call, and "hide" this frame after it re-enters JIT'ed code. This requires several  JSP adjustment in the JIT'ed code and can hurt performance.

runtime/codert_vm/cnathelp.cpp
```diff
@@ -2189,12 +2189,12 @@ fast_jitPreJNICallOffloadCheck(J9VMThread *currentThread)
 		els->calloutVMState = setVMState(currentThread, J9VMSTATE_JNI_FROM_JIT);
 	}
 #endif /* J9VM_PORT_ZOS_CEEHDLRSUPPORT */
-	VM_VMHelpers::beforeJNICall(currentThread);
+    VM_VMHelpers::beforeJNICall(currentThread, method);
```
gacholio: Bad spacing introduced here.

runtime/codert_vm/cnathelp.cpp
```diff
@@ -2209,7 +2209,7 @@ fast_jitPostJNICallOffloadCheck(J9VMThread *currentThread)
 		}
 	}
 #endif /* J9VM_PORT_ZOS_CEEHDLRSUPPORT */
-	VM_VMHelpers::afterJNICall(currentThread);
+    VM_VMHelpers::afterJNICall(currentThread, method);
```
gacholio: and here

runtime/compiler/runtime/Runtime.cpp
```diff
@@ -623,8 +623,8 @@ JIT_HELPER(_nativeStaticHelper);
...
```
gacholio: Do not declare these as JIT_HELPER.  They aren't assembly functions.  They should be externed as proper C functions.

runtime/vm/BytecodeInterpreter.hpp
```diff
@@ -2170,13 +2172,14 @@ done:;
...
-	callCFunction(REGISTER_ARGS_LIST, void * jniMethodStartAddress, void *receiverAddress, UDATA *javaArgs, UDATA **bp, bool isStatic, U_8 *returnType)
+    callCFunction(REGISTER_ARGS_LIST, void * jniMethodStartAddress, void *receiverAddress, UDATA *javaArgs, UDATA **bp, bool isStatic, U_8 *returnType,
+                  J9Method* method)
 	{
 		/* Release VM access (all object pointers are indirect referenced from here on) */
 		UDATA relativeBP = _arg0EA - *bp;
 		updateVMStruct(REGISTER_ARGS);
 		VM_VMAccess::inlineExitVMToJNI(_currentThread);
-		VM_VMHelpers::beforeJNICall(_currentThread);
+        VM_VMHelpers::beforeJNICall(_currentThread, method);
```
gacholio: spacing.

runtime/vm/BytecodeInterpreter.hpp
```diff
@@ -2194,7 +2197,7 @@ done:;
 			}
 		}
 #endif /* J9VM_PORT_ZOS_CEEHDLRSUPPORT */
-		VM_VMHelpers::afterJNICall(_currentThread);
+        VM_VMHelpers::afterJNICall(_currentThread, method);
```
gacholio: spacing

gacholio:
I have to wonder in general how this could really matter.  If the IFA switch is performed on entry and exit, doesn't that vastly overshadow any possible changes in the calling sequence?

NigelYiboYu:
Spacing, line-breaking and stale code issues have been fixed.


> doesn't that vastly overshadow any possible changes in the calling sequence?

IFA calls takes up about 80% of the JNI overhead, and indeed overshadows other JNI overhead. But the save-all and restore-all assembly sequence takes about 10% of the overhead. 10% is somewhat significant and every bit of optimization helps. By replacing this assembly glue code with a C function call, we can at least recover some perf.

runtime/oti/VMHelpers.hpp
```diff
@@ -1133,7 +1133,7 @@ class VM_VMHelpers
 	 * @param[in] currentThread the current J9VMThread
 	 */
 	static VMINLINE void
-	beforeJNICall(J9VMThread *currentThread)
+	beforeJNICall(J9VMThread *currentThread, J9Method* method)
```
fjeremic: Can we update the documentation for this newly added parameter?

runtime/compiler/z/codegen/J9S390CHelperLinkage.cpp
```diff
@@ -266,14 +266,19 @@ class RealRegisterManager
...
+TR::Register * TR::S390CHelperLinkage::buildDirectDispatch(TR::Node * callNode,
+                                                           TR::RegisterDependencyConditions **deps,
+                                                           TR::Register *returnReg,
+                                                           bool forceFastPath,
+                                                           TR_RuntimeHelper helperIndex)
    {
    RealRegisterManager RealRegisters(cg());
    bool isHelperCallWithinICF = deps != NULL;
-   // TODO: Currently only jitInstanceOf is fast path helper. Need to modify following condition if we add support for other fast path only helpers
-   bool isFastPathOnly = callNode->getOpCodeValue() == TR::instanceof;
```
fjeremic: Could we leave this TODO around? This is definitely something we still want to do once we refactor the runtime helpers into a sane table with properties.

runtime/compiler/z/codegen/J9S390CHelperLinkage.hpp
```diff
@@ -122,6 +122,10 @@ class S390CHelperLinkage : public TR::Linkage
       return buildDirectDispatch(callNode, NULL, returnReg);
       }
 
-   TR::Register *buildDirectDispatch(TR::Node *callNode, TR::RegisterDependencyConditions** deps, TR::Register *returnReg=NULL);
+   TR::Register *buildDirectDispatch(TR::Node *callNode,
+                                     TR::RegisterDependencyConditions** deps,
+                                     TR::Register *returnReg=NULL,
+                                     bool forceFastPath = false,
+                                     TR_RuntimeHelper helperIndex = TR_S390numRuntimeHelpers);
```
fjeremic: Could we move the documentation here such that it is with the definition of the API and document `foreFastPath` as well?

fjeremic:
@r30shah anything else you'd like to review on this one? I still see an "x" beside your review but GitHub can't seem to find what it was referencing.

runtime/compiler/z/codegen/J9S390CHelperLinkage.cpp
NigelYiboYu: TODO is added back

runtime/compiler/z/codegen/J9S390CHelperLinkage.hpp
NigelYiboYu: Documentation is moved to the `hpp` file with `forceFastPath` added.

runtime/oti/VMHelpers.hpp
NigelYiboYu: Added the new param to documentation.

fjeremic:
Jenkins test sanity

runtime/compiler/z/codegen/J9S390CHelperLinkage.cpp
```diff
@@ -364,6 +361,17 @@ TR::Register * TR::S390CHelperLinkage::buildDirectDispatch(TR::Node * callNode,
       }
    TR::SymbolReference * callSymRef = callNode->getSymbolReference();
    void * destAddr = callNode->getSymbolReference()->getSymbol()->castToMethodSymbol()->getMethodAddress();
+
+#if defined(J9ZOS390)
```
r30shah: @NigelYiboYu I think my previous comment was for this section. For now we need this part for calling out helpers written in C that means we need to load environment for instanceOf as well. Is that covered? Looking through your other changes. 
 I am going through your changes now.

r30shah: @fjeremic I am thinking of moving some of the context switching code from helper Glue code to in body and see the effect. I expect a small effect on the method footprint. If the effect is not much then we can move this inlined. I believe x86 already do this. We can make our helper linkage more generic this way.

runtime/compiler/z/codegen/J9S390CHelperLinkage.cpp
```diff
@@ -364,6 +361,17 @@ TR::Register * TR::S390CHelperLinkage::buildDirectDispatch(TR::Node * callNode,
...
+   // TR_S390numRuntimeHelpers represents invalid helper index
+   if (helperIndex != TR_S390numRuntimeHelpers)
```
r30shah: @NigelYiboYu I see that a declaration of this function defines `helperIndex` with `TR_S390numRuntimeHelpers`  and for JNI linkage calls (for pre/post offload check) , this function `(buildDirectDispatch)` will be called  with `TR_RuntimeHelper` equal to `TR_S390jitPreJNICallOffloadCheck` or `TR_S390jitPostJNICallOffloadCheck`. Is it correct?

NigelYiboYu: Yes, JNI IFA uses TR_S390jitPreJNICallOffloadCheck and TR_S390jitPostJNICallOffloadCheck.
For the lack of a better choice, I use `TR_S390numRuntimeHelpers` to validate the helper index parameter.

runtime/compiler/z/codegen/J9S390CHelperLinkage.cpp
NigelYiboYu: `InstanceOf` is not covered in this PR. This whole pull request is for JNI and IFA switching. We can come back and change `instanceOf` helper to load environment.

fjeremic: Actually good point @r30shah. Thinking about this again we have duplicate information here. The helper index is already stored in `callNode->getSymbolReference()->getReferenceNumber()` so we shouldn't need to pass it as an argument to the CHelper linkage `buildDirectDispatch`.

This information is already available through the `callNode`. @NigelYiboYu are we able to get rid of this extra parameter? Once this is done we can remove the below check for `if (helperIndex != TR_S390numRuntimeHelpers)` since a helper index must always be available. This will fix the `instanceOf` helper as well.

runtime/compiler/z/codegen/J9S390CHelperLinkage.cpp
r30shah: Okk. I think better choice here would be use of this [isFastPathOnly](https://github.com/NigelYiboYu/openj9/blob/5ee8d1852d8bf8a8608e74bdded9a510737119b2/runtime/compiler/z/codegen/J9S390CHelperLinkage.cpp#L271). I think we only set this when we are calling a C helper function directly from JIT'd code. This should also cover the case for `instanceOf` and load environment register for that.

runtime/compiler/z/codegen/J9S390CHelperLinkage.cpp
NigelYiboYu: The extra helper index is removed.

NigelYiboYu: Update:

Environment pointer is now only loaded if the C helper is fast-path. Non-fastPath helpers jump to glue code, which does its own environment handling.

fjeremic:
@r30shah could you please review again and see if all your concerns were addressed?

fjeremic:
Jenkins test sanity

gacholio:
@fjeremic - you really need to stop this shotgun approach to launching tests, especially considering that the changes in this PR don't affect any of the platforms in PR testing.

fjeremic:
> @fjeremic - you really need to stop this shotgun approach to launching tests, especially considering that the changes in this PR don't affect any of the platforms in PR testing.

Is there a reason you say this? This change modified cnathelp.cpp which gets compiled and runs on all platforms. I've been bitten before by not fully testing changes before merging. Irregardless though IMO the commiters should be running tests on any change before it is merged.

gacholio:
I say this because I've been waiting for a single machine compile for hours, no doubt due to too much stuff being launched. It most certainly is a resource issue as well, but that's unlikely to be fixed soon (if ever). Even the simplest change takes days to get in.

Also worth noting that running 8/9/10/11 testing is completely unnecessary on common code.

fjeremic:
> I say this because I've been waiting for a single machine compile for hours, no doubt due to too much stuff being launched. It most certainly is a resource issue as well, but that's unlikely to be fixed soon (if ever). Even the simplest change takes days to get in.
> 
> Also worth noting that running 8/9/10/11 testing is completely unnecessary on common code.

I see, so do you propose for this change running only JDK8 for example as the right way to go? I would still not feel comfortable merging a change which affects all platforms without running sanity on the said platforms.

I'm not sure what to say about the resource issue. We either need to have more granular tests or get more machines. I don't see a solution to this, and not running tests is not a solution that we should accept.

pshipton:
We are working on getting more machines for some platforms, by sending hardware to UNB and helping to set it up. Hopefully by Sept.

fjeremic:
@pshipton do we have guidelines for committers on how we should be launching testing? If we are unnecessarily stressing the farm too much we should come up with guidelines to minimize this.

gacholio:
Personally, I only ever run jdk8 with a single platform for things that are obviously common (which does occasionally result in problems on windows since the compiler is strict in ways no other one is).

My first proposal is that we eliminate the shotgun from jenkins entirely - one comment per build launch will make people consider what they're doing. It's just too easy to spam the world at the moment.

pshipton:
> do we have guidelines for committers on how we should be launching testing? If we are unnecessarily stressing the farm too much we should come up with guidelines to minimize this.

Not that I'm aware of. I think "test sanity" is the correct approach for JIT unless the change is specific to a platform, or generic across jdk versions. I expect JIT committers know best how to test JIT changes. Perhaps not all of xlinux, xlinux big heap, win and win-32 (jdk8 only) are always required since they are all x. Plus multiple jdk version may just duplicate testing depending on the change.

The automatic launching of the jdk11 builds is a mistake, since they don't work yet. I've opened #2539.

pshipton:
> My first proposal is that we eliminate the shotgun from jenkins entirely - one comment per build launch will make people consider what they're doing.

Works for me, at least for jdk versions. I've accidentally run multiple jdk versions when I didn't mean to.

fjeremic:
> My first proposal is that we eliminate the shotgun from jenkins entirely - one comment per build launch will make people consider what they're doing. It's just too easy to spam the world at the moment.

I agree this would be a step in the right direction.

fjeremic:
JDK11 builds are not supposed to be enabled, see #2539. Windows failures due to #2129.

PR #2055, opened 2018-06-01
Check for boolean array when running bastore bytecode in the JIT

fjeremic:
the array indexed by index"

This commit is the JIT portion of #2046.

### DIFF
--- a/runtime/compiler/ilgen/Walker.cpp
+++ b/runtime/compiler/ilgen/Walker.cpp
@@ -7339,6 +7339,14 @@ TR_J9ByteCodeIlGenerator::storeArrayElement(TR::DataType dataType, TR::ILOpCodes
    {
    TR::Node * value = pop();
 
+   if (dataType == TR::Int8)
+      {
+      // The JVM specification states:
+      // If the arrayref refers to an array whose components are of type boolean, then the int value is narrowed by 
+      // taking the bitwise AND of value and 1; the result is stored as the component of the array indexed by index.
+      value = TR::Node::create(TR::band, 2, value, TR::Node::createConstOne(value, dataType));
+      }
+
    handlePendingPushSaveSideEffects(value);
 
    bool genSpineChecks = comp()->requiresSpineChecks();


andrewcraik:
Jenkins test sanity

runtime/compiler/ilgen/Walker.cpp
```diff
@@ -7339,6 +7339,14 @@ TR_J9ByteCodeIlGenerator::storeArrayElement(TR::DataType dataType, TR::ILOpCodes
    {
    TR::Node * value = pop();
 
+   if (dataType == TR::Int8)
```
DanHeidinga: Is `TR::Int8` only used for boolean?

fjeremic: @DanHeidinga @andrewcraik as far as I can tell this is the only place where we pass `TR::Int8` to the `storeArrayElement` function:

https://github.com/eclipse/openj9/blob/cb042382f25e0c011944abff60bf2bbbbac226d1/runtime/compiler/ilgen/Walker.cpp#L1341

And it is the place that deals with the `bastore` bytecode. Is there something missing?

fjeremic: > Dan is right the Int8 could be a byte no?

You're right. I for some reason associated the 'b' with 'boolean', but it's byte! I don't see a way for us to disambiguate at compile time then between the byte and boolean array case. There is nothing on the symbol reference as far as I can see. @andrewcraik is there a way to do this in the walker?

fjeremic: Thinking about this more, although it may be possible in some cases in general it seems we have to check at runtime what the target arrayref of the store is to check for the boolean case. In which case I'm not sure whether the best place to do this is in the walker or at codegen time. I suppose we may be able to deduce the type of the array and eliminate a path if we do it in the walker so I'm leaning towards that.

andrewcraik: Yes and adding that kind of control flow in the walker is very non-trivial - it will probably have to be done as a post-pass. It will also have a nasty perf overhead if we have to check it all the time so we will have to try and aggressively propagate that type. I don't think this is quite as easy for the JIT as it is for the VM - we will need additional time to look at this I think.

PR #2061, opened 2018-06-04
Modify the translated PII files nls folder

panlei080988:
Check in the Shipment 1 translation package for J182F's TVT

Signed-off-by: Lei Pan <pleinb@cn.ibm.com>


pshipton:
About the copyright failures:

"the translated files are translated according to the English source file. In the English source file, the copyright is 2017, so according to the translating rules the translated files have to keep the same copyright content with English file. So translators cannot change the copyright to 2018 when in the English source file the year is 2017.

The rule is from the whole translation center for all the translation project. For one project they will not change the rule."

Issue #2065, opened 2018-06-04
Windows cmdLineTest_J9test sanity failure due to timeout testing -Xpreloaduser32 -Xprotectcontiguous


pshipton:
This just started occurring over the weekend. I think its intermittent but its blocking omr promotion.
Failures to date have occurred on jdk9 and jdk10.
I thought it was related to having processes hung on the Windows machines, waiting for a dialog, but Adam checked the machines and didn't find any dialogs.

Testing: -Xpreloaduser32 -Xprotectcontiguous
Test start time: 2018/06/04 08:52:44 Central Standard Time
Running command: "C:/Users/jenkins/workspace/Test-Sanity-JDK10-win_x86-64_cmprssptrs/jdk/bin\\java"  -Xcompressedrefs  -Xdump -cp "C:/Users/jenkins/workspace/Test-Sanity-JDK10-win_x86-64_cmprssptrs/openj9/test///..//jvmtest\functional\cmdLineTests\utils\utils.jar" -Xpreloaduser32 -Xprotectcontiguous VMBench.FibBench
Time spent starting: 4 milliseconds
***[TEST INFO 2018/06/04 08:55:44] ProcessKiller detected a timeout after 180000 milliseconds!***
INFO: getUnixPID() has failed indicating this is not a UNIX System.'Debug on timeout' is currently only supported on Linux.
Output from test:
Time spent executing: 2662155 milliseconds
Test result: FAILED
Output from test:
 [OUT] Fibonacci: iterations = 10000
 [OUT] fibonacci(12) = 144
 [OUT] done, time = 12

JasonFengJ9:
Investigating..

pshipton:
@JasonFengJ9 is looking at this.

pshipton:
Removed the blocker label since the test is now excluded by #2068

JasonFengJ9:
Initially the suspect code overrun was around `jvminit.c:preloadUser32Dll()` when `protectContiguous` is true. Instrument JVM doesn't show that code snippet takes longer than successful runs. It appears it is `library j9gc29` takes unusually longer to be loaded.
```
 [ERR] Running initialization stage ALL_LIBRARIES_LOADED
 [ERR] 	for library j9jit29...
 [ERR] 		completed with rc=0 in 1314064 nano sec.
 [ERR] 	for library j9bcv29...
 [ERR] 		completed with rc=0 in 6984 nano sec.
 [ERR] 	for library j9gc29...
[...]

PR #2068, opened 2018-06-04
Exclude Windows J9test -Xpreloaduser32 -Xprotectcontiguous

JasonFengJ9:
Exclude `Windows J9test -Xpreloaduser32 -Xprotectcontiguous`

Excluding `cmdLineTest_J9test subtest -Xpreloaduser32 -Xprotectcontiguous` for `Win64` platform as well.
This avoids intermittent golden failure and allow build promotion before a fix is identified.

Related to: #2065 

Reviewer: @pshipton 
FYI: @DanHeidinga 

### DIFF
--- a/test/functional/cmdLineTests/cmdLineTest_J9tests/j9tests_exclude.xml
+++ b/test/functional/cmdLineTests/cmdLineTest_J9tests/j9tests_exclude.xml
@@ -50,7 +50,7 @@
 <exclude id="-Xthr:secondarySpinForObjectMonitors on command line" platform="aix.*" shouldFix="false"><reason>Only for platforms that three tier spin is enabled</reason></exclude>
 
 <include id="-Xpreloaduser32" platform="win_x86.*" shouldFix="false"><reason>Only available on Windows platforms</reason></include>
-<include id="-Xpreloaduser32 -Xprotectcontiguous" platform="win_x86.*" shouldFix="false"><reason>Only available on Windows platforms</reason></include>
+<include id="-Xpreloaduser32 -Xprotectcontiguous" platform="none" shouldFix="true"><reason>Intermittent failures on Windows platforms and not available on other platforms</reason></include>
 <include id="-Xpreloaduser32 -Xnoprotectcontiguous" platform="win_x86.*" shouldFix="false"><reason>Only available on Windows platforms</reason></include>
 <include id="-Xnopreloaduser32" platform="win_x86.*" shouldFix="false"><reason>Only available on Windows platforms</reason></include>
 <include id="-Xnopreloaduser32 -Xprotectcontiguous" platform="win_x86.*" shouldFix="false"><reason>Only available on Windows platforms</reason></include>


pshipton:
jenkins test sanity win jdk10

JasonFengJ9:
Note: the test failed due to overrun, usually it takes `1-3 seconds` but took more than `180 seconds` in failed cases. 
The Java test code (https://github.com/eclipse/openj9/blob/master/test/functional/cmdLineTests/utils/src/VMBench/FibBench.java) seems simple. The time probably was on following code snippet:
```
		if (protectContiguous) {
			/* Find the largest unallocated memory region to reserve. */
			findLargestUnallocatedRegion(&contiguousRegionStart, &contiguousRegionSize);
			if (0 != contiguousRegionStart) {
				/* Align up contiguousRegionStart to be on a 65k boundary for VirtualAlloc(). */
				const UDATA alignment = 0x10000;
				const UDATA alignmentBits = (alignment - 1);
				UDATA adjustedRegionStart = (contiguousRegionStart + alignmentBits) & ~alignmentBits;

				contiguousRegionSize -= (adjustedRegionStart - contiguousRegionStart);
				contiguousRegionSize &= ~alignmentBits;
				contiguousRegionStart = adjustedRegionStart;
			}
		}
```
Will build an instrumental JVM to verify.

DanHeidinga:
@JasonFengJ9 This is a subtle behaviour change that may look accidental.  Is it worth updating the comment in the exclude file about why this is different (win32 & win64)?

pshipton:
No idea why the management tests failed in the jdk10 sanity run.

test/functional/cmdLineTests/cmdLineTest_J9tests/j9tests_exclude.xml
```diff
@@ -50,7 +50,7 @@
 <exclude id="-Xthr:secondarySpinForObjectMonitors on command line" platform="aix.*" shouldFix="false"><reason>Only for platforms that three tier spin is enabled</reason></exclude>
 
 <include id="-Xpreloaduser32" platform="win_x86.*" shouldFix="false"><reason>Only available on Windows platforms</reason></include>
-<include id="-Xpreloaduser32 -Xprotectcontiguous" platform="win_x86.*" shouldFix="false"><reason>Only available on Windows platforms</reason></include>
+<include id="-Xpreloaduser32 -Xprotectcontiguous" platform="win_x86*" shouldFix="false"><reason>Only available on Windows platforms</reason></include>
```
pshipton: The shouldFix value and the reason need to be updated.

JasonFengJ9:
https://github.com/eclipse/openj9/blob/master/test/functional/J9%20Exclude%20File%20Support/src/com/oti/j9/exclude/XMLInterface.java#L60 shows how `exclude` and `include` work with `platform` and `-xids`:
```
	public void xmlStartElement(String name, Hashtable attr)
	{
		...
		if (name.equals("exclude") || name.equals("include")) {
			String allPlatforms = (String) attr.get("platform");
			String[] platforms = ExcludeList.substrings(allPlatforms, ' ');
			boolean matched = false;

			for (int i=0; i < _fApplicableTests.length; i++) {
				String logicalType = _fApplicableTests[i];
				for (int j=0; j < platforms.length; j++) {
					String aPlatform = platforms[j];
					if (Pattern.matches(aPlatform, logicalType)) {
						matched = true;
						if (name.equals("exclude")) {
							_fTestID = (String) attr.get("id");
							_fExpiry = 0;
							...
							return;
						}
					}
				}
			}
			if (!matched && name.equals("include")) {
				_fTestID = (String) attr.get("id");
				_fExpiry = 0;
				return;
			}
			_fTestID = null;  // ignore this one
			_fExpiry = 0;
			return;
		}
	}
```
Basically `exclude` excludes a test when the platform specified by current entry matches any platform specified by `-xids`, `include` excludes a test when the platform specified by current entry doesn't 
 match any platform specified by `-xids` and allow this test if the platform matches such that `-xids` can controls what platform a test runs.

Previously `<include id="-Xpreloaduser32 -Xprotectcontiguous" platform="win_x86.*" shouldFix="false"><reason>Only available on Windows platforms</reason></include>` allows this test run only on platforms matching `win_x86.*` but exclude it on other platforms. My previous change does work and I verified that the sub-test is excluded with messages like following:
```
[-Xpreloaduser32 -Xprotectcontiguous] expiry: none :: Only available on Windows platforms
Test excluded: -Xpreloaduser32 -Xprotectcontiguous
```
Updated the pattern to `platform="none"` to avoid an impression of accidental changes. Also updated `shouldFix` and `reason`. 
Verified manually the sub-test is excluded as per following job output:
```
[-Xpreloaduser32 -Xprotectcontiguous] expiry: none :: Intermittent failures on Windows platforms and not available on other platforms
Test excluded: -Xpreloaduser32 -Xprotectcontiguous
```

Please review again.

test/functional/cmdLineTests/cmdLineTest_J9tests/j9tests_exclude.xml
JasonFengJ9: Updated `shouldFix` and `reason`, pls see comment below for clarifications.

pshipton:
I know the original patch worked, I just don't understand why it worked. Ideally the test should keep running on 32-bit where I don't think there is a problem, but I'll merge this for now to avoid the intermittent failures.

PR #2074, opened 2018-06-05
Fix warnings that can be fatal

keithc-ca:
This fixes two instances where the return value of `fgets()` was ignored; they are fatal if warnings are treated as errors.

Also use C-style comments in C source files (not all C compilers tolerate C++ comments).


DanHeidinga:
Jenkins test sanity aix jdk8

PR #2075, opened 2018-06-05
Add support for triggering recompilation for HCR

cathyzhyi:
This changeset triggers recompilation of a method body
when any HCR guard in it got patched. Otherwise, the patched method
would keeping doing OSR resulting in high CPU consumption.

To distinguish different recompilations triggered from JIT compiled
method body an extra argument is passed to the vm helper.


cathyzhyi:
@mpirvu Could you please help to review this PR?

runtime/compiler/control/CompilationController.cpp
```diff
@@ -203,7 +203,14 @@ TR_OptimizationPlan *TR::DefaultCompilationStrategy::processEvent(TR_MethodEvent
          // For sync re-compilation we have attached a plan to the persistentBodyInfo
          bodyInfo = TR::Recompilation::getJittedBodyInfoFromPC(event->_oldStartPC);
          methodInfo = bodyInfo->getMethodInfo();
-         if (bodyInfo->getUsesJProfiling() && !bodyInfo->getIsProfilingBody())
+
+         if (methodInfo->getReasonForRecompilation() == TR_PersistentMethodInfo::RecompDueToInlinedMethodRedefinition)
```
mpirvu: Could we combine the code here? It is the same logic for RecompDueToInlinedMethodRedefinition and TR_PersistentMethodInfo::RecompDueToJProfiling. Previously, we had to guess based on the type of the body because we didn't have the reason field.
Similarly, now that you passed the reason for GCR, we could use that test instead of testing `bodyInfo->getUsesGCR()`

mpirvu:
There is a case which I forgot to consider: if redefinitions happen all the time it's possible that we'll compile the same method over and over again. We need some counter that gets incremented every time a method needs to be recompiled due to InlinedMethodRedefinition and if the counter reaches some threshold we should not use OSR anymore (not inline in that method is also an option, but probably more drastic).
The counter could be stored in PersistentMethodInfo as a uint8_t quantity (we don't need more than that). It could be incremented and/tested when a compilation with InlinedMethodRedefinition flag starts.

runtime/compiler/control/CompilationController.cpp
cathyzhyi: Thanks for the suggestion. modified the change.

cathyzhyi:
> There is a case which I forgot to consider: if redefinitions happen all the time it's possible that we'll compile the same method over and over again. We need some counter that gets incremented every time a method needs to be recompiled due to InlinedMethodRedefinition and if the counter reaches some threshold we should not use OSR anymore (not inline in that method is also an option, but probably more drastic).
The counter could be stored in PersistentMethodInfo as a uint8_t quantity (we don't need more than that). It could be incremented and/tested when a compilation with InlinedMethodRedefinition flag starts.

thanks for the suggestion. Changed the code. The part to check the number of recompilation triggered by HCR is in https://github.com/eclipse/omr/pull/2642

cathyzhyi:
@andrewcraik Hey Andrew, could you help to take a look the OSRGuardInsertion part of this change?

cathyzhyi:
Jenkins test sanity depends https://github.com/eclipse/omr/pull/2632

runtime/compiler/control/CompilationController.cpp
```diff
@@ -203,25 +203,21 @@ TR_OptimizationPlan *TR::DefaultCompilationStrategy::processEvent(TR_MethodEvent
          // For sync re-compilation we have attached a plan to the persistentBodyInfo
          bodyInfo = TR::Recompilation::getJittedBodyInfoFromPC(event->_oldStartPC);
          methodInfo = bodyInfo->getMethodInfo();
-         if (bodyInfo->getUsesJProfiling() && !bodyInfo->getIsProfilingBody())
+         if (methodInfo->getReasonForRecompilation() == TR_PersistentMethodInfo::RecompDueToInlinedMethodRedefinition)
+            methodInfo->incrementNumberOfInlinedMethodRedefinition();
```
mpirvu: Is it possible for several threads to enter this code more or less concurrently and therefore increment the counter even though we are actually dealing with the same recompilation event? If that is the case a better solution would be to increment this counter at the beginning of the compilation, right before the code that tests whether we need to disable nextgenHCR

andrewcraik:
Jenkins test sanity

AdamBrousseau:
Jenkins test sanity win jdk8

andrewcraik:
Jenkins test sanity

andrewcraik:
Jenkins test sanity

PR #2081, opened 2018-06-05
DDR: Allow GC ObjectHeapIterator AOL advance to the Top address

dmitripivkine:
Allow GCObjectHeapIteratorAddressOrderedList::advance() move cursor to
the top address of the heap. It would prevent throwing an exception in
case if it advanced to the very top address. This would not create a
logical problem because advance() itself does not return next address.
An APIs like next() or hasNext() still function properly.


dmitripivkine:

dmitripivkine:

PR #2082, opened 2018-06-05
Make java/lang/Class available in AOT

jdmpapin:
Currently getClassClassPointer() refuses to return the J9Class pointer
for java/lang/Class when compiling AOT (ahead of time) even though Class
is practically guaranteed to be identical at AOT load time.

This commit allows callers to opt in to finding Class, in which case
they must ensure that their use of the J9Class pointer is acceptable for
AOT compilation.

### DIFF
--- a/runtime/compiler/compile/J9Compilation.cpp
+++ b/runtime/compiler/compile/J9Compilation.cpp
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2000, 2017 IBM Corp. and others
+ * Copyright (c) 2000, 2018 IBM Corp. and others
  *
  * This program and the accompanying materials are made available under
  * the terms of the Eclipse Public License 2.0 which accompanies this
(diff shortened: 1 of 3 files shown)


runtime/compiler/compile/J9Compilation.cpp
```diff
@@ -1144,9 +1148,36 @@ J9::Compilation::addAsMonitorAuto(TR::SymbolReference* symRef, bool dontAddIfDLT
...
 TR_OpaqueClassBlock *
-J9::Compilation::getClassClassPointer()
+J9::Compilation::getClassClassPointer(bool isVettedForAOT)
    {
-   return _ObjectClassPointer ? self()->fe()->getClassClassPointer(_ObjectClassPointer) : 0;
+   if (!isVettedForAOT)
+      return _ObjectClassPointer ? self()->fe()->getClassClassPointer(_ObjectClassPointer) : 0;
+
+   if (_aotClassClassPointerInitialized)
+      return _aotClassClassPointer;
```
dsouzai: This might be nit-picky, but do we need `_aotClassClassPointerInitialized`? Isn't it sufficient to know that if `_aotClassClassPointer` is NULL it hasn't been initialized?

jdmpapin: It might be unavailable even though we set `isVettedForAOT=true`

andrewcraik:
Jenkins test sanity

andrewcraik:
windows jdk 8 failure was due to #2129

PR #2083, opened 2018-06-05
Allow fabricated loadaddr for classes in AOT

jdmpapin:
More specifically for classes loaded by the bootstrap class loader,
which are unambiguously identifiable at relocation time.

This allows the JIT compiler to make up references to types about which
it has special knowledge. In particular, it's now possible to create a
loadaddr for java/lang/Class "out of thin air," without a corresponding
constant pool entry available.

At a high level, the appropriate relocations are carried out as follows:

  1. Allow TR_ClassAddress ExternalRelocations to be created without a
  constant pool index.

  2. Before grouping into IteratedExternalRelocations, identify these
  and set their kind to the new TR_ArbitraryClassAddress.

  3. Write the relocation record for TR_ArbitraryClassAddress with the
  same data as would be required for TR_ClassPointer.

  4. Relocate TR_ArbitraryClassAddress by finding the class as though
  for TR_ClassPointer, and then patching the code as though for
  TR_ClassAddress (to deal with materialization sequences).

Note that because the compilation has the class pointer, it must have
been through validateArbitraryClass(). In case the class can't be found
at relocation time, the validation record will cause relocation to fail,
so the value used for missing classes is irrelevant.

Power code generation is updated to generate a materialization sequence
and relocation when it encounters a fabricated class symbol reference in
an AOT (ahead of time) compilation. Previously it used an unresolved
snippet, which requires a constant pool index.

### DIFF
--- a/runtime/compiler/arm/codegen/J9AheadOfTimeCompile.cpp
+++ b/runtime/compiler/arm/codegen/J9AheadOfTimeCompile.cpp
@@ -39,6 +39,7 @@
 #include "il/Node_inlines.hpp"
 #include "il/SymbolReference.hpp"
 #include "il/symbol/LabelSymbol.hpp"
+#include "il/symbol/StaticSymbol.hpp"
 
 #define  NON_HELPER         0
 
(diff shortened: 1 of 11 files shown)


jdmpapin:
This is the openj9 part of eclipse/omr#2636.

jdmpapin:
Updated to avoid adding `classAddressUsesReloRecordInfo()` and `getClassAddressSymRef()` to OMR, as described in eclipse/omr#2636.

[Diff from the previous version of this commit](https://github.com/eclipse/openj9/files/2085186/aot-fab-loadaddr.patch.txt)

andrewcraik:
Ok so sanity tests say JDK8 windows build failed due to #2129. Travis build is shown failed because the dependent OMR change had not been merged at that time. The only issue to be resolved here is to fix the copyright failure. @jdmpapin can you please verify the copyright and then do a push to update the change? I don't think we need sanity again - we need travis clean and the copyright check clean provided the only update you make is to the copyright headers

jdmpapin:
Updated the copyright header in J9AheadOfTimeCompile.hpp

andrewcraik:
Build is clean and CI completed previously as noted - merging

PR #2085, opened 2018-06-05
Fix String.contentEquals(StringBuffer buffer)

dhong44:
Fix the String.contentEquals() function to check all bytes
of a string when a UTF16 String and StringBuffer are passed.


fjeremic:
Jenkins test sanity

fjeremic:
Windows failures are due to infrastructure issues. @AdamBrousseau "No space left on device" FYI.

PR #2090, opened 2018-06-06
Improve entropy of reply key

pdbain-ibm:
com.ibm.tools.attach.target.Reply uses a randomly generated key to verify that
the process responding to an attach attempt is the intended target.  The
existing approach used a random value based on time, which makes it
predictable. The key also had only 32 bits of entropy.

Use SecureRandom to generate a longer and more random value when security is
important.

### DIFF
--- a/jcl/src/jdk.attach/share/classes/com/ibm/tools/attach/attacher/OpenJ9VirtualMachine.java
+++ b/jcl/src/jdk.attach/share/classes/com/ibm/tools/attach/attacher/OpenJ9VirtualMachine.java
@@ -385,7 +385,7 @@ final class OpenJ9VirtualMachine extends VirtualMachine implements Response {
 
 				targetServer = new ServerSocket(0); /* select a free port */
 				portNumber = Integer.valueOf(targetServer.getLocalPort());
-				String key = Integer.toHexString((IPC.getRandomNumber()));
+				String key = IPC.getRandomString();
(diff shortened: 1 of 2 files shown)


pshipton:
jenkins test all plinux jdk8

jcl/src/jdk.attach/share/classes/com/ibm/tools/attach/attacher/OpenJ9VirtualMachine.java
```diff
@@ -385,7 +385,7 @@ private void tryAttachTarget(int timeout) throws IOException,
 
 				targetServer = new ServerSocket(0); /* select a free port */
 				portNumber = Integer.valueOf(targetServer.getLocalPort());
-				String key = Integer.toHexString((IPC.getRandomNumber()));
+				String key = IPC.getRandomString();
```
DanHeidinga: Is `IPC.getRandomNumber()` used anymore?  Can it be deleted?

pdbain-ibm: It's used in FileLock.java (ca. line 177) when dealing with stale files.

pdbain-ibm:
Failed test:  58 - - cmdLineTester_jvmtitests_hcr_OSRG_nongold_SE80_0

pshipton:
@DanHeidinga any other comments?

DanHeidinga:
looks good to me

PR #2091, opened 2018-06-06
Missing Register Dependency for Lock Reservation

R-Santhir:
Adding in a missing register dependency. The missing dependency could cause
data that is needed later to be overwritten.


R-Santhir:
@klangman I think I've identified and added the missing reg dep

klangman:
Yep.. Looks right to me. Thanks Ryan!

R-Santhir:
@zl-wang could you take a look and merge when you get a chance?

PR #2092, opened 2018-06-06
Fix JVMTI field event reporting

gacholio:
The JVMTI field watch/modify events should be reported before the object
field is read/written.  JNI modifications to instance fields were
reporting the event after the write.


DanHeidinga:
For the `set` cases, can the hook  modify the value?  If so, are we missing typechecks before the store?

What about reporting the array accesses?  I don't see any hooks triggered in the array get/set

gacholio:
No, there is no facility in JVMTI to modify the written value.  The internal hooks allow it in certain cases because they pass the address of the pending stack slot, but it is not in the hook contract.

Array accesses are not covered by the field watch APIs, only things that can be described by jfieldID.

DanHeidinga:

PR #2095, opened 2018-06-06
Cache j.l.Class package name

JasonFengJ9:
Cache `j.l.Class` package name

Added a private instance field to store the package name;
The package name is interned as well;
Minor refactoring to cache the package name.

Manually verified that `pConfig` still compiles.

Reviewer: @pshipton 
FYI: @DanHeidinga 

### DIFF
--- a/jcl/src/java.base/share/classes/java/lang/Class.java
+++ b/jcl/src/java.base/share/classes/java/lang/Class.java
@@ -139,16 +139,16 @@ public final class Class<T> implements java.io.Serializable, GenericDeclaration,
 	static final Class<?>[] EmptyParameters = new Class<?>[0];
 	
 	/*[PR VMDESIGN 485]*/
-	private long vmRef;
-	private ClassLoader classLoader;
+	private transient long vmRef;
+	private transient ClassLoader classLoader;
 
 	/*[IF Sidecar19-SE]*/
-	private Module module;
+	private transient Module module;
 	/*[ENDIF]*/
 
 	/*[PR CMVC 125822] Move RAM class fields onto the heap to fix hotswap crash */
-	private ProtectionDomain protectionDomain;
-	private String classNameString;
+	private transient ProtectionDomain protectionDomain;
+	private transient String classNameString;
 
 	private static final class AnnotationVars {
 		AnnotationVars() {}
@@ -178,9 +178,9 @@ public final class Class<T> implements java.io.Serializable, GenericDeclaration,
 	private transient EnumVars<T> enumVars;
 	private static long enumVarsOffset = -1;
 	
-	J9VMInternals.ClassInitializationLock initializationLock;
+	transient J9VMInternals.ClassInitializationLock initializationLock;
 	
-	private Object methodHandleCache;
+	private transient Object methodHandleCache;
 	
 	/*[PR Jazz 85476] Address locking contention on classRepository in getGeneric*() methods */
 	private transient ClassRepositoryHolder classRepoHolder;
@@ -1866,7 +1866,7 @@ private static String getNonArrayClassPackageName(Class<?> clz) {
 	String name = clz.getName();
 	int index = name.lastIndexOf('.');
 	if (index >= 0) {
-		return name.substring(0, index);
+		return name.substring(0, index).intern();
 	}
 	return ""; //$NON-NLS-1$
 }


jcl/src/java.base/share/classes/java/lang/Class.java
```diff
@@ -149,6 +149,7 @@
 	/*[PR CMVC 125822] Move RAM class fields onto the heap to fix hotswap crash */
 	private ProtectionDomain protectionDomain;
 	private String classNameString;
+	private String packageName;
```
pshipton: I see classNameString isn't set transient, but many of the other cached variables are. Could mark these as transient to be consistent, although I don't expect it makes any difference since Class is handled specially for serialization.

DanHeidinga:
Cache the value locally and intern() the string.  Seems like overkill to do both.  Either the local cache or interning would be sufficient to ensure the identical object was returned.

JasonFengJ9:
Agreed that `intern()` alone ensures same String object returned, and GC doesn't need work unnecessary if there are lots of calls to create package name objects. Caching the value locally could avoid potential native calls `isPrimitive()` and `isArray()` though there is no application case to show the benefit. 
Attempting a footprint analysis via `Liberty daytrader3` startup test.

jcl/src/java.base/share/classes/java/lang/Class.java
JasonFengJ9: `classNameString` and `packageName` values are consistent after they are computed and could be persisted via serialization. 
It seems there is no specially serialization for `j.l.Class` (maybe I missed something?)

pshipton: These fields are not part of the serialization spec for java.lang.Class and should be transient.

JasonFengJ9: Agreed, marked a few other instance fields as `transient`.

JasonFengJ9:
Run the `Liberty DayTrader3 footprint` benchmark tests. It shows that adding an extra instance field has visible impact (`1 - 4%` increasing) while `intern()` increases footprint by `0.1 - 1.1%`.
With this PR, `j.l.Class.getPackageName()` is going to return same `String` object instead of creating a new `String` object for each call.

Please review again.

jcl/src/java.base/share/classes/java/lang/Class.java
```diff
@@ -1866,7 +1866,7 @@ private static String getNonArrayClassPackageName(Class<?> clz) {
 	String name = clz.getName();
 	int index = name.lastIndexOf('.');
 	if (index >= 0) {
-		return name.substring(0, index);
+		return name.substring(0, index).intern();
```
andrewcraik: The substring call here will create a new String (all be it short lived) which will then be passed to the interning logic. In Java 8 the backing array will be shared between the substring and the new String, but in Java 9 and later the substring will have to copy the backing array as well as create the new String object.

The intern will reduce the number of escaping strings so it isn't strictly worse. The unfortunate thing is that the intern operation will have to compute the string hashcode on the substring every time since substring doesn't derive it from the base string.

We also have to repeat the lastIndexOf and the clz.getName every time which is also computationally expensive.

andrewcraik: If we need the space for perf could one of the other fields (like the EnumVar field or methodHandleCache) move down the hierarchy to free up space for the package name since that is more generic?

JasonFengJ9: First to clarify that intention of this PR is to match RI behaviour discovered during another problem investigation. There is no performance testcase related to this change.
`Liberty DayTrader3 footprint` benchmark tests shows this change has little impact (mainly in noise level).

Agreed that computing expensive methods are invoked before `intern()`. Though it is can be avoided via a local cached instance field (or mechanism alike), will not implement it now until there is a use case showing measurable performance gain.

andrewcraik: Well we should fix the problem which this does - I think the perf improvement is worth doing as a follow-on if you expect this to be called more since we know it is an issue and it would be better to fix it before it becomes a performance problem

JasonFengJ9: It is not expected that `j.l.Class.getPackageName()` is called often enough to have a (throughput) performance impact. There is no testcase to support that. This PR avoids creating a new String object for each `j.l.Class.getPackageName()` call for all classes within same package.

Implementing local cache is going to introduce overhead such as an extra field (or alike) which is not desirable if there is no application demonstrating performance gain.

DanHeidinga:
Relying on String.intern() is a better choice then introducing yet another cache that would need to be managed.  Given that this API isn't known to be performance sensitive, the existing approach seems reasonable.

DanHeidinga:

DanHeidinga:

Issue #1934, opened 2018-05-17
Re-enable toLowerCase / toUpperCase acceleration in the JIT


fjeremic:
As part of #1786 the acceleration for `toUpperCase` / `toLowerCase` was disabled for Java 9 onward. This support needs to be re-enabled. This will require slight altercations to how the recognized methods are laid out in our implementation in the JIT.

PR #2098, opened 2018-06-06
WIP: Re-enable toLowerCase/toUpperCase acceleration

dhong44:
Modify API call so that caseConversionHelper receives
underlying arrays and length of string.

Closes: #1934

### DIFF
--- a/jcl/src/java.base/share/classes/com/ibm/jit/JITHelpers.java
+++ b/jcl/src/java.base/share/classes/com/ibm/jit/JITHelpers.java
@@ -166,6 +166,100 @@ public final class JITHelpers {
 
 
 	/*
+	 * To be recognized by the JIT and returns true if the hardware supports SIMD case conversion.
+	 */
+	public boolean supportsIntrinsicCaseConversion() {
+		return false;
+	}
+
+	/**
+	 * To be used by the JIT when performing SIMD upper case conversion with Latin 1 strings.
+	 * @param value the underlying array for the source string.
+	 * @param output a new array which will be used for the converted string.
+	 * @param length the number of bytes used to represent the string.
+	 * @return True if intrinsic conversion was successful, false if fallback conversion must be used.
+	 */
+	public boolean toUpperIntrinsicLatin1(byte[] value, byte[] output, int length) {
+		return false;
+	}
+
+	/**
+	 * To be used by the JIT when performing SIMD lower case conversion with Latin 1 strings.
+	 * @param value the underlying array for the source string.
+	 * @param output a new array which will be used for the converted string.
+	 * @param length the number of bytes used to represent the string.
+	 * @return True if intrinsic conversion was successful, false if fallback conversion must be used.
+	 */
+	public boolean toLowerIntrinsicLatin1(byte[] value, byte[] output, int length) {
+		return false;
+	}
+
+	/**
+	 * To be used by the JIT when performing SIMD upper case conversion with UTF16 strings.
+	 * @param value the underlying array for the source string.
+	 * @param output a new array which will be used for the converted string.
+	 * @param length the number of bytes used to represent the string.
+	 * @return True if intrinsic conversion was successful, false if fallback conversion must be used.
+	 */
+	public boolean toUpperIntrinsicUTF16(byte[] value, byte[] output, int length) {
+		return false;
+	}
+
+	/**
+	 * To be used by the JIT when performing SIMD lower case conversion with UTF16 strings.
+	 * @param value the underlying array for the source string.
+	 * @param output a new array which will be used for the converted string.
+	 * @param length the number of bytes used to represent the string.
+	 * @return True if intrinsic conversion was successful, false if fallback conversion must be used.
+	 */
+	public boolean toLowerIntrinsicUTF16(byte[] value, byte[] output, int length) {
+		return false;
+	}
+	/**
+	 * To be used by the JIT when performing SIMD upper case conversion with Latin 1 strings.
+	 * @param value the underlying array for the source string.
+	 * @param output a new array which will be used for the converted string.
+	 * @param length the number of bytes used to represent the string.
+	 * @return True if intrinsic conversion was successful, false if fallback conversion must be used.
+	 */
+	public boolean toUpperIntrinsicLatin1(char[] value, char[] output, int length) {
+		return false;
+	}
+
+	/**
+	 * To be used by the JIT when performing SIMD lower case conversion with Latin 1 strings.
+	 * @param value the underlying array for the source string.
+	 * @param output a new array which will be used for the converted string.
+	 * @param length the number of bytes used to represent the string.
+	 * @return True if intrinsic conversion was successful, false if fallback conversion must be used.
+	 */
+	public boolean toLowerIntrinsicLatin1(char[] value, char[] output, int length) {
+		return false;
+	}
+
+	/**
+	 * To be used by the JIT when performing SIMD upper case conversion with UTF16 strings.
+	 * @param value the underlying array for the source string.
+	 * @param output a new array which will be used for the converted string.
+	 * @param length the number of bytes used to represent the string.
+	 * @return True if intrinsic conversion was successful, false if fallback conversion must be used.
+	 */
+	public boolean toUpperIntrinsicUTF16(char[] value, char[] output, int length) {
+		return false;
+	}
+
+	/**
+	 * To be used by the JIT when performing SIMD lower case conversion with UTF16 strings.
+	 * @param value the underlying array for the source string.
+	 * @param output a new array which will be used for the converted string.
+	 * @param length the number of bytes used to represent the string.
+	 * @return True if intrinsic conversion was successful, false if fallback conversion must be used.
+	 */
+	public boolean toLowerIntrinsicUTF16(char[] value, char[] output, int length) {
+		return false;
+	}
+
+	/*
 	 * sun.misc.Unsafe.get* and put* have to generate internal control flow for correctness due to different object shapes. The JIT emitted sequences
 	 * checks for and handles: 1) standard fields in normal objects 2) static fields from classes 3) entries from arrays This sequence is branchy and
 	 * hard for us to optimize. The better solution in cases where we know the type of object is to use case specific accessors which are the methods
--- a/jcl/src/java.base/share/classes/java/lang/String.java
+++ b/jcl/src/java.base/share/classes/java/lang/String.java
@@ -71,7 +71,6 @@ public final class String implements Serializable, Comparable<String>, CharSeque
 /*[IF Sidecar19-SE]*/
 	// DO NOT CHANGE OR MOVE THIS LINE
 	// IT MUST BE THE FIRST THING IN THE INITIALIZATION
-	private static final boolean STRING_OPT_IN_HW = StrCheckHWAvailable();
 	private static final long serialVersionUID = -6849794470754667710L;
 
 	/**
@@ -2709,15 +2708,26 @@ public final class String implements Serializable, Comparable<String>, CharSeque
 			return this;
 		}
 
-		if (StrHWAvailable() && language == "en") { //$NON-NLS-1$
-			String output = new String(lengthInternal());
-			if (toLowerHWOptimized(output))
-				return output;
-		}
+		int sLength = lengthInternal();
 
 		if (enableCompression && (null == compressionFlag || coder == LATIN1)) {
+
+			if (helpers.supportsIntrinsicCaseConversion() && language == "en") { //$NON-NLS-1$
+				byte[] output = new byte[sLength << coder];
+
+				if (helpers.toLowerIntrinsicLatin1(value, output, sLength)) {
+					return new String(output, LATIN1);
+				}
+			}
 			return StringLatin1.toLowerCase(this, value, locale);
 		} else {
+			if (helpers.supportsIntrinsicCaseConversion() && language == "en") { //$NON-NLS-1$
+				byte[] output = new byte[sLength << coder];
+
+				if (helpers.toLowerIntrinsicUTF16(value, output, sLength * 2)) {
+					return new String(output, UTF16);
+				}
+			}
 			return StringUTF16.toLowerCase(this, value, locale);
 		}
 	}
@@ -2755,15 +2765,26 @@ public final class String implements Serializable, Comparable<String>, CharSeque
 			return this;
 		}
 
-		if (StrHWAvailable() && language == "en") { //$NON-NLS-1$
-		   String output = new String(lengthInternal()); 
-		   if (toUpperHWOptimized(output))
-				return output;
-		}
+		int sLength = lengthInternal();
 
 		if (enableCompression && (null == compressionFlag || coder == LATIN1)) {
+
+			if (helpers.supportsIntrinsicCaseConversion() && language == "en") { //$NON-NLS-1$
+				byte[] output = new byte[sLength << coder];
+
+				if (helpers.toUpperIntrinsicLatin1(value, output, sLength)) {
+					return new String(output, LATIN1);
+				}
+			}
 			return StringLatin1.toUpperCase(this, value, locale);
 		} else {
+			if (helpers.supportsIntrinsicCaseConversion() && language == "en") { //$NON-NLS-1$
+				byte[] output = new byte[sLength << coder];
+
+				if (helpers.toUpperIntrinsicUTF16(value, output, sLength * 2)) {
+					return new String(output, UTF16);
+				}
+			}
 			return StringUTF16.toUpperCase(this, value, locale);
 		}
 	}
@@ -3811,33 +3832,6 @@ public final class String implements Serializable, Comparable<String>, CharSeque
 		return stringJoiner.toString();
 	}
 
-	// DO NOT MODIFY THIS METHOD
-	/*
-	 * The method is only called once to setup the flag DFP_HW_AVAILABLE Return value true - when JIT compiled this method, replaces it with loadconst
-	 * 1 if -Xjit:disableHWAcceleratedStringCaseConv hasn't been supplied false - if still interpreting this method or disabled by VM option
-	 */
-	private final static boolean StrCheckHWAvailable() {
-		return false;
-	}
-
-	// DO NOT MODIFY THIS METHOD
-	/*
-	 * Return value true - when JIT compiled this method, replaces it with loadconst 1 if -Xjit:disableHWAcceleratedStringCaseConv hasn't been supplied
-	 * false - if still interpreting this method or disabled by VM option
-	 */
-	private final static boolean StrHWAvailable() {
-		return STRING_OPT_IN_HW;
-	}
-
-	// DO NOT CHANGE CONTENTS OF THESE METHODS
-	private final boolean toUpperHWOptimized(String input) {
-		return false;
-	}
-
-	private final boolean toLowerHWOptimized(String input) {
-		return false;
-	}
-	
 	static void checkBoundsBeginEnd(int begin, int end, int length) {
 		if ((begin >= 0) && (begin <= end) && (end <= length)) {
 			return;
@@ -3946,7 +3940,6 @@ public final class String implements Serializable, Comparable<String>, CharSeque
 /*[ELSE] Sidecar19-SE*/
 	// DO NOT CHANGE OR MOVE THIS LINE
 	// IT MUST BE THE FIRST THING IN THE INITIALIZATION
-	private static final boolean STRING_OPT_IN_HW = StrCheckHWAvailable();
 	private static final long serialVersionUID = -6849794470754667710L;
 
 	/**
@@ -6745,10 +6738,20 @@ written authorization of the copyright holder.
 			return this;
 		}
 
-		if (StrHWAvailable() && language == "en") { //$NON-NLS-1$
-			String output = new String(lengthInternal());
-			if (toLowerHWOptimized(output))
-				return output;
+		if (helpers.supportsIntrinsicCaseConversion() && language == "en") { //$NON-NLS-1$
+			int sLength = lengthInternal();
+
+			if (enableCompression && (null == compressionFlag || count >= 0)) {
+				char[] output = new char[(sLength + 1) / 2];
+				if (helpers.toLowerIntrinsicLatin1(value, output, sLength)) {
+					return new String(output, 0, sLength, true);
+				}
+			} else {
+				char[] output = new char[sLength];
+				if (helpers.toLowerIntrinsicUTF16(value, output, sLength * 2)) {
+					return new String(output, 0, sLength, false);
+				}
+			}
 		}
 
 		return toLowerCaseCore(language);
@@ -7070,10 +7073,20 @@ written authorization of the copyright holder.
 			return this;
 		}
 
-		if (StrHWAvailable() && language == "en") { //$NON-NLS-1$
-		   String output = new String(lengthInternal()); 
-		   if (toUpperHWOptimized(output))
-				return output;
+		if (helpers.supportsIntrinsicCaseConversion() && language == "en") { //$NON-NLS-1$
+			int sLength = lengthInternal();
+
+			if (enableCompression && (null == compressionFlag || count >= 0)) {
+				char[] output = new char[(sLength + 1) / 2];
+				if (helpers.toUpperIntrinsicLatin1(value, output, sLength)){
+					return new String(output, 0, sLength, true);
+				}
+			} else {
+				char[] output = new char[sLength];
+				if (helpers.toUpperIntrinsicUTF16(value, output, sLength * 2)){
+					return new String(output, 0, sLength, false);
+				}
+			}
 		}
 
 		return toUpperCaseCore(language);
@@ -8268,33 +8281,6 @@ written authorization of the copyright holder.
 		}
 
 		return stringJoiner.toString();
-	}
-
-	// DO NOT MODIFY THIS METHOD
-	/*
-	 * The method is only called once to setup the flag DFP_HW_AVAILABLE Return value true - when JIT compiled this method, replaces it with loadconst
-	 * 1 if -Xjit:disableHWAcceleratedStringCaseConv hasn't been supplied false - if still interpreting this method or disabled by VM option
-	 */
-	private final static boolean StrCheckHWAvailable() {
-		return false;
-	}
-
-	// DO NOT MODIFY THIS METHOD
-	/*
-	 * Return value true - when JIT compiled this method, replaces it with loadconst 1 if -Xjit:disableHWAcceleratedStringCaseConv hasn't been supplied
-	 * false - if still interpreting this method or disabled by VM option
-	 */
-	private final static boolean StrHWAvailable() {
-		return STRING_OPT_IN_HW;
-	}
-
-	// DO NOT CHANGE CONTENTS OF THESE METHODS
-	private final boolean toUpperHWOptimized(String input) {
-		return false;
-	}
-
-	private final boolean toLowerHWOptimized(String input) {
-		return false;
 	}
 
 /*[ENDIF] Sidecar19-SE*/
--- a/jcl/src/java.base/share/classes/java/lang/String.java
+++ b/jcl/src/java.base/share/classes/java/lang/String.java
@@ -878,17 +878,6 @@ public final class String implements Serializable, Comparable<String>, CharSeque
 	}
 
 	/**
-	 * Creates a new string from the with specified length
-	 *
-	 * @param numChars
-	 *			  length of new string
-	 */
-	private String(int numChars) {
-		value = new byte[numChars * 2];
-		coder = UTF16;
-	}
-
-	/**
 	 * Creates a string from the contents of a StringBuffer.
 	 *
 	 * @param buffer
@@ -4719,22 +4708,6 @@ public final class String implements Serializable, Comparable<String>, CharSeque
 		value = string.value;
 		count = string.count;
 		hashCode = string.hashCode;
-	}
-
-	/**
-	 * Creates a new string from the with specified length
-	 *
-	 * @param numChars
-	 *			  length of new string
-	 */
-	private String(int numChars) {
-		if (enableCompression) {
-			value = new char[numChars];
-			count = numChars | uncompressedBit;
-		} else {
-			value = new char[numChars];
-			count = numChars;
-		}
 	}
 
 	/**
--- a/runtime/compiler/x/codegen/J9TreeEvaluator.cpp
+++ b/runtime/compiler/x/codegen/J9TreeEvaluator.cpp
@@ -14334,16 +14334,14 @@ J9::X86::TreeEvaluator::directCallEvaluator(TR::Node *node, TR::CodeGenerator *c
       case TR::java_lang_String_andOR:
          return TR::TreeEvaluator::andORStringEvaluator(node, cg);
 
-      case TR::java_lang_String_toUpperHWOptimized:
-      case TR::java_lang_String_toUpperHWOptimizedDecompressed:
-         return TR::TreeEvaluator::decompressedStringToUpperCaseEvalutor(node, cg);
-      case TR::java_lang_String_toUpperHWOptimizedCompressed:
-         return TR::TreeEvaluator::compressedStringToUpperCaseEvalutor(node, cg);
-      case TR::java_lang_String_toLowerHWOptimized:
-      case TR::java_lang_String_toLowerHWOptimizedDecompressed:
-         return TR::TreeEvaluator::decompressedStringToLowerCaseEvalutor(node, cg);
-      case TR::java_lang_String_toLowerHWOptimizedCompressed:
-         return TR::TreeEvaluator::compressedStringToUpperCaseEvalutor(node, cg);
+      case TR::com_ibm_jit_JITHelpers_toUpperIntrinsicUTF16:
+         return TR::TreeEvaluator::toUpperIntrinsicUTF16Evaluator(node, cg);
+      case TR::com_ibm_jit_JITHelpers_toUpperIntrinsicLatin1:
+         return TR::TreeEvaluator::toUpperIntrinsicLatin1Evaluator(node, cg);
+      case TR::com_ibm_jit_JITHelpers_toLowerIntrinsicUTF16:
+         return TR::TreeEvaluator::toLowerIntrinsicUTF16Evaluator(node, cg);
+      case TR::com_ibm_jit_JITHelpers_toLowerIntrinsicLatin1:
+         return TR::TreeEvaluator::toLowerIntrinsicLatin1Evaluator(node, cg);
       default:
          break;
       }
@@ -14665,7 +14663,7 @@ class J9::X86::TreeEvaluator::CaseConversionManager {
 };
 
 TR::Register *
-J9::X86::TreeEvaluator::compressedStringToUpperCaseEvalutor(TR::Node *node, TR::CodeGenerator *cg)
+J9::X86::TreeEvaluator::toUpperIntrinsicLatin1Evaluator(TR::Node *node, TR::CodeGenerator *cg)
    {
    CaseConversionManager manager(true /* isCompressedString */, false /* toLowerCase */);
    return TR::TreeEvaluator::stringCaseConversionHelper(node, cg, manager);
@@ -14673,21 +14671,21 @@ J9::X86::TreeEvaluator::compressedStringToUpperCaseEvalutor(TR::Node *node, TR::
 
 
 TR::Register *
-J9::X86::TreeEvaluator::compressedStringToLowerCaseEvalutor(TR::Node *node, TR::CodeGenerator *cg)
+J9::X86::TreeEvaluator::toLowerIntrinsicLatin1Evaluator(TR::Node *node, TR::CodeGenerator *cg)
    {
    CaseConversionManager manager(true/* isCompressedString */, true /* toLowerCase */);
    return TR::TreeEvaluator::stringCaseConversionHelper(node, cg, manager);
    }
 
 TR::Register *
-J9::X86::TreeEvaluator::decompressedStringToUpperCaseEvalutor(TR::Node *node, TR::CodeGenerator *cg)
+J9::X86::TreeEvaluator::toUpperIntrinsicUTF16Evaluator(TR::Node *node, TR::CodeGenerator *cg)
    {
    CaseConversionManager manager(false /* isCompressedString */, false /* toLowerCase */);
    return TR::TreeEvaluator::stringCaseConversionHelper(node, cg, manager);
    }
 
 TR::Register *
-J9::X86::TreeEvaluator::decompressedStringToLowerCaseEvalutor(TR::Node *node, TR::CodeGenerator *cg)
+J9::X86::TreeEvaluator::toLowerIntrinsicUTF16Evaluator(TR::Node *node, TR::CodeGenerator *cg)
    {
    CaseConversionManager manager(false /* isCompressedString */, true /* toLowerCase */);
    return TR::TreeEvaluator::stringCaseConversionHelper(node, cg, manager);
@@ -14724,17 +14722,16 @@ TR::Register *
 J9::X86::TreeEvaluator::stringCaseConversionHelper(TR::Node *node, TR::CodeGenerator *cg, CaseConversionManager &manager)
    {
    #define iComment(str) if (debug) debug->addInstructionComment(cursor, (const_cast<char*>(str)));
-   TR_ASSERT(node->getNumChildren() == 3, "node has incorrect number of children");
    TR::RegisterDependencyConditions *deps = generateRegisterDependencyConditions((uint8_t)14, (uint8_t)14, cg);
-   TR::Register *srcArray = cg->evaluate(node->getFirstChild());
+   TR::Register *srcArray = cg->evaluate(node->getChild(1));
    deps->addPostCondition(srcArray, TR::RealRegister::NoReg, cg);
    deps->addPreCondition(srcArray, TR::RealRegister::NoReg, cg);
 
-   TR::Register *dstArray = cg->evaluate(node->getSecondChild());
+   TR::Register *dstArray = cg->evaluate(node->getChild(2));
    deps->addPostCondition(dstArray, TR::RealRegister::NoReg, cg);
    deps->addPreCondition(dstArray, TR::RealRegister::NoReg, cg);
 
-   TR::Register *length = cg->intClobberEvaluate(node->getThirdChild());
+   TR::Register *length = cg->intClobberEvaluate(node->getChild(3));
    deps->addPostCondition(length, TR::RealRegister::NoReg, cg);
    deps->addPreCondition(length, TR::RealRegister::NoReg, cg);
 
@@ -14754,7 +14751,8 @@ J9::X86::TreeEvaluator::stringCaseConversionHelper(TR::Node *node, TR::CodeGener
    TR_Debug *debug = cg->getDebug();
    TR::Instruction * cursor = NULL;
 
-   uint32_t strideSize = 16;
+   uint32_t strideSize = 16; 
+   uintptrj_t headerSize = TR::Compiler->om.contiguousArrayHeaderSizeInBytes();
 
    static uint16_t MINUS1[] =
       {
@@ -14789,11 +14787,7 @@ J9::X86::TreeEvaluator::stringCaseConversionHelper(TR::Node *node, TR::CodeGener
 
    generateRegRegInstruction(MOVRegReg(), node, result, dstArray, cg);
 
-   // # charater to # byte conversion
-   if (!manager.isCompressedString()) // non compressedString using char[] which is 2 bytes per charactor
-      cursor = generateRegImmInstruction(SHLRegImm1(), node, length, 1, cg); iComment("# character to # byte conversion");
-
-   // initialize the loop counter
+   // initialize the loop counter 
    cursor = generateRegRegInstruction(XORRegReg(), node, counter, counter, cg); iComment("initialize loop counter");
 
    //calculate the residueStartLength. Later instructions compare the counter with this length and decide when to jump to the residue handling sequence
@@ -14815,7 +14809,7 @@ J9::X86::TreeEvaluator::stringCaseConversionHelper(TR::Node *node, TR::CodeGener
    generateRegRegInstruction(CMPRegReg(), node, counter, residueStartLength, cg);
    generateLabelInstruction(JGE4, node, residueStartLabel, cg);
 
-   auto srcArrayMemRef = generateX86MemoryReference(srcArray, counter, 0, 0, cg);
+   auto srcArrayMemRef = generateX86MemoryReference(srcArray, counter, 0, headerSize, cg);
    generateRegMemInstruction(MOVDQURegMem, node, xmmRegArrayContentCopy0, srcArrayMemRef, cg);
 
    //detect invalid charactors
@@ -14847,7 +14841,7 @@ J9::X86::TreeEvaluator::stringCaseConversionHelper(TR::Node *node, TR::CodeGener
       generateRegRegInstruction(manager.isCompressedString()? PSUBBRegReg: PSUBWRegReg, node,
                                 xmmRegArrayContentCopy2, xmmRegArrayContentCopy1, cg);
 
-   auto dstArrayMemRef = generateX86MemoryReference(dstArray, counter, 0, 0, cg);
+   auto dstArrayMemRef = generateX86MemoryReference(dstArray, counter, 0, headerSize, cg);
    generateMemRegInstruction(MOVDQUMemReg, node, dstArrayMemRef, xmmRegArrayContentCopy2, cg);
    generateRegImmInstruction(ADDRegImms(), node, counter, strideSize, cg);
    generateLabelInstruction(JMP4, node, caseConversionMainLoopLabel, cg);
@@ -14856,7 +14850,7 @@ J9::X86::TreeEvaluator::stringCaseConversionHelper(TR::Node *node, TR::CodeGener
    generateLabelInstruction(LABEL, node, residueStartLabel, cg);
    generateRegRegInstruction(CMPRegReg(), node, counter, length, cg);
    generateLabelInstruction(JGE4, node, endLabel, cg);
-   srcArrayMemRef = generateX86MemoryReference(srcArray, counter, 0, 0, cg);
+   srcArrayMemRef = generateX86MemoryReference(srcArray, counter, 0, headerSize, cg);
    generateRegMemInstruction( manager.isCompressedString()? MOVZXReg4Mem1: MOVZXReg4Mem2, node, singleChar, srcArrayMemRef, cg);
 
    // use unsigned compare to detect invalid range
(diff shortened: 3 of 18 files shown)


dhong44:
@fjeremic Could you please review this

jcl/src/java.base/share/classes/java/lang/String.java
```diff
@@ -2720,9 +2720,17 @@ public String toLowerCase(Locale locale) {
 		}
 
 		if (StrHWAvailable() && language == "en") { //$NON-NLS-1$
-			String output = new String(lengthInternal());
-			if (toLowerHWOptimized(output))
-				return output;
+			int sLength = lengthInternal();
+
+			if (enableCompression && (null == compressionFlag || coder == LATIN1)) {
```
fjeremic: We need to refactor this code and avoid the duplicate if. Can we merge with the path below?

jcl/src/java.base/share/classes/java/lang/String.java
```diff
@@ -2720,9 +2720,17 @@ public String toLowerCase(Locale locale) {
 		}
 
 		if (StrHWAvailable() && language == "en") { //$NON-NLS-1$
```
fjeremic: Since we are in this area of code can we rename this method here to something more sane and place it in the JitHelpers class with proper documentation? Perhaps `supportsIntrinsicCaseConversion` or if you can think of a better name I'm fine with that?

jcl/src/java.base/share/classes/java/lang/String.java
```diff
@@ -2720,9 +2720,17 @@ public String toLowerCase(Locale locale) {
...
-			String output = new String(lengthInternal());
-			if (toLowerHWOptimized(output))
-				return output;
+			int sLength = lengthInternal();
+
+			if (enableCompression && (null == compressionFlag || coder == LATIN1)) {
+				byte[] output = new byte[sLength];
+				if (toLowerHWOptimizedLatin1(value, output, sLength))
```
fjeremic: Perhaps we can match the new name of the check and rename these accordingly to `toLowerIntrinsicLatin1` / `toLowerIntrinsicUTF16` (similarly for `toUpper`).

jcl/src/java.base/share/classes/java/lang/String.java
```diff
@@ -2720,9 +2720,17 @@ public String toLowerCase(Locale locale) {
...
+						return new String(output, LATIN1);
+			} else {
+				byte[] output = new byte[sLength * 2];
```
fjeremic: This can be factored out as `sLength << coder`

jcl/src/java.base/share/classes/java/lang/String.java
```diff
@@ -2720,9 +2720,17 @@ public String toLowerCase(Locale locale) {
...
+				if (toLowerHWOptimizedUTF16(value, output, sLength * 2))
+						return new String(output, UTF16);
```
fjeremic: Let's stay consistent with the rest of the codebase and add curly brackets around if blocks.

jcl/src/java.base/share/classes/java/lang/String.java
```diff
@@ -3839,12 +3855,20 @@ private final static boolean StrHWAvailable() {
 		return STRING_OPT_IN_HW;
 	}
 
-	// DO NOT CHANGE CONTENTS OF THESE METHODS
-	private final boolean toUpperHWOptimized(String input) {
+	// DO NOT CHANGE THE CONTENTS OF THESE METHODS
+	private final static boolean toUpperHWOptimizedLatin1(byte[] value, byte[] output, int length) {
```
fjeremic: Let's move this guy to JitHelpers as well and fully document it. Similarly with others.

jcl/src/java.base/share/classes/java/lang/String.java
```diff
@@ -4569,6 +4593,21 @@ public String(char[] data, int start, int length) {
 		}
 	}
 
+	String(char[] data, int length, boolean compressed) {
```
fjeremic: I don't think there is a need for this constructor to exist. Can we not just reuse the one below with `length == 0`? These things would typically get inlined and the constant will be propagated thus ending up with the same implementation as you have here.

dhong44:
Depends on [eclipse/omr#2650](https://github.com/eclipse/omr/pull/2650)

dhong44:
Addressed your changes request from 3fe5f89 in 5d7021f.

runtime/compiler/x/codegen/J9TreeEvaluator.cpp
```diff
@@ -14765,14 +14765,14 @@ J9::X86::TreeEvaluator::directCallEvaluator(TR::Node *node, TR::CodeGenerator *c
...
-      case TR::java_lang_String_toUpperHWOptimizedUTF16:
-         return TR::TreeEvaluator::decompressedStringToUpperCaseEvaluator(node, cg);
-      case TR::java_lang_String_toUpperHWOptimizedLatin1:
-         return TR::TreeEvaluator::compressedStringToUpperCaseEvaluator(node, cg);
-      case TR::java_lang_String_toLowerHWOptimizedUTF16:
-         return TR::TreeEvaluator::decompressedStringToLowerCaseEvaluator(node, cg);
-      case TR::java_lang_String_toLowerHWOptimizedLatin1:
-         return TR::TreeEvaluator::compressedStringToLowerCaseEvaluator(node, cg);
+      case TR::com_ibm_jit_JITHelpers_toUpperIntrinsicUTF16:
+         return TR::TreeEvaluator::UTF16StringToUpperCaseEvaluator(node, cg);
```
fjeremic: We should probably rename these to `toUpperIntrinsicUTF16Evaluator` for easier search ability. And of course use the same names on Z if needed.

dhong44: Would the name `inlineToUpper` be appropriate or should I change it to something like `toUpperIntrinsic`?

fjeremic: Let's go with `toUpperIntrinsic`. By grepping for the Java method we should be able to find all related methods/functions in the VM/JIT as well. This is the intention.

dhong44: Renamed functions in x and z codegen in 5fd75e8.

runtime/compiler/z/codegen/J9TreeEvaluator.cpp
```diff
@@ -252,405 +252,258 @@ doubleMaxMinHelper(TR::Node *node, TR::CodeGenerator *cg, bool isMaxOp)
...
+   *     The code generator used to generate the instructions
+   *
+   *  \param isToUpper
+   *     Boolean representing case conversion, either to upper or to lower
+   *
+   *  \param isCompressedString
+   *     Boolean representing the string's compression
+   *
+   *  \return
+   *     0 if the conversion is unsuccessful, a non zero value otherwise
```
fjeremic: This is not true. The function returns a register which represents the return value of the Java call. We need to JavaDoc the JIT helpers you added to define what the input parameters represent and what the return value represents. This Doxygen comment needs to be modified.

runtime/compiler/z/codegen/J9TreeEvaluator.cpp
```diff
@@ -252,6 +252,249 @@ doubleMaxMinHelper(TR::Node *node, TR::CodeGenerator *cg, bool isMaxOp)
...
+
+   TR::LabelSymbol* fullVectorConversion = TR::LabelSymbol::create(cg->trHeapMemory(), cg);
+   TR::LabelSymbol* doneLabel = TR::LabelSymbol::create(cg->trHeapMemory(), cg);
+   TR::LabelSymbol* handleInvalidChars = TR::LabelSymbol::create(cg->trHeapMemory(), cg);
+   TR::LabelSymbol* loop = TR::LabelSymbol::create(cg->trHeapMemory(), cg);
+
+   TR::Instruction* instruction;
+
+   const int elementSizeMask = (isCompressedString) ? 0x0 : 0x1;    // byte or halfword mask
+   const int sizeOfVector = 16;
```
fjeremic: We cast this variable to an `int8_t` a little below. Are we able to avoid the cast by redefining the type at source?

runtime/compiler/z/codegen/J9TreeEvaluator.cpp
```diff
@@ -252,6 +252,249 @@ doubleMaxMinHelper(TR::Node *node, TR::CodeGenerator *cg, bool isMaxOp)
...
+   const bool is64 = TR::Compiler->target.is64Bit();
+   uintptrj_t headerSize = TR::Compiler->om.contiguousArrayHeaderSizeInBytes();
+
+   TR::RegisterDependencyConditions * regDeps = new (cg->trHeapMemory()) TR::RegisterDependencyConditions(0, 14, cg);
```
fjeremic: I only count 13 dependencies below. Can we save some space?

runtime/compiler/z/codegen/J9TreeEvaluator.cpp
```diff
@@ -252,6 +252,249 @@ doubleMaxMinHelper(TR::Node *node, TR::CodeGenerator *cg, bool isMaxOp)
...
+   regDeps->addPostCondition(charBufferVector, TR::RealRegister::AssignAny);
+   regDeps->addPostCondition(selectionVector, TR::RealRegister::AssignAny);
+   regDeps->addPostCondition(modifiedCaseVector, TR::RealRegister::AssignAny);
+   regDeps->addPostCondition(charOffsetVector, TR::RealRegister::AssignAny);
+   regDeps->addPostCondition(alphaRangeVector, TR::RealRegister::AssignAny);
+   regDeps->addPostCondition(alphaCondVector, TR::RealRegister::AssignAny);
+   regDeps->addPostCondition(invalidRangeVector, TR::RealRegister::AssignAny);
+   regDeps->addPostCondition(invalidCondVector, TR::RealRegister::AssignAny);
+
+   if (is64){
```
fjeremic: Formatting of braces should be consistent with the rest of the codebase.

runtime/compiler/z/codegen/J9TreeEvaluator.cpp
```diff
@@ -252,6 +252,249 @@ doubleMaxMinHelper(TR::Node *node, TR::CodeGenerator *cg, bool isMaxOp)
...
+
+   // Constant value of 0x20, used to convert between upper and lower
+   generateVRIaInstruction(cg, TR::InstOpCode::VREPI, node, charOffsetVector, static_cast<uint16_t>(0x20), elementSizeMask);
+
+   generateRRInstruction(cg, TR::InstOpCode::LR, node, loadLength, lengthRegister);
+
+   // Check for strings lengths that are too big in 32bit
+   if (!is64)
+      {
+      generateRIInstruction(cg, TR::InstOpCode::TMLH, node, lengthRegister, (uint16_t) 0x8000);
```
fjeremic: Why do we need a cast here? Also doesn't this check if `lengthRegister < 0`. I don't think this is possible. Can you explain what this code is supposed to handle?

runtime/compiler/z/codegen/J9TreeEvaluator.cpp
```diff
@@ -252,6 +252,249 @@ doubleMaxMinHelper(TR::Node *node, TR::CodeGenerator *cg, bool isMaxOp)
...
+      generateVRIaInstruction(cg, TR::InstOpCode::VLEIH, node, invalidCondVector, 0x2000, 0x0);
+      generateVRIaInstruction(cg, TR::InstOpCode::VLEIH, node, invalidCondVector, 0x2000, 0x1);
+      }
+
+   // Constant value of 0x20, used to convert between upper and lower
+   generateVRIaInstruction(cg, TR::InstOpCode::VREPI, node, charOffsetVector, static_cast<uint16_t>(0x20), elementSizeMask);
+
+   generateRRInstruction(cg, TR::InstOpCode::LR, node, loadLength, lengthRegister);
+
+   // Check for strings lengths that are too big in 32bit
```
fjeremic: I don't think the bitness of the JVM matters here. Couldn't the same problem happen on 64-bit?

runtime/compiler/z/codegen/J9TreeEvaluator.cpp
```diff
@@ -252,6 +252,249 @@ doubleMaxMinHelper(TR::Node *node, TR::CodeGenerator *cg, bool isMaxOp)
...
+      instruction = generateS390BranchInstruction(cg, TR::InstOpCode::BRC , TR::InstOpCode::COND_MASK2, node, handleInvalidChars);
+      instruction->setStartInternalControlFlow();
+      }
+
+   generateRRInstruction(cg, TR::InstOpCode::getLoadRegOpCode(), node, loadLength, lengthRegister);
+   generateRILInstruction(cg, TR::InstOpCode::NILF, node, loadLength, 0xF);
+   instruction = generateS390BranchInstruction(cg, TR::InstOpCode::BRC, TR::InstOpCode::COND_BZ, node, fullVectorConversion);
+
+   if (is64)
+      instruction->setStartInternalControlFlow();
```
fjeremic: Let's stay consistent with the rest of the codebase and add curly braces even around single statement if blocks.

runtime/compiler/z/codegen/J9TreeEvaluator.cpp
```diff
@@ -252,6 +252,249 @@ doubleMaxMinHelper(TR::Node *node, TR::CodeGenerator *cg, bool isMaxOp)
...
+
+   // VLL and VSTL take an index, not a count, so subtract the count by 1
+   generateRILInstruction(cg, TR::InstOpCode::getSubtractLogicalImmOpCode(), node, loadLength, 1);
+
+   generateVRSbInstruction(cg, TR::InstOpCode::VLL, node, charBufferVector, loadLength, generateS390MemoryReference(sourceRegister, addressOffset, headerSize, cg));
+
+   // Check for invalid characters, go to fallback individual character conversion implementation
+   if(!isCompressedString)
```
fjeremic: Consistency in space between if and parenthesis. Similarly elsewhere in the function.

runtime/compiler/z/codegen/J9TreeEvaluator.cpp
```diff
@@ -252,6 +252,249 @@ doubleMaxMinHelper(TR::Node *node, TR::CodeGenerator *cg, bool isMaxOp)
...
+      {
+      generateVRRdInstruction(cg, TR::InstOpCode::VSTRC, node, selectionVector, charBufferVector, invalidRangeVector, invalidCondVector, 0x1 , elementSizeMask);
+      generateS390BranchInstruction(cg, TR::InstOpCode::BRC, TR::InstOpCode::COND_CC1, node, handleInvalidChars);
+      }
+
+   generateVRRdInstruction(cg, TR::InstOpCode::VSTRC, node, selectionVector, charBufferVector, alphaRangeVector, alphaCondVector, 0x4, elementSizeMask);
+   generateVRRcInstruction(cg, isToUpper ? TR::InstOpCode::VS : TR::InstOpCode::VA, node, modifiedCaseVector, charBufferVector, charOffsetVector, 0x0, 0x0, elementSizeMask);
+   generateVRReInstruction(cg, TR::InstOpCode::VSEL, node, modifiedCaseVector, modifiedCaseVector, charBufferVector, selectionVector);
+
+   instruction = generateVRSbInstruction(cg, TR::InstOpCode::VSTL, node, modifiedCaseVector, loadLength, generateS390MemoryReference(destRegister, addressOffset, headerSize, cg), 0);
```
fjeremic: Why do we need to set `instruction =` here?

runtime/compiler/z/codegen/J9TreeEvaluator.cpp
```diff
@@ -252,6 +252,249 @@ doubleMaxMinHelper(TR::Node *node, TR::CodeGenerator *cg, bool isMaxOp)
...
+
+   if(!isCompressedString)
+      {
+      generateVRRdInstruction(cg, TR::InstOpCode::VSTRC, node, selectionVector, charBufferVector, invalidRangeVector, invalidCondVector, 0x1 , elementSizeMask);
+      generateS390BranchInstruction(cg, TR::InstOpCode::BRC, TR::InstOpCode::COND_CC1, node, handleInvalidChars);
+      }
+
+   generateVRRdInstruction(cg, TR::InstOpCode::VSTRC, node, selectionVector, charBufferVector, alphaRangeVector, alphaCondVector, 0x4, elementSizeMask);
+   generateVRRcInstruction(cg, isToUpper ? TR::InstOpCode::VS : TR::InstOpCode::VA, node, modifiedCaseVector, charBufferVector, charOffsetVector, 0x0, 0x0, elementSizeMask);
+   instruction = generateVRReInstruction(cg, TR::InstOpCode::VSEL, node, modifiedCaseVector, modifiedCaseVector, charBufferVector, selectionVector);
```
fjeremic: Why do we need to set `instruction =` here?

runtime/compiler/z/codegen/J9TreeEvaluator.cpp
```diff
@@ -252,6 +252,249 @@ doubleMaxMinHelper(TR::Node *node, TR::CodeGenerator *cg, bool isMaxOp)
...
+
+   generateS390BranchInstruction(cg, TR::InstOpCode::BRC, TR::InstOpCode::COND_BRC, node, doneLabel);
+
+   generateS390LabelInstruction(cg, TR::InstOpCode::LABEL, node, handleInvalidChars);
+   cg->generateDebugCounter(isToUpper? "z13/simd/toUpper/null"  : "z13/simd/toLower/null", 1, TR::DebugCounter::Cheap);
+   generateRRInstruction(cg, TR::InstOpCode::getXORRegOpCode(), node, destRegister, destRegister);
+
+   instruction=generateS390LabelInstruction(cg, TR::InstOpCode::LABEL, node, doneLabel);
+   instruction->setEndInternalControlFlow();
+   instruction->setDependencyConditions(regDeps);
```
fjeremic: I believe this line can be folded into the `generateS390LabelInstruction`. There should be an overload accepting register dependencies.

runtime/compiler/z/codegen/J9TreeEvaluator.cpp
```diff
@@ -252,6 +252,249 @@ doubleMaxMinHelper(TR::Node *node, TR::CodeGenerator *cg, bool isMaxOp)
...
+   generateRXInstruction(cg, TR::InstOpCode::getLoadAddressOpCode(), node, addressOffset, generateS390MemoryReference(addressOffset, sizeOfVector, cg));
+   generateS390BranchInstruction(cg, TR::InstOpCode::BRCT, node, loopCounter, loop);
+
+   generateS390BranchInstruction(cg, TR::InstOpCode::BRC, TR::InstOpCode::COND_BRC, node, doneLabel);
+
+   generateS390LabelInstruction(cg, TR::InstOpCode::LABEL, node, handleInvalidChars);
+   cg->generateDebugCounter(isToUpper? "z13/simd/toUpper/null"  : "z13/simd/toLower/null", 1, TR::DebugCounter::Cheap);
+   generateRRInstruction(cg, TR::InstOpCode::getXORRegOpCode(), node, destRegister, destRegister);
+
+   instruction=generateS390LabelInstruction(cg, TR::InstOpCode::LABEL, node, doneLabel);
```
fjeremic: Formatting around `=` needs to be fixed.

runtime/compiler/z/codegen/J9TreeEvaluator.cpp
dhong44: https://github.com/eclipse/openj9/blob/1ca0ab9841f1a6b17fde82402582249853a1e6e7/runtime/compiler/z/codegen/J9TreeEvaluator.cpp#L409-L417

Here there was some justification for this check. It does seem that this is checking for negative lengths, which would never happen so I don't see the need for this. I think this is dead code and can be removed.

runtime/compiler/z/codegen/J9TreeEvaluator.cpp
dhong44: For the same reasons as stated in the comment at Line 407, it seems this whole code block is unnecessary and can be removed.

runtime/compiler/z/codegen/J9TreeEvaluator.cpp
dhong44: 9b53d1e

runtime/compiler/z/codegen/J9TreeEvaluator.cpp
dhong44: 5143440

runtime/compiler/z/codegen/J9TreeEvaluator.cpp
dhong44: 56171e1

runtime/compiler/z/codegen/J9TreeEvaluator.cpp
dhong44: d0a4635

runtime/compiler/z/codegen/J9TreeEvaluator.cpp
dhong44: d0a4635

runtime/compiler/z/codegen/J9TreeEvaluator.cpp
dhong44: 56171e1

runtime/compiler/z/codegen/J9TreeEvaluator.cpp
dhong44: 56171e1

runtime/compiler/z/codegen/J9TreeEvaluator.cpp
dhong44: 4289d81

runtime/compiler/z/codegen/J9TreeEvaluator.cpp
dhong44: 4289d81

runtime/compiler/z/codegen/J9TreeEvaluator.cpp
dhong44: b796109

runtime/compiler/z/codegen/J9TreeEvaluator.cpp
dhong44: 56171e1

runtime/compiler/z/codegen/J9TreeEvaluator.cpp
dhong44: Added JavaDoc comments to JIT helper methods and fixed doxygen comment in 91004aa

dhong44:
```
============================================================
; Live regs: GPR=5 FPR=0 VRF=0 GPR64=0 AR=0 {GPR_0210, &GPR_0152, GPR_0147, &GPR_0033, &GPR_0032}
------------------------------
 n107n    (  0)  treetop                                                                              [     0x3ff69900130] bci=[-1,138,2723] rc=0 vc=394 vn=- li=- udi=- nc=1
 n106n    (  2)    icall  com/ibm/jit/JITHelpers.toUpperIntrinsicUTF16([B[BI)Z[#631  final virtual Method -128] [flags 0x20500 0x0 ] ()  [     0x3ff699000e0] bci=[-1,138,2723] rc=2 vc=394 vn=- li=- udi=- nc=4 flg=0x20
 n97n     (  1)      aload  java/lang/String.helpers Lcom/ibm/jit/JITHelpers;[#664  final Static] (obj1) [flags 0xa0307 0x0 ] (X!=0 sharedMemory )  [     0x3ff698ffe10] bci=[-1,126,2723] rc=1 vc=394 vn=- li=- udi=- nc=0 flg=0x4
 n99n     (  1)      l2a (X!=0 unneededConv sharedMemory )                                            [     0x3ff698ffeb0] bci=[-1,130,2723] rc=1 vc=394 vn=- li=- udi=- nc=1 flg=0x404
 n515n    (  1)        lshl (compressionSequence )                                                    [     0x3ff699080b0] bci=[-1,130,2723] rc=1 vc=394 vn=- li=- udi=- nc=2 flg=0x800
 n514n    (  1)          ==>iu2l (in GPR_0210) (X!=0 unneededConv )
 n512n    (  1)          iconst 2                                                                     [     0x3ff69907fc0] bci=[-1,130,2723] rc=1 vc=394 vn=- li=- udi=- nc=0
 n94n     (  2)      ==>newarray (in &GPR_0152) (X!=0 sharedMemory )
 n105n    (  1)      ishl (X>=0 cannotOverflow )                                                      [     0x3ff69900090] bci=[-1,137,2723] rc=1 vc=394 vn=- li=- udi=- nc=2 flg=0x1100
 n452n    (  1)        ==>iushr (in GPR_0147) (X>=0 cannotOverflow )
 n104n    (  1)        iconst 1 (X!=0 X>=0 )                                                          [     0x3ff69900040] bci=[-1,136,2723] rc=1 vc=394 vn=- li=- udi=- nc=0 flg=0x104
------------------------------
------------------------------
 n107n    (  0)  treetop                                                                              [     0x3ff69900130] bci=[-1,138,2723] rc=0 vc=394 vn=- li=- udi=- nc=1
 n106n    (  1)    icall  com/ibm/jit/JITHelpers.toUpperIntrinsicUTF16([B[BI)Z[#631  final virtual Method -128] [flags 0x20500 0x0 ] (in &GPR_0212) ()  [     0x3ff699000e0] bci=[-1,138,2723] rc=1 vc=394 vn=- li=- udi=- nc=4 flg=0x20
 n97n     (  0)      aload  java/lang/String.helpers Lcom/ibm/jit/JITHelpers;[#664  final Static] (obj1) [flags 0xa0307 0x0 ] (in &GPR_0211) (X!=0 sharedMemory )  [     0x3ff698ffe10] bci=[-1,126,2723] rc=0 vc=394 vn=- li=- udi=- nc=0 flg=0x4
 n99n     (  0)      l2a (in &GPR_0212) (X!=0 unneededConv sharedMemory )                             [     0x3ff698ffeb0] bci=[-1,130,2723] rc=0 vc=394 vn=- li=- udi=- nc=1 flg=0x404
 n515n    (  0)        lshl (in &GPR_0212) (compressionSequence )                                     [     0x3ff699080b0] bci=[-1,130,2723] rc=0 vc=394 vn=- li=- udi=- nc=2 flg=0x800
 n514n    (  0)          ==>iu2l (in GPR_0210) (X!=0 unneededConv )
 n512n    (  0)          iconst 2                                                                     [     0x3ff69907fc0] bci=[-1,130,2723] rc=0 vc=394 vn=- li=- udi=- nc=0
 n94n     (  1)      ==>newarray (in &GPR_0152) (X!=0 sharedMemory )
 n105n    (  1)      ishl (X>=0 cannotOverflow )                                                      [     0x3ff69900090] bci=[-1,137,2723] rc=1 vc=394 vn=- li=- udi=- nc=2 flg=0x1100
 n452n    (  1)        ==>iushr (in GPR_0147) (X>=0 cannotOverflow )
 n104n    (  1)        iconst 1 (X!=0 X>=0 )                                                          [     0x3ff69900040] bci=[-1,136,2723] rc=1 vc=394 vn=- li=- udi=- nc=0 flg=0x104
------------------------------

 [     0x3ff69acedb0]	LGRL    &GPR_0211,0x0000000000000000	
 [     0x3ff69acf000]	SLLG    &GPR_0212,GPR_0210,2
 [     0x3ff69acf180]	LGR     &GPR_0213,&GPR_0152	 ; LR=Clobber_eval
 [     0x3ff69acf980]	LLGFR   &GPR_0213,&GPR_0213	
 [     0x3ff69acfa70]	XGR     GPR_0214,GPR_0214	
 [     0x3ff69acfb60]	VGBM    VRF_0220,0x0
 [     0x3ff69acfc60]	VGBM    VRF_0221,0x0
 [     0x3ff69acfd60]	VGBM    VRF_0222,0x0
 [     0x3ff69acfe60]	VGBM    VRF_0223,0x0
 [     0x3ff69acff60]	VLEIH   VRF_0220,0x617a,0
 [     0x3ff69ad0060]	VLEIH   VRF_0220,0xe0f6,1
 [     0x3ff69ad0160]	VLEIH   VRF_0220,0xf8fe,2
 [     0x3ff69ad0260]	VUPLHB  VRF_0220,VRF_0220
 [     0x3ff69ad0360]	VLEIH   VRF_0221,0xa000,0
 [     0x3ff69ad0460]	VLEIH   VRF_0221,0xc000,1
 [     0x3ff69ad0560]	VLEIH   VRF_0221,0xa000,2
 [     0x3ff69ad0660]	VLEIH   VRF_0221,0xc000,3
 [     0x3ff69ad0760]	VLEIH   VRF_0221,0xa000,4
 [     0x3ff69ad0860]	VLEIH   VRF_0221,0xc000,5
 [     0x3ff69ad0960]	VLEIH   VRF_0222,0xdfdf,0
 [     0x3ff69ad0a60]	VLEIH   VRF_0222,0xb5b5,1
 [     0x3ff69ad0b60]	VLEIH   VRF_0222,0xffff,2
 [     0x3ff69ad0c60]	VUPLHB  VRF_0222,VRF_0222
 [     0x3ff69ad0d60]	VLEIH   VRF_0223,0x8000,0
 [     0x3ff69ad0e60]	VLEIH   VRF_0223,0x8000,1
 [     0x3ff69ad0f60]	VLEIH   VRF_0223,0x8000,2
 [     0x3ff69ad1060]	VLEIH   VRF_0223,0x8000,3
 [     0x3ff69ad1160]	VLEIH   VRF_0223,0x2000,4
 [     0x3ff69ad1260]	VLEIH   VRF_0223,0x2000,5
 [     0x3ff69ad1360]	VREPIH  VRF_0219,0x20
 [     0x3ff69ad1460]	LR      GPR_0215,&GPR_0213	
 [     0x3ff69ad1550]	LGR     GPR_0215,&GPR_0213	
 [     0x3ff69ad1640]	NILF    GPR_0215,15	
 [     0x3ff69ad1760]	BRC     VGNOP(0x8), Label L0080		# (Start of internal control flow)
 [     0x3ff69ad1860]	SLGFI   GPR_0215,1	
 [     0x3ff69ad1b80]	LA      GPR_0224,#672 0(GPR_0214,&GPR_0211)
 [     0x3ff69ad1c70]	VLL     VRF_0216,GPR_0215,#673 8(GPR_0224)
 [     0x3ff69ad1d70]	VSTRCHS VRF_0217,VRF_0216,VRF_0222,VRF_0223,1
 [     0x3ff69ad1e70]	BRC     MASK5(0x4), Label L0082	
 [     0x3ff69ad1f70]	VSTRCH  VRF_0217,VRF_0216,VRF_0220,VRF_0221,4
 [     0x3ff69ad2070]	VSH     VRF_0218,VRF_0216,VRF_0219
 [     0x3ff69ad2170]	VSEL    VRF_0218,VRF_0218,VRF_0216,VRF_0217
 [     0x3ff69ad2470]	LA      GPR_0225,#674 0(GPR_0214,&GPR_0212)
 [     0x3ff69ad2560]	VSTL    VRF_0218,GPR_0215,#675 8(GPR_0225)
 [     0x3ff69ad2720]	LA      GPR_0214,#676 0(GPR_0215,GPR_0214)
 [     0x3ff69ad2810]	ALGFI   GPR_0214,1	
 [     0x3ff69ad2930]	Label L0080:	
 [     0x3ff69ad2a30]	CGIJ    &GPR_0213,Label L0081,16,BM(mask=0x4), 	
 [     0x3ff69ad2b40]	SRLG    &GPR_0213,&GPR_0213,4
 [     0x3ff69ad2c40]	Label L0083:	
 [     0x3ff69ad2e00]	VL      VRF_0216,#677 8(GPR_0214,&GPR_0211)
 [     0x3ff69ad2f00]	VSTRCHS VRF_0217,VRF_0216,VRF_0222,VRF_0223,1
 [     0x3ff69ad3000]	BRC     MASK5(0x4), Label L0082	
 [     0x3ff69ad3100]	VSTRCH  VRF_0217,VRF_0216,VRF_0220,VRF_0221,4
 [     0x3ff69ad3200]	VSH     VRF_0218,VRF_0216,VRF_0219
 [     0x3ff69ad3300]	VSEL    VRF_0218,VRF_0218,VRF_0216,VRF_0217
 [     0x3ff69ad34c0]	VST     VRF_0218,#678 8(GPR_0214,&GPR_0212)
 [     0x3ff69ad3680]	LA      GPR_0214,#679 16(GPR_0214)
 [     0x3ff69ad3770]	BRCT    &GPR_0213,Label L0083	
 [     0x3ff69ad3860]	BRC     NOP(0xf), Label L0081	
 [     0x3ff69ad3960]	Label L0082:	
 [     0x3ff69ad3a60]	XGR     &GPR_0212,&GPR_0212	
 [     0x3ff69ad3d90]	ASSOCREGS
 [     0x3ff69ad3b50]	Label L0081:	
 POST:
 {AssignAny:&GPR_0211:R} {AssignAny:&GPR_0212:R} {AssignAny:&GPR_0213:R} {AssignAny:GPR_0214:R} {AssignAny:GPR_0215:R} {AssignAny:VRF_0216:R} {AssignAny:VRF_0217:R} {AssignAny:VRF_0218:R}
 {AssignAny:VRF_0219:R} {AssignAny:VRF_0220:R} {AssignAny:VRF_0221:R} {AssignAny:VRF_0222:R} {AssignAny:VRF_0223:R}	# (End of internal control flow)
```

runtime/compiler/z/codegen/J9TreeEvaluator.cpp
```diff
@@ -252,405 +252,251 @@ doubleMaxMinHelper(TR::Node *node, TR::CodeGenerator *cg, bool isMaxOp)
...
-   cursor =                  generateRSInstruction          (cg, TR::InstOpCode::getShiftRightLogicalSingleOpCode(), node, rLen, rLen, 4);                                                 iComment("reduce len to multiple of 16 (loop counter)");
+   generateS390LabelInstruction(cg, TR::InstOpCode::LABEL, node, fullVectorConversion);
 
-   // We do this so we can move src index and target index together. It also saves the need of an additional register at the cost of an extra cycle.
-   // The index will begin at 0 and increment in multiples of 16
-   cursor =                  generateRXInstruction          (cg, TR::InstOpCode::getLoadAddressOpCode(), node, rSrcVal, generateS390MemoryReference(rSrcVal, rResidue, 0, cg));                            iComment("source.val += residue");
-   cursor =                  generateRXInstruction          (cg, TR::InstOpCode::getLoadAddressOpCode(), node, rTgtVal, generateS390MemoryReference(rTgtVal, rResidue, 0 ,cg));                            iComment("tgt.val += residue");
+   generateRIEInstruction(cg, is64 ? TR::InstOpCode::CGIJ : TR::InstOpCode::CIJ, node, lengthRegister, sizeOfVector, doneLabel, TR::InstOpCode::COND_BL);
```
fjeremic: Could we use `TR::InstOpCode::getCmpImmBranchRelOpCode()` here?

runtime/compiler/z/codegen/J9TreeEvaluator.cpp
```diff
@@ -252,405 +252,251 @@ doubleMaxMinHelper(TR::Node *node, TR::CodeGenerator *cg, bool isMaxOp)
...
-   cursor =                  generateRIEInstruction         (cg, is64 ? TR::InstOpCode::CGIJ : TR::InstOpCode::CIJ, node, rLen, (int8_t) sizeOfVector, doneLabel, TR::InstOpCode::COND_BNH);                         iComment("len <= 16 ? we're done..")
-   cursor =                  generateRILInstruction         (cg, is64 ? TR::InstOpCode::ALGFI : TR::InstOpCode::ALFI, node, rResidue, 1);                                                                        iComment("re-adjust residue for VLL/VSTL (length -> index)");
+   generateVRSbInstruction(cg, TR::InstOpCode::VSTL, node, modifiedCaseVector, loadLength, generateS390MemoryReference(destRegister, addressOffset, headerSize, cg), 0);
 
-   /* ===================== Step 4.0: Setup for core-loop =====================*/
+   // Increment index by the remainder then add 1, since the loadLength contains the highest index, we must go one past that
+   generateRXInstruction(cg, TR::InstOpCode::getLoadAddressOpCode(), node, addressOffset, generateS390MemoryReference(addressOffset, loadLength, 0, cg));
+   generateRILInstruction(cg, is64 ? TR::InstOpCode::ALGFI : TR::InstOpCode::ALFI, node, addressOffset, 1);
```
fjeremic: Could we introduce a missing `TR::InstOpCode` helper for this instruction pair?

dhong44:
Depends on [eclipse/omr#2656](https://github.com/eclipse/omr/pull/2656)

fjeremic:
> Depends on eclipse/omr#2656

I don't think we need to depend on this. The `ALGFI` should go away after previous review is addressed.

runtime/compiler/z/codegen/J9TreeEvaluator.cpp
dhong44: 3bf85b1

jcl/src/java.base/share/classes/com/ibm/jit/JITHelpers.java
```diff
@@ -54,9 +54,6 @@ private JITHelpers() {
 
 	@CallerSensitive
 	public static JITHelpers getHelpers() {
-		if (Reflection.getCallerClass().getClassLoader() != null) {
-			throw new SecurityException("JITHelpers"); //$NON-NLS-1$
-		}
```
fjeremic: @andrewcraik wanted to point this one out to you. Is this safe? We need a handle to the JITHelpers object in Test_JITHelpers.java so we can do unit testing on the newly added APIs and have full control over expected inputs and outputs to test all corner cases. Unfortunately we cannot get a handle to the object without removing this or going the reflection route which seems a bit dodgy. Are you aware of any reason why this protection was put in place?

andrewcraik: I believe this is here for the same reason sun.misc.Unsafe has the same kind of check - to dissuade others from getting the instance easily. Using a trick similar to unsafe to get hold of the instance in the tests would seem better to me than exposing the API in general.

fjeremic: Cool, that makes sense. @dhong44 can you use a similar trick we use in String.java to get a hold of the JITHelpers object and revert the above deleted lines?

jcl/src/java.base/share/classes/java/lang/String.java
```diff
@@ -2719,15 +2718,26 @@ public String toLowerCase(Locale locale) {
...
-		if (StrHWAvailable() && language == "en") { //$NON-NLS-1$
-			String output = new String(lengthInternal());
-			if (toLowerHWOptimized(output))
-				return output;
-		}
+		int sLength = lengthInternal();
 
 		if (enableCompression && (null == compressionFlag || coder == LATIN1)) {
+
+			if (helpers.supportsIntrinsicCaseConversion() && language == "en") { //$NON-NLS-1$
```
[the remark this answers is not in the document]
fjeremic: I believe so. The interpreter will enter the if block immediately above on line 2728 so we will have profiling data for that path. The JIT will fold away the call to `helpers.supportsIntrinsicCaseConversion()` as a constant at compile time and the flow algorithm in the JIT should propagate the block frequencies to the if at line 2725 to be equal to that of the preceding if. 

Just to be cautious though @dhong44 can we collect a log and verify the block frequencies we compute for the block in which we inline the intrinsic? Given your expertise @andrewcraik is my reasoning here valid?

test/functional/Java8andUp/playlist.xml
```diff
@@ -1042,7 +1042,7 @@
 	<test>
 		<testCaseName>JCL_Test_JIT_Helper</testCaseName>
 		<command>$(JAVA_COMMAND) $(JVM_OPTIONS) --add-exports=java.base/com.ibm.jit=ALL-UNNAMED \
-	-Xbootclasspath/a:$(Q)$(TEST_RESROOT)$(D)TestResources.jar$(Q) \
+	--add-opens=java.base/com.ibm.jit=ALL-UNNAMED -Xbootclasspath/a:$(Q)$(TEST_RESROOT)$(D)TestResources.jar$(Q) \
```
fjeremic: Are you sure this will work for Java 8? I suppose we can test via a build. Also while we're in the area can you rename this test case name to "JCL_Test_JITHelpers" to make it searchable since the test file is Test_JITHelpers.java and the class name is JITHelpers. I would like to be able to find this line in the future by grepping for the class name.

jcl/src/java.base/share/classes/java/lang/String.java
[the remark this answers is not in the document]
andrewcraik: The key question is when the supportsIntrinsicCaseConversion is folded. It looks to be in ilgen so the flow for IProfiling will have a chance to happen - other profilers do not do a flow. The key thing is to check the frequency on the blocks with the custom implementation after cold block marking. I think a log is the best way to verify this is working right.

test/functional/Java8andUp/playlist.xml
dhong44: This test is only for SE90 and up, SE80 has it's own test above. I don't think the `--add-opens` is necessary or is even recognized in Java 8

test/functional/Java8andUp/src/org/openj9/test/com/ibm/jit/Test_JITHelpers.java
```diff
@@ -53,4 +54,194 @@ public static void test_getSuperclass() {
...
+	/**
+	 * @tests com.ibm.jit.JITHelpers#(char[], char[], int)
+	 */
+    public static void test_toUpperIntrinsicUTF16() {
+		char[] buffer;
+
+		for (int j : edgecaseLengths){
+			buffer = new char[j];
+			if (helpers.toUpperIntrinsicUTF16(Arrays.copyOfRange(lowercaseUTF16Char, 0, j), buffer, j * 2)) {
+				AssertJUnit.assertTrue("UTF16 JITHelper upper case conversion with char arrays of " + j + " letters failed",
```
llxia: Instead of using `org.testng.AssertJUnit`, we should directly use `org.testng.Assert`  http://static.javadoc.io/org.testng/testng/6.8.17/org/testng/Assert.html
`org.testng.AssertJUnit` is used when we port/migrate code that uses JUnit's asserts.

jcl/src/java.base/share/classes/java/lang/String.java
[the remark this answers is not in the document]
dhong44:
```
n1398n    BBStart <block_125> (freq 5002)                                                     [     0x3ff4315a4b0] bci=[10,116,2664] rc=0 vc=257 vn=- li=- udi=- nc=0
n1432n    treetop                                                                             [     0x3ff4315af50] bci=[10,122,2664] rc=0 vc=257 vn=- li=- udi=- nc=1
n1431n      newarray  jitNewArray[#176  helper Method] [flags 0x400 0x0 ] (skipZeroInit sharedMemory )  [     0x3ff4315af00] bci=[10,122,2664] rc=2 vc=257 vn=- li=- udi=- nc=2 flg=0x8000
n1429n        ishl                                                                            [     0x3ff4315ae60] bci=[10,121,2664] rc=1 vc=257 vn=- li=- udi=- nc=2
n1426n          iload  <auto slot 3>[#796  Auto] [flags 0x3 0x0 ]                             [     0x3ff4315ad70] bci=[10,116,2664] rc=1 vc=257 vn=- li=- udi=- nc=0
n1428n          iloadi  java/lang/String.coder B[#802  final Shadow +8] [flags 0xa0601 0x0 ]  [     0x3ff4315ae10] bci=[10,118,2664] rc=1 vc=257 vn=- li=- udi=- nc=1
n1427n            aload  <temp slot 3>[#786  Auto] [flags 0x7 0x0 ] (X!=0 sharedMemory )      [     0x3ff4315adc0] bci=[10,117,2664] rc=1 vc=257 vn=- li=- udi=- nc=0 flg=0x4
n1430n        iconst 8   ; array type is byte                                                 [     0x3ff4315aeb0] bci=[10,122,2664] rc=1 vc=257 vn=- li=- udi=- nc=0
n1433n    astore  <auto slot 4>[#810  Auto] [flags 0x7 0x0 ]                                  [     0x3ff4315afa0] bci=[10,124,2664] rc=0 vc=257 vn=- li=- udi=- nc=1
n1431n      ==>newarray
n1438n    compressedRefs                                                                      [     0x3ff4315b130] bci=[10,130,2666] rc=0 vc=257 vn=- li=- udi=- nc=2
n1436n      aloadi  java/lang/String.value [B[#804  final Shadow +4] [flags 0x400a0607 0x0 ]  [     0x3ff4315b090] bci=[10,130,2666] rc=2 vc=257 vn=- li=- udi=- nc=1
n1435n        aload  <temp slot 3>[#786  Auto] [flags 0x7 0x0 ] (X!=0 sharedMemory )          [     0x3ff4315b040] bci=[10,129,2666] rc=1 vc=257 vn=- li=- udi=- nc=0 flg=0x4
n1437n      lconst 0                                                                          [     0x3ff4315b0e0] bci=[10,130,2666] rc=1 vc=257 vn=- li=- udi=- nc=0
n1444n    NULLCHK on n1434n [#30]                                                             [     0x3ff4315b310] bci=[10,138,2666] rc=0 vc=257 vn=- li=- udi=- nc=1
n1443n      icall  com/ibm/jit/JITHelpers.toLowerIntrinsicUTF16([B[BI)Z[#815  final virtual Method -136] [flags 0x20500 0x0 ] ()  [     0x3ff4315b2c0] bci=[10,138,2666] rc=3 vc=257 vn=- li=- udi=- nc=4 flg=0x20
n1434n        aload  java/lang/String.helpers Lcom/ibm/jit/JITHelpers;[#799  final Static] [flags 0xa0307 0x0 ]  [     0x3ff4315aff0] bci=[10,126,2666] rc=1 vc=257 vn=- li=- udi=- nc=0
n1436n        ==>aloadi
n1439n        aload  <auto slot 4>[#810  Auto] [flags 0x7 0x0 ]                               [     0x3ff4315b180] bci=[10,133,2666] rc=1 vc=257 vn=- li=- udi=- nc=0
n1442n        imul                                                                            [     0x3ff4315b270] bci=[10,137,2666] rc=1 vc=257 vn=- li=- udi=- nc=2
n1440n          iload  <auto slot 3>[#796  Auto] [flags 0x3 0x0 ]                             [     0x3ff4315b1d0] bci=[10,135,2666] rc=1 vc=257 vn=- li=- udi=- nc=0
n1441n          iconst 2                                                                      [     0x3ff4315b220] bci=[10,136,2666] rc=1 vc=257 vn=- li=- udi=- nc=0
n1445n    istore  <pending push temp 0>[#793  Auto] [flags 0x3 0x800 ]                        [     0x3ff4315b360] bci=[10,138,2666] rc=0 vc=257 vn=- li=- udi=- nc=1
```

test/functional/Java8andUp/src/org/openj9/test/com/ibm/jit/Test_JITHelpers.java
dhong44: 4a4b2df

test/functional/Java8andUp/src/org/openj9/test/java/lang/Test_String.java
```diff
@@ -1377,6 +1405,34 @@ public void test_toUpperCase() {
 
 		AssertJUnit.assertTrue("Wrong conversion", "\u00df".toUpperCase().equals("SS"));
 
+		// Long strings and strings of lengths near multiples of 16 bytes to test vector HW acceleration edge cases.
+		AssertJUnit.assertTrue("toUpperCase case conversion failed on a long string",
```
llxia: Same here. `org.testng.Assert` should be used.

dhong44:
I ran a bumble bench that called `toLower` on strings of random length between 0 and 1000. With vectorized case conversion the score was about 30 times higher than without.

test/functional/Java8andUp/src/org/openj9/test/java/lang/Test_String.java
dhong44: 09078d4

jcl/src/java.base/share/classes/java/lang/String.java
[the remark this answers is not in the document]
fjeremic: Seems to be working as intended. I validated the log with @dhong44. We seem to have good frequencies on the blocks containing the intrinsic calls.

runtime/compiler/z/codegen/J9TreeEvaluator.cpp
```diff
@@ -252,405 +252,256 @@ doubleMaxMinHelper(TR::Node *node, TR::CodeGenerator *cg, bool isMaxOp)
...
-
-   // We already set start ICF at the branch after the TMLH so don't do it here, to prevent nested ICF issues
-   if (is64)
-      cursor->setStartInternalControlFlow();
-
-   // VLL and VSTL are provided with an index not with a count (length) so we need to offset by 1.
-   cursor =                 generateRILInstruction          (cg, TR::InstOpCode::getSubtractLogicalImmOpCode(), node, rResidue, 1);                                                                         iComment("adjust residue for VLL/VSTL (length -> index)");
-
-   /* ===================== Step 3: Residue processing =====================*/
+   generateRRInstruction(cg, TR::InstOpCode::getLoadRegOpCode(), node, loadLength, lengthRegister);
```
fjeremic: This seems unnecessary given the load above. Can we remove this? Also shouldn't `lengthRegister` always a 32-bit value? Shouldn't we always use an `LR` here? Similarly to above, why can't `loadLength` also be a 32-bit register?

runtime/compiler/z/codegen/J9TreeEvaluator.cpp
```diff
@@ -252,405 +252,256 @@ doubleMaxMinHelper(TR::Node *node, TR::CodeGenerator *cg, bool isMaxOp)
...
-   // Bail out if string is larger than INT_MAX32/2 since # string to # byte conversion will cause overflow.
-   // We could have used 64 bit reg instructions and reg for 31bit JVM with proper sign/zero extension,
-   // but using this keeps the icache footprint down by a little bit. Also enables us to use the HPRs.
-   if (!is64)
+   if (is64)
       {
-      cursor =               generateRIInstruction          (cg, TR::InstOpCode::TMLH, node, rLen, (uint16_t) 0x8000);
-      cursor =               generateS390BranchInstruction  (cg, TR::InstOpCode::BRC , TR::InstOpCode::COND_MASK2, node, handleInvalidChars);
-      cursor->setStartInternalControlFlow();
+      generateRRInstruction(cg, TR::InstOpCode::LLGFR, node, lengthRegister, lengthRegister);
```
fjeremic: Why exactly is this needed? Looking at the instruction selection from https://github.com/eclipse/openj9/pull/2098#issuecomment-395776065 I see that all uses of this register can be made in a 32-bit context. We seem to be forcing it to be 64-bit here however.

test/functional/Java8andUp/src/org/openj9/test/com/ibm/jit/Test_JITHelpers.java
```diff
@@ -53,4 +55,194 @@ public static void test_getSuperclass() {
...
+
+		for (int j : edgecaseLengths) {
+			buffer = new byte[j];
+			if (helpers.toLowerIntrinsicLatin1(Arrays.copyOfRange(uppercaseLatin1Byte, 0, j), buffer, j)) {
+				Assert.assertTrue(Arrays.equals(buffer, Arrays.copyOfRange(lowercaseLatin1Byte, 0, j)),
+					"Latin 1 JITHelper lower case conversion with byte arrays of " + j + " letters failed");
+			}
+        	buffer.getClass();
+		}
+	}
```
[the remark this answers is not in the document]
dhong44: I don't think so, since the result of that method when compiled is dependent on the machine. There's no way we could really test that it returns the correct result.

runtime/compiler/z/codegen/J9TreeEvaluator.cpp
dhong44: cf269b9

runtime/compiler/z/codegen/J9TreeEvaluator.cpp
dhong44: cf269b9

fjeremic:
Jenkins test sanity

fjeremic:
Transferring this to @joransiu as I'll be away for the next two weeks.

PR #2101, opened 2018-06-06
Print Memory Stats

dsouzai:
Added the printing of JIT memory stats at shutdown if the `TR_PrintPersistentMem` env var is set, and every 5 minutes in the vlog if `verbose={jitMemory}` is set.

Depends on https://github.com/eclipse/omr/pull/2542

### DIFF
--- a/runtime/compiler/control/CompilationThread.cpp
+++ b/runtime/compiler/control/CompilationThread.cpp
@@ -2713,6 +2713,13 @@ void TR::CompilationInfo::stopCompilationThreads()
          fprintf(stderr, "Allocated memory for profile info = %d KB\n", getJProfilerThread()->getProfileInfoFootprint()/1024);
       }
 
+   static char * printPersistentMem = feGetEnv("TR_PrintPersistentMem");
+   if (printPersistentMem)
+      {
+      if (trPersistentMemory)
+         trPersistentMemory->printMemStats();
+      }
+
    TR_DataCacheManager::getManager()->printStatistics();
 
    bool aotStatsEnabled = TR::Options::getAOTCmdLineOptions()->getOption(TR_EnableAOTStats);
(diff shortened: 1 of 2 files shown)


dsouzai:
@mpirvu  for your review.

runtime/compiler/control/CompilationThread.cpp
```diff
@@ -2711,6 +2711,9 @@ void TR::CompilationInfo::stopCompilationThreads()
 
       if (getJProfilerThread())
          fprintf(stderr, "Allocated memory for profile info = %d KB\n", getJProfilerThread()->getProfileInfoFootprint()/1024);
+
+      if (trPersistentMemory)
```
mpirvu: These statistics are probably useful only when tracking a memory leak. Also, the list of memory objects is very long.
For these two reasons I would like the printout to be enabled by another environment variable.

dsouzai:
Changes made.

andrewcraik:
@dsouzai a rebase forward should clear up the travis failure (I hope) so we can get this tested and merged.

dsouzai:
Because this depends on https://github.com/eclipse/omr/pull/2542 until that code is merged in openj9-omr, I think travis will still be broken

fjeremic:
Jenkins test sanity depends eclipse/omr#2542

AdamBrousseau:
Jenkins test sanity depends eclipse/omr#master

dsouzai:
@vijaysun-omr This PR should be ok to merge now. I can rebase the changes if you'd like the travis build to run again, but given that sanity tests passed, it  doesn't seem necessary (given how small this change is). The windows failure is unrelated and caused by something else (other PR builds are also failing on windows).

dsouzai:
Windows issue caused by https://github.com/eclipse/openj9/issues/2129

dsouzai:
Rebased onto latest master for clean travis build.

dsouzai:
> Rebased onto latest master for clean travis build.

So much for that... failing now because https://github.com/eclipse/omr/pull/2490 isn't in openj9-omr yet

dsouzai:
@vijaysun-omr This should be good to merge now

PR #2103, opened 2018-06-06
WIP: backOffFromSafePoint in CS

amicic:
Fix a problem where a thread wants to acquire safePointVMaccess, while a
Concurrent Scavenger (CS) cycle is about to start (exclusive access has
already been requested). An example is a thread wanting to acquire
safePointVMaccess is to do class replacement heap walk.

In the transition from the first STW increment to the concurrent phase
of a CS cycle, the master thread would appear that it still holds VM
access (and prevents any other party acquiring safePoint or exclusive VM
access), but in reality the the other party would be oblivious of this
fact. 

The fix is for Master GC thread to call backOffFromSafePoint at the end
of first STW increment, which will transition itself from a Safe Point
state (which has been created when Safe Point access request has been
raised, since Master thread was inactive/safe at that point) to an
unsafe state (and holding VM assess). That transition, will make the
thread that requested safePoint VM access become aware of one more
thread holding VM access and wait for it for respond.

The end effect is, the safePoint VM access will be obatined only when
master GC thread releases VM access, hence there are no active GC
threads mutating the heap and potentially preventing from safely walking
it.


amicic:

PR #2106, opened 2018-06-06
WIP: Tool configuration

keithc-ca:
* use C/C++ compilers more consistently
* remove 'use_ld_to_link' property
* adjust use of 'isCPlusPlus' option
-- it should never be conditional
--  it has no meaning for static libraries


runtime/makelib/targets.mk.aix.inc.ftl
```diff
@@ -99,31 +99,18 @@ ifdef I5_VERSION
...
-  ifeq ($(j9vm_env_data64),1)
-    UMA_DLL_LINK_FLAGS+=-X64
-  else
-    UMA_DLL_LINK_FLAGS+=-X32
-  endif
-  UMA_DLL_LINK_FLAGS+=-p 0 -brtl -G -bernotok -bnoentry -lm
-  UMA_DLL_LINK_FLAGS+=-E $(UMA_TARGET_NAME).exp
+  UMA_DLL_LINK_FLAGS += -q32
 endif
+UMA_DLL_LINK_FLAGS += -qmkshrobj=0 -Wl,-brtl -Wl,-G -Wl,-bernotok -Wl,-bnoentry -lm
```
pshipton: Why is `-p 0` removed?

runtime/makelib/targets.mk.aix.inc.ftl
```diff
@@ -99,31 +99,18 @@ ifdef I5_VERSION
...
+UMA_DLL_LINK_FLAGS += -Wl,-bE:$(UMA_TARGET_NAME).exp
```
pshipton: Why `-bE:` instead of -E? Was -E a mistake?

runtime/makelib/targets.mk.aix.inc.ftl
keithc-ca: `-p 0` is an option understood by `makeC++SharedLib_r`; this PR changes aix*.spec to use the C compiler: the equivalent is `-qmkshrobj=0`.

runtime/makelib/targets.mk.aix.inc.ftl
keithc-ca: Like above, as we switch to linking with the C compiler, we have to speak its language.
The `-E` option of `makeC++SharedLib_r` maps to `-Wl,-bE:$(UMA_TARGET_NAME).exp`.

keithc-ca:
### UMA Tools

UMA uses several kinds of tools via these properties:

| UMA Property             | Usage                                                    |
| :-----------             | :----                                                    |
| uma_make_cmd_cc          | The C compiler command                                   |
| uma_make_cmd_cxx         | The C++ compiler command                                 |
| uma_make_cmd_interp_gcc  | The BytecodeInterpreter compiler command                 |
| uma_make_cmd_ppc_gcc_cxx | The BytecodeInterpreter compiler command (for PPC)       |
| uma_make_cmd_dll_ld      | The linker for shared objects (with no C++ dependencies) |
| uma_make_cmd_cxx_dll_ld  | The linker for shared objects (with C++ dependencies)    |
| uma_make_cmd_exe_ld      | The linker for executables (with no C++ dependencies)    |
| uma_make_cmd_cxx_exe_ld  | The linker for executables (with C++ dependencies)       |

Currently, these tools are essentially independent.

The following tools are discovered at OpenJDK configuration time:

| Tool Name | Usage                    |
| :-------- | :----                    |
| CC        | The C compiler command   |
| CXX       | The C++ compiler command |

The plan is to adjust the OpenJ9 *.spec and module.xml files so this mapping consistently applies:

| UMA Property             | Value  |
| :-----------             | :----  |
| uma_make_cmd_cc          | $(CC)  |
| uma_make_cmd_cxx         | $(CXX) |
| uma_make_cmd_interp_gcc  | $(CXX) |
| uma_make_cmd_ppc_gcc_cxx | $(CXX) |

All of the UMA properties specifying linker commands will be adjusted to use `$(CC)` or `$(CXX)` as appropriate.

To accomplish this:
* each of `uma_make_cmd_cc`, `uma_make_cmd_dll_ld` and `uma_make_cmd_exe_ld` must name a C compiler (not a C++ compiler), otherwise linking may create unwanted references to C++ support libraries
* none of the properties can make unreasonable assumptions about the capabilities of tools named
by other properties (e.g. $(CC) can't be expected to know anything about C++).
* artifacts in module.xml files that specify the UMA option `isCPlusPlus` should assume that implies a dependency on compiler C++ support libraries

runtime/makelib/targets.mk.aix.inc.ftl
```diff
@@ -99,31 +102,19 @@ ifdef I5_VERSION
...
-    UMA_DLL_LINK_FLAGS+=-X64
-  else
-    UMA_DLL_LINK_FLAGS+=-X32
-  endif
-  UMA_DLL_LINK_FLAGS+=-p 0 -brtl -G -bernotok -bnoentry -lm
-  UMA_DLL_LINK_FLAGS+=-E $(UMA_TARGET_NAME).exp
+  UMA_DLL_LINK_FLAGS += -q32
 endif
+# Link options like '-brtl', '-G', etc. must be prefixed by '-Wl,' to make xlc_r pass them through.
+UMA_DLL_LINK_FLAGS += -qmkshrobj -Wl,-brtl -Wl,-G -Wl,-bernotok -Wl,-bnoentry -lm
```
DanHeidinga: Did the ` -bmap:$(UMA_TARGET_NAME).map` get deliberately dropped (why) or accidentally lost?

keithc-ca: I'll restore the map creation.

DanHeidinga:
As discussed today, we need to ensure that this change doesn't cause us to link the c++ standard library.

keithc-ca:
I think this is good to go. I wrote a script to use ldd to compare the dependencies of executables and shared libraries before and after this change: no extra references to C++ libraries.

runtime/makelib/targets.mk.aix.inc.ftl
```diff
@@ -99,41 +102,41 @@ ifdef I5_VERSION
...
-  ifeq ($(j9vm_env_data64),1)
-    UMA_DLL_LINK_FLAGS += -b64
-  else
-    UMA_DLL_LINK_FLAGS += -b32
-  endif
-  UMA_DLL_LINK_FLAGS += -G -bnoentry -bernotok
-  UMA_DLL_LINK_FLAGS += -bmap:$(UMA_TARGET_NAME).map
-  UMA_DLL_LINK_FLAGS += -bE:$(UMA_TARGET_NAME).exp
-  UMA_LINK_LIBRARIES += -lc_r -lC_r -lm -lpthread
+UMA_SYS_LINK_PATH := -L/usr/lib/threads
```
pshipton: Please add a comment to explain the origins of this

keithc-ca:
Comment added as requested.

pshipton:
jenkins test sanity

pshipton:
jenkins test sanity xlinux jdk10

pshipton:
jenkins test sanity xlinux jdk10

pshipton:
jenkins test sanity xlinuxlargeheap jdk10

keithc-ca:
Who can help fix this?
```
Error: java.io.IOException: No space left on device
```

pshipton:
> Who can help fix this?

Created #2161

DanHeidinga:
> As discussed today, we need to ensure that this change doesn't cause us to link the c++ standard library.

@keithc-ca Have you validated that this doesn't link the C++ stdlib anywhere we didn't before?

keithc-ca:
@DanHeidinga Yes, I verified that there are no new references to C++ shared libraries.
See https://github.com/eclipse/openj9/pull/2106#issuecomment-396396414.

PR #2116, opened 2018-06-07
Update labels in line with recent changes

SueChaplain:
doc:whatsnew and doc:releasenotes have been
replaced with doc:externals.

Also added a link to an issue template in the
OpenJ9 docs repo for external changes.

Tidy up doc README.md to cover other types
of doc now in this directory. E.g.
Release notes.

### DIFF
--- a/doc/README.md
+++ b/doc/README.md
@@ -24,13 +24,17 @@ SPDX-License-Identifier: EPL-2.0 OR Apache-2.0 OR GPL-2.0 WITH Classpath-excepti
 Documentation
 =============
 
-This directory contains documentation about various components in OpenJ9
-
-* **`compiler`**: Documentation about the compiler technology, codenamed
-  Testarossa. 
+This directory contains the following types of documentation:
 
 * **`processes`**: Documentation about the processes used in OpenJ9.
 
-Much documentation about OpenJ9 is best considered in tandem with the
+* **`release-notes`**: Release notes for different releases of OpenJ9.
+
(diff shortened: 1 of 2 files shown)


doc/README.md
```diff
@@ -24,13 +24,17 @@ SPDX-License-Identifier: EPL-2.0 OR Apache-2.0 OR GPL-2.0 WITH Classpath-excepti
 Documentation
 =============
 
-This directory contains documentation about various components in OpenJ9
+This directory contains the following types of documentation:
+
+* **`processes`**: Documentation about the processes used in OpenJ9.
```
pshipton: any reason not to put the categories in alphabetical order?

pshipton: If we want to have processes and release-notes first, followed by everything else in alphabetical order, perhaps there should be a division between the two categories of items.

SueChaplain: I was just tidying up to reflect what is there at the moment. I think there is more content that will end up here too. E.g. There's an action item in the community minutes to move the build instructions in here too. I avoided touching those for now as we need to coordinate/update the link at the website.

Maybe you can just let this through for now and we can come up with a better structure when the build instructions move. Seems we should separate the "task oriented" items from the "reference" items ... at least that makes sense to me.

SueChaplain: @pshipton Can this be merged please? The README is a lot more accurate than it was and we can deal with the order of items when I move the build instructions here.

PR #2121, opened 2018-06-07
VarHandle performance

liqunl:
This commit contains the following change:
 - skip HCR guards for classes/methods that are implementation details
   of VarHandle and MethodHandle.
 - Replace the call to MethodHandle.asType with the clone of the
   MethodHandle node if the call is going to return the MethodHandle
   object passed in. This will trigger targeted inlining for VarHandle
   calls.
 - Eliminate calls to DirectHandle.nullCheckIfRequired and
   PrimitiveHandle.initializeClassIfRequired when they're not needed.


liqunl:
depends https://github.com/eclipse/omr/pull/2643

liqunl:

runtime/compiler/optimizer/J9ValuePropagation.cpp
```diff
@@ -639,6 +639,165 @@ J9::ValuePropagation::constrainRecognizedMethod(TR::Node *node)
...
+               if (!asType.hasVMAccess())
+                  break;
+               uintptrj_t mhObject = comp()->fej9()->getStaticReferenceFieldAtAddress((uintptrj_t)mhLocation);
+               uintptrj_t mtInMh = comp()->fej9()->getReferenceFieldAtAddress(mhObject + mtOffset);
+               uintptrj_t mtObject = comp()->fej9()->getStaticReferenceFieldAtAddress((uintptrj_t)mtLocation);
+
+               if (trace())
+                  traceMsg(comp(), "mhLocation %p mtLocation %p mhObject %p mtOffset %d mtInMh %p mtObject %p\n", mhLocation, mtLocation, mhObject, mtOffset, mtInMh, mtObject);
+
+               // AOT doesn't relocate/validate objects, which is why this transformation is disabled under AOT
```
andrewcraik: Where's the code for this comment? I don't see anything obvious about AOT here

andrewcraik:
The gist of the changes is quite reasonable - @jdmpapin can you please take a look at the VP changes since it is an area you are familiar with?

andrewcraik:
@liqunl while that is happening can you rebase forward to get a good travis build?

runtime/compiler/optimizer/J9ValuePropagation.cpp
liqunl: There is an if statement right before the the switch statement to bail out if we're in AOT. There are two switch statements in this function, one for transformations that work in AOT, one for transformations that won't work in AOT. I can merge the two switch statements and add the AOT check before optimizations that don't have AOT support.

runtime/compiler/optimizer/J9ValuePropagation.cpp
```diff
@@ -639,6 +639,165 @@ J9::ValuePropagation::constrainRecognizedMethod(TR::Node *node)
...
+               if (modifierOffset < 0)
+                  break;
+               uintptrj_t* mhLocation = getObjectLocationFromConstraint(mhConstraint);
+
+               TR::VMAccessCriticalSection nullCheckIfRequired(comp(),
+                                                        TR::VMAccessCriticalSection::tryToAcquireVMAccess);
+               if (!nullCheckIfRequired.hasVMAccess())
+                  break;
+               uintptrj_t mhObject = comp()->fej9()->getStaticReferenceFieldAtAddress((uintptrj_t)mhLocation);
+               int32_t modifier = comp()->fej9()->getInt32FieldAt(mhObject, modifierOffset);
```
jdmpapin: This access should be safe because a valid `modifierOffset` was found from the class of `mhObject` :+1: 

Here, if the code is on a dead path, that class might not be `DirectHandle`, but it contains an integer-typed field at `modifierOffset`, and it doesn't matter whether or not the call is removed.

runtime/compiler/optimizer/J9ValuePropagation.cpp
```diff
@@ -639,6 +639,165 @@ J9::ValuePropagation::constrainRecognizedMethod(TR::Node *node)
...
+               {
+               uint32_t defcOffset = J9VMJAVALANGINVOKEPRIMITIVEHANDLE_DEFC_OFFSET(comp()->fej9()->vmThread());
+               uintptrj_t* mhLocation = getObjectLocationFromConstraint(mhConstraint);
+
+               TR::VMAccessCriticalSection initializeClassIfRequired(comp(),
+                                                        TR::VMAccessCriticalSection::tryToAcquireVMAccess);
+               if (!initializeClassIfRequired.hasVMAccess())
+                  break;
+               uintptrj_t mhObject = comp()->fej9()->getStaticReferenceFieldAtAddress((uintptrj_t)mhLocation);
+               uintptrj_t defc = comp()->fej9()->getReferenceFieldAtAddress(mhObject + defcOffset);
```
jdmpapin: Like above, `mhObject` needs to be a `PrimitiveHandle`

runtime/compiler/optimizer/J9ValuePropagation.cpp
```diff
@@ -639,6 +639,165 @@ J9::ValuePropagation::constrainRecognizedMethod(TR::Node *node)
...
+               uintptrj_t* mhLocation = getObjectLocationFromConstraint(mhConstraint);
+               uintptrj_t* mtLocation = getObjectLocationFromConstraint(mtConstraint);
+               uint32_t mtOffset = J9VMJAVALANGINVOKEMETHODHANDLE_TYPE_OFFSET(comp()->fej9()->vmThread());
+
+               TR::VMAccessCriticalSection asType(comp(),
+                                                           TR::VMAccessCriticalSection::tryToAcquireVMAccess);
+               if (!asType.hasVMAccess())
+                  break;
+               uintptrj_t mhObject = comp()->fej9()->getStaticReferenceFieldAtAddress((uintptrj_t)mhLocation);
+               uintptrj_t mtInMh = comp()->fej9()->getReferenceFieldAtAddress(mhObject + mtOffset);
```
jdmpapin: We should check before this access that `mhObject` really is an instance of `MethodHandle`. If we're (unknowingly) looking at a dead path, then it's possible to have some other type of object instead.

liqunl:
depends https://github.com/eclipse/omr/pull/2670
@andrewcraik @jdmpapin Made some changes according to your suggestions, will merge the two switch statements after this round of review. The OMR change is to handle the PassThrough node.

runtime/compiler/optimizer/J9ValuePropagation.cpp
```diff
@@ -639,6 +640,164 @@ J9::ValuePropagation::constrainRecognizedMethod(TR::Node *node)
...
+            TR::VPConstraint* mhConstraint = getConstraint(mh, mhConstraintGlobal);
+            TR::VPConstraint* receiverConstraint = getConstraint(receiver, receiverConstraintGlobal);
+
+            J9VMThread* vmThread = comp()->fej9()->vmThread();
+            TR_OpaqueClassBlock* directHandleClass = calledMethod->classOfMethod();
+            if (!mhConstraint
+                || !mhConstraint->isNonNullObject()
+                || !mhConstraint->getClass()
+                || comp()->fej9()->isInstanceOf(mhConstraint->getClass(), directHandleClass, true, true) != TR_yes)
+               break;
```
jdmpapin: This always requires a constraint on `mh` to guarantee it's a (non-null) `DirectHandle`, even when `receiverConstraint->isNonNullObject()` suffices to eliminate the call without reading `final_modifiers`. Was that intentional?

I think it's still possible to remove the call even if `mh` could be null. The call would be under a `NULLCHK`, and replacing it with `PassThrough` should leave that `NULLCHK` in place, now checking `mh` without calling.

runtime/compiler/optimizer/J9ValuePropagation.cpp
```diff
@@ -639,6 +640,164 @@ J9::ValuePropagation::constrainRecognizedMethod(TR::Node *node)
...
+
+               // AOT doesn't relocate/validate objects, which is why this transformation is disabled under AOT
+               if (mtInMh == mtObject)
+                  {
+                  if (!performTransformation(comp(), "%sTransforming %s on node %p to a PassThrough with child %p\n", OPT_DETAILS, signature, node, mh))
+                     break;
+
+                  anchorAllChildren(node, _curTree);
+                  node->removeAllChildren();
+                  node = TR::Node::recreateWithoutProperties(node, TR::PassThrough, 1, mh);
```
jdmpapin: This looks just like `transformCallNodeToPassThrough()`, which is called in the other new cases. Might it make sense to relax the assertion in that function somehow instead of duplicating its code here?

liqunl: I have an OMR change to extend the usage of that function. Please check https://github.com/eclipse/omr/pull/2679. Will make the change in openj9 shortly.

liqunl:
@jdmpapin commit has been updated and is ready for review.

andrewcraik:
Jenkins test sanity

liqunl:
@andrewcraik This change depends on https://github.com/eclipse/omr/pull/2670 and https://github.com/eclipse/omr/pull/2679 which are pending on your review.

andrewcraik:
Thanks for the reminder - I have commented on those two OMR PRs and @'d an OMR committer to further review and merge.

liqunl:
@andrewcraik The two OMR PRs this PR depends on have been merged. The windows failures are due to https://github.com/eclipse/openj9/issues/2129.

liqunl:
Will merge the two switch statement in a different PR along with other VarHandle changes.

Issue #2124, opened 2018-06-07
Increase highly polymorphic callsite threshold


IBMJimmyk:
When running an internal appserver benchmark it was found that changing the threshold for considering a callsite as being highly polymorphic would improve throughput.

Currently a callsite is considered highly polymorphic if there are 2 or more implementors. The proposed change will increase the threshold to 3. This change lead to changes in inlining decisions at runtime which in turn lead to the increase in throughput.

PR #2125, opened 2018-06-07
Increase highly polymorphic callsite threshold

IBMJimmyk:
This change increases the threshold for considering a callsite as highly
polymorphic from 2 implementors to 3. This change was found to improve
the inlining decisions in an internal appserver benchmark which lead to
an improvement in throughput.

Issue: #2124

### DIFF
--- a/runtime/compiler/env/PersistentCHTable.cpp
+++ b/runtime/compiler/env/PersistentCHTable.cpp
@@ -433,7 +433,7 @@ TR_PersistentCHTable::findSingleInterfaceImplementer(
    }
 
 bool
-TR_PersistentCHTable::hasTwoOrMoreCompiledImplementors(
+TR_PersistentCHTable::hasThreeOrMoreCompiledImplementors(
    TR_OpaqueClassBlock * thisClass, int32_t cpIndex, TR_ResolvedMethod * callerMethod, TR::Compilation * comp, TR_Hotness hotness, bool locked)
    {
    if (comp->getOption(TR_DisableCHOpts))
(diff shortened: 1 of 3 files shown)


runtime/compiler/env/PersistentCHTable.cpp
```diff
@@ -445,8 +445,8 @@ TR_PersistentCHTable::hasTwoOrMoreCompiledImplementors(
    TR_PersistentClassInfo * classInfo = findClassInfoAfterLocking(thisClass, comp, true);
    if (!classInfo) return false;
 
-   TR_ResolvedMethod *implArray[2];
-   return TR_ClassQueries::collectCompiledImplementorsCapped(classInfo,implArray,2,cpIndex,callerMethod,comp,hotness,locked) == 2;
```
andrewcraik: The method is called hasTwoOrMoreCompiledImplementors so the name needs to change if it is becoming 3 or more :)

IBMJimmyk: Good catch. I'm fixing that right now.

andrewcraik:
Jenkins test sanity

IBMJimmyk:
The following numbers are based on 6 baseline iterations and 6 new build iterations that are interleaved.

HiBench measures the time spent to finish a task, so the smaller the score, the better the performance. For this chart, a negative Relative score means a higher time spent. Positive is good, negative is bad.

Benchmark | Relative score | Standard deviation
-- | -- | --
Pagerank | 1.18% | 1.69%
Terasort | 6.82% | 3.46%
Wordcount | -1.12% | 4.14%
Sort | -1.19% | 2.64%
Sleep | 0.24% | 0.09%

The baseline score for Terasort ended with two unusually long runs (out of 6 total). If they are removed, the improvement drops to +1.58%. Other than that, the results are within the noise margin so there's no negative effect on HiBench.

andrewcraik:
Given the performance validation I am fine with this change. It would make sense to be a little less sensitive before declaring a call site to be highly polymorphic. Failures on Win JDK8 are unrelated to this change so this is good to go.

PR #2126, opened 2018-06-07
Add frontend method getVFT entry to read the virtual function table

nwoeanhinnogaehr:
Currently, VP directly reads this from the class pointer (in OMR). We would like to avoid having the optimizer directly read fields from classes. This change is required for JITaaS support.

Signed-off-by: Noah Weninger <noah.weninger@ibm.com>


andrewcraik:
Looks good to me.
Jenkins test sanity

andrewcraik:
the wind JDK8 failures are unrelated to this change; tests passed

PR #2127, opened 2018-06-07
Call OMR::CodeGenPhaseConnector in the extensible class on Z

fjeremic:
The CodeGenPhase extensible class extensions should call OMR::CodeGenPhaseConnector to make sure it calls the platform specific extension point, otherwise we will end up calling the non-platform specific extension and we may miss some switch cases.

This fixes an issue when building a debug build on Linux on Z which complains that:

```
TR::CodeGenPhase 19 doesn't have a corresponding name.
```


fjeremic:
Preemptively launching builds for this simple fix.

Jenkins test sanity zlinux

Issue #2128, opened 2018-06-08
cpConsistency check should ensure cpIndex and cp are from the same version of the class


DanHeidinga:
The change in https://github.com/eclipse/openj9/pull/1752 limited the check to JDK9 and above but didn't fix the underlying issue:

If a redefinition occurs, the `cpIndex` may be from a previous version of the class and should be checked against that class's `cpTypeDescription`.

Mixing a `cpIndex` and `cpTypeDescription` from different versions of the class is invalid.

PR #2130, opened 2018-06-08
Ensure cpTypeDescription and cpIndex are from the same class

DanHeidinga:
Mixing the cpIndex and cpTypeDescription from two different versions
of the same class can result in incorrect data being used as the
constantpool is not kept consistent between two different versions
of the same class.

This commit only updates invokespecial as resolves of
invokestatic and invokevirtual were already correct.

Fixes: #2128


gacholio:
jenkins test sanity xlinux jdk9

gacholio:
jenkins test sanity xlinux jdk9

PR #2131, opened 2018-06-08
 Remove obsolete multi-tenant constants

DanHeidinga:


pshipton:
Copyright failure in runtime/oti/j9dump.h

runtime/oti/j9consts.h
```diff
@@ -287,14 +287,14 @@ extern "C" {
 #define J9_JAVA_CLASS_GC_SPECIAL 0x800000
 #define J9_JAVA_CLASS_HAS_VERIFY_DATA 0x800000
 #define J9_JAVA_METHOD_DEBUGGABLE 0x1000000
-#define J9_JAVA_CLASS_NEEDS_PER_TENANT_INITIALIZATION 0x1000000
+#define J9_JAVA_CLASS_UNUSED_1000000 0x1000000
```
pshipton: There are already constants with the same values, do we need the UNUSED entries?

DanHeidinga: These constants aren't ordered well - there is a mix of class and method constants here

PR #2137, opened 2018-06-08
Add implementation of java.lang.ref.Reference.clone()

babsingh:
Added an implementation of java.lang.ref.Reference.clone().

Closes: #1978

[ci skip] 

### DIFF
--- a/jcl/src/java.base/share/classes/com/ibm/oti/util/ExternalMessages-MasterIndex.properties
+++ b/jcl/src/java.base/share/classes/com/ibm/oti/util/ExternalMessages-MasterIndex.properties
@@ -1282,3 +1282,6 @@ K0703="Negative length"
 K0800="Invalid Unicode code point - {0}"
 K0801="Last character in replacement string can't be \, character to be escaped is required."
 K0802="Last character in replacement string can't be $, group index is required."
+
+#java.lang.ref.Reference
+K0900="Create a new Reference, since a Reference cannot be cloned."
--- a/jcl/src/java.base/share/classes/java/lang/ref/Reference.java
+++ b/jcl/src/java.base/share/classes/java/lang/ref/Reference.java
@@ -235,4 +235,19 @@ public static void reachabilityFence(java.lang.Object ref) {
 }
 /*[ENDIF]*/
 
+/*[IF Java11]*/
+/**
+ * This method will always throw CloneNotSupportedException. A clone of this instance will not be returned 
+ * since a Reference cannot be cloned. Workaround is to create a new Reference.
+ * 
+ * @throws CloneNotSupportedException always since a Reference cannot be cloned
+ *
+ * @since 11
+ */
+@Override
+protected Object clone() throws CloneNotSupportedException {
+	/*[MSG "K0900", "Create a new Reference, since a Reference cannot be cloned."]*/
+	throw new CloneNotSupportedException(com.ibm.oti.util.Msg.getString("K0900")); //$NON-NLS-1$
+}
+/*[ENDIF] Java11 */
 }


babsingh:
Description of _java.lang.ref.Reference.clone()_: http://cr.openjdk.java.net/%7Eiris/se/11/build/latest/api/java.base/java/lang/ref/Reference.html#clone()

fyi - @pshipton

jcl/src/java.base/share/classes/java/lang/ref/Reference.java
```diff
@@ -235,4 +235,19 @@ public static void reachabilityFence(java.lang.Object ref) {
...
+ * throw CloneNotSupportedException. A clone of this instance will not be returned.
+ * 
+ * @return a clone of this instance
+ * 
+ * @throws CloneNotSupportedException always
+ *
+ * @since 11
+ */
+protected Object clone() throws CloneNotSupportedException {
+	throw new CloneNotSupportedException("A Reference cannot be meaningfully cloned. Instead, construct a new Reference."); //$NON-NLS-1$
```
pshipton: Please create an external message for this.

pshipton: See /jcl/src/java.base/share/classes/com/ibm/oti/util/ExternalMessages-MasterIndex.properties
and for example:
```
			/*[MSG "K0626", "Unable to handle type {0}."]*/
			throw new InternalError(com.ibm.oti.util.Msg.getString("K0626", type)); //$NON-NLS-1$
```

babsingh: added an external message, and updated description.

jcl/src/java.base/share/classes/java/lang/ref/Reference.java
```diff
@@ -235,4 +235,17 @@ public static void reachabilityFence(java.lang.Object ref) {
...
+/**
+ * This method will always throw CloneNotSupportedException. A clone of this instance will not be returned 
+ * since a Reference cannot be cloned. Workaround is to create a new Reference.
+ * 
+ * @throws CloneNotSupportedException
+ *
+ * @since 11
+ */
+protected Object clone() throws CloneNotSupportedException {
+	throw new CloneNotSupportedException(com.ibm.oti.util.Msg.getString("K0900")); //$NON-NLS-1$
```
pshipton: The use of K0900 needs to be preceded by a java preprocessor command to declare the external message, otherwise the message won't be defined. The only purpose adding it to ExternalMessages-MasterIndex.properties is so we have a single place that lists all the keys in use.

jcl/src/java.base/share/classes/com/ibm/oti/util/ExternalMessages-MasterIndex.properties
```diff
@@ -1282,3 +1282,6 @@ K0703="Negative length"
 K0800="Invalid Unicode code point - {0}"
 K0801="Last character in replacement string can't be \, character to be escaped is required."
 K0802="Last character in replacement string can't be $, group index is required."
+
+#java.lang.ref.Reference
+K0900="Create a new Reference since a Reference cannot be cloned."
```
pshipton: Please put a comma after the first "Reference".

jcl/src/java.base/share/classes/java/lang/ref/Reference.java
babsingh: fixed.

jcl/src/java.base/share/classes/com/ibm/oti/util/ExternalMessages-MasterIndex.properties
babsingh: fixed.

pshipton:
jenkins compile plinux jdk10

pshipton:
Note the [ci skip] can go in the body of the commit message, it doesn't need to be in the title

babsingh:
for OMR, [ci-skip] tag needs to be in the header. otherwise, it doesn't get recognized by appveyor according to Charlie. i thought it would be the same for OpenJ9.

pshipton:
> i thought it would be the same for OpenJ9.

Putting it in the body works for OpenJ9, which doesn't use appveyor.

jcl/src/java.base/share/classes/java/lang/ref/Reference.java
```diff
@@ -235,4 +235,18 @@ public static void reachabilityFence(java.lang.Object ref) {
 }
 /*[ENDIF]*/
 
+/*[IF Java11]*/
+/**
+ * This method will always throw CloneNotSupportedException. A clone of this instance will not be returned 
+ * since a Reference cannot be cloned. Workaround is to create a new Reference.
+ * 
+ * @throws CloneNotSupportedException
```
pshipton: This causes a warning, please add some text to explain when CloneNotSupportedException is thrown.

jcl/src/java.base/share/classes/java/lang/ref/Reference.java
```diff
@@ -235,4 +235,18 @@ public static void reachabilityFence(java.lang.Object ref) {
...
+ *
+ * @since 11
+ */
+protected Object clone() throws CloneNotSupportedException {
```
pshipton: This causes a warning, please tag with `@Override`

jcl/src/java.base/share/classes/java/lang/ref/Reference.java
babsingh: updated

jcl/src/java.base/share/classes/java/lang/ref/Reference.java
babsingh: added `@Override`

PR #2146, opened 2018-06-11
Prevent the garbage data from being decoded on the stack

ChengJin01:
The change is to avoid decoding the garbage data on
the live stack in generating the error messages if
the stack is loaded from one of stackmaps in which
there is no data existing on the stack.


ChengJin01:
Code reviewer: @DanHeidinga

gacholio:
I'll review merge as Dan is away this week.

gacholio:

PR #2157, opened 2018-06-12
Revert "Test module access checks for method handles"

keithc-ca:
Reverts eclipse/openj9#1763: fails to build on Windows.


smlambert:
Thanks Keith, my bad to have made assumptions and did not look closely to notice it had not been run on Windows before I merged.

PR #2166, opened 2018-06-13
Move BCD and DFP evaluators into separate files

NigelYiboYu:
Move BCD (packed, zoned, and unicode decimal) opCode evaluators into
J9BCDEvaluator.cpp as a separate compilation unit to avoid having large
cpp files.

Also move DFP related evaluators to DFPTreeEvaluators.cpp


NigelYiboYu:
@fjeremic This change separates the functions in the J9TreeEvaluator.

fjeremic:
Jenkins test sanity zlinux

PR #2168, opened 2018-06-13
Tenure char[] for interned String objects

charliegracie:
Interned String objects are created in tenure space. Currently,
the char[] backing these String objects is left in the nursery.
This means that until the char[] is tenure the Remembered Set has
to contain an entry for each new interned String. To reduce the
overhead on the RS we are going to tenure the char[].

Historically this was happening but the change got lost in
refactoring a few years ago.


fjeremic:
Jenkins test sanity

gacholio:
X builds failing for infrastructure reasons - ignore.

pshipton:
Tentatively tagged for the OpenJ9 0.9 release. It seems this may be released shortly, perhaps we should wait for it.

Issue #1166, opened 2018-02-13
Runtime.availableProcessors() ignores cgroup limits


thjaeckle:
When running OpenJ9/OpenJDK8 in Docker (e.g. via image [from adoptopenjdk](https://hub.docker.com/r/adoptopenjdk/openjdk8-openj9/)) a call to `Runtime.availableProcessors()` returns the amount of processors of the Docker host system and not the one defined in cgroup `/sys/fs/cgroup/cpu/cpu.cfs_quota_us`.

Hotspot has/had the same problem: https://bugs.openjdk.java.net/browse/JDK-6515172
which they fixed in 9 and apparently also backported to 8.

For Java 10 they add even better Docker support: https://bugs.openjdk.java.net/browse/JDK-8146115

As OpenJ9 really shines in cloud environments a better "out of the box" Docker container resource usage would be awesome.

pshipton:
@ashu-mehra @dinogun fyi

ashu-mehra:
In addition to Runtime.availableProcessors(), it appears Hotspot has also updated JVMTI API GetAvailableProcessors() to return same value as Runtime.availableProcessors().
I created a docker image using build 10-ea+40 and tested the output of Runtime.availableProcessors() and GetAvailableProcessors():

1) Using cpusets:
```
# docker run --rm --cpuset-cpus=0-3 --volume=/home/ashu/jvmtiTest:/jvmtiTest --entrypoint=/java/jdk-10/bin/java openjdk:jdk10_40 -agentpath:/jvmtiTest/libagent.so -cp /jvmtiTest ProcessorTest
jvmtiGetAvailableProcessors returned: 4
Runtime.availableProcessors: 4

```

2) Using cpu quota:
```
# docker run --rm --cpuset-cpus=0-3 --cpus=2 --volume=/home/ashu/jvmtiTest:/jvmtiTest --entrypoint=/java/jdk-10/bin/java openjdk:jdk10_40 -agentpath:/jvmtiTest/libagent.so -cp /jvmtiTest ProcessorTest
jvmtiGetAvailableProcessors returned: 2
Runtime.availableProcessors: 2
```

DanHeidinga:
OpenJDK has also added a `-XX:ActiveProcessorCount=count` option that enables setting the number of processors returned by `Runtime.availableProcessors()`.  We should support the option as well.

ashu-mehra:
[...]

PR #2170, opened 2018-06-13
Make jvmtiGetAvailableProcessors identical to JVM_ActiveProcessorCount

Ali-Ni:
This change causes `jvmtiGetAvailableProcessors` to behave identically
to `JVM_ActiveProcessorCount`, moving the common code to a separate
internal VM function.

This change addresses the consistency issue between the functions,
also causing `jvmtiGetAvailableProcessors` take into account cgroup limits
when appropriate as discussed in #1166.

Issue: #1166

### DIFF
--- a/runtime/j9vm/jvm.c
+++ b/runtime/j9vm/jvm.c
@@ -5499,7 +5499,6 @@ JVM_RegisterUnsafeMethods(JNIEnv* env, jclass unsafeClz)
 jint JNICALL
 JVM_ActiveProcessorCount(void)
 {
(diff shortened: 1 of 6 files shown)


Ali-Ni:
To be honest, this is completely excessive if this function is going to be used only twice. I was hoping to unify all appropriate CPU count calls with this, (JIT uses getNumTargetCPUs, which returns a cached value, other places call the portlib API directly), but in hindsight this seems to be trying to solve a problem that isn't really a problem at all.

Ali-Ni:
I've opened #2255 for the single line change, and I'll close this and keep the branch in case I come back to it.

PR #2179, opened 2018-06-13
Prevent object corruption in field hooks

gacholio:
JNI get/set field was dereferencing the jobject before calling the event
hooks.  This has two problems if a GC should occur during the hooks:

- upon return from the hooks, the dereferenced object could be incorrect

- if there is more than one agent, the value in the indirect object
pointer passed to the hook will not be updated when a GC occurs, meaning
further hooks could see a bad object value

### DIFF
--- a/runtime/vm/jnifield.cpp
+++ b/runtime/vm/jnifield.cpp
@@ -156,30 +156,30 @@ private:
 		ws->skipCount = 0;
 	}
 public:
-	static void triggerGetEvents(J9VMThread *vmThread, J9JNIFieldID *j9FieldID, j9object_t object)
+	static void triggerGetEvents(J9VMThread *vmThread, J9JNIFieldID *j9FieldID, jobject ref)
 	{
 		J9JavaVM *vm = vmThread->javaVM;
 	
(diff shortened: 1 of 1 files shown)


runtime/vm/jnifield.cpp
```diff
@@ -414,11 +413,10 @@ setObjectField(JNIEnv *env, jobject obj, jfieldID fieldID, jobject valueRef)
 
 	VM_VMAccess::inlineEnterVMFromJNI(vmThread);
 
-	object = J9_JNI_UNWRAP_REFERENCE(obj);
-
 	/* A NULL value is ok */
+	FieldEvents::triggerSetEvents(vmThread, j9FieldID, obj, (NULL == valueRef)? NULL : (void*)valueRef);
```
pshipton: If we pass NULL when valueRef is NULL, why bother checking?

gacholio: Currently, the hook requires a non-null valueAddress, which this breaks, so this is actually incorrect.

gacholio:
I'll close this for now and work on the full solution for the branch tomorrow (this was just supposed to be a stopgap).

Issue #2051, opened 2018-06-01
Eclipse equinox launcher does not recognise OpenJ9 java 10 as modular VM


DanHeidinga:
Mirrored from: https://bugs.eclipse.org/bugs/show_bug.cgi?id=535414

I notice that the Eclipse Photon launcher has a test isModularVM() function here:

https://github.com/eclipse/rt.equinox.framework/blob/master/features/org.eclipse.equinox.executable.feature/library/win32/eclipseWin.c

which tests jvm.dll for PRODUCT_VERSION_KEY to check the product version >= 9

Using Windows Explorer to find jvm.dll, then properties > details shows 

Oracle Java 10 has product version = 10.0.1.0 and file version 10.0.1.0

OpenJ9 has product version R2.9, file version 2.9.0.54

I don't know whether the launcher looks in the compressrefs, j9vm or server directories.

I think the result is that the Eclipse launcher thinks OpenJ9 is not a modular VM and strips off vmargs such as --add-modules=ALL-SYSTEM
To test, change in eclipse.ini --add-modules=ALL-SYSTEM to --add-modules=BADALL-SYSTEM and try to start Eclipse Photon under Oracle Java 10, and under OpenJ9 Java 10.
The Oracle version won't start, the OpenJ9 will.

Who should change - the Eclipse launcher or the OpenJ9 product version? Is there a standard way to test a VM version without starting it? The release file in the JDK root directory has a MODULES= line and
JAVA_VERSION="10.0.1"

OpenJ9, Oracle and IBM JVMs and JREs seem to all have this file. Perhaps it would be better than checking jvm.dll, but does involving finding the root of the jdk/jre.

Also consider Linux.

DanHeidinga:
@theresa-m Can you take a look at this?

keithc-ca:
I think would be better to fix eclipse as I've described in the bugzilla.

DanHeidinga:
> I think would be better to fix eclipse as I've described in the bugzilla.

I don't see your comment in the bugzilla.  Can you mirror it here?

keithc-ca:
From https://bugs.eclipse.org/bugs/show_bug.cgi?id=535414#c3:

I think the eclipse launcher should be improved. Relying on the output of 'java -version' or the version in jvm.dll is at best fragile. Instead I suggest that on all platforms a small Java application should be used that consults the system property 'java.specification.version'.

pshipton:
> The release file in the JDK root directory

[...]

PR #2189, opened 2018-06-14
Fix OpenJ9 jvm.dll version for Windows

theresa-m:
Eclipse equinox launcher does not recognise OpenJ9 Java 10 as modular VM. Equinox checks the product version of jvm.dll which for OpenJ9 is always set to 2.9.x.x.

Resolves: #2051

These extensions prs need to be merged prior to this pr:
Java8: ibmruntimes/openj9-openjdk-jdk8#99
Java9: ibmruntimes/openj9-openjdk-jdk9#165
Java10: ibmruntimes/openj9-openjdk-jdk10#43
Java11: https://github.com/ibmruntimes/openj9-openjdk-jdk/pull/8

### DIFF
--- a/runtime/makelib/mkconstants.mk.ftl
+++ b/runtime/makelib/mkconstants.mk.ftl
@@ -24,8 +24,16 @@
 </#list>
 
 # Define the Java Version we are compiling
-VERSION_MAJOR?=9
+ifndef VERSION_MAJOR
+$(error VERSION_MAJOR is not set from extensions code)
+endif
 export VERSION_MAJOR
+
+# Define full Java Version
+ifndef OPENJDK_VERSION_NUMBER_FOUR_POSITIONS
+$(error OPENJDK_VERSION_NUMBER_FOUR_POSITIONS is not set from extensions code)
+endif
+export OPENJDK_VERSION_NUMBER_FOUR_POSITIONS
 
 # Define a default target of the root directory for all targets.
 ifndef UMA_TARGET_PATH
--- a/sourcetools/com.ibm.uma/com/ibm/j9/uma/platform/PlatformWindows.java
+++ b/sourcetools/com.ibm.uma/com/ibm/j9/uma/platform/PlatformWindows.java
@@ -38,6 +38,8 @@ import com.ibm.uma.util.FileAssistant;
 public class PlatformWindows extends PlatformImplementation {
 	
 	static final String DEFAULT_RC_PRODUCT_NAME="IBM SDK, Java(tm) 2 Technology Edition";
+	/* version number with dots seperating four parts */
+	static final String OPENJDK_VERSION_NUMBER_FOUR_POSITIONS = System.getenv("OPENJDK_VERSION_NUMBER_FOUR_POSITIONS");
 	
 	public PlatformWindows( IConfiguration buildSpec ) {
 		super(buildSpec);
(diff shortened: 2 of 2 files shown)


sourcetools/com.ibm.uma/com/ibm/j9/uma/platform/PlatformWindows.java
```diff
@@ -38,6 +38,7 @@
 public class PlatformWindows extends PlatformImplementation {
 	
 	static final String DEFAULT_RC_PRODUCT_NAME="IBM SDK, Java(tm) 2 Technology Edition";
+	static final String VERSION_MAJOR = System.getenv("VERSION_MAJOR");
```
DanHeidinga: Is this already guaranteed to be set in the environment or is there another change to the build files to add this env var?

pshipton: Note this will have a value of 10 not 10.0.1.0. To get the actual version requires changes to the extensions code in order to make the appropriate version info available. J9VERSION_STRING embeds the version (as 10.0.1) so looking at how this works would be a start.

pshipton:
Why "*Temporarily* resolves: #2051"?

sourcetools/com.ibm.uma/com/ibm/j9/uma/platform/PlatformWindows.java
theresa-m: it looks like the default if its not set is "9" if its not set https://github.com/eclipse/openj9/blob/master/runtime/makelib/mkconstants.mk.ftl#L27.  should that be 11 now?

pshipton: It should always be set by the extensions, not sure if it matters what the default is.

pshipton: Since we expect it to be set by the OpenJDK extensions, probably an error should occur if the value is not set.

theresa-m:
These extensions prs need to be merged prior to this pr:
Java8: https://github.com/ibmruntimes/openj9-openjdk-jdk8/pull/99
Java9: https://github.com/ibmruntimes/openj9-openjdk-jdk9/pull/165
Java10: https://github.com/ibmruntimes/openj9-openjdk-jdk10/pull/43
Java11: TODO

runtime/makelib/mkconstants.mk.ftl
```diff
@@ -24,9 +24,17 @@
...
 
 # Define the Java Version we are compiling
-VERSION_MAJOR?=9
+ifndef VERSION_MAJOR
+$(error VERSION_MAJOR is not set from extensions code)
+endif
 export VERSION_MAJOR
 
+# Define full Java Version
+ifndef VERSION_NUMBER_FOUR_POSITIONS
```
pshipton: I think this name should be more explicit since there are many parts and versions.
i.e. OPENJDK_VERSION_NUMBER_FOUR_POSITIONS

theresa-m: I've updated the name to this.

pshipton:
All the extensions PRs have been merged.

pshipton:
jenkins compile

pshipton:
jenkins compile all jdk9

pshipton:
jenkins test sanity win,win32

pshipton:
jenkins test sanity win

PR #2218, opened 2018-06-18
Fix the condition to decide recompilation level

cathyzhyi:
When a recompilation is triggered by JProfiling, the triggering method
body could be both profiled or non-profiled body. If the triggering
compilation is a profiled compilation we need to promote the
recompilation to the next level.


cathyzhyi:
@mpirvu Hey Marius, could you take a look at this change? thanks!

mpirvu:
jenkins test all

cathyzhyi:
the failures are unrelated to the change.

mpirvu:
I verified that failures are not related to the commit, so merging.

Issue #2129, opened 2018-06-08
Windows JDK8 jvmti tests failed with "NPT ERROR: Cannot open library"


liqunl:
https://ci.eclipse.org/openj9/job/PullRequest-Sanity-JDK8-win_x86-64_cmprssptrs-OpenJ9/36/tapResults/
https://ci.eclipse.org/openj9/job/PullRequest-Sanity-JDK8-win_x86-64_cmprssptrs-OpenJ9/48/tapResults/

Error message:
```
===============================================
Running test cmdLineTester_jvmtitests_extended_3 ...
===============================================
cmdLineTester_jvmtitests_extended_3 Start Time: Wed Jun  6 12:08:27 2018 Epoch Time (ms): 1528304907512
test with Mode154
NPT ERROR: Cannot open library

cmdLineTester_jvmtitests_extended_3_FAILED
```

Failing tests are
```
cmdLineTester_jvmtitests_extended_3
cmdLineTester_jvmtitests_extended_7
cmdLineTester_jvmtitests_extended_8
cmdLineTester_jvmtitests_extended_9
cmdLineTester_jvmtitests_extended_11
decompileAtMethodResolve_0
```

liqunl:
Close this issue as a duplicate of https://github.com/eclipse/openj9/issues/2009

liqunl:
Notice that there are nightly builds whose commits are more advanced than the PRs that failed the tests. I downloaded one nightly build and tried the same tests on my laptop and can't reproduce the error. Since we don't archive builds for PRs, I have no chance to test the failing builds.

I tried running the test on my laptop with `npt.dll` removed, the same error occurred. Thus, I believe the error is not caused by jit/vm/gc changes. If the same tests passed in nightly build but failed in PR builds, I suspect there is a problem in the packaging process for windows builds triggered by PR's jenkins sanity test. @smlambert @llxia

pshipton:
A few PR builds are keep as attachments to the build jobs. You'll need to grab it quickly after a failures occurs.

I couldn't find a failed job that was still available, but as an example see the "Build Artifacts" of  https://ci.eclipse.org/openj9/job/Build-JDK8-win_x86-64_cmprssptrs/84/

liqunl:
[...]

PR #2224, opened 2018-06-19
Support for resetting code cache allocation pointers

mpirvu:
This support could be used for TOSS_CODE scenarios when we compile a method,
but then throw away the compilation result.

CodeCache will add two new fields which will remember the initial position of
_warmCodeAlloc and _coldCodeAlloc.
resetAllocationPointers() will reset _warmCodeAlloc and _coldCodeAlloc to
their initial values.
resetTrampolines() will reset the trampoline status.
resetCodeCache() will reset both allocation pointers and trampolines.

### DIFF
--- a/runtime/compiler/runtime/J9CodeCache.cpp
+++ b/runtime/compiler/runtime/J9CodeCache.cpp
@@ -160,6 +160,7 @@ J9::CodeCache::initialize(TR::CodeCacheManager *manager,
 
    if (!self()->OMR::CodeCache::initialize(manager, codeCacheSegment, codeCacheSizeAllocated, hashEntrySlab))
       return false;
+   self()->setInitialAllocationPointers();
 
    _manager->reportCodeLoadEvents();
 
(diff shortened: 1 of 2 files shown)


mpirvu:
jenkins test all

mpirvu:
@ymanton could you please review this PR? Thanks

ymanton:
Looks reasonable, but is there any value in having more than just `resetCodeCache()` be public? `setInitialAllocationPointers()` seems like an implementation detail, and I don't know if anyone would ever want to call the other two separately either so perhaps it's better to just have `resetCodeCache()` be public? Is there any scenario where one would want to throw away the code but not the trampolines or vice versa and would it be correct to do so?

Re: the Windows JDK8 failures, looks like #2129.

mpirvu:
> is there any value in having more than just resetCodeCache() be public? setInitialAllocationPointers() seems like an implementation detail, 

Unless I am missing something, `setInitialAllocationPointers()` is marked as private and the only new function that is public is `resetCodeCache()`

> Is there any scenario where one would want to throw away the code but not the trampolines or vice versa and would it be correct to do so?

I cannot think of a reason. The code is separated into `resetAllocationPointers()` and  `resetTrampolines()` to be able to reuse `resetTrampolines()` for `onFSDDecompile`

PR #2229, opened 2018-06-19
Optimize JVMTI watched fields

gacholio:
- Stop passing the address of stack slots to JVMTI hook listeners to
avoid issues with the stack growing during event handlers

- Keep object values up-to-date in the JVMTI hook listeners rather than
depending on being passed slots which are scanned by the GC

- Report instance field writes before the write in JNI for consistency

- Report method/location correctly for JNI field accesses inside event
handlers


runtime/jvmti/jvmtiHook.c
```diff
@@ -969,6 +966,10 @@ jvmtiHookFieldAccess(J9HookInterface** hook, UDATA eventNum, void* eventData, vo
 					callback((jvmtiEnv *) j9env, (JNIEnv *) currentThread, threadRef, methodID, (jlocation) location, (jclass) classRef, (jobject) objectRef, fieldID);
 				}
 				finishedEvent(currentThread, JVMTI_EVENT_FIELD_ACCESS, hadVMAccess, javaOffloadOldState);
+				if (NULL != object) {
+					J9VMGetFieldEvent* data = eventData;
+					data->object = *objectRef;
```
DanHeidinga: Shouldn't this be done under vmaccess?

Does it need to occur before finishedEvent?

runtime/jvmti/jvmtiHook.c
```diff
@@ -1072,6 +1070,15 @@ jvmtiHookFieldModification(J9HookInterface** hook, UDATA eventNum, void* eventDa
 					callback((jvmtiEnv *) j9env, (JNIEnv *) currentThread, threadRef, methodID, (jlocation) location, (jclass) classRef, (jobject) objectRef, fieldID, signatureType, newValue);
 				}
 				finishedEvent(currentThread, JVMTI_EVENT_FIELD_MODIFICATION, hadVMAccess, javaOffloadOldState);
+				if (NULL != object) {
```
DanHeidinga: Same comment as above

runtime/oti/j9vm.hdf
```diff
@@ -574,40 +574,40 @@ typedef UDATA (* lookupNativeAddressCallback)(struct J9VMThread *currentThread,
...
 
 	<event>
 		<name>J9HOOK_VM_PUT_FIELD</name>
 		<description>
-			Triggered just before an instance field is written by a JBputfield bytecode.
-			This event is not triggered by JIT methods.
-			Note that reserving or hooking this event incurs a significant performance penalty.
+			Triggered just before an instance field is written by the bytecode interpreter
+			or JNI.  If the listener allows a GC to occur, the object field in the hook
+			data must be kept up-to-date.  Similarly for the contents of valueAddress if it
```
DanHeidinga: `valueAddress` -> `newValue`

runtime/oti/j9vm.hdf
```diff
@@ -620,17 +620,17 @@ typedef UDATA (* lookupNativeAddressCallback)(struct J9VMThread *currentThread,
 	<event>
 		<name>J9HOOK_VM_PUT_STATIC_FIELD</name>
 		<description>
-			Triggered just before a static field is written by a JBputstatic bytecode.
-			This event is not triggered by native or JIT methods.
-			Note that reserving or hooking this event incurs a significant performance penalty.
+			Triggered just before a static field is written by the bytecode interpreter
+			or JNI. If the listener allows a GC to occur and the contents of valueAddress
```
DanHeidinga: `valueAddress` -> `newValue`

runtime/vm/jnifield.cpp
```diff
@@ -33,407 +33,886 @@
...
+		}
+	}
+
+	bool isVolatile = J9_ARE_ANY_BITS_SET(j9FieldID->field->modifiers, J9AccVolatile);
+	if (isVolatile) {
+		VM_AtomicSupport::writeBarrier();
+	}
+
+	j9object_t object = J9_JNI_UNWRAP_REFERENCE(obj);
+	valueOffset += J9_OBJECT_HEADER_SIZE;
```
DanHeidinga: Does this need to do an `& 1` when storing?

gacholio: I think the answer is yes, but it never has done.  Can we leave this to a future PR once we decide how we're going to enforce the type consistency throughout the VM?

runtime/oti/j9vm.hdf
gacholio: Fixed.

runtime/oti/j9vm.hdf
gacholio: Fixed.

runtime/jvmti/jvmtiHook.c
```diff
@@ -1071,6 +1070,16 @@ jvmtiHookFieldModification(J9HookInterface** hook, UDATA eventNum, void* eventDa
 				if (methodID != NULL) {
 					callback((jvmtiEnv *) j9env, (JNIEnv *) currentThread, threadRef, methodID, (jlocation) location, (jclass) classRef, (jobject) objectRef, fieldID, signatureType, newValue);
 				}
+				vm->internalVMFunctions->internalEnterVMFromJNI(currentThread);
+				if (NULL != object) {
+					J9VMPutFieldEvent* data = eventData;
+					data->object = *objectRef;
```
DanHeidinga: ... And this is kept alive because it's been pushed on the stack?  What kind of frame is there that arg0EA -1 is guaranteed to have space and not overwrite something else?  Also, shouldn't we have to adjust a count?

I realize that's all existing code but some clarifications would help

runtime/jvmti/jvmtiHook.c
```diff
@@ -1071,6 +1070,16 @@ jvmtiHookFieldModification(J9HookInterface** hook, UDATA eventNum, void* eventDa
...
+				}
+				if (signatureType == 'L') {
+					if (*((void **) valueAddress) != NULL) {
+						*(j9object_t*)valueAddress = *jvalueStorage;
```
DanHeidinga: Why update `data->newValue` here?  `data` doesn't appear to be used after this point

runtime/vm/jnifield.cpp
DanHeidinga: Fair enough.

DanHeidinga:

DanHeidinga:

DanHeidinga:

runtime/jvmti/jvmtiHook.c
gacholio: data is the hook event data, which may be used by another agent.

runtime/jvmti/jvmtiHook.c
gacholio: Yes, it is pushed into the arg area of the frame created by pushEventFrame (called from prepareForEvent).  The frame is a J9SFJNINativeMethodFrame with a NULL method - the stack walker treats all slots between arg0EA and the frame to be JNI local ref object pushes.

PR #2230, opened 2018-06-20
Prepare Balanced for implicit Master

amicic:
Changes that remove assumption that Master GC thread is always an
explicit/same GC thread. Basically, release any local resources at the
end of a phase (either STW or concurrent), since master GC thread (in
future) may move from one thread to another from a phase to a phase,
within the same  GC cycle.


amicic:

dmitripivkine:

PR #2236, opened 2018-06-20
Remove entitled CPUs

Ali-Ni:
This change removes entitled CPUs so that it can be removed and
replaced in OMR (eclipse/omr#2673) without build failures in OpenJ9.
Most of what is removed here will be added back (with some changes)
in #2196 along with the `-XX:ActiveProcessorCount=` flag.


runtime/tests/port/si_numcpusTest.c
```diff
@@ -117,69 +89,6 @@ int j9sysinfo_numcpus_test4 (J9PortLibrary* portLibrary, UDATA boundTest)
...
-
-	outputComment(PORTLIB,"\nSet number of entitled CPUs to number of physical CPUs.\n");
-	outputComment(PORTLIB,"Get target number of CPUs.\n");
-	outputComment(PORTLIB,"	Expected: OMR_MIN(%d, %d)\n", numberEntitledCPUs, numberBoundCPUs);
-	outputComment(PORTLIB,"	Result:   %d\n", targetNumberCPUs);
-
-	/* Get minimum of numberEntitledCPUs and numberBoundCPUs */
-	minimum = OMR_MIN(numberEntitledCPUs, numberBoundCPUs);
-
-	if (minimum != targetNumberCPUs) {
```
pshipton: I don't believe this test should be removed. It can still check targetNumberCPUs == numberBoundCPUs.

PR #2238, opened 2018-06-20
Fix Escape Analysis's is(NonThis)ArgToCall APIs

cathyzhyi:
EA reserved 8 bits to indicate whether a specific object candidate is
an argument or this argument to a method call at certain depth. 9 bits
are needed when the depths is equal to 8 which is the maximum sniffing
depth. Change the number of bits reserved to 16bits to avoid overflow
when sniffing at maximum depth.

### DIFF
--- a/runtime/compiler/optimizer/EscapeAnalysis.cpp
+++ b/runtime/compiler/optimizer/EscapeAnalysis.cpp
@@ -209,6 +209,7 @@ int32_t TR_EscapeAnalysis::perform()
       _maxInlinedBytecodeSize = 4000 - nodeCount;
       }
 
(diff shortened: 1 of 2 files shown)


runtime/compiler/optimizer/EscapeAnalysis.cpp
```diff
@@ -209,6 +209,7 @@ int32_t TR_EscapeAnalysis::perform()
       _maxInlinedBytecodeSize = 4000 - nodeCount;
       }
 
+   TR_ASSERT_FATAL(_maxSniffDepth < 16, "The argToCall and nonThisArgToCall flags are 16 bits - a depth limit great than 16 will not fit in these flags");
```
andrewcraik: great = greater

andrewcraik:
Jenkins test sanity

andrewcraik:
Jenkins test sanity

andrewcraik:
windows fails are due to #2129

PR #2241, opened 2018-06-20
Fix to remove static data used in Decimal Format Peephole

klangman:
Decimal Format Peephole was using static data that might be
used by more then one compiler thread at the same time resulting
in a compile time crash.


andrewcraik:
Jenkins test sanity

andrewcraik:
Windows failures due to #2129 - build is clean

PR #2248, opened 2018-06-21
Clean up deprecated recognized method inlining X86

0dvictor:
X86 CodeGen has long deprecated inlining of following 3 recognized methods:
1. java_lang_Object_hashCodeImpl
2. java_lang_System_currentTimeMillis
3. java_lang_Integer_rotateLeft
Cleaning-up code related them.


andrewcraik:
I'm having trouble following the logic for the currentTimeMillis removal - where is the point that makes it clearly dead?

0dvictor:
The code is not used under any scenarios and will crash JVM if it kicks in.

1. The only path reaching `inlineCurrentTimeMillis()` is from `VMinlineCallEvaluator()`, which can only be reached from `directCallEvaluator()`.  However, `directCallEvaluator()` does not route `TR::java_lang_System_nanoTime` to `VMinlineCallEvaluator()`.
2. `inlineCurrentTimeMillis()` uses the call node's first (and only) child as the address to hold temporary value; however, the child is actually FastJNI related. If it is `inlineCurrentTimeMillis()` kicks in, JVM may crash.

andrewcraik:
@0dvictor sorry I didn't get back to this one - can you rebase so I can run the build?

0dvictor:
Rebased.

andrewcraik:
Jenkins test sanity

andrewcraik:
Jenkins test sanity

andrewcraik:
Jenkins test sanity

0dvictor:
Windows failures are due to #2129

PR #1685, opened 2018-04-13
Changes to allow code to build on z/OS platform

groeges:
Changes required for building OpenJDK with OpenJ9 on z/OS

* Ensure preprocessor works on z/OS by using EBCDIC as the
default charset.
* Add a .gitattributes file to enable files to be cloned to
z/OS in charset as defined in the file.
* Add definition of SPEC in the buildtools.mk file

Signed-off-by: Steve Groeger <GROEGES@uk.ibm.com>

PR #2254, opened 2018-06-22
Remove .gitattributes file as it breaks travis builds

DanHeidinga:
Travis builds have been updated to newer versions of git which recognize
the `worktree-encoding` attribute.  This means the current files are
invalid as they force the encoding of all files as ebcdic.

Other builds, and developers, will be broken when they update to
newer versions of git.

This solution won't work long term.  Removing the file now so we
can unblock builds while thinking about the right long term solution.

This backs out part of #1685

See also https://github.com/eclipse/omr/pull/2695

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -18,23 +18,6 @@
 #
 # SPDX-License-Identifier: EPL-2.0 OR Apache-2.0 OR GPL-2.0 WITH Classpath-exception-2.0 OR LicenseRef-GPL-2.0 WITH Assembly-exception
 
-# This .gitattributes file will cause all text files EXCEPT for
-# those specifically listed below to be encoded as EBCDIC.
-# Selected binary files will not be translated at all.
-
-# The default for text files
-* git-encoding=iso8859-1 working-tree-encoding=ibm-1047
-
-# Specific types of files remain as ASCII
-*.xml git-encoding=iso8859-1 working-tree-encoding=iso8859-1
-*.dtd git-encoding=iso8859-1 working-tree-encoding=iso8859-1
-*.xsd git-encoding=iso8859-1 working-tree-encoding=iso8859-1
-buildspecs/* git-encoding=iso8859-1 working-tree-encoding=iso8859-1
-
-# git's files (which MUST be ASCII)
-.gitattributes git-encoding=iso8859-1 working-tree-encoding=iso8859-1
-.gitignore git-encoding=iso8859-1 working-tree-encoding=iso8859-1
-
 # Binary files
 *.jpg git-encoding=BINARY working-tree-encoding=BINARY
 *.png git-encoding=BINARY working-tree-encoding=BINARY


DanHeidinga:
FYI @dnakamura @groeges

groeges:
This does mean that z/OS builds will not work anymore as this is why that file was added.

dnakamura:
@groeges  It will indeed break z/OS builds. but as it stands currently builds are broken on all other platforms with new versions of git

dnakamura:
@DanHeidinga  I believe the travis build is failing, because git will fallback to the older version of `.gitattributes` if it cannot find one in the current tree.  Solution is to create an empty `.gitattributes`, tested [here](https://travis-ci.org/dnakamura/openj9/builds/395570201) (note build still fails because of the `.gitattributes` in omr, causing the hookgen definitions to be converted to ebcdic)

Proposed Solution for working with z/OS builds:
- Take current `.gitattributes` and rename to `.gitattributes.zos` or something
- when building on z/os machine manually update the attributes  eg:
```bash
git clone <openj9_repo>
cd openj9
cp .gitattributes.zos .git/info/attributes
git checkout-index -f -a  # update the working tree
```

Note: I'm,using `.git/info/attributes` since it is [non version controlled](https://git-scm.com/docs/gitattributes), but we may not care in this case

smlambert:
Related https://github.com/AdoptOpenJDK/stf/pull/41 (if we want testing on zOS).

dnakamura:
Was thinking about this further, and I don't think we even need the gitattributes for z/os. File tagging means it should just work out of the box

groeges:
File tagging only works with tools / apps that are compiled etc to use it. Java is not one of those tools / apps. So when running java on zOS it doesn’t use the file tags. We tried without the .gitattributes when we first started working on the zOS build but couldn’t get it to work, hence we needed the .gitattributes

dnakamura:
What files in particular are the issue? The compiler supports filetagging, (and if auto conversion is enabled, hookgen + tracegen should work), and AFAIK all the files we need at runtime need to be UTF-8 anyway

DanHeidinga:
> because git will fallback to the older version of .gitattributes if it cannot find one in the current tree.

@dnakamura Can you confirm that's the behaviour?  The doc in [1] says the index will be used first, but I don't see any references to older versions.

> When the .gitattributes file is missing from the work tree, the path in the index is used as a fall-back. During checkout process, .gitattributes in the index is used and then the file in the working tree is used as a fall-back.
[1] https://git-scm.com/docs/gitattributes

dnakamura:
I'm not entirely clear on the "why", however looking at the travis build, its clear its getting an old copy of the attributes from somewhere, and trying to apply them
https://travis-ci.org/eclipse/openj9/builds/395548145#L804

DanHeidinga:
Rather than deleting the file, I've limited it to only describe the binary file types. All the ebcdic parts are removed.

pshipton:
The OMR part is merged. I don't think the travis build can work as it first clones OpenJ9 which contains the incorrect .gitattributes. I'm just going to merge and we can ensure the travis build is working afterwards.

PR #2258, opened 2018-06-24
Optimize JVMTI field watches

gacholio:
- implement VM portion of JIT support for new watch strategy
- control new behaviour with command line option
- remove unused duplicate asmprotos.h


runtime/codert_vm/decomp.cpp
```diff
@@ -1348,31 +1348,43 @@ freeDecompilationRecord(J9VMThread * decompileThread, J9JITDecompilationInfo * i
...
+		/* Remove all breakpoints before resetting the methods */
 
-	removeAllBreakpoints(currentThread);
+		removeAllBreakpoints(currentThread);
 
-	/* Toss the compilation queue */
+		/* Toss the compilation queue */
 
-	currentThread->javaVM->jitConfig->jitClassesRedefined(currentThread, 0, NULL);
+		currentThread->javaVM->jitConfig->jitClassesRedefined(currentThread, 0, NULL);
```
DanHeidinga: `currentThread->javaVM->jitConfig` -> `jitConfig`

Reuse the local fetched above

runtime/codert_vm/decomp.cpp
```diff
@@ -1348,31 +1348,43 @@ freeDecompilationRecord(J9VMThread * decompileThread, J9JITDecompilationInfo * i
...
+
 	/* We have exclusive */
 
 	Trc_Decomp_jitDataBreakpointAdded_Entry(currentThread);
 
-	++(currentThread->javaVM->jitConfig->dataBreakpointCount);
+	++(jitConfig->dataBreakpointCount);
 
-	/* Remove all breakpoints before resetting the methods */
+	bool inlineWatches = J9_ARE_ANY_BITS_SET(vm->extendedRuntimeFlags, J9_EXTENDED_RUNTIME_JIT_INLINE_WATCHES);
```
DanHeidinga: This probably deserves a comment to mention that its a one-way switch.  Once inlineWatches is enabled and data breakpoints have been set, this can;t be disabled (at least not currently)

runtime/jit_vm/ctsupport.c
```diff
@@ -276,22 +277,24 @@ jitCTResolveInstanceFieldRefWithMethod (J9VMThread *vmStruct, J9Method *method,
...
-			vmStruct,
-			result,	/* can be modified */
-			fieldIndex,
-			constantPool,
-			*resolvedField,
-			0,
-			resolveFlags);
-#endif
+
+	if (J9_ARE_NO_BITS_SET(vm->extendedRuntimeFlags, J9_EXTENDED_RUNTIME_JIT_INLINE_WATCHES)) {
```
DanHeidinga: This probably needs a comment to explain why compile time resolves don't need to check for watches anymore (in this mode)

runtime/codert_vm/decomp.cpp
gacholio: This actually isn't an appropriate place for that comment - I'll add a different one, and put the comment you want in the data BP removed code.

runtime/codert_vm/decomp.cpp
gacholio: Fixed.

runtime/jit_vm/ctsupport.c
gacholio: Done.

DanHeidinga:

PR #2261, opened 2018-06-25
Mac OS bringup work

DanHeidinga:
Series of commits to address issues compiling on Mac OSX.

There's a problem compiling with cmake on linux x86 with this change that still needs to be figured out.

I'd like to get the current set of changes reviewed and ready to go while I sort out the cmake issue.


gacholio:
Loads of copyright failures - did you rebase on the latest before opening the PR?

runtime/j9vm/jvm.c
```diff
@@ -4259,8 +4259,12 @@ JVM_Open(const char* filename, jint flags, jint mode)
 #endif
 #endif
 
-#if defined(J9UNIX) || defined(J9ZOS390)
+#if defined(J9UNIX) || defined(J9ZOS390) || defined(OSX)
+#if defined(OSX)
+#define EXTRA_OPEN_FLAGS 0
+#else
 #define EXTRA_OPEN_FLAGS O_LARGEFILE
+#endif
```
gacholio: Should this just be a new ifdef for OSX?  The nesting seems to serve no purpose.

runtime/oti/j9nonbuilder.h
```diff
@@ -5346,7 +5346,11 @@ typedef struct J9JavaVM {
 	struct J9HashTable* classLoadingConstraints;
 	UDATA* vTableScratch;
 	UDATA vTableScratchSize;
+#if defined(J9VM_JIT_CLASS_UNLOAD_RWMONITOR)
 	omrthread_rwmutex_t classUnloadMutex;
+#else
+	omrthread_monitor_t classUnloadMutex;
+#endif
```
gacholio: What does this have to do with OSX?

runtime/oti/xhelpers.m4
```diff
@@ -85,16 +95,18 @@ define({FILE_START},{
 	.text
 })
 
-})	dnl OSX
-
-define({FILE_END})
+define({CALL_C_FUNCTION},$1)
```
gacholio: This is a truly terrible name for the macro, especially given the existing macro CALL_C_FUNC which actually does what its name suggests.  How about C_FUNCTION_SYMBOL?

runtime/oti/xhelpers.m4
```diff
@@ -104,7 +116,11 @@ ifdef({OSX},{
 })	dnl OSX
 })
 
+ifdef({OSX},{
+define({DECLARE_PUBLIC},{.global _$1})
+},{	dnl OSX
 define({DECLARE_PUBLIC},{.global $1})
+})	dnl OSX
```
gacholio: Could be done using the C_FUNCTION_SYMBOL macro.

runtime/port/unix/j9sock.c
```diff
@@ -1225,14 +1227,14 @@ j9sock_getaddrinfo_name(struct J9PortLibrary *portLibrary, j9addrinfo_t handle,
 int32_t
 j9sock_gethostbyaddr(struct J9PortLibrary *portLibrary, char *addr, int32_t length, int32_t type, j9hostent_t handle)
 {
-#if defined(J9ZTPF)
+#if defined(J9ZTPF) || defined(OSX) /* Disable this on OSX to get to version */
```
gacholio: I don't understand the comment.

runtime/port/unix/j9sock.c
```diff
@@ -1347,6 +1349,10 @@ j9sock_gethostbyaddr(struct J9PortLibrary *portLibrary, char *addr, int32_t leng
 int32_t
 j9sock_gethostbyname(struct J9PortLibrary *portLibrary, const char *name, j9hostent_t handle)
 {
+#if defined(OSX) /* Disable this on OSX to get to version */
+	return -1;
+#else
```
gacholio: comment.

runtime/port/unix/j9sock.c
```diff
@@ -1440,6 +1446,7 @@ j9sock_gethostbyname(struct J9PortLibrary *portLibrary, const char *name, j9host
 #undef hostentBuffer
 
 	return 0;
+#endif
```
gacholio: comment

runtime/port/unix/j9sock.c
```diff
@@ -1494,11 +1501,11 @@ j9sock_getnameinfo(struct J9PortLibrary *portLibrary, j9sockaddr_t in_addr, int3
 {
 	OMRPORT_ACCESS_FROM_J9PORT(portLibrary);
 	/* On z/TPF we don't support this option of returning the host name from the in_addr */
-#if defined(J9ZTPF)
+#if defined(J9ZTPF) || defined(OSX)  /* Disable this on OSX to get to version */
```
gacholio: Same obscure comment as above.

runtime/shared_common/Manager.hpp
```diff
@@ -294,7 +294,9 @@ class SH_Manager
 
 private:
 	UDATA _state;
+#if defined(J9SHR_CACHELET_SUPPORT)
 	bool _allCacheletsStarted;
+#endif
```
gacholio: Is this a general bug that OSX exposed by not having this flag?

runtime/shared_common/OSCachevmem.hpp
```diff
@@ -63,6 +63,9 @@ class SH_OSCachevmem : public SH_OSCacheFile
 	virtual bool startup(const char* ctrlDirName, UDATA cacheDirPerm, const char* cacheName, J9SharedClassPreinitConfig* piconfig, IDATA numLocks,
 			UDATA createFlag, UDATA verboseFlags, U_64 runtimeFlags, I_32 openMode, UDATA storageKeyTesting, J9PortShcVersion* versionData, SH_OSCacheInitializer* initializer, UDATA reason);
 
+	/* Work around compile error on mac osx due to hiding the other startup api */
+	using SH_OSCache::startup;
+
```
gacholio: I thought this change (or something very similar) had already been merged?

runtime/vm/xa64/unsafeHelper.m4
```diff
@@ -0,0 +1,123 @@
...
+	DECLARE_PUBLIC(unsafeGet16)
+	DECLARE_PUBLIC(unsafeGet8)
+
+dnl Prototype: void unsafePut64(I_64 *address, I_64 value);
+dnl Defined in: #Args: 2
+
+START_PROC(unsafePut64)
+        push rbp
+        mov rbp, rsp
+	mov [rdi], rsi
```
gacholio: Bad spacing throughout.

runtime/vm/xcinterp.m4
```diff
@@ -163,7 +163,7 @@ START_PROC(c_cInterpreter)
 	mov uword ptr J9TR_ELS_jitGlobalStorageBase[_rax],_rbx
 	lea _rbx,J9TR_cframe_jitFPRs[_rsp]
 	mov uword ptr J9TR_ELS_jitFPRegisterStorageBase[_rax],_rbx
-cInterpreter:
+CALL_C_FUNCTION(cInterpreter):
```
gacholio: This is not a C entrypoint, making use of this macro even more confusing.

runtime/oti/xhelpers.m4
gacholio: What we really need are two macros - one to name public symbols (so the xcinterp.m4 change works), and one to call public symbols (so you can add the _ or PLT).

runtime/oti/j9nonbuilder.h
DanHeidinga: I was building without the JIT and the compiler complained about the incorrect type in some of the calls.  So not strictly OSX related, but found as part of this work.

runtime/port/unix/j9sock.c
DanHeidinga: This is just a marker to come back and look at the right answer for this function.  I've disabled to  "get to java -version" but it'll need to be revisited.

runtime/port/unix/j9sock.c
DanHeidinga: see above

runtime/shared_common/Manager.hpp
DanHeidinga: clang is more picky about this then gcc has been

runtime/port/unix/j9sock.c
gacholio: ok - would be clearer if it  was "-version"

runtime/oti/j9nonbuilder.h
gacholio: ok

runtime/shared_common/Manager.hpp
gacholio: ok

DanHeidinga:

PR #2262, opened 2018-06-25
Pass a field's declaring class to foldFinalFieldIn

liqunl:
Static final field folding is undesirably turned on because the field's
class is passed to foldFinalFieldIn instead of its declaring class.


liqunl:
@andrewcraik Can you review?

andrewcraik:
Jenkins test sanity

andrewcraik:
the newer API usage makes sense.

andrewcraik:
travis failure is infrastructure and the same testing ran in subsequent sanity tests (plus more). Windows failures were due to #2129.

Issue #1999, opened 2018-05-25
cmdLineTester_classesdbgddrext failed on wins


llxia:
```
12:55:45 + build/windows-x86_64-normal-server-release/images/j2sdk-image/bin/java -version
12:55:45 openjdk version "1.8.0_172-internal"
12:55:45 OpenJDK Runtime Environment (build 1.8.0_172-internal-jenkins_2018_05_25_11_17-b00)
12:55:45 Eclipse OpenJ9 VM (build HEAD-65ea150e, JRE 1.8.0 Windows Server 2012 R2 amd64-64-Bit Compressed References 20180525_1 (JIT enabled, AOT enabled)
12:55:45 OpenJ9   - 65ea150e
12:55:45 OMR      - 845000f7
12:55:45 JCL      - 556b4d2928 based on jdk8u172-b11)
```

https://ci.eclipse.org/openj9/job/PullRequest-Extended-JDK8-win_x86-64_cmprssptrs-OpenJ9/1/consoleFull

```
13:07:00 Testing: Run !dumpallromclasslinear 10
13:07:00 Test start time: 2018/05/25 12:07:03 Central Standard Time
13:07:00 Running command: C:/Users/jenkins/workspace/PullRequest-Extended-JDK8-win_x86-64_cmprssptrs-OpenJ9/build/windows-x86_64-normal-server-release/images/j2sdk-image/jre/bin\..\..\bin\jdmpview.exe -core j9core.dmp
13:07:00 Time spent starting: 16 milliseconds
13:07:04 Time spent executing: 3391 milliseconds
13:07:04 Test result: FAILED
13:07:04  [OUT] DTFJView version 4.29.5, using DTFJ version 1.12.29003
13:07:04  [OUT] Loading image from DTFJ...
13:07:04  [OUT] 
13:07:04  [OUT] For a list of commands, type "help"; for how to use "help", type "help help"
13:07:04  [OUT] Available contexts (* = currently selected context) : 
13:07:04  [OUT] 
13:07:04  [OUT] Source : file:///C:/Users/jenkins/workspace/PullRequest-Extended-JDK8-win_x86-64_cmprssptrs-OpenJ9/openj9/test/TestConfig/test_output_15272680044710/cmdLineTester_classesdbgddrext_1/j9core.dmp
[...]

PR #2266, opened 2018-06-26
Improve DDR interoperability

keithc-ca:
Core files created on 32-bit platforms should be readable on 64-bit platforms and vice-versa. Because the debug information of some compilers (e.g. Visual Studio) doesn't distinguish between UDATA and U64 (in 64-bit builds) or between UDATA and U32 (in 32-bit builds), we have to be prepared to encounter a U64 object or a U32 object where the C/C++ source used UDATA.

* The interface of the generated pointer classes must be generalized so that accessors for all U32 and U64 fields are declared to return UDATA so that hand-written code is not prematurely bound to either
  specific type.
* Generated pointer types also cannot be bound to the JVM: they must be derived from each core file (like the structure classes are).
* This requires adjustments to the class hierarchy of IDATA and UDATA. U32 and U64 must both extend UDATA so that the expectation for an UDATA object can be satifsfied by either an U32 object (from a 32-bit core file) or an U64 object (from a 64-bit core file).
* Likewise for the pointer types: both U32Pointer and U64Pointer must extend UDATAPointer.

The same situations exist with respect to IDATA, I32 and I64 and their pointer classes.

Some adjustments will be required to the hand-written code. Places where U32 or U64 (or the related pointer types) were expected from a pointer accessor method, instead UDATA (or UDATAPointer) will be returned.

### DIFF
--- a/debugtools/DDR_VM/src/com/ibm/j9ddr/StructureTypeManager.java
+++ b/debugtools/DDR_VM/src/com/ibm/j9ddr/StructureTypeManager.java
@@ -27,6 +27,7 @@ import java.util.HashMap;
 import java.util.HashSet;
 import java.util.Map;
 import java.util.Set;
+import java.util.regex.Pattern;
 
 import com.ibm.j9ddr.StructureReader.StructureDescriptor;
 
@@ -115,6 +116,14 @@ public class StructureTypeManager
 		
 		localSimpleTypeCodeMap.put("char", TYPE_U8);
 		
+		localSimpleTypeCodeMap.put("bool", TYPE_BOOL);
+		localSimpleTypeCodeMap.put("double", TYPE_DOUBLE);
+		localSimpleTypeCodeMap.put("float", TYPE_FLOAT);
+		localSimpleTypeCodeMap.put("fj9object_t", TYPE_FJ9OBJECT);
+		localSimpleTypeCodeMap.put("iconv_t", TYPE_IDATA);
+		localSimpleTypeCodeMap.put("j9objectclass_t", TYPE_J9OBJECTCLASS);
+		localSimpleTypeCodeMap.put("j9objectmonitor_t", TYPE_J9OBJECTMONITOR);
+
 		simpleTypeCodeMap = Collections.unmodifiableMap(localSimpleTypeCodeMap);
 		
 		Map<Integer, String> localSimpleTypeAccessorMap = new HashMap<Integer, String>();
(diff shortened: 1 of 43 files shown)


pshipton:
jenkins test all plinux jdk8

pshipton:
Do these changes fix #1999?

keithc-ca:
I expect this will fix #1999, but only if the core file is produced on a platform with the same bit-ness; not a mix of 32-bit and 64-bit: there's more work for that.

pshipton:
I'd like to unexclude the Windows 64-bit test for #1999 then, so it can be tested along with these changes.

pshipton:
I'd also like to unexclude for Windows 64-bit new DDR testing which has been recently added.

cmdLineTester_shrcdbgddrext - test/functional/cmdLineTests/shrcdbgddrext/playlist.xml
cmdLineTester_classesdbgddrext - test/functional/cmdLineTests/classesdbgddrext/playlist.xml

pshipton:
and we should also turn on DDR for Windows 32-bit.

test/functional/cmdLineTests/shrcdbgddrext/playlist.xml
```diff
@@ -43,8 +43,8 @@
 	-jar $(CMDLINETESTER_JAR) \
 	-config $(Q)$(TEST_RESROOT)$(D)shrcdbgextddrtests.xml$(Q) -plats all,$(PLATFORM),$(VARIATION) -nonZeroExitWhenError; \
 	${TEST_STATUS}</command>
-		<!-- temporarily disable this test on aix and win github.com/eclipse/openj9/issues/1511 -->
-		<platformRequirements>^os.zos,^os.aix,^os.win</platformRequirements>
+		<!-- temporarily disable this test on AIX, z/OS; github.com/eclipse/openj9/issues/1511 -->
+		<platformRequirements>^os.aix^os.zos</platformRequirements>
```
pshipton: Missing `,` here

keithc-ca: Fixed.

keithc-ca:
DDR enabled in Windows 32-bit builds via https://github.com/ibmruntimes/openj9-openjdk-jdk8/pull/100.

pshipton:
jenkins test all win,win32 depends ibmruntimes/openj9-openjdk-jdk8#100

pshipton:
jenkins copyright check

pshipton:
jenkins test all win depends ibmruntimes/openj9-openjdk-jdk8#100

PR #2270, opened 2018-06-26
Turn on nestmates spec

tajila:
Turn on nestmates spec


keithc-ca:
Does something need to be updated for the cmake build process?

tajila:
I add a cmake cache variable for opt_nestmates opt_valhalla_nestmates

pshipton:
jenkins test sanity

pshipton:
Turning this on will affect all versions of Java. It could affect the footprint of Java 8. Has there been any footprint analysis or performance analysis done for Java 8?

Since the ROMClass format is changed, the shared classes generation number needs to be modified when opt_valhallaNestmates is enabled.

 It seems all `#if defined(J9VM_OPT_VALHALLA_NESTMATES)` should also check the Java version (JAVA_SPEC_VERSION) to avoid affecting early versions of Java which don't need these changes.

pshipton:
It seems buildenv/jenkins/jobs/pull-requests/PullRequest-Sanity-JDK11-linux_ppc-64_cmprssptrs_le_valhalla_nestmates should be deleted as well.

keithc-ca:
It might make sense to simply remove `J9VM_OPT_VALHALLA_NESTMATES` in favour of
```
#if JAVA_SPEC_VERSION >= 11 /* nest-mates */
```

pshipton:
or add something like the following rather than turning on the flag in all the spec files.
```
#if JAVA_SPEC_VERSION >= 11
#define J9VM_OPT_VALHALLA_NESTMATES 
#endif

**Comment by keithc-ca** [comment]
@pshipton I like that suggestion even better.

**Comment by tajila** [comment]
@pshipton I've updated the PR

File: runtime/shared_common/OSCache.hpp [hunk 1]
```diff
@@ -61,7 +61,7 @@
 #define OSCACHE_LOWEST_ACTIVE_GEN 1
 
 /* Always increment this value by 2. For testing we use the (current generation - 1) and expect the cache contents to be compatible. */
-#define OSCACHE_CURRENT_CACHE_GEN 35
+#define OSCACHE_CURRENT_CACHE_GEN 37
```
pshipton: We need a different generation number when J9VM_OPT_VALHALLA_NESTMATES is defined, since the ROMClass format is different.

These functions also need to be modified to understand the new generation number. Its been suggested to change them to use if statements rather than a switch so they don't need to be changed again in the future.

SH_OSCache::getHeaderFieldOffsetForGen
SH_OSCacheFile::getMmapHeaderFieldOffsetForGen
SH_OSCachesysv::getSysvHeaderFieldOffsetForGen

runtime/jcl/uma/se9_exports.xml
```diff
@@ -20,12 +20,8 @@
 	SPDX-License-Identifier: EPL-2.0 OR Apache-2.0 OR GPL-2.0 WITH Classpath-exception-2.0 OR LicenseRef-GPL-2.0 WITH Assembly-exception
 -->
 <exports group="se9">
-	<export name="Java_java_lang_Class_getNestHostImpl">
-		<include-if condition="spec.flags.opt_valhallaNestmates" />
-	</export>
-	<export name="Java_java_lang_Class_getNestMembersImpl">
-		<include-if condition="spec.flags.opt_valhallaNestmates" />
-	</export>
+	<export name="Java_java_lang_Class_getNestHostImpl"/>
```
pshipton: This isn't going to compile/link on Windows without a definition of the function.

pshipton:
@AdamBrousseau what else is required to disable the PullRequest-Sanity-JDK11-linux_ppc-64_cmprssptrs_le_valhalla_nestmates build besides removing the files? Do we need to sync any changes?

runtime/jcl/uma/se9_exports.xml
```diff
@@ -20,12 +20,8 @@
...
+	<export name="Java_java_lang_Class_getNestMembersImpl"/>
```
keithc-ca: These should be moved to se11_exports.xml for inclusion only in Java 11 (or newer). You should expect link errors in earlier version because you can't export functions that don't exist.

AdamBrousseau:
We can archive/disable the Jenkins build(s), that should be all.

runtime/jcl/uma/se11_exports.xml
```diff
@@ -0,0 +1,25 @@
+<!-- 
+	Copyright (c) 2018, 2018 IBM Corp. and others
```
keithc-ca: Can you remove the unnecessary indentation of this copyright notice and trim trailing whitespace, please?

runtime/shared_common/OSCache.hpp
```diff
@@ -61,7 +61,11 @@
 #define OSCACHE_LOWEST_ACTIVE_GEN 1
 
 /* Always increment this value by 2. For testing we use the (current generation - 1) and expect the cache contents to be compatible. */
+#ifdef J9VM_OPT_VALHALLA_NESTMATES
+#define OSCACHE_CURRENT_CACHE_GEN 37
+#else
 #define OSCACHE_CURRENT_CACHE_GEN 35
+#endif /* J9VM_OPT_VALHALLA_NESTMATES */
```
keithc-ca: @pshipton and I discussed this: the generation number does not need to be updated. We don't need to have a separate non-nestmate generation. The Java version (e.g. 8 or 11) is included in the name of the file used in addition to the generation so there won't be confusion between Java versions.

runtime/shared_common/OSCacheFile.cpp
```diff
@@ -563,12 +527,10 @@ SH_OSCacheFile::getMmapHeaderFieldOffsetForGen(UDATA headerGen, UDATA fieldID)
 			}
 		}
 		}
+	} else {
+		Trc_SHR_Assert_ShouldNeverHappen();
```
keithc-ca: This isn't executed for as many situations as previously: Was that intentional?

runtime/shared_common/OSCache.hpp
pshipton: Not incrementing the generation will cause a problem for any existing Java 11 shared caches, but since we don't have any regular builds of Java 11 yet, it seems the better choice to just go ahead without changing the cache generation.

tajila:
@keithc-ca I've updated the PR

keithc-ca:
Jenkins test extended plinux

keithc-ca:
@tajila We should plan to put a mechanism in place so jvmti.h is generated with the appropriate content for the version of Java being built, perhaps along the lines of how `forwarders.m4` works.

tajila:
@keithc-ca I have opened an issue to track this https://github.com/eclipse/openj9/issues/2368.

Issue #36, opened 2017-09-16
Supporting OS X Mac builds


mstoodle:
It has been pointed out that Eclipse OpenJ9 does not yet have OS X builds. Eclipse OMR provides OS X support. In this issue we can discuss what would be needed.

gireeshpunathil:
A first-cut porting outline to kick-start some conversation:
 - API library: already available as openjdk component?
 - JIT: I guess macOS X > 10 (Sierra) supports only x64 so codegen is re-used. Control modules may need to accommodate the new platform
 - GC: no port required?
 - Port library: Major code changes expected here. MacOS is fundamentally a BSD variant with POSIX compliance, so has good affinity with Linux. Threading, networking, file system, process management, memory management, sync. primitives needs to be looked upon on each system call level, rather than build-and- fix-breaks approach I guess.
 - RAS: tracing and dumping should be straightorward, but postmortem tooling needs fresh development (MAT, DDR, IDDE, GCMV, HC..)
 - Build: Not known (to me)
 - Test: Not known

charliegracie:
- GC - minor work since most of the GC code lives in OMR and it already works on OS X. GC policies like metronome and balanced may require work since they have not been moved into OMR yet so they have not been tested on OS X.  Balanced does rely on numa to to improve pause times and object layout so this is something that would have to be investigated.  Also the compressed references option requires being able to directly malloc / reserve memory below the 4G memory boundary.  I have not investigated this yet to see if it is possible on OS X but I assume it is.
- Most of the Port Library is in OMR which already support OS X so this should be minimal work.
- Most of the RAS tooling should just work.  I assume there will be minor tweaks required but this should not be that difficult and likely not required to get a build up and going initially.

jdekonin:
I started playing with this two nights ago.  Please assign to me and I'll have 2 wip pull requests and while I can certainly take comments, there are changes that I have to go bad and investigate/resolve still.  These changes do not compile as yet.  I believe I get through a lot (if not all) of java compilation and but I am failing in compiling openj9 native code.  I get through OMR native compile through so I believe I have the glue layer correct.
 
My current failure is looking like this : 
```
/Library/Developer/CommandLineTools/usr/bin/make  -C ./jilgen -f constgen.mk
[...]

PR #2274, opened 2018-06-26
OSX: Fix BlockPriorityQueue stored element type

nbhuiyan:
In std::priority_queue<T,V,C> where,
  * T = stored element type
  * V = <T,A> = container vector, with A being the allocator
  * C = comparator  

if T is not the same type as V::value_type, it results in
undefined behaviour.

Previously, the priority_queue was being set up as
std::priority_queue<size_t,V,C> with
V = < int32_t , BlockContainerAllocator>.
To Clang, size_t and int32_t aren't equivalent, and that fails
the [static assertion in libc++](https://github.com/llvm-mirror/libcxx/blob/1639392a4af6150c69d77bbda1327b2fbce8c293/include/queue#L437) meant to prevent undefined
behavior that may arise from this.

issue: #36

Also-by: Robert Young <rwy0717@gmail.com>

### DIFF
--- a/runtime/compiler/optimizer/JProfilingBlock.hpp
+++ b/runtime/compiler/optimizer/JProfilingBlock.hpp
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2000, 2017 IBM Corp. and others
+ * Copyright (c) 2000, 2018 IBM Corp. and others
  *
  * This program and the accompanying materials are made available under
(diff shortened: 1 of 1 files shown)


nbhuiyan:
@DanHeidinga @0xdaryl requesting review

0xdaryl:
Jenkins test sanity

0xdaryl:
Windows CI failures appear to be due to #2129.  Merging.

PR #2279, opened 2018-06-27
Modify the translated PII files in nls folder

panlei080988:
Check in the new translation packages from translators after J182F's TVT pass 1 is completed.

Signed-off-by: Lei Pan pleinb@cn.ibm.com


pshipton:

Issue #2135, opened 2018-06-08
Improve x86 JIT wrtbari code generation for gencon garbage collection


hzongaro:
@charliegracie Charlie pointed out that x86 code generated for the `wrtbari` opcode for gencon looks something like this:

```
 if (J9_PRIVATE_FLAGS_CONCURRENT_MARK_ACTIVE)) {
   call the VM helper
   return; // The helper will do both checks anyway right now.
}
if (object is old) {
     if (object is not remembered) {
       call VMHelper;
       return;
     }
}
```
But instead, it should look something like this:
```
if (Object is old) {
   if (J9_PRIVATE_FLAGS_CONCURRENT_MARK_ACTIVE)) {
      call the VM helper;
      return; // The helper will do both checks anyway right now.
   }
   if (value is NOT old) {
     if (object is not remembered) {
       call VMHelper;
       return;
     }
   }
}
```
In addition, Charlie points out that in some cases the generated code checks whether an object is old by comparing with the `lowTenureAddress` and `highTenureAddress` fields of `J9VMThread`.  Instead it should be using `heapBaseForBarrierRange` and `heapSizeForBarrierRange` always.

PR #2280, opened 2018-06-27
Gc mark remembered final

hzongaro:
Rework x86 code generation for wrtbari under gencon GC
    
Reworked code to first check whether the target object of the write barrier is old and whether the value is not old to reduce the frequency of calling the out-of-line VM write barrier helper.  Also, should always check heapBaseBarrierRange and heapSizeBarrierRange instead of lowTenureAddress and highTenureAddress, as the latter will be going away.

Fixes: #2135

### DIFF
--- a/runtime/compiler/x/codegen/J9TreeEvaluator.cpp
+++ b/runtime/compiler/x/codegen/J9TreeEvaluator.cpp
@@ -13204,7 +13204,6 @@ void J9::X86::TreeEvaluator::VMwrtbarWithoutStoreEvaluator(
       || gcMode == TR_WrtbarAlways
       ;
 
-
    bool unsafeCallBarrier = false;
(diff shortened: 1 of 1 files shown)


hzongaro:
Internal build looks good.

andrewcraik:
Can you please squash commits to a set that represents the change to facilitate reviewing. @yanluo7 can you have a look at the implementation since you are familiar with this code? FYI @charliegracie this should implement the change you discussed with us about the wrtbar sequence.

hzongaro:
@andrewcraik wrote:

> Can you please squash commits to a set that represents the change to facilitate reviewing.

Squashed!

andrewcraik:

andrewcraik:
Windows failure due to #2129

PR #2282, opened 2018-06-27
Compile DDR code with Java 8 legacy builds

keithc-ca:
There's no need to limit DDR code to Java 5.


keithc-ca:
The copyright failure for org.eclipse.jdt.core.prefs is to be expected: it didn't have a copyright before and still doesn't.

pshipton:
jenkins test sanity jdk8

pshipton:
jenkins test sanity all jdk8

Issue #2269, opened 2018-06-26
Need JVM support for nestmates


keithc-ca:
The latest merge from jdk-11+19 expects three new JVM functions; builds will fail with link errors until these are defined:
```
JNIEXPORT jboolean JNICALL
JVM_AreNestMates(JNIEnv *env, jclass current, jclass member);

JNIEXPORT jclass JNICALL
JVM_GetNestHost(JNIEnv *env, jclass current);

JNIEXPORT jobjectArray JNICALL
JVM_GetNestMembers(JNIEnv *env, jclass current);
```

theresa-m:
Working on stubs to unblock builds.

JasonFengJ9:
I have a change to add these three methods however adding them alone still won't make raw build -version work due to an j.l.IllegalAccessError.

tajila:
@JasonFengJ9 we need to turn on the spec as well, https://github.com/eclipse/openj9/pull/2270

We also need to do some refactoring for the JVM_* methods so we don't duplicate the code, but that can be done after the stubs are added.

JasonFengJ9:
Enabled the `nestmates` spec as per #2270, now get following error:
```
Exception in thread "main" java/lang/IncompatibleClassChangeError: java/lang/ThreadLocal$ThreadLocalMap.getEntry(Ljava/lang/ThreadLocal;)Ljava/lang/ThreadLocal$ThreadLocalMap$Entry;
	at java/lang/ThreadLocal.get (java.base@9/ThreadLocal.java:165)
	at java/lang/StringCoding.decodeASCII (java.base@9/StringCoding.java:520)
	at java/lang/StringCoding.decode (java.base@9/StringCoding.java:235)
	at java/lang/String.<init> (java.base@9/String.java:586)
	at java/lang/String.<init> (java.base@9/String.java:627)
	at java/lang/System.initProperties (java.base@9/NativeMethod:4294967295)
	at java/lang/System.ensureProperties (java.base@9/System.java:284)
	at java/lang/System.afterClinitInitialization (java.base@9/System.java:128)
	at java/lang/Thread.initialize (java.base@9/Thread.java:371)
	at java/lang/Thread.<init> (java.base@9/Thread.java:153)
```

JasonFengJ9:
Discussed with Tobi, this `IncompatibleClassChangeError` doesn't happen in a `CCM` so it is raw build specific only.

tajila:
[...]

PR #2283, opened 2018-06-27
Add stub methods to bringup jdk-11+19

JasonFengJ9:
Add stub methods to bringup `jdk-11+19`

```
JVM_GetNestHost
JVM_GetNestMembers
JVM_AreNestMates
```

issue: #2269 

Reviewer: @DanHeidinga 

### DIFF
--- a/runtime/j9vm/CMakeLists.txt
+++ b/runtime/j9vm/CMakeLists.txt
@@ -33,7 +33,7 @@ target_sources(jvm_common INTERFACE
 	${CMAKE_CURRENT_SOURCE_DIR}/j7vmi.c
 	${CMAKE_CURRENT_SOURCE_DIR}/j8vmi.c
 	${CMAKE_CURRENT_SOURCE_DIR}/j9memcategories.c
-	${CMAKE_CURRENT_SOURCE_DIR}/java9vmi.c
+	${CMAKE_CURRENT_SOURCE_DIR}/java11vmi.c
 	${CMAKE_CURRENT_SOURCE_DIR}/jvm.c
 	${CMAKE_CURRENT_SOURCE_DIR}/vmi.c
 	#TODO this is gross
--- a/runtime/j9vm/java11vmi.c
+++ b/runtime/j9vm/java11vmi.c
@@ -1797,4 +1797,22 @@ JVM_BeforeHalt()
 {
 	/* To be implemented via https://github.com/eclipse/openj9/issues/1459 */
 }
+JNIEXPORT jclass JNICALL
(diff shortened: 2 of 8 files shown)


runtime/j9vm/java9vmi.c
```diff
@@ -1797,4 +1797,19 @@ JVM_BeforeHalt()
 {
 	/* To be implemented via https://github.com/eclipse/openj9/issues/1459 */
 }
+JNIEXPORT jclass JNICALL
+JVM_GetNestHost(JNIEnv *env, jclass clz)
+{
+	assert(!"JVM_GetNestHost unimplemented");
```
keithc-ca: These methods should still have a return statement; otherwise compilers may issue a warning which may prove fatal.

pshipton:
It would also make sense to add these stubs into a new file java11vmi.c

JasonFengJ9:
How about rename `java9vmi.c` to `java11vmi.c` instead (Java 9 is gone now)?

runtime/j9vm/java9vmi.c
JasonFengJ9: Agreed in principle. But it is not fatal in the build.
```
gcc -DJAVA_SPEC_VERSION=9 -O3 -fno-strict-aliasing -g -MMD -DLINUX -D_REENTRANT -D_FILE_OFFSET_BITS=64 -fPIC -DIPv6_FUNCTION_SUPPORT -DJ9HAMMER -m64 -Wimplicit -Wreturn-type -Werror -I. -I../include -I../oti -I../harmony/include -I../omr/gc/include -I../gc_glue_java -I../hyvm -I../makelib -I../j9vm -I../j9vm -I../j9vm -I../j9vm -I../j9vm -I../j9vm -I../j9vm -I../j9vm -I../nls -I../omr/include_core  -DJ9VM_JAVA9_BUILD=156 -DJ9VM_JCL_SE11  -DUT_DIRECT_TRACE_REGISTRATION -DTR_HOST_X86 -c -o java9vmi.o ../j9vm/java9vmi.c
```
Probably because `-Werror=return-type` is not specified.

pshipton: It could fail on other platforms, such as Windows.

pshipton:
> How about rename java9vmi.c to java11vmi.c instead (Java 9 is gone now)?

Agreed. We still have Java 10 atm but it will become obsolete shortly.

runtime/j9vm/java9vmi.c
JasonFengJ9: Added `return` clauses, and renamed `java9vmi.c` to `java11vmi.c`. Pls have another look.

PR #2284, opened 2018-06-27
Fix compiler warning on mac osx

DanHeidinga:
UDATA variable will always, by definition, be >= 0.
Assigning it a value of U_16 so safe to replace UDATA
with IDATA.

issue: #36


gacholio:

PR #2057, opened 2018-06-01
Check for boolean when running putfield/static bytecode

tajila:
The JVM spec states "If the value is of type int and the field
descriptor type is boolean, then the int value is narrowed by taking the
bitwise AND of value and 1, resulting in value'"

issue: #2049

Signed-off-by: tajila <atobia@ca.ibm.com>

Issue #2049, opened 2018-06-01
Correctly truncate boolean values in the appropriate bytecodes


DanHeidinga:
Bytecodes that need to be updated:   **Complete in 0.11.0**
- [x]  bastore #2910 #2046
- [x] putfield / putstatic #2057 #2285
- [x] returns #1831 #1943

Targeted for 0.12.0:
- [x] Confirm whether JNI follows the same rules
- [x] VM truncate boolean JNI returns: ~~#3678~~ replaced by #3896
- [x] VM Truncate boolean writes in JNI and Unsafe: ~~#3648~~ replaced by #3896
- [x] VM Truncate boolean reads in Unsafe to match RI: #4045
- [x] JIT JNI directToJNI truncate boolean returns on X: #4086
- [x] All JIT Power work: https://github.com/eclipse/openj9/issues/4241
- [x] JNI truncation on Z: https://github.com/eclipse/openj9/issues/4162
- [x] JIT common work:  done in 0.11

- [x] JIT Unsafe truncate boolean writes in JNI / Unsafe to match #4045 & #3896 - jit implementation #6095

### Unsafe / JNI behaviour
As validated in #3896, JNI & Unsafe are checking ` != 0`, not `& 1`

### Check for boolean array when running bastore bytecode:

The JVM spec states: "If the arrayref refers to an array whose
components are of type boolean, then the int value is narrowed by taking
the bitwise AND of value and 1; the result is stored as the component of
the array indexed by index"

### Check for boolean array when running put{field,static} bytecodes:

The JVM spec states "If the value is of type int and the field
descriptor type is boolean, then the int value is narrowed by taking the
bitwise AND of value and 1, resulting in value'"

DanHeidinga:
Opened an issue to connect the VM and JIT PRs for this issue.

DanHeidinga:
@andrewcraik @fjeremic - I'll hold off merging the VM PR until we have an idea of when the JIT side can be ready

fjeremic:
@DanHeidinga do we have a failing test case or unit tests for this one? @andrewcraik sounds like we need to change for from:

```
bstorei
  x
  y
```

to

```
bstorei
  x
  band
    y
    bconst 1
```

fjeremic:
> I'll hold off merging the VM PR until we have an idea of when the JIT side can be ready

I have a fix ready in https://github.com/eclipse/openj9/pull/2055. Awaiting review.

DanHeidinga:
[...]

PR #2285, opened 2018-06-27
Fix putfield/static il trees to comply with JVM spec

cathyzhyi:
The JVM spec states "If the value is of type int and the field
descriptor type is boolean, then the int value is narrowed by taking the
bitwise AND of value and 1, resulting in value'"

This is the JIT part of the change. The VM part is #2057

 issue: #2049

### DIFF
--- a/runtime/compiler/compile/J9SymbolReferenceTable.cpp
+++ b/runtime/compiler/compile/J9SymbolReferenceTable.cpp
@@ -1255,7 +1255,6 @@ J9::SymbolReferenceTable::dynamicMethodSymrefsByCallSiteIndex(int32_t index)
    return _dynamicMethodSymrefsByCallSiteIndex[index];
    }
 
-
 bool
 J9::SymbolReferenceTable::isFieldClassObject(TR::SymbolReference *symRef)
    {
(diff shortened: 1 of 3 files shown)


andrewcraik:
Does this apply in Java 8 and earlier or was this language added to the specification later? If so at what version? If it only applies in a later version we may only want to add this penalty when we are dealing with a newer version which requires this behavior...

andrewcraik:
Jenkins tests sanity

andrewcraik:
probably a generally good thing to do if it doesn't affect perf - @cathyzhyi will comment on perf impact shortly.

andrewcraik:
Jenkins test sanity

andrewcraik:
Jenkins test sanity

cathyzhyi:
tested on major benchmarks without regression.

cathyzhyi:
Windows failures due to #2129

runtime/compiler/compile/J9SymbolReferenceTable.cpp
```diff
@@ -1240,6 +1240,16 @@ J9::SymbolReferenceTable::isFieldClassObject(TR::SymbolReference *symRef)
...
 
+bool
+J9::SymbolReferenceTable::isFieldTypeBool(TR::SymbolReference *symRef)
+   {
+   int32_t len;
+   const char *fieldName = symRef->getOwningMethod(comp())->fieldSignatureChars(symRef->getCPIndex(), len);
+   dumpOptDetails(comp(), "got field signature as %s\n", fieldName);
+   if (fieldName[len-2] != '[' && fieldName[len-1] == 'Z')
+      return true;
+   return false;
```
fjeremic: Why not `return fieldName[len-2] != '[' && fieldName[len-1] == 'Z'` instead of these 3 lines?

cathyzhyi: modified according to the comment. Thanks!

andrewcraik:
@cathyzhyi can we rebase this forward please since it seems to have been lost in the shuffle

cathyzhyi:
@andrewcraik thanks for the reminder. rebased to latest master and is ready to be merged.

andrewcraik:
Jenkins tests sanity

andrewcraik:
Jenkins test sanity

andrewcraik:
Jenkins test sanity jdk8,jdk11 xlinux,plinux,zlinux

andrewcraik:
Jenkins test sanity jdk8,jdk11 xlinux,plinux

andrewcraik:
Jenkins test sanity xlinux,plinux,zlinux jdk8,jdk11

andrewcraik:
@cathyzhyi can you rebase forward so we can try the tests again please?

andrewcraik:

andrewcraik:
CI is clean, sanity had infrastructure issues, but change is quite straight forward so merging

PR #2287, opened 2018-06-27
Explicitly cast to U_8*

DanHeidinga:
OSX compiles require an explicit cast as char * and
U_8* differ in unsignedness.

issue: #36


gacholio:
Trivially correct - not waiting for CI build.

PR #2290, opened 2018-06-28
Fix incorrect usage of TR::Options

harryyu1994:
- during compilation, the per-method TR::Options should be used instead of the global one
- mostly happens at the Optimizer and Code Generation stage
- simplify comp()->getOptions()->getOption(TR_XXX) to comp()->getOptions(TR_XXX)

### DIFF
--- a/runtime/compiler/env/j9method.cpp
+++ b/runtime/compiler/env/j9method.cpp
@@ -1116,7 +1116,7 @@ TR_ResolvedJ9MethodBase::isInlineable(TR::Compilation *comp)
 
 static intptrj_t getInitialCountForMethod(TR_ResolvedMethod *m, TR::Compilation *comp)
    {
-   TR::Options * options = comp->getOptions()->getCmdLineOptions();
+   TR::Options * options = comp->getOptions();
 
    intptrj_t initialCount = m->hasBackwardBranches() ? options->getInitialBCount() : options->getInitialCount();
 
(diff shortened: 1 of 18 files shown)


harryyu1994:
OMR side changes: https://github.com/eclipse/omr/pull/2715

runtime/compiler/env/j9method.cpp
```diff
@@ -1116,7 +1116,7 @@ TR_ResolvedJ9MethodBase::isInlineable(TR::Compilation *comp)
 
 static intptrj_t getInitialCountForMethod(TR_ResolvedMethod *m, TR::Compilation *comp)
    {
-   TR::Options * options = comp->getOptions()->getCmdLineOptions();
+   TR::Options * options = comp->getOptions();
```
mpirvu: I am torn about this one: we are asking about the initial count for method m which is not the same as the method being compiled. It's true that asking for the global options is not correct either because the user could have provided specific options for that method.

mpirvu:
jenkins test all

runtime/compiler/env/j9method.cpp
harryyu1994: Is that true 100% of the time? Are there cases where method m and method being compiled are the same method? 

```
TR_ResolvedMethod *compMethod = comp->getCurrentMethod();
count = compMethod->getInvocationCount();
initialCount = getInitialCountForMethod(compMethod, comp);
```
^ this looks like they are the same method, no?

harryyu1994: So this is the call where method m and method being compiled could be different
```
 intptrj_t initialCount = getInitialCountForMethod(this, comp);
```
is there a way to get `options` from the `this` pointer?

mpirvu:
jenkins test all

harryyu1994:
It does not look like the win test failures are related to this change. I think this is safe to merge (once the linux_390 test finishes).

mpirvu:
Failures are not related to the commit, so I'll merge it

PR #2291, opened 2018-06-28
Set default variables from extensions

theresa-m:
Having this value as 8 I think makes more sense than 9 as before or 11 since our version numbers are changing at such a rapid rate.

Reviewer: @pshipton


pshipton:
I think the correct approach is to ensure the values are set before mkconstants.mk is invoked.

PR #2303, opened 2018-06-29
Remove redundant declarations of parseOptionsBuffer()

keithc-ca:
It is a C function which cannot be overloaded, so the three declarations had to be equivalent, but one should be sufficient.

Also, use modern type names `intptr_t` and `uintptr_t` instead of `IDATA` and `UDATA`.


DanHeidinga:
> Also, use modern type names intptr_t and uintptr_t instead of IDATA and UDATA.

OMR has made that change.  OpenJ9 hasn't yet.  Please stick with the OpenJ9 standard types of I/UDATA

keithc-ca:
All of the other declarations in that file use intptr_t and uintptr_t: this is just making it consistent.

DanHeidinga:
Jenkins compile all jdk8

PR #2308, opened 2018-06-29
Remove definition of `max` macro. Use OMR_MAX instead

DanHeidinga:
Clang warns (and errors) on headers used in cpp files that
define max as it conflicts with std::max.  Change all uses
of `max` to `OMR_MAX` in MethodMetaData.c and delete the
define to address clang warnings.

issue: #36


pshipton:
Note both files need a copyright update.

DanHeidinga:
Fixed copyright dates.  Some day I'll get that right the first time

pshipton:
jenkins test sanity win,plinux jdk8

pshipton:

DanHeidinga:
The windows test failures are jvmti tests: #2129

PR #2310, opened 2018-07-02
Add .gitattributes for zos platform

groeges:
The .gitattributes.zos file will not be picked up by the latest versions of git, so git will not attempt to encode/decode files. 

On a zos platform the openjdk extensions file ./closed/get_j9_source.sh will be modified to copy this .gitattributes.zos file to .git/info/attributes and the files refreshed. This will then emncode/decode the files. 

Signed-off-by: Steve Groeger <GROEGES@uk.ibm.com>

### DIFF
--- a/.gitattributes.zos
+++ b/.gitattributes.zos
@@ -1,0 +1,42 @@
+# Copyright (c) 2000, 2018 IBM Corp. and others
+#
+# This program and the accompanying materials are made available under
+# the terms of the Eclipse Public License 2.0 which accompanies this
+# distribution and is available at https://www.eclipse.org/legal/epl-2.0/
+# or the Apache License, Version 2.0 which accompanies this distribution and
+# is available at https://www.apache.org/licenses/LICENSE-2.0.
+#
+# This Source Code may also be made available under the following
+# Secondary Licenses when the conditions for such availability set
+# forth in the Eclipse Public License, v. 2.0 are satisfied: GNU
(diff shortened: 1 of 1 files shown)


groeges:
@pshipton Please could you review this and merge so that we can get the zos build working again.

keithc-ca:
@groeges Have you verified the behavior that results when both `.gitattributes` *and* `.git/info/attributes` are present and both have a rule for the same file? Which takes precedence? Can we remove duplication (e.g. binary files)?

charliegracie:
I am wondering why there is a different default for the OpenJ9 project and the OMR project in:
#omr/2721

groeges:
@keithc-ca The git docs state:
```
When deciding what attributes are assigned to a path, Git consults $GIT_DIR/info/attributes file (which has the highest precedence), .gitattributes file in the same directory as the path in question, and its parent directories up to the toplevel of the work tree (the further the directory that contains .gitattributes is from the path in question, the lower its precedence). Finally global and system-wide files are considered (they have the lowest precedence).
```

groeges:
@charliegracie I am not sure. All I know is when the zos work was started changes were made just to get the build working. So I presume the omr build worked without the default being ebcdic but OpenJ9 didn't.

groeges:
In order to not impact the OpenJ9 repository with changes for z/OS we have implemented the required change within the z/OS OpenJDK Extensions repository. This PR is therefore no longer required so closing it.

Issue #2045, opened 2018-05-31
Consolidate java/lang/Math acceleration and implement it for Z


fjeremic:
It seems we have duplicated logic in the Walker as well as RecognizedCallTransformer for accelerating `java.lang.Math.min` type operations. We are also missing support for this on Z:

https://github.com/eclipse/openj9/blob/1ca0ab9841f1a6b17fde82402582249853a1e6e7/runtime/compiler/optimizer/J9RecognizedCallTransformer.cpp#L154-L158

https://github.com/eclipse/openj9/blob/1ca0ab9841f1a6b17fde82402582249853a1e6e7/runtime/compiler/ilgen/Walker.cpp#L4180-L4192

We should consolidate this logic into the RecognizedCallTransformer and implement these opcodes for Z if they are not there already.

PR #2320, opened 2018-07-04
Consolidate math call transformations in recognizedCallTransformer

dhong44:
Move transformations from java/lang/math calls to opcodes
into recognized call transformer optimization and add
transformations for Z.

Closes: #2045

### DIFF
--- a/runtime/compiler/optimizer/J9RecognizedCallTransformer.cpp
+++ b/runtime/compiler/optimizer/J9RecognizedCallTransformer.cpp
@@ -48,7 +48,6 @@ void J9::RecognizedCallTransformer::processIntrinsicFunction(TR::TreeTop* treeto
    TR::TransformUtil::removeTree(comp(), treetop);
    }
 
-
 bool J9::RecognizedCallTransformer::isInlineable(TR::TreeTop* treetop)
    {
    auto node = treetop->getNode()->getFirstChild();
@@ -65,11 +64,7 @@ bool J9::RecognizedCallTransformer::isInlineable(TR::TreeTop* treetop)
       case TR::java_lang_Math_min_I:
       case TR::java_lang_Math_max_L:
       case TR::java_lang_Math_min_L:
-         return (TR::Compiler->target.cpu.isX86() ||
-            TR::Compiler->target.cpu.isPower() ||
-            TR::Compiler->target.cpu.isZ() ||
-            TR::Compiler->target.cpu.isARM()) &&
-            !comp()->getOption(TR_DisableMaxMinOptimization);
+         return !comp()->getOption(TR_DisableMaxMinOptimization);
       case TR::java_lang_Math_max_F:
       case TR::java_lang_Math_min_F:
       case TR::java_lang_Math_max_D:
--- a/runtime/compiler/optimizer/J9RecognizedCallTransformer.cpp
+++ b/runtime/compiler/optimizer/J9RecognizedCallTransformer.cpp
@@ -51,9 +51,6 @@ void J9::RecognizedCallTransformer::processIntrinsicFunction(TR::TreeTop* treeto
 
 bool J9::RecognizedCallTransformer::isInlineable(TR::TreeTop* treetop)
    {
-   static bool disableARMMaxMin = true; // (feGetEnv("TR_DisableARMMaxMin") != NULL);
-   static bool disableARMFMaxMin = true; // (feGetEnv("TR_DisableARMFMaxMin") != NULL);
-
    auto node = treetop->getNode()->getFirstChild();
    switch(node->getSymbol()->castToMethodSymbol()->getMandatoryRecognizedMethod())
       {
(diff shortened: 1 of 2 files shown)


dhong44:
Depends on eclipse/omr#2711 and eclipse/omr#2722

dhong44:
@fjeremic

runtime/compiler/optimizer/J9RecognizedCallTransformer.cpp
```diff
@@ -48,22 +48,37 @@ void J9::RecognizedCallTransformer::processIntrinsicFunction(TR::TreeTop* treeto
...
       case TR::java_lang_Math_max_I:
       case TR::java_lang_Math_min_I:
       case TR::java_lang_Math_max_L:
       case TR::java_lang_Math_min_L:
-         return TR::Compiler->target.cpu.isX86() && !comp()->getOption(TR_DisableMaxMinOptimization);
+         return (TR::Compiler->target.cpu.isX86() ||
+            TR::Compiler->target.cpu.isPower() ||
+            TR::Compiler->target.cpu.isZ() ||
+            TR::Compiler->target.cpu.isARM() && !disableARMMaxMin) &&
+            !comp()->getOption(TR_DisableMaxMinOptimization);
```
fjeremic: Are we able to eliminate these platform checks and the `disableARMMaxMin` static variables by simply setting `TR_DisableMaxMinOptimization` on ARM? Similarly below.

0dvictor:
The proposed change implicitly enabled transforming [i/l]abs on X86 while not deleting old related X86 CodeGen logic, it might be better to wait for https://github.com/eclipse/openj9/pull/2360

0dvictor:
Thank you for waiting. https://github.com/eclipse/openj9/pull/2360 has now been merged. This PR may need a rebase to reflect the latest change.

runtime/compiler/optimizer/J9RecognizedCallTransformer.cpp
dhong44: c03fdfd

runtime/compiler/optimizer/J9RecognizedCallTransformer.cpp
```diff
@@ -48,22 +48,34 @@ void J9::RecognizedCallTransformer::processIntrinsicFunction(TR::TreeTop* treeto
    TR::TransformUtil::removeTree(comp(), treetop);
    }
 
+
```
fjeremic: Extra new line not needed.

runtime/compiler/optimizer/J9RecognizedCallTransformer.cpp
```diff
@@ -48,22 +48,34 @@ void J9::RecognizedCallTransformer::processIntrinsicFunction(TR::TreeTop* treeto
...
       case TR::java_lang_Math_max_I:
       case TR::java_lang_Math_min_I:
       case TR::java_lang_Math_max_L:
       case TR::java_lang_Math_min_L:
-         return TR::Compiler->target.cpu.isX86() && !comp()->getOption(TR_DisableMaxMinOptimization);
+         return (TR::Compiler->target.cpu.isX86() ||
+            TR::Compiler->target.cpu.isPower() ||
+            TR::Compiler->target.cpu.isZ() ||
+            TR::Compiler->target.cpu.isARM()) &&
+            !comp()->getOption(TR_DisableMaxMinOptimization);
```
fjeremic: We still need to fold all of these platform checks.

dhong44: a207c22

runtime/compiler/optimizer/J9RecognizedCallTransformer.cpp
dhong44: a207c22

runtime/compiler/optimizer/J9RecognizedCallTransformer.cpp
```diff
@@ -54,16 +54,23 @@ bool J9::RecognizedCallTransformer::isInlineable(TR::TreeTop* treetop)
...
       case TR::java_lang_Math_max_L:
       case TR::java_lang_Math_min_L:
-         return TR::Compiler->target.cpu.isX86() && !comp()->getOption(TR_DisableMaxMinOptimization);
+         return !comp()->getOption(TR_DisableMaxMinOptimization);
+      case TR::java_lang_Math_max_F:
+      case TR::java_lang_Math_min_F:
+      case TR::java_lang_Math_max_D:
+      case TR::java_lang_Math_min_D:
+         return (TR::Compiler->target.cpu.isARM()) &&
+            !comp()->getOption(TR_DisableMaxMinOptimization);
```
fjeremic: This is already set on ARM so we don't need the `isARM` platform check.

dhong44: Since only arm supports transformation of Max and Min calls, without the `isARM` check this would always return false. I intended to leave this code so in the future it could be enabled on ARM. Instead, should the cases for floating point max and min calls just be removed?

fjeremic: > Instead, should the cases for floating point max and min calls just be removed?

I think so yes.

dhong44: 770bfce

fjeremic:
Jenkins test sanity

dhong44:

PR #2322, opened 2018-07-04
Add missing JNI exception check

DanHeidinga:
Adjust formatting to be consistent and update comments to indicate
that an exception will always be pending at that point so no explicit
action is required.


keithc-ca:
Jenkins test extended plinux

keithc-ca:
Merging despite JDK8 failure which seems unrelated to this change.

PR #2328, opened 2018-07-05
Define OMR build env for ARM cross-compile

ymanton:
When building using `OMR_CROSS_CONFIGURE` the tools and environment that the OMR tools will be built with must be specified since they cannot be detected by autoconf.

This fix is needed once https://github.com/eclipse/omr/pull/2406 goes in but can be merged prior in order to avoid breakage.


ymanton:
Jenkins test sanity all all

Issue #2329, opened 2018-07-05
Functional Sanity JDK10 Linux s390x tests suddenly take 7 hours


AdamBrousseau:
First observed on June 28 in [OMR build 574](https://ci.eclipse.org/openj9/job/Pipeline-OMR-Acceptance/574/)
Test before: https://ci.eclipse.org/openj9/job/Test-Sanity-JDK10-linux_390-64_cmprssptrs/200/
Test After: https://ci.eclipse.org/openj9/job/Test-Sanity-JDK10-linux_390-64_cmprssptrs/201/
Typical build time
Compile Test material: 10min
Sanity functional tests: 1.5hrs
After regression
Compile test material: 1hr
Sanity functional tests: 6hrs

Diff between build 573/574
OpenJ9:
https://github.com/eclipse/openj9/compare/693fe845...be52aeb9
No OMR diff between builds
PRs merged
- #2295 CMake: Update exclude list in the compiler component
   - No PR build
- ~#2296 Also fetch branches for checks~
   - Jenkins file changes
- ~#2245 Process zip files up to 4 GB~
   - PR build on xlinux 8 only
   - Revert PR no change in perf
- ~#2271 Prevent recognizing JIT Helpers not for their ISA~
   - PR build was typical time
- ~#2188 [JDK11] Add Dockerfile(s) for z/p Linux and update xLinux Dockerfile~
   - Dockerfile changes
- #2167 Contribute the source code of CDS adapter to OpenJ9
   - PR compile plinux 8 only
- ~#2283 Add stub methods to bringup jdk-11+19~
   - PR build was typical time
- ~#2154 Change Git to SCM step for Copyright and Line Endings checks~
   - Jenkins file chnages

(Crossing off the PRs that have been ruled out)
Also affects PR builds
https://ci.eclipse.org/openj9/job/PullRequest-Sanity-JDK10-linux_390-64_cmprssptrs-OpenJ9/

AdamBrousseau:
Running a revert of #2245 [here](https://ci.eclipse.org/openj9/job/PullRequest-Sanity-JDK10-linux_390-64_cmprssptrs-OpenJ9/146/)

llxia:
It looks like all cmdlinetest are affected. For example:

cmdLineTester_XcheckJNI_0: changed from 8mins to 50mins
cmdLineTester_SCURLClassLoaderNPTests_SE100_1: changed from 9mins to 48mins
[...]

PR #2335, opened 2018-07-05
Correctly identify use of compressed references for CAS write barrier

fjeremic:
The `usingCompressedReferences` variable is misleading as it does not
imply compressed references are actually used. What this variable
actually means is that the value we are storing into the object via
compare and swap is a compressed reference.

Recognized CAS sequences are reduced at lower trees phase [1] where the
last two arguments being the expected value and the new value are
lowered into compressed objects since under compressed references
objects on heap are always compressed.

For concurrent scavenge we need a read barrier on the expected value
hence the read barrier under compressed references need not rely on the
use of the `usingCompressedReferences` variable which means something
totally different.

We take this opportunity to first fix the bug and to improve the
meaning of some of the variables in the modified function so it is not
confusing for future readers.

[1] https://github.com/eclipse/openj9/blob/8a36fcd5f23005bf5028d28b60fd61aa325940a1/runtime/compiler/codegen/J9CodeGenerator.cpp#L675-L683

### DIFF
--- a/runtime/compiler/z/codegen/J9TreeEvaluator.cpp
+++ b/runtime/compiler/z/codegen/J9TreeEvaluator.cpp
@@ -9382,54 +9382,59 @@ VMinlineCompareAndSwap(TR::Node *node, TR::CodeGenerator *cg, TR::InstOpCode::Mn
    TR::Node* oldVNode   = node->getChild(3);
    TR::Node* newVNode   = node->getChild(4);
 
-   bool usingCompressedPointers = false;
-   TR::Node *translatedNode = newVNode;
-   if (isObj && comp->useCompressedPointers() &&
-         translatedNode->getOpCodeValue() == TR::l2i)
+   // The card mark write barrier helper expects the source register to be a decompressed reference. As such if the
+   // value we are storing (last argument) has been lowered we must extract the decompressed reference from the
+   // compression sequence.
+   bool isValueCompressedReference = false;
+
+   TR::Node* decompressedValueNode = newVNode;
+
+   if (isObj && comp->useCompressedPointers() && decompressedValueNode->getOpCodeValue() == TR::l2i)
       {
-      // pattern match the sequence
-      //  icall                         <- node
-      //     thisNode
-      //     objNode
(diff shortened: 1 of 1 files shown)


fjeremic:
@dsouzai @amicic could I please get a review and @amicic could you please be the committer for this PR?

fjeremic:
Jenkins test sanity zlinux

dsouzai:
@joransiu I'm going to have to defer to your expertise for this review.

dsouzai:
I'm ok merging the change as is, as the PR does make sense to me. However, I think it would be worth getting someone from the z codegen team to also take a look at this, maybe @r30shah .

fjeremic:
JDK10 failure is due to #2329. That build has now been suspended. Awaiting review + merge from @amicic.

charliegracie:
@fjeremic @amicic is on vacation for until the end of the month.  I can review this for you if you like?

Looking at the change I like the refactoring (for names, etc.).  The actual fix is:
```
auto guardedLoadMnemonic = comp->useCompressedPointers() ? TR::InstOpCode::LLGFSG : TR::InstOpCode::LGG;
```
which checks to see if compressed references is running and not whether or not this value is compressed correct?

fjeremic:
> @fjeremic @amicic is on vacation for until the end of the month. I can review this for you if you like?
> 
> Looking at the change I like the refactoring (for names, etc.). The actual fix is:
> 
> auto guardedLoadMnemonic = comp->useCompressedPointers() ? TR::InstOpCode::LLGFSG : TR::InstOpCode::LGG;
> 
> which checks to see if compressed references is running and not whether or not this value is compressed correct?

Thanks @charliegracie. Yes this is indeed the fix. Perhaps I should have broken it into two commits to be more clear. Apologies for that. I have manually tested the fix against the broken workload and it does fix  the problem. I've also ran additional testing to the above Jenkins builds and there is no problem.

charliegracie:
@fjeremic no problem it was easy enough to figure out.  Just wanted to make sure I was not missing something :).  Merging as the only failure is a known issue.

Issue #1351, opened 2018-03-06
JavaLangRefAccess.runFinalization() need to be implemented


JasonFengJ9:
`JavaLangRefAccess.runFinalization()` is added since `Java 11 b03` which is invoked by `Runtime.runFinalization()`.

To resolve `Java 11` build error #1267, a stub method has been added at https://github.com/eclipse/openj9/pull/1348 however an actual implementation is still required.

dmitripivkine:
@LinHu2016 Would you please take a look to this when you will have time?

PR #2337, opened 2018-07-05
Implement JavaLangRefAccess.runFinalization()

LinHu2016:
- Implement JavaLangRefAccess.runFinalization for Java 11

Closes: #1351 


jcl/src/java.base/share/classes/java/lang/ref/Reference.java
```diff
@@ -15,6 +15,9 @@
 import sun.misc.SharedSecrets;
 /*[ENDIF]*/
 /*[ENDIF]*/
+/*[IF Java11]*/
+java.lang.ref.Finalizer;
```
pshipton: This isn't a valid statement and won't compile. It its supposed to be an import, no import is required since Finalizer is in the same package.

LinHu2016: have removed it

Issue #1608, opened 2018-04-04
Windows _extended test failures on jdk8/9 


jdekonin:
Failures on Windows64 extended testing in : 
cmdLineTester_verbosetest_0
cmdLineTester_verbosetest_3
cmdLineTester_verbosetest_6

```
[ERR] [TestNG] [ERROR] java.lang.IllegalArgumentException: InputStream cannot be null
[ERR] Exception in thread "main" java.lang.NullPointerException
[ERR] 	at org.testng.TestNG.getStatus(TestNG.java:211)
[ERR] 	at org.testng.TestNG.main(TestNG.java:1324)
```

This is using JDK8/9 variants using the latest from openj9, openj9-omr and openjdk extensions repos.  I haven't tested on jdk10.

DanHeidinga:
Are you able to confirm whether these tests have ever passed on Windows?  

As we migrate the tests and platforms to OpenJ9, we've sometimes hit test failures where the test is excluded on a platform in the internal builds.

jdekonin:
Currently I can only see this internally as eclipse/openj9 doesn't have windows machines available for testing.  But with access to internal build/test resources, I was able to track it back to March 15th for the last passing run.  Which then narrowed the shas to be between 39fd3df74842 and dbb333dde8fb.

A commit that obviously stands out is
https://github.com/eclipse/openj9/commit/b3b3e34f27b7
@TianyuZuo, can you please review?

TianyuZuo:
investigating

DanHeidinga:
This occurred in a build over the weekend: https://ci.eclipse.org/openj9/job/PullRequest-Extended-JDK8-win_x86-OpenJ9/8/consoleText

on many tests:
```
TestManagementAgent_SE80_0_FAILED
TestFileLocking_SE80_0_FAILED
gcNotificationTest_GenCon_0_FAILED
...
```
to name a few.

DanHeidinga:
fyi @llxia

llxia:
I think it is related to PR #1957. It seems that due to the path change, the test cannot find exclude file.
fyi @Ailloviee, could you rollback your change? Thanks.

pshipton:
@llxia you can do this via the Revert button in #1957. Or are you looking for a partial reversion?

PR #2339, opened 2018-07-06
Avoid truncating I64 or U64 values from 32-bit core files

keithc-ca:
When working with 32-bit core files, the constructors for IDATA and UDATA will truncate the value to 32 bits: the constructors for I64 and U64 must retain all 64-bits.

I64 must not inherit longValue() from IDATA which would truncate the return value to 32-bits when working with 32-bit core files.

### DIFF
--- a/debugtools/DDR_VM/src/com/ibm/j9ddr/vm29/types/I64.java
+++ b/debugtools/DDR_VM/src/com/ibm/j9ddr/vm29/types/I64.java
@@ -34,6 +34,9 @@ public class I64 extends IDATA {
 	// Constructors
 	public I64(long value) {
 		super(value);
+		// When working with 32-bit core files, the constructor for IDATA
+		// will truncate the value to 32 bits: we need all 64-bits here.
+		this.data = value;
 	}
 	
 	public I64(Scalar parameter) {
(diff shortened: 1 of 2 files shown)


DanHeidinga:
@keithc-ca Have you considered adding a non-truncating package private constructor to IDATA / UDATA for this case?  It seems a little strange to reassign the this.data in the subclass.

keithc-ca:
I took that path originally, but arrived at this smaller change. I can switch back if you feel strongly about it.

DanHeidinga:

DanHeidinga:
Jenkins test extended win32 jdk8

keithc-ca:
In the PR build, I see two failure modes:
```
```
and
```
[TestNG] [ERROR] java.lang.IllegalArgumentException: InputStream cannot be null
Exception in thread "main" java.lang.NullPointerException
	at org.testng.TestNG.getStatus(TestNG.java:211)
	at org.testng.TestNG.main(TestNG.java:1324)
```
I don't think either of them is related to this change.

On the other hand, all the relevant *dbgddrext* tests now pass.

DanHeidinga:
The `InputStream cannot be null` looks to be the same as #1608

Issue #1275, opened 2018-02-26
ConstantDynamic: Add cycle detection for Constant dynamic args


tajila:

tajila:
@Ailloviee ^^

PR #2340, opened 2018-07-06
[WIP] Cycle Detection for Constant Dynamic

Ailloviee:
Add Cycle Detection for Constant Dynamic to check cycles that might appear in the arguments of a Bootstrap Method of a condy.

Related Issue:  [#1275](https://github.com/eclipse/openj9/issues/1275)


pshipton:
@DanHeidinga is this required for the 0.10.0 release?

DanHeidinga:
I think this can be pushed to 0.11.0 as the existing code should *crosses fingers* just Stackoverflow on cycles (which is explicitly expected by the spec)

DanHeidinga:
Moved to 0.11.0

DanHeidinga:
Moving this to 0.12.0 as the current implementation works - this is just a fast path to the spec mandated stackoverflow

genie-openj9:

PR #2343, opened 2018-07-06
Do not allow NODE to be null

AdamBrousseau:
- If NODE is set as a build parameter but is not set,
  the current logic will set NODE to params.node which
  will be null. This will produce an IllegalArgumentException
  'Null value not allowed as an environment variable: NODE'.
  The updated logic should first look to the param,
  then fetch from the variable file if it is not passed.
  It will then throw an error if neither can be determined.

[skip ci]

### DIFF
--- a/buildenv/jenkins/common/variables-functions
+++ b/buildenv/jenkins/common/variables-functions
@@ -199,9 +199,12 @@ def get_user_credentials_id(KEY) {
 */
 def set_node(job_type) {
     // fetch labels for given platform/spec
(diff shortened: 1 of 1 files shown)


AdamBrousseau:
Jenkins compile xlinux jdk8

AdamBrousseau:
adam test plinux jdk8

AdamBrousseau:
Jenkins compile xlinux jdk8

AdamBrousseau:
adam test sanity plinux jdk8

AdamBrousseau:
adam test sanity plinux jdk8

AdamBrousseau:
@vsebe please review

PR #2346, opened 2018-07-06
Improve jProfiling value sequence trees

r30shah:
This pull request consists following changes for the jProfiling.
1. Improve jPropfiling Value Sequence 
In compilations which use jProfiling, for each value we decide to profile, we generate a code that does the following.
```
if ( value == table.keys[hash(value)])
   ++freqs[hash(value)];
else if ( table.otherIndex >= 0 )
   ++freqs[otherIndex];
else
   call _jProfile(32/64)BitValue
```
In summary we first check if the value exists in hash table and if exists we increase corresponding frequency counter. In cases when value does not exist in the hash table, we need to check if we can acquire lock on table and add new value to it, other wise we need to increase other frequency. This commit optimizes above sequence by using ternary select operation to decide which frequency counter (hash(value) or otherIndex) we need to increment.

With this change, we now only have three blocks per value we profile instead of six. Also in case of profiling VFT pointer, we used to check if the object whose J9Class we are profilig is NULL or not and in case it is NULL, to avoid dereferencing NULL in profiling code we used to generate a fallBack value block. There is no need to generate this as we can simply skip the profiling code if object whose J9Class pointer we are profiling is NULL.

2. If method is queued for recompilation, we do not need to run the profiling code. This commit adds a check around the profiling code to check if method has been queued for recompilation and skips the profiling code.

### DIFF
--- a/runtime/compiler/optimizer/JProfilingValue.cpp
+++ b/runtime/compiler/optimizer/JProfilingValue.cpp
@@ -271,7 +271,7 @@ TR_JProfilingValue::addProfiling(TR::Node *value, TR::TreeTop *tt)
 
    TR_ValueProfileInfo *valueProfileInfo = TR_PersistentProfileInfo::getCurrent(comp())->findOrCreateValueProfileInfo(comp());
    TR_AbstractHashTableProfilerInfo *info = static_cast<TR_AbstractHashTableProfilerInfo*>(valueProfileInfo->getOrCreateProfilerInfo(value->getByteCodeInfo(), comp(), AddressInfo, HashTableProfiler));
-   addProfilingTrees(comp(), tt, value, info, NULL, NULL, true, trace());
+   addProfilingTrees(comp(), tt, value, info, NULL, true, trace());
    }
 
 void
@@ -284,16 +284,12 @@ TR_JProfilingValue::addVFTProfiling(TR::Node *address, TR::TreeTop *tt, bool add
       getSymRefTab()->findOrCreateVftSymbolRef());
 
    TR::Node *check = NULL;
-   TR::Node *fallback = NULL;
    if (addNullCheck)
-      {
       check = TR::Node::createif(TR::ifacmpeq, address, TR::Node::aconst(address, 0));
-      fallback = TR::Node::aconst(address, 0);
-      }
 
    TR_ValueProfileInfo *valueProfileInfo = TR_PersistentProfileInfo::getCurrent(comp())->findOrCreateValueProfileInfo(comp());
    TR_AbstractHashTableProfilerInfo *info = static_cast<TR_AbstractHashTableProfilerInfo*>(valueProfileInfo->getOrCreateProfilerInfo(address->getByteCodeInfo(), comp(), AddressInfo, HashTableProfiler));
-   addProfilingTrees(comp(), tt, vftNode, info, check, fallback, true, trace());
+   addProfilingTrees(comp(), tt, vftNode, info, check, true, trace());
    }
 
 /**
@@ -347,7 +343,7 @@ TR_JProfilingValue::lowerCalls()
          // Extract the arguments and add the profiling trees
          TR::Node *value = child->getFirstChild();
          TR_AbstractHashTableProfilerInfo *table = (TR_AbstractHashTableProfilerInfo*) child->getSecondChild()->getAddress();
-         addProfilingTrees(comp(), cursor, value, table, NULL, NULL, true, trace());
+         addProfilingTrees(comp(), cursor, value, table, NULL, true, trace());
 
          // Remove the original trees and continue from the tree after the profiling
          TR::TransformUtil::removeTree(comp(), cursor);
@@ -360,9 +356,9 @@ TR_JProfilingValue::lowerCalls()
  * Insert the trees and control flow to profile a node after an insertion point.
  * The original block will be split after the insertion point.
  *
- * An optional mapping, with a test and fallback value is supported. An example use of
+ * An optional mapping, with a test is supported. An example use of
  * this is a vft lookup using an address that could be null. A null check is therefore
- * necessary, using a fallback value of 0.
+ * necessary..
  *
  * ------------------
  * | ...            |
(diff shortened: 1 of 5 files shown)


r30shah:
@andrewcraik  These are the changes that improves a jProfiling Value Trees. This is dependent on https://github.com/eclipse/omr/pull/2730

runtime/compiler/runtime/J9Profiler.cpp
```diff
@@ -1643,7 +1643,9 @@ TR_BlockFrequencyInfo::TR_BlockFrequencyInfo(TR_CallSiteInfo *callSiteInfo, int3
    _blocks(blocks),
    _frequencies(frequencies),
    _counterDerivationInfo(NULL),
-   _entryBlockNumber(-1)
+   _entryBlockNumber(-1),
+   _isQueuedForRecompilation(0),
+   _isQueuedForRecompilationSymRef(NULL)
```
andrewcraik: This structure is a persistent structure - the symref should not be stored here - normally we would have a findOrCreate on the SymRefTab - because the pointer will be stale once the compile terminates it certainly doesn't belong here.

runtime/compiler/runtime/J9Profiler.cpp
```diff
@@ -1643,7 +1643,9 @@ TR_BlockFrequencyInfo::TR_BlockFrequencyInfo(TR_CallSiteInfo *callSiteInfo, int3
    _blocks(blocks),
    _frequencies(frequencies),
    _counterDerivationInfo(NULL),
-   _entryBlockNumber(-1)
+   _entryBlockNumber(-1),
+   _isQueuedForRecompilation(0),
```
andrewcraik: I'm guessing there is no easy existing flag for this? If there isn't why is it on the block info and not the body info? This bit will be stale if kept on the blockinfo post recompilation because we keep the last good blockinfo structure around until the next profiling structure is populated with enough data.

runtime/compiler/optimizer/JProfilingValue.cpp
```diff
@@ -491,171 +488,127 @@ TR_JProfilingValue::addProfilingTrees(
...
-      TR::TreeTop::create(comp, fallback->getEntry(), TR::Node::create(example, TR::Goto, 0, quickTest->getEntry()));
-      TR::TreeTop::create(comp, fallback->getEntry(), storeNode(comp, fallbackValue, storedValueSymRef));
+      TR::SymbolReference *storedObjectSymRef = NULL;
+      // We add profiling trees before checkCast or instanceOf node and to avoid commoning out object while splitting blocks,
+      // object is stored in temp slot and subbsequent reference to that object is replaced by loading from that temp slot
+      TR::TreeTop *storeValue = TR::TreeTop::create(comp, insertionPoint, storeNode(comp, value->getFirstChild(), storedObjectSymRef));
+      replaceNode(comp, extendedBlock->getEntry(), storeValue->getNextTreeTop(),
+         value->getFirstChild(), TR::Node::createLoad(value->getFirstChild(), storedObjectSymRef));
+      // VFT pointer is required to stored in temp slot for helper call.
+      quickTest = originalBlock->split(storeValue->getNextTreeTop(), cfg, true, true);
```
andrewcraik: We may not always be profiling a vft pointer - will this logic always be right? I suspect what you mean is that the value to be profiled needs to be stored to a temp so it can be read in the helper call block, but I wanted to check if there was something vft specific

runtime/compiler/optimizer/JProfilingValue.cpp
```diff
@@ -491,171 +488,127 @@ TR_JProfilingValue::addProfilingTrees(
...
+      TR::TreeTop *storeVFTPointer = TR::TreeTop::create(comp, quickTest->getEntry(), storeNode(comp, value, storedValueSymRef));
```
andrewcraik: why is this called storeVFTPointer - I could be profiling an integer value...

runtime/compiler/optimizer/JProfilingValue.cpp
```diff
@@ -491,171 +488,127 @@ TR_JProfilingValue::addProfilingTrees(
...
-   // Load the lock and test if its set
-   TR::Node *lockAddress = TR::Node::aconst(example, table->getBaseAddress() + table->getLockOffset());
-   TR::Node *lock = loadValue(comp, lockType, lockAddress);
-   TR::Node *checkTableLock = TR::Node::createif(TR::ifscmplt, lock, TR::Node::sconst(example, 0));
-   TR::TreeTop::create(comp, slowTest->getEntry(), checkTableLock);
-
-   /********************* slowInc Block *********************/
-   TR::Node *slowIncIndex = convertType(lock, systemType, false);
-   if (!extendBlocks)
+   TR::Block *mainlineReturn = quickInc->split(incTree->getNextTreeTop(), cfg, false, true);
```
andrewcraik: I don't see how you can do fixupCommoning = false and have a merge - you have to uncommon to merge...

runtime/compiler/optimizer/JProfilingValue.cpp
```diff
@@ -491,171 +488,127 @@ TR_JProfilingValue::addProfilingTrees(
...
+   // Test field isQueuedForRecompilation of BlockFrequencyInfo for profiling compilations.
+   // If we have already queued this method for recompilation means we have enough profiling data
+   // s.t we can skip running this code.
+   static bool disableSkipProfilingCode = (feGetEnv("TR_DisableSkipProfilingValueCode") != NULL);
```
andrewcraik: can we give this a more helpful name? TR_DontGenerateJProfilingRecompQueueTest or osmehting? The env var really doesn't say what it is - sounds like we are going to not generate the profiling code or something to me.

runtime/compiler/optimizer/JProfilingValue.cpp
```diff
@@ -491,171 +488,127 @@ TR_JProfilingValue::addProfilingTrees(
...
-   if (trace)
-      traceMsg(comp, "  Slow increment in block_%d\n", slowInc->getNumber());
-
-   // Add the other increment and return to mainline
-   TR::TreeTop::create(comp, slowInc->getEntry(), TR::Node::create(example, TR::Goto, 0, mainlineReturn->getEntry()));
-   TR::Node *counterAddress = TR::Node::aconst(example, table->getBaseAddress() + table->getFreqOffset());
-   TR::TreeTop::create(comp, slowInc->getEntry(), incrementMemory(comp, counterType,
-      effectiveAddress(counterType, counterAddress, slowIncIndex)));
-
+   //mainlineReturn->setIsExtensionOfPreviousBlock();
```
andrewcraik: please clean up dead code

runtime/compiler/runtime/J9Profiler.cpp
r30shah: Removed SymRef : https://github.com/eclipse/openj9/pull/2346/commits/42b3027ce7646151036f0700b8d9ce9e90f40197

runtime/compiler/runtime/J9Profiler.cpp
r30shah: I checked into PersistentMethodBodyInfo and PersistentJittedBodyInfo. Could not find any flag that says this guy is being recompiled now.
When implementing this facility, I found two possible candidates where we can put this information. One is BlockFrequencyInfo and one is PersistentJittedInfo.
I wanted to have it stored in Class that is related to profiling. If I use persistentJittedInfo then I would have increases a size for all methods. Consumer of this field is limited to profiling compilations using jProfiling (JProfilingValue, JProfilingBlock and JProfilingRecompLoopTest) generates test to check if we have queued for recompilation in profiling compilations using jProfiling. So I chose to put this in TR_BlockFrequencyInfo.

r30shah:
@andrewcraik  Made most of the changes requested in last commit.  Can I get feedback on the changes?

runtime/compiler/runtime/J9Profiler.hpp
```diff
@@ -637,6 +637,8 @@ class TR_BlockFrequencyInfo
    bool    isJProfilingData() { return _counterDerivationInfo != NULL; }
    static int32_t *getEnableJProfilingRecompilation() { return &_enableJProfilingRecompilation; }
    static void    enableJProfilingRecompilation() { _enableJProfilingRecompilation = -1; }
+   void setIsQueuedForRecompilation() { _isQueuedForRecompilation = -1; }
```
fjeremic: Why is this not a boolean?

r30shah: This flag is intended to use in JIT compiled code to check if method is Queued for recompilation. It will generate following trees. 
```
ificmpeq -> skipCode
      iload _isQueuedForRecompilation
      iconst -1
```
I am not sure a boolean here serves purpose. Do you have trees in mind which I can use here?

fjeremic: I see. We should document this as it is non-obvious. Curious about the choice of `-1` though. I figure comparing against `0` would be more efficient on most platforms including Z where we can just test this value in memory without loading it in a register at all.

r30shah: On Z We can generate in memory compare immediate if we found the immediate value lies within the range in which we can encode it in the instruction so -1 would be in immediate field to compare with in memory. 
Verifying the instructions generated for above sequence on Z,
```
LARL R, &isQueuedForRecompilation
CHSI 0(R),0xFFFF
```
To answer your question, There is no particular reason behind using -1.

fjeremic: Thanks for clarifying. I forgot about `CHSI`. Indeed it will work with most integers on Z.

andrewcraik:
This code is dormant but compiled on all platforms so I'm going to run sanity to be sure we don't break any builds with this. Once clean I'll merge

andrewcraik:
Jenkins test sanity

r30shah:
Windows errors seems like happening because of this: 
```
01:50:41 NPT ERROR: Cannot open library
01:50:41 
01:50:41 cmdLineTester_jvmtitests_extended_7_FAILED
```
I do not think so it happens because of these changes. Going to squash all the commits.

fjeremic:
> Windows errors seems like happening because of this:

This is the infamous issue in #2129.

fjeremic:
@r30shah please remove the WIP when this is ready to merge.

Jenkins test sanity

fjeremic:

PR #2348, opened 2018-07-09
Add recompilation test in loop in jProfiling

r30shah:
When jProfiling in compilation is enabled, add a test after aynchcheck node in loop to calculate the frequency and trip up recompilation of the method if reached certain threshold.

### DIFF
--- a/runtime/compiler/build/files/common.mk
+++ b/runtime/compiler/build/files/common.mk
@@ -43,6 +43,7 @@ JIT_PRODUCT_BACKEND_SOURCES+=\
     compiler/optimizer/JitProfiler.cpp \
     compiler/optimizer/JProfilingBlock.cpp \
     compiler/optimizer/JProfilingValue.cpp \
+    compiler/optimizer/JProfilingRecompLoopTest.cpp \
     compiler/optimizer/LiveVariablesForGC.cpp \
     compiler/optimizer/LoopAliasRefiner.cpp \
     compiler/optimizer/MonitorElimination.cpp \
--- a/runtime/compiler/optimizer/J9TransformUtil.cpp
+++ b/runtime/compiler/optimizer/J9TransformUtil.cpp
@@ -39,6 +39,22 @@
 #include "ras/DebugCounter.hpp"
 #include "j9.h"
 #include "optimizer/OMROptimization_inlines.hpp"
+#include "optimizer/Structure.hpp"
+
+/**
+ * Walks the TR_RegionStructure counting loops to get the nesting depth of the block
+ */ 
+int32_t J9::TransformUtil::getLoopNestingDepth(TR::Compilation *comp, TR::Block *block)
+   {
+   TR_RegionStructure *region = block->getParentStructureIfExists(comp->getFlowGraph());
+   int32_t nestingDepth = 0;
+   while (region && region->isNaturalLoop())
+      {
+      nestingDepth++;
+      region = region->getParent();
+      }
+   return nestingDepth;
+   }
 
 /*
  * Generate trees for call to jitRetranslateCallerWithPrep to trigger recompilation from JIT-Compiled code.
--- a/runtime/compiler/optimizer/JProfilingBlock.cpp
+++ b/runtime/compiler/optimizer/JProfilingBlock.cpp
@@ -41,7 +41,6 @@
 int32_t TR_JProfilingBlock::nestedLoopRecompileThreshold = 10;
 int32_t TR_JProfilingBlock::loopRecompileThreshold = 250;
 int32_t TR_JProfilingBlock::recompileThreshold = 500;
-int32_t TR_JProfilingBlock::profilingCompileThreshold = 2;
 
 /**
  * Prim's algorithm to compute a Minimum Spanning Tree traverses the edges of the tree
@@ -831,9 +830,9 @@ void TR_JProfilingBlock::dumpCounterDependencies(TR_BitVector **componentCounter
  * appropriate number of method entries has occurred as determined by the raw block
  * count of the first block of the method.
  */   
-void TR_JProfilingBlock::addRecompilationTests(TR_BlockFrequencyInfo *blockFrequencyInfo, TR_BitVector **componentCounters)
+void TR_JProfilingBlock::addRecompilationTests(TR_BlockFrequencyInfo *blockFrequencyInfo)
    {
-   // add invocation check to the top of the method
+  // add invocation check to the top of the method
    int32_t *thresholdLocation = NULL;
    if (comp()->getMethodSymbol()->mayHaveNestedLoops())
       thresholdLocation = &nestedLoopRecompileThreshold;
@@ -842,97 +841,47 @@ void TR_JProfilingBlock::addRecompilationTests(TR_BlockFrequencyInfo *blockFrequ
    else
       thresholdLocation = &recompileThreshold;
 
-   // Profiling compilations have a lower threshold, so that less time is
-   // spent running the high overhead implementation
-   if (comp()->isProfilingCompilation())
-      thresholdLocation = &profilingCompileThreshold;
-
    int32_t startBlockNumber = comp()->getStartBlock()->getNumber();
    blockFrequencyInfo->setEntryBlockNumber(startBlockNumber);
-
    TR::Node *node = comp()->getMethodSymbol()->getFirstTreeTop()->getNode();
-
-   if (componentCounters[startBlockNumber * 2] && (((uintptr_t)componentCounters[startBlockNumber * 2]) & 0x1 == 1 || !componentCounters[startBlockNumber * 2]->isEmpty()))
+   TR::Node *root = blockFrequencyInfo->generateBlockRawCountCalculationSubTree(comp(), startBlockNumber, node);
+   bool isProfilingCompilation = comp()->isProfilingCompilation();
+   if (root != NULL)
       {
-      TR::DebugCounter::incStaticDebugCounter(comp(), TR::DebugCounter::debugCounterName(comp(), "jprofiling.instrument/success/(%s)", comp()->signature()));
-      comp()->getFlowGraph()->setStructure(NULL);
-      // add the positive counters
-      TR::Node *addRoot = NULL;
-      if (((uintptr_t)componentCounters[startBlockNumber * 2]) & 0x1 == 1)
-         {
-         TR::SymbolReference *symRef = comp()->getSymRefTab()->createKnownStaticDataSymbolRef(blockFrequencyInfo->getFrequencyForBlock(((uintptr_t)componentCounters[startBlockNumber * 2]) >> 1), TR::Int32);
-         addRoot = TR::Node::createWithSymRef(node, TR::iload, 0, symRef);
-         }
-      else
-         {
-         TR_BitVectorIterator addBVI(*(componentCounters[startBlockNumber * 2]));
-         while (addBVI.hasMoreElements())
-            {
-            TR::SymbolReference *symRef = comp()->getSymRefTab()->createKnownStaticDataSymbolRef(blockFrequencyInfo->getFrequencyForBlock(addBVI.getNextElement()), TR::Int32);
-            TR::Node *counterLoad = TR::Node::createWithSymRef(node, TR::iload, 0, symRef);
-            if (addRoot)
-               addRoot = TR::Node::create(node, TR::iadd, 2, addRoot, counterLoad);
-            else
-               addRoot = counterLoad;
-            }
-         }
-      TR::Node *subRoot = NULL;
-      if (componentCounters[startBlockNumber * 2 + 1] != NULL)
-         {
-         if (((uintptr_t)componentCounters[startBlockNumber * 2 + 1]) & 0x1 == 1)
-            {
-            TR::SymbolReference *symRef = comp()->getSymRefTab()->createKnownStaticDataSymbolRef(blockFrequencyInfo->getFrequencyForBlock(((uintptr_t)componentCounters[startBlockNumber * 2 + 1]) >> 1), TR::Int32);
-            subRoot = TR::Node::createWithSymRef(node, TR::iload, 0, symRef);
-            }
-         else
-            {
-            TR_BitVectorIterator subBVI(*(componentCounters[startBlockNumber * 2 + 1]));
-            while (subBVI.hasMoreElements())
-               {
-               TR::SymbolReference *symRef = comp()->getSymRefTab()->createKnownStaticDataSymbolRef(blockFrequencyInfo->getFrequencyForBlock(subBVI.getNextElement()), TR::Int32);
-               TR::Node *counterLoad = TR::Node::createWithSymRef(node, TR::iload, 0, symRef);
-               if (subRoot)
-                  {
-                  subRoot = TR::Node::create(node, TR::isub, 2, subRoot, counterLoad);
-                  }
-               else
-                  {
-                  subRoot = counterLoad;
-                  }
-               }
-            }
-         }
-      TR::Node *root = addRoot;
-      if (subRoot)
-         {
-         root = TR::Node::create(node, TR::isub, 2, root, subRoot);
-         }
-
       TR::Block * originalFirstBlock = comp()->getStartBlock();
 
       TR::Block *guardBlock1 = TR::Block::createEmptyBlock(node, comp(), originalFirstBlock->getFrequency());
          {
-         TR::SymbolReference *symRef = comp()->getSymRefTab()->createKnownStaticDataSymbolRef(blockFrequencyInfo->getEnableJProfilingRecompilation(), TR::Int32);
+         // If this is profiling compilation we do not need to check if jProfiling is enabled or not at runtime,
+         // In this case we only check if we have queued for recompilation before comparing against method invocation count.
+         int32_t *loadAddress = isProfilingCompilation ? blockFrequencyInfo->getIsQueuedForRecompilation() : blockFrequencyInfo->getEnableJProfilingRecompilation();
+         TR::SymbolReference *symRef = comp()->getSymRefTab()->createKnownStaticDataSymbolRef(loadAddress, TR::Int32);
          TR::Node *enableLoad = TR::Node::createWithSymRef(node, TR::iload, 0, symRef);
-         TR::Node *enableTest = TR::Node::createif(TR::ificmpeq, enableLoad, TR::Node::iconst(node, 0), originalFirstBlock->getEntry());
+         TR::Node *enableTest = TR::Node::createif(TR::ificmpeq, enableLoad, TR::Node::iconst(node, -1), originalFirstBlock->getEntry());
          TR::TreeTop *enableTree = TR::TreeTop::create(comp(), enableTest);
          enableTest->setIsProfilingCode();
          guardBlock1->append(enableTree);
          }
 
+      static int32_t jProfilingCompileThreshold = comp()->getOptions()->getJProfilingMethodRecompThreshold();
+      if (trace())
+         traceMsg(comp(),"Profiling Compile Threshold for method = %d\n",isProfilingCompilation ? jProfilingCompileThreshold : *thresholdLocation);
       TR::Block *guardBlock2 = TR::Block::createEmptyBlock(node, comp(), originalFirstBlock->getFrequency());
-      TR::Node *recompThreshold = TR::Node::createWithSymRef(node, TR::iload, 0, comp()->getSymRefTab()->createKnownStaticDataSymbolRef(thresholdLocation, TR::Int32));
+      TR::Node *recompThreshold = isProfilingCompilation ? TR::Node::iconst(node, jProfilingCompileThreshold) : TR::Node::createWithSymRef(node, TR::iload, 0, comp()->getSymRefTab()->createKnownStaticDataSymbolRef(thresholdLocation, TR::Int32));
       TR::Node *cmpFlagNode = TR::Node::createif(TR::ificmplt, root, recompThreshold, originalFirstBlock->getEntry());
       TR::TreeTop *cmpFlag = TR::TreeTop::create(comp(), cmpFlagNode);
       cmpFlagNode->setIsProfilingCode();
       guardBlock2->append(cmpFlag);
 
       // construct call block
+      const char * const dc1 = TR::DebugCounter::debugCounterName(comp(),
+         "methodRecomp/(%s)", comp()->signature());
       TR::Block *callRecompileBlock = TR::Block::createEmptyBlock(node, comp(), UNKNOWN_COLD_BLOCK_COUNT);
       callRecompileBlock->setIsCold(true);
       TR::TreeTop *callTree = TR::TransformUtil::generateRetranslateCallerWithPrepTrees(node, TR_PersistentMethodInfo::RecompDueToJProfiling, comp());
       callTree->getNode()->setIsProfilingCode();
       callRecompileBlock->append(callTree);
+      TR::DebugCounter::prependDebugCounter(comp(), dc1, callTree);
       
       comp()->getRecompilationInfo()->getJittedBodyInfo()->setUsesJProfiling();
       TR::CFG *cfg = comp()->getFlowGraph();
@@ -1156,10 +1105,9 @@ int32_t TR_JProfilingBlock::perform()
       
    // dump counter dependency information
    if (trace())
-      dumpCounterDependencies(componentCounters);
-
+      dumpCounterDependencies(componentCounters); 
    // modify the method to add tests to trigger recompilation at runtime
-   addRecompilationTests(blockFrequencyInfo, componentCounters);
+   addRecompilationTests(blockFrequencyInfo);
    return 1;
    }
 
--- a/runtime/compiler/optimizer/JProfilingBlock.hpp
+++ b/runtime/compiler/optimizer/JProfilingBlock.hpp
@@ -51,7 +51,6 @@ class TR_JProfilingBlock : public TR::Optimization
    static int32_t nestedLoopRecompileThreshold;
    static int32_t loopRecompileThreshold;
    static int32_t recompileThreshold;
-   static int32_t profilingCompileThreshold;
    TR_JProfilingBlock(TR::OptimizationManager *manager)
       : TR::Optimization(manager)
       {}
@@ -62,13 +61,12 @@ class TR_JProfilingBlock : public TR::Optimization
 
    virtual int32_t perform();
    virtual const char * optDetailString() const throw();
-
    protected:
    void computeMinimumSpanningTree(BlockParents &parents, BlockPriorityQueue &Q, TR::StackMemoryRegion &stackMemoryRegion);
    int32_t processCFGForCounting(BlockParents &parent, TR::BlockChecklist &countedBlocks, TR::CFGEdge &loopBack);
    TR_BlockFrequencyInfo *initRecompDataStructures();
    void dumpCounterDependencies(TR_BitVector **componentCounters);
-   void addRecompilationTests(TR_BlockFrequencyInfo *blockFrequencyInfo, TR_BitVector **componentCounters);
+   void addRecompilationTests(TR_BlockFrequencyInfo *blockFrequencyInfo);
    };
 
 #endif
--- a/runtime/compiler/optimizer/JProfilingRecompLoopTest.cpp
+++ b/runtime/compiler/optimizer/JProfilingRecompLoopTest.cpp
@@ -1,0 +1,249 @@
+/*******************************************************************************
+ * Copyright (c) 2018, 2018 IBM Corp. and others
+ *
+ * This program and the accompanying materials are made available under
+ * the terms of the Eclipse Public License 2.0 which accompanies this
+ * distribution and is available at https://www.eclipse.org/legal/epl-2.0/
+ * or the Apache License, Version 2.0 which accompanies this distribution and
+ * is available at https://www.apache.org/licenses/LICENSE-2.0.
+ *
+ * This Source Code may also be made available under the following
+ * Secondary Licenses when the conditions for such availability set
+ * forth in the Eclipse Public License, v. 2.0 are satisfied: GNU
+ * General Public License, version 2 with the GNU Classpath
+ * Exception [1] and GNU General Public License, version 2 with the
+ * OpenJDK Assembly Exception [2].
+ *
+ * [1] https://www.gnu.org/software/classpath/license.html
+ * [2] http://openjdk.java.net/legal/assembly-exception.html
+ *
+ * SPDX-License-Identifier: EPL-2.0 OR Apache-2.0 OR GPL-2.0 WITH Classpath-exception-2.0 OR LicenseRef-GPL-2.0 WITH Assembly-exception
+ *******************************************************************************/
+#include "JProfilingRecompLoopTest.hpp"
+#include "il/Block.hpp"
+#include "infra/Cfg.hpp"
+#include "infra/TRCfgEdge.hpp"
+#include "infra/Checklist.hpp"
+#include "infra/ILWalk.hpp"
+#include "infra/List.hpp"
+#include "optimizer/Optimization_inlines.hpp"
+#include "il/Node_inlines.hpp"
+#include "infra/Checklist.hpp"             // for TR::NodeChecklist
+#include "ras/DebugCounter.hpp"
+#include "control/Recompilation.hpp"              // for TR_Recompilation, etc
+#include "control/RecompilationInfo.hpp"              // for TR_Recompilation, etc
+#include "codegen/CodeGenerator.hpp"
+#include "optimizer/TransformUtil.hpp"            // for TransformUtil
+#include "optimizer/JProfilingBlock.hpp"
+
+int32_t TR_JProfilingRecompLoopTest::maxLoopRecompilationThreshold = 10000;
+/**
+ * A utility function that iterates through the TR::list of byte code info and checks if passed bytecode info exists in list
+ * by checking bytecode index and caller index of the bytecode info
+ */ 
+bool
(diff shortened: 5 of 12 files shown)


r30shah:
This PR depends on https://github.com/eclipse/omr/pull/2731

r30shah:
@andrewcraik @fjeremic Can you please review this.

fjeremic:
@r30shah can you please fix the copyright checks?

r30shah:
@fjeremic this complains about copyright issue with `runtime/compiler/runtime/J9Profiler.cpp` .
As I have fixed that in [PR](https://github.com/eclipse/openj9/pull/2346), if I change copyright here as well then it will cause merge conflicts. I have kept both PR as WIP. I will update them and remove `WIP` in sequence s.t we can avoid conflicts.

runtime/compiler/optimizer/JProfilingBlock.hpp
```diff
@@ -62,7 +61,7 @@ class TR_JProfilingBlock : public TR::Optimization
 
    virtual int32_t perform();
    virtual const char * optDetailString() const throw();
-
+   static TR::Node *generateBlockRawCountCalculationNode(TR::Compilation *comp, int32_t blockNumber, TR::Node *node, TR_BlockFrequencyInfo *bfi, TR_BitVector **componentCounters = NULL);
```
andrewcraik: This belongs on the BlockInfo not on the optimization that adds the counters - the block INFO can derive all this...

andrewcraik: If not on the BlockInfo then on a utility class - it is bad form to have one opt reference another which is why this would have been added.

runtime/compiler/optimizer/JProfilingRecompLoopTest.hpp
```diff
@@ -0,0 +1,57 @@
...
+ * where we are spending a lot of time in loops, we are hitting a risk where we
+ * are running into compiled method body with profiling trees if we just rely on
+ * method invocations. This optimization scans the method body and puts 
+ * recompilation test after asynccheck node to consider number of times a loop
+ * is running.
+ */
+class TR_JProfilingRecompLoopTest : public TR::Optimization
+   {
+   public:
+   typedef TR::typed_allocator<std::pair<std::pair<TR::TreeTop*, TR::Block*>, int32_t>, TR::Region &> RecompilationTestLocationAllocator;
```
andrewcraik: can you typedef the std::pair<TR::TreeTop*, TR::Block*> to some kind of helpful name and use it in the typedefs here to make this code easier to understand - what does this pair represent conceptually?

runtime/compiler/optimizer/JProfilingRecompLoopTest.hpp
```diff
@@ -0,0 +1,57 @@
...
+   typedef std::list<std::pair<std::pair<TR::TreeTop*, TR::Block*>, int32_t>, RecompilationTestLocationAllocator> RecompilationTestLocations;
```
andrewcraik: a deque may be more what we want here than a list to save footprint and boost perf...

runtime/compiler/optimizer/JProfilingRecompLoopTest.hpp
```diff
@@ -0,0 +1,57 @@
...
+      : TR::Optimization(manager)
+      {}
+   static TR::Optimization *create(TR::OptimizationManager *manager)
+      {
+      return new (manager->allocator()) TR_JProfilingRecompLoopTest(manager);
+      }
+   virtual int32_t perform();
+   virtual const char *optDetailString() const throw();
+   void addRecompilationTests(TR::Compilation *comp, RecompilationTestLocations &testLocations);
+   int32_t getLoopNestingDepth(TR::Compilation *comp, TR::Block *block);
```
andrewcraik: Isn't the structure's getNestingDepth what you want to use for this? At the very lest if this doesn't yet exist (and it should) then it would need to be pulled off the opt and put somewhere in structure as a utility.

runtime/compiler/optimizer/JProfilingBlock.cpp
```diff
@@ -902,37 +887,59 @@ void TR_JProfilingBlock::addRecompilationTests(TR_BlockFrequencyInfo *blockFrequ
...
          enableTest->setIsProfilingCode();
          guardBlock1->append(enableTree);
          }
 
+      static int32_t profilingCompileThreshold = comp()->getOptions()->getJProfilingMethodRecompThreshold();
+
+      traceMsg(comp(),"Profiling Compile Threshold for method = %d\n", profilingCompileThreshold);
       TR::Block *guardBlock2 = TR::Block::createEmptyBlock(node, comp(), originalFirstBlock->getFrequency());
-      TR::Node *recompThreshold = TR::Node::createWithSymRef(node, TR::iload, 0, comp()->getSymRefTab()->createKnownStaticDataSymbolRef(thresholdLocation, TR::Int32));
+      TR::Node *recompThreshold = TR::Node::iconst(node, profilingCompileThreshold);
```
andrewcraik: Why has this been made a constant unconditionally - being able to vary the compile threshold is certainly something we would want to consider - for a profiled compile the answer may be static but for a non-profiled compile we may need some ability to vary things so I'm not sure that this is an 'improvement' - seems a bit of a hack.

runtime/compiler/optimizer/JProfilingBlock.cpp
```diff
@@ -902,37 +887,59 @@ void TR_JProfilingBlock::addRecompilationTests(TR_BlockFrequencyInfo *blockFrequ
...
+   if (root != NULL)
+      {
       TR::Block * originalFirstBlock = comp()->getStartBlock();
 
       TR::Block *guardBlock1 = TR::Block::createEmptyBlock(node, comp(), originalFirstBlock->getFrequency());
          {
-         TR::SymbolReference *symRef = comp()->getSymRefTab()->createKnownStaticDataSymbolRef(blockFrequencyInfo->getEnableJProfilingRecompilation(), TR::Int32);
-         TR::Node *enableLoad = TR::Node::createWithSymRef(node, TR::iload, 0, symRef);
-         TR::Node *enableTest = TR::Node::createif(TR::ificmpeq, enableLoad, TR::Node::iconst(node, 0), originalFirstBlock->getEntry());
+         TR::Node *enableLoad = TR::Node::createWithSymRef(node, TR::iload, 0, blockFrequencyInfo->getOrCreateSymRefForIsQueuedForRecompilation(comp()));
```
andrewcraik: This makes sense for profiled compilations - can we avoid breaking the non-profiled case?

runtime/compiler/optimizer/JProfilingBlock.cpp
```diff
@@ -35,13 +35,13 @@
 #include "control/Recompilation.hpp"              // for TR_Recompilation, etc
 #include "control/RecompilationInfo.hpp"              // for TR_Recompilation, etc
 #include "optimizer/TransformUtil.hpp"
+#include "optimizer/Structure.hpp"
```
andrewcraik: Why does this now need structure when it didn't need it before?

runtime/compiler/runtime/J9Profiler.cpp
```diff
@@ -1931,6 +1931,39 @@ TR_BlockFrequencyInfo::getRawCount(TR::ResolvedMethodSymbol *resolvedMethod, TR_
...
+ *          It is useful when we are trying to associate the profiling information with the node which might
+ *          have moved to different blocks with different bci
+ *    \param bci TR_ByteCodeInfo for which original block number is searched for
+ *    \param comp Current compilation object
+ *    \return block number of the original block bci belongs to.
+ *            WARNING: If consumer of this API uses this to to get the profiled data in later compilation and
+ *            requested BCI was not inlined before, it returns -1.
+ */
+int32_t
+TR_BlockFrequencyInfo::getOriginalBlockNumberToGetRawCount(TR_ByteCodeInfo &bci, TR::Compilation *comp)
```
andrewcraik: I don't think we want to expose this detail - the APIs that need it may belong here instead.

runtime/compiler/optimizer/JProfilingRecompLoopTest.cpp
```diff
@@ -0,0 +1,267 @@
...
+      }
+   return nestingDepth;
+   }
+
+/**
+ * A utility function that iterates through the TR::list of byte code info and checks if passed bytecode info exists in list
+ * by checking bytecode index and caller index of the bytecode info
+ */ 
+bool
+TR_JProfilingRecompLoopTest::isByteCodeInfoInCurrentTestLocationList(TR_ByteCodeInfo &bci, TR::list<TR_ByteCodeInfo, TR::Region&> &addedLocationBCIList)
```
andrewcraik: Would it be more efficient to just have an array the size of the inlining table and each entry is a pointer to a TR_BitVector initially NULL) and we just set bits in the bitvector for the offsets rather than copying BCInfos everywhere and having a list which allocates nodes etc.

andrewcraik: This function would then become index into the array - check if null, if not null check if the bit is set rather than an O(n) search of the profiled BCInfos...

runtime/compiler/optimizer/JProfilingRecompLoopTest.cpp
```diff
@@ -0,0 +1,267 @@
...
+      if (iterBCI.getByteCodeIndex() == bci.getByteCodeIndex() && iterBCI.getCallerIndex() == bci.getCallerIndex())
+         return true;   
+      }
+   return false;
+   }
+
+int32_t
+TR_JProfilingRecompLoopTest::perform()
+   {
+   if (comp()->getOption(TR_EnableJProfiling))
```
andrewcraik: This is like Homer Simpson's everything is ok alarm - it is going to log that it is working - we should log when it doesn't work and the decisions we made...

runtime/compiler/optimizer/JProfilingRecompLoopTest.cpp
```diff
@@ -0,0 +1,267 @@
...
+         if (currentBlock->getStructureOf()->getContainingLoop() != NULL)
+            {
+            TR_ByteCodeInfo bci = cursor->getNode()->getByteCodeInfo();
+            // If the list of TR_ByteCodeInfo used to track the tree tops in extended basic blocks is empty or the bytecode info
+            // of current asynccheck node does not match with any tracked bytecode info, we will add the test otherwise skip it.   
+            if (addedLocationBCIList.empty() || !isByteCodeInfoInCurrentTestLocationList(bci, addedLocationBCIList))
+               {
+               addedLocationBCIList.push_back(bci);
+               int32_t loopDepth = getLoopNestingDepth(comp(), currentBlock);
+               testLocations.push_back(std::make_pair(std::make_pair(cursor, currentBlock),loopDepth));
```
andrewcraik: why do we have to store this and can't we just get it from structure when needed?

runtime/compiler/optimizer/JProfilingRecompLoopTest.cpp
```diff
@@ -0,0 +1,267 @@
...
+ *                         And corresponding to that location, a loop nesting depth 
+ */
+
+void 
+TR_JProfilingRecompLoopTest::addRecompilationTests(TR::Compilation *comp, RecompilationTestLocations &testLocations)
+   {
+   TR_PersistentProfileInfo *profileInfo = comp->getRecompilationInfo()->findOrCreateProfileInfo();
+   TR_BlockFrequencyInfo *bfi = TR_BlockFrequencyInfo::get(profileInfo);
+   TR::CFG *cfg = comp->getFlowGraph();
+   cfg->setStructure(NULL);
```
andrewcraik: We should be able to fix structure for this really - the sequence is not that hard. I won't require it to get the first version of the code delivered as long as we agree you will look to see if you can fix structure unless it proves more costly to do that than to recompute it - I suspect for some block additions it should be easy to fix up.

runtime/compiler/optimizer/JProfilingRecompLoopTest.cpp
```diff
@@ -0,0 +1,267 @@
...
+
+   // Following environment sets up the base recompilation threshold for the loop.
+   // This base recompile threshold in conjunction with the depth in loop is compared with the raw count of the
+   // loop to decide if we have run this loop enough time to trip method recompilation. 
+   // TODO: Currently set base recompilation threshold needs some tuning.
```
andrewcraik: Not sure what this means - is it really a TODO or should this be some advice that you should tune XX parameter if you see such and such a condition?

runtime/compiler/optimizer/JProfilingRecompLoopTest.cpp
```diff
@@ -0,0 +1,267 @@
...
+   TR_BlockFrequencyInfo *bfi = TR_BlockFrequencyInfo::get(profileInfo);
+   TR::CFG *cfg = comp->getFlowGraph();
+   cfg->setStructure(NULL);
+
+   // Following environment sets up the base recompilation threshold for the loop.
+   // This base recompile threshold in conjunction with the depth in loop is compared with the raw count of the
+   // loop to decide if we have run this loop enough time to trip method recompilation. 
+   // TODO: Currently set base recompilation threshold needs some tuning. 
+   static int32_t recompileThreshold = comp->getOptions()->getJProfilingLoopRecompThreshold();
+   traceMsg(comp, "Loop Recompilation Base Threshold = %d\n",recompileThreshold);
```
andrewcraik: shouldn't this be guarded by if (trace())?

runtime/compiler/optimizer/JProfilingRecompLoopTest.cpp
```diff
@@ -0,0 +1,267 @@
...
+      TR::Node *node = asyncCheckTreeTop->getNode();
+      int32_t depth = testLocationIter->second;
+      if (trace())
+         traceMsg(comp, "block_%d, n%dn, depth = %d\n",originalBlock->getNumber(), asyncCheckTreeTop->getNode()->getGlobalIndex(), depth);
+      TR_ByteCodeInfo bci = asyncCheckTreeTop->getNode()->getByteCodeInfo();
+      
+      TR::Node *root = TR_JProfilingBlock::generateBlockRawCountCalculationNode(comp, bfi->getOriginalBlockNumberToGetRawCount(bci,comp), node, bfi);
+      
+      // If we got a bad counters/ bad block frequency info, above API would return NULL, if that is the case we can not generate a recompilation test for that location.
+      if (!root)
```
andrewcraik: Hmm isn't that a bit of a problem if we are relying on the recomp check happening? I would have hoped this would never happen... Perhaps an assert?

runtime/compiler/optimizer/JProfilingBlock.hpp
r30shah: Changed in https://github.com/eclipse/openj9/pull/2348/commits/3296fd845e8533fd57208409e3a31aa607759f69
Now this utility function belongs to `TR_BlockFrequencyInfo`

runtime/compiler/runtime/J9Profiler.cpp
r30shah: Made this Private in https://github.com/eclipse/openj9/pull/2348/commits/cfd65b3bf32bcc9a412d70c12be24218148e1279

runtime/compiler/optimizer/JProfilingRecompLoopTest.cpp
r30shah: I want to keep track of the ByteCodeInfo of the recompilation test location (asynccheck node) in `extended basic block` which will be emptied after time we encounter a block that is not extension of previous block. For this I need to match both bytecode index and caller index.
Besides this list would be limited to extended basic block containing ByteCode info of test location node.

runtime/compiler/optimizer/JProfilingRecompLoopTest.hpp
r30shah: I need to look in depth how Structure gets nesting depth. I walk parent structure till the point I find acyclic structure.

runtime/compiler/optimizer/JProfilingRecompLoopTest.cpp
r30shah: I do not think we should add an ASSERT here. 
One of the cases where I got NULL here was in method where we might inline constructors that might happen after JProfilingBlock which generates block frequency counters.
We might not have block frequency of these blocks.

runtime/compiler/optimizer/JProfilingRecompLoopTest.cpp
```diff
@@ -0,0 +1,267 @@
...
+      const char * const name = TR::DebugCounter::debugCounterName(comp,
+      "recompilationHelper/(%s)/%d",
+      comp->signature(),depth);
+      TR::DebugCounter::prependDebugCounter(comp, name, callTree);
+      
+      // Code to calculate the raw frequency of loop from block counters and comparing with the adjusted recompilation threshold.
+      // threshold for this particular test location is calculated from the base recompile threshold and the nesting depth of the loop
+      // Putting a higher threshold limit to 10K currently to prohibit running profiling body for too long.
+      int32_t threshold = recompileThreshold << ((depth-1)*1);
+      TR::Node *cmpNode = TR::Node::createif(TR::ificmple, root, TR::Node::iconst(node, threshold <= 10000 ? threshold : 10000), remainingCodeBlock->getEntry());
```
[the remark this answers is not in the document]
r30shah: 10000 limit is derived from  https://github.com/eclipse/openj9/blob/7336f3ab653526218911b922eefa8c173836320b/runtime/compiler/control/RecompilationInfo.hpp#L67 
I gave it a `maxLoopRecompilationThreshold` name in 
https://github.com/eclipse/openj9/pull/2348/commits/0d80ef6fbef26363e11cb5db13ab47a9f965dc4f

runtime/compiler/optimizer/JProfilingRecompLoopTest.cpp
```diff
@@ -0,0 +1,267 @@
...
+      TR::TreeTop *cmpFlag = TR::TreeTop::create(comp, cmpNode);
+      cmpFlag->getNode()->setIsProfilingCode();
+      remainingCodeBlock->getEntry()->insertTreeTopsBeforeMe(callRecompileBlock->getEntry(), callRecompileBlock->getExit());
+      static bool disableSkipRecompilationTestCode = (feGetEnv("TR_DisableJProfilingSkipRecompCode") != NULL);
```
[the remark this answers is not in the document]
r30shah: Changes here: https://github.com/eclipse/openj9/pull/2348/commits/2d71aa41ddd371e0671d12395208051e84ad9a31

runtime/compiler/optimizer/JProfilingRecompLoopTest.hpp
r30shah: I moved this utility function into the `J9TransformUtil` in https://github.com/eclipse/openj9/pull/2348/commits/aa1b8bc31d69fd0268cb59a8d69455505ba8ec2c

runtime/compiler/optimizer/J9TransformUtil.cpp
```diff
@@ -40,6 +40,21 @@
 #include "j9.h"
 #include "optimizer/OMROptimization_inlines.hpp"
 
+/**
+ * Walks the TR_RegionStructure counting loops to get the nesting depth of the block
+ */ 
+int32_t J9::TransformUtil::getLoopNestingDepth(TR::Compilation *comp, TR::Block *block)
```
andrewcraik: is the getNestingDepth on TR_Structure not enough?

r30shah: I tried calling `region->getNestingDepth()` , it returns `0` . Checking in more detail, it just returns `_nestingDepth` set in structure. Going in more detail, I see that a call to setNestingDepths(...) sets the nesting depth till it hits a region which is not `naturalLoop`.
I was expecting by this time in compilation it should be set.

runtime/compiler/optimizer/JProfilingRecompLoopTest.hpp
r30shah: Changed list to deque : https://github.com/eclipse/openj9/pull/2348/commits/279d5b665208f2bb0326e6e48870652562e6adae

runtime/compiler/optimizer/JProfilingRecompLoopTest.hpp
r30shah: I am collecting test location information here that incorporates TreeTop where we want to add test (asynccheck) , enclosing block (instead of calling enclosing block we get that information and store it while walking treetops ) and loop nesting depth. 
The reason I need this information here is because before we add the recompilation test and split the blocks, I set the structure to NULL and I lost information about loop nesting. 
Commenting for, https://github.com/eclipse/openj9/pull/2348#pullrequestreview-135848068
Once this is merged, I will do analysis on what is more costly -> Setting structure to NULL before splitting or fixing up structure every time we add test.
If later is more beneficial, I can clean up depth information from testLocationInfo.

runtime/compiler/runtime/J9Profiler.cpp
```diff
@@ -1931,6 +1931,39 @@ TR_BlockFrequencyInfo::getRawCount(TR::ResolvedMethodSymbol *resolvedMethod, TR_
...
+   int32_t byteCodeToSearch = resolvedMethod->getProfilingByteCodeIndex(bci.getByteCodeIndex());
+   TR_ByteCodeInfo searchBCI = bci;
+   searchBCI.setByteCodeIndex(byteCodeToSearch);
+   bool currentCallSiteInfo = TR_CallSiteInfo::getCurrent(comp) == _callSiteInfo; 
+   for (auto i=0; i < _numBlocks; ++i)
+      {
+      if (currentCallSiteInfo && _callSiteInfo->hasSameBytecodeInfo(_blocks[i], searchBCI, comp) ||
+            (!currentCallSiteInfo && _blocks[i].getCallerIndex() == searchBCI.getCallerIndex() && _blocks[i].getByteCodeIndex() == searchBCI.getByteCodeIndex()))
+         {
+         traceMsg(comp, "Get frequency from original block_%d\n", i);
```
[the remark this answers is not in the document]
r30shah: Fixed in here : https://github.com/eclipse/openj9/pull/2348/commits/849a86a8dee5b163a674b092070399883c69e384

runtime/compiler/optimizer/JProfilingRecompLoopTest.cpp
r30shah: Changed here: https://github.com/eclipse/openj9/pull/2348/commits/849a86a8dee5b163a674b092070399883c69e384

runtime/compiler/optimizer/JProfilingRecompLoopTest.cpp
r30shah: Removed TODO: I believe a method invocation 4000 and loop base threshold 2000 is tuned number that gives us the better performance

runtime/compiler/optimizer/JProfilingRecompLoopTest.cpp
r30shah: changed in https://github.com/eclipse/openj9/pull/2348/commits/849a86a8dee5b163a674b092070399883c69e384

runtime/compiler/optimizer/JProfilingRecompLoopTest.cpp
r30shah: Once we invalidate the structure, we lost the information about loop nesting. That is why we have to store the loopDepth before going on and splitting blocks.

runtime/compiler/optimizer/JProfilingBlock.cpp
r30shah: Changed it here: https://github.com/eclipse/openj9/pull/2348/commits/849a86a8dee5b163a674b092070399883c69e384

runtime/compiler/optimizer/JProfilingBlock.cpp
r30shah: Changed in https://github.com/eclipse/openj9/pull/2348/commits/849a86a8dee5b163a674b092070399883c69e384

r30shah:
@andrewcraik  Tried to address most of the comments via separate commits and comments. Let me know if I am missing something.

runtime/compiler/optimizer/J9TransformUtil.cpp
```diff
@@ -40,6 +40,21 @@
...
+   {
+   TR_RegionStructure *region = block->getParentStructureIfExists(comp->getFlowGraph());
+   int32_t nestingDepth = 0;
+   while (region && !region->isAcyclic())
```
andrewcraik: This means you count improper regions as loops for the purposes of your loop test - is this intentional?

runtime/compiler/runtime/J9Profiler.hpp
```diff
@@ -633,20 +633,22 @@ class TR_BlockFrequencyInfo
    int32_t getFrequencyInfo(TR::Block *block, TR::Compilation *comp);
    int32_t getFrequencyInfo(TR_ByteCodeInfo &bci, TR::Compilation *comp, bool normalizeForCallers = true, bool trace = true);
    void    setCounterDerivationInfo(TR_BitVector **counterDerivationInfo) { _counterDerivationInfo = counterDerivationInfo; }
+   TR_BitVector ** getCounterDerivativeInfo() { return _counterDerivationInfo; }
    void    setEntryBlockNumber(int32_t number) { _entryBlockNumber = number; }
    bool    isJProfilingData() { return _counterDerivationInfo != NULL; }
    static int32_t *getEnableJProfilingRecompilation() { return &_enableJProfilingRecompilation; }
    static void    enableJProfilingRecompilation() { _enableJProfilingRecompilation = -1; }
-
+   TR::Node* generateBlockRawCountCalculationNode(TR::Compilation *comp, int32_t blockNumber, TR::Node *node);
```
andrewcraik: Probably better to say SubTree rather than Node since it is a tree of nodes, but isn't a treetop.

runtime/compiler/runtime/J9Profiler.hpp
```diff
@@ -633,20 +633,22 @@ class TR_BlockFrequencyInfo
    int32_t getFrequencyInfo(TR::Block *block, TR::Compilation *comp);
    int32_t getFrequencyInfo(TR_ByteCodeInfo &bci, TR::Compilation *comp, bool normalizeForCallers = true, bool trace = true);
    void    setCounterDerivationInfo(TR_BitVector **counterDerivationInfo) { _counterDerivationInfo = counterDerivationInfo; }
+   TR_BitVector ** getCounterDerivativeInfo() { return _counterDerivationInfo; }
```
andrewcraik: This isn't called anywhere - why is it exposed?

runtime/compiler/optimizer/J9TransformUtil.cpp
r30shah: I walked over the parent structure till I found acyclic region. Checking the function it will return true when we have a region that does not contain any internal cycles and does not have `backedge`.

runtime/compiler/runtime/J9Profiler.hpp
r30shah: That got out of my attention. No need to have that info. Cleaned up : Changes as per PR#2348 and Fix warnings

runtime/compiler/runtime/J9Profiler.hpp
r30shah: Changed it : https://github.com/eclipse/openj9/pull/2348/commits/dcd4b9def6654084825294d884774145fe894290

runtime/compiler/optimizer/J9TransformUtil.cpp
andrewcraik: Right but do you want to count an improper region - eg one that contains cycles, but which is not a natural loop as a loop for scaling - that seems wrong to me...

runtime/compiler/optimizer/JProfilingRecompLoopTest.cpp
andrewcraik: ok so the list will be small. We may revisit this later, but I'm ok with this for now

runtime/compiler/optimizer/JProfilingRecompLoopTest.cpp
```diff
@@ -0,0 +1,245 @@
...
+      cfg->addNode(callRecompileBlock);
+      const char * const name = TR::DebugCounter::debugCounterName(comp,
+      "recompilationHelper/(%s)/%d",
+      comp->signature(),depth);
+      TR::DebugCounter::prependDebugCounter(comp, name, callTree);
+      
+      // Code to calculate the raw frequency of loop from block counters and comparing with the adjusted recompilation threshold.
+      // threshold for this particular test location is calculated from the base recompile threshold and the nesting depth of the loop
+      // Putting a higher threshold limit to 10K currently to prohibit running profiling body for too long.
+      int32_t threshold = recompileThreshold << ((depth-1)*1);
```
andrewcraik: why the *1? Should there be something to stop overflow if depth gets too deep - unlikely, but possible...

runtime/compiler/optimizer/JProfilingRecompLoopTest.cpp
```diff
@@ -0,0 +1,245 @@
...
+      TR::Node *cmpNode = TR::Node::createif(TR::ificmple, root, TR::Node::iconst(node, threshold <= maxLoopRecompilationThreshold ? threshold : maxLoopRecompilationThreshold), remainingCodeBlock->getEntry());
+      TR::TreeTop *cmpFlag = TR::TreeTop::create(comp, cmpNode);
+      cmpFlag->getNode()->setIsProfilingCode();
+      remainingCodeBlock->getEntry()->insertTreeTopsBeforeMe(callRecompileBlock->getEntry(), callRecompileBlock->getExit());
+      static bool dontGenerateJProfilingRecompQueueTest = (feGetEnv("TR_DontGenerateJProfilingRecompQueueTest") != NULL);
+      if (!dontGenerateJProfilingRecompQueueTest)
```
andrewcraik: Hmm this reads badly - not dontGenerate... Can you get rid of one of these negatives - it requires too much thinking to figure out what is going on...

runtime/compiler/optimizer/JProfilingRecompLoopTest.cpp
```diff
@@ -0,0 +1,245 @@
...
+ *                         And corresponding to that location, a loop nesting depth 
+ */
+
+void 
+TR_JProfilingRecompLoopTest::addRecompilationTests(TR::Compilation *comp, RecompilationTestLocationsInfo &testLocations)
+   {
+   TR_PersistentProfileInfo *profileInfo = comp->getRecompilationInfo()->findOrCreateProfileInfo();
+   TR_BlockFrequencyInfo *bfi = TR_BlockFrequencyInfo::get(profileInfo);
+   TR::CFG *cfg = comp->getFlowGraph();
+   cfg->invalidateStructure();
```
andrewcraik: Can we have a TODO to do a compile time evaluation to see if it is better to fix up structure or invalidate structure since fix-up should be possible given the nature of what is going on here.

andrewcraik:
@r30shah can you please fix the copyright check and address the last few comments. Code looks good. Once done could you ask @fjeremic or another committer to launch sanity? I will try and check tomorrow night to review any revisions and hopefully merge this.

r30shah:
@andrewcraik Thanks a lot for the review. Addressing your comments will be my first priority in the morning.
I will squash commits as well.

runtime/compiler/optimizer/J9TransformUtil.cpp
r30shah: Thanks for explanation. Changed it in https://github.com/eclipse/openj9/pull/2348/commits/113d8283b3fdeb03d4aa6a8ed62d1185301d2a08 
Now routine walks parent structure till it finds a regions that is not natural looop

runtime/compiler/optimizer/JProfilingRecompLoopTest.cpp
r30shah: Changed name: https://github.com/eclipse/openj9/pull/2348/commits/62454581a8c7a177249a55fcecccd3b9a5b793fa

runtime/compiler/optimizer/JProfilingRecompLoopTest.cpp
r30shah: Added TODO

r30shah:
This can be merged after merging https://github.com/eclipse/openj9/pull/2346. That pull request contains some of the supporting functions used here. Current Copyright failure is in `runtime/compiler/runtime/J9Profiler.cpp` which is also addressed in above PR so once that is merged , I will rebase ans squash all commits.

r30shah:
I will wait for https://github.com/eclipse/openj9/pull/2346 to be merged and then squash commits and remove WIP

andrewcraik:
@fjeremic can you handle this one too - I think the code should be good - if could give it a quick once over once updated the merge I would be much obliged.

r30shah:
@fjeremic I have rebased and squashed all the commits. Can you please take a look?

fjeremic:
Approving based off @andrewcraik's review.

fjeremic:
Jenkins test sanity

fjeremic:

runtime/compiler/build/files/common.mk
```diff
@@ -43,6 +43,7 @@ JIT_PRODUCT_BACKEND_SOURCES+=\
     compiler/optimizer/JitProfiler.cpp \
     compiler/optimizer/JProfilingBlock.cpp \
     compiler/optimizer/JProfilingValue.cpp \
+    compiler/optimizer/JProfilingRecompLoopTest.cpp \
```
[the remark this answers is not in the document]
r30shah: Yeah it will be needed. Wasn't aware of this part. Going to create a PR for that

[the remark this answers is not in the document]
r30shah: Created a separate PR : https://github.com/eclipse/openj9/pull/2413 
Thanks @mstoodle  for noticing this.

PR #2349, opened 2018-07-09
Comment and reword assert in Recompilation for jProfiling

r30shah:
jProfiling should not use either sampling or counting to trip recompilation. For this kind of method bodies, we have emitted recompilation test to check method invocation count and loop frequency against tuned threshold to trip recompilation.

### DIFF
--- a/runtime/compiler/control/J9Recompilation.cpp
+++ b/runtime/compiler/control/J9Recompilation.cpp
@@ -207,6 +207,10 @@ J9::Recompilation::beforeOptimization()
    //
    if (self()->isProfilingCompilation()) // this asks the bodyInfo
       {
+      // JProfiling should not use either sampling or counting mechanism to trip recompilation.
+      // Even though _useSampling is set to true here, at the end of compilation we disable sampling
+      // causing us to rely only on the in-body recompilation test that relies on block frequency counter
+      // to trip recompilation making sure we have enough profiling data to use in recompilation.
       _useSampling = _compilation->getProfilingMode() != JitProfiling;
       self()->findOrCreateProfileInfo()->setProfilingCount     (DEFAULT_PROFILING_COUNT);
       self()->findOrCreateProfileInfo()->setProfilingFrequency (DEFAULT_PROFILING_FREQUENCY);
--- a/runtime/compiler/p/codegen/PPCRecompilation.cpp
+++ b/runtime/compiler/p/codegen/PPCRecompilation.cpp
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2000, 2017 IBM Corp. and others
+ * Copyright (c) 2000, 2018 IBM Corp. and others
  *
  * This program and the accompanying materials are made available under
  * the terms of the Eclipse Public License 2.0 which accompanies this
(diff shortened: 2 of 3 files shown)


r30shah:
@andrewcraik  @fjeremic  This PR contains changes related to switching to counting based recompilation for jProfiling body similar to JIT Profiling. Can you please review this?

runtime/compiler/p/codegen/PPCRecompilation.cpp
```diff
@@ -167,7 +167,7 @@ TR::Instruction *TR_PPCRecompilation::generatePrologue(TR::Instruction *cursor)
       else
          {
          // This only applies to JitProfiling, as JProfiling uses sampling
-         TR_ASSERT(_compilation->getProfilingMode() == JitProfiling, "JProfiling should use sampling to trigger recompilation");
+         //TR_ASSERT(_compilation->getProfilingMode() == JitProfiling, "JProfiling should use sampling to trigger recompilation");
```
andrewcraik: Don't comment out asserts - if it is supposed to be removed please delete it

runtime/compiler/x/codegen/X86Recompilation.cpp
```diff
@@ -182,7 +182,7 @@ TR::Instruction *TR_X86Recompilation::generatePrologue(TR::Instruction *cursor)
          else
             {
             // This only applies to JitProfiling, as JProfiling uses sampling
-            TR_ASSERT(_compilation->getProfilingMode() == JitProfiling, "JProfiling should use sampling to trigger recompilation");
+            //TR_ASSERT(_compilation->getProfilingMode() == JitProfiling, "JProfiling should use sampling to trigger recompilation");
```
andrewcraik: Same here

runtime/compiler/control/J9Recompilation.cpp
```diff
@@ -207,7 +207,8 @@ J9::Recompilation::beforeOptimization()
    //
    if (self()->isProfilingCompilation()) // this asks the bodyInfo
       {
-      _useSampling = _compilation->getProfilingMode() != JitProfiling;
+      //_useSampling = _compilation->getProfilingMode() != JitProfiling;
+      _useSampling = false;
```
andrewcraik: Please don't leave old code around - if this should be false make it false.

andrewcraik:
The expert on compilation control is @mpirvu so we should get his thoughts on this change. It seems reasonable to me, but I am too close to the change to make an objective assessment.

andrewcraik:
@r30shah please fix the copyright check

andrewcraik:
@mpirvu I think a review from you on this would be best.

mpirvu:
On the surface the code looks good, but I remember that long time ago we wanted to retire counting compilations completely.
What is the problem that this commit tries to solve? The recompilation is triggered with jitRetranslateCallerWithPrep and not based on actual counting. If sampling gets in the way of the logic we could still create a "sampling" body (the choice between "sampling" and "counting" changes the prologue of the method) but mark the sampling as disabled.

andrewcraik:
@mpirvu So JProfiling wants to use its own counters to determine when the method should be recompiled. We don't want sampling to see the slower profiling body and trigger a compilation before we are ready if that makes sense.

mpirvu:
Sampling for "sampling body" can be disabled with bodyInfo->setDisableSampling(true); we do this in a few places in the code.
Note that I am not opposed to using counting bodies; I just don't find a compelling reason to do so.

andrewcraik:
@r30shah please fix the copyright issue so this can be merged

andrewcraik:
@mpirvu could you handle this one since I'm going to be away?

r30shah:
Fixwed copyright and also title. This just rewords the Assert we had in Recompilation for both x86 and PPC platform in jProfiling.

mpirvu:
jenkins test all

r30shah:
Windows error is because of https://github.com/eclipse/openj9/issues/2129

PR #2353, opened 2018-07-09
Atomic-free JNI work

gacholio:
Implement FlushProcessWriteBuffers for AIX. Note this implementation
differs from Linux in that mlock is not called on the guard page (mlock
on AIX is only available to root-privilege users). 


runtime/vm/FlushProcessWriteBuffers.cpp
```diff
@@ -37,50 +37,60 @@ extern "C" {
...
-	int mprotectrc = mprotect(addr, pageSize, PROT_READ | PROT_WRITE);
-	Assert_VM_true(0 == mprotectrc);
-	VM_AtomicSupport::add((UDATA*)addr, 1);
-	mprotectrc = mprotect(addr, pageSize, PROT_NONE);
-	Assert_VM_true(0 == mprotectrc);
-	omrthread_monitor_exit(vm->flushMutex);
-#else /* LINUX */
+	if (NULL != vm->flushFunction) {
+		((VOID (WINAPI*)(void))vm->flushFunction)();
+	}
```
charliegracie: Should there be some runtime failure if the function does not exist? Similarly below for the mutex?

gacholio: No, the point is to be able to acquire exclusive before it has been initialized, on the assumption that we're single-threaded.  See comment above.

gacholio: If the initialize fails (can't allocate mutex/find function), the VM fails to start.

DanHeidinga:
@charliegracie Are you good with this as well?

DanHeidinga:
Depends on eclipse/omr#2729

runtime/vm/FlushProcessWriteBuffers.cpp
charliegracie: Sorry missed the comment.

DanHeidinga:
Compile-only build was sufficient as this only affects AIX and is under #ifdefs that aren't yet enabled.

PR #2358, opened 2018-07-09
Remove index in memory references used in caseConversionHelper

dhong44:
In case conversion helper for z codegen, VLL and VSTL
instructions used a memory reference with an index register.
Since the instructions do not support an index register,
an implicit LA instruction using a new virtual register was
created. Since addressOffset is zero before the first
VLL and VSTL instruction, the index register can be omitted.


dhong44:
@fjeremic

fjeremic:
Good catch, thanks for this change!

fjeremic:
Good catch, thanks for this change!

Jenkins test sanity zlinux

PR #2359, opened 2018-07-09
Code cleanup related to exception related helpers

0dvictor:
1. Remove following three helpers that is not in-use:
   TR_IncompatibleClassChangeError
   TR_AbstractMethodError
   TR_IllegalAccessError
2. Consolidate TR_throwCurrentException on X86, aligning with other platforms.

### DIFF
--- a/runtime/compiler/runtime/Runtime.cpp
+++ b/runtime/compiler/runtime/Runtime.cpp
@@ -1063,9 +1063,6 @@ void initializeCodeRuntimeHelperTable(J9JITConfig *jitConfig, char isSMP)
 #endif
    SET(TR_throwCurrentException,      (void *)jitThrowCurrentException,      TR_Helper);
 
-   SET(TR_IncompatibleClassChangeError,(void *)jitThrowIncompatibleClassChangeError, TR_Helper);
-   SET(TR_AbstractMethodError,        (void *)jitThrowAbstractMethodError,           TR_Helper);
-   SET(TR_IllegalAccessError,         (void *)jitThrowIllegalAccessError,            TR_Helper);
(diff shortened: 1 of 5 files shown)


andrewcraik:
is this reducing the flexibility of the codegen or implementation to be able to throw these exceptions from the JIT'd code? if so why would we want to get rid of that ability?

0dvictor:
The 3 exception throwing helpers are never used and hence have not been tested for awhile. Currently there is no infrastructure to test any of them so that the future usability is not guaranteed.
In addition, I do not see such a need to throw any of these exceptions in the foreseeable future. Should any of them is needed in the future, it can be easily re-introduced. One should ensure the correctness when any of the helpers is re-introduced.

andrewcraik:
Jenkins test sanity xlinux

andrewcraik:
Jenkins test sanity win

andrewcraik:
Jenkins test sanity win32

0dvictor:
Windows failures due to #2129

andrewcraik:
this actually affects all platforms so I'm going to relaunch a full sanity

andrewcraik:
Jenkins test sanity

andrewcraik:
@fjeremic I think this is ok - at least for x86 - could you review for z @gita-omr FYI

gita-omr:
@0dvictor did you check that those helpers are not use on *any* platform? I just don't want to do all the search if you already done it. I think in the item like that it would be good to indicate what (if anything) other codegens should do.

0dvictor:
> did you check that those helpers are not use on any platform? I just don't want to do all the search if you already done it. I think in the item like that it would be good to indicate what (if anything) other codegens should do.

Yes, I searched the three helpers from the whole OpenJ9 and OMR code, none of the them is currently in-use.

Issue #2373, opened 2018-07-12
JDK11 builds fail with coredump


andrew-m-leonard:
Creating jdk image
Object 000000000086D900 not in heap range [00000000E0000000,0000000100000000)
08:44:36.125 0x847700    j9mm.141    *   ** ASSERTION FAILED ** at ./MarkingScheme.hpp:71: ((false))

keithc-ca:
I have a fix in the works for this: it involves ensuring that `JAVA_SPEC_VERSION` is always defined so the condition in j9cfg.h (`JAVA_SPEC_VERSION >= 11`) is evaluated consistently.

pshipton:
Fixed by #2374

DanHeidinga:
A big thank you to everyone involved in tracking this down - @keithc-ca @tajila @charliegracie and any others I may have missed!

PR #2374, opened 2018-07-12
Ensure JAVA_SPEC_VERSION is defined for OMR compilation

keithc-ca:
This is an interim solution to #2373.


pshipton:
jenkins test sanity plinux,win jdk8,jdk10

pshipton:
The travis ci build hasn't started yet. This change is to fix Java 11 and won't affect cmake, which builds Java 9 atm. There may be an equivalent change required to support cmake, but if so it doesn't need to block merging this.

keithc-ca:
I spoke with @dnakamura. He tells me that OMR is just treated as a subproject of OpenJ9 and should inherit all the macro definitions that belong on the command line: if Java 11 were supported with cmake, it would work without this change.

PR #2376, opened 2018-07-12
Fix incorrect mask

gacholio:
In the extremely unlikely case of failing to enter a the monitor for a
synchronized compiled frame, the mask used to change the resolve frame
type was incorrect.


gacholio:
No test covers this piece of code, which is common to all platforms, so the CI compile test is sufficient.

DanHeidinga:
Jenkins compile xlinux jdk8

DanHeidinga:
The Jenkins compile build passed - merging.

PR #2382, opened 2018-07-12
Enable profiling compilation on jitdumps

dchopra001:
Before queuing the method for recompilation the jit dump mechanism will now check to see if the crash was during a profiling compilation. If so, then it will set the appropriate property, so that the recompilation is also a profiling compilation.

Signed-off-by: Dhruv Chopra <Dhruv.C.Chopra@ibm.com>

### DIFF
--- a/runtime/compiler/control/HookedByTheJit.cpp
+++ b/runtime/compiler/control/HookedByTheJit.cpp
@@ -1887,6 +1887,7 @@ static TR_CompilationErrorCode recompileMethodForLog(
    TR::CompilationInfo *compInfo,
    TR_J9VMBase        *frontendOfThread,
    TR_Hotness          optimizationLevel,
+   bool                profilingCompile,
    void               *oldStartPC,
    TR::FILE *logFile
    )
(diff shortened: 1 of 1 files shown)


dchopra001:
@dsouzai

andrewcraik:
Note that it is possible that a normal compile became a profiling compile part way through due to switch to profiling. This change would mean that the compile, when re-run, would be profiling from the start potentially seeing different trees and transformations. Not saying this is a bad thing, but it is something to be aware of. Under JProfiling the switch to profiling will trigger a compilation abort and restart with profiling so this issue may go away once the default profiler is switched which should happen shortly.

dsouzai:
> Note that it is possible that a normal compile became a profiling compile part way through due to switch to profiling. 

I suppose it's possible for a jitDump compilation to also switch to profiling in the same way, in which case this change will not result in the exact recompile. At the same time however, any compile that had profiling from the beginning was never recompiled properly so this change fixes that problem. Given that JProfiling will solve the switch-to-profiling-midway scenario, this change is a good thing to have. That said, I think JProfiling needs to be jitDump aware so that it doesn't trigger a recompile during a jitDump compilation (though that's a different issue from this PR).

dsouzai:
@mpirvu could you do the Jenkins tests and merge if everything looks good to you.

mpirvu:
jenkins test all

PR #2268, opened 2018-06-26
Tentative class file ValueTypes attribute reading

anikser:
This change implements class file attribute reading for an upcoming
ValueTypes attribute.

Depends on #1772 

Signed-off-by: Mohammad Ali Nikseresht <anikser@ibm.com>

PR #2383, opened 2018-07-12
Add ValueTypes attribute writing to ROM Class

anikser:
This change adds a couple elements to J9ROMClass, where the ValueTypes
attribute data is then written.

Depends on #2268.


tajila:

PR #2384, opened 2018-07-12
Accerlate j/lang/Long.rotateLeft with OMR's lrol

0dvictor:
Transforming intrinsic call java/lang/Long.rotateLeft() to OMR's TR::lrol opcode.


andrewcraik:
Jenkins test sanity

andrewcraik:
@fjeremic I'm going to be away for a bit could you handle this one?

andrewcraik:
@0dvictor can you confirm that on 32bit this will do the right thing? ie with a register pair etc?

fjeremic:
> @0dvictor can you confirm that on 32bit this will do the right thing? ie with a register pair etc?

Will wait for answer to this before merging. I would assume the change is fine w.r.t. 32-bit and if it's not then we're missing something in OMR and we need to add more tril testing there for rotates.

0dvictor:
OpenJ9 functional tests have passed on 32-bit, and OMR's `OMR::X86::I386::TreeEvaluator::integerPairRolEvaluator()` has been manually confirmed that it should generate the desired code.

0dvictor:
Windows failures due to #2129

PR #2387, opened 2018-07-13
WIP : Add implementation of JDK9 Math.fma methods in OpenJ9

shivammittal99:
Signed-off-by: Shivam Mittal <shivammittal99@gmail.com>


runtime/compiler/x/codegen/J9TreeEvaluator.cpp
```diff
@@ -10806,6 +10806,172 @@ inlineMathSQRT(
...
+static bool inlineMathFMAFloat(TR::Node *node, TR::CodeGenerator *cg)
+   {
+   // fma(a, b, c) => (a*b+c)
+   TR::Node *firstChild = node->getChild(0);
+   TR::Node *secondChild = node->getChild(1);
+   TR::Node *thirdChild = node->getChild(2);
+   
+   TR::Register *targetRegister = NULL;
+
+   if(firstChild->getOpCode().isLoadConst() &&
```
andrewcraik: This case should not occur - the simplifier in the compiler should have folded the result prior to this point.

runtime/compiler/x/codegen/J9TreeEvaluator.cpp
```diff
@@ -10806,6 +10806,172 @@ inlineMathSQRT(
...
+static bool inlineMathFMADouble(TR::Node *node, TR::CodeGenerator *cg)
+   {
+   // fma(a, b, c) => (a*b+c)
+   TR::Node *firstChild = node->getChild(0);
+   TR::Node *secondChild = node->getChild(1);
+   TR::Node *thirdChild = node->getChild(2);
+   
+   TR::Register *targetRegister = NULL;
+
+   if(firstChild->getOpCode().isLoadConst() &&
```
andrewcraik: Again this shouldn't happen due to simplification.

runtime/compiler/x/codegen/J9TreeEvaluator.cpp
```diff
@@ -10806,6 +10806,172 @@ inlineMathSQRT(
...
+      }
+   else 
+      {
+      TR::Register *src1Register = cg->evaluate(firstChild);
+      TR::Register *src2Register = cg->evaluate(secondChild);
+      TR::Register *src3Register = cg->evaluate(secondChild);
+
+      targetRegister = src1Register;
+
+      generateRegRegRegInstruction(VFMADD132SSRegRegReg, node, targetRegister, src3Register, src2Register, cg);
```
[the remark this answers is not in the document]
0xdaryl: You should also try and produce the best possible instruction sequence given the context in which this operation occurs.  For instance, in the case of the `vfmadd` instruction there are a number of variants to this instruction that you should apply depending on whether one of its children's registers dies at this instruction or if it is the first and only use of the child.

Looking at what has been added to the instruction list it appears you should add the following to `X86Ops.ins`:
```
VFMADD132RegRegReg (already done)
VFMADD132RegRegMem
VFMADD213RegRegReg
VFMADD213RegRegMem
VFMADD231RegRegReg
VFMADD231RegRegMem
```
Of the 3 children to this `Math.fma()` call, if the reference count on any of those child nodes is 1 prior to their evaluation (i.e., `child->getReferenceCount() == 1 && child->getRegister() != NULL`) then that means that it is the last reference to this child node and any register it is evaluated into dies after this instruction.  If there is such a child then you want that one to be in the `xmm1` position for these instructions because it will always be clobbered by the instruction.  Ultimately this will save some register shuffling.

Furthermore, if a child's reference count is 1 and it has never been evaluated into a register before (i.e., `child->getReferenceCount() == 1 && child->getRegister() == NULL`) then that means this is the first and only evaluation of that child.  In such a case, you want to have this child in the `xmm3` position so that you can use the memory form of the instruction.  Floating point multiplication is commutative, so you should be able to cover all the permutations.  This will save a register because doing it otherwise the child would need to be loaded from memory into a register (potentially evicting something) when its not really required.

In any case, choose the variant of the instruction that allows you to achieve that.  Remember, the order of the operands on the instruction does not have to match the order of the children on the call.

runtime/compiler/x/codegen/J9TreeEvaluator.cpp
shivammittal99: Shifted constant folding to J9Simplifier.cpp

0xdaryl:
Jenkins build jdk9 depends eclipse/omr#2788

genie-openj9:

PR #2398, opened 2018-07-16
Decimal Format Peephole incorrectly detecting opportunity

klangman:
Decimal Format Peephole was not correctly verifying that the
child of the format() call is the direct result of the
BigDecimal.{double/float}Value() call allowing for some
operations to be removed.


fjeremic:
Jenkins test sanity

fjeremic:

PR #2405, opened 2018-07-17
String startsWith fast path for single char comparison

yanluo7:

### DIFF
--- a/jcl/src/java.base/share/classes/java/lang/String.java
+++ b/jcl/src/java.base/share/classes/java/lang/String.java
@@ -2522,6 +2522,12 @@ public final class String implements Serializable, Comparable<String>, CharSeque
 	 *				when prefix is null
 	 */
 	public boolean startsWith(String prefix, int start) {
+		if (prefix.length() == 1) {
+			if (start < 0 || start >= this.length()) {
+				return false;
+			}
+			return charAtInternal(start) == prefix.charAtInternal(0);
+		}
 		return regionMatches(start, prefix, 0, prefix.lengthInternal());
 	}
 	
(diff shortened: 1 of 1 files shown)


yanluo7:
Performance implication:
We observed in several benchmarks that it is often the case `String.startsWith` is called with a single character string. In these cases, going into the `regionMatches` call is much slower for single char comparison. A fast path is implemented to simply perform the character comparison. To preserve correctness, if the array bound check fails, `String.startsWith` simply returns false.

jcl/src/java.base/share/classes/java/lang/String.java
```diff
@@ -2522,6 +2522,11 @@ public boolean startsWith(String prefix) {
 	 *				when prefix is null
 	 */
 	public boolean startsWith(String prefix, int start) {
+                if (prefix.length()==1) {
+                   if (start < 0 || start>=this.length())
+                      return false;
+                   return charAtInternal(start) == prefix.charAtInternal(0);
+                }
```
fjeremic: Please use tabs instead of spaces in JCL files. Also the String.java file is split into a Java 8 and Java 9+ versions. Can we make the change on both sides?

fjeremic:
@keithc-ca could I suggest you for an additional review in this JCL change?

jcl/src/java.base/share/classes/java/lang/String.java
keithc-ca: Also please use braces ({}) for the body of the inner `if` statement and consistently use whitespace around all binary operators.

yanluo7:
@fjeremic All comments are addressed.

fjeremic:
> @fjeremic All comments are addressed.

Please see https://github.com/eclipse/openj9/pull/2405#discussion_r203090264.

jcl/src/java.base/share/classes/java/lang/String.java
```diff
@@ -2522,6 +2522,12 @@ public boolean startsWith(String prefix) {
 	 *				when prefix is null
 	 */
 	public boolean startsWith(String prefix, int start) {
+		if (prefix.length()==1) {
```
keithc-ca: Please use spaces around binary operators (e.g. `==`) in both versions of this method.

fjeremic:
Jenkins test sanity

fjeremic:

PR #2408, opened 2018-07-17
Inlining support of MultiANewArray for 2 dimensional arrays

yanluo7:
- Support inlining of 2 dimensional reference arrays where the second dimension is 0. The inlining sequence allocates N+1 array objects (N being the length of first dimension): 1 "large" array of length N and N "small" arrays of length 0; and the elements of the "large" array of length N point to their respective "small" arrays to form the two dimensions.

### DIFF
--- a/runtime/compiler/x/codegen/J9TreeEvaluator.cpp
+++ b/runtime/compiler/x/codegen/J9TreeEvaluator.cpp
@@ -2459,7 +2459,267 @@ TR::Register *J9::X86::TreeEvaluator::newEvaluator(TR::Node *node, TR::CodeGener
 
 TR::Register *J9::X86::TreeEvaluator::multianewArrayEvaluator(TR::Node *node, TR::CodeGenerator *cg)
    {
-   return TR::TreeEvaluator::performHelperCall(node, NULL, TR::acall, true, cg);
+   static char *useDirectHelperCall = feGetEnv("TR_MultiANewArrayEvaluatorUseDirectCall");
+
+   TR::Node *firstChild = node->getFirstChild();
+   TR::Node *secondChild = node->getSecondChild();
+   TR::Node *thirdChild = node->getThirdChild();
+
+   if (useDirectHelperCall || !secondChild->getOpCode().isLoadConst() || secondChild->getInt()!=2)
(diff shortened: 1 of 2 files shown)


yanluo7:
Performance implication:
We observed in benchmarks a large number of multi-dimension array allocations. Further investigation shows much of these are 2-dimensional, and the second dimension is 0. Inlining of these types of arrays would be beneficial as no helper calls would be needed. We observed day trade throughout to improved by ~0.5%.

yanluo7:
For testing, we ran day trader and other server benchmarks, as well as JLTF testing that exercised the code path (by instrumenting the method signatures for during compilation).
Create issue https://github.com/eclipse/openj9/issues/2424 to implement on P/Z platforms.

yanluo7:
@0dvictor Could you do a review?

yanluo7:
Comments are addressed, could this be merged soon?

runtime/compiler/x/codegen/J9TreeEvaluator.cpp
```diff
@@ -2459,7 +2459,269 @@ TR::Register *J9::X86::TreeEvaluator::newEvaluator(TR::Node *node, TR::CodeGener
...
+   dimsPtrReg = cg->evaluate(firstChild);
+
+   classReg = cg->evaluate(thirdChild);
+
+   generateRegMemInstruction(L4RegMem, node, secondDimLenReg,
+                             generateX86MemoryReference(dimsPtrReg, 0, cg), cg);
+   generateRegMemInstruction(L4RegMem, node, firstDimLenReg,
+                             generateX86MemoryReference(dimsPtrReg, 4, cg), cg);
+
+   //generateRegImm64Instruction(MOV8RegImm64, node, targetReg, 0x1234, cg);
```
0dvictor: Was this for debug purpose? IMO it is quite confusing seeing code being commented out or surrounded with `#if 0`. It might be better just to remove it.

yanluo7: Comment removed.

0dvictor:
LGTM

andrewcraik:

andrewcraik:
LGTM

Issue #2410, opened 2018-07-17
Fix recognition of hashCode function on Power


IBMJimmyk:
There are two issues related to recognizing the function hashCodeImplDecompressed for String on Power.

The first issue is sometimes hashCodeImplDecompressed would get inlined. This prevents the function for being recognized and replaced with hand optimized assembly. It should be changed so that the method is not inlined if the criteria to replace it with hand optimized assembly later on has been met.

The second issue is the hand optimized assembly for hashCodeImplDecompressed does not work with arraylets. Currently the JIT doesn't check if arraylets can be generated when it should. It is possible for incorrect results to be generated if the hand written assembly is used with arraylets. The JIT should be changed to check for arraylets and opt not to use the hand written assembly if arraylets are possible.

PR #2411, opened 2018-07-17
Fix recognition of hashCode function on Power

IBMJimmyk:
This change set fixes two issues related to recognizing the function
hashCodeImplDecompressed for String on Power.

The first issue is sometimes hashCodeImplDecompressed would get inlined.
This prevents the function for being recognized and replaced with hand
optimized assembly. This is being changed so that the function does not
get inlined if the JIT thinks it will be able to replace it with the
assembly later on.

The second issue is the hand optimized assembly for
hashCodeImplDecompressed does not work with arraylets. This was not being
checked for before and could lead to an incorrect result. This is being
changed to check if the generation of arraylets is possible. If so, the
JIT will not try to use the hand optimized assembly.

Issue: #2410

### DIFF
--- a/runtime/compiler/env/ProcessorDetection.cpp
+++ b/runtime/compiler/env/ProcessorDetection.cpp
@@ -562,6 +562,16 @@ TR_J9VMBase::getPPCProcessorType()
    return portLibCall_getProcessorType();
    }
 
+bool
+TR_J9VMBase::getPPCSupportsVSXRegisters()
+   {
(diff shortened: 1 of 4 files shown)


IBMJimmyk:
I see the error. I'm currently working on fixing it.

IBMJimmyk:
@gita-omr Could you take a look at this pull request?

gita-omr:
LGTM

PR #2416, opened 2018-07-17
Disallow using PUSH to preserve registers on X86

0dvictor:
According to Agner Fog's instruction tables, one PUSH instruction performs worse
than the SUB+MOV instruction combinations. In addition, using PUSH instruction
requires stack adjustments before and after preserving registers as they are neither
at the beginning nor at the end of the stack; and hence two extra SUB instructions
cannot be avoided. In fact, X86-64 never uses PUSH instruction for register
preservations, and X86-32 only uses it under limited situations.
This commit removes the logics related to allow using PUSH instruction to preserve
registers, so that follow-up changes can elimite one or two instructions in the prolog.


cathyzhyi:
LGTM

fjeremic:
Jenkins test sanity xlinux

PR #2417, opened 2018-07-17
Replace DataSnippet type with auto on X86

0dvictor:
In case if OMR changes the type of DateSnippet, it is safer to auto instead of the raw type.


cathyzhyi:
LGTM

fjeremic:
Jenkins test sanity xlinux

PR #2420, opened 2018-07-18
load java.instrument module when loading agent

### LINKED ISSUE #2030: Cannot load agent jar in Java 10 modular app
Trying to load an agent jar into OpenJ9 Java 10 using the attach API, where Java 10 is running a module, results in a AgentLoadException error 102. It works with OpenJDK.
```
*** java.lang.instrument ASSERTION FAILED ***: "!errorOutstanding" with message find class on InstrumentationImpl failed at JPLISAgent.c line: 493
*** java.lang.instrument ASSERTION FAILED ***: "success" at InvocationAdapter.c line: 395
Agent failed to start!
JVMJ9TI064E Agent initialization function Agent_OnAttach failed for library instrument, return code 102
```

openjdk version "10.0.1-adoptopenjdk" 2018-04-17
OpenJDK Runtime Environment (build 10.0.1-adoptopenjdk+0-adhoc..openjdk)
Eclipse OpenJ9 VM (build master-6f5a858a, JRE 10 Windows 8.1 amd64-64-Bit Compressed References 20180523_54 (JIT enabled, AOT enabled)
OpenJ9   - 6f5a858a
OMR      - 15290506
JCL      - 228f0625d8 based on jdk-10.0.1+10)

You need to start a Java 10 modular app to connect to, such as jconsole. While running jconsole exhibits the problem you don't get a console. To get a console run "java -m jdk.jconsole/sun.tools.jconsole.JConsole" instead.

Sample loader code.
```
		String vmid = com.ibm.tools.attach.target.AttachHandler.getVmId();
		VirtualMachineDescriptor target = null;
		List<VirtualMachineDescriptor> machines = VirtualMachine.list();
		for (VirtualMachineDescriptor vmd : machines) {
			if (vmd.id().equals(vmid)) continue;
			System.out.println(vmd);
			if (vmd.displayName().contains(args[0])) {
				if (target != null) {
					System.out.println("Duplicates!");
					System.exit(1);
				}
				target = vmd;
			}
		}
		if (target == null) {
			System.out.println("No target found");
			System.exit(1);
		}
		VirtualMachine vm = VirtualMachine.attach(target);
		vm.loadAgent("t:\\peter\\premain.jar");
```

pdbain-ibm:
Java 9 and above need the java.instrument module loaded when
using -javaagent or loading an instrumentation agent via late attach.
Avoid this on legacy versions of Java 9.

Fixes #2030 

### DIFF
--- a/runtime/vm/classloadersearch.c
+++ b/runtime/vm/classloadersearch.c
@@ -34,6 +34,8 @@
 
 static UDATA addToSystemProperty(J9JavaVM * vm, const char * propertyName, const char * segment);
 static UDATA addZipToLoader(J9JavaVM * vm, const char * filename, J9ClassLoader * classLoader, BOOLEAN enforceJarRestriction);
+#define INSTRUMENT_MODULE "java.instrument"
+static UDATA loadModule(J9JavaVM *vm, const char *moduleName, jobject *modulePtr);
 
 
 /**
@@ -226,6 +228,8 @@ addZipToLoader(J9JavaVM * vm, const char * filename, J9ClassLoader * classLoader
 			(*env)->CallVoidMethod(env, classLoaderRef, mid, filenameString);
 			if ((*env)->ExceptionCheck(env)) {
 				rc = CLS_ERROR_OUT_OF_MEMORY;
+			} else if (J2SE_SHAPE(vm) >= J2SE_SHAPE_B165) {
+				rc = loadModule(vm, INSTRUMENT_MODULE, NULL);
 			}
 	cleanup:
 			(*env)->ExceptionClear(env);
@@ -235,5 +239,83 @@ addZipToLoader(J9JavaVM * vm, const char * filename, J9ClassLoader * classLoader
 		}
 	}
 
+	return rc;
+}
+
+jclass
+getJimModules(J9VMThread *currentThread)
+{
+	J9JavaVM *vm = NULL;
+	vm = currentThread->javaVM;
+	if (NULL == vm->jimModules) {
+		const J9InternalVMFunctions *vmFuncs = NULL;
+		/* no need for mutual exclusion since repeating this is harmless */
+		J9Class *j9JimModules = NULL;
+		internalEnterVMFromJNI(currentThread);
+		vmFuncs = vm->internalVMFunctions;
+		j9JimModules = vmFuncs->internalFindKnownClass(currentThread,
+				J9VMCONSTANTPOOL_JDKINTERNALMODULEMODULES,
+				J9_FINDKNOWNCLASS_FLAG_INITIALIZE | J9_FINDKNOWNCLASS_FLAG_NON_FATAL);
+		if (NULL != j9JimModules) {
+			/*
+			 * The address of j9class->classObject is considered to be a JNI global reference.
+			 * Note that the class in question is not a candidate for unloading.
+			 */
+			vm->jimModules = (jclass)&j9JimModules->classObject;
+		} else {
+			currentThread->currentException = NULL;
+			currentThread->privateFlags &= ~J9_PRIVATE_FLAGS_REPORT_EXCEPTION_THROW;
+		}
+		internalExitVMToJNI(currentThread);
+	}
+	return vm->jimModules;
+}
+
+/**
+ * Loads the named module using jdk.internal.module.Modules.loadModule()
+ *
+ * @param [in] vm pointer to JVM
+ * @param [in] moduleName name of module to load
+ * @param [out] modulePtr pointer to return module object.  May be null.
+ *
+ * @return CLS_ERROR_NONE on success, other value on failure
+ */
+static UDATA
+loadModule(J9JavaVM *vm, const char *moduleName, jobject *modulePtr)
+{
+	UDATA rc = CLS_ERROR_NONE;
+	J9VMThread *currentThread = currentVMThread(vm);
+	JNIEnv *env = (JNIEnv *) currentThread;
+	jstring moduleNameString = (*env)->NewStringUTF(env, moduleName);
+	jclass jimModules = getJimModules(currentThread);
+	jobject vmModule = NULL;
+	jmethodID loadModule = NULL;
+
+	if (NULL == moduleNameString) {
+		rc = CLS_ERROR_OUT_OF_MEMORY;
+		goto done;
+	}
+	if (NULL == jimModules) {
+		rc = CLS_ERROR_NOT_FOUND;
+		goto done;
+	}
+
+	if (NULL != modulePtr) {
+		*modulePtr = NULL;
+	}
+
+	loadModule = (*env)->GetStaticMethodID(env, jimModules, "loadModule", "(Ljava/lang/String;)Ljava/lang/Module;");
+	if (NULL == loadModule) {
+		rc = CLS_ERROR_NOT_FOUND;
+		goto done;
+	}
+
+	vmModule = (*env)->CallStaticObjectMethod(env, jimModules, loadModule, moduleNameString);
+	if ((*env)->ExceptionOccurred(env)) {
+		rc = CLS_ERROR_INTERNAL;
+	} else if (NULL != modulePtr) {
+		*modulePtr = vmModule;
+	}
+done:
 	return rc;
 }
--- a/runtime/jvmti/jvmtiModules.c
+++ b/runtime/jvmti/jvmtiModules.c
@@ -640,36 +640,18 @@ jvmtiAddModuleProvides(jvmtiEnv* jvmtiEnv,
 
 		if ((rc == JVMTI_ERROR_NONE) && (FALSE == isUnnamed)) {
 			JNIEnv *env = (JNIEnv *) currentThread;
-			if (NULL == vm->jimModules) {
-				omrthread_monitor_enter(vm->jlmModulesInitMutex);
-				if (NULL == vm->jimModules) {
-					jclass globalModules = NULL;
-					jclass jimModules = (*env)->FindClass(env, "jdk/internal/module/Modules");
-					if (NULL == jimModules) {
-						rc = JVMTI_ERROR_INTERNAL;
-						omrthread_monitor_exit(vm->jlmModulesInitMutex);
-						goto done;
-					}
-
-					globalModules = (*env)->NewGlobalRef(env, jimModules);
-					if (NULL == globalModules) {
-						rc = JVMTI_ERROR_OUT_OF_MEMORY;
-						omrthread_monitor_exit(vm->jlmModulesInitMutex);
-						goto done;
-					}
-
-					vm->jimModules = globalModules;
-				}
-				omrthread_monitor_exit(vm->jlmModulesInitMutex);
+			jclass jimModules = vmFuncs->getJimModules(currentThread);
+			if (NULL == jimModules) {
+				rc = JVMTI_ERROR_INTERNAL;
+				goto done;
 			}
-
 			if (NULL == vm->addProvides) {
 				jmethodID addProvides = NULL;
 
 				if(J2SE_SHAPE(vm) < J2SE_SHAPE_B165) {
-					addProvides = (*env)->GetStaticMethodID(env, vm->jimModules, "addProvides", "(Ljava/lang/reflect/Module;Ljava/lang/Class;Ljava/lang/Class;)V");
+					addProvides = (*env)->GetStaticMethodID(env, jimModules, "addProvides", "(Ljava/lang/reflect/Module;Ljava/lang/Class;Ljava/lang/Class;)V");
 				} else {
-					addProvides = (*env)->GetStaticMethodID(env, vm->jimModules, "addProvides", "(Ljava/lang/Module;Ljava/lang/Class;Ljava/lang/Class;)V");
+					addProvides = (*env)->GetStaticMethodID(env, jimModules, "addProvides", "(Ljava/lang/Module;Ljava/lang/Class;Ljava/lang/Class;)V");
 				}
 				if (NULL == addProvides) {
 					rc = JVMTI_ERROR_INTERNAL;
@@ -678,7 +660,7 @@ jvmtiAddModuleProvides(jvmtiEnv* jvmtiEnv,
 
 				vm->addProvides = addProvides;
 			}
-			(*env)->CallStaticVoidMethod(env, vm->jimModules, vm->addProvides, module, service, implClass);
+			(*env)->CallStaticVoidMethod(env, jimModules, vm->addProvides, module, service, implClass);
 		}
 	}
 
--- a/runtime/oti/classloadersearch.h
+++ b/runtime/oti/classloadersearch.h
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 1991, 2014 IBM Corp. and others
+ * Copyright (c) 1991, 2018 IBM Corp. and others
  *
  * This program and the accompanying materials are made available under
  * the terms of the Eclipse Public License 2.0 which accompanies this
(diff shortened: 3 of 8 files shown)


smlambert:
Jenkins test sanity all

runtime/vm/classloadersearch.c
```diff
@@ -34,6 +34,8 @@
 
 static UDATA addToSystemProperty(J9JavaVM * vm, const char * propertyName, const char * segment);
 static UDATA addZipToLoader(J9JavaVM * vm, const char * filename, J9ClassLoader * classLoader, BOOLEAN enforceJarRestriction);
+#define INSTRUMENT_MODULE "java.instrument"
+UDATA JNICALL loadModule(J9JavaVM *vm, const char *moduleName, jobject *modulePtr);
```
keithc-ca: This method should be static in which case JNICALL should be omitted.

runtime/vm/classloadersearch.c
```diff
@@ -237,3 +242,52 @@ addZipToLoader(J9JavaVM * vm, const char * filename, J9ClassLoader * classLoader
...
+	 * vm->jimModules may have been set already by JVMTI.
+	 * If not, grab a private copy of the class.
+	 */
+	if (NULL == jimModules) {
+		jimModules = (*env)->FindClass(env, "jdk/internal/module/Modules");
+	}
+	if (NULL == jimModules) {
+		rc = CLS_ERROR_NOT_FOUND;
+	} else {
+		jmethodID loadModule =
```
keithc-ca: Shouldn't this save `jimModules` in `vm->jimModules`?

runtime/vm/classloadersearch.c
```diff
@@ -237,3 +242,52 @@ addZipToLoader(J9JavaVM * vm, const char * filename, J9ClassLoader * classLoader
...
+				(*env)->GetStaticMethodID(env, jimModules, "loadModule", "(Ljava/lang/String;)Ljava/lang/Module;");
+		if (NULL == loadModule) {
+			rc = CLS_ERROR_NOT_FOUND;
+		} else {
+			jstring moduleNameString =  (*env)->NewStringUTF(env, moduleName);
```
keithc-ca: A check that this succeeded should be added. Also please fix the formatting (extra space after `=`).

runtime/vm/classloadersearch.c
```diff
@@ -237,3 +242,52 @@ addZipToLoader(J9JavaVM * vm, const char * filename, J9ClassLoader * classLoader
...
+
+
+/**
+ * Loads the named module using jdk.internal.module.Modules.loadModule()
+ *
+ * @param [in] vm pointer to JVM
+ * @param [in] moduleName name of module to load
+ * @param [out] modulePtr pointer to return module object.  May be null.
+ *
+ * @return CLS_ERROR_NONEE on success, other value on failure
```
keithc-ca: Please correct the success code.

runtime/vm/classloadersearch.c
```diff
@@ -227,6 +229,9 @@ addZipToLoader(J9JavaVM * vm, const char * filename, J9ClassLoader * classLoader
 			if ((*env)->ExceptionCheck(env)) {
 				rc = CLS_ERROR_OUT_OF_MEMORY;
 			}
+			if 	((J2SE_VERSION(vm) >= J2SE_19) && (J2SE_SHAPE(vm) >= J2SE_SHAPE_B165)) {
```
gacholio: Some wild spacing here.

gacholio: This code should not run if the above code errored.  Need an else or goto cleanup.

runtime/vm/classloadersearch.c
pdbain-ibm: Thanks.  Fixed.

runtime/vm/classloadersearch.c
gacholio: Not clear what you mean here.  If you're going to cache it, you'll need to move the JVMTI code out into a helper in util (definitely don't duplicate the mutexxed init code),

gacholio: Is there a good reason this isn't just in the VM constant pool?  Could then use findKnownClass to get it.

pdbain-ibm:
I moved the code to load jdk/internal/module/Modules to a utility file per GAC's suggestion
and addressed (I hope) the other comments.  I compiled and did an ad-hoc test.  Could someone run another pull request build?

runtime/vm/classloadersearch.c
pdbain-ibm: That I don't know.  I stole the code from jvmtiModules.c.  Maybe @tajila knows.

smlambert:
Jenkins test sanity all

runtime/util/moduleclasses.c
```diff
@@ -0,0 +1,46 @@
...
+		omrthread_monitor_enter(vm->jlmModulesInitMutex);
+		if (NULL == vm->jimModules) {
+			jclass jimModules = (*env)->FindClass(env,
+					"jdk/internal/module/Modules");
+			if (NULL == jimModules) {
+				result = -1;
+			} else {
+				jclass globalModules = (*env)->NewGlobalRef(env, jimModules);
+				if (NULL == globalModules) {
+					omrthread_monitor_exit(vm->jlmModulesInitMutex);
```
keithc-ca: This yields to two calls `omrthread_monitor_exit`.

runtime/util/moduleclasses.c
```diff
@@ -0,0 +1,46 @@
...
+ *
+ * [1] https://www.gnu.org/software/classpath/license.html
+ * [2] http://openjdk.java.net/legal/assembly-exception.html
+ *
+ * SPDX-License-Identifier: EPL-2.0 OR Apache-2.0 OR GPL-2.0 WITH Classpath-exception-2.0 OR LicenseRef-GPL-2.0 WITH Assembly-exception
+ *******************************************************************************/
+#include "j9.h"
+
+IDATA
+ensureJimModules(J9JavaVM* vm, JNIEnv* env) {
```
gacholio: Formatting - also, don't pass both the thread and the VM.  I also despise the use of IDATA as a return value.

runtime/util/moduleclasses.c
gacholio: Right, this should just be and else case.

runtime/util/moduleclasses.c
gacholio: I suppose IDATA follows existing (if bad) conventions.

runtime/util/moduleclasses.c
pdbain-ibm: Oops.  Missed that.

runtime/vm/classloadersearch.c
gacholio: @tajila Actually, VMCP classes are only loaded eagerly if they have fields/methods declared in the constant pool.  If it's just the class, they are loaded on demand.

gacholio: Discussed with Tobi & Peter, we will go with the CP solution to simplify.

pdbain-ibm:
@jdekonin or @AdamBrousseau sanity builds are failing: `ERROR: Cannot find label value matching JOB_TYPE:'build' SPEC:'aix_ppc-64_cmprssptrs' and SDK_VERSION:'11'
Finished: FAILURE`  Is this a known issue?

runtime/util/moduleclasses.c
pdbain-ibm: ... or I could just delete the whole file and put the class in the VM constant pool, as discussed.

pdbain-ibm:
Compiled this locally, but not tested.  @gacholio could you do a preliminary review?

runtime/vm/classloadersearch.c
```diff
@@ -237,3 +241,71 @@ addZipToLoader(J9JavaVM * vm, const char * filename, J9ClassLoader * classLoader
 
 	return rc;
 }
+
+jclass
+getJimModules(J9VMThread *currentThread) {
```
gacholio: Brace on it's own line,

runtime/vm/classloadersearch.c
```diff
@@ -237,3 +241,71 @@ addZipToLoader(J9JavaVM * vm, const char * filename, J9ClassLoader * classLoader
...
+		goto done;
+	}
+
+	vmModule = (*env)->CallStaticObjectMethod(env, jimModules, loadModule, moduleNameString);
+	if ((*env)->ExceptionOccurred(env)) {
+		rc = CLS_ERROR_INTERNAL;
+	} else if (NULL != modulePtr) {
+		*modulePtr = vmModule;
+	}
+	done:
```
gacholio: No indent.

gacholio: Could do a single assignment to *modulePtr here instead of the two you have.

gacholio:
Other than a few nitpicks, this looks good to me.

runtime/vm/classloadersearch.c
```diff
@@ -237,3 +241,71 @@ addZipToLoader(J9JavaVM * vm, const char * filename, J9ClassLoader * classLoader
 
 	return rc;
 }
+
+jclass
+getJimModules(J9VMThread *currentThread) {
+	J9JavaVM *vm = currentThread->javaVM;
+	if (NULL == vm->jimModules) {
+		const J9InternalVMFunctions *vmFuncs = vm->internalVMFunctions;
+		/* no need for mutual exclusion since repeating this is harmless */
```
gacholio: Need to enterVMFromJNI here.

runtime/vm/classloadersearch.c
```diff
@@ -237,3 +241,71 @@ addZipToLoader(J9JavaVM * vm, const char * filename, J9ClassLoader * classLoader
...
+		J9Class *j9JimModules = vmFuncs->internalFindKnownClass(currentThread,
+				J9VMCONSTANTPOOL_JDKINTERNALMODULEMODULES,
+				J9_FINDKNOWNCLASS_FLAG_INITIALIZE | J9_FINDKNOWNCLASS_FLAG_NON_FATAL);
+		if (NULL != j9JimModules) {
+			vm->jimModules = (jclass)&j9JimModules->classObject;
+		} else {
+			currentThread->currentException = NULL;
+			currentThread->privateFlags &= ~J9_PRIVATE_FLAGS_REPORT_EXCEPTION_THROW;
+		}
```
gacholio: ...and exitVMToJNI here.

runtime/vm/classloadersearch.c
gacholio: So you won't be able to declare and initialize j9JimModules on the same line.

gacholio:
On further reflection, I'm not sure the constant pool was the right choice, as the module stuff is all being done in a JNI context and used in JNI calls.

Using FindClass is probably a better solution. The mutex can be avoided by doing:

enterVMFromJNI
J9Class *clazz = J9VM_J9CLASS_FROM_JCLASS(vmThread, jclassForModule);
jimModule = &clazz->classObject;
exitVMToJNI

runtime/vm/classloadersearch.c
```diff
@@ -237,3 +241,71 @@ addZipToLoader(J9JavaVM * vm, const char * filename, J9ClassLoader * classLoader
...
+getJimModules(J9VMThread *currentThread) {
+	J9JavaVM *vm = currentThread->javaVM;
+	if (NULL == vm->jimModules) {
+		const J9InternalVMFunctions *vmFuncs = vm->internalVMFunctions;
+		/* no need for mutual exclusion since repeating this is harmless */
+		J9Class *j9JimModules = vmFuncs->internalFindKnownClass(currentThread,
+				J9VMCONSTANTPOOL_JDKINTERNALMODULEMODULES,
+				J9_FINDKNOWNCLASS_FLAG_INITIALIZE | J9_FINDKNOWNCLASS_FLAG_NON_FATAL);
+		if (NULL != j9JimModules) {
+			vm->jimModules = (jclass)&j9JimModules->classObject;
```
gacholio: We should add a macro for this later, even though it will look odd to J9CLASS_FROM_JCLASS followed by JCLASS_FROM_J9CLASS.

runtime/vm/classloadersearch.c
pdbain-ibm: Sorry, please clarify.

runtime/vm/classloadersearch.c
pdbain-ibm: > Brace on it's own line,
No apostrophe in the possessive of "it".

runtime/vm/classloadersearch.c
gacholio: I'm saying you can do a single NULL check and assignment at done.

runtime/vm/classloadersearch.c
gacholio: Also probably worth a comment here that the address of j9class->classObject is considered to be a JNI global ref.  It won't keep a class alive, but that's OK here since the class in question is not a candidate for unloading.

pdbain-ibm: https://github.com/eclipse/openj9/issues/2437

gacholio:
jenkins test sanity xlinux jdk9

gacholio:
jenkins test sanity xlinux jdk9

gacholio:
Intentionally cancelled travis build since it will take too long to even start,

pdbain-ibm:
@keithc-ca would you kindly review and approve?
Thanks

runtime/vm/classloadersearch.c
pdbain-ibm: fixed.

runtime/vm/classloadersearch.c
```diff
@@ -237,3 +241,81 @@ addZipToLoader(J9JavaVM * vm, const char * filename, J9ClassLoader * classLoader
 
 	return rc;
 }
+
+jclass
+getJimModules(J9VMThread *currentThread)
+{
+	J9JavaVM *vm = NULL;
+	vm = currentThread->javaVM;
```
keithc-ca: nit: this initialization could be collapsed into the declaration.

runtime/vm/classloadersearch.c
```diff
@@ -237,3 +241,81 @@ addZipToLoader(J9JavaVM * vm, const char * filename, J9ClassLoader * classLoader
...
+ * @param [in] moduleName name of module to load
+ * @param [out] modulePtr pointer to return module object.  May be null.
+ *
+ * @return CLS_ERROR_NONE on success, other value on failure
+ */
+static UDATA
+loadModule(J9JavaVM *vm, const char *moduleName, jobject *modulePtr)
+{
+	UDATA rc = CLS_ERROR_NONE;
+	J9VMThread *currentThread = currentVMThread(vm);
```
keithc-ca: Where this function is used, we've already retrieved the current thread: it would make sense to avoid redoing that by passing `currentThread` or `env` instead of `vm`.

runtime/vm/classloadersearch.c
```diff
@@ -237,3 +241,81 @@ addZipToLoader(J9JavaVM * vm, const char * filename, J9ClassLoader * classLoader
...
+	loadModule = (*env)->GetStaticMethodID(env, jimModules, "loadModule", "(Ljava/lang/String;)Ljava/lang/Module;");
+	if (NULL == loadModule) {
+		rc = CLS_ERROR_NOT_FOUND;
+		goto done;
+	}
+
+	vmModule = (*env)->CallStaticObjectMethod(env, jimModules, loadModule, moduleNameString);
+	if ((*env)->ExceptionOccurred(env)) {
+		rc = CLS_ERROR_INTERNAL;
+	} else if (NULL != modulePtr) {
```
keithc-ca: I like GAC's suggestion to move this after `done` and remove the `*modulePtr = NULL;` above.

keithc-ca:
Jenkins test sanity win jdk8,jdk10

runtime/vm/classloadersearch.c
pdbain-ibm: Will fix in https://github.com/eclipse/openj9/issues/2437

runtime/vm/classloadersearch.c
pdbain-ibm: Agreed.  Will fix in https://github.com/eclipse/openj9/issues/2437

pdbain-ibm:
Per discussion with Keith and GAC, this change is needed ASAP.  Deferring improvements to #2437.

keithc-ca:
The windows failures are due to https://github.com/eclipse/openj9/issues/2440.

keithc-ca:
Enough builds have passed to satisfy me.

PR #2421, opened 2018-07-18
Implement JVM_AreNestMates public native

tajila:
Add implementation for JVM_AreNestMates

### DIFF
--- a/runtime/j9vm/java11vmi.c
+++ b/runtime/j9vm/java11vmi.c
@@ -1797,6 +1797,8 @@ JVM_BeforeHalt()
 {
 	/* To be implemented via https://github.com/eclipse/openj9/issues/1459 */
 }
+#endif /* J9VM_JCL_SE11 */
+#if defined(J9VM_OPT_VALHALLA_NESTMATES)
 JNIEXPORT jclass JNICALL
 JVM_GetNestHost(JNIEnv *env, jclass clz)
 {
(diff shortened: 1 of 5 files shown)


tajila:
@JasonFengJ9 please review

runtime/j9vm/java11vmi.c
```diff
@@ -1812,7 +1812,24 @@ JVM_GetNestMembers(JNIEnv *env, jclass clz)
 JNIEXPORT jboolean JNICALL
 JVM_AreNestMates(JNIEnv *env, jclass clzOne, jclass clzTwo)
 {
-	assert(!"JVM_AreNestMates unimplemented");
+	j9object_t clazzOne = NULL;
+	j9object_t clazzTwo = NULL;
```
JasonFengJ9: These two variable defines could be moved into next `if` block and close to their usages.

runtime/j9vm/java11vmi.c
```diff
@@ -1812,7 +1812,24 @@ JVM_GetNestMembers(JNIEnv *env, jclass clz)
 JNIEXPORT jboolean JNICALL
 JVM_AreNestMates(JNIEnv *env, jclass clzOne, jclass clzTwo)
```
JasonFengJ9: When the stub method was introduced, it wasn't immediately clear what these two `class` variables mean. Now it could be changed to `nestOne` & `nestTwo` or something else more meaningful.

runtime/j9vm/java11vmi.c
```diff
@@ -1812,7 +1812,24 @@ JVM_GetNestMembers(JNIEnv *env, jclass clz)
...
-	assert(!"JVM_AreNestMates unimplemented");
+	j9object_t clazzOne = NULL;
+	j9object_t clazzTwo = NULL;
+
+	if ((NULL != clzOne) && (NULL != clzTwo)) {
+		clazzOne = J9_JNI_UNWRAP_REFERENCE(clzOne);
+		clazzTwo = J9_JNI_UNWRAP_REFERENCE(clzTwo);
+
+		if (clazzOne == clazzTwo) {
+			return JNI_TRUE;
```
JasonFengJ9: Pls use a single return point.

runtime/j9vm/java11vmi.c
```diff
@@ -1812,7 +1812,24 @@ JVM_GetNestMembers(JNIEnv *env, jclass clz)
...
+		} else {
+			J9VMThread *currentThread = (J9VMThread*) env;
+
+			if (J9OBJECT_CLAZZ(currentThread, clazzOne)->nestHost == J9OBJECT_CLAZZ(currentThread, clazzTwo)->nestHost) {
+				return JNI_TRUE;
```
JasonFengJ9: Same here, pls use a single return point.

runtime/j9vm/java11vmi.c
```diff
@@ -1812,7 +1812,24 @@ JVM_GetNestMembers(JNIEnv *env, jclass clz)
...
 {
-	assert(!"JVM_AreNestMates unimplemented");
+	j9object_t clazzOne = NULL;
+	j9object_t clazzTwo = NULL;
+
+	if ((NULL != clzOne) && (NULL != clzTwo)) {
+		clazzOne = J9_JNI_UNWRAP_REFERENCE(clzOne);
+		clazzTwo = J9_JNI_UNWRAP_REFERENCE(clzTwo);
+
+		if (clazzOne == clazzTwo) {
```
JasonFengJ9: Can this `equivalent` check be moved before invoking `J9_JNI_UNWRAP_REFERENCE`?

tajila: JNI references will be pointers to the Java stack, so even if they are the same object it will not return true

runtime/j9vm/java11vmi.c
tajila: classOne and classTwo still makes the most sense as that's what we are comparing. We are trying to see if two classes belong to the same nest

tajila:
Just noticed an issue with this, need to fix it

tajila:
@JasonFengJ9 the PR has been updated

runtime/j9vm/java11vmi.c
```diff
@@ -1809,10 +1809,58 @@ JVM_GetNestMembers(JNIEnv *env, jclass clz)
...
+
+		if (clazzObjectOne == clazzObjectTwo) {
+			result = JNI_TRUE;
+		} else {
+			J9VMThread *currentThread = (J9VMThread*) env;
+			J9Class *clazzOne = J9VM_J9CLASS_FROM_HEAPCLASS(currentThread, clazzObjectOne);
+			J9Class *clazzTwo = J9VM_J9CLASS_FROM_HEAPCLASS(currentThread, clazzObjectTwo);
+
+			if (NULL == clazzOne->nestHost) {
+				if (J9_VISIBILITY_ALLOWED != vmFuncs->loadAndVerifyNestHost(currentThread, clazzOne, J9_LOOK_NO_THROW)) {
```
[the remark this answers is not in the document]
tajila: yes it does but ill change it to be more clear

tajila:
@gacholio can you please review

gacholio:
All the code is under ifdef, so there's no point running extra builds.

PR #2423, opened 2018-07-18
Prep to remove deferredVMAccessRelease

amicic:
Now that MasterGCThread does not use deferred VM access mechanism
(https://github.com/eclipse/omr/pull/2696), we can remove it. This is
just a simple preparation step so that OMR call site can use the
delegate without the argument before that argument is actually removed.


dmitripivkine:

Issue #2109, opened 2018-06-07
jdk11 add String.isBlank() and lines()


pshipton:
http://cr.openjdk.java.net/~iris/se/11/build/latest/diffsFrom10%2b45/java/lang/String.html

fjeremic:
We should check StringLatin1 and StringUTF16 for already existing helpers we can take advantage of when implementing this.

tajila:
@Ali-Ni can you take a look at this

PR #2426, opened 2018-07-18
 Add String.isBlank() implementation (jdk11 only)

Ali-Ni:
This change implements String.isBlank(), a function that determines
whether or not a String contains non-whitespace characters.

This change only affects Java 11.

Issue: #2109 

### DIFF
--- a/jcl/src/java.base/share/classes/java/lang/String.java
+++ b/jcl/src/java.base/share/classes/java/lang/String.java
@@ -2563,6 +2563,26 @@ public final class String implements Serializable, Comparable<String>, CharSeque
 		
 		return (result == null) ? this : result;
 	}
+
+	/**
+	 * Determine if the string contains only white space characters. 
+	 * 
+	 * @return true if the string is empty or contains only white space
+	 * characters, otherwise false.
+	 * 
+	 * @since 11
+	 */
+	public boolean isBlank() {
+		int index;
+
+		if (enableCompression && (null == compressionFlag || coder == LATIN1)) {
+			index = StringLatin1.indexOfNonWhitespace(value);
+		} else {
+			index = StringUTF16.indexOfNonWhitespace(value);
+		}
+		
+		return index >= lengthInternal();
+	}
 /*[ENDIF]*/
 	
 	/**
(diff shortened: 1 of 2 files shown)


jcl/src/java.base/share/classes/java/lang/String.java
```diff
@@ -2563,6 +2563,26 @@ public String stripTrailing() {
...
+	public boolean isBlank() {
+		int index;
+
+		if (enableCompression && (null == compressionFlag || coder == LATIN1)) {
+			index = StringLatin1.indexOfNonWhitespace(value);
+		} else {
+			index = StringUTF16.indexOfNonWhitespace(value);
+		}
+		
+		return (index < lengthInternal()) ? false : true;
```
fjeremic: This can simply be `return index >= lengthInternal()`

Ali-Ni:
@tajila fyi

fjeremic:
Will wait for a review from @tajila before merging. Launching builds preemptively.

Jenkins test sanity

jcl/src/java.base/share/classes/java/lang/String.java
```diff
@@ -2563,6 +2563,26 @@ public String stripTrailing() {
...
+	 * Determine if the string contains only white space characters. 
+	 * 
+	 * @return true if the string is empty or contains only white space
+	 * characters, otherwise false.
+	 * 
+	 * @since 11
+	 */
+	public boolean isBlank() {
+		int index;
+
```
tajila: The first  thing `indexOfNonWhitespace` does is check the length of the array (which is basically the equivalent of `if (count == 0)`. I think we can hoist that check out and skip the rest if its zero. This has benefits when running interpreted, not sure how much this matters to compiled code. 
  
@fjeremic  thoughts?

fjeremic: I don't think there is benefit to duplicating the check. It makes the method larger and less susceptible to inlining at basically no benefit. I don't think we should optimize towards interpreting as nearly everything gets JIT compiled.

tajila: sounds good, no further comments then

fjeremic:

PR #2427, opened 2018-07-18
Fix compile make targets

theresa-m:


keithc-ca:
Dumb question: how does this change help anything?

pshipton:
Given the internal build failures we found without the change, I'm guessing it makes the variable available in more places where it is required.

PR #2430, opened 2018-07-19
Fix unresolve omrsig_handler for j9ddrgen and constgen executables.

jjohnst-us:
z/TPF has strict unresolve checking.  During the z/TPF build both
j9ddrgen and constgen link checks show omrsig_hanlder as being
unresolved and causes a build failure.  Linking in the omrsig
library fixes this issue.


pshipton:
jenkins compile plinux

PR #2432, opened 2018-07-19
Validate zTPF Memory Reserve Factor during Xmx Validation

jjohnst-us:
Validate zTPF Memory Reserve Factor during Xmx Validation.
z/TPF doesn't have mmap and has a relatively small usable
memory footprint compared to linux.  When specifying an Xmx
for the z/TPF platform there is a higher chance of specifying
a higher amount then made available for the process.  This
update does a quick check during validation instead of waiting
for the runtime to detect it an run dump agents related to OOM.

### DIFF
--- a/runtime/gc_modron_startup/mminit.cpp
+++ b/runtime/gc_modron_startup/mminit.cpp
@@ -223,7 +223,7 @@ cleanupMutatorModelJava(J9VMThread* vmThread)
 	J9VMDllLoadInfo* loadInfo;
 	J9JavaVM* vm = vmThread->javaVM;
 	MM_EnvironmentBase *env = MM_EnvironmentBase::getEnvironment(vmThread->omrVMThread);
-	
+
 	if (NULL != env) {
 		/* cleanupMutatorModelJava is called as part of the main vmThread shutdown, which happens after
 		 * gcCleanupHeapStructures has been called. We should therefore only flush allocation caches
@@ -235,7 +235,7 @@ cleanupMutatorModelJava(J9VMThread* vmThread)
 			GC_OMRVMThreadInterface::flushCachesForGC(env);
 		}
 	}
-	
+
 	cleanupMutatorModel(vmThread->omrVMThread, 0);
 
 	vmThread->gcExtensions = NULL;
@@ -267,15 +267,15 @@ reportPrivateHeapDelete(J9JavaVM * javaVM, void * memorySpace)
 void
 gcCleanupHeapStructures(J9JavaVM * vm)
 {
-	/* If shutdown occurs early (due to command line parsing errors, for example) there may not be 
-	 * a J9VMThread, so allocate a fake environment. 
+	/* If shutdown occurs early (due to command line parsing errors, for example) there may not be
+	 * a J9VMThread, so allocate a fake environment.
 	 */
 	MM_EnvironmentBase env(vm->omrVM);
 	MM_GCExtensions *extensions = MM_GCExtensions::getExtensions(vm);
 
 	/* remove hooks installed by Validator */
 	gcCleanupVMHooks(extensions);
-	
+
 	/* Flush any allocation contexts so that their memory is returned to the memory spaces before we tear down the memory spaces */
 	MM_GlobalAllocationManager *gam = extensions->globalAllocationManager;
 	if (NULL != gam) {
@@ -546,7 +546,7 @@ gcInitializeHeapStructures(J9JavaVM *vm)
 	if (!gcInitializeVMHooks(extensions)) {
 		goto error;
 	}
-	
+
 	vm->defaultMemorySpace = defaultMemorySpace;
 
 	return J9VMDLLMAIN_OK;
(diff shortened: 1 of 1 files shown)


pshipton:
@dmitripivkine any concerns?

dmitripivkine:
Why we need this change? I believe memoryMax can not be larger then omrsysinfo_get_addressable_physical_memory() any way:
- in MM_GCExtensionsBase::computeDefaultMaxHeap()
default memory size is selected as min(omrsysinfo_get_addressable_physical_memory()/2, 512m)
- in MM_GCExtensions::computeDefaultMaxHeap() it might be extended to omrsysinfo_get_addressable_physical_memory()/4 if it is larger (2.9 and higher only)

Did I miss something?

runtime/gc_modron_startup/mminit.cpp
```diff
@@ -1089,6 +1096,18 @@ gcInitializeXmxXmdxVerification(J9JavaVM *javaVM, IDATA* memoryParameters, bool
 		}
 	}
 
+#if defined(J9ZTPF)
+
+        usableMemory = omrsysinfo_get_addressable_physical_memory();
+
```
dmitripivkine: The output of omrsysinfo_get_addressable_physical_memory() is stored in extensions->usablePhysicalMemory already. Please use it instead of Port Library call

runtime/gc_modron_startup/mminit.cpp
```diff
@@ -1012,9 +1012,16 @@ gcInitializeXmxXmdxVerification(J9JavaVM *javaVM, IDATA* memoryParameters, bool
 	/* Error reporting */
 	const char *subSpaceTooLargeOption;
 
+#if defined(J9ZTPF)
+    U_64 usableMemory;
+#endif /* defined(J9ZTPF) */
+
```
dmitripivkine: this is not necessary if use extensions->usablePhysicalMemory

runtime/gc_modron_startup/mminit.cpp
```diff
@@ -1012,9 +1012,16 @@ gcInitializeXmxXmdxVerification(J9JavaVM *javaVM, IDATA* memoryParameters, bool
...
 	bool opt_XmxSet = (-1 != memoryParameters[opt_Xmx]);
 	bool opt_XmdxSet = (-1 != memoryParameters[opt_Xmdx]);
 	PORT_ACCESS_FROM_JAVAVM(javaVM);
+#if defined(J9ZTPF)
+	OMRPORT_ACCESS_FROM_J9PORT(PORTLIB);
+#endif /* defined(J9ZTPF) */
```
dmitripivkine: this is not necessary if use extensions->usablePhysicalMemory

runtime/gc_modron_startup/mminit.cpp
```diff
@@ -1089,6 +1096,18 @@ gcInitializeXmxXmdxVerification(J9JavaVM *javaVM, IDATA* memoryParameters, bool
 		}
 	}
 
+#if defined(J9ZTPF)
```
dmitripivkine: I believe this check can be introduced for all platforms, not ZTPF only

runtime/gc_modron_startup/mminit.cpp
```diff
@@ -1089,6 +1096,18 @@ gcInitializeXmxXmdxVerification(J9JavaVM *javaVM, IDATA* memoryParameters, bool
...
+
+        usableMemory = omrsysinfo_get_addressable_physical_memory();
+
+        if (extensions->memoryMax > usableMemory)  {
```
dmitripivkine: Would you please add comment there that this is possible if -Xmx is set only?

jjohnst-us:
@dmitripivkine 
All review comments applied I believe, thank you.

runtime/gc_modron_startup/mminit.cpp
```diff
@@ -1089,6 +1089,15 @@ gcInitializeXmxXmdxVerification(J9JavaVM *javaVM, IDATA* memoryParameters, bool
 		}
 	}
 
+	/* Only if Xmx is specified, verify the value doesn't go over usable memory. */
```
dmitripivkine: I apologize. I took one comment back. Would you please return back defined(J9ZTPF) wrapping for this section?

runtime/gc_modron_startup/mminit.cpp
```diff
@@ -1089,6 +1089,15 @@ gcInitializeXmxXmdxVerification(J9JavaVM *javaVM, IDATA* memoryParameters, bool
...
+	if (opt_XmxSet) {
+        if (extensions->memoryMax > extensions->usablePhysicalMemory)  {
```
dmitripivkine: please use <tab> for indentation. Also should be a single space between ) and {

runtime/gc_modron_startup/mminit.cpp
dmitripivkine: @jjohnst-us Would you please fix last issues from two last comments and this will be good to go?

dmitripivkine:
Skip Sanity build as far as the change is ZTPF only

PR #2433, opened 2018-07-19
Accelerate StringBuilder append for Java8 by removing substrings

msqb:
- Replace the creation of substrings with an arraycopy
  to StringBuilder's value array.

- Remove Java 9 path as it won't be compiled for Java 9+

### DIFF
--- a/jcl/src/java.base/share/classes/java/lang/StringBuilder.java
+++ b/jcl/src/java.base/share/classes/java/lang/StringBuilder.java
@@ -2261,7 +2261,49 @@ public StringBuilder append(CharSequence sequence, int start, int end) {
 	if (sequence == null) {
 		return append(String.valueOf(sequence), start, end);
 	} else if (sequence instanceof String) {
-		return append(((String)sequence).substring(start, end));
+		String sequenceString = (String) sequence;
+		if (start >= 0 && start <= end && end <= sequenceString.lengthInternal()) {
+
+			int currentLength = lengthInternal();
+			int currentCapacity = capacityInternal();
+			int newLength = currentLength + end - start;
+
+			if (String.enableCompression) {
+
+				// Check if the StringBuilder is compressed
+				if (count >= 0 && sequenceString.isCompressed()) {
+					if (newLength > currentCapacity) {
+						ensureCapacityImpl(newLength);
+					}
+					
+					sequenceString.getBytes(start, end, value, currentLength);
+					count = newLength;
+				} else {
+					// Check if the StringBuilder is compressed
+					if (count >= 0) {
+						decompress(newLength);
+						currentCapacity = capacityInternal();
+					}
+					
+					if (newLength > currentCapacity) {
+						ensureCapacityImpl(newLength);
+					}
+					
+					sequenceString.getCharsNoBoundChecks(start, end, value, currentLength);
+					count = newLength | uncompressedBit;
+				}
+			} else {
+				if (newLength > currentCapacity) {
+					ensureCapacityImpl(newLength);
+				}
+				
+				sequenceString.getCharsNoBoundChecks(start, end, value, currentLength);
+				count = newLength;
+			}
+			return this;
+		} else {
+			throw new IndexOutOfBoundsException();
+		}
 	} else if (start >= 0 && end >= 0 && start <= end && end <= sequence.length()) {
 		if (sequence instanceof StringBuffer) {
 			synchronized (sequence) {


jcl/src/java.base/share/classes/java/lang/StringBuilder.java
```diff
@@ -2261,7 +2261,47 @@ public StringBuilder append(CharSequence sequence, int start, int end) {
 	if (sequence == null) {
 		return append(String.valueOf(sequence), start, end);
 	} else if (sequence instanceof String) {
+		/*[IF Sidecar19-SE]*/
```
fjeremic: I don't think we need a Java 9 path here any longer. This file will never be compiled for Java 9+ anyway.

msqb:
Performance analysis was conducted using the DayTrader 3 and DayTrader 7 benchmarks.
Java build used: OpenJDK8-OPENJ9_x64_Linux_201818071200

DayTrader 3, average of 4 runs:
```
            mean        %stdev
no change   [baseline]  0.0104%
change      -0.4367%    0.0041%
```

DayTrader 7, average of 4 runs:
```
            mean        %stdev
no change   [baseline]  0.0085 %
change      -0.1248     0.0044 %
```

The performance of StringBuilder append(CharSequence sequence) reamins unchanged for OpenJDK8-OPENJ9.

jcl/src/java.base/share/classes/java/lang/StringBuilder.java
msqb: Thanks, removed the Java 9 path.

fjeremic:
> Performance analysis was conducted using the DayTrader 3 and DayTrader 7 benchmarks.
> Java build used: OpenJDK8-OPENJ9_x64_Linux_201818071200

Is negative better or worse?

msqb:
Negative percent is worse, added this information to the original post.

jcl/src/java.base/share/classes/java/lang/StringBuilder.java
```diff
@@ -2261,7 +2261,43 @@ public StringBuilder append(CharSequence sequence, int start, int end) {
...
+					decompress(newLength);
+				}
+
+				if (newLength > currentCapacity) {
+					ensureCapacityImpl(newLength);
+				}
+
+				for (int i = start; i < end; ++i) {
+					value[sbBase + i] = sequenceString.charAtInternal(i);
+				}
```
fjeremic: Why can't this be an arraycopy?

msqb: I'll try out arraycopy and see the perf impact. Marking this PR as WIP

msqb:
**Please note that the previously stated benchmark results did not take into account the correct stdev percentage value. The values have been correct and are presented below:**

Performance analysis was conducted using the DayTrader 3 and DayTrader 7 benchmarks.
Java build used: OpenJDK8-OPENJ9_x64_Linux_201818071200

DayTrader 3 Throughput (higher=better), average of 4 runs:
```
            mean        %stdev
no change   [baseline]  1.0422%
change      -0.4367%    0.4132%
```

DayTrader 7 Throughput (higher=better), average of 4 runs:
```
            mean        %stdev
no change   [baseline]  0.8523 %
change      -0.1248%     0.4389 %
```
[both benchmarks show a slightly worse performance for this change]

The performance of StringBuilder append(CharSequence sequence) remains unchanged for OpenJDK8-OPENJ9.

msqb:
```

File: jcl/src/java.base/share/classes/java/lang/StringBuilder.java (hunk 2 above)
Comment (msqb) [reply jcl/src/java.base/share/classes/java/lang/StringBuilder.java]: Results of the ArrayCopy benchmark are posted below.

File: jcl/src/java.base/share/classes/java/lang/StringBuilder.java [hunk 3]
```diff
@@ -2261,7 +2261,43 @@ public StringBuilder append(CharSequence sequence, int start, int end) {
 	if (sequence == null) {
 		return append(String.valueOf(sequence), start, end);
 	} else if (sequence instanceof String) {
-		return append(((String)sequence).substring(start, end));
+		if (start >= 0 && end >= 0 && start <= end && end <= ((String) sequence).lengthInternal()) {
```
DanHeidinga: Can `String sequenceString = (String) sequence;` be declared before the `if` so the redundant cast can be avoided?

jcl/src/java.base/share/classes/java/lang/StringBuilder.java
```diff
@@ -2261,7 +2261,43 @@ public StringBuilder append(CharSequence sequence, int start, int end) {
...
+			int currentLength = lengthInternal();
+			int currentCapacity = capacityInternal();
+			int newLength = lengthInternal() + end - start;
```
DanHeidinga: `lengthInternal() + end - start;` -> `currentLength + end - start;`

jcl/src/java.base/share/classes/java/lang/StringBuilder.java
```diff
@@ -2261,7 +2261,43 @@ public StringBuilder append(CharSequence sequence, int start, int end) {
...
+			int sbBase = currentLength - start;
+			String sequenceString = (String) sequence;
+
+			if (String.enableCompression) {
+				// Check if the StringBuilder is compressed
+				if (count >= 0) {
+					decompress(newLength);
+				}
+
+				if (newLength > currentCapacity) {
```
DanHeidinga: `currentCapacity` may not be correct here if a `decompress()` occurred.  Because this uses `ensureCapacityImpl`, this may result in additional array copies

jcl/src/java.base/share/classes/java/lang/StringBuilder.java
```diff
@@ -2261,7 +2261,43 @@ public StringBuilder append(CharSequence sequence, int start, int end) {
...
+					ensureCapacityImpl(newLength);
+				}
+
+				for (int i = start; i < end; ++i) {
+					value[sbBase + i] = sequenceString.charAtInternal(i);
+				}
+
+				count = newLength;
+			}
+			return this;
```
DanHeidinga: The two branches of the `if` statement are almost identical and can be simplified to:
```java
int newCount = newLength;
if (String.enableCompression) {
	// Check if the StringBuilder is compressed
	if (count >= 0) {
		decompress(newLength);
	}
	newCount |= uncompressedBit;
}

if (newLength > currentCapacity) {
	ensureCapacityImpl(newLength);
}

for (int i = start; i < end; ++i) {
	value[sbBase + i] = sequenceString.charAtInternal(i);
}
count = newCount;
```

Unless there's a compelling reason to duplicate the code?

msqb:
@DanHeidinga Thank you for the feedback. I've made the recommended changes and the results of the DayTrader 3 and 7 benchmarks are as follows:

Java build used: OpenJDK8-OPENJ9_x64_Linux_201818071200

DayTrader 3 Throughput (higher=better), average of 4 runs
```
                            mean        %stdev
no change                   [baseline]  0.558%
current change (for loop)   +0.089%     0.600%
arraycopy                   -0.275%     0.598%  
```

DayTrader 7 Throughput (higher=better), average of 4 runs
```
                            mean        %stdev
no change                   [baseline]  0.204%
current change (for loop)   +0.274%     1.181%
arraycopy                   +1.071%      0.499%   
```

The current change includes the simplification Dan suggested.
Both the current change and the arraycopy variation also get the currentCapacity again after the decompress as Dan mentioned.
```
decompress(newLength);
currentCapacity = capacityInternal();
```

The current change looks to be on par with the baseline, with a slight improvement in mean throughput over 4 runs for both DayTrader 3 and DayTrader 4.

The arraycopy variation shows a 1% increase for DayTrader 7, but a 0.275% regression in DayTrader 3 for mean throughput over 4 runs.

fjeremic:
@msqb the degradation in DT3 is statistically insignificant given the standard deviation for the run. The confidence interval for DT7 with arraycopy shows a clear improvement. The change should be good IMO.

jcl/src/java.base/share/classes/java/lang/StringBuilder.java
```diff
@@ -2261,7 +2261,49 @@ public StringBuilder append(CharSequence sequence, int start, int end) {
...
+		if (start >= 0 && start <= end && end <= sequenceString.lengthInternal()) {
+
+			int currentLength = lengthInternal();
+			int currentCapacity = capacityInternal();
+			int newLength = currentLength + end - start;
+
+			if (String.enableCompression) {
+
+				// Check if the StringBuilder is compressed
+				if (count >= 0 && sequenceString.isCompressed ()) {
```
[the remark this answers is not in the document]
msqb: Fixed, thanks

fjeremic:

PR #2450, opened 2018-07-23
Locking Changes

babsingh:
1) **Revert "Improve locking"**

	This commit reverts the first commit in #553.

	The purpose of the original change was to reduce compare and swap
	operations in the spinlock loop; thus, reducing spinning overhead. The
	premise was to improve throughput by reducing spinning overhead. In the
	initial benchmarking, no throughput improvements were noticed because of
	this change. Recently, this change has resulted in regressing
	throughput. ~30% throughput drop is observed in one of the OpenLiberty
	benchmarks. Thus, it has been decided to revert the original change.

### DIFF
--- a/runtime/vm/ObjectMonitor.cpp
+++ b/runtime/vm/ObjectMonitor.cpp
@@ -489,20 +489,24 @@ spinOnTryEnter(J9VMThread *currentThread, J9ObjectMonitor *objectMonitor, j9obje
 	}
 #endif /* OMR_THR_JLM */
 
+#if defined(OMR_THR_THREE_TIER_LOCKING) && defined(OMR_THR_SPIN_WAKE_CONTROL)
+	bool tryEnterSpin = true;
+	if (monitor->spinThreads < lib->maxSpinThreads) {
+		VM_AtomicSupport::add(&monitor->spinThreads, 1);
+	} else {
+		tryEnterSpinCount1 = 1;
+		tryEnterSpinCount2 = 1;
+		tryEnterYieldCount = 1;
+		tryEnterSpin = false;
+	}
+#endif /* defined(OMR_THR_THREE_TIER_LOCKING) && defined(OMR_THR_SPIN_WAKE_CONTROL) */
+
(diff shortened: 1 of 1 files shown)


babsingh:
This PR depends on https://github.com/eclipse/omr/pull/2783.

babsingh:
fyi - @pshipton @charliegracie @vijaysun-omr

pshipton:
jenkins test sanity plinux depends eclipse/omr#2783

pshipton:
Is the benchmark internal or public? If public please mention it by name for future reference.

babsingh:
not sure if the benchmark is internal or public. i have requested information about the benchmark.

i have accommodated Charlie's request of pushing the new `-Xthr:maxSpinThreads=0` option in a new pull request.

pshipton:
jenkins test sanity plinux depends eclipse/omr#2783

pshipton:

pshipton:
This change is now independent of eclipse/omr#2783

DanHeidinga:
> Recently, this change has resulted in regressing
throughput. ~30% throughput drop 

@babsingh Have you investigated how this change is responsible for the ~30% throughput drop?  
Is the regression related to particular machine configurations (ie: # of cores)?   #553 defined `maxSpinThreads` differently based on the number of cores and I'm wondering if this is might be addressed by adjusting the maxSpinThreads heuristic.

babsingh:
@DanHeidinga I had suggested to vary maxSpinThreads in order to restore the performance but it didn't help. The regression happened on a 2 CPU machine. In #553, value of maxSpinThreads for 1-4 CPUs wasn't changed.

Before #553, non-spinning threads were allowed to perform one CAS operation before exiting the spinlock loop. In #553, this was stopped to reduce spinning overhead, and non-spinning threads exited the spinlock loop without the CAS. At that time, no throughput improvements or regressions were noticed due to this change. Only spinning overhead was reduced. Adding this change back restored the throughput.

babsingh:
correct - **now, OpenJ9 uses more CPU cycles but recovers the ~30% throughput regression**.

Issue #2139, opened 2018-06-08
Test-Sanity-JDK8-linux_ppc-64 cmdLineTester_decompilationTests_1 decomp002 assert


pshipton:
https://ci.eclipse.org/openj9/job/Test-Sanity-JDK8-linux_ppc-64_cmprssptrs_le/289/

Occurred during omr acceptance.

```
Testing: decomp002
Test start time: 2018/06/08 21:55:41 Coordinated Universal Time
Running command: "/home/jenkins/jenkins-agent/workspace/Test-Sanity-JDK8-linux_ppc-64_cmprssptrs_le/j2sdk-image/jre/bin/java"  -Xcompressedrefs -Xcompressedrefs -Xgcpolicy:gencon -Xjit:enableOSR,count=0  -Xdump -Xgcthreads1  -Xdump:system:events=abort  -agentlib:jvmtitest=test:decomp002 -cp "/home/jenkins/jenkins-agent/workspace/Test-Sanity-JDK8-linux_ppc-64_cmprssptrs_le/openj9/test/TestConfig/scripts/testKitGen/../../../../jvmtest/functional/cmdLineTests/jvmtitests/jvmtitest.jar" com.ibm.jvmti.tests.util.TestRunner
Time spent starting: 4 milliseconds
Time spent executing: 9943 milliseconds
Test result: FAILED
Output from test:
 [OUT] *** Testing [1/1]:	testStepping
 [OUT] 	Step 10000000 times
 [ERR] 21:55:50.926 0x12a00j9codertvm(j9ji.104    *   ** ASSERTION FAILED ** at decomp.cpp:584: (((IDATA)pendingStackHeight >= 0))
 [ERR] JVMDUMP039I Processing dump event "traceassert", detail "" at 2018/06/08 21:55:50 - please wait.
 [ERR] JVMDUMP032I JVM requested System dump using '/home/jenkins/jenkins-agent/workspace/Test-Sanity-JDK8-linux_ppc-64_cmprssptrs_le/openj9/test/functional/cmdLineTests/jvmtitests/core.20180608.215550.19560.0001.dmp' in response to an event
 [ERR] JVMPORT030W /proc/sys/kernel/core_pattern setting "|/usr/share/apport/apport %p %s %c %d %P" specifies that the core dump is to be piped to an external program.  Attempting to rename either core or core.19612.
 [ERR] 
[...]

PR #2453, opened 2018-07-23
Use MethodMonitorEntry for inlined callee in OSR

cathyzhyi:
VM expects jitMethodMonitorEntry to be used for monitor enter in
inlined synchronized callee under OSR. Change JIT to call the
correct helper so that VM get the expected stack frame type.

issue: #2139

### DIFF
--- a/runtime/compiler/ilgen/IlGenerator.cpp
+++ b/runtime/compiler/ilgen/IlGenerator.cpp
@@ -987,10 +987,7 @@ TR_J9ByteCodeIlGenerator::prependEntryCode(TR::Block * firstBlock)
       loadMonitorArg();
 
       TR::Node * firstChild = pop();
-      TR::SymbolReference * monEnterSymRef =
-         isOutermostMethod() ?
-           symRefTab()->findOrCreateMethodMonitorEntrySymbolRef(_methodSymbol) :
-           symRefTab()->findOrCreateMonitorEntrySymbolRef(_methodSymbol);
+      TR::SymbolReference * monEnterSymRef = symRefTab()->findOrCreateMethodMonitorEntrySymbolRef(_methodSymbol);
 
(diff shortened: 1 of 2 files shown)


runtime/compiler/ilgen/IlGenerator.cpp
```diff
@@ -988,7 +988,7 @@ TR_J9ByteCodeIlGenerator::prependEntryCode(TR::Block * firstBlock)
 
       TR::Node * firstChild = pop();
       TR::SymbolReference * monEnterSymRef =
-         isOutermostMethod() ?
+         (isOutermostMethod() || comp()->getOption(TR_EnableOSR))?
```
gacholio: Wouldn't it be better to just always use jitMethodMonitorEntry here?  If OSR isn't enabled, there is only the outermost method.

runtime/compiler/ilgen/Walker.cpp
```diff
@@ -6484,7 +6484,7 @@ void
 TR_J9ByteCodeIlGenerator::genMonitorExit(bool isReturn)
    {
    TR::SymbolReference * monitorExitSymbolRef =
-      isReturn && isOutermostMethod() ?
+      isReturn && (isOutermostMethod() || comp()->getOption(TR_EnableOSR)) ?
```
gacholio: Similarly here, isn't isReturn sufficient to select the symbol?

runtime/compiler/ilgen/IlGenerator.cpp
cathyzhyi: Thanks for the suggestion. changed accordingly.

runtime/compiler/ilgen/Walker.cpp
cathyzhyi: changed accordingly.

gacholio:

PR #2454, opened 2018-07-23
Improve test dependency acquisition

keithc-ca:
If a file exists, verify its SHA before skipping the download step: issue https://github.com/eclipse/openj9/issues/2440


llxia:
Tested in AdoptOpenJDK:
https://ci.adoptopenjdk.net/view/work%20in%20progress/job/Grinder/76/

llxia:
jenkins test sanity xlinux,win jdk10

PR #2457, opened 2018-07-23
Add Constant Dynamic in GC Implementation

Ailloviee:
Implemented condy iterator in constant pool and added it to the GC
Implementation
(/runtime/gc_glue_java/CollectorLanguageInterfaceImpl.cpp).

This is dependent on https://github.com/eclipse/openj9/pull/1533


runtime/gc_glue_java/CollectorLanguageInterfaceImpl.cpp
```diff
@@ -428,6 +432,16 @@ MM_CollectorLanguageInterfaceImpl::scavenger_hasIndirectReferentsInNewSpace(MM_E
 				}
 			}
 		}
+		GC_ConstantPoolConstantDynamicSlotIterator constantPoolConstantDynamicSlotIterator(env, classToScan);
+		while(NULL != (slotPtr = (omrobjectptr_t*)constantPoolConstantDynamicSlotIterator.nextSlot())) {
+			omrobjectptr_t condyObjectPtr = *slotPtr;
```
dmitripivkine: why not to to have just "objectPtr" name?

runtime/gc_structs/ConstantPoolConstantDynamicSlotIterator.cpp
```diff
@@ -0,0 +1,74 @@
...
+			_cpDescription = *_cpDescriptionSlots;
+			_cpDescriptionSlots += 1;
+			_cpDescriptionIndex = J9_CP_DESCRIPTIONS_PER_U32;
+		}
+
+		slotType = _cpDescription & J9_CP_DESCRIPTION_MASK;
+		slotPtr = _cpEntry;
+
+		/* Adjust the CP slot and description information */
+		_cpEntry = (J9Object **)( ((U_8 *)_cpEntry) + sizeof(J9RAMConstantPoolItem) );
```
dmitripivkine: please do not put extra spaces between brackets

runtime/gc_glue_java/CollectorLanguageInterfaceImpl.cpp
```diff
@@ -394,6 +394,10 @@ MM_CollectorLanguageInterfaceImpl::scavenger_scavengeIndirectObjectSlots(MM_Envi
 		while((slotPtr = classStaticsIterator.nextSlot()) != NULL) {
 			shouldBeRemembered = _extensions->scavenger->copyObjectSlot(env, slotPtr) || shouldBeRemembered;
 		}
+		GC_ConstantPoolConstantDynamicSlotIterator constantPoolConstantDynamicSlotIterator(env, classToScan);
+		while((slotPtr = constantPoolConstantDynamicSlotIterator.nextSlot()) != NULL) {
```
tajila: add space between `while` and `(`

runtime/gc_glue_java/CollectorLanguageInterfaceImpl.cpp
```diff
@@ -428,6 +432,16 @@ MM_CollectorLanguageInterfaceImpl::scavenger_hasIndirectReferentsInNewSpace(MM_E
 				}
 			}
 		}
+		GC_ConstantPoolConstantDynamicSlotIterator constantPoolConstantDynamicSlotIterator(env, classToScan);
+		while(NULL != (slotPtr = (omrobjectptr_t*)constantPoolConstantDynamicSlotIterator.nextSlot())) {
```
tajila: same here

runtime/gc_glue_java/CollectorLanguageInterfaceImpl.cpp
```diff
@@ -428,6 +432,16 @@ MM_CollectorLanguageInterfaceImpl::scavenger_hasIndirectReferentsInNewSpace(MM_E
...
+			omrobjectptr_t condyObjectPtr = *slotPtr;
+			if (NULL != condyObjectPtr){
```
tajila: add space between `)` and `{`

runtime/gc_glue_java/CollectorLanguageInterfaceImpl.cpp
```diff
@@ -447,6 +461,10 @@ MM_CollectorLanguageInterfaceImpl::scavenger_fixupIndirectObjectSlots(MM_Environ
 		while((slotPtr = classStaticsIterator.nextSlot()) != NULL) {
 			_extensions->scavenger->fixupSlotWithoutCompression(slotPtr);
 		}
+		GC_ConstantPoolConstantDynamicSlotIterator constantPoolConstantDynamicSlotIterator(env, classToScan);
+		while((slotPtr = constantPoolConstantDynamicSlotIterator.nextSlot()) != NULL) {
```
tajila: same here

runtime/gc_glue_java/CollectorLanguageInterfaceImpl.cpp
```diff
@@ -465,6 +483,10 @@ MM_CollectorLanguageInterfaceImpl::scavenger_backOutIndirectObjectSlots(MM_Envir
 		while((slotPtr = classStaticsIterator.nextSlot()) != NULL) {
 			_extensions->scavenger->backOutFixSlotWithoutCompression(slotPtr);
 		}
+		GC_ConstantPoolConstantDynamicSlotIterator constantPoolConstantDynamicSlotIterator(env, classToScan);
+		while((slotPtr = constantPoolConstantDynamicSlotIterator.nextSlot()) != NULL) {
```
tajila: and here

runtime/gc_structs/ConstantPoolConstantDynamicSlotIterator.hpp
```diff
@@ -0,0 +1,83 @@
...
+	U_32 *_cpDescriptionSlots;
+	U_32 _cpDescription;
+	UDATA _cpDescriptionIndex;
+	
+public:
+	GC_ConstantPoolConstantDynamicSlotIterator(MM_EnvironmentBase *env, J9Class *clazz) :
+		_cpEntry((J9Object **)J9_CP_FROM_CLASS(clazz)),
+		_cpEntryCount(clazz->romClass->ramConstantPoolCount)
+	{
+		if(J2SE_VERSION((J9JavaVM*)env->getLanguageVM()) < J2SE_V11) {
```
tajila: add a comment here saying:
```
TODO: detect whether class can have any Condy entries
```

tajila: Also update the commit message saying this is the firs part of two changes, the next change will check if the class contains any Condy cp entries

runtime/gc_structs/ConstantPoolConstantDynamicSlotIterator.hpp
```diff
@@ -0,0 +1,84 @@
...
+		/*TODO: detect whether class can have any Condy entries*/
+			_cpEntryCount = 0;
+		}
+		_cpEntryTotal = _cpEntryCount;
+		if(_cpEntryCount) {
+			_cpDescriptionSlots = SRP_PTR_GET(&clazz->romClass->cpShapeDescription, U_32 *);
+			_cpDescriptionIndex = 0;
+		}
+
```
dmitripivkine: remove this empty line please

amicic:
Another dilemma is: should we just extend nextSlot() for GC_ClassStaticsIterator to iterate these new slots too. so we don't 'pollute' CollectorLanguageInterface?

tajila:
> I have a related question: when you access those fields anywhere in VM code, do you do it directly? I think you should be using J9STATIC_ADDRESS_LOAD/STORE and similar macros/APIs.

Yes, they are accessed directly. We currently do not have any 'barrier' macros for accessing objects in the constantpool

amicic:
I don't have a strong opinion either, so I guess, leaving 'as is' ok for now. 

Btw if the new top level iterator is to be eventaully introduced ClassIndirect  in the name would not be quite right - we are talking about _Object_ Indirect

dmitripivkine:
I noticed we do currently have logic to iterate Constant Dynamic objects in ConstantPoolObjectSlotIterator. So an idea to introduce another ConstantPoolConstantDynamicSlotIterator looks wrong. This duplication of code should be avoided. Looks like right way to go is modify ConstantPoolObjectSlotIterator to add mode to return back Constant Dynamic objects only.

runtime/gc_structs/ConstantPoolObjectSlotIterator.hpp
```diff
@@ -69,9 +69,9 @@ class GC_ConstantPoolObjectSlotIterator
 	 */
 	MMINLINE IDATA getIndex() {
 		return _cpEntryTotal - _cpEntryCount - 1;
-	}
+	};
 
-	j9object_t *nextSlot();
+	j9object_t *nextSlot(Boolean condyOnly, MM_EnvironmentBase *env);
```
tajila: I think by convention `MM_EnvironmentBase *env` should be first

runtime/gc_glue_java/CollectorLanguageInterfaceImpl.cpp
```diff
@@ -384,6 +384,7 @@ bool
 MM_CollectorLanguageInterfaceImpl::scavenger_scavengeIndirectObjectSlots(MM_EnvironmentStandard *env, omrobjectptr_t objectPtr)
 {
 	bool shouldBeRemembered = false;
+	bool condyOnly = true;
```
tajila: don't declare the local, just pass `true` to the function

runtime/gc_structs/ConstantPoolObjectSlotIterator.cpp
```diff
@@ -35,35 +35,69 @@
...
+			/* Adjust the CP slot and description information */
+			_cpEntry = (j9object_t *) (((U_8 *) _cpEntry)
+					+ sizeof(J9RAMConstantPoolItem));
+			_cpEntryCount -= 1;
+
+			_cpDescription >>= J9_CP_BITS_PER_DESCRIPTION;
+			_cpDescriptionIndex -= 1;
+
+			/* Determine if the slot should be processed */
+			switch (slotType) {
```
tajila: I think you can recude the amount of code duplication if you put the if around the `switch` instead

runtime/gc_structs/ConstantPoolObjectSlotIterator.cpp
```diff
@@ -35,18 +35,22 @@
...
  * @return NULL if there are no more object references
+ *
+ * If condyOnly flag is set and Java version is SE11 and above:
+ * @return the next Condy reference from the constant pool
+ * @return NULL if there are no more references
  */
 j9object_t *
-GC_ConstantPoolObjectSlotIterator::nextSlot()
-{
+GC_ConstantPoolObjectSlotIterator::nextSlot(MM_EnvironmentBase *env, Boolean condyOnly) {
```
amicic: condyOnly should be a member of the class, passed only once to the constructor. no need to pass it every time on nextSlot()

gacholio:
Any update here? Can this be staged (iterate CP in scavenger, optimize for no new stores later)?

runtime/gc_structs/ConstantPoolObjectSlotIterator.hpp
```diff
@@ -48,18 +50,41 @@ class GC_ConstantPoolObjectSlotIterator
 	U_32 *_cpDescriptionSlots;
 	U_32 _cpDescription;
 	UDATA _cpDescriptionIndex;
+	bool _condyOnly;
+	bool _condyEnabled;
```
amicic: I'm not strictly opposed to this _condyEnabled flag, but wondring if it's really necessary. Would not iterator even without this flag return null in a case the system is not 'enabled' by simply not finding any slot of Constant Dynamic type?

gacholio: If it's really important, I suspect a compile-time test of the JDK level could be used. Recent discussions with @charliegracie implied that we may want to scan Strings as well (stop tenuring String literals immediately).

amicic: I guess, there is performance implication, since the flag would help avoid the search, but then, I'm not sure how often we would instantiate and use the iterator with condyOnly on a not enabled system?

runtime/gc_structs/ConstantPoolObjectSlotIterator.cpp
```diff
@@ -61,9 +69,17 @@ GC_ConstantPoolObjectSlotIterator::nextSlot()
...
 		_cpDescriptionIndex -= 1;
-
-		/* Determine if the slot should be processed */
-		switch (slotType) {
+		/* If _condyOnly is TRUE, return the next constant dynamic slot in constant pool
+		 * Otherwise return the next object */
+		if (_condyOnly) {
+			/* Determine if the slot is constant dynamic */
+			if (slotType == J9CPTYPE_CONSTANT_DYNAMIC) {
+				result = &(((J9RAMConstantDynamicRef *) slotPtr)->value);
```
tajila: missing the exception check

runtime/gc_structs/ConstantPoolObjectSlotIterator.cpp
```diff
@@ -61,9 +69,17 @@ GC_ConstantPoolObjectSlotIterator::nextSlot()
...
 		_cpDescription >>= J9_CP_BITS_PER_DESCRIPTION;
 		_cpDescriptionIndex -= 1;
-
-		/* Determine if the slot should be processed */
-		switch (slotType) {
+		/* If _condyOnly is TRUE, return the next constant dynamic slot in constant pool
+		 * Otherwise return the next object */
+		if (_condyOnly) {
+			/* Determine if the slot is constant dynamic */
+			if (slotType == J9CPTYPE_CONSTANT_DYNAMIC) {
```
tajila: make a private function for this so the code isn't duplicated

runtime/gc_structs/ConstantPoolObjectSlotIterator.hpp
```diff
@@ -48,18 +50,41 @@ class GC_ConstantPoolObjectSlotIterator
...
+		_condyEnabled(true)
 	{
 		_cpEntryTotal = _cpEntryCount;
 		if(_cpEntryCount) {
 			_cpDescriptionSlots = SRP_PTR_GET(&clazz->romClass->cpShapeDescription, U_32 *);
 			_cpDescriptionIndex = 0;
 		}
+	};
 
+	GC_ConstantPoolObjectSlotIterator(MM_EnvironmentBase *env, J9Class *clazz, bool condyOnly) :
```
tajila: use one constructor

runtime/gc_glue_java/CollectorLanguageInterfaceImpl.cpp
```diff
@@ -394,6 +394,11 @@ MM_CollectorLanguageInterfaceImpl::scavenger_scavengeIndirectObjectSlots(MM_Envi
 		while((slotPtr = classStaticsIterator.nextSlot()) != NULL) {
 			shouldBeRemembered = _extensions->scavenger->copyObjectSlot(env, slotPtr) || shouldBeRemembered;
 		}
+		GC_ConstantPoolObjectSlotIterator constantPoolObjectSlotIterator((J9JavaVM*)_omrVM->_language_vm, classToScan, true);
+		while ((slotPtr = constantPoolObjectSlotIterator.nextSlot()) != NULL) {
+			shouldBeRemembered = _extensions->scavenger->copyObjectSlot(env, slotPtr) || shouldBeRemembered;
+			printf("GC!!\n");
```
DanHeidinga: Looks like the debug `printf()` call snuck into the PR.  Please remove it.

runtime/gc_glue_java/CollectorLanguageInterfaceImpl.cpp
```diff
@@ -428,6 +433,17 @@ MM_CollectorLanguageInterfaceImpl::scavenger_hasIndirectReferentsInNewSpace(MM_E
 				}
 			}
 		}
+		GC_ConstantPoolObjectSlotIterator constantPoolObjectSlotIterator((J9JavaVM*)_omrVM->_language_vm, classToScan, true);
+		while (NULL != (slotPtr = (omrobjectptr_t*)constantPoolObjectSlotIterator.nextSlot())) {
+			omrobjectptr_t objectPtr = *slotPtr;
+			if (NULL != objectPtr) {
+				if (_extensions->scavenger->isObjectInNewSpace(objectPtr)){
+					printf("GC!!\n");
```
DanHeidinga: Another printf

runtime/gc_glue_java/CollectorLanguageInterfaceImpl.cpp
```diff
@@ -447,6 +463,11 @@ MM_CollectorLanguageInterfaceImpl::scavenger_fixupIndirectObjectSlots(MM_Environ
 		while((slotPtr = classStaticsIterator.nextSlot()) != NULL) {
 			_extensions->scavenger->fixupSlotWithoutCompression(slotPtr);
 		}
+		GC_ConstantPoolObjectSlotIterator constantPoolObjectSlotIterator((J9JavaVM*)_omrVM->_language_vm, classToScan, true);
+		while ((slotPtr = constantPoolObjectSlotIterator.nextSlot()) != NULL) {
+			_extensions->scavenger->fixupSlotWithoutCompression(slotPtr);
+			printf("GC!!\n");
```
DanHeidinga: printf

runtime/gc_glue_java/CollectorLanguageInterfaceImpl.cpp
```diff
@@ -465,6 +486,11 @@ MM_CollectorLanguageInterfaceImpl::scavenger_backOutIndirectObjectSlots(MM_Envir
 		while((slotPtr = classStaticsIterator.nextSlot()) != NULL) {
 			_extensions->scavenger->backOutFixSlotWithoutCompression(slotPtr);
 		}
+		GC_ConstantPoolObjectSlotIterator constantPoolObjectSlotIterator((J9JavaVM*)_omrVM->_language_vm, classToScan, true);
+		while ((slotPtr = constantPoolObjectSlotIterator.nextSlot()) != NULL) {
+			_extensions->scavenger->backOutFixSlotWithoutCompression(slotPtr);
+			printf("GC!!\n");
```
DanHeidinga: printf

Ailloviee: sorry I just made some changes earlier this day and I'm still testing it

gacholio:
It's not clear to me from reading the code if both slots in the CP entry are walked for condy (both contain objects).

runtime/gc_structs/ClassIterator.hpp
```diff
@@ -80,7 +80,7 @@ class GC_ClassIterator {
 		, _state(classiterator_state_start)
 		, _scanIndex(0)
 		, _classStaticsIterator(env, clazz)
-		, _constantPoolObjectSlotIterator(clazz)
+		, _constantPoolObjectSlotIterator((J9JavaVM*)env->getLanguageVM(), clazz)
```
dmitripivkine: would you please add space before star (J9JavaVM *) ?

runtime/gc_structs/ClassIterator.hpp
```diff
@@ -92,7 +92,7 @@ class GC_ClassIterator {
 		, _state(classiterator_state_start)
 		, _scanIndex(0)
 		, _classStaticsIterator(extensions, clazz)
-		, _constantPoolObjectSlotIterator(clazz)
+		, _constantPoolObjectSlotIterator((J9JavaVM*)extensions->getOmrVM()->_language_vm, clazz)
```
dmitripivkine: would you please add space before star (J9JavaVM *) ?

tajila:
@gacholio I thought only one of the two slots can contain an object at any point in time

gacholio:
@tajila Maybe only one at a time, but both are object slots - we certainly don't want the GC decoding the resolution state of the entry. Both slots need to be walked.

dmitripivkine:
Jenkins test sanity pLinux jdk11

runtime/gc_structs/ConstantPoolObjectSlotIterator.hpp
```diff
@@ -48,18 +49,44 @@ class GC_ConstantPoolObjectSlotIterator
...
+	 * @return exception otherwise
+	 */
+	MMINLINE j9object_t* scanCondySlot(j9object_t *slotPtr)
+	{
+		j9object_t * result;
+		if (NULL != ((J9RAMConstantDynamicRef *) slotPtr)->value) {
+			result = &(((J9RAMConstantDynamicRef *) slotPtr)->value);
+		} else {
+			result = &(((J9RAMConstantDynamicRef *) slotPtr)->exception);
+		}
```
DanHeidinga: Did GAC's comment about both slots needing to be scanned get addressed?  This looks like it only scans 1 or the other

tajila: @DanHeidinga not yet 

@Ailloviee  you'll need to add  additional state information to return two slots from one cpentry

gacholio: The code is technically correct at the moment, but it's a timebomb waiting to go off.

runtime/gc_structs/ConstantPoolObjectSlotIterator.hpp
```diff
@@ -48,18 +49,52 @@ class GC_ConstantPoolObjectSlotIterator
...
+	MMINLINE j9object_t* scanCondySlot(j9object_t *slotPtr)
+	{
+		j9object_t * result;
+		if (_isPreviousCondy) {
+			result = &(((J9RAMConstantDynamicRef *) slotPtr)->exception);
+			_isPreviousCondy = false;
+		} else {
+			result = &(((J9RAMConstantDynamicRef *) slotPtr)->value);
+			_isPreviousCondy = true;
+	         _cpEntry = (j9object_t *)( ((U_8 *)_cpEntry) - sizeof(J9RAMConstantPoolItem) );
```
tajila: this needs to be tabbed

runtime/gc_structs/ConstantPoolObjectSlotIterator.cpp
```diff
@@ -61,9 +69,17 @@ GC_ConstantPoolObjectSlotIterator::nextSlot()
 
 		_cpDescription >>= J9_CP_BITS_PER_DESCRIPTION;
 		_cpDescriptionIndex -= 1;
-
-		/* Determine if the slot should be processed */
-		switch (slotType) {
+		/* If _condyOnly is TRUE, return the next constant dynamic slot in constant pool
```
tajila: move the loop iteration variables above to the end of the loop and make them conditional based on wether we are staying in the same cpentry or not for the given state

runtime/gc_structs/ConstantPoolObjectSlotIterator.hpp
```diff
@@ -48,18 +49,52 @@ class GC_ConstantPoolObjectSlotIterator
...
+	bool _condyEnabled;
+	bool _isPreviousCondy;
 	
+private:
+	/**
+	 * Scan the current constant dynamic slot
+	 * @return slot value if it is not NULL
+	 * @return exception otherwise
+	 */
+	MMINLINE j9object_t* scanCondySlot(j9object_t *slotPtr)
```
tajila: add an output parameter for this function that returns if it is moving to the next slot or not

runtime/gc_structs/ConstantPoolObjectSlotIterator.hpp
```diff
@@ -48,18 +49,52 @@ class GC_ConstantPoolObjectSlotIterator
...
+	{
+		j9object_t * result;
+		if (_isPreviousCondy) {
+			result = &(((J9RAMConstantDynamicRef *) slotPtr)->exception);
+			_isPreviousCondy = false;
+		} else {
+			result = &(((J9RAMConstantDynamicRef *) slotPtr)->value);
+			_isPreviousCondy = true;
+			_cpEntry = (j9object_t *) (((U_8 *) _cpEntry) - sizeof(J9RAMConstantPoolItem));
```
tajila: once the loop variables are moved these do not need to be here

runtime/gc_structs/ConstantPoolObjectSlotIterator.cpp
```diff
@@ -75,19 +85,27 @@ GC_ConstantPoolObjectSlotIterator::nextSlot()
...
-				} else {
-					result = &(((J9RAMConstantDynamicRef *) slotPtr)->exception);
-				}
+				result = scanCondySlot(slotPtr, _stay);
 				break;
 			default:
 				continue;
+			}
+
+			break;
```
tajila: can't break here anymore

runtime/gc_structs/ConstantPoolObjectSlotIterator.cpp
```diff
@@ -75,19 +85,27 @@ GC_ConstantPoolObjectSlotIterator::nextSlot()
...
 			case J9CPTYPE_CONSTANT_DYNAMIC:
-				if (NULL != ((J9RAMConstantDynamicRef *) slotPtr)->value) {
-					result = &(((J9RAMConstantDynamicRef *) slotPtr)->value);
-				} else {
-					result = &(((J9RAMConstantDynamicRef *) slotPtr)->exception);
-				}
+				result = scanCondySlot(slotPtr, _stay);
 				break;
 			default:
 				continue;
```
tajila: default case cannot continue anymore

runtime/gc_structs/ConstantPoolObjectSlotIterator.cpp
```diff
@@ -75,19 +86,26 @@ GC_ConstantPoolObjectSlotIterator::nextSlot()
...
+			/* Adjust the CP slot and description information */
+			_cpEntry = (j9object_t *)( ((U_8 *)_cpEntry) + sizeof(J9RAMConstantPoolItem) );
+			_cpEntryCount -= 1;
+
+			_cpDescription >>= J9_CP_BITS_PER_DESCRIPTION;
+			_cpDescriptionIndex -= 1;
 		}
 
-		return result;
+		break;
```
dmitripivkine: how it will continue to iterate for case "default" in switch above?

runtime/gc_structs/ConstantPoolObjectSlotIterator.cpp
```diff
@@ -75,19 +86,26 @@ GC_ConstantPoolObjectSlotIterator::nextSlot()
 				result = &(((J9RAMMethodHandleRef *) slotPtr)->methodHandle);
 				break;
 			case J9CPTYPE_CONSTANT_DYNAMIC:
-				if (NULL != ((J9RAMConstantDynamicRef *) slotPtr)->value) {
-					result = &(((J9RAMConstantDynamicRef *) slotPtr)->value);
-				} else {
-					result = &(((J9RAMConstantDynamicRef *) slotPtr)->exception);
-				}
+				result = scanCondySlot(slotPtr, &nextSlot);
```
dmitripivkine: technically we can create another class to iterate objects from single condy slot and move this function there if you like (condySlotObhjecrtIterator?)... shift to next slot when iterator returns NULL. I am neutral about this so it is up to you

Ailloviee:
@dmitripivkine can you review this again please? The test has passed locally

DanHeidinga:
@dmitripivkine Is this ready to be tested and merged?

dmitripivkine:
I just double check latest version... LGTM.

dmitripivkine:
Jenkins test sanity pLinux jdk11

DanHeidinga:

dmitripivkine:

DanHeidinga:
Jenkins test sanity pLinux jdk11

DanHeidinga:

DanHeidinga:
Once the builds pass, who ever sees this first can merge it =)

DanHeidinga:
JDK 11 builds are failing for unrelated issues - naming missing native test libraries.  Merging as is as the the JDK8 build passed

ktbriggs:
@Ailloviee I think the assertion on line 440 of `CollectorLanguageInterfaceImpl.cpp`:

```
				Assert_MM_false(_evacuator->isInEvacuate(objectptr));
```

may fail in abort contexts. @amicic what do you think?

amicic:
@ktbriggs You might be right. That would be in a case of abort occurring during Remembered Set  scanning, which is rather rare. I'll think about how we would test it (cause RS abort to occur), since the current FVT test option would not do it.

amicic:
@ktbriggs The condition leading to this assert isObjectInNewSpace is somwhat misleading in its name- it actually means isObjectinSurvivorSpace (you can look at Scavenger code what it does). So the assert is ok.

However, it still may be an issue for callers of this functionality which may end up with an evacuate object even after abort is handled. For example Concurrent Scavenger (CS), or if we ever introduce object pre-pinning in Scavenger. I don't remember if CS however ends up calling it (after abort is handler)....

Bottom line, it would be wise to introduce a test case with aborted RemSet. I'll think about it more.

ktbriggs:
@amicic I see your point. And my bad for copying and pasting the wrong line into my original comment. I posted it while I was updating evacuator code to incorporate this change.

PR #2460, opened 2018-07-24
Optimized prolog/epilog for private linkage on X86

0dvictor:
Now the prolog/epilog generated by private linkage only allocates/deallocates stack once.


cathyzhyi:
The new code combines padding size with frame size so that only one sub/add instead of 2 is needed in both prolog/epilog. LGTM.

fjeremic:
Approved based on @cathyzhyi's review.

Jenkins test sanity

0dvictor:
Windows failures due to #2129

PR #2463, opened 2018-07-24
Revert "Refactor VTable index/offset variable name"

gacholio:
Reverts eclipse/openj9#2436


pshipton:
What was the problem which required the revert?

gacholio:
Failures observed on ZOS, the revert is just to test whether it might be the cause.  If it isn't I'll re-add.

fjeremic:
Couldn't this have been done in your own personal branch? IMO this is completely unnecessary.

Issue #2437, opened 2018-07-20
Rework getJimModules()


pdbain-ibm:
getJimModules() loads and caches `java/lang/reflect/Module`.  This is currently used only by JVMTI.
Per 
https://github.com/eclipse/openj9/pull/2420#issuecomment-406409243, 
 suggest using 
`enterVMFromJNI
J9Class *clazz = J9VM_J9CLASS_FROM_JCLASS(vmThread, jclassForModule);
jimModule = &clazz->classObject;
exitVMToJNI
`

Also create a macro `JCLASS_FROM_J9CLASS ` to do `(jclass)&someJ9Class->classObject`

pdbain-ibm:
Also, 
https://github.com/eclipse/openj9/pull/2420#pullrequestreview-138839583

pdbain-ibm:
runtime/vm/classloadersearch.c
+loadModule(J9JavaVM *vm, const char *moduleName, jobject *modulePtr)
+{
+	UDATA rc = CLS_ERROR_NONE;
+	J9VMThread *currentThread = currentVMThread(vm);
  @keithc-ca
keithc-ca 8 minutes ago  Contributor
Where this function is used, we've already retrieved the current thread: it would make sense to avoid redoing that by passing currentThread or env instead of vm.

pdbain-ibm:
runtime/vm/classloadersearch.c
+	vmModule = (*env)->CallStaticObjectMethod(env, jimModules, loadModule, moduleNameString);
+	if ((*env)->ExceptionOccurred(env)) {
+		rc = CLS_ERROR_INTERNAL;
+	} else if (NULL != modulePtr) {
  @keithc-ca
keithc-ca 9 minutes ago  Contributor
I like GAC's suggestion to move this after done and remove the *modulePtr = NULL; above.

gacholio:
Not defining the macro I suggested, since the "ref" created doesn't keep the class alive, so the macro would be ripe for accidental misuse.

gacholio:
Changes are here:

https://github.com/gacholio/openj9/tree/rework

I won't open a PR for them until @pdbain-ibm has tested them against the failure he was addressing.

pdbain-ibm:
@gacholio do you have a Linux x86 build in hand?  If not, I will build from your repo.

gacholio:
I'll run a test build to get the SDK.

pdbain-ibm:
[...]

PR #2465, opened 2018-07-24
Rework getJimModules()

gacholio:
- inline single-use function
- handle errors consistently
- load instrument module before adding to classpath
- remove unnecessary use of VM constant pool

Fixes: #2437

### DIFF
--- a/runtime/vm/classloadersearch.c
+++ b/runtime/vm/classloadersearch.c
@@ -34,8 +34,6 @@
 
 static UDATA addToSystemProperty(J9JavaVM * vm, const char * propertyName, const char * segment);
 static UDATA addZipToLoader(J9JavaVM * vm, const char * filename, J9ClassLoader * classLoader, BOOLEAN enforceJarRestriction);
-#define INSTRUMENT_MODULE "java.instrument"
-static UDATA loadModule(J9JavaVM *vm, const char *moduleName, jobject *modulePtr);
 
 
 /**
@@ -166,9 +164,6 @@ addToSystemProperty(J9JavaVM * vm, const char * propertyName, const char * segme
 static UDATA
 addZipToLoader(J9JavaVM * vm, const char * filename, J9ClassLoader * classLoader, BOOLEAN enforceJarRestriction)
 {
-	I_32 zipResult;
-	J9ZipFile zipFile;
-	J9VMThread * currentThread;
 	UDATA rc = CLS_ERROR_NONE;
 
 	/* Handle JCLs which don't have class loaders */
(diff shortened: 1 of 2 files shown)


runtime/vm/classloadersearch.c
```diff
@@ -245,77 +271,26 @@ addZipToLoader(J9JavaVM * vm, const char * filename, J9ClassLoader * classLoader
...
-			currentThread->privateFlags &= ~J9_PRIVATE_FLAGS_REPORT_EXCEPTION_THROW;
+			/* no need for mutual exclusion since repeating this is harmless */
+			J9Class *j9JimModules = NULL;
+			internalEnterVMFromJNI(currentThread);
+			j9JimModules = J9VM_J9CLASS_FROM_JCLASS(currentThread, modulesClass);
+			/* The address of j9class->classObject is considered to be a JNI global reference,
+			 * but it won't keep the class alive.  That's OK in this case since the class in
+			 * question is not a candidate for unloading.
+			 */
+			vm->jimModules = jimModules = (jclass)&j9JimModules->classObject;
```
DanHeidinga: This is iffy as it assumes the J9Class will never be freed (today's behaviour).  We've been talking about freeing the obsolete classes and this cause a hard to find bug.

DanHeidinga: Ah, this is just refactored existing code... something to be cleaned up, not necessarily as part of this PR.  If you're not going to fix this now, let me know and I'll open an issue for it.

runtime/vm/classloadersearch.c
```diff
@@ -228,14 +256,12 @@ addZipToLoader(J9JavaVM * vm, const char * filename, J9ClassLoader * classLoader
...
-				rc = loadModule(vm, INSTRUMENT_MODULE, NULL);
 			}
-	cleanup:
+cleanup:
 			(*env)->ExceptionClear(env);
-			(*env)->DeleteLocalRef(env, classLoaderRef);
-			(*env)->DeleteLocalRef(env, filenameString);
 			(*env)->DeleteLocalRef(env, classLoaderClass);
+			(*env)->DeleteLocalRef(env, filenameString);
+			(*env)->DeleteLocalRef(env, classLoaderRef);
```
keithc-ca: For my own education: why were these calls to `DeleteLocalRef` reordered?

gacholio: I reordered them to be in reverse order of creation, in case that were to ever matter for future optimization of the local ref space (I don't think it does at the moment).

runtime/vm/classloadersearch.c
gacholio: It's iffy for a class that we don't know everything about - in this case, it's a boot class (or whatever the JDK9 equivalent is), so we are sure it's not a candidate for unloading.

DanHeidinga: Note, in JDK9+ not every `java.*` class is loaded by the bootloader.  Some have been moved to the app loader.  (This one is loaded by the bootloader)

My concern is more with future bugs when we allow redefined J9Classes to be GC'd.... though this may not be a concern with FastHCR due to keeping the J9Class stable.

All that to say, we should at minimum open a issue to remember to look at this again,

gacholio: So, I don't want an issue for it - this was a change we made to avoid the need to lock and allocate a global ref.

DanHeidinga: Why are we trying to avoid creating the global ref?  Is there a usecase that demonstrates the current solution is better?  I would expect the extra cost of the global ref to be in the noise.

gacholio: It was to avoid the need for a single-use monitor.  Since you're clearly upset by this, I'll change it back (though I'll reuse an existing monitor).

DanHeidinga:

DanHeidinga:

PR #2466, opened 2018-07-24
Handle SIGBUS properly when executing in PPC helpers

klangman:
The JIT should be converting a SIGBUS into an InternalError
exception when the signal occurs while executing java code
but when executing an a JIT helper (like arraycopy) the
code did not detect that as running in java code and
would fail to convert to an exception resulting in a
javacore/systemdump to be generated instead.

### DIFF
--- a/runtime/compiler/runtime/SignalHandler.c
+++ b/runtime/compiler/runtime/SignalHandler.c
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2000, 2017 IBM Corp. and others
+ * Copyright (c) 2000, 2018 IBM Corp. and others
  *
  * This program and the accompanying materials are made available under
  * the terms of the Eclipse Public License 2.0 which accompanies this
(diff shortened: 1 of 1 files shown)


zl-wang:
do we need consistent behavior across platforms?  i.e. this is not specific to POWER.

runtime/compiler/runtime/SignalHandler.c
```diff
@@ -507,6 +507,26 @@ UDATA jitPPCHandler(J9VMThread* vmThread, U_32 sigType, void* sigInfo)
...
+		   lrPtr = (UDATA *) infoValue;
+		   exceptionTable = jitConfig->jitGetExceptionTableFromPC(vmThread, *lrPtr);
+		   if (exceptionTable) {
+            vmThread->jitException = (J9Object *) (*lrPtr);  /* the lr points at the instruction after the helper call */
+#if (defined(LINUXPPC64) && !defined(__LITTLE_ENDIAN__)) || defined(AIXPPC)
+            *iarPtr = *(UDATA*) ((void *) &jitHandleInternalErrorTrap);  /* get routine address from TOC entry */
+#else
+            *iarPtr = (UDATA) ((void *) &jitHandleInternalErrorTrap);
+#endif
+            return J9PORT_SIG_EXCEPTION_CONTINUE_EXECUTION;
```
zl-wang: Indentation is off for the few lines above

klangman:
This issue would occur on any platform where a helper is used to access the backing storage of a DirectByteBuffer. We have an example of this on PPC, but given how arraycopy is implemented on X86 it's less likely that a helper will be used (they do more inline arraycopy work then POWER does). Not sure about Z, but they have an instruction to copy memory so it's likely to be inline as well in most cases I suspect. We might want to get opinions from the X&Z teams on this matter. @fjeremic @andrewcraik

runtime/compiler/runtime/SignalHandler.c
klangman: I just noticed that as well. This file is using tabs rather then spaces for indenting. My editor was following that conversion until the preprocessor statements where it had no tab convention to follow from the previous line. I'll fix it to use tabs for these lines as well.

fjeremic:
Do we have a unit test which reproduces the issue? This will aid in testing all other platforms to make sure we're conformant. We should also add the said unit test to our test bucket to prevent issues in this area in the future.

klangman:
There is a test-case but it's rather difficult to use since you need to engineer a disk-full situation to see the problem occur.

fjeremic:
> There is a test-case but it's rather difficult to use since you need to engineer a disk-full situation to see the problem occur.

Would it be possible to provide instructions on how to run the testcase? I can try at least manually to reproduce on Z and see the effect.

ymanton:
> Would it be possible to provide instructions on how to run the testcase? I can try at least manually to reproduce on Z and see the effect.

If you create a small ramdisk you can fill it up trivially and probably reproduce the problem. Can't recall if it was AIX, Linux, or both, but that's how the original fix for Power was tested by @AlenBadel IIRC.

klangman:
@zl-wang Can we merge this change now.. @fjeremic is tracking an issue to investigate if the same issue can occur on Z and @andrewcraik is confident that the issue does not effect X86.

zl-wang:
ok then.

DanHeidinga:
@zl-wang please run Jenkins based builds to validate changes prior to merging PRs.  This PR has only been built on x86 using travis.ci.

The syntax to launch a JDK8 plinux build would be: `Jenkins test sanity plinux jdk8`

zl-wang:
Forgot the passed check was only x86 ...

zl-wang:

PR #2473, opened 2018-07-25
Include j9cfg.h for JAVA_SPEC_VERSION

keithc-ca:
j2sever.h relied on j9cfg.h being included before it: today that is true, but we'll avoid possible future subtle problems by removing that assumption.


pshipton:
jenkins test sanity win jdk8

pshipton:
> We really ought to fix that: NPT ERROR: Cannot open library!

https://github.com/eclipse/openj9/pull/2478

PR #2474, opened 2018-07-25
Add feature to ignore spin thread bound

babsingh:
Previously, setting lib->maxSpinThread to 0 was not allowed.

Now onwards, if lib->maxSpinThread is set to 0, then an upper bound
won't be enforced on spin threads, and unlimited spin threads will be
allowed.

Command line option, -Xthr:maxSpinThreads=<number_of_spin_threads> has
been modified to accept 0 as a valid value. Setting
-Xthr:maxSpinThreads=0 will allow unlimited spin threads.

This is an experimental feature, which is disabled by default. It will
be used to create heuristics for locking workloads which benefit from no
bound on the number of spin threads.


babsingh:
fyi - @pshipton @DanHeidinga 

Previously, this change was part of #2450. During benchmarking, this change showed more stable throughput with less fluctuations between consecutive runs for certain Liberty workloads. When disabled, this change doesn't impact spinning behavior since the check is performed outside the spinlock loop.

babsingh:
Dependent OMR changes (https://github.com/eclipse/omr/pull/2794) have been merged.

babsingh:
@gacholio can you please review these changes?

babsingh:
I have rebased and restarted the CI builds. eclipse/omr#2794 shows in https://github.com/eclipse/openj9-omr/commits/openj9. So, the CI builds should pass now.

gacholio:

PR #2478, opened 2018-07-25
Remove path from sun.boot.library.path

theresa-m:
Remove compressedrefs directory from sun.boot.library.path. Our extensions expect this property to have only one path. See full description in #2129.

Fixes: #2129

### DIFF
--- a/runtime/j9vm/jvm.c
+++ b/runtime/j9vm/jvm.c
@@ -2140,10 +2140,8 @@ jint JNICALL JNI_CreateJavaVM(JavaVM **pvm, void **penv, void *vm_args) {
 				/* Add the default options file */
 				(0 != addOptionsDefaultFile(&j9portLibrary, &vmArgumentsList, optionsDefaultFileLocation, localVerboseLevel))
 				|| (0 != addXjcl(&j9portLibrary, &vmArgumentsList, j2seVersion))
-				|| (0 != addBootLibraryPath(&j9portLibrary, &vmArgumentsList, "-Dcom.ibm.oti.vm.bootstrap.library.path=",
-						jvmBufferData(j9binBuffer), jvmBufferData(jrebinBuffer)))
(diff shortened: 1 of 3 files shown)


theresa-m:
copyright fixed and extra space removed

pshipton:
As far as I can tell, OpenJ9 doesn't use sun.boot.library.path for anything (except some tests), it uses com.ibm.oti.vm.bootstrap.library.path instead. The OpenJDK code uses it, but having only jre/bin seems necessary for the extensions code to work properly.

However, I'm reluctant to make this change for IBM Java 8 because we don't know how any application might be using the system property. I think the change should be re-worked so we can more easily keep the original behavior for the IBM builds, while changing OpenJ9.

@DanHeidinga

pshipton:
Exactly where in the extensions code causes the truncation that results in the NPT ERROR?

theresa-m:
The failure occurs during this check https://github.com/ibmruntimes/openj9-openjdk-jdk8/blob/openj9/jdk/src/windows/back/linker_md.c#L111

where `holderlen` can be traced back to Windows MAX_PATH, `pnamelen` is sun.boot.library.path, and `fname` is the name of the dll passed in here https://github.com/ibmruntimes/openj9-openjdk-jdk8/blob/openj9/jdk/src/share/back/debugInit.c#L275

pshipton:
Can we fix this in the extensions code instead? dbgsysBuildLibName() which calls dll_build_name() supports a path, so it seems wrong to limit the output buffer to MAXPATHLEN when the input buffer is longer. It could use the statically allocated buffer if its big enough, but allocate a new one if required.

pshipton:
Actually its not that a bigger buffer is required, as dll_build_name() only uses the buffer for one path at a time. The check of the buffer length against the strlen(pname) in dbgsysBuildLibName() is just wrong.

pshipton:
Once we settle on a fix to the extentions code and deliver it, we'll see about contributing it back to OpenJDK.

theresa-m:
closing this in favor of a change in the extensions repo. just running a personal build for it..

Issue #2267, opened 2018-06-26
Remove APIs Thread.destroy(), stop(Throwable) from Java 11


pshipton:
The deprecated APIs Thread.destroy() and Thread.stop(Throwable) are removed from Java 11.

Java 11 spec: http://cr.openjdk.java.net/~iris/se/11/latestSpec/
Javadoc: http://cr.openjdk.java.net/~iris/se/11/build/latest/api/

pshipton:
@tajila is somebody available to work on this?

tajila:
yes @Ailloviee can work on this

PR #2482, opened 2018-07-25
Remove Thread.destroy() and Thread.stop(Throwable) from JAVA11

Ailloviee:
Removed `destroy()` and `stop(Throwable)` from JAVA11 in `Thread.java`

Closes: #2267

### DIFF
--- a/jcl/src/java.base/share/classes/java/lang/Thread.java
+++ b/jcl/src/java.base/share/classes/java/lang/Thread.java
@@ -538,6 +538,7 @@ public int countStackFrames() {
  */
 public static native Thread currentThread();
 
+/*[IF !Java11]*/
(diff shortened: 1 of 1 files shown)


jcl/src/java.base/share/classes/java/lang/Thread.java
```diff
@@ -548,10 +548,12 @@ public int countStackFrames() {
 /*[ELSE]*/
 @Deprecated
 /*[ENDIF]*/
+/*[IF !Java11]*/
```
pshipton: The above ` @Deprecated` statement and javadoc also needs to be removed for Java 11. Similarly for stop().

pshipton:

PR #2483, opened 2018-07-25
Fix various script issues with Windows JDK11

ChengJin01:
The changes are to resolve all script issues
detected when compiling JDK11 on Windows.

### DIFF
--- a/buildenv/Build_Instructions_V11.md
+++ b/buildenv/Build_Instructions_V11.md
@@ -1,0 +1,457 @@
+<!--
+Copyright (c) 2018, 2018 IBM Corp. and others
+
+This program and the accompanying materials are made available under
+the terms of the Eclipse Public License 2.0 which accompanies this
+distribution and is available at https://www.eclipse.org/legal/epl-2.0/
+or the Apache License, Version 2.0 which accompanies this distribution and
+is available at https://www.apache.org/licenses/LICENSE-2.0.
+
+This Source Code may also be made available under the following
+Secondary Licenses when the conditions for such availability set
+forth in the Eclipse Public License, v. 2.0 are satisfied: GNU
+General Public License, version 2 with the GNU Classpath
+Exception [1] and GNU General Public License, version 2 with the
+OpenJDK Assembly Exception [2].
+
+[1] https://www.gnu.org/software/classpath/license.html
+[2] http://openjdk.java.net/legal/assembly-exception.html
+
+SPDX-License-Identifier: EPL-2.0 OR Apache-2.0 OR GPL-2.0 WITH Classpath-exception-2.0 OR LicenseRef-GPL-2.0 WITH Assembly-exception
+-->
+
+Building OpenJDK Version 11 with OpenJ9
+======================================
+
+Our website describes a simple [build process](http://www.eclipse.org/openj9/oj9_build.html)
+that uses Docker and Dockerfiles to create a build environment that contains everything
+you need to easily build a Linux binary of **OpenJDK V10** with the Eclipse OpenJ9 virtual machine.
+A more complete set of build instructions are included here for multiple platforms:
+
+- [Linux :penguin:](#linux)
+- [AIX :blue_book:](#aix)
+- [Windows :ledger:](#windows)
+- [MacOS :apple:](#macos)
+- [ARM :iphone:](#arm)
+
+----------------------------------
+
+## Linux
+:penguin:
+This build process provides detailed instructions for building a Linux x86-64 binary of **OpenJDK V11** with OpenJ9 on Ubuntu 16.04. The binary can be built directly on your system, in a virtual
+machine, or in a Docker container :whale:.
+
+If you are using a different Linux distribution, you might have to review the list of libraries that are bundled with your distribution and/or modify the instructions to use equivalent commands to the Advanced Packaging Tool (APT). For example, for Centos, substitute the `apt-get` command with `yum`.
+
+If you want to build a binary for Linux on a different architecture, such as Power Systems&trade; or z Systems&trade;, the process is very similar and any additional information for those architectures are included as Notes :pencil: as we go along.
+
+### 1. Prepare your system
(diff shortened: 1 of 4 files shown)


ChengJin01:
The changes mainly addressed the following issues:
0)buildenv/Build_Instructions_V11.md: update the windows section for the compilation with VS2017
1)runtime/compiler/build/toolcfg/win-msvc/common.mk: specify libraries used since VS2015.
  Note: msvcrt was split to msvcrt, ucrt and vcruntime since VS2015.
  Please refer to https://blogs.msdn.microsoft.com/vcblog/2015/03/03/introducing-the-universal-crt/ for details
2)runtime/ddr/blacklist: temporarily exclude these types
3)runtime/ddr/module.xml: fixed the issue in generating jvmtitest.res via omrddrgen
4)runtime/makelib/mkconstants.mk.ftl: export the bin path to MINGW64 if compiled with VS2017

Reviewer: @keithc-ca , @pshipton
FYI: @DanHeidinga

buildenv/Build_Instructions_V11.md
```diff
@@ -0,0 +1,457 @@
...
+
+### 1. Prepare your system
+:ledger:
+You must install a number of software dependencies to create a suitable build environment on your system:
+
+- [Cygwin](https://cygwin.com/install.html), which provides a Unix-style command line interface. Install all packages in the `Devel` category. Included in this package is mingw, a minimalist subset of GNU tools for Microsoft Windows. OpenJ9 uses the mingw/GCC compiler and is tested with version 6.4.0 of same. Other versions are expected to work also. In the `Archive` category, install the packages `zip` and `unzip`. In the `Utils` category, install the `cpio` package. Install any further package dependencies that are identified by the installer. More information about using Cygwin can be found [here](https://cygwin.com/docs.html).
+- [Windows JDK 10](https://adoptopenjdk.net/releases.html?variant=openjdk10#x64_win), which is used as the boot JDK.
+- [Microsoft Visual Studio 2013]( https://go.microsoft.com/fwlink/?LinkId=532495), which is the same compiler level used by OpenJDK; or [Microsoft Visual Studio 2017]( https://visualstudio.microsoft.com/thank-you-downloading-visual-studio/?sku=Community&rel=15), which is now supported.
```
pshipton: Does 2013 work with Java 11?

buildenv/Build_Instructions_V11.md
```diff
@@ -0,0 +1,457 @@
...
+- [Freemarker V2.3.8](https://sourceforge.net/projects/freemarker/files/freemarker/2.3.8/freemarker-2.3.8.tar.gz/download)
+- [mingw-w64](https://sourceforge.net/projects/mingw-w64/files/Toolchains%20targetting%20Win32/Personal%20Builds/mingw-builds/installer/mingw-w64-install.exe)
+
+**Note**
+1) mingw-w64 must be installed if compiled with VS2017 in that the current version of cygwin has not yet integrated the latest changes of mingw-w64 for the moment. In addition, set up the environment variable MINGW64_BIN_PATH to the binary path of mingw-w64. e.g.
```
pshipton: compiled -> compiling

I assume mingw-w64 contains executables that conflict with cygwin?
MINGW64_BIN_PATH is optional? i.e. if mingw-w64 is before cygwin on the path then setting MINGW64_BIN_PATH is not required?

buildenv/Build_Instructions_V11.md
```diff
@@ -0,0 +1,457 @@
...
+```
+export MINGW64_BIN_PATH="C:\mingw-w64\x86_64-8.1.0-win32-seh-rt_v6-rev0\mingw64\bin\"
+```
+2) FreeType2 is not required on Windows Platform (Please refer to https://github.com/ibmruntimes/openj9-openjdk-jdk11/blob/openj9/doc/building.md for details)
```
pshipton: I don't think we need to mention something is not a requirement.

buildenv/Build_Instructions_V11.md
pshipton: It seems 2017 is the default for Java 11? If so "which is the same compiler level used by OpenJDK" isn't appropriate.

ChengJin01: According to openj9-openjdk-jdk11_v1/make/autoconf/toolchain_windows.m4:
```
# The order of these defines the priority by which we try to find them.
VALID_VS_VERSIONS="2017 2013 2015 2012 2010"
...
```
If --with-toolchain-version is not supplied on the command line, it searches the VS version from 2017 by order in VALID_VS_VERSIONS by default,  depending on which VS version is installed on Windows. So it looks for VS2013 if VS2017 doesn't exist.

>It seems 2017 is the default for Java 11
Pretty much yes in such case. I already updated the description there.

buildenv/Build_Instructions_V11.md
ChengJin01: Already removed.

buildenv/Build_Instructions_V11.md
ChengJin01: >compiled -> compiling
Updated.

ChengJin01: > I assume mingw-w64 contains executables that conflict with cygwin? ...
The reason will we choose the latest version of mingw-w64 is we need the new version of x86_64-w64-mingw32-g++ (which is outdated in cygwin) to compile Interpreter related files (including BytecodeInterpreter, MHInterpreter, etc)

There is no need to set up path to Cygwin at all because it knows where to run x86_64-w64-mingw32-g++ in Cygwin, which means there is no way to set up mingw-w64 is before cygwin

ChengJin01: Already recompiled and it works as expected (also no problem with java -version).
So updated the document and removed MINGW64_BIN_PATH in runtime\makelib\mkconstants.mk.ftl.

pshipton:
jenkins compile win

Issue #2047, opened 2018-06-01
Intermittent NullPointerException in TestAttachErrorHandling.deleteIpcDirectories


pdbain-ibm:
Low priority test issue.

```
FAILED: test_startupShutdownRobustness
java.lang.NullPointerException
	at java.io.File.<init>(File.java:371)
	at org.openj9.test.attachAPI.TestAttachErrorHandling.deleteIpcDirectories(TestAttachErrorHandling.java:129)
	at org.openj9.test.attachAPI.TestAttachErrorHandling.test_startupShutdownRobustness(TestAttachErrorHandling.java:113)
```
```
	private void deleteIpcDirectories(File ipcDir,
			ArrayList<TargetManager> targets) {
		for (TargetManager victim: targets) {
			String tgtId = victim.getTargetPid();
>>>>>            File targetDir = new File(ipcDir, victim.targetId);
			if (targetDir.exists()) {
				logger.debug("delete "+tgtId);
				deleteRecursive(targetDir);
			}
		}
	}
```
getTargetPid() should not return null, and ipcDir is used on line 112.

pdbain-ibm:
from https://ci.eclipse.org/openj9/job/PullRequest-Extended-JDK10-aix_ppc-64_cmprssptrs-OpenJ9/18/testReport/junit/org.openj9.test.attachAPI/TestAttachErrorHandling/test_startupShutdownRobustness/
```
java.io.IOException: Error deleting /tmp/test_startupShutdownRobustness/18153680
at org.openj9.test.util.FileUtilities.deleteRecursive(FileUtilities.java:51)
at org.openj9.test.attachAPI.TestAttachErrorHandling.deleteIpcDirectories(TestAttachErrorHandling.java:134)
at org.openj9.test.attachAPI.TestAttachErrorHandling.test_startupShutdownRobustness(TestAttachErrorHandling.java:120)
```

pdbain-ibm:
Excluding the test temporarily: https://github.com/eclipse/openj9/pull/2173

PR #2485, opened 2018-07-26
Improve attach API shutdown robustness

pdbain-ibm:
Increase semaphore count when retrying wait loop termination,
to account for possible missing target directories.
This is tested by TestAttachErrorHandling.test_startupShutdownRobustness().

I ran the test 100 times.  There were a few failures, all due to 
https://github.com/eclipse/openj9/issues/2047  which is a separate issue.

### DIFF
--- a/jcl/src/java.base/share/classes/com/ibm/tools/attach/target/AttachHandler.java
+++ b/jcl/src/java.base/share/classes/com/ibm/tools/attach/target/AttachHandler.java
@@ -402,17 +402,17 @@ public class AttachHandler extends Thread {
 		}
 		FileLock.shutDown();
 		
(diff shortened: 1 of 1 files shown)


pshipton:
Is test_startupShutdownRobustness an excluded test? Ideally it should be included as part of this PR, which likely means fixing #2047 at the same time.

pshipton:
jenkins test all xlinux jdk8

pdbain-ibm:
The test is excluded in the OpenJ9 test suite, but is still running in the internal builds.
I could fix #2047 et al. first in a separate pull request (since they are a different issue from this PR),
then re-submit this commit with the test re-enabled.

PR #2491, opened 2018-07-26
Optimize findNativeMethodFrame

gacholio:
findNativeMethodFrame had two identical implementations, and is so
trivial that it need not be a function call.  Put it in VMHelpers.hpp
instead.


gacholio:

gacholio:
Common code, one build should be enough.

PR #2494, opened 2018-07-26
Optimize j/l/Class.isAssignableFrom on X86

0dvictor:
java/lang/Class.isAssignableFrom() is transformed into a JIT Helper
so that JNI call overhead is avoided.

### DIFF
--- a/runtime/codert_vm/cnathelp.cpp
+++ b/runtime/codert_vm/cnathelp.cpp
@@ -2518,6 +2518,18 @@ fast_jitInstanceOf(J9VMThread *currentThread, J9Class *castClass, j9object_t obj
 }
 
 UDATA J9FASTCALL
+#if defined(J9VM_ARCH_X86) || defined(J9VM_ARCH_S390)
+/* TODO Will be cleaned once all platforms adopt the correct parameter order */
+fast_jitCheckAssignable(J9VMThread *currentThread, J9Class *clazz, J9Class *castClass)
+#else /* J9VM_ARCH_X86 || J9VM_ARCH_S390*/
+fast_jitCheckAssignable(J9VMThread *currentThread, J9Class *castClass, J9Class *clazz)
+#endif /* J9VM_ARCH_X86 || J9VM_ARCH_S390*/
+{
+	JIT_HELPER_PROLOGUE();
+	return VM_VMHelpers::inlineCheckCast(clazz, castClass) ? 1 : 0;
+}
+
+UDATA J9FASTCALL
 fast_jitObjectHashCode(J9VMThread *currentThread, j9object_t object)
 {
 	JIT_HELPER_PROLOGUE();
--- a/runtime/compiler/x/codegen/J9CodeGenerator.cpp
+++ b/runtime/compiler/x/codegen/J9CodeGenerator.cpp
@@ -396,6 +396,13 @@ J9::X86::CodeGenerator::suppressInliningOfRecognizedMethod(TR::RecognizedMethod 
       }
    }
 
+bool
+J9::X86::CodeGenerator::supportsInliningOfIsAssignableFrom()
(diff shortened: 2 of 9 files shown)


0dvictor:
This optimization in theory applies to all platforms; however, some platforms may need extra work in their CodeGen to take advantage of it. For example, Z has its own optimizations in its CodeGen to fast-path certain cases. Should this change be applied to Z blindly, its optimizations would be bypassed and hence the performance may be negatively impacted.
@r30shah FYI.

runtime/compiler/x/codegen/J9CodeGenerator.cpp
```diff
@@ -396,6 +396,13 @@ J9::X86::CodeGenerator::suppressInliningOfRecognizedMethod(TR::RecognizedMethod
       }
    }
 
+bool
+J9::X86::CodeGenerator::supportsInliningOfIsAssignableFrom()
+   {
+   static auto disableInliningOfIsAssignableFrom = (bool)feGetEnv("TR_DisableInliningOfIsAssignableFrom");
```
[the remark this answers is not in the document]
0dvictor: I see your point to make LHS `const`, the code has been updated.

cathyzhyi:
LGTM

runtime/oti/jitprotos.h
```diff
@@ -163,8 +163,10 @@ void J9FASTCALL fast_jitCollapseJNIReferenceFrame(J9VMThread *currentThread);
 #if defined(J9VM_ARCH_X86) || defined(J9VM_ARCH_S390)
 /* TODO Will be cleaned once all platforms adopt the correct parameter order */
 UDATA J9FASTCALL fast_jitInstanceOf(J9VMThread *currentThread, j9object_t object, J9Class *castClass);
+UDATA J9FASTCALL fast_jitCheckAssignable(J9VMThread *currentThread, J9Class *clazz, J9Class *castClass);
 #else /* J9VM_ARCH_X86 || J9VM_ARCH_S390*/
 UDATA J9FASTCALL fast_jitInstanceOf(J9VMThread *currentThread, J9Class *castClass, j9object_t object);
+UDATA J9FASTCALL fast_jitCheckAssignable(J9VMThread *currentThread, J9Class *castClass, J9Class *clazz)
```
fjeremic: I think we may be missing a semicolon here?

0dvictor: Good catch! It is fixed.

runtime/codert_vm/cnathelp.cpp
```diff
@@ -2517,6 +2517,18 @@ fast_jitInstanceOf(J9VMThread *currentThread, J9Class *castClass, j9object_t obj
 	return isInstance;
 }
 
+UDATA J9FASTCALL
+#if defined(J9VM_ARCH_X86) || defined(J9VM_ARCH_S390)
+/* TODO Will be cleaned once all platforms adopt the correct parameter order */
+fast_jitCheckAssignable(J9VMThread *currentThread, J9Class *clazz, J9Class *castClass)
+#else /* J9VM_ARCH_X86 || J9VM_ARCH_S390*/
+fast_jitCheckAssignable(J9VMThread *currentThread, J9Class *castClass, J9Class *clazz)
+#endif /* J9VM_ARCH_X86 || J9VM_ARCH_S390*/
```
fjeremic: Are there some braces missing around this area? Not really sure what is going on but looking at the surrounding code this appears to be incorrect at first glance. Perhaps I'm missing something.

0dvictor: I don't think extra braces are need here. It looks very similar to the method above `fast_jitInstanceOf` and compiles.

0dvictor:
Windows failure due to #2129

PR #2497, opened 2018-07-27
Fix Runtime Instrumentation Control Block Mapping

dhong44:
Remove mapping to suspension control at word 6 byte 3 bit 4
which should be reserved and add mappings for suppression
control and guarded storage.

### DIFF
--- a/runtime/port/common/j9ri.h
+++ b/runtime/port/common/j9ri.h
@@ -36,7 +36,7 @@ typedef struct riControlBlock {
    uint64_t limitAddr;
 
    /* Word 6 - Flags */
-   uint32_t pad29:29;
+   uint32_t padAtWord6Bit0:29;
    uint32_t reportingGroupSize:3;
 
    /* Word 7 */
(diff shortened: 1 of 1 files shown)


runtime/port/common/j9ri.h
```diff
@@ -60,14 +59,18 @@ typedef struct riControlBlock {
    uint32_t branchOnCondAsOtherTypeBrranches:1;
    uint32_t RIEmit:1;
    uint32_t TXAbort:1;
-   uint32_t pad3:2;
+
+   uint32_t guardedStorage:1;
+   uint32_t pad3:1;
```
fjeremic: Could we use a better name for the paddings? It would be much easier to make sure the offsets are correct if for example `pad3` was renamed to be something like `padAtBit15`.

dhong44: Sure, with the first padding, `pad29`, is that name clear since it is a 29 bit padding at the start of word 6 or should we think of something more descriptive?

fjeremic: Let's make the naming consistent. It shouldn't matter whether the pad is at the start on in the middle. The name should be descriptive enough to tell you where the padding is. The bit field will always tell how big the padding is.

dhong44: 424f9aa

fjeremic:
Jenkins test sanity depends eclipse/omr#master

dhong44:
Looks like failures are due to #2129

fjeremic:
Travis builds will fail until we have an OMR promotion. We should be good here.

PR #2499, opened 2018-07-27
WIP: update jitResolveHelper method for invokePrivate

fengxue-IS:
- return the direct method address for private nestmates method
- add a low-bit flag to the method address
- check for low-bit flag during j2i translation


fengxue-IS:
FYI @jdmpapin

runtime/vm/BytecodeInterpreter.hpp
```diff
@@ -697,6 +697,12 @@ class INTERPRETER_CLASS
 	{
 		void* const jitReturnAddress = VM_JITInterface::peekJITReturnAddress(_currentThread, _sp);
 		UDATA jitVTableOffset = VM_JITInterface::jitVTableIndex(jitReturnAddress, interfaceVTableIndex);
+#if defined(J9VM_OPT_VALHALLA_NESTMATES)
+		if (J9_ARE_ANY_BITS_SET(jitVTableIndex, J9_VTABLE_INDEX_DIRECT_METHOD_FLAG)) {
```
gacholio: jitVTableIndex -> jitVTableOffset (and below).

fengxue-IS: fixed, thanks

runtime/vm/BytecodeInterpreter.hpp
```diff
@@ -697,6 +697,12 @@ class INTERPRETER_CLASS
 	{
 		void* const jitReturnAddress = VM_JITInterface::peekJITReturnAddress(_currentThread, _sp);
 		UDATA jitVTableOffset = VM_JITInterface::jitVTableIndex(jitReturnAddress, interfaceVTableIndex);
+#if defined(J9VM_OPT_VALHALLA_NESTMATES)
+		if (J9_ARE_ANY_BITS_SET(jitVTableOffset, J9_VTABLE_INDEX_DIRECT_METHOD_FLAG)) {
+			/* Nestmates: vtable index is really a J9Method to directly invoke */
+			return (J9Method*)(jitVTableOffset & ~J9_VTABLE_INDEX_DIRECT_METHOD_FLAG);
```
gacholio: Multiple returns are discouraged - please declare a variable.

runtime/jilgen/jilconsts.c
```diff
@@ -752,6 +752,9 @@ writeConstants(OMRPortLibrary *OMRPORTLIB, IDATA fd)
 			writeConstant(OMRPORTLIB, fd, "J9TR_ELSSize", sizeof(J9VMEntryLocalStorage)) |
 			writeConstant(OMRPORTLIB, fd, "J9TR_J9_EXTENDED_RUNTIME_DEBUG_MODE", J9_EXTENDED_RUNTIME_DEBUG_MODE) |
 			writeConstant(OMRPORTLIB, fd, "J9TR_J9_EXTENDED_RUNTIME_USE_VECTOR_REGISTERS", J9_EXTENDED_RUNTIME_USE_VECTOR_REGISTERS) |
+#if defined(J9VM_OPT_VALHALLA_NESTMATES)
+			writeConstant(OMRPORTLIB, fd, "J9TR_J9_VTABLE_INDEX_DIRECT_METHOD_FLAG", J9_VTABLE_INDEX_DIRECT_METHOD_FLAG) |
```
gacholio: The JIT may require this to always be defined.  Either that, or the nestmates flag will need an ASM_ equivalent so they can ifdef the PicBuilder code.

runtime/oti/j9consts.h
```diff
@@ -818,6 +818,10 @@ extern "C" {
 #define J9_CLASSLOADER_TYPE_BOOT		1
 #define J9_CLASSLOADER_TYPE_PLATFORM	2
 
+#if defined(J9VM_OPT_VALHALLA_NESTMATES)
+#define J9_VTABLE_INDEX_DIRECT_METHOD_FLAG 0x1
```
gacholio: See above re: always being defined.

runtime/jilgen/jilconsts.c
fengxue-IS: @jdmpapin can you give some insight on this?

runtime/vm/BytecodeInterpreter.hpp
fengxue-IS: fixed

runtime/jilgen/jilconsts.c
jdmpapin: This `writeConstant()` allows us to write `J9TR_J9_VTABLE_INDEX_DIRECT_METHOD_FLAG` symbolically instead of hard-coding 1 in the assembly code in PicBuilder. I think what @gacholio is saying is that there is currently no `J9VM_OPT_VALHALLA_NESTMATES` flag to do conditional assembly in PicBuilder, so either we need to add such a flag, or else `J9TR_J9_VTABLE_INDEX_DIRECT_METHOD_FLAG` has to be unconditionally defined.

runtime/vm/BytecodeInterpreter.hpp
```diff
@@ -695,11 +695,21 @@ class INTERPRETER_CLASS
 	VMINLINE J9Method*
 	j2iVirtualMethod(REGISTER_ARGS_LIST, j9object_t receiver, UDATA interfaceVTableIndex)
 	{
+		J9Method *method;
 		void* const jitReturnAddress = VM_JITInterface::peekJITReturnAddress(_currentThread, _sp);
 		UDATA jitVTableOffset = VM_JITInterface::jitVTableIndex(jitReturnAddress, interfaceVTableIndex);
-		UDATA vTableOffset = sizeof(J9Class) - jitVTableOffset;
-		J9Class *clazz = J9OBJECT_CLAZZ(_currentThread, receiver);
-		return *(J9Method**)((UDATA)clazz + vTableOffset);
+#if defined(J9VM_OPT_VALHALLA_NESTMATES)
```
gacholio: Now that I think about it, I believe this code is completely unncessary - there's no way for the JIT to reach the interpreter via a virtual call to a private method.  The JIT will necessarily have to statically dispatch in these cases.

runtime/jilgen/jilconsts.c
gacholio: Yes, that's exactly what I'm saying, so you guys can decide which path you prefer.

runtime/vm/BytecodeInterpreter.hpp
jdmpapin: This is part of the direct dispatch done by PicBuilder when `invokevirtual` was unresolved at compile time, and the callee is not compiled at runtime

gacholio:
I think this is ready to commit once the j2iVirtual change is removed.

gacholio:
Discussing with Devin, I think we probably will need the j2iVirtual change.

runtime/vm/BytecodeInterpreter.hpp
```diff
@@ -695,11 +695,21 @@ class INTERPRETER_CLASS
 	VMINLINE J9Method*
 	j2iVirtualMethod(REGISTER_ARGS_LIST, j9object_t receiver, UDATA interfaceVTableIndex)
 	{
+		J9Method *method;
```
gacholio: Needs to be initialized.

gacholio:
jenkins compile xlinux jdk11

PR #2500, opened 2018-07-27
Deprecate Access Register (AR) support in Z codegen

fjeremic:
The compiler currently has AR support on some minor spectrum of
usability, though in practice the feature has been turned off for
several years, and has never been used outside of niche areas.

Supporting ARs even when disabled (as is the default) does have
technical debt and compile time overhead as we have to query for them
all the time, and in fact we have a separate register allocation pass
just for ARs.

Given these registers were originally envisioned to be spill slots
however times have changed since their original inception:

1. We don't have as many spills due to codegen and RA improvements
2. Storing out the stack is quite cheap
3. There is a lot of weirdness of using ARs around calls (linkage)
4. No support on Linux on Z

With these in mind this commit purges the codebase of all (modulo any
places I may have missed) AR support.

### DIFF
--- a/runtime/compiler/z/codegen/J9S390CHelperLinkage.cpp
+++ b/runtime/compiler/z/codegen/J9S390CHelperLinkage.cpp
@@ -99,8 +99,7 @@ TR::S390CHelperLinkage::S390CHelperLinkage(TR::CodeGenerator * codeGen,TR_S390Li
       setRegisterFlag(TR::RealRegister::FPR15, Preserved);
       }
 #endif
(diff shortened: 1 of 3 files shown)


fjeremic:
Depends on https://github.com/eclipse/omr/pull/2803.

fjeremic:
eclipse/omr#2803 has been merged. We have a co-dependency here. @0xdaryl could you be the committer for this one as well?

Jenkins test sanity xlinux,zlinux jdk8 depends eclipse/omr#master

0xdaryl:
These are Z-specific changes.  No need to wait for x86 builds.  Merging.

PR #2503, opened 2018-07-29
Excluded Thread.destroy() and Thread.stop(Throwable) test cases in Java11

Ailloviee:
Excluded `destroy()` and `stop(Throwable)`  test cases from JAVA11 in `Thread.java`

[ci skip]

Related: https://github.com/eclipse/openj9/pull/2482

### DIFF
--- a/test/functional/Java8andUp/src/org/openj9/test/java/lang/Test_Thread.java
+++ b/test/functional/Java8andUp/src/org/openj9/test/java/lang/Test_Thread.java
@@ -370,19 +370,6 @@ public class Test_Thread {
 	}
 
 	/**
-	 * @tests java.lang.Thread#destroy()
-	 */
-	@Test
-	public void test_destroy() {
-		try {
-			new Thread().destroy();
-			Assert.fail("NoSuchMethodError expected");
-		} catch (NoSuchMethodError e) {
-			// expected
-		}
-	}
-
-	/**
 	 * @tests java.lang.Thread#enumerate(java.lang.Thread[])
 	 */
 	@Test
@@ -1218,85 +1205,6 @@ public class Test_Thread {
 	 * @tests java.lang.Thread#stop(java.lang.Throwable)
 	 */
 	@Test
-	public void test_stop3() {
-		/* [PR CMVC 94728] AccessControlException on dead Thread */
-		Thread t = new Thread("t");
-		class MySecurityManager extends SecurityManager {
-			public boolean intest = false;
-			public boolean checkAccess = false;
-
-			public void checkAccess(Thread t) {
-				if (intest) {
-					checkAccess = true;
-				}
-			}
-		}
-		MySecurityManager sm = new MySecurityManager();
-		System.setSecurityManager(sm);
-		try {
-			sm.intest = true;
-			try {
-				t.stop(new ThreadDeath());
-			} catch (UnsupportedOperationException e) {
-				// Expected
-			} catch (Throwable e) {
-				Assert.fail("unexpected exception: " + e);
-			}
-			sm.intest = false;
-			AssertJUnit.assertFalse("checkAccess 1", sm.checkAccess);
-			t.start();
-			try {
-				t.join(2000);
-			} catch (InterruptedException e) {
-			}
-			sm.intest = true;
-			sm.checkAccess = false;
-			try {
-				t.stop(new ThreadDeath());
-			} catch (UnsupportedOperationException e) {
-				// Expected
-			} catch (Throwable e) {
-				Assert.fail("unexpected exception: " + e);
-			}
-			AssertJUnit.assertFalse("checkAccess 2", sm.checkAccess);
-			sm.intest = false;
-		} finally {
-			System.setSecurityManager(null);
-		}
-	}
-
-
-	/**
-	 * @tests java.lang.Thread#stop(java.lang.Throwable)
-	 */
-	@Test
-	public void test_stop4() {
-		ResSupThread t = new ResSupThread(Thread.currentThread());
-		synchronized (t) {
-			st = new Thread(t, "StopThread");
-			st.setPriority(Thread.MAX_PRIORITY);
-			st.start();
-			try {
-				t.wait();
-			} catch (InterruptedException e) {
-			}
-		}
-		try {
-			st.stop(new BogusException("Bogus"));
-		} catch (UnsupportedOperationException e) {
-			AssertJUnit.assertTrue("Stopped child with exception not alive", st.isAlive());
-			st.interrupt();
-			return;
-		}
-		st.interrupt();
-		AssertJUnit.assertTrue("Stopped child did not throw exception", false);
-	}
-
-
-	/**
-	 * @tests java.lang.Thread#stop(java.lang.Throwable)
-	 */
-	@Test
 	public void test_stop5() {
 		class StopBeforeStartThread extends Thread {
 			public boolean failed = false;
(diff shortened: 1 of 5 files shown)


test/functional/Java8andUp/src_80/org/openj9/test/java/lang/Test_Thread_Java8to10.java
```diff
@@ -0,0 +1,278 @@
...
+import org.testng.annotations.AfterMethod;
+import org.testng.annotations.Test;
+import org.testng.Assert;
+import org.testng.AssertJUnit;
+import java.lang.ref.WeakReference;
+
+@Test(groups = { "level.sanity" })
+public class Test_Thread_Java8to10 {
+
+	static class SimpleThread implements Runnable {
```
pshipton: SimpleThread doesn't seem to be used.

test/functional/Java8andUp/src_80/org/openj9/test/java/lang/Test_Thread_Java8to10.java
```diff
@@ -0,0 +1,278 @@
...
+
+		}
+
+		public SimpleThread(int d) {
+			if (d >= 0)
+				delay = d;
+		}
+	}
+
+	static class YieldThread implements Runnable {
```
pshipton: YieldThread doesn't seem to be used.

test/functional/Java8andUp/src_80/org/openj9/test/java/lang/Test_Thread_Java8to10.java
```diff
@@ -0,0 +1,278 @@
...
+		} catch (NoSuchMethodError e) {
+			// expected
+		}
+	}
+
+	/**
+	 * @tests java.lang.Thread#stop(java.lang.Throwable)
+	 */
+	@Test
+	public void test_stop3() {
```
pshipton: I don't think we need this test, test_stop6() is sufficient. test_stop3() seems like a modified version of the test that used to test SecurityManager behavior but now doesn't really test anything.

test/functional/Java8andUp/src_80/org/openj9/test/java/lang/Test_Thread_Java8to10.java
```diff
@@ -0,0 +1,278 @@
...
+
+	static class BogusException extends Throwable {
+		public BogusException(String s) {
+			super(s);
+		}
+	}
+
+	Thread st, ct, spinner;
+
+	static boolean calledMySecurityManager = false;
```
pshipton: calledMySecurityManager doesn't seem to be used.

test/functional/Java8andUp/src_80/org/openj9/test/java/lang/Test_Thread_Java8to10.java
```diff
@@ -0,0 +1,278 @@
...
+
+	// used in test_defaultMemoryArea to pass status from thread started
+	// back to thread running test case
+	boolean pass = true;
```
pshipton: pass doesn't seem to be used.

test/functional/Java8andUp/src_80/org/openj9/test/java/lang/Test_Thread_Java8to10.java
```diff
@@ -0,0 +1,278 @@
...
+		} finally {
+			System.setSecurityManager(null);
+		}
+	}
+
+	/**
+	 * @tests java.lang.Thread#stop(java.lang.Throwable)
+	 */
+	@Test
+	public void test_stop4() {
```
pshipton: I don't think we need this test, test_stop6() is sufficient. test_stop4() seems like a modified version of the test that used to validate stop(Throwable) but now doesn't really test anything. Removing this test also obsoletes the ResSupThread class, the BogusException class, and the st variable.

test/functional/Java8andUp/src_80/org/openj9/test/java/lang/Test_Thread_Java8to10.java
```diff
@@ -0,0 +1,278 @@
...
+		}
+	}
+
+	static class BogusException extends Throwable {
+		public BogusException(String s) {
+			super(s);
+		}
+	}
+
+	Thread st, ct, spinner;
```
pshipton: I don't think ct or spinner are used, except in tearDown, and can be removed.

Ailloviee:
I took a different approach for this issue, instead of extracting the specific test methods (`stop(Throwable)` and `destroy()`), I duplicated `Test_Thread.java` and put it under `src_80` , `src_90` , and `src_110`, but in `src_110` the unwanted methods are removed. Through this way it is more straightforward and easier to maintain thread testing in the future in terms of version difference.

llxia:
This is a common question moving forward. Currently, we are facing two issues:
- In general, `src` is for common test material. As we have more and more releases, we will have less test material that is common for all releases. In the case here, we have common test material for only jdk8, jdk9 and jdk10, not jdk11.
- TestNG does not allow to reference any non-existing file in testng.xml. Say, we create a file Test_Thread_extra.java for stop(Throwable) and destroy() tests only. It is for testing jdk8, 9 and 10. It has to be also included into jdk11 compilation (an empty file with same file name). Otherwise, TestNG will complain it cannot find the class during compilation. (We could create a separate testng.xml for this test only, so dummy Test_Thread_extra.java does not needed in src_110. But we need to make changes accordingly in playlist.xml. It feels to be more work)

If our goal is to completely eliminate test duplication in this case, then compilation process (build.xml) will be more complicated. Keep common tests in Test_Thread.java in src and we could put Test_Thread_extra.java file in a "common place" (or just src_80) and include it separately during compiling jdk8, jdk9 and jdk10. For jdk11, we could add an empty Test_Thread_extra.java. Does this make sense?

fyi @smlambert

llxia:
@Ailloviee Could we try above approach for now? And please let me know if you need any help. Thanks.

test/functional/Java8andUp/src_110/org/openj9/test/java/lang/Test_Thread_Extra.java
```diff
@@ -0,0 +1,35 @@
+
+package org.openj9.test.java.lang;
+
+/*******************************************************************************
+ * Copyright (c) 1998, 2018 IBM Corp. and others
```
llxia: Please update the copyright year to 2018. This is a newly created file.

test/functional/Java8andUp/src_110/org/openj9/test/java/lang/Test_Thread_Extra.java
```diff
@@ -0,0 +1,35 @@
...
+ * SPDX-License-Identifier: EPL-2.0 OR Apache-2.0 OR GPL-2.0 WITH Classpath-exception-2.0 OR LicenseRef-GPL-2.0 WITH Assembly-exception
+ *******************************************************************************/
+
+import org.testng.annotations.AfterMethod;
+import org.testng.annotations.Test;
+import org.testng.Assert;
+import org.testng.AssertJUnit;
+import java.lang.ref.WeakReference;
+
+@Test(groups = { "level.sanity" })
```
llxia: Please remove imports and annotation.

test/functional/Java8andUp/src/org/openj9/test/java/lang/Test_Thread.java
```diff
@@ -369,19 +369,6 @@ public void test_currentThread() {
 				Thread.currentThread().getName().equals("main"));
```
llxia: Is there a need to change file permission? It is changed from 100755 → 100644

llxia:
Please test this in a grinder first. Thanks.

llxia:
jenkins test sanity plinux all

llxia:
We tested jdk9 in grinder.

PR #2510, opened 2018-07-30
Check RomClass is in SCC before storing IProf Data

dsouzai:
The following race is possible:

1. Thread 1 calls TR_IPBCDataCallGraph::canBePersisted on Entry A,
   which succeeds. However, at least one class in the list is NULL
2. Thread 2 calls TR_IPBCDataCallGraph::setData on Entry A, which
   also succeeds, setting one of the classes that was previous NULL
   to something non-NULL.
3. Thread 1 calls TR_IPBCDataCallGraph::createPersistentCopy.
   Normally if a class in the list is NULL, the value stored is also
   NULL. However, because Thread 2 updated some previously NULL class,
   Thread 1 will compute the difference of
   (ramClass->romClass - startOfSCC) and potentially get a value
   that's bigger than the SCC.

This commit checks for this race before persisting IProfiler data.

### DIFF
--- a/runtime/compiler/runtime/IProfiler.cpp
+++ b/runtime/compiler/runtime/IProfiler.cpp
@@ -227,7 +227,7 @@ void TR_ReadSampleRequestsHistory::advanceEpoch() // performed by sampling threa
 
 uintptrj_t
 TR_IProfiler::createBalancedBST(uintptrj_t *pcEntries, int32_t low, int32_t high, uintptrj_t memChunk,
-                                TR::Compilation *comp, uintptrj_t cacheStartAddress)
+                                TR::Compilation *comp, uintptrj_t cacheStartAddress, uintptrj_t cacheSize)
    {
    if (high < low)
       return NULL;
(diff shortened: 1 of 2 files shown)


dsouzai:
@mpirvu Could you please review.

runtime/compiler/runtime/IProfiler.cpp
```diff
@@ -3020,11 +3020,43 @@ TR_IPBCDataCallGraph::createPersistentCopy(uintptrj_t cacheStartAddress, TR_IPBC
...
+             *    in the list is NULL, the value stored is also NULL. However, because Thread 2
+             *    updated some previously NULL class, Thread 1 will compute the difference of
+             *    (ramClass->romClass - startOfSCC) and potentially get a value that's bigger than the SCC.
+             *
+             * Because locking the entry in TR_IPBCDataCallGraph::setData would negatively impact
+             * performance, in order to prevent an issue in loadFromPersistentCopy, check again whether
+             * the romClass is within the SCC.
+             */
+            if (romClass >= cacheStartAddress)
+               {
```
mpirvu: Where do we test that the romClass after the SCC?

dsouzai:
@mpirvu Changes made.

On a related note, shouldn't this https://github.com/eclipse/openj9/pull/2510/files#diff-31c18c7318f7ce5d1b5699cf26523d07R2999 be `romClass >= cacheStartAddress+cacheSize`, namely greater than or equal to, instead of just greater than?

mpirvu:
In reply to https://github.com/eclipse/openj9/pull/2510#issuecomment-409220622
the answer is yes you are correct.

dsouzai:
Fixed the copyright and added the `>=` to the line specified in https://github.com/eclipse/openj9/pull/2510#issuecomment-409220622

mpirvu:
jenkins test all

mpirvu:
The few failures are not related to this PR, so merging.

PR #2511, opened 2018-07-31
Correct the binary path of mingw64 for JDK11

ChengJin01:
The minor changes are to move the binary path
of mingw64 to the front of the existing PATH as
cygwin drops its binary path to the frontmost
of the PATH by default.


ChengJin01:
Reviewer: @pshipton

PR #2512, opened 2018-07-31
Limiting Trace Logging by P Codegen

AlenBadel:
Unnecessary output is produced unconditionally during tracing which clogs trace-files. This is an issue specially when profiling for performance.

I've limited some functions in the power codegen to only write to the log if the user is tracing the codegen specifically.

Signed-off-by: Alen Badel <Alen.Badel@ibm.com>


gita-omr:
Just wanted to clarify how it's related to performance. When profiling some benchmarks we sometimes want to generate some JIT tracing for all compiled methods, to aid performance analysis. The extraneous output makes the trace files unmanageable.

PR #2515, opened 2018-07-31
Optimize j/l/Class.isAssignableFrom on X86

0dvictor:
Hook java/lang/Class.isAssignableFrom to existing framework for instanceof/checkcast,
so that certain cases can be further optimized.

### DIFF
--- a/runtime/compiler/x/codegen/J9TreeEvaluator.cpp
+++ b/runtime/compiler/x/codegen/J9TreeEvaluator.cpp
@@ -4504,9 +4504,24 @@ TR::Register *J9::X86::TreeEvaluator::longBitCount(TR::Node *node, TR::CodeGener
 // ----------------------------------------------------------------------------
 inline void generateLoadJ9Class(TR::Node* node, TR::Register* j9class, TR::Register* object, TR::CodeGenerator* cg)
    {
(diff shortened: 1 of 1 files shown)


runtime/compiler/x/codegen/J9TreeEvaluator.cpp
```diff
@@ -4504,9 +4504,25 @@ TR::Register *J9::X86::TreeEvaluator::longBitCount(TR::Node *node, TR::CodeGener
...
+   bool needsNULLCHK = false;
+   switch (node->getOpCodeValue())
+      {
+      case TR::checkcast:
+         break;
+      case TR::checkcastAndNULLCHK:
+         needsNULLCHK = true;
+         break;
+      case TR::instanceof:
+         break;
```
cathyzhyi: nitpicking. Can we group `TR::instanceof` with `TR::checkcast` since they do the same thing?

0dvictor: Sure, updated.

cathyzhyi:
LGTM

fjeremic:
Jenkins test sanity xlinux

0dvictor:
JDK11 failure due to #2536.

Issue #2516, opened 2018-07-31
JVMTI should ignore extra callbacks


gacholio:
JVMTI SetEventCallbacks currently fails if the size of the input callback table exceeds the number of callbacks known to the JVM.  This causes problems if a future (not yet supported) version of JVMTI with more callbacks is used by an agent.

New versions of JVMTI never change the numbering of existing callbacks, so the solution is to simply ignore any callbacks that exceed the known size.

PR #2517, opened 2018-07-31
Ignore unknown JVMTI callbacks

gacholio:
Rather than error, just ignore callbacks that exceed the known table
size.

Fixes: #2516


smlambert:
Is there a test case that can or should be added?

gacholio:
Can, yes, should, not so sure.

pshipton:
jenkins compile plinux

PR #2524, opened 2018-07-31
Print error reason when AOT load fails

0dvictor:
Error message now includes prepareRelocateAOTCodeAndData() return code and its meaning.

### DIFF
--- a/runtime/compiler/control/CompilationThread.cpp
+++ b/runtime/compiler/control/CompilationThread.cpp
@@ -9273,7 +9273,11 @@ TR::CompilationInfo::compilationEnd(J9VMThread * vmThread, TR::IlGeneratorMethod
 
                         if (TR::Options::isAnyVerboseOptionSet(TR_VerboseCompileEnd, TR_VerbosePerformance, TR_VerboseCompFailure))
                            {
-                           TR_VerboseLog::writeLineLocked(TR_Vlog_FAILURE,"Failure while relocating for %s", comp->signature());
+                           TR_VerboseLog::writeLineLocked(TR_Vlog_FAILURE,
+                                                          "Failure while relocating for %s, return code = %d [%s]\n",
+                                                          comp->signature(),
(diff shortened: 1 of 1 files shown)


0dvictor:
@mpirvu do you mind review this change?

runtime/compiler/control/CompilationThread.cpp
```diff
@@ -9273,7 +9273,7 @@ TR::CompilationInfo::compilationEnd(J9VMThread * vmThread, TR::IlGeneratorMethod
 
                         if (TR::Options::isAnyVerboseOptionSet(TR_VerboseCompileEnd, TR_VerbosePerformance, TR_VerboseCompFailure))
                            {
-                           TR_VerboseLog::writeLineLocked(TR_Vlog_FAILURE,"Failure while relocating for %s", comp->signature());
+                           TR_VerboseLog::writeLineLocked(TR_Vlog_FAILURE,"Failure while relocating for %s, return code = %d [%s]\n", comp->signature(), returnCode, compilationErrorNames[returnCode]);
```
dsouzai: This will crash when you have an AOT load that failed to relocate a Debug Counter. That's my fault though; `TR_RelocationRecordDebugCounter::applyRelocation` returns -1, when it should've returned `compilationFailure`. Do you mind changing it so that it doesn't return -1 anymore?

0dvictor: Good point; but IMO changing the return code of `TR_RelocationRecordDebugCounter::applyRelocation` is out of the scope that this PR tries to achieve. I've made update so that it won't crash even if the return code isn't any of the defined one.

mpirvu:
jenkins test all

fjeremic:
JDK11 failures due to #2536.

mpirvu:
Only JDK11 and Windows tests fail for reasons that are not associated with this PR, thus merging.

PR #2530, opened 2018-08-01
WIP: Incorporate Intel Transactional Memory (TSX) into OpenJ9 Locking

doubletea:
Transactional Lock Elision

Currently, the J9JIT (with TLE enabled) transforms monenters/monexits
of synchronized blocks to execute transactionally, if analyisis of
the block passes certain conditions.

The issue arises when some synchronized blocks are transformed to
use TLE, while others are not, AND they are synchronized on the same
lock object.

A thread cannot commit a transaction if the lock is also synchronized
by another thread by non-TLE locking. This creates excessive aborts.

This patch is a work-in-progress implementation of Restricted
Transactional Memory (RTM) as applied to OpenJ9 locks and is intended
to replace the TLE implementation in J9JIT. This approach avoids the
aformentioned problems by eliding locks at the lock object level as
opposed to the lock sites.

This patch depends on patch https://github.com/eclipse/omr/pull/2818

See https://github.com/eclipse/openj9/issues/2484

[skip ci]

Signed-off-by: John Xu <xujohn@ca.ibm.com>

### DIFF
--- a/runtime/vm/BytecodeInterpreter.hpp
+++ b/runtime/vm/BytecodeInterpreter.hpp
@@ -1085,7 +1085,7 @@ obj:;
 	enterObjectMonitor(REGISTER_ARGS_LIST, j9object_t obj)
 	{
 		IDATA rc = (IDATA)obj;
-		if (!VM_ObjectMonitor::inlineFastObjectMonitorEnter(_currentThread, obj)) {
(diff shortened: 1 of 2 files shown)


doubletea:
@0xdaryl This PR will be closed and is serving as documentation to issue eclipse/openj9#2484 for further work. The branch will be kept on my side indefinitely.

PR #2532, opened 2018-08-01
Correct LICENSE typo

ChanHoHo:
Correct typo "2017" to "2018" in LICENSE  @pshipton 

Signed-off-by: ChanHoHo <chanho95@hotmail.com>

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,6 +1,6 @@
-Product: OpenJ9
+﻿Product: OpenJ9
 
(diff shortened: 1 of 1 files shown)


keithc-ca:
The only change I see (besides the copyright date) is the addition of (what I assume is a byte-order mark) `0xEF 0xBB 0xBF` before the first character.

@ChanHoHo Why do you think we want or need that?

PR #2537, opened 2018-08-01
Remove non-ASCII characters

keithc-ca:
Use '(c)' instead of non-ASCII copyright symbol.

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,4 +1,4 @@
-﻿Product: OpenJ9
+Product: OpenJ9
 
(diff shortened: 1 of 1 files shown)


pmhayward:
FWIW, I think you'll find that the correct alternative for the © symbol is (C) rather than (c). i.e uppercase 'C' in parentheses.

keithc-ca:
Perhaps (C) might have been better than (c), but the latter is the convention we've established in every file that contains a copyright notice.

pmhayward:
np! Your call!

PR #2542, opened 2018-08-02
-Djava.security.manager=default should create java.lang.SecurityManager

### LINKED ISSUE #2541: Setting 'java.security.manager=default' doesn't seem to be supported
In OpenJ9 setting "java.security.manager=default" does seem to be supported and causes an exception:
```
Exception in thread "main" java/lang/Error: JVM can't set custom SecurityManager due to java.lang.ClassNotFoundException: default
	at java/lang/ClassLoader.initializeClassLoaders (java.base@9/ClassLoader.java:232)
	at java/lang/Thread.initialize (java.base@9/Thread.java:422)
	at java/lang/Thread.<init> (java.base@9/Thread.java:153)
```
This was found when running some of the JTReg tests on xLinux and z/OS. One test that fails is:
```
jdk/com/sun/security/auth/login/ConfigFile/PropertyExpansion.java
```
When this test executes it sets '-Djava.security.manager=default' on the java command line when executing the test class and hence causes the test to fail.

pshipton:
Setting -Djava.security.manager=default should create an instance of
java.lang.SecurityManager

Fixes #2541

[ci skip]

### DIFF
--- a/jcl/src/java.base/share/classes/java/lang/ClassLoader.java
+++ b/jcl/src/java.base/share/classes/java/lang/ClassLoader.java
@@ -268,7 +268,7 @@ public abstract class ClassLoader {
 		jdk.internal.misc.VM.initLevel(2);
 		String javaSecurityManager = System.internalGetProperties().getProperty("java.security.manager"); //$NON-NLS-1$
 		if (null != javaSecurityManager) {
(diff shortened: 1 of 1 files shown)


pshipton:

PR #2545, opened 2018-08-02
Skip read barriers from hardware generated guarded storage events if …

fjeremic:
…within a transactional region

In the event that the GSECI TX bit indicates that a hardware
generated guarded storage event occured within a transactional
execution region we should abort the read barrier and return back to
the software to abort the transaction. The reason for this is to
avoid the following scenario:

- We have a TX region in which two reference reads happen on o.f
- The first read of object o.f happens via a software read barrier
    - For example from an interpreted method
- The software read barrier will copy the object into the copy cache
- The second read of o.f.g happens via a hardware read barrier
    - This triggers a hardware transaction abort
    - All stores within the transaction are rolled back

In the above scenario before reaching the guarded storage event
handler according to Principles Of Operation S.4 the transaction
will have aborted by the time control is passed off to the event
handler. Because of this the original copy of object o.f in the copy
cache would have been rolled back to what it was prior to the GS
event handler. This location could have arbitrary data, including
a reasonable value which looks like an evacuate address which will
trip up the GC because we are effectively looking at garbage.

The solution here is to skip the read barriers from hardware
generated guarded storage event if we detect that the guarded storage
event was triggered from within a transactional execution region,
which will always result in the transaction aborting before we
reach the GS event handler.

Prior to branching back to the return address (populated by hardware
in this case to point to after the TBEGIN) we set CC=2 to mimic the
behavior of a transient transaction abort, and let the normal TABORT
software path handle the fallback.

### DIFF
--- a/runtime/jilgen/jilconsts.c
+++ b/runtime/jilgen/jilconsts.c
@@ -507,6 +507,7 @@ writeConstants(OMRPortLibrary *OMRPORTLIB, IDATA fd)
 			writeConstant(OMRPORTLIB, fd, "J9TR_VMThread_entryLocalStorage", offsetof(J9VMThread, entryLocalStorage)) |
 			writeConstant(OMRPORTLIB, fd, "J9TR_VMThread_stackWalkState", offsetof(J9VMThread, stackWalkState)) |
 #if defined(OMR_GC_CONCURRENT_SCAVENGER) && defined(J9VM_ARCH_S390)
+			writeConstant(OMRPORTLIB, fd, "J9TR_VMThread_gsParameters_GSECI", offsetof(J9VMThread, gsParameters) + 2) |
 			writeConstant(OMRPORTLIB, fd, "J9TR_VMThread_gsParameters_returnAddr", offsetof(J9VMThread, gsParameters.returnAddr)) |
 #endif /* OMR_GC_CONCURRENT_SCAVENGER */
 
--- a/runtime/oti/zhelpers.m4
+++ b/runtime/oti/zhelpers.m4
@@ -1,4 +1,4 @@
-dnl Copyright (c) 1991, 2017 IBM Corp. and others
+dnl Copyright (c) 1991, 2018 IBM Corp. and others
 dnl
 dnl This program and the accompanying materials are made available under
 dnl the terms of the Eclipse Public License 2.0 which accompanies this
@@ -74,6 +74,7 @@ define({C_GPR},cg)
 define({CHI_GPR},cghi)
 define({CDS_GPR},cdsg)
 define({CL_GPR},clg)
+define({CLFI_GPR},clgfi)
 define({CLM_GPR},clmh)
 define({CLR_GPR},clgr)
 define({CR_GPR},cgr)
@@ -124,6 +125,7 @@ define({C_GPR},c)
 define({CHI_GPR},chi)
 define({CDS_GPR},cds)
 define({CL_GPR},cl)
+define({CLFI_GPR},clfi)
 define({CLM_GPR},clm)
 define({CLR_GPR},clr)
 define({CR_GPR},cr)
--- a/runtime/vm/zcinterp.m4
+++ b/runtime/vm/zcinterp.m4
@@ -153,13 +153,10 @@ END_CURRENT
 })
 
 ifdef({OMR_GC_CONCURRENT_SCAVENGER},{
-dnl Helper to handle guarded storage events and
-dnl software read barriers. 
-dnl When event occurs, hardware populates return address in
-dnl vmThread->gsParameters.returnAddr
-dnl For software read barriers gsParameters.returnAddr is set by JIT
+dnl Helper to handle guarded storage events and software read barriers. When
+dnl the event occurs, hardware/software populates the return address in 
+dnl J9VMThread.gsParameters.returnAddr.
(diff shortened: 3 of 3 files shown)


fjeremic:
Will fix the copyright when final reviews are done. I'm making this as WIP until I've done all the proper testing, but the change should be ready for review @joransiu @amicic.

fjeremic:
The following assembly listing is generated when compiling zcinterp.s:

```
00000170                                     451      DS    0D
00000170                00000170 00000120    452 Z#J9SYM3  CSECT
                                             453 Z#J9SYM3  RMODE ANY
                                             454 Z#J9SYM3  AMODE ANY
00000170 00C300C500C500F1                    455      DC    0D'0',XL8'00C300C500C500F1'
00000178 000000F800000000                    456      DC    AL4(PPAZJ9SYM3-Z#J9SYM3),AL4(0)
00000180                                     457 ZJ9SYM3  DS 0D
                                             458      ENTRY ZJ9SYM3
                                             459 ZJ9SYM3  ALIAS C'handleGuardedStorageEvent'
                                             460 ZJ9SYM3  XATTR SCOPE(X),LINKAGE(XPLINK)
                                             461 ZJ9SYM3  AMODE 64
00000180 9180 DB02      00000B02             464     TM 2818(13),128
00000184 A784 0007               00000192    465     JZ L_GS_CALLHELPER
00000188 C25E 0000 0000          00000000    466     clgfi 5,0
0000018E A7F4 006A               00000262    467     J L_GS_DONE
00000192                                     469 L_GS_CALLHELPER DS 0H
00000192 E34D 0128 0024          00000128    472     stg 4,296(13)
00000198 E34D 0960 0004          00000960    474     lg 4,2400(13)
0000019E D707 D960 D960 00000960 00000960    475     xc 2400(8,13),2400(13)
000001A4 EB0F 4A80 0024          00000A80    478     stmg 0,15,2688(4)
000001AA D207 4AA0 D128 00000AA0 00000128    480     mvc 2720(8,4),296(13)
...
...
00000250 6854 0B28               00000B28    576     ld 5,2856(4)
00000254 6864 0B30               00000B30    577     ld 6,2864(4)
00000258 6874 0B38               00000B38    578     ld 7,2872(4)
0000025C                                     581 L_RV3 DS 0H
0000025C EB0F 4A80 0004          00000A80    586     lmg 0,15,2688(4)
00000262                                     590 L_GS_DONE DS 0H
00000262 E3                                  593     DC X'E3'
00000263 F0                                  594     DC X'F0'
00000264 DB                                  595     DC X'DB'
00000265 28                                  596     DC X'28'
00000266 00                                  597     DC X'00'
00000267 47                                  598     DC X'47'
                        000000E8             601 CODELZJ9SYM3      EQU *-ZJ9SYM3
```

Judging from assembly listing the `TM` instruction reads from offset 2818 off `J9VMThread` and the `BIC` which is hex encoded at the end branches to offset 0xB28 = 2856. 2856 - 2818 = 38 bytes which is exactly the distance from `transactionalExecutionMode` to `returnAddr`:

https://github.com/eclipse/openj9/blob/b2b62046eb4c221ad1773457bae20da74f08ec09/runtime/oti/j9port_generated.h#L77-L134

I think we should be good here, but a second pair of eyes is always good.

fjeremic:
Fixed copyright. This is a Z only change and we don't have external CS testing so a build just on JDK8 to validate we have no compilation issues will suffice. I'll kick one off on the committers behalf.

fjeremic:
I think we should be good to go here. @amicic any other concerns?

joransiu:
Re-reading the scenario documented in the description, I don't think "The second read of o.f happens via a hardware read barrier" will trigger a hardware read barrier; The software barrier copied the object o.f to the copy cache, and would have updated the field reference o.f to the survivor space.

The failing scenario would rather be reading a field value off the o.f object  (i.e. o.f.g) to trigger the hardware read barrier.  o.f would be the new copy in the copy-cache, with o.f.g still pointing in the evacuate space.  All of this will be undone on TX abort.

fjeremic:
> Re-reading the scenario documented in the description, I don't think "The second read of o.f happens via a hardware read barrier" will trigger a hardware read barrier; The software barrier copied the object o.f to the copy cache, and would have updated the field reference o.f to the survivor space.
> 
> The failing scenario would rather be reading a field value off the o.f object (i.e. o.f.g) to trigger the hardware read barrier. o.f would be the new copy in the copy-cache, with o.f.g still pointing in the evacuate space. All of this will be undone on TX abort.

Fixed in 756068c. Also updated a comment in zcinterp.m4 to reference this PR instead of #2666 which I left as a placeholder (chicken and egg before I opened this PR). There were no code changes, so no need to re-run the testing.

dmitripivkine:
@fjeremic FYI Aleks is on vacation this week

fjeremic:
> @fjeremic FYI Aleks is on vacation this week

Ah, ok thanks. @joransiu I'll leave it to your discretion whether awaiting @amicic's review should block the merge of this PR.

joransiu:
> Ah, ok thanks. @joransiu I'll leave it to your discretion whether awaiting @amicic's review should block the merge of this PR.

I propose this is safe to merge.  @amicic had previously agreed with this approach to skip the handler under TX aborts.  The bulk of the change is Z specific assembly.

PR #2546, opened 2018-08-02
Support for dynamically adjusting scratch memory space of compilations

mpirvu:
The JIT is allowed to use up to 256 MB of 'scratch' memory for a compilation
while JSR292 compilations increase this limit to 1.5 GB. These limits create
footprint spikes during an application run which can make resource
provisioning more cumbersome.

This commit adds the following changes:
1) The JIT will use portlib APIs to compute the amount of physical memory
available (taking into consideration any container limits). This value
is cached (to minimize API overhead) and recomputed with a desired
frequency (0.5 sec by default)

2) At the beginning of a compilation the JIT checks the amount of available
physical memory and sets a limit for the scratch space accordingly. If the
amount of physical memory is so low that a relatively cheap compilation cannot
be performed, the compilation request is aborted right there.

3) When a compilation thread needs to allocate a new scratch memory segment
(of size 16 MB) it checks again the available physical memory. If this is low,
the compilation is aborted and will be retried at a lower optimization level.
At the same time a flag will be set which will make one compilation thread
to suspend itself (however, the JIT always makes sure there is at least one
compilation thread active).

### DIFF
--- a/runtime/compiler/control/CompilationThread.cpp
+++ b/runtime/compiler/control/CompilationThread.cpp
@@ -364,6 +364,12 @@ TR_YesNoMaybe TR::CompilationInfo::shouldActivateNewCompThread()
       if ((getNumCompThreadsActive() + 1) * 100 >= (TR::Options::_compThreadCPUEntitlement + 50))
          return TR_no;
       }
+   // Do not activate if we are low on physical memory
+   bool incompleteInfo;
+   uint64_t freePhysicalMemorySizeB = computeAndCacheFreePhysicalMemory(incompleteInfo);
+   if (freePhysicalMemorySizeB != OMRPORT_MEMINFO_NOT_AVAILABLE &&
+       freePhysicalMemorySizeB <= (uint64_t)TR::Options::getSafeReservePhysicalMemoryValue() + TR::Options::getScratchSpaceLowerBound())
+      return TR_no;
    // Do not activate a new thread during graceperiod if AOT is used and first run because
    // we may have too many warm compilations at warm. However, there is no such risk for quickstart
    // Another exception: activate if second run in AOT mode
@@ -792,7 +798,12 @@ TR::CompilationInfo::CompilationInfo(J9JITConfig *jitConfig) :
    // the options have been processed
 
    _vmStateOfCrashedThread = 0;
+
    _cachedFreePhysicalMemoryB = 0;
+   _cachedIncompleteFreePhysicalMemory = false;
+   OMRPORT_ACCESS_FROM_J9PORT(jitConfig->javaVM->portLibrary);
+   _cgroupMemorySubsystemEnabled = (OMR_CGROUP_SUBSYSTEM_MEMORY == omrsysinfo_cgroup_are_subsystems_enabled(OMR_CGROUP_SUBSYSTEM_MEMORY));
+   _suspendThreadDueToLowPhysicalMemory = false;
 
    // Initialize the compilation monitor
    //
@@ -3268,6 +3279,7 @@ TR::CompilationInfoPerThread::waitForWork()
 void
 TR::CompilationInfoPerThread::doSuspend()
    {
+   _compInfo.setSuspendThreadDueToLowPhysicalMemory(false);
    getCompThreadMonitor()->enter();
    setCompilationThreadState(COMPTHREAD_SUSPENDED);
    _compInfo.releaseCompMonitor(getCompilationThread());   // release the queue monitor before waiting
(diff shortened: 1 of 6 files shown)


fjeremic:
@dsouzai FYI for review.

mpirvu:
Note that at the moment the code is not ready to be merged (hence the WIP heading).
Also, the fix from https://github.com/eclipse/omr/pull/2785 is needed for this commit to work correctly.

mpirvu:
@ashu-mehra  FYI for review

runtime/compiler/control/CompilationThread.cpp
```diff
@@ -10462,71 +10456,132 @@ TR::CompilationInfo::increaseUnstoredBytes(U_32 aotBytes, U_32 jitBytes)
 #endif
    }
 
+
+// Compute free physical memory taking into account container limits
+// To limit overhead, this function caches the computed value and
+// only refreshes it if more than 0.5 sec have passed since the last update
+// or if "now" parameter is true
```
dsouzai: Where is the "now" parameter mentioned in this comment?

dsouzai:
The changes lgtm.

runtime/compiler/control/CompilationThread.cpp
```diff
@@ -10462,71 +10456,132 @@ TR::CompilationInfo::increaseUnstoredBytes(U_32 aotBytes, U_32 jitBytes)
...
+      if (memInfo.hostCached != OMRPORT_MEMINFO_NOT_AVAILABLE) // Note 'cached' is not available in containers
+         freeHostPhysicalMemorySizeB += memInfo.hostCached;
+      else
+         incomplete = true;
 #if defined(LINUX)
-            if (memInfo.buffered != OMRPORT_MEMINFO_NOT_AVAILABLE)
-               freePhysicalMemorySizeB += memInfo.buffered;
-            else
-               incomplete = true;
+      if (memInfo.buffered != OMRPORT_MEMINFO_NOT_AVAILABLE)
```
ashu-mehra: As per https://github.com/eclipse/omr/blob/4e352d639ed726de60b827a0e922be5181180729/port/unix/omrsysinfo.c#L1616 `buffered` is not available in containers. Does that mean we would always set `incomplete` to `true` when running in containers?

fjeremic:
@mpirvu what is the state of this PR?

mpirvu:
I just had a stab at it last night but in my local experiments I found the j9sysinfo_get_memory_info API to give the same results for host and container memory (which is wrong). I had a discussion with Ashutosh and the API seems to be working fine for him. Now it's on me to determine why j9sysinfo_get_memory_info gives me bad information.

mpirvu:
With some instrumentation I determined that 
`rc = readCgroupSubsystemFile(portLibrary, OMR_CGROUP_SUBSYSTEM_MEMORY, CGROUP_MEMORY_SWAP_LIMIT_IN_BYTES_FILE, numItemsToRead, "%lu", &cgroupMemInfo->memoryAndSwapLimit);`
return an error code and therefore `retrieveLinuxCgroupMemoryStats` returns error even though the memoryLimit and memoryUsage were correctly read.
In my opinion  `retrieveLinuxCgroupMemoryStats` should always return 'success' and the validity of various memory statistics can be determined by comparison with `OMRPORT_MEMINFO_NOT_AVAILABLE`

mpirvu:
I should also mention that on two Ubuntu 16.04 systems where I tried, when launching a container with a memory limit I get this warning:
`WARNING: Your kernel does not support swap limit capabilities or the cgroup is not mounted. Memory limited without swap.`

runtime/compiler/control/CompilationThread.cpp
mpirvu: Thinking about this, we don't want to set incomplete for container, thus there are two possible solutions
1) In `retrieveLinuxCgroupMemoryStats` always fill with 0 the buffered field instead of UNAVAILABLE
2) In `computeFreePhysicalMemory` check whether we are in a container by using
`if (OMR_CGROUP_SUBSYSTEM_MEMORY == omrsysinfo_cgroup_are_subsystems_enabled(OMR_CGROUP_SUBSYSTEM_MEMORY)) `
I might go with solution (2) because I think I should not add the cache value to the free physical memory. The reason is that on the host the OS can increase the physical memory by releasing some cached pages, but in container the available memory is (limit-used) and any cached pages are not going to help.

mpirvu:
Jenkins test all all jdk8

mpirvu:
Could I get another review please: @ashu-mehra  @dsouzai

runtime/compiler/control/CompilationThread.cpp
```diff
@@ -10466,71 +10483,116 @@ TR::CompilationInfo::increaseUnstoredBytes(U_32 aotBytes, U_32 jitBytes)
...
+         freePhysicalMemory += memInfo.buffered;
+      else
+         incomplete = !_cgroupMemorySubsystemEnabled;
```
dsouzai: Is it possible for `incomplete` to be true from the line 10510, but then set to false in this line? If so, Shouldn't there be a check to not set it to false?

mpirvu: You are right. I need to correct the behavior to never change it to 'false' once we set it to 'true'

runtime/compiler/env/J9SegmentAllocator.cpp
```diff
@@ -50,6 +50,33 @@ J9MemorySegment *
...
+      bool incomplete;
+      TR::CompilationInfo *compInfo = TR::CompilationInfo::get(_javaVM.jitConfig);
+      uint64_t freePhysicalMemory = compInfo->computeAndCacheFreePhysicalMemory(incomplete, 20);
+      if (freePhysicalMemory != OMRPORT_MEMINFO_NOT_AVAILABLE && !incomplete)
+         {
+         if (freePhysicalMemory < (TR::Options::getSafeReservePhysicalMemoryValue() + segmentSize))
+            {
+            // Set a flag that will determine one compilation thread to suspend itself
+            // Even if multiple threads enter his path we still want to suspend only
+            // on thread, not several. This is why we use a flag and not a counter.
```
ashu-mehra: typo: on -> one

ashu-mehra:
Looks fine to me.

ashu-mehra:
> In my opinion retrieveLinuxCgroupMemoryStats should always return 'success' and the validity of various memory statistics can be determined by comparison with OMRPORT_MEMINFO_NOT_AVAILABLE

I think we also need to set `memoryAndSwapLimit` to 0 if the file does not exist. And if we do that then that change alone will fix the issue.

runtime/compiler/env/J9SegmentAllocator.cpp
mpirvu: Fixed

fjeremic:
Jenkins test sanity all JDK8

PR #2548, opened 2018-08-02
WIP: Add Transaction Diagnostic Block for Transactional Memory

sehirst:
When the CPU is under Transactional Execution mode on z, if an abort
happens and a memory location for the TDB is specified, then the
current state will be written to that memory location. The GPRs are
rolled back to the TBEGIN instruction, so the TDB is needed to see the
state at the time of the abort. This is useful for debugging problems
within TX. The TDB also has additional abort related info useful for
performance analysis of TX/TM.

Signed-off-by: Simon Hirst <shirst@ca.ibm.com>


sehirst:
I also deleted a bunch of #ifdef 0 code from an old developer which was experimental and never finished.

fjeremic:
@sehirst whenever this is ready for review please remove the WIP and tag me in a comment. Thanks!

runtime/compiler/env/VMJ9.cpp
```diff
@@ -1865,6 +1865,24 @@ UDATA TR_J9VMBase::thisThreadGetOSRReturnAddressOffset()
...
+ */
+uint16_t TR_J9VMBase::thisThreadGetTDBOffset()
+   {
+#if !defined(J9ZTPF)
+   // This assume is valid to confirm the TDB is on an 8 byte boundary because we know the VMThread is already aligned on a 256 byte boundary
+   // If this assume fails, then that means someone modified the J9VMThread structure which has thrown off the alignment.
+   TR_ASSERT(offsetof(J9VMThread, transactionDiagBlock)%8 == 0, "The Transaction Diagnostic Block must be aligned on a doubleword i.e. 8 byte boundary");
+   return offsetof(J9VMThread, transactionDiagBlock);
+#else
+   return 0;
```
DanHeidinga: Is `0` the right value for the case where there is no `transactionDiagBlock`?  Won't that cause the vmThread to be corrupted if this code was incorrectly run on any other platform?

In VM code, we'd use a TraceAssert to mark this case as illegal.

sehirst: That's from `#if !defined(J9ZTPF)`. I can't comment on that. I didn't add that `ifdef`, the zptf people did. It's been like that for a long time on older pre-open source versions with no issues, so I assume the code works for them.

DanHeidinga: Likely on the assumption that it's never called.  Better to be explicit than to leave potential corruption points in the code.

sehirst: Can you suggest how I should fix this? I don't follow what you mean by "use a TraceAssert to mark this case as illegal".

DanHeidinga: Similar to the trace assert used elsewhere in the JIT.  ie https://github.com/eclipse/openj9/blob/afbe2b92c85fe01ef44df0b6b5ce45421b241f5f/runtime/compiler/control/CompilationThread.cpp#L5265

sehirst: Will that assert fire in a prod/released build?

DanHeidinga: Yes.  Similar to all asserts in the VM / GC code.

sehirst: Upon further reflection, I need to discuss with a ztpf dev about what should be done here. Switching the `return 0` to a fatal assert seems risky without consulting them.

sehirst: Ok. Waiting on some testing, then will remove WIP.

sehirst:
Needed to rebase for testing.

sehirst:
Squashed my commits, ready to review and merge.

runtime/compiler/z/codegen/J9TreeEvaluator.cpp
```diff
@@ -12136,6 +12148,7 @@ J9::Z::TreeEvaluator::tstartEvaluator(TR::Node * node, TR::CodeGenerator * cg)
    //    POST Deps
 
    TR::Compilation *comp = cg->comp();
+   TR_J9VMBase *fej9 = (TR_J9VMBase *)(cg->fe());
```
fjeremic: Let's try an avoid introducing new unsafe C-style casts into the codebase and use `static_cast<TR_J9VMBase*>(cg->fe())` here instead.

runtime/compiler/z/codegen/J9TreeEvaluator.cpp
```diff
@@ -11287,18 +11293,21 @@ inlineConcurrentLinkedQueueTMOffer(
...
-         cursor = generateSILInstruction(cg, TR::InstOpCode::TBEGIN, node, generateS390MemoryReference(cg->machine()->getS390RealRegister(TR::RealRegister::GPR0),0,cg), 0xFF02);
+         // immediate field described in TR::TreeEvaluator::tstartEvaluator
+         cursor = generateSILInstruction(cg, TR::InstOpCode::TBEGIN, node, TDBmemRef, 0xFF02);
 
          cursor = generateS390BranchInstruction(cg, TR::InstOpCode::BRC, TR::InstOpCode::COND_MASK7, node, failLabel);
          }
       else
-         {
-         cursor = generateSILInstruction(cg, TR::InstOpCode::TBEGINC, node, generateS390MemoryReference(cg->machine()->getS390RealRegister(TR::RealRegister::GPR0),0,cg), 0xFF00);  //GPRMask
+         { // No TDB for constrained transactions
```
fjeremic: Comment is on the same line as the brace where as other comments are not. Can we be consistent with the rest of the codebase?

runtime/compiler/env/VMJ9.cpp
```diff
@@ -1865,6 +1865,24 @@ UDATA TR_J9VMBase::thisThreadGetOSRReturnAddressOffset()
...
+/**
+ * @brief TDB is Transaction Diagnostic Block used for Transactional Memory Debugging.
+ * It is a 256 Byte block stored in the J9VMThread and must be 8 byte boundary aligned.
+ */
+uint16_t TR_J9VMBase::thisThreadGetTDBOffset()
+   {
+#if !defined(J9ZTPF)
+   // This assume is valid to confirm the TDB is on an 8 byte boundary because we know the VMThread is already aligned on a 256 byte boundary
+   // If this assume fails, then that means someone modified the J9VMThread structure which has thrown off the alignment.
+   TR_ASSERT(offsetof(J9VMThread, transactionDiagBlock)%8 == 0, "The Transaction Diagnostic Block must be aligned on a doubleword i.e. 8 byte boundary");
```
fjeremic: Spaces around binary operators needed (`%`).

runtime/compiler/env/VMJ9.cpp
fjeremic: Did we ever figure out what to do with this? It seems @DanHeidinga's concern was not addressed.

fjeremic: Ok, I would say lets not bother then however putting in the assert (`Assert_JIT_unreachable();`) will help them catch this bug much faster. So I do support putting in the assert.

runtime/compiler/z/codegen/J9TreeEvaluator.cpp
fjeremic: This change is still missing. Can you squash after this and the one below is done and I'll kick off the Jenkins builds?

sehirst: It's here https://github.com/eclipse/openj9/pull/2548/commits/dfab63d5f5c8b8642886de00cd80a61e8d840b1f#diff-e91b45c91f84f8f7ea1179eb4ac01d36R11310

sehirst:
squashed

runtime/compiler/z/codegen/J9TreeEvaluator.cpp
sehirst: Oh right, the 11533 line

sehirst:
Assert added and other comment fixed.

fjeremic:
@sehirst could you please fix the copyrights and the line endings?

sehirst:
They failed for infra reasons. They were green before. Rebasing and pushing again to trigger new builds.

sehirst:
No new commits, everything's still up to date, so I can't trigger another build by pushing without making a change.

fjeremic:
Jenkins copyright check

fjeremic:
Jenkins line endings check

fjeremic:
Jenkins test sanity JDK8 zlinux

fjeremic:

PR #2554, opened 2018-08-07
Use the right symbol kind for String constructors

liqunl:
Wrong symbol kind on a method will cause numberOfParameters to return a
wrong result. Fix StringPeepholes to use the right symbol kind.

### DIFF
--- a/runtime/compiler/optimizer/StringPeepholes.cpp
+++ b/runtime/compiler/optimizer/StringPeepholes.cpp
@@ -213,9 +213,13 @@ TR::SymbolReference* TR_StringPeepholes::MethodEnumToArgsForMethodSymRefFromName
       {
       if (!fe()->getMethodFromName(classNames[m], methodNames[m], signatures[m], comp()->getCurrentMethod()->getNonPersistentIdentifier()))
          {
-         return comp()->getSymRefTab()->methodSymRefFromName(comp()->getMethodSymbol(), classNames[m], methodNames[m], "([CIIZ)V", TR::MethodSymbol::Static);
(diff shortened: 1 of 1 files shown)


liqunl:
@cathyzhyi @andrewcraik Can you review?

cathyzhyi:
LGTM

runtime/compiler/optimizer/StringPeepholes.cpp
```diff
@@ -213,10 +213,13 @@ TR::SymbolReference* TR_StringPeepholes::MethodEnumToArgsForMethodSymRefFromName
       {
       if (!fe()->getMethodFromName(classNames[m], methodNames[m], signatures[m], comp()->getCurrentMethod()->getNonPersistentIdentifier()))
          {
-         return comp()->getSymRefTab()->methodSymRefFromName(comp()->getMethodSymbol(), classNames[m], methodNames[m], "([CIIZ)V", TR::MethodSymbol::Static);
+         return comp()->getSymRefTab()->methodSymRefFromName(comp()->getMethodSymbol(), classNames[m], methodNames[m], "([CIIZ)V", TR::MethodSymbol::Special);
          }
       }
 
+   if (m >START_STRING_METHODS && m < START_BIG_DECIMAL_HELPER_METHODS)
```
andrewcraik: This concerns me a little bit - if anyone adds a non-init method we could get the wrong symbol kind. Could we at least have an assert that the method is an <init> method which is the issue in play here? Similarly should the other path verify that the method is not an <init> method or something like that? This just seems fragile and prone to break

andrewcraik:
Jenkins test sanity

PR #2558, opened 2018-08-07
Check that PersistentCHTable is not NULL, in jitClassesRedefined

dmitry-ten:
`jitClassesRedefined` calls `methodGotOverridden` on a PersistentCHTable, 
even when the table is NULL.
This can cause undefined behavior, so check that the table is not NULL, before calling the method.


dmitry-ten:
Done.

PR #2560, opened 2018-08-08
Set Maximum Old Space size default value properly

dmitripivkine:
We noticed that maxOldSpaceSize is set to memoryMax by default
This is not correct for Gencon heap geometry where it should be set to
maximum heap memory minus initial Nursery size - obviously the ultimate
case when Nursery never expands and stays on it's initial size and all
remaining memory might be taken for Tenure.
This change is not applicable for case if -Xmox is specified in command
line explicitly.


dmitripivkine:

Issue #1885, opened 2018-05-10
testSCCMLTests1_1 sensitive to PR comment


gacholio:
Test 45: LIR 1445.1 : Debug Area PrintStats Test : Ensure debug area sizes are correct in generated javacore

This test essentially greps a javacore for some success and fail conditions.  One of the fail conditions is:

```>> Failure condition was found: [Output match: corrupt]```

The javacore contains the contents of all environment variables, one of which is the PR comment:

```[OUT] 2CIENVVAR      ghprbPullLongDescription=J9VMThreadPointerUtil.getJ9State() was copied from thrinfo.c but is\r\nmissing some checks.  This can lead to a NULL pointer dereference,\r\ncausing a CorruptDataException to be thrown.\r\n\r\nSigned-off-by: Graham Chapman <graham_chapman@ca.ibm.com>
```

So any PR which contains the word "corrupt" will fail this test.

gacholio:
@smlambert FYI

smlambert:
Ha, I think we all know how much I enjoy string matching as a pass/fail criteria!  We can likely tighten up this particular match immediately.  

We ported all of the cmdlinetester tests almost 'as is', so I've not had a chance to think about how to improve them as a whole.  As this type of failure will continue to happen (irregularly), it would be good to address it for all cmdlinetests.  That will have to come as a separate issue/enhancement.

smlambert:
Actually I see now that I should be able to avoid parse/grep of the variable ghprbPullLongDescription which can solve for all future cases.

PR #2564, opened 2018-08-08
Remove unnecessary failure conditions in shared class CML tests

hangshao0:
"cat" command won't print out the failure messages that are being
currently checked. These checks are sensitive to PR comments (See #1885)
Remove them.


hangshao0:
reviewer: @llxia

llxia:
Jenkins test all xlinux jdk10

llxia:
This is not a fix for #1885. It just these failure checks are not necessary for the tests and they cause false negative result.

Issue #2472, opened 2018-07-25
cmdLineTester_jvmtitests_hcr_OSRG_nongold_1 Test-extended.functional-JDK10-linux_x86-64 rc004 gpf


pshipton:
https://ci.eclipse.org/openj9/job/Test-extended.functional-JDK10-linux_x86-64/26

```
Testing: rc004
Test start time: 2018/07/25 07:30:50 Coordinated Universal Time
Running command: "/home/jenkins/workspace/Test-extended.functional-JDK10-linux_x86-64/openjdkbinary/j2sdk-image/bin/java"  -Xnocompressedrefs -Xjit:enableOSR,enableOSROnGuardFailure,count=1,disableAsyncCompilation -Xgcpolicy:gencon  -Xdump    -agentlib:jvmtitest=test:rc004 -cp "/home/jenkins/workspace/Test-extended.functional-JDK10-linux_x86-64/openjdk-tests/TestConfig/scripts/testKitGen/../../../../jvmtest/functional/cmdLineTests/jvmtitests/jvmtitest.jar:/home/jenkins/workspace/Test-extended.functional-JDK10-linux_x86-64/openjdk-tests/TestConfig/scripts/testKitGen/../../../../jvmtest/TestConfig/lib/asm-all.jar" com.ibm.jvmti.tests.util.TestRunner
Time spent starting: 42 milliseconds
Time spent executing: 7667 milliseconds
Test result: FAILED
Output from test:
 [OUT] *** Testing [1/2]:	testReOrderingStaticFields
 [OUT] *** Test took 871 milliseconds
 [OUT] OK
 [OUT] 
 [OUT] *** Testing [2/2]:	testStaticFieldIDsAfterRedefine
 [ERR] Unhandled exception
 [ERR] Type=Segmentation error vmState=0x00000000
 [ERR] J9Generic_Signal_Number=00000004 Signal_Number=0000000b Error_Value=00000000 Signal_Code=00000001
 [ERR] Handler1=00007FCE9F7ECC10 Handler2=00007FCE9EBE05C0 InaccessibleAddress=0000000000000000
 [ERR] RDI=00007FCDF9D1EB10 RSI=0000000000000020 RAX=0000000000000000 RBX=00007FCE9853AA48
 [ERR] RCX=00007FCE983CD300 RDX=0000000000000000 R8=00007FCE985F4FC0 R9=0000000000000001
[...]

PR #2569, opened 2018-08-09
Clear resolveFrameFlags after each frame walk

gacholio:
Clear resolveFrameFlags in the interpreter stack walker after reporting
the frame. This is consistent with the JIT stack walker and avoids
leaving invalid flags set on entry to the JIT walker in very rare cases.

Also fix a small error in the resumeable stack walker, which has no
effect on the only current consumer of the feature.

Fixes: #2472

### DIFF
--- a/runtime/vm/swalk.c
+++ b/runtime/vm/swalk.c
@@ -346,8 +346,9 @@ UDATA  walkStackFrames(J9VMThread *currentThread, J9StackWalkState *walkState)
 		if (walkFrame(walkState) != J9_STACKWALK_KEEP_ITERATING) {
 			goto terminationPoint;
 		}
-		walkState->previousFrameFlags = walkState->frameFlags;
 resumeInterpreterWalk:
+		walkState->previousFrameFlags = walkState->frameFlags;
+		walkState->resolveFrameFlags = 0;
 
 		/* Call the JIT walker if the current frame is a transition from the JIT to the interpreter */
 


runtime/vm/swalk.c
```diff
@@ -346,8 +346,9 @@ UDATA  walkStackFrames(J9VMThread *currentThread, J9StackWalkState *walkState)
 		if (walkFrame(walkState) != J9_STACKWALK_KEEP_ITERATING) {
 			goto terminationPoint;
 		}
-		walkState->previousFrameFlags = walkState->frameFlags;
 resumeInterpreterWalk:
+		walkState->previousFrameFlags = walkState->frameFlags;
+		walkState->resolveFrameFlags = 0;
```
DanHeidinga: Is the intention that the `resolveFrameFlags` are only set if the current frame is a resolveFrame?  And otherwise cleared?

Overall, I have to wonder why we bother with a separate resolveFrameFlags in that case.

DanHeidinga:
Jenkins test extended xlinux jdk10

runtime/vm/swalk.c
gacholio: Agreed - a lot of this is historical cruft, but fixing that is for a different day.

resolveFrameFlags is set when the previously-walked frame is a resolve frame.  Generally only applicable if the current frame is a JIT frame, but this seemed the safest and most "correct" way to fix it, since there clearly is one case where the resolve frame is not on top of a JIT frame.

DanHeidinga: Thanks for confirming that I understood the fix :)

PR #2572, opened 2018-08-09
Fix incorrectly-held mutex

gacholio:
The JVMTI "check for data breakpoint" hook exits with the JVMTI global
mutex held in cases where the compilation is forced to fail.


DanHeidinga:

PR #2578, opened 2018-08-09
Gencon fixed heap wrtbar range check optimization on X86

yanluo7:
On X86, the current gencon JIT write barrier instruction sequence requires range check to see if an object is in the tenure space or not. It is implemented as a subtract instruction to calculate the offset of the object pointer (`offset = object - heapBase`), followed by a compare instruction that compares `offset` with `heapSize`. If `heapSize` and `heapBase` are fixed (e.g. `-Xmx -Xms` are set at the same value), we can optimize this sequence by using one instruction that simply compares the object with `(heapBase + heapSize)`, whose value can be materialized at compile time and does not change during the lifetime of the JVM process. 


yanluo7:
Performance implication:
For day trader where we specify `-Xmx -Xms` to be the same value, we saw a 0.5% improvement in throughput.

andrewcraik:

Issue #2577, opened 2018-08-09
Java 11 interpreter sometimes skips type test for super invokespecial in interface method


jdmpapin:
Super `invokespecial` requires that the receiver be an instance of the current class (i.e. the class containing the `invokespecial` bytecode instruction). When the current class is an interface, this is not guaranteed by bytecode verification. Instead, the `invokespecial` instruction dynamically checks that the receiver implements the current interface, and if not, it throws `IllegalAccessError`.

The attached [`BadSuperInInterface`](https://github.com/eclipse/openj9/files/2275859/BadSuperInInterface.zip) should print `true`, `false`, `true`, and then throw `IllegalAccessError`. I get the correct result from OpenJ9 if it is built as Java 8, or if it is built as Java 11 and JIT options force the `nosey` method to be compiled. But if `nosey` is interpreted in Java 11, I don't get an exception.

The problem appears to have been introduced by #2434. If I build with that commit reverted, I get the exception again.

I tried to reduce the test case further with [`InterfaceInvokeSpecialTest`](https://github.com/eclipse/openj9/files/2275862/InterfaceInvokeSpecialTest.class.zip), but this one correctly throws `IllegalAccessError`, so the type test is bypassed only sometimes.

gacholio:
@fengxue-IS Can you please take a look at this?  Suggest removing bits and pieces of the change to see what's causing the issue.

jdmpapin:
Yes, for a non-null receiver we need `(receiver instanceof CurrentClass)`, which includes instances of subclasses.

fengxue-IS:
I ll take a look into this, @jdmpapin thanks for providing the test class

fengxue-IS:
Iooking through the code changes made in #2434, this behavior should be due to we are now forcing shared ConstantPool entries between `invokeSpecial` and `invokeVirtual` to be split.

The patched bytecode `JBinvokespecialspilt` in interpreter is executing the method without the access check on the receiver class.

gacholio:
Good catch - sounds like this was already a problem in some circumstances, and just made more frequent by your change.

gacholio:
If possible, please try and common as much of the code as you can between invokespecial and invokespecialsplit. At a glance, once _sendMethod has been fetched, it should be very similar code.

PR #2583, opened 2018-08-11
Add access check to invokespecialsplit

fengxue-IS:
Ensure the receiver class is sub class of the current class for
`JBinvokespecialsplit` bytecode to be consistent with `invokespecial`

Close: #2577 

### DIFF
--- a/runtime/vm/BytecodeInterpreter.hpp
+++ b/runtime/vm/BytecodeInterpreter.hpp
@@ -6495,11 +6495,11 @@ done:
 						restoreGenericSpecialStackFrame(REGISTER_ARGS);
 						if (immediateAsyncPending()) {
 							rc = GOTO_ASYNC_CHECK;
-							goto exit;
+							goto done;
 						}
 						if (VM_VMHelpers::exceptionPending(_currentThread)) {
 							rc = GOTO_THROW_CURRENT_EXCEPTION;
-							goto exit;
+							goto done;
 						}
 					}
 					/* Check the receiver class is the subtype of the current class [= interface] */
@@ -6519,7 +6519,7 @@ done:
 				}
 			}
 		}
-exit:
+done:
 		return rc;
 	}
 
@@ -6534,7 +6534,8 @@ exit:
 		U_16 cpIndex = *(U_16 *)(J9ROMCLASS_SPECIALSPLITMETHODREFINDEXES(clazz->romClass) + splitTableIndex);
 		J9RAMSpecialMethodRef *ramMethodRef = (J9RAMSpecialMethodRef*)ramConstantPool + cpIndex;
 		/* argCount was initialized when we initialized the class (i.e. it is non-volatile), so no memory barrier is required */
-		if (0 == _sp[ramMethodRef->methodIndexAndArgCount & 0xFF]) {
+		j9object_t receiver = ((j9object_t *)_sp)[ramMethodRef->methodIndexAndArgCount & 0xFF];
+		if (NULL == receiver) {
 			rc = THROW_NPE;
 		} else {
 			_sendMethod = clazz->specialSplitMethodTable[splitTableIndex];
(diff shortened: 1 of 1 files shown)


runtime/vm/BytecodeInterpreter.hpp
```diff
@@ -6534,7 +6534,8 @@ done:;
 		U_16 cpIndex = *(U_16 *)(J9ROMCLASS_SPECIALSPLITMETHODREFINDEXES(clazz->romClass) + splitTableIndex);
 		J9RAMSpecialMethodRef *ramMethodRef = (J9RAMSpecialMethodRef*)ramConstantPool + cpIndex;
 		/* argCount was initialized when we initialized the class (i.e. it is non-volatile), so no memory barrier is required */
-		if (0 == _sp[ramMethodRef->methodIndexAndArgCount & 0xFF]) {
+		j9object_t receiver = ((j9object_t *)_sp)[ramMethodRef->methodIndexAndArgCount & 0xFF];
+		if (0 == receiver) {
```
gacholio: Please use NULL rather than 0 for j9object_t.

runtime/vm/BytecodeInterpreter.hpp
```diff
@@ -6546,14 +6547,37 @@ done:;
...
+				if (!instanceOfOrCheckCast(receiverClass, currentClass)) {
+					buildMethodFrame(REGISTER_ARGS, _sendMethod, 0);
+					updateVMStruct(REGISTER_ARGS);
+					setIllegalAccessErrorReceiverNotSameOrSubtypeOfCurrentClass(_currentThread, receiverClass, currentClass);
+					VMStructHasBeenUpdated(REGISTER_ARGS);
+					rc = GOTO_THROW_CURRENT_EXCEPTION;
+				}
+			}
 		}
+exit:
```
gacholio: To keep things similar to the rest of the file, please use "done" as the terminal label.

fengxue-IS: should I also change the label for `invokespecial` as well then (these are the same)?

runtime/vm/BytecodeInterpreter.hpp
fengxue-IS: fixed

runtime/vm/BytecodeInterpreter.hpp
gacholio: Please - this problem will go away when you merge the two invokespecials together.

fengxue-IS:
fixed the label for invokespecial to match rest of the code.

The differences between `invokespecial` and `invokespecialsplit` checks is the location initializing `currentClass` var, would it be better to use a common helper for this (not sure if this may cause performance issue)?

gacholio:
I have an idea for some better commoning, so you can leave that until later.

runtime/vm/BytecodeInterpreter.hpp
```diff
@@ -6546,14 +6547,37 @@ done:;
...
+					goto done;
 				} else if (VM_VMHelpers::exceptionPending(_currentThread)) {
 					rc = GOTO_THROW_CURRENT_EXCEPTION;
+					goto done;
 				} else {
 					profileCallingMethod(REGISTER_ARGS);
 					_sendMethod = method;
 				}
 			}
+
```
gacholio: It seems you're missing a bunch of the code from invokespecial here - fetching host class and seeing if it's an interface.

gacholio:
Also, looking at initialSpecialMethod, it's missing all the checks. Please try making it re-run the current bytecode instead of running the method after successful resolution (just change the default rc).

fengxue-IS:
For methods that require the check, it is resolved inline within `invokespecial` call, so I don't think we need to re-run the bytecode.

another option would be to detect the unresolved case early during `invokespecial` (so remove the inline resolution) and then have `initialSpecialMethod` to re-run the bytecode

gacholio:
Yep, you're right.  I'm not sure if there is an issue with the ordering of the checks, but you certainly haven't introduced any new ones. I'll launch some testing. Please remove the WIP.

gacholio:
jenkins test sanity jdk8,jdk10

gacholio:
jenkins test sanity xlinux jdk8,jdk10

gacholio:
Failure in jdk8 testing seems unlikely, trying again

gacholio:

Issue #2562, opened 2018-08-08
[Condy] Allow invokeWithArguments of varargs methods to absorb extra args into trailing array


DanHeidinga:
```
To relax length restrictions on bootstrap specifiers, the language which defines the invocation of bootstrap methods will be adjusted (with complete backward compatibility) to allow variable arity (ACC_VARARGS) bootstrap methods to absorb, into their trailing arguments, all remaining static arguments, even if there are 2^16-1 of them. (The class-file format already allows this, although there is no way to read over-long bootstrap argument lists.) For consistency, the invokeWithArguments methods of MethodHandle will also be expanded in this way, if the target method has variable arity. In this way the invocation of bootstrap methods can be specified in terms of the weakly typed methods invokeWithArguments and invoke, just as today it is specified in terms of invoke alone.
```

BSM invocation may need to be updated as well if there is an explicit check.  Currently it uses invokeWithArguments().

PR #2584, opened 2018-08-11
[Condy] Update invokeWithArguments varargs

theresa-m:
Allow invokeWithArguments of varargs methods to absorb extra args into trailing array

Fixes: #2562

### DIFF
--- a/jcl/src/java.base/share/classes/java/lang/invoke/VarargsCollectorHandle.java
+++ b/jcl/src/java.base/share/classes/java/lang/invoke/VarargsCollectorHandle.java
@@ -1,6 +1,6 @@
 /*[INCLUDE-IF Sidecar17]*/
 /*******************************************************************************
- * Copyright (c) 2009, 2009 IBM Corp. and others
+ * Copyright (c) 2009, 2018 IBM Corp. and others
  *
  * This program and the accompanying materials are made available under
  * the terms of the Eclipse Public License 2.0 which accompanies this
@@ -23,6 +23,11 @@
 package java.lang.invoke;
 
 import java.lang.invoke.MethodHandles.Lookup;
+/*[IF Java11]*/
+import java.lang.reflect.Array;
+import java.util.Arrays;
+import java.util.Objects;
+/*[ENDIF]*/
 
 import com.ibm.oti.util.Msg;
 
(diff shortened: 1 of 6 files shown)


jcl/src/java.base/share/classes/java/lang/invoke/VarargsCollectorHandle.java
```diff
@@ -99,13 +101,39 @@ static MethodType varargsCollectorType(MethodType nextType, Class<?> arrayType)
...
-			MethodHandle mh = IWAContainer.getMH(args.length);
-			return mh.invokeExact((MethodHandle)this, args);
+		int argsLength = 0;
+		if (args != null) {
+			argsLength = args.length;
+		}
+		if (argsLength < 253) {
+			MethodHandle mh = IWAContainer.getMH(argsLength);
+			return mh.invokeExact((MethodHandle) this, args); 
+		/*[IF Java10]*/
```
DanHeidinga: I think this should be Java11, not 10

jcl/src/java.base/share/classes/java/lang/invoke/VarargsCollectorHandle.java
```diff
@@ -99,13 +101,39 @@ static MethodType varargsCollectorType(MethodType nextType, Class<?> arrayType)
...
+			/* special handling is required if arg count exceeds the allowed number of arguments 
+			 * (254 + trailing array in this case). The additional arguments will be
+			 * stored in the trailing array and the MethodHandle can be treated as fixed arity.
+			 */
+			int mhLength = type.parameterCount();
+			int numTrailingArgs = argsLength - mhLength + 1;
+			Object trailingArgs = Array.newInstance(this.arrayType.getComponentType(), numTrailingArgs);
+			MethodHandle arraySetter = MethodHandles.arrayElementSetter(this.arrayType);
+			for (int i = 0; i < numTrailingArgs; i++) {
+				arraySetter.invoke(trailingArgs, i, args[mhLength - 1 + i]);
```
DanHeidinga: This seems like an expensive way to set the args.  Is there a reason this has to be done with `invoke`?

jcl/src/java.base/share/classes/java/lang/invoke/VarargsCollectorHandle.java
theresa-m: it seems to have been an enhancement in 10 that we missed https://docs.oracle.com/javase/10/docs/api/java/lang/invoke/MethodHandle.html#invokeWithArguments(java.lang.Object...)

jcl/src/java.base/share/classes/java/lang/invoke/VarargsCollectorHandle.java
theresa-m: By using invoke the arguments are cast to the arrayType using "asType conversions", or the proper error is thrown if they cant be.

theresa-m:
It looks like `MethodHandle.resolveInvokeDynamic` is okay as is.

theresa-m:
Added additional test comments & formatting.

theresa-m:
WIP pending internal test verification

theresa-m:
tests are passing

jcl/src/java.base/share/classes/java/lang/invoke/VarargsCollectorHandle.java
```diff
@@ -99,13 +104,39 @@ static MethodType varargsCollectorType(MethodType nextType, Class<?> arrayType)
...
+			Class<?> trailingArrayType = this.arrayType.getComponentType();
+			Object trailingArgs = Array.newInstance(trailingArrayType, numTrailingArgs);
+			MethodHandle arraySetter = MethodHandles.arrayElementSetter(this.arrayType);
+			for (int i = 0; i < numTrailingArgs; i++) {
+				arraySetter.invoke(trailingArgs, i, args[mhLength - 1 + i]);
+			}
+			args = Arrays.copyOf(args, mhLength);
+			args[mhLength - 1] = trailingArgs;
+			return this.asFixedArity().invokeWithArguments(args);
+		} else {
```
DanHeidinga: We can avoid one set of IF Java11 comments by writing the code this way:
```
} else 
/*[ENDIF]*/
{
    ... code here
}
```

test/functional/Jsr292/src_110/com/ibm/j9/jsr292/MethodHandleTestExtra.java
```diff
@@ -0,0 +1,71 @@
+/*******************************************************************************
```
DanHeidinga: If this contains tests only applicable to JDK11+, can the name of the file reflect that?

test/functional/Jsr292/testng.xml
```diff
@@ -37,6 +37,7 @@
 			<class name="com.ibm.j9.jsr292.MethodTypeTests"/>
 			<class name="com.ibm.j9.jsr292.MethodHandleProxiesTest"/>
 			<class name="com.ibm.j9.jsr292.MethodHandleTest"/>
+			<class name="com.ibm.j9.jsr292.MethodHandleTest11plus"/>
```
llxia: - I suspect test will fail when testing against SE80 because of missing `com.ibm.j9.jsr292.MethodHandleTest11plus` in SE80 test material.
- Maybe we can try to create a separate `<test>` for new test so that we can avoid to add dummy files.
- If possible, I would prefer not to include jdk version in the test name - MethodHandleTest11plus

theresa-m:
@llxia ready for another review. tests are passing for me on 8,10,11

llxia:
Jenkins test extended plinux all

llxia:
Jenkins test extended plinux jdk8,jdk10

DanHeidinga:
@llxia This looks good to me and builds are passing.  Any further concerns or can this be merged?

PR #2590, opened 2018-08-13
Fix X86 32-bit opcode bug in wrtbar optimization

yanluo7:
Fixed CMPRegImm op code that was overlooked on 32-bit.


andrewcraik:
Jenkins tests sanity xlinux32,win32

andrewcraik:
Jenkins test sanity xlinux,win32

andrewcraik:
all tests passed and fix to an obvious bug so I'm merging this one to avoid broken builds.

PR #2592, opened 2018-08-13
Add a comment for instanceof/checkcast inlining X86

### LINKED ISSUE #2559: Add a comment to explain why the equality test can be skipped for non-abstract classes
Add a comment to explain why, on line 4718 of J9TreeEvalutor.cpp for the x codegen, the equality test can be skipped for non-abstract classes but is required for TR_checkAssignable? Having the explanation in the code would make it clear to future readers. The need for this was identified by @DanHeidinga. The code change in the PR fixes a major functional issue so I (@andrewcraik) will merge the change, but this item should be looked at by @0dvictor with a follow-on PR to add the necessary comments.

0dvictor:
Added a comment to explain why the equality test can be skipped for abstract classes but is required for TR_checkAssignable.

Closes #2559

### DIFF
--- a/runtime/compiler/x/codegen/J9TreeEvaluator.cpp
+++ b/runtime/compiler/x/codegen/J9TreeEvaluator.cpp
@@ -4977,6 +4977,9 @@ inline void generateInlinedCheckCastOrInstanceOfForClass(TR::Node* node, TR_Opaq
    // Equality test
    if (!fej9->isAbstractClass(clazz) || node->getOpCodeValue() == TR::icall/*TR_checkAssignable*/)
       {
+      // For instanceof and checkcast, LHS is obtained from an instance, which cannot be abstract or interface;
+      // however, LHS for TR_checkAssignable may be abstract or interface as it may be an arbitrary class;
+      // therefore, equality test can be safely skipped for instanceof and checkcast.
       if (use64BitClasses)
          {
          generateRegMemInstruction(CMP8RegMem, node, j9class, generateX86MemoryReference(clazzData, cg), cg);


runtime/compiler/x/codegen/J9TreeEvaluator.cpp
```diff
@@ -4977,6 +4977,9 @@ inline void generateInlinedCheckCastOrInstanceOfForClass(TR::Node* node, TR_Opaq
    // Equality test
    if (!fej9->isAbstractClass(clazz) || node->getOpCodeValue() == TR::icall/*TR_checkAssignable*/)
       {
+      // For instanceof and checkcast, LHS is obtained from an instance, which cannot be abstract or interface;
+      // however, LHS for TR_checkAssignable may be abstract or interface as it may be an arbitrary class;
+      // therefore, equality test can be safely skipped for instanceof and checkcast.
```
DanHeidinga: @0dvictor I'm not sure I follow this code and comment yet.  

The condition says:
`if (!fej9->isAbstractClass(clazz) || node->getOpCodeValue() == TR::icall/*TR_checkAssignable*/)`
If `class` is *not* abstract or it's from a `TR_checkAssignable` node, then we need to do do the equality check but the comment says that we can skip the equality check for instanceof/checkcast.

Unless I'm missing something, the comment and the code don't feel like they are in sync.  

Also, don't instanceof/checkcast need the equality check?  Or is subsumed into the superclass test below?  How else would the check of instances exact class be caught?

DanHeidinga: FYI @andrewcraik

andrewcraik: Quite right my bad - I missed that - I think the confusion is in the end of the sentence. The code I believe is right. @0dvictor can you make the appropriate correction?

DanHeidinga: @0dvictor @andrewcraik Has a PR been opened with the updated comment yet?

0dvictor: @DanHeidinga Sorry that I missed the discussion in this merged PR. I've opened #2698 to update the comments.

PR #2595, opened 2018-08-13
Revert "Build JNI IFA switching helpers as C calls"

fjeremic:
Reverts eclipse/openj9#2044 as it is causing issues on z/OS with Atomic-Free JNI. This path has not been tested well at Eclipse since it was only enabled after this feature was merged.


fjeremic:
@NigelYiboYu FYI.

PR #2607, opened 2018-08-14
remove final field modification flag

fengxue-IS:
The `J9ClassHasIllegalFinalFieldModifications` flag will be set by JIT callback


fengxue-IS:
@gacholio Can you take a look

gacholio:
Verified that the JIT does not currently depend on the bit being set. Trivial enough to merge without testing,

PR #2609, opened 2018-08-14
Update nestmates jni calls for new spec

tajila:
Update nestmates jni calls for new spec

Nestmates JNI calls now triggered classloading for the hostClass
and for the nest members. This PR is based on Talia's work in
https://github.com/eclipse/openj9/pull/1501.

Also added new testcases.

### DIFF
--- a/jcl/src/java.base/share/classes/java/lang/Class.java
+++ b/jcl/src/java.base/share/classes/java/lang/Class.java
@@ -4360,7 +4360,7 @@ static byte[] getExecutableTypeAnnotationBytes(Executable exec) {
 }
 /*[ENDIF] Sidecar19-SE*/
 
-/*[IF Valhalla-NestMates]*/
+/*[IF Java11]*/
 /**
  * Answers the host class of the receiver's nest.
  *
@@ -4408,5 +4408,5 @@ public boolean isNestmateOf(Class<?> that) {
 public Class<?>[] getNestMembers() {
 	return getNestMembersImpl();
 }
-/*[ENDIF] Valhalla-NestMates*/
+/*[ENDIF] Java11 */
 }
--- a/runtime/jcl/common/java_lang_Class.cpp
+++ b/runtime/jcl/common/java_lang_Class.cpp
@@ -1788,19 +1788,45 @@ storePDobjectsHelper(J9VMThread* vmThread, J9Class* arrayClass, J9StackWalkState
 
 #if defined(J9VM_OPT_VALHALLA_NESTMATES)
 static void
-setNestmatesIncompatibleClassChangeError(J9VMThread *vmThread, J9Class *nestMember, J9Class *nestHost)
+setNestmatesIncompatibleClassChangeError(J9VMThread *vmThread, J9Class *nestMember, J9Class *nestHost, IDATA errorCode)
 {
-	J9JavaVM *vm = vmThread->javaVM;
-	J9InternalVMFunctions *vmFuncs = vm->internalVMFunctions;
+	J9InternalVMFunctions *vmFuncs = vmThread->javaVM->internalVMFunctions;
 	PORT_ACCESS_FROM_VMC(vmThread);
 
 	J9UTF8 *nestMemberName = J9ROMCLASS_CLASSNAME(nestMember->romClass);
-	J9UTF8 *nestHostName = J9ROMCLASS_CLASSNAME(nestHost->romClass);
+	J9UTF8 *nestHostName = NULL;
+	UDATA exception = J9VMCONSTANTPOOL_JAVALANGINCOMPATIBLECLASSCHANGEERROR;
+	const char *nlsTemplate = NULL;
 
-	const char *nlsTemplate = j9nls_lookup_message(
-			J9NLS_DO_NOT_PRINT_MESSAGE_TAG | J9NLS_DO_NOT_APPEND_NEWLINE,
-			J9NLS_JCL_NEST_MEMBER_CLAIMS_DIFFERENT_NEST_HOST,
-			"Nest member %2$.*1$s in %4$.*3$s declares a different nest host of %6$.*5$s");
+	if (NULL == nestHost) {
+		nestHostName = J9ROMCLASS_NESTHOSTNAME(nestMember->romClass);
+	} else {
+		nestHostName = J9ROMCLASS_CLASSNAME(nestHost->romClass);
+	}
+
+	switch (errorCode) {
+	case J9_VISIBILITY_NEST_HOST_LOADING_FAILURE_ERROR:
+		exception = J9VMCONSTANTPOOL_JAVALANGNOCLASSDEFFOUNDERROR;
+		nlsTemplate = j9nls_lookup_message(
+				J9NLS_DO_NOT_PRINT_MESSAGE_TAG | J9NLS_DO_NOT_APPEND_NEWLINE,
+				J9NLS_JCL_NESTMATES_CLASS_FAILED_TO_LOAD,
+				NULL);
+		break;
+	case J9_VISIBILITY_NEST_HOST_DIFFERENT_PACKAGE_ERROR:
+		nlsTemplate = j9nls_lookup_message(
+				J9NLS_DO_NOT_PRINT_MESSAGE_TAG | J9NLS_DO_NOT_APPEND_NEWLINE,
+				J9NLS_JCL_NEST_HOST_HAS_DIFFERENT_PACKAGE,
+				NULL);
+		break;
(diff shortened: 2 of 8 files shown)


tajila:
@JasonFengJ9 can you please review these changes

jcl/src/java.base/share/classes/java/lang/Class.java
```diff
@@ -4360,7 +4360,7 @@ ConstantPool getConstantPool() {
 }
 /*[ENDIF] Sidecar19-SE*/
 
-/*[IF Valhalla-NestMates]*/
+/*[IF Java11]*/
```
JasonFengJ9: Is `pConfig VALHALLA` still supported assuming this change removes the natives affected from that configuration?
How about the check decorated by `Valhalla-NestMates` within `j.l.i.MethodHandles.checkAccess()`?

tajila: Because we are not using the UMA spec files to turn on nestmates those flags will never be turned on. 

Ill have to go and fix the other nestmates flags in the jcl. Ill do that in a separate PR.

runtime/jcl/common/java_lang_Class.cpp
```diff
@@ -1788,19 +1788,45 @@ storePDobjectsHelper(J9VMThread* vmThread, J9Class* arrayClass, J9StackWalkState
...
+				J9NLS_DO_NOT_PRINT_MESSAGE_TAG | J9NLS_DO_NOT_APPEND_NEWLINE,
+				J9NLS_JCL_NEST_HOST_HAS_DIFFERENT_PACKAGE,
+				NULL);
+		break;
+	case J9_VISIBILITY_NEST_MEMBER_NOT_CLAIMED_ERROR:
+		nlsTemplate = j9nls_lookup_message(
+				J9NLS_DO_NOT_PRINT_MESSAGE_TAG | J9NLS_DO_NOT_APPEND_NEWLINE,
+				J9NLS_JCL_NEST_MEMBER_NOT_CLAIMED_BY_NEST_HOST,
+				NULL);
+		break;
```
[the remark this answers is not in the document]
tajila: fixed

runtime/jcl/common/java_lang_Class.cpp
```diff
@@ -1815,19 +1841,32 @@ setNestmatesIncompatibleClassChangeError(J9VMThread *vmThread, J9Class *nestMemb
 				J9UTF8_LENGTH(nestMemberName), J9UTF8_DATA(nestMemberName));
```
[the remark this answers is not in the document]
tajila: added null check

runtime/jcl/common/java_lang_Class.cpp
```diff
@@ -1815,19 +1841,32 @@ setNestmatesIncompatibleClassChangeError(J9VMThread *vmThread, J9Class *nestMemb
...
 	}
 
-	vmFuncs->setCurrentExceptionUTF(vmThread, J9VMCONSTANTPOOL_JAVALANGINCOMPATIBLECLASSCHANGEERROR, msg);
+	vmFuncs->setCurrentExceptionUTF(vmThread, exception, msg);
 	j9mem_free_memory(msg);
 }
+#endif /* defined(J9VM_OPT_VALHALLA_NESTMATES) */
```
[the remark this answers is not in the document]
tajila: this is done to work around build issues for older java versions

jcl/src/java.base/share/classes/java/lang/Class.java
tajila: fixed in this PR

runtime/jcl/common/java_lang_Class.cpp
```diff
@@ -1788,46 +1788,90 @@ storePDobjectsHelper(J9VMThread* vmThread, J9Class* arrayClass, J9StackWalkState
...
 	J9Class *nestHost = clazz->nestHost;
+
+	if (NULL == nestHost) {
+		if (J9_VISIBILITY_ALLOWED == vmFuncs->loadAndVerifyNestHost(currentThread, clazz, J9_LOOK_NO_THROW)) {
+			nestHost = clazz->nestHost;
+		} else {
+			/* If there is a failure loading or accessing the nest host, or if this class or interface does
+			 * not specify a nest, then it is considered to belong to its own nest and this is returned as
+			 * the host */
+			nestHost = clazz;
```
JasonFengJ9: Should `clazz->nestHost` be updated to avoid repeating call to `loadAndVerifyNestHost` in this method is invoked later with same `clazz`?

runtime/jcl/common/java_lang_Class.cpp
```diff
@@ -1837,29 +1881,51 @@ Java_java_lang_Class_getNestHostImpl(JNIEnv *env, jobject recv)
...
-	J9ROMClass *romHostClass = nestHost->romClass;
-	U_16 nestMemberCount = romHostClass->nestMemberCount;
 
-	/*  Grab the Class<?> class for result object size */
-	J9Class *jlClass = J9VMJAVALANGCLASS(vm);
-	J9Class *arrayClass = fetchArrayClass(currentThread, jlClass);
+	if (NULL == nestHost) {
+		nestLoadFailure = vmFuncs->loadAndVerifyNestHost(currentThread, clazz, FALSE);
+		if (J9_VISIBILITY_ALLOWED != nestLoadFailure) {
+			nestMember = clazz;
```
JasonFengJ9: Just wondering is this right time to exit with exception? `getNestHost` should return a nestHost or itself if there is an error.

JasonFengJ9:
I opened an issue related to nestmate existing code (not this PR) at https://github.com/eclipse/openj9/issues/2630.

jcl/src/java.base/share/classes/java/lang/Class.java
```diff
@@ -4408,5 +4408,5 @@ public boolean isNestmateOf(Class<?> that) {
 public Class<?>[] getNestMembers() {
```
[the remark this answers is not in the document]
tajila: Another PR will address the JCL changes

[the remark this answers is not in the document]
tajila: https://github.com/eclipse/openj9/pull/1531

runtime/jcl/common/java_lang_Class.cpp
tajila: No, if we set it we will make it a valid nesthost. This means that if we do an access check operation it won't trigger a failure.

runtime/jcl/common/java_lang_Class.cpp
tajila: this is getNestMember not getNestHost. GetNestMember must throw exceptions if it fails to load every member of the nest

JasonFengJ9: Tobi said `RI` has same behaviours when running the tests in this PR as `OpenJ9`, so `GetNestMember` throws exceptions if it fails to load every member of the nest including nest host.

tajila:
@llxia can you please review the test changes

llxia:

llxia:
PR builds failed due to git issue. Even without git issue, jdk11 tests will fail during test compilation due to native test libs location changed.

Since these tests are sanity, did these tests pass in grinder or personal build?

tajila:
@llxia I manually tested it with a CCM build

runtime/jcl/common/java_lang_Class.cpp
```diff
@@ -1881,29 +1947,43 @@ Java_java_lang_Class_getNestMembersImpl(JNIEnv *env, jobject recv)
...
-				J9JAVAARRAYOFOBJECT_STORE(currentThread, resultObject, i + 1, J9VM_J9CLASS_TO_HEAPCLASS(nestMember));
 			}
+			J9JAVAARRAYOFOBJECT_STORE(currentThread, resultObject, i + 1, J9VM_J9CLASS_TO_HEAPCLASS(nestMember));
 		}
 	}
 
 	result = vmFuncs->j9jni_createLocalRef(env, resultObject);
 
 _done:
+	if (J9_VISIBILITY_ALLOWED != nestLoadStatus) {
```
DanHeidinga: This code should be moved into `loadAndVerifyNestHost()`.  Callers who allow `loadAndVerifyNestHost` to throw should expect the ICCE to be pending already in the failure case and shouldn't have to set it themselves.

This makes the contract of `loadAndVerifyNestHost` simpler - an exception will be pending in all failure cases if the caller allows exceptions to be thrown - and simplifies the callers as they don't have to duplicate the exception call everywhere.

Issue #2603, opened 2018-08-14
Add FV mode to stress split bytecodes


gacholio:
Invoke bytecode splitting (different invoke bytecodes targetting the same method ref) doesn't happen very often in the real world.

Add a command-line option to force all possible splits to occur in order to stress the unused codepaths.

PR #2614, opened 2018-08-14
Add FV option to stress bytecode splitting

gacholio:
Add -XX:[+/-]AlwaysSplitBytecodes to force all invokespecial and
invokestatic bytecodes to be split. Clean up BCT_ constants.

Fixes: #2603


runtime/oti/j9nonbuilder.h
```diff
@@ -2030,64 +2030,21 @@ typedef struct J9BCTranslationData {
...
-#define BCT_InitCopyLongs  17
+#define BCT_LittleEndianOutput  0
 #define BCT_BigEndianOutput  1
-#define BCT_ActionReturn  4
-#define BCT_PcFlagLookupSwitch  0x800000
-#define BCT_ErrLocalsMapTooSmall  10
-#define BCT_MajorClassFileVersionMaskShift  24
-#define BCT_ErrIllegalState  6
-#define BCT_InitCopyIntegersW  14
-#define BCT_ErrConstantPoolMapInvalid  4
```
DanHeidinga: Are these really all unused?  I thought some?  all? of the BCT_Err flags were related to verbose verification

gacholio: I searched the codebase for each constant that was removed. I needed a new bit in the flags, and it was less than clear what bits were in use, hence the cleanup.

DanHeidinga: Fair enough.  And the testing passes so I can't argue :)

Issue #1787, opened 2018-04-27
Jenkins: Git clones fail intermittently with passwd file error


AdamBrousseau:
Git clones failing intermittently. This one failed while trying to clone OpenJ9 ([this step](https://github.com/eclipse/openj9/blob/master/buildenv/jenkins/jobs/pull-requests/PullRequest-Sanity-JDK9-linux_ppc-64_cmprssptrs_le#L30)) on the master.

I'll have to look, but I don't think its always on the master that I've seen this.

Console:
```
 > git fetch --tags --progress https://github.com/eclipse/openj9.git +refs/pull/*:refs/remotes/origin/pr/*
ERROR: Error fetching remote repo 'origin'
hudson.plugins.git.GitException: Failed to fetch from https://github.com/eclipse/openj9.git
	at hudson.plugins.git.GitSCM.fetchFrom(GitSCM.java:862)
	at hudson.plugins.git.GitSCM.retrieveChanges(GitSCM.java:1129)
	at hudson.plugins.git.GitSCM.checkout(GitSCM.java:1160)
	at org.jenkinsci.plugins.workflow.steps.scm.SCMStep.checkout(SCMStep.java:113)
	at org.jenkinsci.plugins.workflow.steps.scm.SCMStep$StepExecutionImpl.run(SCMStep.java:85)
	at org.jenkinsci.plugins.workflow.steps.scm.SCMStep$StepExecutionImpl.run(SCMStep.java:75)
	at org.jenkinsci.plugins.workflow.steps.AbstractSynchronousNonBlockingStepExecution$1$1.call(AbstractSynchronousNonBlockingStepExecution.java:47)
	at hudson.security.ACL.impersonate(ACL.java:290)
	at org.jenkinsci.plugins.workflow.steps.AbstractSynchronousNonBlockingStepExecution$1.run(AbstractSynchronousNonBlockingStepExecution.java:44)
	at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:511)
	at java.util.concurrent.FutureTask.run(FutureTask.java:266)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1149)
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:624)
	at java.lang.Thread.run(Thread.java:748)
[...]

PR #2616, opened 2018-08-15
Fix HCRGuardAnalysis

liqunl:
HCRGuardAnalysis uses single bit container as the gen and kill set, thus
when gen/kill is set, kill/gen should be cleared. For exception edges,
gen will flow through the exception edge when there exists one exception
point that can be reached by the gen. The fix also considers non-treetop
nodes that may except.

### DIFF
--- a/runtime/compiler/optimizer/HCRGuardAnalysis.cpp
+++ b/runtime/compiler/optimizer/HCRGuardAnalysis.cpp
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2000, 2016 IBM Corp. and others
+ * Copyright (c) 2000, 2018 IBM Corp. and others
  *
  * This program and the accompanying materials are made available under
  * the terms of the Eclipse Public License 2.0 which accompanies this
@@ -33,6 +33,7 @@
 #include "infra/Assert.hpp"                // for TR_ASSERT
 #include "infra/BitVector.hpp"             // for TR_BitVector
 #include "infra/Checklist.hpp"             // for TR::NodeChecklist
+#include "optimizer/FearPointAnalysis.hpp"
 
 static bool containsPrepareForOSR(TR::Block *block)
    {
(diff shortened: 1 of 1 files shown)


andrewcraik:
since I've looked at this once already offline - a look from @jdmpapin is probably a good idea.

liqunl:
@jdmpapin ping?

runtime/compiler/optimizer/HCRGuardAnalysis.cpp
```diff
@@ -88,80 +88,99 @@ bool TR_HCRGuardAnalysis::shouldSkipBlock(TR::Block *block)
...
-            _regularKillSetInfo[currentBlock->getNumber()]->setAll(getNumberOfBits());
-            if (exceptingTTSeen)
-               _exceptionKillSetInfo[currentBlock->getNumber()]->setAll(getNumberOfBits());
+            _regularKillSetInfo[blockId]->setAll(numBits);
+            _regularGenSetInfo[blockId]->resetAll(numBits);
             }
          }
+
+      bool canNodeRaiseException = false;
+      if (ttNode->getOpCode().canRaiseException())
```
jdmpapin: When the current tree could both yield and throw, I think exception edges should always need to gen, even at OSR points where the analysis kills. It's possible for the tree to yield, allowing assumptions to be invalidated, and then throw afterward. An OSR guard (if one is necessary) would only run after non-exceptional completion of the tree, so it wouldn't stop control from reaching the exception handler.

That said, I'm not sure we need to do anything at all for the exception edges; I thought we had an OSR point at the start of each catch block for this very reason.

runtime/compiler/optimizer/HCRGuardAnalysis.cpp
```diff
@@ -88,80 +88,99 @@ bool TR_HCRGuardAnalysis::shouldSkipBlock(TR::Block *block)
...
          {
-         TR_VirtualGuard *guardInfo = comp()->findVirtualGuardInfo(treeTop->getNode());
+         TR_VirtualGuard *guardInfo = comp()->findVirtualGuardInfo(ttNode);
          if (guardInfo->getKind() != TR_HCRGuard)
             {
-            _regularKillSetInfo[currentBlock->getNumber()]->setAll(getNumberOfBits());
-            if (exceptingTTSeen)
-               _exceptionKillSetInfo[currentBlock->getNumber()]->setAll(getNumberOfBits());
+            _regularKillSetInfo[blockId]->setAll(numBits);
+            _regularGenSetInfo[blockId]->resetAll(numBits);
```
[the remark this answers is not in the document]
liqunl: In order to support killing only on the inlined path, we have to add logic to the analysis and optimizations using the analysis or come up with another data flow to better support our need. Since the taken side usually only has the call with at most some remat trees, optimizing it gives little benefit. Plus there is few optimization that will and can optimize the taken side. As per discussion with @andrewcraik, we decide to avoid the complication and to require optimizations that can generate fear stay away from the taken side.

runtime/compiler/optimizer/HCRGuardAnalysis.cpp
```diff
@@ -88,80 +89,148 @@ bool TR_HCRGuardAnalysis::shouldSkipBlock(TR::Block *block)
...
-      if (treeTop->getNode()->getOpCodeValue() == TR::BBStart)
+      TR::Node* ttNode = treeTop->getNode();
+      TR::Node* osrNode = NULL;
+      isYieldPoint = false;
+      if (ttNode->getOpCodeValue() == TR::BBStart)
          {
-         exceptingTTSeen = false;
          currentBlock = treeTop->getEnclosingBlock();
+         blockId = currentBlock->getNumber();
+         isGen = isKill = hasExceptionGen = hasExceptionKill = false;
```
andrewcraik: I think a comment above this about resetting the walk state would be helpful. I'm not a fan of the chained equals because it makes it much harder to read what this is doing IMO

runtime/compiler/optimizer/HCRGuardAnalysis.cpp
```diff
@@ -88,80 +89,148 @@ bool TR_HCRGuardAnalysis::shouldSkipBlock(TR::Block *block)
...
+            // we require optimizations that can generate fear stay away from taken side such that considering a guard
+            // to be a kill for the taken side is safe.
+            //
+            isKill = true;
+            isGen = false;
+            }
+         }
+
+      bool canNodeRaiseException = false;
+      if (ttNode->getOpCode().canRaiseException())
```
andrewcraik: what about calls under treetop nodes?

andrewcraik: please add {}s

andrewcraik:
a couple of small things in the comments. I'd like @jdmpapin to confirm his concerns are addressed and I think this should then be good to go.

runtime/compiler/optimizer/HCRGuardAnalysis.cpp
```diff
@@ -88,80 +89,154 @@ bool TR_HCRGuardAnalysis::shouldSkipBlock(TR::Block *block)
...
+         // It's possible for the tree to yield, allowing assumptions to be invalidated,
+         // and then throw afterward. An OSR guard (if one is necessary) would only run
+         // after non-exceptional completion of the tree, so it wouldn't stop control from
+         // reaching the exception handler.
+         if (isYieldPoint)
+            hasExceptionGen = true;
+         else
+            {
+            hasExceptionGen = hasExceptionGen || isGen;
+            hasExceptionKill = hasExceptionKill || isKill;
```
jdmpapin: I think we need the state to be "kill" at _all_ exception points in order to kill along the exception path.

If the first interesting point we encounter in the current block is an exception point, then it's possible to invalidate assumptions, enter the block, and follow an exception edge before there has been a chance to run an OSR guard.

liqunl: Added another bool `hasExceptionNotKill` to fix the issue.

liqunl:
@fjeremic Could you help launching the genkins build and merge?

fjeremic:
Jenkins test sanity JDK8

fjeremic:
Jenkins test sanity all JDK8

fjeremic:
Infrastructure failures on the builds due to #1787.

PR #2618, opened 2018-08-15
WIP: Implement j.l.i.MethodHandle.internalMemberName

### LINKED ISSUE #2597: Java 11: j.l.i.MethodHandle.internalMemberName need an implementation
```
java.lang.InternalError: Compile stub invoked! Apart from deliberate reflective access, this should not happen. Please report this to the project so it can be addressed
        at java.lang.invoke.OpenJDKCompileStub.OpenJDKCompileStubThrowError(java.base@11-internal/OpenJDKCompileStub.java:38)
        at java.lang.invoke.MethodHandle.internalMemberName(java.base@11-internal/MethodHandle.java:1168)
        at java.lang.invoke.ConstantBootstraps.getStaticFinal(java.base@11-internal/ConstantBootstraps.java:167)
```

JasonFengJ9:
Implement `j.l.i.MethodHandle.internalMemberName()`

Return `null` for the function not supported by `OpenJ9`

Verified manually that the test doesn't throw `OpenJDKCompileStubThrowError` for this method.

close #2597 

Reviewer: @gacholio 
FYI: @DanHeidinga 

### DIFF
--- a/jcl/src/java.base/share/classes/java/lang/invoke/MethodHandle.java
+++ b/jcl/src/java.base/share/classes/java/lang/invoke/MethodHandle.java
@@ -1276,7 +1276,7 @@ public abstract class MethodHandle {
 	}
 	
 	MemberName internalMemberName() {
-		throw OpenJDKCompileStub.OpenJDKCompileStubThrowError();
+		return null;
 	}
 	
 	BoundMethodHandle rebind() {


gacholio:
Should we be implementing this instead of returning null?

JasonFengJ9:
Ok, looking into it.

JasonFengJ9:
Just realized that returning `null` for `j.l.i.MethodHandle.internalMemberName()` caused NPE within `java.lang.invoke.ConstantBootstraps.getStaticFinal(java.base@11-internal/ConstantBootstraps.java:168)`.
Returning a `MemberName` object requires its APIs to be implemented such as `isFinal()` which requires further implementation of class `MemberName`.  Not sure if that's the direction to go.
Another option would be to implement this new class `j.l.i.ConstantBootstraps` since `Java 11`.

gacholio:
I don't know anything about this API, but it seems like returning null doesn't really make things work, so maybe this PR should just be abandoned.

JasonFengJ9:
All right, moving the discussion to the issue instead.

PR #2622, opened 2018-08-15
Add missing @Deprecated annotations

keithc-ca:
Features whose javadoc says '@deprecated' should also have the '@Deprecated' annotation.


gacholio:
jenkins compile xlinux jdk8

PR #2623, opened 2018-08-15
Fix split bytecode errors in the JIT

gacholio:
- fix incorect cpIndex tagging
- fix invalid use of of cpIndex in invokespecialsplit processing


gacholio:

gacholio:
It's pretty clear that splitting is not tested, so the above sanity test should be sufficient to show no regressions.

andrewcraik:
LGTM and agree with testing statement above

PR #2624, opened 2018-08-15
 Generate PTX code for sort kernels on demand

keithc-ca:
Previously we used `KernelGenerator.java` to generate `SortKernels.cu` which was compiled by `nvcc` to produce `SortKernels.fatbin` with the expectation that would be included in a jar or a module. The compilation by `nvcc` never got enabled in the open. The result is that attempts to use `Maths.sortArray()` yielded `FileNotFoundException(SortKernels.fatbin)`.

Now instead of generating kernels for all four element types, we generate just the ones that are needed which requires less work by the GPU driver to adapt to the actual GPU architecture present on the system. It also removes about 200kB of the footprint of the JDK.

The kernels have a structure almost identical to the CUDA code that was previously generated. The floating-point kernels were fixed to match the behavior of `Arrays.sort()` with respect to infinities, NaNs and -0.0 vs. +0.0.

GPU sort testing is expanded beyond `int[]` to include arrays of `double`, `float` and `long`.

`KernelGenerator.java` and its uses have been removed.


keithc-ca:
Fixed line-ends.

gacholio:
You had said you'd get Devin to have a look. I'll wait for his comments before merge.

Is there any testing for this code?

gacholio:
My real question is if this is tested in PR testing.

keithc-ca:
Yes, I believe PR testing at the 'sanity' level runs that test. See https://ci.eclipse.org/openj9/job/PullRequest-Sanity-JDK8-linux_x86-64_cmprssptrs-OpenJ9/229/testngreports/org.openj9.test.gpu/ (although it needs to run on a system with a GPU to properly exercise it).

gacholio:

gacholio:

gacholio:
jenkins test sanity win jdk8

gacholio:
Hopefully that'll cover it.

keithc-ca:
That should do it, those are all the platforms that support CUDA.

dnakamura:
LGTM

Issue #2347, opened 2018-07-08
Jenkins: Make Slack message colour match build result


AdamBrousseau:
Ie. instead of having all messages red (danger), make messages yellow when build is unstable (failed tests). Also applies to upstream job status.

PR #2625, opened 2018-08-15
Catch aborted et al cases when downstream job doesn't pass

AdamBrousseau:
- If the downstream job is aborted it will not fall into one of
  the existing IF cases. Since we only want to have the parent
  job unstable else fail, we should structure the IF that way.
- Change the IF to UNSTABLE ELSE FAIL

[skip ci]
Issue #2347


AdamBrousseau:
@llxia for review

Issue #2596, opened 2018-08-13
Java 11: j.l.i.MethodHandleNatives.mapLookupExceptionToError need an implementation


JasonFengJ9:
```
"java.lang.InternalError: Compile stub invoked! Apart from deliberate reflective access, this should not happen. Please report this to the project so it can be addressed" was thrown instead
java.lang.InternalError: Compile stub invoked! Apart from deliberate reflective access, this should not happen. Please report this to the project so it can be addressed
        at java.lang.invoke.OpenJDKCompileStub.OpenJDKCompileStubThrowError(java.base@11-internal/OpenJDKCompileStub.java:38)
        at java.lang.invoke.MethodHandleNatives.mapLookupExceptionToError(java.base@11-internal/MethodHandleNatives.java:33)
        at java.lang.invoke.ConstantBootstraps.validateClassAccess(java.base@11-internal/ConstantBootstraps.java:359)
        at java.lang.invoke.ConstantBootstraps.enumConstant(java.base@11-internal/ConstantBootstraps.java:137)
```

PR #2626, opened 2018-08-15
Implement j.l.i.MethodHandleNatives.mapLookupExceptionToError

JasonFengJ9:
Implement `j.l.i.MethodHandleNatives.mapLookupExceptionToError`

Mapped incoming `ReflectiveOperationException` to specific `LinkageError` as per known test failures.

Verified manually that there is no `OpenJDKCompileStubThrowError` caused by `mapLookupExceptionToError`, and the test either pass or failed due to known issues.

close: #2596 

Reviewer: @gacholio 
FYI: @DanHeidinga 

### DIFF
--- a/jcl/src/java.base/share/classes/java/lang/invoke/MethodHandleNatives.java
+++ b/jcl/src/java.base/share/classes/java/lang/invoke/MethodHandleNatives.java
@@ -25,14 +25,23 @@
 
 package java.lang.invoke;
 
-/*
- * Stub class for compilation
- */
-
 class MethodHandleNatives {
 
 	static LinkageError mapLookupExceptionToError(ReflectiveOperationException roe) {
-		throw OpenJDKCompileStub.OpenJDKCompileStubThrowError();
+		String exMsg = roe.getMessage();
+		LinkageError linkageErr;
+		if (roe instanceof IllegalAccessException) {
+			linkageErr = new IllegalAccessError(exMsg);
+		} else if (roe instanceof NoSuchFieldException) {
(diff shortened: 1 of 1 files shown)


jcl/src/java.base/share/classes/java/lang/invoke/MethodHandleNatives.java
```diff
@@ -25,14 +25,22 @@
...
+		if (roe instanceof IllegalAccessException) {
+			linkageErr = new IllegalAccessError(exMsg);
+		} else if (roe instanceof NoSuchFieldException) {
+			linkageErr = new NoSuchFieldError(exMsg);
+		} else if (roe instanceof NoSuchMethodException) {
+			linkageErr = new NoSuchMethodError(exMsg);
+		} else {
+			linkageErr = new IncompatibleClassChangeError(exMsg);
+		}
+		linkageErr.initCause(roe.getCause());
```
gacholio: Should the cause be roe rather than its cause?

JasonFengJ9: My interpretation of this method `mapLookupExceptionToError` is to map (or convert) incoming `ReflectiveOperationException` to a `LinkageError`, hence this `LinkageError` wasn't caused by `ReflectiveOperationException` and they share same causes.
On the other hand, if `roe.getCause()` is `null` then `roe` can be the cause of `linkageErr` instead.

gacholio:
jenkins test sanity xlinux jdk10

PR #2633, opened 2018-08-16
No need to check for "!\" in SharedClassAbstractHelper

### LINKED ISSUE #2632: jar: protocol should not treat \ specially
In https://github.com/eclipse/openj9/pull/2589 changes were made to treat `"!\"` the same as `"!/"`: I don't think this is appropriate. The code is dealing with a URL, which consistently uses forward slashes `/`, even on Windows. Perhaps changes are required elsewhere to map pathnames (that might use backslashes `\`) to `jar:` URLs.

hangshao0:
Back out the check for "!\\" in SharedClassAbstractHelper.java where it is
dealing with a URL.

Fixes #2632


hangshao0:
Reviewer: @keithc-ca

keithc-ca:
Did you also fix the source of the URLs containing backslashes?

hangshao0:
A URL always uses "/". There is no need to check for "!\\", so I am removing the unnecessary check here.

Issue #1779, opened 2018-04-26
Add functional test in OpenJ9 Jenkins


llxia:
We would like to reuse Jenkins files in [AdoptOpenJDK/openjdk-tests](https://github.com/AdoptOpenJDK/openjdk-tests/tree/master/buildenv/jenkins) and add functional test in [OpenJ9 Jenkins](https://ci.eclipse.org/openj9/).

- add functional in OpenJ9 Jenkins
- add it as part of PR build and nightly build (Prereq: PR build needs to have the same flow as nightly https://github.com/eclipse/openj9/issues/1450)

llxia:
https://ci.eclipse.org/openj9/view/Test/job/Test-Sanity-Functional-JDK8-linux_390-64_cmprssptrs/
https://ci.eclipse.org/openj9/view/Test/job/Test-Sanity-Functional-JDK8-linux_ppc-64_cmprssptrs_le/

llxia:
@AdamBrousseau Just to confirm, we will get sdk tar and OPENJ9_SHA from upstream build, correct?

AdamBrousseau:
Yes I believe that is what we decided on.

llxia:
PR: https://github.com/AdoptOpenJDK/openjdk-tests/pull/404

llxia:
Waiting for OpenJ9 Jenkins PR build to change workflow, so we can switch to use Jenkins file in AdoptOpenJDK for PR build.

Issue #1450, opened 2018-03-15
Switch to using Adopt JenkinsFiles for Test


AdamBrousseau:
Currently we have our own [JeknkinsFiles for Test](https://github.com/eclipse/openj9/tree/master/buildenv/jenkins/jobs/tests) purposes ([also here](https://github.com/eclipse/openj9/blob/master/buildenv/jenkins/common/test)). 

We should be pulling the Test Jenkins files from [AdoptOpenJDK/openjdk-tests](https://github.com/AdoptOpenJDK/openjdk-tests/tree/master/buildenv/jenkins). This will benefit us by not having to maintain our own "copies"

This work will require some help and coordination form the test team so we can have a smooth transition.

This may cause us to break out our PR builds into several builds (just like the nightly flow). There are a few issues with this.
1. Moving the SDK around can be slow on some machines (#659)
1. GH-PR-Trigger plugin only works with "1-build". Currently the PR builds run compile & test in a single build/job. The top-level job would be responsible for communicating with GitHub. We could switch to using "Build Status Messages" (see [example](https://github.com/ibmruntimes/openj9-openjdk-jdk8/pull/58#issuecomment-371930546)) instead of the simple "Test result one liner" (which is inaccurate #981).

also related #1409 #1779 #2037

AdamBrousseau:
ref #2037

AdamBrousseau:
#2728 should resolve this

PR #2638, opened 2018-08-16
Remove obsolete test paths for non-Adopt test builds

AdamBrousseau:
- We no longer run JDK9 so we can remove this case.
  If we need to run 9 at some point in the future, test
  jobs will need to be created using the Adopt test scripts.
- Windows is now supported using the Adopt test scripts

[skip ci]
Issue #1450 #1779 AdoptOpenJDK/openjdk-tests#506


AdamBrousseau:
@llxia for review
fyi @vsebe @jdekonin

llxia:
Did not know you are also working on this, but we cannot switch wins atm due to system test failure. https://github.com/eclipse/openj9/issues/2037

llxia:
Also, win32 script still references wrong label, which needs to be fixed.

https://github.com/AdoptOpenJDK/openjdk-tests/blob/2424010ee6835e99b0e9066602929dd2740feb54/buildenv/jenkins/openjdk8_x86-32_windows#L6

llxia:
Systemtest further failed during native test compilation on win: https://github.com/eclipse/openj9-systemtest/issues/38

llxia:
system test failure: https://github.com/AdoptOpenJDK/openjdk-systemtest/issues/165

PR #2639, opened 2018-08-16
Improve z/OS hardware check for Vector Extension Facility

fjeremic:
If the `LOADxx MACHMIG VEF` load parameter is set at IPL time the user
has requested to disable the VEF support in z/OS, as such the CVTVEF
bit will not be enabled for the system.

We currently only check the LE support which says whether LE is
configured to support vector registers. In addition to this check we
also need to validate the CVTVEF bit in the CVT data structure
similarly to other features requiring OS support, such as RI, TX, or GS.

Further we take this opportunity to make improvements in the area:

- Isolate all OS checks into separate functions
  - Use them instead of directly checking CVT bits
- Add documentation on which bits we are checking
- Move checks to corresponding processor areas
- Remove stale comments

### DIFF
--- a/runtime/port/unix/j9sysinfo.c
+++ b/runtime/port/unix/j9sysinfo.c
@@ -702,7 +702,6 @@ testSTFLE(struct J9PortLibrary *portLibrary, uint64_t stfleBit)
 }
 
 #ifdef J9ZOS390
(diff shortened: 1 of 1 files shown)


joransiu:
Jenkins test sanity zlinux

PR #2647, opened 2018-08-17
JIT: Handle resolved private invokevirtual

jdmpapin:
TR_ResolvedJ9Method::getResolvedPossiblyPrivateVirtualMethod() allows
the compiler to query a VirtualMethodRef constant pool entry to
determine whether it has been resolved to a private method. The existing
getResolvedVirtualMethod() refuses to return private methods, so that
the possibility of a private method is only exposed to callers prepared
to deal with it.

When generating intermediate language for an invokevirtual bytecode
instruction, the compiler checks for a private method, and if one is
found, it generates a direct call.

When the inliner encounters an invokevirtual bytecode instruction, it
checks for a private method in order to correctly determine which call
target(s) to consider.

Private invokevirtual is not yet handled in cases where the constant
pool entry is unresolved at compile time. Nor is it handled in AOT
compilations, where for the time being these calls will be treated as
though their constant pool entries are unresolved.

### DIFF
--- a/runtime/compiler/env/j9method.cpp
+++ b/runtime/compiler/env/j9method.cpp
@@ -1622,6 +1622,26 @@ TR_ResolvedRelocatableJ9Method::isUnresolvedMethodHandle(I_32 cpIndex)
    return true;
    }
 
+TR_ResolvedMethod *
+TR_ResolvedRelocatableJ9Method::getResolvedPossiblyPrivateVirtualMethod(
+   TR::Compilation *comp,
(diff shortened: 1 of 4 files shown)


jdmpapin:
Updated the copyright header in `j9method.h

cathyzhyi:
LGTM

fjeremic:
Area of the code this modifies does not have JIT JDK11 support so no need to run the Jenkins tests. The TravisCI is sufficient.

PR #2648, opened 2018-08-17
WIP: JIT: Handle resolved special case invokeinterface

jdmpapin:
This commit adds jitGetImproperInterfaceMethodFromCP() so that the
compiler can query an InterfaceMethod constant pool entry to determine
whether it has been resolved to a method that does not use the usual
interface dispatch, i.e. to a private method or a method of Object. The
existing queries continue to respond in these cases as though the
constant pool entry is unresolved, so that these possibilities are only
exposed to callers prepared to deal with them.

When generating intermediate language for an invokeinterface bytecode
instruction, the compiler checks for these cases and generates a call
using the appropriate kind of dispatch. When not using interface
dispatch, it generates an explicit type check ahead of the call.

When the inliner encounters an invokeinterface bytecode instruction, it
checks for these cases in order to correctly determine which call
target(s) to consider.

Special case invokeinterface is not yet handled in cases where the
constant pool entry is unresolved at compile time. Nor is it handled in
AOT compilations, where for the time being these calls will be treated
as though their constant pool entries are unresolved.


runtime/jit_vm/cthelpers.cpp
```diff
@@ -78,6 +78,48 @@ jitGetInterfaceITableIndexFromCP(J9VMThread *currentThread, J9ConstantPool *cons
...
 
+/* Only returns non-null if the method is not to be dispatched by itable, i.e.
+ * if it is:
+ * - private, using direct dispatch;
+ * - a final method of Object, using direct dispatch; or
+ * - a non-final method of Object, using virtual dispatch.
+ */
+J9Method*
+jitGetImproperInterfaceMethodFromCP(J9VMThread *currentThread, J9ConstantPool *constantPool, UDATA cpIndex)
+	{
```
gacholio: Nitpick: VM files don't indent the opening and closing function braces.

jdmpapin:
From discussion with @gacholio, this needs to be merged before #2006 in order to avoid breaking `invokeinterface` bytecode instructions that target non-final methods of Object. It does depend on the constant pool itable index tag bits, but those have been merged separately in #2651. I'm rebasing to pick up that change, and updating a few copyright header dates, and fixing the outer braces for `jitGetImproperInterfaceMethodFromCP`

runtime/jit_vm/cthelpers.cpp
```diff
@@ -78,6 +78,48 @@ jitGetInterfaceITableIndexFromCP(J9VMThread *currentThread, J9ConstantPool *cons
...
+	J9RAMInterfaceMethodRef *ramMethodRef = (J9RAMInterfaceMethodRef*)constantPool + cpIndex;
+	/* interfaceClass is used to indicate 'resolved'. Make sure to read it first */
+	J9Class* volatile interfaceClass = (J9Class*)ramMethodRef->interfaceClass;
+	VM_AtomicSupport::readBarrier();
+	UDATA volatile methodIndexAndArgCount = ramMethodRef->methodIndexAndArgCount;
+	J9Method *improperMethod = NULL;
+	/* If the ref is resolved, do not call the resolve helper, as it will currently fail if the interface class is not initialized */
+	if (NULL == interfaceClass) {
+		J9RAMInterfaceMethodRef localEntry;
+		if (0 == currentThread->javaVM->internalVMFunctions->resolveInterfaceMethodRefInto(currentThread, constantPool, cpIndex, J9_RESOLVE_FLAG_JIT_COMPILE_TIME, &localEntry)) {
```
DanHeidinga: `resolveInterfaceMethodRefInto()` returns NULL or a `J9Method*`.  Please use NULL rather than 0 in the if condition

gacholio: This is a copy of already bad code, so we should fix both or neither.

DanHeidinga: +1 to both =)

runtime/jit_vm/cthelpers.cpp
jdmpapin: Fixed

runtime/jit_vm/cthelpers.cpp
jdmpapin: Fixed (here and in `jitGetInterfaceITableIndexFromCP`)

jdmpapin:
Removed WIP. Without eclipse/omr#2874, the symbol reference on the `ZEROCHK` for the interface type check will just have more conservative "use-def aliasing" than necessary

andrewcraik:
Jenkins test sanity

gacholio:
This common code change required 20 builds to test it?

andrewcraik:
@gacholio my justification for the builds is that this change a) affects all platforms - the ilgen and optimizer can query codegen specific queries and given the core nature of the change making sure the different combinations of codegen with the common code all work seems prudent b) I felt it prudent to test 8 and 'latest' to ensure the code was compatible in both cases. Perhaps we could run just one of the two JDKs, but I would have lingering concerns. I understand the scarceness of build resources, but given the fundamental changes it seems prudent to ensure good coverage on the testing.

andrewcraik:
@gacholio or @DanHeidinga have the concerns in the VM code been addressed to your satisfaction?

runtime/jit_vm/cthelpers.cpp
```diff
@@ -78,6 +78,48 @@ jitGetInterfaceITableIndexFromCP(J9VMThread *currentThread, J9ConstantPool *cons
...
+		} else if (J9_ARE_ANY_BITS_SET(methodIndexAndArgCount, J9_ITABLE_INDEX_OBJECT)) {
+			/* Object method not in the vTable */
+			improperMethod = jlObject->ramMethods + methodIndex;
+		} else {
+			/* Private interface method */
+			improperMethod = interfaceClass->ramMethods + methodIndex;
+		}
+	}
+done:
+	return improperMethod;
```
DanHeidinga: How does this handle improper but not yet resolved methods?  The current implementation returns null which would imply, given the function's contract, that this is a proper method but it might not be when it's resolved.

gacholio: Unresolved is an entirely different path. This code applies only to the resolved case.

gacholio: This code is actually incorrect - there's no way to determine the method in the virtual case without knowing the receiver class. For reference, the "improper" case from the interpreter is:

```
						/* Object or private interface method invoke */
						if (J9_ARE_ANY_BITS_SET(methodIndexAndArgCount, J9_ITABLE_INDEX_METHOD_INDEX)) {
							if (J9_ARE_ANY_BITS_SET(methodIndexAndArgCount, J9_ITABLE_INDEX_OBJECT)) {
								/* Object method not in the vTable */
								_sendMethod = J9VMJAVALANGOBJECT_OR_NULL(_vm)->ramMethods + methodIndex;
							} else {
								/* Private interface method */
								_sendMethod = interfaceClass->ramMethods + methodIndex;
							}
						} else {
							/* Object method in the vTable */
							_sendMethod = *(J9Method**)((UDATA)receiverClass + methodIndex);
						}```

jdmpapin: In the virtual case, the compiler uses the method to get a VFT offset, and generates a virtual call

gacholio: OK, I understand now.

gacholio: I'll approve - I know Dan would like the code reordered to match the interpreter cases.

jdmpapin: OK, reordering to match the interpreter

fjeremic: @jdmpapin can you add a new commit? We can avoid re-running the testing if we can manually inspect that the change was just simple code movement.

DanHeidinga: Can you also update the comments to include this info in the code?  It'll save someone a lot of digging and frustration in the future if we can record this now while it's fresh.

jdmpapin: Updated

gacholio:
I think this is ready for a squash and merge.

Issue #2571, opened 2018-08-09
SIGINT/SIGTERM doesn't run shutdown hooks in 0.9.0-rc2


thjaeckle:
Hi OpenJ9 team.

Today I tried out the 0.9.0-rc2 on Java 8 (built from AdoptOpenJDK):
https://github.com/AdoptOpenJDK/openjdk8-openj9-nightly/releases/download/jdk8u181-b13-201808010217/OpenJDK8-OPENJ9_x64_Linux_201808010217.tar.gz

My problem now is that for my application (an [Akka](https://akka.io) clustered application) which adds shutdownHooks by itself those are not run and as a consequence the graceful cluster shutdown is not working.

Really puzzling for me is that in simple reproducers (Java with main method registering one or multiple shutdownHooks) the hooks are working as expected and are called both on `SIGINT` and `SIGTERM`.
I also tried a simple Scala program in which a shutdown hook is registered and that also worked.

When starting the same Akka programs with OpenJDK8 Hotspot, the graceful shutdown works as expected, the shutdown hooks are correctly invoked.

I even tried debugging into the `java.lang.Shutdown#runHooks` when sending `SIGTERM`, but this was never called when running with 0.9.0-rc2. It was called when running with Hotspot.

Do you have any ideas or suggestions how I could investigate further?
Would be a shame if our application still can't handle signals like in Hotspot.

Will try to reproduce tomorrow with an Akka application.

DanHeidinga:
@babsingh Can you take a look at this?

thjaeckle:
Still can't reproduce with a simple Akka starter.
Our code in which that doesn't work is here:
https://github.com/eclipse/ditto/

For example, start the following class (e.g. in IntelliJ, running this without any config is possible):
https://github.com/eclipse/ditto/blob/master/services/policies/starter/src/main/java/org/eclipse/ditto/services/policies/starter/PoliciesService.java

Killing the process won't run shutdown hooks and won't terminate the JVM.

thjaeckle:
It appears that four our services, `sun.misc.Signal#dispatch` is not even invoked when the process receives a `SIGTERM`

thjaeckle:
I finally have something.

This is a minimal reproducer with only one dependency:
```xml
<dependency>
  <groupId>com.typesafe.akka</groupId>
  <artifactId>akka-actor_2.12</artifactId>
  <version>2.5.14</version>
</dependency>
```

code:
```java
package test;

[...]

PR #2650, opened 2018-08-17
Fix signal issue when Java main thread terminates

babsingh:
When the Java main thread terminates, DestroyJavaVM gets executed.
DestroyJavaVM sets the J9_RUNTIME_SHUTDOWN_STARTED flag and it invokes
protectedDestroyJavaVM. protectedDestroyJavaVM waits until one
non-daemon thread is alive. So, Java threads can still be alive when
J9_RUNTIME_SHUTDOWN_STARTED is set.

predefinedHandlerWrapper is used to execute Java signal handlers. It
must execute Java signal handlers until Java threads are alive.
Currently, it won't invoke Java signal handlers if
J9_RUNTIME_SHUTDOWN_STARTED is set. Now onwards,
predefinedHandlerWrapper will execute Java signal handlers when
J9_RUNTIME_SHUTDOWN_STARTED is set. It will stop invoking Java signal
handlers only if J9_RUNTIME_EXIT_STARTED is set.

Fixes: #2571

### DIFF
--- a/runtime/vm/jvminit.c
+++ b/runtime/vm/jvminit.c
@@ -6302,9 +6302,9 @@ predefinedHandlerWrapper(struct J9PortLibrary *portLibrary, U_32 gpType, void *g
 		return 1;
 	}
 
-	/* Don't invoke handler if JVM exit or shutdown has started. */
+	/* Don't invoke handler if JVM exit has started. */
 	omrthread_monitor_enter(vm->runtimeFlagsMutex);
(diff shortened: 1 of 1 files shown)


babsingh:
fyi - @DanHeidinga

DanHeidinga:

DanHeidinga:
Reverted in https://github.com/eclipse/openj9/pull/2658 due to new test failure.  `threadMXBeanTestSuite2_21 is failing due to timeouts in a build that only has this change in it.

babsingh:
This change is being repushed through https://github.com/eclipse/openj9/pull/2664.

PR #2651, opened 2018-08-17
Add iTable tag bits

gacholio:
Add bits to enable the JIT to push nestmates code.

### DIFF
--- a/runtime/oti/j9consts.h
+++ b/runtime/oti/j9consts.h
@@ -821,6 +821,24 @@ extern "C" {
 /* Flag for Nestmates invokeVirtual on private methods */
 #define J9_VTABLE_INDEX_DIRECT_METHOD_FLAG 0x1
 
+/* Tag bits for J9RAMInterfaceMethodRef->methodIndexAndArgCount:
+*
+*	J9_ITABLE_INDEX_METHOD_INDEX - index represents a direct method index
+*	J9_ITABLE_INDEX_OBJECT		 - use Object rather than the targetted interfaceClass
+*
+* Private methods in interface will have only J9_ITABLE_INDEX_METHOD_INDEX set, indicating that
+* the underlying index is a method index into the interfaceClass of the ref.
+*
+* vTable methods in Object will have only J9_ITABLE_INDEX_OBJECT set, indicating that
+* the underlying index is a vTable index in Object.
+*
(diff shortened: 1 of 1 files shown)


gacholio:
jenkins compile xlinux jdk8

gacholio:
Whichever compile completes first is sufficient testing.

gacholio:
Yes, but that's in the other branch. Just want to get the bits in so the JIT can reference them.

runtime/oti/j9consts.h
```diff
@@ -821,6 +821,24 @@ extern "C" {
...
+* the underlying index is a method index into the interfaceClass of the ref.
+*
+* vTable methods in Object will have only J9_ITABLE_INDEX_OBJECT set, indicating that
+* the underlying index is a vTable index in Object.
+*
+* Non-vTable methods in Object will have J9_ITABLE_INDEX_OBJECT and J9_ITABLE_INDEX_METHOD_INDEX set,
+* indicating that the underlying index is a method index in Object.
+*/
+#define J9_ITABLE_INDEX_METHOD_INDEX	((UDATA)1 << ((8 * sizeof(UDATA)) - 1))
+#define J9_ITABLE_INDEX_OBJECT			((UDATA)1 << ((8 * sizeof(UDATA)) - 2))
```
fjeremic: Extra tab here.

gacholio: Actually, it's fine in Eclipse, which is where I write my code.

gacholio: Good argument against trying to line things up, I suppose.

gacholio: @DanHeidinga agrees with not lining up things like this. What's your take on single space vs single tab for these cases, @fjeremic ?

fjeremic: I'm indifferent either way, just something that stands out in reviews. It looks like alignment was attempted in this case but was not achieved outside of Eclipse IDE. These sorts of stylistic changes shouldn't block any merge though IMO. But yes you're right, perhaps easiest not to line things up. Can't argue with that!

gacholio: Personally, I would choose space, since the tab alignment is somewhat arbitrary. I've been doing this a lot lately when reordering constants that were previously generated, so they were in no particular order - lining them up in the IDE made things a lot easier to sort.

Issue #2602, opened 2018-08-14
invokespecial updates


gacholio:
See #1730

invokespecial checks for NULL receiver before resolution. It needs a resolution check in the NPE case similar the invokevirtual/invokeinterface.

invokespecialsplit needs a similar change. It's also missing the profiling call in the case where the method is already resolved. Ideally, the code should be almost identical to invokespecial, which is made difficult by the fact that the unresolved method is NULL rather than initialSpecialMethod.

Look into filling the split array with a fake method. If that proves possible, consider doing the same for the static split array (saves a runtime check on every split send).

PR #2652, opened 2018-08-18
Fix and optimize split bytecodes

gacholio:
- commonize logic for invokespecial and invokespecialsplit
- use initial methods instead of NULL for unresolved split bytecodes
- invokespecial/split throw resolve errors before NPE
- add missing profile call to invokespecialsplit

Fixes: #2602

### DIFF
--- a/runtime/vm/BytecodeInterpreter.hpp
+++ b/runtime/vm/BytecodeInterpreter.hpp
@@ -1874,13 +1874,18 @@ done:
 	initialStaticMethod(REGISTER_ARGS_LIST)
 	{
 		VM_BytecodeAction rc = GOTO_RUN_METHOD;
+		bool split = (JBinvokestaticsplit == _pc[0]);
 		U_16 index = *(U_16*)(_pc + 1);
 		J9ConstantPool *ramConstantPool = J9_CP_FROM_METHOD(_literals);
 		buildGenericSpecialStackFrame(REGISTER_ARGS, 0);
 		updateVMStruct(REGISTER_ARGS);
-		J9Method *method = resolveStaticMethodRef(_currentThread, ramConstantPool, index, J9_RESOLVE_FLAG_RUNTIME_RESOLVE | J9_RESOLVE_FLAG_CHECK_CLINIT);
-		if ((J9Method*)-1 == method) {
-			method = ((J9RAMStaticMethodRef*)&_currentThread->floatTemp1)->method;
+		if (split) {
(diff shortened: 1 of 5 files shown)


gacholio:

runtime/vm/BytecodeInterpreter.hpp
```diff
@@ -6607,32 +6560,12 @@ done:;
 	VMINLINE VM_BytecodeAction
 	invokestaticsplit(REGISTER_ARGS_LIST)
 	{
-		VM_BytecodeAction rc = GOTO_RUN_METHOD;
 		U_16 splitTableIndex = *(U_16*)(_pc + 1);
+		profileCallingMethod(REGISTER_ARGS);
```
DanHeidinga: Can this result in profiling the initialStaticMethod?  Is the iprofiler ready to handle that?

runtime/vm/BytecodeInterpreter.hpp
```diff
@@ -1898,19 +1901,24 @@ obj:;
...
+		}
 		VMStructHasBeenUpdated(REGISTER_ARGS);
 		restoreGenericSpecialStackFrame(REGISTER_ARGS);
 		if (immediateAsyncPending()) {
 			rc = GOTO_ASYNC_CHECK;
 		} else if (VM_VMHelpers::exceptionPending(_currentThread)) {
 			rc = GOTO_THROW_CURRENT_EXCEPTION;
 		} else {
-			_sendMethod = method;
+			rc = invokespecialLogic(REGISTER_ARGS, false, split);
```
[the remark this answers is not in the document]
gacholio: The order that must be enforced is:

- resolution errors
- NPE
- interface not implemented errors

The initial method is invoked before the NPE and interface checks (logically), so the logic must run again. See invokevirtualLogic.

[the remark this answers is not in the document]
gacholio: The heart of the "optimization" portion of the change is to run the intial method appropriately instead of doing the NULL method check in invokespecialsplit.

runtime/vm/BytecodeInterpreter.hpp
gacholio: The profile here is for the calling method, not the target method. invokespecial already did this.

```	VMINLINE void
	profileCallingMethod(REGISTER_ARGS_LIST)
	{
		U_8 *profilingCursor = startProfilingRecord(REGISTER_ARGS, sizeof(J9Method*));
		if (NULL != profilingCursor) {
			*(J9Method**)profilingCursor = _literals;
		}
	}
```

gacholio: invokestatic does exactly this.

gacholio:
FYI, I manually tested this with the always split FV option on and off, so I'm confident that the split behaviour matches the normal one.

PR #2656, opened 2018-08-19
Store iTable offset intead of index in interface snippet data

gacholio:
In preparation for invokeinterface of private methods, store the iTable
offset instead of the index in the resolved interface snippet data. This
frees up low bits for use to indicate direct rather than virtual methods
in invokeinterface.

### DIFF
--- a/runtime/codert_vm/cnathelp.cpp
+++ b/runtime/codert_vm/cnathelp.cpp
@@ -999,7 +999,8 @@ old_slow_jitLookupInterfaceMethod(J9VMThread *currentThread)
 	J9ConstantPool *ramConstantPool = ((J9ConstantPool**)indexAndLiteralsEA)[-2];
 	UDATA cpIndex = indexAndLiteralsEA[-1];
 	J9Class *interfaceClass = ((J9Class**)indexAndLiteralsEA)[0];
-	UDATA methodIndex = indexAndLiteralsEA[1];
+	// convert offset back to index for now
+	UDATA methodIndex = (indexAndLiteralsEA[1] - sizeof(J9ITable)) / sizeof(UDATA);
 	UDATA vTableOffset = VM_VMHelpers::convertITableIndexToVTableOffset(currentThread, receiverClass, interfaceClass, methodIndex, ramConstantPool, cpIndex);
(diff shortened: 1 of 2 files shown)


gacholio:

gacholio:

gacholio:

gacholio:
For the record, the only consumers of the iTable index/offset slot in the resolved data are the C implementatin of jitLookupInterfaceMethod (all platforms) and the lastITable check in the PPC PicBuilder.

DanHeidinga:
>  lastITable check in the PPC PicBuilder.

Do the other platforms not do the lastITable check?  Or it doesn't require PicBuilder change on those platforms?  (I thought the lastITable change was first developed on x)

gacholio:
I think various (or all) platforms have the check in the code, which does not use the slot in the interface snippet data, so they are unaffected by this change.

runtime/codert_vm/cnathelp.cpp
```diff
@@ -1516,7 +1518,8 @@ old_slow_jitResolveInterfaceMethod(J9VMThread *currentThread)
 		goto retry;
 	}
 	indexAndLiteralsEA[2] = (UDATA)interfaceClass;
-	indexAndLiteralsEA[3] = methodIndexAndArgCount >> 8; /* remove argCount */
+	// store iTable offset instead of index
+	indexAndLiteralsEA[3] = ((methodIndexAndArgCount >> 8) * sizeof(UDATA)) + sizeof(J9ITable);
```
DanHeidinga: Does this need to clear the tag bits: `J9_ITABLE_INDEX_METHOD_INDEX | J9_ITABLE_INDEX_OBJECT` ?

DanHeidinga: This code writes the offset into `indexAndLiteralsEA[3]` but the other code deals with `[1]`.  Am I missing something?

gacholio: One of the helpers passes the head of the data, the other passes the middle of the data (for historical reasons).

The bits will be dealt with in the follow-up PR - they are currently never set.

PR #2665, opened 2018-08-21
JDK11 nestmate virtual private function call

NigelYiboYu:
- This change enables JIT to handle unresolved invokeVirtual and invokeInterface calls whose
call targets turn out to be private nestmate methods.
The S390VirtualUnresolvedSnippet gets three more data fields to hold
runtime data for private nestmate methods: a J9Method field, a J2I entry
point field, and a private return address field.
The PicBuilder routines are modified to check for private call tags in the
return value of the VM resolve helper. If the resolve helper returns a
low-tagged value, the call is identified as a private nestmate method,
which gets a direct dispatch. Normal virtual/interface call handling does not
change.

- Add tests to exercise this new nestmate private invocation path.

- Create J9S390InterfaceCallDataSnippet to mirror OMR class S390InterfaceCallDataSnippet

### DIFF
--- a/runtime/compiler/z/codegen/J9S390PrivateLinkage.cpp
+++ b/runtime/compiler/z/codegen/J9S390PrivateLinkage.cpp
@@ -1864,7 +1864,7 @@ TR::S390PrivateLinkage::buildVirtualDispatch(TR::Node * callNode, TR::RegisterDe
    TR_J9VMBase *fej9 = (TR_J9VMBase *)(comp()->fe());
 
    // Generate and register a thunk for a resolved virtual function
-   uint8_t *virtualThunk;
+   void *virtualThunk;
    if (methodSymbol && methodSymbol->isComputed())
       {
       switch (methodSymbol->getMandatoryRecognizedMethod())
@@ -1873,10 +1873,10 @@ TR::S390PrivateLinkage::buildVirtualDispatch(TR::Node * callNode, TR::RegisterDe
             {
             char *j2iSignature = fej9->getJ2IThunkSignatureForDispatchVirtual(methodSymbol->getMethod()->signatureChars(), methodSymbol->getMethod()->signatureLength(), comp());
             int32_t signatureLen = strlen(j2iSignature);
-            virtualThunk = (uint8_t*)fej9->getJ2IThunk(j2iSignature, signatureLen, comp());
+            virtualThunk = fej9->getJ2IThunk(j2iSignature, signatureLen, comp());
             if (!virtualThunk)
                {
-               virtualThunk = (uint8_t*)fej9->setJ2IThunk(j2iSignature, signatureLen,
+               virtualThunk = fej9->setJ2IThunk(j2iSignature, signatureLen,
                   TR::S390J9CallSnippet::generateVIThunk(
                      fej9->getEquivalentVirtualCallNodeForDispatchVirtual(callNode, comp()), sizeOfArguments, cg()), comp()); // TODO:JSR292: Is this the right sizeOfArguments?
                }
@@ -1893,9 +1893,9 @@ TR::S390PrivateLinkage::buildVirtualDispatch(TR::Node * callNode, TR::RegisterDe
       }
    else
       {
-      virtualThunk = (uint8_t*)(fej9->getJ2IThunk(methodSymbol->getMethod(), comp()));
+      virtualThunk = fej9->getJ2IThunk(methodSymbol->getMethod(), comp());
       if (!virtualThunk)
-         virtualThunk = (uint8_t*)(fej9->setJ2IThunk(methodSymbol->getMethod(), TR::S390J9CallSnippet::generateVIThunk(callNode, sizeOfArguments, cg()), comp()));
+         virtualThunk = fej9->setJ2IThunk(methodSymbol->getMethod(), TR::S390J9CallSnippet::generateVIThunk(callNode, sizeOfArguments, cg()), comp());
       }
 
    if (methodSymbol->isVirtual() && (!methodSymRef->isUnresolved() && !comp()->compileRelocatableCode()))
@@ -2251,6 +2251,9 @@ TR::S390PrivateLinkage::buildVirtualDispatch(TR::Node * callNode, TR::RegisterDe
       gcPoint = new (trHeapMemory()) TR::S390RRInstruction(TR::InstOpCode::BASR, callNode, RegRA, RegRA, cg());
       gcPoint->setDependencyConditions(preDeps);
 
+      if (unresolvedSnippet != NULL)
+         (static_cast<TR::S390VirtualUnresolvedSnippet *>(unresolvedSnippet))->setIndirectCallInstruction(gcPoint);
+
       if (outlinedSlowPath)
          {
          TR::Instruction * temp = generateS390BranchInstruction(cg(),TR::InstOpCode::BRC,TR::InstOpCode::COND_BRC,callNode,doneVirtualLabel);
@@ -2347,7 +2350,7 @@ TR::S390PrivateLinkage::buildVirtualDispatch(TR::Node * callNode, TR::RegisterDe
 
       TR::LabelSymbol * snippetLabel = generateLabelSymbol(cg());
       TR::S390InterfaceCallSnippet * ifcSnippet = new (trHeapMemory()) TR::S390InterfaceCallSnippet(cg(), callNode,
-           snippetLabel, sizeOfArguments, numInterfaceCallCacheSlots, virtualThunk);
+           snippetLabel, sizeOfArguments, numInterfaceCallCacheSlots, virtualThunk, false);
       cg()->addSnippet(ifcSnippet);
 
       if (numStaticPICs != 0)
--- a/runtime/compiler/z/codegen/S390J9CallSnippet.cpp
+++ b/runtime/compiler/z/codegen/S390J9CallSnippet.cpp
@@ -527,10 +527,39 @@ TR::S390VirtualUnresolvedSnippet::emitSnippetBody()
    *(uintptrj_t *) cursor = (uintptrj_t) callNode->getSymbolReference()->getCPIndexForVM();
    cursor += TR::Compiler->om.sizeofReferenceAddress();
 
+   // instruction to be patched
    *(uintptrj_t *) cursor = (uintptrj_t) (getPatchVftInstruction()->getBinaryEncoding());
    AOTcgDiag1(comp, "add TR_AbsoluteMethodAddress cursor=%x\n", cursor);
    cg()->addExternalRelocation(new (cg()->trHeapMemory()) TR::ExternalRelocation(cursor, NULL, TR_AbsoluteMethodAddress, cg()),
                           __FILE__, __LINE__, callNode);
+   cursor += sizeof(uintptrj_t);
+
+   // Field used by nestmate private calls
+   // J9Method pointer of the callee. Initialized to 0.
+   *(uintptrj_t *) cursor = 0;
+   cursor += sizeof(uintptrj_t);
+
+   // Field used by nestmate private calls
+   // J2I thunk address
+   // TODO: leave out the AOT until AOT supports this.
+   *(uintptrj_t *) cursor = (uintptrj_t) thunkAddress;
+   // AOTcgDiag1(comp, "add ThunkAddress cursor=%x\n", cursor);
+   // cg()->addExternalRelocation(new (cg()->trHeapMemory()) TR::ExternalRelocation(cursor, NULL, TR_AbsoluteMethodAddress, cg()),
+   //                        __FILE__, __LINE__, callNode);
+   cursor += sizeof(uintptrj_t);
+
+   // Field used by nestmate private calls
+   // For private functions, this is the return address.
+   // Add a 2 bytes off set because it's the instruction after the BASR, which is 2-byte long
+   // The assumption here is that S390 J9 private linkage indirect dispatch is emitting BASR.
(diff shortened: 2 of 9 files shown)


NigelYiboYu:
Marking as WIP until fully tested

runtime/compiler/z/codegen/S390J9CallSnippet.hpp
```diff
@@ -117,11 +117,16 @@ class S390VirtualUnresolvedSnippet : public TR::S390VirtualSnippet
    virtual uint8_t *emitSnippetBody();
 
    virtual uint32_t getLength(int32_t estimatedSnippetStart);
+   uint8_t* getJ2IThunk() { return thunkAddress; }
```
fjeremic: Can we rename this to `getJ2IThunkAddress`? It really is extracting the address and perhaps even make it a `void*`?

runtime/compiler/z/codegen/S390J9CallSnippet.hpp
```diff
@@ -133,6 +138,7 @@ class S390InterfaceCallSnippet : public TR::S390VirtualSnippet
 
    public:
 
+   // TODO: remove this ctor. thunk pointer is now mandatory
```
fjeremic: Can we address this TODO in a new commit? We should do it now that we are in the area of this code.

runtime/compiler/z/codegen/S390J9CallSnippet.cpp
```diff
@@ -527,20 +527,45 @@ TR::S390VirtualUnresolvedSnippet::emitSnippetBody()
...
    cursor += sizeof(uintptrj_t);
 
+   // Field used by nestmate private calls
+   // J9Method pointer of the callee. Initialized to 0.
+   *(uintptrj_t *) cursor = 0;
+   cursor += sizeof(uintptrj_t);
+
+   // Field used by nestmate private calls
+   // J2I thunk address
+   // TODO: leave out the AOT until AOT supports this.
```
fjeremic: Similarly to the comment made in the OMR PR do we have an issue opened for this that we can reference?

runtime/compiler/z/codegen/S390J9CallSnippet.cpp
```diff
@@ -527,20 +527,45 @@ TR::S390VirtualUnresolvedSnippet::emitSnippetBody()
...
+   *(uintptrj_t *) cursor = (uintptrj_t) thunkAddress;
+   // AOTcgDiag1(comp, "add ThunkAddress cursor=%x\n", cursor);
+   // cg()->addExternalRelocation(new (cg()->trHeapMemory()) TR::ExternalRelocation(cursor, NULL, TR_AbsoluteMethodAddress, cg()),
+   //                        __FILE__, __LINE__, callNode);
+   cursor += sizeof(uintptrj_t);
+
+   // Field used by nestmate private calls
+   // For private functions, this is the return address.
+   // Add a 2 bytes off set because it's the instruction after the BASR, which is 2-byte long
```
fjeremic: Could you please add a fatal assert that `getIndirectCallInstruction() == TR::InstOpCode::BASR` since we are making the assumption here? Alternatively can't we use `getIndirectCallInstruction()->getBinaryLength()` instead of hardcoding 2 here and avoid adding the assert?

runtime/compiler/z/codegen/S390J9CallSnippet.cpp
```diff
@@ -752,6 +777,18 @@ TR_Debug::print(TR::FILE *pOutFile, TR::S390VirtualUnresolvedSnippet * snippet)
...
+   printPrefix(pOutFile, NULL, bufferPos, sizeof(intptrj_t));
+   trfprintf(pOutFile, "DC   \t%p \t\t# Private J9Method pointer", 0);
+   bufferPos += sizeof(intptrj_t);
+
+   printPrefix(pOutFile, NULL, bufferPos, sizeof(intptrj_t));
+   trfprintf(pOutFile, "DC   \t%p \t\t# J2I thunk address for private", snippet->getJ2IThunk());
+   bufferPos += sizeof(intptrj_t);
+
+   printPrefix(pOutFile, NULL, bufferPos, sizeof(intptrj_t));
+   trfprintf(pOutFile, "DC   \t%p \t\t# RA for private", snippet->getBranchInstruction()->getBinaryEncoding() + 2);
```
fjeremic: Similar comment to above for constant 2.

runtime/compiler/z/codegen/J9S390PrivateLinkage.cpp
```diff
@@ -2251,6 +2251,9 @@ TR::S390PrivateLinkage::buildVirtualDispatch(TR::Node * callNode, TR::RegisterDe
       gcPoint = new (trHeapMemory()) TR::S390RRInstruction(TR::InstOpCode::BASR, callNode, RegRA, RegRA, cg());
       gcPoint->setDependencyConditions(preDeps);
 
+      if (unresolvedSnippet)
+          ((TR::S390VirtualUnresolvedSnippet *)unresolvedSnippet)->setIndirectCallInstruction(gcPoint);
+
```
fjeremic: Can we add some braces here and use `static_cast`? Also `unresolvedSnippet` is of pointer type so we should be comparing against `!= NULL`.

runtime/compiler/z/runtime/s390_asdef.inc
```diff
@@ -61,6 +61,7 @@ define([LNR_GPR],lngr)
 define([LTR_GPR],ltgr)
 define([MSGR_GPR],msgr)
 define([N_GPR],ng)
+define([NR_GPR],ngr)
 define([NIHH_GPR],nihh)
```
fjeremic: Can we alphabetize the newly added line? Similarly below.

runtime/compiler/z/runtime/s390_macros.inc
```diff
@@ -97,21 +97,22 @@ SETVAL(eq_cp_inInterfaceSnippet,8)
...
 
 ZZ For Multi slot case
-SETVAL(eq_lastCachedSlotField_inInterfaceSnippet,48)
-SETVAL(eq_firstSlotField_inInterfaceSnippet,56)
-SETVAL(eq_lastSlotField_inInterfaceSnippet,64)
-SETVAL(eq_firstSlot_inInterfaceSnippet,72)
+SETVAL(eq_lastCachedSlotField_inInterfaceSnippet,56)
+SETVAL(eq_firstSlotField_inInterfaceSnippet,64)
+SETVAL(eq_lastSlotField_inInterfaceSnippet,72)
+SETVAL(eq_firstSlot_inInterfaceSnippet,80)
```
fjeremic: Why are these defined in s390_macros.inc while all the other ones are defined in PicBuilder.m4? Can we consolidate them all into one place, i.e. PicBuilder where they are actually used?

runtime/compiler/z/codegen/S390J9CallSnippet.hpp
NigelYiboYu: OK. I'll change this variable and getter to `void*`, together with some relevant `J9S390PrivateLinkage`  code that has to do with thunk address.

runtime/compiler/z/runtime/s390_asdef.inc
NigelYiboYu: Fixed.

runtime/compiler/z/codegen/S390J9CallSnippet.cpp
NigelYiboYu: Updated this pull request with a new commit that removes these constructors because they are unused and takes no thunkAddress parameter.

NigelYiboYu:
I left a note on a separate issue about the AOT. 
Other concerns should've been addressed in the latest commits.
I'm re-testing the new changes.

runtime/compiler/z/codegen/S390J9CallSnippet.cpp
```diff
@@ -527,20 +527,46 @@ TR::S390VirtualUnresolvedSnippet::emitSnippetBody()
...
+   *(uintptrj_t *) cursor = (uintptrj_t) thunkAddress;
+   // AOTcgDiag1(comp, "add ThunkAddress cursor=%x\n", cursor);
+   // cg()->addExternalRelocation(new (cg()->trHeapMemory()) TR::ExternalRelocation(cursor, NULL, TR_AbsoluteMethodAddress, cg()),
+   //                        __FILE__, __LINE__, callNode);
+   cursor += sizeof(uintptrj_t);
+
+   // Field used by nestmate private calls
+   // For private functions, this is the return address.
+   // Add a 2 bytes off set because it's the instruction after the BASR, which is 2-byte long
+   TR_ASSERT_FATAL(getIndirectCallInstruction() == TR::InstOpCode::BASR, "Unexpected branch instruction in VirtualUnresolvedSnippet.\n");
```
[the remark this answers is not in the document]
NigelYiboYu: Fixed. I caught this in my JIT build.

runtime/compiler/z/codegen/J9S390PrivateLinkage.cpp
```diff
@@ -2251,6 +2251,9 @@ TR::S390PrivateLinkage::buildVirtualDispatch(TR::Node * callNode, TR::RegisterDe
       gcPoint = new (trHeapMemory()) TR::S390RRInstruction(TR::InstOpCode::BASR, callNode, RegRA, RegRA, cg());
       gcPoint->setDependencyConditions(preDeps);
 
+      if (unresolvedSnippet != NULL)
+          (static_cast<TR::S390VirtualUnresolvedSnippet *>(unresolvedSnippet))->setIndirectCallInstruction(gcPoint);
```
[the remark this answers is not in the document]
NigelYiboYu: Fixed.

runtime/compiler/z/runtime/PicBuilder.m4
```diff
@@ -1689,6 +1765,67 @@ LABEL(L_patchBRASL)
...
+LABEL(L_privateMethod)
+    ST_GPR  r2,eq_privMethod_inVUCallSnippet(r14)
+LABEL(L_privateMethodRemoveTag)
+
+ZZ  Load bitwise NOT of the direct method flag
+ifdef([J9ZOS390],[dnl
+ZZ zOS
+    LHI_GPR r0,J9TR_J9_VTABLE_INDEX_DIRECT_METHOD_FLAG
+    LCR_GPR r0,r0
+    AHI_GPR r0,-1
```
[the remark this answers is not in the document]
NigelYiboYu: This `NOT` built-in function is something I couldn't find.

runtime/compiler/z/runtime/PicBuilder.m4
```diff
@@ -2286,6 +2471,100 @@ ZZ # Load address of the lookup class
...
+    L_GPR   r0,eq_intfMethodIndex_inInterfaceSnippet(r14)
+    LR_GPR  r3,r14         # free r14 for RA
+    LR_GPR  r1,r0          # keep low-tagged in r0
+
+ZZ  bitwise NOT the flag
+ifdef([J9ZOS390],[dnl
+ZZ zOS
+    LHI_GPR r2,J9TR_J9_ITABLE_OFFSET_DIRECT
+    LCR_GPR r2,r2
+    AHI_GPR r2,-1
```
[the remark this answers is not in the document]
NigelYiboYu: Done and re-testing

runtime/compiler/z/runtime/PicBuilder.m4
```diff
@@ -1689,6 +1765,67 @@ LABEL(L_patchBRASL)
...
+ifdef([TR_HOST_64BIT],[dnl
+    LGF     r2,-4(r1)     # Load magic word
+],[dnl
+    LR_GPR  r2,r1         # Copy PCStart
+    AHI_GPR r2,-4         # Move up to the magic word
+    L       r2,0(r2)      # Load magic word
+])dnl
+
+ZZ The first half of magic word is jit-to-jit offset, so we need to
+ZZ shift it to lower half
```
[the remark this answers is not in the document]
NigelYiboYu: Comment updated.

runtime/compiler/z/runtime/s390_macros.inc
NigelYiboYu: I moved these back to `s390_macros.inc` because a few of these are used by both the picBulder and valueProf.m4.

runtime/compiler/z/runtime/PicBuilder.m4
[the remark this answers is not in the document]
NigelYiboYu: The `NOT` operator is not supported by z/OS assembler. Reverting back to ~x=-x-1

fjeremic:
Jenkins test sanity zlinux JDK8,JDK11 depends eclipse/omr#master

fjeremic:
JDK11 build is "unstable". Seems there is a crash in some VM port library test. Doesn't seem to affect the code from this PR. I think we're good to go here.

fjeremic:
Actually it seems OMR Acceptance is blocked by this change since we moved the print functions up to OpenJ9. We now have mutual dependency between this PR and eclipse/omr#master so OMR Acceptance cannot promote without merging this. See failure here:

```
JVMJ9VM011W Unable to load j9jit29: /home/jenkins/jenkins-agent/workspace/Test-sanity.functional-JDK8-linux_390-64_cmprssptrs/openjdkbinary/j2sdk-image/jre/lib/s390x/compressedrefs/libj9jit29.so: undefined symbol: _ZN8TR_Debug5printEPN2TR11FilePointerEPNS0_30J9S390InterfaceCallDataSnippetE
```

https://ci.eclipse.org/openj9/blue/rest/organizations/jenkins/pipelines/Test-sanity.functional-JDK8-linux_390-64_cmprssptrs/runs/157/nodes/74/steps/78/log/?start=0

PR #2671, opened 2018-08-21
Refactor nestmates jcl natives

tajila:
Refactor nestmates jcl natives

### DIFF
--- a/runtime/vm/visible.c
+++ b/runtime/vm/visible.c
@@ -29,6 +29,7 @@
 #include "util_api.h"
 #include "vm_internal.h"
 #include "j9protos.h"
+#include "j9vmnls.h"
 
 IDATA
 checkModuleAccess(J9VMThread *currentThread, J9JavaVM* vm, J9ROMClass* srcRomClass, J9Module* srcModule, J9ROMClass* destRomClass, J9Module* destModule, UDATA destPackageID, UDATA lookupOptions)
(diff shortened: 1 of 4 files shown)


tajila:

tajila:
I couldn't move the definition of `romHostClass` because of the `goto done` just below

tajila:
Also, if this is merged before the next release there should be no problem with removing the j9jcl.nls messages

runtime/vm/visible.c
```diff
@@ -185,6 +186,92 @@ checkVisibility(J9VMThread *currentThread, J9Class* sourceClass, J9Class* destCl
...
+		nlsTemplate = j9nls_lookup_message(
+				J9NLS_DO_NOT_PRINT_MESSAGE_TAG | J9NLS_DO_NOT_APPEND_NEWLINE,
+				J9NLS_VM_NEST_MEMBER_NOT_CLAIMED_BY_NEST_HOST,
+				NULL);
+		break;
+	default:
+		Assert_VM_unreachable();
+	}
+
+	char *msg = NULL;
```
DanHeidinga: This can't be declared here due to c89.  It needs to be declared at the top of the smallest enclosing scope - likely the function for this var

runtime/vm/visible.c
```diff
@@ -231,6 +318,8 @@ loadAndVerifyNestHost(J9VMThread *vmThread, J9Class *clazz, UDATA options)
 		/* If a problem occurred in nest host verification then the nest host value is invalid */
 		if  ((J9_VISIBILITY_ALLOWED == result) && (NULL != nestHost)) {
 			clazz->nestHost = nestHost;
+		} else if (throwException) {
```
DanHeidinga: if `result == J9_VISIBILITY_ALLOWED` but `nestHost == NULL`, won't the `setNestmatesIncompatibleClassChangeError()` assert?

In what case would visibility be allowed by the nestHost end up null?

DanHeidinga: I'm not sure I see a path where this can happen which means the `if  ((J9_VISIBILITY_ALLOWED == result) && (NULL != nestHost)) ` is incorrect

tajila:
@DanHeidinga changes are ready for another look

DanHeidinga:
@tajila Can you check the extended test failures?

tajila:
@DanHeidinga There is an existing problem with the jdk11 PR builds. The test material isn't able to compile compile properly because some of the test material is not in the correct place

DanHeidinga:
> The test material isn't able to compile compile properly because some of the test material is not in the correct place

Is there an open issue for this so we can track when it will be fixed?

PR #2675, opened 2018-08-22
Throw IllegalAccessError on resolution with inaccessible types

pdbain-ibm:
Make OpenJ9 behaviour compliant to the reference implementation.

### DIFF
--- a/jcl/src/java.base/share/classes/java/lang/invoke/MethodHandle.java
+++ b/jcl/src/java.base/share/classes/java/lang/invoke/MethodHandle.java
@@ -846,6 +846,14 @@ public abstract class MethodHandle {
 			Objects.requireNonNull(classObject);
 			
 			type = MethodType.vmResolveFromMethodDescriptorString(methodDescriptor, access.getClassloader(classObject), null);
+			final MethodHandles.Lookup lookup = new MethodHandles.Lookup(classObject, false);
+			try {
+				lookup.accessCheckArgRetTypes(type);
+			} catch (IllegalAccessException e) {
+				IllegalAccessError err = new IllegalAccessError();
+				err.initCause(e);
+				throw err;
+			}
 			int bsmIndex = UNSAFE.getShort(bsmData);
 			int bsmArgCount = UNSAFE.getShort(bsmData + BSM_ARGUMENT_COUNT_OFFSET);
 			long bsmArgs = bsmData + BSM_ARGUMENTS_OFFSET;
@@ -856,7 +864,7 @@ public abstract class MethodHandle {
 			}
 			Object[] staticArgs = new Object[BSM_OPTIONAL_ARGUMENTS_START_INDEX + bsmArgCount];
 			/* Mandatory arguments */
-			staticArgs[BSM_LOOKUP_ARGUMENT_INDEX] = new MethodHandles.Lookup(classObject, false);
+			staticArgs[BSM_LOOKUP_ARGUMENT_INDEX] = lookup;
 			staticArgs[BSM_NAME_ARGUMENT_INDEX] = name;
 			staticArgs[BSM_TYPE_ARGUMENT_INDEX] = type;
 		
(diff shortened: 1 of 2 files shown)


jcl/src/java.base/share/classes/java/lang/invoke/MethodHandles.java
```diff
@@ -2503,6 +2510,8 @@ public static MethodHandle varHandleInvoker(VarHandle.AccessMode accessMode, Met
 	public static MethodHandle throwException(Class<?> returnType, Class<? extends Throwable> exception) {
 		MethodType realType = MethodType.methodType(returnType, exception);	
 		MethodHandle handle;
+		System.out.println("PDB_DEBUG  throwException returnType="+returnType
```
JasonFengJ9: This appears a WIP.

jcl/src/java.base/share/classes/java/lang/invoke/MethodHandles.java
```diff
@@ -635,14 +635,21 @@ private MethodHandle findSpecialImpl(Class<?> clazz, String methodName, MethodTy
...
+			}
 			Map<CacheKey, WeakReference<MethodHandle>> cache = HandleCache.getStaticCache(clazz);
 			MethodHandle handle = HandleCache.getMethodFromPerClassCache(cache, methodName, type);
 			if (handle == null) {
 				initCheck(methodName);
 				handle = new DirectHandle(clazz, methodName, type, MethodHandle.KIND_STATIC, null);
 				handle = HandleCache.putMethodInPerClassCache(cache, methodName, type, convertToVarargsIfRequired(handle));
 			}
-			checkAccess(handle, false);
+			checkAccess(handle, true);
```
JasonFengJ9: Pls explain the reason that `skipAccessCheckPara` need to be changed to true.

pdbain-ibm:
@DanHeidinga I reworked per 
https://github.com/eclipse/openj9/pull/2675#pullrequestreview-148735247 to do the checking in `MethodHandle.sendResolveMethodHandle()` and `MethodHandle.resolveInvokeDynamic()` as a separate commit and reran the failing test suite.  If that is acceptable I will squash it.

jcl/src/java.base/share/classes/java/lang/invoke/MethodHandle.java
```diff
@@ -855,8 +855,16 @@ private static final MethodHandle resolveInvokeDynamic(long j9class, String name
 				throw new NullPointerException(Msg.getString("K05cd", classObject.toString(), bsmIndex)); //$NON-NLS-1$
 			}
 			Object[] staticArgs = new Object[BSM_OPTIONAL_ARGUMENTS_START_INDEX + bsmArgCount];
+			final MethodHandles.Lookup lookup = new MethodHandles.Lookup(classObject, false);
```
DanHeidinga: The access check should be moved to immediately after the vmResolveFromMethodDescriptorString call so that access exceptions occur before any exceptions getting the BSM

jcl/src/java.base/share/classes/java/lang/invoke/MethodHandle.java
```diff
@@ -1088,43 +1096,58 @@ private static final MethodHandle sendResolveMethodHandle(
...
 			case 9: /* invokeInterface */
 				type = MethodType.vmResolveFromMethodDescriptorString(typeDescriptor, loader, null);
-				return lookup.findVirtual(referenceClazz, name, type);
+				result = lookup.findVirtual(referenceClazz, name, type);
+				break;
+			default:
+				/* Can never happen */
+				throw new UnsupportedOperationException();
+			}
+			if (!Objects.nonNull(type)) {
```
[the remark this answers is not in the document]
DanHeidinga: Similar concern here - the access check should precede the MH lookup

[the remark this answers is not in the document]
pdbain-ibm: `the access check should precede the MH lookup`: moving the checks into the various `case`s.

jcl/src/java.base/share/classes/java/lang/invoke/MethodHandle.java
pdbain-ibm: Will do.

pdbain-ibm:
Updated and commits squashed.

jcl/src/java.base/share/classes/java/lang/invoke/MethodHandle.java
```diff
@@ -1088,43 +1096,60 @@ private static final MethodHandle sendResolveMethodHandle(
 		try {
 			MethodHandles.Lookup lookup = new MethodHandles.Lookup(currentClass, false);
 			MethodType type = null;
-			
+			MethodHandle result = null;
+
 			switch(cpRefKind){
 			case 1: /* getField */
-				return lookup.findGetter(referenceClazz, name, resolveFieldHandleHelper(typeDescriptor, loader));
+				result = lookup.findGetter(referenceClazz, name, resolveFieldHandleHelper(typeDescriptor, loader));
```
DanHeidinga: `resolveFieldHandleHelper(typeDescriptor, loader)` will have the same issue as `vmResolveFromMethodDescriptorString` and will also need the accessCheck.

pdbain-ibm: Done.  `resolveFieldHandleHelper()` changed to take a `Lookup` object to do the test.
```
	private static final Class<?> resolveFieldHandleHelper(String typeDescriptor, Lookup lookup, ClassLoader loader) throws Throwable {
		MethodType mt = MethodType.vmResolveFromMethodDescriptorString()...
		lookup.accessCheckArgRetTypes(mt);
		return mt.parameterType(0);
```

pdbain-ibm:
@DanHeidinga do you have further comments on this pull request?
Thanks
    -p

DanHeidinga:

Issue #2643, opened 2018-08-17
threadMXBeanTestSuite2_1 hangs


pdbain-ibm:
https://ci.eclipse.org/openj9/job/Test-extended.functional-JDK10-linux_x86-64/48/console:

```
01:11:17 ===============================================
01:11:17 Running test threadMXBeanTestSuite2_1 ...
01:11:17 ===============================================
01:11:17 threadMXBeanTestSuite2_1 Start Time: Fri Aug 17 05:11:17 2018 Epoch Time (ms): 1534482677791
01:11:17 test with Mode301
01:11:17 { "/home/jenkins/workspace/Test-extended.functional-JDK10-linux_x86-64/openjdkbinary/j2sdk-image/bin/java"  -Xnocompressedrefs -Xgcpolicy:metronome -Xnocompressedrefs \
01:11:17 -cp "/home/jenkins/workspace/Test-extended.functional-JDK10-linux_x86-64/openjdk-tests/TestConfig/scripts/testKitGen/../../../../jvmtest/TestConfig/resources:/home/jenkins/workspace/Test-extended.functional-JDK10-linux_x86-64/openjdk-tests/TestConfig/scripts/testKitGen/../../../../jvmtest/TestConfig/lib/testng.jar:/home/jenkins/workspace/Test-extended.functional-JDK10-linux_x86-64/openjdk-tests/TestConfig/scripts/testKitGen/../../../../jvmtest/TestConfig/lib/jcommander.jar:/home/jenkins/workspace/Test-extended.functional-JDK10-linux_x86-64/openjdk-tests/TestConfig/scripts/testKitGen/../../../../jvmtest/functional/Java8andUp/GeneralTest.jar" \
01:11:17 org.testng.TestNG -d "/home/jenkins/workspace/Test-extended.functional-JDK10-linux_x86-64/openjdk-tests/TestConfig/scripts/testKitGen/../../../TestConfig/test_output_15344816822623/threadMXBeanTestSuite2_1" "/home/jenkins/workspace/Test-extended.functional-JDK10-linux_x86-64/openjdk-tests/TestConfig/scripts/testKitGen/../../../../jvmtest/functional/Java8andUp/testng.xml" \
01:11:17 -testnames \
01:11:17 threadMXBeanTestSuite2 \
01:11:17 -groups level.extended \
01:11:17 -excludegroups d.*.linux_x86-64,d.*.arch.x86,d.*.os.linux,d.*.bits.64,d.*.generic-all; \
[...]

PR #2677, opened 2018-08-22
Fix hang with metronome exclusive hand-off

gacholio:
releaseExclusiveVMAccessFromExternalThread (mostly used by the metronome
GC policy) hands off to the normal exclusive while leaving
J9_PUBLIC_FLAGS_NOT_COUNTED_BY_EXCLUSIVE set in the target, which causes
a hang on the next hand off.

Fixes: #2643


gacholio:
Suggest testing on xlinux as metronome is tested there.

DanHeidinga:
The failure in the extended build is in `generalExtendedTest_0` due to `[TestNG] No tests found. Nothing was run`.  This is a known issue being investigated under https://github.com/eclipse/openj9/issues/2644

PR #2680, opened 2018-08-22
Rework classfile version checking

pdbain-ibm:
Reject a classfile if its major version is too large or less than 45,
or its major version is greater than 45 and its minor version is non-zero.


DanHeidinga:
I'd rather leave this code alone until it needs to be modified to support incubator classfile versions.

pshipton:
@DanHeidinga Apparently a fix such as this is needed to resolve a JCK 11 failure.

pdbain-ibm:
Per discussion with @DanHeidinga I'll combine this with http://openjdk.java.net/jeps/12 "JEP 12: Preview Language and VM Features".

> A class file denotes that it depends on the preview VM features of Java SE $N by having a major_version item that corresponds to Java SE $N and a minor_version item that has all 16 bits set. For example, a class file that depends on the preview VM features of Java SE 11 would have version 55.65535.

pdbain-ibm:
Updated and manually tested:
```
$ cfdump ~/HelloWorld.class 
public class HelloWorld extends java.lang.Object 
{
    public void <init>();
    public static void main(java.lang.String[]);
}
$java  HelloWorld
Error: LinkageError occurred while loading main class HelloWorld
	java.lang.UnsupportedClassVersionError: JVMCFRE154 Class file is a preview version but has the wrong major version or preview is not enabled.; class=HelloWorld, offset=6
$ java --enable-preview HelloWorld
Hello, world!
```
@DanHeidinga your review would be appreciated.

runtime/cfdumper/main.c
```diff
@@ -2304,7 +2304,7 @@ static U_32 buildFlags(void)
 	/* Set vm flags to the latest version to pass version validation
 	 * check in j9bcutil_readClassFileBytes.
 	 */
-	flags |= BCT_JavaMaxMajorVersionShifted;
+	flags |= BCT_JavaMaxMajorVersionShifted | BCT_AnyPreviewVersion | BCT_EnablePreview;
```
DanHeidinga: Does this need both `BCT_AnyPreviewVersion | BCT_EnablePreview`?  I would have expected `BCT_AnyPreviewVersion` to be sufficient

runtime/nls/cfre/cfrerr.nls
```diff
@@ -1362,3 +1362,10 @@ J9NLS_CFR_ERR_CP_ENTRY_INVALID_BEFORE_V55.explanation=Please consult the Java Vi
 J9NLS_CFR_ERR_CP_ENTRY_INVALID_BEFORE_V55.system_action=The JVM will throw a verification or classloading-related exception such as java.lang.ClassFormatError.
 J9NLS_CFR_ERR_CP_ENTRY_INVALID_BEFORE_V55.user_response=Contact the provider of the classfile for a corrected version.
 # END NON-TRANSLATABLE
+
+J9NLS_CFR_ERR_PREVIEW_VERSION=Class file is a preview version but has the wrong major version or preview is not enabled.
+# START NON-TRANSLATABLE
+J9NLS_CFR_ERR_MINOR_VERSION.explanation=Classfiles with minor version 65535 must have the same major version as the VM rumtime version and preview must be enabled
+J9NLS_CFR_ERR_MINOR_VERSION.system_action=The JVM will throw a verification or classloading-related exception.
+J9NLS_CFR_ERR_MINOR_VERSION.user_response=Run with the correct Java runtime version and use --enable-preview.
+
```
DanHeidinga: Missing the `# END NON-TRANSLATABLE`

runtime/vm/jvminit.c
```diff
@@ -2035,6 +2035,11 @@ IDATA VMInitStages(J9JavaVM *vm, IDATA stage, void* reserved) {
 				 */
 				vm->runtimeFlags |= J9_RUNTIME_XFUTURE;
 			}
+
+			if (FIND_AND_CONSUME_ARG(EXACT_MATCH, VMOPT_ENABLE_PREVIEW, NULL) >= 0) {
```
DanHeidinga: This option should only be recognized on jdk11+

runtime/bcutil/cfreader.c
```diff
@@ -2131,51 +2131,44 @@ checkAttributes(J9CfrClassFile* classfile, J9CfrAttribute** attributes, U_32 att
...
-		- The 48.* version number identifies JDK 1.4 class files."
-
-	- 49.* are JDK 5.0 class files (aka JDK 1.5)
-
-	Returns -1 on error, 0 on success.
-*/
+ *	Error if:
+ *	- major version < 45
+ *	- major version > maximum allowed
+ *	- major version > 45 and minor version != 0 or 0xFFFF
```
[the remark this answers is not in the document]
pdbain-ibm: Yes.  The code ignores the minor version for major version 45.

runtime/cfdumper/main.c
pdbain-ibm: True.  Fixed.

runtime/nls/cfre/cfrerr.nls
pdbain-ibm: Fixed.

runtime/vm/jvminit.c
pdbain-ibm: Okay, fixed.

Also removed from `checkClassVersion`: `(classfile->majorVersion >= 55)`.  

If the Java version is less than 11, we will fail either `(classfile->majorVersion > max_allowed_version)` because the max_allowed_version is < 55, or `J9_ARE_NO_BITS_SET(flags, BCT_EnablePreview) || (classfile->majorVersion != max_allowed_version)` because the major version is not the runtime version (i.e. illegally less than 55) or `--enable-preview` is disabled.  cfdump can dump preview files with bogus major versions but I think that's okay.

pdbain-ibm:
Code updated. @DanHeidinga could you take another look?

DanHeidinga:
I'm hesitant on this for two reasons, one of them very minor:
1) I've misread the comment on the 45.3 case a number of times.  Its subtle when it doesn't need to be.  (This is the minor one)

2) There is a behaviour change in this code that effects JDK8.  It suddenly makes versions that were previously accepted illegal.  Whether this will hit anyone in practice is unclear.  As is how it compares to the RIs behaviour.

pdbain-ibm:
Will do more testing with the RI and Java 8.

pdbain-ibm:
Change in behaviour from Java 8 to Java 11:
Java 8 accepts non-zero minor versions prior to major version 52 (Java 8).  Java 11 rejects non-zero minor versions for major versions greater than 45.

pdbain-ibm:
Tested and working with both Java 8 and Java 11 class file version tests.

pdbain-ibm:
@DanHeidinga Updated and tested. Your further comments are welcome.

pdbain-ibm:
Rebased, merge conflicts fixed, copyright dates updated to 2019.

DanHeidinga:
Rebased to resolve conflicts.

Pushed an attempt to clarify the classfile version check.  I'd still like to write a test to validate all the minor version cases are handled correctly before merging this.

DanHeidinga:

DanHeidinga:

DanHeidinga:
Jenkins copyright check

DanHeidinga:
Jenkins line endings check

DanHeidinga:
Wrote the following class that attempts to load a simple class of each major version with a 0 minor version and `.3` minor version:
```java
import org.objectweb.asm.*;
import org.objectweb.asm.ClassWriter;
import org.objectweb.asm.Opcodes;

public class VersionLoader extends ClassLoader implements Opcodes {

	VersionLoader vl = new VersionLoader();
	byte[] majors = {45, 46, 47, 48, 49, 50, 51, 52, 53, 54, 55};
	byte[] clazz = generate(45, "A");
	for (int i = 0; i < majors.length; i++) {
		clazz[7] = majors[i];
		try {
			System.out.println(new VersionLoader().defineClass(clazz, 0, clazz.length) + " " + majors[i]);
		} catch(ClassFormatError cfe) {
			System.out.println(" ClassFormatError ");
		}
	}
	for (int i = 0; i < majors.length; i++) {
		clazz[5] = 3;
		clazz[7] = majors[i];
		System.out.println(" " + majors[i] + " " + 3);
		try {
			System.out.println(" " + new VersionLoader().defineClass(clazz, 0, clazz.length) + " " + majors[i] + " " + 3);
		} catch(ClassFormatError cfe) {
			System.out.println(" ClassFormatError ");
		}
	}

}


static byte[] generate(int version, String name) {
	ClassWriter classWriter = new ClassWriter(ClassWriter.COMPUTE_MAXS | ClassWriter.COMPUTE_FRAMES);
	classWriter.visit(version, ACC_PUBLIC | ACC_SUPER, name, null, "java/lang/Object", null);
	classWriter.visitEnd();
	return classWriter.toByteArray();
}

}
```

DanHeidinga:
Hotspot JDK8 | Hotspot 11 | OpenJ9 11
 ------------|-------------|--------
 45 class A | 45 class A | 45 class A 
 46 class A | 46 class A | 46 class A 
 47 class A | 47 class A | 47 class A 
 48 class A | 48 class A | 48 class A 
 49 class A | 49 class A | 49 class A 
 50 class A | 50 class A | 50 class A 
 51 class A | 51 class A | 51 class A 
 52 class A | 52 class A | 52 class A 
 53 ClassFormatError | 53 class A |  53 class A
 54 ClassFormatError | 54 class A |  54 class A
 55 ClassFormatError | 55 class A |  54 class A
 45 3 class A | 45 3 class A |  45 3 class A
 46 3 class A  |  46 3 ClassFormatError |  46 3 class A
 47 3 class A  | 47 3 ClassFormatError  |  47 3 class A
 48 3 class A  | 48 3 ClassFormatError  | 48 3 class A
 49 3 class A  | 49 3 ClassFormatError  |  49 3 class A
 50 3 class A  | 50 3 ClassFormatError  |  50 3 class A
 51 3 class A  | 51 3 ClassFormatError |  51 3 class A
 52 3 ClassFormatError |  52 3 ClassFormatError |  52 3 ClassFormatError
 53 3 ClassFormatError | 53 3 ClassFormatError |  53 3 ClassFormatError
 54 3 ClassFormatError | 54 3 ClassFormatError |  54 3 ClassFormatError
 55 3 ClassFormatError | 55 3 ClassFormatError |  55 3 ClassFormatError

DanHeidinga:
Hotspot's behaviour in JDK11 as shown above doesn't appear to match the spec for dealing with non-0 minor versions:

> For a class file whose major_version is 56 or above, the minor_version must be 0 or 65535.
> For a class file whose major_version is between 45 and 55 inclusive, the minor_version may be any value.
http://cr.openjdk.java.net/~iris/se/12/latestSpec/java-se-12-jvms-diffs.pdf

DanHeidinga:
I've done some further clean up on this code and validated the `--enable-preview` works for local testing.

I think this is ready for PR testing and then merging

DanHeidinga:

DanHeidinga:
Jenkins test extended zlinux jdk11

runtime/nls/cfre/cfrerr.nls
```diff
@@ -1444,5 +1444,10 @@ J9NLS_CFR_ERR_VALUE_TYPES_IS_NOT_SUPPORTED.explanation=ValueTypes is an experime
 J9NLS_CFR_ERR_VALUE_TYPES_IS_NOT_SUPPORTED.system_action=The JVM will throw a verification or classloading-related exception such as java.lang.ClassFormatError.
 J9NLS_CFR_ERR_VALUE_TYPES_IS_NOT_SUPPORTED.user_response=Contact the provider of the classfile for a corrected version or use the -XX:+EnableValhalla option.
 
+J9NLS_CFR_ERR_PREVIEW_VERSION=Class file is a preview version but has the wrong major version or preview is not enabled.
```
pshipton: This should be J9NLS_CFR_ERR_MINOR_VERSION

runtime/bcutil/cfreader.c
```diff
@@ -2168,44 +2168,56 @@ checkAttributes(J9CfrClassFile* classfile, J9CfrAttribute** attributes, U_32 att
...
+ * 	
+ *	To be more precise, throw an error if:
+ *	- major version < 45
+ *	- or major version > maximum allowed
+ *	- or if the runtime version is Java 8 (major version 52) and
+ *		-  major version is 52 and minor version is non-zero
+ *	- or if the runtime version is greater than Java 11 and:
+ *		- [major version > 45] and [minor version is not 0 or -1 (0xFFFF)]
+ *		- or minor version == 0xffff and
+ *			(--enable-preview is not set or major version != runtime version or major version < 55)
```
pshipton: I don't think the case "runtime version is greater than Java 11" and "minor version == 0xffff" and "major version < 55" is reflected in the code. For example, is runtime version is Java 13 and major version is 55 or earlier, an error will not occur.

DanHeidinga:
I've updated the comment and fixed the NLS file

pshipton:

PR #2689, opened 2018-08-23
Make MethodType.lastParameter() public

pdbain-ibm:
Make the method public to comply with the API spec.  Return void instead of
throwing an exception if the parameter list is empty and update callers who
catch the exception.

### DIFF
--- a/jcl/src/java.base/share/classes/java/lang/invoke/MethodType.java
+++ b/jcl/src/java.base/share/classes/java/lang/invoke/MethodType.java
@@ -901,14 +901,21 @@ public final class MethodType implements Serializable {
 		return returnType;
 	}
 	
-	/* Return the last class in the parameterType array.
+	/**
+	 * Return the last class in the parameterType array.
 	 * This is equivalent to:
 	 * 		type.parameterType(type.parameterCount() - 1)
-	 * Will throw ArrayIndexOutOfBounds on bad indexes.  This is a subclass of IndexOutOfBoundsException
-	 * so this is valid when JCKs expect certain exceptions.
+	 * @return class of final parameter, or void.class if there are no parameters
 	 */
-	/* package */ Class<?> lastParameterType() {
-		 return arguments[arguments.length - 1];
+	/*[IF Java10]*/
+	public
+	/*[ENDIF]*/
+	Class<?> lastParameterType() {
+		Class<?> result = void.class;
+		if (arguments.length > 0) {
+			result = arguments[arguments.length - 1];
+		}
+		return result;
 	}
 	
 	/**
--- a/jcl/src/java.base/share/classes/java/lang/invoke/MethodHandle.java
+++ b/jcl/src/java.base/share/classes/java/lang/invoke/MethodHandle.java
@@ -679,11 +679,8 @@ public abstract class MethodHandle {
 		if (isVarArityNeeded) {
 			/* Convert the handle if initially not with variable arity */
 			if (!this.isVarargsCollector()) {
-				Class<?> lastClass = null;
-				try {
-					/* An exception is thrown if the MethodType has no arguments */
-					lastClass = this.type.lastParameterType();
-				} catch(RuntimeException e) {
+				Class<?> lastClass = this.type.lastParameterType();
+				if (void.class == lastClass) { /* there were no arguments */
 					throw new IllegalArgumentException();
 				}
 				/* IllegalArgumentException will be thrown out from asVarargsCollector 


jcl/src/java.base/share/classes/java/lang/invoke/MethodType.java
```diff
@@ -891,14 +891,23 @@ public int parameterCount(){
...
-	 * so this is valid when JCKs expect certain exceptions.
+	 * @return class of final parameter, or void.class for Java 10 and up
 	 */
-	/* package */ Class<?> lastParameterType() {
-		 return arguments[arguments.length - 1];
+	/*[IF Java10]*/
+	public
+	/*[ENDIF]*/
+	Class<?> lastParameterType() {
+		Class<?> result = null;
```
DanHeidinga: `Class<?> result = void.class` and then you don't need the `else` block

pdbain-ibm:
Updated per 
https://github.com/eclipse/openj9/pull/2689#discussion_r212431351.

There are ~8 callers.  Only `withVarargs()` checks the exception; I updated it to check the return value instead.

pdbain-ibm:
Tested and passing the relevant test suite.

DanHeidinga:
> There are ~8 callers.

Are these other 8 callers depending on the exception?

DanHeidinga:
> No, the exception has no effect on the execution of the other (direct) callers of lastArgument().

Looking at a few of them and this one is suspect - what if the DirectHandle has no arguments?
https://github.com/eclipse/openj9/blob/2f295ea91cdd97ff37eb5d164fc63014ea718bc7/jcl/src/java.base/share/classes/java/lang/invoke/PrimitiveHandle.java#L176-L185

DanHeidinga:
Here's an example that shows that `lastParameterType()`'s current AIOOBE does affect current callers.  More investigation needed as to whether it *should*, though the comments on it certainly imply it's required.

```java
import java.lang.invoke.*;
import java.lang.invoke.MethodHandles.*;

public class Varargs {

	MethodHandle mh = MethodHandles.lookup().findStatic(Varargs.class, "test", MethodType.methodType(void.class));
	System.out.println(mh);
	mh.asVarargsCollector(Object[].class);
}

public static void test() { }
}

```

```
$ java Varargs 
MethodHandle()void
Exception in thread "main" java.lang.ArrayIndexOutOfBoundsException: Array index out of range: -1
	at java.lang.invoke.MethodType.lastParameterType(MethodType.java:842)
	at java.lang.invoke.PrimitiveHandle.asVarargsCollector(PrimitiveHandle.java:179)
	at Varargs.main(Varargs.java:10)
```

pdbain-ibm:
@DanHeidinga Thanks for the test case.  I tried it with various makes and models of the JVM:

- OpenJDK 9 throws `java.lang.IllegalArgumentException: array type not assignable to trailing argument: MethodHandle()void, class [Ljava.lang.Object;`
- OpenJDK 10 throws ` java.lang.IllegalArgumentException: bad collect position`
- OpenJ9 Java 8 throws `java.lang.ArrayIndexOutOfBoundsException: Array index out of range: -1`

The public methods which call `lastParameterType()` are:
`asVarargsCollector()` (`PrimitiveHandle` and `MethodHandle`):
```
		Class<?> lastArgType = type().lastParameterType();
		if (!lastArgType.isAssignableFrom(arrayParameter)) {
			throw new IllegalArgumentException();
		}
```

`withVarargs()`:
```
				Class<?> lastClass = this.type.lastParameterType();
				if (void.class == lastClass) { /* there were no arguments */
					throw new IllegalArgumentException();
				}
```

In `VarargsCollector.asType()` this change should prevent a spurious exception:
```
		if (parameterCount == newTypeParameterCount) {
Possible AIOOB here if type and newType have no parameters
			if (type.lastParameterType().isAssignableFrom(newType.lastParameterType())) {
				return next.asType(newType);
			}
		}
		int collectCount = newTypeParameterCount - parameterCount + 1;
		if (collectCount < 0) {
			/*[IF ]*/
			/* Expected by the JCK - should be allowed to just let asCollector throw the IAE */
			/*[ENDIF]*/
			throw new WrongMethodTypeException();
		}
```

pdbain-ibm:
Tested with my changes:
```
$java Varargs
MethodHandle()void
java.lang.IllegalArgumentException
	at java.lang.invoke.PrimitiveHandle.asVarargsCollector(java.base@10.0.2-internal/PrimitiveHandle.java:181)
	at Varargs.main(Varargs.java:10)
```
which bring OpenJ9 in line with the reference implementation.

pdbain-ibm:
`MethodHandle.withVarargs()` formerly caught the AIOOB and threw

jcl/src/java.base/share/classes/java/lang/invoke/MethodType.java
```diff
@@ -901,14 +901,21 @@ public int parameterCount(){
...
 	}
 	
-	/* Return the last class in the parameterType array.
+	/**
+	 * Return the last class in the parameterType array.
 	 * This is equivalent to:
 	 * 		type.parameterType(type.parameterCount() - 1)
-	 * Will throw ArrayIndexOutOfBounds on bad indexes.  This is a subclass of IndexOutOfBoundsException
-	 * so this is valid when JCKs expect certain exceptions.
+	 * @return class of final parameter, or void.class for Java 10 and up
```
DanHeidinga: `@return class of final parameter, or void.class for Java 10 and up` -> `@return class of final parameter, or void.class if there are no parameters`

jcl/src/java.base/share/classes/java/lang/invoke/MethodType.java
```diff
@@ -901,14 +901,21 @@ public int parameterCount(){
...
 	 */
-	/* package */ Class<?> lastParameterType() {
-		 return arguments[arguments.length - 1];
+	/*[IF Java10]*/
+	public
+	/*[ENDIF]*/
+	Class<?> lastParameterType() {
+		Class<?> result  = void.class;
```
DanHeidinga: minor nitpick: extra space before the `=`.  Only worth calling out as the comment also needs to be updated.

pdbain-ibm:
Updated per @DanHeidinga 's comments.

PR #2690, opened 2018-08-23
Optimize JIT transition on PPC

gacholio:
Initialize pseudoTOC on call-in, rather than on every transition.


DanHeidinga:

PR #2695, opened 2018-08-23
Remove tag bits during JIT interface resolution

gacholio:
jitResolveInterfaceMethod is not removing the tag bits from
methodIndexAndArgCount, resulting in wildly large indices.


gacholio:

PR #2697, opened 2018-08-24
Get StringBuilder capacity again after decompress

msqb:
StringBuilder's capicity field is updated after a decompress()
call is made. StringBuilder methods that store the capacity in
a local variable before a call to decompress() should update this
stored value again after the call. This applies only to StringBuilder
methods that use the stored capacity value after a decompress() call.

As an example, if this value is not updated again after decompress(),
the old value will be used as a check on whether or not to run
ensureCapacityImpl(), leading to additional array copies.

decompress() now returns the new capacity, for which the sharedBit
need not be masked off since decompress() creates a new array.

### DIFF
--- a/jcl/src/java.base/share/classes/java/lang/StringBuilder.java
+++ b/jcl/src/java.base/share/classes/java/lang/StringBuilder.java
@@ -78,8 +78,8 @@ public final class StringBuilder extends AbstractStringBuilder implements Serial
 	private transient int count;
 	private transient char[] value;
 	private transient int capacity;
-	
-	private void decompress(int min) {
+
+	private int decompress(int min) {
 		int currentLength = lengthInternal();
 		int currentCapacity = capacityInternal();
 		
@@ -100,6 +100,7 @@ public final class StringBuilder extends AbstractStringBuilder implements Serial
 		capacity = newValue.length;
 		
 		String.initCompressionFlag();
+		return capacity;
 	}
 	
 /**
@@ -208,7 +209,7 @@ public StringBuilder append (char[] chars) {
 		} else {
 			// Check if the StringBuilder is compressed
 			if (count >= 0) {
-				decompress(newLength);
+				currentCapacity = decompress(newLength);
 			}
 			
 			if (newLength > currentCapacity) {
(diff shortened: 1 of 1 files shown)


msqb:
This change was suggested by @DanHeidinga in https://github.com/eclipse/openj9/pull/2433:
> currentCapacity may not be correct here if a decompress() occurred. Because this uses ensureCapacityImpl, this may result in additional array copies

All other methods having the same pattern of using the locally stored capacity after a call to decompress() have been updated as well.

msqb:
I will update this post with the performance implications and remove the WIP from the title thereafter.

andrewcraik:
To be clear having an idea of the performance impact is good, but since this is a correctness issue the change will be required regardless unless a 'cheaper' alternative is available.

msqb:
Performance analysis was conducted using the DayTrader 3 and DayTrader 7 benchmarks.
Java build used: OpenJDK8-OPENJ9_x64_Linux_201818071200

DayTrader 3 Throughput (higher=better), average of 4 runs:
```
            mean        %stdev
baseline    [baseline]  0.276%
change      +0.390%     0.450%
```

DayTrader 7 Throughput (higher=better), average of 4 runs:
```
            mean        %stdev
baseline    [baseline]  1.312%
change      +0.598%     1.457%
```

The performance of the new change remains on par with the baseline

jcl/src/java.base/share/classes/java/lang/StringBuilder.java
```diff
@@ -209,6 +209,7 @@ public StringBuilder append (char[] chars) {
 			// Check if the StringBuilder is compressed
 			if (count >= 0) {
 				decompress(newLength);
+				currentCapacity = capacityInternal();
```
fjeremic: Could we instead modify the `decompress` method to return the new capacity? We can then avoid these additional calls to `capacityInternal()` since the `decompress` method already knows what the new capacity will be.

jcl/src/java.base/share/classes/java/lang/StringBuilder.java
```diff
@@ -208,7 +209,7 @@ public StringBuilder append (char[] chars) {
 		} else {
 			// Check if the StringBuilder is compressed
 			if (count >= 0) {
-				decompress(newLength);
+				currentCapacity = decompress(newLength) & ~sharedBit;
```
fjeremic: Why do we need to mask off the shared bit? `decompress` should always allocate a new array, meaning the array should never be shared at that point.

msqb: Updated the PR to removing masking this bit

fjeremic:

PR #2709, opened 2018-08-27
[mac] Use J9PORT_LIBRARY_SUFFIX instead of ".so" when possible

DanHeidinga:
OSX and Unix versions are mostly compatible if .dylib replaces
.so in sharedlibrary names.  Some of the code, ie jsig,
can't see j9port.h and so has been changed by hand

issue: #36


gacholio:

Issue #2710, opened 2018-08-27
JSR 292 AdaptorTests tests fail with JDK11 due to nestmates


DanHeidinga:
https://github.com/eclipse/openj9/blob/6ce77a0657a28a7ad7b1021d7cf2ac556a39a32a/test/functional/Jsr292/src/com/ibm/j9/jsr292/AdaptorTests.java#L149-L171

This test is no longer valid if compiled with a JDK11 build as the `PrivateMethod` class will be marked as a nest mate of the `AdapterTest` case which allows the private access.

DanHeidinga:
Test is passing now

PR #2711, opened 2018-08-27
Fix JDK11 test failure by ensuring class not a nestmember

DanHeidinga:
PrivateMethod was declared as an inner class of AdaptorTests
which makes it a nest member in JDK11.  Nest members have private
access in JDK11+ which defeats the point of the nest.

This change makes the two classes siblings rather than nestmates.

issue: #2710


DanHeidinga:

DanHeidinga:
@llxia Can you review this?

llxia:
Jenkins test extended xlinux all

PR #2712, opened 2018-08-27
Fix toggling nextSlot value in ConstantPoolObjectSlotIterator.hpp

Ailloviee:
Added toggling `nextSlot` value back to `true` in `scanCondySlot()` when
`_condySlotState` is `condy_slot_exception`, so that the iterator will
go on to the next slot in constant pool if the value of the condy is
`NULL`


Ailloviee:
@dmitripivkine could you review this please?

dmitripivkine:

DanHeidinga:

DanHeidinga:
Launched a JDK8 build as the the 11 builds are still in rough shape

PR #2723, opened 2018-08-28
Fix TLH in Disabled mode path

dmitripivkine:
There are two problems fixed:
- env->isInlineTLHAllocateEnabled() should took to consideration both
TLHs in dual-TLH mode and can not based on zeroed TLH only.
- J9AllocateObjectNoGC and J9AllocateIndexableObjectNoGC should not try
to allocate if TLH is in Disabled state. An attempt to do this might
refresh disabled TLH and re-enable it implicitly
 

dmitripivkine:
Jenkins test extended pLinux jdk8

dmitripivkine:

AdamBrousseau:
Jenkins test extended pLinux jdk8

PR #2727, opened 2018-08-28
Decode new IPIC iTable offset slot in stack walk

jdmpapin:
When encountering a JIT resolve frame for jitLookupInterfaceMethod, the
stack walk uses the itable index to determine the signature and number
of parameters of the callee. However, the IPIC slot is no longer an
itable index. It's now typically an itable offset, although it can also
be a tagged J9Method pointer or VFT offset in case of private methods or
Object methods.

### DIFF
--- a/runtime/codert_vm/jswalk.c
+++ b/runtime/codert_vm/jswalk.c
@@ -1100,12 +1100,23 @@ static void jitWalkResolveMethodFrame(J9StackWalkState *walkState)
 	} else if (resolveFrameType == J9_STACK_FLAGS_JIT_LOOKUP_RESOLVE) {
 		UDATA * interfaceObjectAndISlot = (UDATA *) JIT_RESOLVE_PARM(2);
 		J9Class * interfaceClass = READ_CLASS((J9Class *) READ_UDATA(interfaceObjectAndISlot));
-		UDATA methodIndex = READ_UDATA(interfaceObjectAndISlot + 1);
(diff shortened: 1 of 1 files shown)


gacholio:
Good catch - a matching change will need to be made in DDR. I'm not sure if the bits will be visible in DDR.

gacholio:
DDR code in question:

https://github.com/eclipse/openj9/blob/master/debugtools/DDR_VM/src/com/ibm/j9ddr/vm29/j9/stackwalker/JITStackWalker.java#L406

The constants should be available as we "import static J9Consts.*".

If you're unfamiliar with DDR, I can make the change in a separate PR - this case doesn't really ever come up.

jdmpapin:
@gacholio If you can change DDR separately, that would be a big help

gacholio:
No problem.

gacholio:
It occurs to me that the existing code is completely incorrect for inherited interface methods (a little while back I changed iTables such that they contain entries for all method from all inherited interfaces, rather than just the methods in the resolved interface class).

I'm going to fix that problem, so I may as well roll the other changes from here into that (and do the equivalent DDR work). I'll close this off once the work is complete if you don't object.

DanHeidinga:
@gacholio @jdmpapin Is this ready to go?

gacholio:
@DanHeidinga No - I'm going to fix it along with other things. For the record, this case essentially never comes up (only possibility of a stack walk is when an exception is thrown from jitLookupInterfaceMethod), so it's quite low-priority. It's already broken by my semi-recent change to expand the iTables to include inherited interfaces.

I'll close this to prevent confusion.

PR #2730, opened 2018-08-28
Move interface tag bits

gacholio:
Move the tag bits in J9RAMInterfaceMethodRef->methodIndexAndArgCount to
the low end so that the encoded index can be retrieved with a single
shift, rather than the mask and shift that is currently used.


charliegracie:

charliegracie:
Change is tested via the Sanity build so there is no sense in waiting for the travis build

PR #2733, opened 2018-08-28
Fix for incorrect test generated by ZEROCHK on X86

0dvictor:
ZEROCHK incorrectly generated 64-bit test instruction on X86-32.
In fact, it should always use 32-bit test instruction on both 32-bit and 64-bit,
since its only child must has type TR::Int32.
This changeset fixes the above defect..


gacholio:

PR #2735, opened 2018-08-28
Address diagnostics from gcc 8

keithc-ca:
* avoid possible bounds violations by using j9str_printf instead of sprintf

* Fix 'error=class-memaccess' diagnostics of the form:
```
error: 'void* memset(void*, int, size_t)' clearing an object with no trivial copy-assignment; use assignment or value-initialization instead [-Werror=class-memaccess]
```

* clean up: add missing 'static' modifiers; fix formatting

* correct lengths of predefined trace options

### DIFF
--- a/runtime/rastrace/trcengine.c
+++ b/runtime/rastrace/trcengine.c
@@ -633,11 +633,14 @@ initializeTraceOptions(J9JavaVM *vm, char* opts[])
 	IDATA optionsLength;
 	IDATA rc =  J9VMDLLMAIN_OK;
 
+#define trace_option_helper(option) \
+		splitCommandLineOption(vm, option, LITERAL_STRLEN(option), opts + i)
+
 	/* set up the default options */
-	rc = splitCommandLineOption(vm, UT_MAXIMAL_KEYWORD "=all{level1}", 20, opts);
+	rc = trace_option_helper(UT_MAXIMAL_KEYWORD "=all{level1}");
 	i += 2;
 	if (rc != J9VMDLLMAIN_FAILED) {
-		rc = splitCommandLineOption(vm, UT_EXCEPTION_KEYWORD "=j9mm{gclogger}", 26, opts + i);
+		rc = trace_option_helper(UT_EXCEPTION_KEYWORD "=j9mm{gclogger}");
 		i += 2;
 	}
 
@@ -653,7 +656,7 @@ initializeTraceOptions(J9JavaVM *vm, char* opts[])
 	 * trace they selected but their selection + level 2.)
 	 */
 	if( xtraceIndex >= 0 && !checkAndSetL2EnabledFlag() ) {
-		rc = splitCommandLineOption(vm, UT_MAXIMAL_KEYWORD "=all{level2}", 20, opts + i);
+		rc = trace_option_helper(UT_MAXIMAL_KEYWORD "=all{level2}");
 		i += 2;
 	}
 	while (xtraceIndex >= 0) {
@@ -675,6 +678,8 @@ initializeTraceOptions(J9JavaVM *vm, char* opts[])
 	}
 	opts[i++] = NULL;
 	return rc;
+
+#undef trace_option_helper
 }
 
 /*
--- a/runtime/rastrace/trcengine.c
+++ b/runtime/rastrace/trcengine.c
@@ -93,7 +93,7 @@ static void javaTraceTerm (void* env, UtModuleInfo *modInfo);
 
 typedef omr_error_t (*traceOptionFunction)(J9JavaVM *vm,const char * value,BOOLEAN atRuntime);
 
-/** 
+/**
  * Structure representing a trace option such as stackdepth
  * or trigger
  */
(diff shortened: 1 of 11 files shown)


keithc-ca:
With https://github.com/eclipse/omr/pull/2898 and these changes, I was able to build the VM on xLinux and zLinux with gcc 8.1. (Changes will be required to the extension repos before we can build the whole stack.)

runtime/tests/shared/CacheDirPerm.cpp
```diff
@@ -195,37 +190,36 @@ removeTempDir(J9JavaVM *vm, char *dir)
...
+		j9str_printf(PORTLIB, baseFilePath, J9SH_MAXPATH, "%s", dir);
+		rc = handle = j9file_findfirst(baseFilePath, resultBuffer);
+		while ((UDATA)-1 != rc) {
+			char nextEntry[J9SH_MAXPATH];
+			/* skip current and parent dir */
+			if (resultBuffer[0] == '.') {
+				rc = j9file_findnext(handle, resultBuffer);
+				continue;
+			}
+			j9str_printf(PORTLIB, nextEntry, J9SH_MAXPATH, "%s/%s", baseFilePath, resultBuffer);
```
gacholio: Should be sizeof(nextEntry) rather than J9SH_MAXPATH.

runtime/tests/shared/CacheDirPerm.cpp
```diff
@@ -195,37 +190,36 @@ removeTempDir(J9JavaVM *vm, char *dir)
...
-        	j9str_printf(PORTLIB, nextEntry, J9SH_MAXPATH, "%s/%s", baseFilePath, resultBuffer);
-        	removeTempDir(vm, nextEntry);
-        	rc = j9file_findnext(handle, resultBuffer);
-        }
-        if (handle != (UDATA)-1) {
-        	j9file_findclose(handle);
-        }
-        rc = j9file_unlinkdir(baseFilePath);
-        if ((UDATA)-1 == rc) {
+		j9str_printf(PORTLIB, baseFilePath, J9SH_MAXPATH, "%s", dir);
```
gacholio: Should be sizeof(baseFilePath) rather than J9SH_MAXPATH.

runtime/tests/shared/CacheDirPerm.cpp
```diff
@@ -195,37 +190,36 @@ removeTempDir(J9JavaVM *vm, char *dir)
...
 	}
 	rc = PASS;
 _end:
 	return rc;
-
 }
 
 IDATA
-testCacheDirPerm(J9JavaVM* vm)
+testCacheDirPerm(J9JavaVM *vm)
```
gacholio: These types of cosmetic changes serve no purpose other than to make reviews take longer.

runtime/tests/shared/CacheDirPerm.cpp
```diff
@@ -37,8 +36,8 @@ extern "C"
 #define J9SH_DEFAULT_CTRL_ROOT "/tmp"
 
 #define USER_PERMISSION_STR "0700"
-#define USER_PERMISSION		0700
-#define NEW_DIR_PERM	0760
+#define USER_PERMISSION      0700
+#define NEW_DIR_PERM         0760
```
gacholio: We've opted not to try and line text up like this, since it never works properly between the IDE and github. If you need to make this unnecessary change, I suggest just using a single space.

gacholio:
The two very large changes are essentially impossible to review as they have so many irrelevant changes.

runtime/tests/shared/CacheDirPerm.cpp
keithc-ca: Feel free to make use of the 'Hide whitespace changes' in the 'Diff settings' menu.

runtime/tests/shared/CacheDirPerm.cpp
keithc-ca: Fixed.

runtime/tests/shared/CacheDirPerm.cpp
keithc-ca: Fixed.

runtime/tests/shared/CacheDirPerm.cpp
gacholio: I realize the code was that way before you got here, but considering some of the changes are changing the sizes of the path arrays, this seems safest going forward.

runtime/rastrace/trcengine.c
```diff
@@ -635,17 +634,17 @@ initializeTraceOptions(J9JavaVM *vm, char* opts[])
 	IDATA rc =  J9VMDLLMAIN_OK;
 
 	/* set up the default options */
-	rc = splitCommandLineOption(vm, UT_MAXIMAL_KEYWORD "=all{level1}", 20, opts);
+	rc = splitCommandLineOption(vm, UT_MAXIMAL_KEYWORD "=all{level1}", 19, opts);
```
gacholio: Why are these numbers  changing?

gacholio: This code (from days gone by) is particularly idiotic, in that it uses a macro for some of the string, so there's no way to accurately predict the length. Given there's only a few uses of constant strings, perhaps it would be better to add a new function that strlen's the input and calls the general one.

keithc-ca: The numbers are being corrected; perhaps some part used to be longer and the length didn't get updated.

I will add a macro that can use `sizeof` to compute the length of the string literals.

gacholio: I don't think that will work - sizeof will include the trailing nul character. strlen of a constant string should be folded by the compiler.

keithc-ca: That's ok, I can subtract one. ;-)

gacholio: IIRC, we had issues on some compilers before, where sizeof wasn't consistent on strings. Given that the code clearly doesn't need the correct number, I suppose being off by one won't be terrible.

gacholio:

PR #2736, opened 2018-08-28
Fix warnings from VS2017 due to printf format mismatches

keithc-ca:


gacholio:
jenkins compile win jdk8

PR #2740, opened 2018-08-28
Add access barrier API to copy valueTypes

tajila:
Add access barrier API to copy valueTypes

Add access barrier API to copy valueTypes to copy valueTypes
given a src/dest object and src/dest offset. 

This PR is based on Eric's work in
https://github.com/eclipse/openj9/pull/1770

### DIFF
--- a/runtime/gc_base/ObjectAccessBarrier.cpp
+++ b/runtime/gc_base/ObjectAccessBarrier.cpp
@@ -1345,38 +1345,77 @@ MM_ObjectAccessBarrier::getLockwordAddress(J9VMThread *vmThread, J9Object *objec
  * The new object was just allocated inside the VM, so all fields are NULL.
  * @TODO This does not currently check if the fields that it is reading are volatile.
  */
-void 
+void
 MM_ObjectAccessBarrier::cloneObject(J9VMThread *vmThread, J9Object *srcObject, J9Object *destObject)
 {
+	copyObjectFields(vmThread, J9GC_J9OBJECT_CLAZZ(srcObject), srcObject, J9_OBJECT_HEADER_SIZE, destObject, J9_OBJECT_HEADER_SIZE);
+}
+
+/**
+ * Copy all of the fields of a value class instance to another value class instance.
+ * The source or destination may be a flattened value within another object, meaning
+ * srcOffset and destOffset need not be equal. This is based on cloneObject(...).
+ * @TODO This does not currently check if the fields that it is reading are volatile.
+ *
+ * @oaram objectClass The j9class.
+ * @param srcObject The object containing the value class instance fields being copied.
+ * @param srcOffset The offset of the value class instance fields in srcObject.
+ * @param destValue The object containing the value class instance fields being copied to.
+ * @param destOffset The offset of the value class instance fields in destObject.
+ */
+void
+MM_ObjectAccessBarrier::copyObjectFields(J9VMThread *vmThread, J9Class *objectClass, J9Object *srcObject, UDATA srcOffset, J9Object *destObject, UDATA destOffset)
+{
+	/* For valueTypes we currently do not make a distinction between values that only contain
+	 * primitives and values that may contain a reference (ie. value vs mixed-value
+	 * in packedObject terminology). As a result, we will treat all values as if
+	 * they may contain references. In the future this may change.
+	 *
+	 * Value types have no need or lockwords, however, they are still present in the
+	 * current implementation. For now we will just skip over them by specifying
+	 * appropriate offsets. We will also skip over the bit in the instance description.
+	 */
+	bool isValueType = J9_IS_J9CLASS_VALUETYPE(objectClass);
+
 	j9objectmonitor_t *lockwordAddress = NULL;
-	bool isDestObjectPreHashed = _extensions->objectModel.hasBeenHashed(destObject);
 	I_32 hashCode = 0;
-	if (isDestObjectPreHashed) {
-		hashCode = _extensions->objectModel.getObjectHashCode(vmThread->javaVM, destObject);
-	}	
-	
-	J9Class *clazz = J9GC_J9OBJECT_CLAZZ(srcObject);
-	UDATA *descriptionPtr = (UDATA *)clazz->instanceDescription;
+	bool isDestObjectPreHashed = false;
+
+	if (!isValueType) {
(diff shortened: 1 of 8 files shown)


tajila:
@amicic please review

runtime/gc_base/ObjectAccessBarrier.cpp
```diff
@@ -1410,6 +1410,68 @@ MM_ObjectAccessBarrier::cloneObject(J9VMThread *vmThread, J9Object *srcObject, J
...
+		descriptionBits >>= 1;
+		if(descriptionIndex-- == 0) {
+			descriptionBits = *descriptionPtr++;
+			descriptionIndex = J9_OBJECT_DESCRIPTION_SIZE - 1;
+		}
+		offset += sizeof(fj9object_t);
+	}
+}
+#endif /* defined(J9VM_OPT_VALHALLA_VALUE_TYPES) */
+
```
amicic: Am I right in observing this is a subset of that cloneObject does. Could we share this code within a private method and call it both by cloneObject (with src/dest offsets being 0) and copyValue?

runtime/gc_include/ObjectAccessBarrierAPI.hpp
```diff
@@ -396,6 +396,54 @@ class MM_ObjectAccessBarrierAPI
...
+#if defined(J9VM_OPT_VALHALLA_VALUE_TYPES)
+	VMINLINE UDATA
+	mixedObjectGetFieldSize(J9Class *objectClass)
+	{
+		UDATA size = objectClass->totalInstanceSize;
+		if (0 <= (IDATA)objectClass->lockOffset) {
+			size -= sizeof(j9objectmonitor_t);
+		}
+		return size;
+	}
```
amicic: how about just expanding mixedObjectGetDataSize with another bool argument 'include/exclude' lockwork that would default to 'include'?

runtime/gc_include/ObjectAccessBarrierAPI.hpp
```diff
@@ -396,6 +396,54 @@ class MM_ObjectAccessBarrierAPI
...
+			while (offset < limit) {
+				*(fj9object_t *)((UDATA)destObject + offset + destOffset) = *(fj9object_t *)((UDATA)srcObject + offset + srcOffset);
+				offset += sizeof(fj9object_t);
+			}
+			postBatchStoreObject(vmThread, destObject);
+		}
+#endif /* defined(J9VM_GC_ALWAYS_CALL_OBJECT_ACCESS_BARRIER) */
+	}
+#endif /* defined(J9VM_OPT_VALHALLA_VALUE_TYPES) */
+
```
amicic: again, perhaps we can share it with cloneObject, having a private (it's fine to be inlined, the point is not on saving footprint, but on having more maintainable code) method with an additional argument if lockwork should be copied or not.

tajila:
@amicic please take another look

runtime/gc_base/ObjectAccessBarrier.cpp
```diff
@@ -1345,38 +1345,77 @@ MM_ObjectAccessBarrier::getLockwordAddress(J9VMThread *vmThread, J9Object *objec
  * The new object was just allocated inside the VM, so all fields are NULL.
  * @TODO This does not currently check if the fields that it is reading are volatile.
  */
-void 
+void
 MM_ObjectAccessBarrier::cloneObject(J9VMThread *vmThread, J9Object *srcObject, J9Object *destObject)
 {
+	copyObject(vmThread, J9GC_J9OBJECT_CLAZZ(srcObject), srcObject, sizeof(J9Object), destObject, sizeof(J9Object));
+}
```
amicic: we have this macro: J9_OBJECT_HEADER_SIZE

amicic: I see that original code was using sizeof(J9Object), then it should be ok,  too.

amicic:
looks ok to me, except the method name copyObject (it was copyValue, and than you renamed, right?). either of the two is not self-descriptive enough, and creates confusion in comparison with cloneObject. How about copyObjectPayload (it's not much better even to me tbh)?

@charliegracie, please have a look, too.

tajila:
@amicic I renamed it to `copyObjectFields`

@charliegracie please review this PR

amicic:

tajila:
The CI failure is caused by `The job exceeded the maximum time limit for jobs, and has been terminated.`

runtime/gc_include/ObjectAccessBarrierAPI.hpp
```diff
@@ -357,41 +357,70 @@ class MM_ObjectAccessBarrierAPI
...
+			UDATA limit = mixedObjectGetDataSize(objectClass, isValueType);
+			
 			while (offset < limit) {
-				/* No need for pre-store barrier as we are always overwriting 0's in the new object */
-				*(fj9object_t *)((UDATA)copy + offset) = *(fj9object_t *)((UDATA)original + offset);
+				*(fj9object_t *)((UDATA)destObject + offset + destOffset) = *(fj9object_t *)((UDATA)srcObject + offset + srcOffset);
 				offset += sizeof(fj9object_t);
 			}
-		
+
```
charliegracie: Can we remove the extra tab please?

runtime/gc_include/ObjectAccessBarrierAPI.hpp
```diff
@@ -357,41 +357,70 @@ class MM_ObjectAccessBarrierAPI
...
-		
+			
 #if defined(J9VM_THR_LOCK_NURSERY)
-			/* zero lockword, if present */
-			j9objectmonitor_t *lockwordAddress = getLockwordAddress(copy);
-			if (NULL != lockwordAddress) {
-				*lockwordAddress = 0;
+			if (!isValueType) {
+				/* zero lockword, if present */
+				j9objectmonitor_t *lockwordAddress = getLockwordAddress(destObject);
```
charliegracie: hmm for lockword I am wondering if it makes more sense for getLockwordAddress to just return NULL if the object passed in is a ValueType... That might simply things since all callers should be checking NULL now anyway.. It should not be any different at runtime but it should be safer for callers of getLockwordAddress that may not know about ValueTypes.

amicic:

PR #2745, opened 2018-08-29
Fix crash in gc_rwlocktest

gacholio:
Forked threads may still be running when the test exits. Fix threads to
ensure they are dead before their final check-in with the main thread.


gacholio:
To be a complete fix, omrthread_exit should be changed to release the monitor later, or under some kind of internal lock that prevents the thread library from being destroyed before the thread exits.

charliegracie:

gacholio:

gacholio:
omthread_exit fix is: https://github.com/eclipse/omr/pull/2918

PR #2750, opened 2018-08-30
Fix for possible crash in System/JNI call on X86-32

0dvictor:
Evaluate all children before switching stack to make a System/JNI call.


runtime/compiler/x/i386/codegen/IA32JNILinkage.cpp
```diff
@@ -491,7 +313,7 @@ TR::Register *TR::IA32JNILinkage::buildJNIDispatch(TR::Node *callNode)
       );
 
    if (isJNIGCPoint)
-      instr->setNeedsGCMap((argSize<<14) | getProperties().getPreservedRegisterMapForGC());
+      instr->setNeedsGCMap((argSize<<14) | 0xFF00FFFF);
```
cathyzhyi: could you explain more about this change? could we give the MASK here a meaningful name to indicate its purpose?

cathyzhyi: could you explain why is this change needed here? What's the meaning of the MASK? Is it possible to give it a meaningful name to indicate its purpose?

cathyzhyi:
The change fixes the problem of evaluating node during switching C stack and also change the `JNILinkage` to inherit from `SystemLinkage` rather than `PrivateLinkage` to make it more intuitive. LGTM.

fjeremic:
Jenkins test sanity xlinux,win JDK8

0dvictor:
@fjeremic It seems the sanity tests didn't start, do you mind try once more?

fjeremic:
Jenkins test sanity xlinux,win JDK8

0dvictor:
Sanity tests sill doesn't start. Do you mind double check?

fjeremic:
> Sanity tests sill doesn't start. Do you mind double check?

Yeah sanity testing is broken. Given the TravisCI build passed and this is primarily an x86 change I'll go ahead and merge.

Issue #2072, opened 2018-06-05
Opportunity to stack allocate more objects in OpenJ9


charliegracie:
Along with my investigation into Issue #2070 I also noticed some opportunities for stack allocation that I believe are being missed.

The tight loop in the [benchmark](https://www.excelsiorjet.com/memory-manager-scalability) is using boxed double objects to perform some math operations.
```
public void run() {
      Double result = 1d;
      Double bound = Main.until;
      for (Double f = 0d; f < bound; f++) {
        result += f / (result + 1);
      }
      if (log) System.out.println(result);
    }
```

In this loop only the final value for result actually escapes.  I can verify that even at a scorching compile that loop still allocates some of the intermediate Double objects per iteration.  I would have hoped that when running this test with `-verbose:gc` that at some point GCs would basically stop since no more objects were being created.

@vijaysun-omr this is the test I was discussing with you the other day.

charliegracie:
I attempted to dig through the JIT tracing to see why these objects are believed to be escaping but I very quickly got in too deep and was not able to follow the code without spending a significant amount of time.  I figured I would leave this to the experts.  I can gather logs, test potential fixes, etc. to help out in any way I can.

fjeremic:
> I attempted to dig through the JIT tracing to see why these objects are believed to be escaping but I very quickly got in too deep and was not able to follow the code without spending a significant amount of time.

Escape Analysis only runs for hot and above methods. What level did the method reach? I'm not too familiar with the optimization but I know it has it's limits as it does IL peeking at the moment to determine whether something escapes. A trace log with `traceEscapeAnalysis` will be able to tell.

andrewcraik:
In the current VM we assume that HCR is on by default which prohibits peeking to prove we objects don't escape when passed to a call. A run with -XX:-EnableHCR may help determine if the possibility of method redefinition is part of the escape issue.

Further, what is the declaration of log? If it is static final there is a bunch of work in progress to fold static final fields at compile-time with OSR support for decompiling the method if the field is written, but it is not enabled at present in general. You cold modify the case to simulate the effect of folding the constant to see how the performance changes.

I suspect between the disable of HCR and the enable of static field folding we would be able to get this to perform.

vijaysun-omr:
I had been looking at a test case like the one Charlie mentioned above already. I have some experimental escape analysis code that did the stack allocation in that test case and I will continue that analysis with what Charlie has sent above.

charliegracie:
[...]

PR #2752, opened 2018-08-30
Look for additional opportunities for stack allocation

hzongaro:
Look for additional opportunities for stack allocation
    
Check for cases in which locally allocated objects are only copied to other local variables, without values ever escaping.
    
These escape analysis enhancements were prototyped by Vijay Sundarsen @vijaysun-omr and Mark Stoodley @mstoodle .
    
Fixes:  #2072
    
### DIFF
--- a/runtime/compiler/optimizer/EscapeAnalysis.cpp
+++ b/runtime/compiler/optimizer/EscapeAnalysis.cpp
@@ -142,21 +142,50 @@ TR_EscapeAnalysis::TR_EscapeAnalysis(TR::OptimizationManager *manager)
 
 char *TR_EscapeAnalysis::getClassName(TR::Node *classNode)
    {
-   int32_t  classNameLength;
-   char    *classNameChars = TR::Compiler->cls.classNameChars(comp(), classNode->getSymbolReference(), classNameLength);
-   char   *className       = (char *)trMemory()->allocateStackMemory(classNameLength+1, TR_Memory::EscapeAnalysis);
-   memcpy(className, classNameChars, classNameLength);
-   className[classNameLength] = 0;
+   char *className = NULL;
+
+   if (classNode->getOpCodeValue() == TR::loadaddr)
+      {
+      TR::SymbolReference *symRef = classNode->getSymbolReference();
+
+      if (symRef->getSymbol()->isClassObject())
+         {
+         int32_t  classNameLength;
+         char    *classNameChars = TR::Compiler->cls.classNameChars(comp(), symRef,  classNameLength);
+
+         if (NULL != classNameChars)
+            {
+            className = (char *)trMemory()->allocateStackMemory(classNameLength+1, TR_Memory::EscapeAnalysis);
+            memcpy(className, classNameChars, classNameLength);
+            className[classNameLength] = 0;
+            }
+         }
+      }
    return className;
    }
 
 bool TR_EscapeAnalysis::isImmutableObject(TR::Node *node)
    {
    if (node->getOpCodeValue() != TR::New)
+      {
       return false;
+      }
 
-   if (!strncmp("java/lang/Integer", getClassName(node->getFirstChild()), 17))
+   char *className = getClassName(node->getFirstChild());
+
+   if (NULL != className &&
+          !strncmp("java/lang/", className, 10) &&
+             (!strcmp("Integer", &className[10]) ||
+              !strcmp("Long", &className[10]) ||
+              !strcmp("Short", &className[10]) ||
+              !strcmp("Byte", &className[10]) ||
+              !strcmp("Boolean", &className[10]) ||
+              !strcmp("Character", &className[10]) ||
+              !strcmp("Double", &className[10]) ||
+              !strcmp("Float", &className[10])))
+      {
       return true;
+      }
 
 
    return false;
@@ -210,6 +239,7 @@ int32_t TR_EscapeAnalysis::perform()
       }
 
    TR_ASSERT_FATAL(_maxSniffDepth < 16, "The argToCall and nonThisArgToCall flags are 16 bits - a depth limit greater than 16 will not fit in these flags");
+
    if (getLastRun())
       _maxPassNumber = 0; // Notwithstanding our heursitics, if this is the last run, our max "pass number" is zero (which is the first pass)
 
@@ -506,6 +536,8 @@ int32_t TR_EscapeAnalysis::performAnalysisOnce()
    _parms = NULL;
    _localObjectsValueNumbers = NULL;
    _visitedNodes = NULL;
+   _aliasesOfAllocNode = NULL;
+   _aliasesOfOtherAllocNode = NULL;
    _notOptimizableLocalObjectsValueNumbers = NULL;
    _notOptimizableLocalStringObjectsValueNumbers = NULL;
 
@@ -521,6 +553,8 @@ int32_t TR_EscapeAnalysis::performAnalysisOnce()
       _useDefInfo = optimizer()->getUseDefInfo();
       _blocksWithFlushOnEntry = new (trStackMemory()) TR_BitVector(comp()->getFlowGraph()->getNextNodeNumber(), trMemory(), stackAlloc);
       _visitedNodes = new (trStackMemory()) TR_BitVector(comp()->getNodeCount(), trMemory(), stackAlloc, growable);
+      _aliasesOfAllocNode = new (trStackMemory()) TR_BitVector(comp()->getSymRefCount(), trMemory(), stackAlloc, growable);
+      _aliasesOfOtherAllocNode = new (trStackMemory()) TR_BitVector(comp()->getSymRefCount(), trMemory(), stackAlloc, growable);
 
       if (!_useDefInfo)
          {
@@ -1382,13 +1416,17 @@ void TR_EscapeAnalysis::findCandidates()
          {
          if (node->getOpCodeValue() == TR::New)
             {
-            traceMsg(comp(), "Found [%p] new %s\n", node, getClassName(node->getFirstChild()));
+            const char *className = getClassName(node->getFirstChild());
+            traceMsg(comp(), "Found [%p] new %s\n", node,
+                     className ? className : "<Missing class name>");
             }
          else if (node->getOpCodeValue() == TR::newarray)
             traceMsg(comp(), "Found [%p] newarray of type %d\n", node, node->getSecondChild()->getInt());
          else
             {
-            traceMsg(comp(), "Found [%p] anewarray %s\n", node, getClassName(node->getSecondChild()));
+            const char *className = getClassName(node->getSecondChild());
+            traceMsg(comp(), "Found [%p] anewarray %s\n", node,
+                     className ? className : "<Missing class name>");
             }
          }
 
@@ -1489,7 +1527,10 @@ Candidate *TR_EscapeAnalysis::createCandidateIfValid(TR::Node *node, TR_OpaqueCl
             {
             if (trace())
                {
-               printf("secs Class %s implements Runnable in %s\n", getClassName(classNode), comp()->signature());
+               const char *className = getClassName(classNode->getSecondChild());
+               printf("secs Class %s implements Runnable in %s\n",
+                  className ? className : "<Missing class name>",
+                  comp()->signature());
                traceMsg(comp(), "   Node [%p] failed: class implements the Runnable interface\n", node);
                }
             return NULL;
@@ -2104,17 +2145,17 @@ bool TR_EscapeAnalysis::checkDefsAndUses(TR::Node *node, Candidate *candidate)
                               traceMsg(comp(), "   Look at other defs for use node %p of candidate %p\n", useNode, candidate->_node);
                            ////_otherDefsForLoopAllocation->set(udIndex);
 
-                           if (!checkOtherDefsOfLoopAllocation(useNode, candidate, (next->getFirstChild() == candidate->_node)))
-                              {
-                              if (trace())
-                                 traceMsg(comp(), "   Make [%p] non-local because multiple defs to node [%p]\n", candidate->_node, useNode);
-                              returnValue = false;
-                              }
                            if (!checkOverlappingLoopAllocation(useNode, candidate))
                               {
                               if (trace())
                                  traceMsg(comp(), "   Make [%p] non-local because it overlaps with use [%p]\n", candidate->_node, useNode);
                               /////printf("secs Overlapping loop allocation in %s\n", comp()->signature());
+                              returnValue = false;
+                              }
+                           if (!checkOtherDefsOfLoopAllocation(useNode, candidate, (next->getFirstChild() == candidate->_node)))
+                              {
+                              if (trace())
+                                 traceMsg(comp(), "   Make [%p] non-local because multiple defs to node [%p]\n", candidate->_node, useNode);
                               returnValue = false;
                               }
                            }
@@ -2139,6 +2180,18 @@ bool TR_EscapeAnalysis::checkOtherDefsOfLoopAllocation(TR::Node *useNode, Candid
    // them lead back to the allocation. If they do, it means that generations
    // of the allocation from different loop iterations may be alive at the same
    // time, so the allocation must be done from the heap and not the stack.
+   //
+   // In some limited cases, we can be sure that an object from a prior loop iteration
+   // was not live at the same time as an object from the next loop iteration without expensive analysis.
+   // One such "special" case is when all defs for uses reached by our candidate for stack allocation
+   // were fed by allocations; in this case it's easy to see that it was not an object from a prior iteration
+   // since it is a fresh allocation being done at that program point.
+   //
+   // There is one other special case dealt with in the code below related to a java/lang/Integer cache
+   // where again it's trivial to prove that the value cannot be a candidate allocation from a prior loop iteration
+   //
+   // There may be other such examples that can be added in the future, e.g. if the value is an already stack allocated
+   // object from a prior pass of escape analysis, it obviously cannot be a candidate for stack allocation in this pass.
    //
    int32_t useIndex = useNode->getUseDefIndex();
    if (useIndex <= 0)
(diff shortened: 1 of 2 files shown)


runtime/compiler/optimizer/EscapeAnalysis.cpp
```diff
@@ -155,7 +155,8 @@ bool TR_EscapeAnalysis::isImmutableObject(TR::Node *node)
    if (node->getOpCodeValue() != TR::New)
       return false;
 
-   if (!strncmp("java/lang/Integer", getClassName(node->getFirstChild()), 17))
+   if (!strncmp("java/lang/Integer", getClassName(node->getFirstChild()), 17) ||
```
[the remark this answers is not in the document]
charliegracie: Also what about Float, Long, etc.... Seems like future proofing now to handle all of the primitive wrapper classes might be a good idea.

runtime/compiler/optimizer/EscapeAnalysis.hpp
```diff
@@ -568,6 +576,8 @@ class TR_EscapeAnalysis : public TR::Optimization
    TR_BitVector              *_notOptimizableLocalStringObjectsValueNumbers;
    TR_BitVector              *_blocksWithFlushOnEntry;
    TR_BitVector              *_visitedNodes;
+   TR_BitVector              *_aliasesOfAllocNode;
+   TR_BitVector              *_aliasesOfAllocNode2;
```
fjeremic: This does not seem like a good field name. Could we name it better? Can we add documentation on what this field does exactly? Similarly for others, i.e. `candidate2`

0xdaryl:
This PR will ultimately have a more descriptive commit message, won't it?

runtime/compiler/optimizer/EscapeAnalysis.hpp
hzongaro: @fjeremic Filip, sorry for the long delay in responding.  How about instead of `candidate2` and `_aliasesOfAllocNode2`, we used names like `otherAllocNode` and `_aliasesOfOtherAllocNode`, along with a repetition of Vijay's inline comment describing `_aliasesofAllocNode`?  Something like this:

```
/** 
 * Contains sym refs that are just aliases for a fresh allocation
 * i.e., it is used to track allocations in cases such as
 * ...
 * a = new A() 
 * ...
 * b = a 
 * ...
 * c = b 
 *    
 * In this case a, b and c will all be considered aliases of an alloc node and
 * so a load of any of those sym refs will be treated akin to how the fresh
 * allocation would be treated
 */ 
   TR_BitVector              *_aliasesOfAllocNode;
/**
 * Contains sym refs that are just aliases for a second fresh allocation
 * that is under consideration, as with @ref _aliasesOfAllocNode
 */
   TR_BitVector              *_aliasesOfOtherAllocNode;
```

I know the name _aliasesOfOtherAllocNode doesn't really convey much more about its purpose than _aliasesOfAllocNode2 did, but they're really two vectors tracking aliases of particular candidate allocations that are under consideration at any one point in time.  Do you have any other suggestions?

Perhaps more importantly, we could add a comment over the loop that iterates over `cursor2` - something like this:

```
      // Loop over definitions for this use and over all the candidate allocations
      // If the definition comes directly from a candidate for stack allocation, it's harmless;
      // if it's copied from a variable that's aliased with a candidate for stack allocation that
      // was allocated in the same block, it's harmless
      for (cursor2.SetToFirstOne(); cursor2.Valid(); cursor2.SetToNextOne())
               {
               int32_t defIndex2 = cursor2;
               if (defIndex2 == 0)
                  {
                  allnewsonrhs = false;
                  break;
                  }

               TR::Node *defNode2 = _useDefInfo->getNode(defIndex2);
               TR::Node *firstChild = defNode2->getFirstChild();
               bool rhsIsHarmless = false;
               for (Candidate *candidate2 = _candidates.getFirst(); candidate2; candidate2 = candidate2->getNext())
                  {
                  if (candidate2->_node == firstChild)
                     {
                     rhsIsHarmless = true;
                     break;
```

Does that help?

@vijaysun-omr Vijay, does that capture things accurately?  Any other thoughts?

hzongaro: Via private communication, Vijay agreed with those suggested adjustments to variable names and comments.

hzongaro:
> This PR will ultimately have a more descriptive commit message, won't it?

Vijay @vijaysun-omr , Daryl @0xdaryl, I've finally come back to this PR, squashed the various commits and tried to compose a better description of the changes.  Does that describe things adequately?  Once you're OK with the commit messages, I'll remove the _WIP_ tag.

runtime/compiler/optimizer/EscapeAnalysis.cpp
```diff
@@ -155,7 +155,17 @@ bool TR_EscapeAnalysis::isImmutableObject(TR::Node *node)
    if (node->getOpCodeValue() != TR::New)
       return false;
 
-   if (!strncmp("java/lang/Integer", getClassName(node->getFirstChild()), 17))
+   char *className = getClassName(node->getFirstChild());
+
+   if (!strncmp("java/lang/", className, 10) &&
+          (!strcmp("Integer", &className[10]) ||
+           !strcmp("Long", &className[10]) ||
+           !strcmp("Short", &className[10]) ||
```
fjeremic: Is there really not a better way? Not saying we should fix this here but this looks like a really bad way to looking for particular classes. I know for recognized methods we have things like `TR::java_util_HashMap_all` to capture all HashMap methods. Wondering if something similar exists for classes.

hzongaro: @fjeremic Filip, I'm trying to figure out whether there's a cleaner way of doing this check.

hzongaro: @fjeremic Filip, I don't see any existing mechanism for doing what you're suggesting.  Would you like me to open an issue to track that as an improvement?

fjeremic: Yes, let's open up an issue. I see similar types of `strcmp` code in other parts of the code base. I do think we need some notion on "recognized classes" to avoid this type of cruft. Thanks!

hzongaro: Opened issue #3804

runtime/compiler/optimizer/EscapeAnalysis.cpp
```diff
@@ -2332,6 +2432,32 @@ bool TR_EscapeAnalysis::checkOverlappingLoopAllocation(TR::Node *node, TR::Node
    }
 
 
+
+void TR_EscapeAnalysis::collectAliasesOfAllocations(TR::Node *node, TR::Node *allocNode, bool collectAliases)
```
andrewcraik: I am not a fan of this bool parameter that makes this method just do the visit - basically putting all of the logic in an if... I would rather we had a method to just visit and a method to collect aliases and dropped this boo parameter which is going to make usage harder to understand/follow.

runtime/compiler/optimizer/EscapeAnalysis.cpp
```diff
@@ -2332,6 +2432,32 @@ bool TR_EscapeAnalysis::checkOverlappingLoopAllocation(TR::Node *node, TR::Node
...
+   {
+   if (_visitedNodes->get(node->getGlobalIndex()))
+      return;
+
+   _visitedNodes->set(node->getGlobalIndex());
+
+   if (collectAliases && node->getOpCode().isStore() && node->getSymbol()->isAutoOrParm())
+      {
+      if (node->getFirstChild() == allocNode)
```
andrewcraik: code style - this if ladder should have {}s IMO just due to complexity. Given teh complexity of the conditions here can we put them one per line with the && at the star to the line followed by teh condition? This will reduce the line width and make this much more clear and easier to reason about. This is a style used elsewhere in the compiler when dealing with such complex conditions and has tended to work well.

runtime/compiler/optimizer/EscapeAnalysis.cpp
```diff
@@ -2314,7 +2400,21 @@ bool TR_EscapeAnalysis::checkOverlappingLoopAllocation(TR::Node *node, TR::Node
...
+         _aliasesOfAllocNode->set(node->getSymbolReference()->getReferenceNumber());
+      else if (!_visitedNodes->get(node->getFirstChild()->getGlobalIndex()) && node->getFirstChild()->getOpCode().isLoadVarDirect() && node->getFirstChild()->getSymbol()->isAutoOrParm() &&
+               _aliasesOfAllocNode->get(node->getFirstChild()->getSymbolReference()->getReferenceNumber()))
+         _aliasesOfAllocNode->set(node->getSymbolReference()->getReferenceNumber());
+      else
+         _aliasesOfAllocNode->reset(node->getSymbolReference()->getReferenceNumber());
+      }
+
+   if ((_valueNumberInfo->getValueNumber(node) == _valueNumberInfo->getValueNumber(useNode)) &&
```
andrewcraik: for efficiency perhaps do the node != allocNode first since it is cheaper?

runtime/compiler/optimizer/EscapeAnalysis.cpp
```diff
@@ -2202,17 +2252,49 @@ bool TR_EscapeAnalysis::checkOtherDefsOfLoopAllocation(TR::Node *useNode, Candid
...
+                       if (node->getOpCodeValue() == TR::BBEnd)
+                          break;
+
+                       collectAliasesOfAllocations(treeTop->getNode(), otherAllocNode->_node, trackAliases);
+
+                       if (treeTop == otherAllocNode->_treeTop)
+                           trackAliases = true;
+                       }
+
+                    if (/* (_valueNumberInfo->getValueNumber(defNode2) == _valueNumberInfo->getValueNumber(otherAllocNode->_node)) && */
```
andrewcraik: please remove dead code rather than commenting it out

runtime/compiler/optimizer/EscapeAnalysis.cpp
```diff
@@ -2178,18 +2205,41 @@ bool TR_EscapeAnalysis::checkOtherDefsOfLoopAllocation(TR::Node *useNode, Candid
...
+      // ...
+      // c = b
+      //
+      // In this case a, b and c will all be considered aliases of an alloc node and so a load of
+      // any of those sym refs will be treated akin to how the fresh allocation would have been in the below logic
+      //
+
       bool allnewsonrhs = false;
-      if ((_valueNumberInfo->getValueNumber(defNode) == _valueNumberInfo->getValueNumber(candidate->_node)))
+      //if ((_valueNumberInfo->getValueNumber(defNode) == _valueNumberInfo->getValueNumber(candidate->_node)))
```
andrewcraik: again remove dead code. Once we agree on the change a change to remove the {}s and reduce the indentation is probably also needed.

andrewcraik:
Overall, I think this set of changes needs much better trace. There is nothing printed to the trace logs about what nodes are considered as equivalent in the new scheme or what the analysis is doing. When trace is enabled we need to be able to follow the analysis from the trace for debug purposes since we may only ever get the log as the means of finding and fixing a problem.

Additionally, is there any reasonable way to have an env var or other mechanism to disable this new opt to aid debugging? If we find failures after this is delivered it would be nice to have an easy way to rule this change in or out.

runtime/compiler/optimizer/EscapeAnalysis.cpp
```diff
@@ -155,7 +155,17 @@ bool TR_EscapeAnalysis::isImmutableObject(TR::Node *node)
    if (node->getOpCodeValue() != TR::New)
       return false;
 
-   if (!strncmp("java/lang/Integer", getClassName(node->getFirstChild()), 17))
+   char *className = getClassName(node->getFirstChild());
```
andrewcraik: getClassName may not be new, but I think there needs to be a safety check that the node is one which should have a class as its symbol and that we can get the name. We should handle this returning NULL in cases where it can't be gotten.

hzongaro: Thanks, Andrew @andrewcraik.  I've been pawing around trying to figure out the right way to perform the check you're recommending.  Would this be the right way?

```
if (classNode->getOpCode().hasSymbolReference())
   {
   TR::SymbolReference *symRef = classNode->getSymbolReference();
   TR::Symbol *sym = symRef->getSymbol();
   if (sym->isClassObject() && !sym->addressIsCPIndexOfStatic())
      ...
   }
```

andrewcraik: So the child node will need to be a loadaddr otherwise the symref won't be describing the thing we want (currently news have a loadaddr as a child, but this will be changing). So the first part of the check needs to be that we have a loadaddr.

Usage elsewhere in the compiler suggests sym->isClassObject() is the check you need. I don't know that you need the addressIsCPIndexOfStatic.

runtime/compiler/optimizer/EscapeAnalysis.cpp
hzongaro: Cleaned this up in commit 414b74b10957a8ae1b8a4426f0c29dec2a905275

runtime/compiler/optimizer/EscapeAnalysis.cpp
hzongaro: Cleaned this up in commit 414b74b10957a8ae1b8a4426f0c29dec2a905275

runtime/compiler/optimizer/EscapeAnalysis.cpp
hzongaro: Removed dead code in 414b74b10957a8ae1b8a4426f0c29dec2a905275

runtime/compiler/optimizer/EscapeAnalysis.cpp
hzongaro: Removed dead code and braces, and adjusted indentation in commit 414b74b10957a8ae1b8a4426f0c29dec2a905275

runtime/compiler/optimizer/EscapeAnalysis.cpp
hzongaro: Added checking to `getClassName` and its callers in commit f12434d607be9a31b4674ea1c8bcedceb82558da

runtime/compiler/optimizer/EscapeAnalysis.cpp
hzongaro: Andrew @andrewcraik, Just to be sure I'm clear on what you're suggesting, do you want to see a `visitTree` method that just visits nodes in a tree and a `collectAliasesOfAllocations` that both visits and collects aliases unconditionally?  Or do you want to see a `visitTree` method that visits nodes and also takes a function pointer (or an object) that it calls at each node, where that function would either be no-op or collect aliases?

andrewcraik: for this I'd suggest the former - a `visitTree` that unconditionally visits the tree and marks it as visited and a `collectAliasesofAllocations` that does what it currently does when the Boolean is true. Function pointers can make debugging the compiler harder so I try to avoid them.

hzongaro: Split `collectAliasesOfAllocations` into two methods in commit 665010adf18731078ee8181b05cf4ae0480c1a69, as Andrew suggested

hzongaro:
> Overall, I think this set of changes needs much better trace. There is nothing printed to the trace logs about what nodes are considered as equivalent in the new scheme or what the analysis is doing. When trace is enabled we need to be able to follow the analysis from the trace for debug purposes since we may only ever get the log as the means of finding and fixing a problem.

Andrew @andrewcraik, I have a commit pending that adds additional tracing, though I'll need someone to review for use of terminology, etc.

> Additionally, is there any reasonable way to have an env var or other mechanism to disable this new opt to aid debugging? If we find failures after this is delivered it would be nice to have an easy way to rule this change in or out.

I'm working on using a `TR_disableEALoopAliasAllocationChecking` environment variable for disabling the additional alias checking.  I'll run complete testing with it enabled and disabled.

I was a bit torn on whether to weave checking of the associated bool, `_doLoopAliasAllocationChecking`, throughout the `if` that calculates `allnewsonrhs` with the old and new logic intermingled, but ended up deciding to simply pull an old version of that logic and a new version into new methods for easier maintenance.  Commit pending.

runtime/compiler/optimizer/EscapeAnalysis.cpp
```diff
@@ -1489,7 +1527,10 @@ Candidate *TR_EscapeAnalysis::createCandidateIfValid(TR::Node *node, TR_OpaqueCl
             {
             if (trace())
                {
-               printf("secs Class %s implements Runnable in %s\n", getClassName(classNode), comp()->signature());
+               const char *className = getClassName(classNode->getSecondChild());
+               printf("secs Class %s implements Runnable in %s\n",
```
andrewcraik: please convert to guarded trace

runtime/compiler/optimizer/EscapeAnalysis.cpp
```diff
@@ -1382,13 +1416,17 @@ void TR_EscapeAnalysis::findCandidates()
          {
          if (node->getOpCodeValue() == TR::New)
             {
-            traceMsg(comp(), "Found [%p] new %s\n", node, getClassName(node->getFirstChild()));
+            const char *className = getClassName(node->getFirstChild());
+            traceMsg(comp(), "Found [%p] new %s\n", node,
```
andrewcraik: probably deserves an if(trace())

runtime/compiler/optimizer/EscapeAnalysis.cpp
```diff
@@ -1382,13 +1416,17 @@ void TR_EscapeAnalysis::findCandidates()
...
+                     className ? className : "<Missing class name>");
             }
          else if (node->getOpCodeValue() == TR::newarray)
             traceMsg(comp(), "Found [%p] newarray of type %d\n", node, node->getSecondChild()->getInt());
          else
             {
-            traceMsg(comp(), "Found [%p] anewarray %s\n", node, getClassName(node->getSecondChild()));
+            const char *className = getClassName(node->getSecondChild());
+            traceMsg(comp(), "Found [%p] anewarray %s\n", node,
```
andrewcraik: similar trace guard

runtime/compiler/optimizer/EscapeAnalysis.cpp
```diff
@@ -2178,69 +2231,134 @@ bool TR_EscapeAnalysis::checkOtherDefsOfLoopAllocation(TR::Node *useNode, Candid
...
                {
-               int32_t defIndex2 = cursor2;
-               if (defIndex2 == 0)
+               allnewsonrhs = false;
+               break;
+               }
+
+            TR::Node *defNode2 = _useDefInfo->getNode(defIndex2);
+            TR::Node *firstChild = defNode2->getFirstChild();
+            bool rhsIsHarmless = false;
```
andrewcraik: could we have a comment on what harmless means in this context to provide the intuition for all the checks that follow?

runtime/compiler/optimizer/EscapeAnalysis.cpp
```diff
@@ -2178,69 +2231,134 @@ bool TR_EscapeAnalysis::checkOtherDefsOfLoopAllocation(TR::Node *useNode, Candid
...
+                     int32_t fieldNameLen = -1;
+                     char *fieldName = NULL;
+                     if (underlyingArray && underlyingArray->getOpCode().hasSymbolReference() &&
+                         underlyingArray->getSymbolReference()->getSymbol()->isStaticField())
+                        {
+                        fieldName = underlyingArray->getSymbolReference()->getOwningMethod(comp())->staticName(underlyingArray->getSymbolReference()->getCPIndex(), fieldNameLen, comp()->trMemory());
                         }
+
+                     if (fieldName && (fieldNameLen > 0) &&
+                         !strncmp(fieldName, "java/lang/Integer$IntegerCache.cache", 36))
```
andrewcraik: does long have an equivalent cache that should be treated the same way for symmetry?

runtime/compiler/optimizer/EscapeAnalysis.cpp
hzongaro: @andrewcraik Andrew, are you asking me to replace the `printf` with `traceMsg`?  If so, there are a few other places with calls to `printf` that I'll take care of as well.

runtime/compiler/optimizer/EscapeAnalysis.cpp
hzongaro: It seems it does, as do `java.lang.Byte` and `java.lang.Character`

runtime/compiler/optimizer/EscapeAnalysis.cpp
andrewcraik: yes please - printf is not a nice way to debug anything in the JIT so we should only commit code that logs using the traceMsg mechanism.

runtime/compiler/optimizer/EscapeAnalysis.cpp
hzongaro: That one is actually in an `if (trace())`:
```
      if (trace())
         {
         if (node->getOpCodeValue() == TR::New)
            {
            const char *className = getClassName(node->getFirstChild());
            traceMsg(comp(), "Found [%p] new %s\n", node,
                     className ? className : "<Missing class name>");
            }
         else if (node->getOpCodeValue() == TR::newarray)
```

runtime/compiler/optimizer/EscapeAnalysis.cpp
hzongaro: Expanded the set of caches checked for in commit 92b0cb068aa2ec511d8df1d4cc8a071eac10ca4d

runtime/compiler/optimizer/EscapeAnalysis.cpp
hzongaro: Addressed this in commit aa48bf3fd7a279f3feb25aba345b7ff6850f3c8f

hzongaro:
Andrew @andrewcraik, I think I've addressed all your comments.  I would appreciate it if you could have a look - particularly at those that address your [review comment on `rhsIsHarmless`](https://github.com/eclipse/openj9/pull/2752/#pullrequestreview-182719656) and your [request for additional tracing and a way of disabling the new code](https://github.com/eclipse/openj9/pull/2752/#issuecomment-444921956).

Once you're OK with these changes and any follow on changes, I can collapse these all down to one (or two) commits.

runtime/compiler/optimizer/EscapeAnalysis.cpp
```diff
@@ -2432,8 +2432,14 @@ bool TR_EscapeAnalysis::checkAllNewsOnRHSInLoopWithAliasing(int32_t defIndex, TR
                      fieldName = underlyingArray->getSymbolReference()->getOwningMethod(comp())->staticName(underlyingArray->getSymbolReference()->getCPIndex(), fieldNameLen, comp()->trMemory());
                      }
 
-                  if (fieldName && (fieldNameLen > 0) &&
-                      !strncmp(fieldName, "java/lang/Integer$IntegerCache.cache", 36))
+                  if (fieldName && (fieldNameLen > 10) &&
+                        !strncmp("java/lang/", fieldName, 10) &&
+                          (!strcmp("Integer$IntegerCache.cache", &fieldName[10]) ||
```
andrewcraik: I think we need to use `strncmp` here because names are not guaranteed to be null terminated.

andrewcraik: Otherwise - I think this looks good.

hzongaro: OK.  Now I see `staticName` yields the name and type signature, separated by a space.  I'll fix that.

hzongaro: Corrected that blunder in commit 4bf7ad3bb06a85e54e7d8562d875ffc474c2f8fd

hzongaro:
> I think this might be more clear as disableLoopAliasAllocationChecking == NULL

Fixed in commit 4bf7ad3bb06a85e54e7d8562d875ffc474c2f8fd
@andrewcraik

runtime/compiler/optimizer/EscapeAnalysis.cpp
```diff
@@ -521,6 +556,12 @@ int32_t TR_EscapeAnalysis::performAnalysisOnce()
       _useDefInfo = optimizer()->getUseDefInfo();
       _blocksWithFlushOnEntry = new (trStackMemory()) TR_BitVector(comp()->getFlowGraph()->getNextNodeNumber(), trMemory(), stackAlloc);
       _visitedNodes = new (trStackMemory()) TR_BitVector(comp()->getNodeCount(), trMemory(), stackAlloc, growable);
+      _aliasesOfAllocNode =
+          _doLoopAllocationAliasChecking
+                 ? new (trStackMemory()) TR_BitVector(comp()->getSymRefCount(), trMemory(), stackAlloc, growable) : NULL;
```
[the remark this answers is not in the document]
hzongaro: Adjusted in commit 4bf7ad3bb06a85e54e7d8562d875ffc474c2f8fd

andrewcraik:
Alrighty I'm good with this. @hzongaro can you squash to the final form of commits you expect and @vijaysun-omr to see if he has any concerns? we can then sanity ready for merge.

hzongaro:
Squashed changes down to three commits:

- 4ee42c01289eca51556948f197b4985d5145b596
- 7d903f9b4c8c070809f0764f3b276876d63904a4
- 7d903f9b4c8c070809f0764f3b276876d63904a4

charliegracie:
@andrewcraik and @vijaysun-omr is this ready to go now?

vijaysun-omr:
I'm fine with this now. I believe @hzongaro was planning to check with Andrew before this goes in.

andrewcraik:

andrewcraik:

PR #2753, opened 2018-08-30
Improve interface method resolution

gacholio:
Remove the need for memory barriers in interface method resolution
checks by adding a special initial value for the methodIndex. The new
resolution check is that the interfaceClass is not NULL and the
methodIndex is not the special value. The special value represents a bit
pattern which can not occur in a resolved entry.

### DIFF
--- a/runtime/codert_vm/cnathelp.cpp
+++ b/runtime/codert_vm/cnathelp.cpp
@@ -1541,10 +1541,9 @@ retry:
 	J9ConstantPool *ramConstantPool = ((J9ConstantPool**)indexAndLiteralsEA)[0];
 	UDATA cpIndex = indexAndLiteralsEA[1];
 	J9RAMInterfaceMethodRef *ramMethodRef = (J9RAMInterfaceMethodRef*)ramConstantPool + cpIndex;
-	J9Class* volatile interfaceClass = (J9Class*)ramMethodRef->interfaceClass;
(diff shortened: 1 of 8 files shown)


DanHeidinga:
Changes look good.  I think there's copy of the itable to vtable mapping code in the JIT for JSR 292's InterfaceHandle that needs updating as well:
https://github.com/eclipse/openj9/blob/a6a141215c96220f65360b62edfe2809de82f4e7/runtime/compiler/env/VMJ9.cpp#L9119-L9143

DanHeidinga:

DanHeidinga:

gacholio:
@DanHeidinga I don't think the JIT code needs to be changed. If we ecounter one of the special cases, MethodHandles adapts to the correct kind of invoke, so the JIT code will only apply for the simple interface case.

Look in adaptInterfaceLookupsOfObjectMethodsIfRequired - are we missing the logic to handle both virtual and final Object methods? See findVirtual, which makes the distinction.

DanHeidinga:
We may not have to update the code but I'd rather keep it consistent with the other implementations so that anyone of them is a valid starting point - better still would be to have a single copy of the code.

gacholio:
It's not possible to handle all the cases at this point in the code - that's why Devin added the new imrproper method call instead of changing the existing compile-time resolve helpers.

PR #2754, opened 2018-08-30
JIT common and x86 support for constant dynamic

yanluo7:

### DIFF
--- a/runtime/compiler/env/j9method.cpp
+++ b/runtime/compiler/env/j9method.cpp
@@ -1561,6 +1561,7 @@ cpType2trType(UDATA cpType)
       case J9CPTYPE_ANNOTATION_UTF8:
       case J9CPTYPE_METHOD_TYPE:
       case J9CPTYPE_METHODHANDLE:
+      case J9CPTYPE_CONSTANT_DYNAMIC:
          return TR::Address;
       case J9CPTYPE_INT:
          return TR::Int32;
@@ -5837,6 +5838,54 @@ uint64_t
 TR_ResolvedJ9Method::longConstant(I_32 cpIndex)
    {
    return *((uint64_t *) & romLiterals()[cpIndex]);
+   }
+
+void *
+TR_ResolvedJ9Method::getConstantDynamicTypeFromCP(I_32 cpIndex)
+   {
+   return jitGetConstantDynamicTypeFromCP(fej9()->vmThread(), cp(), cpIndex);
+   }
+
+bool
+TR_ResolvedJ9Method::isConstantDynamic(I_32 cpIndex)
+   {
+   TR_ASSERT_FATAL(cpIndex != -1, "ConstantDynamic cpIndex shouldn't be -1");
+   UDATA cpType = J9_CP_TYPE(J9ROMCLASS_CPSHAPEDESCRIPTION(cp()->ramClass->romClass), cpIndex);
+   return (J9CPTYPE_CONSTANT_DYNAMIC == cpType);
+   }
+
+// Both value and exception slots are of object pointer type.
+// If first slot is non null, the CP entry is resolved to a non-null value.
+// Else if second slot is the class object of j/l/Void, the CP entry is resolved to null (0) value.
+// We retrieve the Void class object via javaVM->voidReflectClass->classObject,
+// which is protected by VMAccessCrtitical section to ensure vm access.
+// Other casese, the CP entry is considered unresolved.
+bool
+TR_ResolvedJ9Method::isUnresolvedConstantDynamic(I_32 cpIndex)
+   {
+   TR_ASSERT(cpIndex != -1, "ConstantDynamic cpIndex shouldn't be -1");
+   if (((J9RAMConstantDynamicRef *) literals())[cpIndex].value != 0)
+      {
+      return false;
+      }
+   if (((J9RAMConstantDynamicRef *) literals())[cpIndex].exception == 0)
+      {
+      return true;
+      }
+
+   TR::VMAccessCriticalSection voidClassObjectCritSec(fej9()); 
+   J9JavaVM * javaVM = fej9()->_jitConfig->javaVM;
+   j9object_t voidClassObject = javaVM->voidReflectClass->classObject;
+   j9object_t slot2 = ((J9RAMConstantDynamicRef *) literals())[cpIndex].exception;
+
+   return (voidClassObject != slot2);
+   }
+
+void *
+TR_ResolvedJ9Method::dynamicConstant(I_32 cpIndex)
+   {
+   TR_ASSERT_FATAL(cpIndex != -1, "ConstantDynamic cpIndex shouldn't be -1");
+   return &((J9RAMConstantDynamicRef *) literals())[cpIndex].value;
    }
 
 bool
--- a/runtime/compiler/ilgen/Walker.cpp
+++ b/runtime/compiler/ilgen/Walker.cpp
@@ -6372,7 +6372,205 @@ TR_J9ByteCodeIlGenerator::loadFromCP(TR::DataType type, int32_t cpIndex)
             loadSymbol(TR::dload, symRefTab()->findOrCreateDoubleSymbol(_methodSymbol, cpIndex));
          break;
       case TR::Address:
-         if (method()->isClassConstant(cpIndex))
+         if (method()->isConstantDynamic(cpIndex))
+            {
+            bool isCondyUnresolved = _methodSymbol->getResolvedMethod()->isUnresolvedConstantDynamic(cpIndex);
+            J9UTF8 *returnTypeUtf8 = (J9UTF8 *)_methodSymbol->getResolvedMethod()->getConstantDynamicTypeFromCP(cpIndex);
+            int returnTypeUtf8Length = J9UTF8_LENGTH(returnTypeUtf8);
+            char* returnTypeUtf8Data = (char *)J9UTF8_DATA(returnTypeUtf8);
+            bool isCondyPrimitive = (1 == returnTypeUtf8Length);
+
+            char* symbolTypeSig = NULL;
+            int32_t symbolTypeSigLength = 0;
+            // For non-primitive condy, resolved or not, we create a static symbol,
+            // store the class info in the symbol so it may be retrieved by optimizer later via
+            // getTypeSignature.
+            if (!isCondyPrimitive)
+               {
+               symbolTypeSig = (char*)comp()->trMemory()->allocateMemory(returnTypeUtf8Length, heapAlloc);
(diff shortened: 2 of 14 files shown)


yanluo7:
Design documentation is detailed in https://github.com/eclipse/openj9/issues/2307

yanluo7:
@cathyzhyi Could you do a review?

yanluo7:
Travis CI build failure is expected due to OMR dependency change.

runtime/compiler/env/j9method.cpp
```diff
@@ -5839,6 +5840,44 @@ TR_ResolvedJ9Method::longConstant(I_32 cpIndex)
...
+   TR_ASSERT(cpIndex != -1, "cpIndex shouldn't be -1");
+   UDATA cpType = J9_CP_TYPE(J9ROMCLASS_CPSHAPEDESCRIPTION(cp()->ramClass->romClass), cpIndex);
+   return (J9CPTYPE_CONSTANT_DYNAMIC == cpType);
+   }
+
+// If first slot is non null, the CP entry is resolved to a non-null value.
+// Else if second slot is Void.class, the CP entry is resolved to null (0) value.
+// Other casese, the CP entry is considered unresolved.
+bool
+TR_ResolvedJ9Method::isUnresolvedConstantDynamic(I_32 cpIndex)
```
cathyzhyi: Does the first slot represent the value of the condy and the second slot represent the resolved class of the condy? Maybe it's better to say this explicitly in the comment to avoid confusion especially when the name of the second slot is called `exception`.

irinarada:
@yanluo7 , @cathyzhyi any ETA for the latest changes? This blocks the z delivery part. @fjeremic , @gita-omr @andrewcraik

runtime/compiler/env/j9method.cpp
yanluo7: Both value and exception slots are of object pointer type. If first slot is non null, the CP entry is resolved to a non-null value. Else if second slot is the class object of j/l/Void, the CP entry is resolved to null (0) value. We retrieve the Void class object via javaVM->voidReflectClass->classObject, which is protected by VMAccessCrtiticalSection to ensure vm access. Other cases, the CP entry is considered unresolved. Added VMAccessCrtiticalSection code and more comments.

andrewcraik:
@yanluo7 asked me to review this yesterday - I have just arrived back from vacation so I will try and do a review today. If the review goes well the merge should be very soon, but I would target end of the week.

runtime/compiler/env/j9method.cpp
```diff
@@ -5839,6 +5840,54 @@ TR_ResolvedJ9Method::longConstant(I_32 cpIndex)
...
+void *
+TR_ResolvedJ9Method::getConstantDynamicTypeFromCP(I_32 cpIndex)
+   {
+   return jitGetConstantDynamicTypeFromCP(fej9()->vmThread(), cp(), cpIndex);
+   }
+
+bool
+TR_ResolvedJ9Method::isConstantDynamic(I_32 cpIndex)
+   {
+   TR_ASSERT(cpIndex != -1, "cpIndex shouldn't be -1");
```
andrewcraik: This should probably be an TR_ASSERT_FATAL - it will never work with cpIndex -1, it doesn't handle the case, and it is just plain wrong to call it with -1. The check is cheap so keeping it in at runtime will make sure we don't miss a bad use of this API.

runtime/compiler/env/j9method.cpp
```diff
@@ -5839,6 +5840,54 @@ TR_ResolvedJ9Method::longConstant(I_32 cpIndex)
...
+   j9object_t voidClassObject = javaVM->voidReflectClass->classObject;
+   j9object_t slot2 = ((J9RAMConstantDynamicRef *) literals())[cpIndex].exception;
+
+   return (voidClassObject != slot2);
+   }
+
+void *
+TR_ResolvedJ9Method::dynamicConstant(I_32 cpIndex)
+   {
+   TR_ASSERT(cpIndex != -1, "cpIndex shouldn't be -1");
```
andrewcraik: same thing for this one

runtime/compiler/ilgen/Walker.cpp
```diff
@@ -6372,7 +6372,202 @@ TR_J9ByteCodeIlGenerator::loadFromCP(TR::DataType type, int32_t cpIndex)
...
+
+            // If condy is primitive type and resolved, load the primitive constant;
+            // Otherwise, load using a CP symol (for resolved and unresolved object type),
+            // and generate subsequent loadi for the unresolved primitive 'value' field if needed (because
+            // for unresolved primitive the resovle helper only returns an autobox'd object).
+            if (isCondyPrimitive)
+               {
+               char *autoboxClassSig = NULL;
+               int32_t autoboxClassSigLength = 0;
+               switch (returnTypeUtf8Data[0])
```
andrewcraik: There should be an assert above this to make sure the signature is valid - ie 1 char.

runtime/compiler/ilgen/Walker.cpp
```diff
@@ -6372,7 +6372,202 @@ TR_J9ByteCodeIlGenerator::loadFromCP(TR::DataType type, int32_t cpIndex)
...
+                                          + comp()->fej9()->getInstanceFieldOffset(typeClassBlock, "value", 5, "S", 1);
+                     break;
+                  case 'Z':
+                     autoboxClassSig = "Ljava/lang/Boolean;";
+                     autoboxClassSigLength = 19;
+                     typeClassBlock = comp()->fej9()->getClassFromSignature(autoboxClassSig, autoboxClassSigLength, method());
+                     valueOffset = comp()->fej9()->getObjectHeaderSizeInBytes()
+                                          + comp()->fej9()->getInstanceFieldOffset(typeClassBlock, "value", 5, "Z", 1);
+                     break;
+                  default:
```
andrewcraik: shouldn't this assert?

andrewcraik:
@yanluo7 the new methods need doxygen comments to describe their operation, their inputs and their return values as well as anything else people using the API need to know / be aware of

runtime/compiler/env/j9method.cpp
yanluo7: Yes, this method should have fatal assertion as this query is the entry point of JIT condy processing logic.

runtime/compiler/env/j9method.cpp
yanluo7: Added fatal assertion here.

runtime/compiler/ilgen/Walker.cpp
yanluo7: Here we are inside the `if (isCondyPrimitive)` condition so it's already implicitly ensured that `turnTypeUtf8Length` is 1 for the type signature UTF8 : `bool isCondyPrimitive = (1 == returnTypeUtf8Length);`

runtime/compiler/ilgen/Walker.cpp
yanluo7: Yes, added a fatal assertion for unrecognized type signature.

runtime/compiler/ilgen/Walker.cpp
```diff
@@ -6372,7 +6372,203 @@ TR_J9ByteCodeIlGenerator::loadFromCP(TR::DataType type, int32_t cpIndex)
...
+                           loadConstant(TR::lconst, *(int64_t*)(obj+valueOffset));
+                           break;
+                        case 'F':
+                           loadConstant(TR::fconst, *(float*)(obj+valueOffset));
+                           break;
+                        case 'D':
+                           loadConstant(TR::dconst, *(double*)(obj+valueOffset));
+                           break;
+                        default:
+                           break;
```
andrewcraik: need an assert here too to match the case above.

runtime/compiler/ilgen/Walker.cpp
```diff
@@ -6372,7 +6372,203 @@ TR_J9ByteCodeIlGenerator::loadFromCP(TR::DataType type, int32_t cpIndex)
...
+                     valueRecogField = TR::Symbol::Java_lang_Short_value;
+                     dt = TR::Int16;
+                     break;
+                  case 'Z':
+                     recogFieldName = "java/lang/Boolean.value Z";
+                     valueRecogField = TR::Symbol::Java_lang_Boolean_value;
+                     dt = TR::Int32;
+                     break;
+                  default:
+                     break;
```
andrewcraik: Also needs an assert

runtime/compiler/ilgen/Walker.cpp
yanluo7: Added assertion.

runtime/compiler/ilgen/Walker.cpp
yanluo7: Added assertion.

andrewcraik:
Jenkins test sanity all jdk11

andrewcraik:
Ok I think that looks good - will let sanity run and then I think this should be good to go.

fjeremic:
We're waiting for an OMR-Acceptance promotion due to a bad OMR merge causing mutual dependencies for Z codny support. Sorry about that. You can launch a Jenkinds depends build against eclipse/omr#master though. That should work fine.

andrewcraik:
Jenkins test sanity all jdk11 depends eclipse/omr#master

andrewcraik:
@fjeremic thanks we'll do that - my bad for not noticing since I've been busy with non-openj9 things

yanluo7:
Added missing `ExternHelper` declaration that should address windows compile.

andrewcraik:
Jenkins test sanity all jdk11 depends eclipse/omr#master

andrewcraik:
@yanluo7 can you take a look at sanity and see if the failures are of concern in your opinion - I will look at them separately and see if we agree

yanluo7:
@andrewcraik Summarized the test failures here:
win_x86-64_cmprssptrs & linux_x86-64_cmprssptrs : `Nestmate_virtual_private_0`
linux_x86-64: `jsr335_interfacePrivateMethod` and `Nestmate_virtual_private`
linux_390-64_cmprssptrs : `org.openj9.test.java.lang.Test_String.test_toLowerCase_lithuanian` and 
`java.lang.ClassFormatError: JVMCFRE145 Nest host must be a constant class;`  There seems to be a few NEST-related failures here.

The entry point to all the constant dynamic logic gated by  `TR_ResolvedJ9Method::isConstantDynamic(cpIndex)`, which checks whether the given CP type is `J9CPTYPE_CONSTANT_DYNAMIC ` , it should not kick in for all of these tests.

andrewcraik:
@yanluo7 I agree - there don't seem to be constant dynamic failures in the above so I will merge this.

PR #2758, opened 2018-08-31
Update error messages to refer to the right events

DanHeidinga:


gacholio:
Not wasting build resource on something that's clearly only a printf change..

PR #2759, opened 2018-08-31
Remove unused and incorrect prototype for flushICache()

shingarov:
Signed-off-by: Boris Shingarov <shingarov@labware.com>


DanHeidinga:

DanHeidinga:
The ip-validation check is failing but not providing results.  Likely eclipse infra related.  It may be necessary to repush the commit to retrigger the check.

DanHeidinga:
@shingarov Can you rebase the commit?  This will cause the ip-validation to rerun.  It's one of the few checks I can't restart directly.

shingarov:
Ok I don't know what gives.  The same check passes fine for OMR pushes, and afaiu from the ecr page, the ecr isn't project-specific although there is some magic there which I don't understand (why did my arm pr's not trigger this? does this have anything to do with committer status in another project? maybe it would make sense to quickly touch on these points in tomorrow's call?)

shingarov:
I am not sure why ip-validation doesn't pass in this PR.  I created https://github.com/eclipse/openj9/pull/3239 to replace this one.

PR #2763, opened 2018-09-04
Move findCorrectInlinedSiteIndex() from OMR and rework it to be usable for non-AOT code

ymanton:
This PR moves `findCorrectInlinedSiteIndex()` from OMR, where it has no uses and no good reason to be, into the J9 compiler and allows it to be called in cases where we are using relocations but not compiling under AOT mode and therefore don't have `compileRelocatableCode() == true`.

The two functions can safely co-exist until the old one is removed from OMR.

### DIFF
--- a/runtime/compiler/codegen/J9AheadOfTimeCompile.cpp
+++ b/runtime/compiler/codegen/J9AheadOfTimeCompile.cpp
@@ -43,6 +43,60 @@ J9::AheadOfTimeCompile::emitClassChainOffset(uint8_t* cursor, TR_OpaqueClassBloc
    return cursor + SIZEPOINTER;
    }
 
+uintptr_t
+J9::AheadOfTimeCompile::findCorrectInlinedSiteIndex(void *constantPool, uintptr_t currentInlinedSiteIndex)
+   {
+   uintptr_t constantPoolForSiteIndex = 0;
+   uintptr_t inlinedSiteIndex = currentInlinedSiteIndex;
+
+   if (inlinedSiteIndex == (uintptr_t)-1)
+      {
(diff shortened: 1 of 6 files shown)


runtime/compiler/codegen/J9AheadOfTimeCompile.cpp
```diff
@@ -43,6 +43,86 @@ J9::AheadOfTimeCompile::emitClassChainOffset(uint8_t* cursor, TR_OpaqueClassBloc
...
 
+uintptr_t
+J9::AheadOfTimeCompile::findCorrectInlinedSiteIndex(void *constantPool, uintptr_t currentInlinedSiteIndex)
+   {
+   uintptr_t constantPoolForSiteIndex = 0;
+   uintptr_t inlinedSiteIndex = currentInlinedSiteIndex;
+
+   if (inlinedSiteIndex == (uintptr_t)-1)
+      {
+      constantPoolForSiteIndex = (uintptr_t)self()->comp()->getCurrentMethod()->constantPool();
```
mpirvu: Small suggestion: save the result of self()->comp() into a local and use that everywhere.

mpirvu:
Jenkins test all all jdk8

mpirvu:
Testing shows some crashes during compilation for some tests that use SCC/AOT.

ymanton:
Crash was caused by a real bug due to a typo; fixed. Also changed to use a local `comp` as suggested.

mpirvu:
Jenkins test all all jdk8

ymanton:
Fixed missing `self()` issues caught by the linter.

mpirvu:
Jenkins test all all jdk8

AdamBrousseau:
Jenkins test all all jdk8

PR #2770, opened 2018-09-05
Private methods do not override interface methods

gacholio:
As of JDK11, private methods expressly do not override methods
inherited from the superclass or superinterfaces, so the correct error
to throw in this case is AbstractMethodError, not IllegalAccessError
that's thrown by older JDK levels.


DanHeidinga:

DanHeidinga:
Testing on 10 as that will validate the code compiles and I'm not aware of any test for this particular corner case.

PR #2774, opened 2018-09-05
Handle final Object methods in InterfaceHandle

gacholio:
The final (non-virtual Object method case was not being handled
correctly in adaptInterfaceLookupsOfObjectMethodsIfRequired.

Also document that
Java_java_lang_invoke_InterfaceHandle_convertITableIndexToVTableIndex
does not need to deal with any special case interface invocations.


DanHeidinga:
Does the Unreflect case need adapting as well? https://github.com/eclipse/openj9/blob/af2d034c44c265f0bf8ec86355fc666fbfc2d96a/jcl/src/java.base/share/classes/java/lang/invoke/MethodHandles.java#L1200-L1201

gacholio:
I think the answer is yes.

gacholio:
I think unreflect just needs to allow private interface methods - it doesn't have any of the invokeinterface special cases. Since we're starting with a Method, it just needs to wrap the appropriate invocation around it (virtual/static/special).

gacholio:
Simply removing the private check completely should be sufficient - let InterfaceHandle deal with the private case, unless there's some reason why unreflect should not be symmetric.

gacholio:
Hmm, might be more complicated if we want to enforce access (allow the private method to only be invoked from within the interface).

gacholio:

gacholio:
jenkins test sanity plinux jdk10

gacholio:

gacholio:
jenkins test sanity plinux jdk10

gacholio:
This may still not be quite right - when throwAbstractMethodErrorForUnreflectPrivateInterfaceMethod is called, the value is returned immediately, rather than going through the restrictReceiver path, which may be incorrect.

gacholio:
Since there's only one caller of the helper, if I'm right, the best answer would be to inline it into the caller, which also avoids the unnecessary rechecking of modifiers.

gacholio:
unreflect fixed

gacholio:

gacholio:
jenkins test sanity plinux jdk10

jcl/src/java.base/share/classes/java/lang/invoke/MethodHandles.java
```diff
@@ -1190,12 +1196,44 @@ public MethodHandle unreflect(Method method) throws IllegalAccessException{
...
+								/*[MSG "K0587", "'{0}' no access to: '{1}'"]*/
+								String message = com.ibm.oti.util.Msg.getString("K0587", this.toString(), declaringClass + "." + method.getName() + ":" + MethodHandle.KIND_INTERFACE + "/invokeinterface");  //$NON-NLS-1$//$NON-NLS-2$ //$NON-NLS-3$ //$NON-NLS-4$
+								throw new IllegalAccessException(message);
+							}
+						}
+
+						/* JDK9 and beyond allow accessible private interface methods.
+						 * JDK8 throws AbstractMethodError for private interface methods.
+						 */
+						/*[IF !Sidecar19-SE]*/
```
DanHeidinga: Is this the best option for saying only jdk8 in the preprocessor?  Yuck.

gacholio: It's in various other places in the code. Perhaps we should switch to m4 for preprocessing...

DanHeidinga:
@gacholio Can you resolve the conflicts so this can be merged?

gacholio:
Rebased (painfully) on master, preserving the message change in the error case that was inlined.

jcl/src/java.base/share/classes/java/lang/invoke/MethodHandles.java
```diff
@@ -1207,12 +1213,44 @@ public MethodHandle unreflect(Method method) throws IllegalAccessException{
...
-						handle = new InterfaceHandle(method);
+					if (Modifier.isPrivate(methodModifiers)) {
+						/* Inlined version of access checking - this need only cover the NO_ACCESS and private_access cases */
+						if (!method.isAccessible()) {
+							if (accessMode == NO_ACCESS) {
+								throw new IllegalAccessException(this.toString());
+							}
+							if (declaringClass != accessClass || !Modifier.isPrivate(accessMode)) {
+								/*[MSG "K0678", "Class '{0}' no access to: '{1}'"]*/
+								String message = com.ibm.oti.util.Msg.getString("K0587", this.toString(), declaringClass + "." + method.getName() + ":" + MethodHandle.KIND_INTERFACE + "/invokeinterface"); //$NON-NLS-1$//$NON-NLS-2$ //$NON-NLS-3$ //$NON-NLS-4$								throw new IllegalAccessException(message);
```
DanHeidinga: Something looks funny here - `throw new IllegalAccessException(message);` is in the comment.

DanHeidinga:
jenkins compile plinux jdk10

DanHeidinga:
Running a single compile to validate this change as its been tested prior to the rebase.

gacholio:
Compile passed - I had to push a new commit to fix a bad message number in the code that I relocated from the abstract error function (the old code got fixed on master after I rebased), so this is basically ready to merge.

pshipton:
This change is causing an assert in jsr335_interfacePrivateMethod (lambda) testing, on Java 9 and above. I'm going to revert it as the fix is not expected until Wed.

`17:03:26.785 0x3ffb0089c00   j9jcl.189    *   ** ASSERTION FAILED ** at ../common/java_dyn_methodhandle.c:671: (((0 != (((methodID->vTableIndex)) & ((((UDATA)((intptr_t)1 << ((sizeof(intptr_t) * 8) - 1)))))))))`

PR #2776, opened 2018-09-05
Add literal pool register to ICF dependencies

fjeremic:
When using the `generateS390CompareAndBranchInstruction` API the passed
in constant may be out of the range of the compare and branch
instructions in which case the API will fall back to generating an
immediate compare and a branch.

In the case where the constant cannot fit even into compare immediate
instructions we fall back to comparing against the constant from the
literal pool. However if the `generateS390CompareAndBranchInstruction`
API is used within an internal control flow (ICF) region we will need
to add the register we allocate to hold the literal pool address to the
dependency conditions of the ICF.

In this case this was not done and an assertion gets triggered when the
register allocator attempts to allocate the literal pool register
within an ICF region. This is a ticking time bomb as it can lead to
undefined behavior if something needed to get spilled to satisfy the
allocation.


joransiu:
jenkins test sanity zlinux

fjeremic:
@joransiu this PR looks to be ready for merging if there are no other comments to address.

PR #2779, opened 2018-09-06
Include CodeGenerator_inlines.hpp in InlinerTempForJ9.cpp

0dvictor:
InlinerTempForJ9.cpp uses CodeGenerator which may need methods from CodeGenerator_inlines.hpp


cathyzhyi:
This change is needed for https://github.com/eclipse/omr/pull/2934 to make linker on ARM work. LGTM.

0dvictor:
> This change is needed for eclipse/omr#2934 to make linker on ARM work. LGTM.

Yes, but this changeset is actually a more general fix. It is a good practice for whatever "*.cpp" includes "CodeGenerator.hpp" also include "CodeGenerator_inlines.hpp" so that the C++ compiler has a better idea about class shape and potentially do better optimization.

fjeremic:

PR #2794, opened 2018-09-06
Prevent class initialization check bypass

gacholio:
If the JIT encounters an unresolved static field at compile-time, the
generated resolve code calls jitResolveClassFromStaticField to resolve
the declaring class of the field. This code was previously filling in
the RAM constant pool entry for the static field, allowing other threads
to bypass the class initialization check, resulting in access to the
static field before the class is fully initialized.

The fix is to resolve the static field into local data instead of the
constant pool. This call will be immediately followed up by the
jitResolveStaticField or jitResolveStaticFieldSetter, which will fully
resolve the constant pool entry in the expected case.


runtime/oti/VMHelpers.hpp
```diff
@@ -1060,6 +1060,20 @@ class VM_VMHelpers
...
+	 *
+	 * @param flagsAndClass[in] field from the ref
+	 * @param valueOffset[in] field from the ref
+	 *
+	 * @returns true if resolved, false if not
+	 */
+	static VMINLINE bool
+	staticFieldRefResolved(IDATA flagsAndClass, UDATA valueOffset)
+	{
+		return ((UDATA)-1 != valueOffset) && (flagsAndClass > 0);
```
DanHeidinga: This helper is a bit of an attractive nuisance as it doesn't differentiate between resolved for put vs get.
See the `putstatic` code in the BytecodeInterpreter: https://github.com/eclipse/openj9/blob/0a25dca436291a5bc981d891b32b4fa763ca487f/runtime/vm/BytecodeInterpreter.hpp#L6260-L6261

gacholio: There isn't a tag for get, so it's really put vs get/dontknoworcare. I can add a bool for put and use is in the interpreter as well if you like.

DanHeidinga: +1 to adding the bool - will hopefully make users think about the purpose and if they need put vs don'tcare semantics

gacholio: In deference to having to think about it, I won't add a default value for the bool.

gacholio:
Still needs work, marking WIP.

gacholio:
Ready for review.

DanHeidinga:

DanHeidinga:
Jenkins test sanity xlinuxlargeheap jdk10

DanHeidinga:
@AdamBrousseau What's the jenkins command to rerun the line endings check?  It failed due to git issues on the executator

AdamBrousseau:
Jenkins line endings check

AdamBrousseau:
Looks like the workspace on master needs a cleanup first

AdamBrousseau:
Jenkins line endings check

AdamBrousseau:

PR #2799, opened 2018-09-07
Fix duplicate MSG key

keithc-ca:
Also fix preprocess to include the existing value (instead of "null").


keithc-ca:
Without this change builds don't get far:
```
jdk11$ make clean
ERROR [preprocessor] MSG command syntax error, command ignored: MSG "K0675", "Unexpected MethodHandle instance: {0} with {1}" - Detail: "K0675" already has the value "null"
    file: /home/keithc/space/jdk11/openj9/jcl/src/java.base/share/classes/java/lang/invoke/MemberName.java
    line: 89 (char 0 to 69)
```

DanHeidinga:
Jenkins compile xlinux jdk11

PR #2803, opened 2018-09-07
Handle string conversion errors in attach API

pdbain-ibm:
If the name of the temporary directory cannot be converted to UTF-16, use the
Java system property.

Fixes https://github.com/eclipse/openj9/issues/2781

### DIFF
--- a/jcl/src/java.base/share/classes/com/ibm/tools/attach/target/IPC.java
+++ b/jcl/src/java.base/share/classes/com/ibm/tools/attach/target/IPC.java
@@ -51,6 +51,7 @@ import static java.nio.file.attribute.PosixFilePermission.OTHERS_WRITE;
  */
 public class IPC {
 
+	private static final String JAVA_IO_TMPDIR = "java.io.tmpdir";
 	/**
 	 * Successful return code from natives.
 	 */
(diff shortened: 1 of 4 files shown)


pdbain-ibm:
@pshipton would you kindly review?

jcl/src/java.base/share/classes/com/ibm/tools/attach/target/IPC.java
```diff
@@ -261,7 +262,8 @@ static private native int createFileWithPermissionsImpl(String path,
 	static String getTmpDir() {
 		String tmpDir = getTempDirImpl();
 		if (null == tmpDir) {
-			tmpDir = com.ibm.oti.vm.VM.getVMLangAccess().internalGetProperties().getProperty("java.io.tmpdir"); //$NON-NLS-1$
+			IPC.logMessage("Could not get system temporary directory. Trying "+JAVA_IO_TMPDIR); //$NON-NLS-1$
+			tmpDir = com.ibm.oti.vm.VM.getVMLangAccess().internalGetProperties().getProperty(JAVA_IO_TMPDIR); //$NON-NLS-1$
```
CodingFabian: will this default to /tmp if not set?

pdbain-ibm: The system property is set by `getPlatformPropertyList()` to the value of the TMPDIR environment variable (sans character set conversion, which is the present problem).  If that does not resolve to a valid path, then it defaults to "/tmp".

PR #2804, opened 2018-09-07
Fix condy constantpool iterator code

tajila:
Fix condy constantpool iterator code

In cases where we are only iterating condy slots, the constant
pool iterator breaks out of the loop which causes us to not 
update the iteration variables. As a result the iterator cycles 
indefinitely.


dmitripivkine:
Jenkins test sanity pLinux jdk11

tajila:
@dmitripivkine failure in the sanity PR build is a pipeline failure and not related to the changes in this PR, can you please re-run the job

PR #2807, opened 2018-09-07
Fix supportsIntrinsicCaseConversion in ilgen

cathyzhyi:
The call node generated was not cleaned up from the oprand stack when
loading const instead. Besides, the code for handling this recognized
method can be moved up to avoid applying the redundant logic for a
general call.

### DIFF
--- a/runtime/compiler/ilgen/Walker.cpp
+++ b/runtime/compiler/ilgen/Walker.cpp
@@ -4643,13 +4643,6 @@ TR_J9ByteCodeIlGenerator::genInvoke(TR::SymbolReference * symRef, TR::Node *indi
 
     */
 
-
-   TR::Node * savedNode = NULL;
-   if (_stack->topIndex() > 0)
-      {
-      savedNode = _stack->element(_stack->topIndex());
-      }
-
 #define DAA_PRINT(a) \
 case a: \
    traceMsg(comp(), "DAA Method found: %s\n", #a); \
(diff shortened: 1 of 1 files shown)


runtime/compiler/ilgen/Walker.cpp
```diff
@@ -4955,7 +4948,17 @@ break
...
-    int32_t constToLoad = 0;
+   int32_t constToLoad = 0;
+   if(symbol->getRecognizedMethod() == TR::com_ibm_jit_JITHelpers_supportsIntrinsicCaseConversion)
+      {
+      if (cg()->getSupportsInlineStringCaseConversion())
+         constToLoad = 1;
+      else
+         constToLoad  = 0; // should already be 0 but just in case..
+      pop(); //pop up the receiver since it's not used
+      loadConstant(TR::iconst, constToLoad);
```
fjeremic: How about `constToLoad = cg()->getSupportsInlineStringCaseConversion() ? 1 : 0` or alternatively just inline the that into `loadConstant`?

cathyzhyi: changed according to the comment. thanks!

runtime/compiler/ilgen/Walker.cpp
```diff
@@ -4955,7 +4948,13 @@ break
        return NULL;
        }
 
-    int32_t constToLoad = 0;
+   int32_t constToLoad = 0;
+   if(symbol->getRecognizedMethod() == TR::com_ibm_jit_JITHelpers_supportsIntrinsicCaseConversion)
```
fjeremic: `constToLoad` is no longer used and can be removed. Also add space after `if`

cathyzhyi: changed according to comments. thanks!

fjeremic:
Jenkins test sanity JDK8

fjeremic:
Jenkins test sanity all JDK8

PR #2809, opened 2018-09-08
Nestmates invokevirtual and invokeinterface work for Power

IBMJimmyk:
I'm implementing this on Power:
https://github.com/eclipse/openj9/issues/2673

Changes were made to support private resolved interface and virtual calls.

For invokeinterface the J2I thunk was added to the PPCInterfaceCallSnippet snipper. The interfaceCallHelper assembly is now able to recognize tagged J9Methods returned from jitResolveInterfaceMethod and new code was added to handle this case.

Similar changes were made for invokevirtual. The J2I thunk and a location to store the J9Method were added to the PPCVirtualUnresolvedSnippet snippet. Furthermore a branch back to the mainline code was added after the bl to the helper. virtualUnresolvedHelper is also now able to recognize and handle tagged J9Methods returned from jitResolveVirtualMethod. A small complication was added to how the value stored to the lr. Under the new private virtual method path, the lr is given a return address back to the new b instruction in the snippet. Under the older, original path, the lr is a return address back to the mainline code. The code tries to jump back to the new b instruction if possible since that is the instruction right after the bl to the helper and there can be a performance hit from not doing this. The original path requires the lr to hold a return address back to mainline code so that was not changed.

### DIFF
--- a/runtime/compiler/p/codegen/CallSnippet.cpp
+++ b/runtime/compiler/p/codegen/CallSnippet.cpp
@@ -525,8 +525,17 @@ uint8_t *TR::PPCVirtualUnresolvedSnippet::emitSnippetBody()
    TR_J9VMBase *fej9 = (TR_J9VMBase *)(comp->fe());
    TR::Node       *callNode = getNode();
    TR::SymbolReference *glueRef = cg()->symRefTab()->findOrCreateRuntimeHelper(TR_PPCvirtualUnresolvedHelper, false, false, false);
-   intptrj_t  distance = (intptrj_t)glueRef->getMethodAddress() - (intptrj_t)cursor;
+   intptrj_t  distance;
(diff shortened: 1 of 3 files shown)


IBMJimmyk:
@gita-omr I think you wanted to take a look at the changes. Please tag anyone else who might be interested as well.

IBMJimmyk:
I modified my implementation of the invokevirtual changes. I added a new branch instruction into the snippet that jumps back to the main line code. Under the private nestmates path, this helper eventually returns back to this branch before going back to the mainline code. The original non-private-nestmate virtualUnresolvedHelper path still goes directly back to the mainline code.

@zl-wang Could you take a look at these changes?

IBMJimmyk:
The tests I've run so far have passed so I've removed the WIP tag. If there aren't any other suggested changes, I think this is ready to be checked in.

gita-omr:
Excellent! I will merge then (Julian've seen all the changes).

PR #2815, opened 2018-09-10
Merge archiveArtifact lines

jdekonin:
* [skip ci]
* on the single line removes the need to wrap a conditional around
    the native-test-libs.tar.gz as its only created in JDK11+


jdekonin:
@AdamBrousseau @llxia, please review.

PR #2818, opened 2018-09-10
Revert "Handle interface special cases for MethodHandle"

pshipton:
Reverts eclipse/openj9#2774

The change is causing an assert in jsr335_interfacePrivateMethod (lambda) testing, on Java 9 and above. I'm reverting it as the fix is not expected until Wed.

`17:03:26.785 0x3ffb0089c00 j9jcl.189 * ** ASSERTION FAILED ** at ../common/java_dyn_methodhandle.c:671: (((0 != (((methodID->vTableIndex)) & ((((UDATA)((intptr_t)1 << ((sizeof(intptr_t) * 8) - 1)))))))))`

### DIFF
--- a/jcl/src/java.base/share/classes/java/lang/invoke/MethodHandles.java
+++ b/jcl/src/java.base/share/classes/java/lang/invoke/MethodHandles.java
@@ -864,13 +864,7 @@ public class MethodHandles {
 					/* Interfaces only inherit *public* methods from Object */
 					throw new NoSuchMethodException(clazz + "." + methodName + type); //$NON-NLS-1$					
 				}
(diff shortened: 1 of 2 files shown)


DanHeidinga:
@pshipton Can you open an issue to track reapplying this change?

pshipton:
@DanHeidinga I don't see the point, just seems like more things to track. @gacholio was already working on a fix which is expected to be ready on Wed.

gacholio:
Agreed - I will re-add the reverted code along with the fix (they're intertwined anyway, and I need the old code to test the fix).

DanHeidinga:
That's fine by me.  I'm just concerned that the author may not always be aware when code is reverted (ie: on vacation, etc).  In general, I'd rather have an extra issue to track getting the right fix released.

PR #2138, opened 2018-06-08
Move all job PROPERTIES to Jenkins Files for Build and Pipeline jobs

AdamBrousseau:
- Add Build Discarder
  - Retrieve values from Variable file
- Add job Parameters
  - Default values for Vendor Params must be set globally
    even if they are blank
- No Trigger for Build and Pipeline jobs
- TODO add trigger for PR builds

[skip ci]
Issue #1017
Signed-off-by: Adam Brousseau <adam.brousseau88@gmail.com>

Depends ~#2093~

PR #2820, opened 2018-09-10
Don't pass params directly to downstream jobs

AdamBrousseau:
- Using params.PARAM will allow null to be passed
  to a downstream job. This causes a failure if the
  downstream job has the parameter configured.
- Null value not allowed as an environment variable
- Adding a check and setting the variable to blank
  should resolve this failure

[skip ci]
Issue #2138


AdamBrousseau:
@vsebe for review
Seems to be working
https://ci.eclipse.org/openj9/job/Pipeline-Build-Test-JDK10-linux_ppc-64_cmprssptrs_le/163/console

PR #2827, opened 2018-09-11
Remove unnecessary synchronization in attach API logging

pdbain-ibm:
Avoid synchronization when  the enabled state of logging is known.

- Remove the `volatile` declaration of the logging status variable.
- If the current thread's copy of `loggingStatus` is in a terminal state (enabled or disabled), avoid  synchronizing. 
- In heavily used functions, test `loggingStatus` for the most probable value directly without a method call.
- Clean up and reorganize some logging variables and methods.

Fixes https://github.com/eclipse/openj9/issues/2048

@pshipton would you kindly review this?  Thank you.

### DIFF
--- a/jcl/src/java.base/share/classes/com/ibm/tools/attach/target/AttachHandler.java
+++ b/jcl/src/java.base/share/classes/com/ibm/tools/attach/target/AttachHandler.java
@@ -1,6 +1,10 @@
 /*[INCLUDE-IF Sidecar16]*/
 package com.ibm.tools.attach.target;
 
+import static com.ibm.tools.attach.target.IPC.LOGGING_DISABLED;
+import static com.ibm.tools.attach.target.IPC.LOGGING_ENABLED;
+import static com.ibm.tools.attach.target.IPC.loggingStatus;
+
 /*******************************************************************************
  * Copyright (c) 2009, 2018 IBM Corp. and others
  *
@@ -114,7 +118,6 @@ public class AttachHandler extends Thread {
 	/* only the attach handler thread uses syncFileLock */
 	/* [PR Jazz 30075] Make syncFileLock an instance variable since it is accessed only by the attachHandler singleton. */
 	FileLock syncFileLock;
-	private PrintStream logStream; 
 
 	/**
 	 * Keep the constructor private
@@ -203,7 +206,7 @@ public class AttachHandler extends Thread {
 				/* The following code needs to hold the lock, so go for a blocking lock. */
 				CommonDirectory.obtainMasterLock();
 			}
-			if (IPC.loggingEnabled ) {
+			if (LOGGING_DISABLED != loggingStatus) {
 				IPC.logMessage("AttachHandler obtained master lock"); //$NON-NLS-1$
 			}
 			CommonDirectory.createNotificationFile();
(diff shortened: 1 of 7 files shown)


jcl/src/java.base/share/classes/com/ibm/tools/attach/target/AttachHandler.java
```diff
@@ -287,15 +290,17 @@ private boolean initialize() throws IOException {
...
-			IPC.loggingEnabled = true;
-			IPC.setDefaultVmId(pidProperty);
-			IPC.logMessage("AttachHandler initialize"); //$NON-NLS-1$
+		synchronized (IPC.accessorMutex) {
+			if ((null == IPC.logStream) && (null != loggingProperty)
+					&& loggingProperty.equalsIgnoreCase("yes")) { //$NON-NLS-1$
+				File logFile = new File(logName + pidProperty + ".log"); //$NON-NLS-1$
+				IPC.logStream = new PrintStream(logFile);
+				IPC.setDefaultVmId(pidProperty);
+				IPC.logMessage("AttachHandler initialize"); //$NON-NLS-1$
```
pshipton: This message won't be logged since logging is not yet enabled.

jcl/src/java.base/share/classes/com/ibm/tools/attach/target/IPC.java
```diff
@@ -365,49 +394,50 @@ public static void logMessage(String string1, int int1, String string2) {
...
+	private static void printLogMessage(final String msg) {
+		synchronized (accessorMutex) {
+			if (!Objects.isNull(logStream)) {
+				printMessageWithHeader(msg, logStream);
+				logStream.flush();
+			}
 		}
 	}
 
+	private static void printMessageWithHeader(String msg, PrintStream log) {
```
pshipton: This method should have a comment that it must be called from inside a synchronized block.

jcl/src/java.base/share/classes/com/ibm/tools/attach/target/AttachHandler.java
```diff
@@ -287,15 +290,17 @@ private boolean initialize() throws IOException {
...
+				loggingStatus = LOGGING_ENABLED;
```
pshipton: I don't think this implementation is 100% proper. In theory the IPC logging code could see that logging is enabled before the logstream is fully initialized.

jcl/src/java.base/share/classes/com/ibm/tools/attach/target/AttachHandler.java
pdbain-ibm: Right.  Changing to unconditional print.

jcl/src/java.base/share/classes/com/ibm/tools/attach/target/IPC.java
pdbain-ibm: Good plan.

jcl/src/java.base/share/classes/com/ibm/tools/attach/target/AttachHandler.java
pshipton: Or you can set LOGGING_ENABLED before calling logMessage()

jcl/src/java.base/share/classes/com/ibm/tools/attach/target/AttachHandler.java
pdbain-ibm: `logStream` is initialized and used inside synchronized blocks, so it cannot be used until initialization is complete.

pshipton:

PR #2828, opened 2018-09-11
Remove unused function set_job_names

AdamBrousseau:
- This function was added but has never been used.
- We set the job names directly in the workflow function.
- Cleanup to avoid confusion.

[skip ci]


AdamBrousseau:
@vsebe for review

AdamBrousseau:
@llxia for final review

PR #2829, opened 2018-09-11
Report illegal write to static final field for Unsafe atomic calls

liqunl:
Add a diamond to take the original call if the field is static final.
The existing API calculateUnsafeAddress does not meet the need of
reporting illegal write, thus move the code to processUnsafeAtomicCall.


0dvictor:
This change itself looks good to me; however I don't think https://github.com/eclipse/omr/pull/2952 is required. If certain functions' transformation is disabled, it should be disabled here in their corresponding `isInlineable()` instead of OMR.

liqunl:
WIP the PR before tests have finished. Victor's comment has been addressed. @andrewcraik Can you take a look at the changes?

runtime/compiler/optimizer/J9RecognizedCallTransformer.cpp
```diff
@@ -102,26 +103,211 @@ void J9::RecognizedCallTransformer::process_java_lang_StringUTF16_toBytes(TR::Tr
...
+      objectNode = unsafeCall->getChild(1);
+      offsetNode = unsafeCall->getChild(2);
+
+      // Null Check
+      if (needsNullCheck)
+         {
+         auto NULLCHKNode = TR::Node::createWithSymRef(TR::NULLCHK, 1, 1,
+                                                       TR::Node::create(TR::PassThrough, 1, objectNode->duplicateTree()),
+                                                       comp()->getSymRefTab()->findOrCreateNullCheckSymbolRef(comp()->getMethodSymbol()));
+         NULLCHKNode->getByteCodeInfo().setCallerIndex(comp()->getCurrentInlinedSiteIndex());
```
andrewcraik: what bytecode offset do we give this? it should match the original unsafe call in case we have to yield I think...

runtime/compiler/optimizer/J9RecognizedCallTransformer.cpp
```diff
@@ -102,26 +103,211 @@ void J9::RecognizedCallTransformer::process_java_lang_StringUTF16_toBytes(TR::Tr
    treetop->insertAfter(TR::TreeTop::create(comp(), TR::Node::create(node, TR::treetop, 1, newCallNode)));
    }
 
-void J9::RecognizedCallTransformer::processUnsafeAtomicCall(TR::TreeTop* treetop, TR::Node* node, TR::SymbolReferenceTable::CommonNonhelperSymbol helper)
+/*
```
andrewcraik: Could you make the names in the code match the names in the diagram (for blocks and treetops) to make checking this easier?

andrewcraik:
Given the complexity of the transformation there needs to be a lot more trace in this code to make it easier to debug. The trace should be guarded in some fashion, but I think it is important to record which blocks are what, what nodes are the treetops of the tests etc in case something goes wrong later on

runtime/compiler/optimizer/J9RecognizedCallTransformer.cpp
liqunl: I passed `unsafeCall` as the originating node to `createWithSymRef`. The null check node should have the bytecode info of the `unsafeCall` now.

runtime/compiler/optimizer/J9RecognizedCallTransformer.cpp
liqunl: Changes made.

liqunl:
@andrewcraik I made some changes to address your comments, can you take a look?

andrewcraik:
Hmm this is the second place where we are testing for a static final field where what we want are the chain of tests needed to say do we need to report this write. I think we may need to factor this code before the sequence of conditions sprays itself around the codebase since we may want to increase the trust set in future.

The trace looks a lot better - thanks for that.

liqunl:
@andrewcraik  I have the environment variable in the commit already. This change is ready for jenkins test and to be merged.

liqunl:
This PR is part of https://github.com/eclipse/openj9/issues/3018

liqunl:
Perf measured on daytrader3 and Hibench, on regression.

runtime/compiler/optimizer/J9RecognizedCallTransformer.cpp
```diff
@@ -102,34 +104,237 @@ void J9::RecognizedCallTransformer::process_java_lang_StringUTF16_toBytes(TR::Tr
...
+start Block_A
+...
+ifacmpeq -> <Block_E>
+  object
+  aconst null
+end Block_A
+
+start Block_B
+iflcmpne -> <Block_E>
+  lload <offset>
```
andrewcraik: There's an and involved here for the tag that is missing in this comment right?

liqunl: Thanks for pointing this out. I've updated the comment as well as the code such that the comparison matches the one in the comment.

andrewcraik:

PR #2830, opened 2018-09-11
Change .i macro annotations to not use @, pt. 2

mikezhang1234567890:
With https://github.com/eclipse/omr/pull/2929, the change to ddrgen to not annotate source files with the @ symbol merged, we can now simplify the sed command to only match the new pattern.

Signed-off-by: Mike Zhang <mike.h.zhang@ibm.com>


mikezhang1234567890:
@keithc-ca can you please review?

keithc-ca:

PR #2835, opened 2018-09-11
Update ldc to load Constant_Dynamic primitives as I_32 type

fengxue-IS:
ldc loads data into a UDATA slot on the stack and primitive values
only uses 32bit on a 64bit system, this change will ensure that
32 bit values will always be written to the lower 32 bit of the slot
(For BE and LE systems)

### DIFF
--- a/runtime/vm/BytecodeInterpreter.hpp
+++ b/runtime/vm/BytecodeInterpreter.hpp
@@ -6116,34 +6116,46 @@ retry:
 resolved:
 		_pc += (1 + parmSize);
 		_sp -= 1;
-		if (J9DescriptionCpTypeClass == romCPEntry->cpType) {
-			value = J9VM_J9CLASS_TO_HEAPCLASS((J9Class*)value);
-		} else if (J9DescriptionCpTypeConstantDynamic == (romCPEntry->cpType & J9DescriptionCpTypeMask)) {
+		
+		if ((J9DescriptionCpTypeConstantDynamic == (romCPEntry->cpType & J9DescriptionCpTypeMask))
+		&& (0 != (romCPEntry->cpType >> J9DescriptionReturnTypeShift))
+		) {
 			/* Constant Dynamic ROM CP entry uses J9DescriptionReturnType* flag to indicate
 			 * different primitive return type that require unboxing before returning the value
 			 */
+			I_32 unboxedValue = 0;
 			switch (romCPEntry->cpType >> J9DescriptionReturnTypeShift) {
 			case J9DescriptionReturnTypeBoolean:
-				value = (j9object_t)(UDATA)J9VMJAVALANGBOOLEAN_VALUE(_currentThread, value);
+				unboxedValue = (I_32)J9VMJAVALANGBOOLEAN_VALUE(_currentThread, value);
 				break;
 			case J9DescriptionReturnTypeByte:
-				value = (j9object_t)(UDATA)J9VMJAVALANGBYTE_VALUE(_currentThread, value);
+				unboxedValue = (I_32)(I_8)J9VMJAVALANGBYTE_VALUE(_currentThread, value);
 				break;
 			case J9DescriptionReturnTypeChar:
-				value = (j9object_t)(UDATA)J9VMJAVALANGCHARACTER_VALUE(_currentThread, value);
+				unboxedValue = (I_32)(U_16)J9VMJAVALANGCHARACTER_VALUE(_currentThread, value);
 				break;
 			case J9DescriptionReturnTypeShort:
-				value = (j9object_t)(UDATA)J9VMJAVALANGSHORT_VALUE(_currentThread, value);
+				unboxedValue = (I_32)(I_16)J9VMJAVALANGSHORT_VALUE(_currentThread, value);
 				break;
 			case J9DescriptionReturnTypeFloat:
-				value = (j9object_t)(UDATA)J9VMJAVALANGFLOAT_VALUE(_currentThread, value);
+				unboxedValue = (I_32)J9VMJAVALANGFLOAT_VALUE(_currentThread, value);
 				break;
 			case J9DescriptionReturnTypeInt:
-				value = (j9object_t)(UDATA)J9VMJAVALANGINTEGER_VALUE(_currentThread, value);
+				unboxedValue = (I_32)J9VMJAVALANGINTEGER_VALUE(_currentThread, value);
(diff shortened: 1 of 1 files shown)


fengxue-IS:

runtime/vm/BytecodeInterpreter.hpp
```diff
@@ -6124,24 +6124,33 @@ done:;
 			 */
 			switch (romCPEntry->cpType >> J9DescriptionReturnTypeShift) {
 			case J9DescriptionReturnTypeBoolean:
-				value = (j9object_t)(UDATA)J9VMJAVALANGBOOLEAN_VALUE(_currentThread, value);
+				*(I_32 *)_sp = (I_32)J9VMJAVALANGBOOLEAN_VALUE(_currentThread, value);
```
DanHeidinga: I would prefer to use `(U_32)` to `(I_32)` - no accidental chance of sign extensions.  Though a boolean should be 0 or 1.

runtime/vm/BytecodeInterpreter.hpp
```diff
@@ -6124,24 +6124,33 @@ done:;
...
 				break;
 			case J9DescriptionReturnTypeByte:
-				value = (j9object_t)(UDATA)J9VMJAVALANGBYTE_VALUE(_currentThread, value);
+				*(I_32 *)_sp = (I_32)J9VMJAVALANGBYTE_VALUE(_currentThread, value);
```
DanHeidinga: Should be `(I_32)(I_8) J9VMJAVALANGBYTE_VALUE `

runtime/vm/BytecodeInterpreter.hpp
```diff
@@ -6124,24 +6124,33 @@ done:;
...
 				break;
 			case J9DescriptionReturnTypeChar:
-				value = (j9object_t)(UDATA)J9VMJAVALANGCHARACTER_VALUE(_currentThread, value);
+				*(I_32 *)_sp = (I_32)J9VMJAVALANGCHARACTER_VALUE(_currentThread, value);
 				break;
 			case J9DescriptionReturnTypeShort:
-				value = (j9object_t)(UDATA)J9VMJAVALANGSHORT_VALUE(_currentThread, value);
+				*(I_32 *)_sp = (I_32)J9VMJAVALANGSHORT_VALUE(_currentThread, value);
```
DanHeidinga: `(I_32)(I_16)`

runtime/vm/BytecodeInterpreter.hpp
```diff
@@ -6124,24 +6124,33 @@ done:;
...
-				value = (j9object_t)(UDATA)J9VMJAVALANGBOOLEAN_VALUE(_currentThread, value);
+				*(I_32 *)_sp = (I_32)J9VMJAVALANGBOOLEAN_VALUE(_currentThread, value);
 				break;
 			case J9DescriptionReturnTypeByte:
-				value = (j9object_t)(UDATA)J9VMJAVALANGBYTE_VALUE(_currentThread, value);
+				*(I_32 *)_sp = (I_32)J9VMJAVALANGBYTE_VALUE(_currentThread, value);
 				break;
 			case J9DescriptionReturnTypeChar:
-				value = (j9object_t)(UDATA)J9VMJAVALANGCHARACTER_VALUE(_currentThread, value);
+				*(I_32 *)_sp = (I_32)J9VMJAVALANGCHARACTER_VALUE(_currentThread, value);
```
DanHeidinga: `(I_32)(U_16)`

runtime/vm/BytecodeInterpreter.hpp
```diff
@@ -6124,24 +6124,33 @@ done:;
...
 				break;
 			case J9DescriptionReturnTypeShort:
-				value = (j9object_t)(UDATA)J9VMJAVALANGSHORT_VALUE(_currentThread, value);
+				*(I_32 *)_sp = (I_32)J9VMJAVALANGSHORT_VALUE(_currentThread, value);
 				break;
 			case J9DescriptionReturnTypeFloat:
-				value = (j9object_t)(UDATA)J9VMJAVALANGFLOAT_VALUE(_currentThread, value);
+				*(I_32 *)_sp = (I_32)J9VMJAVALANGFLOAT_VALUE(_currentThread, value);
```
DanHeidinga: Use `(U_32)` instead of I_32 in both cases.

runtime/vm/BytecodeInterpreter.hpp
```diff
@@ -6124,24 +6124,33 @@ done:;
...
-				value = (j9object_t)(UDATA)J9VMJAVALANGINTEGER_VALUE(_currentThread, value);
+				*(I_32 *)_sp = (I_32)J9VMJAVALANGINTEGER_VALUE(_currentThread, value);
+				break;
+			case J9DescriptionReturnTypeDouble: /* fall through */
+			case J9DescriptionReturnTypeLong:
+				/* double and long value cannot be loaded by ldc as they require 2 slots */
+				Assert_VM_unreachable();
+				break;
+			default:
+				*_sp = (UDATA)value;
```
DanHeidinga: Why add the default when it duplicates the write below on line 6155?

fengxue-IS: this avoid having goto for each of the switch case

DanHeidinga: Fair enough.  The code will be ugly one or the other =)

DanHeidinga:
Jenkins test sanity plinux jdk8,11

DanHeidinga:
Jenkins test sanity zlinux jdk8,11

gacholio:
Personally, I would have used a local for the 32-bit value, and not bothered adding the impossible cases to the case statement (that's real code being generated there unnecessarily).

DanHeidinga:
@fengxue-IS Can you update as per GAC's suggestion?

gacholio:
I suppose it's consistent with the rest of the code to assert in the default case (which is better than picking which cases are unimplmented). We can revisit this later on throughout the interpreter.

fengxue-IS:
updated code, @gacholio can you take a look

gacholio:
I'd still like to see:

- a single I_32 local and final assignment to sp
- default rather than the long cases for the assert

gacholio:
Looks good - please squash (and I suggest adding [ci skip] into the squashed commit message).

gacholio:

PR #2843, opened 2018-09-12
Report illegal write for Unsafe.put

liqunl:
Add a diamond to take the original call if the field is static final.

### DIFF
--- a/runtime/compiler/optimizer/InlinerTempForJ9.cpp
+++ b/runtime/compiler/optimizer/InlinerTempForJ9.cpp
@@ -61,6 +61,7 @@
 #include "runtime/J9Profiler.hpp"
 #include "ras/DebugCounter.hpp"
 #include "j9consts.h"
+#include "optimizer/TransformUtil.hpp"
 
 namespace TR { class SimpleRegex; }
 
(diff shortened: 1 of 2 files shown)


liqunl:
@0dvictor Can you help the review?
Depends https://github.com/eclipse/omr/pull/2959

0dvictor:
LGTM.

andrewcraik:
My main comment is that we need a way to say whether we are needing to report illegal writes - if we never make assumptions on a given class or if we want to disable the opt we need a way to turn that off. Perhaps we need a function somewhere sensible to say are we reporting writes on a given field and perhaps pass the node that is the object so that if we need to be selective in the future we can be.

liqunl:
I have the env var in this commit already. It's `TR_DisableIllegalWriteReport`. This change is ready for a jenkins build.

liqunl:
This PR is part of https://github.com/eclipse/openj9/issues/3018

andrewcraik:

liqunl:
Perf measured on daytrader3 and Hibench, on regression.

runtime/compiler/optimizer/InlinerTempForJ9.cpp
```diff
@@ -1060,91 +1061,62 @@ TR_J9InlinerPolicy::createUnsafePutWithOffset(TR::ResolvedMethodSymbol *calleeSy
...
-                              offset, newSymbolReferenceForAddress, false);
+   static char *disableIllegalWriteReport = feGetEnv("TR_DisableIllegalWriteReport");
+   TR::TreeTop* callTreeForIllegalWriteReport = NULL;
+   // Keep the call for illegal write report
+   if (!disableIllegalWriteReport && !comp()->getOption(TR_DisableGuardedStaticFinalFieldFolding))
+      callTreeForIllegalWriteReport = callNodeTreeTop->duplicateTree();
 
    if(comp()->getOption(TR_TraceUnsafeInlining))
        {
        traceMsg(comp(),"\t After createTempsForUnsafePutGet, callNodeTreeTop dump:\n");
```
andrewcraik: since you removed the createTempsForUnsafePutGet this trace is misleading now.

andrewcraik:
Sanity was previously clean and the last update was a simple removal. With a clean CI this is good to go.

PR #2844, opened 2018-09-12
Clean up pushArguments()

pdbain-ibm:
Remove goto in switch statement which caused PPCLE GCC to generate incorrect code.
Move local variables to inner scopes.
Remove deprecated usage of '--'.

Fixes https://github.com/eclipse/openj9/issues/2788

### DIFF
--- a/runtime/vm/jnicsup.cpp
+++ b/runtime/vm/jnicsup.cpp
@@ -445,10 +445,6 @@ UDATA JNICALL   pushArguments(J9VMThread *vmThread, J9Method* method, void *args
 	jvalue* jvalues;
 	U_8 *sigChar;
 	jobject objArg;
-	jdouble dbl;
-	jlong lng;
-	UDATA* lngOrDblPtr;
-	jfloat *fltPtr;
 	UDATA* sp;
 
 	if ( (UDATA)args & 1 ) {
@@ -465,17 +461,24 @@ UDATA JNICALL   pushArguments(J9VMThread *vmThread, J9Method* method, void *args
 	sp = vmThread->sp;
 
 	for (;;) {
+		BOOLEAN skipSignature = TRUE;
 		switch (*sigChar++) {
 			case '[':
 				/* skip the rest of the signature */
-				while (*sigChar == '[') sigChar++;
-				if (*sigChar++ != 'L') goto dontSkipClassName;
-			case 'L':
+				while ('[' == *sigChar) {
+					sigChar += 1;
+				}
+				skipSignature = ('L' == *sigChar++);
+			case 'L': /* FALLTHROUGH */
 				/* skip the rest of the signature */
-				while (*sigChar++ != ';');
-dontSkipClassName:
+				if (skipSignature) {
+					while (';' != *sigChar) {
+						sigChar += 1;
+					}
+				}
+				sp -= 1;
 				objArg = ARG(jobject, l);
-				*--sp = objArg ? (UDATA)*(j9object_t*)objArg : 0;
+				*sp = (NULL == objArg) ? 0: (UDATA) *((j9object_t*) objArg);
 				break;
 			case 'B':
 				/* byte type */
@@ -500,29 +503,20 @@ dontSkipClassName:
 #ifdef J9VM_INTERP_FLOAT_SUPPORT
 			case 'F':
 				/* float type */
-				fltPtr = (jfloat*)--sp;
-				/* The Linux SH4 compiler needs the next two lines to be two steps.  If you combine them it fails to compile */
-				dbl = ARG(jdouble, f);
-				*fltPtr = (jfloat)dbl;
-				break;
+				sp -= 1;
+				/* jfloat is promoted to double when passed through '...' */
+				*((jfloat*) sp) = (jfloat) ARG(jdouble, f);
+			break;
 			case 'D':
 				/* double type */
-				dbl = ARG(jdouble, d);
-				lngOrDblPtr = (UDATA *) &dbl;
-				goto pushLongOrDouble;
+				sp -= 2;
+				*((jdouble*) sp) = ARG(jdouble, d);
+				break;
 #endif
 			case 'J':
 				/* long type */
-				lng = ARG(jlong, j);
-				lngOrDblPtr = (UDATA *) &lng;
-pushLongOrDouble:
-#ifdef J9VM_ENV_DATA64
-				--sp;
-				*--sp = *(lngOrDblPtr);
-#else
-				*--sp = *(lngOrDblPtr + 1);
-				*--sp = *(lngOrDblPtr);
-#endif
+				sp -= 2;
+				*((jlong*) sp) = ARG(jlong, j);
 				break;
 			case ')':
 				vmThread->sp = sp;


pdbain-ibm:
@gacholio would you kindly review?

runtime/vm/jnicsup.cpp
```diff
@@ -498,30 +494,45 @@ UDATA JNICALL   pushArguments(J9VMThread *vmThread, J9Method* method, void *args
...
 #ifdef J9VM_INTERP_FLOAT_SUPPORT
-			case 'F':
+			case 'F': {
 				/* float type */
-				fltPtr = (jfloat*)--sp;
-				/* The Linux SH4 compiler needs the next two lines to be two steps.  If you combine them it fails to compile */
-				dbl = ARG(jdouble, f);
-				*fltPtr = (jfloat)dbl;
-				break;
+				jfloat *fltPtr = (jfloat*)--sp;
```
gacholio: Why use a local here instead of just assigning after the double fetch?

runtime/vm/jnicsup.cpp
```diff
@@ -498,30 +494,45 @@ UDATA JNICALL   pushArguments(J9VMThread *vmThread, J9Method* method, void *args
...
 			case 'J':
 				/* long type */
-				lng = ARG(jlong, j);
-				lngOrDblPtr = (UDATA *) &lng;
-pushLongOrDouble:
 #ifdef J9VM_ENV_DATA64
-				--sp;
-				*--sp = *(lngOrDblPtr);
+				sp -= 2;
+				*sp = (UDATA) ARG(jlong, j);
```
gacholio: This code should work for both 32 and 64 bit.

gacholio: Rather, `*(U_64*)sp = ARG(...)`

runtime/vm/jnicsup.cpp
```diff
@@ -498,30 +494,45 @@ UDATA JNICALL   pushArguments(J9VMThread *vmThread, J9Method* method, void *args
...
+			}
+			break;
 			case 'D':
 				/* double type */
-				dbl = ARG(jdouble, d);
-				lngOrDblPtr = (UDATA *) &dbl;
-				goto pushLongOrDouble;
+#ifdef J9VM_ENV_DATA64
+				sp -= 2;
+				*sp = (UDATA) ARG(jdouble, d);
```
gacholio: See below.  Also, I think this code may be incorrect. If ARG results in a jdouble, casting it to UDATA will not be correct.

gacholio: `*(jdouble*)sp = ARG(...)`

runtime/vm/jnicsup.cpp
gacholio: Likely a direct translation from builder, which could not deal with values larger than UDATA.

runtime/vm/jnicsup.cpp
gacholio: Yes, if it passes our testing, writing it sensibly is the better option.

pdbain-ibm:
@gacholio I updated the code per your comments and 
https://gcc.gnu.org/bugzilla/show_bug.cgi?id=87289#c8.  Would you kindly re-review?

runtime/vm/jnicsup.cpp
```diff
@@ -468,14 +464,27 @@ UDATA JNICALL   pushArguments(J9VMThread *vmThread, J9Method* method, void *args
...
+				if ('L' == *sigChar++) {
+					/* skip the rest of the signature */
+					while (';' != *sigChar) {
+						sigChar += 1;					
+					}
+				}
+				sp -= 1;
+				objArg = ARG(jobject, l);
+				*sp = (0 == objArg) ? 0: (UDATA) *((j9object_t*) objArg);
+				break;
```
gacholio: I actually prefer the goto in this case (with an additional comment about intentional fall-through when the goto isn't taken).

runtime/vm/jnicsup.cpp
```diff
@@ -468,14 +464,27 @@ UDATA JNICALL   pushArguments(J9VMThread *vmThread, J9Method* method, void *args
...
 				/* skip the rest of the signature */
-				while (*sigChar++ != ';');
-dontSkipClassName:
+				while (';' != *sigChar) {
+					sigChar += 1;
+				}
+				sp -= 1;
 				objArg = ARG(jobject, l);
-				*--sp = objArg ? (UDATA)*(j9object_t*)objArg : 0;
+				*sp = (0 == objArg) ? 0: (UDATA) *((j9object_t*) objArg);
```
gacholio: NULL rather than 0 for the comparison.

runtime/vm/jnicsup.cpp
```diff
@@ -498,31 +507,23 @@ UDATA JNICALL   pushArguments(J9VMThread *vmThread, J9Method* method, void *args
...
+				*((jfloat*) sp) = (jfloat) ARG(jdouble, f);
+			}
+			break;
 			case 'D':
 				/* double type */
-				dbl = ARG(jdouble, d);
-				lngOrDblPtr = (UDATA *) &dbl;
-				goto pushLongOrDouble;
+				sp -= 2;
+				*((jdouble*) sp) = (jdouble) ARG(jdouble, d);
```
gacholio: No need to cast ARG.

runtime/vm/jnicsup.cpp
```diff
@@ -498,31 +507,23 @@ UDATA JNICALL   pushArguments(J9VMThread *vmThread, J9Method* method, void *args
...
-pushLongOrDouble:
-#ifdef J9VM_ENV_DATA64
-				--sp;
-				*--sp = *(lngOrDblPtr);
-#else
-				*--sp = *(lngOrDblPtr + 1);
-				*--sp = *(lngOrDblPtr);
-#endif
+				sp -= 2;
+				*((jlong*) sp) = (jlong) ARG(jlong, j);
```
gacholio: No need to cast ARG.

runtime/vm/jnicsup.cpp
```diff
@@ -468,14 +464,27 @@ UDATA JNICALL   pushArguments(J9VMThread *vmThread, J9Method* method, void *args
 		switch (*sigChar++) {
 			case '[':
 				/* skip the rest of the signature */
-				while (*sigChar == '[') sigChar++;
-				if (*sigChar++ != 'L') goto dontSkipClassName;
+				while (*sigChar == '[') {
+					sigChar++;
```
gacholio: += 1

runtime/vm/jnicsup.cpp
pdbain-ibm: Done. I can remove the cast from 'B', 'Z', et al. and fix the `--sp` too, if you wish.

pdbain-ibm:
Fixed per above comments.

runtime/vm/jnicsup.cpp
gacholio: I believe the casts for smaller-than-int cases are necessary (unless we completely trust the compiler/varargs implementation). Safer to ensure that we are properly truncating the int value and then sign/zero extending it before copying to the stack.

runtime/vm/jnicsup.cpp
```diff
@@ -498,31 +501,23 @@ UDATA JNICALL   pushArguments(J9VMThread *vmThread, J9Method* method, void *args
...
-			case 'F':
+			case 'F': {
 				/* float type */
-				fltPtr = (jfloat*)--sp;
-				/* The Linux SH4 compiler needs the next two lines to be two steps.  If you combine them it fails to compile */
-				dbl = ARG(jdouble, f);
-				*fltPtr = (jfloat)dbl;
-				break;
+				sp -= 1;
+				/* jfloat {aka float}' is promoted to 'double' when passed through '...' */
```
gacholio: last nitpick - unmatched (and unncessary quotes) here - maybe leave the ones around ..., but the others could go.

runtime/vm/jnicsup.cpp
```diff
@@ -498,31 +501,23 @@ UDATA JNICALL   pushArguments(J9VMThread *vmThread, J9Method* method, void *args
 				*(I_32*)(--sp) = (I_32)(jint)ARG(jint, i);	/* NOTE: jint may be larger than int, but jchar, jshort, etc, must be no larger than int */
 				break;
 #ifdef J9VM_INTERP_FLOAT_SUPPORT
-			case 'F':
+			case 'F': {
```
gacholio: scope not needed

runtime/vm/jnicsup.cpp
```diff
@@ -498,31 +501,23 @@ UDATA JNICALL   pushArguments(J9VMThread *vmThread, J9Method* method, void *args
...
 				/* float type */
-				fltPtr = (jfloat*)--sp;
-				/* The Linux SH4 compiler needs the next two lines to be two steps.  If you combine them it fails to compile */
-				dbl = ARG(jdouble, f);
-				*fltPtr = (jfloat)dbl;
-				break;
+				sp -= 1;
+				/* jfloat {aka float}' is promoted to 'double' when passed through '...' */
+				*((jfloat*) sp) = ARG(jdouble, f);
```
gacholio: My mistake = this does need a jfloat cast for the varargs case (which means always cast here).

runtime/vm/jnicsup.cpp
gacholio: could use regular brackets instead of braces here - this is a regular sentence.

runtime/vm/jnicsup.cpp
yongqinglu: Hi, I'm an user of J9 and encountered one issue on jdk80 64bit sr6fp35 occurred in routine pushArguments()  in jnicsup.cpp and seems to be related to the updated line of code here. would you kindly help to review?
The issue was:
When I set  "_export MALLOCDEBUG=catch_overflow,align:0_" to enable memory catch-overflow and a simple Hello java program would crash on the line:  _*sp = (NULL == objArg) ? 0: (UDATA) *((j9object_t*) objArg);_

Several findings by lots of tests：
1, besides my using sr6fp35, many other older and newer versions on jdk8064bit had the issue as well.
2, 32bit java did not have the issue.
3, when option _align_ in MALLOCDEBUG was changed to be 8 or 16 or bigger, the issue did not occur.

I tried to understand the line of code but did not find out obvious overflow. So request your help. Thanks.

pdbain: @yongqinglu I no longer maintain this code.
If you think there is an problem, please open a github issue.
Thank you
             -p

pshipton: Using `align:0`, which really means `align:1`, is invalid / incompatible with OpenJ9 and expected to cause problems. OpenJ9 64-bit expects the allocated memory to be aligned as per the default 2 word boundary. Changing the default to a smaller alignment will cause undefined results.

yongqinglu: Thank you all. Per my understanding of your input, it worked as designed and was not an issue. Thanks. @pshipton

PR #2854, opened 2018-09-13
Clean up java.lang.Access

pdbain-ibm:
Remove incorrect comments and annotations about obsolete methods.
Implement invalidatePackageAccessCache().

Fixes https://github.com/eclipse/openj9/issues/2234.


pshipton:
The PR and commit title is a little misleading, please change it to something like "Implement java.lang.Access.invalidatePackageAccessCache() and cleanup".

jcl/src/java.base/share/classes/java/lang/Access.java
```diff
@@ -232,8 +232,6 @@ public ServicesCatalog createOrGetServicesCatalog(ClassLoader classLoader) {
 		return classLoader.createOrGetServicesCatalog();
 	}
 	
-	/* removed in build 160 */
-	@Deprecated
 	public ServicesCatalog getServicesCatalog(ClassLoader classLoader) {
```
pshipton: This this API is actually unused in the latest? As only getServicesCatalog(ModuleLayer) is used.

pdbain-ibm:
Title changed and getServicesCatalog(ClassLoader classLoader) removed.

pdbain-ibm:
getServicesCatalog() isn't used in Java 8, 10, or 11.  It might be used in Java 9 B148, so I will JPP it out.

pdbain-ibm:
Commit comment updated, preprocessor directive added.

pshipton:
jenkins test sanity zlinux jdk10

PR #2855, opened 2018-09-13
CopyForwardHybrid

LinHu2016:
CopyForwardHybrid is hybrid mode of CopyForward Scheme that 
 GC can run both copyforward and markcompact on the collection set
 simultaneously (for balanced PGC only). During precollection, 
 GC can reserve some regions for markcompact only(non evacuated 
 regions -- such as jni critical regions) and copyforward for 
 the rest of collection set. CopyForwardHybrid intent to avoid 
 markcompact PGC due to nonEvacuated regions in collection set,
 in order to minimize the longer gc pause time and affection to
 concurrent gc.

 - new java Option -XXgc:tarokEnableCopyForwardMarkCompactHybrid for enabling
 CopyForwardHybrid mode, default is disabled.
 - new java Option -XXgc:fvtest_forceCopyForwardMarkCompactHybridRatio=nn,
 (1 <=nn<= 100) nn is the percentage of collection set to be reserved
for markcompact(for margin testing CopyForwardHybrid).
 - report copyforward is running hybrid mode in verbose gc
 - allow jni critical eden regions in nursery collectionSet for
copyforward
  - _noEvacuation flag in HeapRegionDescriptorVLHGC, set it true 
  if the regions can not be evacuated. decide which region can not 
  be evacuated in preProcessRegions.
  - pull scan work from both scanCache and workPacket in CompleteScan
  - synchronize working threads with both scanCache and workPacket
  - if live object in non evacurated regions mark map instead of
copyforward.
  - exception case 1: if abort case happens, stop to pull scan work
  from workPacket
  - exception case 2: if workPacket overflow during CompleteScan
  set it to abort case to handle overflow
 
depends: eclipse/omr#2956

### DIFF
--- a/runtime/gc_vlhgc/CopyForwardScheme.cpp
+++ b/runtime/gc_vlhgc/CopyForwardScheme.cpp
@@ -308,12 +308,15 @@ MM_CopyForwardScheme::MM_CopyForwardScheme(MM_EnvironmentVLHGC *env, MM_HeapRegi
 	, _scanCacheListSize(_extensions->_numaManager.getMaximumNodeNumber() + 1)
 	, _scanCacheWaitCount(0)
 	, _scanCacheMonitor(NULL)
+	, _workQueueWaitCountPtr(&_scanCacheWaitCount)
+	, _workQueueMonitorPtr(&_scanCacheMonitor)
 	, _doneIndex(0)
 	, _markMap(NULL)
 	, _heapBase(NULL)
 	, _heapTop(NULL)
 	, _abortFlag(false)
 	, _abortInProgress(false)
+	, _regionCountCannotBeEvacuated(0)
 	, _cacheLineAlignment(0)
 	, _clearableProcessingStarted(false)
 #if defined(J9VM_GC_DYNAMIC_CLASS_UNLOADING)
@@ -520,16 +523,16 @@ MM_CopyForwardScheme::raiseAbortFlag(MM_EnvironmentVLHGC *env)
 {
 	if (!_abortFlag) {
 		bool didSetFlag = false;
-		omrthread_monitor_enter(_scanCacheMonitor);
+		omrthread_monitor_enter(*_workQueueMonitorPtr);
 		if (!_abortFlag) {
 			_abortFlag = true;
 			didSetFlag = true;
 			/* if any threads are waiting, notify them so that they can get out of the monitor since nobody else is going to push work for them */
-			if (0 != _scanCacheWaitCount) {
-				omrthread_monitor_notify_all(_scanCacheMonitor);
+			if (0 != *_workQueueWaitCountPtr) {
+				omrthread_monitor_notify_all(*_workQueueMonitorPtr);
 			}
 		}
-		omrthread_monitor_exit(_scanCacheMonitor);
+		omrthread_monitor_exit(*_workQueueMonitorPtr);
 
 		if (didSetFlag) {
 			env->_copyForwardStats._aborted = true;
@@ -598,12 +601,15 @@ MM_CopyForwardScheme::preProcessRegions(MM_EnvironmentVLHGC *env)
 
 	UDATA ownableSynchronizerCandidates = 0;
 	UDATA ownableSynchronizerCountInEden = 0;
-	
+
+	_regionCountCannotBeEvacuated = 0;
+
 	while(NULL != (region = regionIterator.nextRegion())) {
 		region->_copyForwardData._survivorBase = NULL;
 
 		if(region->containsObjects()) {
 			region->_copyForwardData._initialLiveSet = true;
+			region->_copyForwardData._evacuateSet = region->_markData._shouldMark;
 			if (region->_markData._shouldMark) {
 				region->getUnfinalizedObjectList()->startUnfinalizedProcessing();
 				ownableSynchronizerCandidates += region->getOwnableSynchronizerObjectList()->getObjectCount();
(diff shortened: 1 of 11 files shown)


runtime/gc_vlhgc/CopyForwardScheme.cpp
```diff
@@ -3269,31 +3323,167 @@ MM_CopyForwardScheme::completeScanForAbort(MM_EnvironmentVLHGC *env)
...
+					ret = getNextWorkUnitOnNode(env, nextNode);
+				}
+				nextNode = (nextNode + 1) % nodeLists;
+			}
+		}
+		if (SCAN_REASON_NONE == ret && (0 != _regionCountCannotBeEvacuated) && !_abortInProgress && !abortFlagRaised()) {
+			if (env->_workStack.retrieveInputPacket(env)) {
+				ret = SCAN_REASON_PACKET;
+			}
+		}
```
amicic: for code cleanness, lines from 3371 to 3391 should go into method called something like getNextWorkUnitNoWait(), that does not take any NUMA node param.

for now, the order we consume things  is ok (prefer copy queue, because realistically more regions are to be evacuated), but in future changes, we have to be fair to both copy and mark work Q. we should somehow interleave. something like: pick from whichever is currently more likely to overflow.

runtime/gc_vlhgc/CopyForwardScheme.hpp
```diff
@@ -118,9 +111,13 @@ class MM_CopyForwardScheme : public MM_BaseNonVirtual
...
 	MM_CopyScanCacheListVLHGC *_cacheScanLists;  /**< An array of per-node caches which contains objects still to be scanned (1+node_count elements in array)*/
 	UDATA _scanCacheListSize;	/**< The number of entries in _cacheScanLists */
-	UDATA _scanCacheWaitCount;	/**< The number of threads currently sleeping on _scanCacheMonitor, awaiting scan cache work */
+	volatile UDATA _scanCacheWaitCount;	/**< The number of threads currently sleeping on _scanCacheMonitor, awaiting scan cache work */
 	omrthread_monitor_t _scanCacheMonitor;	/**< Used when waiting on work on any of the _cacheScanLists */
-	volatile UDATA _doneIndex;	/**< Incremented when _cacheScanLists are empty and we want all threads to fall out of _scanCacheMonitor */
+
+	volatile UDATA* _scanCacheWaitCountPtr;	/**< The number of threads currently sleeping on *_scanCacheMonitorPtr, awaiting scan cache work or work from packets*/
+	omrthread_monitor_t* _scanCacheMonitorPtr;	/**< Used when waiting on work on any of the _cacheScanLists or workPackets*/
+
```
amicic: for the two pointers, since they would be used for both Scan Caches and WorkPackets, i'd give them more generic names _workQueue.... instead of _scanCache....

runtime/gc_vlhgc/CopyForwardScheme.cpp
```diff
@@ -2718,22 +2756,24 @@ MM_CopyForwardScheme::getNextScanCache(MM_EnvironmentVLHGC *env, UDATA preferred
 		}
 
 		if (NULL == cache) {
-			omrthread_monitor_enter(_scanCacheMonitor);
-			_scanCacheWaitCount += 1;
+			omrthread_monitor_enter(*_scanCacheMonitorPtr);
```
amicic: is getNextScanCache() still used? same for getNextScanCacheOnNode.

I hope to remove them, if  not now, than at least when we prove the hybrid scheme is stable.

runtime/gc_vlhgc/CopyForwardScheme.cpp
LinHu2016: done

runtime/gc_vlhgc/CopyForwardScheme.hpp
LinHu2016: done

runtime/gc_vlhgc/CopyForwardScheme.cpp
```diff
@@ -2718,22 +2756,24 @@ MM_CopyForwardScheme::getNextScanCache(MM_EnvironmentVLHGC *env, UDATA preferred
 		}
 
 		if (NULL == cache) {
-			omrthread_monitor_enter(_scanCacheMonitor);
-			_scanCacheWaitCount += 1;
+			omrthread_monitor_enter(*_workQueueMonitorPtr);
+			*_workQueueWaitCountPtr += 1;
```
amicic: It would be cleaner if getNextScanCache() and similar scanCache specific methods use directly scanCache monitor and waitCount directly. I guess, they would be used only in non-hybrid mode. 
But I hope we will get rid of those scanCache specific methods, eventually.

runtime/gc_vlhgc/CopyForwardScheme.cpp
LinHu2016: removed

runtime/gc_vlhgc/CopyForwardScheme.cpp
LinHu2016: I removed method getNextScanCache() and getNextScanCacheOnNode().

runtime/gc_vlhgc/CopyForwardScheme.cpp
```diff
@@ -3269,31 +3227,174 @@ MM_CopyForwardScheme::completeScanForAbort(MM_EnvironmentVLHGC *env)
...
+		if (env->_workStack.retrieveInputPacket(env)) {
+			ret = SCAN_REASON_PACKET;
+		}
+	}
+	return ret;
+}
+
+
+MM_CopyForwardScheme::CopyForwardSchemeScanReason
+MM_CopyForwardScheme::getNextWorkUnit(MM_EnvironmentVLHGC *env, UDATA preferredNumaNode)
```
amicic: try to move this up closer to where getNextScanCache was, to help diff tool

LinHu2016: done

amicic:

dmitripivkine:

dmitripivkine:
known problem (OOM), restarting again

dmitripivkine:

dmitripivkine:

AdamBrousseau:
Jenkins copyright check

AdamBrousseau:
Jenkins line endings check

dmitripivkine:

dmitripivkine:

amicic:
This PR was causing failures when running with debug -XXgc:tarokEnableExpensiveAssertions option that none of Travis or Jenkins test uses. It's reverted till understood if the failure is real or just the debug option is broken.

PR #2857, opened 2018-09-13
Remove use of deprecated API

keithc-ca:
Boolean constructors are deprecated in Java 9+.


DanHeidinga:
Given this is a preprocessor change, a single compile should be sufficient to validate it

DanHeidinga:
Jenkins compile xlinux jdk8

PR #2858, opened 2018-09-13
OpenJ9 fixes for OSX support [Part 1]

babsingh:
**1) Satisfy runtime linker on OSX for `__cxa_pure_virtual's` definition**

On OSX, linker fails to find the definition of `__cxa_pure_virtual`. Existing stub for `__cxa_pure_virtual` is enabled on OSX to satisfy the runtime linker. The stub for `__cxa_pure_virtual` should never be called. 

**2) Fix "Abort trap: 6" on OSX**

On OSX, **Abort trap: 6** happens while copying a name into `J9Statistic->name` using strcpy. Reworked memory allocation of `J9Statistic` in order to prevent **Abort trap: 6**.

**3) private field `_tokenCount` unused in BarrierSynchronization**

While compiling OpenJ9 with Clang 9.1.0 on OSX, the following error is seen:

`./BarrierSynchronization.hpp:48:7: error: private field '_tokenCount' is not used [-Werror,-Wunused-private-field]`

private field `_tokenCount` is removed since it is unused.

**4) Real-time clock is not supported on OSX**

In MetronomeAlarm.hpp, private field `RTCfd` is used to store real-time clock related information.

On OSX, real-time clock is not supported. So, private field `RTCfd` is unused. While compiling OpenJ9 with Clang 9.1.0 on OSX, the following error is seen:

`./MetronomeAlarm.hpp:103:8: error: private field 'RTCfd' is not used
[-Werror,-Wunused-private-field]`

The declaration of private field `RTCfd` is surrounded by `#if defined(LINUX) && !defined(J9ZTPF)`. This resolves the compiler error on OSX.

Also, fixed ifdefs related to `RTCfd` in `MM_RTCAlarm::sleep()`.

### DIFF
--- a/runtime/gc_realtime/MetronomeAlarm.hpp
+++ b/runtime/gc_realtime/MetronomeAlarm.hpp
@@ -1,6 +1,5 @@
-
 /*******************************************************************************
- * Copyright (c) 1991, 2017 IBM Corp. and others
+ * Copyright (c) 1991, 2018 IBM Corp. and others
  *
  * This program and the accompanying materials are made available under
  * the terms of the Eclipse Public License 2.0 which accompanies this
@@ -99,8 +98,10 @@ public:
 
 class MM_RTCAlarm : public MM_Alarm
 {
-private:	
+private:
+#if defined(LINUX) && !defined(J9ZTPF)
 	IDATA RTCfd;
+#endif /* defined(LINUX) && !defined(J9ZTPF) */
 
 public:
 	static MM_RTCAlarm * newInstance(MM_EnvironmentBase *env, MM_OSInterface* osInterface);
--- a/runtime/vm/statistics.c
+++ b/runtime/vm/statistics.c
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 1991, 2016 IBM Corp. and others
+ * Copyright (c) 1991, 2018 IBM Corp. and others
  *
  * This program and the accompanying materials are made available under
  * the terms of the Eclipse Public License 2.0 which accompanies this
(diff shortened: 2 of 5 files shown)


babsingh:
Reviewers:
- @DanHeidinga for VM changes
- @dmitripivkine for GC changes

runtime/gc_realtime/MetronomeAlarm.hpp
```diff
@@ -99,8 +98,10 @@ class MM_HRTAlarm : public MM_Alarm
 
 class MM_RTCAlarm : public MM_Alarm
 {
-private:	
+private:
+#if defined(LINUX) && !defined(J9ZTPF)
 	IDATA RTCfd;
+#endif /* defined(LINUX) && !defined(J9ZTPF) */
```
dmitripivkine: Would you please also add "&& !defined(J9ZTPF)" to MM_RTCAlarm::sleep() in MetronomeAlarm.cpp for consistency ?

babsingh: fixed.

runtime/vm/statistics.c
```diff
@@ -25,23 +25,25 @@
...
-	statistic = j9mem_allocate_memory(size, OMRMEM_CATEGORY_VM);
-	if (statistic != NULL) {
+	statistic = j9mem_allocate_memory(sizeof(J9Statistic), OMRMEM_CATEGORY_VM);
+	if (NULL != statistic) {
+		IDATA stringSize = strlen((char*)name) + 1;
 		statistic->dataSlot = 0;
 		statistic->dataType = (U_32) dataType;
-		strcpy((char*)&(statistic->name), (char*)name);
+
+		statistic->name = j9mem_allocate_memory(stringSize, OMRMEM_CATEGORY_VM);
```
DanHeidinga: The statistics kept the full data in a single allocation.  This doubles the allocations for a statistic.  It looks like the right amount of memory has been allocated originally....

Any insight into why this is aborting?  Are we missing a "no-strict-aliasing" style compile option here?

babsingh: i think the `&` may have been causing issues. i reverted to the old implementation. but, i changed how the size of `J9Statistic` is evaluated and removed the `&` during `strcpy`. now, the code works on OSX. remaining changes are done to satisfy coding standards.

DanHeidinga:

babsingh:
Related to #36.

PR #2862, opened 2018-09-13
Set cache dir to user's home directory

hangshao0:
1. Set cache dir to user's home directory if groupAccess does not
presents on Java 11
2. Add options bootClassesOnly and fatal
3. Set J9SHR_RUNTIMEFLAG_ENABLE_CACHE_NON_BOOT_CLASSES by default,
remove this flag if option bootClassesOnly presents
4. Change the second parameter of j9shmem_getDir().
5. Change the first parameter of getCacheDir().

Doc change eclipse/openj9-docs#103.


pshipton:
For consistency, I expect the shared cache location should always default to the home directory. If we can agree on this, we should also discuss a migration strategy since existing VMs use the temp directory.

hangshao0:
> I expect the shared cache location should always default to the home directory. 

The home directory might not be accessible to other users. This might break users using "groupAccess". Probably we can use home directory if "groupAccess" is not in use.

hangshao0:
@pshipton, Do you agree to set cache directory to user's home only when "groupAccess" is not in use ?

pshipton:
> Do you agree to set cache directory to user's home only when "groupAccess" is not in use ?

Yes, for compatibility it seems necessary.

I think what we need is a list of scenarios and what will occur in each. This will help ensure everything is covered properly, and the documentation is updated appropriately.

Here is a starter list.
- what happens when there is an existing cache in /tmp and you get a new JVM which puts the cache in the home directory? How do caches in /tmp get cleaned up?
- do bootstrap caches and caches created via -Xshareclasses use the same file name?
- how do you delete a bootstrap cache?
- what happens if one JVM creates and uses a bootstrap cache, and another JVM enables -Xshareclasses using the default name?

hangshao0:
>- what happens when there is an existing cache in /tmp and you get a new JVM which puts the cache in the home directory? How do caches in /tmp get cleaned up?

Then the new JVM will use the cache in the home directory. Users can always delete the caches in /tmp using "cacheDir=/tmp,destroy".

>- do bootstrap caches and caches created via -Xshareclasses use the same file name?

Currently yes. 

> - how do you delete a bootstrap cache?

We can use -Xshareclasses:destroy.

> - what happens if one JVM creates and uses a bootstrap cache, and another JVM enables -Xshareclasses using the default name?

They will share the same cache.

pshipton:
> They will share the same cache.

Will JVMs which haven't enabled -Xshareclasses load non-bootstrap classes from the cache?

- what is the default size of a bootstrap cache?

pshipton:
- how do you change the default location a bootstrap cache, or apply other options
- seems nonfatal or the equivalent should automatically be in effect for bootstrap caches, how does this interact with the above
- how do you turn off the bootstrap cache
- I assume changing the location of the cache affects the location of snapshots

hangshao0:
>Will JVMs which haven't enabled -Xshareclasses load non-bootstrap classes from the cache?

No

>what is the default size of a bootstrap cache?

It is the same as the default cache size. 300MB with 64softmx on 64-bit platforms and 16MB on 32 bit platforms.

>how do you change the default location of a bootstrap cache, or apply other options

By changing the setting of user home, or environment variable "HOME". (There is another way to do this , by using -Xshareclasses:CacheDir=/dir/,bootClassesOnly.)

>how do you turn off the bootstrap cache

By using -Xshareclasses:none

>seems nonfatal or the equivalent should automatically be in effect for bootstrap caches, how does this interact with the above

The default class sharing is equivalent to -Xshareclasses:cacheDir=userhome,nonfatal,bootClassesOnly. Nonfatal can be turn on by using -Xshareclasses:cacheDir=userhome,bootClassesOnly.

>I assume changing the location of the cache affects the location of snapshots

Yes.

There will be a trace point if the default shared cache failed to start up. And the size of the default shared cache can be set by -XX:SharedCacheHardLimit= and -Xscmc

hangshao0:
Discussed with Peter, I will try on some real applications with -Xshareclasses:bootClassesOnly to see if the bootstrap shared cache gradually grows in size.

pshipton:
I suggest bootClassesOnly implies nonfatal. A new "fatal" option (although I don't like this name) can be added so -Xshareclasses:bootClassesOnly is equivalent to the default, and -Xshareclasses:bootClassesOnly,fatal can disable it if desired.

pshipton:
Another point we discussed
- what happens if the shared cache cannot be created in the home directory, due to it being on a network drive, or home doesn't exist, or there isn't enough disk space.

Default bootClassesOnly is disabled.

hangshao0:
The change is ready to be reviewed again.

pshipton:
Please open the doc issue, this change cannot be merged without the documentation being ready to go as well.

runtime/port/sysvipc/j9shmem.c
```diff
@@ -1242,30 +1243,90 @@ getControlFilePath(struct J9PortLibrary* portLibrary, const char* cacheDirName,
...
 	OMRPORT_ACCESS_FROM_J9PORT(portLibrary);
-	const char* rootDir = J9SH_DEFAULT_CTRL_ROOT;
+	const char* rootDir = NULL;
+	char homeDirBuf[J9SH_MAXPATH];
+	const char* homeDir = NULL;
+	BOOLEAN appendBaseDir = J9_ARE_ALL_BITS_SET(flags, J9SHMEM_GETDIR_APPEND_BASEDIR);
 
 	Trc_PRT_j9shmem_getDir_Entry();
 
+	memset(homeDirBuf, 0, J9SH_MAXPATH);
```
pshipton: to avoid future problems, the size is better expressed as sizeof(homeDirBuf)

runtime/port/sysvipc/j9shmem.c
```diff
@@ -1242,30 +1243,90 @@ getControlFilePath(struct J9PortLibrary* portLibrary, const char* cacheDirName,
...
-	if (appendBaseDir) {
-		if (omrstr_printf(buffer, bufLength, "%s/%s", rootDir, J9SH_BASEDIR) == bufLength - 1) {
-			Trc_PRT_j9shmem_getDir_ExitFailedOverflow();
-			return -1;
+	} else {
+		if (J9_ARE_ALL_BITS_SET(flags, J9SHMEM_GETDIR_USE_USERHOME)) {
+			Assert_PRT_true(TRUE == appendBaseDir);
+			uintptr_t baseDirLen = strlen(J9SH_BASEDIR);
+			/* From https://linux.die.net/man/3/getpwuid:
+			 * "An application that wants to determine its user's home directory should inspect the value of HOME (rather than the value getpwuid(getuid())->pw_dir)
```
pshipton: Please don't copy the text from the Linux man pages. You can paraphrase.

hangshao0:
OpenJ9 doc issue created: https://github.com/eclipse/openj9-docs/issues/103.

We never documented the debug options from -Xshareclasses:morehelp, I guess we do not need to document -Xshareclasses:bootClassesOnly and -Xshareclasses:fatal either ? (Though "bootClassesOnly" and "fatal" are in the doc change request for now)

runtime/port/sysvipc/j9shmem.c
hangshao0: Fixed

pshipton:
bootClassesOnly and fatal should be first class options shown via -Xshareclasses:help

hangshao0:
Moved bootClassesOnly and fatal to -Xshareclasses:help

pshipton:
jenkins test all plinux jdk11

pshipton:

pshipton:
jenkins test extended plinux jdk11

pshipton:

pshipton:
jenkins compile win,zlinux jdk8

pshipton:
jenkins compile aix jdk8

pshipton:
The compile failures are a result of https://github.com/eclipse/openj9/pull/3158 and can be ignored.

pshipton:
@SueChaplain I've added a review on this for you so you can indicate when the doc changes are ready to go.

pshipton:
jenkins compile aix,zlinux jdk8

pshipton:
jenkins compile xlinux,plinux jdk8

pshipton:
jenkins compile xlinux,plinux jdk8

pshipton:
jenkins compile plinux jdk8

pshipton:
jenkins compile plinux jdk8

pshipton:

PR #2806, opened 2018-09-07
IBM Z Codegen changes to support Condy in jdk11

r30shah:
This commit contains changes to suppport Constant Dynamic
in JIT Compiled code on Z platform.

Signed-off-by: Rahil Shah <rahil@ca.ibm.com>

PR #2863, opened 2018-09-13
Improve condy PicBuilder sequence

fjeremic:
As suggested in [1] we can use a better instruction sequence in our
condy resolution function.

[1] https://github.com/eclipse/openj9/pull/2806#issuecomment-421013666

### DIFF
--- a/runtime/compiler/z/runtime/PicBuilder.m4
+++ b/runtime/compiler/z/runtime/PicBuilder.m4
@@ -1820,8 +1820,7 @@ ZZ into the Literal Pool
     L_GPR r14,eq_codeRA_inDataSnippet(,r14) # Get mainline RA
 
 ZZ Branch instruction in mainline will be patched here with NOP
-    LHI r1,HEX(04)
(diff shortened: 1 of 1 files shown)


fjeremic:
@joransiu @r30shah could I please get a review?

fjeremic:
As an aside there is no doubt a lot of the instruction sequences in PicBuilder can be improved greatly. i.e. similar patching logic happens for normal field resolution and we use an `LHI` sequence there. I think we're due for an overhaul. I'd actually prefer if we started getting rid of some of these PicBuilder sequences. Quite a few of them don't need to be assembler written and if we had to we can use inline asm (once we upgrade compiler levels).

r30shah:
LGTM. I agree with @fjeremic . I think even in the sequences we needed, so many instances where we can reduce number of instructions by choosing better ones. Like we did for condy.

joransiu:

fjeremic:
Same failure here as observed in #2806. I think we're good here @joransiu.

Issue #2418, opened 2018-07-18
Whither x86 JIT assembly files?


0xdaryl:
The OpenJ9 x86 JIT uses a number of hand-written assembly files as part of its runtime.  The choice of assembly allows very precise control of the instructions, registers, and stack in some contexts, as well as customized linkages between sections of code.  There are about 7000 lines of x86 assembly in the JIT and are written using MASM syntax (Microsoft Macro Assembler).  This creates a challenge when building on Linux using the GNU assembler (gas) because it consumes a different syntax altogether.

Rather than maintain two independent but logically similar sets of source files that require dual-maintenance, a tool (`masm2gas.pl`) was written to convert MASM syntax files into AT&T syntax just-in-time during the build that can be consumed by gas.  The decision behind writing a tool rather than using a common assembler on both platforms is historical.  Many of these files were created during the first x86-64 port 15 years ago and at that time there weren't any assemblers mature enough that worked across operating systems that supported all the features the project needed (x86-64 support, macros, preprocessor directives, etc.).  While a tooling solution was able to reduce the dual maintenance of the code, it is a fragile and incomplete solution that needs updates as new assembly is written that uses MASM syntax variants that weren't handled previously.  Furthermore, it is incapable of handling conditional assembly, and to get around that some of the assembly files are actually run through the C preprocessor first (the files have a `.pasm` extension).

With the work underway to port OpenJ9 to macOS these assembly files will need to be built as well.  Unfortunately, the LLVM assembler (llvm-mc) while similar to gas in terms of syntax and command-line options, does have some syntax differences that need to be dealt with.  Rather than simply modifying (hacking) the translation tool to handle LLVM assembly, I think this is a good opportunity to take stock of where we are and evaluate what the long term strategy should be with these files.

Also, for whatever is decided here, I would like to give that strong consideration for the x86 assembly solution in Eclipse OMR.  At present, there are no x86 assembly files but if they were to be introduced then an assembly solution is required.

Here are some options to consider:

1.  *"masm2gas.pl Forever, Dude!"*

[...]

PR #2864, opened 2018-09-13
[WIP] [OSX and Linux on x86] Initial PR for migrating JIT assembly files to NASM 

nbhuiyan:
### Background

Currently, x86 JIT assembly files are handled differently on the 2 supported OSes, neither of which could be easily adapted for Mac OSX. In order to support OpenJ9 on Mac OSX (see #36 ) , it was time we reworked how we handled the x86 assembly files in the JIT. Windows builds consume the MASM files directly, whereas Linux on x86 builds first convert the MASM files to GNU-assembly format using the masm2gas script, and then build them. Some of the assembly files first go through the C preprocessor before being consumed by the masm2gas script, adding to the difficulty with connecting errors to the MASM source line. This rewrite of the X86 JIT assembly files aims to simplify the handling of the assembly files across x86, while making it possible to build the OpenJ9 JIT on OSX. More discussions regarding this occurred in issue #2418 

### How to build them

This PR does not contain CMake-specific changes, so you will have to use autotools. In order to allow Linux builds to make use of the original assembly files, the newly introduced files do not get built by default. In order to build the nasm files, you will have to set the following environment variable:
```sh
export USE_NASM=yes 
```

I've only used NASM versions 2.13.03 and 2.11.08.

### WIP because...
* While building and linking succeeds on Linux (along with `java -version`), there are a number of linker errors on Mac OSX that I am currently working on
* Incomplete 32-bit host support. This was low on the priority list due to being irrelevant to Mac OSX. Currently working on a solution for using the correct register names on 32bit hosts. This was originally handled through the use of x86RegisterMap.inc
* Currently working on auto-generating jilconsts.inc in NASM format. Currently, it is output in MASM format. A replacement exists in this PR, but that is definitely only temporary.
* Currently working on a solution to remove the need for using the C Preprocessor (for .pnasm files)
* Currently working on examining the disassembled code obtained from ELF objects generated by GNU-as and NASM to verify equivalence
* Need to run tests to verify functionality

### DIFF
--- a/runtime/compiler/x/amd64/runtime/AMD64CompressString_nasm.inc
+++ b/runtime/compiler/x/amd64/runtime/AMD64CompressString_nasm.inc
@@ -1,0 +1,239 @@
+; Copyright (c) 2000, 2018 IBM Corp. and others
+;
+; This program and the accompanying materials are made available under
+; the terms of the Eclipse Public License 2.0 which accompanies this
+; distribution and is available at https://www.eclipse.org/legal/epl-2.0/
+; or the Apache License, Version 2.0 which accompanies this distribution and
+; is available at https://www.apache.org/licenses/LICENSE-2.0.
+;
+; This Source Code may also be made available under the following
+; Secondary Licenses when the conditions for such availability set
+; forth in the Eclipse Public License, v. 2.0 are satisfied: GNU
+; General Public License, version 2 with the GNU Classpath
+; Exception [1] and GNU General Public License, version 2 with the
+; OpenJDK Assembly Exception [2].
+;
+; [1] https://www.gnu.org/software/classpath/license.html
+; [2] http://openjdk.java.net/legal/assembly-exception.html
+;
+; SPDX-License-Identifier: EPL-2.0 OR Apache-2.0 OR GPL-2.0 WITH Classpath-exception-2.0 OR LicenseRef-GPL-2.0 WITH Assembly-exception
+
+    segment .text
+
+                global  _compressString
+                global  _compressStringJ
+                global  _compressStringNoCheck
+		global  _compressStringNoCheckJ
+		global  _andORString
+
+                align 16
+
+
+_compressString: ; PROC 
+		shr  rcx, 4
+		add  rsi, rax
+		add  rsi, rax
+		mov  rdx, 0
+eightchars:
+		mov  rax, QWORD  [rsi]  		; load 4 bytes from the source array into EAX (2 chars)
+		or   rdx, rax				;
+		mov  rbx, rax					; copy the loaded value
+		shr  rbx, 8					; shift right by 8
+		or   rax, rbx					; or the 2 values
+		mov  WORD  [rdi], ax			; write the low 2 bytes of the 4 byte value in the destination
+		shr  rax, 32
+		mov  WORD  [rdi+2], ax			; write the low 2 bytes of the 4 byte value in the destination
+
+		mov  rax, QWORD  [rsi+8]  		; load 4 bytes from the source array into EAX (2 chars)
+		or   rdx, rax				;
+		mov  rbx, rax					; copy the loaded value
+		shr  rbx, 8					; shift right by 8
+		or   rax, rbx					; or the 2 values
+		mov  WORD  [rdi+4], ax			; write the low 2 bytes of the 4 byte value in the destination
+		shr  rax, 32
+		mov  WORD  [rdi+6], ax			; write the low 2 bytes of the 4 byte value in the destination
+
+		mov  rax, QWORD  [rsi+16]  		; load 4 bytes from the source array into EAX (2 chars)
+		or   rdx, rax				;
+		mov  rbx, rax					; copy the loaded value
+		shr  rbx, 8					; shift right by 8
+		or   rax, rbx					; or the 2 values
+		mov  WORD  [rdi+8], ax			; write the low 2 bytes of the 4 byte value in the destination
+		shr  rax, 32
+		mov  WORD  [rdi+10], ax			; write the low 2 bytes of the 4 byte value in the destination
+
+
+		mov  rax, QWORD  [rsi+24]  		; load 4 bytes from the source array into EAX (2 chars)
+		or   rdx, rax				;
+		mov  rbx, rax					; copy the loaded value
+		shr  rbx, 8					; shift right by 8
+		or   rax, rbx					; or the 2 values
+		mov  WORD  [rdi+12], ax			; write the low 2 bytes of the 4 byte value in the destination
+		shr  rax, 32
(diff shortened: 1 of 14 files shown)


runtime/oti/jilconsts_nasm.inc
```diff
@@ -0,0 +1,194 @@
...
+%assign J9TR_VMThread_jitReturnAddress 256
+%assign J9TR_VMThread_returnValue 296
+%assign J9TR_VMThread_returnValue2 304
+%assign J9TR_VMThread_entryLocalStorage 616
+%assign J9TR_VMThread_stackWalkState 608
+%assign J9TR_J9StackWalkState_restartPoint 224
+%assign J9TR_JavaVMJitConfig 6384
+%assign J9TR_JavaVM_runtimeFlags 224
+%assign J9TR_JavaVM_cInterpreter 184
+%assign J9TR_JavaVM_bytecodeLoop 192
```
DanHeidinga: This is very fragile and will break the first time someone modifies one of these structures.  We need to have a story that generates the nasm form of this file in addition to the masm form.

DanHeidinga:
@0xdaryl Are you the appropriate person to review the nasm files?  If we can get the rest of the files reviewed while we sort the jilconsts issue, that will the proceed in parallel

0xdaryl:
Yes, I'll review this.  @nbhuiyan is aware of the jilconsts fragility and is working on a solution while others have the chance to look through the body of work he's got done so far.

runtime/oti/jilconsts_nasm.inc
nbhuiyan: Yup, this is only temporary to get the builds to pass. I am working on auto-generating the NASM variant, using values from the generated jilconsts.inc file.

nbhuiyan:
@0xdaryl Ready for review

nbhuiyan:
@DanHeidinga fyi

DanHeidinga:
@0xdaryl Can you review this?

runtime/compiler/x/amd64/runtime/AMD64CompressString_nasm.inc
```diff
@@ -0,0 +1,235 @@
+; Copyright (c) 2000, 2018 IBM Corp. and others
```
fjeremic: These copyrights seem incorrect for the new files added.

runtime/compiler/x/amd64/runtime/AMD64CompressString_nasm.inc
```diff
@@ -0,0 +1,235 @@
...
+		loop eightcharsNoCheckJ
+                ret
+;_compressStringNoCheckJ endp
+;
+
+
+_compressStringNoCheck: ; PROC 
+		shr  rcx, 4
+		add  rsi, rax
+		add  rsi, rax
```
fjeremic: We use tabs in this file but spaces everywhere else. Is this intended?

runtime/compiler/x/amd64/runtime/AMD64CompressString_nasm.inc
nbhuiyan: I have intentionally kept the first copyright dates according to how their masm-equivalent files had it. This is due to the fact that I've taken basically the same code and have re-written it in a different format in a new file. Eventually, the MASM files will be removed once Linux and Windows support is added.

runtime/compiler/x/amd64/runtime/AMD64CompressString_nasm.inc
nbhuiyan: This tabs vs. space inconsistency is due to the fact that for some of the files I've used a heavily modified version of the masm2gas script to do some initial conversion, while others were done without the help of the script. I am working on the formatting and will push an update while it is being reviewed.

0xdaryl:
Before merging, I think we need to make it clear to everyone what the implications are of merging this because it puts us in a less-than-ideal state for about a month.

This PR is not replacing the existing MASM files just yet.  It introduces a temporary duality where NASM files exist for macOS and MASM files exist for Linux and Windows.  This is obviously not ideal because any changes to JIT assembly files will need to be made in both places.  This is necessary because there are additional changes needed on Linux and Windows to convert the existing makefiles and flags to building with NASM.  The short-term focus has been on building for macOS.

Normally this may be tolerable for these kinds of files, but there has been a flurry of activity in the JIT runtime of late which may be impacted by these changes.

@nbhuiyan feels that he can make the necessary changes to Linux and Windows by end of October, so the period the codebase will be in this state should be relatively short-lived.  I also expect the formatting in the NASM files to be straightened up (some of it was pre-existing, but some was introduced by the auto conversion process).  He will also get rid of the need to run the C preprocessor on the assembly files through modifications to jilconsts.

@andrewcraik @0dvictor and others, please raise concerns if you have them.

andrewcraik:
So I'm happy with this if @0xdaryl and @nbhuiyan take responsibility for updating the NASM files until we cut over to the NASM files on all platforms and eliminate the current assembly files. Could a list of files committers need to watch for in reviews be posted here just for reference and who we should @ to make sure that changes are reflected into the NASM versions?

nbhuiyan:
@andrewcraik Yes, we can ensure that the NASM files remain in sync with any changes to the MASM files during this transition. You can @ me to notify of any changes to the MASM files during the transition period.

Here are the list of files that we will need to keep an eye out for:

In [`runtime/compiler/x/runtime`](https://github.com/eclipse/openj9/tree/master/runtime/compiler/x/runtime)
* X86ArrayTranslate.asm
* X86Codert.asm
* X86EncodeUTF16.asm
* X86LockReservation.asm
* X86PicBuilder.inc
* X86PicBuilder.pasm
* X86Unresolveds.pasm

In [`runtime/compiler/x/amd64/runtime`](https://github.com/eclipse/openj9/tree/master/runtime/compiler/x/amd64/runtime)
* AMD64CompressString.asm
* AMD64CompressString.inc
* AMD64Recompilation.asm

In [`runtime/compiler/x/i386/runtime`](https://github.com/eclipse/openj9/tree/master/runtime/compiler/x/i386/runtime)
* IA32CompressString.inc
* IA32Math64.inc
* IA32Recompilation.asm
* J9IA32CompressString.asm
* J9IA32Math64.asm

0xdaryl:
I'm fine with the changes.  @nbhuiyan will push a new set of commits within the next few hours that includes some formatting changes and a re-synchronization of the PicBuilder assembly.  I'll then relaunch a PR build and merge once that is successful.

0dvictor:
I'm fine with the changes and the temporary solution to keep both versions. I don't expect major changes on the assembly files @nbhuiyan listed after this PR before the end of October.

0xdaryl:

PR #2873, opened 2018-09-14
Use general GC_ClassIterator Iterator to scan class

dmitripivkine:
Replaced code was missed to be updated for recent condy related
modifications. There is no reason to duplicate existing Iterator
implementation - it is exactly the same code.


dmitripivkine:
personal build does not show regression

dmitripivkine:
@tajila FYI problem fixed

PR #2672, opened 2018-08-21
Allow relocation of ClassUnload assumptions

dmitry-ten:
AOT and JITaaS require assumptions to be relocated.
Add a new kind of ExternalRelocation to relocate
ClassUnload assumptions.

Signed-off-by: Dmitry Ten <Dmitry.Ten@ibm.com>

PR #2649, opened 2018-08-17
Add rdbar/rdbari opcodes

cathyzhyi:
Add the most basic supports for read barrier opcodes.

issue: #2812

Signed-off-by: Yi Zhang <yizhang@ca.ibm.com>

PR #2874, opened 2018-09-14
Add TR_J2IVirtualThunkPointer relocation kind

jdmpapin:
This relocates a J2I (JIT to interpreter) virtual thunk pointer along
with a nearby constant pool address and index. The relocation is
necessary because with nestmates, the data consumed by PicBuilder now
contains a J2I thunk pointer, which is used for direct dispatch in case
the callee has not yet been compiled.

No such relocations are created yet. That will be done in each code
generator's call snippet code.

### DIFF
--- a/runtime/compiler/arm/codegen/J9AheadOfTimeCompile.cpp
+++ b/runtime/compiler/arm/codegen/J9AheadOfTimeCompile.cpp
@@ -273,6 +273,22 @@ uint8_t *J9::ARM::AheadOfTimeCompile::initializeAOTRelocationHeader(TR::Iterated
          cursor += SIZEPOINTER;
          break;
          }
+
+      case TR_J2IVirtualThunkPointer:
+         {
+         auto info = (TR_RelocationRecordInformation*)relocation->getTargetAddress();
+
+         *(uintptrj_t *)cursor = (uintptrj_t)info->data2; // inlined site index
+         cursor += SIZEPOINTER;
+
+         *(uintptrj_t *)cursor = (uintptrj_t)info->data1; // constantPool
+         cursor += SIZEPOINTER;
+
+         *(uintptrj_t *)cursor = (uintptrj_t)info->data3; // offset to J2I virtual thunk pointer
+         cursor += SIZEPOINTER;
+         }
+         break;
+
       case TR_CheckMethodEnter:
       case TR_CheckMethodExit:
          {
@@ -742,6 +758,7 @@ uint32_t J9::ARM::AheadOfTimeCompile::_relocationTargetTypeToHeaderSizeMap[TR_Nu
    16,                                       // TR_ArbitraryClassAddress               = 58,
    28,                                        // TR_DebugCounter                        = 59
    4,                                        // TR_ClassUnloadAssumption               = 60
+   16,                                       // TR_J2IVirtualThunkPointer              = 61
    };
 
 
--- a/runtime/compiler/runtime/RelocationRecord.cpp
+++ b/runtime/compiler/runtime/RelocationRecord.cpp
@@ -133,12 +133,10 @@ struct TR_RelocationRecordConstantPoolWithIndexBinaryTemplate : public TR_Reloca
 
 typedef TR_RelocationRecordConstantPoolWithIndexBinaryTemplate TR_RelocationRecordClassObjectBinaryTemplate;
 
-struct TR_RelocationRecordThunksBinaryTemplate : public TR_RelocationRecordConstantPoolBinaryTemplate
+struct TR_RelocationRecordJ2IVirtualThunkPointerBinaryTemplate : public TR_RelocationRecordConstantPoolBinaryTemplate
    {
-   UDATA _thunkAddress;
+   IDATA _offsetToJ2IVirtualThunkPointer;
    };
-
-typedef TR_RelocationRecordThunksBinaryTemplate TR_RelocationRecordThunksJXEBinaryTemplate;
 
 struct TR_RelocationRecordInlinedAllocationBinaryTemplate : public TR_RelocationRecordConstantPoolWithIndexBinaryTemplate
    {
@@ -427,6 +425,9 @@ TR_RelocationRecord::create(TR_RelocationRecord *storage, TR_RelocationRuntime *
          break;
       case TR_Thunks:
          reloRecord = new (storage) TR_RelocationRecordThunks(reloRuntime, record);
+         break;
+      case TR_J2IVirtualThunkPointer:
+         reloRecord = new (storage) TR_RelocationRecordJ2IVirtualThunkPointer(reloRuntime, record);
          break;
       case TR_GlobalValue:
          reloRecord = new (storage) TR_RelocationRecordGlobalValue(reloRuntime, record);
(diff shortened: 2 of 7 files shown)


jdmpapin:
Marked WIP because this depends on eclipse/omr#2965

jdmpapin:
@dsouzai @mstoodle Would you mind reviewing? This goes with #2875 and eclipse/omr#2965

jdmpapin:
@irinarada FYI

runtime/compiler/arm/codegen/J9AheadOfTimeCompile.cpp
```diff
@@ -740,7 +756,9 @@ uint32_t J9::ARM::AheadOfTimeCompile::_relocationTargetTypeToHeaderSizeMap[TR_Nu
    0,                                        // TR_NativeMethodAbsolute                = 56,
    0,                                        // TR_NativeMethodRelative                = 57,
    16,                                       // TR_ArbitraryClassAddress               = 58,
-   28                                        // TR_DebugCounter                        = 59
+   28,                                       // TR_DebugCounter                        = 59
+   0,                                        // TR_ClassUnloadAssumption               = 60
```
mstoodle: is TR_ClassUnloadAssumption supposed to be in this pull request?

mstoodle: I see, you're also co-dependent on #2672 (one of you will conflict with whichever is merged first :) ).

mstoodle:
I have to express some regret that this solution is convoluting an already convoluted thing more than it really should be. I would have preferred to evolve this code in a direction of simplicity.

Let me elaborate a little bit. TR_Thunks is currently "over optimized" in my opinion in that  performs a constant pool relocation as well as the work required to relocate the thunk. On top of that, it reads the cpIndex out of the code cache rather than having cpIndex as an element in the relocation record. We should really evolve TR_Thunks to only relocate the thunk, and create a TR_ConstantPool relocation to relocate the constant pool. Then we would have introduced a new relocation to accomplish the relocation needed here. I've spoken offline with @jdmpapin and I believe we're in agreement on this general approach being a better way forward.

Sadly, that's too much work to get this done in time for our impending 0.10 release. So, I will (reluctantly but pragmatically) review this code as-is even though it's not the generally agreed right way forward.

@jdmpapin can I ask you to open an issue to detail the improved design so that we do not lose track of this work, and I'll review. It would be awesome to see that correction made in time for 0.11, but its timeline isn't all that far away either, so I could be convinced we shouldn't muck with it until after 0.11 goes out the door.

runtime/compiler/arm/codegen/J9AheadOfTimeCompile.cpp
jdmpapin: `TR_ClassUnloadAssumption` is an existing `TR_ExternalRelocationTargetKind` with value 60. It wasn't in the size tables previously because it's not for AOT:

    TR_ClassUnloadAssumption               = 60, // this should not be used in AOT relocations

I need to specify a value there in order to get to index 61 (`TR_J2IVirtualThunkPointer`).

runtime/compiler/runtime/RelocationRecord.cpp
```diff
@@ -1799,6 +1809,33 @@ TR_RelocationRecordThunks::relocateAndRegisterThunk(TR_RelocationRuntime *reloRu
...
+   {
+   return sizeof(TR_RelocationRecordJ2IVirtualThunkPointerBinaryTemplate);
+   }
+
+void
+TR_RelocationRecordJ2IVirtualThunkPointer::relocateJ2IVirtualThunkPointer(
+   TR_RelocationTarget *reloTarget,
+   uint8_t *reloLocation,
+   void *thunk)
+   {
```
mstoodle: I'd stick an assert here that `thunk != NULL` just for completeness

runtime/compiler/runtime/RelocationRecord.cpp
```diff
@@ -1799,6 +1809,33 @@ TR_RelocationRecordThunks::relocateAndRegisterThunk(TR_RelocationRuntime *reloRu
...
+   // For uniformity with TR_Thunks, the reloLocation is not the location of the
+   // J2I thunk pointer, but rather the location of the constant pool address.
+   // Find the J2I thunk pointer relative to that.
+   auto recordData = (TR_RelocationRecordJ2IVirtualThunkPointerBinaryTemplate *)_record;
+   reloLocation += recordData->_offsetToJ2IVirtualThunkPointer;
```
mstoodle: I think you're missing the `preparePrivateData()` code here....nothing sets _offsetToJ2IVirtualThunkPointer from the binary relocation data, does it? Or did I miss it?

jdmpapin: Doesn't `_record` point into the binary relocation data? I do get the correct offset (when I have run with the further change in #2875)

runtime/compiler/arm/codegen/J9AheadOfTimeCompile.cpp
```diff
@@ -273,6 +273,22 @@ uint8_t *J9::ARM::AheadOfTimeCompile::initializeAOTRelocationHeader(TR::Iterated
...
+         {
+         auto info = (TR_RelocationRecordInformation*)relocation->getTargetAddress();
+
+         *(uintptrj_t *)cursor = (uintptrj_t)info->data2; // inlined site index
+         cursor += SIZEPOINTER;
+
+         *(uintptrj_t *)cursor = (uintptrj_t)info->data1; // constantPool
+         cursor += SIZEPOINTER;
+
+         *(uintptrj_t *)cursor = (uintptrj_t)info->data3; // offset to J2I virtual thunk pointer
```
fjeremic: Ah, the good ol' `data1`, `data2`, `data3` field names. I definitely agree with @mstoodle we need an overhaul in this area of the code. This seems like too much work just to add a pointer relocation.

jdmpapin:
Updated to address @mstoodle's feedback, and to resolve conflicts since #2672 has been merged

runtime/compiler/runtime/RelocationRecord.cpp
jdmpapin: Assertion added

runtime/compiler/runtime/RelocationRecord.cpp
jdmpapin: Updated to define and call a helper method that uses `reloTarget` to read `_offsetToJ2IVirtualThunkPointer`

jdmpapin:
Removed WIP now that eclipse/omr#2965 has been merged

mstoodle:
@andrewcraik has graciously agreed to usher this PR the rest of the way, since I will not be able to keep a very close eye on it for the next little while. Thanks, Andrew!

andrewcraik:

andrewcraik:
travis failed due to the il change that depended on #2649 which has been merged now to resolve the failure

andrewcraik:
JDK11 windows fails look to be existing known issues/unrelated to this change

andrewcraik:
JDK11 Linux fails look to be existing known issues/unrelated to this change

andrewcraik:
Given the failures are unrelated/known issues I'll merge this now to unblock other changes (eg #2875)

PR #2851, opened 2018-09-13
Dispatch direct (private) calls in x86 PicBuilder

jdmpapin:
In x86 PicBuilder, handle J9TR_J9_VTABLE_INDEX_DIRECT_METHOD_FLAG for
virtual calls and J9TR_J9_ITABLE_OFFSET_DIRECT for interface calls as
required for nestmates unresolved private invoke (#2673).

Issue #2801, opened 2018-09-07
CloneWeakReferenceTest failed with CloneNotSupportedException on linux_x86-64_cmprssptrs in JDK11


llxia:
CloneWeakReferenceTest failed with CloneNotSupportedException on linux_x86-64_cmprssptrs in JDK11

Java Version:

```
8:17:25 openjdk version "11-internal" 2018-09-25
08:17:25 OpenJDK Runtime Environment (build 11-internal+0-adhoc.j9build.Build-JDK11-linuxx86-64cmprssptrs)
08:17:25 Eclipse OpenJ9 VM (build fix_2562_dev-b20f4811, JRE 11 Linux amd64-64-Bit Compressed References 20180906_87 (JIT enabled, AOT enabled)
08:17:25 OpenJ9   - b20f4811
08:17:25 OMR      - ac8c19b2
08:17:25 JCL      - 79da47c4f2f based on jdk-11+28)
```

Output:
```
+++ j9vm.test.clone.CloneWeakReferenceTest: +++
command: /home/j9build/workspace/TestBuild_Sandbox/openjdkbinary/j2sdk-image/bin/java  -Xcompressedrefs -Xgcpolicy:metronome -Xcompressedrefs  -Xdump -Xms64m -Xmx64m    -classpath /home/j9build/workspace/TestBuild_Sandbox/openjdk-tests/TestConfig/scripts/testKitGen/../../../../jvmtest/functional/VM_Test/VM_Test.jar:/home/j9build/workspace/TestBuild_Sandbox/openjdk-tests/TestConfig/scripts/testKitGen/../../../../jvmtest/TestConfig/lib/asm-all.jar  j9vm.test.clone.CloneWeakReferenceTest 

Exception in thread "main" java.lang.CloneNotSupportedException: Create a new Reference, since a Reference cannot be cloned.
	at java.lang.ref.Reference.clone(java.base@11-internal/Reference.java:221)
	at j9vm.test.clone.CloneWeakReferenceTest$SubWeakReference.clone(CloneWeakReferenceTest.java:85)
	at j9vm.test.clone.CloneWeakReferenceTest.main(CloneWeakReferenceTest.java:44)
no-zero exit value: 1
*** Test FAILED *** (j9vm.test.clone.CloneWeakReferenceTest)
```

pdbain-ibm:
From the Javadoc:
```
clone
protected Object clone() throws CloneNotSupportedException
[...]

PR #2875, opened 2018-09-14
Relocate the J2I virtual thunk in PIC data on x86

jdmpapin:


jdmpapin:
Marked WIP because this depends on both #2851 and #2874. For the moment I have copies of those changes in this branch so that it will build, but this PR is just for the change in d281f6b

jdmpapin:
@irinarada FYI

runtime/compiler/x/codegen/CallSnippet.cpp
```diff
@@ -67,12 +67,44 @@ uint8_t *TR::X86PicDataSnippet::encodeConstantPoolInfo(uint8_t *cursor)
...
+      TR_ASSERT(
+         TR::Compiler->target.is64Bit(),
+         "expecting a 64-bit target for thunk relocations");
+
+      auto info =
+         (TR_RelocationRecordInformation *)comp()->trMemory()->allocateMemory(
+            sizeof (TR_RelocationRecordInformation),
+            heapAlloc);
+
+      int offsetToJ2IVirtualThunk = isInterface() ? 0x22 : 0x18;
```
mstoodle: Is there a way these `0x22` and `0x18` numbers could be computed from "cursor" locations rather than these more brittle magic numbers? For example, remembering the cursor of the constant pool address and the cursor of the thunk address as the PICs are laid down, then generating the external relocation afterwards at which point you can just compute the difference?

jdmpapin: They could be, although that would separate the creation of `TR_J2IVirtualThunkPointer` from `TR_ConstantPool` and `TR_Thunks`, and the offset constants would still be necessary [elsewhere](https://github.com/eclipse/openj9/blob/bada7c8b47cd8ec96825d3b04837e3321dec47e0/runtime/compiler/x/runtime/X86PicBuilder.inc#L89-L103).

irinarada:
@jdmpapin - do we have an ETA for this one?

runtime/compiler/x/codegen/CallSnippet.cpp
mstoodle: *sigh* that's why it works better as a separate relocation record. Ok, if it's too much hassle, I suppose you'll have to live with it ;) .

andrewcraik:
Jenkins test xlinux,win jdk11,jdk8

andrewcraik:
Jenkins test sanity xlinux,win jdk11,jdk8

andrewcraik:
Jenkins test sanity xlinux,win jdk11,jdk8

jdmpapin:
The line endings check timed out, likely not due to line ending problems. Can we restart it?

Linux JDK11 sanity had a few failures, none of which are related to my changes:
- test_toLowerCase_lithuanian is #2890;
- CloneWeakReferenceTest is #2801; and
- cmdLineTester_SCCommandLineOptionTests_1 nameOption6 and nameOption18 were (I think) due to  528b82f, which has been reverted.

Windows JDK11 sanity failed test_toLowerCase_lithuanian and CloneWeakReferenceTest, as above.

Linux JDK8 sanity failed nameOption6 and nameOption18, as above.

fjeremic:
#2890 is fixed. #2801 is in progress. I'll let whoever gets assigned to this evaluate whether we should merge or wait for a clean build. Probably @andrewcraik?

Jenkins line endings check

andrewcraik:
given that this is core feature work for 0.10, that the tests in question will not be affected by the changes that are intended to be made in this change, and that we need to complete feature work for 0.10 I'm going to say that I am happy merging this at this time.

Thanks all for the detailed analysis of the results.

Issue #2708, opened 2018-08-26
Test-extended.functional-JDK11-linux_x86-64_cmprssptrs JVMCFRE145 Nest host must be a constant class


JasonFengJ9:
Exception in thread "main" java.lang.ClassFormatError: JVMCFRE145 Nest host must be a constant class; class=com/ibm/j9/jsr292/AdaptorTests$PackageExamplesCrossPackageSubClass, offset=521
	at java.lang.ClassLoader.defineClassImpl(java.base@11-internal/Native Method)
	at java.lang.ClassLoader.defineClass(java.base@11-internal/ClassLoader.java:459)
	at java.security.SecureClassLoader.defineClass(java.base@11-internal/SecureClassLoader.java:174)
	at jdk.internal.loader.BuiltinClassLoader.defineClass(java.base@11-internal/BuiltinClassLoader.java:1098)
	at jdk.internal.loader.BuiltinClassLoader.findClassOnClassPathOrNull(java.base@11-internal/BuiltinClassLoader.java:886)
	at jdk.internal.loader.BuiltinClassLoader.loadClassOrNull(java.base@11-internal/BuiltinClassLoader.java:793)
	at jdk.internal.loader.BuiltinClassLoader.loadClass(java.base@11-internal/BuiltinClassLoader.java:751)
	at jdk.internal.loader.ClassLoaders$AppClassLoader.loadClass(java.base@11-internal/ClassLoaders.java:178)
	at java.lang.ClassLoader.loadClass(java.base@11-internal/ClassLoader.java:1032)
	at java.lang.Class.getDeclaredClassesImpl(java.base@11-internal/Native Method)
	at java.lang.Class.getClasses(java.base@11-internal/Class.java:538)
	at org.testng.internal.ClassInfoMap.registerClass(ClassInfoMap.java:43)
	at org.testng.internal.ClassInfoMap.<init>(ClassInfoMap.java:29)
	at org.testng.internal.ClassInfoMap.<init>(ClassInfoMap.java:21)
	at org.testng.TestRunner.initMethods(TestRunner.java:369)
	at org.testng.TestRunner.init(TestRunner.java:271)
	at org.testng.TestRunner.init(TestRunner.java:241)
	at org.testng.TestRunner.<init>(TestRunner.java:167)
	at org.testng.SuiteRunner$DefaultTestRunnerFactory.newTestRunner(SuiteRunner.java:663)
	at org.testng.SuiteRunner.init(SuiteRunner.java:260)
[...]

PR #2878, opened 2018-09-14
Fix Bug with Condy Primitive Long and Double

r30shah:
While generating IL for byte code "ldc2lw" and "ldc2dw" we were assuming the constant is long and double type respectively. This is not true in case of Constant Dynamic where these primitives are boxed into corresponding object and we need to take special steps for generating appropriate IL.


r30shah:
@fjeremic @andrewcraik @yanluo7  These is the fix needed for JIT to identify correct type for primitive long and double condy

yanluo7:
LGTM. 
The `method()->getLDCType(cpIndex)` upfront in `loadFromCP` should determine the appropriate type and address this issue.

fjeremic:
Jenkins test sanity xlinux,plinux,zlinux JDK11

andrewcraik:
LGTM from me - I'll let @fjeremic marshal it through

fjeremic:
JDK11 failures due to #2708. This was way before we merged any of the JIT changes in this area so it cannot be related to this change.

PR #2879, opened 2018-09-14
[WIP] Fix linker (ld) cannot find omrsig on ARM

babsingh:
On ARM, `j9ddrgen` and `constgen` are build using the `-static` compiler flag.
`-static` prevents shared libraries to be linked.

When `J9VM_PORT_OMRSIG_SUPPORT` is enabled, `j9ddrgen` and `constgen` need to
link to `omrsig`, which is a shared library.

Now onwards, `-static` compiler flag will not be used for `j9ddrgen` and
`constgen` on ARM if `J9VM_PORT_OMRSIG_SUPPORT` is enabled. This will allow
`j9ddrgen` and `constgen` to link to `omrsig`.

### DIFF
--- a/runtime/ddr/module.xml
+++ b/runtime/ddr/module.xml
@@ -276,7 +276,7 @@ SPDX-License-Identifier: EPL-2.0 OR Apache-2.0 OR GPL-2.0 WITH Classpath-excepti
 		</includes>
 		<makefilestubs>
 			<makefilestub data="UMA_EXE_POSTFIX_FLAGS+=-static">
-				<include-if condition="spec.linux_arm.*"/>
+				<include-if condition="spec.linux_arm.* and not spec.flags.J9VM_PORT_OMRSIG_SUPPORT"/>
 			</makefilestub>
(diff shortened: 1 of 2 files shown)


babsingh:
fyi - @pshipton 

Running a personal build to verify if the fix resolves the issue on ARM. Will post a comment after verification.

babsingh:
with the above fix, `j9ddrgen` and `constgen` compile fine and `ld` can successfully locate `omrsig`. but, i am getting another error:
```
/lib/ld-linux.so.3: No such file or directory
```
Shared libraries specific to ARM cross compile tools can't be found.

I have to decided to temporarily disable `J9VM_PORT_OMRSIG_SUPPORT` on ARM until the above issues are fixed: https://github.com/eclipse/openj9/pull/2881.

JamesKingdon:
Ok, so I guess that's a qemu setup issue, but I haven't used it, so no idea what the problem might be. My builds don't build ddr yet, so don't attempt to run any of the built ARM binaries under qemu. It would be interesting to confirm that constgen runs ok on an native ARM hardware to verify that the problem is with qemu.

PR #2883, opened 2018-09-16
Delete unused cache of CodeCacheManager from PersistentInfo

0xdaryl:
The only reference was in DebugExt, but its value was always NULL.


0xdaryl:
Note to reviewers:

This PR is to remove a `CodeCacheManager` dependence in `PersistentInfo`.  Its only use is this debug extension, and it is always NULL, hence the affected function has not been doing anything useful since it couldn't get a `CodeCacheManager`.

The right fix for this is to use the `codeCacheManager` that is cached in the private JITConfig struct (available from `javaVM->jitConfig->privateConfig->codeCacheManager` like kca uses).  However, the `privateConfig` will have to be materialized from the remote VM.  It wasn't done when the local VM was materialized in `dbgext/j9dbgext.c` presumably because it doesn't have access to the `TR_JitPrivateConfig` struct.  Materializing this in the JIT debug extension requires the remote VM which is not presently available.

Fixing this properly is far more work with questionable benefit (no real consumers of the debug extensions).  Hence, I will leave this unimplemented with a comment in the code.  The functionality after this PR is no different than what currently exists (i.e., none).

0xdaryl:
Jenkins test sanity

PR #2885, opened 2018-09-17
Fold VarHandle held in static final fields with fear

liqunl:
Fold static final fields that hold VarHandle object with OSR guard. The
folding occurs in VP and when OSR infrastructure is still available.

### DIFF
--- a/runtime/compiler/compile/J9Compilation.hpp
+++ b/runtime/compiler/compile/J9Compilation.hpp
@@ -315,6 +315,10 @@ class OMR_EXTENSIBLE Compilation : public OMR::CompilationConnector
 
    TR::SymbolValidationManager *getSymbolValidationManager() { return _symbolValidationManager; }
 
+   // Flag to record if any optimization has prohibited OSR over a range of trees
+   void setOSRProhibitedOverRangeOfTrees() { _osrProhibitedOverRangeOfTrees = true; }
+   bool isOSRProhibitedOverRangeOfTrees() { return _osrProhibitedOverRangeOfTrees; }
+
 private:
    enum CachedClassPointerId
       {
@@ -399,6 +403,7 @@ private:
    TR_RelocationRuntime *_reloRuntime;
 
    TR::SymbolValidationManager *_symbolValidationManager;
+   bool _osrProhibitedOverRangeOfTrees;
    };
 
 }
--- a/runtime/compiler/optimizer/J9TransformUtil.cpp
+++ b/runtime/compiler/optimizer/J9TransformUtil.cpp
@@ -1928,30 +1928,29 @@ J9::TransformUtil::createDiamondForCall(TR::Optimization* opt, TR::TreeTop *call
 void J9::TransformUtil::removePotentialOSRPointHelperCalls(TR::Compilation* comp, TR::TreeTop* start, TR::TreeTop* end)
    {
    TR_ASSERT(start->getEnclosingBlock() == end->getEnclosingBlock(), "Does not support range across blocks");
+   TR_ASSERT(comp->supportsInduceOSR() && comp->isOSRTransitionTarget(TR::postExecutionOSR) && comp->getOSRMode() == TR::voluntaryOSR,
+             "removePotentialOSRPointHelperCalls only works in certain modes");
 
    TR::TreeTop* ttAfterEnd = end->getNextTreeTop();
+   TR::TreeTop *tt = start;
 
-   if (comp->getOption(TR_EnableOSR) &&
-       comp->isOSRTransitionTarget(TR::postExecutionOSR) &&
-       comp->getOSRMode() == TR::voluntaryOSR)
+   do
       {
-      TR::TreeTop *tt = start;
-      do
+      TR::Node *osrNode = NULL;
+      if (comp->isPotentialOSRPoint(tt->getNode(), &osrNode))
          {
-         TR::Node *osrNode = NULL;
-         if (comp->isPotentialOSRPoint(tt->getNode(), &osrNode))
+         if (osrNode->isPotentialOSRPointHelperCall())
             {
-            if (osrNode->isPotentialOSRPointHelperCall())
-               {
-               dumpOptDetails(comp, "Remove tt n%dn with potential osr point %p n%dn\n", tt->getNode()->getGlobalIndex(), osrNode, osrNode->getGlobalIndex());
-               TR::TreeTop* prevTT = tt->getPrevTreeTop();
-               TR::TransformUtil::removeTree(comp, tt);
-               tt = prevTT;
-               }
+            dumpOptDetails(comp, "Remove tt n%dn with potential osr point %p n%dn\n", tt->getNode()->getGlobalIndex(), osrNode, osrNode->getGlobalIndex());
+            TR::TreeTop* prevTT = tt->getPrevTreeTop();
+            TR::TransformUtil::removeTree(comp, tt);
+            tt = prevTT;
             }
-         tt = tt->getNextTreeTop();
-         } while (tt != ttAfterEnd);
+         }
+      tt = tt->getNextTreeTop();
       }
+   while (tt != ttAfterEnd);
+
    }
 
 /** \brief
@@ -1973,24 +1972,24 @@ void J9::TransformUtil::removePotentialOSRPointHelperCalls(TR::Compilation* comp
 void J9::TransformUtil::prohibitOSROverRange(TR::Compilation* comp, TR::TreeTop* start, TR::TreeTop* end)
    {
    TR_ASSERT(start->getEnclosingBlock() == end->getEnclosingBlock(), "Does not support range across blocks");
+   TR_ASSERT(comp->supportsInduceOSR() && comp->isOSRTransitionTarget(TR::postExecutionOSR) && comp->getOSRMode() == TR::voluntaryOSR,
+             "prohibitOSROverRange only works in certain modes");
 
    TR::TreeTop* ttAfterEnd = end->getNextTreeTop();
+   TR::TreeTop *tt = start;
 
-   if (comp->getOption(TR_EnableOSR) &&
-       comp->isOSRTransitionTarget(TR::postExecutionOSR) &&
-       comp->getOSRMode() == TR::voluntaryOSR)
+   do
       {
-      TR::TreeTop *tt = start;
-      do
+      TR::Node *osrNode = NULL;
+      if (comp->isPotentialOSRPoint(tt->getNode(), &osrNode))
          {
-         TR::Node *osrNode = NULL;
-         if (comp->isPotentialOSRPoint(tt->getNode(), &osrNode))
-            {
-            dumpOptDetails(comp, "Can no longer OSR at [%p] n%dn\n", osrNode, osrNode->getGlobalIndex());
-            osrNode->getByteCodeInfo().setDoNotProfile(true);
-            }
-         tt = tt->getNextTreeTop();
-         } while (tt != ttAfterEnd);
+         dumpOptDetails(comp, "Can no longer OSR at [%p] n%dn\n", osrNode, osrNode->getGlobalIndex());
+         // Record the prohibition so other opts are aware of the existence of dangerous region
+         comp->setOSRProhibitedOverRangeOfTrees();
+         osrNode->getByteCodeInfo().setDoNotProfile(true);
+         }
+      tt = tt->getNextTreeTop();
       }
-   }
+   while (tt != ttAfterEnd);
 
+   }
--- a/runtime/compiler/optimizer/J9ValuePropagation.cpp
+++ b/runtime/compiler/optimizer/J9ValuePropagation.cpp
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2000, 2018 IBM Corp. and others
+ * Copyright (c) 2000, 2019 IBM Corp. and others
  *
  * This program and the accompanying materials are made available under
  * the terms of the Eclipse Public License 2.0 which accompanies this
@@ -43,6 +43,7 @@
 #include "env/VMAccessCriticalSection.hpp"      // for VMAccessCriticalSection
 #include "runtime/RuntimeAssumptions.hpp"
 #include "env/J9JitMemory.hpp"
+#include "optimizer/HCRGuardAnalysis.hpp"
 
 #define OPT_DETAILS "O^O VALUE PROPAGATION: "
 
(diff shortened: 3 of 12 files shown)


runtime/compiler/optimizer/J9TransformUtil.cpp
```diff
@@ -1982,3 +1982,63 @@ void J9::TransformUtil::prohibitOSROverRange(TR::Compilation* comp, TR::TreeTop*
...
+ *     The tree marking the start of the range.
+ *
+ *  \param end
+ *     The tree marking the end of the range.
+ */
+void J9::TransformUtil::prohibitGuardedStaticFinalFieldFoldingOverRange(TR::Compilation* comp, TR::TreeTop* start, TR::TreeTop* end)
+   {
+   TR_ASSERT(start->getEnclosingBlock() == end->getEnclosingBlock(), "Does not support range across blocks");
+
+   traceMsg(comp, "prohibit final field folding over n%dn - n%dn \n", start->getNode()->getGlobalIndex(), end->getNode()->getGlobalIndex());
```
andrewcraik: this should be dumpOptDetails

runtime/compiler/optimizer/J9TransformUtil.cpp
```diff
@@ -1982,3 +1982,63 @@ void J9::TransformUtil::prohibitOSROverRange(TR::Compilation* comp, TR::TreeTop*
...
+   TR::TreeTop* ttAfterEnd = end->getNextTreeTop();
+
+   TR::NodeChecklist visited(comp);
+
+   if (comp->getOption(TR_EnableOSR) &&
+       comp->isOSRTransitionTarget(TR::postExecutionOSR) &&
+       comp->getOSRMode() == TR::voluntaryOSR)
```
andrewcraik: prohibition only works in certain modes - we should probably FATAL_ASSERT if it isn't supported or have some kind of return code. We shouldn't just silently fail.

runtime/compiler/optimizer/J9TransformUtil.cpp
```diff
@@ -1982,3 +1982,63 @@ void J9::TransformUtil::prohibitOSROverRange(TR::Compilation* comp, TR::TreeTop*
...
+      {
+      TR::TreeTop *tt = start;
+      do
+         {
+         prohibitGuardedStaticFinalFieldFoldingOnNodeAndChildren(comp, tt->getNode(), &visited);
+         tt = tt->getNextTreeTop();
+         } while (tt != ttAfterEnd);
```
andrewcraik: code style - while on next line.

runtime/compiler/optimizer/J9TransformUtil.cpp
```diff
@@ -1982,3 +1982,63 @@ void J9::TransformUtil::prohibitOSROverRange(TR::Compilation* comp, TR::TreeTop*
...
+      return;
+
+   visited->add(node);
+
+   int32_t numChildren = node->getNumChildren();
+   if (numChildren == 0 &&
+       node->getOpCode().isLoadVarDirect() &&
+       node->isLoadOfStaticFinalField())
+      {
+      traceMsg(comp, "prohibit folding on n%dn\n", node->getGlobalIndex());
```
andrewcraik: There should be a some kind of trace flag or something so this doesn't fill the log up.

runtime/compiler/optimizer/J9ValuePropagation.cpp
```diff
@@ -1175,3 +1176,163 @@ J9::ValuePropagation::getParmValues()
...
+         && tt->getNode()->getFirstChild()->isTheVirtualCallNodeForAGuardedInlinedCall())
+         return true;
+      }
+   return false;
+   }
+
+static bool skipFinalFieldFoldingInBlock(TR::Block* block)
+   {
+   if (block->isCold() ||
+       TR_FearPointAnalysis::shouldSkipBlock(block) ||
```
andrewcraik: I'm not a fan of this coupling - it isn't obvious why FearPointAnalysis should be connected to VP. The common concept probably belongs in its own header or something with a more sensible name.

runtime/compiler/optimizer/J9ValuePropagation.cpp
```diff
@@ -1175,3 +1176,163 @@ J9::ValuePropagation::getParmValues()
 
    TR_ASSERT(parmIterator->atEnd() && parmIndex == numParms, "Bad signature for owning method");
    }
+
+static bool isTakenSideOfAVirtualGuard(TR::Block* block)
```
andrewcraik: This seems misleading - what about an OSR guard or a virtual guard which is implemented using OSR... Why not just check if the block's predecessors to see if the end in a branch that is a virtual guard?

runtime/compiler/optimizer/J9ValuePropagation.cpp
```diff
@@ -43,6 +43,7 @@
 #include "env/VMAccessCriticalSection.hpp"      // for VMAccessCriticalSection
 #include "runtime/RuntimeAssumptions.hpp"
 #include "env/J9JitMemory.hpp"
+#include "optimizer/FearPointAnalysis.hpp"
```
andrewcraik: really not a fan of this - see below we need a more sensible header to hold the common query if there isn't already a suitable OSR header.

runtime/compiler/compile/J9SymbolReferenceTable.cpp
```diff
@@ -1400,7 +1400,7 @@ J9::SymbolReferenceTable::findOrCreateStaticSymbol(TR::ResolvedMethodSymbol * ow
          TR_OpaqueClassBlock *declaringClass = owningMethod->getDeclaringClassFromFieldOrStatic(comp(), cpIndex);
          if (declaringClass && fej9->isClassInitialized(declaringClass))
             {
-            static const char *dontFoldVarHandle = feGetEnv("TR_DontFoldVarHandle");
+            static const char *foldVarHandleWithoutGuard = feGetEnv("TR_FoldVarHandleWithoutGuard");
```
andrewcraik: I would prefer we delivered the code and then flipped the sense of this so we only deal with one thing at a time to make narrowing things down easier.

runtime/compiler/optimizer/J9ValuePropagation.cpp
```diff
@@ -1175,3 +1176,163 @@ J9::ValuePropagation::getParmValues()
...
+       !comp()->isOSRTransitionTarget(TR::postExecutionOSR) ||
+       !comp()->getOSRMode() == TR::voluntaryOSR)
+      {
+      return false;
+      }
+
+   if (fieldNode->getByteCodeInfo().doNotProfile() ||
+       skipFinalFieldFoldingInBlock(tree->getEnclosingBlock()) ||
+       !safeToAddFearPointIn(comp(), fieldNode->getByteCodeInfo().getCallerIndex()) ||
+       TR::TransformUtil::canFoldStaticFinalField(comp(), fieldNode) != TR_maybe)
```
andrewcraik: I think this would read better with the ||s a the start of the lines to make the condition relationships more obvious since this is quite complicated.

runtime/compiler/optimizer/J9ValuePropagation.cpp
```diff
@@ -1175,3 +1176,163 @@ J9::ValuePropagation::getParmValues()
...
+   if (trace())
+      {
+      traceMsg(comp(),
+              "Looking at static final field n%dn %.*s declared in class %p\n",
+              fieldNode->getGlobalIndex(), fieldNameLen, fieldName, declaringClass);
+      }
+
+   if (isStaticFinalFieldWorthFolding(comp(), declaringClass, fieldSignature, fieldSigLength))
+      {
+      if (TR::TransformUtil::foldStaticFinalFieldAssumingProtection(comp(), fieldNode))
```
andrewcraik: A perform transformation would seem to be a good idea here...

liqunl: `foldStaticFinalFieldAssumingProtection` does a perform transformation.

runtime/compiler/optimizer/J9ValuePropagation.cpp
liqunl: Fixed.

runtime/compiler/compile/J9SymbolReferenceTable.cpp
liqunl: Change in this file will be in a different PR.

runtime/compiler/optimizer/J9ValuePropagation.cpp
liqunl: Changed to check the last tree of the block's predecessor.

runtime/compiler/optimizer/J9TransformUtil.cpp
liqunl: This is where we set the flag, so I updated the code to use dumpOptDetails here.

runtime/compiler/optimizer/J9TransformUtil.cpp
liqunl: Fixed.

runtime/compiler/optimizer/J9TransformUtil.cpp
liqunl: Added an assertion and StringPeepHoles now checks the mode before prohibition.

runtime/compiler/optimizer/J9ValuePropagation.cpp
liqunl: Renamed the function to isOSRRelatedBlock and move it to OSRUtils.hpp.

runtime/compiler/optimizer/J9TransformUtil.cpp
```diff
@@ -1916,30 +1916,29 @@ J9::TransformUtil::createDiamondForCall(TR::Optimization* opt, TR::TreeTop *call
 void J9::TransformUtil::removePotentialOSRPointHelperCalls(TR::Compilation* comp, TR::TreeTop* start, TR::TreeTop* end)
    {
    TR_ASSERT(start->getEnclosingBlock() == end->getEnclosingBlock(), "Does not support range across blocks");
+   TR_ASSERT(comp->supportsInduceOSR() && comp->isOSRTransitionTarget(TR::postExecutionOSR) && comp->getOSRMode() == TR::voluntaryOSR,
```
andrewcraik: do we also need to check that the OSR infrastructure hasn't been removed? it doesn't make sense to do this after that point I don't think since we have made a lot of complex decisions.

runtime/compiler/optimizer/J9TransformUtil.cpp
```diff
@@ -1961,24 +1960,82 @@ void J9::TransformUtil::removePotentialOSRPointHelperCalls(TR::Compilation* comp
...
+ *  \param start
+ *     The tree marking the start of the range.
+ *
+ *  \param end
+ *     The tree marking the end of the range.
+ */
+void J9::TransformUtil::prohibitGuardedStaticFinalFieldFoldingOverRange(TR::Compilation* comp, TR::TreeTop* start, TR::TreeTop* end)
+   {
+   TR_ASSERT(start->getEnclosingBlock() == end->getEnclosingBlock(), "Does not support range across blocks");
+   TR_ASSERT(comp->supportsInduceOSR() && comp->isOSRTransitionTarget(TR::postExecutionOSR) && comp->getOSRMode() == TR::voluntaryOSR,
```
andrewcraik: isn't it more correct to say we only do static final field folding in specific modes? the prohibition should work in all modes where we fold?

runtime/compiler/optimizer/J9ValuePropagation.cpp
```diff
@@ -1175,3 +1176,167 @@ J9::ValuePropagation::getParmValues()
...
+      }
+
+   if (!comp()->supportsInduceOSR()
+       || !comp()->isOSRTransitionTarget(TR::postExecutionOSR)
+       || !comp()->getOSRMode() == TR::voluntaryOSR)
+      {
+      return false;
+      }
+
+   if (fieldNode->getByteCodeInfo().doNotProfile()
```
andrewcraik: hmm - I'm not sure I like this use of doNotProfile - whether a static final field read is foldable is not related to whether the field read node came from the original program representation so this would seem an overload of the meaning of the flag likely to confuses other opts

runtime/compiler/optimizer/StringPeepholes.cpp
```diff
@@ -2328,3 +2324,21 @@ TR::TreeTop *TR_StringPeepholes::searchForToStringCall(TR::TreeTop *tt, TR::Tree
...
+      if (trace())
+         {
+         TR::Node* startNode = startTree->getNode();
+         TR::Node* endNode = endTree->getNode();
+         traceMsg(comp(), "Post process Trees from %p n%dn to %p n%dn for OSR\n", startNode, startNode->getGlobalIndex(), endNode, endNode->getGlobalIndex());
+         }
+
+      TR::TransformUtil::removePotentialOSRPointHelperCalls(comp(), startTree, endTree);
+      TR::TransformUtil::prohibitOSROverRange(comp(), startTree, endTree);
+      TR::TransformUtil::prohibitGuardedStaticFinalFieldFoldingOverRange(comp(), startTree, endTree);
```
andrewcraik: I don't think this is necessary - we can fold finals if we can protect the OSR points

runtime/compiler/optimizer/OSRUtils.cpp
```diff
@@ -0,0 +1,46 @@
...
+              && tt->getNode()->getFirstChild()->getSymbolReference()->getReferenceNumber() == TR_prepareForOSR)
+             return true;
+          }
+       }
+   return false;
+   }
+
+bool isOSRRelatedBlock(TR::Block *block)
+   {
+   return block->isOSRCatchBlock() || block->isOSRCodeBlock() || containsPrepareForOSR(block);
```
andrewcraik: what blocks contain a prepare that is not a catch or code block?

runtime/compiler/compile/J9Compilation.hpp
```diff
@@ -313,6 +313,10 @@ class OMR_EXTENSIBLE Compilation : public OMR::CompilationConnector
 
    TR::SymbolValidationManager *getSymbolValidationManager() { return _symbolValidationManager; }
 
+   // Flag to record if any optimization has prohibited OSR over a range of trees
+   void setOSRProhibitedOverRangeOfTrees() { _osrProhibitedOverRangeOfTrees = true; }
+   bool isOSRProhibitedOverRangeOfTrees() { return _osrProhibitedOverRangeOfTrees; }
```
andrewcraik: perhaps hasOSRProhibitions would be a better, shorter name?

runtime/compiler/optimizer/J9ValuePropagation.cpp
```diff
@@ -1175,3 +1177,235 @@ J9::ValuePropagation::getParmValues()
...
+static bool isTakenSideOfAVirtualGuard(TR::Compilation* comp, TR::Block* block)
+   {
+   // First block can never be taken side
+   if (block == comp->getMethodSymbol()->getFirstTreeTop()->getEnclosingBlock())
+      return false;
+
+   // Due to VirtualGuardHeadMerger, the taken side of a virtual guard may have more than one predecessors,
+   // each containing a virtual guard that branches to the taken side. It's sufficient to look at only
+   // one predecessor.
+   TR::Node* predLastRealNode = block->getPredecessors().size() > 0 ?
```
andrewcraik: VGHM or other opts may also chain the cold side of guards together - see VirtualGuardTailSplitter - it is not sufficient to look at only one.

liqunl:
@andrewcraik Can you review again?

runtime/compiler/optimizer/J9TransformUtil.cpp
```diff
@@ -1961,24 +1960,24 @@ void J9::TransformUtil::removePotentialOSRPointHelperCalls(TR::Compilation* comp
 void J9::TransformUtil::prohibitOSROverRange(TR::Compilation* comp, TR::TreeTop* start, TR::TreeTop* end)
    {
    TR_ASSERT(start->getEnclosingBlock() == end->getEnclosingBlock(), "Does not support range across blocks");
+   TR_ASSERT(comp->supportsInduceOSR() && comp->isOSRTransitionTarget(TR::postExecutionOSR) && comp->getOSRMode() == TR::voluntaryOSR,
```
andrewcraik: Do we need something to check about if the OSR infrastructure has been removed already since this operation doesn't make sense in that world right?

runtime/compiler/optimizer/J9ValuePropagation.cpp
```diff
@@ -1175,3 +1176,237 @@ J9::ValuePropagation::getParmValues()
...
+       tryFoldStaticFinalFieldAt(_curTree, node))
+      {
+      return true;
+      }
+
+   return false;
+   }
+
+
+static TR_HCRGuardAnalysis* runHCRGuardAnalysisIfNecessary()
```
andrewcraik: This might be better named `runHCRGuardAnalysisIfPossible` since we would like it, but we don't have to do it which I think necessary suggests.

runtime/compiler/optimizer/J9TransformUtil.cpp
liqunl: `comp->supportsInduceOSR()` does the check.

andrewcraik:
Ok so I think this new revision where VP is checking the safety of folding with an optional, but not currently supported hook for running HCRGuard analysis seems like a good design now where we don't rely on node flags and we will be conservatively correct. Given the complexity I'd appreciate a second opinion from @jdmpapin but I'm going to start sanity. If there are additional checks added to asserts please do it as a separate commit that can be squashed so we can avoid having to do a full sanity by inspecting the change prior to the squash.

runtime/compiler/optimizer/J9TransformUtil.cpp
liqunl: `comp->supportsInduceOSR()` does the check.

andrewcraik:

jdmpapin:
I suppose VP catches these things before inlining as part of `methodHandleInvokeInliningGroup`?

runtime/compiler/optimizer/J9ValuePropagation.cpp
```diff
@@ -1175,3 +1176,237 @@ J9::ValuePropagation::getParmValues()
...
+
+   // The caller frame has to be checked first
+   int32_t callerIndex = tt->getNode()->getByteCodeInfo().getCallerIndex();
+   if (cannotAttemptOSRDuring(comp(), callerIndex))
+      {
+      if (trace())
+         {
+         traceMsg(comp(), "Not safe to add fear point because caller frame %d cannot OSR\n", callerIndex);
+         }
+      return TR_no;
```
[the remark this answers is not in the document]
liqunl: No. That's something I didn't think of when changing the code. It was needed in the old version when we never want to run HCR guard analysis. Thanks for pointing this out.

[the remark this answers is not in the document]
jdmpapin: Come to think of it, the `!isOSRProhibitedOverRangeOfTrees()` seems to be broken even though we've checked for `cannotAttemptOSRDuring()`. Suppose we _can_ attempt OSR at the given tree, but the fear flows out from the beginning of the inlined body and into a different inlined body where we _can't_, and furthermore suppose that we did no transformations in string peepholes (or similar). Even without running a full "HCR guard analysis" (now a bit of a misnomer), the local analysis below would conservatively prevent folding, if it were allowed to run. But instead we'll return early with `TR_yes`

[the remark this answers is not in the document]
liqunl: `potentialOSRPointHelper` will prevent a fear flowing to an inlined body that cannot attempt OSR.

runtime/compiler/optimizer/J9ValuePropagation.cpp
```diff
@@ -1175,3 +1176,237 @@ J9::ValuePropagation::getParmValues()
...
+         {
+         if (comp()->isPotentialOSRPointWithSupport(tt))
+            return TR_yes;
+         else
+            return TR_no;
+         }
+      tt = tt->getPrevTreeTop();
+      }
+
+   TR_HCRGuardAnalysis* guardAnalysis = runHCRGuardAnalysisIfNecessary();
```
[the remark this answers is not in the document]
liqunl: We haven't done any compile time measured so don't know when we can afford the analysis now. It'll be updated once we done the performance measurement.

liqunl:
> 
> 
> I suppose VP catches these things before inlining as part of `methodHandleInvokeInliningGroup`?

Yes.

liqunl:
@jdmpapin Made some changes in a separate commit, will squash it before merge. Please review again.

andrewcraik:

PR #2889, opened 2018-09-17
Added s390x ubuntu16 Dockerfile

cwesMills:
Added new Dockerfile, which creates ubuntu16 s390x container with user "jenkins".
Container can build JDK8, 10, 11, and run system and functional tests.
Changed location of authorized_keys file so both Dockerfiles will be able to reference it.


llxia:
@AdamBrousseau could you approve it?

PR #2900, opened 2018-09-17
Update MethodHandle lookup for Nestmates

fengxue-IS:
- Nestmates allow invokeInterface to target private interface methods

### DIFF
--- a/runtime/jcl/common/java_dyn_methodhandle.c
+++ b/runtime/jcl/common/java_dyn_methodhandle.c
@@ -91,8 +91,10 @@ lookupInterfaceMethod(J9VMThread *currentThread, J9Class *lookupClass, J9UTF8 *n
 		J9InternalVMFunctions *vmFuncs = currentThread->javaVM->internalVMFunctions;
 
 		Assert_JCL_true(!J9_ARE_ANY_BITS_SET(J9_ROM_METHOD_FROM_RAM_METHOD(method)->modifiers, J9AccStatic));
-		/* [PR 67082] private interface methods require invokespecial, not invokeinterface.*/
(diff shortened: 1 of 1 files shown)


DanHeidinga:
@gacholio Can you take a look at this?

DanHeidinga:
> would you prefer the runtime check over the ifdef?

Yes, we've tried to keep the code under runtime time checks whenever possible so this should behave the same way.  We may want to revisit that in the future but for now, I'd prefer consistency - especially when it's not in perf sensitive parts of the code.

runtime/jcl/common/java_dyn_methodhandle.c
```diff
@@ -91,11 +91,16 @@ lookupInterfaceMethod(J9VMThread *currentThread, J9Class *lookupClass, J9UTF8 *n
 		J9InternalVMFunctions *vmFuncs = currentThread->javaVM->internalVMFunctions;
 
 		Assert_JCL_true(!J9_ARE_ANY_BITS_SET(J9_ROM_METHOD_FROM_RAM_METHOD(method)->modifiers, J9AccStatic));
+
+/* With Java 11 Nestmates, invokeInterface is allowed to target private interface methods */
+#if !defined(J9VM_OPT_VALHALLA_NESTMATES)
```
gacholio: Please change this to a runtime check:

`if (J2SE_VERSION(vm) >= J2SE_V11)`

pshipton:
@DanHeidinga @gacholio any further concerns? This appears ready for re-review.

DanHeidinga:

DanHeidinga:
Once the build passes, I'm fine with whoever sees it first to merge it.

pshipton:
ppc has a number of "known" problems #2917 #2934.

I'm going to run the jdk11 testing on another platform.

pshipton:
The CloneNotSupportedException failure is expected, otherwise the jdk11 testing looks good on zlinux.

PR #2904, opened 2018-09-17
Fix Nestmates to throw correct exception type

fengxue-IS:
- Check for Nesthost exception before throwing illegalAccessException

### DIFF
--- a/runtime/vm/resolvesupport.cpp
+++ b/runtime/vm/resolvesupport.cpp
@@ -352,7 +352,8 @@ tryAgain:
 	{
 		IDATA checkResult = checkVisibility(vmStruct, J9_CLASS_FROM_CP(ramCP), accessClass, accessModifiers, resolveFlags);
 		if (checkResult < J9_VISIBILITY_ALLOWED) {
-			if (canRunJavaCode) {
+			/* Check for pending exception for (ie. Nesthost class loading/verify), do not overwrite these exceptions */
+			if (canRunJavaCode && (!VM_VMHelpers::exceptionPending(vmStruct))) {
 				char *errorMsg = NULL;
 				PORT_ACCESS_FROM_VMC(vmStruct);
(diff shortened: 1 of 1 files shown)


runtime/vm/resolvesupport.cpp
```diff
@@ -352,17 +352,24 @@ resolveClassRef(J9VMThread *vmStruct, J9ConstantPool *ramCP, UDATA cpIndex, UDAT
...
-			if (canRunJavaCode) {
-				char *errorMsg = NULL;
-				PORT_ACCESS_FROM_VMC(vmStruct);
-				if (J9_VISIBILITY_NON_MODULE_ACCESS_ERROR == checkResult) {
-					errorMsg = illegalAccessMessage(vmStruct, accessModifiers, J9_CLASS_FROM_CP(ramCP), accessClass, J9_VISIBILITY_NON_MODULE_ACCESS_ERROR);
-				} else {
-					errorMsg = illegalAccessMessage(vmStruct, -1, J9_CLASS_FROM_CP(ramCP), accessClass, checkResult);
+#if defined(J9VM_OPT_VALHALLA_NESTMATES)
+			/* Check for pending exception for Nesthost class loading / verify, do not overwrite these exceptions */
+			if (!VM_VMHelpers::exceptionPending(vmStruct)) {
```
DanHeidinga: I think we want to change the order of these checks - `canRunJavaCode` should be first as their had better not be a pending exception if we can't run java code.

DanHeidinga: Also, I don't think this needs to be Nestmate specific.  If `checkVisibility` has a pending exception, we want to throw that one.

gacholio: If canRunJavaCode is false, then we'd better not be setting an exception in checkVisibility - hopefully resolveFlags takes care of that.

DanHeidinga:

DanHeidinga:
The JDK11 failures are the known set (really, really wish that set was 0!) Merging

PR #2905, opened 2018-09-17
Fix for missing SLOW_JIT_HELPER_EPILOGUE()

0dvictor:
SLOW_JIT_HELPER_EPILOGUE() for impl_jitClassCastException(), fixing it.


andrewcraik:

andrewcraik:
The win-x86-64 sanity failures are both j9vm.test.clone.CloneWeakReferenceTest which is a known issue.
The Linux-x86-64 sanity failure are all j9vm.test.clone.CloneWeakReferenceTest which is a known issue.
Sanity is clean and JDK8 sanity is clean so this should be good to go.

PR #2910, opened 2018-09-18
Fix bastore il trees to comply with JVM spec

cathyzhyi:
the array indexed by index"

This is the JIT part of the change. The VM part is #2046

Issue: #2049

### DIFF
--- a/runtime/compiler/ilgen/IlGenerator.cpp
+++ b/runtime/compiler/ilgen/IlGenerator.cpp
@@ -36,6 +36,7 @@
 #include "infra/Checklist.hpp"
 #include "env/VMJ9.h"
 #include "ilgen/J9ByteCodeIlGenerator.hpp"
+#include "optimizer/BoolArrayStoreTransformer.hpp"
 #include "ras/DebugCounter.hpp"
 #include "optimizer/TransformUtil.hpp"
 
@@ -109,7 +110,7 @@ TR_J9ByteCodeIlGenerator::TR_J9ByteCodeIlGenerator(
       _argPlaceholderSignatureOffset = curArg - signatureChars;
       }
    for( int i=0 ; i < _numDecFormatRenames ; i++ )
-      { 
+      {
       _decFormatRenamesDstSymRef[i] = NULL;
       }
    if (comp->getOption(TR_EnableOSR)
@@ -481,11 +482,56 @@ TR_J9ByteCodeIlGenerator::genILFromByteCodes()
    TR_ScratchList<TR::TreeTop> unresolvedInstanceofTops(comp()->trMemory());
    TR_ScratchList<TR::TreeTop> invokeSpecialInterfaceTops(comp()->trMemory());
    TR::NodeChecklist evaluatedInvokeSpecialCalls(comp());
+   TR_BoolArrayStoreTransformer::NodeSet bstoreiUnknownArrayTypeNodes(std::less<TR::Node *>(), comp()->trMemory()->currentStackRegion());
+   TR_BoolArrayStoreTransformer::NodeSet bstoreiBoolArrayTypeNodes(std::less<TR::Node *>(), comp()->trMemory()->currentStackRegion());
+   TR_BoolArrayStoreTransformer boolArrayStoreTransformer(&bstoreiUnknownArrayTypeNodes, &bstoreiBoolArrayTypeNodes);
 
    for (; currTree != NULL; currTree = currTree->getNextTreeTop())
       {
       TR::Node *currNode = currTree->getNode();
       TR::ILOpCode opcode = currNode->getOpCode();
+
+      if ((opcode.isStoreDirect() && opcode.hasSymbolReference() && currNode->getSymbolReference()->getSymbol()->isAutoOrParm()) ||
+          opcode.isCheckCast())
+         {
+         TR::SymbolReference *symRef = currNode->getSymbolReference();
+         TR::Node *typeNode = NULL;
+         if (opcode.isStoreDirect())
+            typeNode = currNode->getFirstChild(); // store auto
+         else typeNode = currNode->getSecondChild(); // checkcast
+         if (boolArrayStoreTransformer.isAnyDimensionBoolArrayNode(typeNode))
+            boolArrayStoreTransformer.setHasBoolArrayAutoOrCheckCast();
+         else if (boolArrayStoreTransformer.isAnyDimensionByteArrayNode(typeNode))
+            boolArrayStoreTransformer.setHasByteArrayAutoOrCheckCast();
+
+         if (opcode.isStoreDirect() && symRef->getSymbol()->isParm() && currNode->getDataType() == TR::Address)
+            {
+            int lhsLength;
+            int rhsLength;
+            const char *lhsSig = currNode->getTypeSignature(lhsLength, stackAlloc, false /* parmAsAuto */);
+            const char *rhsSig = typeNode->getTypeSignature(rhsLength, stackAlloc, true /* parmAsAuto */);
+            if (!lhsSig || !rhsSig || lhsLength != rhsLength || strncmp(lhsSig, rhsSig, lhsLength))
+               boolArrayStoreTransformer.setHasVariantArgs();
+            }
+         }
+      else if (opcode.getOpCodeValue() == TR::bstorei && currNode->getSymbolReference()->getCPIndex() == -1
+            && currNode->getFirstChild()->isInternalPointer())
+         {
+         TR::Node *arrayBase = currNode->getFirstChild()->getFirstChild();
+         if (arrayBase->getOpCode().hasSymbolReference())
+            {
+            if (boolArrayStoreTransformer.isBoolArrayNode(arrayBase))
+               {
+               if (comp()->getOption(TR_TraceILGen))
+                  traceMsg(comp(), "bstorei node n%dn is bool array store\n", currNode->getGlobalIndex());
+               bstoreiBoolArrayTypeNodes.insert(currNode);
+               }
+            else if (!boolArrayStoreTransformer.isByteArrayNode(arrayBase))
+               bstoreiUnknownArrayTypeNodes.insert(currNode);
+            }
+         else
+            bstoreiUnknownArrayTypeNodes.insert(currNode);
+         }
 
       if (currNode->getOpCodeValue() == TR::checkcast
           && currNode->getSecondChild()->getOpCodeValue() == TR::loadaddr
@@ -522,7 +568,7 @@ TR_J9ByteCodeIlGenerator::genILFromByteCodes()
       if(opcode.isStoreDirect() && opcode.hasSymbolReference())
          {
             TR::SymbolReference *symRef = currNode->getSymbolReference();
-            if( symRef && symRef->getSymbol()->isAutoOrParm())
+            if (symRef && symRef->getSymbol()->isAutoOrParm())
                autoOrParmSymRefList.add(symRef);
          }
 
@@ -676,6 +722,9 @@ TR_J9ByteCodeIlGenerator::genILFromByteCodes()
       for (TR::TreeTop *tree = it.getCurrent(); tree != NULL; tree = it.getNext())
          expandInvokeSpecialInterface(tree);
       }
+
+   if (!bstoreiUnknownArrayTypeNodes.empty() || !bstoreiBoolArrayTypeNodes.empty())
+      boolArrayStoreTransformer.perform();
 
    return true;
    }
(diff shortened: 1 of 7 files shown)


andrewcraik:
@cathyzhyi can you fix the copyrights with your next revision?

cathyzhyi:
@andrewcraik fixed copyrights

cathyzhyi:
Tested on major benchmarks and shows no regression in compilation time and throughput.

andrewcraik:

runtime/compiler/optimizer/BoolArrayStoreTransformer.cpp
```diff
@@ -0,0 +1,434 @@
...
+      TR_ASSERT(arrayTypeNode->getOpCode().isLoadConst(), "expect the second child of TR::newarray to be constant " POINTER_PRINTF_FORMAT, arrayTypeNode);
+      if (arrayTypeNode->getInt() == expectedTypeValue)
+         nodeArrayDimension = 1;
+      }
+   else
+      {
+      char expectedTypeChar = boolType? 'Z' : 'B';
+      int length;
+      const char * signature = node->getTypeSignature(length);
+      if (signature && length >= 2 && signature[length-1] == expectedTypeChar && signature[length-2] == '[')
```
cathyzhyi: added checking `length >= 2` and check if the second last char must be `[` to make sure the type is indeed array or multi array.

runtime/compiler/optimizer/BoolArrayStoreTransformer.cpp
```diff
@@ -0,0 +1,434 @@
...
+            else if (isByteArrayNode(rhs))
+               newType = TR_no;
+            else if (rhs->getOpCode().hasSymbolReference())
+               {
+               TR::AutomaticSymbol *rhsLocal = rhs->getSymbolReference()->getSymbol()->getAutoSymbol();
+               if (rhsLocal && currentTypeInfo)
+                  newType = (*currentTypeInfo)[rhsLocal->getLocalIndex()];
+               }
+
+            if (newType != TR_maybe || currentTypeInfo)
```
cathyzhyi: added `|| currentTypeInfo` to catch the case where currentTypeInfo is `[Z` or `[B` but the new type is unknown.

andrewcraik:

runtime/compiler/optimizer/BoolArrayStoreTransformer.cpp
```diff
@@ -0,0 +1,485 @@
...
+      printTypeInfo(first, comp());
+      traceMsg(comp(), "\n");
+      }
+   }
+
+void TR_BoolArrayStoreTransformer::findBoolArrayStoreNodes()
+   {
+   TR::Region currentRegion(comp()->region());
+
+   collectLocals(comp()->getMethodSymbol()->getAutoSymRefs());
```
cathyzhyi: changed the code to get all the autos used in this method.

runtime/compiler/optimizer/BoolArrayStoreTransformer.cpp
```diff
@@ -0,0 +1,485 @@
...
+      traceMsg(comp(), "<BoolArrayStoreTransformer>\n");
+
+   if (!_bstoreiUnknownArrayTypeNodes->empty() && _hasBoolArrayAuto)
+      {
+      if (_hasByteArrayAuto) // only need to iterate CFG if both byte and boolean array exist
+         findBoolArrayStoreNodes();
+      else // if only boolean array exist then all the bstorei nodes are operating on boolean array
+         {
+         _bstoreiBoolArrayTypeNodes->insert(_bstoreiUnknownArrayTypeNodes->begin(), _bstoreiUnknownArrayTypeNodes->end());
+         _bstoreiUnknownArrayTypeNodes->clear();
```
jdmpapin: I think this can erroneously truncate for a store into a `byte[]`:

    Object newByteArray(int n) { return new byte[n]; }
    ...
    void foo(int n, byte b) {
        boolean[] dummy = new boolean[n];
        ((byte[])newByteArray(n))[0] = b;
    }

runtime/compiler/ilgen/IlGenerator.cpp
```diff
@@ -481,12 +482,45 @@ TR_J9ByteCodeIlGenerator::genILFromByteCodes()
...
+      else if (opcode.getOpCodeValue() == TR::bstorei && currNode->getSymbolReference()->getCPIndex() == -1
+            && currNode->getFirstChild()->isInternalPointer())
+         {
+         TR::Node *arrayBase = currNode->getFirstChild()->getFirstChild();
+         if (arrayBase->getOpCode().hasSymbolReference())
+            {
+            if (boolArrayStoreTransformer.isBoolArrayNode(arrayBase))
+               {
+               if (comp()->getOption(TR_TraceILGen))
+                  traceMsg(comp(), "bstorei node n%dn is bool array store\n", arrayBase->getGlobalIndex());
```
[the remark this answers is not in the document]
cathyzhyi: thanks for pointing this out!

runtime/compiler/optimizer/BoolArrayStoreTransformer.cpp
```diff
@@ -0,0 +1,485 @@
...
+   typedef std::stack<TR::Block*, BlockStackContainer> BlockStack;
+   BlockStack blockStack(BlockStackContainer(comp()->trMemory()->currentStackRegion()));
+
+   typedef TR::typed_allocator<std::pair<const int32_t, TypeInfo *>, TR::Region &> ResultAllocator;
+   typedef std::map<int32_t, TypeInfo *, std::less<int32_t>, ResultAllocator> BlockResultMap;
+   BlockResultMap blockStartTypeInfos(std::less<int32_t>(), comp()->trMemory()->currentStackRegion());
+
+   /*
+    * Do a DFS of the CFG as the best effort to figure out types in one traverse
+    */
```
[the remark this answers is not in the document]
cathyzhyi: changed according to the suggestion. Thanks!

runtime/compiler/optimizer/BoolArrayStoreTransformer.cpp
```diff
@@ -0,0 +1,485 @@
...
+
+void TR_BoolArrayStoreTransformer::collectLocals(TR_Array<List<TR::SymbolReference>> *autosListArray)
+   {
+   for (int i = 0; autosListArray && i < autosListArray->size(); i++)
+      {
+      List<TR::SymbolReference> autosList = (*autosListArray)[i];
+      ListIterator<TR::SymbolReference> autosIt(&autosList);
+      for (TR::SymbolReference* symRef = autosIt.getFirst(); symRef; symRef = autosIt.getNext())
+         {
+         TR::DataType dt = symRef->getSymbol()->getDataType();
```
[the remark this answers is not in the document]
cathyzhyi: deleted.

runtime/compiler/optimizer/BoolArrayStoreTransformer.cpp
```diff
@@ -0,0 +1,485 @@
...
+      TR::Node *arrayTypeNode = node->getSecondChild();
+      TR_ASSERT(arrayTypeNode->getOpCode().isLoadConst(), "expect the second child of TR::newarray to be constant " POINTER_PRINTF_FORMAT, arrayTypeNode);
+      if (arrayTypeNode->getInt() == expectedTypeValue)
+         nodeArrayDimension = 1;
+      }
+   else
+      {
+      char expectedTypeChar = boolType? 'Z' : 'B';
+      int length;
+      const char * signature = node->getTypeSignature(length);
```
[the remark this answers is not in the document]
cathyzhyi: Added code so that when there is any store to arg ignore the parm type signature and treat the parm as normal auto. Thanks!

runtime/compiler/optimizer/BoolArrayStoreTransformer.cpp
```diff
@@ -0,0 +1,485 @@
...
+            TR_YesNoMaybe newType = TR_maybe;
+            if (isBoolArrayNode(rhs))
+               newType = TR_yes;
+            else if (isByteArrayNode(rhs))
+               newType = TR_no;
+            else if (rhs->getOpCode().hasSymbolReference())
+               {
+               TR::AutomaticSymbol *rhsLocal = rhs->getSymbolReference()->getSymbol()->getAutoSymbol();
+               if (rhsLocal && currentTypeInfo)
+                  newType = (*currentTypeInfo)[rhsLocal->getLocalIndex()];
```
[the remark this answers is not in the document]
cathyzhyi: Added code to use type info at the load rather than at the use of the load. Thanks!

runtime/compiler/optimizer/BoolArrayStoreTransformer.cpp
cathyzhyi: Added code to handle checkcast as well. thanks!

runtime/compiler/optimizer/BoolArrayStoreTransformer.cpp
```diff
@@ -0,0 +1,678 @@
...
+ *    seen in the trees
+ */
+void TR_BoolArrayStoreTransformer::findLoadAddressAutoAndFigureOutType(TR::Node *node, TypeInfo * typeInfo, TR::NodeChecklist &boolArrayNodes, TR::NodeChecklist &byteArrayNodes, TR::NodeChecklist &loadAutoNodes)
+   {
+   for (int i = 0; i < node->getNumChildren(); i++)
+      findLoadAddressAutoAndFigureOutType(node->getChild(i), typeInfo, boolArrayNodes, byteArrayNodes, loadAutoNodes);
+
+   if (node->getType() == TR::Address && node->getOpCode().isLoadDirect() && node->getOpCode().hasSymbolReference() &&
+         node->getSymbolReference()->getSymbol()->isAutoOrParm() && !loadAutoNodes.contains(node))
+      {
```
jdmpapin: To avoid taking exponential time, `loadAutoNodes` should probably apply to all nodes encountered by this method rather than only the address-typed auto loads.

andrewcraik:
@fjeremic this has mostly been covered by internal testing. I have asked @tajila what testing there is for the VM part of this change in OpenJ9 since it should also serve as a sanity check for these changes.

runtime/compiler/optimizer/BoolArrayStoreTransformer.cpp
cathyzhyi: Fixed according to comments. Thanks!

runtime/compiler/optimizer/BoolArrayStoreTransformer.cpp
jdmpapin: It looks like this can happen again now that `isAnyDimension{Bool,Byte}ArrayNode()` ignore parameters' signatures:

    void foo(byte[] x, int i, byte b) {
        boolean[] dummy = new boolean[1];
        byte[] y = x;
        y[i] = b;
    }

runtime/compiler/optimizer/BoolArrayStoreTransformer.cpp
```diff
@@ -0,0 +1,678 @@
...
+ * \parm parmAsAuto
+ *    For a \parm node with parm symbol, the API is used in two different ways because the slot for dead
+ *    parm symbol can be reused for variables of any type. When \parm parmAsAuto is true, parm is
+ *    treated as other auto and the type signature is ignored. When \parm parmAsAuto is false, the
+ *    type signature of the parm is used for telling the actual type.
+ */
+bool TR_BoolArrayStoreTransformer::isBoolArrayNode(TR::Node *node, bool parmAsAuto)
+   {
+   if (parmAsAuto && node->getOpCode().hasSymbolReference() && node->getSymbolReference()->getSymbol()->isParm())
+      return false;
```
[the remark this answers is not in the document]
cathyzhyi: Thanks for the comment. Added extra argument `parmAsAuto` to `getTypeSignature` of `TR::Node` to solve this issue.

runtime/compiler/optimizer/BoolArrayStoreTransformer.cpp
cathyzhyi: changed` isAnyDimension{Bool,Byte}ArrayNode()` to not use the default value.

andrewcraik:

cathyzhyi:
@fjeremic I pushed some changes for tracing after Andrew started the sanity tests. could you help to start it again? Thanks!

fjeremic:
Jenkins test sanity xlinux,win JDK8,JDK11

andrewcraik:
LGTM

PR #2913, opened 2018-09-18
Remove incorrect transformation of getAndSet

liqunl:
getAndSet returns the old value of the memory address instead of a
boolean. The correspondent atomic symbol hasn't been implemented, thus
remove the incorrent transformation for now.


liqunl:

liqunl:
The equivalent atomic primitive is implemented in https://github.com/eclipse/omr/pull/2174. Created https://github.com/eclipse/openj9/issues/2915 for adding the right mapping.

andrewcraik:
Just waiting for the CI build to finish

PR #2918, opened 2018-09-18
Improvements for Relocatable Compilations Part 1

dsouzai:
Depends on https://github.com/eclipse/omr/pull/2977

This commit introduces the Symbol Validation Manager infrastructure which allows the compiler to run previously (implicitly) disabled optimizations under AOT. The fundamental idea for the Symbol Validation Manager was thought up by @jdmpapin ; the current implementation is a collaboration between myself and @jdmpapin .

Issue tracking all the PRs: https://github.com/eclipse/openj9/issues/2921

Important point worth noting: all of the new AOT validations are NOT enabled by default; they're guarded by an option.

### DIFF
--- a/runtime/compiler/env/PersistentCHTable.cpp
+++ b/runtime/compiler/env/PersistentCHTable.cpp
@@ -312,7 +312,8 @@ TR_PersistentClassInfo *
 TR_PersistentCHTable::findClassInfoAfterLocking(
       TR_OpaqueClassBlock *classId,
       TR::Compilation *comp,
-      bool returnClassInfoForAOT)
+      bool returnClassInfoForAOT,
+      bool validateClassInfo)
    {
    if (comp->fej9()->isAOT_DEPRECATED_DO_NOT_USE() && !returnClassInfoForAOT) // for AOT do not use the class hierarchy
       return NULL;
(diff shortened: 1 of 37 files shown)


andrewcraik:
@dsouzai should this be WIP based on your previous comment?

dsouzai:
@mstoodle @jdmpapin could you please review?

dsouzai:
@mstoodle @jdmpapin could you please review?

runtime/compiler/runtime/RelocationRecord.cpp
```diff
@@ -1221,6 +1221,61 @@ TR_RelocationRecordConstantPoolWithIndex::getInterfaceMethodFromCP(TR_Relocation
...
+      J9Method *method;
+      vTableOffset = javaVM->internalVMFunctions->resolveVirtualMethodRefInto(reloRuntime->currentThread(),
+                                                                              cp,
+                                                                              cpIndex,
+                                                                              J9_RESOLVE_FLAG_AOT_LOAD_TIME,
+                                                                              &method,
+                                                                              NULL);
+      }
+
+   if (abstractClass)
```
jdmpapin: The method resolution might have failed even if the class resolution succeeded. In that case we would pass `J9JIT_INTERP_VTABLE_OFFSET` to `findSingleAbstractImplementer()`, at which point I'm not sure what would happen. It might be safest to check explicitly that both succeeded

runtime/compiler/env/PersistentCHTable.cpp
```diff
@@ -500,7 +500,7 @@ TR_PersistentCHTable::findSingleAbstractImplementer(
    {
    if (comp->getOption(TR_DisableCHOpts))
       return 0;
-   TR_PersistentClassInfo * classInfo = findClassInfoAfterLocking(thisClass, comp);
+   TR_PersistentClassInfo * classInfo = findClassInfoAfterLocking(thisClass, comp, true);
```
jdmpapin: It would be more conventional to introduce a default parameter, and pass it through here. However, it looks to me like you hit all of the callers except for [one in invariant argument preexistence](https://github.com/eclipse/omr/blob/02bfd61c0fd38b2b641828e879f0bc7ad062db8d/compiler/optimizer/LocalOpts.cpp#L7497), which I believe (currently) doesn't run for AOT anyway.

dsouzai:
> Should there be a case for TR_InlinedAbstractMethodWithNopGuard in TR_RelocationRecord::create()?

You are correct; it's there in a later changeset, but when I logically separated them, I guess I missed the creation... sigh; I"ll make the requested changes.

runtime/compiler/env/PersistentCHTable.cpp
dsouzai: I sort of followed what happens in findSingleInterfaceImplementer, which does something similar :P

dsouzai:
Soo... turns out Abstract Method Inlining only worked under AOT because it was running with the new AOT Validations that are introduced in Part 2. So unfortunately I'm going to have to punt over the Abstract Method Inlining change over to Part 2. 

(The issue I'm running into is a crash during class validation, which is probably because the cpIndex is -1 or something like that; the cpIndex being -1 isn't a problem after Part 2)

dsouzai:
@mstoodle friendly reminder for review :)

dsouzai:
In the interest of time, I'm closing this PR since https://github.com/eclipse/openj9/pull/2919 will contain all the relocatable compilation improvements.

PR #2919, opened 2018-09-18
Improvements for Relocatable Compilations

dsouzai:
Depends on https://github.com/eclipse/omr/pull/2978

This PR introduces changes in OpenJ9 required to enable optimizations that are currently implicitly disabled for relocatable compilations. It does so by introducing the Symbol Validation Manager infrastructure. The fundamental idea for the Symbol Validation Manager was thought up by @jdmpapin ; the current implementation is a collaboration between myself and @jdmpapin .

Issue tracking all the PRs: #2921

### DIFF
--- a/doc/compiler/aot/SymbolValidationManager.md
+++ b/doc/compiler/aot/SymbolValidationManager.md
@@ -1,0 +1,173 @@
+<!--
+Copyright (c) 2018, 2018 IBM Corp. and others
+
+This program and the accompanying materials are made available under
+the terms of the Eclipse Public License 2.0 which accompanies this
+distribution and is available at https://www.eclipse.org/legal/epl-2.0/
+or the Apache License, Version 2.0 which accompanies this distribution and
+is available at https://www.apache.org/licenses/LICENSE-2.0.
+
+This Source Code may also be made available under the following
+Secondary Licenses when the conditions for such availability set
+forth in the Eclipse Public License, v. 2.0 are satisfied: GNU
+General Public License, version 2 with the GNU Classpath
+Exception [1] and GNU General Public License, version 2 with the
+OpenJDK Assembly Exception [2].
+
+[1] https://www.gnu.org/software/classpath/license.html
+[2] http://openjdk.java.net/legal/assembly-exception.html
+
+SPDX-License-Identifier: EPL-2.0 OR Apache-2.0 OR GPL-2.0 WITH Classpath-exception-2.0 OR LicenseRef-GPL-2.0 WITH Assembly-exception
+-->
+
+# Symbol Validation Manager
+
+In order for the OpenJ9 Compiler to compile Ahead Of Time (AOT),
+or more accurately, Relocatable, code, it needs to be able to store
+information for the AOT load infrastructure to validate all the 
+assumptions made during the compile run, since the AOT load is 
+performed in a different JVM environment than the one the compilation
+was performed in. For these reasons, the Compiler and the Shared Class
+Cache (SCC) has infrastructure to facilate these validations.
+
+Without the Symbol Validation Manager, the validations involves either
+creating a relocation record to validate an Arbitrary Class, a Class
+from a Constant Pool, an Instance Field, or a Static Field. However,
+when getting a Class By Name, the validation stored would be an 
+Arbitrary Class Validation, which is technically incorrect when the
+Name is retrieved from the signature of a method (since the answer
+depends on what the Class of that method "sees"). In order to deal
+with this, the compiler needs to know where the "beholder" Class comes
+from. The Symbol Validation Manager fixes this problem by making the
+answer to the question "where does this _whatever_ come from?" explicit
+and unambigious.
+
+## How it works
+
+The Symbol Validation Manager (SVM) is very much similar to a symbol table
+that a linker might use. The idea is to store information about where
+a particular symbol (eg `J9Class`, `J9Method`, etc.) came from. For example,
+if the compiler got the `J9Class` of `java/lang/String` by name from the
+parameter list of method `C.foo(Ljava/lang/String)V`, then the SVM will
+store a record stating that the compiler got `J9Class` `java/lang/String`
+By Name as seen by `J9Class` `C`. Then in the load run, the load
+infrastructure can validate these answers the compiler got by redoing the
+same queries *in the same order as during the AOT compilation*. In this
+manner, if the AOT load succeeds validations, the compiled body is 100%
+guaranteed to be compatible in the current environment.
+
+However, since `J9Class`es are artifacts of the current environment, how
+does the SVM enable a new environment to validate it? The core idea is to
+associate unique IDs to each new symbol the compiler gets a hold of. Thus,
+when a symbol is acquired by asking a query using a different symbol, the
+validation record that gets generated refers to these two symbols by their
+IDs, which are environment agnostic. In the load run, the AOT load
+infrastructure uses these IDs to build up its table of symbols that are
+valid in _its_ environment.
+
+To give a concrete example:
+
+The compiler decides to compile `A.foo()`. As part of the compilation, it
+gets class `B` from `A`'s constant pool, gets class `C` from `A`'s 
+constant pool, and finally gets `B` by name as seen by `C`. Thus, the 
+validation records that get generated would be:
+
+1. Root Class Record for `A`; `A` gets ID 1
+2. Method From Class Record for `foo`; `foo` gets ID 2
+3. Class From CP of `A` for `B`; `B` gets ID 3
+4. Class From CP of `A` for `C`; `C` gets ID 4
+5. Class By Name as seen by `C`
+
+When writing out the records, the information would be written out in the
+following manner:
+
+1. Root Class Record: ID=1
+2. Method From Class Record: ID=2, classID=1, index=x
+3. Class From CP Record: ID=3, beholderID=1, cpIndex=y
+4. Class From CP Record: ID=4, beholderID=1, cpIndex=z
+5. Class By Name Record: ID=3, beholder=4
+
+Thus the information written in the validation records are entirely
+environment agnostic. In the load run, the AOT infrastructure will
+go through these validations and build up the symbols that are valid
+in _its_ environment:
+
+1. Map ID=1 to the class of the method being compiled
+2. Get the method at index=x from the class associated with ID=1 and
+map it to ID=2
+3. Get the class from cpIndex=y from the class associated with ID=1
+and map it to ID=3
+4. Get the class from cpIndex=z from the class associated with ID=1
+and map it to ID=4
+5. Get the class by name as seen by the class associated with ID=4
+and verify that it is the same class associated with ID=3
+
+If any of the validations fail, the load is aborted. Thus the general 
+approach in the load run is:
+
+1. Find the symbol as specified by the validation record
+2. Map the ID from the validation record to the symbol
+3. If the map already contains a symbol for that ID, validate that
+the symbol in the map is the same as the symbol currently being
+validated; fail otherwise
+
+
+## Subtleties
+
+### Class Chains
+
+Class Chains are a core component of AOT validation; they ensure that
+the shape of a class is the same as that during the compile run. For
+technical details about Class Chains, see 
+[AOT Class Chains](https://github.com/eclipse/openj9/blob/master/doc/compiler/aot/AOTClassChains.md).
+Thus, for every validation record involving classes, a Class Chain
+Validation Record is also created (unless a record for that class
+already exists).
+
+### Array Classes
+
+Array Classes create a problem for validation. Class Chains can only
+exist for non-array classes. However, if the compiler acquired an 
+array class by name or via profiling, in order to use it during an AOT
+compilation, it still needs to ensure that the shape of that class will
+be valid in the load run. Therefore, for every validation record 
+involving classes, what actually happens is the SVM will get the
+leaf component class, store the record as if the compiler acquired the
+leaf component class from the query, and then store as many 
+Array From Component Class Validation Records as there are array
+dimensions. Therefore, on the load run, when performing a validation, 
+the SVM also has ensure to get the leaf component class in order for
+the validation to be consistent with the compile run. The actual array
+class gets validated via the Array From Component Class records.
+
+### Primitive Classes
+
+Primitive Classes create a different problem for validation. As it turns
+out, the ROM Classes for primitive classes aren't stored in the SCC.
+This is an issue when storing Class By Name records, since the name
+portion of the validation is derived from the ROM Class. In order to
+deal with this problem, the Profiled Class and Class By Name 
+Validations also store a char value representing the primitive type;
+if the char value is '\0', then the class being validated isn't
+primitive and the rest of the validation process continues. If the
+char value is not '\0', then class is primitive, and the Profiled 
+Class or Class By Name record returns validaion succeeded, which is
+correct because Primitive Classes are guaranteed to always be the
+same regardless of how they are acquired.
+
+### Guaranteed IDs
+
+Because Primitive Classes (and their associated array classes) are always
+guaranteed to exist, and because the Root Class and method from root class
+have already been validated prior to even attempting to perfom the AOT
+load, these symbols don't need to be validation records stored in 
+the SCC. Therefore, the SVM, in its constructor, initializes the maps
+with these IDs and the associated symbols.
+
+## Benefits
+
+1. Makes explicit the providence of every symbol acquired by the compiler
+2. Enables code paths previously disabled under AOT
+3. Allows relocations of classes/methods associated with cpIndex=-1
+4. Facilitates the enablement of Known Object Table under AOT
+5. Facilitates the enablement of JSR292 under AOT
--- a/runtime/compiler/codegen/J9AheadOfTimeCompile.cpp
+++ b/runtime/compiler/codegen/J9AheadOfTimeCompile.cpp
@@ -27,6 +27,7 @@
 #include "il/DataTypes.hpp"
 #include "env/VMJ9.h"
 #include "codegen/AheadOfTimeCompile.hpp"
+#include "runtime/RelocationRecord.hpp"
 
 extern bool isOrderedPair(uint8_t reloType);
 
@@ -698,6 +699,7 @@ J9::AheadOfTimeCompile::dumpRelocationData()
          case TR_InlinedSpecialMethodWithNopGuard:
          case TR_InlinedVirtualMethodWithNopGuard:
          case TR_InlinedInterfaceMethodWithNopGuard:
+         case TR_InlinedAbstractMethodWithNopGuard:
          case TR_InlinedHCRMethod:
             cursor++;        // unused field
             if (is64BitTarget)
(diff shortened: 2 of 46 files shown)


runtime/compiler/env/VMJ9.cpp
```diff
@@ -8524,6 +8541,12 @@ TR_J9SharedCacheVM::isClassVisible(TR_OpaqueClassBlock * sourceClass, TR_OpaqueC
           TR_J9VMBase::isClassVisible(sourceClass, destClass);
    }
 
+bool
```
vijaysun-omr: Is there some inherent reason this optimization cannot be done under AOT or is it just something you are not handling at this point in time ?

runtime/compiler/x/codegen/J9AheadOfTimeCompile.cpp
```diff
@@ -211,23 +214,21 @@ uint8_t *J9::X86::AheadOfTimeCompile::initializeAOTRelocationHeader(TR::Iterated
...
-            {
-            // Setup flags field with type of method that needs to be validated at relocation time
-            if (guard->getSymbolReference()->getSymbol()->getMethodSymbol()->isStatic())
-               flags = inlinedMethodIsStatic;
-            if (guard->getSymbolReference()->getSymbol()->getMethodSymbol()->isSpecial())
-               flags = inlinedMethodIsSpecial;
-            if (guard->getSymbolReference()->getSymbol()->getMethodSymbol()->isVirtual())
-               flags = inlinedMethodIsVirtual;
-            }
+         // Setup flags field with type of method that needs to be validated at relocation time
```
vijaysun-omr: Does this get impacted in any way by nest mates in Java11 ? i.e. can we get the "type of method" reliably the way we are getting it here when that feature is in play ?

runtime/compiler/x/codegen/J9AheadOfTimeCompile.cpp
```diff
@@ -253,6 +252,23 @@ uint8_t *J9::X86::AheadOfTimeCompile::initializeAOTRelocationHeader(TR::Iterated
             resolvedMethod = guard->getSymbolReference()->getSymbol()->getResolvedMethodSymbol()->getResolvedMethod();
             }
 
+         if (comp->getOption(TR_UseSymbolValidationManager))
+            {
+            TR_OpaqueMethodBlock *method = resolvedMethod->getPersistentIdentifier();
+            TR_OpaqueClassBlock *thisClass = guard->getThisClass();
+            uint16_t methodID = symValManager->getIDFromSymbol(static_cast<void *>(method));
+            uint16_t receiverClassID = symValManager->getIDFromSymbol(static_cast<void *>(thisClass));
+
```
vijaysun-omr: Just checking why 16 bits for the ID is enough for every compiled method (with inlining) ?

runtime/compiler/x/codegen/J9AheadOfTimeCompile.cpp
```diff
@@ -457,8 +485,20 @@ uint8_t *J9::X86::AheadOfTimeCompile::initializeAOTRelocationHeader(TR::Iterated
          cursor += SIZEPOINTER;
          *(uintptrj_t *)cursor = (uintptrj_t) classSymRef->getOwningMethod(comp)->constantPool();
          cursor += SIZEPOINTER;
-         *(uintptrj_t *)cursor = (uintptrj_t) classSymRef->getCPIndex(); // cp Index
+
+         if (comp->getOption(TR_UseSymbolValidationManager))
+            {
+            TR_OpaqueClassBlock *clazz = (TR_OpaqueClassBlock *)recordInfo->data5;
+            uintptrj_t classID = (uintptrj_t)symValManager->getIDFromSymbol(static_cast<void *>(clazz));
```
vijaysun-omr: Is this ID comparable in any way to the 16 bit IDs I asked about earlier ? Now we are using uintptrj_t which is why I wanted to check.

vijaysun-omr:
Question about something you wrote in one of the commit messages:

```
For example, if the method being compiled is A.foo(), then before the
compilation begins, the compiler creates:

1. Root Class Record for A
2. Method From Class Record for foo()

This will result in A getting ID 1, and foo() getting ID 2. As the
compilation continues, if the compiler gets class B in two different
ways, namely once by name and once via CP, then two other records get
created:

3. Class by Name for B
4. Class from CP for B
```
What happens if we get class B by name again in some other spot in the compilation ? Will we create a separate record for that or are we going to detect that it is a record/query we have seen before i.e is the 3rd record above going to be the only one stored in that case ?

runtime/compiler/runtime/SymbolValidationManager.cpp
```diff
@@ -449,6 +475,8 @@ TR::SymbolValidationManager::addClassFromITableIndexCPRecord(TR_OpaqueClassBlock
    {
    if (!clazz)
       return false;
+   if (inHeuristicRegion())
```
vijaysun-omr: Do we really need to check for heuristic region here ? I thought you were suggesting that we only specify heuristic regions in parts of the code where we are looking at classes in some controlled way, e.g.iterate over subclasses when looking up a single implementor.

runtime/compiler/optimizer/J9ValuePropagation.cpp
```diff
@@ -617,13 +620,16 @@ J9::ValuePropagation::constrainRecognizedMethod(TR::Node *node)
...
 
                TR::KnownObjectTable *knot = comp()->getOrCreateKnownObjectTable();
-               TR::KnownObjectTable::Index knownObjectIndex = knot->getIndexAt((uintptrj_t*)(superClass + comp()->fej9()->getOffsetOfJavaLangClassFromClassField()));
-               addBlockOrGlobalConstraint(node,
-                     TR::VPClass::create(this,
-                        TR::VPKnownObject::createForJavaLangClass(this, knownObjectIndex),
-                        TR::VPNonNullObject::create(this), NULL, NULL,
-                        TR::VPObjectLocation::create(this, TR::VPObjectLocation::JavaLangClassObject)),
-                     classChildGlobal);
+               if (knot)
```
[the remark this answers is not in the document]
dsouzai: ugh, I guess I must've missed this when squashing changes. I'll fix it soon.

runtime/compiler/env/VMJ9.cpp
dsouzai: Oh you're viewing this commit by commit; in 5e6be87f7e9f4847742cbb2ebc8b8442b669caa4 (Add AOT Validations) is where I actually add the necessary validations to make sure this works under AOT

runtime/compiler/x/codegen/J9AheadOfTimeCompile.cpp
dsouzai: @jdmpapin do you know?

runtime/compiler/x/codegen/J9AheadOfTimeCompile.cpp
dsouzai: These IDs are per compilation; ID=1 for the compilation of method `A.foo()` is not going to match ID=1 for the compilation of method `B.bar()`. Thus, @jdmpapin and I figure that 64k unique symbols should be sufficient for any compilation.

runtime/compiler/x/codegen/J9AheadOfTimeCompile.cpp
dsouzai: This is one of those pieces of code that @mstoodle is going to very much dislike. With the symbol validation manager, aot relocations no longer need the cpIndex to find the symbol to validate. However, in order to reuse existing relocation records and the validation work that they already do (and not duplicate code), the new aot infrastructure uses the field that was traditionally used to store the cpIndex to instead store the ID for the symbol. Hence, even though the ID is only 16 bits, it gets stored in a uintptrj_t field. (see https://github.com/eclipse/openj9/pull/2919/files/151e307854fc2ae62b7d3a1a399da9c05644589b..f9c180ad226c0aee6d98e8356979586ea9b9ed61#diff-3f1aba0052032c44a8f47c99b97e30e2R263 for an even more (unfortunately) egregious example of this).

dsouzai:
> What happens if we get class B by name again in some other spot in the compilation ? Will we create a separate record for that or are we going to detect that it is a record/query we have seen before i.e is the 3rd record above going to be the only one stored in that case ?

Yes, we would create another record associated with that compilation. The reason is because if we acquired `B` by name via the parameter list of the method being compiled `A.foo(LB;)`, and then in a different compilation, acquired `B` also by name via the parameter of the method being compiled `C.bar(LB;)`, then in the load run, if `A` and `C` are loaded by different class loaders, it is possible for `A` and `C` not not agree on `B`. Therefore, even if `B` as seen by `A` and `C` have the same ROMClass and the same class chains, there could be two different `J9Class`s - one to describe `B` as seen by `A`, and the other to describe `B` as seen by `C`.

Having this seemingly redundant record to describe `B` by name is to guard against the case where `A` and `C` agree on `B` in the compile run, but disagree on `B` in the load run.

That said, if we were looking up bootstrap classes by name, then the records are definitely redundant because they are never ambiguous.

@jdmpapin anything to add/correct?

runtime/compiler/runtime/SymbolValidationManager.cpp
dsouzai: > I thought you were suggesting that we only specify heuristic regions in parts of the code where we are looking at classes in some controlled way, e.g.iterate over subclasses when looking up a single implementor.

Yes you're right. However, for the sake of consistency, all validation records check whether a heuristic region is active. it is also a proactive means of ensuring that future code that guards other heuristic regions don't have to worry about the kind of validations that might fail within that region since the manager guarantees that all validation records will pass within the region.

jdmpapin:
> What happens if we get class B by name again in some other spot in the compilation ? Will we create a separate record for that or are we going to detect that it is a record/query we have seen before i.e is the 3rd record above going to be the only one stored in that case ?

When looking up a class by name, we need to know how to interpret that name, which is determined by the defining loader of the class in which the name was found. There is no privileged reference frame, as it were. So suppose we found `B` by name from the perspective of some class `X`. We create a `TR_ValidateClassByName` record for it. If we later find the same `B` by name again, we may or may not create a new record. If it was again from the perspective of `X`, no new record is needed, since it would be a duplicate. OTOH, if it was from the perspective of a different class `Y`, we _will_ create a new record. The reason for the extra record is that it's a possibility that the two queries could disagree at load time, and we need to check that they don't.

We could get away with a single record per loader, i.e. only one record if `X` and `Y` had the same defining loader in the example above. However, that would require other validations to check that `X` and `Y` still have the same defining class loader, e.g. by tracking class loaders like we do classes in the symbol validation manager. Doing so might or might not be a win in terms of space, since we would need more records in simple cases where a class is found by name just once. The requirements imposed in order to load would be stricter as well (although probably not _meaningfully_ stricter).

jdmpapin:
> That said, if we were looking up bootstrap classes by name, then the records are definitely redundant because they are never ambiguous.

While certain classes (by name) can only be defined by the bootstrap loader, and any loader that successfully loads such a class does so by delegating to it, I believe a loader can refuse to delegate, and as a result fail to find the class. If so, then we may need validations just to check that the loaders of certain classes haven't started "hiding" the system class we're looking for.

runtime/compiler/x/codegen/J9AheadOfTimeCompile.cpp
jdmpapin: 64k should be enough (distinct symbols) for any one (AOT compilation) :sweat_smile:

vijaysun-omr: Okay the reason I asked was that it seemed like an ID gets assigned every time we compile time query something that is of relevance from a validation perspective. While the number of distinct symbols in a compilation may be well under 64k, it seemed to me that the number of IDs were more dependent on how many times we queried and I was wondering if in a large method with many passes of different optimizations we might get close to 64k queries. Is there some simple safety valve possible anyway to detect and fail the AOT compilation if the ID goes above 64k ?

runtime/compiler/codegen/J9CodeGenerator.cpp
```diff
@@ -2695,12 +2695,20 @@ J9::CodeGenerator::processRelocations()
          }
 
       TR::SymbolValidationManager::SymbolValidationRecordList &validationRecords = self()->comp()->getSymbolValidationManager()->getValidationRecordList();
```
vijaysun-omr: I would like @mpirvu to review this specific commit to make sure the checking to enable/disable is done at the right place and in the right way.

doc/compiler/aot/SymbolValidationManager.md
```diff
@@ -0,0 +1,173 @@
...
+the SCC. Therefore, the SYM, in its constructor, initializes the maps
+with these IDs and the associated symbols.
+
+## Benefits
+
+1. Makes explicit the providence of every symbol acquired by the compiler
+2. Enables code paths previously disabled under AOT
+3. Allows relocations of classes/methods associated with cpIndex=-1
+4. Facilitates the enablement of Known Object Table under AOT
+5. Facilitates the enablement of JSR292 under AOT
```
vijaysun-omr: Thanks for the descrption at this level documented here. Helps understand the motivations for people reading the PR changes.

vijaysun-omr:
I assume JSR 292 is not handled currently (but the design of this new infrastructure allows you handle it better down the line), and the way this is implemented is via the same code that was handling JSR 292 conservatively with the older infrastructure, right ?

runtime/compiler/x/codegen/J9AheadOfTimeCompile.cpp
dsouzai: >While the number of distinct symbols in a compilation may be well under 64k, it seemed to me that the number of IDs were more dependent on how many times we queried 

That's not true, if we queried for class `B` from the constant pool of class `C` at cpIndex `5`, then regardless of how many times times we ran that query, once a class has been assigned an ID, it is unique to that ID; there is a 1-1 mapping between an ID and a `J9Class`. Additionally, there is only record per unique query made. So, as in the example above, even if the query was made 10 times, there would only be one validation record stored, since we only need to check once whether class `B` can be retrieved from the constant pool of class `C` at cpIndex `5`. However, if we retrieved class `B` from the constant pool of class `C` at cpIndices `3`, `4`, and `5`, then we'd need three validation records. but the validation records would look something like this (assuming class `B` <--> ID 1 and class `C` <--> ID 2):

ClassFromCP,ID=2,BeholderID=1,cpIndex=3
ClassFromCP,ID=2,BeholderID=1,cpIndex=4
ClassFromCP,ID=2,BeholderID=1,cpIndex=5

All that said, I do have a check to ensure that we don't overflow the IDs just in case: https://github.com/eclipse/openj9/pull/2919/commits/9a11e2fa6df3fafe57023d05e55df92042f2ba74#diff-3e7f7c3bf56a075860c3736280f74b05R108

runtime/compiler/x/codegen/J9TreeEvaluator.cpp
```diff
@@ -3505,7 +3505,9 @@ inline void generateInlinedCheckCastOrInstanceOfForInterface(TR::Node* node, TR_
    {
    TR_ASSERT(clazz && TR::Compiler->cls.isInterfaceClass(cg->comp(), clazz), "Not a compile-time known Interface.");
 
-   auto use64BitClasses = TR::Compiler->target.is64Bit() && !TR::Compiler->om.generateCompressedObjectHeaders();
+   auto use64BitClasses = TR::Compiler->target.is64Bit() &&
```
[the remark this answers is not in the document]
dsouzai: There is no reason in general to force 64 bit classes for AOT. However, the relocation that gets generated in the DataSnippet is a "SymbolFromManager` record. During the load run, this relo record will add class unload / class redefinition assumptions. A 32-bit pointer requires a different API to register the assumptions and a different patching mechanism (which means a different relo record). In the interest of time and to just get things functional, I decided to take easy route and force 64 bit pointers under AOT for both checkcast inlining and object allocation inlining.

However, there is really no reason to continue do so; in the future we can just add the necessary relocation records to handle both 64-bit and 32-bit pointers.

dsouzai:
>Can isInterpretedForHeuristics be used to enable the "do not inline very large compiled methods" heuristic in the inliner ?

Actually the code that determines this, namely whatever calls `TR_MultipleCallTargetInliner::isLargeCompiledMethod` behaves the same on AOT compiles as it does for JIT'd compiles. This is because it doesn't use the `isInterpreted` query, but rather

```
void * methodAddress = calleeResolvedMethod->startAddressForInterpreterOfJittedMethod();
TR_PersistentJittedBodyInfo * bodyInfo = TR::Recompilation::getJittedBodyInfoFromPC(methodAddress);
```
if the `bodyInfo` is not NULL, then the method is compiled. `startAddressForInterpreterOfJittedMethod` does NOT return NULL in AOT, and instead returns the startPC of the method.

dsouzai:
> I assume JSR 292 is not handled currently (but the design of this new infrastructure allows you handle it better down the line), and the way this is implemented is via the same code that was handling JSR 292 conservatively with the older infrastructure, right ?

Yes, the control logic in CompilationThread.cpp is, where we filter out JSR292 compilations for AOT:
```
      eligibleForRelocatableCompile =
         TR::Options::sharedClassCache() &&
         !details.isNewInstanceThunk() &&
         !entry->isJNINative() &&
         !details.isMethodHandleThunk() &&
         !TR::CompilationInfo::isCompiled(method) &&
         !entry->isDLTCompile() &&
         !entry->_doNotUseAotCodeFromSharedCache &&
         reloRuntime->isRomClassForMethodInSharedCache(method, javaVM) &&
         (!TR::Options::getAOTCmdLineOptions()->getOption(TR_AOTCompileOnlyFromBootstrap) ||
           TR_J9VMBase::get(jitConfig, vmThread)->isClassLibraryMethod((TR_OpaqueMethodBlock *)method), true);
```
specifically the condition `!details.isMethodHandleThunk() `

runtime/compiler/x/codegen/J9AheadOfTimeCompile.cpp
vijaysun-omr: Okay thanks for the clarification and the check for overflow.

runtime/compiler/codegen/J9CodeGenerator.cpp
```diff
@@ -2695,12 +2695,20 @@ J9::CodeGenerator::processRelocations()
...
-         self()->addExternalRelocation(new (self()->trHeapMemory()) TR::ExternalRelocation(NULL,
-                                                                          (uint8_t *)(*it),
-                                                                          (*it)->_kind, self()),
-                                                                          __FILE__, __LINE__, NULL);
+         // Add the flags in TR_AOTMethodHeader on the compile run
+         J9JITDataCacheHeader *aotMethodHeader = (J9JITDataCacheHeader *)self()->comp()->getAotMethodDataStart();
+         TR_AOTMethodHeader *aotMethodHeaderEntry = (TR_AOTMethodHeader *)(aotMethodHeader + 1);
+         aotMethodHeaderEntry->flags |= TR_AOTMethodHeader_UsesSymbolValidationManager;
+
+         for (auto it = validationRecords.begin(); it != validationRecords.end(); it++)
```
mpirvu: Don't we have to add the relocations even if TR_UseSymbolValidationManager is false as well?

runtime/compiler/control/CompilationThread.cpp
```diff
@@ -7414,6 +7414,11 @@ TR::CompilationInfoPerThreadBase::wrappedCompile(J9PortLibrary *portLib, void *
                   options->setOption(TR_EnableGRACostBenefitModel, false);
                }
 
+            if (jitConfig->javaVM->phase != J9VM_PHASE_NOT_STARTUP || options->getOptLevel() < warm)
+               {
```
mpirvu: Assuming the symbolValidationManager is the default, maybe turning it off for AOT loads does not make sense. Let the AOT load do whatever it needs to do based on how the code was generated. This will eliminate the need to set this option later in prepareRelocateAOTCodeAndData

runtime/compiler/control/CompilationThread.cpp
```diff
@@ -7070,6 +7070,12 @@ TR::CompilationInfoPerThreadBase::wrappedCompile(J9PortLibrary *portLib, void *
                options->setOption(TR_EnableAnnotations,false);
             }
 
+         if (vm->canUseSymbolValidationManager() && options->getOption(TR_EnableSymbolValidationManager))
+            {
```
mpirvu: Now I realize that to control this feature externally one has to use the TR_EnableSymbolValidationManager option. This makes TR_UseSymbolValidationManager truly a flag and not an option. Maybe we should put this into flags32_t _flags; in OMRCompilation.

runtime/compiler/env/VMJ9.cpp
```diff
@@ -8536,15 +8545,44 @@ TR_J9SharedCacheVM::isClassVisible(TR_OpaqueClassBlock * sourceClass, TR_OpaqueC
...
+      }
+
+   return (validated ? TR_J9VMBase::isClassVisible(sourceClass, destClass) : false);
    }
 
 bool
 TR_J9SharedCacheVM::stackWalkerMaySkipFrames(TR_OpaqueMethodBlock *method, TR_OpaqueClassBlock *methodClass)
    {
-   return false;
+   bool skipFrames = TR_J9VM::stackWalkerMaySkipFrames(method, methodClass);
```
mstoodle: maybe move this code under `if (comp && comp->getOption(TR_UseSymbolValidationManager))`j ?

runtime/compiler/env/VMJ9.cpp
```diff
@@ -8536,15 +8545,44 @@ TR_J9SharedCacheVM::isClassVisible(TR_OpaqueClassBlock * sourceClass, TR_OpaqueC
...
+   bool validated = false;
+   TR::Compilation *comp = TR::comp();
+   if (comp && comp->getOption(TR_UseSymbolValidationManager))
+      {
+      if (comp->getSymbolValidationManager()->addStackWalkerMaySkipFramesRecord(method, methodClass, skipFrames))
+         validated = true;
+      else
+         TR_ASSERT_FATAL(false, "Failed to validate addStackWalkerMaySkipFramesRecord\n");
```
mstoodle: Is this really such a dire failure condition? Couldn't we just return false in this case?

runtime/compiler/env/VMJ9.cpp
```diff
@@ -8536,15 +8545,44 @@ TR_J9SharedCacheVM::isClassVisible(TR_OpaqueClassBlock * sourceClass, TR_OpaqueC
...
+      }
+
+   if (validated)
```
mstoodle: i find this to be more convoluted than required with two booleans...

runtime/compiler/env/VMJ9.cpp
```diff
@@ -8607,7 +8645,22 @@ TR_J9SharedCacheVM::getInstanceFieldOffset(TR_OpaqueClassBlock * classPointer, c
    TR::Compilation* comp = _compInfo->getCompInfoForCompOnAppThread() ? _compInfo->getCompInfoForCompOnAppThread()->getCompilation() : _compInfoPT->getCompilation();
    TR_ASSERT(comp, "Should be called only within a compilation");
 
-   if (((TR_ResolvedRelocatableJ9Method*) comp->getCurrentMethod())->validateArbitraryClass(comp, (J9Class*)classPointer))
+   bool validated = false;
+
+   if (comp->getOption(TR_UseSymbolValidationManager))
+      {
+      if (comp->getSymbolValidationManager()->inHeuristicRegion() ||
```
mstoodle: when does one know to insert these `inHeuristicRegion()` calls, since they seem to be in some places but not others? It's a shame we couldn't find a way to sink them all inside the SVM or hide them behind existing queries.

runtime/compiler/env/VMJ9.cpp
```diff
@@ -8642,8 +8720,40 @@ TR_J9SharedCacheVM::getResolvedMethods(TR_Memory *trMemory, TR_OpaqueClassBlock
...
+          comp->getSymbolValidationManager()->getIDFromSymbol(static_cast<void *>(classPointer)))
+         validated = true;
+      else
+         TR_ASSERT_FATAL(false, "Class 0x%p should already be validated\n", classPointer);
+      }
+   else
+      {
+      validated = ((TR_ResolvedRelocatableJ9Method *) comp->getCurrentMethod())->validateArbitraryClass(comp, (J9Class *) classPointer);
```
mstoodle: given that the J9Class * is a parameter, I'm not sure why we need to validate it here... No code in the compiler should do anything with a J9Class * in an AOT compilation unless it's been validated already (and, in this spot, you don't really know that it should be an arbitrary class validation anyway because you don't know where it came from).

mstoodle: though i admit, the original code is broken in the same way...

runtime/compiler/env/VMJ9.cpp
mstoodle: it seems to be mostly in places where it's just trying to validate that a class has already been validated....maybe the symbol validation manager needs a `verifySymbolHasID(void *)` service to avoid exposing the heuristic region stuff?

runtime/compiler/control/CompilationThread.cpp
dsouzai: This code is guarded by `if (!that->_methodBeingCompiled->_aotCodeToBeRelocated) // exclude AOT loads` so it never runs under AOT loads. In fact, on an AOT load, regardless of whether we want to use the new infra or not, if the body we're relocating has validations from the new infra, we have to validate it using the new infra (hence the addition of the flag in the AOT Method Header).

runtime/compiler/control/CompilationThread.cpp
```diff
@@ -2846,6 +2846,8 @@ void TR::CompilationInfo::stopCompilationThreads()
       fprintf(stderr, "numVirtualMethodsValidationSucceeded: %d\n", aotStats->virtualMethods.numSucceededValidations);
       fprintf(stderr, "numInterfaceMethodsValidationFailed: %d\n", aotStats->interfaceMethods.numFailedValidations);
       fprintf(stderr, "numInterfaceMethodsValidationSucceeded: %d\n", aotStats->interfaceMethods.numSucceededValidations);
+      fprintf(stderr, "numAbstractMethodsValidationFailed: %d\n", aotStats->abstractMethods.numFailedValidations);
```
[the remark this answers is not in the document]
dsouzai: Initially this was part of a separate commit to enable abstract method inlining. However, if I enabled it in the old infrastructure, it started crashing. Hence, I grouped this change along with the change to add the infra to allow abstract method inlining.

doc/compiler/aot/SymbolValidationManager.md
```diff
@@ -0,0 +1,173 @@
...
+Name is retrieved from the signature of a method (since the answer
+depends on what the Class of that method "sees"). In order to deal
+with this, the compiler needs to know where the "beholder" Class comes
+from. The Symbol Validation Manager fixes this problem by making the
+answer to the question "where does this _whatever_ come from?" explicit
+and unambigious.
+
+## How it works
+
+The Symbol Validation Manager (SYM) is very much similar to a symbol table
```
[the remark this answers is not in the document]
dsouzai: Yes it should :man_facepalming:

doc/compiler/aot/SymbolValidationManager.md
```diff
@@ -0,0 +1,173 @@
...
+and verify that it is the same class associated with ID=3
+
+If any of the validations fail, the load is aborted. Thus the general 
+approach in the load run is:
+
+1. Find the symbol as specified by the validation record
+2. Map the ID from the validation record to the symbol
+3. If the map already contains a symbol for that ID, validate that
+the symbol in the map is the same as the symbol currently being
+validated; fail otherwise
```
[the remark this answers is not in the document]
dsouzai: Profiled Classes are handled by the `ProfiledClassRecord`, which performs the validation in the same manner as  `TR_AribitraryClassValidation`, namely the class chain + class chain of the first class loaded by the class loader validation.

A method can only be derived from a class, so if we have a profiled method, we must have a validation for its class as well as a `MethodFromClassRecord` (or some other `MethodFrom*`) record.

runtime/compiler/env/PersistentCHTable.cpp
```diff
@@ -408,8 +419,27 @@ TR_ResolvedMethod * TR_PersistentCHTable::findSingleImplementer(
...
+      {
+      TR::SymbolValidationManager *svm = comp->getSymbolValidationManager();
+
+      bool validated = svm->addMethodFromSingleImplementerRecord(implementer->getPersistentIdentifier(),
+                                                                 thisClass,
+                                                                 cpIndexOrVftSlot,
+                                                                 callerMethod->getPersistentIdentifier(),
+                                                                 useGetResolvedInterfaceMethod);
+      if (validated)
+         validated = svm->addClassFromMethodRecord(implementer->classOfMethod(), implementer->getPersistentIdentifier());
```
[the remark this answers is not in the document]
dsouzai: There shouldn't be, since every single `addMethodFrom*` method calls `addClassFromMethodRecord`.

runtime/compiler/env/VMJ9.cpp
dsouzai: Yes we should, for the sake of not introducing stupid bugs, all of the fatal asserts were replaced with compilation failures in https://github.com/eclipse/openj9/pull/2919/commits/f51192f3cbab50d521394e55e9932c8deae063e7

dsouzai: We can't just return false because it would give the compiler information that it does not necessarily  reflect the state of the environment. From what @jdmpapin told me, this is something that has to be consistent with the VM. If we compile code with the assumption that the stack walker may NOT skip frames (which is what returning false would imply), if the VM is running with the knowledge that the stack walker MAY skip frames, then we could run into issues.

runtime/compiler/env/VMJ9.cpp
dsouzai: It is also incorrect code, since under the "old" aot infra it would always return false :/ I'll fix this.

runtime/compiler/env/VMJ9.cpp
dsouzai: `validated = ((TR_ResolvedRelocatableJ9Method *) comp->getCurrentMethod())->validateArbitraryClass(comp, (J9Class *) classPointer);` is what the code was doing to begin with, so again, in the interest of minimizing the number of changes, I left it as is. You're right though, it shouldn't require validation (hence why the new code doesn't add a new record but rather checks that that symbol was already validated).

runtime/compiler/env/VMJ9.cpp
dsouzai: > maybe the symbol validation manager needs a verifySymbolHasID(void *) service to avoid exposing the heuristic region stuff

Yeah that would be much cleaner

runtime/compiler/env/VMJ9.cpp
dsouzai: As I realized in https://github.com/eclipse/openj9/pull/2919#discussion_r221742877 , I should not have been returning false to begin with; in the existing code, `stackWalkerMaySkipFrames` should not have returned `false` under "old" AOT; I've fixed this (but haven't pushed the changes yet)

runtime/compiler/runtime/SymbolValidationManager.hpp
```diff
@@ -0,0 +1,1337 @@
...
+#include "j9.h"
+#include "j9nonbuilder.h"
+#include "infra/TRlist.hpp"
+#include "env/TRMemory.hpp"
+#include "env/VMJ9.h"
+#include "runtime/Runtime.hpp"
+
+namespace TR {
+
+struct SymbolValidationRecord
```
mstoodle: are these types needed outside of SymbolValidationManager? If not, I'd suggest declaring them inside SymbolValidationManager.cpp or at least non-public inside SymbolValidationManager

runtime/compiler/runtime/SymbolValidationManager.cpp
```diff
@@ -0,0 +1,2006 @@
...
+   if (cpIndex != -1)
+      classCPIndexOfFieldOrStatic = ((J9ROMFieldRef *)(&romCPBase[cpIndex]))->classRefCPIndex;
+
+   J9Class *definingClass = NULL;
+   J9Class *cpClass = (J9Class*)TR_ResolvedJ9Method::getClassFromCP(fej9, beholderCP, comp, classCPIndexOfFieldOrStatic);
+
+   if (cpClass)
+      {
+      TR::VMAccessCriticalSection getDeclaringClassFromFieldOrStatic(fej9);
+
```
mstoodle: Just picking a random example here. It seems like this code might duplicate some logic present in the JIT somewhere....what's the model for keeping the conditions in this code maintained alongside the equivalent logic elsewhere? e.g. if someone finds a bug in code that looks up the declaring class, do we now need to fix that in more than one place? I don't have any specific examples, but much of this code, for example, "feels" like code in the front end...

runtime/compiler/env/VMJ9.cpp
```diff
@@ -8642,8 +8720,40 @@ TR_J9SharedCacheVM::getResolvedMethods(TR_Memory *trMemory, TR_OpaqueClassBlock
...
+      }
+
+   if (validated)
+      {
+      if (comp->getOption(TR_UseSymbolValidationManager))
+         {
+         TR::VMAccessCriticalSection getResolvedMethods(this); // Prevent HCR
```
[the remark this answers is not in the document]
dsouzai: Hm, I don't think it is any more or less sensitive than the previous infrastructure. I believe that when classes get redefined, we abort all existing compilations and purge the compilation queue. Any inlined method, or class/method pointers get the necessary assumptions registered during the load phase.

runtime/compiler/env/VMJ9.cpp
```diff
@@ -8666,12 +8796,23 @@ TR_J9SharedCacheVM::getMethodFromName(char *className, char *methodName, char *s
...
-      J9Class* methodClass = (J9Class*)getClassFromMethodBlock(omb);
-
       TR::Compilation* comp = _compInfo->getCompInfoForCompOnAppThread() ? _compInfo->getCompInfoForCompOnAppThread()->getCompilation() : _compInfoPT->getCompilation();
       TR_ASSERT(comp, "Should be called only within a compilation");
-      if (!((TR_ResolvedRelocatableJ9Method*) comp->getCurrentMethod())->validateArbitraryClass(comp, methodClass))
-         omb = NULL;
+
+      if (comp->getOption(TR_UseSymbolValidationManager))
+         {
+         bool validated = comp->getSymbolValidationManager()->addMethodByNameRecord(omb, getClassFromMethodBlock(callingMethod)) &&
```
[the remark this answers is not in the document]
dsouzai: I don't think so; in fact I believe that most `addMethodBy*` records also calls `addClassFromMethodRecord` just to be sure.

runtime/compiler/env/VMJ9.cpp
```diff
@@ -8780,102 +8967,259 @@ TR_J9SharedCacheVM::getProfiledClassFromProfiledInfo(TR_ExtraAddressInfo *profil
...
 bool
 TR_J9SharedCacheVM::isPublicClass(TR_OpaqueClassBlock * classPointer)
    {
    TR::Compilation* comp = _compInfo->getCompInfoForCompOnAppThread() ? _compInfo->getCompInfoForCompOnAppThread()->getCompilation() : _compInfoPT->getCompilation();
    TR_ASSERT(comp, "Should be called only within a compilation");
-   if (((TR_ResolvedRelocatableJ9Method *) comp->getCurrentMethod())->validateArbitraryClass(comp, (J9Class *) classPointer))
-      return TR_J9VM::isPublicClass(classPointer);
 
-   return true;
+   bool publicClass = TR_J9VM::isPublicClass(classPointer);
```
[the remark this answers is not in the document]
dsouzai: Ah I see; I added it this code because of the existing call to `validateArbitraryClass`

runtime/compiler/x/codegen/J9AheadOfTimeCompile.cpp
mstoodle: yeah....yuck confirmed...

runtime/compiler/env/j9method.cpp
```diff
@@ -1637,6 +1661,9 @@ TR_ResolvedRelocatableJ9Method::getResolvedPossiblyPrivateVirtualMethod(
          ignoreRtResolve,
          unresolvedInCP);
 
+   if (comp->getOption(TR_UseSymbolValidationManager))
```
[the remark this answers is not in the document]
dsouzai: Because the validation was added in `TR_ResolvedJ9Method::getResolvedPossiblyPrivateVirtualMethod`. It had to be put in there because the validation had to be added before the call `resolvedMethod = createResolvedMethodFromJ9Method(comp, cpIndex, vTableOffset, ramMethod, unresolvedInCP, aotStats);` which will result in an aborted compilation if, in the constructor of `TR_ResolvedRelocatableJ9Method`, the j9method wasn't already validated. Probably some opportunity for cleanup.

runtime/compiler/env/j9method.cpp
```diff
@@ -1793,7 +1820,34 @@ TR_ResolvedRelocatableJ9Method::fieldAttributes(TR::Compilation * comp, int32_t
       if (comp->getOption(TR_DisableAOTInstanceFieldResolution))
```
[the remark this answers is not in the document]
dsouzai: heh, really no reason other than "that's what the code looked like before". It's not a good answer, and, if you think it's something that should be fixed now rather than later I can make the necessary changes.

runtime/compiler/x/codegen/CallSnippet.cpp
```diff
@@ -480,7 +482,7 @@ TR_Debug::print(TR::FILE *pOutFile, TR::X86PicDataSnippet *snippet)
       printLabelInstruction(pOutFile, "jmp", snippet->getDoneLabel());
       bufferPos += 5;
 
-      if (methodSymRef->isUnresolved())
+      if (methodSymRef->isUnresolved() || fej9->forceUnresolvedDispatch())
```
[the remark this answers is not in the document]
dsouzai: Actually in this case, the debug code would sometimes fail to print out the data snippet data under AOT (because it only checked the unresolved condition, and not the "is AOT" condition).

runtime/compiler/codegen/J9AheadOfTimeCompile.cpp
```diff
@@ -1079,9 +1081,729 @@ J9::AheadOfTimeCompile::dumpRelocationData()
...
+            if (is64BitTarget)
+               cursor += 4;     // padding
+            cursor -= sizeof(TR_RelocationRecordBinaryTemplate);
+            TR_RelocationRecordValidateClassByNameBinaryTemplate *binaryTemplate =
+                  reinterpret_cast<TR_RelocationRecordValidateClassByNameBinaryTemplate *>(cursor);
+            if (isVerbose)
+               {
+               traceMsg(self()->comp(), "\n Validate Class By Name: classID=%d beholderID=%d primitiveType=%c romClassOffsetInSCC=%p ",
+                        (uint32_t)binaryTemplate->_classID,
+                        (uint32_t)binaryTemplate->_beholderID,
```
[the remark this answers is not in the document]
dsouzai: I'm not wedded to the name 'beholder' by any means; it came about because when @jdmpapin and I were discussing the semantics of the SVM, we kept talking about the idea of acquiring a symbol "as seen by" another symbol (eg Class By Name), and so "beholder" was the word we ended up using. Any name that conveys the idea that the symbol being validated depends on, or is acquired via, this other symbol is fine by me.

runtime/compiler/codegen/J9AheadOfTimeCompile.cpp
```diff
@@ -1079,9 +1081,729 @@ J9::AheadOfTimeCompile::dumpRelocationData()
...
+            cursor -= sizeof(TR_RelocationRecordBinaryTemplate);
+            TR_RelocationRecordSymbolFromManagerBinaryTemplate *binaryTemplate =
+                  reinterpret_cast<TR_RelocationRecordSymbolFromManagerBinaryTemplate *>(cursor);
+            if (isVerbose)
+               {
+               traceMsg(self()->comp(), "\n Symbol From Manager: symbolID=%d symbolType=%d ",
+                        (uint32_t)binaryTemplate->_symbolID, (uint32_t)binaryTemplate->_symbolType);
+               }
+            cursor += sizeof(TR_RelocationRecordSymbolFromManagerBinaryTemplate);
+            self()->traceRelocationOffsets(cursor, offsetSize, endOfCurrentRecord, orderedPair);
```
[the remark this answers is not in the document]
dsouzai: Yeah, a lot of the debugging I did heavily depended on this debug output. Additionally, @jdmpapin I believe painstakingly went through these.

runtime/compiler/control/CompilationThread.cpp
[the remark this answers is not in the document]
dsouzai: Heh, if it makes you feel any better, abstract method inlining definitely works with the new infrastructure, because it was the source a significant amount of performance gain.

dsouzai:
>just picking on one example at random because the inconsistency is starting to get to me: this is a spot where you don't verify that the incoming classObject parameter has a valid ID, so it seems pretty random whether you trust incoming J9Class (TR_OpaqueClassBlock) pointers or not...
>
>To my mind, anything that produces a J9Class (or TR_OpaqueClassBlock) pointer from something else needs to validate it or immediately change it to NULL. Anything that is handed a J9Class pointer should verify that class has an ID.
>
>We should try to be as consistent about that as possible.

Most of the places where the verification of the incoming symbol appears to missing is because that incoming symbol is passed into the `add*Record` which always verifies the symbols that are required to validate the symbol_to_be_validated. Thus, there is no need to add a second check in the various front end queries.

That said, there are few places where the incoming symbols don't get fed into the `add*Record` calls; I'll add code to verify those in order to maintain consistency.

runtime/compiler/runtime/SymbolValidationManager.cpp
dsouzai: > what's the model for keeping the conditions in this code maintained alongside the equivalent logic elsewhere?

I want to remove duplicate versions of code like this and have one canonical version. Time constraints resulted in duplicated versions, but when I wrote this out, I was cognizant of the fact that in the future I would have to consolidate all of this in order to prevent exactly the situation you're worried about.

runtime/compiler/x/codegen/CallSnippet.cpp
```diff
@@ -1064,7 +1080,12 @@ uint32_t TR::X86CallSnippet::getLength(int32_t estimatedSnippetStart)
       length += codeSize;
       }
 
-   if (methodSymRef->isUnresolved() || fej9->forceUnresolvedDispatch())
+   TR::Compilation *comp = cg()->comp();
```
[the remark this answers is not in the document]
dsouzai: I have a commit (https://github.com/dsouzai/openj9/commit/3146ec2dec6dd72095d13e5570f6d349a491c53a) that attempts to do so; however, it's not something I'm comfortable delivering in the time requirements given that it requires some refactoring work in CallSnippet. Initial tests seem to be fine, but I'd rather leave this till the cleanup work in the coming months.

dsouzai:
@mstoodle @vijaysun-omr Rebased and updated with review changes.

mstoodle:
Jenkins test sanity xlinux,win,win32 jdk8,jdk11 depends eclipse/omr#2978

mstoodle:
jenkins test sanity all jdk8 depends eclipse/omr#2978

mstoodle:
I've assigned myself for the moment just because I've had a lot to say about the commits in this PR and was the one requesting test results. However, I'm away on vacation tomorrow through Tuesday, so I'm leaving this comment to say that I am perfectly ok with another committer "stealing" this PR from me to merge if I don't get it merged today because I know we're pushing hard to make sure this goes into the 0.11 release slated for not very far away.

Note that #2921 documents many outstanding changes identified in this PR that should be scheduled to be done post 0.11 and ideally many of them landing in time for 0.12 .

mstoodle:
Bah, I shouldn't have listed the dependent PR since it's been merged :( .

jenkins test sanity all jdk8

mstoodle:
I was hoping to do another once over on these changes before merging, but the specific comments I requested to be resolved have been, and I will now be away until Tuesday. For that reason, I'm going to delegate my assignment to @mpirvu to decide that this pull request is truly ready to be merged (after having it fully tested once the dependent OMR PR makes its way into `openj9-omr`). On the off chance it hasn't been merged by Tuesday, I am willing to "take it back" to drive to completion.

mstoodle:
Sorry to dump it on you, @mpirvu. If Marius is away or also unable to devote enough time, then I'll ask @vijaysun-omr to either do himself or find another appropriate delegate. Thanks, guys!

mpirvu:
jenkins is misbehaving (since yesterday ) due to an infra bug and Eclipse has not yet acted on it
https://bugs.eclipse.org/bugs/show_bug.cgi?id=539800

runtime/compiler/runtime/SymbolValidationManager.cpp
```diff
@@ -0,0 +1,2329 @@
...
+   int32_t arrayDims = 0;
+   childClass = getBaseComponentClass(childClass, arrayDims);
+
+   if (getIDFromSymbol(superClass) == 0)
+      {
+      TR_ASSERT(false, "superClass 0x%p should have already been validated\n", superClass);
+      TR::comp()->failCompilation<J9::AOTSymbolValidationManagerFailure>("Failed to validate in addConcreteSubClassFromClassRecord");
+      }
+
+   SymbolValidationRecord *record = new (_region) SuperClassFromClassRecord(childClass, superClass);
```
jdmpapin: Should this be a `ConcreteSubClassFromClassRecord`?

dsouzai: :scream: yes it should...

dsouzai: For any reviewers reading this; the consequence of this error isn't anything functional except for that it would result in an AOT load failure.

mpirvu:
jenkins test sanity all jdk8

mpirvu:
ci shows failures:
```
clang++-3.8 -std=c++0x -w -fsyntax-only -ferror-limit=0  -Xclang -load -Xclang /home/travis/build/eclipse/openj9/runtime/omr/tools/compiler/OMRChecker/OMRChecker.so -Xclang -add-plugin -Xclang omr-checker   -DBITVECTOR_BIT_NUMBERING_MSB -DUT_DIRECT_TRACE_REGISTRATION -DJ9_PROJECT_SPECIFIC -DTR_HOST_X86 -DTR_HOST_64BIT -DBITVECTOR_64BIT -DLINUX -DTR_TARGET_X86 -DTR_TARGET_64BIT -DSUPPORTS_THREAD_LOCAL -D_LONG_LONG -DJ9HAMMER -I'/home/travis/build/eclipse/openj9/runtime/omr/include_core' -I'/home/travis/build/eclipse/openj9/runtime/omr/thread' -I'/home/travis/build/eclipse/openj9/runtime/omr/gc/include' -I'../compiler/x/amd64' -I'../compiler/x' -I'../compiler' -I'../omr/compiler/x/amd64' -I'../omr/compiler/x' -I'../omr/compiler' -I'../omr' -I'..' -I'/home/travis/build/eclipse/openj9/runtime/codert_vm' -I'/home/travis/build/eclipse/openj9/runtime/gc_include' -I'/home/travis/build/eclipse/openj9/runtime/include' -I'/home/travis/build/eclipse/openj9/runtime/gc_glue_java' -I'/home/travis/build/eclipse/openj9/runtime/jit_vm' -I'/home/travis/build/eclipse/openj9/runtime/nls' -I'/home/travis/build/eclipse/openj9/runtime/oti' -I'/home/travis/build/eclipse/openj9/runtime/util' -I'/home/travis/build/eclipse/openj9/runtime/build' -I'/home/travis/build/eclipse/openj9/runtime/build/omr' -o ../compiler/codegen/J9AheadOfTimeCompile.cpp.linted -c ../compiler/codegen/J9AheadOfTimeCompile.cpp
In file included from ../compiler/codegen/CodeGenGC.cpp:32:
In file included from ../compiler/x/amd64/codegen/J9CodeGenerator.hpp:41:
In file included from ../compiler/x/codegen/J9CodeGenerator.hpp:26:
In file included from ../compiler/codegen/J9CodeGenerator.hpp:36:
In file included from ../omr/compiler/x/amd64/codegen/OMRCodeGenerator.hpp:36:
In file included from ../omr/compiler/x/codegen/OMRCodeGenerator.hpp:34:
In file included from ../omr/compiler/codegen/OMRCodeGenerator.hpp:45:
In file included from ../compiler/compile/Compilation.hpp:26:
../compiler/compile/J9Compilation.hpp:212:11: error: Implicit this receivers are prohibited in extensible classes
      if (getOption(TR_UseSymbolValidationManager)
          ^
          self()->
```
and many others just like this one

dsouzai:
Sigh, that's probably the linter complaining; it's because I'm not doing `self()->getOption(...)`. I guess there's multiple failures because it's in a hpp file. I won't be able to fix this till probably tomorrow night or maybe even Monday evening (seeing as it's Thanksgiving weekend).

dsouzai:
@mpirvu I moved the definitions of `J9::Compilation::enterHeuristicRegion`, `J9::Compilation::exitHeuristicRegion`, and `J9::Compilation::validateTargetToBeInlined` to the .cpp file (since `self()` can't be used in a .hpp file (because the `TR` namespace isn't resolved at that point)), and I updated the calls to use `self()->`.

dsouzai:
Omg...
```
../compiler/compile/J9Compilation.cpp:173:74: error: Implicit this receivers are prohibited in extensible classes
   _symbolValidationManager = new (region()) TR::SymbolValidationManager(region(), compilee);
                                                                         ^
                                                                         self()->
../compiler/compile/J9Compilation.cpp:173:36: error: Implicit this receivers are prohibited in extensible classes
   _symbolValidationManager = new (region()) TR::SymbolValidationManager(region(), compilee);
                                   ^
                                   self()->
2 errors generated.
```
Missing `self`on the `region()`calls... Ugh, should've just used the member variable directly.

I won't have access to the code till the evening so I'll fix it then (checking this from my phone).

mpirvu:
jenkins test sanity all jdk8

mpirvu:
All checks have passed, so merging.

Issue #2932, opened 2018-09-19
sha1 checksum is fragile


navyxliu:
openj9/test/TestConfig/scripts/tools/getDependencies.pl
my %asmtools = (
	url => 'https://ci.adoptopenjdk.net/view/Dependencies/job/asmtools/lastSuccessfulBuild/artifact/asmtools.jar',
	fname => 'asmtools.jar',
	sha1 => '04cf07c584121c2e5a3d1dad2839fc8ab4828b6d'
);
use lastSuccessfulBuild is very fragile.  Actually, I found a new built succeeded. sha1 checksum changed then. 

I think you need to specify a fixed built number. eg. 107

sophia-guo:
The fixed build number should be stable. The only concern is the specified build could be discarded when more new builds come up. Ideally we should get the jar from more stable resources.

sophia-guo:
I'm not sure which is a better solution.

navyxliu:
that's okay. I also use built 107.

llxia:
The problem is that Jenkins does not keep the build history forever. Based on the history, built 107 will only be there for another 3 months.

I would like to keep this issue open as we need to find a more stable solution.

navyxliu:
sure. no problem.

smlambert:
We could look at having these tools published to the AdoptOpenJDK website (like the binaries) so that we can pull them from there (and get latest published tools from website)

PR #2933, opened 2018-09-19
Download asmtools from the stable build

sophia-guo:
The SHA changes when there is new available one, which we need to update
the SHA and test if the jar works.

Issue #2932 

[ci skip]

### DIFF
--- a/test/TestConfig/scripts/tools/getDependencies.pl
+++ b/test/TestConfig/scripts/tools/getDependencies.pl
@@ -106,7 +106,7 @@ my %jcommander = (
 	sha1 => 'bfcb96281ea3b59d626704f74bc6d625ff51cbce'
 );
 my %asmtools = (
-	url => 'https://ci.adoptopenjdk.net/view/Dependencies/job/asmtools/lastSuccessfulBuild/artifact/asmtools.jar',
+	url => 'https://ci.adoptopenjdk.net/view/Dependencies/job/asmtools/107/artifact/asmtools.jar',
 	fname => 'asmtools.jar',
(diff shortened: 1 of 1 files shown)


sophia-guo:
Before update to [latest successful build 123](https://ci.adoptopenjdk.net/view/Dependencies/job/asmtools/123/) a though sanity test will be needed.

test.Dependency will need to rerun. Currently the job in three jenkins failed. https://ci.eclipse.org/openj9/job/test.getDependency/

sophia-guo:
@llxia

test/TestConfig/scripts/tools/getDependencies.pl
```diff
@@ -106,7 +106,7 @@
 	sha1 => 'bfcb96281ea3b59d626704f74bc6d625ff51cbce'
 );
 my %asmtools = (
-	url => 'https://ci.adoptopenjdk.net/view/Dependencies/job/asmtools/lastSuccessfulBuild/artifact/asmtools.jar',
+	url => 'https://ci.adoptopenjdk.net/view/Dependencies/job/asmtools/107/artifact/asmtools.jar',
```
llxia: Instead of hardcode the build#, I think we should update the sha1 to match with the latest one. Jenkins build does not keep history forever.

sophia-guo: Agreed, see my comment at #2933

sophia-guo:
We need to test if the latest jar work with all jdk versions. So will deliver this change to unblock personal build. Meanwhile do a through test to see if there is any issue with latest successful jar.

llxia:
test.Dependency job is passed at three sites.

PR #2947, opened 2018-09-19
Modify testConfig to add a check for guarded storage support

ebadyano:
Added a check to configure.pl to check for zOS z14 or higher hardware to
enable concurrent scavenger tests.

### DIFF
--- a/test/TestConfig/configure.pl
+++ b/test/TestConfig/configure.pl
@@ -1,7 +1,7 @@
 #!/usr/bin/perl
 
 ##############################################################################
-#  Copyright (c) 2016, 2017 IBM Corp. and others
+#  Copyright (c) 2016, 2018 IBM Corp. and others
 #
 #  This program and the accompanying materials are made available under
 #  the terms of the Eclipse Public License 2.0 which accompanies this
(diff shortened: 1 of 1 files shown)


ebadyano:
@sophia-guo could you please review?

test/TestConfig/configure.pl
```diff
@@ -34,23 +34,35 @@
...
-# virtual infor
-my $KVM_Image=`cat /proc/cpuinfo | grep -i QEMU`;
-my $VMWare_Image=`lspci 2>/dev/null | grep -i VMware` ;
-my $Virt_Sys="";
+if ($platform =~ /OS\/390/i) {
+
+	my $hardware=`uname -m`;
+	if ($hardware >= $ZOSR14) {
+		$capability_Hash{guardedstorage} = 1;
+	}
```
fjeremic: This isn't necessarily true. We can very well be on a z14 but guarded storage disabled due to missing OS support. I guess it's a decent starting point though but we ought to be checking the port library or something for these kinds of queries. This shouldn't block the merge of this PR but something to think about for the future.

smlambert:
While we do not have this hardware in the open, will run a limited PR test to ensure script runs.

smlambert:

smlambert:
jenkins test extended zlinux jdk11

PR #2953, opened 2018-09-20
add RC/release build configure parameters to release_process.md

sxa555:
FYA @pshipton @andrew-m-leonard

### DIFF
--- a/doc/processes/release_process.md
+++ b/doc/processes/release_process.md
@@ -109,7 +109,13 @@ them.  They are merely points in time.
 Update the Extensions branches to pull the tagged levels from the `openj9` 
 & `openj9-omr` release branches.
 1. Rebuild the tagged levels and ensure they pass the quality bar defined by 
-AdoptOpenJDK.
+AdoptOpenJDK. These builds need to have extra configure options which will
+ identify them as release builds.  These options are`--with-milestone=fcs`
+ (JDK8 - don't use the default option for an internal build) or
+ `--without-version-pre --without-version-opt` (JDK9+ - don't set
+ pre-release identifier or OPT field that contains a timestamp) added to the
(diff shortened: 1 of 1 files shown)


doc/processes/release_process.md
```diff
@@ -109,7 +109,9 @@ them.  They are merely points in time.
 Update the Extensions branches to pull the tagged levels from the `openj9` 
 & `openj9-omr` release branches.
 1. Rebuild the tagged levels and ensure they pass the quality bar defined by 
-AdoptOpenJDK.
+AdoptOpenJDK. These builds need to have `--with-milestone=fcs` (JDK8) or
```
pshipton: Please explain why these options are required and what they do.

pshipton:
@sxa555 please note you need to sign the Eclipe ECA (see [here](https://github.com/eclipse/openj9/blob/master/CONTRIBUTING.md))and sign your commit in order for the ip-validation check to pass so we can accept your contribution.

doc/processes/release_process.md
andrew-m-leonard: These configure options are what openjdk's makefiles use to set variables that determine what appears in the java -version output, as discussed a while back.. The default ones will mean you have "user" build info in the strings, username's etc...

pshipton: Thanks Andrew. To be clear, what I meant is that the document should be updated with that information.

sxa555:
Signed and hopefully added enough information - if you need any more let me know @pshipton

doc/processes/release_process.md
sxa555: I've added a bit more info to address this - let me know if it's adequate or if more detail is required.

PR #2956, opened 2018-09-20
WIP: Adapt to recent openjdk changes for Java 12

keithc-ca:
* JavaLangAccess declares a new method that must be implemented: setCause(Throwable, Throwable).

* An element of a classpath may be a directory, an image or a zip file. The code was already using j9file_attr() to detect the directory case; use the return value to avoid errors returned by image or a zip file API when the path does not exist.

* The field 'exception' was removed from ExceptionInInitializerError. To adapt, openj9 needs a distinct JCL version so the constant pool can avoid a reference to that field. Other native code that accessed that field must instead use Throwable.cause instead.

* Each JCL version had separate J9VM_JCL_SEnn macros: these were redundant as the same information is now available via JAVA_SPEC_VERSION, defined in j9cfg.h.

* The macro J9VM_JAVA9_BUILD identified various builds of Java 9; today only two builds are relevant: build 148 and the latest. The definition of J9VM_JAVA9_BUILD was moved to j9cfg.h and derived from JAVA_SPEC_VERSION and OPENJ9_BUILD, for consistency and to simplify build configurations.

* Another recent change to openjdk made all fields of Contructor and Method inaccessible via reflection. Certain fields of those classes are cached by java.lang.Class which now does so by bypassing the filtering that normally applies.

https://github.com/ibmruntimes/openj9-openjdk-jdk/pull/36 must be merged as well.


jcl/src/java.base/share/classes/java/lang/Throwable.java
```diff
@@ -362,7 +361,16 @@ public String toString () {
  * @return		the receiver.
  */
 public synchronized Throwable initCause(Throwable throwable) {
-	return	setCause(throwable);
+	/*[PR CMVC 199629] Exception During Class Initialization Not Handled Correctly */
```
pshipton: Might as well remove this line since its not relevant to OpenJ9, and no longer seems relevant to anyone

runtime/include/j9cfg.h.ftl
```diff
@@ -43,6 +43,21 @@ extern "C" {
 
 #define JAVA_SPEC_VERSION ${uma.spec.properties.JAVA_SPEC_VERSION.value}
 
+/* Ensure J9VM_JAVA9_BUILD is not defined elsewhere. */
```
pshipton: Is similar code required in j9cfg.h.in for cmake?

jcl/src/java.base/share/classes/java/lang/Throwable.java
keithc-ca: Are you suggesting I remove all `/*[PR ... */` comments or just this one?

runtime/include/j9cfg.h.ftl
keithc-ca: In the not-to-distant future, I hope to remove `J9VM_JAVA9_BUILD` entirely.

jcl/src/java.base/share/classes/java/lang/Throwable.java
pshipton: Just this one since the code was modified anyway.

pshipton: If you don't want to remove it that is fine. I don't have any other comments. Ready to run PR testing once the WIP is removed.

keithc-ca: I don't mind removing it, I just wasn't clear on the scope of your request.
I'll squash that and the cmake change into related commits and then remove the WIP.

keithc-ca:
Now ready for PR testing.

pshipton:
jenkins test sanity win,zlinux jdk8,jdk11

pshipton:
jenkins test compile xlinuxcmake jdk9

pshipton:
@keithc-ca build problem, please see https://ci.eclipse.org/openj9/job/PullRequest-Sanity-JDK8-win_x86-64_cmprssptrs-OpenJ9/235/console

keithc-ca:
Ok, this needed more work to adjust exports according to the Java version being built and to avoid trying to create libraries that aren't needed (and which would encounter link errors). I'm testing the result: assuming all goes well, I'll collapse some commits and push them to the branch behind this PR.

keithc-ca:
Superseded by https://github.com/eclipse/openj9/pull/3638; closing.

Issue #2911, opened 2018-09-18
IllegalAccessError setting final fields in <clinit> on JDK 11


aguibert:
In the OpenLiberty project we use trace points injected at runtime (using the newest version of ASM).  This works fine with OpenJDK 11, but fails when using OpenJ9:

```
Stack Dump = java.lang.IllegalAccessError: Attempt to set static final field com/ibm/ws/rsadapter/jdbc/v43/WSJdbc43DataSource.tc from method "<clinit>" that is not <clinit>
    at com.ibm.ws.rsadapter.jdbc.v43.WSJdbc43DataSource.<clinit>(WSJdbc43DataSource.java:30)
    at com.ibm.ws.jdbc.osgi.v43.JDBC43Runtime.newDataSource(JDBC43Runtime.java:71)
    at com.ibm.ws.rsadapter.impl.WSManagedConnectionFactoryImpl.createConnectionFactory(WSManagedConnectionFactoryImpl.java:359)
    at com.ibm.ws.rsadapter.impl.WSManagedConnectionFactoryImpl.createConnectionFactory(WSManagedConnectionFactoryImpl.java:99)
    at com.ibm.ws.jca.cm.AbstractConnectionFactoryService.createResource(AbstractConnectionFactoryService.java:158)
    at com.ibm.ws.injectionengine.osgi.internal.IndirectJndiLookupObjectFactory.createResourceWithFilterPrivileged(IndirectJndiLookupObjectFactory.java:383)
    at com.ibm.ws.injectionengine.osgi.internal.IndirectJndiLookupObjectFactory.access$100(IndirectJndiLookupObjectFactory.java:57)
    at com.ibm.ws.injectionengine.osgi.internal.IndirectJndiLookupObjectFactory$3.run(IndirectJndiLookupObjectFactory.java:365)
    at java.security.AccessController.doPrivileged(java.base@11-adoptopenjdk/AccessController.java:696)
    at com.ibm.ws.injectionengine.osgi.internal.IndirectJndiLookupObjectFactory.createResourceWithFilter(IndirectJndiLookupObjectFactory.java:362)
   ... more stack
```

DanHeidinga:
@gacholio Can you take a look at this?

aguibert:
[...]

PR #2961, opened 2018-09-20
Deal with class redefinition and the final field check

gacholio:
Correctly handle the case where a class is redefined during an
initializer which then sets a final field.

Fixes: #2911 

[ci skip]

### DIFF
--- a/runtime/vm/resolvesupport.cpp
+++ b/runtime/vm/resolvesupport.cpp
@@ -59,6 +59,16 @@ finalFieldSetAllowed(J9VMThread *currentThread, bool isStatic, J9Method *method,
 	bool legal = true;
 	/* NULL method means do not do the access check */
 	if (NULL != method) {
+		/* Handle the scenario where the callerClass is redefeined during <clinit> or <init>.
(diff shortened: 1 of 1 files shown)


runtime/vm/resolvesupport.cpp
```diff
@@ -59,6 +59,16 @@ finalFieldSetAllowed(J9VMThread *currentThread, bool isStatic, J9Method *method,
 	bool legal = true;
 	/* NULL method means do not do the access check */
 	if (NULL != method) {
+		/* Handle the scenario where the callerClass is redefeined during <clinit> or <init>.
+		 * In that case, callerClass will be the obsolete version of the class and fieldClass
+		 * will be the current version (assuming that they represent the same class - if they
+		 * don't, this check fails anyway). It's safe to simply update both classes to their
+		 * current versions, as the classfile version check on fieldClass should take place on
+		 * the current versiion, and the Unsafe flag is preserved during redefinition, so it
```
pshipton: there is a typo in "versiion"

gacholio: Fixed

pshipton:

PR #2963, opened 2018-09-20
Optimized Reference ArrayCopy on X86

0dvictor:
Reference ArrayCopy has been optimized that reduces the code-path length,
and potentially improves performance.

Part of Issue: #3054

### DIFF
--- a/runtime/codert_vm/cnathelp.cpp
+++ b/runtime/codert_vm/cnathelp.cpp
@@ -2062,6 +2062,29 @@ old_slow_jitThrowArrayIndexOutOfBounds(J9VMThread *currentThread)
 }
 
 void* J9FASTCALL
+impl_jitReferenceArrayCopy(J9VMThread *currentThread, UDATA lengthInBytes)
+{
+	SLOW_JIT_HELPER_PROLOGUE();
+	void* exception = NULL;
+	if (-1 != currentThread->javaVM->memoryManagerFunctions->referenceArrayCopy(currentThread,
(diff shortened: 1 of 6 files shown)


0dvictor:
This PR is based on OMR's arraycopy evaluator and OpenJ9's own "VMarrayStoreCheckArrayCopyEvaluator".
The three changes are:
1. J9 takes over the logic related to reference array copy even when store check is not required;
2. Call reference array copy helper when store check is not required but Concurrent Scavenge is active;
3. Create a JIT helper wrapper of GC's reference array copy to reduce JIT code register pressure.

0dvictor:
Performance impact were analyzed in two ways:
1. IBM's internal benchmarks showed no performance impact;
2. Generate assemblies are verified to be identical as what OMR generates under GC policy GenCon, Concurrent Scavenge disabled, when array store check is not required.

andrewcraik:

gacholio:
@andrewcraik I don't like the fact this was merged without involving me. I would certainly have rejected the formatting in cnathelp.cpp and xnathelp.m4. The C code also contains unnecessary (though not strictly incorrect) code. Much like the new CS code, I would likely have insisted on a more sensible way of dealing with the differences in calling conventions.

0dvictor:
@gacholio I apologize for forget to let you know. I've created #3112 base on your comments on the CS code. Could you take a look?

andrewcraik:
@gacholio my sincere apologies - I was reading for correctness and did not notice those changes were in files you normally review - @0dvictor thank you for opening #3112.

PR #2964, opened 2018-09-20
Fix access checking with protected classes

pdbain-ibm:
Treat `protected` inner classes and interfaces as `public` when doing access checks.

### DIFF
--- a/jcl/src/java.base/share/classes/java/lang/invoke/MethodHandles.java
+++ b/jcl/src/java.base/share/classes/java/lang/invoke/MethodHandles.java
@@ -527,16 +527,7 @@ public class MethodHandles {
 				 * the protected flag of this class doesn't exist on the VM level (there is no 
 				 * access flag in the binary form representing 'protected')
 				 */
-				if (Modifier.isProtected(modifiers)) {
-					/* Interfaces are not classes but types, and should therefore not have access to 
-					 * protected methods in java.lang.Object as subclasses.
-					 */
-					if (!accessClass.isInterface()) {
-						return;
-					}
-				
-				/* The following access checking is for a normal class (non-member class) */
-				} else if (Modifier.isPublic(modifiers)) {
+				if (Modifier.isPublic(modifiers) || Modifier.isProtected(modifiers)) {
 					/* Already determined that we have more than "no access" (public access) */
 					return;
 				} else {


jcl/src/java.base/share/classes/java/lang/invoke/MethodHandles.java
```diff
@@ -313,17 +313,19 @@ private static boolean isSamePackage(Class<?> a, Class<?> b){
 				return true;
 			}
 			
+			Class<?> classA = (a.isArray())? a.getComponentType(): a;
+			Class<?> classB = (b.isArray())? b.getComponentType(): b;
```
DanHeidinga: this needs to be a loop:
```
while (b.isArray()) {
   b = b.getComponentType();
}
```

jcl/src/java.base/share/classes/java/lang/invoke/MethodHandles.java
```diff
@@ -522,8 +524,9 @@ private void checkClassAccess(Class<?> targetClass) throws IllegalAccessExceptio
 				if (Modifier.isProtected(modifiers)) {
 					/* Interfaces are not classes but types, and should therefore not have access to 
 					 * protected methods in java.lang.Object as subclasses.
+					 * They should have access to members of the same package.
 					 */
-					if (!accessClass.isInterface()) {
+					if (!accessClass.isInterface() || isSamePackage(accessClass, targetClass)) {
```
DanHeidinga: This needs more thought

jcl/src/java.base/share/classes/java/lang/invoke/MethodHandles.java
pdbain-ibm: Fixed.  Thank you.

jcl/src/java.base/share/classes/java/lang/invoke/MethodHandles.java
pdbain-ibm: We may be excluding protected members from superinterfaces, right?

jcl/src/java.base/share/classes/java/lang/invoke/MethodHandles.java
```diff
@@ -313,17 +313,27 @@ private static boolean isSamePackage(Class<?> a, Class<?> b){
 				return true;
 			}
 			
+			Class<?> classA = a;
```
DanHeidinga: Why introduce `classA` here?  Especially when the loop doesn't use it....

pdbain-ibm: I prefer not to modify method arguments, but I can remove the locals if you wish.  What is the convention?

DanHeidinga: I'm not aware of any convention that says not to update arguments

jcl/src/java.base/share/classes/java/lang/invoke/MethodHandles.java
DanHeidinga: > protected member interface foo.KlassA.InterfaceB.

Can you provide a test case?  I'm not 100% sure I know what a protected member interface is.

pdbain-ibm:
```
package toe.foo;
public class Klass {
	protected interface PI {
		public default int getInt() {
		}
	}
}
```
`PI` is visible only to classes in package `toe.foo` or subclasses of `Klass`.

jcl/src/java.base/share/classes/java/lang/invoke/MethodHandles.java
pdbain-ibm: Fair enough.  I removed the auxiliary local variable.

jcl/src/java.base/share/classes/java/lang/invoke/MethodHandles.java
DanHeidinga: I think the `protected` in this case is a Java Language Spec (JLS) statement.  The VM view of the world doesn't include protected classes - classes are either public or package in the VM.

If we remove this entire check for `protected` and instead treat it as `public` are there any issues with the existing tests?

pdbain-ibm: I am testing now, though this is a caution:
`Interfaces are not classes but types, and should therefore not have access to protected methods in java.lang.Object as subclasses.`

DanHeidinga: @pdbain-ibm This is in `checkClassAccess()`.  There are other `checkAccess` methods for checking access to members.  Which leaves this check as a grey area that needs further investigation.

Thanks for running the extra tests.

DanHeidinga:
@pdbain-ibm One way to move forward here is to split the two parts into separate PRs.  The array part is obviously correct and can be merged anytime.

The interface / samepackage check could be moved to a separate PR and discussion can continue there.

pdbain-ibm:
@DanHeidinga good plan.  Please see https://github.com/eclipse/openj9/pull/3016.

pdbain-ibm:
@DanHeidinga would you kindly review while I finish testing?

DanHeidinga:

Issue #2871, opened 2018-09-14
Writing to closed socket connection kills the JVM


pdbain-ibm:
Recent regression.  Looks like a blocker.  FYI @gacholio .  I am investigating. 

```
$ java -version
openjdk version "10.0.2-internal" 2018-07-17
OpenJDK Runtime Environment (build 10.0.2-internal+0-adhoc.jenkins.Build-JDK10-linuxx86-64cmprssptrs)
Eclipse OpenJ9 VM (build master-e60de42, JRE 10 Linux amd64-64-Bit Compressed References 20180914_446 (JIT enabled, AOT enabled)
OpenJ9   - e60de42
OMR      - 28b6fbc
JCL      - a6243ec based on jdk-10.0.2+13)
28a57f78e63e-root:8$ "/java/bin/java"  -Xcompressedrefs  \
> --add-exports=java.base/com.ibm.tools.attach.target=ALL-UNNAMED \
> -cp "/openj9test/TestConfig/scripts/testKitGen/../../../../jvmtest/TestConfig/resources:/openj9test/TestConfig/scripts/testKitGen/../../../../jvmtest/TestConfig/lib/testng.jar:/openj9test/TestConfig/scripts/testKitGen/../../../../jvmtest/TestConfig/lib/jcommander.jar:/openj9test/TestConfig/scripts/testKitGen/../../../../jvmtest/functional/Java8andUp/GeneralTest.jar" \
> -Dcom.ibm.tools.attach.enable=yes \
> -Dcom.ibm.tools.attach.timeout=15000 \
> -Djava.sidecar="--add-exports=java.base/com.ibm.tools.attach.target=ALL-UNNAMED" \
> org.testng.TestNG -d "/openj9test/TestConfig/scripts/testKitGen/../../../TestConfig/test_output_15369295991794/TestAttachErrorHandling_0" "/openj9test/TestConfig/scripts/testKitGen/../../../../jvmtest/functional/Java8andUp/testng.xml" -testnames TestAttachErrorHandling \
> -groups level.extended \
> -excludegroups d.*.linux_x86-64_cmprssptrs,d.*.arch.x86,d.*.os.linux,d.*.bits.64,d.*.generic-all
[IncludeExcludeTestAnnotationTransformer] [INFO] exclude file is null
...
... TestNG 6.14.2 by C?dric Beust (cedric@beust.com)
...

[...]

PR #2965, opened 2018-09-20
Fix signal handler registration in the JVM

babsingh:
**omrsig.cpp::**`omrsig_primary_sigaction`/`omrsig_primary_signal` doesn't
support registration with SIG_DFL/SIG_IGN/NULL as the signal handler.

1) `SIG_IGN` is passed as the signal handler for `SIGPIPE`.
2) `NULL` is passed as the new signal handler for `SIGILL`.
 
In the above cases, `omrsig_primary_sigaction`/`omrsig_primary_signal`
shouldn't be used for registering a signal handler. 

Now onwards, only **<signal.h>::**`sigaction`/`signal` will used in the above
cases.

`SIGPIPE` needs to be ignored for the Java Attach API to work properly.

Fixes https://github.com/eclipse/openj9/issues/2871

### DIFF
--- a/runtime/vm/jvminit.c
+++ b/runtime/vm/jvminit.c
@@ -825,7 +825,7 @@ freeJavaVM(J9JavaVM * vm)
 #if !defined(WIN32)
 	/* restore any handler we may have overwritten */
 	if (NULL != vm->originalSIGPIPESignalAction) {
-		OMRSIG_SIGACTION(SIGPIPE,(struct sigaction *)vm->originalSIGPIPESignalAction, NULL);
+		sigaction(SIGPIPE,(struct sigaction *)vm->originalSIGPIPESignalAction, NULL);
 		j9mem_free_memory(vm->originalSIGPIPESignalAction);
 		vm->originalSIGPIPESignalAction = NULL;
 	}
@@ -5362,7 +5362,7 @@ protectedInitializeJavaVM(J9PortLibrary* portLibrary, void * userData)
 	newSignalAction.sa_flags = 0;
 #endif /* defined(J9ZTPF) */
 	newSignalAction.sa_handler = SIG_IGN;
-	OMRSIG_SIGACTION(SIGPIPE,&newSignalAction,(struct sigaction *)vm->originalSIGPIPESignalAction);
+	sigaction(SIGPIPE,&newSignalAction,(struct sigaction *)vm->originalSIGPIPESignalAction);
 #endif
 
 #ifdef J9VM_OPT_SIDECAR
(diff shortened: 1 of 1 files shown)


pdbain-ibm:
@pshipton would you kindly run an extended build so we can see it this fixes the attach API error handling failure?
@babsingh did you test it with the test case supplied with issue #2871?

Thank you.

pshipton:
Are there any other places/signals with a similar problem?

jenkins test extended zlinux jdk8

pshipton:
@babsingh does this place also need to be fixed?

https://github.com/eclipse/openj9/blob/4833764455a3a3a9c0873ffb2e353d8b121ba49e/runtime/vm/jvminit.c#L6490

babsingh:
> did you test it with the test case supplied with issue #2871?

yes, it passes.

>  OMRSIG_SIGACTION(SIGILL, NULL, &oldHandler); 

yes, this place would also need to be fixed since `omrsig_primary_sigaction` would return error for NULL signal handler without setting the old handler. so, omrsig lib shouldn't be used in this scenario. i will update the PR.

pshipton:
https://ci.eclipse.org/openj9/job/PullRequest-Extended-JDK8-linux_390-64_cmprssptrs-OpenJ9/72/

babsingh:
@pshipton I have updated two other places where `omrsig_primary_sigaction`/`omrsig_primary_signal` shouldn't be used.

pshipton:
Waiting for https://ci.eclipse.org/openj9/job/PullRequest-Extended-JDK8-linux_390-64_cmprssptrs-OpenJ9/72/ to complete and pass.

jenkins compile xlinux jdk8

babsingh:
Removed changes related to `SIG_RI_INTERRUPT`. 

Opened https://github.com/eclipse/openj9/issues/2967 to resolve the issue related to `SIG_RI_INTERRUPT`.

babsingh:
@pshipton https://ci.eclipse.org/openj9/job/PullRequest-Extended-JDK8-linux_390-64_cmprssptrs-OpenJ9/72/ has passed.

I have verified that `org.openj9.test.attachAPI.TestAttachErrorHandling:test_startupShutdownRobustness` is passing: https://ci.eclipse.org/openj9/job/PullRequest-Extended-JDK8-linux_390-64_cmprssptrs-OpenJ9/72/consoleFull.

babsingh:
> we want to revisit this

Agreed, we should revisit this. right approach: `omrsig_primary_signal/omrsig_primary_sigaction` should behave similar to `signal/sigaction`. `omrsig_primary_signal/omrsig_primary_sigaction` should be updated to handle `SIG_IGN/SIG_DFL/NULL` cases similar to `signal/sigaction`. this way `omrsig_primary_signal/omrsig_primary_sigaction` and `signal/sigaction` can be used interchangeably. @DanHeidinga do you agree? if agreed, i will open an OMR issue.

pshipton:
> omrsig_primary_signal/omrsig_primary_sigaction should behave similar to signal/sigaction

I'm not sure this actually works. There is a difference between user code calling sigaction, and the VM calling sigaction. If the VM has registered a signal we don't want user code to then set it to ignore or default.

DanHeidinga:
@babsingh Can you open an issue and outline the changes that have been made so far, why they were made, and what you think the right path forward is?

Let's have the discussion in a top-level issue rather than in the comments on an already merged PR.  Too easy to lose this discussion :)

PR #2969, opened 2018-09-20
WIP: Add test dependencies to build Dockerfiles

fjeremic:
There are a few additional test dependencies needed to be able to build
and test OpenJ9 from within the same docker container. We add these
test dependencies to our Dockerfile so that a user is able to
simultaneously build and test a JDK from within the same container.

### DIFF
--- a/buildenv/docker/jdk11/ppc64le/ubuntu16/Dockerfile
+++ b/buildenv/docker/jdk11/ppc64le/ubuntu16/Dockerfile
@@ -32,7 +32,10 @@ RUN apt-get update \
   && add-apt-repository ppa:ubuntu-toolchain-r/test \
   && apt-get update \
   && apt-get install -qq -y --no-install-recommends \
+    ant \
+    ant-contrib \
     autoconf \
+    build-essential \
     ca-certificates \
     ccache \
     cpio \
@@ -55,6 +58,7 @@ RUN apt-get update \
     libxt-dev \
     libxtst-dev \
     make \
+    perl \
     pkg-config \
     realpath \
     ssh \
@@ -62,6 +66,9 @@ RUN apt-get update \
     wget \
     zip \
   && rm -rf /var/lib/apt/lists/*
+
+# Install Docker module to run test framework
+RUN echo yes | cpan install JSON Text::CSV
 
 # Install optional tools
 RUN apt-get update \
--- a/buildenv/docker/jdk11/s390x/ubuntu16/Dockerfile
+++ b/buildenv/docker/jdk11/s390x/ubuntu16/Dockerfile
@@ -32,7 +32,10 @@ RUN apt-get update \
   && add-apt-repository ppa:ubuntu-toolchain-r/test \
   && apt-get update \
   && apt-get install -qq -y --no-install-recommends \
+    ant \
+    ant-contrib \
     autoconf \
+    build-essential \
     ca-certificates \
     ccache \
     cmake \
@@ -54,6 +57,7 @@ RUN apt-get update \
     libxt-dev \
     libxtst-dev \
     make \
+    perl \
     pkg-config \
     realpath \
     ssh \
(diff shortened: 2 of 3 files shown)


fjeremic:
I wanted to open this up and mark it as WIP so we can review and discuss this in one place. I generally believe we should move forward in the direction of consolidating our Dockerfiles so as to make it as easy as possible for an arbitrary user to build and test OpenJ9.

Presumably someone building OpenJ9 may want to also run some tests in case they feel up for improving the project. It seems unnecessary to have users jump through hoops to get a testing environment set up. The user should be able to build and test an OpenJ9 JDK from within the same docker container.

Question for reviewers:

Should we backport such changes to all jdk* releases? I can do this no problem.

Tagging @AdamBrousseau @smlambert @llxia for reviews.

AdamBrousseau:
It's probably a question of what are/is the most common use case(s) for the Dockerfile.
If we have the dependencies documented for build/test then one should not be confused or be relying upon dependencies unintentionally. I don't think its expensive to have these few test packages added.

fjeremic:
> It's probably a question of what are/is the most common use case(s) for the Dockerfile.

We have publically available binaries of OpenJ9 over at AdoptOpenJDK so if a user is building from source presumably they want to contribute back to the project or hack their own modifications to test something. In both scenarios users should be encouraged to take advantage of all the tests we provide and having the ability to build and test from within the same docker container is invaluable.

We need to encourage all developers as well to run OpenJ9 tests as well, something I definitely know doesn't happen today. This is a step in the right direction to encourage people to do so as it removes additional steps needed to be able to run the tests available.

keithc-ca:
I think it just plain wrong to mix the build and test dependencies. We *don’t* want to unintentionally rely on things for the build just because they’re required for (some) tests. Likewise, the test should not have to depend on all the tools needed to build a VM.

There's nothing preventing the creation of a Dockerfile that starts with either the build or test Dockerfile and mixes in the content from the other, but, I repeat, we don't want to have extraneous packages installed for either the build or test images.

Perhaps a concrete example will help. Suppose the test image had gcc installed; if testing only occurs in such a world, you could never claim that the JVM you tested works in environments without gcc installed.

fjeremic:
> Perhaps a concrete example will help. Suppose the test image had gcc installed; if testing only occurs in such a world, you could never claim that the JVM you tested works in environments without gcc installed.

Counterexample to that is that our CI builds already bundle the "build and test" dependencies [1].

[1] https://github.com/eclipse/openj9/blob/a65785ba6a75ca53bdb15921808bb339479440c4/buildenv/jenkins/docker-slaves/x86/centOS/Dockerfile#L95-L97

keithc-ca:
I do see the value in helping to make things simpler for contributors, but Dockerfiles have other uses. It was my understanding that we were seriously considering doing builds in docker containers (as seems to be in the plans at https://github.com/AdoptOpenJDK/openjdk-build). Using docker images allows us to be clear about the distinction between build and test dependencies (and avoid the 'separate machine' problem Adam mentioned).

DanHeidinga:
Can this be resolved with two docker files?  The basic build env (which is used to document the minimum build environment) and a build&test one which extends the basic build with the test dependencies?

DanHeidinga:
Might be worth looking at multi-stage docker builds:https://docs.docker.com/develop/develop-images/multistage-build/ Particularly at the named stages and the ability to stop the build at a particular stage. 

Our images could have a `build_env` & a `test_env` stage that differentiate between the build and test requirements.

keithc-ca:
I wasn't aware of the 'stop' option, but that sounds good. I'd propose we have three stages: build, test and both (substitute your favourite better name here).

fjeremic:
Cool! Never knew about multi-stage builds. Thanks @DanHeidinga.

> I wasn't aware of the 'stop' option, but that sounds good. I'd propose we have three stages: build, test and both (substitute your favourite better name here).

In light of this I would agree with Keith. We can have build, test, and build-and-test (name?) stages. Sounds like best of all worlds. I can repurpose this PR to start with an initial solution.

fjeremic:
Closing due to the work already being done. See #5276.

Issue #2088, opened 2018-06-06
JIT Assertion failure when compiling with debug enabled.


pdbain-ibm:
To reproduce, 

1. I checked out the OpenJDK 9 with OpenJ9 project per 
https://www.eclipse.org/openj9/oj9_build.html

2. Enabled debugging per https://blog.openj9.org/2018/06/05/debugging-openj9-in-docker-with-gdb/
```
export UMA_DO_NOT_OPTIMIZE_CCODE="1"
export VMDEBUG="-g3 -fno-inline -O0"
export VMLINK="-g -O0"
export enable_debug=yes
export enable_optimized=no
export enable_optimize=no
export CXXFLAGS="-O0 -g3"
export CFLAGS="-O0 -g3"
export BUILD_CONFIG=debug
```

3. and built it like so:
```
cd /projects/openj9/openj9-openjdk-jdk9/
bash get_source.sh 
bash ./configure --with-freemarker-jar=/projects/freemarker.jar 
make images
```

I get a compilation failure:
```
Compiling 4 files for BUILD_JIGSAW_TOOLS
Assertion failed at /macprojects/openj9/openj9-openjdk-jdk9/build/linux-x86_64-normal-server-fastdebug/vm/compiler/../compiler/optimizer/J9TransformUtil.cpp:374: isFinalFieldPointingAtNativeStruct(symRef, comp)
VMState: 0x000513ff
	dereferenceStructPointerChain should be dealing with reference fields
compiling sun/nio/fs/NativeBuffers.copyCStringToNativeBuffer([BLsun/nio/fs/NativeBuffer;)V at level: warm
#0: function TR_LinuxCallStackIterator::printStackBacktrace(TR::Compilation*)+0x32 [0x2adb9e19143a]
#1: function TR_Debug::printStackBacktrace()+0x1f [0x2adb9e1a3ffb]
#2: /macprojects/openj9/openj9-openjdk-jdk9/build/linux-x86_64-normal-server-fastdebug/jdk/lib/compressedrefs/libj9jit29.so(+0xab0eb2) [0x2adb9de25eb2]
#3: function TR::assertion(char const*, int, char const*, char const*, ...)+0xb3 [0x2adb9de26042]
#4: /macprojects/openj9/openj9-openjdk-jdk9/build/linux-x86_64-normal-server-fastdebug/jdk/lib/compressedrefs/libj9jit29.so(+0x942ece) [0x2adb9dcb7ece]
[...]

PR #2971, opened 2018-09-20
Remove incorrect assertion in dereferenceStructPointerChain

liqunl:
Assertion is not needed in calculating the field address because the
field has been verified by verifyFieldAccess.

Fixes: #2088


liqunl:
@andrewcraik Can you review? This is a low risk change. It only stops the assert in debug build or build with assertion on. Doesn't change other behavior.

andrewcraik:
Just waiting on the CI build.

PR #2979, opened 2018-09-21
Replace mingw with clang in compilation

ChengJin01:
The change is to modify all scripts/code
involved to ensure we use Clang in the place
of mingw to compile the bytecode interpreter
on Windows.

depends on https://github.com/eclipse/openj9/issues/2980


pshipton:
Have you tested building Windows 32-bit?

ChengJin01:
I will try compiling a 32-bit Java 8 build on a Fyre machine to ensure everything works fine.

ChengJin01:
[sanity_test_results.zip](https://github.com/eclipse/openj9/files/2418023/sanity_test_results.zip)

I've already complid Java 8 32 bit with MSVC + Clang (32bit). There is pretty much no change with compiler flags except that the compiler needs to be Clang 32bit (http://releases.llvm.org/6.0.1/LLVM-6.0.1-win32.exe)

[1]  Sanity test(functional)  64bit:
No test failure was detected on both msvc  + mingw and msvc + clang:
TOTAL: 163   EXECUTED: 87   PASSED: 87   FAILED: 0   SKIPPED: 76
ALL TESTS PASSED

[2] Sanity test (functional) 32bit:
No problem was detected on msvc + mingw and 5 test failures (crash?) were detected on msvc + clang
(1) msvc + mingw 
TOTAL: 163   EXECUTED: 94   PASSED: 94   FAILED: 0   SKIPPED: 69
ALL TESTS PASSED

(2) msvc + clang
FAILED test targets:
        cmdLineTester_jvmtitests_hcr_SE80_3
        JCL_Test_SE80_0
        JCL_Test_SE80_1
        jit_jitt_0
        jit_jitt_3

TOTAL: 163   EXECUTED: 94   PASSED: 89   FAILED: 5   SKIPPED: 69

I manually ran one of failing test cmdLineTester_jvmtitests_hcr_SE80_3 and it stopped working in testBCIUsingASM_InjectTryCatch_Level2_injectCatchAndThrowNewAIOBE_catchAIOBEAndThrowNewAE
```
C:/temp_jdk8/perf_win/clang_jdk8_32bit/windows-x86-normal-server-release/images/j2sdk-image/jre/bin/java  -Xgcpolicy:gencon -Xjit:count=0 -Xnocompressedrefs  -Xdump    -agentlib:jvmtitest=test:ta001 -cp "C:/temp_jdk8/openj9_v2/jvmtest/functional/cmdLineTests/jvmtitests/jvmtitest.jar;C:/temp_jdk8/openj9_v2/jvmtest/TestConfig/lib/asm-all.jar" com.ibm.jvmti.tests.util.TestRunner

*** Testing [4/13]:     testBCIUsingASM_InjectTryCatch_Level2_injectCatchAndThrowNewAIOBE_catchAIOBEAndThrowNewAE
Expected NullPointerException received
(stopped working.....)
```

Not yet figured out what happened to the failing tests on the build compiled with msvc + clang

ChengJin01:
I tried replacing __GNUC__ with __clang__ in oti/mingw/stddef.h and oti/mingw_comp.h as Clang already discarded ___GNUC__ in 6.0.1  (the predefined macro remains in older versions but has been removed now) to get Clang accept all definitions previously used by mingw. However the recompiled build still ended up with same failure in testBCIUsingASM_InjectTryCatch_Level2_injectCatchAndThrowNewAIOBE_catchAIOBEAndThrowNewAE, which means the replacement for __GNUC__  is logically incorrect and has nothing to do with the test failures.

Still need to get back to the failing tests for root cause.

ChengJin01:
Investigation shows these test failure were triggered by 3 test classes in which they share the pretty much the same nested try-catch structure. That being said, they belongs to the same issue.

1)  cmdLineTester_jvmtitests_hcr_SE80_3 (executed with -Xjit:count=0)
Running command: "C:/temp_jdk8/perf_win/clang_jdk8_32bit/windows-x86-normal-server-release/images/j2sdk-image/jre/bin\\java"  -Xgcpolicy:gencon **-Xjit:count=0** -Xnocompressedrefs  -Xdump    -agentlib:jvmtitest=test:ta001 -cp "C:/temp_jdk8/openj9_v2/test///..//jvmtest\functional\cmdLineTests\jvmtitests\jvmtitest.jar;C:/temp_jdk8/openj9_v2/test///..//jvmtest\TestConfig\lib\asm-all.jar" com.ibm.jvmti.tests.util.TestRunner

/test/functional/cmdLineTests/jvmtitests/src/com/ibm/jvmti/tests/BCIWithASM/ta001.java
```
testBCIUsingASM_InjectTryCatch_Level2_injectCatchAndThrowNewAIOBE_catchAIOBEAndThrowNewAE {
...
		byte[] trasnformedClassBytes = ASMTransformer.trasnform_injectCatchAndThrowNewAIOBE_catchAIOBEAndThrowNewAE( Source4.class );
		boolean transformed = redefineClass( Source4.class, trasnformedClassBytes.length, trasnformedClassBytes );
		Source4 postState = new Source4();
		try { 
			int postStateResult = postState.returnOne(); <------------ crash here
```
/test/functional/cmdLineTests/jvmtitests/src/com/ibm/jvmti/tests/BCIWithASM/Source4.java
```
public class Source4 {
	public int returnOne() {
		//try {                            <---------------------------- nested try-catch structure
			String str = null;  
			str.equals("a");	
			return 1;
		//} catch ( NullPointerException e1 ) {
		//	try {												<--injected code 
		//		int intArray [] = new int [] {1};   
		//		return intArray[1];	
		//	} catch ( ArrayIndexOutOfBoundsException e2 ) {
		//		return 1/0;
		//	}
		//}
	}
}
```
The crash issue disappeared by removing count=0 (only with JIT on/off)

2) cfgTest in jit_jitt_0/jit_jitt_3  (executed with -Xjit:count=0)
test with -Xjit:noJitUntilMain,**count=0**,assumeStrictFP, .....

```
  <test name="cfgTest">
    <classes>
      <class name="jit.test.jitt.cfg.nestedExceptions"/>
    </classes>
```
/test/functional/JIT_Test/src/jit/test/jitt/cfg/nestedExceptions.java
```
public class nestedExceptions extends jit.test.jitt.Test {
   int tstExcp(int i) {
      try {
         if (i > 10)
            throw new Exception();
      }
      catch(Exception e1) {
         try {
            if (i > 11)
               throw new Exception();
         }
         catch(Exception e2) {
            try {
               if (i > 12)
                  throw new Exception();
            }
            catch(Exception e3) {  
               i = 99;
            }
         }
      }
      return i;
   };
```
The crash issue disappeared by removing count=0 (only with JIT on/off)

3)  JCL_Test_SE80_0/JCL_Test_SE80_1 (executed with JIT on by default)
/test/functional/Java8andUp/playlist.xml
```
		<testCaseName>JCL_Test_SE80</testCaseName>
       -testnames \
         ...
	JCL_TEST_Java-Lang,\
```
/test/functional/Java8andUp/testng.xml
```
	<test name="JCL_TEST_Java-Lang">
		<classes>
			<class name="org.openj9.test.java.lang.Test_Class"/>
			<class name="org.openj9.test.java.lang.Test_ClassCastException"/> <------
			<class name="org.openj9.test.java.lang.Test_Compiler"/>
			<class name="org.openj9.test.java.lang.Test_InstantiationError"/>
			<class name="org.openj9.test.java.lang.Test_InstantiationException"/> <------
```
/test/functional/Java8andUp/src/org/openj9/test/java/lang/Test_ClassCastException.java
```
	public void test_Constructor() {
		// Test for method java.lang.ClassCastException()
		try {      <------------------------------------------ nested try-catch structure
			try {
				String z = (String)new Object();
				if (z == null)
					; //use z to avoid a warning msg
			} catch (ClassCastException e) {
				return;
			}
			AssertJUnit.assertTrue("Failed to generate Exception", false);
		} catch (Exception e) {
			AssertJUnit.assertTrue("Exception during test", false);
		}
	}
```
/test/functional/Java8andUp/src/org/openj9/test/java/lang/Test_InstantiationException.java
```
	public void test_Constructor() {
		try {          <------------------------------------------ nested try-catch structure
			try {
				Class cl = null;
				cl = Class.forName("java.io.OutputStream");
				cl.newInstance();
			} catch (InstantiationException e) {
				return;
			}
			AssertJUnit.assertTrue("Failed to generate Exception", false);
		} catch (Exception e) {
			AssertJUnit.assertTrue("Exception during test", false);
		}
	}
```
The crash issue disappeared by specifying -Xint (with JIT off)

So it turns out JIT doesn't work good with the nested exception structure on the Clang compiled build on Windows 32bit.

I suggest to create a separate issue for JIT to address the issue after merging the changes. Meanwhile, these failing test cases should be temporarily excluded unless the crash issue gets revolved by JIT.

-------------
@pshipton , any other concern at this point?

pshipton:
JIT should take a look at the crashes in advance of merging. We can't just exclude and ignore the problems as they may appear in other ways.
@andrewcraik

pshipton:
or maybe @gacholio can help, its strange there would be a JIT issue when only the bytecode interpreter is compiled with clang.

gacholio:
As with anything that crashes with the JIT and not without, limit the compilations down using -Xjit:verbose,vlog=filename and then -Xjit:limitfile=filename (you can trim the limitfile to get down to the minimal set of compiled methods). Especially effective with count=0.

gacholio:
Personally, I think making a change which causes tests for fail for unknown reasons and excluding the tests is a terrible idea. Who knows what else out there is going to exhibit the issue?

ChengJin01:
It's definitely correct to get JIT guys involved in the analysis before mergin the changes. 
I am wondering how they get started to triage the problem as the changes have not yet been merged.  Should we upload the compiled build to somewhere in vm_archive for downloading?

pshipton:
Guess we can check with them and see how they want to proceed.

ChengJin01:
> As with anything that crashes with the JIT and not without, limit the compilations down using -Xjit:verbose,vlog=filename and then -Xjit:limitfile=filename...

To confirm the crash was caused by JIT/count=0, actually I already tried isolating the failing methods  by removing all methods in the test file before compiling the test case.

e.g.
in /test/functional/cmdLineTests/jvmtitests/src/com/ibm/jvmti/tests/BCIWithASM/ta001.java
remove all other methods except testBCIUsingASM_InjectTryCatch_Level2_injectCatchAndThrowNewAIOBE_catchAIOBEAndThrowNewAE.
to ensure the only the failing method is executed:

1)   it crashed when specifying -Xjit:count=0 
```
Administrator@JCHTORWIN101 /cygdrive/c/temp/test
C:/temp_jdk8/perf_win/clang_jdk8_32bit/windows-x86-normal-server-release/images/j2sdk-image/jre/bin/java  -Xgcpolicy:gencon -Xjit:count=0 -Xnocompressedrefs  -Xdump    -agentlib:jvmtitest=test:ta001 -cp "C:/temp_jdk8/openj9_v2/jvmtest/functional/cmdLineTests/jvmtitests/jvmtitest.jar;C:/temp_jdk8/openj9_v2/jvmtest/TestConfig/lib/asm-all.jar" com.ibm.jvmti.tests.util.TestRunner
*** Testing [1/1]:      testBCIUsingASM_InjectTryCatch_Level2_injectCatchAndThrowNewAIOBE_catchAIOBEAndThrowNewAE
Expected NullPointerException received
(crashed)
```
2)  it passed wen specifying -Xjit or -Xint
```
Administrator@JCHTORWIN101 /cygdrive/c/temp/test
C:/temp_jdk8/perf_win/clang_jdk8_32bit/windows-x86-normal-server-release/images/j2sdk-image/jre/bin/java  -Xgcpolicy:gencon -Xjit -Xnocompressedrefs  -Xdump    -agentlib:jvmtitest=test:ta001 -cp "C:/temp_jdk8/openj9_v2/jvmtest/functional/cmdLineTests/jvmtitests/jvmtitest.jar;C:/temp_jdk8/openj9_v2/jvmtest/TestConfig/lib/asm-all.jar" com.ibm.jvmti.tests.util.TestRunner
*** Testing [1/1]:      testBCIUsingASM_InjectTryCatch_Level2_injectCatchAndThrowNewAIOBE_catchAIOBEAndThrowNewAE
Expected NullPointerException received
Expected ArithmeticException received
*** Test took 31 milliseconds
OK
```

The triage on other 2 failing tests followed the same way: the whole test case passed by removing the failing methods or by specifying -Xjit(removing count=0) or -Xint (disabling JIT)

ChengJin01:
To get things simple, directly compile and run the following class (the class was copied from /test/functional/cmdLineTests/jvmtitests/src/com/ibm/jvmti/tests/BCIWithASM/Source4.java which is used in testBCIUsingASM_InjectTryCatch_Level2_injectCatchAndThrowNewAIOBE_catchAIOBEAndThrowNewAE)
```
public class DivByZero {
	public int returnOne() {
		try {
			String str = null;  
			str.equals("a");	
			return 1;
		} catch ( NullPointerException e1 ) {
			try { 
				int intArray [] = new int [] {1};   
				return intArray[1];	
			} catch ( ArrayIndexOutOfBoundsException e2 ) {
				return 1/0;
			}
		}
	}
	public static void main(String[] args) {
		DivByZero db = new DivByZero();
		db.returnOne();
	   }
}
```
It crashed when specifying -Xjit:count=0.
It passed with the following exception when specifying -Xjit or -Xint.
```
Exception in thread "main" java.lang.ArithmeticException: divide by zero
        at DivByZero.returnOne(DivByZero.java:13)
        at DivByZero.main(DivByZero.java:19)
```

fjeremic:
@0dvictor could you help take a look? Seems to fail consistently with `-Xjit:count=0`. You can work with @ChengJin01 to get the build somehow so you don't have to recompile yourself.

ChengJin01:
@0dvictor ,  I already packaged these compiled builds at T:\cjin\z_jdk11_comp\perf_builds\java8_32bit

Administrator@JCHTORWIN101 /cygdrive/t/cjin/z_jdk11_comp/perf_builds
$ ls *
java8:
build_jdk8_clang.zip  build_jdk8_mingw.zip  (64bit)

java8_32bit:
clang_jdk8_32bit.zip  DivByZero.class  DivByZero.java  mingw_jdk8_32bit.zip

0dvictor:
The failure can be reproduced with `-Xjit:noJitUntilMain,count=0,optlevel=noopt,limit={DivByZero.returnOne*}`. Only one method, DivByZero,returnOne, is compiled. JIT trace logs showed the compiled code, stack atlas, and gc maps are identical for runs from the clang package and the mingw package.
[mingw.log](https://github.com/eclipse/openj9/files/2429167/mingw.log)
[clang.log](https://github.com/eclipse/openj9/files/2429168/clang.log)

ChengJin01:
@keithc-ca , does it mean there is problem with signal handler to deal with the nested exceptions on Windows 32bit?

gacholio:
I suspect the root issue lies in the exception throw entry to the interpreter from the JIT (this path is never taken in pure interpreter mode). If you can run this in the debugger, break at `throwCurrentExceptionFromJIT` and trace from there. You will get exceptions from the divide by 0, which you should use `gh` to continue from. I don't believe there's any problem in the exception handlers or JIT code.

0dvictor:
I ran it in the debugger (Visual Studio), and found that `Unhandled exception at 0x100194C0 (j9vm29.dll) in java.exe: 0xC00001A5: An invalid exception handler routine has been detected (parameters: 0x00000002).` It seems the exception handler wasn't setup properly. It never reached `throwCurrentExceptionFromJIT`.

keithc-ca:
@babsingh Any thoughts about signal handling?

gacholio:
Is there an MSVC ifdef which should be windows?

gacholio:
Hmm, looks like OMR believes both _MSC_VER and __clang__ will be defined at the same time.

ChengJin01:
Only in /runtime/vm/BytecodeInterpreter.cpp:
#if defined(_MSC_VER)
/* MSVC compiler hangs on this file at max opt (/Ox) */
#pragma optimize( "g", off )
#endif /* _MSC_VER */

gacholio:
Is _MSC_VER defined in your builds? If so, the above code is going to cripple the interpreter performance, but should not be incorrect (and certainly has nothing to do with signal handling).

babsingh:
agreed with @gacholio. this is unrelated to signal handling. `_MSC_VER` is only defined when compiling with the Visual Studio C++ compiler.

gacholio:
OMR cotains this (possibly bogus) code:

```
#if defined(_MSC_VER)
  #ifdef __clang__ 
    #pragma clang optimize off
  #else
    #pragma optimize("y",off)
  #endif /* __clang__ */
#elif defined(__MINGW32__)
  #pragma GCC optimize ("O0")
#endif /* _MSC_VER */
```

gacholio:
I think you'll find neither of those statements about clang are true. OMR is compiled with whatever people choose, and _MSC_VER is likely defined on windows clang for compatibility reasons (which is actually pretty annoying). Not clear to me if the MSVC-specific pragma will have any effect on clang or not.

pshipton:
> Not clear to me if the MSVC-specific pragma will have any effect on clang or not.

I'd guess not, as we tested performance of a 64-bit build with clang and it was better than with mingw.

ChengJin01:
> I suspect the root issue lies in the exception throw entry to the interpreter from the JIT (this path is never taken in pure interpreter mode). ....

If so,  at least the code in /runtime/compiler/runtime/SignalHandler.c needs to be analyzed because 64bit and 32bit are handled in different ways from JIT in the case of jitHandleIntegerDivideByZeroTrap:

64bit ---> UDATA jitAMD64Handler(J9VMThread* vmThread, U_32 sigType, void *sigInfo)
32bit ---> UDATA jitX86Handler(J9VMThread* vmThread, U_32 sigType, void* sigInfo)

0dvictor:
> Or there might be something wrong in handler registration on win32 as compared to win64amd since Clang definitely changes something there which is different from mingw.

This is very-likely the case. Otherwise, the debugger should reach VM's master signal handler after divide-by-zero is trapped.

gacholio:
Are only the files currently compiled with mingw now using clang? None of them have anything to do with signal handling.

ChengJin01:
> Are only the files currently compiled with mingw now using clang? 
Yes, they is the only changes we had now.

ChengJin01:
[clang_jdk8_32bit_windbg_log.txt](https://github.com/eclipse/openj9/files/2439817/clang_jdk8_32bit_windbg_log.txt)
[jvmwi3280_windbg_log.txt](https://github.com/eclipse/openj9/files/2439818/jvmwi3280_windbg_log.txt)
[mingw_jdk8_32bit_windbg_log.txt](https://github.com/eclipse/openj9/files/2439819/mingw_jdk8_32bit_windbg_log.txt)

With windbg(x86), I was able to debug from `j9jit29!throwCurrentExceptionFromJIT`
  on jvmwi3280 (created internally) and on mingw_jdk8_32bit (compiled for openjdk8_openj9) 

However, it always failed to set up the same breakpoint on clang compiled build as follows:
```
*** ERROR: Symbol file could not be found.  Defaulted to export symbols for C:\JCHAU\java8_32bit\clang_jdk8_32bit\windows-x86-normal-server-release\images\j2sdk-image\jre\bin\default\j9jit29.dll - 
0:006> bu  j9jit29!throwCurrentExceptionFromJIT
Couldn't resolve error at 'j9jit29!throwCurrentExceptionFromJIT'
```
(the error above never happened on jvmwi3280 and mingw_jdk8_32bit)

Not sure whether this error at the breakpoint has nothing to do with the test failure.


In addition, I used to add -Wno-ignored-attributes to the compiler flag for clang to ignore the error on JNICALL as follows:
```
In file included from ../oti\j9.h:45:
../include\jni.h:227:11: error: stdcall calling convention ignored on variadic function [-Werror,-Wignored-attributes]
        jobject (JNICALL * NewObject)(JNIEnv *env, jclass clazz, jmethodID methodID, ...);
                 ^
../include/jniport.h:29:17: note: expanded from macro 'JNICALL'
#define JNICALL __stdcall
                ^
In file included from BytecodeInterpreter.cpp:24:
```
I am wondering whether this is related to the exception creation via runCallInMethod()
```
void JNICALL
runCallInMethod(JNIEnv *env, jobject receiver, jclass clazz, jmethodID methodID, void* args)
```
as runCallInMethod uses **void\* args** as variadic  arguments, which I suspect might affect the cleanup of the call stack but it doesn't help at this point.
(I tried to change JNICALL to **__cdecl** but the test still ended up with the same failure after compilation)

ChengJin01:
Already talked to @babsingh, we suspect there might be something not correctly set up in terms of -Xrs options in signal handling against the doc at https://www.ibm.com/support/knowledgecenter/en/SSYKE2_8.0.0/com.ibm.java.vm.80.doc/docs/xrs.html. Need to investigate further in there.

gacholio:
No, the interpreter does not do anything but read the ELS (very occasionally). ELS is set up in callin.cpp, which then calls the interpreter.

Note 32-bit windows needs this:

```
#if defined(WIN32) && !defined(J9VM_ENV_DATA64)
extern void __cdecl win32ExceptionHandler(void);
extern UDATA getFS0(void);
extern void setFS0(UDATA);
#endif /* WIN32 && !J9VM_ENV_DATA64 */
```

But if you're not compiling callin.cpp with clang, then I don't see why it would be different.

ChengJin01:
@gacholio,  That's why it only fails on Windows 32bit without J9_SIG_XRS_SYNC because win32ExceptionHandler only gets registered in such case.

>the interpreter does not do anything but read the ELS (very occasionally).

The test passed without the code of setting up win32ExceptionHandler in ELS and it failed after adding the setting-up code back).  If the address of win32ExceptionHandler in ELS didn't get screwed up in interperter, then why win32ExceptionHandler didn't get triggered?  Is there any other reason for this behavior?


The test can be simplified to only one line to trigger the issue.
```
public class DivByZero3 {
	public int returnOne() {
     return 1/0;  <------------------------
	}
	public static void main(String[] args) {
		DivByZero3 db = new DivByZero3();
		db.returnOne();
	   }
}
```

ChengJin01:
The issue with win32ExceptionHandler might be a red herring and there is something else that needs to be checked.

ChengJin01:
After comparing the backtraces form the mingw compiled build,  I managed to reach somewhere in the kernel mode on Windows so as to figure out where the second exception came from.
[clang_debug_RtlIsValidHandler.txt](https://github.com/eclipse/openj9/files/2465477/clang_debug_RtlIsValidHandler.txt)
[mingw_debug_RtlIsValidHandler.txt](https://github.com/eclipse/openj9/files/2465478/mingw_debug_RtlIsValidHandler.txt)

The debugging shows that the unknown exception was triggered in **RtlIsValidHandler** because **RtlpxLookupFunctionTable** went wrong in logic.

Note: RtlIsValidHandler is validating our exception handler when its dispatching exceptions according to SAFESEH
(Please refer to 
https://www.optiv.com/blog/old-meets-new-microsoft-windows-safeseh-incompatibility
https://msdn.microsoft.com/en-us/library/9a89h429.aspx for details as to /SAFESEH option)

```
0:006> k
 # ChildEBP RetAddr  
00 233b21a4 77b42b04 ntdll!RtlIsValidHandler <------- exception thrown out
01 233b223c 77b4c716 ntdll!RtlDispatchException+0x11c
02 233b223c 239c003d ntdll!KiUserExceptionDispatcher+0x26
WARNING: Frame IP not in any known module. Following frames may be wrong.
03 233b2780 53d12890 0x239c003d
04 233b2900 005ae948 j9jit29!returnFromJIT1 [C:\temp_jdk8\openj9-openjdk-jdk8-clang\build\windows-x86-normal-server-release\vm\codert_vm\xnathelp.asm @ 6789] 
05 233b2904 233b279c 0x5ae948
06 233b2908 00000000 0x233b279c

0:006> 
WARNING: Stack pointer is outside the normal stack bounds. Stack unwinding can be inaccurate.
eax=acfa8d2b ebx=583594c0 ecx=583594c0 edx=233b2128 esi=233b2254 edi=233b22a4
eip=77b42bee esp=233b2100 ebp=233b21a4 iopl=0         nv up ei ng nz na pe cy
cs=0023  ss=002b  ds=002b  es=002b  fs=0053  gs=002b             efl=00200287
ntdll!RtlIsValidHandler+0x36:
77b42bee e87da1fdff      call    ntdll!RtlpxLookupFunctionTable (77b1cd70) <---- eax was incorrectly set up (it was supposed to be zero) in RtlpxLookupFunctionTable 
0:006> 
WARNING: Stack pointer is outside the normal stack bounds. Stack unwinding can be inaccurate.
eax=583fb770 ebx=583594c0 ecx=00000011 edx=ebb3ee90 esi=233b2254 edi=233b22a4
eip=77b42bf3 esp=233b2100 ebp=233b21a4 iopl=0         nv up ei pl nz na po nc
cs=0023  ss=002b  ds=002b  es=002b  fs=0053  gs=002b             efl=00200202
ntdll!RtlIsValidHandler+0x3b:
77b42bf3 8b55a0          mov     edx,dword ptr [ebp-60h] ss:002b:233b2144=0000000d
0:006> 
WARNING: Stack pointer is outside the normal stack bounds. Stack unwinding can be inaccurate.
eax=583fb770 ebx=583594c0 ecx=00000011 edx=0000000d esi=233b2254 edi=233b22a4
eip=77b42bf6 esp=233b2100 ebp=233b21a4 iopl=0         nv up ei pl nz na po nc
cs=0023  ss=002b  ds=002b  es=002b  fs=0053  gs=002b             efl=00200202
ntdll!RtlIsValidHandler+0x3e:
77b42bf6 85c0            test    eax,eax
0:006> 
WARNING: Stack pointer is outside the normal stack bounds. Stack unwinding can be inaccurate.
eax=583fb770 ebx=583594c0 ecx=00000011 edx=0000000d esi=233b2254 edi=233b22a4
eip=77b42bf8 esp=233b2100 ebp=233b21a4 iopl=0         nv up ei pl nz na po nc
cs=0023  ss=002b  ds=002b  es=002b  fs=0053  gs=002b             efl=00200202
ntdll!RtlIsValidHandler+0x40:
77b42bf8 0f8492a60300    je      ntdll!RtlIsValidHandler+0x3a6d8 (77b7d290) [br=0]

77b7d2ec e8ef330100      call    ntdll!RtlReportException (77b906e0) <------------------
0:006> 
(2d7c.6c0): Unknown exception - code c00001a5 (!!! second chance !!!)
```

ChengJin01:
Looking at the debugging log in RtlpxLookupFunctionTable:
[clang_debug_RtlpxLookupFunctionTable.txt](https://github.com/eclipse/openj9/files/2465512/clang_debug_RtlpxLookupFunctionTable.txt)
[mingw_debug_RtlpxLookupFunctionTable.txt](https://github.com/eclipse/openj9/files/2465513/mingw_debug_RtlpxLookupFunctionTable.txt)

As compared to mingw:
```
0:006> 
WARNING: Stack pointer is outside the normal stack bounds. Stack unwinding can be inaccurate.
eax=00000011 ebx=030e42d8 ecx=00000011 edx=ebb3ee90 esi=77bf8654 edi=5794da40
eip=77b1ce67 esp=030e42a0 ebp=030e42d0 iopl=0         nv up ei pl zr na pe nc
cs=0023  ss=002b  ds=002b  es=002b  fs=0053  gs=002b             efl=00200246
ntdll!RtlpxLookupFunctionTable+0xf7:
77b1ce67 8b45f4          mov     eax,dword ptr [ebp-0Ch] ss:002b:030e42c4=00000000 <---------
0:006> 
WARNING: Stack pointer is outside the normal stack bounds. Stack unwinding can be inaccurate.
eax=00000000 ebx=030e42d8 ecx=00000011 edx=ebb3ee90 esi=77bf8654 edi=5794da40
eip=77b1ce6a esp=030e42a0 ebp=030e42d0 iopl=0         nv up ei pl zr na pe nc
cs=0023  ss=002b  ds=002b  es=002b  fs=0053  gs=002b             efl=00200246
ntdll!RtlpxLookupFunctionTable+0xfa:
77b1ce6a 85c0            test    eax,eax <---------- eax = 0
0:006> 
WARNING: Stack pointer is outside the normal stack bounds. Stack unwinding can be inaccurate.
eax=00000000 ebx=030e42d8 ecx=00000011 edx=ebb3ee90 esi=77bf8654 edi=5794da40
eip=77b1ce6c esp=030e42a0 ebp=030e42d0 iopl=0         nv up ei pl zr na pe nc
cs=0023  ss=002b  ds=002b  es=002b  fs=0053  gs=002b             efl=00200246
ntdll!RtlpxLookupFunctionTable+0xfc:
77b1ce6c 0f8445020000    je      ntdll!RtlpxLookupFunctionTable+0x347 (77b1d0b7) [br=1]
```

the clang build failed at:
```
0:006> 
WARNING: Stack pointer is outside the normal stack bounds. Stack unwinding can be inaccurate.
eax=00000011 ebx=22f531e0 ecx=00000011 edx=ebb3ee90 esi=77bf8654 edi=583094c0
eip=77b1ce67 esp=22f531a8 ebp=22f531d8 iopl=0         nv up ei pl zr na pe nc
cs=0023  ss=002b  ds=002b  es=002b  fs=0053  gs=002b             efl=00200246
ntdll!RtlpxLookupFunctionTable+0xf7:
77b1ce67 8b45f4          mov     eax,dword ptr [ebp-0Ch] ss:002b:22f531cc={j9vm29!__safe_se_handler_table (583ab770)} <-------- __safe_se_handler_table is never seen in the mingw build
0:006> 
WARNING: Stack pointer is outside the normal stack bounds. Stack unwinding can be inaccurate.
eax=583ab770 ebx=22f531e0 ecx=00000011 edx=ebb3ee90 esi=77bf8654 edi=583094c0
eip=77b1ce6a esp=22f531a8 ebp=22f531d8 iopl=0         nv up ei pl zr na pe nc
cs=0023  ss=002b  ds=002b  es=002b  fs=0053  gs=002b             efl=00200246
ntdll!RtlpxLookupFunctionTable+0xfa:
77b1ce6a 85c0            test    eax,eax <------------ eax=583ab770 is not 0.
0:006> 
WARNING: Stack pointer is outside the normal stack bounds. Stack unwinding can be inaccurate.
eax=583ab770 ebx=22f531e0 ecx=00000011 edx=ebb3ee90 esi=77bf8654 edi=583094c0
eip=77b1ce6c esp=22f531a8 ebp=22f531d8 iopl=0         nv up ei pl nz na po nc
cs=0023  ss=002b  ds=002b  es=002b  fs=0053  gs=002b             efl=00200202
ntdll!RtlpxLookupFunctionTable+0xfc:
77b1ce6c 0f8445020000    je      ntdll!RtlpxLookupFunctionTable+0x347 (77b1d0b7) [br=0]
```
I doubled-checked all code in openjdk8/openj9/omr and didn't find out __safe_se_handler_table.
However, it does appear in j9vm29.dll in the clang build as follows:

```
  10A000 RVA,       68 SizeOfBlock
     D6C  HIGHLOW            10129460  ___security_cookie
     D70  HIGHLOW            1010B770  ___safe_se_handler_table <-----------------
     DFC  HIGHLOW            10124A14  ??_R0?AVCursor@@@8 (class Cursor `RTTI Type Descriptor')
```
That means ___safe_se_handler_table was intentionally inserted by specifying the /SAFESEH option in the compiler or linker flags.

In addition, Google used to compile Chrome with Clang and figured that "Clang is SEH safe because it doesn't support SEH and therefore has no SEH handlers to register in .sxdata"
(Please refer to 
https://bugs.chromium.org/p/chromium/issues/detail?id=82385&cnum=100&cstart=150
http://lists.llvm.org/pipermail/llvm-bugs/2013-September/030264.html for details)

That being said, there might be problem with Clang even though it is safe to specifying /SAFESEH in the compiler/linker option on Windows 32bit.

In our case,  it appears that the inserted **___safe_se_handler_table** messed up the code logic in RtlpxLookupFunctionTable, which eventually led to the second exception.

If ___safe_se_handler_table (never exists in j9vm29.dll of the mingw build) is not expected in our implementation, we need to figure out how to remove it (e.g. remove all /SAFESEH in config of openjdk8/oepnj9/omr) to see whether it works.

ChengJin01:
Search for ___safe_se_handler_table in .pdb files:
[clang_grep_SEH_pdb.txt](https://github.com/eclipse/openj9/files/2465732/clang_grep_SEH_pdb.txt)
[mingw_grep_SEH_pdb.txt](https://github.com/eclipse/openj9/files/2465733/mingw_grep_SEH_pdb.txt)

___safe_se_handler_table is only found in j9vm29.pdb of the clang build:
Binary file j9vm29.pdb matches

Search for ___safe_se_handler_table in .map files:
[clang_grep_SEH_map.txt](https://github.com/eclipse/openj9/files/2465741/clang_grep_SEH_map.txt)
[mingw_grep_SEH_map.txt](https://github.com/eclipse/openj9/files/2465742/mingw_grep_SEH_map.txt)

___safe_se_handler_table doesn't exist in j9vm.map of the mingw build:
vm/j9vm.map: 0000:00000000       ___safe_se_handler_table   00000000     <absolute>
vm/j9vm.map: 0000:00000000       ___safe_se_handler_count   00000000     <absolute>

___safe_se_handler_table is also found j9vm.map of the clang build (a safe_se_handler entry was inserted there):
vm/j9vm.map: 0002:0001b770       ___safe_se_handler_table   1010b770     <linker-defined>
vm/j9vm.map: 0000:00000001       ___safe_se_handler_count   **00000001**     <absolute>

pshipton:
mingw doesn't support safeseh and OpenJ9 should be linking j9vm29.dll without this option. However it should be enabled for other modules. If clang supports it then it can be enabled for j9vm29.dll as well.

ChengJin01:
As mentioned before as to the result from Chrome compiled with Clang,  Clang doesn't support SEH.

gacholio:
The Win64 port library does use SEH. I'm not sure if that means only the port library needs to have the magic flag, or if anyone using signal protection also needs to have it. The actual use of try/except is limited to the port library, I believe.

gacholio:
I have it backwards, it's Win32 that uses SEH, not Win64.

pshipton:
It should be used by all modules except j9vm29.dll

ChengJin01:
I believe I already figured what happened to SEH on the clang build and the way to fix it up.

1) The hyperlink at https://reverseengineering.stackexchange.com/questions/11297/how-does-windows-safeseh-mechanism-work explains how Windows SafeSEH mechanism work.
Upon ntdll.dll loading each PE image, the list of exception handler addresses in the PE image is parsed and each address is inserted into a sorted list that ntdll.dll internally uses. When an exception hits, ntdll.dll!KiUserExceptionDispatcher will try to figure out which exception in the SEH chain to use. In turn, each exception handler address is scanned for in this list (by ntdll.dll). If it is found, then the exception filter is allowed to run. If not, then ntdll.dll will throw exception and essentially fastfail the thread with 0xC00001A5: An invalid exception handler routine has been detected.
<---- This is pretty much what happened in our case as it was unable to locate the correct exception handler (namely win32ExceptionHandler on Windows 32bit)

2) The explanation at https://clang.llvm.org/docs/MSVCCompatibility.html , https://llvm.org/docs/ExceptionHandling.html and https://lld.llvm.org/windows_support.html indicates:
they already implemented Safe Structured Exception Handler (SEH) in a way which works perfect with their own linker (lld-link.exe) rather than MS Linker (they didn't mention how it works with MS Linker, so there might be compatibility issue in our case)

3) According to at https://reviews.llvm.org/D1691,
the object files are safe by emitting a **@feat.00** absolute address symbol. The address is presumably interpreted as a bitfield of features that the compiler would like to enable. Bit 0 is documented in the PE COFF spec to opt in to "registered SEH", which is what /safeseh enables.

Looking at the obj file at BytecodeInterpreter.obj or MHInterpreter.obj on the clang build:
```
    Section length   62, #relocs    0, #linenums    0, checksum 2866B00A
04C 00000001 ABS    notype       Static       | @feat.00 <-----
04D 00000000 SECT1  notype ()    External     | _bytecodeLoop
```
It means Clang added **@feat.00** to the obj file which triggered MS Linker to set up the safe exception handler table in .dll files (Note: this flag doesn't exist on the mingw build as mingw provides no support on SEH,  so there is no safe exception handler table in j9vm29.dll on the mingw build).

4) Looking at the j9vm29.dll for SEH entry (output from dumpbin.exe)
```
Dump of file j9vm29.dll
File Type: DLL
  Section contains the following load config:

            00000048 size
           ... ...
            10129460 Security Cookie
            1010B770 Safe Exception Handler Table
                   1 Safe Exception Handler Count

    Safe Exception Handler Table
          Address
          --------
          100EA1D5  __except_handler4

__except_handler4:
  100EA1D5: 8B FF              mov         edi,edi
  ...
  100EA1E6: 68 C6 A1 0E 10     push        offset @__security_check_cookie@4
  100EA1EB: 68 60 94 12 10     push        offset ___security_cookie
  100EA1F0: E8 65 07 00 00     call        __except_handler4_common <--- defined in vcruntime.lib
```
Based on the setup, when the divide-by-zero trap happens, it searched the exception handler table for the configured handler (__except_handler4 in our case). In such case, __except_handler4_common got invoked to thrown out the second unknown exception which is not our expectation.

So the root cause here is that the SEH table was not set up correctly for our purpose (we except win32ExceptionHandler to be registered rather than __except_handler4 so win32ExceptionHandler should be picked up first). That being said, there should be additional setting to register our handler to get it collected in MS Linker.

To work around this issue, I tried the example at https://docs.microsoft.com/en-us/cpp/assembler/masm/dot-safeseh?view=vs-2017 to create an safeseh.asm at runtime/vm(compiled by ml.exe and linked to the obj files)  to register the required handler.
```
.386
.model  flat
win32ExceptionHandler   proto
.safeseh    win32ExceptionHandler
end
```
The SEH table of j9vm29.dll as follows after compilation:
```
           1010B770 Safe Exception Handler Table
                   2 Safe Exception Handler Count

    Safe Exception Handler Table

          Address
          --------
          100694C0  _win32ExceptionHandler <-------- already set up correctly
          100EA1D5  __except_handler4
```

The trace shows it already hit _win32ExceptionHandler as expected:
```
WARNING: Stack pointer is outside the normal stack bounds. Stack unwinding can be inaccurate.
 # ChildEBP RetAddr  
00 2530067c 77bc9e22 j9vm29!win32ExceptionHandler [c:\temp_jdk8\openj9-openjdk-jdk8-clang\build\windows-x86-normal-server-release\vm\vm\wi32\inlineseh.c @ 67] 
01 253006a0 77bc9df4 ntdll!ExecuteHandler2+0x26
02 2530076c 77bb0ce6 ntdll!ExecuteHandler+0x24
03 2530076c 255e003d ntdll!KiUserExceptionDispatcher+0x26
WARNING: Frame IP not in any known module. Following frames may be wrong.
04 25300c30 63182890 0x255e003d
05 25300c34 243b2200 j9jit29!returnFromJIT1 [C:\temp_jdk8\openj9-openjdk-jdk8-clang\build\windows-x86-normal-server-release\vm\codert_vm\xnathelp.asm @ 6789] 
```

The failing tests also passed with/without -Xr related options:
```
C:\temp\test>C:\temp_jdk8\openj9-openjdk-jdk8-clang\build\windows-x86-normal-server-release\images\j2sdk-image\jre\bin\java.exe     -Xjit:noJitUntilMain,count=0,optlevel=noopt,limit={DivByZero3.returnOne*}  DivByZero3
Exception in thread "main" java.lang.ArithmeticException: divide by zero
        at DivByZero3.returnOne(DivByZero3.java:3)
        at DivByZero3.main(DivByZero3.java:7)

C:\temp\test>C:\temp_jdk8\openj9-openjdk-jdk8-clang\build\windows-x86-normal-server-release\images\j2sdk-image\jre\bin\java.exe  -Xrs   -Xjit:noJitUntilMain,count=0,optlevel=noopt,limit={DivByZero3.returnOne*}  DivByZero3
Exception in thread "main" java.lang.ArithmeticException: divide by zero
        at DivByZero3.returnOne(DivByZero3.java:3)
        at DivByZero3.main(DivByZero3.java:7)

C:\temp\test>C:\temp_jdk8\openj9-openjdk-jdk8-clang\build\windows-x86-normal-server-release\images\j2sdk-image\jre\bin\java.exe  -Xrs:sync   -Xjit:noJitUntilMain,count=0,optlevel=noopt,limit={DivByZero3.returnOne*}  DivByZero3
Exception in thread "main" java.lang.ArithmeticException: divide by zero
        at DivByZero3.returnOne(DivByZero3.java:3)
        at DivByZero3.main(DivByZero3.java:7)

C:\temp\test>C:\temp_jdk8\openj9-openjdk-jdk8-clang\build\windows-x86-normal-server-release\images\j2sdk-image\jre\bin\java.exe  -Xrs:async   -Xjit:noJitUntilMain,count=0,optlevel=noopt,limit={DivByZero3.returnOne*}  DivByZero3
Exception in thread "main" java.lang.ArithmeticException: divide by zero
        at DivByZero3.returnOne(DivByZero3.java:3)
        at DivByZero3.main(DivByZero3.java:7)
```

Meanwhile the _sanity.functional ended up with the same results as mingw
```
TOTAL: 163   EXECUTED: 94   PASSED: 94   FAILED: 0   SKIPPED: 69
ALL TESTS PASSED
++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++
_sanity.functional done
```

ChengJin01:
According to the explanation at https://docs.microsoft.com/en-us/cpp/build/reference/safeseh-image-has-safe-exception-handlers?view=vs-2017, 
Unless /SAFESEH:NO is specified, it doesn't matter whether to specify /SAFESEH for MS Linker as it always produces a table of the image's safe exception handlers (the only difference there is:  If any modules were not compatible with safe exception handling feature, the resulting image will not contain a table of safe exception handlers when /SAFESEH is not specified).

Given that UMA_SAFESEH_NOT_ENABLED is used to enable /SAFESEH and is not specific to j9vm29.dll, it is hard to say whether it works in our way to register win32ExceptionHandler. But it is worth to try it out to see what happens.

ChengJin01:
I already compiled the build with /SAFESEH specified and safeseh.asm removed.
The failing test ended up with same failure as before (no output).

Looking at the dump file from j9vm29.dll as follows, indicating win32ExceptionHandler didn't get registered automatically.
```
            1010B770 Safe Exception Handler Table
                   1 Safe Exception Handler Count
    Safe Exception Handler Table
          Address
          --------
          100EA1D5  __except_handler4
```

The result is correct as  /SAFESEH is already turned on by default and __except_handler4 is set up by MS Linker as the default handler. It's up to developers to registers their own handler if required.

In addition, I already removed UMA_SAFESEH_NOT_ENABLED in changes as it is useless for clang.

pshipton:
Excellent. Maybe we should switch to clang 7 before this is deployed? You could try clang 7 and check if its also working.

keithc-ca:
Why wouldn't we want to adopt a more recent version of clang? My mac has `Apple LLVM version 10.0.0 (clang-1000.11.45.2)`; presumably clang 10 is available on Windows?

babsingh:
probably because there are compiler requirements associated with JDK source: https://wiki.openjdk.java.net/display/Build/Supported+Build+Platforms. Although OpenJ9 may compile with a leading edge clang, JDK source may yield errors.

pshipton:
@keithc-ca clang 7 is the most recent http://releases.llvm.org/download.html
@babsingh we only use clang to compile the bytecode interpreter

ChengJin01:
> Why wouldn't we want to adopt a more recent version of clang? My mac has Apple LLVM version 10.0.0 (clang-1000.11.45.2); presumably clang 10 is available on Windows?

I believe you are talking about Xcode/Clang instead of LLVM/Clang on Windows, which is different in terms of version number (please see https://en.wikipedia.org/wiki/Xcode).
The latest version of LLVM/Clang is 7.0.0 at http://releases.llvm.org/download.html

keithc-ca:
Thanks for the links. @ChengJin01.

ChengJin01:
I already finished compiling Java 8 (32bit & 64bit) and Java 11 64bit with Clang7.0.0 and the _sanity.functional test shows all required tests passed without failure detected.

Java 8 32bit:
```
TOTAL: 163   EXECUTED: 94   PASSED: 94   FAILED: 0   SKIPPED: 69
ALL TESTS PASSED
++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++
_sanity.functional done
Administrator@JCHTORWIN101 /cygdrive/c/temp_jdk8/openj9/test/TestConfig
$ echo $JAVA_HOME
C:\temp_jdk8\perf_win\build_jdk8_clang7_32bit_seh_v1\windows-x86-normal-server-release\images\j2sdk-image
```
Java 8 64bit:
```
TOTAL: 163   EXECUTED: 87   PASSED: 87   FAILED: 0   SKIPPED: 76
ALL TESTS PASSED
++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++
_sanity.functional done
Administrator@JCHTORWIN101 /cygdrive/c/temp_jdk8/openj9/test/TestConfig
$ echo  $JAVA_HOME
C:\temp_jdk8\perf_win\build_jdk8_clang7_64bit_v1\windows-x86_64-normal-server-release\images\j2sdk-image
```

Java 11 64bit:
```
TOTAL: 185   EXECUTED: 106   PASSED: 106   FAILED: 0   SKIPPED: 79
ALL TESTS PASSED
++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++
_sanity.functional done
Administrator@JCHTORWINVM021 /cygdrive/c/temp/openj9-openjdk-jdk11/openj9/test/TestConfig
$ echo $JAVA_HOME
C:\temp\build_jdk11_clang7_64bit_v1\windows-x86_64-normal-server-release\images\jdk

```

As compared to the compilation with Clang6.0.1, the only difference on 7.0.0 is it captures "pragma optimize" as error rather than warning on 6.0.1 (Clang doesn't support this directive even though _MSC_VER is a predefined macro in Clang)
```
/runtime/vm/BytecodeInterpreter.cpp
#if defined(_MSC_VER)
/* MSVC compiler hangs on this file at max opt (/Ox) */
#pragma optimize( "g", off )
#endif /* _MSC_VER */
```

Given that the code here was only used on MSVC and never used on mingw or Clang, I changed the code by disabling Clang to compile on this code so as to maintain the historical background for future reference.
```
#if defined(_MSC_VER) && !defined(__clang__)
```

pshipton:
Next step is the performance check.

ChengJin01:
Will send the packaged builds to Piyush Gupta for performance check (Only for 64bit).

piyush286:
@ChengJin01 Clang7 seems to be around 3% faster for startup and at par for footprint on Liberty DT3.

image

ChengJin01:
Changes are already updated for review if no other concern.

pshipton:
https://github.com/eclipse/openj9/issues/2980 is a prereq to merging this.

ChengJin01:
The work on installing Clang7 will be finished via AdoptOpenJDK/openjdk-infrastructure#514

jdekonin:
jenkins sanity win,win32 jdk8
jenkins sanity win jdk11

jdekonin:
jenkins sanity win,win32 jdk8
jenkins sanity win jdk11

jdekonin:
jenkins sanity win jdk11

pshipton:
jenkins test sanity win,win32 jdk8

pshipton:
jenkins test sanity win jdk11

pshipton:
jenkins test sanity win jdk11

pshipton:
jenkins test sanity win jdk11

pshipton:
> "... and will continue to monitor as we process the backlog of events." 

I wonder if all the PR builds we have requested are queued up ... ?

pshipton:
> I wonder if all the PR builds we have requested are queued up ... ?

Apparently so.

pshipton:
jenkins test sanity win jdk8

pshipton:
@ChengJin01 @jdekonin fyi the build failures

pshipton:
I don't think Joe changed the path to add clang 7, so builds would find clang 5 in cygwin

ChengJin01:
It seems these build failures were caused by the incorrect version of LLVM/Clang on the build machines.
@jdekonin , please ensure the path is correctly set up on 32bit and 64bit.

ChengJin01:
Clang 7 (32bit & 64bit) has been installed on internal/external/Axxon Windows build machines via https://github.ibm.com/runtimes/infrastructure/issues/792. So jenkins test should be safe to go.

pshipton:
After some discussion with @jdekonin, I expect to see https://github.com/eclipse/openj9/pull/3409 as part of this change, rather than as a separate change.

ChengJin01:
Already added the changes at #3409 (mingw removed)

pshipton:
jenkins compile win jdk8

pshipton:
Already referenced above, but fyi the Adopt changes required https://github.com/AdoptOpenJDK/openjdk-build/pull/701

pshipton:
@ChengJin01 please fix the merge conflict

runtime/makelib/targets.mk.win.inc.ftl
```diff
@@ -249,8 +245,8 @@ ifdef UMA_TREAT_WARNINGS_AS_ERRORS
 ifndef UMA_SUPPRESS_WARNINGS_AS_ERRORS
 CFLAGS+=-WX
 CXXFLAGS+=-WX
-  ifdef USE_MINGW
-	MINGW_CXXFLAGS+=-Werror -Wno-deprecated-declarations
+  ifdef USE_CLANG
+	CLANG_CXXFLAGS+=-Werror -Wno-deprecated-declarations -Wno-comment -Wno-ignored-attributes
```
pshipton: Are the `-Wno-comment` and `-Wno-ignored-attributes` options required for the current use of clang to compile the bytecode interpreters? If they are needed, can we change the code to remove the warnings instead?

runtime/oti/clang/stddef.h
```diff
@@ -1,4 +1,23 @@
 /*******************************************************************************
+ * Copyright (c) 2018, 2018 IBM Corp. and others
```
pshipton: This copyright and license should not be added. There is already a license for this file.

runtime/makelib/targets.mk.win.inc.ftl
ChengJin01: These two errors were captured by Clang.

1) -Wno-comment
It was caused by  "/*" in the comment at  /runtime/oti/clang/stddef.h
```
 *		All public headers: include/<star>, arch/<star>/bits/<star>
 *		Startup files: crt/* <-------
 */
```
Add -Wno-comment to let Clang ignore  the **/*** style in checking comment.

2)-Wno-ignored-attributes
The error came from as follows:
```
In file included from ../oti\j9.h:45:
../include\jni.h:227:11: error: stdcall calling convention ignored on variadic function [-Werror,-Wignored-attributes]
        jobject (JNICALL * NewObject)(JNIEnv *env, jclass clazz, jmethodID methodID, ...);
                 ^
../include/jniport.h:29:17: note: expanded from macro 'JNICALL'
#define JNICALL __stdcall
                ^
In file included from BytecodeInterpreter.cpp:24:
```
I don't think we should change the code/logic to fit the stricter compilation rules forced by Clang at this point.

ChengJin01:
Already resolved the conflict in changes

runtime/oti/clang/stddef.h
ChengJin01: Already removed.

runtime/makelib/targets.mk.win.inc.ftl
pshipton: Please change stddef.h from `/*` to `/<start>` and remove the -Wno-comment

@DanHeidinga are you ok with the -Wno-ignored-attributes ? This doesn't seem ideal. We could probably separate out what actually needs to be included, but may take some time.

ChengJin01: > Please change stddef.h from /* to /<star> and remove the -Wno-comment

will update and compile again to make sure it works.

keithc-ca: I think we need to pay attention to the warning about `__stdcall` (which is only relevant for 32-bit ABI): it's a sign the code is not reasonable. The message suggests the compiler is just doing the right thing, so I think the fix could wait (a little while).

pshipton: yes, agreed, I believe Dan does as well.

@ChengJin01 the -Wno-ignored-attributes option needs to be removed. This probably means including more specifically what is required to build the files compiled with clang, so that jni.h doesn't get included.

pshipton: You an also try compiling an example program that uses NewObject with msvc2010, with and without the __stdcall declaration. Look at the code which is generated, and see if its any different.

ChengJin01: Will figure out what needs to be modified to get this compiled  without-Wno-ignored-attributes

ChengJin01:
[test.zip](https://github.com/eclipse/openj9/files/2551090/test.zip)
I already compiled simple tests trailing with  **...)**  (without and without __stdcall) with MSVC2010 (cl.exe 32bit)
```
C:\temp_jdk8\test>cl -v
Microsoft (R) 32-bit C/C++ Optimizing Compiler Version 16.00.40219.01 for 80x86
Copyright (C) Microsoft Corporation.  All rights reserved.

C:\temp_jdk8\test>cl   varargtest.c
Microsoft (R) 32-bit C/C++ Optimizing Compiler Version 16.00.40219.01 for 80x86
Copyright (C) Microsoft Corporation.  All rights reserved.

varargtest.c
Microsoft (R) Incremental Linker Version 10.00.40219.01
Copyright (C) Microsoft Corporation.  All rights reserved.

/out:varargtest.exe
varargtest.obj

C:\temp_jdk8\test>varargtest
 [3.14] [2.72] [1.41]
 [3.99] [2.55] [1.23]

C:\temp_jdk8\test>cl   varargtest2.c
Microsoft (R) 32-bit C/C++ Optimizing Compiler Version 16.00.40219.01 for 80x86
Copyright (C) Microsoft Corporation.  All rights reserved.

varargtest2.c
Microsoft (R) Incremental Linker Version 10.00.40219.01
Copyright (C) Microsoft Corporation.  All rights reserved.

/out:varargtest2.exe
varargtest2.obj

C:\temp_jdk8\test>varargtest2
 [3.14] [2.72] [1.41]
 [3.99] [2.55] [1.23]
```
There is no difference in terms of test result as well as the size of generated obj/exe file (compressed in test.zip)

pshipton:
Please take a look at the function call in the generated assembly code and verify the generated code is the same in both cases.

pshipton:
If it is, the next step is to remove the JNICALL from the var args JNI functions, and check if this removes the need for -Wno-ignored-attributes

ChengJin01:
[varargtest_obj_asm.txt](https://github.com/eclipse/openj9/files/2554254/varargtest_obj_asm.txt)
[varargtest2_obj_asm.txt](https://github.com/eclipse/openj9/files/2554255/varargtest2_obj_asm.txt)

The comparison shows the generated assembly code (via dumpbin) are exactly the same as each other.
Will compile builds on Java 8 and 11 after removing the JNICALL from the var args JNI functions and -Wno-ignored-attributes to see whether the compilations work good as expected.

keithc-ca:
@ChengJin01 One aspect I don't see covered in your test code is the attribute on the function whose address is captured in the function pointer. All the instances of function implementations I found were adorned with the `JNICALL` annotation. We may need to remove those from variadic functions lest clang warn about the mismatch of annotations.

ChengJin01:
As suggested by @keithc-ca,   I double-checked the compilation of test code in which both the function and the function pointer have the same attributes (both are JNICALL in varargtest3 or neither is in varargtest)

Compilation on Clang 32 bit shows varargtest3 with JNICALL ended up with warning which can be bypassed by specifying -Wno-ignored-attributes
```
$ clang -v
clang version 7.0.0 (tags/RELEASE_700/final)
Target: i686-pc-windows-msvc
Thread model: posix
InstalledDir: C:\LLVM_7_32\bin

$ clang  -o varargtest  varargtest.c

$ ./varargtest
 [3.14] [2.72] [1.41]
 [3.99] [2.55] [1.23]

$ clang  -o varargtest3  varargtest3.c
varargtest3.c:5:7: warning: stdcall calling convention ignored on variadic function [-Wignored-attributes]
void (__stdcall *fun_ptr)(int, ...);
      ^
varargtest3.c:7:6: warning: stdcall calling convention ignored on variadic function [-Wignored-attributes]
void __stdcall PrintFloats (int n, ...)
     ^
2 warnings generated.

$ clang  -Wno-ignored-attributes  -o varargtest3  varargtest3.c

$ ./varargtest3
 [3.14] [2.72] [1.41]
 [3.99] [2.55] [1.23]
```

Compilation of varargtest3 on CL 32bit:
```
C:\temp_jdk8\test>cl   varargtest3.c
Microsoft (R) 32-bit C/C++ Optimizing Compiler Version 16.00.40219.01 for 80x86
Copyright (C) Microsoft Corporation.  All rights reserved.

varargtest3.c
Microsoft (R) Incremental Linker Version 10.00.40219.01
Copyright (C) Microsoft Corporation.  All rights reserved.

/out:varargtest3.exe
varargtest3.obj

C:\temp_jdk8\test>varargtest3
 [3.14] [2.72] [1.41]
 [3.99] [2.55] [1.23]
```
[varargtest3.zip](https://github.com/eclipse/openj9/files/2554810/varargtest3.zip)
Also double-check the generated assembly code from varargtest3 which is the same as varargtest.

ChengJin01:
Already compiled on Java 8 (32bit & 64bit) and Java 11 (64bit) without any error.
The new changes include (removed JNICALL from functions and function pointers involved)
```
/runtime/include/j9vmls.h
/runtime/oti/vm_api.h
/runtime/vm/jniinv.c

/runtime/include/jni.h
/runtime/jnichk/jnicwrappers.c

/runtime/include/jvmti.h
/runtime/jvmti/jvmti_internal.h
/runtime/jvmti/jvmtiEventManagement.c

/runtime/oti/clang/stddef.h (changed to crt/<star> in comment)
/runtime/makelib/targets.mk.win.inc.ftl (removed -Wno-ignored-attributes)
```

buildenv/jenkins/variables/defaults.yml
```diff
@@ -347,9 +347,9 @@ win_x86-64_cmprssptrs:
       next: 'hw.arch.x86 && sw.os.windows'
   build_env:
     vars:
-      8: 'PATH+TOOLS=/cygdrive/c/openjdk/nasm-2.13.03/'
-      11: 'PATH+TOOLS=/cygdrive/c/openjdk/nasm-2.13.03/:/cygdrive/c/mingw-w64/x86_64-8.1.0-win32-seh-rt_v6-rev0/mingw64/bin/'
-      next: 'PATH+TOOLS=/cygdrive/c/openjdk/nasm-2.13.03/:/cygdrive/c/mingw-w64/x86_64-8.1.0-win32-seh-rt_v6-rev0/mingw64/bin/'
+      8: 'PATH+TOOLS=/cygdrive/c/openjdk/LLVM64/bin/:/cygdrive/c/openjdk/nasm-2.13.03/'
```
keithc-ca: I think it is unwise to include the trailing `/` in elements of the path.
They should also be removed from the build instructions.

gacholio:
I haven't been following closely, but I'd like a chance to review before the final commit - removing JNICALL seems like a hacky answer (again, not up on all the details).

0dvictor:
I'm not sure if removing `JNICALL` is better than supplying `-Wno-ignored-attributes`. `JNICALL` is defined as `__stdcall` on 32-bit Windows, as a callee-cleanup calling convention, `__stdcall` doesn't support variadic arguments. MSVC resolves the conflicts by treating the calling convention as `cdecl` without generating any warning; on the other hand, clang generates warnings in this situation. By setting `-Wno-ignored-attributes`, clang would have compatible behavior as MSVC.
In addition, remove `JNICALL` from all platforms may be dangerous. What if another platform defines `JNICALL` to something else?

gacholio:
JNICALL is define only for Windows 32-bit. Every other platform, the define is empty.

0dvictor:
If we must avoid setting `-Wno-ignored-attributes` globally, we may also apply the property locally via `pragma` or similar.

0dvictor:
> JNICALL is define only for Windows 32-bit. Every other platform, the define is empty.

True, but whoever writes a JNI method or a method interacts with a JNI method, one shouldn't need to know what `JNICALL` is defined to, not to mention the calling convention. One should be able to blindly add `JNICALL` to the definition of a JNI method.

gacholio:
Agreed - the error sounds like more pointless compiler pedantry, not realizing there's a much larger world out there,

keithc-ca:
> __stdcall doesn't support variadic arguments. MSVC resolves the conflicts ...

Removing JNICALL avoids these 'conflicts' for everyone (the VM and third-party authors of JNI code): it's just the _right thing to do_.

buildenv/jenkins/variables/defaults.yml
ChengJin01: Already removed.

runtime/include/j9vmls.h
```diff
@@ -45,8 +45,8 @@ extern "C" {
  * The VM local storage function table.
  */
 typedef struct J9VMLSFunctionTable {
-    UDATA  (JNICALL *J9VMLSAllocKeys)(JNIEnv * env, UDATA * pInitCount, ...) ;
-    void  (JNICALL *J9VMLSFreeKeys)(JNIEnv * env, UDATA * pInitCount, ...) ;
+    UDATA  (*J9VMLSAllocKeys)(JNIEnv * env, UDATA * pInitCount, ...) ;
+    void  (*J9VMLSFreeKeys)(JNIEnv * env, UDATA * pInitCount, ...) ;
```
gacholio: Why change only two of the four?

ChengJin01: Because the other two functions are not vararg, which is acceptable for Clang.
```
    void*  (JNICALL *J9VMLSGet)(JNIEnv * env, void * key) ;
    void*  (JNICALL *J9VMLSSet)(JNIEnv * env, void ** pKey, void * value) ;
```

ChengJin01:
https://msdn.microsoft.com/en-us/library/zxk0tw93.aspx (__stdcall)
https://msdn.microsoft.com/en-us/library/zkwh89ks.aspx (__cdecl )

According to the explanation on MSDN, MSVC treats __stdcall (JNICALL)  as __cdecl by default on 32bit.
Given that __cdecl is the default calling convention for C and C++ programs on MSVC, whether to remove JNICALL  or not won't affect the result in compilation from MSVC perspective.

If JNICALL must be kept for Spec, another alternative could be: 
add JNICALL2 to /runtime/include/jniport.h and replace  JNICALL with JNICALL2 at all vararg functions in our case.
```
#if defined(WIN32) || ...
...
#if !defined(__clang__)
JNICALL2  __stdcall  (specific to all vararg functions involved in our case)
else
  JNICALL2 
#endif

#else
...
```
That being said,  JNICALL2  is __stdcall  for MSVC and empty for Clang during compilation.

gacholio:
It's the opposite - stdcall is only for 32-bit. We cannot remove this and continue to interact with user JNI libraries.

Given that we only compile 3 files with clang, I think suppressing the warning is the way to go, rather than making far-reaching changes.

gacholio:
@DanHeidinga Do you want to chime in here?

gacholio:
While you may not be wrong, this qualifies as "the hack heard round the world". I'd rather not change 100s of lines of code needlessly.

pshipton:
@ChengJin01 a couple more things to check. Write a JNI example that calls a var args function.
1) Compile it against your JVM that doesn't use JNICALL for var args. Then try to run the JNI example with OpenJDK
2) Compile it using clang, using the OpenJDK jni.h and see if there are any warnings/errors

ChengJin01:
> a couple more things to check. Write a JNI example that calls a var args function.

Already finished the check with 
[jnitest.zip](https://github.com/eclipse/openj9/files/2563876/jnitest.zip)

The test is simply to call NewObject() from jni.h and the results are as follows:
1) task 1 (works good)
```
[1] Compile it against your JVM that doesn't use JNICALL for var args. and c file was compiled by mingw
$ /cygdrive/c/temp_jdk8/build_clang7_32_stdcall/windows-x86-normal-server-release/images/j2sdk-image/bin/javac JniVarArgTest.java -h . 
                
$ i686-w64-mingw32-g++ -Wl,--add-stdcall-alias -I"/cygdrive/c/temp_jdk8/build_clang7_32_stdcall/windows-x86-normal-server-release/images/j2sdk-image/include" -I"/cygdrive/c/temp_jdk8/build_clang7_32_stdcall/windows-x86-normal-server-release/images/j2sdk-image/include/win32" -shared -o JniVarArgTest.dll JniVarArgTest.c

[2] run the binary JNI example with OpenJDK/Hotspot
$ /cygdrive/c/temp_jdk8/jdk8u192-b12_hotspot_win_x86_32/bin/java  JniVarArgTest
result = 0x32 
```

2) task 2  (var args warning with NewObject was captured by Clang )
[clang_warning_error_stdcall.txt](https://github.com/eclipse/openj9/files/2563901/clang_warning_error_stdcall.txt 

Compile java source with my build and  compile the c file using clang 32bit with jni.h from OpenJDK/Hotspot
```
$ /cygdrive/c/temp_jdk8/build_clang7_32_stdcall/windows-x86-normal-server-release/images/j2sdk-image/bin/javac JniVarArgTest.java -h .  

$ clang -Wl,--add-stdcall-alias -I"c:\temp_jdk8\jdk8u192-b12_hotspot_win_x86_32\include" -I"c:\temp_jdk8\jdk8u192-b12_hotspot_win_x86_32\include\win32" -shared -o JniVarArgTest.dll JniVarArgTest.c
In file included from JniVarArgTest.c:1:
In file included from ./JniVarArgTest.h:2:
c:\temp_jdk8\jdk8u192-b12_hotspot_win_x86_32\include\jni.h:277:14: warning: stdcall calling convention ignored on variadic function [-Wignored-attributes]
    jobject (JNICALL *NewObject)
             ^
c:\temp_jdk8\jdk8u192-b12_hotspot_win_x86_32\include\win32\jni_md.h:31:17: note: expanded from macro 'JNICALL'
#define JNICALL __stdcall
                ^
In file included from JniVarArgTest.c:1:
In file included from ./JniVarArgTest.h:2:
c:\temp_jdk8\jdk8u192-b12_hotspot_win_x86_32\include\jni.h:292:14: warning: stdcall calling convention ignored on variadic function [-Wignored-attributes]
    jobject (JNICALL *CallObjectMethod)
             ^
c:\temp_jdk8\jdk8u192-b12_hotspot_win_x86_32\include\win32\jni_md.h:31:17: note: expanded from macro 'JNICALL'
#define JNICALL __stdcall
...
```

ChengJin01:
> a couple more things to check. Write a JNI example that calls a var args function...

Already finished the check with 
[jnitest.zip](https://github.com/eclipse/openj9/files/2563876/jnitest.zip)

The test is simply to call NewObject() from jni.h and the results are as follows:
1) task 1 (**works but images/j2sdk-image/include/jni.h is still from Hotspot rather than the modified jni.h**)
[1] Compile it against your JVM that doesn't use JNICALL for var args. and c file was compiled by mingw
```
$ /cygdrive/c/temp_jdk8/build_clang7_32_stdcall/windows-x86-normal-server-release/images/j2sdk-image/bin/javac JniVarArgTest.java -h . 

$ i686-w64-mingw32-g++ -Wl,--add-stdcall-alias -I"/cygdrive/c/temp_jdk8/build_clang7_32_stdcall/windows-x86-normal-server-release/images/j2sdk-image/include" -I"/cygdrive/c/temp_jdk8/build_clang7_32_stdcall/windows-x86-normal-server-release/images/j2sdk-image/include/win32" -shared -o JniVarArgTest.dll JniVarArgTest.c
```
[2] run the binary JNI example with OpenJDK/Hotspot
```
$ /cygdrive/c/temp_jdk8/jdk8u192-b12_hotspot_win_x86_32/bin/java  JniVarArgTest
result = 0x32 
```

2) task 2  (var args warning with NewObject was captured by Clang )
[clang_warning_error_stdcall.txt](https://github.com/eclipse/openj9/files/2563901/clang_warning_error_stdcall.txt 

Compile the c file using Clang 32bit with jni.h from OpenJDK/Hotspot
```
$ /cygdrive/c/temp_jdk8/build_clang7_32_stdcall/windows-x86-normal-server-release/images/j2sdk-image/bin/javac JniVarArgTest.java -h .  

$ clang -Wl,--add-stdcall-alias -I"c:\temp_jdk8\jdk8u192-b12_hotspot_win_x86_32\include" -I"c:\temp_jdk8\jdk8u192-b12_hotspot_win_x86_32\include\win32" -shared -o JniVarArgTest.dll JniVarArgTest.c
In file included from JniVarArgTest.c:1:
In file included from ./JniVarArgTest.h:2:
c:\temp_jdk8\jdk8u192-b12_hotspot_win_x86_32\include\jni.h:277:14: warning: stdcall calling convention ignored on variadic function [-Wignored-attributes]
    jobject (JNICALL *NewObject)
             ^
c:\temp_jdk8\jdk8u192-b12_hotspot_win_x86_32\include\win32\jni_md.h:31:17: note: expanded from macro 'JNICALL'
#define JNICALL __stdcall
                ^
In file included from JniVarArgTest.c:1:
In file included from ./JniVarArgTest.h:2:
c:\temp_jdk8\jdk8u192-b12_hotspot_win_x86_32\include\jni.h:292:14: warning: stdcall calling convention ignored on variadic function [-Wignored-attributes]
    jobject (JNICALL *CallObjectMethod)
             ^
c:\temp_jdk8\jdk8u192-b12_hotspot_win_x86_32\include\win32\jni_md.h:31:17: note: expanded from macro 'JNICALL'
#define JNICALL __stdcall
...
```
This is because Clang complains any varargs function adorned with __stdcall, no matter where it comes from.

ChengJin01:
The reason that task1 works is because we didn't modify /openj9-openjdk-jdk8/jdk/src/share/javavm/export/jni.h to remove JNICALL from all varargs functions/function pointers. So Clang is still using the unmodified jni.h to compile the test code.

pshipton:
@DanHeidinga thoughts? Should we just use -Wno-ignored-attributes ?

ChengJin01:
After replacing the original jni.h (Hotspot) at  windows-x86-normal-server-release/images/j2sdk-image/include with windows-x86-normal-server-release/vm/include/jni.h (the modified one) ,
I compiled the test code again with mingw and ran with jdk8u192-b12_hotspot_win_x86_32.
The result is the same as task 1 (no problem detected)

DanHeidinga:
> @DanHeidinga thoughts? Should we just use -Wno-ignored-attributes ?

I think that's the right approach so that we stay consistent with OpenJDK. 

If we want to remove the JNICALL from varargs methods, we should also raise a bug at OpenJDK / submit a patch to have the JNI headers updated there and only then make the same change here.

charliegracie:
@gacholio from a quick look at the BytecodeInterpreter it seems as though only `inlAttachmentLoadAgentLibraryImpl` requires `jni.h`.  Is the improvement from having that native as an INL important? My suggestion here is more of a hack than a proper solution of getting OpenJDK to remove the `__stdcall` from the varargs JNI functions but I am still curious.

@ChengJin01 with the move to clang instead of mingw did we test different permutations of the values in `REGISTER_ARGS_LIST` to see if there are any improvements to be had in compile time or in actual throughput of the Interpreter?

@ChengJin01 is Windows still defining `USE_COMPUTED_GOTO` after switching to Clang? I believe it should but I am curious if we verified it.

ChengJin01:
> with the move to clang instead of mingw did we test different permutations of the values ....
We didn't have such tests but we already had perf test to ensure clang meets our expectation at https://github.com/eclipse/openj9/pull/2979#issuecomment-430990446

>  is Windows still defining USE_COMPUTED_GOTO after switching to Clang? I believe it should...
USE_COMPUTED_GOTO is not defined when compiling with Clang because `__GNUC__` is not the predefined macro in Clang (I already verified with test)
[clang_mingw_macros.txt](https://github.com/eclipse/openj9/files/2566454/clang_mingw_macros.txt)

```
/runtime/vm/BytecodeInterpreter.cpp
#if (defined(WIN32) && defined(__GNUC__)) || (defined(LINUX) && defined(J9HAMMER))
#define USE_COMPUTED_GOTO
#else
#undef USE_COMPUTED_GOTO
#endif
```
Meanwhile, defining USE_COMPUTED_GOTO will end up with a bunch of errors during compilation with Clang.
[USE_COMPUTED_GOTO_defined_clang_errors.txt](https://github.com/eclipse/openj9/files/2566464/USE_COMPUTED_GOTO_defined_clang_errors.txt)
```
./BytecodeInterpreter.hpp:9619:4: error: cannot jump from this indirect goto statement to one of its possible targets
                        PERFORM_ACTION(i2b(REGISTER_ARGS));
                        ^
./BytecodeInterpreter.hpp:8535:4: note: expanded from macro 'PERFORM_ACTION'
                        EXECUTE_CURRENT_BYTECODE(); \
                        ^
./BytecodeInterpreter.hpp:8106:36: note: expanded from macro 'EXECUTE_CURRENT_BYTECODE'
#define EXECUTE_CURRENT_BYTECODE() EXECUTE_BYTECODE_NUMBER(*_pc)
                                   ^
./BytecodeInterpreter.hpp:8103:4: note: expanded from macro 'EXECUTE_BYTECODE_NUMBER'
                        goto *(bytecodeTable[number]); \
                        ^
./BytecodeInterpreter.hpp:8786:2: note: possible target of indirect goto statement
        JUMP_TARGET(J9_BCLOOP_SEND_TARGET_INITIAL_STATIC):
        ^
./BytecodeInterpreter.hpp:8097:27: note: expanded from macro 'JUMP_TARGET'
#define JUMP_TARGET(name) c##name
                          ^
<scratch space>:43:1: note: expanded from here
cJ9_BCLOOP_SEND_TARGET_INITIAL_STATIC
./BytecodeInterpreter.hpp:8779:8: note: jump bypasses variable initialization
        void *methodRunAddress = _sendMethod->methodRunAddress;
```

ChengJin01:
I already added -Wno-ignored-attributes back and removed all changes with varargs functions

pshipton:
jenkins test all win jdk8,jdk11

pshipton:
jenkins test all win32 jdk8

pshipton:
jenkins line endings check

pshipton:
jenkins copyright check

pshipton:
jenkins test all win jdk11

pshipton:
jenkins test sanity win jdk11

pshipton:
jenkins test extended win jdk11

gacholio:
Why are there module.xml changes for util? There don't seem to be any clang-compiled files there.

pshipton:
@gacholio note the "runtime/util/mingw_comp.c → runtime/util/clang_comp.c"

ChengJin01:
> Why are there module.xml changes for util? There don't seem to be any clang-compiled files there.

Because `mingw_comp`.h/.c changed to `clang_comp`.h/.c as we should remove all mingw related setting in sript/file naming. So the modulexml has to be modified accordingly.

runtime/oti/clang_comp.h --> runtime/oti/clang_comp.h (name and code changed)
runtime/util/mingw_comp.c  -> runtime/util/clang_comp.c (name changed)

pshipton:
@sxa555 the switch to clang is ready to go. We can coordinate the OpenJ9 change with the Adopt change, maybe on Monday.

pshipton:
@ChengJin01 please fix the conflicts

ChengJin01:
Conflicts in changes have been addressed.

pshipton:
jenkins compile win jdk8

Issue #2957, opened 2018-09-20
Test-sanity.functional-JDK11-linux_x86-64 jsr335_interfacePrivateMethod_mode100_0 Should have thrown an IncompatibleClassChangeError


pshipton:
https://ci.eclipse.org/openj9/job/Test-sanity.functional-JDK11-linux_x86-64/21/tapResults/

```
===============================================
Running test jsr335_interfacePrivateMethod_mode100_0 ...
===============================================
jsr335_interfacePrivateMethod_mode100_0 Start Time: Thu Sep 20 07:17:14 2018 Epoch Time (ms): 1537427834388
test with Mode100
[IncludeExcludeTestAnnotationTransformer] [INFO] exclude file is /home/jenkins/workspace/Test-sanity.functional-JDK11-linux_x86-64/openjdk-tests/TestConfig/scripts/testKitGen/../../../../jvmtest/TestConfig/resources/excludes/latest_exclude_11.txt

...

FAILED: testRunTest_private_non_static_method_with_InvokeVirtual
java.lang.AssertionError: Should have thrown an IncompatibleClassChangeError
	at org.testng.Assert.fail(Assert.java:96)
	at org.openj9.test.jsr335.interfacePrivateMethod.Test_ITestRunner.testRunTest_private_non_static_method_with_InvokeVirtual(Test_ITestRunner.java:112)
	at jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(java.base@11-internal/Native Method)
	at jdk.internal.reflect.NativeMethodAccessorImpl.invoke(java.base@11-internal/NativeMethodAccessorImpl.java:62)
	at jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(java.base@11-internal/DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(java.base@11-internal/Method.java:566)
	at org.testng.internal.MethodInvocationHelper.invokeMethod(MethodInvocationHelper.java:124)
	at org.testng.internal.Invoker.invokeMethod(Invoker.java:580)
	at org.testng.internal.Invoker.invokeTestMethod(Invoker.java:716)
	at org.testng.internal.Invoker.invokeTestMethods(Invoker.java:988)
	at org.testng.internal.TestMethodWorker.invokeTestMethods(TestMethodWorker.java:125)
	at org.testng.internal.TestMethodWorker.run(TestMethodWorker.java:109)
[...]

PR #2981, opened 2018-09-21
Add check before invokePrivate from virtual call

fengxue-IS:
- only allow non-interface method to enter invokePrivate path from `resolveVirtual`
- interface method will be handle by vtable lookup

Close: #2957 

### DIFF
--- a/runtime/vm/resolvesupport.cpp
+++ b/runtime/vm/resolvesupport.cpp
@@ -1574,7 +1574,10 @@ resolveVirtualMethodRefInto(J9VMThread *vmStruct, J9ConstantPool *ramCP, UDATA c
 		if (method != NULL) {
 #if defined(J9VM_OPT_VALHALLA_NESTMATES)
 			J9ROMMethod* romMethod = J9_ROM_METHOD_FROM_RAM_METHOD(method);
-			if (J9_ARE_ALL_BITS_SET(romMethod->modifiers, J9AccPrivate)) {
+			/* Only allow non-interface method to call invokePrivate, private interface method should use "invokeInterface" bytecode
+			 * The else case will throw ICCE for private interface method 
+			 */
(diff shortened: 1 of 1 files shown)


fengxue-IS:
fyi @pshipton

runtime/vm/resolvesupport.cpp
```diff
@@ -1574,7 +1574,10 @@ resolveVirtualMethodRefInto(J9VMThread *vmStruct, J9ConstantPool *ramCP, UDATA c
 		if (method != NULL) {
 #if defined(J9VM_OPT_VALHALLA_NESTMATES)
 			J9ROMMethod* romMethod = J9_ROM_METHOD_FROM_RAM_METHOD(method);
-			if (J9_ARE_ALL_BITS_SET(romMethod->modifiers, J9AccPrivate)) {
+			/* Only allow non-interface method to call invokePrivate, private interface method should use "invokeInterface" bytecode
+			 * The else case will throw ICCE for private interface method 
+			 */
```
DanHeidinga: is this a spec requirement?  Does this match the RIs behaviour?

DanHeidinga:
I'm about to get on an airplane so likely won't get back to this until Saturday at the earliest.
@gacholio would be a good reviewer if this is needed before I'm able to review.

gacholio:
Are sure the test isn't just bogus? Lots of tests needed to change once we starting allowing private methods in JDK11.

pshipton:
Note we only see the failure on xlinuxlargeheap

fengxue-IS:
This test is disabled for compressedrefs, I assume `largeheap` is not using compressedrefs

gacholio:
jenkins test extended xlinuxlargeheap

gacholio:
jenkins test extended xlinuxlargeheap jdk11

runtime/vm/resolvesupport.cpp
fengxue-IS: yes, this does match RIs behaviour

pshipton:
> I assume largeheap is not using compressedrefs

Correct largeheap means non-compressedrefs

PR #2984, opened 2018-09-21
Update resolveConstantDynamic Error wrapping

fengxue-IS:

This is required by JVMS JDK11 Section 5.4.3.6 'Dynamically-Computed Constant and Call Site Resolution'

### DIFF
--- a/jcl/src/java.base/share/classes/java/lang/invoke/MethodHandle.java
+++ b/jcl/src/java.base/share/classes/java/lang/invoke/MethodHandle.java
@@ -853,110 +853,115 @@ public abstract class MethodHandle {
 		Object result = null;
 		Class<?> typeClass = null;
 
+		VMLangAccess access = VM.getVMLangAccess();
+		Object internalRamClass = access.createInternalRamClass(j9class);
+		Class<?> classObject = null;
+		if (JITHELPERS.is32Bit()) {
+			classObject = JITHELPERS.getClassFromJ9Class32((int)j9class);
+		} else {
+			classObject = JITHELPERS.getClassFromJ9Class64(j9class);
+		}
+
+		Objects.requireNonNull(classObject);
+
+		typeClass = fromFieldDescriptorString(fieldDescriptor, access.getClassloader(classObject));
+
+		int bsmIndex = UNSAFE.getShort(bsmData);
+		int bsmArgCount = UNSAFE.getShort(bsmData + BSM_ARGUMENT_COUNT_OFFSET);
+		long bsmArgs = bsmData + BSM_ARGUMENTS_OFFSET;
+		MethodHandle bsm = getCPMethodHandleAt(internalRamClass, bsmIndex);
+		if (null == bsm) {
+			/*[MSG "K05cd", "unable to resolve 'bootstrap_method_ref' in '{0}' at index {1}"]*/
+			throw new NullPointerException(Msg.getString("K05cd", classObject.toString(), bsmIndex)); //$NON-NLS-1$
+		}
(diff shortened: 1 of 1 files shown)


fengxue-IS:
@JasonFengJ9 can you please take a look?

JasonFengJ9:
This change affects all `Java` versions including `Java 8`, please confirm that there is no JCK8 test regression introduced by this PR.

fengxue-IS:
This only affects Java 11 as cp entries that call `resolveConstantDynamic` is only valid starting Java 11

JasonFengJ9:
Added an `new Throwable().printStackTrace();` at beginning of method `java.lang.invoke.MethodHandle.resolveInvokeDynamic()`, with following test code
```
    Consumer<String> c = System.out::println;
    c.accept("Hello!");
```
Run against latest `Java 8` and got following stacktrace:
```
java.lang.Throwable
	at java.lang.invoke.MethodHandle.resolveInvokeDynamic(MethodHandle.java:899)
	at HelloWorld.main(HelloWorld.java:5)
```
This shows that modification to `java.lang.invoke.MethodHandle.resolveInvokeDynamic` affect `Java 8` as well.

fengxue-IS:
@JasonFengJ9 this PR is modifying `ConstantDynamic`, not `invokeDynamic`, please see line 852 just above the first modification

fengxue-IS:
@DanHeidinga please take a look, thanks

jcl/src/java.base/share/classes/java/lang/invoke/MethodHandle.java
```diff
@@ -853,110 +853,110 @@ private static final Object resolveConstantDynamic(long j9class, String name, St
...
+				break;
+			default:
+				// Do nothing. The null check below will throw the appropriate exception.
 			}
 
+			cpEntry.getClass();	// Implicit NPE
+			staticArgs[staticArgIndex] = cpEntry;
+		}
+
+		try {
```
DanHeidinga: Can you add a comment to this try block that references the appropriate part of the JVMS so that we can double check this behaviour in the future?  It makes it much easier to figure out than trying to discern the subtle difference from the code

fengxue-IS:
updated with reference to JVMS, also added a check so exception instance of Error class is not wrapped by extra layer of BootstrapMethodError.

> JVMS 5.4.3.6: If the invocation fails by throwing an instance of Error or a subclass of Error, resolution fails with that exception.

jcl/src/java.base/share/classes/java/lang/invoke/MethodHandle.java
```diff
@@ -853,110 +853,111 @@ private static final Object resolveConstantDynamic(long j9class, String name, St
...
+				break;
+			default:
+				// Do nothing. The null check below will throw the appropriate exception.
 			}
 
+			cpEntry.getClass();	// Implicit NPE
+			staticArgs[staticArgIndex] = cpEntry;
+		}
+
+		/* JVMS JDK11 5.4.3.6 Dynamically-Computed Constant and Call Site Resolution */
```
DanHeidinga:
```
/* JVMS JDK11 5.4.3.6 Dynamically-Computed Constant and Call Site Resolution:
 * requires that only exceptions thrown by the BSM are wrapped in a BootstrapMethodError.
 * Note, subclasses of Error should not be wrapped.
* Exceptions thrown by earlier parts of resolution should be passed through unwrapped.
*/
```

Something like this clearly lays out the rules as they were understood at the time the code was written.  (Please double check I've expressed them correctly).

This makes it easier to know what the code should be doing when debugging something or trying to understand why a particular exception was wrapped / not.

DanHeidinga:

PR #2985, opened 2018-09-21
OpenJ9 fixes for OSX support [Part 3]

babsingh:
**1) Fix value of j9binBuffer/jrebinBuffer [OSX]**

On OSX, `j9binBuffer` and `jrebinBuffer` don't point to the `compressedrefs`
directory. This prevents `jvm.c::preloadLibrary` to derive correct paths
to libraries such as `omrsig`, `j9vm*` etc.

`j9binBuffer` and `jrebinBuffer` are derived using `jvmDLLNameBuffer`.
`jvmDLLNameBuffer` relies on the path to `libj9vm*.so/dll/dylib`. On OSX,
incorrect suffix is used for `libj9vm*.<SUFFIX>`. This causes `j9binBuffer`
and `jrebinBuffer` to be incorrectly set and `jvm.c::preloadLibrary` to
malfunction.

This fix will correctly set the library suffix for `libj9vm*` on OSX.

**2) Fix ifdefs in jvm.c::getj9bin**

Current ifdefs in `jvm.c::getj9bin`:
```
#if (defined(LINUX) && !defined(J9ZTPF)) || defined(OSX)
   /* Code 1 */
#elif defined(J9ZOS390) || defined(J9ZTPF) || defined(OSX)
   /* Code 2 */
#else /* must be AIX / RS6000 */
   /* Code 3 */
#endif
```
Code 1 (#if block) applies to OSX but Code 2 (#elif block) doesn't apply
to OSX. It seems incorrect to include defined(OSX) in both #if and #elif
boolean expressions. defined(OSX) is removed from the #elif boolean
expression.

### DIFF
--- a/runtime/j9vm/jvm.c
+++ b/runtime/j9vm/jvm.c
@@ -1074,7 +1074,7 @@ preloadLibraries(void)
 	/* detect if we're in a subdir or not */
 	jvmDLLNameBuffer = jvmBufferCat(jvmDLLNameBuffer, "/lib");
 	jvmDLLNameBuffer = jvmBufferCat(jvmDLLNameBuffer, vmDllName);
-	jvmDLLNameBuffer = jvmBufferCat(jvmDLLNameBuffer, ".so");
+	jvmDLLNameBuffer = jvmBufferCat(jvmDLLNameBuffer, J9PORT_LIBRARY_SUFFIX);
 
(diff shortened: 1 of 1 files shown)


babsingh:
fyi @DanHeidinga

babsingh:
Related to https://github.com/eclipse/openj9/issues/36.

DanHeidinga:

PR #2987, opened 2018-09-24
Fix stack walker for new interface formats

gacholio:
Fix runtime and DDR stack walkers to understand the expanded iTable
format in the interpreter and the enhanced JIT interface call snippet.

[ci skip]

### DIFF
--- a/runtime/codert_vm/jswalk.c
+++ b/runtime/codert_vm/jswalk.c
@@ -1098,16 +1098,36 @@ static void jitWalkResolveMethodFrame(J9StackWalkState *walkState)
 #endif
 		walkState->unwindSP += getJitRecompilationResolvePushes();
 	} else if (resolveFrameType == J9_STACK_FLAGS_JIT_LOOKUP_RESOLVE) {
-		UDATA * interfaceObjectAndISlot = (UDATA *) JIT_RESOLVE_PARM(2);
-		J9Class * interfaceClass = READ_CLASS((J9Class *) READ_UDATA(interfaceObjectAndISlot));
-		UDATA methodIndex = READ_UDATA(interfaceObjectAndISlot + 1);
-		J9ROMMethod * romMethod = J9ROMCLASS_ROMMETHODS(interfaceClass->romClass);
-
-		while (methodIndex) {
-			romMethod = nextROMMethod(romMethod);
-			--methodIndex;
+		UDATA *interfaceObjectAndISlot = (UDATA *) JIT_RESOLVE_PARM(2);
+		J9Class *resolvedClass = READ_CLASS((J9Class *) READ_UDATA(interfaceObjectAndISlot));
+		UDATA iTableOffset = READ_UDATA(interfaceObjectAndISlot + 1);
+		J9Method *ramMethod = NULL;
+		J9ROMMethod *romMethod = NULL;
+		if (J9_ARE_ANY_BITS_SET(iTableOffset, J9_ITABLE_OFFSET_DIRECT)) {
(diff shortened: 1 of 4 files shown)


pshipton:
@gacholio does this fix any of the open issues? Is there any testing that covers this?

gacholio:
No and No - the code is only executed in error cases.

runtime/codert_vm/jswalk.c
```diff
@@ -1098,16 +1098,36 @@ static void jitWalkResolveMethodFrame(J9StackWalkState *walkState)
...
+			for(;;) {
+				J9Class *interfaceClass = allInterfaces->interfaceClass;
+				UDATA methodCount = interfaceClass->romClass->romMethodCount;
+				if (methodIndex < methodCount) {
+					/* iTable segment located */
+					ramMethod = interfaceClass->ramMethods + methodIndex;
+					break;
+				}
+				methodIndex -= methodCount;
+				allInterfaces = allInterfaces->next;
```
DanHeidinga: Should this assert that allInterfaces is not null?  It might be clearer to have the assert than a segv :)

runtime/codert_vm/jswalk.c
```diff
@@ -1098,16 +1098,36 @@ static void jitWalkResolveMethodFrame(J9StackWalkState *walkState)
...
+		J9Method *ramMethod = NULL;
+		J9ROMMethod *romMethod = NULL;
+		if (J9_ARE_ANY_BITS_SET(iTableOffset, J9_ITABLE_OFFSET_DIRECT)) {
+			ramMethod = (J9Method*)(iTableOffset & ~J9_ITABLE_OFFSET_TAG_BITS);
+		} else if (J9_ARE_ANY_BITS_SET(iTableOffset, J9_ITABLE_OFFSET_VIRTUAL)) {
+			UDATA vTableOffset = iTableOffset & ~J9_ITABLE_OFFSET_TAG_BITS;
+			J9Class *jlObject = J9VMJAVALANGOBJECT_OR_NULL(walkState->walkThread->javaVM);
+			ramMethod = *(J9Method**)((UDATA)jlObject + vTableOffset);
+		} else {
+			UDATA methodIndex = (iTableOffset -  sizeof(J9ITable)) / sizeof(UDATA);
```
DanHeidinga: There's some wonky spacing here but it's only worth fixing if something else needs to be changed in this PR

runtime/codert_vm/jswalk.c
```diff
@@ -1098,16 +1098,36 @@ static void jitWalkResolveMethodFrame(J9StackWalkState *walkState)
...
+			/* Find the appropriate segment for the referenced method within the
+			 * resolvedClass iTable.
```
DanHeidinga: Can you expand this comment to remind the reader that the itable includes not just the methods of the resolvedClass, but all the methods it inherits as well?  otherwise, this walk looks a little strange

runtime/codert_vm/jswalk.c
gacholio: The same could be said of every other pointer fetch in the VM (in particular, the many other places doing the same interface walk)

runtime/codert_vm/jswalk.c
gacholio: I can't see any spacing issues - what specifically are you complaining about?

gacholio: Ah, I see it.

runtime/codert_vm/jswalk.c
DanHeidinga: I can live with that

DanHeidinga:
Jenkins test sanity zlinux,xlinux jdk8

Issue #2978, opened 2018-09-21
security provider testing gets an unexpected AccessControlException


pshipton:
make jtreg_tests TESTDIRS=jdk/sun/security/provider

FAILED: sun/security/provider/PolicyFile/Modules.java
```
java.lang.ExceptionInInitializerError
	at java.base/java.lang.J9VMInternals.ensureError(J9VMInternals.java:191)
	at java.base/java.lang.J9VMInternals.recordInitializationFailure(J9VMInternals.java:180)
	at java.base/jdk.internal.module.SystemModuleFinders$SystemModuleReader.findImageLocation(SystemModuleFinders.java:426)
	at java.base/jdk.internal.module.SystemModuleFinders$SystemModuleReader.find(SystemModuleFinders.java:437)
	at java.base/jdk.internal.loader.BuiltinClassLoader$3.run(BuiltinClassLoader.java:617)
	at java.base/jdk.internal.loader.BuiltinClassLoader$3.run(BuiltinClassLoader.java:614)
	at java.base/java.security.AccessController.doPrivileged(AccessController.java:696)
	at java.base/jdk.internal.loader.BuiltinClassLoader.findResource(BuiltinClassLoader.java:614)
	at java.base/jdk.internal.loader.BuiltinClassLoader.findResource(BuiltinClassLoader.java:394)
	at java.base/jdk.internal.loader.BuiltinClassLoader.findResourceAsStream(BuiltinClassLoader.java:414)
	at java.base/jdk.internal.loader.BootLoader.findResourceAsStream(BootLoader.java:146)
	at java.base/java.lang.Module.getResourceAsStream(Module.java:1570)
	at java.base/com.ibm.oti.vm.MsgHelp.loadMessages(MsgHelp.java:189)
	at java.base/com.ibm.oti.util.PriviAction.run(PriviAction.java:172)
	at java.base/java.security.AccessController.doPrivileged(AccessController.java:640)
	at java.base/com.ibm.oti.util.Msg.<clinit>(Msg.java:63)
	at java.base/java.security.AccessController.throwACE(AccessController.java:176)
	at java.base/java.security.AccessController.checkPermissionHelper(AccessController.java:237)
	at java.base/java.security.AccessController.checkPermission(AccessController.java:373)
	at java.base/java.lang.SecurityManager.checkPermission(SecurityManager.java:322)
	at java.base/java.security.ProtectionDomain.seeAllp(ProtectionDomain.java:471)
[...]

Issue #597, opened 2017-11-15
cmdLineTester_SCHelperCompatabilityTests_unix_1 failure on openj9 openjdk-jdk8


vsebe:
```
Running test cmdLineTester_SCHelperCompatabilityTests_unix_1 ...
===============================================
test with Mode610
{ mkdir -p "/tmp/bld_370249/test/TestConfig/test_output_1510768124/cmdLineTester_SCHelperCompatabilityTests_unix_1"; \
	cd "/tmp/bld_370249/test/TestConfig/test_output_1510768124/cmdLineTester_SCHelperCompatabilityTests_unix_1"; \
	cp "/tmp/bld_370249/test/TestConfig/../../jvmtest/cmdLineTests/shareClassTests/SCHelperCompatabilityTests/HelperCompatabilityTests.jar" .; \
	"/tmp/bld_370249/j2sdk-image/jre/bin/../../bin/jar" xf HelperCompatabilityTests.jar; \
	 \
	"/tmp/bld_370249/j2sdk-image/jre/bin/java"  -Xcompressedrefs -Xcompressedrefs -Xjit -Xgcpolicy:gencon  -DJAVA_EXE='"/tmp/bld_370249/j2sdk-image/jre/bin/java"  -Xcompressedrefs -Xcompressedrefs -Xjit -Xgcpolicy:gencon ' -DJDK_HOME="/tmp/bld_370249/j2sdk-image/jre/bin/../.." -DCPDL=":" -DPROPS_DIR="/tmp/bld_370249/test/TestConfig/../../jvmtest/cmdLineTests/shareClassTests/SCHelperCompatabilityTests/props_unix" -DSCMODE=204 \
	-jar "/tmp/bld_370249/test/TestConfig/../../jvmtest/cmdline_options_tester/cmdlinetester.jar" \
	-config "/tmp/bld_370249/test/TestConfig/../../jvmtest/cmdLineTests/shareClassTests/SCHelperCompatabilityTests/HelperCompatabilityTests_SE80.xml" \
	-xids all,SE80 -xlist "/tmp/bld_370249/test/TestConfig/../../jvmtest/cmdLineTests/shareClassTests/SCHelperCompatabilityTests/exclude.xml" \
	-nonZeroExitWhenError \
	-outputLimit 300; \
[...]

PR #2997, opened 2018-09-24
Preload security sensitive classes

JasonFengJ9:
Preload security sensitive classes

Preload security sensitive classes when current security manager is `null`;
Not adding an extra static field to preload only once;
Minor refactoring to address Eclipse compiler warnings.

Along with https://github.com/ibmruntimes/openj9-openjdk-jdk11/pull/24
https://github.com/ibmruntimes/openj9-openjdk-jdk10/pull/60
and https://github.com/ibmruntimes/openj9-openjdk-jdk/pull/37
fix: #2978 

Manually verified that the tests in question now pass with two PRs referred above.

Reviewer: @pshipton 
FYI: @DanHeidinga 

### DIFF
--- a/jcl/src/java.base/share/classes/java/lang/System.java
+++ b/jcl/src/java.base/share/classes/java/lang/System.java
@@ -743,29 +743,42 @@ public static void setSecurityManager(final SecurityManager s) {
 	final SecurityManager currentSecurity = security;
 	
 	if (s != null) {
-		try {
-			/*[PR 95057] preload classes required for checkPackageAccess() */
-			// Preload classes used for checkPackageAccess(),
-			// otherwise we could go recursive 
-			s.checkPackageAccess("java.lang"); //$NON-NLS-1$
-		} catch (Exception e) {}
+		if (currentSecurity == null) {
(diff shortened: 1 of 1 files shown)


pshipton:
Does this problem also affect Java 10?

pshipton:
Please verify which issue this fixes. It was originally #597 and I changed it to #2978

jcl/src/java.base/share/classes/java/lang/System.java
```diff
@@ -753,16 +753,20 @@ public static void setSecurityManager(final SecurityManager s) {
 			/*[PR 97686] Preload the policy permission */
 			AccessController.doPrivileged(new PrivilegedAction() {
 				public Object run() {
+					ProtectionDomain newDomain = s.getClass().getPDImpl();
+					/*[IF !Java11]*/
 					ProtectionDomain oldDomain = currentSecurity == null ?
```
pshipton: The test should remain for when a security manager is already installed, but being changed. However the test can be modified to pre-initialize when changing the security manager from null to something else.

pshipton:

PR #2998, opened 2018-09-24
Use new OMR variables for installation locations

youngar:


youngar:
This PR needs to be accepted before the OMR side is accepted: https://github.com/eclipse/omr/pull/2996

DanHeidinga:
@youngar Can you provide more details on what this is and why it's needed?  Does it only effect the cmake builds or do the UMA builds need to be updated as well?

I'd also like to see a clear consensus in the OMR PR that this is the right direction before this is merged.  As it stands now, there appears to be a lot of open questions in the OMR issue.

youngar:
@DanHeidinga I updated the commit message with more details, copied below:
```
    Use new OMR variables for installation locations
    
    OMR is removing CMake install logic that is very specific to the OpenJ9
    build system, in favor of a customizable configuration.  To accomplish
    this, OMR is introducing new variables to customize the installation
    directories of OMR artifacts.
    
    OpenJ9 uses the install target in OMR to stage certain artifacts, such
    as omrsig and tracegen, into the build tree. The end result is
    functionally the same from the OpenJ9 perspective.
    
    Signed-off-by: Andrew Young <youngar17@gmail.com>
```

DanHeidinga:
Thanks for the clarification @youngar I'm still trying to understand the usecase that lead to these changes and have raised my questions in the OMR issue as I think they apply more broadly.

youngar:
@DanHeidinga This change is ready to go

DanHeidinga:
@dnakamura As the current cmake expert, can you review this and see how it fits in the cmake build?

pshipton:
eclipse/omr#2996 was merged and omr acceptance is now blocked by a cmake build failure.

pshipton:
@DanHeidinga @charliegracie fyi
https://github.com/eclipse/openj9/pull/2998#issuecomment-429694265
https://github.com/eclipse/openj9/pull/2998#issuecomment-424114092

Issue #2995, opened 2018-09-24
test.getDependency jobs failed to download asm-7.0-beta-20180911.182645-11.jar


llxia:
https://ci.eclipse.org/openj9/job/test.getDependency/ failed to download asm-7.0-beta-20180911.182645-11.jar as it does not exist any more.

```
12:28:10 ERROR: downloading https://repository.ow2.org/nexus/content/repositories/snapshots/org/ow2/asm/asm/7.0-beta-SNAPSHOT/asm-7.0-beta-20180911.182645-11.jar failed, return code: 2048
```

This blocks personal build as personal build downloads dependencies during each build. Regular test builds uses pre-stages test.getDependency Jenkins job, so they are not affected. Once we switch personal build to use same script as regular build, we should not have the issue of unstable dependency links in personal build.

@r30shah I can update the url to the latest one, but it seems they donot keep specific build for long. Is there a place we can download a stable build?

r30shah:
@llxia  I checked out snapshot page for ASM, I do not see stable snapshots. Only release page contains stable asm. I am looking for alternative but in snapshot they do have [maven-metadata.xml](https://repository.ow2.org/nexus/content/repositories/snapshots/org/ow2/asm/asm/7.0-beta-SNAPSHOT/maven-metadata.xml) which can be used to grab latest snapshot. (This will need mvn - So trying to find a way in which we can avoid it). For now yes, please update the url to the latest one.

llxia:
PR is merged. Close this issue.

sophia-guo:
It's broken again as the link only kept for around 10 days. Using pre-staged jars can unblock personal and production build. However docker story will fail. We should re-open this issue and see if there is a way to grab a  stable one to @r30shah .

smlambert:
Link was updated as part of #3089, earlier this evening.  My understanding was there is no stable one yet available from open/public location for the version we require.  I would be happy to hear otherwise.

sophia-guo:
In that case, the link will need to update every 10 days manually otherwise docker test story will fail.

smlambert:
yes, that is the unfortunate assessment Lan arrived at also in her comment above: https://github.com/eclipse/openj9/issues/2995#issuecomment-424129062

r30shah:
This should be fixed by this https://github.com/eclipse/openj9/pull/3095. They have released 7.0 Beta.

PR #2999, opened 2018-09-24
update asm-7.0 to latest version

llxia:
- asm-7.0-beta-20180911.182645-11.jar does not exist anymore
- this is temporary fix. From history, this build will not keep for long
- this blocks personal build at this point

Issue: #2995
[ci skip]


llxia:
Tested in grinder. This change cannot be verified by PR builds as PR builds uses pre-stages jars from test.getDependency Jenkins job.

PR #3004, opened 2018-09-24
Fix incorrect global and extern symbol references on OSX

nbhuiyan:
OSX requires all symbols that are either defined or used in
modules other than itself to have leading underscores
(only relevant when writing assembly code).
This fixes linker errors caused due to the missing underscores
in symbol declarations and references in xnathelp.o

Issue: #36

### DIFF
--- a/runtime/oti/xhelpers.m4
+++ b/runtime/oti/xhelpers.m4
@@ -53,7 +53,7 @@ define({FILE_END},{
 define({START_PROC},{
 	align 16
 	DECLARE_PUBLIC($1)
-	$1 proc
+	GLOBAL_SYMBOL($1) proc
 })
 
 define({END_PROC},{$1 endp})
@@ -66,9 +66,13 @@ define({LABEL},$1)
 
 define({C_FUNCTION_SYMBOL},$1)
 
+define({GLOBAL_SYMBOL},$1)
+
 },{	dnl WIN32
 
 ifdef({OSX},{ 
+
+dnl OSX
 
 define({SHORT_JMP},{})
 
@@ -79,13 +83,11 @@ define({FILE_START},{
 
 define({C_FUNCTION_SYMBOL},_$1)
 
-define({START_PROC},{
-	.align 16
-	DECLARE_PUBLIC($1)
-	_$1:
-})
+define({GLOBAL_SYMBOL},_$1)
 
 },{	dnl OSX
+
+dnl LINUX
 
 define({SHORT_JMP},{short})
 
@@ -97,13 +99,17 @@ define({FILE_START},{
 
 define({C_FUNCTION_SYMBOL},$1)
 
+define({GLOBAL_SYMBOL},$1)
+
+})	dnl OSX
+
+dnl LINUX and OSX
+
 define({START_PROC},{
 	.align 16
 	DECLARE_PUBLIC($1)
-	$1:
+	GLOBAL_SYMBOL($1):
 })
-
-})	dnl OSX
 
 define({FILE_END})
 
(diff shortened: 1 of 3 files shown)


DanHeidinga:
jenkins test sanity xlinux,win jdk8

DanHeidinga:
Change looks reasonable to me.  @gacholio any comments?

gacholio:
Not a huge fan of the new macro name, as the symbols are expressly not C. How about GLOBAL_SYMBOL or something like that. Otherwise, no concerns.

nbhuiyan:
I agree, it could create confusion down the road. I will rename it to GLOBAL_SYMBOL.

nbhuiyan:
@DanHeidinga @gacholio fixed

runtime/oti/xhelpers.m4
```diff
@@ -79,10 +81,12 @@ define({FILE_START},{
 
 define({C_FUNCTION_SYMBOL},_$1)
 
+define({GLOBAL_SYMBOL},_$1)
+
 define({START_PROC},{
```
DanHeidinga: The two copies of START_PROC (osx and linux) are now identical and can be moved out of the ifdef

nbhuiyan:
@DanHeidinga Unified OSX and Linux definition of `START_PROC`. Also fixed an unusual usage of `CALL_C_WITH_VMTHREAD` in xcinterp.m4 that does not work on Mac.

runtime/vm/xcinterp.m4
```diff
@@ -165,7 +165,11 @@ START_PROC(c_cInterpreter)
 	mov uword ptr J9TR_ELS_jitFPRegisterStorageBase[_rax],_rbx
 C_FUNCTION_SYMBOL(cInterpreter):
 	mov _rax,uword ptr J9TR_VMThread_javaVM[_rbp]
+ifdef({OSX},{
```
gacholio: What's the problem here? I'd rather not see platform ifdefs in the common code. Can the macro be fixed?

nbhuiyan: The problem is that on OSX, `CALL_C_WITH_VMTHREAD(uword ptr J9TR_JavaVM_bytecodeLoop[_rax],0)` gets processed as `call _qword ptr 192[rax]`. The underscore prepended to `qword` causes an error. This is a result of `CALL_C_WITH_VMTHREAD` making use of the `CALL_C_FUNC` macro, which then makes use of the `C_FUNCTION_SYMBOL` macro. On OSX:  `define({C_FUNCTION_SYMBOL},_$1)`. It is meant to handle symbol names, as opposed to addresses. It will work on Windows and Linux since they do not need to modify the symbol names, but it will not work on Mac. I will get rid of the platform ifdef in common code problem by introducing a different macro for making such calls, and update this branch.

nbhuiyan:
@DanHeidinga I have addressed the last issue raised by @gacholio . Ready for another review.

@0xdaryl fyi

runtime/oti/xhelpers.m4
```diff
@@ -534,6 +540,24 @@ define({STORE_VIRTUAL_REGISTERS},{
 
 })	dnl ASM_J9VM_ENV_DATA64
 
+ifdef({OSX},{
+	
+	define({FASTCALL_SYMBOL},{_$1})
```
gacholio: The rest of the file does not indent things inside ifdefs.

gacholio:
At some point, I'll likely do a sanitization pass over all the platforms to make the macros more consistent between architectures.

runtime/oti/xhelpers.m4
nbhuiyan: removed the indents.

DanHeidinga:
@gacholio Any further comments?  This looks reasonable to me.

DanHeidinga:

gacholio:
Looks reasonable to me.

PR #3005, opened 2018-09-24
Handle 8-byte call in VPIC snippet for nestmates

### LINKED ISSUE #2955: Test-sanity.functional-JDK11-win_x86-64_cmprssptrs StringPeepholeTest_0 NPE
https://ci.eclipse.org/openj9/job/Test-sanity.functional-JDK11-win_x86-64_cmprssptrs/19/tapResults/

This isn't the first occurrence of the failure.

```
Running test StringPeepholeTest_0 ...
StringPeepholeTest_0 Start Time: Thu Sep 20 01:05:16 2018 Epoch Time (ms): 1537423516714
test with -XX:-EnableHCR -Xjit:count=0,optLevel=hot
Exception in thread "main" java/lang/NullPointerException
	at java/lang/ThreadLocal.set (java.base@9/ThreadLocal.java:222)
	at java/lang/StringCoding.set (java.base@9/StringCoding.java:83)
	at java/lang/StringCoding.encode (java.base@9/StringCoding.java:500)
	at java/lang/System.afterClinitInitialization (java.base@9/System.java:157)
	at java/lang/Thread.initialize (java.base@9/Thread.java:371)
	at java/lang/Thread.<init> (java.base@9/Thread.java:153)

StringPeepholeTest_0_FAILED
```

jdmpapin:
When x86 PicBuilder is carrying out a direct call to a private method of
a nestmate, it sets the return address so that the call will return into
the code in the VPIC snippet, which in the case of a virtual dispatch
would be something like the following:

    call qword ptr [$vftReg+$vftOffset]
    jmp mainline

In particular, the callee should return to the jmp instruction, whose
address has been calculated as 7 + the address of the call instruction.
But the call is sometimes one byte longer. The variability is now dealt
with as follows:

- When overwriting the call, PicBuilder now writes a value that works
  for both 7- and 8-byte offsets to the jmp instruction.

- When doing direct dispatch for nestmates, PicBuilder now detects the
  offset to jmp, in order to compute the correct return address.

Fixes #2955

### DIFF
--- a/runtime/compiler/x/runtime/X86PicBuilder.pasm
+++ b/runtime/compiler/x/runtime/X86PicBuilder.pasm
@@ -1986,14 +1986,22 @@ resolvedToDirectMethodVPic:
       ; via a call though the VFT *unless* RA-5 has 0e8h (direct call relative).
       ; Make sure the VM sees an 0e8h so that it uses the VFT offset register.
       ; Start with a debug trap because the call itself is dead.
-      mov         dword ptr [rdx+eq_VPicData_size], 0e8cccch   ; int 3, int 3, call
+      ;
+      ; By writing 0e8h twice in a row, it doesn't matter here whether the vtable
+      ; call instruction would have been 7 or 8 bytes.
+      mov         dword ptr [rdx+eq_VPicData_size], 0e8e8cccch   ; int 3, int 3, call, call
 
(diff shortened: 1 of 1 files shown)


andrewcraik:

jdmpapin:
@andrewcraik, is this good now?

PR #3016, opened 2018-09-25
Fix package checking for array types

pdbain-ibm:
Compare the root component classes when checking isSamePackage().


DanHeidinga:

PR #3020, opened 2018-09-25
Fix extern "C" naming issues on OSX

nbhuiyan:
This is the other side of #3004, where references from c/cpp to assembly symbols should not have a leading underscore, since Clang adds an underscore to them anyway. Currently, this fix is implemented just for macOS, but in the future, there will be a cleaner, more unified solution once NASM is used across all x86 operating systems.

Signed-off-by: Nazim Uddin Bhuiyan <nazim.uddin.bhuiyan@ibm.com>

Issue: #36

### DIFF
--- a/runtime/compiler/control/CompilationThread.cpp
+++ b/runtime/compiler/control/CompilationThread.cpp
@@ -22,7 +22,7 @@
 
 #define J9_EXTERNAL_TO_VM
 
-#if SOLARIS || AIXPPC || LINUX
+#if SOLARIS || AIXPPC || LINUX || OSX
 #include <strings.h>
 #define J9OS_STRNCMP strncasecmp
 #else
@@ -4707,7 +4707,11 @@ bool TR::CompilationInfo::isQueuedForCompilation(J9Method * method, void *oldSta
    return linkageInfo->isBeingCompiled();
    }
 
+#if defined(OSX)
+JIT_HELPER(initialInvokeExactThunkGlue);
+#else
 JIT_HELPER(_initialInvokeExactThunkGlue);
+#endif
 
 void *TR::CompilationInfo::startPCIfAlreadyCompiled(J9VMThread * vmThread, TR::IlGeneratorMethodDetails & details, void *oldStartPC)
    {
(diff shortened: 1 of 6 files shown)


0xdaryl:

runtime/compiler/control/CompilationThread.cpp
```diff
@@ -4707,7 +4707,11 @@ bool TR::CompilationInfo::isQueuedForCompilation(J9Method * method, void *oldSta
    return linkageInfo->isBeingCompiled();
    }
 
+#if defined(OSX)
+JIT_HELPER(initialInvokeExactThunkGlue);
```
DanHeidinga: Can the JIT_HELPER macro be updated to include `_` on OSX?  That would minimize the other required changes.

nbhuiyan: The leading `_` is needed for platforms other than OSX. So what we would need is to have `JIT_HELPER(x)` defined to `extern "C" void x()` on mac and `extern "C" void _x()` on other platforms. We would then also need to enclose any calls to these functions in C/C++ in another macro, say, `CALL_JIT_HELPER`.

However, I did not go with that solution because the problem is that there is a lack of consistency here. The method calls/declarations with leading underscore in C also are named with leading underscores in the MASM code. Most of the methods are missing a leading underscore (and also in the assembly code), and so using our single `JIT_HELPER` macro that expands differently on OSX would not work. We would then need to divide them into 2 different categories, requiring 4 different macros: `JIT_HELPER(x)`, `JIT_HELPER_UNDERSCORE(x)`, `CALL_JIT_HELPER(x)`, `CALL_JIT_HELPER_UNDERSCORE(x)`.

Did you have a different idea in mind?

DanHeidinga: The VM handles this somewhat differently by hiding the platform differences in the macros.  Looking at the JIT code, the approach there is to handle it with ifdefs throughout.

I don't want this to block this PR from moving forward.  Though I think it would be worth looking at cleaning this up in the future.

nbhuiyan: This will be cleaned up once the NASM assembly files are set up to be used by Windows and xLinux. We will no longer need to have the code in the `#else` blocks of `#if defined(OSX)` in this PR.

0xdaryl:
I agree with @DanHeidinga that hiding the macOS specifics behind the `SET` and `JIT_HELPERS` would be a cleaner solution.  But you've already agreed to do that by end of October so I don't think it's needed in this PR.

Issue #3001, opened 2018-09-24
Update InvokeDynamic Error wrapping


fengxue-IS:
JVMS JDK11 Section 5.4.3.6 'Dynamically-Computed Constant and Call Site Resolution' have updated the requirement of resolution exception/error should be handled.

`resolveInvokeDynamic(...)` method need to be updated to match the specification

fengxue-IS:
> Invokedynamic resolution needs the same change in JDK11.
> This is a change in behaviour for indy pre-jdk11.
@DanHeidinga does this change also apply to JDK8 or is a runtime check need?

DanHeidinga:
I think this can be handled with the java preprocessor so no runtime check is required, but yes to JPP ifdefs.

PR #3022, opened 2018-09-25
Update InvokeDynamic Error wrapping

fengxue-IS:

Close: #3001 


fengxue-IS:

pshipton:
@fengxue-IS please create a PR against the openj9-0.10.0-rc1 branch and we will include this fix in the 0.10.0 release candidate 2.

Issue #2948, opened 2018-09-19
Nestmates visibility exception handling


fengxue-IS:
https://github.com/eclipse/openj9/blob/685530b8f8fe667e0651ed60c750b6c52921f827/runtime/vm/lookupmethod.c#L204-L209

`checkVisibility` could have set non `illegalAccessError` during validation of class nesthost, this exception might be overwritten by the caller of `processMethod` and result in a different exception type being thrown. 

We should reconsider how to handle the exception propagation of `checkVisibility(...)` call.

gacholio:
This is a much larger issue - prior to nestmates, checkVisibility was not an HCR point, now it is. Also, it appears the flags being passed/used in this code are wildly incompatible (call assumes J9_LOOK_ flags, but is getting J9_RESOLVE_ flags in many cases - there is no compatibility between the two).

gacholio:
I've found at least one other place not doing it. I'm working on a large fix.

PR #3027, opened 2018-09-25
Fix visibility checking

gacholio:
- handle exceptions and pop frames correctly after visibility checks
- update new instance prototype access check
- fix confusion between resolve and lookup flags
- add "no java" lookup flag to ensure no java code runs

Fixes: #2948

[ci skip]


runtime/codert_vm/cnathelp.cpp
```diff
@@ -2448,50 +2448,59 @@ old_slow_jitNewInstanceImplAccessCheck(J9VMThread *currentThread)
...
+	/* See if the method is visible - don't use checkVisibility here because newInstance
+	 * has different rules for protected method visibility.
+	 */
+	if (!J9_ARE_ANY_BITS_SET(romMethod->modifiers, J9AccPublic)) {
+		if (J9_ARE_ANY_BITS_SET(romMethod->modifiers, J9AccPrivate)) {
+			/* Private methods are visible only within their declaring class
+			 * unless the two classes are in the same nest (JDK11 and beyond).
+			 */
+#if defined(J9VM_OPT_VALHALLA_NESTMATES)
+			if (thisClass->nestHost != callerClass->nestHost)
```
DanHeidinga: This is insufficient for the nest mates case as the `nestHost` may not have been loaded and validated yet.  It likely needs to call `loadAndVerifyNestHost()` if the nestHosts are null.

runtime/vm/resolvesupport.cpp
```diff
@@ -1876,6 +1895,7 @@ resolveMethodHandleRefInto(J9VMThread *vmThread, J9ConstantPool *ramCP, UDATA cp
 	cpItem = (J9ROMMethodRef *) &(J9_ROM_CP_FROM_CP(ramCP)[fieldOrMethodIndex]);
 	definingClass = ((J9RAMClassRef *) &(ramCP[cpItem->classRefCPIndex]))->value;
 	if (definingClass == NULL) {
+		/* TODO: no exception set here */
```
DanHeidinga: TODO?

runtime/codert_vm/cnathelp.cpp
gacholio: At the start of the function, checkVisibility is called with thisClass and callerClass, which should ensure the loading, shouldn't it?

runtime/vm/resolvesupport.cpp
gacholio: Throwing NPE now, if requested.

runtime/codert_vm/cnathelp.cpp
DanHeidinga: Classes are either `public` or `package` access at the VM level.  Never `private` which is the only case that can cause nest validation to occur.

DanHeidinga:
Once the nestmates check is ironed out, the rest of this looks fine

runtime/codert_vm/cnathelp.cpp
gacholio: Should be fixed now.

DanHeidinga:

DanHeidinga:

pshipton:
@gacholio Note the compile failure.

gacholio:

DanHeidinga:

gacholio:

gacholio:

PR #3034, opened 2018-09-26
Fix RPATH and dylib install_name on OSX

babsingh:
On OSX, -rpath,\\$$ORIGIN doesn't work. -rpath needs to correspond to
either @executable_path, @loader_path or some other path. More
information can be seen in the man pages of dyld. I have chosen to set
rpath as "-rpath,@loader_path/.". @loader_path points to the directory
where dylib being loaded exists. This works since libjvm is able to load
all the dependent dylibs.

When naming dylibs, @rpath/ should be appended to dylib names. Example:
@rpath/libj9thr29.dylib. Appending @rpath to the dylib name instructs
the dynamic linker to search the list of run paths for the dylib.

Reference:
https://wincent.com/wiki/%40executable_path%2C_%40load_path_and_%40rpath

### DIFF
--- a/runtime/makelib/targets.mk.osx.inc.ftl
+++ b/runtime/makelib/targets.mk.osx.inc.ftl
@@ -47,7 +47,7 @@ $(UMA_EXETARGET): $(UMA_OBJECTS) $(UMA_TARGET_LIBRARIES)
 UMA_BEGIN_DASH_L =
 UMA_END_DASH_L =
 
-UMA_EXE_POSTFIX_FLAGS += -lm -liconv -lc -ldl -lutil -Wl,-rpath,\$$ORIGIN
+UMA_EXE_POSTFIX_FLAGS += -lm -liconv -lc -ldl -lutil -Wl,-rpath,@loader_path
 
 <#if uma.spec.processor.amd64>
   UMA_MASM2GAS_FLAGS += --64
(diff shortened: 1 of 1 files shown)


babsingh:
fyi - @DanHeidinga

babsingh:
Related to https://github.com/eclipse/openj9/issues/36

runtime/makelib/targets.mk.osx.inc.ftl
```diff
@@ -47,7 +47,7 @@ $(UMA_EXETARGET): $(UMA_OBJECTS) $(UMA_TARGET_LIBRARIES)
 UMA_BEGIN_DASH_L =
 UMA_END_DASH_L =
 
-UMA_EXE_POSTFIX_FLAGS += -lm -liconv -lc -ldl -lutil -Wl,-rpath,\$$ORIGIN
+UMA_EXE_POSTFIX_FLAGS += -lm -liconv -lc -ldl -lutil -Wl,-rpath,@loader_path/.
```
DanHeidinga: Why `/.` on the end of the path?

babsingh: i think it should work without `/.`. I will verify and remove it.

babsingh: verified; works without `/.`; removed it.

DanHeidinga:
File is osx only which isn't in the builds yet.  Merging as is

PR #3035, opened 2018-09-26
Properly set bytecode info when save node to temp

0dvictor:
Ensure the bytecode info of the load from temp node is same as the
original node.


andrewcraik:
Just want a sanity to verify the change

Issue #2990, opened 2018-09-24
JTreg security test fails due to unexpected 401 return code from test server


ben-walsh:
Test test8077155 of test/jdk/sun/security/krb5/auto/HttpNegotiateServer.java fails with the following exception stack when run against OpenJ9. No such error occurs when run against Hotspot ...

```
Expecting to succeed when executing with the the logged in subject.
GSSException: No valid credentials provided (Mechanism level: No valid credentials provided (Mechanism level: Attempt to obtain new INITIATE credentials failed! (null)))
	at java.security.jgss/sun.security.jgss.spnego.SpNegoContext.initSecContext(SpNegoContext.java:453)
	at java.security.jgss/sun.security.jgss.GSSContextImpl.initSecContext(GSSContextImpl.java:266)
	at java.security.jgss/sun.security.jgss.GSSContextImpl.initSecContext(GSSContextImpl.java:196)
	at java.security.jgss/sun.net.www.protocol.http.spnego.NegotiatorImpl.init(NegotiatorImpl.java:108)
	at java.security.jgss/sun.net.www.protocol.http.spnego.NegotiatorImpl.<init>(NegotiatorImpl.java:117)
	at java.base/jdk.internal.reflect.NativeConstructorAccessorImpl.newInstance0(Native Method)
	at java.base/jdk.internal.reflect.NativeConstructorAccessorImpl.newInstance(NativeConstructorAccessorImpl.java:62)
	at java.base/jdk.internal.reflect.DelegatingConstructorAccessorImpl.newInstance(DelegatingConstructorAccessorImpl.java:45)
	at java.base/java.lang.reflect.Constructor.newInstance(Constructor.java:490)
	at java.base/sun.net.www.protocol.http.Negotiator.getNegotiator(Negotiator.java:65)
	at java.base/sun.net.www.protocol.http.NegotiateAuthentication.firstToken(NegotiateAuthentication.java:234)
	at java.base/sun.net.www.protocol.http.NegotiateAuthentication.setHeaders(NegotiateAuthentication.java:209)
	at java.base/sun.net.www.protocol.http.HttpURLConnection.getServerAuthentication(HttpURLConnection.java:2564)
	at java.base/sun.net.www.protocol.http.HttpURLConnection.getInputStream0(HttpURLConnection.java:1755)
	at java.base/sun.net.www.protocol.http.HttpURLConnection$9.run(HttpURLConnection.java:1501)
[...]

PR #3051, opened 2018-09-27
Set domainCombiner from closest frame in result AccessControlContext

JasonFengJ9:
Set `domainCombiner` from closest frame in result `AccessControlContext` of `j.s.AccessController.getContextHelper()`

For `JEP140` Format result of `j.s.AccessController.getAccSnapshot()`, returning `AccessControlContext.domainCombiner` is set to a `non-null DomainCombiner` of an `AccessControlContext` object from the closest frame objects or `null`.

Verified that the failed testcase passes and `pConfig` still compiles.

close: #2990 

Reviewer: @pshipton
FYI: @DanHeidinga

### DIFF
--- a/jcl/src/java.base/share/classes/java/security/AccessController.java
+++ b/jcl/src/java.base/share/classes/java/security/AccessController.java
@@ -469,6 +469,7 @@ private static AccessControlContext getContextHelper(boolean forDoPrivilegedWith
 	int frameNbr = domains.length / OBJS_ARRAY_SIZE;
 	AccessControlContext accContext = null;
 	AccessControlContext accLower = null;
+	DomainCombiner dc = null; // store a non-null domainCombiner from a closest frame or null
 	for (int j = 0; j < frameNbr; ++j) {
 		AccessControlContext acc = (AccessControlContext) domains[j * OBJS_ARRAY_SIZE];
 		/*[PR JAZZ 66930] j.s.AccessControlContext.checkPermission() invoke untrusted ProtectionDomain.implies */
@@ -486,6 +487,10 @@ private static AccessControlContext getContextHelper(boolean forDoPrivilegedWith
 		}
 		if (null != acc && null != acc.domainCombiner) {
 			accTmp.domainCombiner = acc.domainCombiner;
+			if (dc == null) {
+				// this dc will be set to accContext.domainCombiner
+				dc = acc.domainCombiner;
+			}
 		}
 		if (null != domains[j * OBJS_ARRAY_SIZE + OBJS_INDEX_PERMS_OR_CACHECHECKED]) {
 			// this is frame with limited permissions
@@ -500,7 +505,9 @@ private static AccessControlContext getContextHelper(boolean forDoPrivilegedWith
 		}
 		accLower = accTmp;
 	}
-
+	if ((accContext != null) && (accContext.domainCombiner == null) && (dc != null)) {
+		accContext.domainCombiner = dc;
+	}
 	return accContext;
 }
 


jcl/src/java.base/share/classes/java/security/AccessController.java
```diff
@@ -500,7 +504,9 @@ private static AccessControlContext getContextHelper(boolean forDoPrivilegedWith
 		}
 		accLower = accTmp;
 	}
-
+	if (accContext != null) {
+		accContext.domainCombiner = dc;
```
pshipton: Is accContext.domainCombiner expected to be null here?

jcl/src/java.base/share/classes/java/security/AccessController.java
```diff
@@ -486,6 +487,9 @@ private static AccessControlContext getContextHelper(boolean forDoPrivilegedWith
 		}
 		if (null != acc && null != acc.domainCombiner) {
 			accTmp.domainCombiner = acc.domainCombiner;
```
pshipton: Is this assignment still required, considering dc will always be used?

JasonFengJ9: It is possible that there are multiple `AccessControlContext` object discovered during stack walking along with multiple `non-null domainCombiner`. Returning `AccessControlContext.domainCombiner` will be set to the `domainCombiner` from `AccessControlContext` in the closed frame. For other cases, I feel the `domainCombiner` will be set as before.
One optimization could be as following:
```
			if (dc == null) {
				dc = acc.domainCombiner;
			} else {
				accTmp.domainCombiner = acc.domainCombiner;
			}
``` 
This saves a unnecessary assignment. 
PR is updated.

jcl/src/java.base/share/classes/java/security/AccessController.java
JasonFengJ9: `dc` could be `null`.
Did you mean `accContext` instead? It is not expected to be `null` but the code structure could let it stay `null` if `frameNbr` is zero. (Eclipse has a warning about this).

pshipton: I'm wondering if dc can overwrite the domainCombiner already contained in the accContext.domainCombiner

jcl/src/java.base/share/classes/java/security/AccessController.java
```diff
@@ -485,7 +486,12 @@ private static AccessControlContext getContextHelper(boolean forDoPrivilegedWith
 			accTmp = new AccessControlContext(pDomains, newAuthorizedState);
 		}
 		if (null != acc && null != acc.domainCombiner) {
-			accTmp.domainCombiner = acc.domainCombiner;
+			if (dc == null) {
+				// this dc will be set to accContext.domainCombiner
+				dc = acc.domainCombiner;
+			} else {
+				accTmp.domainCombiner = acc.domainCombiner;
```
pshipton: What I'm wondering is if the accTmp.domainCombiner, other than the one used to assign accContext, is ever used for anything? Just in terms of riskiness for this change, I liked the original code (without the `if (dc == null) ... else ...` better since this version could leave a chained nextStackAcc domainCombiner as null, when before this PR it had a value.

jcl/src/java.base/share/classes/java/security/AccessController.java
JasonFengJ9: `dc` is the first `non-null domainCombiner` discovered so it shouldn't overwrite a different `non-null domainCombiner`. To be safe, adding a `null` check for this assignment.

jcl/src/java.base/share/classes/java/security/AccessController.java
JasonFengJ9: When there are multiple `AccessControlContext` with `domainCombiner` objects discovered during stack walking, only one can be set to returning `AccessControlContext`. It doesn't appear those `domainCombiner` within `AccessControlContext.nextStackAcc` are used. 
For now I would leave it as is to prevent any behaviour changes.
Reverted to original version. Please have another look.

pshipton:
jenkins test all jdk8 zlinux

pshipton:
jenkins test sanity jdk11 plinux

pshipton:
jenkins test all zlinux jdk8

pshipton:

pshipton:
@JasonFengJ9 this PR closed #2990, but @llxia tells me the test for #2990 is still failing. Also it doesn't look like the correct Issue to have closed, was there a typo?

JasonFengJ9:
@pshipton it has been a while (more than two years ago). From the comments within #2990, this appears related, I think it wasn't a typo.

@LongyuZhang is narrowing down if this is a regression.

LongyuZhang:
@JasonFengJ9 Tested with old sdk 20181206 from vmfarm, for jdk_security4 (Internal_grinder 12217), it failed 126 out of 135 tests, for this specific HttpNegotiateServer test, it failed at compilation, same as the original issue (https://github.com/eclipse/openj9/issues/2990#issuecomment-425048410). In addition, I found jdk_security4 test was enabled only since Nov.13, 2018 (https://github.com/AdoptOpenJDK/openjdk-tests/pull/681), after this issue has already been closed (Sep.27, 2018). I will open another issue to add more details. Thanks.

PR #3053, opened 2018-09-27
Remove implicit NULL checks on java/lang/String.value

fjeremic:
The implicit NULL checks on java/lang/String.value are unnecessary
since at IL gen time we check for this particular field and we avoid
generating the NULL checks anyway.

### DIFF
--- a/jcl/src/java.base/share/classes/java/lang/String.java
+++ b/jcl/src/java.base/share/classes/java/lang/String.java
@@ -1347,9 +1347,6 @@ public final class String implements Serializable, Comparable<String>, CharSeque
 
 		byte[] s1Value = s1.value;
 		byte[] s2Value = s2.value;
-		
-		s1Value.getClass(); // Implicit null check
-		s2Value.getClass(); // Implicit null check
 
 		if (enableCompression && (null == compressionFlag || (s1.coder | s2.coder) == LATIN1)) {
 			while (o1 < end) {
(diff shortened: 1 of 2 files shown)


fjeremic:
Leaving this one as WIP until I've collected logs of before/after this change for all the methods modified and manually verified we don't generate the NULL check in both cases.

DanHeidinga:
@fjeremic Are any of those nullchecks in the java source required for correctness in the interpreter?

fjeremic:
> @fjeremic Are any of those nullchecks in the java source required for correctness in the interpreter?

No, because the interpreter will do a null check on the field dereference. The JIT doesn't have to because it "knows" this particular field in this safe JCL class is immutable as defined by the specification.

andrewcraik:
@fjeremic Have you verified the loops below the eliminated NULLCHKs still optimize properly? Any NULLCHKs that would have been in those loops would be eliminated in the new scheme so since this is done at ilgen this should be fine. This pattern was introduced in the early days of Java 8 from what I remember to make sure we didn't end up with NULLCHKs in performance critical loops. I think this new solution should be just better and achieve the same aim, but wanting to make sure that it has checked out in your logs.

fjeremic:
> @fjeremic Have you verified the loops below the eliminated NULLCHKs still optimize properly? 

It appears they are. We have the exact same tree sequence in both situations (before and after this change) at the post optimization trees. The only difference I'm seeing is that in the before case (before this change that is) we have the NULLCHKs from the `.getClass()` lines hanging around until VP gets rid of them. Following VP the trees look identical but in and outside of loops.

fjeremic:
Jenkins test sanity xlinux JDK8,JDK11

fjeremic:
@andrewcraik I think this one should be ready.

PR #3056, opened 2018-09-27
Fix -install_name on OSX

babsingh:
On OSX, <LIB_NAME_DYLIB>.dylib should match the -install_name
<LIB_NAME_INSTALL>. Otherwise, the linker is unable to find the required
library. Currently, LIB_NAME_DYLIB doesn't match LIB_NAME_INSTALL.
LIB_NAME_DYLIB has major and minor build versions appended whereas
LIB_NAME_INSTALL doesn't contain version info. UMA_LIB_NAME will contain
LIB_NAME_INSTALL with major and minor build versions.

UMA_LIB_NAME will be used to specify -install_name. With this change,
the linker is able to find the required libraries on OSX, and OpenJ9 can
be built successfully with the JIT disabled. Other platforms shouldn't
be impacted since UMA_LIB_NAME is only used on OSX.

[ci skip] 

### DIFF
--- a/runtime/makelib/targets.mk.osx.inc.ftl
+++ b/runtime/makelib/targets.mk.osx.inc.ftl
@@ -118,7 +118,7 @@ endif
 </#if>
 
 # https://stackoverflow.com/questions/21907504/how-to-compile-shared-lib-with-clang-on-osx
-UMA_DLL_LINK_FLAGS += -shared -undefined dynamic_lookup -install_name @rpath/lib$(UMA_TARGET_NAME).dylib
+UMA_DLL_LINK_FLAGS += -shared -undefined dynamic_lookup -install_name @rpath/lib$(UMA_LIB_NAME).dylib
 UMA_DLL_LINK_FLAGS += -Xlinker -rpath -Xlinker @loader_path
 
 UMA_DLL_LINK_POSTFLAGS += $(UMA_LINK_STATIC_LIBRARIES)
--- a/sourcetools/com.ibm.uma/com/ibm/uma/UMA.java
+++ b/sourcetools/com.ibm.uma/com/ibm/uma/UMA.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2001, 2017 IBM Corp. and others
+ * Copyright (c) 2001, 2018 IBM Corp. and others
  *
  * This program and the accompanying materials are made available under
  * the terms of the Eclipse Public License 2.0 which accompanies this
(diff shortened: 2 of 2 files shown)


babsingh:
fyi - @DanHeidinga. this is the last change. i was able to built OpenJ9 with no-JIT on OSX without any hacks.

DanHeidinga:
Issue #36

runtime/makelib/targets.mk.osx.inc.ftl
```diff
@@ -118,7 +118,7 @@ endif
 </#if>
 
 # https://stackoverflow.com/questions/21907504/how-to-compile-shared-lib-with-clang-on-osx
-UMA_DLL_LINK_FLAGS += -shared -undefined dynamic_lookup -install_name @rpath/lib$(UMA_TARGET_NAME).dylib
+UMA_DLL_LINK_FLAGS += -shared -undefined dynamic_lookup -install_name @rpath/lib$(UMA_LIB_NAME).dylib
```
DanHeidinga: Since this is an ftl file, can we do with this `$(UMA_TARGET_NAME)${uma.buildinfo.version.major}${uma.buildinfo.version.minor}`?  

I'm loath to add new defines to UMA as we're trying to replace it with cmake

babsingh: Adding `$major$minor` to every dylib causes this error: `JVMJ9VM013W Initialization error in function j9dyn29(7): failed to initialize JImage interface`.

babsingh: **com.ibm.j9.uma.platform.PlatformImplementation.getTargetNameWithRelease:**
```
	public String getTargetNameWithRelease(Artifact artifact) {
		if ( artifact.appendRelease() ) {
			return artifact.getTargetName() + getStream();
		}
		return artifact.getTargetName();
	}
```

**module.xml**: `<artifact name="jvm" ... appendrelease="false">`

If `appendrelease` is true, then `$major$minor` is appended.

DanHeidinga: Is `appendrelease="false"` enabled on the wrong libraries?  While the PR addresses the issue, it feels like it leaves the larger problem lurking.

babsingh: > Is appendrelease="false" enabled on the wrong libraries?

I don't think so. `appendrelease="false"` is set to false for libraries which don't change across Java versions such as `jvm`, `jsig`, `j9omrport` etc. Libraries such as `j9vm`, `j9gc`, `j9prt` ... may differ across Java versions. For such libraries, `appendrelease` is not specified. I think `appendrelease` is set to true by default if it's not specified.

DanHeidinga: Thanks @babsingh for digging through this with me deskside.  I now agree that this is a reasonable solution given the existing hacks in j9lib.h.in / j9lib.h.ftl.

FYI @dnakamura This may affect the cmake build when ported to osx as it adds a new UMA flag `UMA_LIB_NAME`

DanHeidinga:
Jenkins compile xlinux jdk8

Issue #3054, opened 2018-09-27
Concurrent Scavenge support on X86


0dvictor:
Concurrent Scavenge has proved a success on Z and hence X86 will follow. https://github.com/eclipse/omr/issues/2812 provides a good place naturally insert Concurrent Scavenge related read barriers. X86's implementation will take advantage of this new infrastructure and follow the overall principles from Z.

List of works:

- [x] GC to enable Concurrent Scavenge on X86 when `-Xgc:concurrentScavenge` is given; #3070
- [x] Optimizer to insert read barrier where necessary; #3074
- [x] CodeGen to evaluate read barrier correctly and to generate optimal instruction sequences; #3075
- [x] CodeGen to update reference arraycopy inlining sequence to call read barrier when necessary; #2963
- [x] GodeGen to update atomic compare and swap object inlining sequence to include a read barrier; #3111
- [x] Fix for J9VMJAVAUTILCONCURRENTLOCKSABSTRACTOWNABLESYNCHRONIZER; #3256
- [x] Turn on Concurrent Scavenge in the default Linux X86-64 CompressedRefs build spec; #3253

Documentation: https://github.com/eclipse/openj9-docs/issues/99

dsouzai:
If X86 has any TM related instructions, you'll likely have to deal with those too.

0dvictor:
@vijaysun-omr FYI.

andrewcraik:
created https://github.com/eclipse/openj9-docs/issues/99 as the docs issue.

0dvictor:
All pull requests listed in this Issue have been merger, closing.

PR #3070, opened 2018-09-28
Enable Concurrent Scavenge on X86-64 compressedrefs

0dvictor:
Concurrent Scavenge will be enabled if "-Xgc:concurrentScavenge" is set
on X86-64 compressedrefs builds.

Part of Issue #3054

### DIFF
--- a/runtime/gc_modron_startup/mminit.cpp
+++ b/runtime/gc_modron_startup/mminit.cpp
@@ -2853,6 +2853,9 @@ gcInitializeDefaults(J9JavaVM* vm)
 	if (gc_policy_gencon == extensions->configurationOptions._gcPolicy) {
 		/* after we parsed cmd line options, check if we can obey the request to run CS (valid for Gencon only) */
 		if (extensions->concurrentScavengerForced) {
+#if defined(J9VM_ARCH_X86)
+			extensions->softwareEvacuateReadBarrier = true;
+#endif /* J9VM_ARCH_X86 */
 			if (LOADED == (FIND_DLL_TABLE_ENTRY(J9_JIT_DLL_NAME)->loadFlags & LOADED)) {
 				/* If running jitted, it must be on supported h/w */
 				J9ProcessorDesc  processorDesc;
--- a/runtime/gc_modron_startup/mmparseXXgc.cpp
+++ b/runtime/gc_modron_startup/mmparseXXgc.cpp
@@ -787,10 +787,7 @@ gcParseXXgcArguments(J9JavaVM *vm, char *optArg)
 			continue;
 		}
 		if(try_scan(&scan_start, "softwareEvacuateReadBarrier")) {
-			/* Software read barriers are only implemented on s390 for now */
-#if defined(S390) || defined(J9ZOS390)
 			extensions->softwareEvacuateReadBarrier = true;
-#endif /* defined(S390) || defined(J9ZOS390) */
 			continue;
 		}
 #endif /* defined(OMR_GC_CONCURRENT_SCAVENGER) */


dmitripivkine:
@amicic FYI

runtime/gc_modron_startup/mminit.cpp
```diff
@@ -2853,6 +2853,9 @@ gcInitializeDefaults(J9JavaVM* vm)
 	if (gc_policy_gencon == extensions->configurationOptions._gcPolicy) {
 		/* after we parsed cmd line options, check if we can obey the request to run CS (valid for Gencon only) */
 		if (extensions->concurrentScavengerForced) {
+#if defined(J9VM_ARCH_X86) && defined(J9VM_ENV_DATA64) && defined(J9VM_GC_COMPRESSED_POINTERS)
+            extensions->concurrentScavenger = true;
```
DanHeidinga: Please use tabs here rather than spaces so the indentation of this is correct.

0dvictor: Good catch. Updated.

runtime/gc_modron_startup/mminit.cpp
```diff
@@ -2853,6 +2853,9 @@ gcInitializeDefaults(J9JavaVM* vm)
 	if (gc_policy_gencon == extensions->configurationOptions._gcPolicy) {
 		/* after we parsed cmd line options, check if we can obey the request to run CS (valid for Gencon only) */
 		if (extensions->concurrentScavengerForced) {
+#if defined(J9VM_ARCH_X86) && defined(J9VM_ENV_DATA64) && defined(J9VM_GC_COMPRESSED_POINTERS)
+			extensions->concurrentScavenger = true;
+#else /* J9VM_ARCH_X86 && J9VM_ENV_DATA64 && J9VM_GC_COMPRESSED_POINTERS */
```
amicic: it would be simpler code to rely on the branch down below that does:

 						|| (extensions->softwareEvacuateReadBarrier)) {
 					extensions->concurrentScavenger = true;

That would of course require for the two j9sysinfo calls run stable and return false.

runtime/gc_modron_startup/mmparseXXgc.cpp
```diff
@@ -787,10 +787,7 @@ gcParseXXgcArguments(J9JavaVM *vm, char *optArg)
 			continue;
 		}
 		if(try_scan(&scan_start, "softwareEvacuateReadBarrier")) {
-			/* Software read barriers are only implemented on s390 for now */
-#if defined(S390) || defined(J9ZOS390)
 			extensions->softwareEvacuateReadBarrier = true;
-#endif /* defined(S390) || defined(J9ZOS390) */
```
amicic: I'd rather be very specific and allow only for platforms we enable CS.

amicic:
not enabling OMR_GC_CONCURRENT_SCAVENGER build flag yet (or is it part of OMR changes)?

amicic:
Actually, before we enable the build flag, regression testing should be done, if gencon (without CS) has been effected.

The point is, read barrier in the interpreted will be activated even for non-CS runs. It will likely slow down the interpreted runs, but hopefully jitted runs will be unaffected. 

GC (Scavenge cycle) may slow down marginally as well.

runtime/gc_modron_startup/mmparseXXgc.cpp
0dvictor: This code has already be guarded by `OMR_GC_CONCURRENT_SCAVENGER`, which is only defined on platforms that we have enabled CS. IMO, another layer of "#ifdef" only introduces confusion but no actually benefits.

runtime/gc_modron_startup/mminit.cpp
0dvictor: On X86, no CPU feature check is necessary to enable CS. JIT is designed to support CS on all CPUs that OpenJ9 runs.

runtime/gc_modron_startup/mmparseXXgc.cpp
amicic: make sense., I did not check this was already under OMR_GC_CONCURRENT_SCAVENGER

runtime/gc_modron_startup/mminit.cpp
amicic: correct, the CPU check can be skipped (by putting into a separate  GC helper and guard this with something Z specific), but I suspect/hope the check is anyway a no-op (by similar guard on port lib layer).

bottom line I don't like adding another code branch that is X86 specific. CS would be triggered by softwareEvacuateReadBarrier on non h/w assisted platforms. and that branch already exists.

vijaysun-omr:
@amicic what would you suggest we test to ensure that gencon is'nt affected too badly ?

pshipton:
This looks like the appropriate PR. I've added a depends:doc tag. Please ensure an issues is created at https://github.com/eclipse/openj9-docs to handle the documentation requirements.

runtime/gc_modron_startup/mminit.cpp
0dvictor: Good point. I've updated this logic. X86 now sets `softwareEvacuateReadBarrier` as it is indeed a "software evacuate read barrier"; then we can rely on the branch down below.

0dvictor:
> This looks like the appropriate PR. I've added a depends:doc tag. Please ensure an issue is created at https://github.com/eclipse/openj9-docs to handle the documentation requirements.

Thank-you for the reminder. An Issue has been created https://github.com/eclipse/openj9-docs/issues/99 by @andrewcraik

andrewcraik:
@DanHeidinga and @amicic have your concerns been addressed now as this has been open for a little while and we should look to merge soon for 0.11.

DanHeidinga:
@0dvictor If this is ready to go, please remove the "WIP" before its merged.

0dvictor:
> If this is ready to go, please remove the "WIP" before its merged.

The change itself is ready; however, I'd like to let #3075 merged first otherwise runs with "-Xgc:concurrentScavenge" would crash due to lack of CodeGen support.

amicic:
While this looks ok now, I don't like having any platform specific code for s/w based Concurrent Scavenge/ Read Barrier.

Right now, on Z, if you try to run CS on non h/w supported platform we will default to standard Scavenger. We should be uniform and probably change that behaviour (@fjeremic @vijaysun-omr do, you agreee?), that is, both Z and X86 should default to s/w based CS. But since Z s/w based is not quite ready, that new behaviour would lead to crashes. 

So, again I'm ok with #ifdef for X, but I hope that ifdef should go away soon.

vijaysun-omr:
The only issue to consider is whether the s/w based CS scheme has the performance characteristics that make it an attractive choice on z when we do not have the hardware support. Assuming the performance characteristics are not terribly bad, I agree that we should be doing what you suggest @amicic

dmitripivkine:
@amicic How we are going to handle that CS works on Compressed Refs only? If we enable OMR_GC_CONCURRENT_SCAVENGER for CR only we might have problems with incompatibility of CR and Full-64 bit JVMs (command line options recognition for example). <-- Is it true? If we set OMR_GC_CONCURRENT_SCAVENGER for both we should (temporary) prevent run of CS on Full-64 bit and I believe it should be a part of this PR

amicic:
@dmitripivkine While I'm not too happy about CS being enabled on CR only, my undestanding nonCR will be enabled not too far later. 
 don't fully understand your concern. Cmdline option to enable CS is under CS build flag so it wont be visible to nonCR builds, and noone will be able to run it on nonCR.  Not sure how additionally we should 'prevent run of CS on Full-64'?

fjeremic:
> We should be uniform and probably change that behaviour (@fjeremic @vijaysun-omr do, you agreee?), that is, both Z and X86 should default to s/w based CS. But since Z s/w based is not quite ready, that new behaviour would lead to crashes.

I also agree. I'm currently working full force to polish up my solution and get it merged. Can we put this on the radar once the Z portion is in? We've already been bit a few times with the default to gencon. We also need some sort of verbose GC output line which shows whether we're using hardware or software.

dmitripivkine:

fjeremic:
> @amicic I also agree. I'm currently working full force to polish up my solution and get it merged. Can we put this on the radar once the Z portion is in? We've already been bit a few times with the default to gencon. We also need some sort of verbose GC output line which shows whether we're using hardware or software.

This has been taken care of in https://github.com/eclipse/omr/pull/3054.

dmitripivkine:

0dvictor:
I've created a separate pull request for turning CS on in the default build specs (#3253). This PR is only focus on enabling CS in the existing custom build spec (linux_x86-64_cmprssptrs_cs.spec).

PR #3075, opened 2018-09-28
Concurrent Scavenger Read Barrier evaluator on X86-64

0dvictor:
X86-64 compressedrefs now supports read barriers for Concurrent Scavenger.

Part of Issue #3054

### DIFF
--- a/runtime/codert_vm/xnathelp.m4
+++ b/runtime/codert_vm/xnathelp.m4
@@ -967,6 +967,25 @@ END_PROC(jitDecompileOnReturnJ)
 
 })	dnl ASM_J9VM_ENV_DATA64
 
+START_PROC(jitReadBarrier)
+ifdef({OMR_GC_CONCURRENT_SCAVENGER},{
+	SWITCH_TO_C_STACK
+	SAVE_C_VOLATILE_REGS
+	dnl currentThread->javaVM->memoryManagerFunctions->J9ReadBarrier(currentThread,(fj9object_t*)currentThread->floatTemp1);
+	mov _rax,uword ptr J9TR_VMThread_javaVM[_rbp]
+	mov PARM_REG(2),uword ptr J9TR_VMThread_floatTemp1[_rbp]
+	mov _rax,uword ptr J9TR_JavaVM_memoryManagerFunctions[_rax]
+	mov PARM_REG(1),_rbp
+	call uword ptr J9TR_J9MemoryManagerFunctions_J9ReadBarrier[_rax]
+	RESTORE_C_VOLATILE_REGS
+	SWITCH_TO_JAVA_STACK
+	ret
+},{ dnl OMR_GC_CONCURRENT_SCAVENGER
+	dnl not supported
+	int 3
+})	dnl OMR_GC_CONCURRENT_SCAVENGER
+END_PROC(jitReadBarrier)
+
 START_PROC(jitReferenceArrayCopy)
 	FASTCALL_EXTERN(impl_jitReferenceArrayCopy,2)
 	SWITCH_TO_C_STACK
--- a/runtime/compiler/x/codegen/J9TreeEvaluator.cpp
+++ b/runtime/compiler/x/codegen/J9TreeEvaluator.cpp
@@ -960,6 +960,80 @@ TR::Register *J9::X86::AMD64::TreeEvaluator::conditionalHelperEvaluator(TR::Node
 #endif
 
 
+TR::Register *J9::X86::TreeEvaluator::irdbarEvaluator(TR::Node *node, TR::CodeGenerator *cg)
+   {
+   TR_ASSERT(TR::Compiler->om.shouldGenerateReadBarriersForFieldLoads(), "iReadBarrierEvaluator");
+
+#ifdef OMR_GC_CONCURRENT_SCAVENGER
+
+   auto sourceMR = generateX86MemoryReference(node, cg);
+   auto address = TR::TreeEvaluator::loadMemory(node, sourceMR, TR_RematerializableLoadEffectiveAddress, false, cg);
+   address->setMemRef(sourceMR);
+   sourceMR->decNodeReferenceCounts(cg);
+
+   auto object = cg->allocateRegister();
+   auto load = generateRegMemInstruction(L4RegMem, node, object, generateX86MemoryReference(address, 0, cg), cg);
+   cg->setImplicitExceptionPoint(load);
+
+   auto begLabel = generateLabelSymbol(cg);
+   auto endLabel = generateLabelSymbol(cg);
+   auto rdbarLabel = generateLabelSymbol(cg);
+   begLabel->setStartInternalControlFlow();
+   endLabel->setEndInternalControlFlow();
+
+   auto deps = generateRegisterDependencyConditions((uint8_t)2, 2, cg);
+   deps->addPreCondition(object, TR::RealRegister::NoReg, cg);
+   deps->addPreCondition(address, TR::RealRegister::NoReg, cg);
+   deps->addPostCondition(object, TR::RealRegister::NoReg, cg);
+   deps->addPostCondition(address, TR::RealRegister::NoReg, cg);
+
+   generateLabelInstruction(LABEL, node, begLabel, cg);
+
+   generateRegMemInstruction(CMP4RegMem, node, object, generateX86MemoryReference(cg->getVMThreadRegister(), offsetof(J9VMThread, evacuateBaseCompressed), cg), cg);
+   generateLabelInstruction(JAE4, node, rdbarLabel, cg);
+
+   {
+   TR_OutlinedInstructionsGenerator og(rdbarLabel, node, cg);
+
+   generateRegMemInstruction(CMP4RegMem, node, object, generateX86MemoryReference(cg->getVMThreadRegister(), offsetof(J9VMThread, evacuateTopCompressed), cg), cg);
+   generateLabelInstruction(JA4, node, endLabel, cg);
+
+   generateMemRegInstruction(SMemReg(), node, generateX86MemoryReference(cg->getVMThreadRegister(), offsetof(J9VMThread, floatTemp1), cg), address, cg);
+   generateHelperCallInstruction(node, TR_readBarrier, NULL, cg);
+   generateRegMemInstruction(L4RegMem, node, object, generateX86MemoryReference(address, 0, cg), cg);
+   generateLabelInstruction(JMP4, node, endLabel, cg);
+   }
+
+   generateLabelInstruction(LABEL, node, endLabel, deps, cg);
+
+   if (cg->comp()->useCompressedPointers() &&
+       (node->getOpCode().hasSymbolReference() &&
+        node->getSymbolReference()->getSymbol()->getDataType() == TR::Address))
+      {
+      if (!node->getSymbolReference()->isUnresolved() &&
+          (node->getSymbolReference()->getSymbol()->getKind() == TR::Symbol::IsShadow) &&
+          (node->getSymbolReference()->getCPIndex() >= 0) &&
+          (cg->comp()->getMethodHotness()>=scorching))
+         {
+         int32_t len;
+         const char *fieldName = node->getSymbolReference()->getOwningMethod(cg->comp())->fieldSignatureChars(
+                node->getSymbolReference()->getCPIndex(), len);
+
+         if (fieldName && strstr(fieldName, "Ljava/lang/String;"))
+            {
+            generateMemInstruction(PREFETCHT0, node, generateX86MemoryReference(object, 0, cg), cg);
+            }
+         }
+      }
+
+   cg->stopUsingRegister(address);
+   node->setRegister(object);
+   return object;
+#else
+   return NULL;
+#endif
+   }
+
 // Should only be called for pure TR::wrtbar and TR::wrtbari nodes.
 //
 TR::Register *J9::X86::TreeEvaluator::writeBarrierEvaluator(TR::Node *node, TR::CodeGenerator *cg)
@@ -1612,20 +1686,20 @@ TR::Register *J9::X86::TreeEvaluator::evaluateNULLCHKWithPossibleResolve(
       {
       // pattern match the sequence under the l2a
       // NULLCHK        NULLCHK                     <- node
-      //    iaload f      l2a
+      //    aloadi f      l2a
       //       aload O       ladd
       //                       lshl
       //                          i2l
-      //                            iiload f        <- firstChild
-      //                               aload O      <- reference
+      //                            iloadi/irdbari f <- firstChild
+      //                               aload O        <- reference
       //                          iconst shftKonst
       //                       lconst HB
       //
       usingCompressedPointers = true;
 
       TR::ILOpCodes loadOp = comp->il.opCodeForIndirectLoad(TR::Int32);
-
-      while (firstChild->getOpCodeValue() != loadOp)
+      TR::ILOpCodes rdbarOp = comp->il.opCodeForIndirectReadBarrier(TR::Int32);
+      while (firstChild->getOpCodeValue() != loadOp && firstChild->getOpCodeValue() != rdbarOp)
          firstChild = firstChild->getFirstChild();
       reference = firstChild->getFirstChild();
       }
@@ -1946,7 +2020,8 @@ TR::Register *J9::X86::TreeEvaluator::evaluateNULLCHKWithPossibleResolve(
       reference->setIsNonNull(true);
       n = reference->getFirstChild();
       TR::ILOpCodes loadOp = comp->il.opCodeForIndirectLoad(TR::Int32);
-      while (n->getOpCodeValue() != loadOp)
+      TR::ILOpCodes rdbarOp = comp->il.opCodeForIndirectReadBarrier(TR::Int32);
+      while (n->getOpCodeValue() != loadOp && n->getOpCodeValue() != rdbarOp)
          {
          n->setIsNonZero(true);
          n = n->getFirstChild();
@@ -9415,9 +9490,10 @@ static TR::Register* inlineCompareAndSwapObjectNative(TR::Node* node, TR::CodeGe
       }
 
 #ifdef OMR_GC_CONCURRENT_SCAVENGER
+   TR_ASSERT(TR::Compiler->target.is64Bit() && cg->comp()->useCompressedPointers(), "Concurrent Scavenger is currently only supported on 64-bit compressedrefs.");
    if (TR::Compiler->om.shouldGenerateReadBarriersForFieldLoads())
       {
-      generateRegMemInstruction(LRegMem(use64BitClasses), node, tmp, generateX86MemoryReference(object, offset, 0, cg), cg);
+      generateRegMemInstruction(L4RegMem, node, tmp, generateX86MemoryReference(object, offset, 0, cg), cg);
 
       auto begLabel = generateLabelSymbol(cg);
       auto endLabel = generateLabelSymbol(cg);
@@ -9431,20 +9507,20 @@ static TR::Register* inlineCompareAndSwapObjectNative(TR::Node* node, TR::CodeGe
 
       generateLabelInstruction(LABEL, node, begLabel, cg);
 
-      generateRegMemInstruction(CMPRegMem(use64BitClasses), node, tmp, generateX86MemoryReference(cg->getVMThreadRegister(), offsetof(J9VMThread, evacuateBase), cg), cg);
+      generateRegMemInstruction(CMP4RegMem, node, tmp, generateX86MemoryReference(cg->getVMThreadRegister(), offsetof(J9VMThread, evacuateBaseCompressed), cg), cg);
       generateLabelInstruction(JAE4, node, rdbarLabel, cg);
 
-         {
-         TR_OutlinedInstructionsGenerator og(rdbarLabel, node, cg);
+      {
+      TR_OutlinedInstructionsGenerator og(rdbarLabel, node, cg);
 
-         generateRegMemInstruction(CMPRegMem(use64BitClasses), node, tmp, generateX86MemoryReference(cg->getVMThreadRegister(), offsetof(J9VMThread, evacuateTop), cg), cg);
-         generateLabelInstruction(JAE4, node, endLabel, cg);
+      generateRegMemInstruction(CMP4RegMem, node, tmp, generateX86MemoryReference(cg->getVMThreadRegister(), offsetof(J9VMThread, evacuateTopCompressed), cg), cg);
+      generateLabelInstruction(JA4, node, endLabel, cg);
 
-         generateRegMemInstruction(LEARegMem(), node, tmp, generateX86MemoryReference(object, offset, 0, cg), cg);
-         generateMemRegInstruction(SMemReg(), node, generateX86MemoryReference(cg->getVMThreadRegister(), offsetof(J9VMThread, floatTemp1), cg), tmp, cg);
-         generateHelperCallInstruction(node, TR_readBarrier, NULL, cg)->setNeedsGCMap(0xFF00FFFF);
-         generateLabelInstruction(JMP4, node, endLabel, cg);
-         }
+      generateRegMemInstruction(LEARegMem(), node, tmp, generateX86MemoryReference(object, offset, 0, cg), cg);
+      generateMemRegInstruction(SMemReg(), node, generateX86MemoryReference(cg->getVMThreadRegister(), offsetof(J9VMThread, floatTemp1), cg), tmp, cg);
+      generateHelperCallInstruction(node, TR_readBarrier, NULL, cg);
+      generateLabelInstruction(JMP4, node, endLabel, cg);
+      }
 
       generateLabelInstruction(LABEL, node, endLabel, deps, cg);
       }
--- a/runtime/gc_glue_java/CollectorLanguageInterfaceImpl.cpp
+++ b/runtime/gc_glue_java/CollectorLanguageInterfaceImpl.cpp
@@ -700,13 +700,35 @@ MM_CollectorLanguageInterfaceImpl::scavenger_switchConcurrentForThread(MM_Enviro
 			j9tty_printf(PORTLIB, "%p: Nursery [%p,%p] Evacuate [%p,%p] GS [%p,%p] Section size 0x%zx, sections %lu bit offset %lu bit mask 0x%zx\n",
 					vmThread, nurseryBase, nurseryTop, gsBase, gsTop, pageBase, pageTop, _extensions->getConcurrentScavengerPageSectionSize() , sectionCount, startOffsetInBits, bitMask);
 		}
-		vmThread->evacuateBase = gsBase;
-		vmThread->evacuateTop = gsTop;
+		/*
+		 * vmThread->evacuateTop is defined as the last address possible in the evacuate region;
+		 * however, gsTop is the first address above the evacuate region;
+		 * therefore, vmThread->evacuateTop = gsTop - 1.
+		 * In short, the evacuate region is [vmThread->evacuateBase, vmThread->evacuateTop].
+		 */
+		vmThread->evacuateBase = (UDATA)gsBase;
+		vmThread->evacuateTop = (UDATA)gsTop - 1;
+#if defined(J9VM_GC_COMPRESSED_POINTERS)
+		vmThread->evacuateBaseCompressed = _extensions->accessBarrier->convertTokenFromPointer((mm_j9object_t)vmThread->evacuateBase);
+		vmThread->evacuateTopCompressed = _extensions->accessBarrier->convertTokenFromPointer((mm_j9object_t)vmThread->evacuateTop);
+#endif /* J9VM_GC_COMPRESSED_POINTERS */
 		vmThread->privateFlags |= J9_PRIVATE_FLAGS_CONCURRENT_SCAVENGER_ACTIVE;
 		j9gs_enable(&vmThread->gsParameters, _extensions->getConcurrentScavengerPageStartAddress(), _extensions->getConcurrentScavengerPageSectionSize(), bitMask);
     } else {
 		j9gs_disable(&vmThread->gsParameters);
+		/*
+		 * By setting evacuateTop to NULL and evacuateBase to ~evacuateTop
+		 * it gives an empty range that contains no address. Therefore,
+		 * when decide whether to read barrier, a simple range is sufficient and
+		 * checking J9_PRIVATE_FLAGS_CONCURRENT_SCAVENGER_ACTIVE can be skipped.
+		 */
 		vmThread->privateFlags &= ~J9_PRIVATE_FLAGS_CONCURRENT_SCAVENGER_ACTIVE;
+		vmThread->evacuateBase = UDATA_MAX;
+		vmThread->evacuateTop = 0;
+#ifdef J9VM_GC_COMPRESSED_POINTERS
+		vmThread->evacuateBaseCompressed = U_32_MAX;
+		vmThread->evacuateTopCompressed = 0;
+#endif
     }
 }
 #endif /* OMR_GC_CONCURRENT_SCAVENGER */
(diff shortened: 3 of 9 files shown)


0dvictor:
Requires https://github.com/eclipse/omr/pull/2974 and https://github.com/eclipse/omr/pull/3012

runtime/codert_vm/xnathelp.m4
```diff
@@ -967,4 +967,23 @@ END_PROC(jitDecompileOnReturnJ)
...
+	mov    rsi,QWORD PTR [rbp+J9TR_VMThread_floatTemp1]
+	dnl javaVM->memoryManagerFunctions
+	mov    rax,QWORD PTR [rax+0x140]
+	mov    rdi, rbp
+	dnl memoryManagerFunctions->J9ReadBarrier
+	call   QWORD PTR [rax+0x1f8]
+
+	RESTORE_ALL_REGS
+	SWITCH_TO_JAVA_STACK
+	ret
```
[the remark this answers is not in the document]
gacholio: Yes , it would be nice to use consistent spacing and capitalization, as well as using the const[reg] format rather than [reg+const]. Also, "uword ptr" is preferred for pointer-sized operations (not that I think this code will ever be used on 32-bit), and the _reg syntax for fullword registers.

Also, should this new code be under the 64-bit ifdef?

runtime/vm/vmthread.c
```diff
@@ -2364,7 +2364,10 @@ threadAboutToStart(J9VMThread *currentThread)
 	registerSystemStackPointerThreadOffset(currentThread);
 #endif
 
-#if defined(OMR_GC_CONCURRENT_SCAVENGER) && defined(J9VM_ARCH_S390)
+#if defined(OMR_GC_CONCURRENT_SCAVENGER)
+	currentThread->evacuateBase = (void*)-1;
+	currentThread->evacuateTop  = (void*)0;
+#if defined(J9VM_ARCH_S390)
```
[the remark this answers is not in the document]
0dvictor: `j9gs_initializeThread()` sounds like Guarded Storage, which is a Z-only concept. X86 doesn't have any thing similar to Guarded Storage; thus I don't think implement these logic inside `j9gs_initializeThread()` is proper on X86. In addition, these two lines are to initialize `evacuteBase` and `evacuateTop`, which in theory applies to all platforms.

runtime/gc_glue_java/CollectorLanguageInterfaceImpl.cpp
```diff
@@ -700,13 +700,20 @@ MM_CollectorLanguageInterfaceImpl::scavenger_switchConcurrentForThread(MM_Enviro
...
 		vmThread->evacuateBase = gsBase;
 		vmThread->evacuateTop = gsTop;
+#endif /* J9VM_GC_COMPRESSED_POINTERS && J9VM_ARCH_X86 */
 		vmThread->privateFlags |= J9_PRIVATE_FLAGS_CONCURRENT_SCAVENGER_ACTIVE;
 		j9gs_enable(&vmThread->gsParameters, _extensions->getConcurrentScavengerPageStartAddress(), _extensions->getConcurrentScavengerPageSectionSize(), bitMask);
     } else {
 		j9gs_disable(&vmThread->gsParameters);
 		vmThread->privateFlags &= ~J9_PRIVATE_FLAGS_CONCURRENT_SCAVENGER_ACTIVE;
+		vmThread->evacuateBase = (void*)-1;
+		vmThread->evacuateTop = (void*)0;
```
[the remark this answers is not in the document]
0dvictor: `(void*)-1` gives the max value that a `void*` possibly be, and `(void*)0` gives the min value that a `void*` possibly be. By setting `evacuateBase` to `(void*)-1` and `evacuateTop` to `(void*)0`, it gives an empty range that contains no address. Therefore, when decide whether to read barrier, a simple range is sufficient and checking `J9_PRIVATE_FLAGS_CONCURRENT_SCAVENGER_ACTIVE` can be skipped.

runtime/codert_vm/xnathelp.m4
```diff
@@ -967,4 +967,23 @@ END_PROC(jitDecompileOnReturnJ)
...
+START_PROC(jitReadBarrier)
+	SWITCH_TO_C_STACK
+	SAVE_ALL_REGS
+
+	dnl currentThread->javaVM->memoryManagerFunctions->J9ReadBarrier(currentThread, (fj9object_t*)currentThread->floatTemp1);
+
+	mov    rax,QWORD PTR [rbp+J9TR_VMThread_javaVM]
+	mov    rsi,QWORD PTR [rbp+J9TR_VMThread_floatTemp1]
+	dnl javaVM->memoryManagerFunctions
+	mov    rax,QWORD PTR [rax+0x140]
```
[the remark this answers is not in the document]
0dvictor: Thank-you for pointing this out, this part is not mature yet, the offset magical number will be updated to an entry from jilconsts in the final version.

runtime/codert_vm/xnathelp.m4
```diff
@@ -967,4 +967,23 @@ END_PROC(jitDecompileOnReturnJ)
...
+	mov    rdi, rbp
+	dnl memoryManagerFunctions->J9ReadBarrier
+	call   QWORD PTR [rax+0x1f8]
```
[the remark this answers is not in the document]
0dvictor: Thank-you for pointing this out, this part is not mature yet, the offset magical number will be updated to an entry from jilconsts in the final version.

runtime/codert_vm/xnathelp.m4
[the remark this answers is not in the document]
0dvictor: Thank-you for pointing this out, this part is not mature yet, the format will be updated according to both comments.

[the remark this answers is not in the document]
gacholio: Also, would it be more efficient to either:

- use a cnathelp.cpp helper with an inline (or mostly inline) version of the barrier to avoid the cross-module call?

- call the MM function directly from the compiled code

Another point - this clearly isn't a GC point, so there's no need to save all the registers - you can get away with only saving the C volatile registers (`SAVE_C_VOLATILE_REGS` / `RESTORE_C_VOLATILE_REGS`)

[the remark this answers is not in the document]
0dvictor: Good suggestions. Will definitely evaluate both approaches.
I thought read barrier was a GC point thus I saved all registers. Will update it to only save volatile ones.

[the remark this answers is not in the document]
gacholio: If it is a GC point, then you would need to build a resolve frame (in cnathelp) before calling the function. I don't see anything similar on Z - wonder what they're doing?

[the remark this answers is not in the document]
0dvictor: Good point.
@fjeremic Are read barriers considered as GC points on Z?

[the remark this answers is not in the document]
gacholio: From a scan of the J9ReadBarrier implementation, it is certainly not a GC point.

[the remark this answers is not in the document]
gacholio: The path on Z is quite inefficient - too many registers saved, multiple cross-library calls. There really shouldn't be wildly different paths for X and Z, so I suspect a cleanup of the existing code along with your changes would benefit both architectures.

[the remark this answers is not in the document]
gacholio: @amicic @dmitripivkine I think this would benefit from an inline .hpp version of the barrier, calling to the GC only when necessary, which could be used in cnathelp.cpp.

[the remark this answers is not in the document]
amicic: In ObjectAccessBarrierAPI.hpp there are already APIs like internalPreReadObject/preMixedObjectReadObject/preIndexableObjectReadObject (there is a static variant too, currently not active, but will be eventually be).

However, they end up doing a cross module call. Yes we could inline the implementation to save on cross module (possible one more virtual dispatch and/or another minor check or two). Still, I would not expect much - the barrier itself is relatively heavy.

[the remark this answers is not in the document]
gacholio: I suppose it also depends on how much of the barrier work is inlined in the compiled code - maybe the call is only for unexpected cases.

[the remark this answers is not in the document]
amicic: Fairly little is inlined in compiled code - it's the range check. this is what I meant by 'minor checks' that could be avoided in the OABAPI implementation.

If JIT called the barrier that already means there is non-trivial work to do. as a min (if object is already copied) it will be a ref. update (which is an atomic), or worse to win the copying (atomic) and  copy the object. there are other scenarios, but these two are most common, split about 50-50.

The best savings would come from speeding up the range checks, but this is where only specialized h/w could further help.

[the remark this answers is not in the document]
0dvictor: @gacholio I've updated this method, do you mind take a look once more?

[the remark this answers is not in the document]
0dvictor: > use a cnathelp.cpp helper with an inline (or mostly inline) version of the barrier to avoid the cross-module call?

From the discussions above, the only thing JIT should be inlined is the range check, which is already inlined be JIT. Hence creating a cnathelp.cpp helper is not necessary.

> call the MM function directly from the compiled code

Calling read barrier is considered rare, calling the MM function from JITted code would create too much register pressure and uses too much unnecessary code cache. The whole purpose of this xnathelp.m4 wrapper is to reduce register pressure and save JIT code cache.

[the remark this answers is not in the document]
gacholio: The spacing is still different. The registers loads are linux-specific, whereas this file is shared with windows as well - might be worth defining a few arg register macros to make this common. All but the first comment seem redundant.

runtime/jilgen/jilconsts.c
```diff
@@ -521,12 +521,14 @@ writeConstants(OMRPortLibrary *OMRPORTLIB, IDATA fd)
 			writeConstant(OMRPORTLIB, fd, "J9TR_JavaVM_bytecodeLoop", offsetof(J9JavaVM, bytecodeLoop)) |
 			writeConstant(OMRPORTLIB, fd, "J9TR_JavaVM_extendedRuntimeFlags", offsetof(J9JavaVM, extendedRuntimeFlags)) |
 			writeConstant(OMRPORTLIB, fd, "J9TR_JavaVMInternalFunctionTable", offsetof(J9JavaVM, internalVMFunctions)) |
+            writeConstant(OMRPORTLIB, fd, "J9TR_JavaVM_memoryManagerFunctions", offsetof(J9JavaVM, memoryManagerFunctions)) |
```
gacholio: Spacing! Same  comment below.

runtime/vm/vmthread.c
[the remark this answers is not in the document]
gacholio: At the very least, this is worth a detailed comment.

runtime/gc_glue_java/CollectorLanguageInterfaceImpl.cpp
```diff
@@ -700,13 +700,20 @@ MM_CollectorLanguageInterfaceImpl::scavenger_switchConcurrentForThread(MM_Enviro
 			j9tty_printf(PORTLIB, "%p: Nursery [%p,%p] Evacuate [%p,%p] GS [%p,%p] Section size 0x%zx, sections %lu bit offset %lu bit mask 0x%zx\n",
 					vmThread, nurseryBase, nurseryTop, gsBase, gsTop, pageBase, pageTop, _extensions->getConcurrentScavengerPageSectionSize() , sectionCount, startOffsetInBits, bitMask);
 		}
+#if defined(J9VM_GC_COMPRESSED_POINTERS) && defined(J9VM_ARCH_X86)
```
gacholio: I'd like to see a GC reviewer on this bit.

runtime/vm/vmthread.c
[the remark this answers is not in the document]
0dvictor: Updated.

runtime/jilgen/jilconsts.c
0dvictor: Both have been updated.

runtime/gc_glue_java/CollectorLanguageInterfaceImpl.cpp
0dvictor: @amicic Could you take a look?

runtime/codert_vm/xnathelp.m4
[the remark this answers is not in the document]
0dvictor: Updated. Could you take a look once more?

runtime/codert_vm/xnathelp.m4
```diff
@@ -967,6 +967,25 @@ END_PROC(jitDecompileOnReturnJ)
 
 })	dnl ASM_J9VM_ENV_DATA64
 
+START_PROC(jitReadBarrier)
+ifdef({OMR_GC_CONCURRENT_SCAVENGER},{
+	SWITCH_TO_C_STACK
+	SAVE_C_VOLATILE_REGS
+	dnl currentThread->javaVM->memoryManagerFunctions->J9ReadBarrier(currentThread, (fj9object_t*)currentThread->floatTemp1);
+	mov _rax, uword ptr J9TR_VMThread_javaVM[_rbp]
```
gacholio: Since I'm being incredibly nitpicky - the rest of the code does not have space after comma.

gacholio:
I'd forgotten about the PARM_REG macro - doesn't work on 32-bit, but I'll implement it now that we're using fastcall (at your original suggestion), so at least things with 2 or fewer parameters will work everywhere, and will error out properly if we try to use more than 2.

runtime/gc_glue_java/CollectorLanguageInterfaceImpl.cpp
gacholio: In particular, would it make sense to have some `#error` in here when CS is defined, but the platforms options don't have a good answer for the values (and some comments).

runtime/vm/vmthread.c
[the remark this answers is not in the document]
gacholio: I dislike the fact that this code exists in the GC and VM - is there a way to commonize it? Not your fault by any means. Magic values in more than one place are bad.

runtime/codert_vm/xnathelp.m4
0dvictor: Thank-you for pointing it out. It has been updated.

runtime/vm/vmthread.c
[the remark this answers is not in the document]
0dvictor: Totally agreed. Let me think about it.

runtime/codert_vm/xnathelp.m4
```diff
@@ -967,6 +967,25 @@ END_PROC(jitDecompileOnReturnJ)
...
+	mov _rax,uword ptr J9TR_JavaVM_memoryManagerFunctions[_rax]
+	mov PARM_REG(1),_rbp
+	call uword ptr J9TR_J9MemoryManagerFunctions_J9ReadBarrier[_rax]
+	RESTORE_C_VOLATILE_REGS
+	SWITCH_TO_JAVA_STACK
+	ret
+},{ dnl OMR_GC_CONCURRENT_SCAVENGER
+	dnl not supported
+	int 3
+})
```
gacholio: endif comment here please.

andrewcraik:
@DanHeidinga could you have a look now that @gacholio has approved to see if you are happy - I'm going to do a review now from the JIT side.

@0dvictor could you confirm if this still needs to be WIP? The dependencies have been merged I believe and it looks like we should be done a review soon so I'd like to make sure this is ready to go soon if possible.

0dvictor:
> could you confirm if this still needs to be WIP? The dependencies have been merged I believe and it looks like we should be done a review soon so I'd like to make sure this is ready to go soon if possible.

It is currently marked as WIP only because I have a test run pending. I'll remove the WIP as soon as it cleared.

runtime/compiler/x/codegen/J9TreeEvaluator.cpp
```diff
@@ -960,6 +960,79 @@ TR::Register *J9::X86::AMD64::TreeEvaluator::conditionalHelperEvaluator(TR::Node
 #endif
 
 
+TR::Register *J9::X86::TreeEvaluator::irdbarEvaluator(TR::Node *node, TR::CodeGenerator *cg)
+   {
+   TR_ASSERT(TR::Compiler->om.shouldGenerateReadBarriersForFieldLoads(), "iReadBarrierEvaluator");
```
andrewcraik: Hmm - this will be changed soon by the field watch work so I'm wondering if we should structure this more like if shouldGenerateReadBarriersForFieldLoads then do this sequence otherwise just treat it like an aloadi and other can add other meanings and uses?

runtime/gc_glue_java/CollectorLanguageInterfaceImpl.cpp
```diff
@@ -700,13 +700,20 @@ MM_CollectorLanguageInterfaceImpl::scavenger_switchConcurrentForThread(MM_Enviro
...
+		vmThread->evacuateBase = (void*)(((uintptr_t)gsBase) >> vmThread->javaVM->compressedPointersShift);
+		vmThread->evacuateTop = (void*)(((uintptr_t)gsBase) >> vmThread->javaVM->compressedPointersShift);
```
DanHeidinga: Why don't the calls to get the initial value just return this constant?  Then the z an x code can be the same:

ie: `void* gsBase = _extensions->scavenger->getEvacuateBase();`

runtime/vm/vmthread.c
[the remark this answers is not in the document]
DanHeidinga: This really belongs in a common helper - either by implementing the `j9gs_initializeThread` call on x or by implementing a new common call to intialize concurrent scavenger on both x & z

DanHeidinga:
@amicic Are you OK with the initialization code?  You're most familiar with the existing sequence on z

amicic:
@DanHeidinga I put a bunch of comments last night but they stayed at pending state. Just pushed them through with 'submit reveiw'

runtime/gc_glue_java/CollectorLanguageInterfaceImpl.cpp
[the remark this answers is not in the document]
0dvictor: > Second one is incorrect, should be based on gsTop.

Good catch, fixed.

> It's cleaner to use _extensions->accessBarrier->convertTokenFromPointer() API. It already has #if defined(J9VM_GC_COMPRESSED_POINTERS) in it.

Good suggestion, updated.

runtime/vm/vmthread.c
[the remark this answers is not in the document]
amicic: @fjeremic These two VMThread fields are turning into compressed refs. I hope you are ok with it, and that Z platfrom s/w barrier will use them it such a form too. Note that they are currently used in non compressed form by the first attempt (peephole pass), but I assume that code will go away.

andrewcraik:
@0dvictor have the concerns from @amicic been addressed? If so please @DanHeidinga directly to get him to complete his review.

0dvictor:
> @0dvictor have the concerns from @amicic been addressed? If so please @DanHeidinga directly to get him to complete his review.

Sorry, I haven't fully addressed his concern yet. I need some research on initializing `evacuateBase` in the right place as well as see think about whether we want to have another pair for compressed base/top.

runtime/gc_glue_java/CollectorLanguageInterfaceImpl.cpp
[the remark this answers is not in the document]
0dvictor: The magic number has been avoided.

runtime/gc_glue_java/CollectorLanguageInterfaceImpl.cpp
0dvictor: Updated to keep both uncompressed and compressed ranges, so that X and Z code are now the same.

runtime/vm/vmthread.c
```diff
@@ -210,6 +210,16 @@ allocateVMThread(J9JavaVM * vm, omrthread_t osThread, UDATA privateFlags, void *
...
 #endif
 
+#ifdef OMR_GC_CONCURRENT_SCAVENGER
+	/* Initialize fields used by Concurrent Scavenger */
+	newThread->evacuateTop = NULL;
+	newThread->evacuateBase = (void*)~(UDATA)newThread->evacuateTop;
+#ifdef J9VM_GC_COMPRESSED_POINTERS
+	newThread->evacuateTopCompressed = 0;
+	newThread->evacuateBaseCompressed = ~newThread->evacuateTopCompressed;
+#endif
```
DanHeidinga: nitpick - add the closing `/* J9VM_GC_COMPRESSED_POINTERS */` to endif's

runtime/vm/vmthread.c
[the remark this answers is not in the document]
0dvictor: I realized that the range shouldn't be part of "thread initialization"; instead it should be part of initialization of vmthread structure along with other fields. I've moved the code and avoided the magical numbers.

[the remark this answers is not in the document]
0dvictor: @amicic I added another pair of range that dedicate for compressed refs. Now we have both uncompressed and compressed ranges.

runtime/vm/vmthread.c
0dvictor: Updated.

runtime/oti/j9nonbuilder.h
```diff
@@ -5000,6 +5000,10 @@ typedef struct J9VMThread {
 	struct J9GSParameters gsParameters;
 	void* evacuateBase;
 	void* evacuateTop;
+#if defined(J9VM_GC_COMPRESSED_POINTERS)
+	U_32 evacuateBaseCompressed;
+	U_32 evacuateTopCompressed;
+#endif /* J9VM_GC_COMPRESSED_POINTERS */
```
amicic: This isn't the way I exactly imagined, but this could work, too. Like this, the code generation needs to check if it generates the code for 64bit CR or nonCR and use one set of bounds or the other (which it isn't as I can see - it just uses the CR set of bounds).

BTW I imagined that that even evacuateBase(Top)Compressed would also be 64bit (basically it would be of type fj9object_t *) and would be used for code that deals with heap references (that could be, but not necessarily are compressed, like if we run 64bit nonCR), and the original set of evacuateBase/Top (basically of type j9object_t *) would be used with code that deal with off heap structures, locals, native method arguments etc, that are known to deal with already uncompressed form of a heap reference.

gacholio: Not clear to me why we need separate variables for compressed and not - only one can be true in any given build.

0dvictor: I see your point, but IMO for nonCR build, there is no need to keep two identical ranges. The JIT evaluator need a separate evaluator for nonCR build anyway, so that I don't see the advantage keep two identical ranges on nonCR.

amicic: In CR build we need both forms (CR and nonCR) for the ranges. When jitted code wants to check a range for a reference just read from an object in heap it wants to compare it with compressed from of ranges .
If jitted code (or any other VM code) wants to check a range for reference just read (for example) from class static it wants to compare it with non-compressed form of ranges.

amicic: @0dvictor Yes, understand your point. A possible advantage of the other approach is to avoid runtime checkes in non generated code (when running CR) that need to distinguish between the two set of ranges. Maybe there are not many (or at all, now), but in future CR and non-CR can actually get merged.

I don't have a strong preference, both approaches are ok to me. If someone has stronger preference let's hear it.

runtime/compiler/x/codegen/J9TreeEvaluator.cpp
0dvictor: Good suggestion; however without seeing the actual change of field watch work, it is hard to image exactly what need be checked. If field watch work is merged before this PR, I would definitely updated this change accordingly; otherwise, I suggest to massage the code when field watch is delivered.

runtime/gc_glue_java/CollectorLanguageInterfaceImpl.cpp
```diff
@@ -702,11 +702,27 @@ MM_CollectorLanguageInterfaceImpl::scavenger_switchConcurrentForThread(MM_Enviro
...
+		 * it gives an empty range that contains no address. Therefore,
+		 * when decide whether to read barrier, a simple range is sufficient and
+		 * checking J9_PRIVATE_FLAGS_CONCURRENT_SCAVENGER_ACTIVE can be skipped.
+		 */
 		vmThread->privateFlags &= ~J9_PRIVATE_FLAGS_CONCURRENT_SCAVENGER_ACTIVE;
+		vmThread->evacuateTop = NULL;
+		vmThread->evacuateBase = (void*)~(UDATA)vmThread->evacuateTop;
+#ifdef J9VM_GC_COMPRESSED_POINTERS
+		vmThread->evacuateTopCompressed = 0;
+		vmThread->evacuateBaseCompressed = ~vmThread->evacuateTopCompressed;
```
amicic: I actually prefer using (void *)UDATA_MAX, it's more in coding spirit with the rest of GC code.

(I would actually even more prefer UINT_MAX, but the code is still heavily using UDATA_MAX, so let's be consistent)

0dvictor: Updated.

runtime/compiler/x/codegen/J9TreeEvaluator.cpp
```diff
@@ -960,6 +960,80 @@ TR::Register *J9::X86::AMD64::TreeEvaluator::conditionalHelperEvaluator(TR::Node
...
+
+   auto sourceMR = generateX86MemoryReference(node, cg);
+   auto address = TR::TreeEvaluator::loadMemory(node, sourceMR, TR_RematerializableLoadEffectiveAddress, false, cg);
+   address->setMemRef(sourceMR);
+   sourceMR->decNodeReferenceCounts(cg);
+
+   auto object = cg->allocateRegister();
+   auto load = generateRegMemInstruction(L4RegMem, node, object, generateX86MemoryReference(address, 0, cg), cg);
+   cg->setImplicitExceptionPoint(load);
+
```
vijaysun-omr: What does setting as an implicit exception point do exactly ? What is the exception that's possible ?

The other question I had was whether a read barrier opcode should only appear under a node that is at the top of the tree at codegen time (i.e. after trees have been lowered). Looking at some IL from logs with the concurrent scavenger enabled, I did not see it appear only under a treetop and so I was wondering if this was by design.

runtime/compiler/x/codegen/J9TreeEvaluator.cpp
```diff
@@ -9405,20 +9480,20 @@ static TR::Register* inlineCompareAndSwapObjectNative(TR::Node* node, TR::CodeGe
 
       generateLabelInstruction(LABEL, node, begLabel, cg);
 
-      generateRegMemInstruction(CMPRegMem(use64BitClasses), node, tmp, generateX86MemoryReference(cg->getVMThreadRegister(), offsetof(J9VMThread, evacuateBase), cg), cg);
+      generateRegMemInstruction(CMPRegMem(use64BitClasses), node, tmp, generateX86MemoryReference(cg->getVMThreadRegister(), offsetof(J9VMThread, evacuateBaseCompressed), cg), cg);
```
vijaysun-omr: Will this code work on 64-bit uncompressed or 32-bit ?

0dvictor: Good point. Concurrent Scavenger is currently only supported on 64-bit compressedrefs. I've update this part to avoid confusion.

runtime/compiler/x/codegen/J9TreeEvaluator.cpp
0dvictor: The `irdbari` node may be under a `NULLCHK` and hence this is the implicit NULLCHK point.

There is no restriction or necessity for read barrier to have its own treetop. The discussion following https://github.com/eclipse/openj9/pull/3075/files#r221504019 concluded read barrier is not a GC point.

gacholio: If this is the same node which will be used for field watch, it most certainly will be a GC point in that case.

0dvictor: Field watch is irrelevant to this Concurrent Scavenger change. Even it is a GC point for field watch, it is NOT a GC point when field watch is disabled/off. This evaluator does NOT assume whether `irdbari` is a GC point, and should work regardless. It is up to the Optimizer generating the trees, and this evaluator generates instructions for Concurrent Scavenger regardless whether `irdbari` has its own treetop or not. Whether the Optimizer should generate `irdbari` under a dedicate treetop is beyong the scope of this changeset.

vijaysun-omr: > 
> 
> The `irdbari` node may be under a `NULLCHK` and hence this is the implicit NULLCHK point.
> 
> There is no restriction or necessity for read barrier to have its own treetop. The discussion following https://github.com/eclipse/openj9/pull/3075/files#r221504019 concluded read barrier is not a GC point.

Is that the right link ?

0dvictor: > Is that the right link ?

Yes, the link is correct; but I guess this is a bug in GitHub preventing it jumping to the correct comment. If you click the link, then click "Show conversation" right under "runtime/codert_vm/xnathelp.m4", it should show the correct discussion.

runtime/compiler/x/codegen/J9TreeEvaluator.cpp
```diff
@@ -9405,20 +9481,20 @@ static TR::Register* inlineCompareAndSwapObjectNative(TR::Node* node, TR::CodeGe
 
       generateLabelInstruction(LABEL, node, begLabel, cg);
 
-      generateRegMemInstruction(CMPRegMem(use64BitClasses), node, tmp, generateX86MemoryReference(cg->getVMThreadRegister(), offsetof(J9VMThread, evacuateBase), cg), cg);
+      generateRegMemInstruction(CMP4RegMem, node, tmp, generateX86MemoryReference(cg->getVMThreadRegister(), offsetof(J9VMThread, evacuateBaseCompressed), cg), cg);
```
amicic: On Z platforms CS runs on both 64 CR and nonCR, but not on 32. Same should be on X. But if you want to stage it and enable only CR now and later for nonCR that's ok, but nonCR should be supported eventually.

0dvictor: Thank you for reminding. This version only supports 64 CR. There are more changes required beyond this PR to make CS on nonCR work. CS support on nonCR will be added after CR is fully enabled.

runtime/gc_glue_java/CollectorLanguageInterfaceImpl.cpp
```diff
@@ -700,13 +700,35 @@ MM_CollectorLanguageInterfaceImpl::scavenger_switchConcurrentForThread(MM_Enviro
...
-		vmThread->evacuateBase = gsBase;
-		vmThread->evacuateTop = gsTop;
+		/*
+		 * vmThread->evacuateTop is defined as the last address possible in the evacuate region;
+		 * however, gsTop is the first address above the evacuate region;
+		 * therefore, vmThread->evacuateTop = gsTop - 1.
+		 * In short, the evacuate region is [vmThread->evacuateBase, vmThread->evacuateTop].
+		 */
+		vmThread->evacuateBase = (UDATA)gsBase;
+		vmThread->evacuateTop = (UDATA)gsTop - 1;
```
amicic: I understand that this line has been modified after I approved the PR. I approve this additional change, too. 

This change is necessary for CR config in case gsTop is at highest possible for CR range, in which case evacuateTopCompressed  would be 0x100000000 after the shift is applied, which would not fit into a 32bit variable/register. Like this highest possible evacuateTopCompressed would be 0xffffffff.

andrewcraik:
Jenkins test sanity,extended xlinux,win jdk8,jdk11

andrewcraik:

andrewcraik:
This is all reviewed and all approved so I am going to merge this. Sanity is clean and test have run successfully with CS enabled in additional testing. Note this does not enable the feature it is the implementation - #3070 is the enablement

dmitripivkine:
> This is all reviewed and all approved so I am going to merge this. Sanity is clean and test have run successfully with CS enabled in additional testing. Note this does not enable the feature it is the implementation - #3070 is the enablement

please do not miss the fact that even #3070 would not activate CS at x86. OMR_GC_CONCURRENT_SCAVENGER should be set as well

Issue #2994, opened 2018-09-24
Segmentation Error When compiling with debug enabled on PPC


AlenBadel:
This only occurs when building on PPCLE, and with java versions 10,11. Java 8 builds perfectly. 

This occurs during the DDR stage, when it uses the built jdk. 

Passes with -Xint. 

To reproduce,

I checked out the OpenJDK 11 with OpenJ9 project per
https://www.eclipse.org/openj9/oj9_build.html

Enabled debugging per https://blog.openj9.org/2018/06/05/debugging-openj9-in-docker-with-gdb/

```
export UMA_DO_NOT_OPTIMIZE_CCODE="1"
export VMDEBUG="-g3 -fno-inline -O0"
export VMLINK="-g -O0"
export enable_debug=yes
export enable_optimized=no
export enable_optimize=no
export CXXFLAGS="-O0 -g3"
export CFLAGS="-O0 -g3"
export BUILD_CONFIG=debug
```

To build:

```
cd /projects/openj9/openj9-openjdk-jdk11/
bash get_source.sh 
bash ./configure --with-freemarker-jar=/projects/freemarker.jar 
make all
```

We get the following error

```
Type=Segmentation error vmState=0x00000000
J9Generic_Signal_Number=00000004 Signal_Number=0000000b Error_Value=00000000 Signal_Code=00000001
Handler1=000071479F7829A0 Handler2=000071479F417384
R0=0000714768CC06EC R1=00007147A01CAD30 R2=00007147A0760E28 R3=0000000000078200
R4=000000000007A100 R5=0000000000000001 R6=000000000007A100 R7=0000000000000000
R8=0000000000000001 R9=0000000000078200 R10=0000000000000000 R11=000071479EA25658
R12=00007147A0770E28 R13=00007147A01D68E0 R14=0000000000044220 R15=0000000000044500
R16=0000714768250038 R17=FFFFFFFFFFFFFFFF R18=000071479826FD5C R19=000071479800DCD0
R20=000071479826FD5C R21=0000714768032754 R22=00007147680304C0 R23=0000000000000000
R24=00007147A01CAFA0 R25=0000000000074E00 R26=00007147980B93A8 R27=0000714798006BD0
R28=0000000000000000 R29=00007147A01CAFA0 R30=00007147A01CB1B0 R31=00007147A01CAD30
NIP=000071479DCD7B10 MSR=900000010280F033 ORIG_GPR3=000071479EA256B0 CTR=000071479EA25658
[...]

PR #3083, opened 2018-10-01
PPC ABI updates and optimizations

gacholio:
- c_cInterpreter does not use TOC, so don't load it
- c_cInterpreter now follows PPC LE ABI for entry
- unnecessary nop removed from direct calls
- fetch jitTOC from J9VMThread rather than J9JavaVM
- inline ptrgl for AIX
- don't save/restore TOC across calls as they are in the same module

There were no issues with the incorrect ABI in c_cInterpreter on PPC LE,
but it seems wise to follow the ABI in case future code needs the TOC,
or c_cInterpreter is called from another module.

See: #2994

[ci skip]

### DIFF
--- a/runtime/oti/phelpers.m4
+++ b/runtime/oti/phelpers.m4
@@ -1,4 +1,4 @@
-dnl Copyright (c) 1991, 2017 IBM Corp. and others
+dnl Copyright (c) 1991, 2018 IBM Corp. and others
 dnl
 dnl This program and the accompanying materials are made available under
 dnl the terms of the Eclipse Public License 2.0 which accompanies this
@@ -100,8 +100,7 @@ define({staddrx},stdx)
 define({staddru},stdu)
 define({cmpliaddr},{cmpli 0,1,})
 define({ADDR},.llong)
-define({ALen},8)  
-
+define({ALen},{J9CONST(8,$1,$2)})
 define({J9VMTHREAD},{r15})
 
 },{ dnl ASM_J9VM_ENV_DATA64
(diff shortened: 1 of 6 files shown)


runtime/oti/phelpers.m4
```diff
@@ -135,7 +135,10 @@ define({FUNC_PTR},{r11})
 
 define({CALL_INDIRECT},{
 	ifelse($1,,,{mr FUNC_PTR,$1})
-	bl ._ptrgl[pr]
+	laddr r2,0(FUNC_PTR)
```
DanHeidinga: Can you add a comment that says this is an inlined version of `_ptrgl`?

runtime/oti/phelpers.m4
```diff
@@ -135,7 +135,10 @@ define({FUNC_PTR},{r11})
...
+	mtctr r2
+	laddr r2,eval(ALen)(FUNC_PTR)
```
DanHeidinga: what is `eval(ALen)`?

gacholio: It's an m4ism - ALen is a macro defined to a number, so if Alen is 4, Alen(reg) evaluates to 4. I'll change Alen to the fancy constant macro that handles argument properly.

runtime/oti/phelpers.m4
gacholio: Done.

gacholio:
Personal testing across all PPC platforms passes with the latest changes.

DanHeidinga:
Does the jit debug build succeed with this change?

DanHeidinga:
Jenkins test sanity plinux,aix jdk8

PR #3085, opened 2018-10-01
Fix javadoc for toGenericString()

keithc-ca:
Fix UMA conditional content and use `&lt;` instead of `&#60;`, etc.


pshipton:
jenkins compile zlinux jdk8,jdk11

keithc-ca:
Apparently javadoc doesn't like `&commat;`; reverted to `&#64;`.

pshipton:
jenkins compile jdk11

PR #3086, opened 2018-10-01
Allow some Unsafe atomic ops to be inlined on POWER

hzongaro:
Allow for the possibility that the parent OMR class will want to
inline a direct call by delegating to the parent class's implementation
of inlineDirectCall.  In particular, we will want to do that for some
atomic operations in Unsafe.

This depends on https://github.com/eclipse/omr/pull/3018/


runtime/compiler/optimizer/J9RecognizedCallTransformer.cpp
```diff
@@ -139,10 +139,19 @@ bool J9::RecognizedCallTransformer::isInlineable(TR::TreeTop* treetop)
       {
       case TR::sun_misc_Unsafe_getAndAddInt:
       case TR::sun_misc_Unsafe_getAndSetInt:
-         return TR::Compiler->target.cpu.isX86() && !comp()->getOption(TR_DisableUnsafe) && !comp()->compileRelocatableCode() && !TR::Compiler->om.canGenerateArraylets();
+         return (TR::Compiler->target.cpu.isX86()
+                       || TR::Compiler->target.cpu.isPower())
+                    && !comp()->getOption(TR_DisableUnsafe)
+                    && !comp()->compileRelocatableCode()
+                    && !TR::Compiler->om.canGenerateArraylets();
```
fjeremic: These queries will need updating after https://github.com/eclipse/omr/pull/2999 is merged. I suggest we mark this PR as WIP until this is done.

andrewcraik:
@hzongaro eclipse/omr#2999 has been merged - can you update to account for that change so that review can progress?

fjeremic:
> @hzongaro [eclipse/omr#2999](https://github.com/eclipse/omr/pull/2999) has been merged - can you update to account for that change so that review can progress?

I think a large part of this PR is made redundant by #3195. @hzongaro given #3195 I actually don't think this PR should exist at all.

runtime/compiler/p/codegen/J9TreeEvaluator.cpp
```diff
@@ -11824,6 +11824,10 @@ J9::Power::CodeGenerator::inlineDirectCall(TR::Node *node, TR::Register *&result
       resultReg = inlineSinglePrecisionFP(node, TR::InstOpCode::fsqrts, cg);
       return true;
       }
+   else if (OMR::Power::CodeGenerator::inlineDirectCall(node, resultReg))
+      {
+      return true;
+      }
```
fjeremic: This is not the right way to call the underlying OMR API as there could be other extensions on top of on top of Power extension point in OMR. The proper way to call this is via a connector, i.e. 

```cpp
OMR::CodeGeneratorConnector::inlineDirectCall(node, resultReg)
```

andrewcraik:
@hzongaro can you look at the comments from @fjeremic and update the PR?

runtime/compiler/p/codegen/J9TreeEvaluator.cpp
fjeremic: We still need a fix here. Once it's done this should be ready.

hzongaro: Thanks, Filip.  Revised in change https://github.com/eclipse/openj9/pull/3086/commits/545a95fa85e2d80a2492b2bc3bd60e955da9806a

hzongaro:
I believe all review comments have been addressed, so I will remove "WIP" from the title.

fjeremic:
Jenkins test plinux,aix JDK8

fjeremic:
Jenkins test sanity plinux,aix JDK8

andrewcraik:
@fjeremic were you going to merge this or do we need to get another opinion before you are ready to merge it?

gita-omr:
I would like to take a look before merge.

fjeremic:
> @fjeremic were you going to merge this or do we need to get another opinion before you are ready to merge it?

I'll delegate to @gita-omr given the above comment.

gita-omr:
To follow up on our discussion @hzongaro @fjeremic @0dvictor adding an ASSERT in the evaluator (inlineDirectCall || !supportsNonHelper) would be slightly misleading since supportsNonHelper (at least from its name) does not imply that the helper has to be inlined.
It's not the first place where we have some kind of handshake between IL gen, optimizer, and codegen and things need to be kept manually in sync. But I think the more accurate the name of the query is the better. If we call it supportsNonHelper let's assume that codegen will handle it one way or another (maybe by inlining or maybe by generating a call to something else). If we want to indicate that this nonHelper definitely needs to be inlined, then let's call it inlinesNonHelper().

Also, I think checking for 64bit should be done inside supports/inlinesNonHelper().

PR #3060, opened 2018-09-27
Potential deadlock for calling GarbageCollectorMXBean.getLastGcInfo()

LinHu2016:
- poolNames(Java String array) are constant after jvm is initialized,
retrieve them from Java side instead of constructing it in JNI.
 - change the memoryPool order in internal MemoryPoolList(Heap memory
 Pools first) in order to matching the memorypools order in GcInfo.
 - use jniGetPrimitiveArrayCritical to avoid Potential deadlock
  condition.

 fix: #3057

Signed-off-by: Lin Hu <linhu@ca.ibm.com>

PR #3090, opened 2018-10-01
Remove unused stringClass variable

fjeremic:
JDK11 test run with -Werror and this variable is set but not used and
thus causing a build break.


fjeremic:
Regression introduced in #3060.

gacholio:
Merging without test, as this is obviously correct.

PR #3091, opened 2018-10-01
Handle error thrown by nestmate visibility check

fengxue-IS:
`checkVisibility` used to only return error code, Nestmates imposed
extra checks on the nesthost class which may throw errors.

This change addresses potential incorrect exception handling that
sets wrong type of exception during method lookup.

### DIFF
--- a/runtime/vm/lookupmethod.c
+++ b/runtime/vm/lookupmethod.c
@@ -379,6 +379,12 @@ javaResolveInterfaceMethods(J9VMThread *currentThread, J9Class *targetClass, J9R
 			BOOLEAN shouldAdd = TRUE;
 			J9Class* newMethodClass = NULL;
 			newMethod = processMethod(currentThread, lookupOptions, foundMethod, interfaceClass, &data->exception, &data->exceptionClass, &data->errorType, nameAndSig, senderClass, targetClass);
+
+			/* Exit loop if exception have been directly set since direct exception must be from access control (Nestmates) */
+			if (NULL != currentThread->currentException) {
+				return NULL;
+			}
+
 			if (NULL == newMethod) {
 				iTable = iTable->next;
 				continue;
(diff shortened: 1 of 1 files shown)


fengxue-IS:
@gacholio can you please take a look?

runtime/vm/lookupmethod.c
```diff
@@ -379,7 +379,9 @@ javaResolveInterfaceMethods(J9VMThread *currentThread, J9Class *targetClass, J9R
 			BOOLEAN shouldAdd = TRUE;
 			J9Class* newMethodClass = NULL;
 			newMethod = processMethod(currentThread, lookupOptions, foundMethod, interfaceClass, &data->exception, &data->exceptionClass, &data->errorType, nameAndSig, senderClass, targetClass);
-			if (NULL == newMethod) {
+			if (NULL != currentThread->currentException) {
+				return NULL;
```
gacholio: Can this one be a goto like the others rather than an inline return?

gacholio:
Has this behaviour been verified against the RI?

runtime/vm/lookupmethod.c
```diff
@@ -379,7 +379,9 @@ javaResolveInterfaceMethods(J9VMThread *currentThread, J9Class *targetClass, J9R
...
+			} else if (NULL == newMethod) {
```
gacholio: No real need for elseif if the previous returns or goto's.

runtime/vm/lookupmethod.c
fengxue-IS: this function uses inline return consistently, would you like me to change all usage to `goto end`?

fengxue-IS:
This fixed a JDK11 nestmates test failure to match RI behaviour

gacholio:

PR #3096, opened 2018-10-02
Update InvokeDynamic Error wrapping

fengxue-IS:


fengxue-IS:
copying #3022 to include in v0.10.0 rc2
@pshipton FYI

PR #3098, opened 2018-10-02
Improve option parsing

keithc-ca:
The primary change is to do option pattern matching just once. Instead of:
```
if ( $argv =~ /^\-\-graphSpecs=/ ) {
    ($graphSpecs) = $argv =~ /^\-\-graphSpecs=(.*)/;
}   
```
we can use:
        
```
if ( $argv =~ /^--graphSpecs=(.*)/ ) { 
    $graphSpecs = $1;
} 
```
which ensures a proper option match and captures the option value at the same time.
Also, `-` doesn't need to be escaped in this context.


smlambert:
Tested via Grinder (internal due to issues with openj9 Jenkins server today).

PR #3099, opened 2018-10-02
Improve DDR support

### LINKED ISSUE #2507: Must run jdmpview from a 32-bit JVM to look at a 32-bit core file
Until it can be fixed, you must run jdmpview from a 32-bit JVM to look at a 32-bit core file. Similarly you must run jdmpview from a 64-bit JVM to look at a 64-bit core file.

keithc-ca:
DDR support in OpenJ9 makes use of the `ddrgen` tool from OMR which operates on debug information generated by compilers.

Not all compilers retain references to typedefs. For example a field involving type `UDATA` may instead be recorded as using `uint32_t` or `uint64_t` depending on the definition of `uintptr_t`. This means that we cannot rely on generating DDR pointer classes at build time because they would reflect the type definitions in that build which would be incompatible with core files produced by a VM with a different pointer size.

This makes the necessary foundation changes to allow DDR pointer classes to be created dynamically according to the core file being examined. Small changes will be required in the extension repositories to adopt this new capability.

doc issue https://github.com/eclipse/openj9-docs/issues/97
Fixes #2507


pshipton:
@keithc-ca please create an issue at https://github.com/eclipse/openj9-docs in order to coordinate updating the release notes

debugtools/DDR_VM/src/com/ibm/j9ddr/vm29/j9/OptInfo.java
```diff
@@ -220,15 +221,15 @@ private static String getOption(J9ROMClassPointer romClass, long option) throws
 	}
 
 	private static VoidPointer getStructure(J9ROMClassPointer romClass, long option) throws CorruptDataException {
-		SelfRelativePointer ptr = getSRPPtr(romClass.optionalInfo(), romClass.optionalFlags(), option);
+		SelfRelativePointer ptr = getSRPPtr(J9ROMClassHelper.optionalInfo(romClass), romClass.optionalFlags(), option);
```
pshipton: What happens if there is code outside this project using optionalInfo(), cpShapeDescription()? jdmpview supports adding debug extensions at runtime via the plugins command. There are many extensions that exist outside of this project.

debugtools/DDR_VM/src/com/ibm/j9ddr/BytecodeGenerator.java
```diff
@@ -21,95 +21,1724 @@
...
+
+	/* Sample generated code:
+	 *
+	 * public ObjectMonitorReferencePointer objectMonitorLookupCacheEA() throws CorruptDataException {
+	 *     return ObjectMonitorReferencePointer.cast(nonNullFieldEA(J9VMThread._objectMonitorLookupCacheOffset_));
+	 * }
+	 */
+	private void doArrayMethod(FieldDescriptor field) {
+		String fieldType = field.getType();
+		String componentType = fieldType.substring(0, fieldType.lastIndexOf('[')).trim();
```
pshipton: Assumes arrays are always single dimension.

debugtools/DDR_VM/src/com/ibm/j9ddr/vm29/j9/OptInfo.java
keithc-ca: I don't think there's a way to avoid that problem - it's a consequence of the behavior of ddrgen (being unable to distinguish among `U32Pointer`, `U64Pointer` and `UDATAPointer`).

debugtools/DDR_VM/src/com/ibm/j9ddr/BytecodeGenerator.java
keithc-ca: No. If the field type is `Whatever[4][5]` it will analyze `Whatever[4]` and notice the component type has type `TYPE_ARRAY`. This is the same as the behavior found in `PointerGenerator.java`.

debugtools/DDR_VM/src/com/ibm/j9ddr/vm29/j9/OptInfo.java
pshipton: What will happen?

runtime/ddr/types.cpp
```diff
@@ -0,0 +1,41 @@
...
+ *******************************************************************************/
+
+#include "j9.h"
+#include "j9port_generated.h"
+#include "j9generated.h"
+#include "j9nongenerated.h"
+#include "ddrhelp.h"
+
+/*
+ * This struct has fields correspnding to StructureTypeManager.TYPE_XXX categories
```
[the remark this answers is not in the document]
keithc-ca: Perhaps the fact that this file is in a folder called `ddr` provides that context?

debugtools/DDR_VM/src/com/ibm/j9ddr/vm29/j9/OptInfo.java
pshipton: Keith created https://github.com/eclipse/openj9/issues/3131 to deal with this.

pshipton:
jenkins test all win,zlinux jdk8

pshipton:
@keithc-ca note the test failures

keithc-ca:
I see two failures of the form
```
[FindVMOutputParser] [ERROR] ParserUtil: missing line J9WSRP(struct J9PoolPuddleList) puddleList
```
where in the new output the `struct` tag is omitted: `J9WSRP(J9PoolPuddleList)`: I think the test should be adjusted to match either format.

keithc-ca:
The newest commit fixes the test failure.

pshipton:
jenkins test sanity win jdk8

pshipton:
jenkins test extended win jdk8

pshipton:
jenkins test all zlinux jdk8

pshipton:
jenkins test all win,zlinux jdk8

pshipton:
jenkins test sanity win,zlinux jdk8

pshipton:
jenkins test sanity win,zlinux jdk8

pshipton:
jenkins test extended win,zlinux jdk8

pshipton:
jenkins test all win,zlinux jdk8

pshipton:
Looks like the win jdk8 sanity failure is a jenkins issue and all the testing passed.

PR #3059, opened 2018-09-27
Add test evironment setup and teardown

renfeiw:
- add testEnv.mk
- if TEST_FLAG is JITAAS, startup and kill jitaasserver server

Signed-off-by: Renfei Wang <renfeiw@ca.ibm.com>

Issue #2893, opened 2018-09-17
Testing for out-of-process JIT compilation (aka JITaaS)


harryyu1994:
While it’s true that JITaaS can run the same set of tests that OpenJ9 runs, we are going to need two “java” instances running instead of one. (One act as the server and one act as the client)

For example:
We need to first launch the server like this:
./java -XX:JITaaSServer &

Then runs the OpenJ9 tests with this extra option.
make _sanity EXTRA_OPTIONS=” -XX:JITaaSClient ”

Could this setup be done? 

@llxia @smlambert @vsebe @mpirvu

llxia:
I think we can have a variable, say `PRE_TEST_CMD` set in https://github.com/AdoptOpenJDK/openjdk-tests/blob/f887cbaeaecfecc2dcf8f8b644c41b58cff9328a/maketest.sh#L19

By default, `PRE_TEST_CMD` is empty. If JITaaS build, Jenkins build/local run can pass in the cmd it needs run before running any test.

fyi @renfeiw

harryyu1994:
Also I think we need to somehow kill/clean up the server after the test run is over.

vsebe:
Depends on https://github.com/AdoptOpenJDK/openjdk-infrastructure/issues/493

harryyu1994:
Hi, I'm wondering roughly when this will be completed. Thanks!

llxia:
For test part, if this is urgent, please feel free to modify [maketest.sh](https://github.com/AdoptOpenJDK/openjdk-tests/blob/f887cbaeaecfecc2dcf8f8b644c41b58cff9328a/maketest.sh) as needed.

llxia:
I saw the PR in https://github.com/AdoptOpenJDK/openjdk-tests/pull/534. Do we plan to run tests both in docker and directly on machines?

harryyu1994:
Running tests directly on machines are more important to us. Testing in docker is a nice-to-have.

harryyu1994:
[...]

PR #2836, opened 2018-09-11
Convert PullRequests to pipeline of jobs

AdamBrousseau:
- PRs currently run in separate, stand-alone jobs
  and this is not maintenance friendly.
- There are many benefits to switching to the common
  workflow of "pipeline of jobs" used by all other builds.
- This PR adds support for PR builds to use
  Pipeline-Build-Test-All as a top-level job.
- This also adds support in pipeline-functions to accomodate
  various functionality needed by PRs.
- New features:
   - Github commit status updates before and after a Compile or Test job
   - Auto-Cancel Test jobs when new ones are launched from the same PR
   - New Parameter BULD_IDENTIFIER which is useful for determining
     what a build belongs to (Nightly, OMR, PR#123 etc)

Issue #2728
[skip ci]

Signed-off-by: Adam Brousseau <adam.brousseau88@gmail.com>

Issue #2728, opened 2018-08-28
Convert PullRequest builds to be "pipeline of jobs"


AdamBrousseau:
Related #1450 

Currently our PR builds execute in a single job. We use our OpenJ9 build AND test pipeline scripts in order to complete the build. This is inefficient because it is a different process than how all our other builds are now working. In order to complete #1450 and remove our test pipeline code, we need to convert the PR builds to use the [Pipeline-Build-Test-Any-Platform](https://github.com/eclipse/openj9/blob/master/buildenv/jenkins/jobs/pipelines/Pipeline-Build-Test-Any-Platform). This will also allow for other PR work to proceed. There are two parts in order to get this working
1. Reconfig the PR jobs to use the above file instead of the matching PR files
1a. Add the PLATFORM, SDK_VERSION and TESTS_TARGETS (not TEST_TARGET or TEST_TARGETS) parameters to each PR job
1. Modify the pipeline files ([1](https://github.com/eclipse/openj9/blob/master/buildenv/jenkins/common/pipeline-functions#L218)) to pass the PR related parameters we need in the build.
   - ghprbPullId
   - ghprbGhRepository
   - ghprbCommentBody
   - ghprbTargetBranch

/cc @llxia @jdekonin @vsebe

AdamBrousseau:
Summarizing here since the PR is pretty messy at this point. I have booked a meeting next week to discuss this. Hopefully we can get this resolved soon.

#### Advantages
- Use Adopt Test Pipelines
   - No longer maintain our own test pipeline scripts
- Align with the same build & test story as our nightly, omr and personal builds.
- Allows for SVT and other testing in PRs (JitaaS etc)
- Reduced master disk usage. See **Concerns**.
- Reduced releng maintenance
   - No complicated regexs
   - Less builds to duplicate each new java release
   - No dup test pipelines to maintain/debug
- Unblocks future work
   - Stale pipeline code cleanup #2753
   - Grinder-style PR builds (no compiles) #2859
   - Move Jenkins params to pipeline code for PRs #2744
- Reduced overall machine time
   - 1 Compile per build rather than per test

#### Disadvantages
- Overall build time will likely increase in most scenarios.
   - Machines will be less busy, so when the farm is loaded builds may get through faster. But when the farm is idle the build time will be increased by the amount of time to transfer the SDK(s) from the compile machine to the Jenkins master and then back down to the test machine.
[...]

PR #3101, opened 2018-10-02
Parse test target for jitaas flag

AdamBrousseau:
- Adds ability to pass a TESTS_TARGETS with '+jitaas' in the name.
- Sets flag TEST_FLAG=JITAAS and passes to Test job

Issue #2893
Related AdoptOpenJDK/openjdk-tests#584, #3059
[skip ci]


AdamBrousseau:
This is what we have come up with for a proposed way to handle turning on jitaas testing via either a PR or regular build.
In a regular build, you would add `.jitaas` to any test target you wanted to run. In a PR build you would basically do the same. For example `Jenkins test sanity.functional.jitaas xlinux jdk11`. We still call down to the same sanity.functional job but with the parameter `TEST_FLAG=JITAAS`. This allows you to run both jitaas tests and non jitaas tests in the same PR command. Ex `Jenkins test sanity.functional,sanity.functional.jitaas xlinux jdk11`

Please note: This will not work via PRs until #2728 is resolved
cc @smlambert @renfeiw @jdekonin @DanHeidinga @pshipton @harryyu1994

smlambert:
As discussed with @AdamBrousseau the caveat of this approach is that committers/users may be confused and believe that sanity.functional.jitaas is an actual test target that they could run, which it is most definitely not.  

This approach does overcome the barrier of how to format and retrieve input from a git comment to trigger PR builds, and given our short window to enable this testing, we should likely proceed with it.  

I do not want to see us continue to overload the test targets via the PR build plugin everytime a new feature request comes in.  It will be good for us to consciously keep the PR inputs aligned with the actual manual/commandline input for testing, otherwise make it a maintenance/doc nightmare to explain the mappings introduced over time.  We see that in the internal/proprietary system we are still pulling ourselves out of, there are far too many cases of "we mapped this to this to this to this..." to the point where we no longer even understand how that evolution/hackery ever occurred.

AdamBrousseau:
The alternative would be something like `Jenkins test sanity.functional xlinux jdk11 TEST_FLAG=JITAAS depends ....`
Perhaps we should update the `depends` syntax to be `DEPENDS=eclipse/omr#1,ibmruntimes/openj9-openjdk-jdk11#2` since the code currently takes everything after depends as a space separated list of dependencies. You could probably argue that PLATFORMS, VERSIONS and TESTS_TARGETS should be used this way as well. That way order no longer matters.

pshipton:
A slight modification would be sanity.functional+jitaas which maybe keeps the parsing easy for faster delivery, but is more explainable, less confusing, to users trying to figure out the test target.

And then agree on a longer term strategy such as suggested in https://github.com/eclipse/openj9/pull/3101#issuecomment-426377412, although I'm not a fan of having to specify PLATFORMS=, VERSIONS=, etc.

smlambert:
sanity.functional+jitaas  
that is a good compromise... ship it!

smlambert:
Jenkins test sanity+jitaas xlinux jdk11

AdamBrousseau:
That syntax would work if this and #2836 was merged ;)

harryyu1994:
sorry forgot to ask this earlier, has the JVM EXTRA_OPTION "-XX:JITaaSClient " already been dealt with as well?

AdamBrousseau:
Good question. I don't see where it would have been added yet. I thought it would have been in #3059 or AdoptOpenJDK/openjdk-tests#584 but it isn't there. I do however see [this](https://github.com/AdoptOpenJDK/openjdk-tests/blob/76cea9961f7c627afa1ebf97eee74504c228e242/buildenv/jenkins/JenkinsfileBase#L59). So I may have to make another change to also pass `EXTRA_OPTIONS` or `JVM_OPTIONS` to the test jobs. @smlambert or @renfeiw please confirm.

smlambert:
I don't think this was clearly communicated in the initial request (and I have not seen a reply to my question, how are customers expected to invoke this feature)...

But anyway, there number of ways to do this.  Testing with additional cmdline options, can be done by introducing new variants in playlist.xml file.  I sense the expectation here is that you assume we will apply this commandline option across the board to all tests (that all tests should be able to be run with jitaas).  

If so, then we can leverage EXTRA_OPTIONS env var and set the 
export EXTRA_OPTIONS="-XX:JITaaSClient"

(JVM_OPTIONS replaces existing options which is not desirable in this case, EXTRA_OPTIONS appends additional commandline options).

smlambert:
@renfeiw - when we setup jitaas server, can we also append -XX:JITaaSClient ?

AdamBrousseau:
Ie. Set `EXTRA_OPTIONS` [here](https://github.com/eclipse/openj9/blob/master/test/TestConfig/testEnv.mk#L30)

smlambert:
@AdamBrousseau - thanks.  We know where to set it, I was asking if @renfeiw can prep a PR.  You do not need to help chase down test solutions, I know your plate is full in releng/build space.

renfeiw:
I can add `EXTRA_OPTIONS=-XX:JITaaSClient` in the testEnv.mk, but my concern is this hides the EXTRA_OPTIONS away. It will break if a user explicitly exports another EXTRA_OPTIONS and forget to append `-XX:JITaaSClient` in the export. My understanding is, when a user wants to locally run the tests with jitaas, he/she has to provide TEST_FLAG and EXTRA_OPTIONS manually. In the build environment, the EXTRA_OPTIONS can be set along with TEST_FLAG when `+jitaas` exists. Another way is to introduce a second option JITAAS_OPTIONS, and define it in the testEnv.mk.

smlambert:
yes, we can proceed with the separate JITAAS_OPTIONS approach.

AdamBrousseau:
Can we not do something like `EXTRA_OPTIONS+=...` in the makefile?

renfeiw:
Yes, we can. It's the same idea with the approach `JITAAS_OPTIONS`, instead of directly appending it in the mkgen.pl, we append it in the EXTRA_OPTIONS. I guess we need to use `override EXTRA_OPTIONS+=...`, as otherwise the cmdline argument will override this makefile assignment. I would prefer the `JITAAS_OPTIONS`, as `EXTRA_OPTIONS` was designed for user's extra option only.

AdamBrousseau:
I guess my concern is from the build side, we're already passing TEST_FLAG=JITAAS, I'd rather not pass another, somewhat redundant, option.

renfeiw:
You don't need to pass another variable. I will define `JITAAS_OPTIONS` in the testEnv.mk, and use it in the mkgen.pl.

PR #3095, opened 2018-10-02
Update link to get ASM-7.0 from Beta release page

r30shah:
Fixes: #2995

Signed-off-by: Rahil Shah <rahil@ca.ibm.com>

PR #3103, opened 2018-10-02
Fix failures in Condy Test

r30shah:
When GC kicks in when constant dynamic object is resolved, VM helper code expects all the registers are stored on Java Stack and while resolving frames it takes this adjustment into account while building frames. Condy Test was failing as it was not storing registers on Java Stack producing corrupted stack frame.

Changes originate from #3089 and #3095


r30shah:
@pshipton @irinarada @fjeremic PR for Condy and asm link fix.

andrewcraik:
@r30shah why is there an unrelated commit about ASM 7.0 in this PR?

Issue #3008, opened 2018-09-25
JTReg test: sun/security/util/FilePermCompat/CompatImpact.java fails with OpenJ9


groeges:
The `./jdk/sun/security/util/FilePermCompat/CompatImpact.java` test fails when running with the latest OpenJ9/OpenJDK 11 build. This same test passes when using Hotspot/OpenJDK11 build.

The issue seems to be related to the java option `-Djdk.security.filePermCompat=true`

The main error that is generated is:
```
java.security.AccessControlException: Access denied ("java.io.FilePermission" "f" "read")
```
The full JTReg output for the failure is:
```
command: main -Djdk.security.filePermCompat=true CompatImpact mine
reason: User specified action: run main/othervm -Djdk.security.filePermCompat=true CompatImpact mine 
Mode: othervm [/othervm specified]
Additional options from @modules: --add-modules java.base --add-exports java.base/jdk.internal.misc=ALL-UNNAMED
elapsed time (seconds): 0.746
configuration:
Boot Layer
  add modules: java.base                   
  add exports: java.base/jdk.internal.misc ALL-UNNAMED

STDOUT:
[...]

PR #3104, opened 2018-10-02
Use ProtectionDomain.impliesWithAltFilePerm()

JasonFengJ9:
Use `ProtectionDomain.impliesWithAltFilePerm()`

Uses `j.s.ProtectionDomain.impliesWithAltFilePerm()` to ensure `FilePermission` compatibility after [JDK-8164705](https://bugs.java.com/view_bug.do?bug_id=JDK-8164705).
This change only applies to `Java 9` and onward, decorate it with current level `Java11`. 

Manually verified the failed testcase (#3008) passes with this PR.

close: #3008 

Reviewer: @pshipton 
FYI: @DanHeidinga 

### DIFF
--- a/jcl/src/java.base/share/classes/java/security/AccessControlContext.java
+++ b/jcl/src/java.base/share/classes/java/security/AccessControlContext.java
@@ -521,7 +521,11 @@ static int checkPermWithCachedPDsImplied(Permission perm, Object[] toCheck, Acce
 					}
 				}
 			}
+			/*[IF Sidecar19-SE-OpenJ9]*/
+			if (!((ProtectionDomain) domain).impliesWithAltFilePerm(perm)) {
+			/*[ELSE]*/
 			if (!((ProtectionDomain) domain).implies(perm)) {
+			/*[ENDIF] Sidecar19-SE-OpenJ9*/
 				return i; // NOT implied
 			}
 		}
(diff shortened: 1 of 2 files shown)


jcl/src/java.base/share/classes/java/security/AccessControlContext.java
```diff
@@ -521,7 +521,11 @@ static int checkPermWithCachedPDsImplied(Permission perm, Object[] toCheck, Acce
 					}
 				}
 			}
+			/*[IF Java11]*/
```
pshipton: Please use Java 9 instead of 11, no reason not to fix these releases as well. Although Java 9 is out of service its still somewhat alive for cmake and macOS atm. Also there will be one more build of Java 10 in the 0.11.0 release.

JasonFengJ9: Updated JPP flag to `Sidecar19-SE-OpenJ9`. Manually verified that `pConfig` still compiles with updated `ProtectionDomain` class.
Please have another look.

pshipton:
jenkins test all zlinux jck11

pshipton:
jenkins test sanity zlinux jck11

pshipton:
The other implies() in AccessControlContext was removed from the commit?

JasonFengJ9:
There was a `Permission.implies` changed to `Permission. impliesWithAltFilePerm` which wasn't intended. Somehow it escaped the sanity check.
Current changes are verified via personal builds.

pshipton:
jenkins test all zlinux jdk11

pshipton:
jenkins test extended zlinux jdk11

PR #3105, opened 2018-10-02
Remove obsolete Panama build spec

keithc-ca:


DanHeidinga:
fyi @tajila

DanHeidinga:
It would have been nice to give @tajila a chance to weigh in on this as he's the primary user of the Panama spec...

keithc-ca:
We did discuss this with @tajila.

DanHeidinga:
Glad to hear it!  When there's no evidence of a discussion (even a +1 somewhere) or a reference to an issue about how future panama testing will be handled, this looks abrupt.

PR #3106, opened 2018-10-02
Updated Dockerfile to copy known_hosts file

cwesMills:
Dockerfile will copy known_hosts file provided by user, into the docker container
The purpose of this is to prevent an error while cloning from github.


cwesMills:
@AdamBrousseau

smlambert:
I am guessing you can use ci skip for these PRs, to avoid triggering travis build.

PR #3109, opened 2018-10-02
Fix inverted FSD test in DTFJ

gacholio:
The logic for determining if FSD is enabled is backwards.

[ci skip]


gacholio:
This really doesn't require even basic testing. It's a single-character change that's obviously correct syntax, and the feature itself is not tested.

DanHeidinga:
Merging without a build as it's obviously correct

PR #3112, opened 2018-10-03
Style update of jitReferenceArrayCopy on X86

0dvictor:
Update the code style to be consistent with nearby code.

### DIFF
--- a/runtime/codert_vm/xnathelp.m4
+++ b/runtime/codert_vm/xnathelp.m4
@@ -969,29 +969,17 @@ END_PROC(jitDecompileOnReturnJ)
 
 START_PROC(jitReferenceArrayCopy)
 	FASTCALL_EXTERN(impl_jitReferenceArrayCopy,2)
-	pop uword ptr J9TR_VMThread_jitReturnAddress[_rbp]
 	SWITCH_TO_C_STACK
-	SAVE_ALL_REGS
-
-	mov    uword ptr [_rbp+J9TR_VMThread_floatTemp3], _rsi
-	mov    uword ptr [_rbp+J9TR_VMThread_floatTemp4], _rdi
-ifdef({ASM_J9VM_ENV_DATA64},{
-ifdef({WIN32},{
-	mov    _rdx, _rcx
-	mov    _rcx, _rbp
-},{
-	mov    _rsi, _rcx
-	mov    _rdi, _rbp
-})
-},{
-	mov    _rdx, _rcx
-	mov    _rcx, _rbp
-})
-	call   FASTCALL_SYMBOL(impl_jitReferenceArrayCopy,2)
-	test   _rax, _rax
-	RESTORE_ALL_REGS
+	SAVE_C_VOLATILE_REGS
+	mov uword ptr J9TR_VMThread_floatTemp3[_rbp],_rsi
+	mov uword ptr J9TR_VMThread_floatTemp4[_rbp],_rdi
+	mov PARM_REG(2),_rcx
+	mov PARM_REG(1),_rbp
+	call FASTCALL_SYMBOL(impl_jitReferenceArrayCopy,2)
+	dnl set ZF if succeed
+	test _rax,_rax
+	RESTORE_C_VOLATILE_REGS
 	SWITCH_TO_JAVA_STACK
-	push uword ptr J9TR_VMThread_jitReturnAddress[_rbp]
 	ret
 END_PROC(jitReferenceArrayCopy)
 
--- a/runtime/oti/xhelpers.m4
+++ b/runtime/oti/xhelpers.m4
@@ -205,6 +205,8 @@ dnl C linkage (windows and linux)
 dnl	argument GPRs: none (stdcall) / ECX EDX (fastcall)
 dnl	preserved: EBX EDI ESI, no XMM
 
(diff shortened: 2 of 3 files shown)


0dvictor:
@gacholio Could you take a look at this change? It updates the code style to match nearby code based on your comments from #3075.

runtime/codert_vm/xnathelp.m4
```diff
@@ -972,23 +972,12 @@ START_PROC(jitReferenceArrayCopy)
...
-	mov    _rcx, _rbp
-})
-	call   FASTCALL_SYMBOL(impl_jitReferenceArrayCopy,2)
-	test   _rax, _rax
+	mov uword ptr J9TR_VMThread_floatTemp3[_rbp], _rsi
+	mov uword ptr J9TR_VMThread_floatTemp4[_rbp], _rdi
+	mov PARM_REG(2),_rcx
+	mov PARM_REG(1),_rbp
+	call FASTCALL_SYMBOL(impl_jitReferenceArrayCopy,2)
+	test _rax, _rax
```
gacholio: Space after comma again (I really need to get over this!).

runtime/oti/xhelpers.m4
```diff
@@ -205,6 +205,8 @@ dnl C linkage (windows and linux)
 dnl	argument GPRs: none (stdcall) / ECX EDX (fastcall)
 dnl	preserved: EBX EDI ESI, no XMM
 
+define({PARM_REG},{ifelse($1,1,_rcx,$1,2,_rdx,{ERROR})})
```
gacholio: Thanks for saving me having to do this.

gacholio:
The other change I'd like to see here is in cnathelp.cpp - the indenting of the arguments to referenceArrayCopy is extremely unreadable in git (and not great in eclipse). In general, matching indents can only be done if there are not characters preceding, so it you were to move the currentThread argument onto the next line indented, the code would read a lot better.

I'd also like

```
    ))
	{
```

to be

```
    )) {
```

gacholio:
Thank you for putting up with my quirks.

runtime/codert_vm/xnathelp.m4
0dvictor: Updated. Not a problem at all. I actually like your idea of keeping consistent style inside one file.

0dvictor:
I've updated `cnathelp.cpp` as well. Do you mind take a look once more?

runtime/codert_vm/xnathelp.m4
```diff
@@ -972,23 +972,12 @@ START_PROC(jitReferenceArrayCopy)
...
-})
-},{
-	mov    _rdx, _rcx
-	mov    _rcx, _rbp
-})
-	call   FASTCALL_SYMBOL(impl_jitReferenceArrayCopy,2)
-	test   _rax, _rax
+	mov uword ptr J9TR_VMThread_floatTemp3[_rbp],_rsi
+	mov uword ptr J9TR_VMThread_floatTemp4[_rbp],_rdi
+	mov PARM_REG(2),_rcx
```
gacholio: This is great, but I'm a little concerned that since we don't know here what arg1 and arg2 are, it's possible we might be smashing registers unintentionally (I have no good solution for this).

As with the readBarrier code, you can probably get away with only saving C volatile registers here, not all registers. Also, I don't think you need the return address pop/push.

gacholio:
In cnathelp.cpp, you can use JIT_HELPER_PROLOGUE without the SLOW, and remove the epilogiue macro (I don't have a bare macro that yet, though I should). The SLOW variants save/restore the JIT return address which should not be necessary in this case.

gacholio:
The indentation of the C call arguments still needs to be fixed (move currentThread to new line to avoid overindent).

0dvictor:
Updated. Could you take a look again?

runtime/codert_vm/xnathelp.m4
```diff
@@ -972,23 +972,12 @@ START_PROC(jitReferenceArrayCopy)
...
+	mov PARM_REG(1),_rbp
+	call FASTCALL_SYMBOL(impl_jitReferenceArrayCopy,2)
+	test _rax,_rax
```
gacholio: I think this test may have been copied from another helper - it doesn't seem to serve any purpose.

runtime/codert_vm/xnathelp.m4
gacholio: See above re: C volatile regs and return address.

runtime/codert_vm/xnathelp.m4
0dvictor: The test is to set ZF if reference array copy failed so that JIT code can throw an array store exception.

gacholio: I see - a little fragile since it assumes the macros will not affect the flags. Could you please add a comment here?

runtime/codert_vm/xnathelp.m4
0dvictor: Good suggestion, updated.

runtime/codert_vm/xnathelp.m4
0dvictor: Done.

gacholio:
jenkins test sanity xlinux,win jdk8

PR #3116, opened 2018-10-03
Inlining changes for MethodHandle

liqunl:
- Remove inlining heuristics for MethodHandle.asType because VP now can
  fold the whole call to a constant with known object constraint.
- Always fold Unsafe methods to reveal the native Unsafe methods.

### DIFF
--- a/jcl/src/java.base/share/classes/java/lang/invoke/MethodHandle.java
+++ b/jcl/src/java.base/share/classes/java/lang/invoke/MethodHandle.java
@@ -240,13 +240,6 @@ public abstract class MethodHandle {
 	// Until the JIT can synthesize calls to virtual methods, we must synthesize calls to these static ones instead
 	/*[ENDIF]*/
 	private static MethodHandle asType(MethodHandle mh, MethodType newType) {
-		/*
-		 * JIT can easily propagate type information and fold the if when it can prove early return always happen.
-		 * The early return also saves the JIT from having to inline full asType call
-		 */
-		if (mh.type == newType) {
-			return mh;
-		}
 		return mh.asType(newType);
 	}
 	/*[IF Sidecar19-SE]*/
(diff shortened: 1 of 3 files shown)


jcl/src/java.base/share/classes/java/lang/invoke/MethodHandle.java
```diff
@@ -240,13 +240,6 @@ MethodHandle cloneWithNewType(MethodType newType) {
 	// Until the JIT can synthesize calls to virtual methods, we must synthesize calls to these static ones instead
 	/*[ENDIF]*/
 	private static MethodHandle asType(MethodHandle mh, MethodType newType) {
-		/*
```
andrewcraik: Why is this being deleted?

andrewcraik:
I don't see the HCR guard change - where is it?

andrewcraik:
@0dvictor I assume the Unsafe inlining changes aren't going to affect the chances you need for CAS support under concurrent scavenge?

0dvictor:
> I assume the Unsafe inlining changes aren't going to affect the chances you need for CAS support under concurrent scavenge?

No, I don't think this change will affect the concurrent scavenge one or vise versa.

jcl/src/java.base/share/classes/java/lang/invoke/MethodHandle.java
liqunl: That code was added because we wanted to inine `asType` and let VP fold the comparison to a goto. We don't need it now because VP recognize and fold the outer method.

liqunl:
>I don't see the HCR guard change - where is it?

Sorry, just found that we already fold methods of DirectHandle and PrimitiveHandle.

jcl/src/java.base/share/classes/java/lang/invoke/MethodHandle.java
andrewcraik: Is the VM ever running this code and having that if was saving us from creating new handles or was it only ever called from the JIT?

liqunl: No, having that won't save us from creating new handles. The intention for this code was to aid the optimization in the JIT. But now the JIT can do the same without the early return. The other MethodHandle.asType calls to MethodType.equals which does the same check.

andrewcraik:

Issue #1017, opened 2018-01-24
Jenkins log and artifact cleanup schedule


AdamBrousseau:
Today we took down Eclipse's [hipp5](https://ci.eclipse.org/) which is where our OpenJ9 Jenkins master lives. We were consuming 264G space according to the [Bugzilla](https://bugs.eclipse.org/bugs/show_bug.cgi?id=530259) and have been asked to reduce.

What we store:
- Build logs (console output)
- SDKs from Builds and PRs

Jenkins has the following options, which are configured on a per job basis
```
Days to keep builds, if not empty, build records are only kept up to this number of days    
Max # of builds to keep, if not empty, only up to this number of build records are kept    
Days to keep artifacts, if not empty, artifacts from builds older than this number of days will be deleted, but the logs, history, reports, etc for the build will be kept    
Max # of builds to keep with artifacts, if not empty, only up to this number of builds have their artifacts retained
```
Currently, our nightly (and OMR acceptance) builds are setup with the following:
- Days to keep builds: 60
- Max # of builds to keep: 100
- Days to keep artifacts:
- Max # of builds to keep with artifacts: 10

Our PR builds are not configured to discard (keep everything).

#### Propose:
Nightlies and OMR Acceptance:
- Days to keep builds: 60
- Max # of builds to keep: 100
- Days to keep artifacts: ~7~
- Max # of builds to keep with artifacts: ~10~

PRs:
- Days to keep builds: 60
- Max # of builds to keep: 100
- Days to keep artifacts: ~14~
- Max # of builds to keep with artifacts: ~25~

Please review @pshipton 

Also, I have a script to check disk space I could add a job for and send a Slack notification if we're over X. Could open as a separate issue if we're interested.

pshipton:
I'm ok with this, or with smaller numbers even. Do you know about how many builds consume the 264G?

Perhaps others will have comments.

pshipton:
I checked a full Java 9 build and it was 4G uncompressed. This includes the git repos, all the source, build artifacts, and images. I assume the SDK kept around must be smaller than this. The jdk image itself is 323M and zipped its 202M, so around 20G for 100 builds.

pshipton:
We probably want a bigger number for PRs, "Max # of builds to keep with artifacts", say 100 to match the "Max # of builds to keep".

AdamBrousseau:
Biggest offenders currently are the PR Sanity builds. (no cleanuo yet)
688 SDKs being stored across all jobs.
[...]

PR #3117, opened 2018-10-03
WIP: Remove build artifacts in pipeline after tests finish

AdamBrousseau:
- Delete all Build artifacts after the overall pipeline has completed
- Always remove artifacts even if tests fail

Related #1017
[skip ci]


AdamBrousseau:
A hack way to remove the SDKs after the tests finish. You would only be able to access them from the build jobs when the tests are running. Would currently save us 35G on master.
Thoughts @DanHeidinga @pshipton @smlambert @jdekonin ?

AdamBrousseau:
I could also wrap this in a parameter so we have the option to turn it off at launch time or via variables config file.

AdamBrousseau:
In addition. We configure the build discarder via pipeline code for Build jobs.

AdamBrousseau:
Closing this for now as I don't believe it is desired behaviour.

PR #3119, opened 2018-10-03
Change heuristics regarding AOT compilations

mpirvu:
With the current heuristics only methods that have been downgraded from warm
to cold optimization level are eligible for AOT compilations. Moreover,
methods that are deemed "important for start-up" (string/zip/hash) are not
downgraded and therefore not AOTed.

This commit changes the heuristics such that the methods "important for
start-up" become subject to the downgrading policy (like any other method)
and therefore can be AOTed. However, when generating an AOT body for such
an "important method for start-up" the optimization level is raised back
to warm.

Performance experiments with Liberty Applicatioon Server show a 5-6% start-up
time improvement from this change.

### DIFF
--- a/runtime/compiler/control/CompilationThread.cpp
+++ b/runtime/compiler/control/CompilationThread.cpp
@@ -501,10 +501,7 @@ bool TR::CompilationInfo::shouldDowngradeCompReq(TR_MethodToBeCompiled *entry)
                  !TR::Options::getCmdLineOptions()->getOption(TR_DisableDowngradeToColdOnVMPhaseStartup))
                )
                {
-#if !defined(J9ZOS390)  // disable for zOS because we don't want to increase CPU time
-               if (!importantMethodForStartup(method))
-#endif
-                  doDowngrade = true;
+               doDowngrade = true;
                }
             // Downgrade if RI based recompilation is enabled
             else if (persistentInfo->isRuntimeInstrumentationRecompilationEnabled() && // RI and RI Recompilation is functional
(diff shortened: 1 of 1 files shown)


mpirvu:
@ymanton could you please review this commit? Thanks

runtime/compiler/control/CompilationThread.cpp
```diff
@@ -501,10 +501,7 @@ bool TR::CompilationInfo::shouldDowngradeCompReq(TR_MethodToBeCompiled *entry)
                  !TR::Options::getCmdLineOptions()->getOption(TR_DisableDowngradeToColdOnVMPhaseStartup))
                )
                {
-#if !defined(J9ZOS390)  // disable for zOS because we don't want to increase CPU time
-               if (!importantMethodForStartup(method))
-#endif
```
ymanton: Does removing this mean that on zOS we'll be compiling "important for start-up" methods at AOT-warm now, where as before they were compiled at AOT-cold? Since the comment for this `#ifdef` says it's purpose is to reduce CPU consumption perhaps we need someone to clarify exactly what was desired. This patch should result in less CPU consumption when the important for start-up methods are in the cache, but more if they have to be compiled since they're now at a higher opt level.

Is zOS OK with this change or do we really want to maintain the slightly different behaviour for it? Personally I'd prefer it if heuristics were the same on all platforms, it makes life easier for everyone in the long run, but I'm not qualified to comment on zOS issues.

If "tweaks" like this need to exist they should be guarded with an option rather than an `#ifdef` since there is really nothing architecture-specific about the heuristic or the tweak and it would make it easier to know that the tweak even exists in the first place.

Any comments @fjeremic and/or @joransiu?

mpirvu: Traditionally on zOS customers care a lot about CPU burnt by compilation threads; this is was the reason or the existing exclusion. With this PR there will be more CPU burnt in the cold run, but not in the warm run.
Note that we are not talking about huge amounts of CPU since "important method for startup" only targets "small" String/zip/hash methods and only during start-up.

Agreed about eliminating the #ifdef and this commit actually does so making the heuristic more uniform across platforms.

fjeremic: I'm ok with the change given the small scope on the classes affected.

mpirvu:
Jenkins test sanity.functional all jdk8

mpirvu:
Jenkins test sanity all jdk8

PR #3120, opened 2018-10-03
Create remote calls to get constant strings from resolved method

dmitry-ten:
Make remote calls for `isUnresolvedString` and `constantString`.
This is needed to correctly determine whether a String sym reference
is resolved. Previously, `isUnresolvedString` was always returning true,
which caused all corresponding sym references to be marked as
unresolved. This caused some blocks to be incorrectly marked as cold,
which prevented inlinining and caused a throughput regression.


dmitry-ten:
@mpirvu Could you review, please?

PR #3122, opened 2018-10-03
Read Barrier aware ref array GC copy helper

amicic:
Introduce optimized variant of reference array GC copy helper for
Concurrent Scavenger. It will execute RB for every single ref slot of
the source arrays (and as for regular Gencon config will just do post
batch WB).

JITed code in H/W based CS was not relying on this, but now that we are
introducing S/W variant (https://github.com/eclipse/openj9/issues/3054)
this may be called more often.

### DIFF
--- a/runtime/gc_modron_standard/StandardAccessBarrier.cpp
+++ b/runtime/gc_modron_standard/StandardAccessBarrier.cpp
@@ -603,6 +603,43 @@ MM_StandardAccessBarrier::getJNICriticalRegionCount(MM_GCExtensions *extensions)
 }
 
 #if defined(J9VM_GC_ARRAYLETS)
+#if defined(OMR_GC_CONCURRENT_SCAVENGER)
+I_32
+MM_StandardAccessBarrier::doCopyContiguousBackwardWithReadBarrier(J9VMThread *vmThread, J9IndexableObject *srcObject, J9IndexableObject *destObject, I_32 srcIndex, I_32 destIndex, I_32 lengthInSlots)
+{
+	srcIndex += lengthInSlots;
+	destIndex += lengthInSlots;
+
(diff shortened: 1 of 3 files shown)


amicic:

amicic:

dmitripivkine:

runtime/gc_modron_standard/StandardAccessBarrier.cpp
```diff
@@ -603,6 +603,43 @@ MM_StandardAccessBarrier::getJNICriticalRegionCount(MM_GCExtensions *extensions)
...
+{
+	srcIndex += lengthInSlots;
+	destIndex += lengthInSlots;
+
+	fj9object_t *srcSlot = (fj9object_t *)indexableEffectiveAddress(vmThread, srcObject, srcIndex, sizeof(fj9object_t));
+	fj9object_t *destSlot = (fj9object_t *)indexableEffectiveAddress(vmThread, destObject, destIndex, sizeof(fj9object_t));
+	fj9object_t *srcEndSlot = srcSlot - lengthInSlots;
+
+	while (srcSlot > srcEndSlot) {
+		preObjectRead(vmThread, (J9Object *)srcObject, srcSlot);
```
charliegracie: in the code below you are doing a pre decrement correct?  Shouldn't the preObjectRead also check the value at `srcSlot - 1`?

amicic: thanks for spotting this!

would this be too ugly (trying to save on one substract) ?

		preObjectRead(vmThread, (J9Object *)srcObject, --srcSlot);
		*--destSlot = *srcSlot;

amicic: or ?

	while (srcSlot-- > srcEndSlot) {
		preObjectRead(vmThread, (J9Object *)srcObject, srcSlot);
		*--destSlot = *srcSlot;
	}

amicic:

PR #3123, opened 2018-10-03
Eliminate remote calls for isWarmCallGraphTooBig

dmitry-ten:
Previously, calls to `isWarmCallGraphTooBig`
always made remote calls to the client, which used IProfiler to
determine
whether the callgraph is too big to be inlined from the bytecode entry.
However, since we are already caching IProfiler bytecode entries on the
server, we can just access the cached result.
We ensure that `isWarmCallGraphTooBig` always returns the correct value
by making `setWarmCallGraphTooBig` update the value in cached
entries, as well as on the client. This way, even if a current server
crashes, the new server will still receive correct values.

### DIFF
--- a/runtime/compiler/control/JITaaSCompilationThread.cpp
+++ b/runtime/compiler/control/JITaaSCompilationThread.cpp
@@ -1437,14 +1437,6 @@ bool handleServerMessage(JITaaS::J9ClientStream *client, TR_J9VM *fe)
          client->write(mirror->isInlineable(comp));
          }
          break;
-      case J9ServerMessageType::ResolvedMethod_isWarmCallGraphTooBig:
-         {
-         auto recv = client->getRecvData<TR_ResolvedJ9Method *, uint32_t>();
-         TR_ResolvedJ9Method *mirror = std::get<0>(recv);
-         uint32_t bcIndex = std::get<1>(recv);
-         client->write(mirror->isWarmCallGraphTooBig(bcIndex, comp));
-         }
-         break;
       case J9ServerMessageType::ResolvedMethod_setWarmCallGraphTooBig:
          {
          auto recv = client->getRecvData<TR_ResolvedJ9Method *, uint32_t>();
--- a/runtime/compiler/env/j9methodServer.cpp
+++ b/runtime/compiler/env/j9methodServer.cpp
@@ -1041,16 +1041,12 @@ TR_ResolvedJ9JITaaSServerMethod::isInlineable(TR::Compilation *comp)
       }
    }
 
-bool
-TR_ResolvedJ9JITaaSServerMethod::isWarmCallGraphTooBig(uint32_t bcIndex, TR::Compilation *comp)
-   {
-   _stream->write(JITaaS::J9ServerMessageType::ResolvedMethod_isWarmCallGraphTooBig, _remoteMirror, bcIndex);
-   return std::get<0>(_stream->read<bool>()); 
-   }
-
 void
 TR_ResolvedJ9JITaaSServerMethod::setWarmCallGraphTooBig(uint32_t bcIndex, TR::Compilation *comp)
    {
+   // set the variable in the server IProfiler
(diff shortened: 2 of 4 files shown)


runtime/compiler/rpc/protos/compile.proto
```diff
@@ -91,7 +91,7 @@ enum J9ServerMessageType
    ResolvedMethod_isBigDecimalConvertersMethod = 137;
    ResolvedMethod_osrFrameSize = 138;
    ResolvedMethod_isInlineable = 139;
-   ResolvedMethod_isWarmCallGraphTooBig = 140;
+   // ResolvedMethod_isWarmCallGraphTooBig = 140;
```
mpirvu: Same as above, we may still need this message when caching is disabled

runtime/compiler/control/JITaaSCompilationThread.cpp
```diff
@@ -1437,14 +1437,6 @@ bool handleServerMessage(JITaaS::J9ClientStream *client, TR_J9VM *fe)
          client->write(mirror->isInlineable(comp));
          }
          break;
-      case J9ServerMessageType::ResolvedMethod_isWarmCallGraphTooBig:
```
mpirvu: I think we still need this code, just in case caching is not used

runtime/compiler/env/j9methodServer.cpp
```diff
@@ -1041,16 +1041,12 @@ TR_ResolvedJ9JITaaSServerMethod::isInlineable(TR::Compilation *comp)
       }
    }
 
-bool
```
mpirvu: We may still needs this code when caching is disabled. 
When caching is enabled we need to check the cache. Note that the structure of the cache at the server is different than the structure of the IProfiler at the client, thus you cannot simply call `TR_ResolvedJ9Method::isWarmCallGraphTooBig`

runtime/compiler/env/j9methodServer.cpp
```diff
@@ -1041,16 +1041,12 @@ TR_ResolvedJ9JITaaSServerMethod::isInlineable(TR::Compilation *comp)
...
-   {
-   _stream->write(JITaaS::J9ServerMessageType::ResolvedMethod_isWarmCallGraphTooBig, _remoteMirror, bcIndex);
-   return std::get<0>(_stream->read<bool>()); 
-   }
-
 void
 TR_ResolvedJ9JITaaSServerMethod::setWarmCallGraphTooBig(uint32_t bcIndex, TR::Compilation *comp)
    {
+   // set the variable in the server IProfiler
+   TR_ResolvedJ9Method::setWarmCallGraphTooBig(bcIndex, comp);
```
mpirvu: By calling `TR_ResolvedJ9Method::setWarmCallGraphTooBig` the server will try to set the flag into its own **global** IProfiler which is the wrong thing. The server needs to access the per client cache )which has a different structure than the global IProfiler hashtable.

dmitry-ten: `setWarmCallGraphTooBig` sets the flag in the bytecode entry returned by `profilingSample`.
I thought that calling `profilingSample` on the server already returns an entry from the per-client cache. Wouldn't this make calling `TR_ResolvedJ9Method::setWarmCallGraphTooBig` ok?

mpirvu: You are right. I missed that `setWarmCallGraphTooBig` use the overridden `profilingSample` which knows how to access the information per client

runtime/compiler/env/j9methodServer.cpp
dmitry-ten: Same thing as below, since it is calling `profilingSample`, it would automatically fetch an entries from the client/server as needed.

mpirvu: correct

dmitry-ten:
So, is there anything that needs to be done for this PR, other than testing?

mpirvu:
All is good. Please test it and let me know of the outcome.

PR #3129, opened 2018-10-03
Only set Artifactory env variables if they are not null

AdamBrousseau:
[skip ci]
Fixes #3114


smlambert:
Both failed checks are due to outofmemory on proxy-worker, looking at copyright and line-ending myself, they are good, so will merge.

PR #3130, opened 2018-10-03
Check for software read barrier before initializing guarded storage

fjeremic:
In case we're running on z14+ hardware but we still want to run with
software read barriers we need an extra check before initializing
guarded storage on the particular thread.

### DIFF
--- a/runtime/vm/guardedstorage.c
+++ b/runtime/vm/guardedstorage.c
@@ -50,10 +50,13 @@ j9gs_initializeThread(J9VMThread *vmThread)
 		success = 1;
 	} else {
 		BOOLEAN supportsGuardedStorageFacility = FALSE;
+		J9JavaVM *vm = vmThread->javaVM;
+		J9MemoryManagerFunctions *mmFuncs = vm->memoryManagerFunctions;
 		J9ProcessorDesc  processorDesc;
 		j9sysinfo_get_processor_description(&processorDesc);
 		if (j9sysinfo_processor_has_feature(&processorDesc, J9PORT_S390_FEATURE_GUARDED_STORAGE) &&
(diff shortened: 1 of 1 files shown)


dmitripivkine:
@amicic FYI

dmitripivkine:
I believe we should figure out how we are planning to use -XX:gc=softwareEvacuateReadBarrier option.
Is there any reason to run CS on z14 and NOT use HW support?

fjeremic:
@amicic could I please get a review and you assigned as the committer for this one? The issue here is that with `-XXgc:softwareEvacuateReadBarrier` we don't initialize the GS handler address with the software handler version and we still go through the hardware path which executes a `BIC` instruction which is incorrect if we're running in a z13 (we'll get a SIGILL).

fjeremic:
> Is there any reason to run CS on z14 and NOT use HW support?

Yes, for testing and if the Guarded Storage facility is turned off at the OS level (or support is not available).

dmitripivkine:
ok, but why we need to call j9gs_initializeThread() for case if softwareEvacuateReadBarrier is enabled? I believe any j9gs_* helpers should not be called at all.

dmitripivkine:
If it is a temporary fix to solve immediate problem we can do it

fjeremic:
> If it is a temporary fix to solve immediate problem we can do it

Yes this is a temporary fix to allow us to even run with software read barriers. We cannot even run it on z14 today because the GS handler address gets initialized to the wrong thing! I'm in favor of the refactoring however given we now have a CS solution for x86. We should converge on the same implementation on the GC side, as well as the JIT of course.

dmitripivkine:

fjeremic:

dmitripivkine:
Yes, we need to unify GC/VM side of code for CS for all platforms for software read barriers path. We are planning to start work on it as soon as x86 part will be in

fjeremic:
@dmitripivkine I think this one is good to go.

Issue #3050, opened 2018-09-27
Test-extended.system-JDK10-linux_x86-64_cmprssptrs SharedClasses.SCM01.MultiThread_0 assertion CallSnippet.cpp:143: j2iThunk != NULL


pshipton:
https://ci.eclipse.org/openj9/job/Test-extended.system-JDK10-linux_x86-64_cmprssptrs/48

```
MT5 stderr Assertion failed at /home/jenkins/workspace/Build-JDK10-linux_x86-64_cmprssptrs/build/linux-x86_64-normal-server-release/vm/compiler/../compiler/x/codegen/CallSnippet.cpp:143: j2iThunk != NULL
MT5 stderr VMState: 0x0005ff0a
MT5 stderr 	null virtual j2i thunk
MT5 stderr compiling java/util/Objects.equals(Ljava/lang/Object;Ljava/lang/Object;)Z at level: cold
```

irinarada:
@andrewcraik - can we look at this?

irinarada:
Thanks @liqunl for looking at this!

andrewcraik:
FYI @jdmpapin

liqunl:
@pshipton Where can I download the test?

liqunl:
Found the test in https://github.com/eclipse/openj9-systemtest/tree/master/openj9.test.sharedClasses

pshipton:
Given the failure rate, it seems this can be deferred to the next release.

pshipton:
Another occurrence https://ci.eclipse.org/openj9/job/Test-extended.system-JDK10-linux_x86-64_cmprssptrs/53

pshipton:
Fixed by #3135

PR #3135, opened 2018-10-03
Skip getJ2IThunk() when writing x86 PIC data

jdmpapin:
X86PicDataSnippet already has its J2I thunk in _thunkAddress, so there
is no need to use getJ2IThunk() to find it.

This also sidesteps a problem whereby getJ2IThunk() would (somehow)
sometimes fail to find the thunk even though it was already generated,
particularly in ahead of time compilations.


jdmpapin:
I expect this to fix #3050, although I haven't been able to repro

andrewcraik:

PR #3138, opened 2018-10-04
Add wrtbar/wrtbari opcodes

cathyzhyi:
The xwrtbar/xwrtbari opcodes are added to represent both the store
as well as its side effects like checking object address against
tenure address range for GC, notifying the VM of the store or
other similar activities.

This PR adds the most basic supports for write barrier opcodes in openj9.
The exist wrtbar/wrtbari are change to awrtbar/awrtbari for consistency.
By default write barriers use store handlers and evaluators.

issue: #2948

### DIFF
--- a/runtime/compiler/codegen/CodeGenGPU.cpp
+++ b/runtime/compiler/codegen/CodeGenGPU.cpp
@@ -66,13 +66,13 @@ static const char * nvvmOpCodeNames[] =
    "load",          // TR::bload
    "load",          // TR::sload
    "load",          // TR::lload
-   "rdbar",         // TR::irdbar
-   "rdbar",         // TR::frdbar
-   "rdbar",         // TR::drdbar
-   "rdbar",         // TR::ardbar
-   "rdbar",         // TR::brdbar
-   "rdbar",         // TR::srdbar
-   "rdbar",         // TR::lload
+   NULL,            // TR::irdbar
+   NULL,            // TR::frdbar
+   NULL,            // TR::drdbar
+   NULL,            // TR::ardbar
+   NULL,            // TR::brdbar
+   NULL,            // TR::srdbar
+   NULL,            // TR::lrdbar
    "load",          // TR::iloadi
    "load",          // TR::floadi
    "load",          // TR::dloadi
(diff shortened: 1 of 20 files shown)


runtime/compiler/codegen/CodeGenGPU.cpp
```diff
@@ -93,19 +93,33 @@ static const char * nvvmOpCodeNames[] =
    "store",          // TR::fstore
    "store",          // TR::dstore
    "store",          // TR::astore
-
-   NULL,          // TR::wrtbar
-
    "store",          // TR::bstore
    "store",          // TR::sstore
+
+   "wrtbar",         // TR::iwrtbar
```
andrewcraik: why are these not printing their proper names? making them all print the same seems a bad idea

andrewcraik:
@cathyzhyi please fix the copyrights

andrewcraik:
The basic change is pretty mechanical and looks fine - I'll wait to review further until the OMR change has been reviewed and agreed.

AdamBrousseau:
Jenkins copyright check

gita-omr:
why are these not printing their proper names? making them all print the same seems a bad idea

These are not for Debug. These are NNVM IR names

runtime/compiler/codegen/CodeGenGPU.cpp
cathyzhyi: Thanks for pointing this out. changed to NULL according to Gita's comments. https://github.com/eclipse/openj9/pull/3138#issuecomment-427132679

AdamBrousseau:
Jenkins copyright check

AdamBrousseau:
Jenkins line endings check

andrewcraik:
@cathyzhyi please force push to restart testing.

andrewcraik:
Once OMR design review completes we should coordinate the delivery of these two changes. FYI @fjeremic FYI @vijaysun-omr since this is the OpenJ9 part of the wrtbar opcode additions

fjeremic:
Jenkins test sanity all JDK8 depends eclipse/omr#3027

PR #3139, opened 2018-10-04
Change heuristics for activating compilation threads

mpirvu:
Currently the logic for activating compilation threads does not work as
intended when the number of compilation threads is specified by the user.
Ideally, when the user sets the number of compilation threads with
`-XcompilationThreads<n>` option, the JVM should activate new threads up to
'n' based on its internal activation thresholds. However, currently the JIT
will activate only (numProc-1) compilation threads (regardless of how many
the user specified) and the extra compilation threads could be activated
only if 'starvation' of the compilation threads is detected.

This commit will change the logic as follows:
1) If the user specifies the number of compilation threads the JIT will
assume that the user knows best and will activate the compilation threads
based on its internal activation threasholds irrespective of the number
of CPUs the JVM is allowed to run on.

2) If the user does not specify the number of compilation threads, the JVM
will activate up to numProc-1 compilation threads based on its internal
activation thresholds. Moreover, if more compilation threads are present
and starvation of compilation is detected (this could happen in linux)
then the additional existent compilation threads are activated based on
a more conservative activation policy.

### DIFF
--- a/runtime/compiler/control/CompilationThread.cpp
+++ b/runtime/compiler/control/CompilationThread.cpp
@@ -388,22 +388,23 @@ TR_YesNoMaybe TR::CompilationInfo::shouldActivateNewCompThread()
    // determined based on the number of CPUs, then the upper bound of comp threads is _numTargetCPUs-1
    // However, if the compilation threads are starved (on Linux) we may want
    // to activate additional comp threads irrespective of the CPU entitlement
-   if (TR::Options::_useCPUsToDetermineMaxNumberOfCompThreadsToActivate ||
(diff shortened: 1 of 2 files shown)


ymanton:
LGTM.

Jenkins test sanity all jdk8

ymanton:
Jenkins copyright check
Jenkins line endings check

AdamBrousseau:
Jenkins copyright check

AdamBrousseau:
Jenkins line endings check

mpirvu:
The two test failures on Windows are due to infra problems.

PR #3143, opened 2018-10-04
Prevent allocation of replaced classes

gacholio:
Objects on the heap must always point to the most current version of the
J9Class. Prevent allocation of replaced classes by causing fast path
allocations to fail for replaced classes. The final (slow path)
allocator will use the most current version of the requested class to
fix any callers who unwittingly passes a replaced class.

[ci skip]


gacholio:
jenkins test all zlinux jdk8

runtime/gc_base/JavaObjectAllocationModel.hpp
```diff
@@ -77,7 +77,7 @@ class MM_JavaObjectAllocationModel : public MM_AllocateInitialization
 		if (NULL != objectPtr) {
 			/* Initialize class pointer in object header -- preserve flags set by base class */
 			MM_GCExtensions *extensions = MM_GCExtensions::getExtensions(env);
-			extensions->objectModel.setObjectClass(objectPtr, _class);
+			extensions->objectModel.setObjectClass(objectPtr, getClass());
```
DanHeidinga: Should the protected `_class` field become private to prevent subclass's from reaching for it directly?  And likely commented to indicate the issue.

gacholio: Making the field private means modifying a lot of code. Any users of the field before a GC has occurred really don't need to get the current class (the OM is used in the faster path NoGC code as well).

gacholio: Tempted to add a new accessor getCurrentClass() for people who are aware that the class may change during a GC, and leave getClass() as it is (there are actually no callers - all access is by direct field fetch).

DanHeidinga: @dmitripivkine I'm OK with taking this change as is.  Do you want a followup issue to make the `_class` field private?

DanHeidinga:
Once the testing passes, whoever sees it first can merge

runtime/gc_base/JavaObjectAllocationModel.hpp
gacholio: I will not make the accessor change now - you can decide if you want it when changing the field visibility.

PR #2939, opened 2018-09-19
Enable DDR tests for AIX

mikezhang1234567890:
Enables tests for DDR on AIX, closes https://github.com/eclipse/openj9/issues/1769
- enable callsitedbgddrext, classesdbgddrext, shrcdbgddrext, and
  modularityddrtests cmdlinetests for AIX
- enable DDR_Test for AIX
- create new AIX and zOS targets in playlist.xml for the DDR cmdline
  tests and use correct arguments to sysvcleanup.pl script. Pass in zos
  option since zos and aix options are identical in sysvcleanup.pl
- limit heap size in DDR tests to 256MB, since on AIX the core file
  scales with heap size, which can hit filesystem limits

Signed-off-by: Mike Zhang <mike.h.zhang@ibm.com>

This PR depends on the following extensions repo PRs, since the DDR tests do not compile without DDR enabled on AIX:
https://github.com/ibmruntimes/openj9-openjdk-jdk8/pull/116
https://github.com/ibmruntimes/openj9-openjdk-jdk9/pull/178
https://github.com/ibmruntimes/openj9-openjdk-jdk10/pull/59
https://github.com/ibmruntimes/openj9-openjdk-jdk11/pull/23
https://github.com/ibmruntimes/openj9-openjdk-jdk/pull/35

Doc issue is https://github.com/eclipse/openj9-docs/issues/92

PR #3152, opened 2018-10-04
Add lib path env variable to find libomrsig on AIX

mikezhang1234567890:
Add LD_LIBRARY_PATH and LIBPATH env variables when running ddrgen due to its new dependency on libomrsig. On AIX, the search path is not saved in the executable when it is created since we do not have a direct equivalent to setting `rpath=$ORIGIN` as a compiler flag.

Signed-off-by: mikezhang <mike.h.zhang@ibm.com>

This PR should be merged before the extension repo PRs enabling DDR by default:
https://github.com/ibmruntimes/openj9-openjdk-jdk8/pull/116
https://github.com/ibmruntimes/openj9-openjdk-jdk9/pull/178
https://github.com/ibmruntimes/openj9-openjdk-jdk10/pull/59
https://github.com/ibmruntimes/openj9-openjdk-jdk11/pull/23
https://github.com/ibmruntimes/openj9-openjdk-jdk/pull/35

PR should also be tested with the extension repo PR since DDR isn't enabled by default on AIX without them.

### DIFF
--- a/runtime/ddr/run_omrddrgen.mk.ftl
+++ b/runtime/ddr/run_omrddrgen.mk.ftl
@@ -42,6 +42,13 @@ DDR_INPUT_FILES := $(addprefix $(TOP_DIR),$(foreach module,$(DDR_INPUT_MODULES),
 DDR_INPUT_FILES := $(addsuffix .dbg,$(DDR_INPUT_DEPENDS))
 </#if>
 
+# workaround to find libomrsig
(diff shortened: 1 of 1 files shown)


pshipton:
@mikezhang1234567890 please combine this into the same PR as #2939 (you can keep it as two commits) so I can PR test it properly.

mikezhang1234567890:
Alright, merging with that PR.

PR #3155, opened 2018-10-04
Fix jit crash in Symbol staticName

yanluo7:
Caller of Symbol::staticName typically assumes the underlying symbol
represents a static field. However, the condition that guards the
call currently only checks for a staic symbol, which is insufficient
as other types like Constant Dynamic. Method Handle, Method Type all
use static symbols. Use the correct check Symbol::isStaticField which
precludes the above mentioned non-static field types.


andrewcraik:

andrewcraik:

AdamBrousseau:

AdamBrousseau:
Jenkins copyright check

AdamBrousseau:
Jenkins line endings check

andrewcraik:
Jenkins test sanity win jdk8

PR #3157, opened 2018-10-05
Removed ccache install from Dockerfile

cwesMills:
Docker containers will not be running long enough to make good use of ccache, so it is being removed from the dockerfile.
[skip ci]

### DIFF
--- a/buildenv/jenkins/docker-slaves/s390x/ubuntu16/Dockerfile
+++ b/buildenv/jenkins/docker-slaves/s390x/ubuntu16/Dockerfile
@@ -45,7 +45,6 @@ RUN apt-get update \
     autoconf \
     build-essential \
     ca-certificates \
-    ccache \
     cmake \
     cpio \
     curl \


cwesMills:
@AdamBrousseau

keithc-ca:
Should this change be applied to all the other Dockerfiles?

cwesMills:
Ideally yes, the others should be changed.

cwesMills:
I've removed ccache from this dockerfile because the docker container won't be running long enough to make good use of it.

cwesMills:
@smlambert

smlambert:
Seems fine, btw, is there a way to test these dockerfiles?  A node label one calls to invoke one (that I could pass into a grinder)?

cwesMills:
This specific Dockerfile is set up on jenkins with labels: ci.project.openj9 hw.arch.s390x sw.os.linux sw.os.ubuntu sw.os.ubuntu.16

AdamBrousseau:
Unfortunately we don't have the resources at the moment to have these setup on the Eclipse OpenJ9 Jenkins instance. We also have yet to setup any kind of automation around creating the images and updating the "good" ones, nor is the ability to "test changes" via a PR style build. We have been prototyping this on a private Jenkins instance for now. Each docker cloud agent template has the same labels as the other existing machines. Because of this, the docker slaves will only spin-up when the existing machines are full. If you wish to guarantee execution on a docker slave simply add the label `hw.arch.docker` to your list of labels in your grinder job (eg. hw.arch.s390x && sw.os.ubuntu.16 && hw.arch.docker`). I believe the zlinux one has been working for a few weeks. The ub16 &cent7 on ppc64le are likely ready. I will check with Colton on Monday and get them updated with the standard labels once ready.

smlambert:
Great!  Looking forward to trying these out!  :)

PR #3159, opened 2018-10-05
Refactor transformDirectLoad

liqunl:
Pull the code of static final field folding out of transformDirectLoad
and have its own API.


liqunl:
Depends https://github.com/eclipse/omr/pull/3030

andrewcraik:
@liqunl the use of the TR_RestrictStaticFieldFolding option seems to be missing post refactor - was that intentional? I would think this option is one we would want to continue to use for now?

liqunl:
@andrewcraik Yes, it was intentional. I wanted to use this option for guarded final field folding. Since we have added another option for that, we can keep using this one for its original purpose, which is to enable folding without guards.

liqunl:
@andrewcraik The change this PR depends on has been merged. Could you continue the review?

andrewcraik:
So at a highlevel I'm not sure it makes sense to expose `foldStaticFinalField` as a public method. I am fine with the new `canFoldStaticFinalField` because that is just saying could we fold it. The objection is because `transformDirectLoad` has additional checking and book keeping which should always be done when folding and so calling `foldStaticFinalField` outside of the class doesn't make sense. I would, therefore,  like to see `foldStaticFinalField` being made protected, the doxygen reflecting it is an internal API and the proper public API getting doxygen to show people the right API to use rather than this internal API which lacks safety checks.

andrewcraik:
The functional parts of the refactoring look to be straightforward breakups of the original code from what I can see.

liqunl:
@andrewcraik `foldStaticFinalField` is public because it is to be called in field folding opt that I'm going to add. I don't think there's a good way to do the safety checks since the caller may like to fold fields that need OSR protection before the OSR protection is available. I think the caller is responsible for the safety checks or the caller has to make sure that subsequent optimizations will add protection for the folding.

runtime/compiler/optimizer/J9TransformUtil.cpp
```diff
@@ -1006,184 +1005,232 @@ J9::TransformUtil::transformIndirectLoad(TR::Compilation *comp, TR::Node *node)
    return NULL;
    }
 
+bool
+J9::TransformUtil::foldReliableStaticFinalField(TR::Compilation *comp, TR::Node *node)
+   {
+   TR_ASSERT(node->getOpCode().isLoadVarDirect() && node->isLoadOfStaticFinalField(),
```
andrewcraik: I'm pretty sure that in sat least some cases on z static final loads can be represented as indirect loads. Doesn't this assert need to account for that?

runtime/compiler/optimizer/J9TransformUtil.cpp
```diff
@@ -1006,184 +1005,232 @@ J9::TransformUtil::transformIndirectLoad(TR::Compilation *comp, TR::Node *node)
...
+
+   return false;
+   }
 
 bool
-J9::TransformUtil::transformDirectLoad(TR::Compilation *comp, TR::Node *node)
+J9::TransformUtil::foldStaticFinalFieldAssumingProtection(TR::Compilation *comp, TR::Node *node)
    {
-   TR_ASSERT(node->getOpCode().isLoadVarDirect(), "Expecting direct load; found %s %p", node->getOpCode().getName(), node);
+   TR_ASSERT(node->getOpCode().isLoadVarDirect() && node->isLoadOfStaticFinalField(),
```
andrewcraik: Ditto

runtime/compiler/optimizer/J9TransformUtil.cpp
andrewcraik: If they can still occur right now there needs to be a conditional return false rather than an assert until that deprecation is done.

liqunl: @andrewcraik Sure, I've updated the commit so that we return false instead of assert

andrewcraik:

PR #3162, opened 2018-10-05
Fold static final fields that are VarHandle

liqunl:
Fold static final fields that are VarHandle such that the calls to
VarHandle access methods can be inlined by methodHandleInvokeInlining
opts.


liqunl:

PR #3165, opened 2018-10-05
Implement vectorized String.indexOf() on Power

aviansie-ben:
An evaluator for the JITHelpers.intrinsicIndexOf* methods has been
implemented on Power that makes use of vector instructions to improve
the performance of String.indexOf().

### DIFF
--- a/runtime/compiler/p/codegen/J9TreeEvaluator.cpp
+++ b/runtime/compiler/p/codegen/J9TreeEvaluator.cpp
@@ -11797,6 +11797,313 @@ static TR::Register *inlineEncodeUTF16(TR::Node *node, TR::CodeGenerator *cg)
    return outputLenReg;
    }
 
+static TR::Register *inlineIntrinsicIndexOf(TR::Node *node, bool isLatin1, TR::CodeGenerator *cg)
+   {
+   auto vectorCompareOp = isLatin1 ? TR::InstOpCode::vcmpeubr : TR::InstOpCode::vcmpeuhr;
+
+   TR::Register *arrObjectAddress = cg->evaluate(node->getChild(1));
+   TR::Register *targetScalar = cg->evaluate(node->getChild(2));
+   TR::Register *startIndex = cg->evaluate(node->getChild(3));
+   TR::Register *endIndex = cg->evaluate(node->getChild(4));
+
+   TR::Register *cr0 = cg->allocateRegister(TR_CCR);
+   TR::Register *cr6 = cg->allocateRegister(TR_CCR);
+
+   TR::Register *zeroRegister = cg->allocateRegister();
+   TR::Register *result = cg->allocateRegister();
+   TR::Register *arrAddress = cg->allocateRegister();
+   TR::Register *currentAddress = cg->allocateRegister();
+   TR::Register *endAddress = cg->allocateRegister();
+
+   TR::Register *targetVector = cg->allocateRegister(TR_VRF);
+   TR::Register *targetVectorNot = cg->allocateRegister(TR_VRF);
+   TR::Register *searchVector = cg->allocateRegister(TR_VRF);
+   TR::Register *permuteVector = cg->allocateRegister(TR_VRF);
+
+   TR_PPCScratchRegisterManager *srm = cg->generateScratchRegisterManager();
+
+   TR::LabelSymbol *startLabel = generateLabelSymbol(cg);
+   TR::LabelSymbol *notSmallLabel = generateLabelSymbol(cg);
+   TR::LabelSymbol *vectorLoopLabel = generateLabelSymbol(cg);
+   TR::LabelSymbol *residueLabel = generateLabelSymbol(cg);
+   TR::LabelSymbol *notFoundLabel = generateLabelSymbol(cg);
+   TR::LabelSymbol *foundLabel = generateLabelSymbol(cg);
+   TR::LabelSymbol *endLabel = generateLabelSymbol(cg);
+
+   startLabel->setStartInternalControlFlow();
+   endLabel->setEndInternalControlFlow();
+
+   generateLabelInstruction(cg, TR::InstOpCode::label, node, startLabel);
+
+   // Special case for empty strings, which always return -1
+   generateTrg1Src2Instruction(cg, TR::InstOpCode::cmp4, node, cr0, startIndex, endIndex);
+   generateConditionalBranchInstruction(cg, TR::InstOpCode::beq, node, notFoundLabel, cr0);
+
+   generateTrg1ImmInstruction(cg, TR::InstOpCode::li, node, zeroRegister, 0);
+
+   // Calculate the actual addresses of the start and end points
+   generateTrg1Src1ImmInstruction(cg, TR::InstOpCode::addi, node, arrAddress, arrObjectAddress, TR::Compiler->om.contiguousArrayHeaderSizeInBytes());
+   generateTrg1Src2Instruction(cg, TR::InstOpCode::add, node, endAddress, arrAddress, endIndex);
+   generateTrg1Src2Instruction(cg, TR::InstOpCode::add, node, currentAddress, arrAddress, startIndex);
+
+   if (!isLatin1)
+      {
+      generateTrg1Src2Instruction(cg, TR::InstOpCode::add, node, endAddress, endAddress, endIndex);
+      generateTrg1Src2Instruction(cg, TR::InstOpCode::add, node, currentAddress, currentAddress, startIndex);
+      }
+
+   // Prefetch the string
+   generateMemInstruction(cg, TR::InstOpCode::dcbt, node, new (cg->trHeapMemory()) TR::MemoryReference(zeroRegister, currentAddress, 16, cg));
+
+   // Splat the value to be compared against and its bitwise complement into two vector registers
+   // for later use
+   generateTrg1Src1Instruction(cg, TR::InstOpCode::mtvsrwz, node, targetVector, targetScalar);
+   if (isLatin1)
+      generateTrg1Src1ImmInstruction(cg, TR::InstOpCode::vspltb, node, targetVector, targetVector, 7);
+   else
+      generateTrg1Src1ImmInstruction(cg, TR::InstOpCode::vsplth, node, targetVector, targetVector, 3);
+   generateTrg1Src2Instruction(cg, TR::InstOpCode::vnor, node, targetVectorNot, targetVector, targetVector);
+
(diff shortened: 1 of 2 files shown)


runtime/compiler/p/codegen/J9TreeEvaluator.cpp
```diff
@@ -11797,6 +11797,338 @@ static TR::Register *inlineEncodeUTF16(TR::Node *node, TR::CodeGenerator *cg)
...
+      deps->addPostCondition(zeroRegister, TR::RealRegister::NoReg);
+      deps->addPostCondition(result, TR::RealRegister::NoReg);
+      deps->addPostCondition(currentAddress, TR::RealRegister::NoReg);
+      deps->addPostCondition(endAddress, TR::RealRegister::NoReg);
+
+      deps->addPostCondition(targetVector, TR::RealRegister::NoReg);
+      deps->addPostCondition(targetVectorNot, TR::RealRegister::NoReg);
+      deps->addPostCondition(searchVector, TR::RealRegister::NoReg);
+      deps->addPostCondition(permuteVector, TR::RealRegister::NoReg);
+
```
zl-wang: 1) There are 4 registers in the dependency to be marked/set as excluding gr0, since they are sources of addi instructions. Something like:  deps->getPostConditions()->getRegisterDependency(index)->setExcludeGPR0();
     This will not cause any side effect, but guarantee the correctness, since you are using so many registers anyway (in case gr0 being free, it will be assigned to other registers not marked).

2) I am wondering if you can minimize the number of registers being used by checking if the children nodes' registers will not be reused.

3) I am looking at the code sequence to see whether I can spot something to account for the anomaly (7th-match vs. 8th-match).  I am suspecting short-latency VSX instructions (or/and  and  permute) being starved of the pipeline by long-latency instructions.  Not sure though.

aviansie-ben: 1. My bad, I meant to go through and mark registers as excluding gr0 as necessary before submitting but forgot. Will fix this.
2. I think it should be possible for me to check if `targetScalar`, `startIndex`, or `endIndex` won't be reused by checking the child refcount and donate them to the scratch register manager after use if that holds. I'm not sure whether I can do that for `arrObjectAddress` though, since it's marked as containing a collected reference and I don't know enough about how that works to say whether that would be safe or not.
3. As I've said before, the drop in speedup between 7th and 8th character matches has nothing to do with the new vectorized code becoming slower there and is instead caused by the old code unexpectedly becoming faster.

zl-wang: 2.  you can, since there is no GC point in that sequence (i.e  branch out to call another method, jitAsyncMessage or jitThrowException).  In addition, you are implicitly creating quite a few internal-pointers anyway (i.e. addresses pointing at the middle of object) ... that is un-safe, if they lived across GC points.

3. Ok, recall it now.  Make each match-point loop long enough ... and profile the two cases to see the tick distribution. That is the only way I can think of following it up.

aviansie-ben: 1 and 2 should be fixed now. As for 3, since the problem seems to be related to something about the scalar version, I think it's out of scope for this work and should be dealt with as an entirely separate matter (if it needs followup at all).

runtime/compiler/p/codegen/J9TreeEvaluator.cpp
```diff
@@ -11797,6 +11797,357 @@ static TR::Register *inlineEncodeUTF16(TR::Node *node, TR::CodeGenerator *cg)
...
+      deps->stopUsingDepRegs(cg, result);
+
+      cg->decReferenceCount(node->getChild(0));
+      cg->decReferenceCount(node->getChild(1));
+      cg->decReferenceCount(node->getChild(2));
+      cg->decReferenceCount(node->getChild(3));
+      cg->decReferenceCount(node->getChild(4));
+      }
+
+   node->setRegister(result);
```
fjeremic: FYI though it can't happen here because you allocated the register it's usually good practice to set the node's register _before_ calling `decReferenceCount` to avoid the situation where the `result` is obtained by calling `evaluate(...)` on one of `node`'s children. In such a scenario if you call `decReferenceCount` before `setRegister` you could end up killing a register because it's use count decrements to 0.

runtime/compiler/p/codegen/J9TreeEvaluator.cpp
zl-wang: One potential correctness issue:
  The same register(s) could be added into the dependency list multiple times.  The later code had better look like:
          if (....->referenceCount() > 1)
             {
             deps->addPostDependency() ...
             (setExcludeGPR0 if necessary)
             }

Another minor issue:
    I didn't see child(0) being evaluated.  I think child(0) is the String object itself(?).  If child(1) evaluation depends on the existence of child(0) result register, there might be a problem.  I am not sure though.

zl-wang: for the case of referenceCount() <= 1,  the register(s) are in srm.  They will be added to deps with excluding GPR0 later anyway.

aviansie-ben: Looked back at the code for `TR_PPCScratchRegisterManager` and it seems I got confused since `srm->stopUsingRegisters()` ignores donated registers but `srm->addScratchRegistersToDependencyList(...)` does not. I'll go ahead and fix that.

As for the 0th child, that child is simply a `JITHelpers` object on which the helper is being called. Its only reason for existing is to be checked against `null` to ensure that untrusted code cannot call these helpers unsafely since they can't get a reference to one.

aviansie-ben: Register dependencies should be fixed now.

zl-wang: Looks good ...

zl-wang:
Jenkins test sanity,extended aix,plinux jdk8,jdk11

fjeremic:
@zl-wang FYI [1]. I think we have the use `all` to get both sanity and extended.

[1] https://github.com/eclipse/openj9/tree/master/buildenv/jenkins

fjeremic:
@aviansie-ben we seem to have a crash in a JIT compiled method in the extended tests. Can you please take a look if it's related to this change? I'm also reassigning to @gita-omr since she self-requested a review and is more familiar with Power codegen.

https://ci.eclipse.org/openj9/job/PullRequest-Extended-JDK8-linux_ppc-64_cmprssptrs_le-OpenJ9/141/tapResults/

aviansie-ben:
Based on a quick look at the javacore, that failure definitely looks to be related to these changes. I'll take a closer look when I get a chance...

runtime/compiler/p/codegen/J9TreeEvaluator.cpp
```diff
@@ -11797,6 +11797,355 @@ static TR::Register *inlineEncodeUTF16(TR::Node *node, TR::CodeGenerator *cg)
...
+   // Then count the number of leading zeroes from the obtained result. This tells us the index (in
+   // bytes) of the first matching element in the vector. Note that there is no way to count leading
+   // zeroes of a half-word, so we count leading zeroes of a word and subtract 16 since the value
+   // we're interested in is in the least significant half-word.
+   generateTrg1Src1Instruction(cg, TR::InstOpCode::cntlzw, node, result, result);
+   generateTrg1Src1ImmInstruction(cg, TR::InstOpCode::addi, node, result, result, -16);
+
+   // Finally, combine this with the address of the last vector load to find the address of the
+   // first matching element in the string. Finally, use this to calculate the corresponding index.
+   generateTrg1Src2Instruction(cg, TR::InstOpCode::add, node, result, result, currentAddress);
```
zl-wang: When the match is found in the first vector, the byte-index should be (result - (arrAddress & 0xF)) disregarding currentAddress. (result + (currentAddress - arrAddress))  is only correct for matching in none first vector.  A simple fix is to add one instruction before this add:  currentAddress = currentAddress & ~(0xF);

aviansie-ben: I think you've misunderstood the part of the code responsible for loading the first misaligned vector. This is the correct way to calculate the address of the match because the data in `searchVector` has _already_ been shifted backwards by `currentAddress & 0xf` bytes earlier by the code responsible for replacing the garbage loaded before the string data. Because of this, the first byte of `searchVector` corresponds to the byte at `(currentAddress & ~0xf) + (currentAddress & 0xf) = currentAddress` _not_ the first byte that was actually loaded from memory (which would be at `currentAddress & ~0xf`).

zl-wang: Fair enough ...

aviansie-ben:
The previous test failure should be fixed now

aviansie-ben:
Not quite sure why the Travis build failed. From the logs, it seems like the build stalled before it even started... Can someone retry that for me?

gita-omr:
> Can someone retry that for me?

I am trying to figure out how to do it.

runtime/compiler/p/codegen/J9TreeEvaluator.cpp
```diff
@@ -11797,6 +11797,356 @@ static TR::Register *inlineEncodeUTF16(TR::Node *node, TR::CodeGenerator *cg)
...
+      srm->donateScratchRegister(arrObjectAddress);
+
+   // Handle the first character using a simple scalar compare. Otherwise, first character matches
+   // would be slower than the old scalar comparison loop. This is a problem since first character
+   // matches are common in certain contexts, e.g. StringTokenizer where the default first character
+   // to each String.indexOf call is ' ', which is the most common whitespace character.
+      {
+      TR::Register *value = srm->findOrCreateScratchRegister();
+
+      generateTrg1MemInstruction(cg, scalarLoadOp, node, value, new (cg->trHeapMemory()) TR::MemoryReference(arrAddress, result, isLatin1 ? 1 : 2, cg));
```
zl-wang: Please switch the base/index positions of arrAddress and result registers in this MemoryReference constructor, since arrAddress register is not set excludingGPR0 but result register is.

aviansie-ben: Fixed

zl-wang:
Jenkins test sanity,extended aix,plinux jdk8,jdk11

zl-wang:

gita-omr:
Since @zl-wang did the merge I will remove myself from assignee.

PR #3166, opened 2018-10-05
Deprecate ShrinkWrapping - Part I

fjeremic:
ShrinkWrapping is currently disabled on all platforms due to functional
issues. It has been disabled for several years and it's benefit is
questionable at best given the compile time costs.

This commit folds away some code related to ShrinkWrapping so as to not
cause build breaks when the optimization is removed from OMR. There
will be a subsequent "Part II" PR which will remove all the rest of the
code which is dependent on OMR.

Issue: https://github.com/eclipse/omr/issues/2107

### DIFF
--- a/runtime/compiler/p/codegen/PPCPrivateLinkage.cpp
+++ b/runtime/compiler/p/codegen/PPCPrivateLinkage.cpp
@@ -348,11 +348,6 @@ TR::PPCPrivateLinkage::PPCPrivateLinkage(TR::CodeGenerator *cg)
    _properties._computedCallTargetRegister  = TR::RealRegister::gr0; // gr11 = interface, gr12 = virtual, so we need something else for computed
    _properties._vtableIndexArgumentRegister = TR::RealRegister::gr12;
    _properties._j9methodArgumentRegister    = TR::RealRegister::gr3; // TODO:JSR292: Confirm
-
-   // shrink wrapping
-   setRegisterSaveSize(0);
-   _mapRegsToStack = (int32_t *) trMemory()->allocateHeapMemory(TR::RealRegister::NumRegisters*sizeof(int32_t));
-   memset(_mapRegsToStack, -1, (TR::RealRegister::NumRegisters*sizeof(int32_t)));
    }
 
 const TR::PPCLinkageProperties& TR::PPCPrivateLinkage::getProperties()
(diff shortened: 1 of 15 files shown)


fjeremic:
Preemptively launching builds here:

Jenkins test sanity all JDK8

runtime/compiler/p/codegen/PPCPrivateLinkage.cpp
```diff
@@ -1193,26 +1074,20 @@ void TR::PPCPrivateLinkage::createPrologue(TR::Instruction *cursor)
...
            TR::RealRegister::LastGPR - intSavedFirst <= 3))
          {
-         // shrink wrapping
-         TR_BitVector *p = cg()->getPreservedRegsInPrologue();
          for (regIndex=intSavedFirst; regIndex<=TR::RealRegister::LastGPR; regIndex=(TR::RealRegister::RegNum)((uint32_t)regIndex+1))
             {
-            if (!p || p->get((uint32_t)regIndex))
-               cursor = generateMemSrc1Instruction(cg(),TR::InstOpCode::Op_st, firstNode, new (trHeapMemory()) TR::MemoryReference(stackPtr, argSize, TR::Compiler->om.sizeofReferenceAddress(), cg()), machine->getPPCRealRegister(regIndex), cursor);
+            cursor = generateMemSrc1Instruction(cg(),TR::InstOpCode::Op_st, firstNode, new (trHeapMemory()) TR::MemoryReference(stackPtr, argSize, TR::Compiler->om.sizeofReferenceAddress(), cg()), machine->getPPCRealRegister(regIndex), cursor);
+
```
0xdaryl: This comment is not related to the changes in this PR but more an observation I'll make about the code that's already there and want to document it somewhere.

The use of `om.sizeofReferenceAddress()` in this register preserving code seems wrong to me.  This code should be preserving the full register size for the target bitness rather than asking the object model what the size of a reference is.  As luck (design?) would have it, on 64-bit this always returns '8' and on 32-bit returns '4' regardless of whether compressed refs is enabled or not.

There are some curiosities that need to be sorted out here (like what should `sizeOfReferenceAddress` be returning from the object model, what breaks if that changes).  I'll open a separate issue for that investigation.

fjeremic: @gita-omr FYI the above comment.

gita-omr: Thanks. Will wait for the issue that @0xdaryl is going to open.

PR #3173, opened 2018-10-05
Jitaas development messages

satbir-ca:
@mpirvu please review

### DIFF
--- a/runtime/compiler/control/JITaaSCompilationThread.cpp
+++ b/runtime/compiler/control/JITaaSCompilationThread.cpp
@@ -2717,9 +2717,10 @@ JITaaSHelpers::cacheRemoteROMClass(ClientSessionData *clientSessionData, J9Class
    bool classInitialized = std::get<9>(classInfo);
    uint32_t byteOffsetToLockword = std::get<10>(classInfo);
    TR_OpaqueClassBlock * leafComponentClass = std::get<11>(classInfo);
+   void *classLoader = std::get<12>(classInfo);
    clientSessionData->getROMClassMap().insert({ clazz, { romClass, methods,
       baseComponentClass, numDims,
-      nullptr, nullptr, parentClass, interfaces, classHasFinalFields, classDepthAndFlags, classInitialized, byteOffsetToLockword, leafComponentClass } });
+      nullptr, nullptr, parentClass, interfaces, classHasFinalFields, classDepthAndFlags, classInitialized, byteOffsetToLockword, leafComponentClass, classLoader} });
    uint32_t numMethods = romClass->romMethodCount;
    J9ROMMethod *romMethod = J9ROMCLASS_ROMMETHODS(romClass);
    for (uint32_t i = 0; i < numMethods; i++)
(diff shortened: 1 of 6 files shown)


runtime/compiler/control/JITaaSCompilationThread.cpp
```diff
@@ -2758,7 +2757,8 @@ JITaaSHelpers::packRemoteROMClassInfo(J9Class *clazz, TR_J9VM *fe, TR_Memory *tr
    bool classHasFinalFields = fe->hasFinalFieldsInClass((TR_OpaqueClassBlock *)clazz);
    uintptrj_t classDepthAndFlags = fe->getClassDepthAndFlagsValue((TR_OpaqueClassBlock *)clazz);
    bool inlineClass = fe->canAllocateInlineClass((TR_OpaqueClassBlock *)clazz);
-   return std::make_tuple(packROMClass(clazz->romClass, trMemory), methodsOfClass, baseClass, numDims, parentClass, TR::Compiler->cls.getITable((TR_OpaqueClassBlock *) clazz), methodTracingInfo, classHasFinalFields, classDepthAndFlags, inlineClass);
+   bool classInitialized =  fe->isClassInitialized((TR_OpaqueClassBlock *)clazz);
```
mpirvu: What do we need 'inlineClass' for?

runtime/compiler/env/VMJ9Server.cpp
```diff
@@ -514,6 +514,13 @@ TR_J9ServerVM::isPrimitiveClass(TR_OpaqueClassBlock * clazz)
 bool
 TR_J9ServerVM::isClassInitialized(TR_OpaqueClassBlock * clazz)
    {
+   OMR::CriticalSection getRemoteROMClass(_compInfoPT->getClientData()->getROMMapMonitor());
```
mpirvu: Critical section needs to be as small as possible

runtime/compiler/env/VMJ9Server.cpp
```diff
@@ -514,6 +514,13 @@ TR_J9ServerVM::isPrimitiveClass(TR_OpaqueClassBlock * clazz)
...
+   auto it = _compInfoPT->getClientData()->getROMClassMap().find((J9Class*) clazz);
```
mpirvu: If the class is not initialized we don't want to cache and return that answer forever. Instead a new message should be sent to the client in the hope that by now the class has been initialized. So the code would look something like:
```
check cache
if class is cached
   if class is initialized
      return true
    else
        send message
        cache answer
        return answer
else
    send message
    return answer
```
Make sure you don't hold the lock while sending the message

runtime/compiler/env/VMJ9Server.cpp
```diff
@@ -726,9 +733,17 @@ TR_J9ServerVM::canAllocateInlineClass(TR_OpaqueClassBlock *clazz)
    auto it = _compInfoPT->getClientData()->getROMClassMap().find((J9Class*) clazz);
    if (it != _compInfoPT->getClientData()->getROMClassMap().end())
       {
-      return it->second.inlineClass;
+      if (it->second.classInitialized)
```
mpirvu: Make critical section as small as possible.

runtime/compiler/env/VMJ9Server.cpp
```diff
@@ -726,9 +733,17 @@ TR_J9ServerVM::canAllocateInlineClass(TR_OpaqueClassBlock *clazz)
...
+         {
+         return (it->second.romClass->modifiers & (J9_JAVA_ABSTRACT | J9_JAVA_INTERFACE ));
+         }
+      else
+         {
+         JITaaS::J9ServerStream *stream = _compInfoPT->getMethodBeingCompiled()->_stream;
```
mpirvu: Here we also need to return the correct answer to the caller. If the class is initialized at this point we check againthe romClass->modifiers, otherwise we return false

mpirvu:
Is the commit called 'Reduce number of canAllocateInlineClass messages ' actually needed?
Also the commit 'Caching the class inline status' needs a correction in the title and there are a few typos in the commit message.

runtime/compiler/env/VMJ9Server.cpp
```diff
@@ -712,7 +776,36 @@ TR_J9ServerVM::isCloneable(TR_OpaqueClassBlock *clazzPointer)
...
+      OMR::CriticalSection getRemoteROMClass(_compInfoPT->getClientData()->getROMMapMonitor());
+      auto it = _compInfoPT->getClientData()->getROMClassMap().find((J9Class*) clazz);
+      if (it != _compInfoPT->getClientData()->getROMClassMap().end())
+         {
+            it->second.classInitialized = classInitialized;
+            return ((it->second.romClass->modifiers & (J9_JAVA_ABSTRACT | J9_JAVA_INTERFACE ))?
+                     true:false);
+         }
+      }
+
```
mpirvu: If the class is not initialized even when you've just asked, then we can return false. No need to send another message.

PR #3187, opened 2018-10-09
Implement vector String.indexOf(String)

NigelYiboYu:
Implement the String.indexOf(String) intrinsic for Java 8 and Java 9 JCL
for z13 and z14. This is able to detect the intrinsic methods and inline
them with a vector assembly sequence to accelerate indexOf(String) search.

For Java 8 JCL, refactor Java 8 indexOf(String) APIs and introduce
JITHelper intrinsic functions. These new JITHelper APIs will be recognized
as intrinsic and inlined with the aforementioned vector sequence.

### DIFF
--- a/jcl/src/java.base/share/classes/com/ibm/jit/JITHelpers.java
+++ b/jcl/src/java.base/share/classes/com/ibm/jit/JITHelpers.java
@@ -568,6 +568,97 @@ public final class JITHelpers {
 		}
 	}
 
+	/**
+	 * Searches in the source character array for the index of the target character array.
+	 * Both arrays must be Latin1 arrays consisting of compressible characters ranging from \u0000 to \u00FF.
+	 * The source and target character arrays have their lengths specified. The search for the array starts
+	 * at the specified offset and moves towards the end of source array.
+	 * 
+	 * @param s1Value
+	 * 		 	the source character array to serach in
+	 * @param s1len
+	 * 		    the length of the source array
+	 * @param s2Value
+	 * 			the target character array to search for
+	 * @param s2len
+	 * 			the length of the target array
+	 * @param start
+	 * 			the starting offset
+	 * 
+	 * @return the index of the target array inside the source array, or -1 if the target array is not found
+	 * */
+	public int intrinsicIndexOfStringLatin1(char[] s1Value, int s1len, char[] s2Value, int s2len, int start) {
+		char firstChar = byteToCharUnsigned(getByteFromArrayByIndex(s2Value, 0));
+
+		while (true) {
+			int i = -1;
+			if (start < s1len) {
+				i = intrinsicIndexOfLatin1(s1Value, (byte)firstChar, start, s1len);
+			}
+						
+			// Handles subCount > count || start >= count
+			if (i == -1 || s2len + i > s1len) {
+				return -1;
+			}
+
+			int o1 = i;
+			int o2 = 0;
+
+			while (++o2 < s2len && getByteFromArrayByIndex(s1Value, ++o1) == getByteFromArrayByIndex(s2Value, o2))
+				;
+
+			if (o2 == s2len) {
+				return i;
+			}
+
+			start = i + 1;
+		}
+	}
+	
+	/**
+	 * Searches in the source character array for the index of the target character array.
+	 * Both arrays consisting of UTF16 characters.
+	 * The source and target character arrays have their lengths specified. The search for the array starts
+	 * at the specified offset and moves towards the end of source array.
+	 * 
+	 * @param s1Value
+	 * 		 	the source character array to serach in
+	 * @param s1len
+	 * 		    the length of the source array
+	 * @param s2Value
+	 * 			the target character array to search for
+	 * @param s2len
+	 * 			the length of the target array
+	 * @param start
+	 * 			the starting offset
+	 * 
+	 * @return the index of the target array inside the source array, or -1 if the target array is not found
+	 * */
+	public int intrinsicIndexOfStringUTF16(char[] s1Value, int s1len, char[] s2Value, int s2len, int start) {
+		char firstChar = s2Value[0];
+
+		while (true) {
+			int i = intrinsicIndexOfUTF16(s1Value, firstChar, start, s1len);
+			
+			// Handles subCount > count || start >= count
+			if ((i == -1) || (s2len + i > s1len)) {
+				return -1;
+			}
+
+			int o1 = i;
+			int o2 = 0;
+
+			while ((++o2 < s2len) && (s1Value[++o1] == s2Value[o2]))
+				;
+
+			if (o2 == s2len) {
+				return i;
+			}
+
+			start = i + 1;
+		}
+	}
+	
 	public int intrinsicIndexOfLatin1(Object array, byte ch, int offset, int length) {
 		for (int i = offset; i < length; i++) {
 			if(getByteFromArrayByIndex(array, i) == ch) {
--- a/jcl/src/java.base/share/classes/java/lang/String.java
+++ b/jcl/src/java.base/share/classes/java/lang/String.java
@@ -2011,6 +2011,10 @@ public final class String implements Serializable, Comparable<String>, CharSeque
 	 * @see #lastIndexOf(String, int)
 	 */
 	public int indexOf(String subString, int start) {
+		if (subString.length() == 1) {
+			return indexOf(subString.charAtInternal(0), start);
+		}
+		
 		return indexOf(value, coder, lengthInternal(), subString, start);
 	}
 
@@ -6013,6 +6017,10 @@ public final class String implements Serializable, Comparable<String>, CharSeque
 	 * @see #lastIndexOf(String, int)
 	 */
 	public int indexOf(String subString, int start) {
+		if (subString.length() == 1) {
+			return indexOf(subString.charAtInternal(0), start);
+		}
+		
 		if (start < 0) {
 			start = 0;
 		}
@@ -6032,29 +6040,13 @@ public final class String implements Serializable, Comparable<String>, CharSeque
 			char[] s2Value = s2.value;
 
 			if (enableCompression && (null == compressionFlag || (s1.count | s2.count) >= 0)) {
-				char firstChar = helpers.byteToCharUnsigned(helpers.getByteFromArrayByIndex(s2Value, 0));
-
-				while (true) {
-					int i = indexOf(firstChar, start);
-
-					// Handles subCount > count || start >= count
-					if (i == -1 || s2len + i > s1len) {
-						return -1;
-					}
-
-					int o1 = i;
-					int o2 = 0;
-
-					while (++o2 < s2len && helpers.getByteFromArrayByIndex(s1Value, ++o1) == helpers.getByteFromArrayByIndex(s2Value, o2))
-						;
-
-					if (o2 == s2len) {
-						return i;
-					}
-
-					start = i + 1;
-				}
+				// both s1 and s2 are compressed
+				return helpers.intrinsicIndexOfStringLatin1(s1Value, s1len, s2Value, s2len, start);
+			} else if (s1.count < 0 && s2.count < 0) {
+				// both s1 and s2 are decompressed
+				return helpers.intrinsicIndexOfStringUTF16(s1Value, s1len, s2Value, s2len, start);
 			} else {
+				// s1 decompressed and s2 compressed
 				char firstChar = s2.charAtInternal(0, s2Value);
 
 				while (true) {
--- a/runtime/compiler/env/j9method.cpp
+++ b/runtime/compiler/env/j9method.cpp
@@ -3384,6 +3384,8 @@ TR_ResolvedJ9Method::TR_ResolvedJ9Method(TR_OpaqueMethodBlock * aMethod, TR_Fron
       {
       {x(TR::com_ibm_jit_JITHelpers_is32Bit,                                  "is32Bit", "()Z")},
       {x(TR::com_ibm_jit_JITHelpers_isArray,                                  "isArray", "(Ljava/lang/Object;)Z")},
+      {x(TR::com_ibm_jit_JITHelpers_intrinsicIndexOfStringLatin1,             "intrinsicIndexOfStringLatin1", "([CI[CII)I")},
+      {x(TR::com_ibm_jit_JITHelpers_intrinsicIndexOfStringUTF16,              "intrinsicIndexOfStringUTF16", "([CI[CII)I")},
       {x(TR::com_ibm_jit_JITHelpers_intrinsicIndexOfLatin1,                   "intrinsicIndexOfLatin1", "(Ljava/lang/Object;BII)I")},
       {x(TR::com_ibm_jit_JITHelpers_intrinsicIndexOfUTF16,                    "intrinsicIndexOfUTF16", "(Ljava/lang/Object;CII)I")},
 #ifdef TR_TARGET_32BIT
@@ -3477,10 +3479,17 @@ TR_ResolvedJ9Method::TR_ResolvedJ9Method(TR_OpaqueMethodBlock * aMethod, TR_Fron
       { TR::unknownMethod}
       };
 
+   static X StringLatin1Methods[] =
+      {
+      { x(TR::java_lang_StringLatin1_indexOf,                                 "indexOf",       "([BI[BII)I")},
+      { TR::unknownMethod }
+      };
+
    static X StringUTF16Methods[] =
       {
-      { x(TR::java_lang_StringUTF16_getChar,                                  "getChar", "([BI)C")},
-      { x(TR::java_lang_StringUTF16_toBytes,                                  "toBytes", "([CII)[B")},
+      { x(TR::java_lang_StringUTF16_getChar,                                  "getChar",        "([BI)C")},
+      { x(TR::java_lang_StringUTF16_indexOf,                                  "indexOf",        "([BI[BII)I")},
+      { x(TR::java_lang_StringUTF16_toBytes,                                  "toBytes",        "([CII)[B")},
       { TR::unknownMethod }
       };
 
@@ -4331,6 +4340,7 @@ TR_ResolvedJ9Method::TR_ResolvedJ9Method(TR_OpaqueMethodBlock * aMethod, TR_Fron
       { "java/text/NumberFormat", NumberFormatMethods },
       { "com/ibm/jit/JITHelpers", JITHelperMethods},
       { "java/lang/StringCoding", StringCodingMethods },
+      { "java/lang/StringLatin1", StringLatin1Methods },
       { 0 }
       };
 
--- a/runtime/compiler/z/codegen/J9CodeGenerator.cpp
+++ b/runtime/compiler/z/codegen/J9CodeGenerator.cpp
@@ -3950,6 +3950,8 @@ extern TR::Register *inlineBigDecimalToPackedConverter(TR::Node * node, TR::Code
 extern TR::Register *toUpperIntrinsic(TR::Node * node, TR::CodeGenerator * cg, bool isCompressedString);
 extern TR::Register *toLowerIntrinsic(TR::Node * node, TR::CodeGenerator * cg, bool isCompressedString);
 
+extern TR::Register* inlineVectorizedStringIndexOf(TR::Node* node, TR::CodeGenerator* cg, bool isCompressed);
+
 extern TR::Register *intrinsicIndexOf(TR::Node * node, TR::CodeGenerator * cg, bool isCompressed);
 
 extern TR::Register *inlineDoubleMax(TR::Node *node, TR::CodeGenerator *cg);
@@ -4235,17 +4237,25 @@ J9::Z::CodeGenerator::inlineDirectCall(
 
    if (cg->getSupportsInlineStringIndexOf())
       {
-         switch (methodSymbol->getRecognizedMethod())
-            {
-            case TR::com_ibm_jit_JITHelpers_intrinsicIndexOfLatin1:
-               resultReg = intrinsicIndexOf(node, cg, true);
+      switch (methodSymbol->getRecognizedMethod())
+         {
+         case TR::com_ibm_jit_JITHelpers_intrinsicIndexOfLatin1:
+            resultReg = intrinsicIndexOf(node, cg, true);
+            return true;
+         case TR::com_ibm_jit_JITHelpers_intrinsicIndexOfUTF16:
+            resultReg = intrinsicIndexOf(node, cg, false);
+            return true;
+         case TR::java_lang_StringLatin1_indexOf:
+         case TR::com_ibm_jit_JITHelpers_intrinsicIndexOfStringLatin1:
+               resultReg = inlineVectorizedStringIndexOf(node, cg, false);
                return true;
-            case TR::com_ibm_jit_JITHelpers_intrinsicIndexOfUTF16:
-               resultReg = intrinsicIndexOf(node, cg, false);
+         case TR::java_lang_StringUTF16_indexOf:
+         case TR::com_ibm_jit_JITHelpers_intrinsicIndexOfStringUTF16:
+               resultReg = inlineVectorizedStringIndexOf(node, cg, true);
                return true;
-            default:
-               break;
-            }
+         default:
+            break;
+         }
       }
 
       if (!comp->getOption(TR_DisableSIMDDoubleMaxMin) && cg->getSupportsVectorRegisters())
--- a/runtime/compiler/z/codegen/J9TreeEvaluator.cpp
+++ b/runtime/compiler/z/codegen/J9TreeEvaluator.cpp
@@ -211,6 +211,356 @@ doubleMaxMinHelper(TR::Node *node, TR::CodeGenerator *cg, bool isMaxOp)
    return node->getRegister();
    }
 
+/**
+ * \brief
+ *
+ * Use vector instructions to find the index of a sub-string inside
+ * a string assuming both strings have the same element size. Each element
+ * is 1-byte for compact strings and 2-bytes for non-compressed strings.
+ *
+ * \details
+ *
+ * The vector sequence searches for the first character of the sub-string
+ * inside the source/master string. If the first character is located, it'll
+ * perform iterative vector binary compares to match the rest of the sub-string
+ * starting from the first character position.
+ *
+ * This evaluator inlines the following Java intrinsic methods:
+ *
+ * <verbatim>
+ * For Java 9 and above:
+ *
+ * StringLatin1.indexOf(s1Value, s1Length, s2Value, s2Length, fromIndex);
+ * StringUTF16.indexOf(s1Value, s1Length, s2Value, s2Length, fromIndex);
+ *
+ * For Java 8:
+ * com.ibm.jit.JITHelpers.intrinsicIndexOfStringLatin1(char[] s1Value, int s1len, char[] s2Value, int s2len, int start);
+ * com.ibm.jit.JITHelpers.intrinsicIndexOfStringUTF16(char[] s1Value, int s1len, char[] s2Value, int s2len, int start);
+ *
+ * Assumptions:
+ *
+ * -# 0 <= fromIndex < s1Length
+ * -# s1Length could be anything: positive, negative or 0.
+ * -# s2Length > 0
+ * -# s1Value and s2Value are non-null arrays and are interpreted as byte arrays.
+ * -# s1Length and s2Length are not related. i.e. s1Length could be smallers than s2Length.
+ *
+ * <\verbatim>
+ *
+ * \param node the intrinsic function call node
+ * \param cg the code generator
+ * \param isUTF16 true if the string is a decompressed string.
+ *
+ * \return a register for that contains the indexOf() result.
+*/
+TR::Register*
+inlineVectorizedStringIndexOf(TR::Node* node, TR::CodeGenerator* cg, bool isUTF16)
+   {
+   // Element size is 2-byte if both s1 and s2 are decompressed strings and it is 12 otherwise.
+   // 0 for byte, 1 for halfword
+   const uint32_t elementSizeMask = isUTF16 ? 1 : 0;
+   const int8_t vectorSize = cg->machine()->getVRFSize();
+   const uintptrj_t headerSize = TR::Compiler->om.contiguousArrayHeaderSizeInBytes();
+   TR::Compilation* comp = cg->comp();
+
+   if (comp->getOption(TR_TraceCG))
+      traceMsg(comp, "inlineVectorizedStringIndexOf. Is isUTF16 %d\n", isUTF16);
+
+   // Get call parameters
+   // s1Value and s2Value are byte arrays
+   // This evaluator function handles different indexOf() instrinsics, some of which
+   // are static calls without a receiver. Hence, the need for static call check.
+   const bool isStaticCall = node->getSymbolReference()->getSymbol()->castToMethodSymbol()->isStatic();
+   const uint8_t firstCallArgIdx = isStaticCall ? 0 : 1;
+
+   TR::Register* s1ValueReg   = cg->evaluate          (node->getChild(firstCallArgIdx));
+   TR::Register* s1LenReg     = cg->gprClobberEvaluate(node->getChild(firstCallArgIdx+1));
+   TR::Register* s2ValueReg   = cg->evaluate          (node->getChild(firstCallArgIdx+2));
+   TR::Register* s2LenReg     = cg->gprClobberEvaluate(node->getChild(firstCallArgIdx+3));
+   TR::Register* fromIndexReg = cg->gprClobberEvaluate(node->getChild(firstCallArgIdx+4));
+
+   // Registers
+   TR::Register* resultReg          = cg->allocateRegister();
+   TR::Register* maxIndexReg        = cg->allocateRegister();
+   TR::Register* s1VecStartIndexReg = fromIndexReg;
+   TR::Register* s2VecStartIndexReg = cg->allocateRegister();
+   TR::Register* loadLenReg         = cg->allocateRegister();
+   TR::Register* char1IndexReg      = resultReg;
+
+   TR::Register* s1PartialVReg      = cg->allocateRegister(TR_VRF);
+   TR::Register* s2PartialVReg      = cg->allocateRegister(TR_VRF);
+   TR::Register* s2Char1RepVReg     = cg->allocateRegister(TR_VRF);
+   TR::Register* tmpVReg            = cg->allocateRegister(TR_VRF);
+
+   // Register dependencies
+   TR::RegisterDependencyConditions * regDeps = new (cg->trHeapMemory()) TR::RegisterDependencyConditions(0, 13, cg);
+   regDeps->addPostCondition(s1ValueReg  , TR::RealRegister::AssignAny);
+   regDeps->addPostCondition(s1LenReg    , TR::RealRegister::AssignAny);
+   regDeps->addPostCondition(s2ValueReg  , TR::RealRegister::AssignAny);
+   regDeps->addPostCondition(s2LenReg    , TR::RealRegister::AssignAny);
+   regDeps->addPostCondition(fromIndexReg, TR::RealRegister::AssignAny);
+
+   regDeps->addPostCondition(resultReg         , TR::RealRegister::AssignAny);
+   regDeps->addPostCondition(maxIndexReg       , TR::RealRegister::AssignAny);
+   regDeps->addPostCondition(s2VecStartIndexReg, TR::RealRegister::AssignAny);
+   regDeps->addPostCondition(loadLenReg        , TR::RealRegister::AssignAny);
+
+   regDeps->addPostCondition(s1PartialVReg , TR::RealRegister::AssignAny);
+   regDeps->addPostCondition(s2PartialVReg , TR::RealRegister::AssignAny);
+   regDeps->addPostCondition(s2Char1RepVReg, TR::RealRegister::AssignAny);
+   regDeps->addPostCondition(tmpVReg       , TR::RealRegister::AssignAny);
+
+   // Labels
+   TR::LabelSymbol* labelStart               = generateLabelSymbol(cg);
+   TR::LabelSymbol* labelFindS2Head          = generateLabelSymbol(cg);
+   TR::LabelSymbol* labelLoadLen16           = generateLabelSymbol(cg);
+   TR::LabelSymbol* labelLoadLenDone         = generateLabelSymbol(cg);
+   TR::LabelSymbol* labelExtractFirstCharPos = generateLabelSymbol(cg);
+   TR::LabelSymbol* labelMatchS2Loop         = generateLabelSymbol(cg);
+   TR::LabelSymbol* labelMatchS2LoopSetup    = generateLabelSymbol(cg);
+   TR::LabelSymbol* labelS2PartialMatch      = generateLabelSymbol(cg);
+   TR::LabelSymbol* labelLoadResult          = generateLabelSymbol(cg);
+   TR::LabelSymbol* labelResultDone          = generateLabelSymbol(cg);
+   TR::LabelSymbol* labelStringNotFound      = generateLabelSymbol(cg);
+   TR::LabelSymbol* labelDone                = generateLabelSymbol(cg);
+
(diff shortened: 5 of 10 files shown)


NigelYiboYu:
@fjeremic ready for review.

fjeremic:
@NigelYiboYu could you please post some performance numbers on this change? Also if we have any benchmarks to check in let's do so (BumbleBench for example).

fjeremic:
Requesting review from @DanHeidinga for the JCL portion as well to get a second pair of eyes on it. Feel free to delegate if needed if there is someone else that should review. Thanks!

NigelYiboYu:
Bumblebench `String.indexOf(String) benchmark` uses different long strings as `master strings`, creates sub-strings from it and searches for them. The input master strings are around 64K characters long and can either be compressible random strings or a compressible human-readable string pulled from a Wikipedia article. Bumblebench produced mixed good and bad results and indicates that vector implementation performance is very workload dependent: the vector implementation has a degraded performance **(0% to -30%**)  in random strings but is more performant searching inside the human-readable wiki string.

The next table shows the vector vs. non-vector performance delta in percentage. Wikipedia article was used as the master string.

The x-axis is the sub-string length and the y-axis is the index of the sub-string in the master string. The lengths and indices are in the number of characters. For instance, a string of ~10 in length at an index of ~10 means it's approximately 10 characters long and is located at index 10 of the master string.

Negative numbers here mean that vector string.indexof(string) being slower than non-vector implementation.

  |   | sub-string   length |   |  
-- | -- | -- | -- | --
perf delta % |   | ~10 | ~100 | ~1K | ~10K
  | ~10 | 49% | 0% | 296% | -79%
index | ~100 | 5% | -6% | 0% | 0%
  | ~1K | 0% | -13% | 0% | 0%
  | ~50K |   | 0% | -9% | 0%


The table below shows compact string results.

  |   | sub-string   length |   |  
-- | -- | -- | -- | --
perf delta % |   | ~10 | ~100 | ~1K | ~10K
  | ~10 | 83% | 39% | -5% | -18%
index | ~100 | 42% | 40% | 15% | 4%
  | ~1K | 188% | 155% | 218% | 234%
  | ~50K |   | 10537% | 10795% | 11135%

jcl/src/java.base/share/classes/java/lang/String.java
```diff
@@ -6054,30 +6060,15 @@ public int indexOf(String subString, int start) {
 			s1Value.getClass(); // Implicit null check
 			s2Value.getClass(); // Implicit null check
 
+			// Nigel changes
```
fjeremic: Can we remove these references?

runtime/compiler/env/j9method.cpp
```diff
@@ -4624,6 +4634,25 @@ TR_ResolvedJ9Method::TR_ResolvedJ9Method(TR_OpaqueMethodBlock * aMethod, TR_Fron
             setRecognizedMethodInfo(TR::java_util_ArrayList_all);
          else if ((classNameLen == 19) && !strncmp(className, "java/util/Hashtable", 19))
             setRecognizedMethodInfo(TR::java_util_Hashtable_all);
+         else if ((classNameLen == 21) && !strncmp(className, "java/lang/StringUTF16", 21))
+            {
+            // StringUTF16 has two methods that start with 'indexOf'. For vector support, we only
+            // recognize and inline indexOf().
+            if ((strncmp(name, "indexOf", 7) == 0) && (strncmp(name, "indexOfLatin1", 13) != 0))
+               setRecognizedMethodInfo(TR::java_lang_StringUTF16_indexOf);
+            }
```
fjeremic: Can you explain why exactly we need this chunk of code and the subsequent `else if` blocks that were added?

runtime/compiler/z/codegen/J9TreeEvaluator.cpp
```diff
@@ -211,6 +211,372 @@ doubleMaxMinHelper(TR::Node *node, TR::CodeGenerator *cg, bool isMaxOp)
...
+ * -# s2Length > 0
+ * -# s1Value and s2Value are non-null arrays and are interpreted as byte arrays.
+ * -# s1Length and s2Length are not related. i.e. s1Length could be smallers than s2Length.
+ *
+ * <\verbatim>
+ *
+ *
+ * \param node the intrinsic function call node
+ * \param cg the code generator
+ * \param elementSize element size in bytes. This is 1 for compact strings of Latin1 letters, and it's 2 for non-compact strings.
```
fjeremic: I would avoid giving user the option here. The evaluator only seems to support `1` and `2` so why even give the user the option to accidentally pass `4` for example? Why not follow the footsteps of `intrinsicIndexOf` and supply a boolean dictating whether the `char` is compressed or not? This would also add symmetry to the APIs.

runtime/compiler/z/codegen/J9CodeGenerator.cpp
```diff
@@ -4235,17 +4237,33 @@ J9::Z::CodeGenerator::inlineDirectCall(
...
+         {
+         case TR::com_ibm_jit_JITHelpers_intrinsicIndexOfLatin1:
+            resultReg = intrinsicIndexOf(node, cg, true);
+            return true;
+         case TR::com_ibm_jit_JITHelpers_intrinsicIndexOfUTF16:
+            resultReg = intrinsicIndexOf(node, cg, false);
+            return true;
+         case TR::java_lang_StringLatin1_indexOf:
+         case TR::com_ibm_jit_JITHelpers_intrinsicIndexOfStringLatin1:
+            if (!comp->getOption(TR_DisableVectorStringIndexOfString))
```
fjeremic: Don't we already have a `TR_DisableFastStringIndexOf` option for this? Do we need the new one?

runtime/compiler/z/codegen/J9TreeEvaluator.cpp
```diff
@@ -211,6 +211,372 @@ doubleMaxMinHelper(TR::Node *node, TR::CodeGenerator *cg, bool isMaxOp)
...
+   TR::LabelSymbol* label_match_s2_loop_prep   = TR::LabelSymbol::create(cg->trHeapMemory(), cg);
+   TR::LabelSymbol* label_s2_partial_match     = TR::LabelSymbol::create(cg->trHeapMemory(), cg);
+   TR::LabelSymbol* label_load_result          = TR::LabelSymbol::create(cg->trHeapMemory(), cg);
+   TR::LabelSymbol* label_done                 = TR::LabelSymbol::create(cg->trHeapMemory(), cg);
+
+   // Ops
+   TR::InstOpCode::Mnemonic AHI  = is64Bit ? TR::InstOpCode::AGHI : TR::InstOpCode::AHI;
+   TR::InstOpCode::Mnemonic AHIK = is64Bit ? TR::InstOpCode::AGHIK : TR::InstOpCode::AHIK;
+   TR::InstOpCode::Mnemonic ARK  = is64Bit ? TR::InstOpCode::AGRK : TR::InstOpCode::ARK;
+   TR::InstOpCode::Mnemonic AR   = is64Bit ? TR::InstOpCode::AGR : TR::InstOpCode::AR;
```
fjeremic: We should be using the `TR::InstOpCode` APIs for these. For example this one should be `TR::InstOpCode::getAddRegOpCode()`. I suggest inlining them as well to keep symmetry with all other evaluators.

runtime/compiler/z/codegen/J9TreeEvaluator.cpp
```diff
@@ -211,6 +211,372 @@ doubleMaxMinHelper(TR::Node *node, TR::CodeGenerator *cg, bool isMaxOp)
...
+   TR::Register* r_tmp2            = cg->allocateRegister();
+   TR::Register* r_1stCharIndex    = cg->allocateRegister();
+
+   TR::Register* v_s1Partial       = cg->allocateRegister(TR_VRF);
+   TR::Register* v_s2Partial       = cg->allocateRegister(TR_VRF);
+   TR::Register* v_s2Char1Rep      = cg->allocateRegister(TR_VRF);
+   TR::Register* v_tmp             = cg->allocateRegister(TR_VRF);
+
+   // register dependencies
+   TR::RegisterDependencyConditions * regDeps = new (cg->trHeapMemory()) TR::RegisterDependencyConditions(0, 17, cg);
```
fjeremic: This is a scary number of dependencies. I count 13 GPRs which is almost more than we have available. The register pressure here is going to be huge causing a ton of spilling of live values. Is there a way to reduce the number of registers you need for this evaluator?

runtime/compiler/z/codegen/J9TreeEvaluator.cpp
```diff
@@ -211,6 +211,372 @@ doubleMaxMinHelper(TR::Node *node, TR::CodeGenerator *cg, bool isMaxOp)
...
+   regDeps->addPostCondition(r_tmp2           , TR::RealRegister::AssignAny);
+   regDeps->addPostCondition(r_1stCharIndex   , TR::RealRegister::AssignAny);
+
+   regDeps->addPostCondition(v_s1Partial , TR::RealRegister::AssignAny);
+   regDeps->addPostCondition(v_s2Partial , TR::RealRegister::AssignAny);
+   regDeps->addPostCondition(v_s2Char1Rep, TR::RealRegister::AssignAny);
+   regDeps->addPostCondition(v_tmp       , TR::RealRegister::AssignAny);
+
+   // Labels
+   TR::LabelSymbol* label_find_1st_char        = TR::LabelSymbol::create(cg->trHeapMemory(), cg);
```
fjeremic: Please use the `generateLabelSymbol` API for these to be consistent with the rest of the codebase.

runtime/compiler/z/codegen/J9TreeEvaluator.cpp
```diff
@@ -211,6 +211,372 @@ doubleMaxMinHelper(TR::Node *node, TR::CodeGenerator *cg, bool isMaxOp)
...
+   TR::Instruction* cursor = generateRIInstruction(cg, TR::InstOpCode::LHI, node, r_resultReg, -1);
+   generateRRInstruction(cg, TR::InstOpCode::LGFR, node, r_s1Len, r_s1Len);
+   generateRRInstruction(cg, TR::InstOpCode::LGFR, node, r_s2Len, r_s2Len);
+   generateRRInstruction(cg, TR::InstOpCode::LGFR, node, r_fromIndex, r_fromIndex);
+
+   if (isUTF16)
+      {
+      generateRSInstruction(cg, SLL, node, r_s1Len, r_s1Len, 1);         // these are byte lengths
+      generateRSInstruction(cg, SLL, node, r_s2Len, r_s2Len, 1);
+      generateRSInstruction(cg, SLL, node, r_fromIndex, r_fromIndex, 1);
```
fjeremic: Can't we use `RISBG` instructions here to do the `LGFR` + `SLL` all in one shot? Search around the codebase there is an API to do this. @dchopra001 should be familiar with it if you need help.

runtime/compiler/z/codegen/J9TreeEvaluator.cpp
```diff
@@ -211,6 +211,372 @@ doubleMaxMinHelper(TR::Node *node, TR::CodeGenerator *cg, bool isMaxOp)
...
+      }
+
+   cursor = generateRRRInstruction(cg, SRK, node, r_maxIndex, r_s1Len, r_s2Len);
+   cursor = generateRIEInstruction(cg, CRJ, node, r_maxIndex, r_fromIndex, label_done, TR::InstOpCode::COND_BLR);
+   cursor->setStartInternalControlFlow();
```
fjeremic: I think we're trying to avoid setting ICF on instructions and only do them on labels. We want to deprecate the non-label API in fact. See https://github.com/eclipse/omr/issues/2833 for details. We'll need a fix here.

runtime/compiler/z/codegen/J9TreeEvaluator.cpp
```diff
@@ -211,6 +211,372 @@ doubleMaxMinHelper(TR::Node *node, TR::CodeGenerator *cg, bool isMaxOp)
...
+
+   if (comp->getOption(TR_TraceCG))
+      traceMsg(comp, "inlineVectorizedStringIndexOf element size %d\n", elementSize);
+
+   // Get call parameters
+   // s1Value and s2Value are byte arrays
+   // This evaluator function handles different indexOf() instrinsics, some of which are static calls without a receiver. Hence, the need for static call check.
+   bool isStaticCall = node->getSymbolReference()->getSymbol()
+           && node->getSymbolReference()->getSymbol()->castToMethodSymbol()
+           && node->getSymbolReference()->getSymbol()->castToMethodSymbol()->isStatic();
```
fjeremic: If `node` doesn't have a method symbol associated with it I would think this whole evaluator would be wrong, hence I don't think we need to null check `node->getSymbolReference()->getSymbol()->castToMethodSymbol()` here. If you're paranoid add an assertion that it's non-null and avoid the two extra checks here.

However I suppose a better question here is why is `intrinsicIndexOfStringUTF16` not defined as `static`? Can't we just change the signature so we make sure all nodes that pass through this evaluator are `static` and we can get rid of this check and fold it away?

runtime/compiler/env/j9method.cpp
NigelYiboYu: This new checks are all under `      if (TR_Method::getMandatoryRecognizedMethod() == TR::unknownMethod)` where multiple methods all have the same signature and very similar names. The OpenJ9 recognized method mechanism could not distinguish Java functions like these.

Take the `StringUTF16` class for example, it has these two methods:
```
    public static int indexOf(byte[] value, int valueCount, byte[] str, int strCount, int fromIndex) 
    public static int indexOfLatin1(byte[] src, int srcCount, byte[] tgt, int tgtCount, int fromIndex) 
```

They have the same signature and the names all start with `indexOf`.  If this `if check` were not added, the JIT would recognize both `indexOf` and `indexOfLatin1` as the same method. I've seen this incorrect behavior in a method trace and added this check to fix that.

The JITHelper recognized method checks are added for the same reason.

runtime/compiler/z/codegen/J9CodeGenerator.cpp
NigelYiboYu: OK. I'll remove the new option and re-use `TR_DisableFastStringIndexOf`.

runtime/compiler/z/codegen/J9TreeEvaluator.cpp
NigelYiboYu: Yes. I'm worried about this, too. I can probably eliminate one GPR and see if there are other opportunities.

The problem here is that that we have 5 parameters, and two strings to point into. The number of variables here is naturally large.

runtime/compiler/z/codegen/J9TreeEvaluator.cpp
NigelYiboYu: Yeah, I was being paranoid about symbol checking. 

`intrinsicIndexOfStringUTF16` is not static because most methods in `com.ibm.JITHelper`  are not static. I was trying to be consistent with most JITHelper methods. 

I'll make relevant JITHelper methods static for this PR.

andrewcraik: The method should not be static since we use the JITHelpers instance singleton as part of the security checks around JITHelpers. All JITHelper calls in the String class should be calls to the helper instance methods which necessitates having an instance of JITHelpers and so the singleton accessor security checks will have run.

fjeremic: Ah right, forgot about that. Thanks!

jcl/src/java.base/share/classes/com/ibm/jit/JITHelpers.java
```diff
@@ -568,6 +568,59 @@ public char getCharFromArrayByIndex(Object obj, int index) {
...
 	}
 
+	public int intrinsicIndexOfStringLatin1(char[] s1Value, int s1len, char[] s2Value, int s2len, int start) {
+		char firstChar = byteToCharUnsigned(getByteFromArrayByIndex(s2Value, 0));
+
+		while (true) {
+			int i = -1;
+			if (start < s1len) {
+				i = intrinsicIndexOfLatin1(s1Value, (byte)firstChar, start, s1len);
+				}
```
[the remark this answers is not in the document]
NigelYiboYu: Fixed.

jcl/src/java.base/share/classes/java/lang/String.java
```diff
@@ -2017,6 +2017,9 @@ public int indexOf(String string) {
 	 * @see #lastIndexOf(String, int)
 	 */
 	public int indexOf(String subString, int start) {
+		if (subString.length() == 1)
+			return indexOf(subString.charAt(0), start);
```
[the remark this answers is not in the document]
NigelYiboYu: Fixed.

runtime/compiler/z/codegen/J9TreeEvaluator.cpp
NigelYiboYu: Changed the `elementSize ` parameter to a boolean `isCompact`.

runtime/compiler/z/codegen/J9TreeEvaluator.cpp
NigelYiboYu: Using inlined `TR::InstOpCode` getters everywhere.

runtime/compiler/z/codegen/J9TreeEvaluator.cpp
NigelYiboYu: Reduced the number of GPRs from 13 to 10.

runtime/compiler/z/codegen/J9TreeEvaluator.cpp
NigelYiboYu: Fixed.

runtime/compiler/z/codegen/J9TreeEvaluator.cpp
NigelYiboYu: Updated the code to use `generateShiftAndKeepSelected64Bit` to generate `RISBGs`.

runtime/compiler/z/codegen/J9TreeEvaluator.cpp
NigelYiboYu: Added a start label and used it as ICF start.

runtime/compiler/z/codegen/J9TreeEvaluator.cpp
NigelYiboYu: Removed the symbol null checks. But the static check stays.

NigelYiboYu:
Marking as WIP until tests are good.

fjeremic:
This PR depends on https://github.com/eclipse/omr/pull/3041.

fjeremic:
The bumblebench for this work is being contributed in https://github.com/AdoptOpenJDK/bumblebench/pull/10.

runtime/compiler/env/j9method.cpp
fjeremic: Ok I think I understand. Why do we need to check for `"java/lang/StringLatin1"` below then? It doesn't have these duplicates. Also why do we need to check `"com/ibm/jit/JITHelpers"` it also doesn't have duplicates.

runtime/compiler/z/codegen/J9CodeGenerator.cpp
```diff
@@ -4235,17 +4237,33 @@ J9::Z::CodeGenerator::inlineDirectCall(
...
+         {
+         case TR::com_ibm_jit_JITHelpers_intrinsicIndexOfLatin1:
+            resultReg = intrinsicIndexOf(node, cg, true);
+            return true;
+         case TR::com_ibm_jit_JITHelpers_intrinsicIndexOfUTF16:
+            resultReg = intrinsicIndexOf(node, cg, false);
+            return true;
+         case TR::java_lang_StringLatin1_indexOf:
+         case TR::com_ibm_jit_JITHelpers_intrinsicIndexOfStringLatin1:
+            if (!comp->getOption(TR_DisableFastStringIndexOf))
```
fjeremic: I don't think we need this check. We already account for this. `cg->getSupportsInlineStringIndexOf()` will return false if this option is active.

runtime/compiler/z/codegen/J9TreeEvaluator.cpp
```diff
@@ -211,6 +211,362 @@ doubleMaxMinHelper(TR::Node *node, TR::CodeGenerator *cg, bool isMaxOp)
...
+      if (isUTF16)
+         {
+         generateRSInstruction(cg, TR::InstOpCode::SLL, node, r_s1Len, 1);
+         generateRSInstruction(cg, TR::InstOpCode::SLL, node, r_s2Len, 1);
+         generateRSInstruction(cg, TR::InstOpCode::SLL, node, r_fromIndex, 1);
+         }
+      }
+
+   cursor = generateRRRInstruction(cg, TR::InstOpCode::getSubtractThreeRegOpCode(), node, r_maxIndex, r_s1Len, r_s2Len);
+   cursor = generateRIEInstruction(cg, TR::InstOpCode::getCmpRegAndBranchRelOpCode(), node, r_maxIndex, r_fromIndex, label_string_not_found, TR::InstOpCode::COND_BLR);
```
fjeremic: Similarly:

```suggestion
   generateRIEInstruction(cg, TR::InstOpCode::getCmpRegAndBranchRelOpCode(), node, r_maxIndex, r_fromIndex, label_string_not_found, TR::InstOpCode::COND_BLR);
```

runtime/compiler/z/codegen/J9TreeEvaluator.cpp
```diff
@@ -211,6 +211,362 @@ doubleMaxMinHelper(TR::Node *node, TR::CodeGenerator *cg, bool isMaxOp)
...
+
+      if (isUTF16)
+         {
+         generateRSInstruction(cg, TR::InstOpCode::SLL, node, r_s1Len, 1);
+         generateRSInstruction(cg, TR::InstOpCode::SLL, node, r_s2Len, 1);
+         generateRSInstruction(cg, TR::InstOpCode::SLL, node, r_fromIndex, 1);
+         }
+      }
+
+   cursor = generateRRRInstruction(cg, TR::InstOpCode::getSubtractThreeRegOpCode(), node, r_maxIndex, r_s1Len, r_s2Len);
```
fjeremic: I don't think `cursor` is used until the end of the ICF:

```suggestion
   generateRRRInstruction(cg, TR::InstOpCode::getSubtractThreeRegOpCode(), node, r_maxIndex, r_s1Len, r_s2Len);
```

runtime/compiler/z/codegen/J9TreeEvaluator.cpp
```diff
@@ -211,6 +211,362 @@ doubleMaxMinHelper(TR::Node *node, TR::CodeGenerator *cg, bool isMaxOp)
...
+ *
+ * -# 0 <= fromIndex < s1Length
+ * -# s1Length could be anything: positive, negative or 0.
+ * -# s2Length > 0
+ * -# s1Value and s2Value are non-null arrays and are interpreted as byte arrays.
+ * -# s1Length and s2Length are not related. i.e. s1Length could be smallers than s2Length.
+ *
+ * <\verbatim>
+ *
+ *
```
fjeremic: Extra newline.

runtime/compiler/z/codegen/J9TreeEvaluator.cpp
```diff
@@ -211,6 +211,362 @@ doubleMaxMinHelper(TR::Node *node, TR::CodeGenerator *cg, bool isMaxOp)
...
+ * \param node the intrinsic function call node
+ * \param cg the code generator
+ * \param isCompressed true if the string is compressed.
+ *
```
fjeremic: I think we need a `\return` clause here.

runtime/compiler/z/codegen/J9TreeEvaluator.cpp
```diff
@@ -211,6 +211,362 @@ doubleMaxMinHelper(TR::Node *node, TR::CodeGenerator *cg, bool isMaxOp)
...
+   uintptrj_t headerSize = TR::Compiler->om.contiguousArrayHeaderSizeInBytes();
+   TR::Compilation* comp = cg->comp();
+
+   if (comp->getOption(TR_TraceCG))
+      traceMsg(comp, "inlineVectorizedStringIndexOf. is compressed %d\n", isCompressed);
+
+   // Get call parameters
+   // s1Value and s2Value are byte arrays
+   // This evaluator function handles different indexOf() instrinsics, some of which are static calls without a receiver. Hence, the need for static call check.
+   bool isStaticCall = node->getSymbolReference()->getSymbol()->castToMethodSymbol()->isStatic();
```
fjeremic: This can be `const` as can the next line below this.

runtime/compiler/z/codegen/J9TreeEvaluator.cpp
NigelYiboYu: done

runtime/compiler/z/codegen/J9TreeEvaluator.cpp
NigelYiboYu: Fixed

runtime/compiler/z/codegen/J9TreeEvaluator.cpp
NigelYiboYu: Removed `cursor=` from these two lines.

runtime/compiler/z/codegen/J9TreeEvaluator.cpp
NigelYiboYu: Added `const` here and some other places.

runtime/compiler/z/codegen/J9TreeEvaluator.cpp
```diff
@@ -211,6 +211,361 @@ doubleMaxMinHelper(TR::Node *node, TR::CodeGenerator *cg, bool isMaxOp)
...
+   // s1Value and s2Value are byte arrays
+   // This evaluator function handles different indexOf() instrinsics, some of which are static calls without a receiver. Hence, the need for static call check.
+   const bool isStaticCall = node->getSymbolReference()->getSymbol()->castToMethodSymbol()->isStatic();
+   const uint8_t firstCallArgIdx = isStaticCall ? 0 : 1;
+
+   TR::Register* r_s1Value   = cg->evaluate          (node->getChild(firstCallArgIdx));
+   TR::Register* r_s1Len     = cg->gprClobberEvaluate(node->getChild(firstCallArgIdx+1));
+   TR::Register* r_s2Value   = cg->evaluate          (node->getChild(firstCallArgIdx+2));
+   TR::Register* r_s2Len     = cg->gprClobberEvaluate(node->getChild(firstCallArgIdx+3));
+   TR::Register* r_fromIndex = cg->gprClobberEvaluate(node->getChild(firstCallArgIdx+4));
```
fjeremic: I suggest we don't deviate from the naming scheme of the rest of the codebase. In particular I don't think we've ever used Hungarian notation for local variables. Let's try not to deviate too much and perhaps use a "Reg" suffix (or "Register" which seems the be the next most common though needlessly longer):

https://github.com/eclipse/openj9/blob/8d06ec3edc1c5ad262b9d6221974a410ae20220b/runtime/compiler/z/codegen/J9TreeEvaluator.cpp#L1527-L1528

https://github.com/eclipse/openj9/blob/8d06ec3edc1c5ad262b9d6221974a410ae20220b/runtime/compiler/z/codegen/J9TreeEvaluator.cpp#L1841

https://github.com/eclipse/openj9/blob/8d06ec3edc1c5ad262b9d6221974a410ae20220b/runtime/compiler/z/codegen/J9TreeEvaluator.cpp#L1897

https://github.com/eclipse/openj9/blob/8d06ec3edc1c5ad262b9d6221974a410ae20220b/runtime/compiler/z/codegen/J9TreeEvaluator.cpp#L2228-L2229

https://github.com/eclipse/openj9/blob/8d06ec3edc1c5ad262b9d6221974a410ae20220b/runtime/compiler/z/codegen/J9TreeEvaluator.cpp#L2228-L2229

etc.

fjeremic: Similar comments for labels apply.

NigelYiboYu: Changed all variables names.

runtime/compiler/z/codegen/J9TreeEvaluator.cpp
```diff
@@ -211,6 +211,361 @@ doubleMaxMinHelper(TR::Node *node, TR::CodeGenerator *cg, bool isMaxOp)
...
+   // decrement the receiver, which is not used in this intrinsic sequence
+   if (!isStaticCall)
+      cg->decReferenceCount(node->getChild(0));
+
+   // decrement the call parameters.
+   cg->decReferenceCount(node->getChild(firstCallArgIdx));
+   cg->decReferenceCount(node->getChild(firstCallArgIdx+1));
+   cg->decReferenceCount(node->getChild(firstCallArgIdx+2));
+   cg->decReferenceCount(node->getChild(firstCallArgIdx+3));
+   cg->decReferenceCount(node->getChild(firstCallArgIdx+4));
```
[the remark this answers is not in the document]
NigelYiboYu: Done.

NigelYiboYu:
> see some test cases added covering corner case

Will check in some `String.indexOf()` tests to the test directory.

NigelYiboYu:
@fjeremic  Updated the PR with a functional sanity test in `openj9/test` folder.

test/functional/Java8andUp/src/org/openj9/test/string/StringIndexOfString.java
```diff
@@ -0,0 +1,236 @@
...
+				subStringByteLen = (VECTOR_SIZE * 6 + 6);
+				break;
+			case 19:
+				// same as case 15, sub-string to be modified.
+				// test case for string not found. mismatch at the end
+				subStringByteIndex = (VECTOR_SIZE * 7 + 8);
+				subStringByteLen = (VECTOR_SIZE * 6 + 6);
+				break;
+			default:
+				System.out.println("unknown test case number");
```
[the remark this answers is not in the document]
NigelYiboYu: Used JUnit.AssertTrue instead.

test/functional/Java8andUp/playlist.xml
```diff
@@ -1909,7 +1909,29 @@
 			<subset>11</subset>
 		</subsets>
 	</test>
-
+	<test>
+		<testCaseName>testStringIndexOfString</testCaseName>
+		<command>$(ADD_JVM_LIB_DIR_TO_LIBPATH) $(JAVA_COMMAND) '-Xjit:disableAsyncCompilation,tryToInline={java/lang/String.indexOf*},limit={org/openj9/test/string/StringIndexOfString.doTest*},{org/openj9/test/string/StringIndexOfString.doTest*}(count=50)' $(JVM_OPTIONS) \
```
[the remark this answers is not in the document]
NigelYiboYu: Added `NoOptions` and this custom JIT option as two variations of the test.

test/functional/Java8andUp/src/org/openj9/test/string/StringIndexOfString.java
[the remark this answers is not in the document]
fjeremic: I don't see the updated change. Perhaps you forgot to push?

test/functional/Java8andUp/src/org/openj9/test/string/StringIndexOfString.java
```diff
@@ -0,0 +1,236 @@
+/*******************************************************************************
+ * Copyright (c) 2001, 2018 IBM Corp. and others
```
fjeremic: This is the wrong copyright date. I believe this file is new.

NigelYiboYu: Fixed.

test/functional/Java8andUp/src/org/openj9/test/string/StringIndexOfString.java
```diff
@@ -0,0 +1,236 @@
...
+			}
+
+			expectedResults[testCaseNum - 1] = expectedResult;
+			inputStrings[testCaseNum - 1] = subString;
+		}
+	}
+
+	@Test
+	public void testStringIndexOfString() {
+		generateInputs();
```
fjeremic: I don't see anything in `generateInputs` that requires runtime input, hence we should be moving it to a static constructor so these inputs are only generated once, not once per test since this test may be executed many, many times within the test bucket.

test/functional/Java8andUp/src/org/openj9/test/string/StringIndexOfString.java
```diff
@@ -0,0 +1,236 @@
...
+import org.testng.Assert;
+import org.testng.AssertJUnit;
+import java.lang.reflect.Field;
+import java.lang.management.ManagementFactory;
+import java.lang.management.RuntimeMXBean;
+import java.util.List;
+
+/**
+ * Designed to test SIMD accelerated String.indexOf(String) on platforms such as
+ * the z/Architecture. The vector register size is assumed to be 16 bytes.
```
fjeremic: Should be "such as z/Architecture".

test/functional/Java8andUp/src/org/openj9/test/string/StringIndexOfString.java
```diff
@@ -0,0 +1,236 @@
...
+ */
+@Test(groups = { "level.sanity" })
+public class StringIndexOfString {
+
+	public static final Logger logger = Logger.getLogger(StringIndexOfString.class);
+
+	private static final String shortString200Char = "Canada ic *a zzthe xxwwthi xx country located in ...zzthe xxwwthi xx northern part of North Americann. Its ten provinces and three territories extend from the Atlantic to the e xxwwthi xx northern part of North America. Its te Pacific and northward into the Arctic Ocean, covering 9.98 million square kilometres (3.85 million square miles), making it the world's second-largest country by total area. \";\r\n";
+	private static final int NUM_TEST_CASE = 19;
+	private static int VECTOR_SIZE = 16;
```
fjeremic: Missing `final` keyword here.

test/functional/Java8andUp/src/org/openj9/test/string/StringIndexOfString.java
```diff
@@ -0,0 +1,236 @@
...
+/**
+ * Designed to test SIMD accelerated String.indexOf(String) on platforms such as
+ * the z/Architecture. The vector register size is assumed to be 16 bytes.
+ */
+@Test(groups = { "level.sanity" })
+public class StringIndexOfString {
+
+	public static final Logger logger = Logger.getLogger(StringIndexOfString.class);
+
+	private static final String shortString200Char = "Canada ic *a zzthe xxwwthi xx country located in ...zzthe xxwwthi xx northern part of North Americann. Its ten provinces and three territories extend from the Atlantic to the e xxwwthi xx northern part of North America. Its te Pacific and northward into the Arctic Ocean, covering 9.98 million square kilometres (3.85 million square miles), making it the world's second-largest country by total area. \";\r\n";
```
fjeremic: I would make this public and avoid using `final` since the JIT can perform static final field folding and fold away all operations on this string.

runtime/compiler/z/codegen/J9TreeEvaluator.cpp
```diff
@@ -211,6 +211,355 @@ doubleMaxMinHelper(TR::Node *node, TR::CodeGenerator *cg, bool isMaxOp)
...
+ *
+ * \return a register for that contains the indexOf() result.
+*/
+TR::Register*
+inlineVectorizedStringIndexOf(TR::Node* node, TR::CodeGenerator* cg, bool isCompressed)
+   {
+   // element size is 2-byte if both s1 and s2 are decompressed strings and it is 12 otherwise.
+   // 0 for byte, 1 for halfword
+   const uint32_t elementSizeMask = isCompressed ? 0 : 1;
+   const bool isUTF16 = !isCompressed;
```
fjeremic: Why complicate things and not just inline this, or modify the parameter to be `isUTF16`?

runtime/compiler/z/codegen/J9TreeEvaluator.cpp
```diff
@@ -211,6 +211,355 @@ doubleMaxMinHelper(TR::Node *node, TR::CodeGenerator *cg, bool isMaxOp)
...
+   TR::LabelSymbol* labelMatchS2LoopSetup    = generateLabelSymbol(cg);
+   TR::LabelSymbol* labelS2PartialMatch      = generateLabelSymbol(cg);
+   TR::LabelSymbol* labelLoadResult          = generateLabelSymbol(cg);
+   TR::LabelSymbol* labelResultDone          = generateLabelSymbol(cg);
+   TR::LabelSymbol* labelStringNotFound      = generateLabelSymbol(cg);
+   TR::LabelSymbol* labelDone                = generateLabelSymbol(cg);
+
+   // Start
+   TR::Instruction* cursor = generateS390LabelInstruction(cg, TR::InstOpCode::LABEL, node, labelStart);
+   cursor->setStartInternalControlFlow();
```
fjeremic: See https://github.com/eclipse/omr/pull/3106. We're deprecating these APIs and only setting ICF on labels. We need to change these accordingly.

runtime/compiler/z/codegen/J9TreeEvaluator.cpp
```diff
@@ -211,6 +211,355 @@ doubleMaxMinHelper(TR::Node *node, TR::CodeGenerator *cg, bool isMaxOp)
...
+      generateS390BranchInstruction(cg, TR::InstOpCode::BRC, TR::InstOpCode::COND_BRC, node, labelResultCheckDone);
+
+      generateS390LabelInstruction(cg, TR::InstOpCode::LABEL, node, labelResultGT100);
+      cg->generateDebugCounter("indexOfString/result/above-100", 1, TR::DebugCounter::Cheap);
+
+      generateS390LabelInstruction(cg, TR::InstOpCode::LABEL, node, labelResultCheckDone);
+      }
+   // END debug counters
+
+   /************************************** Done ******************************************/
```
fjeremic: I don't think these comments are particularly useful given we have scoped blocks encapsulating the debug counter code and we have a done label right below.

runtime/compiler/z/codegen/J9TreeEvaluator.cpp
```diff
@@ -211,6 +211,355 @@ doubleMaxMinHelper(TR::Node *node, TR::CodeGenerator *cg, bool isMaxOp)
...
+
+   cursor = generateS390LabelInstruction(cg, TR::InstOpCode::LABEL, node, labelDone, regDeps);
+   cursor->setEndInternalControlFlow();
```
fjeremic: Similar comment to above regarding ICF and labels.

runtime/compiler/z/codegen/J9TreeEvaluator.cpp
NigelYiboYu: OK. Changed to `labelStart->setStartInternalControlFlow()`.

test/functional/Java8andUp/src/org/openj9/test/string/StringIndexOfString.java
NigelYiboYu: I changed this string to ``public static`. But static folding isn't the focus of this test nor does it affect the outcome, right?

test/functional/Java8andUp/src/org/openj9/test/string/StringIndexOfString.java
NigelYiboYu: Changed `generateInputs()`  to a static block.
This is a sanity test and all test functions annotated by `@Test` will be executed only once. Hence, this `generateInputs` function is only invoked once. So static block or not makes no difference here.

test/functional/Java8andUp/src/org/openj9/test/string/StringIndexOfString.java
fjeremic: Like all other things in the JIT it will not affect semantics of the user program, however static final field folding may affect what you want to test here, which is a codegen path that the JIT will generate.

test/functional/Java8andUp/src/org/openj9/test/string/StringIndexOfString.java
NigelYiboYu: Fixed.

runtime/compiler/z/codegen/J9TreeEvaluator.cpp
NigelYiboYu: Fixed.

fjeremic:

PR #3188, opened 2018-10-09
fix invalid assertion check in test

LinHu2016:
- collection Usage of the same memoryPool might be different
	between getCollectionUsage() calls.

### DIFF
--- a/test/functional/JLM_Tests/src/org/openj9/test/java/lang/management/TestManagementFactory.java
+++ b/test/functional/JLM_Tests/src/org/openj9/test/java/lang/management/TestManagementFactory.java
@@ -833,7 +833,7 @@ public class TestManagementFactory {
 			MemoryUsage directUsage = realBean.getCollectionUsage();
 			validateMemoryUsage(proxyUsage);
 			validateMemoryUsage(directUsage);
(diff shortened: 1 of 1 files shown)


test/functional/JLM_Tests/src/org/openj9/test/java/lang/management/TestManagementFactory.java
```diff
@@ -833,7 +833,8 @@ public void testMemoryPoolMXBeanProxy() {
 			MemoryUsage directUsage = realBean.getCollectionUsage();
 			validateMemoryUsage(proxyUsage);
 			validateMemoryUsage(directUsage);
-			AssertJUnit.assertEquals(directUsage, proxyUsage);
+			/* directUsage and proxyUsage could be different, if gc happened between getCollectionUsage() calls*/
+//			AssertJUnit.assertEquals(directUsage, proxyUsage);
```
smlambert: Are you keeping it in the code as a comment to come back to it?  If it is invalid, do you not simply want to remove it?

LinHu2016: removed the assertion check and leave comment for a reminder

PR #3190, opened 2018-10-09
Use properties file to control whether pointer classes are generated

keithc-ca:
Both pointer and structure classes must be included in j9ddr.jar to continue supporting debug extensions (which must be compiled using j9ddr.jar).

With this change, `J9DDRClassLoader` will use pointer classes in j9ddr.jar unless it contains `pointers/generated/dynamic.properties` which defines `generate.pointers` with the value `true`, in which case, pointer classes will be generated dynamically based on the blob in the related core file.

The related pull requests will be updated accordingly:
https://github.com/ibmruntimes/openj9-openjdk-jdk8/pull/117
https://github.com/ibmruntimes/openj9-openjdk-jdk11/pull/28


pshipton:
jenkins test sanity zlinux

PR #3195, opened 2018-10-09
Add support for atomic symbol reduction for Z

fjeremic:
In addition we make use of the new `supportsNonHelper` API defined at
the OMR level to remove platform checks in the common code generator
and delegate the question to the platform specific code generators as
to whether we support inlining of a particular intrinsic.

### DIFF
--- a/runtime/compiler/optimizer/J9RecognizedCallTransformer.cpp
+++ b/runtime/compiler/optimizer/J9RecognizedCallTransformer.cpp
@@ -138,11 +138,17 @@ bool J9::RecognizedCallTransformer::isInlineable(TR::TreeTop* treetop)
    switch(node->getSymbol()->castToMethodSymbol()->getMandatoryRecognizedMethod())
       {
       case TR::sun_misc_Unsafe_getAndAddInt:
+         return !comp()->getOption(TR_DisableUnsafe) && !comp()->compileRelocatableCode() && !TR::Compiler->om.canGenerateArraylets() && 
+            cg()->supportsNonHelper(TR::SymbolReferenceTable::atomicFetchAndAddSymbol);
       case TR::sun_misc_Unsafe_getAndSetInt:
-         return TR::Compiler->target.cpu.isX86() && !comp()->getOption(TR_DisableUnsafe) && !comp()->compileRelocatableCode() && !TR::Compiler->om.canGenerateArraylets();
+         return !comp()->getOption(TR_DisableUnsafe) && !comp()->compileRelocatableCode() && !TR::Compiler->om.canGenerateArraylets() && 
+            cg()->supportsNonHelper(TR::SymbolReferenceTable::atomicSwapSymbol);
       case TR::sun_misc_Unsafe_getAndAddLong:
+         return !comp()->getOption(TR_DisableUnsafe) && !comp()->compileRelocatableCode() && !TR::Compiler->om.canGenerateArraylets() && TR::Compiler->target.is64Bit() && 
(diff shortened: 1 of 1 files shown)


fjeremic:
Leaving as WIP until I've done some more internal testing. Reviewers please feel to proceed with the reviews however. I'd like to request one from @0dvictor here.

runtime/compiler/optimizer/J9RecognizedCallTransformer.cpp
```diff
@@ -170,16 +172,16 @@ void J9::RecognizedCallTransformer::transform(TR::TreeTop* treetop)
...
-         processUnsafeAtomicCall(treetop, node, TR::SymbolReferenceTable::atomicFetchAndAdd32BitSymbol);
+         processUnsafeAtomicCall(treetop, node, TR::SymbolReferenceTable::atomicFetchAndAddSymbol);
          break;
       case TR::sun_misc_Unsafe_getAndSetInt:
-         processUnsafeAtomicCall(treetop, node, TR::SymbolReferenceTable::atomicSwap32BitSymbol);
+         processUnsafeAtomicCall(treetop, node, TR::SymbolReferenceTable::atomicSwapSymbol);
          break;
       case TR::sun_misc_Unsafe_getAndAddLong:
-         processUnsafeAtomicCall(treetop, node, TR::SymbolReferenceTable::atomicFetchAndAdd64BitSymbol);
+         processUnsafeAtomicCall(treetop, node, TR::SymbolReferenceTable::atomicFetchAndAddSymbol);
```
0dvictor: Why not common this entry with `TR::sun_misc_Unsafe_getAndAddInt`?

runtime/compiler/x/codegen/J9CodeGenerator.cpp
```diff
@@ -403,6 +403,28 @@ J9::X86::CodeGenerator::suppressInliningOfRecognizedMethod(TR::RecognizedMethod
...
+J9::X86::CodeGenerator::supportsNonHelper(TR::SymbolReferenceTable::CommonNonhelperSymbol symbol)
+   {
+   bool result = false;
+
+   switch (symbol)
+      {
+      case TR::SymbolReferenceTable::atomicFetchAndAddSymbol:
+      case TR::SymbolReferenceTable::atomicSwapSymbol:
+         {
+         if (!self()->comp()->getOption(TR_DisableUnsafe) && !self()->comp()->compileRelocatableCode() && !TR::Compiler->om.canGenerateArraylets())
```
0dvictor: IMO, these checks shouldn't belong to CodeGen. These conditions are to determine whether a Unsafe method should be transformed into NonHelper; they are NOT whether such NonHelper is supported. For example, other intrinsic methods may be transformed into the NonHelper but don't need checks for any of the tree condition.
In addition, the CodeGen actually do support for the two NonHelper, it is the ILGen or Optimizer that cannot transform the Unsafe calls to NonHelper under certain modes.

runtime/compiler/optimizer/J9RecognizedCallTransformer.cpp
fjeremic: Fixed in 24ee533

runtime/compiler/x/codegen/J9CodeGenerator.cpp
fjeremic: I think I agree. I've greatly simplified things in 24ee533

fjeremic:
I've finished my testing on my side and convinced myself this change is correct. Off to reviewers / committers next. @andrewcraik could you please take a look?

andrewcraik:
Jenkins test sanity xlinux,zlinux,win,plinux jdk11

Issue #3062, opened 2018-09-28
Investigate limiting System.identityHashcode() to positive numbers


DanHeidinga:
OpenJ9 allows both positive and negative identity hashcodes.  This is problematic for programs that incorrectly assume hashcodes can only be positive.

Investigate adding an option to limit hashcode to be positive and determine the performance overhead of this.

Note, any such option would limit the range of hashcodes and may lead to more collisions.

charliegracie:
Are there open/closed issues where consumers of OpenJ9 were having this issue?

DanHeidinga:
The likely place to hack for this is https://github.com/eclipse/openj9/blob/2f295ea91cdd97ff37eb5d164fc63014ea718bc7/runtime/oti/ObjectHash.hpp#L177

Changing it here will ensure the option + mask is only rechecked until the object is moved.  All the fast paths that read the hashcode from moved objects shouldn't require any modification.

Will attempt to add an -XX option and investigate the performance implications of this change.

vijaysun-omr:
Maybe if the result of the -XX option is stored in a private static final boolean variable and if that variable is checked in the identityHashCode at the spot @DanHeidinga pointed out, then the JIT should be able to fold away the extra test at compile time via it's usual static final optimizations.

vijaysun-omr:
I assume you were thinking of an -XX option to work around this particular migration issue in terms of different JCL behavior. One question is whether you prefer different -XX options for different JCL issues such as this one (assuming there are more of them we will find down the line) or have one option that changes all the JCL behaviours needed to make migration easier.

DanHeidinga:
@vijaysun-omr I've looked for how to push the option as low in the code as possible - down to the native hash operation -  to ensure there is only 1 place that needs to be modified and that there's never a chance of an inconsistent hashcode being returned.

> One question is whether you prefer different -XX options for different JCL issues such as this one (assuming there are more of them we will find down the line) or have one option that changes all the JCL behaviours needed to make migration easier.

That's an interesting idea.  I think we should start with a single option. If we end up with multiple migration related options, they can be grouped later under a single -XX:EasyMigration (or similar) option.

vijaysun-omr:
Thanks @DanHeidinga I am in agreement with your approach to wait for the "second" migration related option before considering grouping under one option. 

Is there a PR for the change to the native hash ?

DanHeidinga:
[...]

PR #3198, opened 2018-10-09
Add option to limit identity hash code to positive values

gacholio:
Add command line option "-XX:[+|-]PositiveIdentityHash" to
enable/disable forcing identity hash code to be a positive value.

Fixes: #3062


gacholio:

pshipton:
@gacholio please open an issue at https://github.com/eclipse/openj9-docs for the doc change.

vijaysun-omr:
Thanks @DanHeidinga and @gacholio for this change. 

I think a small unit test that computes indentityHashCode many times should tell us the relative cost of adding the option check on this code path. The baseline would be a build without this change. 

I can also get Liberty tried with this change just to make sure it does not regress. I believe some code in a Lucene test we had exercised identity hash codes but I cannot be sure but I'll ask the team members I was working with at that time if we still have the test.

vijaysun-omr:
I asked and we don't have the Lucene testing I was thinking of anymore. So I will just get this evaluated on Liberty.

gacholio:
-Xjit results (10x the iteration count of -Xint):

Before:

1000000000 iterations took 9923 ms
1000000000 iterations took 9785 ms
1000000000 iterations took 9789 ms

After:

1000000000 iterations took 10815 ms
1000000000 iterations took 10812 ms
1000000000 iterations took 10865 ms

Seems consistently slower.

vijaysun-omr:
I assume this is comparing a build without the option vs a build with the option (but the option is not specified on the command line). ?

gacholio:
@vijaysun-omr The JIT seems to recognize and replaced identityHashCode, but it looks like that's guarded by an option that is not set, and the replacement code is unimplmeneted. Can you confirm?

I can't explain why the JIT would be slower (the JITs were not identical in the builds, but they were no more than a day different).

gacholio:
Yes, the "before" is a build from today and the "after" is a build from yesterday with this PR added. The new option was not enabled.

vijaysun-omr:
I've never seen identityHashCode take more than a couple of percent in any profile really; it's much lower than that in the profiles of benchmarks we work with usually. So I am thinking this is acceptable since the overhead of the extra check you added would be < 0.2% in those benchmarks.

gacholio:
There is a Fast JNI implementation in use by the JIT (and that will get the change from this PR).

vijaysun-omr:
Is it possible to swap the JIT module between these 2 builds and see if we can get a run with literally the same JIT ?

gacholio:
@vijaysun-omr For the Liberty bench, it would be good to know the cost of enabling the new option, not just the cost of the new code.

gacholio:
Copying the "before" JIT into the "after" VM:

1000000000 iterations took 10844 ms
1000000000 iterations took 10856 ms
1000000000 iterations took 10876 ms

yields essentially identical results.

DanHeidinga:
@vijaysun-omr @gacholio I'm OK with merging this now so it makes the 0.11.0 release and we can update with the Liberty results when they're available.

Any objections?

gacholio:
I guess I can just say "unknowable performance impact" in the doc.

vijaysun-omr:
Given that I don't expect any measurable regression in Liberty I have no issue merging this now without waiting for the Liberty result.

vijaysun-omr:
@gacholio yes we don't do anything special for the identityHashCode computation in the JIT as far as I know.

gacholio:
Doc issue: https://github.com/eclipse/openj9-docs/issues/109

PR #3199, opened 2018-10-09
Use ITEMTYPE(item) to check for orphan classes

hangshao0:
Using RcLinkedListImpl::_isOrphan to check for orphan classes is not
reliable. There is a small time hole in
SH_ROMClassManagerImpl::storeNew() where orphan is added to the hash
table by hllTableUpdate() but _isOrphan is not set to true yet. So use
ITEMTYPE(item) to check for orphan classes instead.


pshipton:
Is there any point in keeping and setting the _isOrphan field?

pshipton:
Ok, it can be done separately.

PR #3208, opened 2018-10-10
CopyForwardHybrid

LinHu2016:
CopyForwardHybrid is hybrid mode of CopyForward Scheme that 
 GC can run both copyforward and markcompact on the collection set
 simultaneously (for balanced PGC only). During precollection, 
 GC can reserve some regions for markcompact only(non evacuated 
 regions -- such as jni critical regions) and copyforward for 
 the rest of collection set. CopyForwardHybrid intent to avoid 
 markcompact PGC due to nonEvacuated regions in collection set,
 in order to minimize the longer gc pause time and affection to
 concurrent gc.

 - new java Option -XXgc:tarokEnableCopyForwardMarkCompactHybrid for
enabling
 CopyForwardHybrid mode, default is disabled.
 - new java Option
-XXgc:fvtest_forceCopyForwardMarkCompactHybridRatio=nn,
 (1 <=nn<= 100) nn is the percentage of collection set to be reserved
for markcompact(for margin testing CopyForwardHybrid).
 - report copyforward is running hybrid mode in verbose gc
 - allow jni critical eden regions in nursery collectionSet for
copyforward
  - _noEvacuation flag in HeapRegionDescriptorVLHGC, set it true 
  if the regions can not be evacuated. decide which region can not 
  be evacuated in preProcessRegions.
  - pull scan work from both scanCache and workPacket in CompleteScan
  - synchronize working threads with both scanCache and workPacket
  - if live object in non evacurated regions mark map instead of
copyforward.
  - exception case 1: if abort case happens, stop to pull scan work
  from workPacket
  - exception case 2: if workPacket overflow during CompleteScan
  set it to abort case to handle overflow
 
### DIFF
--- a/runtime/gc_modron_startup/mmparseXXgc.cpp
+++ b/runtime/gc_modron_startup/mmparseXXgc.cpp
@@ -632,6 +632,11 @@ gcParseXXgcArguments(J9JavaVM *vm, char *optArg)
 
 			continue;
 		}
+		if (try_scan(&scan_start, "tarokEnableCopyForwardMarkCompactHybrid")) {
+			extensions->tarokEnableCopyForwardHybrid = true;
+			continue;
+		}
+
 #endif /* defined (J9VM_GC_VLHGC) */
 
 		if(try_scan(&scan_start, "packetListLockSplit=")) {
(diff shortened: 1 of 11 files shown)


dmitripivkine:

PR #3218, opened 2018-10-10
Add missing conversion for load generated by UnsafeFastPath

liqunl:
Convert the result of load if the result type doesn't match the Unsafe
call's return type. The conversion is needed for accessing arrays whose
element type is narrower than 32 bits.

### DIFF
--- a/runtime/compiler/optimizer/UnsafeFastPath.cpp
+++ b/runtime/compiler/optimizer/UnsafeFastPath.cpp
@@ -538,12 +538,12 @@ int32_t TR_UnsafeFastPath::perform()
                      }
                   else
                      {
-                     if (isByIndex && value->getDataType() != type)
+                     if (value->getDataType() != type)
                         {
                         TR::ILOpCodes conversionOpCode = TR::ILOpCode::getProperConversion(value->getDataType(), type, true);
 
                         // Sanity check for future modifications
-                        TR_ASSERT(conversionOpCode != TR::BadILOp, "Could not get proper conversion for a JITHelpers byIndex operation.\n");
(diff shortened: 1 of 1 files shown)


liqunl:
@fjeremic @andrewcraik Can I have a review?

runtime/compiler/optimizer/UnsafeFastPath.cpp
```diff
@@ -560,7 +560,7 @@ int32_t TR_UnsafeFastPath::perform()
                else
                   {
                   //This is a load
-                  if (isByIndex && node->getDataType() != type)
+                  if (node->getDataType() != type)
```
fjeremic: There seems to be an assert below guarding for future modifications. Are you certain we won't hit this assert? Note it's not a `FATAL` assert so it won't fire by default (outside of debug builds).

liqunl: Not for Unsafe methods. Is there similar problem to JITHelpers methods like `getByteFromArray`? The method return `byte` but in trees we use `icallx`, so a conversion is needed on the load.

liqunl: Now that you mention, I'll update the assertion message to not limit to JITHelpers.

liqunl:
Added change to also fix store.

runtime/compiler/optimizer/UnsafeFastPath.cpp
liqunl: @fjeremic Have my comments answered your questions? Any more concerns?

andrewcraik:
CI should be sufficient - merging

PR #3219, opened 2018-10-10
Fix incorrect recognized method name in VarHandleTransformer

liqunl:
A few methods got wrong method name in the recognized methods table.


liqunl:
@andrewcraik Please review

PR #3225, opened 2018-10-11
Fix bug in canAllocateInlineClass

mpirvu:
Due to a bug canAllocateInlineClass returns true only if the class is
abstract or interface. It should be the other way around.


mpirvu:
@ymanton Please review and merge. Thanks!

Issue #3226, opened 2018-10-11
Enable idle tuning options by default when running in container


ashu-mehra:
OpenJ9 docker images produced at AdoptOpenJDK already enable `-XX:+IdleTuningGcOnIdle` and `-XX:+IdleTuningCompactOnIdle`. See https://github.com/AdoptOpenJDK/openjdk-docker/blob/master/8/jdk/ubuntu/Dockerfile.openj9.releases.full.
It would be beneficial if we enable these options at runtime when OpenJ9 is running in container so that any product that uses OpenJ9 but does not use the docker images from Adopt, also gets the benefits of these options without the need to explicitly add these options in their Dockerfile.

@mpirvu @dinogun

ashu-mehra:
Doc issue: https://github.com/eclipse/openj9-docs/issues/112

DanHeidinga:
@ashu-mehra Can you expand on:
* the rationale for this, 
* the differences users might see, 
* whether these options are supported on all platforms or not, 
* any limitations
* any performance impacts to either startup or throughput when this is enabled

ashu-mehra:
> the rationale for this,

I added my understanding of why we want to do this in the first comment i.e. `It would be beneficial if we enable these options at runtime when OpenJ9 is running in container so that any product that uses OpenJ9 but does not use the docker images from Adopt, also gets the benefits of these options without the need to explicitly add these options in their Dockerfile.`
Generally when someone includes OpenJ9 in their Dockerfile, they are unlikely to enable these options unless they are aware of them. So this PR would be useful in such scenarios.
@mpirvu do you want to add anything more to it?

 > the differences users might see,

I don't expect any differences observable by user other than during idle state when the physical memory footprint may drop down.

 > whether these options are supported on all platforms or not,

AFAIK these options are supported on all platforms except windows, but this PR only enables them when OpenJ9 in running in container which currently covers Linux platforms only.

> any limitations

Same limitations that apply to these options (eg only work with gencon policy), nothing new added by this PR.

> any performance impacts to either startup or throughput when this is enabled

I don't think any throughput impact was observed in earlier performance runs done when these options were added. That should hold true if we enable these options by default.
For startup, Marius has shared his script for gathering Liberty startup numbers (as in https://github.com/eclipse/openj9/issues/2263#issuecomment-427010152). I will use that and will post the startup numbers here.

ashu-mehra:
[...]

PR #3228, opened 2018-10-11
Enable options for idle tuning when running in container

ashu-mehra:
OpenJ9 docker images produced at AdoptOpenJDK already enable
-XX:+IdleTuningGcOnIdle and -XX:+IdleTuningCompactOnIdle.
It would be beneficial if we enable these options at runtime by detecting
if OpenJ9 is running in container, so that any product that uses OpenJ9
but does not use Adopt docker images also gets the benefits of these
options without the need to explicitly add these options in their Dockerfile.

Doc issue eclipse/openj9-docs#112
Fixes: #3226

### DIFF
--- a/runtime/vm/jvminit.c
+++ b/runtime/vm/jvminit.c
@@ -1603,6 +1603,7 @@ IDATA VMInitStages(J9JavaVM *vm, IDATA stage, void* reserved) {
 	BOOLEAN lockwordWhat = FALSE;
 	UDATA rc = 0;
 	PORT_ACCESS_FROM_JAVAVM(vm);
+	OMRPORT_ACCESS_FROM_J9PORT(PORTLIB);
 
 	switch(stage) {
 		case PORT_LIBRARY_GUARANTEED :
@@ -1823,19 +1824,40 @@ IDATA VMInitStages(J9JavaVM *vm, IDATA stage, void* reserved) {
 				IDATA argIndexCompactOnIdleDisable = FIND_AND_CONSUME_ARG(EXACT_MATCH, VMOPT_XXIDLETUNINGCOMPACTONIDLEDISABLE, NULL);
 				IDATA argIndexIgnoreUnrecognizedOptionsEnable = FIND_AND_CONSUME_ARG(EXACT_MATCH, VMOPT_XXIDLETUNINGIGNOREUNRECOGNIZEDOPTIONSENABLE, NULL);
 				IDATA argIndexIgnoreUnrecognizedOptionsDisable = FIND_AND_CONSUME_ARG(EXACT_MATCH, VMOPT_XXIDLETUNINGIGNOREUNRECOGNIZEDOPTIONSDISABLE, NULL);
+				BOOLEAN enableGcOnIdle = FALSE;
+				BOOLEAN inContainer = FALSE;
+				int32_t errorCode = 0;
+				
+				inContainer = omrsysinfo_is_running_in_container(&errorCode);
 
 				/* 
-				 * idle heap tuning is enabled only if -XX:+IdleTuningGcOnIdle is set 
-				 * or if java version is 9 or above and -Xtune:virtualized is set as VM option
+				 * GcOnIdle is enabled only if:
+				 * 1. -XX:+IdleTuningGcOnIdle is set, or
+				 * 2. running in container, or
+				 * 3. if java version is 9 or above and -Xtune:virtualized is set as VM option
 				 */
-				if ((argIndexGcOnIdleEnable > argIndexGcOnIdleDisable) 
-				|| ((J2SE_VERSION(vm) >= J2SE_19) && (argIndexGcOnIdleDisable == -1) && J9_ARE_ANY_BITS_SET(vm->runtimeFlags, J9_RUNTIME_TUNE_VIRTUALIZED))
-				) {
+				if (argIndexGcOnIdleEnable > argIndexGcOnIdleDisable) {
+					enableGcOnIdle = TRUE;
+				} else if (-1 == argIndexGcOnIdleDisable) {
+					if (inContainer
+					|| ((J2SE_VERSION(vm) >= J2SE_19) && J9_ARE_ANY_BITS_SET(vm->runtimeFlags, J9_RUNTIME_TUNE_VIRTUALIZED))
+					) {
+						enableGcOnIdle = TRUE;
+					}
+				}
+				if (TRUE == enableGcOnIdle) {
 					vm->vmRuntimeStateListener.idleTuningFlags |= (UDATA)J9_IDLE_TUNING_GC_ON_IDLE;
 				} else {
 					vm->vmRuntimeStateListener.idleTuningFlags &= ~(UDATA)J9_IDLE_TUNING_GC_ON_IDLE;
 				}
-				if (argIndexCompactOnIdleEnable > argIndexCompactOnIdleDisable) {
+				/*
+				 * CompactOnIdle is enabled only if:
+				 * 1. -XX:+IdleTuningCompactOnIdle is set, or
+				 * 2. running in container
+				 */
+				if ((argIndexCompactOnIdleEnable > argIndexCompactOnIdleDisable)
+				|| ((-1 == argIndexCompactOnIdleDisable) && inContainer)
+				) {
 					vm->vmRuntimeStateListener.idleTuningFlags |= (UDATA)J9_IDLE_TUNING_COMPACT_ON_IDLE;
 				} else {
 					vm->vmRuntimeStateListener.idleTuningFlags &= ~(UDATA)J9_IDLE_TUNING_COMPACT_ON_IDLE;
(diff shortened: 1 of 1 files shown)


ashu-mehra:
fyi - @mpirvu 
@DanHeidinga please review.

DanHeidinga:
@ashu-mehra I've asked some questions about the "big picture" on this change in the https://github.com/eclipse/openj9/issues/3226#issuecomment-428956355

I'd like to see those questions addressed before reviewing this change.  Note, that we're on a very tight timeline for the 0.11.0 release and we may chose to delay this until after that release is out.

pshipton:
Using this PR to do some unrelated aix build testing.
jenkins compile aix jdk11

pshipton:
jenkins compile aix jdk11

pshipton:
As its past the feature complete date for the 0.11.0 milestone, targeting this to the next release.

ashu-mehra:
@DanHeidinga can you please review it.

pshipton:
jenkins test sanity xlinux,win,aix jdk11

runtime/vm/jvminit.c
```diff
@@ -1818,19 +1819,39 @@ IDATA VMInitStages(J9JavaVM *vm, IDATA stage, void* reserved) {
...
 					vm->vmRuntimeStateListener.idleTuningFlags |= (UDATA)J9_IDLE_TUNING_GC_ON_IDLE;
 				} else {
 					vm->vmRuntimeStateListener.idleTuningFlags &= ~(UDATA)J9_IDLE_TUNING_GC_ON_IDLE;
 				}
-				if (argIndexCompactOnIdleEnable > argIndexCompactOnIdleDisable) {
+				/*
+				 * CompactOnIdle is enabled only if:
+				 * 1. -XX:+IdleTuningCompactOnIdle is set, or
+				 * 2. running in container
+				 */
```
DanHeidinga: Why different conditions for CompactOnIdle vs IdleTuningGcOnIdle?

runtime/vm/jvminit.c
```diff
@@ -1818,19 +1819,39 @@ IDATA VMInitStages(J9JavaVM *vm, IDATA stage, void* reserved) {
 				IDATA argIndexCompactOnIdleDisable = FIND_AND_CONSUME_ARG(EXACT_MATCH, VMOPT_XXIDLETUNINGCOMPACTONIDLEDISABLE, NULL);
 				IDATA argIndexIgnoreUnrecognizedOptionsEnable = FIND_AND_CONSUME_ARG(EXACT_MATCH, VMOPT_XXIDLETUNINGIGNOREUNRECOGNIZEDOPTIONSENABLE, NULL);
 				IDATA argIndexIgnoreUnrecognizedOptionsDisable = FIND_AND_CONSUME_ARG(EXACT_MATCH, VMOPT_XXIDLETUNINGIGNOREUNRECOGNIZEDOPTIONSDISABLE, NULL);
+				BOOLEAN enableGcOnIdle = FALSE;
+				BOOLEAN inContainer = FALSE;
+				
+				omrsysinfo_is_running_in_container(&inContainer);
```
DanHeidinga: Does this have a return code that needs to be checked?

runtime/vm/jvminit.c
```diff
@@ -1818,19 +1819,39 @@ IDATA VMInitStages(J9JavaVM *vm, IDATA stage, void* reserved) {
...
+				 * 3. if java version is 9 or above and -Xtune:virtualized is set as VM option
 				 */
-				if ((argIndexGcOnIdleEnable > argIndexGcOnIdleDisable) 
-				|| ((J2SE_VERSION(vm) >= J2SE_19) && (argIndexGcOnIdleDisable == -1) && J9_ARE_ANY_BITS_SET(vm->runtimeFlags, J9_RUNTIME_TUNE_VIRTUALIZED))
-				) {
+				if (argIndexGcOnIdleEnable > argIndexGcOnIdleDisable) {
+					enableGcOnIdle = TRUE;
+				} else if (-1 == argIndexGcOnIdleDisable) {
+					if (inContainer
+						|| ((J2SE_VERSION(vm) >= J2SE_19) && J9_ARE_ANY_BITS_SET(vm->runtimeFlags, J9_RUNTIME_TUNE_VIRTUALIZED))
```
DanHeidinga: It's a nitpick but can you format the `if` block like below as it makes the conditions block line up?
```
	if (inContainer
	|| ((J2SE_VERSION(vm) >= J2SE_19) && J9_ARE_ANY_BITS_SET(vm->runtimeFlags, J9_RUNTIME_TUNE_VIRTUALIZED))
	) {
```

runtime/vm/jvminit.c
```diff
@@ -1818,19 +1819,39 @@ IDATA VMInitStages(J9JavaVM *vm, IDATA stage, void* reserved) {
...
 					vm->vmRuntimeStateListener.idleTuningFlags &= ~(UDATA)J9_IDLE_TUNING_GC_ON_IDLE;
 				}
-				if (argIndexCompactOnIdleEnable > argIndexCompactOnIdleDisable) {
+				/*
+				 * CompactOnIdle is enabled only if:
+				 * 1. -XX:+IdleTuningCompactOnIdle is set, or
+				 * 2. running in container
+				 */
+				if ((argIndexCompactOnIdleEnable > argIndexCompactOnIdleDisable)
+					|| ((-1 == argIndexCompactOnIdleDisable) && inContainer)
```
DanHeidinga: same nitpick as above

DanHeidinga:
Some minor formatting concerns and a question regarding the difference in how the defaults are set need to be addressed before merging

pshipton:
There is also a compiler error in jvminit.c, the following is from the AIX build.
`09:33:50 "jvminit.c", line 1830.68: 1506-280 (W) Function argument assignment between types "int*" and "unsigned int*" is not allowed.`

DanHeidinga:
Compile error is likely due to the change in signature for `omrsysinfo_is_running_in_container` in OMR

pshipton:
@ashu-mehra will you be addressing the review comments?

ashu-mehra:
Sorry, I was away for past few days. I will soon address the comments.

runtime/vm/jvminit.c
ashu-mehra: Only missing condition for CompactOnIdle is presence of -Xtune:virtualized option. I am not sure why we did not add that condition for CompactOnIdle. PR where IdleTuningGcOnIdle was enabled for -Xtune:virtualized is:  https://github.com/eclipse/openj9/pull/125.
But I think that should be considered separately from this PR.

ashu-mehra:
@DanHeidinga Updated the change with the review comments

runtime/vm/jvminit.c
pshipton: @ashu-mehra if it is to be considered separately, there should be an issue for it.

runtime/vm/jvminit.c
```diff
@@ -1823,19 +1824,41 @@ IDATA VMInitStages(J9JavaVM *vm, IDATA stage, void* reserved) {
 				IDATA argIndexCompactOnIdleDisable = FIND_AND_CONSUME_ARG(EXACT_MATCH, VMOPT_XXIDLETUNINGCOMPACTONIDLEDISABLE, NULL);
 				IDATA argIndexIgnoreUnrecognizedOptionsEnable = FIND_AND_CONSUME_ARG(EXACT_MATCH, VMOPT_XXIDLETUNINGIGNOREUNRECOGNIZEDOPTIONSENABLE, NULL);
 				IDATA argIndexIgnoreUnrecognizedOptionsDisable = FIND_AND_CONSUME_ARG(EXACT_MATCH, VMOPT_XXIDLETUNINGIGNOREUNRECOGNIZEDOPTIONSDISABLE, NULL);
+				BOOLEAN enableGcOnIdle = FALSE;
+				BOOLEAN inContainer = FALSE;
+				int32_t errorCode = 0;
+				
+				inContainer = omrsysinfo_is_running_in_container(&errorCode);
+				inContainer = ((0 == errorCode) && (TRUE == inContainer));
```
pshipton: This seems reversed. The return value is in32_t and zero for success. The parameter is a BOOLEAN set to TRUE when running in a container.

ashu-mehra: @pshipton Yeah it looks reversed, but it is as per the api `omrsysinfo_is_running_in_container`. See https://github.com/eclipse/omr/blob/85e2a6fd463e232126565956ca30cfe3340b483e/port/common/omrsysinfo.c#L962

runtime/vm/jvminit.c
ashu-mehra: @pshipton ok, I will open an issue to address this.

runtime/vm/jvminit.c
pshipton: Guess the api changed at some point and I had an old copy checked out.

pshipton:
@DanHeidinga any further concerns?

runtime/vm/jvminit.c
ashu-mehra: Issue to address this concern #4004

runtime/vm/jvminit.c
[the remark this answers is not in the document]
ashu-mehra: It is not clear from the API what is the return value when error happens. During error condition, it could return TRUE or FALSE. That is why the explicit check for error code.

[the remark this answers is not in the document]
DanHeidinga: I've checked all the implementations and opened a PR at OMR to clarify the functions contract: https://github.com/eclipse/omr/pull/3380

The function is most useful it returns a consistent answer (FALSE) on an error and that's what it does.

I think we can drop checking the error code here.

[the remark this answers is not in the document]
ashu-mehra: Ok, thanks for clarifying the contract. I have updated the change to remove the error check.

DanHeidinga:
Jenkins test all zlinux jdk11

DanHeidinga:

DanHeidinga:
Merging as is.  The failure in the extended build is unrelated to this PR

PR #3229, opened 2018-10-11
Fix resource access checks in getResource and getResourceAsStream

pdbain-ibm:
Allow getResource et al. to search the current module if the package in
question is accessible by the caller's module.

Fixes https://github.com/eclipse/openj9/issues/3136

### DIFF
--- a/jcl/src/java.base/share/classes/java/lang/Class.java
+++ b/jcl/src/java.base/share/classes/java/lang/Class.java
@@ -1911,10 +1911,10 @@ String getPackageName() {
 }
 
 /**
- * Answers a read-only stream on the contents of the
+ * Answers a URL referring to the
  * resource specified by resName. The mapping between
- * the resource name and the stream is managed by the
- * class' class loader.
+ * the resource name and the URL is managed by the
+ * class's class loader.
  *
  * @param		resName 	the name of the resource.
  * @return		a stream on the resource.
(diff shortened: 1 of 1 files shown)


jcl/src/java.base/share/classes/java/lang/Class.java
```diff
@@ -1964,25 +1966,66 @@ public URL getResource(String resName) {
...
+ * Look up the resource in the module if the module is named 
+ * and is the same module as the caller or the package is open to the caller.
+ * The default package (i.e. resources at the root of the module) is considered open.
+ * 
+ * @param absoluteResName name of resource, including package
+ * @param thisModule module of the current class
+ * @param callerClass class of method calling getResource() or getResourceAsStream()
+ * @return true if modular lookup should be used.
+ */
+private boolean useModularSearch(String absoluteResName, Module thisModule, Class callerClass) {
```
DanHeidinga: `Class callerClass` -> `Class<?> callerClass`

jcl/src/java.base/share/classes/java/lang/Class.java
```diff
@@ -1964,25 +1966,66 @@ public URL getResource(String resName) {
...
+	boolean visible = false;
+
+	if (thisModule.isNamed()) {
+		final Module callerModule = callerClass.getModule();
+		visible = (thisModule == callerModule);
+		if (!visible) {
+			if (absoluteResName.endsWith(".class")) {
```
DanHeidinga: To match the pattern of code above, use:

```java
visible = absoluteResName.endsWith(".class");
if (!visible) {.....
```

jcl/src/java.base/share/classes/java/lang/Class.java
```diff
@@ -1964,25 +1966,66 @@ public URL getResource(String resName) {
...
+				visible = true;
+			} else {
+				// extract the package name
+				int lastSlash = absoluteResName.lastIndexOf('/');
+				if (-1 == lastSlash) { // no package name
+					visible = true;
+				} else {
+					String result = absoluteResName.substring(0, lastSlash).replace('/', '.');
+					visible = thisModule.isOpen(result, callerModule);
```
DanHeidinga: Does `isOpen` cover all cases or should there also be an `isExported` check?

pdbain-ibm: From the javadoc for `isExported()`:
> A package that is open to the given module is considered exported to that module at run-time and so this method returns true if the package is open to the given module.

so `isExported()` would be redundant.

jcl/src/java.base/share/classes/java/lang/Class.java
pdbain-ibm: Good idea. Done.

jcl/src/java.base/share/classes/java/lang/Class.java
pdbain-ibm: Right.  Done.
I will push the updates when my personal build completes.

jcl/src/java.base/share/classes/java/lang/Class.java
```diff
@@ -1927,31 +1927,33 @@ String getPackageName() {
...
 		}
-	} else
-/*[ENDIF] Sidecar19-SE */
+	}
+	if (null == result)
+		/*[ENDIF] Sidecar19-SE */
 	{
 		if (loader == ClassLoader.bootstrapClassLoader) {
-			return ClassLoader.getSystemResource(absoluteResName);
+			result =  ClassLoader.getSystemResource(absoluteResName);
```
DanHeidinga: nitpick - there's some odd double spacing here and throughout the new code.  Can you address that before the merge?

pdbain-ibm: Fixed in 5 locations.

DanHeidinga:
Jenkins test sanity plinux jdk8,11

DanHeidinga:
Jenkins test extended xlinux jdk10

DanHeidinga:

pdbain-ibm:
@DanHeidinga all tests passed.

PR #3230, opened 2018-10-11
Remove unused compile_ddr.xml file from DDR_VM debugtools

acrowthe:
Remove unused compilation file from DDR_VM.

Original message and reason for changing file:
Remove unused dependency from compile_ddr.xml due to issues with Java 8
compilation


pshipton:
As this file is not used by openj9 it should just be deleted.

tajila:
@acrowthe can you modify this PR to delete the file instead

acrowthe:
@pshipton @tajila File deleted

pshipton:
This was not a good idea as this file is still needed, reverting.

PR #3233, opened 2018-10-11
Update DDR method references to passed classes

acrowthe:
Remove the requirement for Java 8 from the ddr modularity commands


pshipton:
The Java 8 restriction has been removed.

PR #3235, opened 2018-10-11
Add caching for resolved methods

dmitry-ten:
- To reduce the number of remote messages for resolved methods, cache return values of `getResolved[Virtual|Static|Special|Interface]Method` in a per-compilation, per-resolved method cache.

- Fix formatting of `TR_RemoteROMStringKey` hash function, replace `typedef` of `TR_ResolvedJ9JITaaSServerMethodInfo` with `using`.

### DIFF
--- a/runtime/compiler/control/J9Options.cpp
+++ b/runtime/compiler/control/J9Options.cpp
@@ -45,6 +45,7 @@
 #include "control/CompilationRuntime.hpp"
 #include "control/CompilationThread.hpp"
 #include "runtime/IProfiler.hpp"
+#include "env/j9methodServer.hpp"
 
 #if defined(J9VM_OPT_SHARED_CLASSES)
 #include "j9jitnls.h"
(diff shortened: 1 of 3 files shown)


dmitry-ten:

dmitry-ten:
- Added underscores to members of new structs
- Moved `_useCaching` variable to CompilationInfo, so that it's set only once.

dmitry-ten:
Rebased to latest jitaas

runtime/compiler/env/j9methodServer.cpp
```diff
@@ -1186,6 +1209,37 @@ TR_ResolvedJ9JITaaSServerMethod::setAttributes(TR_OpaqueMethodBlock * aMethod, T
    setMandatoryRecognizedMethod(mandatoryRm);
    setRecognizedMethod(rm);
 
+   _useCaching = true; // TODO: figure out how to change this variable without retrieving env variable each time a resolved method is created (i.e. feGenEnv(TR_DisableResolvedMethodsCaching))
```
mpirvu: The solution is to define _useCaching as a static. If initializing this static does not work with 
`bool TR_ResolvedJ9JITaaSServerMethod::_useCaching = (feGetEnv("TR_ResolvedJ9JITaaSServerMethodCaching") != NULL);`
Then you could set it to true and overwrite it during options processing when we set options for JITaaS. In the end we may want an option rather than an env var.

mpirvu:
The code looks good overall. If you change _useCaching to a static the 
code should be ready to be merged.

dmitry-ten:
Updated

runtime/compiler/control/J9Options.cpp
```diff
@@ -2817,6 +2818,9 @@ J9::Options::unpackOptions(char *clientOptions, size_t clientOptionsSize, TR_Mem
    if (rtResolve)
       jitConfig->runtimeFlags |= J9JIT_RUNTIME_RESOLVE;
 
+   // disable caching of resolved methods if env variable TR_DisableResolvedMethodsCaching is set
+   TR_ResolvedJ9JITaaSServerMethod::_useCaching = !feGetEnv("TR_DisableResolvedMethodsCaching");
```
mpirvu: TODO: add an option flag in omr which will be set at client and propagated automatically at the server. Then we can check the option directly instead of having this `useCaching` field

PR #3238, opened 2018-10-11
Reduce number of messages between client and server

satbir-ca:
@mpirvu please review the changes

### DIFF
--- a/runtime/compiler/control/JITaaSCompilationThread.hpp
+++ b/runtime/compiler/control/JITaaSCompilationThread.hpp
@@ -58,6 +58,7 @@ class ClientSessionData
       void *classLoader;
       TR_OpaqueClassBlock * hostClass;
       TR_OpaqueClassBlock * componentClass; // caching the componentType of the J9ArrayClass
+      TR_OpaqueClassBlock * arrayClass;
       };
 
    struct J9MethodInfo
@@ -172,7 +173,7 @@ class JITaaSHelpers
    public:
    // NOTE: when adding new elements to this tuple, add them to the end,
    // to not mess with the established order.
-   using ClassInfoTuple = std::tuple<std::string, J9Method *, TR_OpaqueClassBlock *, int32_t, TR_OpaqueClassBlock *, std::vector<TR_OpaqueClassBlock *>, std::vector<std::tuple<bool, bool>>, bool, uintptrj_t , bool, uint32_t, TR_OpaqueClassBlock *, void *, TR_OpaqueClassBlock *, TR_OpaqueClassBlock *>;
+   using ClassInfoTuple = std::tuple<std::string, J9Method *, TR_OpaqueClassBlock *, int32_t, TR_OpaqueClassBlock *, std::vector<TR_OpaqueClassBlock *>, std::vector<std::tuple<bool, bool>>, bool, uintptrj_t , bool, uint32_t, TR_OpaqueClassBlock *, void *, TR_OpaqueClassBlock *, TR_OpaqueClassBlock *, TR_OpaqueClassBlock *>;
    static ClassInfoTuple packRemoteROMClassInfo(J9Class *clazz, TR_J9VM *fe, TR_Memory *trMemory);
    static void cacheRemoteROMClass(ClientSessionData *clientSessionData, J9Class *clazz, J9ROMClass *romClass, ClassInfoTuple *classInfoTuple);
    static J9ROMClass *getRemoteROMClassIfCached(ClientSessionData *clientSessionData, J9Class *clazz);
--- a/runtime/compiler/control/JITaaSCompilationThread.hpp
+++ b/runtime/compiler/control/JITaaSCompilationThread.hpp
@@ -57,6 +57,7 @@ class ClientSessionData
       TR_OpaqueClassBlock * leafComponentClass;
       void *classLoader;
       TR_OpaqueClassBlock * hostClass;
+      TR_OpaqueClassBlock * componentClass; // caching the componentType of the J9ArrayClass
       };
 
    struct J9MethodInfo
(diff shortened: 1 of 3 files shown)


mpirvu:
Could we please change the title of the PR to something more readable? Maybe: "Reduce number of messages between client and server"

runtime/compiler/control/JITaaSCompilationThread.hpp
```diff
@@ -57,6 +57,7 @@ class ClientSessionData
       TR_OpaqueClassBlock * leafComponentClass;
       void *classLoader;
       TR_OpaqueClassBlock * hostClass;
+      TR_OpaqueClassBlock * componentClass;
```
mpirvu: Because the name of this field is not the same as the name of the field in the j9class I would like a comment showing what exactly it is caching (J9ArrayClass::componentType) and the fact that is applicable only for array classes

runtime/compiler/env/VMJ9Server.cpp
```diff
@@ -639,9 +639,28 @@ TR_J9ServerVM::getNumInnerClasses(TR_OpaqueClassBlock * clazz)
...
+            {
+            return J9ROMCLASS_IS_PRIMITIVE_TYPE(it_comp->second.romClass) ? true : false;
+            }
+         }
+      }
+
    JITaaS::J9ServerStream *stream = _compInfoPT->getMethodBeingCompiled()->_stream;
    stream->write(JITaaS::J9ServerMessageType::VM_isPrimitiveArray, clazz);
    return std::get<0>(stream->read<bool>());
+
```
mpirvu: We could delete this extra line.

mpirvu:
Please let me know when testing is done.

satbir-ca:
@mpirvu added the three new commits along with the previous review comments update

satbir-ca:
sanity testing is done

runtime/compiler/env/VMJ9Server.cpp
```diff
@@ -1062,6 +1062,28 @@ TR_J9ServerVM::sameClassLoaders(TR_OpaqueClassBlock * class1, TR_OpaqueClassBloc
 bool
 TR_J9ServerVM::isUnloadAssumptionRequired(TR_OpaqueClassBlock *clazz, TR_ResolvedMethod *methodBeingCompiled)
    {
+   TR_OpaqueClassBlock * classOfMethod = static_cast<TR_ResolvedJ9JITaaSServerMethod *>(methodBeingCompiled)->classOfMethod();
+      {
```
mpirvu: Let's optimize this code: if clazz==classOfMethod then we can return false immediately; no need to search the hashtables. Next you can do a search for `clazz`. With it you can evaluate whether getClassLoader(clazz)==getSystemClassLoader as well as isAnonymousClass(clazz). If you cannot answer false then you must search for classOfMethod

runtime/compiler/env/VMJ9Server.cpp
```diff
@@ -1062,6 +1062,28 @@ TR_J9ServerVM::sameClassLoaders(TR_OpaqueClassBlock * class1, TR_OpaqueClassBloc
...
+      OMR::CriticalSection getRemoteROMClass(_compInfoPT->getClientData()->getROMMapMonitor());
+      auto it_class = _compInfoPT->getClientData()->getROMClassMap().find((J9Class*) clazz);
+      auto it_classOfMethod = _compInfoPT->getClientData()->getROMClassMap().find((J9Class*) classOfMethod);
+      if ((it_class != _compInfoPT->getClientData()->getROMClassMap().end()) &&
+          (it_classOfMethod != _compInfoPT->getClientData()->getROMClassMap().end()))
+         {
+         if (clazz == classOfMethod ||
+            (it_class->second.classLoader == getSystemClassLoader()) ||
+            ((it_class->second.classLoader == it_classOfMethod->second.classLoader) &&
+            !isAnonymousClass(clazz)))
```
mpirvu: Is `isAnonymousClass` working correctly? Have you tested this? I see it defined as 
`isAnonymousClass(TR_OpaqueClassBlock *j9clazz) { return (J9_ARE_ALL_BITS_SET(((J9Class*)j9clazz)->romClass->extraModifiers, J9AccClassAnonClass)); }`
but dereferencing the j9clazz at the server should cause problems in my opinion.
I know it's safe to use the other definition:
` isAnonymousClass(J9ROMClass *romClass) { return (J9_ARE_ALL_BITS_SET(romClass->extraModifiers, J9AccClassAnonClass)); }`
where you first obtain the romClass from the cache for clazz

runtime/compiler/env/VMJ9Server.cpp
```diff
@@ -1064,6 +1064,43 @@ TR_J9ServerVM::sameClassLoaders(TR_OpaqueClassBlock * class1, TR_OpaqueClassBloc
...
+      {
+      return false;
+      }
+
+      {
+      OMR::CriticalSection getRemoteROMClass(_compInfoPT->getClientData()->getROMMapMonitor());
+      auto it_class = _compInfoPT->getClientData()->getROMClassMap().find((J9Class*) clazz);
+      if (it_class != _compInfoPT->getClientData()->getROMClassMap().end())
+         {
+         if (it_class->second.classLoader == getSystemClassLoader())
```
mpirvu: The logic is not correct, because there is an `and` with `isAnonymous()`

runtime/compiler/env/VMJ9Server.cpp
```diff
@@ -1015,6 +1064,43 @@ TR_J9ServerVM::sameClassLoaders(TR_OpaqueClassBlock * class1, TR_OpaqueClassBloc
...
+            if (it_classOfMethod != _compInfoPT->getClientData()->getROMClassMap().end())
+               {
+               if (it_class->second.classLoader == it_classOfMethod->second.classLoader)
+                  {
+                  return false;
+                  }
+               else
+                  {
+                  return true;
+                  }
```
fjeremic: How about `return it_class->second.classLoader != it_classOfMethod->second.classLoader` to cut 8 lines into 1.

runtime/compiler/env/VMJ9Server.cpp
```diff
@@ -1015,6 +1064,28 @@ TR_J9ServerVM::sameClassLoaders(TR_OpaqueClassBlock * class1, TR_OpaqueClassBloc
...
+         if (!isAnonymousClass(it_class->second.romClass))
+            {
+            if (it_class->second.classLoader == getSystemClassLoader())
+               return false;
+            auto it_classOfMethod = _compInfoPT->getClientData()->getROMClassMap().find((J9Class*) classOfMethod);
+            if (it_classOfMethod != _compInfoPT->getClientData()->getROMClassMap().end())
+               return (it_class->second.classLoader != it_classOfMethod->second.classLoader);
+            }
+         else
+            return true;
```
mpirvu: Please use curly brackets around `return true`. The convention is that the `if` and the `else` blocks have to look the same: either both have curly brackets or none of them has.

PR #3237, opened 2018-10-11
Fix JIT on macOS crash in compressedrefs builds

nbhuiyan:
Fixed by: 

1. Using `%define` instead of `equ` to define constants in jilconsts.inc
2. Fixing the order of file inclusion in X86Unresolveds.pnasm

Closes: #3223 
Issue: #36 

Signed-off-by: Nazim Uddin Bhuiyan <nazim.uddin.bhuiyan@ibm.com>

PR #3243, opened 2018-10-11
Remove the use of the C Preprocessor on NASM assembly files

nbhuiyan:
NASM assembly files will no longer need the C Preprocessor. Going forward, there will be no intermediate source files to look at when debugging any of the JIT assembly code. Furthermore, OMR will not need to update its CMake files to handle the temporary `pnasm` extention (see eclipse/omr#3042)

Depends on: #3237 

~WIP because of dependency on the PR above which enables writing constants in `jilconsts.inc` with `%define` directives.~


0xdaryl:
`J9VM_OPT_TEMP_NEW_INTERFACE_INVOCATION` is not defined in any build spec (confirmed by @pshipton in the OpenJ9 Slack discussion [here](https://openj9.slack.com/archives/C8312LCV9/p1536771417000100).  Can you simply remove it and fold code assuming it's not defined?

Other changes look good.

nbhuiyan:
@0xdaryl addressed your concerns regarding the never-defined macro, rebased, and did some local testing. Ready for review.

0xdaryl:
Jenkins test sanity

0xdaryl:
macOS failure is infrastructure, and the Win32 tests all passed but failed notification (infra issue).  I don't see any problems related to this PR.  Merging...

runtime/compiler/x/runtime/X86PicBuilder.nasm
```diff
@@ -371,7 +363,7 @@ mergePopulateIPicClass:
       LoadClassPointerFromObjectHeader eax, edx, edx              ; receiver class
       mov         eax, dword  [edx+J9TR_J9Class_classLoader]      ; receivers class loader
 
-      mov         ebx, dword  [esi+eq_IPicData_interfaceClass]    ; ebx == IClass from IPic data area
+      mov         ebx, dword  [esi+eq_IPicData_interfaceClass]
```
0xdaryl: Why is that comment removed?  Seems still relevant to me.

0xdaryl: That commit in the .pasm file was the one that introduced the esi offset, and then deleted the comment.  For some reason your pnasm file still had the comment along with the new code with the offset (it was in some hybrid state).   That's OK--I think you can leave the code as-is.

PR #3246, opened 2018-10-11
Remove emma build tool from DDR compilation

acrowthe:
Emma was removed due to
- Causing issues with Java 8 builds
- Not being used, so can be removed without issues


pshipton:
Cancelled the unnecessary travis build.

PR #3250, opened 2018-10-12
Enable zLinux JDK8 compiles to use openssl 1.1.1

### LINKED ISSUE #3118: Enable and bundle an openssl library with OpenJ9 builds
The OpenJ9 0.11.0 release is coming in Oct, with final builds created for Oct 19.

One of the features [1] of the release is using the openssl 1.1.x library, which we'd like to bundle with the OpenJ9 Java 8 build for the time being.
[1] ibmruntimes/openj9-openjdk-jdk8#112

There is a choice to use a pre-built openssl library installed on the build machines (including header files), however this means the machines will need to be updated every time there is a security fix required. This affects all platforms: Windows 32-bit, 64-bit, Linux, AIX, macOS (later).

If you use pre-built OpenSSL binary, then

    use the following configure option, --with-openssl=path_to_openssl_library and --enable-openssl-bundling

If you are building openssl binary with openJDK8, then

    use --openssl-version=1.1.XX with ‘bash get_source.sh’ command. (NOTE : For Windows platform, building openssl from source along with OpenJDK8 had a known issue and has to use pre-built openssl binary); and
    use the following configure option, --with-openssl=fetched and --enable-openssl-bundling

jdekonin:
* [skip ci]
* fixes #3118

fyi @pshipton

### DIFF
--- a/buildenv/jenkins/variables/defaults.yml
+++ b/buildenv/jenkins/variables/defaults.yml
@@ -132,6 +132,8 @@ linux_390-64_cmprssptrs:
     vars:
       11: 'CC=gcc-7 CXX=g++-7'
       next: 'CC=gcc-7 CXX=g++-7'
+  extra_configure_options:
+    8: '--with-openssl=/home/jenkins/openssl-1.1.1 --enable-openssl-bundling'
 #========================================#
 # AIX PPC 64bits Compressed Pointers
 #========================================#


pshipton:
jenkins compile zlinux jdk8

pshipton:
Tested in https://github.com/ibmruntimes/openj9-openjdk-jdk8/pull/140, which was a prereq for this change.

PR #3251, opened 2018-10-12
Pass if JIT is loaded to Verbose GC

amicic:
Pass info to Initialized hook (effectively to Verbose GC that is the
only registered user) if JIT is loaded. The info (among some other) is
used to report if Concurrent Scavenger is H/W assisted or not.

### DIFF
--- a/runtime/gc_modron_startup/mminit.cpp
+++ b/runtime/gc_modron_startup/mminit.cpp
@@ -2970,9 +2970,9 @@ hookAcquireVMAccess(J9HookInterface** hook, UDATA eventNum, void* voidEventData,
 jint
 triggerGCInitialized(J9VMThread* vmThread)
 {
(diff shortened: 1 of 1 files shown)


fjeremic:
Assigning to myself given the OMR dependency. @dmitripivkine could you please review?

amicic:
There is a minor dependency on https://github.com/eclipse/omr/pull/3054. No compile dependacy, just runtime.
Without the OMR change, info that we pass (if JIT is loaded) will be incorrectly interpreted (as if CS is enabled), and associated Verbose GC line may be incorrect or confusing. But there will be no effect on program/JVM execution.

fjeremic:

fjeremic:

amicic:
Added some body in the commit and PR descriptons. (no code changes)

Issue #3252, opened 2018-10-12
Revert mac osx default to -Xrs when signal handler support fixed


DanHeidinga:
Temporarily default mac osx to use `-Xrs` while the signal handler issues are being investigated.  This allows most things to run but with reduced use of signals by the VM.

This item needs to stay open until this change has been reverted.

related to issue: #36

keithc-ca:
#3254 wasn't needed, so this won't be needed either.

PR #3254, opened 2018-10-12
Temporarily default to -Xrs on mac os x

DanHeidinga:
The signalhandlers are hanging under some conditions on mac and this is
a temporary workaround while we find the root case.

issue: #3252 has been opened to track reverting this change once
the root cause is fixed.

issue: #36


DanHeidinga:
FYI @keithc-ca I've put together a temporary work around for the signal handling issue you're investigating.  I'll hold off on this for a bit until we have a better idea of where we're at with a real fix

DanHeidinga:
Removing the WIP from the PR as we don't have a solution to the signal handler issue yet and we're out of runway for the 0.11.0 release

DanHeidinga:
Jenkins compile xlinux jdk8

DanHeidinga:
A single non-osx compile should be sufficient to show this doesn't break other platforms.

keithc-ca:
https://github.com/eclipse/omr/pull/3070 fixes the handling of the divide by zero exception discussed in #36: closing this.

PR #3260, opened 2018-10-13
Remove retry(5) around checkouts on master

AdamBrousseau:
- #1787 is resolved so this should be removed
- Leaving it in can hide future problems
- Killing a build in a retry will not actually
  stop the build, it will just loop into the next
  retry iteration which is problematic and annoying.


AdamBrousseau:
@pshipton for review

PR #3261, opened 2018-10-13
Remove stale function definition on X86

0dvictor:
reverseLoadEvaluator has only definition but no implementation,
removing it.


andrewcraik:
Jenkins test sanity xlinux,win jdk8.jdk11

andrewcraik:

PR #3262, opened 2018-10-13
Don't fingerprint artifacts

AdamBrousseau:
- Fingerprinting will store md5 checksums for artifacts in a DB
- This takes time to execute
- It is useful for tracking artifact dependencies around builds
- We don't use this functionality so we should remove it to save time
- More info: https://wiki.jenkins.io/display/JENKINS/Fingerprint

[skip ci]


AdamBrousseau:
@pshipton for review

PR #3267, opened 2018-10-14
Migrate from wrtbar snippet to OOL instructions on X86

0dvictor:
Use out-of-line instructions to call write barrier helpers,
and deprecate write barrier related snippets.


andrewcraik:

PR #3290, opened 2018-10-15
Proper set SupportsReferenceArrayCopy in CodeGen X86

0dvictor:
OMR incorrectly set SupportsReferenceArrayCopy and is going to
remove it; therefore, OpenJ9 should set the flag own its own.


andrewcraik:

PR #3291, opened 2018-10-15
Deprecate ShrinkWrapping - Part III

fjeremic:
Remove some last bits and pieces of Shrink Wrapping code in the stack
walker.

Issue: https://github.com/eclipse/omr/issues/2107

### DIFF
--- a/debugtools/DDR_VM/src/com/ibm/j9ddr/vm29/j9/stackwalker/MethodMetaData.java
+++ b/debugtools/DDR_VM/src/com/ibm/j9ddr/vm29/j9/stackwalker/MethodMetaData.java
@@ -27,7 +27,6 @@ import static com.ibm.j9ddr.vm29.pointer.generated.J9StackWalkFlags.J9SW_REGISTE
 import static com.ibm.j9ddr.vm29.structure.J9StackWalkConstants.J9SW_POTENTIAL_SAVED_REGISTERS;
 import static com.ibm.j9ddr.vm29.structure.J9StackWalkConstants.J9SW_REGISTER_MAP_MASK;
 import static com.ibm.j9ddr.vm29.structure.MethodMetaDataConstants.INTERNAL_PTR_REG_MASK;
-import static com.ibm.j9ddr.vm29.structure.MethodMetaDataConstants.J9TR_SHRINK_WRAP;
 import static com.ibm.j9ddr.vm29.structure.J9JITExceptionTable.*;
 
 import java.util.LinkedList;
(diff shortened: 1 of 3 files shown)


fjeremic:
Changes are rather simple but wouldn't hurt a review from either @gacholio or @keithc-ca given changes in their area of expertise. FYI this removes the last bit of code related to shrink wrapping in OpenJ9 (AFAIK).

gacholio:
Looks reasonable to me - I don't think you should edit the DDR artifacts. @keithc-ca ?

keithc-ca:
The changes to files in `DDR_Artifacts` have no effect on OpenJ9 builds, they're only relevant to the legacy tooling used in IBM builds and they're not needed (or wanted) there.

keithc-ca:
I don't know what 'shrink wrapping' means in this context, but if _any_ JVM based on OpenJ9 exists that might generate a core file to be examined, the support for interpreting such core files should remain.

fjeremic:
> The changes to files in `DDR_Artifacts` have no effect on OpenJ9 builds, they're only relevant to the legacy tooling used in IBM builds and they're not needed (or wanted) there.

Ok, I'll revert the spec changes then. On another note if these files have no effect on OpenJ9 builds can we migrate them to the IBM internal repo that cares about them? Why pollute the open source repo if the files are meaningless? Should we open up an issue for this? If you can specify which files can be migrated I will gladly do the work.

fjeremic:
Jenkins test sanity all JDK8

debugtools/DDR_VM/src/com/ibm/j9ddr/vm29/j9/stackwalker/MethodMetaData.java
```diff
@@ -451,43 +448,23 @@ public void jitAddSpilledRegisters(WalkState walkState, VoidPointer stackMap) th
...
-						savedGPRs = registerSaveDescription.bitAnd(new UDATA(31));
-						saveOffset = registerSaveDescription.rightShift(11);
-						lowestRegister = new UDATA(32).sub(savedGPRs);
-					}
+					/*
+						* see PPCLinkage for a description of the RSD 
+						* the save offset is located from bits 18-32 
+						* so first mask it off to get the bit vector 
+						* corresponding to the saved GPRS
+						*/
```
keithc-ca: Could you please fix the indentation of this comment so is consistent?

fjeremic: Fixed in 19913fe

0xdaryl:
Jenkins test sanity all JDK11

Launching again because the record of whether the previous PR build was successful was lost on the last push, and I did not see the results.

PR #3296, opened 2018-10-16
Do not set a call count for bytecodes tracked by IProfiler

mpirvu:
The optimizer may try to artificially create IProfiler entries for bytecodes
that it does not track. This should be the case for "invokeStatic" and
"invokeSpecial". However, we see this happening for "invokeVirtual" as well.
In this commit we will add a filter such that setCallCount() will do nothing
for calls that are already tracked by IProfiler.

### DIFF
--- a/runtime/compiler/runtime/JITaaSIProfiler.cpp
+++ b/runtime/compiler/runtime/JITaaSIProfiler.cpp
@@ -631,6 +631,13 @@ TR_JITaaSIProfiler::setCallCount(TR_OpaqueMethodBlock *method, int32_t bcIndex, 
    {
    uintptr_t methodStart = TR::Compiler->mtd.bytecodeStart(method);
    uint8_t bytecode = *((uint8_t*)(methodStart+bcIndex));
+   // The following bytecode types are tracked by IProfiler, so we
+   // don't need to artificially set a call count for them
+   if (bytecode == JBinvokevirtual ||
+       bytecode == JBinvokeinterface ||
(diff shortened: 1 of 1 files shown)


ymanton:
The change itself LGTM.

Is it possible to do the check in the optimizer instead if it calls `setCallCount` in only a few places? That way the change could be committed to master and we'd have one less patch to carry and it would also keep the behaviour consistent between local and remote compilations.

mpirvu:
The root problem is in the inliner which (according to a comment) is supposed to setCallCount only for direct calls.
```
                  // if call-graph profiling is enabled and the call is special or static (!indirect)
                  // then update the block frequency information because we don't profile predictable calls
                  if (!callSites[i]->isIndirectCall())
                     {
                     profileManager->updateCallGraphProfilingCount( currentBlock, calltarget->_calleeMethod->getPersistentIdentifier(), i, comp());
                     heuristicTrace(tracer(),"Depth %d: Updating Call Graph Profiling Count for calltarget %p count = %d",_recursionDepth, calltarget,profileManager->getCallGraphProfilingCount(calltarget->_calleeMethod->getPersistentIdentifier(), i, comp()));
                     }
```
Somehow the `isIndirectCall()` query responds false for virtualInvokes and I think this is a bug that needs to be fixed in the main branch. I will open an issue for it, but, since this bug does not result in a functional or performance problem I am guessing that it is not going to be addressed very soon. 
The solution in this PR ca be eliminated once the bug is fixed, but it is not harmful even if it stays on.

PR #3300, opened 2018-10-16
Update J9Package exportToAll data type build settings

acrowthe:
Remove entries for J9Package.exportToAll and
J9Package.exportToAllUnnamed from spec files
- Prevent generation as exportToAll_v1 and exportToAllUnnamed_v1
- Required to be able to consistantly utilize these variables

When the J9PackagePointer file was being generated, the contained
variables exportToAll and exportToAllUnnamed were sometimes
generated as exportToAll_v1 and exportToAllUnnamed_v1. This was
determined to be due to these variables having different types on
different platforms: UDATA on windows and BOOLEAN everywhere else.

Removing these mentions from spec files results in the correct variables
being generated, without _v1. This allows for consistent access to these
variables, without compile time variance.

Due to these two variables having potentially difering types, it is
important that any uses of them function on both UDATA and BOOLEAN data
types.


tajila:
@keithc-ca please take a look

acrowthe:
Aside from runtime/ddr/overrides-vm and runtime/ddr/vmddrstructs.properties, none of the changed files have copy write notices

keithc-ca:
Jenkins test sanity win,zlinux jdk8,jdk11

PR #3305, opened 2018-10-16
Update j9ddr build target directory

acrowthe:
- Build j9ddr.jar to java8_j9ddr.jar
- Fixes build issues related to old vm versions expecting j9ddr.jar to
have been compiled using older versions of java

### DIFF
--- a/debugtools/DDR_VM/generate.xml
+++ b/debugtools/DDR_VM/generate.xml
@@ -563,7 +563,7 @@
 		<if>
 		  	<equals arg1="${product}" arg2="java8" />
 			<then>
-				<copy file="${output.dir}/libs/j9ddr.jar" todir="/bluebird/tools/ddr/" />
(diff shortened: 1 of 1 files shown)


debugtools/DDR_VM/generate.xml
```diff
@@ -563,7 +563,7 @@
 		<if>
 		  	<equals arg1="${product}" arg2="java8" />
 			<then>
-				<copy file="${output.dir}/libs/j9ddr.jar" todir="/bluebird/tools/ddr/" />
+				<copy file="${output.dir}/libs/j9ddr.jar" tofile="/bluebird/tools/ddr/java8_j9ddr.jar" />
```
pshipton: I'd prefer to keep the prefix as j9ddr.
java8 is not unique as there is a jvm.28 version and a jvm.29 version, so how about using the name j9ddr_jvm29.jar

PR #3306, opened 2018-10-16
Software RangeCheck Read Barrier

amicic:
Renaming flag and command line option S/W Evacuate RB to S/W RangeCheck
RB. Evacuate is an action that is required by RB in Concurrent
Scavenger. It's always executed in software. What has alternative
implementation in S/W vs H/W is the trigger mechanism, which is the
range check against Evacuate area boundaries (or a set of areas).


dmitripivkine:
Waiting for promotion of https://github.com/eclipse/omr/pull/3066

amicic:

PR #3309, opened 2018-10-16
Simplify cleanWs commands

AdamBrousseau:
- CleanWs commands at the end of builds were inconsistent
- Adding notFailBuild: true so that the build will
  not fail if the cleanWs fails
- Adding disableDeferredWipeout: true to avoid
  asynchronous WS cleanups failing and getting left behind

[skip ci]
https://jenkins.io/doc/pipeline/steps/ws-cleanup/

### DIFF
--- a/buildenv/jenkins/common/build
+++ b/buildenv/jenkins/common/build
@@ -335,13 +335,7 @@ def build_all() {
                 archive()
             }
         } finally {
-            try {
-                cleanWs()
-            } catch(e) {
-                echo e.toString()
-                echo e.printStackTrace()
-                sh "ps -ef"
-            }
+            cleanWs notFailBuild: true, disableDeferredWipeout: true
         }
     }
 }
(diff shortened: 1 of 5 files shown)


AdamBrousseau:
Jenkins compile zlinux jdk8

AdamBrousseau:

AdamBrousseau:
PR builds will cover `build` and `test` files.

AdamBrousseau:

AdamBrousseau:
@vsebe and @pshipton for review

AdamBrousseau:
Not sure why the test job has 2 clean ws. Looking at previous builds it seems this behaviour isn't new. I will investigate and remove the duplicate.

AdamBrousseau:
There are 2 cleanWs in a PR test build because 1 is from the `test` file and the second is from the each PR file. I think this is fine as both will be removed with #2728.

AdamBrousseau:
Conflict resolved

PR #3310, opened 2018-10-16
Improve MethodHandles.lookup() field accesses

JasonFengJ9:
Improve `MethodHandles.lookup()` field accesses

Perform loader constraints check when accessing a field via a `FindGetter` or `FindSetter` `MethodHandle`. This applies to both `instance` and `static` fields.
Added tests to check `IllegalAccessException` with `LinkageError` cause thrown in following cases:
* `MH instance/static field getter/setter` for `reference` and `type` classes loaded by different class loaders
* `MH instance/static field getter/setter` for `reference/type` and `access` classes loaded by different class loaders

Reviewer: @DanHeidinga 


DanHeidinga:

Issue #3285, opened 2018-10-15
JITaaS: reducing the size of Persistent CH table updates


dmitry-ten:
Currently, CH table updates are the biggest messages, accounting for about a third of network traffic from client to the server. On average, updates are very big because for each dirty class, we serialize the class and pointers to all of its subclasses. But some classes have hundreds of subclasses, which results in a large message.
After looking at the code, I see that a list of subclasses is modified only in `TR_PersistentCHTable::classGotExtended`. This means that we don't need to send a list of subclasses to the server in other cases. 
I suggest that we keep track of when `classGotExtended` is called, and send subclasses only then.
This would also save some CPU, because computing the number of subclasses is expensive.

We could also keep track of classes that got added, and only send them, instead of all subclasses.

mpirvu:
Fixed by #3527

PR #3311, opened 2018-10-16
WIP: Optimize CH table serialization

dmitry-ten:
When serializing modifications to the persistent CH table, serialize subclasses only if a class got extended or it is the first CH table commit. Otherwise, we know that class hierarchy hasn't changed, so there is no point in sending subclasses. Also, only send subclasses that got
added.
This should improve CPU usage, because some classes can have hundreds
of subclasses, which makes CH table commits very large.

### DIFF
--- a/runtime/compiler/control/JITaaSCompilationThread.cpp
+++ b/runtime/compiler/control/JITaaSCompilationThread.cpp
@@ -1996,7 +1996,7 @@ bool handleServerMessage(JITaaS::J9ClientStream *client, TR_J9VM *fe)
          client->getRecvData<JITaaS::Void>();
          auto table = (TR_JITaaSClientPersistentCHTable*)comp->getPersistentInfo()->getPersistentCHTable();
          auto encoded = table->serializeUpdates();
-         client->write(encoded.first, encoded.second);
+         client->write(encoded.first, encoded.second, table->_commitNumber);
          }
          break;
       case J9ServerMessageType::CHTable_clearReservable:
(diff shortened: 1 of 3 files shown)


dmitry-ten:
This addresses #3285

dmitry-ten:
@mpirvu could you review?

mpirvu:
I haven't yet done a proper review, but can your code handle the case where a class is added and then removed and then an update is sent from the client to server?

mpirvu:
The large size of a CHTable update could also be due to the fact that `TR_JITaaSClientPersistentCHTable::findClassInfo(TR_OpaqueClassBlock * classId)` marks the class as dirty and therefore subject to being sent (together with all its subclasses).
It is a conservative solution because once we have a pointer to `TR_PersistentClassInfo` it's hard to determine how the caller will change this entry.

dmitry-ten:
- Yes, I agree that the current way to track added subclasses is pretty fragile but I did it this way to avoid any  additional copying. What we could do instead, is to put all added subclasses in a vector, and then send that vector as a part of `FlatPersistentClassInfo`.

 - We should be able to tell when a newly added subclass is deleted by creating a check for each superclass in `markSuperClassesAsDirty`.

mpirvu:
Is this PR obsolete now? If so let's close it.

dmitry-ten:
The idea of sending just the subclasses that got added could still be implemented,
but I don't think there is a point in trying at this moment, as CH table commits were made small enough.

PR #3317, opened 2018-10-17
Fix a compiler warning on X86

0dvictor:
guessClass has type uintptrj_t and hence 0 should be used instead of
NULL.


andrewcraik:

PR #3322, opened 2018-10-17
Resolve issue using "../java" when looking for a shared library doesn't work on zOS

groeges:
Use different code dependent on the OS we are building on, zOS vs all others.

Signed-off-by: Steve Groeger <GROEGES@uk.ibm.com>

### DIFF
--- a/runtime/jcl/common/system.c
+++ b/runtime/jcl/common/system.c
@@ -407,8 +407,12 @@ jstring getEncoding(JNIEnv *env, jint encodingType)
 			{
 				UDATA handle = 0;
 				PORT_ACCESS_FROM_ENV(env);
-				/* libjava.[so|dylib] is in the jdk/lib/ directory, one level up from the default/ & compressedrefs/ directories */
-				if (0 == j9sl_open_shared_library("../java", &handle, J9PORT_SLOPEN_DECORATE)) {
+#if defined(J9ZOS390)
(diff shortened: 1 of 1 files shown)


pshipton:
@DanHeidinga are you ok with this change, at least temporarily until another solution is found?

DanHeidinga:
I don't think this is the right approach and would prefer not to enable it in the code base as it makes it harder to reason about the code longer term.

zOS is not yet supported in OpenJ9 so these kinds of temporary patches can be kept in the local branches of those working on zOS.  

I would prefer to focus this investigation on the code doing the library lookup rather than adding this kind of temporary hack.

groeges:
Ok. I just wanted to get the zOS build working but am happy to cancel this PR. 

Will raise an issue to get this fixed in a proper way.  

Is this an OMR or OpenJ9 issue?

DanHeidinga:
Its worth raising the issue here as it relates to OpenJ9.  Its not clear yet where - openj9 vs omr -  the fix would be needed

groeges:
Closing as temporary fix is not sufficient. Issue raised to get this fixed properly: https://github.com/eclipse/openj9/issues/3392

groeges:
Closing as temporary fix is not sufficient.
Raised issue in order to get this fixed: https://github.com/eclipse/openj9/issues/3392

PR #3328, opened 2018-10-17
JDK8 OSX <arch> (=amd64) not defined

babsingh:
Don't remove \<arch\> (=amd64) from j9libBuffer. JDK8 on OSX doesn't have
amd64 directory. Removing \<arch\> invalidates the directory path stored
in j9libBuffer. This prevents java.home system property to be properly
defined. Incorrect java.home system property leads to JVM initialization
error. Paths to JAR files are incorrect; thus, the JAR files can't be
loaded.

I am guessing that (J9VM_JAVA9_BUILD < 150) won't have amd64 directory
on OSX since JDK8 doesn't have one. Also, (J9VM_JAVA9_BUILD < 150) is no
longer supported.
 
Issue: https://github.com/eclipse/openj9/issues/3327

### DIFF
--- a/runtime/j9vm/jvm.c
+++ b/runtime/j9vm/jvm.c
@@ -1096,10 +1096,15 @@ preloadLibraries(void)
 		jrebinBuffer = jvmBufferCat(NULL, jvmBufferData(j9binBuffer));
 	}
 	j9libBuffer = jvmBufferCat(NULL, jvmBufferData(jrebinBuffer));
-#if J9VM_JAVA9_BUILD < 150
+#if !defined(OSX)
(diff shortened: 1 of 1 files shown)


DanHeidinga:
Can you add the comment into the code that describes why this is needed?

It might be clearer to break it into two parts
```
#if !defined(OSX)
/* OSX requires the arch directory always */
#if (build blah < 150) 
/* Explain why removed for less than 150 */
#endif
#endif
```

babsingh:
related to #36.

DanHeidinga:

DanHeidinga:
We don't have osx jdk8 pr builds yet so running any jdk 8 & 11 build should be sufficient

DanHeidinga:

PR #3330, opened 2018-10-17
Fix ICF assert in pdcmpgt vector evaluator

fjeremic:
The vector registers in question are marked as part of the ICF which is
certainly not the case. RA will assert that we are trying to assign
registers within ICF. To alleviate this we add a label right after the
vector compare instruction and set ICF on the label.

### DIFF
--- a/runtime/compiler/z/codegen/J9BCDTreeEvaluator.cpp
+++ b/runtime/compiler/z/codegen/J9BCDTreeEvaluator.cpp
@@ -2701,8 +2701,8 @@ J9::Z::TreeEvaluator::pdcmpVectorEvaluatorHelper(TR::Node *node, TR::CodeGenerat
    generateRRInstruction(cg, TR::Compiler->target.is64Bit() ? TR::InstOpCode::XGR : TR::InstOpCode::XR, node, resultReg, resultReg);
    generateLoad32BitConstant(cg, node, 1, resultReg, true);
 
-   TR::RegisterDependencyConditions* deps = new(cg->trHeapMemory()) TR::RegisterDependencyConditions(0, 1, cg);
-   TR::LabelSymbol* doneCompareLabel = TR::LabelSymbol::create(cg->trHeapMemory(), cg);
+   TR::RegisterDependencyConditions* deps = new(cg->trHeapMemory()) TR::RegisterDependencyConditions(0, 2, cg);
+   deps->addPostConditionIfNotAlreadyInserted(resultReg, TR::RealRegister::AssignAny);
 
    TR::Node* pd1Node = node->getFirstChild();
    TR::Node* pd2Node = node->getSecondChild();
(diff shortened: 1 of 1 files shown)


fjeremic:
@NigelYiboYu could I please get a review?

runtime/compiler/z/codegen/J9BCDTreeEvaluator.cpp
```diff
@@ -2713,40 +2713,45 @@ J9::Z::TreeEvaluator::pdcmpVectorEvaluatorHelper(TR::Node *node, TR::CodeGenerat
...
-         generateS390BranchInstruction(cg, TR::InstOpCode::BRC, TR::InstOpCode::COND_CC0, node, doneCompareLabel);
-         generateS390BranchInstruction(cg, TR::InstOpCode::BRC, TR::InstOpCode::COND_CC2, node, doneCompareLabel);
+         cursor = generateS390BranchInstruction(cg, TR::InstOpCode::BRC, TR::InstOpCode::COND_CC0, node, icfEndLabel);
+         cursor = generateS390BranchInstruction(cg, TR::InstOpCode::BRC, TR::InstOpCode::COND_CC2, node, icfEndLabel);
          break;
       default:
          TR_ASSERT(0, "Unrecognized op code in pd cmp vector evaluator helper.");
       }
 
-   cursor = generateLoad32BitConstant(cg, node, 0, resultReg, true);
```
fjeremic: I actually think we should assert inside of these APIs that can take in dependencies for ICF. i.e. if you have ICF start already marked and you call this API and you don't pass in dependencies then assert because you're doing something wrong. This API can potentially allocate a register depending on the constant being loaded and that register has to be added to the dependencies. Food for thought.

fjeremic:

pshipton:
@fjeremic is this is must fix for the 0.11 release? Does it / will it affect zLinux?

runtime/compiler/z/codegen/J9BCDTreeEvaluator.cpp
```diff
@@ -2701,8 +2701,8 @@ J9::Z::TreeEvaluator::pdcmpVectorEvaluatorHelper(TR::Node *node, TR::CodeGenerat
    generateRRInstruction(cg, TR::Compiler->target.is64Bit() ? TR::InstOpCode::XGR : TR::InstOpCode::XR, node, resultReg, resultReg);
    generateLoad32BitConstant(cg, node, 1, resultReg, true);
 
-   TR::RegisterDependencyConditions* deps = new(cg->trHeapMemory()) TR::RegisterDependencyConditions(0, 1, cg);
-   TR::LabelSymbol* doneCompareLabel = TR::LabelSymbol::create(cg->trHeapMemory(), cg);
+   TR::RegisterDependencyConditions* deps = new(cg->trHeapMemory()) TR::RegisterDependencyConditions(0, 2, cg);
```
NigelYiboYu: Why 2 post dependencies?  I only see one call of `addPostCondition`.

NigelYiboYu: Is it because  `generateLoad32BitConstant` can potentially allocate another register?

NigelYiboYu: The implicit dependency changes in APIs like `generateLoad32BitConstant()` can be tricky.. How do we know the number of dependencies they are adding?

fjeremic: > Is it because `generateLoad32BitConstant` can potentially allocate another register?

Yes, this is precisely it.

fjeremic:
@pshipton do you want to be the committer for this one? I see we have not split the 0.11.0 branch so I assume I don't have to double deliver?

pshipton:
Correct, the branch has not yet occurred.

Issue #3326, opened 2018-10-17
OpenJ9 JDK8 OSX g++4.9 -Werror=header-guard unrecognized


babsingh:
When building OpenJ9 JDK8 on OSX, the following compiler option was unrecognized: `-Werror=header-guard`.

This compiler option is used by the JIT [runtime/compiler/build/toolcfg/gnu/common.mk](https://github.com/eclipse/openj9/blob/master/runtime/compiler/build/toolcfg/gnu/common.mk#L179):
```
ifeq ($(C_COMPILER),clang)
    CX_FLAGS+=-Wno-parentheses -Werror=header-guard
endif
```

When building JDK8 on OSX, `gcc-4.9` was used. The above `ifdefs` only add the compiler option if `C_COMPILER == clang`. Since `gcc-4.9` was used, `-Werror=header-guard` shouldn't be added. Perhaps, a setup issue.

After removing `-Werror=header-guard`, OpenJ9 JDK8 built without errors on OSX.

DanHeidinga:
@nbhuiyan mentioned he'll pick this up

PR #3215, opened 2018-10-10
Add OSX settings for PR builds

jdekonin:
* [skip ci]
* JDK8 and JDK11 only

Signed-off-by: Joe deKoning <joe_dekoning@ca.ibm.com>

#3124

PR #3331, opened 2018-10-17
Add amd64-osx-gcc.mk platform file for the JIT

nbhuiyan:
This is needed for building OpenJ9 with gcc when using older versions of
XCode.

amd64-osx-clang.mk is still going to be the default makefile used
unless "PLATFORM" environment variable is set.

Closes: #3326 

Note: You must set the PLATFORM variable as `amd64-osx-gcc` in order to make use of this makefile.

### DIFF
--- a/runtime/compiler/build/platform/host/amd64-osx-gcc.mk
+++ b/runtime/compiler/build/platform/host/amd64-osx-gcc.mk
@@ -1,0 +1,27 @@
+###############################################################################
+# Copyright (c) 2018, 2018 IBM Corp. and others
+#
+# This program and the accompanying materials are made available under
+# the terms of the Eclipse Public License 2.0 which accompanies this
+# distribution and is available at http://eclipse.org/legal/epl-2.0
+# or the Apache License, Version 2.0 which accompanies this distribution
+# and is available at https://www.apache.org/licenses/LICENSE-2.0.
+#
+# This Source Code may also be made available under the following Secondary
+# Licenses when the conditions for such availability set forth in the
+# Eclipse Public License, v. 2.0 are satisfied: GNU General Public License,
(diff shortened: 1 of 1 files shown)


nbhuiyan:
@babsingh agreed that he will set the PLATFORM variable in the JDK8 extensions repo

@0xdaryl @DanHeidinga fyi

DanHeidinga:
@nbhuiyan See also the change in this PR to ensure that PRODUCT is set - https://github.com/eclipse/openj9/pull/3215

Does that need to be changed to work with this?

nbhuiyan:
@DanHeidinga, changes to `runtime/compiler/makefile.ftl` proposed in #3215 would prevent us from making use of this file and would not solve the problem with building using gcc on older XCode, as this would override whatever we set `PLATFORM` to and prevent building OpenJ9 with JDK8. One (hacky) way to deal with this is to check for the `VERSION_MAJOR` variable in `makefile.ftl`, and set `PLATFORM` to `amd64-osx-gcc` for that. This would also remove the need to set it in the jdk extensions or manually when building with JDK8. How do you think this can be handled?

babsingh:
I incorporated `runtime/compiler/build/platform/host/amd64-osx-gcc.mk` in `openj9`. Set `export PLATFORM := amd64-osx-gcc`. Compiler option `-Werror=header-guard` wasn't used. JDK8 built properly on OSX.

DanHeidinga:
Change looks fine to me.  @nbhuiyan Can you remove the WIP?

DanHeidinga:
issue #36

DanHeidinga:
Jenkins compile xlinux jdk8

0xdaryl:
Posthumously looks fine to me.

Issue #3288, opened 2018-10-15
Reduce the number of `classOfStatic` messages


mpirvu:
Currently, one of the most frequently queries sent by the JITaaS server to the client is related to this method:
```
TR_OpaqueClassBlock *
TR_ResolvedJ9Method::classOfStatic(I_32 cpIndex, bool returnClassForAOT)
   {
   TR::VMAccessCriticalSection classOfStatic(fej9());
   TR_OpaqueClassBlock *result;
   result = _fe->convertClassPtrToClassOffset(cpIndex >= 0 ? jitGetClassOfFieldFromCP(_fe->vmThread(), cp(), cpIndex) : 0);
   return result;
   }
```
This method is very simple and it boils down to the implementation of ` jitGetClassOfFieldFromCP` which is
```
J9Class*
jitGetClassOfFieldFromCP(J9VMThread *vmStruct, J9ConstantPool *constantPool, UDATA fieldIndex) 
{
	J9RAMStaticFieldRef *ramRefWrapper;
	J9Class *result = NULL;

	/* romConstantPool is a J9ROMConstantPoolItem */
	ramRefWrapper = ((J9RAMStaticFieldRef*) constantPool) + fieldIndex;
	if (J9RAMSTATICFIELDREF_IS_RESOLVED(ramRefWrapper)) {
		J9Class *classWrapper = J9RAMSTATICFIELDREF_CLASS(ramRefWrapper);
		UDATA initStatus = classWrapper->initializeStatus;
		if ((J9ClassInitSucceeded == initStatus) || ((UDATA) vmStruct == initStatus)) {
			result = classWrapper;
		}
	}
	return result;
}
```
We could reduce the number of `classOfStatic` queries at the server by employing some caching mechanism: Add a hashtable with cpIndex-->classOfStatic mappings to ClassInfo struct. 
From the resolvedMethod we know the `_ramClass` this method belongs to. 
We search the hashtable of j9classes for my `_ramClass` and then search the hashtable of cpIndex-->classOfStatic mappings. This implies 2 hashtable searches, but the advantage is that when a class gets unloaded or modified, the hashtable of cpIndex-->classOfStatic mappings simply goes away.
If my desired mapping is not found, then we send a `classOfStatic` message at the client and if the client replies with a non-null answer we cache that answer.

PR #3336, opened 2018-10-17
Add caching for classes of statics

dmitry-ten:
Add a cache to ClassInfo that stores results of remote calls of `TR_ResolvedJ9JITaaSServerMethod::classOfStatic`.

Addresses:#3288

### DIFF
--- a/runtime/compiler/env/j9methodServer.cpp
+++ b/runtime/compiler/env/j9methodServer.cpp
@@ -166,8 +166,34 @@ TR_ResolvedJ9JITaaSServerMethod::classOfStatic(I_32 cpIndex, bool returnClassFor
    // if method is unresolved, return right away
    if (cpIndex < 0)
       return nullptr;
+
+   TR::CompilationInfoPerThread *compInfoPT = _fe->_compInfoPT;
+      {
+      OMR::CriticalSection getRemoteROMClass(compInfoPT->getClientData()->getROMMapMonitor()); 
+      auto &classOfStaticCache = getJ9ClassInfo(compInfoPT, _ramClass)._classOfStaticCache;
+      if (!classOfStaticCache)
+         {
+         // initialize cache, called once per ram class
+         classOfStaticCache = new (PERSISTENT_NEW) PersistentUnorderedMap<int32_t, TR_OpaqueClassBlock *>(PersistentUnorderedMap<int32_t, TR_OpaqueClassBlock *>::allocator_type(TR::Compiler->persistentAllocator()));
+         }
+      
+      auto it = classOfStaticCache->find(cpIndex);
+      if (it != classOfStaticCache->end())
+         return it->second;
+      }
+
    _stream->write(JITaaS::J9ServerMessageType::ResolvedMethod_classOfStatic, _remoteMirror, cpIndex, returnClassForAOT);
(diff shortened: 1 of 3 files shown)


dmitry-ten:
@mpirvu could you review this, when you have time?

runtime/compiler/env/j9methodServer.cpp
```diff
@@ -166,8 +166,22 @@ TR_ResolvedJ9JITaaSServerMethod::classOfStatic(I_32 cpIndex, bool returnClassFor
    // if method is unresolved, return right away
    if (cpIndex < 0)
       return nullptr;
+
+   auto &classOfStaticCache = getJ9ClassInfo(_fe->_compInfoPT, _ramClass)._classOfStaticCache;
```
mpirvu: You need a critical section here or two clients can allocate at the same time.

mpirvu: classOfStaticCache is a pointer. Do you need the `&` ?

runtime/compiler/env/j9methodServer.cpp
```diff
@@ -166,8 +166,22 @@ TR_ResolvedJ9JITaaSServerMethod::classOfStatic(I_32 cpIndex, bool returnClassFor
...
+      }
+   
+   auto it = classOfStaticCache->find(cpIndex);
+   if (it != classOfStaticCache->end())
+      return it->second;
+
    _stream->write(JITaaS::J9ServerMessageType::ResolvedMethod_classOfStatic, _remoteMirror, cpIndex, returnClassForAOT);
-   return std::get<0>(_stream->read<TR_OpaqueClassBlock *>());
+   TR_OpaqueClassBlock *classOfStatic = std::get<0>(_stream->read<TR_OpaqueClassBlock *>());
+   classOfStaticCache->insert({cpIndex, classOfStatic});
```
mpirvu: Caching should take place only if the client returns a valid class. It's possible that at this time the cp entry is not resolved or that the class is not yet fully initialized, but things may change in the future.

runtime/compiler/env/j9methodServer.cpp
dmitry-ten: Yes, because if classOfStaticCache is NULL, we need to initialize it, which means we need to change the value of the pointer in ClassInfo.

dmitry-ten:
- Surrounded ROM map access with critical sections
- Only caching non-NULL results now

mpirvu:
Please let me know when testing is done.

dmitry-ten:
Testing is done, another rebase complete

runtime/compiler/env/j9methodServer.cpp
```diff
@@ -856,6 +882,7 @@ TR_ResolvedJ9JITaaSServerMethod::fieldOrStaticName(I_32 cpIndex, int32_t & len,
    TR::CompilationInfoPerThread *threadCompInfo = _fe->_compInfoPT;
    {
       OMR::CriticalSection getRemoteROMClass(threadCompInfo->getClientData()->getROMMapMonitor()); 
+      // if class of static cache exists, free it
```
mpirvu: Is this a stray comment?

dmitry-ten: yes

PR #3345, opened 2018-10-18
Improve generated PTX code

keithc-ca:
The previous version would generate (see `compare()`) the sequences like:
```
testp.number.f64 p0,vl1;
setp.gtu.and.f64 p0,vl0,vl1,p0;
```
Using `p0` as both an input and an output (in the second instruction) seems not to work in all environments (failures have been seen with CUDA 5.5 on a K1100M and with CUDA 8.0 on a GTX960).

Using distinct predicate registers avoids the problem:
```
testp.number.f64 p1,vl1;
setp.gtu.and.f64 p0,vl0,vl1,p1;
```

Some instructions can be predicated to avoid unnecessary work.

This also fixes pattern for `-srand` test option: often the seed will be negative; such values should be accepted.


pshipton:
There aren't any tests for this, not sure its even worth a compile, but just in case...

pshipton:
jenkins compile xlinux

pshipton:
Adding to the 0.11.0 milestone in case anyone is building/running with gpu enabled.

PR #3024, opened 2018-09-25
Implement DDR findmodules command

acrowthe:
Implement DDR findmodules command

Added new command: !findmodules. Display all modules matching one of
the below options:
- all (default). Lists all modules loaded by the runtime
- name <module name>. Lists all modules with the same name as the parameter
- requires <module name>. Lists all modules that require the input module
- package <package name>. Lists the loaded module that owns the input package

Implements part of the new proposal within #1859

Created ModularityHelper.java
- Will assist in implementing other parts of the second proposal,
such as !dumppackage and !dumpmodule
- Used to traverse module and package lists
- Provides filter and output support

Example modules:
moduleA
```
module moduleA {
	exports moduleA.packageA;
	requires moduleB;
}
```
moduleB
```
module moduleB {
	exports moduleB.packageB to moduleA;
}
```
Example calls
!findmodules name moduleA
```
moduleA                        !j9module 0x00007FAC203C55B8
```
!findmodules requires moduleB
```
moduleA                        !j9module 0x00007FCA242F0960
```
!findmodules package moduleA.packageA
```
moduleA                        !j9module 0x00007FAC203C55B8
```
!findmodules
or
!findmodules all
```
jdk.naming.rmi                 !j9module 0x00007FAC20335E48
java.desktop                   !j9module 0x00007FAC203A1CB8
java.xml                       !j9module 0x00007FAC203B4D10
java.security.sasl             !j9module 0x00007FAC20335C98
jdk.management.jfr             !j9module 0x00007FAC203CD958
java.logging                   !j9module 0x00007FAC20335A58
java.datatransfer              !j9module 0x00007FAC20399068
java.naming                    !j9module 0x00007FAC20399410
jdk.jfr                        !j9module 0x00007FAC203A1490
java.rmi                       !j9module 0x00007FAC202EECC8
java.prefs                     !j9module 0x00007FAC20335788
openj9.sharedclasses           !j9module 0x00007FAC203358F0
java.management                !j9module 0x00007FAC20336118
java.management.rmi            !j9module 0x00007FAC203A1B08
jdk.management                 !j9module 0x00007FAC202EF148
java.base                      !j9module 0x00007FAC2008EAC8
java.compiler                  !j9module 0x00007FAC203998D8
java.security.jgss             !j9module 0x00007FAC203C5690
jdk.localedata                 !j9module 0x00007FAC202EF6E8
jdk.security.auth              !j9module 0x00007FAC202EEAD0
[...]

PR #3107, opened 2018-10-02
Add ddr dumppackage command

acrowthe:
Implement !dumppackage DDR modularity command

Usage: `!dumppackage [option] <packageAddress>`
Includes options to list the following details about a package:
- all: List `exportsTo`
- exportsTo: Lists all modules that the given module is exported to
- classes: Lists all loaded classes within the given package

Implements part of second proposal within #1859.
~Blocked by #3024 and #3072.~
No longer blocked

Example Commands:
`!dumppackage <package hex address>` or `!dumppackage all <package hex address>` or `!dumppackage exportsTo <package hex address>`
- When package is exported from a module (using example packageA):
```
> !dumppackage 0x00007FA2D43A9D88
Exported to:
Package is exported to all
> !dumppackage exportsTo 0x00007FA2D43A9D88
Package is exported to all
```
- When package is not exported from a module (using example packageNonExported):
```
> !dumppackage 0x00007FA2D43A9DD0
Exported to:
Package is not exported
> !dumppackage exportsTo 0x00007FA2D43A9DD0
Package is not exported
```
- When package is exported from a module to another module (using example packageB)
```
> !dumppackage 0x00007FA2D43360D8
Exported to:
moduleA                        !j9module 0x00007FA2D43A9D40
```
- When package is exported from a module to all unnamed modules (using the java `--add-exports <package>=ALL-UNNAMED` option)
```
> !dumppackage 0x00007F47043BB2F8
Exported to:
ALL-UNNAMED
```
`!dumppackage classes <package hex address>`
- Using packageA, with classA loaded and classNonLoaded not loaded:
```
> !dumppackage classes 0x00007FA93C2F1868
moduleA/packageA/classA        !j9class 0x000000000259CD00
Found 1 class
```

Modules for Examples:
```
moduleA
 | packageA
    | classA
    \ classNonLoaded
 | packageNonExported
    \ classNonExported
 \ module-info.java
moduleB
 | packageB
    \ classB
 \ module-info.java
```
moduleA module-info file:
```
module moduleA{
	exports moduleA.packageA;
	requires moduleB;
}
```
moduleB module-info file:
```
module moduleB{
	exports moduleB.packageB to moduleA;
}
```

Signed-off-by: Andrew Crowther acrowthe3388@gmail.com

PR #3072, opened 2018-09-28
Add ddr dumpmodules command

acrowthe:
Implement !dumpmodule DDR modularity command

Usage: !dumpmodule [option] <moduleAddress>
Includes options to list the following details about a module:
- all: List requires, exports and opens
- requires: Lists all modules required by given module
- exports: Lists all packages exported by given module
- classes: Lists all loaded classes within the given module
- packages: Lists all packages within the given module

Implements part of second proposal within #1859.

Example Commands:
`!dumpmodule <moduleA hex address>` or `!dumpmodule all <moduleA hex address>`
```
> !dumpmodule 0x00007FA2D43A9D40
Module:
moduleA                        !j9module 0x00007FA2D43A9D40
Requires:
java.base                      !j9module 0x00007FA2D408EAC8
moduleB                        !j9module 0x00007FA2D4336090
moduleC                        !j9module 0x00007FA2D43B2498
Exports:
Exports moduleA/packageA                              !j9package 0x00007FA2D43A9D88
```
`!dumpmodule exports <moduleA hex address>`
```
> !dumpmodule exports 0x00007FA2D43A9D40
Exports moduleA/packageA                              !j9package 0x00007FA2D43A9D88
Exports moduleA/packageExportUnnamed                  !j9package 0x00007F47043BB2F8
     to ALL-UNNAMED
```
`!dumpmodule exports <moduleB hex address>`
```
Exports moduleB/packageB                              !j9package 0x00007FA2D43360D8
     to moduleA                        !j9module 0x00007FA2D43A9D40
```
`!dumpmodule classes <moduleA hex address>`
- With both classes loaded:
```
> !dumpmodule classes 0x00007FA2D43A9D40
moduleA/packageA/classA        !j9class 0x000000000188CD00
moduleA/packageNonExported/classNonExported !j9class 0x000000000188D000
```
- Without classNonExported loaded:
```
> !dumpmodule classes 0x00007FA2D43A9D40
moduleA/packageA/classA        !j9class 0x000000000188CD00
```
!dumpmodule packages <moduleA hex address>
```
> !dumpmodule packages 0x00007FA2D43A9D40
moduleA/packageA                              !j9package 0x00007FA2D43A9D88
moduleA/packageNonExported                    !j9package 0x00007FA2D43A9DD0
```

Modules for Examples:
```
moduleA
 | packageA
    \ classA
 | packageNonExported
    \ classNonExported
 | packageExportUnnamed
    \ classExportUnnamed
 \ module-info.java
[...]

PR #3350, opened 2018-10-18
Add tests for ddr !findmodules command

acrowthe:
Add tests for the !findmodules ddr command implemented in #3024

### DIFF
--- a/test/functional/cmdLineTests/modularityddrtests/modularityddrtests.xml
+++ b/test/functional/cmdLineTests/modularityddrtests/modularityddrtests.xml
@@ -26,13 +26,13 @@
 
 <suite id="J9 modularity ddr command Tests" timeout="600">
 
- <variable name="PROGRAM" value="-version" />
- <variable name="DUMPFILE" value="j9core.dmp" />
- <variable name="DUMPDIR" value="dumpdir" />
- <variable name="XDUMP" value="-Xdump:system:file=$DUMPFILE$,events=vmstop" />
+	<variable name="PROGRAM" value="-version" />
+	<variable name="DUMPFILE" value="j9core.dmp" />
+	<variable name="DUMPDIR" value="dumpdir" />
+	<variable name="XDUMP" value="-Xdump:system:file=$DUMPFILE$,events=vmstop" />
 
- <!-- override the -Xdump command on z/OS -->
- <variable name="XDUMP" value="-Xdump:system:opts=IEATDUMP,dsn=%uid.J9CORE.DMP,events=vmstop,request=exclusive+compact" platforms="zos.*" />
+	<!-- override the -Xdump command on z/OS -->
+	<variable name="XDUMP" value="-Xdump:system:opts=IEATDUMP,dsn=%uid.J9CORE.DMP,events=vmstop,request=exclusive+compact" platforms="zos.*" />
  
  <test id="Create core file">
   <exec command="tso delete J9CORE.DMP.*" platforms="zos_390-64.*" />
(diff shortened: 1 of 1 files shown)


acrowthe:
Once #3072 and #3107 are merged, this will have to be rebased as it is presently blocked by them

Most of the changes in this PR are from these PR's.

Only changes made in the `test/functional/cmdLineTests/modularityddrtests` folder should be commented on here: All other changes should be commented on in their respective PR.

acrowthe:
> 
> 
> can you please split this into three PRs one for each command

Will do, once I get the tests fully functional.

acrowthe:
This pull request has been modified and now only implements the tests for the `!findmodules` command. As a result, it is no longer dependent on #3072 and #3107 and is ready for review

acrowthe:
@keithc-ca Could you please review this?

test/functional/cmdLineTests/modularityddrtests/modularityddrtests.xml
```diff
@@ -372,5 +372,184 @@
...
  </test>
- 
- </suite>
\ No newline at end of file
+
+
+
+ <!-- Test !findmodules and all options -->
+ <test id="Run !findmodules">
+  <exec command="sh" capture="LOGNAME" platforms="zos.*" >
```
keithc-ca: Please indent consistently: one tab (or a fixed number of spaces) per level. I'll let you decide whether you want to 'fix' the rest of the file.

Also, can you remove the extra blank lines above (one should be sufficient).

acrowthe: In the future the rest of the tests will be removed, leaving just new tests. As a result, I have only fixed indentation on new tests.

test/functional/cmdLineTests/modularityddrtests/modularityddrtests.xml
```diff
@@ -372,5 +372,181 @@
...
+		<exec command="cp //'$LOGNAME$.J9CORE.DMP.X001' $DUMPFILE$" platforms="zos_390-64.*" />
+		<exec command="cp //'$LOGNAME$.J9CORE.DMP' $DUMPFILE$" platforms="zos_390-31.*" />
+		<exec command="tso delete J9CORE.DMP.*" platforms="zos_390-64.*" />
+		<exec command="tso delete J9CORE.DMP" platforms="zos_390-31.*" />
+		<command command="$JDMPVIEW_EXE$">
+			<arg>-core $DUMPFILE$</arg>
+			<input>!findmodules name thisdoesnotexist</input>
+			<input>quit</input>
+		</command>
+		<output regex="no" type="success">> ></output>
```
keithc-ca: This not legal xml (the unescaped `>` characters). There are two more instances of this problem below.

test/functional/cmdLineTests/modularityddrtests/modularityddrtests.xml
```diff
@@ -372,5 +372,181 @@
...
+
+	<test id="Run !findmodules requires">
+		<exec command="sh" capture="LOGNAME" platforms="zos.*" >
+			<arg>-c</arg>
+			<arg>echo $$LOGNAME</arg>
+		</exec>
+		<exec command="cp //'$LOGNAME$.J9CORE.DMP.X001' $DUMPFILE$" platforms="zos_390-64.*" />
+		<exec command="cp //'$LOGNAME$.J9CORE.DMP' $DUMPFILE$" platforms="zos_390-31.*" />
+		<exec command="tso delete J9CORE.DMP.*" platforms="zos_390-64.*" />
+		<exec command="tso delete J9CORE.DMP" platforms="zos_390-31.*" />
```
keithc-ca: Won't this (on z/OS) remove the core file you're about to examine? Each of the tests below is the same.

acrowthe: The lines above the `tso delete` line appear to copy the dumpfile to `$DUMPFILE$`. In testing, `$DUMPFILE$` appears to be `j9core.dmp`, which would not be removed.

These lines seem to just be deleting the old copy of the dumpfile, after renaming it to what other platforms use.

acrowthe:
@keithc-ca These tests had been run on a linux machine without issue. I have confirmed that the new changes do run fine on a linux_x86-64_cmprssptrs machine as well.

Both concerns you raised have been addressed, either by fixing or explination.

test/functional/cmdLineTests/modularityddrtests/modularityddrtests.xml
acrowthe: While I have not tested this changes on z/OS, other tests within this file use the same logic and have run fine in past builds

test/functional/cmdLineTests/modularityddrtests/modularityddrtests.xml
acrowthe: `> >` has been replaced by `&gt &gt` in all 3 cases for compatibility purposes

test/functional/cmdLineTests/modularityddrtests/modularityddrtests.xml
```diff
@@ -372,5 +372,181 @@
...
+		<command command="$JDMPVIEW_EXE$">
+			<arg>-core $DUMPFILE$</arg>
+			<input>!findmodules name thisdoesnotexist</input>
+			<input>quit</input>
+		</command>
+		<output regex="no" type="success">&gt &gt</output>
```
keithc-ca: I think you mean `&gt;` instead of `&gt` (the semicolon is missing).

test/functional/cmdLineTests/modularityddrtests/modularityddrtests.xml
```diff
@@ -372,5 +372,181 @@
...
+		<exec command="cp //'$LOGNAME$.J9CORE.DMP.X001' $DUMPFILE$" platforms="zos_390-64.*" />
+		<exec command="cp //'$LOGNAME$.J9CORE.DMP' $DUMPFILE$" platforms="zos_390-31.*" />
+		<exec command="tso delete J9CORE.DMP.*" platforms="zos_390-64.*" />
+		<exec command="tso delete J9CORE.DMP" platforms="zos_390-31.*" />
+		<command command="$JDMPVIEW_EXE$">
+			<arg>-core $DUMPFILE$</arg>
+			<input>!findmodules requires thisdoesnotexist</input>
+			<input>quit</input>
+		</command>
+		<output regex="no" type="success">&gt &gt</output>
```
keithc-ca: Same as above: `&gt;` (and another below). Does the tool that parses this understand xml entities?

acrowthe: It definitely understands variables. Going to test entities, but will be a little while. This parser seems to be pretty lenient as both `>` and `&gt` were interpreted fine.
```
 <variable name="PROGRAM" value="-version" />
```

acrowthe: The tool does not seem to support entities.

keithc-ca: Then either the tool will need to be taught to process things like `&gt;` properly, or the test will need to specify a different success string.

acrowthe: To clarify: The testing tool does support default entity escapes like `&gt;` and `&lt;` (see line 107 for instances of those from before). I got confused and saw some xml extension that allowed custom entities, like `<!ENTITY example "example text"><output>&example</output>`, which are not supported.

Summary: `&gt;` is supported, I just got confused

acrowthe:
This has been updated to take into account #3534. @keithc-ca, your concerns with regards to no results tests should have been resolved by checking for `Found 0 modules` instead

test/functional/cmdLineTests/modularityddrtests/modularityddrtests.xml
```diff
@@ -372,5 +372,182 @@
...
+		<exec command="sh" capture="LOGNAME" platforms="zos.*" >
+			<arg>-c</arg>
+			<arg>echo $$LOGNAME</arg>
+		</exec>
+		<exec command="cp //'$LOGNAME$.J9CORE.DMP.X001' $DUMPFILE$" platforms="zos_390-64.*" />
+		<exec command="cp //'$LOGNAME$.J9CORE.DMP' $DUMPFILE$" platforms="zos_390-31.*" />
+		<exec command="tso delete J9CORE.DMP.*" platforms="zos_390-64.*" />
+		<exec command="tso delete J9CORE.DMP" platforms="zos_390-31.*" />
+		<command command="$JDMPVIEW_EXE$">
+			<arg>-core $DUMPFILE$</arg>
```
keithc-ca: Perhaps the current framework doesn't care, but `-core` and `$DUMPFILE$` should be in separate `<arg>` elements. (Here and below).

acrowthe:
Now reliant on #3586 due to invalid argument failure case message.

acrowthe:
@keithc-ca Changes from #3586 have been taken into account and your previous review has been resolved. This is ready for review again

test/functional/cmdLineTests/modularityddrtests/modularityddrtests.xml
```diff
@@ -372,5 +372,262 @@
...
+			<arg>echo $$LOGNAME</arg>
+		</exec>
+		<exec command="cp //'$LOGNAME$.J9CORE.DMP.X001' $DUMPFILE$" platforms="zos_390-64.*" />
+		<exec command="cp //'$LOGNAME$.J9CORE.DMP' $DUMPFILE$" platforms="zos_390-31.*" />
+		<exec command="tso delete J9CORE.DMP.*" platforms="zos_390-64.*" />
+		<exec command="tso delete J9CORE.DMP" platforms="zos_390-31.*" />
+		<command command="$JDMPVIEW_EXE$">
+			<arg>-core</arg>
+			<arg>$DUMPFILE$</arg>
+			<input>!findmodules package java/io</input>
```
keithc-ca: Shouldn't this be `java.io` (`.` instead of `/`)?

acrowthe: It looks like `/` is required instead of `.`. I just tested manually:
```
> !findmodules package java/io
java.base                      !j9module 0x00007F92E808FA68
Found 1 module
> !findmodules package java.io
Found 0 modules
```
it seems like internally it is stored with a `\`: the `findmodules` command is comparing directly to the internal packageName:
```
return J9UTF8Helper.stringValue(packagePtr.packageName()).equals(targetName);
```
If this is not the desired functionality, !`findmodules package` will need to be modified

acrowthe: This will be kept in mind for improving the user experience, but is not for now an issue.

keithc-ca: Yes, minimizing the translation required by humans would be good.

keithc-ca:

PR #3354, opened 2018-10-18
Fixing Proper ABI Linkage on PPC when calling CHelpers from JIT

AlenBadel:
PPC ABI is not being followed correctly on Linux & AIX on PPC, when calling CHelpers from JIT compiled code.

In the case of PPCLE64, the ABI behavior requires that the caller of the C/C++ compiled method pass the CHelper's address in r12 before calling.
Otherwise for PPCBE64, and AIX the caller of the C/C++ compiled method is required to pass the TOC, or jitTOC in this case within r15(PPCBE64, AIX64) or r13(AIX32).

This commit implements PPC ABI compatability, making sure that instructions are generated to reflect loading the proper content in the respecive registers prior to calling the CHelpers.

Signed-off-by: Alen Badel Alen.Badel@ibm.com.

### DIFF
--- a/runtime/compiler/p/codegen/PPCPrivateLinkage.cpp
+++ b/runtime/compiler/p/codegen/PPCPrivateLinkage.cpp
@@ -1913,6 +1913,42 @@ int32_t TR::PPCPrivateLinkage::buildPrivateLinkageArgs(TR::Node                 
 	    addDependency(dependencies, NULL, (TR::RealRegister::RegNum)i, TR_VSX_SCALAR, cg());
       }
 
+
+
+   /*
+   PPC ABI --
+   AIX 32-bit BE & AIX 64-bit BE & Linux 64-bit BE: 
+   These follow aix-style linkage. 
+   There is a TOC-base register (gr2) that needs to be loaded from function descriptor of that function that is called. 
+   If the callee function is in libj9jitxx.so module, (as in the case of CHelper functions are) then the module's TOC-base can be found in the J9VMThread jitTOC field.
+   The TOC can be loaded from the jitTOC field of J9VMThread (pointed to by metaDataRegister).
+  
+   Linux 64-bit LE: There exists a TOC (gr2), but there is no such function descriptor. 
+   The TOC-base set-up is defined by ABI to go through a function's global-entry (relocation). 
+   The global-entry address will be set up in gr12.
+
+   Registers r12, and r2 are always added to the dependency in the situations that they're used. 
+   R12 is added at all times, and is an unreserved register across all platforms. 
+   R2 is added to the dependency only if we're not dealing with a linux 32 bit platform, where it's system reserved. 
+   Ultimately, we don't need an assert or NULL check when searching for these registers within the dependency. 
+   */
+   if(linkage == TR_CHelper)
+   {
+      if(TR::Compiler->target.isLinux() && TR::Compiler->target.is64Bit() && TR::Compiler->target.cpu.isLittleEndian())
+      {
+         int32_t helperOffset = (callNode->getSymbolReference()->getReferenceNumber() - 1)*sizeof(intptrj_t);
+         generateTrg1MemInstruction(cg(), TR::InstOpCode::Op_load, callNode, dependencies->searchPreConditionRegister(TR::RealRegister::gr12), 
+            new(trHeapMemory()) TR::MemoryReference(cg()->getTOCBaseRegister(), helperOffset, TR::Compiler->om.sizeofReferenceAddress(), cg()));
+
+      }
+      else if (TR::Compiler->target.isAIX() || (TR::Compiler->target.isLinux() && TR::Compiler->target.is64Bit()))
+      {
+         generateTrg1MemInstruction(cg(), TR::InstOpCode::Op_load, callNode, dependencies->searchPreConditionRegister(TR::RealRegister::gr2), 
+            new(trHeapMemory()) TR::MemoryReference(cg()->getMethodMetaDataRegister(), offsetof(J9VMThread, jitTOC), TR::Compiler->om.sizeofReferenceAddress(), cg()));
+      }
+   }
+
+
    // This is a performance hack. CCRs are rarely live across calls,
    // particularly helper calls, but because we create scratch virtual
    // regs and re-use them during evaluation they appear live.


AlenBadel:
See https://github.com/eclipse/openj9/issues/2994

AlenBadel:
Unfortunately, both techniques to generate load on aix are failing. 

I'm taking a look at both getPPCRealRegister, and searchPreConditionRegister. 

getPPCRealRegister - Fails on AIX32
searchPreConditionRegister - Works on AIX64

AlenBadel:
Only non-reserved or special registers can be added to dependencies. Therefore we must use getPPCRealRegister to refer to the vmthread 13/15 on x32/x64

I've made the respective change. 

@gita-omr I would appreciate if you could look at this and forward me any feedback.

runtime/compiler/p/codegen/PPCPrivateLinkage.cpp
```diff
@@ -1913,6 +1913,27 @@ int32_t TR::PPCPrivateLinkage::buildPrivateLinkageArgs(TR::Node
...
+      {
+         uintptrj_t targetAddress = (uintptrj_t)callSymbol->getMethodAddress();
+         loadConstant(cg(), callNode, (int64_t)targetAddress,dependencies->searchPreConditionRegister(TR::RealRegister::gr12));
+      }
+      else if ((TR::Compiler->target.isAIX() || TR::Compiler->target.isLinux()) && TR::Compiler->target.is64Bit())
+      {
+         generateTrg1MemInstruction(cg(),TR::InstOpCode::Op_load, callNode, dependencies->searchPreConditionRegister(TR::RealRegister::gr2), 
+            new(trHeapMemory()) TR::MemoryReference(cg()->machine()->getPPCRealRegister(TR::RealRegister::gr15), offsetof(J9VMThread, jitTOC), TR::Compiler->om.sizeofReferenceAddress(), cg()));
+      }
+      else if (TR::Compiler->target.isAIX() && TR::Compiler->target.is32Bit())
```
gita-omr: We can also have Linux BE 32-bit (just double checked with @zl-wang). It also has J9VMThread in gpr13.  Please put this summary into the design and change the code accordingly:

**AIX 32-bit BE**: set gpr2 with TOC from  VMthread in gpr13
**AIX 64-bit BE**: set gpr2 with TOC from  VMThread in gpr15
**Linux 32-bit BE**: set gpr2 with TOC from  VMThread in gpr13
**Linux 64-bit LE**: set  gpr12 with target address
**Linux 64-bit BE**: set gpr2 with TOC from  VMThread in gpr15

I hope the table is correct, but @zl-wang please double check.

gita-omr: Sorry, I only added a confusion. Linux 32-bit BE does not have TOC, so we should not set gpr2. @zl-wang. I think having a comment in the code and a summary in the description would've helped.

zl-wang: The J9VMThread register should be retrieved by cg->getMethodMetaDataRegister().   It is gr13 on 32bit JVM and gr15 on 64bit JVM respectively.  It is a locked down register (dedicated register). It can be in dependency condition, just not depending on any other real register.

There are many examples of using method metadata register.

zl-wang: AIX 32-bit BE:
AIX 64-bit BE:
Linux 64-bit BE:
       These are aix-style linkage.  There is a TOC-base register (gr2) from function descriptor if that function is called.  You can load it from jitTOC field of J9VMThread (pointed to by metaDataRegister).

Linux 32bit BE:   no TOC.  Don't touch gr2.  gr2 is reserved.

Linux 64-bit LE:   there is TOC (gr2), but there is no function descriptor.  TOC-base set-up is defined by ABI to go through a function's global-entry (relocation).  The global-entry address should be set up in gr12.

runtime/compiler/p/codegen/PPCPrivateLinkage.cpp
```diff
@@ -1913,6 +1913,27 @@ int32_t TR::PPCPrivateLinkage::buildPrivateLinkageArgs(TR::Node
...
+      {
+         uintptrj_t targetAddress = (uintptrj_t)callSymbol->getMethodAddress();
+         loadConstant(cg(), callNode, (int64_t)targetAddress,dependencies->searchPreConditionRegister(TR::RealRegister::gr12));
+      }
+      else if ((TR::Compiler->target.isAIX() || TR::Compiler->target.isLinux()) && TR::Compiler->target.is64Bit())
+      {
+         generateTrg1MemInstruction(cg(),TR::InstOpCode::Op_load, callNode, dependencies->searchPreConditionRegister(TR::RealRegister::gr2), 
+            new(trHeapMemory()) TR::MemoryReference(cg()->getMethodMetaDataRegister(), offsetof(J9VMThread, jitTOC), TR::Compiler->om.sizeofReferenceAddress(), cg()));
+      }
+      else if (TR::Compiler->target.isAIX())
```
gita-omr: We can have just two cases, as @zl-wang summarized. Please include his summary as a comment.

runtime/compiler/p/codegen/PPCPrivateLinkage.cpp
```diff
@@ -1913,6 +1913,42 @@ int32_t TR::PPCPrivateLinkage::buildPrivateLinkageArgs(TR::Node
...
+   The global-entry address should be set up in gr12. (although TOC-base is recorded in jitTOC field of J9VMThread as well ... none-C helpers are using it).
+
+   */
+   
+   if(linkage == TR_CHelper)
+   {
+      if(TR::Compiler->target.isLinux() && TR::Compiler->target.is64Bit() && TR::Compiler->target.cpu.isLittleEndian())
+      {
+         uintptrj_t targetAddress = (uintptrj_t)callSymbol->getMethodAddress();
+         loadConstant(cg(), callNode, (int64_t)targetAddress,dependencies->searchPreConditionRegister(TR::RealRegister::gr12));
```
zl-wang: This needs a few instructions to materialize the address constant into gr12.
There is a better way:  the helper global-entry is in pseudo-TOC.  pseudo-TOC is based off gr16 (cg->getTOCBaseRegister()).  the index (offset) is 8 * helperSymbolIndex (I think).

AlenBadel: loadConstant seems to just generate one load instruction directly from gr16 into gr12.

zl-wang: It might be, but that consumes pseudo-TOC space unnecessarily (i.e.  loadConstant doesn't know targetAddress being a helper address, treats it as a new constant, and allocates a new pseudo-TOC entry for it). On the other hand, targetAddress is already in pseudo-TOC for creating helper trampoline purpose.

Read code in runtime/compiler/runtime/Trampoline.cpp (createHelperTrampolines) and runtime/compiler/p/codegen/J9PPCInstruction.cpp (TR::PPCDepImmSymInstruction::generateBinaryEncoding()).

zl-wang: you can verify the theory above by looking at the offset for the load you mentioned.  the offset is likely fairly big (more than 1200 for example).  the existing helper offset is likely less than 1200 (e.g.  helper index is less than 150).

runtime/compiler/p/codegen/PPCPrivateLinkage.cpp
```diff
@@ -1913,6 +1913,42 @@ int32_t TR::PPCPrivateLinkage::buildPrivateLinkageArgs(TR::Node
...
+      }
+      else if (TR::Compiler->target.isAIX() || (TR::Compiler->target.isLinux() && TR::Compiler->target.is64Bit()))
+      {
+         generateTrg1MemInstruction(cg(),TR::InstOpCode::Op_load, callNode, dependencies->searchPreConditionRegister(TR::RealRegister::gr2),
```
gita-omr: Why are we sure that searchPreConditionRegister(TR::RealRegister::gr2) will not return NULL? Apparently it was tested and works somehow, but I am not sure how.

gita-omr:
Yeah, that’s true, assuming volatiles have been added already. Might be good to put a comment or even an assert.

runtime/compiler/p/codegen/PPCPrivateLinkage.cpp
AlenBadel: The method searchPreConditionRegister essentially just checks if the register was added as a dependency. It returns the register if it was added, and null if it was never added as a dependency. 

We can see in buildPrivateLinkageArgs that in the case of CHelpers it always is added to the dependencies. In the case of Linux x32, it isn't since getPreserved will return True. 

```
   if (!properties.getPreserved(TR::RealRegister::gr2))
      {
      // Helper linkage preserves all registers that are not argument registers, so we don't need to spill them.
      if (linkage != TR_Helper)
         addDependency(dependencies, NULL, TR::RealRegister::gr2, TR_GPR, cg());
      }
```

gita-omr: Thanks. Adding a brief comment or even an ASSERT could be helpful.

runtime/compiler/p/codegen/PPCPrivateLinkage.cpp
AlenBadel: @zl-wang That's right. It did in fact create a new entry. I've made the respective changes, and tested them.

runtime/compiler/p/codegen/PPCPrivateLinkage.cpp
```diff
@@ -1913,6 +1913,47 @@ int32_t TR::PPCPrivateLinkage::buildPrivateLinkageArgs(TR::Node
...
+   R12 is added at all times, and is an unreserved register across all platforms. 
+   R2 is added to the dependency only if we're not dealing with a linux 32 bit platform. 
+   Ultimately, we don't need an assert or NULL check when searching for these registers within the dependency. 
+   */
+   if(linkage == TR_CHelper)
+   {
+      if(TR::Compiler->target.isLinux() && TR::Compiler->target.is64Bit() && TR::Compiler->target.cpu.isLittleEndian())
+      {
+         int32_t helperOffset = (callNode->getSymbolReference()->getReferenceNumber() - 1)*sizeof(intptrj_t);
+         generateTrg1MemInstruction(cg(), TR::InstOpCode::Op_load, callNode, dependencies->searchPreConditionRegister(TR::RealRegister::gr12),
```
gita-omr: Just curious: here we just set r12. There must be a code somewhere that does exactly the  same (loads helper address from the pseudo-TOC) and jumps to the helper. Which register is used there to contain the address? Should we just use r12 when we are jumping to the helper?

AlenBadel: That's done in the Trampoline. 

 openj9/runtime/compiler/runtime/Trampoline.cpp

#if defined(TR_TARGET_64BIT)
         // ld gr11, [grPTOC, 8*(cookie-1)]
         *(int32_t *)buffer = 0xe9700000 | (((cookie-1)*sizeof(intptrj_t)) & 0x0000ffff);
buffer += 4;

On Java JIT PPCLE64 the same instruction is done to load the target address into r11, in which we use to branch directly to the target. 

There is a version in OMR also that uses r12, which follows the ABI perfectly. 

I was told by @zl-wang that r11 is used for other reasons as well here.

gita-omr:
Right, thanks for reminding.

gita-omr:

AlenBadel:
I'm not entirely sure why the cmdLineTester_jep178_staticLinking_SE80_0 test failed. I've ran it over a hundred times with both a build not containing my changes and a build with my changes without any issues.

AlenBadel:

pdbain-ibm:
Likely known issue https://github.com/eclipse/openj9/issues/3672.

gita-omr:

PR #3356, opened 2018-10-18
Add tests for the ddr !dumpmodules command

acrowthe:
Implement tests for the DDR command that is implemented in #3072

Requires #3350 due to variable usage

### DIFF
--- a/test/functional/cmdLineTests/modularityddrtests/modularityddrtests.xml
+++ b/test/functional/cmdLineTests/modularityddrtests/modularityddrtests.xml
@@ -580,10 +580,11 @@
 		<output regex="no" type="required">!findmodules name</output>
 		<output regex="no" type="required">!findmodules requires</output>
 		<output regex="no" type="required">!findmodules package</output>
+		<output regex="no" type="failure">Argument failed to parse or was parsed to an unhandled subcommand.</output>
 		<output regex="no" type="failure">DDRInteractiveCommandException</output>
 	</test>
 
-	<test id="Run !findmodules with an invalid option">
+	<test id="Run !findmodules with an invalid subcommand">
 		<exec command="sh" capture="LOGNAME" platforms="zos.*" >
 			<arg>-c</arg>
 			<arg>echo $$LOGNAME</arg>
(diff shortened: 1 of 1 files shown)


acrowthe:
Blocked by #3350 due to requiring it's changes to be functional. Once #3350 is merged, this branch will have to be rebased and squashed to avoid conflicts.

acrowthe:
As all required PR's have been merged, this has been squashed and is ready to be merged.

@keithc-ca, when you have time, could you please review this?

test/functional/cmdLineTests/modularityddrtests/modularityddrtests.xml
```diff
@@ -630,4 +631,304 @@
...
+		<exec command="tso delete J9CORE.DMP.*" platforms="zos_390-64.*" />
+		<exec command="tso delete J9CORE.DMP" platforms="zos_390-31.*" />
+		<command command="$JDMPVIEW_EXE$">
+			<arg>-core</arg>
+			<arg>$DUMPFILE$</arg>
+			<input>!dumpmodule packages $javaBaseModuleAddr$</input>
+			<input>quit</input>
+		</command>
+		<saveoutput regex="no" type="success" saveName="nonExportedPackageAddr" splitIndex="1" splitBy="!j9package ">jdk/internal/platform </saveoutput>
+		<saveoutput regex="no" type="required" saveName="javaUtilPackageAddr" splitIndex="1" splitBy="!j9package ">java/util </saveoutput>
```
keithc-ca: Are the trailing spaces (e.g. `>java/util `) actually present? I would hope they're not, in which case this will fail to find the package. We also need to make sure it won't find a subpackage like `java/util/concurrent`.

test/functional/cmdLineTests/modularityddrtests/modularityddrtests.xml
```diff
@@ -630,4 +631,304 @@
...
+			<input>!dumpmodule exports $javaBaseModuleAddr$</input>
+			<input>quit</input>
+		</command>
+		<output regex="no" type="success">Exports java/nio/file/spi</output>
+		<output regex="no" type="required">!j9package 0x</output>
+		<output regex="no" type="required">Exports sun/util/logging</output>
+		<output regex="no" type="required">to java.logging</output>
+		<output regex="no" type="required">!j9module 0x</output>
+		<output regex="no" type="required">to ALL-UNNAMED</output>
+		<saveoutput regex="no" type="required" saveName="sunAnnotationPackageAddr" splitIndex="1" splitBy="!j9package ">sun/reflect/annotation </saveoutput>
```
keithc-ca: References to internal packages like `sun/util/logging` and `sun/reflect/annotation` will make this test fragile: it will break if/when they are renamed or refactored. Also, it appears `sunAnnotationPackageAddr` is unused.

test/functional/cmdLineTests/modularityddrtests/modularityddrtests.xml
```diff
@@ -630,4 +631,304 @@
...
+		<exec command="cp //'$LOGNAME$.J9CORE.DMP' $DUMPFILE$" platforms="zos_390-31.*" />
+		<exec command="tso delete J9CORE.DMP.*" platforms="zos_390-64.*" />
+		<exec command="tso delete J9CORE.DMP" platforms="zos_390-31.*" />
+		<command command="$JDMPVIEW_EXE$">
+			<arg>-core</arg>
+			<arg>$DUMPFILE$</arg>
+			<input>!dumpmodule 0x0</input>
+			<input>quit</input>
+		</command>
+		<output regex="no" type="success">Memory Fault reading 0x00000000 :</output>
```
keithc-ca: Addresses may display differently on 32-bit and 64-bit systems: perhaps this should use `regex="yes"`.

test/functional/cmdLineTests/modularityddrtests/modularityddrtests.xml
acrowthe: The output of `!dumpmodule package` outputs package name, then a series of spaces for alignment, then the package address, as shown below. Thus, the space catches the end of the package name, preventing sub packages from being caught.
```
//              v Last matched character, a space. This line would match "`moduleA/packageA `"
moduleA/packageA                              !j9package 0x00007FA2D43A9D88
//              v Last matched character, an s. This line would fail (correctly)
moduleA/packageA/subPackage                   !j9package 0x00007FA2D43A9D88
```
Is there something you would like me to change? Presently, it is working like intended.

acrowthe: This doesn't directly match a package name: instead it matches the format that DDR outputs the package name in, which is always space padded.

Thus, the space acts as matching the end of the package, rather than a character within the package name.

keithc-ca: I thought the package name was at the end of a line (which shouldn't have any trailing whitespace). Because it's at the beginning (followed by at least one space), we're good.

test/functional/cmdLineTests/modularityddrtests/modularityddrtests.xml
acrowthe: Switched to a regex searching for `0x0+ :` instead of `0x00000000 :`

test/functional/cmdLineTests/modularityddrtests/modularityddrtests.xml
acrowthe: References to sun packages have been replaced with `com/ibm/gpu/spi`. `sunAnnotationPackageAddr` has been removed

keithc-ca:
Jenkins test extended win,zlinux jdk11

PR #3358, opened 2018-10-18
Caching the instance size for the class

satbir-ca:
Caching the total instance size for the class and reducing the messages for classInstanceSize and
getAllocationSize.


satbir-ca:
@mpirvu can you please review the code

Issue #3338, opened 2018-10-18
Concurrent Scavenge support on X86


0dvictor:
#3054 has built a framework for X86 to support Concurrent Scavenge, and brought up the support on linux_x86-64_cmprssptrs.spec. This Issue is a follow-up to bring Concurrent Scavenge to all X86-64 specs.

List of works:
- [x] Optimizer to insert read barrier where necessary (for non-compressedref); #3362
- [x] CodeGen to enhance its read barrier related evaluators to support non-compressedref; #3363
- [x] Turn on Concurrent Scavenge in the Linux X86-64 build spec; #3365

0dvictor:
All pull requests listed in this Issue have been merged, closing.

SueChaplain:
@pshipton This is referenced in the 0.12.0 Eclipse plan for XL heap. Is there another issue for that piece that is still open please?

pshipton:
@0dvictor or @andrewcraik please open a issue at https://github.com/eclipse/openj9-docs to cover the documentation changes for this.

Is there a plan to enable Windows?

andrewcraik:
I believe the windows specific parts of the change are likely to be related to the GC so we need to be guided by @amicic on when they plan to do that work. @0dvictor will provide any necessary support once we know what is needed beyond that already implemented.

amicic:
I'm not aware of any changes on GC side. Once JIT side is ready, we just need to turn on the build flag. If testing shows problem both sides will investigate as appropriate.

BTW, this is how forcing s/w barrier on startup looks like (nothing Linux or Windows specific)

#if defined(J9VM_ARCH_X86)
			extensions->softwareRangeCheckReadBarrier = true;

PR #3362, opened 2018-10-18
Concurrent Scavenger Read Barrier evaluator on X86-64

0dvictor:
X86-64 default now supports read barriers for Concurrent Scavenger.

Part of Issue #3338


andrewcraik:

PR #3363, opened 2018-10-18
Generate read barrier for Concurrent Scavenger

0dvictor:
Under Concurrent Scavenger, read barriers when accessing
non-compressed instance fields will be generated.

Part of Issue #3338 


andrewcraik:

andrewcraik:
@0dvictor Why is the code living here rather than with the rest of the rdbar addition logic which isn't in this particular area of the codegen?

dsouzai:
@fjeremic Once Z has the `ardbari` evaluator, this code pretty much removes the need to anchor `aload`s under CS...though we still need something similar for `aload`s for statics.

0dvictor:
> Why is the code living here rather than with the rest of the rdbar addition logic which isn't in this particular area of the codegen?

If you mean the existing rdbar addition logic for compressedrefs, it locates inside `lowerCompressedRefs`, which does not apply to non-compressedrefs case.

andrewcraik:
Right I see

andrewcraik:

PR #3364, opened 2018-10-18
Override the isAnonymousClass at JITaaS server

satbir-ca:
This frontend method access J9Class members , needs to be overridden on the server to work properly.
Adding a new server message(VM_isAnonymousClass) to protobuf.

### DIFF
--- a/runtime/compiler/env/VMJ9Server.cpp
+++ b/runtime/compiler/env/VMJ9Server.cpp
@@ -1459,3 +1459,19 @@ TR_J9ServerVM::transformJlrMethodInvoke(J9Method *callerMethod, J9Class *callerC
    stream->write(JITaaS::J9ServerMessageType::VM_transformJlrMethodInvoke, callerMethod, callerClass);
    return std::get<0>(stream->read<bool>());
    }
+
+bool
+TR_J9ServerVM::isAnonymousClass(TR_OpaqueClassBlock *j9clazz)
+   {
+      {
+      OMR::CriticalSection getRemoteROMClass(_compInfoPT->getClientData()->getROMMapMonitor());
+      auto it = _compInfoPT->getClientData()->getROMClassMap().find((J9Class*) j9clazz);
+      if (it != _compInfoPT->getClientData()->getROMClassMap().end())
+         {
+         return (J9_ARE_ALL_BITS_SET(it->second.romClass->extraModifiers, J9AccClassAnonClass));
+         }
+      }
+   JITaaS::J9ServerStream *stream = _compInfoPT->getMethodBeingCompiled()->_stream;
+   stream->write(JITaaS::J9ServerMessageType::VM_isAnonymousClass, j9clazz);
+   return std::get<0>(stream->read<bool>());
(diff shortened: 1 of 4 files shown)


runtime/compiler/env/VMJ9Server.hpp
```diff
@@ -160,6 +160,8 @@ class TR_J9ServerVM: public TR_J9VM
    virtual uintptrj_t getFieldOffset(TR::Compilation * comp, TR::SymbolReference* classRef, TR::SymbolReference* fieldRef) override { return 0; } // safe answer
    virtual bool instanceOfOrCheckCast(J9Class *instanceClass, J9Class* castClass) override;
    virtual bool transformJlrMethodInvoke(J9Method *callerMethod, J9Class *callerClass) override;
+   using TR_J9VM :: isAnonymousClass;
```
mpirvu: What is the purpose of this `using` statement?

runtime/compiler/env/VMJ9Server.cpp
```diff
@@ -1459,3 +1459,19 @@ TR_J9ServerVM::transformJlrMethodInvoke(J9Method *callerMethod, J9Class *callerC
    stream->write(JITaaS::J9ServerMessageType::VM_transformJlrMethodInvoke, callerMethod, callerClass);
    return std::get<0>(stream->read<bool>());
    }
+
+bool
+TR_J9ServerVM::isAnonymousClass(TR_OpaqueClassBlock *j9clazz)
+   {
+      {
+         OMR::CriticalSection getRemoteROMClass(_compInfoPT->getClientData()->getROMMapMonitor());
```
mpirvu: In this method indentation does not follow our coding convention.

mpirvu:
Let's the drop the `the` in the title of the commit and remove ` to protobuf` from the commit message because protobuf is just a serialization mechanism.

runtime/compiler/env/VMJ9Server.hpp
satbir-ca: Once a name match is found in the a class scope, it doesn't look further up the inheritance tree for overloads. By specifying the 'using' declaration, you bring all of the overloads of 'isAnonymousClass' from 'TR_J9VM' into the scope of 'TR_J9ServerVM'. Then overloading works properly.

satbir-ca: Otherwise we have to override all the function of the same name.

mpirvu: > Then overloading works properly.

Are you saying that without `using TR_J9VM :: isAnonymousClass;` the code does not work correctly?

satbir-ca: It does not compile. The definition of the virtual bool isAnonymousClass(TR_OpaqueClassBlock *j9clazz) will be visible but the virtual bool isAnonymousClass(J9ROMClass *romClass) will not be visible.

satbir-ca:
@mpirvu updated the code with the review comments

PR #3368, opened 2018-10-19
Prepare for JVMTI ExceptionCatch not forcing FSD

gacholio:
- use getCurrentByteCodeIndexAndIsSameReceiver instead of
getJitPCOffsetFromExceptionHandler to fetch the bytecode PC of exception
handlers (this avoids reading the bytecode PC from the JIT exception
handler data which is only filled in for FSD)

- report exceptions caught by JIT if the callback is filled in rather
than explicitly checking for FSD mode

### DIFF
--- a/runtime/codert_vm/CodertVMHelpers.cpp
+++ b/runtime/codert_vm/CodertVMHelpers.cpp
@@ -27,6 +27,7 @@
 #include "VMHelpers.hpp"
 #include "AtomicSupport.hpp"
 #include "MethodMetaData.h"
+#include "ut_j9codertvm.h"
 
 extern "C" {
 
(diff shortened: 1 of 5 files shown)


gacholio:
jenkins test all zlinux jdk8

gacholio:
Note that the decomp003 test runs these codepaths, so I'm confident the change is correct.

gacholio:
Related: #3373

runtime/codert_vm/CodertVMHelpers.cpp
```diff
@@ -319,28 +320,29 @@ jitUpdateInlineAttribute(J9VMThread *currentThread, J9Class * classPtr, void *ji
...
 {
-	UDATA inlinedCatcherFound = FALSE;
+	J9Method *method = metaData->ramMethod;
+	void *stackMap = NULL;
+	void *inlineMap = NULL;
+	void *inlinedCallSite = NULL;
+	/* Note we need to add 1 to the JIT PC here in order to get the correct map at the exception handler
+	 * because jitGetMapsFromPC is expecting a return address, so it subtracts 1.  The value passed in is
+	 * the start address of the compiled exception handler.
+	 */
```
fjeremic: Thanks for the comment. This is definitely non-obvious!

PR #3383, opened 2018-10-19
Fix decoding of string parameters

keithc-ca:
`com.ibm.jvm.format.Util.constructString()` returned a string whose length was longer than the number of bytes consumed from the buffer whenever it escaped certain ASCII control characters. Then callers like `com.ibm.jvm.format.Message.processPercents()` used the length of the resulting string to compute the offset of the next parameter leaving too few bytes remaining to be decoded successfully.

In a failing test, a string parameter has the 3 character value `"\u0008\u00ef\u00f8"` (using Java literal notation) which is expanded to `"\\u0010\u00ef\u00f8"` (an 8 character string).

A second problem is that the encoding should use hex rather than octal (e.g. `"\u0008"` should not be presented as `"\u0010"`).

This fixes these problems by escaping control characters _after_ advancing the offset into the decode buffer.

### DIFF
--- a/jcl/src/openj9.traceformat/share/classes/com/ibm/jvm/format/Message.java
+++ b/jcl/src/openj9.traceformat/share/classes/com/ibm/jvm/format/Message.java
@@ -1,6 +1,6 @@
 /*[INCLUDE-IF Sidecar18-SE]*/
 /*******************************************************************************
- * Copyright (c) 2000, 2016 IBM Corp. and others
+ * Copyright (c) 2000, 2018 IBM Corp. and others
  *
  * This program and the accompanying materials are made available under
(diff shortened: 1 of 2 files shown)


pshipton:
jenkins test all plinux jdk8

pshipton:
PR testing
https://ci.eclipse.org/openj9/job/PullRequest-Sanity-JDK8-linux_ppc-64_cmprssptrs_le-OpenJ9/625/
https://ci.eclipse.org/openj9/job/PullRequest-Extended-JDK8-linux_ppc-64_cmprssptrs_le-OpenJ9/143/

pshipton:
Not sure how "\\u0010\u00ef\u00f8" is an 8 character string. Did you mean 18?

PR #3384, opened 2018-10-19
Correct path to redirector subdirectory for OSX

keithc-ca:
* mirrors changes to jvm.c in #3328


DanHeidinga:
Jenkins test sanity xlinux,osx jdk11

PR #3385, opened 2018-10-19
Use generateRXInstruction API

fjeremic:
Use generateRXInstruction API instead of manually allocating an RX
instruction as the generate API may return an RXY instruction if the
displacement is constrained.


fjeremic:
@joransiu could I please get a review and you assigned as the committer? This change is in preparation for work in OMR which uses `generate[RX|RS|SI]Instruction` APIs to check for long displacement at instruction selection time and instead return an RXY,RSY,SIY instruction.

Issue #3373, opened 2018-10-19
Agent support for exception catch information


fjeremic:
This Epic tracks the work to enable agents to request exception catch information without the JVM going into Full Speed Debug (FSD) mode. Most of the work here is on the VM side but there are JIT portions as well to ensure we have a stack map at the start of every exception catch block and to ensure we don't enable throw to goto transformations when this mode is enabled.

Other than the obvious desire to enable as many features as possible without going into FSD, the motivation behind this work is the upcoming cooperation with OverOps - their agent needs to get exception events, but can not tolerate the overhead of FSD being imposed on production systems.

fjeremic:
@gacholio I've opened up the Epic to track the work but I'm not too familiar with the VM portion or the high level. I've written down what I could but I encourage you to add some more detail here if possible. I'll open up the JIT issues for us to tackle against this Epic. I've also tagged it for 0.12.0 as I asumme the work will be done for the next release. @pshipton FYI.

gacholio:
The main issue I ran into when doing the VM side of the work  was that bytecode PC information is stored in the JIT exception handler data only in FSD mode.

- jitExceptionHandlerSearch updates userData1 with the bytecode PC if present
- getJitPCOffsetFromExceptionHandler assumes the bytecode PC data is there

Both of the above are used in the JVMTI Exception event (i.e. exception throw) to identify the catching location.

The JVMTI ExceptionCatch event already relies on the stack map at the catch point and gets the bytecode PC information from getCurrentByteCodeIndexAndIsSameReceiver. This works in FSD, but not in normal mode.

https://github.com/eclipse/openj9/pull/3368

changes the VM to always use the stack map approach, obviating the need for bytecode PC information in the JIT handler data completely. getJitPCOffsetFromExceptionHandler is no longer called from anywhere in the codebase, and the userData1 value from jitExceptionHandlerSearch is ignored.

The above change is safe to make today, given that ExceptionCatch forces FSD.

I modified the JIT not to force FSD for ExceptionCatch, and ran into the two issues that need to be addressed in the JIT:

1) Platform-dependent code tests FSD mode before generating stack maps at catch points (J9TreeEvalutator -> VMgenerateCatchBlockBBStartPrologue on X, unsure of other platforms)

2) Common code tests FSD mode and merges stack maps with different bytecode PCs (MetaData.cpp -> mapsAreIdentical)

[...]

PR #3389, opened 2018-10-19
ExceptionCatch no longer forces FSD

gacholio:
The JVMTI ExceptionCatch event no longer requires FSD. If ExceptionCatch
is enabled, fill in the function pointer that allows compiled catchers
to fix the decompilation stack and report the event.

[ci skip]

### DIFF
--- a/runtime/compiler/control/J9Options.cpp
+++ b/runtime/compiler/control/J9Options.cpp
@@ -2168,8 +2168,7 @@ bool J9::Options::feLatePostProcess(void * base, TR::OptionSet * optionSet)
        (*vmHooks)->J9HookDisable(vmHooks, J9HOOK_VM_PUT_FIELD) ||
        (*vmHooks)->J9HookDisable(vmHooks, J9HOOK_VM_GET_STATIC_FIELD) ||
        (*vmHooks)->J9HookDisable(vmHooks, J9HOOK_VM_PUT_STATIC_FIELD) ||
-       (*vmHooks)->J9HookDisable(vmHooks, J9HOOK_VM_SINGLE_STEP) ||
-       (*vmHooks)->J9HookDisable(vmHooks, J9HOOK_VM_EXCEPTION_CATCH))
+       (*vmHooks)->J9HookDisable(vmHooks, J9HOOK_VM_SINGLE_STEP))
       {
         static bool TR_DisableFullSpeedDebug = feGetEnv("TR_DisableFullSpeedDebug")?1:0;
       #if defined(J9VM_JIT_FULL_SPEED_DEBUG)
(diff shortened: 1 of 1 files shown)


gacholio:
Marking WIP as this cannot be merged until the JIT work is done.

Realted: #3373

runtime/compiler/control/J9Options.cpp
```diff
@@ -2180,8 +2179,17 @@ bool J9::Options::feLatePostProcess(void * base, TR::OptionSet * optionSet)
       #endif
       }
 
-   if ((*vmHooks)->J9HookDisable(vmHooks, J9HOOK_VM_EXCEPTION_CATCH) ||
-       (*vmHooks)->J9HookDisable(vmHooks, J9HOOK_VM_EXCEPTION_THROW))
+   bool disableThrowToGoto = false;
+   if ((*vmHooks)->J9HookDisable(vmHooks, J9HOOK_VM_EXCEPTION_CATCH))
+      {
+      jitConfig->jitExceptionCaught = jitExceptionCaught;
```
fjeremic: Perhaps I'm not understanding this correctly but why are we not always initializing this and checking in the VM [1] whether the hook is actually enabled and calling this function? Unless we should only be calling `jitExceptionCaught` under FSD or when the catch hook is enabled?

It also appears the comment regarding FSD in [1] has not aged well and should be fixed, since after this change it's not only for FSD but for hooks as well.

[1] https://github.com/eclipse/openj9/blob/3ab115dd60822be79b146721059e025b64a0e531/runtime/vm/BytecodeInterpreter.hpp#L2427-L2434

gacholio: jitExceptionCaught does several things:

- keeps the decompilation stack in sync
- reports the event
- fixes decompile if necessary

It's only filled in if the exception catch event is being reported, or FSD is enabled (it's set in intializeFSD as well). It's not sufficient to test this for the new feature - we need proper stack maps if exception throw is hooked as well (this is currently a bug in the JIT - catch should be forcing FSD at the moment, but isn't).

fjeremic: Thanks for the explanation. I've suggested an alternate route in #3462 where the JIT query should be implemented. We just need to settle on a query on the JIT side and the rest will follow easily. I'll push on the JIT front so we can wrap this work up.

runtime/compiler/control/J9Options.cpp
```diff
@@ -2180,8 +2179,17 @@ bool J9::Options::feLatePostProcess(void * base, TR::OptionSet * optionSet)
       #endif
       }
 
-   if ((*vmHooks)->J9HookDisable(vmHooks, J9HOOK_VM_EXCEPTION_CATCH) ||
-       (*vmHooks)->J9HookDisable(vmHooks, J9HOOK_VM_EXCEPTION_THROW))
+   bool disableThrowToGoto = false;
+   if ((*vmHooks)->J9HookDisable(vmHooks, J9HOOK_VM_EXCEPTION_CATCH))
+      {
```
0dvictor: FYI, once #3463 is merged, we also need set `J9HookExceptionCatch` here. i.e. adding one line `self()->enableHook(J9HookExceptionCatch);`.

gacholio: @0dvictor Don't forget we need to maintain the stack maps if either exception catch or throw are hooked.

0dvictor: FYI #3463 has merged. We need set `self()->setReportByteCodeInfoAtCatchBlock();` here and for throw hook.

gacholio: Thanks, this PR has been updated to include the new call.

fjeremic:
@gacholio I believe this can be unmarked now that all the JIT support has been delivered.

gacholio:
Testing it now.

gacholio:
New test (as well as everything else) seems to be passing on all architectures - ready for review.

fjeremic:
Jenkins test sanity all JDK8

PR #3396, opened 2018-10-22
Do not use 64-bit atomic intrinsics on 32-bit platform

liqunl:
64-bit atomic intrinsics are not implemented on 32-bit platforms, using
them will result in an illegal instruction at run time.

This is a workaround, issue #3404 is created to track the complete fix.

### DIFF
--- a/runtime/compiler/optimizer/UnsafeFastPath.cpp
+++ b/runtime/compiler/optimizer/UnsafeFastPath.cpp
@@ -80,11 +80,13 @@ static TR::SymbolReferenceTable::CommonNonhelperSymbol equivalentAtomicIntrinsic
    switch (rm)
       {
       case TR::sun_misc_Unsafe_getAndSetInt:
-      case TR::sun_misc_Unsafe_getAndSetLong:
            return TR::SymbolReferenceTable::atomicSwapSymbol;
+      case TR::sun_misc_Unsafe_getAndSetLong:
+           return TR::Compiler->target.is64Bit() ? TR::SymbolReferenceTable::atomicSwapSymbol : TR::SymbolReferenceTable::lastCommonNonhelperSymbol;
       case TR::sun_misc_Unsafe_getAndAddInt:
-      case TR::sun_misc_Unsafe_getAndAddLong:
            return TR::SymbolReferenceTable::atomicFetchAndAddSymbol;
+      case TR::sun_misc_Unsafe_getAndAddLong:
+           return TR::Compiler->target.is64Bit() ? TR::SymbolReferenceTable::atomicFetchAndAddSymbol : TR::SymbolReferenceTable::lastCommonNonhelperSymbol;
       default:
          break;
       }


liqunl:
@andrewcraik ci build is not triggered because github is repairing a data storage system https://status.github.com/messages

andrewcraik:
Given there are two 64bit sensitive operations wouldn't it make sense to pull them out into their own switch inside and is64bit test? That would make the 64bit dependence more clear than a ternary a the return (at least in my opinion)...

runtime/compiler/optimizer/UnsafeFastPath.cpp
```diff
@@ -80,11 +80,13 @@ static TR::SymbolReferenceTable::CommonNonhelperSymbol equivalentAtomicIntrinsic
...
       case TR::sun_misc_Unsafe_getAndSetInt:
-      case TR::sun_misc_Unsafe_getAndSetLong:
            return TR::SymbolReferenceTable::atomicSwapSymbol;
+      case TR::sun_misc_Unsafe_getAndSetLong:
+           return TR::Compiler->target.is64Bit() ? TR::SymbolReferenceTable::atomicSwapSymbol : TR::SymbolReferenceTable::lastCommonNonhelperSymbol;
       case TR::sun_misc_Unsafe_getAndAddInt:
-      case TR::sun_misc_Unsafe_getAndAddLong:
            return TR::SymbolReferenceTable::atomicFetchAndAddSymbol;
+      case TR::sun_misc_Unsafe_getAndAddLong:
+           return TR::Compiler->target.is64Bit() ? TR::SymbolReferenceTable::atomicFetchAndAddSymbol : TR::SymbolReferenceTable::lastCommonNonhelperSymbol;
```
fjeremic: IMO multiple return points make it hard see the exit points of functions where there is a default "do nothing" return value. Perhaps this may be more readible:

```
static TR::SymbolReferenceTable::CommonNonhelperSymbol equivalentAtomicIntrinsic(TR::RecognizedMethod rm)
   {
   auto result = TR::SymbolReferenceTable::lastCommonNonhelperSymbol;

   switch (rm)
      {
         result = TR::SymbolReferenceTable::atomicSwapSymbol;
         break;
      case TR::sun_misc_Unsafe_getAndSetLong:
            {
            result = TR::SymbolReferenceTable::atomicSwapSymbol;
            }
         break;
      case TR::sun_misc_Unsafe_getAndAddInt:
         result =  TR::SymbolReferenceTable::atomicFetchAndAddSymbol;
         break;
      case TR::sun_misc_Unsafe_getAndAddLong:
            {
            result = TR::SymbolReferenceTable::atomicFetchAndAddSymbol;
            }
        break;
      default:
         break;
      }

   return result;
}
```

AdamBrousseau:
Jenkins copyright check

fjeremic:
Also we need a space in the title. "Donot" should be "Do not".

runtime/compiler/optimizer/UnsafeFastPath.cpp
andrewcraik: that is one style - some of the JIT recognition paths would be more complicated with a single return style so a multi return can be more clear, but making the 64bit condition clear is the main request I have rather than using an easily missed inline ternary.

AdamBrousseau:
Jenkins copyright check

liqunl:
@andrewcraik The two switch statements will have the same cases, maintaining two switch statements is more prone to defects IMO. Plus the function is used only in UnsafeFastPath, it is not a public API. And I expect the change to be reverted as soon as we support 64bit data on ia32.

liqunl:
Created https://github.com/eclipse/openj9/issues/3404 to as the todo once the missing implementation is added.

andrewcraik:
Can we reference the issue in the PR message please for completeness? Otherwise I think I'm fine with the form of the change as a temporary workaround.

andrewcraik:
@fjeremic any further comments - the style is a little less critical in my opinion given the temporary nature of what is being proposed while OMR support is added for the 64bit on 32bit case...

andrewcraik:
Jenkins test sanity xlinux,xlinux32,win jdk8,jdk11

andrewcraik:

fjeremic:
> @fjeremic any further comments - the style is a little less critical in my opinion given the temporary nature of what is being proposed while OMR support is added for the 64bit on 32bit case...

No objections here. Please proceed.

PR #3402, opened 2018-10-22
Fake region should have containsInternalCycles set

liqunl:
Since the construction of a fake region does not do a full structural
analysis, it should always assume the method contains internal cycles
such that the data flow analysis check for every change for correctness.


andrewcraik:
Given the likely performance implications of this in terms of compile-time could you provide details of your compile-time evaluation? If the compile overhead is significant would there be benefit from using the compilation's mayHaveLoops test to gate setting this? If we have a more narrow setting does that improve the overhead?

andrewcraik:
So the overhead was in the noise margin when measuring DayTrader3? if so what was the noise margin as a % for the compile-time and footprint when you measured it?

andrewcraik:

PR #3407, opened 2018-10-22
Fold away TR_DisableLongDispStackSlot as false

fjeremic:
In preparation for folding the query away in OMR we fold away the
TR_DisableLongDispStackSlot as false, meaning it is always available.

Issue: https://github.com/eclipse/omr/issues/2848


fjeremic:
@joransiu could I please get a review and a merge? This is in preparation for https://github.com/eclipse/omr/issues/2848 and a whole bunch of work coming in OMR to handle long displacement properly. The long displacement slot is always available and we should always use it given how infrequently we need huge displacement support.

fjeremic:
Given the simplicity of the change I'm preemptively launching builds here.

PR #3408, opened 2018-10-23
Ensure all ICF flags are on labels in z/codegen

dhong44:
Ensure all ICF flags are on labels in z/codegen


dhong44:
@fjeremic Could I please get a review

dhong44:
This is part of https://github.com/eclipse/omr/issues/2833 and should be merged before setting ICF flags on labels is deprecated.

runtime/compiler/z/codegen/J9BCDTreeEvaluator.cpp
```diff
@@ -267,7 +267,7 @@ J9::Z::TreeEvaluator::udsl2pdEvaluator(TR::Node *node, TR::CodeGenerator *cg)
 
    if (!node->getOpCode().isSetSign())
       {
-      TR::LabelSymbol * startLabel = TR::LabelSymbol::create(cg->trHeapMemory(),cg);
+      TR::LabelSymbol * cFlowRegionStart = TR::LabelSymbol::create(cg->trHeapMemory(),cg);
       TR::LabelSymbol * doneLabel = TR::LabelSymbol::create(cg->trHeapMemory(),cg);
```
fjeremic: According to https://github.com/eclipse/omr/issues/2833#issuecomment-432058904 this one should have been `cFlowRegionEnd` as it doesn't have an otherwise meaningful name. There are several occurrences of this. Do we need to be more consistent?

Also while we are in the area of modifying these names could you also change all the `TR::LabelSymbol::create` calls to `generateLabelSymbol` calls which doesn't need memory passed?

fjeremic:
@dhong44 as a high level comment, would it be possible to use the same commit title for both OMR and OpenJ9 changes given they're pretty much the same change had OMR and OpenJ9 lived in the same repository. It makes searching much easier.

dhong44:
Sure, I've changed the name of this PR to match https://github.com/eclipse/omr/pull/3106 and also the issue in omr, https://github.com/eclipse/omr/issues/2833.

runtime/compiler/z/codegen/J9BCDTreeEvaluator.cpp
dhong44: I've tried to replace all the names of labels that have end ICF flags with `cFlowRegionEnd` where applicable in b34fad6, hopefully I haven't missed any. Regarding your second, I've replaced all calls to `TR::LabelSymbol::create` with `generateLabelSymbol` in 0a8b4ca.

fjeremic:

fjeremic:
Argh! There was a merge conflict which I've resolved in 6ccbc2b but I'm not a fan of that commit polluting the commit history. I'll let the builds finish but I'd like to squash that commit away before we merge. I'll follow the build and approve once they pass, but I would like to see that commit squashed away before we merge. @dhong44 can you take care of this?

dhong44:
Looks like the in these tests is affecting other open pull requests right now as well. Seems like it's unrelated to my changes. I will rebase and squash like you've suggested.

PR #3414, opened 2018-10-23
Finish adding SETUP_LABEL param to all Pipeline and Build jobs

AdamBrousseau:
- Switch from LABEL to SETUP_LABEL
- Pass SETUP_LABEL down from PBT-ALL to Pipelines and Builds
- Add SETUP_LABEL to the Param list
- This will allow jobs to stop doing setup and Master and move it
  to worker machine(s)
- Work was originally done in PBT-All (#2160), this adds it to the rest

[skip ci]


AdamBrousseau:
Violeta questioned whether we should restrict infra type tasks to this single worker machine (with 4 executors) or if we should allow things to run on any node (other than possibly aix/win/zos). My response was that question is somewhat of a separate issue than the scope of this change. We've created the `worker` machine at Eclipse/OpenJ9 Jenkins specifically for these tasks because we are limited on machine resources and didn't want to block "real" builds. Perhaps with more machines coming online we should reconsider running these tasks on "any" node, but I still think that should be a separate discussion.

buildenv/jenkins/jobs/pipelines/Pipeline-Build-Test-All
```diff
@@ -224,6 +224,7 @@ def build(OPENJDK_REPO, OPENJDK_BRANCH, SHAS, OPENJ9_REPO, OPENJ9_BRANCH, OMR_RE
                     string(name: 'SLACK_CHANNEL', value: SLACK_HANDLE),
                     string(name: 'RESTART_TIMEOUT', value: RESTART_TIMEOUT),
                     string(name: 'RESTART_TIMEOUT_UNITS', value: RESTART_TIMEOUT_UNITS)]
```
vsebe: This is not the last element of the array, remove `]` add `,`:
`string(name: 'RESTART_TIMEOUT_UNITS', value: RESTART_TIMEOUT_UNITS),`

AdamBrousseau: Good catch, likely a result of a merge conflict. Fixed.

AdamBrousseau:
@pshipton for merge.

PR #3415, opened 2018-10-23
WIP: Move compiler-specific fields of jitConfig to jitConfigPrivate

dmitry-ten:
J9JITConfig should be used to store properties that are accessed
by both VM and compiler. However, many of its fields are used
only inside the compiler code, so there is no reason for them
to be in jitConfig.
Move all such fields from jitConfig to jitConfigPrivate.


0xdaryl:
While I agree with and support the spirit of this PR, I do want to caution that we have experienced performance issues in the past when moving jitConfig fields around (due to cache misses mostly).  While you may actually realize some improvements with a denser, JIT-specific packing, I do think a performance run on all platforms is warranted so that in the event that something slows down you can take steps to mitigate it now.

andrewcraik:
A measurement of throughput, compile-time and footprint is certainly called for on a suite of benchmarks to ensure there is no impact on high opt, low opt, or other configs.

mpirvu:
> I do want to caution that we have experienced performance issues in the past when moving jitConfig fields around (due to cache misses mostly)

@0xdaryl Do you remember what benchmarks were affected so that we can try those first?

0xdaryl:
It's been a long while so my recollection is a bit fuzzy, but my focus at the time was on WAS/Liberty/DayTrader.  Ultimately, however, the modern workloads that OpenJ9 currently cares about for performance are what matters for evaluating this.

dmitry-ten:
It seems that a lot of the function pointers from jitConfig are not used anywhere.
Can they be removed?
Some examples: `hasMoreInlinedMethods`, `getByteCodeIndex`, `doAESInHardware`

mpirvu:
Do we even have a definition for those functions? If not I would say, yes, remove them.

dmitry-ten:
Removed unused function pointers.

PR #3419, opened 2018-10-23
Fold away only use of ExtCodeBase register

fjeremic:
Since long displacement stack slot is being folded away as always being
available this means the extended code base register is never needed
and is hence slated for deprecation.

All paths using the extended code base register have either a check for
whether we have long displacement slot available (always true) or a
check for `isExtCodeBaseFreeForAssignment` which under the hood also
returns false if long long displacement slot is available.


fjeremic:
Depends on https://github.com/eclipse/omr/pull/3113.

@joransiu leaving this one as WIP to prevent accidental merge. We'll need to wait until the OMR change propagates to avoid build breaks.

fjeremic:
We just had an OMR-Acceptance promotion with the right SHA. Unleashing builds:

Issue #3273, opened 2018-10-15
JTreg test fail : java/lang/invoke/8177146/TestMethodHandleBind.java


ben-walsh:
Test https://github.com/ibmruntimes/openj9-openjdk-jdk11/blob/openj9/test/jdk/java/lang/invoke/8177146/TestMethodHandleBind.java has two failures ... 

```
test TestMethodHandleBind.testInstanceOfReceiverClass(): failure
java.lang.AssertionError: IllegalAccessException expected
	at org.testng.Assert.fail(Assert.java:94)
	at TestMethodHandleBind.testInstanceOfReceiverClass(TestMethodHandleBind.java:61)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.base/java.lang.reflect.Method.invoke(Method.java:566)
	at org.testng.internal.MethodInvocationHelper.invokeMethod(MethodInvocationHelper.java:85)
	at org.testng.internal.Invoker.invokeMethod(Invoker.java:639)
	at org.testng.internal.Invoker.invokeTestMethod(Invoker.java:821)
	at org.testng.internal.Invoker.invokeTestMethods(Invoker.java:1131)
	at org.testng.internal.TestMethodWorker.invokeTestMethods(TestMethodWorker.java:125)
	at org.testng.internal.TestMethodWorker.run(TestMethodWorker.java:108)
	at org.testng.TestRunner.privateRun(TestRunner.java:773)
	at org.testng.TestRunner.run(TestRunner.java:623)
	at org.testng.SuiteRunner.runTest(SuiteRunner.java:357)
	at org.testng.SuiteRunner.runSequentially(SuiteRunner.java:352)
	at org.testng.SuiteRunner.privateRun(SuiteRunner.java:310)
	at org.testng.SuiteRunner.run(SuiteRunner.java:259)
	at org.testng.SuiteRunnerWorker.runSuite(SuiteRunnerWorker.java:52)
	at org.testng.SuiteRunnerWorker.run(SuiteRunnerWorker.java:86)
	at org.testng.TestNG.runSuitesSequentially(TestNG.java:1185)
[...]

PR #3423, opened 2018-10-23
Refactor MH.bind to use standard handle search

fengxue-IS:
- remove internalPrivilegedLookup call
- always restrict the receiver of 'bind'

Close: #3273 

### DIFF
--- a/jcl/src/java.base/share/classes/java/lang/invoke/MethodHandles.java
+++ b/jcl/src/java.base/share/classes/java/lang/invoke/MethodHandles.java
@@ -253,10 +253,13 @@ public class MethodHandles {
 			Class<?> receiverClass = receiver.getClass();
 			MethodHandle handle = handleForMHInvokeMethods(receiverClass, methodName, type);
 			if (handle == null) {
-				// Use the priviledgedLookup to allow probe the findSpecial cache without restricting the receiver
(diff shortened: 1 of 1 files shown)


fengxue-IS:
> What about the lost caching in findSpecialImpl?

Can you clarify what you meant by lost caching, this change should not have affected how the MethodHandle caching works, it only removes lookup privilege so receiver of the handle will always be restricted

fengxue-IS:
@DanHeidinga can you take another look?

DanHeidinga:
The path through `findSpecialImpl` is something like:
```
* Create DirectHandle
* Restrict receiver if not internalPrivilegedLookup  (new MH creation)
* Convert to varargs (new MH)
* cache
```

and then in `bind`,
```
* findSpecial
* Remove varargs (grr!)
* call bindTo which
**  If DirectHandle, replace with ReceiverBoundHandle, else MHs.InsertArgs
```

By not using the `internalPrivilegedLookup`, we add an extra MH creation that will be effectively removed later if we hit the DH->RBH case in `bind`.

While I'm not 100% happy with this approach, it does make use consistent with the RI.

I'll launch testing on it and if it passes, we'll merge this and we can look at improving that path in the future.

DanHeidinga:
Jenkins test all zlinux jdk11

PR #3425, opened 2018-10-24
Task thread count

amicic:
Comply with new API for prepareThreadsForTask (part of Dispatcher
framework to run a parallel GC task) that takes an extra argument that
specifies explict parallel thread count per Task basis. 

Metronome does not really make use of this API, but the old one is to be
removed.


dmitripivkine:

PR #3427, opened 2018-10-24
Move job result status echos into common function area

AdamBrousseau:
- We had echos in workflow() as well as build_with_slack(),
  move all to build_with_slack().
- Messages were confusing becasue we would get an echo for
  WARNING: unstable, then an echo for PASSED on the same job.
- Standardize verbiage a bit more.
- Remove try/catch around downstream job method calls as
  they are unnecessary.

[skip ci]


pshipton:
There is a merge conflict.

Issue #2856, opened 2018-09-13
Add tests for "Reject illegal entries in the bootstrap argument array"


DanHeidinga:
With the changes in https://github.com/eclipse/openj9/pull/2700, openj9 correctly handles condy's in the bsm list.  We should add some tests that validate that other cp items are rejected.

DanHeidinga:
@ChengJin01 Can you take a look at this?

ChengJin01:
Will figure out how to write these tests (might need  asm tools to get this work at the bytecode level)

PR #3428, opened 2018-10-24
Test cases for BoostrapMethod arguments

ChengJin01:
The changes are to add some test cases  to check
that invalid constant pool entries stored in 
BoostrapMethod arguments must be captured during 
static verification.

issue: #2856

### DIFF
--- a/test/functional/cmdLineTests/bootstrapMethodArgumentTest/playlist.xml
+++ b/test/functional/cmdLineTests/bootstrapMethodArgumentTest/playlist.xml
@@ -1,0 +1,48 @@
+<?xml version="1.0" encoding="UTF-8"?>
+
+<!--
+  Copyright (c) 2019, 2019 IBM Corp. and others
+
+  This program and the accompanying materials are made available under
+  the terms of the Eclipse Public License 2.0 which accompanies this
+  distribution and is available at https://www.eclipse.org/legal/epl-2.0/
+  or the Apache License, Version 2.0 which accompanies this distribution and
+  is available at https://www.apache.org/licenses/LICENSE-2.0.
+
+  This Source Code may also be made available under the following
+  Secondary Licenses when the conditions for such availability set
+  forth in the Eclipse Public License, v. 2.0 are satisfied: GNU
+  General Public License, version 2 with the GNU Classpath
+  Exception [1] and GNU General Public License, version 2 with the
+  OpenJDK Assembly Exception [2].
+
+  [1] https://www.gnu.org/software/classpath/license.html
+  [2] http://openjdk.java.net/legal/assembly-exception.html
+
+  SPDX-License-Identifier: EPL-2.0 OR Apache-2.0 OR GPL-2.0 WITH Classpath-exception-2.0 OR LicenseRef-GPL-2.0 WITH Assembly-exception
+-->
+
+<playlist xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:noNamespaceSchemaLocation="../../../TestConfig/playlist.xsd">
+	<test>
+		<testCaseName>cmdLineTester_bootstrapMethodArgumentTest</testCaseName>
+		<variations>
+			<variation>NoOptions</variation>
+		</variations>
+		<command>$(JAVA_COMMAND) $(JVM_OPTIONS) -DJAVA_EXE=$(JAVA_COMMAND) \
+		-DTESTSJARPATH=$(Q)$(TEST_RESROOT)$(D)bsmargstest.jar$(Q) \
+		-DRESJAR=$(CMDLINETESTER_RESJAR) -jar $(CMDLINETESTER_JAR) \
+		-config $(Q)$(TEST_RESROOT)$(D)bsmargstest.xml$(Q) \
+		-xids all,$(PLATFORM) \
+		-nonZeroExitWhenError; \
+		$(TEST_STATUS)</command>
+		<levels>
+			<level>sanity</level>
+		</levels>
+		<groups>
+			<group>functional</group>
+		</groups>
+		<subsets>
+			<subset>11</subset>
+		</subsets>
+	</test>
+</playlist>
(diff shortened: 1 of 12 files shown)


ChengJin01:
The test cases are written in .jcod files which is based on the 
RejectInvalidCPTestTemplate.jcod with one line modification as follows:
```
    Attr(#70) { // BootstrapMethods
      [] { // bootstrap_methods
        {  //  bootstrap_method
          #71; // bootstrap_method_ref
          [] { // bootstrap_arguments
            #XX;  <---------------Modified to an invalid constant pool entry (UTF8, Fieldref, etc)
          }  //  bootstrap_arguments
        }  //  bootstrap_method
      }
```
The test set covers all invalid entries against VM Spec:
```
CONSTANT_Utf8
CONSTANT_Fieldref
CONSTANT_Methodref
CONSTANT_NameAndType
CONSTANT_InvokeDynamic
CONSTANT_InterfaceMethodref
```
except
```
CONSTANT_Module
CONSTANT_Package
```
which only occurs in module-info.class

Test cases were already verified manually and in personal build:
[manual_run.txt](https://github.com/eclipse/openj9/files/2511332/manual_run.txt)
[test_log_personal_build.txt](https://github.com/eclipse/openj9/files/2511346/test_log_personal_build.txt)

ChengJin01:
Test reviewer: @DanHeidinga 
Test framework reviewer: @renfeiw

test/functional/cmdLineTests/bootstrapMethodArgumentTest/playlist.xml
```diff
@@ -0,0 +1,49 @@
...
+		-nonZeroExitWhenError; \
+		$(TEST_STATUS)</command>
+		<levels>
+			<level>sanity</level>
+		</levels>
+		<groups>
+			<group>functional</group>
+		</groups>
+		<subsets>
+			<subset>SE110</subset>
```
renfeiw: SE version will be deprecated soon, could you please change it to `<subset>11</subset>`.

ChengJin01: Already updated and verified in personal build.

test/functional/cmdLineTests/bootstrapMethodArgumentTest/playlist.xml
```diff
@@ -0,0 +1,49 @@
...
+<playlist xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:noNamespaceSchemaLocation="../../../TestConfig/playlist.xsd">
+	<test>
+		<testCaseName>cmdLineTester_bootstrapMethodArgumentTest</testCaseName>
+		<variations>
+			<variation>NoOptions</variation>
+		</variations>
+		<command>$(JAVA_COMMAND) $(JVM_OPTIONS) -DJAVA_EXE=$(JAVA_COMMAND) -DJAVA_HOME=$(Q)$(JAVA_HOME)$(Q) \
+		-DJAVATEST_ROOT=$(JAVATEST_ROOT) -DJVM_TEST_ROOT=$(JVM_TEST_ROOT) \
+		-DTESTSJARPATH=$(Q)$(TEST_RESROOT)$(D)bsmargstest.jar$(Q) \
+		-DRESJAR=$(CMDLINETESTER_RESJAR) -jar $(CMDLINETESTER_JAR) \
```
renfeiw: It seems that you don't really need to pass in JAVA_HOME, JAVATEST_ROOT, JVM_TEST_ROOT and RESJAR. Other than that, the test framework part looks good to me.

ChengJin01: Already removed.

DanHeidinga:

DanHeidinga:
Jenkins test sanity osx jdk11

DanHeidinga:
@ChengJin01 What's the issue with the default format string?  It accepts all the correct arguments...

ChengJin01:
> What's the issue with the default format string? It accepts all the correct arguments...

@DanHeidinga, 
Originally, there should be no problem with the the default format string as it returns the whole string template. But the logic changed with eclipse/omr#3912, in which case only the format string returns if it is non-NULL instead of the whole message template (it needs to call `parse_catalog `to get the whole template), the same issue detected in test cases above.

e.g. J9NLS_BCV_ERROR_TEMPLATE_NO_PC=`%1$s; class=%3$.*2$s, method=%5$.*4$s%7$.*6$s`
the passed-in default format string is `%s;%.*s,%.*s%.*s`.
1) originally it returns the whole template: `%1s; class=%.*s, method=%.*s%.*s`
2) with eclipse/omr#3912, it only returns the format string: `%s;%.*s,%.*s%.*s`, without `class=` and `method=`. In the final error message, it looks like something like `JVMCRFY114; classA, foo` instead of
`JVMCRFY114; class=classA, method=foo  ` 

So we need to do the same thing (pass in a NULL default string to `j9nls_lookup_message `to avoid skipping `parse_catalog`) to get rid of these format issues.

DanHeidinga:
> So we need to do the same thing (pass in a NULL default string to j9nls_lookup_message to avoid skipping parse_catalog) to get rid of these format issues.

Or pass in the english string as the default string.  This gets the right format and avoids the NLS lookup in english locales

ChengJin01:
> Or pass in the english string as the default string. This gets the right format and avoids the NLS lookup in english locales

Passing in the whole string template as the default should work in this case. What about the cases of non-English locale? 

e.g.  for J9NLS_BCV_ERROR_TEMPLATE_NO_PC=`%1$s; class=%3$.*2$s, method=%5$.*4$s%7$.*6$s`
we just directly pass in `%1s; class=%.*s, method=%.*s%.*s` in order to return the same template for message generation in term of the English locale. 
For non-English locale, it still goes to `parse_catalog `with `%1s; class=%.*s, method=%.*s%.*s`. I am not too sure whether it returns the correct template including the non-English translation of  `class=` and `method=`.

DanHeidinga:
The OMR change was to not lookup the NLS message in the case where:
* there is a default message
* the VM is in an english locale

For non-english locales the VM will still lookup the message.

ChengJin01:
> For non-english locales the VM will still lookup the message.

If so, there should be no problem with the whole string template as the default string. I will create a PR to fix all the code involved.

PR #3429, opened 2018-10-24
Deprecate ConcurrentHashMap TM support in the JIT

fjeremic:
The Java class support does not exist in OpenJ9 and we've found from
past experimentation that the ConcurrentHashMap transactional memory
support did not provide a measurable benefit.

Given the maintenance cost and the fact that the code has not been
running in the last several years we are dropping the JIT support.


fjeremic:
@vijaysun-omr, @zl-wang, @joransiu could I get a formal approval for this one from each of you? Thanks!

runtime/compiler/p/codegen/J9CodeGenerator.cpp
```diff
@@ -351,19 +351,6 @@ bool J9::Power::CodeGenerator::suppressInliningOfRecognizedMethod(TR::Recognized
       return true;
       }
 
-   // Transactional Memory
-   if (self()->getSupportsTM())
-      {
-      if (method == TR::java_util_concurrent_ConcurrentHashMap_tmEnabled ||
-          method == TR::java_util_concurrent_ConcurrentHashMap_tmPut ||
-          method == TR::java_util_concurrent_ConcurrentLinkedQueue_tmOffer ||
```
zl-wang: Need to keep CLQ methods around?  A little too much are removed here.

fjeremic: This is a mistake. Thanks for catching that. I've generated a diff of the change before and after and looked at all the + and - lines and checked that I'm not making a similar mistake elsewhere. Fixed in 5e429ad

fjeremic:
Given two approvals I'm preemptively launching builds for this @vijaysun-omr. I've also assigned you as the commiter since you are the last person who needs to sign off.

Jenkins test sanity all JDK8

fjeremic:
@vijaysun-omr I've rebased to eliminate the merge conflict that arose since this PR was opened.

Jenkins test sanity all JDK8

fjeremic:
@vijaysun-omr any other comments for this one?

vijaysun-omr:
LGTM. Thanks for cleaning up the (not small) amounts of code we had for this, @fjeremic

PR #3432, opened 2018-10-24
Remove setting specialized for direction flag in Z CodeGen

r30shah:
Remove setting flag that Value Propagation checks to generate array copy trees for specialized direction with test to decide direction as it is not java specific.
 

r30shah:
Depends on https://github.com/eclipse/omr/pull/3118 so marking it WIP. FYI @fjeremic

PR #3433, opened 2018-10-24
SPMDParallelizer BadILOp Fix

R-Santhir:
Prior to this fix a BadILOp node will be valid for SIMDizing if the node's first
child is loop invariant. Now if the treetop is a BadILOp it will not be
processed to be SIMDized

### DIFF
--- a/runtime/compiler/optimizer/SPMDParallelizer.cpp
+++ b/runtime/compiler/optimizer/SPMDParallelizer.cpp
@@ -492,6 +492,12 @@ bool TR_SPMDKernelParallelizer::visitTreeTopToSIMDize(TR::TreeTop *tt, TR_SPMDKe
                }
             }
 
+         TR::ILOpCode scalarOp = node->getOpCode();
+         TR::ILOpCodes vectorOpCode = TR::ILOpCode::convertScalarToVector(scalarOp.getOpCodeValue());
+
+         if (isCheckMode && vectorOpCode == TR::BadILOp)
+            return false;
+
          if (loop->isExprInvariant(node->getFirstChild()))
             {
             if (isCheckMode)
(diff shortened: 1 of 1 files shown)


R-Santhir:
@gita-omr @IBMJimmyk

IBMJimmyk:
Did you intent to move the check or duplicate the check?  It looks like the check is done again in the old location as well.

R-Santhir:
That was my mistake, let me change that, did not intend to duplicate it

runtime/compiler/optimizer/SPMDParallelizer.cpp
```diff
@@ -492,6 +492,14 @@ bool TR_SPMDKernelParallelizer::visitTreeTopToSIMDize(TR::TreeTop *tt, TR_SPMDKe
                }
             }
 
+         TR::ILOpCode scalarOp = node->getOpCode();
+         TR::ILOpCodes vectorOpCode = TR::ILOpCode::convertScalarToVector(scalarOp.getOpCodeValue());
+         TR::SymbolReference *symRef = node->getSymbolReference();
+         TR::SymbolReference *vecSymRef = pSPMDInfo->getVectorSymRef(symRef);
+
+         if (isCheckMode && vectorOpCode == TR::BadILOp)
```
andrewcraik: would it be better for this to read more like:
```if (vectorOpCode == TR::BadIlOp)
    {
    TR_ASSERT_FATAL(isCheckMode, "BadILOp encountered during SPMD parallelization which was not seen during checking mode at node n%dn!", node->getGlobalIndex());
     return false;
    }```
This would guard against a bad IL op that might pop up later

Comment (gita-omr) [reply runtime/compiler/optimizer/SPMDParallelizer.cpp]: We have an assert for this purpose few lines down. Maybe we should make it FATAL. But please note that this was not the problem in this case. In this case, we skipped the check all together for nodes whose RHS is loop invariant.

File: runtime/compiler/optimizer/SPMDParallelizer.cpp [hunk 2]
```diff
@@ -520,14 +528,6 @@ bool TR_SPMDKernelParallelizer::visitTreeTopToSIMDize(TR::TreeTop *tt, TR_SPMDKe
...
 
-         TR::ILOpCode scalarOp = node->getOpCode();
-         TR::ILOpCodes vectorOpCode = TR::ILOpCode::convertScalarToVector(scalarOp.getOpCodeValue());
-         TR::SymbolReference *symRef = node->getSymbolReference();
-         TR::SymbolReference *vecSymRef = pSPMDInfo->getVectorSymRef(symRef);
-
-         if (isCheckMode && vectorOpCode == TR::BadILOp)
-            return false;
-
          TR_ASSERT(vectorOpCode != TR::BadILOp, "BAD IL Opcode to be assigned during transformation");
```
gita-omr: Please make it FATAL assert

runtime/compiler/optimizer/SPMDParallelizer.cpp
andrewcraik: I see - if there is an assert later on that is fine - I was just trying to make sure the BadILOp was caught with an assert so I'm happy to yield to the judgement of @gita-omr on whether this is needed. Making the assert fatal may help ensure we get compile crashes at the time of generating a bad opcode rather than falling over later in codegen trying to evaluate it.

gita-omr:
Jenkins test all aix,plinux,zlinux jdk8,jdk11

PR #3434, opened 2018-10-24
Fix SharedCache Throughput Regression with IPdata

harryyu1994:
Currently when we send iprofile data from client to server,
we only grab ipdata from the hashtable. Changing it to grab ipdata from
hashtable and also the sharedcache.

[skip ci]

### DIFF
--- a/runtime/compiler/runtime/JITaaSIProfiler.cpp
+++ b/runtime/compiler/runtime/JITaaSIProfiler.cpp
@@ -383,7 +383,7 @@ TR_JITaaSClientIProfiler::TR_JITaaSClientIProfiler(J9JITConfig *jitConfig)
  */
 uint32_t
 TR_JITaaSClientIProfiler::walkILTreeForIProfilingEntries(uintptrj_t *pcEntries, uint32_t &numEntries, TR_J9ByteCodeIterator *bcIterator,
-                                                       TR_OpaqueMethodBlock *method, TR_BitVector *BCvisit, bool &abort)
+                                                       TR_OpaqueMethodBlock *method, TR_BitVector *BCvisit, bool &abort, TR::Compilation *comp)
    {
    abort = false; // optimistic
    uint32_t bytesFootprint = 0;
@@ -395,7 +395,7 @@ TR_JITaaSClientIProfiler::walkILTreeForIProfilingEntries(uintptrj_t *pcEntries, 
          {
          uintptrj_t thisPC = getSearchPCFromMethodAndBCIndex(method, bci);
 
-         TR_IPBytecodeHashTableEntry *entry = profilingSample(thisPC, 0, false);
+         TR_IPBytecodeHashTableEntry *entry = profilingSample(method, bci, comp);
          BCvisit->set(bci);
          if (entry && !invalidateEntryIfInconsistent(entry))
             {
@@ -516,7 +516,7 @@ TR_JITaaSClientIProfiler::serializeAndSendIProfileInfoForMethod(TR_OpaqueMethodB
       // numEntries will indicate how many entries have been populated
       // These profiling entries have been 'locked' so we must remember to unlock them
       bool abort = false;
-      bytesFootprint = walkILTreeForIProfilingEntries(pcEntries, numEntries, &bci, method, BCvisit, abort);
+      bytesFootprint = walkILTreeForIProfilingEntries(pcEntries, numEntries, &bci, method, BCvisit, abort, comp);
       if (numEntries && !abort)
          {
          // Serialize the entries
--- a/runtime/compiler/runtime/JITaaSIProfiler.hpp
+++ b/runtime/compiler/runtime/JITaaSIProfiler.hpp
@@ -95,7 +95,7 @@ class TR_JITaaSClientIProfiler : public TR_IProfiler
       // the base class. It may be better not to override any methods though
     
       uint32_t walkILTreeForIProfilingEntries(uintptrj_t *pcEntries, uint32_t &numEntries, TR_J9ByteCodeIterator *bcIterator,
-                                              TR_OpaqueMethodBlock *method, TR_BitVector *BCvisit, bool &abort);
+                                              TR_OpaqueMethodBlock *method, TR_BitVector *BCvisit, bool &abort, TR::Compilation *comp);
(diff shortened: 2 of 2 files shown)


harryyu1994:
There are 2 variations of `profilingSampe()`, one only grabs ipdata from hashtable and the other one grabs from both hashtable and sharedcache. I have yet to confirm this is the only cause of throughput regression as my performance experiments on vms have been extremely unstable. FYI @mpirvu

harryyu1994:
```
// We have two sources of data. Must pick the best one
         else
            {
            // Remember that we already looked into the SCC for this PC
            currentEntry->setPersistentEntryRead();
            int32_t currentCount = getSamplingCount(currentEntry, comp);
            int32_t persistentCount = getSamplingCount(persistentEntry, comp);
            if(currentCount >= persistentCount)
               {
               _STATS_IPEntryChoosePersistent++;
               currentEntry->copyFromEntry(persistentEntry, comp);
               return currentEntry;
               }
            else
               {
               _STATS_IPEntryChoosePersistent++;
               currentEntry->copyFromEntry(persistentEntry, comp);
               return currentEntry;
               }
            }
```
This is code from IProfiler.cpp's profilingSample() function, it's comparing HashTable & SharedCache IProfileData, both if & else branch are identical, I think for the if branch, we should just return currentEntry. Bug in OpenJ9.

harryyu1994:
We're getting multiple "CHTable WARNING: classInfo for subclass 0x2423500 of 0x189cd01 is null, ignoring" when with a sharedcache, and jitaas without sharedcache runs don't seem to have this warning. This could be another cause of the throughput regression. Will look into it.

harryyu1994:
JITaaS with SCC without my change: 3900/s
JITaaS with SCC with my change: 4200/s
While we definitely improved performance with this change, we are still far away from baseline.
JITaaS without SCC is around 4960/s and non-JITaaS without SCC is around 5170/s.
Going to look into the warnings I mentioned in the above comment.

fjeremic:
Minor suggestion to try to avoid using "(JITaaS)" prefixes. This is what the label is intended for.

mpirvu:
Related to "CHTable WARNING:..." this looks like a bug. We shouldn't ignore this. The code is adding a new list of subclasses for a particular j9class superclass, but one of the subclasses does not exist in its own right in the CHTable.
I suggest we trace the behavior and understand what is going on. I am guessing that the ordering of events my make do the wrong thing.

harryyu1994:
Should definitely open a separate issue for "CHTable WARNING ..." as it's independent of AOT SCC. It's not the cause of regression for this issue.

harryyu1994:
Performance actually improves from 4300/s to 4800/s with this change, while without SCC we get 4900/s. Non-jitaas is around 5100~5200.  (I've updated the comment from above). I'm still trying to chase down the gap between jitaas with SCC and jitaas without SCC.

mpirvu:
If you remove WIP I am willing to merge this PR and then we can investigate the remaining regression separately.
To summarize:
AOT+JITaaS = 4800
JITaaS = 4900
non-JITaaS = 5100-5200
AOT + non-JITaaS = ?

harryyu1994:
Removed WIP, I will try with AOT + non-JITaaS.

harryyu1994:
For today's run: 
non-JITaaS = 5350/s
AOT + non-JITaaS = 5230/s

I guess that means the gap between AOT+JITaaS and JITaaS might be reasonable. I'm just seeing a bunch of `validateCachedIPEntry` errors with AOT+JITaaS that I want to try to figure out why they exist.

harryyu1994:
When running AOT+JITaaS, we get a ton of 
```
Error for cached IP data: client sent us something but we have no cached entry. isMethodBeingCompiled=0 cnt=26873^M
       Method=0xc549c0 methodStart=0x7f64d6171594 bci=8 ID=1^M

Missmatch dominant class client=0xbe6e00 server=(nil)
```
and JITaaS without AOT doesn't have these messages.

mpirvu:
Those `Error for cached IP data:` messages come from a feature which is enabled only in debug mode. Are you running in debug mode?
The feature tries to validate the caching of IProfiler data at the server. To minimize number of messages, when the server asks for IP info for a bytecode, the client will send info for all bytecodes from that method and the server will cache that information. When the server wants IP info and finds that info in its cache, it will also send a message to the client to validate that whatever finds in the cache is the same as the info at the client.

harryyu1994:
The performance was run with release mode. Yes this check is performed in debug mode. I'm seeing a huge amount of messages that indicates we do have things cached in the server but it's empty. 
That's what this means right? Client side has the data and server doesn't. But if server doesn't have it, why would it be in the cache? I did see that there's a function that caches empty ip data entry so that we dont ask the client again. But I disabled that function and still seeing these errors. Trying to get to the bottom of this...
```
Error for cached IP data: client sent us something but we have no cached entry. isMethodBeingCompiled=0 cnt=26873^M
       Method=0xc549c0 methodStart=0x7f64d6171594 bci=8 ID=1^M
```

mpirvu:
>But if server doesn't have it, why would it be in the cache?

The client sends IProfiler info for an entire method, so while there is something cached at the server, not all bytecodes for the method may have data.

harryyu1994:
So could this be explained like this? 

1. Server asks Client for IProfiler data for some bytecodes
2. Client immediately sends IProfiler info for the entire method because the method is already AOT compiled(just guessing here). The data client sends likely doesn't contain iprofiler info for all the entries because they are mostly grabbed from the SCC
3. Server caches the "incomplete" IProfiler data (containg empty entires)
4. As the client has been running for a while, its IProfiler has gathered more data
5. And when the validation is performed, we can potentially have data at the client and empty cached entry at the server

We get about 713 "Error for cached IP data: client sent us something but we have no cached entry" for a 10 minute run.
And we get about 28 "Error for cached IP data: client sent us something but we have no cached entry"  for a 10 minute run without using SCC.

mpirvu:
Currently, the client sends IP information about an entire method only if the method has already been compiled in which case new IProfiler info cannot be collected.
There is one exception, the client also sends all IP info for methods that are being compiled, the idea being that there cannot be too much new IProfiler info. However, looking at your example `isMethodBeingCompiled=0` which means that this is not your case.

PR #3437, opened 2018-10-24
Adjust _safeReservePhysicalMemoryValue based on amount of free memory

mpirvu:
When allocating scratch memory the JIT will avoid allocating the last
TR::Options::_safeReservePhysicalMemoryValue of free physical memory. This
field is initialized to 50 MB, but it makes sense to initialize it to
smaller values if the JVM is running in containers that are very small.
In this commit TR::Options::_safeReservePhysicalMemoryValue is going
to be initialized to a value proportional to the free memory found at the
start of the JVM, but no more than 32 MB. The "free memory" readout already
takes into account the limits imposed on containers.


vijaysun-omr:
This change has the desired effect of having lesser "safe physical reserve memory" on smaller containers. On the other hand, the scaling that we do in this commit carries a greater risk that we will exceed the memory envelope at lower container sizes (and the container gets killed) but maybe this is okay since we are anyway risking instability by running inside very small containers to begin with. With a more reasonable sized container, the scaling leads to a fairly substantial safe physical reserve memory value anyway and it's in these cases that we want to avoid exceeding the memory envelope.

mpirvu:
Jenkins test sanity all JDK8

mpirvu:
@vijaysun-omr Sanity tests passed. The failure on MAC is not due to this PR.

Issue #3349, opened 2018-10-18
JITaaS: Aggressively cache information for j9classes at the server


mpirvu:
Currently, when the JITaaS server inquires about some field of a j9class the following pattern is used:
```
if j9class is cached
   return desired field
else
    send a message to client to get the value of the desired field
```
The proposal is to fetch all the info about a j9class (not just the desired field) when it is not cached. This transfers a lot more data, but the assumption is that future queries about j9class fields will be avoided, thus saving CPU. 
We need to measure whether this technique actually saves CPU and network traffic.

fjeremic:
Do we have data on which fields we usually request info about? Is it only specific classes (recognized fields for example or JCL classes) or is it arbitrary classes?

mpirvu:
The classes are arbitrary.
Examples of fields the optimizer wants to know can be found in https://github.com/eclipse/openj9/issues/3201

PR #3438, opened 2018-10-24
 Cache information for j9Classes at server

satbir-ca:
The implementation is to fetch all the information of the class
when it is not cached.

Fixes: #3349

### DIFF
--- a/runtime/compiler/control/JITaaSCompilationThread.cpp
+++ b/runtime/compiler/control/JITaaSCompilationThread.cpp
@@ -2725,30 +2725,47 @@ ClientSessionHT::printStats()
 void 
 JITaaSHelpers::cacheRemoteROMClass(ClientSessionData *clientSessionData, J9Class *clazz, J9ROMClass *romClass, ClassInfoTuple *classInfoTuple)
    {
+   ClientSessionData::ClassInfo classInfo;
    OMR::CriticalSection cacheRemoteROMClass(clientSessionData->getROMMapMonitor());
+   auto it = clientSessionData->getROMClassMap().find((J9Class*)clazz);
+   if (it == clientSessionData->getROMClassMap().end())
+      {
+      JITaaSHelpers::cacheRemoteROMClass(clientSessionData, clazz, romClass, classInfoTuple, classInfo);
+      }
+   }
+
+void
+JITaaSHelpers::cacheRemoteROMClass(ClientSessionData *clientSessionData, J9Class *clazz, J9ROMClass *romClass, ClassInfoTuple *classInfoTuple, ClientSessionData::ClassInfo &classInfoStruct)
+   {
    ClassInfoTuple &classInfo = *classInfoTuple;
-   J9Method *methods = std::get<1>(classInfo);
-   TR_OpaqueClassBlock *baseComponentClass = std::get<2>(classInfo);
-   int32_t numDims = std::get<3>(classInfo);
-   TR_OpaqueClassBlock *parentClass = std::get<4>(classInfo);
+
+   classInfoStruct.romClass = romClass;
+   classInfoStruct.methodsOfClass = std::get<1>(classInfo);
+   J9Method *methods = classInfoStruct.methodsOfClass;
+   classInfoStruct.baseComponentClass = std::get<2>(classInfo);
+   classInfoStruct.numDimensions = std::get<3>(classInfo);
+   classInfoStruct._remoteROMStringsCache = nullptr;
+   classInfoStruct._fieldOrStaticNameCache = nullptr;
+   classInfoStruct.parentClass = std::get<4>(classInfo);
    auto &tmpInterfaces = std::get<5>(classInfo);
-   auto interfaces = new (PERSISTENT_NEW) PersistentVector<TR_OpaqueClassBlock *>
+   classInfoStruct.interfaces = new (PERSISTENT_NEW) PersistentVector<TR_OpaqueClassBlock *>
       (tmpInterfaces.begin(), tmpInterfaces.end(),
        PersistentVector<TR_OpaqueClassBlock *>::allocator_type(TR::Compiler->persistentAllocator()));
    auto &methodTracingInfo = std::get<6>(classInfo);
-   bool classHasFinalFields = std::get<7>(classInfo);
-   uintptrj_t classDepthAndFlags = std::get<8>(classInfo);
-   bool classInitialized = std::get<9>(classInfo);
-   uint32_t byteOffsetToLockword = std::get<10>(classInfo);
-   TR_OpaqueClassBlock * leafComponentClass = std::get<11>(classInfo);
-   void *classLoader = std::get<12>(classInfo);
-   TR_OpaqueClassBlock * hostClass = std::get<13>(classInfo);
-   TR_OpaqueClassBlock * componentClass = std::get<14>(classInfo);
-   TR_OpaqueClassBlock * arrayClass = std::get<15>(classInfo);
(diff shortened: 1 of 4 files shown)


satbir-ca:
@mpirvu Please have a look at the initial implementation.
Still need to implement for the remaining class info fields.

runtime/compiler/control/JITaaSCompilationThread.cpp
```diff
@@ -2816,3 +2829,11 @@ JITaaSHelpers::getRemoteROMClass(J9Class *clazz, JITaaS::J9ServerStream *stream,
    *classInfoTuple = std::get<0>(recv);
    return romClassFromString(std::get<0>(*classInfoTuple), trMemory->trPersistentMemory());
    }
+
+void
+JITaaSHelpers::getRemoteRomClassReturnClassinfo(J9Class * clazz, JITaaS::J9ServerStream *stream, ClientSessionData::ClassInfo &classInfo)
```
mpirvu: Maybe a better name is `getAndCacheRAMClassInfo`

runtime/compiler/control/JITaaSCompilationThread.cpp
```diff
@@ -2816,3 +2829,11 @@ JITaaSHelpers::getRemoteROMClass(J9Class *clazz, JITaaS::J9ServerStream *stream,
...
+   {
+   JITaaSHelpers::ClassInfoTuple classInfoTuple;
+   auto romClass = JITaaSHelpers::getRemoteROMClass(clazz, stream, TR::comp()->trMemory(), &classInfoTuple);
```
mpirvu: I would like to minimize the use of thread locals if at all possible because they are expensive operations. trMemory and clientData could be available at the caller of this method.

runtime/compiler/control/JITaaSCompilationThread.cpp
```diff
@@ -2816,3 +2829,11 @@ JITaaSHelpers::getRemoteROMClass(J9Class *clazz, JITaaS::J9ServerStream *stream,
...
+   JITaaSHelpers::cacheRemoteROMClassReturnClassinfo(TR::compInfoPT->getClientData(), clazz, romClass, &classInfoTuple, classInfo);
```
mpirvu: Maybe we need a test on the romClass first.

mpirvu:
The ip validation check is failing because you need the signoff footer.

satbir-ca:
@mpirvu updated code pushed

runtime/compiler/env/VMJ9Server.cpp
```diff
@@ -1062,22 +1073,52 @@ TR_J9ServerVM::allocateCodeMemory(TR::Compilation * comp, uint32_t warmCodeSize,
...
+   void *class2Loader = nullptr;
+   do
       {
-      OMR::CriticalSection getRemoteROMClass(_compInfoPT->getClientData()->getROMMapMonitor());
-      auto it_class1 = _compInfoPT->getClientData()->getROMClassMap().find((J9Class*) class1);
-      if (it_class1 != _compInfoPT->getClientData()->getROMClassMap().end())
          {
-         auto it_class2 = _compInfoPT->getClientData()->getROMClassMap().find((J9Class*) class2);
-         if (it_class2 != _compInfoPT->getClientData()->getROMClassMap().end())
+         OMR::CriticalSection getRemoteROMClass(_compInfoPT->getClientData()->getROMMapMonitor());
```
mpirvu: I suggest using a helper that finds the classloader of a j9method. This could be called twice here and once in `isClassLoadedBySystemClassLoader`

satbir-ca:
@mpirvu  updated code pushed

fjeremic:
The commit within this PR needs to conform to the commit guidelines:

>The first line describes the change made. It is written in the imperative mood, and should say what happens when the patch is applied. Keep it short and simple. The first line should be less than 70 characters, where reasonable, and should be written in sentence case *preferably not ending in a period*. Leave a blank line between the first line and the message body.
>
> *The body should be wrapped at 72 characters*, where reasonable.

[1] https://github.com/eclipse/openj9/blob/master/CONTRIBUTING.md#commit-guidelines

runtime/compiler/control/JITaaSCompilationThread.cpp
```diff
@@ -2816,3 +2833,162 @@ JITaaSHelpers::getRemoteROMClass(J9Class *clazz, JITaaS::J9ServerStream *stream,
    *classInfoTuple = std::get<0>(recv);
    return romClassFromString(std::get<0>(*classInfoTuple), trMemory->trPersistentMemory());
    }
+
+bool
+JITaaSHelpers::getAndCacheRAMClassInfo(J9Class *clazz, ClientSessionData *clientSessionData, JITaaS::J9ServerStream *stream, ClassInfoDataType dataType, void *data)
+   {
+   JITaaSHelpers::ClassInfoTuple classInfoTuple;
+   ClientSessionData::ClassInfo classInfo;
+   if (!clazz)
```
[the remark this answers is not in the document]
satbir-ca: I have see calls being made with the class = NULL that was the reason why I was seeing crash. if you check the function TR_J9VMBase::isClassInitialized you will find the same check for class == NULL.

The circumstance under which this will happen I don't know. If you want I can debug further but since the code is already there is good chances of being some valid reason.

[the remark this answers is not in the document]
mpirvu: If only `TR_J9VMBase::isClassInitialized` has this behavior, then we could place the check in that function alone.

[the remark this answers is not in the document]
satbir-ca: also TR_J9VMBase::getByteOffsetToLockword.
If we are targeting function by function then if in future the JIT changes its usage like any other function with NULL class pointer, then JITaaS will crash for that unless we make the appropriate change.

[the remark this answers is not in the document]
mpirvu: ok. If it's more than one place then the code should stay the way it is.

runtime/compiler/control/JITaaSCompilationThread.cpp
```diff
@@ -2816,3 +2833,165 @@ JITaaSHelpers::getRemoteROMClass(J9Class *clazz, JITaaS::J9ServerStream *stream,
...
+         *(TR_OpaqueClassBlock **)data = classInfo.arrayClass;
+         }
+         break;
+      case CLASSINFO_TOTAL_INSTANCE_SIZE :
+         {
+         *(uintptrj_t *)data = classInfo.totalInstanceSize;
+         }
+         break;
+      default :
+         break;
```
mpirvu: I suggest an ASSERT for the default case. This will catch errors when we add new fields.

mpirvu:
@satbir-ca  Please drop the "WIP" prefix when ready and develop more the description of this commit. You could copy directly from the connected issue.

mpirvu:
@satbir-ca   Applying this commit results in
```
JITaaS server ready to accept incoming requests
*** stack smashing detected ***: /home/mpirvu/JITaaS/sdk/jre/bin/java terminated
JVMDUMP039I Processing dump event "abort", detail "" at 2018/11/08 18:15:32 - please wait.
```
even when running "java -version" with a prod build. It does not happen with a debug build.

mpirvu:
The code writes 64 bits
 `*(uintptrj_t *)data = classInfo.totalInstanceSize;`

into a local variable that is 32 bits
`uint32_t totalInstanceSize = 0`

mpirvu:
Bug fixed in #3615

PR #3439, opened 2018-10-24
Fix monitor object access sites

amicic:
Use proper API to access (read) monitor's object, which would trigger GC
object access (read) barrier, if needed.


amicic:

Issue #3369, opened 2018-10-19
FindDeadlockTest test hangs when native library not present (jdk11 OSX)


smlambert:
Will need to provide the test natives for functional tests.  Can not get through the functional test suite due to this hang:
https://ci.adoptopenjdk.net/view/Test_weekly/job/openjdk11_j9_functionaltest_x86-64_macos/5/console

04:46:53 +++ j9vm.test.thread.FindDeadlockTest: +++
04:46:53 command: /Users/jenkins/workspace/openjdk11_j9_functionaltest_x86-64_macos/openjdkbinary/j2sdk-image/bin/java -Xcompressedrefs -XX:+UseCompressedOops -Xjit -Xgcpolicy:balanced  -Xdump -Xms64m -Xmx64m    -classpath /Users/jenkins/workspace/openjdk11_j9_functionaltest_x86-64_macos/openjdk-tests/TestConfig/scripts/testKitGen/../../../../jvmtest/functional/VM_Test/VM_Test.jar:/Users/jenkins/workspace/openjdk11_j9_functionaltest_x86-64_macos/openjdk-tests/TestConfig/scripts/testKitGen/../../../../jvmtest/TestConfig/lib/asm-all.jar  j9vm.test.thread.FindDeadlockTest 
04:46:53 
04:46:54 Locker got java.lang.Object@c4459101
04:46:54 Blocker Thread[b3,5,main] locking java.lang.Object@c4459101
04:46:54 Blocker Thread[b1,5,main] locking java.lang.Object@c4459101
04:46:54 Blocker Thread[b2,5,main] locking java.lang.Object@c4459101
04:46:56 Blocker Thread[c1,5,main] locking Thread[b1,5,main]
04:46:57 Blocker Thread[c2,5,main] locking Thread[c1,5,main]
04:46:58 Blocker Thread[c3,5,main] locking Thread[c2,5,main]
04:47:00 Testing no deadlocks...No natives for j9vm.test.thread tests
04:47:00 Exception in thread "main" java.lang.UnsatisfiedLinkError: j9vm/test/thread/NativeHelpers.findDeadlockedThreads()[Ljava/lang/Thread;
04:47:00 	at j9vm.test.thread.FindDeadlockTest.checkDeadlocks(FindDeadlockTest.java:177)
04:47:00 	at j9vm.test.thread.FindDeadlockTest.testNoDeadlocks(FindDeadlockTest.java:301)
[...]

PR #3443, opened 2018-10-25
Force VM exit if exception thrown while threads deadlocked

pdbain-ibm:
Fixes https://github.com/eclipse/openj9/issues/3369


pshipton:
Not sure we should close #3369. Is there another issue to handle the native library problem?

pshipton:

pdbain-ibm:
>  Is there another issue to handle the native library problem?

Yes: https://github.com/eclipse/openj9/issues/3292

smlambert:
thanks!

PR #3448, opened 2018-10-25
Limit heap size for jvmtitests

pdbain-ibm:
Some jvmti tests are exceedingly slow if run on machines with large amounts of
RAM.


pdbain-ibm:
Tested and working 10/10 on the problematic machine.

pshipton:

Issue #2024, opened 2018-05-30
The wrong logic for the disclaim work in the shared classes


ChengJin01:
The current implementation of disclaiming the meta data in the shared cache is incorrect in that it only disclaims the area section(min/max value) set up by the "used" meta data (only when the meta data entry is retrieved) rather than the stored meta data. 

The problem was detected in cold/warm runs of LibertyStartup in which case only a few stored meta data entries got retrieved in the cold runs while a huge amount of stored meta data (e.g. for AOT methods) got retrieved in subsequent warm runs before exiting the startup stage.  So the VM ended up with much higher footprint in cold runs than in the warm runs.

The intent of the disclaim work is to release all stored meta data after exiting the startup stage, whether it is in the cold run or in the warm run (we assume the meta data is used as long as it is stored there). 
Note: the disclaimed memory can be loaded back after that if needed.

To achieve this, the min/max value of the disclaimed meta data section (initially updated by the stored meta data) should be modified to ensure these values are shared in both cold runs and warm runs.

From the footprint perspective, the score of cold runs should be pretty much equal to that of warm runs after fixing this issue.

ChengJin01:
FYI: @pshipton , @DanHeidinga , @hangshao0, @pdbain-ibm

ChengJin01:
@pshipton,    I talked to Hang. Basically there are two solutions to disclaim the meta data as follows:

1) call updateAccessedShrCacheMetadataBounds() whenever storing the meta data (AOT method, attachData, ROM class, etc) , in which case it triggers to disclaim the meta data in cold run ( the existing code already called updateAccessedShrCacheMetadataBounds when finding the meta data in warm runs). 
2) no need to call updateAccessedShrCacheMetadataBounds() at all and directly disclaim all stored meta data in both cold and warm runs as the left boundary/min value (UPDATEPTR(_theca)) and  the right boundary/max value (CADEBUGSTART(_theca)) of the stored meta data area are already stored in the header of the shared cache.

There is no fundamental difference for both solutions.  I am wondering which one you prefer to fix the issue. Correct me if anything I overlooked.

vijaysun-omr:
FYI @mpirvu

pshipton:
I think option 2 would work, but only for the cold run. The logic for the warm run should not be changed. We shouldn't disclaim from UPDATEPTR(_theca) in a warm run because this would AOT which hasn't yet been used.

pshipton:
[...]

PR #3454, opened 2018-10-25
Disclaim the stored meta data in the shared cache

ChengJin01:
The change is to ensure all stored meta data
are disclaimed when exiting the startup stage
in the cold run.

Fix: #2024

### DIFF
--- a/runtime/shared_common/CacheMap.cpp
+++ b/runtime/shared_common/CacheMap.cpp
@@ -1662,6 +1662,20 @@ SH_CacheMap::allocateROMClass(J9VMThread* currentThread, const J9RomClassRequire
 	
 	pieces->romClass = (void *) allocateROMClassOnly(currentThread, romclassSizeToUse, classnameLength, classnameData, cpw, partitionInCache, modContextInCache, callerHelperID, modifiedNoContext, newItemInCache, cacheAreaForAllocate);
 
+	if ((NULL != newItemInCache) && _ccHead->isNewCache()) {
+		/* Update the min/max boundary with the stored metadata entry only when the cache is
+		 * being created by the current VM.
+		 */
+#if !defined(J9ZOS390) && !defined(AIXPPC)
+#if defined(LINUX)
+		if (J9_ARE_ALL_BITS_SET(*_runtimeFlags, J9SHR_RUNTIMEFLAG_ENABLE_PERSISTENT_CACHE))
+#endif
+		{
+			updateAccessedShrCacheMetadataBounds(currentThread, (uintptr_t *) ITEMDATA(newItemInCache));
+		}
+#endif /* !defined(J9ZOS390) && !defined(AIXPPC) */
+	}
+
 	if ((true == allocatedDebugMem) && (NULL == pieces->romClass)) {
 		Trc_SHR_CM_allocateROMClass_FailedToRomClassRollbackDebug_Event(currentThread, classnameLength, classnameData, sizes->lineNumberTableSize, pieces->lineNumberTable, sizes->localVariableTableSize, pieces->localVariableTable);
 		this->rollbackClassDebugData(currentThread, classnameLength, classnameData);
@@ -2306,6 +2320,20 @@ SH_CacheMap::commitMetaDataROMClassIfRequired(J9VMThread* currentThread, Classpa
 		goto done;
 	}
 
+	/* Update the min/max boundary with the stored metadata entry only when the cache is
+	 * being created by the current VM.
+	 */
+	if (_ccHead->isNewCache()) {
+#if !defined(J9ZOS390) && !defined(AIXPPC)
+#if defined(LINUX)
+		if (J9_ARE_ALL_BITS_SET(*_runtimeFlags, J9SHR_RUNTIMEFLAG_ENABLE_PERSISTENT_CACHE))
+#endif
+		{
+			updateAccessedShrCacheMetadataBounds(currentThread, (uintptr_t *) ITEMDATA(itemInCache));
+		}
+#endif /* !defined(J9ZOS390) && !defined(AIXPPC) */
+	}
+
 	retval = commitROMClass(currentThread, itemInCache, cacheAreaForAllocate, cpw, cpeIndex, partitionInCache, modContextInCache, romClassBuffer, false);
 	goto done_skipHashUpdate;
 
@@ -2789,6 +2817,24 @@ SH_CacheMap::storeROMClassResource(J9VMThread* currentThread, const void* romAdd
 		result = resourceDescriptor->unWrap(resourceWrapper);
 	}
 
+	/* Update the min/max boundary with the stored metadata entry only when the cache is
+	 * being created by the current VM.
+	 */
+	if ((NULL != result)
+	&& ((void*)J9SHR_RESOURCE_STORE_FULL != result)
+	&& ((void*)J9SHR_RESOURCE_STORE_ERROR != result)
+	&& _ccHead->isNewCache()
+	) {
+#if !defined(J9ZOS390) && !defined(AIXPPC)
+#if defined(LINUX)
+		if (J9_ARE_ALL_BITS_SET(*_runtimeFlags, J9SHR_RUNTIMEFLAG_ENABLE_PERSISTENT_CACHE))
+#endif
+			{
+				updateAccessedShrCacheMetadataBounds(currentThread, (uintptr_t *) result);
+			}
+#endif /* !defined(J9ZOS390) && !defined(AIXPPC) */
+	}
+
 	_ccHead->exitWriteMutex(currentThread, fnName);
 	Trc_SHR_CM_storeROMClassResource_Exit4(currentThread, result);
 	return result;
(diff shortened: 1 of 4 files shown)


ChengJin01:
The fix is based on the Option 1 already comfirmed at https://github.com/eclipse/openj9/issues/2024
The expected result is: the disclaimed memory in cold run should be very close to that in warm run

With the fix, the test on LibertyStartup shows it meets our expectation:

1) Without the fix:
```
trace.jdk11_openj9_nofix.12987_id.out.fmt (cold run):
20:34:24.949599000  0x00000000025f3e00 j9shr.2183          Event       SH_CacheMap::updateAccessedShrCacheMetadataBounds: new minimum 0x7fe80fb39928
20:34:24.949599000  0x00000000025f3e00 j9shr.2184          Event       SH_CacheMap::updateAccessedShrCacheMetadataBounds: new maximum 0x7fe80fb39928
...
20:34:27.946487000  0x0000000002aa8600 j9shr.2183          Event       SH_CacheMap::updateAccessedShrCacheMetadataBounds: new minimum 0x7fe80f9bb6b0

MAX - MIN = 0x7fe80fb39928 - 0x7fe80f9bb6b0 = 17E278 = 1,565,304 = 1.565MB

trace.jdk11_openj9_nofix.12987_id.out.fmt (warm run):
20:34:55.142439000  0x000000000250ae00 j9shr.2183          Event       SH_CacheMap::updateAccessedShrCacheMetadataBounds: new minimum 0x7f5c7fb3aef0
20:34:55.142439000  0x000000000250ae00 j9shr.2184          Event       SH_CacheMap::updateAccessedShrCacheMetadataBounds: new maximum 0x7f5c7fb3aef0
...
20:34:55.971428000  0x000000000250ea00 j9shr.2183          Event       SH_CacheMap::updateAccessedShrCacheMetadataBounds: new minimum 0x7f5c7f765734

MAX - MIN = 0x7f5c7fb3aef0 - 0x7f5c7f765734 = 3D57BC = 4,020,156 = 4.02MB
```

2) With the fix:
```
trace.jdk11_openj9_fix.14307_id.out.fmt (cold run):
21:07:09.688931000  0x0000000000cbee00 j9shr.2183          Event       SH_CacheMap::updateAccessedShrCacheMetadataBounds: new minimum 0x7fe57fb3aef0
21:07:09.688931000  0x0000000000cbee00 j9shr.2184          Event       SH_CacheMap::updateAccessedShrCacheMetadataBounds: new maximum 0x7fe57fb3aef0
...
21:07:27.989654000  0x0000000000cc2a00 j9shr.2183          Event       SH_CacheMap::updateAccessedShrCacheMetadataBounds: new minimum 0x7fe57f773dd0

MAX - MIN = 0x7fe57fb3aef0 - 0x7fe57f773dd0 = 3C7120 = 3,961,120 = 4MB

trace.jdk11_openj9_fix.14624_id.out.fmt (warm run):
21:07:40.388673000  0x0000000002308e00 j9shr.2183          Event       SH_CacheMap::updateAccessedShrCacheMetadataBounds: new minimum 0x7fe7d3b3aef0
21:07:40.388674000  0x0000000002308e00 j9shr.2184          Event       SH_CacheMap::updateAccessedShrCacheMetadataBounds: new maximum 0x7fe7d3b3aef0
...
21:07:55.044459000  0x000000000230ca00 j9shr.2183          Event       SH_CacheMap::updateAccessedShrCacheMetadataBounds: new minimum 0x7fe7d372e8a0

MAX - MIN = 0x7fe7d3b3aef0 - 0x7fe7d372e8a0 = 40C650 = 4,245,072 = 4.245MB
```

ChengJin01:
Reviewer: @pshipton 

FYI: @DanHeidinga

hangshao0:
> The change is to ensure all stored meta data
> are disclaimed when exiting the startup stage
> in the cold run.

I don't see the check for cold run. I guess this change affects both warm and cold runs.  

Can this be done in commitUpdateHelper() ?

ChengJin01:
> I don't see the check for cold run. I guess this change affects both warm and cold runs

The changes are based on the fact that the stored metadata are seldomly used in cold runs before existing the startup stage.

> Can this be done in commitUpdateHelper() ?

That looks weird as we already called updateAccessedShrCacheMetadataBounds() in find operation (e.g. in findCompiledMethod and findROMClass.

hangshao0:
Not sure if we want to change the behavior of warm runs. 

If we are only recording the boundary of storing operation of romclass metadata, AOT and JIT data on the cold run, then commitUpdateHelper might not be the right place. 

For romclass metadata, there is another place it is stored:
https://github.com/eclipse/openj9/blob/master/runtime/shared_common/CacheMap.cpp#L2298to#L2300

ChengJin01:
> Not sure if we want to change the behavior of warm runs.

As mentioned above, the store operations mostly happen in cold run (I believe most of them are JIT related). So the assumption is that it is used when stored, which is why we can safely disclaim them. 
It doesn't matter whether it happens in colds run or warm run.

runtime/shared_common/CacheMap.cpp
```diff
@@ -2789,6 +2801,16 @@ SH_CacheMap::storeROMClassResource(J9VMThread* currentThread, const void* romAdd
 		result = resourceDescriptor->unWrap(resourceWrapper);
 	}
 
+	/* Update the min/max boundary with the stored metadata entry if allocated successfully */
+#if !defined(J9ZOS390) && !defined(AIXPPC)
+#if defined(LINUX)
+		if (J9_ARE_ALL_BITS_SET(*_runtimeFlags, J9SHR_RUNTIMEFLAG_ENABLE_PERSISTENT_CACHE))
```
pshipton: Indenting seems wrong here.

runtime/shared_common/CacheMap.cpp
```diff
@@ -2789,6 +2801,16 @@ SH_CacheMap::storeROMClassResource(J9VMThread* currentThread, const void* romAdd
...
+#endif
+		{
+			updateAccessedShrCacheMetadataBounds(currentThread, (uintptr_t *) result);
```
pshipton: This place and findCompiledMethod() call updateAccessedShrCacheMetadataBounds without checking that result is valid first.

pshipton:
> It doesn't matter whether it happens in colds run or warm run.

Actually it does. The idea is to madvise the memory that was already used during startup. Updating the metadata bounds for a store in the warm run will end up madvising much more than desired, and this could have undesired side affects since the VM will need to read some of that memory as it continues to run.

ChengJin01:
Already discussed with Hang.  It seems we can take advantage of isCacheInitComplete() to determine whether the current cache is a new one or not (which we assume to be in a cold run)
```
/runtime/shared_common/CompositeCache.cpp
bool
SH_CompositeCacheImpl::isCacheInitComplete(void)
{
	return ((_theca != NULL) && (0 != (_theca->ccInitComplete & CC_STARTUP_COMPLETE)));
}
```
So we update the boundary of metadata in store only when isCacheInitComplete() returns false, which means it is still in the new state.

hangshao0:
> So we can update the boundary of metadata in store only when isCacheInitComplete() returns false

Not exactly. isCacheInitComplete() always returns true after the shared cache is started up. You might want to cache this variable and check this variable:
https://github.com/eclipse/openj9/blob/master/runtime/shared_common/CompositeCache.cpp#L1358

ChengJin01:
> Not exactly. isCacheInitComplete() always ...

Right.  The idea is updating the boundaries only when the cache is created by the current VM this time; otherwise, the updating should be rejected (including the cases: 1) not created by this VM in cold run 2) already created by a VM in cold run ).

runtime/shared_common/CacheMap.cpp
ChengJin01: Already fixed.

runtime/shared_common/CacheMap.cpp
ChengJin01: Already added checking code.

ChengJin01:
I modified the changes based the logic above and need to verify it works as expected.

ChengJin01:
I already checked the result on LibertyStartup with the latest changes as follows:
```
trace.jdk11_openj9_fix.32051_id.out.fmt (cold run):

19:30:41.814752000  0x0000000002165e00 j9shr.2183          Event       SH_CacheMap::updateAccessedShrCacheMetadataBounds: new minimum 0x7f5aebb3aef0
19:30:41.814753000  0x0000000002165e00 j9shr.2184          Event       SH_CacheMap::updateAccessedShrCacheMetadataBounds: new maximum 0x7f5aebb3aef0
...
19:31:00.140240000  0x0000000002169a00 j9shr.2183          Event       SH_CacheMap::updateAccessedShrCacheMetadataBounds: new minimum 0x7f5aeb76bc0c

MAX - MIN = 0x7f5aebb3aef0 - 0x7f5aeb76bc0c = 3CF2E4 = 3,994,340 = 3.994MB

trace.jdk11_openj9_fix.32372_id.out.fmt (warm run):

19:31:12.360423000  0x00000000012e6e00 j9shr.2183          Event       SH_CacheMap::updateAccessedShrCacheMetadataBounds: new minimum 0x7f1107b3aef0
19:31:12.360424000  0x00000000012e6e00 j9shr.2184          Event       SH_CacheMap::updateAccessedShrCacheMetadataBounds: new maximum 0x7f1107b3aef0
...
19:31:13.984535000  0x00000000012eaa00 j9shr.2183          Event       SH_CacheMap::updateAccessedShrCacheMetadataBounds: new minimum 0x7f110776c130

MAX - MIN = 0x7f1107b3aef0 - 0x7f110776c130 = 3CEDC0 = 3,993,024 = 3.993MB
```
As compared to the previous result (4.245MB) with the old changes,  the result in the warm run was literally reduced to 3.993MB, which is pretty close to the value in cold run.  So it means the new changes successfully prevented updating the boundaries of meta data for a store in the warm run.

runtime/shared_common/CompositeCache.cpp
```diff
@@ -1325,7 +1326,7 @@ SH_CompositeCacheImpl::startup(J9VMThread* currentThread, J9SharedClassPreinitCo
 				}
 			}
 			if (!isCacheCorrupt()) {
-				bool initializingNewCache = false;
+				_initializingNewCache = false;
```
hangshao0: It is already initialized to false. No need to set to false here.

runtime/shared_common/CacheMap.cpp
```diff
@@ -3016,43 +3062,46 @@ SH_CacheMap::findCompiledMethod(J9VMThread* currentThread, const J9ROMMethod* ro
...
+				}
 			}
-		}
-	} else
+		} else
 #if defined(LINUX)
 		if (J9_ARE_ALL_BITS_SET(*_runtimeFlags, J9SHR_RUNTIMEFLAG_ENABLE_PERSISTENT_CACHE))
 #endif
 		{
 			updateAccessedShrCacheMetadataBounds(currentThread, (uintptr_t *) result);
```
pshipton: Please fix the bug and check result is not NULL before calling updateAccessedShrCacheMetadataBounds().

runtime/shared_common/CompositeCache.cpp
```diff
@@ -6861,3 +6862,14 @@ SH_CompositeCacheImpl::increaseUnstoredBytes(U_32 blockBytes, U_32 aotBytes, U_3
...
+
+/* Query whether the cache is being created by the current VM,
+ * which we assume to be in cold run.
+ *
+ * @return true if cache is being created, false otherwise
+ */
+bool
+SH_CompositeCacheImpl::isNewCache(void)
+{
+	return _initializingNewCache;
```
pshipton: This should have a check
```
	if (!_started) {
		Trc_SHR_Assert_ShouldNeverHappen();
		return false;
	}
```

runtime/shared_common/CompositeCache.cpp
```diff
@@ -1355,7 +1356,7 @@ SH_CompositeCacheImpl::startup(J9VMThread* currentThread, J9SharedClassPreinitCo
 				 *  - store information regarding -Xnolinenumbers
 				 */
 				if (isCacheInitComplete() == false) {
-					initializingNewCache = true;
+					_initializingNewCache = true;
```
hangshao0: You can consider adding a trace point here.

runtime/shared_common/CacheMap.cpp
pshipton: Sorry, missed the null check added above.

runtime/shared_common/CompositeCache.cpp
ChengJin01: Already removed.

runtime/shared_common/CompositeCache.cpp
ChengJin01: Added as suggested above.

runtime/shared_common/CompositeCache.cpp
ChengJin01: Already added an event.

pshipton:
jenkins test all xlinux jdk8

PR #3456, opened 2018-10-26
Fix profilingSample SharedClassCache bug

harryyu1994:
When we have profiling data in both HashTable and SCC, we pick
the one with higher sampling count. Should return HashTable's entry
when its sampling count is higher. Currently we are returning SCC's entry
for both cases.

### DIFF
--- a/runtime/compiler/runtime/IProfiler.cpp
+++ b/runtime/compiler/runtime/IProfiler.cpp
@@ -1576,8 +1576,6 @@ TR_IProfiler::profilingSample (TR_OpaqueMethodBlock *method, uint32_t byteCodeIn
             int32_t persistentCount = getSamplingCount(persistentEntry, comp);
             if(currentCount >= persistentCount)
                {
-               _STATS_IPEntryChoosePersistent++;
-               currentEntry->copyFromEntry(persistentEntry, comp);
                return currentEntry;
                }
             else


harryyu1994:
@mpirvu Hi Marius, could you help reviewing this?

mpirvu:
@cathyzhyi could you please review this PR since you are familiar with this code?

cathyzhyi:
the old code is wrong and the change looks correct.

harryyu1994:
Okay will do

mpirvu:
How about performance for DT7 (throughput and startup time)?

harryyu1994:
```
config: iprofiling
mean:  14916.8
std:  23.7    0.00158881 %
raw throughput
14893.
14940.
config: original
mean:  14636.1
std:  53.75    0.00367241 %
raw throughput
14689.
14582.
```
Seeing an improvement for DayTrader. I will have more runs once the machines are available again.

mpirvu:
Improvement is good :-)
This is not necessarily unexpected because the original idea was good.

mpirvu:
Jenkins test sanity.functional all jdk8

mpirvu:
Jenkins test sanity all jdk8

mpirvu:
@harryyu1994 Do you plan to run more performance tests? If not, I could merge this item.

harryyu1994:
@mpirvu I'm planning on run more performance runs tonight.

harryyu1994:
A question, should we also update the SCC entry to be the HashTable entry if the HashTable entry was chosen?

mpirvu:
> should we also update the SCC entry to be the HashTable entry if the HashTable entry was chosen?

If I understand correctly you want to write to the SCC an updated version of the entry. As far I know SCC is mostly "write once", so it cannot be done.

harryyu1994:
```
config: iprofiling
mean:  13733.1
std:  69.7809    0.00508124 %
raw throughput
13759.
13806.
13583.
13757.
13747.
13744.
config: original
mean:  13937.6
std:  83.8865    0.0060187 %
raw throughput
13910.
13784.
13941.
14054.
14004.
13930.
```

The previous runs I forgot to use the shareclasses options. This one is will shareclasses and it looks like we regressed with this change.

mpirvu:
Could you please try one more time? I wonder how much of this is fluctuation given that previous results showed an improvement while the performance should have been neutral because SCC was not used.

BTW, liberty sets sharedclasses options by default.

harryyu1994:
Using default liberty settings we get:
```
config: iprofiling
raw throughput
14926.4
14789.8
14829.4
14740.1
14852.2
config: original
raw throughput
14661.8
14687.1
14746.9
14726.2
14547.5
```

These are also valid results with liberty shareclasses default settings:
```
config: iprofiling
raw throughput
14893.
14940.
config: original
raw throughput
14689.
14582.
```

I think the reason we got worse results for the change was due to my options. 
I had "-Xshareclasses:name=original,cacheDir=/home/harryayu/shareclasses,enableBCI -Xscmx200M -Xscmaxaot100M".

@mpirvu I think it's safe to merge this one now.

mpirvu:

mpirvu:
CI errors are not related to this PR, so merging.

Issue #3400, opened 2018-10-22
JVMTI GetOwnedMonitorStackDepthInfo does not halt target thread


gacholio:
GetOwnedMonitorStackDepthInfo walks the stack of the target thread without suspending execution of the thread, resulting in crashes or inconsistent data.

PR #3430, opened 2018-10-24
Halt thread before getting owned monitor data

gacholio:
- halt target thread in GetOwnedMonitorStackDepthInfo
- add test for targetting non-current thread
- remove forced -Xint from monitor tests

Note the test validates nothing - it just calls the API repeatedly on a
foreign thread. This crashes on unfixed VMs.

Fixes: #3400

[ci skip]

Signed-off-by: Graham Chapman <graham_chapman@ca.ibm.com>

PR #3458, opened 2018-10-26
Halt thread before getting owned monitor data

gacholio:
- halt target thread in GetOwnedMonitorStackDepthInfo
- add test for targetting non-current thread

Note the test validates nothing - it just calls the API repeatedly on a
foreign thread. This crashes on unfixed VMs.

Fixes: #3400

[ci skip]


gacholio:
@pshipton This is #3430 without removing the -Xint (and adding it for the new test).

Issue #3148, opened 2018-10-04
Use NASM Assembler across all x86 operating systems


nbhuiyan:
During the effort to port the OpenJ9 JIT on macOS, it became necessary to re-write the x86 assembly files in NASM format. Discussions around this topic occurred in #2418 

One of the main advantages of using NASM is that it is supported on all x86 platforms, meaning a unified assembler solution can be used throughout x86, consuming the same source files without requiring intermediate conversions. While the current assembly files are set up to work on macOS (see #2864), they are not set up to be used on xLinux and Windows yet. This issue will be used for tracking the work around enabling use of NASM across all x86 operating systems.

### The transition period

To enable macOS support, the NASM assembly files currently coexist with the MASM files that are used on Windows and Linux (after being processed by `masm2gas.pl`). This dual implementation will have to exist until support is added for Linux and Windows, which is expected to be completed within October 2018. In the meantime, any changes to the MASM files will have to be synced with the NASM files, and I will ensure that they remain synced. Here are the list of files that will need to be synced with their NASM versions, and will be removed once the migration is complete:

In [`runtime/compiler/x/runtime`](https://github.com/eclipse/openj9/tree/master/runtime/compiler/x/runtime)
* X86ArrayTranslate.asm
* X86Codert.asm
* X86EncodeUTF16.asm
* X86LockReservation.asm
* X86PicBuilder.inc
* X86PicBuilder.pasm
* X86RegisterMap.inc
* X86Unresolveds.pasm

In [`runtime/compiler/x/amd64/runtime`](https://github.com/eclipse/openj9/tree/master/runtime/compiler/x/amd64/runtime)
* AMD64CompressString.asm
* AMD64CompressString.inc
* AMD64Recompilation.asm

In [`runtime/compiler/x/i386/runtime`](https://github.com/eclipse/openj9/tree/master/runtime/compiler/x/i386/runtime)
* IA32CompressString.inc
* IA32Math64.inc
* IA32Recompilation.asm
* J9IA32CompressString.asm
* J9IA32Math64.asm

After the migration is complete, all new changes should be made in the NASM files. Syntactically, MASM and NASM are quite similar. There is an excellent [quick-start guide for MASM developers to migrate to NASM](https://www.nasm.us/doc/nasmdoc2.html#section-2.2) that highlights some of the major differences.

### List of tasks to complete migration

- [x] OSX Support (#2864)
- [x] Linux on x86-64 support (#3293)
[...]

PR #3459, opened 2018-10-26
Use NASM assembler on x86-64 Windows and update NASM set-up process on OSX and xLinux makefiles

nbhuiyan:
Depends on: #3293 , #3461 , #3493 

This PR builds on top of the changeset in #3293 , and contains commits that are specific to x86_64 Linux, and will be rebased once #3293 is merged.

Issue: #3148 

Signed-off-by: Nazim Uddin Bhuiyan <nazim.uddin.bhuiyan@ibm.com>

### DIFF
--- a/doc/build-instructions/Build_Instructions_V11.md
+++ b/doc/build-instructions/Build_Instructions_V11.md
@@ -1,5 +1,5 @@
 <!--
-Copyright (c) 2018, 2018 IBM Corp. and others
+Copyright (c) 2018, 2019 IBM Corp. and others
 
 This program and the accompanying materials are made available under
 the terms of the Eclipse Public License 2.0 which accompanies this
@@ -366,10 +366,16 @@ You must install a number of software dependencies to create a suitable build en
 - [Microsoft Visual Studio 2017]( https://visualstudio.microsoft.com/thank-you-downloading-visual-studio/?sku=Community&rel=15), which is the default compiler level used by OpenJDK11; or [Microsoft Visual Studio 2013]( https://go.microsoft.com/fwlink/?LinkId=532495), which is chosen to compile if VS2017 isn't installed.
 - [Freemarker V2.3.8](https://sourceforge.net/projects/freemarker/files/freemarker/2.3.8/freemarker-2.3.8.tar.gz/download)
 - [LLVM/Clang](http://releases.llvm.org/7.0.0/LLVM-7.0.0-win64.exe)
+- [NASM Assembler v2.13.03 or newer](https://www.nasm.us/pub/nasm/releasebuilds/?C=M;O=D)
 
 Add the binary path of Clang to the `PATH` environment variable to override the older version of clang integrated in Cygwin. e.g.
 ```
 export PATH="/cygdrive/c/LLVM/bin:$PATH" (in Cygwin)
+```
+
+Add the path to `nasm.exe` to the `PATH` environment variable to override the older version of NASM installed in Cygwin. e.g.
+```
+export PATH="/cygdrive/c/NASM:$PATH" (in Cygwin)
 ```
 
 Update your `LIB` and `INCLUDE` environment variables to provide a path to the Windows debugging tools with the following commands:
--- a/runtime/compiler/build/rules/win-msvc/filetypes.mk
+++ b/runtime/compiler/build/rules/win-msvc/filetypes.mk
@@ -1,4 +1,4 @@
-# Copyright (c) 2000, 2017 IBM Corp. and others
+# Copyright (c) 2000, 2019 IBM Corp. and others
 #
 # This program and the accompanying materials are made available under
 # the terms of the Eclipse Public License 2.0 which accompanies this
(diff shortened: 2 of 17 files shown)


nbhuiyan:
@0xdaryl Ready for review

0xdaryl:
Jenkins test sanity all jdk11

0xdaryl:
Jenkins test sanity all jdk11

runtime/compiler/runtime/asmprotos.h
```diff
@@ -64,7 +64,7 @@ extern "C" {
  * is set for OS variants that currently use NASM and is used to
  * decide the version of helper names to use.
  */
-#if defined(OSX) || (defined(LINUX) && defined(TR_HOST_X86) && defined(TR_HOST_64BIT))
+#if (defined(TR_HOST_X86) && defined(TR_HOST_64BIT)) && (defined(OSX) || defined(LINUX) || defined(WINDOWS))
```
0dvictor: Should we remove `defined(OSX) || defined(LINUX) || defined(WINDOWS)`? OpenJ9 now only supports these three platforms, and hence this `defined` seems unnecessary.

runtime/compiler/build/rules/win-msvc/filetypes.mk
```diff
@@ -89,6 +89,22 @@ endef # DEF_RULE.pasm
 
 RULE.pasm=$(eval $(DEF_RULE.pasm))
 
+#
+# Compile .nasm file into .obj file
+#
+define DEF_RULE.nasm
+$(1): $(2) | jit_createdirs
+	$(NASM_CMD) $(NASM_OBJ_FORMAT) $$(patsubst %,-D%=1,$$(NASM_DEFINES)) $$(patsubst %,-I'%/',$$(subst \,/,$$(NASM_INCLUDES))) -o $$(subst \,/,$$@) $$(subst \,/,$$<)
```
0dvictor: The "/" vs "\\" should be the reason why sanity test failed. Replace it with following line works on my machine.
```$(NASM_CMD) $(NASM_OBJ_FORMAT) $$(patsubst %,-D%=1,$$(NASM_DEFINES)) $$(patsubst %,-I%,$$(call FIXPATH,$$(NASM_INCLUDES))) -o $$(subst \,/,$$@) $$(subst \,/,$$<)```

**Comment by nbhuiyan** [comment]
The CI build failures on win64 so far has been due to how `-I` include path search directives are handled in NASM 2.13, and the fact that we are using cygwin for the jenkins builds. Prior to NASM version 2.14, `-I` include paths provided in the command line would not undergo any further processing, and would simply be concatenated with the filename specified in the `%include` preprocessor directive, requiring us to have a trailing `\` or `/` in the include file search paths.

When running NASM through Cygwin, it appears as though the full paths enclosed in quotes for the `-I` command line arguments get processed into a version without the trailing slash, which results in being unable to open the file to include due to the missing path separator between the file name and the containing directory when NASM naively concatenates them. Having absolute paths enclosed in quotes is a good idea in the situation that a parent directory contains a space in its name.

This will not be a problem in NASM version 2.14 and newer, as the `-I` option semantics have been vastly improved and we are no longer required to ensure a trailing slash in the -I paths. If want to use the full path in -I, which is not a requirement, even a slightly older version of NASM will be unable to build the assembly files. Given that NASM 2.14 was released very recently, perhaps it is too soon to make that a requirement when there is a workaround to it.

The workaround (currently in this PR) that I have been able to come up with supports NASM version 2.14 as well as the previous releases. This involves not enclosing the -I paths in quotes, and using relative paths with a trailing slash from `compiler` directory to the include file search directories instead of absolute paths. That way, we do not have to worry about the possibility of spaces in a parent directory resulting in issues. The target file and the source files are still going to be enclosed in quotes without any issues. This will allow us to build the JIT using both Cygwin and native Windows shells.

I would still like the transition to NASM v2.14 to happen soon, as it would clean up some hacks that were needed in CMake, but with this approach, we can make the transition smoother as it would work across the 2 versions.

Currently, the `%include` directives for Windows have been ifdef'd out as I experimented with different command formats. If we can agree on my current implementation, I will apply the same approach to macOS and xLinux as well and clean this up.

**Comment by 0dvictor** [comment]
I don't think requiring NASM 2.14 is unreasonable. The workaround being applies has serve limitations: 1. it is quite common seeing spaces in the path, for example the Desktop folder in Windows 7 is under "Documents and Settings"; 2. search paths of `.inc` files are different from the old PASM ones and other C/C++ files. In addition, none of existing code requires any early version of NASM and hence I don't think requiring the latest version is too soon. As a reference, we introduced the requirement of Clang 7 only one month after it was released.

However, if upgrading/installing the required NASM 2.14 cannot make the deadline of a targeted release, I accept this workaround with two extra notes in the building instructions and/or release notes:
1. Mark NASM 2.14 as a requirement;
2. Document the limitation that no space can exist in build path.

File: doc/build-instructions/Build_Instructions_V11.md [hunk 3]
```diff
@@ -366,6 +366,7 @@ You must install a number of software dependencies to create a suitable build en
 - [Microsoft Visual Studio 2017]( https://visualstudio.microsoft.com/thank-you-downloading-visual-studio/?sku=Community&rel=15), which is the default compiler level used by OpenJDK11; or [Microsoft Visual Studio 2013]( https://go.microsoft.com/fwlink/?LinkId=532495), which is chosen to compile if VS2017 isn't installed.
 - [Freemarker V2.3.8](https://sourceforge.net/projects/freemarker/files/freemarker/2.3.8/freemarker-2.3.8.tar.gz/download)
 - [LLVM/Clang](http://releases.llvm.org/7.0.0/LLVM-7.0.0-win64.exe)
+- [NASM Assembler](https://www.nasm.us/), which would be installed as one of the packages in the `Devel` category in Cygwin.
```
0dvictor: Is there any reason why the Cygwin is picked instead of the official release from https://www.nasm.us/ ?

nbhuiyan: Since the instructions ask to install **all** packages in the Devel category, and the NASM version that would be installed as part of the packages is capable of building it, I thought that would be enough. However, to give us an easier time upgrading to NASM v2.14 in the near future, I have updated the instructions to download from the official website.

0xdaryl:
Jenkins test sanity all jdk11

0xdaryl:
Jenkins test sanity all jdk11

0xdaryl:
Sanity tests have passed, as well as additional stress tests performed within IBM.  This PR is also needed to enable #4249.  Merging.

Issue #3401, opened 2018-10-22
JITaaS: Reduce number of CompInfo_isCompiled messages


mpirvu:
Once JITaaS started to effectively use the code in `isLargeCompiledMethod` the number of `CompInfo_isCompiled` increased substantially. One way to reduce the number of these messages is to piggy-back on the information available in a ResolvedMethod.
The proposal is to replace the code:
```
   TR_OpaqueMethodBlock* methodCallee = calleeResolvedMethod->getPersistentIdentifier();
   if (TR::Compiler->mtd.isCompiledMethod(methodCallee))
```
with a call to 
```
! calleeResolvedMethod->isInterpreted()
```
because the 'interpreted' information is already cached.

PR #3460, opened 2018-10-26
Reduce number of messages related to inlining

dmitry-ten:
JITaaS part of #3506

- Reduce number of calls to IProfiler_searchForMethodSample.
Override methods of ResolvedJ9Method that access fanin info.
This enables us to get number and weight of callers, total weight and
other bucket weight at the creation of client-side mirror
and cache these values on the server.
The number of callers and their weights might change
during compilation, so cached information will not always be
up to date, but it should be a good enough approximation.
Since fanin information is only used at warm and above opt levels,
do not send this information to the server at levels below warm. 

Addresses:#3401

### DIFF
--- a/runtime/compiler/env/j9methodServer.cpp
+++ b/runtime/compiler/env/j9methodServer.cpp
@@ -1144,6 +1144,66 @@ TR_ResolvedJ9JITaaSServerMethod::getExistingJittedBodyInfo()
    }
 
 void
+TR_ResolvedJ9JITaaSServerMethod::getFaninInfo(uint32_t *count, uint32_t *weight, uint32_t *otherBucketWeight)
+   {
+   uint32_t i = 0;
+   uint32_t w = 0;
+   if (otherBucketWeight)
+      *otherBucketWeight = 0;
+
+   TR_IPMethodHashTableEntry *entry = _iProfilerMethodEntry;
+   if (entry)
+      {
+      w = entry->_otherBucket.getWeight();
+      // Iterate through all the callers and add their weight
+      for (TR_IPMethodData* it = &entry->_caller; it; it = it->next)
+         {
+         w += it->getWeight();
+         i++;
+         }
+      if (otherBucketWeight)
+         *otherBucketWeight = entry->_otherBucket.getWeight();
+      }
+   *weight = w;
+   *count = i;
+   }
+
+bool
+TR_ResolvedJ9JITaaSServerMethod::getCallerWeight(TR_ResolvedJ9Method *caller, uint32_t *weight, uint32_t pcIndex)
+   {
+   TR_OpaqueMethodBlock *callerMethod = caller->getPersistentIdentifier();
+   bool useTuples = (pcIndex != ~0);
+
+   //adjust pcIndex for interface calls (see getSearchPCFromMethodAndBCIndex)
+   //otherwise we won't be able to locate a caller-callee-bcIndex triplet
+   //even if it is in a TR_IPMethodHashTableEntry
+   TR_IProfiler *iProfiler = _fe->getIProfiler();
+   if (!iProfiler)
+      return false;
+
+   uintptrj_t pcAddress = iProfiler->getSearchPCFromMethodAndBCIndex(callerMethod, pcIndex, 0);
+
+   TR_IPMethodHashTableEntry *entry = _iProfilerMethodEntry;
+
+   if(!entry)   // if there are no entries, we have no callers!
+      {
+      *weight = ~0;
+      return false;
+      }
+   for (TR_IPMethodData* it = &entry->_caller; it; it = it->next)
+      {
+      if( it->getMethod() == callerMethod && (!useTuples || ((((uintptrj_t) it->getPCIndex()) + TR::Compiler->mtd.bytecodeStart(callerMethod)) == pcAddress)))
+         {
+         *weight = it->getWeight();
+         return true;
+         }
+      }
+
+   *weight = entry->_otherBucket.getWeight();
+   return false;
+   }
+
+void
 TR_ResolvedJ9JITaaSServerMethod::createResolvedJ9MethodMirror(TR_ResolvedJ9JITaaSServerMethodInfo &methodInfo, TR_OpaqueMethodBlock *method, uint32_t vTableSlot, TR_ResolvedMethod *owningMethod, TR_FrontEnd *fe, TR_Memory *trMemory)
    {
    TR_ResolvedJ9Method *resolvedMethod = new (trMemory->trHeapMemory()) TR_ResolvedJ9Method(method, fe, trMemory, owningMethod, vTableSlot);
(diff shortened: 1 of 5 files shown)


dmitry-ten:
Did a performance run, this change does not introduce any throughput regression.

dmitry-ten:
@mpirvu could you review?

mpirvu:
Thinking more about this: should we just serialize all the information when creating a mirror ResolvedMethod and cache this at the server? This way we can answer even `getCallerWeight` queries and all messages to method profiling info would go away.

dmitry-ten:
So you are suggesting that we serialize all callers of a method?
Would it be worth it though? With this change, the number of calls to `searchForMethodSample` is pretty small, if we serialize callers every time a resolved method is created, our network traffic will increase.

mpirvu:
If you determined that the remaining number of calls to searchForMethodSample is reduced very much, then we could skip serialization. My suggestion was based on my recollection that typically the list of callers is just 1.

dmitry-ten:
Number of calls went down from 250k to 20k in AcmeAir, which is not as low as I thought.
I'll measure the average number of callers, to see if it's close to 1.

andrewcraik:
Because there are changes which affect general inliner code a full perf verification will be needed to check the impact of this on non JaaS situations before the merge to mainline occurs - you may want to consider tracking such changes to make the validation ahead of merge easier.

mpirvu:
@andrewcraik FYI: The proposed changes do not change the flow of the non-JITaaS code in any way. Instead of calling some functionality from the frontend we propose to call the same functionality from the ResolvedMethod. Thus, except for intended bugs, all performance metrics should stay the same.

andrewcraik:
Well I do see the new `otherWeight` parameter in the inliner which may just be a straight refactor, but still worthy of a verification. Thank you for the clarification that this is intended to not change behavior.

dmitry-ten:
Yes, `otherWeight` is just a refactor of `getOtherBucketWeight`

dmitry-ten:
I measured the average number of callers during AcmeAir run, it's 18.77.
It is a lot, but the benefit of completely eliminating this message might still be worth it.
@mpirvu what do you think we should do?

mpirvu:
18.77 is a surprise for me. I think that the number of messages is more important that their size, so in the end it might be a win. There is no way to tell unless we measure both.

I wonder if can do something dynamic because cold compilations do not ask for fanin info, so the information sent would be a waste. This means that at the server we need the ability to check whether fanin info was cached, and if not to specifically ask for it.

dmitry-ten:
We could check optLevel, and only send fanin info if it is at least warm.

mpirvu:
Agree on checking the opt level

dmitry-ten:
@mpirvu should be ready for another review.
I'm caching an entire IProfiler method entry in a resolved method now.
This eliminates all messages to `searchForMethodSample` other than the case when `getCallerWeight` is called from `getCallCount`, but that happens really infrequently.

dmitry-ten:
@mpirvu this should be ready for another review.
The only remaining message to `searchForMethodSample` is when `getCallerWeight` gets called from `getCallCount`, but that happens really infrequently.

runtime/compiler/env/j9method.cpp
```diff
@@ -5313,6 +5313,34 @@ TR_ResolvedJ9Method::isJITInternalNative()
...
+TR_ResolvedJ9Method::getFaninInfo(uint32_t *count, uint32_t *weight, uint32_t *otherBucketWeight)
+   {
+   TR_IProfiler *profiler = _fe->getIProfiler();
+
+   if (!profiler)
+      {
+      return;
+      }
+
+   TR_OpaqueMethodBlock *method = getPersistentIdentifier();
```
mpirvu: My suggestion would be to create a getFaninInfo at the IProfiler as well and call it here. Thus we can have one virtual call instead of two.
`profiler->getNumberofCallersAndTotalWeight()` and `profiler->getOtherBucketWeight()` can be eliminated then.

dmitry-ten:
I want to deliver the mainline part first.
Once that is done and jitaas is rebased, I will keep just JITaaS specific code in this PR.

runtime/compiler/env/j9methodServer.hpp
```diff
@@ -62,10 +63,13 @@ TR_ResolvedJ9JITaaSServerMethodInfoStruct
    bool virtualMethodIsOverridden;
    void *addressContainingIsOverriddenBit;
    J9ClassLoader *classLoader;
+   uint32_t numCallers;
```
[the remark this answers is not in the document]
dmitry-ten: Nope, just forgot to delete them.

runtime/compiler/env/j9methodServer.cpp
```diff
@@ -1185,6 +1245,11 @@ TR_ResolvedJ9JITaaSServerMethod::createResolvedJ9MethodMirror(TR_ResolvedJ9JITaa
    std::string methodInfoStr = bodyInfo ? std::string((char*)bodyInfo->getMethodInfo(), sizeof(TR_PersistentMethodInfo)) : std::string();
    std::get<1>(methodInfo) = jbi;
    std::get<2>(methodInfo) = methodInfoStr;
+
+   // set IP method data string.
+   // fanin info is not used at cold opt level, so there is no point sending this information to the server
+   TR_JITaaSClientIProfiler *iProfiler = (TR_JITaaSClientIProfiler *)((TR_J9VMBase *) fe)->getIProfiler();
+   std::get<3>(methodInfo) = (TR::comp() && TR::comp()->getOptLevel() >= warm && iProfiler) ? iProfiler->serializeIProfilerMethodEntry(method) : std::string();
```
mpirvu: Is it ever possible for TR::comp() to return 0 at this point?

dmitry-ten: Yes, at the beginning of compilation, a resolved method for the method being compiled gets created, at that point we don't have a comp object yet.

runtime/compiler/env/j9methodServer.cpp
```diff
@@ -1236,6 +1301,11 @@ TR_ResolvedJ9JITaaSServerMethod::setAttributes(TR_OpaqueMethodBlock * aMethod, T
   
    setMandatoryRecognizedMethod(mandatoryRm);
    setRecognizedMethod(rm);
+
+   TR_JITaaSIProfiler *iProfiler = (TR_JITaaSIProfiler *) ((TR_J9VMBase *) fe)->getIProfiler();
+   const std::string entryStr = std::get<3>(methodInfo);
+   const auto serialEntry = (TR_ContiguousIPMethodHashTableEntry*) &entryStr[0];
+   _iProfilerMethodEntry = (TR::comp() && iProfiler && !entryStr.empty()) ? iProfiler->deserializeMethodEntry(serialEntry) : nullptr;
```
mpirvu: I think that here TR::comp() can return nullptr because when first create the `compilee` and then we create the compilation object. However, we don't want to have _iProfilerMethodEntry null for the method to be compiled. I think that the only reason you want the test for TR:comp() is because `deserializeMethodEntry` needs to allocate heap memory. In this case it could be easier if we provide it with a pointer to trMemory object available in this routine.

PR #3462, opened 2018-10-26
WIP: Ensure GC map is generated at BB entry on X86

### LINKED ISSUE #3374: Ensure stack maps are generated for exception catch blocks on x86
Please see the parent Epic for further details. We need to ensure that under the new mode we generate stack maps at the start of every exception catch block and ensure that we don't enable throw to goto transformations when this mode is enabled.

0dvictor:
Stack maps are required at the entry of exception catch
blocks under FSD mode and when Agent is attached.

Closes #3374

### DIFF
--- a/runtime/compiler/x/codegen/J9TreeEvaluator.cpp
+++ b/runtime/compiler/x/codegen/J9TreeEvaluator.cpp
@@ -11506,9 +11506,9 @@ VMgenerateCatchBlockBBStartPrologue(
    TR::Compilation *comp = cg->comp();
    TR_J9VMBase *fej9 = (TR_J9VMBase *)(cg->fe());
 
-   if (comp->getOption(TR_FullSpeedDebug))
+   if (comp->getOptions()->getReportByteCodeInfoAtCatchBlock())
       {
-      fenceInstruction->setNeedsGCMap(); // a catch entry is a gc point in FSD mode
+      node->getBlock()->getFirstInstruction()->setNeedsGCMap();
       }
 
    if (comp->getJittedMethodSymbol()->usesSinglePrecisionMode() &&


runtime/compiler/x/codegen/J9TreeEvaluator.cpp
```diff
@@ -11551,9 +11551,9 @@ VMgenerateCatchBlockBBStartPrologue(
    TR::Compilation *comp = cg->comp();
    TR_J9VMBase *fej9 = (TR_J9VMBase *)(cg->fe());
 
-   if (comp->getOption(TR_FullSpeedDebug))
+   if (comp->getOption(TR_FullSpeedDebug) || comp->getOption(TR_DisableThrowToGoto))
```
andrewcraik: This should be checking the appropriate VM flag - whether the opt is disabled or not should not change the map since we may disable the opt for debugging.

0dvictor: Totally agree, and that's why this pull request is currently marked as WIP. Once we decide what the "proper" flag is, I'll update this check. https://github.com/eclipse/openj9/issues/3373#issuecomment-433514777

fjeremic: According to #3389 the right "flag" should be checking `fej9 >getJ9JITConfig()->jitExceptionCaught != NULL` since we do not initialize this callback if we do not need to call it. Or in other words we initialize the callback when FSD is on or when we request exception catch hook via agent.

fjeremic: This seems to be the correct way to do things as it is exactly the same thing the VM checks [1]. Additionally if you check the contents of #3389 and the code sorrounding thet change you will see we only initialize `jitExceptionCaught` if we have to call it in the VM, i.e. under FSD or when a hook is attached. We should document this in the `J9JITConfig` struct. I can take care of that.

[1] https://github.com/eclipse/openj9/pull/3368/files#diff-5ca144eb3d6b1334698c9840002b59c3R2432

fjeremic: Having thought about this some more I'm not sure what the VM is doing is the best thing here. I don't quite understand why we're not always initializing `jitExceptionCaught` and always calling it. 

From the JIT point of view we need an explicit "do we need to ensure correct BCI info is present at catch and throw locations". Checking if `jitExceptionCaught` is non-NULL is not a great solution here. There is a similar approach taken in [2] where we set the `TR_ReportMethodEnter` and `TR_ReportMethodExit` options. Perhaps we need to introduce `TR_ReportMethodThrow` and `TR_ReportMethodCatch` accordingly and use them?

[2] https://github.com/eclipse/openj9/blob/87d9bea1aef33f69500e3b8035254a20987721d2/runtime/compiler/control/J9Options.cpp#L2205-L2223

fjeremic: @0dvictor @andrewcraik thoughts on above? We need to settle on a JIT query to proceed with all the platforms. Whatever solution we come up with should be consistent w.r.t. other hooks available since we'll soon have field read and write hooks as well in non-FSD.

0dvictor:
This PR requires #3463

runtime/compiler/x/codegen/J9TreeEvaluator.cpp
0dvictor: The code has been updated according to changes in #3463. For discussions related to the flag to query the status of JVM hook, please use #3373 and/or #3463.

runtime/compiler/x/codegen/J9TreeEvaluator.cpp
```diff
@@ -11551,9 +11551,9 @@ VMgenerateCatchBlockBBStartPrologue(
    TR::Compilation *comp = cg->comp();
    TR_J9VMBase *fej9 = (TR_J9VMBase *)(cg->fe());
 
-   if (comp->getOption(TR_FullSpeedDebug))
+   if (comp->getOption(TR_FullSpeedDebug) || comp->getOptions()->isHookEnabled(TR::Options::J9HookExceptionCatch))
```
fjeremic: Does FSD imply that we should always configure the JIT method bodies for all the hooks? In which case we should be checking only the `TR::Options::J9HookExceptionCatch` here. Where FSD is enabled we should enabled all the corresponding hooks that we need.

0dvictor: Good suggestion. It has been updated accordingly.

fjeremic:
Jenkins test sanity win,xlinux JDK8

fjeremic:
Travis build failed because #3463 was merged after the build ran. I don't envision this commit chaning so launched builds preemptively. I won't merge until WIP is removed.

0dvictor:
I rebased the changes so that Travis CI build is relaunched.

fjeremic:
Jenkins test sanity win,xlinux JDK8

fjeremic:
#3598 should handle all the platforms at a common point. @0dvictor if you agree we can close this PR and I'll appreciate a formal review in #3598.

0dvictor:
Totally agreed, #3598 supersets what this PR achieves. I'm marking this PR as WIP for now and will close it once #3598 is merged.

PR #3463, opened 2018-10-26
Ensure BCI in catch block stack maps are correct

0dvictor:
Ensure bytecode info at the entry of an exception
catch block is correct when Agent is attached.


0dvictor:
This PR requires #3462

runtime/compiler/runtime/MetaData.cpp
```diff
@@ -700,6 +703,8 @@ mapsAreIdentical(
        mapCursor->getMapSizeInBytes() == nextMapCursor->getMapSizeInBytes() &&
        mapCursor->getRegisterMap() == nextMapCursor->getRegisterMap() &&
        !memcmp(mapCursor->getMapBits(), nextMapCursor->getMapBits(), mapCursor->getMapSizeInBytes()) &&
+       (nonmergeableBCI.find(mapCursor) == nonmergeableBCI.end() ||
+        mapCursor->isByteCodeInfoIdenticalTo(nextMapCursor)) &&
```
fjeremic: Doesn't adding the `isByteCodeInfoIdenticalTo` check make the `!comp->getOption(TR_FullSpeedDebug)` check redundant (and the comment as well)? Can we replace the FSD check with this `isByteCodeInfoIdenticalTo` check and pull it out of this `||` statement?

runtime/compiler/control/J9Options.hpp
```diff
@@ -64,6 +70,35 @@ class OMR_EXTENSIBLE Options : public OMR::OptionsConnector
...
+    */
+   static inline void disableHook(J9HookTypes hook) { _hookStatus &= ~hook; }
+   /** \brief
+    *     Check if a JVM hook is enabled
+    *
+    *  \param hook
+    *     The JVM hook to be checked
+    *
+    *  \return
+    *     True of the JVM hook is enabled otherwise false
```
fjeremic: I don't think this is true. If `J9HookDisable` returns a non-zero value for a particular hook it implies that a hook `can` be enabled, not that it is. This API should answer the question of whether a particular hook _can be enabled_. That is, the JIT should do what it needs to do to make sure that when a hook is enabled the JIT methods can report the hooked event.

fjeremic: @gacholio can you confirm the above? If `(*vmHooks)->J9HookDisable(vmHooks, J9HOOK_VM_EXCEPTION_CATCH)` returns a non-NULL value does it imply the hook is enabled or that it _may_ be enabled at some point in the future?

DanHeidinga: If J9HookDisable returns 0, the hook is disabled and can't be enabled in the future.  If it returns non-zero, the hook is either enabled or reserved

https://github.com/eclipse/omr/blob/f8fc704032f95cbc0f337eb6be54f2b06e58a773/util/hookable/hookable.cpp#L277-L286

gacholio: Also note the need for exception throw to be mapped here (both throw and catch require the accurate stack maps), so I'm not sure this is the direction you want to go in.

fjeremic: Thanks @DanHeidinga! So my suspicions are correct, we don't know whether it's actually hooked or reserved, so this new code is ill-defined. @0dvictor can we make appropriate changes to API?

@gacholio yes we will have a `J9HookExceptionThrow` as well. I think at the moment we already generate accurate stack maps for throws, but we will double check.

gacholio: The throw events reports both the thrower and the likely catcher, which is why I need the accurate catch data for throw. This is an existing bug in that throw does not currently force FSD.

gacholio: Do we need this much overdesign for this feature? Surely a single option describing that exception catch maps need to be correct would be simpler and more accurate. At best, this new hook data can apply only to things disabled by JIT startup - it's not going to be a current view of the actual hook world.

0dvictor: Good point, I guess we don't need a query for whether a hook may present. All we need is a property indicating whether we need preserve the bytecode info at the beginning of a Catch block.

0dvictor: I changed `_hookStatus` flag into one simple property `_reportByteCodeInfoAtCatchBlock`, so that it isn't tight to the definition of `J9HookDisable` returning non-zero. What do you guys think?

fjeremic: Sounds good to me! Thanks @0dvictor!

0dvictor:
I've updated the code based on the discussion above.

fjeremic:
Jenkins test sanity all JDK8

0dvictor:
The change is updated so that `ReportByteCodeInfoAtCatchBlock` is set under Full Speed Debug mode, as FSD mode also requires bytecode info at a catch block.

fjeremic:
Jenkins test sanity all JDK8

PR #3464, opened 2018-10-26
Remove redundant checks for SSE/SSE2 on X86

0dvictor:
SSE/SSE2 is part of prerequisites of OpenJ9 on X86, and
JVM guarantees the availablity of SSE/SSE2; therefore,
checking for SSE/SSE2 is redundant and should be removed.

### DIFF
--- a/runtime/compiler/x/codegen/X86PrivateLinkage.cpp
+++ b/runtime/compiler/x/codegen/X86PrivateLinkage.cpp
@@ -1118,15 +1118,6 @@ TR::X86PrivateLinkage::buildDirectDispatch(
    //
    buildCallArguments(site);
 
-   if (spillFPRegs && !cg()->useSSEForDoublePrecision())
-      {
-      // Force the FP register stack to be spilled.
-      //
-      TR::RegisterDependencyConditions *fpSpillDependency = generateRegisterDependencyConditions(1, 0, cg());
-      fpSpillDependency->addPreCondition(NULL, TR::RealRegister::AllFPRegisters, cg());
-      generateInstruction(FPREGSPILL, callNode, fpSpillDependency, cg());
-      }
-
    // Remember where internal control flow region should start,
    // and create labels
    //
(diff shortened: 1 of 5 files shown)


runtime/compiler/x/codegen/X86PrivateLinkage.cpp
```diff
@@ -2377,23 +2359,20 @@ TR::Register *TR::X86PrivateLinkage::buildCallPostconditions(TR::X86CallSite &si
...
-      if (cg()->useSSEForSinglePrecision() || cg()->useSSEForDoublePrecision())
+      TR_LiveRegisters *lr = cg()->getLiveRegisters(TR_FPR);
+      if(!lr || lr->getNumberOfLiveRegisters() > 0)
          {
-         TR_LiveRegisters *lr = cg()->getLiveRegisters(TR_FPR);
-         if(!lr || lr->getNumberOfLiveRegisters() > 0)
+         for (regIndex = TR::RealRegister::FirstXMMR; regIndex <= TR::RealRegister::LastXMMR; regIndex = (TR::RealRegister::RegNum)(regIndex + 1))
             {
-            for (regIndex = TR::RealRegister::FirstXMMR; regIndex <= TR::RealRegister::LastXMMR; regIndex = (TR::RealRegister::RegNum)(regIndex + 1))
+            TR_ASSERT(regIndex != highReturnRegIndex, "assertion failure");
```
andrewcraik: While you are here can you fix the assert message please?

0dvictor: I've updated the assert message; though I'm not sure if this assertion has any meaningful usage here. `highReturnRegIndex` is the register for returning high 32 bits of a 64-bit primitive on 32-bit ISA. It should only be a GPR and hence not an XMM register.

andrewcraik: Do we need a better assert?

0dvictor: Probably not. This loop is to kill XMM registers that are neither the return register nor any of the preserved registers. I guess the original author wanted to be defensive so that it can detect a misused "high return register"?

andrewcraik:

andrewcraik:
JDK11 win failure is infra - the rest of the builds are fine.

PR #3465, opened 2018-10-26
Move Guarded Load test in OpenJ9

r30shah:
Test to decide if we need to generate Guareded Load in OOL for Concurrent Scavenge for Array Copy is moved to OpenJ9.

### DIFF
--- a/runtime/compiler/z/codegen/J9TreeEvaluator.cpp
+++ b/runtime/compiler/z/codegen/J9TreeEvaluator.cpp
@@ -9652,6 +9652,42 @@ J9::Z::TreeEvaluator::VMinlineCallEvaluator(TR::Node * node, bool indirect, TR::
    return callWasInlined;
    }
 
+void
+J9::Z::TreeEvaluator::genGuardedLoadOOL(TR::Node *node, TR::CodeGenerator *cg, TR::Register *byteSrcReg, TR::Register *byteDstReg, TR::Register *byteLenReg, TR::LabelSymbol *mergeLabel, TR_S390ScratchRegisterManager *srm, bool isForward)
+   {
+   TR_ASSERT_FATAL(J9_PRIVATE_FLAGS_CONCURRENT_SCAVENGER_ACTIVE == 0x20000,
+               "GSCS: The OOL sequence branch is dependant on the flag being 0x20000");
+   TR_ASSERT_FATAL(TR::Compiler->target.is64Bit(),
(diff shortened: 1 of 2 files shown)


r30shah:
This depends on https://github.com/eclipse/omr/pull/3126

r30shah:
@fjeremic  Can I please get review for this? I moved Concurrent Scavenge test from OMR to here.

runtime/compiler/z/codegen/J9TreeEvaluator.cpp
```diff
@@ -9609,6 +9609,42 @@ J9::Z::TreeEvaluator::VMinlineCallEvaluator(TR::Node * node, bool indirect, TR::
...
+   TR::LabelSymbol* slowPathLabel = generateLabelSymbol(cg);
+
+   TR::Register *vmReg = cg->getMethodMetaDataRealRegister();
+   TR::MemoryReference *privFlagMR = generateS390MemoryReference(vmReg,
+   TR::Compiler->vm.thisThreadGetConcurrentScavengeActiveByteAddressOffset(cg->comp()), cg);
+
+   generateSIInstruction(cg, TR::InstOpCode::TM, node, privFlagMR, 0x00000002);
+   generateS390BranchInstruction(cg, TR::InstOpCode::BRC, TR::InstOpCode::COND_CC3, node, slowPathLabel);
+   // Generate OOL Slow Path
+   TR_S390OutOfLineCodeSection* outOfLineCodeSection = new (cg->trHeapMemory()) TR_S390OutOfLineCodeSection(slowPathLabel, mergeLabel, cg);
```
fjeremic: Please avoid using placement `new` as there are APIs to handle this. Similarly below for dependencies.

r30shah: @fjeremic I looked at the available API for OutOfLine Code Section. I could not find one that generates `TR_S390OutOfLineCodeSection` . Also for the register dependency, I think we need to add more `generate*` to the API . In this case I need to construct new Dependency using available deps. We have constructor to do that but no `generate*` API. We should certainly add new functions to RegisterDependency API but that is out of this PR's scope and we need separate PR which adds new functions and updates code.

fjeremic: Ah you're right!

r30shah:
With https://github.com/eclipse/omr/pull/3167 merged, this is a safe change to merge.

fjeremic:

Issue #3276, opened 2018-10-15
Software based Concurrent Scavenge support on Z


fjeremic:
This Epic tracks the work needed to enable software based concurrent scavenge (CS) support on Z.

Currently we only support Concurrent Scavenge on z14+ processors which have Guarded Storage (GS) support on certain levels of OS. This imposes quite a restriction on the environment needed to run with CS on Z. The software based read barrier solution will enable us to run with CS on z13- hardware and will aid in testing of the CS feature on our build farm.

pshipton:
@fjeremic now that #3283 is closed, can this also be closed?

fjeremic:
> @fjeremic now that #3283 is closed, can this also be closed?

Indeed. All related issues have been addressed.

PR #3474, opened 2018-10-29
Rename VMThread evacuate fields

amicic:
Renaming evacuateBase/Top fields to readBarrierRangeCheckBase/Top.
Evacuate is a specific action of Read Barrier for Concurrent Scavegner
GC, while the trigger is range check against the bounds. This check can
be used to trigger the barrier that will do completely different action
(for some diffrent type of GC, some profiler/debug tool etc). 

Similar things have been done already for GC specific atributes:
https://github.com/eclipse/omr/pull/3066

### DIFF
--- a/runtime/gc_glue_java/CollectorLanguageInterfaceImpl.cpp
+++ b/runtime/gc_glue_java/CollectorLanguageInterfaceImpl.cpp
@@ -680,16 +680,16 @@ MM_CollectorLanguageInterfaceImpl::scavenger_switchConcurrentForThread(MM_Enviro
 		void* top = _extensions->scavenger->getEvacuateTop();
 
 		/*
-		 * vmThread->evacuateTop is defined as the last address possible in the evacuate region;
+		 * vmThread->readBarrierRangeCheckTop is defined as the last address possible in the evacuate region;
 		 * however, top is the first address above the evacuate region;
-		 * therefore, vmThread->evacuateTop = top - 1.
-		 * In short, the evacuate region is [vmThread->evacuateBase, vmThread->evacuateTop].
+		 * therefore, vmThread->readBarrierRangeCheckTop = top - 1.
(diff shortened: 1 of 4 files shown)


amicic:
@fjeremic I did not change the names of JIT helpers like thisThreadGetEvacuateBaseAddressOffset. I believe they should be renamed too, but did not want to go too far into JIT code.

runtime/gc_glue_java/CollectorLanguageInterfaceImpl.cpp
```diff
@@ -721,17 +721,17 @@ MM_CollectorLanguageInterfaceImpl::scavenger_switchConcurrentForThread(MM_Enviro
 			j9gs_disable(&vmThread->gsParameters);
 		}
 		/*
-		 * By setting evacuateTop to NULL and evacuateBase to ~evacuateTop
+		 * By setting readBarrierRangeCheckTop to NULL and readBarrierRangeCheckBase to ~readBarrierRangeCheckTop
```
DanHeidinga: This comment isn't accurate with the code below.  Can you update it?

amicic: updated

DanHeidinga:
Jenkins test sanity plinux,zlinux jdk8

DanHeidinga:

DanHeidinga:
@fjeremic Any comments on this PR?  Otherwise I'll merge as is

fjeremic:
> @fjeremic I did not change the names of JIT helpers like thisThreadGetEvacuateBaseAddressOffset. I believe they should be renamed too, but did not want to go too far into JIT code.

Thanks I'm referencing this with #3276 so we know to clean it up, though there is already an item opened for similar cleanup. I'll take care of it.

> @fjeremic Any comments on this PR? Otherwise I'll merge as is

Should be good to go!

PR #3479, opened 2018-10-29
Fix debug build break

mpirvu:
Some TR_ASSERT needed to be placed after the statement that defines the
data it uses, not before.


mpirvu:
@ymanton Younes, please review and merge this asap. My bad on compiling a prod build instead of debug one.

PR #3486, opened 2018-10-30
Return true for canGCAndReturn for rdbar/wrtbar

cathyzhyi:
GC can happen at field watch helpers.


cathyzhyi:
waiting for jenkins builds pick up depending change https://github.com/eclipse/omr/pull/3008

fjeremic:
I think we're missing an OMR change for the option?

cathyzhyi:
@fjeremic yes, I just pushed again to see if Jenkins picked up the change in omr containing the option.

PR #3487, opened 2018-10-30
Check for MODULE access on MethodHandle

pdbain-ibm:
Lookup objects without MODULE access cannot a package in the same module or in
other modules unless the package is unconditionally exported.  They also cannot
access packages in modules they do not read.


pdbain-ibm:
Manually tested on the JTreg test for https://github.com/eclipse/openj9/issues/3411.

smlambert:
Jenkins compile xLinux jdk11

pdbain-ibm:
@DanHeidinga and @tajila would you kindly review this?

pdbain-ibm:
@pshipton  or @gacholio and @tajila would you kindly run a pull request build and review this?
Thank you.

jcl/src/java.base/share/classes/java/lang/invoke/MethodHandles.java
```diff
@@ -752,17 +752,24 @@ static void checkClassModuleVisibility(int accessMode, Module accessModule, Clas
...
 						throw new IllegalAccessException(Msg.getString("K0587", getModuleName(accessModule), targetClassPackageName, getModuleName(targetModule))); //$NON-NLS-1$
 					}
-				} else if (!(
-					Objects.equals(accessModule, targetModule)
-					|| (targetModule.isExported(targetClassPackageName, accessModule) && accessModule.canRead(targetModule)))
-				) {
-					if (!targetModule.isExported(targetClassPackageName, accessModule)) {
-						/*[MSG "K0677", "Module '{0}' no access to: package '{1}' which is not exported by module '{2}' to module '{0}'"]*/
-						throw new IllegalAccessException(Msg.getString("K0677", getModuleName(accessModule), targetClassPackageName, getModuleName(targetModule))); //$NON-NLS-1$
+				} else if (Objects.nonNull(targetModule)) { // targetModule may be null during bootstrapping
```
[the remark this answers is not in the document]
tajila: accessModule may also be null during bootstraping

gacholio:
Waiting for Tobi's comments to be addressed.

pdbain-ibm:
@gacholio and @tajila Updated and ready for review.

jcl/src/java.base/share/classes/java/lang/invoke/MethodHandles.java
```diff
@@ -746,28 +746,33 @@ private static String getModuleName(Module module) {
 		 * @param targetClass Class which the referring class is accessing
 		 * @throws IllegalAccessException if the targetClass is not visible
 		 */
-		static void checkClassModuleVisibility(int accessMode, Module accessModule, Class<?> targetClass) throws IllegalAccessException {
+		static void checkClassModuleVisibility(int accessMode, Module accessModule, Class<?> targetClass) 
+				throws IllegalAccessException {
```
gacholio: Unnecessary (and doesn't look any better).

jcl/src/java.base/share/classes/java/lang/invoke/MethodHandles.java
```diff
@@ -746,28 +746,33 @@ private static String getModuleName(Module module) {
...
-					}
-				} else if (!(
-					Objects.equals(accessModule, targetModule)
-					|| (targetModule.isExported(targetClassPackageName, accessModule) && accessModule.canRead(targetModule)))
-				) {
-					if (!targetModule.isExported(targetClassPackageName, accessModule)) {
-						/*[MSG "K0677", "Module '{0}' no access to: package '{1}' which is not exported by module '{2}' to module '{0}'"]*/
-						throw new IllegalAccessException(Msg.getString("K0677", getModuleName(accessModule), targetClassPackageName, getModuleName(targetModule))); //$NON-NLS-1$
+				// modules may be null during bootstrapping
+				if (Objects.nonNull(targetModule) && Objects.nonNull(targetModule)) {
```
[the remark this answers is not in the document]
gacholio: Also, is there a compelling reason not to simply use `null == variable` here?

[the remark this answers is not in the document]
pdbain-ibm: > Did you intend to null check accessModule here

Oops.  fixed.

> null == variable 

No compelling reason, other than clarity.  I can change it back if you want.

jcl/src/java.base/share/classes/java/lang/invoke/MethodHandles.java
```diff
@@ -750,24 +750,28 @@ static void checkClassModuleVisibility(int accessMode, Module accessModule, Clas
...
-					if (!targetModule.isExported(targetClassPackageName, accessModule)) {
-						/*[MSG "K0677", "Module '{0}' no access to: package '{1}' which is not exported by module '{2}' to module '{0}'"]*/
-						throw new IllegalAccessException(Msg.getString("K0677", getModuleName(accessModule), targetClassPackageName, getModuleName(targetModule))); //$NON-NLS-1$
+				// modules may be null during bootstrapping
+				if ((null != targetModule) && (null != accessModule)) { 
+					String targetClassPackageName = targetClass.getPackageName();
+					if ((UNCONDITIONAL & accessMode) == UNCONDITIONAL) {
+						/* publicLookup objects can see all unconditionally exported packages. */
+						if (!targetModule.isExported(targetClassPackageName)) {
+							/*[MSG "K0676", "Module '{0}' no access to: package '{1}' which is not exported by module '{2}'"]*/
```
gacholio: The message number in the comment does not match the code below.

jcl/src/java.base/share/classes/java/lang/invoke/MethodHandles.java
```diff
@@ -750,24 +750,28 @@ static void checkClassModuleVisibility(int accessMode, Module accessModule, Clas
...
+							throw new IllegalAccessException(Msg.getString("K0587", getModuleName(accessModule), targetClassPackageName, getModuleName(targetModule))); //$NON-NLS-1$
```
DanHeidinga: There are three `throw new Illegal...` clauses in this method which generates a lot of bytecode.  It's often cleaner to pull this out into a method like:

```java
IllegalAccessException throwNewIAEForModuleAccess(String message, Module accessModule, String targetClassPackageName, Module targetModule) {
 throw new IllegalAccessException(Msg.getString(message, getModuleName(accessModule), targetClassPackageName, getModuleName(targetModule)));
}
```
and then call with code like:
```java
throw throwNewIAEForModuleAccess(.....);
```
as this makes it clear to the reader that execution can't continue past that call.

jcl/src/java.base/share/classes/java/lang/invoke/MethodHandles.java
```diff
@@ -750,24 +750,28 @@ static void checkClassModuleVisibility(int accessMode, Module accessModule, Clas
...
+						}
+					} else if (!accessModule.canRead(targetModule)) {
+						/*[MSG "K0679", "Module '{0}' no access to: package '{1}' because module '{0}' can't read module '{2}'"]*/
+						throw new IllegalAccessException(Msg.getString("K0679", getModuleName(accessModule), targetClassPackageName, getModuleName(targetModule))); //$NON-NLS-1$					
+					} else {
+						boolean isPackageModuleConditionallyExported = ((MODULE & accessMode) == MODULE) // Need MODULE access to see packages not unconditionally exported
+								&& (Objects.equals(accessModule, targetModule) || targetModule.isExported(targetClassPackageName, accessModule)
+										);
+						if (!targetModule.isExported(targetClassPackageName)
+								&& !isPackageModuleConditionallyExported) {
```
DanHeidinga: Given the line lengths in this file, just do this on a single line

jcl/src/java.base/share/classes/java/lang/invoke/MethodHandles.java
pdbain-ibm: Also legacy code, but I really like your idea.

jcl/src/java.base/share/classes/java/lang/invoke/MethodHandles.java
```diff
@@ -750,29 +750,38 @@ static void checkClassModuleVisibility(int accessMode, Module accessModule, Clas
...
+						/* publicLookup objects can see all unconditionally exported packages. */
+						if (!targetModule.isExported(targetClassPackageName)) {
+							throw newIllegalAccessException(accessModule, targetModule, targetClassPackageName, "K0676"); //$NON-NLS-1$
+						}
+					} else if (!accessModule.canRead(targetModule)) {
+						throw newIllegalAccessException(accessModule, targetModule, targetClassPackageName, "K0679"); //$NON-NLS-1$
+					} else if (!targetModule.isExported(targetClassPackageName)) {
+						// Need MODULE access to see packages conditionally exported
+						if (((MODULE & accessMode) != MODULE) 
+								|| (!Objects.equals(accessModule, targetModule)
```
gacholio: Is this just a slower way of saying == ?

jcl/src/java.base/share/classes/java/lang/invoke/MethodHandles.java
```diff
@@ -750,29 +750,38 @@ static void checkClassModuleVisibility(int accessMode, Module accessModule, Clas
```
gacholio: Blank line here should be removed.

pdbain-ibm: Will do.

jcl/src/java.base/share/classes/java/lang/invoke/MethodHandles.java
```diff
@@ -750,29 +750,38 @@ static void checkClassModuleVisibility(int accessMode, Module accessModule, Clas
...
 			}
 		}
-		/*[ENDIF]*/
+
+		private static IllegalAccessException newIllegalAccessException(Module accessModule, Module targetModule,
+				String targetClassPackageName, String msgId) throws IllegalAccessException {
+			/*[MSG "K0676", "Module '{0}' no access to: package '{1}' which is not exported by module '{2}'"]*/
+			/*[MSG "K0677", "Module '{0}' no access to: package '{1}' which is not exported by module '{2}' to module '{0}'"]*/
+			/*[MSG "K0679", "Module '{0}' no access to: package '{1}' because module '{0}' can't read module '{2}'"]*/
+			return new IllegalAccessException(Msg.getString(msgId, getModuleName(accessModule), targetClassPackageName, getModuleName(targetModule))); //$NON-NLS-1$
```
DanHeidinga: `//$NON-NLS-1$` unneeded

jcl/src/java.base/share/classes/java/lang/invoke/MethodHandles.java
```diff
@@ -750,29 +750,38 @@ static void checkClassModuleVisibility(int accessMode, Module accessModule, Clas
...
-					// here accessModule.canRead(targetModule) must be false
-					/*[MSG "K0679", "Module '{0}' no access to: package '{1}' because module '{0}' can't read module '{2}'"]*/
-					throw new IllegalAccessException(Msg.getString("K0679", getModuleName(accessModule), targetClassPackageName, getModuleName(targetModule))); //$NON-NLS-1$
 				}
 			}
 		}
-		/*[ENDIF]*/
+
+		private static IllegalAccessException newIllegalAccessException(Module accessModule, Module targetModule,
+				String targetClassPackageName, String msgId) throws IllegalAccessException {
```
DanHeidinga: The rest of this file uses long lines - please match the existing style

jcl/src/java.base/share/classes/java/lang/invoke/MethodHandles.java
```diff
@@ -750,29 +750,38 @@ static void checkClassModuleVisibility(int accessMode, Module accessModule, Clas
...
+						if (!targetModule.isExported(targetClassPackageName)) {
+							throw newIllegalAccessException(accessModule, targetModule, targetClassPackageName, "K0676"); //$NON-NLS-1$
+						}
+					} else if (!accessModule.canRead(targetModule)) {
+						throw newIllegalAccessException(accessModule, targetModule, targetClassPackageName, "K0679"); //$NON-NLS-1$
+					} else if (!targetModule.isExported(targetClassPackageName)) {
+						// Need MODULE access to see packages conditionally exported
+						if (((MODULE & accessMode) != MODULE) 
+								|| (!Objects.equals(accessModule, targetModule) 
+										&& !targetModule.isExported(targetClassPackageName, accessModule))) {
```
DanHeidinga: The indentation on this `if` block is strange as well - why the double indents?

jcl/src/java.base/share/classes/java/lang/invoke/MethodHandles.java
pdbain-ibm: Only if the `Module` objects for a given module are guaranteed to be unique.  Is that the case @DanHeidinga ?

DanHeidinga: I'm not aware of any specification requiring Modules to be interned but it's an obvious implementation strategy.  

IMHO, the Objects.equals() call is fine here

gacholio: Is it better than a simple call to equals() ?

pdbain-ibm: `Object.equals()` would be safe if one or both arguments were null, though they aren't in this case.

gacholio: In the absence of a good reason to use the Objects methods, I'd prefer to see normal java code to keep the interpreted execution as fast as possible.

gacholio:

jcl/src/java.base/share/classes/java/lang/invoke/MethodHandles.java
```diff
@@ -747,32 +747,39 @@ private static String getModuleName(Module module) {
...
 				}
 			}
 		}
-		/*[ENDIF]*/
+
+		private static IllegalAccessException newIllegalAccessException(Module accessModule, Module targetModule, String targetClassPackageName, String msgId) throws IllegalAccessException {
+			/*[MSG "K0676", "Module '{0}' no access to: package '{1}' which is not exported by module '{2}'"]*/
+			/*[MSG "K0677", "Module '{0}' no access to: package '{1}' which is not exported by module '{2}' to module '{0}'"]*/
+			/*[MSG "K0679", "Module '{0}' no access to: package '{1}' because module '{0}' can't read module '{2}'"]*/
+			return new IllegalAccessException(Msg.getString(msgId, getModuleName(accessModule), targetClassPackageName, getModuleName(targetModule)));
```
DanHeidinga: This should throw rather than return so that anyone forgetting to throw the exception gets it automatically.  My bad on suggesting the original method name - something like `throwNewIllegalAccessException` would be more clear.

The reason to have the caller add a throw as well is to make it clear that execution can't continue past that call.

DanHeidinga:

Issue #3398, opened 2018-10-22
Always fill in return values in JVMTI


gacholio:
GetOwnedMonitorStackDepthInfo (and quite likely others) does not always fill in the return buffer pointer (in particular, if there are no monitors, the call returns success, fills in the buffer size to 0, but does not fill in the buffer pointer with NULL.

All JVMTI calls return an error code. Any other returned information is done via pointers passed to the function. The specification states that the return values will be filled in - it does not say that the call has to return success in order to do so.

PR #3492, opened 2018-10-31
Ensure JVMTI return values are filled in

gacholio:
Fix all appropriate JVMTI calls to store return value information in
local variables which are initialized to consistent values, and assign
those values at the end of the function unconditionally to any return
pointer which is not NULL.

This does introduce some extra NULL checks for return pointers, but
should have no noticeable performance impact.

Notably not included in this are parameters which describe a
user-allocated buffer (the buffers may or may not be fully or partially
filled in in error situations).

Fixes: #3398

[ci skip]


DanHeidinga:
Jenkins test all xlinux jdk8

PR #3499, opened 2018-10-31
Update class visibility checking

pdbain-ibm:
Fixes https://github.com/eclipse/openj9/issues/3047


pdbain-ibm:
Travis build failed due to lint issue.  Passed on restart.  Please run functional sanity and extended tests.  Linux X86 should be sufficient.

pdbain-ibm:
Let's wait until 
https://github.com/eclipse/openj9/pull/3487 is merged before we test this.

pdbain-ibm:
Ready for review.

jcl/src/java.base/share/classes/java/lang/invoke/MethodHandles.java
```diff
@@ -512,33 +512,27 @@ private void checkAccess(Class<?> targetClass, String name, int memberModifiers,
...
-				if (accessClass == targetClass) {
-					return;
-				}
-				
-				int modifiers = targetClass.getModifiers();
-				
-				/* A protected class (must be a member class) is compiled to a public class as
+			int targetModifiers = targetClass.getModifiers();
+			boolean accessible = false;
+			if (Modifier.isPublic(targetModifiers) || Modifier.isProtected(targetModifiers)) {
```
DanHeidinga: Classes at the VM level are either public or package.  Are there tests that should `protected` should be treated like public?

jcl/src/java.base/share/classes/java/lang/invoke/MethodHandles.java
```diff
@@ -512,33 +512,27 @@ private void checkAccess(Class<?> targetClass, String name, int memberModifiers,
...
+				/*[IF Sidecar19-SE]*/
```
DanHeidinga: Doesn't this refactoring also mean that a Lookup with `NO_ACCESS` can still access public types?

jcl/src/java.base/share/classes/java/lang/invoke/MethodHandles.java
```diff
@@ -512,33 +512,27 @@ private void checkAccess(Class<?> targetClass, String name, int memberModifiers,
...
-				if (Modifier.isPublic(modifiers) || Modifier.isProtected(modifiers)) {
-					/* Already determined that we have more than "no access" (public access) */
-					return;
-				} else {
-					if (((PACKAGE == (accessMode & PACKAGE)) || Modifier.isPrivate(accessMode)) && isSamePackage(accessClass, targetClass)) {
-						return;
-					}
-				}
+				accessible = ((accessClass == targetClass)
+						|| ((PACKAGE == (accessMode & PACKAGE)) && isSamePackage(accessClass, targetClass))
```
DanHeidinga: The code is clearer if this is written as a couple of if statements rather than a set of conditions assigned to the variable.

It also has that floating `);`

jcl/src/java.base/share/classes/java/lang/invoke/MethodHandles.java
pdbain-ibm: Yes.  There is a lot of redundant checking for `NO_ACCESS` in the existing code.  I am 
looking at refactoring it.

DanHeidinga: Let's get the fix for this in before looking at a larger refactoring

jcl/src/java.base/share/classes/java/lang/invoke/MethodHandles.java
pdbain-ibm: There is a JTReg test for Lookup on protected classes: 
https://github.com/eclipse/openj9/issues/3301#issuecomment-435983845.  I suggest we concentrate on closing that, since it fixes a JTReg test, then revisit this.

pdbain-ibm:
Parking this while I update it.

pshipton:
@pdbain-ibm any outlook for this?

pdbain-ibm:
@pshipton It is affected by https://github.com/eclipse/openj9/pull/3592, which I believe is higher priority as it fixes a jtreg failure https://github.com/eclipse/openj9/issues/3301. Would you kindly review #3592?

jcl/src/java.base/share/classes/java/lang/invoke/MethodHandles.java
pdbain-ibm: Good plan.  Will fix.

jcl/src/java.base/share/classes/java/lang/invoke/MethodHandles.java
pdbain-ibm: Okay.

pdbain-ibm:
Verified in a personal build.

genie-openj9:

PR #3502, opened 2018-10-31
AArch64: Implement unsafe helper functions

knn-k:
This commit implements the unsafe helper functions for aarch64.

### DIFF
--- a/runtime/vm/xr64/unsafeHelper.s
+++ b/runtime/vm/xr64/unsafeHelper.s
@@ -1,0 +1,101 @@
+# Copyright (c) 2019, 2019 IBM Corp. and others
+#
+# This program and the accompanying materials are made available under
+# the terms of the Eclipse Public License 2.0 which accompanies this
+# distribution and is available at https://www.eclipse.org/legal/epl-2.0/
+# or the Apache License, Version 2.0 which accompanies this distribution and
+# is available at https://www.apache.org/licenses/LICENSE-2.0.
+#
+# This Source Code may also be made available under the following
+# Secondary Licenses when the conditions for such availability set
+# forth in the Eclipse Public License, v. 2.0 are satisfied: GNU
+# General Public License, version 2 with the GNU Classpath
+# Exception [1] and GNU General Public License, version 2 with the
+# OpenJDK Assembly Exception [2].
+#
+# [1] https://www.gnu.org/software/classpath/license.html
+# [2] http://openjdk.java.net/legal/assembly-exception.html
+#
+# SPDX-License-Identifier: EPL-2.0 OR Apache-2.0 OR GPL-2.0 WITH Classpath-exception-2.0 OR LicenseRef-GPL-2.0 WITH Assembly-exception
+
+	.text
+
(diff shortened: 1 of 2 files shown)


knn-k:
The Travis CI build failed with:
```
DeadTreesElimination.cpp:742:39: error: ‘class TR::ILOpCode’ has no member named ‘isPackedExponentiation’
                if (child->getOpCode().isPackedExponentiation())
```
This is not caused by this commit.

hangshao0:
I guess you also need to add vpath to module.xml for AArch64:
https://github.com/eclipse/openj9/blob/master/runtime/vm/module.xml#L148toL150

runtime/vm/xr64/unsafeHelper.s
```diff
@@ -0,0 +1,92 @@
...
+# Prototype: void unsafePut64(I_64 *address, I_64 value);
+# Defined in: #Args: 2
+	.align	2
+	.global	unsafePut64
+	.type	unsafePut64, %function
+unsafePut64:
+	str	x1, [x0]
+	ret
+	.size	unsafePut64, .-unsafePut64
+# Prototype: void unsafePut32(I_32 *address, I_32 value);
```
DanHeidinga: It's a fairly minor thing but can you add a blank line between the different functions being defined in this file?  It makes it little easier to pick out where one ends and the next begins

hangshao0:
There are also changes required in https://github.com/eclipse/openj9/blob/master/runtime/vm/gphandle.c#L238 that handles SIGBUS from the unsafe helpers.

The idea is that when SIGBUS is triggered from accessing the invalid address in the unsafe helpers, we know the UNSAFE_TARGET_REGISTER (x0 here) has the invalid address, then we update this register to an writable/readable but garbage address on stack, so that we can bypass the offending instruction and turn SIGBUS into an exception. 

If you are not making the above change in this PR, please put some comments in unsafeHelper.s and gphandle.c saying that this needs to be done on AArch64.

knn-k:
@hangshao0 Thanks for your comments.

I will add vpath to module.xml.  What should the spec name be for AArch64?  I would say `linux_aarch64`.  I guess I also have to put a new file in openj9/buildspecs with the same name -- Correct?

runtime/vm/xr64/unsafeHelper.s
knn-k: Sure.

hangshao0:
>What should the spec name be for AArch64? I would say linux_aarch64

I am fine with this name.

> I guess I also have to put a new file in openj9/buildspecs with the same name (I am not going to do that in this PR) -- Correct?

Yes.

DanHeidinga:
Jenkins compile xlinux jdk8

DanHeidinga:
@knn-k can you rebase to resolve the conflict?  I think it should be fairly straightforward

knn-k:
Resolved merge conflict.

knn-k:
Corrected copyright years.

DanHeidinga:
Jenkins compile xlinux jdk8

knn-k:
I would like this to be merged if the code is OK.

PR #3503, opened 2018-10-31
Migrate async check snippet calls to OOL on X86

0dvictor:
Use out-of-line instructions to call async check helper.

### DIFF
--- a/runtime/compiler/x/codegen/J9TreeEvaluator.cpp
+++ b/runtime/compiler/x/codegen/J9TreeEvaluator.cpp
@@ -1194,10 +1194,11 @@ TR::Register *J9::X86::TreeEvaluator::asynccheckEvaluator(TR::Node *node, TR::Co
       {
       TR::TreeEvaluator::asyncGCMapCheckPatching(node, cg, snippetLabel);
       }
-   else if (secondChild->getOpCode().isLoadConst())
+   else
       {
-      TR::MemoryReference *mr = generateX86MemoryReference(compareNode->getFirstChild(), cg);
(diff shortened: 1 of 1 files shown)


andrewcraik:

runtime/compiler/x/codegen/J9TreeEvaluator.cpp
```diff
@@ -1194,10 +1194,11 @@ TR::Register *J9::X86::TreeEvaluator::asynccheckEvaluator(TR::Node *node, TR::Co
       {
       TR::TreeEvaluator::asyncGCMapCheckPatching(node, cg, snippetLabel);
       }
-   else if (secondChild->getOpCode().isLoadConst())
+   else
       {
-      TR::MemoryReference *mr = generateX86MemoryReference(compareNode->getFirstChild(), cg);
+      TR_ASSERT(secondChild->getOpCode().isLoadConst(), "unrecognized asynccheck test: special async check value is not a constant");
```
andrewcraik: Actually - this needs to be fatal - previously we would return NULL without generating the check - now we would try to generate the checking code assuming a const load if the assert were not enabled.

andrewcraik:
Sanity is clean so if the assert just becomes fatal a CI should be fine to merge this.

runtime/compiler/x/codegen/J9TreeEvaluator.cpp
0dvictor: Good point. I've updated it to `TR_ASSERT_FATAL`.

PR #3506, opened 2018-11-01
Refactor code related to method fanin

dmitry-ten:
- Refactor interpreter profiling methods `getNumberofCallersAndTotalWeight`
and `getOtherBucketWeight` into one method - `getFaninInfo`.
These 2 methods are usually called together, so it makes sense
to make them one method. This also reduces the number of times
a hashtable of method samples is searched.

- Move calls that access fanin info from the front-end to resolved method.
This information is only accessed from the inliner, where we always have
a resolved method. This also makes fanin info more easily accessible for
a downstream project.

### DIFF
--- a/runtime/compiler/env/VMJ9.cpp
+++ b/runtime/compiler/env/VMJ9.cpp
@@ -5797,55 +5797,6 @@ TR_J9VMBase::getMaxCallGraphCallCount()
    return profiler->getMaxCallCount();
    }
 
-void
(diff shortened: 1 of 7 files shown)


mpirvu:
I will wait for the performance results before merging this PR.

mpirvu:
jenkins test sanity all jdk8

dmitry-ten:
Performance testing done, throughput is the same.

PR #3507, opened 2018-11-01
Rename evacuate RB type

amicic:
Rename 'evacuate' Read Barrier type to 'range check'. It's more
appropriate since it's discribing the trigger, rather than action done
by RB specific to Concurrent Scavenger. The same trigger could be used
for other GCs or profiling/debug tools with an action that has nothing
with object evacuation.

OMR side of the change: https://github.com/eclipse/omr/pull/3132


dmitripivkine:

amicic:

amicic:

PR #3511, opened 2018-11-01
Add build instructions for OpenJDK11 macOS

SueChaplain:
New section for building OpenJDK 11 with OpenJ9
for macOS.

### DIFF
--- a/doc/build-instructions/Build_Instructions_V11.md
+++ b/doc/build-instructions/Build_Instructions_V11.md
@@ -31,7 +31,7 @@ A more complete set of build instructions are included here for multiple platfor
 - [Linux :penguin:](#linux)
 - [AIX :blue_book:](#aix)
 - [Windows :ledger:](#windows)
-- [MacOS :apple:](#macos)
+- [macOS :apple:](#macos)
 - [ARM :iphone:](#arm)
 
 ----------------------------------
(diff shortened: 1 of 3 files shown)


SueChaplain:
@DanHeidinga 

I based this off some notes that Steve Groeger made but there are a few things before this is ready:

- which boot JDK should I point to (currently pointing to OpenJDK 10 Hotspot page)
- please confirm the names of the directories where the binaries are built
- please can I have a java -version output 

Let me know if anything else needs changing. Thanks

doc/build-instructions/Build_Instructions_V11.md
```diff
@@ -448,11 +448,96 @@ JCL      - a786f96b13 based on jdk-11+21)
...
 :apple:
+The following instructions guide you through the process of building a macOS **OpenJDK V11** binary that contains Eclipse OpenJ9. This process can be used to build binaries for macOS 10.
 
-:construction:
-We haven't created a full build process for macOS yet? Watch this space!
+### 1. Prepare your system
+:apple:
+You must install a number of software dependencies to create a suitable build environment on your system:
+
+- [Xcode 4.6.2](https://xcodereleases.com/), which should be installed into the directory `/Applications/Xcode4/`.
```
nbhuiyan: AFAIK, Xcode 4.6.2 is necessary only for OpenJDK8 builds. JDK9 and up can be build using the latest Xcode versions (9.x, 10.x), which are very easy to set up. Downloading and setting up Xcode 4.6.2 is not as straightforward, and in my experience, does not seem to work with macOS versions newer than macOS 10.11 El Capitan.

SueChaplain: @nbhuiyan  @DanHeidinga Which version do we want to quote in here for OpenJDK 11? I see Xcode 7 mentioned in #36?

DanHeidinga: @SueChaplain #36 is a bit of a mess as it's the ongoing investigation into building JDK8 mixed into the JDK11 builds.

For JDK11, we should match the OpenJDK suggested build environment of `xcode 9.4` and `10.13.5 (Yosemite)` OS version.  See https://wiki.openjdk.java.net/display/Build/Supported+Build+Platforms

As this will ensure that the OpenJDK code compiles successfully as well.

DanHeidinga:
>which boot JDK should I point to (currently pointing to OpenJDK 10 Hotspot page)

Assuming it's an AdoptOpenJDK link, that's correct.  OpenJ9 has mac support on JDK11 and soon JDK8

doc/build-instructions/Build_Instructions_V11.md
```diff
@@ -448,11 +448,96 @@ JCL      - a786f96b13 based on jdk-11+21)
...
+- [macOS OpenJDK 10](https://adoptopenjdk.net/archive.html?variant=openjdk10&jvmVariant=hotspot), which is used as the boot JDK.
+- [nasm 2.13.03](https://formulae.brew.sh/formula/nasm), which can be installed by using [Homebrew](https://brew.sh/).
+- [bash 4.4.23](https://formulae.brew.sh/formula/bash), which can be installed by using [Homebrew](https://brew.sh/).
+- [Freemarker V2.3.8](https://sourceforge.net/projects/freemarker/files/freemarker/2.3.8/freemarker-2.3.8.tar.gz/download), which can be obtained and installed with the following commands:
```
DanHeidinga: I believe we also need `gnu tar` installed through homebrew.

doc/build-instructions/Build_Instructions_V11.md
```diff
@@ -448,11 +448,96 @@ JCL      - a786f96b13 based on jdk-11+21)
...
+:apple:
+Now you're ready to build OpenJDK with OpenJ9:
+
+```
+make all
+```
+
+A binary for the full developer kit (jdk) is built and stored in the following directory:
+
+- **build/macos-x86_64-normal-server-release/images/jdk**
```
DanHeidinga: This is right.

doc/build-instructions/Build_Instructions_V11.md
```diff
@@ -448,11 +448,96 @@ JCL      - a786f96b13 based on jdk-11+21)
...
+```
+Run:
+```
+./bin/java -version
+```
+
+Here is some sample output:
+
+```
+openjdk version "11-internal" 2018-09-25
```
DanHeidinga: java -version from one of my builds:
```
$ ./build/macosx-x86_64-normal-server-release/images/jdk/bin/java -version
openjdk version "11-internal" 2018-09-25
OpenJDK Runtime Environment (build 11-internal+0-adhoc.heidinga.openj9-openjdk-jdk11)
Eclipse OpenJ9 VM (build djh/libjava-72338d7a1, JRE 11 Mac OS X amd64-64-Bit Compressed References 20181104_000000 (JIT enabled, AOT enabled)
OpenJ9   - 72338d7a1
OMR      - d4cd7c31
JCL      - 9da99f8b97 based on jdk-11+28)
```

doc/build-instructions/Build_Instructions_V11.md
SueChaplain: Changing to xcode 9.4. Do we need to tell them where to install xcode? If so, presumably this is different for v9.4

>..  which should be installed into the directory `/Applications/Xcode4/`.

doc/build-instructions/Build_Instructions_V11.md
SueChaplain: Added.

doc/build-instructions/Build_Instructions_V11.md
DanHeidinga: We don't need to tell them where to install it.  The default location is fine.

doc/build-instructions/Build_Instructions_V11.md
SueChaplain: Thanks. Added

doc/build-instructions/Build_Instructions_V11.md
SueChaplain: For now I changed this to /Applications/Xcode9/ .... so please let me know if it needs further adjustment (or removing).

SueChaplain: OK. Removed the directory. Thanks

SueChaplain:
Last question (related) - should I be updating https://www.eclipse.org/openj9/docs/openj9_support/#openjdk-11_1 to change the OS level from 10.9.0 to 10.13.5?

DanHeidinga:
> Last question (related) - should I be updating https://www.eclipse.org/openj9/docs/openj9_support/#openjdk-11_1 to change the OS level from 10.9.0 to 10.13.5?

Yes please.  That's the preferred OS level from OpenJDK and we should match their minimum.

doc/build-instructions/Build_Instructions_V11.md
```diff
@@ -448,11 +448,97 @@ JCL      - a786f96b13 based on jdk-11+21)
...
 :apple:
+The following instructions guide you through the process of building a macOS **OpenJDK V11** binary that contains Eclipse OpenJ9. This process can be used to build binaries for macOS 10.
 
-:construction:
-We haven't created a full build process for macOS yet? Watch this space!
+### 1. Prepare your system
+:apple:
+You must install a number of software dependencies to create a suitable build environment on your system:
+
+- [Xcode 9.4](https://xcodereleases.com/).
```
nbhuiyan: Should we be suggesting downloading from an unofficial website, when an official one exists? https://developer.apple.com/download/more/ (requires an account)

SueChaplain: Ah, certainly not. It was the first one I found. Altering now.

SueChaplain:
@DanHeidinga  Can this be merged please? I'll put the OpenJDK8 instructions in a separate PR when the steps are finalised.

Issue #3500, opened 2018-10-31
JVMTI GetOwnedMonitorInfo inefficient


gacholio:
GetOwnedMonitorInfo is acquiring exclusive VM access before walking the target thread. This should probably be changed to use haltThreadForInspection instead.

PR #3512, opened 2018-11-01
Optimize GetOwnedMonitorInfo

gacholio:
Use haltThreadForInspection instead of exclusive VM access to halt the
target thread.

Add a test to call GetOwnedMonitorInfo repeatedly on a running foreign
thread.

Fixes: #3500


DanHeidinga:
Is there anything in the JVMTI spec that requires exclusive?  

If two threads attempt to inspect the same targetThread, will the releaseVMThread calls conflict in any way?

DanHeidinga:
Jenkins test all plinux jdk8

PR #1234, opened 2018-02-20
Add TODO comment to L-World opcode remap workaround

eric98zhang:
Signed-off-by: Eric Zhang <eric99.zhang@gmail.com>

PR #3517, opened 2018-11-01
Give up optimizing Unsafe CAS on arraylets

liqunl:
Codegen doesn't support inlining Unsafe CAS for arraylets.

### DIFF
--- a/runtime/compiler/optimizer/UnsafeFastPath.cpp
+++ b/runtime/compiler/optimizer/UnsafeFastPath.cpp
@@ -114,10 +114,23 @@ static bool isVarHandleOperationMethodOnArray(TR::RecognizedMethod rm)
    return false;
    }
 
+static bool isVarHandleOperationMethodOnNonStaticField(TR::RecognizedMethod rm)
+   {
+   switch (rm)
+      {
+      case TR::java_lang_invoke_InstanceFieldVarHandle_InstanceFieldVarHandleOperations_OpMethod:
+      case TR::java_lang_invoke_ArrayVarHandle_ArrayVarHandleOperations_OpMethod:
+      case TR::java_lang_invoke_ByteArrayViewVarHandle_ByteArrayViewVarHandleOperations_OpMethod:
+         return true;
+      default:
+         return false;
+      }
+   return false;
+   }
 
 /**
  * \brief
- *    Transform unsafe atomic method called from VarHandle to codegen intrinsic
+ *    Try transform unsafe atomic method called from VarHandle to codegen intrinsic
  *
  * \parm callTree
  *    Tree containing the call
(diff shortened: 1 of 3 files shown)


runtime/compiler/optimizer/UnsafeFastPath.cpp
```diff
@@ -132,6 +132,19 @@ static bool isVarHandleOperationMethodOnArray(TR::RecognizedMethod rm)
...
+   // Give up on arraylet
+   //
+   if (isVarHandleOperationMethodOnArray(callerMethod)
+       && TR::Compiler->om.canGenerateArraylets())
+      {
+      if (trace())
+         {
+         traceMsg(comp(), "Call %p n%dn is accessing an element from an array that might be arraylet, quit\n", node, node->getGlobalIndex());
+         }
+      return;
```
fjeremic: This seems a little odd to me. The function documentation and the function name seems to imply the transformation will always be made, yet we bail out here. Doesn't this code belong to something like `isTransformableUnsafeAtomic`? Otherwise I'd like to see the documentation and the name changed to reflect what is actually happening.

liqunl: Good point. `isTransformableUnsafeAtomic` was first introduced as a public API in j9method, since it lives in UnsafeFastPath now, will make it more tailored for VarHandle.

liqunl: @fjeremic I've updated the transform API as well as the documention, can you continue the review?

runtime/compiler/optimizer/UnsafeFastPath.cpp
```diff
@@ -142,16 +154,24 @@ void TR_UnsafeFastPath::transformUnsafeAtomicCallInVarHandleAccessMethod(TR::Tre
...
          if (trace())
             {
             traceMsg(comp(), "Found Unsafe CAS node %p n%dn on non-static field, set the flag\n", node, node->getGlobalIndex());
             }
+
+         return true;
          }
 
-      return;
+      return false;
```
fjeremic: The code following this point is dead. Was this intentional?

runtime/compiler/optimizer/UnsafeFastPath.cpp
```diff
@@ -142,16 +154,24 @@ void TR_UnsafeFastPath::transformUnsafeAtomicCallInVarHandleAccessMethod(TR::Tre
       {
       // codegen doesn't optimize CAS on a static field
       //
-      if (callerMethod != TR::java_lang_invoke_StaticFieldVarHandle_StaticFieldVarHandleOperations_OpMethod)
+      if (callerMethod != TR::java_lang_invoke_StaticFieldVarHandle_StaticFieldVarHandleOperations_OpMethod &&
```
fjeremic: This seems error prone. How are you certain there won't be new VarHandle methods introduced which operate on static fields? If this happens (and it's certainly not unlikely) this check would pass and we would perform an unsafe operation.

Rather than opt-out for statics, can we instead opt-in for non-statics? i.e. turn the `!=` into a set of `==`? Is this feasible? Do you understand my concern?

runtime/compiler/optimizer/UnsafeFastPath.cpp
```diff
@@ -128,12 +128,24 @@ static bool isVarHandleOperationMethodOnArray(TR::RecognizedMethod rm)
  * \parm calleeMethod
  *    The unsafe method
  *
+ * \return True if the call is transformed, otherwise false
+ *
  */
-void TR_UnsafeFastPath::transformUnsafeAtomicCallInVarHandleAccessMethod(TR::TreeTop* callTree, TR::RecognizedMethod callerMethod, TR::RecognizedMethod calleeMethod)
+bool TR_UnsafeFastPath::transformUnsafeAtomicCallInVarHandleAccessMethod(TR::TreeTop* callTree, TR::RecognizedMethod callerMethod, TR::RecognizedMethod calleeMethod)
```
fjeremic: Not a fan of the function name. It is written in the imperative mood, meaning that it should always succeed, yet it returns a `bool` indicating it can sometimes fail. This function really attempts to perform a transformation. I suggest we dictate that in the name and perhaps rename to `tryTransformUnsafeAtomicCallInVarHandleAccessMethod`?

runtime/compiler/optimizer/UnsafeFastPath.cpp
liqunl: Yes. We don't have intrinsics for Unsafe CAS right now. The flag is a workaround and CAS will be processed as the same way once the CAS atomic symbol is in. For that I'll create a separate PR.

fjeremic: Ok, I'd like to see a TODO left here describing exactly why we have a dead end point here. The TODO should reference an issue (for example `TODO (#1234): Temporarily disable the following code until CAS work is done`) which will solve the problem.

liqunl: Sure, I've created https://github.com/eclipse/openj9/issues/3532 as the follow-up item.

liqunl:
@andrewcraik Can I have a review from you?

runtime/compiler/optimizer/UnsafeFastPath.cpp
andrewcraik: I don't think that is feasible since there are a large number of other VarHandle sub-types. The static case is specific so I think this is reasonable unless @liqunl has some specific idea on how to group 'everything else'.

liqunl: There're not many VarHandle sub-types. The thing is that whoever adds a sub-type has to let the JIT know, otherwise the jitted code will crash. So I think either way is fine.

andrewcraik:
Ok change seems reasonable - can we make sure we have an issue to come back and fix this for arraylets please?

andrewcraik:
Jenkins test sanity xlinux,plinux,win jdk11

liqunl:
@andrewcraik I don't think it's related to me, this commit is to disable a transformation under some gc modes. The transformation is not required by functionality.

liqunl:
https://github.com/eclipse/openj9/issues/3556 is created to make the transformation work for arraylets.

PR #3527, opened 2018-11-02
Make CH table updates more granular

dmitry-ten:
Create a subclass of `TR_PersistentClassInfo` for the client.
Override every method that modifies class info so that it is marked
as dirty/for removal before calling base method.

This enables us to detect changes to persistent CH table at a finer
granularity. We used to mark class as dirty every time class info
is accessed, whether or not it was modified. This is too conservative
and makes CH table updates larger than necessary.

It also makes code cleaner and less prone to bugs, because we don't need
to think about class hierarchy anymore. Since dirty/removed set gets
updated every time a class info is modified, cases like unloading a subclass of
a loaded class are handled automatically, so we don't need
`markSuperClasesAsDirty` method.

### DIFF
--- a/runtime/compiler/env/JITaaSPersistentCHTable.cpp
+++ b/runtime/compiler/env/JITaaSPersistentCHTable.cpp
@@ -388,7 +388,6 @@ TR_JITaaSClientPersistentCHTable::TR_JITaaSClientPersistentCHTable(TR_Persistent
 TR_PersistentClassInfo *
 TR_JITaaSClientPersistentCHTable::findClassInfo(TR_OpaqueClassBlock * classId)
    {
-   markDirty(classId);
    return TR_PersistentCHTable::findClassInfo(classId);
    }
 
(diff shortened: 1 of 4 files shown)


dmitry-ten:
@mpirvu you were right, this change removes most of the updates to CH table.
Total size of updates went down by 85%.
Throughput didn't change.

dmitry-ten:
Rebased and made a pointer to CH table static, so that only one pointer is allocated for all instances of class info.

dmitry-ten:
Functional testing passed the same tests as without this PR,
testing performance.

dmitry-ten:
Throughput seems to be fine, but @mpirvu maybe you want to verify on your own machine.

mpirvu:
Will do.

mpirvu:
Throughput is unchanged after this change.
CPU at the client is reduced by about 4 percentage points which is very nice.

PR #3528, opened 2018-11-02
Return IMMUTABLE IntStream from String.chars()

pdbain-ibm:
Make the resulting IntStream return immutable spliterator for efficiency.

Fixes https://github.com/eclipse/openj9/issues/3202

### DIFF
--- a/jcl/src/java.base/share/classes/java/lang/String.java
+++ b/jcl/src/java.base/share/classes/java/lang/String.java
@@ -33,6 +33,9 @@ import java.util.Formatter;
 import java.util.StringJoiner;
 import java.util.Iterator;
 import java.nio.charset.Charset;
+import java.util.Spliterator;
+import java.util.Spliterators;
+import java.util.stream.StreamSupport;
 
 /*[IF Sidecar19-SE]*/
 import jdk.internal.misc.Unsafe;
@@ -3876,8 +3879,15 @@ public final class String implements Serializable, Comparable<String>, CharSeque
 	
 	@Override
 	public IntStream chars() {
-		/* Following generic CharSequence method invoking need to be updated with optimized implementation specifically for this class */
-		return CharSequence.super.chars();
+		final int len = lengthInternal();
+		char charBuff[] = new char[len];
+		getCharsNoBoundChecks(0, len, charBuff, 0);
+		int intBuff[] = new int[len];
+		for (int i = 0; i < len; ++i) {
+			intBuff[i] = charBuff[i];
+		}
+		return StreamSupport.intStream(Spliterators.spliterator(intBuff,Spliterator.IMMUTABLE | Spliterator.ORDERED),
+				false);
 	}
 	
 	@Override
--- a/jcl/src/java.base/share/classes/java/lang/String.java
+++ b/jcl/src/java.base/share/classes/java/lang/String.java
@@ -31,10 +31,12 @@ import java.util.regex.Pattern;
 import java.util.regex.PatternSyntaxException;
 import java.util.Formatter;
 import java.util.StringJoiner;
+import java.util.function.IntConsumer;
+import java.util.function.IntFunction;
+import java.util.function.IntUnaryOperator;
 import java.util.Iterator;
 import java.nio.charset.Charset;
 import java.util.Spliterator;
-import java.util.Spliterators;
 import java.util.stream.StreamSupport;
 
 /*[IF Sidecar19-SE]*/
@@ -3877,23 +3879,60 @@ public final class String implements Serializable, Comparable<String>, CharSeque
 		throw newStringIndexOutOfBoundsException(begin, end, length);
 	}
 	
+	private static final class StringSpliteratorOfInt implements Spliterator.OfInt {
+		private final int start; // index of first character in the array
+		private int end; // index after the final character
+		private int cursor; // next character to read
+		private final IntUnaryOperator func;
+		
+		StringSpliteratorOfInt(int start, int end, IntUnaryOperator func) {
+			this.start = start;
+			this.end = end;
+			this.func = func;
+			cursor = start;
+		}
+
+		@Override
+		public long estimateSize() {
+			return end - start;
+		}
+
+		@Override
+		public int characteristics() {
+			return IMMUTABLE | ORDERED | SIZED | SUBSIZED;
+		}
+
+		@Override
+		public OfInt trySplit() {
+			int newEnd = cursor + (end - cursor + 1) / 2;
+			StringSpliteratorOfInt newSpliterator = null;
+			if (newEnd < end) {
+				newSpliterator = new StringSpliteratorOfInt(newEnd, end, func);
+				end = newEnd;
+			}
+			return newSpliterator;
+		}
+
+		@Override
+		public boolean tryAdvance(IntConsumer action) {
+			boolean result = false;
+			if (cursor < end) {
+				action.accept(func.applyAsInt(cursor));
+				result = true;
+				++cursor;
+			}
+			return result;
+		}
+	}
+	
 	@Override
 	public IntStream chars() {
-		final int len = lengthInternal();
-		char charBuff[] = new char[len];
-		getCharsNoBoundChecks(0, len, charBuff, 0);
-		int intBuff[] = new int[len];
-		for (int i = 0; i < len; ++i) {
-			intBuff[i] = charBuff[i];
-		}
-		return StreamSupport.intStream(Spliterators.spliterator(intBuff,Spliterator.IMMUTABLE | Spliterator.ORDERED),
-				false);
+		return StreamSupport.intStream(new StringSpliteratorOfInt(0, lengthInternal(), i -> charAt(i)), false);
 	}
 	
 	@Override
 	public IntStream codePoints() {
-		/* Following generic CharSequence method invoking need to be updated with optimized implementation specifically for this class */
-		return CharSequence.super.codePoints();
+		return StreamSupport.intStream(new StringSpliteratorOfInt(0, lengthInternal(), i -> codePointAt(i)), false);
 	}
 
 	/*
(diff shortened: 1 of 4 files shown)


pdbain-ibm:
@keithc-ca would you kindly review this?
Thank you.

jcl/src/java.base/share/classes/java/lang/String.java
```diff
@@ -3876,8 +3879,15 @@ static void checkBoundsBeginEnd(int begin, int end, int length) {
...
-		return CharSequence.super.chars();
+		final int len = lengthInternal();
+		char charBuff[] = new char[len];
+		getCharsNoBoundChecks(0, len, charBuff, 0);
+		int intBuff[] = new int[len];
+		for (int i = 0; i < len; ++i) {
+			intBuff[i] = charBuff[i];
+		}
+		return StreamSupport.intStream(Spliterators.spliterator(intBuff,Spliterator.IMMUTABLE | Spliterator.ORDERED),
+				false);
```
keithc-ca: Rather than copying the characters, we could return a stream based on either a byte or char spliterator which access the underlying data. As is, an ASCII string requires 6 times additional space and a lot of copying before returning.

pdbain-ibm: Keith
Thanks for the idea. I will rework per your suggestion and add some unit tests.
Marking this WIP.

jcl/src/java.base/share/classes/java/lang/String.java
```diff
@@ -3874,16 +3878,60 @@ static void checkBoundsBeginEnd(int begin, int end, int length) {
 		throw newStringIndexOutOfBoundsException(begin, end, length);
 	}
 	
+	class StringSpliteratorOfInt implements Spliterator.OfInt {
+		int start; // index of first character in the array
+		int end; // index after the final character
+		int cursor; // next character to read
+		IntFunction func;
```
keithc-ca: This class should be `private static final` and all its fields should be `private final`.
Using `IntUnaryOperator` will avoid the boxing/unboxing implied via `IntFunction`.

jcl/src/java.base/share/classes/java/lang/String.java
```diff
@@ -3874,16 +3878,60 @@ static void checkBoundsBeginEnd(int begin, int end, int length) {
...
+		}
+
+		@Override
+		public int characteristics() {
+			return IMMUTABLE | ORDERED | SIZED | SUBSIZED;
+		}
+
+		@Override
+		public OfInt trySplit() {
+			int newEnd = (cursor + end + 1)/2;
```
keithc-ca: This may overflow for large strings, `cursor + ((end - cursor + 1) / 2)` will not.

test/functional/Java8andUp/playlist.xml
```diff
@@ -1936,6 +1936,30 @@
...
+	$(TEST_STATUS)</command>
+		<levels>
+			<level>sanity</level>
+		</levels>
+		<groups>
+			<group>functional</group>
+		</groups>
+		<subsets>
+			<subset>8</subset>
+			<subset>11</subset>
```
keithc-ca: Why is this only enabled for 8 and 11? It should be for all versions (i.e. 8+) so it doesn't need to be change for Java 12, for example.

test/functional/Java8andUp/src/org/openj9/test/string/StringStreams.java
```diff
@@ -0,0 +1,168 @@
...
+	public void testStringCharsCharacteristics() {
+		Spliterator<?> s = DIGITS.chars().spliterator();
+		int expectedCharacteristics = Spliterator.ORDERED | Spliterator.SIZED | Spliterator.SUBSIZED;
+		if (VersionCheck.major() >= 11 ) {  /* IMMUTABLE not set in reference implementation Java 8 */
+			expectedCharacteristics |= Spliterator.IMMUTABLE; 
+		}
+		int characteristics = s.characteristics();
+		assertTrue((characteristics & expectedCharacteristics) == expectedCharacteristics,
+				"chars() spliterator characteristics 0x" + Integer.toHexString(characteristics)  //$NON-NLS-1$
+				+ "\ndid not have expected characteristics 0x"+Integer.toHexString(expectedCharacteristics)); //$NON-NLS-1$
```
keithc-ca: formatting nit: spaces around binary operators

test/functional/Java8andUp/src/org/openj9/test/string/StringStreams.java
```diff
@@ -0,0 +1,168 @@
...
+	}
+
+	@Test
+	public void testStringCodePointssCharacteristics() {
```
keithc-ca: Typo? `...Pointss...`

test/functional/Java8andUp/src/org/openj9/test/string/StringStreams.java
```diff
@@ -0,0 +1,168 @@
...
+		Spliterator.OfInt s = EMPTY.chars().spliterator();
+		final ArrayList<Integer> actualIntegers = new ArrayList<>();
+		s.forEachRemaining((IntConsumer) i -> actualIntegers.add(Integer.valueOf(i)));
+		checkDigitsOrdered(EMPTY, actualIntegers);
+	}
+
+	@Test
+	public void testStringCharsSpliteratorSplitting() {
+		OfInt s = DIGITS.chars().spliterator();
+		ArrayList<OfInt> sArray = new ArrayList<Spliterator.OfInt>();
```
keithc-ca: Redundant type parameter in `ArrayList<Spliterator.OfInt>`.

test/functional/Java8andUp/src/org/openj9/test/string/StringStreams.java
pdbain-ibm: Yup.  As indicated, this is a work in progress.

pdbain-ibm:
Personal build had some failures.  Running a new build, will advise when this is ready for review.

test/functional/Java8andUp/playlist.xml
pdbain-ibm: In that case, I will remove the `<subsets>` element.

pdbain-ibm:
@keithc-ca and @smlambert or @llxia would you kindly review?
Thank you.

jcl/src/java.base/share/classes/java/lang/String.java
```diff
@@ -3874,16 +3879,60 @@ static void checkBoundsBeginEnd(int begin, int end, int length) {
 		throw newStringIndexOutOfBoundsException(begin, end, length);
 	}
 	
+	private static final class StringSpliteratorOfInt implements Spliterator.OfInt {
+		private int start; // index of first character in the array
+		private int end; // index after the final character
+		private int cursor; // next character to read
+		IntUnaryOperator func;
```
keithc-ca: Fields `start` and `func` should be `final`; `func` should also be `private`.

jcl/src/java.base/share/classes/java/lang/String.java
```diff
@@ -3874,16 +3879,60 @@ static void checkBoundsBeginEnd(int begin, int end, int length) {
...
+		}
+
+		@Override
+		public int characteristics() {
+			return IMMUTABLE | ORDERED | SIZED | SUBSIZED;
+		}
+
+		@Override
+		public OfInt trySplit() {
+			int newEnd = cursor + (end - cursor + 1)/2;
```
keithc-ca: Formatting: spaces around binary operators.

jcl/src/java.base/share/classes/java/lang/String.java
```diff
@@ -3874,16 +3879,60 @@ static void checkBoundsBeginEnd(int begin, int end, int length) {
...
+				end = newEnd;
+			}
+			return newSpliterator;
+		}
+
+		@Override
+		public boolean tryAdvance(IntConsumer action) {
+			boolean result = false;
+			if (cursor < end) {
+				action.accept((int) func.applyAsInt(cursor));
```
keithc-ca: The cast here should be redundant.

test/functional/Java8andUp/src/org/openj9/test/string/StringStreams.java
```diff
@@ -0,0 +1,161 @@
...
+		checkDigits(actualIntegers);
+	}
+
+	/**
+	 * Consume a bit of the spliterator, then split it.
+	 */
+	@Test
+	public void testStringCharsSpliteratorSplittingAfterIterating() {
+		OfInt s = DIGITS.chars().spliterator();
+		ArrayList<OfInt> sArray = new ArrayList<Spliterator.OfInt>();
```
keithc-ca: Please remove redundant type arguments (for consistency).

test/functional/Java8andUp/src/org/openj9/test/string/StringStreams.java
```diff
@@ -0,0 +1,161 @@
...
+			sArray.add(s);
+			s = s.trySplit();
+		}
+		for (OfInt e: sArray) {
+			e.forEachRemaining(consumer);
+		}
+		checkDigits(actualIntegers);
+	}
+
+	private void checkDigitsOrdered(String expectedString, ArrayList<Integer> actualIntegers) {
```
keithc-ca: This method can be `static`; the second parameter should be more general: `List<Integer>`.

test/functional/Java8andUp/src/org/openj9/test/string/StringStreams.java
```diff
@@ -0,0 +1,161 @@
...
+		for (int i = 0; i < expectedString.length(); ++i) {
+			assertEquals(actualIntegers.get(i), Integer.valueOf(expectedString.charAt(i)), 
+					"Wrong value at position " + i); //$NON-NLS-1$
+		}
+	}
+	
+	private void checkDigits(Set<Integer> actualIntegers) {
```
keithc-ca: This method can be `static`.

pdbain-ibm:
Do the reviewers have any other comments?

keithc-ca:
Jenkins test sanity zlinux jdk8,jdk11

Issue #3392, opened 2018-10-22
Using "../java" in system.c when looking for a shared library doesn't work on zOS


groeges:
A change went in recently: https://github.com/eclipse/openj9/pull/3176 to fix an issue with MacOS but this fix has broken the zOS builds. See issue: https://github.ibm.com/runtimes/openjdk-contribution/issues/631 for details of issue.   This needs to be resolved in order to get the zOS builds working again.

tajila:
Ill take a look at this

groeges:
@tajila Any progress on this?

tajila:
I have a PR https://github.com/eclipse/omr/pull/3148

pshipton:
Does the original change to use `../` only work because it fails to find the library at that path and falls back to search the portlib directory?

tajila:
@pshipton yes

pshipton:
I think I'm still missing something. If "../java" always fails on Mac, and the fallback makes it work, wouldn't just using "java" have been sufficient on the Mac, as it could also fail and then use the fallback?

tajila:
I'm not sure what happens on mac, I never tried it there. But on xa64 `../java` only works because of the fallback.

pshipton:
Perhaps I misunderstood. Does the fallback still use the `../java` path, but from the portlib directory?

DanHeidinga:
> Does the fallback still use the ../java path, but from the portlib directory?

Yes. it makes the path relative to the portlib directory as libjava.so is up one level from the compressed | default directories

pshipton:
#3535 was reverted
#3542 fixes the problem.

PR #3535, opened 2018-11-03

DanHeidinga:

issue: #3392


gacholio:

gacholio:
Reverted due to compile errors on non-Linux platforms.

DanHeidinga:
```
ERROR CCN3045 ./../common/system.c:416   Undeclared identifier PATH_MAX.
ERROR CCN4324 ./../common/system.c:416   Array size must have integer type.
```

gacholio:
Probably want to use `EsMaxPath` instead of `PATH_MAX`.

DanHeidinga:
testing using `EsMaxPath` out now on my mac

PR #3538, opened 2018-11-04
Defer acquiring heap walk capability until necessary

gacholio:
Rather than enabling the heap walk GC capability when can_tag_objects is
acquired, defer it until the first use of the heap walk APIs. This
allows object tags to be used without the GC time penalty if the heap
walk APIs are avoided.

Also remove the last vestiges of autotagging (GC and most JVMTI support
was removed some time ago).

[ci skip]

### DIFF
--- a/runtime/jvmti/jvmtiCapability.c
+++ b/runtime/jvmti/jvmtiCapability.c
@@ -302,6 +302,7 @@ jvmtiAddCapabilities(jvmtiEnv* env,
 	jvmtiCapabilities newCapabilities;
 	UDATA i;
 	jvmtiError rc = JVMTI_ERROR_NOT_AVAILABLE;
+	J9VMThread *currentThread = NULL;
 
 	Trc_JVMTI_jvmtiAddCapabilities_Entry(env);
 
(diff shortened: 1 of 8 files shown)


gacholio:
jenkins test all xlinux jdk8

runtime/jvmti/jvmtiCapability.c
```diff
@@ -302,126 +302,97 @@ jvmtiAddCapabilities(jvmtiEnv* env,
...
-				}
-			}
-			vmFuncs->releaseExclusiveVMAccess(currentThread);
-			vmFuncs->internalExitVMToJNI(currentThread);
-		}
-	}
+	rc = getCurrentVMThread(vm, &currentThread);
+	if (rc == JVMTI_ERROR_NONE) {
+		rc = JVMTI_ERROR_NOT_AVAILABLE;
+		vm->internalVMFunctions->internalEnterVMFromJNI(currentThread);
```
DanHeidinga: Does this need vm access?

runtime/jvmti/jvmtiHelpers.c
```diff
@@ -1697,3 +1697,23 @@ findDecompileInfo(J9VMThread *currentThread, J9VMThread *targetThread, UDATA dep
 	currentThread->javaVM->walkStackFrames(currentThread, walkState);
 	return (UDATA)walkState->userData1;
 }
+
+void
+ensureHeapWalkable(J9VMThread *currentThread)
+{
+	J9JavaVM *vm = currentThread->javaVM;
+	/* If heap walk is already enabled, nothing need be done */
+	if (J9_ARE_NO_BITS_SET(vm->requiredDebugAttributes, J9VM_DEBUG_ATTRIBUTE_ALLOW_USER_HEAP_WALK)) {
```
DanHeidinga: Can we add an assert that current thread must have exclusive?

runtime/jvmti/jvmtiHeap10.c
```diff
@@ -668,17 +649,15 @@ wrapObjectIterationCallback(j9object_t *slotPtr, j9object_t referrer, void *user
 
 		/* Update the tag only if not autotagging or if it is a class, i.e., if the object is taggable */
 
-		if ( OBJECT_IS_TAGGABLE(iteratorData->env, vm, object) ) {
```
DanHeidinga: `OBJECT_IS_TAGGABLE` also checked `J9VM_IS_INITIALIZED_HEAPCLASS_VM` - still required?

runtime/jvmti/jvmtiHeap10.c
```diff
@@ -568,28 +562,26 @@ wrapHeapIterationCallback(J9JavaVM * vm, J9MM_IterateObjectDescriptor *objectDes
 					iteratorData->userData);
 
 			/* Update the tag only if not autotagging or if it is a class, i.e., if the object is taggable */
-			if ( OBJECT_IS_TAGGABLE(iteratorData->env, vm, object) ) {
```
DanHeidinga: I see use of `autotagging` in some of the comments in this file.  Can you nuke them as well since you're removing autotagging?

runtime/jvmti/jvmtiCapability.c
gacholio: VM access was added here as a lock for requiredDebugAttributes - it's read/written by the ensureHeapWalkable code under exclusive. There are other places in the code which update this field without any protection, but they're all startup-related.

runtime/jvmti/jvmtiHeap10.c
gacholio: The macro checked !autotag || initializedClass - since autotag is always false, all objects are taggable.

runtime/jvmti/jvmtiHeap10.c
gacholio: Done.

runtime/jvmti/jvmtiHelpers.c
gacholio: Done.

runtime/jvmti/jvmtiHeap10.c
gacholio: "!a || b" doesn't even evaluate b if a is false.

runtime/jvmti/jvmtiHeap.c
```diff
@@ -184,12 +184,6 @@ jvmtiGetTag(jvmtiEnv* env,
 		ENSURE_JOBJECT_NON_NULL(object);
 		ENSURE_NON_NULL(tag_ptr);
 
-		if (AUTOTAGGING_OBJECTS(env)) {
-			if ( !J9VM_IS_INITIALIZED_HEAPCLASS(currentThread, (*(j9object_t *)object)) ) {
```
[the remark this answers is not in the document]
gacholio: No - this was a hack to treat Class objects like normal objects for object tagging when jprof autotagging was on.

gacholio:
re: initialized classes - there's no requirement in the spec for this. It's just a hack for jprof which is completely unnecessary now.

re: vm access - I don't think this implies an existing bug (though there may well be some due to the largely unprotected access to requiredDebugAttributes). The capability code also acquires a mutex while fiddling with this stuff, so it was protected from concurrent capability acquisition. Having moved the check code into the exclusive path of the heap APIs, I now want to stop the field being modified by the capability code while I'm checking the bit in the heap walk code.

DanHeidinga:

PR #3541, opened 2018-11-05
WIP: Enable CPU info for z/os

irisbaron:
This change enable Process and System cpu info. 
The Process cpu relies on existing code. And the System cpu info was added.
The process cpu mechanism retrieve the info as cpu time, which is the expected data in the current framework for cpu calculation.
However the system info retrieve percentage, and thus required an additional field in the SystemInfo object to pass this value back to the mbean and return it as the final answer.
This change relies on a prior OMR change.


jcl/src/jdk.management/share/classes/com/ibm/lang/management/internal/CpuUtilizationHelper.java
```diff
@@ -64,7 +64,12 @@ synchronized double getSystemCpuLoad() {
 
 		/* calculate using the most recent value in the history */
 		if ((latestTime.getTimestamp() - interimTime.getTimestamp()) >= MINIMUM_INTERVAL) {
-			cpuLoad = calculateCpuLoad(latestTime, interimTime);
+	            if (latestTime.getCpuLoad() == -1) {
```
DanHeidinga: For consistency, please use tabs rather than spaces throughout all JCL and VM code

DanHeidinga:
> This change relies on a prior OMR change.

Can you indicate which OMR change this depends on?

jcl/src/jdk.management/share/classes/com/ibm/lang/management/internal/CpuUtilizationHelper.java
```diff
@@ -79,7 +84,13 @@ synchronized double getSystemCpuLoad() {
 		}
 
 		if ((latestTime.getTimestamp() - oldestTime.getTimestamp()) >= MINIMUM_INTERVAL) {
-			cpuLoad = calculateCpuLoad(latestTime, oldestTime);
+		
+		    if (latestTime.getCpuLoad() == -1) {
```
DanHeidinga: Instead of duplicating this logic, can you add it to the `calculateCpuLoad()` method?  That way all future callers will get the right behaviour without having to code this extra check themselves

runtime/jcl/common/mgmtos.c
```diff
@@ -218,14 +218,14 @@ Java_com_ibm_lang_management_internal_SysinfoCpuTime_getCpuUtilizationImpl(JNIEn
...
-		tempMethod = (*env)->GetMethodID(env, clazz, "<init>", "(JJII)V");
+		tempMethod = (*env)->GetMethodID(env, clazz, "<init>", "(JJIII)V");
 		if (NULL == tempMethod) {
 			return NULL;
 		}
 
 		JCL_CACHE_SET(env, MID_com_ibm_lang_management_SysinfoCpuTime_getCpuUtilization_init, tempMethod);
 	}
-
+printf("Found tempMethod - getting time now\n");fflush(stdout);
```
DanHeidinga: There's a printf here and in other places that needs to be removed.

runtime/jcl/common/mgmtos.c
```diff
@@ -245,8 +245,14 @@ Java_com_ibm_lang_management_internal_SysinfoCpuTime_getCpuUtilizationImpl(JNIEn
 			break;
 		}
 	}
+#if defined(J9ZOS390)
+	result = (*env)->NewObject(env, clazz, tempMethod,
+			cpuTime.timestamp, cpuTime.cpuTime , cpuTime.numberOfCpus, myStatus, cpuTime.cpuLoad);
```
DanHeidinga: What is the default value for `cpuTime.cpuLoad` from the OMR code?  Is it `-1` when not supported?  Is the `#ifdef` really required?

test/functional/JLM_Tests/src/org/openj9/test/management/TestGetSystemCpuLoad.java
```diff
@@ -46,7 +46,7 @@
 	@BeforeMethod
 	protected void setUp() throws Exception {
 		String osName = System.getProperty("os.name");
-		if ((null == osName) || osName.equalsIgnoreCase("z/OS")) {
+		if (null == osName) {
```
DanHeidinga: This whole method can be removed as it no longer serves any purpose - every platform should have an non-null `os.name`

fjeremic:
@irisbaron thanks for your contribution. There seems to be quite a few commits with the same name here. Can we squash it down to a single commit?

jcl/src/jdk.management/share/classes/com/ibm/lang/management/internal/CpuUtilizationHelper.java
```diff
@@ -64,7 +64,12 @@ synchronized double getSystemCpuLoad() {
 
 		/* calculate using the most recent value in the history */
 		if ((latestTime.getTimestamp() - interimTime.getTimestamp()) >= MINIMUM_INTERVAL) {
-			cpuLoad = calculateCpuLoad(latestTime, interimTime);
+	            if (latestTime.getCpuLoad() == -1) {
+		        cpuLoad = calculateCpuLoad(latestTime, interimTime);
+                    } else {
+		        cpuLoad = (double)latestTime.getCpuLoad()/100;
```
fjeremic: Spaces needed around mathematical operators for readability. Similarly elsewhere in this PR.

jcl/src/jdk.management/share/classes/com/ibm/lang/management/internal/SysinfoCpuTime.java
```diff
@@ -29,19 +29,23 @@
 	private int numberOfCpus; /* should be at least 1 */
 	private int status;
 	private long processCpuTime; /* in ns. The process Cpu Time, used for getProcessCpuLoad() */
-
+	private int cpuLoad;
```
fjeremic: Can we get some Javadoc documentation here? It is certanly not obvious why we have both `cpuLoad` and `processCpuTime` and when one should use one vs. the other. This would be a great place to mention the meaning of the `-1` `cpuLoad` value.

jcl/src/jdk.management/share/classes/com/ibm/lang/management/internal/SysinfoCpuTime.java
```diff
@@ -29,19 +29,23 @@
...
 	 */
-	SysinfoCpuTime(long timestamp, long cpuTime, int numberOfCpus, int status) {
+	SysinfoCpuTime(long timestamp, long cpuTime, int numberOfCpus, int status, int cpuLoad) {
 		super();
 		this.timestamp = timestamp;
 		this.cpuTime = cpuTime;
 		this.numberOfCpus = numberOfCpus;
 		this.status = status;
+		this.cpuLoad = cpuLoad;
+
```
fjeremic: Extra space here.

jcl/src/jdk.management/share/classes/com/ibm/lang/management/internal/SysinfoCpuTime.java
```diff
@@ -94,5 +98,13 @@ void setProcessCpuTime(long processCpuTime) {
 	long getProcessCpuTime() {
 		return processCpuTime;
 	}
+	
+	/**
+         * @return cumulative CPU time in precentage (<1)
```
fjeremic: What does "`<1`" here mean? Also the method is named `getCpuLoad`, so is it time or load (the description says "time", but it doesn't reflect what is returned).

jcl/src/jdk.management/share/classes/com/ibm/lang/management/internal/CpuUtilizationHelper.java
irisbaron: Changes applied in commit ffdf65f8be493319c906b24c60b752782364cb35

jcl/src/jdk.management/share/classes/com/ibm/lang/management/internal/CpuUtilizationHelper.java
irisbaron: Comment addressed in  ffdf65f8be493319c906b24c60b752782364cb35

runtime/jcl/common/mgmtos.c
irisbaron: Fix applied in ffdf65f8be493319c906b24c60b752782364cb35

runtime/jcl/common/mgmtos.c
irisbaron: Made the change to support cpuLoad for all platforms. Value is initialized to -1 in OMR and updated with real cpuLoad only for zos.  #ifdefs removed. 
Change applied in ffdf65f8be493319c906b24c60b752782364cb35 and the corresponding change in OMR.

test/functional/JLM_Tests/src/org/openj9/test/management/TestGetSystemCpuLoad.java
irisbaron: Change applied in ffdf65f8be493319c906b24c60b752782364cb35

jcl/src/jdk.management/share/classes/com/ibm/lang/management/internal/CpuUtilizationHelper.java
irisbaron: Change applied in ffdf65f8be493319c906b24c60b752782364cb35

jcl/src/jdk.management/share/classes/com/ibm/lang/management/internal/SysinfoCpuTime.java
irisbaron: Fixed in ffdf65f8be493319c906b24c60b752782364cb35

jcl/src/jdk.management/share/classes/com/ibm/lang/management/internal/SysinfoCpuTime.java
irisbaron: Change applied in 5edae7db32e8a09931cf979411bb18bf46ad95b2

jcl/src/jdk.management/share/classes/com/ibm/lang/management/internal/SysinfoCpuTime.java
irisbaron: Fix in 5edae7db32e8a09931cf979411bb18bf46ad95b2

DanHeidinga:
@irisbaron Can you provide a link to the OMR PR?

It's important to ensure we don't merge things before the OMR change is available and it helps with reviewing the whole change.

irisbaron:
> @irisbaron Can you provide a link to the OMR PR?
> 
> It's important to ensure we don't merge things before the OMR change is available and it helps with reviewing the whole change.

@DanHeidinga 
OMR PR: https://github.com/eclipse/omr/pull/3152

DanHeidinga:
Thanks @irisbaron 

I've marked this WIP until the conversation in the OMR pr is resolved as it will affect the final design here.

irisbaron:
Closing this openj9 as the design for the omr dependency changed.

PR #3542, opened 2018-11-05

DanHeidinga:

issue: #3392


groeges:
@DanHeidinga Have tried this fix on my zOS box as well as my Mac OS and Linux box and all seem to build and work OK, so I think we should be OK with this fix, as well as the omr fix from Tobi.

PR #3547, opened 2018-11-05
Check the j9str_printf fit within the allowed buffer

DanHeidinga:
Add an assert for this rather than a dynamic allocation as the likelihood
of ever needing to fallback to the allocation is so low the extra code
isn't worth writing (given it will never be tested)

### DIFF
--- a/runtime/jcl/common/system.c
+++ b/runtime/jcl/common/system.c
@@ -414,9 +414,12 @@ jstring getEncoding(JNIEnv *env, jint encodingType)
 			UDATA handle = 0;
 			J9JavaVM * const vm = ((J9VMThread*)env)->javaVM;
 			char dllPath[EsMaxPath];
+			UDATA written = 0;
 			PORT_ACCESS_FROM_ENV(env);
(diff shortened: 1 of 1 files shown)


pshipton:
Given the travis build succeeded and ran -version, I don't think any more testing is required.

runtime/jcl/common/system.c
```diff
@@ -414,9 +414,12 @@ jstring getEncoding(JNIEnv *env, jint encodingType)
 			UDATA handle = 0;
 			J9JavaVM * const vm = ((J9VMThread*)env)->javaVM;
 			char dllPath[EsMaxPath];
+			UDATA written = 0;
 			PORT_ACCESS_FROM_ENV(env);
 			/* libjava.[so|dylib] is in the jdk/lib/ directory, one level up from the default/ & compressedrefs/ directories */
```
pshipton: btw this comment only applies to jdk11 and not jdk8, where the libjava.so is in the `lib/<arch>` directory

pshipton: and since there is a `#if defined(J9VM_JCL_SE11)` just above, this is appropriate :)

PR #3557, opened 2018-11-05
Remove unnecessary type cast on X86

0dvictor:
In this case, casting TR::Instruction* to TR::X86RegInstruction*
is unnecessary, removing.


andrewcraik:

PR #3559, opened 2018-11-06
AArch64: Implement arm64cinterp.m4 and arm64helpers.m4

knn-k:
This commit adds two asm files for aarch64, implementing c_cInterpreter.

### DIFF
--- a/runtime/vm/arm64cinterp.m4
+++ b/runtime/vm/arm64cinterp.m4
@@ -1,0 +1,70 @@
+dnl Copyright (c) 2019, 2019 IBM Corp. and others
+dnl
+dnl This program and the accompanying materials are made available under
+dnl the terms of the Eclipse Public License 2.0 which accompanies this
+dnl distribution and is available at https://www.eclipse.org/legal/epl-2.0/
+dnl or the Apache License, Version 2.0 which accompanies this distribution and
+dnl is available at https://www.apache.org/licenses/LICENSE-2.0.
+dnl
+dnl This Source Code may also be made available under the following
+dnl Secondary Licenses when the conditions for such availability set
+dnl forth in the Eclipse Public License, v. 2.0 are satisfied: GNU
+dnl General Public License, version 2 with the GNU Classpath
+dnl Exception [1] and GNU General Public License, version 2 with the
+dnl OpenJDK Assembly Exception [2].
+dnl
+dnl [1] https://www.gnu.org/software/classpath/license.html
+dnl [2] http://openjdk.java.net/legal/assembly-exception.html
+dnl
+dnl SPDX-License-Identifier: EPL-2.0 OR Apache-2.0 OR GPL-2.0 WITH Classpath-exception-2.0 OR LicenseRef-GPL-2.0 WITH Assembly-exception
+
+include(arm64helpers.m4)
+
+	.file "arm64cinterp.s"
+
+DECLARE_PUBLIC(cInterpreter)
+
+START_PROC(c_cInterpreter)
+	sub sp, sp, {#}CINTERP_STACK_SIZE
+	stp x19, x20, [sp, GPR_SAVE_OFFSET(19)]
+	stp x21, x22, [sp, GPR_SAVE_OFFSET(21)]
+	stp x23, x24, [sp, GPR_SAVE_OFFSET(23)]
+	stp x25, x26, [sp, GPR_SAVE_OFFSET(25)]
+	stp x27, x28, [sp, GPR_SAVE_OFFSET(27)]
+	stp x29, x30, [sp, GPR_SAVE_OFFSET(29)]
+	stp d8, d9, [sp, FPR_SAVE_OFFSET(8)]
+	stp d10, d11, [sp, FPR_SAVE_OFFSET(10)]
+	stp d12, d13, [sp, FPR_SAVE_OFFSET(12)]
+	stp d14, d15, [sp, FPR_SAVE_OFFSET(14)]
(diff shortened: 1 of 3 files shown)


knn-k:
Should the file names be arm64* or aarch64*?

pshipton:
@gacholio @DanHeidinga ^ do you have a preference?

gacholio:
I'd prefer this not be in another file - will it not be 99% similar to the 32-bit implementation? I was hoping not to generate another branch of files for 64-bit (we haven't needed to do it on any other platforms).

I'd prefer to see ifdefs on our existing 64-bit flag (the poorly-name ENV_DATA64) in ARM files rather than calling out AARCH64 as an entirely different entity.

@knn-k See if you can fix/ifdef the existing m4 files.

gacholio:
Can this difference be resolved by way of alias (in much the same way that _rbx on X is EBX or RBX on different bitness)? 99% of all of the instructions in the m4 files are fullword-sized.

knn-k:
@gacholio It won't be simple instruction/register aliases.  I guess I will have to define more macros to put both 32-bit and 64-bit ARM.

From the existing armcinterp.m4:
```
START_PROC(c_cInterpreter)
	sub r13,r13,{#}CINTERP_STACK_SIZE
	stmia r13,{{ r4,r5,r6,r7,r8,r9,r10,r11,r14 }}
	add r3,r13,{#}FPR_SAVE_OFFSET(0)
	vstmia.64 r3,{{ D8-D15 }}
	mov J9VMTHREAD,r0
	ldr r4,[J9VMTHREAD,{#}J9TR_VMThread_entryLocalStorage]	
	add r3,r13,{#}JIT_GPR_SAVE_OFFSET(0)
	str r3,[r4,{#}J9TR_ELS_jitGlobalStorageBase]
```

AArch64 version (in preparation):
```
START_PROC(c_cInterpreter)
	sub sp, sp, {#}CINTERP_STACK_SIZE
	stp x29, x30, [sp, GPR_SAVE_OFFSET(29)]
	stp x27, x28, [sp, GPR_SAVE_OFFSET(27)]
	stp x25, x26, [sp, GPR_SAVE_OFFSET(25)]
	stp x23, x24, [sp, GPR_SAVE_OFFSET(23)]
	stp x21, x22, [sp, GPR_SAVE_OFFSET(21)]
	stp x19, x20, [sp, GPR_SAVE_OFFSET(19)]
	stp d14, d15, [sp, FPR_SAVE_OFFSET(14)]
	stp d12, d13, [sp, FPR_SAVE_OFFSET(12)]
	stp d10, d11, [sp, FPR_SAVE_OFFSET(10)]
	stp d8, d9, [sp, FPR_SAVE_OFFSET(8)]
	mov J9VMTHREAD, x0
	ldr x27, [J9VMTHREAD,{#}J9TR_VMThread_entryLocalStorage]
	add x28, sp, {#}JIT_GPR_SAVE_OFFSET(0)
	str x28, [x27,{#}J9TR_ELS_jitGlobalStorageBase]
```

gacholio:
jenkins compile xlinux jdk8

knn-k:
Updated module.xml.

gacholio:
jenkins compile xlinux jdk8

gacholio:
`spec.flags.arch_aarch64` is not defined. I suggest you use `arch_arm` and `env_data64`.

gacholio:
While you're at it, you may as well exclude the 32-bit file on 64-bit builds in preparation for having a build.

knn-k:
How is the spec.flags determined during the OpenJ9 build process?

knn-k:
I updated the files a little.
Not working yet, of course, but `libj9vm29.so` and `vmtest` are built successfully for aarch64 in my build environment.

knn-k:
Updated the files adding actual code to `c_cInterpreter`.
It depends on the changes in oti/j9nonbuilder.h and jilgen/jilconsts.c in #4447.

knn-k:
I can run `HelloWorld` on aarch64 Linux in the interpreter mode now.

knn-k:
Removed unnecessary comments from the .m4 files.
Some macros in `arm64helpers.m4`, marked as "To be filled", are not implemented yet.  I am going to write their code later in a separate PR along with the JIT work.

DanHeidinga:
@gacholio can you give this another review?

knn-k:
I would like this PR to be merged if the code is OK.
The aarch64 VM runs fine in the interpreter mode with the `c_cInterpreter` in this PR.

gacholio:
Sorry, I will get to this soon.

runtime/vm/arm64cinterp.m4
```diff
@@ -0,0 +1,70 @@
...
+
+include(arm64helpers.m4)
+
+	.file "arm64cinterp.s"
+
+DECLARE_PUBLIC(cInterpreter)
+
+START_PROC(c_cInterpreter)
+	sub sp, sp, {#}CINTERP_STACK_SIZE
+	stp x19, x20, [sp, GPR_SAVE_OFFSET(19)]
```
gacholio: Why isn't x20 going into slot 20? This will be extremely confusing.

gacholio: Sorry, I see it's a dual store now.

gacholio:
jenkins compile xlinux jdk8

runtime/vm/arm64cinterp.m4
knn-k: Yes, aarch64 `stp`/`ldp` instructions handle register pairs.

PR #3575, opened 2018-11-06
Add support for DefaultValue bytecode

tajila:
Add support for DefaultValue bytecode

Adds an initial implementation for defaultValue. This PR does not
update the verifier, that work will be done in a subsequent PR. 
As a result ValueTypes can only used with -Xverify:none for the 
time being. 

A subsequent PR will be added to provide tests for this bytecode.

This based on Eric's work in 
https://github.com/eclipse/openj9/pull/1235.

### DIFF
--- a/runtime/bcutil/ClassFileOracle.cpp
+++ b/runtime/bcutil/ClassFileOracle.cpp
@@ -35,6 +35,10 @@
 #include "ut_j9bcu.h"
 #include "util_api.h"
 
+#if defined(J9VM_OPT_VALHALLA_VALUE_TYPES)
+#define VALUE_TYPES_MAJOR_VERSION 55
+#endif /* J9VM_OPT_VALHALLA_VALUE_TYPES */
+
 /* The array entries must be in same order as the enums in ClassFileOracle.hpp */
 ClassFileOracle::KnownAnnotation ClassFileOracle::_knownAnnotations[] = {
 #define FRAMEITERATORSKIP_SIGNATURE "Ljava/lang/invoke/MethodHandle$FrameIteratorSkip;"
@@ -1739,6 +1743,16 @@ ClassFileOracle::walkMethodCodeAttributeCode(U_16 methodIndex)
 			break;
 		}
 
+#if defined(J9VM_OPT_VALHALLA_VALUE_TYPES)
+		case CFR_BC_defaultvalue:
+			if (_classFile->majorVersion >= VALUE_TYPES_MAJOR_VERSION) {
+				cpIndex = PARAM_U16();
+				addBytecodeFixupEntry(entry++, codeIndex + 1, cpIndex, ConstantPoolMap::DEFAULT_VALUE);
+				markClassAsUsedByDefaultValue(cpIndex);
+			}
+			break;
+#endif /* defined(J9VM_OPT_VALHALLA_VALUE_TYPES) */
+
 		default:
 			/* Do nothing */
 			break;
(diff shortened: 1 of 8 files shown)


tajila:
@gacholio please review this PR

runtime/vm/BytecodeInterpreter.hpp
```diff
@@ -8058,8 +8058,57 @@ done:;
...
+				}
+				newObject = instance;
+				goto done;
+			}
+			doInitializeClass = TRUE;
+		}
+
+		buildGenericSpecialStackFrame(REGISTER_ARGS, 0);
+		updateVMStruct(REGISTER_ARGS);
+		if (doInitializeClass) {
```
gacholio: I'm not sure all this complexity is worth it - can the initialize be part of the resolution?

runtime/vm/BytecodeInterpreter.hpp
```diff
@@ -8058,8 +8058,57 @@ done:;
...
+			/* Class requires initialization */
+			initializeClass(_currentThread, resolvedClass);
+		} else {
+			/* Class is unresolved or is not a value type */
+			resolveClassRef(_currentThread, ramConstantPool, index, J9_RESOLVE_FLAG_RUNTIME_RESOLVE | J9_RESOLVE_FLAG_INIT_CLASS | J9_RESOLVE_FLAG_INSTANTIABLE | J9_RESOLVE_FLAG_CHECK_VALUE_CLASS);
```
gacholio: Is there going to be another use of the CHECK_VALUE_CLASS flag? A failed resolution here will prevent any future resolution of the class ref.

gacholio:
I think I'm wrong here - given that the classref can already be resolved, I suppose you can't make any assumption about initialization or being a value type

runtime/vm/BytecodeInterpreter.hpp
gacholio: Actually, VT failures will not mark the CP entry (which may or may not be the correct behaviour). Still not convinced that we shouldn't just do the VT check here, and not pollute resolution with it.

runtime/bcutil/ClassFileOracle.cpp
```diff
@@ -1739,6 +1739,14 @@ ClassFileOracle::walkMethodCodeAttributeCode(U_16 methodIndex)
 			break;
 		}
 
+#if defined(J9VM_OPT_VALHALLA_VALUE_TYPES)
```
gacholio: Should this be a classfile version check?

runtime/vm/BytecodeInterpreter.hpp
gacholio: Given the first check is "resolved && valuetype", how about rewriting this in stages - check resolution, check valuetype, check class init, instead of nesting them together for no real savings.

gacholio: This does mean it now makes sense to move the VT check out of resolution completely since you have to do it here anyway.

runtime/vm/BytecodeInterpreter.hpp
gacholio: See above.

tajila:
Made a minor update to remove the `J9_RESOLVE_FLAG_CHECK_VALUE_CLASS` as it is not needed anymore

runtime/vm/BytecodeInterpreter.hpp
```diff
@@ -8058,8 +8058,56 @@ done:;
...
+		updateVMStruct(REGISTER_ARGS);
+
+		if (NULL == resolvedClass) {
+			resolveClassRef(_currentThread, ramConstantPool, index, J9_RESOLVE_FLAG_RUNTIME_RESOLVE | J9_RESOLVE_FLAG_INIT_CLASS | J9_RESOLVE_FLAG_INSTANTIABLE | J9_RESOLVE_FLAG_CHECK_VALUE_CLASS);
+		} else if (!J9_IS_J9CLASS_VALUETYPE(resolvedClass)) {
+			J9UTF8 *badClassName = J9ROMCLASS_CLASSNAME(resolvedClass->romClass);
+			setCurrentExceptionNLSWithArgs(_currentThread, J9NLS_VM_ERROR_BYTECODE_CLASSREF_MUST_BE_VALUE_TYPE, J9VMCONSTANTPOOL_JAVALANGINCOMPATIBLECLASSCHANGEERROR, J9UTF8_LENGTH(badClassName), J9UTF8_DATA(badClassName));
+		} else if (VM_VMHelpers::classRequiresInitialization(_currentThread, resolvedClass)) {
+			initializeClass(_currentThread, resolvedClass);
+		}
```
gacholio: I don't think this logic is correct - all of the unexpected conditions could be true, so elseif isn't appropriate.

gacholio: I suppose the code will eventually loop around and hit all the conditions.

tajila:
I've updated the PR

tajila:

Issue #3558, opened 2018-11-05
zlinux/plinux is not statically linking libstdc++.so


pshipton:
Unlike the xlinux builds, zlinux/plinux builds are not statically linking libstdc++.so

I checked a zlinux jdk11 build (the 0.11.0 release) downloaded from Adopt
This build uses gcc 7.3.
Also looked at a jdk8 build using gcc 4.8.5 and it is the same.

`ldd libj9jit29.so` shows
`libstdc++.so.6 => /usr/lib64/libstdc++.so.6 (0x000003fffc8e8000)`

`nm libj9jit29.so | grep GLIBCXX` shows
```
                 U _ZNSt8__detail15_List_node_base7_M_hookEPS0_@@GLIBCXX_3.4.15
                 U _ZNSt8__detail15_List_node_base9_M_unhookEv@@GLIBCXX_3.4.15
                 U _ZNSt9bad_allocD1Ev@@GLIBCXX_3.4
                 U _ZNSt9bad_allocD2Ev@@GLIBCXX_3.4
                 U _ZNSt9exceptionD2Ev@@GLIBCXX_3.4
                 U _ZSt18_Rb_tree_decrementPSt18_Rb_tree_node_base@@GLIBCXX_3.4
                 U _ZSt18_Rb_tree_incrementPKSt18_Rb_tree_node_base@@GLIBCXX_3.4
                 U _ZSt18_Rb_tree_incrementPSt18_Rb_tree_node_base@@GLIBCXX_3.4
                 U _ZSt20__throw_length_errorPKc@@GLIBCXX_3.4
                 U _ZSt24__throw_out_of_range_fmtPKcz@@GLIBCXX_3.4.20
                 U _ZSt28_Rb_tree_rebalance_for_erasePSt18_Rb_tree_node_baseRS_@@GLIBCXX_3.4
                 U _ZSt29_Rb_tree_insert_and_rebalancebPSt18_Rb_tree_node_baseS0_RS_@@GLIBCXX_3.4
                 U _ZSt7nothrow@@GLIBCXX_3.4
                 U _ZTVSt9bad_alloc@@GLIBCXX_3.4
```

pshipton:
Note the user encountered problem #3580 on plinux.

0dvictor:
I've investigated the issue on plinux. Removing `stdc++` from following line resolves the issue.

[...]

PR #3579, opened 2018-11-06
Enable gcc-7 on Linux x86-64 platforms

vsebe:
- remove uma_make_cmd_interp_gcc property and clean up its references

[ci skip]


buildspecs/linux_x86-64.spec
```diff
@@ -101,7 +101,7 @@ SPDX-License-Identifier: EPL-2.0 OR Apache-2.0 OR GPL-2.0 WITH Classpath-excepti
 		<property name="uma_make_cmd_cxx_exe_ld" value="$(CC)"/>
 		<property name="uma_make_cmd_dll_ld" value="$(CC)"/>
 		<property name="uma_make_cmd_exe_ld" value="$(CC)"/>
-		<property name="uma_make_cmd_interp_gcc" value="gcc-4.6"/>
```
pshipton: We can just remove this variable, and the places that use it, in mkconstants.mk.ftl line 44 and targets.mk.inux.inc.ftl line 511

pshipton:
> Blocked by: #3558

Actually I think this is blocked by #3594

buildspecs/linux_x86-64.spec
vsebe: done

PR #3589, opened 2018-11-07
Add a flag to switch on and off report dir staging

renfeiw:
- Added flag KEEP_REPORTDIR in setting.mk to control staging

[ci skip]

### DIFF
--- a/test/TestConfig/settings.mk
+++ b/test/TestConfig/settings.mk
@@ -206,7 +206,12 @@ REPORTDIR = $(Q)$(TESTOUTPUT)$(D)$@$(Q)
 #######################################
 # TEST_STATUS
 #######################################
-TEST_STATUS=if [ $$? -eq 0 ] ; then $(ECHO) $(Q)$(Q); $(ECHO) $(Q)$@$(Q)$(Q)_PASSED$(Q); $(ECHO) $(Q)$(Q); $(CD) $(TEST_ROOT); $(RM) -r $(REPORTDIR); else $(ECHO) $(Q)$(Q); $(ECHO) $(Q)$@$(Q)$(Q)_FAILED$(Q); $(ECHO) $(Q)$(Q); fi
+RM_REPORTDIR=
+KEEP_REPORTDIR?=true
+ifeq ($(KEEP_REPORTDIR), false)
+	RM_REPORTDIR=$(RM) -r $(REPORTDIR);
+endif
+TEST_STATUS=if [ $$? -eq 0 ] ; then $(ECHO) $(Q)$(Q); $(ECHO) $(Q)$@$(Q)$(Q)_PASSED$(Q); $(ECHO) $(Q)$(Q); $(CD) $(TEST_ROOT); $(RM_REPORTDIR) else $(ECHO) $(Q)$(Q); $(ECHO) $(Q)$@$(Q)$(Q)_FAILED$(Q); $(ECHO) $(Q)$(Q); fi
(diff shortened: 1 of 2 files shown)


renfeiw:
reviewer: @llxia

test/TestConfig/settings.mk
```diff
@@ -206,7 +206,11 @@ REPORTDIR = $(Q)$(TESTOUTPUT)$(D)$@$(Q)
 #######################################
 # TEST_STATUS
 #######################################
-TEST_STATUS=if [ $$? -eq 0 ] ; then $(ECHO) $(Q)$(Q); $(ECHO) $(Q)$@$(Q)$(Q)_PASSED$(Q); $(ECHO) $(Q)$(Q); $(CD) $(TEST_ROOT); $(RM) -r $(REPORTDIR); else $(ECHO) $(Q)$(Q); $(ECHO) $(Q)$@$(Q)$(Q)_FAILED$(Q); $(ECHO) $(Q)$(Q); fi
+REPORT_CLEAN_CMD=
+ifeq ($(REPORT_CLEANUP), true)
```
llxia: Could we have a similar name as `KEEP_WORKSPACE`? 

```
ifeq ($(KEEP_REPORTDIR), false)
...

```

Similarly, could we have `RM_REPORTDIR` instead of `REPORT_CLEAN_CMD`?

renfeiw: I will update them. A question, KEEP_REPORTDIR can be true or false by default, which one do you think is better?

smlambert: My preference is to KEEP_REPORTDIR true by default, as that is how it originally was designed, that is what a user running it locally would expect (and its the behaviour other underlying test frameworks also have).  

The main reason we added support to remove the report dir is to alleviate storage issues on Jenkins.  So, for the automated Jenkins jobs we can set KEEP_REPORTDIR to false.

Use case 1: local testing, expectation is keeping report dir (we should keep the number of variables individual devs have to set, so default setting of true makes sense)
Use case 2: Jenkins testing, expectation is removing report dir on successful test run - this is a one time change in scripts (low-cost, low aggravation)

renfeiw: Updated

renfeiw:
@AdamBrousseau We are adding this REPORT_CLEANUP to control if the passing test reports need to be cleaned up. After this PR is merged, the default behavior will be keeping all reports. Please feel free to set it to false in openJ9 builds if space issue occurs.

llxia:

AdamBrousseau:
Thanks for at-ing me. I'm trying to understand exactly what this is doing. Is this cleaning the REPORTDIR on the agent before the test results are archived? Does it clean just the passed test reports and not the failed reports? If the answer is yes, then I see how this will reduce the artifact size that we save in the case of an unstable build.

smlambert:
Adds a flag around deletion of successful test results.  As we had already delivered a change that would remove all successful test reports, @renfeiw put a flag on it, because we should only be removing successful test reports at the Jenkins servers that are space-constrained.  This PR adds a flag, and sets the default setting to keep the successful test reports, but still allows for them to be deleted by setting KEEP_REPORTDIR to false.

AdamBrousseau:
>already delivered a change that would remove all successful test reports

Ok so that explains my confusion :)
So then I believe my assumption about the artifact is correct. I will open a PR to update the Pipelines that call the test jobs.

PR #3591, opened 2018-11-07
Disable Vector Facility for z/TPF

jjohnst-us:
z/TPF does not currently support Vector Instructions and JVM
initializaiton will fail if enabled for z/TPF.  This PR configures zTPF
not to use it.

### DIFF
--- a/runtime/port/unix/j9sysinfo.c
+++ b/runtime/port/unix/j9sysinfo.c
@@ -269,7 +269,7 @@ static intptr_t getAIXPPCDescription(struct J9PortLibrary *portLibrary, J9Proces
 #elif (defined(S390) || defined(J9ZOS390) || defined(J9ZTPF))
 static BOOLEAN testSTFLE(struct J9PortLibrary *portLibrary, uint64_t stfleBit);
 static intptr_t getS390Description(struct J9PortLibrary *portLibrary, J9ProcessorDesc *desc);
-#endif
+#endif /* defined(S390) || defined(J9ZOS390) || defined(J9ZTPF) */
 
 #if (defined(LINUXPPC) || defined(AIXPPC))
 static J9ProcessorArchitecture mapPPCProcessor(const char *processorName);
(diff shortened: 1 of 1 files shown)


runtime/port/unix/j9sysinfo.c
```diff
@@ -1053,8 +1053,9 @@ getS390Description(struct J9PortLibrary *portLibrary, J9ProcessorDesc *desc)
 		if (J9_ARE_ALL_BITS_SET(auxvFeatures, J9PORT_HWCAP_S390_VXRS))
 #endif
 		{
+#if !defined(J9ZTPF)
```
DanHeidinga: Would it be cleaner to have `getS390zOS_supportsVectorExtensionFacility()` return false for zTPF?

jjohnst-us: @DanHeidinga Should I piggyback off of the existing zOS functions or should I create z/TPF versions... (e.g., getS390TPF_supportVectorExtensionFacility)?  The only reason I ask is those function are guarded with z/OS with some structure defintions.

Whatever you prefer.  tia..

DanHeidinga: I'm not sure it makes sense to add a new function just to return false.  My suggestion for using the zOS functions was they help document the reason why (no support for vectors) this is being disabled.

jjohnst-us: @DanHeidinga Actually we will try to fit into the Linux version using query_auxv.  Maybe hardcoding the flags as we don't have query_auxv on TPF.  We may implement one as well so that we completely are transparent with Linux configuration.  If that you are okay with that.

DanHeidinga: That makes sense.

jjohnst-us: This version sets the auxvFeatures variable as if it were set by an query_auxv call.  This allowed removing the z/tpF guard macros and making use of the J9_ARE_ALL_BITS_SET as if we were LINUX s390.

runtime/port/unix/j9sysinfo.c
```diff
@@ -838,24 +838,31 @@ getS390Description(struct J9PortLibrary *portLibrary, J9ProcessorDesc *desc)
 	if (J9_ARE_NO_BITS_SET(*(int*) 200, S390_STFLE_BIT)) {
 		return -1;
 	}
-#elif defined(LINUX) && !defined(J9ZTPF) /* LINUX S390*/
+#elif defined(J9ZTPF)  /* ZTPF */
+	// First turn off all features
+	unsigned long auxvFeatures = 0;
```
DanHeidinga: The values can be assigned directly.  No need for the initial 0 assignment first.

jjohnst-us: Done.

runtime/port/unix/j9sysinfo.c
```diff
@@ -838,24 +838,33 @@ getS390Description(struct J9PortLibrary *portLibrary, J9ProcessorDesc *desc)
 	if (J9_ARE_NO_BITS_SET(*(int*) 200, S390_STFLE_BIT)) {
 		return -1;
 	}
-#elif defined(LINUX) && !defined(J9ZTPF) /* LINUX S390*/
+#elif defined(J9ZTPF)  /* ZTPF */
```
DanHeidinga: It's a nitpick but the `#if / #elif / #else` close comments need to be adjusted to indicate the end of the previous one.

`#elif defined(J9ZTPF)  /* J9ZOS390 */`
and

`#elif defined(LINUX) /* J9ZTPF */`

DanHeidinga:
Sorry for @jjohnst-us for missing this the first time around.  Once that's changed, I can launch the builds on this.

runtime/port/unix/j9sysinfo.c
jjohnst-us: Thanks, done.

DanHeidinga:
Jenkins test sanity zlinux,plinux jdk8

PR #3592, opened 2018-11-07
Fix checking method handle access

pdbain-ibm:
Use the defining class of a class member or the target class of the lookup as
appropriate.  Also reinstate missing external message definition and replace duplicated
code with a common method.

Fixes https://github.com/eclipse/openj9/issues/3301

### DIFF
--- a/jcl/src/java.base/share/classes/java/lang/invoke/MethodHandles.java
+++ b/jcl/src/java.base/share/classes/java/lang/invoke/MethodHandles.java
@@ -383,8 +383,8 @@ public class MethodHandles {
 				throw new IllegalAccessException(this.toString());
 			}
 			
-			final Class<?> definingClass = handle.getDefiningClass();
-			checkAccess(definingClass, definingClass, handle.getFieldName(), handle.getModifiers(), null, skipAccessCheckPara);
+			/* VarHandles have no reference class */
+			checkAccess(handle.getDefiningClass(), null, handle.getFieldName(), handle.getModifiers(), null, skipAccessCheckPara);
 		}
 		
 		void checkAccess(Class<?> clazz) throws IllegalAccessException {
@@ -406,7 +406,8 @@ public class MethodHandles {
 		 * Equivalent of visible.c checkVisibility();
 		 * 
 		 * @param definingClass The {@link Class} that defines the member being accessed.
-		 * @param referenceClass The {@link Class} class through which the the member is accessed, which may be a subtype of the defining class.
+		 * @param referenceClass The {@link Class} class through which the the member is accessed, 
+		 * which must be the defining class or a subtype.  May be null.
 		 * @param name The name of member being accessed.
 		 * @param memberModifiers The modifiers of the member being accessed.
 		 * @param handle A handle object (e.g. {@link MethodHandle} or {@link VarHandle}), if applicable.
@@ -418,6 +419,9 @@ public class MethodHandles {
 		 * @throws IllegalAccessError If a handle argument or return type is not accessible.
 		 */
 		private void checkAccess(Class<?> definingClass, Class<?> referenceClass, String name, int memberModifiers, MethodHandle handle, boolean skipAccessCheckPara) throws IllegalAccessException {
+			if (null == referenceClass) {
+				referenceClass = definingClass;
+			}
 			checkClassAccess(referenceClass);
 
 			/*[IF Sidecar19-SE]*/
--- a/jcl/src/java.base/share/classes/java/lang/invoke/MethodHandles.java
+++ b/jcl/src/java.base/share/classes/java/lang/invoke/MethodHandles.java
@@ -364,7 +364,12 @@ public class MethodHandles {
 				throw new IllegalAccessException(this.toString());
 			}
 
-			checkAccess(handle.getDefc(), handle.getMethodName(), handle.getModifiers(), handle, skipAccessCheckPara);
+			/* defc (defining class) may be a superclass of the references class.
+			 * Note that we need to check against the requested class, which may have inherited
+			 * the method from an inaccessible class. Protected members of a subclass are accessible to 
+			 * a superclass provided they are inherited from the superclass. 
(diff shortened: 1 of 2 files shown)


pdbain-ibm:
Will update to show the correspondence with bytecode behaviour.

pdbain-ibm:
`resolveField.cpp:findFieldAndCheckVisibility()` calls `checkVisibility()` on the _defining_ class of the target field (ca. line 324).  `visible.c:checkVisibility()` succeeds if the target field is `private` and in the same class as the access class (ca. line 134).  This corresponds to `MethodHandles.checkAccess()` ca. line 435.

checkVisibility() passes if the field is `protected` and in a superclass of the access class (ca. line 159). This corresponds to `MethodHandles.checkAccess()` ca. line 460.

Minor tweak to commit, tested with JTReg test and Jsr292 tests.
Would a committer please run sanity and extended builds.  Any platform will do.
Thank you

pdbain-ibm:
Updating and retesting this...

pdbain-ibm:
Ready for review @DanHeidinga

jcl/src/java.base/share/classes/java/lang/invoke/MethodHandles.java
```diff
@@ -407,7 +409,7 @@ void checkAccess(Class<?> clazz) throws IllegalAccessException {
 		 * @throws IllegalAccessException If the member is not accessible.
 		 * @throws IllegalAccessError If a handle argument or return type is not accessible.
 		 */
-		private void checkAccess(Class<?> targetClass, String name, int memberModifiers, MethodHandle handle, boolean skipAccessCheckPara) throws IllegalAccessException {
+		private void checkAccess(Class<?> targetClass, Class<?> definingClass, String name, int memberModifiers, MethodHandle handle, boolean skipAccessCheckPara) throws IllegalAccessException {
```
DanHeidinga: I really dislike changing the meaning of the variable name here as it makes it hard to review whether changes were deliberate or not.

DanHeidinga:
Sorry, just realized I hadn't completed the review yet.  It's been sitting in a pending state for the last few days

pdbain-ibm:
@DanHeidinga 
Thanks for the feedback.  I reverted non-essential changes and renamed the argument for clarity.
Tested on the jtreg testcase.

jcl/src/java.base/share/classes/java/lang/invoke/MethodHandles.java
```diff
@@ -434,22 +441,22 @@ private void checkAccess(Class<?> targetClass, String name, int memberModifiers,
...
 				/* Ensure that the accessMode is not restricted (public-only) */
 				if (PUBLIC != accessMode) {
-					if (targetClass.isArray()) {
+					if (definingClass.isArray()) {
 						/* The only methods array classes have are defined on Object and thus accessible */
 						return;
 					}
 					
-					if (isSamePackage(accessClass, targetClass)) {
+					if (isSamePackage(accessClass, referenceClass)) {
```
ChengJin01: The API Spec at https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/lang/invoke/MethodHandles.Lookup.html
```
If the desired member is protected, the usual JVM rules apply, including the requirement that the lookup class 
must be either be in the same package as the desired member, or must inherit that member. 
(See the Java Virtual Machine Specification, sections 4.9.2, 5.4.3.5, and 6.4.) In addition, if the desired member
 is a non-static field or method in a different package, the resulting method handle may only be applied to 
 objects of the lookup class or one of its subclasses. 
```
So changing to referenceClass seems incorrect at this point.

jcl/src/java.base/share/classes/java/lang/invoke/MethodHandles.java
```diff
@@ -497,7 +506,7 @@ private void checkAccess(Class<?> targetClass, String name, int memberModifiers,
 				}
 			} else {
 				/* default (package access) */
-				if ((PACKAGE == (accessMode & PACKAGE)) && isSamePackage(accessClass, targetClass)){
+				if ((PACKAGE == (accessMode & PACKAGE)) && isSamePackage(accessClass, referenceClass)){
```
ChengJin01: I am not too sure whether it is correct to change to referenceClass.

ChengJin01:
Introducing the reference class to checkAccess() is definitely correct to address the case but how to deal with the relationship between the defining class and the reference class is up to us as the API Spec doesn't explain in details at this point.

jcl/src/java.base/share/classes/java/lang/invoke/MethodHandles.java
pdbain-ibm: In this case, the method is not public, so we need to check for package access.  Consider the case where the lookup class is in the same package as the reference class, but a different package from the defining class.  If we check against the defining class we would fail, even though the method is visible via the reference class.

Note also that the reference class may broaden the visibility of the method.

jcl/src/java.base/share/classes/java/lang/invoke/MethodHandles.java
pdbain-ibm: `referenceClass` must be a subtype of `definingClass` in order to inherit the method.

pdbain-ibm:
In reply to 
https://github.com/eclipse/openj9/pull/3592#issuecomment-453568838, most of the uses of `definingClass` pertain to private methods or access via inheritance from the defining class to the lookup class in cases where the lookup class does not extend the reference class.

pdbain-ibm:
Found an issue when testing with a VarHandles test case: the defining class can sometimes be null.
Will update the changeset.

pdbain-ibm:
I added a second commit to fix the null reference class problem.  I will squash it when this is approved.

DanHeidinga:
Jenkins test all xlinux jdk11,jdk8

DanHeidinga:

DanHeidinga:

PR #3598, opened 2018-11-07
Ensure GC map is generated at BB entry on all platforms

### LINKED ISSUE #3374: Ensure stack maps are generated for exception catch blocks on x86

### LINKED ISSUE #3375: Ensure stack maps are generated for exception catch blocks on Power
Please see the parent Epic for further details. We need to ensure that under the new mode we generate stack maps at the start of every exception catch block and ensure that we don't enable throw to goto transformations when this mode is enabled.

### LINKED ISSUE #3376: Ensure stack maps are generated for exception catch blocks on Z
Please see the parent Epic for further details. We need to ensure that under the new mode we generate stack maps at the start of every exception catch block and ensure that we don't enable throw to goto transformations when this mode is enabled.

fjeremic:
Stack maps are required at the entry of exception catch
blocks under FSD mode and when Agent is attached.

Closes #3374
Closes #3375
Closes #3376
Closes #3462

### DIFF
--- a/runtime/compiler/codegen/J9CodeGenerator.cpp
+++ b/runtime/compiler/codegen/J9CodeGenerator.cpp
@@ -4526,8 +4526,17 @@ J9::CodeGenerator::generateCatchBlockBBStartPrologue(
       TR::Node *node,
       TR::Instruction *fenceInstruction)
    {
+   if (self()->comp()->getOptions()->getReportByteCodeInfoAtCatchBlock())
+      {
+      // Note we should not use `fenceInstruction` here because it is not the first instruction in this BB. The first
+      // instruction is a label that incoming branches will target. We will use this label (first instruction in the
+      // block) in `createMethodMetaData` to populate a list of non-mergable GC maps so as to ensure the GC map at the
+      // catch block entry is always present if requested.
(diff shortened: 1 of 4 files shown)


fjeremic:
ARM currently has a problem as the code which does the catch block handling is only partially implemented and it sits in OMR. It's also missing a call to the common code generator. I've fixed this in https://github.com/eclipse/omr/pull/3158 so this should cover ARM support as well.

fjeremic:
Preemptively launching builds here. @andrewcraik could I ask you to be the committer for this one?

Jenkins test sanity all JDK8

gacholio:
osx is not ready, should not be part of "all" for jdk8.

fjeremic:
@gacholio also feel free to assign yourself to this PR if you feel comfortable merging given the simplicity of the change.

gacholio:
If @andrewcraik doesn't get to it when the builds finish, I may do that.

andrewcraik:
I will look at results early tomorrow so @gacholio can monitor more closely if he wants to move this through before then.

Issue #3597, opened 2018-11-07
Memory leak for options allocated at the JITaaS server


mpirvu:
Options from the client are serialized and sent to the JITaaS server which in turn will materialize them using persistent memory. This block of persistent memory is attached to a compilation request entry. At the end of the compilation the options need to be freed when the compilation request entry is put back into the pool for recycling (`void TR::CompilationInfo::recycleCompilationEntry(TR_MethodToBeCompiled *entry)`).
However, current code only frees the options when the compilation request entry is totally deallocated (this happens if the pool of free entries grows too much).

PR #3599, opened 2018-11-07
Fix JITaaS Options deallocation

harryyu1994:
Previously JITaaS Options strings are deallocated in MethodToBeCompiled::shutdown() when entries are destroyed. However, we often recycle MethodToBeCompiled entries for use and that does not call shutdown(). We should therefore deallocate the strings when entries are recycled.

Issue: #3597
[skip ci]

### DIFF
--- a/runtime/compiler/control/CompilationThread.cpp
+++ b/runtime/compiler/control/CompilationThread.cpp
@@ -5044,6 +5044,7 @@ void TR::CompilationInfo::recycleCompilationEntry(TR_MethodToBeCompiled *entry)
    entry->_freeTag |= ENTRY_IN_POOL_NOT_FREE;
    if (entry->_numThreadsWaiting == 0)
       entry->_freeTag |= ENTRY_IN_POOL_FREE;
+   entry->cleanupJITaaS();
 
    entry->_next = _methodPool;
    _methodPool = entry;
--- a/runtime/compiler/control/MethodToBeCompiled.hpp
+++ b/runtime/compiler/control/MethodToBeCompiled.hpp
@@ -65,6 +65,7 @@ struct TR_MethodToBeCompiled
    void setRemoteCompReq() { _remoteCompReq = true; }
    bool isOutOfProcessCompReq() const { return _stream != nullptr; } // at the server
    uint64_t getClientUID() const;
+   void cleanupJITaaS();
 
    TR_MethodToBeCompiled *_next;
    TR::IlGeneratorMethodDetails _methodDetailsStorage;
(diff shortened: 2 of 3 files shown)


harryyu1994:
@mpirvu Hi Marius, this is ready for review.

A question about this, would it be safe to deallocate the strings when the entry is still in `ENTRY_IN_POOL_NOT_FREE` state. (aka already in the pool but entry->_numThreadsWaiting hasn't reached 0). If that's the case, we can avoid calling the deallocation in two places.

Also I removed the code in `shutDown()` because `shutDown()` is only called on `ENTRY_IN_POOL_FREE` entries and `ENTRY_IN_POOL_FREE` entries should have their clientOptions string deallocated already.

harryyu1994:
I think I'm wrong about the `shutDown()` being only called on ENTRY_IN_POOL_FREE entries. Still need to clean up in `shutDown()`.

mpirvu:
>>  would it be safe to deallocate the strings when the entry is still in ENTRY_IN_POOL_NOT_FREE state.

Yes, because compilation is done, we only need the monitor from the entry to synchronize with other threads waiting for the compilation to be finished.

runtime/compiler/control/MethodToBeCompiled.hpp
```diff
@@ -65,6 +65,7 @@ struct TR_MethodToBeCompiled
    void setRemoteCompReq() { _remoteCompReq = true; }
    bool isOutOfProcessCompReq() const { return _stream != nullptr; } // at the server
    uint64_t getClientUID() const;
+   void TR_MethodToBeCompiled::cleanUpJITaaS();
```
mpirvu: remove the class name here

runtime/compiler/control/CompilationThread.cpp
```diff
@@ -5036,7 +5036,10 @@ void TR::CompilationInfo::recycleCompilationEntry(TR_MethodToBeCompiled *entry)
                    "recycling an improper entry\n");
    entry->_freeTag |= ENTRY_IN_POOL_NOT_FREE;
    if (entry->_numThreadsWaiting == 0)
+      {
```
mpirvu: I would move the cleanup outside the `if` statement because otherwise you have another cleanup in another part of the code when entry moves `FREE` state (which you did)

harryyu1994:
Updated as requested.

mpirvu:
Will merge after bringing jitaas branch up-to-date with master openj9 branch

harryyu1994:
Rebased to latest jitaas branch

PR #3605, opened 2018-11-07
Eliminate redundant bytecode size table

gacholio:
Information in JavaBytecodeRelocation is available in
J9JavaInstructionSizeAndBranchActionTable instead.

Also makes the C and DDR code more congruent.


gacholio:

pshipton:
bcsize.c has changes for J9VM_OPT_VALHALLA_VALUE_TYPES that are not in pcstack.c

gacholio:
Seeing as how I'm deleting the table, I don't think any of the above comments apply. The table was used in exactly one place (decompiler treating allocation bytecodes).

DanHeidinga:
The value for `JBunimplemented` is `0` in pcstack but `1` in bcsizes.  I don't think it matters as the verifier will detect any use of `JBunimplemented` and throw VerifyError.

pshipton:
What about the J9VM_OPT_VALHALLA_VALUE_TYPES sizes that are removed by this change?

DanHeidinga:
> What about the J9VM_OPT_VALHALLA_VALUE_TYPES sizes that are removed by this change?

@pshipton  Can you provide a link to the line number?  I don't see this define at all in the changeset....

pshipton:
> Can you provide a link to the line number? I don't see this define at all in the changeset....

nm, I had it backwards. The J9VM_OPT_VALHALLA_VALUE_TYPES changes are in pcstack.c but not in bcsize.c This change lgtm

Issue #3565, opened 2018-11-06
ASSERTION FAILED ** at classsupport.c:1244 running Build-JDK11-win_x86-64_cmprssptrs


pshipton:
https://ci.eclipse.org/openj9/job/Build-JDK11-win_x86-64_cmprssptrs/159
```
00:54:12 JVMJ9VM019E Unrecoverable error: Unable to find and initialize required class java/lang/Byte
00:54:12 JVMJ9VM023I This may indicate that JAVA_HOME is incorrect, or that class libraries are not installed
00:54:12 05:55:10.107 0x19a00    j9vm.546    *   ** ASSERTION FAILED ** at classsupport.c:1244: ((0 ))
```

pshipton:
1244:	Assert_VM_internalFindKnownClassFailure();

gacholio:
Likely filesystem flakiness. None of the changes in the PR would have caused this. Java ran successfully before the error.

pshipton:
Agreed, but I wanted to record it in case it occurs again.

pshipton:
Occurred again, looking like a real issue
https://ci.eclipse.org/openj9/job/Build-JDK11-win_x86-64_cmprssptrs/160

pshipton:
Blocking omr acceptance builds as well

pshipton:
#3542 is the suspect change

DanHeidinga:
I've opened PR #3609 as a way to test whether this change was the culprit as I'm running into issues building windows locally.

PR #3609, opened 2018-11-08
Revert to loading `../java` now that OMR is updated

DanHeidinga:
Current behaviour of using the full path is causing odd
behaviour on windows.  This worked before so reverting.

fixes: #3565

### DIFF
--- a/runtime/jcl/common/system.c
+++ b/runtime/jcl/common/system.c
@@ -412,15 +412,9 @@ jstring getEncoding(JNIEnv *env, jint encodingType)
 #if defined(J9VM_JCL_SE11)
 		{
 			UDATA handle = 0;
-			J9JavaVM * const vm = ((J9VMThread*)env)->javaVM;
-			char dllPath[EsMaxPath];
-			UDATA written = 0;
 			PORT_ACCESS_FROM_ENV(env);
 			/* libjava.[so|dylib] is in the jdk/lib/ directory, one level up from the default/ & compressedrefs/ directories */
-			written = j9str_printf(PORTLIB, dllPath, sizeof(dllPath), "%s/../java", vm->j9libvmDirectory);
-			/* Assert the number of characters written (not including the null) fit within the dllPath buffer */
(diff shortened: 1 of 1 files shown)


DanHeidinga:
Jenkins test sanity win jdk11

DanHeidinga:

DanHeidinga:
The windows build is running testing now so this will unblock the build.

I still want to find out why the original change was causing problems and add a tracepoint for the failure case.

runtime/jcl/common/system.c
```diff
@@ -412,15 +412,9 @@ jstring getEncoding(JNIEnv *env, jint encodingType)
...
-			char dllPath[EsMaxPath];
-			UDATA written = 0;
 			PORT_ACCESS_FROM_ENV(env);
 			/* libjava.[so|dylib] is in the jdk/lib/ directory, one level up from the default/ & compressedrefs/ directories */
-			written = j9str_printf(PORTLIB, dllPath, sizeof(dllPath), "%s/../java", vm->j9libvmDirectory);
-			/* Assert the number of characters written (not including the null) fit within the dllPath buffer */
-			Assert_JCL_true(written < (sizeof(dllPath) - 1));
-			if (0 == j9sl_open_shared_library(dllPath, &handle, J9PORT_SLOPEN_DECORATE)) {
+			if (0 == j9sl_open_shared_library("../java", &handle, J9PORT_SLOPEN_DECORATE)) {
 				void (*nativeFuncAddrJNU)(JNIEnv *env, const char *str) = NULL;
```
keithc-ca: This function pointer should have the `JNICALL` attribute.

pshipton: Created new issue #3614 to address this so I can merge this blocker fix.

DanHeidinga:
PR https://github.com/eclipse/openj9/pull/3616 adds the JNICALL decoration (even though it's technically unnecessary for JDK11 as it doesn't support win32)

keithc-ca:
Someone will try to build for 32-bit Windows. ;-)

PR #3612, opened 2018-11-08
Enable gcc-7 on zLinux 64bits platforms

vsebe:
Update graph_enable_gcc7_cmd property to enable gcc-7 environment


fjeremic:
Is this gcc 7.3 that is being enabled here?

vsebe:
@fjeremic - yes, it is.

pshipton:
There is a potential performance concern, setting WIP for now.

vsebe:
Rebased, resolved merge conflict.

pshipton:
The -Xint performance concern is now understood, related to the gcc LRA (the register allocator). It can be worked around by using the `-mno-lra` debug option. Since we know using `-mno-lra` is a possibility, if necessary, I'm going ahead with this change.

PR #3617, opened 2018-11-09
Disable Atomic-Free JNI on z/TPF

jjohnst-us:
z/TPF does not support mprotect nor mlock which are used as part of this
new feature.

### DIFF
--- a/buildspecs/linux_ztpf_390-64.spec
+++ b/buildspecs/linux_ztpf_390-64.spec
@@ -117,8 +117,8 @@ SPDX-License-Identifier: EPL-2.0 OR Apache-2.0 OR GPL-2.0 WITH Classpath-excepti
 	</source>
 	<flags>
 		<flag id="interp_atomicFreeJni" value="true"/>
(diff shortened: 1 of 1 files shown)


DanHeidinga:
@jjohnst-us The longer term plan is to remove the old code paths and standardize on the new way.  zTPF will need to look at alternative apis to support the new atomic-free jni code paths.

fyi @gacholio

gacholio:
The solution here is to not use the "uses flush" flag. Atomic-free works with memory barriers as well (much like ZOS).

gacholio:
This also means that you don't need "two pass" either.

jjohnst-us:
Made update per your recommendation.

Let me know if you need anything else.

gacholio:
My merging is a bit premature - you also need to change `linux_ztpf_390-64_cmprssptrs.spec`

jjohnst-us:
We don't support cmprssptrs.

gacholio:
Then we should delete the spec rather than let it get stale (easy enough to resurrect it from the large heap spec if need be).

PR #3619, opened 2018-11-09
Allow setter method handles to final fields

pdbain-ibm:
Allow unreflection to get setters on a final field if it is set accessible.

Fix minor unrelated javadoc error to avoid compilation warning.

Fixes https://github.com/eclipse/openj9/issues/3175

### DIFF
--- a/jcl/src/java.base/share/classes/java/lang/invoke/MethodHandles.java
+++ b/jcl/src/java.base/share/classes/java/lang/invoke/MethodHandles.java
@@ -1444,8 +1444,10 @@ public class MethodHandles {
 			Class<?> declaringClass = field.getDeclaringClass();
 			Class<?> fieldType = field.getType();
 			String fieldName = field.getName();
-			
-			if (Modifier.isFinal(modifiers)) {
+			/* https://github.com/eclipse/openj9/issues/3175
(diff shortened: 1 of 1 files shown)


pdbain-ibm:
Updated to prohibit `unreflectSetter()` on a `static final` field, for consistency with `Field.set()`.
Would a committer please test & review this?

Thank you

pdbain-ibm:
@gacholio would you kindly review this and run a test build? Any platform should be sufficient.

gacholio:

jzhn:
Thanks for fixing this issue! Is there any chance to get this fix backported to 0.11?

Our springboot microservice is using spring-data which would hit this issue when some of the fields are final: 
https://github.com/spring-projects/spring-data-commons/blob/master/src/main/java/org/springframework/data/mapping/model/ClassGeneratingPropertyAccessorFactory.java#L241-L243

This is currently the blocker issue for us to move to openj9. Any advice is greatly appreciated!

DanHeidinga:
The 0.12 release (which will cover both JDK 8 & 11) is targeted for the end of January.  See the release plan: https://projects.eclipse.org/projects/technology.openj9/releases/0.12.0/plan

PR #3623, opened 2018-11-09
Add multithreaded backward arraycopy test

r30shah:
This commits adds a Java Test to verify atomicity of doing array copy in babckward direction.

### DIFF
--- a/test/functional/VM_Test/src/j9vm/test/arraycopy/MultiThreadedBackwardArrayCopyTest.java
+++ b/test/functional/VM_Test/src/j9vm/test/arraycopy/MultiThreadedBackwardArrayCopyTest.java
@@ -28,84 +28,84 @@ import java.util.HashSet;
  * Test atomicity of backward arraycopy in a multithreaded environment.
  */
 
-public class MultiThreadedBackwardArrayCopyTest implements Runnable{
-   static Object[] source = new Object[] {
-      createValidValue(1),
-      createValidValue(2),
-      createValidValue(3),
-      createValidValue(4),
-      createValidValue(5),
-      createValidValue(6),
-      createValidValue(7),
-      createValidValue(8),
-      createValidValue(9),
-      createValidValue(10),
-      createValidValue(11),
-      createValidValue(12),
-      createValidValue(13),
-      createValidValue(14),
-      createValidValue(15),
-      createValidValue(16),
-      createValidValue(17),
-      createValidValue(18),
-      createValidValue(19),
-      createValidValue(20),
-      createValidValue(21),
-      createValidValue(22),
-      createValidValue(23),
-      createValidValue(24),
-      createValidValue(25),
-      createValidValue(26),
-      createValidValue(27),
-      createValidValue(28),
-      createValidValue(29),
-      createValidValue(30),
-      createValidValue(31),
-      createValidValue(32),
-      createValidValue(33),
-      createValidValue(34),
-   };
+public class MultiThreadedBackwardArrayCopyTest implements Runnable {
+	static Object[] source = new Object[] {
+		createValidValue(1),
+		createValidValue(2),
+		createValidValue(3),
+		createValidValue(4),
+		createValidValue(5),
+		createValidValue(6),
+		createValidValue(7),
+		createValidValue(8),
+		createValidValue(9),
+		createValidValue(10),
+		createValidValue(11),
+		createValidValue(12),
+		createValidValue(13),
+		createValidValue(14),
+		createValidValue(15),
+		createValidValue(16),
+		createValidValue(17),
+		createValidValue(18),
+		createValidValue(19),
+		createValidValue(20),
+		createValidValue(21),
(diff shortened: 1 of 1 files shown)


r30shah:
FYI @fjeremic  @DanHeidinga  @dmitripivkine . Can I please get review on this. I created this test to check my changes for accelerating backward array copy sequence. I checked https://github.com/eclipse/openj9/blob/master/test/functional/Java8andUp/src/j9vm/test/arrayCopy/ArrayCopyTest.java 
This will test array copy in both backward and forward direction in single thread. I used logic from https://github.com/eclipse/openj9/blob/master/test/functional/VM_Test/src/j9vm/test/arraycopy/MultiThreadedArrayCopyTest.java where we test atomicity of forward array copy where multiple threads are concurrently accesses same array. For backward array copy I needed specific overlap to force it so my implementation simply rotates array right.

test/functional/VM_Test/src/j9vm/test/arraycopy/MultiThreadedBackwardArrayCopyTest.java
```diff
@@ -0,0 +1,111 @@
...
+   private static Object createValidValue(int i){
+		// Create array of bytes to spreadout the objects
+      byte [] unused = new byte[17832];
+      return Integer.toString(i);
+   }
+
+	// Rotate source right forcing it to do array copy in backward direction
+   public static void rotateSource(){
+      Object temp = source[source.length-1]; // Getting last element in temp
+      System.arraycopy(source, 0, source, 1, source.length-1);
```
dmitripivkine: Would you please try to copy array starting element [1], not [0] and run it in Compressed Refs JVM? The start address of array (as well as start address of data for contiguous array &array[0]) is aligned to 8 bytes always. However size of reference array element can be 4 for Compressed Refs, so an address of &array[1] would not be aligned to 8. Would it be atomic copy in this case?

test/functional/VM_Test/src/j9vm/test/arraycopy/MultiThreadedBackwardArrayCopyTest.java
```diff
@@ -0,0 +1,111 @@
...
+
+	// Create a unique String object
+	private static Object createValidValue(int i) {
+		// Create array of bytes to spreadout the objects
+		byte [] unused = new byte[17832];
+		return Integer.toString(i);
+	}
+
+	// Rotate source right forcing it to do array copy in backward direction
+	public static void rotateSource() {
```
r30shah: @dmitripivkine  Made changes as suggested. Ran test locally with comressesdref SPEC on Linux on Z. 
Test passes. I see it runs test in two configuration. 
1. `-Xcompressedrefs -Xjit -Xgcpolicy:balanced`
2. `-Xcompressedrefs -Xjit -Xgcpolicy:gencon`
verified through JIT method log and jit verbose log. `rotateSource()` is compiled. And it executes backward array copy path. (`src < dst && src+length > dst`)

test/functional/VM_Test/src/j9vm/test/arraycopy/MultiThreadedBackwardArrayCopyTest.java
```diff
@@ -28,84 +28,84 @@
...
 	// Rotate source right forcing it to do array copy in backward direction
-   public static void rotateSource(){
-      Object temp = source[source.length-1]; // Getting last element in temp
-      System.arraycopy(source, 0, source, 1, source.length-1);
-      source[0] = temp;
-   }
-
+	public static void rotateSource() {
+		Object temp = source[source.length-1]; // Getting last element in temp
+		System.arraycopy(source, 1, source, 2, source.length-2);
```
fjeremic: Spaces needed around arithmetic operators here.

r30shah: Changed in https://github.com/eclipse/openj9/pull/3623/commits/254a0edac69d3b5254e783c2997c388a1a5055e6

fjeremic:
@r30shah do we need to add this test to one of the xml files so it actually runs as part of PR testing?

r30shah:
@fjeremic I think it would run it as a part of VM_Test which I think is ran with every sanity test. Looking at https://github.com/eclipse/openj9/blob/82275daf0c7145f8ce63878ca43961b8b2e17039/test/functional/VM_Test/playlist.xml#L37 , I do not think we need to add it to playlist.xml
@llxia Am I correct? As I have added this test in VM_Test , it will be run when we launch sanity test on PR right?

llxia:
yes, it will be part of sanity test in `J9vmTest`. In this case, no playlist modification is needed.

In PR build console output, you should see the test name and test detail.
```
+++ j9vm.test.arraycopy.MultiThreadedBackwardArrayCopyTest: +++
 command: ....
```

fjeremic:
Jenkins test sanity all JDK8

Issue #3504, opened 2018-11-01
Reduce the number of Recompilation_getExistingMethodInfo messages


mpirvu:
The JITaaS server will issue `Recompilation_getExistingMethodInfo` messages in `J9::Recompilation::setupMethodInfo()` which is executed during the constructor of `TR_X86Recompilation` object. Since any compilation that can be followed by a re-compilation has a recompilation object, it may make sense for the client to send the existing method info together with the compilation request and save one message pair. We just need a place to stash the desired method info until is needed.

dmitry-ten:
We are already sending `TR_PersistentMethodInfo` when creating a resolved method mirror.
Since at that point we already have a resolved method for method being compiled, we should be able to just access method info from there.

dmitry-ten:
Then it makes sense to send resolved method info for method being compiled with compilation request, this would save us another message pair.

mpirvu:
Note that for a TR_ResolvedJ9Method representing a method undergoing compilation
there could be two PersistentJittedBodyInfos at the server
(1) One that describes the previous compilation (uses heap memory) and it is cached in the resolvedMethod. This one is NULL for first time compilations (which are the vast majority of compilations).
  We should access this through compilee->getExistingJittedBodyInfo()

(2) Another one that describes current compilation (uses persistent memory)
  We should access this through the recompilation object
  compiler->getRecompilationInfo()->getJittedBodyInfo().  Recompilation_getExistingMethodInfo messages are sent for this one.

dmitry-ten:
I see. Then we could store persistent method info in `TR_MethodToBeCompiled`.

PR #3624, opened 2018-11-09
Eliminate Recompilation_getExistingMethodInfo messages

dmitry-ten:
Persistent method info is initialized in Recompilation
object, which is created for every compilation
that can be recompiled. Thus, it makes sense to send this
method info as a part of compile request.
Once server receives it, method info is allocated and stashed in
`TR::CompilationInfoPerThreadRemote`, until it is time to create a recompilation object.

Addresses:#3504

### DIFF
--- a/runtime/compiler/control/J9Recompilation.cpp
+++ b/runtime/compiler/control/J9Recompilation.cpp
@@ -63,18 +63,17 @@ J9::Recompilation::setupMethodInfo()
    //
    TR_OptimizationPlan * optimizationPlan =  _compilation->getOptimizationPlan();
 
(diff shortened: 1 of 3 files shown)


dmitry-ten:
@mpirvu could you review?

mpirvu:
This item can be resumed now. `_recompilationMethodInfo` could be stored into `CompilationInfoPerThreadRemote` in `TR::CompilationInfoPerThreadRemote::processEntry(TR_MethodToBeCompiled &entry, J9::J9SegmentProvider &scratchSegmentProvider)` when we read the compilation request from the client

dmitry-ten:
@mpirvu this is ready for another review.

PR #3625, opened 2018-11-09
Add -XX:DebugLocalMapper to force using the debug stack mapper

DanHeidinga:
The debugLocalMapper (debuglocalmapper.c) is much more conservative
about object lifetimes than the regular local mapper.

The regular mapper walks from the current pc to the end of the method
and only reports locals containing objects as alive if they will be
used as objects again.

The debug local mapper walks from the start of the method to the current
pc and determines if the slot previously contained a live object.  This
is the behaviour typically wanted when debugging to ensure that the
debugger can show you the current object values in locals.

Occasionally, we want to force the use of the debug mapper to test
whether the shortened object lifetimes are affecting poorly written user
code, usually due to the use of finalizers.

This PR adds a new option `-XX:DebugLocalMapper` which uses the debug
local mapper to help diagnose and debug these kinds of issues.

### DIFF
--- a/runtime/oti/jvminit.h
+++ b/runtime/oti/jvminit.h
@@ -328,6 +328,7 @@ enum INIT_STAGE {
 #define VMOPT_XXHANDLESIGXFSZ "-XX:+HandleSIGXFSZ"
 #define VMOPT_XXHEAPDUMPONOOM "-XX:+HeapDumpOnOutOfMemoryError"
 #define VMOPT_XXNOHEAPDUMPONOOM "-XX:-HeapDumpOnOutOfMemoryError"
+#define VMOPT_XXDEBUGLOCALMAPPER "-XX:DebugLocalMapper"
 
 #define VMOPT_XSOFTREFTHRESHOLD "-XSoftRefThreshold"
 #define VMOPT_XAGGRESSIVE "-Xaggressive"
--- a/runtime/vm/jvminit.c
+++ b/runtime/vm/jvminit.c
@@ -81,6 +81,7 @@
 #include "bcnames.h"
 #include "jimagereader.h"
 #include "vendor_version.h"
+#include "stackmap_api.h"
 
 #ifdef J9VM_OPT_ZIP_SUPPORT
 #include "zip_api.h"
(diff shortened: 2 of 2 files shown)


DanHeidinga:
I don't intend to document this option as it purely for development debugging.

runtime/oti/jvminit.h
```diff
@@ -328,6 +328,7 @@ enum INIT_STAGE {
 #define VMOPT_XXHANDLESIGXFSZ "-XX:+HandleSIGXFSZ"
 #define VMOPT_XXHEAPDUMPONOOM "-XX:+HeapDumpOnOutOfMemoryError"
 #define VMOPT_XXNOHEAPDUMPONOOM "-XX:-HeapDumpOnOutOfMemoryError"
+#define VMOPT_XXDEBUGLOCALMAPPER "-XX:DebugLocalMapper"
```
pshipton: Should the option be -XX:+DebugLocalMapper ?

pshipton:
Could add a simple test to ensure the option doesn't cause a crash or throw exceptions or other undesired behavior.

runtime/oti/jvminit.h
DanHeidinga: No, a `+` version implies that there should also be a `-` version.  I've deliberately coded this so the debug local mapper can't be disabled once enabled as doing so would bad when running in debug mode

DanHeidinga:
> Could add a simple test to ensure enabling the option doesn't cause a crash or throw exceptions or other undesired behavior.

I'll have to think about what to test for this as existing jvmti tests should cover the debug local mapper.

DanHeidinga:
@gacholio Based on your comment in #3946, are you suggesting this PR should also set:
`(vm->requiredDebugAttributes & J9VM_DEBUG_ATTRIBUTE_CAN_ACCESS_LOCALS)) ` as well?

gacholio:
No - if that is set, the JIT will go into FSD mode.

liqunl:
`-XX:DebugLocalMapper` will cause problems under OSR in non-FSD mode. There is an optimization in the JIT to not zero out some dead o slots, so those slots will likely contain garbage. This optimization is needed for performance in production build with production mode. Can the VM write zero to dead (in the sense of java bytecode) o-slots during OSR transition?

gacholio:
So maybe the answer is to set the debug attribute as well, forcing the JIT into FSD mode which will work with the debug mapper.

@liqunl - we could zero the whole OSR buffer before starting, so any value you don't fill in is zero, but this doesn't work if you copy garbage from the stack into the OSR buffer on the assumption that the value is dead. Doing real liveness tracking on transition is impractical and defeats the purpose of this change.

liqunl:
The JIT won't copy any value into the OSR buffer that will crash the GC after transition. Under voluntary OSR, for dead o-slot, we either omit the write or write a zero to the OSR buffer; for dead primitive slot, we omit the write or write `0xdead` to the OSR buffer. We omit the write when the slot is dead on all OSR paths.

Regarding the two approaches you proposed, neither one requires JIT change, so I'm fine with both.

gacholio:
The VM already zeroes the OSR buffer on allocation, so I suspect we'll just work.

runtime/vm/jvminit.c
```diff
@@ -2960,6 +2961,12 @@ processVMArgsFromFirstToLast(J9JavaVM * vm)
 			vm->extendedRuntimeFlags |= J9_EXTENDED_RUNTIME_POSITIVE_HASHCODE;
 		} else if (0 == strcmp(testString, VMOPT_XXDISABLEPOSITIVEHASHCODE)) {
 			vm->extendedRuntimeFlags &= ~(UDATA)J9_EXTENDED_RUNTIME_POSITIVE_HASHCODE;
+		} else if (0 == strcmp(testString, VMOPT_XXDEBUGLOCALMAPPER)) {
```
keithc-ca: This needs to be rebased: `testString` no longer exists.

DanHeidinga: Agreed - it's been blocked on me finding time to write the test case

genie-openj9:

Issue #3394, opened 2018-10-22
JTreg test fail : java/lang/invoke/lambda/LambdaStackTrace.java


ben-walsh:
Test https://github.com/ibmruntimes/openj9-openjdk-jdk11/blob/openj9/test/jdk/java/lang/invoke/lambda/LambdaStackTrace.java fails with ...

```
Actual trace did not match expected trace at frame 1
Expected frame patterns:
  0: LambdaStackTrace\\..*
  1: LambdaStackTrace.testBasic
Actual frames:
  0: LambdaStackTrace.lambda$testBasic$0(LambdaStackTrace.java:63)
  1: LambdaStackTrace$$Lambda$20.000000000000D7F0.run(Unknown Source)
java.lang.Exception: Incorrect stack frames found
	at LambdaStackTrace.verifyFrames(LambdaStackTrace.java:191)
	at LambdaStackTrace.testBasic(LambdaStackTrace.java:75)
	at LambdaStackTrace.main(LambdaStackTrace.java:53)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.base/java.lang.reflect.Method.invoke(Method.java:566)
	at com.sun.javatest.regtest.agent.MainWrapper$MainThread.run(MainWrapper.java:115)
	at java.base/java.lang.Thread.run(Thread.java:825)
```

The synthetic frame (i.e. "LambdaStackTrace$$Lambda$20.000000000000D7F0.run(Unknown Source)") should not be present in the exception stack. See OpenJDK bug https://bugs.openjdk.java.net/browse/JDK-8025636.

Test passes against Hotspot.

Log archive : [LambdaStackTraceProb.tar.gz](https://github.com/eclipse/openj9/files/2501513/LambdaStackTraceProb.tar.gz)

tajila:
@fengxue-IS can you please take a look at this

tajila:
The code responsible for creating stack traces is in `/runtime/jcl/common/jclexception.c`, specifically the `getStackTraceIterator`.

[...]

PR #3627, opened 2018-11-09
Support -XX:[+/-]ShowHiddenFrames option

fengxue-IS:
- removed unused VM runtime flags
- added hiddenFrames count info to J9GetStackTraceUserData
- added new VM option -XX:[+/-]ShowHiddenFrames
- added new VM runtime flag J9_RUNTIME_SHOW_HIDDEN_FRAMES for new option
- Add check to skip hidden frames during walkFrame() call
- Add new flag to all fillInStackTrace implementations

Fixes: #3394, Fixes: #14132


fengxue-IS:
@tajila can you take a look?

runtime/jcl/common/jclexception.c
```diff
@@ -23,13 +23,15 @@
...
 #include "j9cp.h"
+#include "j9protos.h"
 #include "jclexception.h"
 #include "jvminit.h"
 #include "objhelp.h"
 #include "omrgcconsts.h"
 #include "rommeth.h"
 #include "vm_api.h"
 
+
```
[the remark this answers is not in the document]
fengxue-IS: fixed

runtime/jcl/common/jclexception.c
```diff
@@ -97,6 +99,14 @@ getStackTraceIterator(J9VMThread * vmThread, void * voidUserData, J9ROMClass * r
 		return FALSE;
 	}
 
+	/* Skip the frame if */
+	if (J9_ARE_NO_BITS_SET(vm->runtimeFlags, J9_RUNTIME_SHOW_HIDDEN_FRAMES) && J9_ARE_ALL_BITS_SET(romClass->extraModifiers, J9AccClassAnonClass) && (NULL != romMethod) && J9ROMMETHOD_HAS_EXTENDED_MODIFIERS(romMethod)) {
```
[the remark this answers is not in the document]
fengxue-IS: updated code, moved `ShowHiddenFrames` condition to inner if statement. updated comments  to reflect the code behavior

fengxue-IS:
Senior review : @DanHeidinga please take a look

runtime/jcl/common/jclexception.c
```diff
@@ -97,6 +98,15 @@ getStackTraceIterator(J9VMThread * vmThread, void * voidUserData, J9ROMClass * r
 		return FALSE;
 	}
 
+	/* Check for extended modifiers on the ROM method */
+	if ((NULL != romMethod) && J9ROMMETHOD_HAS_EXTENDED_MODIFIERS(romMethod)) {
+		UDATA extraModifiers = getExtendedModifiersDataFromROMMethod(romMethod);
+		/* Skip the frame if @Hidden annotation is detected and -XX:+ShowHiddenFrames option has not been set */
+		if (J9_ARE_NO_BITS_SET(vm->runtimeFlags, J9_RUNTIME_SHOW_HIDDEN_FRAMES) && J9ROMMETHOD_HAS_LAMBDAFROM_HIDDEN_ANNOTATION(extraModifiers)) {
```
DanHeidinga: This allows user code to add the annotation and hide itself in stack traces.  Does this match the RI's behaviour?

runtime/oti/j9consts.h
```diff
@@ -328,9 +328,9 @@ extern "C" {
 #define J9_RUNTIME_EXTENDED_METHOD_BLOCK 0x200000
 #define J9_RUNTIME_SHUTDOWN_STARTED 0x400000
 #define J9_RUNTIME_EXIT_STARTED 0x800000
-#define J9_RUNTIME_AGGRESSIVE_VERIFICATION 0x1000000
+#define J9_RUNTIME_UNUSED_2 0x1000000
 #define J9_RUNTIME_NO_PRIORITIES 0x2000000
-#define J9_RUNTIME_DFPBD 0x4000000
+#define J9_RUNTIME_SHOW_HIDDEN_FRAMES 0x4000000
```
DanHeidinga: I see you've removed / repurposed existing bits.  Have you validated that these bits are used by existing code?  Sadly there may be code that's depending on the bit value and not the named constant....

runtime/jcl/common/jclexception.c
fengxue-IS: tested with forcing annotation on non-lambda method, this matches RI's behaviour

runtime/oti/j9consts.h
fengxue-IS: inspected all codes that accesses `vm->runtimeFlags`, didn't see any hardcoded values used with the flag. Sanity testing passed, extended testing passed before rebasing, retesting now

fengxue-IS: extended tests passed

fengxue-IS:
@DanHeidinga can I get a re-review please?

M-Davies:
Any update on this @fengxue-IS?

fengxue-IS:
@DanHeidinga Can you take a look at this please

genie-openj9:

tajila:
@DanHeidinga did you have any objections to these changes?

We have noticed some JDK18 tests that require this behaviour

fengxue-IS:
@tajila Can you please take a look.
I made a revert commit for the the original changes and build a new approach as I noticed the old logic will generate a `StackTraceElement[]` array with null entries at the end for hidden frames removed which breaks some expected behavior of `getStackTrace`. 
The new approach will filter the hidden frames when the internal stacktrace array is created during `fillInStackTrace` by specifying the newly added flag `J9_STACKWALK_SKIP_HIDDEN_FRAMES `.

Few points to note:
1. which Java version should this behavior be enabled on?
2. which APIs other than `Throwable.fillInStackTrace` should we specify `J9_STACKWALK_SKIP_HIDDEN_FRAMES` (I noticed that we have a different impl for `j.l.Thread.getStackTrace` which doesn't use `Throwable.fillInStackTrace` to initialize)
3. should we update StackWalker natives to use the new flag or keeping filtering in the iterator?

We could also continue with the old approach by building a new iterator for the first pass of `iterateStackTrace` call when calculating the number of frames (https://github.com/eclipse-openj9/openj9/blob/master/runtime/jcl/common/jclexception.cpp#L324).

tajila:
> which Java version should this behavior be enabled on?

all versions

> which APIs other than Throwable.fillInStackTrace should we specify J9_STACKWALK_SKIP_HIDDEN_FRAMES (I noticed that we have a different impl for j.l.Thread.getStackTrace which doesn't use Throwable.fillInStackTrace to initialize)

I think we should try to match the RI as much as possible

> should we update StackWalker natives to use the new flag or keeping filtering in the iterator?

Yes, lets handle this in a uniform manner

tajila:
We should also update `cInterpGetStackClassJEP176Iterator`

tajila:
Re @gacholio comment

> This isn't as simple as setting the skip bit or not - the bit is still required for security checks, etc.

I looks like we use `Class::getStackClass/Classes` for security checks which check the `J9AccMethodFrameIteratorSkip` to skip. Maybe skipping hidden/anon here as well is not correct. Perhaps we need two stackwalker flags, one for hidden/anon and one for frameiteratorskip.

runtime/oti/j9consts.h
```diff
@@ -276,7 +276,7 @@ extern "C" {
 #define J9_RUNTIME_SHUTDOWN 0x100
 #define J9_RUNTIME_XFUTURE 0x200
 #define J9_RUNTIME_PAINT_STACK 0x400
-#define J9_RUNTIME_AOT_STRIPPED 0x800
+#define J9_RUNTIME_UNUSED 0x800
```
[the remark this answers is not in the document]
fengxue-IS: This is intensional, these flags are part of the cleanup commit which removes unused flags in the VM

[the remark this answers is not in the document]
tajila: okay, thanks

tajila:
@DanHeidinga do you have any thoughts on the comment above?

fengxue-IS:
@tajila can you please take another look?
I've updated the code to have the swalk caller check `-XX:ShowHiddenFrames` option and set walk flag. Also update the StackWalker API to use the same logic/flags. Personal build passed on sanity.functional/openjdk

Locally tested on my VM that tests below passed:
```
java/lang/StackTraceElement/WithClassLoaderName.java
java/lang/reflect/MethodHandleAccessorsTest.java#id0
java/lang/reflect/IllegalArgumentsTest.java
```

runtime/jcl/common/java_lang_StackWalker.cpp
```diff
@@ -93,6 +90,12 @@ Java_java_lang_StackWalker_walkWrapperImpl(JNIEnv *env, jclass clazz, jint flags
 	walkState->walkThread = vmThread;
 	walkState->flags = J9_STACKWALK_ITERATE_FRAMES | J9_STACKWALK_WALK_TRANSLATE_PC
 			|  J9_STACKWALK_INCLUDE_NATIVES | J9_STACKWALK_VISIBLE_ONLY;
+	/* If -XX:+ShowHiddenFrames and StackWalker.SHOW_HIDDEN_FRAMES option has not been set, skip hidden method frames */
+	if (J9_ARE_NO_BITS_SET(vm->runtimeFlags, J9_RUNTIME_SHOW_HIDDEN_FRAMES)
+	&& J9_ARE_NO_BITS_SET((UDATA)flags, SHOW_HIDDEN_FRAMES)
```
tajila, line 95: Just to confirm if `Stackwalker.Option::SHOW_HIDDEN_FRAMES` is requested ojdk doesn't show hidden frames unless `-XX:+ShowHiddenFrame` is specified as well?

tajila:
@gacholio please review these changes

runtime/oti/j9consts.h
gacholio, line 279: The naming convention for flags like this includes the value, i.e. `J9_RUNTIME_UNUSED_0x800`. Same comment for the unused below.

gacholio:
Was our behaviour incorrect before (assuming the option defaults to off)?

runtime/oti/j9consts.h
fengxue-IS: fixed

fengxue-IS:
> Was our behaviour incorrect before (assuming the option defaults to off)?

OpenJ9 by default doesn't skip hidden frames, though behavior in terms of handling hidden frames is unspecified, the reason for this change is due to many OJDK tests depends on this behavior so we are matching the RI's behavior

gacholio:
For consistency, please add the print to the start of the stack walker for the new flag bit.

gacholio:
And please squash the commits once that change is made.

fengxue-IS:
fixed all review comments as suggested, re-compiled locally and passed JTreg test associated.

gacholio:
jenkins test sanity zlinux jdk17

gacholio (approved):

pshipton:
@fengxue-IS please create a https://github.com/eclipse-openj9/openj9-docs issue and PR to update the user guide. The new option also needs to be mentioned in the OpenJ9 release notes. https://github.com/eclipse-openj9/openj9/blob/master/doc/release-notes/0.31/0.31.md

PR #3628, opened 2018-11-09
Correct JVM_ARCH_DIR for ppc64le

keithc-ca:
For little-endian Linux on POWER, `JVM_ARCH_DIR` should be `"ppc64le"`.

Also, fix compile errors if `DEBUG` is defined in `jvm.c`.


pshipton:
Note JVM_ARCH_DIR / J9VM_LIB_ARCH_DIR is only used in relation to a "classic" directory.

PR #3639, opened 2018-11-12
Fix invalid class pointer

dmitry-ten:
In `runFEMacro`, when invoke_SpreadHandle_arrayarg is encountered, leaf class pointer wasn't being assigned in JITaaS mode. Fix by requesting this pointer from the client.
In addition, refactor code to reduce the number of messages for this
case by 2


dmitry-ten:
@mpirvu could you review, please?

dmitry-ten:
I don't see any comments, but I assume you mean using `const std::stirng &`, instead of `const std::string`.

PR #3648, opened 2018-11-12
Truncate boolean writes in JNI and Unsafe

tajila:
Truncate boolean writes in JNI and Unsafe

### DIFF
--- a/runtime/oti/UnsafeAPI.hpp
+++ b/runtime/oti/UnsafeAPI.hpp
@@ -421,7 +421,7 @@ public:
 	static VMINLINE void
 	putBoolean(J9VMThread *currentThread, MM_ObjectAccessBarrierAPI *objectAccessBarrier, j9object_t object, UDATA offset, bool isVolatile, U_8 value)
 	{
-		put32(currentThread, objectAccessBarrier, object, offset, isVolatile, 0, false, (I_32)value);
+		put32(currentThread, objectAccessBarrier, object, offset, isVolatile, 0, false, (I_32)(value & 1));
 	}
 
 	static VMINLINE I_16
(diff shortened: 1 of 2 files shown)


tajila:
The static boolean JNI writes already behave in accordance with the spec

tajila:
@DanHeidinga can you please review this

runtime/vm/BytecodeInterpreter.hpp
```diff
@@ -3280,7 +3280,7 @@ done:;
 		buildInternalNativeStackFrame(REGISTER_ARGS);
 		updateVMStruct(REGISTER_ARGS);
 		_currentThread->privateFlags2 |= J9_PRIVATE_FLAGS2_UNSAFE_HANDLE_SIGBUS;
-		VM_UnsafeAPI::putBoolean(_currentThread, &_objectAccessBarrier, obj, offset, isVolatile, value);
+		VM_UnsafeAPI::putBoolean(_currentThread, &_objectAccessBarrier, obj, offset, isVolatile, value & 1);
```
keithc-ca: Shouldn't this be `(0 != value)` rather than `value & 1`?

tajila: The spec says `the int value is narrowed by taking the bitwise AND of value and 1, resulting in value`.   
  
I don't see the difference between either approach since C++ bools will be 0 or 1.

However, the current implementation is consistent with what we did earlier in the interpreter.

keithc-ca: Perhaps this change should be made in UnsafeAPI.hpp so all users benefit?

gacholio: I think there only is one caller, but putting the mask in putBoolean instead makes sense.

tajila:
@keithc-ca @gacholio I've updated the PR

gacholio:

PR #3650, opened 2018-11-13

DanHeidinga:

Add an assert for this rather than a dynamic allocation as the likelihood
of ever needing to fallback to the allocation is so low the extra code
isn't worth writing (given it will never be tested)

issue: #3392
issue: #3609

### DIFF
--- a/runtime/jcl/common/system.c
+++ b/runtime/jcl/common/system.c
@@ -412,9 +412,15 @@ jstring getEncoding(JNIEnv *env, jint encodingType)
 #if defined(J9VM_JCL_SE11)
 		{
 			UDATA handle = 0;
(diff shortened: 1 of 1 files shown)


DanHeidinga:
Jenkins test sanity all jdk11

DanHeidinga:
Manually tested on windows and mac - both work.

jdekonin:
jenkins test sanity osx jdk11

jdekonin:
jenkins test sanity osx jdk11

New osx machine added caused the previous failure. Respin again, looks like network connectivity to machine dropped.

jdekonin:
jenkins test sanity osx jdk11

pshipton:
The osx build succeeded building, and ran the sanity tests until it hung running the abort test, which I understand is a known osx test problem.

PR #3657, opened 2018-11-13
Deprecate use of `generate(RXY|RSY|SIY)Instruction` API

fjeremic:
Now that eclipse/omr#3113 has been merged we can safely deprecate the
use of long displacement APIs in favor or RX,RS,SI APIs since the long
displacement versions map right down to the non-long displacement APIs.

This eliminates all uses of such long displacement APIs in all
downstream projects so that we can fold the API away entirely at the
OMR level and simplify the generation of many instructions in the Z
code generator without having the user worry whether they need long
displacement support or not.


fjeremic:

fjeremic:

fjeremic:
Ah just realized we need to wait for OMR-Acceptance to promote with the change from eclipse/omr#3113. I'm expecting the build to fail, so I'll mark this as WIP until we get the promotion and I'll give the green light and re-launch the builds then.

fjeremic:
We just had an OMR-Acceptance promotion with the right SHA. I had to rebase due to a conflict in J9TreeEvaluator.cpp though. Unleashing builds:

PR #3662, opened 2018-11-13
Reduce number of getOSRFrameSizeInBytes messages

satbir-ca:
ROMMethod are already cached so acquire the ROMMethod for the J9Method from the J9MethodMap.
New method osrFrameSizeRomMethod takes ROMMethod as input and return the frame size.

### DIFF
--- a/runtime/codert_vm/decomp.cpp
+++ b/runtime/codert_vm/decomp.cpp
@@ -1765,6 +1765,20 @@ UDATA
 osrFrameSize(J9Method *method)
 {
 	J9ROMMethod *romMethod = J9_ROM_METHOD_FROM_RAM_METHOD(method);
+	return osrFrameSizeRomMethod(romMethod);
+}
+
+
+/**
+ * Compute the number of bytes required for a single OSR frame.
+ *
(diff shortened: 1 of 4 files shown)


runtime/codert_vm/decomp.cpp
```diff
@@ -1765,6 +1765,20 @@ UDATA
 osrFrameSize(J9Method *method)
 {
 	J9ROMMethod *romMethod = J9_ROM_METHOD_FROM_RAM_METHOD(method);
+	return osrFrameSizeRomMethod(romMethod);
+}
```
mpirvu: To minimize discrepancies between jitaas and mainline this code needs to be delivered to the master branch.

runtime/compiler/env/j9method.cpp
```diff
@@ -792,6 +792,15 @@ TR_J9MethodBase::osrFrameSize(J9Method* j9Method)
    {
    if (auto stream = TR::compInfoPT->getStream())
       {
+         {
```
mpirvu: I don't see this method being used anywhere, so we should delete it from the master branch as well.

PR #3664, opened 2018-11-13
Reduce number of getOSRFrameSizeInBytes messages

satbir-ca:
ROMMethod are already cached so acquire the ROMMethod for the J9Method from the J9MethodMap.
New method osrFrameSizeRomMethod takes ROMMethod as input and return the frame size.

### DIFF
--- a/runtime/compiler/env/VMJ9Server.cpp
+++ b/runtime/compiler/env/VMJ9Server.cpp
@@ -19,7 +19,7 @@
  *
  * SPDX-License-Identifier: EPL-2.0 OR Apache-2.0 OR GPL-2.0 WITH Classpath-exception-2.0 OR LicenseRef-GPL-2.0 WITH Assembly-exception
  *******************************************************************************/
-
+#include "oti/jitprotos.h"
 #include "rpc/J9Server.hpp"
 #include "VMJ9Server.hpp"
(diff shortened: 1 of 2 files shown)


satbir-ca:

runtime/codert_vm/decomp.cpp
```diff
@@ -1765,6 +1765,20 @@ UDATA
 osrFrameSize(J9Method *method)
 {
 	J9ROMMethod *romMethod = J9_ROM_METHOD_FROM_RAM_METHOD(method);
+	return osrFrameSizeRomMethod(romMethod);
```
mpirvu: This code needs to be delivered in the master branch.

runtime/compiler/env/VMJ9Server.cpp
```diff
@@ -519,6 +519,15 @@ TR_J9ServerVM::isClassInitialized(TR_OpaqueClassBlock * clazz)
 UDATA
 TR_J9ServerVM::getOSRFrameSizeInBytes(TR_OpaqueMethodBlock * method)
    {
+      {
```
mpirvu: ... while this code needs to be delivered only in the jitaas branch

satbir-ca:
@mpirvu This commit depend on the master branch PR #3674.

PR #3665, opened 2018-11-13
Fix for incorrect kind of AllocPrefetchSnippet on X86

0dvictor:
X86AllocPrefetchSnippet incorrectly reports itself as IsJNIPause,
fixing.


andrewcraik:

PR #3668, opened 2018-11-14
Change how compilation dispatch works in JITaaS

mpirvu:
Currently, the dispatcher thread at the JITaaS server performs too much
work which could hinder connections from other clients. One of the main
problems is that the listener thread, which accepts incoming connection
requests, also perform a blocking read to determine the parameters for
the incoming compilation request. Thus, while waiting for this request
to come, no new connections can be established.
Another problem is that processing the unloaded classes could take some
time (also done under a monitor) which again could delay new connections.

The proposal is to queue just the stream that is created as a consequence
of the connection being accepted. Then, the reading of the compilation
request sent by the client could be handled by a dedicated compilation
thread. This way, the job of the dispatcher thread is much simplified.


mpirvu:
I eliminated the two obsolete comments.

PR #3678, opened 2018-11-14
Truncate JNI boolean returns

tajila:
Truncate JNI boolean returns

### DIFF
--- a/runtime/oti/VMHelpers.hpp
+++ b/runtime/oti/VMHelpers.hpp
@@ -1381,7 +1381,7 @@ exit:
 			}
 			break;
 		case J9NtcBoolean:
-			*returnStorage = (UDATA)(U_8)*returnStorage;
+			*returnStorage = ((UDATA)(U_8)*returnStorage) & 1;
 			break;
 		case J9NtcByte:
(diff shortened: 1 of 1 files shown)


runtime/oti/VMHelpers.hpp
```diff
@@ -1381,7 +1381,7 @@ class VM_VMHelpers
 			}
 			break;
 		case J9NtcBoolean:
-			*returnStorage = (UDATA)(U_8)*returnStorage;
+			*returnStorage = (UDATA)(U_8)(*returnStorage & 1);
```
DanHeidinga: It's clearer if the `& 1` is bubbled out of the read/cast brackets: `((UDATA)(U_8)*returnStorage) & 1;`

tajila:
@DanHeidinga I've made the change

tajila:
@andrewcraik The JIT will need similar changes. Likewise with unsafe boolean stores

DanHeidinga:
Discussed with @andrewcraik & @gacholio -  the prefered approach is to have the JIT update the code generated for the directToJNI case to include the `& 1` as this covers all cases and adds minimal instructions only in the boolean return case.

PR #3683, opened 2018-11-14
Enable Linux PPC compile with gcc-7

vsebe:
- update uma_make_cmd_ppc_gcc_cxx to the build environment gcc (gcc
v7.3.0)
- update graph build requirement

[ci skip]

### DIFF
--- a/buildspecs/linux_ppc-64.spec
+++ b/buildspecs/linux_ppc-64.spec
@@ -40,7 +40,7 @@ SPDX-License-Identifier: EPL-2.0 OR Apache-2.0 OR GPL-2.0 WITH Classpath-excepti
 		<property name="graph_commands.chroot" value=""/>
 		<property name="graph_commands.unix.remote_host" value=""/>
 		<property name="graph_datamines" value="commands.unix.datamine,site-ottawa.datamine,use.local.datamine"/>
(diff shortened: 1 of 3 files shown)


buildspecs/linux_ppc-64.spec
```diff
@@ -96,7 +96,7 @@ SPDX-License-Identifier: EPL-2.0 OR Apache-2.0 OR GPL-2.0 WITH Classpath-excepti
 		<property name="uma_make_cmd_cxx_exe_ld" value="$(CC)"/>
 		<property name="uma_make_cmd_dll_ld" value="$(CC)"/>
 		<property name="uma_make_cmd_exe_ld" value="$(CC)"/>
-		<property name="uma_make_cmd_ppc_gcc_cxx" value="gcc-4.6"/>
```
pshipton: This should be deleted.

vsebe: If  uma_make_cmd_ppc_gcc_cxx is deleted then PPC_GCC_CXX would be set to xlc (mkconstants.mk.ftl). Shouldn't BytecodeInterpreter be compiled with gcc?

pshipton: yup

PR #3684, opened 2018-11-14
Add bytecode support for Q Type getter and setters

acrowthe:
- Fix two cases where argCount was incorrect on Q Types
- Mark Q types as objects to prevent putField and getFieldLogic issues

### DIFF
--- a/runtime/bcutil/ClassFileOracle.cpp
+++ b/runtime/bcutil/ClassFileOracle.cpp
@@ -839,11 +839,20 @@ ClassFileOracle::computeSendSlotCount(U_16 methodIndex)
 			while ((index < count) && ('[' == bytes[index])) {
 				++index;
 			}
-			if ((index >= count) || ('L' != bytes[index])) {
+			if ((index >= count)
+#if defined(J9VM_OPT_VALHALLA_VALUE_TYPES)
+				|| (('L' != bytes[index]) && ('Q' != bytes[index]))
+#else
+				|| ('L' != bytes[index])
+#endif /* J9VM_OPT_VALHALLA_VALUE_TYPES */
+			) {
 				break;
 			}
 			/* fall through */
 		case 'L':
+#if defined(J9VM_OPT_VALHALLA_VALUE_TYPES)
+		case 'Q':
+#endif /* J9VM_OPT_VALHALLA_VALUE_TYPES */
 			++index;
 			while ((index < count) && (';' != bytes[index])) {
 				++index;
--- a/runtime/util/sendslot.c
+++ b/runtime/util/sendslot.c
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 1991, 2014 IBM Corp. and others
+ * Copyright (c) 1991, 2018 IBM Corp. and others
  *
  * This program and the accompanying materials are made available under
  * the terms of the Eclipse Public License 2.0 which accompanies this
@@ -37,13 +37,20 @@ getSendSlotsFromSignature(const U_8* signature)
 		case '[':
 			/* skip all '['s */
 			for (i++; signature[i] == '['; i++);
-			if (signature[i] == 'L') {
+			if ((signature[i] == 'L')
+#if defined(J9VM_OPT_VALHALLA_VALUE_TYPES)
+				|| (signature[i] == 'Q')
+#endif /* J9VM_OPT_VALHALLA_VALUE_TYPES */
+			) {
 				/* FALL THRU */
 			} else {
 				sendArgs++;
 				break;
 			}
 		case 'L':
+#if defined(J9VM_OPT_VALHALLA_VALUE_TYPES)
+		case 'Q':
+#endif /* J9VM_OPT_VALHALLA_VALUE_TYPES */
 			for (i++; signature[i] != ';'; i++);
 			sendArgs++;
 			break;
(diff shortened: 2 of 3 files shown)


tajila:
wrap your changes with the `J9VM_OPT_VALHALLA_VALUE_TYPES` macro. ie `#if defined(J9VM_OPT_VALHALLA_VALUE_TYPES)`

runtime/bcutil/ClassFileOracle.cpp
```diff
@@ -835,11 +835,18 @@ ClassFileOracle::computeSendSlotCount(U_16 methodIndex)
 			while ((index < count) && ('[' == bytes[index])) {
 				++index;
 			}
+#if defined(J9VM_OPT_VALHALLA_VALUE_TYPES)
+			if ((index >= count) || !(('L' == bytes[index]) || ('Q' == bytes[index]))) {
```
tajila: It is easier to read as `(index >= count) || ('L' != bytes[index]) || ('Q' != bytes[index])`

runtime/util/sendslot.c
```diff
@@ -44,6 +44,9 @@ getSendSlotsFromSignature(const U_8* signature)
 				break;
```
tajila: Need to also address the array case above

tajila:
Can you also please test this with a Q-array

runtime/bcutil/ClassFileOracle.cpp
acrowthe: I believe the correct logic, in a more readable way, would be:
`(index >= count) || (('L' != bytes[index]) && ('Q' == bytes[index]))`
The desired logic is as follows:
> If (index >= count) or (the next character isn't a Q or L)

acrowthe:
Fails with Q arrays. Tests I have used are located at https://github.com/acrowthe/openj9/tree/qtype-fixes-1-tests, to be cherry-picked.

acrowthe:
Q type arrays will be looked into in a different PR

acrowthe:
Build failing for `win_x86-64_cmprssptrs`

tajila:
@gacholio Please review these changes.
  
This PR contains the minimal set of changes required to support fields with Q signatures thus enabling us to start making progress on field flattening. 

A follow on PR will be created to support Qarrays arrays. 

A more comprehensive PR is currently being worked on to address Qtypes for JNI, reflection, verifier, method trace, etc.

runtime/bcutil/ClassFileOracle.cpp
```diff
@@ -839,11 +839,18 @@ ClassFileOracle::computeSendSlotCount(U_16 methodIndex)
 			while ((index < count) && ('[' == bytes[index])) {
 				++index;
 			}
+#if defined(J9VM_OPT_VALHALLA_VALUE_TYPES)
+			if ((index >= count) || (('L' != bytes[index]) && ('Q' != bytes[index]))) {
+#else
 			if ((index >= count) || ('L' != bytes[index])) {
+#endif /* J9VM_OPT_VALHALLA_VALUE_TYPES */
```
gacholio: J9 prefers a single open brace outside the ifdef to avoid confusing formatting/parsing tools. Given that, the Q case could be added undef ifdef without the need to duplicate the common code:

```
if (common)
#ifdef VT
|| VT code
#endif
{
```

with appropriate formatting.

acrowthe: Due to the nature of the conditions, I don't think the condition can be split into `(common)` and `VT code`. This is because the condition needs `(not L) AND (not Q)`.

I can refactor the code to the below code. Is that sufficient?
```
			if ((index >= count) || (('L' != bytes[index]) && ('Q' != bytes[index]))) {
#else
			if ((index >= count) || ('L' != bytes[index]))
#endif /* J9VM_OPT_VALHALLA_VALUE_TYPES */
			{
```

acrowthe: Another alternative (lets call this 2, and the previous 1):
```
			if (index >= count)
				|| (('L' != bytes[index]) && ('Q' != bytes[index]))
#else
				|| ('L' != bytes[index])
#endif /* J9VM_OPT_VALHALLA_VALUE_TYPES */
			{
```
This would keep the `if` outside the `if defined`'s

gacholio: Sorry, didn't read the bracketting properly.

gacholio: I like option 2 better. @tajila ?

tajila: Im fine with option 2

acrowthe: Option 2 has been implemented

acrowthe:
Encountering issues building with the suggested changes: Looking into that now.

runtime/bcutil/ClassFileOracle.cpp
acrowthe: Looks like code in the format
```
f (index == 1)
	|| (index == 2)
{
```
does not compile.
Adding `()` around the entire conditional fixes this:

```
if ((index == 1)
	|| (index == 2)
) {
```
Fixed option 2 is now (with shortened whitespace)
```
	if ((index >= count)
		|| (('L' != bytes[index]) && ('Q' != bytes[index]))
#else
		|| ('L' != bytes[index])
#endif /* J9VM_OPT_VALHALLA_VALUE_TYPES */
	) {
		break;
	}
```

gacholio:

PR #3687, opened 2018-11-15
Fix incorrect map argument

gacholio:
For accurate bytecode information,
getCurrentByteCodeIndexAndIsSameReceiver needs the inline map, not the
stack map. Fix exception event and decompilation to use the proper map,
and update the variable name in other code to be consistent.

Closes: #3373

[ci skip]


gacholio:
Testing on AIX since that's where the problem showed up (inexplicably) - failure was about 1/1000 so I'm not expecting an issue here.

PR #3689, opened 2018-11-15
Set KEEP_REPORTDIR false for tests archiving

AdamBrousseau:
- KEEP_REPORTDIR is a flag in the test jobs that controls
  whether or not we delete the successful test reports before
  we archive the reports.
- By default the setting is true (don't delete).
- Setting it to false will delete successful repots.
- This will reduce the size of the test artifact.
- Set to false for both Test jobs that archive to Artifactory and Master.

[skip ci]
Related #3589


AdamBrousseau:
cc @smlambert @renfeiw 
We need to determine if we want to flip this flag for all test jobs or just the ones that upload to Jenkins.

AdamBrousseau:
@renfeiw for review

AdamBrousseau:
@llxia for merge

PR #3692, opened 2018-11-15
J9::Power::TreeEvaluator::VMarrayCheckEvaluator() spilling

klangman:
J9::Power::TreeEvaluator::VMarrayCheckEvaluator() is not correctly
excluding the use of GPR0 for one register which results in
spilling inside internal control flow.


gita-omr:

gita-omr:
This is the same issue that @aviansie-ben found, he has some ideas how to ASSERT at compile time if that happens.

klangman:
> This is the same issue that @aviansie-ben found, he has some ideas how to ASSERT at compile time if that happens.

The fix should be in before the 18_04u2 release, maybe the assert should be a separate PR if it would delay the fix.

aviansie-ben:
The issue I found was the same _type_ of issue (missing GPR0 exclude), but it wasn't in this specific evaluator and has already been corrected. The assert that I've been working on to check for missing GPR0 excludes isn't going to be ready for a while, as I've been sidetracked with other work. However, it is in progress and should hopefully be able to catch future issues of this nature at compile-time.

gita-omr:
I meant maybe not exactly the same place, but the same issue of GPR0 not being excluded in ICF.

PR #3695, opened 2018-11-15
Enable gcc-7 on Linux x86-32 platform

vsebe:

[ci skip]


vsebe:
see https://github.com/eclipse/openj9/pull/3695

PR #3696, opened 2018-11-15
Deleted known_hosts file

cwesMills:
The known_hosts file was used for the docker containers
to make a copy of. Now that there is a job to build docker
images and automatically generate the known_hosts file, there
is no more need for this file.


cwesMills:
@AdamBrousseau

AdamBrousseau:
@smlambert for merge

charliegracie:
Hmm interesting that this file does not have a copyright header. It *should* in my opinion.  Also I do not think this is the right approach. The jenkins job that builds the docker image should be providing this file and not the repo IMHO.

AdamBrousseau:
If the Jenkins job is providing the file, wouldn't this just push the file to a different folder in the source. Unless we put it into Jenkins variable or credential store. That would be the opposite direction of trying to get as much config into code as possible.
I'm fine with adding the Copyright header.

DanHeidinga:
I agree with @charliegracie - this change doesn't belong in OpenJ9 as it introduces new IBMisms into the source.

We need to find a different approach than adding `github.ibm.com` keys to OpenJ9 repo.

AdamBrousseau:
@cwesMills based on how the job works I think this file can be removed now?

PR #3706, opened 2018-11-16
Validate class chain for defining class of static

jdmpapin:
Since the defining class of a static doesn't escape from the point at
which we request the validation, we're able to get away with checking
only the ROM class, rather than the entire class chain. However, there
are a few issues with this:

- It's different from every other class validation.

- We may need to validate the class chain anyway, if the same class is
  encountered in any other way during compilation, in which case the ROM
  class validation is redundant.

Furthermore, it's unlikely that validating the class chain will cause
many additional load failures.

As such, queries for the defining class of a static will now request a
class chain validation for that class.

### DIFF
--- a/runtime/compiler/runtime/SymbolValidationManager.cpp
+++ b/runtime/compiler/runtime/SymbolValidationManager.cpp
@@ -153,19 +153,14 @@ TR::SymbolValidationManager::getBaseComponentClass(TR_OpaqueClassBlock *clazz, i
 
 
 void *
-TR::SymbolValidationManager::storeClassChain(TR_J9VMBase *fej9, TR_OpaqueClassBlock *clazz, bool isStatic)
+TR::SymbolValidationManager::storeClassChain(TR_J9VMBase *fej9, TR_OpaqueClassBlock *clazz)
    {
    void *classChain = NULL;
    bool validated = false;
 
    classChain = fej9->sharedCache()->rememberClass(clazz);
    if (classChain)
-      {
-      if (isStatic)
-         validated = addRomClassRecord(clazz);
-      else
-         validated = addClassChainRecord(clazz, classChain);
-      }
+      validated = addClassChainRecord(clazz, classChain);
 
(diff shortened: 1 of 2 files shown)


jdmpapin:
I've rebased this so that it's more up to date

andrewcraik:
Jenkins test sanity xlinux,win.plinux jdk8

andrewcraik:
Jenkins test sanity xlinux,win,plinux jdk8

andrewcraik:
Jenkins test sanity xlinux jdk8,jdk11

andrewcraik:
So ordinarily I would want to run a set of test more likely to stress AOT but pending resolution of #3774 we don't have a better test bucket. @jdmpapin can you sketch the testing you did beyond sanity to ensure this has been exercised?

jdmpapin:
Sorry for the delayed response.

I've run the IBM SDK for Java's internal CI testing, which includes more tests for AOT.

I've run `java -version` with

    forceAOT,disableAsyncCompilation,initialOptLevel=warm,inhibitRecompilation

using low counts and various inlining settings:
- `count=5,dontInline={*}`
- `count=5`
- `count=20`
- `count=20,tryToInline={*}`

Each set of options was run twice: once to AOT compile and store, and once to AOT load. Then each set of options was run twice again with `-Xtune:virtualized`.

I've also run a nontrivial app server benchmark (not for measurement, but just to see that it runs without issue). Again I ran twice with and twice without `-Xtune:virtualized` for each of a variety of options (all including `forceAOT`):
- `count=0`
- `count=5`
- `count=20`
- `useHigherMethodCounts`

Finally, I found a method in which there was a class first encountered via `DefiningClassFromCP`. Before this change, the `DefiningClassFromCP` generated a `RomClass` record. The same class was subsequently found in a different way, necessitating a `ClassChain` record. After this change, the `DefiningClassFromCP` generated the `ClassChain` record in the first place, as expected.

PR #3708, opened 2018-11-16
Slightly reword point in SVM documentation

jdmpapin:
"Provenance" seems to have been intended here.


andrewcraik:
This is a single word change in a markdown file. @dsouzai approves so I'm going to merge this.

PR #3709, opened 2018-11-16
Print _class for ClassChainRecord in SVM

jdmpapin:
The symbol validation manager (SVM) class chain record has a class field
but has not been printing it in its tracing. This field is now shown.


andrewcraik:
We should do a sanity before merging since this is a code change.

andrewcraik:

PR #3711, opened 2018-11-16
Concurrent Scavenger yield critera

amicic:
Introduce a criteria to yield during concurrent phase of a CS cycle:
obey exclusive VM acceess request.

This will clearly cover explicit requests from thread party (like JVMTI
or System GC).

But also this will cover an indirect request to complete concurrent
phase if hybrid allocate/survivor space is (in rare and unfortunat case)
exhausted -> AF is raised by a mutator, which will request exclusive VM
access. It is indeed prefered to complete cycle in STW mode in this
case, because we have more GC threads at disposal than in concurrent
phase.


amicic:
Related, by not dependent on: https://github.com/eclipse/omr/pull/3205

dmitripivkine:

PR #3705, opened 2018-11-15
Re-enable ppc64le tests

keithc-ca:
Re-enable tests that were disabled for ibmruntimes/openj9-openjdk-jdk8#190.

PR #3715, opened 2018-11-16
Remove requirement for 'spec.flags.build_j2se' in jep178 tests

keithc-ca:
The jep178 tests were failing on Linux ppc64le (using gcc) because the associated build specs specified `build_j2se = false` . The tests don't need to depend on that flag. This change removes the requirement that flag be set and adjusts `UMA_EXE_POSTFIX_FLAGS` to suit gcc.

### DIFF
--- a/runtime/tests/redirector/jep178/module.xml
+++ b/runtime/tests/redirector/jep178/module.xml
@@ -1,31 +1,27 @@
 <?xml version="1.0" encoding="UTF-8"?>
-
 <!--
-   Copyright (c) 2014, 2017 IBM Corp. and others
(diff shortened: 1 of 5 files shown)


keithc-ca:
There are a number of issues related to jep178 tests, but I couldn't find one specific to the ppc64le issue.

keithc-ca:
An example failure was seen for #3705: https://ci.eclipse.org/openj9/job/PullRequest-Extended-JDK8-linux_ppc-64_cmprssptrs_le-OpenJ9/151.

pshipton:
jenkins test extended plinux

PR #3730, opened 2018-11-19
Reduce compinfo messages

satbir-ca:
Caching the JNINative status in resolve method info
and this eliminate the isJNINative message.

ROM Class is already cached. Using the cached info
to reduce the getMethodBytecodeSize and isJSR292 messages


satbir-ca:

runtime/compiler/env/j9methodServer.cpp
```diff
@@ -105,8 +105,7 @@ TR_ResolvedJ9JITaaSServerMethod::isJNINative()
    {
    if (supportsFastJNI(_fe))
       return _jniTargetAddress != NULL;
-   _stream->write(JITaaS::J9ServerMessageType::ResolvedMethod_isJNINative, _remoteMirror);
-   return std::get<0>(_stream->read<bool>());
+   return _isJNINative;
```
mpirvu: If we already cache the right result, do we need the `supportsFastJNI` check above?

mpirvu:
Please rephrase the first sentence from the commit message to something like:
"Cache the JNINative status in resolved method info
to eliminate isJNINative messages."

runtime/compiler/env/j9methodServer.cpp
satbir-ca: The supportsFastJNI is for TR_TARGET_S390 TR_TARGET_X86 and TR_TARGET_POWER. If it is for any of these target then  _jniTargetAddress is used to make decision and for other target  isJNINative is used.

satbir-ca:
@mpirvu updated the comment

Issue #3582, opened 2018-11-07
Test-sanity.system-JDK8 MauveMultiThreadLoadTest_0 BigInteger


pshipton:
https://ci.eclipse.org/openj9/job/Test-sanity.system-JDK8-linux_ppc-64_cmprssptrs_le/135
```
LT  07:17:05.028 - First failure detected by thread: load-0. Running test: Mauve[gnu.testlet.java.math.BigInteger.add]. Creating java dumps.
LT  stderr JVMDUMP034I User requested Heap dump using '/home/jenkins/jenkins-agent/workspace/Test-sanity.system-JDK8-linux_ppc-64_cmprssptrs_le/openjdk-tests/TestConfig/test_output_15415746471851/MauveMultiThreadLoadTest_0/20181107-071649-MauveMultiThreadLoadTest/results/firstfailure.heapdump.20181107.071705.31257.0001.phd' through com.ibm.jvm.Dump.heapDumpToFile
LT  stderr JVMDUMP010I Heap dump written to /home/jenkins/jenkins-agent/workspace/Test-sanity.system-JDK8-linux_ppc-64_cmprssptrs_le/openjdk-tests/TestConfig/test_output_15415746471851/MauveMultiThreadLoadTest_0/20181107-071649-MauveMultiThreadLoadTest/results/firstfailure.heapdump.20181107.071705.31257.0001.phd
LT  stderr JVMDUMP034I User requested Java dump using '/home/jenkins/jenkins-agent/workspace/Test-sanity.system-JDK8-linux_ppc-64_cmprssptrs_le/openjdk-tests/TestConfig/test_output_15415746471851/MauveMultiThreadLoadTest_0/20181107-071649-MauveMultiThreadLoadTest/results/firstfailure.javacore.20181107.071705.31257.0002.txt' through com.ibm.jvm.Dump.javaDumpToFile
LT  stderr JVMDUMP010I Java dump written to /home/jenkins/jenkins-agent/workspace/Test-sanity.system-JDK8-linux_ppc-64_cmprssptrs_le/openjdk-tests/TestConfig/test_output_15415746471851/MauveMultiThreadLoadTest_0/20181107-071649-MauveMultiThreadLoadTest/results/firstfailure.javacore.20181107.071705.31257.0002.txt
[...]

PR #3733, opened 2018-11-19
NULLCHK argument of dead call to BigDecimal or BigInteger

hzongaro:
If the result of a call to one of the BigDecimal or BigInteger arithmetic operations is removed because its result is not used, ensure a NULLCHK of its argument is still preserved as a side effect.

Fixes:  #3582

### DIFF
--- a/runtime/compiler/optimizer/J9Simplifier.cpp
+++ b/runtime/compiler/optimizer/J9Simplifier.cpp
@@ -479,28 +479,72 @@ TR::Node *
 J9::Simplifier::simplifyaCallMethods(TR::Node * node, TR::Block * block)
    {
    if ((node->getOpCode().isCallDirect()) && !node->getSymbolReference()->isUnresolved() &&
-       (node->getSymbol()->isResolvedMethod()) &&
-       ((node->getSymbol()->getResolvedMethodSymbol()->getRecognizedMethod() == TR::java_math_BigDecimal_valueOf) ||
-        (node->getSymbol()->getResolvedMethodSymbol()->getRecognizedMethod() == TR::java_math_BigDecimal_add) ||
-        (node->getSymbol()->getResolvedMethodSymbol()->getRecognizedMethod() == TR::java_math_BigDecimal_subtract) ||
-        (node->getSymbol()->getResolvedMethodSymbol()->getRecognizedMethod() == TR::java_math_BigDecimal_multiply) ||
-        (node->getSymbol()->getResolvedMethodSymbol()->getRecognizedMethod() == TR::java_math_BigInteger_add) ||
-        (node->getSymbol()->getResolvedMethodSymbol()->getRecognizedMethod() == TR::java_math_BigInteger_subtract) ||
-        (node->getSymbol()->getResolvedMethodSymbol()->getRecognizedMethod() == TR::java_math_BigInteger_multiply)) &&
-       (node->getReferenceCount() == 1) &&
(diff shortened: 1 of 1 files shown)


hzongaro:
Marking this as WIP while I run testing to ensure this doesn't cause any problems, but I believe it fixes the issue.

andrewcraik:

runtime/compiler/optimizer/J9Simplifier.cpp
```diff
@@ -479,28 +479,73 @@ TR::Node *
...
+                  comp()->getSymRefTab()->findOrCreateNullCheckSymbolRef(
+                                                comp()->getMethodSymbol());
+            TR::TreeTop::create(comp(), _curTree,
+                  TR::Node::createWithSymRef(node, TR::NULLCHK, 1,
+                        TR::Node::create(node, TR::PassThrough, 1, secondChild),
+                                         nullCheckSR));
+            _alteredBlock = true;
+            }
+
+         int i;
```
andrewcraik: can we please move the declaration of i into the for loop to limit its scope since you aren't using it outside the loop

runtime/compiler/optimizer/J9Simplifier.cpp
```diff
@@ -479,28 +479,73 @@ TR::Node *
...
-        (node->getSymbol()->getResolvedMethodSymbol()->getRecognizedMethod() == TR::java_math_BigInteger_subtract) ||
-        (node->getSymbol()->getResolvedMethodSymbol()->getRecognizedMethod() == TR::java_math_BigInteger_multiply)) &&
-       (node->getReferenceCount() == 1) &&
-       performTransformation(comp(), "%sRemoved dead BigDecimal/BigInteger call node [" POINTER_PRINTF_FORMAT "]\n", optDetailString(), node))
+       (node->getSymbol()->isResolvedMethod()))
       {
-      TR::Node *firstChild = node->getFirstChild();
-      anchorChildren(node, _curTree);
+      bool isSimplifiableMethod = false;
+      bool isIntOrDecimalNonDivArithOp = false;
```
andrewcraik: this is more rightly called isInstanceMethod since instance methods are the ones which need to NULLCHK the receiver.

hzongaro: @andrewcraik Andrew, as we discussed, I changed the name to `requiresArgNULLCHK` to clarify why we're tracking these particular methods.  Fixed in ed97fb6168f0220071a8694fe1b19cc919d3af73

If you're OK with this change, I can squash the two commits prior to delivery.

runtime/compiler/optimizer/J9Simplifier.cpp
hzongaro: Fixed in https://github.com/eclipse/openj9/commit/ed97fb6168f0220071a8694fe1b19cc919d3af73

andrewcraik:
@hzongaro to keep the convention uniform can you move the comment from the end of the bool to above it? Once done I'm happy with the rest.

andrewcraik:
@hangshao0 ok that looks good please squash - we don't need a new sanity because the second commit is an obvious rename and fix-up that is functionally identical. Once squashed we'll wait for the CI build and this can then be merged.

hzongaro:
> @hzongaro ok that looks good please squash

Squashed into a25163f730d1670535a123ba5f9709b593b9e1e0

Issue #3531, opened 2018-11-02
Use potentialOSRPointHelper call as a marker of OSR point


liqunl:
OSR points are places in a jitted code where transition to the interpreter is possible. Currently, OSR points are calls, asyncchecks and monitor enters. Calls can be inlined or optimized aways, which adds the difficulty in locating OSR points. Serving as a marker, `potentialOSRPointHelper` is to help locating OSR points. With more OSR points, we can do more OSR dependent optimizations.
 
To use potentialOSRPointHelper as OSR point marker, we have to
   1. Insert potentialOSRPointHelper call after each OSR point in ILGen
   2. Remove potentialOSRPointHelper calls that are no longer safe OSR points due to StringPeepHoles transformation.
   3. Remove redudant potentialOSRPointHelper calls before inserting OSR guards.
   4. Clean up potentialOSRPointHelper calls after OSR guards have been inserted.

Depends on https://github.com/eclipse/omr/pull/3030

PR #3744, opened 2018-11-20
Use potentialOSRPointHelper call as OSR point marker

liqunl:
This pull request contains three commits to use potentialOSRPointHelper
call as OSR point marker.

 - Insert a helper call right after each OSR point.
 - Remove the helper calls where OSR is not possible due to
    StringPeepHoles transformations.
 - Remove redundant helper calls in OSRGuardInsertion.

Details can be found in the commit message of each commit.

Closes: #3531

### DIFF
--- a/runtime/compiler/ilgen/Walker.cpp
+++ b/runtime/compiler/ilgen/Walker.cpp
@@ -5380,6 +5380,27 @@ break
          }
          _intrinsicErrorHandling = false;
       }
+
+   // The call may be transformed into a non-OSR point. Check if bookkeeping is needed
+   // before the transformation.
+   bool needOSRBookkeeping = false;
+   int32_t osrInductionOffset;
+
+   // callNodeTreeTop may be null if this call should not be placed in the trees
+   if (callNodeTreeTop
+       && comp()->getOption(TR_EnableOSR)
+       && !comp()->isPeekingMethod()
+       && comp()->isOSRTransitionTarget(TR::postExecutionOSR)
+       && comp()->isPotentialOSRPoint(callNodeTreeTop->getNode())
+       && !_methodSymbol->cannotAttemptOSRAt(callNode->getByteCodeInfo(), NULL, comp())
+       && !_cannotAttemptOSR)
+      {
+      needOSRBookkeeping = true;
+      // callNode may become a non-OSR point after the transformation, calling getOSRInductionOffset
+      // on a non-OSR point will trigger the assertion, thus get the induction offset here.
+      osrInductionOffset = comp()->getOSRInductionOffset(callNode);
+      }
+
    TR::Node * resultNode = 0;
 
    TR::ResolvedMethodSymbol * resolvedMethodSymbol = symbol->getResolvedMethodSymbol();
@@ -5535,17 +5556,15 @@ break
       push(resultNode);
       }
 
-   // callNodeTreeTop may be null if this call should not be placed in the trees
-   if (callNodeTreeTop
-       && comp()->getOption(TR_EnableOSR)
-       && !comp()->isPeekingMethod()
-       && comp()->isOSRTransitionTarget(TR::postExecutionOSR)
-       && comp()->isPotentialOSRPoint(callNodeTreeTop->getNode())
-       && !_methodSymbol->cannotAttemptOSRAt(callNode->getByteCodeInfo(), NULL, comp())
-       && !_cannotAttemptOSR)
+   if (needOSRBookkeeping)
       {
       saveStack(-1, !comp()->pendingPushLivenessDuringIlgen());
-      stashPendingPushLivenessForOSR(comp()->getOSRInductionOffset(callNode));
+      stashPendingPushLivenessForOSR(osrInductionOffset);
+      if (comp()->supportsInduceOSR() && comp()->getOSRMode() == TR::voluntaryOSR)
+         {
+         TR::Node* callToPotentialOSRPoint = TR::Node::createPotentialOSRPointHelperCallInILGen(callNodeTreeTop->getNode(), osrInductionOffset);
+         _block->append(TR::TreeTop::create(comp(), TR::Node::create(TR::treetop, 1, callToPotentialOSRPoint)));
+         }
       }
 
    if (cg()->getEnforceStoreOrder() && calledMethod->isConstructor())
--- a/runtime/compiler/optimizer/OSRGuardInsertion.cpp
+++ b/runtime/compiler/optimizer/OSRGuardInsertion.cpp
@@ -30,63 +30,221 @@
 #include "optimizer/Optimization_inlines.hpp"
 #include "optimizer/FearPointAnalysis.hpp"
 #include "optimizer/RematTools.hpp"
-#include "optimizer/HCRGuardAnalysis.hpp"
 #include "optimizer/TransformUtil.hpp"
 #include "ras/DebugCounter.hpp"
+#include "infra/Checklist.hpp"             // for TR::NodeChecklist
+
+TR_Structure* fakeRegion(TR::Compilation *comp);
 
 static bool generatesFear(TR::Compilation *comp, TR_FearPointAnalysis &fearAnalysis, TR::Block *block)
    {
    return fearAnalysis._blockAnalysisInfo[block->getNumber()]->get(0);
    }
 
-int32_t TR_OSRGuardInsertion::perform()
+
+static bool hasHCRGuard(TR::Compilation *comp)
    {
+   TR::list<TR_VirtualGuard*> &virtualGuards = comp->getVirtualGuards();
+   for (auto itr = virtualGuards.begin(), end = virtualGuards.end(); itr != end; ++itr)
+      {
+      if ((*itr)->getKind() == TR_HCRGuard || (*itr)->mergedWithHCRGuard())
+         return true;
+      }
+
+   return false;
+   }
+
+static bool hasUnsupportedPotentialOSRPoint(TR::Compilation *comp, bool trace)
+   {
+   for (TR::TreeTop *treeTop = comp->getStartTree(); treeTop; treeTop = treeTop->getNextTreeTop())
+      {
+      TR::Node *ttNode = treeTop->getNode();
+
+      if (comp->isPotentialOSRPoint(ttNode) &&
+          !comp->isPotentialOSRPointWithSupport(treeTop))
+         {
+         if (trace)
+            traceMsg(comp, "Found an unsupported potential OSR point at n%dn\n", ttNode->getGlobalIndex());
+         return true;
+         }
+      }
+
+   return false;
+   }
+
+void TR_OSRGuardInsertion::cleanUpPotentialOSRPointHelperCalls()
+   {
+   for (TR::TreeTop *treeTop = comp()->getStartTree(); treeTop; treeTop = treeTop->getNextTreeTop())
+      {
+      TR::Node *ttNode = treeTop->getNode();
+      if (ttNode->getNumChildren() != 1)
+         continue;
+
+      TR::Node *node = ttNode->getFirstChild();
+      if (node->isPotentialOSRPointHelperCall())
+         {
+         TR::TreeTop* prevTreeTop = treeTop->getPrevTreeTop();
+         TR::TransformUtil::removeTree(comp(), treeTop);
+         treeTop = prevTreeTop;
+         }
+      }
+   }
+
+
+/** \brief
+ *     Remove redundant potentialOSRPointHelper calls using HCRGuardAnalysis.
+ *
+ *     A potentialOSRPointHelper call not reached by a gen is considered redundant.
+ *     The effect of a potentialOSRPointHelper call in data flow is to kill. If no
+ *     gen reaches it, it doesn't matter if we kill there. Thus removing the redundant
+ *     potentialOSRPoint helper calls will not change the result of HCRGuardAnalysis if
+ *     we were to repeat it after the removal.
+ *
+ *  \param guardAnalysis
+ *     Result of HCRGuardAnalysis used to determine which helper call is redundant.
+ *
+ */
+void TR_OSRGuardInsertion::removeRedundantPotentialOSRPointHelperCalls(TR_HCRGuardAnalysis* guardAnalysis)
+   {
+   bool protectedByOSRPoints = false;
+   TR::NodeChecklist visited(comp());
+
+   for (TR::TreeTop *treeTop = comp()->getStartTree();
+        treeTop;
+        treeTop = treeTop->getNextRealTreeTop())
+      {
+      TR::Node *ttNode = treeTop->getNode();
+
+      if (ttNode->getOpCodeValue() == TR::BBStart)
+         {
+         TR::Block* block = treeTop->getEnclosingBlock();
+         protectedByOSRPoints = !guardAnalysis || guardAnalysis->_blockAnalysisInfo[block->getNumber()]->isEmpty();
+         continue;
+         }
+
+      TR::Node *osrNode = NULL;
+      if (comp()->isPotentialOSRPoint(ttNode, &osrNode))
+         {
+         if (visited.contains(osrNode))
+            continue;
+
+         if (protectedByOSRPoints &&
+             osrNode->isPotentialOSRPointHelperCall())
+            {
+            dumpOptDetails(comp(), "Remove redundant potentialOSRPointHelper call n%dn %p\n", osrNode->getGlobalIndex(), osrNode);
+
+            TR::TreeTop* prevTree = treeTop->getPrevTreeTop();
+            TR::TransformUtil::removeTree(comp(), treeTop);
+            treeTop = prevTree;
+            }
+         else if (comp()->isPotentialOSRPointWithSupport(treeTop))
+            {
+            if (!protectedByOSRPoints && trace())
+               traceMsg(comp(), "treetop n%dn is an OSR point with support\n", ttNode->getGlobalIndex());
+
+            protectedByOSRPoints = true;
+            }
+         else
+            {
+            if (protectedByOSRPoints && trace())
+               traceMsg(comp(), "treetop n%dn is an OSR point without support\n", ttNode->getGlobalIndex());
+
+            protectedByOSRPoints = false;
+            }
+
+         visited.add(osrNode);
+         continue;
+         }
+      }
+   }
+
+static bool skipOSRGuardInsertion(TR::Compilation* comp)
+   {
+   // Use with caution. There can be fear points generated before OSRGuardInsertion,
+   // disabling it will leave those fear points unprotected.
+   //
    static char *disableOSRGuards = feGetEnv("TR_DisableOSRGuards");
 
    // Currently, OSR guard insertion is only needed when in the OSR HCR mode
-   if (disableOSRGuards || !comp()->isOSRTransitionTarget(TR::postExecutionOSR))
+   if (disableOSRGuards || !comp->isOSRTransitionTarget(TR::postExecutionOSR))
       {
-      if (trace())
-         traceMsg(comp(), "HCR Guards are only removed when in OSR HCR mode - skipping\n");
-      return 0;
+      return true;
       }
 
    // Even under NextGenHCR, OSR may have been disabled for this compilation at runtime
-   if (!comp()->supportsInduceOSR())
+   if (!comp->supportsInduceOSR())
       {
-      if (trace())
-         traceMsg(comp(), "OSR induce is not supported during this compilation - skipping\n");
-      return 0;
+      return true;
       }
 
-   // Detect if there are no HCR guards
-   TR::list<TR_VirtualGuard*> &virtualGuards = comp()->getVirtualGuards();
-   auto guard = virtualGuards.begin();
-   for (; guard != virtualGuards.end(); ++guard)
+   return false;
+   }
+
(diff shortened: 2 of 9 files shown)


liqunl:
Performance evaluation
1. Daytrader3

Relative throughput (compared to baseline), the higher the score, the better the performance

| Benchmark | Relative score| Standard deviation|
| daytrader3  | -0.1%  | 0.65%

With the change, we'll do HCRGuardAnalysis when there exists a bad yield point, which is the right thing to do but cost more compile time. The compile time overhead is 2.3% on daytrader3. The noise margin for compile time measurement is 0.43

Thus I have measured the performance on HiBench since it's compile time sensitive.

2. HiBench

| Benchmark | Relative score| Standard deviation|
| Pagerank  | -1.7%  |3.34%
| Terasort | -0.7% |2.28%
|Wordcount|+0.3%|1.56%|
|Sort|+0.3%|1.64%|

In conclusion, the change shows no regression on daytrader3 and HiBench

liqunl:

runtime/compiler/ilgen/Walker.cpp
```diff
@@ -5478,17 +5499,15 @@ break
...
-       && comp()->isOSRTransitionTarget(TR::postExecutionOSR)
-       && comp()->isPotentialOSRPoint(callNodeTreeTop->getNode())
-       && !_methodSymbol->cannotAttemptOSRAt(callNode->getByteCodeInfo(), NULL, comp())
-       && !_cannotAttemptOSR)
+   if (needOSRBookkeeping)
       {
       saveStack(-1, !comp()->pendingPushLivenessDuringIlgen());
-      stashPendingPushLivenessForOSR(comp()->getOSRInductionOffset(callNode));
+      stashPendingPushLivenessForOSR(osrInductionOffset);
+      if (comp()->supportsInduceOSR())
```
andrewcraik: I think you might want getOSRMode() == TR::voluntaryOSRMode to be sure we don't try and do this for involuntary OSR.

runtime/compiler/ilgen/Walker.cpp
```diff
@@ -5323,6 +5323,27 @@ break
...
+   // The call may be transformed into a non-OSR point. Check if bookkeeping is needed
+   // before the transformation.
+   bool needOSRBookkeeping = false;
+   int32_t osrInductionOffset;
+
+   // callNodeTreeTop may be null if this call should not be placed in the trees
+   if (callNodeTreeTop
+       && comp()->getOption(TR_EnableOSR)
+       && !comp()->isPeekingMethod()
+       && comp()->isOSRTransitionTarget(TR::postExecutionOSR)
```
andrewcraik: what about if the mode is preExecutionOSR do we need another helper marker for that?

runtime/compiler/optimizer/StringPeepholes.cpp
```diff
@@ -79,6 +79,53 @@
...
+      TR::TreeTop *tt = start;
+      do
+         {
+         TR::Node *osrNode = NULL;
+         if (comp()->isPotentialOSRPoint(tt->getNode(), &osrNode))
+            {
+            if (osrNode->isPotentialOSRPointHelperCall())
+               {
+               if (trace())
+                  traceMsg(comp(), "Remove tt n%dn with potential osr point %p n%dn\n", tt->getNode()->getGlobalIndex(), osrNode, osrNode->getGlobalIndex());
```
andrewcraik: this should be a dumpOptDetails

runtime/compiler/optimizer/StringPeepholes.cpp
```diff
@@ -79,6 +79,53 @@
 #include "env/PersistentCHTable.hpp"           // for CLASSHASHTABLE_SIZE, etc
 #include "env/PersistentInfo.hpp"              // for PersistentInfo
 
+void TR_StringPeepholes::removePotentialOSRPointHelperCalls(TR::TreeTop* start, TR::TreeTop* end)
+   {
+   if (comp()->isOSRTransitionTarget(TR::postExecutionOSR) && comp()->getOption(TR_EnableOSR))
```
andrewcraik: we should only do this for voluntary OSR modes

runtime/compiler/optimizer/StringPeepholes.cpp
```diff
@@ -79,6 +79,53 @@
...
+      // or on the surrounding trees.
+      //
+      TR::TreeTop *tt = start;
+      do
+         {
+         TR::Node *osrNode = NULL;
+         if (comp()->isPotentialOSRPoint(tt->getNode(), &osrNode))
+            {
+            if (trace())
+               traceMsg(comp(), "Can no longer OSR at [%p] n%dn\n", osrNode, osrNode->getGlobalIndex());
```
andrewcraik: again dumpOptDetails rather than trace for this I think.

runtime/compiler/optimizer/StringPeepholes.cpp
```diff
@@ -309,7 +356,7 @@ void TR_StringPeepholes::processBlock(TR::Block *block)
    {
    TR::TreeTop *exit = block->getExit();
 
-   //traceMsg(comp(), "looking at block_%d\n", block->getNumber());
+   traceMsg(comp(), "looking at block_%d\n", block->getNumber());
```
andrewcraik: please guard or clean up

runtime/compiler/optimizer/StringPeepholes.cpp
```diff
@@ -2092,9 +2143,25 @@ bool TR_StringPeepholes::checkMethodSignature(TR::SymbolReference *symRef, const
  */
 bool TR_StringPeepholes::skipNodeUnderOSR(TR::Node *node)
    {
-   return comp()->getOption(TR_EnableOSR)
-      && comp()->isOSRTransitionTarget(TR::postExecutionOSR)
-      && comp()->getMethodSymbol()->isOSRRelatedNode(node);
+   bool result = false;
+   if (comp()->getOption(TR_EnableOSR)
+      && comp()->isOSRTransitionTarget(TR::postExecutionOSR))
```
andrewcraik: I don't know why the transition target affects this any OSR helper for a pre or post transitionw ould need to be skipable.

runtime/compiler/optimizer/StringPeepholes.cpp
```diff
@@ -79,6 +79,53 @@
...
+               TR::TransformUtil::removeTree(comp(), tt);
+               tt = prevTT;
+               }
+            }
+         tt = tt->getNextTreeTop();
+         } while (tt != end);
+      }
+   }
+
+void TR_StringPeepholes::fixUpOSRBookkeeping(TR::TreeTop* start, TR::TreeTop* end)
```
andrewcraik: This is a bit misleading I think this needs to be something like prohibitOSROverRange or something like that since the bookkeeping is the stuff in the OMR layer tracking the live slots etc.

runtime/compiler/optimizer/StringPeepholes.cpp
```diff
@@ -79,6 +79,53 @@
...
+   {
+   if (comp()->isOSRTransitionTarget(TR::postExecutionOSR) && comp()->getOption(TR_EnableOSR))
+      {
+      // Walk region and ensure all possible OSR transition points has doNotProfile set on their bci
```
andrewcraik: Can you please make this into doxygen for the method?

runtime/compiler/optimizer/StringPeepholes.cpp
```diff
@@ -332,7 +379,11 @@ void TR_StringPeepholes::processBlock(TR::Block *block)
 
             TR::TreeTop *newTree = detectPattern(block, tt, true);
             if (newTree)
+               {
```
andrewcraik: I am concerned by the sequence which follows here. You are removing the OSR infrastructure before we have 'matched' the pattern and passed the point where we do the perform transformation to say that we are going to replace the peephole sequence with a new String of a private constructor targeting the specific sequence of appends we have found. This would be much more correct and much nicer if we skipped the OSR nodes as part of the pattern matching, ran the performTransformation and then did the remove of the OSR points before we did the manipulation to switch to the custom constructor. This pattern could see OSR infrastructure removed needlessly and could mean that doing a lastOptSubIndex in StringPeepholes could have unintended consequences on the OSR points and transitions seen by the OSR infrastructure which runs later.

runtime/compiler/optimizer/OSRGuardInsertion.cpp
```diff
@@ -30,63 +30,205 @@
...
+      if ((*guard)->getKind() == TR_HCRGuard || (*guard)->mergedWithHCRGuard())
+         return true;
+      }
 
-   // Currently, OSR guard insertion is only needed when in the OSR HCR mode
-   if (disableOSRGuards || !comp()->isOSRTransitionTarget(TR::postExecutionOSR))
+   return true;
+   }
+
+static bool hasYieldPointCannotOSR(TR::Compilation *comp)
```
andrewcraik: I would prefer hasUnsupportedPotentialOSRPoint since this is about OSR points and the relationship to yield points is not determined - we are only concerned about transition points (which are normally yield points, but we don't need to conflate the two concepts).

runtime/compiler/optimizer/OSRGuardInsertion.cpp
```diff
@@ -30,63 +30,205 @@
...
+   bool needHCRGuardRemoval = hasHCRGuard(comp());
+   bool canInsertOSRGuards = !skipOSRGuardInsertion(comp());
+
+   if (canInsertOSRGuards && needHCRGuardRemoval)
       {
-      if (trace())
-         traceMsg(comp(), "No fear generating nodes - skipping\n");
-      comp()->getFlowGraph()->invalidateStructure();
-      return 0;
+      bool hasBadYieldPoint = hasYieldPointCannotOSR(comp());
```
andrewcraik: hasBadYieldPoint should be something like hasUnsupportedOSRPoint or something like that...

runtime/compiler/optimizer/OSRGuardInsertion.cpp
```diff
@@ -30,63 +30,205 @@
...
+      // Future fear generating optimizations
+      //
+      if (!fearGeneratingNodes.isEmpty())
+         {
+         // Redudant potentialOSRPointHelper calls may result in more OSR guards than needed
+         //
+         if (hasBadYieldPoint)
+            removeRedundantPotentialOSRPointHelperCalls(guardAnalysis);
+         else
+            cleanUpPotentialOSRPointHelperCalls();
```
andrewcraik: Why clean up here when there is a cleanup later on - this doesn't quite make sense to me... An explanatory comment is warranted.

runtime/compiler/optimizer/OSRGuardInsertion.cpp
```diff
@@ -30,63 +30,205 @@
...
+         treeTop = prevTreeTop;
+         }
       }
+   }
 
-   // Detect if there are no HCR guards
-   TR::list<TR_VirtualGuard*> &virtualGuards = comp()->getVirtualGuards();
-   auto guard = virtualGuards.begin();
-   for (; guard != virtualGuards.end(); ++guard)
+void TR_OSRGuardInsertion::removeRedundantPotentialOSRPointHelperCalls(TR_HCRGuardAnalysis* guardAnalysis)
```
andrewcraik: Please provide a comment with an intuition for how this is working - I'm still working through the code to understand its interplay with the analyses in this opt pass.

runtime/compiler/optimizer/OSRGuardInsertion.cpp
```diff
@@ -30,63 +30,205 @@
...
+      TR::Node *osrNode = NULL;
+      if (comp()->isPotentialOSRPoint(ttNode, &osrNode))
+         {
+         if (visited.contains(osrNode))
+            continue;
+
+         if (protectedByOSRPoints &&
+             osrNode->isPotentialOSRPointHelperCall())
+            {
+            if (trace())
```
andrewcraik: Is this optional? If so it should be a perform transformation. If not this is probably a dumpOptDetails kind of thing.

andrewcraik:
I think some discussion of this change in relation to the pre and post OSR models is warranted - I think we need to be clear on when the helpers are needed and why since some fo the same issues affecting postExcecutionOSR could affect preExecutionOSR.

runtime/compiler/optimizer/OSRGuardInsertion.cpp
jdmpapin: While your suggested name would be more consistent with the naming of the queries in terms of which this is implemented, I think it's those queries that are confusing. "OSR point" sounds like a point at which OSR is possible, a transition point as you put it. An example of the confusion this causes: this function is not concerned with transition points at all. It is concerned instead with points that allow assumptions to be invalidated but which are specifically _not_ transition points. In the context of OSR specifically, I think the "yield point" terminology does make it clear which property is relevant, although it does conflict somewhat with the more general notion of a yield point, and with the existing queries `isPotentialOSRPoint*`.

Maybe we should try to move things consistently to a set of clearer terminology? e.g. abandon "OSR point" and speak only of "OSR transition point" and "OSR yield point"/"OSR invalidation point." Not as part of this PR, of course.

runtime/compiler/optimizer/StringPeepholes.cpp
jdmpapin: `detectPattern()` both matches the pattern and, if it finds a match, transforms the code - see the end of the function [starting here](https://github.com/eclipse/openj9/blob/7ebb4d4bef739fde5b4fcd6085f15449ffb25048/runtime/compiler/optimizer/StringPeepholes.cpp#L1863-L1890). I expect the other `detect*Pattern()` methods work similarly, considering that there doesn't seem to be any further transformation after skipping ahead to `newTree` in any of `processBlock`, `process`, or `perform` (except an attempt to inline some calls in the last of these).

liqunl: Yes. all the `detect*Pattern()` methods detect and transform the code at the same time, then return the last tree in the pattern.

runtime/compiler/ilgen/Walker.cpp
```diff
@@ -5380,6 +5380,27 @@ break
...
+       && comp()->isPotentialOSRPoint(callNodeTreeTop->getNode())
+       && !_methodSymbol->cannotAttemptOSRAt(callNode->getByteCodeInfo(), NULL, comp())
+       && !_cannotAttemptOSR)
+      {
+      needOSRBookkeeping = true;
+      // callNode may become a non-OSR point after the transformation, calling getOSRInductionOffset
+      // on a non-OSR point will trigger the assertion, thus get the induction offset here.
+      osrInductionOffset = comp()->getOSRInductionOffset(callNode);
```
andrewcraik: Can you please share a case where the call node becomes something other than a call and we need the OSR point - ie we will still have a point where we want to decompile and that point is at an invalidation point? If it isn't an invalidation point why do we want the transition point?

runtime/compiler/optimizer/StringPeepholes.cpp
```diff
@@ -79,6 +79,80 @@
...
+ *     as they will not be able to reconstruct their stacks due to the optimizations done on them
+ *     or on the surrounding trees.
+ *
+ *  \param start
+ *     The tree marking the start of the range.
+ *
+ *  \param end
+ *     The tree marking the end of the range.
+ */
+void TR_StringPeepholes::prohibitOSROverRange(TR::TreeTop* start, TR::TreeTop* end)
```
andrewcraik: This also feels like a more general utility. I think it needs a check for voluntary OSR because it is not valid in involuntary OSR.

runtime/compiler/optimizer/StringPeepholes.cpp
andrewcraik: I think this needs a fast return for involuntary OSR where no skip is needed.

runtime/compiler/ilgen/Walker.cpp
liqunl: `J9VMInternals.getSuperclass` will be transformed to a different call. A call to `JITHelpers.jitHelpers()` is also inserted. In this case, the invalidation point still exists but the calls are not OSR points. Even if the invalidation point is removed, there is no harm in doing the OSR bookkeeping and inserting the OSR helper call. Any redundant helper calls will be removed before inserting OSR guards.

liqunl:
@andrewcraik could you review again?

runtime/compiler/ilgen/Walker.cpp
andrewcraik: right it just adds overhead, but given the perf numbers I guess we can live with that

andrewcraik:
Ok I think this is looking good - I'm going to launch the builds and would just like a second opinion on correctness from @jdmpapin

andrewcraik:

andrewcraik:

runtime/compiler/optimizer/OSRGuardInsertion.cpp
```diff
@@ -30,63 +30,222 @@
...
+      //
+      TR_Structure *structure = requiresAnalysis ? fakeRegion(comp()) : NULL;
+      comp()->getFlowGraph()->setStructure(structure);
+
+      TR_HCRGuardAnalysis *guardAnalysis = requiresAnalysis ? new (comp()->allocator()) TR_HCRGuardAnalysis(comp(), optimizer(), structure) : NULL;
+      TR_BitVector fearGeneratingNodes(comp()->getNodeCount(), trMemory(), stackAlloc);
+
+      // Handle fear nodes for HCR guards, branch to slow path might be removed
+      //
+      if (needHCRGuardRemoval)
```
jdmpapin: `needHCRGuardRemoval` is always true here. I think the logic is right as-is, but the `if` is misleading

liqunl:
Failure is not related to this PR. There is a closed issue with the same test failure. https://github.com/eclipse/openj9/issues/594

runtime/compiler/optimizer/OSRGuardInsertion.cpp
```diff
@@ -30,63 +30,222 @@
...
+   TR::list<TR_VirtualGuard*> &virtualGuards = comp->getVirtualGuards();
+   for (auto itr = virtualGuards.begin(), end = virtualGuards.end(); itr != end; ++itr)
+      {
+      if ((*itr)->getKind() == TR_HCRGuard || (*itr)->mergedWithHCRGuard())
+         return true;
+      }
 
-   // Currently, OSR guard insertion is only needed when in the OSR HCR mode
-   if (disableOSRGuards || !comp()->isOSRTransitionTarget(TR::postExecutionOSR))
+   return true;
```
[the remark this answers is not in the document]
liqunl: The query is `skipOSRGuardInsertion`

runtime/compiler/optimizer/StringPeepholes.cpp
```diff
@@ -1945,9 +1964,9 @@ TR::TreeTop *TR_StringPeepholes::detectPattern(TR::Block *block, TR::TreeTop *tt
             // create a new treetop for String.valueOf()
             //                                                      nodeToConvert
             TR::SymbolReference * symRef = getSymRefTab()->findOrCreateMethodSymbol((appendedString[j] && appendedString[j]->getOpCode().hasSymbolReference()) ? appendedString[j]->getSymbolReference()->getOwningMethodIndex() : JITTED_METHOD_INDEX, -1, valueOfSymRef[j]->getSymbol()->getResolvedMethodSymbol()->getResolvedMethod(), TR::MethodSymbol::Static);
-            TR::Node *valueOf = TR::Node::createWithSymRef(TR::acall, 1, 1, appendedString[j], symRef);
+            TR::Node *valueOf = TR::Node::createWithSymRef(stringNode, TR::acall, 1, appendedString[j], symRef);
             TR::TreeTop::create(comp(), appendTree[j],
-                               TR::Node::create(TR::treetop, 1, valueOf));
+                               TR::Node::create(stringNode, TR::treetop, 1, valueOf));
```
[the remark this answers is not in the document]
liqunl: The `append` calls are to be removed, the bci on `stringNode` will be used on the final tree, so I choose to use bci `stringNode`.

jdmpapin:
> I believe we have some code somewhere that detects cases where the result of a call is immediately stored into a local, and moves the transition point for that call to after the store, so that the pending push will no longer be live at the transition. I couldn't easily find it, however.

@liqunl pointed me to [the code I was thinking of](https://github.com/eclipse/openj9/blob/fca3de3964302be44834593f7fe9c7705b0a1888/runtime/compiler/optimizer/OSRGuardInsertion.cpp#L378-L390), and it turns out I was mistaken. The target bytecode instruction for the transition is left unchanged, and the stores are repeated in the interpreter. In particular, the pending push temp holding the result of the call remains live. We simply move the OSR guard past repeatable stores so that splitting the block won't cause the call result to be stored and reloaded through yet *another* temp. This should just work for calls to `potentialOSRPointHelper`.

liqunl:
WIP this PR as I need to add change to skip the helper call in StringBuilderTransformer

liqunl:
@andrewcraik @jdmpapin Sorry for the last minute change. I added a change in StringBuilderTransformer to skip the helper call during pattern matching. Could you review again?

liqunl:
@andrewcraik @jdmpapin ping?

jdmpapin:
Still lgtm

andrewcraik:
I think that looks reasonable - launching sanity again

andrewcraik:

liqunl:
WIP this PR as it seems to have exposed a bug.

PR #3754, opened 2018-11-20
WIP: JEP 331 : Low-Overhead Heap Profiling Stage 1

JasonFengJ9:
`JEP 331` : Low-Overhead Heap Profiling Stage 1

**Overview**:

This is the initial change for `JEP 331` which aims to provide the functionality and doesn't affect performance when the feature is not enabled. Further update is to add more tests and ensure the performance targets are met as per https://github.com/eclipse/openj9/issues/3436.

The sampling feature is going to be enabled when `can_generate_sampled_object_alloc_events` capability is requested. 
Depends on the sampling interval requested (512KB by default), two different approaches are used to determine how to report a sampling event. 
1. When the interval requested is less than minimum `TLH` size (~0.5KB), `disableInlineTLHAllocate()` is applied hence allocation has to always go through slow path. 
2. When the interval is not less than minimum `TLH` size, `GCExtensionsBase.tlhActiveMaximumSize` is set to the interval size or the actual maximum `TLH` size (~128KB) whichever less, further `TLH` allocation is based on this size, when a `TLH` refresh is performed, current `TLH` allocation size is recorded, `disableInlineTLHAllocate()` is invoked if the sampling interval has been reached.
After `disableInlineTLHAllocate()` call, next allocation is out-of-line at `mgcalloc.cpp J9AllocateObject() & J9AllocateIndexableObject()` where `TRIGGER_J9HOOK_SAMPLED_OBJECT_ALLOCATE` is invoked. 

**Outline of changes:**

A new callback method:
```
	void JNICALL SampledObjectAlloc(jvmtiEnv *jvmti_env, JNIEnv* jni_env, jthread thread, jobject object, jclass object_klass, jlong size)
```
A new jvmti method:
```
	jvmtiError SetHeapSamplingInterval(jvmtiEnv* env, jint sampling_interval)
```
A new event type: `JVMTI_EVENT_SAMPLED_OBJECT_ALLOC`
A new capability: `can_generate_sampled_object_alloc_events`
A new hook event: `J9HOOK_SAMPLED_OBJECT_ALLOCATE`
`j9gc_set_allocation_sampling_interval`
`j9gc_set_allocation_sampling`

Adding a testcase to verify this feature in a simple allocation case. 

Note: `jvmti.h` change affects `Java 8` and version before `Java 11` which will be addressed by https://github.com/eclipse/openj9/issues/2368.

Depends on https://github.com/eclipse/omr/pull/3220

Peer reviewer: @tajila 
Reviewer: @DanHeidinga 


runtime/gc_base/modronapi.cpp
```diff
@@ -786,6 +786,44 @@ j9gc_allocation_threshold_changed(J9VMThread *currentThread)
...
+ * Set the flag to enable or disable allocation sampling
+ * 
+ * @parm[in] vmThread The current VM Thread
+ * @parm[in] flag true - enable allocation sampling, false otherwise
+ */
+void 
+j9gc_set_allocation_sampling(J9VMThread *vmThread, BOOLEAN flag)
+{
+	J9JavaVM *vm = vmThread->javaVM;
+	MM_GCExtensions *ext = MM_GCExtensions::getExtensions(vm);
```
charliegracie: What happens if this API is called and `j9gc_set_allocation_sampling_interval` was not called?

JasonFengJ9: The sampling interval stays as default value 512KB which is larger than maximum TLH size 128KB, and `tlhActiveMaximumSize` stays as default value which is same as `tlhMaximumSize`.

runtime/jvmti/jvmtiMemory.c
```diff
@@ -80,4 +80,21 @@ jvmtiDeallocate(jvmtiEnv* env,
...
-
+jvmtiError JNICALL 
+jvmtiSetHeapSamplingInterval(jvmtiEnv *env, 
+	jint samplingInterval)
+{
+	jvmtiError rc = JVMTI_ERROR_NONE;
+	J9VMThread *currentThread = NULL;
+	
+	Trc_JVMTI_jvmtiSetHeapSamplingInterval_Entry(env, samplingInterval);
+
```
tajila: need to ensure that this is called during the OnLoad or the live phase

tajila: need to check that sampling interval is non-negative

tajila: what happens if this is called and the JVM doesn't posses the capability to do heap sampling?

runtime/jvmti/jvmtiMemory.c
```diff
@@ -80,4 +80,21 @@ jvmtiDeallocate(jvmtiEnv* env,
...
+	jvmtiError rc = JVMTI_ERROR_NONE;
+	J9VMThread *currentThread = NULL;
+	
+	Trc_JVMTI_jvmtiSetHeapSamplingInterval_Entry(env, samplingInterval);
+	
+	rc = getCurrentVMThread(((J9JVMTIEnv *)env)->vm, &currentThread);
+	if ((JVMTI_ERROR_NONE == rc) && (NULL != currentThread)) {
+		J9MemoryManagerFunctions *fns = currentThread->javaVM->memoryManagerFunctions;
+		if (NULL != fns) {
+			fns->j9gc_set_allocation_sampling_interval(currentThread, samplingInterval);
```
tajila: what happens if we fail (eg. pass in a sampling interval of 1TB).
  
should this return an internal error? or fail silently

runtime/jvmti/jvmtiHook.c
```diff
@@ -2776,6 +2785,40 @@ jvmtiHookCompilingEnd(J9HookInterface** hook, UDATA eventNum, void* eventData, v
...
+	if (NULL != callback) {
+		J9SampledObjectAllocateEvent *data = eventData;
+		J9VMThread * currentThread = data->currentThread;
+		jthread threadRef = NULL;
+		UDATA hadVMAccess = 0;
+		UDATA javaOffloadOldState = 0;
+
+		if (prepareForEvent(j9env, currentThread, currentThread, JVMTI_EVENT_SAMPLED_OBJECT_ALLOC, &threadRef, &hadVMAccess, TRUE, 2, &javaOffloadOldState)) {
+			j9object_t * objectRef = (j9object_t*) currentThread->arg0EA;
+			j9object_t * classRef = (j9object_t*) (objectRef - 1);
```
tajila: Nitpick, might be more clear if you were fetching classRef from a relative offset off of arg0EA

runtime/jvmti/jvmtiHook.c
```diff
@@ -2776,6 +2785,40 @@ jvmtiHookCompilingEnd(J9HookInterface** hook, UDATA eventNum, void* eventData, v
...
+			J9Class* clazz = NULL;
+			J9JavaVM * vm = currentThread->javaVM;
+
+			*objectRef = data->object;
+			clazz = J9OBJECT_CLAZZ(currentThread, data->object);
+			*classRef = J9VM_J9CLASS_TO_HEAPCLASS(clazz);
+			vm->internalVMFunctions->internalExitVMToJNI(currentThread);
+			callback((jvmtiEnv *) j9env, (JNIEnv *) currentThread, threadRef, (jobject) objectRef, (jclass) classRef, (jlong) data->size);
+			currentThread->javaVM->internalVMFunctions->internalEnterVMFromJNI(currentThread);
+			data->object = J9_JNI_UNWRAP_REDIRECTED_REFERENCE(objectRef);
```
tajila: why is this being set? is data still used after the callback function has run

runtime/jvmti/jvmtiMemory.c
JasonFengJ9: Added `ENSURE_PHASE_ONLOAD_OR_LIVE`, `ENSURE_CAPABILITY` and `ENSURE_NON_NEGATIVE` accordingly.

runtime/jvmti/jvmtiMemory.c
JasonFengJ9: Any value larger than `Integer.MAX_VALUE` will be casted to `jint`, and the negative value is going to be caught by `ENSURE_NON_NEGATIVE(samplingInterval)`.

runtime/jvmti/jvmtiHook.c
JasonFengJ9: Fixed.

runtime/jvmti/jvmtiHook.c
JasonFengJ9: No, that's not the case. Removed.

JasonFengJ9:
@tajila sorry for late response (was distracted by other workitems), please have another look.

JasonFengJ9:
superseded by https://github.com/eclipse/openj9/pull/4461, https://github.com/eclipse/openj9/pull/4618, https://github.com/eclipse/openj9/pull/5869, https://github.com/eclipse/openj9/pull/5968.

PR #3755, opened 2018-11-20
Use correct type conversion in UnsafeFastPath

liqunl:
Unsigned conversion is only needed for JITHelpers methods that load a
char from an array.


liqunl:
Will remove the WIP once my personal build finishes. @fjeremic Can you take a look?

fjeremic:
Jenkins test sanity plinux,xlinux,zlinux JDK8,JDK11

andrewcraik:
Can we verify that the use of this helper in String is correctly using the convertByteToCharUnsigned helper before we merge this to avoid exposing a second round of bugs?

liqunl:
@andrewcraik Checked String.java, all the places treating the result of `getByteFromArrayByIndex` as a char have called `byteToCharUnsigned`.

andrewcraik:
Thanks for the verification - merging

PR #3758, opened 2018-11-20
Switch to the proper OMR_COMPILER flag

charliegracie:
The OMR_JIT flag should be removed once the OMR_COMPILER flag is
used everywhere.


charliegracie:
FYI @dnakamura

charliegracie:
@gacholio since this passed the Travis build, which is the only CMake build, I am very confident this change is correct :)

PR #1253, opened 2018-02-22
Add initial implementation of withfield bytecode

eric98zhang:
Add initial implementation for the Valhalla L-World withfield bytecode.

This implementation accepts a reference to a heap-allocated value class
instance from the stack and clones it on the heap. The desired field of
the clone is mutated and a reference to the new object is left on the
stack. The implementation details are similar to those of
inlInternalsPrimitiveClone() and putfield() in BytecodeInterpreter.hpp.

IncompatibleClassChangeError is thrown if the class of the instance is
not a value class.

Signed-off-by: Eric Zhang <eric99.zhang@gmail.com>

PR #3760, opened 2018-11-20
Add withfield bytecode support

tajila:
Add support for the withfield bytecode

Based on Eric's changes in 
#1253


tajila:
@gacholio Please review this PR

runtime/bcutil/ConstantPoolMap.hpp
```diff
@@ -275,6 +275,7 @@ class ConstantPoolMap
 	void markClassAsUsedByNew(U_16 classCfrCPIndex)            { mark(classCfrCPIndex, NEW); }
 #if defined(J9VM_OPT_VALHALLA_VALUE_TYPES)
 	void markClassAsUsedByDefaultValue(U_16 classCfrCPIndex)   { mark(classCfrCPIndex, DEFAULT_VALUE); }
+	void markFieldRefAsUsedByWithField(U_16 fieldRefCfrCPIndex){ mark(fieldRefCfrCPIndex, WITH_FIELD); }
```
gacholio: No space before {

runtime/oti/j9consts.h
```diff
@@ -250,6 +250,9 @@ extern "C" {
 #define J9_RESOLVE_FLAG_JCL_CONSTANT_POOL 0x100
 #define J9_RESOLVE_FLAG_REDEFINE_CLASS 0x200
 #define J9_RESOLVE_FLAG_NO_CLASS_INIT 0x400
+#if defined(J9VM_OPT_VALHALLA_VALUE_TYPES)
+#define J9_RESOLVE_FLAG_FIELD_WITHER 0x800
```
gacholio: That's a very odd name.

runtime/vm/BytecodeInterpreter.hpp
```diff
@@ -8114,8 +8114,91 @@ done:;
 	VMINLINE VM_BytecodeAction
 	withfield(REGISTER_ARGS_LIST)
 	{
+retry:
 		VM_BytecodeAction rc = EXECUTE_BYTECODE;
-		// TODO: Implement bytecode
+		const U_16 index = *(U_16 *)(_pc + 1);
```
gacholio: Please put const after the type (here and elsewhere), rather than before.

runtime/vm/BytecodeInterpreter.hpp
```diff
@@ -8114,8 +8114,91 @@ done:;
...
+		/* Must be done after resolution as we don't know where the object ref is
+		 * until we have resolved the fieldref
+		 */
+		if (NULL == originalObjectRef) {
+			rc = THROW_NPE;
+			goto done;
+		}
+		objectRefClass = J9OBJECT_CLAZZ(_currentThread, originalObjectRef);
+
+		if (!J9_IS_J9CLASS_VALUETYPE(objectRefClass)) {
```
gacholio: Can this be ensure by verification instead?

runtime/vm/resolvesupport.cpp
```diff
@@ -939,7 +939,15 @@ resolveInstanceFieldRefInto(J9VMThread *vmStruct, J9Method *method, J9ConstantPo
 					}
 					goto done;
 				}
-				if (!finalFieldSetAllowed(vmStruct, false, method, definingClass, classFromCP, field, canRunJavaCode)) {
+			if (
+#if defined(J9VM_OPT_VALHALLA_VALUE_TYPES)
+				/* The withfield bytecode is allowed to set a final field. However, the invoker of withfield must have private access
+				 * to the field (similar to a constructor setting a final field) */
+				J9_ARE_NO_BITS_SET(resolveFlags, J9_RESOLVE_FLAG_FIELD_WITHER) &&
```
gacholio: This code doesn't seem to match the comment - all withfield of final fields will be allowed.

runtime/oti/j9consts.h
tajila: It a naming convention used in the earlier days of VT. (ie. getter, setter & wither).  
  
Ill change it to withfield for more clarity.

runtime/vm/BytecodeInterpreter.hpp
tajila: The verifier will only see the signatures. An 'L' signature may be a VT or just a regular Ref type so we cant do the check there.

runtime/vm/resolvesupport.cpp
tajila: The private access check is done above. So at this point all that needs to be checked is if it is a with_field.

I will update the comment to make that more clear

runtime/vm/BytecodeInterpreter.hpp
```diff
@@ -8114,8 +8114,91 @@ done:;
...
+				rc = GOTO_THROW_CURRENT_EXCEPTION;
+				goto done;
+			}
+			restoreGenericSpecialStackFrame(REGISTER_ARGS);
+			goto retry;
+		}
+
+		/* Must be done after resolution as we don't know where the object ref is
+		 * until we have resolved the fieldref
+		 */
```
gacholio: This comment appears untrue. It must be done after resolution because resolve errors should take precedence over NPE.

runtime/vm/resolvesupport.cpp
```diff
@@ -741,6 +741,7 @@ resolveStaticFieldRefInto(J9VMThread *vmStruct, J9Method *method, J9ConstantPool
 					}
 					goto done;
 				}
+
```
gacholio: Unnecessary change.

runtime/vm/BytecodeInterpreter.hpp
tajila: Thats true. What I was trying to say is, we wont know if its a single or double slot field until after resolution so we wont know whether objectref is _sp + 1 or _sp + 2

gacholio: Got it - this would be a lot clearer if the ref weren't fetched before the resolution!

tajila:
@gacholio updated the PR

gacholio:

tajila:

PR #3761, opened 2018-11-20
Enable COMPUTED_GOTO on zLinux

babsingh:
`USE_COMPUTED_GOTO` is enabled on `zLinux` when compiling with `gcc-7+`. When
OpenJ9 is built using `gcc-7` on zLinux, `USE_COMPUTED_GOTO` improves startup
performance by `~10%` for DayTrader.

When compiling with `USE_COMPUTED_GOTO` enabled and `gcc version > 4.6`,
`VMINLINE` shouldn't be used with `VM_BytecodeInterpreter::run`. Otherwise,
the following compile error is encountered:
```
BytecodeInterpreter.hpp:8136:2: error: function ‘UDATA
VM_BytecodeInterpreter::run(J9VMThread*)’ can never be copied because it
saves address of local label in a static variable
```

### DIFF
--- a/runtime/vm/BytecodeInterpreter.cpp
+++ b/runtime/vm/BytecodeInterpreter.cpp
@@ -34,11 +34,16 @@
 #define LOOP_NAME bytecodeLoop
 #endif
 
-/* USE_COMPUTED_GOTO on Windows is a performance improvement of 15% */
-/* USE_COMPUTED_GOTO on Linux amd64 has a performance improvement of 3% */
+/* USE_COMPUTED_GOTO on Windows has a performance improvement of 15%.
+ * USE_COMPUTED_GOTO on Linux amd64 has a performance improvement of 3%.
+ * USE_COMPUTED_GOTO is enabled on Linux s390 when compiling with gcc7+.
+ * USE_COMPUTED_GOTO improves startup performance by ~10% on Linux s390.
+ */
 #if (defined(WIN32) && defined(__GNUC__))
 #define USE_COMPUTED_GOTO
 #elif (defined(LINUX) && defined(J9HAMMER))
+#define USE_COMPUTED_GOTO
+#elif (defined(LINUX) && defined(S390) && (__GNUC__ > 7))
 #define USE_COMPUTED_GOTO
 #elif defined(OSX)
 #define USE_COMPUTED_GOTO
--- a/runtime/vm/BytecodeInterpreter.hpp
+++ b/runtime/vm/BytecodeInterpreter.hpp
@@ -8124,7 +8124,7 @@ protected:
 
 public:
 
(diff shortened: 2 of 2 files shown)


babsingh:
fyi - @pshipton @DanHeidinga

runtime/vm/BytecodeInterpreter.cpp
```diff
@@ -38,7 +38,7 @@
 /* USE_COMPUTED_GOTO on Linux amd64 has a performance improvement of 3% */
 #if (defined(WIN32) && defined(__GNUC__))
 #define USE_COMPUTED_GOTO
-#elif (defined(LINUX) && defined(J9HAMMER))
+#elif (defined(LINUX) && (defined(J9HAMMER) || defined(S390)))
```
charliegracie: What happens if you try to compile the BytecodeInterpreter on Linux 390 with xlC? Is this possible with our current makefiles?

runtime/vm/BytecodeInterpreter.hpp
```diff
@@ -8124,7 +8124,7 @@ done:;
 
 public:
 
-#if defined(J9VM_ARCH_X86) && defined(__GNUC__) && ((__GNUC__> 4) || (__GNUC__ == 4 && __GNUC_MINOR__ > 6))
+#if (defined(J9VM_ARCH_X86) || defined(S390)) && defined(__GNUC__) && ((__GNUC__ > 4) || ((__GNUC__ == 4) && (__GNUC_MINOR__ > 6)))
```
charliegracie: At this point do we want to just remove the VMINLINE all of the time? With these changes it will be removed from Linux x86 and Linux 390 when using any new GCC. On OS X with clang and computed goto enabled (the new default) the run method is NOT inlined even with VMINLINE specified.

My assumptions are:
1. now that we compile the BytecodeInterpreter with clang on Windows if computed goto is enabled it will increase performance and force the run method to not be inlined.
2. If we try computed goto on Linux P with GCC 7 computed goto will improve performance as well so we will need to add that to this list of ever growing defines that turn off the VMINLINE
3. Removing the VMINLINE all of the time likely only impacts platforms where the BytecodeInterpreter is compiled with xlC. Are there any of those platforms anymore?

pshipton: AIX and z/OS are the only platforms that compile the BytecodeInterpreter with xlC.
Note Windows isn't using clang yet, not until #2979 is merged.

runtime/vm/BytecodeInterpreter.cpp
pshipton: Not sure this has ever been tried.

pshipton:
jenkins compile zlinux jdk8

pshipton:
jenkins compile zlinux jdk8

pshipton:
jenkins test all zlinux jdk11

PR #3763, opened 2018-11-20
Reload evacuated object in the hardware read barrier helper

fjeremic:
In the case of the unlikely event that the call to J9ReadBarrier fails
to evacuate the object we may end up in an infinite loop because the
hardware read barrier helper currently returns back into the JIT
mainline to re-execute the guarded load instruction.

The JIT was under the assumption that J9ReadBarrier will always succeed
which may not be the case if we completely run out of space on the heap
to evacuate the object. In such a scenario the GC must reach a
synchronization point and issue a full global collection, but to do
that we must be able to return to the JIT mainline and load the
non-evacuated object. This means we cannot re-execute the guarded load
instruction in the JIT mainline.

To avoid this issue we reload the evacuated object in the hardware read
barrier helper and jump past the guarded load instruction upon exiting
the helper.

As a side-effect of this change we can no longer do constrained TX with
hardware concurrent scavenge because the J9ReadBarrier can fail and we
cannot get out of the CTX loop if that happens because the execution
will resume at the beginning of the transaction.

### DIFF
--- a/runtime/vm/zcinterp.m4
+++ b/runtime/vm/zcinterp.m4
@@ -153,10 +153,97 @@ END_CURRENT
 })
 
 ifdef({OMR_GC_CONCURRENT_SCAVENGER},{
-dnl Helper to handle guarded storage events and software read barriers. When
-dnl the event occurs, hardware/software populates the return address in 
-dnl J9VMThread.gsParameters.returnAddr.
-define({HANDLE_GS_EVENT_HELPER},{
+define({HANDLE_HARDWARE_READ_BARRIER},{
+BEGIN_HELPER($1)
+    SAVE_ALL_REGS($1)
+
+dnl Test the TX bit of GSECI (bit 16) to see if the CPU was in transactional-
+dnl execution mode when the guarded storage event was recognized. Subsequently
+dnl test the CX bit of GSECI (bit 17) to see which type of transaction we were
+dnl executing. At that point we have two cases:
+dnl
+dnl 1. TX was a non-constrained transaction:
+dnl    If this was the case do not execute the read barrier and simply return to
+dnl    the fallback transaction abort JIT code to handle it by forcing condition
+dnl    code to be 2 with the below CL(G)FI instruction indicating a transitent 
+dnl    transaction abort.
+dnl
+dnl 2. TX was a constrained transaction:
+dnl    This should not happen because we cannot handle constrained transactions
+dnl    in the hardware read barrier case. Because executing any of the following
+dnl    code will be functionally incorrect we force a crash here so we fail fast
+dnl    and can manually debug.
+dnl
+dnl See the following PRs for more details:
+dnl eclipse/openj9#2545
+dnl eclipse/openj9#2790
+dnl eclipse/openj9#3763
+    TM J9TR_VMThread_gsParameters_GSECI(J9VMTHREAD),128
+    JZ LABEL_NAME(L_GS_CALL_HELPER)
+    TM J9TR_VMThread_gsParameters_GSECI(J9VMTHREAD),64
+    JNZ LABEL_NAME(L_GS_FORCE_CRASH)
+    CLFI_GPR J9SP,0
+    J LABEL_NAME(L_GS_SKIP_HELPER)
+
+PLACE_LABEL(L_GS_CALL_HELPER)
+    ST_GPR J9SP,J9TR_VMThread_sp(J9VMTHREAD)
+    LR_GPR CARG1,J9VMTHREAD
+    L_GPR CRA,J9TR_VMThread_javaVM(J9VMTHREAD)
+    L_GPR CRA,J9TR_JavaVM_invokeJ9ReadBarrier(CRA)
+    CALL_INDIRECT(CRA)
+
+dnl Load the updated object pointer into CARG2
+    LG CARG1,J9TR_VMThread_gsParameters_operandAddr(J9VMTHREAD)
+ifdef({ASM_J9VM_GC_COMPRESSED_POINTERS},{
+dnl Some objects may be stored in a decompressed format on the heap.
+dnl A typical example of this may be static objects. For such objects
+dnl we must not decompress via the compressedPointersShift value. To
+dnl check for this we disassemble the guarded load instruction in the
+dnl mainline and check whether it is an LGG or LLGFSG instruction.
+    LG CARG2,J9TR_VMThread_gsParameters_instructionAddr(J9VMTHREAD)
+    TM 5(CARG2),4
+    JNZ LABEL_NAME(L_GS_SKIP_SHIFT)
+    LLGF CARG1,0(CARG1)
+    LG CARG2,J9TR_VMThread_javaVM(J9VMTHREAD)
+    LG CARG2,J9TR_JavaVM_compressedPointersShift(CARG2)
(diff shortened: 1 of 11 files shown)


runtime/vm/zcinterp.m4
```diff
@@ -153,70 +153,101 @@ END_CURRENT
...
-    HANDLE_GS_EVENT_HELPER($1)
-PLACE_LABEL(L_GS_DONE)
-dnl Branch back to the instruction that triggered guarded storage event.
-    BRANCH_INDIRECT_ON_CONDITION(15,J9TR_VMThread_gsParameters_returnAddr,0,J9VMTHREAD)
-END_CURRENT
-})
+dnl See the following PRs for more details:
+dnl eclipse/openj9#2545
+dnl eclipse/openj9#2790
+dnl eclipse/openj9#9999
```
fjeremic: Uuups. Just realized I left this as an eyecatcher because I didn't know what the PR ID was going to be before opening this. Fixing this momentarily.

fjeremic: Fixed in 6260445

runtime/vm/zcinterp.m4
```diff
@@ -153,70 +153,101 @@ END_CURRENT
...
+
+dnl Extract the target address register
+        LG r8,J9TR_VMThread_gsParameters_instructionAddr(J9VMTHREAD)
+        LLGH r8,0(r8)
+        NILF r8,240
+
+dnl We combine a shift right by 4 and a multiply by 8 to a shift
+dnl right by 1 to get the byte offset into the save area where to
+dnl store the updated object pointer
+        SRLG r8,r8,1
```
joransiu: Minor optimization: You can combine the `NILF r8, 240` and `SRLG r8,r8,1` into a single `RISBGN R8,R8,56,59,63` (might have to double check the indices calculation), which has a nicer bypass into the subsequent STG memref.

runtime/vm/zcinterp.m4
```diff
@@ -153,70 +153,101 @@ END_CURRENT
...
-    ST_GPR J9SP,JIT_GPR_SAVE_SLOT(J9SP)
-    RESTORE_ALL_REGS_AND_SWITCH_TO_JAVA_STACK($1)
-})
+    define({HANDLE_HARDWARE_READ_BARRIER},{
+    BEGIN_HELPER($1)
+        SAVE_ALL_REGS($1)
 
-define({HANDLE_HARDWARE_READ_BARRIER},{
-BEGIN_FUNC($1)
 dnl Test the TX bit of GSECI (bit 16) to see if the CPU was in transactional-
```
joransiu: Should fix up the comment to say we test both TX and CTX bits of GSECI (bit 16 and 17).

charliegracie:
I will let someone else launch the testing as I think there is common, p and z code modified and this is not my area of expertise :)

runtime/vm/zcinterp.m4
fjeremic: I must admit this is mostly laziness on my part since encoding an `RISBGN` requires hand crafting a m4 macro which generates the hex encoding for the instruction. However you are completely correct. I'll take care of this tomorrow.

gacholio:
There's a lot of inconsistent spacing in the m4 changes (spaces instead of tabs, indented ifdef/label directives, etc).

gita-omr:
@IBMJimmyk I imagine software implementation will be different but something we should consider on Power as well.

gita-omr:
Does it mean that JIT essentially should not assume the object was moved and proceed the same way whether it was evacuated or not? @0dvictor will it affect Intel as well?

joransiu:
> Does it mean that JIT essentially should not assume the object was moved and proceed the same way whether it was evacuated or not?

Right.  As I understand, J9ReadBarrier may or may not evacuate the object.  However, it does guarantee on return of a J9ReadBarrier call, the object referenced in the field is valid and not stale.  Hence, it is safe to load the field's value, and proceed after the ReadBarrier call.

0dvictor:
> will it affect Intel as well?

No, on X86, we don't check the range after returning from read barrier and hence do not have this issue.

IBMJimmyk:
I don't think this will be a problem for Power either. Similar to x86, the range is not checked again after returning from the read barrier.

gita-omr:
Thanks, just wanted to make sure.

runtime/vm/zcinterp.m4
fjeremic: Fixed in f95c0c2

runtime/vm/zcinterp.m4
fjeremic: Fixed in f95c0c2

fjeremic:
> There's a lot of inconsistent spacing in the m4 changes (spaces instead of tabs, indented ifdef/label directives, etc).

Fixed in 039c726. Given the changes in f95c0c2 and the often confusing `RISBGN` instruction I'd like to launch another personal build against my change to be 100% certain there is no regression introduced. I'm doing this now. Reviews feel free to proceed further with reviews meanwhile.

runtime/vm/zcinterp.m4
```diff
@@ -205,12 +206,10 @@ dnl Load the updated object pointer into R9
 dnl Extract the target address register
         LG r8,J9TR_VMThread_gsParameters_instructionAddr(J9VMTHREAD)
         LLGH r8,0(r8)
-        NILF r8,240
-
 dnl We combine a shift right by 4 and a multiply by 8 to a shift
 dnl right by 1 to get the byte offset into the save area where to
 dnl store the updated object pointer
-        SRLG r8,r8,1
+		ENCODE_RISBGN(r8,r8,57,60,63)
```
joransiu: You probably should set to set the Z bit to 1 in the I4 operand, to ensure the remaining bits of the first operand are set to zeros.  Otherwise, you might accidentally get some odd bits set for the extra bits loaded by LLGH.

fjeremic: Argh yes good catch, my builds were falling all over the place and this was the cause. Thanks!

Fixed in 892129c

fjeremic:
Removing WIP. @joransiu I have added two new commits to address two previously observed problems during my testing:

1. We cannot use R8 and R9 in the implementation which we were doing before. This is because `SAVE_ALL_REGS` assumes these two registers are valid so that it can reuse them when restoring FPRs via `RESTORE_ALL_REGS_AND_SWITCH_TO_JAVA_STACK`

2. There is a problem with the compressedrefs path if the guarded load was an `LGG` instead of a `LLGFSG` in the mainline. This can happen for static objects which are not compressed. Previously we assumed all objects need to be decompressed which is not true.

All my testing shows there are no more problems. Please review again. @gacholio FYI on 1. as it really tripped me up for a whole day. Perhaps we should be re-calculating R8 and R9 in the restore macro rather than relying on those registers to be correct throughout the entire helper.

gacholio:
r8 and r9 are used to do the vector register enabled check. In the general case, those registers remain live throughout helpers (they are preserved by C calls), so it seemed like a waste to refetch/recompute the mask after the C call.

fjeremic:
OSX and Travis failures are due to infrastructure issues. I think we should be good to go here given the change is only really targeting Z.

fjeremic:
Squashed and rebased to get rid of the extra commits. I've saved the commit messages in the squashed commit for searchability.

PR #3773, opened 2018-11-21
Fix typo in Memory Documentation

dsouzai:
The existing documentation refers to a "J9::SegmentProvider", which
doesn't exist; what it was referring to was a "J9::J9SegmentProvider".
This commit updates the documentation to be correct.

### DIFF
--- a/doc/compiler/memory/MemoryManager.md
+++ b/doc/compiler/memory/MemoryManager.md
@@ -70,11 +70,11 @@ provider of memory.
 --------------------------------------------------------------------
 
 
-                +---------------------+
-                |                     |
-                | J9::SegmentProvider |
-                |                     |
-                +-----+-----------+---+
+                +-----------------------+
+                |                       |
+                | J9::J9SegmentProvider |
+                |                       |
+                +-----+-----------+-----+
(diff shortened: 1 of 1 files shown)


dsouzai:
@mpirvu or @fjeremic could you guys please review and merge?

mpirvu:
No testing needed since this is just doc changes.

dsouzai:
> Is there a reason this class has J9 in front of it, but others like SystemSegmentProvider do not?

I don't know if there is a good reason for it. However, I think it's probably better that it be named this way. There is a `TR::SegmentProvider` which is NOT related by inheritance to `J9::J9SegmentProvider`. As such, having a `J9::SegmentProvider` would, with the fact that OpenJ9/OMR uses extensible classes, be a source of confusion. This problem already exists with `TR::SegmentAllocator` and `J9::SegmentAllocator`; the latter is NOT related to the former via inheritance, and as such, should really be renamed to `J9::J9SegmentAllocator` to remove the confusion.

All in all, I think the entire memory infrastructure probably deserves some effort to rename the various components. Too many people (myself included) have had to deal with the confusion caused by it.

PR #3780, opened 2018-11-21
Handle disabled Known Object Table in invokeFilterArgumentsHandle

dmitry-ten:
Previously, if a filter arguments method handler was compiled, known object table had to be enabled, otherwise the compiler would crash.
However, Known Object Table is only used to improve an existing symbol
reference, it's not required for compilation.
This change enables to run with disabled known object table, by
using a list of filters, instead of a list of filter indices, to check if a null filter is used.

### DIFF
--- a/runtime/compiler/env/j9method.cpp
+++ b/runtime/compiler/env/j9method.cpp
@@ -8338,6 +8338,8 @@ TR_J9ByteCodeIlGenerator::runFEMacro(TR::SymbolReference *symRef)
 
          int32_t startPos;
          TR::KnownObjectTable::Index *filterIndexList;
+         bool *haveFilter = NULL;
+         bool knotEnabled = !comp()->getOption(TR_DisableKnownObjectTable);
          char *nextSignature;
 
             {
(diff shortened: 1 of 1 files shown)


dmitry-ten:
@andrewcraik could you review?
@mpirvu fyi, should I make a PR for JITaaS part now or after the next rebase?

mpirvu:
> should I make a PR for JITaaS part now or after the next rebase?

You can create the PR now. We can cherry pick commits from the mainline so that we don't have to wait for the next rebase.

runtime/compiler/env/j9method.cpp
```diff
@@ -8392,9 +8399,10 @@ TR_J9ByteCodeIlGenerator::runFEMacro(TR::SymbolReference *symRef)
                TR::Node *receiverHandle = pop();
                if (receiverHandle->getOpCode().hasSymbolReference() && receiverHandle->getSymbolReference()->getSymbol()->isArrayShadowSymbol())
                   {
-                  if (thunkDetails->isShareable())
+                  if (!knotEnabled || thunkDetails->isShareable())
```
jdmpapin: I have a general comment about this—not something that needs to be addressed in the code here. If we're somehow able to generate custom thunks without a known object table, and if we're actually doing so, we should stop. Custom thunks are pointless without known objects.

dmitry-ten: Sorry, I don't know enough about custom thunks to tell if what you described is happening.

runtime/compiler/env/j9method.cpp
```diff
@@ -8338,18 +8338,25 @@ TR_J9ByteCodeIlGenerator::runFEMacro(TR::SymbolReference *symRef)
...
-            TR::KnownObjectTable *knot = comp()->getOrCreateKnownObjectTable();
+            filterIndexList = knotEnabled ? (TR::KnownObjectTable::Index *) comp()->trMemory()->allocateMemory(sizeof(TR::KnownObjectTable::Index) * numFilters, stackAlloc) : NULL;
+            filtersList = (uintptrj_t *) comp()->trMemory()->allocateMemory(sizeof(uintptrj_t) * numFilters, stackAlloc);
+            TR::KnownObjectTable *knot = knotEnabled ? comp()->getOrCreateKnownObjectTable() : NULL;
             for (int i = 0; i <numFilters; i++)
                {
-               filterIndexList[i] = knot->getIndex(fej9->getReferenceElement(filters, i));
+               // copy filters in a list, so that we don't have to use filterIndexList
+               // to determine if a filter is null later on
+               filtersList[i] = fej9->getReferenceElement(filters, i);
```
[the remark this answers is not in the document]
dmitry-ten: Good point, done.

dmitry-ten:
@andrewcraik this is ready to be merged.

andrewcraik:

andrewcraik:
@dmitry-ten a merge can happen once sanity has passed.

PR #3782, opened 2018-11-21
Fix for missing NULLCHK related to checkcast on X86

0dvictor:
checkcastAndNULLCHK was not evaluated properly so that it missed
throwing NPE when NULLCHK is required. Fixing.


andrewcraik:

0dvictor:
Copied the incorrect change, now it is fixed.

PR #3783, opened 2018-11-21
Fix MethodHandle related bugs

dmitry-ten:
- fix `filterArguments`, server tried to get and use filter indices
from the known object table, which is not support in JITaaS.
Use pointers to filters instead.

- `FoldHandle_argIndices` would put indices in the reverse order on
the client, and then reverse them again on the server, resulting
in the non-reversed order, which is incorrect. Fix by not reversing on
the client.

- `X86PrivateLinkage::buildIndirectDispatch` recreates thunks every time
we try to get one in JITaaS. It's not clear to me why it's done, because
these thunks are already allocated in code memory, and they do not seem
to be persistent.
I don't think this behavior is entirely correct, because recreating
`invokeExactJ2IThunk` leads to errors in method handle tests.

JITaaS part of: #3780

### DIFF
--- a/runtime/compiler/env/j9method.cpp
+++ b/runtime/compiler/env/j9method.cpp
@@ -8223,8 +8223,10 @@ TR_J9ByteCodeIlGenerator::runFEMacro(TR::SymbolReference *symRef)
             arity = std::get<2>(recv);
             leafClass = std::get<3>(recv);
             const std::string &leafClassNameStr = std::get<4>(recv);
-            leafClassNameChars = (char *) &leafClassNameStr[0];
             leafClassNameLength = leafClassNameStr.length();
+            leafClassNameChars = (char *) alloca(leafClassNameLength + 1);
+            memcpy(leafClassNameChars, leafClassNameStr.data(), leafClassNameLength);
+            leafClassNameChars[leafClassNameLength] = 0;
             isPrimitiveClass = std::get<5>(recv);
             }
          else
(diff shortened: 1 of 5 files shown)


dmitry-ten:
@mpirvu please review.
@ymanton do you know why we regenerate thunks every time we access them, even within the same compilation? Seems like it leads to some errors.

ymanton:
See the comment for commit https://github.com/eclipse/openj9/commit/cac61c17a87c80a7b8fc7136767ac597f95a4a5d.

Based on the comment and my recollection, it was done because we didn't know if the client had the thunk or not. IIRC the regular JIT compiler keeps track of the thunks via a persistent hash table, so creating the same thunk multiple times should not happen, it should be found in the hash table if it was previously created. For JITaaS that doesn't tell us if the current client has the thunk so we always create it so we can send it. If the client has it already it should discard the one the server sends.

That was the situation at the time the change was made, however some things may have changed since then.

dmitry-ten:
Thanks, I'll look into this code.

dmitry-ten:
So the problem was that server would always generate `invokeExactJ2IThunk`, even when it's not needed at all. Resolved this issue by overriding `TR_J9VMBase::needsInvokeExactJ2IThunk` on the server so that it returns true when a thunk is actually needed.
The fact that we can send duplicate thunks is irrelevant, because an old thunk will get overwritten by an identical new thunk on the client.

runtime/compiler/env/VMJ9.cpp
```diff
@@ -4726,6 +4726,9 @@ TR_J9VMBase::needsInvokeExactJ2IThunk(TR::Node *callNode, TR::Compilation *comp)
       && (  method->getMandatoryRecognizedMethod() == TR::java_lang_invoke_MethodHandle_invokeExact
          || method->isArchetypeSpecimen()))
       {
+      if (comp->isOutOfProcessCompilation())
```
mpirvu: If you have overridden this method, while do you still need to modify the base class?

runtime/compiler/env/j9method.cpp
```diff
@@ -8246,8 +8246,10 @@ TR_J9ByteCodeIlGenerator::runFEMacro(TR::SymbolReference *symRef)
             arity = std::get<2>(recv);
             leafClass = std::get<3>(recv);
             const std::string &leafClassNameStr = std::get<4>(recv);
-            leafClassNameChars = (char *) &leafClassNameStr[0];
             leafClassNameLength = leafClassNameStr.length();
+            leafClassNameChars = (char *) alloca(leafClassNameLength + 1);
```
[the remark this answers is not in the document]
dmitry-ten: Yes, buffer for the string will be discarded once we leave the if statement where it's initialized.

runtime/compiler/env/VMJ9.cpp
dmitry-ten: No, this is a typo, forgot to delete it.

dmitry-ten:
Fixed problems

mpirvu:
If this double commit solution is going to create merge problems later on ( @ymanton ), I would prefer to have the mainline done first and then cherry-pick that commit.

ymanton:
I don't think there is a way to avoid conflicts here because you're making more changes on top of the changes you're making to master.

However it looks like even the "common" parts are slightly different between this and https://github.com/eclipse/openj9/pull/3780; it would be easier if you just added that commit to this PR and added commits on top of it to minimize the number of different lines.

dmitry-ten:
@mpirvu rebased and should be ready for merge.

PR #3785, opened 2018-11-21
Improve JITaaSIProfiler caching

harryyu1994:
1. Cache only empty method iprofiler data entries instead both empty and invalidated entries.
2. If whole method iprofiler data sending fails, we should still attempt to send the data for the bcIndex requested.
[skip ci]


harryyu1994:
Older build throughput with sharedclasses: 4062/s to 4218/s range  4140

Older build throughput without sharedclasses: 4331/s to 4431/s range

With this change throughput with sharedclasses: 4222/s to 4399/s range . 4310

With this change throughput without sharedclasses: 4397/s to 4404/s range

It seems like there is some improvement with this change. (4% improvement)

@mpirvu this change is ready for review.

runtime/compiler/control/JITaaSCompilationThread.cpp
```diff
@@ -123,9 +124,9 @@ void handler_IProfiler_profilingSample(JITaaS::J9ClientStream *client, TR_J9VM *
    if (wholeMethodInfo)
       {
       // Serialize all the information related to this method
-      iProfiler->serializeAndSendIProfileInfoForMethod(method, comp, client);
+      iProfiler->serializeAndSendIProfileInfoForMethod(method, comp, client, abort);
       }
-   else  // Send information just for this entry
+   else if (!wholeMethodInfo || abort) // Send information just for this entry
```
mpirvu: Why is this test needed? If `wholeMethodInfo` is false, `!wholeMethodInfo` is going to be true and we enter this `if` statement anyway.

runtime/compiler/control/JITaaSCompilationThread.cpp
```diff
@@ -123,9 +124,9 @@ void handler_IProfiler_profilingSample(JITaaS::J9ClientStream *client, TR_J9VM *
    if (wholeMethodInfo)
       {
       // Serialize all the information related to this method
-      iProfiler->serializeAndSendIProfileInfoForMethod(method, comp, client);
+      iProfiler->serializeAndSendIProfileInfoForMethod(method, comp, client, abort);
```
mpirvu: If you want `abort` to be returned, it's better to just use `return abort` rather than using a parameter passed by reference.

runtime/compiler/runtime/JITaaSIProfiler.cpp
```diff
@@ -514,9 +515,11 @@ TR_JITaaSClientIProfiler::serializeAndSendIProfileInfoForMethod(TR_OpaqueMethodB
       // Walk all bytecodes and populate the sorted array of interesting PCs (pcEntries)
       // numEntries will indicate how many entries have been populated
       // These profiling entries have been 'locked' so we must remember to unlock them
-      bool abort = false;
       bytesFootprint = walkILTreeForIProfilingEntries(pcEntries, numEntries, &bci, method, BCvisit, abort, comp);
-      if (numEntries && !abort)
+      if (abort)
```
mpirvu: Even though abort is set, some of the entries could have been locked, so we need to run the sequence that unlocks them below.

harryyu1994:
Code changed as requested.

Issue #3762, opened 2018-11-20
JITaaS Server not using compressedReferenceShift from JITaaS Client


harryyu1994:
Currently JITaaS Server is not getting compressedReferenceShift information from the client.

Originally discovered by running Liberty AcmeAir, 
Reduced testcase to show the problem:

```
./java -XX:JITaaSServer
^start up a server

./java -XX:JITaaSClient -Xgc:preferredHeapBase=0x100000000 '-Xjit:count=5,disableAsyncCompilation,{java/util/HashMap.putVal(ILjava/lang/Object;Ljava/lang/Object;ZZ)Ljava/lang/Object;}(traceFull,log=trace.log)' -version
^start up a client
```

Client crashes due to incorrect compressed reference shift.

JITaaS Server obtains the compressedReferenceShift information through this call:
`TR::Compiler->om.compressedReferenceShift()`

`TR::Compiler` is a global object of type `CompilerEnv`. Currently Client's `CompilerEnv` is not communicated to Server. Server needs to somehow stop using the global object and instead use a per-client `TR::Compiler` object.

harryyu1994:
Since the JIT gets compressedRefShift info through `TR::Compiler->om.compressedReferenceShift()`, the original plan was to serialize `TR::Compiler`(CompilerEnv object) and send it from Client to Server, and Server caches the CompilerEnv object for each client. However, after I looked into the call `TR::Compiler->om.compressedReferenceShift()`

```
int32_t
J9::ObjectModel::compressedReferenceShift()
   {
#if defined(J9VM_GC_COMPRESSED_POINTERS)
   J9JavaVM *javaVM = TR::Compiler->javaVM;
   if (!javaVM)
      return 0;

   J9VMThread *vmThread = javaVM->internalVMFunctions->currentVMThread(javaVM);
   J9MemoryManagerFunctions * mmf = javaVM->memoryManagerFunctions;
   int32_t result = mmf->j9gc_objaccess_compressedPointersShift(vmThread);
   return result;
#else
   return 0;
#endif

   }
```
It does not look like this function uses anything from the J9::ObjectModel `om` itself. Which means even if Server has Client's version of `TR::Compiler->om`, we still won't get the correct compressedReferenceShift.

harryyu1994:
Will add in a field called "_compressedReferenceShift" here and Server will query this info from ClientSessionData
```
class ClientSessionData
   {
   // This struct contains information about VM that does not change during its lifetime.
   struct VMInfo
[...]

PR #3793, opened 2018-11-21
Store and use client's compressedRefShift at server

harryyu1994:
- Server should be using client's compressedRefShift for compilation
- previously using server's own compressedRefShift
[skip ci]
Issue: #3762


harryyu1994:
@mpirvu Ready for review, confirmed that original issue disappeared with this change.

PR #3798, opened 2018-11-22
Fix for missing NULLCHK under Concurrent Scavenge

0dvictor:
NULLCHK may be missing under concurrent scavenge on non-compressedrefs
build; fixing.


andrewcraik:
@0dvictor Why isn't the NULLCHK case being handled with a simple check of if the treetop is NULLCHK then we need to create a NULLCHK of the grandchild (whatever it is) then anchor the child? I guess why are we only looking for two patterns?

0dvictor:
We only should create NULLCHK when anchoring `ardbari` for these two patterns. When `ardbari` is neither the child nor the grandchild of a NULLCHK node, it should be anchored under `treetop` as it should be evaluated before the grandchild of the NULLCHK node.

andrewcraik:

andrewcraik:
win jdk11 failed due to infra. Everything else is clean and the test did pass on JDK8 win so I'm happy to merge since this should not be jdk specific - just GC mode sensitive.

PR #3801, opened 2018-11-22
Temporarily turn off mode service

renfeiw:
[ci skip]

### DIFF
--- a/test/TestConfig/scripts/testKitGen/makeGenTool/mkgen.pl
+++ b/test/TestConfig/scripts/testKitGen/makeGenTool/mkgen.pl
@@ -61,7 +61,8 @@ sub runmkgen {
 		$testRoot = $output;
 	}
 
-	my $includeModesService = 1;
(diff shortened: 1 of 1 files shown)


renfeiw:
reviewer: @llxia

DanHeidinga:
@llxia @renfeiw Is the mode service used / usable outside of IBM?  If not, this may be code that needs a deeper look to determine it's future.

llxia:
No, mode service is not used outside (as it cannot be accessed outside). 
We are going to have a further discussion with @smlambert to determine its future.

smlambert:
My opinion is that there would be no reason why this 'modes service' could not be used externally or internally, and is part of a larger story of feeding data to the test systems we use from services running in the cloud.  

At the moment, I do not think we should remove it completely.  I would like to explore options to connect to public cloud services (where this and other services could be hosted).

llxia:

PR #3803, opened 2018-11-22
Allow base pointer with zero value

JasonFengJ9:
Allow base pointer with zero value

Only check the base pointer value before it is going to be used.
Remove unnecessary checking of the base pointer to allow stack walking when the base pointer is zero (particularly when `-fomit-frame-pointer` is enabled).

Reviewer: @DanHeidinga 

### DIFF
--- a/debugtools/DDR_VM/src/com/ibm/j9ddr/corereaders/elf/ELFDumpReader.java
+++ b/debugtools/DDR_VM/src/com/ibm/j9ddr/corereaders/elf/ELFDumpReader.java
@@ -1210,8 +1210,10 @@ public abstract class ELFDumpReader implements ILibraryDependentCore
 				instructionPointer = maskInstructionPointer(getLinkRegisterFrom(registers));
 			}
 
-			if (0 != instructionPointer && 0 != basePointer	&& isValidAddress(instructionPointer)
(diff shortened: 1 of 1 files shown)


debugtools/DDR_VM/src/com/ibm/j9ddr/corereaders/elf/ELFDumpReader.java
```diff
@@ -1210,8 +1210,10 @@ public void walkStack() {
 				instructionPointer = maskInstructionPointer(getLinkRegisterFrom(registers));
 			}
 
-			if (0 != instructionPointer && 0 != basePointer	&& isValidAddress(instructionPointer)
```
pshipton: The basePointer check may be important for 32-bit code.

JasonFengJ9: The check for `0 != basePointer` is moved down before the value is actually used. I assume retrieving stack frames via `unwindTable` works regardless the value of `basePointer`. Also verified that `xi3280 jdmpview !nstack` shows native stacks, and passes `VM_Extended.testDTFJ` test.

pshipton:
jenkins test all xlinux jdk8

Issue #3736, opened 2018-11-19
Crash in handleServerMessage


mpirvu:
While trying to run AcmeAir with hot compilations I hit a crash with the following stack trace:
```
#14 0x00007fc7f6514cb9 in handleServerMessage (client=0x7fc7e14c4dd0, fe=0x7fc7d4007300)
    at /home/mpirvu/JITaaS/openj9/runtime/compiler/../compiler/control/JITaaSCompilationThread.cpp:2046
#15 0x00007fc7f65162ae in remoteCompile (vmThread=0x11d9f00, compiler=0x7fc7bb000000, compilee=0x7fc7e14c5a88, method=0x120f588,
    details=..., compInfoPT=0x7fc7f25dc470)
    at /home/mpirvu/JITaaS/openj9/runtime/compiler/../compiler/control/JITaaSCompilationThread.cpp:2138
#16 0x00007fc7f62e922d in TR::CompilationInfoPerThreadBase::compile (this=0x7fc7f25dc470, vmThread=0x11d9f00, compiler=0x7fc7bb000000,
    compilee=0x7fc7e14c5a88, vm=..., optimizationPlan=0x7fc7f25fbdd0, scratchSegmentProvider=...)
    at /home/mpirvu/JITaaS/openj9/runtime/compiler/../compiler/control/CompilationThread.cpp:8602
#17 0x00007fc7f62e7557 in TR::CompilationInfoPerThreadBase::wrappedCompile (portLib=0x7fc7fcc16640 <j9portLibrary>,
--Type <RET> for more, q to quit, c to continue without paging--
    opaqueParameters=0x7fc7e14c5800) at /home/mpirvu/JITaaS/openj9/runtime/compiler/../compiler/control/CompilationThread.cpp:8116
#18 0x00007fc7f7da50a6 in omrsig_protect (portLibrary=0x7fc7fcc16640 <j9portLibrary>,
    fn=0x7fc7f62e4efa <TR::CompilationInfoPerThreadBase::wrappedCompile(J9PortLibrary*, void*)>, fn_arg=0x7fc7e14c5800,
```
The problem is in this code:
```
      case J9ServerMessageType::IProfiler_setCallCount:
         {
         auto recv = client->getRecvData<TR_OpaqueMethodBlock*, int32_t, int32_t>();
         auto method = std::get<0>(recv);
         auto bcIndex = std::get<1>(recv);
         auto count = std::get<2>(recv);
[...]

PR #3806, opened 2018-11-22
Store Client's isIProfilerEnabled info at Server

harryyu1994:
There are cases where a client disables its IProfiler. Server needs to keep track of this information for each client so that it only uses it for client that has IProfiler enabled.
[skip ci]

Closes: #3736


runtime/compiler/env/VMJ9.cpp
```diff
@@ -8337,6 +8338,14 @@ TR_J9VM::getClassFromMethodBlock(TR_OpaqueMethodBlock *mb)
 TR_IProfiler *
 TR_J9VM::getIProfiler()
    {
+   if (auto stream = TR::CompilationInfo::getStream())
```
mpirvu: I would prefer to override TR_J9VM implementation. Is there a reason that is not possible?

PR #3482, opened 2018-10-30
Added x86 Dockerfiles for cent6.9 and ubuntu16

cwesMills:
The cent file can compile JDK 8/11, and run system and functional tests.
The cent file sets max user processes to unlimited in order to avoid an out of resources error.
The ubuntu file is test only, and uses the gdb package to allow generation of core files.

Signed-off-by: Colton Mills <millscolt3@gmail.com>

PR #3812, opened 2018-11-23
Updated x86 cent6 Dockerfile for JDK12. Build jdk11 with jdk11

cwesMills:
JDK11 will now be built using bootJDK11. This allows JDK12 to be built using the same bootJDK. The x86 cent6 Dockerfile, and the default variable file have been updated to reflect this change.

Depends on #3482, and should be merged afterward


cwesMills:
@AdamBrousseau

PR #3814, opened 2018-11-23
WIP: Use correct byte code index on asynccheck

liqunl:
Ilgen may generate an if compare node to combine a compare bytecode and
its suceeding conditional branch bytecode. If an asynccheck node is
generated before the if compare node, it should have the bytecode index
of the compare bytecode. This is important in OSR because asynccheck is
an OSR point whose bytecode index represents a transition target. Having
it wrong will result in the interpreter resuming execution from a wrong
place after an OSR transition.

### DIFF
--- a/runtime/compiler/ilgen/J9ByteCodeIlGenerator.hpp
+++ b/runtime/compiler/ilgen/J9ByteCodeIlGenerator.hpp
@@ -91,7 +91,9 @@ private:
    // GenBranch
    //
    int32_t      genGoto(int32_t);
-   int32_t      genIf(TR::ILOpCodes, bool singleStackOp = false);
+   int32_t      genIfOneOperand(TR::ILOpCodes);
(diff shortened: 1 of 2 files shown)


liqunl:
Marking WIP until the personal builds have finished. @andrewcraik could you take a look?

andrewcraik:
For correctness I would prefer that the bytecode index is correct at the point where we generate the asynccheck tree rather than trying to modify the tree after it is created since that could have a variety of unintended consequences on the OSR infrastructure. It would probably be best to take the top little bit of genIf and turn it into genIfOneOperand, genIfTwoOperand and the asynccheck code for the combined cmp and if can move up into the method where you are currently fixing the tree. Those three places can then call genIfImpl or something like that.

liqunl:
@andrewcraik Created different entry points for genIf. Can you review again?

andrewcraik:

andrewcraik:

PR #3817, opened 2018-11-23
Track unloaded classes on the server

ymanton:
When sending a list of unloaded classes to the server,
the client will also send the range of memory the
methods of each class occupy. This info will be used
to maintain per-client `TR_AddressSet`s of unloaded
classes and methods on the server.

Signed-off-by: Younes Manton <ymanton@ca.ibm.com>

### DIFF
--- a/runtime/compiler/control/JITaaSCompilationThread.cpp
+++ b/runtime/compiler/control/JITaaSCompilationThread.cpp
@@ -38,6 +38,8 @@
 #include "env/j9methodServer.hpp"
 #include "exceptions/AOTFailure.hpp" // for AOTFailure
 
+extern TR::Monitor *assumptionTableMutex;
+
 uint32_t serverMsgTypeCount[JITaaS::J9ServerMessageType_ARRAYSIZE] = {};
 
 size_t methodStringsLength(J9ROMMethod *method)
@@ -201,6 +203,18 @@ bool handleServerMessage(JITaaS::J9ClientStream *client, TR_J9VM *fe)
       case J9ServerMessageType::compilationCode:
          done = true;
          break;
+
+      case J9ServerMessageType::getUnloadedClassRanges:
+         {
+         auto unloadedClasses = comp->getPersistentInfo()->getUnloadedClassAddresses();
+         std::vector<TR_AddressRange> ranges(unloadedClasses->getNumberOfRanges());
+            {
+            OMR::CriticalSection getAddressSetRanges(assumptionTableMutex);
+            unloadedClasses->getRanges(ranges);
+            }
+         client->write(ranges, unloadedClasses->getMaxRanges());
+         break;
+         }
 
       case J9ServerMessageType::VM_isClassLibraryClass:
          {
@@ -2397,7 +2411,9 @@ ClientSessionData::ClientSessionData(uint64_t clientUID, uint32_t seqNo) :
    _chTableClassMap(decltype(_chTableClassMap)::allocator_type(TR::Compiler->persistentAllocator())),
    _romClassMap(decltype(_romClassMap)::allocator_type(TR::Compiler->persistentAllocator())),
    _J9MethodMap(decltype(_J9MethodMap)::allocator_type(TR::Compiler->persistentAllocator())),
-   _systemClassByNameMap(decltype(_systemClassByNameMap)::allocator_type(TR::Compiler->persistentAllocator()))
+   _systemClassByNameMap(decltype(_systemClassByNameMap)::allocator_type(TR::Compiler->persistentAllocator())),
+   _unloadedClassAddresses(NULL),
+   _requestUnloadedClasses(true)
    {
    updateTimeOfLastAccess();
    _javaLangClassPtr = nullptr;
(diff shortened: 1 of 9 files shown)


runtime/compiler/control/JITaaSCompilationThread.hpp
```diff
@@ -130,6 +132,10 @@ class ClientSessionData
    int8_t  _inUse;  // Number of concurrent compilations from the same client 
                     // Accessed with compilation monitor in hand
    VMInfo *_vmInfo; // info specific to a client VM that does not change, nullptr means not set
+
+   // Per-client versions of the unloaded class and method addresses kept in J9PersistentInfo
+   TR_AddressSet _unloadedClassAddresses;
+   TR_AddressSet _unloadedMethodAddresses;
```
mpirvu: We don't need _unloadedMethodAddresses at the server. The reason is that this is used only in `isInUnloadedMethod` which is only used in `invalidateEntryIfInconsistent` that looks at `entry->getPC()`. We already pass lists of unloaded classes from the client to server and clean-up the IP hashtable at the server (we invalidate all j9mthods for a class that gets unloaded)

runtime/compiler/control/HookedByTheJit.cpp
```diff
@@ -4106,7 +4106,9 @@ static void jitHookClassLoad(J9HookInterface * * hookInterface, UDATA eventNum,
          allocFailed = true;
       }
 
-   compInfo->getPersistentInfo()->ensureUnloadedAddressSetsAreInitialized();
+   // In JITaaS server mode unloaded addresses are maintained per client
+   if (compInfo->getPersistentInfo()->getJITaaSMode() != SERVER_MODE)
```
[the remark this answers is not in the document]
ymanton: I checked that the hook was being called on the server and it is unfortunately, presumably because bootstrap classes are still loaded. There are various other JITaaS mode checks in that code so it seems to have been the case all along at least.

Anyway I've moved the check into `ensureUnloadedAddressSetsAreInitialized()` instead to make sure the single persistent instances are never created on the server regardless of where it's called from.

ymanton:
Updated the patch to get the complete state of unloaded classes from the server when needed, as well as dropping the tracking of unloaded methods, and various other cleanups.

runtime/compiler/runtime/ClassUnloadAssumption.cpp
```diff
@@ -1280,6 +1292,23 @@ J9::PersistentInfo::addUnloadedClass(
...
+   {
+   jitPersistentFree(_addressRanges);
+   }
+
+void TR_AddressSet::getRanges(std::vector<TR_AddressRange> &ranges)
+   {
+   ranges.insert(ranges.begin(), _addressRanges, _addressRanges + _numAddressRanges);
+   }
+
+void TR_AddressSet::setRanges(std::vector<TR_AddressRange> &ranges)
```
mpirvu: Parameter 'ranges' can be made const here.

runtime/compiler/runtime/ClassUnloadAssumption.cpp
```diff
@@ -1280,6 +1292,23 @@ J9::PersistentInfo::addUnloadedClass(
...
+   {
+   TR_ASSERT(ranges.size() <= _maxAddressRanges, "Setting too many ranges");
```
mpirvu: This is very tricky. `_maxAddressRanges` could be increased at the client with a command line option, so the ASSERT is not enough.
I think that the dynamic allocation of _unloadedClassAddresses would have helped here: have a null pointer and allocate the object only `_maxAddressRanges` is known which happens when the message `getUnloadedClassRanges` is sent in `processUnloadedClasses`.

runtime/compiler/control/JITaaSCompilationThread.cpp
```diff
@@ -2415,20 +2431,35 @@ ClientSessionData::~ClientSessionData()
...
+      _unloadedClassAddresses.setRanges(unloadedClassRanges);
+      _requestUnloadedClasses = false;
+      }
+
    OMR::CriticalSection processUnloadedClasses(getROMMapMonitor());
-   for (TR_OpaqueClassBlock *clazz : classes)
+
+   for (auto clazz : classes)
       {
+      _unloadedClassAddresses.add((uintptrj_t)clazz);
```
mpirvu: This line could be avoided when `_requestUnloadedClasses` was found true because we just initialized the most up-to-date set of ranges.

runtime/compiler/runtime/ClassUnloadAssumption.cpp
```diff
@@ -1236,9 +1241,16 @@ J9::PersistentInfo::isUnloadedClass(
...
-   bool result = (_unloadedClassAddresses && _unloadedClassAddresses->mayContain((uintptrj_t)v));
-   return result;
+   if (getJITaaSMode() != SERVER_MODE)
+      {
+      OMR::CriticalSection isUnloadedClass(assumptionTableMutex);
+      return _unloadedClassAddresses && _unloadedClassAddresses->mayContain((uintptrj_t)v);
+      }
+   else
+      {
+      OMR::CriticalSection isUnloadedClass(TR::compInfoPT->getClientData()->getROMMapMonitor());
```
mpirvu: TR::compInfoPT uses thread local storage which is more expensive to access and here we call it twice in a row. I would prefer to cache the clientSession to a local and use only one TLS access.

runtime/compiler/runtime/IProfiler.cpp
```diff
@@ -1208,6 +1208,10 @@ TR_IPMethodHashTableEntry::add(TR_OpaqueMethodBlock *caller, TR_OpaqueMethodBloc
 bool
 TR_IProfiler::invalidateEntryIfInconsistent(TR_IPBytecodeHashTableEntry *entry)
    {
+   // Invalid entries are purged elsewhere in JITaaS server mode
+   if (_compInfo->getPersistentInfo()->getJITaaSMode() == SERVER_MODE)
```
mpirvu: Instead of this check we could make this method virtual and redefine it as an empty one in TR_JITaaSIProfiler

runtime/compiler/runtime/RuntimeAssumptions.hpp
```diff
@@ -229,6 +229,10 @@ class TR_AddressSet
       _maxAddressRanges(maxAddressRanges)
       {}
 
+   void destroy();
+   void getRanges(std::vector<TR_AddressRange> &ranges);
+   void setRanges(std::vector<TR_AddressRange> &ranges);
```
mpirvu: As mentioned elsewhere the `ranges` vector could be marked `const` for `setRanges`

Issue #3342, opened 2018-10-18
MacOS: sigxfszHandlingTests fail


pdbain-ibm:
```
[transcript.txt](https://github.com/eclipse/openj9/files/2492049/transcript.txt)
Testing: -XX:+HandleSIGXFSZ
Running command: "/Users/pdbain/vm/openj9-openjdk-jdk11/build/macosx-x86_64-normal-server-release/images/jdk/bin//java" -Xcompressedrefs  -cp "/Users/pdbain/macprojects/openj9test/openjdk-tests/TestConfig/scripts/testKitGen/../../../../jvmtest/functional/cmdline_options_testresources/cmdlinetestresources.jar" -XX:+HandleSIGXFSZ j9vm.test.sigxfsz.SignalXfszTest test.txt
Test result: FAILED
Output from test:
 [OUT] Starting SignalXfszTest
 [OUT] File is written successfully
>> Success condition was not found: [Output match: Failed to write the file]
```

babsingh:
- [Test source code](https://github.com/eclipse/openj9/blob/master/test/functional/cmdline_options_testresources/src/j9vm/test/sigxfsz/SignalXfszTest.java)
- [Test parameters](https://github.com/eclipse/openj9/blob/master/test/functional/cmdLineTests/sigxfszHandlingTest/sigxfszHandlingTest.xml#L45)

babsingh:
More details about this issue:
- For the `SIGXFSZ` tests to successfully work, `ulimit -f 4000` needs to be set.
- `ulimit -f 4000` restricts file size so that `SIGXFSZ` is raised. Then, tests evaluate JVM's behavior w.r.t `-XX:+/-HandleSIGXFSZ`.
- On OSX, `ulimit -f 4000` is not persistent. `sigxfszHandlingTest` suite has five tests. `ulimit -f 4000` is set only once before running the five tests. After the first test is run, `ulimit -f` automatically changes from `4000` to `unlimited` on OSX. This causes the remaining four tests to fail.

Solution: Set `ulimit -f 4000` before running each test.

Issue #3779, opened 2018-11-21
sigxfszHandlingTest should restore the orignal ulimit setting when finished


hangshao0:
The file size ulimit is changed by the sigxfszHandlingTest 
https://github.com/eclipse/openj9/blob/b2bb724bfcbe817f3106000c521eb84be6b9b133/test/functional/cmdLineTests/sigxfszHandlingTest/playlist.xml#L32

But the test did not restore the original value after it is finished. It might cause future test to fail.

hangshao0:
FYI: @llxia

llxia:
Thanks @hangshao0

This test needs to be updated to run in 3 steps:
- store the default `ulimit` value
- run test
- reset `ulimit` value back to the stored value (regardless actual test result)

PR #3818, opened 2018-11-23
Fix sigxfszHandlingTests on OSX

babsingh:
For the `SIGXFSZ` tests to successfully work, file size limit (`ulimit -f 4000`) 
needs to be set. Restricting file size limit causes `SIGXFSZ` to be
raised. On OSX, `ulimit -f 4000` is not persistent. `ulimit -f 4000` is
invoked only once before running the `SIGXFSZ` tests. After the first
`SIGXFSZ` test is run, file size limit automatically changes from `4000` to
`unlimited` on OSX. This causes the remaining `SIGXFSZ` tests to fail.

Instead of setting file size limit once before running all `SIGXFSZ` tests,
file size limit (`ulimit -f 4000`) will be set before running each `SIGXFSZ`
test in order to fix `SIGXFSZ` test failures on OSX.

On Linux, file size limit (`ulimit -f 4000`) is persistent. This can cause
`non-SIGXFSZ` tests to fail. Now onwards, each `SIGXFSZ` test will be
run in a new shell. Thus, `non-SIGXFSZ` tests won't be impacted by 
`ulimit -f 4000`. `runSIGXFSZTest.sh` has been added to run `SIGXFSZ`
tests in a new shell.

`SIGXFSZ` tests are not run on Windows since `SIGXFSZ` is not supported on
Windows. So, `*.bat` variant of `runSIGXFSZTest.sh` is not needed.

Fixes: https://github.com/eclipse/openj9/issues/3342
Fixes: https://github.com/eclipse/openj9/issues/3779

### DIFF
--- a/test/functional/cmdLineTests/sigxfszHandlingTest/sigxfszHandlingTest.xml
+++ b/test/functional/cmdLineTests/sigxfszHandlingTest/sigxfszHandlingTest.xml
@@ -32,7 +32,7 @@
 <!-- Add -Xshareclasses:none to the command lines of this test suite to ensure that signal SIGXFSZ is triggered on test.txt creation rather than 
 	on shared cache file creation (if shared classes is enabled by default). See issue: https://github.com/eclipse/openj9/issues/3333 -->
  <test id="Default">
-	<command>$EXE$ -Xshareclasses:none $CP$ j9vm.test.sigxfsz.SignalXfszTest test.txt</command>
+	<command>$RUN_SCRIPT$ $TESTDIR$$PATHSEP$runSIGXFSZTest$SCRIPT_SUFFIX$ "$EXE$ -Xshareclasses:none $CP$ j9vm.test.sigxfsz.SignalXfszTest test.txt"</command>
 	<output regex="no" type="required">Starting SignalXfszTest</output>
 	<output regex="no" type="success">Failed to write the file</output>
 	<output regex="no" type="required">java.io.IOException</output>
@@ -45,7 +45,7 @@
  </test>
 
  <test id="-XX:+HandleSIGXFSZ">
-	<command>$EXE$ -Xshareclasses:none $CP$ $HANDLE$ j9vm.test.sigxfsz.SignalXfszTest test.txt</command>
+	<command>$RUN_SCRIPT$ $TESTDIR$$PATHSEP$runSIGXFSZTest$SCRIPT_SUFFIX$ "$EXE$ -Xshareclasses:none $CP$ $HANDLE$ j9vm.test.sigxfsz.SignalXfszTest test.txt"</command>
 	<output regex="no" type="required">Starting SignalXfszTest</output>
 	<output regex="no" type="success">Failed to write the file</output>
 	<output regex="no" type="required">java.io.IOException</output>
@@ -58,7 +58,7 @@
  </test>
 
 <test id="-XX:-HandleSIGXFSZ -XX:+HandleSIGXFSZ">
-	<command>$EXE$ -Xshareclasses:none $CP$ $NOHANDLE$ $HANDLE$ j9vm.test.sigxfsz.SignalXfszTest test.txt</command>
+	<command>$RUN_SCRIPT$ $TESTDIR$$PATHSEP$runSIGXFSZTest$SCRIPT_SUFFIX$ "$EXE$ -Xshareclasses:none $CP$ $NOHANDLE$ $HANDLE$ j9vm.test.sigxfsz.SignalXfszTest test.txt"</command>
 	<output regex="no" type="required">Starting SignalXfszTest</output>
 	<output regex="no" type="success">Failed to write the file</output>
 	<output regex="no" type="required">java.io.IOException</output>
@@ -71,7 +71,7 @@
  </test>
 
 <test id=" -Xrs">
-	<command>$EXE$ -Xshareclasses:none $CP$ -Xrs j9vm.test.sigxfsz.SignalXfszTest test.txt</command>
+	<command>$RUN_SCRIPT$ $TESTDIR$$PATHSEP$runSIGXFSZTest$SCRIPT_SUFFIX$ "$EXE$ -Xshareclasses:none $CP$ -Xrs j9vm.test.sigxfsz.SignalXfszTest test.txt"</command>
 	<output regex="no" type="required">Starting SignalXfszTest</output>
 	<output regex="no" type="success">Failed to write the file</output>
 	<output regex="no" type="required">java.io.IOException</output>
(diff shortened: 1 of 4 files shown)


babsingh:
fyi - @llxia

babsingh:
fyi - @llxia @DanHeidinga @pshipton

llxia:
we actually have an issue about this test (#3779). If the test does not set `ulimit` back to original value, future tests may fail.

DanHeidinga:
@babsingh ulimit settings should persist for the life of the shell and it's children.  Are we creating new shells between each of these commands?

babsingh:
@DanHeidinga No new shells are created. 

`ulimit` behavior on OSX:
```
ulimit -f 4000
ulimit <- 4000
java -version 
ulimit <- unlimited
```

`ulimit` behavior on Linux:
```
ulimit -f 4000
ulimit <- 4000
java -version 
ulimit <- 4000
```

@llxia for #3779, i will update the PR to cache and restore the `ulimit` value after running the test.

babsingh:
Adding `ulimit -f 4000 && ulimit` in `sigxfszHandlingTest.xml` (`<command>...</command>`) doesn't work. `ulimit` can't be invoked from `sigxfszHandlingTest.xml`. fyi - @llxia 

On OSX, resource limits are set using `launchctl`, `ulimit`, and `kern.<VARIABLES`. Also, instructions on setting resource limits vary depending upon OSX version:
1) [open-files-limit/#mac-os-x-el-capitan](https://docs.basho.com/riak/kv/2.0.8/using/performance/open-files-limit/#mac-os-x-el-capitan)
2) [open-files-limit/#mac-os-x-yosemite](https://docs.basho.com/riak/kv/2.0.8/using/performance/open-files-limit/#mac-os-x-yosemite)
3) [open-files-limit/#mac-os-x-older-versions](https://docs.basho.com/riak/kv/2.0.8/using/performance/open-files-limit/#mac-os-x-older-versions)

I didn't find a resource online which explains the following `ulimit` behavior:
```
ulimit -f 4000
ulimit -a # file size = 4000
ls
ulimit -a # file size <- unlimited
```

For `pwd`, `ulimit` behavior is different:
```
ulimit -f 4000
ulimit -a # file size = 4000
pwd
ulimit -a # file size = 4000
```

**Guess:** If `file size (-f)` setting ` from ulimit` is insufficient for a process, then the global `file size (-f)` setting `from launchctl` is used to override the local `file size (-f)` setting after the process completes.

llxia:
rm https://github.com/eclipse/openj9/pull/3818#issuecomment-441303991, I do not see this strange behavior with hotspot on my mac (Mojave):

```
lans-mbp:jdk_mac lanxia$ ulimit
unlimited
lans-mbp:jdk_mac lanxia$ ulimit -f 4000
lans-mbp:jdk_mac lanxia$ ulimit
4000
lans-mbp:jdk_mac lanxia$ java -version
java version "1.8.0_181"
Java(TM) SE Runtime Environment (build 1.8.0_181-b13)
Java HotSpot(TM) 64-Bit Server VM (build 25.181-b13, mixed mode)
lans-mbp:jdk_mac lanxia$ ulimit
4000
```

llxia:
>Adding ulimit -f 4000 && ulimit in sigxfszHandlingTest.xml (<command>...</command>) doesn't work. ulimit can't be invoked from sigxfszHandlingTest.xml.


could you try to use `<exec command=`? For example:

https://github.com/eclipse/openj9/blob/d779403041e4f03b37a617a5990f8f147d0b657c/test/functional/cmdLineTests/callsitedbgddrext/callsiteddrtests.xml#L64-L71

babsingh:
> I do not see this strange behavior with hotspot 

I think the strange behavior will only be seen if file operations, which violate the file size limit, are involved. `pwd` won't show the strange behavior whereas `ls` will demonstrate the strange behavior.

> could you try to use <exec command=?

`<exec command=...` command doesn't work. `ulimit` setting doesn't persist till `<command>...</command>` is invoked. `<exec command=...` and `<command>...` are invoked in separate shells.

llxia:
Should we have the same behavior as hotspot in this case?

babsingh:
> Should we have the same behavior as hotspot in this case?

no. you can reproduce the strange behavior in hotspot as well:
```
ulimit -f 4000
ulimit # file size = 4000
java -Xffff
ulimit # file size = unlimited
```
you need to involve file operations that violate the file size limit to reproduce the strange behavior in hotspot.

llxia:
We can use a script in `<command>`. Something like: 
```
<command>sh /path/to/test.sh</command>
```
You can define your test inside `test.sh`.

Simliar to https://github.com/eclipse/openj9/blob/b48b6c77b66f5a7fee0b4e0a2d5188ec3be5e558/test/functional/cmdLineTests/URLClassLoaderTests/URLClassLoaderTests.xml#L53

babsingh:
WIP removed. Test has been updated to fix #3342 and #3779. Pull request is ready for merge. fyi - @llxia @DanHeidinga.

test/functional/cmdLineTests/sigxfszHandlingTest/sigxfszHandlingTest.xml
```diff
@@ -84,7 +84,7 @@
  </test>
 
  <test id="-XX:+HandleSIGXFSZ -Xrs">
-	<command>$EXE$ -Xshareclasses:none $CP$ $HANDLE$ -Xrs j9vm.test.sigxfsz.SignalXfszTest test.txt</command>
+	<command>$RUN_SCRIPT$ $TESTDIR$$PATHSEP$runSIGXFSZTest$SCRIPT_SUFFIX$ "$EXE$ -Xshareclasses:none $CP$ $HANDLE$ -Xrs j9vm.test.sigxfsz.SignalXfszTest test.txt"</command>
```
DanHeidinga: Some of the tests are platform limited like the one below - `platforms="linux.*">`.  I'm assuming these tests only run on unix platforms so a `!#/bin/sh` script is valid.

babsingh: Yes, these tests only run on Unix platforms. So,`!#/bin/sh` is valid. Checked a AIX and a zOS machine to confirm `!#/bin/sh`. Existing test code for Unix platforms also utilizes `!#/bin/sh`. Example: 
https://github.com/eclipse/openj9/blob/451b01ecbe5276c0b566faae9adda9338c5d950e/test/functional/cmdLineTests/URLClassLoaderTests/Sanity/buildVerify.sh#L1

DanHeidinga:
@llxia I'll defer to you on this one.

llxia:
jenkins test sanity xlinux,plinux,aix,zlinux jdk11

llxia:
jenkins test sanity osx jdk11

llxia:
jenkins test sanity xlinux,aix jdk8

llxia:
`cmdLineTest_sigxfszHandlingTest_0` passed in osx PR build. The failures in osx builds are not related to this PR.

PR #3823, opened 2018-11-23
Preventing out-of-order updates in JITaaS

mpirvu:
The client sends class unloading and CHTable updates with each compilation
request and the JITaaS server needs to apply those updates in the order
they were generated. Unfortunatelly, the updates can arrive out-of-order
at the server and this can lead to functional problems. The solution
implemented by this commit uses a sequencing scheme based on ticket numbers.

The client obtains a sequence number (a ticket) that defines the update
and attaches it to the compilation request.
The server obtains a pointer to the clientSession and then enters a
critical section protected by the "sequencingMonitor" where compilation
requests are reordered. The clientSession maintains an "expectedSeqNo"
variable and only a request with a matching sequence number can pass
through. Any request with a larger sequence number is put onto a
waiting list (we say that handling threads are parked). Waiting is done
on a monitor associated with the entry (this scheme allows us to pick
which threads to wake up).
Any request with a smaller sequence number is cancelled (this can happen
at the very beginning when a later compilation request arrives first and
creates the clientSession).
The thread that goes through applies its updates to the CHTable and the
hashtable with ramClass-->romClass mappings and then increments
expectedSeqNo to allow the next thread to pass. Also, it needs to notify
any potential thread that is parked waiting for its turn.
The first thread that goes through the sequencing critical section and
finds an empty CHTable is responsible for fetching the entire CHTable
from the client (this is done only once).

A big complication is introduced by messages that are lost. If a message
never arrives, the threads that are parked are going to time-out, but
since the lost message could carry crucial information, these threads
cannot continue safely. In order to avoid cancelling these compilation
requests, when the thread at the front of the waiting list times-out,
it waits for any active threads (threads that actually perform
compilations) to drain out, it clears the caches for this clientSession
and then it changes the expectedSeqNo to allow itself to go through the
sequencing critical section. Parked threads are notified as before.

### DIFF
--- a/runtime/compiler/control/CompilationRuntime.hpp
+++ b/runtime/compiler/control/CompilationRuntime.hpp
@@ -951,6 +951,9 @@ public:
    void setClientSessionHT(ClientSessionHT *ht) { _clientSessionHT = ht; }
    PersistentVector<TR_OpaqueClassBlock*> *getUnloadedClassesTempList() { return _unloadedClassesTempList; }
    void setUnloadedClassesTempList(PersistentVector<TR_OpaqueClassBlock*> *it) { _unloadedClassesTempList = it; }
+   TR::Monitor *getSequencingMonitor() { return _sequencingMonitor; }
+   uint32_t getCompReqSeqNo() const { return _compReqSeqNo; }
+   uint32_t incCompReqSeqNo() { return ++_compReqSeqNo; }
    PersistentUnorderedMap<TR_OpaqueClassBlock*, uint8_t> *getNewlyExtendedClasses() { return _newlyExtendedClasses; }
    void classGotNewlyExtended(TR_OpaqueClassBlock* clazz)
       {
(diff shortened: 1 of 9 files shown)


ymanton:
The implementation as described sounds OK to me, I can't think of anything that's amiss, and I didn't spot anything wrong with the code changes either.

But a somewhat related issue comes to mind; if an earlier compilation request never arrives we'll have issues accurately tracking unloaded classes won't we? Clearing the caches is sufficient for the caches, but the unloaded class addresses range set will have missed an update and wont it?

mpirvu:
> Clearing the caches is sufficient for the caches, but the unloaded class addresses range set will have missed an update and wont it?

This code was implemented before class address ranges came into the picture. We have to deal with them in them in the same way we deal with CHTable updates, i.e. at the beginning we need the full content of the class address ranges, not just an update. This is also needed when the server dies and a new one is started. The new server needs to receive the entire content of the CHTable and class address ranges.

PR #3828, opened 2018-11-25
Fix Inlined Method Relocations

dsouzai:
This PR adds a few related fixes for inlined method relocations:

1. A check to guard against getting -1 when acquring a method pointer from the inline site table in the metadata.
2. When using the SVM, inlined site table is relocated regardless of whether the relocation failed.

### DIFF
--- a/runtime/compiler/runtime/RelocationRecord.cpp
+++ b/runtime/compiler/runtime/RelocationRecord.cpp
@@ -1528,7 +1528,7 @@ TR_RelocationRecordDataAddress::findDataAddress(TR_RelocationRuntime *reloRuntim
          {
          ramMethod = reloRuntime->method();
          }
-      if (ramMethod && (ramMethod != static_cast<J9Method *>(-1)))
+      if (ramMethod && (ramMethod != reinterpret_cast<J9Method *>(-1)))
          address = (uint8_t *)jitCTResolveStaticFieldRefWithMethod(vmThread, ramMethod, cpindex, false, &fieldShape);
       }
 
@@ -2157,31 +2157,31 @@ TR_RelocationRecordInlinedMethod::fixInlinedSiteInfo(TR_RelocationRuntime *reloR
 void
 TR_RelocationRecordInlinedMethod::preparePrivateData(TR_RelocationRuntime *reloRuntime, TR_RelocationTarget *reloTarget)
    {
-   J9Method *ramMethod = NULL;
-   bool failValidation = true;
+   TR_OpaqueMethodBlock *ramMethod = NULL;
+   bool inlinedSiteIsValid = inlinedSiteValid(reloRuntime, reloTarget, &ramMethod);
 
-   if (validateClassesSame(reloRuntime, reloTarget, ((TR_OpaqueMethodBlock **)&ramMethod)) &&
-       !reloRuntime->options()->getOption(TR_DisableCHOpts))
+   if (ramMethod)
       {
       // If validate passes, no patching needed since the fall-through path is the inlined code
-      fixInlinedSiteInfo(reloRuntime, reloTarget, (TR_OpaqueMethodBlock *) ramMethod);
-      failValidation = false;
+      fixInlinedSiteInfo(reloRuntime, reloTarget,ramMethod);
       }
 
    TR_RelocationRecordInlinedMethodPrivateData *reloPrivateData = &(privateData()->inlinedMethod);
-   reloPrivateData->_ramMethod = (TR_OpaqueMethodBlock *) ramMethod;
-   reloPrivateData->_failValidation = failValidation;
-   RELO_LOG(reloRuntime->reloLogger(), 5, "\tpreparePrivateData: ramMethod %p failValidation %d\n", ramMethod, failValidation);
+   reloPrivateData->_ramMethod = ramMethod;
+   reloPrivateData->_failValidation = !inlinedSiteIsValid;
+   RELO_LOG(reloRuntime->reloLogger(), 5, "\tpreparePrivateData: ramMethod %p inlinedSiteIsValid %d\n", ramMethod, inlinedSiteIsValid);
    }
 
 
 bool
-TR_RelocationRecordInlinedMethod::validateClassesSame(TR_RelocationRuntime *reloRuntime, TR_RelocationTarget *reloTarget, TR_OpaqueMethodBlock **theMethod)
+TR_RelocationRecordInlinedMethod::inlinedSiteValid(TR_RelocationRuntime *reloRuntime, TR_RelocationTarget *reloTarget, TR_OpaqueMethodBlock **theMethod)
    {
+   J9Method *currentMethod = NULL;
+   bool inlinedSiteIsValid = true;
    J9Method *callerMethod = (J9Method *) getInlinedSiteCallerMethod(reloRuntime);
    if (callerMethod == (J9Method *)-1)
       {
-      RELO_LOG(reloRuntime->reloLogger(), 6, "\tvalidateClassesSame: caller failed relocation so cannot validate inlined method\n");
+      RELO_LOG(reloRuntime->reloLogger(), 6, "\tinlinedSiteValid: caller failed relocation so cannot validate inlined method\n");
       *theMethod = NULL;
       return false;
       }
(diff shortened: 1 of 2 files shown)


dsouzai:

@mpirvu could you please review https://github.com/eclipse/openj9/pull/3828/commits/be6c96d34160d6decbbc5af269eb35b63e060640 and also verify that this commit https://github.com/eclipse/openj9/pull/3828/commits/e77958aa2636a7bcf0581ae3e28b0a276d9a9a7e#diff-ddfa88e7e962afc796f166ed3cce5d5eR933 is accurate?

dsouzai:
> Is it possible to acquire/release monitor for each relocation individually?

Hm, I think it might be possible to do it here https://github.com/eclipse/openj9/pull/3828/files#diff-ddfa88e7e962afc796f166ed3cce5d5eL144 . 

What I'll do though, is move the class table mutex commit to its own PR.

runtime/compiler/runtime/RelocationRecord.cpp
```diff
@@ -2157,30 +2184,36 @@ void
...
-   reloPrivateData->_failValidation = failValidation;
-   RELO_LOG(reloRuntime->reloLogger(), 5, "\tpreparePrivateData: ramMethod %p failValidation %d\n", ramMethod, failValidation);
+   reloPrivateData->_failValidation = !inlinedSiteIsValid;
+   RELO_LOG(reloRuntime->reloLogger(), 5, "\tpreparePrivateData: ramMethod %p inlinedSiteIsValid %d\n", ramMethod, inlinedSiteIsValid);
    }
 
 
 bool
-TR_RelocationRecordInlinedMethod::validateClassesSame(TR_RelocationRuntime *reloRuntime, TR_RelocationTarget *reloTarget, TR_OpaqueMethodBlock **theMethod)
+TR_RelocationRecordInlinedMethod::inlinedSiteValid(TR_RelocationRuntime *reloRuntime, TR_RelocationTarget *reloTarget, TR_OpaqueMethodBlock **theMethod)
```
jdmpapin: Both this method and its caller (`preparePrivateData`) are working with `J9Method*`. Perhaps this should accept `J9Method **theMethod` instead of `TR_OpaqueMethodBlock**`?

runtime/compiler/runtime/RelocationRecord.cpp
```diff
@@ -1528,7 +1554,8 @@ TR_RelocationRecordDataAddress::findDataAddress(TR_RelocationRuntime *reloRuntim
          {
          ramMethod = reloRuntime->method();
          }
-      address = (uint8_t *)jitCTResolveStaticFieldRefWithMethod(vmThread, ramMethod, cpindex, false, &fieldShape);
+      if (ramMethod && (ramMethod != static_cast<J9Method *>(-1)))
```
jdmpapin: `reinterpret_cast`, and similarly at the other occurrence

runtime/compiler/runtime/RelocationRecord.cpp
dsouzai: It turns out that `preparePrivateData` initialized a `J9Method` but always worked with a `TR_OpaqueMethodBlock`, so I just updated `preparePrivateData` to use an opaque method block.

dsouzai:
@jdmpapin could you re-review; I made the requested changes.

dsouzai:
@mstoodle could you take a look at this?

runtime/compiler/runtime/RelocationRecord.cpp
```diff
@@ -2157,31 +2157,31 @@ TR_RelocationRecordInlinedMethod::fixInlinedSiteInfo(TR_RelocationRuntime *reloR
...
+   bool inlinedSiteIsValid = inlinedSiteValid(reloRuntime, reloTarget, &ramMethod);
 
-   if (validateClassesSame(reloRuntime, reloTarget, ((TR_OpaqueMethodBlock **)&ramMethod)) &&
-       !reloRuntime->options()->getOption(TR_DisableCHOpts))
+   if (ramMethod)
       {
       // If validate passes, no patching needed since the fall-through path is the inlined code
-      fixInlinedSiteInfo(reloRuntime, reloTarget, (TR_OpaqueMethodBlock *) ramMethod);
-      failValidation = false;
+      fixInlinedSiteInfo(reloRuntime, reloTarget,ramMethod);
```
mstoodle: I had a concern with this approach where the bug exposing this problem was due to sharing a symbol reference between two different inlined sites. So it seemed like a safer fix would be to either 1) not share the symbol reference between different inlined sites, or 2) only share "up": i.e. never use the symbol reference for an "inner" method in a reference in an "outer" method.

However, while chatting with Irwin, I convinced myself that 1 isn't desirable since it introduces a difference between AOT and JIT that would pre-req alias analysis to overcome, and 2 isn't always possible because you could have two "sibling" inlines that end up sharing a symbol reference between them. In this case, there's no "right" symbol reference to necessarily choose.

So, I like this solution best of the alternatives, I think. Thanks for putting it together!

runtime/compiler/runtime/RelocationRecord.cpp
```diff
@@ -904,10 +904,32 @@ TR_RelocationRecordWithInlinedSiteIndex::ignore(TR_RelocationRuntime *reloRuntim
...
+         {
+         // -1 means this inlined method isn't active, so can ignore relocations associated with it
+         return true;
+         }
+      }
 
+   /* It is safe to return true here because if classes were unloaded, then
+    * the compilation will be aborted and the potentially unrelocated sections
+    * of code will never be executed.
+    */
```
mstoodle: nice informational comment, thanks for that!

mstoodle:

dsouzai:
Most of the osx tests look like they're failing of an error when trying to create the SCC file on disk. 

`cmdLineTest_sigxfszHandlingTest_0` isn't AOT specific (since it's being run with `-Xshareclasses:none`)

There are some failures though such as
```
Testing: gste001_multiple
Test start time: 2018/11/30 12:42:43 Eastern Standard Time
Running command: "/Users/jenkins/workspace/PullRequest-Sanity-JDK11-osx_x86-64_cmprssptrs-OpenJ9/build/macosx-x86_64-normal-server-release/images/jdk/bin/java" -Xcompressedrefs -Xjit -Xgcpolicy:gencon -Xshareclasses:cacheRetransformed,name=retransformenabled  -Xdump    -Xjit:count=0 -agentlib:jvmtitest=test:gste001 -agentlib:jvmtitest=test:gste001 -cp "/Users/jenkins/workspace/PullRequest-Sanity-JDK11-osx_x86-64_cmprssptrs-OpenJ9/openj9/test/TestConfig/scripts/testKitGen/../../../../jvmtest/functional/cmdLineTests/jvmtitests/jvmtitest.jar" com.ibm.jvmti.tests.util.TestRunner
Time spent starting: 19 milliseconds
Time spent executing: 3065 milliseconds
Test result: FAILED
Output from test:
 [ERR] java.lang.NullPointerException
 [ERR] 	at java.base/java.util.Set.of(Set.java:693)
 [ERR] 	at java.base/jdk.internal.module.Builder.requires(Builder.java:193)
 [ERR] 	at java.base/jdk.internal.module.SystemModules$default.moduleDescriptors(Unknown Source)
 [ERR] 	at java.base/jdk.internal.module.SystemModuleFinders.of(SystemModuleFinders.java:146)
 [ERR] 	at java.base/jdk.internal.module.ModuleBootstrap.boot(ModuleBootstrap.java:193)
 [ERR] 	at java.base/java.lang.ClassLoader.initializeClassLoaders(ClassLoader.java:217)
 [ERR] 	at java.base/java.lang.Thread.initialize(Thread.java:422)
 [ERR] 	at java.base/java.lang.Thread.<init>(Thread.java:153)
>> Success condition was not found: [Return code: 0]
```
which I've seen when trying to reproduce https://github.com/eclipse/openj9/issues/3333#issuecomment-440327270 . It isn't something that is caused by the SVM; I see the issue happening with the "old" AOT. I guess I'll have to sort that issue out before we're able to deliver this PR.

mstoodle:
Thanks @dsouzai! You're right, we won't be able to merge this PR until the discovered issues have been resolved, even if they turn out not to be related to your changes I'm afraid. Please reach out if you require any assistance.

dsouzai:
@mstoodle https://github.com/eclipse/omr/pull/3303 has promoted to `openj9-omr:openj9`. Could you please retest and merge?

PR #3834, opened 2018-11-26
Allow executables to allocate memory below 4GB

babsingh:
Native executables, that invoke OpenJ9 in compressedrefs mode, should be
able to allocate memory below 4GB. But, __PAGEZERO segment takes the
first 4GB by default on OSX 64-bit architectures.

With -pagezero_size 0x1000, __PAGEZERO segment size is reduced to 4KB.
This allows OpenJ9 in compressedrefs mode to allocate memory below 4GB.

-pagezero_size 0x1000 can be used for both compressedrefs and
non-compressedrefs variants. This will allow executables to work with
both compressedrefs and non-compressedrefs OpenJ9 librarires.

Fixes: https://github.com/eclipse/openj9/issues/3795

### DIFF
--- a/runtime/makelib/targets.mk.osx.inc.ftl
+++ b/runtime/makelib/targets.mk.osx.inc.ftl
@@ -48,6 +48,9 @@ $(UMA_EXETARGET): $(UMA_OBJECTS) $(UMA_TARGET_LIBRARIES)
 UMA_BEGIN_DASH_L =
 UMA_END_DASH_L =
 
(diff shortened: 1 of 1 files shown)


babsingh:
fyi @DanHeidinga

DanHeidinga:
Jenkins test sanity osx jdk11

DanHeidinga:
Tests failures are known issues.  Merging as is.

PR #3845, opened 2018-11-26
Fix a bug in SymbolValidationManager

dmitry-ten:
SymbolValidationManager retrieves a pointer to boolean array classes
from javaVM struct and then extracts a component classes from them.
In JITaaS, the pointer to array classes is taken from the server VM,
which is incorrect and leads to crashes.
Fix by making a remote call to the client.

Should fix:
DaaLoadTest_daa1_ConcurrentScavenge_0
DaaLoadTest_daa2_ConcurrentScavenge_0
DaaLoadTest_daa3_ConcurrentScavenge_0

### DIFF
--- a/runtime/compiler/runtime/SymbolValidationManager.cpp
+++ b/runtime/compiler/runtime/SymbolValidationManager.cpp
@@ -69,14 +69,17 @@ TR::SymbolValidationManager::SymbolValidationManager(TR::Region &region, TR_Reso
      _seenSymbolsSet((SeenSymbolsComparator()), _region)
    {
    TR::Compilation* comp = TR::comp();
-   J9VMThread *vmThread = comp->j9VMThread();
-   TR_J9VM *fej9 = (TR_J9VM *)TR_J9VMBase::get(vmThread->javaVM->jitConfig, vmThread);
-
+   TR_J9VM *fej9 = comp->fej9vm();
    struct J9Class ** arrayClasses = &fej9->getJ9JITConfig()->javaVM->booleanArrayClass;
    uint16_t id;
    for (int32_t i = 4; i <= 11; i++)
       {
-      TR_OpaqueClassBlock *arrayClass = reinterpret_cast<TR_OpaqueClassBlock *>(arrayClasses[i - 4]);
+      TR_OpaqueClassBlock *arrayClass;
+      if (TR::CompilationInfo::getStream())
+         // JITaaS: shouldn't use array classes from the server VM, get them from client
+         arrayClass = fej9->getClassFromNewArrayType(i);
+      else
+         arrayClass = reinterpret_cast<TR_OpaqueClassBlock *>(arrayClasses[i - 4]);
       TR_OpaqueClassBlock *component = fej9->getComponentClassFromArrayClass(arrayClass);
 
       id = getNewSymbolID();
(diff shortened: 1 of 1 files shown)


dmitry-ten:
@mpirvu could you review?

runtime/compiler/runtime/SymbolValidationManager.cpp
```diff
@@ -1534,11 +1537,19 @@ bool
...
    {
    TR::Compilation* comp = TR::comp();
-   J9VMThread *vmThread = comp->j9VMThread();
-   TR_J9VM *fej9 = (TR_J9VM *)TR_J9VMBase::get(vmThread->javaVM->jitConfig, vmThread);
+   TR_J9VM *fej9 = comp->fej9vm();
+   TR_OpaqueClassBlock *arrayClass;
+   if (TR::CompilationInfo::getStream())
+      {
+      // JITaaS: shouldn't use array classes from the server VM, get them from client
+      arrayClass = fej9->getClassFromNewArrayType(arrayClassIndex);
```
mpirvu: Same observation as above.

dmitry-ten:
I agree that this change should be in the mainline.
However, I've just realized that `SymbolValidationManager` is part of a new class validation infrastructure for AOT, and in the current implementation of JITaaS, the only method of validation manager called on the server is the constructor.
Maybe we should not initialize a `SymbolValidationManager` in JITaaS mode, that would make all other bugs go away.

mpirvu:
We will need the symbol validation manager when we will generate AOT at the server. As a temporary workaround we can disable it for now, but in parallel I would like those changes pushed into mainline. This will prevent other people copy/paste code that needs JITaaS changes.

runtime/compiler/runtime/SymbolValidationManager.cpp
```diff
@@ -69,14 +69,17 @@ TR::SymbolValidationManager::SymbolValidationManager(TR::Region &region, TR_Reso
      _seenSymbolsSet((SeenSymbolsComparator()), _region)
    {
    TR::Compilation* comp = TR::comp();
-   J9VMThread *vmThread = comp->j9VMThread();
-   TR_J9VM *fej9 = (TR_J9VM *)TR_J9VMBase::get(vmThread->javaVM->jitConfig, vmThread);
-
+   TR_J9VM *fej9 = comp->fej9vm();
```
dsouzai: There's a reason the SVM used `TR_J9VM *fej9 = (TR_J9VM *)TR_J9VMBase::get(vmThread->javaVM->jitConfig, vmThread);` instead of `TR_J9VM *fej9 = comp->fej9vm();`.

The front end from the comp is the `TR_J9SharedCacheVM`; this will result in problems when calling `getComponentClassFromArrayClass` below because it will invoke the SVM (which is still being constructed). In general, the SVM uses the `TR_J9VM` object specifically because the query it's making is not something that requires a validation (because the SVM is performing the validation).

dsouzai:
@dmitry-ten , after talking with @mpirvu , the best solution looks to be not creating the SVM if the current compilation is an AOT compile or AOT load. ie, in the constructor of `J9::Compilation`

```
if (((TR_J9VMBase *)fe)->canUseSymbolValidationManager() || optimizationPlan->getIsAotLoad())
   _symbolValidationManager = new (self()->region()) TR::SymbolValidationManager(self()->region(), compilee);
else
   _symbolValidationManager = NULL;
```

runtime/compiler/runtime/SymbolValidationManager.cpp
dmitry-ten: I see. That means when we start generating AOT at the server, we will have to do an if statement to get the server VM, every time `TR_J9VMBase::get` is called on the server.

dsouzai: Well I think when AOT is generated at the server, you're going to have to use the base JaaS VM class that gets the information from the client. We'll have to think about how to structure the `TR_J9SharedCacheVM` and a potential `TR_JaaSSharedCacheVM` to prevent duplicated code.

dmitry-ten:
Alright, then this check for AOT compile/load should probably go in the mainline code, I guess?

dsouzai:
> Alright, then this check for AOT compile/load should probably go in the mainline code, I guess?
Yeah I think; what you have at https://github.com/eclipse/openj9/pull/3855 is good, you can either add another commit to that PR or create a new one.

PR #3849, opened 2018-11-26
WIP: Wrap gc suballocator size check in CompressedPointers scope

fengxue-IS:

### DIFF
--- a/runtime/gc_modron_startup/mminit.cpp
+++ b/runtime/gc_modron_startup/mminit.cpp
@@ -1558,7 +1558,9 @@ combinationMemoryParameterVerification(J9JavaVM *javaVM, IDATA* memoryParameters
 	bool opt_XmosSet = (-1 != memoryParameters[opt_Xmos]);
 	bool opt_XmoxSet = (-1 != memoryParameters[opt_Xmox]);
 	bool opt_XsoftmxSet = (-1 != memoryParameters[opt_Xsoftmx]);
+#if defined(J9VM_GC_COMPRESSED_POINTERS)
 	bool opt_XmcrsSet = (-1 != memoryParameters[opt_Xmcrs]);
+#endif /* J9VM_GC_COMPRESSED_POINTERS */
 
 	PORT_ACCESS_FROM_JAVAVM(javaVM);
 
@@ -2285,6 +2287,7 @@ combinationMemoryParameterVerification(J9JavaVM *javaVM, IDATA* memoryParameters
 	}
 #endif /* J9VM_GC_VLHGC */
 
+#if defined(J9VM_GC_COMPRESSED_POINTERS)
 	if (opt_XmcrsSet) {
 		/* Silently handle a size mismatch; don't report an error about undocumented options
 		 * if the user has specified the official one. */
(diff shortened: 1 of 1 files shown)


runtime/gc_modron_startup/mminit.cpp
```diff
@@ -1558,7 +1558,9 @@ combinationMemoryParameterVerification(J9JavaVM *javaVM, IDATA* memoryParameters
 	bool opt_XmosSet = (-1 != memoryParameters[opt_Xmos]);
 	bool opt_XmoxSet = (-1 != memoryParameters[opt_Xmox]);
 	bool opt_XsoftmxSet = (-1 != memoryParameters[opt_Xsoftmx]);
+#if defined(J9VM_GC_COMPRESSED_POINTERS)
```
charliegracie: What is the documentation for this option?

dmitripivkine:
What the reason for this change? Suballocator as feature is not limited (but happen to be used only) to Compressed Refs spec currently.

runtime/gc_modron_startup/mminit.cpp
dmitripivkine:
```
Sets an initial size for an area in memory that is reserved for any native classes, monitors, and threads that are used by compressed references within the lowest 4 GB memory area.
```
https://www.ibm.com/support/knowledgecenter/SSYKE2_8.0.0/openj9/xmcrs/index.html

fengxue-IS:
@dmitripivkine The suballocator values are currently only initialized for compressed refs specs in OMR GC library, this check however runs unconditionally which fails when running with -Xcheck:memory.

I proposed this change in openj9 since OMR only used the suballocator values within compressed refs scope. (therefore correct by it self)

dmitripivkine:
>The suballocator values are currently only initialized for compressed refs specs in OMR GC library, this check however runs unconditionally which fails when running with -Xcheck:memory.
> 
Please provide more details how it relates to failure when running with -Xcheck:memory

dmitripivkine:
ok, I see where the problem is:
in OMR in GCExtensionsBase.hpp
```
	uintptr_t suballocatorInitialSize; /**< the initial chunk size in bytes for the J9Heap suballocator (enabled with the -Xgc:suballocatorInitialSize option) */
	uintptr_t suballocatorCommitSize; /**< the commit size in bytes for the J9Heap suballocator (enabled with the -Xgc:suballocatorCommitSize option) */
```
are defined unconditionally however an initialization of these variables in Constructor done for Compressed Refs only:
```
#if defined(OMR_GC_COMPRESSED_POINTERS)
		, suballocatorInitialSize(SUBALLOCATOR_INITIAL_SIZE) /* default for J9Heap suballocator initial size is 200 MB */
		, suballocatorCommitSize(SUBALLOCATOR_COMMIT_SIZE) /* default for J9Heap suballocator commit size is 50 MB */
```
So in my opinion proposed fix is a hack and it is not good enough. 
We should fix root of the problem and make code consistent. In order to do this we should make initialization of these variables in Constructor unconditional for all platforms () which require definitions for SUBALLOCATOR_INITIAL_SIZE and SUBALLOCATOR_COMMIT_SIZE be for all platforms too (or alternatively set both to 0). Another possibility is opposite: make variables be defined for CR only (and as result add missed define checks everywhere missed). 
I am voting for first solution because it makes code more general.
@charliegracie What is your opinion?

charliegracie:
-Xcheck:memory fails because those two values are not initialized?  Can you provide more details here?

dmitripivkine:
Fixed at OMR side  eclipse/omr#3250

PR #3855, opened 2018-11-27
Improve SymbolValidationManager compatibility with JITaaS

dmitry-ten:
Compiler code should avoid using attributes of javaVM,
and use front-end instead, because it makes code compatible with JITaaS.
In this case, accessing javaVM->booleanArrayClass will lead to bugs on
JITaaS server, because classes from the server VM will be accessed, but
for compilation we need classes from the client.
On the other hand, using front-end call is fine, because JITaaS subclasses `TR_J9VM`,
and will fetch classes from the client.

Also add a guard to only initialize SVM when doing an AOT compile or load.

Related to: #3845

### DIFF
--- a/runtime/compiler/codegen/J9CodeGenerator.cpp
+++ b/runtime/compiler/codegen/J9CodeGenerator.cpp
@@ -2842,20 +2842,23 @@ J9::CodeGenerator::processRelocations()
             }
          }
 
-      TR::SymbolValidationManager::SymbolValidationRecordList &validationRecords = self()->comp()->getSymbolValidationManager()->getValidationRecordList();
-      if (!validationRecords.empty() && self()->comp()->getOption(TR_UseSymbolValidationManager))
+      if (self()->comp()->getOption(TR_UseSymbolValidationManager))
          {
-         // Add the flags in TR_AOTMethodHeader on the compile run
-         J9JITDataCacheHeader *aotMethodHeader = (J9JITDataCacheHeader *)self()->comp()->getAotMethodDataStart();
-         TR_AOTMethodHeader *aotMethodHeaderEntry = (TR_AOTMethodHeader *)(aotMethodHeader + 1);
-         aotMethodHeaderEntry->flags |= TR_AOTMethodHeader_UsesSymbolValidationManager;
-
-         for (auto it = validationRecords.begin(); it != validationRecords.end(); it++)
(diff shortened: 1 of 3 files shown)


dmitry-ten:
@mpirvu @dsouzai could you review?

dmitry-ten:
I found that in `J9CodeGenerator.cpp` SVM was being accessed before we check if it's enabled,
which was causing a crash now that we don't always initialize it. Moved the guard up

mpirvu:
I am not entirely happy with the code right now because it creates a specific frontend:
`TR_J9VM *fej9 = (TR_J9VM *)TR_J9VMBase::get(vmThread->javaVM->jitConfig, vmThread);`
This means that we cannot use any virtual calls for JITaaS and we'll need to add a test for JITaaS server when we'll allow the server to produce AOT code. However, I am sure we'll forget about this code by then and go through a painful debug process.

dsouzai:
Perhaps `TR_J9VMBase::get` could be modified to return the JaaS version of `TR_J9VM` when running under JaaS? That would fix any issues from an SVM point of view. However, that all depends on where the validation/relocation happens. If it happens on the client side, then I suppose the client shouldn't use the JaaS version of `TR_J9VM`.

jdmpapin:
> I am not entirely happy with the code right now because it creates a specific frontend:
```TR_J9VM *fej9 = (TR_J9VM *)TR_J9VMBase::get(vmThread->javaVM->jitConfig, vmThread);```
This means that we cannot use any virtual calls for JITaaS and we'll need to add a test for JITaaS server when we'll allow the server to produce AOT code. However, I am sure we'll forget about this code by then and go through a painful debug process.

The trouble is that, as @dsouzai mentioned in #3845, `SymbolValidationManager` can't use the default front-end for the compilation, because in an AOT compilation it will be a `TR_J9SharedCacheVM` which will inappropriately call back into the SVM.

> Perhaps `TR_J9VMBase::get` could be modified to return the JaaS version of `TR_J9VM` when running under JaaS? That would fix any issues from an SVM point of view. However, that all depends on where the validation/relocation happens (in an AOTaaS scenario). If it happens on the client side, then I suppose the client shouldn't use the JaaS version of `TR_J9VM`.

One of my forthcoming changes is to grab the front-end only once, when constructing the SVM. After that there will be a single location where SVM can be modified to choose which (non-AOT) `TR_J9VM` should be used, without needing to change `TR_J9VMBase::get()`.

Issue #3248, opened 2018-10-12
AIX performance is affected by -g option added to support DDR


pshipton:
https://github.com/eclipse/openj9/pull/2309 added the -g compile option on AIX. This affected performance and needs to be removed.

The following will be reverted and then -g can be removed.
ibmruntimes/openj9-openjdk-jdk8#116
ibmruntimes/openj9-openjdk-jdk9#178
ibmruntimes/openj9-openjdk-jdk10#59
ibmruntimes/openj9-openjdk-jdk11#23
ibmruntimes/openj9-openjdk-jdk#35
eclipse/openj9#2939

PR #3856, opened 2018-11-27
Add extra assertions for atomic non-helpers on Power

hzongaro:
If a call to codegen inlineDirectCall from directCallEvaluator does not inline a call, added an assertion that the symbol is not one of the atomic non-helpers.

### DIFF
--- a/runtime/compiler/p/codegen/J9TreeEvaluator.cpp
+++ b/runtime/compiler/p/codegen/J9TreeEvaluator.cpp
@@ -12482,6 +12482,17 @@ TR::Register *J9::Power::TreeEvaluator::directCallEvaluator(TR::Node *node, TR::
 
    if (!cg->inlineDirectCall(node, returnRegister))
       {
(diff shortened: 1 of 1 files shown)


hzongaro:
@gita-omr Gita, this is a follow-on to [OMR pull request #3248](https://github.com/eclipse/omr/pull/3248), and it depends that pull request.  This adds an assertion to verify that if a method was not inlined by code gen., it must not be an atomic non-helper that code gen. claims to support.

See also your [comment in PR 3086](https://github.com/eclipse/openj9/pull/3086#issuecomment-439155813) in case you think this should be done differently.

**cc:**  Victor @0dvictor, Filip @fjeremic - FYI

hzongaro:
Now that the pull request that this depends on -- [OMR PR 3248](https://github.com/eclipse/omr/pull/3248) -- has been merged, I've removed the "WIP" from the title.

gita-omr:
Did not realize jenkens tests have not been ran yet.

gita-omr:

PR #3731, opened 2018-11-19
 Update Q Type support throughout VM

acrowthe:
Where appropriate, add support for Q types.
Previously, many locations check for L types, either through character comparisons or through an ascii character map.
Q types have been added to these checks when appropriate, so that they will be treated like objects as opposed to causing crashes or unexpected behavior.

Signed-off-by: Andrew Crowther <acrowthe3388@gmail.com>

PR #3857, opened 2018-11-27
Implement bytecode verification for new bytecodes

acrowthe:
Update bytecode verification to support bytecodes:
- 203, defaultvalue
- 204, withfield


acrowthe:
Currently trying to trace a related issue that my initial testing didn't catch

runtime/bcverify/rtverify.c
```diff
@@ -1786,6 +1795,16 @@ verifyBytecodes (J9BytecodeVerificationData * verifyData)
 				/* put a uninitialized object of the correct type on the stack */
 				PUSH(BCV_SPECIAL_NEW | (start << BCV_CLASS_INDEX_SHIFT));
 				break;
+			
+#if defined(J9VM_OPT_VALHALLA_VALUE_TYPES)
+			case JBdefaultvalue: /* TODO: ValueTypes perform actual verification */
+				index = PARAM_16(bcIndex, 1);
+				info = &constantPool[index];
+				utf8string = J9ROMSTRINGREF_UTF8DATA((J9ROMStringRef *) info);
+				type = convertClassNameToStackMapType(verifyData, J9UTF8_DATA(utf8string), J9UTF8_LENGTH(utf8string), 0, 0);
```
tajila: this can be `pushClassType`

runtime/bcverify/rtverify.c
```diff
@@ -1437,6 +1440,13 @@ verifyBytecodes (J9BytecodeVerificationData * verifyData)
 					/* Jazz 82615: Save the location of receiver */
```
tajila: in this part of the code it can only be putstatic, putfield or withfield. So you can do something like `bc != JBputstatic` since saving the location of the receiver is common for both putfield and withfield

runtime/bcverify/bcverify.c
```diff
@@ -1674,13 +1674,20 @@ simulateStack (J9BytecodeVerificationData * verifyData)
...
+#endif
+			) {
 				DROP(1);
 				if ((*J9UTF8_DATA(utf8string) == 'D') || (*J9UTF8_DATA(utf8string) == 'J')) {
 					DROP(1);
 				}
-
+#if defined(J9VM_OPT_VALHALLA_VALUE_TYPES)
+				if (JBwithfield == bc) {
+					stackTop = pushFieldType(verifyData, utf8string, stackTop);
```
tajila: this should also push receiver type

runtime/bcverify/bcverify.c
```diff
@@ -1668,19 +1668,27 @@ simulateStack (J9BytecodeVerificationData * verifyData)
...
+#endif
+			) {
 				DROP(1);
 				if ((*J9UTF8_DATA(utf8string) == 'D') || (*J9UTF8_DATA(utf8string) == 'J')) {
 					DROP(1);
 				}
-
+#if defined(J9VM_OPT_VALHALLA_VALUE_TYPES)
+				if (JBwithfield == bc) {
+					type = POP;
```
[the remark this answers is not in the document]
tajila: my mistake, should not be putfield, receiver should be saved instead

tajila:
@DanHeidinga please review this PR

runtime/bcutil/jsrinliner.c
```diff
@@ -1501,7 +1501,12 @@ evaluateCodeBlock (I_32 hasRET, J9JSRICodeBlock ** blockParentPointer, U_32 star
 			break;
 
 		default:
-			if (bc > CFR_BC_jsr_w) {
+#if defined(J9VM_OPT_VALHALLA_VALUE_TYPES)
```
DanHeidinga: I don't think this change is needed as JSRs are illegal in Java 7+ classfiles and these new bytecodes won't be in old classfiles.

DanHeidinga: The entry point to this function is `inlineJSRs` which is only called if the method has been tagged to have a JSR.... and clearly a method with valhalla bytecodes won't

runtime/bcverify/rtverify.c
```diff
@@ -1786,6 +1794,15 @@ verifyBytecodes (J9BytecodeVerificationData * verifyData)
 				/* put a uninitialized object of the correct type on the stack */
 				PUSH(BCV_SPECIAL_NEW | (start << BCV_CLASS_INDEX_SHIFT));
 				break;
+			
+#if defined(J9VM_OPT_VALHALLA_VALUE_TYPES)
+			case JBdefaultvalue: /* TODO: ValueTypes perform actual verification */
+				index = PARAM_16(bcIndex, 1);
+				info = &constantPool[index];
+				utf8string = J9ROMSTRINGREF_UTF8DATA((J9ROMStringRef *) info);
+				stackTop = pushClassType(verifyData, utf8string, stackTop);
```
DanHeidinga: Does the type being pushed here include the `Q`?  How does it differentiate from pushing the `L` version of the class?

acrowthe:
Specification for value types can be found here:
http://cr.openjdk.java.net/~fparain/L-world/LW2-JVMS-draft-20181009.pdf
Bytecode specification begins on page 51.

Note that as this specification process is ongoing, this link will eventually go out of date.

runtime/bcverify/rtverify.c
acrowthe: Normally, when instantiating a new `L` object, it will first be pushed (`JBnew`), then later initialized. `JBdefaultValue` should push a reference to a allocated instance of the value class which has been initialized to default values.

As was noted in #3731, `vrfyhelp.c`'s `pushClassType` method does not differentiate between Q and L types.

acrowthe: Looks like `utf8string` does not contain `Q` or `;`, just the class name. However, `convertClassNameToStackMapType` which does the eventual conversion from string to type searches for the romClass matching the name. As a result, the Q or L type data should be retrievable at the point where it is required.

genie-openj9:

PR #3859, opened 2018-11-27
Health Check for Jitaas

satbir-ca:
The idea is to have some sort of validation that
JITaaS is doing its job and add a thread which print
the JITaas Statistics regularly with some tunable frequency.

### DIFF
--- a/runtime/compiler/control/HookedByTheJit.cpp
+++ b/runtime/compiler/control/HookedByTheJit.cpp
@@ -78,6 +78,8 @@
 #include "env/SystemSegmentProvider.hpp"
 #include "control/JITaaSCompilationThread.hpp"
 #include "runtime/JITaaSIProfiler.hpp"
+#include "runtime/Listener.hpp"
+#include "runtime/StatisticsThread.hpp"
 
 extern "C" {
 struct J9JavaVM;
@@ -4766,6 +4768,12 @@ void JitShutdown(J9JITConfig * jitConfig)
    if (!vm->isAOT_DEPRECATED_DO_NOT_USE())
       stopSamplingThread(jitConfig);
 
+   TR_StatisticsThread *statisticsThread = ((TR_JitPrivateConfig*)(jitConfig->privateConfig))->statisticsThread;
+   if (statisticsThread)
+      {
+      statisticsThread->stopStatisticsThread(jitConfig);
+      }
+
    if (jitConfig->runtimeFlags & J9JIT_DUMP_STATS)
       dumpStats(jitConfig);
 
@@ -7098,11 +7106,17 @@ int32_t setUpHooks(J9JavaVM * javaVM, J9JITConfig * jitConfig, TR_FrontEnd * vm)
    
    if (compInfo->getPersistentInfo()->getJITaaSMode() == SERVER_MODE)
       {
-      TR_Listener *listener = ((TR_JitPrivateConfig*)(jitConfig->privateConfig))->listener; 
+      TR_Listener *listener = ((TR_JitPrivateConfig*)(jitConfig->privateConfig))->listener;
       listener->startListenerThread(javaVM);
 
       if (TR::Options::getVerboseOption(TR_VerboseJITaaS))
          TR_VerboseLog::writeLineLocked(TR_Vlog_JITaaS, "Started JITaaSServer listener thread: %p ", listener->getListenerThread());
+
+      if (jitConfig->samplingFrequency != 0)
+         {
+         TR_StatisticsThread *statisticsThread = ((TR_JitPrivateConfig*)(jitConfig->privateConfig))->statisticsThread;
+         statisticsThread->startStatisticsThread(javaVM);
+         }
 
       // Give the JIT a chance to do stuff after the VM is initialized
       if ((*vmHooks)->J9HookRegisterWithCallSite(vmHooks, J9HOOK_VM_INITIALIZED, jitHookVMInitialized, OMR_GET_CALLSITE(), NULL))
--- a/runtime/compiler/env/VMJ9.h
+++ b/runtime/compiler/env/VMJ9.h
@@ -44,7 +44,6 @@
 #include "env/J9SharedCache.hpp"
 #include "infra/Array.hpp"
 #include "env/CompilerEnv.hpp"
-#include "runtime/Listener.hpp"
 #include "env/PersistentCollections.hpp"
 
 class TR_CallStack;
@@ -62,6 +61,8 @@ class TR_ExternalProfiler;
 class TR_JitPrivateConfig;
 class TR_DataCacheManager;
 class TR_EstimateCodeSize;
+class TR_Listener;
+class TR_StatisticsThread;
 struct TR_CallSite;
 struct TR_CallTarget;
 namespace J9 { class ObjectModel; }
@@ -139,6 +140,7 @@ typedef struct TR_JitPrivateConfig
    TR_HWProfiler *hwProfiler;
    TR_JProfilerThread  *jProfiler;
    TR_Listener   *listener;
+   TR_StatisticsThread   *statisticsThread;
    TR_LMGuardedStorage *lmGuardedStorage;
    TR::CodeCacheManager *codeCacheManager; // reachable from JitPrivateConfig for kca's benefit
    TR_DataCacheManager *dcManager;  // reachable from JitPrivateConfig for kca's benefit
--- a/runtime/compiler/runtime/StatisticsThread.cpp
+++ b/runtime/compiler/runtime/StatisticsThread.cpp
@@ -1,0 +1,174 @@
+/*******************************************************************************
+ * Copyright (c) 2018, 2018 IBM Corp. and others
+ *
+ * This program and the accompanying materials are made available under
+ * the terms of the Eclipse Public License 2.0 which accompanies this
+ * distribution and is available at https://www.eclipse.org/legal/epl-2.0/
+ * or the Apache License, Version 2.0 which accompanies this distribution and
+ * is available at https://www.apache.org/licenses/LICENSE-2.0.
+ *
+ * This Source Code may also be made available under the following
+ * Secondary Licenses when the conditions for such availability set
+ * forth in the Eclipse Public License, v. 2.0 are satisfied: GNU
+ * General Public License, version 2 with the GNU Classpath
+ * Exception [1] and GNU General Public License, version 2 with the
+ * OpenJDK Assembly Exception [2].
+ *
+ * [1] https://www.gnu.org/software/classpath/license.html
+ * [2] http://openjdk.java.net/legal/assembly-exception.html
+ *
+ * SPDX-License-Identifier: EPL-2.0 OR Apache-2.0 OR GPL-2.0 WITH Classpath-exception-2.0 OR LicenseRef-GPL-2.0 WITH Assembly-exception
+ *******************************************************************************/
+
+#include "runtime/StatisticsThread.hpp"
+#include "env/VMJ9.h"
+#include "runtime/CompileService.hpp"
+#include "control/JITaaSCompilationThread.hpp"
+
+TR_StatisticsThread::TR_StatisticsThread()
+   : _statisticsThread(NULL), _statisticsThreadMonitor(NULL), _statisticsOSThread(NULL),
+   _statisticsThreadAttachAttempted(false), _statisticsThreadExitFlag(false), _statisticsFrequency(TR::Options::getStatisticsFrequency())
+   {
+   }
+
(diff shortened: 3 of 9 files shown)


satbir-ca:
@mpirvu Please review the code for the Statistics thread for JITaaS.

This is work in progress and still need to add the option for enabling the statistics thread and to control statistics printing frequency.

runtime/compiler/runtime/StatisticsThread.cpp
```diff
@@ -0,0 +1,197 @@
...
+      syncDrift = 0;
+      }
+
+   if (TR::Options::getVerboseOption(TR_VerboseJITaaS))
+      TR_VerboseLog::writeLineLocked(TR_Vlog_JITaaS, "Detaching JITaaSServer listening thread");
+
+   vm->internalVMFunctions->DetachCurrentThread((JavaVM *) vm);
+   statisticsThread->setStatisticsThread(NULL);
+   statisticsThread->getStatisticsThreadMonitor()->enter();
+   statisticsThread->getStatisticsThreadMonitor()->notifyAll();
```
mpirvu: Missing statisticsThread->setListenerThreadExitFlag();

runtime/compiler/runtime/StatisticsThread.cpp
```diff
@@ -0,0 +1,197 @@
...
+   uint64_t lastStatistics = 0;
+   uint64_t syncDrift = 0;
+   while(!statisticsThread->getStatisticsThreadExitFlag())
+      {
+      while(!statisticsThread->getStatisticsThreadExitFlag() && j9thread_sleep_interruptable((IDATA) samplingPeriod - syncDrift, 0) == 0)
+         {
+         persistentInfo->updateElapsedTime(samplingPeriod);
+         crtTime += samplingPeriod;
+
+         //adjusting the sampling period window for sync drift.
```
mpirvu: I do not understand how time correction can be done without reading the time from the OS.

runtime/compiler/control/HookedByTheJit.cpp
```diff
@@ -7104,6 +7112,12 @@ int32_t setUpHooks(J9JavaVM * javaVM, J9JITConfig * jitConfig, TR_FrontEnd * vm)
       if (TR::Options::getVerboseOption(TR_VerboseJITaaS))
          TR_VerboseLog::writeLineLocked(TR_Vlog_JITaaS, "Started JITaaSServer listener thread: %p ", listener->getListenerThread());
 
+      TR_StatisticsThread *statisticsThread = ((TR_JitPrivateConfig*)(jitConfig->privateConfig))->statisticsThread; 
+      statisticsThread->startStatisticsThread(javaVM);
+
+      if (TR::Options::getVerboseOption(TR_VerboseJITaaS))
```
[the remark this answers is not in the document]
satbir-ca: Starting of thread has nothing to do with the printing. We can start thread and not print if the statisticsFrequency is not given.
This is just for info like the message for Listener thread.

runtime/compiler/runtime/StatisticsThread.cpp
satbir-ca: we will reach here only when the statisticsThreadExitFlag is set from stopStatisticsThread().

runtime/compiler/runtime/StatisticsThread.cpp
```diff
@@ -0,0 +1,197 @@
...
+
+   J9VMThread *statThread = NULL;
+   PORT_ACCESS_FROM_JAVAVM(vm);
+
+   int rc = vm->internalVMFunctions->internalAttachCurrentThread(vm, &statThread, NULL,
+                                  J9_PRIVATE_FLAGS_DAEMON_THREAD | J9_PRIVATE_FLAGS_NO_OBJECT |
+                                  J9_PRIVATE_FLAGS_SYSTEM_THREAD | J9_PRIVATE_FLAGS_ATTACHED_THREAD,
+                                  statisticsThread->getStatisticsOSThread());
+   if (rc != JNI_OK)
+      return JNI_ERR; // attaching the JITaaS Statistics thread failed
```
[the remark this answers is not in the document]
satbir-ca: I thought when the attached will fail we will land in the error part. Will update it.

satbir-ca:
@mpirvu update the final code.

runtime/compiler/control/HookedByTheJit.cpp
```diff
@@ -7098,12 +7103,17 @@ int32_t setUpHooks(J9JavaVM * javaVM, J9JITConfig * jitConfig, TR_FrontEnd * vm)
...
+      int32_t statisticsFrequency = TR::Options::getStatisticsFrequency();
+      TR_Listener *listener = ((TR_JitPrivateConfig*)(jitConfig->privateConfig))->listener;
+
       listener->startListenerThread(javaVM);
 
       if (TR::Options::getVerboseOption(TR_VerboseJITaaS))
          TR_VerboseLog::writeLineLocked(TR_Vlog_JITaaS, "Started JITaaSServer listener thread: %p ", listener->getListenerThread());
 
+      TR_StatisticsThread *statisticsThread = ((TR_JitPrivateConfig*)(jitConfig->privateConfig))->statisticsThread;
+      statisticsThread->startStatisticsThread(javaVM, statisticsFrequency);
```
mpirvu: The statisticsThread could get the frequency by itself since that is a static global. No need to pass it as a parameter.

runtime/compiler/runtime/StatisticsThread.cpp
mpirvu: Correct. The listener does not have a stop routine.

runtime/compiler/runtime/StatisticsThread.cpp
```diff
@@ -0,0 +1,189 @@
...
+ * SPDX-License-Identifier: EPL-2.0 OR Apache-2.0 OR GPL-2.0 WITH Classpath-exception-2.0 OR LicenseRef-GPL-2.0 WITH Assembly-exception
+ *******************************************************************************/
+
+#include "runtime/StatisticsThread.hpp"
+#include "env/VMJ9.h"
+#include "runtime/CompileService.hpp"
+
+TR_StatisticsThread::TR_StatisticsThread()
+   : _statisticsThread(NULL), _statisticsThreadMonitor(NULL), _statisticsOSThread(NULL),
+   _statisticsThreadAttachAttempted(false), _statisticsThreadExitFlag(false), _statisticsFrequency(0)
```
mpirvu: You can initialize _statisticsFrequency in the constructor because `getStatisticsFrequency()` is a static. No need to wait until the thread is started.

runtime/compiler/runtime/StatisticsThread.cpp
mpirvu: I don't think this new code is going to work. Also, if the thread is reading the correct time every time it wakes up there is no need to have any drift value.
The original sampling thread wakes up every 10 ms, but reads the correct time every 100 ms, updating the elapsed time. The pseudocode is something like
```
crtTime = 0;
while(sleep_interruptable(samplingPeriod) == 0)
   crtTime += samplingPeriod
   if (100 ms or more have passed since last sync time)
      crtTime = j9time_current_time_millis() - startTime
      update time when last sync happened
```

satbir-ca: Sampler thread sleep for  max(_minSamplingPeriod, jitconfig->samplingFrequency) so if we put any value on command line greater then 10msec it will sleep for that time.
Or depend the state of the VM, sampling frequency in idle state is 1000 msec.

Both the implementation are essentially doing the same thing.

Sample thread is updating the elapsed time every iteration with sampling period which might not be the right current time and then after 100 msec update the elapsed time with the correct time i.e. sync.

Statistics thread is updating the elapsed time every time with the correct time.
More over it take into the account of the extra time taken by the thread to execute due to system latency and adjust the sleep window to accommodate this latency. If every time the thread wake up at right time then there will be no drift and not adjustment.

runtime/compiler/env/VMJ9.h
```diff
@@ -45,6 +45,7 @@
 #include "infra/Array.hpp"
 #include "env/CompilerEnv.hpp"
 #include "runtime/Listener.hpp"
+#include "runtime/StatisticsThread.hpp"
```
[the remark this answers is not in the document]
satbir-ca: Should we remove the Listener.hpp also.

runtime/compiler/runtime/StatisticsThread.cpp
```diff
@@ -0,0 +1,172 @@
...
+      {
+      while(!statisticsThread->getStatisticsThreadExitFlag() && j9thread_sleep_interruptable((IDATA) samplingPeriod, 0) == 0)
+         {
+         crtTime = j9time_current_time_millis();
+         persistentInfo->setElapsedTime(crtTime - persistentInfo->getStartTime());
+
+         if ((statisticsThread->getStatisticsFrequency()) && ((crtTime - lastStatistics) > statisticsThread->getStatisticsFrequency()))
+            {
+         //   TR_VerboseLog::writeLineLocked(TR_Vlog_JITaaS, "No of clients : 0x%x \n", ClientSessionHT::_clientSessionMap->size());
+            TR_VerboseLog::writeLineLocked(TR_Vlog_JITaaS, "Total thread : %d \n", compInfo->getNumUsableCompilationThreads());
```
mpirvu: Let's use plural here:  "Total compilation threads"

runtime/compiler/runtime/StatisticsThread.cpp
```diff
@@ -0,0 +1,172 @@
...
+            TR_VerboseLog::writeLineLocked(TR_Vlog_JITaaS, "Active Thread : %d \n",compInfo->getNumCompThreadsActive());
```
mpirvu: Same here: "Active compilation threads:

runtime/compiler/runtime/StatisticsThread.cpp
```diff
@@ -0,0 +1,172 @@
...
+            CpuUtilization *cpuUtil = compInfo->getCpuUtil();
+            if (cpuUtil->isFunctional())
+               {
+
+               TR_VerboseLog::writeLineLocked(TR_Vlog_JITaaS, "CpuLoad %d%% (AvgUsage %d%%) JvmCpu %d%% \n", cpuUtil->getCpuUsage(), cpuUtil->getAvgCpuUsage(), cpuUtil->getVmCpuUsage());
+               }
+            lastStatistics = crtTime;
+            }
+            oldSyncTime = crtTime;
```
mpirvu: `oldSyncTime` is not needed anymore because you read the correct time every single time

runtime/compiler/runtime/StatisticsThread.cpp
```diff
@@ -0,0 +1,172 @@
...
+         }
+      // This thread has been interrupted or StatisticsThreadExitFlag flag was set
+      }
+
+   if (TR::Options::getVerboseOption(TR_VerboseJITaaS))
+      TR_VerboseLog::writeLineLocked(TR_Vlog_JITaaS, "Detaching JITaaSServer listening thread");
```
mpirvu: The message is wrong here: change "listener" to "statistics"

runtime/compiler/runtime/StatisticsThread.cpp
```diff
@@ -0,0 +1,175 @@
...
+   persistentInfo->setStartTime(crtTime);
+   persistentInfo->setElapsedTime(0);
+   while(!statisticsThread->getStatisticsThreadExitFlag())
+      {
+      while(!statisticsThread->getStatisticsThreadExitFlag() && j9thread_sleep_interruptable((IDATA) samplingPeriod, 0) == 0)
+         {
+         crtTime = j9time_current_time_millis();
+         persistentInfo->setElapsedTime(crtTime - persistentInfo->getStartTime());
+
+         if ((statisticsThread->getStatisticsFrequency()) && ((crtTime - lastStatistics) > statisticsThread->getStatisticsFrequency()))
```
mpirvu: It occurred to me that the default value for `getStatisticsFrequency` is 0, so stats may be printed every single time. Am I missing something? We agreed that 0 means no printing at all.

Same for samplingFrequency. If that is 0 there should be no statistics thread created.

runtime/compiler/runtime/StatisticsThread.cpp
```diff
@@ -0,0 +1,175 @@
...
+            {
+            TR_VerboseLog::writeLineLocked(TR_Vlog_JITaaS, "Number of clients : %d \n", compInfo->getClientSessionHT()->getClientSessionMap().size());
```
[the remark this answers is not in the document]
satbir-ca: No, surprisingly compiler is not giving me warning for this. I have changes it to %u.

PR #3860, opened 2018-11-27
Getting OSR frame size from ROMMethod

mpirvu:
osrFrameSizeRomMethod is introduced to return the OSR frame size
for the ROMMethod.
This is needed to reduce the messages in the JITaaS.

Also removed the unused function "TR_J9MethodBase::osrFrameSize".


mpirvu:
Code already checked-in by another PR, hence closing.

PR #3862, opened 2018-11-27
Add cfdump support for value type bytecodes

acrowthe:
Adds cfdump support for the new bytecodes
    - 203, defaultvalue
    - 204, withvalue

cfdump assumes unkown bytecodes have a sizeof 1 byte. As these bytecodes are 3 bytes in size, cases need to be added for them in cfdump.

In addition, more information was added to cfdump in case of future unimplemented bytecodes. Previously, unimplemented bytecodes were printed as `JBunimplmented`. Now, the bytecode number is appended, as in `JBunimplmented_204`.


acrowthe:
@tajila While fixing the bytecode size issues related to cfdump, I encountered several other files where bytecode sizes had to be fixed (`runtime\util\sunbcrel.c`, `runtime\util\jbcmap.c`, `debugtools\DDR_VM\src\com\ibm\j9ddr\vm29\j9\PCStack.java`)

Would you like me to keep those fixes here, or move them to a separate PR?

acrowthe:
This is no longer reliant on #3857 as the changes were not necessary.

runtime/cfdumper/main.c
```diff
@@ -1437,12 +1437,20 @@ static void printDisassembledMethod(J9CfrClassFile* classfile, J9CfrMethod* meth
...
-				j9tty_printf( PORTLIB, "%s ", sunJavaBCNames[bc]);
+				j9tty_printf( PORTLIB, "%s", sunJavaBCNames[bc]);
 			}
 			else
 			{
-				j9tty_printf( PORTLIB, "%5i %s ", pc, sunJavaBCNames[bc]);
+				j9tty_printf( PORTLIB, "%5i %s", pc, sunJavaBCNames[bc]);
 			}
+
+			if(0 == strcmp(sunJavaBCNames[bc], "JBunimplemented")) {
```
tajila: This file does not follow the current coding standards, but we should be consistent with what is already there.

the `{` should be on the next line

runtime/cfdumper/main.c
```diff
@@ -1437,12 +1437,20 @@ static void printDisassembledMethod(J9CfrClassFile* classfile, J9CfrMethod* meth
...
+				j9tty_printf( PORTLIB, "_%d ", bc);
+			}
+			else {
```
tajila: same here

tajila:
The bytecode changes are small enough that they can remain here.

Please rename the title of the PR to something that mentions updating cfdump

runtime/cfdumper/main.c
```diff
@@ -1660,6 +1668,7 @@ static void printDisassembledMethod(J9CfrClassFile* classfile, J9CfrMethod* meth
 				case CFR_BC_putstatic:
 				case CFR_BC_getfield:
 				case CFR_BC_putfield:
+				case CFR_BC_withfield:
```
tajila: these changes should be ifdef'd. If the JVM is not VT enabled, the withfield, defaultValue bytecodes should be treated as unimplemented bytecodes

tajila:

runtime/cfdumper/main.c
```diff
@@ -1437,12 +1437,22 @@ static void printDisassembledMethod(J9CfrClassFile* classfile, J9CfrMethod* meth
 			NEXT_U8_ENDIAN(bigEndian, bc, bcIndex);
 			if(wide)
 			{
-				j9tty_printf( PORTLIB, "%s ", sunJavaBCNames[bc]);
+				j9tty_printf( PORTLIB, "%s", sunJavaBCNames[bc]);
```
gacholio: Why the spacing change (in the string)?

runtime/cfdumper/main.c
```diff
@@ -1437,12 +1437,22 @@ static void printDisassembledMethod(J9CfrClassFile* classfile, J9CfrMethod* meth
...
 			}
 			else
 			{
-				j9tty_printf( PORTLIB, "%5i %s ", pc, sunJavaBCNames[bc]);
+				j9tty_printf( PORTLIB, "%5i %s", pc, sunJavaBCNames[bc]);
```
gacholio: Same here.

runtime/cfdumper/main.c
acrowthe: The spacing change is so that the bytecode can be appended to the bytecode name, in the case of JBunimplemented. This allows for easier debugging.

Example of new unimplemented output (prior to handling `204` as `JBwithfield`):
```
    2 JBunimplmented_204 26 FlattenedLine2D.en QPoint2D;
```

Space was removed to avoid confusion between the bytecode (204) and other data (26)

runtime/cfdumper/main.c
acrowthe: Same reason as above: For including bytecode number in output

gacholio:
Can you explain the action value changes?

acrowthe:
@gacholio I'm unsure which change you are referring to by that. Could you clarify for me please?

gacholio:
The byte value changes in the tables.

acrowthe:
These tables are all used to look up a size or offset based on a given bytecode.

Specifically, `jbcmap.c` and `PCStack.java` are used to map a given bytecode to it's size in bytes. As `defaultvalue` is a 3 byte bytecode, these needed to be changed so that iterators would know the bytes 2 and 3 of `defaultvalue` are not different bytecodes.

I am unsure what `sunbcrel.c` is used for: There I mirrored the changes to withfield as their signatures are the same. @tajila may know more about that table.

gacholio:
Quick scan of the code says only the low 3 bits of the bytes in sunbcrel.c are used at all, so I'll approve this.

gacholio:
jenkins test all xlinux jdk8

gacholio:
Does the ROM class dumper in DDR also need to be updated? Personally, I use that at least as often as cfdump.

acrowthe:
Out of scope for this PR, but I can certainly give that a look, if @tajila agrees. There are still several other bytecode issues I am working on right now though.

gacholio:
I'll accept any changes in a future PR, but it really isn't "out of scope". DDR has to be kept in line with the main codebase.

tajila:
> Does the ROM class dumper in DDR also need to be updated? Personally, I use that at least as often as cfdump.

`!dumpromclass` is working for the new bytecodes with Q and L signatures.

gacholio:
OK, cool, means `!bytecodes` will work as well (if they don't back-end onto the same code, I'll be very upset).

tajila:
Yes `!bytecodes` works as well

gacholio:
pltest failure in sanity obviously unrelated to this PR.

gacholio:
Test failure in extended also appears to be nonsense.

PR #3866, opened 2018-11-28
Remove Xgc:debugLOAResize option

LinHu2016:
- remove -Xgc:debugLOAResize option, use trace point for the related debug output


dmitripivkine:

charliegracie:
The commit title and the contents of the change seem different.

The commit could simple state that it is removing the option and then the commit body would just need the signed-off-by.

PR #3867, opened 2018-11-28
Fix rare VM shutdown assertion

gacholio:
If VM startup fails to complete, but gets far enough to have reported VM
start, then VM shutdown must be reported. The hook caller for this case
was calling the hook in the JNI context, whereas the other two callers
have entered the VM and released VM access. Make the startup failure
case consistent with the other callers.

[ci skip]


gacholio:
jenkins test sanity aix jdk8

Issue #3723, opened 2018-11-17
Reduce number of IProfiler related messages


mpirvu:
Currently the JITaaS server caches interpreter profiling information for methods that have already been compiled. This can be done safely because compiled methods cannot accumulate more interpreter profiling information. However, even with this level of caching messages related to Iprofiler are by far the most frequent ones.
The proposal advanced by this issue is to allow caching of Iprofiler information for interpreted methods only for the duration of the compilation that requested that info.
More specifically, when a compilation wants to know the Iprofiler info for a particular bytecode, the JITaaS server will send a message to the client which will respond with Iprofiler info for the entire method (the idea is that the server will ask for info on other bytecodes of this method soon). The server will cache this whole-method-iprofiler info using heap memory which is freed at the end of the compilation. Even though more IProfiler information can be accumulated while the compilation is in progress we surmise that the change is small enough such that the decisions taken by the compiler will not change.

Experimental results need to confirm that
1) Throughput is not degraded
2) Number of messages is reduced enough to cause a drop in CPU consumed by the client

PR #3870, opened 2018-11-28
Cache IProfiler info for interpreted methods

dmitry-ten:
Previously, IProfiler would only cache profiling information
for methods that have already been compiled, and for the method currently being compiled.
This change starts caching info for interpreted methods in a per-compilation cache.
When a server requests info for a particular bytecode, client sends profiling info
for the entire method, which is then cached until the end of the compilation.
Even though IProfiler data might get updated during compilation,
the hope is that the change will be small enough to not affect performance.

Addresses:#3723

### DIFF
--- a/runtime/compiler/control/CompilationThread.cpp
+++ b/runtime/compiler/control/CompilationThread.cpp
@@ -9298,6 +9298,9 @@ TR::CompilationInfo::compilationEnd(J9VMThread * vmThread, TR::IlGeneratorMethod
    TR_DataCache *dataCache = NULL;
    const J9JITDataCacheHeader *storedCompiledMethod = nullptr;
    TR::CompilationInfo *compInfo = TR::CompilationInfo::get();
+   if (comp->getPersistentInfo()->getJITaaSMode() == SERVER_MODE)
+      // end of compilation, clear per-compilation IProfiler cache
+      ((TR::CompilationInfoPerThreadRemote *) entry->_compInfoPT)->clearIProfilerMap(comp->trMemory());
 
    if (details.isNewInstanceThunk())
       {
--- a/runtime/compiler/control/JITaaSCompilationThread.cpp
+++ b/runtime/compiler/control/JITaaSCompilationThread.cpp
@@ -112,19 +112,21 @@ void handler_IProfiler_profilingSample(JITaaS::J9ClientStream *client, TR_J9VM *
 
    bool isCompiled = TR::CompilationInfo::isCompiled((J9Method*)method);
    bool isInProgress = comp->getMethodBeingCompiled()->getPersistentIdentifier() == method;
-   bool wholeMethodInfo = (isCompiled || isInProgress) && (data == 0);
    bool abort = false;
+   // used to tell the server if a profiled entry should be stored in persistent or heap memory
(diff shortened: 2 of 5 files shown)


dmitry-ten:
For AcmeAir,
`IProfiler_profilingSample` message count went from 216k to 82k.
My testing didn't show any throughput regression, but additional testing needs to be done.

mpirvu:
I'll give it a spin when I have the time. Meanwhile please go through the sanity buckets making sure there are no functional problems.

dmitry-ten:
`_IPDataPerComp` wasn't being freed after unsuccessful compilation, fix by moving call to `clearIProfilerMap` to `TR::CompilationInfo::compilationEnd`, which should get called at the end of every compilation.

dmitry-ten:
The reason @mpirvu didn't observe any CPU improvement is because I wasn't caching for methods where IProfiler entries for all bytecodes are NULL. 
This means that whenever server would request profiling information for a method with no info, client would step through all of the bytecodes, find out that there is no profiling information, tell that to the server, only for the server to not cache it, and ask the client again next time.
I changed the code to cache methods with no info, and the number of records client has to go through went down to a value close to baseline JITaaS.
The number of `profilingSample` messages decreased even further, to 19k in AcmeAir.

I also enabled proper handling of different caches in `setCallCount`, which wasn't working correctly with heap caches before.

dmitry-ten:
@mpirvu this is ready for another review.

mpirvu:
Code looks good. Performance is as expected this time. I measured a 10% improvement in CPU consumed at the client bringing the JITaaS advantage from 50% to 55%. Throughput and footprint are unchanged.
```

Results for JDK=/home/mpirvu/JITaaS/sdk jvmOpts=-XX:JITaaSClient -XXjitdirectory=/home/mpirvu/JITaaS -Xjit:dontDowngradeToCold,sampleThreshold=0,scorchingSampleThreshold=0 -Xmx256m -Xshareclasses:none
Throughput      avg=13553.42    min=13415.10    max=13721.30    stdDev=92.7     maxVar=2.28%    confInt=0.40%   samples=10
CompTime        avg=11415.80    min=11272.00    max=11818.00    stdDev=164.4    maxVar=4.84%    confInt=0.83%   samples=10
Footprint       avg=230321.60   min=227808.00   max=232132.00   stdDev=1604.4   maxVar=1.90%    confInt=0.40%   samples=10

Baseline
Results for JDK=/home/mpirvu/JITaaS/sdk jvmOpts=-XXjitdirectory=/home/mpirvu/JITaaS -Xjit:sampleThreshold=0,scorchingSampleThreshold=0,dontDowngradeToCold -Xmx256m -Xshareclasses:none
Throughput      avg=14458.71    min=14032.60    max=14628.00    stdDev=165.3    maxVar=4.24%    confInt=0.66%   samples=10
CompTime        avg=25716.00    min=24742.00    max=26595.00    stdDev=609.5    maxVar=7.49%    confInt=1.37%   samples=10
Footprint       avg=220413.60   min=218984.00   max=221276.00   stdDev=735.9    maxVar=1.05%    confInt=0.19%   samples=10

dmitry-ten:
Resolved conflict, I think this is ready to be merged, @mpirvu

Issue #3768, opened 2018-11-21
Recent Java12 changes have moved some classes causing builds to now fail


groeges:
The recent merge of JDK12 code to the extensions repository has caused the JDK12 builds to fail.
This seems to be related to Oracle moving some files from `src/java.base/share/classes/jdk/internal/misc` to `src/java.base/share/classes/jdk/internal/access`. 

The files that seem to have caused the OpenJ9 jcl source build to fail are: 
```
JavaLangRefAccess.java
SharedSecrets.java
```
as the package locations have changed and hence the import statements in the OpenJ9 jcl code that references these classes is incorrect and causes issues:
```
/home/groeges/source/openj9jdk12/openj9-openjdk-jdk/build/linux-x86_64-server-release/support/j9jcl_sources/java.base/share/classes/java/lang/invoke/MethodHandle.java:41: error: cannot find symbol
import jdk.internal.misc.SharedSecrets;
                        ^
  symbol:   class SharedSecrets
  location: package jdk.internal.misc
/home/groeges/source/openj9jdk12/openj9-openjdk-jdk/build/linux-x86_64-server-release/support/j9jcl_sources/java.base/share/classes/java/lang/System.java:33: error: cannot find symbol
import jdk.internal.misc.SharedSecrets;
                        ^
  symbol:   class SharedSecrets
  location: package jdk.internal.misc
/home/groeges/source/openj9jdk12/openj9-openjdk-jdk/build/linux-x86_64-server-release/support/j9jcl_sources/java.base/share/classes/java/lang/Access.java:49: error: cannot find symbol
import jdk.internal.misc.JavaLangAccess;
```

groeges:
These need to be fixed in order for the builds to work correctly.

groeges:
The following files needed to be updated to resolve the issues with the missing imports:
```
	modified:   jcl/src/java.base/share/classes/java/lang/Access.java
	modified:   jcl/src/java.base/share/classes/java/lang/ClassLoader.java
	modified:   jcl/src/java.base/share/classes/java/lang/System.java
	modified:   jcl/src/java.base/share/classes/java/lang/invoke/MethodHandle.java
	modified:   jcl/src/java.base/share/classes/java/lang/invoke/MethodHandles.java
	modified:   jcl/src/java.base/share/classes/java/lang/ref/Reference.java
	modified:   jcl/src/java.management/share/classes/com/ibm/java/lang/management/internal/BufferPoolMXBeanImpl.java
```

[...]

PR #3875, opened 2018-11-29
Remaining stub methods for Java 12 builds

theresa-m:
Issue: #3768

+ vm compiles
- vm cannot start without implementation of Unsafe natives
- some jmods cannot compile without Unsafe and MethodHandles.filterArgumentsWithCombiner stubs implemented


theresa-m:
Unsafe reference methods are set replace Unsafe Object methods so setting the reference natives to use Object methods is the correct behavior.

PR #3876, opened 2018-11-29
Fix boolean return in SCC API

tajila:
Fix boolean return in SCC API


pshipton:

DanHeidinga:
Related to https://github.com/eclipse/openj9/pull/3678

DanHeidinga:
@tajila Can you also search for other JNI natives in the OpenJ9 that return jboolean and ensure they return clean booleans?

tajila:
I checked `/runtime/jcl` and `/runtime/j9vm` and the boolean returns look correct

PR #3877, opened 2018-11-29
Expand MethodHandle invoke calls after walker

liqunl:
One or more extra calls are generated for MethodHandle invoke in walker,
adding the difficulty to do OSR bookkeeping. This commit will move the
extra call creations to post-walker.

### DIFF
--- a/runtime/compiler/ilgen/IlGenerator.cpp
+++ b/runtime/compiler/ilgen/IlGenerator.cpp
@@ -39,6 +39,7 @@
 #include "optimizer/BoolArrayStoreTransformer.hpp"
 #include "ras/DebugCounter.hpp"
 #include "optimizer/TransformUtil.hpp"
+#include "env/JSR292Methods.h"
 
 #define OPT_DETAILS "O^O ILGEN: "
 
@@ -62,7 +63,12 @@ TR_J9ByteCodeIlGenerator::TR_J9ByteCodeIlGenerator(
      _invokeSpecialInterfaceCalls(NULL),
      _invokeSpecialSeen(false),
      _couldOSRAtNextBC(false),
-     _processedOSRNodes(NULL)
+     _processedOSRNodes(NULL),
+     _invokeHandleCalls(NULL),
+     _invokeHandleGenericCalls(NULL),
+     _invokeDynamicCalls(NULL),
+     _ilGenMacroInvokeExactCalls(NULL),
+     _methodHandleInvokeCalls(NULL)
    {
    static const char *noLookahead = feGetEnv("TR_noLookahead");
    _noLookahead = (noLookahead || comp->getOption(TR_DisableLookahead)) ? true : false;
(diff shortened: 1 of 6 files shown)


liqunl:
@andrewcraik @jdmpapin Could you take a look?

runtime/compiler/ilgen/IlGenerator.cpp
```diff
@@ -42,6 +42,20 @@
 
 #define OPT_DETAILS "O^O ILGEN: "
 
+#define JSR292_MethodHandle   "java/lang/invoke/MethodHandle"
```
andrewcraik: These defines are duplicating defines in j9method.cpp I think we should have a header for the JSR292 defines and include the header in the files where we want the defines to apply so that the definitions are just in once place.

andrewcraik:
Overall I think this looks good, but I would like to see more trace (my usual refrain :) ) just to make it more clear what is happening. @cathyzhyi could you do a review?

andrewcraik:
@liqunl do the NodeChecklists need to be in heap memory? Is there an ilgen region or a stack region they could be allocated from since I don't think we need them once the walker is done and it would save space for them to be cleaned up.

runtime/compiler/ilgen/Walker.cpp
```diff
@@ -4132,6 +4108,33 @@ TR_J9ByteCodeIlGenerator::genInvokeHandle(TR::SymbolReference *invokeExactSymRef
...
+TR::Node *
+TR_J9ByteCodeIlGenerator::genILGenMacroInvokeExact(TR::SymbolReference *invokeExactSymRef)
+   {
+   if (comp()->compileRelocatableCode())
+      {
+      comp()->failCompilation<J9::AOTHasInvokeHandle>("COMPILATION_AOT_HAS_INVOKEHANDLE 1");
+      }
+
+   if (comp()->getOption(TR_FullSpeedDebug) && !isPeekingMethod())
+      comp()->failCompilation<J9::FSDHasInvokeHandle>("FSD_HAS_INVOKEHANDLE 1");
```
jdmpapin: I think these could be assertions rather than conditional compilation failures. AOT or FSD compilations should have failed before we start trying to `runMacro()`/`runFEMacro()`

runtime/compiler/ilgen/J9ByteCodeIlGenerator.hpp
```diff
@@ -379,6 +389,11 @@ class TR_J9ByteCodeIlGenerator : public TR_IlGenerator, public TR_J9ByteCodeIter
    // JSR292
    int32_t                           _argPlaceholderSlot;
    int32_t                           _argPlaceholderSignatureOffset;
+   TR::NodeChecklist                *_methodHandleInvokeNodes;
+   TR::NodeChecklist                *_invokeHandleNodes;
+   TR::NodeChecklist                *_invokeHandleGenericNodes;
+   TR::NodeChecklist                *_invokeDynamicNodes;
+   TR::NodeChecklist                *_ilGenMacroInvokeExactNodes;
```
jdmpapin: Using `NodeChecklist` for this will break if a node of interest is duplicated, e.g. because of a catch block catching multiple exception types. This happened before with `expandInvokeSpecialInterface()`, which is why `_invokeSpecialInterfaceCalls` identifies the calls by their bytecode indices.

liqunl:
@andrewcraik @jdmpapin Changes are made according to your comments, could you review again?

liqunl:
WIP this PR as it seems to have exposed a bug.

liqunl:
@andrewcraik @cathyzhyi is developing a fix to the bug exposed by this PR and #3744, we hope we can get a fix for the bug on Tuesday, could you continue reviewing this PR?

andrewcraik:
I think this looks good - it actually makes the 292 expansion easier to visualize. Has there been validation that perf and compile-time are not adversely affected by the change as part of the functional verification. Once we have evidence to say this is performant we should be good to go.

andrewcraik:
Jenkins test sanity xlinux,win,plinux jdk8.jdk11

andrewcraik:
enkins test sanity xlinux,win,plinux jdk8,jdk11

andrewcraik:

PR #3880, opened 2018-11-29
Update JITaaS build documentations

harryyu1994:
Add in "How to prepare your system for JITaaS". Previous doc was missing protobuf and other dependency installation procedures.
[skip ci]


harryyu1994:
@mpirvu Hi Marius, could you help review this?

mpirvu:
BTW, the commit message does not have to be in bullet form because logically there is just one bullet. The second one is the reason why we implemented the first.

PR #3883, opened 2018-11-29
Enforce final field setting rules

gacholio:
If the interpreter successfully resolves a putfield or putstatic of a
final field in one method, another method in the same class will see the
ref as resolved and not perform the final field check, which allows
final fields to be set from non-constructor methods.

Revise the resolved check such that final field puts are always
considered unresolved. Update putfield and putstatic to perform the
method check every time in the final field case.

[ci skip]


runtime/oti/VMHelpers.hpp
```diff
@@ -1450,6 +1458,41 @@ class VM_VMHelpers
...
+	/**
+	 * Checks whether a ROM method is <clinit> or <init>
+	 *
+	 * @param romMethod[in] the J9ROMMethod to test
+	 * @param isStatic[in] true to check for <clinit>, false to check for <init>
+	 *
+	 * @returns true if the method is a constructor, false if not
+	 */
+	static VMINLINE bool
+	romMethodIsConstructor(J9ROMMethod *romMethod, bool isStatic)
```
DanHeidinga: Constructor doesn't seem like the right name here - Initializer maybe?

runtime/vm/BytecodeInterpreter.hpp
```diff
@@ -6263,26 +6262,37 @@ done:;
...
+			/* The above test will fail if the ref has been resolved for put and the target
+			 * field is final. In this case, ensure that the current method is a constructor
+			 * if the class is enforcing the final field setting rules.
+			 *
+			 * If the check fails, run the resolve again to throw the exception.
+			 */
+			if (!J9_ARE_ALL_BITS_SET(flagsAndClass, (UDATA)(J9StaticFieldRefPutResolved | J9StaticFieldRefFinal) << (8 * sizeof(UDATA) - J9_REQUIRED_CLASS_SHIFT))
+				|| (VM_VMHelpers::romClassChecksFinalStores(ramConstantPool->ramClass->romClass)
+					&& !VM_VMHelpers::romMethodIsConstructor(J9_ROM_METHOD_FROM_RAM_METHOD(_literals), false)
+				   )
```
DanHeidinga: Not a fan of this formatting but also not a fan of any alternative I've been able to come up with

runtime/oti/VMHelpers.hpp
gacholio: Updated.

runtime/vm/BytecodeInterpreter.hpp
gacholio: Agreed - it looks better in Eclipse.

DanHeidinga:
@gacholio To confirm, @tajila was writing tests that covered this?

DanHeidinga:

gacholio:
Opened #3886 to cover the new tests.

PR #3887, opened 2018-11-29
Fix crash in verifier

gacholio:
There were a few places in the jsr inliner which fail to set the PC
value when setting an error code (leaving the PC at its default value of
-1). After various casts, the end result is that the error reporting 
framework gets a PC value of 0xFFFF, which is used to index into the
bytecode array, resulting in a possible crash on unreadable memory.

[ci skip]


DanHeidinga:

PR #3888, opened 2018-11-29
Add Infrastructure to retry AOT loads

dsouzai:
Currently, when an AOT load fails, the compiler will perform a JIT
compilation. An AOT compilation is generally performed with count=1000,
bcount=250, but an AOT load is generally performed with scount=20. This
means that there will likely be AOT load failures caused by validation
failures which occur because what's being validated isn't resolved yet.
This PR provides the infrastructure to postpone the AOT load (that
fail under specific conditions likely caused by the low scount) and
retry the load some time later.

The general design is as follows. The first time an AOT load fails,
the J9Method is added to a map. The key is the J9Method pointer itself,
and the value is the new invocation count. The extra field of the
J9Method is then updated with the new invocation count, and the
compilation ends. On each subsequent load failure, the invocation count
doubles, up to half of the initial default count/bcount for that
method; the value in the map is used for this purpose and updated
accordingly. The purpose of this is to ensure that the total invocation
count set for this method does not exceed its default count/bcount. If
the count cannot be increased anymore, the method is JIT compiled.

In order to minimize the cost of querying the map (since the map is
only required for AOT load failures), a method is only removed
from the map when the compiler is unable to increase the invocation
count. This is ok because the map is only checked during an AOT load
failure.

### DIFF
--- a/runtime/compiler/control/CompilationRuntime.hpp
+++ b/runtime/compiler/control/CompilationRuntime.hpp
@@ -880,6 +880,9 @@ public:
    bool getSuspendThreadDueToLowPhysicalMemory() const { return _suspendThreadDueToLowPhysicalMemory; }
    void setSuspendThreadDueToLowPhysicalMemory(bool b) { _suspendThreadDueToLowPhysicalMemory = b; }
 
+   bool shouldRetryAOTLoad(void *j9method);
+   bool retryAOTLoadIfPossible(void *j9method, Compilation *comp);
+
    static int32_t         VERY_SMALL_QUEUE;
    static int32_t         SMALL_QUEUE;
    static int32_t         MEDIUM_LARGE_QUEUE;
@@ -1006,6 +1009,12 @@ private:
    int32_t                _numSeriousFailures; // count failures where method needs to continue interpreted
 
    TR::Monitor           *_gpuInitMonitor;
+
+
+   typedef TR::typed_allocator<std::pair<void* const, uint32_t>, TR_TypedPersistentAllocatorBase&> AOTLoadRetryAllocator;
+   typedef std::less<void*> AOTLoadRetryComparator;
+   typedef std::map<void*, uint32_t, AOTLoadRetryComparator, AOTLoadRetryAllocator> AOTLoadRetryMap;
+   AOTLoadRetryMap        _aotLoadRetryMap;
 
    enum // flags
       {
--- a/runtime/compiler/control/CompilationThread.cpp
+++ b/runtime/compiler/control/CompilationThread.cpp
@@ -784,7 +784,8 @@ TR::CompilationInfoPerThread::CompilationInfoPerThread(TR::CompilationInfo &comp
 TR::CompilationInfo::CompilationInfo(J9JITConfig *jitConfig) :
    _persistentMemory(pointer_cast<TR_PersistentMemory *>(jitConfig->scratchSegment)),
    _reloRuntime(jitConfig),
-   _samplingThreadWaitTimeInDeepIdleToNotifyVM(-1)
+   _samplingThreadWaitTimeInDeepIdleToNotifyVM(-1),
+   _aotLoadRetryMap((AOTLoadRetryComparator()), TR_TypedPersistentAllocatorBase())
    {
    // The object is zero-initialized before this method is called
    //
@@ -1768,6 +1769,8 @@ bool TR::CompilationInfo::shouldRetryCompilation(TR_MethodToBeCompiled *entry, T
    TR::IlGeneratorMethodDetails & details = entry->getMethodDetails();
    J9Method *method = details.getMethod();
 
+   TR::CompilationInfo *compInfo = TR::CompilationInfo::get();
+
    // when the compilation fails we might retry
    //
    if (entry->_compErrCode != compilationOK &&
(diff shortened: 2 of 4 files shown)


dsouzai:
@mpirvu could you please review.

dsouzai:
> The places where the new hashtable is accessed are protected by the compilation monitor, so the new monitor is not needed.

Ooh good catch, that'll simply the commit.

> When setting the invocation count we may want to double check that the method is actually interpreted. 

Right, that's fair; I'll update the PR with the changes.

dsouzai:
@mpirvu I made the requested changes.

mpirvu:
The next step would be to measure the impact of these changes. Only then we can decide if we want to deliver it.

dsouzai:
I ran liberty startup:

```
Results for JDK=/home/dsouzai/sdks/openj9BuildForRetry jvmOpts=-Xmx256m -Xshareclasses:enableBCI,name=liberty -Xscmx60M -Xscmaxaot8m -Xjit:enableAOTStats -Xaot:enableAOTStats -XXjitdirectory=/home/dsouzai/jitdlls/openj9BuildForRetry/orig
StartupTime     avg=5777        min=5504        max=6299        stdDev=160.4    maxVar=14.5%    confInt=0.58%   samples= 64
Footprint       avg=243089      min=232720      max=253688      stdDev=5352.0   maxVar=9.0%     confInt=0.46%   samples= 64
CThreadTime     avg=2673        min=2277        max=3227        stdDev=190.8    maxVar=41.7%    confInt=1.49%   samples= 64

Results for JDK=/home/dsouzai/sdks/openj9BuildForRetry jvmOpts=-Xmx256m -Xshareclasses:enableBCI,name=liberty -Xscmx60M -Xscmaxaot8m -Xjit:enableAOTStats -Xaot:enableAOTStats -XXjitdirectory=/home/dsouzai/jitdlls/openj9BuildForRetry/changes
StartupTime     avg=5773        min=5511        max=6193        stdDev=161.4    maxVar=12.4%    confInt=0.58%   samples= 64
Footprint       avg=242411      min=231952      max=253084      stdDev=5096.3   maxVar=9.1%     confInt=0.44%   samples= 64
CThreadTime     avg=2644        min=2382        max=3212        stdDev=205.8    maxVar=34.8%    confInt=1.62%   samples= 64
```

These changes have no impact on startup.

dsouzai:
Throughput results:

```
Results for JDK=/home/dsouzai/sdks/openj9BuildForRetry jvmOpts=-Xshareclasses:name=liberty -Xscmx60M -Xscmaxaot8M -Xmx1G -XXjitdirectory=/home/dsouzai/jitdlls/openj9BuildForRetry/orig
Throughput      avg=8678.05     min=8592.30     max=8771.10     stdDev=73.7     maxVar=2.08%    confInt=1.00%   samples= 4
Intermediate results:
Run 0   775.1   8328.5  8726.6  8685.4  Avg=8685        CPU=89261 ms  Footprint=917308 KB
Run 1   789.6   8208.3  8594.3  8663.4  Avg=8663        CPU=84426 ms  Footprint=901240 KB
Run 2   762.3   8336.5  8696.5  8771.1  Avg=8771        CPU=86892 ms  Footprint=913728 KB
Run 3   761.3   8017.4  8601.0  8592.3  Avg=8592        CPU=88070 ms  Footprint=893844 KB
CompTime        avg=87162.25    min=84426.00    max=89261.00    stdDev=2064.7   maxVar=5.73%    confInt=2.79%   samples= 4
Footprint       avg=906530.00   min=893844.00   max=917308.00   stdDev=10907.1  maxVar=2.63%    confInt=1.42%   samples= 4

Results for JDK=/home/dsouzai/sdks/openj9BuildForRetry jvmOpts=-Xshareclasses:name=liberty -Xscmx60M -Xscmaxaot8M -Xmx1G -XXjitdirectory=/home/dsouzai/jitdlls/openj9BuildForRetry/changes
Throughput      avg=8672.33     min=8655.10     max=8695.50     stdDev=17.1     maxVar=0.47%    confInt=0.23%   samples= 4
Intermediate results:
Run 0   796.0   8231.2  8632.6  8655.1  Avg=8655        CPU=85323 ms  Footprint=925188 KB
Run 1   795.5   8278.7  8633.5  8665.8  Avg=8666        CPU=89432 ms  Footprint=903620 KB
Run 2   781.1   8251.1  8599.4  8695.5  Avg=8696        CPU=91262 ms  Footprint=913692 KB
Run 3   776.0   8270.4  8544.5  8672.9  Avg=8673        CPU=85293 ms  Footprint=914932 KB
CompTime        avg=87827.50    min=85293.00    max=91262.00    stdDev=3003.7   maxVar=7.00%    confInt=4.02%   samples= 4
Footprint       avg=914358.00   min=903620.00   max=925188.00   stdDev=8819.8   maxVar=2.39%    confInt=1.13%   samples= 4
```

No impact on throughput either.

dsouzai:
@mpirvu The changes don't have any impact on performance. FWIW, this feature isn't enabled by default.

mpirvu:
The goal of this feature was to improve start-up time by reducing the number of AOT load failures. Did the number of AOT load failures decrease?

dsouzai:
Using this to improve start-up time is something that's going to require tuning to figure out the right retry invocation count. However, it does reduce the number of AOT load failures, by which I mean, it reduces the number of AOT loads that had to be JIT compiled:

In the orig build:
Cold run: 2 failures
Warm run1: 115 failures
Warm run8: 22 failures

Using `retryAOTLoadInitialInvocationCount=50`:
Cold run: 8 failures
Warm run1: 28 failures
Warm run2: 15 failures

Because this PR results in a vlog line outputted (https://github.com/eclipse/openj9/pull/3888/files#diff-08779930c70bee9d159bc2c125154c30R11662), I got the number of failures by modifying your vlog summary script to decrement the number of failures by one when it encounters this line.

mpirvu:
The existing heuristic uses SCC hints to find the 'right' invocation count for AOT loads. The problem with this approach is that needs a few 'training runs'. The value of the new scheme could be better shown if we do small batches: 1 cold run and 1 warm run after which SCC is destroyed, thus the command line for the script should be something like:
perl start_stop_liberty 2 50

dsouzai:
Archiving.

PR #3896, opened 2018-11-30
Fix unsafe and JNI bool writes and returns

tajila:
Fix unsafe and JNI bool writes and returns

The spec rules on bool's `the int value is narrowed by taking the
bitwise AND of value and 1, resulting in value` only apply to JVM
bytecodes. The Unsafe behaviour in the RI writes `(0 != value)`. 
JNI returns behave the same way.


DanHeidinga:
We need to validate the Unsafe case as it's backing VarHandles and it would be strange for Varhandles and the bytecodes to have different behaviour.

tajila:
@DanHeidinga changes are ready for review

pshipton:
jenkins test sanity all jdk8

DanHeidinga:
Thanks @tajila.  It's slightly concerning that writing a boolean with varhandles vs bytecode & jni can have different results but I agree that we should match the RI here.

PR #3899, opened 2018-11-30
Update JITaaS testing docs

dmitry-ten:
Add an explanation for how to get ant-contrib.jar,
as it's something people have struggled with previously

### DIFF
--- a/doc/compiler/jitaas/Testing.md
+++ b/doc/compiler/jitaas/Testing.md
@@ -31,32 +31,45 @@ These are the steps to run the tests on your machine.
 
 1. Compile a debug build of openj9 from scratch. To do this pass the flag `--with-debug-level=slowdebug` to `configure`.
 
-2. Compile  the tests (only need to compile once):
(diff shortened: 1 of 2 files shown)


dmitry-ten:
@mpirvu I remember that you and I both struggled with getting this jar file.

smlambert:
Cool, I did not know you created this.  :)

A few notes:

- When you merge jitaas into master, it will be good to reference this from the https://github.com/eclipse/openj9/blob/master/test/docs/OpenJ9TestUserGuide.md

- In your section 2, you could just add:
export EXTRA_OPTIONS="-XX:JITaaSClient"
to avoid having to pass it on each test target execution in the Run tests section

- ant-contrib is one of the test-prereqs, feel free to reference https://github.com/eclipse/openj9/blob/master/test/docs/Prerequisites.md or point to https://github.com/eclipse/openj9/blob/master/buildenv/docker/test/Dockerfile#L57-L68 for install example

dmitry-ten:
Applied suggestions. @mpirvu should be ready for merge.

PR #3901, opened 2018-12-01
Fix for JITaaS sequencer code

mpirvu:
Previous sequencer code had a problem described here:
1. One thread times-out and clears the caches from the server
2. Another waiting thread times-out very soon after and is now
waiting to enter the critical section
3. The first thread proceeds through the critical section and
at some point it unlocks the monitor to do a remote read.
4. The second thread enters the critical section and thinks it
needs to clear the caches as well because it is a thread that
timed-out.

The solution is for the first thread to flag the next in line
thread that everything is ok and it should go and wait again
on the monitor until it gets notified.


ymanton:
Is the problem here that a thread N that has already cleared the caches will release the sequence monitor then proceed to apply it's updates to the caches while thread N+1 that also timed out is clearing them at the same time?

PR #3913, opened 2018-12-03
Fix typo: USE_COMPUTED_GOTO for gcc 7 on zLinux

keithc-ca:
I expect the intent of #3761 was to enable computed goto for gcc 7 on zLinux (not require gcc 8+).


pshipton:
jenkins test all zlinux jdk11

PR #3915, opened 2018-12-03
Modify tar actions to omit owner information

jdekonin:
* [skip ci]


jdekonin:
My test got lucky I think.  The `-o` should be on the extraction not the creation of the file.

PR #3919, opened 2018-12-04
Return the same helper index on method symbol

cathyzhyi:
Fix getInterpretedDispatchHelper on PPC to return the same
helper index on method symbol reference since now
induceOSR helper can be either TR_induceOSRAtCurrentPC or
TR_induceOSRAtCurrentPCAndRecompile.


andrewcraik:
@gita-omr this is a blocker fix so can you have a look - it seems right to me.

andrewcraik:

andrewcraik:
Checks all clean and appropriate reviews completed - merging

PR #3920, opened 2018-12-04
Fix instance/static field resolution confusion

gacholio:
The change in #3883 introduced some confusion in the putstatic bytecode
where it might consider a field resolved when it should not.

Fix the resolve/final logic in putstatic and putfield.

[ci skip]

### DIFF
--- a/runtime/codert_vm/cnathelp.cpp
+++ b/runtime/codert_vm/cnathelp.cpp
@@ -1390,7 +1390,7 @@ old_slow_jitResolveClassFromStaticField(J9VMThread *currentThread)
 	J9RAMStaticFieldRef *ramStaticFieldRef = (J9RAMStaticFieldRef*)ramConstantPool + cpIndex;
 	IDATA flagsAndClass = ramStaticFieldRef->flagsAndClass;
 	UDATA valueOffset = ramStaticFieldRef->valueOffset;
-	if (J9_UNEXPECTED(!VM_VMHelpers::staticFieldRefIsResolved(flagsAndClass, valueOffset, false))) {
+	if (J9_UNEXPECTED(!VM_VMHelpers::staticFieldRefIsResolved(flagsAndClass, valueOffset))) {
 		buildJITResolveFrameWithPC(currentThread, J9_SSF_JIT_RESOLVE_DATA, parmCount, true, 0, jitEIP);
 		J9RAMStaticFieldRef localRef;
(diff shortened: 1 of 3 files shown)


gacholio:
Some tests failing - more investigation required.

gacholio:

runtime/codert_vm/cnathelp.cpp
```diff
@@ -1429,31 +1429,24 @@ old_fast_jitResolvedFieldIsVolatile(J9VMThread *currentThread)
 static VMINLINE void*
 old_slow_jitResolveFieldImpl(J9VMThread *currentThread, UDATA parmCount, J9ConstantPool *ramConstantPool, UDATA cpIndex, void *jitEIP, bool isSetter, bool direct)
 {
-	void *addr = NULL;
-	J9RAMFieldRef *ramFieldRef = (J9RAMFieldRef*)ramConstantPool + cpIndex;
-	UDATA valueOffset = ramFieldRef->valueOffset;
-	UDATA flags = ramFieldRef->flags;
-	if (J9_UNEXPECTED(!VM_VMHelpers::instanceFieldRefIsResolved(flags, valueOffset, isSetter))) {
```
DanHeidinga: Why don't we check if the CP entry is already resolved?

runtime/codert_vm/cnathelp.cpp
```diff
@@ -1512,41 +1505,35 @@ old_slow_jitResolveFieldSetterDirect(J9VMThread *currentThread)
...
-		}
-		field = (UDATA)vm->internalVMFunctions->resolveStaticFieldRef(currentThread, method, ramConstantPool, cpIndex, resolveFlags, NULL);
-		if ((UDATA)-1 == field) {
-			/* fetch result from floatTemp - stashed there in clinit case */
-			J9RAMStaticFieldRef *fakeRef = (J9RAMStaticFieldRef*)&currentThread->floatTemp1;
-			/* Resolve call indicated success - no need for extra putstatic checks, and the ref
-			 * must contain valid resolved information.
-			 */
-			field = (UDATA)VM_VMHelpers::staticFieldAddressFromRef(fakeRef, false);
-			Assert_CodertVM_false(0 == field);
```
DanHeidinga: We lose the assert here - intentional?

runtime/codert_vm/cnathelp.cpp
gacholio: Rather than duplicate the complexity of the final check, it's much simpler to just run the resolve.

runtime/codert_vm/cnathelp.cpp
gacholio: Yes, the helper used to return 0 for unresolved, now it only applies to resolved.

PR #3925, opened 2018-12-04
Fix and use withfield in value type tests

acrowthe:
Previously, withfield was not correctly implemented due to an incorrect
return type and naming convention. Support was also added for a
`withGeneric` method for testing.

In addition, value type tests used setters instead of withers, causing
errors as `putfield` was used to overwrite final variables.

Previously unused value type testing was re-enabled due to now being
supported.


tajila:
@llxia please review these changes

llxia:
Will forgo PR builds as they will not compile/execute Valhalla tests

tajila:

PR #3928, opened 2018-12-05
VarHandles class initialization

pdbain-ibm:
Initialize the class when creating a VårHandle to a static field.  This change
is to match the behaviour of the reference implementation.

### DIFF
--- a/jcl/src/java.base/share/classes/java/lang/invoke/StaticFieldVarHandle.java
+++ b/jcl/src/java.base/share/classes/java/lang/invoke/StaticFieldVarHandle.java
@@ -1,6 +1,6 @@
 /*[INCLUDE-IF Sidecar19-SE]*/
 /*******************************************************************************
- * Copyright (c) 2016, 2018 IBM Corp. and others
+ * Copyright (c) 2016, 2019 IBM Corp. and others
  *
  * This program and the accompanying materials are made available under
  * the terms of the Eclipse Public License 2.0 which accompanies this
@@ -90,9 +90,10 @@ final class StaticFieldVarHandle extends FieldVarHandle {
 				
 		return populateMHs(operationsClass, lookupTypes, exactTypes);
 	}
-
+	
 	/**
 	 * Constructs a VarHandle to a static field.
+	 * Ensure that the class is initialized to match behaviour of the reference implementation.
 	 * 
 	 * @param lookupClass The class where we start the lookup of the field
 	 * @param fieldName The field name
@@ -101,6 +102,7 @@ final class StaticFieldVarHandle extends FieldVarHandle {
 	 */
 	StaticFieldVarHandle(Class<?> lookupClass, String fieldName, Class<?> fieldType, Class<?> accessClass) {
 		super(lookupClass, fieldName, fieldType, accessClass, true, COORDINATE_TYPES, populateMHs(fieldType));
+		_unsafe.ensureClassInitialized(lookupClass);
 	}
 	
 	/**


DanHeidinga:
@pdbain-ibm Can you drop the "Version 2" from the commit messages?  There's no commits with "Version 1" in them related to this space

pdbain-ibm:
@DanHeidinga In reply to 
https://github.com/eclipse/openj9/pull/3928#issuecomment-446232095, this is a replacement for https://github.com/eclipse/openj9/pull/192 `WIP: VarHandles class initializaiton`.

DanHeidinga:
Which was never merged.  Having a `Version 2` in the git changelog is unnecessary

jcl/src/java.base/share/classes/java/lang/invoke/StaticFieldVarHandle.java
```diff
@@ -101,6 +111,7 @@
 	 */
 	StaticFieldVarHandle(Class<?> lookupClass, String fieldName, Class<?> fieldType, Class<?> accessClass) {
 		super(lookupClass, fieldName, fieldType, accessClass, true, COORDINATE_TYPES, populateMHs(fieldType));
+		ensureClassIsInitialized(lookupClass);
```
DanHeidinga: There's no need for the JITHelpers calls here.  A direct call to `_unsafe.ensureClassInitialized(clazz)` is sufficient as this only called on VarHandle creation.

DanHeidinga: Please also add a comment indicating that the change is to match the RI's behaviour, even though its inconsistent with the javadoc.

pdbain-ibm: Done.  Thanks.

jcl/src/java.base/share/classes/java/lang/invoke/StaticFieldVarHandle.java
```diff
@@ -90,9 +93,17 @@
 				
 		return populateMHs(operationsClass, lookupTypes, exactTypes);
 	}
+	
+	private static final JITHelpers jitHelpers = JITHelpers.getHelpers();
```
DanHeidinga: I think this code can all be removed now.

jcl/src/java.base/share/classes/java/lang/invoke/StaticFieldVarHandle.java
```diff
@@ -22,6 +22,9 @@
  *******************************************************************************/
 package java.lang.invoke;
 
+import com.ibm.jit.JITHelpers;
+import com.ibm.oti.vm.VM;
```
DanHeidinga: This can be removed now as well

pdbain-ibm: All requested changes done.

jcl/src/java.base/share/classes/java/lang/invoke/StaticFieldVarHandle.java
pdbain-ibm: Done.

JasonFengJ9:
This PR is going to fix https://github.com/eclipse/openj9/issues/3205.

`org/openj9/test/varhandle/StaticFieldVarHandleTests.java:testFinalField()` would need to be updated to remove `StaticHelper.reset();` at  https://github.com/eclipse/openj9/blob/master/test/functional/Java9andUp/src/org/openj9/test/varhandle/StaticFieldVarHandleTests.java#L2271.
This verifies that `StaticFieldVarHandle getter` initializes the static fields if they haven't been initialized before.

DanHeidinga:

DanHeidinga:

JasonFengJ9:
The test change mentioned meant to support this fix but not required for this PR.
Will open another PR instead.

PR #3931, opened 2018-12-05
Avoid NPEs running 'info system' when no JRE is found

keithc-ca:
JVM start time (milliseconds and nanoseconds) are only available if a JVM was found.


DanHeidinga:

Issue #94, opened 2017-09-21
javacore does not identify z/OS hardware


pshipton:
The javacore for z/OS shows "OS Level" like "z/OS 02.02.00", and "Architecture   : s390x" but no other details about the hardware being used. Joran has suggested adding in the raw STFLE bits.

joransiu:
In addition to STFLE bits, we'd probably like to document in OS section the auxvector list of features on Linux and relevant CVT bits on z/OS -- would give us the necessary information to understand what the underlying stack support are for the HW features we exploit.

fjeremic:
We read the `STFLE` bits in j9sysinfo.c in function `testSTFLE(struct J9PortLibrary *portLibrary, uint64_t stfleBit)`. This is a good starting place. It looks like we cache the bits in `PPG_stfleCache` which is defined as:

```
#define PPG_stfleCache (portLibrary->portGlobals->platformGlobals.stfleCache)
```

Sounds like we can use this cache to print the `STFLE` double words when we generate the javacore. It looks like we save the CVT bits and auxvector as well in the same file in a nearby place.

PR #3956, opened 2018-12-06
Unify processor detection on Z

fjeremic:
This PR contains several incremental changes. The end goal was to really print the machine id if the processor was unknown. Previously we extract the machine id and compare against a list of known ids. If no id matches we return 0, and when verbose printout is requested via `-Xjit:verbose` we print 0 as the machine id.

This is not ideal as we would like to print the machine id even if we did not match it to one of the known ones. We fix this problem here by eliminating the use of `TR_S390MachineType` enum and just use the machine id as an `int32_t`. In doing so we identify several other opportunities in the area of processor detection, such as severe code duplication, and responsibility movement to the port library.

Please see each commit for a description of what was changed and why.


fjeremic:

runtime/compiler/z/env/J9CPU.cpp
```diff
@@ -49,84 +49,25 @@ namespace J9
...
+   struct utsname info;
 
-static const int S390UnsupportedMachineTypes[] =
-   {
-   TR_G5, TR_MULTIPRISE7000
-   };
+   if (::uname(&info) >= 0)
+      {
+      return atoi(info.machine);
+      }
```
fjeremic: Note we cannot use `utsname` on Linux as it simply returns `"s390x"` in `info.machine`. On z/OS we can indeed use it [1].

[1] https://www.ibm.com/support/knowledgecenter/en/SSLTBW_2.1.0/com.ibm.zos.v2r1.bpxbd00/rtuna.htm

runtime/compiler/z/env/J9CPU.cpp
```diff
@@ -49,84 +49,25 @@ namespace J9
...
-      }
-
-   // Scan list of supported machines.
-   for (int i = 0; i < sizeof(S390MachineTypes) / sizeof(int); ++i)
-      {
-      if (machine == S390MachineTypes[i])
-         {
-         ret_machine = (TR_S390MachineType)machine;
-         }
-      }
```
fjeremic: This is the logic that existed prior to this PR which would force us to return 0 as the machine id even though we did extract it.

runtime/compiler/z/env/J9CPU.cpp
```diff
@@ -415,215 +219,78 @@ CPU::getS390SupportsGuardedStorageFacility()
 ////////////////////////////////////////////////////////////////////////////////
 
 void
-CPU::initializeS390zLinuxProcessorFeatures()
+CPU::initializeS390ProcessorFeatures()
```
fjeremic: `initializeS390zLinuxProcessorFeatures` and `initializeS390zOSProcessorFeatures` are now unified into this function. I suggest to reviewers to compare the before and after for both functions against this new one and convince themselves that they are indeed the same semantically.

runtime/compiler/z/env/J9CPU.cpp
```diff
@@ -415,215 +219,78 @@ CPU::getS390SupportsGuardedStorageFacility()
...
    {
    J9ProcessorDesc *processorDesc = TR::Compiler->target.cpu.TO_PORTLIB_getJ9ProcessorDesc();
    J9PortLibrary *privatePortLibrary = TR::Compiler->portLib;
 
    // The following conditionals are dependent on each other and must occur in this order
+   TR::Compiler->target.cpu.setS390SupportsZ9();
```
fjeremic: This was added as our ALS is z9.

fjeremic:
I'll fix the copyright checks when I start addressing reviews so as to not cancel the currently running build.

runtime/compiler/env/ProcessorDetection.cpp
```diff
@@ -771,14 +663,94 @@ int32_t TR_J9VM::getCompInfo(char *processorName, int32_t stringLength)
...
+         case TR_Z13s:
+            returnValue = snprintf(processorName, stringLength, "z13s (%d)", machineId);
+         break;
+
+         case TR_Z14:
+            returnValue = snprintf(processorName, stringLength, "z14 (%d)", machineId);
+         break;
+
+         case TR_Z14s:
+            returnValue = snprintf(processorName, stringLength, "z14s (%d)", machineId);
```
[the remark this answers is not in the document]
fjeremic: So should we remove it and map both processor ids to z14?

runtime/compiler/ras/DebugExt.cpp
```diff
@@ -2289,9 +2289,6 @@ void TR_DebugExt::dxPrintCompilationInfo(TR::CompilationInfo *remoteCompInfo)
       _dbgPrintf("bool                                  _compBudgetSupport               = %s\n", localCompInfo->_compBudgetSupport ? "true" : "false");
       _dbgPrintf("bool                                  _rampDownMCT                     = %s\n", localCompInfo->_rampDownMCT ? "true" : "false");
       _dbgPrintf("flags32_t                             _flags                           = %u\n", localCompInfo->_flags.getValue());
-#ifdef TR_TARGET_S390
-      _dbgPrintf("TR_S390MachineType                    _s390MachineType                 = %d\n", TR::Compiler->target.cpu.getS390MachineType());
```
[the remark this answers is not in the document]
fjeremic: I don't think it's used. I could not find any reference to this field in our entire codebase. As part of #94 we'll be adding STFLE bits and processor info to the javacore making this redundant. @dchopra001 is currently working on this.

runtime/compiler/z/env/J9CPU.cpp
fjeremic: Fixed in 332b5f3

runtime/compiler/env/ProcessorDetection.cpp
```diff
@@ -771,14 +663,94 @@ int32_t TR_J9VM::getCompInfo(char *processorName, int32_t stringLength)
 #if defined(TR_TARGET_S390)
```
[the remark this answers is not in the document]
fjeremic: The issue here is that the `TO_PORTLIB_get390MachineId` is only defined in the env/z/J9CPU.cpp file which does not get included in the IPATH for other platforms so they will fail to compile if we remove the `#ifdef`. The solution here is to migrate this to the port library in OMR where all this processor detection really belongs. But that's an issue outside the scope of this PR (it's coming though - I'm working on it).

runtime/port/unix/j9sysinfo.c
```diff
@@ -860,15 +860,13 @@ getS390Description(struct J9PortLibrary *portLibrary, J9ProcessorDesc *desc)
 #endif /* defined(S390) && defined(LINUX) */
 
 	/* HIGH_GPRS support */
-#if defined(OMR_ENV_DATA64)
```
fjeremic: Pointing this out explicitly to @joransiu and @NigelYiboYu who has recently got bitten by this.

fjeremic: I'm going to need to think some more about this one. I'm not sure we actually want HPRs to be enabled on 31-bit platforms because there are other dependencies in the VM for example where we do not support collectable references being in HPRs. I'm not sure we want to go down this road to be honest. I'll have to think some more.

fjeremic: Marking this WIP until tomorrow. Going to sleep on it.

runtime/compiler/env/ProcessorDetection.cpp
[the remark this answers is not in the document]
joransiu: Given that the JIT doesn't do anything different between z14 and z14 ZR1, we could just go with referencing both as z14, to be in sync with the official nomenclature.  However, I could also go either way here if we desire to have a clear distinction between the two model codes.

joransiu:
I'll be out of office with questionable internet access.  Perhaps @vijaysun-omr or @gita-omr would be able to take over this PR's review? 

I think the only outstanding issue is related to HIGH_GPR support settings.

runtime/compiler/env/ProcessorDetection.cpp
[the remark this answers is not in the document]
fjeremic: Sure, we can go with just one. I prefer that anyway as both are treated as z14. Similarly for other processors as well.

runtime/port/unix/j9sysinfo.c
fjeremic: So I've done extensive thinking and testing on this and I've decided it is a good idea to enable HPR support on 31-bit Linux as well. This has been completed in 4522048

fjeremic:
Unmarking WIP as everything has been addressed. @vijaysun-omr / @gita-omr are you able to take ownership of this PR as per the above comment by @joransiu?

gita-omr:
Assigned to myself. Which tests should we run?

fjeremic:
> Assigned to myself. Which tests should we run?

This touches some common files as well. So to be conservative let's run everything:

Jenkins test sanity all JDK8

fjeremic:
So it's been a while since this guy has been ready, however I will be away next week and back on the 13th. I would dislike if merging this change causes problems while I'm away and not here to fix anything. Having said that I think it's best to wait until my return before merging this.

I'll continue background testing to make sure nothing has regressed since.

fjeremic:
@joransiu I have removed the "enabling HPRs on Linux on Z 31-bit" commit from this PR. The reason is because I have another rather large change coming to remove `TR_GPR64` from the codebase and consolidate handling of 64-bit registers across all Z platforms. Because this is a massive RA change (~2000 lines or so) I don't feel comfortable making this HPR change beforehand. I'd rather bundle it as part of the RA change to make sure nothing gets broken.

This PR should be ready to go. Launching builds on your behalf.

Jenkins test sanity all JDK8

Issue #3955, opened 2018-12-06
Crash in JIT when `-Xjit:version,verbose` option is applied


fjeremic:
Using a recent build:

```
> jdk8u192-b12-jre-11-14/bin/java -version
openjdk version "1.8.0_192"
OpenJDK Runtime Environment (build 1.8.0_192-201811140430-b12)
Eclipse OpenJ9 VM (build master-7aa10144a, JRE 1.8.0 Linux s390x-64-Bit Compressed References 20181113_90 (JIT enabled, AOT enabled)
OpenJ9   - 7aa10144a
OMR      - fdc56958
JCL      - 4fb7da1e23 based on jdk8u192-b12)
```

adding the `-Xjit:version,verbose` will cause a crash:

```
> jdk8u192-b12-jre-11-14/bin/java -Xjit:version,verbose -Xdump:none -version
JIT: using build "Nov 13 2018 23:35:07"
JIT level: 7aa10144a

#INFO:  _______________________________________
#INFO:  Version Information:
#INFO:       JIT Level  - 7aa10144a
#INFO:       JVM Level  - 20181113_90
#INFO:       GC Level   - 7aa10144a
#INFO:
#INFO:  Processor Information:
#INFO:          Name: z13 (2964)
#INFO:        Vendor: IBM
#INFO:       numProc: 8
#INFO:           DFP: 1
#INFO:           FPE: 1
#INFO:          HPRD: 1
#INFO:            RI: 1
#INFO:            TM: 1
#INFO:            VF: 1
#INFO:            GS: 0
#INFO:
#INFO:  _______________________________________
#INFO:  AOT
#INFO:  options specified:
#INFO:       samplingFrequency=2,numInterfaceCallCacheSlots=4
#INFO:
#INFO:  options in effect:
#INFO:       numInterfaceCallCacheSlots=4
#INFO:       verbose=1Unhandled exception
Type=Segmentation error vmState=0x00000000
J9Generic_Signal_Number=00000004 Signal_Number=0000000b Error_Value=00000002 Signal_Code=00000001
Handler1=000003FFBC9AD8E8 Handler2=000003FFBC7A7798 InaccessibleAddress=0000000000000000
gpr0=0000000000000000 gpr1=0000000000000000 gpr2=000003FFB7D40DE8 gpr3=000003FFB80B1A10
gpr4=000003FFB7968378 gpr5=000003FF00000002 gpr6=000003FFB7D64080 gpr7=000003FFB7B0D5FC
[...]

PR #3965, opened 2018-12-07
Fix invalid option definitions causing crashes in the JIT

fjeremic:
The `TR::OptionTable::msg` field must be non-null, otherwise when
printing the options via the `verbose` command we run into null pointer
dereferences which causes the JIT to crash.

Fixes: #3955


fjeremic:
@dsouzai could I please get a review?

runtime/compiler/control/J9Options.cpp
```diff
@@ -981,13 +981,13 @@ TR::OptionTable OMR::Options::_feOptions[] = {
    {"verbose=",           "L{regex}\tlist of verbose output to write to vlog or stdout",
         TR::Options::setVerboseBitsInJitPrivateConfig, offsetof(J9JITConfig, privateConfig), 0, "F"},
    {"version",            "L\tdisplay the jit build version",
-        TR::Options::versionOption, 0, 0},
+        TR::Options::versionOption, 0, 0, "F"},
```
dsouzai: Isn't the second `0` supposed to be replaced with the `"F"?

runtime/compiler/control/J9Options.cpp
```diff
@@ -981,13 +981,13 @@ TR::OptionTable OMR::Options::_feOptions[] = {
...
    {"veryHotSampleThreshold=",          "R<nnn>\tThe maximum number of global samples taken during a sample interval for which the method will be recompiled at hot with normal priority",
         TR::Options::setStaticNumeric, (intptrj_t)&TR::Options::_veryHotSampleThreshold, 0, " %d", NOT_IN_SUBSET},
    {"vlog=",              "L<filename>\twrite verbose output to filename",
         TR::Options::setString,  offsetof(J9JITConfig,vLogFileName), 0, "F%s"},
    {"vmState=",           "L<vmState>\tdecode a given vmState",
-        TR::Options::vmStateOption, 0, 0},
+        TR::Options::vmStateOption, 0, 0, "F"},
```
dsouzai: Same comment as above.

runtime/compiler/control/J9Options.cpp
fjeremic: I don't think so. If you take a look at the `"verbose="` option above and the actual definition of `TR::OptionTable` [1] you will see the order is, `name`, `helpText`, `fcn`, `parm1`, `parm2`, and then `msg` which is not specified here (so assumed `NULL` - see [2] for a live example).

[1] https://github.com/eclipse/omr/blob/7cf542c14aa04cc5e5baafa4cd88a7b3fbef6f98/compiler/control/OptionsUtil.hpp#L48-L137
[2] https://ideone.com/03Ediv

mstoodle:
A very safe fix and no platform specific aspect. Thanks @fjeremic !

PR #3967, opened 2018-12-07
Don't build JitBuilderRecorder files from OMR yet

mstoodle:
An OMR pull request will be merged soon that introduces four new
files as part of the JitBuilder library. While I am in the process
of bringing JitBuilder support into OpenJ9, I am not sure if these
files will compile or not as part of OpenJ9, so let's leave them
out for now. They are useless to OpenJ9 now anyway.

I'll bring these files back in as part of the larger effort
to bring JitBuilder into OpenJ9.

https://github.com/eclipse/omr/pull/3019 is the OMR pull request.


charliegracie:
This blacklist of classes needs to be handled in a better manner.  Adding a new file in OMR that isn’t used by OpenJ9 should not require a change to OpenJ9.  I don’t think this PR needs to make that change but this needs to become a priority so we do not have this churn in OpenJ9.

charliegracie:
I believe the Travis (non Lint) build is using CMake so this change should be good to go without further testing.

Issue #3851, opened 2018-11-27
IllegalAccessException when calling default methods on interfaces


mseele:
There seem's to be an issue in openj9 when calling default methods on interfaces. 
Please see https://github.com/lviggiano/owner/issues/243

Here is a snippet to reproduce the error:

```
import org.aeonbits.owner.ConfigFactory;

import com.automation.remarks.video.SystemUtils;
import com.automation.remarks.video.recorder.VideoConfiguration;

public class Test {

	public static void main(String[] args) {
		ConfigFactory.setProperty("os.type", SystemUtils.getOsType()); //$NON-NLS-1$
		VideoConfiguration videoConfiguration = ConfigFactory.create(VideoConfiguration.class, System.getProperties());
		System.err.println(videoConfiguration.screenSize());
	}

}
```

...and here is the maven dependencies to run this code:

```
<dependencies>
	<dependency>
		<groupId>com.automation-remarks</groupId>
		<artifactId>video-recorder-core</artifactId>
		<version>1.8</version>
	</dependency>
	<dependency>
		<groupId>org.aeonbits.owner</groupId>
		<artifactId>owner</artifactId>
		<version>1.0.10</version>
	</dependency>
	<dependency>
		<groupId>org.aeonbits.owner</groupId>
		<artifactId>owner-java8</artifactId>
		<version>1.0.10</version>
	</dependency>
</dependencies>
```

It work's well with openjdk-11 hotspot (https://adoptopenjdk.net/?variant=openjdk11&jvmVariant=hotspot) and java 8.

DanHeidinga:
Copying the stack trace here:
```
Exception in thread "main" java.lang.reflect.UndeclaredThrowableException
	at com.sun.proxy.$Proxy7.screenSize(Unknown Source)
	at Test.main(Test.java:13)
Caused by: java.lang.IllegalAccessException: org.aeonbits.owner.util.Java8SupportImpl could not access com.automation.remarks.video.recorder.VideoConfiguration - private access required
	at java.base/java.lang.invoke.MethodHandles$Lookup.checkSpecialAccess(MethodHandles.java:493)
	at java.base/java.lang.invoke.MethodHandles$Lookup.findSpecial(MethodHandles.java:517)
	at org.aeonbits.owner.util.Java8SupportImpl.invokeDefaultMethod(Java8SupportImpl.java:47)
	at org.aeonbits.owner.util.Reflection.invokeDefaultMethod(Reflection.java:68)
[...]

PR #3978, opened 2018-12-10
Fix the access check in findSpecial

ChengJin01:
The change is to add the access check
in invoking interface's methods via
findSpecial.

Fix: #3851

### DIFF
--- a/jcl/src/java.base/share/classes/java/lang/invoke/MethodHandles.java
+++ b/jcl/src/java.base/share/classes/java/lang/invoke/MethodHandles.java
@@ -549,12 +549,19 @@ public class MethodHandles {
 			throw new IllegalAccessException(com.ibm.oti.util.Msg.getString("K0680", accessClass.getName(), targetClass.getName()));  //$NON-NLS-1$
 		}
 		
-		private void checkSpecialAccess(Class<?> callerClass) throws IllegalAccessException {
+		private void checkSpecialAccess(Class<?> declaringClass, Class<?> callerClass) throws IllegalAccessException {
 			if (INTERNAL_PRIVILEGED == accessMode) {
 				// Full access for use by MH implementation.
 				return;
 			}
-			if (isWeakenedLookup() || accessClass != callerClass) {
+			/* Given that findSpecial emulates the behavior of invokespecial,
+			 * calling interface methods of declaringClass should be allowed as long as 
+			 * declaringClass is an interface and callerClass can be assigned to it when
+			 * callerClass is not the same as the lookup class.
+			 */
+			if (isWeakenedLookup() 
+			|| ((accessClass != callerClass) && !(declaringClass.isInterface() && declaringClass.isAssignableFrom(callerClass)))
+			) {
 				/*[MSG "K0585", "{0} could not access {1} - private access required"]*/
 				throw new IllegalAccessException(com.ibm.oti.util.Msg.getString("K0585", accessClass.getName(), callerClass.getName())); //$NON-NLS-1$
 			}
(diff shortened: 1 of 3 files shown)


ChengJin01:
The changes cover 2 two cases
1) if callerClass == lookup class, the lookup class must be the same or a subclass of the current class.
2) if callerClass != lookup class, the declaring class must be an interface which the callerClass can be assigned to.

New changes was verified on Java 8 & Java 11 in personal builds.
New test cases were verfied on Both our build and OpenJDK/Hotspot.

Reviewer: @DanHeidinga

jcl/src/java.base/share/classes/java/lang/invoke/MethodHandles.java
```diff
@@ -549,12 +549,20 @@ private void checkClassAccess(Class<?> targetClass) throws IllegalAccessExceptio
...
 			}
-			if (isWeakenedLookup() || accessClass != callerClass) {
+			/* Given that findSpecial emulates the behavior of invokespecial,
+			 * calling interface methods of declaringClass should be allowed as long as 
+			 * declaringClass is an interface and callerClass can be assigned to it when
+			 * callerClass is not the same as the lookup class.
+			 */
+			if (isWeakenedLookup() 
+			|| ((accessClass != callerClass) 
+			&& !((null != declaringClass) && declaringClass.isInterface() && declaringClass.isAssignableFrom(callerClass)))
```
DanHeidinga: When can `declaringClass` be null?  I'm not sure I saw a case that allowed that.

test/functional/Jsr292/src/com/ibm/j9/jsr292/LookupAPITests_Find.java
```diff
@@ -2098,6 +2099,35 @@ public void test_FindSpecial_Public_CrossPackage_InnerClass() throws Throwable {
...
+	 * method handle will be bound to) is in a different package.
+	 * Note: when the lookup class is not equal to the caller class, the declaring class
+	 * must be an interface which the caller class can be assigned to.
+	 * @throws Throwable
+	 */
+	@Test(groups = { "level.extended" })
+	public void test_FindSpecial_Public_CrossPackage_Not_Interface() throws Throwable {
+		boolean illegalAccessExceptionThrown = false;
+		try {
+			MethodHandles.lookup().findSpecial(PackageExamples.class, "addPublic", MethodType.methodType(int.class, int.class, int.class), PackageExamples.class);
```
DanHeidinga: The preferred pattern for this kind of test is to use a `fail` method after the code that's expected to throw the exception like:
https://github.com/eclipse/openj9/blob/b8cdef32c53c4c2fd1e427e1692a49ac8c4d5634/test/functional/Jsr292/src/com/ibm/j9/jsr292/LookupAPITests_Find.java#L375-L378

jcl/src/java.base/share/classes/java/lang/invoke/MethodHandles.java
```diff
@@ -580,12 +588,15 @@ private void checkSpecialAccess(Class<?> callerClass) throws IllegalAccessExcept
...
+			checkSpecialAccess(clazz, specialToken);	/* Must happen before method resolution */
 			MethodHandle handle = null;
 			try {
 				handle = findSpecialImpl(clazz, methodName, type, specialToken);
 				Class<?> handleDefc = handle.getDefc();
-				if ((handleDefc != accessClass) && !handleDefc.isAssignableFrom(accessClass)) {
+				/* Check the relationship between the lookup class and the current class
+				 * only when the requested caller class is the same as the lookup class.
+				 */
+				if ((specialToken == accessClass) && !handleDefc.isAssignableFrom(accessClass)) {
```
DanHeidinga: I'm not sure I follow this logic.  The specialToken could be a subclass of the accessClass.  I would still expect it to have to be assignable in that case.

DanHeidinga:
Jenkins test all zlinux jdk8

jcl/src/java.base/share/classes/java/lang/invoke/MethodHandles.java
ChengJin01: Already removed as unnecessary.

test/functional/Jsr292/src/com/ibm/j9/jsr292/LookupAPITests_Find.java
ChengJin01: Updated the pattern as suggested.

jcl/src/java.base/share/classes/java/lang/invoke/MethodHandles.java
ChengJin01: I need to double-check the case to see whether it works for all existing related test cases.

ChengJin01: I already verified all JSR292 related test case in personal build on Java 8 & 11 and nothing wrong was detected. So updated to `accessClass.isAssignableFrom(specialToken)` as specialToken should be assignable to the accessClass in our case.

DanHeidinga:
Jenkins test all zlinux jdk8

PR #3979, opened 2018-12-10
Stop adding ext repo shas to omr promote tag

AdamBrousseau:
- No longer add extentions repo SHA metadatata to the omr promotion tags
- Fix mising comma in list of arguments for build_with_slack function call


AdamBrousseau:
@DanHeidinga for review

DanHeidinga:
Summarizing our slack discussion here for posterity:
* Originally tagged all repos so that we could checkout the same version after a promotion build to be able to understand how something escaped through the build.
* We're not aware of anyone doing this 
* Extensions repos change much slower than the openj9 / omr repos so we can reasonably "guess" the shas of the relevant extensions repos
* Current tagging hasn't kept up with the new 6 month JDK release cadence... and no one has noticed

Seems fine to tag just omr / openj9 now

PR #3980, opened 2018-12-10
Fix Java 12 build failures

theresa-m:
Errors in java.lang.System caused by internal apis

JEP 334 stubs for Java 12 build needed for compilation
- stubs for java.lang.Class
- stubs for java.lang.invoke.VarHandle.VarHandleDesc

12 builds with extensions up to 0b941d8:
- new unsafe methods xxReference and natives, invokeCleaner
- MethodHandle.filterArgumentsWithCombiner()
- MethodHandle.foldArgumentsWithCombiner()

Fixes: #3768
Fixes: #4158 


theresa-m:
investigating reason for pr build failure

theresa-m:
update jcl version 12 to vmconstantpool.xml additions. pr build should pass

theresa-m:
I've added a commit to fix the compile errors described in #4158 to this pull request. This branch should now work again to compile Java 12.

theresa-m:
update copyright dates to 2019

theresa-m:
rerunning ci build. build is working fine in jenkins

theresa-m:
fyi @keithc-ca

jcl/src/java.base/share/classes/java/lang/Class.java
```diff
@@ -38,12 +38,18 @@
...
+import java.util.Optional;
+/*[ENDIF]*/
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.atomic.AtomicInteger;
 import java.security.AccessController;
 import java.security.PrivilegedExceptionAction;
 import java.security.PrivilegedAction;
 import java.lang.ref.*;
+/*[IF Java12]*/
+import java.lang.constant.ClassDesc;
```
keithc-ca: nit: if this import is in a separate conditional, please move it into the `java.lang` territory.

jcl/src/java.base/share/classes/java/lang/Class.java
```diff
@@ -4545,4 +4551,14 @@ public boolean isNestmateOf(Class<?> that) {
...
+
+/*[IF Java12]*/
+	public Class<?> arrayType() {
+		throw new UnsupportedOperationException("Stub for Java 12 compilation");
+	}
+
+	public Optional<ClassDesc> describeConstable() {
+		throw new UnsupportedOperationException("Stub for Java 12 compilation");
+	}
+/*[ENDIF]*/
```
keithc-ca: I'd prefer that these closing directives were `/*[ENDIF] Java12 */` to match the style used in C/C++ code which makes it easier to associate with the opening directive.

jcl/src/java.base/share/classes/java/lang/System.java
```diff
@@ -397,41 +395,84 @@ private static void ensureProperties() {
...
+
+	if (isInitialization) {
+		/*[PR CMVC 179976] System.setProperties(null) throws IllegalStateException */
+		try {
+		/*[IF Java12]*/
+			VM.saveProperties(initializedProperties);
+		/*[ELSE]
+			VM.saveAndRemoveProperties(initializedProperties);
+		/*[ENDIF]*/
+		} catch (NoSuchMethodError e) {
```
keithc-ca: Please add a comment explaining why this exception is being ignored.

jcl/src/java.base/share/classes/java/lang/invoke/FilterArgumentsWithCombinerHandle.java
```diff
@@ -0,0 +1,91 @@
...
+        if (!ILGenMacros.isCustomThunk()) {
+            doCustomizationLogic();
+        }
+
+        return ILGenMacros.invokeExact_X(handle, ILGenMacros.placeholder(
+            ILGenMacros.firstN(filterPosition(), argPlaceholder), 
+            ILGenMacros.invokeExact(combiner, argumentsForCombiner(argumentIndices(), argPlaceholder)),
+            ILGenMacros.lastN(numSuffixArgs(), argPlaceholder)));
+    }
+}
```
keithc-ca: Please add a final new-line to this file.

runtime/vm/MHInterpreter.cpp
```diff
@@ -64,6 +64,7 @@ static const char * names[] = {
 /* #ifdef J9VM_OPT_PANAMA */
 	"J9_METHOD_HANDLE_KIND_NATIVE",
 /* #endif J9VM_OPT_PANAMA */
+	"J9_METHOD_HANDLE_KIND_FILTER_ARGUMENTS_WITH_COMBINER",
```
keithc-ca: A placeholder is needed for `J9_METHOD_HANDLE_KIND_NATIVE` even if `J9VM_OPT_PANAMA` is not defined else this table will be too short.

jcl/src/java.base/share/classes/java/lang/System.java
theresa-m: I actually have no idea what this is for. This was something I copied over when I refactored this method to make it compatible with 12. I can see what happens when its missing in an 8 build..

theresa-m: I've applied your suggested changes in a new commit except for this one atm. thanks for the review

keithc-ca: I suggest we remove the try-catch surroundings: It's hard to imagine why `NoSuchMethodError` would be thrown and, for me, harder to explain why it should be ignored.

theresa-m: okay. it has been removed.

jcl/src/java.base/share/classes/java/lang/System.java
```diff
@@ -397,41 +395,81 @@ private static void ensureProperties() {
...
+/*[ENDIF] Java12 */
 	/*[IF Sidecar19-SE]*/
-	useLegacyVerPresents = systemProperties.containsKey("use.legacy.version");
+	useLegacyVerPresents = initializedProperties.containsKey("use.legacy.version");
 	/*[ENDIF]*/
+
+	if (isInitialization) {
+		/*[PR CMVC 179976] System.setProperties(null) throws IllegalStateException */
+		/*[IF Java12]*/
+			VM.saveProperties(initializedProperties);
```
keithc-ca: nit: please remove the indentation (that _was_ appropriate before the try-catch was removed).

theresa-m: woops. updated

theresa-m:
Thanks for the review @ChengJin01. The vm was not able to start without the new MethodHandles methods so I am reasonable confident it is working. I will work on some tests around edge cases to submit in a later pr. There is pr https://github.com/eclipse/openj9/pull/4531 to address Java 12 tests compilation failures that needs to be in before new tests. (tracking this here https://github.com/eclipse/openj9/issues/4571)

andrewcraik:
@jdmpapin @liqunl @cathyzhyi this change adds some Unsafe APIs and some new MethodHandles. Could you please take a look - there may need to be follow-on PRs or suggestions to @theresa-m on how to add the appropriate acceleration paths etc in the JIT.

jcl/src/java.base/share/classes/java/lang/invoke/MethodHandles.java
```diff
@@ -2820,7 +2820,99 @@ public static MethodHandle filterArguments(MethodHandle handle, int startPositio
 		MethodHandle result = FilterArgumentsHandle.get(handle, startPosition, filters, newType);
 		return result;
 	}
+
+/*[IF Java12]*/
+	/**
+	 * Produce a MethodHandle that preprocesses some of the arguments by calling the preprocessor handle.
```
liqunl: Can this be more descriptive? An example would help a lot. I have to read the implementation to guess what the MethodHandle returned by this API is supposed to do. The MethodHandle seems to replace only the argument of `handle` at `foldPosition` with the result of the combiner handle. The combiner handle takes arguments at positions specified by `argumentIndices`.

jcl/src/java.base/share/classes/java/lang/invoke/FilterArgumentsWithCombinerHandle.java
liqunl: This class has to implement `compareWith`. https://github.com/eclipse/openj9/blob/master/jcl/src/java.base/share/classes/java/lang/invoke/MethodHandle.java#L1421

I think it should be similar or exactly the same as `FoldHandle`'s implementation

liqunl:
The new native Unsafe APIs should be mapped to the Object version in j9method.cpp. An example is
```
       {x(TR::sun_misc_Unsafe_compareAndExchangeObject_jlObjectJjlObjectjlObject_Z, "compareAndExchangeObject", "(Ljava/lang/Object;JLjava/lang/Object;Ljava/lang/Object;)Ljava/lang/Object;")},
+      {x(TR::sun_misc_Unsafe_compareAndExchangeObject_jlObjectJjlObjectjlObject_Z, "compareAndExchangeReference", "(Ljava/lang/Object;JLjava/lang/Object;Ljava/lang/Object;)Ljava/lang/Object;")},
```

theresa-m:
thanks @liqunl I've made changes as per your review

liqunl:
I have looked at the implementation of `FilterArgumentsWithCombinerHandle` in java code and the implementation for its JIT macros, both look good. The newly added `FoldHandle` API just calls to existing method, that one looks good too. The recognition of the newly added Unsafe APIs are also good.

jcl/src/java.base/share/classes/java/lang/invoke/FilterArgumentsWithCombinerHandle.java
```diff
@@ -0,0 +1,107 @@
...
+        return new FilterArgumentsWithCombinerHandle(handle, filterPosition, combiner, argumentIndices);
+    }
+
+    @Override
+    MethodHandle cloneWithNewType(MethodType newType) {
+        return new FilterArgumentsWithCombinerHandle(this, newType);
+    }
+
+    private static Object[] infoAffectingThunks(MethodType combinerType, int filterPosition, int...argumentIndices) {
+        Object[] result = {combinerType, filterPosition, argumentIndices};
```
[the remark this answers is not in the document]
theresa-m: I'll add this change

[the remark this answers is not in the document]
theresa-m: thanks for the review

theresa-m:
thanks everyone who participating in reviewing from the JIT team. @DanHeidinga any last comments?

jcl/src/java.base/share/classes/jdk/internal/misc/Unsafe.java
```diff
@@ -4092,6 +4324,68 @@ public final Object getAndSetObjectAcquire(Object obj, long offset, Object value
...
+	 * 
+	 * @param obj object into which to set the value
+	 * @param offset location to set value in obj
+	 * @param value to set in obj memory
+	 * @return value of field in obj at offset before update
+	 */
+	public final Object getAndSetReference(Object obj, long offset, Object value) {
+		for (;;) {
+			Object objectAtOffset = getReferenceVolatile(obj, offset);
+			if (weakCompareAndSetReference(obj, offset, objectAtOffset, value)) {
```
DanHeidinga: `weakCompareAndSetReference` is this spec'd to have `weak` memory ordering effects?  Then it won't meet the `get/setVolatile` semantics.

jcl/src/java.base/share/classes/java/lang/invoke/FilterArgumentsWithCombinerHandle.java
```diff
@@ -0,0 +1,108 @@
...
+ *
+ * [1] https://www.gnu.org/software/classpath/license.html
+ * [2] http://openjdk.java.net/legal/assembly-exception.html
+ *
+ * SPDX-License-Identifier: EPL-2.0 OR Apache-2.0 OR GPL-2.0 WITH Classpath-exception-2.0 OR LicenseRef-GPL-2.0 WITH Assembly-exception
+ *******************************************************************************/
+package java.lang.invoke;
+
+final class FilterArgumentsWithCombinerHandle extends MethodHandle {
+    protected final MethodHandle handle;
```
DanHeidinga: The convention is to name the next methodHandle `next`

jcl/src/java.base/share/classes/java/lang/invoke/FilterArgumentsWithCombinerHandle.java
```diff
@@ -0,0 +1,108 @@
...
+    protected final MethodHandle combiner;
+    private final int filterPosition;
+    private final int[] argumentIndices;
+
+    FilterArgumentsWithCombinerHandle(MethodHandle handle, int filterPosition, MethodHandle combiner, int... argumentIndices) {
```
DanHeidinga: Can you add a comment either to the constructor or the class that describes the algorithm?  Is it combine then filter?  or filter than combine?

jcl/src/java.base/share/classes/java/lang/invoke/MethodHandles.java
```diff
@@ -2791,7 +2791,112 @@ public static MethodHandle filterArguments(MethodHandle handle, int startPositio
...
 	
+		if (preprocessorTypeParamCount != argIndexCount) {
+			/*[MSG "K0638", "The count of argument indices: {0} must be equal to the parameter count of the combiner: {1}"]*/
+			throw new IllegalArgumentException(Msg.getString("K0638", new Object[] { //$NON-NLS-1$
+							Integer.toString(argIndexCount),
+							Integer.toString(preprocessorTypeParamCount)}));
+		}
+		
+		for (int i = 0; i < argIndexCount; i++) {
+			if ((argumentIndices[i] < 0) || (argumentIndices[i] >= handleTypeParamCount)) {
```
DanHeidinga: We need to make a defensive copy of `argumentIndices` or the user could modify the array after we've validated it causing corruption / incorrect behaviour

runtime/vm/MHInterpreter.cpp
```diff
@@ -1353,6 +1366,127 @@ VM_MHInterpreter::getArgSlotsBeforePosition(j9object_t argumentTypes, U_32 argPo
...
+	U_32 argumentIndicesCount = J9INDEXABLEOBJECT_SIZE(_currentThread, argumentIndices);
+	UDATA *spFirstFilterArgSlot = _currentThread->sp + filterArgSlots;
+
+	/* [... filterHandle args descriptionBytes MethodTypeFrame] */
+	UDATA *spPriorToFrameBuild = _currentThread->sp;
+	(void)buildMethodTypeFrame(_currentThread, filterHandleType);
+
+	/* [... filterHandle args descriptionBytes MethodTypeFrame filterHandle args] */
+	_currentThread->sp -= (filterArgSlots  + 1);
+	memcpy(_currentThread->sp, spPriorToFrameBuild, sizeof(UDATA) * (filterArgSlots+1));
```
DanHeidinga: I don't think this memcpy of the arguments is actually used.  There's also nothing that describes them on the stack so if they were used, we'd have a potential issue post-gc.

Can you try removing this?

runtime/vm/MHInterpreter.cpp
```diff
@@ -1353,6 +1366,127 @@ VM_MHInterpreter::getArgSlotsBeforePosition(j9object_t argumentTypes, U_32 argPo
...
+	((j9object_t *)_currentThread->sp)[filterArgSlots] = combinerHandle;
+	UDATA *spCombinerSlot = _currentThread->sp + filterArgSlots;
+
+	/* Copy all arguments specified by argumentIndices from filterHandle to the stack slots used by combinerHandler */
+	U_32 arrayIndex = 0;
+	for (arrayIndex = 0; arrayIndex < argumentIndicesCount; arrayIndex++) {
+		U_32 argumentTypeIndex = (U_32)J9JAVAARRAYOFINT_LOAD(_currentThread, argumentIndices, arrayIndex);
+
+		/* Determine the slot index of each argument specified by the index in argumentIndices */
+		U_32 argumentTypeSlots = getArgSlotsBeforePosition(argumentTypes, argumentTypeIndex);
```
DanHeidinga: This is `N*M` in terms of the number of arguments.  How many arguments are commonly used here?  

If the common case from the class libraries is 2 or 3 arguments, this isn't so bad.  If it's more than that, we'll want to do something better here.

runtime/vm/MHInterpreter.cpp
```diff
@@ -1353,6 +1366,127 @@ VM_MHInterpreter::getArgSlotsBeforePosition(j9object_t argumentTypes, U_32 argPo
...
+			|| (_vm->longReflectClass == argumentClassAtIndex)
+		) {
+			spCombinerSlot -= 2;
+			*(U_64*)spCombinerSlot = *(U_64*)(spFirstFilterArgSlot - argumentTypeSlots - 2);
+		} else {
+			spCombinerSlot -= 1;
+			*spCombinerSlot = *(spFirstFilterArgSlot - argumentTypeSlots - 1);
+		}
+	}
+	_currentThread->sp = spCombinerSlot; /* after loop spCombinerSlot is at the top of the combiner arguments */
```
DanHeidinga: This shouldn't be necessary as `UDATA *spCombinerSlot = _currentThread->sp + filterArgSlots;`

Maybe add an assert that `_currentThread->sp == spCombinerSlot`?

runtime/vm/MHInterpreter.cpp
```diff
@@ -1353,6 +1366,127 @@ VM_MHInterpreter::getArgSlotsBeforePosition(j9object_t argumentTypes, U_32 argPo
...
+	}
+
+	/* Determine the number of argument slots starting after the filterPosition */
+	U_32 remainingArgSlots = filterArgSlots - argumentSlots - combinerReturnSlots;
+
+	/* Set local variables to restore state of the _currentThread during updateVMStruct */
+	UDATA *mhPtr = UNTAGGED_A0(frame);
+
+	/* Advance past the frame and single argument (now filterHandle is at the top of the stack)) */
+	bp = (UDATA *)(((J9SFStackFrame*)(bp+1)) + 1);
```
DanHeidinga: This assumes that the J9SFMethodTypeFrame is immediately after the PlaceHolderFrame:
```
J9SFStackFrame
MethodHandle
J9SFMethodTypeFrame
```
I don't see how this works with the extra copy of the args from the `filterArgumentsWithCombiner` method.

DanHeidinga:
Are there test cases for the new MethodHandles?  In a separate PR is fine.

jcl/src/java.base/share/classes/jdk/internal/misc/Unsafe.java
theresa-m: `weakCompareAndSetReference` just calls the `compareAndSetReference` native. I can update this method with `compareAndSetReference` to make the memory ordering more clear.

DanHeidinga: We should call the method defined to have the memory semantics we want as the JIT may optimize some of these calls.  It would introduce a bug if we call the "wrong" one because we know the implementation and the JIT replaces it with an intrinsics with different memory semantics

theresa-m: ah okay. in that case the other `getAndSetXXX` methods are wrong as well. I will fix them in a separate pr

keithc-ca: Why a separate PR?

theresa-m: mostly because this pr has only java 12 changes. i'll change it here if you prefer

keithc-ca: My understanding was that @DanHeidinga was pointing out changes required to methods _added_ by this PR for Java 12. I'd prefer that we don't start with versions we know are wrong.

DanHeidinga: Fix the Java 12 ones in this PR.  The other APIs can be a separate one.

I'd like to see our 12 builds up and running asap

runtime/vm/MHInterpreter.cpp
theresa-m: I'm seeing argument lengths of only 2 & 3

runtime/vm/MHInterpreter.cpp
theresa-m: It's necessary to do this here because the sp is previously set at the number of arguments for the "next" handle after the memcpy. This sets it to the correct number of arguments for the combiner handle which might be different. I will apply this comment to the version I'm working on without the memcpy though.

runtime/vm/MHInterpreter.cpp
theresa-m: I think the extra args in the stack comment is a mistake it should be `/* [... filterHandle args descriptionBytes MethodTypeFrame filterHandle PlaceHolderFrame combinerReturnValue] */` similar to what it was in the filterArgumentsWithCombinerHandle method

theresa-m:
Thanks for the review @DanHeidinga I've just pushed a commit that applies your changes. looking at this java.lang.System merge conflict that came up

runtime/vm/MHInterpreter.cpp
DanHeidinga: Ok.  Let's leave it for now as the JIT implementation won't have this problem.  If it shows up in profiles, we can look for other ways to optimize it

theresa-m:
Merged my java.lang.System changes with those for https://github.com/eclipse/openj9/issues/4660. Confirmed that this test is still passing.

runtime/vm/MHInterpreter.cpp
```diff
@@ -1353,6 +1366,126 @@ VM_MHInterpreter::getArgSlotsBeforePosition(j9object_t argumentTypes, U_32 argPo
...
+		} else {
+			spCombinerSlot -= 1;
+			*spCombinerSlot = *(spFirstFilterArgSlot - argumentTypeSlots - 1);
+		}
+	}
+	/* after loop spCombinerSlot is at the top of the combiner arguments */
+	Assert_VM_true(spCombinerSlot == _currentThread->sp);
+
+	/* [... filterHandle args descriptionBytes MethodTypeFrame filterHandle PlaceHolderFrame combinerHandle combinerArgs] */
+	insertPlaceHolderFrame2(combinerArgSlots, methodHandle, J9VMJAVALANGINVOKEMETHODHANDLE_FILTERARGUMENTSWITHCOMBINERPLACEHOLDER_METHOD(_vm));
```
DanHeidinga: Rather than introducing the new helper, write the combiner MH into the stack after making the placeholder frame.
```C
insertPlaceHolderFrame(combinerArgSlots, methodHandle, J9VMJAVALANGINVOKEMETHODHANDLE_FILTERARGUMENTSWITHCOMBINERPLACEHOLDER_METHOD(_vm));
((j9object_t *)_currentThread->sp)[combinerArgSlots] = combinerHandle;
```

keithc-ca:
To be certain I don't misunderstand your comments @DanHeidinga, I've reassigned this to you to merge when you're satisfied.

runtime/vm/MHInterpreter.cpp
theresa-m: updated

DanHeidinga:

DanHeidinga:

pshipton:
Running a jdk12 sanity with these changes and #4531
https://ci.eclipse.org/openj9/view/Pipelines/job/Pipeline-Release-Build/40/

pshipton:
Note the known issues with jdk12 sanity in https://github.com/eclipse/openj9/projects/12

DanHeidinga:
Given the jdk11 build passed and the jdk12 compiles passed in Peter's build, I'm inclined to merge this.  Any additional sanity test failures in 12 can be addressed in separate PRs

theresa-m:
It looks to me like the failing pr build is not actually using these changes:
```
17:12:28 Using OPENJ9_REPO = https://github.com/eclipse/openj9.git OPENJ9_BRANCH = master OPENJ9_SHA = 89b39276bc647c6eafa5127db9dd3e3040e9a72b
```

pshipton:
jdk12 testing shows only know issues.

cmdLineTester_CryptoTest_0 - openssl not ported to jdk12 yet

https://github.com/eclipse/openj9/issues/4658
TestFlushReflectionCache_0

https://github.com/eclipse/openj9/issues/4661
openj9_jsr292Test_0
openj9_jsr292Test_JitCount0_0

PR #3983, opened 2018-12-10
Add `cfdump -x` support for bytecodes 203 and 204

acrowthe:
Previously, `cfdump -x` and `cfdump -xm` incorrectly printed the new `withfield` and `defaultvalue` bytecodes. Instead of being treated as a 3 byte bytecode, all 3 bytes were interpreted and printed separately, leading to incorrect output.

Now, all 3 bytes will be recognized as the same bytecode. In addition, more relevant information will be output alongside the bytecode name.


tajila:
@pshipton please review

PR #3991, opened 2018-12-11
Add GC map for loading J9Class for rd/wrtbar

cathyzhyi:
The indirect load, loading J9Class from the object, generated for rdbar
and wrtbar can cause NPE and needs a GC map.

### DIFF
--- a/runtime/compiler/x/codegen/J9TreeEvaluator.cpp
+++ b/runtime/compiler/x/codegen/J9TreeEvaluator.cpp
@@ -137,29 +137,37 @@ inline void generateLoadJ9Class(TR::Node* node, TR::Register* j9class, TR::Regis
    {
    bool needsNULLCHK = false;
    TR::ILOpCodes opValue = node->getOpCodeValue();
-   switch (opValue)
+
+   if (node->getOpCode().isReadBar() || node->getOpCode().isWrtBar())
+      needsNULLCHK = true;
+   else
       {
-      case TR::checkcastAndNULLCHK:
-         needsNULLCHK = true;
-         break;
-      case TR::icall: // TR_checkAssignable
-         return; // j9class register already holds j9class
-      default:
-         TR_ASSERT(node->getOpCode().isReadBar() ||
-                   node->getOpCode().isWrtBar() ||
-                   opValue == TR::checkcast ||
-                   opValue == TR::instanceof,
-                   "Unexpected opCode for generateLoadJ9Class %s.", node->getOpCode().getName());
-         break;
+      switch (opValue)
+         {
+         case TR::checkcastAndNULLCHK:
+            needsNULLCHK = true;
+            break;
+         case TR::icall: // TR_checkAssignable
+            return; // j9class register already holds j9class
+         default:
+            TR_ASSERT(opValue == TR::checkcast ||
+                      opValue == TR::instanceof,
+                     "Unexpected opCode for generateLoadJ9Class %s.", node->getOpCode().getName());
+            break;
+         }
       }
+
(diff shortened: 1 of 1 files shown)


cathyzhyi:
@0dvictor could you review this change? thanks!

runtime/compiler/x/codegen/J9TreeEvaluator.cpp
```diff
@@ -152,14 +152,20 @@ inline void generateLoadJ9Class(TR::Node* node, TR::Register* j9class, TR::Regis
                    "Unexpected opCode for generateLoadJ9Class %s.", node->getOpCode().getName());
          break;
       }
+   if (node->getOpCode().isReadBar() || node->getOpCode().isWrtBar())
```
0dvictor: It might be better to move this into `default` case of the switch statement, as checking whether the node is read/write barrier is unnecessary when its opcode falls into the other cases. In addition, the assertion can also be simplified.

runtime/compiler/x/codegen/J9TreeEvaluator.cpp
```diff
@@ -152,14 +152,20 @@ inline void generateLoadJ9Class(TR::Node* node, TR::Register* j9class, TR::Regis
...
    auto use64BitClasses = TR::Compiler->target.is64Bit() && !TR::Compiler->om.generateCompressedObjectHeaders();
    auto instr = generateRegMemInstruction(LRegMem(use64BitClasses), node, j9class, generateX86MemoryReference(object, TR::Compiler->om.offsetOfObjectVftField(), cg), cg);
    if (needsNULLCHK)
       {
       cg->setImplicitExceptionPoint(instr);
       instr->setNeedsGCMap(0xFF00FFFF);
-      instr->setNode(cg->comp()->findNullChkInfo(node));
       }
+
+   if (opValue == TR::checkcastAndNULLCHK)
```
0dvictor: We may also want to move this inside the earlier `if` statement. We don't need check the opcode when `needsNULLCHK` is false, since when `opValue == TR::checkcastAndNULLCHK` `needsNULLCHK` must be true.

runtime/compiler/x/codegen/J9TreeEvaluator.cpp
cathyzhyi: Changed according to comments.

runtime/compiler/x/codegen/J9TreeEvaluator.cpp
cathyzhyi: Changed according to comments.

runtime/compiler/x/codegen/J9TreeEvaluator.cpp
```diff
@@ -145,21 +145,28 @@ inline void generateLoadJ9Class(TR::Node* node, TR::Register* j9class, TR::Regis
       case TR::icall: // TR_checkAssignable
          return; // j9class register already holds j9class
       default:
+         if (node->getOpCode().isReadBar() || node->getOpCode().isWrtBar())
+            needsNULLCHK = true;
+
          TR_ASSERT(node->getOpCode().isReadBar() ||
```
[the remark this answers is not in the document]
cathyzhyi: changed the entire switch to an else statement to avoid calling `isReadBar()` and `isWrtBar()` twice.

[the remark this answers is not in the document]
0dvictor: Good idea.

andrewcraik:

andrewcraik:
Looks like infra issues with sanity - will retry now

andrewcraik:

cathyzhyi:
@andrewcraik Hey Andrew, could you help to merge this one? Thanks!

PR #4000, opened 2018-12-11
Free client specific data on VM shutdown

satbir-ca:
JITaaS server caches the information for the JITaaS client
in an internal hash map per client.This data is purged if
not used for sometime(1000 minutes) but this can grow very
large if the number of client increases to large number.
To fix this issue the client data needs to be deleted when
the client compilation thread teminates.

On VM shutdown, stopCompilationThreads get called which terminate
the client thread.On thread temination a dummy remote compilation
request is build with NULL J9Method. On seeing this request the server
check if the inUse for the client session is 0 and removes the client
session data from the hash map. If the inUse is not zero then the
the client data is marked to be deleted and the onus of deleting
the client data lies with whoever make the inUse 0.

### DIFF
--- a/runtime/compiler/control/JITaaSCompilationThread.cpp
+++ b/runtime/compiler/control/JITaaSCompilationThread.cpp
@@ -2428,6 +2428,7 @@ ClientSessionData::ClientSessionData(uint64_t clientUID, uint32_t seqNo) :
    _sequencingMonitor = TR::Monitor::create("JIT-JITaaSSequencingMonitor");
    _vmInfo = nullptr;
    _staticMapMonitor = TR::Monitor::create("JIT-JITaaSStaticMapMonitor");
+   _markedForDeletion = false;
    }
 
 ClientSessionData::~ClientSessionData()
@@ -2730,6 +2731,30 @@ ClientSessionHT::findOrCreateClientSession(uint64_t clientUID, uint32_t seqNo, b
          }
       }
    return clientData;
+   }
+// Search the clientSessionHashtable for the given clientUID and delete  the
+// data corresponding to the client. Return true if client data found and deleted.
+// Must have compilation monitor in hand when calling this function.
+
+bool
+ClientSessionHT::deleteClientSession(uint64_t clientUID, bool forDeletion)
+   {
+   ClientSessionData *clientData = nullptr;
+   auto clientDataIt = _clientSessionMap.find(clientUID);
+   if (clientDataIt != _clientSessionMap.end())
+      {
+      clientData = clientDataIt->second;
+      if (forDeletion)
+         clientData->markForDeletion();
+
+      if ((clientData->getInUse() == 0) && clientData->isMarkedForDeletion())
+         {
+         ClientSessionData::destroy(clientData); // delete the client data
+         _clientSessionMap.erase(clientDataIt); // delete the mapping from the hashtable
+         return true;
+         }
+      }
+   return false;
    }
 
 // Search the clientSessionHashtable for the given clientUID and return the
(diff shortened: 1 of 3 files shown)


satbir-ca:

runtime/compiler/control/JITaaSCompilationThread.cpp
```diff
@@ -3260,175 +3289,196 @@ TR::CompilationInfoPerThreadRemote::processEntry(TR_MethodToBeCompiled &entry, J
...
-      } // end critical section
-
-     
-      // We must process unloaded classes lists in the same order they were generated at the client
-      // Use a sequencing scheme to re-order compilation requests
-      //
-      clientSession->getSequencingMonitor()->enter();
-      clientSession->updateMaxReceivedSeqNo(seqNo);
-      if (seqNo > clientSession->getExpectedSeqNo()) // out of order messages
+      if (!ramMethod)
```
mpirvu: Just for aesthetics could we leave the string of std::get<> uninterrupted by this `if` statement?

runtime/compiler/control/JITaaSCompilationThread.cpp
```diff
@@ -3260,175 +3289,196 @@ TR::CompilationInfoPerThreadRemote::processEntry(TR_MethodToBeCompiled &entry, J
...
+            {
+            if (TR::Options::getVerboseOption(TR_VerboseJITaaS))
+               TR_VerboseLog::writeLineLocked(TR_Vlog_JITaaS,"client (%lu) not deleted", clientId);
+            }
+         else
+            {
+            if (TR::Options::getVerboseOption(TR_VerboseJITaaS))
+               TR_VerboseLog::writeLineLocked(TR_Vlog_JITaaS,"client (%lu) deleted", clientId);
+            }
+         noCompilation = true;   // This is a null message for termination so no compilation needed.
```
mpirvu: Suggestion: maybe a simpler solution is to set `abortCompilation = true` and let the code take that path.

runtime/compiler/control/JITaaSCompilationThread.cpp
```diff
@@ -2731,6 +2732,32 @@ ClientSessionHT::findOrCreateClientSession(uint64_t clientUID, uint32_t seqNo, b
...
+   auto clientDataIt = _clientSessionMap.find(clientUID);
+   if (clientDataIt != _clientSessionMap.end())
+      {
+      clientData = clientDataIt->second;
+      if (forDeletion)
+         clientData->markedForDeletion();
+
+      if ((clientData->getInUse() == 0) && clientData->isMarkedForDeletion())
+         {
+         // if clientData found in hashtable, update the access time before returning it
```
mpirvu: This comment needs to be deleted.

runtime/compiler/control/JITaaSCompilationThread.hpp
```diff
@@ -135,6 +135,9 @@ class ClientSessionData
    void decNumActiveThreads() { --_numActiveThreads;  }
    void printStats();
 
+   void markedForDeletion() { _markedForDeletion = true;}
```
mpirvu: Please change the name to setMarkedForDeletion() or markForDeletion(). The current name suggests a query.

runtime/compiler/control/JITaaSCompilationThread.hpp
```diff
@@ -135,6 +135,9 @@ class ClientSessionData
...
+   bool isMarkedForDeletion() { return _markedForDeletion; }
```
mpirvu: Please make this a const member function.

runtime/compiler/control/JITaaSCompilationThread.cpp
mpirvu: Maybe it's not working only in this context. I have seen compilations that take this path and everything is well. Anyway, if there is something wrong with it we need to fix it.

mpirvu: The way the code stands now we are leaking compilation entries because `compInfo->recycleCompilationEntry(&entry);` is not called.

PR #4002, opened 2018-12-11
Add extra dependency to CHelper to preserve HPRs

NigelYiboYu:
Fix a bug where HPRs are not preserved across calls built by CHelper
linkage.

All HPRs are volatiles for 32 bit SDKs. If 64-bit registers can
be used in 32-bit SDKs, the RA should be informed of 64-bit registers
that are in use and preserve them across native calls. This is achieved
by adding a KillVolHighRegs to the post dependency group of a CHelper
call.


NigelYiboYu:
Marking as WIP until fully tested.

fjeremic:

PR #4006, opened 2018-12-12
Remove obsolete jcl MAX link

JasonFengJ9:
Remove obsolete `jcl MAX` link

`jcl Max` configuration is obsoleted and the link is invalid.
Also removed a few other internal links.
Note: this was discovered while eyeballing JVM flags.

Reviewer: @DanHeidinga 

### DIFF
--- a/runtime/bcutil/ROMClassWriter.cpp
+++ b/runtime/bcutil/ROMClassWriter.cpp
@@ -901,9 +901,19 @@ private:
 			 * (i.e., number of dimensions of the array - 1)
 			 * in the next 2 bytes in Big Endian (since we are maintaining Sun StackMapTable format).
 			 *
-			 * See: https://jtcjazz.ottawa.ibm.com:9443/jazz/resource/itemName/com.ibm.team.workitem.WorkItem/18860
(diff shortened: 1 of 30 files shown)


runtime/bcutil/ROMClassWriter.cpp
```diff
@@ -900,10 +900,7 @@ class ROMClassWriter::Helper :
 			/* Also, encode the arity beyond 1
 			 * (i.e., number of dimensions of the array - 1)
 			 * in the next 2 bytes in Big Endian (since we are maintaining Sun StackMapTable format).
-			 *
-			 * See: https://jtcjazz.ottawa.ibm.com:9443/jazz/resource/itemName/com.ibm.team.workitem.WorkItem/18860
-			 * for explanation of why arity -1 is encoded instead of arity.
-			 *  */
```
DanHeidinga: Can you add the explanation from the internal item?  Better to have the details here in the code rather than losing the link to them.

JasonFengJ9: Sure, updated the explanation from the internal item.
@DanHeidinga please have another look.

DanHeidinga:

PR #4009, opened 2018-12-12
Guarantee functional correctness of rdbar/wrtbar

cathyzhyi:
This change guarantees the correctness of rdbar/wrtbar from the following
aspects:
1. Consider rdbar/wrtbar as GC points
2. Make optimizations consider shape of rdbar/wrtbar when doing
transformations
3. Disable opts not supporting newly added rdbar/wrtbar when seeing
the opcodes

### DIFF
--- a/runtime/compiler/codegen/J9CodeGenerator.cpp
+++ b/runtime/compiler/codegen/J9CodeGenerator.cpp
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2000, 2018 IBM Corp. and others
+ * Copyright (c) 2000, 2019 IBM Corp. and others
  *
  * This program and the accompanying materials are made available under
  * the terms of the Eclipse Public License 2.0 which accompanies this
@@ -351,7 +351,7 @@ J9::CodeGenerator::lowerCompressedRefs(
 
       // base object
       address = loadOrStoreNode->getFirstChild();
-      loadOrStoreOp = TR::Compiler->om.shouldGenerateReadBarriersForFieldLoads() ? self()->comp()->il.opCodeForIndirectReadBarrier(TR::Int32) :
+      loadOrStoreOp = TR::Compiler->om.shouldGenerateReadBarriersForFieldLoads() || loadOrStoreNode->getOpCode().isReadBar() ? self()->comp()->il.opCodeForIndirectReadBarrier(TR::Int32) :
                                                                                    self()->comp()->il.opCodeForIndirectLoad(TR::Int32);
       }
    else if ((loadOrStoreNode->getOpCode().isStoreIndirect() ||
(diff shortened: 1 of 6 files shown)


runtime/compiler/compile/J9Compilation.hpp
```diff
@@ -311,6 +311,8 @@ class OMR_EXTENSIBLE Compilation : public OMR::CompilationConnector
 
    TR_RelocationRuntime *reloRuntime() { return _reloRuntime; }
 
+   bool incompleteOptimizerSupportForReadWriteBarriers();
```
fjeremic: Is this API really necessary to pollute the compilation object? I see only one use of it and it is very specific.

cathyzhyi: The other optimization using the query are in omr level https://github.com/eclipse/omr/pull/3224.

fjeremic: Ah right, I had not considered that. Thanks!

runtime/compiler/codegen/J9CodeGenerator.cpp
```diff
@@ -2925,16 +2925,17 @@ J9::CodeGenerator::compressedReferenceRematerialization()
...
+   bool generateReadBarrier = self()->comp()->getOption(TR_EnableFieldWatch);
+
+   // The compressedrefs remat opt removes decompression/compression sequences from
+   // loads/stores where there doesn't exist a gc point between the load and the store,
+   // and the load doesn't need to be dereferenced.
+   // The opt needs to be disabled for the following cases:
+   // 1. In Guarded Storage, we can't not do a guarded load because the object that is loaded may
+   // not be in the root set, and as a consequence, may get moved.
+   // 2. For read barriers, the vmhelpers are GC points and therefore the object might be moved
+   if (TR::Compiler->om.shouldGenerateReadBarriersForFieldLoads() || generateReadBarrier)
```
andrewcraik: why isn't the first part of this || part of the bool added above? it would seem you would have to repeat this idiom no? Doesn't this warrant a trace statement of some kind to say we disabling because of this?

runtime/compiler/optimizer/IdiomRecognition.cpp
```diff
@@ -2302,6 +2302,9 @@ TR_CISCNode *
 TR_CISCTransformer::addAllSubNodes(TR_CISCGraph *const graph, TR::Block *const block, TR::TreeTop *const top,
                                    TR::Node *const parent, TR::Node *const node, const int32_t dagId)
    {
+   //IdiomRecognition doesn't know how to handle rdbar/wrtbar for now
+   if ((node->getOpCode().isReadBar() || node->getOpCode().isWrtBar()) && comp()->incompleteOptimizerSupportForReadWriteBarriers())
```
andrewcraik: can we reverse the order of the && to make this more clear and break onto a new line at the && so it looks more like
```if (comp()->incompleteOptimizerSupportForReadWriteBarriers()
       && (node->getOpCode().isReadBar() || node->getOpCode().isWrtBar()))

andrewcraik:
Once the final code formatting issue is sorted I think this will be good to test and merge.

runtime/compiler/optimizer/IdiomRecognition.cpp
cathyzhyi: changed according to the comment. Thanks!

andrewcraik:
Jenkins test sanity xlinux,win,plinux jdk8.jdk11

andrewcraik:

andrewcraik:
Failure on x86-64 is DDR related and unrelated to this change- merging

PR #4010, opened 2018-12-12
Generate readbar and write bar when ilgen

cathyzhyi:
Generate rdbar/wrtbar during ilgen for static and instance field load
and store. The rdbar must hang under a treetop which is different from a
normal read.

Refer to https://github.com/eclipse/omr/issues/2812 and https://github.com/eclipse/omr/issues/2948 for syntax of rdbar/wrtbar opcodes


runtime/compiler/ilgen/Walker.cpp
```diff
@@ -7219,9 +7229,9 @@ TR_J9ByteCodeIlGenerator::storeInstance(int32_t cpIndex)
    // code to handle volatiles moved to CodeGenPrep
    //
    TR::Node * node;
-   if (type == TR::Address && _generateWriteBarriers)
+   if ((type == TR::Address && _generateWriteBarriers) || comp()->getOption(TR_EnableFieldWatch))
```
fjeremic: We seem to be orthogonal here w.r.t. write and read barriers. Write barriers are controlled via `_generateWriteBarriers` which is initialized once in the constructor while we check for read barriers via `TR_EnableFieldWatch` every time. Not only this but we seem to store the result of the `getOption` call in a `generateReadBarrier` variable above, but here we query it directly.

I'm not a fan of this inconsistency as it makes the code harder to read and follow. Can we stick to one approach? Are we able to introduce a `_generateReadBarriers` field analogous to `_generateWriteBarriers`?

cathyzhyi: The other `_generateWriteBarriers` is for `comp->getOptions()->getGcMode() != TR_WrtbarNone` and wrtbar are only needed for address type in this case so one bool field for all wrtbars won't work.  I can change it to  `_generateWriteBarriersForGC` to be more clear and create another field called `_generateReadBarrierForFieldWatch` and `_generateWriteBarrierForFieldWatch` and all of them got initialized once at constructor.

fjeremic: Sounds like a reasonable, and much more consistent approach.

cathyzhyi: changed the existing field to be  `_generateWriteBarriersForGC` 
 and  added `_generateReadBarrierForFieldWatch` `_generateWriteBarrierForFieldWatch`

andrewcraik:
@cathyzhyi most of these changes won't be tested by sanity - what testing have you done. I will run sanity to make sure the case that does not generate the readbars and writebars still works fine.

andrewcraik:
Jenkins test sanity xlinux,win,plinux jdk11

runtime/compiler/ilgen/Walker.cpp
```diff
@@ -5909,7 +5910,7 @@ TR_J9ByteCodeIlGenerator::loadInstance(int32_t cpIndex)
       }
    else if (!address->isNonNull())
       treeTopNode = genNullCheck(load);
```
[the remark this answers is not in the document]
cathyzhyi: NULLCHK of readbar doesn't need special handling. The child of NULLCHK is still the rdbar and NULLCHK by default check the grandchild of its child.  This is the same as existing NULLCHK on indirect stores.

cathyzhyi:
> @cathyzhyi most of these changes won't be tested by sanity - what testing have you done. I will run sanity to make sure the case that does not generate the readbars and writebars still works fine.

I've run personal builds with forcing generating rdbar/wrtbar and some large benchmarks with field watch enabled but this change itself will not generaet rdbar/wrtbar for field watch yet because `TR_EnableFieldWatch` is not set.

andrewcraik:
ok - the code seems reasonable, has had separate testing, and full testing in OpenJ9 can be completed as part of the enablement PR which I assume will follow shortly.

PR #4011, opened 2018-12-12
Pass isOptServer value to the server

dmitry-ten:
`_isOptServer` can be set either by `TR_Server` option, or it
is set automatically, when the number of loaded classes exceeds
a certain threshold. However, for JITaaS client, automatic set
will not happen because it doesn't know the number of classes loaded
on the client.
To fix this limitation, check whether `_isOptServer` is set on the client,
before sending compilation request, and if it is set, set `TR_Server` option,
so that it gets passed to a server.


dmitry-ten:
@mpirvu please review.

PR #4016, opened 2018-12-12
Enable TGC root scan times for Metronome

amicic:
Reduce the prints to only threads/increments that really have non-0
values to report (note that most of increments in Metronome do not deal
with roots)

Since scan time for one entity can span over several increments, new API
is introduced for suspending/resuming within one entity, to properly
include only GC increment time (and exclude mutator times in between).

Each time GC yields to mutator or blocks a thread on synchronization
barrier, suspend API is invoked, and conversely after awakening on any
of those two points resume API is invoked.

The API will maintain start/end event of the increment, and add the diff
(duration of increment) to the total time of entity scan time. Also
maximum increment of any entity for each threads is maintained and
reported by TGC.

A yield point in incremental roots is moved up to scan class entity(s).
Now, all explicit yield points are within a root scan entity, which
makes life simpler for reporting purposes.

An example of TGC report with two new max increment fields (expectedly,
time is around or bellow 3ms):

```
<scan timestamp="Dec 12 21:17:40 2018">
        <thread id="0" ownablesynchronizerobjects="2.240"
stringtable="0.109" maxincrementtime="2.240"
maxincremententity="ownablesynchronizerobjects"/>
        <thread id="1" ownablesynchronizerobjects="2.151"
stringtable="0.094" maxincrementtime="2.150"
maxincremententity="ownablesynchronizerobjects"/>
        <thread id="2" ownablesynchronizerobjects="2.154"
stringtable="0.090" monitorrefs="6.525"
monitorreferenceobjectscomplete="0.001" maxincrementtime="2.964"
maxincremententity="monitorrefs"/>
        <total ownablesynchronizerobjects="6.545" stringtable="0.293"
monitorrefs="6.525" monitorreferenceobjectscomplete="0.001"/>
</scan>
```

### DIFF
--- a/runtime/gc_realtime/Scheduler.cpp
+++ b/runtime/gc_realtime/Scheduler.cpp
@@ -712,7 +712,11 @@ MM_Scheduler::condYieldFromGC(MM_EnvironmentBase *envBase, U_64 timeSlack)
 	if (!internalShouldGCYield(env, timeSlack)) {
 		return false;
 	}
+
+	env->reportScanningSuspended();
 	yieldFromGC(env, true);
+	env->reportScanningResumed();
+
 	env->resetCurrentDistanceToYieldTimeCheck();
 
 	return true;
--- a/runtime/gc_base/RootScanner.cpp
+++ b/runtime/gc_base/RootScanner.cpp
@@ -1,6 +1,6 @@
 
 /*******************************************************************************
- * Copyright (c) 1991, 2017 IBM Corp. and others
+ * Copyright (c) 1991, 2018 IBM Corp. and others
  *
  * This program and the accompanying materials are made available under
  * the terms of the Eclipse Public License 2.0 which accompanies this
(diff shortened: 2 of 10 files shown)


amicic:
dependant on https://github.com/eclipse/omr/pull/3331

dmitripivkine:
Jenkins test sanity xLinux jdk8 depends https://github.com/eclipse/omr/pull/3331

charliegracie:
Jenkins test sanity xLinux jdk8 depends eclipse/omr#3331

amicic:
more code added, please re-review.

runtime/gc_realtime/Scheduler.cpp
```diff
@@ -45,6 +45,7 @@
 #include "Metronome.hpp"
 #include "MetronomeAlarmThread.hpp"
 #include "RealtimeGC.hpp"
+#include "RealtimeRootScanner.hpp"
```
dmitripivkine: Do we need this include?

amicic: removed

dmitripivkine:

PR #4023, opened 2018-12-12
Fix GPU sorting of large arrays

keithc-ca:
PTX code was generated for compute capability 2.0, yet the grid dimensions were computed assuming the kernel supported compute capability 3.0 resulting in an exception:

> com.ibm.gpu.GPUSortException: invalid argument

This changes the generator to produce compute capability 3.0 when appropriate.

The test was adjusted to include a size that would trigger the above exception (before this fix).


pshipton:
jenkins compile xlinux jdk8

PR #4025, opened 2018-12-12
Change cache directory default permission on OpenJ9

hangshao0:
1. Use 700 on new cache directory created without "groupAccess"
2. Use 750 on new cache directory created with "groupAcess"
3. Do not change the permission of existing cache directory on OpenJ9.
4. Update the testCacheDirPerm

This change has no effect if "cacheDirPerm=" presents.


runtime/port/sysvipc/j9sharedhelper.c
```diff
@@ -139,7 +139,7 @@ createDirectory(struct J9PortLibrary *portLibrary, char *pathname, uintptr_t per
 		
 	current = strchr(tempPath+1, DIR_SEPARATOR); /* skip the first '/' */
 
-	if (J9SH_DIRPERM_ABSENT == permission) {
+	if (J9SH_DIRPERM_ABSENT == permission || J9SH_DIRPERM_ABSENT_GROUPACCESS == permission) {
```
DanHeidinga: formatting - please enclose each condition in its own parenthesis so that readers don't have to think about order of operations.

runtime/port/sysvipc/j9shmem.c
```diff
@@ -1348,10 +1348,14 @@ j9shmem_createDir(struct J9PortLibrary* portLibrary, char* cacheDirName, uintptr
 		Trc_PRT_j9shmem_createDir_ExitFailedFile();
 		break;
 	case EsIsDir:
-#if !defined(J9ZOS390)
-		if (J9SH_DIRPERM_ABSENT == cacheDirPerm) {
+#if !defined(J9ZOS390) && !defined(OPENJ9_BUILD)
+		if (J9SH_DIRPERM_ABSENT == cacheDirPerm || J9SH_DIRPERM_ABSENT_GROUPACCESS == cacheDirPerm) {
```
DanHeidinga: Same thing here - formatting

runtime/port/sysvipc/j9shmem.c
```diff
@@ -1348,10 +1348,14 @@ j9shmem_createDir(struct J9PortLibrary* portLibrary, char* cacheDirName, uintptr
...
 			/* cacheDirPerm option is not specified. Change permission to J9SH_DIRPERM. */
```
DanHeidinga: comment needs to be updated to match the new code

runtime/port/sysvipc/j9shmem.c
```diff
@@ -1348,10 +1348,14 @@ j9shmem_createDir(struct J9PortLibrary* portLibrary, char* cacheDirName, uintptr
 		Trc_PRT_j9shmem_createDir_ExitFailedFile();
 		break;
 	case EsIsDir:
-#if !defined(J9ZOS390)
-		if (J9SH_DIRPERM_ABSENT == cacheDirPerm) {
+#if !defined(J9ZOS390) && !defined(OPENJ9_BUILD)
```
DanHeidinga: Why are we introducing new `!OPENJ9_BUILD` code here?  Why wouldn't OpenJ9 update the directory permissions in this case?

runtime/port/sysvipc/j9sharedhelper.c
hangshao0: Fixed

runtime/port/sysvipc/j9shmem.c
hangshao0: Fixed

runtime/port/sysvipc/j9shmem.c
hangshao0: Fixed

runtime/port/sysvipc/j9shmem.c
hangshao0: If user created a directory and set permission for that directory. I think we shouldn't overwrite the permission in this case.

DanHeidinga: I guess my question is - why is this behaviour different between builds that define OPENJ9_BUILD and anything else (IBM Java?)?

hangshao0: I am fine with keeping the behavior here consistent regardless of OPENJ9_BUILD. 
Then we need to choose whether we want to overwrite the user's permission setting on an existing directory. I preference is not to overwrite.

hangshao0: Code updated to make the behavior consistent on existing directory.

runtime/port/sysvipc/j9shmem.c
```diff
@@ -1348,23 +1348,8 @@ j9shmem_createDir(struct J9PortLibrary* portLibrary, char* cacheDirName, uintptr
 		Trc_PRT_j9shmem_createDir_ExitFailedFile();
 		break;
 	case EsIsDir:
-#if !defined(J9ZOS390)
-		if (J9SH_DIRPERM_ABSENT == cacheDirPerm) {
-			/* cacheDirPerm option is not specified. Change permission to J9SH_DIRPERM. */
-			rc2 = changeDirectoryPermission(portLibrary, cacheDirName, J9SH_DIRPERM);
```
pshipton: I think this behavior is still valid, but perhaps only when using a default /tmp/javasharedresources directory. I don't think it should be removed completely. Otherwise it could stop other users on the machine from creating or using caches.

runtime/oti/j9port.h
```diff
@@ -177,7 +177,15 @@
 #define J9SH_SEMAPHORE_ID "_semaphore_"
 
 #define J9SH_DIRPERM_ABSENT ((uintptr_t)-2)
+#define J9SH_DIRPERM_ABSENT_GROUPACCESS ((uintptr_t)-3)
+#if defined(OPENJ9_BUILD)
+#define J9SH_DIRPERM (0700)
+#define J9SH_DIRPERM_GROUPACCESS (0750)
+#else
 #define J9SH_DIRPERM (0777)
```
pshipton: Like the changeDirectoryPermission() comment, I'm thinking 777 should only be used when using a default /tmp/javasharedresources directory, anything else can use 700. Similarly for group access.

pshipton:
I added the depends:doc tag since the changes in behavior will require documentation.

hangshao0:
This change can be reviewed again.

pshipton:
jenkins test all zlinux jdk8

runtime/port/sysvipc/j9shmem.c
```diff
@@ -1339,8 +1339,16 @@ j9shmem_createDir(struct J9PortLibrary* portLibrary, char* cacheDirName, uintptr
 	OMRPORT_ACCESS_FROM_J9PORT(portLibrary);
 	intptr_t rc, rc2;
 	char pathCopy[J9SH_MAXPATH];
+	char defaultTmpDir[J9SH_MAXPATH];
```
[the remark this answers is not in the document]
hangshao0: Renamed pathCopy and use it as the buffer.

runtime/port/sysvipc/j9shmem.c
```diff
@@ -1339,8 +1339,16 @@ j9shmem_createDir(struct J9PortLibrary* portLibrary, char* cacheDirName, uintptr
...
+	BOOLEAN usingDefaultTmp = FALSE;
 
 	Trc_PRT_j9shmem_createDir_Entry();
+	memset(defaultTmpDir, 0, J9SH_MAXPATH);
```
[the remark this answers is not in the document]
hangshao0: Removed memset here.

pshipton:

PR #4027, opened 2018-12-13
Add static symbol for constant pool address

cathyzhyi:
Create a static symbol to represent the constant pool address for a
resolved method so that a loadaddr node can be created when the CP
address is needed in trees and AOT can relocate the address value
correctly.


dsouzai:
> AOT can relocate the address value correctly.

I can't tell from just this commit whether that's true. Is there code somewhere already where you check that this `loadaddr` is for a CP Address and generate the appropriate relocation for it?

cathyzhyi:
> I can't tell from just this commit whether that's true. Is there code somewhere already where you check that this loadaddr is for a CP Address and generate the appropriate relocation for it?

the check for CP Address is in this PR https://github.com/eclipse/omr/pull/3334

PR #4047, opened 2018-12-13
Fix anewarray AOT bug

dsouzai:
When the compiler encounters the TR::anewarray op code, it will ignore
the class register when calling genInitArrayHeader in favour of
determining the array class itself. However, during an AOT compile, it
does use the class register; this is because when not using the SVM, the
compiler would generate code to acquire the array class from the
component class. It did so because prior to the SVM, there wasn't a way
to materialize the array class without creating a specialized relocation
record.

With the SVM, the compiler does not need to generate the code to acquire
the array class from the component class, and thus needs to also go
through the code path that ignores the class register.


dsouzai:
@jdmpapin @andrewcraik could you please review?

andrewcraik:
This seems reasonable, but I defer to @jdmpapin for a review since I am less familiar with the SVM. I will launch sanity now in the hope we can merge this since the change seems correct and trivial.

andrewcraik:

PR #4049, opened 2018-12-13
WIP: Fix bug in Array Store Check

dsouzai:
This PR fixes a bug in the Array Store Check evaluator that should have been comparing the component class of the array class with `java/lang/Object`, but instead was incorrectly doing the comparison with the array class, which means the test will never succeed. AOT support is also added.

### DIFF
--- a/runtime/compiler/x/codegen/J9TreeEvaluator.cpp
+++ b/runtime/compiler/x/codegen/J9TreeEvaluator.cpp
@@ -7908,27 +7908,47 @@ J9::X86::TreeEvaluator::VMarrayStoreCHKEvaluator(
       // -------------------------------------------------------------------------
       //
       // If the component type is java.lang.Object then the store always succeeds.
+      // On AOT, only do this when using the SVM; otherwise, it's safer to just
+      // skip this check than 1) incorrectly relocate the java/lang/Object class
+      // or 2) crash while trying to relocate.
       //
       // -------------------------------------------------------------------------
-
-      TR_OpaqueClassBlock *objectClass = fej9->getSystemClassFromClassName("java/lang/Object", 16);
-      objectClass = fej9->getArrayClassFromComponentClass(objectClass);
-
-      if (TR::Compiler->target.is64Bit())
+      if (!cg->needClassAndMethodPointerRelocations() || comp->getOption(TR_UseSymbolValidationManager))
          {
-         TR_ASSERT((((uintptr_t)objectClass) >> 32) == 0, "TR_OpaqueClassBlock must fit on 32 bits when using class pointer compression");
-         instr = generateRegImmInstruction(CMP4RegImm4, node, destComponentClassReg, (uint32_t) ((uint64_t) objectClass), cg);
-         }
-      else
-         {
-         instr = generateRegImmInstruction(CMP4RegImm4, node, destComponentClassReg, (int32_t)(uintptr_t) objectClass, cg);
-         }
+         TR_OpaqueClassBlock *objectClass = fej9->getSystemClassFromClassName("java/lang/Object", 16);
+         objectClass = fej9->getArrayClassFromComponentClass(objectClass);
 
-      generateLabelInstruction(JE4, node, wrtbarLabel, cg);
+         if (TR::Compiler->target.is64Bit())
+            {
+            TR_ASSERT((((uintptr_t)objectClass) >> 32) == 0, "TR_OpaqueClassBlock must fit on 32 bits when using class pointer compression");
 
-      // HCR in VMarrayStoreCHKEvaluator
-      if (cg->wantToPatchClassPointer(objectClass, node))
-         comp->getStaticHCRPICSites()->push_front(instr);
+            if (cg->needClassAndMethodPointerRelocations())
+               {
+               TR::Register *objectClassReg = scratchRegisterManager->findOrCreateScratchRegister();
+               // MOV8RegImm64 because the relocation infrastructure can't currently relocate a 32-bit pointer
+               instr = generateRegImm64Instruction(MOV8RegImm64, node, objectClassReg, (uintptr_t) objectClass, cg, TR_ClassPointer);
+               generateRegRegInstruction(CMP4RegReg, node, destComponentClassReg, objectClassReg, cg);
+               scratchRegisterManager->reclaimScratchRegister(objectClassReg);
+               }
+            else
+               {
+               instr = generateRegImmInstruction(CMP4RegImm4, node, destComponentClassReg, (uint32_t) ((uint64_t) objectClass), cg);
+               }
+            }
+         else
+            {
+            if (cg->needClassAndMethodPointerRelocations())
+               instr = generateRegImmInstruction(CMP4RegImm4, node, destComponentClassReg, (int32_t)(uintptr_t) objectClass, cg, TR_ClassPointer);
+            else
+               instr = generateRegImmInstruction(CMP4RegImm4, node, destComponentClassReg, (int32_t)(uintptr_t) objectClass, cg);
+            }
+
+         generateLabelInstruction(JE4, node, wrtbarLabel, cg);
+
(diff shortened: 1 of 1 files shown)


dsouzai:
WIP because I need to run more tests on this.

dsouzai:
@jdmpapin @0dvictor could you please review?

fjeremic:
> This PR fixes a bug in the Array Store Check evaluator that should have been comparing the component class of the array class with `java/lang/Object`, but instead was incorrectly doing the comparison with the array class, which means the test will never succeed. AOT support is also added.

Not sure I'm understanding this correctly. What we were doing before sounds functionally incorrect. Or is there a fallback path? Can you elaborate?

dsouzai:
This is what was generated before:
```
   0x7fffd542fc5d:      mov    edi,DWORD PTR [rcx]
   0x7fffd542fc5f:      and    rdi,0xffffffffffffff00
   0x7fffd542fc66:      cmp    edi,0xbfea00
   0x7fffd542fc6c:      je     0x7fffd542fca9
   0x7fffd542fc6e:      mov    rdi,QWORD PTR [rdi+0x58]
   0x7fffd542fc72:      mov    r8d,DWORD PTR [r13+0x0]
   0x7fffd542fc76:      and    r8,0xffffffffffffff00
   0x7fffd542fc7d:      cmp    edi,r8d
   0x7fffd542fc80:      je     0x7fffd542fca9
```
after
```
   0x7fffd542fc5d:      mov    edi,DWORD PTR [rcx]
```
`edi` contains the array class who's component class we want to compare against `java/lang/Object`. `0xbfea00` is the `J9Class *` of  `java/lang/Object`. Thus, this will never succeed. 

```
   0x7fffd542fc6e:      mov    rdi,QWORD PTR [rdi+0x58]
```
is where we grab the component class from the array class.

fjeremic:
Ok, so this is a performance issue. i.e. the first four instructions in the above seuqnece are useless because the `cmp` at `0x7fffd542fc66` will never be true and the subsequent `je` will never be taken, so these instructions can be omitted altogether. Is this correct?

dsouzai:
> so these instructions can be omitted altogether. Is this correct?

Not quite; the idea for those instructions is to eliminate the store check if the component type is `java/lang/Object`. Therefore, the fix that this PR implements is to move 

```
   0x7fffd542fc66:      cmp    edi,0xbfea00
   0x7fffd542fc6c:      je     0x7fffd542fca9
```
after
```
   0x7fffd542fc6e:      mov    rdi,QWORD PTR [rdi+0x58]
```
so that the `cmp    edi,0xbfea00` instruction now compares the component class with `0xbfea00`.

jdmpapin:
I think it would be better (and I suspect it was the original intention) to compare the array type to `Object[]` instead of comparing the component type to `Object`

fjeremic:
> herefore, the fix that this PR implements is to move
> 
> ```
>    0x7fffd542fc66:      cmp    edi,0xbfea00
>    0x7fffd542fc6c:      je     0x7fffd542fca9
> ```
> 
> after
> 
> ```
>    0x7fffd542fc6e:      mov    rdi,QWORD PTR [rdi+0x58]
> ```

Thanks. This makes sense now.

> I think it would be better (and I suspect it was the original intention) to compare the array type to `Object[]` instead of comparing the component type to `Object`

This would make most sense to me. @dsouzai is there a reason we aren't doing this instead?

dsouzai:
>This would make most sense to me. @dsouzai is there a reason we aren't doing this instead?

No reason, just lack of expertise in this part of the code :). I'll make the requested changes.

dsouzai:
Changes made; @jdmpapin @0dvictor could you guys please review?

runtime/compiler/x/codegen/J9TreeEvaluator.cpp
```diff
@@ -7908,42 +7908,47 @@ J9::X86::TreeEvaluator::VMarrayStoreCHKEvaluator(
...
-      TR_OpaqueClassBlock *objectClass = fej9->getSystemClassFromClassName("java/lang/Object", 16);
-
-      if (TR::Compiler->target.is64Bit())
+      if (!cg->needClassAndMethodPointerRelocations() || comp->getOption(TR_UseSymbolValidationManager))
          {
-#ifdef J9VM_INTERP_COMPRESSED_OBJECT_HEADER
-         TR_ASSERT((((uintptr_t)objectClass) >> 32) == 0, "TR_OpaqueClassBlock must fit on 32 bits when using class pointer compression");
-         instr = generateRegImmInstruction(CMP4RegImm4, node, destComponentClassReg, (uint32_t) ((uint64_t) objectClass), cg);
+         TR_OpaqueClassBlock *objectClass = fej9->getSystemClassFromClassName("java/lang/Object", 16);
+         objectClass = fej9->getArrayClassFromComponentClass(objectClass);
```
jdmpapin: Is it possible to fail to find the array class here?

jdmpapin: Should `objectClass` (and `objectClassReg` below) perhaps be renamed now that it isn't `Object` but `Object[]`?

I realize that `destComponentClassReg` is similarly misleading at this point, but in the following code the dest array class in it will be overwritten with the component class.

runtime/compiler/x/codegen/J9TreeEvaluator.cpp
```diff
@@ -8006,51 +8011,6 @@ J9::X86::TreeEvaluator::VMarrayStoreCHKEvaluator(
...
-         {
-         instr = generateRegImmInstruction(CMP4RegImm4, node, destComponentClassReg, (int32_t)(uintptr_t) objectClass, cg);
-         }
-
-      generateLabelInstruction(JE4, node, wrtbarLabel, cg);
-
-      // HCR in VMarrayStoreCHKEvaluator
-      if (cg->wantToPatchClassPointer(objectClass, node))
-         comp->getStaticHCRPICSites()->push_front(instr);
-      */
```
jdmpapin: :+1:

aviansie-ben:
Regarding this issue under Power: it looks like we don't have this particular problem. We did, however, seem to be adding this check under old AOT without adding a relocation or doing anything to fix the address of `j/l/Object`, which will need to be fixed. We also do the comparison after loading the component type rather than simply comparing the array type, which I'll be changing to match the behaviour you've used here as it should be more efficient.

runtime/compiler/x/codegen/J9TreeEvaluator.cpp
```diff
@@ -7908,27 +7908,47 @@ J9::X86::TreeEvaluator::VMarrayStoreCHKEvaluator(
...
-         {
-         instr = generateRegImmInstruction(CMP4RegImm4, node, destComponentClassReg, (int32_t)(uintptr_t) objectClass, cg);
-         }
+            if (cg->needClassAndMethodPointerRelocations())
+               {
+               TR::Register *objectClassReg = scratchRegisterManager->findOrCreateScratchRegister();
+               // MOV8RegImm64 because the relocation infrastructure can't currently relocate a 32-bit pointer
+               instr = generateRegImm64Instruction(MOV8RegImm64, node, objectClassReg, (uintptr_t) objectClass, cg, TR_ClassPointer);
+               generateRegRegInstruction(CMP4RegReg, node, destComponentClassReg, objectClassReg, cg);
+               scratchRegisterManager->reclaimScratchRegister(objectClassReg);
```
0dvictor: Alternatively, we can store `objectClass` in a data snippet and use `CMP4RegMem` so that the scratch register can be avoided.

runtime/compiler/x/codegen/J9TreeEvaluator.cpp
aviansie-ben: I tried to do the same thing on the Power codegen and, as it turns out, it _is_ possible to fail to find the class for `Object[]`...

genie-openj9:

PR #4061, opened 2018-12-14
Better code sharing for TGC root scan times

amicic:
Implicitly, also better handling of timestamp overflow.


dmitripivkine:

PR #4068, opened 2018-12-15
WIP: Perform AOT compilation at cheap warm level

mpirvu:
This commit will force AOT compilations to be performed at a "cheap"
warm level. A "cheap" warm level is an optimization level where the
inliner has been subdued somewhat (the same inliner changes are applied
to the Xtune:virtualized mode).
This change can be disabled by using -Xaot:disableAOTAtCheapWarm

### DIFF
--- a/runtime/compiler/control/CompilationThread.cpp
+++ b/runtime/compiler/control/CompilationThread.cpp
@@ -498,7 +498,7 @@ bool TR::CompilationInfo::shouldDowngradeCompReq(TR_MethodToBeCompiled *entry)
                  getMethodQueueSize() >= TR::Options::_qszThresholdToDowngradeOptLevelDuringStartup) ||
                  // Downgrade if AOT and startup, 
                 (TR::Options::getCmdLineOptions()->sharedClassCache() &&
-                 _jitConfig->javaVM->phase == J9VM_PHASE_STARTUP &&
+                 _jitConfig->javaVM->phase != J9VM_PHASE_NOT_STARTUP &&
                  !TR::Options::getCmdLineOptions()->getOption(TR_DisableDowngradeToColdOnVMPhaseStartup))
                )
                {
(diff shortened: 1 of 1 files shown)


mpirvu:
Jenkins copyright check

mpirvu:
Jenkins line endings check

mpirvu:
Jenkins test sanity all jdk8

runtime/compiler/control/CompilationThread.cpp
```diff
@@ -498,7 +498,7 @@ bool TR::CompilationInfo::shouldDowngradeCompReq(TR_MethodToBeCompiled *entry)
                  getMethodQueueSize() >= TR::Options::_qszThresholdToDowngradeOptLevelDuringStartup) ||
                  // Downgrade if AOT and startup, 
                 (TR::Options::getCmdLineOptions()->sharedClassCache() &&
-                 _jitConfig->javaVM->phase == J9VM_PHASE_STARTUP &&
```
vijaysun-omr: I understand that we don't want to downgrade during startup as per the intent of this change. 
I'm curious why we changed the condition such that it would downgrade when we are not in startup though. 
i.e. why do we want to change anything if we are not in startup phase ?

vijaysun-omr: What is the difference between J9VM_PHASE_STARTUP and !J9VM_PHASE_NOT_STARTUP ?

runtime/compiler/control/CompilationThread.cpp
```diff
@@ -7058,6 +7043,22 @@ TR::CompilationInfoPerThreadBase::wrappedCompile(J9PortLibrary *portLib, void *
...
+            // In some circumstances AOT compilations are performed at warm
+            if ((TR::Options::getCmdLineOptions()->getAggressivityLevel() == TR::Options::AGGRESSIVE_AOT ||
+               that->getCompilationInfo()->importantMethodForStartup((J9Method*)method) ||
+               !TR::Options::getAOTCmdLineOptions()->getOption(TR_DisableAotAtCheapWarm)) &&
+               p->_optimizationPlan->isOptLevelDowngraded() &&
+               p->_optimizationPlan->getOptLevel() == cold) // Is this test really needed?
+               {
+               p->_optimizationPlan->setOptLevel(warm);
+               p->_optimizationPlan->setOptLevelDowngraded(false);
+               aotCompilationReUpgradedToWarm = true;
```
vijaysun-omr: Is this complication in the logic the best way to implement this ? 
I'm asking specifically about the opt level having been downgraded first (the condition that was checked before we entered this code) and then "reupgraded" afterwards ? i.e. it almost seems like we went through a warm -> cold -> warm transition in the decision making. Is it necessary to go through these hops to arrive at the same answer we started from, namely "warm". 
Maybe this is unavoidable given some other complexities of the code as it stands but I thought I would at least ask the question.

runtime/compiler/control/CompilationThread.cpp
mpirvu: From the JVM point of view there are 3 phases, EARLY_STARTUP, STARTUP and NOT_STARTUP.
The condition was changed to include the EARLY_STARTUP phase.

runtime/compiler/control/CompilationThread.cpp
mpirvu: I felt that we should not apply the cheap-warm strategy to methods that would not be subject to downgrading, i.e. if I decided that this method should stay at warm, then it should be a 'true' warm optimization level (in particular, if -Xjit:dontDowngradeToCold is used I don't want the switch to cheap-warm). Also, if the user specified -Xjit:initialOptlevel=cold (or even optLevel=cold) I don't want the push to cheap-warm.
Thus, first I need `shouldDowngradeOptLevel()` to run its course and arrive to a decision and only then override that decision. This overriding could have happened in `shouldDowngradeOptLevel()` but in this case I have to remember this decision until the options object is created later on. I had such a solution where I added another flag in OptimizationPlan and then I changed my mind because OptimizationPlan is defined in omr and I wanted to avoid 'polluting' the omr layer with openj9 specific flags

vijaysun-omr: Okay yes, I was expecting an answer that touched on the difference between "real" warm and "cheap" warm. Thanks, I understand the motivation for doing it this way.

PR #4072, opened 2018-12-17
Extract defineGuaranteedID() method in SVM

jdmpapin:
This deduplicates some repetitive symbol validation manager logic.


jdmpapin:
Yes, it should. I have a separate commit that does so ("(h) Add symbols of guaranteed IDs to _seenSymbolsSet"). I only realized after testing this commit that it could have been folded in here. Currently that fix has an incidental dependency (missing from my dependency graph :sweat_smile:) on "(f) Store the SVM ID to symbol mapping as a vector," just based on the order in which I wrote things. If you like I can reorder to include that change in this PR, but it is coming down the pipe regardless

dsouzai:
> If you like I can reorder to include that change in this PR, but it is coming down the pipe regardless

Given that the SVM as it currently stands doesn't add a symbol associated with a guaranteed ID to the seen symbols set, this PR isn't going to introduce a regression. Since the fix is in the pipeline, I'm fine waiting; no point adding unnecessary churn given the number of PRs that we still have to  get through.

andrewcraik:

PR #4074, opened 2018-12-17
Check that SVM symbol <-> ID lookups succeed

jdmpapin:
This series changes `getIDFromSymbol()` and `getSymbolFromID()` so that they require as a precondition that the requested key exists. Looking up an unknown symbol or ID results in an assertion failure (`SVM_ASSERT`). Explicit checks are removed from callers, since they're now redundant.


andrewcraik:

PR #4075, opened 2018-12-17
Fix MethodFromClassAndSigRecord parameter order

jdmpapin:
For this validation type these values are conventionally listed in the following order: `method`, `methodClass`, `beholder`. This conventional order is used everywhere other than this constructor's parameter list:

- `MethodFromClassAndSigRecord`'s field declarations,
- this constructor's initializer list,
- `MethodFromClassAndSigRecord::printFields()`,
- `MethodFromClassAndSigRecord::operator==()`,
- `initializeAOTRelocationHeader()`, and
- (crucially) the only caller: `addMethodFromClassAndSignatureRecord()`.

The last of these had the effect of swapping the `methodClass` and `beholder` IDs in any `TR_ValidateMethodFromClassAndSig` records.

Rearranging the arguments at the call site instead would work as well, but this way the order is consistent everywhere.


andrewcraik:

PR #4077, opened 2018-12-17
Use regular getClassClassPointer() when using SVM

jdmpapin:
It is not necessary to take the `_aotClassClassPointer` path when the symbol validation manager (SVM) is in use.


andrewcraik:

andrewcraik:
based on the review of @dsouzai I'm merging this since sanity is clean.

PR #4085, opened 2018-12-17
Enable local AOT compilations at the client

harryyu1994:


harryyu1994:
FYI @mpirvu, I'm going to run some performance tests to verify the change. And it's ready for review.

PR #4086, opened 2018-12-18
Ensure boolean JNI methods on x86 produce a result of zero or one

hzongaro:
In calling a JNI boolean method, if the native implementation returns
zero, the result is false; if it returns a non-zero value, the result
is true.

This behaviour is consistent with the reference implementation.

Progresses issue #2049

### DIFF
--- a/runtime/compiler/compile/J9SymbolReferenceTable.cpp
+++ b/runtime/compiler/compile/J9SymbolReferenceTable.cpp
@@ -1269,6 +1269,13 @@ static bool isSignatureTypeBool(const char *fieldSignature, int32_t len)
    return len == 1 && fieldSignature[0] == 'Z';
    }
 
+static bool isSignatureReturnTypeBool(const char *methodSignature, int32_t len)
+   {
+   TR_ASSERT(len > 1, "Method signature is unexpectedly short %d", len);
+   // Method signature must end with ")Z" to have a boolean result type
+   return len > 1 && (')' == methodSignature[len-2]) && ('Z' == methodSignature[len-1]);
+   }
+
 bool
 J9::SymbolReferenceTable::isFieldTypeBool(TR::SymbolReference *symRef)
    {
@@ -1285,6 +1292,16 @@ J9::SymbolReferenceTable::isStaticTypeBool(TR::SymbolReference *symRef)
    const char *fieldSignature = symRef->getOwningMethod(comp())->staticSignatureChars(symRef->getCPIndex(), len);
    dumpOptDetails(comp(), "got static signature as %s\n", fieldSignature);
    return isSignatureTypeBool(fieldSignature, len);
+   }
+
+bool
+J9::SymbolReferenceTable::isReturnTypeBool(TR::SymbolReference *symRef)
+   {
+   TR_Method *method = symRef->getSymbol()->castToResolvedMethodSymbol()->getMethod();
+   char *methodSignature = method->signatureChars();
+   const int32_t len = method->signatureLength();
+   dumpOptDetails(comp(), "got method signature as %.*s\n", len, methodSignature);
+   return isSignatureReturnTypeBool(methodSignature, len);
    }
 
 static bool parmSlotCameFromExpandingAnArchetypeArgPlaceholder(int32_t slot, TR::ResolvedMethodSymbol *sym, TR_Memory *mem)
--- a/runtime/compiler/x/amd64/codegen/AMD64JNILinkage.cpp
+++ b/runtime/compiler/x/amd64/codegen/AMD64JNILinkage.cpp
@@ -1181,14 +1181,25 @@ void TR::AMD64JNILinkage::cleanupReturnValue(
       // type so that we sign and zero extend the narrower integer return types properly.
       //
       TR_X86OpCodes op;
+      TR::SymbolReference      *callSymRef = callNode->getSymbolReference();
       TR::ResolvedMethodSymbol *callSymbol = callNode->getSymbol()->castToResolvedMethodSymbol();
       TR_ResolvedMethod *resolvedMethod = callSymbol->getResolvedMethod();
 
       bool isUnsigned = resolvedMethod->returnTypeIsUnsigned();
+
       switch (resolvedMethod->returnType())
          {
          case TR::Int8:
-            if (isUnsigned)
+            if (comp()->getSymRefTab()->isReturnTypeBool(callSymRef))
+               {
+               // For bool return type, must check whether value returned by
+               // JNI is zero (false) or non-zero (true) to yield Java result
+               generateRegRegInstruction(TEST1RegReg, callNode,
+                     linkageReturnReg, linkageReturnReg, cg());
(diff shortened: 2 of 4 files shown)


runtime/compiler/x/amd64/codegen/AMD64JNILinkage.cpp
```diff
@@ -1181,14 +1181,25 @@ void TR::AMD64JNILinkage::cleanupReturnValue(
       // type so that we sign and zero extend the narrower integer return types properly.
       //
       TR_X86OpCodes op;
+      TR::SymbolReference      *callSymRef = callNode->getSymbolReference();
       TR::ResolvedMethodSymbol *callSymbol = callNode->getSymbol()->castToResolvedMethodSymbol();
       TR_ResolvedMethod *resolvedMethod = callSymbol->getResolvedMethod();
 
       bool isUnsigned = resolvedMethod->returnTypeIsUnsigned();
+bool isBool;
```
hzongaro: Left behind unused declaration from debugging

hzongaro: Fixed in commit https://github.com/eclipse/openj9/pull/4086/commits/6758341778bba8d756cea2bfa472e9fa49d3165b

andrewcraik:
@0dvictor can you review the x86 changes please?

runtime/compiler/x/amd64/codegen/AMD64JNILinkage.cpp
```diff
@@ -1181,14 +1181,25 @@ void TR::AMD64JNILinkage::cleanupReturnValue(
...
          case TR::Int8:
-            if (isUnsigned)
+            if (comp()->getSymRefTab()->isReturnTypeBool(callSymRef))
+               {
+               // For bool return type, must check whether value returned by
+               // JNI is zero (false) or non-zero (true) to yield Java result
+               generateRegRegInstruction(TEST1RegReg, callNode,
+                     linkageReturnReg, linkageReturnReg, cg());
+               generateRegInstruction(SETNE1Reg, callNode, linkageReturnReg, cg());
+               op = TR::Compiler->target.is64Bit() ? MOVZXReg8Reg1 : MOVZXReg4Reg1;
```
0dvictor: I like the way you keep this consistent with the rest of codes.
Just a side note: I'm not sure if checking `TR::Compiler->target.is64Bit()` as this file is only consumed by 64-bit JIT. We may want to revisit this file later and cleanup the unnecessary bitness check.

runtime/compiler/x/i386/codegen/IA32JNILinkage.cpp
```diff
@@ -338,8 +338,19 @@ TR::Register *TR::IA32JNILinkage::buildJNIDispatch(TR::Node *callNode)
...
       case TR::Int8:
-         generateRegRegInstruction(isUnsigned ? MOVZXReg4Reg1 : MOVSXReg4Reg1,
-                                   callNode, ecxReal, eaxReal, cg());
+         if (comp()->getSymRefTab()->isReturnTypeBool(callSymRef))
+            {
+            // For bool return type, must check whether value returned by
+            // JNI is zero (false) or non-zero (true) to yield Java result
+            generateRegRegInstruction(TEST1RegReg, callNode, eaxReal, eaxReal, cg());
+            generateRegInstruction(SETNE1Reg, callNode, ecxReal, cg());
+            generateRegRegInstruction(MOVZXReg4Reg1, callNode, ecxReal, ecxReal, cg());
```
0dvictor: This line is nearly identical (minus the `isUnsigned` check) with the `else` case, maybe we can common them into a fall-through. I'm not sure the value of `isUnsigned` here; however, `ecxReal` can only be 0 or 1 in this case, zero-extension and sign-extension yield same result.

runtime/compiler/compile/J9SymbolReferenceTable.cpp
```diff
@@ -1269,6 +1269,12 @@ static bool isSignatureTypeBool(const char *fieldSignature, int32_t len)
    return len == 1 && fieldSignature[0] == 'Z';
    }
 
+static bool isSignatureReturnTypeBool(const char *methodSignature, int32_t len)
+   {
+   // Method signature must end with ")Z" to have a boolean result type
+   return len > 1 && (')' == methodSignature[len-2]) && ('Z' == methodSignature[len-1]);
```
0dvictor: We may want to have an assertion here to make sure `strlen(methodSignature)` and `len` is at least two.

runtime/compiler/x/amd64/codegen/AMD64JNILinkage.cpp
hzongaro: Yes, I was surprised by the number of places in this file that check `target.is64Bit()`.  Were developers thinking at some point earlier on that there could be just one version of the code gen for x86 for both 64 bit and 32 bit?  That's how things seem to be arranged for Power.

Should I open an issue to track that suggestion for cleanup?

runtime/compiler/x/i386/codegen/IA32JNILinkage.cpp
hzongaro: I thought about doing that, but then I decided to put the result of `SETNE` into `ecxReal` instead of overwriting `eaxReal`.  I'll switch back to use your suggestion.

runtime/compiler/compile/J9SymbolReferenceTable.cpp
hzongaro: I think the value of `methodSignature` is not necessarily null-terminated, so I will just assert that `len > 1`

runtime/compiler/x/amd64/codegen/AMD64JNILinkage.cpp
0dvictor: Yup, I think an Issue to track the cleanup is good for now.

runtime/compiler/compile/J9SymbolReferenceTable.cpp
hzongaro: Revised in commit https://github.com/eclipse/openj9/pull/4086/commits/9e34a9d6f838efda22f40c737d833e5c60f4ed00

runtime/compiler/x/i386/codegen/IA32JNILinkage.cpp
hzongaro: Revised in commit https://github.com/eclipse/openj9/pull/4086/commits/fefc9a56c812601030c115d541c1a6c8f606f6bb

andrewcraik:

runtime/compiler/x/amd64/codegen/AMD64JNILinkage.cpp
hzongaro: Created issue #4140

hzongaro:
I will collapse these changes into a single commit

andrewcraik:
No new sanity failures - merging

PR #4087, opened 2018-12-18
Add libffi preconf files for AArch64

knn-k:
This commit adds two header files in runtime/libffi/preconf/ for
aarch64.


knn-k:
I chose the directory name `xr64` for Linux ARM 64-bit.  Should it be anything else?

knn-k:
> Is that naming scheme consistent with the m4 files added to support the Unsafe(?) helpers?

Yes, I added it in #3502 (not merged yet).

DanHeidinga:

DanHeidinga:
Testing an x86 compile to ensure that these files aren't picked up by the build system on other platforms. (don't expect them to be)

knn-k:
I added changes for aarch64 support to libffi/module.xml.

PR #4095, opened 2018-12-18
Do not update prexAssumptions with client values

dmitry-ten:
Number of prexAssumptions is updated and accessed only by the optimizer,
so CH table update shouldn't overwrite it by values from the client.


dmitry-ten:
@mpirvu could you review?

dmitry-ten:
Performance:
With this change (JITaaS): 11871 (average over 3 runs)
Before this change (JITaaS): 11757 (average over 5 runs)
So it seems there is a ~1% improvement.
The only optimization this affects is `InvariantArgumentPreexistence`, and it's only used in comparison to the max number of pre-existing assumptions, so I would not expect a large improvement.

dmitry-ten:
Yes, it's initialized to 0 in the constructor of `TR_PersistentClassInfo`.

PR #4096, opened 2018-12-18
Support AOT testing

renfeiw:
- turn on aot testing when TEST_FLAG contains AOT
- generate test commands for multiple times when AOT is on

[ci skip]


renfeiw:
reviewer: @llxia

dsouzai:
I'm not sure if it needs to happen in this PR, but in order to stress test AOT, the tests should also do 

```
export TR_Options="forceAOT"
export TR_OptionsAOT="forceAOT"
```

prior to running the tests. The reason for using the env var instead of `-Xjit:forceAOT -Xaot:forceAOT` is because if a test already has `-Xjit`/`-Xaot` options, then adding more `-Xjit`/`-Xaot` options to the end of the command line will cause the JIT to only process the last ones.

Adding the env vars might be something that is already planned, but I figured I'd mention it just in case :)

renfeiw:
Thanks for pointing out the env vars, I have added them into the commit.

test/TestConfig/scripts/testKitGen/testKitGen.pl
```diff
@@ -79,9 +82,9 @@
 			. "                            If the modesXml is not provided, the program will try to find modes.xml under projectRootDir/TestConfig/resources.\n"
 			. "  --ottawaCsv=<path>        Path to ottawa.csv file.\n"
 			. "                            If the ottawaCsv is not provided, the program will try to find ottawa.csv under projectRootDir/TestConfig/resources.\n"
-			. "  --buildList=<paths>       Comma separated project paths to search playlist.xml. The paths are relative to projectRootDir.\n"
```
llxia: Why `buildList` is removed? Is it intended?

test/TestConfig/scripts/testKitGen/testKitGen.pl
```diff
@@ -79,9 +82,9 @@
...
-			. "                            If the buildList is not provided, the program will search all files under projectRootDir.\n";
-		die "Please specify valid options!";
+			. "  --iterations=<number>     The iterations to repeat for each generated test command.\n"
```
llxia: Repeatedly generate test command based on iteration number. The default value is 1.

test/TestConfig/featureSettings.mk
```diff
@@ -0,0 +1,30 @@
...
+#  Exception [1] and GNU General Public License, version 2 with the
+#  OpenJDK Assembly Exception [2].
+#
+#  [1] https://www.gnu.org/software/classpath/license.html
+#  [2] http://openjdk.java.net/legal/assembly-exception.html
+#
+#  SPDX-License-Identifier: EPL-2.0 OR Apache-2.0 OR GPL-2.0 WITH Classpath-exception-2.0 OR LicenseRef-GPL-2.0 WITH Assembly-exception
+##############################################################################
+
+ITERATIONS = 1
```
llxia: `ITERATIONS` is used in Jenkins file as Grinder iterations. 
https://github.com/AdoptOpenJDK/openjdk-tests/blob/7f436b093a45cfba07562ef92b77ba08b4fbc017/buildenv/jenkins/JenkinsfileBase#L242

We may want to choose another name to avoid confusion. What about `TEST_ ITERATIONS`?

llxia:
Grinder failed: https://ci.eclipse.org/openj9/view/Test/job/Test-Grinder/128/console

```
00:01:01.575 JIT: env var TR_OptionsAOT is set to "forceAOT"
00:01:01.575 <AOT: unrecognized option --> '"forceAOT"'>
00:01:01.671 JVMJ9VM015W Initialization error for library j9jit29(11): cannot initialize JIT
00:01:01.671 Error: Could not create the Java Virtual Machine.
00:01:01.671 Error: A fatal exception has occurred. Program will exit.
00:01:01.671 <AOT: fatal error, invalid command line>
00:01:01.671 makefile:82: recipe for target 'compile' failed
```

renfeiw:
I have exported the two env vars before compilation, looks like they can only be exported before running the tests? @dsouzai

test/TestConfig/scripts/testKitGen/testKitGen.pl
renfeiw: updated

test/TestConfig/scripts/testKitGen/testKitGen.pl
renfeiw: updated

test/TestConfig/featureSettings.mk
renfeiw: updated

dsouzai:
I think it's because the environment variable has `"` around it; I thought when exporting the environment variable, the `"` characters got removed, but I guess not? Try doing

```
export TR_Options=forceAOT
export TR_OptionsAOT=forceAOT
```

llxia:
jenkins test sanity xlinux,aix,win jdk11

llxia:
Grinder test with `TEST_FLAG=AOT`:
https://ci.eclipse.org/openj9/view/Test/job/Test-Grinder/131/consoleFull

dsouzai:
Yeah the output looks good; I do see:

1. The cache destroyed prior to and after running the tests (but not between the iterations)
2. The test run twice with `-Xshareclasses:name=floatSanityTests_0 -Xscmx400M -Xscmaxaot256m`
3. The `forceAOT` option is recognized by the JIT

So I believe the test is performing as per the requirements. Thanks for this :)

llxia:
@renfeiw Docs (i.e., readme, wiki, etc) need to be updated for AOT testing. This can be done in another PR.

PR #4097, opened 2018-12-18
Fix local AOT compilation bug

harryyu1994:
[skip ci]


harryyu1994:
Strange thing is that even when we specify "-Xshareclasses:none", `TR::Options::sharedClassCache()` still returns true, looking into why that's the case.

PR #4101, opened 2018-12-18
Properly find machine base pointer from C stack on X86

0dvictor:
The previous approach is improper and dangerous, fixing.


andrewcraik:

PR #4104, opened 2018-12-18
Setup for Java12andUp tests (jep334)

theresa-m:
see https://github.com/eclipse/openj9/issues/4195 for dependencies

- including playlist.xml entry for jep334 tests

### DIFF
--- a/test/functional/Java12andUp/playlist.xml
+++ b/test/functional/Java12andUp/playlist.xml
@@ -1,0 +1,48 @@
+<?xml version="1.0" encoding="UTF-8"?>
+
+<!--
+  Copyright (c) 2018, 2019 IBM Corp. and others
+
+  This program and the accompanying materials are made available under
+  the terms of the Eclipse Public License 2.0 which accompanies this
+  distribution and is available at https://www.eclipse.org/legal/epl-2.0/
+  or the Apache License, Version 2.0 which accompanies this distribution and
+  is available at https://www.apache.org/licenses/LICENSE-2.0.
+
+  This Source Code may also be made available under the following
+  Secondary Licenses when the conditions for such availability set
+  forth in the Eclipse Public License, v. 2.0 are satisfied: GNU
+  General Public License, version 2 with the GNU Classpath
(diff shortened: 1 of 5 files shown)


theresa-m:
@smlambert can I have a review from your team?

theresa-m:
update copyright, merge conflict

DanHeidinga:
@llxia Can you or a delegate review this?

theresa-m:
updated to resolve conflicting files

test/functional/Java12andUp/playlist.xml
```diff
@@ -0,0 +1,48 @@
...
+                  $(TEST_STATUS)
+          </command>
+          <levels>
+                  <level>sanity</level>
+          </levels>
+          <groups>
+                  <group>functional</group>
+          </groups>
+          <subsets>
+                  <subset>12</subset>
```
smlambert: I am guessing these tests would be for 12+ subsets/versions.

As discussed in separate review, this can stay implementation-agnostic, until someone adds a variant/mode that contains a command line option that hotspot doesn't recognize.  With NoOptions, its good as is.

theresa-m: yes it should be 12+. I will fix that

theresa-m: updated with that change

smlambert:
Launched Grinder: https://ci.eclipse.org/openj9/view/Test/job/Test-Grinder/272
to verify that testng doesn't complain if no test classes specified.

@theresa-m - have you tried running this test target locally ?

theresa-m:
ooh just tried the test frame work does not like that its empty. I can lump this commit in with one of my prs that actually adds a test. I made this separate because I wasn't sure which one would get merged first

theresa-m:
Adding this commit to https://github.com/eclipse/openj9/pull/4105 instead to avoid the compilation problem.

smlambert:
no worries, we could have additionally added a dummy class in this PR, but looks like you have a good approach to shift it anyway :)  Thanks for strengthening the jdk12 testing story, very much appreciated!!

PR #4106, opened 2018-12-18
Fix stack shape calling out from JIT helpers on X86

0dvictor:
The stack shape seen in JIT helper's C implementations should be same as
if the C function is called directly from JIT.

### DIFF
--- a/runtime/codert_vm/xnathelp.m4
+++ b/runtime/codert_vm/xnathelp.m4
@@ -970,6 +970,7 @@ END_PROC(jitDecompileOnReturnJ)
 
 START_PROC(jitReadBarrier)
 ifdef({OMR_GC_CONCURRENT_SCAVENGER},{
+	pop uword ptr J9TR_VMThread_jitReturnAddress[_rbp]
 	SWITCH_TO_C_STACK
 	SAVE_C_VOLATILE_REGS
 	dnl currentThread->javaVM->memoryManagerFunctions->J9ReadBarrier(currentThread,(fj9object_t*)currentThread->floatTemp1);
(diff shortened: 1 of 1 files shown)


runtime/codert_vm/xnathelp.m4
```diff
@@ -988,6 +990,7 @@ ifdef({OMR_GC_CONCURRENT_SCAVENGER},{
 END_PROC(jitReadBarrier)
 
 START_PROC(jitReferenceArrayCopy)
+	pop uword ptr J9TR_VMThread_jitReturnAddress[_rbp]
```
DanHeidinga: This should probably go after the `FASTCALL_EXTERN`.  It makes more sense (at least to me) to group the executable code together.

gacholio:
Looks reasonable - same comment as Dan.

runtime/codert_vm/xnathelp.m4
0dvictor: Good suggestion, it has been updated.

DanHeidinga:

DanHeidinga:
Jenkins test extended win jdk11

runtime/codert_vm/xnathelp.m4
0dvictor: Good point, I've moved it outside. Though I don't think it would cause any failures on other platforms as this M4 file is only consumed by X86.

gacholio: Yep - vacation brain :)

gacholio:
This passed 90+% of the testing before the last push, so we'll call this good.

PR #4107, opened 2018-12-18
Properly setup C stack shape on X86-64 Windows

0dvictor:
Microsoft x64 calling convention requires 4 slots register parameter
register area on stack before calling into any functions.

### DIFF
--- a/runtime/oti/j9nonbuilder.h
+++ b/runtime/oti/j9nonbuilder.h
@@ -5658,6 +5658,9 @@ typedef struct J9CInterpreterStackFrame {
 	U_8 jitFPRs[16 * 8]; /* fpr0-15 */
 #endif /* J9VM_ENV_DATA64 */
 #elif defined(J9VM_ARCH_X86) /* J9VM_ARCH_ARM */
+#if defined(J9VM_ENV_DATA64) && defined(WIN32)
+	UDATA arguments[4]; /* outgoing arguments shadow */
+#endif /*J9VM_ENV_DATA64 && WIN32*/
 	UDATA vmStruct;
 	UDATA machineBP;
 #if defined(J9VM_ENV_DATA64)
@@ -5683,7 +5686,7 @@ typedef struct J9CInterpreterStackFrame {
 		} named;
 	} jitGPRs;
 	U_64 jitFPRs[16];
-	#if defined(WIN32)
+#if defined(WIN32)
 	/* Windows x86-64
 	 *
 	 * Stack must be 16-byte aligned.


0dvictor:
Marking the PR as WIP for now, as it requires #4101.

runtime/oti/j9nonbuilder.h
```diff
@@ -5658,6 +5658,9 @@ typedef struct J9CInterpreterStackFrame {
 	U_8 jitFPRs[16 * 8]; /* fpr0-15 */
 #endif /* J9VM_ENV_DATA64 */
 #elif defined(J9VM_ARCH_X86) /* J9VM_ARCH_ARM */
+#if defined(J9VM_ENV_DATA64) && defined(WIN32)
+	UDATA arguments[4]; /* outgoing arguments shadow */
+#endif /*J9VM_ENV_DATA64 && WIN32*/
 	UDATA vmStruct;
 	UDATA machineBP; /* JIT has this offset hard-coded - do not move this field */
```
charliegracie: By adding the `arguments` field above have you changed the offset for this field? If so does it require JIT changes as well?

charliegracie: nvm I missed the changes in #4101

0dvictor: Yes, #4101 ensures JIT picks the correct offset.

andrewcraik:
@DanHeidinga I think this is one of the files you normally review? This is needed to fix the linkage on win64 and get CS working properly.

DanHeidinga:
@gacholio FYI

DanHeidinga:
@0dvictor Do you have a link to documentation that describes the Windows stack frame layout?

DanHeidinga:
For doc, I found: https://docs.microsoft.com/en-us/cpp/build/x64-calling-convention?view=vs-2017 which states:
>The caller is responsible for allocating space for parameters to the callee, and must always allocate sufficient space to store four register parameters, even if the callee doesn’t take that many parameters. This convention simplifies support for unprototyped C-language functions and vararg C/C++ functions. For vararg or unprototyped functions, any floating point values must be duplicated in the corresponding general-purpose register. Any parameters beyond the first four must be stored on the stack after the shadow store prior to the call. Vararg function details can be found in Varargs. Unprototyped function information is detailed in Unprototyped Functions.

Working though the code to determine if further adjustments are required for:
> Any parameters beyond the first four must be stored on the stack after the shadow store prior to the call.

DanHeidinga:
Code flow is:
* Call one of the call-in methods (all taking 5 parameters) from callin.cpp 
https://github.com/eclipse/openj9/blob/1e7ab9b2abe374133f3f0d98c0160b9711eb3f30/runtime/vm/callin.cpp#L1038

* Eventually calling the cInterpreter: https://github.com/eclipse/openj9/blob/1e7ab9b2abe374133f3f0d98c0160b9711eb3f30/runtime/vm/callin.cpp#L1073

* cInterpreter sets up the call-in frame
https://github.com/eclipse/openj9/blob/2f295ea91cdd97ff37eb5d164fc63014ea718bc7/runtime/vm/xcinterp.m4#L158-L159
https://github.com/eclipse/openj9/blob/2f295ea91cdd97ff37eb5d164fc63014ea718bc7/runtime/vm/xcinterp.m4#L33-L43

* and then jump to the bytecode loop:
https://github.com/eclipse/openj9/blob/2f295ea91cdd97ff37eb5d164fc63014ea718bc7/runtime/vm/xcinterp.m4#L168

* which takes a single parameter:
https://github.com/eclipse/openj9/blob/2f295ea91cdd97ff37eb5d164fc63014ea718bc7/runtime/vm/BytecodeInterpreter.cpp#L97

For the interpreter, there is only a very short time the cInterpreter frame is the current frame - just while calling the cInterpreter.

For the JIT, this frame is live while executing until / unless the JIT builds a new frame.

From an interpreter point of view, the 4 args are sufficient.  
@0dvictor Can you confirm that the JIT doesn't make any c calls on the c stack taking more than 4 parameters?

DanHeidinga:
Jenkins test all xlinux jdk8

DanHeidinga:
Jenkins test sanity win jdk11

0dvictor:
@DanHeidinga JIT only uses this frame for JIT helpers; none of JIT helpers takes more than 4 arguments and hence the 4 arguments are sufficient.

DanHeidinga:
Change looks good.  I'll merge once the PR testing is complete

gacholio:
Yep, looks reasonable.

DanHeidinga:
@gacholio A bit of an aside, but given the code flow I outlined above, do you see any reason to continue to ensure all call-ins are 5 parameters?

gacholio:
No, there's no longer any reason for that. Also, we may be able to shrink the frames on the other architectures - I don't know if direct JNI depends on the 32 argument slots being there. The interpreter certainly doesn't (FFI builds its own stack frame for each call).

Opened #4134 to investigate.

0dvictor:
> I don't know if direct JNI depends on the 32 argument slots being there

On X86, JIT always buys stack before calling JNI, and hence does not use the 32-byte argument slots. Though we probably can do some optimization taking advantage of the slots if the callee has no more than 4 parameters.

PR #4110, opened 2018-12-18
Store the SVM ID to symbol mapping as a vector

jdmpapin:
Here `vector` should be more efficient than `map` because symbol IDs are small sequential integers.

### DIFF
--- a/runtime/compiler/runtime/SymbolValidationManager.cpp
+++ b/runtime/compiler/runtime/SymbolValidationManager.cpp
@@ -65,7 +65,7 @@ TR::SymbolValidationManager::SymbolValidationManager(TR::Region &region, TR_Reso
      _region(region),
      _symbolValidationRecords(_region),
      _symbolToIdMap((SymbolToIdComparator()), _region),
-     _idToSymbolsMap((IdToSymbolComparator()), _region),
+     _idToSymbolTable(_region),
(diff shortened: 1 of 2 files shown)


runtime/compiler/runtime/SymbolValidationManager.cpp
```diff
@@ -92,7 +92,17 @@ TR::SymbolValidationManager::defineGuaranteedID(void *symbol)
...
    _symbolToIdMap.insert(std::make_pair(symbol, id));
-   _idToSymbolsMap.insert(std::make_pair(id, symbol));
+   setSymbolOfID(id, symbol);
+   }
+
+void
+TR::SymbolValidationManager::setSymbolOfID(uint16_t id, void *symbol)
+   {
+   if (id >= _idToSymbolTable.size())
+      _idToSymbolTable.resize(id + 1, NULL);
```
dsouzai: To clarify, do the elements (that haven't already been set) get initialized to `0`?

jdmpapin: Yes, the second argument (`NULL` in this case) is the value to copy into new elements when `resize` increases the size of the vector.

andrewcraik:

jdmpapin:
The PPC failure looks unrelated; it also happened in #4111.

andrewcraik:
This is good to go - merging.

PR #4111, opened 2018-12-18
Lazily initialize class pointers on Compilation

jdmpapin:
When the symbol validation manager is in use, this lazy initialization limits the generated by-name validation records for these system classes to the ones that are actually requested (via `getObjectClassPointer()`, etc.) during compilation.

### DIFF
--- a/runtime/compiler/compile/J9Compilation.cpp
+++ b/runtime/compiler/compile/J9Compilation.cpp
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2000, 2018 IBM Corp. and others
+ * Copyright (c) 2000, 2019 IBM Corp. and others
  *
  * This program and the accompanying materials are made available under
  * the terms of the Eclipse Public License 2.0 which accompanies this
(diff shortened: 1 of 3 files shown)


dsouzai:
Travis CI is failing because the linter is complaining about missing `self()` calls :)

jdmpapin:
Updated to appease the extensible class lint

andrewcraik:

jdmpapin:
Git error in Windows JDK8 sanity test

jdmpapin:
Need another `self()->` in `J9Compilation.cpp`. I'll add that.

The PPC failure looks unrelated; it also happened in #4110.

I don't have the code that was crashing in x86-64 Linux JDK11, since for PR builds the binaries get deleted. I'm unsure how to investigate

jdmpapin:
Rebased, but the copyright header check is wishing me a happy new year. Leaving this comment before fixing the dates because I've seen GH collapse a series of two force pushes, showing only the second. Hopefully this will preserve the first diff link.

jdmpapin:
Updated the copyright headers

jdmpapin:
The x86-64 Linux JDK11 test failures are due to #4135.

The Windows tests ran out of disk space before starting.

andrewcraik:
Jenkins test sanity win jdk8,jdk11

jdmpapin:
Windows JDK11 tests ran out of disk space again

andrewcraik:
windows jdk11 failed due to space as noted, jdk11 linux x86 failed due to #4135. Change looks good.

Issue #4090, opened 2018-12-18
Test-extended.functional-JDK11-osx_x86-64_cmprssptrs testSoftMxDisclaimMemory Heap can't shrink


pshipton:
https://ci.eclipse.org/openj9/job/Test-extended.functional-JDK11-osx_x86-64_cmprssptrs/20
testSoftMxDisclaimMemory_2
testSoftMxDisclaimMemory_LP4k_2
```
===============================================
Running test testSoftMxDisclaimMemory_2 ...
===============================================
testSoftMxDisclaimMemory_2 Start Time: Tue Dec 18 04:43:31 2018 Epoch Time (ms): 1545126211692
variation: Mode601
JVM_OPTIONS: -Xcompressedrefs -Xcompressedrefs -Xjit -Xgcpolicy:optthruput 
[IncludeExcludeTestAnnotationTransformer] [INFO] exclude file is /Users/jenkins/workspace/Test-extended.functional-JDK11-osx_x86-64_cmprssptrs/openjdk-tests/TestConfig/scripts/testKitGen/../../../../jvmtest/TestConfig/resources/excludes/latest_exclude_11.txt
...
... TestNG 6.14.2 by Cédric Beust (cedric@beust.com)
...

[SoftmxAdvanceTest] [ERROR] 	FAIL: Memory didn't release back to OS! Post memory size 3823534080 is less than or equal to previous memory size 3821236224
[SoftmxAdvanceTest] [WARN] 	Generate Java core dump after forcing GC.
JVMJ9VM079W triggerOneOffDump(java) requires j9dmp29
JVMJ9VM079W triggerOneOffDump(heap) requires j9dmp29
JVMJ9VM079W triggerOneOffDump(system) requires j9dmp29
[SoftmxAdvanceTest] [ERROR] 	FAIL: Heap can't shrink to the reset max heap size within 5 minutes!
[SoftmxAdvanceTest] [WARN] 	Generate Java core dump after forcing GC.
JVMJ9VM079W triggerOneOffDump(java) requires j9dmp29
JVMJ9VM079W triggerOneOffDump(heap) requires j9dmp29
JVMJ9VM079W triggerOneOffDump(system) requires j9dmp29
[SoftmxAdvanceTest] [ERROR] 	FAIL: Heap can't shrink to the reset max heap size within 5 minutes!
Exception in thread "MemoryMXBean notification dispatcher" java.lang.OutOfMemoryError: Java heap space
[...]

PR #4116, opened 2018-12-19
remove -Xdump:none and add verbosegclog for all SoftMx tests

llxia:
Issue: #4090
[ci skip]

### DIFF
--- a/test/functional/JLM_Tests/playlist.xml
+++ b/test/functional/JLM_Tests/playlist.xml
@@ -209,7 +209,7 @@
 		</variations>
 		<command>$(ADD_JVM_LIB_DIR_TO_LIBPATH) \
 	$(JAVA_COMMAND) $(JVM_OPTIONS) \
-	-Xdump:none -Xmx1024m -Xsoftmx512m -Xmn1m \
+	-Xmx1024m -Xsoftmx512m -Xmn1m -verbose:gc -Xverbosegclog:$(REPORTDIR)$(D)vgc.log \
 	-cp $(Q)$(RESOURCES_DIR)$(P)$(TESTNG)$(P)$(TEST_RESROOT)$(D)jlm_tests.jar$(Q) \
 	org.testng.TestNG -d $(REPORTDIR) $(Q)$(TEST_RESROOT)$(D)testng.xml$(Q) \
 	-testnames testSoftMxLocal \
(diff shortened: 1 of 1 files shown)


llxia:
jenkins test extended xlinux,aix,win jdk11

llxia:
jenkins test extended xlinux,win jdk8

DanHeidinga:
@llxia Can you check the internal history for the rationale on why we added `-Xdump:none` in the first place?  If I had to guess, it would be related to the test generating cores too frequently for a case that was generally expected to occur.

llxia:
Unfortunately, we cannot find the rationale from the history. It seems the tests were migrated from cvs to rtc in 2013 and we do not have cvs history anymore.

How about we still keep `-Xdump:none` in command and move `$(JVM_OPTIONS)` after hardcoded options? 

```
$(JAVA_COMMAND) \
-Xdump:none -Xmx1024m -Xsoftmx512m -Xmn1m -verbose:gc -Xverbosegclog:$(REPORTDIR)$(D)vgc.log $(JVM_OPTIONS) \
```

In this way, the tests will not generate cores and users can debug it by specifying `-Xdump:XXXX` in `EXTRA_OPTIONS` in Grinder.

DanHeidinga:
`-Xdump:system:none:events=systhrow,filter=java/lang/OutOfMemoryError` would disable the system core on OOM - https://www.eclipse.org/openj9/docs/xdump/ - and leave the javacore.

@dmitripivkine Would that be sufficient to give you info you need to validate the GC functionality isn't broken?

dmitripivkine:
> Failures are intermittent and usually we need to make a decision on every failed case. So I believe it is important to have (at least) javacore to verify that GC Softmx functionality is not broken but OS environment somehow does not provide report about returned memory in time

... or may be GC verbose log might help in this case and we can suppress dumps if it is critical

dmitripivkine:
> `-Xdump:system:none:events=systhrow,filter=java/lang/OutOfMemoryError` would disable the system core on OOM - https://www.eclipse.org/openj9/docs/xdump/ - and leave the javacore.
> 
> @dmitripivkine Would that be sufficient to give you info you need to validate the GC functionality isn't broken?

The OOM is not usual for this test. It something needs to be investigated. So it would not help unfortunately. Cores are requested explicitly from https://github.com/eclipse/openj9/blob/master/test/functional/JLM_Tests/src/j9vm/test/softmx/SoftmxAdvanceTest.java lines 167-169. This code can be changed (may be we have more copy/paste places)

dmitripivkine:
Just to summarize: if we believe that check of free memory in OS has a value we can add gc verbose log generation and move -Xdump:none to external options it should be sufficient for now.

dmitripivkine:
> @dmitripivkine It sounds like we should proceed with @llxia's change as the reason for the `-Xdump:none` is lost. It can always be modified if we find there's a problem.

Yes, this is my point exactly.

PR #4133, opened 2018-12-20
Reorder interpreter jump targets for performance

pdbain-ibm:
Make the order of the JUMP_TARGETs match the bytecode numbers
better.

This causes an approximately 2% improvement in Liberty startup time with -Xint
on Linux PPC.

Signed-off-by: Peter Bain peter_bain@ca.ibm.com

### DIFF
--- a/runtime/vm/BytecodeInterpreter.hpp
+++ b/runtime/vm/BytecodeInterpreter.hpp
@@ -9993,6 +9993,9 @@ executeBytecodeFromLocal:
 		JUMP_TARGET(JBmultianewarray):
 			SINGLE_STEP();
 			PERFORM_ACTION(multianewarray(REGISTER_ARGS));
+		JUMP_TARGET(JBiincw):
+			SINGLE_STEP();
+			PERFORM_ACTION(iinc(REGISTER_ARGS, 2));
 		JUMP_TARGET(JBaloadw):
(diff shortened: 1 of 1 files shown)


DanHeidinga:
@pdbain-ibm Can you add the signed-off by tag to your commit?

Please also add context on how this was found, the performance difference seen with & without it, etc.  I understand there may be details that can't be shared due to the test cases used, but that shouldn't prevent sharing the improvement seen by making this change and the platforms its been tested on.

pdbain-ibm:
@DanHeidinga Done.  Let me know if you ant more details.

DanHeidinga:
@pdbain-ibm Has this change been tested on any other platforms?

DanHeidinga:
Jenkins test sanity plinux,zlinux,win jdk8

DanHeidinga:
Jenkins test sanity plinux,zlinux,win jdk8

DanHeidinga:
@pdbain-ibm Did you get the performance results with this change from the other platforms?  Would be good to record that here if possible

pdbain-ibm:
Our internal performance team reports:

> The iincw changes have been merged. Xint performance has improved as a result. Testing was done for Linux x86 and PPCLE64 with no ill effects observed.

PR #4141, opened 2018-12-20
Add value type static preloading and tests

acrowthe:
Implement the following upon preloading a valueType:
- Preload static fields with a Q signature
- Throw an `IncompatibleClassChangeError` if the loaded class is not a value type
- Throw a `NoClassDefError` if the class is not defined
In addition, tests were added for these cases. These tests verify that the errors are correctly thrown when a Q type matching these issues is loaded.

Presently, the test for IncompatibleClassChangeError cannot be run as
support for generating objects with Q signatures is not present, just
support for Q signature fields. As a result, this test case has been
commented out.


acrowthe:
This is currently blocked by the test changes in https://github.com/eclipse/openj9/pull/3960 and https://github.com/eclipse/openj9/pull/3857. Once those have been merged, this will need to be rebased to remove commits 14cbf80 and c725f6c.

DanHeidinga:
@tajila Is this still relevant for valhalla?

tajila:
work was completed in a another PR

PR #4145, opened 2018-12-21
Revert recent EscapeAnalysis changes from PR 2752

hzongaro:
Pull request #2752 appears to have caused (or exposed) a problem in Escape Analysis that results in intermittent failures running `ClassLibraryLambdaTests` on z/OS.  This change reverts that pull request while we investigate the source of the problem.


andrewcraik:

andrewcraik:
Given the issues seen I think this revert is warranted pending further investigation.

charliegracie:
@hzongaro now that the fix has been released can we give these EA changes another try? :)

hzongaro:
@charliegracie Yes, along with further improvements from @andrewcraik. . . .

PR #4148, opened 2018-12-23
Tolerate upcoming OMR compiler memory allocator changes

mstoodle:
An effort to simplify and clean up the lowest level compiler memory
allocation story in OMR will break OpenJ9. To avoid this breakage,
the changes will be staged across the two projects. Any project
like OpenJ9 that relies on the current compiler memory allocation
scheme can define the uninspiringly named OLD_MEMORY macro to
preserve the older memory scheme. This macro will only be needed in
the compiler's build system, so I contained it there.

Once this change has been merged, OMR will be free to update
its memory allocator without affecting OpenJ9's allocation
story. For more details on what those changes currently look like,
please see https://github.com/eclipse/omr/pull/3396 .

The second stage will be to modify OpenJ9's JIT compiler to
apply the same refactoring as OMR so it will no longer need
to define OLD_MEMORY. Once OLD_MEMORY is removed from OpenJ9,
then OMR will remove the old memory allocation files and lock
in the new scheme. The goal is to avoid any co-dependent pull
requests, although obviously pull request merges across the
two projects will need to be synchronized. I also intend to
create the full sequence of changes before merging anything
new into OMR, to minimize the amount of time OMR will have
two allocation implementations. But this PR is needed to kick
the whole thing off...


mstoodle:

That's overkill, but let's just make absolutely certain given the fairly fundamental changes coming...

mstoodle:
jenkins test sanity xlinux,osx,sin,aix jdk8,jdk11

mstoodle:

mstoodle:
lots of changes have occurred recently in the underlying OMR pull request, so earlier testing results are no longer relevant (plus lots of the builds failed for unrelated reasons)...hopefully things will be cleaner now, so I'm going to go back to basics and re-test this change set without the new OMR pull request, and then test against the pull request

mstoodle:

mstoodle:

PR #4118, opened 2018-12-19
Remove unsupported compiler options for clang

pdbain-ibm:
Remove -fno-signaling-nans and -fno-rounding-math on clang.
These options were causing warnings on MacOS:

> clang: warning: optimization flag '-fno-signaling-nans' is not supported [-Wignored-optimization-argument]
> clang: warning: optimization flag '-fno-rounding-math' is not supported [-Wignored-optimization-argument]
> 

Signed-off-by: Peter Bain <peter_bain@ca.ibm.com>

PR #4156, opened 2018-12-31
Remove unnecessary gcc-specific optimization flags

nbhuiyan:
Using -fno-rounding-math and -fno-signalling-nans flags on OSX
results in Clang warning about these unsupported options for
every use of this option.

Edit: Also removed these options for GCC due to them being defaults.

Edit2: Since #4118 addressed the clang warning issues, this PR has been modified to address just removing the options on GCC.

### DIFF
--- a/runtime/compiler/build/toolcfg/gnu/common.mk
+++ b/runtime/compiler/build/toolcfg/gnu/common.mk
@@ -178,8 +178,6 @@ endif
 
 ifeq ($(C_COMPILER),clang)
     CX_FLAGS+=-Wno-parentheses -Werror=header-guard
-else
-    CX_FLAGS+=-fno-rounding-math -fno-signaling-nans
 endif
 
 ifeq ($(BUILD_CONFIG),debug)


DanHeidinga:
Jenkins test sanity osx jdk8,jdk11

andrewcraik:
@DanHeidinga and @nbhuiyan  - we do need to be careful that the floating point rounding mode is correct without the -f option and that NANs don't trap. These bugs can be really subtle so we need to be sure what the Clang mode is with these options off otherwise the JIT can encounter incorrect floating point behaviour which is an absolute nightmare to debug (speaking from experience). FYI @0dvictor since you added these options originally.

runtime/compiler/build/toolcfg/gnu/common.mk
```diff
@@ -125,7 +125,13 @@ CX_OPTFLAG?=$(CX_DEFAULTOPT)
 CX_FLAGS_PROD+=$(CX_OPTFLAG)
 
 ifeq ($(HOST_ARCH),x)
-    CX_FLAGS+=-mfpmath=sse -msse -msse2 -fno-math-errno -fno-rounding-math -fno-trapping-math -fno-signaling-nans
+    CX_FLAGS+=-mfpmath=sse -msse -msse2 -fno-math-errno -fno-trapping-math
+
+    # -fno-rounding-math and -fno-signalling-nans 
+    # optimization options are not supported on OSX
+    ifneq ($(OS),osx)
+        CX_FLAGS+=-fno-rounding-math -fno-signaling-nans
```
0dvictor: This can be removed for GCC, as both are default, per https://gcc.gnu.org/onlinedocs/gcc/Optimize-Options.html

0dvictor:
Quote from LLVM's documentation:
>The default LLVM floating-point environment assumes that floating-point instructions do not have side effects. Results assume the round-to-nearest rounding mode. No floating-point exception state is maintained in this environment. Therefore, there is no attempt to create or preserve invalid operation (SNaN) or division-by-zero exceptions.

The behavior should be same as GCC's `-fno-rounding-math` and `-fno-signaling-nans`. In addition, as `-fno-rounding-math` and `-fno-signaling-nans` are default on GCC, we can safely remove them for GCC as well.

nbhuiyan:
Thanks for looking into this, @0dvictor! 

If we were to remove these options for gcc as well, I think it's necessary to understand the original motivations behind specifying `-fno-signaling-nans` and `-fno-rounding-math` options for gcc. Was it necessary to specify these options in the version of gcc used at the time they were added? In that case, we may need to verify that the versions of gcc currently used for OpenJ9 builds do have these options by default.

0dvictor:
> Was it necessary to specify these options in the version of gcc used at the time they were added?

No, it wasn't necessary. The four floating point related GCC options are directly inferred from The Java Language Specification on floating point semantics, and hence were added.
IMO, when this file was only consumed when compiling with GCC, it is fine to add extra but non-essential options. Now Clang also uses this file and not support these two options, it is completely unnecessary to selectively add these two to GCC. It also introduces confusion as readers may think Clang behaves differently from GCC.

> we may need to verify that the versions of gcc currently used for OpenJ9 builds do have these options by default.
Good suggestion, I've checked GCC versions from 4.4.7 to 8.2, `-fno-signaling-nans` and `-fno-signaling-nans` are all default.

DanHeidinga:
@0dvictor Are you OK with the updated change?

DanHeidinga:

DanHeidinga:
I see #4118 was merged which addresses the same issue as this PR.  The only difference being that it kept the options for non-clang compiles.

If it's important to remove the options for gcc as well, then @nbhuiyan can you rebase this PR and remove the bit of from #4118?

nbhuiyan:
@DanHeidinga 
As pointed out by @0dvictor , selectively using these options on GCC could potentially cause confusion about the default behavior on Clang and GCC. Therefore, I have removed these options entirely from the makefile.

DanHeidinga:

PR #4157, opened 2018-12-31
Methods and tests for jep334 Class

theresa-m:
- Class.componentType
- Class.describeConstable
- Class.descriptorString
- Class arrayType

Depends on (see https://github.com/eclipse/openj9/issues/4195): 
- Fix Java 12 build failures: https://github.com/eclipse/openj9/pull/3980
- Constants API stubs: https://github.com/eclipse/openj9/pull/4030
- Java12andUp test setup: https://github.com/eclipse/openj9/pull/4104

### DIFF
--- a/jcl/src/java.base/share/classes/java/lang/Class.java
+++ b/jcl/src/java.base/share/classes/java/lang/Class.java
@@ -4561,20 +4561,77 @@ public Class<?>[] getNestMembers() throws LinkageError, SecurityException {
 /*[ENDIF] Java11 */
 
 /*[IF Java12]*/
+	/**
+	 * Create class of an array. The component type will be this Class instance.
+	 * 
+	 * @return array class where the component type is this Class instance
+	 */
 	public Class<?> arrayType() {
-		throw new UnsupportedOperationException("Stub for Java 12 compilation (Jep334)");
+		if (this == void.class) {
+			throw new IllegalArgumentException();
+		}
+		return arrayTypeImpl();
 	}
 
+	private native Class<?> arrayTypeImpl();
+
+	/**
+	 * Answers a Class object which represents the receiver's component type if the receiver 
+	 * represents an array type. The component type of an array type is the type of the elements 
+	 * of the array.
+	 *
+	 * @return the component type of the receiver. Returns null if the receiver does 
+	 * not represent an array.
+	 */
 	public Class<?> componentType() {
-		throw new UnsupportedOperationException("Stub for Java 12 compilation (Jep334)");
+		return getComponentType();
 	}
 
+	/**
+	 * Returns the nominal descriptor of this Class instance, or an empty Optional 
+	 * if construction is not possible.
+	 * 
+	 * @return Optional with a nominal descriptor of Class instance
+	 */
 	public Optional<ClassDesc> describeConstable() {
-		throw new UnsupportedOperationException("Stub for Java 12 compilation (Jep334)");
+		ClassDesc classDescriptor = ClassDesc.ofDescriptor(this.descriptorString());
+		return Optional.of(classDescriptor);
 	}
 
+	/**
+	 * Return field descriptor of Class instance.
+	 * 
+	 * @return field descriptor of Class instance
+	 */
 	public String descriptorString() {
-		throw new UnsupportedOperationException("Stub for Java 12 compilation (Jep334)");
+		/* see MethodType.getBytecodeStringName */
+
+		if (this.isPrimitive()) {
+			if (this == int.class) {
+				return "I"; //$NON-NLS-1$
+			} else if (this == long.class) {
+				return "J"; //$NON-NLS-1$
+			} else if (this == byte.class) {
+				return "B"; //$NON-NLS-1$
+			} else if (this == boolean.class) {
+				return "Z"; //$NON-NLS-1$
+			} else if (this == void.class) {
+				return "V"; //$NON-NLS-1$
+			} else if (this == char.class) {
+				return "C"; //$NON-NLS-1$
+			} else if (this == double.class) {
+				return "D"; //$NON-NLS-1$
+			} else if (this == float.class) {
+				return "F"; //$NON-NLS-1$
+			} else if (this == short.class) {
+				return "S"; //$NON-NLS-1$
+			}
+		}
+		String name = this.getName().replace('.', '/');
+		if (this.isArray()) {
+			return name;
+		}
+		return "L"+ name + ";"; //$NON-NLS-1$ //$NON-NLS-2$
 	}
 /*[ENDIF] Java12 */
 }
--- a/runtime/vm/BytecodeInterpreter.hpp
+++ b/runtime/vm/BytecodeInterpreter.hpp
@@ -2721,6 +2721,26 @@ done:
 		return EXECUTE_BYTECODE;
 	}
 
+	/* java.lang.Class: private native Class<?> arrayTypeImpl(); */
+	VMINLINE VM_BytecodeAction
+	inlClassArrayTypeImpl(REGISTER_ARGS_LIST)
+	{
+		VM_BytecodeAction rc = EXECUTE_BYTECODE;
+		J9Class *componentClazz = J9VM_J9CLASS_FROM_HEAPCLASS(_currentThread, *(j9object_t*)_sp);
+		J9Class *arrayClazz = componentClazz->arrayClass;
+		if (NULL == arrayClazz) {
+			arrayClazz = internalCreateArrayClass(_currentThread, 
+				(J9ROMArrayClass *) J9ROMIMAGEHEADER_FIRSTCLASS(_currentThread->javaVM->arrayROMClasses), 
+				componentClazz);
+		}
+		if (NULL == arrayClazz) {
+			rc = GOTO_THROW_CURRENT_EXCEPTION;
+		} else {
+			returnObjectFromINL(REGISTER_ARGS, J9VM_J9CLASS_TO_HEAPCLASS(arrayClazz), 1);
+		}
+		return rc;
+	}
+
 	/* java.lang.Class: private native String getSimpleNameImpl(); */
 	VMINLINE VM_BytecodeAction
 	inlClassGetSimpleNameImpl(REGISTER_ARGS_LIST)
@@ -8718,6 +8738,7 @@ public:
 		JUMP_TABLE_ENTRY(J9_BCLOOP_SEND_TARGET_VARHANDLE),
 		JUMP_TABLE_ENTRY(J9_BCLOOP_SEND_TARGET_INL_THREAD_ON_SPIN_WAIT),
 		JUMP_TABLE_ENTRY(J9_BCLOOP_SEND_TARGET_OUT_OF_LINE_INL),
+		JUMP_TABLE_ENTRY(J9_BCLOOP_SEND_TARGET_CLASS_ARRAY_TYPE_IMPL),
 	};
 #endif /* !defined(USE_COMPUTED_GOTO) */
 
@@ -9277,6 +9298,8 @@ runMethod: {
 		PERFORM_ACTION(inlThreadOnSpinWait(REGISTER_ARGS));
 	JUMP_TARGET(J9_BCLOOP_SEND_TARGET_OUT_OF_LINE_INL):
 		PERFORM_ACTION(outOfLineINL(REGISTER_ARGS));
+	JUMP_TARGET(J9_BCLOOP_SEND_TARGET_CLASS_ARRAY_TYPE_IMPL):
+		PERFORM_ACTION(inlClassArrayTypeImpl(REGISTER_ARGS));
 #if !defined(USE_COMPUTED_GOTO)
 	default:
 		Assert_VM_unreachable();
--- a/runtime/vm/FastJNI_java_lang_Class.cpp
+++ b/runtime/vm/FastJNI_java_lang_Class.cpp
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2001, 2018 IBM Corp. and others
+ * Copyright (c) 2001, 2019 IBM Corp. and others
  *
  * This program and the accompanying materials are made available under
  * the terms of the Eclipse Public License 2.0 which accompanies this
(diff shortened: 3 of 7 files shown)


runtime/vm/BytecodeInterpreter.hpp
```diff
@@ -2721,6 +2721,20 @@ done:;
 		return EXECUTE_BYTECODE;
 	}
 
+	/* java.lang.Class: public native Class<?> arrayType(); */
+	VMINLINE VM_BytecodeAction
+	inlClassArrayType(REGISTER_ARGS_LIST)
+	{
+		J9Class *componentClazz = J9VM_J9CLASS_FROM_HEAPCLASS(_currentThread, *(j9object_t*)_sp);
```
gacholio: Should fetch arrayClass here and NULL check to avoid the call.

runtime/vm/FastJNI_java_lang_Class.cpp
```diff
@@ -180,6 +180,19 @@ Fast_java_lang_Class_getModifiersImpl(J9VMThread *currentThread, j9object_t rece
 	return (jint)modifiers;
 }
 
+/* java.lang.Class: public native Class<?> arrayType(); */
+j9object_t JNICALL
+Fast_java_lang_Class_arrayType(J9VMThread *currentThread, j9object_t receiverObject)
+{
+	J9Class *componentClazz = J9VM_J9CLASS_FROM_HEAPCLASS(currentThread, receiverObject);
```
gacholio: Similar comment as above.

runtime/vm/FastJNI_java_lang_Class.cpp
```diff
@@ -212,6 +227,9 @@ J9_FAST_JNI_METHOD_TABLE(java_lang_Class)
 	J9_FAST_JNI_METHOD("getComponentType", "()Ljava/lang/Class;", Fast_java_lang_Class_getComponentType,
 		J9_FAST_JNI_RETAIN_VM_ACCESS | J9_FAST_JNI_NOT_GC_POINT | J9_FAST_JNI_NO_NATIVE_METHOD_FRAME | J9_FAST_JNI_NO_EXCEPTION_THROW |
 		J9_FAST_JNI_NO_SPECIAL_TEAR_DOWN | J9_FAST_JNI_DO_NOT_WRAP_OBJECTS)
+	J9_FAST_JNI_METHOD("arrayType", "()Ljava/lang/Class;", Fast_java_lang_Class_arrayType,
```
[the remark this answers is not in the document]
theresa-m: updated

runtime/vm/BytecodeInterpreter.hpp
```diff
@@ -2721,6 +2721,22 @@ done:;
 		return EXECUTE_BYTECODE;
 	}
 
+	/* java.lang.Class: public native Class<?> arrayType(); */
+	VMINLINE VM_BytecodeAction
+	inlClassArrayType(REGISTER_ARGS_LIST)
+	{
+		j9object_t arrayObject = NULL;
+		J9Class *componentClazz = J9VM_J9CLASS_FROM_HEAPCLASS(_currentThread, *(j9object_t*)_sp);
```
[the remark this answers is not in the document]
theresa-m: oh I see what you mean. I will fix that.

runtime/vm/BytecodeInterpreter.hpp
```diff
@@ -2721,6 +2721,22 @@ done:;
...
+		J9Class *arrayClazz = internalCreateArrayClass(_currentThread, 
+				(J9ROMArrayClass *) J9ROMIMAGEHEADER_FIRSTCLASS(_currentThread->javaVM->arrayROMClasses), 
+				componentClazz);
+		if (NULL != arrayClazz) {
+			arrayObject = J9VM_J9CLASS_TO_HEAPCLASS(arrayClazz);
+		}
```
gacholio: In the NULL case, there should be an exception pending, which requires a different return code. I suggest initializing an rc to `GOTO_THROW_CURRENT_EXCEPTION` and setting the rc to `EXECUTE_BYTECODE` in the success case (if the arrayClazz is non-NULL). The `returnObjectFromINL` should also only be performed in the success case.

runtime/vm/BytecodeInterpreter.hpp
```diff
@@ -2721,6 +2721,29 @@ done:;
...
+				(J9ROMArrayClass *) J9ROMIMAGEHEADER_FIRSTCLASS(_currentThread->javaVM->arrayROMClasses), 
+				componentClazz);
+			componentClazz->arrayClass = arrayClazz;
+		}
+		if (NULL == arrayClazz) {
+			rc = GOTO_THROW_CURRENT_EXCEPTION;
+		} else {
+			arrayObject = J9VM_J9CLASS_TO_HEAPCLASS(arrayClazz);
+		}
+		returnObjectFromINL(REGISTER_ARGS, arrayObject, 1);
```
gacholio: This needs to move into the non-NULL arrayClazz case.

runtime/vm/FastJNI_java_lang_Class.cpp
```diff
@@ -180,6 +180,25 @@ Fast_java_lang_Class_getModifiersImpl(J9VMThread *currentThread, j9object_t rece
...
+Fast_java_lang_Class_arrayType(J9VMThread *currentThread, j9object_t receiverObject)
+{
+	j9object_t arrayObject = NULL;
+	J9Class *componentClazz = J9VM_J9CLASS_FROM_HEAPCLASS(currentThread, receiverObject);
+	J9Class *arrayClazz = componentClazz->arrayClass;
+	if (NULL == arrayClazz) {
+		arrayClazz = internalCreateArrayClass(currentThread, 
+			(J9ROMArrayClass *) J9ROMIMAGEHEADER_FIRSTCLASS(currentThread->javaVM->arrayROMClasses), 
+			componentClazz);
+		componentClazz->arrayClass = arrayClazz;
```
gacholio: This assignment is unnecessary - the call will do it.

gacholio:
@andrewcraik New native Class.arrayType should be recognized by the JIT if we this think it will be called a lot. The expected case is a few instructions, so making a call every time seems wasteful. The sequence should be:

```
result = clazz->arrayClass;
if (NULL == result) {
   result = clazz.arrayClass();
}
```

theresa-m:
fixing error in testng.xml

runtime/vm/BytecodeInterpreter.hpp
```diff
@@ -2721,6 +2721,28 @@ done:;
 		return EXECUTE_BYTECODE;
 	}
 
+	/* java.lang.Class: public native Class<?> arrayType(); */
+	VMINLINE VM_BytecodeAction
+	inlClassArrayType(REGISTER_ARGS_LIST)
+	{
+		VM_BytecodeAction rc = EXECUTE_BYTECODE;
+		j9object_t arrayObject = NULL;
```
gacholio: Local can be better scoped, or removed entirely (assigned and used only on the very next line).

pshipton:
Class implements `TypeDescriptor.OfField<Class<?>>, Constable`
http://cr.openjdk.java.net/~iris/se/12/build/latest/api/java.base/java/lang/Class.html

theresa-m:
extra implements are included here https://github.com/eclipse/openj9/pull/4030

test/functional/Java12andUp/src/org/openj9/test/java_lang/Test_Class.java
```diff
@@ -0,0 +1,130 @@
...
+	@Test(groups = { "level.sanity" })
+	public void testClassDescribeConstable() {
+		describeConstableTestGeneral("testClassDescribeConstable (primitive)", primitiveTest);
+		describeConstableTestGeneral("testClassDescribeConstable (object)", objectTest);
+		describeConstableTestGeneral("testClassDescribeConstable (array)", arrayTest);
+		describeConstableTestGeneral("testClassDescribeConstable (array 2)", arrayTest2);
+	}
+
+	private void describeConstableTestGeneral(String testName, Class<?> testClass) {
+		Optional<ClassDesc> optionalDesc = testClass.describeConstable();
```
DanHeidinga: use `.orElseThrow()` and let the test code throw if the Optional is empty.  The test harness knows how to fail tests for unexpected exceptions.

test/functional/Java12andUp/src/org/openj9/test/java_lang/Test_Class.java
```diff
@@ -0,0 +1,130 @@
...
+ * This test Java.lang.Class API added in Java 12 and later version.
+ *
+ * new methods:
+ * - describeConstable()
+ * - descriptorString()
+ * - arrayType() 
+ * - componentType(): wrapper for Class.getComponentType(), not tested
+ */
+public class Test_Class {
+    public static Logger logger = Logger.getLogger(Test_Class.class);
```
DanHeidinga: tabs vs spaces

test/functional/Java12andUp/src/org/openj9/test/java_lang/Test_Class.java
```diff
@@ -0,0 +1,130 @@
...
+		Optional<ClassDesc> optionalDesc = testClass.describeConstable();
+
+		if (optionalDesc.isPresent()) {
+			ClassDesc desc = optionalDesc.get();
+
+			/* verify that descriptor can be resolved */
+			Class<?> resolvedClass;
+			try {
+				resolvedClass = (Class<?>)desc.resolveConstantDesc(MethodHandles.lookup());
+			} catch(ReflectiveOperationException e) {
```
DanHeidinga: Let this code throw - the test framework knows how to fail tests for this

test/functional/Java12andUp/src/org/openj9/test/java_lang/Test_Class.java
```diff
@@ -0,0 +1,130 @@
...
+	Class<?> arrayTest2 = new double[0][][].getClass();
+
+	/*
+	 * Test Java 12 API Class.descriptorString()
+	 */
+	@Test(groups = { "level.sanity" })
+	public void testClassDescriptorString() {
+		/* primitive type test */
+		logger.debug("testClassDescriptorString: Primitive test result is: " + primitiveTest.descriptorString() + " expected: " + primitiveResult);
+		Assert.assertTrue(primitiveTest.descriptorString().equals(primitiveResult));
```
DanHeidinga: There are a small number of primitives - Z, B, C, S, I, F, D, J, & V.  Seems like we should test them all:

Something like:
```
Object[][] primitiveData = {
	{ boolean.class, "Z" },
	{ byte.class, "B" },
	{ char.class, "C" },
	{ short.class, "S" },
	{ int.class, "I" },
	{ float.class, "F" },
	{ double.class, "D" },
	{ long.class, "J" },
	{ void.class, "V" }
};
for (Object[] params : primitiveData) {
	Assert.assertEquals(((Class<?>)params[0]).descriptorString(), params[1]);
}
```

test/functional/Java12andUp/src/org/openj9/test/java_lang/Test_Class.java
```diff
@@ -0,0 +1,130 @@
...
+
+		/* ObjectType test */
+		logger.debug("testClassDescriptorString: Object test result is: " + objectTest.descriptorString() + " expected: " + objectResult);
+		Assert.assertTrue(objectTest.descriptorString().equals(objectResult));
+
+		/* ArrayType test */
+		logger.debug("testClassDescriptorString: Array test result is: " + arrayTest.descriptorString() + " expected: " + arrayResult);
+		Assert.assertTrue(arrayTest.descriptorString().equals(arrayResult));
```
DanHeidinga: Something similar to the suggestion above can be done for the primitive array cases.... actually, the Object subclasses can be rolled into the same kind of tests

test/functional/Java12andUp/src/org/openj9/test/java_lang/Test_Class.java
```diff
@@ -0,0 +1,130 @@
...
+		arrayTypeTestGeneral("testClassArrayType (object)", objectTest, objectResult);
+		arrayTypeTestGeneral("testClassArrayType (embedded array)", arrayTest, arrayResult);
+	}
+	
+	private void arrayTypeTestGeneral(String testName, Class<?> testClass, String result) {
+		Class<?> array = testClass.arrayType();
+		String arrayString = array.descriptorString();
+		logger.debug(testName + ": Descriptor of component is: " + result + " descriptor of array is: " + arrayString);
+		Assert.assertTrue(arrayString.equals("[" + result));
+	}
```
DanHeidinga: Can you also add a test case for the class of a lambda expression?

```
Runnable r = () -> System.out.println("a");
// This should succeed
ClassDesc desc = r.getClass().describeConstanble();

desc.getDescriptorString() should also succeed but the value returned will be different in each run

but resolveConstantDesc(MethodHandles.lookup()) should fail
```

theresa-m:
I added a helper for the arrayType() method that throws an IllegalArgumentException when the instance class is void.class to replicate the ri's behavior.

jcl/src/java.base/share/classes/java/lang/Class.java
```diff
@@ -4561,20 +4561,77 @@ public boolean isNestmateOf(Class<?> that) {
...
+	 */
 	public Class<?> arrayType() {
-		throw new UnsupportedOperationException("Stub for Java 12 compilation (Jep334)");
+		if (this == void.class) {
+			throw new IllegalArgumentException();
+		}
+		return arrayTypeHelper();
 	}
 
+	private native Class<?> arrayTypeHelper();
```
DanHeidinga: The more typical naming convention is adding `impl` to the method name

theresa-m: I just pushed the rename.

DanHeidinga:
Jenkins compile win jdk11

DanHeidinga:
Jenkins test sanity zlinux jdk12

theresa-m:
sorry that push was a mistake :(

DanHeidinga:
@theresa-m  Is this branch back to being ready to PR test and merge?

theresa-m:
Right now it's dependent on https://github.com/eclipse/openj9/pull/4105 only for the reason that the rest of the Java12andUp xml files are there. I just assumed String would be merged first since it was approved at the time. I can also move that commit with the other test files here so the tests will run properly

theresa-m:
okay I've just added the rest of the test infra. now it won't matter which one goes in first

test/functional/Java12andUp/src/org/openj9/test/java_lang/Test_Class.java
```diff
@@ -0,0 +1,173 @@
...
+
+		/*
+		 * verify that descriptor can be resolved. Otherwise exception will be thrown.
+		 */
+		Class<?> resolvedClass = (Class<?>)desc.resolveConstantDesc(MethodHandles.lookup());
+
+		String originalDescriptor = testClass.descriptorString();
+		String newDescriptor = resolvedClass.descriptorString();
+		logger.debug(testName + ": Descriptor of original class is: " + originalDescriptor + " descriptor of ClassDesc is: " + newDescriptor);
+		Assert.assertTrue(testClass.equals(resolvedClass));
```
theresa-m: one last improvement to this line. previously was comparing strings
```
Assert.assertTrue(originalDescriptor.equals(newDescriptor));
```

DanHeidinga:
Jenkins test sanity zlinux jdk12

theresa-m:
failing tests are related to #4661

DanHeidinga:
Same known failures - `com.ibm.j9.jsr292.api.MethodHandleAPI_defineClass.test_defineClass_ClassBytes_CorruptedArray`

PR #4161, opened 2019-01-02
WIP: [X86] Add AOT support for field watch datablock

cathyzhyi:
When reporting field read/write JIT code needs to pass a data block
containing data like fieldAddress, offset and owning method. This change
makes sure the data block contains correct value for AOT compilation.
Runtime stores are used to store the data into relavant field of data
block. Will add appropriate relocation types for data snippet when
performance becomes a concern in the future.

For relocating owning method, the code should work in new AOT world
which is not completely enabled yet. An AOT compilation exception is
added to abort AOT compilation for field watch for now. Code throwing
the exception in ilgen when generating rdbar/wrtbar can be deleted once
the new AOT is enabled completely.

### DIFF
--- a/runtime/compiler/exceptions/AOTFailure.hpp
+++ b/runtime/compiler/exceptions/AOTFailure.hpp
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2000, 2018 IBM Corp. and others
+ * Copyright (c) 2000, 2019 IBM Corp. and others
  *
  * This program and the accompanying materials are made available under
  * the terms of the Eclipse Public License 2.0 which accompanies this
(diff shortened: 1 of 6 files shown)


runtime/compiler/exceptions/AOTFailure.hpp
```diff
@@ -80,6 +80,15 @@ class AOTSymbolValidationManagerFailure : public virtual TR::CompilationExceptio
    virtual const char* what() const throw() { return "AOT Symbol Validation Manager Failure"; }
    };
 
+/**
+ * Thrown when certain code path would have worked under new AOT but not old AOT.
+ * This exception can be deleted when new AOT is completely enabled.
+ */
+class AOTNoSupportForOldAOTFailure : public virtual TR::CompilationException
+   {
+   virtual const char* what() const throw() { return "This code only works under new AOT"; }
```
dsouzai: Maybe a message like

> Code not supported on AOT without the SVM

might be better.

dsouzai:
I had an offline chat with @cathyzhyi; the better approach we came up with is:

1. Disable Field Watch for AOT (with or without SVM)
2. When using the SVM, relocate the method pointer in the data snippet
3. Enable Field Watch only for AOT with the SVM

This will prevent the need for generating the unresolved snippet unnecessarily under AOT.

andrewcraik:
I hae marked this WIP based on the last comment - @cathyzhyi if this PR is going to be replaced could you close this one?

cathyzhyi:
closed based on comments to rework this part.

Issue #4135, opened 2018-12-20
JIT crash compiling WeakHashMap.put in JDK11


DanHeidinga:
Crashes on multiple platforms: Z & X.  

Failures seen in the `cmdLineTester_jvmtitests_*` test bucket in the gctcti001, gtgc001, and gtgc002 subtests.

Example failure output:
```
Testing: gctcti001
Test start time: 2018/12/20 10:43:29 Eastern Standard Time
Running command: "/home/jenkins/jenkins-agent/workspace/PullRequest-Extended-JDK11-linux_390-64_cmprssptrs-OpenJ9/build/linux-s390x-normal-server-release/images/jdk/bin/java"  -Xcompressedrefs -Xjit -Xgcpolicy:gencon -Xshareclasses:name=bcienabled,enableBCI  -Xdump    -agentlib:jvmtitest=test:gctcti001 -cp "/home/jenkins/jenkins-agent/workspace/PullRequest-Extended-JDK11-linux_390-64_cmprssptrs-OpenJ9/openj9/test/TestConfig/scripts/testKitGen/../../../../jvmtest/functional/cmdLineTests/jvmtitests/jvmtitest.jar" com.ibm.jvmti.tests.util.TestRunner
Time spent starting: 3 milliseconds
Time spent executing: 3667 milliseconds
.....
 [ERR] Unhandled exception
 [ERR] Type=Segmentation error vmState=0x0005ffff
......
[ERR] Method_being_compiled=java/util/WeakHashMap.put(Ljava/lang/Object;Ljava/lang/Object;)Ljava/lang/Object;
```

https://ci.eclipse.org/openj9/job/PullRequest-Sanity-JDK11-linux_x86-64_cmprssptrs-OpenJ9/248/
https://ci.eclipse.org/openj9/job/Test-extended.functional-JDK11-linux_390-64_cmprssptrs/111
https://ci.eclipse.org/openj9/job/Test-sanity.functional-JDK11-osx_x86-64_cmprssptrs/24
https://ci.eclipse.org/openj9/job/Test-sanity.functional-JDK11-linux_x86-64_cmprssptrs/329

DanHeidinga:
FYI @andrewcraik as this is cross-platform.

andrewcraik:
@cathyzhyi since you were last in the JVMTI code fixing another OSR bug this might be related...

pshipton:
Set as a blocker since it blocks omr acceptance builds.

[...]

PR #4164, opened 2019-01-02
Track and check types for SVM symbol IDs

jdmpapin:
This is an attempt to ensure that if types are mismatched due to a bug, we will no longer blindly cast and access memory through a pointer to an unrelated type, which is undefined behaviour.


andrewcraik:

jdmpapin:
The test failures are due to #4135

andrewcraik:
Sanity looks good - failures look like #4135 as noted - merging.

Issue #3635, opened 2018-11-10
Implement ability to double compile a method: remote and locally


mpirvu:
It would be nice if the client could submit a logging compilation request to the JITaaS server and then repeat the same compilation locally, again with logging.
This would allow us to compare the two logs (remote and local) for the purpose of tracking changes that lead to performance gaps.

PR #4167, opened 2019-01-02
Enable double compile of methods, remote and local

harryyu1994:
Allows JITaaS developers to compare remote and local compilations of a method for the purpose of tracking differences that lead to performance gaps. This "double compilation" is activated with `-Xjit:enableJITaaSDoLocalCompilesForRemoteCompiles`
[skip ci]
Depends on: https://github.com/eclipse/omr/pull/3451
Issue: #3635 

### DIFF
--- a/runtime/compiler/control/JITaaSCompilationThread.cpp
+++ b/runtime/compiler/control/JITaaSCompilationThread.cpp
@@ -2242,7 +2242,21 @@ remoteCompile(
 
          TR_ASSERT(!metaData->startColdPC, "coldPC should be null");
 
-         compiler->getOptions()->writeLogFileFromServer(logFileStr);
+         int compilationSequenceNumber = compiler->getOptions()->writeLogFileFromServer(logFileStr);
+         if (TR::Options::getCmdLineOptions()->getOption(TR_EnableJITaaSDoLocalCompilesForRemoteCompiles) && compilationSequenceNumber)
+            {
+            // if compilationSequenceNumber is not 0, vlog from server is enabled
+            // double compile is also controlled by TR_EnableJITaaSDoubleCompile option
+            // try to perform an equivalent local compilation and generate vlog
+            // append the same compilationSequenceNumber in the filename
+            compiler->getOptions()->setLogFileForClientOptions(compilationSequenceNumber);
+            if (compiler->compile() != COMPILATION_SUCCEEDED)
+               {
+               TR_ASSERT(false, "Compiler returned non zero return code %d\n", rtn);
+               compiler->failCompilation<TR::CompilationException>("Compilation Failure");
+               }
+            compiler->getOptions()->closeLogFileForClientOptions();
+            }
 
          if (TR::Options::getVerboseOption(TR_VerboseJITaaS))
             {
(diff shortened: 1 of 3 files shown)


harryyu1994:
Example output vlog with `./java -XX:JITaaSClient '-Xjit:{*}(traceFull,log=trlog)' -version`

trlog.1.2733.66688.20190102.220448.2733  
trlog.1.REMOTE.20190102.220448.2733      
trlog.2.2733.66688.20190102.220448.2733
trlog.2.REMOTE.20190102.220448.2733

The number right after trlog is the compilation sequence number. It's unique to each compilation.
Local and remote compilation logs are produced in separate files with matching compilation sequence number.  The remote log has "REMOTE" in the filename instead of the PID.

runtime/compiler/control/JITaaSCompilationThread.cpp
```diff
@@ -2259,7 +2259,20 @@ remoteCompile(
...
-         compiler->getOptions()->writeLogFileFromServer(logFileStr);
+         int compilationSequenceNumber = compiler->getOptions()->writeLogFileFromServer(logFileStr);
+         if (compilationSequenceNumber)
+            {
+            // if compilationSequenceNumber is not 0, vlog from server is enabled
+            // try to perform an equivalent local compilation and generate vlog
+            // append the same compilationSequenceNumber in the filename
+            compiler->getOptions()->setLogFileForClientOptions(compilationSequenceNumber);
+            if (compiler->compile() != COMPILATION_SUCCEEDED)
+               {
```
mpirvu: I do not understand how the local compilation was triggered.
Do you actually see two compilations for this method taking place?

harryyu1994: `compiler->compile()` I think this command triggers the local compilation, I do see 2 compilation logs generated.

harryyu1994: In this case, we first open the log file with  `setLogFileForClientOptions`, then triggers local compilation with `compiler->compile()`, in the process trees will be written to the log, after the compilation we close the log.

mpirvu: Got it. I missed that compiler->compile().
I didn't think such a solution was possible, but if it works, it works.
I would like this strategy to be done only under some option.

harryyu1994:
Adding a JITaaSDoubleCompile option for this (https://github.com/eclipse/omr/pull/3451).

harryyu1994:
There's an issue with this change at the moment that needs more investigation: Since the local compilation is performed after remote compilation of the same method, does the local compiled code get loaded and effectively overwrite the loaded remote code? I think I need to make sure that it's only for producing a log and should not load code to JITaaS client.

mpirvu:
The omr change that implements the needed option has been delivered a week ago. Has teh change reached openj9-omr? If so we could rely on a merge to bring it in.

harryyu1994:
Yes, it has reached openj9-omr. @ymanton @mpirvu This change can be merged as soon as we update openj9-omr.

ymanton:
Noted.

harryyu1994:
@mpirvu Hi Marius, this can be reviewed and merged.

Issue #3585, opened 2018-11-07
Small reduction in data sent for method tracing


mpirvu:
When the client sends information about a j9class to the JITaaS server it packs the method tracing information into a vector of a pair of boolean values. Because the number of methods may be large for some classes, it may worthwhile to do a better packing by using 2 bits (thus just one byte) per method. Relevant code:
```
JITaaSHelpers::ClassInfoTuple
JITaaSHelpers::packRemoteROMClassInfo(J9Class *clazz, TR_J9VM *fe, TR_Memory *trMemory)
{
...
  for(uint32_t i = 0; i < numMethods; ++i)
      {
      methodTracingInfo.push_back(std::make_tuple(
         fe->isMethodEnterTracingEnabled((TR_OpaqueMethodBlock *) &methodsOfClass[i]),
         fe->isMethodExitTracingEnabled((TR_OpaqueMethodBlock *) &methodsOfClass[i])
         ));
      }
...
}
```

dmitry-ten:
Looking at the VM code, it is identical for both enter and exit tracing.
```
UDATA
jitMethodEnterTracingEnabled(J9VMThread *currentThread, J9Method *method)
{
	J9JavaVM *vm = currentThread->javaVM;

	if (vm->extendedRuntimeFlags & J9_EXTENDED_RUNTIME_METHOD_TRACE_ENABLED) {
		U_8 *pmethodflags = fetchMethodExtendedFlagsPointer(method);
		if (J9_RAS_IS_METHOD_TRACED(*pmethodflags)) {
			return TRUE;
		}
	}
	return FALSE;
}

/**
 * Check if method exit tracing is enabled
 * @param currentThread	The current J9VMThread
 * @param method		The method to be checked if exit tracing is required
 * @return 				non-zero if the method exit tracing is required, otherwise 0 is returned
 */
UDATA
jitMethodExitTracingEnabled(J9VMThread *currentThread, J9Method *method)
{
	J9JavaVM *vm = currentThread->javaVM;

	if (vm->extendedRuntimeFlags & J9_EXTENDED_RUNTIME_METHOD_TRACE_ENABLED) {
		U_8 *pmethodflags = fetchMethodExtendedFlagsPointer(method);
		if (J9_RAS_IS_METHOD_TRACED(*pmethodflags)) {
			return TRUE;
		}
	}
	return FALSE;
}
```
So we can just send one boolean per method, instead of 2.

mpirvu:
[...]

PR #4169, opened 2019-01-02
Combine enter/exit tracing queries into one method

a7ehuo:
TR_J9VMBase::isMethodEnterTracingEnabled() and TR_J9VMBase::isMethodExitTracingEnabled()
are implemented as the same. Combine them into one method
TR_J9VMBase::isMethodTracingEnabled(). The tracing info sent between the JITaas client
and the server is reduced to a vector of uint8_t values instead of a vector of tuples
of bool values.

Issue: #3585

### DIFF
--- a/runtime/compiler/env/VMJ9Server.cpp
+++ b/runtime/compiler/env/VMJ9Server.cpp
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2000, 2018 IBM Corp. and others
+ * Copyright (c) 2000, 2019 IBM Corp. and others
  *
  * This program and the accompanying materials are made available under
  * the terms of the Eclipse Public License 2.0 which accompanies this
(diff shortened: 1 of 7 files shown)


a7ehuo:
@mpirvu Please help take a look at this change. thanks.

runtime/compiler/env/VMJ9Server.cpp
```diff
@@ -167,12 +167,12 @@ TR_J9ServerVM::isMethodEnterTracingEnabled(TR_OpaqueMethodBlock *method)
       auto it = _compInfoPT->getClientData()->getJ9MethodMap().find((J9Method*) method);
       if (it != _compInfoPT->getClientData()->getJ9MethodMap().end())
          {
-         return it->second._isMethodEnterTracingEnabled;
+         return it->second._isMethodTracingEnabled;
```
mpirvu: `isMethodEnterTracingEnabled` and `isMethodExitTracingEnabled` could be deleted and `isMethodTracingEnabled` could be overridden following this pattern. Thus, when somebody calls `isMethodEnterTracingEnabled` the J9Base implementation will be called which in turn calls the virtual implementation of `isMethodTracingEnabled` which knows how to get the information from the client.

a7ehuo:
@mpirvu Changed the status to WIP. Looking at a compilation error that only occurs when building the whole JVM.

a7ehuo:
@mpirvu Clarified my build environment issue. All looks good now. Please help review the latest change. thank you.

PR #4172, opened 2019-01-03
Define a guaranteed ID for null in the SVM

jdmpapin:
This allows ID fields of validation and relocation records to be made optional. Previously, all ID lookups required as a precondition that the mapped value is not null, simply due to the fact that null was not stored in the mappings. Now they will continue to require this by default, so that the lookup does not unexpectedly return null. In cases where the value is optional, callers must pass `SymOptional` to explicitly allow null.

### DIFF
--- a/runtime/compiler/x/codegen/J9AheadOfTimeCompile.cpp
+++ b/runtime/compiler/x/codegen/J9AheadOfTimeCompile.cpp
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2000, 2018 IBM Corp. and others
+ * Copyright (c) 2000, 2019 IBM Corp. and others
  *
  * This program and the accompanying materials are made available under
  * the terms of the Eclipse Public License 2.0 which accompanies this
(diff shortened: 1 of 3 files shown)


runtime/compiler/x/codegen/J9AheadOfTimeCompile.cpp
```diff
@@ -261,12 +261,7 @@ uint8_t *J9::X86::AheadOfTimeCompile::initializeAOTRelocationHeader(TR::Iterated
             TR_OpaqueMethodBlock *method = resolvedMethod->getPersistentIdentifier();
             TR_OpaqueClassBlock *thisClass = guard->getThisClass();
             uint16_t methodID = symValManager->getIDFromSymbol(static_cast<void *>(method));
-            uint16_t receiverClassID = 0;
-            if (relocation->getTargetKind() == TR_InlinedInterfaceMethodWithNopGuard ||
-                relocation->getTargetKind() == TR_InlinedAbstractMethodWithNopGuard)
-               {
-               receiverClassID = symValManager->getIDFromSymbol(static_cast<void *>(thisClass));
-               }
+            uint16_t receiverClassID = symValManager->getIDFromSymbol(static_cast<void *>(thisClass));
```
dsouzai: At first I wasn't sure how this would work, since there is no change to pass in `SymOptional` to the associated `getClassFromID` in the load infra code, but then I saw

https://github.com/eclipse/openj9/blob/2f2d9a6df07350f2045e6f240a5190f41c238eb6/runtime/compiler/runtime/RelocationRecord.cpp#L2230-L2233

so it all makes sense.

dsouzai:
Jenkins copyright check

PR #4174, opened 2019-01-03
Abort AOT load for field watch

cathyzhyi:
When reporting field read/write JIT code needs to pass a data block
containing data like fieldAddress, offset and owning method. Proper AOT
support needs to be implemented.

For now, an AOT compilation exception is added to abort AOT compilation
for field watch. Code throwing the exception in ilgen when generating
rdbar/wrtbar can be changed to check only AOT without SVM once the
proper support is completed.


cathyzhyi:
https://github.com/eclipse/openj9/issues/4173 has been opened to track future AOT work for data block.

andrewcraik:

cathyzhyi:
Some tests are skipped but there is no failing test. I am sure why it shows the build failed.

andrewcraik:
All green and approved - look fine to me too. Merging.

PR #4177, opened 2019-01-04
Fix rdbar/wrtbar X86 evaluators for field watch

cathyzhyi:
Fix a few places in the evaluators for field watch
1. Generate correct sequence for floating point type field
2. Remove the fancy nested outOfLine sequence and ICF because RA cannot
handle them properly.
3. Fix the logic for getting valueNode for indirect wrtbar. Previous
version assumes the second child of wrtbar is the field value to be
written which is incorrect for compressedrefs
4. Stop hacking the evaluator table for wrtbars in
TEMPORARY_initJ9X86TreeEvaluatorTable and use the proper evaluator
table entry

### DIFF
--- a/runtime/compiler/x/codegen/J9TreeEvaluator.cpp
+++ b/runtime/compiler/x/codegen/J9TreeEvaluator.cpp
@@ -603,8 +603,6 @@ static TR::Instruction *generatePrefetchAfterHeaderAccess(TR::Node              
 extern void TEMPORARY_initJ9X86TreeEvaluatorTable(TR::CodeGenerator *cg)
    {
    TR_TreeEvaluatorFunctionPointer *tet = cg->getTreeEvaluatorTable();
-   tet[TR::awrtbar] =               TR::TreeEvaluator::writeBarrierEvaluator;
-   tet[TR::awrtbari] =              TR::TreeEvaluator::writeBarrierEvaluator;
    tet[TR::monent] =                TR::TreeEvaluator::monentEvaluator;
    tet[TR::monexit] =               TR::TreeEvaluator::monexitEvaluator;
    tet[TR::monexitfence] =          TR::TreeEvaluator::monexitfenceEvaluator;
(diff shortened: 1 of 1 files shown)


runtime/compiler/x/codegen/J9TreeEvaluator.cpp
```diff
@@ -12268,50 +12346,72 @@ void generateFillInDataBlockSequenceForUnresolvedField(TR::Node *node, J9Method*
...
-      TR::Node *helperCallNode = TR::Node::createWithSymRef(TR::acall, 2, 2, cpIndexNode, cpAddressNode,
-            cg->comp()->getSymRefTab()->findOrCreateRuntimeHelper(helperIndex, true /* canGCandReturn */, true /*canGCandExcept*/, false /* preservesAllRegisters*/));
-      helperCallNode->incReferenceCount();
-      TR::Register *resultReg = TR::TreeEvaluator::performCall(helperCallNode, false /* isIndirect */, false /* spillFPRegs */, cg);
+      if (is64Bit)
+         {
+         TR::Register *cpAddressReg = cg->allocateRegister();
+         TR::Register *cpIndexReg = cg->allocateRegister();
+         generateRegImm64SymInstruction(MOV8RegImm64, node, cpAddressReg, (uintptrj_t) methodSymbol->getResolvedMethod()->constantPool(), cg->comp()->getSymRefTab()->findOrCreateConstantPoolAddressSymbolRef(methodSymbol), cg);
+         generateRegImm64Instruction(MOV8RegImm64, node, cpIndexReg, (uintptrj_t)symRef->getCPIndex(), cg);
```
0dvictor: `getCPIndex()` has value type `int32_t`, it seems we can safely `MOV8RegImm4` unless it need be zero-extended to a 64-bit value instead of sign-extension.

cathyzhyi: Thanks for the suggestion. Changed accordingly.

andrewcraik:
@0dvictor given the updates could you please continue your review?

0dvictor:
Thank you for reminding. The change looks good to me after the update.

andrewcraik:

cathyzhyi:
Just found one problem with the use of `getDummyLocalMR`, WIP while fixing it.

runtime/compiler/x/codegen/J9TreeEvaluator.cpp
```diff
@@ -12359,73 +12461,135 @@ void generateReportFieldAccessOutlinedInstructions(TR::Node *node, TR::LabelSymb
...
-      generateRegInstruction(PUSHReg, node, valueReg, cg);
-      TR::MemoryReference *mr = generateX86MemoryReference(espReg, 0, cg);
-      generateRegMemInstruction(LEARegMem(), node, valueReferenceReg, mr, cg);
+      valueMR = cg->machine()->getDummyLocalMR(node->getType());
+      if (!valueReg->getRegisterPair())
+         {
+         if (valueReg->getKind() == TR_GPR)
+            {
+            TR::AutomaticSymbol *autoSymbol = valueMR->getSymbolReference().getSymbol()->getAutoSymbol();
+            generateMemRegInstruction(SMemReg(autoSymbol->getRoundedSize() == 8), node, valueMR, valueReg, cg);
```
cathyzhyi: fixed the instruction storing to the `dummyLocal` to use the correct size.

cathyzhyi:
fixed the `getDummyLocalMR` problem.

andrewcraik:

andrewcraik:
@0dvictor can you take another look?

0dvictor:
The updated version looks good to me.

runtime/compiler/x/codegen/J9TreeEvaluator.cpp
```diff
@@ -1041,6 +1039,86 @@ TR::Register* J9::X86::TreeEvaluator::performHeapLoadWithReadBarrier(TR::Node* n
...
+      // iwrtbar f
+      //    aload O
+      //    l2i
+      //      a2l
+      //        value  <- sourceObject
+      //
+      TR::Node *translatedNode = sourceObject;
+      if (translatedNode->getOpCode().isConversion())
+         translatedNode = translatedNode->getFirstChild();
+      if (translatedNode->getOpCode().isRightShift()) // optional
```
andrewcraik: does this need a check that there is a compressed ref shift amount - otherwise might the address be generated from some other shift if the address was not a collected reference - wrtbars do work on collected references atm, but at the very least an assert?

runtime/compiler/x/codegen/J9TreeEvaluator.cpp
```diff
@@ -1041,6 +1039,86 @@ TR::Register* J9::X86::TreeEvaluator::performHeapLoadWithReadBarrier(TR::Node* n
...
+   }
+
+/*
+ * Return the j9class of the static field for the direct wrtbar
+ */
+static TR::Node *getDirectWrtbarJ9ClassNode(TR::Node *node)
+   {
+   TR_ASSERT(!node->getOpCode().isIndirect() && node->getOpCode().isWrtBar(), "getDirectWrtbarJ9ClassChild expects direct wrtbar nodes only n%dn (%p)\n", node->getGlobalIndex(), node);
+   if (TR::Compiler->cls.classesOnHeap())
+      return node->getSecondChild()->getFirstChild(); // the direct second child is JavaLangClass which is loaded from the j9class
```
andrewcraik: should there be asserts about the symrefs on these nodes for safety?

runtime/compiler/x/codegen/J9TreeEvaluator.cpp
cathyzhyi: Added assertion to check the shift amount equal to `om.compressedReferenceShiftOffset`.

runtime/compiler/x/codegen/J9TreeEvaluator.cpp
cathyzhyi: Added assert to check the symbol on the second child and its child are as expected.

andrewcraik:
Jenkins tests sanity xlinux,win jdk8,jdk11

andrewcraik:

cathyzhyi:
The failures are unrelated.

andrewcraik:

PR #4178, opened 2019-01-04
Cache repeatedly fetched values in SVM

jdmpapin:
Currently the symbol validation manager (SVM) contains highly repetitive code that repeatedly fetches the compilation, thread, front-end, memory, and class hierarchy table objects. With this commit, these objects will now be fetched only once, and (more importantly) only in one place.


dsouzai:

dsouzai:
xlinux jdk11 build failure due to https://github.com/eclipse/openj9/issues/4166

PR #4179, opened 2019-01-04
Implement X86 codegen support for dynamic array clone

yanluo7:
Dynamic array cloning transforms refArrayObject.clone()
into `anewarray` followed by `arraycopy`, where refArrayObject
is a resolved, not fixed class. For code gen, It requires the
class (2nd) child of anewarray to be runtime evaluated instead
of compile time constant, as actual refArrayObject type is not
known at compiled time (other than it's an ref array). This
change implements support in X86 codegen to handle a runtime
evaluated class child of anewarray.

This is prerequisite for the optimizer change that performs the
object.clone() call transformation.

### DIFF
--- a/runtime/compiler/compile/J9Compilation.cpp
+++ b/runtime/compiler/compile/J9Compilation.cpp
@@ -648,8 +648,15 @@ J9::Compilation::canAllocateInline(TR::Node* node, TR_OpaqueClassBlock* &classIn
    else if (node->getOpCodeValue() == TR::anewarray)
       {
       classRef      = node->getSecondChild();
-      classSymRef   = classRef->getSymbolReference();
 
+      // In the case of dynamic array allocation, return 0 indicating variable dynamic array allocation
+      if (classRef->getOpCodeValue() != TR::loadaddr)
+         {
+         classInfo = NULL;
+         return 0;
+         }
+
+      classSymRef   = classRef->getSymbolReference();
       // Can't skip the allocation if the class is unresolved
       //
       clazz = self()->fej9vm()->getClassForAllocationInlining(self(), classSymRef);
--- a/runtime/compiler/x/codegen/J9TreeEvaluator.cpp
+++ b/runtime/compiler/x/codegen/J9TreeEvaluator.cpp
@@ -6558,6 +6558,7 @@ static void genInitObjectHeader(TR::Node             *node,
                                 TR::Register         *objectReg,
                                 TR::Register         *tempReg,
                                 bool                 isZeroInitialized,
+                                bool                 isDynamicAllocation,
                                 TR::CodeGenerator    *cg)
    {
    TR::Compilation *comp = cg->comp();
@@ -6567,12 +6568,7 @@ static void genInitObjectHeader(TR::Node             *node,
                            (!TR::Compiler->om.generateCompressedObjectHeaders() ||
                             (cg->comp()->compileRelocatableCode() && cg->comp()->getOption(TR_UseSymbolValidationManager)));
 
-   // This code was moved to this point so that the romClass can be used by the AOT
-   // portion without calling the method again.
-   //
-   J9ROMClass *romClass = 0;
-   TR_ASSERT(clazz, "Cannot have a null OpaqueClassBlock\n");
-   romClass = TR::Compiler->cls.romClassOf(clazz);
+   TR_ASSERT((isDynamicAllocation || clazz), "Cannot have a null clazz while not doing dynamic array allocation\n");
 
    // --------------------------------------------------------------------------------
    //
@@ -6584,8 +6580,16 @@ static void genInitObjectHeader(TR::Node             *node,
 
    TR::Register * clzReg = classReg;
 
+   // For dynamic array allocation, load the array class from the component class and store into clzReg
+   if (isDynamicAllocation)
+      {
+      TR_ASSERT((node->getOpCodeValue() == TR::anewarray), "Dynamic allocation currently only supports reference arrays");
+      TR_ASSERT(classReg, "must have a classReg for dynamic allocation");
+      clzReg = tempReg;
+      generateRegMemInstruction(LRegMem(), node, clzReg, generateX86MemoryReference(classReg, offsetof(J9Class, arrayClass), cg), cg);
+      }
    // TODO: should be able to use a TR_ClassPointer relocation without this stuff (along with class validation)
-   if (cg->needClassAndMethodPointerRelocations() && !comp->getOption(TR_UseSymbolValidationManager))
+   else if (cg->needClassAndMethodPointerRelocations() && !comp->getOption(TR_UseSymbolValidationManager))
       {
       TR::Register *vmThreadReg = cg->getVMThreadRegister();
       if (node->getOpCodeValue() == TR::newarray)
@@ -6599,13 +6603,6 @@ static void genInitObjectHeader(TR::Node             *node,
          generateMemRegInstruction(opSMemReg, node,
              generateX86MemoryReference(objectReg, TR::Compiler->om.offsetOfObjectVftField(), cg),
              tempReg, cg);
-         clzReg = tempReg;
-         }
-      else if (node->getOpCodeValue() == TR::anewarray)
-         {
-         TR_ASSERT(classReg, "must have a classReg for TR::anewarray in AOT mode");
-         generateRegMemInstruction(LRegMem(), node, tempReg,
-             generateX86MemoryReference(classReg, offsetof(J9Class, arrayClass), cg), cg);
          clzReg = tempReg;
          }
       else
@@ -6661,6 +6658,7 @@ static void genInitObjectHeader(TR::Node             *node,
 
 #ifndef J9VM_INTERP_FLAGS_IN_CLASS_SLOT
    // Enable macro once GC-Helper is fixed
+   J9ROMClass *romClass = TR::Compiler->cls.romClassOf(clazz);
    if (romClass)
       {
       orFlags |= romClass->instanceShape;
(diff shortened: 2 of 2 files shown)


yanluo7:
The desired tree after transformation looks like following: 
Note the `aloadi  <componentClass>`  child of `anewarray` node where we obtain the component class from the class of the receiver array object of `object.clone()` :

```
treetop ()
  anewarray  jitANewArray
    iloadi  <contiguous-array-size>
      ==>l2a (array object to be cloned)
    aloadi  <componentClass>
      ==>aloadi (vftload of array object)
treetop ()
  arraycopy  <arraycopy>
    ==>l2a
    ==>anewarray
    aladd
      ==>l2a
      lconst 8
    aladd 
      ==>anewarray
      ==>lconst 8
    lshl 
       i2l 
         ==>iloadi
       iconst 2 
```

runtime/compiler/x/codegen/J9TreeEvaluator.cpp
```diff
@@ -6567,12 +6567,11 @@ static void genInitObjectHeader(TR::Node             *node,
...
-   // portion without calling the method again.
-   //
-   J9ROMClass *romClass = 0;
-   TR_ASSERT(clazz, "Cannot have a null OpaqueClassBlock\n");
-   romClass = TR::Compiler->cls.romClassOf(clazz);
+   // Dynamic array cloning is true when the class (2nd) child node of the anewarray node is a component class load from
+   // the original array object (array object to be copied). i.e. we are loading the component class from the original array class
+   // and pass the component class as 2nd child of anewarray. So the 2nd child must be evaluated into a reg.
+   bool dynamicArrayClone = (node->getOpCodeValue() == TR::anewarray) && (node->getSecondChild()->getSymbolReference()) &&
+         (node->getSecondChild()->getSymbolReference() == comp->getSymRefTab()->findArrayComponentTypeSymbolRef());
```
andrewcraik: this seems a little optimistic - shouldn't we say that if it isn't a loadaddr (or possibly aconst) then it is dynamic and assert that we have clazz if the array is not dynamic?

runtime/compiler/x/codegen/J9TreeEvaluator.cpp
```diff
@@ -6682,23 +6688,42 @@ static void genInitObjectHeader(TR::Node             *node,
...
-         if (initReservable)
-            initialLwValue = OBJECT_HEADER_LOCK_RESERVED;
+         bool initLw = (node->getOpCodeValue() != TR::New) || initReservable;
+         int lwOffset = fej9->getByteOffsetToLockword(clazz);
+         if (lwOffset == -1)
+            initLw = false;
 
-         TR_X86OpCodes op = (TR::Compiler->target.is64Bit() && fej9->generateCompressedLockWord()) ? S4MemImm4 : SMemImm4();
-         generateMemImmInstruction(op, node, generateX86MemoryReference(objectReg, lwOffset, cg), initialLwValue, cg);
+         if (initLw)
```
andrewcraik: I think some comments about lock word handling are warranted along with a description of the interaction with lock reservation

runtime/compiler/x/codegen/J9TreeEvaluator.cpp
yanluo7: Yes, added extra condition that checks for the class node op code for dynamic array, also assertion that rensure clazz cannot be null when not dynamic array.

runtime/compiler/x/codegen/J9TreeEvaluator.cpp
yanluo7: Added more comments to address lock reservation case.

andrewcraik:
@0dvictor could I get your thoughts on this please?

runtime/compiler/x/codegen/J9TreeEvaluator.cpp
```diff
@@ -6682,23 +6691,44 @@ static void genInitObjectHeader(TR::Node             *node,
...
-      int lwOffset = fej9->getByteOffsetToLockword(clazz);
-      if (lwOffset == -1)
-         initLw = false;
-
-      if (initLw)
+      TR::LabelSymbol *doneLabel = generateLabelSymbol(cg);
+#if defined (J9VM_THR_LOCK_NURSERY)
+      generateRegMemInstruction(LRegMem(), node, tempReg, generateX86MemoryReference(clzReg, offsetof(J9ArrayClass, lockOffset), cg), cg);
+#else
+      generateRegImmInstruction(MOVRegImm4(), node, tempReg, TMP_OFFSETOF_J9OBJECT_MONITOR, cg);
```
0dvictor: In this case, `tempReg` holds a compile-time constant, so that the branching logic below can be fold away.

runtime/compiler/x/codegen/J9TreeEvaluator.cpp
```diff
@@ -6682,23 +6691,44 @@ static void genInitObjectHeader(TR::Node             *node,
...
+#endif
+      generateRegImmInstruction(CMPRegImm4(), node, tempReg, (int32_t)-1, cg);
+      generateLabelInstruction (JE4, node, doneLabel, true, cg);
```
0dvictor: `generateLabelInstruction` with `needsVMThreadRegister` parameter has been deprecated; hence this line should be:
```generateLabelInstruction (JE4, node, doneLabel, cg); ```

File: runtime/compiler/x/codegen/J9TreeEvaluator.cpp [hunk 5]
```diff
@@ -6682,23 +6691,44 @@ static void genInitObjectHeader(TR::Node             *node,
...
+      generateRegMemInstruction(LEARegMem(), node, tempReg, generateX86MemoryReference(objectReg, tempReg, 0, cg) ,cg);
+      TR_X86OpCodes op = (TR::Compiler->target.is64Bit() && fej9->generateCompressedLockWord()) ? S4MemImm4 : SMemImm4();
```
0dvictor: This logic may be simplified as:
```SMemImm4(TR::Compiler->target.is64Bit() && !fej9->generateCompressedLockWord())```

File: runtime/compiler/x/codegen/J9TreeEvaluator.cpp [hunk 6]
```diff
@@ -6682,23 +6691,44 @@ static void genInitObjectHeader(TR::Node             *node,
...
 
-         TR_X86OpCodes op = (TR::Compiler->target.is64Bit() && fej9->generateCompressedLockWord()) ? S4MemImm4 : SMemImm4();
-         generateMemImmInstruction(op, node, generateX86MemoryReference(objectReg, lwOffset, cg), initialLwValue, cg);
+         if (initLw)
+            {
+            int32_t initialLwValue = 0;
+            if (initReservable)
+               initialLwValue = OBJECT_HEADER_LOCK_RESERVED;
+
+            TR_X86OpCodes op = (TR::Compiler->target.is64Bit() && fej9->generateCompressedLockWord()) ? S4MemImm4 : SMemImm4();
```
0dvictor: This logic may be simplified as:
SMemImm4(TR::Compiler->target.is64Bit() && !fej9->generateCompressedLockWord())

runtime/compiler/x/codegen/J9TreeEvaluator.cpp
```diff
@@ -6643,6 +6645,12 @@ static void genInitObjectHeader(TR::Node             *node,
       }
    else
       {
+      // For dynamic array clone, load the array class from the component class and store into clzReg
+      if (dynamicArrayClone)
+         {
+         clzReg = tempReg;
+         generateRegMemInstruction(LRegMem(), node, clzReg, generateX86MemoryReference(classReg, offsetof(J9Class, arrayClass), cg), cg);
+         }
```
jdmpapin: As it stands, this will generate the load twice if dynamic array clone occurs in "old" AOT. Could it be combined with the path that generates the corresponding load above? e.g.
```
if (classReg != NULL && node->getOpCodeValue() == TR::anewarray)
   {
   // Variable component type, due either to dynamicArrayClone or (non-SVM) AOT.
   // Load the array type from it at runtime.
   clzReg = tempReg;
   generateRegMemInstruction(LRegMem(), node, clzReg, generateX86MemoryReference(classReg, offsetof(J9Class, arrayClass), cg), cg);
   }
else if (cg->needClassAndMethodPointerRelocations() && !comp->getOption(TR_UseSymbolValidationManager))
   {
   // ... [check for TR::newarray and TR::New, since TR::anewarray has been ruled out] ...
   }
```

runtime/compiler/x/codegen/J9TreeEvaluator.cpp
jdmpapin: This still seems pretty special-purpose. Do we need to require that the component type child be either constant or in this restricted form? Or could we use something like the following?
```
node->getOpCodeValue() == TR::anewarray && clazz == NULL
```
I think `canAllocateInline()` could set `clazz = NULL` whenever the child isn't `loadaddr`.

runtime/compiler/x/codegen/J9TreeEvaluator.cpp
```diff
@@ -6682,23 +6691,44 @@ static void genInitObjectHeader(TR::Node             *node,
...
+#if defined (J9VM_THR_LOCK_NURSERY)
+      generateRegMemInstruction(LRegMem(), node, tempReg, generateX86MemoryReference(clzReg, offsetof(J9ArrayClass, lockOffset), cg), cg);
+#else
+      generateRegImmInstruction(MOVRegImm4(), node, tempReg, TMP_OFFSETOF_J9OBJECT_MONITOR, cg);
+#endif
+      generateRegImmInstruction(CMPRegImm4(), node, tempReg, (int32_t)-1, cg);
+      generateLabelInstruction (JE4, node, doneLabel, true, cg);
+      generateRegMemInstruction(LEARegMem(), node, tempReg, generateX86MemoryReference(objectReg, tempReg, 0, cg) ,cg);
+      TR_X86OpCodes op = (TR::Compiler->target.is64Bit() && fej9->generateCompressedLockWord()) ? S4MemImm4 : SMemImm4();
+      generateMemImmInstruction(op, node, generateX86MemoryReference(tempReg, 0, cg), 0, cg);
```
jdmpapin: Can we get away without the `lea` just above, and generate
```
mov ... ptr [rObject + rTemp], 0
```
instead of
```
lea rTemp, [rObject + rTemp]
mov ... ptr [rTemp], 0
```
?

runtime/compiler/x/codegen/J9TreeEvaluator.cpp
yanluo7: I revised the logic for evaluating the `isDynamicArray` boolean. Now we only initialize the boolean once in the top level NewEvaluator using the return values from `canAllocateInline()`. This way checking the `ArrayComponentType` symbol occurs only once in `canAllocateInline()`. And the NewEvaluator passes this flag down to helper routines which only need to check the flag to do the right thing. Currently the boolean is named as array-specific since only dynamic arrays are supported. When we implement dynamic object we can rename it properly -- just `isDynamicAllocation`.

andrewcraik:
@yanluo7 could you comment on what has changed and respond to the review questions above so we can continue reviewing this change?

yanluo7:
@andrewcraik yes, forgot to hit `submit`

andrewcraik:
@0dvictor and @jdmpapin could you review the latest changes to see if they address your comments on this item previously? thanks!

runtime/compiler/x/codegen/J9TreeEvaluator.cpp
```diff
@@ -7293,7 +7314,7 @@ J9::X86::TreeEvaluator::VMnewEvaluator(
    TR_OpaqueClassBlock *clazz      = NULL;
    TR::Register        *classReg   = NULL;
    bool                 isArrayNew = false;
-
+   bool     dynamicArrayAllocation = false;
```
jdmpapin: Since this isn't used between here and the real initialization below, would you mind moving the declaration down? Right now it looks like something might read this `false`

runtime/compiler/x/codegen/J9TreeEvaluator.cpp
yanluo7: Fixed

runtime/compiler/x/codegen/J9TreeEvaluator.cpp
```diff
@@ -7697,6 +7723,7 @@ J9::X86::TreeEvaluator::VMnewEvaluator(
             dataOffset,
             tempReg,
             monitorSlotIsInitialized,
+            dynamicArrayAllocation,
```
jdmpapin: If you pass `true` here, then the path in `genInitObjectHeader()` that loads the array type for "old" AOT can be deleted, since we'll start using the path that does it for `dynamicArrayAllocation` instead. Personally I think that would be more straightforward because effectively what old AOT is doing here is forcing the allocation to be dynamic

runtime/compiler/compile/J9Compilation.cpp
```diff
@@ -650,6 +650,14 @@ J9::Compilation::canAllocateInline(TR::Node* node, TR_OpaqueClassBlock* &classIn
       classRef      = node->getSecondChild();
       classSymRef   = classRef->getSymbolReference();
 
+      // In the case of dynamic array cloning, second child is the array component type load,
+      // return 0 indicating variable dynamic array allocation
+      if (classSymRef && (classRef->getOpCodeValue() == TR::aloadi) && (classSymRef == self()->getSymRefTab()->findArrayComponentTypeSymbolRef()))
```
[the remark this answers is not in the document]
yanluo7: More generic, done.

runtime/compiler/x/codegen/J9TreeEvaluator.cpp
```diff
@@ -6584,8 +6580,14 @@ static void genInitObjectHeader(TR::Node             *node,
 
    TR::Register * clzReg = classReg;
 
+   // For dynamic array allocation, load the array class from the component class and store into clzReg
+   if (isDynamicAllocation)
+      {
+      clzReg = tempReg;
```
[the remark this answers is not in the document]
yanluo7: Done

runtime/compiler/x/codegen/J9TreeEvaluator.cpp
yanluo7: Done

runtime/compiler/x/codegen/J9TreeEvaluator.cpp
yanluo7: Good point.

yanluo7:
@andrewcraik Victor and Devin have approved the change. If you're OK with it, this can be merged.

andrewcraik:

PR #4185, opened 2019-01-07
Query node property to catch loads in runFEMacro

cathyzhyi:
Change the switch to queries on node property to catch rdbars and do the
intented transformation when folding bytecodes in runFEMacro.


andrewcraik:

PR #4186, opened 2019-01-07
Fix for client crash when killing server

satbir-ca:
In stopcompilationthread the dummy message is send to server
for cleaning the client data session. If the server is killed
before that the dummy messagei, send will fail and an exception
is generated. Adding the catch for the exception in the
stopcompilationThread to avoid crash.

### DIFF
--- a/runtime/compiler/control/CompilationThread.cpp
+++ b/runtime/compiler/control/CompilationThread.cpp
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2000, 2018 IBM Corp. and others
+ * Copyright (c) 2000, 2019 IBM Corp. and others
  *
  * This program and the accompanying materials are made available under
  * the terms of the Eclipse Public License 2.0 which accompanies this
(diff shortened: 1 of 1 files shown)


satbir-ca:

runtime/compiler/control/CompilationThread.cpp
```diff
@@ -3351,23 +3351,30 @@ void TR::CompilationInfo::stopCompilationThreads()
...
+         uint32_t romMethodOffset = 0;
+         JITaaS::J9ClientStream client(getPersistentInfo());
+         client.buildCompileRequest(getPersistentInfo()->getJITaaSId(), romMethodOffset,
                      ramMethod, clazz, cold, detailsStr, J9::EMPTY,
                      unloadedClasses, classTuple, optionsStr, recompMethodInfoStr, seqNo);
-
+         }
+      catch (const JITaaS::StreamFailure &e)
+         {
+         if (TR::Options::getVerboseOption(TR_VerboseJITaaS))
```
mpirvu: A more descriptive message detailed this specific case would be helpful here.

satbir-ca:
updated

PR #4187, opened 2019-01-07
Health check information as one block

satbir-ca:
Health check involve multiple lines. Printing with
writeLineLock will cause them to interleave with
print from other threads. Using VlogAcquire and
VlogRelease will make the information one block.


satbir-ca:

satbir-ca:
updated

PR #4189, opened 2019-01-07
Declare and implement helper method J9ClassEnv isEnumClass

yanluo7:
Add helper method isEnumClass in J9ClassEnv, which returns if a
given class is an Enum class or not. This API will be used in
Value Propagation code that optimizes j/l/Class.getEnumConstants
path.


andrewcraik:

PR #4191, opened 2019-01-07
Improve robustness of checking tags of allocated memory blocks

keithc-ca:
J9MemTag.allocSize has type uintptr_t (aka UDATA), which is mapped to U64 by MSVC on Windows (or with the changes in eclipse/omr#3355 on all 64-bit platforms). A random address treated as a pointer to a J9MemTag has a 50% chance of referring to a J9MemTag where allocSize >= 2^63 which would trigger an InvalidDataTypeException from U64.longValue(). This change modifies ROUNDED_FOOTER_OFFSET() so it always returns a UDATA object independent of the actual
argument type, thus avoiding that exception for blocks that otherwise would fail the sanity checks.

https://github.com/eclipse/omr/pull/3422 replays eclipse/omr#3355.

### DIFF
--- a/debugtools/DDR_VM/src/com/ibm/j9ddr/vm29/pointer/helper/J9MemTagHelper.java
+++ b/debugtools/DDR_VM/src/com/ibm/j9ddr/vm29/pointer/helper/J9MemTagHelper.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2010, 2014 IBM Corp. and others
+ * Copyright (c) 2010, 2019 IBM Corp. and others
  *
  * This program and the accompanying materials are made available under
  * the terms of the Eclipse Public License 2.0 which accompanies this
@@ -72,9 +72,8 @@ public class J9MemTagHelper
 	 */
 	public static IDATA j9mem_check_tags(VoidPointer memoryPointer, long headerEyecatcher, long footerEyecatcher) throws J9MemTagCheckError 
 	{
-		J9MemTagPointer headerTagAddress, footerTagAddress = J9MemTagPointer.NULL;
-
-		headerTagAddress = j9mem_get_header_tag(memoryPointer);
+		J9MemTagPointer headerTagAddress = j9mem_get_header_tag(memoryPointer);
+		J9MemTagPointer footerTagAddress = J9MemTagPointer.NULL;
 
 		try {
 			footerTagAddress = j9mem_get_footer_tag(headerTagAddress);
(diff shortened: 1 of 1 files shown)


keithc-ca:
Jenkins test all zlinux jdk11 depends eclipse/omr#3422

debugtools/DDR_VM/src/com/ibm/j9ddr/vm29/pointer/helper/J9MemTagHelper.java
```diff
@@ -266,8 +265,9 @@ public static VoidPointer j9mem_get_footer_padding(J9MemTagPointer headerEyeCatc
 	
 	public static UDATA ROUNDED_FOOTER_OFFSET(UDATA number) 
 	{
-		UDATA a = (number.add(ROUNDING_GRANULARITY - 1)).add(J9MemTag.SIZEOF);
-		UDATA b = new UDATA(ROUNDING_GRANULARITY - 1).bitNot();
+		UDATA mask = new UDATA(ROUNDING_GRANULARITY - 1);
+		UDATA a = mask.add(number).add(J9MemTag.SIZEOF);
+		UDATA b = mask.bitNot();
```
DanHeidinga: Based on the PR description
> This change modifies ROUNDED_FOOTER_OFFSET() so it always returns a UDATA object independent of the actual
argument type, thus avoiding that exception for blocks that otherwise would fail the sanity checks.

I'm not clear on why the original code wouldn't have already always returned a UDATA.  Can you explain why this change is required?

keithc-ca: U64.add() returns a U64 while UDATA.add(UScalar) returns UDATA; UDATA.longValue() doesn't throw an exception.

DanHeidinga: Which variable is a U64 here? `J9MemTag.SIZEOF`?

pshipton:
Jenkins test all win jdk11 depends eclipse/omr#3422

pshipton:
jenkins test sanity win jdk11 depends eclipse/omr#3422

pshipton:
jenkins test extended win jdk11 depends eclipse/omr#3422

debugtools/DDR_VM/src/com/ibm/j9ddr/vm29/pointer/helper/J9MemTagHelper.java
DanHeidinga: Thanks Keith.  I had forgotten that `U64` was changed to be a subclass of `UDATA`

PR #4209, opened 2019-01-08
cmake: explicitly add OMR_PLATFORM compile options to TR

rwy0717:
For the compiler, we're blowing away the cmake build flags, to start
with a fresh set of options. In doing so, we're removing the flags
installed by OmrPlatform, which are required on every target.

This PR adds those missing options back in to the compiler targets, and
fixes a recent OMR integration issue where some options were moved from
TR to the OmrPlatform variables.


pshipton:
Set as a blocker since this is blocking omr acceptance builds.

Note the copyright check failed.

PR #4218, opened 2019-01-09
Fix InnerPreexistence test bug

### LINKED ISSUE #4217: test/functional/JIT_Test/src/jit/test/tr/chtable/InnerPreexistence.java might never complete
The core test method looks like this:

        @Test
        public void testInnerPreexistence() {
                timer.schedule(new NotifyTask(TASK_LOAD_C, this), 8000);
                while (keepOnGoing)
                        foo();
                AssertJUnit.assertTrue(foo() == 'c');
        }

So a NotifyTask is being scheduled with a 8 seconds delay and a loop is entered based on `keepOnGoing`. Looking at wakeUp we see:

        //Should not be run as a test, therefore it is disabled
        @Test(enabled = false)
        public void wakeUp(int eventId) {
                someB = new C();
                dummySync(); // to ensure correct ordering on weak memory systems
                keepOnGoing = false;
        }

And dymmySync() is as follows:

       private synchronized void dummySync() {}

There seems to be an incorrect assumption that a synchronized method has some effect visibility of `keepOnGoing`. This test might spin forever because the change to `keepOnGoing` will never be visible:

        while (keepOnGoing)
             foo();

In fact, this can be reproduced on a hotspot JVM as follows:

    $ git clone https://github.com/jerboaa/inner-preexisting-reproducer.git
    $ cd inner-preexisting-reproducer
    $ javac InnerPreexistence.java
    $ java -showversion InnerPreexistence
    openjdk version "1.8.0_191"
    OpenJDK Runtime Environment (build 1.8.0_191-b12)
    OpenJDK 64-Bit Server VM (build 25.191-b12, mixed mode)

     DEBUG: Waking up: 1
     <hangs forever>

jerboaa:
Declare keepOnGoing volatile which enforces re-reading of the
variable upon a write.

Closes #4217

### DIFF
--- a/test/functional/JIT_Test/src/jit/test/tr/chtable/InnerPreexistence.java
+++ b/test/functional/JIT_Test/src/jit/test/tr/chtable/InnerPreexistence.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2017, 2018 IBM Corp. and others
+ * Copyright (c) 2017, 2019 IBM Corp. and others
  *
  * This program and the accompanying materials are made available under
  * the terms of the Eclipse Public License 2.0 which accompanies this
@@ -37,7 +37,7 @@ public class InnerPreexistence implements Notifiable {
 	public static final int TASK_LOAD_C = 1;
 
 	private Timer timer;
-	private boolean keepOnGoing = true;
+	private volatile boolean keepOnGoing = true;
 	private A someA;
 	private B someB;
 	private B otherB;


jerboaa:
@smlambert Thanks for the review! I've updated the copyright. I work for Red Hat, isn't the ECA covered by some company agreement?

smlambert:
To answer the question on the ECA, I believe everyone needs to sign it (I know we all had to).  I am not aware that there is a company agreement that overrides it.

smlambert:
ip-validation check still fails, I think is expecting a "Signed-off-by: your name <youremail>" at the end of the commit comment

jerboaa:
OK, trying that now.

jerboaa:
@smlambert Thanks! Seems to have worked. Any more comments?

smlambert:
https://ci.eclipse.org/openj9/view/Test/job/Test-Grinder/252/

PR #4226, opened 2019-01-10
Generate ProfiledClassRecord for new classes only

jdmpapin:
A `ProfiledClassRecord` for a class that has already been determined causes the validation process to choose arbitrarily from the classes that match the given class chain, and then to verify that this new choice agrees with the existing one. Eliminating the `ProfiledClassRecord` in this case saves space, and causes the validation process to behave as though it always makes the correct choice.


dsouzai:

PR #3995, opened 2018-12-11
Re-enable DDR for AIX

keithc-ca:
* restore use of `-g` on AIX except for interpreter source files
* re-enable DDR testing on AIX

Depends on:
* https://github.com/ibmruntimes/openj9-openjdk-jdk8/pull/223
* https://github.com/ibmruntimes/openj9-openjdk-jdk11/pull/76
* https://github.com/ibmruntimes/openj9-openjdk-jdk/pull/59

Closes #1769 

Doc issue https://github.com/eclipse/openj9-docs/issues/92

Issue #4120, opened 2018-12-19
aix build machine compile issues


pshipton:
https://ci.eclipse.org/openj9/job/PullRequest-Compile-JDK8-aix_ppc-64_cmprssptrs-OpenJ9/42/
aix71-p8-4
https://ci.eclipse.org/openj9/job/PullRequest-Compile-JDK8-aix_ppc-64_cmprssptrs-OpenJ9/44
aix71-p8-1
```
11:49:28 Creating images/lib/jconsole.jar
11:50:03 gmake[2]: *** virtual memory exhausted.  Stop.
11:50:03 gmake[2]: *** Waiting for unfinished jobs....
11:50:03 gmake[2]: *** virtual memory exhausted.  Stop.
11:50:03 gmake[1]: *** [BuildJdk.gmk:106: images] Error 2
11:50:03 gmake[1]: INTERNAL: Exiting with 6 jobserver tokens available; should be 8!
11:50:03 make: *** [/home/u0020236/workspace/PullRequest-Compile-JDK8-aix_ppc-64_cmprssptrs-OpenJ9//make/Main.gmk:141: images-only] Error 2
```

https://ci.eclipse.org/openj9/job/PullRequest-Compile-JDK8-aix_ppc-64_cmprssptrs-OpenJ9/43
aix71-p8-2
```
13:23:14 Building lib/uma.jar
13:23:16 J9VM version: ad53cf012
13:23:16 #
13:23:16 # A fatal error has been detected by the Java Runtime Environment:
13:23:16 #
13:23:16 #  SIGBUS (0xa) at pc=0x000000000000e898, pid=9961620, tid=258
13:23:16 #
13:23:16 # JRE version:  (7.0) (build )
13:23:16 # Java VM: OpenJDK 64-Bit Server VM (24.95-b00 mixed mode aix-ppc64 compressed oops)
13:23:16 # Problematic frame:
13:23:16 # C  0x000000000000e898
13:23:16 #
13:23:16 # Core dump written. Default location: /home/u0020236/workspace/PullRequest-Compile-JDK8-aix_ppc-64_cmprssptrs-OpenJ9/build/aix-ppc64-normal-server-release/vm/core or core.9961620
```

pshipton:
@AdamBrousseau

AdamBrousseau:
Heppened again, disabled aix71-p8-2

pshipton:
[...]

PR #4229, opened 2019-01-10
AIX: strip symbols and debugging information from shared libraries

keithc-ca:
Even though core files seem to have similar sizes, `dtfj` often takes much longer to open an AIX core file created by a VM with unstripped shared libraries.

#3995 added the need to start stripping the shared libraries.


pshipton:
jenkins test sanity aix jdk8,jdk11

pshipton:
I expect the jdk8 build to fail due to #4120, although it would be interesting if it doesn't fail due to this change.

PR #4230, opened 2019-01-10
Default SubSpace active size API

amicic:
Simplify usage of  getActiveMemorySize() against default MemorySubSpace.
Since default subspace can be either Tenure/OLD or Nursery/NEW, now any
callers that relied on getActiveMemorySize(OLD | NEW) could just simply
call getActiveMemorySize() which is the total size of Subspace (for
Nursery it specifically mean both Allocate and Survivor)


amicic:
Dependent on https://github.com/eclipse/omr/pull/3436

dmitripivkine:
Jenkins test sanity xLinux jdk8 depends eclipse/omr#3436

amicic:

amicic:
resolved conflicts that showed up meanwhile

PR #4233, opened 2019-01-10
Remove unused constants: J9Init*

keithc-ca:
This was noticed because the macro definition
  `#define J9InitEnd 0x-1`
elicits numerous warnings from xlc:
  `... 1540-0887 (S) The hexadecimal literal "0x" is not valid.`
during DDR macro processing.


gacholio:
jenkins compile zlinux jdk8

PR #4234, opened 2019-01-10
Fix for getting CPU utilization on server as zero

satbir-ca:
updateCpuUtil needed to be called before quering the
getCpuUsage.

### DIFF
--- a/runtime/compiler/runtime/StatisticsThread.cpp
+++ b/runtime/compiler/runtime/StatisticsThread.cpp
@@ -83,14 +83,22 @@ static int32_t J9THREAD_PROC statisticsThreadProc(void * entryarg)
 
          if ((statisticsThread->getStatisticsFrequency() != 0) && ((crtTime - lastStatistics) > statisticsThread->getStatisticsFrequency()))
             {
+            int32_t cpuUsage = 0, avgCpuUsage = 0, vmCpuUsage = 0;
+            CpuUtilization *cpuUtil = compInfo->getCpuUtil();
+            if (cpuUtil->isFunctional())
+               {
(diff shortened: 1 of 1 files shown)


satbir-ca:

runtime/compiler/runtime/StatisticsThread.cpp
```diff
@@ -83,14 +83,24 @@ static int32_t J9THREAD_PROC statisticsThreadProc(void * entryarg)
...
+               vmCpuUsage = cpuUtil->getVmCpuUsage();
+               isFunctional = true;
+               }
             TR_VerboseLog::vlogAcquire();
             TR_VerboseLog::writeLine(TR_Vlog_JITaaS, "Number of clients : %u", compInfo->getClientSessionHT()->size());
             TR_VerboseLog::writeLine(TR_Vlog_JITaaS, "Total compilation threads : %d", compInfo->getNumUsableCompilationThreads());
             TR_VerboseLog::writeLine(TR_Vlog_JITaaS, "Active compilation threads : %d",compInfo->getNumCompThreadsActive());
-            CpuUtilization *cpuUtil = compInfo->getCpuUtil();
-            if (cpuUtil->isFunctional())
+            if (isFunctional)
```
mpirvu: Nit-pick: it's not clear what `isFunctional` refers to without reading a bunch of the preceding code. One possibility is to change the name of the variable to `isCpuUtilizationFunctional`, but a better approach is to eliminate this variable and directly test cpuUtil->isFunctional(). This is a getter that should be easily inlined by the c++ compiler, so there is no runtime overhead.

satbir-ca:
updated

Issue #4241, opened 2019-01-10
All JIT Power work for "Correctly truncate boolean values in the appropriate bytecodes #2049"


gita-omr:
Power work for  https://github.com/eclipse/openj9/issues/2049

- [x] JNI methods: https://github.com/eclipse/openj9/issues/4244

gita-omr:
@DanHeidinga is this all that is required from JIT for https://github.com/eclipse/openj9/issues/2049 ?

DanHeidinga:
My understanding was the conversions for `bastore`, `putfield`, `putstatic` & and truncated return bytecodes were complete in `0.11.0`.   Can you confirm this is the case?

For JIT behaviour, `0.12.0` added checking jni boolean returns using ` != 0`.  

The Interpreter has also added similar checks for Unsafe access which the JIT will need to match in 0.13.0

gita-omr:
@andrewcraik could you confirm comment above? I added 'JIT common work' bullet to the parent issue.

andrewcraik:
yes bastore, putfield, putstatic and truncated return were done in `0.11.0`. The codegen change needed was the truncation of the jni Boolean returns where direct invocation was performed (the VM truncates already due to another change included in `0.12.0`). The work in `0.11.0` was done at the trees level and did not require specific codegen changes - we just added the appropriate iand operations.

gita-omr:
Ok, so closing then.

PR #4243, opened 2019-01-10
Ensure boolean JNI methods on Power produce a result of zero or one

R-Santhir:
When a calling a JNI boolean method, if the native implementation returns zero
the result is false, and true otherwise.

Progresses issue #4241


R-Santhir:
@gita-omr can you please review when you get a chance

gita-omr:

PR #4247, opened 2019-01-11
Remove Xdiagnosticscollector 1192

linyuehan:
Removes the the -Xdiagnosticscollector option

Fixes: eclipse#1192

Signed-off-by: Yuehan Lin <Yuehan.Lin@ibm.com>


pshipton:
We don't remove NLS messages, as this causes the remaining messages to be renumbered. The documentation and tests may refer to NLS message numbers.

pshipton:
See the comments at the top of the NLS files.

PR #4249, opened 2019-01-11
Optimized unresolved static/special glue on X86-64

0dvictor:
Setup RDI with desired values and return to next instruction directly, avoiding re-run the patch code.


0dvictor:
Although not related, this PR assumes https://github.com/eclipse/openj9/pull/3459 will be merged soon, so that deprecating code is not updated.

andrewcraik:

PR #4253, opened 2019-01-11
Enable JITInlineFieldwatch when option specified

cathyzhyi:
This change enables JIT inlined field watch on X86 when
-XX:+JITInlineWatches is specified in the Java command line.

### DIFF
--- a/runtime/compiler/control/J9Options.cpp
+++ b/runtime/compiler/control/J9Options.cpp
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2000, 2018 IBM Corp. and others
+ * Copyright (c) 2000, 2019 IBM Corp. and others
  *
  * This program and the accompanying materials are made available under
  * the terms of the Eclipse Public License 2.0 which accompanies this
@@ -2313,6 +2313,12 @@ bool J9::Options::feLatePostProcess(void * base, TR::OptionSet * optionSet)
       {
       self()->setOption(TR_DisableNextGenHCR);
       }
+
+#if !defined(TR_HOST_X86)
+   //The bit is set when -XX:+JITInlineWatches is specified
+   if (J9_ARE_ANY_BITS_SET(javaVM->extendedRuntimeFlags, J9_EXTENDED_RUNTIME_JIT_INLINE_WATCHES))
+      TR_ASSERT_FATAL(false, "this platform doesn't support JIT inline field watch");
+#endif
 
    // GCR and JProfiling are disabled under FSD for a number of reasons
    // First, there is confusion between the VM and the JIT as to whether a call to jitRetranslateCallerWithPreparation is a decompilation point.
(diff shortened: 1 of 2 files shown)


runtime/compiler/control/J9Options.cpp
```diff
@@ -2314,6 +2314,12 @@ bool J9::Options::feLatePostProcess(void * base, TR::OptionSet * optionSet)
       self()->setOption(TR_DisableNextGenHCR);
       }
 
+#if !defined(TR_HOST_X86)
+   //The bit is set when -XX:+JITInlineWatches is specified
+   if (J9_ARE_ANY_BITS_SET(javaVM->extendedRuntimeFlags, J9_EXTENDED_RUNTIME_JIT_INLINE_WATCHES))
+      TR_ASSERT_FATAL(false, "this platform doesn't support JIT inline field watch");
```
andrewcraik: I'm not sure a fatal assert and crash is the right thing to do - isn't there a more graceful way to fail?

cathyzhyi:
@gacholio Currently we assert on platforms not supporting field watch and it's better if we can fall back to old field watch without crashing. Is there a place in VM we can do that?

gacholio:
If we had a build flag in the spec, we could do it. I'm not sure it's worth the effort, as I don't plan to publicize the option to enable field watch (once it's working on all platforms, we'll remove the old code).

@DanHeidinga Any thoughts? I'm fine with the assertion in the short term.

DanHeidinga:
I'm on board with failing to start the JVM if an invalid option is specified.  An assert feels like a strange way to do that.  Isn't there an existing way in the JIT to fail to start due to invalid options?

Also, `-XX:+` options always have a `-XX:-` version as well to allow disabling the option.  

Thinking about this further, we should only check for the `-XX:[+-]JITInlineWatches` option on platforms that support it.  Unrecognized -XX options are ignored.

gacholio:
We can do that (though the options are soon not going to be ignored, iirc). Better that we fail early with a vaguely useful message than assert.

DanHeidinga:
I misread this PR.  It's not adding the option, only testing the result.

In that case, asserting is fine as we should never enable the bit on a platform that doesn't support the option.

gacholio:
@cathyzhyi If you remove your check and assert, you can add `#if defined(J9VM_ARCH_X86)` here:

https://github.com/eclipse/openj9/blob/ffbffce5ca61db3c775b19f37ee3e11a55c6d5eb/runtime/vm/jvminit.c#L2977-L2980

gacholio:
For performance testing, asserting is better than silently ignoring the option.

gacholio:
Looks like the final decision is to leave the assert the way it is.

runtime/compiler/control/J9Options.cpp
cathyzhyi: It's a new option for users to experiment with and will not be publicized so It's probably better to keep the assertion here and let it fail fast in our testing environment.

PR #4257, opened 2019-01-11
Revert "Fix incorrect -Xdump:dump:defaults output"

keithc-ca:
Reverts eclipse/openj9#563 which changed the default `-Xdump` options. This will restore the old defaults and we can explore fixing the help output later.


pshipton:
As this is a generated revert, I'm going to ignore the copyright failure.

PR #4259, opened 2019-01-11
Use a set in SymbolValidationManager::recordExists

jdmpapin:
This reduces the running time of `recordExists` from linear to log time. Furthermore, by using a separate data structure here, it's now possible to seed the set with records that should not be generated.

An ordering comparison is defined on validation records in order to allow them to be kept in a set.

### DIFF
--- a/runtime/compiler/runtime/SymbolValidationManager.cpp
+++ b/runtime/compiler/runtime/SymbolValidationManager.cpp
@@ -1121,6 +1121,52 @@ static void printClass(TR_OpaqueClassBlock *clazz)
       }
    }
 
+namespace // file-local
+   {
+   class LexicalOrder
+      {
+      enum Result
+         {
+         LESS,
+         EQUAL,
+         GREATER,
+         };
+
+      public:
+      LexicalOrder() : _result(EQUAL) { }
+      LexicalOrder(const LexicalOrder &other) : _result(other._result) { }
+
+      static LexicalOrder by(void *a, void *b) { return LexicalOrder().thenBy(a, b); }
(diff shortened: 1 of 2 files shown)


runtime/compiler/runtime/SymbolValidationManager.cpp
```diff
@@ -1121,6 +1121,52 @@ static void printClass(TR_OpaqueClassBlock *clazz)
...
+         LESS,
+         EQUAL,
+         GREATER,
+         };
+
+      public:
+      LexicalOrder() : _result(EQUAL) { }
+      LexicalOrder(const LexicalOrder &other) : _result(other._result) { }
+
+      static LexicalOrder by(void *a, void *b) { return LexicalOrder().thenBy(a, b); }
```
dsouzai: Should the return value be `LexicalOrder &`?

jdmpapin: There's no `LexicalOrder` we could reference here that lives past the return

dsouzai: I see; then I guess my question is, why have `thenBy` return a reference when passed in `void *` but return by value when passed in `uintptr_t `?

jdmpapin: Oh, thanks - I didn't mean to copy in `thenBy(uintptr_t, uintptr_t)`

jdmpapin: Fixed

dsouzai:

dsouzai:

dsouzai:
Jenkins copyright check

dsouzai:
Jenkins line endings check

PR #4260, opened 2019-01-11
Create fewer ClassInstanceOfClass records

jdmpapin:
No record is necessary when the inheritance relationship between the types is known in advance.


dsouzai:

dsouzai:

PR #4272, opened 2019-01-11
WIP Remove Xdiagnosticscollector

Yuehan-Lin:
Removes the the -Xdiagnosticscollector option

Fixes: eclipse#1192

Signed-off-by: Yuehan-Lin <Yuehan.Lin@ibm.com>


pshipton:
@Yuehan-Lin we don't delete obsolete NLS message, see the comment at the top of the file.
```
```

pshipton:
A doc issue needs to be created at https://github.com/eclipse/openj9-docs/issues/new/choose
The option is documented, so the documentation needs to change as well.

https://www.eclipse.org/openj9/docs/xdiagnosticscollector/

For the record, the option to print the warning has been in the code since Apr 24, 2013.

Issue #4281, opened 2019-01-14
Intermittent failure running ClassLibraryLambdaTest on Z


hzongaro:
We saw an intermittent failure running ClassLibraryLambdaTest on z/OS.  The failure was exposed by changes in Escape Analysis merged in pull request #2752.  Those changes were subsequently reverted in pull request #4145 to fix that failure.

This issue is being opened to track the root cause.

PR #4282, opened 2019-01-14
Avoid mishandling of immutable objects that escape in cold blocks

hzongaro:
In fixupNode, force non-contiguous, immutable objects that escape in
cold blocks to go through fixupFieldAccess processing to ensure the
temporary that will replace it will be properly initialized.

Then, in fixupFieldAccessForNonContiguousAllocation, ensure the store
to the original object is preserved along with the store to the temporary
that's being created to replace its field.  The former will be needed
if we escape, and the latter is needed if we don't escape.

Fixes:  #4281

### DIFF
--- a/runtime/compiler/optimizer/EscapeAnalysis.cpp
+++ b/runtime/compiler/optimizer/EscapeAnalysis.cpp
@@ -4731,7 +4731,12 @@ bool TR_EscapeAnalysis::fixupNode(TR::Node *node, TR::Node *parent, TR::NodeChec
 
             if (fieldIsPresentInObject)
                {
-               if (candidate->escapesInColdBlock(_curBlock))
+               // Special case handling of non-contiguous immmutable object:
+               // if it escapes in a cold block, need to ensure the temporary
+               // that replaces it is correctly initialized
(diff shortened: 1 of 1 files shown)


hzongaro:
@vijaysun-omr Vijay, may I ask you to review this pull request?

vijaysun-omr:
Looks good to me @hzongaro The logic makes sense to keep both auto and field store when we have the special case of an immutable discontiguous allocation that escapes in the given block.

hzongaro:
Thanks, Vijay @vijaysun-omr , for all your guidance, help and advice in fixing this problem!

andrewcraik:

andrewcraik:
Jenkins test sanity xlinux jdk11,jdk8

DanHeidinga:
This is targeted for post-`0.12.0`.  I've assigned it to the `0.13.0` milestone so we can ensure it gets merged as soon as possible after we open up master again

charliegracie:
Can this be merged now?

andrewcraik:
I believe so - I'm happy with this given the clean sanity and review from @vijaysun-omr.

PR #4286, opened 2019-01-14
Use strip '-t' option to remove debug symbols

keithc-ca:
Using `strip -t` doesn't appear to have the problems encountered with `strip -r` in https://github.com/eclipse/openj9/pull/4276.


keithc-ca:
Jenkins test sanity aix jdk8,jdk11

keithc-ca:
Jenkins test sanity aix jdk8 depends ibmruntimes/openj9-openjdk-jdk8#237

PR #4289, opened 2019-01-14
Enforce rule override for .S files on zLinux 32 bit for libffi

vsebe:
Update freemarker if condition to override rule for .S files for zLinux
32 bit platforms.
 
[ci skip]


vsebe:
@pshipton please review, thank you!

pshipton:

PR #4290, opened 2019-01-14
Removing the -xml parameter from jdmpview help

### LINKED ISSUE #1193: jdmpview doesn't need -xml parameter
jdmpview help still shows -xml although this is obsolete. The help output can be removed, and likely any remaining supporting code.

samuel-rubin:
Removed the -xml parameter from the help message and the supporting code in the file.

Fixes #1193 

### DIFF
--- a/jcl/src/openj9.dtfjview/share/classes/com/ibm/jvm/dtfjview/Session.java
+++ b/jcl/src/openj9.dtfjview/share/classes/com/ibm/jvm/dtfjview/Session.java
@@ -1,6 +1,6 @@
 /*[INCLUDE-IF Sidecar18-SE]*/
 /*******************************************************************************
- * Copyright (c) 2004, 2018 IBM Corp. and others
+ * Copyright (c) 2004, 2019 IBM Corp. and others
  *
  * This program and the accompanying materials are made available under
  * the terms of the Eclipse Public License 2.0 which accompanies this
@@ -89,7 +89,6 @@ public class Session implements ISession {
 	public static String prompt = "> ";
 	
 	private static final String ARG_CORE = "-core";
-	private static final String ARG_XML = "-xml";
 	private static final String ARG_ZIP = "-zip";
 	private static final String ARG_VERSION = "-version";
 	private static final String ARG_VERBOSE = "-verbose";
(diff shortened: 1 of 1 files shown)


samuel-rubin:
This is the output that I get when I try to input an XML file 

```
The command -xml was not recognised
Usage: "dtfjview -core <core_file> [-verbose]" or
       "dtfjview -zip <zip_file> [-verbose]" or
       "dtfjview -version"

To analyze dumps from DDR-enabled JVMs, dtfjview only requires a core file

The default ImageFactory is null. To change the ImageFactory use:
	 -J-Dcom.ibm.dtfj.image.factoryclass=<classname>

dtfjview can also be used in 'batch mode' whereby commands can be issued without entering an interactive prompt.
This processing is controlled via the following command line options :
        -cmdfile <path to command file> : will read and sequentially execute
                                          a series of dtfjview commands in the specified file
        -charset <character set name>   : specifies the character set for the commands in -cmdfile.
        -outfile <path to output file>  : file to store any output generated by commands
        -overwrite                      : specifies that the file specified by -outfile should be overwritten
        -append                         : specifies that the file specified by -outfile should be appended.

It is possible to execute a single command without specifying a command file by appending the command to the end
 of the command line which executes dtfjview
e.g. dtfjview -core mycore.dmp info class

Finally, some commands which take the * parameter will need to have that specified on the command line as ALL
This is to ensure correct processing by the OS e.g. dtfjview -core mycore.dmp info thread ALL
```

samuel-rubin:
@babsingh for review

jcl/src/openj9.dtfjview/share/classes/com/ibm/jvm/dtfjview/Session.java
```diff
@@ -467,12 +465,10 @@ private void printHelp() {
...
 		
 		out.print(
-				"Usage: \"" + launcher + " -core <core_file> [-xml <xml_file>] [-verbose]\" or \n" +
+				"Usage: \"" + launcher + " -core <core_file> [-verbose]\" or \n" +
 				"       \"" + launcher + " -zip <zip_file> [-verbose]\" or \n" +
 				"       \"" + launcher + " -version\"\n\n" +
-				"To analyze dumps from DDR-enabled JVMs, " + launcher + " only requires a core file, no XML file " +
-				"is needed.\n\nTo analyze dumps from non DDR-enabled JVMs, " + launcher + " requires both core and " +
-				"XML files (jextract must be run on the dump first, to produce the XML file).\n\n" +
+				"To analyze dumps from DDR-enabled JVMs, " + launcher + " only requires a core file\n\n" +
```
babsingh: **a core file.** - full stop (punctuation) is missing.

pshipton:
jenkins test all zlinux jdk8

samuel-rubin:
@pshipton I rebased my change.

pshipton:
jenkins test all zlinux jdk8

PR #4302, opened 2019-01-15
Disable vmcheck test on MacOS temporarily

pdbain-ibm:
Disable until https://github.com/eclipse/openj9/issues/4279 is fixed.

### DIFF
--- a/test/functional/Java8andUp/playlist.xml
+++ b/test/functional/Java8andUp/playlist.xml
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="UTF-8"?>
 
 <!--
-  Copyright (c) 2016, 2018 IBM Corp. and others
+  Copyright (c) 2016, 2019 IBM Corp. and others
 
   This program and the accompanying materials are made available under
   the terms of the Eclipse Public License 2.0 which accompanies this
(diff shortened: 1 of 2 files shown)


pdbain-ibm:
FYI @llxia @pshipton 
This test runs on my Mac and on other platforms.  Excluding the test while I investigate.

llxia:
jenkins test extended osx jdk11

pshipton:
This test only needs to be excluded for jdk8. It wasn't failing or hanging on jdk11. Is that possible?

pshipton:
Meanwhile I'll run a jdk8 build so we can see how that stands without this test.

pshipton:
jenkins test extended osx jdk8

llxia:
jdk8 failed at git fetch
https://ci.eclipse.org/openj9/job/PullRequest-Extended-JDK8-osx_x86-64_cmprssptrs-OpenJ9/6/console

Going to rerun jdk8

llxia:
jenkins test extended osx jdk8

llxia:
> This test only needs to be excluded for jdk8. It wasn't failing or hanging on jdk11. Is that possible?


yes, we can use `latest_exclude_8.txt` to exclude the TestNG test only for jdk8. Since `testvmcheck` is using testNG, so we should be able to use it.
https://github.com/eclipse/openj9/blob/master/test/TestConfig/resources/excludes/latest_exclude_8.txt

I initially suggested playlist.xml because it is the easiest way to do temporary exclude. But yes, it will exclude both jdk11 and jdk8 on osx in this case.

pdbain-ibm:
I hope to have a fix for this today.  If so, I will close this PR.

pshipton:
Created https://github.com/eclipse/openj9/issues/4316 for the jdk8 extended test failure.

pdbain-ibm:
Testing the fix.  Won't need this.

PR #4303, opened 2019-01-15
Remove hardcoded system property java.util.prefs.PreferencesFactory

JasonFengJ9:
Remove hardcoded system property `java.util.prefs.PreferencesFactory`

This only affects `J9VM_UNIX` platforms.

This modification conforms to specification, and allows a `PreferencesFactory` implementation from a `jar` file containing a provider-configuration file named `java.util.prefs.PreferencesFactory` in
the resource directory `META-INF/services` when this system property is not set.

In case neither the system property nor an extension jar file is provided, a default `PreferencesFactory` implementation for the underlying platform is provided. For a platform not `Windows` or `OS X`, a default `PreferencesFactory` is `java.util.prefs.FileSystemPreferencesFactory` which was hardcoded via system property java.util.prefs.PreferencesFactory.

References: https://docs.oracle.com/en/java/javase/11/docs/api/java.prefs/java/util/prefs/Preferences.html - "Implementation Note"
https://github.com/ibmruntimes/openj9-openjdk-jdk11/blob/openj9/src/java.prefs/share/classes/java/util/prefs/Preferences.java#L293-L302

Note: Not able to trace back why this system property was hardcoded within `VM`. There are couple of `VmArgumentTests` that verify this system property can be override by command line options, which is irrelevant to this PR. (a person build with this PR still passes the test in question) 
https://github.ibm.com/runtimes/openj9/blob/ibm_sdk/test/functional/Java8andUp/src/org/openj9/test/vmArguments/VmArgumentTests.java#L215-L273

closes: https://github.com/eclipse/openj9/issues/3232 

Reviewer: @DanHeidinga 

### DIFF
--- a/runtime/jcl/common/vm_scar.c
+++ b/runtime/jcl/common/vm_scar.c
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 1998, 2018 IBM Corp. and others
+ * Copyright (c) 1998, 2019 IBM Corp. and others
  *
  * This program and the accompanying materials are made available under
  * the terms of the Eclipse Public License 2.0 which accompanies this
@@ -202,12 +202,6 @@ addBFUSystemProperties(J9JavaVM* javaVM)
 	}
 #endif
 
-	if (J9SYSPROP_ERROR_NOT_FOUND == vmfunc->getSystemProperty(javaVM, "java.util.prefs.PreferencesFactory", NULL)) {
-		rc = vmfunc->addSystemProperty(javaVM, "java.util.prefs.PreferencesFactory", "java.util.prefs.FileSystemPreferencesFactory", 0);
-		if (J9SYSPROP_ERROR_NONE != rc) {
-			return rc;
-		}
-	}
 	if (fontPathSize >= 0) {
 		if (J9SYSPROP_ERROR_NOT_FOUND == vmfunc->getSystemProperty(javaVM, "sun.java2d.fontpath", NULL)) {
 			rc = vmfunc->addSystemProperty(javaVM, "sun.java2d.fontpath", fontPathBuffer, 0);


DanHeidinga:
@JasonFengJ9 I traced it back internally to an old cmvc defect: 162032 but I don't have access to that system right now.  There might be more background there if you have a chance to check it

DanHeidinga:
Also does this apply to both jdk8 & 11?  The doc link above is to the 11 doc

JasonFengJ9:
@DanHeidinga This does apply to `JDK8` as well. Please refer following links:
https://docs.oracle.com/javase/8/docs/api/java/util/prefs/Preferences.html
https://github.com/ibmruntimes/openj9-openjdk-jdk8/blob/openj9/jdk/src/share/classes/java/util/prefs/Preferences.java#L291-L308

Looking at `cmvc 162032`.

JasonFengJ9:
Thanks @DanHeidinga for identifying `cmvc 162032` (created ~9 years ago :)
Here is the summary about `cmvc 162032`:
It was a defect against a regression introduced by `J9 Java 7 JVM` with `Harmony` version of  `java.util.prefs.Preferences`. There was a `LIR1029.7` which changed the system property value for `java.util.prefs.PreferencesFactory` from `java.util.prefs.FileSystemPreferencesFactory` to an empty string `""`. 
The fix for this defect restored the system property value in question to `java.util.prefs.FileSystemPreferencesFactory`. 
So the system property value was hardcoded to this value even before `cmvc 162032` time.
My guess is that this probably was the behaviour of earlier `RI` as well, and this function was moved from natives to java code `java.util.prefs.Preferences` later, i.e., the logic becomes:
1. Check if there is a system property `java.util.prefs.PreferencesFactory` set (value `null` or not);
2. If the system property value was `null` (not set), look at provider-configuration file named `java.util.prefs.PreferencesFactory` in the `jar` file resource directory `META-INF/services` (this is what https://github.com/eclipse/openj9/issues/3232 was testing);
3. At last, if neither such system property was set nor there is a provider-configuration file, use default platform factory implementation according to the `OS`, i.e., `java.util.prefs.WindowsPreferencesFactory` for `Windows`, `java.util.prefs.MacOSXPreferencesFactory` for `OS X`, and `java.util.prefs.FileSystemPreferencesFactory` for others like `UNIX`.

DanHeidinga:
Jenkins test sanity plinux,osx,win jdk8

PR #4304, opened 2019-01-15
Remove redundant access mode checks in Lookup

pdbain-ibm:
Various access checking methods in MethodHandles.Lookup test for NO_ACCESS and
PRIVILEGED_ACCESS the same way.  In some case, the same check is performed
multiple times on the same handle.  Factor the repeated code out to a utility
method. Ensure it is called in all paths.  Rename for emphasis the methods
implementing the access checks and remove redundant checks from them.


pdbain-ibm:
@pshipton or @DanHeidinga would you kindly review this? It will simplify the changes for 
https://github.com/eclipse/openj9/pull/3499

genie-openj9:

PR #4305, opened 2019-01-15
Switch Linux platforms to compile with gcc 7

pshipton:
Doc issue https://github.com/eclipse/openj9-docs/issues/191


pshipton:
jenkins test sanity xlinux,plinux,zlinux jdk8

pshipton:
@DanHeidinga @keithc-ca I think is ready to go.

pshipton:
@jdekonin fyi

DanHeidinga:
+1 from me as this makes the compilers consistent with what Adopt is using.

DanHeidinga:
fyi @dnakamura the cmake scripts (lists?) probably need to be updated for this as well

PR #4309, opened 2019-01-15
Update relocation header to support SVM On Z

r30shah:
Update Relocation header to add support for SymbolValidationManager if it is enabled on Z.

### DIFF
--- a/runtime/compiler/z/codegen/J9AheadOfTimeCompile.cpp
+++ b/runtime/compiler/z/codegen/J9AheadOfTimeCompile.cpp
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2000, 2018 IBM Corp. and others
+ * Copyright (c) 2000, 2019 IBM Corp. and others
  *
  * This program and the accompanying materials are made available under
  * the terms of the Eclipse Public License 2.0 which accompanies this
@@ -153,6 +153,8 @@ uint8_t *J9::Z::AheadOfTimeCompile::initializeAOTRelocationHeader(TR::IteratedEx
    // This has to be created after the kind has been written into the header
    TR_RelocationRecord storage;
    TR_RelocationRecord *reloRecord = TR_RelocationRecord::create(&storage, reloRuntime, reloTarget, reinterpret_cast<TR_RelocationRecordBinaryTemplate *>(relocation->getRelocationData()));
+   TR::Compilation *comp = _cg->comp();
+   TR::SymbolValidationManager *symValManager = comp->getSymbolValidationManager();
 
    switch (relocation->getTargetKind())
       {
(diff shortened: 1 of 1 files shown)


r30shah:
@dsouzai  Can I please get a review on this? This commit simply contains changes to update Relocation header on Z to support SVM. I think I have considered all changes recently made in https://github.com/eclipse/openj9/blob/master/runtime/compiler/x/codegen/J9AheadOfTimeCompile.cpp . Need to double verify so marked it WIP. 
FYI @fjeremic

runtime/compiler/z/codegen/J9AheadOfTimeCompile.cpp
```diff
@@ -280,7 +282,17 @@ uint8_t *J9::Z::AheadOfTimeCompile::initializeAOTRelocationHeader(TR::IteratedEx
          *(uintptrj_t *)cursor = (uintptrj_t) classSymRef->getOwningMethod(self()->comp())->constantPool();
          cursor += SIZEPOINTER;
 
-         *(uintptrj_t *)cursor = (uintptrj_t) classSymRef->getCPIndex(); // cp Index
+         if (comp->getOption(TR_UseSymbolValidationManager))
+            {
+            TR_OpaqueClassBlock *clazz = (TR_OpaqueClassBlock *)recordInfo->data5;
+            uintptrj_t classID = (uintptrj_t)symValManager->getIDFromSymbol(static_cast<void *>(clazz));
+            TR_ASSERT_FATAL(classID, "classID should exist!\n");
```
dsouzai: I'm gonna defer to @fjeremic 's thoughts on this one; I just figure it's worth mentioning that though this does mean the AOT compile is compromised, it doesn't necessarily mean the JVM is. Just a `TR_ASSERT` and failing the compilation here might be the better thing to do in a non-debug build.

runtime/compiler/z/codegen/J9AheadOfTimeCompile.cpp
```diff
@@ -305,7 +317,17 @@ uint8_t *J9::Z::AheadOfTimeCompile::initializeAOTRelocationHeader(TR::IteratedEx
          *(uintptrj_t *)cursor = (uintptrj_t) classSymRef->getOwningMethod(self()->comp())->constantPool();
          cursor += SIZEPOINTER;
 
-         *(uintptrj_t *)cursor = (uintptrj_t) classSymRef->getCPIndex(); // cp Index
+         if (comp->getOption(TR_UseSymbolValidationManager))
+            {
+            TR_OpaqueClassBlock *classOfMethod = (TR_OpaqueClassBlock *)recordInfo->data5;
+            uintptrj_t classID = (uintptrj_t)symValManager->getIDFromSymbol(static_cast<void *>(classOfMethod));
+            TR_ASSERT_FATAL(classID, "classID should exist!\n");
```
dsouzai: Same comment as above.

runtime/compiler/z/codegen/J9AheadOfTimeCompile.cpp
fjeremic: Given that we don't have debug builds where such a `TR_ASSERT` would fire and because this is new code I'd rather us fail completely so we can fix the obvious bug than suffer a silent never to be caught performance issue.

fjeremic:

PR #4310, opened 2019-01-15
Update relocation for Object Allocation inlining

r30shah:
Update New object allocation inlining to add relocation under SVM if it is enabled.

### DIFF
--- a/runtime/compiler/z/codegen/J9TreeEvaluator.cpp
+++ b/runtime/compiler/z/codegen/J9TreeEvaluator.cpp
@@ -8153,7 +8153,9 @@ genInitObjectHeader(TR::Node * node, TR::Instruction *& iCursor, TR_OpaqueClassB
 
       // a pointer to the virtual register that will actually hold the class pointer.
       TR::Register * clzReg = classReg;
-
+      // TODO: Following approach for initializing object header for array of objects in AOT is conservative.
+      // We need support for relocation in generating RIL type instruction. If we have support, we can use
+      // same sequence generated in JIT which saves us a load and store.
       if (comp->compileRelocatableCode())
          {
          if (node->getOpCodeValue() == TR::newarray)
@@ -8861,20 +8863,7 @@ J9::Z::TreeEvaluator::VMnewEvaluator(TR::Node * node, TR::CodeGenerator * cg)
       current = cg->getAppendInstruction();
 
       TR_ASSERT(current != NULL, "Could not get current instruction");
-
-      if (comp->compileRelocatableCode() && (opCode == TR::New || opCode == TR::anewarray))
-         {
-         iCursor = firstInstruction = generateS390BranchInstruction(cg, TR::InstOpCode::BRC, TR::InstOpCode::COND_VGNOP, node, callLabel, current);
-         if(!firstBRCToOOL)
-            {
-            firstBRCToOOL = iCursor;
-            }
-         else
-            {
-            secondBRCToOOL = iCursor;
-            }
-         }
-
+      
       if (outlineNew)
          {
          if (isVariableLen)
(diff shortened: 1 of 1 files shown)


r30shah:
@fjeremic @dsouzai  Can I please get review on this change?

runtime/compiler/z/codegen/J9TreeEvaluator.cpp
```diff
@@ -8195,7 +8195,9 @@ genInitObjectHeader(TR::Node * node, TR::Instruction *& iCursor, TR_OpaqueClassB
 
       // a pointer to the virtual register that will actually hold the class pointer.
       TR::Register * clzReg = classReg;
-
+      // TODO: Following approach for initializing object header for array of objects in AOT is conservative.
```
r30shah: @fjeremic Currently on Z, whenever we are loading a constant address, we use snippet in case we are generating relocatable code.  We should add relocation support for RIL type instructions which can be used to load address so we can have better AOT compiled body.

runtime/compiler/z/codegen/J9TreeEvaluator.cpp
```diff
@@ -9092,8 +9099,15 @@ J9::Z::TreeEvaluator::VMnewEvaluator(TR::Node * node, TR::CodeGenerator * cg)
             {
             classSymRef = node->getSecondChild()->getSymbolReference();
             reloKind = TR_VerifyRefArrayForAlloc;
-            }
 
+            if (comp->getOption(TR_UseSymbolValidationManager))
+               classToValidate = comp->fej9()->getComponentClassFromArrayClass(classToValidate);
```
dsouzai: Might be worth having a comment here stating why this is necessary. I don't think there was a comment on the x codegen because I dropped the ball there :).

The reason we need to validate the component class is because in AOT w/o SVM, we validate the class by pulling it out of the constant pool; however that class is not the array class (because the `anewarray` bytecode refers to the component class). However, in the evaluator, we directly refer to the array class. In order to have the relo infrastructure be consistent, we need to remember to validate the component class (since the relo infrastructure is expecting the component class).

runtime/compiler/z/codegen/J9TreeEvaluator.cpp
r30shah: Opened up issue : https://github.com/eclipse/openj9/issues/4391

fjeremic:
@r30shah still a couple of things left here from both reviewers.

runtime/compiler/z/codegen/J9TreeEvaluator.cpp
```diff
@@ -8908,6 +8910,10 @@ J9::Z::TreeEvaluator::VMnewEvaluator(TR::Node * node, TR::CodeGenerator * cg)
 
       if (comp->compileRelocatableCode() && (opCode == TR::New || opCode == TR::anewarray))
          {
+         // We need to validate calculated object size for given class in AOT load run to validate the AOT compiled body.
+         // Relocation record will be added at this NOP instruction later in this evaluator to validate calculated size of object for given class.
+         // TODO: Instead of generating NOP instruction and add relocation record there, can we add relocation record to the instruction which
+         // loads register with calculated object size (which is immediate value) ?
```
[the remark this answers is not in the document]
r30shah: Made change to test if we can remove NOP : https://github.com/eclipse/openj9/pull/4310/commits/27e4fa395c0e39d074854c42553f892f33d9c3d3

Marking this one as WIP till I finish testing it. Will add removal of NOP to this PR as well if testing goes fine.

r30shah:
@fjeremic Local testing on removing NOP works. I have combined those changes with comments regarding to validating component class for SVM AOT is added in https://github.com/eclipse/openj9/pull/4310/commits/21bba9fff8ca2bca7c8bbd4a86e3240f416382a9
Let me know if changes are fine, will squash them. Removing WIP.

fjeremic:

fjeremic:
> Squashed commits. Should be good to merge once jenkins sanity test passes

Unfortunately force pushing deletes Jenkins builds. Will need to launch them again.

PR #4320, opened 2019-01-16
Enable String.indexOf transformation for JITaaS

dmitry-ten:
`J9TransformUtil::transformStringIndexofCall` was not working, because `criticalsection.hasVMAccess()` always returns false on the server, which aborts this transformation.
Enable this transformation by continuing without VM access in JITaaS, and overriding front-end `getStringLength` and `getStringCharacter` for the server

### DIFF
--- a/runtime/compiler/control/JITaaSCompilationThread.cpp
+++ b/runtime/compiler/control/JITaaSCompilationThread.cpp
@@ -975,6 +975,20 @@ bool handleServerMessage(JITaaS::J9ClientStream *client, TR_J9VM *fe)
          client->write(fe->dereferenceStaticFinalAddress(address, addressType));
          }
          break;
+      case J9ServerMessageType::VM_getStringLength:
+         {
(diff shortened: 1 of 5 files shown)


dmitry-ten:
In AcmeAir on lara, it doesn't seem to do anything, because there is no fast String.indexOf method in the class library.
I assumed that this transformation would be easy to enable, but with all the things mentioned above, I guess it's not worth it.

mpirvu:
The correct implementation is complicated and the rewards are small, thus I propose we close this PR for now.

fjeremic:
> As far as I understand the code `transformStringIndexOfCall` tries to replace `java/lang/String.indexOf(String)` with another `index` method that has a different signature `"(Ljava/lang/String;Ljava/lang/String;IIC)I"`. This other method may or may not exist in the class library. On my installation I could not find it.

This method does not exist and is deprecated. We need to purge the corresponding JIT code.

Issue #4301, opened 2019-01-15
JTReg test fails - java/util/Spliterator/SpliteratorTraversingAndSplittingTest.java and java/lang/String/Chars.java


ben-walsh:
Test https://github.com/ibmruntimes/openj9-openjdk-jdk11/blob/openj9/test/jdk/java/util/Spliterator/SpliteratorTraversingAndSplittingTest.java consistently fails with multiple failures like this ...

```
test SpliteratorTraversingAndSplittingTest.testIntForEach(""𐀀".codePoints().spliterator()", [65536], SpliteratorTraversingAndSplittingTest$SpliteratorOfIntCharDataBuilder$$Lambda$212/000000007C214660@7d9ec70c): failure
java.lang.AssertionError: expected [1] but found [2]
	at org.testng.Assert.fail(Assert.java:94)
	at org.testng.Assert.failNotEquals(Assert.java:496)
	at org.testng.Assert.assertEquals(Assert.java:125)
	at org.testng.Assert.assertEquals(Assert.java:267)
	at org.testng.Assert.assertEquals(Assert.java:277)
	at java.base/java.util.SpliteratorTestHelper.testForEach(SpliteratorTestHelper.java:267)
	at java.base/java.util.SpliteratorTestHelper.testForEach(SpliteratorTestHelper.java:174)
	at SpliteratorTraversingAndSplittingTest.testIntForEach(SpliteratorTraversingAndSplittingTest.java:875)
	at jdk.internal.reflect.GeneratedMethodAccessor13.invoke(Unknown Source)
	at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.base/java.lang.reflect.Method.invoke(Method.java:566)
	at org.testng.internal.MethodInvocationHelper.invokeMethod(MethodInvocationHelper.java:85)
	at org.testng.internal.Invoker.invokeMethod(Invoker.java:639)
	at org.testng.internal.Invoker.invokeTestMethod(Invoker.java:821)
	at org.testng.internal.Invoker.invokeTestMethods(Invoker.java:1131)
	at org.testng.internal.TestMethodWorker.invokeTestMethods(TestMethodWorker.java:125)
	at org.testng.internal.TestMethodWorker.run(TestMethodWorker.java:108)
	at org.testng.TestRunner.privateRun(TestRunner.java:773)
	at org.testng.TestRunner.run(TestRunner.java:623)
[...]

PR #4321, opened 2019-01-16
Fix implementation of String.codePoints()

keithc-ca:
The class `StringSpliteratorOfInt` worked fine for `chars()`, but not for `codePoints()` which must deal with the pairing of characters to form single code points. That class was removed in favour of using helper classes from `StringLatin1` and `StringUTF16`.

Also:
* use `Character.isHighSurrogate()` and `Character.isLowSurrogate()` to detect surrogate characters
* use `Character.toCodePoint()` to combine surrogates into code points
* fix tests: `((i + 1) < len)` can fail where `(i < (len - 1))` won't as long as `(len >= 0)`

Fixes: #4301

### DIFF
--- a/jcl/src/java.base/share/classes/java/lang/String.java
+++ b/jcl/src/java.base/share/classes/java/lang/String.java
@@ -2844,7 +2844,7 @@ public final class String implements Serializable, Comparable<String>, CharSeque
 		int end = last;
 
 		// Check if the String is compressed
-		if (enableCompression &&  (null == compressionFlag || coder == LATIN1)) {
+		if (enableCompression && (null == compressionFlag || coder == LATIN1)) {
 			while ((start <= end) && (helpers.byteToCharUnsigned(helpers.getByteFromArrayByIndex(value, start)) <= ' ')) {
 				start++;
 			}
@@ -3247,7 +3247,7 @@ public final class String implements Serializable, Comparable<String>, CharSeque
 			byte[]
 			chars = this.value;
 
-			final boolean compressed = enableCompression &&  (null == compressionFlag || coder == LATIN1);
+			final boolean compressed = enableCompression && (null == compressionFlag || coder == LATIN1);
 
 			int start = 0, current = 0, end = lengthInternal();
 			if (regex.lengthInternal() == 1 || singleEscapeLiteral) {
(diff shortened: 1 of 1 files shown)


jcl/src/java.base/share/classes/java/lang/String.java
```diff
@@ -3891,61 +3891,136 @@ static void checkBoundsBeginEnd(int begin, int end, int length) {
...
+				}
+
+				/*
+				 * If newEnd is the character index of a low surrogate, we include
+				 * the containing code point in this spliterator and only if there
+				 * are code points after it do we return a new spliterator.
+				 */
+				if (newEnd > start
+						&& Character.isHighSurrogate(charAtInternal(newEnd - 1))
+						&& Character.isLowSurrogate(charAtInternal(newEnd))) {
```
DanHeidinga: Do we need both tests?  If the `char[newEnd - 1]` is a high surrogate, then splitting in the middle of the pair seems wrong

jcl/src/java.base/share/classes/java/lang/String.java
```diff
@@ -3891,61 +3891,136 @@ static void checkBoundsBeginEnd(int begin, int end, int length) {
...
+
+				char high = charAtInternal(index);
+				int code = high;
+				int width = 1;
+
+				if (available > 1 && Character.isHighSurrogate(high)) {
+					char low = charAtInternal(index + 1);
+
+					if (Character.isLowSurrogate(low)) {
+						code = Character.MIN_SUPPLEMENTARY_CODE_POINT
```
DanHeidinga: Wouldn't calling `Character. toCodePoint(char high, char low)` be better here?

jcl/src/java.base/share/classes/java/lang/String.java
keithc-ca: The condition is consistent with treatment of code points elsewhere. See `codePointAt()` for example where a high surrogate must be followed by a low surrogate to be treated as a pair.

jcl/src/java.base/share/classes/java/lang/String.java
keithc-ca: I wasn't aware of that method: I like the suggestion.

keithc-ca: Addressed in second commit.

DanHeidinga:

jcl/src/java.base/share/classes/java/lang/String.java
```diff
@@ -3891,61 +3879,134 @@ static void checkBoundsBeginEnd(int begin, int end, int length) {
...
-	@Override
-	public IntStream chars() {
-		return StreamSupport.intStream(new StringSpliteratorOfInt(0, lengthInternal(), i -> charAt(i)), false);
-	}
-	
+
 	@Override
 	public IntStream codePoints() {
-		return StreamSupport.intStream(new StringSpliteratorOfInt(0, lengthInternal(), i -> codePointAt(i)), false);
+		final class CodeSpliterator implements Spliterator.OfInt {
```
fjeremic: This seems effectively a duplicate of `java/lang/StringLatin1.CharsSpliterator` [1]. There is also a `StringUTF16` version as well. I think we should be able to reuse these classes rather than rolling out our own. Can we check if this is possible?

[1] https://github.com/ibmruntimes/openj9-openjdk-jdk11/blob/007fb486206c01105c4652c4f7c1c69e1ee4a624/src/java.base/share/classes/java/lang/StringLatin1.java#L737-L792

fjeremic: Similarly for `chars()` as well.

fjeremic:
> My primary concern is with the change from direct checks to method calls as this performance-sensitive code. @fjeremic Any concerns?

This should be fine as the `Character.is(High|Low)Surrogate` methods are one liners with no side-effects. This JIT should always inline such methods.

jcl/src/java.base/share/classes/java/lang/String.java
keithc-ca: Another good suggestion: it seems to work just fine.

keithc-ca:
Implemented @fjeremic's suggestion, rebased and squashed.

keithc-ca:

keithc-ca:

keithc-ca:
Updated description to match the latest commit.

Issue #3610, opened 2018-11-08
Deprecate `getAccessStaticsIndirectly`


fjeremic:
@liqunl has been doing some work in the area of static final fields. She came to as with a question about `getAccessStaticsIndirectly` which is used in the Walker to determine which of the following tree sequences we generate for accessing static fields:

```
direct load <static field address symref>
```

or

```
indirect load <static field symref>
   loadaddr <class statics>
```

The code in question which controls this can be found in [1]. Currently the only setter of this is in the Z codegen on z9 or when compiling AOT [2]. I don't see a particular reason for this restriction however. @vijaysun-omr gave some historical context here:

>I asked Vijay if he recall any reason of the indirect load for static. He said it's not for functional purpose, it's an optimization. We use indirect load such that the `class statics` child can be commoned. He said materialization of a static address was expensive at the time this opt was implemented, but the perf gain wasn't much. If this opt is only used on Z9 and under non-AOT, he has no objection if Z team wants to clean it up.

We should clean this up and simplify the code. This can only be done after #3577 is addressed however.

[1] https://github.com/eclipse/openj9/blob/22e0009fda531be602d728ea128f4ad9b4d289fd/runtime/compiler/ilgen/Walker.cpp#L6199-L6206

[2] https://github.com/eclipse/openj9/blob/22e0009fda531be602d728ea128f4ad9b4d289fd/runtime/compiler/z/codegen/J9CodeGenerator.cpp#L199-L203

fjeremic:
@dhong44 since you are completing #3577 we can tackle this one after.

liqunl:
[`isLoadOfStaticFinalField`](https://github.com/eclipse/omr/blob/master/compiler/il/OMRNode.cpp#L8699) needs to have a check for direct load once this is done.

PR #4322, opened 2019-01-16
Set cpIndex to static Class sym ref

r30shah:
If cpIndex is passed while creating sym ref for static Class use it instead of setting it to -1.


r30shah:
@dsouzai @jdmpapin  Can I get your view on this change? I checked caller of `findOrCreateClassStaticsSymbol` and can only find two occurrences. 
1. https://github.com/eclipse/openj9/blob/91e786ab3d0ccbc6734636c855f690883089cabe/runtime/compiler/ilgen/Walker.cpp#L6245
2. https://github.com/eclipse/openj9/blob/91e786ab3d0ccbc6734636c855f690883089cabe/runtime/compiler/ilgen/Walker.cpp#L7405 
Which may even not going to get executed on Z with [this](https://github.com/eclipse/openj9/pull/4311/files#diff-eeab0f6a82448f563e51b3e7529826f7) changes as it will set `AccessStaticsIndirectly` on Z when SVM is enabled on Z. I thought still there is no reason why we do not set cpIndex to symRef if it is known.

dsouzai:
LGTM, but since this is out of my area of knowledge, I'll wait for @jdmpapin 's seal of approval.

jdmpapin:
Those are the only two callers that I see as well.

Based on the linked change from #4311, I was wondering first why `accessStaticsIndirectly` was still set for "old" AOT, but then more importantly, how these "class statics" symrefs ever worked in AOT in the first place.

The answer is that they aren't really used in AOT. Both call sites were guarded to require `getAccessStaticsIndirectly()` *and* `!compileRelocatableCode()`. The latter condition has been relaxed to allow AOT when the SVM is in use, but currently that only happens on x86, which uses direct static accesses (unless the environment variable `TR_AccessStaticsIndirectly` is set, which must be broken now if it wasn't already). Once the SVM is enabled on z, then based on the above-linked change, compilations using it won't enter here unless targeting z9 or below, or `disableDirectStaticAccessOnZ` is specified in the JIT options. For the time being, I don't expect these cases to work any better under the SVM than they would have in "old" AOT, since we haven't changed the way in which statics are relocated. So if the SVM is enabled on z before indirect static access is removed (#3610), the conditions guarding the call sites will likely have to be reverted so that they disallow all AOT again.

The callers also require the static to be resolved, which leads me to believe we don't have a run-time resolution path for it.

I don't think it's necessarily a *problem* to put the given CP index onto the symref here, but I do think it's misleading, because it must be unused: we need it neither for resolution, nor for AOT, and (I assume...) the indirect accesses have been working fine with the CP index explicitly missing. Furthermore, the (inlined call site index, CP index) pair is non-unique. Two or more (resolved) static field references for statics belonging to the same class will share the same "class statics" symref. So if we wanted to start using the CP index for something, we would probably need further changes to prevent any sharing that's incompatible with the new use case.

fjeremic:
> I don't think it's necessarily a _problem_ to put the given CP index onto the symref here, but I do think it's misleading, because it must be unused: we need it neither for resolution, nor for AOT, and (I assume...) the indirect accesses have been working fine with the CP index explicitly missing. Furthermore, the (inlined call site index, CP index) pair is non-unique. Two or more (resolved) static field references for statics belonging to the same class will share the same "class statics" symref. So if we wanted to start using the CP index for something, we would probably need further changes to prevent any sharing that's incompatible with the new use case.

Given this I'm inclined to merge this as having a correct but unused CP index available is IMO better than having an invalid one that also shouldn't be used. I'll wait until EOD and if there is no other concerns I'll go ahead with the merge.

Regarding the z9 or below issue. Yeah this is a mess. It's unlikely it worked before either. Thankfully z9 support won't be around for much longer I think.

jdmpapin:
> ... having a correct but unused CP index available is IMO better than having an invalid one that also shouldn't be used.

On later reflection I think the latter part of my comment was more pertinent, repeated here:
>  Furthermore, the (inlined call site index, CP index) pair is non-unique. Two or more (resolved) static field references for statics belonging to the same class will share the same "class statics" symref.

What this means is that the CP index is not just unused, but I expect it to be unus*able*. If we put a valid CP index there, it will *look* usable, but that will just be an illusion in any case where we actually care whose CP entry it is, and generally we do care whenever a CP index is needed at all. Using the CP index might even seem at first to work correctly, if somebody is working on a case where the sharing doesn't kick in, only for the bug to get exposed later.

fjeremic:
I see. Ok, let's reprupose this PR then. @r30shah let's leave the CP index as -1 and put a comment there instead summarizing the explanations provided by @jdmpapin here. You can also point the reader to this PR (#4322) if they wish to get more information.

r30shah:
Tried to accommodate comments made in https://github.com/eclipse/openj9/pull/4322#issuecomment-456176121 in https://github.com/eclipse/openj9/pull/4322/commits/96eaf7e1f25a8e52f4d9ad9b6aa3b47c15ab9948 
@jdmpapin  let me know if this is fine?

jdmpapin:
I'm okay with the comment. Can the commits be squashed so that we don't have a pair of commits that first change -1 to `cpIndex`, and then change it back again?

PR #4325, opened 2019-01-16
Respect resolving static method low-tag on X86-32

0dvictor:
Interpreter may low-tag the RAM method when resolving static method.
When the lowest bit is set, JIT should skip code patch.


andrewcraik:

Issue #4312, opened 2019-01-16
Implement JVM_CopySwapMemory


lumpfish:
With the 19_01 updates, openjdk has introduced a new JVM method JVM_CopySwapMemory which openj9 needs to implement.

```
07:01:03 /home/jenkins/workspace/Build-JDK8-linux_x86-64_cmprssptrs/build/linux-x86_64-normal-server-release/jdk/objs/libjava/Bits.o: In function `Java_java_nio_Bits_copySwapMemory0':
07:01:03 /home/jenkins/workspace/Build-JDK8-linux_x86-64_cmprssptrs/jdk/src/share/native/java/nio/Bits.c:33: undefined reference to `JVM_CopySwapMemory'
```

The previous build had the following natives which were merged into `Java_java_nio_Bits_copySwapMemory0;`
```
Java_java_nio_Bits_copyFromShortArray;
Java_java_nio_Bits_copyToShortArray;
Java_java_nio_Bits_copyFromIntArray;
Java_java_nio_Bits_copyToIntArray;
Java_java_nio_Bits_copyFromLongArray;
Java_java_nio_Bits_copyToLongArray;
```

DanHeidinga:
We need a description of this function to be able to implement it.

lumpfish:
function prototype is

JVM_CopySwapMemory(JNIEnv *env, jobject srcObj, jlong srcOffset, jobject dstObj, jlong dstOffset, jlong size, jlong elemSize)

The function needs to copy size bytes of data from the srcObj+srcOffset location to the dstObj+dstOffset location

elemSize must be 2, 4, or 8 and denotes whether the data contains short, long or double data types.

As the data is copied it is reversed, so for
elemSize=2, byte order 1,2 becomes 2,1
elemSize=4, byte order 1,2,3,4 becomes 4,3,2,1
elemSize=8, byte order 1,2,3,4,5,6,7,8 becomes 8,7,6,5,4,3,2,1

The data may be in native or heap memory.

The original openjdk bug the new method addresses concerned the jdk ('C) compiler not necessarily aligning addresses at a boundary for the particular data type.  So the implementation needs to make sure that alignment is guaranteed (presumably in the destination buffer).

gacholio:
I'll implement this ASAP.

gacholio:
Assuming the function is "void JNICALL".

gacholio:
[...]

PR #4328, opened 2019-01-16
Initial implementation of JVM_CopySwapMemory

gacholio:
New function to copy and endian flip memory of various sizes. The
current implementation has given no thought to performance.

Closes: #4312

[ci skip]

### DIFF
--- a/runtime/j9vm/j8vmi.c
+++ b/runtime/j9vm/j8vmi.c
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2002, 2017 IBM Corp. and others
+ * Copyright (c) 2002, 2019 IBM Corp. and others
  *
  * This program and the accompanying materials are made available under
  * the terms of the Eclipse Public License 2.0 which accompanies this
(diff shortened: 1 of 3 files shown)


DanHeidinga:
@lumpfish Can you test the Extensions code with this fix?

runtime/j9vm/j8vmi.c
```diff
@@ -99,3 +99,114 @@ JVM_GetTemporaryDirectory(JNIEnv *env)
...
+void JNICALL
+JVM_CopySwapMemory(JNIEnv *env, jobject srcObj, jlong srcOffset, jobject dstObj, jlong dstOffset, jlong size, jlong elemSize)
+{
+	void *srcBytes = NULL;
+	void *dstBytes = NULL;
+	U_8 *srcAddr = NULL;
+	U_8 *dstAddr = NULL;
+	/* Ensure that size is a non-negative multiple of elemSize */
+	Assert_SC_true(size >= 0);
+	Assert_SC_true(0 == (size & elemSize));
```
keithc-ca: I think you meant `0 == (size % elemSize)`.

runtime/j9vm/j8vmi.c
```diff
@@ -99,3 +99,114 @@ JVM_GetTemporaryDirectory(JNIEnv *env)
...
+			elemCount -= 1;
+		}
+		break;
+	}
+	default:
+		Assert_SC_unreachable();
+		break;
+	}
+	if (NULL != srcObj) {
+		(*env)->ReleasePrimitiveArrayCritical(env, srcObj, srcBytes, 0);
```
keithc-ca: Shouldn't the final argument to `ReleasePrimitiveArrayCritical()` be `JNI_COMMIT` here and in the similar call below?

runtime/j9vm/j8vmi.c
gacholio: I did, and noticed it in a review with Dan and forgot to fix...

runtime/j9vm/j8vmi.c
gacholio: No. ABORT means do nothing, COMMIT means copy back but don't free the back buffer (if it was allocated). 0 (no special flags) is the right answer virtually 100% of the time.

runtime/j9vm/j8vmi.c
gacholio: Fixed.

DanHeidinga:
Jenkins test sanity win,plinux jdk8

DanHeidinga:

lumpfish:
Building and testing the extensions code with branch gacholio/jvm.  jdk built ok, tests running now.

DanHeidinga:
I'm going to merge this.  Any additional changes required based on @lumpfish's testing can be done in separate prs

lumpfish:
Got some test failures running the openjdk jdk_nio tests.  @gachiolio is investigating.

runtime/j9vm/j8vmi.c
gacholio: Sorry, you are in fact correct. Since this is the source array, there's no need to copy back to the heap. I'm not sure it matters in this case, since we release the destination array afterwards, so if the source and dest were the same array, only one copy back would take place. I've update this in a later PR.

keithc-ca: To make sure we're on the same page, the source array should be released with `JNI_ABORT` and the destination with mode `0`, right?

gacholio: Right - ABORT means free the buffer but don't copy back to heap (assuming that we didnt get dirtect access to the heap). Since we're not modifying the source array unless it's identical to the dest array, there's no need to copy back.

Issue #4093, opened 2018-12-18
JITaaS client should perform synchronous compilations locally


mpirvu:
For synchronous compilations the application needs to wait for the compilation result, therefore the latency of such compilations is crucial for performance. Since remote compilations (performed at the server) have much higher latency due to network communication, it may be beneficial to schedule these synchronous compilations at the client instead. However, this cancels the advantage of JITaaS, namely elimination of CPU and memory spikes which happen during compilation. To address this issue I propose we schedule a cheap (cold) synchronous compilation at the client immediately followed by an expensive re-compilation performed at the server.
This heuristic should be enabled only when -Xjit:enableJITaaSHeuristics is present so that we can do proper JITaaS testing otherwise.

mpirvu:
This item was implemented, so closing the issue

PR #4330, opened 2019-01-16
Schedule remote sync compilation to local first

a7ehuo:
Remote synchronous compilations have much higher latency due to network communication,
It’s proposed to schedule a cheap (cold) synchronous compilation at the client
immediately followed by an expensive re-compilation performed at the server.
This heuristic should be enabled only when -Xjit:enableJITaaSHeuristics is present.

Issue: #4093

### DIFF
--- a/runtime/compiler/control/CompilationThread.cpp
+++ b/runtime/compiler/control/CompilationThread.cpp
@@ -4604,7 +4604,7 @@ TR::CompilationInfo::addMethodToBeCompiled(TR::IlGeneratorMethodDetails & detail
       increaseQueueWeightBy(entryWeight);
 
       if (TR::Options::getCmdLineOptions()->getVerboseOption(TR_VerboseCompileRequest))
-         TR_VerboseLog::writeLineLocked(TR_Vlog_CR, "%p     Added entry %p of weight %d to comp queue. Now Q_SZ=%d weight=%d",
+         TR_VerboseLog::writeLineLocked(TR_Vlog_CR, "%p   Added entry %p of weight %d to comp queue. Now Q_SZ=%d weight=%d",
             _jitConfig->javaVM->internalVMFunctions->currentVMThread(_jitConfig->javaVM), cur, entryWeight, getMethodQueueSize(), getQueueWeight());
 
       // Examine if we need to activate a new thread
@@ -5624,6 +5624,7 @@ void *TR::CompilationInfo::compileOnSeparateThread(J9VMThread * vmThread, TR::Il
 
    TR_ASSERT(asynchronousCompilation() || requireAsyncCompile != TR_yes, "We cannot require async compilation when it is turned off");
 
+   bool forcedSync = false;
    bool async = asynchronousCompilation() && requireAsyncCompile != TR_no;
    if (async)
       {
@@ -5678,6 +5679,7 @@ void *TR::CompilationInfo::compileOnSeparateThread(J9VMThread * vmThread, TR::Il
       if (!async)
          {
          debugPrint("forced synchronous");
+         forcedSync = true;
          }
       else
          {
@@ -5882,6 +5884,7 @@ void *TR::CompilationInfo::compileOnSeparateThread(J9VMThread * vmThread, TR::Il
 
    if (entry == NULL)
       {
+      releaseCompMonitor(vmThread);
       if (compErrCode)
          *compErrCode = compilationFailure;
       return 0;  // We couldn't add a method entry to be compiled.
@@ -5923,8 +5926,10 @@ void *TR::CompilationInfo::compileOnSeparateThread(J9VMThread * vmThread, TR::Il
    // for an asynchronous compilation (very rare event possible due to some races)
    //
    if (!(*queued) && entry->_changedFromAsyncToSync && requireAsyncCompile != TR_yes)
+      {
       async = false;
-
+      forcedSync = true;
+      }
    entry->_async = async;
 
    // no need to grab slot monitor or releasing VM access if we are doing async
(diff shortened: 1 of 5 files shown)


runtime/compiler/control/CompilationThread.cpp
```diff
@@ -6846,6 +6921,37 @@ TR::CompilationInfoPerThreadBase::preCompilationTasks(J9VMThread * vmThread,
...
                doLocalCompilation = true;
 
+#if DEV_DBG_ENABLE_CHANGE
+            // If it is a remote sync compilation, change it to a local sync compilation.
+            // After the local compilation is completed, a remote async compilation will be scheduled
+            if (!doLocalCompilation &&
+                !entry->_async &&
+                !entry->_changedToLocalSyncComp &&
+                TR::Options::getCmdLineOptions()->getOption(TR_EnableJITaaSHeuristics) &&
+                !TR::CompilationInfo::isCompiled(method) &&
```
mpirvu: `TR::CompilationInfo::isCompiled(method)` test should be eliminated. We don't want to do this only for first time sync compilations.

runtime/compiler/control/CompilationThread.cpp
```diff
@@ -6846,6 +6921,37 @@ TR::CompilationInfoPerThreadBase::preCompilationTasks(J9VMThread * vmThread,
...
+                !TR::Options::getCmdLineOptions()->getOption(TR_DontDowngradeToCold) &&
```
mpirvu: Replace the test on `TR_DontDowngradeToCold` with a test on opt level. If the optLevel is 'warm' or higher then we don't to do this optimization. In theory the test on optlevel is not even required because of the piece that transforms cheap compilations into local ones, but the code becomes more explicit that way.

runtime/compiler/control/CompilationThread.cpp
```diff
@@ -6035,6 +6070,44 @@ void *TR::CompilationInfo::compileOnSeparateThread(J9VMThread * vmThread, TR::Il
...
+      // queue a new entry for this method as a remote async compilation
+      acquireCompMonitor(vmThread);
+      if ((startPC && startPC != oldStartPC) &&
+           entry->_changedToLocalSyncComp)
+         {
+         TR::IlGeneratorMethodDetails methodDetails(method);
+         CompilationPriority compPriority = CP_ASYNC_NORMAL;
+         bool queuedForRemote = false;
+         bool isAsync = true;
+         TR_OptimizationPlan *plan = TR_OptimizationPlan::alloc(entry->_optimizationPlan->getOptLevel());
```
mpirvu: Don't we need to take the optLevel from optimizationPlan->_origOptLevel?

a7ehuo: The entry's level is restored in `postCompilationTasks()`:
https://github.com/eclipse/openj9/compare/jitaas..4f62ac3ab1e93a140183dd71f62ef13d1f0fa61f#diff-08779930c70bee9d159bc2c125154c30R7120

We could also directly use `entry->_origOptLevel` here to create the new optimization plan.

mpirvu:
@a7ehuo please let me know when the PR is ready for another review. Thanks

a7ehuo:
@mpirvu will do, testing the new update.

a7ehuo:
@mpirvu Please check the latest change. Thank you

runtime/compiler/control/CompilationThread.cpp
```diff
@@ -6849,6 +6896,29 @@ TR::CompilationInfoPerThreadBase::preCompilationTasks(J9VMThread * vmThread,
                (TR::Options::getCmdLineOptions()->getOption(TR_EnableJITaaSHeuristics) || localColdCompilations))
                doLocalCompilation = true;
 
+            // If this is a remote sync compilation, change it to a local sync compilation.
+            // After the local compilation is completed successfully, a remote async compilation
+            // will be scheduled in compileOnSeparateThread().
+            TR::IlGeneratorMethodDetails & details = entry->getMethodDetails();
+            if (!doLocalCompilation &&
+                !entry->_async &&
+                !entry->_changedToLocalSyncComp &&
```
mpirvu: Do we need the test on _changedToLocalSyncComp? I guess the idea of the test to prevent lowering the optLevel for the recompilation, but maybe the _async test already does that.

runtime/compiler/control/CompilationThread.cpp
```diff
@@ -6028,6 +6033,48 @@ void *TR::CompilationInfo::compileOnSeparateThread(J9VMThread * vmThread, TR::Il
       if (compErrCode)
          *compErrCode = (TR_CompilationErrorCode)entry->_compErrCode;
 
+      // If a remote sync compilation has been changed to a local sync compilation,
```
mpirvu: I wonder if we need to put all this code somewhat below when we acquire comp monitor. There is this part of the code
```
      // Release the monitor now
      //
      debugPrint(vmThread, "\tapplication thread releasing AppMonitor\n");
      debugPrint(vmThread, "-AM-", entry);

      entry->releaseSlotMonitor(vmThread);

      // Grab the compilation mutex to be able to decrement numThreadsWaiting
      // and access entryShouldBeDeallocated
      // When numThreadsWaiting reaches 0, the entry can be reused or deallocated
      //
      acquireCompMonitor(vmThread);
      entry->_numThreadsWaiting--;
```
which suggests that `_numThreadsWaiting` is protected by the compilation mutex and when you test it we don't have it locked. This change also prevents us from acquiring the comp monitor then releasing it and then acquiring again.

runtime/compiler/control/CompilationThread.cpp
```diff
@@ -6028,6 +6033,48 @@ void *TR::CompilationInfo::compileOnSeparateThread(J9VMThread * vmThread, TR::Il
...
+      //   compilation from an async compilation, it probably means we can’t do an
+      //   async compilation. Don't queue the second async remote compilation.
+      if (entry->_changedToLocalSyncComp &&
+          (entry->_numThreadsWaiting <= 1) &&
+          !forcedSync &&
+          (startPC && startPC != oldStartPC))
+         {
+         acquireCompMonitor(vmThread);
+
+         void *currOldStartPC = startPCIfAlreadyCompiled(vmThread, details, oldStartPC);
```
mpirvu: startPCIfAlreadyCompiled(.., oldStartPC) returns NULL if the method body defined by oldStartPC has not been recompiled. In this case we want to use the existing startPC for addMethodToBeCompiled(). If startPCIfAlreadyCompiled(.., oldStartPC) return something, then it means that a recompilation already happened and we don't have to do anything

runtime/compiler/control/CompilationThread.cpp
```diff
@@ -6028,6 +6033,48 @@ void *TR::CompilationInfo::compileOnSeparateThread(J9VMThread * vmThread, TR::Il
...
+               {
+               entryRemoteCompReq->_async = isAsync;
+
+               if (getMethodQueueSize() <= 1 ||
+                  getNumCompThreadsJobless() > 0) // send notification if any thread is sleeping on comp monitor waiting for suitable work
+                  {
+                  getCompilationMonitor()->notifyAll();
+                  }
+               }
+            if ((!queuedForRemote) || (!entryRemoteCompReq))
```
mpirvu: The test on queuedForRemote is enough because if `addMethodToBeCompiled` cannot allocate an entry, then it's not going to be queued for sure.

runtime/compiler/control/CompilationThread.cpp
a7ehuo: I can move `acquireCompMonitor` and `releaseCompMonitor` outside the `if` condition check so that checking `_numThreadsWaiting` is guarded with locks. Or pull the code before `entry->_numThreadsWaiting--`. Let me test it and see how it goes.

PR #4340, opened 2019-01-17
Reducing messages for ResolvedMethod_getClassFromConstantPool

satbir-ca:
Persistent map cache is created in the ClassInfo for the constant class pool.
Cache is checked before posting message to the client.

### DIFF
--- a/runtime/compiler/env/j9methodServer.cpp
+++ b/runtime/compiler/env/j9methodServer.cpp
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2018, 2018 IBM Corp. and others
+ * Copyright (c) 2018, 2019 IBM Corp. and others
  *
  * This program and the accompanying materials are made available under
  * the terms of the Eclipse Public License 2.0 which accompanies this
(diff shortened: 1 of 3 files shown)


satbir-ca:
@mpirvu please review the code.
On the "-version" test i am seeing the 77% message drop.
Complete sanity testing is still to run.

runtime/compiler/env/j9methodServer.cpp
```diff
@@ -146,8 +146,31 @@ TR_ResolvedJ9JITaaSServerMethod::staticAttributes(TR::Compilation * comp, I_32 c
 TR_OpaqueClassBlock *
 TR_ResolvedJ9JITaaSServerMethod::getClassFromConstantPool(TR::Compilation * comp, uint32_t cpIndex, bool)
    {
+   TR::CompilationInfoPerThread *compInfoPT = _fe->_compInfoPT;
```
mpirvu: The original code returns null if cpIndex=-1
```
   if (cpIndex != -1 &&
       !((fej9->_jitConfig->runtimeFlags & J9JIT_RUNTIME_RESOLVE) &&
         !comp->ilGenRequest().details().isMethodHandleThunk() && // cmvc 195373
         performTransformation(comp, "Setting as unresolved class from CP cpIndex=%d\n",cpIndex) )&&
       (resolvedClass = fej9->_vmFunctionTable->resolveClassRef(fej9->vmThread(), cp, cpIndex, J9_RESOLVE_FLAG_JIT_COMPILE_TIME)))
      {
      result = fej9->convertClassPtrToClassOffset(resolvedClass);
      }
   else
      {
      INCREMENT_COUNTER(fej9, unresolvedClassRefs);
      }
```
It's a strange comparison given that cpIndex is declared as uint32_t.
We should add that condition to avoid sending messages whenever cpIndex is -1.

satbir-ca:
updated

PR #4342, opened 2019-01-17
Fixes for JVM_CopySwapMemory

gacholio:
Change the native to safely copy the memory using memmove, then flip
all the elements in the destination. Fix the primitive array case to
remove the Unsafe.arrayBaseOffset() added by the java caller.

Related: #4312

[ci skip]

### DIFF
--- a/runtime/j9vm/j8vmi.c
+++ b/runtime/j9vm/j8vmi.c
@@ -103,6 +103,9 @@ JVM_GetTemporaryDirectory(JNIEnv *env)
 /**
  * Copies memory from one place to another, endian flipping the data.
  *
+ * Implementation of native java.nio.Bits.copySwapMemory0(). The single java caller
+ * has ensured all of the parameters are valid.
+ *
  * @param [in] env Pointer to JNI environment
  * @param [in] srcObj Source primitive array (NULL means srcOffset represents native memory)
  * @param [in] srcOffset Offset in source array / address in native memory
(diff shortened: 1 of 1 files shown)


DanHeidinga:
Not much point in running PR testing on this as its only called by code not yet delivered.

@gacholio Can you confirm when the internal (?) build you launched passes?

gacholio:
Will do - I have already manually compiled the file. Hopefully the testing will show improved results.

gacholio:
I've tested the code in isolation (i.e. not as part of nio) and it seems to perform as expected - overlap works, flipping seems to be happening.

runtime/j9vm/j8vmi.c
```diff
@@ -204,7 +193,7 @@ JVM_CopySwapMemory(JNIEnv *env, jobject srcObj, jlong srcOffset, jobject dstObj,
 		break;
 	}
 	if (NULL != srcObj) {
-		(*env)->ReleasePrimitiveArrayCritical(env, srcObj, srcBytes, 0);
+		(*env)->ReleasePrimitiveArrayCritical(env, srcObj, srcBytes, JNI_ABORT);
```
DanHeidinga: Why `JNI_ABORT`?

DanHeidinga:
@lumpfish Can you test with this change as well?

DanHeidinga:
Sufficient to see this compiled once as the corresponding OJDK change isn't available in Extensions yet

Jenkins compile zlinux jdk8

runtime/j9vm/j8vmi.c
gacholio: Because the source array is not modified, there's no need to copy back to the heap (in the case where we didnt get direct access).

gacholio:
Looks like the issue is that array offsets have Unsafe.arrayBaseOffset added to them, which the native is not accounting for.

lumpfish:
Rerun the openjdk nio tests with a jdk built with the jdk8u202-b08 updates and @gacholio 's openj9 overlap branch.  Tests now pass.

PR #4344, opened 2019-01-17
Update JIT bytecode enum to match a change from VM

cathyzhyi:
A recent change in VM added 2 new opcodes and changed the
internal bytecodes. Update the JIT counter part in this
commit.

https://github.com/eclipse/openj9/pull/3416

### DIFF
--- a/runtime/compiler/env/VMJ9.cpp
+++ b/runtime/compiler/env/VMJ9.cpp
@@ -22,6 +22,10 @@
 
 #define J9_EXTERNAL_TO_VM
 
+#if defined (_MSC_VER) && (_MSC_VER < 1900)
+#define snprintf _snprintf
+#endif
+
 #include "env/VMJ9.h"
 
 #include <algorithm>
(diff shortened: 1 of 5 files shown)


runtime/compiler/env/VMJ9.cpp
```diff
@@ -4722,6 +4722,14 @@ TR_J9VMBase::getOverflowSafeAllocSize()
    return result;
    }
 
+void
+TR_J9VMBase::unsupportedByteCode(TR::Compilation * comp, U_8 opcode)
+   {
+   char errMsg[40];
+   sprintf(errMsg, "bytecode %d not supported by JIT", opcode);
```
andrewcraik: snprintf with some error checking would be much safer here

andrewcraik: should we enhance the unknonByteCode case like this as well - it seems a much better error message since you are "in the area"

andrewcraik:
@DanHeidinga Should there be a comment referencing the JIT enum somewhere in the VM to try and prevent something like this happening again?

DanHeidinga:
Already done - I opened a PR for shortly after I became aware of the issue: https://github.com/eclipse/openj9/pull/4341

runtime/compiler/env/VMJ9.cpp
cathyzhyi: Updated the code according to comments. Thanks!

cathyzhyi: Please ingore the comment about breakpoint in `unknownByteCode` for now. I will change that comment when adding assert_fatal in a separate PR.

andrewcraik:
Ok changes look ok, but I am going to wait for @cathyzhyi to give me a thumbs up on this being done because I think three is another change alluded to in one of her previous comments.

cathyzhyi:
> Ok changes look ok, but I am going to wait for @cathyzhyi to give me a thumbs up on this being done because I think three is another change alluded to in one of her previous comments.

This is one is ready to merge. I will open a separate PR to add assert_fatal when we hit unknown opcodes.

andrewcraik:

cathyzhyi:
@andrewcraik seems windows build doesn't support `snprintf`. I will change it back to `sprintf` then?

cathyzhyi:
added ifdef check for _MSC_VER to fix windows build failure.

andrewcraik:
https://stackoverflow.com/questions/2915672/snprintf-and-visual-studio-2010 might be worth looking at since not having the length bound is poor practice in general.

cathyzhyi:
the failures are because of infrastructure issues.

DanHeidinga:

DanHeidinga:
Jenkins copyright check

DanHeidinga:
Jenkins line endings check

andrewcraik:
Issue in x builds seems to be a repeat of a problem with javac - bytecode renumbering looks good - merging.

PR #3679, opened 2018-11-14
Consolidate Relocation Records

dsouzai:
https://github.com/eclipse/openj9/issues/3031

This PR consolidates a few Relocation Records, removing code duplication. It is the first of several PRs.

PR #4345, opened 2019-01-17
Add TR_MethodCallAddress implementation for x86

ymanton:
This patch introduces an x86 implementation for the newly added OMR relocation `TR_MethodCallAddress`, which is suitable for relocating a relative call to a fixed, known target address.

Depends on https://github.com/eclipse/omr/pull/3427.

### DIFF
--- a/runtime/compiler/codegen/J9AheadOfTimeCompile.cpp
+++ b/runtime/compiler/codegen/J9AheadOfTimeCompile.cpp
@@ -203,6 +203,15 @@ J9::AheadOfTimeCompile::initializeCommonAOTRelocationHeader(TR::IteratedExternal
          }
          break;
 
+      case TR_MethodCallAddress:
+         {
+         TR_RelocationRecordMethodCallAddress *mcaRecord = reinterpret_cast<TR_RelocationRecordMethodCallAddress *>(reloRecord);
+
+         mcaRecord->setEipRelative(reloTarget);
+         mcaRecord->setAddress(reloTarget, relocation->getTargetAddress());
+         }
+         break;
+
       default:
          return cursor;
       }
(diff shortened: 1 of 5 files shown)


ymanton:
Jenkins test sanity all jdk8 depends eclipse/omr#3427

ymanton:
> Also, to save me the trouble of eventually having to clean this up, do you mind using the newer way of storing the relo header data (eg see #3679)

Done. Let me know if that's the change you had in mind.

> What's an example of when the call target is always going to be at that fixed offset in multiple (AOT) runs? I figure for JaaS there is no issue because the environment hasn't changed, but in a general relocatable compile, what ensures that the call target is always going to be at a fixed offset?

Outside of a few niche cases (e.g. vsyscalls on Linux and millicode routines on AIX are always at fixed, known addresses) there aren't any that I know of.

I don't have any plans to use this for AOT, if someone wants to do that they'll have to change the payload data and how it's applied to make it work. My interest is for JITaaS only, where the target's address is sufficient.

runtime/compiler/codegen/J9AheadOfTimeCompile.cpp
```diff
@@ -1742,6 +1750,19 @@ J9::AheadOfTimeCompile::dumpRelocationData()
             }
             break;
 
+         case TR_MethodCallAddress:
```
dsouzai: This should be added to `J9::AheadOfTimeCompile::dumpRelocationHeaderData` rather than here.

ymanton: Can you clarify why we have methods like `initializeCommonAOTRelocationHeader` and `dumpRelocationHeaderData` that manipulate the entire record, not just the header?

In the previous case I was writing to both a header field and the record's data portion, and in this case I'm dumping the record's data, not the header.

dsouzai: I'm currently in the process of consolidating the way we write out the relocation header data to be consistent with the way we read out the relocation header data (https://github.com/eclipse/openj9/issues/3031). The latter is done via the `TR_RelocationRecord`s (and sub classes), so the former should also use the same interface. Eventually I want to clean up the way we use raw pointer manipulation in `dumpRelocationData` but first I wanted to consolidate everything before I started cleaning that up.

dsouzai: I guess another reason for using the `TR_RelocationRecord`s is for consistency; plus, eventually @mstoodle  wants the the binary templates moved to `RelocationRecord.cpp` and so this function will not be able to use them anymore.

ymanton: Changes made.

dsouzai:
Jobs failed because of https://github.com/eclipse/openj9/issues/4397

dsouzai:
Jenkins test sanity osx,xlinux,win jdk11

dsouzai:
Jenkins test sanity osx,xlinux,win jdk11

PR #4349, opened 2019-01-18
Delete package-private getChars( byte[]) method

DanHeidinga:
As discussed in https://github.com/eclipse/openj9/pull/4224#issuecomment-454577251 , this method was necessary for zOS but not anymore.  Now that 9b148 is removed, it can be deleted.


fjeremic:
Thanks for following up on this @DanHeidinga! It is much appreciated.

PR #4350, opened 2019-01-18
AArch64: Add files needed for linux aarch64 build

knn-k:
This commit adds files needed for building for linux aarch64.

### DIFF
--- a/buildspecs/linux_aarch64.spec
+++ b/buildspecs/linux_aarch64.spec
@@ -1,0 +1,246 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+Copyright (c) 2019, 2019 IBM Corp. and others
+
+This program and the accompanying materials are made available under
+the terms of the Eclipse Public License 2.0 which accompanies this
+distribution and is available at https://www.eclipse.org/legal/epl-2.0/
+or the Apache License, Version 2.0 which accompanies this distribution and
+is available at https://www.apache.org/licenses/LICENSE-2.0.
+
+This Source Code may also be made available under the following
+Secondary Licenses when the conditions for such availability set
+forth in the Eclipse Public License, v. 2.0 are satisfied: GNU
+General Public License, version 2 with the GNU Classpath
+Exception [1] and GNU General Public License, version 2 with the
+OpenJDK Assembly Exception [2].
+
+[1] https://www.gnu.org/software/classpath/license.html
+[2] http://openjdk.java.net/legal/assembly-exception.html
+
+SPDX-License-Identifier: EPL-2.0 OR Apache-2.0 OR GPL-2.0 WITH Classpath-exception-2.0 OR LicenseRef-GPL-2.0 WITH Assembly-exception
+-->
+<spec xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xmlns="http://www.ibm.com/j9/builder/spec" xsi:schemaLocation="http://www.ibm.com/j9/builder/spec spec-v1.xsd" id="linux_aarch64">
+	<name>Linux AArch64</name>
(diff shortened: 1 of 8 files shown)


buildspecs/linux_aarch64.spec
```diff
@@ -0,0 +1,246 @@
...
+		<flag id="jit_gcOnResolveSupport" value="true"/>
+		<flag id="jit_newInstancePrototype" value="true"/>
+		<flag id="jit_supportsDirectJNI" value="true"/>
+		<flag id="module_algorithm_test" value="true"/>
+		<flag id="module_bcutil" value="true"/>
+		<flag id="module_bcverify" value="true"/>
+		<flag id="module_cassume" value="true"/>
+		<flag id="module_cfdumper" value="true"/>
+		<flag id="module_codegen_common" value="true"/>
+		<flag id="module_codegen_ia32" value="true"/>
```
keithc-ca: If `ia32` is a reference to Intel, it would seem this (and similar flags) should be false.

knn-k: Thank you for pointing it out.  I have updated the file, and I also added lines to `buildspecs/j9.flags`.

I used `linux_arm_linaro.spec` as the base, and those `ia32` lines came from it.  I wonder why it has been OK for 32-bit ARM.

keithc-ca: I couldn't find any use of `module_codegen_ia32`, so it shouldn't matter what value is included in this PR.  I opened https://github.com/eclipse/openj9/issues/4398 to track cleaning up such unused flags.

keithc-ca:
I think this is ready. @knn-k, please remove the WIP prefix from the heading after you've made any final adjustments, if any.

knn-k:
Thank you @keithc-ca.  I removed the WIP.  I will add more changes in other PRs.

buildspecs/linux_aarch64.spec
```diff
@@ -0,0 +1,246 @@
...
+		<feature id="se60_26"/>
+		<feature id="se7"/>
+		<feature id="se70_27"/>
+	</features>
+	<source>
+		<project id="com.ibm.jvmti.tests"/>
+		<project id="compiler"/>
+	</source>
+	<flags>
+		<flag id="arch_arm" value="true"/>
```
DanHeidinga: This shouldn't be `true` for aarch64

knn-k: Thank you for pointing it out.  I corrected it, and the `j9vm_arch_*=`section in makelib/mkconstants.mk is now OK.

knn-k:
I changed configure_linux_aarch64.mk for `--enable-OMR_ENV_DATA64` and `CONFIGURE_ARGS += 'OMR_TARGET_DATASIZE=64'`.

runtime/makelib/mkconstants.mk.ftl
```diff
@@ -157,6 +157,8 @@ endif
 TR_HOST=TR_HOST_X86
 <#elseif uma.spec.processor.arm>
 TR_HOST=TR_HOST_ARM
+<#elseif uma.spec.processor.aarch64>
+TR_HOST=TR_HOST_ARM64
```
keithc-ca: Perhaps this should be `TR_HOST=TR_HOST_AARCH64` for consistency?

knn-k: OMR code already uses `TR_HOST_ARM64`.

keithc-ca: Yes, I see it now. I think that's an unfortunate situation, but not one to be addressed in this PR.

keithc-ca:
Jenkins compile zlinux jdk11

Issue #2368, opened 2018-07-10
Generate jvmti.h header for vm29 


tajila:
Since vm29 will support Java8, 9, 10, 11, etc. we will need to generate the jvmti header so that it is specific to the Java release version.

`#ifdefs` will not work because the j9 flags (j9cfg.h) are not included when compiling the OpenJDK natives.

The jvmti header file can be generated with m4.

pshipton:
@tajila is this likely to be completed for the 0.10.0 release, in the next couple of weeks? If not, where should this be targeted? The 0.11.0 release will be for end of Oct, and 0.12.0 for Jan.

pshipton:
@tajila is this likely to be completed for the 0.11.0 release, in the next week or so?

MarkQingGuo:
Just want to make sure if "JAVA_SPEC_VERSION" is the variable to represent the version of java? Does vm29 support JAVA12?

tajila:
> Just want to make sure if "JAVA_SPEC_VERSION" is the variable to represent the version of java? 
yes
> Does vm29 support JAVA12? 
yes
> Do I need to rewrite macros in J9cfg.h.in and J9cfg.h.ftl then insert into jvmti.h?
No, we will need a jvmti.h.m4 (or something similar). The `JAVA_SPEC_VERSION` will be passed to jvmti.h.m4 file and youll be able to do something like:
```
_IF([JAVA_SPEC_VERSION  > 10],
	// add java 11 code here ...
```

PR #4353, opened 2019-01-18
Create jvmti.h.m4 to generate different jvmti.h header for vm29

MarkQingGuo:
-Since #ifdefs will not work because the j9 flags (j9cfg.h) are not included when compiling the    OpenJDK natives, use m4 to generate different jvmti.h header
- Change runtime/module.xml and runtime/include/module.xml

Closes: #2368 


MarkQingGuo:
@tajila jvmti.h is not generated when building. Does it mean I have to modify CMakefile under `runtime`?

PR #4354, opened 2019-01-18
Disable DAA optimization in AOT mode

NigelYiboYu:
The AOT can not relocate direct calls in the DAA OOL paths.
If AOT is on and DAA makes a call in the OOL path, the target
won't be reached and it seg faults. Disable DAA for now until a proper
fix is ready.

### DIFF
--- a/runtime/compiler/optimizer/DataAccessAccelerator.cpp
+++ b/runtime/compiler/optimizer/DataAccessAccelerator.cpp
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2000, 2018 IBM Corp. and others
+ * Copyright (c) 2000, 2019 IBM Corp. and others
  *
  * This program and the accompanying materials are made available under
  * the terms of the Eclipse Public License 2.0 which accompanies this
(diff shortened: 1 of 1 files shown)


runtime/compiler/optimizer/DataAccessAccelerator.cpp
```diff
@@ -95,7 +95,10 @@ int32_t TR_DataAccessAccelerator::perform()
        !comp()->getOption(TR_MimicInterpreterFrameShape) &&
 
        // We cannot handle arraylets because hardware intrinsics act on contiguous memory
-       !comp()->generateArraylets()&& !TR::Compiler->om.useHybridArraylets())
+       !comp()->generateArraylets()&& !TR::Compiler->om.useHybridArraylets() &&
+       // AOT can not relocate direct calls in the DAA OOL paths. If AOT is on and DAA makes
+       // a call in the OOL path, the target won't be reached and it seg-faults. Disable DAA for now.
+       !(comp()->compileRelocatableCode() && !comp()->getOption(TR_UseSymbolValidationManager)))
```
r30shah: @fjeremic  We won't need to revert this back for SVM as this optimization will be disabled when in AOT it does not use SVM.  When we looked into the failure, it seems like AOT is doing right thing (Relocationg correct resolved ramCP) but fails in VM routine resolving this DAA method which we still don't understand why and needs some more digging.

fjeremic: @NigelYiboYu can you please dump your investigation details in an issue so we can tag someone from the VM side to help with the investigation if you are seeing crashes in the VM resolve code?

fjeremic: I'd also like to the comment here changed to reference this issue you will open. I'd like to see this fixed eventually and this code reverted.

NigelYiboYu: VM helper crash issue is here: https://github.com/eclipse/openj9/issues/4363

fjeremic: Thanks @NigelYiboYu can you please `TODO (#4363): AOT can not relocate...` directly in the code so that we know which issue to look at in the future to get this fixed?

NigelYiboYu: Added the TODO tag.

fjeremic:
Merging this as only the comment was changed and the previous builds passed.

PR #4357, opened 2019-01-18
Anchor arguments of MethodHandle invoke before expansion

liqunl:
The expansion will insert calls before the original call. Anchor the
arguments such that the evaluation of arguments is before the execution
of the call.

Notice that missing the anchoring does not cause functional issues
because the inserted calls are internal to the JIT and have no side
effect on arguments. However, it prevents us from folding argument that
is a static final field because the inserted calls are unsupported
potential OSR points.


liqunl:
@andrewcraik Can I have you review?

andrewcraik:

andrewcraik:

andrewcraik:

andrewcraik:
error looks to be a repeat of existing javac issue and unrelated to change - merging

PR #4358, opened 2019-01-19
Handle setCallCount if client IProfiler is disabled

dmitry-ten:
If interpreter profiling is disabled on the client,
the profiler instance is NULL. Handle this case by
checking if client IProfiler exists.


dmitry-ten:
@mpirvu please review

dmitry-ten:
Apparently, #3806 has already solved this issue. I encountered this again, because I was running an older build.
There is no need for this change, but we could still do it, because other IProfiler message handlers have this check.

PR #4359, opened 2019-01-21
Fix compile error

gacholio:
It's illegal to add an integer to a void pointer.

[ci skip]


DanHeidinga:
Jenkins compile zlinux,win jdk8

PR #4360, opened 2019-01-21
Replace enter/exit tracing method with isMethodTracingEnabled

a7ehuo:
In practice we never enable methodEnterTracing without methodExitTracing.
These two methods should be replaced with isMethodTracingEnabled().

### DIFF
--- a/runtime/compiler/rpc/protos/compile.proto
+++ b/runtime/compiler/rpc/protos/compile.proto
@@ -115,8 +115,6 @@ enum J9ServerMessageType
    VM_getStaticReferenceFieldAtAddress = 207;
    VM_getSystemClassFromClassName = 208;
    VM_isMethodTracingEnabled = 209;
-   // VM_isMethodEnterTracingEnabled = 209;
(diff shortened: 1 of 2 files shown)


a7ehuo:
@mpirvu This change is specific for the `jitaas` branch. Replacing enter/exit tracing method with isMethodTracingEnabled() for the rest of the code should take place on `master` branch.

a7ehuo:
@mpirvu These enter/exit methods are called in `prepareRelocateJITCodeAndData()` in `runtime/compiler/runtime/RelocationRuntime.cpp`.  `TR_RelocationRuntime::prepareRelocateJITCodeAndData()` only exists in 'jitaas' branch since it's called by `remoteCompile()` in `JITaaSCompilationThread.cpp`.  Therefore, these two locations have to be updated on `jitaas`. 

There are another two calls to enter/exit methods in the same file `runtime/compiler/runtime/RelocationRuntime.cpp` but in `prepareRelocateAOTCodeAndData()` , which shall be updated in `master` branch.

mpirvu:
Got it

PR #4361, opened 2019-01-21
Use the appropriate JIT API when enabling field watches

gacholio:
Use the old jitClassesRedefined API to flush the compilation queue if
inline field watches are not enabled. If they are enabled, use the new
jitFlushCompilationQueue API.

[ci skip]

### DIFF
--- a/runtime/codert_vm/decomp.cpp
+++ b/runtime/codert_vm/decomp.cpp
@@ -1374,7 +1374,13 @@ jitDataBreakpointAdded(J9VMThread * currentThread)
 
 		/* Toss the compilation queue */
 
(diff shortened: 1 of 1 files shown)


gacholio:
FYI @cathyzhyi

DanHeidinga:
Should the other block of code checking inlineWatches be rolled into this one?
```
		if (inlineWatches) {
			/* Make all future compilations inline the field watch code */
			jitConfig->inlineFieldWatches = TRUE;
		}
```

gacholio:
Suggested change made.

gacholio:

gacholio:
This should go into the current releases.

pshipton:
As it is merged, it will be in the 0.12 release.

PR #4371, opened 2019-01-21
Do not check if IProfiler is NULL on the client

dmitry-ten:
PR #3806 disables IProfiler on the server, if client has disabled it,
so the case when a server sends IProfiler message to the client, but the client
doesn't have an instance of IProfiler is no longer possible.
Hence remove all IProfiler checks on the client


dmitry-ten:
@mpirvu fyi

mpirvu:
Jenkins copyright check

mpirvu:
Jenkins line endings check

mpirvu:
Jenkins copyright check

Issue #2916, opened 2018-09-18
Crashes with -Xshareclasses enabled testing on x


pshipton:
https://ci.eclipse.org/openj9/job/Test-sanity.functional-JDK11-win_x86-64_cmprssptrs/17/tapResults/
https://ci.eclipse.org/openj9/job/Test-sanity.functional-JDK11-linux_x86-64_cmprssptrs/23/tapResults/

example
```
Testing: gaste001_multiple
Test start time: 2018/09/18 07:36:06 Coordinated Universal Time
Running command: "/home/jenkins/workspace/Test-sanity.functional-JDK11-linux_x86-64_cmprssptrs/openjdkbinary/j2sdk-image/bin/java"  -Xcompressedrefs -Xjit -Xgcpolicy:gencon -Xshareclasses:name=bcienabled,enableBCI  -Xdump    -Xjit:count=0 -agentlib:jvmtitest=test:gaste001 -agentlib:jvmtitest=test:gaste001 -cp "/home/jenkins/workspace/Test-sanity.functional-JDK11-linux_x86-64_cmprssptrs/openjdk-tests/TestConfig/scripts/testKitGen/../../../../jvmtest/functional/cmdLineTests/jvmtitests/jvmtitest.jar" com.ibm.jvmti.tests.util.TestRunner
Time spent starting: 3 milliseconds
Time spent executing: 8969 milliseconds
Test result: FAILED
Output from test:
 [ERR] Unhandled exception
 [ERR] Type=Segmentation error vmState=0x00040000
 [ERR] J9Generic_Signal_Number=00000004 Signal_Number=0000000b Error_Value=00000000 Signal_Code=00000001
 [ERR] Handler1=00007F9B69399E10 Handler2=00007F9B6877E740 InaccessibleAddress=00007F8AD67D8E1C
 [ERR] RDI=00007F8AD67D8E1C RSI=00000000FFE09A88 RAX=00000000FFEB4288 RBX=00000000FFE09A88
 [ERR] RCX=00000000010EF800 RDX=0000000001112800 R8=0000000001112641 R9=00007F9B445FE6F8
 [ERR] R10=0000000000000000 R11=00007F9AF40008D0 R12=00007F9AF40008D8 R13=00007F9AF40008C8
 [ERR] R14=00007F9B6A730020 R15=00007F9B445FB2D0
 [ERR] RIP=00007F8AD67D8E1C GS=0000 FS=0000 RSP=00000000011D2238
[...]

PR #4378, opened 2019-01-22
Consolidate 64-bit vs. 32-bit virtual register handling on Z codegen

fjeremic:
Due to the sheer size of the change it has been broken up into logical pieces each of which is in a separate commit to make reviews and code archeology easier to manage. Each commit body details what and why the change was made.

At a high level this change consolidates how we handle 64-bit virtual registers in the Z codegen. Previously we had two ways of making this distinction:

1. Using the `Is64BitReg` register flag
2. Using the `TR_GPR64` register kind

This change deprecates the `TR_GPR64` register kind in favor of the `Is64BitReg` register flag. The reason behind this choice is that the `TR_GPR64` register kind was "bolted on" to the Z register allocator at the same time we added support for using 64-bit registers on 32-bit. The register kind was only meant to be used by the 32-bit targets. The 64-bit targets were completely unaware of the `TR_GPR64` register kind, because they used a different mechanism to tell apart 64-bit vs. 32-bit virtual registers.

This caused all sorts of problems where the register allocator would get "out of sync" when both of these concepts were used to track 64-bit virtual registers. A catalyst which uncovered all these issues was enabling the use of 64-bit registers on 32-bit as part of #2916.

In the new world both 32-bit and 64-bit targets will use the `Is64BitReg` register flag to determine whether a particular virtual register holds a 32-bit or a 64-bit value. All registers allocated are assumed to be 32-bit. When a register is used within a 64-bit instruction (as marked in OMRInstOpCodeProperties.hpp) the virtual register will automatically be marked as a 64-bit register. Outside from this the developer need not care whether a virtual register is 32-bit or 64-bit as everything is handled automatically. The only exceptions to this is around special evaluators such as the GRA evaluators (`lRegLoad`, `aRegLoad`, etc.) and the extension/narrowing evaluators (`l2i`, `a2l`, etc.).


fjeremic:
Depends on eclipse/omr#3488

fjeremic:
Given the Z-specifics of this change and the sheer breadth I don't expect any one single committer to fully review the change so I'm kindly requesting reviews from @joransiu, @NigelYiboYu, @r30shah, @dchopra001, and @sehirst if you are familiar with any piece of the register allocator which I changed.

runtime/compiler/z/codegen/J9S390PrivateLinkage.cpp
```diff
@@ -200,7 +200,6 @@ TR::S390PrivateLinkage::initS390RealRegisterLinkage()
       freeingSSPDisabled = false;
 
    if (freeingSSPDisabled || disableFreeJITSSP != NULL ||
-       !(cg()->supportsJITFreeSystemStackPointer()) ||
```
joransiu: Was the removal of `!(cg()->supportsJITFreeSystemStackPointer())` an intended change in this commit?

fjeremic: Yes it was. If you look at a few lines above you see the following:

```
   bool freeingSSPDisabled = true;

   TR_J9VMBase *fej9 = (TR_J9VMBase *)(fe());

   if (cg()->supportsJITFreeSystemStackPointer())
      freeingSSPDisabled = false;
```

Thus `freeingSSPDisabled == !cg()->supportsJITFreeSystemStackPointer()` and `freeingSSPDisabled` is already part of the OR chain of the if statement on line 202 so querying `!(cg()->supportsJITFreeSystemStackPointer())` is redundant and hence can be removed.

fjeremic:
My internal testing is looking clean. I'm unleashing the builds. Note once the OMR merge happens OpenJ9 will stay broken until this is merged. I could not find a sane way of making a 10k line change without temporarily breaking OpenJ9. I think it's the lesser of two evils.

Jenkins test sanity all JDK8 depends eclipse/omr#3488

fjeremic:
Windows issues seem unrelated. I've opened #4495 to track this.

PR #4379, opened 2019-01-22
Fix VM info deallocation bug

dmitry-ten:
VM info was always freed, even when it wasn't initialized in the first place,
which leads to occasional segfaults. Fix by checking if it's not NULL first.


dmitry-ten:
@mpirvu please review

PR #4389, opened 2019-01-22
Donot process Unsafe call from cold block

liqunl:
Unsafe call from cold block could be a result of optimization, thus
should be not processed.


liqunl:

andrewcraik:

andrewcraik:
x failures are not new and the change looks simple - LGTM

PR #4392, opened 2019-01-23
Initial setup for JITaaS remote AOT

harryyu1994:
JITaaS remote AOT is by default disabled. To enable it, set environment variable "TR_EnableJITaaSRemoteAOT" to 1.

Issue: #4401 
[skip ci]

### DIFF
--- a/runtime/compiler/control/CompilationThread.cpp
+++ b/runtime/compiler/control/CompilationThread.cpp
@@ -898,6 +898,7 @@ TR::CompilationInfoPerThread::CompilationInfoPerThread(TR::CompilationInfo &comp
    _lastTimeThreadWasSuspended = 0;
    _lastTimeThreadWentToSleep = 0;
    _serverVM = nullptr;
+   _sharedCacheServerVM = nullptr;
 
    if (compInfo.getPersistentInfo()->getJITaaSMode() == SERVER_MODE)
       {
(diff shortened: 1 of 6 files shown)


harryyu1994:
FYI @mpirvu ; Hi Marius, I was reading over the code just to get a sense before taking on a sharedCacheServerVM API, and figured since this setup needs to be done anyways eventually, I wrote this piece of code. Could I please get an review? 
I feel like this can be merged once it's reviewed because the code isn't enabled yet.

runtime/compiler/control/CompilationThread.cpp
```diff
@@ -6772,7 +6774,7 @@ TR::CompilationInfoPerThreadBase::preCompilationTasks(J9VMThread * vmThread,
       TR::IlGeneratorMethodDetails & details = entry->getMethodDetails();
 
       eligibleForRelocatableCompile =
-            _compInfo.getPersistentInfo()->getJITaaSMode() != SERVER_MODE && // we do not allow AOT compilations in JITaaS Server Mode yet
+            _compInfo.getPersistentInfo()->getJITaaSMode() != SERVER_MODE && // code for out-of-process AOT compilation is Work In Progress
```
mpirvu: We should probably remove this comment altogether

runtime/compiler/control/CompilationThread.cpp
```diff
@@ -6824,11 +6826,24 @@ TR::CompilationInfoPerThreadBase::preCompilationTasks(J9VMThread * vmThread,
...
+         entry->setRemoteCompReq();
+         }
       }
    else if (entry->isOutOfProcessCompReq())
       {
-      _vm = TR_J9VMBase::get(_jitConfig, vmThread, TR_J9VMBase::J9_SERVER_VM);
-      entry->_useAotCompilation = false;
+      // is this really needed? TR::CompilationInfoPerThreadRemote::processEntry(...) already sets the _vm for us
+      // feels like this is redundant code for JITaaS Server, in my opinion we should remove this one, as the other one might be necessary
+      if (entry->_useAotCompilation)
```
mpirvu: I agree that we can avoid setting  the VM again. Needs serious testing though in all variations (with AOT/without, etc)
We should keep a comment that the VM has been set in processEntry()

mpirvu:
I can merge this once the minor observations are implemented and some testing is done.

harryyu1994:
Testings are done. Currently need "remote AOT" to be disabled in order to not affect existing functionalities. Ready for review and merge.

mpirvu:
What is "remote AOT"? Is this an option of some sort?

harryyu1994:
"remote AOT" is AOT at the server. I'm basically delivering this change but disabling it at the same time. Since AOT is enabled by default, if we enable this change, it will break JITaaS.

PR #4393, opened 2019-01-23
Don't inline Unsafe.put from cold block

liqunl:
Trivial inliner doesn't look at block info, it inlines calls from cold
block. Unsafe call from a cold block can be a product of Unsafe
inlining, thus should not be inlined.

Also add debug counter for illegal write notification.


andrewcraik:

DanHeidinga:
The two windows builds failed due to #4397

DanHeidinga:
@andrewcraik is this ready to be merged?

DanHeidinga:
This will need to be double delivered - both to the `master` branch and the `v0.12.0-release` branch

andrewcraik:
@liqunl this is good to merge can you open a PR against the 0.12 release branch please?

andrewcraik:
dependent change is ready and failures are fails due to #4397 so merging

PR #4366, opened 2019-01-21
DDR: Fix printing of class static field values

keithc-ca:
* access a minimum of 32-bits (as the VM does) to handle byte ordering properly
* `!j9statics` output now suggests use of `!j9romstaticfieldshape` instead of non-existent commands `!j9romstaticdoublefieldshape`, `!j9romstaticsinglefieldshape` or `!j9romstaticstringfieldshape`
* remove unnecessary casts
* fix test for NULL pointer
* correct C/C++ source references

PR #4403, opened 2019-01-23
Remove unused imports

keithc-ca:
Uses of those classes were removed by #4366.


pshipton:
jenkins compile zlinux

PR #4404, opened 2019-01-23
Donot process Unsafe call from cold block

liqunl:
Unsafe call from cold block could be a result of optimization, thus
should be not processed.


liqunl:
@andrewcraik Double deliver for https://github.com/eclipse/openj9/pull/4389

andrewcraik:
4389 is already merged and the failure in CI is infra related - change same on inspection so merging.

PR #4405, opened 2019-01-23
Don't inline Unsafe.put from cold block

liqunl:
Trivial inliner doesn't look at block info, it inlines calls from cold
block. Unsafe call from a cold block can be a product of Unsafe
inlining, thus should not be inlined.

Also add debug counter for illegal write notification.


liqunl:
https://github.com/eclipse/openj9/pull/4393

liqunl:
@andrewcraik

andrewcraik:
release delivery of 4393 - merging now that the mainline is merged

PR #4406, opened 2019-01-23
Function implementation for the TR_J9SharedCacheServerVM class

satbir-ca:

### DIFF
--- a/runtime/compiler/env/VMJ9Server.cpp
+++ b/runtime/compiler/env/VMJ9Server.cpp
@@ -31,6 +31,7 @@
 #include "runtime/CodeCacheExceptions.hpp"
 #include "control/JITaaSCompilationThread.hpp"
 #include "env/PersistentCHTable.hpp"
+#include "exceptions/AOTFailure.hpp"
 
 bool
 TR_J9ServerVM::isClassLibraryMethod(TR_OpaqueMethodBlock *method, bool vettedForAOT)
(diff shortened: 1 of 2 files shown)


satbir-ca:

runtime/compiler/env/VMJ9Server.cpp
```diff
@@ -1417,3 +1417,113 @@ TR_J9ServerVM::dereferenceStaticFinalAddress(void *staticAddress, TR::DataType a
...
+bool
+TR_J9SharedCacheServerVM::stackWalkerMaySkipFrames(TR_OpaqueMethodBlock *method, TR_OpaqueClassBlock *methodClass)
+   {
+   bool skipFrames = TR_J9ServerVM::stackWalkerMaySkipFrames(method, methodClass);
+   TR::Compilation *comp = TR::comp();
+   if (comp && comp->getOption(TR_UseSymbolValidationManager))
+      {
+      if (!comp->getSymbolValidationManager()->addStackWalkerMaySkipFramesRecord(method, methodClass, skipFrames))
+         {
+         TR_ASSERT(false, "Failed to validate addStackWalkerMaySkipFramesRecord\n");
```
mpirvu: Why ASSERT instead of failing the compilation like the original?

runtime/compiler/env/VMJ9Server.cpp
```diff
@@ -1417,3 +1417,113 @@ TR_J9ServerVM::dereferenceStaticFinalAddress(void *staticAddress, TR::DataType a
...
+         }
+      }
+   return skipFrames;
+   }
+
+bool
+TR_J9SharedCacheServerVM::isMethodEnterTracingEnabled(TR_OpaqueMethodBlock *method)
+   {
+   // We want to return the same answer as TR_J9VMBase unless we want to force it to allow tracing
+   return TR_J9VMBase::isMethodEnterTracingEnabled(method) || TR::Options::getAOTCmdLineOptions()->getOption(TR_EnableAOTMethodEnter);
```
mpirvu: This should ask `TR_J9ServerVM`

mpirvu: The test on cmdLineOptions is not going to work here because those are common for all clients. We may need to fetch the comp object and find the options for this client

runtime/compiler/env/VMJ9Server.cpp
```diff
@@ -1417,3 +1417,113 @@ TR_J9ServerVM::dereferenceStaticFinalAddress(void *staticAddress, TR::DataType a
...
+   }
+
+bool
+TR_J9SharedCacheServerVM::isMethodExitTracingEnabled(TR_OpaqueMethodBlock *method)
+   {
+   return TR_J9VMBase::isMethodExitTracingEnabled(method) || TR::Options::getAOTCmdLineOptions()->getOption(TR_EnableAOTMethodExit);
```
mpirvu: Same comments as above

runtime/compiler/env/VMJ9Server.cpp
```diff
@@ -1417,3 +1417,113 @@ TR_J9ServerVM::dereferenceStaticFinalAddress(void *staticAddress, TR::DataType a
...
+TR_J9SharedCacheServerVM::traceableMethodsCanBeInlined()
+   {
+   return true;
+   }
+
+bool
+TR_J9SharedCacheServerVM::canMethodEnterEventBeHooked()
+   {
+   // We want to return the same answer as TR_J9ServerVM unless we want to force it to allow tracing
+   return TR_J9ServerVM::canMethodEnterEventBeHooked() || TR::Options::getAOTCmdLineOptions()->getOption(TR_EnableAOTMethodEnter);
```
mpirvu: You should not test cmdLineOptions because they are global and thus common for all clients

runtime/compiler/env/VMJ9Server.cpp
```diff
@@ -1417,3 +1417,113 @@ TR_J9ServerVM::dereferenceStaticFinalAddress(void *staticAddress, TR::DataType a
...
+   }
+
+bool
+TR_J9SharedCacheServerVM::canMethodExitEventBeHooked()
+   {
+   // We want to return the same answer as TR_J9ServerVM unless we want to force it to allow tracing
+   return TR_J9ServerVM::canMethodExitEventBeHooked() || TR::Options::getAOTCmdLineOptions()->getOption(TR_EnableAOTMethodExit);
```
mpirvu: You should not test cmdLineOptions because they are global and thus common for all clients

runtime/compiler/env/VMJ9Server.cpp
```diff
@@ -1417,3 +1417,113 @@ TR_J9ServerVM::dereferenceStaticFinalAddress(void *staticAddress, TR::DataType a
...
+         TR_ASSERT(false, "Class 0x%p should already be validated\n", classPointer);
+         }
+      }
+   else
+      {
+      validated = ((TR_ResolvedRelocatableJ9Method*) comp->getCurrentMethod())->validateArbitraryClass(comp, (J9Class*)classPointer);
+      }
+
+   if (validated)
+      return this->TR_J9ServerVM::getInstanceFieldOffset (classPointer, fieldName, fieldLen, sig, sigLen, options);
```
mpirvu: Is `this` required here?

runtime/compiler/env/VMJ9Server.cpp
```diff
@@ -1417,3 +1417,113 @@ TR_J9ServerVM::dereferenceStaticFinalAddress(void *staticAddress, TR::DataType a
...
+         validated = true;
+         }
+      else
+         {
+         TR_ASSERT(false, "Class 0x%p should already be validated\n", classPointer);
+         }
+      }
+   else
+      {
+      validated = ((TR_ResolvedRelocatableJ9Method*) comp->getCurrentMethod())->validateArbitraryClass(comp, (J9Class*)classPointer);
```
mpirvu: The cast should be made to a TR_ResolvedRelocatableJITaaSJ9ServerMethod which hasn't been created yet

runtime/compiler/env/VMJ9Server.cpp
```diff
@@ -1417,3 +1417,113 @@ TR_J9ServerVM::dereferenceStaticFinalAddress(void *staticAddress, TR::DataType a
...
    }
 
+bool
+TR_J9SharedCacheServerVM::isClassLibraryMethod(TR_OpaqueMethodBlock *method, bool vettedForAOT)
+   {
+   TR_ASSERT(vettedForAOT, "The TR_J9SharedCacheServerVM version of this method is expected to be called only from isClassLibraryMethod.\n"
+                                          "Please consider whether you call is vetted for AOT!");
+
+   if (getSupportsRecognizedMethods())
+      return this->TR_J9ServerVM::isClassLibraryMethod(method, vettedForAOT);
```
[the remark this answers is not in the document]
satbir-ca: being used in the implementation in the J9VM class.

runtime/compiler/env/VMJ9Server.cpp
satbir-ca: we are failing on the server side will the failure propagate to the client?

runtime/compiler/env/VMJ9Server.cpp
satbir-ca: Not able to find the TR_J9ServerVM::isMethodEnterTracingEnabled.
On client side also using the TR_J9VMBase::isMethodEnterTracingEnabled version not the TR_J9VM::isMethodEnterTracingEnabled.

runtime/compiler/env/VMJ9Server.cpp
satbir-ca: Do we need to do any JITaaS specific in this drived class or it will an exact replica of the TR_ResolvedRelocatableJ9Method.

runtime/compiler/env/VMJ9Server.cpp
```diff
@@ -1417,3 +1417,113 @@ TR_J9ServerVM::dereferenceStaticFinalAddress(void *staticAddress, TR::DataType a
...
+      }
+   else
+      {
+      validated = ((TR_ResolvedRelocatableJ9Method*) comp->getCurrentMethod())->validateArbitraryClass(comp, (J9Class*)classPointer);
+      }
+
+   if (validated)
+      return this->TR_J9ServerVM::getInstanceFieldOffset (classPointer, fieldName, fieldLen, sig, sigLen, options);
+
+   return ~0;
```
dsouzai: Heh, if this is not a typo, maybe it should be fixed to just be `return -1`.

mpirvu: The function returns an unsigned data type so ~0 makes more sense here.

harryyu1994:
@satbir-ca Hi Satbir, could you please add the APIs you touched in the box file? Just making sure the team don't do redundant work.

satbir-ca:
updated. 
This depends on the merge of the #4509

runtime/compiler/env/VMJ9Server.cpp
```diff
@@ -1429,3 +1430,118 @@ TR_J9ServerVM::methodTrampolineLookup(TR::Compilation *comp, TR::SymbolReference
...
+bool
+TR_J9SharedCacheServerVM::stackWalkerMaySkipFrames(TR_OpaqueMethodBlock *method, TR_OpaqueClassBlock *methodClass)
+   {
+   bool skipFrames = TR_J9ServerVM::stackWalkerMaySkipFrames(method, methodClass);
+   TR::Compilation *comp = TR::comp();
+   if (comp && comp->getOption(TR_UseSymbolValidationManager))
+      {
+      if (!comp->getSymbolValidationManager()->addStackWalkerMaySkipFramesRecord(method, methodClass, skipFrames))
+         {
+         TR_ASSERT(false, "Failed to validate addStackWalkerMaySkipFramesRecord\n");
```
mpirvu: The new code uses a SVM_ASSERT here

PR #4416, opened 2019-01-23
Change J9VMGCRememberedSet to OMR's MM_GCRememberedSet

jason-hall:
After adding this typedef to use OMR's new MM_GCRememberedSet structure, usage of J9VMGCRememberedSet in both OMR and OpenJ9 can be changed to MM_GCRememberedSet. This structure is used in OMR but was previously missing.


dmitripivkine:
Add WIP until OMR part is promoted

jason-hall:
Looks like I mixed something up and had the wrong changes here originally

dmitripivkine:

Issue #3789, opened 2018-11-21
gpTest suite failures on OSX with JDK8


babsingh:
With JDK11 on OSX, `gpTest` suite has failures because of `UnsatisfiedLinkError`: https://github.com/eclipse/openj9/issues/3686.

With JDK8 on OSX, `gpTest` failures are different. The following tests are failing in the `gpTest` suite:
```
23:22:56 ---SUMMARY OF FAILED TESTS---
23:22:56 abort  // Process timed out. Looks like a JVM hang.
23:22:56 hardwareFloat // [ERR] Survived hardware-triggered SIGFPE!
23:22:56 abort thread // Process timed out. Looks like a JVM hang.
23:22:56 hardwareFloat thread  // [ERR] Survived hardware-triggered SIGFPE!
```

**OSX JDK8 Test output:**
```
23:11:54 Testing: abort  
23:11:54 Test start time: 2018/11/20 23:11:54 Eastern Standard Time
23:11:54 Running command: "/Users/jenkins/workspace/Test-sanity.functional-JDK8-osx_x86-64_cmprssptrs/openjdkbinary/j2sdk-image/jre/bin/java" -Xcompressedrefs  -Xdump:system:none -cp "/Users/jenkins/workspace/Test-sanity.functional-JDK8-osx_x86-64_cmprssptrs/openjdk-tests/TestConfig/scripts/testKitGen/../../../../jvmtest/functional/cmdline_options_testresources/cmdlinetestresources.jar" VMBench.GPTests.GPTest   abort
23:11:54 Time spent starting: 3 milliseconds
23:17:37 ***[TEST INFO 2018/11/20 23:17:23] ProcessKiller detected a timeout after 300000 milliseconds!***
23:17:37 INFO: The current OS is 'Mac OS X'. 'Debug on timeout' is currently only supported on Linux.
23:17:37 ***[TEST INFO 2018/11/20 23:17:23] executing kill -ABRT 82233***
23:17:37 ***[TEST INFO 2018/11/20 23:17:23] kill -ABRT signal sent***
23:17:37 Time spent executing: 329288 milliseconds
23:17:37 Test result: FAILED
23:17:37 ***[TEST INFO 2018/11/20 23:17:23] ABRT completed***
23:17:37 Output from test:
23:17:37  [ERR] JVMDUMP039I Processing dump event "abort", detail "" at 2018/11/20 23:11:54 - please wait.
[...]

PR #4420, opened 2019-01-24
Fix the hang during abort and re-enable the gpTest suite on OSX

babsingh:
Mirror of https://github.com/eclipse/openj9/pull/4402 for the **0.12.0** release branch.

**1. Disable the call to `protectedWriteThreadsWithNativeStacks`**

While writing the thread section for the javacore, the call to
`protectedWriteThreadsWithNativeStacks` has been disabled since the
`omrintrospect` library causes a hang and fails to retrieve the backtraces
and symbols for native threads on OSX.

**2. Re-enable the gpTest suite on OSX**

All the issues, reported in #3789, have been resolved. Thus, the gpTest
suite is re-enabled on OSX.


babsingh:
fyi - @pshipton @DanHeidinga

Issue #4337, opened 2019-01-17
NULL loop parent causes segfault in SPMDParallelizer


IBMJimmyk:
The GPU transformation might make the block we are trying to vectorize unreachable due to creating and using a new GPU path. This occurs if the Xjit option enableGPU={enforce} is set since it forces the GPU path to always be taken. If vectorization is attempted anyways, it will segfault due to the loop block now having a parent of NULL.

SPMDParallelizer should be updated to check for this case and skip the vectorization attempt.

PR #4428, opened 2019-01-24
Added NULL parent check to SPMDParallelizer

IBMJimmyk:
A problem arises if the enableGPU={enforce} Xjit option is used and
vectorization is attempted on the non-GPU path that is parallel to a newly
created JIT GPU path. Due to the enforce option, the non-GPU path is never
taken and the block is unreachable. However, the SIMD part of the
optimization still knows about the non-GPU path blocks.

This change adds a check before vectorization is done to check if the
loop to be vectorized is unreachable. It does this by checking if the
loop's parent is NULL. If it is NULL, vectorization is not performed.

Issue: #4337


andrewcraik:
CI failure expected - targeting a release branch

Issue #4397, opened 2019-01-23
Failures running Javac for DDR_Test


andrewcraik:
PR builds are failing regularly trying to run the DDR_Tests. Failures report a compile step failed related to javac with cryptic messages similar to:
```10:48:33 /home/u0020236/workspace/PullRequest-Sanity-JDK11-aix_ppc-64_cmprssptrs-OpenJ9/openj9/test/functional/DDR_Test/build.xml:59: Error running /home/jenkins/jenkins-agent/workspace/PullRequest-Sanity-JDK11-aix_ppc-64_cmprssptrs-OpenJ9/build/aix-ppc64-normal-server-release/images/jdk/bin/javac compiler
10:48:33 
10:48:33 Total time: 1 second
10:48:33 makefile:82: recipe for target 'compile' failed
10:48:33 make: *** [compile] Error 1
[Pipeline] }
```

see https://ci.eclipse.org/openj9/job/PullRequest-Sanity-JDK11-aix_ppc-64_cmprssptrs-OpenJ9/85/console

andrewcraik:
@DanHeidinga @pshipton FYI. @keithc-ca could this be related to any of the recent DDR changes? Not trying to place blame just trying to figure out if this is a problem with javac or a problem with DDR - any thoughts?

andrewcraik:
FYI @0xdaryl and @liqunl due to interest on the OpenJ9 triage slack channel

DanHeidinga:
I think I've seen it on non-DDR builds as well - looking for a link now.

keithc-ca:
Yes, I think the test code is being compiled with the brand new JDK. I agree it's serious if javac in that JDK doesn't work.

pshipton:
I think we found the culprit
https://github.com/eclipse/openj9/pull/4017

DanHeidinga:
@pshipton Is this fixed now that #4017 has been merged?  If so we should close it out.

AdamBrousseau:
No, 4017 is what caused this issue. I am still determining why though. I would like to figure it out so we can get in a fix rather than simply reverting the change.

DanHeidinga:
@AdamBrousseau What's the downside of reverting now and re-introducing the change later?  Currently PR builds are broken which affects all PRs.

PR #4429, opened 2019-01-24
Preserve null check on Unsafe

liqunl:
Preserve the null check before Unsafe inlining.

Also move `TR_J9ByteCodeIlGenerator::separateNullCheck` to
`TransformUtil` so that it can be used in both ILGen and optimizer.


liqunl:

andrewcraik:

andrewcraik:
Failures due to #4397 - good to go.

andrewcraik:
@liqunl can you create the PR to target 0.12 since this is now good to go and we'll get it merged? Thanks!

Issue #2977, opened 2018-09-21
Options starting with -XX: should not be ignored


pshipton:
OpenJ9 ignores any options starting with -XX: that it doesn't understand. While this may be useful if trying to use an OpenJDK command line containing options that don't apply to OpenJ9, it makes it hard to tell when an option starting with -XX: is valid for OpenJ9 or not. OpenJDK does not have this behavior.

https://github.com/eclipse/openj9/blob/a65785ba6a75ca53bdb15921808bb339479440c4/runtime/vm/jvminit.c#L184

I suggest changing the OpenJ9 default behavior so unrecognized options starting with -XX: cause an error. There is a option -XX:+IgnoreUnrecognizedVMOptions that can be used to ignore any unrecognized (top level) option, which includes those starting with -XX:.

Another option could be added if necessary to just ignore unrecognized options starting with -XX:

DanHeidinga:
This has been requested previously: https://github.com/eclipse/openj9/issues/526

> Another option could be added if necessary to just ignore unrecognized options starting with -XX:

OpenJ9's current behaviour makes migration easier but is less explicit for users that want to be notified about the mismatch.

I think the new option is the right path forward as it allows existing users to maintain the behaviour they've come to know and love(?) without requiring them to modify their existing commandlines.

pdbain-ibm:
I was palnning to do some argument hacking.  Would you like me to work on this?

luchsh:
From a user's perspective, the current behavior is too confusing.
HotSpot has got an option '-XX:+IgnoreUnrecognizedVMOptions', can OpenJ9 work in the same way?
or maybe you can create a whitelist of valid '-XX' options for compatibility, and report errors when cannot find a match in the whitelist.

AidanHa:
Hello, I will be working on this issue.

For now, I will implement the new option, where -XX:-IgnoreUnrecognizedVMOptions reports all unrecognized options, and -XX:+IgnoreUnrecognizedVMOptions ignores them.

I'm still a bit confused on how the current options work, but I will figure it out...

pshipton:
Note https://github.com/eclipse/openj9/issues/2977#issuecomment-423578791, we want to introduce a new option here, rather than changing the behavior of IgnoreUnrecognizedVMOptions.

> I think the new option is the right path forward as it allows existing users to maintain the behaviour they've come to know and love(?) without requiring them to modify their existing commandlines.

AidanHa:
[...]

PR #4435, opened 2019-01-24
Enhancement to report any unrecognized options that begin with "-XX:" (Issue2977)

AidanHa:
Add an enhancement option that allows users to enter the option '-XX:-IgnoreUnrecognizedXXColonOptions' to ignore all unrecognized options that start with '-XX:'

Steps:
Removed ' { VMOPT_XX, STARTSWITH_MATCH }'  from the Ignore table in runtime/vm/jvminit.c (Line 184)
Add a new flag in j9options that is set when the argument is supposed to be ignored (ARG_IGNORED)
Add necessary macros to support the new flag
Modify CheckArgsConsumed and findArgInVMArgs to be able to set and read the new flag

Doc Issue https://github.com/eclipse/openj9-docs/issues/212

Fixes: #2977


runtime/util/vmargs.c
```diff
@@ -200,7 +199,11 @@ findArgInVMArgs(J9PortLibrary *portLibrary, J9VMInitArgs* j9vm_args, UDATA match
 							j9vm_args->j9Options[optionCntr].flags = NOT_CONSUMABLE_ARG /* override any flags that already exist */
 									| (j9vm_args->j9Options[optionCntr].flags & ARG_MEMORY_ALLOCATION);
 						} else {
-							j9vm_args->j9Options[optionCntr].flags |= ARG_CONSUMED;
+							if (doConsumeIgnoredArgs) {
+								j9vm_args->j9Options[optionCntr].flags |= ARG_IGNORED;
```
tajila: ignored args still have to be consumed so no one else can process them

runtime/vm/jvminit.c
```diff
@@ -2331,27 +2330,36 @@ static UDATA checkArgsConsumed(J9PortLibrary* portLibrary, J9VMInitArgs* j9vm_ar
...
 	jboolean ignoreUnrecognized = j9vm_args->actualVMArgs->ignoreUnrecognized;
+	jboolean ignoreUnrecongizedXXOption = JNI_TRUE;
 	jboolean ignoreUnrecongizedTopLevelOption = JNI_FALSE;
 	IDATA xxIgnoreUnrecognizedVMOptionsEnableIndex = -1;
 	IDATA xxIgnoreUnrecognizedVMOptionsDisableIndex = -1;
+	IDATA xxIgnoreUnrecognizedXXOptionsEnableIndex = -1;
+	IDATA xxIgnoreUnrecognizedXXOptionsDisableIndex = -1;
 
-	if (findArgInVMArgs( PORTLIB, j9vm_args, EXACT_MATCH, VMOPT_XXVM_IGNOREUNRECOGNIZED, NULL, TRUE) >= 0) {
+	if (findArgInVMArgs( PORTLIB, j9vm_args, EXACT_MATCH, VMOPT_XXVM_IGNOREUNRECOGNIZED, NULL, TRUE, FALSE) >= 0) { //-XXvm:ignoreUnrecognized
```
tajila: comments need to be with `/* */` instead of `//` for compiler compatability

runtime/vm/jvminit.c
```diff
@@ -2363,11 +2371,15 @@ static UDATA checkArgsConsumed(J9PortLibrary* portLibrary, J9VMInitArgs* j9vm_ar
 			} else if (HAS_MAPPING( j9vm_args, i) && (MAPPING_FLAGS( j9vm_args, i ) & INVALID_OPTION)) {
 				j9nls_printf(PORTLIB, J9NLS_ERROR, J9NLS_VM_INVALID_CMD_LINE_OPT, optString);
 			} else {
-				/* If ignoreUnrecongizedTopLevelOption is set to JNI_TRUE, we should ignore any unrecognized top-level option */
+				/*If ignoreUnrecognizedXXOption is set to JNI_TRUE, we should ignore any options that start with -XX */
+				if (ignoreUnrecongizedXXOption) {
```
tajila: do you need to check that the option is actually `-XX:` before you continue ?

tajila:
Can you please update the commit message and the PR description so it matches the most recent changes

tajila:
Also, I think you can greatly simplify this PR by introducing a new function called `findArgInVMArgsImpl`. This function will take in the extra parameter for `ignoredArgs`.

The existing `findArgInVMArgs` will have the same existing method params and will call the new function and pass in false for `ignoredArgs`. In the one place where you need the new behaviour you can call the new function. In all other places the code stays the same as it is.

runtime/util/vmargs.c
```diff
@@ -221,6 +221,152 @@ findArgInVMArgs(J9PortLibrary *portLibrary, J9VMInitArgs* j9vm_args, UDATA match
 	return returnVal;
 }
 
+IDATA
+findArgInVMArgsImpl(J9PortLibrary *portLibrary, J9VMInitArgs* j9vm_args, UDATA match, const char* optionName, const char* optionValue, UDATA doConsumeArgs, UDATA doConsumeIgnoredArgs) {
```
tajila: Changes look good, however instead of having two versions of this, what I meant was `findArgInVMArgs` would simply call `findArgInVMArgsImpl` and pass in false for `doConsumeIgnoredArgs`

runtime/oti/j9nonbuilder.h
```diff
@@ -4521,6 +4521,7 @@ typedef struct J9InternalVMFunctions {
 	IDATA  ( *objectMonitorEnterBlocking)(struct J9VMThread *currentThread) ;
 	void  ( *fillJITVTableSlot)(struct J9VMThread *vmStruct, UDATA *currentSlot, struct J9Method *ramMethod) ;
 	IDATA  ( *findArgInVMArgs)(J9PortLibrary *portLibrary, struct J9VMInitArgs* j9vm_args, UDATA match, const char* optionName, const char* optionValue, UDATA doConsumeArgs) ;
+	IDATA  ( *findArgInVMArgsImpl)(J9PortLibrary *portLibrary, struct J9VMInitArgs* j9vm_args, UDATA match, const char* optionName, const char* optionValue, UDATA doConsumeArgs, UDATA doConsumeIgnoredArgs) ;
```
tajila: this function isn't called from outside the j9vm lib, so it doesnt need to be in the internal function table

tajila:
Can you show the results of running the following `-XInvalid`, `-XX:Invalid`,  `-XInvalid -XX:Invalid` and `-XInvalid -XX:Invalid` with:

1) Default options
2) -XX:+IgnoreUnrecognizedVMOptions
3) -XX:-IgnoreUnrecognizedVMOptions
4) -XX:+IgnoreUnrecognizedXXVMOptions
5) -XX:-IgnoreUnrecognizedXXVMOptions

AidanHa:
Sure, the changes I implemented are giving errors while building the jdk, so I will look into that first.

AidanHa:
image
image

tajila:
Thanks running that @AidanHa. 

I made an error in my last comment. The last one was supposed to be `-XX:Invalid -XInvalid` not `-XInvalid -XX:Invalid`. Can you please show those results as well

AidanHa:
@tajila 

image

runtime/util/vmargs.c
```diff
@@ -80,6 +80,13 @@ static char * getStartOfOptionValue(J9VMInitArgs* j9vm_args, IDATA element, cons
 
 IDATA
 findArgInVMArgs(J9PortLibrary *portLibrary, J9VMInitArgs* j9vm_args, UDATA match, const char* optionName, const char* optionValue, UDATA doConsumeArgs) {
+	PORT_ACCESS_FROM_PORT(portLibrary);
```
tajila: this line is not needed, can simply pass port from the parameter

runtime/oti/j9nonbuilder.h
tajila: still needs to be addressed

runtime/vm/intfunc.c
```diff
@@ -173,6 +173,7 @@ J9InternalVMFunctions J9InternalFunctions = {
 	objectMonitorEnterBlocking,
 	fillJITVTableSlot,
 	findArgInVMArgs,
+	findArgInVMArgsImpl,
```
tajila: this is not needed either, since `findArgInVMArgsImpl` is only being called from the j9vm lib

runtime/vm/jvminit.c
```diff
@@ -2363,6 +2373,10 @@ static UDATA checkArgsConsumed(J9PortLibrary* portLibrary, J9VMInitArgs* j9vm_ar
 			} else if (HAS_MAPPING( j9vm_args, i) && (MAPPING_FLAGS( j9vm_args, i ) & INVALID_OPTION)) {
 				j9nls_printf(PORTLIB, J9NLS_ERROR, J9NLS_VM_INVALID_CMD_LINE_OPT, optString);
 			} else {
+				/*If ignoreUnrecognizedXXOption is set to JNI_TRUE, we should ignore any options that start with -XX */
+				if (ignoreUnrecongizedXXOption && !strncmp(optString, "-XX:", 4)) {
```
tajila: can you make the strncmp check more explicit, ie. `(0 == strncmp)`

runtime/vm/jvminit.c
```diff
@@ -2363,6 +2373,10 @@ static UDATA checkArgsConsumed(J9PortLibrary* portLibrary, J9VMInitArgs* j9vm_ar
 			} else if (HAS_MAPPING( j9vm_args, i) && (MAPPING_FLAGS( j9vm_args, i ) & INVALID_OPTION)) {
 				j9nls_printf(PORTLIB, J9NLS_ERROR, J9NLS_VM_INVALID_CMD_LINE_OPT, optString);
 			} else {
+				/*If ignoreUnrecognizedXXOption is set to JNI_TRUE, we should ignore any options that start with -XX */
```
tajila: Minor, needs space after `/*`

tajila:
Some minor comments, rest LGTM

pshipton:
We'll need a doc item created for this.

AidanHa:
Doc Issue at https://github.com/eclipse/openj9-docs/issues/212

tajila:

pshipton:
@tajila please resolve your comments which have been addressed.

The test suite needs to be updated with additional tests for the new option.

Please explain how this works. It is not apparent to me why ARG_IGNORED is needed. It seems setting ARG_CONSUMED should be sufficient.

runtime/vm/jvminit.c
```diff
@@ -2345,8 +2344,11 @@ static UDATA checkArgsConsumed(J9PortLibrary* portLibrary, J9VMInitArgs* j9vm_ar
 	PORT_ACCESS_FROM_PORT(portLibrary);
 	jboolean ignoreUnrecognized = j9vm_args->actualVMArgs->ignoreUnrecognized;
 	jboolean ignoreUnrecongizedTopLevelOption = JNI_FALSE;
+	jboolean ignoreUnrecongizedXXOption = JNI_TRUE;
```
pshipton: This should be named `ignoreUnrecongizedXXOptions` with an `s`.

tajila:
@pshipton 

> Please explain how this works. It is not apparent to me why ARG_IGNORED is needed. It seems setting ARG_CONSUMED should be sufficient.

We need a way to tell if the arg was consumed because it was valid, or if it was consumed because it was in the ignore list. (ie `findArgInVMArgs` in `registerIgnoredOptions` vs all other usages of `findArgInVMArgs`)

Also, I am unable to resolve the old comments, the option to do so does not appear for me. @AidanHa are you able to resolve the old comments?

AidanHa:
I have resolved the  old comments on your behalf if that's okay.

tajila:
@AidanHa thanks

tajila:
> The test suite needs to be updated with additional tests for the new option.

I think we will need tests for the 4 cases I listed above. Also, we need to test `-XX:+IgnoreUnrecognizedXXOptions` with valid -XX: options.

Also, @AidanHa, did you run a sanity build with the latest changes?

test/functional/cmdLineTests/ignoreUnrecognizedVMOptions/ignoreunrecognizedxxoptions.xml
```diff
@@ -0,0 +1,83 @@
...
+  General Public License, version 2 with the GNU Classpath
+  Exception [1] and GNU General Public License, version 2 with the
+  OpenJDK Assembly Exception [2].
+
+  [1] https://www.gnu.org/software/classpath/license.html
+  [2] http://openjdk.java.net/legal/assembly-exception.html
+
+  SPDX-License-Identifier: EPL-2.0 OR Apache-2.0 OR GPL-2.0 WITH Classpath-exception-2.0 OR LicenseRef-GPL-2.0 WITH Assembly-exception
+-->
+
```
tajila: Will also need to test `-XX:+IgnoreUnrecognizedOptions`

test/functional/cmdLineTests/ignoreUnrecognizedVMOptions/ignoreunrecognizedxxoptions.xml
```diff
@@ -0,0 +1,83 @@
...
+<!DOCTYPE suite SYSTEM "cmdlinetester.dtd">
+
+<suite id="IgnoreUnrecognizedXXOptions Command-Line Option Tests" timeout="2400">
+
+ <variable name="IGNOREUNRECOGNIZEDXXOPTIONS" value="-XX:+IgnoreUnrecognizedXXOptions" />
+ <variable name="NOIGNOREUNRECOGNIZEDXXOPTIONS" value="-XX:-IgnoreUnrecognizedXXOptions" />
+ 
+ <variable name="BADOPTION1" value="-XX:INVALID" />
```
tajila: will also need to test good options

test/functional/cmdLineTests/ignoreUnrecognizedVMOptions/ignoreunrecognizedxxoptions.xml
AidanHa: @tajila did you mean `-XXvm:ignoreUnrecognized`. I couldn't find the option you gave me

tajila: I meant `-XX:+IgnoreUnrecognizedVMOptions` and `-XX:-IgnoreUnrecognizedVMOptions`

AidanHa: There are tests that exist for that seperately in `/test/functional/cmdLineTests/ignoreUnrecognizedVMOptions/ignoreunrecognizedvmoptions.xml` If you're asking to create tests to include both options, then which file should I add them to?

I'll assume that I add them in the suite that test for xx options

test/functional/cmdLineTests/ignoreUnrecognizedVMOptions/playlist.xml
```diff
@@ -31,7 +31,22 @@
 		<command>$(JAVA_COMMAND) $(JVM_OPTIONS) -DTESTDIR=$(Q)$(TEST_RESROOT)$(Q) -DRESJAR=$(CMDLINETESTER_RESJAR) -DEXE=$(SQ)$(JAVA_COMMAND)$(SQ) -jar $(CMDLINETESTER_JAR) -config $(Q)$(TEST_RESROOT)$(D)ignoreunrecognizedvmoptions.xml$(Q) -verbose -explainExcludes -nonZeroExitWhenError; \
 	$(TEST_STATUS)</command>
 		<levels>
-			<level>sanity</level>
+			<level>extended</level>
```
pshipton: FYI, I asked for this change, to move this testing to extended. I don't think it needs to be in sanity, as the implementation would be rarely changed and unlikely to break normally.

test/functional/cmdLineTests/ignoreUnrecognizedVMOptions/ignoreunrecognizedxxoptions.xml
tajila: Ah okay, I wasn't aware that those tests existed, you can ignore this comment

runtime/vm/jvminit.c
```diff
@@ -2360,8 +2362,16 @@ static UDATA checkArgsConsumed(J9PortLibrary* portLibrary, J9VMInitArgs* j9vm_ar
...
+	if (xxIgnoreUnrecognizedXXOptionsDisableIndex >= 0) {
+		xxIgnoreUnrecognizedXXOptionsEnableIndex = findArgInVMArgs( PORTLIB, j9vm_args, EXACT_MATCH, VMOPT_XXIGNOREUNRECOGNIZEDXXOPTIONSENABLE, NULL, TRUE);
+		if (xxIgnoreUnrecognizedXXOptionsDisableIndex > xxIgnoreUnrecognizedXXOptionsEnableIndex) {
+			ignoreUnrecongizedXXOption = JNI_FALSE;
+		}
+	}
+
 	for (i=0; i<j9vm_args->nOptions; i++) {
-		if (IS_CONSUMABLE( j9vm_args, i ) && !IS_CONSUMED( j9vm_args, i )) {
+		if (IS_IGNORED(j9vm_args,i) || (IS_CONSUMABLE( j9vm_args, i ) && !IS_CONSUMED( j9vm_args, i ))) {
```
DanHeidinga: Shouldn't this be `if (!ignored || (consumable && !consumed))`?  ie: only process args that haven't been ignored or already consumed?

runtime/vm/jvminit.c
```diff
@@ -2376,6 +2386,10 @@ static UDATA checkArgsConsumed(J9PortLibrary* portLibrary, J9VMInitArgs* j9vm_ar
 			} else if (HAS_MAPPING( j9vm_args, i) && (MAPPING_FLAGS( j9vm_args, i ) & INVALID_OPTION)) {
 				j9nls_printf(PORTLIB, J9NLS_ERROR, J9NLS_VM_INVALID_CMD_LINE_OPT, optString);
 			} else {
+				/* If ignoreUnrecognizedXXOption is set to JNI_TRUE, we should ignore any options that start with -XX */
+				if (ignoreUnrecongizedXXOption && (0 == strncmp(optString, "-XX:", 4))) {
```
DanHeidinga: comment and code don't match - comment says `-XX`, code says `-XX:`.  Which is the intended behaviour?

runtime/oti/vm_api.h
```diff
@@ -1422,6 +1422,19 @@ exitJavaVM(J9VMThread * vmThread, IDATA rc);
...
 
+/**
+* @brief
+* @param j9vm_args
+* @param match
+* @param optionName
+* @param optionValue
+* @param doConsumeArgs
+* @param doConsumeIgnoredArgs
+* @return IDATA
```
DanHeidinga: Can you fill in the values for these doc annotations?  I know existing code hasn't done a great job of actually provide descriptions of the function / arguments / return value but let's try to do better =)

test/functional/cmdLineTests/ignoreUnrecognizedVMOptions/ignoreunrecognizedxxoptions.xml
```diff
@@ -0,0 +1,144 @@
...
+
+<!DOCTYPE suite SYSTEM "cmdlinetester.dtd">
+
+<suite id="IgnoreUnrecognizedXXOptions Command-Line Option Tests" timeout="2400">
+
+ <variable name="IGNOREUNRECOGNIZEDOPTIONS" value="-XXvm:ignoreUnrecognized" />
+ <variable name="IGNOREUNRECOGNIZEDXXOPTIONS" value="-XX:+IgnoreUnrecognizedXXOptions" />
+ <variable name="NOIGNOREUNRECOGNIZEDXXOPTIONS" value="-XX:-IgnoreUnrecognizedXXOptions" />
+ 
+ <variable name="BADOPTION" value="-XX:INVALID" />
```
DanHeidinga: Should we test more than 1 bad option pattern?  What about `-XXFoo`?  or `-XX:+Foo` or -XX:-Foo`

pshipton:
@tajila mentioned to me that some `-XX:` options are parsed but not consumed. It seems that needs to be fixed first.

runtime/vm/jvminit.c
AidanHa: ah, that is correct, I will address this is my next push

runtime/vm/jvminit.c
AidanHa: -XX:, I'll change the comment

test/functional/cmdLineTests/ignoreUnrecognizedVMOptions/ignoreunrecognizedxxoptions.xml
AidanHa: Gotcha, I'll delete the tests that I have wrote for this function

test/functional/cmdLineTests/ignoreUnrecognizedVMOptions/ignoreunrecognizedxxoptions.xml
AidanHa: good point, I will address this

runtime/vm/jvminit.c
DanHeidinga: I expected the answer would go the other way and the option would ignore all -XX options.  Why require the `:`?

AidanHa: If I understand correctly, the original issue/enhancement was to report/ignore any issues that start with `-XX:` and not `-XX`

runtime/vm/jvminit.c
AidanHa: Looking back at this, I think the proper boolean statement is 

`!IS_IGNORED(j9vm_args,i) && IS_CONSUMABLE( j9vm_args, i ) && !IS_CONSUMED( j9vm_args, i )`

runtime/vm/jvminit.c
DanHeidinga: That's an oversight.  Most `-XX` options are of the form `-XX:[+-]Something` so I think the original issue is overly specific.  Let's actually match on `-XX` to ensure the behaviour matches the option name

AidanHa: Will do. I'll change the PR accordingly.

runtime/vm/jvminit.c
AidanHa: I've tested the new boolean statement, and it seems to work as intended. I'll confirm with an internal grinder instance

runtime/vm/jvminit.c
pshipton: This would be a change in behavior. The VM has only ever ignored options that start with `-XX:`

https://github.com/eclipse/openj9/blob/278be6c0ff617069c6fa2038ffac7afbf4cb5989/runtime/vm/jvminit.c#L183

https://github.com/eclipse/openj9/blob/278be6c0ff617069c6fa2038ffac7afbf4cb5989/runtime/oti/jvminit.h#L292

pshipton: A change in the default behavior.

runtime/vm/jvminit.c
pshipton: I don't think this code is necessary. As I understand it, the problem is that some options are parsed without being consumed. Once this problem is fixed, I don't expect the IGNORED flag will be needed. Until the problem is fixed, the IGNORED flag as written isn't going to make it work properly for all cases.

runtime/vm/jvminit.c
DanHeidinga: I'm OK with changing the corner case here so the behaviour is more consistent.

A quick test with the RI shows that `java -XXFoo -XX:+IgnoreUnrecognizedVMOptions -version` starts just fine.  We should have have the same behaviour when ignoring options.

pshipton: And we do. `java -XXFoo -XX:+IgnoreUnrecognizedVMOptions -version` starts with with OpenJ9 as well. That isn't what this PR is about.

runtime/vm/jvminit.c
AidanHa: Okay, I will be removing the IGNORED flag since it is no longer needed.

DanHeidinga:
Summarizing my understanding here:
* VM has historically ignored `-XX:` options by default
* All VM options can be ignored (or not) with `-XX:[+-]IgnoreUnrecognizedVMOptions`
* This PR adds a new `-XX:[+-]IgnoreUnrecognizedXXOptions`option to toggle the historic default behaviour.

Given that, my comments on modifying the default behaviour to also ignore `-XX` options are probably step to far for this PR.

I'm not 100% sold on the option name given the behaviour is related to `-XX:` but don't have a good suggestion for alternative names.  If anyone has suggestions....

pshipton:
Not friendly, but `-XX:[+-]IgnoreUnrecognizedXX:Options`

AidanHa:
I personally do not like the `:` in the name but if no one has any better option, I will change the option to `-XX:[+-]IgnoreUnrecognizedXX:Options`

DanHeidinga:
I'm opposed to `-XX:[+-]IgnoreUnrecognizedXX:Options` as it implies other suboptions and messes with the reading of the command.

Given this option won't be commonly used, let's pick a more explicit name like: `-XX:[+-]IgnoreUnrecognizedXXColonOptions`.  The option is already verbose so making it a little longer won't hurt it.

runtime/vm/jvminit.c
```diff
@@ -2387,6 +2397,10 @@ static UDATA checkArgsConsumed(J9PortLibrary* portLibrary, J9VMInitArgs* j9vm_ar
 			} else if (HAS_MAPPING( j9vm_args, i) && (MAPPING_FLAGS( j9vm_args, i ) & INVALID_OPTION)) {
 				j9nls_printf(PORTLIB, J9NLS_ERROR, J9NLS_VM_INVALID_CMD_LINE_OPT, optString);
 			} else {
+				/* If ignoreUnrecognizedXXOption is set to JNI_TRUE, we should ignore any options that start with -XX */
```
DanHeidinga: This should go back to `-XX:` so as not to change the default behaviour

pshipton:
`-XX:[+-]IgnoreUnrecognizedXXColonOptions` is fine with me

AidanHa:
I ran a grinder instance with the new tests and all seems well. Is this PR good to go?

pshipton:
I'd like to see https://github.com/eclipse/openj9/pull/4435#issuecomment-467289346 addressed first before this is merged.

AidanHa:
> @tajila mentioned to me that some `-XX:` options are parsed but not consumed. It seems that needs to be fixed first.

Hello, this has been fixed. These options are now recognized and consumed. I tested several of these options out locally, but I can run some more grinder tests if necessary.

pshipton:
Where was it fixed?

runtime/vm/jvminit.c
```diff
@@ -2902,59 +2916,59 @@ processVMArgsFromFirstToLast(J9JavaVM * vm)
 
 		/* Since these two options are -XX, we don't need to consume them as they will be ignored */
 
-		if (strcmp(testString, VMOPT_XXNOSTACKTRACEINTHROWABLE) == 0) {
+		if (FIND_AND_CONSUME_ARG(EXACT_MATCH, VMOPT_XXNOSTACKTRACEINTHROWABLE, NULL) != -1) {
```
AidanHa: @pshipton The fix was made here where `XX` options are consumed here (where previously, they were not)

pshipton: ok, sorry. I didn't realize your earlier comment meant you had updated the code in this PR.

DanHeidinga:

AidanHa:
Seems like one of the extended tests failed :https://ci.eclipse.org/openj9/job/PullRequest-Extended-JDK8-linux_ppc-64_cmprssptrs_le-OpenJ9/172/console

The test id was `test -XX:+PageAlignDirectMemory -XX:-PageAlignDirectMemory` 

I tried rebasing my fork to see if anything has changed...

DanHeidinga:

runtime/vm/jvminit.c
```diff
@@ -2387,6 +2397,10 @@ static UDATA checkArgsConsumed(J9PortLibrary* portLibrary, J9VMInitArgs* j9vm_ar
 			} else if (HAS_MAPPING( j9vm_args, i) && (MAPPING_FLAGS( j9vm_args, i ) & INVALID_OPTION)) {
 				j9nls_printf(PORTLIB, J9NLS_ERROR, J9NLS_VM_INVALID_CMD_LINE_OPT, optString);
 			} else {
+				/* If ignoreUnrecognizedXXOptions is set to JNI_TRUE, we should ignore any options that start with -XX: */
+				if (ignoreUnrecongizedXXOptions && (0 == strncmp(optString, "-XX:", 4))) {
```
pshipton: This should use the macro VMOPT_XX instead of `"-XX:"`

runtime/vm/jvminit.c
pshipton: There are two problems with these changes to use FIND_AND_CONSUME_ARG, the second of which is causing the extended test failure.
- this change is inefficient, this code is in a loop. It will check for these options every time through the loop, however FIND_AND_CONSUME_ARG only needs to be called once for each option
- the order of options is not taken into account by FIND_AND_CONSUME_ARG. For options that have both a + and - variation specified, we need to check which one comes last and use that one

AidanHa: I have changed the structure for each if statement to something like the following:
`if ((FIND_ARG_IN_VMARGS(EXACT_MATCH, VMOPT_XXNOSTACKTRACEINTHROWABLE, NULL) != -1) || (FIND_ARG_IN_VMARGS(EXACT_MATCH, VMOPT_XXSTACKTRACEINTHROWABLE, NULL) != -1)) {
		if ((FIND_ARG_IN_VMARGS(EXACT_MATCH, VMOPT_XXNOSTACKTRACEINTHROWABLE, NULL) != -1) && (FIND_ARG_IN_VMARGS(EXACT_MATCH, VMOPT_XXSTACKTRACEINTHROWABLE, NULL) != -1)) {
			if (FIND_AND_CONSUME_ARG(EXACT_MATCH, VMOPT_XXNOSTACKTRACEINTHROWABLE, NULL) > FIND_AND_CONSUME_ARG(EXACT_MATCH, VMOPT_XXSTACKTRACEINTHROWABLE, NULL)) {
				vm->runtimeFlags |= J9_RUNTIME_OMIT_STACK_TRACES;
			} else {
				vm->runtimeFlags &= ~(UDATA)J9_RUNTIME_OMIT_STACK_TRACES;
			}
		} else {
			if(FIND_AND_CONSUME_ARG(EXACT_MATCH, VMOPT_XXNOSTACKTRACEINTHROWABLE, NULL) != -1) {
				vm->runtimeFlags |= J9_RUNTIME_OMIT_STACK_TRACES;
			} else if (FIND_AND_CONSUME_ARG(EXACT_MATCH, VMOPT_XXSTACKTRACEINTHROWABLE, NULL) != -1) {
				vm->runtimeFlags &= ~(UDATA)J9_RUNTIME_OMIT_STACK_TRACES;
			}
		}	
	}`

AidanHa: On a second look, I found it much easier to do the following: 

image

Where I simply just add a check where I would check if both options exist first and consume the last one if both existed.

pshipton: Please find, and consume, each argument only once. Every time FIND_ is called there is a search through all the options.

AidanHa: Okay, so I guess each statement would look like the following
`IDATA noStackTraceInThrowable = FIND_AND_CONSUME_ARG(EXACT_MATCH, VMOPT_XXNOSTACKTRACEINTHROWABLE, NULL);
`IDATA StackTraceInThrowable = FIND_AND_CONSUME_ARG(EXACT_MATCH, VMOPT_XXSTACKTRACEINTHROWABLE, NULL);
if ((noStackTraceInThrowable > stackTraceInThrowable) && (noStackTraceInThrowable != -1)) {
    vm->runtimeFlags |= J9_RUNTIME_OMIT_STACK_TRACES;
} else if ((stackTraceInThrowable > noStackTraceInThrowable) && (stackTraceInThrowable != -1)) {
    vm->runtimeFlags &= ~(UDATA)J9_RUNTIME_OMIT_STACK_TRACES;
}`

AidanHa: Looking ahead, it seems like space is an issue if we declare an IDATA variable for each if statement. Perhaps we can have a temporary variable that can be reused for each if statement? If so, what should that variable be called?

pshipton: You don't need to check `<variable> != -1`. If it is greater than the alternate variable then it can't be -1.

Please use a different set of local variables for each option. Clarity and lack of bugs is more important that saving some stack space. If you use a different scope (i.e. `{` `}`) for checking each option, it will make it easier for the compiler to take care of space optimization.

runtime/vm/jvminit.c
```diff
@@ -2963,65 +3035,127 @@ processVMArgsFromFirstToLast(J9JavaVM * vm)
...
-		} else if (0 == strcmp(testString, VMOPT_XXRESTRICTCONTENDED)) {
+		}
+	}
+
+	{
+		IDATA restrictContended = FIND_AND_CONSUME_ARG(EXACT_MATCH, VMOPT_XXRESTRICTCONTENDED, NULL);
+		IDATA noRestrictContended = FIND_AND_CONSUME_ARG(EXACT_MATCH, VMOPT_XXNORESTRICTCONTENDED, NULL);
+		if (restrictContended > noRestrictContended) {
 			vm->extendedRuntimeFlags &= ~(UDATA)J9_EXTENDED_RUNTIME_ALLOW_APPLICATION_CONTENDED_FIELDS;
-		} else if (0 == strcmp(testString, VMOPT_XXNORESTRICTCONTENDED)) { /* enabling application contended fields implicitly turns on bootstrap contended fields */
```
pshipton: Please keep the comment.

runtime/vm/jvminit.c
```diff
@@ -2963,65 +3035,127 @@ processVMArgsFromFirstToLast(J9JavaVM * vm)
...
+		} else if (restrictContended < noRestrictContended) {
 			vm->extendedRuntimeFlags |= (UDATA)J9_EXTENDED_RUNTIME_ALLOW_APPLICATION_CONTENDED_FIELDS;
 			vm->extendedRuntimeFlags |= (UDATA)J9_EXTENDED_RUNTIME_ALLOW_CONTENDED_FIELDS;
-		} else if (0 == strcmp(testString, VMOPT_XXNOCONTENDEDFIELDS)) { /* disabling bootstrap contended fields implicitly turns off application contended fields  */
+		}
+	}
+
+	if (FIND_AND_CONSUME_ARG(EXACT_MATCH, VMOPT_XXREDUCECPUMONITOROVERHEAD, NULL) != -1) {
```
pshipton: This is already checked above.

test/functional/cmdLineTests/ignoreUnrecognizedVMOptions/ignoreunrecognizedxxcolonoptions.xml
```diff
@@ -0,0 +1,133 @@
...
+ <variable name="BADOPTION3" value="-XX:+INVALID" />
+ <variable name="GOODOPTION" value="-XX:+DisableExplicitGC" />
+
+ <test id="test -XX:+IgnoreUnrecognizedXXColonOptions, bad option: -XX:INVALID">
+  <command>$EXE$ $IGNOREUNRECOGNIZEDXXOPTIONS$ $BADOPTION1$ -version</command>
+  <output type="success" caseSensitive="no" regex="yes" javaUtilPattern="yes">(java|openjdk) version</output>
+  <output type="failure" caseSensitive="no" regex="no">Command-line option unrecognised</output>
+ </test>
+
+ <test id="test -XX:+IgnoreUnrecognizedXXColonOptions, bad option: -XX:-INVALID">
```
pshipton: This test and the next seem redundant, the test above already tested this. Adding the `-` or `+` doesn't make any difference.

test/functional/cmdLineTests/ignoreUnrecognizedVMOptions/ignoreunrecognizedxxcolonoptions.xml
```diff
@@ -0,0 +1,133 @@
...
+  <output type="failure" caseSensitive="no" regex="no">Command-line option unrecognised</output>
+ </test>
+ 
+ <test id="test -XX:-IgnoreUnrecognizedXXColonOptions, bad option: -XX:INVALID">
+  <command>$EXE$ $NOIGNOREUNRECOGNIZEDXXOPTIONS$ $BADOPTION1$ -version</command>
+  <output type="success" caseSensitive="no" regex="no">Command-line option unrecognised</output>
+  <output type="failure" caseSensitive="no" regex="yes" javaUtilPattern="yes">(java|openjdk) version</output>
+ </test>
+
+ <test id="test -XX:-IgnoreUnrecognizedXXColonOptions, bad option: -XX:-INVALID">
```
pshipton: This and the next are redundant.

test/functional/cmdLineTests/ignoreUnrecognizedVMOptions/ignoreunrecognizedxxcolonoptions.xml
```diff
@@ -0,0 +1,133 @@
...
+  <output type="failure" caseSensitive="no" regex="no">Command-line option unrecognised</output>
+ </test>
+
+ <test id="test -XX:-IgnoreUnrecognizedXXColonOptions -XX:+IgnoreUnrecognizedXXColonOptions, bad option: -XX:INVALID">
+  <command>$EXE$ $NOIGNOREUNRECOGNIZEDXXOPTIONS$ $BADOPTION1$ $IGNOREUNRECOGNIZEDXXOPTIONS$ -version</command>
+  <output type="success" caseSensitive="no" regex="yes" javaUtilPattern="yes">(java|openjdk) version</output>
+  <output type="failure" caseSensitive="no" regex="no">Command-line option unrecognised</output>
+ </test>
+
+ <test id="test -XX:-IgnoreUnrecognizedXXColonOptions -XX:+IgnoreUnrecognizedXXColonOptions, bad option: -XX:-INVALID">
```
pshipton: This and the next are redundant.

test/functional/cmdLineTests/ignoreUnrecognizedVMOptions/ignoreunrecognizedxxcolonoptions.xml
```diff
@@ -0,0 +1,133 @@
...
+  <output type="failure" caseSensitive="no" regex="no">Command-line option unrecognised</output>
+ </test>
+
+ <test id="test -XX:+IgnoreUnrecognizedXXColonOptions -XX:-IgnoreUnrecognizedXXColonOptions, bad option: -XX:INVALID">
+  <command>$EXE$ $IGNOREUNRECOGNIZEDXXOPTIONS$ $BADOPTION1$ $NOIGNOREUNRECOGNIZEDXXOPTIONS$ -version</command>
+  <output type="success" caseSensitive="no" regex="no">Command-line option unrecognised</output>
+  <output type="failure" caseSensitive="no" regex="yes" javaUtilPattern="yes">(java|openjdk) version</output>
+ </test>
+
+ <test id="test -XX:+IgnoreUnrecognizedXXColonOptions -XX:-IgnoreUnrecognizedXXColonOptions, bad option: -XX:-INVALID">
```
pshipton: This and the next are redundant.

test/functional/cmdLineTests/ignoreUnrecognizedVMOptions/ignoreunrecognizedxxcolonoptions.xml
```diff
@@ -0,0 +1,133 @@
...
+  <output type="failure" caseSensitive="no" regex="no">Command-line option unrecognised</output>
+ </test>
+
+ <test id="test -XX:+IgnoreUnrecognizedXXColonOptions, bad option: -XX:-INVALID">
+  <command>$EXE$ $IGNOREUNRECOGNIZEDXXOPTIONS$ $BADOPTION2$ -version</command>
+  <output type="success" caseSensitive="no" regex="yes" javaUtilPattern="yes">(java|openjdk) version</output>
+  <output type="failure" caseSensitive="no" regex="no">Command-line option unrecognised</output>
+ </test>
+
+ <test id="test -XX:+IgnoreUnrecognizedXXColonOptions, bad option: -XX:+INVALID">
```
pshipton: We don't need so many tests. They do take some time to run. Let's remove all the single option variants like this one, and only keep the double option tests like `-XX:-IgnoreUnrecognizedXXColonOptions -XX:+IgnoreUnrecognizedXXColonOptions`

runtime/vm/jvminit.c
```diff
@@ -2387,6 +2397,10 @@ static UDATA checkArgsConsumed(J9PortLibrary* portLibrary, J9VMInitArgs* j9vm_ar
 			} else if (HAS_MAPPING( j9vm_args, i) && (MAPPING_FLAGS( j9vm_args, i ) & INVALID_OPTION)) {
 				j9nls_printf(PORTLIB, J9NLS_ERROR, J9NLS_VM_INVALID_CMD_LINE_OPT, optString);
 			} else {
+				/* If ignoreUnrecognizedXXOptions is set to JNI_TRUE, we should ignore any options that start with -XX: */
+				if (ignoreUnrecongizedXXOptions && (0 == strncmp(optString, VMOPT_XX, 4))) {
```
pshipton: The hard coded 4 should be `sizeof(VMOPT_XX) - 1`. The `- 1` removes the trailing null which is included in the sizeof.

runtime/vm/jvminit.c
AidanHa: Strange.. not too sure why this was here. Sorry about that

test/functional/cmdLineTests/ignoreUnrecognizedVMOptions/ignoreunrecognizedxxcolonoptions.xml
AidanHa: From this comment, I'll assume that we only need one `BADOPTION`

AidanHa:
The fixes have been addressed.

pshipton:

pshipton:
@llxia can you please review the test material

PR #4440, opened 2019-01-24
Preserve null check on Unsafe

liqunl:
Preserve the null check before Unsafe inlining.

Also move `TR_J9ByteCodeIlGenerator::separateNullCheck` to
`TransformUtil` so that it can be used in both ILGen and optimizer.


liqunl:
Deliver to v0.12.0-release for #4429

liqunl:
@andrewcraik

andrewcraik:
CI build willf ail for release target PR

PR #4454, opened 2019-01-25
[WIP] Removing GCC from the x86/ub16 container on Jenkins

samuel-rubin:
I want to see if we can still test the jdks tests without gcc

[skip ci]


samuel-rubin:
jenkins build docker x86 ubuntu16

PR #4472, opened 2019-01-28
Remove unused legacy tracegen tool

keithc-ca:
We have been using the OMR tool for some time; the Java implementation is no longer needed.


pshipton:
Travis build is sufficient.

PR #4474, opened 2019-01-28
Remove GCC 4 from containers 

samuel-rubin:
Removing GCC 4 because it is no longer being used to compile java
Installing GCC-7 where it is not already being installed
Adding libstdc++.so* to ldconfig when using precompiled gcc-7

[skip ci]

### DIFF
--- a/buildenv/jenkins/docker-slaves/ppc64le/centos7/Dockerfile
+++ b/buildenv/jenkins/docker-slaves/ppc64le/centos7/Dockerfile
@@ -50,8 +50,6 @@ RUN yum -y update \
     fontconfig \
     fontconfig-devel \
     freetype-devel \
-    gcc-4.8.5 \
-    gcc-c++-4.8.5 \
     gettext \
     glibc \
     glibc-common \
@@ -106,16 +104,14 @@ RUN wget -O - http://cpanmin.us | perl - --self-upgrade \
   && cpanm Text::CSV \
   && cpanm JSON
 
-# Setup links to GCC-4.8.5
-RUN rm /usr/bin/c++ /usr/bin/cc \
-  && ln -s g++ /usr/bin/c++ \
-  && ln -s gcc /usr/bin/cc
-
 # Install GCC-7.3.1
 RUN cd /usr/local \
   && wget -O gcc-7.tar.xz "https://ci.adoptopenjdk.net/userContent/gcc/gcc730+ccache.ppc64le.tar.xz" \
   && tar -xJf gcc-7.tar.xz --strip-components=1 \
-  && rm -rf gcc-7.tar.xz
+  && rm -rf gcc-7.tar.xz \
+  && ln -s /usr/local/bin/gcc-7.3 /usr/bin/cc \
+  && ln -s /usr/local/bin/gcc-7.3 /usr/bin/gcc \
+  && ln -s /usr/local/bin/g++-7.3 /usr/bin/g++ 
 
 # Edit ldconfig to connect this library
 RUN echo "/usr/local/lib64" >> /etc/ld.so.conf.d/local.conf \
--- a/buildenv/jenkins/docker-slaves/ppc64le/ubuntu16/Dockerfile
+++ b/buildenv/jenkins/docker-slaves/ppc64le/ubuntu16/Dockerfile
@@ -48,8 +48,8 @@ RUN apt-get update \
     autoconf \
     build-essential \
     curl \
-    g++-4.8 \
-    gcc-4.8 \
+    g++-7 \
+    gcc-7 \
     gdb \
     git \
     make \
--- a/buildenv/jenkins/docker-slaves/s390x/ubuntu16/Dockerfile
+++ b/buildenv/jenkins/docker-slaves/s390x/ubuntu16/Dockerfile
@@ -48,8 +48,6 @@ RUN apt-get update \
     cpio \
     curl \
     file \
-    g++-4.8 \
-    gcc-4.8 \
     gdb \
     git \
     git-core \
@@ -81,8 +79,10 @@ RUN apt-get update \
     zip \
   && rm -rf /var/lib/apt/lists/*
 
-# Install Docker module to run test framework
-RUN echo yes | cpan install JSON Text::CSV
+# Dependency required by test framework
+RUN wget -O - http://cpanmin.us | perl - --self-upgrade \
+  && cpanm Text::CSV \
+  && cpanm JSON
 
 # Install GCC-7.4
 RUN cd /usr/local \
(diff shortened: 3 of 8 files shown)


samuel-rubin:
@AdamBrousseau for review

samuel-rubin:
jenkins build docker ppc64le centos7

samuel-rubin:
I forgot to link GCC-7 on the centos containers

samuel-rubin:
I recreated the links for GCC in the centos containers

samuel-rubin:
jenkins build docker ppc64le centos7

buildenv/jenkins/docker-slaves/x86/centos6.9/Dockerfile
```diff
@@ -111,24 +111,15 @@ RUN wget -O - http://cpanmin.us | perl - --self-upgrade \
...
+# Install GCC-7.3
+RUN yum -y update \
   && yum -y install centos-release-scl \
   && yum -y install devtoolset-7-gcc devtoolset-7-binutils \
   && yum -y install devtoolset-7-gcc-c++ devtoolset-2-gcc-gfortran \
-  && yum clean all
+  && yum clean all \
+  && ln -s /opt/rh/devtoolset-7/root/usr/bin/gcc /usr/bin/cc \
+  && ln -s /opt/rh/devtoolset-7/root/usr/bin/gcc /usr/bin/gcc \
+  && ln -s /opt/rh/devtoolset-7/root/usr/bin/g++ /usr/bin/g++
```
AdamBrousseau: Is it necessary to symlink default gcc? I.e. does something need a default gcc?

buildenv/jenkins/docker-slaves/ppc64le/centos7/Dockerfile
```diff
@@ -106,16 +104,14 @@ RUN wget -O - http://cpanmin.us | perl - --self-upgrade \
...
-
 # Install GCC-7.3.1
 RUN cd /usr/local \
   && wget -O gcc-7.tar.xz "https://ci.adoptopenjdk.net/userContent/gcc/gcc730+ccache.ppc64le.tar.xz" \
   && tar -xJf gcc-7.tar.xz --strip-components=1 \
-  && rm -rf gcc-7.tar.xz
+  && rm -rf gcc-7.tar.xz \
+  && ln -s /usr/local/bin/gcc-7.3 /usr/bin/cc \
+  && ln -s /usr/local/bin/gcc-7.3 /usr/bin/gcc \
+  && ln -s /usr/local/bin/g++-7.3 /usr/bin/g++
```
AdamBrousseau: Is it necessary to symlink default gcc? I.e. does something need a default gcc?

buildenv/jenkins/docker-slaves/x86/centos6.9/Dockerfile
samuel-rubin: git requires cc and Protobuf requires gcc.

buildenv/jenkins/docker-slaves/ppc64le/centos7/Dockerfile
samuel-rubin: I need to a link to cc to install git, but I don't need the rest of the symlinks

buildenv/jenkins/docker-slaves/x86/centos6.9/Dockerfile
AdamBrousseau: ok good enough. I doubt compiler versions matter too much for those dependencies. Unless they fail outright.

AdamBrousseau:
jenkins build docker ppc64le ubuntu16

AdamBrousseau:
Related #4305

AdamBrousseau:
Talked to Shelley. I think ideally we should keep the same gcc version as the vm compiles (currently 7.3/7.4). Since we already "know" how to build a container with it.

samuel-rubin:
On Ubuntu18, the default gcc is 7.3. and for Ubuntu16 on ppc64le and x86, can I just use the ubuntu-toolchain and that will currently use gcc-7 (7.4). Are these good enough?

AdamBrousseau:
I think we need to stick to the same solution as #4568 . We need to freeze the gcc versions. Just because the default version is the one we want today, doesn't mean it will be tomorrow.
(cc @jdekonin since we were discussing gcc yesterday)

buildenv/jenkins/docker-slaves/ppc64le/centos7/Dockerfile
jdekonin: you could just use a `cc=/usr/local/bin/gcc-7.3` before the git compilation, but I guess this works too

buildenv/jenkins/docker-slaves/x86/centos6.9/Dockerfile
jdekonin: I'm not surprised this didn't cause a problem.  Running gcc out of the devtoolset requires running a script to setup the environment and libpaths for newer libraries.  If this generates a docker container that compiles openj9, I guess we leave it be though.

samuel-rubin:
I have updated the docker containers such that the test containers install gcc from Adopt

buildenv/jenkins/docker-slaves/ppc64le/ubuntu16/Dockerfile
```diff
@@ -40,36 +40,41 @@ RUN apt-get update \
...
     wget \
     xvfb \
+    xz-utils \
   && rm -rf /var/lib/apt/lists/*
 
-# Install Docker module to run test framework
-RUN echo yes | cpan install JSON Text::CSV
+# Install GCC-7.3
+RUN cd /usr/local \
+  && wget -O gcc-7.tar.xz "https://ci.adoptopenjdk.net/userContent/gcc/gcc730+ccache.ppc64le.tar.xz" \
```
jdekonin: Using these pre-compiled you will need to also add libstdc++.so* to either the LD_LIBRARY_PATH or set it up with ldconfig.  Same applies to other instance below.

samuel-rubin: I added the libstdc++.so* from the precompiled binaries to ldconfig

samuel-rubin:
@AdamBrousseau

buildenv/jenkins/docker-slaves/s390x/ubuntu16/Dockerfile
```diff
@@ -48,9 +48,6 @@ RUN apt-get update \
     cpio \
     curl \
     file \
-    g++-4.8 \
-    gcc-4.8 \
-    gdb \
```
AdamBrousseau: removing gdb intentionally? Why is it not needed?

samuel-rubin: It was not intentional. I'll go and test out the container to see if it is actually needed.

samuel-rubin:
jenkins build docker s390x ubuntu16

buildenv/jenkins/docker-slaves/s390x/ubuntu16/Dockerfile
pshipton: This comment is still pending resolution.

AdamBrousseau: We tried to test this but were blocked by not having xml parser module in the container. @jdekonin  recalls gdb being added by Colton for testing purposes in order to generate cores and we didn't think it was needed anymore. We could leave it in for now until the parser is added and we can test it properly. Thoughts?

pshipton: If you want to remove it, it can always be quickly added back if a problem is found. Although, according to that explanation we could miss getting a core file for a failure if it is needed. Leaving it in is less likely to break anything.

samuel-rubin: I added gdb back in

PR #4352, opened 2019-01-18
Remove references to legacy Java 9 classes

pdbain-ibm:
Remove references to java.lang.reflect.Module and java.lang.reflect.Layer.

Signed-off-by: Peter Bain <peter_bain@ca.ibm.com>

PR #4477, opened 2019-01-28
Fix read of FPC register

keithc-ca:
Update JIT to match https://github.com/eclipse/omr/pull/3509.

### DIFF
--- a/runtime/compiler/runtime/SignalHandler.c
+++ b/runtime/compiler/runtime/SignalHandler.c
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2000, 2018 IBM Corp. and others
+ * Copyright (c) 2000, 2019 IBM Corp. and others
  *
  * This program and the accompanying materials are made available under
  * the terms of the Eclipse Public License 2.0 which accompanies this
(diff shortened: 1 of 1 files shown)


runtime/compiler/runtime/SignalHandler.c
```diff
@@ -920,7 +920,7 @@ UDATA jit390Handler(J9VMThread* vmThread, U_32 sigType, void* sigInfo)
       const char* infoName;
       UDATA *controlPC;
       UDATA controlPCValue;
-      UDATA *controlFPC;
+      U_32 *controlFPC;
```
fjeremic: While we're in this area of code we can remove this variable all together. All we care about is the value of the `DXC` on line 948. We can fold `infoValue` directly into it and remove the `controlFPC` variable.

keithc-ca: Sure, will do.

keithc-ca: Done.

fjeremic:
Jenkins test sanity zlinux JDK8 depends eclipse/omr#3509

fjeremic:
Relaunching the build because of issues introduced in #4352.

Jenkins test sanity zlinux JDK8 depends eclipse/omr#3509

fjeremic:
Relaunching the build because of issues introduced in #4352.

Jenkins test sanity zlinux JDK8 depends eclipse/omr#3509

PR #4481, opened 2019-01-28
Remove source folder for tracegen

keithc-ca:
This should have been included in #4472.


pshipton:
There is a copyright failure

keithc-ca:
There wasn't a copyright notice in that file and if we add one, eclipse would remove it the next time it's modified.

PR #4483, opened 2019-01-28
Use class chains for by-name class records

jdmpapin:
By using the class chain rather than the ROM class, `ClassByNameRecord` and `SystemClassByNameRecord` can now validate the class chain themselves, so that there is no need to store a separate `ClassChainRecord`.

Currently `ClassChainRecord` is avoided only when the class was *first* found by one of these two records. Further improvements are possible in the future:

- A `ClassChainRecord` could be eliminated if the same class is found by name before the first use of that class that needs its class chain to match, e.g. for constant pool lookup.

- All `ClassChainRecord`s for classes that were ever found by name could be eliminated by deferring them to the end, and having by-name validations suppress them. This would require constant pool lookups to bound-check the constant pool index, and type check the constant pool entry found there.

### DIFF
--- a/doc/compiler/aot/SymbolValidationManager.md
+++ b/doc/compiler/aot/SymbolValidationManager.md
@@ -134,18 +134,14 @@ generates records relating the array class to each of its (transitive)
 component classes until the leaf component class is reached, and then
 generates a class chain validation for the leaf component class.
 
-Some class records use a ROM class offset or a class chain offset in
-order to identify the class by name:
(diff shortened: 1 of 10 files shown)


dsouzai:
fyi @r30shah @aviansie-ben

jdmpapin:
Rebased to pick up #4309 and updated z/codegen/J9AheadOfTimeCompile.cpp

dsouzai:
Jenkins test sanity xlinux,win,plinux,zlinux jdk8,jdk11

PR #4484, opened 2019-01-29
AArch64: Enable env_data64 in buildspec

knn-k:
This commit adds a line for env_data64 in buildspec/linux_aarch64.spec.
In addition, it removes a line for gc_alignObjects because it conflicts
with env_data64.


DanHeidinga:
Merging without build as we don't have the arm builds yet

PR #4485, opened 2019-01-29
Add flattening support for valuetypes

tajila:
Add flattening support for valuetypes 

This PR does not address volatile access of flattened fields
This PR does not address flattened value hidden fields
This PR does not allow flattened value contended fields
This PR does not address flattened value static fields
This PR does not address flattened value arrays

This PR introduces an option to -XX:ValueTypeFlatteningThreshold 
to set the threshold at which a flattenable value type is no 
longer flattened. 

Fields are ordered in the following manner:
- FlattenedTypeContainingDoubles
- Doubles
- FlattenedTypeContainingObjects
- Objects
- FlattenedTypeContainingSingles
- Singles

A valueType with a Q signature is flattenable. However, the JVM 
will only flatten it if the J9ClassIsFlattened bit is set in the 
classFlags (this is determined by the flattening threshold). If 
the instance field is resolved, flattened fields will have the 
J9FieldFlagFlattened bit set in the ram CP flags slot. If this
bit is set, the offset contained in the ram CP entry is not the 
field offset, rather it is an offset into the 
clazz->flattenedFieldCache[] table. Each entry in this table 
contains the J9Class for the flattened field and the real field 
offset.

When a field is flattened it does not contain a the lockword 
or padding, only instance fields. In order to do anything with a 
field one must first retrieve it with the getField bytecode, this
bytecode will create a new object and copy the contents of the 
flattened field into the new object, then place a reference to 
the new object on the stack. The new object will not have a 
lockword. If the first field requires an 8 byte alignment padding 
will be added after the header. If a type with padding bytes is 
flattened, the padding bytes will be ommitted from the the 
flattened representation. In order to determine if a ValueType 
has padding bytes, the backfilloffset is used to denote the 
starting offset of the first field. For a type without padding
this will be zero.

Some of this work is based on the PackedObjects tech preview and 
the work done by Eric Zhang.

Some of this work is based on the PackedObjects tech preview and 
the work done by Eric Zhang.

### DIFF
--- a/runtime/gc_base/ObjectAccessBarrier.cpp
+++ b/runtime/gc_base/ObjectAccessBarrier.cpp
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 1991, 2018 IBM Corp. and others
+ * Copyright (c) 1991, 2019 IBM Corp. and others
  *
  * This program and the accompanying materials are made available under
  * the terms of the Eclipse Public License 2.0 which accompanies this
@@ -1401,9 +1401,10 @@ MM_ObjectAccessBarrier::copyObjectFields(J9VMThread *vmThread, J9Class *objectCl
 	UDATA limit = objectClass->totalInstanceSize;
 
 	if (isValueType) {
-		if (0 <= (IDATA)objectClass->lockOffset) {
-			/* skip lockword */
-			limit -= sizeof(j9objectmonitor_t);
+		U_32 firstFieldOffset = (U_32) objectClass->backfillOffset;
+		if (0 != firstFieldOffset) {
+			/* subtract padding */
+			offset += firstFieldOffset;
 			descriptionBits >>= 1;
 			descriptionIndex -= 1;
 		}
--- a/runtime/gc_include/ObjectAccessBarrierAPI.hpp
+++ b/runtime/gc_include/ObjectAccessBarrierAPI.hpp
@@ -351,16 +351,9 @@ public:
 	}
 
 	VMINLINE UDATA
-	mixedObjectGetDataSize(J9Class *objectClass, bool dontIncludeLockword)
+	mixedObjectGetDataSize(J9Class *objectClass)
 	{
-		UDATA size = objectClass->totalInstanceSize;
-		if (dontIncludeLockword) {
-			if ((0 <= (IDATA)objectClass->lockOffset)) {
-				size -= sizeof(j9objectmonitor_t);
-			}
-		}
-
-		return size;
+		return objectClass->totalInstanceSize;
 	}
 
 	VMINLINE void
@@ -395,8 +388,11 @@ public:
 			}
 		} else {
 			UDATA offset = 0;
-			bool isValueType = J9_IS_J9CLASS_VALUETYPE(objectClass);
-			UDATA limit = mixedObjectGetDataSize(objectClass, isValueType);
+			UDATA limit = mixedObjectGetDataSize(objectClass);
+			
+			if (J9_IS_J9CLASS_VALUETYPE(objectClass)) {
+				offset += objectClass->backfillOffset;
+			}
 			
 			while (offset < limit) {
 				*(fj9object_t *)((UDATA)destObject + offset + destOffset) = *(fj9object_t *)((UDATA)srcObject + offset + srcOffset);
--- a/runtime/jvmti/jvmtiHook.c
+++ b/runtime/jvmti/jvmtiHook.c
@@ -1116,7 +1116,14 @@ findFieldIndexFromOffset(J9VMThread *currentThread, J9Class *clazz, UDATA offset
 		J9ROMClass * const romClass = clazz->romClass;
 		J9Class * const superclazz = GET_SUPERCLASS(clazz);
 		J9ROMFieldOffsetWalkState state;
-		J9ROMFieldOffsetWalkResult *result = vmFuncs->fieldOffsetsStartDo(vm, romClass, superclazz, &state, walkFlags);
+		J9ROMFieldOffsetWalkResult *result = NULL;
+
+#if defined(J9VM_OPT_VALHALLA_VALUE_TYPES)
+		result = vmFuncs->fieldOffsetsStartDo(vm, romClass, superclazz, &state, walkFlags, clazz->flattenedClassCache);
+#else /* defined(J9VM_OPT_VALHALLA_VALUE_TYPES) */
+		result = vmFuncs->fieldOffsetsStartDo(vm, romClass, superclazz, &state, walkFlags);
+#endif /* defined(J9VM_OPT_VALHALLA_VALUE_TYPES) */
+
 		while (NULL != result->field) {
 			if (staticBit == (result->field->modifiers & J9AccStatic)) {
 				if (offset == result->offset) {
--- a/runtime/vm/BytecodeInterpreter.hpp
+++ b/runtime/vm/BytecodeInterpreter.hpp
@@ -6415,12 +6415,47 @@ retry:
 				{
 					UDATA const newValueOffset = valueOffset + J9_OBJECT_HEADER_SIZE;
 					bool isVolatile = (0 != (flags & J9AccVolatile));
+
 					if (flags & J9FieldSizeDouble) {
 						_sp += (slotsToPop - 2);
 						*(U_64*)_sp = _objectAccessBarrier.inlineMixedObjectReadU64(_currentThread, objectref, newValueOffset, isVolatile);
 					} else if (flags & J9FieldFlagObject) {
 						_sp += (slotsToPop - 1);
-						*(j9object_t*)_sp = _objectAccessBarrier.inlineMixedObjectReadObject(_currentThread, objectref, newValueOffset, isVolatile);
+#if defined(J9VM_OPT_VALHALLA_VALUE_TYPES)
+						if (flags & J9FieldFlagFlattened) {
+							J9FlattenedClassCache *cache = J9OBJECT_CLAZZ(_currentThread, objectref)->flattenedClassCache + valueOffset;
+							J9Class *flattenedFieldClass = cache->clazz;
+							j9object_t newObjectRef = _objectAllocate.inlineAllocateObject(_currentThread, flattenedFieldClass, false, false);
+
+							if (NULL == newObjectRef) {
+								buildGenericSpecialStackFrame(REGISTER_ARGS, 0);
+								pushObjectInSpecialFrame(REGISTER_ARGS, objectref);
+								updateVMStruct(REGISTER_ARGS);
+								newObjectRef = _vm->memoryManagerFunctions->J9AllocateObject(_currentThread, flattenedFieldClass, J9_GC_ALLOCATE_OBJECT_NON_INSTRUMENTABLE);
+								VMStructHasBeenUpdated(REGISTER_ARGS);
+								objectref = popObjectInSpecialFrame(REGISTER_ARGS);
+								restoreGenericSpecialStackFrame(REGISTER_ARGS);
+								if (J9_UNEXPECTED(NULL == newObjectRef)) {
+									rc = THROW_HEAP_OOM;
+									goto done;
+								}
+								flattenedFieldClass = VM_VMHelpers::currentClass(flattenedFieldClass);
+							}
+
+							_objectAccessBarrier.copyObjectFields(_currentThread,
+												flattenedFieldClass,
+												objectref,
+												cache->offset + J9_OBJECT_HEADER_SIZE,
+												newObjectRef,
+												J9_OBJECT_HEADER_SIZE);
+
+							_sp += (slotsToPop - 1);
+							*(j9object_t*)_sp = newObjectRef;
+						} else
+#endif /* defined(J9VM_OPT_VALHALLA_VALUE_TYPES) */
+						{
+							*(j9object_t*)_sp = _objectAccessBarrier.inlineMixedObjectReadObject(_currentThread, objectref, newValueOffset, isVolatile);
+						}
 					} else {
 						_sp += (slotsToPop - 1);
 						*(U_32*)_sp = _objectAccessBarrier.inlineMixedObjectReadU32(_currentThread, objectref, newValueOffset, isVolatile);
@@ -6516,6 +6551,7 @@ resolve:
 		{
 			bool isVolatile = (0 != (flags & J9AccVolatile));
 			UDATA const newValueOffset = valueOffset + J9_OBJECT_HEADER_SIZE;
+
 			if (flags & J9FieldSizeDouble) {
 				j9object_t objectref = *(j9object_t*)(_sp + 2);
 				if (NULL == objectref) {
@@ -6530,7 +6566,22 @@ resolve:
 					rc = THROW_NPE;
 					goto done;
 				}
-				_objectAccessBarrier.inlineMixedObjectStoreObject(_currentThread, objectref, newValueOffset, *(j9object_t*)_sp, isVolatile);
+#if defined(J9VM_OPT_VALHALLA_VALUE_TYPES)
+				if (flags & J9FieldFlagFlattened) {
+					J9FlattenedClassCache *cache = J9OBJECT_CLAZZ(_currentThread, objectref)->flattenedClassCache + valueOffset;
+
+					_objectAccessBarrier.copyObjectFields(_currentThread,
+										cache->clazz,
+										*(j9object_t*)_sp,
+										J9_OBJECT_HEADER_SIZE,
+										objectref,
+										cache->offset + J9_OBJECT_HEADER_SIZE);
+
+				} else
+#endif /* defined(J9VM_OPT_VALHALLA_VALUE_TYPES) */
+				{
+					_objectAccessBarrier.inlineMixedObjectStoreObject(_currentThread, objectref, newValueOffset, *(j9object_t*)_sp, isVolatile);
+				}
 				_sp += 2;
 			} else {
 				j9object_t objectref = *(j9object_t*)(_sp + 1);
@@ -8182,6 +8233,7 @@ retry:
 		UDATA const flags = ramFieldRef->flags;
 		UDATA const valueOffset = ramFieldRef->valueOffset;
 		j9object_t copyObjectRef = NULL;
+		J9Class *objectRefClass = NULL;
 
 		/* In a resolved field, flags will have the J9FieldFlagResolved bit set, thus
 		 * having a higher value than any valid valueOffset.
@@ -8210,7 +8262,6 @@ retry:
 		}
 		{
 			j9object_t originalObjectRef = *(j9object_t *)(_sp + (J9_ARE_ALL_BITS_SET(flags, J9FieldSizeDouble) ? 2 : 1));
-			J9Class *objectRefClass = NULL;
 
 			if (NULL == originalObjectRef) {
 				rc = THROW_NPE;
@@ -8234,6 +8285,7 @@ retry:
 				copyObjectRef = _vm->memoryManagerFunctions->J9AllocateObject(_currentThread, objectRefClass, J9_GC_ALLOCATE_OBJECT_NON_INSTRUMENTABLE);
 				VMStructHasBeenUpdated(REGISTER_ARGS);
 				originalObjectRef = popObjectInSpecialFrame(REGISTER_ARGS);
+				restoreGenericSpecialStackFrame(REGISTER_ARGS);
 				if (J9_UNEXPECTED(NULL == copyObjectRef)) {
 					rc = THROW_HEAP_OOM;
 					goto done;
@@ -8250,7 +8302,17 @@ retry:
 				_objectAccessBarrier.inlineMixedObjectStoreU64(_currentThread, copyObjectRef, newValueOffset, *(U_64*)_sp, isVolatile);
 				_sp += 2;
 			} else if (J9_ARE_ALL_BITS_SET(flags, J9FieldFlagObject)) {
-				_objectAccessBarrier.inlineMixedObjectStoreObject(_currentThread, copyObjectRef, newValueOffset, *(j9object_t*)_sp, isVolatile);
+				if (J9_ARE_ALL_BITS_SET(flags, J9FieldFlagFlattened)) {
+					J9FlattenedClassCache *cache = objectRefClass->flattenedClassCache + valueOffset;
+					_objectAccessBarrier.copyObjectFields(_currentThread,
+										cache->clazz,
+										*(j9object_t*)_sp,
+										J9_OBJECT_HEADER_SIZE,
+										copyObjectRef,
+										cache->offset + J9_OBJECT_HEADER_SIZE);
+				} else {
+					_objectAccessBarrier.inlineMixedObjectStoreObject(_currentThread, copyObjectRef, newValueOffset, *(j9object_t*)_sp, isVolatile);
+				}
 				_sp += 1;
 			} else {
 				_objectAccessBarrier.inlineMixedObjectStoreU32(_currentThread, copyObjectRef, newValueOffset, *(U_32*)_sp, isVolatile);
--- a/runtime/vm/ObjectFieldInfo.cpp
+++ b/runtime/vm/ObjectFieldInfo.cpp
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2015, 2017 IBM Corp. and others
+ * Copyright (c) 2015, 2019 IBM Corp. and others
  *
  * This program and the accompanying materials are made available under
  * the terms of the Eclipse Public License 2.0 which accompanies this
@@ -25,6 +25,7 @@
 #include "j9modifiers_api.h"
 #include "ObjectFieldInfo.hpp"
 #include "util_api.h"
+#include "vm_api.h"
 
 /**
  * Calculate the number of various types of fields.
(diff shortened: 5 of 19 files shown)


tajila:

runtime/jvmti/jvmtiHook.c
```diff
@@ -1112,7 +1112,14 @@ findFieldIndexFromOffset(J9VMThread *currentThread, J9Class *clazz, UDATA offset
 		J9ROMClass * const romClass = clazz->romClass;
 		J9Class * const superclazz = GET_SUPERCLASS(clazz);
 		J9ROMFieldOffsetWalkState state;
-		J9ROMFieldOffsetWalkResult *result = vmFuncs->fieldOffsetsStartDo(vm, romClass, superclazz, &state, walkFlags);
+		J9ROMFieldOffsetWalkResult *result = NULL;
+
+#if defined(J9VM_OPT_VALHALLA_VALUE_TYPES)
+		state.classLoader = clazz->classLoader;
```
gacholio: Wouldn't it be safe to just do this all the time?

tajila: The classloader is only needed for value types. If we assign it here we will have to do the same in all the other places.

gacholio: Which we should do, since those places will fall over on value types otherwise. Why do we need the loader?

gacholio:
I'd prefer to see no lockword for VTs, as I think this is the way things will eventually have to work. Either way, there's no value embedding them in the object - we should use the side table if we need to lock VTs at all.

tajila:
I'll remove the lock word and post an update

runtime/gc_include/ObjectAccessBarrierAPI.hpp
```diff
@@ -351,16 +351,9 @@ class MM_ObjectAccessBarrierAPI
...
 	{
-		UDATA size = objectClass->totalInstanceSize;
-		if (dontIncludeLockword) {
-			if ((0 <= (IDATA)objectClass->lockOffset)) {
-				size -= sizeof(j9objectmonitor_t);
-			}
-		}
-
-		return size;
+		return objectClass->totalInstanceSize;
```
gacholio: This is MM_MixedObjectModel::getSizeInBytesWithoutHeader

runtime/jvmti/jvmtiHook.c
```diff
@@ -1112,7 +1112,12 @@ findFieldIndexFromOffset(J9VMThread *currentThread, J9Class *clazz, UDATA offset
 		J9ROMClass * const romClass = clazz->romClass;
 		J9Class * const superclazz = GET_SUPERCLASS(clazz);
 		J9ROMFieldOffsetWalkState state;
-		J9ROMFieldOffsetWalkResult *result = vmFuncs->fieldOffsetsStartDo(vm, romClass, superclazz, &state, walkFlags);
+		J9ROMFieldOffsetWalkResult *result = NULL;
+
+		state.classLoader = clazz->classLoader;
```
gacholio: As usual, I'm flip-flopping here. If this point is guaranteed to not need the loader, let's set it to NULL and assert in the offsets code is you need to load something and the loader has not been provided.

The callers of this code can not handle java code running, so let's make the NULL be the indicator for that.

runtime/vm/BytecodeInterpreter.hpp
```diff
@@ -7221,7 +7272,7 @@ done:;
 		J9ConstantPool *ramConstantPool = J9_CP_FROM_METHOD(_literals);
 		J9RAMClassRef *ramCPEntry = ((J9RAMClassRef*)ramConstantPool) + index;
 		J9Class* volatile resolvedClass = ramCPEntry->value;
-		if ((NULL != resolvedClass) && (resolvedClass->romClass->modifiers & (J9AccAbstract | J9AccInterface)) == 0) {
+		if ((NULL != resolvedClass) && (resolvedClass->romClass->modifiers & (J9AccAbstract | J9AccInterface )) == 0) {
```
gacholio: Incorrect spacing change, and will conflict with something about to be merged - please removed.

runtime/vm/ObjectFieldInfo.cpp
```diff
@@ -44,6 +45,23 @@ ObjectFieldInfo::countInstanceFields(void)
 		const U_32 modifiers = field->modifiers;
 		if (J9_ARE_NO_BITS_SET(modifiers, J9AccStatic) ) {
 
+#ifdef J9VM_OPT_VALHALLA_VALUE_TYPES
+			J9UTF8 *fieldSig = J9ROMFIELDSHAPE_SIGNATURE(field);
+			U_8 *fieldSigBytes = J9UTF8_DATA(J9ROMFIELDSHAPE_SIGNATURE(field));
+			if ('Q' == *fieldSigBytes) {
+				J9Class *fieldClass = hashClassTableAt(_classLoader, fieldSigBytes + 1, J9UTF8_LENGTH(fieldSig) - 2);
```
gacholio: No possibility of failure to find?

runtime/vm/createramclass.cpp
```diff
@@ -1634,49 +1643,64 @@ loadSuperClassAndInterfaces(J9VMThread *vmThread, J9ClassLoader *classLoader, J9
...
-			} else {
-				J9ROMClass *valueROMClass = valueClass->romClass;
-
-				if (J9_ARE_NO_BITS_SET(valueROMClass->modifiers, J9AccValueType)) {
-					J9UTF8 *badClass = NNSRP_GET(valueROMClass->className, J9UTF8*);
-					setCurrentExceptionNLSWithArgs(currentThread, J9NLS_VM_ERROR_QTYPE_NOT_VALUE_TYPE, J9VMCONSTANTPOOL_JAVALANGINCOMPATIBLECLASSCHANGEERROR, J9UTF8_LENGTH(badClass), J9UTF8_DATA(badClass));
+		if (J9_ARE_NO_BITS_SET(modifiers, J9AccStatic)) {
+			J9UTF8 *signature = J9ROMFIELDSHAPE_SIGNATURE(field);
+			U_8 *signatureChars = J9UTF8_DATA(signature);
+			if (('Q' == signatureChars[0]) && J9_ARE_NO_BITS_SET(modifiers, J9AccStatic)) {
```
gacholio: Extra static check here. Would this read better as a switch statement instead of elseif?

runtime/vm/createramclass.cpp
```diff
@@ -1634,49 +1643,64 @@ loadSuperClassAndInterfaces(J9VMThread *vmThread, J9ClassLoader *classLoader, J9
...
+				J9Class *valueClass = internalFindClassUTF8(currentThread, signatureChars + 1, J9UTF8_LENGTH(signature) - 2, classLoader, classPreloadFlags);
+				if (NULL == valueClass) {
 					result = FALSE;
 					goto done;
-				}
+				} else {
+					J9ROMClass *valueROMClass = valueClass->romClass;
 
-				bool classIsPublic = J9_ARE_ALL_BITS_SET(valueROMClass->modifiers, J9AccPublic);
+					if (J9_ARE_NO_BITS_SET(valueROMClass->modifiers, J9AccValueType)) {
```
gacholio: Should these checks move into a new helper for loading VT classes?

gacholio: i.e. a helper to load a single VT class by name from another class.

runtime/vm/createramclass.cpp
```diff
@@ -2872,16 +2921,20 @@ internalCreateRAMClassFromROMClass(J9VMThread *vmThread, J9ClassLoader *classLoa
 
 	if (!loadSuperClassAndInterfaces(vmThread, hostClassLoader, romClass, elementClass, packageID, hotswapping, classPreloadFlags, &superclass, module)
 #if defined(J9VM_OPT_VALHALLA_VALUE_TYPES)
-		|| !loadFlattenableFieldValueClasses(vmThread, hostClassLoader, romClass, classPreloadFlags, packageID, module)
+		|| !loadFlattenableFieldValueClasses(vmThread, hostClassLoader, romClass, classPreloadFlags, packageID, module, &largestFieldType, &flattenableInstanceFieldCount)
```
gacholio: Haven't all the classes been loaded already by the field offset walk? I'd rather see the VT classes loaded ASAP and then not pass the loader to the field offset walk (so that it cannot fail).

runtime/vm/resolvefield.cpp
```diff
@@ -608,14 +611,26 @@ fieldOffsetsStartDo(J9JavaVM *vm, J9ROMClass *romClass, J9Class *superClazz, J9R
 {
 
 	Trc_VM_romFieldOffsetsStartDo_Entry( NULL, romClass, superClazz, flags );
+#ifdef J9VM_OPT_VALHALLA_VALUE_TYPES
+	/* save the classLaoder */
+	J9ClassLoader *classLoader = state->classLoader;
```
gacholio: Very much yuck. Perhaps investigate passing the loader as a parameter - it's more explicit.

gacholio: Discussed with @DanHeidinga and we agree - let's have all the input data as parameters to the functions - feel free to store it in the state for later use.

runtime/jvmti/jvmtiHook.c
gacholio: I'm wrong - we always need the loader, but we'll be passing it as a parameter to the function (see below).

runtime/vm/resolvefield.cpp
```diff
@@ -909,8 +937,34 @@ fieldOffsetsFindNext(J9ROMFieldOffsetWalkState *state, J9ROMFieldShape *field)
 							state->result.offset = state->backfillOffsetToUse;
 							state->walkFlags &= ~(UDATA)J9VM_FIELD_OFFSET_WALK_BACKFILL_OBJECT_FIELD;
 						} else {
-							state->result.offset = state->firstObjectOffset + state->objectsSeen * sizeof( fj9object_t );
-							state->objectsSeen++;
+#ifdef J9VM_OPT_VALHALLA_VALUE_TYPES
+							J9UTF8 *fieldSig = J9ROMFIELDSHAPE_SIGNATURE(field);
+							U_8 *fieldSigBytes = J9UTF8_DATA(fieldSig);
+							if ('Q' == *fieldSigBytes) {
+								J9Class *fieldClass = hashClassTableAt(state->classLoader, fieldSigBytes + 1, J9UTF8_LENGTH(fieldSig) - 2);
```
gacholio: Please assert that we found the class here.

runtime/vm/resolvesupport.cpp
```diff
@@ -972,7 +972,31 @@ resolveInstanceFieldRefInto(J9VMThread *vmStruct, J9Method *method, J9ConstantPo
 		
 			if (ramCPEntry != NULL) {
 				UDATA valueOffset = fieldOffset;
+#if defined(J9VM_OPT_VALHALLA_VALUE_TYPES)
+				if ('Q' == *J9UTF8_DATA(signature)) {
+					J9Class *flattenableClass = hashClassTableAt(classFromCP->classLoader, J9UTF8_DATA(signature) + 1, J9UTF8_LENGTH(signature) - 2);
```
gacholio: Assert again,

runtime/vm/ObjectFieldInfo.cpp
gacholio: I'd rather see this is a call to internalFindClassUTF8 with flags preventing java code from running (find existing only, no exception throw). Should also assert that the returned value is not NULL.

runtime/vm/resolvefield.cpp
gacholio: With VTs as a possibility, we will always need to pass a classLoader in every caller, but will not need to run java code (assuming we prepopulate the hash table).

gacholio:
@charliegracie - I went over the GC changes with @dmitripivkine and he suggested that you have a quick glance.

runtime/gc_include/ObjectAccessBarrierAPI.hpp
charliegracie: So before this was only messing with the size for value type objects since that was the only place where `dontIncludeLockword` was true?

runtime/gc_include/ObjectAccessBarrierAPI.hpp
```diff
@@ -395,8 +388,11 @@ class MM_ObjectAccessBarrierAPI
 			}
 		} else {
 			UDATA offset = 0;
-			bool isValueType = J9_IS_J9CLASS_VALUETYPE(objectClass);
-			UDATA limit = mixedObjectGetDataSize(objectClass, isValueType);
+			UDATA limit = mixedObjectGetDataSize(objectClass);
+			
+			if (J9_IS_J9CLASS_VALUETYPE(objectClass)) {
+				offset += objectClass->backfillOffset;
```
charliegracie: `backfillOffset` is the start of the first field for value types?

charliegracie: if so can we just make it `offset = objectClass->backfillOffset` as I think that is clearer.  Anyone disagree?

charliegracie:
I have made some comments and suggested 1 simple change.

runtime/gc_include/ObjectAccessBarrierAPI.hpp
gacholio: I don't think two assignments is clearer (though clearly more efficient), unless there's clearly in an if/else or ternary path.

charliegracie: My concern is that if someone changes the default value of the `offset` variable we may do the incorrect thing by accident.  Another suggestion then.  Since offset is always 0 for non value types and backfillOffset for value types maybe should just add backfillOffset to [src,dest]Offset when it is a value type?

gacholio: Assigning would have the opposite issue (if offset were meaningful before the VT test, an assignment would lose any previous computation).

As the VT test is only doing this addition, would a ternary assignment be better for you?

charliegracie: Hmmm I don't really like the ternary operations lol.  We discussed it and there is technically nothing wrong with the current code so I say just leave it.

runtime/gc_include/ObjectAccessBarrierAPI.hpp
gacholio: Before, VTs were all getting lockwords. Now they are never getting them, so this is the same as the default.

charliegracie: I dug a bit deeper and I am ok with this.

tajila:
@gacholio changes are ready for another look

runtime/gc_base/ObjectAccessBarrier.cpp
```diff
@@ -1401,9 +1401,14 @@ MM_ObjectAccessBarrier::copyObjectFields(J9VMThread *vmThread, J9Class *objectCl
...
-		if (0 <= (IDATA)objectClass->lockOffset) {
-			/* skip lockword */
-			limit -= sizeof(j9objectmonitor_t);
+		/* For value types the backFillOffset is the start of the first field.
+		 * If this is not zero then we have added padding bytes to satisfy alignment
+		 * of double slot fields. Skip the padding bytes when copying objects over.
+		 */
+		U_32 firstFieldOffset = (U_32) objectClass->backfillOffset;
+		if (0 != firstFieldOffset) {
+			/* subtract padding */
```
gacholio: Comment says subtract, code says add...

gacholio: The comment above is complete, so this one could probably just be deleted.

runtime/vm/BytecodeInterpreter.hpp
```diff
@@ -8226,7 +8278,16 @@ done:;
 			bool const isVolatile = (0 != (flags & J9AccVolatile));
 			UDATA const newValueOffset = valueOffset + J9_OBJECT_HEADER_SIZE;
 
-			if (J9_ARE_ALL_BITS_SET(flags, J9FieldSizeDouble)) {
+			if (flags & J9FieldFlagFlattened) {
```
gacholio: Please retain the macro to match the other cases in this code.

runtime/vm/ObjectFieldInfo.cpp
```diff
@@ -44,6 +45,26 @@ ObjectFieldInfo::countInstanceFields(void)
 		const U_32 modifiers = field->modifiers;
 		if (J9_ARE_NO_BITS_SET(modifiers, J9AccStatic) ) {
 
+#ifdef J9VM_OPT_VALHALLA_VALUE_TYPES
+			J9UTF8 *fieldSig = J9ROMFIELDSHAPE_SIGNATURE(field);
+			U_8 *fieldSigBytes = J9UTF8_DATA(J9ROMFIELDSHAPE_SIGNATURE(field));
+			if ('Q' == *fieldSigBytes) {
```
gacholio: Why can modifiers not be used here?

gacholio: I realize now -  but can this be nested inside the object case below? I think making Q behaviour default to L behaviour unless otherwise directed makes sense.

runtime/vm/resolvefield.cpp
```diff
@@ -602,9 +608,48 @@ addHiddenInstanceField(J9JavaVM *vm, const char *className, const char *fieldNam
...
 #endif /* !defined(J9VM_OUT_OF_PROCESS) */
 
+#ifdef J9VM_OPT_VALHALLA_VALUE_TYPES
+J9Class *
+findJ9ClassInFlattenedClassCache(J9FlattenedClassCache *flattenedClassCache, U_8 *className, UDATA classNameLength)
+{
+	/* first field indicates the number of classes in the cache */
+	UDATA length = flattenedClassCache->offset;
+	J9Class *clazz = NULL;
+	for (UDATA i = 1; i < (length + 1); i++) {
```
gacholio: This might read better as `<= length` (similarly below).

runtime/vm/resolvesupport.cpp
```diff
@@ -972,7 +972,30 @@ resolveInstanceFieldRefInto(J9VMThread *vmStruct, J9Method *method, J9ConstantPo
...
 				UDATA valueOffset = fieldOffset;
-
+#if defined(J9VM_OPT_VALHALLA_VALUE_TYPES)
+				if ('Q' == *J9UTF8_DATA(signature)) {
+					J9FlattenedClassCache *flattenedClassCache = classFromCP->flattenedClassCache;
+					J9Class *flattenableClass = NULL;
+					UDATA index = 0;
+
+					Assert_VM_notNull(flattenedClassCache);
+					index = findIndexInFlattenedClassCache(flattenedClassCache, nameAndSig);
```
gacholio: Is there ever a case where you expect not to find the index? Wondering if this asserts could move into the callee.

runtime/vm/createramclass.cpp
```diff
@@ -1633,53 +1643,87 @@ loadSuperClassAndInterfaces(J9VMThread *vmThread, J9ClassLoader *classLoader, J9
...
-				result = FALSE;
-				goto done;
-			} else {
-				J9ROMClass *valueROMClass = valueClass->romClass;
+		if (J9_ARE_NO_BITS_SET(modifiers, J9AccStatic)) {
+			J9UTF8 *signature = J9ROMFIELDSHAPE_SIGNATURE(field);
+			U_8 *signatureChars = J9UTF8_DATA(signature);
+			switch (signatureChars[0]) {
+			case 'Q':
+				{
```
gacholio: I'd just as soon not see another level of indent here. I'd either treat the case like an if (my preference), or move the curlies to the same indent as the case.

runtime/vm/createramclass.cpp
```diff
@@ -1633,53 +1643,87 @@ loadSuperClassAndInterfaces(J9VMThread *vmThread, J9ClassLoader *classLoader, J9
...
+
+						J9FlattenedClassCache *entry = flattenedClassCache + (flattenableInstanceFieldCount + 1);
+						entry->clazz = valueClass;
+						entry->nameAndSignature = &(field->nameAndSignature);
+						entry->offset = 0;
+						flattenableInstanceFieldCount += 1;
+					}
+					/* Dont check if previously set, or in the values to avoid
+					 * extra branching, we'll only write one of them into the classFlags.
+					 */
```
gacholio: I don't understand the comment at all. Have you started using `classFlags` (a bad idea)?

runtime/vm/createramclass.cpp
```diff
@@ -2877,19 +2956,43 @@ internalCreateRAMClassFromROMClass(J9VMThread *vmThread, J9ClassLoader *classLoa
 			}
 		}
 	}
+#if defined(J9VM_OPT_VALHALLA_VALUE_TYPES)
+	if (romFieldCount > DEFAULT_SIZE_OF_CLASS_CACHE) {
+		flattenedClassCache = (J9FlattenedClassCache *) j9mem_allocate_memory(sizeof(J9FlattenedClassCache) * (romFieldCount + 1), J9MEM_CATEGORY_CLASSES);
```
gacholio: Maybe put the alloc/zero size in to a local.

runtime/vm/createramclass.cpp
```diff
@@ -2913,6 +3016,16 @@ internalCreateRAMClassFromROMClass(J9VMThread *vmThread, J9ClassLoader *classLoa
 #if defined(J9VM_OPT_VALHALLA_VALUE_TYPES)
 		if (J9_ARE_ALL_BITS_SET(romClass->modifiers, J9AccValueType)) {
 			classFlags |= J9ClassIsValueType;
+			if ((result->totalInstanceSize <= javaVM->valueFlatteningThreshold) && !J9ROMCLASS_IS_CONTENDED(romClass)) {
```
gacholio: Is there no room left in `classDepthAndFlags`?

gacholio:
@pdbain-ibm Can you please have a look at the ObjectFieldInfo/description changes?

runtime/vm/resolvesupport.cpp
```diff
@@ -972,7 +972,30 @@ resolveInstanceFieldRefInto(J9VMThread *vmStruct, J9Method *method, J9ConstantPo
...
+					Assert_VM_true(index != 0);
+					flattenableClass = flattenedClassCache[index].clazz;
+					Assert_VM_notNull(flattenableClass);
+
+					if (J9_ARE_ALL_BITS_SET(flattenableClass->classFlags, J9ClassIsFlattened)) {
```
gacholio: Should we nest flattened under object?

runtime/vm/createramclass.cpp
gacholio: I see they are already in use. Either way, you should not be using the whole field, you need to mask the bits you actually care about (if you're masking in the caller, I suppose this isn't necessary, though it's a bit odd returning a pile of meaningless bits).

runtime/vm/createramclass.cpp
gacholio: Ignore this.

tajila:
@gacholio I have updated the PR

runtime/vm/resolvesupport.cpp
gacholio: Still waiting for an answer here - if there is no case where you test the return value other than asserting it, the asserts may as well be in the callee.

runtime/vm/createramclass.cpp
gacholio: Is the size (fieldCount + 1) used other than to be mutitplied? If not, the full value may as well be in the local.

runtime/vm/resolvesupport.cpp
tajila: I moved the assertion to the callee, just forgot to remove it from the caller. I have made that change now.

runtime/vm/createramclass.cpp
tajila: I have moved the full value to the local

runtime/vm/resolvesupport.cpp
```diff
@@ -972,7 +972,27 @@ resolveInstanceFieldRefInto(J9VMThread *vmStruct, J9Method *method, J9ConstantPo
 		
 			if (ramCPEntry != NULL) {
 				UDATA valueOffset = fieldOffset;
+#if defined(J9VM_OPT_VALHALLA_VALUE_TYPES)
+				if ('Q' == *J9UTF8_DATA(signature)) {
+					J9FlattenedClassCache *flattenedClassCache = classFromCP->flattenedClassCache;
+					J9Class *flattenableClass = NULL;
+					UDATA index = 0;
+
+					Assert_VM_notNull(flattenedClassCache);
```
gacholio: Should this assert be removed?

tajila: I've removed the assertion

runtime/vm/resolvesupport.cpp
```diff
@@ -972,7 +972,26 @@ resolveInstanceFieldRefInto(J9VMThread *vmStruct, J9Method *method, J9ConstantPo
 		
 			if (ramCPEntry != NULL) {
 				UDATA valueOffset = fieldOffset;
+#if defined(J9VM_OPT_VALHALLA_VALUE_TYPES)
+				if ('Q' == *J9UTF8_DATA(signature)) {
+					J9FlattenedClassCache *flattenedClassCache = classFromCP->flattenedClassCache;
+					J9Class *flattenableClass = NULL;
+					UDATA index = 0;
```
gacholio: Last nitpick - declare and assign index.

tajila: done

gacholio:

gacholio:
Does DDR need to be modified to understand the new layouts?

PR #4487, opened 2019-01-29
AArch64: Add changes for building for AArch64

knn-k:
This commit fixes various compilation errors found during AArch64
Linux builds.


knn-k:
See also #4087, #4350 and #4484.

knn-k:
Added a change in `runtime/gc_glue_java/configure_includes/configure_linux_aarch64.mk`.
Without the change, the build process fails in generating `omr/tools/toolconfigure.mk`.

knn-k:
Added a change in the `Dockerfie` for aarch64 cross-compilation, updating the URL for adoptopenjdk.net.  Fetching the JDK now fails with `400: Bad Request` if the URL does not have `&type=jdk` in it.

knn-k:
I would like to have this PR reviewed, please.  Thanks.

keithc-ca:
Jenkins compile zlinux jdk11

PR #4491, opened 2019-01-29
Stop emitting SVM records for primitive arrays

jdmpapin:
Even with guaranteed IDs for the primitive types and their (1-dimensional) array types, the symbol validation manager (SVM) could still emit redundant records for these array-component relationships. For example, if `byte[]` were found by name, `addClassByNameRecord()` would create an `ArrayClassFromComponentClass` record, and the record would actually be emitted because it isn't a duplicate.

Since such records are never necessary, the `_alreadyGeneratedRecords` set will now be seeded with them so that they are no longer emitted.


dsouzai:

PR #4493, opened 2019-01-29
Fix jdmpview check for valid logger name

keithc-ca:
* print the intended message rather than throw a NullPointerException


pshipton:
jenkins test all plinux jdk8

PR #4496, opened 2019-01-29
Fix DDR test for free tag

keithc-ca:
Instead of the test succeeding, it would throw:
> InvalidDataTypeException: U64 contains value larger than Long.MAX_VALUE


keithc-ca:
This is the next issue anticipated by https://github.com/eclipse/openj9/issues/4114#issuecomment-452073708.

pshipton:
jenkins test all zlinux jdk8

PR #4499, opened 2019-01-29
Add logic to specify test with different aot tags

renfeiw:
- test can be tagged with aot applicable, nonapplicable and explicit
- when TEST_FLAG contains aot, the tests will behave differently as follows
- nonapplicable test will not be generated
- applicable test will be specially generated (iterations and aot options)
- explicit test will be generated as is

[ci skip]


renfeiw:
reviewer: @llxia

llxia:

PR #4501, opened 2019-01-29
Implement some methods of TR_J9SharedCacheServerVM

dmitry-ten:
Methods implemented:
```
getSuperClass
getResolvedMethodForNameAndSignature
isPrimitiveArray
isReferenceArray
getClassClassPointer
getInlinedCallSiteMethod
isUnloadAssumptionRequired
classHasBeenExtended.
```

### DIFF
--- a/runtime/compiler/env/VMJ9Server.cpp
+++ b/runtime/compiler/env/VMJ9Server.cpp
@@ -33,6 +33,26 @@
 #include "env/PersistentCHTable.hpp"
 #include "exceptions/AOTFailure.hpp"
 
+void
+TR_J9ServerVM::getResolvedMethodsAndMethods(TR_Memory *trMemory, TR_OpaqueClassBlock *classPointer, List<TR_ResolvedMethod> *resolvedMethodsInClass, J9Method **methods, uint32_t *numMethods)
+   {
+   JITaaS::J9ServerStream *stream = _compInfoPT->getMethodBeingCompiled()->_stream;
+   stream->write(JITaaS::J9ServerMessageType::VM_getResolvedMethodsAndMirror, classPointer);
+   auto recv = stream->read<J9Method *, std::vector<TR_ResolvedJ9JITaaSServerMethodInfo>>();
+   auto methodsInClass = std::get<0>(recv);
+   auto &methodsInfo = std::get<1>(recv);
+   if (methods)
+      *methods = methodsInClass;
+   if (numMethods)
+      *numMethods = methodsInfo.size();
+   for (int i = 0; i < methodsInfo.size(); ++i)
+      {
+      // create resolved methods, using information from mirrors
+      auto resolvedMethod = new (trMemory->trHeapMemory()) TR_ResolvedJ9JITaaSServerMethod((TR_OpaqueMethodBlock *) &(methodsInClass[i]), this, trMemory, methodsInfo[i], 0);
+      resolvedMethodsInClass->add(resolvedMethod);
+      }
+   }
+
 bool
 TR_J9ServerVM::isClassLibraryMethod(TR_OpaqueMethodBlock *method, bool vettedForAOT)
    {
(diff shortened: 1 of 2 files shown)


dmitry-ten:
@mpirvu fyi

runtime/compiler/env/VMJ9Server.cpp
```diff
@@ -1417,3 +1429,188 @@ TR_J9ServerVM::dereferenceStaticFinalAddress(void *staticAddress, TR::DataType a
...
+         }
+      else
+         {
+         TR_ASSERT(false, "Class 0x%p should already be validated\n", classPointer);
+         comp->failCompilation<J9::AOTSymbolValidationManagerFailure>("Failed to validate in getResolvedMethods");
+         }
+      }
+   else
+      {
+      validated = ((TR_ResolvedRelocatableJ9Method *) comp->getCurrentMethod())->validateArbitraryClass(comp, (J9Class *) classPointer);
```
mpirvu: Here we should use the TR_ResolvedRelocatbaleJ9ServerMethod that knows how to talk to the client.

dmitry-ten:
Rebased and addressed comments.

dmitry-ten:
Rebased again

runtime/compiler/env/VMJ9Server.cpp
```diff
@@ -1542,3 +1554,166 @@ TR_J9SharedCacheServerVM::getInstanceFieldOffset(TR_OpaqueClassBlock * classPoin
...
+   {
+   TR::Compilation* comp = _compInfo->getCompInfoForCompOnAppThread() ? _compInfo->getCompInfoForCompOnAppThread()->getCompilation() : _compInfoPT->getCompilation();
+   TR_ASSERT(comp, "Should be called only within a compilation");
+
+   bool validated = false;
+   TR_ResolvedMethod *resolvedMethod = TR_J9ServerVM::getResolvedMethodForNameAndSignature(trMemory, classPointer, methodName, signature);
+
+   if (comp->getOption(TR_UseSymbolValidationManager))
+      {
+      TR_OpaqueMethodBlock *method = (TR_OpaqueMethodBlock *)((TR_ResolvedJ9Method *)resolvedMethod)->ramMethod();
```
mpirvu: The cast should be to a `TR_ResolvedJ9JITaaSServerMethod`

runtime/compiler/env/VMJ9Server.cpp
```diff
@@ -1542,3 +1554,166 @@ TR_J9SharedCacheServerVM::getInstanceFieldOffset(TR_OpaqueClassBlock * classPoin
...
+      return isRefArray;
+   else
+      return false;
+   }
+
+TR_OpaqueClassBlock *
+TR_J9SharedCacheServerVM::getClassClassPointer(TR_OpaqueClassBlock *objectClassPointer)
+   {
+   TR_OpaqueClassBlock *ccPointer = TR_J9ServerVM::getClassClassPointer(objectClassPointer);
+   TR::Compilation *comp = TR::comp();
```
mpirvu: We should get comp from `_compInfoPT->getCompilation()` as it is cheaper. Then add the assert that the comp object must exist like in all other cases. It must exist because we are at the server where all actions are done within a compilation.

runtime/compiler/env/VMJ9Server.cpp
```diff
@@ -1542,3 +1554,166 @@ TR_J9SharedCacheServerVM::getInstanceFieldOffset(TR_OpaqueClassBlock * classPoin
...
+      if (!comp->getSymbolValidationManager()->addClassClassRecord(ccPointer, objectClassPointer))
+         ccPointer = NULL;
+      }
+   return ccPointer;
+   }
+
+TR_OpaqueMethodBlock *
+TR_J9SharedCacheServerVM::getInlinedCallSiteMethod(TR_InlinedCallSite *ics)
+   {
+   return (TR_OpaqueMethodBlock *)((TR_AOTMethodInfo *)(ics->_vmMethodInfo))->resolvedMethod->getPersistentIdentifier();
```
mpirvu: Will this work in JITaaS server? We get a `resolvedMethod` and try to fish the persistentIdentifier from it. Should this be cast to a TR_ResolvedJ9JITaaSServerMethod?

runtime/compiler/env/VMJ9Server.cpp
```diff
@@ -1542,3 +1554,166 @@ TR_J9SharedCacheServerVM::getInstanceFieldOffset(TR_OpaqueClassBlock * classPoin
 
    return ~0;
    }
+
+TR_OpaqueClassBlock *
+TR_J9SharedCacheServerVM::getSuperClass(TR_OpaqueClassBlock * classPointer)
+   {
+   TR::Compilation* comp = _compInfo->getCompInfoForCompOnAppThread() ? _compInfo->getCompInfoForCompOnAppThread()->getCompilation() : _compInfoPT->getCompilation();
```
mpirvu: The JITaaS server does not support compilation on application thread (there are no application threads at the server). Thus, we can simplify the code to: `TR::Compilation* comp = _compInfoPT->getCompilation()`
Similar observation for other methods below.

runtime/compiler/env/VMJ9Server.cpp
dmitry-ten: Wouldn't this call a JITaaS version of the method through polymorphism, as long as resolved method was instantiated as`TR_ResolvedJ9JITaaSServerMethod`?

mpirvu: You are correct

PR #4506, opened 2019-01-29
Make SVM ArrayClassFromComponentClass bidirectional

jdmpapin:
Two records `ArrayClassFromComponentClassRecord(array, component)` and `ComponentClassFromArrayClassRecord(component, array)` represent the exact same relationship between `array` and `component`, so it is never necessary to have both for the same `array`, `component` pair. This PR unifies the two by allowing `ArrayClassFromComponentClassRecord` to populate the symbol table in either direction. Now any redundancies of the kind described above will manifest as duplicated records, which will be naturally deduplicated.

In a second commit, `TR_ValidateComponentClassFromArrayClass` is removed because it is now redundant.


dsouzai:
fyi @r30shah @aviansie-ben

runtime/compiler/runtime/SymbolValidationManager.cpp
```diff
@@ -81,6 +79,12 @@ TR::SymbolValidationManager::defineGuaranteedID(void *symbol, TR::SymbolType typ
    _seenSymbolsSet.insert(symbol);
    }
 
+bool
+TR::SymbolValidationManager::isDefinedID(uint16_t id)
```
dsouzai: Should some of the existing `isAlreadyValidated` / `tryGetIDFromSymbol` be replaced with this (faster) query?

jdmpapin: `isAlreadyValidated` takes a pointer in order to check during compilation that we've already established how loads will find the corresponding pointer later. OTOH `isDefinedID` takes an ID and tells us whether the value for that ID has been determined at load time

dsouzai:
Jenkins test sanity xlinux,win,plinux,zlinux jdk8,jdk11

jdmpapin:
All checks have passed except for Travis, which failed almost immediately due to an infrastructure problem
```
Resolving api.adoptopenjdk.net (api.adoptopenjdk.net)... 104.17.15.123, 104.17.14.123, 104.17.12.123, ...
Connecting to api.adoptopenjdk.net (api.adoptopenjdk.net)|104.17.15.123|:443... connected.
HTTP request sent, awaiting response... 400 Bad Request
2019-01-30 22:22:31 ERROR 400: Bad Request.
```

Issue #4385, opened 2019-01-22
Enable Unsafe illegal write notification


liqunl:
Notification is turned off in #4383. Current implementation of Unsafe illegal write notification has caused a significant regression on start up. The initial investigation shows that splitting the block and creating temps seem to have large impact on compilation time. More study is needed to understand the regression. We also need to come up with a better design to enable the notification without losing the start up performance.

PR #4514, opened 2019-01-30
Report final field modification by calling runtime helper

liqunl:
Keeping the Unsafe call as the slow path is suboptimal compared to using
the runtime helper.
- Unsafe call consumes more arguments, which means more registers.
- Unsafe call has more use and def aliases and not transparent to the
  optimizer.
- For Unsafe call that returns a result, a temp has to be created to
  store result from the fast path and from the slow path.

Therefore, report the ilegal modification through the runtime helper
jitReportFinalFieldModified.

Also Enable the illegal write notification in this PR.

closes: #4385

### DIFF
--- a/runtime/compiler/optimizer/InlinerTempForJ9.cpp
+++ b/runtime/compiler/optimizer/InlinerTempForJ9.cpp
@@ -1061,9 +1061,6 @@ Unsafe.getShort.
 bool
 TR_J9InlinerPolicy::createUnsafePutWithOffset(TR::ResolvedMethodSymbol *calleeSymbol, TR::ResolvedMethodSymbol *callerSymbol, TR::TreeTop * callNodeTreeTop, TR::Node * unsafeCall, TR::DataType type, bool isVolatile, bool needNullCheck, bool isOrdered)
    {
-   if (callNodeTreeTop->getEnclosingBlock()->isCold())
-      return false;
-
    if (isVolatile && type == TR::Int64 && TR::Compiler->target.is32Bit() && !comp()->cg()->getSupportsInlinedAtomicLongVolatiles())
       return false;
    TR_ASSERT(TR::Compiler->cls.classesOnHeap(), "Unsafe inlining code assumes classes are on heap\n");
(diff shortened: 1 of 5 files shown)


liqunl:
Depends on https://github.com/eclipse/omr/pull/3523 and https://github.com/eclipse/omr/pull/3528.
WIP this PR, will add a commit to enable notification once perf evaluation is done.

liqunl:
Performance comparison between proposed change and baseline. Positive number means better than baseline, negative number means worse than baseline.

The performance difference between the proposed change and the baseline is within the noise margin for the following benchmarks.

| Benchmark | Relative perf| Standard deviation|
| daytrader3 | +0.02% | 0.38% |
| AcmeAir | -0.32% | 0.8% |
| ILOG 300RuleSet | +0.11% | 3.65% |
| ILOG 5Path| -0.15% | 1.43% |
| Pagerank  | -0.45%  | 1.34% |
| Sort | +0.01% | 2.07% |
| Terasort | +0.08% | 1.15%
|Wordcount|+0.4%|1.55%|

liqunl:

andrewcraik:
Have you looked at any startup metrics to see what impact the compiel-time has on them specifically? I suspect it should be fine since HiBench is ok but an explicit measure would be good. Actual changes look ok to me.

liqunl:
@andrewcraik Yes, looked at liberty start up sufpdt7, there is no regression. 

| Benchmark | Relative perf | Standard deviation|
| sufpdt7 | +0.04% | 1.4% |

andrewcraik:

andrewcraik:
@mpirvu any object to turning this on again? @liqunl has fixed the regressions and the data looks good. If any perf issues surface again we'll have a look for sure, but I think this should be good to go now.

mpirvu:
If start-up regressions have been eliminated I am ok with the change.

liqunl:
@mpirvu Yes, the start up is verified on sufpdt7, a liberty DT7 benchmark.

PR #4521, opened 2019-01-30
Add options to print a javacore dump to STDOUT and STDERR

### LINKED ISSUE #2837: Provide option to print javacore output to stdout/stderr instead of to a file
When sent a SIGQUIT, HotSpot prints a thread dump on stdout. In response to the same signal, OpenJ9 writes a javacore dump to a file, and prints the filename and dump event details on stderr.

Writing the data to a file is a better approach in most scenarios, but under certain circumstances it would be preferable to write the output to stdout/stderr instead, so it would be useful to have to the option to do that.

AidanHa:
Add the option -Xdump:java:file=/STDOUT/ to print a javacore dump to STDOUT
Add the option -Xdump:java:file=/STDERR/ to print a javacore dump to STDERR

Steps:
-Change the filedescriptor to 1 to print to STDOUT if the inputed filename is "/STDOUT"
-Change the filedescriptor to 2 to print to STDERR if the inputed filename is "/STDERR"

Doc issue https://github.com/eclipse/openj9-docs/issues/204

Fixes #2837

### DIFF
--- a/runtime/rasdump/TextFileStream.cpp
+++ b/runtime/rasdump/TextFileStream.cpp
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2003, 2014 IBM Corp. and others
+ * Copyright (c) 2003, 2019 IBM Corp. and others
  *
  * This program and the accompanying materials are made available under
  * the terms of the Eclipse Public License 2.0 which accompanies this
@@ -51,7 +51,11 @@ void
 TextFileStream::open(const char* fileName, bool cacheWrites)
 {
 	PORT_ACCESS_FROM_PORT(_PortLibrary);
-	if (fileName[0] != '-' ) {
+	if (NULL != strstr(fileName, "/STDOUT/")) {
+		_FileHandle = 1; 
+	} else if (NULL != strstr(fileName, "/STDERR/")) {
+		_FileHandle = 2;
+	} else if (fileName[0] != '-' ) {
 		_FileHandle = j9file_open(fileName, EsOpenWrite | EsOpenCreate | EsOpenTruncate | EsOpenCreateNoTag, 0666);
 	}
 	if(!cacheWrites) {
--- a/runtime/rasdump/dmpagent.c
+++ b/runtime/rasdump/dmpagent.c
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 1991, 2018 IBM Corp. and others
+ * Copyright (c) 1991, 2019 IBM Corp. and others
  *
  * This program and the accompanying materials are made available under
  * the terms of the Eclipse Public License 2.0 which accompanies this
@@ -886,11 +886,13 @@ static omr_error_t
 doJavaDump(J9RASdumpAgent *agent, char *label, J9RASdumpContext *context)
 {
 	J9JavaVM *vm = context->javaVM;
-
-	if (makePath(vm, label) == OMR_ERROR_INTERNAL) {
-		/* Nowhere available to write the dump, we are done, makePath() will have issued error message */
-		return OMR_ERROR_INTERNAL;
+	if ((0 != strcmp(label, "/STDOUT/")) && (0 != strcmp(label, "/STDERR/"))) {
+		if (makePath(vm, label) == OMR_ERROR_INTERNAL) {
+			/* Nowhere available to write the dump, we are done, makePath() will have issued error message */
+			return OMR_ERROR_INTERNAL;
+		}
 	}
+	
 	runJavadump(label, context, agent);
 
 	return OMR_ERROR_NONE;
--- a/test/functional/cmdLineTests/cmdLineTest_J9tests/j9tests.xml
+++ b/test/functional/cmdLineTests/cmdLineTest_J9tests/j9tests.xml
@@ -1,0 +1,48 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no" ?>
+
+<!--
+Copyright (c) 2019, 2019 IBM Corp. and others
+
+This program and the accompanying materials are made available under
+the terms of the Eclipse Public License 2.0 which accompanies this
+distribution and is available at https://www.eclipse.org/legal/epl-2.0/
+or the Apache License, Version 2.0 which accompanies this distribution and
+is available at https://www.apache.org/licenses/LICENSE-2.0.
+
+This Source Code may also be made available under the following
+Secondary Licenses when the conditions for such availability set
+forth in the Eclipse Public License, v. 2.0 are satisfied: GNU
+General Public License, version 2 with the GNU Classpath
+Exception [1] and GNU General Public License, version 2 with the
+OpenJDK Assembly Exception [2].
+
+[1] https://www.gnu.org/software/classpath/license.html
+[2] http://openjdk.java.net/legal/assembly-exception.html
+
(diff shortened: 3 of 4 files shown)


runtime/rasdump/dmpagent.c
```diff
@@ -886,8 +886,7 @@ static omr_error_t
 doJavaDump(J9RASdumpAgent *agent, char *label, J9RASdumpContext *context)
 {
 	J9JavaVM *vm = context->javaVM;
-
-	if (makePath(vm, label) == OMR_ERROR_INTERNAL) {
+	if (makePath(vm, label) == OMR_ERROR_INTERNAL && 0!= strcmp(label, "/STDOUT/") && 0!= strcmp(label, "/STDERR/")) {
```
pshipton: coding style. Adding brackets and spaces, this should be
`if ((makePath(vm, label) == OMR_ERROR_INTERNAL) && (0 != strcmp(label, "/STDOUT/")) && (0 != strcmp(label, "/STDERR/"))) {`

pshipton: Seems makePath() shouldn't be called at all if the label is `/STDOUT/` or `/STDERR/`. Are these names magic? Won't makePath() have side affects and create a directory `/STDOUT` or `/STDERR`?

pshipton:
fyi https://github.com/eclipse/openj9/issues/2837#issuecomment-458884303

runtime/rasdump/dmpagent.c
AidanHa: The character `/` is an illegal character for filenames and/or directories. So makePath with the label `/STDOUT/` shouldn't be possible. Regardless, we shouldn't be making a directory, if we want to write to STDOUT/ERR.

AidanHa:
Documentation issue at: https://github.com/eclipse/openj9-docs/issues/204

runtime/rasdump/TextFileStream.cpp
```diff
@@ -51,7 +52,11 @@ void
 TextFileStream::open(const char* fileName, bool cacheWrites)
 {
 	PORT_ACCESS_FROM_PORT(_PortLibrary);
-	if (fileName[0] != '-' ) {
+	if (NULL != strstr(fileName, "/STDOUT/")) {
+		_FileHandle = 1;
```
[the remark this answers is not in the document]
AidanHa: image
Tested this locally on my Mac, and it seems to work (not really sure how to prove it works besides building it). Currently waiting on the windows tests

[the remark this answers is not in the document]
pshipton: What I'm looking for is an automated test added as part of this PR so we can run it on all platforms before merging. In particular z/OS might be different.

AidanHa:
@pshipton hopefully I understood you correctly this time. Could you give this automated test a review so we can try launching the PR builds?

test/functional/cmdLineTests/cmdLineTest_J9tests/j9tests.xml
```diff
@@ -0,0 +1,48 @@
...
+SPDX-License-Identifier: EPL-2.0 OR Apache-2.0 OR GPL-2.0 WITH Classpath-exception-2.0 OR LicenseRef-GPL-2.0 WITH Assembly-exception
+-->
+
+<!DOCTYPE suite SYSTEM "cmdlinetester.dtd">
+
+<suite id="J9 Command-Line Option Tests" timeout="180">
+    
+    <test id="Verify Generate a javacore to STDOUT">
+        <command>$EXE$ -Xdump:java:events=vmstart,file=/STDOUT/</command>
+        <output type="success" caseSensitive="yes" regex="yes">TITLE subcomponent dump routine</output>
```
pshipton: This and the next can be `regex="no"`

pshipton:
> Could you give this automated test a review so we can try launching the PR builds?

Seems reasonable. Have you tried running the test via a grinder on the internal machines? You should be able to try all the platforms that way first. Running extended.functional PR testing on all platforms consumes a lot of machine time.

pshipton:
@AidanHa I'm not sure from your wording, note with a grinder you can run just your new test suite and don't need to run the entire extended test suite.

llxia:
We do need an SDK with Aidan's changes to run Grinder. We should be able to use the SDK from his personal build.

AidanHa:
> @AidanHa I'm not sure from your wording, note with a grinder you can run just your new test suite and don't need to run the entire extended test suite.

Sorry about that. I took your advice and used grinders to help run the new tests. I've sent you the internal links to each of the tests.

pshipton:
@llxia are you good with the test update? The internal testing looks good, I think this is ready to be merged.

llxia:
yes, the test changes look good to me.

test/functional/cmdLineTests/cmdLineTest_J9tests/playlist.xml
```diff
@@ -62,4 +62,21 @@
...
+	-config $(Q)$(TEST_RESROOT)$(D)j9tests.xml$(Q) \
+	-xids all,$(PLATFORM) -plats all,$(PLATFORM) -xlist $(Q)$(TEST_RESROOT)$(D)j9tests_exclude.xml$(Q) \
+	-explainExcludes -nonZeroExitWhenError; \
+	$(TEST_STATUS)</command>
+		<levels>
+			<level>extended</level>
+		</levels>
+		<groups>
+			<group>functional</group>
+		</groups>
```
llxia: ah, sorry, one more thing, please add following as this test only applies to ibm and openj9 java impl

```
<impls>
	<impl>ibm</impl>
	<impl>openj9</impl>
</impls>
```

AidanHa: @llxia  this has been done! I've squashed all of the commits for convenience sake. Could you check if I implemented your change correctly?

pshipton:
jenkins test all zlinux jdk8

pshipton:
jenkins compile win jdk8

runtime/rasdump/TextFileStream.cpp
```diff
@@ -51,7 +51,11 @@ void
 TextFileStream::open(const char* fileName, bool cacheWrites)
 {
 	PORT_ACCESS_FROM_PORT(_PortLibrary);
-	if (fileName[0] != '-' ) {
+	if (NULL != strstr(fileName, "/STDOUT/")) {
+		_FileHandle = 1; 
+	} else if (NULL != strstr(fileName, "/STDERR/")) {
+		_FileHandle = 2;
```
keithc-ca: * we should be using the constants `OMRPORT_TTY_OUT` and `OMRPORT_TTY_ERR` which also avoids problems on platforms where `FD_BIAS != 0`

* we should not close `_FileHandle` for stdout or stderr (because we did not open it)

AidanHa: *  Where exactly are`OMRPORT_TTY_OUT` and `OMRPORT_TTY_ERR` defined? As well @pshipton, is this change necessary if all of the grinder instances passed?

keithc-ca: Those macros are defined in `omr/include_core/omrport.h`.

A fix to not close stdout or stderr is necessary.

pshipton: Yes, the change to use OMRPORT_TTY_OUT and OMRPORT_TTY_ERR should also be done, in case something changes in the future.

pshipton: It would also be a good idea to create #define's for `/STDOUT/` and `/STDERR/` and use the defines instead of the strings.

PR #4525, opened 2019-01-31
Replace enter/exit tracing method with isMethodTracingEnabled

a7ehuo:
In practice we never enable methodEnterTracing without methodExitTracing.
These two methods are replaced with isMethodTracingEnabled().

### DIFF
--- a/runtime/compiler/env/VMJ9.cpp
+++ b/runtime/compiler/env/VMJ9.cpp
@@ -391,7 +391,7 @@ void releaseVMaccessIfNeeded(J9VMThread *vmThread, bool haveAcquiredVMAccess)
 bool
 TR_J9VMBase::isAnyMethodTracingEnabled(TR_OpaqueMethodBlock *method)
    {
-   return isMethodEnterTracingEnabled(method) || isMethodExitTracingEnabled(method);
+   return isMethodTracingEnabled(method);
    }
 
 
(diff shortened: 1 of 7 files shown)


a7ehuo:
@mpirvu Plan to combine stage 3 and 4 together: replacing and removing enter/exit methods in this one change in `openJ9/master`.

mpirvu:
What about the omr layer? For instance the inliner in omr has this piece of code
```
  if (comp()->getOption(TR_EnableJProfiling) ||
       (firstBlock->getPredecessors().size() > 1) ||
       firstBlock->hasExceptionSuccessors() ||
       comp()->fe()->isMethodEnterTracingEnabled(calleeResolvedMethod->getPersistentIdentifier()) ||
       TR::Compiler->vm.canMethodEnterEventBeHooked(comp()))
      {
      int32_t freq = firstBlock->getFrequency();
      firstBlock = _calleeSymbol->prependEmptyFirstBlock();
      firstBlock->setFrequency(freq);
      }
```
If you delete the corresponding implementation from openj9 the code will use the default implementation which will always return false;

a7ehuo:
Yeah, I was debating about this one place in omr where it calls enter/exit method tracing. As you mentioned `TR_FrontEnd::isMethodEnterTracingEnabled` will return false after this change. I assumed it'd be low risk though. But I agree, the better practice is to not combine replacement and removal in one change.

a7ehuo:
@mpirvu Please help check the latest update. Thank you.

mpirvu:
Jenkins test sanity all jdk8

PR #4533, opened 2019-01-31
Initialize jdk.internal.reflect.langReflectAccess to avoid NPE

JasonFengJ9:
Initialize `jdk.internal.reflect.langReflectAccess` to avoid `NPE`

`Modifier.isPublic(Modifier.PUBLIC)` is going to initialize the field `jdk.internal.reflect.langReflectAccess` right away to avoid `NPE` later such as https://github.com/ibmruntimes/openj9-openjdk-jdk11/pull/113#issuecomment-459417810.
This is a workaround of an `OpenJDK` issue in which `jdk.internal.reflect.ReflectionFactory` refers the field `langReflectAccess` directly instead of via `langReflectAccess()`.

Manually verified that the `pConfig` still compile.

Note: this change is not required by `Java 8` since the `OpenJDK` issue only presents in later versions.

Related to https://github.com/ibmruntimes/openj9-openjdk-jdk11/pull/113

Reviewer: @pshipton 

### DIFF
--- a/jcl/src/java.base/share/classes/java/lang/ClassLoader.java
+++ b/jcl/src/java.base/share/classes/java/lang/ClassLoader.java
@@ -28,6 +28,7 @@ import java.lang.reflect.*;
 import java.security.cert.Certificate;
 
 /*[IF Sidecar19-SE]
+import java.lang.reflect.Modifier;
 import java.util.Spliterator;
 import java.util.Spliterators;
 import java.util.stream.StreamSupport;
@@ -46,7 +47,7 @@ import sun.reflect.CallerSensitive;
 /*[ENDIF]*/
 
 /*******************************************************************************
- * Copyright (c) 1998, 2018 IBM Corp. and others
+ * Copyright (c) 1998, 2019 IBM Corp. and others
  *
  * This program and the accompanying materials are made available under
  * the terms of the Eclipse Public License 2.0 which accompanies this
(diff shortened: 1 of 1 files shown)


pshipton:
jenkins test sanity zlinux jdk11 depends ibmruntimes/openj9-openjdk-jdk11#113

JasonFengJ9:
Looking at the build failure. It appears the initialization is required even earlier.

pshipton:
@JasonFengJ9 the build failed https://ci.eclipse.org/openj9/job/PullRequest-Sanity-JDK11-linux_390-64_cmprssptrs-OpenJ9/175/

JasonFengJ9:
Moved `Modifier.isPublic(Modifier.PUBLIC)` before `jdk.internal.module.ModuleBootstrap.boot()`, and verified that `functional/Java9andUp` (failed in PR build) compiles at `xa64` platform.

@pshipton please review and launch another PR build.

pshipton:
jenkins test sanity zlinux jdk11 depends ibmruntimes/openj9-openjdk-jdk11#113

jcl/src/java.base/share/classes/java/lang/ClassLoader.java
```diff
@@ -259,6 +260,12 @@ static final void initializeClassLoaders() {
 		
 		/*[IF Sidecar19-SE]*/
 		jdk.internal.misc.VM.initLevel(1);
+		/* 
+		 * Keep following code as first call of this method which ensures that the field jdk.internal.reflect.langReflectAccess
```
pshipton: I think the wording of this comment needs to be changed. It is no longer the first call of this method. The testing passed, so once that it done I will merged this and the extension changes.

pshipton:
PR build passed https://ci.eclipse.org/openj9/job/PullRequest-Sanity-JDK11-linux_390-64_cmprssptrs-OpenJ9/176/

jcl/src/java.base/share/classes/java/lang/ClassLoader.java
JasonFengJ9: Good catch! the comments are updated.

PR #4541, opened 2019-02-01
Add test for VarHandle MethodHandleInfo

pdbain-ibm:
Depends on 
https://github.com/eclipse/openj9/pull/4458

### DIFF
--- a/test/functional/Java11andUp/playlist.xml
+++ b/test/functional/Java11andUp/playlist.xml
@@ -252,4 +252,28 @@
 			<subset>11+</subset>
 		</subsets>
 	</test>
+	<test>
+		<testCaseName>JCL_TEST_Java-Lang-Invoke</testCaseName>
+		<command>$(JAVA_COMMAND) $(JVM_OPTIONS) \
+			-cp $(Q)$(RESOURCES_DIR)$(P)$(TESTNG)$(P)$(TEST_RESROOT)$(D)GeneralTest.jar$(Q) \
+			org.testng.TestNG -d $(REPORTDIR) $(Q)$(TEST_RESROOT)$(D)testng.xml$(Q) \
+			-testnames JCL_TEST_Java-Lang-Invoke \
+			-groups $(TEST_GROUP) \
+			-excludegroups $(DEFAULT_EXCLUDE); \
+			$(TEST_STATUS)
+		</command>
+		<levels>
+			<level>extended</level>
+		</levels>
+		<groups>
+			<group>functional</group>
+		</groups>
+		<impls>
+			<impl>openj9</impl>
+			<impl>ibm</impl>
+		</impls>
+		<subsets>
+			<subset>11+</subset>
+		</subsets>
+	</test>
 </playlist>
--- a/test/functional/Java11andUp/src/org/openj9/test/varhandle/TestVarHandleInfo.java
+++ b/test/functional/Java11andUp/src/org/openj9/test/varhandle/TestVarHandleInfo.java
@@ -1,0 +1,118 @@
+/*******************************************************************************
+ * Copyright (c) 2019, 2019 IBM Corp. and others
+ *
+ * This program and the accompanying materials are made available under
+ * the terms of the Eclipse Public License 2.0 which accompanies this
+ * distribution and is available at https://www.eclipse.org/legal/epl-2.0/
+ * or the Apache License, Version 2.0 which accompanies this distribution and
+ * is available at https://www.apache.org/licenses/LICENSE-2.0.
+ *
+ * This Source Code may also be made available under the following
+ * Secondary Licenses when the conditions for such availability set
+ * forth in the Eclipse Public License, v. 2.0 are satisfied: GNU
+ * General Public License, version 2 with the GNU Classpath
+ * Exception [1] and GNU General Public License, version 2 with the
+ * OpenJDK Assembly Exception [2].
+ *
+ * [1] https://www.gnu.org/software/classpath/license.html
+ * [2] http://openjdk.java.net/legal/assembly-exception.html
+ *
+ * SPDX-License-Identifier: EPL-2.0 OR Apache-2.0 OR GPL-2.0 WITH Classpath-exception-2.0 OR LicenseRef-GPL-2.0 WITH Assembly-exception
+ *******************************************************************************/
+
+package org.openj9.test.varhandle;
+
+import static org.testng.AssertJUnit.assertEquals;
+import java.lang.invoke.MethodHandle;
+import java.lang.invoke.MethodHandleInfo;
+import java.lang.invoke.MethodHandles;
+import java.lang.invoke.MethodHandles.Lookup;
+import java.lang.reflect.Modifier;
+import java.lang.invoke.MethodType;
+import java.lang.invoke.VarHandle;
+import java.util.Arrays;
+import java.util.HashMap;
+import java.util.List;
+import java.util.stream.Collectors;
+
+import org.testng.annotations.BeforeSuite;
+import org.testng.annotations.Test;
+import org.testng.log4testng.Logger;
+
+@Test(groups = { "level.extended" })
+public class TestVarHandleInfo {
+
+	protected static Logger logger = Logger.getLogger(TestVarHandleInfo.class);
+	private static final int MODIFIER_SYNTHETIC = 0x01000; /* see Table 5.5-a of the Java 8 VM spec */
+	/* the JVM adds the SYNTHETIC modifier automatically */
+	private static final int expectedModifiers = 
+			Modifier.PUBLIC + Modifier.FINAL + Modifier.NATIVE 
+			+ MODIFIER_SYNTHETIC;
+	private static final String EXPECTED_MODIFIERS_STRING = Modifier.toString(expectedModifiers);
+	private static final int expectedReferenceKind = MethodHandleInfo.REF_invokeVirtual;
+	
+	Lookup 	mylookup = MethodHandles.lookup();
+	Integer instanceField; /* the target of the varhandle */
+	List<String> accessModes;
+	HashMap<String, MethodType> infos;
+
+	@Test
+	public void testAccessModes() throws NoSuchFieldException, IllegalAccessException {
+		/* grab a VarHahnlde purely to check the accessModeType */
+		VarHandle vh = mylookup.findVarHandle(TestVarHandleInfo.class, "instanceField", Integer.class); //$NON-NLS-1$
+		for (String methName : accessModes) {
+			logger.debug(methName);
+			VarHandle.AccessMode am = VarHandle.AccessMode.valueFromMethodName(methName);
+			
+			MethodHandle mh = MethodHandles.varHandleExactInvoker(am, vh.accessModeType(am));
+			checkMethodHandleInfo(mylookup.revealDirect(mh), infos.get(methName), methName);
+			
+			mh = MethodHandles.varHandleInvoker(am, vh.accessModeType(am));
+			checkMethodHandleInfo(mylookup.revealDirect(mh), infos.get(methName), methName);
+		}
+	}
+
+	private static void checkMethodHandleInfo(MethodHandleInfo info, MethodType expectedMethodType,
+			String expectedMethodName) {
+		assertEquals("Incorrect declaring class", VarHandle.class, info.getDeclaringClass()); //$NON-NLS-1$
+
+		assertEquals("Incorrect methodType", expectedMethodType, info.getMethodType()); //$NON-NLS-1$
+
+		int actualModifiers = info.getModifiers();
+		assertEquals("Missing modifiers: expected \"" + EXPECTED_MODIFIERS_STRING //$NON-NLS-1$
+				+ "\" (" + Integer.toHexString(expectedModifiers) //$NON-NLS-1$
+				+ "), actual \"" + Modifier.toString(actualModifiers) //$NON-NLS-1$
+				+ "\" (" + Integer.toHexString(actualModifiers) + ")", //$NON-NLS-1$ //$NON-NLS-2$
+				expectedModifiers, actualModifiers);
+
+		assertEquals("Incorrect name ", expectedMethodName, info.getName()); //$NON-NLS-1$
+
+		assertEquals("Incorrect reference kind, expected " //$NON-NLS-1$
+				+ MethodHandleInfo.referenceKindToString(expectedReferenceKind), expectedReferenceKind,
+				info.getReferenceKind());
+	}
+
+	@BeforeSuite
+	public void makeInfos() {
+		infos = new HashMap<>();
+		accessModes = Arrays.asList(VarHandle.AccessMode.values())
+				.stream()
+				.map(m -> m.methodName())
+				.collect(Collectors.toList());
+
+		MethodType mtInt = MethodType.methodType(Integer.class, TestVarHandleInfo.class);
+		accessModes.stream().filter(m -> m.contains("get")).forEach(s -> infos.put(s, mtInt)); //$NON-NLS-1$
+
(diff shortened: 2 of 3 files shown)


test/functional/Java11andUp/src/org/openj9/test/varhandle/TestVarHandleInfo.java
```diff
@@ -0,0 +1,115 @@
...
+		assertEquals("Incorrect declaring class", VarHandle.class, info.getDeclaringClass()); //$NON-NLS-1$	
+
+		assertEquals("Incorrect methodType", expectedMethodType, info.getMethodType()); //$NON-NLS-1$
+		
+		int actualModifiers = info.getModifiers();
+		assertTrue("Missing modifiers: expected \""+EXPECTED_MODIFIERS_STRING  //$NON-NLS-1$
+				+ "\" (" + Integer.toHexString(expectedModifiers)  //$NON-NLS-1$
+		+ "), actual \"" + Modifier.toString(actualModifiers)  //$NON-NLS-1$
+		+ "\" ("+Integer.toHexString(actualModifiers)+")",  //$NON-NLS-1$ //$NON-NLS-2$
+		(expectedModifiers & actualModifiers) == expectedModifiers);
```
DanHeidinga: Should this be an `assertEquals` between the expectedModifiers and the actualModifiers?  Why the `&`?

test/functional/Java11andUp/src/org/openj9/test/varhandle/TestVarHandleInfo.java
```diff
@@ -0,0 +1,115 @@
...
+
+	@BeforeSuite
+	public void makeInfos() {
+		infos = new HashMap<>();
+		mylookup = MethodHandles.lookup();
+		accessModes = Arrays.asList(VarHandle.AccessMode.values())
+				.stream().map(m -> m.methodName()).collect(Collectors.toList());
+		
+		accessModes.stream().filter(m -> m.contains("get")) //$NON-NLS-1$
+		.forEach(s -> infos.put(s, MethodType.fromMethodDescriptorString("("+TEST_VAR_HANDLE_INFO +")" + INTEGER, null))); //$NON-NLS-1$ //$NON-NLS-2$
```
DanHeidinga: As the MethodType is constant for this loop, why not compute it once outside the forEach and reuse it as creating MTs is expensive?

```
MethodType getType = MethodType.fromMethodDescriptorString("("+TEST_VAR_HANDLE_INFO +")" + INTEGER, null);
accessModes.stream().filter(m -> m.contains("get")).forEach(s -> infos.put(s, getType));
```

I think the more imperative version of this code would be more readable and would occur in a single pass:
```java
for (AccessMode am : VarHandle.AccessMode.values()) {
  String m = am.methodName();
  if (m.contains("get") { 
      infos.put(m, getType); 
  } else if (m.contains("set")) {
     infos.put(m, setType);
  } else if ....
}
```
but won't insist on it being changed

test/functional/Java11andUp/src/org/openj9/test/varhandle/TestVarHandleInfo.java
```diff
@@ -0,0 +1,115 @@
...
+
+	protected static Logger logger = Logger.getLogger(TestVarHandleInfo.class);
+	private static final String TEST_VAR_HANDLE_INFO = "Lorg/openj9/test/varhandle/TestVarHandleInfo;"; //$NON-NLS-1$
+	private static final String INTEGER = "Ljava/lang/Integer;"; //$NON-NLS-1$
+	private static final int PUBLIC_FINAL_NATIVE = Modifier.PUBLIC + Modifier.FINAL + Modifier.NATIVE;
+	private static final int expectedModifiers = PUBLIC_FINAL_NATIVE;
+	private static final String EXPECTED_MODIFIERS_STRING = Modifier.toString(expectedModifiers);
+	private static final int expectedReferenceKind = MethodHandleInfo.REF_invokeVirtual;
+	
+	Lookup mylookup;
```
DanHeidinga: This can be initialized here once as it won't change - no need to delay to the BeforeSuite method

test/functional/Java11andUp/playlist.xml
```diff
@@ -252,4 +252,24 @@
...
+		</command>
+		<levels>
+			<level>extended</level>
+		</levels>
+		<groups>
+			<group>functional</group>
+		</groups>
+		<subsets>
+			<subset>11+</subset>
+		</subsets>
```
llxia: Could you add impls?

```
		<impls>
			<impl>openj9</impl>
			<impl>ibm</impl>
		</impls>
```

test/functional/Java11andUp/src/org/openj9/test/varhandle/TestVarHandleInfo.java
```diff
@@ -0,0 +1,115 @@
...
+	List<String> accessModes;
+	HashMap<String, MethodType> infos;
+
+	@Test
+	public void testAccessModes() throws NoSuchFieldException, IllegalAccessException {
+		VarHandle vh = mylookup.findVarHandle(TestVarHandleInfo.class, "instanceField", Integer.class); //$NON-NLS-1$
+		for (String methName: accessModes) {
+			logger.debug(methName);
+			VarHandle.AccessMode am = VarHandle.AccessMode.valueFromMethodName(methName);
+			MethodHandle mh = MethodHandles.varHandleExactInvoker(am,vh.accessModeType(am));
```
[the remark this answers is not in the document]
pdbain: Thanks.

test/functional/Java11andUp/src/org/openj9/test/varhandle/TestVarHandleInfo.java
```diff
@@ -0,0 +1,115 @@
...
+ * General Public License, version 2 with the GNU Classpath
+ * Exception [1] and GNU General Public License, version 2 with the
+ * OpenJDK Assembly Exception [2].
+ *
+ * [1] https://www.gnu.org/software/classpath/license.html
+ * [2] http://openjdk.java.net/legal/assembly-exception.html
+ *
+ * SPDX-License-Identifier: EPL-2.0 OR Apache-2.0 OR GPL-2.0 WITH Classpath-exception-2.0 OR LicenseRef-GPL-2.0 WITH Assembly-exception
+ *******************************************************************************/
+//package java.lang.invoke;
```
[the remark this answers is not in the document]
pdbain: Thank you.

test/functional/Java11andUp/src/org/openj9/test/varhandle/TestVarHandleInfo.java
```diff
@@ -0,0 +1,115 @@
...
+	private static final int expectedReferenceKind = MethodHandleInfo.REF_invokeVirtual;
+	
+	Lookup mylookup;
+	Integer instanceField; /* the target of the varhandle */
+	List<String> accessModes;
+	HashMap<String, MethodType> infos;
+
+	@Test
+	public void testAccessModes() throws NoSuchFieldException, IllegalAccessException {
+		VarHandle vh = mylookup.findVarHandle(TestVarHandleInfo.class, "instanceField", Integer.class); //$NON-NLS-1$
```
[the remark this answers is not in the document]
pdbain: Will do.

test/functional/Java11andUp/src/org/openj9/test/varhandle/TestVarHandleInfo.java
pdbain: > Should this be an `assertEquals` between the expectedModifiers and the actualModifiers?

Yes.  Will fix.

> Why the `&`?

The `actualModifiers` can have some extra (undocumented) bits set.

pdbain-ibm:
@DanHeidinga and @llxia 
Updated per your comments and re-tested.

In re efficiency, this test runs in less than 2 seconds, so I favoured clarity over speed.
Thanks for the input on functional vs. imperative: I will keep that in mind in future.

test/functional/Java11andUp/src/org/openj9/test/varhandle/TestVarHandleInfo.java
```diff
@@ -0,0 +1,118 @@
...
+				expectedReferenceKind, info.getReferenceKind());
+	}
+
+	@BeforeSuite
+	public void makeInfos() {
+		infos = new HashMap<>();
+		accessModes = Arrays.asList(VarHandle.AccessMode.values())
+				.stream().map(m -> m.methodName()).collect(Collectors.toList());
+
+		{
```
DanHeidinga: Can you use a unique name for each of the MTs rather than adding this extra scoping?  It will be more clear for readers what's happening

test/functional/Java11andUp/src/org/openj9/test/varhandle/TestVarHandleInfo.java
```diff
@@ -0,0 +1,118 @@
...
+		assertEquals("Incorrect declaring class", VarHandle.class, info.getDeclaringClass()); //$NON-NLS-1$	
+
+		assertEquals("Incorrect methodType", expectedMethodType, info.getMethodType()); //$NON-NLS-1$
+		
+		int actualModifiers = info.getModifiers();
+		assertEquals("Missing modifiers: expected \""+EXPECTED_MODIFIERS_STRING  //$NON-NLS-1$
+				+ "\" (" + Integer.toHexString(expectedModifiers)  //$NON-NLS-1$
+		+ "), actual \"" + Modifier.toString(actualModifiers)  //$NON-NLS-1$
+		+ "\" ("+Integer.toHexString(actualModifiers)+")",  //$NON-NLS-1$ //$NON-NLS-2$
+		expectedModifiers, (expectedModifiers & actualModifiers));
```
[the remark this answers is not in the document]
pdbain-ibm: The extra flag is `SYNTHETIC` (0x1000).  
The error is:

> expected "public final native" (111), actual "public final native" (1111)

I could mask just that one bit out if you prefer.

pdbain-ibm:
> Can you use a unique name for each of the MTs rather than adding this extra scoping? It will be more clear for readers what's happening

Done.

> nitpick - missing space between `am,vh`

Fixed.

test/functional/Java11andUp/src/org/openj9/test/varhandle/TestVarHandleInfo.java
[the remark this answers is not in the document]
DanHeidinga: Does the RI return the modifiers for `public final native synthetic` as well?  If so, the test should check for synthetic.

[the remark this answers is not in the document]
pdbain-ibm: > Does the RI return the modifiers for `public final native synthetic` as well? 

Yes. Updated per your request.

test/functional/Java11andUp/src/org/openj9/test/varhandle/TestVarHandleInfo.java
```diff
@@ -0,0 +1,115 @@
...
+import org.testng.annotations.Test;
+import org.testng.log4testng.Logger;
+
+@Test(groups = { "level.extended" })
+public class TestVarHandleInfo {
+
+	protected static Logger logger = Logger.getLogger(TestVarHandleInfo.class);
+	private static final String TEST_VAR_HANDLE_INFO = "Lorg/openj9/test/varhandle/TestVarHandleInfo;"; //$NON-NLS-1$
+	private static final String INTEGER = "Ljava/lang/Integer;"; //$NON-NLS-1$
+	private static final int MODIFIER_SYNTHETIC = 0x01000;  /* get this from dumping java/lang/reflect/Modifier */
```
DanHeidinga:
```
/* get this from dumping java/lang/reflect/Modifier */
```
It's also in the VM spec :)

pdbain-ibm: Thanks @DanHeidinga 
I need a newer copy of the spec -:).
Code updated with chapter and verse.

pdbain-ibm:
Updated per reviewers' comments.

pdbain-ibm:
@llxia or @DanHeidinga do you have any further comments?

test/functional/Java11andUp/src/org/openj9/test/varhandle/TestVarHandleInfo.java
DanHeidinga: I still see the `/* get this from dumping java/lang/reflect/Modifier */` comment

test/functional/Java11andUp/src/org/openj9/test/varhandle/TestVarHandleInfo.java
```diff
@@ -0,0 +1,115 @@
...
+
+	private static void checkMethodHandleInfo(MethodHandleInfo info, MethodType expectedMethodType, String expectedMethodName) {	
+		assertEquals("Incorrect declaring class", VarHandle.class, info.getDeclaringClass()); //$NON-NLS-1$	
+
+		assertEquals("Incorrect methodType", expectedMethodType, info.getMethodType()); //$NON-NLS-1$
+		
+		int actualModifiers = info.getModifiers();
+		assertEquals("Missing modifiers: expected \""+EXPECTED_MODIFIERS_STRING  //$NON-NLS-1$
+				+ "\" (" + Integer.toHexString(expectedModifiers)  //$NON-NLS-1$
+		+ "), actual \"" + Modifier.toString(actualModifiers)  //$NON-NLS-1$
```
DanHeidinga: The formatting seems off here.

test/functional/Java11andUp/src/org/openj9/test/varhandle/TestVarHandleInfo.java
```diff
@@ -0,0 +1,115 @@
...
+				expectedReferenceKind, info.getReferenceKind());
+	}
+
+	@BeforeSuite
+	public void makeInfos() {
+		infos = new HashMap<>();
+		accessModes = Arrays.asList(VarHandle.AccessMode.values())
+				.stream().map(m -> m.methodName()).collect(Collectors.toList());
+
+		MethodType mtInt = MethodType.fromMethodDescriptorString("("+TEST_VAR_HANDLE_INFO +")" + INTEGER, null); //$NON-NLS-1$ //$NON-NLS-2$
```
DanHeidinga:
```java
MethodType mtInt = MethodType.methodType(Integer.class, TestVarHandleInfo.class);
```
Which avoids the static strings and the conversions from String -> Class

test/functional/Java11andUp/src/org/openj9/test/varhandle/TestVarHandleInfo.java
```diff
@@ -0,0 +1,115 @@
...
+		accessModes.stream().filter(m -> m.contains("get")).forEach(s -> infos.put(s, mtInt)); //$NON-NLS-1$
+
+		MethodType mtIntVoid = MethodType.fromMethodDescriptorString("("+TEST_VAR_HANDLE_INFO + INTEGER + ")V", null); //$NON-NLS-1$ //$NON-NLS-2$
```
DanHeidinga:
```java
MethodType.methodType(void.class, TestVarHandleInfo.class, Integer.class);
```

test/functional/Java11andUp/src/org/openj9/test/varhandle/TestVarHandleInfo.java
pdbain-ibm: Replaced with reference to VM spec.

pdbain-ibm:
Updated per reviewer's comments.  I also trimmed out some redundant code and cleaned up formatting. Tested on MacOS and pushed.

DanHeidinga:

DanHeidinga:
Jenkins test sanity osx jdk12

pdbain-ibm:
@DanHeidinga could you restart the sanity build? It appears there was a machine problem.

DanHeidinga:
Jenkins test sanity osx jdk12

PR #4543, opened 2019-02-01
Enable guarded devirtualization for JITaaS

dmitry-ten:
Now that we can make direct calls properly, devirtualization
can be safely enabled.


dmitry-ten:
@ymanton could you review please?

dmitry-ten:
No, but I could try to measure it.

dmitry-ten:
I measured a throughput improvement of 1%.

ymanton:
Merged. Good catch, thanks.

PR #4546, opened 2019-02-01
Use evacuate top to check if CS is active on X86

0dvictor:
Evacuate Top is null when CS is inactive, and non-null otherwise;
therefore, J9_PRIVATE_FLAGS_CONCURRENT_SCAVENGER_ACTIVE can be
deprecated.


andrewcraik:

andrewcraik:
Jenkins test sanity win jdk11

PR #4555, opened 2019-02-02
Fix CryptoTest to accept "native" or "Native" in output

pshipton:
[ci skip]

Signed-off-by: Peter Shipton <Peter_Shipton@ca.ibm.com>

PR #4547, opened 2019-02-01
Remove obsolete OpenSSL test

llxia:
[ci skip]

Signed-off-by: lanxia <lan_xia@ca.ibm.com>

PR #4558, opened 2019-02-03
Update openssl testing for the 0.12.1 release

pshipton:
Required so openssl (CryptoTest) testing passes with https://github.com/ibmruntimes/openj9-openjdk-jdk11/pull/122 and https://github.com/ibmruntimes/openj9-openjdk-jdk8/pull/263

Corresponds to #4551, #4547 and #4555


pshipton:

Issue #4530, opened 2019-01-31
Crash in NativeCrypto.DigestUpdate (openssl native)


pshipton:
```
3XMTHREADINFO3           Java callstack:
4XESTACKTRACE                at jdk/crypto/jniprovider/NativeCrypto.DigestUpdate(Native Method)
4XESTACKTRACE                at sun/security/provider/NativeDigest.engineUpdate(NativeDigest.java:103)
4XESTACKTRACE                at java/security/MessageDigest$Delegate.engineUpdate(MessageDigest.java:613)
4XESTACKTRACE                at java/security/MessageDigest.update(MessageDigest.java:355)
4XESTACKTRACE                at com/sun/crypto/provider/HmacCore.engineUpdate(HmacCore.java:180)
4XESTACKTRACE                at javax/crypto/Mac.update(Mac.java:505)
4XESTACKTRACE                at sun/security/ssl/Authenticator$MacImpl.compute(Authenticator.java:510)
4XESTACKTRACE                at sun/security/ssl/Authenticator$TLS10Mac.compute(Authenticator.java:581)
4XESTACKTRACE                at sun/security/ssl/Authenticator$MAC.compute(Authenticator.java:444)
4XESTACKTRACE                at sun/security/ssl/SSLCipher.addMac(SSLCipher.java:2089)
4XESTACKTRACE                at sun/security/ssl/SSLCipher$T11BlockWriteCipherGenerator$BlockWriteCipher.encrypt(SSLCipher.java:1449)
4XESTACKTRACE                at sun/security/ssl/OutputRecord.t10Encrypt(OutputRecord.java:462)
4XESTACKTRACE                at sun/security/ssl/OutputRecord.encrypt(OutputRecord.java:394)
4XESTACKTRACE                at sun/security/ssl/SSLSocketOutputRecord.encodeAlert(SSLSocketOutputRecord.java:78)
5XESTACKTRACE                   (entered lock: sun/security/ssl/SSLSocketOutputRecord@0x00000000C0C0EDE8, entry count: 2)
4XESTACKTRACE                at sun/security/ssl/TransportContext.warning(TransportContext.java:243)
4XESTACKTRACE                at sun/security/ssl/SSLSocketImpl.duplexCloseOutput(SSLSocketImpl.java:553)
5XESTACKTRACE                   (entered lock: sun/security/ssl/SSLSocketOutputRecord@0x00000000C0C0EDE8, entry count: 1)
[...]

PR #4565, opened 2019-02-04
Add release notes for 0.12.1

SueChaplain:
New topic in the 0.12 release folder to
cover the notable change for 0.12.1 and
to record it as an issue in the table.

### DIFF
--- a/doc/release-notes/0.12/0.12.1.md
+++ b/doc/release-notes/0.12/0.12.1.md
@@ -22,19 +22,17 @@
 * Classpath-exception-2.0 OR LicenseRef-GPL-2.0 WITH Assembly-exception
 -->
 
-# Eclipse OpenJ9 version 0.12.0 release notes
+# Eclipse OpenJ9 version 0.12.1 release notes
 
-These release notes support the [Eclipse OpenJ9 0.12.0 release plan](https://projects.eclipse.org/projects/technology.openj9/releases/0.12.0/plan).
+These release notes support the [Eclipse OpenJ9 0.12.1 release plan](https://projects.eclipse.org/projects/technology.openj9/releases/0.12.1/plan).
 
 
 ## Binaries and supported environments
 
-OpenJ9 release 0.12.0 supports OpenJDK 8 and OpenJDK 11. Binaries are available at the AdoptOpenJDK project:
+OpenJ9 release 0.12.1 supports OpenJDK 8 and OpenJDK 11. Binaries are available at the AdoptOpenJDK project:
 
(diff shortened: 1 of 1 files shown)


doc/release-notes/0.12/0.12.1.md
```diff
@@ -0,0 +1,151 @@
+<!--
+* Copyright (c) 2017, 2019 IBM Corp. and others
```
pshipton: The start date should be 2019 for this new file.

doc/release-notes/0.12/0.12.1.md
```diff
@@ -0,0 +1,151 @@
...
+<th valign="bottom">Workaround</th>
+</tr>
+</thead>
+<tbody>
+
+<tr><td valign="top"><a href="https://github.com/eclipse/openj9/issues/4530">#4530</a></td>
+<td valign="top">Intermittent segmentation faults occur when using OpenSSL V1.1.x acceleration for the Digest algorithm</td>
+<td valign="top">All platforms</td>
+<td valign="top">You cannot use cryptographic acceleration for the Digest algorithm.</td>
+<td valign="top">Use an alternative algorithm. Currently, Digest is disabled by default.</td>
```
pshipton: Do we need #4530 listed twice?
Re `Use an alternative algorithm.` this is confusing. The user doesn't need to do anything since the openssl digest support is disabled by default.

pshipton: `You cannot use cryptographic acceleration for the Digest algorithm.` sounds like it may be a concerning problem, while the estimated the acceleration was only worth about ~5% perf and unlikely to be noticeable.

SueChaplain: >Do we need #4530 listed twice?

It is both a change and an issue (regression). I figured it deserved a mention in both places.

>Re Use an alternative algorithm. this is confusing. The user doesn't need to do anything since the openssl digest support is disabled by default.

I was guessing users might want cryptographic acceleration but I'll remove it

> You cannot use cryptographic acceleration for the Digest algorithm. sounds like it may be a concerning problem, while we estimated the acceleration was only worth about ~5% perf and unlikely to be noticeable.

Really? Wow. If noone was likely to be bothered about it, I guess it is a regression that can be ignored. In which case I'll remove it from the issues table completely.

pshipton:
The copyright wasn't fixed. Otherwise, ready for merge.

SueChaplain:
>The copyright wasn't fixed. Otherwise, ready for merge.

Fixed for this file and the 0.12.md release notes, which got merged unnoticed.

PR #4567, opened 2019-02-04
Fix TR_J9SharedCacheServerVM::stackWalkerMaySkipFrames

mpirvu:
Bring the implementation of TR_J9SharedCacheServerVM::stackWalkerMaySkipFrames
in-line with the new code from TR_J9SharedCacheVM::stackWalkerMaySkipFrames


mpirvu:
I only ran "java -version" for testing because the new code can only be tested when we have implemented enough methods to actually perform AOT compilations.

Issue #4208, opened 2019-01-08
Dockerfile for JDK11 installs gcc 7.4 by default


AdamBrousseau:
https://github.com/eclipse/openj9/blob/master/buildenv/docker/jdk11/x86_64/ubuntu16/Dockerfile

Default gcc7 on ubuntu16 is 7.4. Should we install gcc7.3 instead of default 7.4 from the package manager?

@pshipton

AdamBrousseau:
@samuel-rubin as a good first issue, please update the Dockerfile(s) in the buildenv/docker/jdk11 directories to install gcc 7.3 instead of the default one from the package manager.
We should probably also check the ub16/390 Dockerfile used for the buildfarm containers
https://github.com/eclipse/openj9/blob/master/buildenv/jenkins/docker-slaves/s390x/ubuntu16/Dockerfile

pshipton:
Using gcc 7.4 for s390/zlinux is fine. 7.4 is the desired level for zlinux, and it has already been verified for performance.

AdamBrousseau:
Confirmed the Container for ub16/390 already has 7.4
https://hub.docker.com/r/eclipse/openj9-jenkins-agent-s390x-ubuntu16

samuel-rubin:
The current repository for gcc-7 only holds onto the latest version which is currently 7.4.0. I also looked into Installing GCC 7.3 from the source but that took too long (over an hour). Something that we can do is copy the gcc binary from the AdoptOpenJDK jenkins like in #3864.

samuel-rubin:
I am going to copy to copy the 7.3 binaries (and links so they work) to the ub16 machines on x86 and ppc64le. On ub18 for x86 and ppc64le, I am thinking of using the ubuntu repo and restricting the version to 7.3. On s390 containers and x86 centos7 container, I'm thinking of keeping the same method, but requiring the version the be 7.4. If the package updates, the docker build will fail.

samuel-rubin:
Jenkins Docker files: 

|           | x86                   | s390x            | ppc64le               |
|-----------|-----------------------|------------------|-----------------------|
| Ubuntu16  | AdoptOpenJDK Binaries | Ubuntu toolchain | AdoptOpenJDK Binaries |
| Ubuntu18  | Ubuntu default        | Ubuntu toolchain | Ubuntu Default        |
| Centos6.9 | Centos devtoolset     | Not Available    | Not Available         |
[...]

PR #4568, opened 2019-02-04
Install gcc 7.3/7.4 from Adopt in docker files

samuel-rubin:
Dev Containers:
- Changes the way GCC is installed on x86, ppcle and s390x
  - GCC is now pulled from AdoptOpenJDK instead of the ubuntu toolchain
  - Added links to the default C libraries and include files
  - Links were added because GCC was not able to find them

Jenkins Containers:
- Specifies which version will be installed on the x86 build container
- Changes the way GCC is installed (same way as Dev Containers)

[skip ci]

Related #4208

### DIFF
--- a/buildenv/docker/jdk11/s390x/ubuntu16/Dockerfile
+++ b/buildenv/docker/jdk11/s390x/ubuntu16/Dockerfile
@@ -1,4 +1,4 @@
-# Copyright (c) 2018, 2018 IBM Corp. and others
+# Copyright (c) 2018, 2019 IBM Corp. and others
 #
 # This program and the accompanying materials are made available under
 # the terms of the Eclipse Public License 2.0 which accompanies this
@@ -29,7 +29,6 @@ RUN apt-get update \
   && apt-get install -qq -y --no-install-recommends \
     software-properties-common \
     python-software-properties \
-  && add-apt-repository ppa:ubuntu-toolchain-r/test \
   && apt-get update \
   && apt-get install -qq -y --no-install-recommends \
     autoconf \
@@ -38,8 +37,6 @@ RUN apt-get update \
     cmake \
     cpio \
     file \
-    g++-7 \
-    gcc-7 \
     git \
     git-core \
     libasound2-dev \
@@ -48,6 +45,7 @@ RUN apt-get update \
     libelf-dev \
     libfontconfig1-dev \
     libfreetype6-dev \
+    libmpc3 \
     libx11-dev \
     libxext-dev \
     libxrender-dev \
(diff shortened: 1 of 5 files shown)


samuel-rubin:
@AdamBrousseau

buildenv/docker/jdk11/s390x/ubuntu16/Dockerfile
```diff
@@ -39,7 +39,7 @@ RUN apt-get update \
     cpio \
     file \
     g++-7 \
-    gcc-7 \
+    gcc-7=7.4.0-1ubuntu1~16.04~ppa1 \
```
AdamBrousseau: Will 7.4 always be in the repo? If not I think we should pull the Adopt binary.

buildenv/jenkins/docker-slaves/s390x/ubuntu16/Dockerfile
```diff
@@ -52,7 +52,7 @@ RUN apt-get update \
     g++-4.8 \
     g++-7 \
     gcc-4.8 \
-    gcc-7 \
+    gcc-7=7.4.0-1ubuntu1~16.04~ppa1 \
```
AdamBrousseau: Same comment

buildenv/jenkins/docker-slaves/x86/centos6.9/Dockerfile
```diff
@@ -126,8 +126,8 @@ RUN cd /etc/yum.repos.d \
   && ln -s /opt/rh/devtoolset-2/root/usr/bin/gcc-4.8 /usr/bin/cc \
   && yum -y update \
   && yum -y install centos-release-scl \
-  && yum -y install devtoolset-7-gcc devtoolset-7-binutils \
-  && yum -y install devtoolset-7-gcc-c++ devtoolset-2-gcc-gfortran \
+  && yum -y install devtoolset-7-gcc-7.3* devtoolset-7-binutils \
```
AdamBrousseau: Similar comment

buildenv/docker/jdk11/s390x/ubuntu16/Dockerfile
samuel-rubin: I'm not sure. I don't see any other versions of gcc-7 in it, so I do not believe that It will stay.

buildenv/jenkins/docker-slaves/x86/centos6.9/Dockerfile
samuel-rubin: I saw multiple versions of gcc-7.3 and gcc-7.2 in that repo so I think that it will remain in that repo.

samuel-rubin: It is similar to what Adopt is doing

samuel-rubin:
Updated the s390x containers to have GCC installed from the AdoptOpenJDK binary

samuel-rubin:
I also added libmpc3 to the s390x containers because I got this message from gcc: 
> libmpc.so.3: cannot open shared object file: No such file or directory

AdamBrousseau:
Jenkins build container s390x ubuntu16

AdamBrousseau:
Jenkins build docker s390x ubuntu16

pshipton:
Please fix the commit and PR title to better describe this change, i.e. like `Install gcc 7.3/7.4 from Adopt in docker files`

samuel-rubin:
Ok. I changed the title.

pshipton:
@JasonFengJ9 or @samuel-rubin
the jdk12 docker files that Jason just added need the same changes.

pshipton:
and actually the jdk8 docker files do as well.

Issue #4193, opened 2019-01-08
Image files are not displayed properly on SwingSet2 demo program


takiguc:
Hello.

I used OpenJDK JDK11 + OpenJ9 for Linux amd64.
When I executed SwingSet2 demo, GIF images were not displayed properly.
And the situation was random, 4 GIF images were displayed on JInternalFrame Demo.
2 or 3 images were broken.

Version Information: RHEL 7.5
```
$ ~/OpenJ9/jdk-11.0.1+13/bin/java -version
openjdk version "11.0.1" 2018-10-16
OpenJDK Runtime Environment AdoptOpenJDK (build 11.0.1+13)
Eclipse OpenJ9 VM AdoptOpenJDK (build openj9-0.11.0, JRE 11 Linux amd64-64-Bit Compressed References 20181115_18 (JIT enabled, AOT enabled)
OpenJ9   - 090ff9dcd
OMR      - ea548a66
JCL      - d4455071ce based on jdk-11.0.1+13)
```
Test instruction:
1. Run following command
```
$ ~/OpenJ9/jdk-11.0.1+13/bin/java -jar ~/OpenJ9/jdk-11.0.1+13/demo/jfc/SwingSet2/SwingSet2.jar
```
2. Check Frame 0 - 4 windows. some of images are not displayed properly

This issue only happens on Linux amd64.
Linux ppc64le and s390x builds were OK.
It worked fine on following AdoptOpenJDK build.
```
$ ~/Hotspot/jdk-11.0.1+13/bin/java -version
openjdk version "11.0.1" 2018-10-16
OpenJDK Runtime Environment AdoptOpenJDK (build 11.0.1+13)
OpenJDK 64-Bit Server VM AdoptOpenJDK (build 11.0.1+13, mixed mode)
```
It seemed GIF image treated as interlace GIF.
But I checked interlace flag on GIF.
```
$ identify -verbose resources/images/misc/duke.gif | grep Interlace:
  Interlace: None
```
I checked the latest IBM Java8 for Linux amd64, but I could not recreate this issue.
I assume IBM Java8 applied a fix against class library on GIF image decoder related code.
But it seems register or memory area was not clean up properly.

![internalframe](https://user-images.githubusercontent.com/33543753/50816589-58dc4a00-1364-11e9-9b94-45980a5bed05.png)

ChengJin01:
Will investigate to see what happens.

pshipton:
Doe it still occur with -Xint?

[...]

PR #4274, opened 2019-01-11
Add tests for boolean truncation

tajila:
Add boolean tests

Signed-off-by: tajila <atobia@ca.ibm.com>

PR #4569, opened 2019-02-04
Extend booleans to the int type in the native

ChengJin01:
The change is to ensure booleans can be handled
appropriately as the int type when calling the
JNI natives.

Close: #4193

### DIFF
--- a/runtime/tests/jniarg/args_10.c
+++ b/runtime/tests/jniarg/args_10.c
@@ -1,0 +1,207 @@
+
+/*******************************************************************************
+ * Copyright (c) 2019, 2019 IBM Corp. and others
+ *
+ * This program and the accompanying materials are made available under
+ * the terms of the Eclipse Public License 2.0 which accompanies this
+ * distribution and is available at https://www.eclipse.org/legal/epl-2.0/
+ * or the Apache License, Version 2.0 which accompanies this distribution and
+ * is available at https://www.apache.org/licenses/LICENSE-2.0.
+ *
+ * This Source Code may also be made available under the following
+ * Secondary Licenses when the conditions for such availability set
+ * forth in the Eclipse Public License, v. 2.0 are satisfied: GNU
+ * General Public License, version 2 with the GNU Classpath
+ * Exception [1] and GNU General Public License, version 2 with the
+ * OpenJDK Assembly Exception [2].
+ *
+ * [1] https://www.gnu.org/software/classpath/license.html
+ * [2] http://openjdk.java.net/legal/assembly-exception.html
+ *
+ * SPDX-License-Identifier: EPL-2.0 OR Apache-2.0 OR GPL-2.0 WITH Classpath-exception-2.0 OR LicenseRef-GPL-2.0 WITH Assembly-exception
+ *******************************************************************************/
+
+#include "jniargtests.h"
+
+jboolean JNICALL Java_JniArgTests_nativeIZrZ(JNIEnv *p_env, jobject p_this, jint value, jbooleanArray backChannel)
+{
+	J9JavaVM *javaVM = getJ9JavaVM(p_env);
+	jboolean converted = (jboolean)value;
+	jboolean elements[] = {JNI_FALSE};
+	PORT_ACCESS_FROM_JAVAVM(javaVM);
+
+	jniTests++;
+
+	if (converted) {
+		/* if C could would see this as true, set the backchannel to true */
+		elements[0] = JNI_TRUE;
+	}
+
+	(*p_env)->SetBooleanArrayRegion(p_env, backChannel, 0, 1, elements);
+
+	return value;   /* return the value as a boolean */
+}
+
+jboolean JNICALL Java_JniArgTests_nativeZIZIZIZIZIrZ(JNIEnv *p_env, jobject p_this, jint arg1, jint arg2, jint arg3, jint arg4, jint arg5, jint arg6, jint arg7, jint arg8, jint arg9, jint arg10)
+{
+	J9JavaVM *javaVM = getJ9JavaVM(p_env);
+	PORT_ACCESS_FROM_JAVAVM(javaVM);
+
+	jniTests++;
+
+	if (arg1 != test_jboolean[0]) {
+		cFailure_jint(PORTLIB, "nativeZIZIZIZIZIrZ", 1, arg1, test_jboolean[0]);
+	}
+	if (arg2 != test_jint[1]) {
+		cFailure_jint(PORTLIB, "nativeZIZIZIZIZIrZ", 2, arg2, test_jint[1]);
+	}
+	if (arg3 != test_jboolean[1]) {
+		cFailure_jint(PORTLIB, "nativeZIZIZIZIZIrZ", 3, arg3, test_jboolean[1]);
+	}
+	if (arg4 != test_jint[2]) {
+		cFailure_jint(PORTLIB, "nativeZIZIZIZIZIrZ", 4, arg4, test_jint[2]);
+	}
+	if (arg5 != test_jboolean[0]) {
+		cFailure_jint(PORTLIB, "nativeZIZIZIZIZIrZ", 5, arg5, test_jboolean[0]);
+	}
+	if (arg6 != test_jint[3]) {
+		cFailure_jint(PORTLIB, "nativeZIZIZIZIZIrZ", 6, arg6, test_jint[3]);
+	}
+	if (arg7 != test_jboolean[1]) {
+		cFailure_jint(PORTLIB, "nativeZIZIZIZIZIrZ", 7, arg7, test_jboolean[1]);
+	}
+	if (arg8 != test_jint[4]) {
+		cFailure_jint(PORTLIB, "nativeZIZIZIZIZIrZ", 8, arg8, test_jint[4]);
+	}
+	if (arg9 != test_jboolean[0]) {
+		cFailure_jint(PORTLIB, "nativeZIZIZIZIZIrZ", 9, arg9, test_jboolean[0]);
+	}
+	if (arg10 != test_jint[5]) {
+		cFailure_jint(PORTLIB, "nativeZIZIZIZIZIrZ", 10, arg10, test_jint[5]);
+	}
+
+	return (JNI_TRUE == test_jboolean[0]);
+}
+
+jboolean JNICALL Java_JniArgTests_nativeIZIZIZIZIZrZ(JNIEnv *p_env, jobject p_this, jint arg1, jboolean arg2, jint arg3, jboolean arg4, jint arg5, jboolean arg6, jint arg7, jboolean arg8, jint arg9, jboolean arg10)
+{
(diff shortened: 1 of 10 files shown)


ChengJin01:
The changes mainly include:
1) extend boolean to int (4 byte) when calling JNI native
2) address the return boolean in the case of little endian & big-endian

ChengJin01:
Reviewer: @DanHeidinga 
FYI: @pshipton , @tajila

DanHeidinga:
@ChengJin01 Can you add the tests you used to validate the results as well?  I believe @tajila has contributed tests for the other boolean behaviour checks so you can build off that framework.

The return behaviour was driven by C's willingness to treat any non-zero value as "true".  I'm concerned that this change may result in "booleans" that appear true in the native code but false on return.  

I don't follow the parameter part of the change yet.  Can you expand on why this is correct?  The part I'm unclear on is that openj9 only adjusts the booleans / bytes / chars / etc when storing into a field or when returning them.  We don't normally adjust them when making a call.  Does this change that behaviour?

Does this change match the RI's behaviour?

ChengJin01:
@DanHeidinga 

[1] I'm concerned that this change may result in "booleans" that appear true in the native code but false on return.

---> This won't happen because the changes already addressed the issue with little/big-endian to ensure only the actual boolean value is fetched on return.


[2]I don't follow the parameter part of the change yet. Can you expand on why this is correct? The part I'm unclear on ... We don't normally adjust them when making a call. Does this change that behaviour? Does this change match the RI's behaviour?

---> let me illustrate what happened in the issue at #4193 on x86 as follows.
```
  java code            FFI code: copying stack arguments by the type size               JNI native
boolean (1byte) ---->    copy 1 byte (padded 7 byte for alignment)   -------->    int (4 bytes)
```
The problem there is FFI code only copied 1 byte for boolean to the native sack and left other 7 bytes (from alignment) uninitialized.  In such case, the casting from boolean to int was done automatically by including 1 byte for the actual boolean value plus other 3 garbage bytes due to 8 byte alignment, which led to incorrect result. That's why I initially proposed to initialize the whole argument stack with 0 to ensure there is no garbage data introduced when casting; otherwise, the garbage bytes are inevitably cast to the int value.

Actually, boolean is already stored as int (4 bytes) in our code, which follows the VM Spec. It is ffi_type that truncates or adjusts boolean to 1 byte in size.  That being said, we were able to copy 4 bytes for boolean but only copied 1 byte to the native stack, which should be addressed on our side.

 So, changing to `uint32 `in ffi_type is just to explicitly declare boolean as 4 bytes, in which case there is no need to do initialization on aligned bytes on the native argument stack in the FFI code.
```
  java code            FFI code: copying stack arguments by the type size               JNI native
boolean (4bytes) ------------>    copy 4 bytes        ----------------------->    int (4 bytes)
```

The changes are not adjusting but correcting something we missed in the code, which won't affect the existing behavior at this point.

ChengJin01:
@DanHeidinga ,  tobi's test framework at #4274 have not yet get merged to the repo. So my test cases are added to the existing framework at test/functional/NativeTest/jniargtests as it only focuses on the verification of native's arguments & return values, which is more suitable for our situation.

All added test cases have been verified in personal builds.

runtime/tests/jniarg/args_10.c
```diff
@@ -0,0 +1,184 @@
...
+ *
+ * [1] https://www.gnu.org/software/classpath/license.html
+ * [2] http://openjdk.java.net/legal/assembly-exception.html
+ *
+ * SPDX-License-Identifier: EPL-2.0 OR Apache-2.0 OR GPL-2.0 WITH Classpath-exception-2.0 OR LicenseRef-GPL-2.0 WITH Assembly-exception
+ *******************************************************************************/
+
+#include "jniargtests.h"
+
+jboolean JNICALL Java_JniArgTests_nativeZIZIZIZIZIrZ(JNIEnv *p_env, jobject p_this, int arg1, int arg2, int arg3, int arg4, int arg5, int arg6, int arg7, int arg8, int arg9, int arg10)
```
DanHeidinga: `int` should that be `jint`?  and `jboolean`?  The `ZIZIZIZIZI` signature doesn't match the declared types of the args

ChengJin01: @DanHeidinga ,  the behavior of test cases are based on the code changes in this PR.

Given that changing `uint8 `(1byte) to `uint32 `(4bytes) for boolean type affects both arguments and return type in size, we need to make sure the following cases are correct with my fix:
e.g.
```
definition in java                                             declaration/implementation in the native
1) nativeZIrZ(int arg1, boolean arg2) --->  nativeZIrZ(jint arg1, jint arg2)  (extend boolean to int in the native, which is exactly the issue we fixed in this PR)
2) nativeZIrZ(int arg1, boolean arg2) --->  nativeZIrZ(jint arg1, jboolean arg2)  (remains boolean  in the native to ensure it still works good)
3) return boolean in different ways (directly TRUE/FALSE or indirectly with == statement which only set value to AL in x86) to ensure it returns correctly on both little-endian and big-endian.
```

I will added the test case you mentioned above (which I overlooked) to see whether it works  good.

ChengJin01: @DanHeidinga ,  the behavior of test cases are based on the code changes in this PR.

Given that changing `uint8 `(1byte) to `uint32 `(4bytes) for boolean type affects both arguments and return type in size, we need to make sure the following cases are correct with my fix:
e.g.
```
definition in java                                             declaration/implementation in the native
1) nativeZIrZ(int arg1, boolean arg2) --->  nativeZIrZ(jint arg1, jint arg2)  (extend boolean to int in the native, which is exactly the issue we fixed in this PR)
2) nativeZIrZ(int arg1, boolean arg2) --->  nativeZIrZ(jint arg1, jboolean arg2)  (remains boolean  in the native to ensure it still works good)
3) return boolean in different ways (directly TRUE/FALSE or indirectly with == statement which only set value to AL in x86) to ensure it returns correctly on both little-endian and big-endian.
```

I will added the test case you mentioned above (which I overlooked) to see whether it works  good.

ChengJin01:
@DanHeidinga ,  the behavior of test cases are based on the code changes in this PR.

Given that changing `uint8 `(1byte) to `uint32 `(4bytes) for boolean type affects both arguments and return type in size, we need to make sure the following cases are correct with my fix:
e.g.
```
definition in java                                             declaration/implementation in the native
1) nativeZIrZ(int arg1, boolean arg2) --->  nativeZIrZ(jint arg1, jint arg2)  (extend boolean to int in the native, which is exactly the issue we fixed in this PR)
2) nativeZIrZ(int arg1, boolean arg2) --->  nativeZIrZ(jint arg1, jboolean arg2)  (remains boolean  in the native to ensure it still works good)
3) return boolean in different ways (directly TRUE/FALSE or indirectly with == statement which only set value to AL in x86) to ensure it returns correctly on both little-endian and big-endian.
```

I will added the test case you mentioned above (which I overlooked) to see whether it works  good.

ChengJin01:
@DanHeidinga ,   I already added the test case above (renamed to nativeIZrZ() /it should be SetBooleanArrayRegion rather than SetIntArrayRegion in the test code; otherwise the compilation fails due to incompatible type) and verified it with both our build (in personal build) and Hotspot Java 11 as follows:

[jnitest_hotspot_java11.zip](https://github.com/eclipse/openj9/files/2862085/jnitest_hotspot_java11.zip)
```
jdk_11.0.1_13_hotspot/bin/java -version
openjdk version "11.0.1" 2018-10-16
OpenJDK Runtime Environment AdoptOpenJDK (build 11.0.1+13)
OpenJDK 64-Bit Server VM AdoptOpenJDK (build 11.0.1+13, mixed mode)

jdk_11.0.1_13_hotspot/bin/java -Djava.library.path=".../jnitest"   JniBoolArgTest
Argument: value = 0 (0x0), converted = 0 (0x0), elements[0] = 0
1: retval_boolean = false, backChannel[0] = false

Argument: value = 1 (0x1), converted = 1 (0x1), elements[0] = 1
2: retval_boolean = true, backChannel[0] = true

Argument: value = 254 (0xFE), converted = 254 (0xFE), elements[0] = 1
3: retval_boolean = true, backChannel[0] = true

Argument: value = 3840 (0xF00), converted = 0 (0x0), elements[0] = 0 <---- the converted boolean value is false
4: retval_boolean = false, backChannel[0] = false <----- the return boolean value is false

Argument: value = 4094 (0xFFE), converted = 254 (0xFE), elements[0] = 1
5: retval_boolean = true, backChannel[0] = true
```

Our build and Hotspot ended up with the same results above, which means:
1) for a passed-in int value,  converting it to a boolean value only depends on the lowest byte of the int value, whether or not it is greater than 1 byte ( 255 or 0xFF).
2) the rule also applies to the int value on return.
That being said, the lowest byte of the int value is truncated to a boolean variable and only the lowest byte of the int value determines whether the boolean value is true or not on both conversion and return.
e.g (in the test above)
```
int value = 3840 (0xF00),  jboolean converted = (jboolean)value = 0 (0x0), return the int value as false (the lowest byte of the int value is 0x0)
int value = 4094 (0xFFE), jboolean converted = (jboolean)value = 254 (0xFE), return the int value as true ( 254 !=0)
```
So,  our build with my code changes matches Oracles's behavior at this point.

runtime/tests/jniarg/args_10.c
```diff
@@ -0,0 +1,207 @@
...
+{
+	J9JavaVM *javaVM = getJ9JavaVM(p_env);
+	jboolean converted = (jboolean)value;
+	jboolean elements[] = {JNI_FALSE};
+	PORT_ACCESS_FROM_JAVAVM(javaVM);
+
+	jniTests++;
+
+	if (converted) {
+		// if C could would see this as true, set the backchannel to true
```
DanHeidinga: Some of our compilers historically haven't supported `//` comments yet.  Can you change this to `/* */` in C files?

DanHeidinga:
@ChengJin01 Can you confirm the tests pass with `-Xjit:count=0` to ensure the DirectToJNI path is correct as well?

@fjeremic @andrewcraik can you comment on whether this change is likely to affect the DirectToJNI path?

fjeremic:
> The changes mainly include:
> 
>     1. extend boolean to int (4 byte) when calling JNI native
>     2. address the return boolean in the case of little endian & big-endian

I don't see why this would have any effect on the JIT. We should just add an `-Xjit:count=0` variation though as every platform could be different.

runtime/tests/jniarg/args_10.c
ChengJin01: Already updated with `/* */` and will double-check test cases with `-Xjit:count=0`.

ChengJin01:
Already verified in personal builds (Java 8 & 11) with` -Xjit:count=0` added for jniargtests in /test/functional/NativeTest/playlist.xml

DanHeidinga:
Jenkins test sanity plinux,zlinux jdk8

DanHeidinga:

DanHeidinga:
Jenkins test sanity aix,win jdk12

PR #4563, opened 2019-02-04
Delete references to obsolete classes and methods

pdbain-ibm:
Remove `include`s and methods required for Java 9 Build 148.

Signed-off-by: Peter Bain <peter_bain@ca.ibm.com>

PR #4570, opened 2019-02-04
Fold opening of getBootLayer() onto a single line

keithc-ca:
Perhaps this should have been included in #4563 with the removal of the conditional return type.


pshipton:
jenkins compile zlinux jdk11

PR #4581, opened 2019-02-04
Add assert fatal for unknown bytecode for JIT

cathyzhyi:
Change the code to recognize J9BCbreakpoint opcode rather pretend it is
unknown and abort compilation. Add assert fatal when hit real unknown
opcodes.


andrewcraik:

PR #4594, opened 2019-02-05
Implement storeValidationRecordIfNecessary for JITaaS

dmitry-ten:
[skip ci]

### DIFF
--- a/runtime/compiler/env/j9methodServer.cpp
+++ b/runtime/compiler/env/j9methodServer.cpp
@@ -1355,3 +1355,111 @@ TR_ResolvedJ9JITaaSServerMethod::getCachedResolvedMethod(TR_ResolvedMethodKey ke
       }
    return false;
    }
+
+bool
+TR_ResolvedRelocatableJ9JITaaSServerMethod::storeValidationRecordIfNecessary(TR::Compilation * comp, J9ConstantPool *constantPool, int32_t cpIndex, TR_ExternalRelocationTargetKind reloKind, J9Method *ramMethod, J9Class *definingClass)
+   {
+   TR_J9VMBase *fej9 = (TR_J9VMBase *) comp->fe();
+
+   bool storeClassInfo = true;
+   bool fieldInfoCanBeUsed = false;
+   TR_AOTStats *aotStats = ((TR_JitPrivateConfig *)fej9->_jitConfig->privateConfig)->aotStats;
+   bool isStatic = false;
+
+   TR::CompilationInfo *compInfo = TR::CompilationInfo::get(fej9->_jitConfig);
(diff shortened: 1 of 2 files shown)


dmitry-ten:

runtime/compiler/env/j9methodServer.cpp
```diff
@@ -1355,3 +1355,111 @@ TR_ResolvedJ9JITaaSServerMethod::getCachedResolvedMethod(TR_ResolvedMethodKey ke
...
+   {
+   TR_J9VMBase *fej9 = (TR_J9VMBase *) comp->fe();
+
+   bool storeClassInfo = true;
+   bool fieldInfoCanBeUsed = false;
+   TR_AOTStats *aotStats = ((TR_JitPrivateConfig *)fej9->_jitConfig->privateConfig)->aotStats;
+   bool isStatic = false;
+
+   TR::CompilationInfo *compInfo = TR::CompilationInfo::get(fej9->_jitConfig);
+   TR_RelocationRuntime *reloRuntime = compInfo->getCompInfoForThread(fej9->vmThread())->reloRuntime();
```
mpirvu: We should make sure that the reloRuntime stored in compInfo is of type TR_SharedCacheJITaaSRelocationRuntime

runtime/compiler/env/j9methodServer.cpp
```diff
@@ -1355,3 +1355,111 @@ TR_ResolvedJ9JITaaSServerMethod::getCachedResolvedMethod(TR_ResolvedMethodKey ke
...
+
+   isStatic = (reloKind == TR_ValidateStaticField);
+
+   TR_J9ServerVM *feServer = (TR_J9ServerVM *) fej9;
+   J9Class *clazz = (J9Class *) feServer->TR_J9ServerVM::getClassOfMethod((TR_OpaqueMethodBlock *) ramMethod);
```
mpirvu: Would it be bad if we called directly feServer->getClassOfMethod()?

dmitry-ten: If I'm not wrong, `fej9` will be instantiated to `TR_SharedCacheServerVM`, so calling `feServer->getClassOfMethod()` will call the AOT of the method, which we do not want.

runtime/compiler/env/j9methodServer.cpp
dmitry-ten: Yeah, but that should probably go in a different commit, the one that implements RelocationRuntime.

runtime/compiler/env/j9methodServer.cpp
mpirvu: I agree

PR #4596, opened 2019-02-05
Fix a bug in packed decimal shift

NigelYiboYu:
Fix a problem in packed decimal left shift to make it produce the correct results.
Ensure that PD shift operations can cause HW overflow exceptions and can
fall back to Java implementation so that Java exceptions can be thrown.


fjeremic:

PR #4409, opened 2019-01-23
Create jvmti.h.m4 to generate different versions of jvmti.h

MarkQingGuo:
we use m4 to generate different jvmti header files to support different JDKs

we need to generate jvmti header by using m4 so that it is specific to
the Java release version, since vm will support Java 8, 9, 10, 11, 12.
Some PRs in extension repositories are created.

There are some changes:

- use m4 preprocessor and macros to generate different header
- add macros in other jvmti implementation files so that it matches header files
- modify dependency in `runtime/CMakeList.txt` to support cmake build
- modify dependency in `runtime/buildtools.mk` to support Uma build
- modify dependency in `runtime/compiler/linter.mk` to support linter build

Closes: #2368
 
Signed-off-by: MarkQingGuo <Qing.Guo@ibm.com>

Issue #116, opened 2017-09-22
Travis PR builds broken


dnakamura:
Travis pull request builds are broken. They time out cloning `openj9-openjdk-jdk9`

```Cloning into 'openj9-openjdk-jdk9'...
remote: Counting objects: 565618, done.
remote: Compressing objects: 100% (13/13), done.
Receiving objects:  98% (558063/565618), 776.86 MiB | 160.00 KiB/s   
The job exceeded the maximum time limit for jobs, and has been terminated.
```
https://travis-ci.org/eclipse/openj9/builds/278682619

~~Builds on branches still seem ok~~ I have seen the issue strike on branch builds but much less often than on PR builds

AdamBrousseau:
Can this be closed?

DanHeidinga:
I think so.  There has been other travis build flakiness but that's been related to source forge downloads of free marker.

PR #4605, opened 2019-02-05
Add a new target to generate jvmti.h when build with linter in travis

MarkQingGuo:
A new target(java_spec_headers) is added in build-on-travis.mk

A new target is necessary because it generates jvmti header files. Since we need to generate different jvmti header files for different JDKs and building with linter doesn't depend on extension repository, we have to modify makefiles so that it generates jvmti header files when build with linter.

PRs should be commited in such order:

- jvmti.h.m4 which generates jvmti header files(#4409)
- changes in extension repositories(JDK8, JDK11, JDK12)
- changes in JDK11(#116) in its specific extension repository
- changes in how openj9 are build with linter(#4605)

There are some changes:
- A new target(java_spec_headers) is added
- A new recipe of target java_spec_headers is added: it gets java_spec_version from j9cfg.h then uses it as a variable

### DIFF
--- a/buildenv/travis/build-on-travis.sh
+++ b/buildenv/travis/build-on-travis.sh
@@ -60,7 +60,7 @@ if test "x$RUN_LINT" = "xyes"; then
   mkdir $CMAKE_BUILD_DIR
   cd $CMAKE_BUILD_DIR
   cmake -C $J9SRC/cmake/caches/linux_x86-64_cmprssptrs.cmake ..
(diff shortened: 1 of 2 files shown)


MarkQingGuo:
@keithc-ca here maybe other changes we need, if we want to build on travis.

keithc-ca:
I am hoping that #4409 will be updated to handle cmake builds so that this is unnecessary.

PR #4602, opened 2019-02-05
Fix missing atomic-free JNI check

gacholio:
printThreadInfo was not handling atomic-free JNI. Copy the conditional
acquire/release code from RAS.

[ci skip]

Signed-off-by: Graham Chapman <graham_chapman@ca.ibm.com>

PR #4611, opened 2019-02-06
Fix unused variable warning when J9VM_INTERP_ATOMIC_FREE_JNI is not defined

keithc-ca:
Companion to #4602


pshipton:
jenkins compile zlinux jdk8

PR #4615, opened 2019-02-06
Fix missing return value assignment

gacholio:
Missing `*` on assignment, not caught by compiler due to the use of void
pointers. Also fix the associated test.

[ci skip]

Signed-off-by: Graham Chapman <graham_chapman@ca.ibm.com>

PR #4617, opened 2019-02-06
Ensure out pointer is correct assigned in jvmtiGetJ9method

DanHeidinga:
Fix for issue introduced in #3492 which ensured the values were always
initialized


gacholio:
Dup of #4615

PR #4627, opened 2019-02-06
Clean up deprecated code resulting from migrating to a unified assembler across x86

nbhuiyan:
This PR removes the MASM assembly files and cleans up the dual (or triple, if you want to count masm2gas) assembler solutions that existed during the transition period.

Closes: #3148


nbhuiyan:
@0xdaryl ready for review

0xdaryl:
Jenkins test sanity win,xlinux,xlinuxlargeheap jdk11

0xdaryl:

0xdaryl:
Changes look reasonable to me.  One unfortunate consequence is that we lose the history on the deleted assembly files without doing a bit of git work to dig it out.  However, this probably isn't such a big deal in practice because there isn't much of a history on these assembly files in the 1.5 years this code has been open sourced.

@0dvictor , I'd appreciate a +1 from you too.

0dvictor:
Changes look good to me. I totally agree deleting these deprecated files; they are slowly out-of-sync and likely stop work in the future. Despite the short history of these files, all histories are still in git though it takes more effort retrieving it. I don't anticipate frequent need of these histories.

0xdaryl:
Jenkins test sanity plinux,zlinux jdk11

PR #4628, opened 2019-02-06
Ensure _Filehandle does not close if printing to STDOUT/ERR

### LINKED ISSUE #2837: Provide option to print javacore output to stdout/stderr instead of to a file

AidanHa:
Replace '1' and '2' with J9PORT_TTY_OUT and J9PORT_TTY_ERR respectively
Replace '/STDOUT/' and '/STDERR/' with J9RAS_STDOUT_NAME and J9RAS_STDERR_NAME
Add support for choosing a file name with `-` to print to STDOUT
Add support for choose a file name `/stdout/` and `/stderr/` to print to STDOUT and STDERR respectively.
Add a boolean to the class `IsOpen` that is set to true when the file stream is open.

Doc issue [eclipse/openj9-docs#204](https://github.com/eclipse/openj9-docs/issues/204)

Fixes #2837

### DIFF
--- a/runtime/rasdump/TextFileStream.cpp
+++ b/runtime/rasdump/TextFileStream.cpp
@@ -25,9 +25,12 @@
 #include "TextFileStream.hpp"
 #include "j9.h"
 #include "j9port.h"
+#include "rasdump_internal.h"
 
 /* Constructor */
 TextFileStream::TextFileStream(J9PortLibrary* portLibrary) :
+	_Buffer(NULL),
+	_IsOpen(false),
 	_BufferPos(0),
 	_BufferSize(16*1024),
 	_PortLibrary(portLibrary),
@@ -51,12 +54,16 @@ void
 TextFileStream::open(const char* fileName, bool cacheWrites)
 {
 	PORT_ACCESS_FROM_PORT(_PortLibrary);
-	if (NULL != strstr(fileName, "/STDOUT/")) {
-		_FileHandle = 1; 
-	} else if (NULL != strstr(fileName, "/STDERR/")) {
-		_FileHandle = 2;
-	} else if (fileName[0] != '-' ) {
+	if (0 == strcmp(fileName, J9RAS_STDOUT_NAME)) {
+		_FileHandle = J9PORT_TTY_OUT; 
+	} else if (0 == strcmp(fileName, J9RAS_STDERR_NAME)) {
+		_FileHandle = J9PORT_TTY_ERR;
+	} else {
 		_FileHandle = j9file_open(fileName, EsOpenWrite | EsOpenCreate | EsOpenTruncate | EsOpenCreateNoTag, 0666);
+		if (_FileHandle != -1) {
+			_IsOpen = true;
+		}
+		
 	}
 	if(!cacheWrites) {
 		_BufferSize = 0;
@@ -73,7 +80,10 @@ TextFileStream::close(void)
 			j9file_write_text(_FileHandle, _Buffer, _BufferPos);
 		}
 		j9file_sync(_FileHandle);
-		j9file_close(_FileHandle);
+		if (_IsOpen) {
+			/*If we don't open the file stream, it means that we don't close it either */
+			j9file_close(_FileHandle);
+		}
 	}
 
 	_FileHandle = -1;	
--- a/runtime/rasdump/dmpagent.c
+++ b/runtime/rasdump/dmpagent.c
@@ -886,7 +886,12 @@ static omr_error_t
 doJavaDump(J9RASdumpAgent *agent, char *label, J9RASdumpContext *context)
 {
 	J9JavaVM *vm = context->javaVM;
-	if ((0 != strcmp(label, "/STDOUT/")) && (0 != strcmp(label, "/STDERR/"))) {
+
+	if ((0 == strcmp("-", label)) || (0 == j9_cmdla_stricmp(label, J9RAS_STDOUT_NAME))) {
+		strcpy(label, J9RAS_STDOUT_NAME);
+	} else if (0 == j9_cmdla_stricmp(label, J9RAS_STDERR_NAME)) {
+		strcpy(label, J9RAS_STDERR_NAME);
+	} else {
 		if (makePath(vm, label) == OMR_ERROR_INTERNAL) {
 			/* Nowhere available to write the dump, we are done, makePath() will have issued error message */
 			return OMR_ERROR_INTERNAL;
--- a/runtime/rasdump/rasdump_internal.h
+++ b/runtime/rasdump/rasdump_internal.h
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 1991, 2018 IBM Corp. and others
+ * Copyright (c) 1991, 2019 IBM Corp. and others
  *
  * This program and the accompanying materials are made available under
  * the terms of the Eclipse Public License 2.0 which accompanies this
@@ -125,6 +125,9 @@ void setAllocationThreshold(J9VMThread *vmThread, UDATA min, UDATA max);
 #endif
 
 #define J9RAS_DUMP_EXCEPTION_EVENT_GROUP (J9RAS_DUMP_ON_EXCEPTION_THROW | J9RAS_DUMP_ON_EXCEPTION_SYSTHROW | J9RAS_DUMP_ON_EXCEPTION_CATCH | J9RAS_DUMP_ON_EXCEPTION_DESCRIBE)
+
+#define J9RAS_STDOUT_NAME "/STDOUT/"
+#define J9RAS_STDERR_NAME "/STDERR/"
 
 #ifdef __cplusplus
 }
--- a/runtime/rasdump/TextFileStream.hpp
+++ b/runtime/rasdump/TextFileStream.hpp
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2003, 2014 IBM Corp. and others
+ * Copyright (c) 2003, 2019 IBM Corp. and others
  *
  * This program and the accompanying materials are made available under
  * the terms of the Eclipse Public License 2.0 which accompanies this
@@ -67,6 +67,7 @@ private :
 	TextFileStream(const TextFileStream& source);
 	TextFileStream& operator=(const TextFileStream& source);
 	char *_Buffer;
+	bool _IsOpen;
 	UDATA _BufferPos;
 	UDATA _BufferSize;
 


runtime/oti/j9port.h
```diff
@@ -765,6 +765,9 @@ typedef struct J9CacheInfoQuery {
 #define J9PORT_TTY_OUT OMRPORT_TTY_OUT
 #define J9PORT_TTY_ERR OMRPORT_TTY_ERR
 
+#define J9PRINT_TO_STDOUT "/STDOUT/"
```
pshipton: I don't think this file is the appropriate place for these defines. They are specific to the rasdump code and not the port library. Please put them in an include file specific to the rasdump code.
The defines should also be used in dmpagent.c

runtime/rasdump/TextFileStream.cpp
```diff
@@ -68,7 +68,7 @@ void
 TextFileStream::close(void)
 {
 	PORT_ACCESS_FROM_PORT(_PortLibrary);
-	if (_FileHandle != -1) {
+	if (_FileHandle != -1 && _FileHandle != OMRPORT_TTY_OUT && _FileHandle != OMRPORT_TTY_ERR) {
```
pshipton: The check for TTY_OUT/ERR needs to done around the close. The j9file_write_text() and j9file_sync() calls are still desired.

runtime/oti/j9port.h
AidanHa: I've pushed new changes to address your comments. Hopefully I chose the proper file to put the defines in.

runtime/rasdump/rasdump_internal.h
```diff
@@ -126,6 +126,9 @@ void setAllocationThreshold(J9VMThread *vmThread, UDATA min, UDATA max);
 
 #define J9RAS_DUMP_EXCEPTION_EVENT_GROUP (J9RAS_DUMP_ON_EXCEPTION_THROW | J9RAS_DUMP_ON_EXCEPTION_SYSTHROW | J9RAS_DUMP_ON_EXCEPTION_CATCH | J9RAS_DUMP_ON_EXCEPTION_DESCRIBE)
 
+#define J9PRINT_TO_STDOUT "/STDOUT/"
```
pshipton: How about J9RAS_PRINT_TO...

runtime/rasdump/TextFileStream.cpp
```diff
@@ -73,7 +74,9 @@ TextFileStream::close(void)
 			j9file_write_text(_FileHandle, _Buffer, _BufferPos);
 		}
 		j9file_sync(_FileHandle);
-		j9file_close(_FileHandle);
+		if (_FileHandle != OMRPORT_TTY_OUT && _FileHandle != OMRPORT_TTY_ERR) {
```
pshipton: Please be consistent in the constant names. Above on line 56, 58, J9PORT is used instead of OMRPORT. I think using OMRPORT is fine. The J9PORT constants were added to avoid having to fix all the code when the port library moved to OMR and got renamed from J9 to OMR.

AidanHa: Sorry, I must've copied the variable name wrong... Clumsy mistake

runtime/rasdump/rasdump_internal.h
AidanHa: I Shall change the macros to : 
J9RAS_PRINT_TO_STDOUT
J9RAS_PRINT_TO_STDERR

Is the location of the macro fine?

pshipton: @keithc-ca are you good with the location?

runtime/rasdump/TextFileStream.cpp
```diff
@@ -51,10 +52,10 @@ void
...
 {
 	PORT_ACCESS_FROM_PORT(_PortLibrary);
-	if (NULL != strstr(fileName, "/STDOUT/")) {
-		_FileHandle = 1; 
-	} else if (NULL != strstr(fileName, "/STDERR/")) {
-		_FileHandle = 2;
+	if (NULL != strstr(fileName, J9PRINT_TO_STDOUT)) {
+		_FileHandle = J9PORT_TTY_OUT; 
+	} else if (NULL != strstr(fileName, J9PRINT_TO_STDERR)) {
+		_FileHandle = J9PORT_TTY_ERR;
```
keithc-ca: This should use the same names as below (`OMRPORT_TTY_OUT ` and `OMRPORT_TTY_ERR `).

runtime/rasdump/TextFileStream.cpp
keithc-ca: Better than checking the value of `_FileHandle`, I think a boolean variable set if the call to `j9file_open` succeeds should be the condition for calling `j9file_close`.

runtime/rasdump/rasdump_internal.h
keithc-ca: I'm fine with the location, but I'd suggest the names be `J9RAS_STDOUT_NAME` and `J9RAS_STDERR_NAME`.
I'm not a fan of the slashes (`/`) in values; it may be better if the names couldn't be confused for path names (e.g. `[stdout]` or some other characters that are illegal (at least on Windows)).

runtime/rasdump/TextFileStream.cpp
```diff
@@ -51,10 +52,10 @@ void
 TextFileStream::open(const char* fileName, bool cacheWrites)
 {
 	PORT_ACCESS_FROM_PORT(_PortLibrary);
-	if (NULL != strstr(fileName, "/STDOUT/")) {
-		_FileHandle = 1; 
-	} else if (NULL != strstr(fileName, "/STDERR/")) {
-		_FileHandle = 2;
+	if (NULL != strstr(fileName, J9PRINT_TO_STDOUT)) {
```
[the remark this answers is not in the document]
AidanHa: The reason why we use strstr is because `fileName` actually includes the path of the file (Eg: /root/build/..../FILENAME. If we used a regular strncmp, it would return false. I think the code to manually parse for the filename would be too tedious, and I don't think that there's any implemented function to get the direct file name (without the path).

Also @keithc-ca, the reason why we used slashes in the option is because it is illegal to name a file with slashes, and when we use strstr, there wont be a file with the substring `/STDOUT/`.

If there is a better solution, please let me know!

runtime/rasdump/rasdump_internal.h
AidanHa: Okay I shall change the macros to :
J9RAS_STDOUT_NAME
J9RAS_STDERR_NAME

runtime/rasdump/TextFileStream.cpp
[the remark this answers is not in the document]
keithc-ca: We _want_ it to treat a path like `.../STDOUT/whatever/...` as it always has; that is why `strstr` is inappropriate.

We need to use something like `strcasecmp` for a case insensitive comparison.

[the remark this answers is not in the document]
AidanHa: Gotcha, is there a way to grab the current directory path in openJ9? I will be needing that when using `strcasecmp` since the filaname includes the path to the file (and for strcasecmp to work, we need to compare the file name to `path+"STDOUT"`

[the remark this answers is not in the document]
keithc-ca: I'm sure there is, but that doesn't factor into the test we want; we only want to look at the `file=xxx` part of the command-line argument.

AidanHa:
Ive added support for `/stdout/`, `/stderr/` and `-`. Hopefully I've understood the request correctly.

runtime/rasdump/TextFileStream.cpp
```diff
@@ -73,7 +74,9 @@ TextFileStream::close(void)
 			j9file_write_text(_FileHandle, _Buffer, _BufferPos);
 		}
 		j9file_sync(_FileHandle);
-		j9file_close(_FileHandle);
+		if (_FileHandle != J9PORT_TTY_OUT && _FileHandle != J9PORT_TTY_ERR) {
```
keithc-ca: This condition should be base on a local variable which is set only after a successful call to `j9file_open` and not make assumptions about the values that function returns.

runtime/rasdump/dmpagent.c
```diff
@@ -886,7 +886,16 @@ static omr_error_t
...
+	
+	if ('-' == label[0] || 0 == strcasecmp(label, J9RAS_STDOUT_NAME)) {
+                label = J9RAS_STDOUT_NAME;
+        } 
+        
+	if (0 == strcasecmp(label, J9RAS_STDERR_NAME)) {
+                label = J9RAS_STDERR_NAME;
+        }
+	
+	if ((0 != strcmp(label, J9RAS_STDOUT_NAME)) && (0 != strcmp(label, J9RAS_STDERR_NAME))) {
```
keithc-ca: Please fix the indentation (use leading tabs).
Please add parentheses around the parts of compound conditions.
Use `if` / `else if` to avoid repeating (the incorrect) checks for stdout/stderr.
This function is expected to write into the buffer addressed by `label` the name of the output file; use `strcpy` for the stdout and stderr cases.

AidanHa: Sloppy coding on my end, that was my fault. I'll run a personal build with my changes before pushing the changes.

runtime/rasdump/TextFileStream.cpp
AidanHa: I was thinking :
`if ((_FileHandle != J9PORT_TTY_OUT) && (_FileHandle != J9PORT_TTY_ERR) && _FileHandle != -1)`
rather than adding a local variable. Is there a more elegant solution?

runtime/rasdump/dmpagent.c
AidanHa: If you were talking about the repetition on line 898, we need to have a separate `if` statement since the first 2 if statements will change `-`, `/stdout/` and `/stderr/` into `/STDOUT/` and `/STDERR/` respectively.

runtime/rasdump/TextFileStream.cpp
keithc-ca: A local variable is not enough; we need a field to capture that state. I don't know about 'elegant', but it's definitely cleaner than making assumptions about `j9file_open` return values.

AidanHa: It seems like the only time where `j9file_open` is not open is when `_FileHandle` = -1. Can't we just simply do a check for `_FileHandle == -1` ? That check would simply be isOpen().

I might be missing something here... Please let me know if that's the case

pshipton: I don't really follow either. The code is setting _FileHandle to J9PORT_TTY_OUT and J9PORT_TTY_ERR and not calling j9file_open(). If j9file_open() was called and somehow returned J9PORT_TTY_OUT or J9PORT_TTY_ERR, wouldn't closing these be wrong?

keithc-ca: Yes, if j9file_open() returned J9PORT_TTY_OUT or J9PORT_TTY_ERR, then _not_ closing it would be wrong.

This can really happen. Take a Linux system for example, and imagine file descriptor 2 was closed. The next call to j9file_open() might return 2 which should be closed, but the code as currently written would not.

AidanHa: > Yes, if j9file_open() returned J9PORT_TTY_OUT or J9PORT_TTY_ERR, then _not_ closing it would be wrong.

I'm confused. If that's the case then when do we ever close it?

AidanHa: https://github.com/eclipse/openj9/pull/4521#discussion_r254310704

@keithc-ca Didn't you say in this comment that we should not close the file stream if we are printing to STDOUT/STDERR. 


> Yes, if j9file_open() returned J9PORT_TTY_OUT or J9PORT_TTY_ERR, then _not_ closing it would be wrong.
> 
> This can really happen. Take a Linux system for example, and imagine file descriptor 2 was closed. The next call to j9file_open() might return 2 which should be closed, but the code as currently written would not.

Isn't file descriptor 2 the same as returning J9_TTY_ERR, and thus, the file stream was not even open to begin with?

AidanHa:
@keithc-ca Hopefully these changes are too your liking. I've built the changes on my local machine, and they work like a charm.

runtime/rasdump/dmpagent.c
```diff
@@ -886,7 +886,12 @@ static omr_error_t
 doJavaDump(J9RASdumpAgent *agent, char *label, J9RASdumpContext *context)
 {
 	J9JavaVM *vm = context->javaVM;
-	if ((0 != strcmp(label, "/STDOUT/")) && (0 != strcmp(label, "/STDERR/"))) {
+
+	if (('-' == label[0]) || (0 == strcasecmp(label, J9RAS_STDOUT_NAME))) {
```
keithc-ca: This matches anything that starts with `'-'`: you probably want to use `strcmp` instead.

runtime/rasdump/dmpagent.c
```diff
@@ -886,7 +886,12 @@ static omr_error_t
...
+			strcpy(label, J9RAS_STDOUT_NAME);
```
keithc-ca: Please fix the indentation here and two lines below.

runtime/rasdump/TextFileStream.cpp
```diff
@@ -25,9 +25,11 @@
 #include "TextFileStream.hpp"
 #include "j9.h"
 #include "j9port.h"
+#include "rasdump_internal.h"
 
 /* Constructor */
 TextFileStream::TextFileStream(J9PortLibrary* portLibrary) :
+	_IsOpen(false),
```
keithc-ca: Please also initialize `_Buffer`.

runtime/rasdump/dmpagent.c
AidanHa: Sorry, maybe I should stop using Vim to do my code changes....

runtime/rasdump/TextFileStream.cpp
AidanHa: Isn't `_Buffer` initialized on line 39?

`_Buffer = (char *) OMRPORT_FROM_J9PORT(_PortLibrary)->mem_allocate_memory(OMRPORT_FROM_J9PORT(_PortLibrary), _BufferSize, "TextFileStream::TextFileStream", OMRMEM_CATEGORY_VM);`

keithc-ca: Ideally all fields should be constructed before the body of the constructor begins; adding `_Buffer(NULL),` before `_IsOpen(false),` does that.

AidanHa:
@keithc-ca I've made some changes, and the options still seem to work as expected.

AidanHa:
@pshipton I've updated the doc issue at https://github.com/eclipse/openj9-docs/issues/204 as well

runtime/rasdump/TextFileStream.cpp
```diff
@@ -87,7 +97,7 @@ TextFileStream::close(void)
 /* Methods for getting the object's status */
 bool TextFileStream::isOpen(void) const
 {
-	return _FileHandle != -1;
+	return _IsOpen;
```
keithc-ca: I'm not sure this is appropriate. When writing to stdout/stderr, `javadump.cpp` will report an error:
> J9NLS_DMP_NO_CREATE=Aborting: Cannot create file (%s)

We may need to expand the API of this class to avoid inappropriate diagnostics.

AidanHa: I see. So should I revert this change, then perhaps create another issue for IsOpen()?

keithc-ca: I think it makes sense to address the javadump behavior in this PR (there is no issue without these changes).

AidanHa:
I used Is_STDOUTERR rather than isFile since i found that it gets really easy to confuse isOpen=false and isFile=false with the current implementation...

AidanHa:
Also @keithc-ca, when I was adding another print option in `j9dmp.nls` and `j9dmp.tdf` I found that there were other files that contained the same string in different languages (EG: runtime/nls/dump/j9dmp_ja.nls has `J9NLS_DMP_WRITTEN_DUMP_STR` in japanese). IF I were to add another string/option (not to sure what I should call this), then would I need to add translation for each language?

runtime/rasdump/TextFileStream.cpp
```diff
@@ -90,6 +100,11 @@ bool TextFileStream::isOpen(void) const
 	return _FileHandle != -1;
 }
 
+bool TextFileStream::isSTDOUTERR(void) const
+{
+	return (!_IsOpen && (_FileHandle != -1));/* Returns true if writting to STDOUT/ERR*/
```
keithc-ca: Please remove the outer-most parentheses, and fix the typo: writing has only one 't'.

runtime/rasdump/javadump.cpp
```diff
@@ -520,6 +523,9 @@ JavaCoreDumpWriter::JavaCoreDumpWriter(
 	if (_Error) {
 		j9nls_printf(PORTLIB, J9NLS_ERROR | J9NLS_STDERR, J9NLS_DMP_ERROR_IN_DUMP_STR, "Java", _FileName);
 		Trc_dump_reportDumpError_Event2("Java", _FileName);
+	} else if (_IsSTDOUTERR) {
+		j9nls_printf(PORTLIB, J9NLS_INFO | J9NLS_STDERR, J9NLS_DMP_WRITTEN_DUMP_STDOUTERR_STR, "Java");
```
keithc-ca: This should be written to stdout if the report was produced on stdout.

runtime/rasdump/javadump.cpp
```diff
@@ -333,6 +333,7 @@ private :
 	const char*       _FileName;
 	TextFileStream    _OutputStream;
 	bool              _FileMode;
+	bool              _IsSTDOUTERR;
```
keithc-ca: This doesn't need to be a field - it's only used in one method.

runtime/rasdump/javadump.cpp
pshipton: I think this need to continue to be written to stderr. All the NLS error codes and messages have always gone to stderr, I don't think it should be changed for one particular message. For example, `-Xdump:java:events=vmstart` produces the following messages.
```
JVMDUMP039I Processing dump event "vmstart", detail "" at 2019/02/14 13:45:20 - please wait.
JVMDUMP032I JVM requested Java dump using 'jdk-11.0.2+8/javacore.20190214.134520.3456.0001.txt' in response to an event
JVMDUMP010I Java dump written to jdk-11.0.2+8/javacore.20190214.134520.3456.0001.txt
JVMDUMP013I Processed dump event "vmstart", detail "".
```

pshipton:
New NLS messages are only added in the master file in English. New messages get translated by IBM about once a year.

runtime/nls/dump/j9dmp.nls
```diff
@@ -157,6 +157,16 @@ J9NLS_DMP_WRITTEN_DUMP_STR.link=
 
 # END NON-TRANSLATABLE
 
+J9NLS_DMP_WRITTEN_DUMP_STDOUTERR_STR=%1$s dump written
+# START NON-TRANSLATABLE
+J9NLS_DMP_WRITTEN_DUMP_STR.explanation=The dump was written to STDOUT/ERR.
+J9NLS_DMP_WRITTEN_DUMP_STR.system_action=The JVM continues.
+J9NLS_DMP_WRITTEN_DUMP_STR.user_response=To understand how to interpret a java dump file, refer to diagnostics information.
+J9NLS_DMP_WRITTEN_DUMP_STR.sample_input_1=Java
+J9NLS_DMP_WRITTEN_DUMP_STR.link=
```
keithc-ca: Given @pshipton's suggestion to just use `/STDOUT/` or `/STDERR/` for the 'file name', I don't think we need or want a new NLS message.

runtime/rasdump/j9dmp.tdf
```diff
@@ -29,6 +29,7 @@ TraceEvent=Trc_dump_J9VMDllMain_Event2 NoEnv Overhead=1 Level=1 Template="Dump c
 TraceEvent=Trc_dump_reportDumpStart_Event1 NoEnv Overhead=1 Level=1 Template="JVM Requesting %s Dump using filename=%s"
 TraceEvent=Trc_dump_reportDumpError_Event2 NoEnv Overhead=1 Level=1 Template="Error in %s Dump: %s"
 TraceEvent=Trc_dump_reportDumpEnd_Event2 NoEnv Overhead=1 Level=1 Template="%s Dump written to filename=%s"
+TraceEvent=Trc_dump_reportDumpEnd_Event3 NoEnv Overhead=1 Level=1 Template="%s Dump finished writing"
```
keithc-ca: New trace events must be added at the end (to retain ids of existing events).

runtime/rasdump/javadump.cpp
keithc-ca: In light of this, I think we should stick with the existing message (`_FileName` will be `/STDOUT/` or `/STDERR/`).
I think this also means we don't need the new trace event nor the `isSTDOUTERR()` method.

runtime/rasdump/j9dmp.tdf
AidanHa: I guess since we'll be stick with the existing messages (Ie, what we had before). Is it still nessesary to create a new trace event, or can we use the same event as priting to a file (Dump using filename...)

keithc-ca: Agreed (see my next comment).

AidanHa:
Changes have been reverted. The PR is a lot simpler now

keithc-ca:

keithc-ca:
The Windows build failed because `dmpagent.c` is missing a declaration of `strcasecmp`. We may need to do like in `decomp003.c`:
```
#if defined(_MSC_VER)
#define strcasecmp _stricmp
#endif /* _MSC_VER */

File: runtime/rasdump/dmpagent.c [hunk 18]
```diff
@@ -886,7 +886,12 @@ static omr_error_t
 doJavaDump(J9RASdumpAgent *agent, char *label, J9RASdumpContext *context)
 {
 	J9JavaVM *vm = context->javaVM;
-	if ((0 != strcmp(label, "/STDOUT/")) && (0 != strcmp(label, "/STDERR/"))) {
+
+	if ((0 == strcmp("-", label)) || (0 == j9_cmdla_stricmp(label, J9RAS_STDOUT_NAME))) {
+		strcpy(label, J9RAS_STDOUT_NAME);
+	} else if (0 == strcasecmp(label, J9RAS_STDERR_NAME)) {
```
keithc-ca: All uses of `strcasecmp`, including this one, will need to switch to `j9_cmdla_stricmp`.

AidanHa: Sorry I forgot about this one

keithc-ca: @AidanHa Will you push your changes shortly?

AidanHa: Sorry, I pushed the changes yesterday morning, but they didn't show up until i did a rebase...

pshipton:
Jenkins test extended plinux jdk11

pshipton:
jenkins compile win jdk11

pshipton:
Jenkins test extended plinux jdk11

PR #4630, opened 2019-02-06
Auto-detect Visual Studio version

0dvictor:
Try `VisualStudioVersion` env var when `MSVC_VERSION` is not set; allowing more reliably detect Visual Studio version.


andrewcraik:

andrewcraik:
This should be fine as an automake only change since CMake already handles the auto detection for us.

andrewcraik:
infra fail on jdk8 linux64 - we can try this one again

andrewcraik:

PR #4635, opened 2019-02-06
Fix SVM Assert

dsouzai:
When enabling SVM asserts for AOT w/ SVM, an assert will
trigger during the relocation of an inlined site if its caller is a
profiled inlined method who's site was disabled (as part of _its_
relocation). This is because, under the SVM, relocation of the inlined
table entries are guaranteed (meaning, though the inlined site may be
disabled, the J9Method pointer in the inlined table in the metadata will
be updated to **not** be -1). However, when a profiled inlined method is
relocated, it does not go through the same code path as other inlined
method relocations, and so did not fix up the metadata when the site was
disabled.

This PR refactors the profiled inlined method relocation to use the
class ID from the SVM. It also relocates the inlined table entry
regardless of whether the inlined site can be activated.

### DIFF
--- a/runtime/compiler/runtime/RelocationRecord.cpp
+++ b/runtime/compiler/runtime/RelocationRecord.cpp
@@ -2181,6 +2181,28 @@ TR_RelocationRecordInlinedMethod::preparePrivateData(TR_RelocationRuntime *reloR
    RELO_LOG(reloRuntime->reloLogger(), 5, "\tpreparePrivateData: ramMethod %p inlinedSiteIsValid %d\n", ramMethod, inlinedSiteIsValid);
    }
 
+bool
+TR_RelocationRecordInlinedMethod::inlinedSiteCanBeActivated(TR_RelocationRuntime *reloRuntime,
+                                                            TR_RelocationTarget *reloTarget,
+                                                            J9Method *currentMethod)
+   {
+   TR::SimpleRegex * regex = reloRuntime->options()->getDisabledInlineSites();
+   if (regex && TR::SimpleRegex::match(regex, inlinedSiteIndex(reloTarget)))
+      {
+      RELO_LOG(reloRuntime->reloLogger(), 6, "\tinlinedSiteCanBeActivated: inlined site forcibly disabled by options\n");
+      return false;
+      }
+
+   if (reloRuntime->fej9()->isAnyMethodTracingEnabled((TR_OpaqueMethodBlock *) currentMethod) ||
+       reloRuntime->fej9()->canMethodEnterEventBeHooked() ||
+       reloRuntime->fej9()->canMethodExitEventBeHooked())
+      {
+      RELO_LOG(reloRuntime->reloLogger(), 6, "\tinlinedSiteCanBeActivated: target may need enter/exit tracing so disabling inline site\n");
+      return false;
+      }
+
+   return true;
+   }
 
 bool
 TR_RelocationRecordInlinedMethod::inlinedSiteValid(TR_RelocationRuntime *reloRuntime, TR_RelocationTarget *reloTarget, TR_OpaqueMethodBlock **theMethod)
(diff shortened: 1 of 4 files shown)


dsouzai:
fyi @aviansie-ben

fjeremic:
At a quick glance the PR title does not match what the commit is doing.

dsouzai:
Heh yeah it's not immediately obvious, but i still have to write up proper commit messages, as well as a proper PR description (hence the WIP).

dsouzai:

runtime/compiler/runtime/RelocationRecord.cpp
```diff
@@ -2277,27 +2293,22 @@ TR_RelocationRecordInlinedMethod::inlinedSiteValid(TR_RelocationRuntime *reloRun
             }
          else
             {
-            inlinedSiteIsValid = false;
```
jdmpapin: Is there any particular significance in moving this line below the assertion? Just curious; it looks harmless

runtime/compiler/runtime/RelocationRecord.cpp
```diff
@@ -2698,66 +2709,75 @@ TR_RelocationRecordProfiledInlinedMethod::preparePrivateData(TR_RelocationRuntim
...
-         RELO_LOG(reloRuntime->reloLogger(), 6,"\tpreparePrivateData: inlined class valid\n");
-         reloPrivateData->_inlinedCodeClass = inlinedCodeClass;
-         uintptrj_t *chainData = (uintptrj_t *) reloRuntime->fej9()->sharedCache()->pointerFromOffsetInSharedCache((void *) classChainForInlinedMethod(reloTarget));
-         if (reloRuntime->fej9()->sharedCache()->classMatchesCachedVersion(inlinedCodeClass, chainData))
-            {
-            RELO_LOG(reloRuntime->reloLogger(), 6,"\tpreparePrivateData: classes match\n");
-            TR_OpaqueMethodBlock *inlinedMethod = * (TR_OpaqueMethodBlock **) (((uint8_t *)reloPrivateData->_inlinedCodeClass) + vTableSlot(reloTarget));
+         inlinedMethod = * (TR_OpaqueMethodBlock **) (((uint8_t *)reloPrivateData->_inlinedCodeClass) + vTableSlot(reloTarget));
+         if (!inlinedMethod)
+            inlinedSiteIsValid = false;
```
jdmpapin: I don't think it's possible for the vtable slot to be null

runtime/compiler/runtime/RelocationRecord.cpp
```diff
@@ -2698,66 +2709,75 @@ TR_RelocationRecordProfiledInlinedMethod::preparePrivateData(TR_RelocationRuntim
...
+         setupInlinedMethodData(reloRuntime, reloTarget);
          }
+
+      /* Even if the inlined site is disabled, the inlined site table in the metadata
+       * should still be populated under AOT w/ SVM
+       */
+      if (inlinedMethod)
+         fixInlinedSiteInfo(reloRuntime, reloTarget, inlinedMethod);
+      else if (reloRuntime->comp()->getOption(TR_UseSymbolValidationManager))
+         SVM_ASSERT(inlinedMethod != NULL, "inlinedMethod should not be NULL when using the SVM!");
```
jdmpapin: :+1:

runtime/compiler/x/codegen/J9AheadOfTimeCompile.cpp
```diff
@@ -294,6 +294,7 @@ uint8_t *J9::X86::AheadOfTimeCompile::initializeAOTRelocationHeader(TR::Iterated
       case TR_ProfiledMethodGuardRelocation :
          {
          guard = (TR_VirtualGuard *)relocation->getTargetAddress2();
+         TR_OpaqueClassBlock *inlinedCodeClass = guard->getThisClass();
```
jdmpapin: This is the profiled class, yes? Not the defining class of the method, which may have been inherited

runtime/compiler/runtime/RelocationRecord.cpp
dsouzai: Initially, I had moved the code that checks `compileRomClass == currentRomClass` into `inlinedSiteCanBeActivated` at whcih point the `inlinedSiteIsValid = false;` got changed to a `return false`, and so had to be moved after the check for SVM. It turns out commoning that code to also be used for profiled inlined methods didn't work (because the rom class stored in the relorecord was not the rom class of the method that was inlined), so I had to un-refactor the code. I guess that resulted in the `inlinedSiteIsValid = false` ended up after the if :).

runtime/compiler/x/codegen/J9AheadOfTimeCompile.cpp
dsouzai: Yeah that's right, we don't store the romclass of the defining class of the method.

runtime/compiler/runtime/RelocationRecord.cpp
dsouzai: I see, do you think it's fine as is, or would you prefer I rebase with this check removed?

jdmpapin: I'd ask for it to be changed if there was something else to necessitate a round of updates, but I think it can be left alone. It doesn't really hurt

dsouzai:
@fjeremic do you mind shepherding this PR (since you're already here :P)

dsouzai:
jenkins test sanity xlinux,win,plinux,zlinux,jdk8,jdk11

dsouzai:

dsouzai:

PR #4651, opened 2019-02-07
Add ValueType support to the if_acmp bytecodes

MarkQingGuo:
Modify behaviour of Bytecodes if_acmps 

we need to modify the behaviour of these two bytecodes when value types are involved:
- the result of if_acmpeq is true when these two valuetypes are substitutable and false otherwise
- the result of if_acmpne is true when these two valuetypes are not substitutable and false otherwise
- others remain the same for normal objects
 
### DIFF
--- a/runtime/vm/BytecodeInterpreter.hpp
+++ b/runtime/vm/BytecodeInterpreter.hpp
@@ -7183,6 +7183,56 @@ done:
 		return rc;
 	}
 
+	/*
+	 * Determine if the two objects are substitutable
+	 *
+	 * @param[in] lhs the lhs object of acmp bytecodes
+	 * @param[in] rhs the rhs object of acmp bytecodes
+	 * return true if they are substituable and false otherwise
+	 */
+	VMINLINE bool
+	acmp(j9object_t lhs, j9object_t rhs)
+	{
+#if defined(J9VM_OPT_VALHALLA_VALUE_TYPES)
+		bool acmpResult = false;
+		if (rhs == lhs) {
+			acmpResult = true;
+		} else if ((NULL == rhs) || (NULL == lhs)) {
+			acmpResult = false;
+		} else {
+			J9Class * lhsClass = J9OBJECT_CLAZZ(_currentThread, lhs);
+			J9Class * rhsClass = J9OBJECT_CLAZZ(_currentThread, rhs);
+			if ((J9_IS_J9CLASS_VALUETYPE(rhsClass)
+				&& J9_IS_J9CLASS_VALUETYPE(lhsClass))
+				&& (rhsClass == lhsClass)
+			) {
+				acmpResult = isSubstitutable(lhs, rhs);
+			}
+		}
+		return acmpResult;
+#else /* J9VM_OPT_VALHALLA_VALUE_TYPES */
+		return (rhs == lhs);
+#endif /* J9VM_OPT_VALHALLA_VALUE_TYPES */
+	}
+
+#if defined(J9VM_OPT_VALHALLA_VALUE_TYPES)
+	/*
+	 * Determine if the two valueTypes are substitutable when rhs.class equals lhs.class
+	 *
+	 * @param[in] lhs the lhs object of acmp bytecodes and it's a valueType
+	 * @param[in] rhs the rhs object of acmp bytecodes and it's a valueType
+	 * return true if they are substitutable and false otherwise
+	 */
+	VMINLINE bool
+	isSubstitutable(j9object_t lhs, j9object_t rhs)
+	{
+		/*
+		 * TODO: this will be updated in a future PR.
+		 */
+		return false;
+	}
+#endif /* J9VM_OPT_VALHALLA_VALUE_TYPES */
+
 	/* ..., lhs, rhs => ... */
 	VMINLINE VM_BytecodeAction
 	ifacmpeq(REGISTER_ARGS_LIST)
@@ -7192,7 +7242,7 @@ done:
 		j9object_t lhs = *(j9object_t*)(_sp + 1);
 		U_8 *profilingCursor = startProfilingRecord(REGISTER_ARGS, sizeof(U_8));
 		_sp += 2;
-		if (lhs == rhs) {
+		if(acmp(lhs, rhs)) {
 			_pc += *(I_16*)(_pc + 1);
 			if (NULL != profilingCursor) {
 				*profilingCursor = 1;
@@ -7216,7 +7266,7 @@ done:
 		j9object_t lhs = *(j9object_t*)(_sp + 1);
 		U_8 *profilingCursor = startProfilingRecord(REGISTER_ARGS, sizeof(U_8));
 		_sp += 2;
-		if (lhs != rhs) {
+		if(!acmp(lhs, rhs)) {
 			_pc += *(I_16*)(_pc + 1);
 			if (NULL != profilingCursor) {
 				*profilingCursor = 1;
--- a/test/functional/Valhalla/src/org/openj9/test/lworld/ValueTypeTests.java
+++ b/test/functional/Valhalla/src/org/openj9/test/lworld/ValueTypeTests.java
@@ -362,7 +362,38 @@ public class ValueTypeTests {
 			Assert.fail("should throw error. Class does not exist");
 		} catch (NoClassDefFoundError e) {}
 	}
-	
+
+	/*
+	 * Test ifacmp on value class
+	 * 
+	 * class TestIfacmpOnValueClass {}
+	 */
+	@Test(priority=2)
+	static public void TestIfacmpOnValueClass() throws Throwable {
+		int x = 0;
+		int y = 0;
+		
+		Object valueType = makePoint2D.invoke(x, y);
+
+		Object refType = (Object) x;
+		
+		Assert.assertFalse((valueType == refType), "An identity (==) comparison that contains a valueType should always return false");
+		
+		Assert.assertFalse((refType == valueType), "An identity (==) comparison that contains a valueType should always return false");
+
+		Assert.assertFalse((valueType == valueType), "An identity (==) comparison that contains a valueType should always return false");
+
+		Assert.assertTrue((refType == refType), "An identity (==) comparison on the same refType should always return true");
+
+		Assert.assertTrue((valueType != refType), "An identity (!=) comparison that contains a valueType should always return true");
+
+		Assert.assertTrue((refType != valueType), "An identity (!=) comparison that contains a valueType should always return true");
+
+		Assert.assertTrue((valueType != valueType), "An identity (!=) comparison that contains a valueType should always return true");
+
+		Assert.assertFalse((refType != refType), "An identity (!=) comparison on the same refType should always return false");
+	}
+		
 	static MethodHandle generateGetter(Class<?> clazz, String fieldName, Class<?> fieldType) {
 		try {
 			return lookup.findVirtual(clazz, "get"+fieldName, MethodType.methodType(fieldType));


MarkQingGuo:
@tajila please review. please tell me where is wrong if this is not what you want.
The test for correctness is still in progress.

tajila:
I think the logic can be simplified if you create a new inline function `ie. acmp_v2(lhs, rhs)` that does the following:
1) check if lhs != rhs, if so return false
2) check if lhs is null, if so return true
3) check if lhs is a valueType if so return false
4) return true

you if_acmpeq code can use `acmp_v2` and the if_acmpneq can use `!acmp_v2`

MarkQingGuo:
I have changed the code, please review @tajila

runtime/vm/BytecodeInterpreter.hpp
```diff
@@ -7107,6 +7107,19 @@ done:;
 		return rc;
 	}
 
+#if defined(J9VM_OPT_VALHALLA_VALUE_TYPES)
+	VMINLINE bool
+	acmp(j9object_t lhs, j9object_t rhs) 
+	{
+		if(lhs != rhs) return false;
```
tajila: please dont use inline if statements. See https://github.com/eclipse/omr/blob/master/doc/CodingStandard.md

tajila: Also, only have a single return statement in a function, use a temp variable if you have to.

runtime/vm/BytecodeInterpreter.hpp
```diff
@@ -7107,6 +7107,19 @@ done:;
...
+		if(lhs == NULL) return true;
+		J9Class * lhsClass = J9OBJECT_CLAZZ(_currentThread, lhs);
+		if(J9_IS_J9CLASS_VALUETYPE(lhsClass)) return false;
+		return true;
+		}
```
tajila: Pay attention to the indentation here

runtime/vm/BytecodeInterpreter.hpp
```diff
@@ -7116,6 +7129,20 @@ done:;
 		j9object_t lhs = *(j9object_t*)(_sp + 1);
 		U_8 *profilingCursor = startProfilingRecord(REGISTER_ARGS, sizeof(U_8));
 		_sp += 2;
+#if defined(J9VM_OPT_VALHALLA_VALUE_TYPES)
```
tajila: put the ifdefs in the acmp help, not the existing ifacmp* bytecodes

test/functional/Valhalla/src/org/openj9/test/lworld/ValueTypeTests.java
```diff
@@ -362,7 +364,47 @@ static public void testWithFieldOnNonExistentClass() throws Throwable {
...
+	 * Test ifacmp on value class
+	 * 
+	 * class TestIfacmpOnValueClass {
+	 *  
+	 * }
+	 */
+@Test(priority=1)
+  	static public void TestIfacmpOnValueClass() throws Throwable {
+    		String fields[] = {"IntField:I"};
+    		Class<?> acmpClass = ValueTypeGenerator.generateValueClass("acmpClass", fields);
```
tajila: You dont need to make a new class, you can use one of the existing ones, like Point2D

MarkQingGuo:
@tajila please review, 
- add some comments and change indentation: It seems my old vim configuration makes the tab key equal 2 spaces
- use a temporary variable in amp() function and use macros inside it: there are a little duplicate code but I think it makes it more readable
- use existing Point2D class

runtime/vm/BytecodeInterpreter.hpp
```diff
@@ -7107,6 +7107,41 @@ done:;
...
+            /* two objects are null */
+            acmpResult = true;
+        } else {
+            /* at least one object is value type */
+            J9Class * lhsClass = J9OBJECT_CLAZZ(_currentThread, lhs);
+            if (J9_IS_J9CLASS_VALUETYPE(lhsClass)) {
+                acmpResult = false;
+            }
+        }
+#else /* J9VM_OPT_VALHALLA_VALUE_TYPES */
```
tajila: the code in the else can simply do `return rhs == lhs`, but the VT code shuold have one return point

runtime/vm/BytecodeInterpreter.hpp
```diff
@@ -7107,6 +7107,41 @@ done:;
...
+/*
+ * Determine the equality of two objects
+ * @param[in] lhs the lhs object of acmp bytecodes
+ * @param[in] rhs the rhs object of acmp bytecodes
+ * return true if they are equal and false otherwise
+ */
+	VMINLINE bool
+	acmp(j9object_t lhs, j9object_t rhs) 
+	{
+        bool acmpResult = true;
```
tajila: please put this variable in the ifdef

test/functional/Valhalla/src/org/openj9/test/lworld/ValueTypeTests.java
```diff
@@ -362,7 +362,53 @@ static public void testWithFieldOnNonExistentClass() throws Throwable {
...
+	@Test(priority=1)
+	static public void TestIfacmpOnValueClass() throws Throwable {
+		String fields[] = {"x:I", "y:I"};
+		point2DClass = ValueTypeGenerator.generateValueClass("Point2DInAcmp", fields);
+		makePoint2D = lookup.findStatic(point2DClass, "makeValue", MethodType.methodType(point2DClass, int.class, int.class));
+
+		int x = 0;
+		int y = 0;
+		
+		Object point2D = makePoint2D.invoke(x, y);
```
tajila: you can call point2D valueType

test/functional/Valhalla/src/org/openj9/test/lworld/ValueTypeTests.java
```diff
@@ -362,7 +362,53 @@ static public void testWithFieldOnNonExistentClass() throws Throwable {
...
+
+		Object normalClass = (Object) x;
```
tajila: you can this one refType

test/functional/Valhalla/src/org/openj9/test/lworld/ValueTypeTests.java
```diff
@@ -362,7 +362,53 @@ static public void testWithFieldOnNonExistentClass() throws Throwable {
...
+		/*
```
tajila: you dont need this comment

test/functional/Valhalla/src/org/openj9/test/lworld/ValueTypeTests.java
```diff
@@ -362,7 +362,53 @@ static public void testWithFieldOnNonExistentClass() throws Throwable {
...
+         * Test if bytecode if_acmpne behaves correctly
+         */
+		if (point2D == normalClass) {
+		    Assert.fail("should throw error. Acmpne passes one valuetype and one normal object!");
+		}
+		if (point2D == point2D) {
+            Assert.fail("should throw error. Acmpne passes two normal objects!");
```
tajila: fix indentation

test/functional/Valhalla/src/org/openj9/test/lworld/ValueTypeTests.java
```diff
@@ -362,7 +362,53 @@ static public void testWithFieldOnNonExistentClass() throws Throwable {
...
+		int y = 0;
+		
+		Object point2D = makePoint2D.invoke(x, y);
+
+		Object normalClass = (Object) x;
+		/*
+         * Test if bytecode if_acmpne behaves correctly
+         */
+		if (point2D == normalClass) {
+		    Assert.fail("should throw error. Acmpne passes one valuetype and one normal object!");
```
tajila: you can modify comment to say, "Equality comparisons with valueTypes should always return false"

test/functional/Valhalla/src/org/openj9/test/lworld/ValueTypeTests.java
tajila: same here with the failure message

test/functional/Valhalla/src/org/openj9/test/lworld/ValueTypeTests.java
```diff
@@ -362,7 +362,53 @@ static public void testWithFieldOnNonExistentClass() throws Throwable {
...
 		} catch (NoClassDefFoundError e) {}
 	}
-	
+
+	/*
+	 * Test ifacmp on value class
+	 * 
+	 * class TestIfacmpOnValueClass {
+	 *  
+	 * }
```
tajila: I think you need to test the following
1) VT with VT ==
2) VT with VT !=
3) VT with Ref ==
4) VT with Ref !=
5) Ref with VT ==
6) Ref with VT !=
7) Ref with Ref ==
8) Ref with Ref !=

MarkQingGuo: I write these tests as conditional statements in this function, do you mean I have to spilt it into functions respectively?

MarkQingGuo:
@tajila please review, 
- change indentation, I use IDE this time and indentation is fixed by using it.
- move local variable inside #ifdefs
- change names in ValueTypesTest.java and rearrange the tests

runtime/vm/BytecodeInterpreter.hpp
```diff
@@ -7107,6 +7107,37 @@ done:;
...
+ * @param[in] rhs the rhs object of acmp bytecodes
+ * return true if they are equal and false otherwise
+ */
+	VMINLINE bool
+	acmp(j9object_t lhs, j9object_t rhs) 
+	{
+#if defined(J9VM_OPT_VALHALLA_VALUE_TYPES)
+		bool acmpResult = true;
+		if (rhs != lhs) {
+		/* two objects are not equal */
```
tajila: the comment is not needed

runtime/vm/BytecodeInterpreter.hpp
```diff
@@ -7107,6 +7107,37 @@ done:;
...
+			acmpResult = false;
+		} else if (NULL == lhs) {
+			/* two objects are null */
```
tajila: same here

runtime/vm/BytecodeInterpreter.hpp
```diff
@@ -7107,6 +7107,37 @@ done:;
...
+			acmpResult = true;
+		} else {
+		/* at least one object is value type */
```
tajila: and here

test/functional/Valhalla/src/org/openj9/test/lworld/ValueTypeTests.java
```diff
@@ -362,7 +362,60 @@ static public void testWithFieldOnNonExistentClass() throws Throwable {
...
+
+		int x = 0;
+		int y = 0;
+		
+		Object valueType = makePoint2D.invoke(x, y);
+
+		Object refType = (Object) x;
+		
+		if (valueType == refType) {
+			Assert.fail("should throw error. Acmpne passes one valueType and one refType!");
```
tajila: `should throw error` is misleading.

It doesnt throw an error. Its probably simpler if you dont mention acmp in the test. You can simply say:

"An identity (==) comparison that contains a valueType should always return false" and vice versa for the "!=" case

MarkQingGuo:
@tajila  please review.
- change comments

test/functional/Valhalla/src/org/openj9/test/lworld/ValueTypeTests.java
```diff
@@ -362,7 +362,60 @@ static public void testWithFieldOnNonExistentClass() throws Throwable {
...
-	
+
+	/*
+	 * Test ifacmp on value class
+	 * 
+	 * class TestIfacmpOnValueClass {
+	 *  
+	 * }
+	 */
+	@Test(priority=1)
```
tajila: Where you able to run this test? I think this has to be priority 2 because testCreatePoint2D is priority 1

runtime/vm/BytecodeInterpreter.hpp
```diff
@@ -7107,6 +7107,34 @@ done:;
 		return rc;
 	}
 
+/*
```
tajila: this comment needs to have the same indentation as the function

test/functional/Valhalla/src/org/openj9/test/lworld/ValueTypeTests.java
```diff
@@ -362,7 +362,60 @@ static public void testWithFieldOnNonExistentClass() throws Throwable {
...
+		
+		if (valueType == valueType) {
+			Assert.fail("An identity (==) comparison that contains a valueType should always return false");
+		}
+		
+		if (!(refType == refType)) {
+			Assert.fail("An identity (==) comparison on the same refType should always return true");
+		}
+        
+		if (!(valueType != refType)) {
```
tajila: you need to double check that javac will not optimize `(!(valueType != refType))` into `(valueType == refType)`

test/functional/Valhalla/src/org/openj9/test/lworld/ValueTypeTests.java
MarkQingGuo: yes, because I actually create a new value type here, but I will test it again and tell you the result later

MarkQingGuo: I have retested it and my test is able to run

test/functional/Valhalla/src/org/openj9/test/lworld/ValueTypeTests.java
MarkQingGuo: ``if (false == (valueType != refType))``
Does java optimize this code? I think it may not.

test/functional/Valhalla/src/org/openj9/test/lworld/ValueTypeTests.java
tajila: `point2DClass = ValueTypeGenerator.generateValueClass("Point2DInAcmp", fields);` this overwrites a shared global that is used by other test cases. Its not a problem now, but may be in the future.

You can simply change the priority to 2, and use the global `makePoint2D` instead of re-assigning here

MarkQingGuo: This has been resolved.

test/functional/Valhalla/src/org/openj9/test/lworld/ValueTypeTests.java
tajila: you can check by using javap with the -c option to see what bytecodes the javac compiler generated

tajila:
can you also please change the title of the PR to say "Add ValueType support to the if_acmp bytecodes". And also make a similar change to the commit message. 
  
Dont forget the guidlines for writing commit messages regarding the length of the title (https://github.com/eclipse/openj9/blob/master/CONTRIBUTING.md).

MarkQingGuo:
@tajila 
- edit commit message
- use Assert.assertTrue() and Assert.assertFail() to test the result of operands directly so as to avoid optimization

runtime/vm/BytecodeInterpreter.hpp
```diff
@@ -7183,6 +7183,52 @@ done:;
...
+#if defined(J9VM_OPT_VALHALLA_VALUE_TYPES)
+	/*
+	 * Determine the equality of two valueTypes
+	 * @param[in] lhs the lhs object of acmp bytecodes and it's a valueType
+	 * @param[in] rhs the rhs object of acmp bytecodes and it's a valueType
+	 * return true if they are equal and false otherwise
+	 */
+	VMINLINE bool
+	isSubstitutable(j9object_t lhs, j9object_t rhs)
+	{
```
tajila: add a comment saying 

```
/*
 * TODO: this will be updated in a future PR.
 */

File: runtime/vm/BytecodeInterpreter.hpp [hunk 21]
```diff
@@ -7183,6 +7183,52 @@ done:;
...
 	}
 
+	/*
+	 * Determine the equality of two objects
+	 * @param[in] lhs the lhs object of acmp bytecodes
+	 * @param[in] rhs the rhs object of acmp bytecodes
+	 * return true if they are equal and false otherwise
+	 */
+	VMINLINE bool
+	acmpSub(j9object_t lhs, j9object_t rhs)
```
tajila: you can keep the name as acmp

runtime/vm/BytecodeInterpreter.hpp
```diff
@@ -7183,6 +7183,52 @@ done:;
...
+#if defined(J9VM_OPT_VALHALLA_VALUE_TYPES)
+		bool acmpResult = false;
+		if (rhs == lhs) {
+			acmpResult = true;
+		} else if ((NULL == lhs) || (NULL == rhs)) {
+			acmpResult = false;
+		} else {
+			J9Class * lhsClass = J9OBJECT_CLAZZ(_currentThread, lhs);
+			J9Class * rhsClass = J9OBJECT_CLAZZ(_currentThread, rhs);
+			if(J9_IS_J9CLASS_VALUETYPE(lhsClass)
```
tajila: needs to be a space between `if` and `(`

tajila: since you have already gotten the class, you can add another contition that says `lhsClass == rhsClass` before checking if they are substitutable

runtime/vm/BytecodeInterpreter.hpp
```diff
@@ -7183,6 +7183,52 @@ done:;
 		return rc;
 	}
 
+	/*
+	 * Determine the equality of two objects
```
tajila: This should be changed to, determine if the two objects are substitutable

tajila:
@MarkQingGuo You need to sign the eclipse agreement again, it was recently updated.

runtime/vm/BytecodeInterpreter.hpp
MarkQingGuo: please review. I added a condition that check if they are substitutable when `lhsClass == rhsClass` and return false directly when `lhsClass != rhsClass`, so function isSubstitutable() will be called with precondition that `lhsClass == rhsClass`.

runtime/vm/BytecodeInterpreter.hpp
```diff
@@ -7183,6 +7183,60 @@ done:;
...
+		}
+		return acmpResult;
+#else /* J9VM_OPT_VALHALLA_VALUE_TYPES */
+		return (rhs == lhs);
+#endif /* J9VM_OPT_VALHALLA_VALUE_TYPES */
+	}
+
+#if defined(J9VM_OPT_VALHALLA_VALUE_TYPES)
+	/*
+	 * Determine if the two valueTypes are substitutable when rhs.class equals lhs.class
```
tajila: Also, put a space after the first line

runtime/vm/BytecodeInterpreter.hpp
```diff
@@ -7183,6 +7183,60 @@ done:;
 		return rc;
 	}
 
+	/*
+	 * Determine if the two objects are substitutable
```
tajila: add a newline after this one

runtime/vm/BytecodeInterpreter.hpp
```diff
@@ -7183,6 +7183,60 @@ done:;
...
+			acmpResult = true;
+		} else if ((NULL == lhs) || (NULL == rhs)) {
+			acmpResult = false;
+		} else {
+			J9Class * lhsClass = J9OBJECT_CLAZZ(_currentThread, lhs);
+			J9Class * rhsClass = J9OBJECT_CLAZZ(_currentThread, rhs);
+			if (J9_IS_J9CLASS_VALUETYPE(lhsClass)
+					&& J9_IS_J9CLASS_VALUETYPE(rhsClass)
+			) {
+				if (lhsClass == rhsClass) {
```
tajila: you can combine this with the previous if statement

runtime/vm/BytecodeInterpreter.hpp
```diff
@@ -7183,6 +7183,59 @@ done:;
...
+		} else {
+			J9Class * lhsClass = J9OBJECT_CLAZZ(_currentThread, lhs);
+			J9Class * rhsClass = J9OBJECT_CLAZZ(_currentThread, rhs);
+			if ((J9_IS_J9CLASS_VALUETYPE(lhsClass)
+					&& J9_IS_J9CLASS_VALUETYPE(rhsClass))
+					&& (lhsClass == rhsClass)
+			) {
+				acmpResult = isSubstitutable(lhs, rhs);
+			} else {
+				acmpResult = false;
```
tajila: this else case is not required

runtime/vm/BytecodeInterpreter.hpp
```diff
@@ -7183,6 +7183,59 @@ done:;
...
+		bool acmpResult = false;
+		if (rhs == lhs) {
+			acmpResult = true;
+		} else if ((NULL == lhs) || (NULL == rhs)) {
+			acmpResult = false;
+		} else {
+			J9Class * lhsClass = J9OBJECT_CLAZZ(_currentThread, lhs);
+			J9Class * rhsClass = J9OBJECT_CLAZZ(_currentThread, rhs);
+			if ((J9_IS_J9CLASS_VALUETYPE(lhsClass)
+					&& J9_IS_J9CLASS_VALUETYPE(rhsClass))
```
tajila: i think you need one less tab here

tajila: and below

MarkQingGuo:
- remove unnecessary else case
- remove extra tab: it seems ide add an extra tab automatically in that case
- change PR description

tajila:
@gacholio please review.

runtime/vm/BytecodeInterpreter.hpp
```diff
@@ -7183,6 +7183,56 @@ done:;
...
+	 */
+	VMINLINE bool
+	acmp(j9object_t lhs, j9object_t rhs)
+	{
+#if defined(J9VM_OPT_VALHALLA_VALUE_TYPES)
+		bool acmpResult = false;
+		if (rhs == lhs) {
+			acmpResult = true;
+		} else if ((NULL == rhs) || (NULL == lhs)) {
+			acmpResult = false;
```
gacholio: This could be optimized given the initial value is false, but I'm not sure how much value there is there (I really hope the compiler would just do the right thing here).

gacholio:

Issue #4401, opened 2019-01-23
Generate AOT code at JITaaS server


mpirvu:
This issue proposes to allow the JITaaS server to perform AOT compilations, but the AOT code is going to be stored at the client where the shared class cache resides.

A few infrastructure changes need to be implemented:
1) Extend `TR_J9ServerVM` with `TR_J9SharedCacheServerVM` and teach the new frontend how to find the relevant pieces of information remotely. This means that we have to reimplement all the methods in `TR_J9SharedCacheVM`
2) Extend `TR_ResolvedJ9JITaaSServerMethod` with `TR_ResolvedRelocatableJ9JITaaSServerMethod` and reimplement all the methods in `TR_ResolvedRelocatableJ9Method` teaching them how to fetch relevant information from the remote client
3) Extend `TR_RelocationRuntime` with `TR_SharedCacheJITaaSRelocationRuntime` and reimplement all the methods in `TR_SharedCacheRelocationRuntime`

To implement all this incrementally I propose to start by making a simple AOT compilation work (like a method that just return 3).

mpirvu:
Will need to look into `classChain = fej9->sharedCache()->rememberClass(definingClass);`

dsouzai:
It might almost be worth extending the shared cache object to make a network call.

dsouzai:
It might be worth updating the jitaas branch before doing a bunch of work implementing the sharedvm queries. For example: 

```
   if (comp->getOption(TR_UseSymbolValidationManager))
      {
      if (comp->getSymbolValidationManager()->verifySymbolHasBeenValidated(static_cast<void *>(classPointer)))
         {
         validated = true;
         }
      else
         {
         TR_ASSERT(false, "Class 0x%p should already be validated\n", classPointer);
         }
      }
```
(https://github.com/eclipse/openj9/pull/4406/files#diff-770fc7473e17fd36db57f82532b90d82R1510)

has been changed to 
https://github.com/eclipse/openj9/blob/fc206904ce524f8575628a8d9a217bf6c1548066/runtime/compiler/env/VMJ9.cpp#L8597-L8601

dsouzai:
Given the remaining PRs in https://github.com/eclipse/openj9/issues/3752 there might be additional changes to the frontend queries (though @jdmpapin can confirm that).

harryyu1994:
[...]

PR #4664, opened 2019-02-08
Implement TR_J9SharedCacheServerVM::getMethodFromName

harryyu1994:
Issue: #4401
[skip ci]


harryyu1994:
The reason is that `getMethodFromName` calls `getSystemClassFromClassName` and `getMethodFromClass` and I think we need the `TR_J9SharedCacheServerVM` version of these two APIs to be called.

harryyu1994:
Or maybe we do get `TR_J9SharedCacheServerVM` version of the two APIs called even using VMJ9's version.

harryyu1994:
Okay we shouldn't redefine them in TR_J9SharedCacheServerVM

harryyu1994:
Change updated.

PR #4666, opened 2019-02-08
Symbol validation manager well-known classes

jdmpapin:
Some system classes are loaded very early and are referred to by most JIT compilations. With this commit the symbol validation manager has a list of such classes, termed "well-known classes." It will store a single copy of their class chain offsets in the shared class cache (SCC). For each AOT-compiled method using the symbol validation manager, the emitted relocation data needs just one offset into the SCC in order to refer to the shared data. This not only deduplicates the offsets; it also allows many records to be eliminated:

- `SystemClassByName` records are never needed for well-known classes,

- `ClassClass` records are now completely unnecessary, and

- `ClassByName` and `ClassFromCP` records for well-known classes are made unnecessary by checking eagerly that the well-known classes are visible to all possible "beholders."

`ClassClass` is removed since it's unneeded

### DIFF
--- a/doc/compiler/aot/SymbolValidationManager.md
+++ b/doc/compiler/aot/SymbolValidationManager.md
@@ -165,6 +165,46 @@ load, these symbols don't need to be validation records stored in
 the SCC. Therefore, the SVM, in its constructor, initializes the maps
 with these IDs and the associated symbols.
 
+### Well-known Classes
+
+Some system classes are loaded very early and are referred to by most
+JIT compilations. The SVM has a list of such classes, termed "well-known
+classes." It will store a single copy of their class chain offsets in
+the SCC. For each AOT-compiled method using the SVM, the emitted
+relocation data needs just one offset into the SCC in order to refer to
+the shared data. This not only deduplicates the offsets; it also allows
+many otherwise necessary records to be eliminated.
+
+In particular, well-known classes never need SystemClassByName,
+ClassByName, or ClassFromCP records. Because these latter two have
+beholders, eliminating them is a bit more difficult than eliminating
+SystemClassByName. In order to make them unnecessary, the SVM checks for
+every class it encounters that the class can see every well-known class,
+failing otherwise. Furthermore, the well-known classes must have names
+that only the bootstrap loader is allowed to define. Then any
+ClassByName or ClassFromCP record naming a well-known class is
+definitely redundant: its beholder can see a class with the given name,
+and that class must have been defined by the bootstrap loader, so it
+must be the same class whose class chain was validated at the beginning
+of the AOT load.
+
+In early compilations, it's possible that not all of the well-known
+classes are available. In order to avoid spurious compilation failures,
+the SVM permits some of the classes to be missing, in which case the
+array of class chain offsets represents only a subset of the possible
+well-known classes. The key under which it is stored specifies the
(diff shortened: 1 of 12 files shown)


runtime/compiler/runtime/SymbolValidationManager.cpp
```diff
@@ -79,6 +87,195 @@ TR::SymbolValidationManager::defineGuaranteedID(void *symbol, TR::SymbolType typ
...
+      "java/lang/StringBuffer",
+      "java/lang/System",
+      "java/lang/ref/Reference",
+      "com/ibm/jit/JITHelpers",
+      };
+
+   unsigned int includedClasses = 0;
+
+   static_assert(
+      sizeof (names) / sizeof (names[0]) == WELL_KNOWN_CLASS_COUNT,
```
dsouzai: :+1:

doc/compiler/aot/SymbolValidationManager.md
```diff
@@ -165,6 +165,46 @@ load, these symbols don't need to be validation records stored in
...
+the SCC. For each AOT-compiled method using the SVM, the emitted
+relocation data needs just one offset into the SCC in order to refer to
+the shared data. This not only deduplicates the offsets; it also allows
+many otherwise necessary records to be eliminated.
+
+In particular, well-known classes never need SystemClassByName,
+ClassByName, or ClassFromCP records. Because these latter two have
+beholders, eliminating them is a bit more difficult than eliminating
+SystemClassByName. In order to make them unnecessary, the SVM checks for
+every class it encounters that the class can see every well-known class,
```
dsouzai: >the SVM checks for every class it encounters that the class can see every well-known class

This doesn't result in potential load failures (outside of the class not being loaded yet) because the names of these well known classes are only allowed to be loaded by the bootstrap classloader right?

runtime/compiler/x/codegen/J9AheadOfTimeCompile.cpp
```diff
@@ -73,13 +73,37 @@ void J9::X86::AheadOfTimeCompile::processRelocations()
       {
       self()->addToSizeOfAOTRelocations(r->getSizeOfRelocationData());
       }
+
    // now allocate the memory  size of all iterated relocations + the header (total length field)
-   if (self()->getSizeOfAOTRelocations() != 0)
+
+   // Note that when using the SymbolValidationManager, the well-known classes
+   // must be checked even if no explicit records were generated, since they
+   // might be responsible for the lack of records.
```
dsouzai: :+1:

dsouzai:

dsouzai:
fyi @aviansie-ben

doc/compiler/aot/SymbolValidationManager.md
jdmpapin: They'll only be *defined* by the bootstrap loader, but in order for code belonging to other loaders to name these classes, those loaders have to be able to find them (by delegating).

Since we're not allowed to run Java code, it might be possible to fail to find a class even from a loader that *will* delegate and get it at runtime, if that loader hasn't previously been asked about that class. If so, we could fail loads as a result.

However, the entire "well-known classes" technique carries an increased risk of load failure, because it creates dependencies on classes that the compilation hasn't naturally queried for.

PR #4672, opened 2019-02-08
Create generic test process launcher

pdbain-ibm:
TargetManager launches a specific Java application (TargetVM).  Extract the
code which launches the native application and connects to stdin/stdout/stderr.
This is to allow testing of new utilities such as jps
https://github.com/eclipse/openj9/issues/4655.

### DIFF
--- a/test/functional/TestUtilities/src/org/openj9/test/attachAPI/ProcessLauncher.java
+++ b/test/functional/TestUtilities/src/org/openj9/test/attachAPI/ProcessLauncher.java
@@ -1,0 +1,114 @@
+/*******************************************************************************
+ * Copyright (c) 2019, 2019 IBM Corp. and others
+ *
+ * This program and the accompanying materials are made available under
+ * the terms of the Eclipse Public License 2.0 which accompanies this
+ * distribution and is available at https://www.eclipse.org/legal/epl-2.0/
+ * or the Apache License, Version 2.0 which accompanies this distribution and
+ * is available at https://www.apache.org/licenses/LICENSE-2.0.
+ *
+ * This Source Code may also be made available under the following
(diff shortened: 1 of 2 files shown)


smlambert:
This is a utility that launches processes, what is the expectation for use, and for destroying launched processes?

pdbain-ibm:
This is extracted from the existing code to launch and query `TargetVM` processes, generalized to launch executables other than the Java VM. It is  is intended for launching and checking the output of new tools such as `jps`, `jstack`, etc., as well as its existing role launching `TargetVM`.

The mechanism for destroying processes is up to the user: in the case of `TargetVM` used by the atach API tests, the parent process sends a command to `stdin`; in the case of `jps`, the process runs and exits when it completes.

test/functional/TestUtilities/src/org/openj9/test/attachAPI/ProcessLauncher.java
```diff
@@ -0,0 +1,114 @@
...
+	protected BufferedReader targetErrReader;
+	protected String errOutput = ""; //$NON-NLS-1$
+	protected String outOutput = ""; //$NON-NLS-1$
+
+	/**
+	 * Launch a process
+	 * @param argBuffer executable name followed by arguments
+	 * @return Child process or null if error
+	 */
+	public Process launchProcess(List<String> argBuffer) {
```
DanHeidinga: Can this be replaced by `java.lang.ProcessBuilder`?  It seems to do the same kinds of things.

pdbain-ibm: Hmmm.... Good catch.  I will give it a try.

pdbain-ibm:
Not required.

PR #4678, opened 2019-02-10
Revert "Remove duplicate macros between j9nonbuilder.h and j9consts.h"

gacholio:
Reverts eclipse/openj9#4287

DDR compile failures in the internal builds.


keithc-ca:
Thanks, @gacholio, I was just about to do this.

PR #4647, opened 2019-02-07
Create just one JVM shared library

keithc-ca:
* remove j9vm_b150
* remove j9vm_b156
* remove j9vm_jdk11
* remove j9vm_jdk12
* remove util_b156

This requires coordinated changes to
* jdk11 - https://github.com/ibmruntimes/openj9-openjdk-jdk11/pull/128
* jdk12 - https://github.com/ibmruntimes/openj9-openjdk-jdk12/pull/10
* jdk13 - https://github.com/ibmruntimes/openj9-openjdk-jdk/pull/72

PR #4682, opened 2019-02-11
Run jdk11 cmake build instead of jdk9, add jdk8 as well

pshipton:
[ci skip]


pshipton:
jdk9 no longer works since #4647, and we should be testing jdk11 anyway instead of jdk9.

pshipton:
Currently blocking omr acceptance builds from promoting.

buildenv/jenkins/jobs/pipelines/Pipeline-Build-Test-All
```diff
@@ -83,7 +83,7 @@ SPECS = ['aix_ppc-64_cmprssptrs'      : CURRENT_RELEASES,
          'linux_ppc-64_cmprssptrs_le' : CURRENT_RELEASES,
          'linux_x86-64'               : CURRENT_RELEASES,
          'linux_x86-64_cmprssptrs'    : CURRENT_RELEASES,
-         'linux_x86-64_cmprssptrs_cmake' : ['9'],
+         'linux_x86-64_cmprssptrs_cmake' : ['11'],
```
keithc-ca: Perhaps this should be `'8', '11'` now that `cmake` builds work for JDK8 and JDK11?

pshipton: It will take another ~45 min of xlinux machine to to build 8. Not sure if it is a good idea or not, but I've added for now. We can always remove it again later if problematic.

keithc-ca: That sounds fair.

Issue #4597, opened 2019-02-05
JTReg test fail - Linux s390 : java/lang/String/UnicodeCasingTest.java


ben-walsh:
Consistent multiple failures of test https://github.com/ibmruntimes/openj9-openjdk-jdk11/blob/openj9/test/jdk/java/lang/String/UnicodeCasingTest.java are observed on Linux s390 of the form ...

```
Testing on en_NU locale....
toUpperCase() failed.
	Original: 0xFF 
	Got:      0xFF 
	Expected: 0x178 
```

See https://ci.adoptopenjdk.net/view/Test_openjdk/job/openjdk11_j9_openjdktest_s390x_linux/140/testReport/junit/java_lang_String_UnicodeCasingTest/java/UnicodeCasingTest

No such failures are observed on other platforms ...
e.g. Linux PPC ...
https://ci.adoptopenjdk.net/view/Test_openjdk/job/openjdk11_j9_openjdktest_ppc64le_linux/130/testReport/java_lang_String_UnicodeCasingTest/java/UnicodeCasingTest
e.g. Windows ...
https://ci.adoptopenjdk.net/view/Test_openjdk/job/openjdk11_j9_openjdktest_x86-64_windows/130/testReport/java_lang_String_UnicodeCasingTest/java/UnicodeCasingTest

It seems to be all the locale variants of English that are failing, nothing else.

pshipton:
@fjeremic

fjeremic:
Can we get some machine information? Which processor did this test run and fail on? Going to try to reproduce manually if I can find the documentation on how to do so.

fjeremic:
I cannot reproduce UnicodeCasingTest on my z13 using the latest nightly build of JDK11. We'll need more information on the machine this ran on. It is likely a JIT issue given the test runs fine on my machine.

fjeremic:
I can reproduce on an internal grinder. Going to try using the test infrastructure to reproduce locally too if I can. If not I can try to debug via grinders.

ben-walsh:
From the top of this page - https://ci.adoptopenjdk.net/view/Test_openjdk/job/openjdk11_j9_openjdktest_s390x_linux/140 ... the test was executed on test-marist-ubuntu1604-s390x-1. The details for this test machine can be found at https://ci.adoptopenjdk.net/computer/test-marist-ubuntu1604-s390x-1.
I will discuss with @sxa555 to obtain access sufficient to gain the information you require on the machine.

[...]

PR #2318, opened 2018-07-04
 Re-enable toLowerCase/toUpperCase acceleration 

dhong44:
Modify API call so that caseConversionHelper receives
underlying arrays and length of string.

Closes: #1934
Closes: #777

Signed-off-by: Daniel Hong <daniel.hong@live.com>

PR #4688, opened 2019-02-11
Fix vectorized toLowerCase and toUppercase intrinsics on Z

fjeremic:
In #4597 we identified cases on which `toUpperCase` fails on Linux on 
Z. This is due to some faulty logic in the evaluator for these 
intrinsics on Z. Bugs seem to have been introduced in #2318 for certain 
characters. Namely the test in #4597 fails to convert the `\u00FF` 
character. The bug is obvious, since `toUpperCase` of such a character 
cannot be handled and the JIT evaluator determines the invalid 
character range by checking if the character is greater than 0xFF, but 
it should be checking if greater than or equal to.

In addition the compressed String evaluator also fails on 0xB5 as 
determined by the unit tests. Both issues are fixed.

Fixes: #4597

### DIFF
--- a/runtime/compiler/z/codegen/J9TreeEvaluator.cpp
+++ b/runtime/compiler/z/codegen/J9TreeEvaluator.cpp
@@ -640,12 +640,13 @@ TR::Register * caseConversionHelper(TR::Node* node, TR::CodeGenerator* cg, bool 
    generateVRIaInstruction(cg, TR::InstOpCode::VGBM, node, invalidRangeVector, 0, 0);
    generateVRIaInstruction(cg, TR::InstOpCode::VGBM, node, invalidCondVector, 0, 0);
 
-   // Letters a-z (0x61-0x7A) when to upper and A-Z (0x41-0x5A) when to lower
+   // Characters a-z (0x61-0x7A) when to upper and A-Z (0x41-0x5A) when to lower
    generateVRIaInstruction (cg, TR::InstOpCode::VLEIH, node, alphaRangeVector, isToUpper ? 0x617A : 0x415A, 0x0);
    // Characters àáâãäåæçèéêëìíîïðñòóôõö (0xE0-0xF6) when to upper and ÀÁÂÃÄÅÆÇÈÉÊËÌÍÎÏÐÑÒÓÔÕÖ (0xC0-0xD6) when to lower
    generateVRIaInstruction (cg, TR::InstOpCode::VLEIH, node, alphaRangeVector, isToUpper ? 0xE0F6 : 0xC0D6, 0x1);
    // Characters øùúûüýþ (0xF8-0xFE) when to upper and ØÙÚÛÜÝÞ (0xD8-0xDE) when to lower
    generateVRIaInstruction (cg, TR::InstOpCode::VLEIH, node, alphaRangeVector, isToUpper ? 0xF8FE : 0xD8DE, 0X2);
+
    if (!isCompressedString)
       {
       generateVRRaInstruction(cg, TR::InstOpCode::VUPLH, node, alphaRangeVector, alphaRangeVector, 0, 0, 0, 0);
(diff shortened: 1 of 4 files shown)


fjeremic:
I've improved the tests by including an exhaustive list of tests for uppercasing and lowercasing. There were two issues:

1. The previous testing never tested all characters so some conversions were wrong and uncaught
2. The whole point of Test_JITHelpers is to compile the intrinsics and we should be testing with `-Xjit:count=0` as a variation

Both issues will be fixed. There will be a subsequent commit fixing the issues, but before I add it I'd like to kick off sanity testing with the new tests and ensure the failures are caught on the farm. In addition to the problem observed in #4597 we have one additional character we're not converting properly (0xB5). Hopefully both failures show up on the builds.

Jenkins test sanity xlinux,zlinux JDK8 JDK11

fjeremic:
I've improved the tests by including an exhaustive list of tests for uppercasing and lowercasing. There were two issues:

1. The previous testing never tested all characters so some conversions were wrong and uncaught
2. The whole point of Test_JITHelpers is to compile the intrinsics and we should be testing with `-Xjit:count=0` as a variation

Both issues will be fixed. There will be a subsequent commit fixing the issues, but before I add it I'd like to kick off sanity testing with the new tests and ensure the failures are caught on the farm. In addition to the problem observed in #4597 we have one additional character we're not converting properly (0xB5). Hopefully both failures show up on the builds.

Jenkins test sanity xlinux,zlinux JDK8,JDK11

fjeremic:
As expected the tests failed as follows [1]:

```
FAILED: test_toUpperIntrinsicLatin1
java.lang.AssertionError: Failed to correctly uppercase character 0xb5, actual = 0xb5 expected = 0x39c expected [true] but found [false]
	at org.testng.Assert.fail(Assert.java:96)
	at org.testng.Assert.failNotEquals(Assert.java:776)
	at org.testng.Assert.assertTrue(Assert.java:44)
	at org.openj9.test.com.ibm.jit.Test_JITHelpers.test_toUpperIntrinsicLatin1(Test_JITHelpers.java:297)
...
FAILED: test_toUpperIntrinsicUTF16
java.lang.AssertionError: Failed to correctly uppercase character 0xff, actual = 0xff expected = 0x178 expected [true] but found [false]
	at org.testng.Assert.fail(Assert.java:96)
	at org.testng.Assert.failNotEquals(Assert.java:776)
	at org.testng.Assert.assertTrue(Assert.java:44)
	at org.openj9.test.com.ibm.jit.Test_JITHelpers.test_toUpperIntrinsicUTF16(Test_JITHelpers.java:261)
```

The second failure is the one observed in #4597. I'll test my fix locally tomorrow morning and attach it as a new commit to this PR.

[1] https://ci.eclipse.org/openj9/job/PullRequest-Sanity-JDK11-linux_390-64_cmprssptrs-OpenJ9/198/tapResults/

fjeremic:
And the fix is in. I've tested manually the tests pass but for good measure we'll test on the farm as well. Requesting reviews from one of @smlambert / @llxia for the test piece of the PR and one of @NigelYiboYu / @dhong44 for the Z codegen piece.

Jenkins test sanity xlinux,zlinux JDK8,JDK11

runtime/compiler/z/codegen/J9TreeEvaluator.cpp
```diff
@@ -668,36 +669,36 @@ TR::Register * caseConversionHelper(TR::Node* node, TR::CodeGenerator* cg, bool
...
 
-   // Can't toUpper \u00df (capital sharp s) nor \u00b5 (mu) with a simple addition of 0x20
-   // Condition code equal for capital sharp and mu (bit 0=0x80) and greater than (bit 2=0x20) for codes larger than 0xFF
    if (isToUpper)
       {
+      // Can't uppercase \u00DF (capital sharp s) nor \u00B5 (mu) with a simple addition of 0x20 so we do an equality
+      // comparison (bit 0) and greater than or equal comparison (bits 0 and 2) for codes larger than or equal to 0xFF
+      generateVRIaInstruction(cg, TR::InstOpCode::VLEIH, node, invalidRangeVector, 0xDFDF, 0x0);
+      generateVRIaInstruction(cg, TR::InstOpCode::VLEIH, node, invalidRangeVector, 0xB5B5, 0x1);
+      generateVRIaInstruction(cg, TR::InstOpCode::VLEIH, node, invalidRangeVector, 0xFFFF, 0x2);
```
NigelYiboYu: LGTM

dhong44: LGTM

fjeremic:
@andrewcraik @0dvictor FYI we'll need to take a look at these x86 failures as my unit tests have caught what are likely bugs in the x86 `toUpperCase` and `toLowerCase` intrinsics. The output seems correct however so perhaps something else went wrong on the x86 side.

andrewcraik:
per discussion the issues on x86 are that we need to truncate the return value of the helper to be 0 or 1 per recent spec updates, ensure we check the `disableHWAcceleratedStringCaseConv` flag. With that the functional issues should be resolved and a remaining perf item will be to expand the number of characters converted by the Latin1 helpers which are currently limited to 0-127 (eg core ASCII). @0dvictor is having a look and coordinating with @fjeremic directly.

fjeremic:
Jenkins test sanity xlinux,zlinux JDK8,JDK11

fjeremic:
Jenkins test sanity xlinux,zlinux JDK8,JDK11

fjeremic:
All issues have been fixed. This is ready for final review from committers.

0xdaryl:
@andrewcraik, can you give your +1 if you're fine with the x86 changes please?

Issue #4571, opened 2019-02-04
Test edge cases for new Java12 MethodHandles changes


theresa-m:
see https://github.com/eclipse/openj9/pull/3980

depends on VM and test compilation prs https://github.com/eclipse/openj9/pull/3980 and https://github.com/eclipse/openj9/pull/4531

- MethodHandles.filterArgumentsWithCombiner tests

- MethodHandles.foldArgumentsWithCombiner tests

PR #4689, opened 2019-02-11
Test Java12 MethodHandle changes

theresa-m:
Fixes: #4571

fyi @ChengJin01 @DanHeidinga

### DIFF
--- a/test/functional/Jsr292/playlist.xml
+++ b/test/functional/Jsr292/playlist.xml
@@ -103,6 +103,36 @@
 	</test>
 
 	<test>
+		<testCaseName>jsr292Test_jdk12</testCaseName>
+		<variations>
+				<variation>NoOptions</variation>
+				<variation>Mode195</variation>
+		</variations>
+		<command>$(JAVA_COMMAND) $(JVM_OPTIONS) \
+	--add-opens=java.base/java.lang=ALL-UNNAMED \
+	-Djava.security.policy=$(Q)$(TEST_RESROOT)$(D)java.policy$(Q) \
+	-cp $(Q)$(TEST_RESROOT)$(D)jsr292test.jar$(P)$(RESOURCES_DIR)$(P)$(TESTNG)$(P)$(LIB_DIR)$(D)asm-all.jar$(Q) \
+	org.testng.TestNG -d $(REPORTDIR) $(Q)$(TEST_RESROOT)$(D)testng.xml$(Q) \
+	-testnames jsr292Test_Java12up \
+	-groups $(TEST_GROUP) \
+	-excludegroups $(DEFAULT_EXCLUDE); \
+	$(TEST_STATUS)</command>
+		<levels>
+			<level>extended</level>
+		</levels>
+		<groups>
+			<group>functional</group>
+		</groups>
+		<subsets>
+			<subset>12+</subset>
+		</subsets>
+		<impls>
+			<impl>openj9</impl>
+			<impl>ibm</impl>
+		</impls>
+	</test>
+
+	<test>
 		<testCaseName>jsr292Test_JitCount0_jdk8</testCaseName>
 		<variations>
 			<variation>-Xjit:count=0</variation>
--- a/test/functional/Jsr292/src/com/ibm/j9/jsr292/FilterArgumentsWithCombinerTest.java
+++ b/test/functional/Jsr292/src/com/ibm/j9/jsr292/FilterArgumentsWithCombinerTest.java
@@ -1,0 +1,391 @@
+/*******************************************************************************
+ * Copyright (c) 2019, 2019 IBM Corp. and others
+ *
+ * This program and the accompanying materials are made available under
+ * the terms of the Eclipse Public License 2.0 which accompanies this
+ * distribution and is available at https://www.eclipse.org/legal/epl-2.0/
+ * or the Apache License, Version 2.0 which accompanies this distribution and
+ * is available at https://www.apache.org/licenses/LICENSE-2.0.
+ *
+ * This Source Code may also be made available under the following
+ * Secondary Licenses when the conditions for such availability set
+ * forth in the Eclipse Public License, v. 2.0 are satisfied: GNU
+ * General Public License, version 2 with the GNU Classpath
+ * Exception [1] and GNU General Public License, version 2 with the
+ * OpenJDK Assembly Exception [2].
+ *
+ * [1] https://www.gnu.org/software/classpath/license.html
+ * [2] http://openjdk.java.net/legal/assembly-exception.html
+ *
+ * SPDX-License-Identifier: EPL-2.0 OR Apache-2.0 OR GPL-2.0 WITH Classpath-exception-2.0 OR LicenseRef-GPL-2.0 WITH Assembly-exception
+ *******************************************************************************/
+package com.ibm.j9.jsr292;
+
+import org.testng.annotations.Test;
+import org.testng.AssertJUnit;
+
+import java.lang.invoke.MethodHandle;
+import java.lang.invoke.MethodHandles;
+import java.lang.invoke.MethodHandles.Lookup;
+import java.lang.invoke.MethodType;
+import java.lang.invoke.WrongMethodTypeException;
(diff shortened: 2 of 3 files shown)


test/functional/Jsr292/src/com/ibm/j9/jsr292/FilterArgumentsWithCombinerTest.java
```diff
@@ -0,0 +1,435 @@
...
+            fawcMethod.setAccessible(true);
+        } catch(Throwable e) {
+            AssertJUnit.fail("FilterArgumentsWithCombinerTest setup failure" + e.getMessage());
+        }
+    }
+
+    /* test filterArgumentsWithCombiner for data type byte */
+    @Test(groups = { "level.extended" })
+	public void test_filterArgumentsWithCombiner_Byte() throws WrongMethodTypeException, Throwable {
+        MethodHandle result;
```
ChengJin01: Better declare when using it in the test.
e.g. MethodHandle  result = (MethodHandle)fawcMethod.invoke(...)
      result = ...

test/functional/Jsr292/src/com/ibm/j9/jsr292/FilterArgumentsWithCombinerTest.java
```diff
@@ -0,0 +1,435 @@
...
+        result = (MethodHandle)fawcMethod.invoke(mhClazz, target, 0, combiner, new int[]{0, 2}); 
+        AssertJUnit.assertEquals(Byte.MIN_VALUE, (byte)result.invokeExact(constByte1, Byte.MIN_VALUE, constByte2));
+
+        /* verify medium boundary, filter position is last handle parameter */
+        result = (MethodHandle)fawcMethod.invoke(mhClazz, target, 2, combiner, new int[]{1, 2});
+        AssertJUnit.assertEquals(constByte1, (byte)result.invokeExact(Byte.MAX_VALUE, constByte1, constByte2));
+    }
+
+    /* combiner returns highest byte value */
+    public static byte combinerByte(byte v1, byte v2) {
```
ChengJin01: All test methods like this should be moved to a separate file as this file only contains test cases.

test/functional/Jsr292/src/com/ibm/j9/jsr292/FilterArgumentsWithCombinerTest.java
```diff
@@ -0,0 +1,435 @@
...
+    private final static long constLong2 = (long) 1.26663739519795e+018;
+    private final static float constFloat1 = (float) 6.12483306976318e+037F;
+    private final static float constFloat2 = (float) 1.53640056404114e+038F;
+    private final static double constDouble1 = (double) 7.33110759312901e+307;
+    private final static double constDouble2 = (double) 1.10819706189633e+307;
+    private final static boolean constBoolMax = true;
+    private final static boolean constBoolMin = false;
+
+    private final static String MHClassName = "java.lang.invoke.MethodHandles";
+    private static Class<?> mhClazz;
```
DanHeidinga:
```
private static Class<?> mhClazz = MethodHandles.class;`
```

test/functional/Jsr292/src/com/ibm/j9/jsr292/FilterArgumentsWithCombinerTest.java
```diff
@@ -0,0 +1,435 @@
...
+    }
+
+    /* combiner returns highest char value */
+    public static char combinerChar(char v1, char v2) {
+        return (v1 > v2) ? v1 : v2;
+    }
+
+    /* return the lowest value char */
+    public static char targetChar(char v1, char v2, char v3) {
+       if ((v1 <= v2) & (v1 <= v3)) {
```
DanHeidinga: This can be written in fewer comparisons and ends up with simpler to read logic

```java
public static char targetChar(char v1, char v2, char v3) {
  char lowest = v1;
  if (v2 < lowest) {
   lowest = v2;
  }
  if (v3 < lowest) {
   lowest = v3;
  }
  return lowest;
}
```

DanHeidinga: I've found for tests like this, that sometimes it more clear what is happening if the test either returns a string that concats all the arguments into a list or it returns an array of the arguments.  It helps with debugging the tests when there is a problem.

test/functional/Jsr292/playlist.xml
```diff
@@ -102,6 +102,32 @@
 		</subsets>
 	</test>
 
+	<test>
+		<testCaseName>jsr292Test_jdk12</testCaseName>
+		<variations>
+				<variation>NoOptions</variation>
+				<variation>Mode195</variation>
```
DanHeidinga: is Mode195 a JIT count=0 mode?  These tests will need to be run both normally and ensuring they've been jitted.

theresa-m: `-Xjit` is set for this mode

test/functional/Jsr292/src/com/ibm/j9/jsr292/FilterArgumentsWithCombinerTest.java
theresa-m: that definitely seems more straightforward. I will do something your suggestion.

theresa-m:
thanks for the reviews @ChengJin01 and @DanHeidinga I've addressed your comments

test/functional/Jsr292/playlist.xml
```diff
@@ -102,6 +102,32 @@
...
+	$(TEST_STATUS)</command>
+		<levels>
+			<level>extended</level>
+		</levels>
+		<groups>
+			<group>functional</group>
+		</groups>
+		<subsets>
+			<subset>12+</subset>
+		</subsets>
```
llxia: Could you please add impls? Thanks

```
		<impls>
			<impl>openj9</impl>
			<impl>ibm</impl>
		</impls>
```

theresa-m: these changes should not be j9 specific.

smlambert: Ideally try running these tests against a hotspot build, have you been able to verify that?
If so and they run fine, then you can leave out the `<impls>` block, and impl specifiers, so that these tests will get run against all implementations?

theresa-m: the NoOptions run passes with hotspot but the Mode195 test does not.

smlambert: So presumably a hotspot build chokes on the option -XX:RecreateClassfileOnload which [Mode195 contains](https://github.com/eclipse/openj9/blob/master/test/TestConfig/resources/modes.xml#L1979).

theresa-m: makes sense. its added now

llxia: In this case, should we break this test into two? one for NoOptions which is for all impls, and one for Mode195 which is openj9 and ibm specific?

smlambert: I was not going to suggest it, because it has implications for all other targets (as test team already discussed, and have deferred decision), we may want to consider other options than nearly duplicate test targets for the multi-impl test approach.

smlambert: but yes, it could broken it in 2.  

Eventually, we will need to do a pass of all playlists, when we do that, we may not want to break all the test targets into 2 to achieve multi-impl testing.  We may want a rule that says NoOptions gets run on all impls, or some back-end check for options that strips out modes that are not applicable.

DanHeidinga:

theresa-m:
rebased to include build fix

pshipton:
@DanHeidinga can this be merged before the 0.13 code split?

pshipton:
Jenkins test sanity plinux jdk12

pshipton:
Jenkins test extended plinux jdk12

pshipton:
The reviewers have approved and the testing is passing, except for expected sanity failures. Merging.

PR #4694, opened 2019-02-12
Eliminate redundant ClassFromCP/ClassByName records

jdmpapin:
This series reduces the number of records generated for ClassFromCP and ClassByName:
- self-referential ClassFromCP and ClassByName records are unnecessary;
- we don't need both ClassFromCP and ClassByName for the same class, beholder pair; and
- using ClassByName instead of ClassFromCP for fresh classes saves a ClassChain record.

### DIFF
--- a/runtime/compiler/runtime/SymbolValidationManager.cpp
+++ b/runtime/compiler/runtime/SymbolValidationManager.cpp
@@ -631,7 +631,12 @@ TR::SymbolValidationManager::addClassFromCPRecord(TR_OpaqueClassBlock *clazz, J9
    if (recordExists(&byName))
       return true; // already have an equivalent ClassByName
 
-   bool added = addClassRecord(clazz, new (_region) ClassFromCPRecord(clazz, beholder, cpIndex));
+   bool added;
+   if (!isAlreadyValidated(clazz)) // save a ClassChainRecord
+      added = addClassRecordWithChain(new (_region) ClassByNameRecord(clazz, beholder));
+   else
+      added = addClassRecord(clazz, new (_region) ClassFromCPRecord(clazz, beholder, cpIndex));
+
    if (added)
       _classesFromAnyCPIndex.insert(ClassFromAnyCPIndex(clazz, beholder));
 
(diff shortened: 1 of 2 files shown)


runtime/compiler/runtime/SymbolValidationManager.hpp
```diff
@@ -848,6 +849,37 @@ class SymbolValidationManager
...
+
+      ClassFromAnyCPIndex(TR_OpaqueClassBlock *clazz, TR_OpaqueClassBlock *beholder)
+         : clazz(clazz), beholder(beholder) { }
+      };
+
+   struct LessClassFromAnyCPIndex
+      {
+      bool operator()(const ClassFromAnyCPIndex &a, const ClassFromAnyCPIndex &b) const
+         {
+         std::less<TR_OpaqueClassBlock*> lt;
```
dsouzai: Curiosity question: how come you didn't use `LexicalOrder` here?

jdmpapin: I wanted to keep `LexicalOrder` file-local in `SymbolValidationManager.cpp`. It could be useful more generally, but we'd have to figure out where to put it—certainly not in `SymbolValidationManager.hpp`. I suppose we could also move the implementation of `LessClassFromAnyCPIndex::operator()` into the cpp file

runtime/compiler/runtime/SymbolValidationManager.cpp
```diff
@@ -631,7 +631,12 @@ TR::SymbolValidationManager::addClassFromCPRecord(TR_OpaqueClassBlock *clazz, J9
    if (recordExists(&byName))
       return true; // already have an equivalent ClassByName
 
-   bool added = addClassRecord(clazz, new (_region) ClassFromCPRecord(clazz, beholder, cpIndex));
+   bool added;
+   if (!isAlreadyValidated(clazz)) // save a ClassChainRecord
```
dsouzai: The idea here is that 
1. because a ClassFromCP+ClassChain is larger than ClassByName, it is better to generate a ClassByName if we've never generated a ClassChain record. 
2. because ClassFromCP is smaller than ClassByName, it is better to generate a ClassFromCP if we **have** generated a ClassChain record.

Is that right?

runtime/compiler/runtime/SymbolValidationManager.hpp
dsouzai: I see, I think it's fine as is for now; if ever there's a need for another "less" operator, we can then move the `LexicalOrder` class somewhere more appropriate.

runtime/compiler/runtime/SymbolValidationManager.cpp
jdmpapin: Yes, especially since in 2 the ClassFromCP will suppress any later ClassByName that would be equivalent

dsouzai:

PR #4695, opened 2019-02-12
Avoid unnecessary SVM records for field references

jdmpapin:
Certain records generated by the symbol validation manager (SVM) identify the class in which an (instance or static) field was found, based on a constant pool field reference. If the beholder refers to one of its own fields, or to a field of a well-known class, and if it does so by naming the class that declares the field (not a subclass), then no record is necessary. The class reference is unambiguous and it must succeed, and because the class chain of the named class will have been validated, the field must be found again directly in that class.

### DIFF
--- a/runtime/compiler/runtime/SymbolValidationManager.cpp
+++ b/runtime/compiler/runtime/SymbolValidationManager.cpp
@@ -570,6 +570,39 @@ TR::SymbolValidationManager::addMethodRecord(TR::MethodValidationRecord *record)
    }
 
 bool
+TR::SymbolValidationManager::skipFieldRefClassRecord(
+   TR_OpaqueClassBlock *definingClass,
+   TR_OpaqueClassBlock *beholder,
+   uint32_t cpIndex)
+   {
+   // If the beholder refers to one of its own fields, or to a field of a
+   // well-known class, and if it does so by naming the class that declares the
+   // field (not a subclass), then no record is necessary.
+   if (definingClass == beholder || isWellKnownClass(definingClass))
+      {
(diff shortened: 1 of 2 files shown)


runtime/compiler/runtime/SymbolValidationManager.cpp
```diff
@@ -569,6 +569,39 @@ TR::SymbolValidationManager::addMethodRecord(TR::MethodValidationRecord *record)
...
+bool
+TR::SymbolValidationManager::skipFieldRefClassRecord(
+   TR_OpaqueClassBlock *definingClass,
+   TR_OpaqueClassBlock *beholder,
+   uint32_t cpIndex)
+   {
+   // If the beholder refers to one of its own fields, or to a field of a
+   // well-known class, and if it does so by naming the class that declares the
+   // field (not a subclass), then no record is necessary.
+   if (definingClass == beholder || isWellKnownClass(definingClass))
```
dsouzai: If the `definingClass` is equal to the `beholder`, do we need to go through this check? Is it possible for the `beholder` to name a subclass even though the field is declared by the `beholder`?

jdmpapin: Sure:
```
// Foo.java
class Foo {
    int x;
    int getBarX(Bar bar) { return bar.x; } // <---
}

// Bar.java
class Bar extends Foo {}
```

dsouzai:

dsouzai:
zlinux build failed because of https://github.com/eclipse/openj9/issues/4702

dsouzai:

PR #4697, opened 2019-02-12
Add J9Symbol API that returns owning class name of recognized field

yanluo7:
Declare and implement API owningClassNameCharsForRecognizedField that
returns the name chars of the owning class of a given recognized field
symbol.

### DIFF
--- a/runtime/compiler/il/symbol/J9Symbol.cpp
+++ b/runtime/compiler/il/symbol/J9Symbol.cpp
@@ -49,10 +49,6 @@
 #include "infra/Flags.hpp"                     // for flags32_t
 #include "ras/Debug.hpp"                       // for TR_DebugBase
 
-TR::Symbol::RecognizedField
-J9::Symbol::searchRecognizedField(TR::Compilation * comp, TR_ResolvedMethod * owningMethod, int32_t cpIndex, bool isStatic)
-   {
-
     struct F
       {
       TR::Symbol::RecognizedField id;
@@ -166,6 +162,9 @@ J9::Symbol::searchRecognizedField(TR::Compilation * comp, TR_ResolvedMethod * ow
        {TR::Symbol::UnknownField}
       };
 
+TR::Symbol::RecognizedField
+J9::Symbol::searchRecognizedField(TR::Compilation * comp, TR_ResolvedMethod * owningMethod, int32_t cpIndex, bool isStatic)
+   {
    struct FP
       {
       struct F   *fieldInfo;
@@ -182,6 +181,7 @@ J9::Symbol::searchRecognizedField(TR::Compilation * comp, TR_ResolvedMethod * ow
        {0,0,0},{0,0,0},{0,0,0},{0,0,0},{0,0,0},{0,0,0},
        {recognizedFieldName_j, 16, 52}
       };
+
    const char minClassPrefix = 'c';
    const char maxClassPrefix = 'j';
 
@@ -269,6 +269,38 @@ J9::Symbol::getRecognizedField()
       return self()->getRecognizedStaticSymbol()->getRecognizedField();
    else
       return TR::Symbol::UnknownField;
+   }
+
+/**
+ * Return the owning class name of this recognized field.
+ * Return null if this symbol does not have a recognized field.
+ */
+const char *
+J9::Symbol::owningClassNameCharsForRecognizedField(int32_t & length)
+   {
+   TR_ASSERT(isShadow(), "Must be a shadow symbol");
+   TR::Symbol::RecognizedField recognizedField = self()->getRecognizedField();
+   TR_ASSERT(TR::Symbol::UnknownField != recognizedField, "Symbol should have a valid recognized field");
+   for (int i = 0; recognizedFieldName_c[i].id != TR::Symbol::UnknownField; ++i)
+      {
+      F &knownField = recognizedFieldName_c[i];
+      if (knownField.id == recognizedField)
+         {
+         length = knownField.classLen;
+         return knownField.classStr;
+         }
+      }
+   for (int i = 0; recognizedFieldName_j[i].id != TR::Symbol::UnknownField; ++i)
+      {
+      F &knownField = recognizedFieldName_j[i];
+      if (knownField.id == recognizedField)
+         {
+         length = knownField.classLen;
+         return knownField.classStr;
+         }
+      }
+
+   return NULL;
    }
 
 /**
--- a/runtime/compiler/il/symbol/J9Symbol.hpp
+++ b/runtime/compiler/il/symbol/J9Symbol.hpp
@@ -172,6 +172,7 @@ public:
 
    static RecognizedField searchRecognizedField(TR::Compilation *, TR_ResolvedMethod * owningMethodSymbol, int32_t cpIndex, bool isStatic);
    RecognizedField  getRecognizedField();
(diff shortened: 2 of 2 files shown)


yanluo7:
This PR is in support of https://github.com/eclipse/omr/pull/3433, which utilizes the API implemented that safe-checks a known object against its associated recognized field owing class in VP.

andrewcraik:
@yanluo7 what is the merge order requirement for this PR and the OMR PR you mentioned above?

yanluo7:
@andrewcraik Merge order is: this PR, followed by https://github.com/eclipse/omr/pull/3433, followed by https://github.com/eclipse/openj9/pull/4346

runtime/compiler/il/symbol/J9Symbol.cpp
```diff
@@ -182,6 +178,10 @@ J9::Symbol::searchRecognizedField(TR::Compilation * comp, TR_ResolvedMethod * ow
        {0,0,0},{0,0,0},{0,0,0},{0,0,0},{0,0,0},{0,0,0},
        {recognizedFieldName_j, 16, 52}
       };
+
+TR::Symbol::RecognizedField
```
andrewcraik: why did this move? some comment related to this is warranted in the commit mesage

runtime/compiler/il/symbol/J9Symbol.cpp
```diff
@@ -271,6 +271,44 @@ J9::Symbol::getRecognizedField()
...
    }
 
+/**
+ * Return the owning class name of this recognized field.
+ * Return null if this symbol does not have a recognized field.
+ */
+const char *
+J9::Symbol::owningClassNameCharsForRecognizedField(int32_t & length)
+   {
+   length = 0;
```
andrewcraik: why always clobber length if the real check is supposed to be for a NULL return?

runtime/compiler/il/symbol/J9Symbol.cpp
```diff
@@ -271,6 +271,44 @@ J9::Symbol::getRecognizedField()
...
+   TR::Symbol::RecognizedField recognizedField = getRecognizedField();
+   if (TR::Symbol::UnknownField == recognizedField)
+      {
+      return NULL;
+      }
+   int i;
```
andrewcraik: why is `i` not scoped to the loops? I don't see that it should carry across the loops...

runtime/compiler/il/symbol/J9Symbol.cpp
```diff
@@ -271,6 +271,44 @@ J9::Symbol::getRecognizedField()
...
+   F *knownField;
```
andrewcraik: same why is this not loop scoped?

runtime/compiler/il/symbol/J9Symbol.cpp
```diff
@@ -271,6 +271,44 @@ J9::Symbol::getRecognizedField()
...
+   for (i = 0, knownField = &recognizedFieldName_c[i];
+         knownField->id != TR::Symbol::UnknownField;
```
andrewcraik: is it always safe to deref knownField?

runtime/compiler/il/symbol/J9Symbol.cpp
```diff
@@ -271,6 +271,44 @@ J9::Symbol::getRecognizedField()
       return TR::Symbol::UnknownField;
    }
 
+/**
+ * Return the owning class name of this recognized field.
+ * Return null if this symbol does not have a recognized field.
+ */
+const char *
+J9::Symbol::owningClassNameCharsForRecognizedField(int32_t & length)
```
andrewcraik: Is there some kind of assert to make sure that we are calling this on a field of some kind? is there any kind of assertion we should have to make sure we find the recognizedField for something that should be recognized?

runtime/compiler/il/symbol/J9Symbol.cpp
yanluo7: The static recognized field tables were scoped inside a function that uses them. I moved them outside the function so that new API implementation can access them.

runtime/compiler/il/symbol/J9Symbol.cpp
yanluo7: Removed this line.

runtime/compiler/il/symbol/J9Symbol.cpp
yanluo7: I tried to initialize both variables in the loop condition, but compiler complained. This is OK if both variables are of same type, but there seems to be a limitation if they are different types.

runtime/compiler/il/symbol/J9Symbol.cpp
yanluo7: Please see last comment.

runtime/compiler/il/symbol/J9Symbol.cpp
yanluo7: I believe it is safe as the last table entry struct `{TR::Symbol::UnknownField}`, always has the first `id` field. As soon as `TR::Symbol::UnknownField` is read from `id` field, loop terminates and no other field is read from this entry,

runtime/compiler/il/symbol/J9Symbol.cpp
yanluo7: Added assertion to make sure the receiver symbol is 1) a shadow symbol, and 2) has a valid recognized field.

runtime/compiler/il/symbol/J9Symbol.cpp
yanluo7: Also declare just one of them inside the loop doesn't work either.
`for (int i =0, knownField = &recognizedFieldName_c[i]; ...`
As compiler thinks `knownField` is also an int.

andrewcraik: There are a few alternative choices - probably the best is to do something like:
```
for (int i = 0; recognizedFieldName_c[i].id != TR::Symbol::UnknownField; ++i) 
     {
     if (recognizedFieldName_c[i].id == recognizedField) {
        {
        ...
        return ...
        }
     }
```
The variables crossing scopes is bad form for maintenance. You could create the knownField temp within the loop body to simplify the code there and that would only be one load and check in the loop header and an additional load in the loop which is worth it for the maintenance benefits IMO.

runtime/compiler/il/symbol/J9Symbol.cpp
andrewcraik: The deref would be made obviously correct by using a referenece type rather than a pointer for `knownField`

runtime/compiler/il/symbol/J9Symbol.cpp
yanluo7: Simplified the loop conditions.

runtime/compiler/il/symbol/J9Symbol.cpp
yanluo7: Using a reference type.

yanluo7:
@andrewcraik All review comments are addressed. Can this be merged?

andrewcraik:

andrewcraik:
I'm happy as soon as sanity is complete - I have marked my approval.

yanluo7:
The linux_ppc-64_cmprssptrs_le JDK11 Sanity failure appears to be infrastructure issue, not problem in the code.

andrewcraik:
Agreed - good to merge.

PR #4706, opened 2019-02-12
Add DigestLoop test class

### LINKED ISSUE #4686: Suggestion: New edge test case for Native Digest
Here is a test that reproduces the crash in https://github.com/eclipse/openj9/issues/4530 and https://github.com/ibmruntimes/openj9-openjdk-jdk11/issues/125 

The following code attempts to perform different digest operations at the same time to the same digest object. 
This causes segfault when nativeDigest is enabled.
(When nativeDigest is disabled, the code would throw exceptions as expected due to digest object not being thread safe)

The test should pass with nativeDigest enabled after these changes: https://github.com/ibmruntimes/openj9-openjdk-jdk8/pull/268 and https://github.com/ibmruntimes/openj9-openjdk-jdk11/pull/132

```
$ cat DigestLoop.java

import java.security.MessageDigest;
import java.security.NoSuchAlgorithmException;
import java.security.NoSuchProviderException;
import java.util.Arrays;
import java.util.Random;

public final class DigestLoop extends Thread {

    static final int     len = 16;
    static final String  alg = "SHA-1";

    static byte[]  data;
    static MessageDigest md;
    static java.security.Provider java_provider;
    public static void main(String args[]) throws InterruptedException
    {
        data = new byte[len];
        Random r = new Random(10);
        r.nextBytes(data);

        try {

            java_provider = java.security.Security.getProvider("SUN");

            md = MessageDigest.getInstance(alg, java_provider);
            System.out.println("Using Provider " + md.getProvider().getName());
            System.out.println("Payload size: "+ data.length + " bytes");
        } catch (Exception e) {
            e.printStackTrace();
        }
        int num_threads=20;

        for (int i = 0 ; i < num_threads ; i++){
            (new DigestLoop()).start();
        }

    }
    public void run() {
        try {
            my_loop(1000000);
        } catch (Exception e){
            e.printStackTrace();
        }
    }

    protected long my_loop(long numIterations) throws InterruptedException {

        Thread currentThread = Thread.currentThread();
        byte[] result = null;
        Random r2 = new Random(currentThread.getId());
        int switch_int;
        for (int i = 0; i < numIterations; i++) {
            int a = r2.nextInt();
            switch_int = ((a >= 0) ? a : -a) % 6;
            try{
                if (switch_int == 1) {
                    //md = MessageDigest.getInstance(alg, java_provider);
                } else if (switch_int == 2) {
                    md.reset();
                } else if (switch_int == 3) {
                    result = md.digest(data);
                } else if (switch_int == 4) {
                     md.clone();
                } else if (switch_int == 5) {
                     md = (MessageDigest) md.clone();
                }
            }catch (Exception e){
                e.printStackTrace();
            }
        }

        return numIterations;
    }
}
```

smlambert:
Fixes #4686 
[skip ci]
Signed-off-by: smlambert <slambert@gmail.com>

### DIFF
--- a/test/functional/Java8andUp/src/org/openj9/test/java/security/DigestLoop.java
+++ b/test/functional/Java8andUp/src/org/openj9/test/java/security/DigestLoop.java
@@ -1,0 +1,102 @@
+package org.openj9.test.java.security;
+
+/*******************************************************************************
+ * Copyright (c) 2019, 2019 IBM Corp. and others
+ *
+ * This program and the accompanying materials are made available under
+ * the terms of the Eclipse Public License 2.0 which accompanies this
+ * distribution and is available at https://www.eclipse.org/legal/epl-2.0/
+ * or the Apache License, Version 2.0 which accompanies this distribution and
+ * is available at https://www.apache.org/licenses/LICENSE-2.0.
+ *
+ * This Source Code may also be made available under the following
+ * Secondary Licenses when the conditions for such availability set
+ * forth in the Eclipse Public License, v. 2.0 are satisfied: GNU
+ * General Public License, version 2 with the GNU Classpath
+ * Exception [1] and GNU General Public License, version 2 with the
+ * OpenJDK Assembly Exception [2].
+ *
+ * [1] https://www.gnu.org/software/classpath/license.html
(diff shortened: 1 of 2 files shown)


llxia:
Jenkins test sanity xlinux,aix jdk11

PR #4708, opened 2019-02-12
Fix uninitialized variable

dmitry-ten:
Uninitialized variable was used in assert, which crashes
JITaaS build in debug. Initialize the variable.


dmitry-ten:
@mpirvu fyi

Issue #3421, opened 2018-10-23
Software based Concurrent Scavenge support on P


gita-omr:
The software based read barrier solution for pauseless GC on P

Related issue for Z: https://github.com/eclipse/openj9/issues/3276

Doc issue https://github.com/eclipse/openj9-docs/issues/134

Pull requests:
https://github.com/eclipse/openj9/pull/4159
https://github.com/eclipse/openj9/pull/4712

Dependencies:
 https://github.com/eclipse/openj9/issues/3453

SueChaplain:
Hi @gita-omr 

Assuming we will need some documentation to support this.  Please add doc:externals label to this and  open an issue at openj9-docs repo: https://github.com/eclipse/openj9-docs/issues/new?template=new-documentation-change.md so we can plan for this in the 0.12.0 list.

gita-omr:
An update:
While working on non-compressed version, @jimmyk found a problem with the way read barriers are generated for NULLCHK trees. @0dvictor is working on a different solution right now. @andrewcraik @fjeremic. I imagine this might affect all read/write barrier opcodes @cathyzhyi ...

0dvictor:
The NULLCHK issue will be fixed with https://github.com/eclipse/openj9/pull/3798.

gita-omr:
@pshipton : @IBMJimmyk made some cosmetic changes after another code review and we would like to run more tests. They will be done by Monday morning. Could the code split wait for that?

pshipton:
Yes, since we got not much delivered today because of the Jenkins issues it seems we need to delay another day.

pshipton:
Deferred to the next release.

gita-omr:
@pshipton we are ready to deliver it to 0.13.

pshipton:
@gita-omr when do you expect the merges to occur?

gita-omr:
@pshipton @amicic I updated Dependencies in the description -- JVMTI fixes for CS.

gita-omr:
Just merged #4159

fjeremic:
> Just merged #4159

@gita-omr left some extra review comments on that PR. There are some changes required.

vij-singh:
@gita-omr are we still ok for 0.13.0 ?

SueChaplain:
@gita-omr is this now moved to 0.14.0 please?

gita-omr:
No, it's in 0.13. We generally try to keep milestone field up to date.

gita-omr:
[...]

PR #4712, opened 2019-02-12
Clean up of Concurrent Scavenge code on Power

IBMJimmyk:
Comment regarding compatibility of TM and concurrent scavenge was updated.
At the current time there are no plans to be able to support both at the
time time so it was removed as being a TODO.

stopUsingRegister already performs a NULL check on the register so the
NULL checks before the calls were unnecessary.

Issue: #3421


IBMJimmyk:
Travis CI builds looked like they failed because:
"ERROR: cannot verify api.adoptopenjdk.net's certificate"

gita-omr:
Oops, I think we have not merged it. Will merge now.

gita-omr:
@pshipton can we get this into 0.13?

pshipton:
@gita-omr yes, there was nothing else blocking it, so I've pushed this to the 0.13 branch.

PR #4713, opened 2019-02-12
Add assertions for the UTF8 SRPs of J9ROMClass

hangshao0:

### DIFF
--- a/runtime/shared_common/CacheMap.cpp
+++ b/runtime/shared_common/CacheMap.cpp
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2001, 2018 IBM Corp. and others
+ * Copyright (c) 2001, 2019 IBM Corp. and others
  *
  * This program and the accompanying materials are made available under
  * the terms of the Eclipse Public License 2.0 which accompanies this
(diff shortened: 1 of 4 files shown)


hangshao0:
I didn't walk all the romMethods and romFields to assert the UTF8's there. Doing so could have more impact on the performance.

pshipton:
jenkins test all zlinux jdk8

runtime/shared_common/CacheMap.cpp
```diff
@@ -7400,3 +7406,50 @@ SH_CacheMap::updateLocalHintsData(J9VMThread* currentThread, J9SharedLocalStartu
...
+			innerClassNames++;
+		}
+	}
+
+#if defined(J9VM_OPT_VALHALLA_NESTMATES)
+	Trc_SHR_Assert_True((UDATA)J9ROMCLASS_NESTHOSTNAME(romClass) < romClassEnd);
+
+	if (romClass->nestMemberCount > 0) {
+		J9SRP *nestMemberNames = J9ROMCLASS_NESTMEMBERS(romClass);
+		for (i = 0; i < (U_32)nestMemberCount; i++) {
```
pshipton: nestMemberCount looks like an undeclared variable.

hangshao0: Fixed. It is inside #ifdef so compiler didn't capture this.

pshipton:
jenkins test all zlinux jdk11

DanHeidinga:
@hangshao0 Can you add some details describing why this change was made?  Also, were any startup performance tests run?

hangshao0:
> Can you add some details describing why this change was made?

I have added more details in the description. 

> Also, were any startup performance tests run?

No. I guess the performance impact is minimal. There are 3 + interfaceCount + innerClassCount + nestMemberCount of pointer comparisons added when writing a romClass to the shared cache.

DanHeidinga:
Thanks @hangshao0

Issue #3410, opened 2018-10-23
Mockito inlining appears to fail on J9/windows but work on J9/mac


planetf1:
I tried building http://github.com/odpi/egeria using the JDK 8 192 J9 SDK from adoptopenjdk on Windows 10. A few of the tests use mockito inline mocking and failed (see log below)

I then tried using JDK 8 192 SDK from adoptopenjdk on MacOS Mojave and it all seemed to work just fine, as indeed does both the oracle jdk on windows, and openjdk/hotspot on windows.

So this looks like an issue with the Windows 10 JDK

I do not know if this is a build issue, or the code.

original discussion was at https://adoptopenjdk.slack.com/archives/C09NLQQAV/p1540215988000100

Error I hit was:

```
[INFO] Running org.odpi.openmetadata.accessservices.governanceengine.client.GovernanceEngineImplTest
SLF4J: Failed to load class "org.slf4j.impl.StaticLoggerBinder".
SLF4J: Defaulting to no-operation (NOP) logger implementation
SLF4J: See http://www.slf4j.org/codes.html#StaticLoggerBinder for further details.
[ERROR] Tests run: 26, Failures: 0, Errors: 26, Skipped: 0, Time elapsed: 2.352 s <<< FAILURE! - in org.odpi.openmetadata.accessservices.governanceengine.client.GovernanceEngineImplTest
[ERROR] testGetGovernedAssetComponentListBadConstructorEmpty  Time elapsed: 1.862 s  <<< ERROR!
org.mockito.exceptions.base.MockitoException:

Mockito cannot mock this class: class org.springframework.web.client.RestTemplate.

If you're not sure why you're getting this error, please report to the mailing list.


Java               : 1.8
JVM vendor name    : Eclipse OpenJ9
JVM vendor version : openj9-0.11.0
JVM name           : Eclipse OpenJ9 VM
JVM version        : 1.8.0_192-b12
JVM info           : JRE 1.8.0 Windows 10 amd64-64-Bit Compressed References 20181019_105 (JIT enabled, AOT enabled)
OpenJ9   - 090ff9dc
OMR      - ea548a66
JCL      - 51609250b5 based on jdk8u192-b12
OS name            : Windows 10
OS version         : 10.0


You are seeing this disclaimer because Mockito is configured to create inlined mocks.
You can learn about inline mocks and their limitations under item #39 of the Mockito class javadoc.

[...]

PR #4714, opened 2019-02-12
Allow redefinition of Object in fast HCR

gacholio:
Allow java.lang.Object to be redefined (or retransformed) in fast HCR
mode (i.e. runtime HCR, not extended HCR used by the java debugger).

Related: #3410

[ci skip]

### DIFF
--- a/runtime/util/hshelp.c
+++ b/runtime/util/hshelp.c
@@ -98,7 +98,6 @@ static jvmtiError jitEventInitialize(J9VMThread * currentThread, jint redefinedC
 static void jitEventAddMethod(J9VMThread * currentThread, J9JVMTIHCRJitEventData * eventData, J9Method * oldMethod, J9Method * newMethod, UDATA equivalent);
 static void jitEventAddClass(J9VMThread * currentThread, J9JVMTIHCRJitEventData * eventData, J9Class * originalRAMClass, J9Class * replacementRAMClass);
 #endif
-static void removeFromSubclassHierarchy(J9JavaVM *javaVM, J9Class *clazzPtr);
 static void swapClassesForFastHCR(J9Class *originalClass, J9Class *newClass);
 
 #undef  J9HSHELP_DEBUG_SANITY_CHECK
@@ -351,7 +350,6 @@ fixJNIMethodID(J9VMThread *currentThread, J9Method *oldMethod, J9Method *newMeth
 					goto done;
 				}
 			}
-
 
 			oldMethodID = oldJNIIDs[oldMethodIndex];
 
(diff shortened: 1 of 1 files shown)


gacholio:
Note there is currently no testing for this because:

a) Proper testing requires generating class files (more difficult than just writing .java files)
b) Any test would fail at -Xint today (since that implies extended HCR)

gacholio:
jenkins test all plinux jdk8

gacholio:
I have hand-tested the virtual and non-virtual cases for Object (which also covers the private interface method case). Tests will eventually be added to cover all of these cases.

fjeremic:
@andrewcraik FYI so we can think about whether JIT optimizer makes certain assumptions about `java.lang.Object` and it's methods.

andrewcraik:
There are definitely assumptions in some places - we need to prioritize the dev work to do this against other items since it will take a bit of time to find all the possible dependencies.

gacholio:
@andrewcraik I'd call the JIT work low-priority at this point - we already allow Object to be transformed at load time, so any assumptions you're making can already be broken (and the VM has plenty of its own assumptions around Object).

andrewcraik:
So just caaturing some thoughts discussion related to this:
- what are we going to do if Object's finalize method gains a body - is the GC going to start running these? what are we going to do with JITCODE that has been produced assuming that no finalization is necessary?
- the JIT is going to need to stop special casing Object.hashCode and rely on forcing the inlining to get fastIdentityHashCode exposed to get good performance - I think this may be done, but we need to check
- we will need to worry about Object.clone changing - if clone gains side-effects they may not be respected by JIT optimizations to object cloning to force it inline - we'll need an appropriate guard
Given the impacts of the above can there at least be some kind of debug option to disable allowing Object redefinition in case of perf or functional problems?

DanHeidinga:
For the case of clone / finalize, can we, as you've elegantly suggested in the past, "throw the car battery in the pool" and dump the entire code cache if those methods are redefined?

andrewcraik:
certainly possible - depends on the perf behaviours that are expected.

gacholio:
The fix is to not recognize the Object methods at all, and recognize the primitives that they call (and we're not allowing to be modified).

For other things like finalize, we will report the modification to the JIT, so it can break assumptions, etc, as needed.

finalize() in particular is interesting, since we use the tag when we allocate. There are several options:

1)  Disallow Object becoming finalizeable

I like this since there's no way we're going to add finalization to arrays.

2) Allow it, propagate the finalize bits down the hierarchy

2a) New instances are finalizeable

Maybe.

2b) New and existing instances are finalizeable

Seems like a lot of work for nothing.

gacholio:
Note that the things we're talking about here have nothing to do with HCR - Object can be modified in any of the unpleasant ways we've discussed at load time, so we need to handle it in both cases, which means we should remove as many of our hard-coded assumptions as possible.

I'll open a new issue to deal with our expectations for Object.

gacholio:
Please continue any discussion in #4792

runtime/util/hshelp.c
```diff
@@ -1663,7 +1665,7 @@ reresolveHotSwappedConstantPool(J9ConstantPool * ramConstantPool, J9VMThread * c
 }
 
 static void
-fixRAMConstantPoolForFastHCR(J9ConstantPool *ramConstantPool, J9HashTable *classHashTable, J9HashTable *methodHashTable)
+fixRAMConstantPoolForFastHCR(J9ConstantPool *ramConstantPool, J9HashTable *classHashTable, J9HashTable *methodHashTable, J9Class *objectClass)
```
DanHeidinga: Won't this pass in a potentially stale object class ref?  I'm not clear on why passing it is better than fetching it from the VM here instead of in the caller?

gacholio: I pass it to ensure the original class pointer value (I want the stale one in order to compute indices into the method array).

DanHeidinga: Ok, that makes sense

Issue #4470, opened 2019-01-27
Test-sanity.functional-JDK11-linux_ppc-64_cmprssptrs_le UnsafeTests_0 crash Compiled_method=org/openj9/test/unsafe/UnsafeTestBase.testGetInt


pshipton:
https://ci.eclipse.org/openj9/job/Test-sanity.functional-JDK11-linux_ppc-64_cmprssptrs_le/450
```
21:51:56 ===============================================
21:51:56 Running test UnsafeTests_0 ...
21:51:56 ===============================================
21:51:56 UnsafeTests_0 Start Time: Sun Jan 27 02:51:56 2019 Epoch Time (ms): 1548557516026
21:51:56 "/home/jenkins/jenkins-agent/workspace/Test-sanity.functional-JDK11-linux_ppc-64_cmprssptrs_le/openjdkbinary/j2sdk-image/bin/java" -Xshareclasses:destroyAll; "/home/jenkins/jenkins-agent/workspace/Test-sanity.functional-JDK11-linux_ppc-64_cmprssptrs_le/openjdkbinary/j2sdk-image/bin/java" -Xshareclasses:groupAccess,destroyAll; echo "cache cleanup done";
21:51:56 JVMSHRC005I No shared class caches available
21:51:56 JVMSHRC005I No shared class caches available
21:51:56 cache cleanup done
21:51:56 variation: -DScenario=Regular
21:51:56 JVM_OPTIONS:  -Xcompressedrefs -DScenario=Regular 
21:51:56 { itercnt=1; \
21:51:56 mkdir -p "/home/jenkins/jenkins-agent/workspace/Test-sanity.functional-JDK11-linux_ppc-64_cmprssptrs_le/openjdk-tests/TestConfig/scripts/testKitGen/../../../TestConfig/test_output_15485542235723/UnsafeTests_0"; \
21:51:56 cd "/home/jenkins/jenkins-agent/workspace/Test-sanity.functional-JDK11-linux_ppc-64_cmprssptrs_le/openjdk-tests/TestConfig/scripts/testKitGen/../../../TestConfig/test_output_15485542235723/UnsafeTests_0"; \
21:51:56 "/home/jenkins/jenkins-agent/workspace/Test-sanity.functional-JDK11-linux_ppc-64_cmprssptrs_le/openjdkbinary/j2sdk-image/bin/java"  -Xcompressedrefs -DScenario=Regular  \
[...]

PR #4721, opened 2019-02-13
Add range check test around profiled guard method test

r30shah:
In OpenJ9 we take advantage of the profiled information in inlining method for virtual call. While inlining the method, we generate a guard around inlined code to make sure that the inlined implementation is indeed what we need to execute. For this `TR_ProfiledGuard**`
kind of guard it either `TR_VftTest` or `TR_MethodTest`.

In case of `TR_MethodTest`, it records the offset of the inlined method in the Virtual Function Table of the class in which that method belongs to and compares the value at the same offset of class of receiver object with the `J9Method` of inlined method. This comparison introduces a functional bug.
Some classes would have smaller function table and recorded offset w.r.t the class
might point to a garbage value or even might be inaddressible (Out of mapped memory range). This commit adds a range check test around the TR_MethodTest to verify that address that we are going to reference points to valid address.

### DIFF
--- a/runtime/compiler/codegen/J9CodeGenerator.cpp
+++ b/runtime/compiler/codegen/J9CodeGenerator.cpp
@@ -53,6 +53,7 @@
 #include "il/symbol/LabelSymbol.hpp"
 #include "infra/Assert.hpp"
 #include "infra/BitVector.hpp"
+#include "infra/ILWalk.hpp"
 #include "infra/List.hpp"
 #include "optimizer/Structure.hpp"
 #include "optimizer/TransformUtil.hpp"
@@ -3915,6 +3916,132 @@ J9::CodeGenerator::collectSymRefs(
 
    return true;
    }
+
+  /** \brief 
+    *       Following codegen phase walks the blocks in the CFG and checks for the virtual guard performing TR_MethodTest
+    *       and guarding an inlined interface call. 
+    *
+    * \details
+    *       Virtual Guard performing TR_MethodTest would look like following.
+    *       n1n BBStart <block_X>
+    *       ...      
+    *       n2n ifacmpne goto -> nXXn 
+    *       n3n    aloadi <offset of inlined method in VTable>
+    *       n4n       aload <vft>
+    *       n5n    aconst <J9Method of inlined method>
+    *       n6n BBEnd <block_X>
+    *       For virtual dispatch sequence, we know that this is the safe check but in case of interface call, classes implementing
+    *       that interface would have different size of VTable. This makes executing above check unsafe when VTable of the class of
+    *       the receiver object is smaller, effectively making reference in n3n to pointing to a garbage location which might lead 
+    *       to a segmentation fault if the reference in not memory mapped or if bychance it contains J9Method  pointer of same inlined
+    *       method then it will execute a code which should not be executed.
+    *       For this kind of Virtual guards which are not nop'd we need to add a range check to make sure the address we are going to
+    *       access is pointing to a valid location in VTable. There are mainly two ways we can add this range check test. First one is
+    *       during the conception of the virtual guard. There are many downsides of doing so especially when other optimizations which
+    *       can moved guards around (for example loop versioner, virtualguard head merger, etc) needs to make sure to move range check 
+    *       test around as well. Other way is to scan for this type of guards after optimization is finished like here in CodeGen Phase
+    *       and add a range check test here.
+    *       At the end of this function, we would have following code around them method test.
+    *       BBStart <block_X>    
+    *       ...      
+    *       ifacmple goto nXXn
+    *          aloadi <offset of VTableHeader.size from J9Class*>
+    *             aload <vft>
+    *          aconst <Index of the inlined method in VTable of class of inlined method>
+    *       BBEnd <block_X>
+    * 
+    *       BBStart <block_Y>
+    *       ifacmpne goto -> nXXn 
+    *          aloadi <offset of inlined method in VTable>
+    *             aload <vft>
+    *          aconst <J9Method of inlined method>
+    *       BBEnd <block_Y>
+    */  
+void
+J9::CodeGenerator::fixUpProfiledInterfaceGuardTest()
+   {
+   TR::Compilation *comp = self()->comp();
+   TR::CFG * cfg = comp->getFlowGraph();
+   TR::NodeChecklist checklist(comp);
+   for (TR::AllBlockIterator iter(cfg, comp); iter.currentBlock() != NULL; ++iter)
+      {
+      TR::Block *block = iter.currentBlock();
+      TR::TreeTop *treeTop = block->getLastRealTreeTop();
+      TR::Node *node = treeTop->getNode();
+      if (node->getOpCode().isIf() && node->isTheVirtualGuardForAGuardedInlinedCall() && !checklist.contains(node))
+         {
+         TR_VirtualGuard *vg = comp->findVirtualGuardInfo(node);
+         // Mainly we need to make sure that virtual guard which performs the TR_MethodTest and can be NOP'd are needed the range check.
+         if (vg && vg->getTestType() == TR_MethodTest && 
+            !(comp->performVirtualGuardNOPing() && (node->isNopableInlineGuard() || comp->isVirtualGuardNOPingRequired(vg))))
+            {
+            TR::SymbolReference *callSymRef = vg->getSymbolReference();
+            TR_ASSERT_FATAL(callSymRef != NULL, "Guard n%dn for the inlined call should have stored symbol reference for the call", node->getGlobalIndex());
+            if (callSymRef->getSymbol()->castToMethodSymbol()->isInterface())
+               {
+               TR::DebugCounter::incStaticDebugCounter(comp, TR::DebugCounter::debugCounterName(comp, "profiledInterfaceTest/({%s}{%s})", comp->signature(), comp->getHotnessName(comp->getMethodHotness())));
+               dumpOptDetails(comp, "Need to add a rangecheck before n%dn in block_%d\n",node->getGlobalIndex(), block->getNumber());
+
+               // We need a VFT Load of the receiver object to get the VTableHeader.size to check the range. As this operation is happening during codegen phase, only
+               // known concrete way we can have this information is through aloadi child of the guard that has single child which is vft load of receiver object.
+               // Now instead of accessing VFT load from the child of the aloadi, we could have treetop's the aloadi during inlining where we generate the virtual guard
+               // to access information from the treetop. Because of the same reasons lined up behind adding range check test during codegen phase in the description of this function,
+               // we would need to make changes in all optimizations moving Virtual Guard around to keep that treetop together before guard which will be very difficult to enforce.
+               // Also as children of virtual guard is very self contained and atm it is very unlikely that other optimizations are going to find opportunity of manipulating them and
+               // Because of the fact that it is very unlikely that we will have another aloadi node with same VTable offset of same receiver object, this child would not be commoned out
+               // and have only single reference in this virtual guard therefore splitting of block will not store it to temp slot.
+               // In rare case child of the virtual guard is manipulated then illegal memory reference load would hace occured before the Virtual Guard which 
+               // is already a bug as mentioned in the description of this function and it would be safer to fail compilation. 
+               TR::Node *vTableLoad = node->getFirstChild();
+               if (!(vTableLoad->getOpCodeValue() == TR::aloadi && comp->getSymRefTab()->isVtableEntrySymbolRef(vTableLoad->getSymbolReference())))
+                  comp->failCompilation<TR::CompilationException>("Abort compilation as Virtual Guard has generated illegal memory reference");
+               TR::Node *vTableSizeOfReceiver = NULL;
+               TR::Node *rangeCheckTest = NULL;
+               if (TR::Compiler->target.is64Bit())
+                  {
+                  vTableSizeOfReceiver = TR::Node::createWithSymRef(TR::lloadi, 1, 1, vTableLoad->getFirstChild(),
+                                                                           comp->getSymRefTab()->findOrCreateVtableEntrySymbolRef(comp->getMethodSymbol(),
+                                                                                                                                    sizeof(J9Class)+ offsetof(J9VTableHeader, size)));
+                  rangeCheckTest = TR::Node::createif(TR::iflcmple, vTableSizeOfReceiver, 
+                                                                  TR::Node::lconst(node,  (vTableLoad->getSymbolReference()->getOffset() - sizeof(J9Class) - sizeof(J9VTableHeader)) / sizeof(UDATA)) , 
+                                                                  node->getBranchDestination());
+                  }
+               else
+                  {
+                  vTableSizeOfReceiver = TR::Node::createWithSymRef(TR::iloadi, 1, 1, vTableLoad->getFirstChild(),
+                                                                           comp->getSymRefTab()->findOrCreateVtableEntrySymbolRef(comp->getMethodSymbol(),
+                                                                                                                                    sizeof(J9Class)+ offsetof(J9VTableHeader, size)));
+                  rangeCheckTest = TR::Node::createif(TR::ificmple, vTableSizeOfReceiver, 
+                                                                  TR::Node::iconst(node,  (vTableLoad->getSymbolReference()->getOffset() - sizeof(J9Class) - sizeof(J9VTableHeader)) / sizeof(UDATA)) , 
+                                                                  node->getBranchDestination());
+                  }
+               TR::TreeTop *rangeTestTT = TR::TreeTop::create(comp, treeTop->getPrevTreeTop(), rangeCheckTest);
+               TR::Block *newBlock = block->split(treeTop, cfg, false, false);
+               cfg->addEdge(block, node->getBranchDestination()->getEnclosingBlock());
+               newBlock->setIsExtensionOfPreviousBlock();
+               if (node->getNumChildren() == 3)
+                  {
+                  TR::Node *currentBlockGlRegDeps = node->getChild(2);
+                  TR::Node *exitGlRegDeps = TR::Node::create(TR::GlRegDeps, currentBlockGlRegDeps->getNumChildren());
+                  for (int i = 0; i < currentBlockGlRegDeps->getNumChildren(); i++)
+                     {
+                     TR::Node *child = currentBlockGlRegDeps->getChild(i);
+                     exitGlRegDeps->setAndIncChild(i, child);
+                     }
+                  rangeCheckTest->addChildren(&exitGlRegDeps, 1);
+                  }
+               // While walking all blocks in CFG, when we find the location to add the range check, it will split the original block and
+               // We will have actual Virtual Guard in new block. As Block Iterator guarantees to visit all block in the CFG, 
+               // While going over the blocks, we will encounter same virtual guard in newly created block after split.  
+               // We need to make sure we are not examining already visited guard. 
+               // Add checked virtual guard node to NodeChecklist to make sure we check all the nodes only once.
+               checklist.add(node);
+               }
+            }
+         }
+      }
+   }
+
 
 
 void
--- a/runtime/compiler/codegen/CodeGenPhaseToPerform.hpp
+++ b/runtime/compiler/codegen/CodeGenPhaseToPerform.hpp
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2000, 2018 IBM Corp. and others
+ * Copyright (c) 2000, 2019 IBM Corp. and others
  *
  * This program and the accompanying materials are made available under
  * the terms of the Eclipse Public License 2.0 which accompanies this
@@ -28,6 +28,7 @@
 
 
     ReserveCodeCachePhase,
+    FixUpProfiledInterfaceGuardTest,
 
 
     InliningReportPhase,
--- a/runtime/compiler/codegen/J9CodeGenPhase.cpp
+++ b/runtime/compiler/codegen/J9CodeGenPhase.cpp
@@ -58,6 +58,11 @@ J9::CodeGenPhase::getNumPhases()
    return static_cast<int>(TR::CodeGenPhase::LastJ9Phase);
    }
 
+void 
+J9::CodeGenPhase::performFixUpProfiledInterfaceGuardTestPhase(TR::CodeGenerator *cg, TR::CodeGenPhase *phase)
+   {
+   cg->fixUpProfiledInterfaceGuardTest();
+   }
 
 void
 J9::CodeGenPhase::performAllocateLinkageRegistersPhase(TR::CodeGenerator * cg, TR::CodeGenPhase * phase)
@@ -145,6 +150,8 @@ J9::CodeGenPhase::getName(TR::CodeGenPhase::PhaseValue phase)
 	      return "IdentifyUnneededByteConvsPhase";
       case LateSequentialConstantStoreSimplificationPhase:
          return "LateSequentialConstantStoreSimplification";
+      case FixUpProfiledInterfaceGuardTest:
+         return "FixUpProfiledInterfaceGuardTest";
       default:
          return OMR::CodeGenPhaseConnector::getName(phase);
       }
--- a/runtime/compiler/codegen/J9CodeGenPhase.hpp
+++ b/runtime/compiler/codegen/J9CodeGenPhase.hpp
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2000, 2018 IBM Corp. and others
+ * Copyright (c) 2000, 2019 IBM Corp. and others
  *
  * This program and the accompanying materials are made available under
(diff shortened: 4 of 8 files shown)


r30shah:
@fjeremic @andrewcraik @jdmpapin  Can I please get a review? Currently I am doing a performance evaluation of this change and also I need to add documentation of what this is doing. so marked this as `WIP`.
Looking at the function where I can possibly add this test in lower trees is https://github.com/eclipse/openj9/blob/99e024af1e763969da4a3030ec72543dd3d4ea88/runtime/compiler/codegen/J9CodeGenerator.cpp#L784. 
It is a massive function doing all sorts of things, and it looks like this kind of change needs proper documentation and separate phase generating this test require us to split the block and copy `GlRegDeps`.    
So I have created a separate CodeGen Phase to add the range check test instead of the doing it in `lowerTrees`.  I would like to get all reviewer's view on this.

runtime/compiler/codegen/J9CodeGenerator.cpp
```diff
@@ -3917,6 +3917,53 @@ J9::CodeGenerator::collectSymRefs(
...
 
 
+
+void
+J9::CodeGenerator::fixUpProfiledInterfaceGuardTest()
+   {
+   TR::Compilation *comp = self()->comp();
+   TR::CFG * cfg = comp->getFlowGraph();
+   TR::Block *currentBlock = NULL;
+   for (TR::TreeTop * treeTop = comp->getStartTree(); treeTop; treeTop = treeTop->getNextTreeTop())
```
andrewcraik: We are you walking all TreeTops - a profiled guard has to be at the end of a basic block so you can walk the blocks in treetop order checking their exits which will save time...

runtime/compiler/codegen/J9CodeGenerator.cpp
```diff
@@ -3917,6 +3917,53 @@ J9::CodeGenerator::collectSymRefs(
...
+      if (node->getOpCodeValue() == TR::BBStart )
+         {
+         currentBlock = node->getBlock();
+         }
+      else if (node->getOpCode().isIf() && node->isTheVirtualGuardForAGuardedInlinedCall())
+         {
+         TR_VirtualGuard *vg = comp->findVirtualGuardInfo(node);
+         if (vg && vg->getTestType() == TR_MethodTest && vg->getIsInterfaceMethodGuard())
+            {
+            traceMsg(comp, "Need to add a rangecheck before n%dn in block_%d\n",node->getGlobalIndex(), currentBlock->getNumber());
```
andrewcraik: this should be a dumpOptDetails rather than just a trace message I would have thought...

runtime/compiler/codegen/J9CodeGenerator.cpp
```diff
@@ -3917,6 +3917,53 @@ J9::CodeGenerator::collectSymRefs(
...
+            TR::Node* vTableSizeOfReceiver = TR::Node::createWithSymRef(TR::iloadi, 1, 1, node->getFirstChild()->getFirstChild(),
```
andrewcraik: How are you guaranteed that the grandchild exists?

andrewcraik:
I'm not sure on the compile-time implication of a new phase - @fjeremic thoughts? The clarity certainly seems useful rather than weaving it in with a bunch of other code. That being said some more comments wouldn't hurt :)

r30shah:
OMR Counterpart: https://github.com/eclipse/omr/pull/3564

runtime/compiler/codegen/J9CodeGenerator.cpp
0xdaryl: Consider using one of the Block iterators available in ILWalk.cpp to accomplish this walk.

fjeremic:
> I'm not sure on the compile-time implication of a new phase - @fjeremic thoughts? The clarity certainly seems useful rather than weaving it in with a bunch of other code. That being said some more comments wouldn't hurt :)

We've introduced other codegen phases that do a treetop walk (example [1]) and they have never been visible on a startup libj9jit29.so profile. I'm in support of the logical separation.

[1] https://github.com/eclipse/openj9/blob/35c44ead49a53f8c0ad1ca3e63c18711d3a7b26f/runtime/compiler/z/codegen/CodeGenPhaseToPerform.hpp#L51

runtime/compiler/codegen/J9CodeGenerator.cpp
```diff
@@ -3917,6 +3917,53 @@ J9::CodeGenerator::collectSymRefs(
    }
 
 
+
+void
+J9::CodeGenerator::fixUpProfiledInterfaceGuardTest()
```
fjeremic: Can we please Doxygen document precisely what this phase does? Let's take this opportunity to clean up the triple newline immediately before as well.

runtime/compiler/codegen/J9CodeGenerator.cpp
```diff
@@ -3917,6 +3917,53 @@ J9::CodeGenerator::collectSymRefs(
...
+      TR::Node *node = treeTop->getNode();
+
+      if (node->getOpCodeValue() == TR::BBStart )
+         {
+         currentBlock = node->getBlock();
+         }
+      else if (node->getOpCode().isIf() && node->isTheVirtualGuardForAGuardedInlinedCall())
+         {
+         TR_VirtualGuard *vg = comp->findVirtualGuardInfo(node);
+         if (vg && vg->getTestType() == TR_MethodTest && vg->getIsInterfaceMethodGuard())
```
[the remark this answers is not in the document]
andrewcraik: Someone will need to check but a combination of `isNopable` and the codegen supporting noping is what I would expect to be the right test.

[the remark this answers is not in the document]
r30shah: Looking at the `virtualGuardHelper` 
https://github.com/eclipse/omr/blob/6e7568d6532997b3de2e41da2560f9af79cd9799/compiler/z/codegen/ControlFlowEvaluator.cpp#L139
I see we have a codegen query to know if VGNOP is supported or not : 
`cg->getSupportsVirtualGuardNOPing()`

runtime/compiler/codegen/J9CodeGenerator.cpp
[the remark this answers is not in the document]
r30shah: Changed load to `aloadi` to make sure, it loads correct size from the J9VTableHeader. https://github.com/eclipse/openj9/pull/4721/commits/36292aa9f2fcefeedc9fe76a5f0929a1ce9c2d68

runtime/compiler/codegen/J9CodeGenerator.cpp
```diff
@@ -3917,6 +3917,53 @@ J9::CodeGenerator::collectSymRefs(
...
+            {
+            traceMsg(comp, "Need to add a rangecheck before n%dn in block_%d\n",node->getGlobalIndex(), currentBlock->getNumber());
+            TR::Node* vTableSizeOfReceiver = TR::Node::createWithSymRef(TR::iloadi, 1, 1, node->getFirstChild()->getFirstChild(),
+                                                                           comp->getSymRefTab()->findOrCreateVtableEntrySymbolRef(comp->getMethodSymbol(),
+                                                                                                                                    sizeof(J9Class)+ offsetof(J9VTableHeader, size)));
+            TR::Node * rangeCheckTest = TR::Node::createif(TR::ificmplt, vTableSizeOfReceiver, TR::Node::iconst(node, node->getFirstChild()->getSymbolReference()->getOffset()), node->getBranchDestination());
```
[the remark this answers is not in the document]
r30shah: Thanks for catching this. Changed to adjust the constant offset to point to index in table . 
https://github.com/eclipse/openj9/pull/4721/commits/36292aa9f2fcefeedc9fe76a5f0929a1ce9c2d68

runtime/compiler/codegen/J9CodeGenerator.cpp
```diff
@@ -3917,6 +3918,118 @@ J9::CodeGenerator::collectSymRefs(
...
+               // we would need to make changes in all optimizations moving Virtual Guard around to keep that treetop together before guard which will be very difficult to enforce.
+               // Also as children of virtual guard is very self contained and atm it is very unlikely that other optimizations are going to find opportunity of manipulating them and
+               // Because of the fact that it is very unlikely that we will have another aloadi node with same VTable offset of same receiver object, this child would not be commoned out
+               // and have only single reference in this virtual guard therefore splitting of block will not store it to temp slot.
+               // Due to the above reasons we have a fatal assert in case the first child of the guard node performing is not aloadi. If there exists any case we need to reconsider our choice for range check test.
+               TR_ASSERT_FATAL(node->getFirstChild()->getOpCodeValue() == TR::aloadi, "Virtual Guard perfoming TR_MethodTest should have aloadi as first child, to extract the vftPointer of the receiver object");
```
r30shah: @andrewcraik  @jdmpapin  I have added a comment here regarding to why we have the FATAL_ASSERT if child of the guard is not `aloadi` .

runtime/compiler/codegen/J9CodeGenerator.cpp
r30shah: Added change to walk blocks in CFG and checking Exits (https://github.com/eclipse/openj9/pull/4721/commits/36292aa9f2fcefeedc9fe76a5f0929a1ce9c2d68)

runtime/compiler/codegen/J9CodeGenerator.cpp
r30shah: changed in https://github.com/eclipse/openj9/pull/4721/commits/36292aa9f2fcefeedc9fe76a5f0929a1ce9c2d68

runtime/compiler/codegen/J9CodeGenerator.cpp
```diff
@@ -3917,6 +3918,118 @@ J9::CodeGenerator::collectSymRefs(
...
+      {
+      TR::Block *block = iter.currentBlock();
+      TR::TreeTop *treeTop = block->getLastRealTreeTop();
+      TR::Node *node = treeTop->getNode();
+      if (node->getOpCode().isIf() && node->isTheVirtualGuardForAGuardedInlinedCall() && !checklist.contains(node))
+         {
+         TR_VirtualGuard *vg = comp->findVirtualGuardInfo(node);
+         // Mainly we need to make sure that virtual guard which performs the TR_MethodTest and can be NOP'd are needed the range check.
+         if (vg && vg->getTestType() == TR_MethodTest && 
+            !(comp->performVirtualGuardNOPing() && (node->isNopableInlineGuard() || comp->isVirtualGuardNOPingRequired(vg))))
```
r30shah: @andrewcraik @jdmpapin  Combining query from CodeGen where it generates the VGNOP. Basically if VGNOP is supported and node or guard requires NOPing then we won't add range check test.

runtime/compiler/codegen/J9CodeGenerator.cpp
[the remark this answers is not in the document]
jdmpapin: Can we instead choose between `iloadi`/`iconst`/`icmp` and `lloadi`/`lconst`/`lcmp` as appropriate? Neither of the values being compared here is really an address, and the ordering `acmp` opcodes are a bit unusual

runtime/compiler/codegen/J9CodeGenerator.cpp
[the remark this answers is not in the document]
jdmpapin: The comparison should still change to `le` from `lt`

runtime/compiler/codegen/J9CodeGenerator.cpp
jdmpapin: I've recently seen PRE hoist an `aloadi <vft-symbol>` out of a loop, leaving a profiled guard with a VFT test of this form:
```
ifacmpne --> slow path
  aload <temp slot ...>
  aconst 0x... (isClassPointerConstant)
```
There might be circumstances where `aloadi <vtable-entry-symbol>` looks like it will happen unconditionally, so that normally it would be safe to hoist, but because it's part of one of these guards it would now be implicitly conditional.

runtime/compiler/codegen/J9CodeGenerator.cpp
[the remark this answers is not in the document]
r30shah: Changed in https://github.com/eclipse/openj9/pull/4721/commits/29a608cd75d687e9b64955c6fde4c7f1b3a031e6

runtime/compiler/codegen/J9CodeGenerator.cpp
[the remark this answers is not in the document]
r30shah: Changed in https://github.com/eclipse/openj9/pull/4721/commits/29a608cd75d687e9b64955c6fde4c7f1b3a031e6

runtime/compiler/codegen/J9CodeGenerator.cpp
r30shah: @andrewcraik  I discussed this with @jdmpapin  as well, although it is possible to optimization to hoist out  `aloadi` we would still left with the issue where that `aloadi` is not safe as it is followed by invisible condition.  Another solution Devin came up with was , we could have new type of conditional load node in trees which we can identify during codegen phase and generate a range check test. Even in this case, if some optimization hoist id out of loop, we would have difficulty generating a range check test. 
It seems if some optimization tries to work with these type of load, it needs to be cautious not to mess up these kind of invisible conditional node. 
@andrewcraik  I would like to get your view on this. ATM I have strengthen out the ASSERT by also checking the type of symref associated with `aloadi` which should be of type vTable Entry Symbol in https://github.com/eclipse/openj9/pull/4721/commits/ddbed9715f1f2fd5b5f4b123bbbe809d09bee454

r30shah:
Ran a personal test and I am not seeing any failures. Also finished performance runs with the last changes incorporated. 

`DayTrader-7` on Linux on Z (4 Core No SMT) results. 

1. `Compilation Time Results`

| Config | Avg. Compilation Time | % Standard Deviation | % Perf |
|--------|-----------------------|----------------------|--------|
|Baseline| 156956.7              | 3.65                 | 100    |
|RangeTest| 158405.4             | 3.72                 | 99.08  |

2. `Throughput Results`

| Config | Avg. Throughput       | % Standard Deviation | % Perf |
|--------|-----------------------|----------------------|--------|
|Baseline| 10309.89              | 0.49                 | 100    |
|RangeTest| 10300.23             | 0.75                 | 99.90  |

`Spark` Results

1. `HiBench/Pagerank` Performance

| Config | Time | % Standard Deviation | % Perf |
|--------|------|----------------------|--------|
| Baseline | 39.39 | 1.45              | 100    |
| RangeTest| 39.62 | 3.93              | 99.41  |

2. `HiBench/Terasort` Performance

| Config | Time | % Standard Deviation | % Perf |
|--------|------|----------------------|--------|
| Baseline | 32.35 | 7.60              | 100    |
| RangeTest| 31.53 | 0.83              | 102.6  |

In summary I do not see any effect of having these extra test in both compilation time as well as throughput.

runtime/compiler/codegen/J9CodeGenerator.cpp
```diff
@@ -3916,6 +3917,131 @@ J9::CodeGenerator::collectSymRefs(
...
+               // we would need to make changes in all optimizations moving Virtual Guard around to keep that treetop together before guard which will be very difficult to enforce.
+               // Also as children of virtual guard is very self contained and atm it is very unlikely that other optimizations are going to find opportunity of manipulating them and
+               // Because of the fact that it is very unlikely that we will have another aloadi node with same VTable offset of same receiver object, this child would not be commoned out
+               // and have only single reference in this virtual guard therefore splitting of block will not store it to temp slot.
+               // Due to the above reasons we have a fatal assert in case the first child of the guard node performing is not aloadi. If there exists any case we need to reconsider our choice for range check test.
+               TR::Node *vTableLoad = node->getFirstChild();
+               TR_ASSERT_FATAL(vTableLoad->getOpCodeValue() == TR::aloadi &&
```
andrewcraik: Given that Devin has seen the aloadi happen I don't think we can tolerate a fatal assert in production - would you crash the JVM? I think the right thing is to abort the compilation... It should be rare but a fatal assert when we have a known example makes me worry that we would generate issues for users needlessly

andrewcraik: Another alternative to failing the compile is to do a type test or to jump unconiditionally to the failed side of the guard - the latter probably being the easiest.

runtime/compiler/codegen/J9CodeGenerator.cpp
```diff
@@ -3916,6 +3917,131 @@ J9::CodeGenerator::collectSymRefs(
    return true;
    }
 
+  /** \brief
```
andrewcraik: this doesn't seem brief...

runtime/compiler/codegen/J9CodeGenerator.cpp
andrewcraik: I think this is ok - @jdmpapin concerns?

runtime/compiler/codegen/J9CodeGenerator.cpp
r30shah: I am getting inclined towards the idea jump unconditionally to prevent JVM for crashing. This would definitely have lesser perf impact then failing compilation but I have one concern with that. In case, some one saves `vtable-entry-table` symbol in temp slot (hoisting out of the loop or splitting block, etc), that means we would still have a bug present in the compiled body where illegal reference appears well before of the guard. Skipping over the whole guard and inlined body in that case will not prevent the illegal memory reference.

runtime/compiler/codegen/J9CodeGenerator.cpp
r30shah: Fixed in https://github.com/eclipse/openj9/pull/4721/commits/657326068c58cbe715a3d1d1af251eaa1ab3d371

runtime/compiler/codegen/J9CodeGenerator.cpp
fjeremic: > Another alternative to failing the compile is to do a type test or to jump unconiditionally to the failed side of the guard - the latter probably being the easiest.

I agree. The safe thing to do is always bail to the slow path. We can place a static debug counter here to keep track of how often we need fall back to the slow path if we want such data (in customer scenarios). 

@andrewcraik do we have some sort of debug counter name prepend which is meant for "bad performance" things that we're doing?

andrewcraik: not at present - we were going to look to do that as part of the assert conversion work - we hadn't agreed a convention yet.

jdmpapin: I haven't actually seen `aloadi <vtable-entry-symbol>` get replaced with a load of a temp, but I have seen the corresponding thing happen to `aloadi <vft-symbol>`, and I know of a couple of situations in which it would be reasonable to do that transformation with `<vtable-entry-symbol>` as well, unless we decide that it's illegal to do so.

----

The first situation is as follows. We have two calls, one after the other, calling the same method of the same object.
```
treetop
  calli Foo.bar()V
    aloadi <vft-symbol>
      aload obj
    ==>aload obj
treetop
  calli Foo.bar()V
    aloadi <vft-symbol>
      aload obj
    ==>aload obj
```
We inline both with the same profiled guard with method test, and virtual guard tail splitter causes the second guard to be dominated by the first inlined body. After block extension, local CSE commons the children of the guards:
```
BBStart
ifacmpne --> slow call 0
  aloadi <vtable-entry-symbol>
    aloadi <vft-symbol>
      aload obj
  aconst 0x...
BBEnd
BBStart (extension)
<...inlined Foo.bar()V 0...>
BBEnd
BBStart (extension)
ifacmpne --> slow call 1
  ==>aloadi <vtable-entry-symbol>
  ==>aconst 0x...
BBEnd
BBStart (extension)
<...inlined Foo.bar()V 1...>
BBEnd
```
If `<...inlined Foo.bar()V 0...>` is split and a merge point is inserted, then `aloadi <vtable-entry-symbol>` will be spilled into a temp.

Assuming this kind of thing is sufficiently unlikely, I believe it's fine to deal with it by unconditionally failing the guard, as mentioned upthread.

----

The second scenario is code motion, as @r30shah mentioned. To elaborate a bit, the safety requirements for `aloadi <vtable-entry-symbol>` are stricter than those for `<vft-symbol>`, and I don't know whether anything checks for those and moves these loads. However, it would be reasonable to hoist any (invariant) load whatsoever if, upon entering the loop, the load would necessarily run, e.g.
```
do {
    x = obj.baz;  // NB. if loop has been entered, this *always* runs
    ...
} while(...);
```
can in general be safely rewritten as
```
tmp = obj.baz;
do {
    x = tmp;
    ...
} while(...);
```
because if the load is disallowed at the new location, it would also have been disallowed at its original location, no matter what its safety conditions are.

The issue would be that with this change, when `aloadi <vtable-entry-symbol>` occurs in a profiled guard, the vtable entry isn't necessarily loaded every time we reach the guard. It has an implicit safety condition that is tested at runtime. Because the test is implicit, it might look as though the load can be hoisted, but hoisting it would remove it from the guard, and as a result separate it from the implicit test.

To ensure that we don't hit this kind of problem in the future, I think we have to do one of the following:
1. decree that no optimization may speculatively load `<vtable-entry-symbol>`, at least when its original occurrence was under a profiled guard; or
2. decree that no optimization may speculatively load `<vtable-entry-symbol>` without first checking the size of the vtable, _even if_ the existing control flow appears to guarantee that the vtable is large enough; or
3. create a new node that encapsulates the conditional, which would be safe to speculatively evaluate; or
4. make the safety test explicit in the CFG.

Personally I'm not a fan of restrictions like 1 or 2 (out of which 2 seems less objectionable to me), but I understand if given the alternatives one of them seems to be the best option, or at least the best option for the moment.

If we do want to go with something like 1 or 2, I'm not suggesting that we have to immediately audit all optimizations that could insert problematic speculative loads. This is more of a request to consider and possibly document the direction of the design w.r.t. this sort of issue

r30shah: Considering everything, It seems like jumping to slower path would not prevent the bug.
As we are checking both opcode of the child as well as type of symbol reference, If we do not have both, it would be safer to fail compilation. 
I have changed the code in https://github.com/eclipse/openj9/pull/4721/commits/0f44e093a81b6d6cb75ce55ccfef8dd1033e9a69

runtime/compiler/codegen/J9CodeGenerator.cpp
```diff
@@ -3916,6 +3917,132 @@ J9::CodeGenerator::collectSymRefs(
...
+               // we would need to make changes in all optimizations moving Virtual Guard around to keep that treetop together before guard which will be very difficult to enforce.
+               // Also as children of virtual guard is very self contained and atm it is very unlikely that other optimizations are going to find opportunity of manipulating them and
+               // Because of the fact that it is very unlikely that we will have another aloadi node with same VTable offset of same receiver object, this child would not be commoned out
+               // and have only single reference in this virtual guard therefore splitting of block will not store it to temp slot.
+               // In rare case child of the virtual guard is manipulated then illegal memory reference load would hace occured before the Virtual Guard which 
+               // is already a bug as mentioned in the description of this function and it would be safer to fail compilation. 
+               TR::Node *vTableLoad = node->getFirstChild();
+               if (!(vTableLoad->getOpCodeValue() == TR::alaodi && comp->getSymRefTab()->isVtableEntrySymbolRef(vTableLoad->getSymbolReference())))
```
jdmpapin: `TR::aloadi` (the `a` and `o` are switched)

runtime/compiler/codegen/J9CodeGenerator.cpp
jdmpapin: Something like this is probably the best we'll get without some mechanism to get codegen to commit in advance to using `vgnop`. I think it's good for now

runtime/compiler/codegen/J9CodeGenerator.cpp
```diff
@@ -3916,6 +3917,132 @@ J9::CodeGenerator::collectSymRefs(
...
+               // we would need to make changes in all optimizations moving Virtual Guard around to keep that treetop together before guard which will be very difficult to enforce.
+               // Also as children of virtual guard is very self contained and atm it is very unlikely that other optimizations are going to find opportunity of manipulating them and
+               // Because of the fact that it is very unlikely that we will have another aloadi node with same VTable offset of same receiver object, this child would not be commoned out
+               // and have only single reference in this virtual guard therefore splitting of block will not store it to temp slot.
+               // In rare case child of the virtual guard is manipulated then illegal memory reference load would hace occured before the Virtual Guard which 
+               // is already a bug as mentioned in the description of this function and it would be safer to fail compilation. 
+               TR::Node *vTableLoad = node->getFirstChild();
+               if (!(vTableLoad->getOpCodeValue() == TR::aloadi && comp->getSymRefTab()->isVtableEntrySymbolRef(vTableLoad->getSymbolReference())))
```
r30shah: Changed.

andrewcraik:

r30shah:
Seems like PPCLE failure  is same as https://github.com/eclipse/openj9/issues/4470

andrewcraik:
Given the failure is the same as #4470 I think this should be good to go.

PR #4722, opened 2019-02-13
Handle the receiving of AOT code on JITaaS client

harryyu1994:
- rename original remoteCompilationEnd() to outOfProcessCompilationEnd() because in our terms, remoteCompile indicates that they are client routines and outOfProcessCompile indicates they are server routines. In this case, the original remoteCompilationEnd() is a server routine that deals with the packing and sending of out-of-process compiled code.
- create new remoteCompilationEnd(), a client routine, to handle the installation of compiled code for both remote JIT and remote AOT compilations. For AOT compilations, they first get written to the Shared Class Cache if possible.

Issue: #4401
[skip ci]

### DIFF
--- a/runtime/compiler/control/JITaaSCompilationThread.cpp
+++ b/runtime/compiler/control/JITaaSCompilationThread.cpp
@@ -38,6 +38,11 @@
 #include "j9port.h" // for j9time_current_time_millis
 #include "env/j9methodServer.hpp"
 #include "exceptions/AOTFailure.hpp" // for AOTFailure
+#include "env/TRMemory.hpp" // for jitPersistentAlloc and jitPersistentFree
+
+#if defined(J9VM_OPT_SHARED_CLASSES)
+#include "j9jitnls.h"
+#endif
 
 extern TR::Monitor *assumptionTableMutex;
 
@@ -2265,30 +2270,11 @@ remoteCompile(
             compiler->exitHeuristicRegion();
             }
 
-         compInfoPT->reloRuntime()->setReloStartTime(compInfoPT->getTimeWhenCompStarted());
-
          TR_ASSERT(codeCacheStr.size(), "must have code cache");
          TR_ASSERT(dataCacheStr.size(), "must have data cache");
 
-         metaData = compInfoPT->reloRuntime()->prepareRelocateJITCodeAndData(vmThread, compiler->fej9vm(),
-               compiler->cg()->getCodeCache(), (uint8_t *)&codeCacheStr[0], (J9JITDataCacheHeader *)&dataCacheStr[0],
-               method, false, TR::comp()->getOptions(), TR::comp(), compilee);
-
-         TR::compInfoPT->setMetadata(metaData);
-
-         if (!metaData)
-            {
-            if (TR::Options::getVerboseOption(TR_VerboseJITaaS))
-               {
-               TR_VerboseLog::writeLineLocked(TR_Vlog_JITaaS,
-                                              "JITaaS Relocation failure: %d",
-                                              compInfoPT->reloRuntime()->returnCode());
-               }
-            // relocation failed, fail compilation
-            compInfoPT->getMethodBeingCompiled()->_compErrCode = compInfoPT->reloRuntime()->returnCode();
-            compiler->failCompilation<J9::AOTRelocationFailed>("Failed to relocate");
-            }
-         // TR_ASSERT(metaData, "relocation must succeed");
+         // relocate the received compiled code
+         metaData = remoteCompilationEnd(vmThread, compiler, compilee, method, compInfoPT, codeCacheStr, dataCacheStr);
 
          if (!TR::comp()->getOption(TR_DisableCHOpts))
             {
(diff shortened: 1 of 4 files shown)


harryyu1994:
Not sure if these hooks are absolutely needed, they can be added to `remoteCompilationEnd()` if they are:
```
if (J9_EVENT_IS_HOOKED(jitConfig->javaVM->hookInterface, J9HOOK_VM_DYNAMIC_CODE_LOAD))
                        {
                        OMR::CodeCacheMethodHeader *ccMethodHeader;
                        ALWAYS_TRIGGER_J9HOOK_VM_DYNAMIC_CODE_LOAD(
                           jitConfig->javaVM->hookInterface,
                           vmThread,
                           method,
                           reinterpret_cast<void *>(relocatedMetaData->startPC),
                           relocatedMetaData->endWarmPC - relocatedMetaData->startPC,
                           "JIT warm body",
                           relocatedMetaData
                           );
                        if (relocatedMetaData->startColdPC)
                           {
                           // Register the cold code section too
                           //
                           ALWAYS_TRIGGER_J9HOOK_VM_DYNAMIC_CODE_LOAD(
                              jitConfig->javaVM->hookInterface,
                              vmThread,
                              method,
                              reinterpret_cast<void *>(relocatedMetaData->startColdPC),
                              relocatedMetaData->endPC - relocatedMetaData->startColdPC,
                              "JIT cold body",
                              relocatedMetaData
                              );
                           }
                        ccMethodHeader = getCodeCacheMethodHeader(
                           reinterpret_cast<char *>(relocatedMetaData->startPC),
                           32,
                           relocatedMetaData
                           );
                        if (ccMethodHeader && (relocatedMetaData->bodyInfo != NULL))
                           {
                           TR_LinkageInfo *linkageInfo = TR_LinkageInfo::get(reinterpret_cast<void *>(relocatedMetaData->startPC));
                           if (linkageInfo->isRecompMethodBody())
                              {
                              ALWAYS_TRIGGER_J9HOOK_VM_DYNAMIC_CODE_LOAD(
                                 jitConfig->javaVM->hookInterface,
                                 vmThread,
                                 method,
                                 ccMethodHeader->_eyeCatcher + 4,
                                 relocatedMetaData->startPC - reinterpret_cast<uintptr_t>(ccMethodHeader->_eyeCatcher + 4),
                                 "JIT method header",
                                 relocatedMetaData
                                 );
                              }
                           }
                        }
```

Another question is: are compilation/slot monitors required for any of the operations here?

harryyu1994:
I accidentally closed the original PR and then did a force push. So the original PR can't be re-opened now. Creating this new one.

harryyu1994:
@mpirvu @ymanton This change is ready for a review.

ymanton:
> Another question is: are compilation/slot monitors required for any of the operations here?

Not sure about slot, but in a lot of places where we expect to have the compilation monitor the code will check with an assert, hopefully that is sufficient to catch any missed cases once everything comes together.

harryyu1994:
Thanks for the review, I'm going to run some testing now before removing the WIP tag. The other question I had was if compilation/slot monitors are required for any of the operations here. In the original `compilationEnd` routine it's called with those locks, and released before calling `prepareRelocateAOTCodeAndData`. I didn't feel like they were needed for any of the operations here but just wanna make sure.

harryyu1994:
Okay so we should put the compilation monitor in for now?

harryyu1994:
Ok got it. Leaving the monitors out for now.

runtime/compiler/control/JITaaSCompilationThread.cpp
```diff
@@ -2363,8 +2349,230 @@ remoteCompile(
...
+      if (canRelocateMethod)
+         {
+         J9JITDataCacheHeader *cacheEntry;
+
+         TR_ASSERT_FATAL(comp->cg(), "CodeGenerator must be allocated");
+
+         // Use same code cache as AOT compile
+         //
+         TR::CodeCache *aotMCCRuntimeCodeCache = comp->cg()->getCodeCache();
+         TR_ASSERT(aotMCCRuntimeCodeCache, "Must have a reserved codeCache");
```
dmitry-ten: When I tried running with this PR, this assert was triggered.
Normally, JIT would reserve code cache during compilation, but for JITaaS we only reserve the code cache on the server. I'm not sure if it's safe to just remove this assert here.

harryyu1994: I think we might need to reserve a code cache for JITaaS client remote AOT compilations then.  (If it's used for what I think it's used for, storing the relocated code). I might have disabled it somewhere in this commit. Looking into it.

harryyu1994: If there isn't a code cache available, allocateSpaceInCodeCache is going to allocate one for us, so I guess it's safe. But I'm not sure if we should still reserve it first. Going to find out.
```
U_8 *
TR_SharedCacheRelocationRuntime::allocateSpaceInCodeCache(UDATA codeSize)
   {
   TR_J9VMBase *fej9 = (TR_J9VMBase *)_fe;
   TR::CodeCacheManager *manager = TR::CodeCacheManager::instance();

   int32_t compThreadID = fej9->getCompThreadIDForVMThread(_currentThread);
   if (!codeCache())
      {
      int32_t numReserved;

      _codeCache = manager->reserveCodeCache(false, codeSize, compThreadID, &numReserved);  // Acquire a cold/warm code cache.
...
```

runtime/compiler/control/JITaaSCompilationThread.cpp
```diff
@@ -2363,8 +2349,230 @@ remoteCompile(
...
+            TR_VerboseLog::writeLineLocked(
+               TR_Vlog_JITaaS,
+               "Applying JITaaS remote AOT relocations to newly AOT compiled body for %s @ %s",
+               comp->signature(),
+               comp->getHotnessName()
+               );
+            }
+         try
+            {
+            TR::CompilationInfoPerThreadBase::InterruptibleOperation(*entry->_compInfoPT);
```
dmitry-ten: This line should be removed, I think.
At this point we are alredy inside Interruptible operation region, which is started in `TR::CompilationInfoPerThreadBase::compile`, so trying to enter it again triggers assert.

runtime/compiler/control/JITaaSCompilationThread.cpp
harryyu1994: For normal JIT, we reserve the code cache at `performReserveCodeCachePhase` stage. We shouldn't have a code cache at this point for a remote compile. I think we can safely remove the assert.
```
void
OMR::CodeGenPhase::performReserveCodeCachePhase(TR::CodeGenerator * cg, TR::CodeGenPhase * phase)
   {
   cg->reserveCodeCache();
   }
```

harryyu1994:
The hooks will be added in another PR. This PR is ready for merge.

PR #4723, opened 2019-02-13
Enable SVM during startup

dsouzai:
Fixes https://github.com/eclipse/openj9/issues/4138
Fixes https://github.com/eclipse/openj9/issues/4201

### DIFF
--- a/runtime/compiler/control/CompilationThread.cpp
+++ b/runtime/compiler/control/CompilationThread.cpp
@@ -7442,12 +7442,6 @@ TR::CompilationInfoPerThreadBase::wrappedCompile(J9PortLibrary *portLib, void * 
                   options->setOption(TR_EnableGRACostBenefitModel, false);
                }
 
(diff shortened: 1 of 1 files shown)


dsouzai:

dsouzai:
I verified that enabling the SVM during startup does not result in a startup regression, thanks to all of @jdmpapin 's excellent work in https://github.com/eclipse/openj9/issues/3752

dsouzai:
@mpirvu for your review.

mpirvu:
There is one failure in Java11 LinuxAMD64
```
18:51:23         FAILED: testAnonClassCodePaths
18:51:23         java.lang.NullPointerException
18:51:23         	at java.base/java.lang.Integer.valueOf(Integer.java:1049)
```
Is there any chance the failure is due to SVM being enabled during startup?

dsouzai:
No, doesn't look like it. The test was run with
```
JVM_OPTIONS: -Xnocompressedrefs -Xjit -Xgcpolicy:gencon -Xnocompressedrefs 
```
so there was no AOT code.

PR #4724, opened 2019-02-13
Enforce a return value of 0 or 1 for JNI boolean

NigelYiboYu:
Make sure JNI boolean return value is either 0 or 1 by loading a 'one' into
the return value register if it's not zero.

Slightly refactor JNI linkage return value processing code.

### DIFF
--- a/runtime/compiler/z/codegen/J9S390PrivateLinkage.cpp
+++ b/runtime/compiler/z/codegen/J9S390PrivateLinkage.cpp
@@ -2823,6 +2823,55 @@ void TR::J9S390JNILinkage::checkException(TR::Node * callNode,
    generateS390LabelInstruction(self()->cg(), TR::InstOpCode::LABEL, callNode, exceptionRestartLabel);
    }
 
+void
+TR::J9S390JNILinkage::processJNIReturnValue(TR::Node * callNode,
+                                            TR::CodeGenerator* cg,
+                                            TR::Register* javaReturnRegister)
+   {
+   auto resolvedMethod = callNode->getSymbol()->castToResolvedMethodSymbol()->getResolvedMethod();
+   auto returnType = resolvedMethod->returnType();
+   const bool isUnwrapAddressReturnValue = !((TR_J9VMBase *)fe())->jniDoNotWrapObjects(resolvedMethod)
+                                            && (returnType == TR::Address);
+
+   TR::LabelSymbol *cFlowRegionStart = NULL, *cFlowRegionEnd = NULL;
+
+   if (isUnwrapAddressReturnValue)
+      {
+      cFlowRegionStart = generateLabelSymbol(cg);
+      cFlowRegionEnd = generateLabelSymbol(cg);
+
+      generateS390LabelInstruction(cg, TR::InstOpCode::LABEL, callNode, cFlowRegionStart);
+      cFlowRegionStart->setStartInternalControlFlow();
+      generateS390CompareAndBranchInstruction(cg, TR::InstOpCode::getCmpOpCode(), callNode, javaReturnRegister, 0, TR::InstOpCode::COND_BE, cFlowRegionEnd);
+      generateRXInstruction(cg, TR::InstOpCode::getLoadOpCode(), callNode, javaReturnRegister,
+                            generateS390MemoryReference(javaReturnRegister, 0, cg));
+
+      generateS390LabelInstruction(cg, TR::InstOpCode::LABEL, callNode, cFlowRegionEnd);
+      cFlowRegionEnd->setEndInternalControlFlow();
+      }
+   else if ((returnType == TR::Int8) && comp()->getSymRefTab()->isReturnTypeBool(callNode->getSymbolReference()))
+      {
+      if (cg->getS390ProcessorInfo()->supportsArch(TR_S390ProcessorInfo::TR_z13))
+         {
+         generateRIInstruction(cg, TR::InstOpCode::getCmpHalfWordImmOpCode(), callNode, javaReturnRegister, 0);
+         generateRIEInstruction(cg, TR::Compiler->target.is64Bit() ? TR::InstOpCode::LOCGHI : TR::InstOpCode::LOCHI,
+                                callNode, javaReturnRegister, 1, TR::InstOpCode::COND_BNE);
+         }
+      else
+         {
+         cFlowRegionStart = generateLabelSymbol(cg);
+         cFlowRegionEnd = generateLabelSymbol(cg);
+
+         generateS390LabelInstruction(cg, TR::InstOpCode::LABEL, callNode, cFlowRegionStart);
+         cFlowRegionStart->setStartInternalControlFlow();
+         generateS390CompareAndBranchInstruction(cg, TR::InstOpCode::getCmpOpCode(), callNode, javaReturnRegister,
+                                                 0, TR::InstOpCode::COND_BE, cFlowRegionEnd);
+         generateRIInstruction(cg, TR::InstOpCode::getLoadHalfWordImmOpCode(), callNode, javaReturnRegister, 1);
+         generateS390LabelInstruction(cg, TR::InstOpCode::LABEL, callNode, cFlowRegionEnd);
+         cFlowRegionEnd->setEndInternalControlFlow();
+         }
+      }
+   }
 
 TR::Register * TR::J9S390JNILinkage::buildDirectDispatch(TR::Node * callNode)
    {
@@ -2887,7 +2936,6 @@ TR::Register * TR::J9S390JNILinkage::buildDirectDispatch(TR::Node * callNode)
    bool isAcquireVMAccess = isReleaseVMAccess;
    bool isCollapseJNIReferenceFrame = !fej9->jniNoSpecialTeardown(resolvedMethod);
    bool isCheckException = !fej9->jniNoExceptionsThrown(resolvedMethod);
-   bool isUnwrapAddressReturnValue = !fej9->jniDoNotWrapObjects(resolvedMethod);
    bool isKillAllUnlockedGPRs = isJNIGCPoint;
 
    killMask = killAndAssignRegister(killMask, deps, &methodAddressReg, (TR::Compiler->target.isLinux()) ?  TR::RealRegister::GPR1 : TR::RealRegister::GPR9 , codeGen, true);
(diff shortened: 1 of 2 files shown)


NigelYiboYu:
Marked as WIP until all tests are clear.

NigelYiboYu:
@fjeremic  FYI

runtime/compiler/z/codegen/J9S390PrivateLinkage.cpp
```diff
@@ -2823,6 +2823,73 @@ void TR::J9S390JNILinkage::checkException(TR::Node * callNode,
    generateS390LabelInstruction(self()->cg(), TR::InstOpCode::LABEL, callNode, exceptionRestartLabel);
    }
 
+/**
+ * \brief JNI return value processing:
+ * 1) Unwrap return value for objects or
+ * 2) Enforce a return value of 0 or 1 for boolean return type.
+*/
```
fjeremic: Could we Doxygen document this new API and move the documentation to the API definition (header file)?

runtime/compiler/z/codegen/J9S390PrivateLinkage.cpp
```diff
@@ -2823,6 +2823,73 @@ void TR::J9S390JNILinkage::checkException(TR::Node * callNode,
...
+                                            TR::CodeGenerator* cg,
+                                            TR::Register* javaReturnRegister,
+                                            TR::RegisterDependencyConditions* deps)
+   {
+   auto resolvedMethod = callNode->getSymbol()->castToResolvedMethodSymbol()->getResolvedMethod();
+   auto returnType = resolvedMethod->returnType();
+   bool isUnwrapAddressReturnValue = !((TR_J9VMBase *)fe())->jniDoNotWrapObjects(resolvedMethod)
+                                     && (returnType == TR::Address)
+                                     && (javaReturnRegister!= NULL)
+                                     && (javaReturnRegister->getKind() == TR_GPR);
```
fjeremic: Why do we need to check both the `returnType` and `javaReturnRegister->getKind()`? If the `returnType` is an address we should be guaranteed for it to be in a GPR.

runtime/compiler/z/codegen/J9S390PrivateLinkage.cpp
```diff
@@ -2823,6 +2823,73 @@ void TR::J9S390JNILinkage::checkException(TR::Node * callNode,
...
+   TR::LabelSymbol *cFlowRegionStart = NULL, *cFlowRegionEnd = NULL;
+
+   if (isUnwrapAddressReturnValue)
+      {
+      cFlowRegionStart = generateLabelSymbol(cg);
+      cFlowRegionEnd = generateLabelSymbol(cg);
+
+      generateS390LabelInstruction(cg, TR::InstOpCode::LABEL, callNode, cFlowRegionStart);
+      cFlowRegionStart->setStartInternalControlFlow();
+      generateS390CompareAndBranchInstruction(cg, TR::InstOpCode::getCmpOpCode(), callNode, javaReturnRegister, (signed char)0, TR::InstOpCode::COND_BE, cFlowRegionEnd);
```
fjeremic: I don't think we need to cast to a `(signed char)` here.

runtime/compiler/z/codegen/J9S390PrivateLinkage.cpp
```diff
@@ -2823,6 +2823,73 @@ void TR::J9S390JNILinkage::checkException(TR::Node * callNode,
...
+
+         generateS390LabelInstruction(cg, TR::InstOpCode::LABEL, callNode, cFlowRegionStart);
+         cFlowRegionStart->setStartInternalControlFlow();
+         generateS390CompareAndBranchInstruction(cg, TR::InstOpCode::getCmpOpCode(), callNode, javaReturnRegister,
+                                                 (signed char)0, TR::InstOpCode::COND_BE, cFlowRegionEnd);
+         generateRIInstruction(cg, TR::InstOpCode::getLoadHalfWordImmOpCode(), callNode, javaReturnRegister, 1);
+         int32_t regPos = deps->searchPostConditionRegisterPos(javaReturnRegister);
+         TR::RealRegister::RegNum realReg = deps->getPostConditions()->getRegisterDependency(regPos)->getRealRegister();
+         TR::RegisterDependencyConditions * postDeps = new (trHeapMemory()) TR::RegisterDependencyConditions(0, 1, cg);
+         postDeps->addPostCondition(javaReturnRegister, realReg);
```
fjeremic: Why exactly do we need this post condition? The Java return register is already in the `deps` post conditions which I think are generated on the last label of the JNI dispatch, thus making this post condition redundant. Can you generate a JIT trace of what the evaluator looks like and post it here so we can confirm?

runtime/compiler/z/codegen/J9S390PrivateLinkage.cpp
```diff
@@ -2823,6 +2823,73 @@ void TR::J9S390JNILinkage::checkException(TR::Node * callNode,
...
+      postDeps->addPostCondition(javaReturnRegister, realReg);
+
+      generateS390LabelInstruction(cg, TR::InstOpCode::LABEL, callNode, cFlowRegionEnd, postDeps);
+      cFlowRegionEnd->setEndInternalControlFlow();
+      }
+   else if ((returnType == TR::Int8) && comp()->getSymRefTab()->isReturnTypeBool(callNode->getSymbolReference()))
+      {
+      if (cg->getS390ProcessorInfo()->supportsArch(TR_S390ProcessorInfo::TR_z13))
+         {
+         generateRIInstruction(cg, TR::InstOpCode::getCmpHalfWordImmOpCode(), callNode, javaReturnRegister, (int32_t) 0);
```
fjeremic: I don't think we need a `(int32_t)` cast here.

runtime/compiler/z/codegen/J9S390PrivateLinkage.cpp
fjeremic: Similarly below.

runtime/compiler/z/codegen/J9S390PrivateLinkage.cpp
NigelYiboYu: Right. This is moved from the old source and I didn't take a second look at this. I'll remove the GPR check.

runtime/compiler/z/codegen/J9S390PrivateLinkage.cpp
NigelYiboYu: This post dependency is unnecessary and again is something inherited from the old code. The JNI dispatch has a huge `deps` at the very end that already contains the return register. I've removed and re-tested the change.

runtime/compiler/z/codegen/J9S390PrivateLinkage.cpp
NigelYiboYu: Fixed.

runtime/compiler/z/codegen/J9S390PrivateLinkage.cpp
NigelYiboYu: Fixed.

runtime/compiler/z/codegen/J9S390PrivateLinkage.cpp
NigelYiboYu: Done.

runtime/compiler/z/codegen/J9S390PrivateLinkage.cpp
```diff
@@ -2823,6 +2823,56 @@ void TR::J9S390JNILinkage::checkException(TR::Node * callNode,
...
+void
+TR::J9S390JNILinkage::processJNIReturnValue(TR::Node * callNode,
+                                            TR::CodeGenerator* cg,
+                                            TR::Register* javaReturnRegister)
+   {
+   auto resolvedMethod = callNode->getSymbol()->castToResolvedMethodSymbol()->getResolvedMethod();
+   auto returnType = resolvedMethod->returnType();
+   bool isUnwrapAddressReturnValue = !((TR_J9VMBase *)fe())->jniDoNotWrapObjects(resolvedMethod)
+                                     && (returnType == TR::Address)
+                                     && (javaReturnRegister!= NULL);
```
fjeremic: We also don't need a `javaReturnRegister!= NULL` check here because if the type is address then it must have a return value, and hence a return register. Also we can make this variable assignment `const`.

fjeremic:
```suggestion

```

runtime/compiler/z/codegen/J9S390PrivateLinkage.cpp
```diff
@@ -2823,6 +2823,56 @@ void TR::J9S390JNILinkage::checkException(TR::Node * callNode,
...
 
+void
+TR::J9S390JNILinkage::processJNIReturnValue(TR::Node * callNode,
+                                            TR::CodeGenerator* cg,
+                                            TR::Register* javaReturnRegister)
+   {
+   auto resolvedMethod = callNode->getSymbol()->castToResolvedMethodSymbol()->getResolvedMethod();
+   auto returnType = resolvedMethod->returnType();
+   bool isUnwrapAddressReturnValue = !((TR_J9VMBase *)fe())->jniDoNotWrapObjects(resolvedMethod)
+                                     && (returnType == TR::Address)
```
fjeremic:
```suggestion
                                     && (returnType == TR::Address);
```

runtime/compiler/z/codegen/J9S390PrivateLinkage.cpp
```diff
@@ -2823,6 +2823,56 @@ void TR::J9S390JNILinkage::checkException(TR::Node * callNode,
...
    }
 
+void
+TR::J9S390JNILinkage::processJNIReturnValue(TR::Node * callNode,
+                                            TR::CodeGenerator* cg,
+                                            TR::Register* javaReturnRegister)
+   {
+   auto resolvedMethod = callNode->getSymbol()->castToResolvedMethodSymbol()->getResolvedMethod();
+   auto returnType = resolvedMethod->returnType();
+   bool isUnwrapAddressReturnValue = !((TR_J9VMBase *)fe())->jniDoNotWrapObjects(resolvedMethod)
```
fjeremic:
```suggestion
   const bool isUnwrapAddressReturnValue = !((TR_J9VMBase *)fe())->jniDoNotWrapObjects(resolvedMethod)
```

fjeremic:

PR #4726, opened 2019-02-13
Remove TR_SharedCacheJITaaSRelocationRuntime and add asserts

harryyu1994:
Issue: #4401
[skip ci]

### DIFF
--- a/runtime/compiler/runtime/RelocationRuntime.hpp
+++ b/runtime/compiler/runtime/RelocationRuntime.hpp
@@ -401,32 +401,22 @@ public:
       TR_ALLOC(TR_Memory::Relocation)
       void * operator new(size_t, J9JITConfig *);
       TR_JITaaSRelocationRuntime(J9JITConfig *jitCfg) : TR_RelocationRuntime(jitCfg) {}
+// The following public APIs should not be used with this class
+      virtual bool storeAOTHeader(J9JavaVM *javaVM, TR_FrontEnd *fe, J9VMThread *curThread)  override { TR_ASSERT(0, "Should not be called in this RelocationRuntime!"); return 0;}
+      virtual TR_AOTHeader *createAOTHeader(J9JavaVM *javaVM, TR_FrontEnd *fe)  override { TR_ASSERT(0, "Should not be called in this RelocationRuntime!"); return 0;}
+      virtual bool validateAOTHeader(J9JavaVM *javaVM, TR_FrontEnd *fe, J9VMThread *curThread)  override { TR_ASSERT(0, "Should not be called in this RelocationRuntime!"); return 0;}
+
+      virtual void *isROMClassInSharedCaches(UDATA romClassValue, J9JavaVM *javaVM)  override { TR_ASSERT(0, "Should not be called in this RelocationRuntime!"); return 0; }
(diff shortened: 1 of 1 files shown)


runtime/compiler/runtime/RelocationRuntime.hpp
```diff
@@ -401,32 +401,22 @@ class TR_JITaaSRelocationRuntime : public TR_RelocationRuntime {
...
+      virtual TR_AOTHeader *createAOTHeader(J9JavaVM *javaVM, TR_FrontEnd *fe)  override { TR_ASSERT(0, "Should not be called in this RelocationRuntime!"); return 0;}
+      virtual bool validateAOTHeader(J9JavaVM *javaVM, TR_FrontEnd *fe, J9VMThread *curThread)  override { TR_ASSERT(0, "Should not be called in this RelocationRuntime!"); return 0;}
+
+      virtual void *isROMClassInSharedCaches(UDATA romClassValue, J9JavaVM *javaVM)  override { TR_ASSERT(0, "Should not be called in this RelocationRuntime!"); return 0; }
+      virtual bool isRomClassForMethodInSharedCache(J9Method *method, J9JavaVM *javaVM)  override { TR_ASSERT(0, "Should not be called in this RelocationRuntime!"); return 0; }
+      virtual TR_YesNoMaybe isMethodInSharedCache(J9Method *method, J9JavaVM *javaVM)  override { TR_ASSERT(0, "Should not be called in this RelocationRuntime!");  return TR_no; }
+      virtual TR_OpaqueClassBlock *getClassFromCP(J9VMThread *vmThread, J9JavaVM *javaVM, J9ConstantPool *constantPool, I_32 cpIndex, bool isStatic)  override { TR_ASSERT(0, "Should not be called in this RelocationRuntime!"); return 0; }
```
ymanton: I'm confused by the comment and the asserts. Are the asserts going to be replaced with code that does something on the JITaaS client or are these methods not supposed to be called?

harryyu1994: These methods are not supposed to be called at all. We realized that a `TR_SharedCacheJITaaSRelocationRuntime` class is not needed for JITaaS server and JITaaS client can just use the existing `TR_SharedCacheRelocationRuntime`.  Added these APIs & asserts to make sure JITaaS server never calls these APIs, and just in case things change in the future.

harryyu1994: JITaaS client calls the existing `TR_SharedCacheRelocationRuntime` version of the APIs.

runtime/compiler/runtime/RelocationRuntime.hpp
```diff
@@ -401,32 +401,22 @@ class TR_JITaaSRelocationRuntime : public TR_RelocationRuntime {
       TR_ALLOC(TR_Memory::Relocation)
       void * operator new(size_t, J9JITConfig *);
       TR_JITaaSRelocationRuntime(J9JITConfig *jitCfg) : TR_RelocationRuntime(jitCfg) {}
+// following public APIs should only be called on JITaaS client
```
ymanton:
```suggestion
// The following public APIs should not be used with this class
```

runtime/compiler/runtime/RelocationRuntime.hpp
ymanton: OK thanks, the comment is really confusing and made it should like the client should be calling these methods in this class.

harryyu1994:
Comments updated.

PR #4727, opened 2019-02-13
Disable callerClassLoader optimization under AOT

aviansie-ben:
Previously, `j/l/ClassLoader.callerClassLoader` was optimized during IL
generation if it was inlined into a method which itself was inlined into
another method. Under these circumstances, the address of the calling
method's class loader would be hardcoded into the instruction stream.
This causes problems under AOT, since the address of the caller's class
loader was not being properly relocated. Since there is not yet any
relocation that could handle this, this particular optimization has been
disabled under AOT for the time being.


dsouzai:

PR #4728, opened 2019-02-13
Ensure case conversion inrinsic return zero or one on X86

0dvictor:
com/ibm/jit/JITHelpers.to{Upper,Lower}Intrinsic{UTF16,Latin1} returns a boolean and hence the returned value should be either zero or one per current Java specs.


fjeremic:
This is already fixed as part of #4688 which contains a more complete fix. If you agree please close the PR and have a review at the x86 commit in #4688.

0dvictor:
Closing this PR as #4688 contains the required changes.

PR #4729, opened 2019-02-13
Fix improper J9Class materialization during IL generation

aviansie-ben:
There were a couple of instances during IL generation where the
internals of a J9Class were used to get access to a related J9Class
without using the proper APIs to do so. This was causing problems under
SVM AOT, since no validation records would be emitted when this
occurred. These have been fixed to use the proper APIs, which should
ensure that the correct validation records get emitted.

### DIFF
--- a/runtime/compiler/env/j9fieldsInfo.cpp
+++ b/runtime/compiler/env/j9fieldsInfo.cpp
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2000, 2018 IBM Corp. and others
+ * Copyright (c) 2000, 2019 IBM Corp. and others
  *
  * This program and the accompanying materials are made available under
  * the terms of the Eclipse Public License 2.0 which accompanies this
@@ -149,11 +149,12 @@ TR_VMFieldsInfo::TR_VMFieldsInfo(TR::Compilation * comp, J9Class *aClazz, int bu
       }
 
    //supers
-   J9Class **supClasses = aClazz->superclasses;
    int numSupClasses = J9CLASS_DEPTH(aClazz);
+   J9Class *supClass = aClazz;
    for (i=numSupClasses-1; i>=0; i--)
       {
-      J9Class *supClass = supClasses[i];
+      supClass = (J9Class*)comp->fej9()->getSuperClass((TR_OpaqueClassBlock*)supClass);
+      TR_ASSERT(supClass, "Found NULL supClass in inheritance chain");
       romCl = supClass->romClass;
 
       // iterate through the fields creating TR_VMField and inserting them into a List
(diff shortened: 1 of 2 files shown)


runtime/compiler/env/j9fieldsInfo.cpp
```diff
@@ -149,11 +149,12 @@ TR_VMFieldsInfo::TR_VMFieldsInfo(TR::Compilation * comp, J9Class *aClazz, int bu
...
 
    //supers
-   J9Class **supClasses = aClazz->superclasses;
    int numSupClasses = J9CLASS_DEPTH(aClazz);
+   J9Class *supClass = aClazz;
    for (i=numSupClasses-1; i>=0; i--)
       {
-      J9Class *supClass = supClasses[i];
+      supClass = (J9Class*)comp->fej9()->getSuperClass((TR_OpaqueClassBlock*)supClass);
+      TR_ASSERT(supClass, "Found NULL supClass in inheritance chain");
```
dsouzai: :+1:

dsouzai:
@jdmpapin could I get a review from you?

runtime/compiler/env/j9fieldsInfo.cpp
```diff
@@ -149,11 +149,12 @@ TR_VMFieldsInfo::TR_VMFieldsInfo(TR::Compilation * comp, J9Class *aClazz, int bu
       }
 
    //supers
-   J9Class **supClasses = aClazz->superclasses;
    int numSupClasses = J9CLASS_DEPTH(aClazz);
+   J9Class *supClass = aClazz;
    for (i=numSupClasses-1; i>=0; i--)
       {
-      J9Class *supClass = supClasses[i];
+      supClass = (J9Class*)comp->fej9()->getSuperClass((TR_OpaqueClassBlock*)supClass);
```
jdmpapin: At least at the use of this in ilgen, it's unfortunate to dig through all these superclasses just to [filter out fields inherited from them](https://github.com/eclipse/openj9/blob/35c44ead49a53f8c0ad1ca3e63c18711d3a7b26f/runtime/compiler/ilgen/Walker.cpp#L5515). If we find that this generates too many superclass records, we may want to put this use into a heuristic region. The information we're extracting is whether or not `methodClass` declares any of its own final (instance) fields, and while the use of that information isn't strictly heuristic, it is completely determined by the class chain of `methodClass`

runtime/compiler/env/j9method.cpp
```diff
@@ -8008,7 +8008,7 @@ TR_J9ByteCodeIlGenerator::runFEMacro(TR::SymbolReference *symRef)
          // Construct the signature string for the array class
          //
          UDATA arity = arrayJ9Class->arity;
-         J9Class *leafClass = arrayJ9Class->leafComponentType;
+         J9Class *leafClass = (J9Class*)fej9->getLeafComponentClassFromArrayClass((TR_OpaqueClassBlock*)arrayJ9Class);
```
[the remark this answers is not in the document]
aviansie-ben: I'm honestly not sure if this ends up getting executed during AOT (I haven't seen it executed myself). I noticed this while fixing the other issue and figured I'd change this to use the front-end API pre-emptively to prevent any unwanted surprises if this ever does end up getting executed during AOT.

dsouzai:
jenkins test sanity all jdk11

dsouzai:
jenkins test sanity all jdk11

PR #4731, opened 2019-02-13
Reimplement storeValidationRecordIfNecessary

dmitry-ten:
Since we are no longer overriding methods of J9SharedCache,
and J9SharedCacheRelocationRuntime, calls to their methods
should be handled in the front-end callers.
storeValidationRecordIfNecessary will now make a remote call
to remember class and get class from CP


dmitry-ten:
@mpirvu @ymanton could you please review?

PR #4736, opened 2019-02-14
Initialize TEST_FLAG to blank

AdamBrousseau:
- Avoid Jenkins error trying to pass parameter with null value

[skip ci]

### DIFF
--- a/buildenv/jenkins/common/pipeline-functions
+++ b/buildenv/jenkins/common/pipeline-functions
@@ -380,7 +380,7 @@ def workflow(SDK_VERSION, SPEC, SHAS, OPENJDK_REPO, OPENJDK_BRANCH, OPENJ9_REPO,
             echo "Using ARTIFACTORY_SERVER:'${ARTIFACTORY_SERVER}'"
         }
         for (name in TARGET_NAMES) {
-            def TEST_FLAG
+            def TEST_FLAG = ''
             if (name.contains('+jitaas')) {
                 name -= '+jitaas'
                 TEST_FLAG = 'JITAAS'


AdamBrousseau:
Yesterday we added an option to one of our Jenkins instances which allows parameters to be passed to a job without the job having such parameters defined and not writing a warning to the main Jenkins log.

Example log entry
```
hudson.model.ParametersAction filter WARNING: Skipped parameter `VARIABLE_FILE` as it is undefined on `Test-sanity.functional-JDK11-linux_390-64_cmprssptrs`. Set `-Dhudson.model.ParametersAction.keepUndefinedParameters=true` to allow undefined parameters to be injected as environment variables or `-Dhudson.model.ParametersAction.safeParameters=[comma-separated list]` to whitelist specific parameter names, even though it represents a security breach or `-Dhudson.model.ParametersAction.keepUndefinedParameters=false` to no longer show this message.
```
Adding the option `-Dhudson.model.ParametersAction.keepUndefinedParameters=true` prevents these warnings flooding the log.

However, it seemed to have the unintended side effect of not allowing `null` parameters to be passed.
We are seeing this error on Test jobs
```
java.lang.IllegalArgumentException: Null value not allowed as an environment variable: TEST_FLAG
```

Initializing this variable to blank resolves the error.

@pshipton or @smlambert for merge

DanHeidinga:
@AdamBrousseau Can you link this to the issue / pr for adding `-Dhudson.model.ParametersAction.keepUndefinedParameters=true`?  It'll make it easier to understand the dependencies between these changes in the future.

AdamBrousseau:
There was no issue. It was on an internal Jenkins instance. It came about after chasing a disk space issue with the support team on Slack and realizing the Jenkins log file was eating 50G of disk mostly due to these warnings. The support team added the option and I noticed Test jobs failing the next day so I am assuming the link between the change and the failures.

DanHeidinga:
@AdamBrousseau Isn't adding that option a security concern?  https://jenkins.io/blog/2016/05/11/security-update/ it suggests that not passing the variables was a low / medium security concern as it allows injecting extra env vars to Jenkins processes.  Do any of the suggested alternatives in that doc better address the root issue?

AdamBrousseau:
Not really a security concern imo.
1) It was added to an internal jenkins.
2) You still need permissions to launch builds in order to launch a build with rogue params. If someone has permissions to launch builds, they've been approved the access and are trusted. Even if we added it to the Eclipse Jenkins only committers would be able to take advantage. 

AdoptOpenJDK/openjdk-tests#837 was opened to address this, but it looks like the better option which is being investigated is too auto generate jobs that would have all the necessary parameters.

DanHeidinga:
There's no way to PR build this.  Merging as is.

Issue #4734, opened 2019-02-13
java.security.manager=disallow in jdk12


pshipton:
See setSecurityManager() in http://cr.openjdk.java.net/~iris/se/12/build/latest/api/java.base/java/lang/System.html

```
Implementation Note:
In the JDK implementation, if the Java virtual machine is started with the system property java.security.manager set to the special token "disallow" then the setSecurityManager method cannot be used to set a security manager
```

pshipton:
@JasonFengJ9 @DanHeidinga

JasonFengJ9:
Will take this.

pshipton:
Also see the table in the description for http://cr.openjdk.java.net/~iris/se/12/build/latest/api/java.base/java/lang/SecurityManager.html

"allow" is also permitted.

pshipton:
Might check with a test case to see if jdk8 should support "disallow" and "allow".

PR #4738, opened 2019-02-14
JDK12 setSecurityManager according to java.security.manager special tokens

JasonFengJ9:
`JDK12` `setSecurityManager()` according to system property `java.security.manager` special tokens

`JDK12` throws `UnsupportedOperationException` when the system property `java.security.manager` is set to special token `disallow` during startup via command line options;
In addition, special token `allow` permits `setSecurityManager` dynamically instead of an actual `SecurityManager` implementation class name;
Replaced deprecated (since `Java9`) `Class.newInstance()` with `Class.getDeclaredConstructor().newInstance()` (not other occurrences, maybe another PR) ;
Minor `javadoc` updates.

Note: `JDK8` and `JDK11` don't support these special tokens `allow` and `disallow`.

Manually verified behaviors matching `RI`.
Expected JTReg tests to provide test coverage such as following.
https://github.com/ibmruntimes/openj9-openjdk-jdk12/blob/e969f57ff8031485b00ccf2ed10f87a627c7bd68/test/jdk/java/lang/System/AllowSecurityManager.java#L24-L32

Fixes: #4734 

Reviewer: @pshipton 
FYI: @DanHeidinga 

### DIFF
--- a/jcl/src/java.base/share/classes/com/ibm/oti/util/ExternalMessages-MasterIndex.properties
+++ b/jcl/src/java.base/share/classes/com/ibm/oti/util/ExternalMessages-MasterIndex.properties
@@ -1305,3 +1305,6 @@ K0900="Create a new Reference, since a Reference cannot be cloned."
 #java.lang.invoke.MethodHandle
 K0A00="Failed to resolve Constant Dynamic entry with j9class: {0}, name: {1}, descriptor: {2}, bsmData: {3}"
 K0A01="Constant_Dynamic references bootstrap method '{0}' does not have java.lang.invoke.MethodHandles.Lookup as first parameter."
+
+#java.lang.System
+K0B00="`-Djava.security.manager=disallow` has been specified."
--- a/jcl/src/java.base/share/classes/java/lang/ClassLoader.java
+++ b/jcl/src/java.base/share/classes/java/lang/ClassLoader.java
@@ -278,12 +278,18 @@ public abstract class ClassLoader {
 		/*[ENDIF]*/
 		jdk.internal.misc.VM.initLevel(2);
 		String javaSecurityManager = System.internalGetProperties().getProperty("java.security.manager"); //$NON-NLS-1$
-		if (null != javaSecurityManager) {
-			if (javaSecurityManager.isEmpty() || "default".equals(javaSecurityManager)) {
+		if ((javaSecurityManager != null) 
+		/*[IF Java12]*/
+			/* See the SecurityManager javadoc for details about special tokens. */
+			&& !javaSecurityManager.equals("disallow") //$NON-NLS-1$ /* special token to disallow SecurityManager */
+			&& !javaSecurityManager.equals("allow") //$NON-NLS-1$ /* special token to allow SecurityManager */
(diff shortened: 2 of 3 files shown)


jcl/src/java.base/share/classes/com/ibm/oti/util/ExternalMessages-MasterIndex.properties
```diff
@@ -1305,3 +1305,6 @@ K0900="Create a new Reference, since a Reference cannot be cloned."
 #java.lang.invoke.MethodHandle
 K0A00="Failed to resolve Constant Dynamic entry with j9class: {0}, name: {1}, descriptor: {2}, bsmData: {3}"
 K0A01="Constant_Dynamic references bootstrap method '{0}' does not have java.lang.invoke.MethodHandles.Lookup as first parameter."
+
+#java.lang.System
+K0B00="Runtime is specified to disallow setting security manager"
```
pshipton: This can be more explicit. i.e. `-Djava.security.manager=disallow` has been specified

jcl/src/java.base/share/classes/java/lang/ClassLoader.java
```diff
@@ -278,12 +278,17 @@ static final void initializeClassLoaders() {
 		/*[ENDIF]*/
 		jdk.internal.misc.VM.initLevel(2);
 		String javaSecurityManager = System.internalGetProperties().getProperty("java.security.manager"); //$NON-NLS-1$
-		if (null != javaSecurityManager) {
-			if (javaSecurityManager.isEmpty() || "default".equals(javaSecurityManager)) {
+		if ((javaSecurityManager != null) 
+		/*[IF Java12]*/
+			&& !javaSecurityManager.equals("disallow") //$NON-NLS-1$ /* special token to disallow SecurityManager */
```
pshipton: Please add a comment to refer to the SecurityManager javadoc.

jcl/src/java.base/share/classes/com/ibm/oti/util/ExternalMessages-MasterIndex.properties
JasonFengJ9: Updated the message.

jcl/src/java.base/share/classes/java/lang/ClassLoader.java
JasonFengJ9: `JDK 12` doc is still an [early access](https://download.java.net/java/early_access/jdk12/docs/api/java.base/java/lang/SecurityManager.html), added a comment `More details are at JDK 12 javadoc jdk12/docs/api/java.base/java/lang/SecurityManager.html` instead.

Please have another look.

pshipton: I didn't mean to include a direct link. In fact, I don't think the path is desired either. I'm fine with a comment at says "See the SecurityManager javadoc for details."

JasonFengJ9: Sure, the comment is updated.

pshipton:

PR #4740, opened 2019-02-14
Functions implemented for TR_ResolvedRelocatableJ9JITaaSServerMethod

satbir-ca:


satbir-ca:

ymanton:
Is this code supposed to execute on the server or the client? If it's supposed to be on the server I'm not sure how this implementation can work as it is, e.g. calling functions like `jitCTResolveStaticFieldRefWithMethod` on the server will not have the intended effect.

satbir-ca:
updated

PR #4744, opened 2019-02-14
Add SVM profiled class records for checkcast/instanceof

aviansie-ben:
The common infrastructure to get the top profiled classes for checkcast
and instanceof inlining has been updated to add the correct profiled
class records where necessary and to skip any profiled class for which
a record could not be added when run under AOT with SVM enabled.


dsouzai:
jenkins test sanity xlinux,plinux,zlinux jdk8,jdk11

Issue #3850, opened 2018-11-27
New methods in String for jdk12


pshipton:
indent(int)
transform​(Function<? super String,? extends R> f)
http://cr.openjdk.java.net/~iris/se/12/build/latest/api/java.base/java/lang/String.html

JasonFengJ9:
@theresa-m does your `Java 12` PR cover the methods specified here?

JasonFengJ9:
There are couple of new additions:
```
public Optional<String> describeConstable()
public String resolveConstantDesc​(MethodHandles.Lookup lookup)
```

theresa-m:
I have not covered the align/indent/transform methods. describeConstable and resolveConstantDesc are part of jep334. the pr for them is https://github.com/eclipse/openj9/pull/4105

JasonFengJ9:
It appears `align()` & `align(int)` have been removed.

theresa-m:
@sharon-wang

pshipton:
@JasonFengJ9 @theresa-m is somebody looking at transform?

fjeremic:
> @JasonFengJ9 @theresa-m is somebody looking at transform?

In addition I'd like to add to please take a look at the corresponding StringLatin1 and StringUTF16 classes to see if OpenJDK has already added helpers for these APIs which we can reuse. IIRC that was th ecase for some of the previous Java 12 methods.

JasonFengJ9:
> is somebody looking at transform?

Not that I am aware of.

pshipton:
@DanHeidinga we need an implementation for `String.transform​(Function<? super String,? extends R> f)`

@theresa-m @sharon-wang any outlook for indent(int)? The feature complete date is Feb 25.
There is no StringLatin1, StringUTF16 helper.

JasonFengJ9:
Will take `String.transform​(Function<? super String,? extends R> f)`.

theresa-m:
We're on it. @sharon-wang is working on some tests for the method.

PR #4745, opened 2019-02-14
JDK12 add new method java.lang.String.transform()

JasonFengJ9:
`JDK12` add new method `java.lang.String.transform()`

Adding `java.lang.String.transform()` for `JDK12`.

Manually verified it matches `RI` results in a simple testcase as following:
```
    String strOne = "StringOne";
    System.out.println((Object)strOne.transform(string -> string.length()));
```
Expecting the test coverage is provided by JTReg tests 
https://github.com/ibmruntimes/openj9-openjdk-jdk12/blob/e969f57ff8031485b00ccf2ed10f87a627c7bd68/test/jdk/java/lang/String/Transform.java#L24-L28

Related to: #3850 

Reviewer: @pshipton 
FYI: @DanHeidinga 

### DIFF
--- a/jcl/src/java.base/share/classes/java/lang/String.java
+++ b/jcl/src/java.base/share/classes/java/lang/String.java
@@ -37,6 +37,7 @@ import java.util.function.IntUnaryOperator;
 import java.util.Iterator;
 import java.nio.charset.Charset;
 /*[IF Java12]*/
+import java.util.function.Function;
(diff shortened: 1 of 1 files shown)


pshipton:
Please add your test to the test suite.

pshipton:
Given the implementation is so simple and there is coverage elsewhere, I'll merge without a test.

PR #4746, opened 2019-02-14
Replace calls to indexedTrampolineLookup with findHelperTrampoline

0xdaryl:
Ask the CodeCacheManager directly for the helper trampoline address.

Change the return type of `CodeCacheManager::findHelperTrampoline` to an `intptrj_t`
and fix up references.  Introduce a J9 version of `findHelperTrampoline` to
prevent breakages while upstream OMR is changed.

Change some invocations of `findHelperTrampoline` to use the new form where the
helper index is the first parameter.


0xdaryl:
Depends on eclipse/omr#3570

0xdaryl:
Jenkins test sanity all jdk11

0xdaryl:
Un-WIPing as upstream changes have made it to openj9-omr.

runtime/compiler/arm/runtime/Recomp.cpp
```diff
@@ -237,19 +238,16 @@ void J9::Recompilation::methodHasBeenRecompiled(void *oldStartPC, void *newStart
 
 uint32_t encodeRuntimeHelperBranchAndLink(TR_RuntimeHelper helper, uint8_t *cursor, TR_FrontEnd *fe)
    {
```
0xdaryl: This function needed a bit more love because of its inconsistent use of a single data type to represent the target address.

runtime/compiler/arm/codegen/ARMRecompilationSnippet.cpp
```diff
@@ -64,103 +64,3 @@ uint32_t TR::ARMRecompilationSnippet::getLength(int32_t estimatedSnippetStart)
    {
    return  12;
    }
-/*
```
0xdaryl: This commented out block of code has long been deprecated.  It contained a reference to the `indexedTrampolineLookup` so I simply deleted the whole thing.

runtime/compiler/runtime/J9CodeCacheManager.cpp
```diff
@@ -723,3 +723,9 @@ J9::CodeCacheManager::printOccupancyStats()
       }
    }
 
+
+intptrj_t
```
0xdaryl: Provide a version of `findHelperTrampoline` that returns an `intptrj_t` type rather than `void *`.  This eliminates a number of casts that would be immediately required when using this function.

0xdaryl:
@dsouzai @fjeremic @ymanton @mpirvu : I'd appreciate a review and merge from at least one of you please.

runtime/compiler/z/codegen/S390J9CallSnippet.cpp
```diff
@@ -213,14 +214,14 @@ TR::S390J9CallSnippet::generateInvokeExactJ2IThunk(TR::Node * callNode, int32_t
       cursor += 2;
 
       TR::SymbolReference *helper = cg->symRefTab()->findOrCreateRuntimeHelper(TR_methodHandleJ2IGlue, false, false, false);
-      uintptrj_t destAddr = (uintptrj_t)helper->getMethodAddress();
+      intptrj_t destAddr = (uintptrj_t)helper->getMethodAddress();
```
dsouzai: nitpick, the cast is `(uintptrj_t)` but the type of `destAddr` is `intptrj_t`.

dsouzai:
> If anyone can think of the reason why it's needed then its easy for me to add back in as a specialization in the J9 version of CodeCacheManager.

I thought the critical section was using a monitor, but it's a VMAccessCriticalSection. I really can't think of why the JIT would need VM Access for the code cache. If I were grasping for straws, I'd say the only situation it _might_ matter is under FSD (because of the jettisoning of the code cache)?

0xdaryl:
> re/ the `findHelperTrampoline` in J9::CodeCacheManager:

My intention is to change the return type of this function.  I introduced this version of the function temporarily in OpenJ9 to prevent build breakages because of the coordinated changes required in OMR.

Next steps:
* Once this PR is merged, in OMR replace uses of `indexedTrampolineLookup` with direct calls to `findHelperTrampoline` and add this `findHelperTrampoline` API to OMR
* Once that PR is merged, in OpenJ9 remove the FrontEnd implementation of `indexedTrampolineLookup` and this temporary function
* Once that PR is merged, remove deprecated FrontEnd API in OMR

runtime/compiler/z/codegen/S390J9CallSnippet.cpp
0xdaryl: Not a nit pick.  Thanks for pointing that out.  I'll fix and re-push.

0xdaryl:
Jenkins test sanity all jdk11

PR #4752, opened 2019-02-15
Remove volatile from return type

rwy0717:
The volatile return type used here has no effect.


rwy0717:
I'm closing this PR, as this usage of volatile may be a workaround for a bug in xlc.

PR #4755, opened 2019-02-15
Add J9HookVMDynamicCodeLoad for JITaaS remote AOT relocation

harryyu1994:
Issue: #4401
[skip ci]

### DIFF
--- a/runtime/compiler/control/CompilationRuntime.hpp
+++ b/runtime/compiler/control/CompilationRuntime.hpp
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2000, 2018 IBM Corp. and others
+ * Copyright (c) 2000, 2019 IBM Corp. and others
  *
  * This program and the accompanying materials are made available under
  * the terms of the Eclipse Public License 2.0 which accompanies this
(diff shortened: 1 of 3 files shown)


runtime/compiler/control/JITaaSCompilationThread.cpp
```diff
@@ -2562,7 +2564,63 @@ remoteCompilationEnd(
...
 
-         if (!relocatedMetaData)
+         if (relocatedMetaData) 
+            {
+            if (TR::Options::getVerboseOption(TR_VerboseJITaaS))
+               {
+               TR_VerboseLog::writeLineLocked(TR_Vlog_JITaaS, "JITaaS Client successfully relocated metadata for %s", comp->signature());
+               }
+
+            if (J9_EVENT_IS_HOOKED(jitConfig->javaVM->hookInterface, J9HOOK_VM_DYNAMIC_CODE_LOAD))
```
mpirvu: Since this part of the code is identical to the one use for non-JITaaS compilations can we refactor it into a separate function that is going to be called in two different places?

harryyu1994:
@mpirvu  Hi Marius, ready for a review along with https://github.com/eclipse/openj9/pull/4788 <-- another refactor item. 
Will remove "WIP" tag after testing finishes.

runtime/compiler/control/CompilationRuntime.hpp
```diff
@@ -966,6 +966,12 @@ class CompilationInfo
    uint32_t getLocalGCCounter() { return _localGCCounter; }
    void incrementLocalGCCounter() { _localGCCounter++; }
 
+#if defined(J9VM_INTERP_AOT_COMPILE_SUPPORT) && defined(J9VM_OPT_SHARED_CLASSES) && (defined(TR_HOST_X86) || defined(TR_HOST_POWER) || defined(TR_HOST_S390) || defined(TR_HOST_ARM))
```
mpirvu: Do we need the `ifdef` here? I am in favor of eliminating these ifdefs as much as possible. The only place where we need them is when we use SCC APIs that only exist under such `ifdef` in the VM code.

harryyu1994:
Tested and ready for merge.

PR #4756, opened 2019-02-15
Fix multiple problems with resolved methods

dmitry-ten:
- Implement version of `TR_ResolvedRelocatableJ9JITaaSServerMethod::createResolvedMethodFromJ9Method`
that is passed method info tuple as an argument. This version is used for most creations of resolved methods,
and we forgot to override it for AOT.

- When creating client-side mirror, need to differentiate 2 cases:
	1. Server called constructor of resolved method directly.
	2. Server called `createResolvedMethodFromJ9Method.`
Before this change we treated these cases as one.
It doesn't matter for non-AOT, because `createResolvedMethodFromJ9Method` just calls the constructor,
but it matters for AOT, because the above methods needs to do some checks, like verifying that classes
were loaded by the same class loader.

- Because of the previous point, creating a mirror can now fail.
Modify `packMethodInfo` to handle this case.

- Simplify constructor of `TR_ResolvedRelocatableJ9JITaaSServerMethod`.
We used to track on the server whether client remembered containing class of the
resolved method. This is not actually necessary, because all it affects is the
recognized method, and the correct version of that is already sent to the server.

- In TR_ResolvedRelocatableJ9Method constructor, check if owner method is not NULL,
before trying to validate class. Needed because different order of compilation object creation in JITaaS.

### DIFF
--- a/runtime/compiler/env/j9methodServer.cpp
+++ b/runtime/compiler/env/j9methodServer.cpp
@@ -1232,10 +1232,86 @@ TR_ResolvedJ9JITaaSServerMethod::getCallerWeight(TR_ResolvedJ9Method *caller, ui
    }
 
 void
-TR_ResolvedJ9JITaaSServerMethod::createResolvedJ9MethodMirror(TR_ResolvedJ9JITaaSServerMethodInfo &methodInfo, TR_OpaqueMethodBlock *method, uint32_t vTableSlot, TR_ResolvedMethod *owningMethod, TR_FrontEnd *fe, TR_Memory *trMemory)
+TR_ResolvedJ9JITaaSServerMethod::createResolvedMethodMirror(TR_ResolvedJ9JITaaSServerMethodInfo &methodInfo, TR_OpaqueMethodBlock *method, uint32_t vTableSlot, TR_ResolvedMethod *owningMethod, TR_FrontEnd *fe, TR_Memory *trMemory)
    {
-   TR_ResolvedJ9Method *resolvedMethod = new (trMemory->trHeapMemory()) TR_ResolvedJ9Method(method, fe, trMemory, owningMethod, vTableSlot);
+   // Create resolved method mirror on the client.
+   // Should be called to mirror a call to resolved method constructor,
+   // will work when method doesn't have an owning method (compilee method).
+   // Resolved method should not be NULL
+   TR_ResolvedJ9Method *resolvedMethod = NULL;
+   if (!((TR_J9VMBase *) fe)->isAOT_DEPRECATED_DO_NOT_USE())
+      resolvedMethod = new (trMemory->trHeapMemory()) TR_ResolvedJ9Method(method, fe, trMemory, owningMethod, vTableSlot);
+   else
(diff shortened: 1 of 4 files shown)


dmitry-ten:
@mpirvu could you do a review?

ymanton:
I'll look at it soon, but in the meantime @satbir-ca FYI this probably overlaps with what you are doing in #4740.

runtime/compiler/env/j9methodServer.cpp
```diff
@@ -1232,17 +1232,102 @@ TR_ResolvedJ9JITaaSServerMethod::getCallerWeight(TR_ResolvedJ9Method *caller, ui
...
 
    packMethodInfo(methodInfo, resolvedMethod, fe);
    }
 
+void
+TR_ResolvedJ9JITaaSServerMethod::createResolvedMethodFromJ9MethodMirror(TR_ResolvedJ9JITaaSServerMethodInfo &methodInfo, TR_OpaqueMethodBlock *method, uint32_t vTableSlot, TR_ResolvedMethod *owningMethod, TR_FrontEnd *fe, TR_Memory *trMemory)
+   {
+   // Create resolved method mirror on the client.
+   // Should be called to mirror a call to TR_Resolved(Relocatable)MethodFromJ9Method::createResolvedMethodFromJ9Method.
+   // Requires owning method to exist to work properly.
```
mpirvu: Let's put an assert that `owningMethod` is not NULL

runtime/compiler/env/j9methodServer.cpp
```diff
@@ -1232,17 +1232,102 @@ TR_ResolvedJ9JITaaSServerMethod::getCallerWeight(TR_ResolvedJ9Method *caller, ui
...
 
 void
-TR_ResolvedJ9JITaaSServerMethod::createResolvedJ9MethodMirror(TR_ResolvedJ9JITaaSServerMethodInfo &methodInfo, TR_OpaqueMethodBlock *method, uint32_t vTableSlot, TR_ResolvedMethod *owningMethod, TR_FrontEnd *fe, TR_Memory *trMemory)
-   {
-   TR_ResolvedJ9Method *resolvedMethod = new (trMemory->trHeapMemory()) TR_ResolvedJ9Method(method, fe, trMemory, owningMethod, vTableSlot);
+TR_ResolvedJ9JITaaSServerMethod::createResolvedMethodMirror(TR_ResolvedJ9JITaaSServerMethodInfo &methodInfo, TR_OpaqueMethodBlock *method, uint32_t vTableSlot, TR_ResolvedMethod *owningMethod, TR_FrontEnd *fe, TR_Memory *trMemory)
+   {
+   // Create resolved method mirror on the client.
+   // Should be called to mirror a call to resolved method constructor,
+   // will work when method doesn't have an owning method (compilee method).
```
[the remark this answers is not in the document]
dmitry-ten: No, it means that owningMethod can be NULL, but it can also be not NULL.

runtime/compiler/env/j9methodServer.cpp
dmitry-ten: ok

dmitry-ten:
Done.

PR #4760, opened 2019-02-15
Remove initializeAotRuntimeInfo() call from TR_JITaaSRelocationRuntime

harryyu1994:
This API is actually used by JITaaS client and we should not put an assert here
`initializeAotRuntimeInfo` is used by `prepareRelocateJITCodeAndData`.

Issue: #4401
[skip ci]

### DIFF
--- a/runtime/compiler/runtime/RelocationRuntime.cpp
+++ b/runtime/compiler/runtime/RelocationRuntime.cpp
@@ -543,7 +543,6 @@ TR_RelocationRuntime::prepareRelocateJITCodeAndData(J9VMThread* vmThread,
 
    if (_relocationStatus == RelocationNoError)
       {
-      initializeAotRuntimeInfo();
       AOTcgDiag5(comp, "%s: relocating code %p -> %p (start pc %p -> %p)\n",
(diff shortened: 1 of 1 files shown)


harryyu1994:
@ymanton Hi Younes, so I was wrong in that the class TR_JITaaSRelocationRuntime is only used by the server. We also use this relocationRuntime on the client for non-AOT. One of the uses on client I see is `prepareRelocateJITCodeAndData`, and it calls `initializeAotRuntimeInfo` (which gets to the base class TR_RelocationRuntime's empty initializeAotRuntimeInfo API ). Since it's not AOT, we intentionally want  `initializeAotRuntimeInfo` to do nothing. However, we can also just remove the call in `prepareRelocateJITCodeAndData` and still leave this assert in. Which way is better?

ymanton:
> However, we can also just remove the call in `prepareRelocateJITCodeAndData` and still leave this assert in.

That's probably the better way. The base class should probably be refactored at some point such that it doesn't have any AOT-specific APIs, they should all be in the AOT version of the RR. You may as well remove the call from `prepareRelocateJITCodeAndData` now.

harryyu1994:
@ymanton change updated and ready for review & merge.

PR #4762, opened 2019-02-15
Add further SVM documentation

dsouzai:
[ci-skip]

### DIFF
--- a/doc/compiler/aot/SymbolValidationManager.md
+++ b/doc/compiler/aot/SymbolValidationManager.md
@@ -212,3 +212,33 @@ in the same way, to match the IDs assigned during compilation.
 3. Allows relocations of classes/methods associated with cpIndex=-1
 4. Facilitates the enablement of Known Object Table under AOT
 5. Facilitates the enablement of JSR292 under AOT
+
+## When to create a new Validation Record
+
+Any time a new front end query is created, or an existing query is 
+modified, the appropriate validation record should be created, or 
+modified, respectively<sup>1</sup>. "Front end query" means, a query that the 
+compiler (the "back end") makes of the runtime (the "front end") in order
+to get information about the environment such that execution of the
+compiled method will be functionally correct. The SVM uses the 
+validation record to redo the query in order to perform the validation. 
+Therefore, the validation record needs to contain all the information
+necessary to redo the query in a different JVM instance. Each unique
+front end query will have its own associated validation record; this is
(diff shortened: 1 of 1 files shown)


dsouzai:
Force pushed to change `[ci-skip]` to `[ci skip]`...

dsouzai:
@jdmpapin This look ok to you?

andrewcraik:
Perhaps add a further qualification saying something like "Note the validation is required where the query result is used in a way that could affect **functional correctness** - queries used only for heuristic purposes do not necessarily _need_ to be validated since, by nature, they do not affect the correctness of the program. All omissions on the basis of heuristic usage should be documented as such in the code to make the omission clearly intentional and the basis for that omission clear."

dsouzai:
@andrewcraik updated with the proposed changes.

dsouzai:
@andrewcraik  @jdmpapin made the requested changes.

doc/compiler/aot/SymbolValidationManager.md
```diff
@@ -212,3 +212,33 @@ in the same way, to match the IDs assigned during compilation.
...
+by nature, they do not affect the correctness of the program. 
+All omissions on the basis of heuristic usage should be documented as 
+such in the code to make the omission clearly intentional and the basis 
+for that omission clear. The `enterHeuristicRegion`/`exitHeuristicRegion`
+APIs are used to facilitate making frontend queries without generating
+validation records.
+
+<hr/>
+1. We can have "non-primitive" queries (i.e. queries that answer questions
+about anything that isn't a primitive) that simply combine other
```
jdmpapin: This clarification ("i.e. [...]") makes it sound like it matters whether we're querying properties of primitive types. Maybe it was confusing for me to use the word "primitive" earlier. The distinction I was trying to make was between queries that have to be part of the "front end" and those that could be implemented outside of it. I think the real distinction is whether or not the query inspects the run-time environment (classes, etc.) _itself_. A "non-primitive" query would be one that inspects the run-time environment only through other queries

andrewcraik: Might I suggest a better word might be compound / non-compound. So it would read "We can have compound queries, queries that aggregate responses from other queries, …"

dsouzai:
Changes made.

dsouzai:
@andrewcraik could you merge if everything looks good to you?

PR #4105, opened 2018-12-18
Methods and tests for jep334 String

theresa-m:
- Java12andUp test framework
- String.describeConstable & tests
- String.resolveConstantDesc & tests

Depends on (see https://github.com/eclipse/openj9/issues/4195): 
- Fix Java 12 build failures: https://github.com/eclipse/openj9/pull/3980
- Constants API stubs: https://github.com/eclipse/openj9/pull/4030
- Java12andUp test setup: https://github.com/eclipse/openj9/pull/4104

Signed-off-by: Theresa Mammarella <Theresa.T.Mammarella@ibm.com>

PR #4767, opened 2019-02-16
Implement new JDK12 method java.lang.String.indent(int)

sharon-wang:
Fixes: #3850.

Dependent on: #4105 test framework commit - sanity test for indent(int) will be moved to Java12andUp once it is available.

### DIFF
--- a/jcl/src/java.base/share/classes/java/lang/String.java
+++ b/jcl/src/java.base/share/classes/java/lang/String.java
@@ -8299,10 +8299,15 @@ written authorization of the copyright holder.
 
 /*[IF Java12]*/
 	/**
-	 * Apply a function to this string. The function expects a single String input and returns an R.
+	 * Apply a function to this string. The function expects a single String input
+	 * and returns an R.
 	 * 
-	 * @param f - the functional interface to be applied
+	 * @param f 
+	 *			  the functional interface to be applied
+	 *
 	 * @return the result of application of the function to this string
+	 * 
+	 * @since 12
 	 */
 	public <R> R transform(Function<? super String, ? extends R> f) {
 		return f.apply(this);
@@ -8313,20 +8318,75 @@ written authorization of the copyright holder.
 	 * if construction is not possible.
 	 * 
 	 * @return Optional with nominal descriptor of String instance
+	 * 
+	 * @since 12
 	 */
 	public Optional<String> describeConstable() {
 		return Optional.of(this);
 	}
 
 	/**
-	 * Resolves this ConstantDesc instance
+	 * Resolves this ConstantDesc instance.
 	 * 
-	 * @param lookup parameter is ignored
+	 * @param lookup 
+	 *			  parameter is ignored
 	 * 
 	 * @return the resolved Constable value
+	 * 
+	 * @since 12
 	 */
 	public String resolveConstantDesc(MethodHandles.Lookup lookup) {
 		return this;
 	}
+
+	/**
+	 * Indents each line of the string depending on the value of n, and normalizes
+	 * line terminators to the newline character "\n".
+	 * 
+	 * @param n 
+	 *			  the number of spaces to indent the string
+	 *
+	 * @return the indented string with normalized line terminators
+	 * 
+	 * @since 12
+	 */
+	public String indent(int n) {
+		Stream<String> lines = lines();
+		Iterator<String> iter = lines.iterator();
+		StringBuilder builder = new StringBuilder();
+
+		String spaces = n > 0 ? " ".repeat(n) : null;
+		int absN = Math.abs(n);
+
+		while (iter.hasNext()) {
+			String currentLine = iter.next();
+
+			if (n > 0) {
+				builder.append(spaces);
+			} else if (n < 0) {
+				int start = 0;
+
+				while ((currentLine.length() > start) 
+					&& (Character.isWhitespace(currentLine.charAt(start)))
+				) {
+					start++;
+
+					if (start >= absN) {
+						break;
+					}
+				}
+				currentLine = currentLine.substring(start);
+			}
+
+			/**
+			 * Line terminators are removed when lines() is called. A newline character is
+			 * added to the end of each line, to normalize line terminators.
+			 */
+			builder.append(currentLine);
+			builder.append("\n");
+		}
+
+		return builder.toString();
+	}	
 /*[ENDIF] Java12 */
 }
--- a/test/functional/Java11andUp/src/org/openj9/test/java_lang/Test_String.java
+++ b/test/functional/Java11andUp/src/org/openj9/test/java_lang/Test_String.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2018, 2018 IBM Corp. and others
+ * Copyright (c) 2018, 2019 IBM Corp. and others
  *
  * This program and the accompanying materials are made available under
  * the terms of the Eclipse Public License 2.0 which accompanies this
--- a/test/functional/Java12andUp/playlist.xml
+++ b/test/functional/Java12andUp/playlist.xml
@@ -45,4 +45,27 @@
                   <subset>12+</subset>
           </subsets>
   </test>
+
+  <test>
+          <testCaseName>StringIndentTests</testCaseName>
+          <variations>
+                  <variation>NoOptions</variation>
+          </variations>
+          <command>$(JAVA_COMMAND) $(JVM_OPTIONS) \
+                  -cp $(Q)$(RESOURCES_DIR)$(P)$(TESTNG)$(P)$(TEST_RESROOT)$(D)GeneralTest.jar$(Q) \
+                  org.testng.TestNG -d $(REPORTDIR) $(Q)$(TEST_RESROOT)$(D)testng.xml$(Q) -testnames StringIndentTests \
+                  -groups $(TEST_GROUP) \
+                  -excludegroups $(DEFAULT_EXCLUDE); \
+                  $(TEST_STATUS)
+          </command>
+          <levels>
+                  <level>sanity</level>
+          </levels>
+          <groups>
+                  <group>functional</group>
+          </groups>
+          <subsets>
+                  <subset>12+</subset>
+          </subsets>
+  </test>
 </playlist>
--- a/test/functional/Java12andUp/src/org/openj9/test/java_lang/Test_String.java
+++ b/test/functional/Java12andUp/src/org/openj9/test/java_lang/Test_String.java
@@ -82,5 +82,4 @@ public class Test_String {
 		String resolvedTest2 = test.resolveConstantDesc(null);
 		Assert.assertTrue(test.equals(resolvedTest2));
 	}
-	 
 }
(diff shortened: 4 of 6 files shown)


jcl/src/java.base/share/classes/java/lang/String.java
```diff
@@ -8315,5 +8315,48 @@ public String resolveConstantDesc(MethodHandles.Lookup lookup) {
...
+	/**
+	 * Adjusts the indentation of each line of this string based on the value of n, 
+	 * and normalizes line termination characters. 
+	 * 
+	 * @param n
+	 *			  the number of spaces to indent the string	 
+	 *
+	 * @return the string with indentation adjusted and line endings normalized
+	 * 
+	 * @since 12
```
pshipton: Can you please add this to the transform() method above, it was missed.

theresa-m:
you can change the pr description to include "Fixes: #3850" since this is the last change needed to finish the issue.

jcl/src/java.base/share/classes/java/lang/String.java
```diff
@@ -8315,5 +8315,48 @@ public String resolveConstantDesc(MethodHandles.Lookup lookup) {
...
+                        start++;
+                    }
+                    else {
+                        break;
+                    }
+               }
+               currentLine = currentLine.substring(start);
+           }
+
+           builder.append(currentLine + "\n");
```
theresa-m: It looks like this would just add an extra newline character to the line. not sure thats correct but it should be fairly obvious when you get a change to test against the RI (RI = reference implementation = hotspot)

jcl/src/java.base/share/classes/java/lang/String.java
```diff
@@ -8315,5 +8315,48 @@ public String resolveConstantDesc(MethodHandles.Lookup lookup) {
...
+        while (iter.hasNext()) {
+           String currentLine = iter.next();
+
+           if (n > 0) {
+               for (int i = 0; i < n; i++) {
+                   currentLine = " " + currentLine;
+               }
+           }
+           else {
+               int start = 0;
```
theresa-m: you can use `String.isBlank()` to see if your string has any non whitespace characters first. If all characters are whitespace you know that `start = n` without having to go through the loop.

test/functional/Java11andUp/src/org/openj9/test/java_lang/Test_String.java
```diff
@@ -268,4 +268,59 @@ public void testLines() {
 			// Expected exception
 		}
 	}
+	
+	/*	
+	 * Test Java 12 API String.indent
+	 */
+	// TODO: move to Java12andUp
+	@Test(groups = { "level.sanity" })
+	public void testIndent() {
```
theresa-m: There's a lot of duplicated code in your tests. I would make a helper function to improve maintainability.

test/functional/Java11andUp/src/org/openj9/test/java_lang/Test_String.java
```diff
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2018, 2018 IBM Corp. and others
+ * Copyright (c) 2018, 2019 IBM Corp. and others
```
theresa-m: some other tests you could consider are
- what happens when `|n|` is greater or less than the number of characters in your string?
- I think it would be good to have tests for different kinds of strings like the empty string and string without a line ending etc

test/functional/Java11andUp/src/org/openj9/test/java_lang/Test_String.java
```diff
@@ -268,4 +268,59 @@ public void testLines() {
...
+		while (iter.hasNext()) {
+			s = iter.next();
+			Assert.assertTrue(s.equals(latin1LinesArray[i]) || s.equals(latin1LinesArray[i].substring(-n)),
+					"indent: failed to indent " + n + " spaces - failed on " + latin1LinesArray[i] + " with: " + s);
+			i++;
+		}
+
+		// allWhiteSpace
+		s = allWhiteSpace.indent(n);
+		Assert.assertEquals(s, allWhiteSpace.substring(-n) + "\n",
```
theresa-m: same comment about `-n` as before

jcl/src/java.base/share/classes/java/lang/String.java
```diff
@@ -8307,13 +8307,67 @@ public String resolveConstantDesc(MethodHandles.Lookup lookup) {
...
+				int absN = Math.abs(n);
+				if (currentLine.isBlank()) {
+					if (currentLine.length() > absN) {
+						currentLine = currentLine.substring(absN);
+					} else {
+						currentLine = "";
+					}
+				} else {
+					int start = 0;
+					while ((currentLine.length() > start) && (Character.isWhitespace(currentLine.charAt(start)))) {
```
theresa-m: nitpick to break this into two separate lines as its running off of view. we make it look like this where the && or || operator starts the next line
```
while ((currentLine.length() > start) 
     && (Character.isWhitespace(currentLine.charAt(start)))
) {
```

jcl/src/java.base/share/classes/java/lang/String.java
```diff
@@ -8307,13 +8307,67 @@ public String resolveConstantDesc(MethodHandles.Lookup lookup) {
...
+						if (start >= absN) {
+							break;
+						}
+					}
+					currentLine = currentLine.substring(start);
+				}
+			}
+
+			// TODO: check if this adds an extra newline or if it is correct
+			builder.append(currentLine + "\n");
```
theresa-m: can you add a comment to explain what's happening here?

jcl/src/java.base/share/classes/java/lang/String.java
```diff
@@ -8307,13 +8307,71 @@ public String resolveConstantDesc(MethodHandles.Lookup lookup) {
...
+						}
+					}
+					currentLine = currentLine.substring(start);
+				}
+			}
+			/**
+			 * Newline characters are removed when lines() is called. The newline character
+			 * "\n" is added back for each line, to normalize line terminators.
+			 */
+			builder.append(currentLine + "\n");
```
DanHeidinga: What's the expected behaviour on windows?  Appending `\n`?  or `\r\n`?  I expect using `System.lineSeparator()` will do the right thing here

jcl/src/java.base/share/classes/java/lang/String.java
```diff
@@ -8307,13 +8307,71 @@ public String resolveConstantDesc(MethodHandles.Lookup lookup) {
...
+		Stream<String> lines = lines();
+		Iterator<String> iter = lines.iterator();
+		StringBuilder builder = new StringBuilder();
+
+		while (iter.hasNext()) {
+			String currentLine = iter.next();
+
+			if (n > 0) {
+				int spaceOffset = n + currentLine.length();
+				currentLine = String.format("%" + spaceOffset + "s", currentLine);
```
DanHeidinga: Does this match the RI's behaviour?  Some jshell examples show that the RI is prepending a fixed amount of space:
```
jshell> "ab".indent(1)
$28 ==> " ab\n"

jshell> " ab".indent(1)
$29 ==> "  ab\n"
```
Note, the second example starts with a space and indent(1) still adds a space.

If the method should always add some number of spaces, the code create the string of spaces using String.repeat() and prepend that string always.

test/functional/Java11andUp/src/org/openj9/test/java_lang/Test_String.java
```diff
@@ -268,4 +268,105 @@ public void testLines() {
 			// Expected exception
 		}
 	}
+	
+	/*
+	 * Test Java 12 API String.indent
+	 */
+	// TODO: move to Java12andUp
```
DanHeidinga: Please address the TODO

test/functional/Java11andUp/src/org/openj9/test/java_lang/Test_String.java
```diff
@@ -268,4 +268,105 @@ public void testLines() {
...
+	@Test(groups = { "level.sanity" })
+	public void testIndent() {
```
DanHeidinga: Do these tests pass on the RI?

jcl/src/java.base/share/classes/java/lang/String.java
```diff
@@ -8307,13 +8307,71 @@ public String resolveConstantDesc(MethodHandles.Lookup lookup) {
...
+	public String indent(int n) {
+		Stream<String> lines = lines();
+		Iterator<String> iter = lines.iterator();
+		StringBuilder builder = new StringBuilder();
+
+		while (iter.hasNext()) {
+			String currentLine = iter.next();
+
+			if (n > 0) {
+				String spaces = " ".repeat(n);
```
DanHeidinga: This is loop invariant for `n > 0` case and can be calculated outside the loop

jcl/src/java.base/share/classes/java/lang/String.java
```diff
@@ -8307,13 +8307,71 @@ public String resolveConstantDesc(MethodHandles.Lookup lookup) {
...
+				currentLine = spaces + currentLine;
+			} else if (n < 0) {
+				int absN = Math.abs(n);
```
DanHeidinga: This is also loop invariant

jcl/src/java.base/share/classes/java/lang/String.java
```diff
@@ -8307,13 +8307,71 @@ public String resolveConstantDesc(MethodHandles.Lookup lookup) {
...
+				if (currentLine.isBlank()) {
```
DanHeidinga: Does this need to know if the line is blank?  checking for a blank line is equivalent to the `Character.isWhitespace(currentLine.charAt(start))` loop below.  I think the space case can be removed and handled by the substring() call

test/functional/Java11andUp/src/org/openj9/test/java_lang/Test_String.java
sharon-wang: Will move to Java12andUp once the #4105 test framework commit goes through

jcl/src/java.base/share/classes/java/lang/String.java
sharon-wang: Looks like the jshell output for RI on Windows doesn't use `System.lineSeparator()` and just uses `\n`. Will be changing it back to `\n`.

test/functional/Java11andUp/src/org/openj9/test/java_lang/Test_String.java
DanHeidinga: @sharon-wang #4105 has been merged.  Can you rebase this and address the TODO?

sharon-wang:
@DanHeidinga should be good to go!

DanHeidinga:

DanHeidinga:
only build failure is the known `com.ibm.j9.jsr292.api.MethodHandleAPI_defineClass.test_defineClass_ClassBytes_CorruptedArray` issue

DanHeidinga:
Congrats @sharon-wang on your first OpenJ9 PR!

PR #4779, opened 2019-02-19
Fix code cache issues with remote AOT

dmitry-ten:
- Use `prepareRelocateJITCodeAndData` instead of `prepareRelocateAOTCodeAndData`,
even in remote AOT mode. This is because both methods do the same thing, except
the former is passed a code cache string received from the server, while the latter
assumes that code cache is contiguously allocated right after data cache.
For JITaaS, that assumption is incorrect and we always need to use code cache from the server.

- Tell `prepareRelocateJITCodeAndData` to allocate code cache by passing
`shouldUseCompiledCopy=false`. In a local compilation, code cache is allocated during codegen,
so it can be later accessed when relocating code. In JITaaS, however, we allocate code cache on the
server and need to allocate it on the client right before relocating.

### DIFF
--- a/runtime/compiler/control/JITaaSCompilationThread.cpp
+++ b/runtime/compiler/control/JITaaSCompilationThread.cpp
@@ -2386,9 +2386,17 @@ remoteCompilationEnd(
       {
       compInfoPT->reloRuntime()->setReloStartTime(compInfoPT->getTimeWhenCompStarted());
 
-      relocatedMetaData = compInfoPT->reloRuntime()->prepareRelocateJITCodeAndData(vmThread, fe,
-            comp->cg()->getCodeCache(), (uint8_t *)&codeCacheStr[0], (J9JITDataCacheHeader *)&dataCacheStr[0],
-            method, false, TR::comp()->getOptions(), TR::comp(), compilee);
+      relocatedMetaData = compInfoPT->reloRuntime()->prepareRelocateAOTCodeAndData(
+         vmThread,
+         fe,
+         comp->cg()->getCodeCache(),
+         (J9JITDataCacheHeader *)&dataCacheStr[0],
+         method,
+         false,
+         comp->getOptions(),
+         comp,
+         compilee,
+         (uint8_t *)&codeCacheStr[0]);
 
       if (!relocatedMetaData)
          {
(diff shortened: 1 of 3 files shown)


ymanton:
FYI #4760 just removed the `initializeAotRuntimeInfo()` call from `prepareRelocateJITCodeAndData()` because it wasn't needed. I don't think it affects you, but just in case you have some unpushed changes that depend on it being called you'll need to re-work them.

dmitry-ten:
`prepareRelocateJITCodeAndData` does not affect which relocations get done, both call `relocateAOTCodeAndData` which applies relocations. 

I think we do need `initializeAOTRuntimeInfo` for remote AOT. I'll check if we need to add it.

dmitry-ten:
So we do need to call `initializeAOTRuntimeInfo`.
I could add it back to `prepareRelocateJITCodeAndData` and execute it only for AOT.
Or I could modify `prepareRelocateAOTCodeAndData` and pass it code from the server as an argument.
Which one do you think is better?

harryyu1994:
@ymanton Hi Younes, your takes on this? Dmitry switched to using `prepareRelocateJITCodeAndData` for remote AOT as well and now `initializeAOTRuntimeInfo` is no longer an empty function and needed.

ymanton:
OK, I guess you should put it back and provide an implementation in `TR_JITaaSRelocationRuntime` that does nothing for remote JIT and does whatever `TR_SharedCacheRelocationRuntime` currently does for remote AOT.

dmitry-ten:
Re-added `initializeAotRuntimeInfo` and removed assert.

mpirvu:
As far as I can tell, `prepareRelocateJITCodeAndData` and `prepareRelocateAOTCodeAndData` are 99% the same. I would prefer to have only one function with a small `if` statement where needed.

dmitry-ten:
@mpirvu Done, also changed commit message.

mpirvu:
@dmitry-ten Could you please reformat the commit text to have shorter lines? It doesn't look so nice when seen in the web browser. I believe that the recommended line length is 72 (were appropriate). Thanks

dmitry-ten:
Done.

PR #4795, opened 2019-02-19
Remove Java 9/10 native supports

JasonFengJ9:
Remove `Java 9/10` native supports

Replaced `J2SE_19` with `J2SE_V11`;
Removed redundant `J2SE_VERSION(vm) &` with `J2SE_RELEASE_MASK` or `J2SE_VERSION_MASK`.

Reviewer: @DanHeidinga 

### DIFF
--- a/runtime/jcl/common/vm_scar.c
+++ b/runtime/jcl/common/vm_scar.c
@@ -278,7 +278,6 @@ scarInit(J9JavaVM * vm)
 #if defined(WIN32) && !defined(OPENJ9_BUILD)
 	switch (jclVersion) {
 	case J2SE_18:
-	case J2SE_19:	/* This will need to be modified when the JCL team provides a dbgwrapper90 */
 		dbgwrapperStr = "dbgwrapper80";
 		break;
 	default:
@@ -448,7 +447,7 @@ scarPreconfigure(J9JavaVM * vm)
 		return JNI_ERR;
 	}
 
-	if (J2SE_VERSION(vm) < J2SE_19) {
+	if (J2SE_VERSION(vm) < J2SE_V11) {
 		char *vmSpecificDirectoryPath = "jclSC180";
 
 		rc = addVMSpecificDirectories(vm, &i, vmSpecificDirectoryPath);
--- a/runtime/vm/jnimisc.cpp
+++ b/runtime/vm/jnimisc.cpp
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2012, 2018 IBM Corp. and others
+ * Copyright (c) 2012, 2019 IBM Corp. and others
  *
  * This program and the accompanying materials are made available under
  * the terms of the Eclipse Public License 2.0 which accompanies this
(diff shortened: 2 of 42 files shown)


runtime/jcl/common/vm_scar.c
```diff
@@ -278,7 +278,7 @@ scarInit(J9JavaVM * vm)
 #if defined(WIN32) && !defined(OPENJ9_BUILD)
 	switch (jclVersion) {
 	case J2SE_18:
-	case J2SE_19:	/* This will need to be modified when the JCL team provides a dbgwrapper90 */
+	case J2SE_V11:	/* This will need to be modified when the JCL team provides a dbgwrapper90 */
```
pshipton: Java 11 doesn't support dbgwrapper and probably never will. It was only used for an IBM internal Java 9 that was obsoleted by OpenJ9.

runtime/vm/jnimisc.cpp
```diff
@@ -394,13 +394,8 @@ getObjectClass(JNIEnv *env, jobject obj)
 jint JNICALL
 getVersion(JNIEnv *env)
 {
-	J9VMThread *currentThread = (J9VMThread*)env;
-	J9JavaVM *vm = currentThread->javaVM;
-
-	if (J2SE_VERSION(vm) >= J2SE_V10) {
+	if (J2SE_VERSION(((J9VMThread*)env)->javaVM) >= J2SE_V11) {
 		return JNI_VERSION_10;
```
pshipton: This needs to be checked, should it be 11?

pshipton:
I think we should change all the runtime checks to #ifdef's. The build is already created specific to a Java version, I don't think there is any advantage in using runtime checks.

runtime/jcl/common/vm_scar.c
JasonFengJ9: Will remove `J2SE_V11` as well then.

runtime/vm/jnimisc.cpp
JasonFengJ9: `JDK11` still has `JNI_VERSION_10`. By the way, I didn't update JVMTI version for `JDK11` and leave it to JEP331 PR instead.

JasonFengJ9:
> I think we should change all the runtime checks to #ifdef's. The build is already created specific to a Java version, I don't think there is any advantage in using runtime checks.

Agreed, it does make sense to replace the runtime check with `#ifdef`.

DanHeidinga:
The only caution with #ifdefs is I want to ensure that we don't end up maintaining N-variants of the code ifdef'd differently.

The costs of runtime checks helps force us to use them sparingly and therefore make as little code version specific as possible.

runtime/jcl/common/vm_scar.c
JasonFengJ9: PR has been updated, please have another look.

pshipton:
jenkins tests sanity win jdk8,jdk11

pshipton:
jenkins test sanity win jdk8,jdk11

PR #4797, opened 2019-02-20
Temporarily lower thread priority when spinning

DanHeidinga:
Make the `spinOnTryEnter` match the `spinOnFlatLock` case which
already had calls to `dropSMTThreadPriority` and `restoreSMTThreadPriority`
around the tight spin loop.


DanHeidinga:
@babsingh Can you review this change?

DanHeidinga:
I've searched the history, going back to the Builder code, and don't know why we only applied this to one of the two spins. 

There's another copy of similar code in OMR that should likely be updated in the same way:
https://github.com/eclipse/omr/blob/fb3c9e04d4f32578ad8c45ceb6ca470c3087d5ca/thread/common/threadhelpers.cpp#L103-L106

DanHeidinga:
fyi @zl-wang

DanHeidinga:
When the original change was made to the `spinOnTryEnter` case, the spin count for the lock was also dropped from 256 to 96 when running on PPC.

I've reached out to @ymanton to see if he knows why we only modified one case originally.

zl-wang:
@DanHeidinga  @babsingh  @ymanton 
While we are at this issue, the so-called AtomicSupport drop/restore priority primitives need to be different for AIX and pLinux, because they choose different default hardware priority.   AIX default priority is Medium (i.e. 4), while pLinux default is Medium-Low (i.e. 3).

The current implementation is suitable to AIX (dropping to Low -- i.e. 2, restoring to Medium).  It is not suitable for pLinux (restoring to Medium which is higher than the platform default).  The suggestion for  pLinux is:  dropping to Very-Low -- i.e. 1;  restoring to Medium-Low which is the platform default).

The instruction to drop to Very-Low is:              or r31, r31, r31;
The instruction to restore to Medium-Low is:    or r6, r6, r6;

zl-wang:
Likely need some experiments to land on the appropriate (default) spinCount(s).  We can start with the previous count @ymanton identified.  Same ratio to begin with (Medium minus Low is 2;  while Medium-Low minus Very-Low is 2 as well.   That means CPU core dispatches instructions at a rate 4 times less than the default priority thread[s]).

DanHeidinga:
@zl-wang Do you have the encodings of the instructions on ppc for xlc?  I'm looking at updating the OMR code and it would big help if you could provide the equivalent of https://github.com/eclipse/omr/blob/fb3c9e04d4f32578ad8c45ceb6ca470c3087d5ca/include_core/AtomicSupport.hpp#L52-L60

zl-wang:
@DanHeidinga   binary encoding for
or r31,r31,r31 is 0x7FFFFB78
or r6,r6,r6       is 0x7CC63378

DanHeidinga:
Jenkins test sanity plinux,aix jdk8

genie-openj9:

PR #4805, opened 2019-02-20
Remove TR::X86UnresolvedVirtualCallSnippet print

jdmpapin:
`TR::X86UnresolvedVirtualCallSnippet` itself has already been removed, and the definition of `print` is only here to prevent linker errors. Now that the only (dead) call site has been removed from OMR, this definition is unreferenced and can be deleted.

### DIFF
--- a/runtime/compiler/x/codegen/CallSnippet.cpp
+++ b/runtime/compiler/x/codegen/CallSnippet.cpp
@@ -39,10 +39,6 @@
 #include "il/symbol/StaticSymbol.hpp"
 #include "x/codegen/X86PrivateLinkage.hpp"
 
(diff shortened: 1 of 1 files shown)


0xdaryl:
Jenkins line endings check

0xdaryl:
Jenkins copyright check

0xdaryl:

This is an x86-specific change.  No need to test non-x86 platforms.

0xdaryl:

jdmpapin:
The build failed with a compiler error in an unrelated part of the code

0xdaryl:

0xdaryl:
Failures are unrelated to this simple change.  Merging.

PR #4807, opened 2019-02-20
Fix version conditions for cmake 3.4

keithc-ca:
Replace uses of `GREATER_EQUAL` (which requires cmake version 3.7 or better) with `NOT LESS` (which is available in 3.4), the minimum required version.


keithc-ca:
Jenkins copyright check

keithc-ca:
Jenkins lineendings check

keithc-ca:
Jenkins line endings check

Issue #4776, opened 2019-02-19
OpenJDK8-OpenJ9 throws UnsatisfiedLinkError: sun/management/FileSystemImpl.isSecuritySupported0 when starting TomEE via IntelliJ IDEA


alexbde:
Trying to deploy ear file on TomEE 1.7.5/7.1.0 via IntelliJ IDEA 2018.3.4 (TomEE run configuration) on Windows 7 x64 throws following exception:

```
C:\Users\example\apache-tomee\bin\catalina.bat run
C:\Users\example\jdk8u202-b08_openj9\bin\java.exe -Dfile.encoding=windows-1252 -classpath [...] com.intellij.javaee.oss.process.JavaeeProcess 50708 org.jetbrains.idea.tomcat.agent.TomEEAgent
Exception in thread "main"# START NON-TRANSLATABLE java/lang/UnsatisfiedLinkError: sun/management/FileSystemImpl.isSecuritySupported0(Ljava/lang/String;)Z
	in sun/management/FileSystemImpl.supportsFileSecurity (FileSystemImpl.java:37)
	in sun/management/jmxremote/ConnectorBootstrap.checkPasswordFile (ConnectorBootstrap.java:572)
	in sun/management/jmxremote/ConnectorBootstrap.startRemoteConnectorServer (ConnectorBootstrap.java:426)
	in sun/management/Agent.startAgent (Agent.java:262)
	in sun/management/Agent.startAgent (Agent.java:452)
	in java/lang/System.startSNMPAgent (NativeMethod:4294967295)
	in java/lang/Thread.completeInitialization (Thread.java:168)
	in java/lang/J9VMInternals.completeInitialization (J9VMInternals.java:72)
Using CATALINA_BASE:   "C:\Users\example\.IntelliJIdea2018.3\system\tomcat\Unnamed_3"
Using CATALINA_HOME:   "C:\Users\example\apache-tomee"
Using CATALINA_TMPDIR: "C:\Users\example\apache-tomee\temp"
Using JRE_HOME:        "C:\Users\example\jdk8u202-b08_openj9"
Using CLASSPATH:       "C:\Users\example\apache-tomee\bin\bootstrap.jar;C:\Users\example\apache-tomee\bin\tomcat-juli.jar"
```

It is working with Oracle OpenJDK and OpenJDK Hotspot (from AdoptOpenJDK).

This exception might be connected to #3909 (not quite sure).

DanHeidinga:
[...]

PR #4808, opened 2019-02-20
Add FileSystemImpl.isSecuritySupported0()

keithc-ca:
* generalize `translatePathToPlatform()` as `convertPath()`
* remove redundant uses of `JNIEXPORT` and `WIN64`

Fixes: #4776


keithc-ca:
Fixes #4776.

pshipton:
jenkins test sanity win jdk8

pshipton:

pshipton:
Reverting as this doesn't compile on some platforms
`"mgmtinit.c", line 113.35: 1506-280 (W) Function argument assignment between types "void*" and "const char*" is not allowed.`

PR #4810, opened 2019-02-20
Implemented TR_J9SharedCacheVM methods

dmitry-ten:
Implement all methods related to storing/retrieving thunks:
getJ2IThunk
setJ2IThunk
persistJ2IThunk
persistThunk
findPersistentThunk

### DIFF
--- a/runtime/compiler/env/VMJ9Server.cpp
+++ b/runtime/compiler/env/VMJ9Server.cpp
@@ -2202,3 +2202,72 @@ TR_J9SharedCacheServerVM::getClassFromNewArrayType(int32_t arrayType)
       return TR_J9ServerVM::getClassFromNewArrayType(arrayType);
    return NULL;
    }
+
+void *
+TR_J9SharedCacheServerVM::getJ2IThunk(char *signatureChars, uint32_t signatureLength, TR::Compilation *comp)
+   {
+   return (void *)findPersistentThunk(signatureChars, signatureLength);
+   }
+
+void *
+TR_J9SharedCacheServerVM::setJ2IThunk(char *signatureChars, uint32_t signatureLength, void *thunkptr, TR::Compilation *comp)
+   {
+   return persistThunk(signatureChars, signatureLength, (U_8*)thunkptr - 8 /* start of thunk */, *((U_32 *)((U_8*)thunkptr -8)) + 8 /* size of thunk */);
+   }
+
+void *
+TR_J9SharedCacheServerVM::persistJ2IThunk(void *thunk)
+   {
+   return persistThunk(((TR_J2IThunk *)thunk)->terseSignature(), strlen(((TR_J2IThunk *)thunk)->terseSignature()), (uint8_t*)thunk, ((TR_J2IThunk *)thunk)->totalSize());
+   }
+
+void *
+TR_J9SharedCacheServerVM::persistThunk(char *signatureChars, uint32_t signatureLength, uint8_t *thunkStart, uint32_t totalSize)
+   {
+#if defined(J9VM_INTERP_AOT_COMPILE_SUPPORT) && defined(J9VM_OPT_SHARED_CLASSES) && (defined(TR_HOST_X86) || defined(TR_HOST_POWER) || defined(TR_HOST_S390) || defined(TR_HOST_ARM))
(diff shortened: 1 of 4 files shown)


runtime/compiler/env/VMJ9Server.cpp
```diff
@@ -2202,3 +2202,72 @@ TR_J9SharedCacheServerVM::getClassFromNewArrayType(int32_t arrayType)
...
+void *
+TR_J9SharedCacheServerVM::getJ2IThunk(char *signatureChars, uint32_t signatureLength, TR::Compilation *comp)
+   {
+   return (void *)findPersistentThunk(signatureChars, signatureLength);
+   }
+
+void *
+TR_J9SharedCacheServerVM::setJ2IThunk(char *signatureChars, uint32_t signatureLength, void *thunkptr, TR::Compilation *comp)
+   {
+   return persistThunk(signatureChars, signatureLength, (U_8*)thunkptr - 8 /* start of thunk */, *((U_32 *)((U_8*)thunkptr -8)) + 8 /* size of thunk */);
```
mpirvu: This code dereferences a pointer (thunkptr-8). Does the information reside at the server?

runtime/compiler/env/VMJ9Server.cpp
```diff
@@ -2202,3 +2202,72 @@ TR_J9SharedCacheServerVM::getClassFromNewArrayType(int32_t arrayType)
...
+   }
+
+void *
+TR_J9SharedCacheServerVM::persistJ2IThunk(void *thunk)
+   {
+   return persistThunk(((TR_J2IThunk *)thunk)->terseSignature(), strlen(((TR_J2IThunk *)thunk)->terseSignature()), (uint8_t*)thunk, ((TR_J2IThunk *)thunk)->totalSize());
```
mpirvu: Same question here: does the think exist at the server at this point?

runtime/compiler/env/VMJ9Server.hpp
```diff
@@ -251,10 +251,11 @@ class TR_J9SharedCacheServerVM: public TR_J9ServerVM
...
    virtual TR::CodeCache * getDesignatedCodeCache(TR::Compilation *comp) override;
-   virtual void * getJ2IThunk(char *signatureChars, uint32_t signatureLength,  TR::Compilation *comp) override { TR_ASSERT(0, "called");  return 0; }
-   virtual void * setJ2IThunk(char *signatureChars, uint32_t signatureLength, void *thunkptr,  TR::Compilation *comp) override { TR_ASSERT(0, "called");  return 0; }
-   virtual void * persistJ2IThunk(void *thunk) override { TR_ASSERT(0, "called");  return 0; }
-   void * persistThunk(char *signatureChars, uint32_t signatureLength, uint8_t *thunkStart, uint32_t totalSize) { TR_ASSERT(0, "called");  return 0; }
+   virtual void * getJ2IThunk(char *signatureChars, uint32_t signatureLength,  TR::Compilation *comp) override;
+   virtual void * setJ2IThunk(char *signatureChars, uint32_t signatureLength, void *thunkptr,  TR::Compilation *comp) override;
+   virtual void * persistJ2IThunk(void *thunk) override;
+   void * persistThunk(char *signatureChars, uint32_t signatureLength, uint8_t *thunkStart, uint32_t totalSize);
```
mpirvu: This is marked virtual in `TR_J9SharedCacheVM`. Is the absence of `virtual` on purpose? I do not see a strong reason to be virtual, unless we want to extend the hierarchy in the future.

dmitry-ten: I was not meaning to drop `virtual`. I was removing `override` and accidentally deleted `virtual` as well. I think it's easier to make it virtual, just in case.

runtime/compiler/env/VMJ9Server.cpp
dmitry-ten: Thunks are generated during codegen on the server, so dereferencing it should be fine.

runtime/compiler/env/VMJ9Server.cpp
dmitry-ten: Same as above.

dmitry-ten:
Added `virtual` to `persistThunk`

PR #4812, opened 2019-02-20
Add avgCompTime.pl script

dsouzai:
This script, written by @mpirvu, is used to parse and summarize
verbose logs generated by the JIT. To run:

```
   perl avgCompTime.pl <vlog>
```

[ci skip]

### DIFF
--- a/runtime/compiler/tools/avgCompTime.pl
+++ b/runtime/compiler/tools/avgCompTime.pl
@@ -1,0 +1,305 @@
+##############################################################################
+#  Copyright (c) 2019, 2019 IBM Corp. and others
+#
+#  This program and the accompanying materials are made available under
+#  the terms of the Eclipse Public License 2.0 which accompanies this
+#  distribution and is available at https://www.eclipse.org/legal/epl-2.0/
+#  or the Apache License, Version 2.0 which accompanies this distribution and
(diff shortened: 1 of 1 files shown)


dsouzai:
@mpirvu would you do the honours?

fjeremic:
Any reason why we can't just aggregate all this data in the JIT itself rather than using a post-processing script?

mpirvu:
The JIT does aggregation of some metrics which are printed to stderr at the end of the run. I use them as a quick sanity check of the run. For instance:
```
Number of compilations per level:
Level=1 numComp=1375
Level=2 numComp=167
Level=3 numComp=2
NumJNIMethodsCompiled=197
NumMethodsFoundInSharedCache=3186
NumMethodsTakenFromSharedCache=3021
NumAotedMethods=13
NumberOfAotedMethodsThatWereRecompiled=3 (forced=3)
Classes loaded=15469
GCR bodies=1016, GCRSaves=172 GCRRecomp=0
Compilation queue peak size = 85
Allocated memory for code cache = 8192 KB       Limit = 262144 KB
Allocated memory for data cache = 8192 KB       Limit = 393216 KB
Time spent in compilation thread =1998 ms
```
However, not all runs finish nicely, so we may lose them.

PR #4822, opened 2019-02-21
[WIP] JVM_GetVmArguments returns User arguments only

### LINKED ISSUE #2974: Distinguish user arguments from generated arguments
Currently it cannot be distinguished whether arguments in J9VMInitArgs are created by the vm or by the user. `JVM_GetVmArguments` is expected to return only the user generated arguments. `JVM_GetVmArguments` should be updated to reflect this when the information distinguishing user args or generated args is available. @pdbain-ibm is working on making this information available.

AidanHa:
Add a `USER_ARG` flag for J9vmInitArg, and set the flag in addLauncherArgs (when processing arguments by the user).

WIP since option files and other methods of user inputs have been considered.

Fixes #2974


pshipton:
Note the copyright check failure

pdbain-ibm:
Wouldn't it be easier to put `USER_ARG` constant directly in the statement, instead of assigning it to a variable?

AidanHa:
Could I get a review for this PR. I think the solution I implemented could use a lot of work. Here is what I've done.

-Added a flag `USER_ARG` to differentiate user arguments from VM arguments. The following are considered "User Arguments"
    -User args from `addLauncherArgs` with the exception of`-Djava.class.path=.`, `-Dsun.java.command=Test`, `-Dsun.java.launcher=SUN_STANDARD` and `-Dsun.java.launcher.pid=75488`. This includes args entered from the command line
     -User args from the `OPENJ9_JAVA_OPTIONS=` environment variable.
    -User args from any Jar Arguments (`addJarArguments`).
    -User args from any Optionfiles except the default one.

runtime/util/vmargs.c
```diff
@@ -1496,11 +1499,18 @@ addLauncherArgs(J9PortLibrary * portLib, JavaVMInitArgs *launcherArgs, UDATA lau
 		}
 
 		if (TRUE == convertEncoding) {
-			optArg = newJavaVMArgInfo(vmArgumentsList, transcodedString, ARG_MEMORY_ALLOCATION|consumableFlag);
+			if ((strncmp(optString, "-Djava.class.path=", strlen("-Djava.class.path=")) != 0) && (strncmp(optString, "-Dsun.java.command=", strlen("-Dsun.java.command=")) != 0) && (strncmp(optString, "-Dsun.java.launcher=", strlen("-Dsun.java.launcher=")) != 0) && (strncmp(optString, "-Dsun.java.launcher.pid=", strlen("-Dsun.java.launcher.pid=")) != 0)) {
```
AidanHa: Not too sure on how I can exclude the default options in this if statement. Does anyone have any ideas?

runtime/jcl/common/com_ibm_oti_vm_VM.c
```diff
@@ -69,11 +71,12 @@ Java_com_ibm_oti_vm_VM_getVMArgsImpl(JNIEnv *env, jobject recv)
 					if (NULL == option) {
 						/* Don't use break here to avoid triggering the assertion below */
 						return NULL;
+					} else if (IS_USER_ARG( j9vm_args, i )) {
```
theresa-m: I would add an extra comment in here somewhere to explain the third case

runtime/util/vmargs.c
theresa-m: I don't think this will catch all the system arguments based on the output from #2974 such as `-Xjcl:jclse11_29`. what was your command line output?

runtime/oti/vmargs_core_api.h
```diff
@@ -74,6 +76,19 @@ newJavaVMArgInfo(J9JavaVMArgInfoList *vmArgumentsList, char *optString, uintptr_
...
+ * This allocates memory for the options string and contents of the file
+ * @param portLib port library
+ * @param xOptionsfileArg -Xoptionsfile=<path>
+ * @param vmArgumentsList current list of arguments
+ * @param verboseFlags set to VERBOSE_INIT for verbosity
+ * @param isDefault set to TRUE if using default optionfile
+ * @return 0 on success, negative value on failure
+ */
+intptr_t
+addXOptionsFileImpl(J9PortLibrary* portLib, const char *xOptionsfileArg, J9JavaVMArgInfoList *vmArgumentsList, uintptr_t verboseFlags, UDATA isDefault);
```
theresa-m: Modify addXOptionsFile and parseOptionsBuffer directly instead of creating a wrapper function. It's harder to read this way in my opinion

runtime/util/vmargs.c
theresa-m: do you have any ideas for doing this other than hard coding @pdbain-ibm ?

AidanHa: Without the changes, the output for me was:

-Xoptionsfile=/root/openj9-openjdk-jdk11/build/linux-x86_64-normal-server-release/images/jdk/lib/options.default, -Xlockword:mode=default,noLockword=java/lang/String,noLockword=java/util/MapEntry,noLockword=java/util/HashMap$Entry,noLockword=org/apache/harmony/luni/util/ModifiedMap$Entry,noLockword=java/util/Hashtable$Entry,noLockword=java/lang/invoke/MethodType,noLockword=java/lang/invoke/MethodHandle,noLockword=java/lang/invoke/CollectHandle,noLockword=java/lang/invoke/ConstructorHandle,noLockword=java/lang/invoke/ConvertHandle,noLockword=java/lang/invoke/ArgumentConversionHandle,noLockword=java/lang/invoke/AsTypeHandle,noLockword=java/lang/invoke/ExplicitCastHandle,noLockword=java/lang/invoke/FilterReturnHandle,noLockword=java/lang/invoke/DirectHandle,noLockword=java/lang/invoke/ReceiverBoundHandle,noLockword=java/lang/invoke/DynamicInvokerHandle,noLockword=java/lang/invoke/FieldHandle,noLockword=java/lang/invoke/FieldGetterHandle,noLockword=java/lang/invoke/FieldSetterHandle,noLockword=java/lang/invoke/StaticFieldGetterHandle,noLockword=java/lang/invoke/StaticFieldSetterHandle,noLockword=java/lang/invoke/IndirectHandle,noLockword=java/lang/invoke/InterfaceHandle,noLockword=java/lang/invoke/VirtualHandle,noLockword=java/lang/invoke/PrimitiveHandle,noLockword=java/lang/invoke/InvokeExactHandle,noLockword=java/lang/invoke/InvokeGenericHandle,noLockword=java/lang/invoke/VarargsCollectorHandle,noLockword=java/lang/invoke/ThunkTuple, -Xjcl:jclse29, 
-Dcom.ibm.oti.vm.bootstrap.library.path=/root/openj9-openjdk-jdk11/build/linux-x86_64-normal-server-release/images/jdk/lib/compressedrefs:/root/openj9-openjdk-jdk11/build/linux-x86_64-normal-server-release/images/jdk/lib, -Dsun.boot.library.path=/root/openj9-openjdk-jdk11/build/linux-x86_64-normal-server-release/images/jdk/lib/compressedrefs:/root/openj9-openjdk-jdk11/build/linux-x86_64-normal-server-release/images/jdk/lib, -Djava.library.path=/root/openj9-openjdk-jdk11/build/linux-x86_64-normal-server-release/images/jdk/lib/compressedrefs:/root/openj9-openjdk-jdk11/build/linux-x86_64-normal-server-release/images/jdk/lib:/usr/lib64:/usr/lib, -Djava.home=/root/openj9-openjdk-jdk11/build/linux-x86_64-normal-server-release/images/jdk, 
-Duser.dir=/root/openj9-openjdk-jdk11/build/linux-x86_64-normal-server-release/images/jdk/bin, 
--add-exports=java.base/com.ibm.oti.vm=ALL-UNNAMED, 
-Djava.class.path=., 
-Dsun.java.command=Test, 
-Dsun.java.launcher=SUN_STANDARD, 
-Dsun.java.launcher.pid=49456]

theresa-m: right so those extra few 
`-Xoptionsfile=/root/openj9-openjdk-jdk11/build/linux-x86_64-normal-server-release/images/jdk/lib/options.default,
-Xlockword:mode=default,noLockword=java/lang/String,noLockword=java/util/MapEntry,noLockword=java/util/HashMap$Entry,noLockword=org/apache/harmony/luni/util/ModifiedMap$Entry,noLockword=java/util/Hashtable$Entry,noLockword=java/lang/invoke/MethodType,noLockword=java/lang/invoke/MethodHandle,noLockword=java/lang/invoke/CollectHandle,noLockword=java/lang/invoke/ConstructorHandle,noLockword=java/lang/invoke/ConvertHandle,noLockword=java/lang/invoke/ArgumentConversionHandle,noLockword=java/lang/invoke/AsTypeHandle,noLockword=java/lang/invoke/ExplicitCastHandle,noLockword=java/lang/invoke/FilterReturnHandle,noLockword=java/lang/invoke/DirectHandle,noLockword=java/lang/invoke/ReceiverBoundHandle,noLockword=java/lang/invoke/DynamicInvokerHandle,noLockword=java/lang/invoke/FieldHandle,noLockword=java/lang/invoke/FieldGetterHandle,noLockword=java/lang/invoke/FieldSetterHandle,noLockword=java/lang/invoke/StaticFieldGetterHandle,noLockword=java/lang/invoke/StaticFieldSetterHandle,noLockword=java/lang/invoke/IndirectHandle,noLockword=java/lang/invoke/InterfaceHandle,noLockword=java/lang/invoke/VirtualHandle,noLockword=java/lang/invoke/PrimitiveHandle,noLockword=java/lang/invoke/InvokeExactHandle,noLockword=java/lang/invoke/InvokeGenericHandle,noLockword=java/lang/invoke/VarargsCollectorHandle,noLockword=java/lang/invoke/ThunkTuple, -Xjcl:jclse29,
`
will need excluding as well

AidanHa: Right, with my changes the extra few options are already excluded (since those arguments are not added to the the argument array (`vm_args`) when `addLauncherArgs()` is called. It's just that unfortunately, the arguments that are added to `vm_args` before `addLauncherArgs()` is called, are all reported with my current implementation.

AidanHa: Perhaps if anyone could help me identify where the arguments (`-Djava.class`...etc.) are added, I might be able to find a better solution. But ATM, the only solution I have is hard coding it.

pdbain: > do you have any ideas for doing this other than hard coding @pdbain-ibm ?

Not that I can think of just now.  These come from the Java launcher, which comes from OpenJDK.

You should also include arguments from `IBM_JAVA_OPTIONS` and `JAVA_TOOL_OPTIONS`

AidanHa: I was thinking of creating an "Ignored  user options table" (similiar to `J9VMIgnoredOption ignoredOptionTable[]` in jvminit.c) where arguments in the table are parsed beforehand so that they are not returned. While is is still "hard coded", it's a lot more elegant than our current implementation...

AidanHa: @pdbain I've pushed changes to include arguments from `IBM_JAVA_OPTIONS` and `JAVA_TOOL_OPTIONS` :)

pdbain-ibm: Sounds like a good plan.  Keep it in sorted order so you can search in O(lg N).

runtime/util/vmargs.c
```diff
@@ -76,8 +76,41 @@
...
+UDATA
+findArgInIgnoreTable(const char *optionName, IDATA start, IDATA end) {
+	IDATA mid = (start + end) / 2;
+	if (strncmp(JavaLauncherOptionTable[mid].optionName, optionName, strlen(JavaLauncherOptionTable[mid].optionName)) == 0) {
+	/* The search function found the given option name in the table. */
+		return TRUE;
+	} else if (start == end) {
+		/* The search function did not find the given option name in the table. */
+		return FALSE;
+	} else if (strncmp(JavaLauncherOptionTable[mid].optionName, optionName, strlen(JavaLauncherOptionTable[mid].optionName)) < 0) {
```
pdbain-ibm: Please precompute the lengths and put them in a separate table.  `sizeof(theString) -1` is the safest way.  Use `#define`s so you don't need to type the string twice.

runtime/util_core/vmargs_core.c
```diff
@@ -71,7 +71,7 @@ newJavaVMArgInfo(J9JavaVMArgInfoList *vmArgumentsList, char *optString, uintptr_
  */
 
 intptr_t
-parseOptionsFileText(J9PortLibrary* portLibrary, const char* fileText, J9JavaVMArgInfoList *vmArgumentsList, uintptr_t verboseFlags)
+parseOptionsFileText(J9PortLibrary* portLibrary, const char* fileText, J9JavaVMArgInfoList *vmArgumentsList, uintptr_t verboseFlags, UDATA isDefault)
```
pdbain-ibm: Make this a more generic name, like `containsUserArgs`.

runtime/util_core/vmargs_core.c
```diff
@@ -201,8 +201,14 @@ parseOptionsFileText(J9PortLibrary* portLibrary, const char* fileText, J9JavaVMA
 					 */
 					memcpy(nextOptionPosition, tokenStart, argumentLength);
 					/* The first argument is at the head of the buffer and is the same pointer as the allocated buffer */
-					optArg = newJavaVMArgInfo(vmArgumentsList, nextOptionPosition,
+					if (isDefault) {
```
pdbain-ibm: These two arms are the same apart from the third argument, right?  If so, use the ternary operator to generate the third argument.

runtime/util_core/vmargs_core.c
```diff
@@ -277,8 +283,13 @@ addXOptionsFile(J9PortLibrary* portLib, const char *xOptionsfileArg, J9JavaVMArg
 		return -1;
 	}
 
-	/* add the -Xoptionsfile argument */
-	optArg = newJavaVMArgInfo(vmArgumentsList, NULL, ARG_MEMORY_ALLOCATION|CONSUMABLE_ARG);
+	/* add the -Xoptionsfile argument. Sets the USER_ARG flag if the optionfile is not the default optionfile */
+	if (isDefault) {
```
pdbain-ibm: Ibid.

runtime/vm/jvminit.c
```diff
@@ -2606,34 +2606,34 @@ modifyDllLoadTable(J9JavaVM * vm, J9Pool* loadTable, J9VMInitArgs* j9vm_args)
...
 		if (strcmp(testString, VMOPT_XINT)==0) {
 			if (xjit || xnojit || xnoaot || xaot) {
-				SET_FLAG_AT_INDEX( NOT_CONSUMABLE_ARG, i );
+				SET_FLAG_AT_INDEX( NOT_CONSUMABLE_ARG | (j9vm_args->j9Options[i].flags & USER_ARG), i );
 			} else {
 				xint = TRUE;
 			}
 		} else if ((strcmp(testString, VMOPT_XJIT)==0) ||
 					(strstr(testString, VMOPT_XJIT_COLON)==testString)) {
 			if (xint || xnojit) {
```
pdbain-ibm: Can you grab `(j9vm_args->j9Options[i].flags & USER_ARG)` into a local variable and use that instead?

runtime/util/vmargs.c
AidanHa: Should I put the precomputed length in a separate table? it might be easier to include them in the `JavaLauncherOption` object...

pdbain-ibm: > it might be easier to include them in the JavaLauncherOption object

Good idea.

runtime/util/vmargs.c
```diff
@@ -76,8 +76,41 @@
...
+	"-Dsun.java.launcher=",
+	"-Xjcl:jclse11_29",
+};
+
+#define ignoredOptionTableEnd (sizeof(JavaLauncherOptionTable) / sizeof(struct JavaLauncherOption) - 1)
+
 static char * getStartOfOptionValue(J9VMInitArgs* j9vm_args, IDATA element, const char *optionName);
 
+UDATA
+findArgInIgnoreTable(const char *optionName, IDATA start, IDATA end) {
```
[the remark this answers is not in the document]
AidanHa: The comment will be added in `vmargs_api.h`

runtime/util_core/vmargs_core.c
AidanHa: This option is set to TRUE if it is a default option file. Perhaps a name like "containsDefaultArgs" might be less confusing.

AidanHa:
Hello,

> I was thinking of creating an "Ignored user options table" (similar to J9VMIgnoredOption ignoredOptionTable[] in jvminit.c) where arguments in the table are parsed beforehand so that they are not returned. While is is still "hard coded", it's a lot more elegant than our current implementation...

The table I implemented seems to work as expected. However the issue with this solution is that if the user adds an option from this table into their command line (eg: `-Djava.class.path=.`, then that option would not be returned in `JVM_GetVmArguments()`.

I was wondering if those generated options can be called by the user? Or are those options strictly for the OpenJDK Java Launcher?

theresa-m:
the changes look good to me. can you also add a test to Java8AndUp to verify that VM.getVMArgs is returning the expected output?

runtime/util/vmargs.c
```diff
@@ -76,8 +76,46 @@
...
 static char * getStartOfOptionValue(J9VMInitArgs* j9vm_args, IDATA element, const char *optionName);
 
+UDATA
+isGeneratedArgument(const char *optionName) {
+	IDATA start = 0;
+	IDATA end = ignoredOptionTableEnd;
+	IDATA mid = 0;
+	while (start <= end) {
+		mid = (start + end) / 2;
+		IDATA optionCompare = strncmp(optionName, JavaLauncherOptionTable[mid].optionName, JavaLauncherOptionTable[mid].optionNameLength);
```
pdbain-ibm: This probably won't compile on Windows.  declare `optionCompare` before any executable code.  `compareResult` would be amore meaningful name.

runtime/util/vmargs.c
```diff
@@ -76,8 +76,46 @@
...
+		if (optionCompare == 0) {
+			/* The search function found the given option name in the table. */
+			return TRUE;
```
pdbain-ibm: A single return point is preferred. Declare and return a status variable; initialize it to FALSE, if you find a match set it to TRUE and break the loop.

runtime/util/vmargs.c
```diff
@@ -1496,11 +1537,20 @@ addLauncherArgs(J9PortLibrary * portLib, JavaVMInitArgs *launcherArgs, UDATA lau
 		}
 
 		if (TRUE == convertEncoding) {
-			optArg = newJavaVMArgInfo(vmArgumentsList, transcodedString, ARG_MEMORY_ALLOCATION|consumableFlag);
+			if(!isGeneratedArgument(transcodedString)) {
+				/* We did not find the argument in the table of Java Launcher Args, so we treat it as a user arg. */
+				optArg = newJavaVMArgInfo(vmArgumentsList, transcodedString, ARG_MEMORY_ALLOCATION|CONSUMABLE_ARG|USER_ARG);
+			} else {
+				optArg = newJavaVMArgInfo(vmArgumentsList, transcodedString, ARG_MEMORY_ALLOCATION|CONSUMABLE_ARG);
```
pdbain-ibm: Duplicated code: use the ternary operator to add either `CONSUMABLE_ARG|USER_ARG` or `CONSUMABLE_ARG`.

runtime/util/vmargs.c
```diff
@@ -1496,11 +1537,20 @@ addLauncherArgs(J9PortLibrary * portLib, JavaVMInitArgs *launcherArgs, UDATA lau
...
+			}
 		} else {
-
 			strcpy(cursor, optString); /* safe because the lengths of all strings have been measured and included in argumentLength */
-			optArg = newJavaVMArgInfo(vmArgumentsList, cursor, (consumableFlag | ((0 == optInd) ? ARG_MEMORY_ALLOCATION: 0)));
+			if(!isGeneratedArgument(cursor)) {
+				/* We did not find the argument in the table of Java Launcher Args, so we treat it as a user arg. */
+				optArg = newJavaVMArgInfo(vmArgumentsList, cursor, (USER_ARG | CONSUMABLE_ARG | ((0 == optInd) ? ARG_MEMORY_ALLOCATION: 0)));
+			} else {
+				optArg = newJavaVMArgInfo(vmArgumentsList, cursor, (CONSUMABLE_ARG | ((0 == optInd) ? ARG_MEMORY_ALLOCATION: 0)));
```
pdbain-ibm: ibid.

AidanHa:
For reference This is an example command line that I used to test the changes:

Environment variables:
`IBM_JAVA_OPTIONS=-XX:+LazySymbolResolution
JAVA_TOOL_OPTIONS=-XX:+HandleSIGXFSZ
OPENJ9_JAVA_OPTIONS=-Dmysysprop1=tcpip -Dmysysprop2=wait -Xdisablejavadump'

Optionsfile content: `-XX:+UseNoGC`

META-INF/MANIFEST.MF content: `-Xshareclasses:name=mycache,nonfatal,cacheDirPerm=1000`

Command line : `./java --add-exports java.base/com.ibm.oti.vm=ALL-UNNAMED -Djava.class.path=. -Xoptionsfile=aidan -jar Test.jar`

Output:
`-Xshareclasses:name=mycache,nonfatal,cacheDirPerm=1000` 
`-XX:+HandleSIGXFSZ` 
`-Dmysysprop1=tcpip` 
`-Dmysysprop2=wait` 
`-Xdisablejavadump` 
`-XX:+LazySymbolResolution` 
`--add-exports=java.base/com.ibm.oti.vm=ALL-UNNAMED` 
`-Xoptionsfile=aidan` 
`-XX:+UseNoGC`

pdbain-ibm:
Please include your unit tests in this pull request.  I suggest augmenting `VmArgumentTests`.

runtime/util/vmargs.c
AidanHa: Ah sorry, must've missed this

runtime/util/vmargs.c
AidanHa: I was thinking of simply breaking the loop `if (optionCompare == 0)` and return `(optionCompare == 0)? TRUE : FALSE;` 
This way, we don't need to introduce a new variable for the return status

pdbain-ibm: You could do that, provided you initialize the variable suitably. 
Style comments: 
- put the rvalue (e.g. a constant) on the left side of a `==` or `!=` for safety.
- `(optionCompare == 0)? TRUE : FALSE` is redundant, since `(optionCompare == 0)` evaluates to `TRUE` or `FALSE`.

AidanHa:
Could I get a review for this PR? Note that the copyright check fails since I added a new MANIFEST.MF file, and AFAIK, there is no way to comment on a manifest file...

test/functional/Java8andUp/src/org/openj9/test/contendedfields/ContendedFieldsTests.java
```diff
@@ -63,7 +65,8 @@ static void getVMConfig() { /* ideally we should have an mxbean to get this info
 			CACHE_LINE_SIZE = 64;			
 		}
 		jep142Restricted = true;
-		for (String vmarg: ManagementFactory.getRuntimeMXBean().getInputArguments()) {
+		RuntimeMXBean bean = (RuntimeMXBean)ManagementFactory.getRuntimeMXBean();
```
pdbain-ibm: `-XX:-RestrictContended` is a user argument, so it should show up in the regular `getInputArguments()`.

test/functional/Java8andUp/src/org/openj9/test/vmArguments/UserArgumentDumper.java
```diff
@@ -0,0 +1,50 @@
...
+import java.util.Set;
+
+@SuppressWarnings("nls")
+public class UserArgumentDumper {
+    public static final String USERARG_TAG = "2CIUSERARG";
+    public static void main(String[] args) {
+
+        System.out.println(USERARG_TAG);
+        RuntimeMXBean bean = (RuntimeMXBean)ManagementFactory.getRuntimeMXBean();
+        for (String a: bean.getInputArguments()) {
```
pdbain-ibm: Assuming most of this code is common with `ArgumentDumper.java`, you can extract it to an abstract superclass which calls an abstract `getVMArguments()` method.

jcl/src/java.base/share/classes/com/ibm/oti/vm/VM.java
```diff
@@ -375,25 +374,42 @@ private static void shutdown() {
...
+	byte[][] byteArgs = (getUserArgsOnly) ? getVMArgsImpl() : getAllVMArgsImpl();
+	if (byteArgs == null) return null;
+	String[] result = new String[byteArgs.length];
+	for (int i=0; i<byteArgs.length; i++) {
+		if (byteArgs[i] == null) {
+			result[i] = null;
+		} else {
+			result[i] = Util.toString(byteArgs[i]);
 		}
-		cachedVMArgs = result;
```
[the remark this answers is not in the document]
AidanHa: Strange, I thought I left a comment...
The reason why we don't cache the VM arguments is because if we were to call getVMArgs() and getAllVMArgs() in the same java file, the function that is called second would return the results from the first.

Eg, if i called getVMArgs(), then getAllVMArgs() right after, getVMArgs() would return the user arguments only (since only the user arguments are cached)

runtime/util/vmargs.c
```diff
@@ -76,8 +76,47 @@
...
 
+struct JavaLauncherOption {
+	char *optionName;
+	int optionNameLength;
+};
+
+/* options in this table will be not treated as a user Argument 
+Note, these options need to be in lexical order (ASCII values) for the binary search function to properly work */
+static const struct JavaLauncherOption JavaLauncherOptionTable[] = {
+	{"-Djava.class.path=", strlen("-Djava.class.path=")},
```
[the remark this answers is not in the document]
AidanHa: Should I use the macros in `jvminit.h`? For example `-Xjcl:` is defined as `VMOPT_XJCL_COLON`. In my opinion it's a bit unclear

jcl/src/java.base/share/classes/com/ibm/oti/vm/VM.java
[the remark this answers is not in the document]
pdbain-ibm: You could cache both sets: user args and all args.  If you want to be clever, you could store the disjoint sets (user and non-user) and return  their unions if you request all args.

runtime/util/vmargs.c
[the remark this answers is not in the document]
pdbain-ibm: If you have a macro already defined, then by all means use it.

test/functional/JIT_Test/src/jit/test/jitt/codecache/TestManager.java
```diff
@@ -282,8 +282,8 @@ private static boolean isAOTVMOptionsSet() {
 	 * @return true if -Xshareclasscache VM option is in use, false otherwise
 	 */
 	public static boolean isShareClassesSet() {
-		RuntimeMXBean RuntimemxBean = ManagementFactory.getRuntimeMXBean();
-		List<String> aList=RuntimemxBean.getInputArguments();
+		RuntimeMXBean RuntimemxBean = (RuntimeMXBean)ManagementFactory.getRuntimeMXBean();
+		List<String> aList=RuntimemxBean.getAllInputArguments();
```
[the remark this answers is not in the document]
AidanHa: How exactly should I implement a helper method here? In my opinion, this might be out of the scope of the PR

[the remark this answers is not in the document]
AidanHa: Rather than a helper method, I was thinking of declaring `RuntimemxBean` and `aList` as a variable/member of the class. I would call getAllInputArguments() only once.

runtime/vm/jvminit.c
```diff
@@ -2625,36 +2625,38 @@ modifyDllLoadTable(J9JavaVM * vm, J9Pool* loadTable, J9VMInitArgs* j9vm_args)
 		}
 #endif
 
+		uintptr_t argUserFlag = j9vm_args->j9Options[i].flags & USER_ARG;
+
 		if (strcmp(testString, VMOPT_XINT)==0) {
 			if (xjit || xnojit || xnoaot || xaot) {
-				SET_FLAG_AT_INDEX( NOT_CONSUMABLE_ARG, i );
+				SET_FLAG_AT_INDEX( NOT_CONSUMABLE_ARG | argUserFlag, i );
```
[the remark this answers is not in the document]
AidanHa: I was a bit unclear on this part of the code as well... I thought it was wrriten here for a reason, so I didn't change. Should I just remove this if statement?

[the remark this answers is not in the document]
AidanHa: This still needs to be addressed. From  my point of view, this line of code were added for a reason, and I don't want to remove it until I'm 100 percent certain that it's fine doing so.

jcl/src/java.base/share/classes/com/ibm/oti/vm/VM.java
```diff
@@ -389,22 +391,29 @@ private static void shutdown() {
...
  */
 private static String [] getVMArgsHelper(boolean getUserArgsOnly) {
 	/*[PR CMVC 112580] Java cannot handle non-ascii character arguments */
-	byte[][] byteArgs = (getUserArgsOnly) ? getVMArgsImpl() : getAllVMArgsImpl();
-	if (byteArgs == null) return null;
-	String[] result = new String[byteArgs.length];
-	for (int i=0; i<byteArgs.length; i++) {
-		if (byteArgs[i] == null) {
-			result[i] = null;
+	if (((cachedVMUserArgs == null) && getUserArgsOnly) || ((cachedVMGeneratedArgs == null) && !getUserArgsOnly)) {
```
pdbain-ibm: The ternary operator would be clearer.

jcl/src/java.base/share/classes/com/ibm/oti/vm/VM.java
```diff
@@ -389,22 +391,29 @@ private static void shutdown() {
...
+		}
+		if (getUserArgsOnly) {
+			cachedVMUserArgs = result;
 		} else {
-			result[i] = Util.toString(byteArgs[i]);
+			cachedVMGeneratedArgs = result;
 		}
 	}
-	return (String[])result.clone();
+	return (String[])((getUserArgsOnly)? cachedVMUserArgs : cachedVMGeneratedArgs).clone();
```
pdbain-ibm: Use `Arrays.copyOf()`.

runtime/oti/vmargs_api.h
```diff
@@ -60,7 +60,7 @@
  * The entries in this table represent arguments that are generated by the Java Launcher in OpenJDK.
  * Note that this is a binary search function. This means that the entries must be in lexical order.
  */
-UDATA
+int
 isGeneratedArgument(const char *optionName);
```
pdbain-ibm: If this is a boolean function, why not return BOOLEAN?

runtime/util/vmargs.c
```diff
@@ -76,32 +76,37 @@
...
 
 /* Binary Search function that searches if optionName is in JavaLauncherOptionTable */
-UDATA
+int
 isGeneratedArgument(const char *optionName) {
-	IDATA start = 0;
-	IDATA end = ignoredOptionTableEnd;
-	IDATA mid = 0;
-	IDATA compareResults = 0;
+	int start = 0;
```
pdbain-ibm: Check with GAC or Keith as to the preferred type in this case.

test/functional/Java8andUp/src/org/openj9/test/vmArguments/UserArgumentDumper.java
```diff
@@ -26,25 +26,24 @@
...
-        System.out.print("\n===============\n\nSystem properties\n\n");
-        Properties props = System.getProperties();
-        Set<String> pns = props.stringPropertyNames();
-        for (String pn: pns) {
-            String p = props.getProperty(pn).replaceAll("\n", "\\n");
-            System.out.print(pn);
-            System.out.print("=");
-            System.out.println(p);
-        }
+        AbstractDumper dumper = new AbstractUserArgumentDumper();
```
pdbain-ibm: You cannot create instances of abstract classes.

runtime/util/vmargs.c
AidanHa: I was recommended to use UDATA

AidanHa: I was recommended to use UDATA

runtime/util/vmargs.c
AidanHa: I've decided to switch to a linear search as requested by Keith for the following:
-It might be overkill to implement a binary search since were only reducing the number of loops we’re doing by 2 (MAX)
-With a linear search, the only variable we need to declare is a pointer, which is much cleaner than having a start/end/mid variable.
-The table will not need to be in Lexical Order (edited)

AidanHa: Regarding your comment earlier about having a single return point, would you still like me to implement that? I find that it might not be as effective for the following reasons:

-If we decide to have a boolean variable storing the return value upon exiting the loop, then we introduce another variable in the function
-If we don't introduce a new variable, then the only other solution is to return another call of `strncmp()`...

runtime/util/vmargs.c
```diff
@@ -76,8 +76,41 @@
...
+	{SUN_JAVA_LAUNCHER_EQUALS, LITERAL_STRLEN(SUN_JAVA_LAUNCHER_EQUALS)},
+	{VMOPT_XJCL_COLON, LITERAL_STRLEN(VMOPT_XJCL_COLON)}, /* macro from jvminit.h */
+	{NULL, 0} /* The last entry in the table must be Null in order for the search function to exit.*/
+};
+
 static char * getStartOfOptionValue(J9VMInitArgs* j9vm_args, IDATA element, const char *optionName);
 
+/* Binary Search function that searches if optionName is in JavaLauncherOptionTable */
+BOOLEAN
+isGeneratedArgument(const char *optionName) {
```
AidanHa: I've decided to switch to a linear search as requested by Keith for the following:
-It might be overkill to implement a binary search since were only reducing the number of loops we’re doing by 2 (MAX)
-With a linear search, the only variable we need to declare is a pointer, which is much cleaner than having a start/end/mid variable.
-The table will not need to be in Lexical Order

AidanHa:
Up and ready for another review!

runtime/jcl/common/com_ibm_oti_vm_VM.c
```diff
@@ -46,7 +62,7 @@ Java_com_ibm_oti_vm_VM_getVMArgsImpl(JNIEnv *env, jobject recv)
 	/* Count only options that begin with "-" */
 
 	for (i = 0; i < originalSize; ++i) {
-		if ('-' == options[i].optionString[0]) {
+		if (('-' == options[i].optionString[0]) && (!getUserArgsOnly || IS_USER_ARG( j9vm_args, i ))) {
```
tajila: Minor nitpick: `IS_USER_ARG( j9vm_args, i )` is how macros were used with the old coding guidelines, typically for new code we just do `IS_USER_ARG(j9vm_args, i)`

runtime/oti/vmargs_api.h
```diff
@@ -55,6 +55,13 @@
 #define ENV_LD_LIB_PATH "LD_LIBRARY_PATH"
 #endif /* defined(OSX) */
 
+/**
+ * Walks the JavaLauncherOptionTable to find the entry for a given argument.
+ * The entries in this table represent arguments that are generated by the Java Launcher in OpenJDK.
```
tajila: also describe parameters and returns, see `addOptionsDefaultFile` for an example

runtime/util/vmargs.c
```diff
@@ -76,8 +76,41 @@
...
+	const struct JavaLauncherOption *tableOptionName = JavaLauncherOptionTable;
+	while (tableOptionName->optionName != NULL) {
+		if (0 == strncmp(optionName, tableOptionName->optionName, tableOptionName->optionNameLength)) {
```
tajila: strncmp will match substrings, use `J9UTF8_DATA_EQUALS` instead

runtime/util/vmargs.c
```diff
@@ -1355,8 +1388,12 @@ addEnvironmentVariableArguments(J9PortLibrary* portLib, const char* envVarName,
 	}
 	JVMINIT_VERBOSE_INIT_TRACE1(verboseFlags, "Parsing environment variable %s\n", envVarName);
 	j9sysinfo_get_env(envVarName, argumentBuffer, envVarSize);
-
-	return parseOptionsBuffer(portLib, argumentBuffer, vmArgumentsList, verboseFlags, TRUE);
+	if ((0 == strcmp(ENVVAR_JAVA_TOOL_OPTIONS, envVarName)) || (0 == strcmp(ENVVAR_OPENJ9_JAVA_OPTIONS, envVarName)) || (0 == strcmp(ENVVAR_IBM_JAVA_OPTIONS, envVarName))) {
```
tajila: same here

runtime/util/vmargs.c
```diff
@@ -1512,6 +1547,11 @@ addLauncherArgs(J9PortLibrary * portLib, JavaVMInitArgs *launcherArgs, UDATA lau
 	return 0;
 }
 
+#undef JAVA_CLASS_PATH_EQUALS
```
tajila: why are these undef'd

runtime/util_core/vmargs_core.c
```diff
@@ -360,11 +370,11 @@ parseOptionsBuffer(J9PortLibrary* portLib, char* argumentBuffer, J9JavaVMArgInfo
 		if (tokenEnd) {
 			if ((0 == strncmp(tokenStart, VMOPT_XOPTIONSFILE_EQUALS, strlen(VMOPT_XOPTIONSFILE_EQUALS))) && (TRUE == parseOptionsFileFlag)) {
 				/* addXOptionsFile allocates a new buffer for the -Xoptionsfile argument and the file contents */
-				if (0 != addXOptionsFile(portLib, tokenStart, vmArgumentsList, verboseFlags)) {
+				if (0 != addXOptionsFile(portLib, tokenStart, vmArgumentsList, verboseFlags, TRUE)) {
 					return -1;
 				}
 			} else {
-				J9JavaVMArgInfo *optArg = newJavaVMArgInfo(vmArgumentsList, NULL, CONSUMABLE_ARG);
+				optArg = newJavaVMArgInfo(vmArgumentsList, NULL, CONSUMABLE_ARG | (parseEnvVar)? USER_ARG : 0);
```
tajila: Nitpick: parseEnvVar doesn't need the brackets. Also there should be a space before the `?`

runtime/vm/jvminit.c
```diff
@@ -2625,36 +2625,38 @@ modifyDllLoadTable(J9JavaVM * vm, J9Pool* loadTable, J9VMInitArgs* j9vm_args)
 		}
 #endif
 
+		uintptr_t argFlag = ((NOT_CONSUMABLE_ARG) | ((j9vm_args->j9Options[i].flags) & (USER_ARG)));
```
tajila: same here with the extra brackets

test/functional/JIT_Test/src/jit/test/jitt/codecache/TestManager.java
```diff
@@ -69,6 +69,10 @@
 	/*Total number of code caches being used at startup, this is equivalent to -Xjit:numOfCodeCachesOnStartup*/
 	private static int numCodeCacheOnStartup = 0;
 
+	/* list of all input arguments from RuntimemxBean */
+	private static RuntimeMXBean RuntimemxBean = (RuntimeMXBean)ManagementFactory.getRuntimeMXBean();
+	private static List<String> aList=RuntimemxBean.getAllInputArguments();
```
tajila: can you please use a more descriptive name for this variable

runtime/util/vmargs.c
AidanHa: We need it to check the substring since some arguments might have parameters (EG: -Djava.class.path=/User/...). If we did an exact comparison, we wouldn't get a match unless the user argument is the exact same as the one in the table

runtime/util/vmargs.c
AidanHa: Since we don't use these macros anywhere else in the code, may as well undef them here no? I can remove this if it doesn't make sense

test/functional/JIT_Test/src/jit/test/jitt/codecache/TestManager.java
AidanHa: It was previously called `alist` in the rest of the code. I feel like keeping the variable name the same would be better in terms of clarity.

runtime/util/vmargs.c
tajila: > Since we don't use these macros anywhere else in the code

This is true, but we typically don't do this. There are a lot of macros in the codebase that we only use once.

runtime/vm/jvminit.c
AidanHa: I get an error when I remove the brackets around `((j9vm_args->j9Options[i].flags) & (USER_ARG)))`, so atm, I'll just remove the brackets around `NOT_CONSUMABLE_ARG`

tajila: what kind of error?

AidanHa: it didn't like how there was a `|` and a `&` in the same expression, and asked me to brackets for the & case

So it ended up looking like: 

```uintptr_t argFlag = NOT_CONSUMABLE_ARG | (j9vm_args->j9Options[i].flags & (USER_ARG));	```

File: runtime/util/vmargs.c (hunk 29 above)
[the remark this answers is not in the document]
Comment (tajila) [reply runtime/util/vmargs.c]: looks  like this function is only called by `addEnvironmentVariables` which is only used in jvm.c. 

If there is only one caller, and the possible args for `envVarName` are known, then a comparision that take into account the length would be more optimal

File: runtime/util_core/vmargs_core.c [hunk 34]
```diff
@@ -277,8 +278,10 @@ addXOptionsFile(J9PortLibrary* portLib, const char *xOptionsfileArg, J9JavaVMArg
 		return -1;
 	}
 
-	/* add the -Xoptionsfile argument */
-	optArg = newJavaVMArgInfo(vmArgumentsList, NULL, ARG_MEMORY_ALLOCATION|CONSUMABLE_ARG);
+	/* add the -Xoptionsfile argument. Sets the USER_ARG flag if the optionfile is not the default optionfile */
+	optArg = newJavaVMArgInfo(vmArgumentsList, NULL, ARG_MEMORY_ALLOCATION | ((containsUserArgs ? USER_ARG : 0) | CONSUMABLE_ARG));
```
tajila: nitpick: extra bracket not needed. i.e. `ARG_MEMORY_ALLOCATION | (containsUserArgs ? USER_ARG : 0) | CONSUMABLE_ARG`

AidanHa:
Up for another review!

runtime/util/vmargs.c
```diff
@@ -1355,8 +1388,13 @@ addEnvironmentVariableArguments(J9PortLibrary* portLib, const char* envVarName,
 	}
 	JVMINIT_VERBOSE_INIT_TRACE1(verboseFlags, "Parsing environment variable %s\n", envVarName);
 	j9sysinfo_get_env(envVarName, argumentBuffer, envVarSize);
-
-	return parseOptionsBuffer(portLib, argumentBuffer, vmArgumentsList, verboseFlags, TRUE);
+	UDATA envVarNameLength = LITERAL_STRLEN(envVarName);
+	if ((J9UTF8_DATA_EQUALS(ENVVAR_JAVA_TOOL_OPTIONS, LITERAL_STRLEN(ENVVAR_JAVA_TOOL_OPTIONS), envVarName, envVarNameLength)) || (J9UTF8_DATA_EQUALS(ENVVAR_OPENJ9_JAVA_OPTIONS, LITERAL_STRLEN(ENVVAR_OPENJ9_JAVA_OPTIONS), envVarName, envVarNameLength)) || (J9UTF8_DATA_EQUALS(ENVVAR_IBM_JAVA_OPTIONS, LITERAL_STRLEN(ENVVAR_IBM_JAVA_OPTIONS), envVarName, envVarNameLength))) {
```
tajila: Nitpick: when you have a large conditional statement like this put the `||` in a new line 

ie.

```
if (condition1
     || condition2
     ...
     || conditionN
) {
```

genie-openj9:

PR #4826, opened 2019-02-21
Implement TR_J9JITaaSServerSharedCache

dmitry-ten:
- Initialize TR_J9JITaaSServerSharedCache on the server when
    `TR_EnableJITaaSRemoteAOT` environment variable is set.
    It will be used to make remote calls/cache information to/from
    shared class cache on the client.

 - Implement most methods that are used by shared cache on the server:
    pointerFromOffsetInSharedCache
    offsetInSharedCacheFromPointer
    persistIprofileInfo (dummy implementation, does nothing)
    rememberClass
    getClassChainOffsetOfIdentifyingLoaderForClazzInSharedCache

- Enable AOT-specific relocations that require shared cache access

### DIFF
--- a/runtime/compiler/env/J9SharedCache.cpp
+++ b/runtime/compiler/env/J9SharedCache.cpp
@@ -35,6 +35,8 @@
 #include "env/j9method.h"
 #include "runtime/IProfiler.hpp"
 #include "env/ClassLoaderTable.hpp"
+#include "control/CompilationThread.hpp" // for TR::compInfoPT
+#include "control/JITaaSCompilationThread.hpp" // for ClientSessionData
 
 #define   LOG(n,c) \
    if (_logLevel >= (3*n)) \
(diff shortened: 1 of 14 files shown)


dmitry-ten:
This PR contains commit merged into master recently, so it should be merged after the next rebase.

dmitry-ten:
Ready for review, but one thing that could definitely be improved is constructor.
I had to modify constructor of `TR_J9SharedCache`, because it assumes that SCC exists on the server, which is not the case for remote AOT.
Not sure what would be a better thing to do instead.

runtime/compiler/x/codegen/J9AheadOfTimeCompile.cpp
```diff
@@ -199,7 +199,7 @@ uint8_t *J9::X86::AheadOfTimeCompile::initializeAOTRelocationHeader(TR::Iterated
          cursor += SIZEPOINTER;
 
          TR_OpaqueClassBlock *j9class = (TR_OpaqueClassBlock *) aconstNode->getAddress();
-         uintptrj_t classChainOffsetInSharedCache = sharedCache->lookupClassChainOffsetInSharedCacheFromClass(j9class);
+         uintptrj_t classChainOffsetInSharedCache = sharedCache->getClassChainOffsetOfIdentifyingLoaderForClazzInSharedCache(j9class);
```
mpirvu: Why the change in name from `lookupClassChainOffsetInSharedCacheFromClass` to `getClassChainOffsetOfIdentifyingLoaderForClazzInSharedCache`

runtime/compiler/env/J9SharedCache.cpp
```diff
@@ -797,3 +771,47 @@ TR_J9SharedCache::getClassChainOffsetOfIdentifyingLoaderForClazzInSharedCache(TR
    uintptrj_t classChainOffsetInSharedCache = (uintptrj_t) offsetInSharedCacheFromPointer(classChainIdentifyingLoaderForClazz);
    return classChainOffsetInSharedCache;
    }
+
+void *
+TR_J9JITaaSServerSharedCache::pointerFromOffsetInSharedCache(void *offset)
+   {
```
mpirvu: Maybe we should store the stream into `TR_J9JITaaSServerSharedCache` object so that we don't have to call a TLS routine every time.

runtime/compiler/x/codegen/J9AheadOfTimeCompile.cpp
dmitry-ten: `lookupClassChainOffsetInSharedCacheFromClass` was not present in mainline.
When I added it, `getClassChainOffsetOfIdentifyingLoaderForClazzInSharedCache` is the name Irwin and I agreed on.

runtime/compiler/env/J9SharedCache.cpp
dmitry-ten: Will do

dmitry-ten:
Done

PR #4829, opened 2019-02-21
Tgc root scan for Metronome

amicic:
Miscellaneous improvements for TGC root scan times for Metronome:

1) add suspend/resume pair (for tracking increment durations per thread)
to the slave wait point in getPacket (the logic is identical to the wait
point in synchronization barriers, which already have suspent/resume)

2) suspend/resume pair in yield point is brought down to more specific
points for master and slave points separately. This is to avoid
reporting a false increment end when we decide to for example double
beat.

3) report 'unknown' root entity. These are typically synchronization
points not residing with existing known root entities, but somewhere
between them. Or anything else that we miss to reported as a known
entity.

4) report start/end scanWeakReferencesComplete to avoid reporting it as
unknown (an example of one that we miss to properly report).

### DIFF
--- a/runtime/gc_base/RootScanner.hpp
+++ b/runtime/gc_base/RootScanner.hpp
@@ -201,15 +201,15 @@ protected:
 		}
 	}
 	
-	MMINLINE void updateScanStats(uint64_t endTime) {
-	
+	MMINLINE void updateScanStats(uint64_t endTime)
+	{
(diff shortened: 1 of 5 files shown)


dmitripivkine:
Jenkins test sanity jdk8

dmitripivkine:

dmitripivkine:
Jenkins compile win

PR #4831, opened 2019-02-22
Temporarily blacklist OMRs SwitchAnalyzer.cpp

charliegracie:
The SwitchAnalyzer optimization is being moved from OpenJ9 to OMR.
While it is being moved it will be in both repos at the same time.
For now blacklist the OMR one in OpenJ9 so that it does not find
duplicate symbols. Once a newer OMR has been accepted into OpenJ9
the OMR copy should be used and the OpenJ9 copy should be deleted.


DanHeidinga:

DanHeidinga:
@AdamBrousseau indicated that we can't run the cmake build in PRs yet.  And the Travis build won't include the OMR conflict.

Will need to merge this as is.

DanHeidinga:
I've cancelled the PR build

DanHeidinga:
Based on discussion on Slack, we'll wait for the travis build (provided it launches in a reasonable time frame)

PR #4836, opened 2019-02-22
Rename JVM_GetClassName

keithc-ca:
* name changed to JVM_InitClassName for Java 11 and Java 13+

### DIFF
--- a/runtime/j9vm/j9vmnatives.xml
+++ b/runtime/j9vm/j9vmnatives.xml
@@ -75,7 +75,10 @@ SPDX-License-Identifier: EPL-2.0 OR Apache-2.0 OR GPL-2.0 WITH Classpath-excepti
 		<export name="_JVM_GetClassConstantPool@8" />
 		<export name="_JVM_GetClassContext@4" />
 		<export name="_JVM_GetClassLoader@8" />
-		<export name="_JVM_GetClassName@8" />
+		<export name="_JVM_GetClassName@8">
+			<include-if condition="spec.java12 or not spec.java11"/>
+			<exclude-if condition="spec.java13"/>
+		</export>
 		<export name="_JVM_GetClassSignature@8" />
 		<export name="_JVM_GetEnclosingMethodInfo@8" />
 		<export name="_JVM_GetInterfaceVersion@0" />
@@ -329,6 +332,9 @@ SPDX-License-Identifier: EPL-2.0 OR Apache-2.0 OR GPL-2.0 WITH Classpath-excepti
 		<export name="JVM_GetNestHost"/>
 		<export name="JVM_GetNestMembers"/>
 		<export name="JVM_AreNestMates"/>
+		<export name="JVM_InitClassName">
+			<exclude-if condition="spec.java12 and not spec.java13"/>
+		</export>
 	</exports>
 
 	<exports group="jdk12">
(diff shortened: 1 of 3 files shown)


runtime/j9vm/j9vmnatives.xml
```diff
@@ -329,6 +332,9 @@ SPDX-License-Identifier: EPL-2.0 OR Apache-2.0 OR GPL-2.0 WITH Classpath-excepti
 		<export name="JVM_GetNestHost"/>
 		<export name="JVM_GetNestMembers"/>
 		<export name="JVM_AreNestMates"/>
+		<export name="JVM_InitClassName">
+			<exclude-if condition="spec.java12 and not spec.java13"/>
```
pshipton: This doesn't seem right, it should be excluded for java8 and java12

runtime/j9vm/jvm.c
```diff
@@ -5491,10 +5491,18 @@ JVM_ActiveProcessorCount(void)
...
+ *
+ * The name was changed from JVM_GetClassName to JVM_InitClassName
+ * in interface version 6.
  */
 jstring JNICALL
-JVM_GetClassName(JNIEnv *env, jclass theClass)
+#if (JAVA_SPEC_VERSION < 11) || (JAVA_SPEC_VERSION == 12)
+JVM_GetClassName
+#else /* (JAVA_SPEC_VERSION < 11) || (JAVA_SPEC_VERSION == 12) */
+JVM_InitClassName
```
pshipton: How do we know the implementation for JVM_InitClassName is the same as for JVM_GetClassName?

runtime/j9vm/j9vmnatives.xml
```diff
@@ -75,7 +75,10 @@ SPDX-License-Identifier: EPL-2.0 OR Apache-2.0 OR GPL-2.0 WITH Classpath-excepti
 		<export name="_JVM_GetClassConstantPool@8" />
 		<export name="_JVM_GetClassContext@4" />
 		<export name="_JVM_GetClassLoader@8" />
-		<export name="_JVM_GetClassName@8" />
+		<export name="_JVM_GetClassName@8">
+			<include-if condition="spec.java12 or not spec.java11"/>
```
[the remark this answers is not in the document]
keithc-ca: No, `spec.java12` implies `JAVA_SPEC_VERSION >= 12`.

runtime/j9vm/j9vmnatives.xml
keithc-ca: This group, `jdk11`, is only used for Java 11+ (which excludes it for Java 8).

See my comment above: The condition excludes the export for Java 12+, but not Java 13+ (which means only for Java 12).

runtime/j9vm/jvm.c
keithc-ca: Class.java seems to use the methods the same way: I think openjdk just changed the name not its behavior.

pshipton: Note that since this isn't used, a stub that asserts would be fine.

keithc-ca:
Updated `JVM_InitClassName` to save the result in the `name` field.

pshipton:
jenkins test sanity win,plinux jdk8,jdk11,jdk12

PR #4844, opened 2019-02-22
Remove the unneeded if statement in ObjectMonitor.cpp::spinOnFlatLock

babsingh:
**Referring to [runtime/vm/vmthread.c#L516-L521](https://github.com/eclipse/openj9/blob/master/runtime/vm/vmthread.c#L516-L521):**
> In `ObjectMonitor.cpp:objectMonitorEnterNonBlocking`, we converted "goto
statements" into "three nested for loops". Due to this change, we can no
longer let spin counts be set to 0. With spin counts set to zero, there will
be no attempts to acquire a lock. To allow atleast one attempt to
acquire a lock, minimum value of spin counts is set to 1.

So, `spinCount1`/`thrMaxSpins1BeforeBlocking` won't be set to 0. Thus, the
unneeded if statement, `(0 != spinCount1)`, is removed.

### DIFF
--- a/runtime/vm/ObjectMonitor.cpp
+++ b/runtime/vm/ObjectMonitor.cpp
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2001, 2018 IBM Corp. and others
+ * Copyright (c) 2001, 2019 IBM Corp. and others
  *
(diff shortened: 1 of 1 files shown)


gacholio:

Issue #4751, opened 2019-02-15
Test-sanity.functional-JDK11-aix_ppc-64_cmprssptrs cmdLineTester_jvmtitests_hcr_4 crash vmState=0x00050080


pshipton:
omr acceptance build, but unrelated to the omr changes.
Note the test didn't fail, but I don't think it is supposed to crash.
https://ci.eclipse.org/openj9/job/Test-sanity.functional-JDK11-aix_ppc-64_cmprssptrs/478
cmdLineTester_jvmtitests_hcr_4
```
Testing: rc017
Test start time: 2019/02/15 08:46:26 Central Standard Time
Running command: "/home/u0020236/workspace/Test-sanity.functional-JDK11-aix_ppc-64_cmprssptrs/openjdkbinary/j2sdk-image/bin/java" -Xcompressedrefs -Xjit -Xgcpolicy:gencon -Xshareclasses:name=bcienabled,enableBCI  -Xdump    -agentlib:jvmtitest=test:rc017 -cp "/home/u0020236/workspace/Test-sanity.functional-JDK11-aix_ppc-64_cmprssptrs/openjdk-tests/TestConfig/scripts/testKitGen/../../../../jvmtest/functional/cmdLineTests/jvmtitests/jvmtitest.jar:/home/u0020236/workspace/Test-sanity.functional-JDK11-aix_ppc-64_cmprssptrs/openjdk-tests/TestConfig/scripts/testKitGen/../../../../jvmtest/TestConfig/lib/asm-all.jar" com.ibm.jvmti.tests.util.TestRunner
Time spent starting: 10 milliseconds
+0x358
masterSynchSignalHandler+0x60c
+0x0
Unhandled exception
Type=Segmentation error vmState=0x00050080
J9Generic_Signal_Number=00000004 Signal_Number=0000000b Error_Value=00000000 Signal_Code=00000032
Handler1=09001000B7612808 Handler2=09001000B75E9538
R0=0000000000000000 R1=00000001209A2A30 R2=08001000A01CDD48 R3=00000001209A2B40
R4=0000000000000000 R5=0000000000058F10 R6=E8A10078E86304B0 R7=000000000026259F
R8=E8A10078E86304B4 R9=00000000128500C1 R10=8000000000001032 R11=0000000000000000
R12=090000031FDDC3E8 R13=00000001209B8800 R14=00000001209A65F0 R15=00000001209A64C8
[...]

PR #4864, opened 2019-02-25
Fix crash in AOT during field iteration

aviansie-ben:
The code in the constructor for TR_VMFieldsInfo was making the
assumption that the superclass for a class can always be found. While
this is always true when JIT compiling a method, it is possible under
certain rare circumstances for this to fail during an AOT compilation.
For instance, this can fail under SVM if the SCC runs out of space
while trying to store the class chain for the parent class.

When an unexpected failure was encountered, this would then cause the
JVM to crash. This has now been fixed so that under AOT this will cause
the compilation to fail instead of crashing the JVM.

Fixes: #4751

### DIFF
--- a/runtime/compiler/env/j9fieldsInfo.cpp
+++ b/runtime/compiler/env/j9fieldsInfo.cpp
@@ -154,7 +154,17 @@ TR_VMFieldsInfo::TR_VMFieldsInfo(TR::Compilation * comp, J9Class *aClazz, int bu
    for (i=numSupClasses-1; i>=0; i--)
       {
       supClass = (J9Class*)comp->fej9()->getSuperClass((TR_OpaqueClassBlock*)supClass);
-      TR_ASSERT(supClass, "Found NULL supClass in inheritance chain");
+
+      if (comp->compileRelocatableCode())
+         {
+         if (!supClass)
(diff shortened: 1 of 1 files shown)


dsouzai:
jenkins test sanity xlinux,plinux,aix jdk11

dsouzai:
@pshipton according to @gita-omr 's update, this should be in the 0.13.0 release; can this change be cherry picked, or does @aviansie-ben need to open another PR to deliver into the 0.13.0 release branch?

gita-omr:
As discussed with @aviansie-ben I just wanted some assurance that this is a unique place affected by the issue and we check !superClass in all other places.

pshipton:
@dsouzai This will have to be merged first. We might be able to push it to the branch if nothing else we don't want has been merged. Otherwise you do have to cherry pick it and create another PR for the 0.13 branch.

dsouzai:
@aviansie-ben once this is merged, could you coordinate with @pshipton to see if you need to create another PR.

aviansie-ben:
@gita-omr I've taken a look and haven't found any other uses of `getSuperClass` that stand out to me as having problems if it returns `NULL`.

gita-omr:
Excellent, thanks!

pshipton:
Tthere was nothing else blocking it, so I've pushed this to the 0.13 branch.

Issue #4465, opened 2019-01-26
windows10&openj9-jdk8??JVM_EnqueueOperation() stubbed


iver3on:
![image](https://user-images.githubusercontent.com/12184988/51782060-4da75d80-215d-11e9-9449-0e424c3c5d07.png)
what's wrong???!!!

pshipton:
This error means that an OpenJDK+Hotspot / Oracle JVM is attempting to use Attach API to connect to an OpenJ9 VM. This is not supported.

pshipton:
We could update the assertion error message with a description of the problem, rather than "JVM_EnqueueOperation() stubbed!".

DanHeidinga:
@sharon-wang is looking at updating the message

DanHeidinga:
This didn't make the `0.13.0` milestone.  Retargeting to `0.14.0`

Issue #3441, opened 2018-10-25
JVM_EnqueueOperation() stubbed


planetf1:
When trying to build github.com/odpi/egeria with 'mvn clean install' using maven 3.5.4 on windows 10, within IntelliJ IDEA, I get:

Microsoft Visual C++ Runtime Library
X Assertion failed!

Program: c:\java\jdku192-b12\bin\java.exe
j7vmi.c
Line: 2713

Expression: !"JVM_EnqueueOperation() stubbed!"

The JDK was downloaded from adoptopenjdk.net

I can ignore the popup, and the process does continue

A similar issue occurs on MacOS, but the exception is different:

DanHeidinga:
@ChengJin01 Can you look at this?

ChengJin01:
Will investigate to see what happened in there.

ChengJin01:
I don't think the issue here has anything to do with #3442 as JVM_SocketShutdown() was already implemented on UNIX-based system (So it just got enabled for OSX).

JVM_EnqueueOperation() was not implemented on any platform. So I need to figure out how this got invoked on Windows first.

pshipton:
It seems something is trying to use the OpenJDK attach mechanism instead of the OpenJ9 one. i.e. it must be hard coded to use sun.tools.attach.WindowsVirtualMachine (or similar concrete class) instead of using the generic classes as it should.

ChengJin01:
I already noticed that at  openj9-openjdk-jdk8/jdk/src/windows/native/sun/tools/attach/WindowsVirtualMachine.c.  
```
/* exported function in target VM */
typedef jint (WINAPI* EnqueueOperationFunc)
    (const char* cmd, const char* arg1, const char* arg2, const char* arg3, const char* pipename);
```

 If so, it looks like there are some setting to be modified so as to avoid sun.tools.attach.WindowsVirtualMachine in favor of our build.

pshipton:
> it looks like there are some setting to be modified so as to avoid sun.tools.attach.WindowsVirtualMachine in favor of our build.

Note the OpenJ9 build is configured correctly, I think this problem would only occur if something is hard coded to refer to the Hotspot attach API implementation rather than the generic api classes https://docs.oracle.com/javase/8/docs/jdk/api/attach/spec/index.html

ChengJin01:
> I think this problem would only occur if something is hard coded to refer to the Hotspot attach API implementatation...

[...]

PR #4871, opened 2019-02-25
Update assertion error message for JVM_EnqueueOperation()

sharon-wang:
Fixes: #4465. 
Related: #3441. 

The existing assertion error message does not indicate the likely cause of the error.
The cause of the error is likely due to the use of an OpenJDK+Hotspot/Oracle VM Attach API to connect to an OpenJ9 VM, which is not supported.

Suggestions on the wording of the message are welcomed!

### DIFF
--- a/runtime/j9vm/j7vmi.c
+++ b/runtime/j9vm/j7vmi.c
@@ -2709,7 +2709,7 @@ JVM_DefineClassWithSourceCond(jint arg0, jint arg1, jint arg2, jint arg3, jint a
 jobject JNICALL
 JVM_EnqueueOperation(jint arg0, jint arg1, jint arg2, jint arg3, jint arg4)
 {
-	assert(!"JVM_EnqueueOperation() stubbed!");
+	assert(!"A HotSpot VM Attach API is attempting to connect to an OpenJ9 VM. This is not supported.");
 	return NULL;
 }
 


runtime/j9vm/j7vmi.c
```diff
@@ -2709,7 +2709,7 @@ JVM_DefineClassWithSourceCond(jint arg0, jint arg1, jint arg2, jint arg3, jint a
 jobject JNICALL
 JVM_EnqueueOperation(jint arg0, jint arg1, jint arg2, jint arg3, jint arg4)
 {
-	assert(!"JVM_EnqueueOperation() stubbed!");
+	assert(!"An OpenJDK Attach API may have been used to connect to an OpenJ9 VM. This is not supported.");
```
pshipton: I think this needs to mention Hotspot, otherwise it is unclear what it means because both Hotspot and OpenJ9 use OpenJDK.

sharon-wang: Would it be more clear to say: "A HotSpot JVM Attach API is attempting to connect to an OpenJ9 VM. This is not supported."?

pshipton: Works for me. Although you can use "HotSpot VM" instead of "HotSpot JVM"

sharon-wang:
@pshipton Just confirming that this is what we want to see.

**Note:** the message will now say "A HotSpot VM Attach API is attempting to connect to an OpenJ9 VM. This is not supported."

image

pshipton:
Works for me. Maybe others will have an opinion as well if we wait a bit.
@DanHeidinga

DanHeidinga:
I'm OK with the wording of this as well.  The unique message + any stacktrace should make it clear which JVM_ function was called.

DanHeidinga:
Jenkins test sanity win jdk8

DanHeidinga:
This didn't make the `0.13.0` milestone.  Retargeting to `0.14.0`

Issue #2768, opened 2018-09-05
Prevent instrumentation of com.ibm.jit.JITHelpers


pshipton:
As per https://github.com/eclipse/openj9/issues/2175#issuecomment-397068794, https://github.com/eclipse/openj9/issues/2175#issuecomment-397069745 and https://github.com/eclipse/openj9/issues/2175#issuecomment-418807896 @gacholio wrote.

"JITHelpers should probably be in the list of non-instrumentable classes.

jvmtiHook.c : canClassBeInstrumented to prevent load-time modification.

hshelp.c : classIsModifiable to prevent redefinition

It seems sensible to prevent instrumentation (and we can probably come up with more classes that it would be disastrous to instrument).
"

pshipton:
@gacholio would you like to volunteer to make the changes, or leave this for somebody else?

gacholio:
This isn't critical, so let's gather the data before making any changes.

gacholio:
@andrewcraik @vijaysun-omr Are there other classes that will be an issue if instrumented? Not concerned about loss of performance, just with incorrectness/crash. Bear in mind the granularity of the decision is at the class level, not method. I was wondering in particular about MethodHandle implementation classes - I suspect instrumenting the thunk implementation methods would lead to badness.

gacholio:
@DanHeidinga - any classes you can think of that we should not instrument?

DanHeidinga:
VarHandleInternal for one.

vijaysun-omr:
@gacholio I think you are right about the thunk archetypes that they should not be modifiable since they are really "convenience" methods to make the job of ILgen of thunks easier. Arbitrary changes in those methods would break functionality.

I wonder if all the classes in com.ibm.jit package (such as com_ibm_jit_DecimalFormatHelper_formatAsDouble or the DAA/crypto classes) should not be allowed to instrument.

vijaysun-omr:
@jdmpapin could you please give your opinion too ?

vijaysun-omr:
com.ibm.Compiler.Internal.Quad might be another class that cannot be instrumented.

gacholio:
Once #4876 is merged, any bootstrap class can be annotated with `com.ibm.oti.vm.J9INTERNAL_CLASS` to prevent it from being modified during load, or redefined later.

gacholio:
Note the annotation is `NOT` inherited - every class that wants it must declare it.

PR #4876, opened 2019-02-26
Use a class annotation to prevent instrumentation

gacholio:
New annotation: com.ibm.oti.vm.J9UnmodifiableClass

Rather than using a hard-coded (and inconsistent!) list of classes which
are not allowed to be modified by the JVMTI ClassFileLoadHook or
Re[define/transform]Classes APIs, annotate the classes which are private
to the VM implementation.

For performance, only classes loaded by the bootstrap loader will be
scanned for the annotation before ClassFileLoadHook (and that scan is
optimistic - any class file containing the bytes representing a constant
pool entry UTF8 matching the annotation signature will be tagged). This
is safe because the VM is largely in control of the boot path (it's
beyond unlikely that this pattern could match accidentally in classes
added to the boot path by the application).

Any class with access to com.ibm.oti.vm may use the annotation to prevent being redefined. This prevents application classes from using the annotation.

This also fixes the fact that we were not reporting ClassFileLoadHook for any class that was not explicitly-named in the defineClass call.

Fixes: #2768

[ci skip]

### DIFF
--- a/debugtools/DDR_VM/src/com/ibm/j9ddr/vm29/tools/ddrinteractive/commands/J9BCUtil.java
+++ b/debugtools/DDR_VM/src/com/ibm/j9ddr/vm29/tools/ddrinteractive/commands/J9BCUtil.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2001, 2018 IBM Corp. and others
+ * Copyright (c) 2001, 2019 IBM Corp. and others
  *
  * This program and the accompanying materials are made available under
  * the terms of the Eclipse Public License 2.0 which accompanies this
@@ -556,7 +556,7 @@ public class J9BCUtil {
 
 		out.append(String.format("Sun Access Flags (0x%s): ", Long.toHexString(romClass.modifiers().longValue())));
 		dumpModifiers(out, romClass.modifiers().longValue(), MODIFIERSOURCE_CLASS, ONLY_SPEC_MODIFIERS);
-		
+		out.append(nl);
 		out.append(String.format("J9  Access Flags (0x%s): ", Long.toHexString(romClass.extraModifiers().longValue())));
 		dumpClassJ9ExtraModifiers(out, romClass.extraModifiers().longValue());
 		out.append(nl);
@@ -708,6 +708,8 @@ public class J9BCUtil {
 			out.append("(preverified) ");
 		if ((accessFlags & J9JavaAccessFlags.J9AccClassAnonClass) != 0)
 			out.append("(anonClass) ");
+		if ((accessFlags & J9JavaAccessFlags.J9AccClassIsUnmodifiable) != 0)
+			out.append("(unmodifiable) ");
 	}
 
 	private static void dumpEnclosingMethod(PrintStream out, J9ROMClassPointer romClass, long flags) throws CorruptDataException {
--- a/runtime/jvmti/jvmtiHook.c
+++ b/runtime/jvmti/jvmtiHook.c
@@ -2231,26 +2231,59 @@ jvmtiHookGCCycleEnd(J9HookInterface** hook, UDATA eventNum, void* eventData, voi
 }
 
 /**
- * Determines if the class described by <tt>classLoadData</tt> can 
+ * Determines if the class described by <tt>data</tt> can 
  * be instrumented.
- * @param classLoadData 
+ * @param data The class file load hook event data
  * @return TRUE if the class can be instrumented. 
  */
 static BOOLEAN
-canClassBeInstrumented(J9VMClassLoadHookEvent* classLoadData)
+canClassBeInstrumented(J9VMClassLoadHookEvent* data)
 {
-	BOOLEAN result = TRUE;
-
-	/* Skip any class that does not have a name specified */
-	if (classLoadData->className == NULL) {
-		result = FALSE;
-	} else if (strcmp(classLoadData->className, "java/lang/J9VMInternals") == 0) {
-		result = FALSE;
-	} else if (strcmp(classLoadData->className, "java/lang/invoke/VarHandleInternal") == 0) {
-		result = FALSE;
+	BOOLEAN modifiable = TRUE;
+	/* Only clasess on the bootpath may be marked unmodifiable */
+	if (data->classLoader == data->currentThread->javaVM->systemClassLoader) {
+		U_8 *classData = data->classData;
+		UDATA const classDataLength = data->classDataLength;
+		UDATA const annotationLength = sizeof(J9_UNMODIFIABLE_CLASS_ANNOTATION_DATA);
+		/* Before the constant pool in the class file, there's at least:
+		 *
+		 *	u4 magic
+		 *	u2 minor_version
+		 *	u2 major_version
+		 *	u2 constant_pool_count
+		 *
+ 		 * Totalling 10 bytes.
+		 */
+		UDATA const minimumSize = 10;
+		if (classDataLength > (annotationLength + minimumSize)) {
+			/* Make sure this at least looks like a class file */
+			if ((0xCA == classData[0]) && (0xFE == classData[1]) && (0xBA == classData[2]) && (0xBE == classData[3])) {
+				U_8 *dataEnd = classData + classDataLength - annotationLength;
+				classData += minimumSize;
+				/* Search for the bytes representing the UTF8 annotation name */
+				while (classData < dataEnd) {
+					static J9_UNMODIFIABLE_CLASS_ANNOTATION_DATA annotBytes = {
+						(char)1, /* tag byte for UTF8 in the constant pool */
+						(char)(LITERAL_STRLEN(J9_UNMODIFIABLE_CLASS_ANNOTATION) >> 8), /* high byte of size (big endian) */
+						(char)(LITERAL_STRLEN(J9_UNMODIFIABLE_CLASS_ANNOTATION) & 0xFF), /* low byte of size (big endian) */
+						J9_UNMODIFIABLE_CLASS_ANNOTATION /* UTF8 data */
+					};
+					UDATA scanSize = annotationLength;
+					U_8 *scanBytes = (U_8*)&annotBytes;
+					do {
+						if (*scanBytes++ != *classData++) {
+							goto notFound;
+						}
+						scanSize -= 1;
+					} while (0 != scanSize);
+					modifiable = FALSE;
+					break;
+notFound: ;
+				}
+			}
+		}
 	}
-
-	return result;
+	return modifiable;
 }
 
 
--- a/jcl/src/java.base/share/classes/com/ibm/jit/JITHelpers.java
+++ b/jcl/src/java.base/share/classes/com/ibm/jit/JITHelpers.java
@@ -3,7 +3,7 @@
 package com.ibm.jit;
 
 /*******************************************************************************
- * Copyright (c) 1998, 2018 IBM Corp. and others
+ * Copyright (c) 1998, 2019 IBM Corp. and others
  *
  * This program and the accompanying materials are made available under
  * the terms of the Eclipse Public License 2.0 which accompanies this
@@ -24,6 +24,7 @@ package com.ibm.jit;
  * SPDX-License-Identifier: EPL-2.0 OR Apache-2.0 OR GPL-2.0 WITH Classpath-exception-2.0 OR LicenseRef-GPL-2.0 WITH Assembly-exception
  *******************************************************************************/
 
+import com.ibm.oti.vm.J9UnmodifiableClass;
 import java.lang.reflect.Field;
 import java.lang.reflect.Array;
 import com.ibm.oti.vm.VM;
@@ -41,6 +42,7 @@ import sun.reflect.CallerSensitive;
  * The <code>JITHelpers</code> class contains methods used by the JIT to optimize certain primitive operations.
  */
 @SuppressWarnings("javadoc")
+@J9UnmodifiableClass
 public final class JITHelpers {
 
 	private static final JITHelpers helpers;
--- a/jcl/src/java.base/share/classes/com/ibm/oti/vm/J9UnmodifiableClass.java
+++ b/jcl/src/java.base/share/classes/com/ibm/oti/vm/J9UnmodifiableClass.java
@@ -1,0 +1,41 @@
+/*[INCLUDE-IF Sidecar17]*/
+/*******************************************************************************
+ * Copyright (c) 2019, 2019 IBM Corp. and others
+ *
+ * This program and the accompanying materials are made available under
+ * the terms of the Eclipse Public License 2.0 which accompanies this
+ * distribution and is available at https://www.eclipse.org/legal/epl-2.0/
+ * or the Apache License, Version 2.0 which accompanies this distribution and
+ * is available at https://www.apache.org/licenses/LICENSE-2.0.
+ *
+ * This Source Code may also be made available under the following
+ * Secondary Licenses when the conditions for such availability set
+ * forth in the Eclipse Public License, v. 2.0 are satisfied: GNU
+ * General Public License, version 2 with the GNU Classpath
+ * Exception [1] and GNU General Public License, version 2 with the
+ * OpenJDK Assembly Exception [2].
+ *
+ * [1] https://www.gnu.org/software/classpath/license.html
+ * [2] http://openjdk.java.net/legal/assembly-exception.html
+ *
+ * SPDX-License-Identifier: EPL-2.0 OR Apache-2.0 OR GPL-2.0 WITH Classpath-exception-2.0 OR LicenseRef-GPL-2.0 WITH Assembly-exception
+ *******************************************************************************/
+
+package com.ibm.oti.vm;
+
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+@Retention(RetentionPolicy.RUNTIME)
+@Target({ElementType.TYPE})
+
+// WARNING: Do not rename or repackage this class without changing
+// J9_UNMODIFIABLE_CLASS_ANNOTATION in j9.h to match. Under no
+// circumstances include any character whose ASCII value is
+// less than that of '/' (see canClassBeInstrumented in jvmtiHook.c).
+
+public @interface J9UnmodifiableClass {
+}
+
--- a/jcl/src/java.base/share/classes/java/lang/J9VMInternals.java
+++ b/jcl/src/java.base/share/classes/java/lang/J9VMInternals.java
@@ -2,6 +2,7 @@
 
 package java.lang;
 
+import com.ibm.oti.vm.J9UnmodifiableClass;
 import java.lang.ref.SoftReference;
 import java.lang.reflect.*;
 import java.security.AccessController;
@@ -30,7 +31,7 @@ import com.ibm.oti.util.Msg;
 
 
 /*******************************************************************************
- * Copyright (c) 1998, 2018 IBM Corp. and others
+ * Copyright (c) 1998, 2019 IBM Corp. and others
  *
  * This program and the accompanying materials are made available under
  * the terms of the Eclipse Public License 2.0 which accompanies this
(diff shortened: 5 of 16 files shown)


gacholio:
jenkins test all xlinux jdk8

jcl/src/java.base/share/classes/com/ibm/oti/vm/J9INTERNAL_CLASS.java
```diff
@@ -0,0 +1,35 @@
...
+package com.ibm.oti.vm;
+
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+@Retention(RetentionPolicy.RUNTIME)
+@Target({ElementType.TYPE})
+public @interface J9INTERNAL_CLASS {
```
fjeremic: Odd naming convention for an annotation. Shouldn't we follow class/annotation naming conventions here (`J9InternalClass`)?

fjeremic: HotSpot has a similar `@HotSpotIntrinsicCandidate` as an example.

gacholio: It's name similarly to our other internal annoatation:

https://github.com/eclipse/openj9/blob/278be6c0ff617069c6fa2038ffac7afbf4cb5989/jcl/src/java.base/share/classes/java/lang/invoke/VMCONSTANTPOOL_CLASS.java

I'm certainly not married to the name, and we should be moving these things out of com.ibm at some point.

gacholio: Also, I want something fairly obtuse because I'm cheating the annotation lookup in the ClassFileLoadHook code.

gacholio:
jenkins test all xlinux jdk8

jcl/src/java.base/share/classes/com/ibm/oti/vm/J9INTERNAL_CLASS.java
gacholio: @DanHeidinga can comment on this during his review.

runtime/jvmti/jvmtiHook.c
```diff
@@ -2231,26 +2231,40 @@ jvmtiHookGCCycleEnd(J9HookInterface** hook, UDATA eventNum, void* eventData, voi
...
+	/* Only clasess on the bootpath may be marked unmodifiable */
+	if (data->classLoader == data->currentThread->javaVM->systemClassLoader) {
+		U_8 *classData = data->classData;
+		UDATA classDataLength =  data->classDataLength;
+		U_8 const annotationLength = LITERAL_STRLEN(J9_UNMODIFIABLE_CLASS_ANNOTATION);
+		if (classDataLength > (annotationLength + 4)) {
+			/* Make sure this at least looks like a class file */
+			if ((classData[0] = 0xCA) && (classData[1] = 0xFE) && (classData[2] = 0xBA) && (classData[3] = 0xBE)) {
+				U_8 *dataEnd = classData + classDataLength - annotationLength;
+				classData += 4;
```
DanHeidinga: This can skip ahead another 6 bytes {u2 major, u2 minor, u2 cp count}.

runtime/jvmti/jvmtiHook.c
```diff
@@ -2231,26 +2231,40 @@ jvmtiHookGCCycleEnd(J9HookInterface** hook, UDATA eventNum, void* eventData, voi
...
-		result = FALSE;
-	} else if (strcmp(classLoadData->className, "java/lang/invoke/VarHandleInternal") == 0) {
-		result = FALSE;
+canClassBeInstrumented(J9VMClassLoadHookEvent* data)
+{
+	BOOLEAN modifiable = TRUE;
+	/* Only clasess on the bootpath may be marked unmodifiable */
+	if (data->classLoader == data->currentThread->javaVM->systemClassLoader) {
+		U_8 *classData = data->classData;
+		UDATA classDataLength =  data->classDataLength;
```
DanHeidinga: one extra space after the = only worth mentioning because of other changes to this method.

runtime/jvmti/jvmtiHook.c
```diff
@@ -2231,26 +2231,40 @@ jvmtiHookGCCycleEnd(J9HookInterface** hook, UDATA eventNum, void* eventData, voi
...
+				U_8 *dataEnd = classData + classDataLength - annotationLength;
+				classData += 4;
+				while (classData < dataEnd) {
+					/* Look for u8 tag (1 for UTF8) followed by u16 string length */
+					if ((1 == classData[0]) && (0 == classData[1]) && (annotationLength == classData[2])) {
+						/* Now look for the annotation name */
+						if (0 == memcmp(classData + 3, J9_UNMODIFIABLE_CLASS_ANNOTATION, annotationLength)) {
+							modifiable = FALSE;
+							break;
+						}
```
DanHeidinga: I think an extra `classData += 1;` can be done if we made into the if block as the block is looking for `{ 1 0 length }`

jcl/src/java.base/share/classes/com/ibm/oti/vm/J9INTERNAL_CLASS.java
DanHeidinga: The VMCONSTANTPOOL naming convention is my mess.  I'm not super keen to propagate it.

I'd propose a name that more closely matches the intention and name of the internal bit - something like `J9UnmodifiableClass`?

gacholio:
> Have you done any startup performance tests with this change? I don't expect to see much of a difference unless someone has attached an agent

I have not done any measurements.

runtime/jvmti/jvmtiHook.c
gacholio: Given that the string is short (and will not contain any matching substrings), I could completely optimize this by not using memcmp, and updating classData with each character compared.

DanHeidinga:
> What package would you like for the annotation?

I've been trying to find the list of restricted packages and am failing miserably.  I expect `com.ibm.oti.vm.` to be on the restricted list but can't confirm that yet.

I'd like it to be in a restricted package if possible as it adds a minor barrier that helps prevent everyone from using the annotation.

jcl/src/java.base/share/classes/com/ibm/oti/vm/J9INTERNAL_CLASS.java
```diff
@@ -0,0 +1,35 @@
...
+
+package com.ibm.oti.vm;
+
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+@Retention(RetentionPolicy.RUNTIME)
+@Target({ElementType.TYPE})
```
DanHeidinga: Should probably add a comment to this class warning against renaming it or moving it without updating the native code.

runtime/jvmti/jvmtiHook.c
DanHeidinga: Sounds like a good experiment - perf test both versions and see which is actually faster.

jcl/src/java.base/share/classes/com/ibm/oti/vm/J9INTERNAL_CLASS.java
gacholio: Done.

runtime/jvmti/jvmtiHook.c
gacholio: Fixed.

jcl/src/java.base/share/classes/com/ibm/oti/vm/J9INTERNAL_CLASS.java
gacholio: Renamed.

runtime/jvmti/jvmtiHook.c
gacholio: Will roll this into the improved compare algorithm.

fjeremic:
@gacholio we may want to squash those commits as it looks like they have no body and only a sign-off title.

gacholio:
@fjeremic Note the WIP - I don't expect this to be merged. I always squash before commit.

fjeremic:
> @fjeremic Note the WIP - I don't expect this to be merged. I always squash before commit.

Ah good call! I had a case of the selective vision. Sorry for the churn, carry on!

gacholio:
The Liberty bench does not appear to use ClassFileLoadHook, so this is only measuring the cost of parsing the annotation in the ROM class builder:

ORIGINAL

StartupTime	avg= 980	min= 900	max=1125	stdDev=53.1	maxVar=25.0%	confInt=1.01%	samples= 80
Footprint	avg=59372	min=57128	max=63328	stdDev=1265.1	maxVar=10.9%	confInt=0.40%	samples= 80
CThreadTime	avg= 302	min= 217	max= 516	stdDev=78.7	maxVar=137.8%	confInt=4.91%	samples= 78
	Outlier values:  523 519
Stats for cold run:
StartupTime	avg=2313	min=2291	max=2362	stdDev=23.2	maxVar=3.1%	confInt=0.58%	samples= 10
Footprint	avg=62156	min=59632	max=66480	stdDev=2020.4	maxVar=11.5%	confInt=1.88%	samples= 10
CThreadTime	avg=2108	min=2076	max=2137	stdDev=21.4	maxVar=2.9%	confInt=0.59%	samples= 10


MY CODE

StartupTime	avg= 978	min= 897	max=1102	stdDev=44.0	maxVar=22.9%	confInt=0.85%	samples= 77
	Outlier values:  1148.81 1113.962 1115.252
Footprint	avg=59536	min=55636	max=64468	stdDev=1719.4	maxVar=15.9%	confInt=0.54%	samples= 80
CThreadTime	avg= 312	min= 237	max= 561	stdDev=88.9	maxVar=136.7%	confInt=5.29%	samples= 80
Stats for cold run:
StartupTime	avg=2309	min=2284	max=2354	stdDev=22.4	maxVar=3.1%	confInt=0.56%	samples= 10
Footprint	avg=61981	min=60328	max=63940	stdDev=1131.4	maxVar=6.0%	confInt=1.06%	samples= 10
CThreadTime	avg=2114	min=2102	max=2145	stdDev=13.6	maxVar=2.0%	confInt=0.37%	samples= 10

To me, this indicates no regression (@mpirvu can you confirm?)

I will test the ClassFileLoadHook using a microbench.

gacholio:
Measuring the load hook time in isolation is not practical (basic startup time is too variable to be useful). Instead, I've added an agent to Liberty startup which gets every ClassFileLoadHook (the new VM also detects the annotation at ROM class build time, as above, so this a measurement of the complete overhead of this change).

ORIGINAL

StartupTime	avg=1217	min=1127	max=1378	stdDev=46.2	maxVar=22.3%	confInt=0.71%	samples= 80
Footprint	avg=60398	min=56688	max=64468	stdDev=1812.3	maxVar=13.7%	confInt=0.56%	samples= 80
CThreadTime	avg= 323	min= 238	max= 548	stdDev=88.4	maxVar=130.3%	confInt=5.09%	samples= 80
Stats for cold run:
StartupTime	avg=2333	min=2286	max=2404	stdDev=34.6	maxVar=5.2%	confInt=0.86%	samples= 10
Footprint	avg=62901	min=60584	max=64368	stdDev=1138.8	maxVar=6.2%	confInt=1.05%	samples= 10
CThreadTime	avg=2116	min=2079	max=2145	stdDev=21.6	maxVar=3.2%	confInt=0.59%	samples= 10


MY CODE

StartupTime	avg=1232	min=1165	max=1364	stdDev=47.8	maxVar=17.1%	confInt=0.72%	samples= 80
Footprint	avg=60210	min=56368	max=63352	stdDev=1591.0	maxVar=12.4%	confInt=0.49%	samples= 80
CThreadTime	avg= 321	min= 222	max= 566	stdDev=92.6	maxVar=155.0%	confInt=5.36%	samples= 80
Stats for cold run:
StartupTime	avg=2323	min=2276	max=2356	stdDev=25.3	maxVar=3.5%	confInt=0.63%	samples= 10
Footprint	avg=63182	min=61420	max=65432	stdDev=1188.8	maxVar=6.5%	confInt=1.09%	samples= 10
CThreadTime	avg=2099	min=2065	max=2129	stdDev=18.7	maxVar=3.1%	confInt=0.52%	samples= 10

So the overhead of my check is minimal.

gacholio:
It's also worth noting that I'll soon be pushing a much-improved version of the load hook scan, so the overhead should be even less.

runtime/jvmti/jvmtiHook.c
gacholio: See below.

gacholio:
Ready for re-review.

gacholio:
I've manually tested this everywhere, launching one here for the record.

gacholio:
More on the bench scores. With load hook enabled on the unmodified build, the startup time is 1.24x.

With both of my changes active, the startup time is 1.26x.

mpirvu:
Regarding experiments in https://github.com/eclipse/openj9/pull/4876#issuecomment-470740549 there is a 1.2% regression with 0.7% confidence interval.

gacholio:
I'm willing to eat that loss for this case - the slowdown from using ClassFileLloadHook is large enough that I don't believe the extra overhead matters.

gacholio:
Reformatted the struct initializer to include the comments inline.

gacholio:
> Regarding experiments in [#4876 (comment)](https://github.com/eclipse/openj9/pull/4876#issuecomment-470740549) there is a 1.2% regression with 0.7% confidence interval.

@DanHeidinga What do you think?

gacholio:

gacholio:

gacholio:
@pshipton Is there a way I can further filter the classes using the ProtectionDomain? I really only want to inspect classes that come from the vm.jar rather than all classes on the boot path. That would essentially eliminate the overhead.

gacholio:

pshipton:
> Is there a way I can further filter the classes

For Java 8, you can retrieve the class path entry and check if it is the vm.jar
For Java 11 and later, there is no distinction. All you can check automatically is if the class is from the java.base module.

pshipton:
Actually, there is no vm.jar in OpenJ9 builds. You'll have to check for the rt.jar

gacholio:

gacholio:
New benchmark results from the latest code (and run on a different machine than before, so the scores are not comparable).

ORIGINAL (load hook NOT enabled)

StartupTime	avg=1234	min=1100	max=1645	stdDev=122.8	maxVar=49.5%	confInt=1.85%	samples= 80
Footprint	avg=61140	min=58052	max=65296	stdDev=1296.7	maxVar=12.5%	confInt=0.39%	samples= 80
CThreadTime	avg= 469	min= 306	max= 959	stdDev=137.2	maxVar=213.4%	confInt=5.44%	samples= 80
Stats for cold run:
StartupTime	avg=2848	min=2797	max=2954	stdDev=45.9	maxVar=5.6%	confInt=1.00%	samples=  9
	Outlier values:  3102.627
Footprint	avg=63906	min=61748	max=66940	stdDev=1682.5	maxVar=8.4%	confInt=1.53%	samples= 10
CThreadTime	avg=2738	min=2691	max=2825	stdDev=37.7	maxVar=5.0%	confInt=0.80%	samples= 10

ORIGINAL (load hook enabled)

StartupTime	avg=1496	min=1401	max=1729	stdDev=74.4	maxVar=23.4%	confInt=0.94%	samples= 78
	Outlier values:  2189.835 2176.52
Footprint	avg=62997	min=59924	max=66988	stdDev=1485.0	maxVar=11.8%	confInt=0.44%	samples= 80
CThreadTime	avg= 493	min= 328	max= 837	stdDev=137.4	maxVar=155.2%	confInt=5.18%	samples= 80
Stats for cold run:
StartupTime	avg=2880	min=2813	max=3055	stdDev=70.9	maxVar=8.6%	confInt=1.53%	samples=  9
	Outlier values:  3169.132
Footprint	avg=64390	min=62104	max=65360	stdDev=1063.4	maxVar=5.2%	confInt=1.02%	samples=  9
	Outlier values:  68392
CThreadTime	avg=2840	min=2679	max=3042	stdDev=140.1	maxVar=13.5%	confInt=2.86%	samples= 10

MY CODE (load hook enabled)

StartupTime	avg=1502	min=1396	max=1774	stdDev=86.2	maxVar=27.1%	confInt=1.10%	samples= 76
	Outlier values:  2069.799 2019.756 1837.436 1901.916
Footprint	avg=62596	min=59504	max=66544	stdDev=1776.0	maxVar=11.8%	confInt=0.53%	samples= 80
CThreadTime	avg= 460	min= 321	max= 923	stdDev=126.2	maxVar=187.5%	confInt=5.28%	samples= 75
	Outlier values:  1025 979 1035 952 993
Stats for cold run:
StartupTime	avg=2963	min=2823	max=3210	stdDev=134.6	maxVar=13.7%	confInt=2.63%	samples= 10
Footprint	avg=64821	min=62524	max=66872	stdDev=1476.6	maxVar=7.0%	confInt=1.32%	samples= 10
CThreadTime	avg=2826	min=2661	max=3076	stdDev=151.7	maxVar=15.6%	confInt=3.11%	samples= 10

Looks like the new code has virtually no overhead.

DanHeidinga:
@mpirvu Any comments on the results in https://github.com/eclipse/openj9/pull/4876#issuecomment-471012382  ?  Otherwise, I'll merge this Monday

gacholio:
I see a typo in one of the static asserts which I'm fixing now.

gacholio:
...and I had the C++ ifdef wrong (missing underscore), so we'll see if this compiles now.

gacholio:

gacholio:
jenkins compile win jdk8

gacholio:

gacholio:

gacholio:
PPC build is failing due to what looks like network issues - please ignore.

gacholio:
Last trivial fix incoming - missing comment on C++ endif.

gacholio:
One last build for the record...

jenkins test all xlinux jdk8

gacholio:
Ready for merge.

mpirvu:
> @mpirvu Any comments on the results in #4876 (comment) ? Otherwise, I'll merge this Monday

@DanHeidinga The results look neutral from startup point of view, so I am ok with the change.

debugtools/DDR_VM/src/com/ibm/j9ddr/vm29/tools/ddrinteractive/commands/J9BCUtil.java
```diff
@@ -708,6 +708,8 @@ private static void dumpClassJ9ExtraModifiers(PrintStream out, long accessFlags)
 			out.append("(preverified) ");
 		if ((accessFlags & J9JavaAccessFlags.J9AccClassAnonClass) != 0)
 			out.append("(anonClass) ");
+		if ((accessFlags & J9JavaAccessFlags.J9AccClassIsUnmodifiable) != 0)
```
keithc-ca: This will fail with `NoSuchFieldError` when accessing a core file generated prior to the addition of `J9AccClassIsUnmodifiable`.

This should use a static field initialized with
> J9ConstantHelper.getLong(J9JavaAccessFlags.class, "J9AccClassIsUnmodifiable", 0)

gacholio: The bit won't actually be set in older cores. Should all of the flag accesses not be using the new mechanism?

keithc-ca: It's fine to use fields of the generated structure types when the constant has always been around; it's only when new ones are added that we need to be careful.

This got me thinking that perhaps there's a more general way to handle this, say, by introducing a way to cause `J9DDRClassLoader` to include new fields even for old core files with a suitable default value.

gacholio: Constant fix is in https://github.com/eclipse/openj9/pull/5118

Further discussions can take place elsewhere.

PR #4882, opened 2019-02-26
A fix for TR::CompilationInfo::compileOnSeparateThread()

knn-k:
This commit moves the declaration of a variable `useCodeFromSharedCache`
in TR::CompilationInfo::compileOnSeparateThread(), because it is
declared in a `#if` - `#endif` block and later used outside of `#if` -
`#endif`.
This commit also fixes minor typos.


0xdaryl:
Jenkins test sanity all jdk11

PR #4884, opened 2019-02-26
Release notes for 0.13.0 milestone 1

SueChaplain:
Created a new release note, with work in progress
for milestone 1.

[ci skip]

### DIFF
--- a/doc/release-notes/0.13/0.13.md
+++ b/doc/release-notes/0.13/0.13.md
@@ -1,0 +1,152 @@
+<!--
+* Copyright (c) 2019, 2019 IBM Corp. and others
+*
+* This program and the accompanying materials are made
+* available under the terms of the Eclipse Public License 2.0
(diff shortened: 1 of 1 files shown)


SueChaplain:
@pshipton - I've added all the content for milestone 1 now. One question. There is an issue about bundling OpenSSL for AIX. Do you want me to remove this, now that our intention is not to bundle going forward?

pshipton:
@SueChaplain AIX doesn't work with OpenSSL yet. Steve mentioned he might have a fix. I think we need to wait and confirm it is working before removing the note.

pshipton:
This also means that "OpenSSL 1.0.2 support *(all platforms)*" isn't quite true.

SueChaplain:
@pshipton - I removed the issue from the table about us not being able to bundle OpenSSL on AIX due to an issue as we no longer want to.

I think these notes are ready for the milestone build.

PR #4889, opened 2019-02-26
Remove TR_J9VMBase::traceAssumeFailure

rwy0717:
This function is disabled, and has no callsites in J9. In OMR, there is
one remaining call inside the implemention of TR_ASSERT. Once this call
is removed in eclipse/omr#3613, this function can be deleted.


rwy0717:
The OMR callsite was removed, so it is now safe to remove this function. Rebased and ready for review.

0xdaryl:
Jenkins test sanity all jdk11

PR #4890, opened 2019-02-26
Fix group tag in testKitGen

renfeiw:
- remove `<BS>` character that was accidentally added
- remove duplicated semicolon

[ci skip]


renfeiw:
reviewer: @llxia

Issue #4885, opened 2019-02-26
Crash due to isFinalizable test left behind for non-contiguous stack allocated object


hzongaro:
We're seeing an intermittent crash running the `TreeMapFactory` system test.

hzongaro:
The problem appears to be with an inline `isFinalizable` test that's left behind for an object that was made discontiguous, stack-allocated by Escape Analysis.

Before Escape Analysis we have these trees:

```
n2516n    treetop                                                                             [0x7f80f9501230] bci=[30,4,12] rc=0 vc=1186 vn=471 li=- udi=- nc=1
n2515n      new  jitNewObject[#88  helper Method] [flags 0x400 0x0 ]                          [0x7f80f95011e0] bci=[30,4,12] rc=3 vc=1186 vn=75 li=- udi=- nc=1
n2514n        loadaddr  java/lang/Integer[#704  Static] [flags 0x18307 0x0 ]                  [0x7f80f9501190] bci=[30,4,12] rc=1 vc=1186 vn=74 li=- udi=- nc=0
n2659n    astore  <temp slot 6>[#728  Auto] [flags 0x7 0x0 ]                                  [0x7f80f9503ee0] bci=[30,4,12] rc=0 vc=1186 vn=75 li=102 udi=30 nc=1
n2515n      ==>new
n2638n    astore  <temp slot 3>[#725  Auto] [flags 0x20000007 0x0 ] (privatizedInlinerArg sharedMemory )  [0x7f80f9503850] bci=[30,4,12] rc=0 vc=1186 vn=75 li=103 udi=31 nc=1 flg=0x2000
n2515n      ==>new
...
n2613n    ificmpne --> block_249 BBStart at n2619n ()                                         [0x7f80f9503080] bci=[33,0,39] rc=0 vc=1186 vn=478 li=- udi=- nc=2 flg=0x20
n3407n      iand                                                                              [0x7f80f95b28b0] bci=[33,0,39] rc=1 vc=1186 vn=86 li=- udi=- nc=2
n3406n        l2i                                                                             [0x7f80f95b2860] bci=[33,0,39] rc=1 vc=1186 vn=85 li=- udi=- nc=1
n2609n          lloadi  <isClassAndDepthFlags>[#367  Shadow +24] [flags 0x603 0x0 ]           [0x7f80f9502f40] bci=[33,0,39] rc=1 vc=1186 vn=84 li=- udi=- nc=1
[...]

PR #4891, opened 2019-02-26
Revise tree pattern used to match inline finalizable test

hzongaro:
The pattern used to match an inline finalizable test might be incorrect in 64-bit mode as tree simplification might result in 32-bit operations being used.  This change accounts for that possibility by checking whether an `l2i` was applied to the result of the load of `isClassAndDepthFlags`, and checks for 32-bit operations accordingly.

Fixes:  Issue #4885

### DIFF
--- a/runtime/compiler/optimizer/EscapeAnalysis.cpp
+++ b/runtime/compiler/optimizer/EscapeAnalysis.cpp
@@ -3742,32 +3742,59 @@ static bool isFinalizableInlineTest(TR::Compilation *comp, TR::Node *candidate, 
    TR_J9VMBase *fej9 = (TR_J9VMBase *)(comp->fe());
 
    bool is64Bit = TR::Compiler->target.is64Bit();
-   TR::ILOpCodes ifOp = is64Bit ? TR::iflcmpne : TR::ificmpne;
-   TR::ILOpCodes andOp = is64Bit ? TR::land : TR::iand;
+
+   // root
+   //   r1      - first child of root
+   //     r11   - first child of first child of root
+   //     r12   - second child of first child of root
+   //   r2      - second child of root
+   TR::Node *r1 = (root->getNumChildren() > 0) ? root->getFirstChild() : NULL;
+   TR::Node *r2 = (root->getNumChildren() > 1) ? root->getSecondChild() : NULL;
+   TR::Node *r11 = (r1 && r1->getNumChildren() > 0) ? r1->getFirstChild() : NULL;
+   TR::Node *r12 = (r1 && r1->getNumChildren() > 1) ? r1->getSecondChild() : NULL;
+
+   bool castDownToInt = is64Bit && r11 && (r11->getOpCodeValue() == TR::l2i);
+   bool usesLongOps = is64Bit && !castDownToInt;
+
+   TR::ILOpCodes ifOp = usesLongOps ? TR::iflcmpne : TR::ificmpne;
+   TR::ILOpCodes andOp = usesLongOps ? TR::land : TR::iand;
    TR::ILOpCodes loadOp = is64Bit ? TR::lloadi : TR::iloadi;
-   TR::ILOpCodes constOp = is64Bit ? TR::lconst : TR::iconst;
+   TR::ILOpCodes constOp = usesLongOps ? TR::lconst : TR::iconst;
+
+   TR::Node *loadNode = castDownToInt ? r11->getFirstChild() : r11;
 
    /*
       Looking for a pattern of:
+
       if[i/l]cmpne                                      <root>
-         [i/l]and
-            [i/l]loadi <classAndDepthFlags>
+         [i/l]and                                       <r1>
+            [i/l]loadi <classAndDepthFlags>             <r11/loadNode>
                aloadi <vft-symbol>                      <vftLoad>
                   ...                                   <ref>
-            [i/l]const <FlagValueForFinalizerCheck>
-         [i/l]const 0
+            [i/l]const <FlagValueForFinalizerCheck>     <r12>
+         [i/l]const 0                                   <r2>
+
+       or
+
+      ificmpne                                          <root>
+         iand                                           <r1>
+            l2i                                         <r11>
+               lloadi <classAndDepthFlags>              <loadNode>
+                  aloadi <vft-symbol>                   <vftLoad>
+                     ...                                <ref>
+            iconst <FlagValueForFinalizerCheck>         <r12>
+         iconst 0                                       <r2>
    */
 
    return root->getOpCodeValue() == ifOp &&
-          root->getFirstChild()->getOpCodeValue() == andOp &&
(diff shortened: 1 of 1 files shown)


runtime/compiler/optimizer/EscapeAnalysis.cpp
```diff
@@ -3742,32 +3742,60 @@ static bool isFinalizableInlineTest(TR::Compilation *comp, TR::Node *candidate,
...
+   TR::ILOpCodes ifOp = usesLongOps ? TR::iflcmpne : TR::ificmpne;
+   TR::ILOpCodes andOp = usesLongOps ? TR::land : TR::iand;
    TR::ILOpCodes loadOp = is64Bit ? TR::lloadi : TR::iloadi;
-   TR::ILOpCodes constOp = is64Bit ? TR::lconst : TR::iconst;
+   TR::ILOpCodes constOp = usesLongOps ? TR::lconst : TR::iconst;
+
+   TR::Node *loadNode = castDownToInt ? r11->getFirstChild() : r11;
 
    /*
       Looking for a pattern of:
```
fjeremic: FYI if you did not know we have quite a handy utility `TR_PatternBuilder` which is a very concise and readable way of doing tree pattern matching. There are several examples if you search the code base (eg. [1]).

[1] https://github.com/eclipse/omr/blob/88fd3ecdbbdc0ef7ace859dacc069b2c38e9ef27/compiler/codegen/OMRTreeEvaluator.cpp#L363-L380

hzongaro: Thanks, Filip!  I did not know, and I had been thinking that something like that was needed.  Why doesn't its use seem to be particularly widespread?

fjeremic: Probably because not many people know about it. I've been encouraging it's use in PR reviews if I spot an opportunity. If it works well for you perhaps you can advocate for it as well? Thanks!

hzongaro: @fjeremic Filip, I was playing around with the `TR_PatternBuilder`.  I think in this case using it might be a bit messy because we want to match either `ificmpne` or `iflcmpne`, `iand` or `land`, etc.  So either I would have to build the patterns with `pb.choice` throughout, or build three different patterns:  one pure `int` case, one pure `long` case and one mixed with an `i2l` in the middle.  We might want to think about having some built-in mechanisms for matching either `int` or `long` operations to simplify this kind of scenario.

The other thing I found is that tree pattern matching seems to be currently disabled.  We'll investigate why that is, but Andrew @andrewcraik  suggested I leave it aside.  If you don't have any objection, that's what I'll do for now.  I will definitely look for further opportunities to use the pattern builder once it's enabled.

(As an aside, I really miss XPath!)

fjeremic: That sounds reasonable.

andrewcraik: I think the disable of the matching is conflating the disable of OverflowCHK usage with the pattern matcher - we'll need to separate the two concepts to allow this stuff to be used elsewhere.

runtime/compiler/optimizer/EscapeAnalysis.cpp
```diff
@@ -4986,7 +5014,15 @@ bool TR_EscapeAnalysis::fixupNode(TR::Node *node, TR::Node *parent, TR::NodeChec
                TR::Node *andNode = _curTree->getNode()->getFirstChild();
                andNode->removeAllChildren();
 
-               if (TR::Compiler->target.is64Bit())
+               TR_ASSERT(andNode->getOpCodeValue() == TR::iand
```
andrewcraik: This seems like it should be an assert fatal in the if ladder below processing andNode. Is there anything sensible to be done if we don't have an `iand` or `land` or do we need to bring the VM down?

runtime/compiler/optimizer/EscapeAnalysis.cpp
```diff
@@ -3742,32 +3742,60 @@ static bool isFinalizableInlineTest(TR::Compilation *comp, TR::Node *candidate,
...
-          (is64Bit ? root->getSecondChild()->getLongInt() == 0 : root->getSecondChild()->getInt() == 0) &&
-          root->getFirstChild()->getFirstChild()->getOpCodeValue() == loadOp &&
-          root->getFirstChild()->getSecondChild()->getOpCodeValue() == constOp &&
-          (is64Bit ?
-           root->getFirstChild()->getSecondChild()->getLongInt() == fej9->getFlagValueForFinalizerCheck() :
-           root->getFirstChild()->getSecondChild()->getInt() == fej9->getFlagValueForFinalizerCheck()) &&
-          root->getFirstChild()->getFirstChild()->getFirstChild() == vftLoad /*&& (implied by the above assume)
+          r1->getOpCodeValue() == andOp &&
+          r2->getOpCodeValue() == constOp &&
+          (usesLongOps ? r2->getLongInt() == 0 : r2->getInt() == 0) &&
```
andrewcraik: would this be more clear as:

`(usesLongOps ? r2->getLongInt() : r2->getInt()) == 0` - that would imply some kind of type conversion but that wouldn't affect the value 0 which is what we would be looking for...

runtime/compiler/optimizer/EscapeAnalysis.cpp
```diff
@@ -3742,32 +3742,60 @@ static bool isFinalizableInlineTest(TR::Compilation *comp, TR::Node *candidate,
...
+          loadNode->getOpCodeValue() == loadOp &&
+          r12->getOpCodeValue() == constOp &&
+          (usesLongOps
```
andrewcraik: could the == be factored out safely here as well or would the widening cause a problem?

andrewcraik:
@hzongaro any update since the issue this is fixing is tagged as 0.13?

runtime/compiler/optimizer/EscapeAnalysis.cpp
hzongaro: There's probably nothing sensible that we can do.  I'll rearrange the position of the assertion, and make it a `TR_ASSERT_FATAL`

hzongaro:
> @hzongaro any update since the issue this is fixing is tagged as 0.13?
In the process of running some sanity testing after addressing those comments

andrewcraik:
Please squash - I think that should address my concerns so I'll start sanity.

hzongaro:
Thanks, Andrew @andrewcraik .  Just squashed as commit https://github.com/eclipse/openj9/commit/dba4ac6da60bedfdaf29d102dc6cf64351e5f29f

andrewcraik:

andrewcraik:
sanity was an infra issue, the rest of the checks pass so merging - @hzongaro can you please create PR for the 0.13 release branch? FYI @pshipton

PR #4892, opened 2019-02-26
Do not ignore any relocation records

dsouzai:
Currently there is a coupling between the existence of guards and the
relocation of the inlined table in the metadata. This means that if
the JIT does not produce a guard for certain inlined sites, the
associated entry in the inlined table in the metadata will not be
relocated. This has the consequence of causing other relocations that
need information via this entry to not have the means to do so. The
only way to proceed is to fail the AOT load - if the compiler cannot
materialize a value to relocate a location that **MUST** be
relocated, the code **CANNOT** be loaded and executed.

Fixes https://github.com/eclipse/openj9/issues/4201 (again (I hope))

### DIFF
--- a/runtime/compiler/runtime/RelocationRecord.cpp
+++ b/runtime/compiler/runtime/RelocationRecord.cpp
@@ -932,6 +932,19 @@ TR_RelocationRecordWithInlinedSiteIndex::ignore(TR_RelocationRuntime *reloRuntim
 
    if (method == reinterpret_cast<J9Method *>(-1))
       {
+      /* Currently there is a coupling between the existence of guards and the
+       * relocation of the inlined table in the metadata. This means that if the
+       * JIT does not produce a guard for certain inlined sites, the associated
+       * entry in the inlined table in the metadata will not be relocated. This
+       * has the consequence of causing other relocations that need information
+       * via this entry to not have the means to do so. The only way to proceed
+       * is to fail the AOT load - if the compiler cannot materialize a value to
+       * relocate a location that **MUST** be relocated, the code **CANNOT** be
+       * loaded and executed.
+       */
+      RELO_LOG(reloRuntime->reloLogger(), 6, "\tAborting Load: cannot ignore relocation!\n");
+      reloRuntime->comp()->failCompilation<J9::AOTNoSupportForAOTFailure>("inlined site method is -1");
+
       if (reloRuntime->comp()->getOption(TR_UseSymbolValidationManager))
          {
          /* With the SVM, it isn't possible for the method to be equal to -1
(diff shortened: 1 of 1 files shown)


dsouzai:

dsouzai:
@mstoodle could you also review?

mstoodle:

dsouzai:
Most builds are failing because of 

```
13:24:14  Creating images/lib/ext/dtfj.jar
13:24:15  'u' flag requires manifest, 'e' flag or input files to be specified!
13:24:15  Usage: jar {ctxui}[vfmn0PMe] [jar-file] [manifest-file] [entry-point] [-C dir] files ...
13:24:15  Options:
13:24:15      -c  create new archive
13:24:15      -t  list table of contents for archive
13:24:15      -x  extract named (or all) files from archive
13:24:15      -u  update existing archive
13:24:15      -v  generate verbose output on standard output
13:24:15      -f  specify archive file name
13:24:15      -m  include manifest information from specified manifest file
13:24:15      -n  perform Pack200 normalization after creating a new archive
13:24:15      -e  specify application entry point for stand-alone application 
13:24:15          bundled into an executable jar file
13:24:15      -0  store only; use no ZIP compression
13:24:15      -P  preserve leading '/' (absolute path) and ".." (parent directory) components from file names
13:24:15      -M  do not create a manifest file for the entries
13:24:15      -i  generate index information for the specified jar files
13:24:15      -C  change to the specified directory and include the following file
13:24:15  If any file is a directory then it is processed recursively.
13:24:15  The manifest file name, the archive file name and the entry point name are
13:24:15  specified in the same order as the 'm', 'f' and 'e' flags.
13:24:15  
13:24:15  Example 1: to archive two class files into an archive called classes.jar: 
13:24:15         jar cvf classes.jar Foo.class Bar.class 
13:24:15  Example 2: use an existing manifest file 'mymanifest' and archive all the
13:24:15             files in the foo/ directory into 'classes.jar': 
13:24:15         jar cvfm classes.jar mymanifest -C foo/ .
13:24:15  
13:24:15  make[2]: *** [/Users/jenkins/workspace/PullRequest-Sanity-JDK8-osx_x86-64_cmprssptrs-OpenJ9/build/macosx-x86_64-normal-server-release/images/lib/ext/dtfj.jar] Error 1
13:24:15  make[2]: *** Waiting for unfinished jobs....
13:24:19  make[1]: *** [images] Error 2
13:24:19  make: *** [images-only] Error 2
```
:man_shrugging:

dsouzai:
jenkins test sanity all jdk8

dsouzai:
Suspected change causing build issue was reverted https://github.com/eclipse/openj9/pull/4913

dsouzai:
@mstoodle all checks passed :)

mstoodle:
thanks for the ping. merging now

PR #4894, opened 2019-02-26
Fix issues with thunks in remote AOT mode

dmitry-ten:
JITaaS generates all thunks inside the code cache and later
relocates them in one batch on the client at the end of the compilation.
This means get/set/persist thunk logic should be slightly different
from mainline.
We do not need shared cache versions of `setJ2IThunk`
on the server, because client already receives all thunks,
so we can store them to shared class cache at the time of relocation.

To this end,
- Remove `TR_J9SharedCacheServerVM` versions of `setJ2IThunk`,
`persistThunk`, `persistJ2IThunk`.
- In `TR::CompilationInfoPerThread::relocateThunks()` call the shared
cache version of `setJ2IThunk` in addition to the base version.
- Override `TR_ResolvedRelocatableJ9JITaaSServerMethod::constantPool()`
to match `TR_ResolvedRelocatableJ9Method` implementation. It needs to
return ROM literals and without it thunk relocation will not work.


dmitry-ten:
`get/setJ2IThunk` will be called, but we need to use non-AOT implementation.
I'll add asserts for other methods.

dmitry-ten:
Done.

PR #4895, opened 2019-02-26
Refactor fold native MHInterpreter

theresa-m:
- Eliminates an extra memcpy in foldArguments MethodHandle calls
- Makes code easier to understand - reason for copying is to make stack compatible with insertPlaceHolderFrame

@DanHeidinga as discussed previously for #3980 

### DIFF
--- a/runtime/vm/MHInterpreter.cpp
+++ b/runtime/vm/MHInterpreter.cpp
@@ -1386,10 +1386,10 @@ VM_MHInterpreter::filterArgumentsWithCombiner(j9object_t methodHandle)
 	j9object_t combinerType = getMethodHandleMethodType(combinerHandle);
 	U_32 combinerArgSlots = getMethodTypeArgSlots(combinerType);
 
-	/* Add combinerHandle to top of stack and adjust sp to make room for combinerHandle.type.argSlots */
-	/* [... filterHandle args descriptionBytes MethodTypeFrame combinerHandle combinerArgs] */
+	/* Adjust sp to make room for combinerHandle and combinerArgs. There is no point in actually adding the 
+	 * combinerHandle to the stack here because it will later be overwritten by insertPlaceHolderFrame. */
+	/* [... filterHandle args descriptionBytes MethodTypeFrame combinerHandle(empty) combinerArgs(empty)] */
 	_currentThread->sp -= (combinerArgSlots + 1);
-	((j9object_t *)_currentThread->sp)[combinerArgSlots] = combinerHandle;
 	UDATA *spCombinerSlot = _currentThread->sp + combinerArgSlots;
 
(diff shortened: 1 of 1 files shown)


runtime/vm/MHInterpreter.cpp
```diff
@@ -1507,48 +1507,36 @@ VM_MHInterpreter::foldForFoldArguments(j9object_t methodHandle)
...
 	U_32 combinerArgSlots = getMethodTypeArgSlots(combinerType);
 
-	/* Replace foldHandle with combinerHandle and adjust sp according to combinerHandle.type.argSlots */
-	/* [... foldHandle args descriptionBytes MethodTypeFrame foldHandle PlaceHolderFrame combinerHandle args] */
-	((j9object_t *)_currentThread->sp)[foldArgSlots] = combinerHandle;
-	UDATA *spCombinerSlot = _currentThread->sp + foldArgSlots;
+	/* Add combinerHandle to top of stack and adjust sp to make room for combinerHandle.type.argSlots */
+	/* [... foldHandle args descriptionBytes MethodTypeFrame combinerHandle combinerArgs] */
+	_currentThread->sp -= (combinerArgSlots  + 1);
+	((j9object_t *)_currentThread->sp)[combinerArgSlots] = combinerHandle;
```
DanHeidinga: Is this required?  If it's overwritten by the insertPlaceHolderFrame below, then I think we can skip it here

theresa-m: you're right its not required. I've removed a similar line in `filterArgumentsWithCombiner` as well

runtime/vm/MHInterpreter.cpp
```diff
@@ -1600,9 +1586,18 @@ VM_MHInterpreter::foldForFoldArguments(j9object_t methodHandle)
...
 		}
-		_currentThread->sp = spCombinerSlot;
+		/* after loop spCombinerSlot is at the top of the combiner arguments */
+		Assert_VM_true(spCombinerSlot == _currentThread->sp);
 	}
 
+	/* [... foldHandle args descriptionBytes MethodTypeFrame foldHandle PlaceHolderFrame foldHandle combinerArgs] */
+	insertPlaceHolderFrame(combinerArgSlots, methodHandle, J9VMJAVALANGINVOKEMETHODHANDLE_FOLDHANDLEPLACEHOLDER_METHOD(_vm));
+
+	/* insertPlaceHolderFrame will replace combinerHandle with methodHandle. This should be undone so the stack is correctly
```
DanHeidinga: This comment isn't correct any more as the MH isn't written until the end.  Can you update it here and in the other case as well?

theresa-m: updated.

DanHeidinga:
Jenkins test sanity aix,win,xlinux jdk11

PR #4899, opened 2019-02-26
Add range check test around profiled guard method test

r30shah:
In OpenJ9 we take advantage of the profiled information in inlining method for virtual call. While inlining the method, we generate a guard around inlined code to make sure that the inlined
implementation is indeed what we need to execute. For this TR_ProfiledGuard** kind of guard it either TR_VftTest or TR_MethodTest.

In case of TR_MethodTest, it records the offset of the inlined method in the Virtual Function Table of the class in which that method belongs to and compares the value at the same offset of class of receiver object with the J9Method of inlined method. This comparison introduces a functional bug.

Some classes would have smaller function table and recorded offset w.r.t the class might point to a garbage value. This commit adds a range check test around the TR_MethodTest to verify that address that we are going to reference points to valid address.


r30shah:
FYI @andrewcraik @pshipton  PR to port changes from PR #4721 to 0.13 release branch.

PR #4900, opened 2019-02-27
Add AOT upgrades based on SCC hints to the LPQ

dsouzai:
Forced AOT upgrades are recompilations that are based on hints stored in
the SCC. These hints are used to identify methods that were AOT compiled
(or loaded), but were recompiled because the sampling mechanism
determined that they were important methods. However, these upgrades are
added immediately to the queue when the method is loaded from the SCC,
which can potentially impact an application's rampup phase.

This PR defers these upgrades until free cycles are available by adding them to
the low priority queue.

### DIFF
--- a/runtime/compiler/control/CompilationThread.cpp
+++ b/runtime/compiler/control/CompilationThread.cpp
@@ -6635,8 +6635,19 @@ TR::CompilationInfoPerThreadBase::postCompilationTasks(J9VMThread * vmThread,
             // Now let's see if we need to schedule an AOT upgrade
             if (hints)
                {
-               entry->_newStartPC = startPC; // must do this before calling queueForcedAOTUpgrade
-               _compInfo.queueForcedAOTUpgrade(entry, hints, _vm);
+               // must do this before queueing for an upgrade
+               entry->_newStartPC = startPC;
+
+               static char *disableQueueLPQAOTUpgrade = feGetEnv("TR_DisableQueueLPQAOTUpgrade");
+               if (TR::Options::getAOTCmdLineOptions()->getOption(TR_EnableSymbolValidationManager)
+                   && !disableQueueLPQAOTUpgrade)
+                  {
+                  _compInfo.getLowPriorityCompQueue().addUpgradeReqToLPQ(getMethodBeingCompiled());
+                  }
+               else
+                  {
+                  _compInfo.queueForcedAOTUpgrade(entry, hints, _vm);
+                  }
                }
             }
          }
--- a/runtime/compiler/control/J9Options.cpp
+++ b/runtime/compiler/control/J9Options.cpp
@@ -2446,6 +2446,17 @@ bool J9::Options::feLatePostProcess(void * base, TR::OptionSet * optionSet)
       self()->setOption(TR_DisableIntrinsics);
       }
 
+   /* Temporary (until SVM is the default)
+    *
+    * If the SVM is not enabled, use the old _coldUpgradeSampleThreshold
+    * for AGGRESSIVE_AOT
+    */
+   if (!self()->getOption(TR_EnableSymbolValidationManager)
+       && TR::Options::_aggressivenessLevel == TR::Options::AGGRESSIVE_AOT)
+      {
+      TR::Options::_coldUpgradeSampleThreshold = 10;
+      }
+
    return true;
    }
 


dsouzai:

dsouzai:
fyi @vijaysun-omr

vijaysun-omr:
By "steady state" in your description, I assume you just mean "when there are free CPU cycles" ? 
I don't see any checking of the "phase" that the JVM is in in your change (and that's fine) which is why I am asking the question.

mpirvu:
LPQ requests are processed only under some conditions that are independent of the phase
```
bool TR::CompilationInfo::canProcessLowPriorityRequest()
   {
   // To avoid overhead cycling through all the threads, we should first
   // check if there is anything in that low priority queue
   //
   if (!getLowPriorityCompQueue().hasLowPriorityRequest()) // access _firstLPQentry without monitor in hand
      return false;

   // If the main compilation queue has requests, we should serve those first
   if (getMethodQueueSize() != 0)
      return false;

   // To process a request from the low priority queue we need to have
   // (1) no other compilation in progress (not required if TR_ConcurrentLPQ is enabled)
   // (2) some idle CPU
   // (3) current JVM not taking its entire share of CPU entitlement
   //
   if (TR::Options::getCmdLineOptions()->getOption(TR_ConcurrentLPQ) &&
       _jitConfig->javaVM->phase == J9VM_PHASE_NOT_STARTUP) // ConcurrentLPQ is too damaging to startup
      {
      if ((getCpuUtil() && getCpuUtil()->isFunctional() &&
         getCpuUtil()->getCpuIdle() > idleThreshold() && // enough idle time
         getJvmCpuEntitlement() - getCpuUtil()->getVmCpuUsage() >= 200)) // at least 2 cpus should be not utilized
         return true;
      // Fall through the next case where we allow LPQ if the JVM leaves 50% of a CPU
      // unutilized, but there should be nothing in the main queue
      }

   // Cycle through all the compilation threads
   for (uint8_t i = 0; i < getNumUsableCompilationThreads(); i++)
      {
      TR::CompilationInfoPerThread *curCompThreadInfoPT = _arrayOfCompilationInfoPerThread[i];
      TR_ASSERT(curCompThreadInfoPT, "a thread's compinfo is missing\n");

      if (curCompThreadInfoPT->getMethodBeingCompiled())
         return false;
      }
   return (getCpuUtil() && getCpuUtil()->isFunctional() &&
      getCpuUtil()->getCpuIdle() > idleThreshold() && // enough idle time
           getJvmCpuEntitlement() - getCpuUtil()->getVmCpuUsage() > 50); // at least half a processor should be empty
   }
```

dsouzai:
> By "steady state" in your description, I assume you just mean "when there are free CPU cycles" ?
I don't see any checking of the "phase" that the JVM is in in your change (and that's fine) which is why I am asking the question.

Ah, yes, you're right; I was mixing it up with a change that I thought was also required (ie only process the LPQ during the STEADY_STATE jitstate); good catch. I'll update the PR/commit message.

> We may want to have an option to switch between the old and the new behavior.

yeah good point, I'll make the requested changes.

dsouzai:
@mpirvu @vijaysun-omr I made the requested changes.

mpirvu:
jenkins test sanity all jdk8

vijaysun-omr:
LGTM to me too.

dsouzai:
jenkins test sanity all jdk8

dsouzai:
don't know why jenkins didn't trigger before :S

dsouzai:
Builds are failing because of 

```
 > git fetch --tags --progress https://github.com/eclipse/openj9.git +refs/pull/*:refs/remotes/origin/pr/*
ERROR: Error fetching remote repo 'origin'
hudson.plugins.git.GitException: Failed to fetch from https://github.com/eclipse/openj9.git
    at hudson.plugins.git.GitSCM.fetchFrom(GitSCM.java:894)
    at hudson.plugins.git.GitSCM.retrieveChanges(GitSCM.java:1161)
    at hudson.plugins.git.GitSCM.checkout(GitSCM.java:1192)
    at org.jenkinsci.plugins.workflow.steps.scm.SCMStep.checkout(SCMStep.java:120)
    at org.jenkinsci.plugins.workflow.cps.CpsScmFlowDefinition.create(CpsScmFlowDefinition.java:144)
    at org.jenkinsci.plugins.workflow.cps.CpsScmFlowDefinition.create(CpsScmFlowDefinition.java:67)
    at org.jenkinsci.plugins.workflow.job.WorkflowRun.run(WorkflowRun.java:293)
    at hudson.model.ResourceController.execute(ResourceController.java:97)
    at hudson.model.Executor.run(Executor.java:429)
Caused by: hudson.plugins.git.GitException: Command "git fetch --tags --progress https://github.com/eclipse/openj9.git +refs/pull/*:refs/remotes/origin/pr/*" returned status code 128:
stdout: 
stderr: error: RPC failed; curl 18 transfer closed with outstanding read data remaining
fatal: The remote end hung up unexpectedly
    at org.jenkinsci.plugins.gitclient.CliGitAPIImpl.launchCommandIn(CliGitAPIImpl.java:2042)
    at org.jenkinsci.plugins.gitclient.CliGitAPIImpl.launchCommandWithCredentials(CliGitAPIImpl.java:1761)
    at org.jenkinsci.plugins.gitclient.CliGitAPIImpl.access$400(CliGitAPIImpl.java:72)
    at org.jenkinsci.plugins.gitclient.CliGitAPIImpl$1.execute(CliGitAPIImpl.java:442)
    at hudson.plugins.git.GitSCM.fetchFrom(GitSCM.java:892)
    ... 8 more
ERROR: Error fetching remote repo 'origin'
```

dsouzai:
Builds seem ok in a different PR

dsouzai:
jenkins test sanity all jdk8

dsouzai:
@mpirvu all checks passed :)

PR #4905, opened 2019-02-27
Add missing indirection

keithc-ca:
* discovered via compiler warning [-Wpointer-compare]
	`comparison between pointer and zero character constant`


DanHeidinga:
Jenkins compile zlinux jdk8

PR #4909, opened 2019-02-27
master: Fix memory semantics for Unsafe.getAndSet*

theresa-m:
This issue was identified in #3980 and created as a separate pr because this code is not Java 12 specific fyi @DanHeidinga 

Note that float, double and boolean are not updated because they call into getAndSet for int, long, byte respectively.


DanHeidinga:

pshipton:
Does this need to be delivered to the 0.13 branch as well?

theresa-m:
that would probably be best. I'll open another pr for the v0.13.0-release branch

Issue #3742, opened 2018-11-20
Test-sanity.functional-JDK11-linux_x86-64 UnsafeTests_0 testArrayGetAndOpInt NullPointerException


pshipton:
https://ci.eclipse.org/openj9/job/Test-sanity.functional-JDK11-linux_x86-64/192
occurred in OMR acceptance, but seems like an intermittent problem and not related to the OMR changes.
https://github.com/eclipse/omr/compare/69b1b77...0c0f19c

```
===============================================
Running test UnsafeTests_0 ...
===============================================
UnsafeTests_0 Start Time: Tue Nov 20 05:27:16 2018 Epoch Time (ms): 1542691636257
variation: -DScenario=Regular
JVM_OPTIONS: -Xnocompressedrefs -DScenario=Regular 
[IncludeExcludeTestAnnotationTransformer] [INFO] exclude file is /home/jenkins/workspace/Test-sanity.functional-JDK11-linux_x86-64/openjdk-tests/TestConfig/scripts/testKitGen/../../../../jvmtest/TestConfig/resources/excludes/latest_exclude_11.txt
...
... TestNG 6.14.2 by Cédric Beust (cedric@beust.com)
...

FAILED: testArrayGetAndOpInt (Regular)
java.lang.NullPointerException
	at org.openj9.test.unsafe.TestUnsafeGetAndOp.testArrayGetAndOpInt(TestUnsafeGetAndOp.java:135)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.base/java.lang.reflect.Method.invoke(Method.java:566)
	at org.testng.internal.MethodInvocationHelper.invokeMethod(MethodInvocationHelper.java:124)
	at org.testng.internal.Invoker.invokeMethod(Invoker.java:580)
	at org.testng.internal.Invoker.invokeTestMethod(Invoker.java:716)
	at org.testng.internal.Invoker.invokeTestMethods(Invoker.java:988)
	at org.testng.internal.TestMethodWorker.invokeTestMethods(TestMethodWorker.java:125)
	at org.testng.internal.TestMethodWorker.run(TestMethodWorker.java:109)
[...]

PR #4918, opened 2019-02-27
Add VM function to query if the JVM supports VTs

tajila:
Add VM function to query if the JVM supports VTs

ValueTypes is an experimental feature and must be enabled with
the -XX:+EnableValhalla option. This PR also adds a vm helper for
other modules to query whether valuesTypes have been enabled on 
the JDK.


tajila:

tajila:
CI build failure was caused by https://github.com/eclipse/openj9/pull/5398. I rebased to to trigger another run

runtime/oti/j9nonbuilder.h
```diff
@@ -5363,6 +5365,7 @@ typedef struct J9JavaVM {
 #ifdef J9VM_OPT_VALHALLA_VALUE_TYPES
 	UDATA valueFlatteningThreshold;
 #endif /* defined(J9VM_OPT_VALHALLA_VALUE_TYPES) */
+	U_64 extendedRuntimeFlags2;
```
DanHeidinga: please put this with the other runtimeFlags field near the top of the structure.  We've seen surprising overheads reading this far into the J9JavaVM for status flags

runtime/oti/jvminit.h
```diff
@@ -374,6 +374,11 @@ enum INIT_STAGE {
 #define VMOPT_XXENABLEPOSITIVEHASHCODE "-XX:+PositiveIdentityHash"
 #define VMOPT_XXDISABLEPOSITIVEHASHCODE "-XX:-PositiveIdentityHash"
 
+#if defined(J9VM_OPT_VALHALLA_VALUE_TYPES)
+#define VMOPT_XXENABLEVALHALLA "-XX:+EnableValhalla"
+#define VMOPT_XXDISABLEVALHALLA "-XX:-EnableValhalla"
```
DanHeidinga: `EnableValhalla` -> `EnableValueTypes`.  Name the feature for the particular part thats being enabled / disabled rather than the overall project.  Valhalla might be used for LW10 now, and LW20 in the future.

runtime/vm/jvminit.c
```diff
@@ -588,6 +588,12 @@ cleanCustomSpinOptions(void *element, void *userData)
 }
 #endif /* J9VM_INTERP_CUSTOM_SPIN_OPTIONS */
 
+BOOLEAN
+areValueTypesEnabled(J9JavaVM *vm)
+{
+	return J9_ARE_ALL_BITS_SET(vm->extendedRuntimeFlags2, J9_EXTENDED_RUNTIME2_ENABLE_VALHALLA);
```
DanHeidinga: `J9_EXTENDED_RUNTIME2_ENABLE_VALHALLA` -> `J9_EXTENDED_RUNTIME2_ENABLE_VALUETYPES`

runtime/oti/j9nonbuilder.h
gacholio: Also not a fan of U_64. Until we get smart and start using proper bitfields, let's keep flags fields as U_32 or UDATA.

DanHeidinga: Not UDATA as it wastes 1/2 the range on 64 bit.

runtime/oti/jvminit.h
tajila: @DanHeidinga I choose the naming to be consistent with the LW1 prototype on the RI. 

Does it matter that the -XX option names are the same?

DanHeidinga: That's a reasonable approach and is better than having conflicting options for testing LW10 functionality.  

The reason to make the runtime flag and option consistent is that it's easier to find the all the references in a single search.

tajila:
I've updated the PR

runtime/vm/jvminit.c
```diff
@@ -588,6 +588,12 @@ cleanCustomSpinOptions(void *element, void *userData)
 }
 #endif /* J9VM_INTERP_CUSTOM_SPIN_OPTIONS */
 
+BOOLEAN
```
gacholio: Why have this as a function instead of a macro? Is there someone that really needs to call it indirectly?

tajila: The JIT will need it

DanHeidinga:

DanHeidinga:
Unsafe test failures are a known issue - see #3742

pshipton:
This is causing some test failure, reverting while we figure it out.

Issue #4661, opened 2019-02-07
Java 12 functional sanity failure openj9_jsr292Test test_defineClass_ClassBytes_CorruptedArray


pshipton:
I ran a Java 12 functional sanity test using the changes from #3980 and #4531.

https://ci.eclipse.org/openj9/job/Test-sanity.functional-JDK12-linux_390-64_cmprssptrs/3
openj9_jsr292Test_0
openj9_jsr292Test_JitCount0_0
```
FAILED: test_defineClass_ClassBytes_CorruptedArray
org.testng.TestException: 
Expected exception of type class java.lang.LinkageError but got java.lang.AssertionError: The test case failed to detect a corrupted class byte array
	at org.testng.internal.ExpectedExceptionsHolder.wrongException(ExpectedExceptionsHolder.java:68)
	at org.testng.internal.Invoker.considerExceptions(Invoker.java:1129)
	at org.testng.internal.Invoker.invokeMethod(Invoker.java:612)
	at org.testng.internal.Invoker.invokeTestMethod(Invoker.java:716)
	at org.testng.internal.Invoker.invokeTestMethods(Invoker.java:988)
	at org.testng.internal.TestMethodWorker.invokeTestMethods(TestMethodWorker.java:125)
	at org.testng.internal.TestMethodWorker.run(TestMethodWorker.java:109)
	at org.testng.TestRunner.privateRun(TestRunner.java:648)
	at org.testng.TestRunner.run(TestRunner.java:505)
	at org.testng.SuiteRunner.runTest(SuiteRunner.java:455)
	at org.testng.SuiteRunner.runSequentially(SuiteRunner.java:450)
	at org.testng.SuiteRunner.privateRun(SuiteRunner.java:415)
	at org.testng.SuiteRunner.run(SuiteRunner.java:364)
	at org.testng.SuiteRunnerWorker.runSuite(SuiteRunnerWorker.java:52)
	at org.testng.SuiteRunnerWorker.run(SuiteRunnerWorker.java:84)
	at org.testng.TestNG.runSuitesSequentially(TestNG.java:1208)
	at org.testng.TestNG.runSuitesLocally(TestNG.java:1137)
	at org.testng.TestNG.runSuites(TestNG.java:1049)
	at org.testng.TestNG.run(TestNG.java:1017)
	at org.testng.TestNG.privateMain(TestNG.java:1354)
	at org.testng.TestNG.main(TestNG.java:1323)
[...]

Issue #3641, opened 2018-11-12
Remove workaround for Java 12 in java.lang.Access.defineClass()


keithc-ca:
https://github.com/eclipse/openj9/pull/3638 introduces a workaround that should be removed. Ideally the behavior of that method should be the same for all versions of Java.

pshipton:
@keithc-ca what is the next step to remove the workaround?

DanHeidinga:
@keithc-ca Can you point to examples of why the work around was required?  We're running out of time to have this sorted for 0.13.0

keithc-ca:
Creating a `jmod` fails without the workaround because the validation done by `ClassLoader.defineClass` fails for class names containing `/`:
```
Exception in thread "main" java.lang.NoClassDefFoundError: jdk/internal/reflect/GeneratedMethodAccessor1
	at java.base/java.lang.ClassLoader.defineClass(ClassLoader.java:470)
	at java.base/java.lang.Access.defineClass(Access.java:244)
	at java.base/jdk.internal.reflect.ClassDefiner.defineClass(ClassDefiner.java:65)
	at java.base/jdk.internal.reflect.MethodAccessorGenerator$1.run(MethodAccessorGenerator.java:400)
	at java.base/jdk.internal.reflect.MethodAccessorGenerator$1.run(MethodAccessorGenerator.java:394)
	at java.base/java.security.AccessController.doPrivileged(AccessController.java:647)
	at java.base/jdk.internal.reflect.MethodAccessorGenerator.generate(MethodAccessorGenerator.java:393)
	at java.base/jdk.internal.reflect.MethodAccessorGenerator.generateMethod(MethodAccessorGenerator.java:75)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:53)
	at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.base/java.lang.reflect.Method.invoke(Method.java:567)
	at java.base/java.lang.Class.copyConstructors(Class.java:4155)
	at java.base/java.lang.Class.cacheConstructors(Class.java:4199)
	at java.base/java.lang.Class.getDeclaredConstructors(Class.java:859)
	at java.base/java.lang.invoke.InnerClassLambdaMetafactory$1.run(InnerClassLambdaMetafactory.java:194)
[...]

PR #4919, opened 2019-02-27
Remove workaround in Access.defineClass()

keithc-ca:
* Add RAM class flag J9ClassIsExemptFromValidation which is set for MagicAccessorImpl and its subclasses which makes them exempt for some access and validation rules (like classes defined via unsafe API).
* Add and use macro: J9CLASS_IS_EXEMPT_FROM_VALIDATION.
* New flag, J9_FINDCLASS_FLAG_NAME_IS_INVALID, is used to signal an invalid class name if the actual loaded class is not exempt. If the class being defined is not exempt, defineClassCommon() will now return NULL without a pending exception. ClassLoader_defineClassImpl() throws an exception only in that case.
* Allow JavaLangAccess to define a class with a null ProtectionDomain.
* Map proxy classes to the proper module.
* defineClassCommon() now calls fixUnsafeMethods() as necessary
* fix loop start index in verifyQualifiedName()

Fixes: #3641
Fixes: #4661

### DIFF
--- a/jcl/src/java.base/share/classes/java/lang/ClassLoader.java
+++ b/jcl/src/java.base/share/classes/java/lang/ClassLoader.java
@@ -538,6 +538,18 @@ protected final Class<?> defineClass (
 		ProtectionDomain protectionDomain) 
 		throws java.lang.ClassFormatError 
 {
+	return defineClassInternal(className, classRep, offset, length, protectionDomain, false /* allowNullProtectionDomain */);
+}
+
+final Class<?> defineClassInternal(
+		final String className, 
+		final byte[] classRep, 
+		final int offset, 
+		final int length, 
+		ProtectionDomain protectionDomain,
+		boolean allowNullProtectionDomain)
+		throws java.lang.ClassFormatError 
+{
 	Certificate[] certs = null; 
 	if (protectionDomain != null) {
 		final CodeSource cs = protectionDomain.getCodeSource();
@@ -555,7 +567,7 @@ protected final Class<?> defineClass (
 		throw new ArrayIndexOutOfBoundsException();
 	}
 
-	if (protectionDomain == null)	{
+	if ((protectionDomain == null) && !allowNullProtectionDomain) {
 		protectionDomain = getDefaultProtectionDomain();
 	}
 	
--- a/runtime/jcl/common/clsldr.cpp
+++ b/runtime/jcl/common/clsldr.cpp
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 1998, 2018 IBM Corp. and others
+ * Copyright (c) 1998, 2019 IBM Corp. and others
  *
  * This program and the accompanying materials are made available under
  * the terms of the Eclipse Public License 2.0 which accompanies this
@@ -37,29 +37,43 @@ jboolean JNICALL Java_java_lang_ClassLoader_isVerboseImpl(JNIEnv *env, jclass cl
 	return ( (javaVM->verboseLevel & VERBOSE_CLASS) == VERBOSE_CLASS );
 }
 
-
-jclass JNICALL 
+jclass JNICALL
 Java_java_lang_ClassLoader_defineClassImpl(JNIEnv *env, jobject receiver, jstring className, jbyteArray classRep, jint offset, jint length, jobject protectionDomain)
 {
-#ifdef J9VM_OPT_DYNAMIC_LOAD_SUPPORT
-	if (className != NULL) {
-		J9VMThread *currentThread = (J9VMThread *)env;
-		J9JavaVM *vm = currentThread->javaVM;
-		J9InternalVMFunctions *vmFuncs = vm->internalVMFunctions;
+	J9VMThread *currentThread = (J9VMThread *)env;
+	J9InternalVMFunctions *vmFuncs = currentThread->javaVM->internalVMFunctions;
+	UDATA options = 0;
 
+#ifdef J9VM_OPT_DYNAMIC_LOAD_SUPPORT
+	if (NULL != className) {
 		vmFuncs->internalEnterVMFromJNI(currentThread);
 
 		if (CLASSNAME_INVALID == vmFuncs->verifyQualifiedName(currentThread, J9_JNI_UNWRAP_REFERENCE(className))) {
-			vmFuncs->setCurrentException(currentThread, J9VMCONSTANTPOOL_JAVALANGNOCLASSDEFFOUNDERROR, (UDATA *)*(j9object_t*)className);
-			vmFuncs->internalExitVMToJNI(currentThread);
-			return NULL;
+			/*
+			 * We don't yet know if the class being defined is exempt. Setting this option tells
+			 * defineClassCommon() to fail if it discovers that the class is not exempt. That failure
+			 * is distinguished by returning NULL with no exception pending.
+			 */
+			options |= J9_FINDCLASS_FLAG_NAME_IS_INVALID;
 		}
 
 		vmFuncs->internalExitVMToJNI(currentThread);
 	}
-#endif
+#endif /* J9VM_OPT_DYNAMIC_LOAD_SUPPORT */
 
-	return defineClassCommon(env, receiver, className, classRep, offset, length, protectionDomain, 0, NULL);
+	jclass result = defineClassCommon(env, receiver, className, classRep, offset, length, protectionDomain, options, NULL);
+
+	if (J9_ARE_ANY_BITS_SET(options, J9_FINDCLASS_FLAG_NAME_IS_INVALID) && (NULL == result) && (NULL == currentThread->currentException)) {
+		/*
+		 * Now that we know the class is not exempt, throw NoClassDefFoundError as we
+		 * would have liked to have done above before we called defineClassCommon().
+		 */
+		vmFuncs->internalEnterVMFromJNI(currentThread);
+		vmFuncs->setCurrentException(currentThread, J9VMCONSTANTPOOL_JAVALANGNOCLASSDEFFOUNDERROR, (UDATA *)*(j9object_t *)className);
+		vmFuncs->internalExitVMToJNI(currentThread);
+	}
+
+	return result;
 }
 
 jboolean JNICALL
--- a/runtime/oti/j9nonbuilder.h
+++ b/runtime/oti/j9nonbuilder.h
@@ -76,6 +76,7 @@
 #define J9ClassIsValueType 0x400
 #define J9ClassLargestAlignmentConstraintReference 0x800
 #define J9ClassLargestAlignmentConstraintDouble 0x1000
+#define J9ClassIsExemptFromValidation 0x2000
 
 /* @ddr_namespace: map_to_type=J9FieldFlags */
 
--- a/runtime/vm/createramclass.cpp
+++ b/runtime/vm/createramclass.cpp
@@ -149,7 +149,7 @@ static UDATA checkPackageAccess(J9VMThread *vmThread, J9Class *foundClass, UDATA
 static void setCurrentExceptionForBadClass(J9VMThread *vmThread, J9UTF8 *badClassName, UDATA exceptionIndex);
 static BOOLEAN verifyClassLoadingStack(J9VMThread *vmThread, J9ClassLoader *classLoader, J9ROMClass *romClass);
 static void popFromClassLoadingStack(J9VMThread *vmThread);
-static VMINLINE BOOLEAN loadSuperClassAndInterfaces(J9VMThread *vmThread, J9ClassLoader *classLoader, J9ROMClass *romClass, J9Class *elementClass, UDATA packageID, BOOLEAN hotswapping, UDATA classPreloadFlags, J9Class **superclassOut, J9Module *module);
+static VMINLINE BOOLEAN loadSuperClassAndInterfaces(J9VMThread *vmThread, J9ClassLoader *classLoader, J9ROMClass *romClass, UDATA options, J9Class *elementClass, UDATA packageID, BOOLEAN hotswapping, UDATA classPreloadFlags, J9Class **superclassOut, J9Module *module);
 static J9Class* internalCreateRAMClassDropAndReturn(J9VMThread *vmThread, J9ROMClass *romClass, J9CreateRAMClassState *state);
 static J9Class* internalCreateRAMClassDoneNoMutex(J9VMThread *vmThread, J9ROMClass *romClass, UDATA options, J9CreateRAMClassState *state);
 static J9Class* internalCreateRAMClassDone(J9VMThread *vmThread, J9ClassLoader *classLoader, J9ROMClass *romClass, UDATA options, J9Class *elementClass,
@@ -172,6 +172,16 @@ static I_32 interfaceDepthCompare(const void *a, const void *b);
 static void checkForCustomSpinOptions(void *element, void *userData);
 #endif /* J9VM_INTERP_CUSTOM_SPIN_OPTIONS */
 static void trcModulesSettingPackage(J9VMThread *vmThread, J9Class *ramClass, J9ClassLoader *classLoader, J9UTF8 *className);
+
+/*
+ * A class which extends (perhaps indirectly) the 'magic'
+ * accessor class is exempt from the normal access rules.
+ */
+#if JAVA_SPEC_VERSION == 8
+#define MAGIC_ACCESSOR_IMPL "sun/reflect/MagicAccessorImpl"
+#else /* JAVA_SPEC_VERSION == 8 */
+#define MAGIC_ACCESSOR_IMPL "jdk/internal/reflect/MagicAccessorImpl"
+#endif /* JAVA_SPEC_VERSION == 8 */
 
 /**
  * Mark all of the interfaces supported by this class, including all interfaces
@@ -1519,11 +1529,11 @@ popFromClassLoadingStack(J9VMThread *vmThread)
  * appropriate Java error on the VM.
  */
 static VMINLINE BOOLEAN
-loadSuperClassAndInterfaces(J9VMThread *vmThread, J9ClassLoader *classLoader, J9ROMClass *romClass, J9Class *elementClass,
+loadSuperClassAndInterfaces(J9VMThread *vmThread, J9ClassLoader *classLoader, J9ROMClass *romClass, UDATA options, J9Class *elementClass,
 	UDATA packageID, BOOLEAN hotswapping, UDATA classPreloadFlags, J9Class **superclassOut, J9Module *module)
 {
 	J9JavaVM *vm = vmThread->javaVM;
-	const BOOLEAN isROMClassUnsafe = (J9ROMCLASS_IS_UNSAFE(romClass) != 0);
+	BOOLEAN isExemptFromValidation = J9_ARE_ANY_BITS_SET(options, J9_FINDCLASS_FLAG_UNSAFE);
 	J9UTF8 *className = J9ROMCLASS_CLASSNAME(romClass);
 	J9UTF8 *superclassName = NULL;
 	J9Class *superclass = NULL;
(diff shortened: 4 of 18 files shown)


runtime/oti/j9javaaccessflags.h
```diff
@@ -67,7 +67,7 @@
 #define J9AccClassReferenceSoft 0x20000000
 #define J9AccClassReferenceWeak 0x10000000
 #define J9AccClassRomToRamMask 0xF3000000
-#define J9AccClassUnsafe 0x40000
+#define J9AccClassUnsafe 0x40000 /* obsolete - do not use */
```
gacholio: Typcially we define J9AccClassUnused0x4000 or something like that. If you haven't deleted the constant, you can't prove you found all the references.

runtime/vm/KeyHashTable.c
```diff
@@ -447,7 +446,7 @@ hashPkgTableIDFor(J9VMThread *vmThread, J9ClassLoader *classLoader, J9ROMClass *
 
 	key.tag = (UDATA)romClass | TAG_ROM_CLASS;
 
-	if (isSystemClassLoader && (J9ROMCLASS_IS_UNSAFE(romClass) || (LOAD_LOCATION_UNKNOWN == locationType))) {
+	if (isSystemClassLoader && (LOAD_LOCATION_UNKNOWN == locationType)) {
```
gacholio: This change is not equivalent code.

runtime/vm/createramclass.cpp
```diff
@@ -3010,6 +3028,18 @@ internalCreateRAMClassFromROMClass(J9VMThread *vmThread, J9ClassLoader *classLoa
...
 			classFlags |= J9ClassIsAnonymous;
 		}
+		if (J9_ARE_ANY_BITS_SET(options, J9_FINDCLASS_FLAG_UNSAFE)) {
+			classFlags |= J9ClassIsExemptFromValidation;
+		}
+		if (J9_ARE_NO_BITS_SET(classFlags, J9ClassIsExemptFromValidation)) {
+			const U_16 magicLength = sizeof(MAGIC_ACCESSOR_IMPL) - 1; /* exclude NUL */
+			J9UTF8 *className = J9ROMCLASS_CLASSNAME(romClass);
+
+			if ((magicLength == J9UTF8_LENGTH(className)) &&
```
gacholio: We have macros for this - `J9UTF8_LITERAL_EQUALS(data1, length1, cString)`

gacholio:
Let's make `(J9_ARE_NO_BITS_SET(resultClass->classFlags, J9ClassIsExemptFromValidation)` into a macro, which would make many of your unnecessary formatting changes even more unnecessary.

runtime/oti/j9javaaccessflags.h
keithc-ca: It's still referenced in DDR code for reading older core files.

runtime/vm/KeyHashTable.c
keithc-ca: I realize that; it's not clear what better we can do given the 'unsafe' bit is no longer available (and wasn't entirely reliable).

keithc-ca:
I added `J9CLASS_IS_EXEMPT_FROM_VALIDATION` in `j9.h` and used it as suggested.
I also made use of `J9UTF8_LITERAL_EQUALS`.

runtime/jcl/common/clsldr.cpp
```diff
@@ -30,36 +30,42 @@
...
 	return ( (javaVM->verboseLevel & VERBOSE_CLASS) == VERBOSE_CLASS );
 }
 
-
-jclass JNICALL 
+jclass JNICALL
 Java_java_lang_ClassLoader_defineClassImpl(JNIEnv *env, jobject receiver, jstring className, jbyteArray classRep, jint offset, jint length, jobject protectionDomain)
 {
+	J9VMThread *currentThread = (J9VMThread *)env;
+	jclass result = defineClassCommon(env, receiver, className, classRep, offset, length, protectionDomain, 0, NULL);
```
DanHeidinga: Why did you refactor this code to define the class before doing the verification of the name?  Once the class has been defined, it gets added to classloader tables and the like - I would expect other threads to be able to find it as though it was successfully defined even though it might fail below now.

runtime/vm/createramclass.cpp
```diff
@@ -1519,11 +1529,11 @@ popFromClassLoadingStack(J9VMThread *vmThread)
  * appropriate Java error on the VM.
  */
 static VMINLINE BOOLEAN
-loadSuperClassAndInterfaces(J9VMThread *vmThread, J9ClassLoader *classLoader, J9ROMClass *romClass, J9Class *elementClass,
+loadSuperClassAndInterfaces(J9VMThread *vmThread, J9ClassLoader *classLoader, J9ROMClass *romClass, UDATA options, J9Class *elementClass,
 	UDATA packageID, BOOLEAN hotswapping, UDATA classPreloadFlags, J9Class **superclassOut, J9Module *module)
 {
 	J9JavaVM *vm = vmThread->javaVM;
-	const BOOLEAN isROMClassUnsafe = (J9ROMCLASS_IS_UNSAFE(romClass) != 0);
+	BOOLEAN isExemptFromValidation = J9_ARE_ANY_BITS_SET(options, J9_FINDCLASS_FLAG_UNSAFE);
```
DanHeidinga: why was const removed?

runtime/vm/createramclass.cpp
```diff
@@ -1571,22 +1580,28 @@ loadSuperClassAndInterfaces(J9VMThread *vmThread, J9ClassLoader *classLoader, J9
...
 				 * the same package as the class being loaded.
 				 * 2) The superClass class is public but the class being loaded
 				 * belongs to a module that doesn't have access to the module that
 				 * owns the superClass class.
 				 */
 				bool superClassIsPublic = J9_ARE_ALL_BITS_SET(superclass->romClass->modifiers, J9AccPublic);
-				if ((!superClassIsPublic && (packageID != superclass->packageID))
-					|| (superClassIsPublic && (J9_VISIBILITY_ALLOWED != checkModuleAccess(vmThread, vm, romClass, module, superclass->romClass, superclass->module, superclass->packageID, 0)))
+				if (superClassIsPublic
+						? (J9_VISIBILITY_ALLOWED != checkModuleAccess(vmThread, vm, romClass, module, superclass->romClass, superclass->module, superclass->packageID, 0))
```
DanHeidinga: Please revert this refactoring - embedding a ternary if inside an if statement is painful from a readability perspective.  Repeating the single boolean check is fine if the compiler isn't smart enough to remove it

runtime/vm/createramclass.cpp
```diff
@@ -1614,18 +1629,19 @@ loadSuperClassAndInterfaces(J9VMThread *vmThread, J9ClassLoader *classLoader, J9
...
 						 * the same package as the class being loaded.
 						 * 2) The interface class is public but the class being loaded
 						 * belongs to a module that doesn't have access to the module that
 						 * owns the interface class.
 						 */
 						bool interfaceIsPublic = J9_ARE_ALL_BITS_SET(interfaceClass->romClass->modifiers, J9AccPublic);
-						if ((!interfaceIsPublic && (packageID != interfaceClass->packageID))
-							|| (interfaceIsPublic && (J9_VISIBILITY_ALLOWED != checkModuleAccess(vmThread, vm, romClass, module, interfaceClass->romClass, interfaceClass->module, interfaceClass->packageID, 0)))
+						if (interfaceIsPublic
+								? (J9_VISIBILITY_ALLOWED != checkModuleAccess(vmThread, vm, romClass, module, interfaceClass->romClass, interfaceClass->module, interfaceClass->packageID, 0))
```
DanHeidinga: Same here - embedding ternary ifs in if statements is a readability concern - please revert this refactoring

runtime/jcl/common/clsldr.cpp
keithc-ca: Whether the class is exempt from the usual name rules is dependent on whether it extends `MagicAccessorImpl`: that information isn't available before the call to `
defineClassCommon`. I think you're right, though, if the name is bad, it ought to be removed from those tables.

runtime/vm/createramclass.cpp
keithc-ca: It will be set below if the superclass is exempt.

runtime/vm/createramclass.cpp
keithc-ca: I disagree: I find it clearer and easier to understand this way. I will revert if you insist.

runtime/jcl/common/clsldr.cpp
keithc-ca: Those class cannot be found because `Class.forName()` also makes use of `verifyQualifiedName()`.

DanHeidinga: Need to check the resolveClassRef path as well

keithc-ca: I've restored validation of the name before the call, but deferred throwing the exception until we know the class is not exempt.

runtime/vm/createramclass.cpp
keithc-ca: I will revert this change (and the similar one below): we can return to this discussion later.

runtime/vm/KeyHashTable.c
keithc-ca: I've restored setting the 'unsafe' bit via that path to the tag bit can continue to be set in the same situations.

keithc-ca:
I believe I've addressed all the comments (so far) in a new branch (extending this one) and will push to this branch once I've verified it works as expected.

keithc-ca:
Ready for another round of reviews.

DanHeidinga:
Jenkins test sanity plinux,win jdk11

DanHeidinga:

DanHeidinga:

DanHeidinga:

jdekonin:
jenkins test sanity win jdk11

jdekonin:
jenkins test sanity xlinux jdk12

keithc-ca:
I'd like a chance to squash this before it's merged, please.

pshipton:
Technically the 0.13 PR should "fix" the listed Issues.

DanHeidinga:
Running the builds first as I'm reasonably certain this looks right but wanted a broad set of tests first.

@keithc-ca Will ping you to squash to before merge

DanHeidinga:
@keithc-ca I'm seeing some jvmti redefinition-related test failures in the extended jdk8 build.  Can you take a look?

keithc-ca:
I am reworking the changes to `clsldr.cpp` and `createramclass.cpp` to restore the exception behavior as expected by the failing test (and to match the RI).

keithc-ca:
I have the right exception behavior; now looking at the JVMTI redefinition issue.

gacholio:
Another thing I thought of over the weekend - classes which extend a class defined via Unsafe do not currently inherit the validation exclusion. This case is unlikely, but I think it's definitely possible.

keithc-ca:
The redefinition issue should now be fixed.

keithc-ca:

gacholio:

keithc-ca:
I've reworked this so classes which extend those defined via unsafe API aren't exempted from the normal validation rules. After running local testing, I'll push to this branch.

gacholio:
Sanity build failed because master is unstable. Will try again later.

gacholio:
OK, I'll try testing again after review.

gacholio:
May be worth rebasing on master after your latest changes to avoid build issues.

keithc-ca:
Testing went well, so I squashed and rebased. Hopefully this is in its final state.

pshipton:
jenkins test sanity zlinux jdk8,jdk12

pshipton:

pshipton:
jenkins test extended plinux jdk8,jdk11,jdk12

pshipton:
jenkins test sanity win jdk8

pshipton:
The jdk12 sanity testing is failing.

runtime/vm/createramclass.cpp
```diff
@@ -1584,15 +1594,26 @@ loadSuperClassAndInterfaces(J9VMThread *vmThread, J9ClassLoader *classLoader, J9
...
+						isExemptFromValidation = TRUE;
+					} else {
+						Trc_VM_CreateRAMClassFromROMClass_superclassNotVisible(vmThread, superclass, superclass->classLoader, classLoader);
+						setCurrentExceptionForBadClass(vmThread, superclassName, J9VMCONSTANTPOOL_JAVALANGILLEGALACCESSERROR);
+						return FALSE;
+					}
 				}
 			}
 
+			if (J9_ARE_ANY_BITS_SET(options, J9_FINDCLASS_FLAG_NAME_IS_INVALID) && !isExemptFromValidation) {
```
gacholio: Should this be setting the bit and letting the caller deal with the exception? The exception below is a much better one than is being thrown in the new calling code.

keithc-ca: When you say 'below', do you mean the one in `clsldr.cpp`? Just returning `FALSE` here with no exception seems to work (and is reasonable given the caller has claimed responsibility for dealing with the invalid name).

gacholio: "below" I mean the special exception setter call which has good detail info. The one being thrown outside is bare (no message). I'd move the good call out to the caller and check there (which you indicate works).

keithc-ca: I spoke too soon: calling `ClassLoader.loadClass()` with a `/` name referring to an already loaded class yields a `LinkageError` instead of the more specific `NoClassDefFoundError`. I'll see if there's a straight-forward fix for that.

gacholio: Strictly speaking, that's fine - you are throwing a LinkageError. Unless there's a test querying the exact class of the exception, I think that should be fine.

Comment welcome from @DanHeidinga @pshipton

DanHeidinga: While GAC is likely technically correct, the javadoc for loadClass() only mentions `ClassNotFoundException` so I expect there's a bunch of code that knows to map `NoClassDefFoundError` -> `ClassNotFoundException` and may not handle LinkageError correctly

keithc-ca: OpenJ9 has at least one test that expects `NoClassDefFoundError` and considers the more general `LinkageError` a failure: that expected behavior is consistent with the RI.

keithc-ca: I only know of the reverse mapping (`ClassNotFoundException` -> `NoClassDefFoundError`) that used to be involved in the compiled code for expressions of the form `Whatever.class` before class constants were a thing in class files: the compiler knew about the mentioned class and so generates code to catch `ClassNotFoundException` (a checked exception) thrown by `Class.forName()` and map it to `NoClassDefFoundError`.

keithc-ca:
This version passes all sanity testing in both jdk11 and jdk12.

pshipton:
jenkins test extended plinux jdk11,jdk12

gacholio:
jenkins test all xlinux jdk8

gacholio:
jenkins test sanity zlinux jdk11,jdk12

gacholio:
jenkins compile win jdk12

sourcetools/j9constantpool/com/ibm/oti/VMCPTool/ConstantPoolStream.java
```diff
@@ -329,7 +329,7 @@ private void writeHeader(int size) {
 			out.println("} _jclROMClass = {");
 
 			String shapeDescriptionOffset = "sizeof(J9ROMClass) - offsetof(J9ROMClass, cpShapeDescription) + " + size;
-			out.println("\t{0,0,0,0,0,J9AccClassUnsafe, 0,0,0,0,0,0,0,0, " + itemCount + ", " + itemCount + ", 0,0,0, " + shapeDescriptionOffset + ", 0,},");
+			out.println("\t{0,0,0,0,0,0, 0,0,0,0,0,0,0,0, " + itemCount + ", " + itemCount + ", 0,0,0, " + shapeDescriptionOffset + ", 0,},");
```
gacholio: This change is unnecessary now (since we're keeping unsafe).

runtime/oti/j9nonbuilder.h
```diff
@@ -2879,7 +2880,6 @@ typedef struct J9ROMImageHeader {
 
 /* @ddr_namespace: map_to_type=J9ClassLocation */
 
-struct J9Class;
```
gacholio: Unnecessary.

runtime/jcl/common/stdinit.c
```diff
@@ -111,6 +111,9 @@ standardInit( J9JavaVM *vm, char *dllName)
 
 	TOC_STORE_TOC(vm->jclTOC, standardInit);
 
+	memset(&jclFakeClass, 0, sizeof(jclFakeClass));
+	/* the fake class must be exempt because it references package-private classes */
+	jclFakeClass.classFlags = J9ClassIsExemptFromValidation;
```
gacholio: This should also be unnecessary since we're not removing unsafe.

gacholio:
The comments in the header are out-of-date (not removing unsafe).

I'd like @DanHeidinga to have a look at this before merge.

keithc-ca:
Description updated.

pshipton:
For the record, all the testing started today has passed.

runtime/vm/createramclass.cpp
```diff
@@ -2934,21 +2957,41 @@ internalCreateRAMClassFromROMClass(J9VMThread *vmThread, J9ClassLoader *classLoa
...
-					module = findModuleForPackage(vmThread, classLoader, J9UTF8_DATA(className), pkgNameLength);
-					omrthread_monitor_exit(javaVM->classLoaderModuleAndLocationMutex);
 				} else {
-					module = javaVM->unamedModuleForSystemLoader;
+					bool findModule = false;
+
+					if (classLoader != javaVM->systemClassLoader) {
+						findModule = true;
+					} else if ((LOAD_LOCATION_PATCH_PATH == locationType)
+							|| (LOAD_LOCATION_MODULE == locationType)) {
```
DanHeidinga: Formatting:
```c
} else if ((LOAD_LOCATION_PATCH_PATH == locationType)
|| (LOAD_LOCATION_MODULE == locationType)
) {
        ........
}
```

runtime/vm/createramclass.cpp
```diff
@@ -2934,21 +2957,41 @@ internalCreateRAMClassFromROMClass(J9VMThread *vmThread, J9ClassLoader *classLoa
...
+					} else {
+						J9UTF8 *superclassName = J9ROMCLASS_SUPERCLASSNAME(romClass);
+
+						if ((NULL != superclassName)
+						&& J9UTF8_LITERAL_EQUALS(J9UTF8_DATA(superclassName), J9UTF8_LENGTH(superclassName), "java/lang/reflect/Proxy")
+						) {
+							/*
+							 * Proxy classes loaded by the system classloader may belong to a dynamically
+							 * created module (which has been granted access to the interfaces it must implement).
+							 */
```
DanHeidinga: is this a bug fix for something beyond the Access.defineClass issue?  If so, please add it to the issue description

DanHeidinga:
overall, this looks good to me.  Was there something you specific you wanted me to look at @gacholio ?

gacholio:
> overall, this looks good to me. Was there something you specific you wanted me to look at @gacholio ?

Nope, just wanted to make sure you agreed with the overall strategy.

runtime/vm/createramclass.cpp
keithc-ca: Fixed.

runtime/jcl/common/stdinit.c
keithc-ca: Reverted.

runtime/oti/j9nonbuilder.h
keithc-ca: Reverted.

sourcetools/j9constantpool/com/ibm/oti/VMCPTool/ConstantPoolStream.java
keithc-ca: Reverted.

jcl/src/java.base/share/classes/java/lang/ClassLoader.java
```diff
@@ -537,6 +537,18 @@ private String checkClassName(String className) {
 		final int length, 
 		ProtectionDomain protectionDomain) 
 		throws java.lang.ClassFormatError 
+{
+	return defineClassInternal(className, classRep, offset, length, protectionDomain, false);
```
gacholio: Extreme nitpick - code above uses a variable for the boolean, this code does not.

gacholio:
Some local test failures are being investigated before this can be merged.

jcl/src/java.base/share/classes/java/lang/ClassLoader.java
gacholio: I'd personally prefer an inline value with a comment to minimize bytecode size when running in the interpreter.

keithc-ca: Comments added at both call sites.

keithc-ca:
None of those local test failures are introduced by this PR.

PR #4926, opened 2019-02-28
Remove deprecated unsigned Constant IL Opcodes

PeXXeR:
Get rid of deprecated opcodes.

https://github.com/eclipse/omr/issues/2657

### DIFF
--- a/runtime/compiler/optimizer/SwitchAnalyzer.cpp
+++ b/runtime/compiler/optimizer/SwitchAnalyzer.cpp
@@ -1224,7 +1224,7 @@ TR::Block *TR::SwitchAnalyzer::addGotoBlock(TR::TreeTop *dest)
    }
 TR::Block *TR::SwitchAnalyzer::addIfBlock(TR::ILOpCodes opCode, CASECONST_TYPE val, TR::TreeTop *dest)
    {
-   TR::Node *constNode = TR::Node::create(_switch, _isInt64 ? (_signed ? TR::lconst : TR::luconst) : (_signed ? TR::iconst : TR::iuconst), 0);
+   TR::Node *constNode = TR::Node::create(_switch, _isInt64 ? TR::lconst : TR::iconst, 0);
    constNode->set64bitIntegralValue(val);
    TR::Node *node = TR::Node::createif(opCode,
                                      TR::Node::createLoad(_switch, _temp),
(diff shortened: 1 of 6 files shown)


0xdaryl:
@jdmpapin, I'd appreciate a review from you on this as well please.

andrewcraik:
Given the large amount of unsigned operations in the SwitchAnalyzer I would prefer that the opt was cleaned up in its entirety and, if appropriate, _signed removed or usage made clear rather than piecemeal deleting parts of the code checking _signed without looking at the whole.

0xdaryl:
@andrewcraik, I think that's a reasonable approach.  Are there other optimizations you'd like to see done this way?

andrewcraik:
Not that spring to mind - I'd have to go audit the code. I think my rule would that if the signedness / unsignedness pervades the optimization then we should clean the opt as a while. If, however, like IdiomRecognition or TreeSimplification the unsignedness is just in contained parts of the opt I'd be fine with those parts being cleaned up as the opcodes are removed. @jdmpapin feel free to jump in.

runtime/compiler/optimizer/SwitchAnalyzer.cpp
```diff
@@ -1269,10 +1269,6 @@ TR::Block *TR::SwitchAnalyzer::addTableBlock(SwitchInfo *dense)
       node->setAndIncChild(0, TR::Node::create(TR::isub, 2,
                                            (_isInt64 ? TR::Node::create(TR::l2i, 1, TR::Node::createLoad(_switch, _temp)) : TR::Node::createLoad(_switch, _temp)),
                                            TR::Node::create(_switch, TR::iconst, 0, dense->_min)));
-   else
-      node->setAndIncChild(0, TR::Node::create(TR::iusub, 2,
-                                           (_isInt64 ? TR::Node::create(TR::l2i, 1, TR::Node::createLoad(_switch, _temp)) : TR::Node::createLoad(_switch, _temp)),
-                                           TR::Node::create(_switch, TR::iuconst, 0, dense->_min)));
```
jdmpapin: The corresponding code just above shouldn't be guarded by `_signed` anymore. The only difference between the `_signed` and `!_signed` cases was the choice of `isub` vs. `iusub` and of `iconst` vs. `iuconst`. Substituting in the sign-agnostic opcodes for the unsigned ones, both branches are the same, so the one currently under `if (_signed)` should run whether or not `_signed` is set

jdmpapin:
While it might be a good idea to check the meaning/usage of `_signed` in `SwitchAnalyzer`, I'm not sure that should block the removal of these opcodes. Substituting in the sign-agnostic opcodes doesn't change the behaviour of the generated trees, and if there is any problem with `_signed`, that problem should be unrelated to this change

PeXXeR:
@jdmpapin and @andrewcraik Thanks a lot for reviewing the code changes.
I will proceed with complete set of changes for only SwitchAnalyzer as it contains a lot of signedness stuff.
Also, I am closing this PR here and will create new one with fresh changes which will include the aobve review comments as well.
Since, SwitchAnalyzer.cpp has also been removed from OpenJ9 now and getting used from OMR, it seems better to make changes which comprises all above considerations and comments.

PR #4929, opened 2019-02-28
Fix instantiating valueTypes

MarkQingGuo:
Remove flag J9_RESOLVE_FLAG_INSTANTIABLE in  defaultValue bytecode

- Remove flag J9_RESOLVE_FLAG_INSTANTIABLE in  function resolveClassRef()

### DIFF
--- a/runtime/vm/BytecodeInterpreter.hpp
+++ b/runtime/vm/BytecodeInterpreter.hpp
@@ -8252,7 +8252,7 @@ retry:
 		updateVMStruct(REGISTER_ARGS);
 
 		if (NULL == resolvedClass) {
-			resolveClassRef(_currentThread, ramConstantPool, index, J9_RESOLVE_FLAG_RUNTIME_RESOLVE | J9_RESOLVE_FLAG_INIT_CLASS | J9_RESOLVE_FLAG_INSTANTIABLE);
+			resolveClassRef(_currentThread, ramConstantPool, index, J9_RESOLVE_FLAG_RUNTIME_RESOLVE | J9_RESOLVE_FLAG_INIT_CLASS);
 		} else if (!J9_IS_J9CLASS_VALUETYPE(resolvedClass)) {
 			J9UTF8 *badClassName = J9ROMCLASS_CLASSNAME(resolvedClass->romClass);
 			setCurrentExceptionNLSWithArgs(_currentThread, J9NLS_VM_ERROR_BYTECODE_CLASSREF_MUST_BE_VALUE_TYPE, J9VMCONSTANTPOOL_JAVALANGINCOMPATIBLECLASSCHANGEERROR, J9UTF8_LENGTH(badClassName), J9UTF8_DATA(badClassName));


MarkQingGuo:
@tajila please review.

runtime/vm/resolvesupport.cpp
```diff
@@ -259,7 +259,14 @@ resolveClassRef(J9VMThread *vmStruct, J9ConstantPool *ramCP, UDATA cpIndex, UDAT
 	ramClassRefWrapper = (J9RAMClassRef *)&ramCP[cpIndex];
 	resolvedClass = ramClassRefWrapper->value;
 	/* If resolving for "new", check if the class is instantiable */
-	if ((NULL != resolvedClass) && (J9_ARE_NO_BITS_SET(resolveFlags, J9_RESOLVE_FLAG_INSTANTIABLE) || J9ROMCLASS_ALLOCATES_VIA_NEW(resolvedClass->romClass))) {
+	if ((NULL != resolvedClass) && (J9_ARE_NO_BITS_SET(resolveFlags, J9_RESOLVE_FLAG_INSTANTIABLE) ||
+#if defined(J9VM_OPT_VALHALLA_VALUE_TYPES)
+                (J9ROMCLASS_ALLOCATES_VIA_NEW(resolvedClass->romClass) || 
+                 (J9_IS_J9CLASS_VALUETYPE(resolvedClass) && J9_ARE_ANY_BITS_SET(resolvedFlags, J9_RESOLVE_FLAG_VALUETYPE)))
```
tajila: use J9_ARE_ALL_BITS_SET

tajila: also the || should be on the next line. Look at the coding guidelines to see how to do this properly

runtime/vm/resolvesupport.cpp
```diff
@@ -398,7 +405,11 @@ resolveClassRef(J9VMThread *vmStruct, J9ConstantPool *ramCP, UDATA cpIndex, UDAT
 
 	accessModifiers = resolvedClass->romClass->modifiers;
 	if (J9_ARE_ANY_BITS_SET(resolveFlags, J9_RESOLVE_FLAG_INSTANTIABLE)) {
-		if (!J9ROMCLASS_ALLOCATES_VIA_NEW(resolvedClass->romClass)) {
+		if (!J9ROMCLASS_ALLOCATES_VIA_NEW(resolvedClass->romClass)
+#if defined(J9VM_OPT_VALHALLA_VALUE_TYPES)
+                || (J9_IS_J9CLASS_VALUETYPE(resolvedClass) && J9_ARE_NO_BITS_SET(resolveFlags, J9_RESOLVE_FLAG_VALUETYPE))
```
tajila: I think there is still an issue with this. 

The test write now says:
`if (notInstantiableViaNew || (isValueType && !resolveValueType)`

The problem is a valueType is not instantiable via new so the first part of the condition is always true

tajila:
The simple solution is to remove the `J9_RESOLVE_FLAG_INSTANTIABLE` flag from the defaultValue bytecode. This way we don't need to add another resolve flag or change any of the existing resolve code

tajila:
@gacholio please review this change

gacholio:
Too trivial to waste test resource on.

PR #4931, opened 2019-02-28
Function implemented for TR_ResolvedRelocatableJ9JITaaSServerMethod

satbir-ca:
Functions implemented
1. getResolvedImproperInterfaceMethod
2. handleUnresolvedStaticMethodInCP
3. handleUnresolvedSpecialMethodInCP
4. handleUnresolvedVirtualMethodInCP
5. fieldAttributes
6. staticAttributes

small bugfix in the TR_ResolvedJ9JITaaSServerMethod::
getResolvedImproperInterfaceMethod

### DIFF
--- a/runtime/compiler/control/JITaaSCompilationThread.cpp
+++ b/runtime/compiler/control/JITaaSCompilationThread.cpp
@@ -1597,6 +1597,54 @@ bool handleServerMessage(JITaaS::J9ClientStream *client, TR_J9VM *fe)
          client->write(clazz, definingClass, classChain);
          }
          break;
+      case J9ServerMessageType::ResolvedRelocatableMethod_fieldAttributes:
+         {
+         UDATA ltype = 0;
+         J9ROMFieldShape * fieldShape = 0;
+         auto recv = client->getRecvData<J9Method *, TR_ResolvedJ9Method*, int32_t, bool, J9ConstantPool *>();
+         auto ramMethod = std::get<0>(recv);
+         TR_ResolvedJ9Method *method= std::get<1>(recv);
+         auto cpIndex = std::get<2>(recv);
+         bool isStore = std::get<3>(recv);
+         auto constantPool = std::get<4>(recv);
+
+         J9Class *definingClass = (J9Class *) compInfoPT->reloRuntime()->getClassFromCP(fe->vmThread(), fe->_jitConfig->javaVM, constantPool, cpIndex, false);
+         IDATA offset = jitCTResolveInstanceFieldRefWithMethod(fe->vmThread(), ramMethod, cpIndex, isStore, &fieldShape);
+         bool unresolvedInCP = method->getUnresolvedFieldInCP(cpIndex);
+         if (offset >= 0)
+            ltype = fieldShape->modifiers;
+         client->write((IDATA)offset, unresolvedInCP, ltype, definingClass);
+         }
+         break;
+      case J9ServerMessageType::ResolvedRelocatableMethod_getFieldType:
+         {
+         auto recv = client->getRecvData<int32_t, TR_ResolvedJ9Method *>();
+         auto cpIndex = std::get<0>(recv);
+         TR_ResolvedJ9Method *method= std::get<1>(recv);
+         UDATA ltype = getFieldType((J9ROMConstantPoolItem *)(method->romLiterals()), cpIndex);
+         client->write(ltype);
+         }
+         break;
+      case J9ServerMessageType::ResolvedRelocatableMethod_staticAttributes:
+         {
+         J9ROMFieldShape * fieldShape = 0;
+         UDATA ltype = 0;
+         auto recv = client->getRecvData<J9Method *, TR_ResolvedJ9Method*, int32_t, bool, J9ConstantPool *>();
+         auto ramMethod = std::get<0>(recv);
+         TR_ResolvedJ9Method *method= std::get<1>(recv);
+         int32_t cpIndex = std::get<2>(recv);
+         bool isStore = std::get<3>(recv);
+         auto constantPool = std::get<4>(recv);
+
+         J9Class *definingClass = (J9Class *) compInfoPT->reloRuntime()->getClassFromCP(fe->vmThread(), fe->_jitConfig->javaVM, constantPool, cpIndex, true);
+         void * offset = jitCTResolveStaticFieldRefWithMethod(fe->vmThread(), ramMethod, cpIndex, isStore, &fieldShape);
+         bool unresolvedInCP = method->getUnresolvedFieldInCP(cpIndex);
+         if (offset)
+            ltype = fieldShape->modifiers;
+         client->write(offset, unresolvedInCP, ltype, definingClass);
+         }
+         break;
+
       case J9ServerMessageType::CompInfo_isCompiled:
          {
          J9Method *method = std::get<0>(client->getRecvData<J9Method *>());
--- a/runtime/compiler/env/j9method.h
+++ b/runtime/compiler/env/j9method.h
@@ -73,6 +73,7 @@ static bool supportsFastJNI(TR_FrontEnd *fe)
 #endif
    }
 
+UDATA getFieldType(J9ROMConstantPoolItem * cp, I_32 cpIndex);
 inline char *nextSignatureArgument(char *currentArgument)
    {
    char *result = currentArgument;
@@ -596,6 +597,7 @@ public:
 
    virtual bool                  storeValidationRecordIfNecessary(TR::Compilation * comp, J9ConstantPool *constantPool, int32_t cpIndex, TR_ExternalRelocationTargetKind reloKind, J9Method *ramMethod, J9Class *definingClass=0);
 
+   static void                   setAttributeResult(bool, bool, uintptr_t, int32_t, int32_t, int32_t, TR::DataType *, bool *, bool *, bool *, void ** );
 protected:
    virtual TR_ResolvedMethod *   createResolvedMethodFromJ9Method(TR::Compilation *comp, int32_t cpIndex, uint32_t vTableSlot, J9Method *j9Method, bool * unresolvedInCP, TR_AOTInliningStats *aotStats);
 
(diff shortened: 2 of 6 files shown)


satbir-ca:
@mpirvu Please revise the code

runtime/compiler/env/j9methodServer.cpp
```diff
@@ -1935,8 +2010,233 @@ TR_ResolvedRelocatableJ9JITaaSServerMethod::getUnresolvedSpecialMethodInCP(I_32
...
+#endif
+      }
+
+   setAttributeResult(true, theFieldIsFromLocalClass, ltype, volatileFlag, finalFlag, privateFlag, type, volatileP, isFinal, isPrivate, address);
+
+   return theFieldIsFromLocalClass;
+   }
+
+void
+TR_ResolvedRelocatableJ9JITaaSServerMethod::setAttributeResult(bool isStaticField, bool result, UDATA ltype, I_32 myVolatile, I_32 myFinal, I_32 myPrivate, TR::DataType * type, bool * volatileP, bool * isFinal, bool * isPrivate, void ** fieldOffset)
```
mpirvu: This method looks entirely identical to TR_ResolvedRelocatableJ9Method version. Isn't better just to redirect there?

runtime/compiler/env/j9methodServer.cpp
```diff
@@ -27,6 +27,66 @@
 #include "control/MethodToBeCompiled.hpp"
 #include "control/JITaaSCompilationThread.hpp"
 #include "exceptions/DataCacheError.hpp"
+#include "env/VMAccessCriticalSection.hpp"
+
+static inline UDATA getFieldType(J9ROMConstantPoolItem * cp, I_32 cpIndex)
```
mpirvu: Duplicating code may get us in trouble if the other copy changes.

runtime/compiler/env/j9methodServer.cpp
```diff
@@ -650,7 +710,7 @@ TR_ResolvedJ9JITaaSServerMethod::getResolvedImproperInterfaceMethod(TR::Compilat
       if (j9method == nullptr)
          return nullptr;
       else
-         createResolvedMethodFromJ9Method(comp, cpIndex, 0, j9method, nullptr, nullptr, methodInfo);
+         return createResolvedMethodFromJ9Method(comp, cpIndex, 0, j9method, nullptr, nullptr, methodInfo);
```
dmitry-ten: That's a good catch. I wonder if it might improve performance.

runtime/compiler/env/j9methodServer.cpp
```diff
@@ -1935,8 +2010,233 @@ TR_ResolvedRelocatableJ9JITaaSServerMethod::getUnresolvedSpecialMethodInCP(I_32
...
+TR_ResolvedRelocatableJ9JITaaSServerMethod::fieldAttributes(TR::Compilation * comp, int32_t cpIndex, uint32_t * fieldOffset, TR::DataType * type, bool * volatileP, bool * isFinal, bool * isPrivate, bool isStore, bool * unresolvedInCP, bool needAOTValidation)
+   {
+   TR_ASSERT(cpIndex != -1, "cpIndex shouldn't be -1");
+
+   UDATA ltype;
+   I_32 volatileFlag = 0, finalFlag = 0, privateFlag = 0;
+
+   bool result = false;
+   bool theFieldIsFromLocalClass = false;
+   J9ConstantPool *constantPool = (J9ConstantPool *)(J9_CP_FROM_METHOD(ramMethod()));
```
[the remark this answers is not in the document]
dmitry-ten: Actually, just remembered that we cache constant pool pointer in the resolved method.
So you should do
`J9ConstantPool *constantPool = (J9ConstantPool *) literals()`

runtime/compiler/env/j9methodServer.cpp
```diff
@@ -27,6 +27,66 @@
...
+      default:// object
+        return (j9FieldType |= J9FieldFlagObject);
+      }
+
+   }
+
+// takes a J9FieldType and returns the DataTypes associated
+// along with the size.
+//
+static TR::DataType decodeType(U_32 inType)
```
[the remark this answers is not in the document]
satbir-ca: decodeType and getfieldType are both define in the CPP file. using them means we need to move to Header file. Since it is being used only here I don't think its is of use to move to header file.

runtime/compiler/env/j9methodServer.cpp
satbir-ca: setAttributeResult is a protected method of 
TR_ResolvedRelocatableJ9Method. 
TR_ResolvedRelocatableJ9JITaasServerMethod is derived from TR_ResolvedJ9JITaasServerMethod. Hence the class TR_ResolvedRelocatableJ9Method does not come in the hierarchy tree.

runtime/compiler/env/j9methodServer.cpp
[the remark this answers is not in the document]
mpirvu: If we remove the static qualifier the linker should do the job.

runtime/compiler/env/j9methodServer.cpp
mpirvu: Reading the code again, there is nothing related to the state of  `TR_ResolvedRelocatableJ9Method` so we could make this method static and public. The same goes for the other two static methods `decodeType` and `getFieldType`

runtime/compiler/env/j9methodServer.cpp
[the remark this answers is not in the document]
satbir-ca: will do that

runtime/compiler/env/j9methodServer.cpp
```diff
@@ -1935,8 +2010,233 @@ TR_ResolvedRelocatableJ9JITaaSServerMethod::getUnresolvedSpecialMethodInCP(I_32
...
+   bool resolveField = true;
+
+   _stream->write(JITaaS::J9ServerMessageType::ResolvedRelocatableMethod_ResolveInstanceFieldRefWithMethod, ramMethod(), getRemoteMirror(), cpIndex, isStore);
+   auto recv = _stream->read<IDATA, bool, UDATA >();
+   offset = std::get<0>(recv);
+   *unresolvedInCP = std::get<1>(recv);
+   ltype = std::get<2>(recv);
+
+      {
+      TR::VMAccessCriticalSection fieldAttributes(fej9());
```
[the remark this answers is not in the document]
satbir-ca: will remove this

satbir-ca:
updated

dmitry-ten:
Looks fine to me.
Could you test if it works though? If you rebase to the latest jitaas branch, you should be able to run `-version` with AOT and see if your change works.

mpirvu:
@ymanton Younes, please review and merge this item as appropriately. Thanks

satbir-ca:
I am seeing a crash. work on it

ymanton:
> I'm getting `libj9jit29.so: undefined symbol: getFieldType` by simply running `java -version`. If `getFieldType()` is defined as `inline`, its definition needs to be moved to the header file. If it's kept in the .`cpp`, it should not be marked as `inline`.

Good catch, `static inline` is fine in a .cpp file, just `inline` is not. @satbir-ca just remove the `inline` keyword, those functions are not suitable for headers.

satbir-ca:
updated the code. Fixed couple of issue but still seeing some crash on client side. Working on it.

satbir-ca:
@ymanton 
updated the code.
the compilation is successful.

runtime/compiler/control/JITaaSCompilationThread.cpp
```diff
@@ -1597,6 +1597,57 @@ bool handleServerMessage(JITaaS::J9ClientStream *client, TR_J9VM *fe)
...
+         IDATA offset = jitCTResolveInstanceFieldRefWithMethod(fe->vmThread(), ramMethod, cpIndex, isStore, &fieldShape);
+         bool unresolvedInCP = method->getUnresolvedFieldInCP(cpIndex);
+         if (offset >= 0)
+            ltype = fieldShape->modifiers;
+         client->write((IDATA)offset, unresolvedInCP, ltype, definingClass);
+         }
+         break;
+      case J9ServerMessageType::ResolvedRelocatableMethod_getFieldType:
+         {
+         UDATA ltype = 0;
```
ymanton: Can we delete this and define `ltype` below when we retrieve the field type?

runtime/compiler/control/JITaaSCompilationThread.cpp
```diff
@@ -1597,6 +1597,57 @@ bool handleServerMessage(JITaaS::J9ClientStream *client, TR_J9VM *fe)
...
+         auto recv = client->getRecvData<J9Method *, TR_ResolvedJ9Method*, int32_t, bool, J9ConstantPool *>();
+         auto ramMethod = std::get<0>(recv);
+         TR_ResolvedJ9Method *method= std::get<1>(recv);
+         int32_t cpIndex = std::get<2>(recv);
+         bool isStore = std::get<3>(recv);
+         auto constantPool = std::get<4>(recv);
+
+         J9Class *definingClass = (J9Class *) compInfoPT->reloRuntime()->getClassFromCP(fe->vmThread(), fe->_jitConfig->javaVM, constantPool, cpIndex, true);
+
```
ymanton: Can we get rid of this whitespace? It makes it easier to compare with the `fieldAttributes` message case since they are both similar.

runtime/compiler/control/JITaaSCompilationThread.cpp
```diff
@@ -1597,6 +1597,57 @@ bool handleServerMessage(JITaaS::J9ClientStream *client, TR_J9VM *fe)
...
+         void * offset = jitCTResolveStaticFieldRefWithMethod(fe->vmThread(), ramMethod, cpIndex, isStore, &fieldShape);
+
```
ymanton: Can we get rid of this whitespace? It makes it easier to compare with the fieldAttributes message case since they are both similar.

runtime/compiler/env/j9method.h
```diff
@@ -596,6 +597,7 @@ class TR_ResolvedRelocatableJ9Method : public TR_ResolvedJ9Method
 
    virtual bool                  storeValidationRecordIfNecessary(TR::Compilation * comp, J9ConstantPool *constantPool, int32_t cpIndex, TR_ExternalRelocationTargetKind reloKind, J9Method *ramMethod, J9Class *definingClass=0);
 
+   static void                          setAttributeResult(bool, bool, uintptr_t, int32_t, int32_t, int32_t, TR::DataType *, bool *, bool *, bool *, void ** );
```
ymanton: Can you fix the spacing so that `setAttributeResult` lines up with the adjacent methods?

runtime/compiler/env/j9methodServer.cpp
```diff
@@ -28,6 +28,7 @@
 #include "control/JITaaSCompilationThread.hpp"
 #include "exceptions/DataCacheError.hpp"
 
+
```
ymanton: Let's get rid of this whitespace change.

satbir-ca:
updated

Issue #4098, opened 2018-12-18
Windows OpenJ9 builds of JDK10/11 not working after jlink'ing from linux


pshipton:
Originally reported via https://github.com/AdoptOpenJDK/openjdk-build/issues/807

copied here:

Platform:
Windows

Architecture:
x64

Please describe the issue that you are having with our binaries and if possible how to recreate it.

I'm trying to jlink a windows JRE from my linux. Using both JDK10 and JDK11 builds (OpenJ9) this seems to produce a VM which is not healthy.

Running this on my linux box:

mduft@fril0041 /ssd/other/openjdk $ ./jdk-11.0.1+13-j9/bin/jlink --output test-jre --add-modules java.base,java.compiler,java.datatransfer,java.desktop,java.instrument,java.logging,java.management,java.naming,java.sql,java.xml,jdk.jdwp.agent,jdk.jdi,jdk.zipfs,jdk.jcmd --compress=2 --no-header-files --no-man-pages --module-path win64/jdk-11.0.1+13-j9/jmods:mlib

Then running the bin/java.exe from a windows 10:
```
F:\other\openjdk\test-jre>.\bin\java.exe -version
Unhandled exception
Type=Segmentation error vmState=0x00000000
Windows_ExceptionCode=c0000005 J9Generic_Signal=00000004 ExceptionAddress=0000000000000000 ContextFlags=0010005f
Handler1=00007FFFFB556020 Handler2=00007FF80525ECA0 DEPAddress=0000000000000000
RDI=000002A40F84F42E RSI=000002A40F84F42E RAX=000002A40F84F44B RBX=000002A410384300
RCX=000002A40F84F44B RDX=00000000000003C0 R8=000002A410384300 R9=00000000000007AE
R10=000000B724EFE350 R11=0000000000000000 R12=00000000000007AE R13=000002A40C2B9930
R14=000002A410383AD0 R15=000002A410383AD0
RIP=0000000000000000 RSP=000000B724EFE2D8 RBP=000000B724EFE3F0 GS=002B
FS=0053 ES=002B DS=002B
XMM0 000000000010000e (f: 1048590.000000, d: 5.180723e-318)
XMM1 656a624f2f676e61 (f: 795307648.000000, d: 3.421278e+180)
XMM2 0000000000000000 (f: 0.000000, d: 0.000000e+00)
XMM3 0000000000000000 (f: 0.000000, d: 0.000000e+00)
[...]

PR #4935, opened 2019-02-28
Load the zip native library earlier on Windows

ChengJin01:
The change is to get zip.dll loaded earlier
after successfully loading jimage.dll before
invoking GetModuleHandle on Windows

Close: #4098

### DIFF
--- a/runtime/jcl/common/stdinit.c
+++ b/runtime/jcl/common/stdinit.c
@@ -137,16 +137,6 @@ standardInit( J9JavaVM *vm, char *dllName)
 		goto _fail;
 	}
 
-	vmFuncs->internalAcquireVMAccess(vmThread);
-	result = (jint)initializeRequiredClasses(vmThread, dllName);
-
-	if (0 == result) {
-		result = vmFuncs->initializeHeapOOMMessage(vmThread);
-	}
-
-	/* CANNOT hold VM Access while calling registerBootstrapLibrary */
-	vmFuncs->internalReleaseVMAccess(vmThread);
-
 	/* Add the correct "zip" natives to the path:
 	 * Configuration	Load Zip
 	 *  Java_6			yes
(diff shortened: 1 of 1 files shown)


ChengJin01:
Already verified in personal builds on Java 8 & 11.

Reviewer: @pshipton

runtime/jcl/common/stdinit.c
```diff
@@ -155,21 +145,29 @@ standardInit( J9JavaVM *vm, char *dllName)
...
+	{
 		UDATA handle = 0;
 		result = (jint)vmFuncs->registerBootstrapLibrary(vm->mainThread, "zip", (J9NativeLibrary **)&handle, FALSE);
 	}
 #endif /* !J9VM_INTERP_MINIMAL_JCL */
 #endif /* J9VM_OPT_SIDECAR */
 
 	vmFuncs->internalAcquireVMAccess(vmThread);
+	
+	if (0 == result) {
```
pshipton: Could move this result check before the internalAcquireVMAccess() and put everything to the internalReleaseVMAccess() inside the block, since nothing is going to run if result != 0.

ChengJin01: Already updated the code as suggested above.

pshipton:
jenkins test sanity win jdk8

pshipton:

runtime/jcl/common/stdinit.c
```diff
@@ -155,62 +145,72 @@ standardInit( J9JavaVM *vm, char *dllName)
...
+					J9VMCONSTANTPOOL_JAVALANGTHREADDEATH,
+					J9_FINDKNOWNCLASS_FLAG_INITIALIZE | J9_FINDKNOWNCLASS_FLAG_NON_FATAL);
+				if (vmThread->currentException) {
+					result = JNI_ERR;
+				}
 			}
 		}
+		vmFuncs->internalReleaseVMAccess(vmThread);
+	} else {
+		goto _fail;
```
pshipton: This is not equivalent to the original code. Any one of the calls above could fail, and execution should not continue past here.

ChengJin01: I previously overlooked the failing cases inside the block. Already changed it back to the original one.

pshipton:

PR #4942, opened 2019-03-01
Max pause  TS

amicic:
For Metronome GC policy report max pause timestamp relative to the
beginning of the heartbeat, rather than relative to the beginning of a
run.


pshipton:
@amicic does this change need an update to the docs?

amicic:
It (attribute maxTimestampMs) does not seem to be documented https://www.ibm.com/support/knowledgecenter/en/SSYKE2_8.0.0/com.ibm.java.vm.80.doc/docs/mm_gc_pd_mgc_verbose.html.

Verbose GC for Metronome is not documented for OpenJDK variant at all.

All in all, no change is needed. It is also rather low-level piece of info - it's not a must to be added into docs.

dmitripivkine:

dmitripivkine:

PR #4955, opened 2019-03-04
Add ValueType support to the MonitorEnter and MonitorExit  bytecodes

MarkQingGuo:
we need to modify the behaviour of these two bytecodes when value types are involved:

- These two bytecodes should throw illegalMonitorStateException when they are used on valueTypes
- others remain the same for nonValueType object
- new nls message are added for the exception when these bytecodes are used on valuetypes
- new tests are added: monitorenter on valuetypes, monitorexit on valuetypes, monitorenter on nonvaluetypes, monitorexit on nonvaluetypes, and monitorenterandmonitorexit on nonvaluetypes

### DIFF
--- a/runtime/nls/j9vm/j9vm.nls
+++ b/runtime/nls/j9vm/j9vm.nls
@@ -1,5 +1,5 @@
 #
-# Copyright (c) 2000, 2018 IBM Corp. and others
+# Copyright (c) 2000, 2019 IBM Corp. and others
 #
 # This program and the accompanying materials are made available under
 # the terms of the Eclipse Public License 2.0 which accompanies this
@@ -1870,4 +1870,13 @@ J9NLS_VM_ERROR_BYTECODE_OBJECTREF_MUST_BE_VALUE_TYPE.sample_input_2=Foo
 J9NLS_VM_ERROR_BYTECODE_OBJECTREF_MUST_BE_VALUE_TYPE.explanation=The class has specified a bytecode operation that can only be performed on a value type but the object on stack is not a value type.
 J9NLS_VM_ERROR_BYTECODE_OBJECTREF_MUST_BE_VALUE_TYPE.system_action=The JVM will throw a IncompatibleClassChangeError.
 J9NLS_VM_ERROR_BYTECODE_OBJECTREF_MUST_BE_VALUE_TYPE.user_response=Contact the provider of the classfile for a corrected version.
+# END NON-TRANSLATABLE
+
+J9NLS_VM_ERROR_BYTECODE_OBJECTREF_CANNOT_BE_VALUE_TYPE=bad object type %2$.*1$s: object that is synchronized is a value type
+# START NON-TRANSLATABLE
+J9NLS_VM_ERROR_BYTECODE_OBJECTREF_CANNOT_BE_VALUE_TYPE.sample_input_1=3
+J9NLS_VM_ERROR_BYTECODE_OBJECTREF_CANNOT_BE_VALUE_TYPE.sample_input_2=Foo
+J9NLS_VM_ERROR_BYTECODE_OBJECTREF_CANNOT_BE_VALUE_TYPE.explanation=The class has specified the bytecode monitorenter or monitorexit operation which cannot be performed on a value type but the object on stack is a value type.
+J9NLS_VM_ERROR_BYTECODE_OBJECTREF_CANNOT_BE_VALUE_TYPE.system_action=The JVM will throw an IllegalMonitorStateException.
+J9NLS_VM_ERROR_BYTECODE_OBJECTREF_CANNOT_BE_VALUE_TYPE.user_response=Contact the provider of the classfile for a corrected version.
 # END NON-TRANSLATABLE
--- a/runtime/vm/BytecodeInterpreter.hpp
+++ b/runtime/vm/BytecodeInterpreter.hpp
@@ -7605,28 +7605,42 @@ done:
 		if (NULL == obj) {
 			rc = THROW_NPE;
 		} else {
-			IDATA monitorRC = enterObjectMonitor(REGISTER_ARGS, obj);
-			/* Monitor enter can only fail in the nonblocking case, which does not
-			 * release VM access, so the immediate async and failed enter cases are
-			 * mutually exclusive.
-			 */
-			if (0 == monitorRC) {
-				rc = THROW_MONITOR_ALLOC_FAIL;
-			} else {
-				if (J9_UNEXPECTED(!VM_ObjectMonitor::recordBytecodeMonitorEnter(_currentThread, (j9object_t)monitorRC, _arg0EA))) {
-					objectMonitorExit(_currentThread, obj);
+#if defined(J9VM_OPT_VALHALLA_VALUE_TYPES)
+			J9Class * objClass = J9OBJECT_CLAZZ(_currentThread, obj);
+			if (J9_IS_J9CLASS_VALUETYPE(objClass)) {
+				J9UTF8 *badClassName = J9ROMCLASS_CLASSNAME(objClass->romClass);
+				buildInternalNativeStackFrame(REGISTER_ARGS);
+				updateVMStruct(REGISTER_ARGS);
+				prepareForExceptionThrow(_currentThread);
+				setCurrentExceptionNLSWithArgs(_currentThread, J9NLS_VM_ERROR_BYTECODE_OBJECTREF_CANNOT_BE_VALUE_TYPE, J9VMCONSTANTPOOL_JAVALANGILLEGALMONITORSTATEEXCEPTION, J9UTF8_LENGTH(badClassName), J9UTF8_DATA(badClassName));
+				VMStructHasBeenUpdated(REGISTER_ARGS);
+				rc = GOTO_THROW_CURRENT_EXCEPTION;
+			} else
+#endif /* J9VM_OPT_VALHALLA_VALUE_TYPES */
+			{
+				IDATA monitorRC = enterObjectMonitor(REGISTER_ARGS, obj);
+				/* Monitor enter can only fail in the nonblocking case, which does not
+				 * release VM access, so the immediate async and failed enter cases are
+				 * mutually exclusive.
+				 */
+				if (0 == monitorRC) {
 					rc = THROW_MONITOR_ALLOC_FAIL;
+				} else {
+					if (J9_UNEXPECTED(!VM_ObjectMonitor::recordBytecodeMonitorEnter(_currentThread, (j9object_t)monitorRC, _arg0EA))) {
+						objectMonitorExit(_currentThread, obj);
+						rc = THROW_MONITOR_ALLOC_FAIL;
+						if (immediateAsyncPending()) {
+							rc = GOTO_ASYNC_CHECK;
+						}
+						goto done;
+					}
 					if (immediateAsyncPending()) {
 						rc = GOTO_ASYNC_CHECK;
+						goto done;
 					}
-					goto done;
+					_pc += 1;
+					_sp += 1;
 				}
-				if (immediateAsyncPending()) {
-					rc = GOTO_ASYNC_CHECK;
-					goto done;
-				}
-				_pc += 1;
-				_sp += 1;
 			}
 		}
 done:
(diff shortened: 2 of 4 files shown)


MarkQingGuo:
@tajila please review.

test/functional/Valhalla/src/org/openj9/test/lworld/ValueTypeGenerator.java
```diff
@@ -174,7 +179,79 @@ private static void testWithFieldOnNonExistentClass(ClassWriter cw, String class
 		mv.visitMaxs(1, 2);
 		mv.visitEnd();
 	}
-	
+
+	 /* 
+	  * This function should only be called in the
+	  * TestMonitorEnterOnValueType test
+	  */
+	private static void testMonitorEnterOnValueType(ClassWriter cw, String className, String[] fields) {
```
tajila: you dont need the value and ref variants since they are both the same

runtime/vm/BytecodeInterpreter.hpp
```diff
@@ -7590,8 +7611,29 @@ done:;
 		VM_BytecodeAction rc = EXECUTE_BYTECODE;
 		j9object_t obj = *(j9object_t*)_sp;
 		_sp += 1;
+
+#if defined(J9VM_OPT_VALHALLA_VALUE_TYPES)
+		J9Class * objClass = NULL;
+
+		if (NULL == obj) {
```
tajila: This is the same with or without valuetypes so it can be taken out of the ifdef

runtime/vm/BytecodeInterpreter.hpp
```diff
@@ -7601,6 +7643,7 @@ done:;
 				_pc += 1;
 			}
 		}
+done:
```
tajila: and you wont need this

runtime/vm/BytecodeInterpreter.hpp
tajila: same for monitor enter

MarkQingGuo:
Please don't review this. I'm testing the correctness of these code.

MarkQingGuo:
I think we might need these goto statements to increase readability of the code, otherwise, the code will becomes really messy. Please tell me if there is a better way.

runtime/vm/BytecodeInterpreter.hpp
```diff
@@ -7640,17 +7656,33 @@ done:;
...
+		if (J9_IS_J9CLASS_VALUETYPE(objClass)) {
+			J9UTF8 *badClassName = J9ROMCLASS_CLASSNAME(objClass->romClass);
+			buildInternalNativeStackFrame(REGISTER_ARGS);
+			updateVMStruct(REGISTER_ARGS);
+			prepareForExceptionThrow(_currentThread);
+			setCurrentExceptionNLSWithArgs(_currentThread, J9NLS_VM_ERROR_BYTECODE_OBJECTREF_CANNOT_BE_VALUE_TYPE, J9VMCONSTANTPOOL_JAVALANGILLEGALMONITORSTATEEXCEPTION, J9UTF8_LENGTH(badClassName), J9UTF8_DATA(badClassName));
+			VMStructHasBeenUpdated(REGISTER_ARGS);
+			rc = GOTO_THROW_CURRENT_EXCEPTION;
+			goto done;
+		}
```
tajila: I think if you make this `} else ` you dont need the goto

MarkQingGuo:
@tajila Please review. That's what I'm thinking with minimum changes.
goto statement might be necessary here, otherwise we will use a monitor on a value type.

runtime/vm/BytecodeInterpreter.hpp
tajila:
```
if (NULL = obj) {
   ...
} else {
ifdef VT
    if (obj.isValue) {
        ...
    } else 
endif VT
   {
        ... // rest of code
   }
return ...

Comment (tajila) [reply runtime/vm/BytecodeInterpreter.hpp]: just modified it, I think its fine to fetch the objClass twice since its an exception path. Will need a comment to indicate why that is happening

File: runtime/vm/BytecodeInterpreter.hpp [hunk 5]
```diff
@@ -7605,28 +7605,42 @@ done:;
...
-			 * release VM access, so the immediate async and failed enter cases are
-			 * mutually exclusive.
-			 */
-			if (0 == monitorRC) {
-				rc = THROW_MONITOR_ALLOC_FAIL;
-			} else {
-				if (J9_UNEXPECTED(!VM_ObjectMonitor::recordBytecodeMonitorEnter(_currentThread, (j9object_t)monitorRC, _arg0EA))) {
-					objectMonitorExit(_currentThread, obj);
+#if defined(J9VM_OPT_VALHALLA_VALUE_TYPES)
+			J9Class * objClass = J9OBJECT_CLAZZ(_currentThread, obj);
```
tajila: same thing here, but you can keep the goto, sine its already being used

runtime/vm/BytecodeInterpreter.hpp
MarkQingGuo: I have modified it. emm, I have never seen a bracket used like this before : ) .

tajila: not quite, the `{` after the first else should be removed.

And youll have to do `if (J9_IS_J9CLASS_VALUETYPE(J9OBJECT_CLAZZ(_currentThread, obj))`

tajila: The purpose is to ensure that non-VT remains unchanged

MarkQingGuo: please review. Just understand the purpose of such modification, hope this is what you want.

runtime/vm/BytecodeInterpreter.hpp
```diff
@@ -7642,7 +7655,20 @@ done:;
 		_sp += 1;
 		if (NULL == obj) {
 			rc = THROW_NPE;
-		} else {
+		} else
```
tajila: should be 
```
} else
ifdef 
if (obj.isValue) {
   ...
}
endif
{
    ... // rest of code
}

Comment (MarkQingGuo) [reply runtime/vm/BytecodeInterpreter.hpp]: please review. Tabbing is fixed.

**Comment by tajila** [comment]
@MarkQingGuo please resolve all the comments

File: test/functional/Valhalla/src/org/openj9/test/lworld/ValueTypeTests.java [hunk 7]
```diff
@@ -397,7 +397,116 @@ static public void testWithFieldOnNonExistentClass() throws Throwable {
...
+	static public void testMonitorExitOnValueType() throws Throwable {
+		int x = 1;
+		int y = 1;
+		Object valueType = makePoint2D.invoke(x, y);
+
+		String fields[] = {"longField:J"};
+		Class<?> testMonitorExitOnValueType = ValueTypeGenerator.generateRefClass("TestMonitorExitOnValueType", fields);
+		MethodHandle monitorExitOnValueType = lookup.findStatic(testMonitorExitOnValueType, "testMonitorExitOnObject", MethodType.methodType(void.class, Object.class));
+		try {
+			monitorExitOnValueType.invoke(valueType);
```
tajila: Were you able to confirm that the NLS message is printed properly with the className etc?

MarkQingGuo: yes, I checked it before. The NLS message is printed.

runtime/nls/j9vm/j9vm.nls
```diff
@@ -1871,3 +1871,12 @@ J9NLS_VM_ERROR_BYTECODE_OBJECTREF_MUST_BE_VALUE_TYPE.explanation=The class has s
 J9NLS_VM_ERROR_BYTECODE_OBJECTREF_MUST_BE_VALUE_TYPE.system_action=The JVM will throw a IncompatibleClassChangeError.
 J9NLS_VM_ERROR_BYTECODE_OBJECTREF_MUST_BE_VALUE_TYPE.user_response=Contact the provider of the classfile for a corrected version.
 # END NON-TRANSLATABLE
+
+J9NLS_VM_ERROR_BYTECODE_OBJECTREF_CANNOT_BE_VALUE_TYPE=bad object type %2$.*1$s: object that is synchronized is a value type
+# START NON-TRANSLATABLE
+J9NLS_VM_ERROR_BYTECODE_OBJECTREF_CANNOT_BE_VALUE_TYPE.sample_input_1=3
+J9NLS_VM_ERROR_BYTECODE_OBJECTREF_CANNOT_BE_VALUE_TYPE.sample_input_2=Foo
+J9NLS_VM_ERROR_BYTECODE_OBJECTREF_CANNOT_BE_VALUE_TYPE.explanation=The class has specified the bytecode monitorenter and monitorexit operation that cannot be performed on a value type but the object on stack is a value type.
```
tajila: The class has specified the monitorenter and monitorexit bytecode operations which

tajila:
minor comments, the rest LGTM

tajila:
@gacholio please review

MarkQingGuo:
@tajila I have changed it. feels bad for my grammar : (

runtime/vm/BytecodeInterpreter.hpp
```diff
@@ -7642,7 +7655,20 @@ done:;
...
+#if defined(J9VM_OPT_VALHALLA_VALUE_TYPES)
+		if (J9_IS_J9CLASS_VALUETYPE(J9OBJECT_CLAZZ(_currentThread, obj))) {
+			J9Class * objClass = J9OBJECT_CLAZZ(_currentThread, obj);
```
gacholio: double fetch of the class

runtime/nls/j9vm/j9vm.nls
```diff
@@ -1871,3 +1871,12 @@ J9NLS_VM_ERROR_BYTECODE_OBJECTREF_MUST_BE_VALUE_TYPE.explanation=The class has s
 J9NLS_VM_ERROR_BYTECODE_OBJECTREF_MUST_BE_VALUE_TYPE.system_action=The JVM will throw a IncompatibleClassChangeError.
 J9NLS_VM_ERROR_BYTECODE_OBJECTREF_MUST_BE_VALUE_TYPE.user_response=Contact the provider of the classfile for a corrected version.
 # END NON-TRANSLATABLE
+
+J9NLS_VM_ERROR_BYTECODE_OBJECTREF_CANNOT_BE_VALUE_TYPE=bad object type %2$.*1$s: object that is synchronized is a value type
+# START NON-TRANSLATABLE
+J9NLS_VM_ERROR_BYTECODE_OBJECTREF_CANNOT_BE_VALUE_TYPE.sample_input_1=3
+J9NLS_VM_ERROR_BYTECODE_OBJECTREF_CANNOT_BE_VALUE_TYPE.sample_input_2=Foo
+J9NLS_VM_ERROR_BYTECODE_OBJECTREF_CANNOT_BE_VALUE_TYPE.explanation=The class has specified the bytecode monitorenter and monitorexit operation which cannot be performed on a value type but the object on stack is a value type.
```
gacholio: enter or exit, rather than and

runtime/vm/BytecodeInterpreter.hpp
MarkQingGuo: If we don't double fetch the class, we might have to use a goto statement here. Is it fine?

runtime/nls/j9vm/j9vm.nls
MarkQingGuo: I will correct this.

runtime/vm/BytecodeInterpreter.hpp
tajila: I think GAC just means that the class should be cached instead of fetched twice. 

I proposed this approach because it would allow us to keep the non-VT code unchanged. Also the double fetch is on the exception path, so even if the compile doesn't cache it for us, it is a slow path regardless.

That being said we could re-write as:

```
if (NULL == obj) {
   ...
} else {
ifdef VT
   Class clazz = //fetch and save class
   if (class.isVT) {
      ...
   } else
endif VT
  {
      //rest of code
  }
}
```

MarkQingGuo:
I have changed this.
@gacholio please review.

runtime/vm/BytecodeInterpreter.hpp
```diff
@@ -7605,28 +7605,42 @@ done:;
...
+			if (J9_IS_J9CLASS_VALUETYPE(objClass)) {
+				J9UTF8 *badClassName = J9ROMCLASS_CLASSNAME(objClass->romClass);
+				buildInternalNativeStackFrame(REGISTER_ARGS);
+				updateVMStruct(REGISTER_ARGS);
+				prepareForExceptionThrow(_currentThread);
+				setCurrentExceptionNLSWithArgs(_currentThread, J9NLS_VM_ERROR_BYTECODE_OBJECTREF_CANNOT_BE_VALUE_TYPE, J9VMCONSTANTPOOL_JAVALANGILLEGALMONITORSTATEEXCEPTION, J9UTF8_LENGTH(badClassName), J9UTF8_DATA(badClassName));
+				VMStructHasBeenUpdated(REGISTER_ARGS);
+				rc = GOTO_THROW_CURRENT_EXCEPTION;
+			}
```
tajila: needs to be an else here, (ie. `} else`). 

Same below

MarkQingGuo: I have correct this.

MarkQingGuo:
@gacholio please review.

gacholio:

PR #4968, opened 2019-03-04
WIP: Marshal relo data

dsouzai:
The data that is stored in the `_targetAddress`/`_targetAddress2` may be
different on different platforms. However, relocation header data that
is stored to to the SCC is common across platforms. In order to fully
consolidate the code that writes the header information, this PR
introduces a method that has to be implemented on each platform that is
used to demarshal the platform specific data from the
`TR::IteratedExternalRelocation` object and marshal it into the platform
agnostic form that is used to write out the header information.

### DIFF
--- a/runtime/compiler/arm/codegen/J9AheadOfTimeCompile.cpp
+++ b/runtime/compiler/arm/codegen/J9AheadOfTimeCompile.cpp
@@ -101,6 +101,21 @@ void J9::ARM::AheadOfTimeCompile::processRelocations()
       }
    }
 
+void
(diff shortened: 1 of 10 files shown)


dsouzai:
@jdmpapin @mstoodle This is what I was thinking of doing to deal with having a common point to write out the relocation header data due the fact that some relocations such as `TR_DataAddress` have different data in the `TR::IteratedExternalRelocation` on different platforms.

What do you guys think?

dsouzai:
I had a discussion about this with @jdmpapin offline. It became clear to me that there are different ways to approach this problem. However, since there are only a few relocation types that differ across platforms, and the majority that don't have yet to be consolidated, I'm going to close this PR in favour of re-evaluating a solution when the consolidation work for everything that can be consolidated is done.

PR #4975, opened 2019-03-04
Fix NPE for extra_getsource_options

vsebe:
Add check for options per spec.

[ci skip]


vsebe:
@AdamBrousseau please review

PR #4978, opened 2019-03-05
AArch64: Add #if defined(TR_HOST_ARM64)

knn-k:
This commit adds `defined(TR_HOST_ARM64)` to `#if` lines so that
the JIT files can be compiled for AArch64.


knn-k:
There are more places to add `defined(TR_HOST_AARCH64)`, but this PR gives the minimum set of changes that are needed to compile platform-independent JIT files.

0xdaryl:
Jenkins test sanity all jdk11

knn-k:
Corrected the title and the comment of the commit.

PR #4982, opened 2019-03-05
Inline TypeCoercion calls in RecognizedCallTransformer instead

dchopra001:
Java's floatToIntBits, floatToRawIntBits, doubleToLongBits, and doubleToRawLongBits are currently inlined inside TR_J9VM::inlineNativeCall. These should be inlined inside RecognizedCallTransformer instead. This will also improve performance because the Walker currently does not inline floatToIntBits/doubleToLongBits calls because they are not natives.

### DIFF
--- a/runtime/compiler/env/VMJ9.cpp
+++ b/runtime/compiler/env/VMJ9.cpp
@@ -7685,38 +7685,10 @@ TR_J9VM::inlineNativeCall(TR::Compilation * comp, TR::TreeTop * callNodeTreeTop,
          if (comp->cg()->getSupportsInliningOfTypeCoersionMethods())
             TR::Node::recreate(callNode, TR::ibits2f);
          return callNode;
-      case TR::java_lang_Float_floatToIntBits:
-         if (comp->cg()->getSupportsInliningOfTypeCoersionMethods())
-            {
-            TR::Node::recreate(callNode, TR::fbits2i);
-            callNode->setNormalizeNanValues(true);
-            }
-         return callNode;
-      case TR::java_lang_Float_floatToRawIntBits:
-         if (comp->cg()->getSupportsInliningOfTypeCoersionMethods())
-            {
-            TR::Node::recreate(callNode, TR::fbits2i);
-            callNode->setNormalizeNanValues(false);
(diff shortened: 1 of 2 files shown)


andrewcraik:
There is a danger in moving the non-natives to the recognized call transformer - we may need HCR guards for them which I believe would be omitted so you likely need to do something to handle that.

dchopra001:
@andrewcraik 

I totally forgot about this! The 4 calls we are inlining here are hotspot intrinsic candidates. So I don't think we will need HCR guards for them. 

The annotation itself was introduced in JDK9, so I'm providing links to each one here as an example as well as for documentation purposes:

https://github.com/ibmruntimes/openj9-openjdk-jdk9/blob/df4c51090b2352a212ea93de97257c3361fbcd95/jdk/src/java.base/share/classes/java/lang/Float.java#L766-L767

https://github.com/ibmruntimes/openj9-openjdk-jdk9/blob/df4c51090b2352a212ea93de97257c3361fbcd95/jdk/src/java.base/share/classes/java/lang/Float.java#L809-L810

https://github.com/ibmruntimes/openj9-openjdk-jdk9/blob/df4c51090b2352a212ea93de97257c3361fbcd95/jdk/src/java.base/share/classes/java/lang/Double.java#L853-L854

https://github.com/ibmruntimes/openj9-openjdk-jdk9/blob/df4c51090b2352a212ea93de97257c3361fbcd95/jdk/src/java.base/share/classes/java/lang/Double.java#L897-L898

andrewcraik:
@dchopra001 ok if they have the annotation I think that is fine. Ideally, there would be a check in the code to make sure the function does have the annotation, but we could get away without it for now if we need to get this in for performance reasons. If the functions exist in Java 8 do they need HCR guards or were they introduced in Java 9?

dchopra001:
> If the functions exist in Java 8 do they need HCR guards or were they introduced in Java 9?

The functions existed in Java 8 as well, and I believe they were intrinsics then as well (But this annotation itself was only introduced in Java 9). I can do a check to verify if you prefer.

genie-openj9:

PR #4989, opened 2019-03-05
Make String.indexOf(String) handle compact strings correctly

NigelYiboYu:
This commit fixes a bug in String.indexOf(String) API where certain optimal
paths (intrinsics that can be hardware accelerated) can never be reached. 
As a result of this, the String.indexOf(string) API always defaults to the sub-optimal
string search logic.
This applies to both compact and non-compact string modes.

### DIFF
--- a/jcl/src/java.base/share/classes/java/lang/String.java
+++ b/jcl/src/java.base/share/classes/java/lang/String.java
@@ -6066,36 +6066,41 @@ public final class String implements Serializable, Comparable<String>, CharSeque
 			char[] s1Value = s1.value;
 			char[] s2Value = s2.value;
 
-			if (enableCompression && (null == compressionFlag || (s1.count | s2.count) >= 0)) {
-				// both s1 and s2 are compressed
-				return helpers.intrinsicIndexOfStringLatin1(s1Value, s1len, s2Value, s2len, start);
-			} else if (s1.count < 0 && s2.count < 0) {
-				// both s1 and s2 are decompressed
-				return helpers.intrinsicIndexOfStringUTF16(s1Value, s1len, s2Value, s2len, start);
-			} else {
-				// s1 decompressed and s2 compressed
-				char firstChar = s2.charAtInternal(0, s2Value);
+			if (enableCompression) {
+				if (null == compressionFlag || (s1.count | s2.count) >= 0) {
+					// Both s1 and s2 are compressed.
+					return helpers.intrinsicIndexOfStringLatin1(s1Value, s1len, s2Value, s2len, start);
+				} else if ((s1.count & s2.count) < 0) {
+					// Both s1 and s2 are decompressed.
+					return helpers.intrinsicIndexOfStringUTF16(s1Value, s1len, s2Value, s2len, start);
+				} else {
+					// Mixed case.
+					char firstChar = s2.charAtInternal(0, s2Value);
 
-				while (true) {
-					int i = indexOf(firstChar, start);
+					while (true) {
+						int i = indexOf(firstChar, start);
 
-					// Handles subCount > count || start >= count
-					if (i == -1 || s2len + i > s1len) {
-						return -1;
+						// Handles subCount > count || start >= count.
+						if (i == -1 || s2len + i > s1len) {
+							return -1;
+						}
+
+						int o1 = i;
+						int o2 = 0;
+
+						while (++o2 < s2len && s1.charAtInternal(++o1, s1Value) == s2.charAtInternal(o2, s2Value))
+							;
+
+						if (o2 == s2len) {
+							return i;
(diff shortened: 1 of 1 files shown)


NigelYiboYu:
FYI @fjeremic Tested OK. Ready to review.

fjeremic:
Let's also get a second pair of eyes on this so I'll request a review from one of @keithc-ca or @DanHeidinga.

Jenkins test sanity xlinux,zlinux JDK8,JDK11

jcl/src/java.base/share/classes/java/lang/String.java
```diff
@@ -6066,36 +6066,41 @@ public int indexOf(String subString, int start) {
...
-				// both s1 and s2 are decompressed
-				return helpers.intrinsicIndexOfStringUTF16(s1Value, s1len, s2Value, s2len, start);
-			} else {
-				// s1 decompressed and s2 compressed
-				char firstChar = s2.charAtInternal(0, s2Value);
+			if (enableCompression) {
+				if (null == compressionFlag || (s1.count | s2.count) >= 0) {
+					// Both s1 and s2 are compressed
+					return helpers.intrinsicIndexOfStringLatin1(s1Value, s1len, s2Value, s2len, start);
+				} else if (s1.count < 0 && s2.count < 0) {
```
keithc-ca: Following the example above, this could be written as `(s1.count & s2.count) < 0`.

jcl/src/java.base/share/classes/java/lang/String.java
```diff
@@ -6066,36 +6066,41 @@ public int indexOf(String subString, int start) {
...
-				return helpers.intrinsicIndexOfStringLatin1(s1Value, s1len, s2Value, s2len, start);
-			} else if (s1.count < 0 && s2.count < 0) {
-				// both s1 and s2 are decompressed
-				return helpers.intrinsicIndexOfStringUTF16(s1Value, s1len, s2Value, s2len, start);
-			} else {
-				// s1 decompressed and s2 compressed
-				char firstChar = s2.charAtInternal(0, s2Value);
+			if (enableCompression) {
+				if (null == compressionFlag || (s1.count | s2.count) >= 0) {
+					// Both s1 and s2 are compressed
```
keithc-ca: nit: Either don't change the comment, or, if you want to make it a sentence (starting with a capital), then end it with a period.

jcl/src/java.base/share/classes/java/lang/String.java
NigelYiboYu: Fixed.

jcl/src/java.base/share/classes/java/lang/String.java
NigelYiboYu: Fixed.

jcl/src/java.base/share/classes/java/lang/String.java
```diff
@@ -6066,36 +6066,41 @@ public int indexOf(String subString, int start) {
...
-				// both s1 and s2 are decompressed
-				return helpers.intrinsicIndexOfStringUTF16(s1Value, s1len, s2Value, s2len, start);
-			} else {
-				// s1 decompressed and s2 compressed
-				char firstChar = s2.charAtInternal(0, s2Value);
+			if (enableCompression) {
+				if (null == compressionFlag || (s1.count | s2.count) >= 0) {
+					// Both s1 and s2 are compressed.
+					return helpers.intrinsicIndexOfStringLatin1(s1Value, s1len, s2Value, s2len, start);
+				} else if (s1.count & s2.count < 0) {
```
DanHeidinga: This needs brackets to clarify the order of operations - don't make the reader try to remember what it is.

NigelYiboYu: Brackets added.

DanHeidinga:
Jenkins test sanity xlinux,zlinux JDK8,JDK11

DanHeidinga:
Jenkins test sanity xlinux,zlinux jdk8

DanHeidinga:

keithc-ca:
Jenkins test sanity xlinux,zlinux jdk8,jdk11

fjeremic:
@keithc-ca we have a dependant change in flight. See committers slack channel. We'll have to wait until OMR-Acceptnce propagates. Should be less than 30 min now. I'll relaunch the build as soon as it completes. Sorry for the churn.

fjeremic:
Jenkins test sanity xlinux,zlinux jdk8,jdk11

keithc-ca:
> @keithc-ca we have a dependant change in flight.

@fjeremic No worries. I was just helping to move this along. Given this is assigned to you, I was planning to let you merge when you're ready.

PR #4992, opened 2019-03-05
Check the stack size of a stackmap frame

ChengJin01:
The change is to check the stack size of a stackmap frame
against the maximal stack size specified in the class file
so as to capture invalid stack size during static verification.

### DIFF
--- a/runtime/bcverify/staticverify.c
+++ b/runtime/bcverify/staticverify.c
@@ -1313,6 +1313,12 @@ checkStackMap (J9CfrClassFile* classfile, J9CfrMethod * method, J9CfrAttributeCo
 					slotCount = 0;
 					if ((frameType >= CFR_STACKMAP_SAME_LOCALS_1_STACK) && (frameType <= CFR_STACKMAP_SAME_LOCALS_1_STACK_EXTENDED)) {
 						slotCount = 1;
+
+						/* The stackmap entry is invalid if the size of stack (1 slot) exceeds the size of the max stack.*/
+						if (code->maxStack < slotCount) {
+							errorCode = FATAL_CLASS_FORMAT_ERROR;
+							goto _failedCheck;
+						}
 					}
 					if (frameType >= CFR_STACKMAP_CHOP_3) {
 						slotCount = (IDATA) frameType - CFR_STACKMAP_APPEND_BASE;


ChengJin01:
Reviewer: @DanHeidinga

runtime/bcverify/staticverify.c
```diff
@@ -1313,6 +1313,12 @@ checkStackMap (J9CfrClassFile* classfile, J9CfrMethod * method, J9CfrAttributeCo
 					slotCount = 0;
 					if ((frameType >= CFR_STACKMAP_SAME_LOCALS_1_STACK) && (frameType <= CFR_STACKMAP_SAME_LOCALS_1_STACK_EXTENDED)) {
 						slotCount = 1;
+
+						/* The stackmap entry is invalid if the size of stack (1 slot) exceeds the size of the max stack.*/
+						if (code->maxStack < slotCount) {
```
DanHeidinga: Is this the right place for this code?  Does the `CFR_STACKMAP_CHOP_3` stackmap type (below) need this check as well

ChengJin01: The code was added against the VM Spec as follows:
```
4.7.4 The StackMapTable Attribute
union stack_map_frame {
same_frame;  <-- the operand stack is empty
same_locals_1_stack_item_frame;  <--- the operand stack has one entry
same_locals_1_stack_item_frame_extended; <--- the operand stack has one entry
chop_frame;  <-- the operand stack is empty
same_frame_extended; <-- the operand stack is empty
append_frame;  <-- the operand stack is empty
full_frame;  <-- the operand stack is number_of_stack_items defined in full_frame
}
```
It means the stack size of all types of frames (including chop_frame) is 0 except `same_locals_1_stack_item_frame` (1 stack slot), `same_locals_1_stack_item_frame_extended ` (1 stack slot) and `full_frame ` which holds stack slots which can be checked against the max stack size.

In addition, if frameType >= CFR_STACKMAP_CHOP_3, the slots (if > 0) belong to local variables rather than the operation stack. So there is no need to check there.

I didn't add the stack size check for the full_frame (as follows if we need) as there is no test case for this for now.
```
/* full frame */
/* Executed with StackMap or StackMapTable */
NEXT_U16(slotCount, entries);
localSlots = slotCount;
errorCode = checkStackMapEntries (classfile, code, map, &entries, slotCount, end);
if (0 != errorCode) {
		goto _failedCheck;
}
NEXT_U16(slotCount, entries);  <------------------ this is the stack size in the full_frame
	if (code->maxStack < slotCount) { 
		errorCode = FATAL_CLASS_FORMAT_ERROR;
		goto _failedCheck;
	}
```

DanHeidinga: The full frame checks are handled by the checkStackMapEntries calls.

DanHeidinga:
Jenkins test sanity xlinux jdk8,jdk12

ChengJin01:
@DanHeidinga , the PR against 0.13 has been created at https://github.com/eclipse/openj9/pull/5079

PR #4993, opened 2019-03-05
Use client's ROM class pointer in a front-end query

dmitry-ten:
[skip ci]
`TR_J9VM::getPersistentClassPointerFromClassPointer` returns a pointer
to a ROM class. However, on the server, we store a copy of client's
ROM class, and calling this method would return a pointer to that copy.
But this method is always used to measure the offset of a ROM class
from the beginning of shared cache.
Thus, we need to get a pointer to the ROM class on the client side.
Do it by adding a pointer to ClassInfo struct, so that for all cached
ROM classes we know its pointer on the client.

### DIFF
--- a/runtime/compiler/control/JITaaSCompilationThread.cpp
+++ b/runtime/compiler/control/JITaaSCompilationThread.cpp
@@ -3204,6 +3204,7 @@ JITaaSHelpers::cacheRemoteROMClass(ClientSessionData *clientSessionData, J9Class
    classInfoStruct.totalInstanceSize = std::get<16>(classInfo);
    classInfoStruct._classOfStaticCache = nullptr;
    classInfoStruct._constantClassPoolCache = nullptr;
+   classInfoStruct.remoteRomClass = std::get<17>(classInfo);
 
    clientSessionData->getROMClassMap().insert({ clazz, classInfoStruct});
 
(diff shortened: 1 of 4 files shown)


dmitry-ten:
@mpirvu this is ready for review.

dmitry-ten:
Yeah, we don't really have any guarantees that we won't try to dereference the remote pointer on the server.
In the example you provided though, shouldn't it be fine though, because it's during codegen?

dmitry-ten:
We could create a new front-end query, that 
does
```
void *romClass = (void *) fej9->getPersistentClassPointerFromClassPointer(clazz)l
return sharedCache->offsetInSharedCacheFromPointer(romClass)
```
and call it something like `getOffsetInSharedCacheFromROMClassPointer`

mpirvu:
> In the example you provided though, shouldn't it be fine though, because it's during codegen?

Codegen is executed on the server. If `romClassPtr` becomes a client pointer, then we may read bad memory

I don't understand what problem the new fe query will solve.

dmitry-ten:
I assumed that if we write it to `cursor`, then it won't be dereferenced on the server. I don't know if it's true though.

The new query would abstract away getting offset of ROM class from shared cache.
In the base VM it could execute the code quoted above.
In the server VM it would retrieve the ROM class pointer from the client.
In this case we wouldn't have to override `getPersistentClassPointerFromClassPointer` and assume that its return value is never going to be dereferenced.

This wouldn't help with the case you mentioned in s390 codegen. For that, I guess, we just need to figure out what it's going to do with the pointer.

PR #4995, opened 2019-03-05
Add tracepoints when setting then file soft limit

pdbain-ibm:

### DIFF
--- a/runtime/vm/j9vm.tdf
+++ b/runtime/vm/j9vm.tdf
@@ -779,3 +779,7 @@ TraceExit=Trc_VM_sendResolveConstantDynamic_Exit Overhead=1 Level=2 Template="se
 TraceException=Trc_VM_CreateRAMClassFromROMClass_nestedValueClassNotVisible Overhead=1 Level=1 Template="Nested field (RAM class=%p, classloader=%p, this classloader=%p) is not visible. Throw IllegalAccessError"
 
 TraceEvent=Trc_VM_CreateRAMClassFromROMClass_valueTypeIsFlattened Overhead=1 Level=7 Template="ValueType is eligible for flattening name=%.*s j9class=%p"
+
+TraceEvent=Trc_VM_SetSoftFileLimit Overhead=1 Level=3 Template="Set max number of files to %llu"
+TraceEvent=Trc_VM_SetSoftFileLimitResult Overhead=1 Level=3 Template="Set max number of files status=%d"
+TraceException=Trc_VM_SetSoftFileLimitUnknown Overhead=1 Level=1 Template="Cannot determine file descriptor maximum: status=%d"
--- a/runtime/vm/jvminit.c
+++ b/runtime/vm/jvminit.c
@@ -1824,7 +1824,11 @@ IDATA VMInitStages(J9JavaVM *vm, IDATA stage, void* reserved) {
 				uint64_t limit = 0;
 				uint32_t rc = omrsysinfo_get_limit(OMRPORT_RESOURCE_FILE_DESCRIPTORS | J9PORT_LIMIT_HARD, &limit);
 				if (OMRPORT_LIMIT_UNKNOWN != rc ) {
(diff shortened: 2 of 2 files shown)


runtime/vm/j9vm.tdf
```diff
@@ -779,3 +779,6 @@ TraceExit=Trc_VM_sendResolveConstantDynamic_Exit Overhead=1 Level=2 Template="se
 TraceException=Trc_VM_CreateRAMClassFromROMClass_nestedValueClassNotVisible Overhead=1 Level=1 Template="Nested field (RAM class=%p, classloader=%p, this classloader=%p) is not visible. Throw IllegalAccessError"
 
 TraceEvent=Trc_VM_CreateRAMClassFromROMClass_valueTypeIsFlattened Overhead=1 Level=7 Template="ValueType is eligible for flattening name=%.*s j9class=%p"
+
+TraceEvent=Trc_VM_SetSoftFileLimit Overhead=1 Level=3 Template="Set max number of files to %zu"
```
pshipton: `%zu` is not correct for uint64_t as this expect 32-bit for a 32-bit JVM, I think it's `%llu`

runtime/vm/jvminit.c
```diff
@@ -1824,7 +1824,9 @@ IDATA VMInitStages(J9JavaVM *vm, IDATA stage, void* reserved) {
 				uint64_t limit = 0;
 				uint32_t rc = omrsysinfo_get_limit(OMRPORT_RESOURCE_FILE_DESCRIPTORS | J9PORT_LIMIT_HARD, &limit);
 				if (OMRPORT_LIMIT_UNKNOWN != rc ) {
-					omrsysinfo_set_limit(OMRPORT_RESOURCE_FILE_DESCRIPTORS | OMRPORT_LIMIT_SOFT, limit);
+					Trc_VM_SetSoftFileLimit(vm->mainThread, limit);
+					rc = omrsysinfo_set_limit(OMRPORT_RESOURCE_FILE_DESCRIPTORS | OMRPORT_LIMIT_SOFT, limit);
+					Trc_VM_SetSoftFileLimitResult(vm->mainThread, rc);
 				}
```
pshipton: Should there be another tracepoint in the case where omrsysinfo_get_limit fails (OMRPORT_LIMIT_UNKNOWN )?

runtime/vm/jvminit.c
```diff
@@ -1824,7 +1824,9 @@ IDATA VMInitStages(J9JavaVM *vm, IDATA stage, void* reserved) {
 				uint64_t limit = 0;
 				uint32_t rc = omrsysinfo_get_limit(OMRPORT_RESOURCE_FILE_DESCRIPTORS | J9PORT_LIMIT_HARD, &limit);
 				if (OMRPORT_LIMIT_UNKNOWN != rc ) {
-					omrsysinfo_set_limit(OMRPORT_RESOURCE_FILE_DESCRIPTORS | OMRPORT_LIMIT_SOFT, limit);
+					Trc_VM_SetSoftFileLimit(vm->mainThread, limit);
+					rc = omrsysinfo_set_limit(OMRPORT_RESOURCE_FILE_DESCRIPTORS | OMRPORT_LIMIT_SOFT, limit);
+					Trc_VM_SetSoftFileLimitResult(vm->mainThread, rc);
```
pshipton: omrsysinfo_set_limit() contains a tracepoint, do we need another one here?

pdbain-ibm:
> `%zu` is not correct for uint64_t as this expect 32-bit for a 32-bit JVM, I think it's `%llu`

Fixed

> omrsysinfo_set_limit() contains a tracepoint, do we need another one here?

Not necessarily.  I added it for convenience but can take it out

> Should there be another tracepoint in the case where omrsysinfo_get_limit fails (OMRPORT_LIMIT_UNKNOWN )?

Good idea.

pshipton:
I am reluctant to accept this change because omrsysinfo_get_limit() and omrsysinfo_set_limit() contain basically the same tracepoints.

PR #5000, opened 2019-03-05
Array class can get methods

JasonFengJ9:
`Array` class can get methods

`java.lang.Class.getMethodImpl()` is modified to allow `array` class to find any public method inherited by the array type from `Object` except method `clone()`.

**Java doc** [java.base/java/lang/Class.html#getMethod(java.lang.String,java.lang.Class...)](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/lang/Class.html#getMethod(java.lang.String,java.lang.Class...))
```
If this Class object represents an array type, then this method finds any public method inherited by the array type from Object except method clone().
```

Reviewer: @pshipton 
FYI: @DanHeidinga 

### DIFF
--- a/runtime/jcl/common/java_lang_Class.cpp
+++ b/runtime/jcl/common/java_lang_Class.cpp
@@ -894,15 +894,16 @@ Java_java_lang_Class_getMethodImpl(JNIEnv *env, jobject recv, jobject name, jobj
 	J9InternalVMFunctions *vmFuncs = vm->internalVMFunctions;
 	j9object_t resultObject = NULL;
 	vmFuncs->internalEnterVMFromJNI(currentThread);
-	J9Class *clazz = J9VM_J9CLASS_FROM_HEAPCLASS(currentThread, J9_JNI_UNWRAP_REFERENCE(recv));
 
 	PORT_ACCESS_FROM_VMC(currentThread);
 	if ((NULL == name) || (NULL == partialSignature)) {
 		vmFuncs->setCurrentExceptionUTF(currentThread, J9VMCONSTANTPOOL_JAVALANGNULLPOINTEREXCEPTION, NULL);
 	} else {
+		J9Class *clazz = J9VM_J9CLASS_FROM_HEAPCLASS(currentThread, J9_JNI_UNWRAP_REFERENCE(recv));
 		J9ROMClass *romClass = clazz->romClass;
-		/* primitives/arrays don't have local methods */
-		if (!J9ROMCLASS_IS_PRIMITIVE_OR_ARRAY(romClass)) {
+
+		/* primitives doesn't have local methods */
+		if (!J9ROMCLASS_IS_PRIMITIVE_TYPE(romClass)) {
 			J9Method *currentMethod = NULL;
 			j9object_t nameObject = J9_JNI_UNWRAP_REFERENCE(name);
 			j9object_t signatureObject = J9_JNI_UNWRAP_REFERENCE(partialSignature);


runtime/jcl/common/java_lang_Class.cpp
```diff
@@ -894,15 +894,22 @@ Java_java_lang_Class_getMethodImpl(JNIEnv *env, jobject recv, jobject name, jobj
...
 	} else {
-		J9ROMClass *romClass = clazz->romClass;
-		/* primitives/arrays don't have local methods */
-		if (!J9ROMCLASS_IS_PRIMITIVE_OR_ARRAY(romClass)) {
+		J9Class *clazz = J9VM_J9CLASS_FROM_HEAPCLASS(currentThread, J9_JNI_UNWRAP_REFERENCE(recv));
+		J9ROMClass *romClass = NULL;
+		
+		if (J9CLASS_IS_ARRAY(clazz)) {
+			/* Use underlying type for arrays */
+			romClass = ((J9ArrayClass*)clazz)->leafComponentType->romClass;
```
pshipton: I read the javadoc as allowing array classes to find methods in Object, not methods in the component type. Do you have any test cases?

Also, how does this restrict clone()?

DanHeidinga:
@JasonFengJ9 Is this specific to a particular java release?  Or does it apply to 8+?

JasonFengJ9:
@DanHeidinga though the initial test failure was reported against `JDK12`, `RI JDK8` has same behaviours, so this PR intended to apply to `8+`.

runtime/jcl/common/java_lang_Class.cpp
pshipton: I see, you are modifying the !J9ROMCLASS_IS_PRIMITIVE_TYPE test. This also has side effects because the romClass is used to set the lookupFLags. Just change the !J9ROMCLASS_IS_PRIMITIVE_TYPE test directory rather than modifying the romClass.

pshipton:
Intentionally added to 0.14 milestone and not 0.13. It is a long standing problem, and I'm not aware that anybody has reported a problem because of it.

runtime/jcl/common/java_lang_Class.cpp
JasonFengJ9: > I read the javadoc as allowing array classes to find methods in Object, not methods in the component type. Do you have any test cases?

Manually verified that `RI` has following behaviours:
`System.out.println(new TestClassReflect[2].getClass().getMethod("getClass")`
returns 
`class [Latestse.java_lang_class.TestClassReflect;.getMethod(getClass) = public final native java.lang.Class java.lang.Object.getClass()`,
and 
`System.out.println(new TestClassReflect[2].getClass().getMethod("clone")` 
throws 
`java.lang.NoSuchMethodException: [Latestse.java_lang_class.TestClassReflect;.clone()`
and 
`System.out.println(new TestClassReflect[2].getClass().getMethod("method0")` 
throws 
`java.lang.NoSuchMethodException: [Latestse.java_lang_class.TestClassReflect;.method0()`
in which `method0` is a public method within class `TestClassReflect`.
`OpenJ9` with this PR matches the behaviours above.
Note: these behaviours are observed in `RI JDK8` as well.      

> Also, how does this restrict clone()?

`java.lang.Class.getMethod()` only find `public` method while `clone` is a `protected` method.

JasonFengJ9: > Just change the !J9ROMCLASS_IS_PRIMITIVE_TYPE test directly rather than modifying the romClass.

Agreed, PR has been updated. Also verified that `OpenJ9` with this change matches `RI` behaviours in question. Please have another look.

pshipton:
jenkins test all plinux jdk8,jdk11

pshipton:
jenkins test all plinux jdk11

pshipton:

pshipton:
jenkins test extended plinux jdk11

Issue #4940, opened 2019-03-01
JTReg Test Failure : java/lang/constant/CondyDescTest.java


ben-walsh:
Consistent failures ...

```
[2019-02-27T15:39:53.713Z] STDOUT:
[2019-02-27T15:39:53.713Z] [TestNG] Running:
[2019-02-27T15:39:53.713Z]   java/lang/constant/CondyDescTest.java
[2019-02-27T15:39:53.713Z] 
[2019-02-27T15:39:53.713Z] test CondyDescTest.testDynamicConstant(): success
[2019-02-27T15:39:53.713Z] test CondyDescTest.testEnumDesc(): success
[2019-02-27T15:39:53.713Z] test CondyDescTest.testLifting(): failure
[2019-02-27T15:39:53.713Z] java.lang.AssertionError: expected [DynamicConstantDesc[ConstantBootstraps::staticFieldVarHandle(sf/ClassDesc[CondyDescTest$MyClass],PrimitiveClassDesc[int])VarHandle]] but found [VarHandleDesc[static CondyDescTest$MyClass.sf:int]]
[2019-02-27T15:39:53.713Z] 	at org.testng.Assert.fail(Assert.java:94)
[2019-02-27T15:39:53.713Z] 	at org.testng.Assert.failNotEquals(Assert.java:496)
[2019-02-27T15:39:53.713Z] 	at org.testng.Assert.assertEquals(Assert.java:125)
[2019-02-27T15:39:53.713Z] 	at org.testng.Assert.assertEquals(Assert.java:167)
[2019-02-27T15:39:53.713Z] 	at CondyDescTest.assertLifted(CondyDescTest.java:212)
[2019-02-27T15:39:53.713Z] 	at CondyDescTest.testLifting(CondyDescTest.java:239)
[2019-02-27T15:39:53.713Z] 	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
[2019-02-27T15:39:53.713Z] 	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
[2019-02-27T15:39:53.713Z] 	at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
[2019-02-27T15:39:53.713Z] 	at java.base/java.lang.reflect.Method.invoke(Method.java:567)
[...]

PR #5003, opened 2019-03-06
Fix VarHandleDesc constructor setup

theresa-m:
- fix logic error in VarHandle.VarHandleDesc.Kind.getBootstrap
- fix VarHandleDesc constructor superclass call to DynamicConstantDesc
- remove declaringClassDesc field in VarHandleDesc. This information can be retrieved from the superclass bootstrap arguments
- add private helper methods to retrieve information from superclass bootstrap arguments
- add additional tests for DynamicConstantDesc to VarHandle description <-> resolution

Fixes: #4940

This will also need to be delivered to 0.13 branch, I'll create a separate pr once this has been reviewed

fyi @DanHeidinga

### DIFF
--- a/jcl/src/java.base/share/classes/java/lang/invoke/VarHandle.java
+++ b/jcl/src/java.base/share/classes/java/lang/invoke/VarHandle.java
@@ -351,7 +351,6 @@ public abstract class VarHandle extends VarHandleInternal
 			return false;
 		}
 
-
 		/* argument comparison */
 		VarHandle that = (VarHandle)obj;
 		if (!(this.fieldType.equals(that.fieldType) 
@@ -1037,7 +1036,6 @@ public abstract class VarHandle extends VarHandleInternal
 	/* nominal descriptor of a VarHandle constant */
 	public static final class VarHandleDesc extends DynamicConstantDesc<VarHandle> implements ConstantDesc {
 		private Kind type = null;
-		private ClassDesc declaringClassDesc = null;
 
 		private static enum Kind {
 			ARRAY,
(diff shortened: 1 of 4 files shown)


jcl/src/java.base/share/classes/java/lang/invoke/VarHandle.java
```diff
@@ -1065,14 +1077,13 @@ DirectMethodHandleDesc getBootstrap() {
...
 		 * @param declaringClassDesc ClassDesc describing the declaring class for the field (null for array)
-		 * @param name unqualified String name of the field (null for array)
-		 * @param varHandleType ClassDesc describing the field type
+		 * @param name unqualified String name of the field (blank for array)
+		 * @param varHandleType ClassDesc describing the field type (arrayType for array)
 		 * @throws NullPointerException if there is a null argument
 		 */
 		private VarHandleDesc(Kind type, ClassDesc declaringClassDesc, String name, ClassDesc varHandleDesc) throws NullPointerException {
-			super(type.getBootstrap(), name, varHandleDesc, ConstantDescs.CD_VarHandle);
+			super(type.getBootstrap(), name, ConstantDescs.CD_VarHandle, type.getBootstrapArgs(declaringClassDesc, varHandleDesc));
```
DanHeidinga: `NullPointerException if there is a null argument`  

Is this code missing some null checks / `Objects.requireNonNull()` calls?

jcl/src/java.base/share/classes/java/lang/invoke/VarHandle.java
```diff
@@ -1090,7 +1101,7 @@ public static VarHandleDesc ofArray(ClassDesc arrayClass) throws NullPointerExce
 				/*[MSG "K0625", "{0} is not an array type."]*/
 				throw new IllegalArgumentException(com.ibm.oti.util.Msg.getString("K0625", arrayClass.descriptorString())); //$NON-NLS-1$
 			}
-			return new VarHandleDesc(Kind.ARRAY, arrayClass, "_", arrayClass.componentType()); //$NON-NLS-1$
+			return new VarHandleDesc(Kind.ARRAY, null, "_", arrayClass); //$NON-NLS-1$
```
DanHeidinga: The constructor says it will throw an NPE for any null args.

Also, it says the name should be `(blank for array)`.  Does that mean `"_"` or `""` (empty string)?  Its not clear which is correct to me

jcl/src/java.base/share/classes/java/lang/invoke/VarHandle.java
```diff
@@ -1196,8 +1205,41 @@ public String toString() {
 		 * @return ClassDesc of variable type
 		 */
 		public ClassDesc varType() {
-			return constantType();
+			return (type == Kind.ARRAY) ? getArrayTypeClassDesc().componentType() : getFieldTypeClassDesc();
+		}
+
+		/* helper method. declaring class is always the first bootstrap argument for a field type
```
DanHeidinga: nitpick; Please start the sentence with a capital letter

jcl/src/java.base/share/classes/java/lang/invoke/VarHandle.java
```diff
@@ -1196,8 +1205,41 @@ public String toString() {
...
+		 * @throws InternalError if VarHandleDesc was created improperly
+		 */
+		private ClassDesc getFieldDeclaringClassDesc() {
+			ConstantDesc[] args = bootstrapArgs();
+			if ((args.length < 2) || !(args[0] instanceof ClassDesc)) {
+				throw new InternalError();
```
DanHeidinga: Please add a description to this exception so that it will be easier to understand what went wrong form a stacktrace

jcl/src/java.base/share/classes/java/lang/invoke/VarHandle.java
```diff
@@ -1196,8 +1205,41 @@ public String toString() {
...
+			return (ClassDesc)args[0];
+		}
+
+		/* helper method. fieldType is always the second bootstrap argument for a field type
+		 * @throws InternalError if VarHandleDesc was created improperly
+		 */
+		private ClassDesc getFieldTypeClassDesc() {
+			ConstantDesc[] args = bootstrapArgs();
+			if ((args.length < 2) || !(args[1] instanceof ClassDesc)) {
+				throw new InternalError();
```
DanHeidinga: same here

jcl/src/java.base/share/classes/java/lang/invoke/VarHandle.java
```diff
@@ -1196,8 +1205,41 @@ public String toString() {
...
+		/* helper method. array component type is always the first bootstrap argument for an array type
+		 * @throws InternalError if VarHandleDesc was created improperly
+		 */
+		private ClassDesc getArrayTypeClassDesc() {
+			ConstantDesc[] args = bootstrapArgs();
+			if ((args.length == 0) || !(args[0] instanceof ClassDesc) || (type != Kind.ARRAY)) {
+				throw new InternalError();
+			}
+			return (ClassDesc)args[0];
+		}
```
DanHeidinga: All three of these methods need their javadoc cleaned up and messages added to their exceptions

test/functional/Java12andUp/src/org/openj9/test/java_lang_constant/Test_DynamicConstantDesc.java
```diff
@@ -45,14 +50,77 @@
...
 		logger.debug("testDynamicConstantDescResolveConstantDesc: resolved class is: " + resolvedClass.toString());
 		Assert.assertTrue(resolvedClass.equals(int.class));
+
+		/* test equals */
+		logger.debug("original is: " + desc.toString() + " resolved is: " + resDesc.toString());
+		Assert.assertTrue(desc.equals(resDesc));
+	}
+
+	/* helpers for VarHandle tests */
+	private static ClassDesc helperClass = ClassDesc.of("org.openj9.test.java_lang_constant", "Test_DynamicConstantDesc")
```
DanHeidinga: The formatting looks off here.  Better to avoid heavy levels of indenting by choosing a different point to break the line or just use a long line.

```java
private static ClassDesc helperClass = 
	ClassDesc.of("org.openj9.test.java_lang_constant", "Test_DynamicConstantDesc").nested("NestedHelper")
```

jcl/src/java.base/share/classes/java/lang/invoke/VarHandle.java
theresa-m: yes.. previously it was covered by the superclass. will update

theresa-m:
ready for another review

DanHeidinga:
Jenkins test sanity zlinux jdk12

DanHeidinga:
@theresa-m Can you create the PR against the 0.13.0 branch for this change?

PR #5005, opened 2019-03-06
Fix fields/staticsAreSame for remote AOT

dmitry-ten:
In mainline, we use fe->sameClassLoaders(..) to verify
that classes to which fields/statics compared in
`TR_ResolvedJ9Method::fields/staticsAreSame` belong are loaded by
the same class loader. For AOT that method always returns false.
In JITaaS implementation, we replaced a front-end call with
`getClassLoader() != method2->getClassLoader()`, because server
resolved methods always cache their class loader, so this allows
us to avoid a remote message.
However, this doesn't work correctly in AOT mode, so I added a guard
to always return false if doing an AOT compile.


dmitry-ten:
@mpirvu could you review?

PR #5007, opened 2019-03-06
Create a new front-end query for getting array classes

dmitry-ten:
Create `TR_J9VMBase::getClassFromNewArrayTypeNonNull` query,
which does the same thing as `TR_J9VMBase::getClassFromNewArrayType`,
except the former always returns an array class, while the latter
returns NULL in AOT mode.
This query is needed for cases when we need to get a pointer to boolean
array class during AOT compile.


dmitry-ten:
@mpirvu could you review?
I don't like the name of the new query, any suggestions?
Also, the non-JITaaS part of the code should be merged into mainline, but that would mean we need another rebase, so maybe we can just merge this into jitaas for now.

PR #5010, opened 2019-03-06
Increase _coldUpgradeSampleThreshold to 20

dsouzai:
`_coldUpgradeSampleThreshold` is currently only used in the OpenJ9 project. Because the SVM has resulted in better AOT code quality, under `AGGRESSIVE_AOT`, this PR makes the upgrade sample threshold more conservative by increasing it to 20.

### DIFF
--- a/runtime/compiler/control/J9Options.cpp
+++ b/runtime/compiler/control/J9Options.cpp
@@ -2456,15 +2456,14 @@ bool J9::Options::feLatePostProcess(void * base, TR::OptionSet * optionSet)
       self()->setOption(TR_DisableIntrinsics);
       }
 
-   /* Temporary (until SVM is the default)
-    *
-    * If the SVM is not enabled, use the old _coldUpgradeSampleThreshold
+   /* If the SVM is enabled, use a higher _coldUpgradeSampleThreshold
     * for AGGRESSIVE_AOT
     */
-   if (!self()->getOption(TR_EnableSymbolValidationManager)
-       && TR::Options::_aggressivenessLevel == TR::Options::AGGRESSIVE_AOT)
+   if (TR::Options::_aggressivenessLevel == TR::Options::AGGRESSIVE_AOT &&
+       self()->getOption(TR_EnableSymbolValidationManager) &&
+       TR::Options::sharedClassCache())
       {
-      TR::Options::_coldUpgradeSampleThreshold = 10;
+      TR::Options::_coldUpgradeSampleThreshold = 20;
       }
 
    return true;


dsouzai:
@mpirvu @vijaysun-omr I realized that I could address the concerns brought up in https://github.com/eclipse/omr/pull/3618 entirely in OpenJ9.

runtime/compiler/control/J9Options.cpp
```diff
@@ -2456,15 +2456,14 @@ bool J9::Options::feLatePostProcess(void * base, TR::OptionSet * optionSet)
...
     * for AGGRESSIVE_AOT
     */
-   if (!self()->getOption(TR_EnableSymbolValidationManager)
-       && TR::Options::_aggressivenessLevel == TR::Options::AGGRESSIVE_AOT)
+   if (TR::Options::_aggressivenessLevel == TR::Options::AGGRESSIVE_AOT &&
+       self()->getOption(TR_EnableSymbolValidationManager) &&
+       TR::Options::sharedClassCache())
       {
-      TR::Options::_coldUpgradeSampleThreshold = 10;
+      TR::Options::_coldUpgradeSampleThreshold = 20;
```
dsouzai: Hm...I guess we need to take into account the fact that the user could set this differently.

dsouzai: Order of options processing:

JIT:
```
JIT_INITIALIZED
    onLoadInternal
        processOptionsAOT/processOptionsJIT
            fePreProcess
            jitPreProcess
            processOptions
```
SCC:
```
JIT_INITIALIZED
    if ((rc = j9shr_init(vm, loadFlags, &nonfatal)) != J9VMDLLMAIN_OK)
        vm->sharedClassConfig->runtimeFlags |= J9SHR_RUNTIMEFLAG_CACHE_INITIALIZATION_COMPLETE;
```
            
JIT:
```
ABOUT_TO_BOOTSTRAP
    aboutToBootstrap
        latePostProcessAOT/latePostProcessJIT
            latePostProcess
                feLatePostProcess
                    TR::Options::_coldUpgradeSampleThreshold = 20;
```
Basically:
`JIT_INITIALIZED`: preset options values and process user set options, but we don't know if we can AOT
`ABOUT TO BOOTSTRAP`: we know we can AOT, but we don't know if user specified options

Heh, @mpirvu could we just bite the bullet that under `-Xtune:virtualized` the user can't set `-Xjit:coldUpgradeSampleThreshold`?

dsouzai: I guess I could add an env var to prevent the override of this... Do you think it's worth it?

mpirvu: There is already this piece of code that tries to modify the threshold when AOT is used with SCC:
```
            // For AOT with GCR enabled we can be more conservative with
            // upgrades through sampling because we can rely on GCR
            // In this case we'll use the same threshold value as the one
            // employed in `setGlobalAggressiveAOT` (note that larger values
            // means more conservative upgrades and that the default value is
            // currently TR_DEFAULT_COLD_UPGRADE_SAMPLE_THRESHOLD==3)
            if (!self()->getOption(TR_DisableGuardedCountingRecompilations) &&
              _coldUpgradeSampleThreshold == TR_DEFAULT_COLD_UPGRADE_SAMPLE_THRESHOLD)
               _coldUpgradeSampleThreshold = 10;
```
You could do something very similar for your case, just test the Xtune:option. Of course you may want to test whether the user has set some other value by checking against the value set by Xtune (I would like something symbolic in this case, like TR_XTUNE_COLD_UPGRADE_SAMPLE_THRESHOLD)
This would be an OMR change.

dsouzai: Yeah looks reasonable, though I guess it would still not work if the user set the value to be the default set in `preprocess` (ie 10, since we wouldn't be able to tell if it was preset to 10 or user-set to 10; in the latter case we ideally shouldn't override it to 20)

> This would be an OMR change.

:sob:

andrewcraik:
@dsouzai is this dead? if so please close

dsouzai:
Yeah I don't know when I'm going to get around to validate this change, so I'll close for now and make a note to revisit this sometime.

Issue #5013, opened 2019-03-06
JIT GPU cudaInfo object still in use after being freed.


IBMJimmyk:
In the single GPU kernel case, the regionExitGPU C Helper calls freeGPUScope which frees the memory for the cudaInfo object. However, right after that is a call to the getStateGPU C Helper that uses the cudaInfo object that was just freed. This needs to be fixed such that cudaInfo is only freed after it is not needed anymore.

PR #5014, opened 2019-03-06
Fixed freeing of memory for cudaInfo for JIT GPU

IBMJimmyk:
In the single GPU kernel case, regionExitGPU calls freeGPUScope
which frees the memory for the cudaInfo object. The code then
calls getStateGPU with the exact same cudaInfo object that was
just freed. This is now changed so that the cudaInfo object is freed
inside getStateGPU instead.

regionExitGPU would also flush data to the CPU in some cases even after an
error case has been triggered. This was changed to no longer happen.

This change also cleans up the code by removing an unused parameter to
regionExitGPU and changing int return types from NULL_DEVICE_PTR to 0.

Closes: #5013

### DIFF
--- a/runtime/compiler/runtime/GPUHelpers.cpp
+++ b/runtime/compiler/runtime/GPUHelpers.cpp
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2000, 2018 IBM Corp. and others
+ * Copyright (c) 2000, 2019 IBM Corp. and others
  *
  * This program and the accompanying materials are made available under
  * the terms of the Eclipse Public License 2.0 which accompanies this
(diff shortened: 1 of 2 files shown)


IBMJimmyk:
WIP since I'm currently running some final tests.

IBMJimmyk:
The tests I ran all passed so I removed the WIP tag. Right now I just need the code reviewed before getting checked in.

runtime/compiler/runtime/GPUHelpers.cpp
```diff
@@ -1887,41 +1887,50 @@ bool freeGPUScope(CudaInfo *cudaInfo)
...
 
    return failedGPUSession;
    }
 
 extern "C" int
-regionExitGPU(CudaInfo *cudaInfo, uint8_t *startPC, int ptxSourceID, int flushBlock, void **liveRef)
+regionExitGPU(CudaInfo *cudaInfo, uint8_t *startPC, int ptxSourceID, void **liveRef)
    {
-   returnOnTrue(!cudaInfo, NULL_DEVICE_PTR);
+   returnOnTrue(!cudaInfo, 1);
```
gita-omr: Here we return 1 while we return 0 in other places. We definitely should not use a constant but some macro. I think we have some to indicate error/ok condition.

IBMJimmyk: Would it be better to have it return 0 instead of 1 on an error? Right now nothing looks at the return value. I changed this line to return 1 on error but it is also possible to do the reverse and change this line:
```
return (failedGPUSession ? 1 : 0);
``` 
to return 0 on error instead.

gita-omr: It's exactly my point:  I am not sure what 0 or 1 means in each situation. I think we need to return some macro or an enum, similar to some similar calls in this file.

gita-omr: Since @IBMJimmyk will open another PR for cleaning up return codes I am closing this.

gita-omr:
Jenkins test sanity aix,plinux jdk8,jdk11

IBMJimmyk:
The ECA check failed. I resubmitted my ECA. I'm not exactly sure how long it takes to get processed but the check should pass when that happens.

runtime/compiler/runtime/GPUHelpers.cpp
IBMJimmyk: I'm going to leave this as just a magic number of 1 for now since unclear is better than NULL_DEVICE_PTR which was wrong. A more detailed clean up of error codes is outside the scope of this PR. I will shortly be submitting a new PR with the more detailed error code clean up.

gita-omr:
Jenkins test extended plinux jdk8,jdk1

IBMJimmyk:
The ECA failure has been dealt with.

PR #4789, opened 2019-02-19
Refactor CompilationEnd and remoteCompilationEnd

harryyu1994:
Issue: #4401
[skip ci]

Signed-off-by: Harry Yu <harryyu1994@gmail.com>

PR #5025, opened 2019-03-07
Fail remote AOT compilation when relocation fails

dmitry-ten:
In `remoteCompilationEnd` we should be failing a compilation
if `prepareRelocateAOTCodeAndData` returned NULL, as it means
that relocation has failed. We already do that for non-AOT
compilations but forgot to add it for AOT.

### DIFF
--- a/runtime/compiler/control/JITaaSCompilationThread.cpp
+++ b/runtime/compiler/control/JITaaSCompilationThread.cpp
@@ -2589,6 +2589,14 @@ remoteCompilationEnd(
             }
          else
             {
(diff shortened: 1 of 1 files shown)


dmitry-ten:
Partially fixes relocation of remote AOT compiles. The other fix is in #4789.

harryyu1994:
https://github.com/eclipse/openj9/pull/5028 created a new PR for this. Just in case https://github.com/eclipse/openj9/pull/5028 doesn't get approved, then we would be going back to https://github.com/eclipse/openj9/pull/4789.

PR #5026, opened 2019-03-07
Implement TR_J9JITaaSServerSharedCache::addHint

dmitry-ten:
- We previously assumed that `addHint` is never called
on the server, but that assumption is incorrect.
Implemented the method.
- Fix caching of `J9ServerStream` pointer inside shared cache.
It's not correct to always use the stream pointer initialized in
`J9JITaaSServerSharedCache` constructor, because that constructor
is only called when front-end is created, which happens once
on the server.
But we create a new stream for each compilation, so the cached pointer
is incorrect. Update it at the beginning of each remote compilation.


dmitry-ten:
I'm now running into a segfault with this change that I haven't seen before. 
Investigating.

dmitry-ten:
The problem was that we were not updating stream object cached inside shared cache at the beginning of every compilation.
@ymanton, this is ready for review.

PR #5028, opened 2019-03-07
Set JITaaS client to not delay relocation of remote AOT

harryyu1994:
[skip ci]

### DIFF
--- a/runtime/compiler/control/JITaaSCompilationThread.cpp
+++ b/runtime/compiler/control/JITaaSCompilationThread.cpp
@@ -2493,45 +2493,8 @@ remoteCompilationEnd(
 
 #if defined(J9VM_INTERP_AOT_RUNTIME_SUPPORT)
 
-      TR_Debug *debug = TR::Options::getDebug();
-      bool canRelocateMethod = false;
-      if (debug)
-         {
-         TR_FilterBST *filter = NULL;
-         J9UTF8 *className = ((TR_ResolvedJ9Method*)comp->getCurrentMethod())->_className;
-         J9UTF8 *name = ((TR_ResolvedJ9Method*)comp->getCurrentMethod())->_name;
-         J9UTF8 *signature = ((TR_ResolvedJ9Method*)comp->getCurrentMethod())->_signature;
-         char *methodSignature;
-         char arr[1024];
-         int32_t len = J9UTF8_LENGTH(className) + J9UTF8_LENGTH(name) + J9UTF8_LENGTH(signature) + 3;
-         if (len < 1024)
-            methodSignature = arr;
-         else
-            methodSignature = (char *) TR_MemoryBase::jitPersistentAlloc(len);
-
-         if (methodSignature)
-            {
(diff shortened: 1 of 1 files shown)


harryyu1994:
@ymanton @dsouzai @mpirvu @dmitry-ten Dmitry and I have different opinions on this change. I think we should preserve the behaviour of how non-JITaaS AOT handles AOT relocations (I see that there are cases where we delay the AOT relocations to collect more Iprofiler Info before AOT load). Dmitry suggests that we should maybe always do AOT relocations right away after we receive the code from server. (Which is what this PR does) We are asking for your opinions on what's the correct way to deal with this since you guys know more about AOT/delay relocations/Iprofiler.

dsouzai:
The main (only?) purpose of delaying the load of an AOT body that has been compiled in the current run is for IProfiler information. If that's something that's important in the JITaaS world then you should preserve the behaviour. However, if the notion of sharing the IProfiler information from different Clients on the Server is something that's considered for the future, then you could relocate it immediately.

Since @mpirvu originally implemented the load delay, I'll defer this to him, but my feeling is that you want to preserve the behaviour until you have something like the ability to share IProfiler info.

dmitry-ten:
My main question here is: do we immediately run AOT compiles, or is it possible to compile but not actually run it until the next run? 
Also, when we run AOT compiles right after they finish, do we run them from code cache, or do we load them from SCC?
I assumed that we run from code cache, in which case we cannot delay relocations.

dsouzai:
> My main question here is: do we immediately run AOT compiles, or is it possible to compile but not actually run it until the next run?

By default we don't immediately run AOT compiles, we'll store it to the SCC, and after some time load it from the SCC into the code cache. You could easily have it never load by setting the extra field to something like -5 or whatever that value is to never translate.

> Also, when we run AOT compiles right after they finish, do we run them from code cache, or do we load them from SCC?

We load them from the SCC, validate/relocate them, and then put it into the code cache.

dmitry-ten:
I see. Then I guess we might want to support delayed relocations.
Then we'd need to handle some things differently. For example, CH table commit needs to know startPC of the relocated code. Not sure how to handle that if we delay relocations.

dsouzai:
> For example, CH table commit needs to know startPC of the relocated code. Not sure how to handle that if we delay relocations.

Isn't that something that happens in the non-JITaaS world too? Shouldn't JITaaS Client follow the same approach?

harryyu1994:
Maybe treat it the same as a "failed" attempt and not do anything for the CH table commit? and do the CH table commit when we do the actual AOT load and relocation. If that's how the non-JITaaS case does it.

dmitry-ten:
Ah, I see. we use `startPC` for creating method override assumptions. But in non-JITaaS AOT these assumptions are just not supported.

mpirvu:
The default behavior is to delay the AOT relocations which means that for some methods it's actually possible to create the AOT, but never load it (because the new counts that we encode haven't reached 0). However, normal JIT also supports non-delayed AOT relocations (through an options) and JITaaS must support both these cases as well.

harryyu1994:
I think the option is `TR_DisableDelayRelocationForAOTCompilations`. We already do support both cases, going to close this PR.

PR #5030, opened 2019-03-07
Enable CH table for remote AOT 

dmitry-ten:
- When fetching the entire persistent CH table
from the client, fetch it even if in AOT mode.
Before that, we wouldn't use persistent CH table
in AOT mode, and I see no reason why we shouldn't
be able to.
- Create SVM validation records when calling
`findClassInfoAfterLocking`.


dmitry-ten:
@ymanton ready for review.

ymanton:
I'm not up to date on how the regular JIT uses the CHTable in AOT mode these days. Does the regular JIT enable CHTable-based opts in AOT mode and are you making this change so that remote AOT behaves the same way?

dmitry-ten:
Yes, with SVM enabled, I believe that regular JIT should be able to perform all CH opts.
Without SVM it should still be able to run some Persistent CH table queries, although not all.

dmitry-ten:
For example,
`TR_PersistentCHTable::findInterfaceImplementers` should work even without SVM,
but `TR_PersistentCHTable::findSingleAbstractImplementer` requires SVM to be used.

PR #5035, opened 2019-03-08
AArch64: Add files in compiler/aarch64/{codegen,env}/ to aarch64.mk

knn-k:
This commit add .cpp files in compiler/aarch64/{codegen,env}/ to
JIT_PRODUCT_SOURCE_FILES in aarch64.mk.


knn-k:
These files are already in CMakeLists.txt in those directories.

0xdaryl:
This is an AArch64 JIT-specific change for which builds are not available yet.  No point running Jenkins tests.

PR #5036, opened 2019-03-08
Another fix to storeValidationRecordIfNecessary

dmitry-ten:
Don't store defining class to shared cache
if the class is NULL.


dmitry-ten:
@ymanton this is also ready for review.

Issue #5016, opened 2019-03-07
JITaaS in remote AOT mode doesn't disable unsupported optimizations


dmitry-ten:
Unsupported optimizations are disabled in `J9::Options::fePostProcessJIT`.
However, when run in AOT mode, they are re-enabled somewhere.
I observed that this can lead to crashes, for example, when server tries to use Known Object Table.
Since options specified on command line work as expected,
I suspect that the problem is due to option processing of `-Xaot` command line options.
We probably override disabled options sometime during processing of AOT options.

a7ehuo:
@dmitry-ten @mpirvu I'll take a look at this one.

a7ehuo:
The root cause of this issue is: although `TR_DisableKnownObjectTable` is set true for jitaas in `J9::Options::fePostProcessJIT`, the option is not the one that's used for the compilation request. A new option object is created for each compilation in `TR::CompilationInfoPerThreadBase::wrappedCompile()`. https://github.com/eclipse/openj9/blob/af8eeaa5638bcf67f812da613f5023bc300aa09d/runtime/compiler/control/CompilationThread.cpp#L7669.

PR #5040, opened 2019-03-08
Update options for compilation request

a7ehuo:
Update options for compilation request

A new option object is created for each compilation in `wrappedCompile()`. 
The new option object is constructed from `_aotCmdLineOptions` 
instead of `_jitCmdLineOptions` when AOT is used. However options for jitaa
were only set in `fePostProcessJIT()` but not in `fePostProcessAOT()`.
Add a helper method `setupJITaasOptions()` to be called by both JIT and AOT
option post processes. 

[skip ci]

Fixes: #5016


a7ehuo:
@ymanton @dmitry-ten FYI

dmitry-ten:
LGTM,
My only nitpick is that it's better to spell `JITaas` as `JITaaS`.

a7ehuo:
@ymanton Could you help review?

PR #5041, opened 2019-03-08
Separate java.vendor and java.vm.vendor

JasonFengJ9:
Separate `java.vendor` and `java.vm.vendor`

Removed the setting of `java.vm.vendor` with `java.vm` value;
Added `JAVA_VM_VENDOR` to be set for `java.vm.vendor` regardless of `java.vendor` values.

Additional notes:
1. For `JDK12+`, it can be built with `--with-vendor-name` and `--with-vendor-url`, and the values specified will be set to `java.vendor` and `java.vendor.url` via `java.lang.VersionProps.init(systemProperties)`;
2. For `JDK8 & 11`, change to following defines are required:
https://github.com/eclipse/openj9/blob/0bd089b36679fe0e3c5bfd63df0fcfc22c16c36d/runtime/include/vendor_version.h#L57-L58
just like [AdoptOpenJDK patch](https://github.com/AdoptOpenJDK/openjdk-build/blob/aefabfd2ac9f0eec39bcd78a4ac1633afce32605/git-hg/jdk8u/patches/0001-Set-vendor-information.patch#L18-L23);
3. Attempted to use configuration options `--with-vendor-name` for JDK 8 & 11 builds (particularly for JDK 11), it requires passing the value to `closed/OpenJ9.gmk` and to `include/j9cfg.h.ftl` via `uma.spec.properties` which requires more changes for legacy builds without incoming makefile options. `JDK12+` provides a generic way to set a customized `java.vendor` and it doesn't seem worth to duplicate the effort for earlier versions;
4. System property `java.vendor.url.bug` was set by `RI` but not `OpenJ9`, not adding it in this PR. 

closes: https://github.com/eclipse/openj9/issues/5002

Reviewer: @pshipton 
FYI: @DanHeidinga @smlambert  

### DIFF
--- a/jcl/src/java.base/share/classes/java/lang/System.java
+++ b/jcl/src/java.base/share/classes/java/lang/System.java
@@ -400,7 +400,6 @@ private static void ensureProperties(boolean isInitialization) {
 	java.lang.VersionProps.init(initializedProperties);
 	/* VersionProps.init(systemProperties) above sets java.specification.version value which is used to set java.vm.specification.version. */
 	initializedProperties.put("java.vm.specification.version", initializedProperties.get("java.specification.version")); //$NON-NLS-1$ //$NON-NLS-2$
-	initializedProperties.put("java.vm.vendor", initializedProperties.get("java.vendor")); //$NON-NLS-1$ //$NON-NLS-2$
 /*[ELSE]
 	/* VersionProps.init requires systemProperties to be set */
 	systemProperties = initializedProperties;
--- a/runtime/include/vendor_version.h
+++ b/runtime/include/vendor_version.h
@@ -48,9 +48,11 @@
 
 #define VENDOR_SHORT_NAME "OpenJ9"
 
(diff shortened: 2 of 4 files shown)


pshipton:
As I mentioned in https://github.com/eclipse/openj9/issues/5002#issuecomment-470240725
The following will also need to be fixed to use java.vm.vendor
https://github.com/eclipse/openj9/blob/bd0a4839cccedb3bc28716c6e67a67333ac983e2/test/functional/Java8andUp/src/org/openj9/test/support/Support_Exec.java#L143

JasonFengJ9:
@pshipton it has been fixed. Please have another look.

pshipton:
jenkins test sanity plinux jdk11,jdk12

pshipton:
@JasonFengJ9 please create a PR for the 0.13 branch.

pshipton:
The jdk12 failures are unrelated to this change.

PR #5050, opened 2019-03-11
Change and refactor targeted inlining heuristic

cathyzhyi:
This change refactored the logic in checkIfTargetInlineable to make
the code clearer. The heuristic around jsr292 methods are separated into
different inlining policy.

A few recognized small getters are added as inlineable targets in
targeted inlining.

https://github.com/eclipse/openj9/issues/4837

### DIFF
--- a/runtime/compiler/optimizer/J9Inliner.hpp
+++ b/runtime/compiler/optimizer/J9Inliner.hpp
@@ -36,9 +36,9 @@ class TR_J9InnerPreexistenceInfo;
  * Class TR_MultipleCallTargetInliner
  * ==================================
  *
- * Exception Directed Optimization (EDO) enables more aggressive 
- * inlining that the JIT does in cases when the call graph of the 
- * callee being inlined has some throw statements that would get 
+ * Exception Directed Optimization (EDO) enables more aggressive
+ * inlining that the JIT does in cases when the call graph of the
+ * callee being inlined has some throw statements that would get
  * caught in a catch block of the caller.
  *
  * The end goal is to give the JIT optimizer (value propagation in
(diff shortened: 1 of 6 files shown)


runtime/compiler/optimizer/J9Inliner.hpp
```diff
@@ -239,12 +239,16 @@ class TR_J9InlinerPolicy : public OMR_InlinerPolicy
...
+   friend class TR_J9InlinerPolicy;
    friend class TR_InlinerBase;
    friend class TR_MultipleCallTargetInliner;
    public:
       TR_J9JSR292InlinerPolicy(TR::Compilation *comp);
    protected:
       virtual TR_InlinerFailureReason checkIfTargetInlineable(TR_CallTarget* target, TR_CallSite* callsite, TR::Compilation* comp);
+      static bool isJSR292Methods(TR_ResolvedMethod *resolvedMethod);
+      static bool isJSR292AlwaysInlineableMethods(TR_ResolvedMethod *resolvedMethod);
+      static bool isJSR292SmallGetterMethods(TR_ResolvedMethod *resolvedMethod);
```
fjeremic: The pluralized methods does not make sense here given the parameter is only a single method. These APIs need to be documented.

runtime/compiler/optimizer/J9Inliner.hpp
```diff
@@ -239,12 +239,16 @@ class TR_J9InlinerPolicy : public OMR_InlinerPolicy
 class TR_J9JSR292InlinerPolicy : public TR_J9InlinerPolicy
    {
    friend class TR_J9InlinerUtil;
+   friend class TR_J9InlinerPolicy;
```
andrewcraik: Why the friend class? couldn't the helper methods be moved up if they are needed in the regular and 292 policies?

cathyzhyi: Thought conceptually JSR292 related queries should be in `TR_J9JSR292InlinerPolicy` but you are right this doesn't make sense in software design's point of view. Moved up the helper methods to TR_J9InlinerPolicy.

runtime/compiler/optimizer/J9Inliner.hpp
cathyzhyi: Changed query names as suggested and added documentation. thanks!

andrewcraik:
@cathyzhyi before we continue with the review could you provide some performance numbers to justify the change and any regressions it might cause. It makes sense to handle the method handles after the general methods to eliminate the handle chains. There is a risk of regression which we may have to accept for other gains. Is there an option that would revert the behaviour in case we need to test for a regression?

cathyzhyi:
> @cathyzhyi before we continue with the review could you provide some performance numbers to justify the change and any regressions it might cause. It makes sense to handle the method handles after the general methods to eliminate the handle chains. There is a risk of regression which we may have to accept for other gains. Is there an option that would revert the behaviour in case we need to test for a regression?

This change only refactors the code to use inliner policy to make the logic of deciding what can be inlined in each inlining more clear.  There is no change in code behavior.

andrewcraik:
@cathyzhyi is this still needed?

cathyzhyi:
Closing this PR as it's no longer needed.

PR #5052, opened 2019-03-11
Change RAT to lazy clean-up for performance

andrewcraik:
This change centralizes assumption Runtime Assumption Table clean-up logic by
converting all sites which remove assumptions to simply mark assumptions for
later removal. RAT manipulation functions in RuntimeAssumption are now protected
and accessed by friendship to prevent incorrect manipulation of the next pointers.
Finally, the clean-up routine has gained the ability to do a partial clean-up
which will be used to rate-limit clean-up to amortize the cost.

Rather than cleaning all assumptions at the end of a GC we will now
clean a limited number of assumptions at GC start (both scavenge and
global) to amortize the clean-up cost over time and minimize pauses.
When the JIT control state enters IDLE take the opportunity to eagerly
clean the RAT of any remaining dead assumptions since the CPU is not
required for the application at that moment. Clean-up should be short
especially since during GC cycles we will have already begun cleaning
the RAT.

depends eclipse/omr/pull/3657

### DIFF
--- a/runtime/compiler/runtime/ClassUnloadAssumption.cpp
+++ b/runtime/compiler/runtime/ClassUnloadAssumption.cpp
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2000, 2018 IBM Corp. and others
+ * Copyright (c) 2000, 2019 IBM Corp. and others
  *
  * This program and the accompanying materials are made available under
  * the terms of the Eclipse Public License 2.0 which accompanies this
--- a/runtime/compiler/runtime/ClassUnloadAssumption.cpp
+++ b/runtime/compiler/runtime/ClassUnloadAssumption.cpp
@@ -231,18 +231,28 @@ bool OMR::RuntimeAssumption::isAssumingMethod(void *md, bool reclaimPrePrologueA
 void OMR::RuntimeAssumption::dequeueFromListOfAssumptionsForJittedBody()
    {
    // We should not try to detach TR::SentinelRuntimeAssumption
-   TR_ASSERT(getAssumptionKind() != RuntimeAssumptionSentinel, "Sentinel assumptions cannot be detached");
-   OMR::RuntimeAssumption *crt = this->getNextAssumptionForSameJittedBody();
+   TR_ASSERT_FATAL(getAssumptionKind() != RuntimeAssumptionSentinel, "Sentinel assumptions cannot be detached");
+   OMR::RuntimeAssumption *crt = this->getNextAssumptionForSameJittedBodyEvenIfDead();
    OMR::RuntimeAssumption *prev = this;
    TR_ASSERT(crt, "Assumption must be queued when trying to detach");
    while (crt != this)
       {
-      // any such assumption (except the sentinel) must have an assumption location belonging to this body
-      prev = crt;
-      crt = crt->getNextAssumptionForSameJittedBody();
+      if (crt->isMarkedForDetach())
+         {
+         OMR::RuntimeAssumption *next = crt->getNextAssumptionForSameJittedBodyEvenIfDead();
+         prev->setNextAssumptionForSameJittedBody(next);
+         crt->setNextAssumptionForSameJittedBody(NULL);
+         crt = next;
+         }
+      else
+         {
+         // any such assumption (except the sentinel) must have an assumption location belonging to this body
+         prev = crt;
+         crt = crt->getNextAssumptionForSameJittedBodyEvenIfDead();
+         }
       }
    // Now I completed a full circle
-   prev->setNextAssumptionForSameJittedBody(crt->getNextAssumptionForSameJittedBody());
+   prev->setNextAssumptionForSameJittedBody(crt->getNextAssumptionForSameJittedBodyEvenIfDead());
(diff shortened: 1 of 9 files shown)


andrewcraik:
This change is marked WIP because performance evaluation is ongoing, but I would like some early feedback from reviewers on the implementation while the evaluation is completed. FYI @mpirvu

andrewcraik:
Because of the extensive API changes I don't think there is a good way to stage this change - it requires breaking changes in both OMR and OpenJ9 so some kind of coordination is going to be required to avoid breaking the OpenJ9 build.

andrewcraik:
Travis fail is just due to the dependent change in OMR

andrewcraik:
documentation updated

andrewcraik:
@mpirvu could I get an opinion on this change before I do the full perf regression?

runtime/compiler/runtime/ClassUnloadAssumption.cpp
```diff
@@ -379,53 +365,61 @@ void TR_RuntimeAssumptionTable::markForDetachFromRAT(OMR::RuntimeAssumption *ass
...
 
    assumptionTableMutex->enter();
-   for (kind=0; kind < LastAssumptionKind; kind++) // for each table
+   for (int kind=0; _marked > 0 && cleanupCount != 0 && kind < LastAssumptionKind; kind++) // for each table
       {
       if (_detachPending[kind] == true)  // Is there anything to remove from this table?
          {
-         hashTable = _tables + kind;
-         for (size_t i = 0; i < hashTable->_spineArraySize; ++i) // for each bucket in the table
+         TR_RatHT *hashTable = _tables + kind;
```
mpirvu: Just an idea: every invocation of `reclaimMarkedAssumptionsFromRAT` could start from a different table (or different bucket) to avoid the situation where the later buckets/tables get larger because they have a lower priority in evacuation.

runtime/compiler/runtime/HookHelpers.cpp
```diff
@@ -413,12 +413,16 @@ void jitRemoveAllMetaDataForClassLoader(J9VMThread * vmThread, J9ClassLoader * c
    classLoader->jitMetaDataList = NULL;
    }
 
-void jitReclaimMarkedAssumptions()
+void jitReclaimMarkedAssumptions(bool isEager)
    {
-   static const char *useOldRAReclaim = feGetEnv("TR_useOldRAReclaim");
-   if (!useOldRAReclaim)
+   static char *forceAggressiveRATCleaning = feGetEnv("TR_forceAggressiveRATCleaning");
+   if (isEager && forceAggressiveRATCleaning)
```
andrewcraik:
```suggestion
   if (isEager || forceAggressiveRATCleaning)
```

runtime/compiler/runtime/HookHelpers.cpp
```diff
@@ -413,12 +413,16 @@ void jitRemoveAllMetaDataForClassLoader(J9VMThread * vmThread, J9ClassLoader * c
    classLoader->jitMetaDataList = NULL;
    }
 
-void jitReclaimMarkedAssumptions()
+void jitReclaimMarkedAssumptions(bool isEager)
    {
-   static const char *useOldRAReclaim = feGetEnv("TR_useOldRAReclaim");
-   if (!useOldRAReclaim)
+   static char *forceAggressiveRATCleaning = feGetEnv("TR_forceAggressiveRATCleaning");
```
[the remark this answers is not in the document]
andrewcraik: no you are right there is a bug - the isEager && forceAggressiveRATCleaning should be isEager || forceAggressiveRATCleaning and the else if can become an else I'll make an update shortly

runtime/compiler/runtime/ClassUnloadAssumption.cpp
andrewcraik: Could I do this as a follow-on change since I worry non-determinism or extra complexity could make debugging harder?

andrewcraik:
I have fixed the logic in `jitReclaimMarkedAssumptions` per suggestion.

runtime/compiler/runtime/ClassUnloadAssumption.cpp
mpirvu: Sure.It was a suggestion anyway.

andrewcraik:
Performance testing on DayTrader 3 shows no regression either for throughput or startup. I have checked other IBM benchmarks and there are no regressions seen so I think this is ready to move forward. Note it does REQUIRE a coordinated merge because the API definition lives in OMR but all of the implementation of that API is in OpenJ9 and the changes to the API are complex and non-trivial to stage.

andrewcraik:
@fjeremic would you mind doing a review of the delivery of this and launch a build if you are happy. The delivery between OMR and OpenJ9 would need to be delivered because the API header file for the RAT is in OMR and all the implementation is in OpenJ9 at present. Staging this could be quite messy so I'd like a coordinated merge.

fjeremic:
Jenkins test sanity all depends eclipse/omr/#3657

fjeremic:
Jenkins test sanity all depends eclipse/omr#3657

pshipton:
jenkins restarted, please rerun PR testing tomorrow

pshipton:
Jenkins test sanity all depends eclipse/omr#3657

PR #5054, opened 2019-03-11
Changed AUTO_DETECT variable from on/off into true/false

Yuehan-Lin:
[ci skip]
Signed-off-by: Yuehan-Lin <Yuehan.Lin@ibm.com>


llxia:
Jenkins test sanity all jdk8

llxia:
Jenkins test sanity all jdk11

PR #5057, opened 2019-03-11
Fix calculateFieldDataStart in flat field case

tajila:
Fix calculateFieldDataStart in flat field case

Fix the `calculateFieldDataStart` code to account for flattened 
fields that contain doubles.

### DIFF
--- a/runtime/vm/ObjectFieldInfo.hpp
+++ b/runtime/vm/ObjectFieldInfo.hpp
@@ -342,11 +342,15 @@ public:
 	{
 		U_32 fieldDataStart = 0;
 		if (!isContendedClassLayout()) {
-			fieldDataStart = getSuperclassFieldsSize();
+			bool doubleAlignment = (_totalDoubleCount > 0);
+#ifdef J9VM_OPT_VALHALLA_VALUE_TYPES
+			doubleAlignment = doubleAlignment || (_totalFlatFieldDoubleBytes > 0);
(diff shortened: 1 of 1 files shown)


tajila:

tajila:

gacholio:
jenkins test sanity xlinux jdk12

runtime/vm/ObjectFieldInfo.hpp
```diff
@@ -342,11 +342,16 @@ class ObjectFieldInfo {
 	{
 		U_32 fieldDataStart = 0;
 		if (!isContendedClassLayout()) {
-			fieldDataStart = getSuperclassFieldsSize();
+#ifdef J9VM_OPT_VALHALLA_VALUE_TYPES
```
gacholio: I think this might read better with an |= in the VT case.

tajila: C++ doesn't have logical assignments, so I did `|=`.

gacholio: Is that strictly valid? I suppose a bitwise OR does give the correct result.

gacholio: Maybe "x = x || y" instead...

tajila: done

gacholio:
Earlier testing was sufficient so I'm not launching any more.

PR #5064, opened 2019-03-12
Modifty Metronome.hpp for moving into OMR

jason-hall:

### DIFF
--- a/runtime/gc_realtime/Metronome.hpp
+++ b/runtime/gc_realtime/Metronome.hpp
@@ -1,6 +1,5 @@
-
 /*******************************************************************************
- * Copyright (c) 1991, 2017 IBM Corp. and others
+ * Copyright (c) 1991, 2019 IBM Corp. and others
  *
  * This program and the accompanying materials are made available under
  * the terms of the Eclipse Public License 2.0 which accompanies this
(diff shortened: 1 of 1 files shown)


runtime/gc_realtime/Metronome.hpp
```diff
@@ -25,18 +24,15 @@
...
+#include "omr.h"
+#include "omrcfg.h"
 
 #include <assert.h>
 #define ASSERT_LEVEL 0
 #define assert1(expr) assert((ASSERT_LEVEL < 1) || (expr));
 #define assert2(expr) assert((ASSERT_LEVEL < 2) || (expr));
 
-#define MAX_UINT ((UDATA) (-1))
+#define MAX_UINT ((uintptr_t) (-1))
```
dmitripivkine: This definition exists in OMR code already in `AllocationContextSegregated.cpp`
Also OMR has `UDATA_MAX` definition for this used widely.
Also there are multiple places in the code with `((uintptr_t) (-1))` and `((UDATA) (-1))`
@charliegracie I wondering may be it is time to do cleanup and use single proper definition everywhere (done by two separate PRs - one for OMR another for OpenJ9) ?

PR #5065, opened 2019-03-12
Correctly check if string compression is enabled

dmitry-ten:
- Add a front-end query to check if client has
string compression enabled in VM.
- Replace dereference of client-side address with
a front-end query.


dmitry-ten:
Ok, I'll change the name.

dmitry-ten:
@ymanton could you review?

PR #5067, opened 2019-03-12
Remove workaround in Access.defineClass()

keithc-ca:
This is a replay of #4919 for the v0.13.0-release branch.

* Add RAM class flag J9ClassIsExemptFromValidation which is set for
  MagicAccessorImpl and its subclasses which makes them exempt for
  some access and validation rules (like classes defined via unsafe
  API).

* Add and use macro: J9CLASS_IS_EXEMPT_FROM_VALIDATION.

* New flag, J9_FINDCLASS_FLAG_NAME_IS_INVALID, is used to signal
  an invalid class name if the actual loaded class is not exempt.
  If the class being defined is not exempt, defineClassCommon()
  will now return NULL without a pending exception.
  ClassLoader_defineClassImpl() throws an exception only
  in that case.

* Allow JavaLangAccess to define a class with a null ProtectionDomain.

* Map proxy classes to the proper module.

* defineClassCommon() now calls fixUnsafeMethods() as necessary

* fix loop start index in verifyQualifiedName()


gacholio:

gacholio:
PR builds in branches don't work, so I'm going to merge this and look at the nightly.

Issue #5069, opened 2019-03-12
Assert found shared cache with option disabled 


a7ehuo:
Hitting the following assert on the JITaaS server side when killing the server. 

```
Assertion failed at /root/home/ahuo/src/openj9-openjdk-jdk8/openj9/runtime/compiler/env/VMJ9.cpp:767: TR::Options::sharedClassCache()
	Found shared cache with option disabled
```

This assert might not apply to the server since -Xshareclasses is not specified on the server side but on the client side.  The shared cache is created as long as `TR_EnableJITaaSRemoteAOT` is set on the server side

PR #5070, opened 2019-03-12
Fix the assert that does not apply to JITaaS remote AOT

a7ehuo:
When freeing the shared cache in `TR_J9VMBase::freeSharedCache()`,
the assert checks if `-Xshareclasses` is enabled. This does not
apply to the JITaaS server which uses the environment
variable `TR_EnableJITaaSRemoteAOT`.

[skip ci]

Fixes: #5069

### DIFF
--- a/runtime/compiler/env/VMJ9.cpp
+++ b/runtime/compiler/env/VMJ9.cpp
@@ -735,8 +735,9 @@ TR_J9VMBase::TR_J9VMBase(
          }
 
    _sharedCache = NULL;
+   static bool isRemoteAOT = feGetEnv("TR_EnableJITaaSRemoteAOT") != NULL;
    if (TR::Options::sharedClassCache() ||
-       (compInfo->getPersistentInfo()->getJITaaSMode() == SERVER_MODE && feGetEnv("TR_EnableJITaaSRemoteAOT")))
+       (compInfo->getPersistentInfo()->getJITaaSMode() == SERVER_MODE && isRemoteAOT))
       // shared classes and AOT must be enabled, or we should be on the JITaaS server with remote AOT enabled
       {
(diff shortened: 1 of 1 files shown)


a7ehuo:
@ymanton Could you help review? Thanks.

runtime/compiler/env/VMJ9.cpp
```diff
@@ -764,7 +764,14 @@ TR_J9VMBase::freeSharedCache()
    {
    if (_sharedCache)        // shared classes and AOT must be enabled
       {
-      TR_ASSERT(TR::Options::sharedClassCache(), "Found shared cache with option disabled");
+      if (_compInfo && (_compInfo->getPersistentInfo()->getJITaaSMode() == SERVER_MODE))
+         {
+         TR_ASSERT(feGetEnv("TR_EnableJITaaSRemoteAOT"), "Found shared cache with remote AOT disabled.");
```
ymanton: The `freeSharedCache` method looks like it will be called every time a thread is destroyed, which can add up to a lot of calls. Unfortunately `feGetEnv` leaks memory so it's best to declare a static variable to hold the return value so that `feGetEnv` will only be called once.

```suggestion
         static bool isRemoteAOT = feGetEnv("TR_EnableJITaaSRemoteAOT") != NULL;
         TR_ASSERT(isRemoteAOT, "Found shared cache with remote AOT disabled.");
```

Can you also fix this instance while you're at it?

https://github.com/eclipse/openj9/blob/69199c55d85917c4b885ec098dfab672be263f6e/runtime/compiler/env/VMJ9.cpp#L739

a7ehuo:
@ymanton Updated. Please check. Thanks.

PR #5072, opened 2019-03-12
Fix incorrect iTable build

gacholio:
The iTable building code was considering all methods in the interface
class, not just the ones that are appropriate for the iTable. This
results in an incorrect iTable shape (and very possibly writing off the
end of the allocated iTable chunk).

Process only methods from the interface which make up the interface
contract (public methods that appear in the vTable).

[ci skip]


gacholio:

gacholio:

PR #5075, opened 2019-03-13
Remove reference to non-existent jar

keithc-ca:
The project references a JRE; we don't need the jar that no longer exists.


DanHeidinga:
Jenkins compile xlinux jdk8

PR #5079, opened 2019-03-13
Check the stack size of a stackmap frame (v0.13)

ChengJin01:
The change is to check the stack size of a stackmap frame
against the maximal stack size specified in the class file
so as to capture invalid stack size during static verification.


ChengJin01:
Reviewer: @DanHeidinga

PR #5080, opened 2019-03-13
JDK12 j.l.Class.toGenericString() appends type parameters

### LINKED ISSUE #4938: JTReg Test Failure : java/lang/Class/GenericStringTest.java
Consistent failures ...

```
[2019-02-27T15:39:06.434Z] STDERR:
[2019-02-27T15:39:06.434Z] Unexpected Class.toGenericString output; expected 
[2019-02-27T15:39:06.434Z] 	'public class java.util.EnumMap<K extends java.lang.Enum<K>,V>',
[2019-02-27T15:39:06.434Z]  got 
[2019-02-27T15:39:06.434Z] 	'public class java.util.EnumMap<K,V>'.
[2019-02-27T15:39:06.434Z] Unexpected Class.toGenericString output; expected 
[2019-02-27T15:39:06.434Z] 	'public abstract class java.util.EventListenerProxy<T extends java.util.EventListener>',
[2019-02-27T15:39:06.434Z]  got 
[2019-02-27T15:39:06.434Z] 	'public abstract class java.util.EventListenerProxy<T>'.
[2019-02-27T15:39:06.434Z] Unexpected Class.toGenericString output; expected 
[2019-02-27T15:39:06.434Z] 	'public abstract class java.lang.Enum<E extends java.lang.Enum<E>>',
[2019-02-27T15:39:06.434Z]  got 
[2019-02-27T15:39:06.434Z] 	'public abstract class java.lang.Enum<E>'.
[2019-02-27T15:39:06.434Z] java.lang.RuntimeException
[2019-02-27T15:39:06.434Z] 	at GenericStringTest.main(GenericStringTest.java:78)
[2019-02-27T15:39:06.434Z] 	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
[2019-02-27T15:39:06.434Z] 	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
[2019-02-27T15:39:06.434Z] 	at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
[2019-02-27T15:39:06.434Z] 	at java.base/java.lang.reflect.Method.invoke(Method.java:567)
[2019-02-27T15:39:06.434Z] 	at com.sun.javatest.regtest.agent.MainActionHelper$AgentVMRunnable.run(MainActionHelper.java:298)
[2019-02-27T15:39:06.434Z] 	at java.base/java.lang.Thread.run(Thread.java:825)
```

... observed with test https://github.com/ibmruntimes/openj9-openjdk-jdk12/blob/openj9/test/jdk/java/lang/Class/GenericStringTest.java against OpenJDK12+OpenJ9 build.

See [consoleText.txt](https://github.com/eclipse/openj9/files/2919012/consoleText.txt) and [consoleText.txt](https://github.com/eclipse/openj9/files/2919051/consoleText.txt).

See https://github.com/ibmruntimes/openj9-openjdk-jdk12/commit/d6b3d6b959d8381d0f240bc710ac6797e71dff44#diff-dcb8c538740ef99677c90b8e39b63199, suspect something is off with the JCL classes.

JasonFengJ9:
JDK12 `j.l.Class.toGenericString()` appends type parameters

For each type parameter, skip the case that it has only one bound and the bound is `java.lang.Object`, otherwise append ` extends ` following by type name and ` & ` if there are more than one bound.

Manually verified that this PR passes the test in https://github.com/eclipse/openj9/issues/4938

fixes #4938 

Reviewer: @DanHeidinga 

### DIFF
--- a/jcl/src/java.base/share/classes/java/lang/Class.java
+++ b/jcl/src/java.base/share/classes/java/lang/Class.java
@@ -2337,6 +2337,18 @@ private void appendTypeParameters(StringBuilder nameBuilder) {
 			if (comma) nameBuilder.append(',');
 			nameBuilder.append(t);
 			comma = true;
+/*[IF Java12]*/
+			Type[] types = t.getBounds();
+			if (types.length == 1 && types[0].equals(Object.class)) {
+				// skip in case the only bound is java.lang.Object
+			} else {
+				String prefix = " extends "; //$NON-NLS-1$
+				for (Type type : types) {
+					nameBuilder.append(prefix).append(type.getTypeName());
+					prefix = " & "; //$NON-NLS-1$
+				}
+			}
+/*[ENDIF] Java12 */
 		}
 		nameBuilder.append('>');
 	}


pshipton:
jenkins test sanity zlinux jdk12

jcl/src/java.base/share/classes/java/lang/Class.java
```diff
@@ -2337,6 +2337,23 @@ private void appendTypeParameters(StringBuilder nameBuilder) {
...
+				// skip in case the only bound is java.lang.Object
+			} else {
+				boolean addExtends = true;
+				for (Type type : types) {
+					if (addExtends) {
+						nameBuilder.append(" extends "); //$NON-NLS-1$
+						addExtends = false;
+					} else {
+						nameBuilder.append(" & "); //$NON-NLS-1$
+					}
```
keithc-ca: I don't remember who initially suggested the pattern to me, but if you replace `boolean addExtends` with `String prefix` the loop is much simpler with in internal branching:
```
String prefix = " extends "; //$NON-NLS-1$
for (Type type : types) {
	nameBuilder.append(prefix).append(type.getTypeName());
	prefix = " & "; //$NON-NLS-1$
}
```

JasonFengJ9: Thanks for the suggestion. The code snippet proposed is going to create a string object for `" & "` regardless it is required to be appended or not (the usage without `&` are more common than those with it). This code carries more footprint than a primitive `boolean` variable which probably is not desirable.

keithc-ca: No, the string literal is in the constant pool in both your version and my proposal.

JasonFengJ9: Respectfully disagree. This PR does not create a string object `" & "` when there is only one `type` to be processed. The proposed code always creates such an object even when that string is not required to be appended.

DanHeidinga: @JasonFengJ9 you're correct that if no one calls toGenericString() on a class with multiple type parameters the `" & "` will never be needed.  @keithc-ca is also correct that it's a one time cost - the first user of multiple type parameters pays the cost to create / intern it and everyone else benefits there after.

My preference is code patterns that are simpler to read & debug so I prefer Keith's as most code interacting with this api will pay for that second allocation at some point.

JasonFengJ9: Agreed. The string is going to be created/intern for future usage of `" & "`. Allowing this one time cost to favour the code pattern.
PR has been updated, please review.

PR #5081, opened 2019-03-13
Modify OSInterface for moving into OMR

jason-hall:


dmitripivkine:

PR #5084, opened 2019-03-13
use UDATA to replace int to avoid windows compiling issue

LinHu2016:
- use UDATA to replace int to avoid windows compiling issue


amicic:
jenkins compile win jdk11

PR #5086, opened 2019-03-14
AArch64: Add arm64nathelp.m4 in runtime/codert_vm

knn-k:
This commit adds a new file arm64nathelp.m4 in runtime/codert_vm,
and make changes to related files.  The declared functions do not
work at this point.


gacholio:
The only thing here that will affect PR builds is the UMA file exclusion, which I'll just say is correct by inspection, so no builds required.

Issue #5077, opened 2019-03-13
Remote AOT segfault


dmitry-ten:
I'm frequently getting the following segfault when running AcmeAir on lara perf machine:
```
#0  0x00007f39854dd9b1 in pthread_kill () from /lib64/libpthread.so.0
#1  0x00007f3982f83eed in omrdump_create (portLibrary=0x7f3983dc9380 <j9portLibrary>, filename=0x7f398063f880 "/home/dmitry-ten/", dumpType=<optimized out>, userData=<optimized out>)
    at ../../omr/port/unix/omrosdump.c:188
#2  0x00007f39826e84e2 in doSystemDump (agent=0x7f397c025ad0, label=0x7f398063f880 "/home/dmitry-ten/", context=0x7f398063fd30) at dmpagent.c:718
#3  0x00007f39826e4365 in protectedDumpFunction (portLibrary=portLibrary@entry=0x7f3983dc9380 <j9portLibrary>, userData=userData@entry=0x7f398063f7e0) at dmpagent.c:2876
#4  0x00007f3982f56ec8 in omrsig_protect (portLibrary=0x7f3983dc9380 <j9portLibrary>, fn=0x7f39826e4350 <protectedDumpFunction>, fn_arg=0x7f398063f7e0, handler=0x7f39826e4370 <signalHandler>, handler_arg=0x0,
    flags=125, result=0x7f398063f7d8) at ../../omr/port/unix/omrsignal.c:384
#5  0x00007f39826e796b in runDumpFunction (agent=<optimized out>, label=0x7f398063f880 "/home/dmitry-ten/", context=<optimized out>) at dmpagent.c:2850
#6  0x00007f39826e7aec in runDumpAgent (vm=vm@entry=0x7f397c010310, agent=agent@entry=0x7f397c025ad0, context=context@entry=0x7f398063fd30, state=state@entry=0x7f398063fd28,
    detail=detail@entry=0x7f398063fdb0 "", timeNow=timeNow@entry=1552420882963) at dmpagent.c:2776
#7  0x00007f39826fe55b in triggerDumpAgents (vm=0x7f397c010310, self=0x22a3400, eventFlags=8192, eventData=<optimized out>) at trigger.c:1028
#8  0x00007f398366f6f2 in generateDiagnosticFiles (portLibrary=portLibrary@entry=0x7f3983dc9380 <j9portLibrary>, userData=userData@entry=0x7f3980640290) at gphandle.c:1170
[...]

PR #5099, opened 2019-03-14
Always use non-AOT front-end in packRemoteROMClassInfo

dmitry-ten:
Previously, if `packRemoteROMClassInfo` was called
during AOT compilation, all of the values for class info
would be obtained by calling front-end methods of a
Shared Cache VM.
This is a problem, because shared cache versions of the
methods attempt to validate classes and may return NULL.
This is wrong for 2 reasons:
1. We perform validation on the server.
2. Values packed in `packRemoteROMClassInfo` are cached
on the server, for the entire lifetime of a client.
If a shared cache version of a method sets class info
parameter to NULL, then it will remain NULL on the server,
even for non-AOT compilations. This is wrong, because
non-AOT requires most parameters to be non-NULL.

Fixes: #5077

### DIFF
--- a/runtime/compiler/control/JITaaSCompilationThread.cpp
+++ b/runtime/compiler/control/JITaaSCompilationThread.cpp
@@ -1001,7 +1001,7 @@ bool handleServerMessage(JITaaS::J9ClientStream *client, TR_J9VM *fe)
       case J9ServerMessageType::ResolvedMethod_getRemoteROMClassAndMethods:
          {
          J9Class *clazz = std::get<0>(client->getRecvData<J9Class *>());
-         client->write(JITaaSHelpers::packRemoteROMClassInfo(clazz, fe, trMemory));
+         client->write(JITaaSHelpers::packRemoteROMClassInfo(clazz, fe->vmThread(), trMemory));
          }
          break;
       case J9ServerMessageType::ResolvedMethod_isJNINative:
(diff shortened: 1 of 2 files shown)


ymanton:
Do AOT compilations on the server look for or depend on some of these values being NULL, though? Or can they tolerate these values being non-NULL? I would not be surprised if we had code all over the JIT that checks some value and assumes that if NULL the compilation has to be an AOT compilation, otherwise a JIT compilation.

dmitry-ten:
As far as I can tell, all of the cached values are only used in front-end methods like this one:
```
TR_OpaqueClassBlock *
TR_J9SharedCacheServerVM::getBaseComponentClass(TR_OpaqueClassBlock * classPointer, int32_t & numDims)
   {
   TR::Compilation* comp = _compInfoPT->getCompilation();
   TR_ASSERT(comp, "Should be called only within a compilation");

   bool validated = false;
   TR_OpaqueClassBlock *baseComponent = TR_J9ServerVM::getBaseComponentClass(classPointer, numDims);

   if (comp->getOption(TR_UseSymbolValidationManager))
      {
      SVM_ASSERT_ALREADY_VALIDATED(comp->getSymbolValidationManager(), baseComponent);
      validated = true;
      }
   else
      {
      if (((TR_ResolvedRelocatableJ9JITaaSServerMethod *) comp->getCurrentMethod())->validateArbitraryClass(comp, (J9Class *) classPointer))
         validated = true;
      }

   if (validated)
      return baseComponent;
   else
      return classPointer;  // not sure about this return value, but it's what we used to return before we got "smart"
   }
```
That is, we always call the non-AOT version of the method, which won't return NULL, and determine whether the return value should be NULL depending on whether we could validate the class. Thus, we will return NULL in AOT compilations when needed.

PR #5101, opened 2019-03-14
Look at field declaring class when verify fields

cathyzhyi:
All the fields with the same declaring class are aliased together.
For example, class A is the parent class having field A.f and class B
and class C are 2 subclasses of A. A.f, B.f C.f are all aliased
together. Therefore to verify if a known object x is compatible with a
field symRef C.f we should check if object is instance of the declaring
class A but not the symRef C.f 's field class C. The object would still be
valid if it's class B.


andrewcraik:

PR #5104, opened 2019-03-14
Bridge methods should also be 'synthetic'

keithc-ca:
Without the synthetic bit, eclipse treats bridge methods as if they introduce ambiguity.
The intent is to put the generated classes in a jar that can be on the classpath for the DDR_VM project.


DanHeidinga:
@keithc-ca Any suggestions on the right level of PR build for this change?

keithc-ca:
> @keithc-ca Any suggestions on the right level of PR build for this change?

I don't expect this to affect anything yet. The old pointer class generator is still being used at build time and pointers are not yet created dynamically at runtime. Even just a basic compile should be sufficient.

DanHeidinga:
Jenkins compile xlinux jdk8

PR #5106, opened 2019-03-14
Fix handling of 'volatile' qualifier

keithc-ca:
* there isn't necessarily a space after 'const' or 'volatile'
  e.g. the owner field of a monitor has type 'J9Thread*volatile'


pshipton:

PR #5117, opened 2019-03-15
Add J9ARRAYCLASS_SET_STRIDE and J9ARRAYCLASS_GET_STRIDE macros

### LINKED ISSUE #5021: Valhalla: Add support for flattened valueType arrays
When creating a non-primitive array, all fields are currently ref sized (32 or 64bits). Each slot in the array contains a reference to the element.

With valueTypes the lack of identity enables the JVM to create flattened arrays. These are arrays where the array elements are not stored as references, rather, they are stored directly (flattened) into the array.

Not flattened
```
----------------------
|       array header   |
----------------------                  ----------------------  
|       reference       |  -->          |        element         |
----------------------                  ----------------------  
|       reference       |  -->          |        element        |
----------------------                  ---------------------- 
```

Flattened
 ```
----------------------
|       array header  |
----------------------          
|      element        |  
----------------------          
|      element       |  
----------------------          
```

For now we will treat all valueType arrays as flattenable (later when we add the distinction between Q and L values in the constantPool we can change this). Note: flattenable does not mean flattened, it means it is allowed to be flattened.

1) Array Classes
------------------
A valueType is flattened if the `J9ClassIsFlattened` flag is set in the J9Class->classFlags. This is the same flag that will determine if the valueType array will be flattened or not. In order to do this for arrays we need to make sure that the classFlags of the elementType are copied over to the arrayRamClass (this may already be the case, but it is worth double checking).

Also, we need to add the `flattenedClassCache` to `J9ArrayClass`, this was an oversight in a previous PR.

2) Creating arrays
--------------------
Add a simple test to ValueTypeTests.java that creates a Value type array. ie.

`Array.newInstance(FlattenedLine2Dclass, ....)` 

Array creation is done with the anewarray bytecode. There are two functions used to create arrays in the interpreter `inlineAllocateIndexableObject` and `J9AllocateIndexableObject` (which uses `MM_IndexableObjectAllocationModel` for the actual allocation). In both cases the arrayClass is used to determine the elementType. And in both cases the size calculation must be changed if the element type is flattened. We can introduce a new macro for this (ie `J9_IS_J9CLASS_FLATTENED`) see j9.h for examples.

Classes may have padding bytes added before the first field in order to satisfy alignment constraints. The flattened layout may or may not include this padding. As a result, in order to determine the flattened size of a valuetype, one must take the `J9Class->totalInstanceSize` and subtract it by the `J9Class->backfillOffset`. A macro can also be added for this.

When calculating the size of the each element, one must ensure that the alignment constraints of the type are met (see `J9ClassLargestAlignmentConstraintDouble` and `J9ClassLargestAlignmentConstraintReference` in J9Class->classFlags). This might mean that extra space must be allocated for each element in order to make sure that all alignment constraints are met. This should also be taken into account when calculating the size of the array. 

Size calculation is done as follows:
```
if VT contains doubles //long, double
    elementSize = ROUND_UP_TO(instanceSize - offsetOfFirstField, sizeof(U_64))
else if VT contains ref
    elementSize = ROUND_UP_TO(instanceSize - offsetOfFirstField, sizeof(fj9object_t))
else  // VT only contains singles (int, short, etc) at this point
    elementSize = instanceSize - offsetOfFirstField
```
This ensures that we pad up the size of the elements so alignment constraints are met.

Note: J9 indexable object headers are guaranteed to end on a 8byte boundary, so the starting point of the array elements should always satisfy alignment constraints, we only need to worry about the sizing of the elements.

The `J9ARRAYCLASS_SET\GET_STRIDE` have been added to set/get the stride length on array objects. All array classes will need their stride lengths set upon creation. This can be done in `internalCreateRAMClassFromROMClass`. For all non-VT array classes the stride length will be `1 << elementClass->romClass->instanceShape`. For VT-arrays stride length will be calculated as demonstrated above.

The `dataSize` calculation in `inlineAllocateIndexableObject` will have to be updated to use `strideLength` from the arrayClass. Like wise with `GC_ArrayletObjectModel::getDataSizeInBytes(..)`.

3) Accessing array elements
-----------------------------
Accessing object array elements is done with the aaload and aastore bytecodes. Both of these use access barrier functions (ObjectAccessBarrierAPI.hpp) to read/store array elements. These functions need to detect if the array is flattened so that it read/stores the elements at the correct location.

Flattened array elements are variable length, so they must be read and written to with the `copyObjectFields` helper. 

It is possible that the size of a flattened type in an array will differ than the size of a flattened type in a field (or on its own). In order to keep track of the element size we can use the flattenedClassCache (currently unsued) field in the arrayRamClass to store that information.

4) Scanning flattened arrays
------------------------------
Finally, the GC object scanners need to be updated to take into account that array elements may contained mixed objects. Typically, arrays are either primitive, in which case the GC can ignore the elements when scanning for objects, or they are object arrays in which case the GC scans every slot. With flattened arrays, it is a combination of the two. The GC needs to use the instance description bits of the element type in order to determine which slots to scan.

5) Testing flattened arrays
---------------------------
We will need basic testing for flattened array support, this includes creating an array, then reading and writing fields out of it.  We will need further testing to ensure that the GC is able to scan array elements properly. We will need to test multiple GC policies (ie, gencon, balanced and metronome) for this.

------------------------------
Note: This work should be done in stages (1 - 5), at least 1 PR for each stage. Also, this is design is subject to change and will likely be modified in the future to accommodate the Q and L array distinction. I will also add more details on scanning array fields.

March 8: Updates made to section 1, 2 and 4
March 18: Udates made to section 2

### PR DESCRIPTION
**Pull request by AidanHa**
Add J9ARRAYCLASS_SET_STRIDE and J9ARRAYCLASS_GET_STRIDE macros

Add UDATA flattenedElementSize in J9ArrayClass (temporarily) for consistency

Also, separately confirm that the classFlags of the elementTypes are copied over to the arrayRamClass (#5021 (comment))

Related #5021 step 1

### DIFF
--- a/runtime/oti/j9nonbuilder.h
+++ b/runtime/oti/j9nonbuilder.h
@@ -2954,6 +2954,9 @@ typedef struct J9Class {
 #define J9INTERFACECLASS_ITABLEMETHODCOUNT(clazz) ((clazz)->totalInstanceSize)
 #define J9INTERFACECLASS_SET_ITABLEMETHODCOUNT(clazz, value) (clazz)->totalInstanceSize = (value)
 
+#define J9ARRAYCLASS_SET_STRIDE(clazz, strideLength) ((clazz)->flattenedClassCache) = (J9FlattenedClassCache*)(UDATA)(strideLength)
+#define J9ARRAYCLASS_GET_STRIDE(clazz) ((UDATA)((clazz)->flattenedClassCache))
+
 typedef struct J9ArrayClass {
 	UDATA eyecatcher;
 	struct J9ROMClass* romClass;
@@ -3003,6 +3006,10 @@ typedef struct J9ArrayClass {
 #if defined(J9VM_OPT_VALHALLA_NESTMATES)
 	struct J9Class* nestHost;
 #endif /* defined(J9VM_OPT_VALHALLA_NESTMATES) */
+#if defined(J9VM_OPT_VALHALLA_VALUE_TYPES)
+	/* Added temporarily for consistency */
+	UDATA flattenedElementSize;
+#endif /* defined(J9VM_OPT_VALHALLA_VALUE_TYPES) */
 } J9ArrayClass;
 
 typedef struct J9HookedNative {

### CODE REVIEW

**Comment by tajila** [comment]
@gacholio Please review this change

**Comment by gacholio** [comment]
In fact, re-using totalInstanceSize for array classes is probably your best bet.

Note that I plan to remove the extra class types in favour of macros in the near future.

File: runtime/oti/j9nonbuilder.h [hunk 1]
```diff
@@ -2954,6 +2954,11 @@ typedef struct J9Class {
 #define J9INTERFACECLASS_ITABLEMETHODCOUNT(clazz) ((clazz)->totalInstanceSize)
 #define J9INTERFACECLASS_SET_ITABLEMETHODCOUNT(clazz, value) (clazz)->totalInstanceSize = (value)
 
+#if defined(J9VM_OPT_VALHALLA_VALUE_TYPES)
+#define J9ARRAYCLASS_SET_STRIDE(clazz, strideLength) *((UDATA*)&(clazz->flattenedClassCache)) = (UDATA) strideLength;
```
tajila: semi-colons are not needed for both macros

gacholio: I would prefer as cast to the final type, rather than casting the result location. Casting pointers to odd things can result in aliasing errors.

runtime/oti/j9nonbuilder.h
```diff
@@ -2954,6 +2954,11 @@ typedef struct J9Class {
 #define J9INTERFACECLASS_ITABLEMETHODCOUNT(clazz) ((clazz)->totalInstanceSize)
 #define J9INTERFACECLASS_SET_ITABLEMETHODCOUNT(clazz, value) (clazz)->totalInstanceSize = (value)
 
+#if defined(J9VM_OPT_VALHALLA_VALUE_TYPES)
+#define J9ARRAYCLASS_SET_STRIDE(clazz, strideLength) *((UDATA*)&(clazz->flattenedClassCache)) = (UDATA) strideLength
+#define J9ARRAYCLASS_GET_STRIDE(clazz) *((UDATA*)&(clazz->flattenedClassCache))
```
gacholio: Better to fetch the value and cast it at the end, rather than casting the source.

runtime/oti/j9nonbuilder.h
```diff
@@ -2954,6 +2954,11 @@ typedef struct J9Class {
 #define J9INTERFACECLASS_ITABLEMETHODCOUNT(clazz) ((clazz)->totalInstanceSize)
 #define J9INTERFACECLASS_SET_ITABLEMETHODCOUNT(clazz, value) (clazz)->totalInstanceSize = (value)
 
+#if defined(J9VM_OPT_VALHALLA_VALUE_TYPES)
+#define J9ARRAYCLASS_SET_STRIDE(clazz, strideLength) *((UDATA*)&(clazz->flattenedClassCache)) = (UDATA) strideLength
```
gacholio: Cast the result and store, rather than casting the target location.

runtime/oti/j9nonbuilder.h
```diff
@@ -2954,6 +2954,11 @@ typedef struct J9Class {
 #define J9INTERFACECLASS_ITABLEMETHODCOUNT(clazz) ((clazz)->totalInstanceSize)
 #define J9INTERFACECLASS_SET_ITABLEMETHODCOUNT(clazz, value) (clazz)->totalInstanceSize = (value)
 
+#if defined(J9VM_OPT_VALHALLA_VALUE_TYPES)
```
gacholio: This should probably be available always, not just with VT.

gacholio:
Should probably update the title & description to reflect the new code.

runtime/oti/j9nonbuilder.h
gacholio: The stride for normal arrays is 1 << arrayROMClass->instanceShape.

gacholio:
Random aside - it's always worth adding `[ci skip]` to the PR description to avoid the automated build - the CI build only tests cmake at the moment, so if you haven't added/removed files, there's really no need for it, since most PRs will run real compile/test before merge.

runtime/oti/j9nonbuilder.h
```diff
@@ -2954,6 +2954,9 @@ typedef struct J9Class {
 #define J9INTERFACECLASS_ITABLEMETHODCOUNT(clazz) ((clazz)->totalInstanceSize)
 #define J9INTERFACECLASS_SET_ITABLEMETHODCOUNT(clazz, value) (clazz)->totalInstanceSize = (value)
 
+#define J9ARRAYCLASS_SET_STRIDE(clazz, strideLength) (clazz->flattenedClassCache) = (J9FlattenedClassCache*)(UDATA) strideLength
```
gacholio: Last nitpick (hopefully) - macro parameters should be bracketed in the macro definitions.

`(clazz)->flattenedClassCache`

gacholio:
If this PR is not completely resolving #5021 then the description should not say "Fixes" - I personally use "Related".

runtime/oti/j9nonbuilder.h
AidanHa: Should I be adding another macro for normal arrays in that case?

gacholio: I don't think so - the existing macros are fine. Once you start setting the stride for VT arrays, you may as well set it for the other array types as well.

tajila:
@AidanHa `strideLength` also needs the brackets. 

Also, can you please make the commit message the same as the PR description

tajila:
@gacholio please take another look at these changes

runtime/oti/j9nonbuilder.h
```diff
@@ -2954,6 +2954,9 @@ typedef struct J9Class {
 #define J9INTERFACECLASS_ITABLEMETHODCOUNT(clazz) ((clazz)->totalInstanceSize)
 #define J9INTERFACECLASS_SET_ITABLEMETHODCOUNT(clazz, value) (clazz)->totalInstanceSize = (value)
 
+#define J9ARRAYCLASS_SET_STRIDE(clazz, strideLength) ((clazz)->flattenedClassCache) = (J9FlattenedClassCache*)(UDATA) (strideLength)
```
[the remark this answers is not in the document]
AidanHa: Fixed

gacholio:
jenkins test compile win jdk8

gacholio:
jenkins compile win jdk8

PR #5119, opened 2019-03-15
Modify WorkPacketsRealtime for moving into OMR

jason-hall:

### DIFF
--- a/runtime/gc_realtime/WorkPacketsRealtime.cpp
+++ b/runtime/gc_realtime/WorkPacketsRealtime.cpp
@@ -1,4 +1,3 @@
-
 /*******************************************************************************
  * Copyright (c) 1991, 2019 IBM Corp. and others
  *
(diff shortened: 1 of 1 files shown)


runtime/gc_realtime/WorkPacketsRealtime.cpp
```diff
@@ -21,19 +20,16 @@
...
+#include "omrcfg.h"
 #include "omrthread.h"
 
-#include "Debug.hpp"
 #include "EnvironmentRealtime.hpp"
-#include "GCExtensions.hpp"
+#include "GCExtensionsBase.hpp"
 #include "IncrementalOverflow.hpp"
 #include "IncrementalParallelTask.hpp"
-#include "RememberedSetSATB.hpp"
```
dmitripivkine: Well, if you remove this line you should add `#include "WorkPacketsSATB.hpp"` explicitly here

dmitripivkine: ok, never mind. I missed the fact that MM_WorkPacketsSATB is it's parent.

dmitripivkine:
Jenkins compile win

PR #5122, opened 2019-03-15
Fix temp slot bug in valueType generator

tajila:
Fix temp slot bug in valueType generator

When creating a reference type, the valueType generator does not 
account for doubles when placing the new ref type in a temp slot.

In addition the max locals specified for the method does not take
into account the temp slot required for the new ref type. This PR
fixes both those issues.


tajila:
@llxia Please review this change

llxia:
will forgo PR build as it will not test Valhalla tests atm.

Issue #5108, opened 2019-03-14
JITaaSCHTableCommit crash in jitaas remote AOT mode


harryyu1994:
JITaaS Client side crash.
`metaData=0x0` dereferencing a null pointer in `JITaaSCHTableCommit` caused the crash

```
#13 0x00007ff8470f9baf in JITaaSCHTableCommit (comp=0x7ff843d78000, 
    metaData=0x0, data=std::tuple containing = {...})
    at /root/Harry_JITaaS/openj9-openjdk-jdk8/openj9/runtime/compiler/env/JITaaSCHTable.cpp:233
#14 0x00007ff847073a83 in remoteCompile (vmThread=0x2125300, 
    compiler=0x7ff843d78000, compilee=0x7ff844e74948, method=0x2150f28, 
    details=..., compInfoPT=0x7ff844e8a070)
    at /root/Harry_JITaaS/openj9-openjdk-jdk8/openj9/runtime/compiler/control/JITaaSCompilationThread.cpp:2368
#15 0x00007ff846e284f0 in TR::CompilationInfoPerThreadBase::compile (
    this=0x7ff844e8a070, vmThread=0x2125300, compiler=0x7ff843d78000, 
    compilee=0x7ff844e74948, vm=..., optimizationPlan=0x7ff844ef24e0, 
    scratchSegmentProvider=...)
    at /root/Harry_JITaaS/openj9-openjdk-jdk8/openj9/runtime/compiler/control/CompilationThread.cpp:8738
```

dmitry-ten:
This is the same problem that #5028 tried to deal with.
The crash happens because relocations are not done at the end of the compilation, but
`remoteCompile` assumes that either relocations are successful and `metaData` is not NULL, or compilation fails.

harryyu1994:
Working on a fix for this one.

harryyu1994:
It turns out `CHTable` is not used by AOT. CHTable commit should return true right away AOT compilations.

harryyu1994:
Going to need another change to disable the sending of chtable data generated by compilation from server to client. Looking into it.

harryyu1994:
Fixed by #5126

PR #5126, opened 2019-03-15
Disable JITaaSCHTableCommit for Remote AOT

harryyu1994:
- CHTable commit is not used by AOT, thus disabling it for JITaaS Remote AOT to prevent the null pointer dereference crash
- Remove one line of unused code

Issue: #5108

### DIFF
--- a/runtime/compiler/env/JITaaSCHTable.cpp
+++ b/runtime/compiler/env/JITaaSCHTable.cpp
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2018, 2018 IBM Corp. and others
+ * Copyright (c) 2018, 2019 IBM Corp. and others
  *
  * This program and the accompanying materials are made available under
  * the terms of the Eclipse Public License 2.0 which accompanies this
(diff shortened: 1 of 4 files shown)


harryyu1994:
@ymanton @mpirvu This one needs an review.

ymanton:
If we don't need to call `JITaaSCHTableCommit()` then is there any reason to execute the rest of the code in the following block for remote AOT?

https://github.com/eclipse/openj9/blob/3d1af1337bb99f039f1dbbedcf7ea9fb8af484cf/runtime/compiler/control/JITaaSCompilationThread.cpp#L2348-L2378

harryyu1994:
I think we can skip over the whole thing, didn't catch that.

harryyu1994:
Change updated

runtime/compiler/env/JITaaSCHTable.cpp
```diff
@@ -221,6 +221,9 @@ bool JITaaSCHTableCommit(
       TR_MethodMetaData *metaData,
       CHTableCommitData &data)
    {
+   if (comp->fej9()->isAOT_DEPRECATED_DO_NOT_USE())
+      return true;
```
ymanton: Can we replace this with an assert now since we don't expect to call it in AOT mode?

harryyu1994:
Change updated

ymanton:
Merged, thanks.

PR #5133, opened 2019-03-18
AArch64: Implement arm64nathelp.m4 and arm64helpers.m4

knn-k:
This commit fills the unimplemented parts of arm64nathelp.m4 and
arm64helpers.m4.

### DIFF
--- a/runtime/codert_vm/arm64nathelp.m4
+++ b/runtime/codert_vm/arm64nathelp.m4
@@ -34,8 +34,9 @@ define({CALL_SLOW_PATH_ONLY_HELPER_NO_RETURN_VALUE},{
 	SAVE_ALL_REGS
 	str x30,[J9VMTHREAD,{#}J9TR_VMThread_jitReturnAddress]
 	CALL_C_WITH_VMTHREAD(old_slow_$1)
-	cbz x0,+8
+	cbz x0,.L_done_$1
 	ret x0
+.L_done_$1:
 	RESTORE_ALL_REGS
 	RESTORE_FPLR
 	SWITCH_TO_JAVA_STACK
@@ -126,7 +127,12 @@ define({OLD_DUAL_MODE_HELPER},{
 	CALL_C_WITH_VMTHREAD(old_fast_$1)
 	cbz x0,.L_done_$1
 	SAVE_C_NONVOLATILE_REGS
-	hlt {#}0
+	mov x1,x0
+	mov x0,J9VMTHREAD
+	blr x1
+	cbz x0,.L_old_slow_$1
+	ret x0
+.L_old_slow_$1:
 	RESTORE_C_NONVOLATILE_REGS
 .L_done_$1:
 	RESTORE_C_VOLATILE_REGS
@@ -145,7 +151,12 @@ define({OLD_DUAL_MODE_HELPER_NO_RETURN_VALUE},{
 	CALL_C_WITH_VMTHREAD(old_fast_$1)
 	cbz x0,.L_done_$1
 	SAVE_C_NONVOLATILE_REGS
-	hlt {#}0
+	mov x1,x0
+	mov x0,J9VMTHREAD
+	blr x1
+	cbz x0,.L_old_slow_$1
+	ret x0
+.L_old_slow_$1:
 	RESTORE_C_NONVOLATILE_REGS
 .L_done_$1:
 	RESTORE_C_VOLATILE_REGS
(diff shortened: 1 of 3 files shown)


knn-k:
This is a followup to #5086.
The instruction `hlt #0` was inserted in `arm64nathelp.m4` to indicate unimplemented parts, and now they are replaced with actual code.

runtime/oti/arm64helpers.m4
```diff
@@ -82,41 +82,121 @@ define({CALL_C_WITH_VMTHREAD},{
...
+	ldp d22,d23,JIT_GPR_SAVE_SLOT(22)
+	ldp d24,d25,JIT_GPR_SAVE_SLOT(24)
+	ldp d26,d27,JIT_GPR_SAVE_SLOT(26)
+	ldp d28,d29,JIT_GPR_SAVE_SLOT(28)
+	ldp d30,d31,JIT_GPR_SAVE_SLOT(30)
 })
 
-dnl To be filled
 define({SAVE_C_NONVOLATILE_REGS},{
+	stp x16,x17,JIT_GPR_SAVE_SLOT(16)
```
0xdaryl: x16, x17, and x18 are special purpose registers in the 64-bit ABI, but they are temporary registers (I'm reading this in section 5.1.1).  According to the ABI, this makes them caller-preserved.  Hence, I don't think they belong here.

runtime/oti/arm64helpers.m4
```diff
@@ -82,41 +82,121 @@ define({CALL_C_WITH_VMTHREAD},{
...
-define({SAVE_PRESERVED_REGS},{
-})
-
 define({RESTORE_ALL_REGS},{
 	RESTORE_C_VOLATILE_REGS
 	RESTORE_C_NONVOLATILE_REGS
 })
 
-dnl To be filled
+define({SAVE_PRESERVED_REGS},{
```
0xdaryl: If my understanding is correct, `SAVE_PRESERVED_REGS` and `RESTORE_PRESERVED_REGS` refers to the JIT private linkage.  If you're not doing anything different from the ABI linkage then my comments above about special purpose registers apply here as well.

runtime/codert_vm/arm64nathelp.m4
```diff
@@ -126,7 +126,11 @@ define({OLD_DUAL_MODE_HELPER},{
 	CALL_C_WITH_VMTHREAD(old_fast_$1)
 	cbz x0,.L_done_$1
 	SAVE_C_NONVOLATILE_REGS
-	hlt {#}0
+	mov x1,x0
+	mov x0,J9VMTHREAD
+	blr x1
+	cbz x0,+8
```
0xdaryl: Rather than hard-coding the branch displacement `+8`, would it be clearer to branch to an actual named label?  It makes the code easier to follow.  Unless you have another reason for hard-coding this?  This comment applies to the other `cbz` instructions in the other macros as well.

runtime/oti/arm64helpers.m4
```diff
@@ -82,41 +82,121 @@ define({CALL_C_WITH_VMTHREAD},{
...
 define({SAVE_C_VOLATILE_REGS},{
+	stp x0,x1,JIT_GPR_SAVE_SLOT(0)
+	stp x2,x3,JIT_GPR_SAVE_SLOT(2)
+	stp x4,x5,JIT_GPR_SAVE_SLOT(4)
+	stp x6,x7,JIT_GPR_SAVE_SLOT(6)
+	stp x8,x9,JIT_GPR_SAVE_SLOT(8)
+	stp x10,x11,JIT_GPR_SAVE_SLOT(10)
+	stp x12,x13,JIT_GPR_SAVE_SLOT(12)
+	stp x14,x15,JIT_GPR_SAVE_SLOT(14)
+	stp d0,d1,JIT_GPR_SAVE_SLOT(0)
```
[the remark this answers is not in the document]
knn-k: That is correct.  It is a simple mistake from copy&paste.

runtime/codert_vm/arm64nathelp.m4
knn-k: Added labels based on those in `pnathelp.m4`.

runtime/oti/arm64helpers.m4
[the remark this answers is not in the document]
knn-k: `JIT_FPR_SAVE_SLOT(n)` exceeds the range of the immediate offset in `stp` instruction when `n` is 12 or larger.  Changing the code.

runtime/oti/arm64helpers.m4
knn-k: Moved x16-x18 to volatile registers.

runtime/oti/arm64helpers.m4
knn-k: Removed x16-x18 from preserved registers.
I will visit this again when I define the JIT private linkage.

knn-k:
I removed the lines for `d8`-`d15` from `SAVE_C_NONVOLATILE_REGS` and `RESTORE_C_NONVOLATILE_REGS`, from looking at those macros for `p` and `arm`.  I will revisit it when I define the JIT private linkage.

0xdaryl:
@gacholio, I'd appreciate your approval on this one too please.

Issue #5131, opened 2019-03-15
JITaaS remote AOT mode doesn't handle delayed AOT load properly


harryyu1994:
After an AOT compilation finishes, a lot of times AOT relocations are intentionally failed to delay the loading of AOT code. Current JITaaS remote AOT mode does not consider this case.

harryyu1994:
Currently looking at this issue.

The following code snippet needs to be incorporated into JITaaS remote AOT
```
// AOT compilations can fail on purpose because we want to load
// the AOT body later on. This case is signalled by having a metaData != 0
// but a startPC == entry->_oldStartPC == 0
if (metaData && !entry->_oldStartPC && !startPC && _compInfo.getPersistentInfo()->getJITaaSMode() != SERVER_MODE)
   {
   TR_ASSERT(_vm->isAOT_DEPRECATED_DO_NOT_USE(), "compilationEnd() can fail only for relocating AOT compilations\n");
   TR_ASSERT(!entry->_oldStartPC, "We expect compilationEnd() to fail only for AOT compilations which are first time compilations\n");

   // Replenish the counts of the method
   // We are holding the compilation monitor at this point
   //
   J9ROMMethod *romMethod = J9_ROM_METHOD_FROM_RAM_METHOD(method);
   if (!(romMethod->modifiers & J9AccNative))// Never change the extra field of a native method
      {
      int32_t methodVMExtra = TR::CompilationInfo::getJ9MethodVMExtra(method);
      if (methodVMExtra == 1 || methodVMExtra == J9_JIT_QUEUED_FOR_COMPILATION)
         {
         // We want to use high counts unless the user specified counts on the command line
         // or he used useLowerMethodCounts (or Xquickstart)
         int32_t count;
         if (TR::Options::getCountsAreProvidedByUser() || TR::Options::startupTimeMatters() == TR_yes)
            count = getCount(romMethod, _compiler->getOptions(), _compiler->getOptions());
         else
            count = J9ROMMETHOD_HAS_BACKWARDS_BRANCHES(romMethod) ? TR_DEFAULT_INITIAL_BCOUNT : TR_DEFAULT_INITIAL_COUNT;

         TR::CompilationInfo::setInvocationCount(method, count);
         if (TR::Options::getCmdLineOptions()->getVerboseOption(TR_VerbosePerformance))
            {
            // compiler must exist because startPC != NULL
[...]

PR #5143, opened 2019-03-18
Handle delayed AOT load for JITaaS Remote AOT

harryyu1994:
AOT compilations can fail on purpose because we want to load the AOT body later on. This case is signalled by having a successful compilation but canRelocateMethod == false. When we get a delayed AOT load, we want to manipulate the invocation count (just like how non-JITaaS AOT handles it). Also, logCompilationSuccess requires additional handling as this special case has a NULL metaData.
[skip ci]
Issue: #5131

### DIFF
--- a/runtime/compiler/control/JITaaSCompilationThread.cpp
+++ b/runtime/compiler/control/JITaaSCompilationThread.cpp
@@ -2379,7 +2379,7 @@ remoteCompile(
 
          TR::compInfoPT->relocateThunks();
 
-         TR_ASSERT(!metaData->startColdPC, "coldPC should be null");
+         TR_ASSERT(!metaData || !metaData->startColdPC, "coldPC should be null");
 
          int compilationSequenceNumber = compiler->getOptions()->writeLogFileFromServer(logFileStr);
          if (TR::Options::getCmdLineOptions()->getOption(TR_EnableJITaaSDoLocalCompilesForRemoteCompiles) && compilationSequenceNumber)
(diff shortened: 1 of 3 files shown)


harryyu1994:
The difference between JITaaS remote AOT and non-jitaas AOT is that JITaaS remote AOT doesn't have a metaData created before AOT relocation while the non-JITaaS AOT does. The compilation of the method generates the metadata and, in JITaaS's case, JITaaS server generates it while the client won't have access to it until client relocates it. And this is why we need additional handling for remote AOT.

harryyu1994:
`logCompilationSuccess` needs metaData which we don't have at the client when we don't relocate. We need to either skip `logCompilationSuccess` for remote AOT client or modify things in it.

runtime/compiler/control/JITaaSCompilationThread.cpp
```diff
@@ -2573,6 +2573,37 @@ remoteCompilationEnd(
             comp->failCompilation<J9::AOTRelocationFailed>("Failed to relocate");
             }
          }
+      else
+         {
```
mpirvu: This code is a copy-paste from CompilationThread.cpp. Let's create a method that is called twice in relevant places. Also, in line 2503 there is a debug object being defined but not being used anywehere.

runtime/compiler/control/CompilationThread.hpp
```diff
@@ -49,6 +49,11 @@
 #include "env/J2IThunk.hpp"
 #include "env/PersistentCollections.hpp"
 
+extern "C" {
+   int32_t getCount(J9ROMMethod *romMethod, TR::Options *optionsJIT, TR::Options *optionsAOT);
```
mpirvu: I would prefer this not to be included in the hpp file. If you refactor the code like I suggested below and put the refactored method in CompilationThread.cpp, then you can eliminate these lines

runtime/compiler/control/JITaaSCompilationThread.cpp
harryyu1994: The debug object was spotted and removed in another PR that hasn't been merged yet.

harryyu1994:
Ready for review and merge. @mpirvu

Issue #5012, opened 2019-03-06
StackTraceElement.toString() missing lock information


pdbain-ibm:
The reference implementation includes lock information in the stack trace (see below for test program):
```
$ java locker.java 

testMonitorsThread:
"testMonitorsThread" prio=5 Id=12 BLOCKED on java.lang.Object@2d1ef81a owned by "main" Id=1
	at locker.syncMethod(locker.java:39)
	-  blocked on java.lang.Object@2d1ef81a
	at locker$1.run(locker.java:16)
	at java.base@11/java.lang.Thread.run(Thread.java:834)

main:
"main" prio=5 Id=1 RUNNABLE
	at java.management@11/sun.management.ThreadImpl.dumpThreads0(Native Method)
	at java.management@11/sun.management.ThreadImpl.getThreadInfo(ThreadImpl.java:466)
	at locker.printThreadInfo(locker.java:34)
	at locker.meth(locker.java:25)
	-  locked java.lang.Object@2d1ef81a
	at locker.main(locker.java:46)
	...
```

but not OpenJ9:
```

testMonitorsThread:
"testMonitorsThread" prio=5 Id=25 BLOCKED
	at locker.syncMethod(locker.java:39)
	at locker$1.run(locker.java:16)
	at java.base/java.lang.Thread.run(Thread.java:825)


main:
"main" prio=5 Id=1 RUNNABLE
	at java.management/com.ibm.java.lang.management.internal.ThreadMXBeanImpl.getMultiThreadInfoImpl(Native Method)
	at java.management/com.ibm.java.lang.management.internal.ThreadMXBeanImpl.getThreadInfoCommon(ThreadMXBeanImpl.java:322)
	at java.management/com.ibm.java.lang.management.internal.ThreadMXBeanImpl.getThreadInfo(ThreadMXBeanImpl.java:283)
	at locker.printThreadInfo(locker.java:34)
	at locker.meth(locker.java:25)
	at locker.main(locker.java:46)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.base/java.lang.reflect.Method.invoke(Method.java:566)
[...]

PR #5150, opened 2019-03-19
Implement jstack

pdbain-ibm:
Create general tooling to convert mxbeans to properties and vice versa.  Add
directories and modules as required: for Java 8 some modules go into rt.jar and
others to tools.jar.  Add service providers to allows java.base to access
classes from foreign modules.  Update module exports, classpath and jpp
configuration.  Reorganize tools directory to ease filtering source
directories.  Add tests for jstack.

Needs coordinated changes in extensions repositories.

- https://github.com/ibmruntimes/openj9-openjdk-jdk8/pull/279
- https://github.com/ibmruntimes/openj9-openjdk-jdk11/pull/146
- https://github.com/ibmruntimes/openj9-openjdk-jdk12/pull/35
- https://github.com/ibmruntimes/openj9-openjdk-jdk/pull/90

Fixes https://github.com/eclipse/openj9/issues/4858

### DIFF
--- a/jcl/src/java.base/share/classes/com/ibm/tools/attach/target/Attachment.java
+++ b/jcl/src/java.base/share/classes/com/ibm/tools/attach/target/Attachment.java
@@ -36,6 +36,9 @@ import java.security.PrivilegedAction;
 import java.util.Objects;
 import java.util.Properties;
 
+import java.util.ServiceLoader;
+import openj9.tools.attach.diagnostics.base.DiagnosticProperties;
+import openj9.tools.attach.diagnostics.spi.TargetDiagnosticsProvider;
 /*[IF Sidecar19-SE]*/
 import jdk.internal.vm.VMSupport;
 /*[ELSE] Sidecar19-SE
@@ -60,6 +63,7 @@ final class Attachment extends Thread implements Response {
 	private String attachError;
 	private final AttachHandler handler;
 	private final String key;
+	private TargetDiagnosticsProvider diagProvider;
 	private static final String START_REMOTE_MANAGEMENT_AGENT = "startRemoteManagementAgent"; //$NON-NLS-1$
 	private static final String START_LOCAL_MANAGEMENT_AGENT = "startLocalManagementAgent"; //$NON-NLS-1$
 
@@ -106,7 +110,7 @@ final class Attachment extends Thread implements Response {
 		portNumber = rc.getPortNumber();
 		this.key = rc.getKey();
 		this.handler = attachHandler;
-
+		diagProvider = null;
 		setDaemon(true);
 	}
 
@@ -199,6 +203,12 @@ final class Attachment extends Thread implements Response {
 					AttachmentConnection.streamSend(respStream, Response.ERROR
 							+ " " + attachError); //$NON-NLS-1$
 				}
+			} else if (cmd.startsWith(Command.GET_THREAD_GROUP_INFO)) {
+				try {
+					replyWithProperties(getDiagnosticsProvider().getThreadGroupInfo());
+				} catch (Exception e) {
+					replyWithProperties(makeExceptionProperties(e));
+				}
 			} else if (cmd.startsWith(Command.GET_SYSTEM_PROPERTIES)) {
 				Properties internalProperties = com.ibm.oti.vm.VM.getVMLangAccess().internalGetProperties();
 				String argumentString = String.join(" ", com.ibm.oti.vm.VM.getVMArgs()); //$NON-NLS-1$
@@ -260,8 +270,35 @@ final class Attachment extends Thread implements Response {
 		return false;
 	}
 
+	private static Properties makeExceptionProperties(Exception e) {
+		Properties props = new Properties();
+		props.put(IPC.PROPERTY_DIAGNOSTICS_ERROR, Boolean.toString(true));
+		props.put(IPC.PROPERTY_DIAGNOSTICS_ERRORTYPE, e.getClass().getName());
+		String msg = e.getMessage();
+		if (null != msg) {
+			props.put(IPC.PROPERTY_DIAGNOSTICS_ERRORMSG, msg);
+		}
+		return props;
+	}
+
+	private void replyWithProperties(DiagnosticProperties props) throws IOException {
+		replyWithProperties(props.toProperties());
+	}
+
 	private void replyWithProperties(Properties props) throws IOException {
 		IPC.sendProperties(props, responseStream);
+	}
+	
+	private TargetDiagnosticsProvider getDiagnosticsProvider() {
+		if (diagProvider == null) {
+			for (TargetDiagnosticsProvider p: ServiceLoader.load(TargetDiagnosticsProvider.class)) {
+				if (null != p) {
+					diagProvider = p;
+					break;
+				}
+			}
+		}
+		return diagProvider;
 	}
 
 	/**
--- a/jcl/src/java.base/share/classes/com/ibm/tools/attach/target/IPC.java
+++ b/jcl/src/java.base/share/classes/com/ibm/tools/attach/target/IPC.java
@@ -2,7 +2,7 @@
 package com.ibm.tools.attach.target;
 
 /*******************************************************************************
- * Copyright (c) 2009, 2018 IBM Corp. and others
+ * Copyright (c) 2009, 2019 IBM Corp. and others
  *
  * This program and the accompanying materials are made available under
  * the terms of the Eclipse Public License 2.0 which accompanies this
@@ -31,6 +31,7 @@ import java.io.OutputStream;
 import java.io.PrintStream;
 import java.security.SecureRandom;
 import java.io.File;
+import java.nio.charset.StandardCharsets;
 import java.nio.file.Files;
 import java.nio.file.LinkOption;
 import java.nio.file.Paths;
@@ -45,13 +46,14 @@ import static java.nio.file.attribute.PosixFilePermission.GROUP_READ;
 import static java.nio.file.attribute.PosixFilePermission.GROUP_WRITE;
 import static java.nio.file.attribute.PosixFilePermission.OTHERS_READ;
 import static java.nio.file.attribute.PosixFilePermission.OTHERS_WRITE;
+import static openj9.tools.attach.diagnostics.base.DiagnosticsInfo.OPENJ9_DIAGNOSTICS_PREFIX;
 
 /**
  * Utility class for operating system calls
  */
 public class IPC {
 
-	private static final String JAVA_IO_TMPDIR = "java.io.tmpdir";
+	private static final String JAVA_IO_TMPDIR = "java.io.tmpdir"; //$NON-NLS-1$
 	/**
 	 * Successful return code from natives.
 	 */
@@ -67,6 +69,13 @@ public class IPC {
 	static final int LOGGING_UNKNOWN = 0;
 	static final int LOGGING_DISABLED = 1;
 	static final int LOGGING_ENABLED = 2;
+	/**
+	 * Prefixes and names for Diagnostics properties
+	 */
+	public static final String INCOMPATIBLE_JAVA_VERSION = "OPENJ9_INCOMPATIBLE_JAVA_VERSION"; //$NON-NLS-1$
+	public final static String PROPERTY_DIAGNOSTICS_ERROR = OPENJ9_DIAGNOSTICS_PREFIX + "error"; //$NON-NLS-1$
+	public final static String PROPERTY_DIAGNOSTICS_ERRORTYPE = OPENJ9_DIAGNOSTICS_PREFIX + "errortype"; //$NON-NLS-1$
+	public final static String PROPERTY_DIAGNOSTICS_ERRORMSG = OPENJ9_DIAGNOSTICS_PREFIX + "errormsg"; //$NON-NLS-1$
 	/**
 	 * True if operating system is Windows.
 	 */
@@ -483,6 +492,13 @@ public class IPC {
 		defaultVmId = id;
 	}
 
+	/**
+	 * Send a properties file as a stream of bytes
+	 * 
+	 * @param props     Properties file
+	 * @param outStream destination of the bytes
+	 * @throws IOException on communication error
+	 */
 	public static void sendProperties(Properties props, OutputStream outStream)
 			throws IOException {
 		ByteArrayOutputStream propsBuffer = new ByteArrayOutputStream();
@@ -492,14 +508,20 @@ public class IPC {
 	}
 
 	/**
-	 * @param inStream  Source for the properties
-	 * @param requireNull Indicates if the input message must be null-terminated (typically for socket input) or end of file (typically for re-parsing an already-received message) 
+	 * @param inStream    Source for the properties
+	 * @param requireNull Indicates if the input message must be null-terminated
+	 *                    (typically for socket input) or end of file (typically for
+	 *                    re-parsing an already-received message)
 	 * @return properties object
-	 * @throws IOException
+	 * @throws IOException on communication error
 	 */
 	public static Properties receiveProperties(InputStream inStream, boolean requireNull)
 			throws IOException {
 		byte msgBuff[] = AttachmentConnection.streamReceiveBytes(inStream, 0, requireNull);
+		if (IPC.isLoggingEnabled()) {
+			String propsString = new String(msgBuff, StandardCharsets.UTF_8);
+			IPC.logMessage("Received properties file:", propsString); //$NON-NLS-1$
+		}
 		Properties props = new Properties();
 		props.load(new ByteArrayInputStream(msgBuff));
 		return props;
--- a/jcl/src/java.base/share/classes/openj9/tools/attach/diagnostics/base/DiagnosticProperties.java
+++ b/jcl/src/java.base/share/classes/openj9/tools/attach/diagnostics/base/DiagnosticProperties.java
@@ -1,0 +1,277 @@
+/*[INCLUDE-IF Sidecar18-SE]*/
+/*******************************************************************************
+ * Copyright (c) 2019, 2019 IBM Corp. and others
+ *
+ * This program and the accompanying materials are made available under
+ * the terms of the Eclipse Public License 2.0 which accompanies this
+ * distribution and is available at https://www.eclipse.org/legal/epl-2.0/
+ * or the Apache License, Version 2.0 which accompanies this distribution and
+ * is available at https://www.apache.org/licenses/LICENSE-2.0.
+ *
+ * This Source Code may also be made available under the following
+ * Secondary Licenses when the conditions for such availability set
+ * forth in the Eclipse Public License, v. 2.0 are satisfied: GNU
+ * General Public License, version 2 with the GNU Classpath
+ * Exception [1] and GNU General Public License, version 2 with the
+ * OpenJDK Assembly Exception [2].
+ *
+ * [1] https://www.gnu.org/software/classpath/license.html
+ * [2] http://openjdk.java.net/legal/assembly-exception.html
+ *
+ * SPDX-License-Identifier: EPL-2.0 OR Apache-2.0 OR GPL-2.0 WITH Classpath-exception-2.0 OR LicenseRef-GPL-2.0 WITH Assembly-exception
+ *******************************************************************************/
+
+package openj9.tools.attach.diagnostics.base;
+
+import java.io.IOException;
+import java.io.PrintWriter;
+import java.io.StringWriter;
+import java.util.Comparator;
+import java.util.Properties;
+
+/**
+ * Augments Properties with convenience methods to add ints, booleans, and
+ * longs.
+ *
+ */
+public class DiagnosticProperties {
+	private final Properties baseProperties;
+
+	private static final String JAVA_LANG_STRING = "java.lang.String"; //$NON-NLS-1$
+
+	/**
+	 * Set this to "true" to enable verbose mode
+	 */
+	public static final String DEBUG_PROPERTY = "openj9.tools.attach.diagnostics.debug"; //$NON-NLS-1$
+
+	/**
+	 * For development use only
+	 */
+	public static boolean isDebug = Boolean.getBoolean(DEBUG_PROPERTY);
+
+	/**
+	 * @param props Properties object received from the target.
+	 */
+	public DiagnosticProperties(Properties props) {
+		baseProperties = props;
+	}
+
+	/**
+	 * Create an empty list of properties.
+	 */
+	public DiagnosticProperties() {
+		baseProperties = new Properties();
+	}
+
+	/**
+	 * Add a property with an integer value.
+	 * 
+	 * @param key   property name
+	 * @param value property value
+	 */
+	public void put(String key, int value) {
+		baseProperties.setProperty(key, Integer.toString(value));
+	}
+
+	/**
+	 * Add a property with a String value.
+	 * 
+	 * @param key   property name
+	 * @param value property value
+	 */
+	public void put(String key, String value) {
+		baseProperties.setProperty(key, value);
+	}
+
+	/**
+	 * Add a property with an long value.
+	 * 
+	 * @param key   property name
+	 * @param value property value
+	 */
+	public void put(String key, long value) {
+		baseProperties.setProperty(key, Long.toString(value));
+	}
+
+	/**
+	 * Add a property with a boolean value.
+	 * 
+	 * @param key   property name
+	 * @param value property value
+	 */
+	public void put(String key, boolean value) {
+		baseProperties.setProperty(key, Boolean.toString(value));
+	}
+
+	private void checkExists(String key) throws IOException {
+		if (!containsField(key)) {
+			throw new IOException("key " + key + " not found"); //$NON-NLS-1$ //$NON-NLS-2$
+		}
+	}
+
+	/**
+	 * Test if the given property is present.
+	 * 
+	 * @param key property name
+	 * @return true if the property name is found
+	 */
+	public boolean containsField(String key) {
+		return baseProperties.containsKey(key);
+	}
+
+	/**
+	 * Retrieve the value of a property and convert it to an int.
+	 * 
+	 * @param key name of the property
+	 * @return value of the property as an int
+	 * @throws NumberFormatException if the value is not a string representing a
+	 *                               number.
+	 * @throws IOException           if the property is missing
+	 */
+	public int getInt(String key) throws NumberFormatException, IOException {
+		checkExists(key);
+		return Integer.parseInt(baseProperties.getProperty(key));
+	}
+
+	/**
+	 * Retrieve the value of a property and convert it to an long.
+	 * 
+	 * @param key name of the property
+	 * @return value of the property as a long
+	 * @throws NumberFormatException if the value is not a string representing a
+	 *                               number.
+	 * @throws IOException           if the property is missing
+	 */
+	public long getLong(String key) throws NumberFormatException, IOException {
+		checkExists(key);
+		return Long.parseLong(baseProperties.getProperty(key));
+	}
+
+	/**
+	 * Retrieve the value of a property and convert it to a boolean.
+	 * 
+	 * @param key name of the property
+	 * @return value of the property as a boolean
+	 * @throws IOException if the property is missing
+	 */
+	public boolean getBoolean(String key) throws IOException {
+		checkExists(key);
+		return Boolean.parseBoolean(baseProperties.getProperty(key));
+	}
+
+	/**
+	 * Return a property value for the given key.
+	 * 
+	 * @param key property name
+	 * @return property value or null if the property is not found
+	 */
+	public String getPropertyOrNull(String key) {
+		return baseProperties.getProperty(key);
+	}
+
+	/**
+	 * Retrieve a value of a simple type, i.e. boxed scalar type, from a specified
+	 * property. Return null if the type is not recognized, the value is absent, or the
+	 * value is the empty string (unless the type is String).
+	 * 
+	 * @param typeName Fully qualified name of the type
+	 * @param key      Name of the property containing the value
+	 * @return object containing the value or null on error.
+	 * @throws NumberFormatException if the property is malformed
+	 */
+	public Object getSimple(String typeName, String key) throws NumberFormatException {
+		Object value = null;
+		String valueString = baseProperties.getProperty(key);
+		if ((null != valueString) && 
+				(JAVA_LANG_STRING.equals(typeName) || !valueString.isEmpty())) {
+			switch (typeName) {
+			case "java.lang.Boolean": //$NON-NLS-1$
+				value = Boolean.valueOf(valueString);
+				break;
+			case "java.lang.Character": //$NON-NLS-1$
+				value = Character.valueOf(valueString.charAt(0));
+				break;
+			case "java.lang.Byte": //$NON-NLS-1$
+				value = Byte.valueOf(valueString);
+				break;
+			case "java.lang.Short": //$NON-NLS-1$
+				value = Short.valueOf(valueString);
+				break;
+			case "java.lang.Integer": //$NON-NLS-1$
+				value = Integer.valueOf(valueString);
+				break;
+			case "java.lang.Long": //$NON-NLS-1$
+				value = Long.valueOf(valueString);
+				break;
+			case "java.lang.Float": //$NON-NLS-1$
+				value = Float.valueOf(valueString);
+				break;
+			case "java.lang.Double": //$NON-NLS-1$
+				value = Double.valueOf(valueString);
+				break;
+			case JAVA_LANG_STRING:
+				value = valueString;
+				break;
+			default:
+				break;
+			}
+		}
+		return value;
+	}
+
+	/**
+	 * Print a DiagnosticProperties object.
+	 * 
+	 * @param msg   Message to print before properties. May be null.
+	 * @param props DiagnosticProperties object to dump
+	 */
+	public static void dumpPropertiesIfDebug(String msg, DiagnosticProperties props) {
+		dumpPropertiesIfDebug(msg, props.baseProperties);
+	}
+
+	/**
+	 * Print a list of properties in sorted order if debugging is enabled.
+	 * 
+	 * @param msg   Message to print before properties. May be null.
+	 * @param props Properties object to dump
+	 */
+	public static void dumpPropertiesIfDebug(String msg, final Properties props) {
+		if (DiagnosticProperties.isDebug) {
+			if (null != msg) {
+				System.err.println(msg);
+			}
+			if (null != props) {
+				final String result = printProperties(props);
+				System.err.print(result);
+			}
+		}
+	}
+
+	/**
+	 * Print a Properties object in sorted key order.
+	 * 
+	 * @param props object to dump
+	 * @return String representation of props
+	 */
+	public static String printProperties(final Properties props) {
+		final StringWriter buff = new StringWriter(1000);
+		try (PrintWriter buffWriter = new PrintWriter(buff)) {
+			props.entrySet().stream().sorted(Comparator.comparing(e -> e.getKey().toString())).forEach(theEntry -> {
+				buffWriter.print(theEntry.getKey());
+				buffWriter.print("="); //$NON-NLS-1$
+				buffWriter.println((String) theEntry.getValue());
+			});
+		}
+		final String result = buff.toString();
+		return result;
+	}
+
+	/**
+	 * Return the underlying properties object by reference.
+	 * 
+	 * @return Properties object
+	 */
+	public Properties toProperties() {
+		return baseProperties;
+	}
+}
--- a/jcl/src/java.management/share/classes/java/lang/management/ThreadInfo.java
+++ b/jcl/src/java.management/share/classes/java/lang/management/ThreadInfo.java
@@ -771,8 +771,8 @@ public class ThreadInfo {
 				result.append(String.format(" on %s owned by \"%s\" Id=%d",  //$NON-NLS-1$
 						lockName, lockOwnerName, Long.valueOf(lockOwnerId)));
 			}
+			result.append(ls);
 			if (stackTraces != null && stackTraces.length > 0) {
-				result.append(ls);
 				MonitorInfo[] lockList = getLockedMonitors();
 				MonitorInfo[] lockArray = new MonitorInfo[stackTraces.length];
 				for (MonitorInfo mi: lockList) {
@@ -792,9 +792,6 @@ public class ThreadInfo {
 					}
 					++stackDepth;
 				}
-			} else {
-				result.append(" null"); //$NON-NLS-1$
-				result.append(ls);
 			}
 			TOSTRING_VALUE = result.toString();
 		}
--- a/jcl/src/jdk.attach/share/classes/module-info.java.extra
+++ b/jcl/src/jdk.attach/share/classes/module-info.java.extra
@@ -24,3 +24,4 @@
 /*[REM] This file must not use tabs because the dependency recognition code in openjdk does not support them. */
 
 provides com.sun.tools.attach.spi.AttachProvider with com.ibm.tools.attach.attacher.OpenJ9AttachProvider;
+exports com.ibm.tools.attach.attacher to jdk.jcmd;
--- a/jcl/src/jdk.jcmd/share/classes/module-info.java
+++ b/jcl/src/jdk.jcmd/share/classes/module-info.java
@@ -1,6 +1,6 @@
 /*[INCLUDE-IF Sidecar19-SE]*/
 /*******************************************************************************
- * Copyright (c) 2016, 2019 IBM Corp. and others
+ * Copyright (c) 2019, 2019 IBM Corp. and others
  *
  * This program and the accompanying materials are made available under
  * the terms of the Eclipse Public License 2.0 which accompanies this
@@ -23,5 +23,10 @@
 
 /*[REM] This file must not use tabs because the dependency recognition code in openjdk does not support them. */
 
-exports com.ibm.java.lang.management.internal to jdk.jcmd, jdk.management;
-uses com.ibm.sharedclasses.spi.SharedClassProvider;
+module jdk.jcmd {
+  exports openj9.tools.attach.diagnostics.info;
+  provides openj9.tools.attach.diagnostics.spi.TargetDiagnosticsProvider with openj9.tools.attach.diagnostics.target.TargetDiagnosticsProviderImpl;
+  requires java.base;
+  requires java.management;
+  requires jdk.attach;
+}
--- a/jcl/src/jdk.jcmd/share/classes/openj9/tools/attach/diagnostics/attacher/AttacherDiagnosticsProvider.java
+++ b/jcl/src/jdk.jcmd/share/classes/openj9/tools/attach/diagnostics/attacher/AttacherDiagnosticsProvider.java
@@ -1,0 +1,127 @@
+/*[INCLUDE-IF Sidecar18-SE]*/
+/*******************************************************************************
+ * Copyright (c) 2019, 2019 IBM Corp. and others
+ *
+ * This program and the accompanying materials are made available under
+ * the terms of the Eclipse Public License 2.0 which accompanies this
+ * distribution and is available at https://www.eclipse.org/legal/epl-2.0/
+ * or the Apache License, Version 2.0 which accompanies this distribution and
+ * is available at https://www.apache.org/licenses/LICENSE-2.0.
+ *
+ * This Source Code may also be made available under the following
+ * Secondary Licenses when the conditions for such availability set
+ * forth in the Eclipse Public License, v. 2.0 are satisfied: GNU
+ * General Public License, version 2 with the GNU Classpath
+ * Exception [1] and GNU General Public License, version 2 with the
+ * OpenJDK Assembly Exception [2].
+ *
+ * [1] https://www.gnu.org/software/classpath/license.html
+ * [2] http://openjdk.java.net/legal/assembly-exception.html
+ *
+ * SPDX-License-Identifier: EPL-2.0 OR Apache-2.0 OR GPL-2.0 WITH Classpath-exception-2.0 OR LicenseRef-GPL-2.0 WITH Assembly-exception
+ *******************************************************************************/
+
+package openj9.tools.attach.diagnostics.attacher;
+
+import static com.ibm.oti.util.Msg.getString;
+
+import java.io.IOException;
+import java.util.Properties;
+
+import com.ibm.tools.attach.attacher.OpenJ9AttachProvider;
+import com.ibm.tools.attach.attacher.OpenJ9VirtualMachine;
+import com.ibm.tools.attach.target.IPC;
+import com.sun.tools.attach.AttachNotSupportedException;
+
+import openj9.tools.attach.diagnostics.base.DiagnosticProperties;
+import openj9.tools.attach.diagnostics.base.DiagnosticsInfo;
+import openj9.tools.attach.diagnostics.info.ThreadGroupInfo;
+
+/**
+ * This class allows a Attach API attacher to query a target JVM about
+ * diagnostic information such as JIT compilation, threads, classes, etc.
+ * Instances must be created using the getDiagnostics() factory method.
+ *
+ */
+public class AttacherDiagnosticsProvider {
+
+	private OpenJ9VirtualMachine vm;
+
+	/**
+	 * Acquire the stacks of all running threads in the current VM and encode them
+	 * into a set of properties.
+	 * 
+	 * @param printSynchronizers indicate if ownable synchronizers should be printed
+	 * @return properties object
+	 * @throws IOException on communication error
+	 */
+	public DiagnosticsInfo getThreadGroupInfo(boolean printSynchronizers) throws IOException {
+		IPC.logMessage("enter getRemoteThreadGroupInfo"); //$NON-NLS-1$
+		checkAttached();
+		Properties threadInfo = vm.getThreadInfo();
+		DiagnosticProperties.dumpPropertiesIfDebug("Properties from target:", threadInfo); //$NON-NLS-1$
+		ThreadGroupInfo info = new ThreadGroupInfo(threadInfo, printSynchronizers);
+		IPC.logMessage("exit getRemoteThreadGroupInfo"); //$NON-NLS-1$
+		return info;
+	}
+
+	/**
+	 * Call equivalent com.sun.tools.attach.VirtualMachine method.
+	 * 
+	 * @param vmid ID of target
+	 * @throws IOException on communication error
+	 */
+	public void attach(String vmid) throws IOException {
+		OpenJ9AttachProvider attachProv = new OpenJ9AttachProvider();
+		IPC.logMessage("DiagnosticsProviderImpl attaching to ", vmid); //$NON-NLS-1$
+		try {
+			vm = attachProv.attachVirtualMachine(vmid);
+		} catch (AttachNotSupportedException e) {
+			/*[MSG "K0809", "Exception connecting to {0}"] */
+			throw new IOException(getString("K0809", vmid)); //$NON-NLS-1$
+		}
+		IPC.logMessage("DiagnosticsProviderImpl attached to ", vmid); //$NON-NLS-1$
+	}
+
+	/**
+	 * Call equivalent com.sun.tools.attach.VirtualMachine method.
+	 * 
+	 * @throws IOException on communication error
+	 */
+	public void detach() throws IOException {
+		if (null != vm) {
+			vm.detach();
+			vm = null;
+		}
+	}
+
+	/**
+	 * Call equivalent com.sun.tools.attach.VirtualMachine method.
+	 * 
+	 * @return properties from target VM
+	 * @throws IOException on communication error
+	 */
+	public Properties getSystemProperties() throws IOException {
+		checkAttached();
+		return vm.getSystemProperties();
+	}
+
+	/**
+	 * Call equivalent com.sun.tools.attach.VirtualMachine method.
+	 * 
+	 * @return properties from target VM
+	 * @throws IOException on communication error
+	 */
+	public Properties getAgentProperties() throws IOException {
+		checkAttached();
+		return vm.getAgentProperties();
+	}
+
+	private void checkAttached() throws IOException {
+		if (null == vm) {
+			/*[MSG "K0544", "Target not attached"] */
+			throw new IOException(getString("K0554")); //$NON-NLS-1$
+		}
+	}
+
+}
--- a/jcl/src/jdk.jcmd/share/classes/openj9/tools/attach/diagnostics/info/JvmInfo.java
+++ b/jcl/src/jdk.jcmd/share/classes/openj9/tools/attach/diagnostics/info/JvmInfo.java
@@ -1,0 +1,169 @@
+/*[INCLUDE-IF Sidecar18-SE]*/
+/*******************************************************************************
+ * Copyright (c) 2019, 2019 IBM Corp. and others
+ *
+ * This program and the accompanying materials are made available under
+ * the terms of the Eclipse Public License 2.0 which accompanies this
+ * distribution and is available at https://www.eclipse.org/legal/epl-2.0/
+ * or the Apache License, Version 2.0 which accompanies this distribution and
+ * is available at https://www.apache.org/licenses/LICENSE-2.0.
+ *
+ * This Source Code may also be made available under the following
+ * Secondary Licenses when the conditions for such availability set
+ * forth in the Eclipse Public License, v. 2.0 are satisfied: GNU
+ * General Public License, version 2 with the GNU Classpath
+ * Exception [1] and GNU General Public License, version 2 with the
+ * OpenJDK Assembly Exception [2].
+ *
+ * [1] https://www.gnu.org/software/classpath/license.html
+ * [2] http://openjdk.java.net/legal/assembly-exception.html
+ *
+ * SPDX-License-Identifier: EPL-2.0 OR Apache-2.0 OR GPL-2.0 WITH Classpath-exception-2.0 OR LicenseRef-GPL-2.0 WITH Assembly-exception
+ *******************************************************************************/
+
+package openj9.tools.attach.diagnostics.info;
+
+import java.io.IOException;
+import java.util.Set;
+
+import javax.management.openmbean.ArrayType;
+import javax.management.openmbean.CompositeData;
+import javax.management.openmbean.CompositeDataSupport;
+import javax.management.openmbean.CompositeType;
+import javax.management.openmbean.OpenDataException;
+import javax.management.openmbean.OpenType;
+import javax.management.openmbean.SimpleType;
+
+import com.ibm.java.lang.management.internal.LockInfoUtil;
+import com.ibm.java.lang.management.internal.MonitorInfoUtil;
+import com.ibm.java.lang.management.internal.StackTraceElementUtil;
+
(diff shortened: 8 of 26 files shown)


pshipton:
I don't think we want the new modules openj9.attach and openj9.diagnostics. The code should be added to the existing modules jdk.attach and jdk.jcmd

test/functional/Java8andUp/src/org/openj9/test/attachAPI/TestJstack.java
```diff
@@ -0,0 +1,237 @@
...
+	/* basic sanity: fetch and print the thread information */
+	public void testAllThreads() throws IOException {
+		Properties props = ThreadGroupInfo.getThreadInfoProperties().toProperties();
+		logProperties(props);
+		ThreadGroupInfo grpInfo = new ThreadGroupInfo(props);
+		log(grpInfo.toString());
+	}
+
+	@Test(groups = { "disable.generic-all" })
+	// TODO re-enable this when https://github.com/eclipse/openj9/issues/5012 is resolved
```
fjeremic: We should make note of this in #5012 so we don't forget.

test/functional/Java8andUp/src/org/openj9/test/attachAPI/TestJstack.java
```diff
@@ -0,0 +1,237 @@
...
+	}
+
+	void method1() {
+		method2();
+	}
+
+	void method2() {
+		synchronized (syncObject) {
+			try {
+				log("method2: Wait"); // TODO convert this to log //$NON-NLS-1$
```
fjeremic: What do these TODOs mean?

pdbain-ibm:
> I don't think we want the new modules openj9.attach and openj9.diagnostics. The code should be added to the existing modules jdk.attach and jdk.jcmd

The diagnostics tools require access to the `java.management` and `jdk.management` modules.  As discussed, I will overwrite the OpenJDK `jdk.jcmd` and add the necessary `require`s.

test/functional/Java8andUp/src/org/openj9/test/attachAPI/TestJstack.java
pdbain-ibm: Legacy code.  Fixed.

pshipton:
This seems unlikely to be delivered before the 0.14 code split Monday. However we will make an assessment of the change when it is ready to determine if it can go into 0.14 after the code split.

pdbain-ibm:
I have designer tested it and am running personal builds.  I'm still polishing the code and hope to have it ready for review today.

pdbain-ibm:
Tested and working in personal builds.  Ready for review.

pdbain-ibm:
Sample output:
```
"testMonitorsThread" Id=20 BLOCKED on java.lang.Object@2f0a8076 owned by "main" Id=1 
    at org.openj9.test.attachAPI.TestJstack.syncMethod(TestJstack.java:224)
    at org.openj9.test.attachAPI.TestJstack$2.run(TestJstack.java:100)
    at java.lang.Thread.run(Thread.java:813)
    CPU time: 5804078 ns
    User  time: 5804078 ns

"main" Id=1 RUNNABLE
    at com.ibm.java.lang.management.internal.ThreadMXBeanImpl.dumpAllThreadsImpl(Native Method)
    at com.ibm.java.lang.management.internal.ThreadMXBeanImpl.dumpAllThreadsCommon(ThreadMXBeanImpl.java:643)
    at com.ibm.java.lang.management.internal.ThreadMXBeanImpl.dumpAllThreads(ThreadMXBeanImpl.java:626)
    at openj9.tools.attach.diagnostics.info.ThreadGroupInfo.getThreadInfoProperties(ThreadGroupInfo.java:96)
    at org.openj9.test.attachAPI.TestJstack.testMonitors(TestJstack.java:111)
    - locked java.lang.Object@2f0a8076
    at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)

"GC Slave" Id=15 RUNNABLE null
    CPU time: 12376169 ns
    User  time: 12376169 ns

"process reaper" Id=22 RUNNABLE
    at java.lang.UNIXProcess.waitForProcessExit(Native Method)
    at java.lang.UNIXProcess.lambda$initStreams$3(UNIXProcess.java:289)
    at java.lang.UNIXProcess$$Lambda$16.000000007C386F80.run(Unknown Source)
    at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1149)
    at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:624)
    at java.lang.Thread.run(Thread.java:813)
Ownable synchronizers
- java.util.concurrent.ThreadPoolExecutor$Worker@652df21e
    CPU time: 2784317 ns
    User  time: 2784317 ns
```

jcl/src/jdk.jcmd/share/classes/openj9/tools/attach/diagnostics/info/ThreadGroupInfo.java
```diff
@@ -0,0 +1,207 @@
...
+ * This mimics the behaviour of Thread
+ */
+public class ThreadGroupInfo implements DiagnosticsInfo {
+
+	private static final String TIME_FORMAT_STRING = "\t%s time: %d ns%n";
+	static final String THREAD_GROUP_PREFIX = "threadgroup."; //$NON-NLS-1$
+	static final String CPU_TIME = "cpu_time"; //$NON-NLS-1$
+	static final String USER_TIME = "user_time"; //$NON-NLS-1$
+	private static final String THREAD_GROUP_VERSION = THREAD_GROUP_PREFIX + "version"; //$NON-NLS-1$
+	private static final String THREAD_GROUP_JAVA_INFO = THREAD_GROUP_PREFIX + "java_info"; //$NON-NLS-1$
```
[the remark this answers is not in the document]
pdbain-ibm: Good plan.

jcl/src/java.management/share/classes/java/lang/management/ThreadInfo.java
```diff
@@ -610,7 +610,7 @@ public static ThreadInfo from(CompositeData cd) {
 				daemonVal = ((Boolean) cd.get("daemon")).booleanValue(); //$NON-NLS-1$
 				priorityVal = ((Integer) cd.get("priority")).intValue(); //$NON-NLS-1$
 				/*[ENDIF]*/
-			} catch (InvalidKeyException e) {
+			} catch (NullPointerException | InvalidKeyException e) {
```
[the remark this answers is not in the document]
pdbain-ibm: This handles composite data from incompatible foreign JVMs more gracefully.  Separate pullreq opened: https://github.com/eclipse/openj9/pull/5226.

jcl/src/jdk.jcmd/share/classes/openj9/tools/attach/diagnostics/attacher/AttacherDiagnosticsProvider.java
```diff
@@ -0,0 +1,160 @@
...
+		} else {
+			osBean = null;
+		}
+	}
+	
+	/**
+	 * @return Attach API virtual machine ID of the current process
+	 */
+	public String getMyVmId() {
+		return (null != runtimeBean) ? runtimeBean.getVmId() : null;
```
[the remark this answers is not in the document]
pdbain-ibm: `jdk.management` is still needed by `TargetDiagnosticsProviderImpl`. You are however correct that `getMyVmId()` and `isProcessRunning()` are redundant.

pdbain-ibm:
Thanks @pshipton !
Updated and tested on my computer.  Running a personal build.

jcl/src/jdk.jcmd/share/classes/openj9/tools/attach/diagnostics/attacher/AttacherDiagnosticsProvider.java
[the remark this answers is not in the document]
pshipton: Are you going to fix them? I don't see why TargetDiagnosticsProviderImpl needs jdk.management otherwise.

jcl/src/java.base/share/classes/com/ibm/tools/attach/target/Attachment.java
```diff
@@ -60,6 +66,7 @@
 	private String attachError;
 	private final AttachHandler handler;
 	private final String key;
+	private TargetDiagnosticsProvider diagProvider;
```
pshipton: This is specific to Java 11

jcl/src/java.base/share/classes/com/ibm/tools/attach/target/Attachment.java
```diff
@@ -106,7 +113,7 @@
 		portNumber = rc.getPortNumber();
 		this.key = rc.getKey();
 		this.handler = attachHandler;
-
+		diagProvider = null;
```
pshipton: This is specific to Java 11

jcl/src/jdk.jcmd/share/classes/openj9/tools/attach/diagnostics/info/JvmThreadInfo.java
```diff
@@ -0,0 +1,98 @@
...
+
+import openj9.tools.attach.diagnostics.base.DiagnosticProperties;
+
+/**
+ * Container for information on a foreign JVM's thread
+ *
+ */
+public class JvmThreadInfo  extends JvmInfo {
+
+	private static final String THREAD_INFO = THREAD_GROUP_PREFIX + "thread_info."; //$NON-NLS-1$
```
pshipton: There is only the one use of "thread_info." (for THREAD_NUM), is it really needed?

jcl/src/jdk.jcmd/share/classes/openj9/tools/attach/diagnostics/info/ThreadGroupInfo.java
```diff
@@ -0,0 +1,209 @@
...
+ * This mimics the behaviour of Thread
+ */
+public class ThreadGroupInfo implements DiagnosticsInfo {
+
+	private static final String TIME_FORMAT_STRING = "\t%s time: %d ns%n";
+	static final String THREAD_GROUP_PREFIX = "threadgroup."; //$NON-NLS-1$
+	static final String CPU_TIME = "cpu_time"; //$NON-NLS-1$
+	static final String USER_TIME = "user_time"; //$NON-NLS-1$
+	private static final String THREAD_GROUP_VERSION = THREAD_GROUP_PREFIX + "version"; //$NON-NLS-1$
+	private static final String THREAD_GROUP_JAVA_INFO = THREAD_GROUP_PREFIX + DiagnosticsInfo.JAVA_INFO; //$NON-NLS-1$
```
pshipton: Does this really need the THREAD_GROUP_PREFIX? `System.getProperty("java.vm.info")` will be the same for every command,

jcl/src/jdk.jcmd/share/classes/openj9/tools/attach/diagnostics/info/ThreadGroupInfo.java
```diff
@@ -0,0 +1,207 @@
...
+		return props;
+	}
+
+	/**
+	 * Acquire stack traces for selected threads
+	 * @param threadIds list of threads to query
+	 * @return Properties objects thread information
+	 * @note some entries my be null if a specified thread no longer exists
+	 */
+	public static DiagnosticProperties getThreadInfoProperties(long[] threadIds) {
```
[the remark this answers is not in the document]
pshipton: There was no reply to this comment.

jcl/src/jdk.jcmd/share/classes/openj9/tools/attach/diagnostics/info/ThreadGroupInfo.java
```diff
@@ -0,0 +1,209 @@
...
+			if (myProps.containsField(IPC.PROPERTY_DIAGNOSTICS_ERROR) && myProps.getBoolean(IPC.PROPERTY_DIAGNOSTICS_ERROR)) {
+				final String targetMsg = myProps.getProperty(IPC.PROPERTY_DIAGNOSTICS_ERRORMSG);
+				String msg = myProps.getProperty(IPC.PROPERTY_DIAGNOSTICS_ERRORTYPE); 
+				if ((null != targetMsg) && !targetMsg.isEmpty()) {
+					msg = msg+ ": " +  targetMsg; //$NON-NLS-1$
+				}
+				throw new IOException(msg);
+			}
+			numThreads = myProps.getInt(NUM_THREADS);
+			groupInfoVersion = props.getProperty(THREAD_GROUP_VERSION);
```
pshipton: Seems there should be a check of the version, with an appropriate action.

jcl/src/jdk.management/share/classes/module-info.java.extra
```diff
@@ -1,6 +1,6 @@
 /*[INCLUDE-IF Sidecar19-SE]*/
 /*******************************************************************************
- * Copyright (c) 2016, 2018 IBM Corp. and others
+ * Copyright (c) 2016, 2019 IBM Corp. and others
```
pshipton: The copyright change in the only change to this file.

jcl/src/java.management/share/classes/module-info.java.extra
```diff
@@ -20,5 +20,5 @@
  *
  * SPDX-License-Identifier: EPL-2.0 OR Apache-2.0 OR GPL-2.0 WITH Classpath-exception-2.0 OR LicenseRef-GPL-2.0 WITH Assembly-exception
  *******************************************************************************/
-	exports com.ibm.java.lang.management.internal to jdk.management;
+	exports com.ibm.java.lang.management.internal to jdk.management, jdk.jcmd;
```
pshipton: The other changes are done alphabetically, this one doesn't match.

jcl/src/jdk.jcmd/share/classes/openj9/tools/attach/diagnostics/info/JvmThreadInfo.java
pdbain-ibm: Not really.  Deleted it and THREAD_NUM.

jcl/src/jdk.jcmd/share/classes/openj9/tools/attach/diagnostics/info/ThreadGroupInfo.java
pdbain-ibm: I will change it to a generic prefix.

jcl/src/jdk.jcmd/share/classes/openj9/tools/attach/diagnostics/info/ThreadGroupInfo.java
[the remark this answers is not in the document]
pdbain-ibm: I would like to leave it as-is because:
- it simplifies a unit test which works on a known thread by not having to filter out the other threads
- It allows other utilities which would work on s specific thread or set of threads
- The extra no-argument variant of this method method is a wrapper and there is no duplicated code.

jcl/src/jdk.jcmd/share/classes/openj9/tools/attach/diagnostics/attacher/AttacherDiagnosticsProvider.java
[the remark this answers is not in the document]
pdbain-ibm: Good plan.  Will do.

jcl/src/jdk.jcmd/share/classes/openj9/tools/attach/diagnostics/info/ThreadGroupInfo.java
pdbain-ibm: Adding a check.

jcl/src/jdk.jcmd/share/classes/openj9/tools/attach/diagnostics/info/ThreadGroupInfo.java
```diff
@@ -0,0 +1,220 @@
...
+public class ThreadGroupInfo implements DiagnosticsInfo {
+
+	private static final String CURRENT_GROUPINFO_VERSION = "0"; //$NON-NLS-1$
+	private static final String TIME_FORMAT_STRING = "\t%s time: %d ns%n"; //$NON-NLS-1$
+	static final String THREAD_GROUP_PREFIX = "threadgroup."; //$NON-NLS-1$
+	static final String CPU_TIME = "cpu_time"; //$NON-NLS-1$
+	static final String USER_TIME = "user_time"; //$NON-NLS-1$
+	private static final String THREAD_GROUP_VERSION = THREAD_GROUP_PREFIX + "version"; //$NON-NLS-1$
+	private static final String NUM_THREADS = THREAD_GROUP_PREFIX+"num_threads"; //$NON-NLS-1$
+	private static ThreadMXBean threadBean;
```
keithc-ca: This is not thread-safe: it should be passed to static methods that need it.

jcl/src/jdk.jcmd/share/classes/openj9/tools/attach/diagnostics/info/ThreadGroupInfo.java
```diff
@@ -0,0 +1,220 @@
...
+				}
+			} catch (NumberFormatException | IOException e) {
+				/* EMPTY: ignore missing optional information */
+			}
+			buff.append(newline);
+		}
+		String result = buff.toString();
+		if (result.isEmpty()) {
+			result = "Incompatible data from target VM:\n"  //$NON-NLS-1$
+		+ DiagnosticProperties.printProperties(myProps.toProperties());
```
keithc-ca: nit: indentation

jcl/src/jdk.jcmd/share/classes/openj9/tools/attach/diagnostics/tools/Jps.java
```diff
@@ -21,7 +21,7 @@
  * SPDX-License-Identifier: EPL-2.0 OR Apache-2.0 OR GPL-2.0 WITH Classpath-exception-2.0 OR LicenseRef-GPL-2.0 WITH Assembly-exception
  *******************************************************************************/
 
-package openj9.tools.attach.diagnostics;
+package openj9.tools.attach.diagnostics.tools;
```
keithc-ca: This will require the launcher scripts to be updated.

jcl/src/jdk.jcmd/share/classes/openj9/tools/attach/diagnostics/tools/Jstack.java
```diff
@@ -0,0 +1,156 @@
...
+				}
+			} finally {
+				try {
+					diagProvider.detach();
+				} catch (IOException e) {
+					/* ignore */
+				}
+			}
+		}
+		System.exit(rc);
```
keithc-ca: This doesn't appear to be useful: I don't see any changes after `int rc = 0;`.

jcl/src/jdk.jcmd/share/classes/openj9/tools/attach/diagnostics/tools/Jstack.java
```diff
@@ -0,0 +1,156 @@
...
+			} else if (a.equals("-l")) {
+				printSynchronizers = true;
+			} else {
+				err.printf(HELPTEXT, "");
+				okay = false;
+				break;
+			}
+		}
+
+		if (okay && (0 == vmids.size())) {
```
keithc-ca: Why not `vmids.isEmpty()`?

test/functional/Java8andUp/src/org/openj9/test/attachAPI/TestJstack.java
```diff
@@ -0,0 +1,236 @@
...
+@Test(groups = { "level.extended" })
+public class TestJstack extends AttachApiTest {
+
+	private static final String JSTACK_COMMAND = "jstack"; //$NON-NLS-1$
+	Object syncObject = new Object();
+
+	@Test
+	public void testSingleThread() throws IOException {
+		try {
+			Thread testThread = new Thread(new Runnable() {
```
keithc-ca: This could use a method handle: `this::method1`.

jcl/src/java.base/share/classes/com/ibm/tools/attach/target/Attachment.java
```diff
@@ -260,9 +277,39 @@ boolean doCommand(InputStream cmdStream, OutputStream respStream) {
...
+			props.put(IPC.PROPERTY_DIAGNOSTICS_ERRORMSG, msg);
+		}
+		return props;
+	}
+
+	private void replyWithProperties(DiagnosticProperties props) throws IOException {
+		replyWithProperties(props.toProperties());
+	}
+
 	private void replyWithProperties(Properties props) throws IOException {
```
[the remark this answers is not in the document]
pdbain-ibm: It makes reference to  the instance variable `responseStream` so I believe it must be a virtual method.

jcl/src/jdk.jcmd/share/classes/openj9/tools/attach/diagnostics/tools/Jps.java
pdbain-ibm: Thanks for pointing that out.  I have been updating and testing the launcher scripts.

jcl/src/jdk.jcmd/share/classes/openj9/tools/attach/diagnostics/tools/Jstack.java
pdbain-ibm: True.  Removed it and the variable.

test/functional/Java8andUp/src/org/openj9/test/attachAPI/TestJstack.java
pdbain-ibm: Cool.  Thanks.

jcl/src/java.base/share/classes/openj9/tools/attach/diagnostics/base/DiagnosticProperties.java
```diff
@@ -0,0 +1,250 @@
...
+
+	/**
+	 * Print a Properties object in sorted key order
+	 * @param props object to dump
+	 * @return String representation of props
+	 */
+	public static String printProperties(final Properties props) {
+		StringBuffer buff = new StringBuffer(1000);
+		String newLine = System.lineSeparator();
+		props.keySet().stream().sorted().forEach(k -> {
```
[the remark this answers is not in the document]
pdbain-ibm: Thanks.

jcl/src/jdk.jcmd/share/classes/openj9/tools/attach/diagnostics/info/ThreadGroupInfo.java
pdbain-ibm: Good plan.

pdbain-ibm:
Updated per Keith's suggestions, not yet tested.

keithc-ca:
In your example, what is `null` in
> "GC Slave" Id=15 RUNNABLE null

pdbain:
> In your example, what is `null` in
> 
> > "GC Slave" Id=15 RUNNABLE null

That indicates that there is no Java stack.  It comes from the ThreadInfo `toString()`:
```
			if (stackTraces != null && stackTraces.length > 0) {
...
			} else {
				result.append(" null"); //$NON-NLS-1$
				result.append(ls);
			}
```

keithc-ca:
Is that what the reference implementation does when there's no stack? I think the output could be clearer, for example:
```
"GC Slave" Id=15 RUNNABLE
    [no stack]
    CPU time: 12376169 ns
    User  time: 12376169 ns
```

pdbain-ibm:
> Is that what the reference implementation does when there's no stack? I think the output could be clearer, for example:
> 
> ```
> "GC Slave" Id=15 RUNNABLE
>     [no stack]
>     CPU time: 12376169 ns
>     User  time: 12376169 ns
> ```

OpenJ9 (in Java 11) prints native threads thus:
`"JIT Compilation Thread-6 Suspended" prio=10 Id=9 RUNNABLE null`

the RI prints:
`"C1 CompilerThread3" #8 daemon prio=9 os_prio=31 tid=0x00007ff3b3800000 nid=0x3e03 waiting on condition [0x0000000000000000]`

Printing `[no Java stack]` in lieu of `null` sounds like a good idea. @pshipton your thoughts?

pshipton:
Printing nothing, similar to the RI, seems best.

pdbain-ibm:
Updated and designer tested. I added a JPP flag to fix compilation failures in the Java 8 special builds.

pdbain-ibm:
Did another pass and cleaned up some spacing issues.  @keithc-ca have I addressed all your comments?

pdbain-ibm:
Requires changes in extensions repo:

- [ ] https://github.com/ibmruntimes/openj9-openjdk-jdk8/pull/279
- [ ] https://github.com/ibmruntimes/openj9-openjdk-jdk11/pull/146
- [ ] https://github.com/ibmruntimes/openj9-openjdk-jdk12/pull/35
- [ ] https://github.com/ibmruntimes/openj9-openjdk-jdk/pull/90

jcl/src/java.base/share/classes/com/ibm/tools/attach/target/IPC.java
```diff
@@ -67,6 +68,17 @@
 	static final int LOGGING_UNKNOWN = 0;
 	static final int LOGGING_DISABLED = 1;
 	static final int LOGGING_ENABLED = 2;
+	/**
+	 * Prefixes and names for Diagnostics properties
+	 */
+	public static final String PROPERTY_DIAGNOSTICS_STATUS_PREFIX =
+			"org.openj9.diagnostics.status."; //$NON-NLS-1$
```
pshipton: Just to shorten things up, I think the "org" and the "status" parts can be removed. Seems this could use the OPENJ9_DIAGNOSTICS_PREFIX defined elsewhere.

jcl/src/java.management/share/classes/java/lang/management/ThreadInfo.java
```diff
@@ -793,7 +793,6 @@ public String toString() {
 					++stackDepth;
 				}
 			} else {
-				result.append(" null"); //$NON-NLS-1$
 				result.append(ls);
```
pshipton: You can move the `append(ls)` above the if case, then the else case isn't needed.

jcl/src/jdk.attach/share/classes/module-info.java.extra
```diff
@@ -21,3 +21,4 @@
  * SPDX-License-Identifier: EPL-2.0 OR Apache-2.0 OR GPL-2.0 WITH Classpath-exception-2.0 OR LicenseRef-GPL-2.0 WITH Assembly-exception
  *******************************************************************************/
 	provides com.sun.tools.attach.spi.AttachProvider with com.ibm.tools.attach.attacher.OpenJ9AttachProvider;
+	exports com.ibm.tools.attach.attacher to jdk.jcmd;
```
pshipton: See https://github.com/eclipse/openj9/pull/5260

jcl/src/jdk.jcmd/share/classes/openj9/tools/attach/diagnostics/info/JvmThreadInfo.java
```diff
@@ -0,0 +1,97 @@
...
+
+	static void put(ThreadInfo info, DiagnosticProperties props, int threadNum) {
+		CompositeData threadData = ThreadInfoUtil.toCompositeData(info);
+		String keyPrefix = thrdNumPrefix(threadNum);
+		final CompositeType compType = ThreadInfoUtil.getCompositeType();
+		addFields(props, keyPrefix, compType, threadData);
+	}
+
+	static String thrdNumPrefix(int threadNum) {
+		return THREAD_GROUP_PREFIX + "thread_info.thread_num."  //$NON-NLS-1$
```
pshipton: "thread_info." seems unnecessary.

jcl/src/jdk.jcmd/share/classes/openj9/tools/attach/diagnostics/target/TargetDiagnosticsProviderImpl.java
```diff
@@ -0,0 +1,55 @@
...
+ * This class allows a Attach API attacher to query a target JVM about diagnostic information
+ * such as JIT compilation, threads, classes, etc.
+ * Instances must be created using the getDiagnostics() factory method.
+ *
+ */
+public class TargetDiagnosticsProviderImpl implements TargetDiagnosticsProvider {
+
+	
+	@Override
+	public String getMyVmId() {
```
pshipton: I don't find any callers of this.

pshipton:
Note the merge conflicts, due to #5260

jcl/src/jdk.jcmd/share/classes/openj9/tools/attach/diagnostics/target/TargetDiagnosticsProviderImpl.java
pdbain-ibm: Removed.

pdbain-ibm:
Updated per @pshipton 's comments, merge conflicts in `module-info` files fixed.

jcl/src/jdk.jcmd/share/classes/module-info.java
```diff
@@ -0,0 +1,30 @@
...
+ * Exception [1] and GNU General Public License, version 2 with the
+ * OpenJDK Assembly Exception [2].
+ *
+ * [1] https://www.gnu.org/software/classpath/license.html
+ * [2] http://openjdk.java.net/legal/assembly-exception.html
+ *
+ * SPDX-License-Identifier: EPL-2.0 OR Apache-2.0 OR GPL-2.0 WITH Classpath-exception-2.0 OR LicenseRef-GPL-2.0 WITH Assembly-exception
+ *******************************************************************************/
+module jdk.jcmd {
+	exports openj9.tools.attach.diagnostics.info;
```
pshipton: Still some tabs in here, and please add the comment about no tabs.

jcl/src/java.base/share/classes/openj9/tools/attach/diagnostics/base/DiagnosticProperties.java
```diff
@@ -0,0 +1,260 @@
...
+	/**
+	 * Print a Properties object in sorted key order
+	 * @param props object to dump
+	 * @return String representation of props
+	 */
+	public static String printProperties(final Properties props) {
+		final StringWriter buff = new StringWriter(1000);
+		try (PrintWriter buffWriter = new PrintWriter(buff)) {
+			props.entrySet().stream().sorted((e1, e2) 
+					-> e1.getKey().toString().compareTo(e2.getKey().toString()))
```
keithc-ca: This can be simplified:
```
props.entrySet().stream().sorted(Comparator.comparing(e -> e.getKey().toString()))
```

jcl/src/java.base/share/classes/openj9/tools/attach/diagnostics/base/DiagnosticProperties.java
```diff
@@ -0,0 +1,260 @@
...
+			.forEach(k -> {
```
keithc-ca: `k` was never a good name; now it's worse because it refers to a `Map.Entry`.

jcl/src/jdk.jcmd/share/classes/module-info.java
pdbain-ibm: Replaced tabs with four spaces.  Is that the correct substitution?

jcl/src/jdk.jcmd/share/classes/openj9/tools/attach/diagnostics/info/JvmInfo.java
```diff
@@ -0,0 +1,168 @@
...
+	 * @throws IOException if the properties are invalid
+	 */
+	protected static CompositeData extractFields(DiagnosticProperties props, 
+			String propertyPrefix, CompositeType compType) throws IOException {
+		Set<String> keys = compType.keySet();
+		final int numKeys = keys.size();
+		String[] itemNames = keys.toArray(new String[numKeys]);
+		Object itemValues[] = new Object[numKeys];
+		try {
+			for (int cursor = 0; cursor < numKeys; ++ cursor) {
```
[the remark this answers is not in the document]
pdbain-ibm: Could you clarify?  I don't see an `entrySet()` method in `CompositeType`. https://docs.oracle.com/javase/8/docs/api/javax/management/openmbean/CompositeType.html.
Thank you.

jcl/src/jdk.jcmd/share/classes/openj9/tools/attach/diagnostics/info/ThreadGroupInfo.java
```diff
@@ -0,0 +1,219 @@
...
+				buff.setLength(0);
+				break;
+			}
+			buff.append(ti.toString());
+			LockInfo[] syncs = ti.getLockedSynchronizers();
+			if (addSynchronizers) {
+				buff.append(String.format("\n\tLocked ownable synchronizers: %d\n",  //$NON-NLS-1$
+						Integer.valueOf(syncs.length)));
+				for (LockInfo s: syncs) {
+					buff.append(String.format("\t- %s\n", s.toString())); //$NON-NLS-1$
```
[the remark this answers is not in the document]
pdbain-ibm: Thanks for catching that.

jcl/src/java.base/share/classes/openj9/tools/attach/diagnostics/base/DiagnosticProperties.java
pdbain-ibm: Thanks!

jcl/src/jdk.jcmd/share/classes/module-info.java
keithc-ca: Please follow the pattern in other module-info files: indent with two spaces (so they won't easily get automatically converted to tabs).

jcl/src/jdk.jcmd/share/classes/openj9/tools/attach/diagnostics/info/JvmInfo.java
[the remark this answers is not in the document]
keithc-ca: My mistake: I thought `compType` was a `Map`.

pdbain-ibm:
Updated and pushed.

pshipton:
Please provide an example of what is printed when trying to use jstack on a JVM which doesn't support it. There should be an explanatory message and not just an exception.

pdbain-ibm:
> Please provide an example of what is printed when trying to use jstack on a JVM which doesn't support it. There should be an explanatory message and not just an exception.

Good idea.  Done.  It now prints `Error getting data from 7315: incompatible target JVM`.

jcl/src/jdk.jcmd/share/classes/openj9/tools/attach/diagnostics/info/ThreadGroupInfo.java
```diff
@@ -0,0 +1,220 @@
...
+	 */
+	public ThreadGroupInfo(Properties props, boolean printSynchronizers) throws  IOException  {
+		addSynchronizers = printSynchronizers;
+		try {
+			myProps = new DiagnosticProperties(props);
+			if (myProps.containsField(IPC.PROPERTY_DIAGNOSTICS_ERROR) && myProps.getBoolean(IPC.PROPERTY_DIAGNOSTICS_ERROR)) {
+				final String targetMsg = myProps.getProperty(IPC.PROPERTY_DIAGNOSTICS_ERRORMSG);
+				String msg = myProps.getProperty(IPC.PROPERTY_DIAGNOSTICS_ERRORTYPE); 
+				if ((null != targetMsg) && !targetMsg.isEmpty()) {
+					msg = msg+ ": " + targetMsg; //$NON-NLS-1$
```
keithc-ca: formatting: space around binary operators

jcl/src/jdk.jcmd/share/classes/openj9/tools/attach/diagnostics/info/ThreadGroupInfo.java
```diff
@@ -0,0 +1,220 @@
...
+			try {
+				final String cpuTimeKey = prefix + CPU_TIME;
+				final String userTimeKey = prefix + USER_TIME;
+				if (myProps.containsField(cpuTimeKey)) {
+					long cpuTime = myProps.getLong(cpuTimeKey);
+					buff.append(String.format(TIME_FORMAT_STRING, "CPU", Long.valueOf(cpuTime))); //$NON-NLS-1$
+				}
+				if (myProps.containsField(userTimeKey)) {
+					long userTime = myProps.getLong(userTimeKey);
+					buff.append(String.format(TIME_FORMAT_STRING, "User ", Long.valueOf(userTime))); //$NON-NLS-1$
```
keithc-ca: Why the space in `"User "`?

jcl/src/jdk.jcmd/share/classes/openj9/tools/attach/diagnostics/tools/Jstack.java
```diff
@@ -0,0 +1,157 @@
...
+				okay = false;
+				break;
+			}
+		}
+
+		if (okay && vmids.isEmpty()) {
+			/* grab the VMIDs from stdin. Omit my own ID. */
+			try (BufferedReader jpsOutReader = new BufferedReader(new InputStreamReader(System.in))) {
+				vmids = jpsOutReader.lines()
+						.map(s -> s.trim()).filter(s -> !s.isEmpty())
```
keithc-ca: This should also be split before `.filter`.

test/functional/Java8andUp/src/org/openj9/test/attachAPI/AttachApiTest.java
```diff
@@ -200,5 +215,26 @@ protected boolean checkLoadAgentOutput(String errOutput,
...
+	protected void getJdkUtilityPath(final String commandRoot) {
+		char fs = java.io.File.separatorChar;
+		final String commandWithSuffix = 
+				PlatformInfo.isWindows() ? commandRoot + ".exe": commandRoot; //$NON-NLS-1$
+		String binDir = System.getProperty("java.home") + fs + "bin" + fs; //$NON-NLS-1$ //$NON-NLS-2$
+		commandName = binDir + commandWithSuffix;
+		File  commandFile = new File(commandName);
+		if (!commandFile.exists()) { /* Java 8 has 2 bin directories, and JAVA_HOME is probably pointing at jre/bin */
+			log("Did not find jps in "+binDir); //$NON-NLS-1$
+			binDir = System.getProperty("java.home") + fs + ".." + fs + "bin" + fs; //$NON-NLS-1$ //$NON-NLS-2$ //$NON-NLS-3$
```
keithc-ca: This should use `File` API rather than constructing paths as strings.

test/functional/Java8andUp/src/org/openj9/test/attachAPI/TestJstack.java
```diff
@@ -0,0 +1,231 @@
...
+	Object syncObject = new Object();
+
+	@Test
+	public void testSingleThread() throws IOException {
+		try {
+			Thread testThread = new Thread(this::method1);
+			testThread.start();
+			while (testThread.getState() != State.WAITING) {
+				try {
+					Thread.sleep(1);
```
keithc-ca: A millisecond is a very short sleep; perhaps you meant 1000 (for one second)?

test/functional/Java8andUp/src/org/openj9/test/attachAPI/TestJstack.java
```diff
@@ -0,0 +1,231 @@
...
+				} catch (InterruptedException e) {
+					/* ignore */
+				}
+			}
+			Properties props = ThreadGroupInfo.getThreadInfoProperties().toProperties();
+			logProperties(props);
+			ThreadGroupInfo grpInfo = new ThreadGroupInfo(props, false);
+			final String jstackOutput = grpInfo.toString();
+			log("jstack output:\n"+jstackOutput); //$NON-NLS-1$
```
keithc-ca: formatting: space around binary operators

jcl/src/java.base/share/classes/openj9/tools/attach/diagnostics/base/DiagnosticProperties.java
```diff
@@ -0,0 +1,260 @@
...
+	/**
+	 * Print a Properties object in sorted key order
+	 * @param props object to dump
+	 * @return String representation of props
+	 */
+	public static String printProperties(final Properties props) {
+		final StringWriter buff = new StringWriter(1000);
+		try (PrintWriter buffWriter = new PrintWriter(buff)) {
+			props.entrySet().stream().sorted(Comparator.comparing(e -> e.getKey().toString()))
+			.forEach(theEntry -> {
```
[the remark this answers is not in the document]
pdbain-ibm: Will do.  Do you know how to make Eclipse use that style?

jcl/src/jdk.jcmd/share/classes/openj9/tools/attach/diagnostics/info/ThreadGroupInfo.java
pdbain-ibm: Thanks for catching that.

test/functional/Java8andUp/src/org/openj9/test/attachAPI/TestJstack.java
pdbain-ibm: Actually I meant that, since the thread should go into the waiting state very quickly.  Will add a comment.

test/functional/Java8andUp/src/org/openj9/test/attachAPI/TestJstack.java
pdbain-ibm: Fixed here and elsewhere in the file.

jcl/src/java.base/share/classes/com/ibm/tools/attach/target/Attachment.java
```diff
@@ -260,9 +277,43 @@ boolean doCommand(InputStream cmdStream, OutputStream respStream) {
...
+			for (TargetDiagnosticsProvider p: ServiceLoader.load(TargetDiagnosticsProvider.class)) {
+				if (null != p) {
+					diagProvider = p;
+				}
+			}
+		}
+		return diagProvider;
+/*[ELSE]*/ } /* Java 8 */
+/*[IF !NoAttachAPI]*/ if (true) { /* Full Java 8 */
+		return new TargetDiagnosticsProviderImpl();
```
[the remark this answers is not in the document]
pdbain-ibm: Will do.

jcl/src/java.base/share/classes/openj9/tools/attach/diagnostics/base/DiagnosticProperties.java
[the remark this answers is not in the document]
keithc-ca: In my personal projects, I use `// <br/>` to break lines and guide the formatter, like this:
```
props.entrySet().stream() // <br/>
	.sorted(Comparator.comparing(e -> e.getKey().toString())) // <br/>
	.forEach(theEntry -> {
```
This isn't a popular strategy, so it's up to you to make it easy to follow the structure of the code you write.

test/functional/Java8andUp/src/org/openj9/test/attachAPI/TestJstack.java
keithc-ca: To be clear, such suggestions are meant to apply to the pull request as a whole.

pdbain-ibm:
Updated and sanity tested on Java 8. Will push the Java 8 extensions changes tomorrow.

pdbain-ibm:
Pushed changes to Java 8 extensions to implement service provider.  Fixed a few more hygiene issues.
Testing in a personal build now.

jcl/src/java.base/share/classes/openj9/tools/attach/diagnostics/base/DiagnosticProperties.java
```diff
@@ -0,0 +1,256 @@
...
+	public Object getSimple(String typeName, String key) throws NumberFormatException {
+		Object value = null;
+		String valueString = baseProperties.getProperty(key);
+		if (null != valueString) {
+			switch (typeName) {
+			case "java.lang.Boolean": //$NON-NLS-1$
+				value = Boolean.valueOf(valueString);
+				break;
+			case "java.lang.Character": //$NON-NLS-1$
+				value = Character.valueOf(valueString.charAt(0));
```
keithc-ca: This might throw `StringIndexOutOfBoundsException`.

jcl/src/jdk.jcmd/share/classes/openj9/tools/attach/diagnostics/info/JvmInfo.java
```diff
@@ -0,0 +1,168 @@
...
+					}
+				}
+			}
+		}
+	}
+
+	/**
+	 * Get the CompositeType object based on a name
+	 * @param typeIdentifier Name of the type
+	 * @return CompositeType for that type
```
keithc-ca: This may return null: please document that.

jcl/src/jdk.jcmd/share/classes/openj9/tools/attach/diagnostics/tools/Jstack.java
```diff
@@ -0,0 +1,157 @@
...
+					out.println("System properties:"); //$NON-NLS-1$
+					out.println(diagProvider.getSystemProperties());
+					out.println("Agent properties:"); //$NON-NLS-1$
+					out.println(diagProvider.getAgentProperties());
+				}
+			} catch (Exception e) {
+				System.err.printf("Error getting data from %s", vmid); //$NON-NLS-1$
+				final String msg = e.getMessage();
+				if (null == msg) {
+					System.err.println(""); //$NON-NLS-1$
```
keithc-ca: `println()` (with no arguments) does the same thing (and is preferred).

jcl/src/jdk.jcmd/share/classes/openj9/tools/attach/diagnostics/tools/Jstack.java
```diff
@@ -0,0 +1,157 @@
...
+		vmids = new ArrayList<>();
+		for (String a: args) {
+			if (!a.startsWith("-")) {
+				vmids.add(a);
+			} else if (a.equals("-p")) {
+				printProperties = true;
+			} else if (a.equals("-l")) {
+				printSynchronizers = true;
+			} else {
+				System.err.printf(HELPTEXT, "");
```
keithc-ca: The second argument is unused and not required.

test/functional/Java8andUp/src/org/openj9/test/attachAPI/AttachApiTest.java
```diff
@@ -200,5 +215,25 @@ protected boolean checkLoadAgentOutput(String errOutput,
...
+		char fs = java.io.File.separatorChar;
+		final String commandWithSuffix = 
+				PlatformInfo.isWindows() ? commandRoot + ".exe": commandRoot; //$NON-NLS-1$
+		final String javaHomeString = System.getProperty("java.home");
+		File javaHome = new File(javaHomeString);
+		File javaHomeBin = new File(javaHome, "bin");
+		File  commandFile = new File(javaHomeBin, commandWithSuffix);
+		if (!commandFile.exists()) { /* Java 8 has 2 bin directories, and JAVA_HOME is probably pointing at jre */
+			log("Did not find " + commandFile.getAbsolutePath()); //$NON-NLS-1$
+			File javaHomeBinParentBin = new File(javaHomeBin.getParentFile(), "bin");
```
keithc-ca: This is the same as `javaHome`; I think you meant (also notice the variable name change):
> File javaHomeParentBin = new File(javaHome.getParentFile(), "bin");

test/functional/Java8andUp/src/org/openj9/test/attachAPI/TestJstack.java
keithc-ca: I don't see that comment you promised.

test/functional/Java8andUp/src/org/openj9/test/attachAPI/TestJstack.java
```diff
@@ -0,0 +1,231 @@
...
+	}
+
+	@Test
+	public void testMonitors() throws IOException {
+		synchronized (syncObject) {
+			Thread testThread = new Thread(new Runnable() {
+				@Override
+				public void run() {
+					syncMethod();
+				}
```
keithc-ca: The `Runnable` can be replaced by `this::syncMethod`.

test/functional/Java8andUp/src/org/openj9/test/attachAPI/TestJstack.java
```diff
@@ -0,0 +1,231 @@
...
+		jpsOutput.forEach(s -> {buff.append(s); buff.append("\n");}); //$NON-NLS-1$
+		log(buff.toString());
+		Optional<String> searchResult =  search(testName, jpsOutput);
+		assertTrue(searchResult.isPresent(), "Method name missing"); //$NON-NLS-1$
+	}
+
+	@Test
+	public void testSynchronizers() throws IOException {
+		String myId = TargetManager.getVmId();
+		ArrayList<String> args = new ArrayList<>();
```
keithc-ca: The variable type can be just `List<String>` like below: please be consistent.

test/functional/Java8andUp/src/org/openj9/test/attachAPI/TestJstack.java
```diff
@@ -0,0 +1,231 @@
...
+		assertEquals(thrInfo.getThreadName(), testThread.getName(), "wrong name:"); //$NON-NLS-1$
+		assertEquals(thrInfo.getThreadState(), testThread.getState(), "wrong state:"); //$NON-NLS-1$
+		StackTraceElement[] expectedStackTrace = testThread.getStackTrace();
+		StackTraceElement[] actualStackTrace = testThread.getStackTrace();
+		compareStackTraces(expectedStackTrace, actualStackTrace);
+	}
+
+	protected void compareStackTraces(StackTraceElement[] expectedStackTrace, StackTraceElement[] actualStackTrace) {
+		boolean match = (expectedStackTrace.length == actualStackTrace.length);
+		if (match) {
```
keithc-ca: This is mostly a reimplementation of `Arrays.equals()` which should be used instead.

jcl/src/java.base/share/classes/openj9/tools/attach/diagnostics/base/DiagnosticProperties.java
pdbain-ibm: Right.  Checking for empty strings.

test/functional/Java8andUp/src/org/openj9/test/attachAPI/AttachApiTest.java
pdbain-ibm: Yup.  Found that in testing.

pdbain-ibm:
Thanks. Updated and retested.

jcl/src/java.base/share/classes/openj9/tools/attach/diagnostics/base/DiagnosticProperties.java
```diff
@@ -0,0 +1,256 @@
...
+		return Boolean.parseBoolean(baseProperties.getProperty(key));
+	}
+
+	/**
+	 * Return a property value for the given key
+	 * @param key property name
+	 * @return property value or null if the property is not found
+	 */
+	public String getProperty(String key) {
+		return baseProperties.getProperty(key);
```
keithc-ca: Why does this method not include a call to `checkExists(key)`?

jcl/src/java.base/share/classes/openj9/tools/attach/diagnostics/base/DiagnosticProperties.java
keithc-ca: Now this doesn't support an empty `java.lang.String`: might that be a problem?

jcl/src/jdk.jcmd/share/classes/openj9/tools/attach/diagnostics/attacher/AttacherDiagnosticsProvider.java
```diff
@@ -0,0 +1,122 @@
...
+
+/**
+ * This class allows a Attach API attacher to query a target JVM about diagnostic information
+ * such as JIT compilation, threads, classes, etc.
+ * Instances must be created using the getDiagnostics() factory method.
+ *
+ */
+public class AttacherDiagnosticsProvider {
+
+	OpenJ9VirtualMachine vm;
```
keithc-ca: Why is this field not `private`?

jcl/src/jdk.jcmd/share/classes/openj9/tools/attach/diagnostics/info/JvmInfo.java
```diff
@@ -0,0 +1,168 @@
...
+ * Methods for converting between properties and CompositeTypes
+ *
+ */
+public class JvmInfo {
+	protected static final String ARRAY_SIZE = "array_size"; //$NON-NLS-1$
+	protected static void addFields(DiagnosticProperties props, String keyPrefix,
+			final CompositeType compType, CompositeData compData) {
+		for (String k: compType.keySet()) {
+			final String keyString = keyPrefix + k;
+			Object value = compData.get(k);
```
keithc-ca: Why is this local not `final` (like the others)?

jcl/src/jdk.jcmd/share/classes/openj9/tools/attach/diagnostics/info/JvmInfo.java
```diff
@@ -0,0 +1,168 @@
...
+				if (JvmInfo.isSimpleType(valueType)) {
+					value = props.getSimple(valueType.getTypeName(), keyString);
+				} else {
+					boolean valuePresent = props.getBoolean(keyString);
+					if (valuePresent) {
+						if (valueType.isArray()) {
+							final int arrayLength = props.getInt(keyString + ARRAY_SIZE);
+							ArrayType<?> valueArrayType = (ArrayType<?>) valueType;
+							CompositeType elementType = getCompositeType(valueArrayType.getElementOpenType());
+							CompositeData valueArray[] = new CompositeData[arrayLength];
```
keithc-ca: The brackets should be with the type: `CompositeData[] valueArray = ...`.

jcl/src/jdk.jcmd/share/classes/openj9/tools/attach/diagnostics/tools/Jstack.java
```diff
@@ -0,0 +1,157 @@
...
+				printSynchronizers = true;
+			} else {
+				System.err.printf(HELPTEXT);
+				okay = false;
+				break;
+			}
+		}
+
+		if (okay && vmids.isEmpty()) {
+			/* grab the VMIDs from stdin. Omit my own ID. */
```
keithc-ca: The code below does *not* omit the executor's id. Please fix the code or the comment.

test/functional/Java8andUp/src/org/openj9/test/attachAPI/AttachApiTest.java
```diff
@@ -200,5 +215,26 @@ protected boolean checkLoadAgentOutput(String errOutput,
...
 	}
+
+	/**
+	 * Find an executable in the jdk directory and fail if it is missing.
+	 * @param commandRoot Base name of the command, without ".exe" 
+	 */
+	protected void getJdkUtilityPath(final String commandRoot) {
+		char fs = java.io.File.separatorChar;
+		final String commandWithSuffix = 
+				PlatformInfo.isWindows() ? commandRoot + ".exe": commandRoot; //$NON-NLS-1$
```
keithc-ca: `fs` appears to be unused: please remove it.
formatting: spaces around operators (`:` here).
Please be consistent about the use of `//$NON-NLS-n$`; the comment appears twice below but there are more string literals without the comment.

jcl/src/java.base/share/classes/openj9/tools/attach/diagnostics/base/DiagnosticProperties.java
pdbain-ibm: There are places where the caller wants to do the checking.

pdbain-ibm:
> Now this doesn't support an empty java.lang.String: might that be a problem?

https://github.com/eclipse/openj9/pull/5150#discussion_r270889673

In this case, an empty string is illegal.

jcl/src/jdk.jcmd/share/classes/openj9/tools/attach/diagnostics/info/JvmInfo.java
pdbain-ibm: Good idea.  Fixed a number of other cases too.

jcl/src/jdk.jcmd/share/classes/openj9/tools/attach/diagnostics/info/JvmInfo.java
pdbain-ibm: Okay.  Fixed a few other places too.

test/functional/Java8andUp/src/org/openj9/test/attachAPI/AttachApiTest.java
pdbain-ibm: Thanks. NLS warnings are suppressed in the file, so I will leave out the annotations for now.  I will fix them in a separate change set.

pdbain-ibm:
Thanks for the tip on cmd-shift-F @keithc-ca . Reformatted the new code and classes, cleaned up comments. 
`Now this doesn't support an empty java.lang.String: might that be a problem?` True for the case of typeName=String.  Thanks for catching that.  Retested in personal build.

jcl/src/java.base/share/classes/com/ibm/tools/attach/target/Attachment.java
```diff
@@ -425,7 +462,7 @@ static String saveLocalConnectorAddress() {
 		synchronized (systemProperties) {
 			addr = systemProperties.getProperty(LOCAL_CONNECTOR_ADDRESS);
 			if (Objects.isNull(addr)) {
-				addr = VMSupport.getAgentProperties().getProperty(LOCAL_CONNECTOR_ADDRESS);
+				addr = VMSupport.getAgentProperties().getPropertyOrNull(LOCAL_CONNECTOR_ADDRESS);
```
keithc-ca: This won't compile: `Properties` has no such method.

keithc-ca:
Jenkins test extended plinux,win jdk8,jdk11

jcl/src/jdk.jcmd/share/classes/module-info.java
```diff
@@ -0,0 +1,32 @@
...
+ *******************************************************************************/
+
+/*[REM] This file must not use tabs because the dependency recognition code in openjdk does not support them. */
+
+module jdk.jcmd {
+  exports openj9.tools.attach.diagnostics.info;
+  provides openj9.tools.attach.diagnostics.spi.TargetDiagnosticsProvider with openj9.tools.attach.diagnostics.target.TargetDiagnosticsProviderImpl;
+  requires java.base;
+  requires java.management;
+  requires jdk.attach;
```
keithc-ca: Adding a `requires` class for `jdk.internal.jvmstat` allows jdk11 to build without changes to that extension repo.

keithc-ca:
Jenkins test extended plinux,win jdk8 depends ibmruntimes/openj9-openjdk-jdk8#279

keithc-ca:
Jenkins test extended plinux,win jdk11 depends ibmruntimes/openj9-openjdk-jdk11#146

pshipton:
@pdbain-ibm please create a PR for the 0.14 branch.

pdbain-ibm:
> @pdbain-ibm please create a PR for the 0.14 branch.

https://github.com/eclipse/openj9/pull/5336

PR #5151, opened 2019-03-19
Improve the error reporting when j9shmem_getDir() failed.

### LINKED ISSUE #5134: Having a NFS home directory gives confusing shared cache error message
If the user's home directory is set to a network (NFS) location, trying to use/create a shared cache results in the following error, which is cryptic. Better to have an error which is understandable, without having to look any further.

`JVMSHRC404E Failed to get a directory for the cache`

Further, this message code has the following explanation, which isn't appropriate for all the failure cases.

`J9NLS_SHRC_OSCACHE_GETCACHEDIR_FAILED.explanation=The directory path requested is bigger than the internal buffer space`

hangshao0:
1. Change j9shmem_getDir() to return different error code for different
failures.
2. Change the caller of j9shmem_getDir()to print new NLS messages
according to the error code returned from j9shmem_getDir().
3. Update the test code to check if j9shmem_getDir() returns a negative
value instead of -1.

Fixes #5134


hangshao0:
Manually set HOME to NFS:

```
java -Xshareclasses:name=Cache1 -version
JVMSHRC831E Failed to use user's home as the default shared cache directory. It is on a Network File System. Please set another directory via environment variable "HOME" or fix the home directory in the password file entry.
JVMSHRC831E Failed to use user's home as the default shared cache directory. It is on a Network File System. Please set another directory via environment variable "HOME" or fix the home directory in the password file entry.
JVMJ9VM015W Initialization error for library j9shr29(11): JVMJ9VM009E J9VMDllMain failed
Error: Could not create the Java Virtual Machine.
Error: A fatal exception has occurred. Program will exit.

java -Xshareclasses:name=Cache1,silent -version
JVMJ9VM015W Initialization error for library j9shr29(11): JVMJ9VM009E J9VMDllMain failed
Error: Could not create the Java Virtual Machine.
Error: A fatal exception has occurred. Program will exit.

java -Xshareclasses:name=Cache1,nonfatal -version
JVMSHRC831E Failed to use user's home as the default shared cache directory. It is on a Network File System. Please set another directory via environment variable "HOME" or fix the home directory in the password file entry.
JVMSHRC831E Failed to use user's home as the default shared cache directory. It is on a Network File System. Please set another directory via environment variable "HOME" or fix the home directory in the password file entry.
JVMSHRC686I Failed to startup shared class cache. Continue without using it as -Xshareclasses:nonfatal is specified
openjdk version "1.8.0_202-internal"
OpenJDK Runtime Environment (build 1.8.0_202-internal-jenkins_2019_03_18_14_23-b00)
Eclipse OpenJ9 VM (build FixSCC-a41867c5f, JRE 1.8.0 Linux amd64-64-Bit Compressed References 20190318_863 (JIT enabled, AOT enabled)
OpenJ9   - 9f1746bad
OMR      - 9b73e2bd
JCL      - 6894a182ad1 based on jdk8u202-b08)
```

The reason JVMSHRC831E shows up twice is because we try getting the cache dir both in checkIfCacheExists() and SH_OSCache::commonStartup(). Without this change, there were also be 2 NLS messages printed out in checkIfCacheExists() and SH_OSCache::commonStartup().

runtime/nls/shrc/j9shr.nls
```diff
@@ -6694,3 +6694,46 @@ J9NLS_SHRC_CM_PRINTSTATS_SUMMARY_NUM_STARTUP_HINTS.explanation=NOTAG
...
 # END NON-TRANSLATABLE
+
+J9NLS_SHRC_GET_DIR_BUF_OVERFLOW=The shared cache directory path is too long.
+# START NON-TRANSLATABLE
+J9NLS_SHRC_GET_DIR_BUF_OVERFLOW.explanation=The directory path requested is bigger than the internal buffer space
+J9NLS_SHRC_GET_DIR_BUF_OVERFLOW.system_action=The JVM terminates, unless you have specified the nonfatal option with "-Xshareclasses:nonfatal", in which case the JVM continues without using Shared Classes.
+J9NLS_SHRC_GET_DIR_BUF_OVERFLOW.user_response=Please try using a shorter shared cache directory.
+# END NON-TRANSLATABLE
+
+J9NLS_SHRC_GET_DIR_FAILED_TO_GET_HOME=Failed to use user's home as the default shared cache directory. Cannot get home directory. Please set another directory via environment variable "HOME" or fix the home directory in the password file entry.
```
pshipton: Setting the cacheDir is also an option. Similarly for the following messages.

runtime/shared/shrclssup.c
```diff
@@ -181,7 +181,31 @@ IDATA J9VMDllMain(J9JavaVM* vm, IDATA stage, void* reserved)
...
+					if (0 != verboseFlags) {
+						switch(rc) {
+						case J9PORT_ERROR_SHMEM_GET_DIR_BUF_OVERFLOW:
+							SHRCLSSUP_ERR_TRACE(verboseFlags, J9NLS_SHRC_GET_DIR_BUF_OVERFLOW);
+							break;
+						case J9PORT_ERROR_SHMEM_GET_DIR_FAILED_TO_GET_HOME:
+							SHRCLSSUP_ERR_TRACE(verboseFlags, J9NLS_SHRC_GET_DIR_FAILED_TO_GET_HOME);
+							break;
+						case J9PORT_ERROR_SHMEM_GET_DIR_HOME_BUF_OVERFLOW:
+							j9nls_printf(PORTLIB, J9NLS_ERROR, J9NLS_SHRC_GET_DIR_HOME_BUF_OVERFLOW);
```
pshipton: This one using j9nls_printf() is inconsistent with the others. Although they could all use j9nls_printf() since the verboseFlags are checked above.

runtime/shared_common/OSCache.cpp
```diff
@@ -210,7 +210,32 @@ SH_OSCache::getCacheDir(J9JavaVM* vm, const char* ctrlDirName, char* buffer, UDA
 #endif /*defined(OPENJ9_BUILD) */
 
 	rc = j9shmem_getDir(ctrlDirName, flags, buffer, bufferSize);
-	if (rc == -1) {
+
+	if (rc < 0) {
+		if (0 != vm->sharedCacheAPI->verboseFlags) {
```
pshipton: Seems this code could be in a helper function, so it's not duplicated twice.

pshipton:
Is it much work to fix the error to only print once?

runtime/shared_common/OSCache.cpp
hangshao0: Changed the code to call new helper function j9shr_getCacheDir().

runtime/shared/shrclssup.c
hangshao0: Changed the code to call new helper function j9shr_getCacheDir().

hangshao0:
> Is it much work to fix the error to only print once?

We need to change the signature of SH_OSCache::getCacheDir() to control whether the NLS messages should be printed. There are about 20 callers.

pshipton:
jenkins test sanity win,plinux jdk11

pshipton:
> We need to change the signature of SH_OSCache::getCacheDir() to control whether the NLS messages should be printed. There are about 20 callers.

Seems better done separately.

runtime/shared/shrclssup.c
```diff
@@ -178,10 +178,18 @@ IDATA J9VMDllMain(J9JavaVM* vm, IDATA stage, void* reserved)
...
+				}
+				/* set runtimeFlags and verboseFlags here as they will be used later in SH_OSCache::getCacheDir() */
+				vm->sharedCacheAPI->runtimeFlags = runtimeFlags;
+				vm->sharedCacheAPI->verboseFlags = verboseFlags;
+
 #if !defined(WIN32) && !defined(WIN64)
 				/* Get platform default cache directory */
-				rc = j9shmem_getDir(NULL, J9SHMEM_GETDIR_APPEND_BASEDIR, defaultCacheDir, J9SH_MAXPATH);
-				if (-1 == rc) {
+				if (-1 == j9shr_getCacheDir(vm, NULL, defaultCacheDir, J9SH_MAXPATH, vm->sharedCacheAPI->cacheType)) {
```
keithc-ca: This condition will always be false.
>  * @internal J9PORT_ERROR_SHMEM* range from at -170 to -199 to match hyporterror.h codes.

pshipton: I think that is intentional? If it did happen we'd hit the assert.

pshipton: j9shr_getCacheDir() and SH_OSCache::getCacheDir() returns -1 on failure. I think the assert needs to be removed.

hangshao0: I will remove this assert in the next PR.

hangshao0: PR: https://github.com/eclipse/openj9/pull/5202

pshipton: You found a problem anyway :)

PR #5161, opened 2019-03-19
Add Flattened ValueType Array Implementation

AidanHa:
Steps:
Add Macros J9_IS_J9CLASS_FLATTENED and J9_VALUETYPE_FLATTENED_SIZE to help determine if the J9Class instance is flatenned
Add Stride calculation for Array classes in creatorramclass.cpp. Utilize the J9ARRAYCLASS_SET_STRIDE that was previously implemented
Modify the space allocation calculation for inlineAllocateIndexableObject and getDataSizeInBytes using the J9ARRAYCLASS_GET_STRIDE Macro

Related #5021

### DIFF
--- a/runtime/gc_glue_java/ArrayletObjectModel.hpp
+++ b/runtime/gc_glue_java/ArrayletObjectModel.hpp
@@ -1,6 +1,6 @@
 
 /*******************************************************************************
- * Copyright (c) 1991, 2017 IBM Corp. and others
+ * Copyright (c) 1991, 2019 IBM Corp. and others
  *
  * This program and the accompanying materials are made available under
  * the terms of the Eclipse Public License 2.0 which accompanies this
@@ -54,6 +54,7 @@ public:
 * Function members
 */
 private:
+	void AssertBadElementSize();
 protected:
 	/* forward declare methods from parent class to avoid namespace issues */
 	MMINLINE UDATA
@@ -300,12 +301,11 @@ public:
 	 */
 	MMINLINE UDATA
 	getDataSizeInBytes(J9Class *clazzPtr, UDATA numberOfElements)
-	{
+	{	
+		UDATA stride = J9ARRAYCLASS_GET_STRIDE(clazzPtr);
+		UDATA size = numberOfElements * stride;
 		UDATA alignedSize = UDATA_MAX;
-		J9ROMArrayClass *romArrayClass = (J9ROMArrayClass *)clazzPtr->romClass;
-		UDATA shift = romArrayClass->arrayShape & 0x0000FFFF;
-		UDATA size = numberOfElements << shift;
-		if ((size >> shift) == numberOfElements) {
+		if ((size / stride) == numberOfElements) {
 			alignedSize = MM_Math::roundToSizeofUDATA(size);
 			if (alignedSize < size) {
 				alignedSize = UDATA_MAX;
@@ -402,14 +402,13 @@ public:
 	/*MMINLINE*/ void
 	memcpyFromArray(void *destData, J9IndexableObject *srcObject, I_32 elementIndex, I_32 elementCount)
 	{
-		J9ROMArrayClass *romArrayClass = (J9ROMArrayClass*)J9GC_J9OBJECT_CLAZZ(srcObject)->romClass;
-		UDATA elementScale = romArrayClass->arrayShape & 0x0000FFFF;
+		UDATA elementSize = J9ARRAYCLASS_GET_STRIDE(J9GC_J9OBJECT_CLAZZ(srcObject));
 		if (isInlineContiguousArraylet(srcObject)) {
 			// If the data is stored contiguously, then a simple copy is sufficient.
 			void* srcData = getDataPointerForContiguous(srcObject);
-			switch(elementScale)
+			switch(elementSize)
 			{
-			case 0:
+			case 1:
 				// byte/boolean
 				{
 					U_8* srcCursor = ((U_8*)srcData) + elementIndex;
@@ -421,7 +420,7 @@ public:
 				}
 				break;
 
-			case 1:
+			case 2:
 				// short/char
 				{
 					U_16* srcCursor = ((U_16*)srcData) + elementIndex;
@@ -433,7 +432,7 @@ public:
 				}
 				break;
 
-			case 2:
+			case 4:
 				// int/float
 				{
 					U_32* srcCursor = ((U_32*)srcData) + elementIndex;
@@ -445,7 +444,7 @@ public:
 				}
 				break;
 
-			case 3:
+			case 8:
 				// long/double
 				{
 					U_64* srcCursor = ((U_64*)srcData) + elementIndex;
@@ -456,12 +455,19 @@ public:
 					}
 				}
 				break;
+				
+			default :
+				// unreachable since we currently do not expect anything other than primitive arrays to be using them.
+				{
+					AssertBadElementSize();
+				}
+				break;
 			}
 		} else {
 			fj9object_t *srcArraylets = getArrayoidPointer(srcObject);
 			void *outerDestCursor = destData;
 			U_32 outerCount = elementCount;
-			UDATA arrayletLeafElements = _omrVM->_arrayletLeafSize >> elementScale;
+			UDATA arrayletLeafElements = _omrVM->_arrayletLeafSize / elementSize;
 			UDATA arrayletIndex = elementIndex / arrayletLeafElements;
 			UDATA arrayletElementOffset = elementIndex % arrayletLeafElements;
 			while(outerCount > 0) {
@@ -473,9 +479,9 @@ public:
 					// Copy as much as we can
 					innerCount = (U_32)(arrayletLeafElements - arrayletElementOffset);
 				}
-				switch(elementScale)
+				switch(elementSize)
 				{
-				case 0:
+				case 1:
 					// byte/boolean
 					{
 						U_8* srcCursor = ((U_8*)srcLeafAddress) + arrayletElementOffset;
@@ -489,7 +495,7 @@ public:
 					}
 					break;
 
-				case 1:
+				case 2:
 					// short/char
 					{
 						U_16* srcCursor = ((U_16*)srcLeafAddress) + arrayletElementOffset;
@@ -503,7 +509,7 @@ public:
 					}
 					break;
 
-				case 2:
+				case 4:
 					// int/float
 					{
 						U_32* srcCursor = ((U_32*)srcLeafAddress) + arrayletElementOffset;
@@ -517,7 +523,7 @@ public:
 					}
 					break;
 
-				case 3:
+				case 8:
 					// long/double
 					{
 						U_64* srcCursor = ((U_64*)srcLeafAddress) + arrayletElementOffset;
@@ -528,6 +534,13 @@ public:
 						}
 						outerDestCursor = (void*)destCursor;
 						arrayletElementOffset = 0;
+					}
+					break;
+
+				default :
+					// unreachable since we currently do not expect anything other than primitive arrays to be using them.
+					{
+						AssertBadElementSize();
 					}
 					break;
 				}
@@ -547,14 +560,13 @@ public:
 	MMINLINE void
 	memcpyToArray(J9IndexableObject *destObject, I_32 elementIndex, I_32 elementCount, void *srcData)
 	{
-		J9ROMArrayClass *romArrayClass = (J9ROMArrayClass*)J9GC_J9OBJECT_CLAZZ(destObject)->romClass;
-		UDATA elementScale = romArrayClass->arrayShape & 0x0000FFFF;
+		UDATA elementSize = J9ARRAYCLASS_GET_STRIDE(J9GC_J9OBJECT_CLAZZ(destObject));
 		if (isInlineContiguousArraylet(destObject)) {
 			// If the data is stored contiguously, then a simple copy is sufficient.
 			void* destData = getDataPointerForContiguous(destObject);
-			switch(elementScale)
+			switch(elementSize)
 			{
-			case 0:
+			case 1:
 				// byte/boolean
 				{
 					U_8* srcCursor = (U_8*)srcData;
@@ -566,7 +578,7 @@ public:
 				}
 				break;
 
-			case 1:
+			case 2:
 				// short/char
 				{
 					U_16* srcCursor = (U_16*)srcData;
@@ -578,7 +590,7 @@ public:
 				}
 				break;
 
-			case 2:
+			case 4:
 				// int/float
 				{
 					U_32* srcCursor = (U_32*)srcData;
@@ -590,7 +602,7 @@ public:
 				}
 				break;
 
-			case 3:
+			case 8:
 				// long/double
 				{
 					U_64* srcCursor = (U_64*)srcData;
@@ -601,12 +613,19 @@ public:
 					}
 				}
 				break;
+
+			default :
+				// unreachable since we currently do not expect anything other than primitive arrays to be using them.
+				{
+					AssertBadElementSize();
+				}
+				break;
 			}
 		} else {
 			fj9object_t *destArraylets = getArrayoidPointer(destObject);
 			void *outerSrcCursor = srcData;
 			U_32 outerCount = elementCount;
-			UDATA arrayletLeafElements = _omrVM->_arrayletLeafSize >> elementScale;
+			UDATA arrayletLeafElements = _omrVM->_arrayletLeafSize / elementSize;
 			UDATA arrayletIndex = elementIndex / arrayletLeafElements;
 			UDATA arrayletElementOffset = elementIndex % arrayletLeafElements;
 			while(outerCount > 0) {
@@ -618,9 +637,9 @@ public:
 					// Copy as much as we can
 					innerCount = (U_32)(arrayletLeafElements - arrayletElementOffset);
 				}
-				switch(elementScale)
+				switch(elementSize)
 				{
-				case 0:
+				case 1:
 					// byte/boolean
 					{
 						U_8* srcCursor = (U_8*)outerSrcCursor;
@@ -634,7 +653,7 @@ public:
 					}
 					break;
 
-				case 1:
+				case 2:
 					// short/char
 					{
 						U_16* srcCursor = (U_16*)outerSrcCursor;
@@ -648,7 +667,7 @@ public:
 					}
 					break;
 
-				case 2:
+				case 4:
 					// int/float
 					{
 						U_32* srcCursor = (U_32*)outerSrcCursor;
@@ -662,7 +681,7 @@ public:
 					}
 					break;
 
-				case 3:
+				case 8:
 					// long/double
 					{
 						U_64* srcCursor = (U_64*)outerSrcCursor;
@@ -673,6 +692,13 @@ public:
 						}
 						outerSrcCursor = (void*)srcCursor;
 						arrayletElementOffset = 0;
+					}
+					break;
+					
+				default :
+					// unreachable since we currently do not expect anything other than primitive arrays to be using them.
+					{
+						AssertBadElementSize();
 					}
 					break;
 				}
@@ -815,8 +841,7 @@ public:
 	getHashcodeOffset(J9Class *clazzPtr, ArrayLayout layout, UDATA numberOfElements)
 	{
 		/* Don't call getDataSizeInBytes() since that rounds up to UDATA. */
-		J9ROMArrayClass *romArrayClass = (J9ROMArrayClass *)clazzPtr->romClass;
-		UDATA dataSize = numberOfElements << (romArrayClass->arrayShape & 0x0000FFFF);
+		UDATA dataSize = numberOfElements * J9ARRAYCLASS_GET_STRIDE(clazzPtr);
 		UDATA numberOfArraylets = numArraylets(dataSize);
 		bool alignData = shouldAlignSpineDataSection(clazzPtr);
 		UDATA spineSize = getSpineSize(clazzPtr, layout, numberOfArraylets, dataSize, alignData);
--- a/runtime/gc_include/ObjectAllocationAPI.hpp
+++ b/runtime/gc_include/ObjectAllocationAPI.hpp
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 1991, 2018 IBM Corp. and others
+ * Copyright (c) 1991, 2019 IBM Corp. and others
  *
  * This program and the accompanying materials are made available under
  * the terms of the Eclipse Public License 2.0 which accompanies this
@@ -168,14 +168,13 @@ public:
 		if (0 != size) {
 			/* Contiguous Array */
 
-			UDATA scale = ((J9ROMArrayClass*)(arrayClass->romClass))->arrayShape;
 
 #if !defined(J9VM_ENV_DATA64)
-			if (!sizeCheck || (size < ((U_32)J9_MAXIMUM_INDEXABLE_DATA_SIZE >> scale)))
+			if (!sizeCheck || (size < ((U_32)J9_MAXIMUM_INDEXABLE_DATA_SIZE / J9ARRAYCLASS_GET_STRIDE(arrayClass))))
 #endif /* J9VM_ENV_DATA64 */
 			{
 				/* Calculate the size of the object */
-				UDATA dataSize = ((UDATA)size) << scale;
+				UDATA dataSize = ((UDATA)size) * J9ARRAYCLASS_GET_STRIDE(arrayClass);
 				UDATA allocateSize = (dataSize + sizeof(J9IndexableObjectContiguous) + _objectAlignmentInBytes - 1) & ~(UDATA)(_objectAlignmentInBytes - 1);
 				if (allocateSize < J9_GC_MINIMUM_OBJECT_SIZE) {
 					allocateSize = J9_GC_MINIMUM_OBJECT_SIZE;
--- a/runtime/oti/j9.h
+++ b/runtime/oti/j9.h
@@ -325,9 +325,13 @@ static const struct { \
 
 /* Macros for ValueTypes */
 #ifdef J9VM_OPT_VALHALLA_VALUE_TYPES
-#define J9_IS_J9CLASS_VALUETYPE(clazz) J9_ARE_ALL_BITS_SET(clazz->classFlags, J9ClassIsValueType)
+#define J9_IS_J9CLASS_VALUETYPE(clazz) J9_ARE_ALL_BITS_SET((clazz)->classFlags, J9ClassIsValueType)
+#define J9_IS_J9CLASS_FLATTENED(clazz) J9_ARE_ALL_BITS_SET((clazz)->classFlags, J9ClassIsFlattened)
+#define J9_VALUETYPE_FLATTENED_SIZE(clazz)((clazz)->totalInstanceSize - (clazz)->backfillOffset)
 #else /* J9VM_OPT_VALHALLA_VALUE_TYPES */
 #define J9_IS_J9CLASS_VALUETYPE(clazz) FALSE
+#define J9_IS_J9CLASS_FLATTENED(clazz) FALSE
+#define J9_VALUETYPE_FLATTENED_SIZE(clazz)((UDATA) 0) /* It is not possible for this macro to be used since we always check J9_IS_J9CLASS_FLATTENED before ever using it. */
 #endif /* J9VM_OPT_VALHALLA_VALUE_TYPES */
 
 #if defined(OPENJ9_BUILD)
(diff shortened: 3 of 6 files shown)


runtime/vm/createramclass.cpp
```diff
@@ -2947,6 +2947,17 @@ internalCreateRAMClassFromROMClass(J9VMThread *vmThread, J9ClassLoader *classLoa
 				 * which was used for the elementClass.
 				 */
 				module = elementClass->module;
+				if (J9_IS_J9CLASS_VALUETYPE(elementClass) && J9_IS_J9CLASS_FLATTENED(elementClass)) {
+					if (J9_ARE_ALL_BITS_SET(elementClass->classFlags, J9ClassLargestAlignmentConstraintDouble)) {
+						J9ARRAYCLASS_SET_STRIDE(romClass, ROUND_UP_TO(J9_VALUETYPE_FLATTENED_SIZE(elementClass), sizeof(U_64)));
```
tajila: stride should be set on ramClass not romClass

runtime/vm/createramclass.cpp
```diff
@@ -2947,6 +2947,17 @@ internalCreateRAMClassFromROMClass(J9VMThread *vmThread, J9ClassLoader *classLoa
...
+					} else if (J9_ARE_ALL_BITS_SET(elementClass->classFlags, J9ClassLargestAlignmentConstraintReference)) {
+						J9ARRAYCLASS_SET_STRIDE(romClass, ROUND_UP_TO(J9_VALUETYPE_FLATTENED_SIZE(elementClass), sizeof(fj9object_t)));
+					} else { // VT only contains singles (int, short, etc) at this point
```
tajila: comments should be done with `/* ... */`

runtime/gc_glue_java/ArrayletObjectModel.hpp
```diff
@@ -300,16 +300,12 @@ class GC_ArrayletObjectModel : public GC_ArrayletObjectModelBase
...
+	{		
 		J9ROMArrayClass *romArrayClass = (J9ROMArrayClass *)clazzPtr->romClass;
-		UDATA shift = romArrayClass->arrayShape & 0x0000FFFF;
-		UDATA size = numberOfElements << shift;
-		if ((size >> shift) == numberOfElements) {
-			alignedSize = MM_Math::roundToSizeofUDATA(size);
-			if (alignedSize < size) {
-				alignedSize = UDATA_MAX;
-			}
+		UDATA size = (numberOfElements * (J9ARRAYCLASS_GET_STRIDE(arrayClass)));
```
tajila: extra brackets around `J9ARRAYCLASS_GET_STRIDE` are not needed

AidanHa:
Looking at the `J9ARRAYCLASS_SET_STRIDE` and `J9ARRAYCLASS_GET_STRIDE`, I have some comments:

These macros ONLY work if  `J9VM_OPT_VALHALLA_VALUE_TYPES` is defined. This is because `flattenedClassCache` is only declared as a member of the struct if `J9VM_OPT_VALHALLA_VALUE_TYPES` is defined. During compile time, these macros returned an error if I didn't add a `if defined` statement before using them.

 We should add either add a `#if defined(J9VM_OPT_VALHALLA_VALUE_TYPES)` in front of these macros, or declare flattenedClassCache even if flattenedClassCache is not defined.

tajila:
remove the ifdefs around flattenedclassCache

runtime/vm/createramclass.cpp
```diff
@@ -3081,6 +3082,19 @@ internalCreateRAMClassFromROMClass(J9VMThread *vmThread, J9ClassLoader *classLoa
 #endif /* defined(J9VM_OPT_VALHALLA_VALUE_TYPES) */
 
 		result->classFlags = classFlags;
+		if (J9ROMCLASS_IS_ARRAY(romClass)) {
```
tajila: also, its better to put this code with the rest of the array specific code in line 2800

runtime/oti/j9.h
```diff
@@ -325,7 +325,9 @@ static const struct { \
 
 /* Macros for ValueTypes */
 #ifdef J9VM_OPT_VALHALLA_VALUE_TYPES
-#define J9_IS_J9CLASS_VALUETYPE(clazz) J9_ARE_ALL_BITS_SET(clazz->classFlags, J9ClassIsValueType)
+#define J9_IS_J9CLASS_VALUETYPE(clazz) J9_ARE_ALL_BITS_SET((clazz)->classFlags, J9ClassIsValueType)
```
tajila: you'll need to add the macros `#else /* J9VM_OPT_VALHALLA_VALUE_TYPES */` case as well

AidanHa: Would this work @tajila :

```#define J9_IS_J9CLASS_FLATTENED(clazz) FALSE
#define J9_VALUETYPE_FLATTENED_SIZE(clazz)(0)```

Comment (AidanHa) [reply runtime/oti/j9.h]: Not too sure on what to define  `J9_VALUETYPE_FLATTENED_SIZE` as, but maybe we won't even need it if `J9_IS_J9CLASS_FLATTENED(clazz)` is false

File: runtime/vm/createramclass.cpp [hunk 6]
```diff
@@ -2842,6 +2846,17 @@ internalCreateRAMClassFromROMClassImpl(J9VMThread *vmThread, J9ClassLoader *clas
...
+				if (J9_IS_J9CLASS_VALUETYPE(elementClass) && J9_IS_J9CLASS_FLATTENED(elementClass)) {
+					if (J9_ARE_ALL_BITS_SET(elementClass->classFlags, J9ClassLargestAlignmentConstraintDouble)) {
+						J9ARRAYCLASS_SET_STRIDE(ramClass, ROUND_UP_TO(J9_VALUETYPE_FLATTENED_SIZE(elementClass), sizeof(U_64)));
+					} else if (J9_ARE_ALL_BITS_SET(elementClass->classFlags, J9ClassLargestAlignmentConstraintReference)) {
+						J9ARRAYCLASS_SET_STRIDE(ramClass, ROUND_UP_TO(J9_VALUETYPE_FLATTENED_SIZE(elementClass), sizeof(fj9object_t)));
+					} else { /* VT only contains singles (int, short, etc) at this point */
+						J9ARRAYCLASS_SET_STRIDE(ramClass, J9_VALUETYPE_FLATTENED_SIZE(elementClass));
+					}
+				} else {
+					J9ARRAYCLASS_SET_STRIDE(ramClass, (1 << (((J9ROMArrayClass*)romClass)->arrayShape & 0x0000FFFF)));
```
AidanHa: If it's not a value type, then the stride is `(1 << (((J9ROMArrayClass*)romClass)->arrayShape & 0x0000FFFF))`

I ran a test to make sure that the size calculation for non VT elements remained the same

runtime/gc_glue_java/ArrayletObjectModel.hpp
```diff
@@ -300,18 +300,16 @@ class GC_ArrayletObjectModel : public GC_ArrayletObjectModelBase
 	 */
 	MMINLINE UDATA
 	getDataSizeInBytes(J9Class *clazzPtr, UDATA numberOfElements)
-	{
-		UDATA alignedSize = UDATA_MAX;
-		J9ROMArrayClass *romArrayClass = (J9ROMArrayClass *)clazzPtr->romClass;
-		UDATA shift = romArrayClass->arrayShape & 0x0000FFFF;
-		UDATA size = numberOfElements << shift;
-		if ((size >> shift) == numberOfElements) {
```
gacholio: This check has been removed from the new code.

runtime/gc_include/ObjectAllocationAPI.hpp
```diff
@@ -168,14 +168,13 @@ class MM_ObjectAllocationAPI
...
-			UDATA scale = ((J9ROMArrayClass*)(arrayClass->romClass))->arrayShape;
 
 #if !defined(J9VM_ENV_DATA64)
-			if (!sizeCheck || (size < ((U_32)J9_MAXIMUM_INDEXABLE_DATA_SIZE >> scale)))
+			if (!sizeCheck || (size < ((U_32)J9_MAXIMUM_INDEXABLE_DATA_SIZE >> (((J9ROMArrayClass*)(arrayClass->romClass))->arrayShape))))
 #endif /* J9VM_ENV_DATA64 */
 			{
 				/* Calculate the size of the object */
-				UDATA dataSize = ((UDATA)size) << scale;
+				UDATA dataSize = (((UDATA)size) * (J9ARRAYCLASS_GET_STRIDE(arrayClass)));
```
gacholio: stride should be in a local since you read it twice.

runtime/vm/createramclass.cpp
```diff
@@ -32,6 +32,9 @@
 #include "../util/ut_module.h"
 #undef UT_MODULE_LOADED
 #undef UT_MODULE_UNLOADED
+#include "../shared_common/AbstractMemoryPermission.hpp"
+#undef UT_MODULE_LOADED
+#undef UT_MODULE_UNLOADED
```
gacholio: What is this for?

AidanHa: This is so I could use the `ROUND_UP_TO` macro. Is there a better way for me to access this macro from `AbstractMemoryPermission.hpp`

runtime/gc_glue_java/ArrayletObjectModel.hpp
AidanHa: Yea, I was wondering when will this check ever return false...

runtime/vm/createramclass.cpp
tajila: Use the round up to macro in j9.h (`ROUND_UP_TO_POWEROF2`). That file is already included so you dont need to add any includes.

gacholio: Agreed.

runtime/gc_include/ObjectAllocationAPI.hpp
gacholio: Not sure what I was seeing here.

runtime/gc_glue_java/ArrayletObjectModel.hpp
gacholio: Ask @dmitripivkine - it may have something to do with overflow detection.

dmitripivkine: This is overflow detection on 32-bit platforms

AidanHa: Will it still need to be kept? If so, would this be a good substitution? 

```
	MMINLINE UDATA
	getDataSizeInBytes(J9Class *clazzPtr, UDATA numberOfElements)
	{	
		UDATA stride = J9ARRAYCLASS_GET_STRIDE(clazzPtr);
		UDATA size = numberOfElements * stride;
		UDATA alignedSize = UDATA_MAX;
		if ((size/stride) == numberOfElements) {
			alignedSize = MM_Math::roundToSizeofUDATA(size);
			if (alignedSize < size) {
				alignedSize = UDATA_MAX;
			}
		}
		return alignedSize;

	}
```

dmitripivkine: Logically this is code should work. However my concern here is an introduction of real division operation might be expensive some platforms. There is no way for compiler to replace it to shift any kind.
@charliegracie Any thoughts?

gacholio: To check for overflow, simply do:

orig = val
val = val * size
if (val < orig) overflow

dmitripivkine: It would not work for all cases. For example:
orig = 0x7FFFFFFF
size = 0x18
val = (orig * size) = 0x(B)FFFFFFE8

val is larger then orig but this is overflow

dmitripivkine: I see a few solutions but all of them not better then using division

AidanHa: I've decided to stick with using division

runtime/gc_include/ObjectAllocationAPI.hpp
tajila: brackets are not needed around `J9ARRAYCLASS_GET_STRIDE(arrayClass)`

runtime/oti/j9.h
tajila: change J9_VALUETYPE_FLATTENED_SIZE(clazz) to `((UDATA) -1)`.

tajila:
Changes LGTM

@gacholio and @dmitripivkine Please take another look at this PR.

runtime/gc_glue_java/ArrayletObjectModel.hpp
```diff
@@ -300,12 +300,11 @@ class GC_ArrayletObjectModel : public GC_ArrayletObjectModelBase
...
-	{
+	{	
+		UDATA stride = J9ARRAYCLASS_GET_STRIDE(clazzPtr);
+		UDATA size = numberOfElements * stride;
 		UDATA alignedSize = UDATA_MAX;
-		J9ROMArrayClass *romArrayClass = (J9ROMArrayClass *)clazzPtr->romClass;
-		UDATA shift = romArrayClass->arrayShape & 0x0000FFFF;
-		UDATA size = numberOfElements << shift;
-		if ((size >> shift) == numberOfElements) {
+		if ((size/stride) == numberOfElements) {
```
dmitripivkine: please add spaces around `/` `(size / stride)`

runtime/gc_include/ObjectAllocationAPI.hpp
```diff
@@ -168,14 +168,13 @@ class MM_ObjectAllocationAPI
 		if (0 != size) {
 			/* Contiguous Array */
 
-			UDATA scale = ((J9ROMArrayClass*)(arrayClass->romClass))->arrayShape;
 
 #if !defined(J9VM_ENV_DATA64)
-			if (!sizeCheck || (size < ((U_32)J9_MAXIMUM_INDEXABLE_DATA_SIZE >> scale)))
```
tajila: Dmitri pointed out that this size check must be consistent with the size check below. 

We can not rely on the scale as that will not work for valuetypes

AidanHa: Would the following make sense? 

```((U_32)J9_MAXIMUM_INDEXABLE_DATA_SIZE / J9ARRAYCLASS_GET_STRIDE(arrayClass)))```

**Comment by AidanHa** [comment]
@dmitripivkine @tajila I've pushed changes to address your comments. Currently running some tests to ensure the changes compile properly on wi and xa

File: runtime/gc_include/ObjectAllocationAPI.hpp [hunk 12]
```diff
@@ -168,14 +168,13 @@ class MM_ObjectAllocationAPI
...
+			if (!sizeCheck || (size < ((U_32)J9_MAXIMUM_INDEXABLE_DATA_SIZE / J9ARRAYCLASS_GET_STRIDE(arrayClass))))
 #endif /* J9VM_ENV_DATA64 */
 			{
 				/* Calculate the size of the object */
-				UDATA dataSize = ((UDATA)size) << scale;
+				UDATA dataSize = (((UDATA)size) * J9ARRAYCLASS_GET_STRIDE(arrayClass));
```
dmitripivkine: outside brackets are not necessary

dmitripivkine:
there are a lot of places where `arrayShape & 0x0000FFFF` is used for array element size detection. If I understand correctly all these places must be changed as well

AidanHa:
@dmitripivkine it feels like that change might be out of the scope of this PR. Should I be putting that change into another Issue/PR? ATM, `J9ARRAYCLASS_SET_STRIDE` for non VT elements is set to `arrayShape & 0x0000FFFF` anyways, so it might be better to make another PR to fix it

dmitripivkine:
There are three other places in `ArrayletObjectModel.hpp` uses `arrayShape & 0x0000FFFF` except modified `getDataSizeInBytes()`. I understand you concern that change might be speed wide but in this case we need to be consistent.

AidanHa:
Git it, it that case, I'll also be changing any other instances where `arrayShape & 0x0000FFFF` is used for element size detection (outside of `ArrayletObjectModel.hpp`)

runtime/gc_glue_java/ArrayletObjectModel.hpp
```diff
@@ -402,14 +401,13 @@ class GC_ArrayletObjectModel : public GC_ArrayletObjectModelBase
...
 	memcpyFromArray(void *destData, J9IndexableObject *srcObject, I_32 elementIndex, I_32 elementCount)
 	{
-		J9ROMArrayClass *romArrayClass = (J9ROMArrayClass*)J9GC_J9OBJECT_CLAZZ(srcObject)->romClass;
-		UDATA elementScale = romArrayClass->arrayShape & 0x0000FFFF;
+		UDATA elementSize = J9ARRAYCLASS_GET_STRIDE(J9GC_J9OBJECT_CLAZZ(srcObject));
 		if (isInlineContiguousArraylet(srcObject)) {
 			// If the data is stored contiguously, then a simple copy is sufficient.
 			void* srcData = getDataPointerForContiguous(srcObject);
-			switch(elementScale)
+			switch(elementSize)
```
tajila: @dmitripivkine I dont think this is correct, A valuetype containing a single object reference will be treated like an int/float

dmitripivkine: you should add `default:` case to all these switches to handle array sizes other then 1,2,4,8. You can add your own implementation (for example using `memcpy()`) or modify `case 1:`  to do proper byte by byte copy:
```
			case 1:
                        default:
					// byte/boolean or flattened type
					{
						U_8* srcCursor = ((U_8*)srcData) + elementIndex * elementSize;
					U_8* destCursor = (U_8*)destData;
					I_32 count = elementCount  * elementSize;
					while(count--) {
						*destCursor++ = *srcCursor++;
					}
					}
					break;
```

dmitripivkine:
@tajila BTW there is a potential problem with copy of flattened array elements. If we are not going inside of element structure at the time of copy we should guarantee that object reference slots (must, crash otherwise) and primitive slots (if not data might be corrupted) are copied in atomic units. For example object pointer must be copied by 8 bytes copy instruction for non-CR 64 and at least 4 bytes copy instruction for CR 64 bit and 32 bit. Otherwise (for example in case of copy by bytes) object pointer might be read at the time is partially updated (like half old half new) which is fatal. Similarly primitive data types read might return wrong result. The solution might be always copy by largest chinks to make copy atomic for each element. However this is tricky from alignment point of view. 
I am ok with byte copy for this implementation but it should be reviewed in the future

gacholio:
Atomicity of access is guaranteed by certain APIs (particularly if the field is volatile). I'm not sure what volatile means (if it's even legal) on flattened instances. Anything not declared volatile is not guaranteed atomic updates, afaik (@DanHeidinga can you confirm/deny?).

Given that VTs are immutable, I'm not sure this is an actual concern.

DanHeidinga:
The simplest implementation for a volatile flattenable field is to refuse to flatten it.  (Note, flattenable is not a guarantee that it will be flattened)

We can leave all volatile fields as pointers so that volatile operations are straightforward.  This handles the use site properties (ie: field definitions) and there will likely be a way to mark an entire value type as volatile.

Once we have the simple semantics for volatile values working, we can look at optimizing any the indirection for any volatile values that will fit into a single slot.

gacholio:
I can't recall how the flattening works, if I flatten:

class DoubleDouble {
   double x;
   double y;
} 

into:

class Whatever {
   double z;
   int x;
   DoubleDouble dd;
}

is dd in it's class order (i.e. is `getfield dd` a memcpy of dd into the new object, or are the fields ordered as if they were declared locally, so the 3 double would sort together)?

gacholio:
If the flattening is such that the embedded object is in order, I'm not sure atomicity can be guaranteed, unless we align all flat objects to the maximum alignment (8 today). If the flattened fields are scattered across the object, reconstituting it will be unreasonably difficult/expensive.

DanHeidinga:
> (i.e. is getfield dd a memcpy of dd into the new object,

This.  We keep the same shape for DoubleDouble whether its flattened into something else or not.

> I'm not sure atomicity can be guaranteed, unless we align all flat objects to the maximum alignment (8 today)

If you want atomicity, declare it volatile and live with the indirection (for now)

gacholio:
Not sure that's reasonable - anyone would expect atomic ops to work on a field of the appropriate size, regardless of how it's embedded - think migration - if atomic ops fail some very small percentage of the time, that would be nearly impossible to track down.

gacholio:
Since VTs are immuteable, this matters a lot less. As @DanHeidinga says, any volatile field needs to be aligned, which implies that all embedded objects must be aligned. I assume volatile fields are allowed within VTs.

Since volatile requires alignment, I believe this means that flattened objects (with fields that require alignment) must be aligned appropriately at the start (i.e. alignemnt must be propagated outwards).

gacholio:
Alternatively, any VT with a volatile field (does that even mean anything in a read-only object?) is simply ineligible for flattening.

runtime/gc_glue_java/ArrayletObjectModel.hpp
AidanHa: Will there ever be a case where to elementsize is not 1, 2, 4, 8?
References have size 4, and the biggest primitive is a double which is byte size 8...

AidanHa: @dmitripivkine I've also added default cases for the discontiguous case of `memcpyFromArray` as well as `memcpyToArray` an its discontiguous case. Could you review my changes to see if I've done this correctly?

runtime/gc_glue_java/ArrayletObjectModel.hpp
```diff
@@ -530,6 +540,20 @@ class GC_ArrayletObjectModel : public GC_ArrayletObjectModelBase
...
 					}
 					break;
+
+				default :
+					// array size other than 1/2/4/8 (mainly for flattened types)
+					{
+						U_8* srcCursor = ((U_8*)srcLeafAddress) + arrayletElementOffset * elementSize;
+						U_8* destCursor = (U_8*)outerDestCursor;
+						outerCount -= innerCount;	// Decrement the outer count before looping
+						while(innerCount--) {
```
dmitripivkine: we need double check. I believe this is not correct. `innerCount` counts in elements but you need in bytes, should be `innerCount * elementSize`

runtime/gc_glue_java/ArrayletObjectModel.hpp
```diff
@@ -530,6 +540,20 @@ class GC_ArrayletObjectModel : public GC_ArrayletObjectModelBase
 						arrayletElementOffset = 0;
 					}
 					break;
+
+				default :
+					// array size other than 1/2/4/8 (mainly for flattened types)
+					{
+						U_8* srcCursor = ((U_8*)srcLeafAddress) + arrayletElementOffset * elementSize;
```
tajila: Currently, these API's are only used for primitive array copies (i.e. jniGetPrimitiveArrayCritical) so the only allowable sizes are 1,2,4,8 bytes.

There is no spec on the JNI behaviour for valuetypes so right now dont know if we will even need a VT copy array helper for JNI. However, we might need one for Unsafe. 

In any case, VTs will need special handling for array copies since the elements have varying alignment constraints and may or may not contain references. For VT arrays we should create a separate helper that is optimized to recognize the cases where it can do a fast copy and when it must look at the instance description and do slot sized copies with the required barriers.

For now we should simply ensure that the primitive array copies behave correctly with the new stride calculation.

The default case in these APIs should assertUnimplemented (or assertUnreachable) since we currently do not expect anything other than primitive arrays to be using them.

tajila:
We can also take a look at the VM primitve array copy helpers in `ArrayCopyHelpers.hpp` and `alignedmemcpy.c`. These helpers currently use arrayShape and can be modified to use the stride instead. 

There are quite a few of them so I suggest that this should be done in a separate PR.

AidanHa:
>For now we should simply ensure that the primitive array copy APIs behave correctly with the new stride calculation.

Okay then the only thing I will do in this PR regarding array copying is implement the `J9ARRAYCLASS_GET_STRIDE` macro, change the calculations of `elementSize`, and add a default case to Assert `MM_unreachable`.

runtime/gc_glue_java/ArrayletObjectModel.hpp
```diff
@@ -456,12 +454,19 @@ class GC_ArrayletObjectModel : public GC_ArrayletObjectModelBase
 					}
 				}
 				break;
+				
+			default :
+				// unreachable since we currently do not expect anything other than primitive arrays to be using them.
+				{
+					Assert_MM_unreachable();
```
dmitripivkine: To make this correct you need to add `#include "ModronAssertions.h"` at the beginning of the file

AidanHa: Yea, I realized that when I was testing my changes. I'm currently rebuilding my changes to see if i missed anything else

runtime/gc_glue_java/ArrayletObjectModel.hpp
```diff
@@ -33,6 +33,9 @@
 #include "j9consts.h"
 #include "modron.h"
 #include "modronopt.h"
+#include "ModronAssertions.h"
+#undef UT_MODULE_LOADED
+#undef UT_MODULE_UNLOADED
```
dmitripivkine: Why you need this two lines?

AidanHa: I tried building without those 2 lines, but I get an error saying how `UT_MODULE_LOADED` and 
`UT_MODULE_UNLOADED` are already defined.

This meant that in `ArrayletObjectModelBase.hpp`, `Bits.hpp`, `Math,hpp` or `SlotObject.hpp`, `UT_MODULE_LOADED` and `UT_MODULE_UNLOADED` are already defined.

dmitripivkine: ok, remove all three lines and rely on proper includes in higher level

gacholio: Sounds like a bug in a header file, possibly a missing include gate.

dmitripivkine: yes this is old problem with `#include "ut_j9mm.h"`

runtime/gc_glue_java/ArrayletObjectModel.hpp
```diff
@@ -852,6 +876,11 @@ class GC_ArrayletObjectModel : public GC_ArrayletObjectModelBase
 	 * Tear down the receiver
 	 */
 	void tearDown(MM_GCExtensionsBase *extensions);
+
+	/**
+	 * Assert MM_unreachable if the element size is invalid
+	 */
+	void AssertBadElementSize();
```
AidanHa: Any name recommendations for this function?

dmitripivkine: Move function to the private section (close to line 56). Also comment is not relevant

dmitripivkine:
Jenkins compile win

dmitripivkine:

runtime/oti/j9.h
gacholio: UDATA_MAX might be better than the cast value.

gacholio: I see the value is still 0 - would it be better to simply not define the macro if VTs are not in the compile? The "is flat" macro can stay, but the size macro should go.

runtime/oti/j9.h
```diff
@@ -325,9 +325,13 @@ static const struct { \
...
+#define J9_IS_J9CLASS_FLATTENED(clazz) J9_ARE_ALL_BITS_SET((clazz)->classFlags, J9ClassIsFlattened)
+#define J9_VALUETYPE_FLATTENED_SIZE(clazz)((clazz)->totalInstanceSize - (clazz)->backfillOffset)
 #else /* J9VM_OPT_VALHALLA_VALUE_TYPES */
 #define J9_IS_J9CLASS_VALUETYPE(clazz) FALSE
+#define J9_IS_J9CLASS_FLATTENED(clazz) FALSE
+#define J9_VALUETYPE_FLATTENED_SIZE(clazz)((UDATA) 0) /* It is not possible for this macro to be used since we always check J9_IS_J9CLASS_FLATTENED before ever using it. */
```
gacholio: Please remove this line (and ensure there are no unguarded callers).

runtime/oti/j9.h
AidanHa: I think I had a compile error when the size macro was not defined... I'll try to reproduce it and paste the error here

gacholio:

pshipton:
This is failing to compile on Windows. Needs to be fixed asap or backed out. 
@AidanHa @gacholio 
```
15:28:01  createramclass.cpp(2860): error C2220: warning treated as error - no 'object' file generated
15:28:01  createramclass.cpp(2860): warning C4334: '<<': result of 32-bit shift implicitly converted to 64 bits (was 64-bit shift intended?)
```

AidanHa:
On it

AidanHa:
Strange since the previous Windows build passed...

pshipton:
Here is the failure link I should have provided https://ci.eclipse.org/openj9/job/Build-JDK11-win_x86-64_cmprssptrs/739/

PR #5166, opened 2019-03-19
Add IgnoreDecimalOverflowException codegen flag

fjeremic:
This codegen flag determines whether decimal overflow or fixed point
overflow checks should be generated for instructions which support such
by-passes.

### DIFF
--- a/runtime/compiler/z/codegen/J9CodeGenerator.cpp
+++ b/runtime/compiler/z/codegen/J9CodeGenerator.cpp
@@ -178,6 +178,8 @@ J9::Z::CodeGenerator::CodeGenerator() :
       {
       self()->setHasFixedFrameC_CallingConvention();
       }
+
+   cg->setIgnoreDecimalOverflowException(false);
    }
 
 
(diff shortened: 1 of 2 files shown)


fjeremic:

runtime/compiler/z/codegen/J9CodeGenerator.cpp
```diff
@@ -178,6 +178,8 @@ J9::Z::CodeGenerator::CodeGenerator() :
       {
       self()->setHasFixedFrameC_CallingConvention();
       }
+
+   setIgnoreDecimalOverflowException(false);
```
0xdaryl: Missing `self()->`

runtime/compiler/z/codegen/J9CodeGenerator.hpp
```diff
@@ -300,6 +300,40 @@ class OMR_EXTENSIBLE CodeGenerator : public J9::CodeGenerator
...
+      return ignoreDecimalOverflowException;
+      }
+
+   private:
+
+   /** \brief
+    *     Determines whether decimal overflow or fixed point overflow checks should be generated for instructions which
+    *     support such by-passes.
+    */
+   bool ignoreDecimalOverflowException;
```
0xdaryl: Other classes use a preceding `_` for private fields (though arguably it isn't super consistent throughout the code base).  Just pointing it out though.

fjeremic: Fixed in cbc0737

runtime/compiler/z/codegen/J9CodeGenerator.cpp
fjeremic: Fixed in cbc0737

fjeremic:

Issue #5147, opened 2019-03-18
Crash in relocateThunks in JITaaS remote AOT mode


harryyu1994:
```
#0: function TR_LinuxCallStackIterator::printStackBacktrace(TR::Compilation*)+0x45 [0x7fdc2b76609b]
#1: function TR_Debug::printStackBacktrace()+0x2f [0x7fdc2b778e6f]
#2: /root/Harry_JITaaS/sdk/jre/lib/amd64/compressedrefs/libj9jit29.so(+0x9cab2f) [0x7fdc2ae0eb2f]
#3: /root/Harry_JITaaS/sdk/jre/lib/amd64/compressedrefs/libj9prt29.so(+0x2116f) [0x7fdc32fdb16f]
#4: /lib/x86_64-linux-gnu/libpthread.so.0(+0x12890) [0x7fdc35626890]
#5: function TR::CompilationInfoPerThread::relocateThunks()+0xfe [0x7fdc2ae0fe9a]
#6: function remoteCompile(J9VMThread*, TR::Compilation*, TR_ResolvedMethod*, J9Method*, TR::IlGeneratorMethodDetails&, TR::CompilationInfoPerThreadBase*)+0xc00 [0x7fdc2b073c0b]
#7: function TR::CompilationInfoPerThreadBase::compile(J9VMThread*, TR::Compilation*, TR_ResolvedMethod*, TR_J9VMBase&, TR_OptimizationPlan*, TR::SegmentAllocator const&)+0xe8e [0x7fdc2ae284b0]
#8: function TR::CompilationInfoPerThreadBase::wrappedCompile(J9PortLibrary*, void*)+0x26ca [0x7fdc2ae266f4]
#9: /root/Harry_JITaaS/sdk/jre/lib/amd64/compressedrefs/libj9prt29.so(+0x21ee8) [0x7fdc32fdbee8]
#10: function TR::CompilationInfoPerThreadBase::compile(J9VMThread*, TR_MethodToBeCompiled*, J9::J9SegmentProvider&)+0x70d [0x7fdc2ae23bb5]
#11: function TR::CompilationInfoPerThread::processEntry(TR_MethodToBeCompiled&, J9::J9SegmentProvider&)+0x4fd [0x7fdc2ae1a373]
#12: function TR::CompilationInfoPerThread::processEntries()+0x247 [0x7fdc2ae19637]
#13: function TR::CompilationInfoPerThread::run()+0x60 [0x7fdc2ae18ed8]
[...]

PR #5167, opened 2019-03-19
Persist thunks to SCC for delayed AOT load in JITaaS Remote AOT mode

harryyu1994:
Current code assumes that AOT relocation always happens right after AOT compilation and always relocates thunks, this change adds the appropriate handle for delayed AOT load in JITaaS Remote AOT mode.

[skip ci]
closes: #5147

### DIFF
--- a/runtime/compiler/control/CompilationThread.cpp
+++ b/runtime/compiler/control/CompilationThread.cpp
@@ -698,6 +698,8 @@ j9ThunkInvokeExactHelperFromTerseSignature(void * jitConfig, UDATA signatureLeng
 void
 TR::CompilationInfoPerThread::relocateThunks()
    {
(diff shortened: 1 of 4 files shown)


runtime/compiler/control/CompilationThread.cpp
```diff
@@ -698,6 +698,7 @@ j9ThunkInvokeExactHelperFromTerseSignature(void * jitConfig, UDATA signatureLeng
 void
 TR::CompilationInfoPerThread::relocateThunks()
    {
+   // only used by JITaaS
```
mpirvu: Instead of a comment let's put an assert

ymanton:
Looks OK to me as well, aside from the asserts.

harryyu1994:
Ready for review and merge. @mpirvu

PR #5168, opened 2019-03-19
caching the attributes for AOT

satbir-ca:
Caching the attributes and checking the cache
before sending the message to the client.

removing warning due to rememberDebugCounterName.

### DIFF
--- a/runtime/compiler/env/j9methodServer.hpp
+++ b/runtime/compiler/env/j9methodServer.hpp
@@ -29,7 +29,7 @@
 #include "runtime/JITaaSIProfiler.hpp"
 
 struct
-TR_J9MethodFieldAttributes
+TR_J9MethodResolvedFieldAttributes
    {
    // for field attributes, use fieldOffset,
    // for static attributes, use address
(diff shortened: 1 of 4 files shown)


satbir-ca:
@ymanton Please review the code

runtime/compiler/env/j9methodServer.hpp
```diff
@@ -45,6 +45,27 @@ TR_J9MethodFieldAttributes
...
+TR_J9MethodResolvedRelocatableFieldAttributes
+   {
+   // for field attributes, use fieldOffset,
+   // for static attributes, use address
+   union
+      {
+      IDATA fieldOffset;
+      void *address;
+      };
+   bool unresolvedInCP;
```
mpirvu: Please move `unresolvedInCP` after `ltype`, otherwise the compiler will insert padding and increase the size of the struct unnecessarily.

mpirvu:
```
Satbir Singh (satbir.singh1@****.com) is not covered by the necessary legal agreements in order to proceed.
```

@satbir-ca You many need to resign ECA. I had a similar request in my email box.

Also, since we can now run with AOT you should test the new code.

satbir-ca:
updated and tested with AOT.

satbir-ca:
ECA also done

ymanton:
ECA check is still red. Can you rebase your change and force-push?

satbir-ca:
updated

PR #60, opened 2017-09-18
Add new cpEntry for modified anonClassName

tajila:
Add new cpEntry for modified anonClassName

Currently, when an anonClass is created, the entry in the constantPool
that contains the className utf is modified and is allocated a larger
buffer for the name (which re-written during ramClass creation -
className/addressOfRomClass). A problem arises when there are multiple
references to the utf8 (ie. Constant_Class ref and a Constant_String
ref) in the constantPool. If an anonClass happens to have a string
literal that is the same as its class name, this string literal becomes
corrupted because there is only one shared reference to the utf in the
constantPool. 

This patch creates a new constantPool entry for the className if we are
creating an anonClass and it is referenced by both a Class ref and a
String or nameAndType ref. This way all other references point to the
original utf8 but the classref points to the modified utf entry. If the
constantPool size is at U_16_MAX then we revert to doing what we are
currently doing.

Signed-off-by: Tobi Ajila <tobi_ajila@ca.ibm.com>

Reviewer: @DanHeidinga

PR #5169, opened 2019-03-19
Correct the field order for J9DbgROMClassBuilder

keithc-ca:
the inconsistency was introduced when fields were added in #60


pshipton:
Why does the order matter? Does this fix something?

DanHeidinga:
It appears to be in relation to the comment here: https://github.com/eclipse/openj9/blob/master/runtime/bcutil/ROMClassBuilder.hpp#L112

but I'm not clear on how this would surface.  DDR issue?

DanHeidinga:
Jenkins test all zlinux jdk8

keithc-ca:
DDR uses `J9DbgROMClassBuilder` rather than `ROMClassBuilder` (perhaps due to a limitation of legacy tooling). The VM actually uses they layout of `ROMClassBuilder`: this just makes them match.

Issue #5124, opened 2019-03-15
JITaaS should not cache array class of RAM class on the server


dmitry-ten:
Whenever we cache a ROM class on the server, we also cache some of its RAM attributes in `ClassInfo` struct, including `ClassInfo.arrayClass`. That is, we cache a pointer to an array class that has elements of the class being cached. 
When we implemented caching we assumed the array class will not change while a class is loaded, however, that is not true.
For example, it is changed in this method in `runtime/vm/createramclass.cpp`
```
internalCreateRAMClassDone(J9VMThread *vmThread, J9ClassLoader *classLoader, J9ROMClass *romClass,
        UDATA options, J9Class *elementClass, J9UTF8 *className, J9CreateRAMClassState *state) {
    ...
    if (J9ROMCLASS_IS_ARRAY(romClass)) {
        ((J9ArrayClass *)elementClass)->arrayClass = state->ramClass;
        ...
     }
    ...
 }
```
From what I can tell, this means when an array class gets loaded, `arrayClass` pointer of its component class is set to the newly loaded class. I'm not sure what happens when multiple array classes have the same component type, but that doesn't really matter to us.  What matters is that if we previously cached NULL for `arrayClass`, we'll now have invalid information on the server.
I encountered this bug while working on enabling SVM for remote AOT, but it's obviously present in a regular JITaaS as well.
The simplest solution would be to stop caching array classes but that would increase CPU consumption, so maybe we can think of something better.

PR #5172, opened 2019-03-20
Do not cache array class of RAM class

dmitry-ten:
Array class of RAM class can be initialized
    when the class is already loaded, so it is not
    safe to cache NULL as array class.
    Only cache non-NULL values, make remote call otherwise.
    If a remote call returns non-NULL, it means that array
    class has been initiailized. It should not change anymore
    so it's safe to cache it.

Fixes: #5124

### DIFF
--- a/runtime/compiler/env/VMJ9Server.cpp
+++ b/runtime/compiler/env/VMJ9Server.cpp
@@ -781,9 +781,24 @@ TR_J9ServerVM::canAllocateInlineClass(TR_OpaqueClassBlock *clazz)
 TR_OpaqueClassBlock *
 TR_J9ServerVM::getArrayClassFromComponentClass(TR_OpaqueClassBlock *componentClass)
    {
-   TR_OpaqueClassBlock *arrayClass = nullptr;
    JITaaS::J9ServerStream *stream = _compInfoPT->getMethodBeingCompiled()->_stream;
+   TR_OpaqueClassBlock *arrayClass = NULL;
    JITaaSHelpers::getAndCacheRAMClassInfo((J9Class *)componentClass, _compInfoPT->getClientData(), stream, JITaaSHelpers::CLASSINFO_ARRAY_CLASS, (void *)&arrayClass);
+   if (!arrayClass)
+      {
+      stream->write(JITaaS::J9ServerMessageType::VM_getArrayClassFromComponentClass, componentClass);
+      arrayClass = std::get<0>(stream->read<TR_OpaqueClassBlock *>());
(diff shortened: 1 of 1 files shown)


ymanton:
> @mpirvu @ymanton do you think there is a way we can at least partially cache array class?
> If not, I'll also have to remove the entry for array class from `ClassInfo` struct and all related code.

We have various class loading hooks in the JIT that get called when classes are loaded; I assume the hooks would be called when an array class gets created. We could keep track of changing array classes and send deltas to the server on each compilation, but I don't know if it's really worth it.

It's a little surprising to me that the array class of a class can change. Are you sure that the `internalCreateRAMClassDone` code you pointed out in #5124 can overwriting an existing array class, or is it simply initializing it? Can you figure out under which conditions this happens? Then maybe we can figure out how important it is to optimize.

dmitry-ten:
I think you are right.
When we initialize a new array class, `internalCreateRAMClassDone` is the last function called in a long call stack that begins in `ByteCodeInterpreter.hpp` with call to `internalCreateArrayClass`, from functions such as  `inlArrayNewArrayIMpl`, `anewarray`, `multianewarray` etc.
Possible callstack:
```
# internalCreateRAMClassDone
# ...
# <more functions>
# ...
# internalCreateArrayClass
# anewArray
```
In all of these functions, we first check if `componentClass->arrayClass` is NULL, and only if it is NULL, we create a new array class, which results in `componentClass->arrayClass` pointer being set.
Otherwise, we use an already existing array class.
Example:
```
        VMINLINE VM_BytecodeAction
        inlClassArrayTypeImpl(REGISTER_ARGS_LIST)
        {
                J9Class *componentClazz = J9VM_J9CLASS_FROM_HEAPCLASS(_currentThread, *(j9object_t*)_sp);
                J9Class *arrayClazz = componentClazz->arrayClass;
                if (NULL == arrayClazz) {
                        arrayClazz = internalCreateArrayClass(_currentThread,
                                (J9ROMArrayClass *) J9ROMIMAGEHEADER_FIRSTCLASS(_currentThread->javaVM->arrayROMClasses),
                                componentClazz);
                }
                if (NULL == arrayClazz) {
                } else {
                        returnObjectFromINL(REGISTER_ARGS, J9VM_J9CLASS_TO_HEAPCLASS(arrayClazz), 1);
                }
                return rc;
        }
```

dmitry-ten:
So in this case, the fix is to only cache non-NULL array classes.

dmitry-ten:
Updated.

PR #5176, opened 2019-03-20
Dynamic checkcast evaluation on x86

yanluo7:
Currently on X checkcast evaluator only supports inline evaluation for
constant cast classes. If the cast class is not known at compile time a VM
helper call is generated. With upcoming optimizer changes where certain
instanceOf nodes are transformed into checkcast with dynamic cast class,
there is a need to support the inline evaluation for these new cases. This
PR provides such support. It inlines the checkcast instructions needed for
interface and class case; for array case, a VM helper call is generated,
but this is the same as in constant cast class case since array cases are
believed to be uncommon. As the cast class is dynamic, runtime tests need
to be generated to tell if the cast class is array, interface, or class.
For interface, a tight loop through the itable is generated; for class,
the necessary equality test and superclass test are generated.
Consideration was made to add some form of cache slot similar to the
constant case; but given that the cast class is dynamic, 1) instead of
just 1 slot, we would need 2 slots to match both the cast class and the
cached value in order to produce a hit, and 2) given the dynamic nature
where the cast class will potentially vary every time the checkcast is
executed, the hit rate for the cached pair are likely low. So this as an
initial support only inlines the necessary tests without the cached slots,
which can be added as a followup change in the future.

### DIFF
--- a/runtime/compiler/x/codegen/J9TreeEvaluator.cpp
+++ b/runtime/compiler/x/codegen/J9TreeEvaluator.cpp
@@ -3664,6 +3664,148 @@ TR::Register *J9::X86::TreeEvaluator::longBitCount(TR::Node *node, TR::CodeGener
    return resultReg;
    }
 
+inline void generateInlinedCheckCastForDynamicCastClass(TR::Node* node, TR::CodeGenerator* cg)
+   {
+   auto use64BitClasses = TR::Compiler->target.is64Bit() &&
+               (!TR::Compiler->om.generateCompressedObjectHeaders() ||
+               (cg->comp()->compileRelocatableCode() && cg->comp()->getOption(TR_UseSymbolValidationManager)));
+   TR::Register *ObjReg = cg->evaluate(node->getFirstChild());
+   TR::Register *castClassReg = cg->evaluate(node->getSecondChild());
+   TR::Register *temp1Reg = cg->allocateRegister();
+   TR::Register *temp2Reg = cg->allocateRegister();
+   TR::Register *objClassReg = cg->allocateRegister();
+  
+   bool isCheckCastAndNullCheck = (node->getOpCodeValue() == TR::checkcastAndNULLCHK);
+
+   TR::LabelSymbol *startLabel = generateLabelSymbol(cg);
+   TR::LabelSymbol *fallThruLabel = generateLabelSymbol(cg);
+   TR::LabelSymbol *outlinedCallLabel = generateLabelSymbol(cg);
+   TR::LabelSymbol *throwLabel = generateLabelSymbol(cg);
+   TR::LabelSymbol *isClassLabel = generateLabelSymbol(cg);
+   TR::LabelSymbol *iTableLoopLabel = generateLabelSymbol(cg);
+   startLabel->setStartInternalControlFlow();
+   fallThruLabel->setEndInternalControlFlow();
+   
+   generateLabelInstruction(LABEL, node, startLabel, cg);
+   
+   TR_OutlinedInstructions *outlinedHelperCall = new (cg->trHeapMemory()) TR_OutlinedInstructions(node, TR::call, NULL, outlinedCallLabel, fallThruLabel, cg);
+   cg->getOutlinedInstructionsList().push_front(outlinedHelperCall);
+
+   // objClassReg holds object class also serves as null check
+   if (isCheckCastAndNullCheck)
+      generateLoadJ9Class(node, objClassReg, ObjReg, cg);
+
+   // temp2Reg holds romClass of cast class, for testing array, interface class type
+   generateRegMemInstruction(LRegMem(), node, temp2Reg, generateX86MemoryReference(castClassReg, offsetof(J9Class, romClass), cg), cg);
+   
+   // If cast class is array, call out of line helper
+   generateMemImmInstruction(TEST4MemImm4, node,
+       generateX86MemoryReference(temp2Reg, offsetof(J9ROMClass, modifiers), cg), J9AccClassArray, cg);
+   generateLabelInstruction(JNE4, node, outlinedCallLabel, cg);
+
+   // objClassReg holds object class  
+   if (!isCheckCastAndNullCheck)
+      {
+      generateRegRegInstruction(TESTRegReg(), node, ObjReg, ObjReg, cg);
+      generateLabelInstruction(JE4, node, fallThruLabel, cg);
+      generateLoadJ9Class(node, objClassReg, ObjReg, cg);
+      }
+
+   // Object not array, inline checks
+   // Check cast class is interface
+   generateMemImmInstruction(TEST4MemImm4, node,
+       generateX86MemoryReference(temp2Reg, offsetof(J9ROMClass, modifiers), cg), J9AccInterface, cg);
+   generateLabelInstruction(JE4, node, isClassLabel, cg);
+   
+   // Obtain I-Table
+   // temp1Reg holds head of J9Class->iTable of obj class
+   generateRegMemInstruction(LRegMem(), node, temp1Reg, generateX86MemoryReference(objClassReg, offsetof(J9Class, iTable), cg), cg);
+   // Loop through I-Table
+   // temp1Reg holds iTable list element through the loop
+   generateLabelInstruction(LABEL, node, iTableLoopLabel, cg);
+   generateRegRegInstruction(TESTRegReg(), node, temp1Reg, temp1Reg, cg);
+   generateLabelInstruction(JE4, node, throwLabel, cg);
+   auto interfaceMR = generateX86MemoryReference(temp1Reg, offsetof(J9ITable, interfaceClass), cg);
+   generateMemRegInstruction(CMPMemReg(), node, interfaceMR, castClassReg, cg);
+   generateRegMemInstruction(LRegMem(), node, temp1Reg, generateX86MemoryReference(temp1Reg, offsetof(J9ITable, next), cg), cg);
+   generateLabelInstruction(JNE4, node, iTableLoopLabel, cg);
+   
+   // Found from I-Table
+   generateLabelInstruction(JMP4, node, fallThruLabel, cg);
+   
+   // cast class is non-interface class
+   generateLabelInstruction(LABEL, node, isClassLabel, cg);
+   // equality test
+   generateRegRegInstruction(CMPRegReg(use64BitClasses), node, objClassReg, castClassReg, cg);
+   generateLabelInstruction(JE4, node, fallThruLabel, cg);
+   
+   // class not equal
+   // temp2 holds cast class depth
+   // class depth mask must be low 16 bits to safely load without the mask.
+   static_assert(J9AccClassDepthMask == 0xffff, "J9_JAVA_CLASS_DEPTH_MASK must be 0xffff");
+   generateRegMemInstruction(TR::Compiler->target.is64Bit()? MOVZXReg8Mem2 : MOVZXReg4Mem2, node,
+            temp2Reg, generateX86MemoryReference(castClassReg, offsetof(J9Class, classDepthAndFlags), cg), cg);
+   
+   // cast class depth >= obj class depth, throw 
+   generateRegMemInstruction(CMP2RegMem, node, temp2Reg, generateX86MemoryReference(objClassReg, offsetof(J9Class, classDepthAndFlags), cg), cg);
+   generateLabelInstruction(JAE4, node, throwLabel, cg);
+   
+   // check obj class's super class array entry
+   // temp1Reg holds superClasses array of obj class
+   // An alternative sequences requiring one less register may be:
+   // SHL temp2Reg, 3 for 64-bit or 2 for 32-bit
+   // ADD temp2Reg, [temp3Reg, superclasses offset]
+   // CMP classClassReg, [temp2Reg]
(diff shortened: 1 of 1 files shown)


dsouzai:
Since this enabled under AOT (w/ SVM), please do some sanity tests to make sure the necessarily relocations are handled :)

yanluo7:
@0dvictor hi Victor, could you also review this?

yanluo7:
@dsouzai Ran forceAOT on tests from personal builds, no problem was observed.

runtime/compiler/x/codegen/J9TreeEvaluator.cpp
```diff
@@ -3974,45 +3974,193 @@ TR::Register *J9::X86::TreeEvaluator::checkcastinstanceofEvaluator(TR::Node *nod
          break;
       }
    auto clazz = TR::TreeEvaluator::getCastClassAddress(node->getChild(1));
-   if (clazz &&
-       !TR::Compiler->cls.isClassArray(cg->comp(), clazz) && // not yet optimized
-       (!cg->comp()->compileRelocatableCode() || cg->comp()->getOption(TR_UseSymbolValidationManager)) &&
-       !cg->comp()->getOption(TR_DisableInlineCheckCast)  &&
-       !cg->comp()->getOption(TR_DisableInlineInstanceOf))
+   if (isCheckCast && !clazz && (!cg->comp()->compileRelocatableCode() || cg->comp()->getOption(TR_UseSymbolValidationManager)))
```
0dvictor: We may want to disable the optimization when `TR_DisableInlineCheckCast` is set.

runtime/compiler/x/codegen/J9TreeEvaluator.cpp
```diff
@@ -3974,45 +3974,193 @@ TR::Register *J9::X86::TreeEvaluator::checkcastinstanceofEvaluator(TR::Node *nod
...
       {
-      cg->evaluate(node->getChild(0));
-      if (TR::Compiler->cls.isInterfaceClass(cg->comp(), clazz))
-         {
-         generateInlinedCheckCastOrInstanceOfForInterface(node, clazz, cg, isCheckCast);
-         }
-      else
+      auto use64BitClasses = TR::Compiler->target.is64Bit() &&
```
0dvictor: The code would look cleaner if these logic is extracted into a separate static function.

yanluo7: Refactored into a static function `is64BitClasses`

runtime/compiler/x/codegen/J9TreeEvaluator.cpp
yanluo7: Yes, added the option.

andrewcraik:
@0dvictor can you please review the updates to see if you are now happy with this change? @yanluo7 is there an issue to track doing this for other codegens? FYI @fjeremic @gita-omr

fjeremic:
> @yanluo7 is there an issue to track doing this for other codegens? FYI @fjeremic @gita-omr

@r30shah can you confirm whether we already have this on Z with the common instanceof infrastructure?

yanluo7:
@fjeremic @gita-omr https://github.com/eclipse/openj9/issues/5291 created to track P/Z implementation.

r30shah:
@fjeremic  ATM with `checkcast` we do not have support for dynamic cast class (https://github.com/eclipse/openj9/blob/5f146b60acce228f37dedc630d60ab677c25e922/runtime/compiler/codegen/J9TreeEvaluator.cpp#L269 , Looking into the changes needed I think it won't be that difficult to add support.

runtime/compiler/x/codegen/J9TreeEvaluator.cpp
```diff
@@ -3664,6 +3664,151 @@ TR::Register *J9::X86::TreeEvaluator::longBitCount(TR::Node *node, TR::CodeGener
...
+inline void generateInlinedCheckCastForDynamicCastClass(TR::Node* node, TR::CodeGenerator* cg)
+   {
+   auto use64BitClasses = TR::Compiler->target.is64Bit() &&
+               (!TR::Compiler->om.generateCompressedObjectHeaders() ||
+               (cg->comp()->compileRelocatableCode() && cg->comp()->getOption(TR_UseSymbolValidationManager)));
+   TR::Register *ObjReg = cg->evaluate(node->getFirstChild());
+   TR::Register *castClassReg = cg->evaluate(node->getSecondChild());
+   TR::Register *temp1Reg = cg->allocateRegister();
+   TR::Register *temp2Reg = cg->allocateRegister();
+   TR::Register *temp3Reg = cg->allocateRegister();
```
0dvictor: `temp3Reg` seems always holds the `J9Class` of `ObjReg`. It may be better to reflect this from its name to improve readability.
`temp1Reg` and `temp2Reg` are legit temporary registers. I agree to keep the name; however it may be easier to read the code if there are some comments describing their content wherever used, OR aliasing the variable with a meaning name before using it.

runtime/compiler/x/codegen/J9TreeEvaluator.cpp
```diff
@@ -3664,6 +3664,151 @@ TR::Register *J9::X86::TreeEvaluator::longBitCount(TR::Node *node, TR::CodeGener
...
+   
+   TR_OutlinedInstructions *outlinedHelperCall = new (cg->trHeapMemory()) TR_OutlinedInstructions(node, TR::call, NULL, outlinedCallLabel, fallThruLabel, cg);
+   cg->getOutlinedInstructionsList().push_front(outlinedHelperCall);
+   
+   generateRegMemInstruction(LRegMem(), node, temp2Reg, generateX86MemoryReference(castClassReg, offsetof(J9Class, romClass), cg), cg);
+   
+   // If cast class is array, call out of line helper
+   generateMemImmInstruction(TEST4MemImm4, node,
+       generateX86MemoryReference(temp2Reg, offsetof(J9ROMClass, modifiers), cg), J9AccClassArray, cg);
+   generateLabelInstruction(JE4, node, notArrayLabel, cg);
```
0dvictor: This part effectively generates following code:
```
    JE notArrayLabel
    Logics related to NULLCHK
    JMP OutlinedLabel
notArrayLabel:
    ...
```

We can move NULLCHK logics into outlined code so that this could become:
```
    JNE OutlinedLabel
    ...
```

runtime/compiler/x/codegen/J9TreeEvaluator.cpp
```diff
@@ -3664,6 +3664,151 @@ TR::Register *J9::X86::TreeEvaluator::longBitCount(TR::Node *node, TR::CodeGener
...
+   generateLabelInstruction(JE4, node, fallThruLabel, cg);
+   
+   // class not equal,
+   // if cast class depth >= min depth, compare with object class depth
+   // store cast class depth in temp2 for later use
+   // class depth mask must be low 16 bits to safely load without the mask.
+   static_assert(J9AccClassDepthMask == 0xffff, "J9_JAVA_CLASS_DEPTH_MASK must be 0xffff");
+   generateRegMemInstruction(TR::Compiler->target.is64Bit()? MOVZXReg8Mem2 : MOVZXReg4Mem2, node,
+            temp2Reg, generateX86MemoryReference(castClassReg, offsetof(J9Class, classDepthAndFlags), cg), cg);
+   generateRegImmInstruction(CMP2RegImm2, node, temp2Reg, ((int16_t)(cg->comp()->getOptions()->_minimumSuperclassArraySize)), cg);
```
0dvictor: Checking if `cast class depth >= min depth` is unnecessary, as the checking if `cast class depth >= obj class depth` is sufficient to conclude if testing superclass is required.

runtime/compiler/x/codegen/J9TreeEvaluator.cpp
```diff
@@ -3664,6 +3664,151 @@ TR::Register *J9::X86::TreeEvaluator::longBitCount(TR::Node *node, TR::CodeGener
...
+   generateLabelInstruction(JB4, node, checkSuperClassArrayLabel, cg);
+   
+   // cast class depth >= obj class depth, fail
+   generateRegMemInstruction(CMP2RegMem, node, temp2Reg, generateX86MemoryReference(temp3Reg, offsetof(J9Class, classDepthAndFlags), cg), cg);
+   generateLabelInstruction(JAE4, node, throwLabel, cg);
+   
+   // check obj class's super class array entry
+   generateLabelInstruction(LABEL, node, checkSuperClassArrayLabel, cg);
```
0dvictor: Just FYI, an alternative sequences requiring one less instruction may be:
```
SHL temp2Reg, 3 for 64-bit or 2 for 32-bit
ADD temp2Reg, [temp3Reg, superclasses offset]
CMP classClassReg, [temp2Reg]
```
I'm not sure which one is better. Your approach has one less instruction but uses one more register comparing to this alternative.

runtime/compiler/x/codegen/J9TreeEvaluator.cpp
```diff
@@ -3664,6 +3664,151 @@ TR::Register *J9::X86::TreeEvaluator::longBitCount(TR::Node *node, TR::CodeGener
...
+   generateRegMemInstruction(LRegMem(), node, temp1Reg, generateX86MemoryReference(temp3Reg, offsetof(J9Class, superclasses), cg), cg);
+   generateRegMemInstruction(CMPRegMem(use64BitClasses), node, castClassReg,
+       generateX86MemoryReference(temp1Reg, temp2Reg, TR::Compiler->target.is64Bit()?3:2, cg), cg);
+   generateLabelInstruction(JNE4, node, throwLabel, cg);
+   
+   // throw classCastException
+   {
+      TR_OutlinedInstructionsGenerator og(throwLabel, node, cg);
+      //generateLabelInstruction(LABEL, node, throwLabel, cg);
```
0dvictor: This comment seems unnecessary, I'd prefer removing it.

andrewcraik:
@yanluo7 please address comments from @0dvictor when you can and at us once updated so we can continue the review - thanks!

runtime/compiler/x/codegen/J9TreeEvaluator.cpp
yanluo7: Addressed.

runtime/compiler/x/codegen/J9TreeEvaluator.cpp
yanluo7: Thanks for pointing it out. On 64 bit, the extra reg isn't likely to cause significant register pressure. On 32 bit, it could put more register pressure due to limited number of regs. Since 64-bit is more prevalent, I think we could optimize for 64bit and leave as is, and I added comments here to reflect this tradeoff made.

runtime/compiler/x/codegen/J9TreeEvaluator.cpp
yanluo7: Addressed. Removed the extraneous min depth check.

runtime/compiler/x/codegen/J9TreeEvaluator.cpp
yanluo7: Addressed. We can just use the VFT load and push it ahead of the array class test, this way if the cast class is array, it serves as a null check; and if the cast class is not array, it serves as both a null check and loading the object class. With this, for the array case the explicit TEST4MemImm4 test instruction is deleted and can safely JNE to outlined helper.

runtime/compiler/x/codegen/J9TreeEvaluator.cpp
yanluo7: Addressed. Renamed temp3Reg and added more verbose comments when temp1/2 is (re)defined.

yanluo7:
@andrewcraik @0dvictor Comments have been addressed.

0dvictor:
LGTM.

andrewcraik:
@yanluo7 looks good - could you please create an issue to track adding this since z and p will need support too. I will start sanity.

andrewcraik:

yanluo7:
@andrewcraik https://github.com/eclipse/openj9/issues/5291 has already been created few days back. @fjeremic and @gita-omr are aware of the change.

andrewcraik:
Ok thanks I missed that scrolling through the history.

yanluo7:
The failure in JDK8 - linux_x86-64_cmprssptrs Sanity testing comes from cmdLineTester_test_0: the subtest themselves actually passed, but the pattern matching doesn't recognize the passing string for some tests:
```
 [OUT] Time spent starting: 95 milliseconds
 [OUT] Time spent executing: 99 milliseconds
 [OUT] Test result: PASSED
 [OUT] 
 [OUT] 
 [OUT] ---TEST RESULTS---
 [OUT] Number of PASSED tests: 1 out of 1
 [OUT] Number of FAILED tests: 0 out of 1
>> Required condition was found: [Return code: 0]
>> Success condition was not found: [Output match: Number of FAILED tests: 0 out of \d]
```
In this build, although checkcast dynamic cast class is supported but there should not be any such cases as the CFG simplifier change that enables non-constant cast class isn't yet merged. 
With my personal build that does have the non-constant dynamic cast class enabled plus this change, a 10-run grinder build with the same test all passed. @andrewcraik

andrewcraik:
Ok looks like there is no failure related - this is an enablement step for another optimization. Merging

PR #5146, opened 2019-03-18
Pass test pipelines ARTIFACTORY_REPO

AdamBrousseau:
- Test Pipelines require this or they default to
  sys-rt-generic-local which is undesirable.
- Add Jenkins Domain to Artifactory path
- Test Pipelines already have the domain as
  part of the path. This will match that behaviour.
- See openjdk-tests/JenkinsfileBase#L478

Issue #2637
Related #4705
[skip ci]

Signed-off-by: Adam Brousseau <adam.brousseau88@gmail.com>

PR #4705, opened 2019-02-12
Add Artifactory details for Eclipse OpenJ9 Jenkins

AdamBrousseau:
- This will enable builds to push to Artifactory

Issue #2637
Related #5146 
[skip ci]

Signed-off-by: Adam Brousseau <adam.brousseau88@gmail.com>

Issue #2637, opened 2018-08-16
Investigate Options for Jenkins artifact Storage


AdamBrousseau:
Related #1017 
Our space is limited on our Jenkins master. Eclipse has asked us to target 100G and we are currently using almost double of that [[1]](https://ci.eclipse.org/openj9/view/Infrastructure/job/Check-Master-Disk-Usage/). [Artifactory](https://jfrog.com/artifactory/) is an artifact repository manager and could allow us to stash (temporarily) our SDKs somewhere other than master. They offer a [free version](https://jfrog.com/open-source/) ([Apply](https://jfrog.com/artifactory/saas-gcp/)). This would allow us to keep more builds (logs) on master and more artifacts on a separate server. There is also an Artifactory plugin for Jenkins that would allow us to easily switch to archivingLast time I looked I believe there was an option to setup Artifactory yourself for "free"?

We have an existing machine ([proxy_worker](https://ci.eclipse.org/openj9/computer/proxy_worker/)) attached to Jenkins that we already use for running infra type builds as well as a proxy into other machines we use that are behind a vpn. We could try setting up Artifactory on it. This machine is provisioned via an OpenStack service and has the option to create volumes (up to 1T total) and attach them to running instance(s). This would give us the storage.

**NOTE**: 
- [x] We need to determine if this type of usage of the plinux machine is aligned with and doesn't break any agreement with the machine provider (OSU)

Joe already has a Pull Request open to add Artifactory support to the builds.
#2272 
AdoptOpenJDK/openjdk-tests#512

AdamBrousseau:
I went ahead and created a 1T volume in the OpenStack Dashboard and attached it to proxy_worker (/dev/vdb). I needed to format etc the drive and mount it on the machine.
ref: https://askubuntu.com/questions/154180/how-to-mount-a-new-drive-on-startup

```
sudo fdisk /dev/sdb
```
Press o and press Enter (creates a new table)
Press n and press Enter (creates a new partition)
Press p and press Enter (makes a primary partition)
Then press 1 and press Enter (creates it as the 1st partition)
`First sector (2048-2097151999, default 2048):` Enter
`Last sector, +sectors or +size{K,M,G,T,P} (2048-2097151999, default 2097151999):` Enter
Finally, press w (this will write any changes to disk)

Create a filesystem
```
sudo mkfs.ext4 /dev/vdb1
```
Add it to /etc/fstab
```
#device        mountpoint             fstype    options  dump   fsck
[...]

PR #5179, opened 2019-03-20
No longer pass Artifactory details to Test Pipelines

AdamBrousseau:
- Test jobs will have default params set

Related #5146 #2637 #4705
[skip ci]


AdamBrousseau:
@jdekonin and @llxia for review. Note this will need to merge after at least the internal Test jobs are regenerated.

llxia:
We have set `ARTIFACTORY_SERVER ` in internal test jobs already. Did we miss any test job?

AdamBrousseau:
True. So this can be merged now since the test code has default repo value as `sys-rt-generic-local`. And once the external jobs are regenerated with server and repo then Artifactory will be enabled for test jobs there.

PR #5180, opened 2019-03-20
Disable SVM during Startup

dsouzai:
Mild admission of defeat, but since there are issues in startup on multiple platforms in Java 11, this is a temporary solution until a better one is implemented. For the most part it seems the reason SVM regresses startup is because of increased load failures that comes from compiling at invocation count 1000 (or 250) but loading at invocation count 20 (https://github.com/eclipse/openj9/issues/4897).


dsouzai:
FYI @vijaysun-omr 

dsouzai:

dsouzai:
@mpirvu all failures were infra failures. For three of them the tests passed but the jenkins job failed; the last job failed because of 
```
error: RPC failed; curl 18 transfer closed with outstanding read data remaining
fatal: The remote end hung up unexpectedly
fatal: early EOF
fatal: index-pack failed
```

This should be ok to merge now.

PR #5188, opened 2019-03-20
Salt on Nursery resize

amicic:
Fix incorrect update of low/high Nursery bounds for Hash Salt
calculation purposes.

On Nursery contract, the new bounds would end up being the range of
contracted area. An example:
old Nursery bounds hashData1/2 ff740000/100000000
low/highAddress (range being contracted) ff740000/ff760000
new Nursery bounds hashData1/2 ff740000/ff760000

Instead, the correct new bounds of Nursery should be: ff760000/100000000

This is relatively benign problem for STW Scavenger (there would be less
variety in Salt value). However, this a malign problem for Concurrent
Scavenger. The salt value (hence, hash value, too) would change for
objects that are allocated during active CS cycle and has been just
hashed, if there was a Nursery contract at the end of that cycle.
Such an object would reside in Nursery both before and after the
second/last STW increment of CS cycle, but salt value would be
different, because the Nursery range for Salt calculation was
incorrectly updated, and that object would be incorrectly treated as out
of Nursery, after the contract.

Some other minor ranges in the logic is to ensure that range being
contracted/expanded is always exactly at the border of exactly one of
the lower or higher bound.


amicic:
jenkins compile win jdk11

dmitripivkine:

dmitripivkine:

dmitripivkine:

dmitripivkine:

PR #5202, opened 2019-03-21
Remove the assertion and return if j9shr_getCacheDir() failed

hangshao0:
There is no need to assert and return if we cannot get the default cache
directory in shrcssup.c. We are checking if the directory passed in via
"cacheDir=" is the same as the default directory. We ignore the
"cacheDirPerm=" if it is the default directory. So there is no need to
assert and return if user is not using the default directory at all. 


pshipton:

PR #5211, opened 2019-03-22
Fix and re-enable vectorized String.indexOf on 64-bit Power

aviansie-ben:
When it was first added, the vectorized versions of the `intrinsicIndexOf` helpers for Power had to be disabled due to mysterious difficult-to-reproduce segfaults it appeared to be causing. These have since been tracked down to the code falsely assuming that the upper word of a 64-bit register containing a 32-bit value were zeroes, which is not guaranteed. This PR corrects that issue, fixes a tangentially related sign-extension issue with the same helpers, and re-enables the vectorized versions of these helpers on 64-bit Power.

### DIFF
--- a/runtime/compiler/p/codegen/J9TreeEvaluator.cpp
+++ b/runtime/compiler/p/codegen/J9TreeEvaluator.cpp
@@ -11782,15 +11782,15 @@ static TR::Register *inlineIntrinsicIndexOf(TR::Node *node, bool isLatin1, TR::C
    generateTrg1Src2Instruction(cg, TR::InstOpCode::cmp4, node, cr0, startIndex, endIndex);
    generateConditionalBranchInstruction(cg, TR::InstOpCode::beq, node, notFoundLabel, cr0);
 
+   // IMPORTANT: The upper 32 bits of a 64-bit register containing an int are undefined. Since the
+   // indices are being passed in as ints, we must ensure that their upper 32 bits are not garbage.
+   generateTrg1Src1Instruction(cg, TR::InstOpCode::extsw, node, result, startIndex);
+   generateTrg1Src1Instruction(cg, TR::InstOpCode::extsw, node, endAddress, endIndex);
+
    if (!isLatin1)
       {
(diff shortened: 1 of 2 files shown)


runtime/compiler/p/codegen/J9TreeEvaluator.cpp
```diff
@@ -11808,15 +11808,21 @@ static TR::Register *inlineIntrinsicIndexOf(TR::Node *node, bool isLatin1, TR::C
    // to each String.indexOf call is ' ', which is the most common whitespace character.
       {
       TR::Register *value = srm->findOrCreateScratchRegister();
+      TR::Register *zxTargetScalar = srm->findOrCreateScratchRegister();
+
+      // Since we're going to do a load followed immediately by a comparison, we need to ensure that
```
gita-omr: So you think we only need to do that for the first character and not for the rest (in the vector loop) ?

aviansie-ben: That's correct. Only the first comparison is done using a 32-bit scalar comparison operator. It doesn't matter what the upper bits of this register are when performing the vector comparison, as they will have been overwritten when the value is splatted into the vector register.

gita-omr:
Jenkins test sanity aix,plinux jdk8,jdk11

gita-omr:
@pshipton this is a very good performance improvement to String.indexOf on Power. We would like to get it into 0.14.

pshipton:
@gita-omr no objections, as long as it is promptly backed out or fixed if it causes any further problems

pshipton:
@gita-omr perhaps wait until after the split for 0.14 is completed on Monday, then deliver to the master branch. If all looks good after a few days, then deliver it to the 0.14 branch as well.

gita-omr:
Jenkins test extended aix,plinux jdk8,jdk11

gita-omr:
@aviansie-ben @zl-wang please follow the process @pshipton described above.

PR #5213, opened 2019-03-22
Look for additional opportunities for stack allocation

hzongaro:
This introduces some further refinements to detect opportunities for Escape Analysis (EA) in loops where references are copied between local variables.

This change considers the right-hand side of an assignment to be harmless for the purposes of EA if it was an object that was locally allocated by a previous pass of EA, not only if it is a candidate for local allocation in the current pass of EA.

This change was prototyped by Andrew Craik <ajcraik@ca.ibm.com>.

Also, fixed a bug in trace code that was getting the class name of a `TR::New` operation incorrectly.

### DIFF
--- a/runtime/compiler/optimizer/EscapeAnalysis.cpp
+++ b/runtime/compiler/optimizer/EscapeAnalysis.cpp
@@ -543,7 +543,8 @@ int32_t TR_EscapeAnalysis::performAnalysisOnce()
    _fixedVirtualCallSites.setFirst(NULL);
 
    _parms = NULL;
-   _localObjectsValueNumbers = NULL;
+   _nonColdLocalObjectsValueNumbers = NULL;
+   _allLocalObjectsValueNumbers = NULL;
    _visitedNodes = NULL;
    _aliasesOfAllocNode = NULL;
    _aliasesOfOtherAllocNode = NULL;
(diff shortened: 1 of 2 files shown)


andrewcraik:

andrewcraik:
these changes look like a refined version of my prototype. As someone who originally wrote the code I think it would be poor form for me to do a review. @vijaysun-omr could you please review as one of the other experts in EA?

andrewcraik:
@vijaysun-omr ping?

vijaysun-omr:
If we ignore the uses under prepareForOSR blocks, how do we ensure that the interpreter does not see a stack allocated object from the JIT stack after an OSR transition ?

hzongaro:
@andrewcraik Andrew, are you able to respond to [Vijay's comment](https://github.com/eclipse/openj9/pull/5213#issuecomment-477628015)?  In particular, was the part of the change that records as ignoreable the uses that appear under calls to `prepareForOSR` dependent on another change of yours that's still pending?  If so, I can remove that.

vijaysun-omr:
@hzongaro I was also thinking the same, i.e. whether there was a pending change that is part of the full picture here.

andrewcraik:
That is related to another change that is coming - it shouldn't be in this change.

hzongaro:
Vijay @vijaysun-omr, Andrew @andrewcraik, I have removed the part of the change related to checking for uses under calls to `PrepareForOSR`.

Sorry about all the `git push --force` churn - I needed to rebase on a more recent upstream master in order to rerun testing, and that ended up with github seeming to think there were changes unrelated to this one that were being pull requested.

vijaysun-omr:
Looks okay to me. @andrewcraik can also review and check that the right amount of testing has been done.

andrewcraik:

Issue #3656, opened 2018-11-13
Missing support for ARM processors


keithc-ca:
At least these classes, in `debugtools/DDR_VM/src/com/ibm/j9ddr/vm29/j9/stackwalker`, need changes to support ARM processors:
```
JITRegMap
StackWalker
StackWalkerUtils
```
instead of throwing `IllegalArgumentException("Unsupported platform")`.

pshipton:
#53
@knn-k fyi

knn-k:
Thank you for letting me know.

JamesKingdon:
Do we have a more general issue for enabling DDR in arm builds? My concern is that I'm not sure it's possible to do the ddr parts of the build in a cross compile (maybe with qemu?), and a ddr enabled native build could be very time consuming unless we have server grade hardware (and good parallelism in the DDR build steps).

keithc-ca:
I'm not aware of a more general issue for enabling DDR for arm builds.

There would be some OMR work to reclassify ddrgen as a 'host' tool so that it gets built properly to run on the host and analyze target objects (which will be challenging when the host and target don't both use DWARF or the same formats).

PR #5220, opened 2019-03-24
AArch64: Update Dockerfile for cross-compilation

knn-k:
This commit updates the Dockerfile for aarch64 cross-compilation.
- Update the file names for missing libcups libraries
- Remove unnecessary libraries for x86

### DIFF
--- a/buildenv/docker/jdk11/aarch64_CC/arm-linux-aarch64/Dockerfile
+++ b/buildenv/docker/jdk11/aarch64_CC/arm-linux-aarch64/Dockerfile
@@ -42,18 +42,6 @@ RUN apt-get update \
     gcc-7 \
     git \
     git-core \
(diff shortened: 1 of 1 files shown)


DanHeidinga:
I don't see  a`libelf-dev` equivalent in the .deb files being installed.  Does DDR successfully build on aarch64?  

If it doesn't, I'm OK with merging it as DDR will likely require additional work to become function on aarch64 and adding it to the docker file can be done then

knn-k:
AArch64 is not building DDR so far.  You can find the following line in `linux_aarch64_cmprssptrs.spec`:
```
  <flag id="module_ddr" value="false"/>
```
32-bit ARM is not building DDR, either. (#3656)

knn-k:
I opened #5221 for future AArch64 DDR support.

PR #5222, opened 2019-03-25
AArch64: Add libssl files to Dockerfile

knn-k:
This commit adds libssl files to the Dockerfile for AArch64 cross-compilation.
It also updates the Build Instructions with the information on how to build
the runtime with OpenSSL support.

### DIFF
--- a/doc/build-instructions/Build_Instructions_V11.md
+++ b/doc/build-instructions/Build_Instructions_V11.md
@@ -716,7 +716,7 @@ bash configure --openjdk-target=${OPENJ9_CC_PREFIX} \
 
   where:
 
-  - `path_to_library` uses a custom OpenSSL v1.1.x library that's already built.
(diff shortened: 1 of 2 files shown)


doc/build-instructions/Build_Instructions_V11.md
```diff
@@ -716,7 +716,7 @@ bash configure --openjdk-target=${OPENJ9_CC_PREFIX} \
 
   where:
 
-  - `path_to_library` uses a custom OpenSSL v1.1.x library that's already built.
+  - `path_to_library` uses an OpenSSL v1.1.x library that's already built.  You can use `${OPENJ9_CC_DIR}/${OPENJ9_CC_PREFIX}/libc/usr` as `path_to_library` when you are configuring in the Docker container.
```
pshipton: The same change should be added for other versions (8, 12) as well.

knn-k: The other versions don't have the `AArch64` section in the Build Instruction yet.  So this is only for 11 right now.

PR #5226, opened 2019-03-25
Handle NullPointerExceptions in ThreadInfo.from()

pdbain-ibm:
If the CompositeData is missing a key the parsing code will throw an NPE.
catch it and translate it to an IllegalArgumentException.


pshipton:
jenkins compile zlinux jdk8

PR #5232, opened 2019-03-25
Don't copy monitor object in Read Barrier

amicic:
Monitor table is weak (clearable) root and Read Barrier should not copy
objects, since info if objects is live or not is not known in a middle
of a GC cycle. The barrier only needs to update the reference slot if
object happened to be already copied.

For API with vmThread provided, also impose the assert that object is
already copied, which is true for typical synchronizing object
(vmThread->blockingEnterObject). The other API with VM provided should
be used when accessing random monitor table slots with unknown liveness
info.


dmitripivkine:
Jenkins test sanity zLinux jdk11

dmitripivkine:
Jenkins compile win

PR #5234, opened 2019-03-25
Read Barrier for atomics on statics

amicic:
Invoke read barrier prior to execution atomic operations (like
compare&swap) on static class fields. We've already had proper barriers
for both:
- non atomic read of statics
- atomic access of non-statics
but we missed to add it for this combination.

This is a pre-req for statics scanning being removed from roots scanning
for GCs that rely on read barrier (like Concurrent Scavenger).


dmitripivkine:

dmitripivkine:
Jenkins compile win

Issue #4945, opened 2019-03-01
JTReg Test Failure : java/lang/ProcessBuilder/Basic.java


ben-walsh:
Consistent failures ...

```
[2019-03-01T14:31:57.603Z] STDOUT:
[2019-03-01T14:31:57.603Z] This appears to be a Unix system.
[2019-03-01T14:31:57.603Z] Testing ASCII environment values
[2019-03-01T14:31:57.603Z] Testing Latin1 environment values
[2019-03-01T14:31:57.603Z] Testing Unicode environment values
[2019-03-01T14:31:57.603Z] Testing directory .
[2019-03-01T14:31:57.603Z] Testing directory ..
[2019-03-01T14:31:57.603Z] Testing directory /
[2019-03-01T14:31:57.603Z] Testing directory /bin
[2019-03-01T14:31:57.603Z] cmdp
[2019-03-01T14:31:57.603Z] cmdp 0: /home/jenkins/workspace/Grinder/openjdkbinary/j2sdk-image/bin/java
[2019-03-01T14:31:57.603Z] cmdp 1: -XX:+DisplayVMOutputToStderr
[2019-03-01T14:31:57.603Z] cmdp 2: -classpath
[2019-03-01T14:31:57.603Z] cmdp 3: /home/jenkins/workspace/Grinder/jvmtest/openjdk_regression/work/classes/0/java/lang/ProcessBuilder/Basic_id0.d:/home/jenkins/workspace/Grinder/openjdk-tests/openjdk_regression/openjdk-jdk/test/jdk/java/lang/ProcessBuilder:/home/jenkins/workspace/Grinder/jvmtest/openjdk_regression/jtreg/lib/javatest.jar:/home/jenkins/workspace/Grinder/jvmtest/openjdk_regression/jtreg/lib/jtreg.jar
[2019-03-01T14:31:57.603Z] cmdp 4: Basic$JavaChild
[2019-03-01T14:31:57.603Z] cmdp 5: System.getenv()
[2019-03-01T14:31:57.603Z] envp
[2019-03-01T14:31:57.603Z] envp 0: LC_ALL=C��
[2019-03-01T14:31:57.603Z] envp 1: FO�=B�R
[2019-03-01T14:31:57.603Z]  waitFor process: delta: 139618
[2019-03-01T14:31:57.603Z] 
[2019-03-01T14:31:57.603Z] Passed = 9021, failed = 2
[2019-03-01T14:31:57.603Z] 
[...]

PR #5239, opened 2019-03-25
Add support for -XX:OnOutOfMemoryError=string

sharon-wang:
Map -XX:OnOutOfMemoryError=string to the existing OpenJ9 equivalent -Xdump:tool:events=systhrow,filter=java/lang/OutOfMemoryError,exec=string

Fixes: #4945.

Documentation Issue: https://github.com/eclipse/openj9-docs/issues/242

### DIFF
--- a/runtime/rasdump/dmpsup.c
+++ b/runtime/rasdump/dmpsup.c
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 1991, 2018 IBM Corp. and others
+ * Copyright (c) 1991, 2019 IBM Corp. and others
  *
  * This program and the accompanying materials are made available under
  * the terms of the Eclipse Public License 2.0 which accompanies this
@@ -577,7 +577,42 @@ configureDumpAgents(J9JavaVM *vm)
 		}
 		if ( IS_CONSUMABLE(j9vm_args, xdumpIndex) && !IS_CONSUMED(j9vm_args, xdumpIndex) )
 		{
-			GET_OPTION_VALUE(xdumpIndex, ':', &optionString);
+			BOOLEAN isMappedToolDump = FALSE;
+			/* Handle mapped tool dump options */
+			if (HAS_MAPPING(j9vm_args, xdumpIndex)) {
+				char *mappingJ9Name = MAPPING_J9NAME(j9vm_args, xdumpIndex);
+				char *toolString = ":tool:";
+				char *toolCursor = strstr(mappingJ9Name, toolString);
+				if (NULL != toolCursor) {
+					char *optionValue = NULL;
+					/* The mapped option specifies the tool command to run after the equals */
+					GET_OPTION_VALUE(xdumpIndex, '=', &optionValue);
+					
+					/* Move toolCursor past ":tool:" */
+					toolCursor += strlen(toolString);
+
+					if (NULL != optionValue) {
+						size_t toolCursorLength = strlen(toolCursor);
+						size_t optionValueLength = strlen(optionValue);
+						size_t optionStringMemAlloc = toolCursorLength + optionValueLength + 1;
+
+						/* Construct optionString by combining the J9 tool dump command with the mapped option */
+						optionString = (char *) j9mem_allocate_memory(optionStringMemAlloc, OMRMEM_CATEGORY_VM);
+						
+						if (NULL != optionString) {
+							strcpy(optionString, toolCursor);
+							strcat(optionString + toolCursorLength, optionValue);
+							isMappedToolDump = TRUE;
+						} else {
+							char *mappingMapName = MAPPING_MAPNAME(j9vm_args, xdumpIndex);
+							j9tty_err_printf(PORTLIB, "Unable to map %s to J9 %s - Could not allocate the requested size of memory %zu for optionString\n", mappingMapName, mappingJ9Name, optionStringMemAlloc);
+							return J9VMDLLMAIN_FAILED;
+						}
+					}
+				}
+			} else {
+				GET_OPTION_VALUE(xdumpIndex, ':', &optionString);
+			}
 			if (!optionString) {
 				/* ... silent option ... */
 			} else if( strncmp(optionString, "none", strlen("none") ) == 0 ){
@@ -589,6 +624,13 @@ configureDumpAgents(J9JavaVM *vm)
 					agentOpts[agentNum].pass = J9RAS_DUMP_OPTS_PASS_ONE;
 					agentNum++;
 				}
+			} else if (isMappedToolDump) {
+				char * toolString = "tool";
+				agentOpts[agentNum].kind = scanDumpType(&toolString);
+				agentOpts[agentNum].flags = J9RAS_DUMP_OPT_ARGS_ALLOC;
+				agentOpts[agentNum].args = optionString;
+				agentOpts[agentNum].pass = J9RAS_DUMP_OPTS_PASS_ONE;
+				agentNum++;
 			} else {
 				char *typeString = optionString;
 
--- a/runtime/vm/jvminit.c
+++ b/runtime/vm/jvminit.c
@@ -4020,6 +4020,10 @@ registerVMCmdLineMappings(J9JavaVM* vm)
 	if (registerCmdLineMapping(vm, MAPOPT_XXINITIALHEAPSIZE_EQUALS, VMOPT_XMS, EXACT_MAP_WITH_OPTIONS) == RC_FAILED) {
 		return RC_FAILED;
 	}
+	/* Map -XX:OnOutOfMemoryError= to -Xdump:tool:events=systhrow,filter=java/lang/OutOfMemoryError,exec= */ 
+	if (registerCmdLineMapping(vm, MAPOPT_XXONOUTOFMEMORYERROR_EQUALS, VMOPT_XDUMP_TOOL_OUTOFMEMORYERROR_EXEC_EQUALS, EXACT_MAP_WITH_OPTIONS) == RC_FAILED) {
+		return RC_FAILED;
+	}
 
 	return 0;
 }
--- a/test/functional/cmdLineTests/cmdLineTest_J9tests/build.xml
+++ b/test/functional/cmdLineTests/cmdLineTest_J9tests/build.xml
@@ -1,7 +1,7 @@
 <?xml version="1.0"?>
 
 <!--
-  Copyright (c) 2016, 2018 IBM Corp. and others
+  Copyright (c) 2016, 2019 IBM Corp. and others
 
   This program and the accompanying materials are made available under
   the terms of the Eclipse Public License 2.0 which accompanies this
@@ -30,16 +30,43 @@
 
 	<!-- set properties for this build -->
 	<property name="DEST" value="${BUILD_ROOT}/functional/cmdLineTests/cmdLineTest_J9tests" />
-	<property name="src" location="." />
+	<property name="PROJECT_ROOT" location="." />
+	<property name="src" location="${PROJECT_ROOT}/src" />
+	<property name="build" location="${PROJECT_ROOT}/bin" />
 
-	<target name="dist" description="generate the distribution">
+	<target name="init">
+		<mkdir dir="${DEST}" />
+		<mkdir dir="${build}" />
+	</target>
+
+	<target name="compile" depends="init" description="Using java ${JDK_VERSION} to compile the source ">
+		<echo>Ant version is ${ant.version}</echo>
+		<echo>============COMPILER SETTINGS============</echo>
+		<echo>===fork:                         yes</echo>
+		<echo>===executable:                   ${compiler.javac}</echo>
+		<echo>===debug:                        on</echo>
+		<echo>===destdir:                      ${DEST}</echo>
+		<javac srcdir="${src}" destdir="${build}" debug="true" fork="true" executable="${compiler.javac}" includeAntRuntime="false" encoding="ISO-8859-1">
+		</javac>
+	</target>
+
+	<target name="dist" depends="compile" description="generate the distribution">
+		<jar jarfile="${DEST}/cmdLineTest_J9tests.jar" filesonly="true">
+			<fileset dir="${build}" />
+			<fileset dir="${src}" />
+		</jar>
 		<copy todir="${DEST}">
-			<fileset dir="${src}" includes="*.xml"/>
-			<fileset dir="${src}" includes="*.mk"/>
+			<fileset dir="${PROJECT_ROOT}" includes="*.xml" />
+			<fileset dir="${PROJECT_ROOT}" includes="*.mk" />
 		</copy>
 	</target>
 	
+	<target name="clean" depends="dist" description="clean up">
+		<!-- Delete the ${build} directory trees -->
+		<delete dir="${build}" />
+	</target>
+
 	<target name="build" >
-		<antcall target="dist" inheritall="true" />
+		<antcall target="clean" inheritall="true" />
 	</target>
 </project>
--- a/test/functional/cmdLineTests/cmdLineTest_J9tests/j9tests.xml
+++ b/test/functional/cmdLineTests/cmdLineTest_J9tests/j9tests.xml
@@ -24,11 +24,17 @@ SPDX-License-Identifier: EPL-2.0 OR Apache-2.0 OR GPL-2.0 WITH Classpath-excepti
 
 <!DOCTYPE suite SYSTEM "cmdlinetester.dtd">
 
-<suite id="J9 Command-Line Option Tests" timeout="180">
+<suite id="J9 Command-Line Option Tests" timeout="600">
     <variable name="READIPINFOFORRAS" value="-XX:+ReadIPInfoForRAS"/>
     <variable name="NOREADIPINFOFORRAS" value="-XX:-ReadIPInfoForRAS"/>
     <variable name="READIPINFOFORRAS_MESSAGE" value="enabled network query to determine host name and IP address for RAS."/>
     <variable name="NOREADIPINFOFORRAS_MESSAGE" value="disabled network query to determine host name and IP address for RAS."/>
+
+    <variable name="JVM_HEAP_LIMIT" value="-Xmx512m"/>
+    <variable name="ONOUTOFMEMORYERROR_EQUALS" value="-XX:OnOutOfMemoryError="/>
+    <variable name="ONOUTOFMEMORYERROR_JAR" value="-cp $Q$$JARPATH$$Q$ OnOutOfMemoryErrorTest"/>
+    <variable name="JAVALANGOUTOFMEMORYERROR" value="java.lang.OutOfMemoryError:"/>
+    <variable name="JVMREQUESTINGTOOLDUMP" value="JVM Requesting Tool dump using 'java -version'"/>
 
     <test id="Verify Generate a javacore to STDOUT">
         <command>$EXE$ -Xdump:java:events=vmstart,file=/STDOUT/</command>
@@ -58,6 +64,13 @@ SPDX-License-Identifier: EPL-2.0 OR Apache-2.0 OR GPL-2.0 WITH Classpath-excepti
         <command>$EXE$ $READIPINFOFORRAS$ $NOREADIPINFOFORRAS$ -verbose:init -version</command>
         <output type="success" caseSensitive="yes" regex="no">$NOREADIPINFOFORRAS_MESSAGE$</output>
         <output type="failure" caseSensitive="yes" regex="no">$READIPINFOFORRAS_MESSAGE$</output>
+    </test>
+
+    <test id="test -XX:OnOutOfMemoryError=">
+        <command>$EXE$ $JVM_HEAP_LIMIT$ $ONOUTOFMEMORYERROR_EQUALS$"java -version" $ONOUTOFMEMORYERROR_JAR$</command>
+        <output type="success" caseSensitive="yes" regex="no">$JAVALANGOUTOFMEMORYERROR$</output>
+        <output type="required" caseSensitive="yes" regex="no">$JVMREQUESTINGTOOLDUMP$</output>
+        <output type="required" caseSensitive="no" regex="yes" javaUtilPattern="yes">(java|openjdk) version</output>
     </test>
 </suite>
 
--- a/test/functional/cmdLineTests/cmdLineTest_J9tests/src/OnOutOfMemoryErrorTest.java
+++ b/test/functional/cmdLineTests/cmdLineTest_J9tests/src/OnOutOfMemoryErrorTest.java
@@ -1,0 +1,33 @@
+/*******************************************************************************
+ * Copyright (c) 2019, 2019 IBM Corp. and others
+ *
+ * This program and the accompanying materials are made available under
+ * the terms of the Eclipse Public License 2.0 which accompanies this
+ * distribution and is available at https://www.eclipse.org/legal/epl-2.0/
+ * or the Apache License, Version 2.0 which accompanies this distribution and
+ * is available at https://www.apache.org/licenses/LICENSE-2.0.
+ *
+ * This Source Code may also be made available under the following
+ * Secondary Licenses when the conditions for such availability set
+ * forth in the Eclipse Public License, v. 2.0 are satisfied: GNU
+ * General Public License, version 2 with the GNU Classpath
+ * Exception [1] and GNU General Public License, version 2 with the
+ * OpenJDK Assembly Exception [2].
+ *
+ * [1] https://www.gnu.org/software/classpath/license.html
+ * [2] http://openjdk.java.net/legal/assembly-exception.html
+ *
+ * SPDX-License-Identifier: EPL-2.0 OR Apache-2.0 OR GPL-2.0 WITH Classpath-exception-2.0 OR LicenseRef-GPL-2.0 WITH Assembly-exception
+ *******************************************************************************/
+
+/**
+ * @file OnOutOfMemoryErrorTest.java
+ * @brief Generates a java.lang.OutOfMemoryError to test the command line option
+ *        -XX:OnOutOfMemoryError=
+ */
+
+class OnOutOfMemoryErrorTest {
+    public static void main(String[] args) {
+        Object test = new double[Integer.MAX_VALUE];
+    }
+}
--- a/runtime/oti/jvminit.h
+++ b/runtime/oti/jvminit.h
@@ -268,6 +268,7 @@ enum INIT_STAGE {
 #define VMOPT_XDUMP  "-Xdump"
 #define VMOPT_XDUMP_NONE  "-Xdump:none"
 #define VMOPT_XDUMP_DIRECTORY_EQUALS  "-Xdump:directory="
+#define VMOPT_XDUMP_TOOL_OUTOFMEMORYERROR_EXEC_EQUALS "-Xdump:tool:events=systhrow,filter=java/lang/OutOfMemoryError,exec="
 #define VMOPT_XARGENCODING "-Xargencoding"
 #define VMOPT_XARGENCODINGCOLON "-Xargencoding:"
 #define VMOPT_XARGENCODINGUTF8 "-Xargencoding:utf8"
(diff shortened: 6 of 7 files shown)


sharon-wang:
Currently working on adding a test to: https://github.com/eclipse/openj9/blob/master/test/functional/cmdLineTests/xxargtest/XXArgumentTesting.xml 

I will open an issue in https://github.com/eclipse/openj9-docs for the new option after these changes are reviewed.

runtime/rasdump/dmpsup.c
```diff
@@ -577,7 +577,19 @@ configureDumpAgents(J9JavaVM *vm)
...
 		{
-			GET_OPTION_VALUE(xdumpIndex, ':', &optionString);
+			/* Handle mapped tool dump options */
+			if (HAS_MAPPING(j9vm_args, xdumpIndex)) {
+				/* Construct optionString by combining the J9 tool dump command with the mapped option */
+				char *toolCursor = strstr(MAPPING_J9NAME(j9vm_args, xdumpIndex), ":tool");
+				if (toolCursor != NULL) {
+					/* The mapped option specifies the tool command to run after the equals */
+					GET_OPTION_VALUE(xdumpIndex, '=', &optionString);
+					strcat(toolCursor++, optionString);
```
pdbain-ibm: - Please don't use side-effect operators line `toolCursor++` (except in for-loops).
- check for buffer overflow

runtime/rasdump/dmpsup.c
```diff
@@ -577,7 +577,18 @@ configureDumpAgents(J9JavaVM *vm)
...
 		{
-			GET_OPTION_VALUE(xdumpIndex, ':', &optionString);
+			/* Handle mapped tool dump options */
+			if (HAS_MAPPING(j9vm_args, xdumpIndex)) {
+				/* Construct optionString by combining the J9 tool dump command with the mapped option */
+				char *toolCursor = strstr(MAPPING_J9NAME(j9vm_args, xdumpIndex), ":tool");
+				if (toolCursor != NULL) {
+					/* The mapped option specifies the tool command to run after the equals */
+					GET_OPTION_VALUE(xdumpIndex, '=', &optionString);
+					optionString = strncat(toolCursor, optionString, (sizeof(toolCursor) - strlen(toolCursor) - 1)) + 1;
```
sharon-wang: @pdbain-ibm this is now using `strncat` instead of `strcat` to address buffer overflow - does the maximum source bytes value look right? (`sizeof(toolCursor) - strlen(toolCursor) - 1`)

pdbain-ibm: Since `toolCursor` is declared as `char *`, `sizeof(toolCursor)` will return the size of a pointer, not the string length.  

You will also need to handle the case where the new option string is longer than the old (which might be always).  In that case you will need to allocate (and free later) a sufficiently long buffer.

sharon-wang:
@pdbain-ibm I've made a few changes based on your suggestions - would appreciate a review when you have a moment!

runtime/rasdump/dmpsup.c
```diff
@@ -577,7 +578,27 @@ configureDumpAgents(J9JavaVM *vm)
 		}
 		if ( IS_CONSUMABLE(j9vm_args, xdumpIndex) && !IS_CONSUMED(j9vm_args, xdumpIndex) )
 		{
-			GET_OPTION_VALUE(xdumpIndex, ':', &optionString);
+			/* Handle mapped tool dump options */
+			if (HAS_MAPPING(j9vm_args, xdumpIndex)) {
+				char *toolCursor = strstr(MAPPING_J9NAME(j9vm_args, xdumpIndex), ":tool") + 1;
+				if (toolCursor) {
```
pdbain-ibm: Please use an explicit comparand, e.g. `NULL != toolCursor`.  Remember to put an rvalue on the left of equality and inequality operators.

runtime/rasdump/dmpsup.c
```diff
@@ -577,7 +578,27 @@ configureDumpAgents(J9JavaVM *vm)
...
+					/* The mapped option specifies the tool command to run after the equals */
+					char *optionValue = NULL;
+					GET_OPTION_VALUE(xdumpIndex, '=', &optionValue);
+					
+					/* Construct optionString by combining the J9 tool dump command with the mapped option */
+					int toolCursorLength = strlen(toolCursor);
+					int optionValueLength = strlen(optionValue);
+					optionString = (char *) j9mem_allocate_memory(toolCursorLength + optionValueLength + 1, OMRMEM_CATEGORY_VM);
+					strcpy(optionString, toolCursor);
+					strncat(optionString + toolCursorLength, optionValue, optionValueLength);
```
pdbain-ibm: - Please check for memory allocation failure and handle is the same as other places.
- You don't need `strncat` since you know that `optionValue` will fit.

runtime/rasdump/dmpsup.c
```diff
@@ -714,6 +735,9 @@ configureDumpAgents(J9JavaVM *vm)
 	}
 	j9mem_free_memory(agentOpts);
 
+	/* Free optionString memory at original address for a mappped tool dump */
+	if (optionStringOriginal) {j9mem_free_memory(optionStringOriginal);}
```
pdbain-ibm: - Same comment about comparison
- Please put the `j9mem_free_memory(optionStringOriginal)` statement on its own line.

runtime/vm/jvminit.c
```diff
@@ -3989,6 +3989,10 @@ registerVMCmdLineMappings(J9JavaVM* vm)
 	if (registerCmdLineMapping(vm, MAPOPT_XXINITIALHEAPSIZE_EQUALS, VMOPT_XMS, EXACT_MAP_WITH_OPTIONS) == RC_FAILED) {
 		return RC_FAILED;
 	}
+	/* Map -XX:OnOutOfMemoryError= to -Xdump:tool:events=systhrow,filter=java/lang/OutOfMemoryError,exec= */ 
+	if (registerCmdLineMapping(vm, MAPOPT_XXONOUTOFMEMORYERROR_EQUALS, VMOPT_XDUMP_TOOL_OUTOFMEMORYERROR_EXEC_EQUALS, EXACT_MAP_WITH_OPTIONS) == RC_FAILED) {
+		return RC_FAILED;
```
pdbain-ibm: Do you have a test case for this feature?

sharon-wang: I will add a test to check for failure cases of this feature. The JTReg test `Basic.java` that is currently failing in the associated issue (https://github.com/eclipse/openj9/issues/4945) will pass with this feature implemented - would this be sufficient for testing the success case of this feature?

runtime/rasdump/dmpsup.c
```diff
@@ -577,7 +578,32 @@ configureDumpAgents(J9JavaVM *vm)
...
+					GET_OPTION_VALUE(xdumpIndex, '=', &optionValue);
+					
+					/* Construct optionString by combining the J9 tool dump command with the mapped option */
+					int toolCursorLength = strlen(toolCursor);
+					int optionValueLength = strlen(optionValue);
+					optionString = (char *) j9mem_allocate_memory(toolCursorLength + optionValueLength + 1, OMRMEM_CATEGORY_VM);
+
+					if (NULL == optionString) {
+						j9tty_err_printf(PORTLIB, "Storage for mapped tool dump optionString not available, unable to process option\n");
+						return J9VMDLLMAIN_FAILED;
```
sharon-wang: @pdbain-ibm This is the added check for memory allocation failure. I went with `J9VMDLLMAIN_FAILED` for the error code since it was being used for a similar check in the method, but I noticed that `J9VMDLLMAIN_SILENT_EXIT_VM` and `OMR_ERROR_OUT_OF_NATIVE_MEMORY` are also used in other checks. Is there a particular error code I should use for this situation?

pdbain-ibm: Looks okay to me, thought the committers may have other opinions.
As discussed, please see if there is a test for the existing mapped option and extend that.

sharon-wang:
@pshipton Which release target should I specify in the documentation issue?

pshipton:
@sharon-wang the 0.15 release.

pshipton:
> is the jtreg test from the associated issue sufficient to test this option, or should a new test be added?

Please add new cmdline test(s) in the extended suite.

sharon-wang:
@pshipton I have added a new command line test - will need a review from the test team. Afterwards, PR should be ready to go.

sharon-wang:
@llxia Could you please review the test changes?

runtime/rasdump/dmpsup.c
```diff
@@ -577,7 +578,32 @@ configureDumpAgents(J9JavaVM *vm)
...
+					char *optionValue = NULL;
+					GET_OPTION_VALUE(xdumpIndex, '=', &optionValue);
+					
+					/* Construct optionString by combining the J9 tool dump command with the mapped option */
+					int toolCursorLength = strlen(toolCursor);
+					int optionValueLength = strlen(optionValue);
+					optionString = (char *) j9mem_allocate_memory(toolCursorLength + optionValueLength + 1, OMRMEM_CATEGORY_VM);
+
+					if (NULL == optionString) {
+						j9tty_err_printf(PORTLIB, "Storage for mapped tool dump optionString not available, unable to process option\n");
```
pshipton: This error would be better if it outputs the actual option that cannot be mapped, and the size of memory which is being allocated.

runtime/rasdump/dmpsup.c
```diff
@@ -559,6 +559,7 @@ configureDumpAgents(J9JavaVM *vm)
 	 */
 
 	xdumpIndex = FIND_ARG_IN_VMARGS_FORWARD(OPTIONAL_LIST_MATCH, VMOPT_XDUMP, NULL);
+	char *optionStringOriginal = NULL;
```
pshipton: This could be named better.

runtime/rasdump/dmpsup.c
```diff
@@ -577,7 +578,32 @@ configureDumpAgents(J9JavaVM *vm)
...
+					optionString = (char *) j9mem_allocate_memory(toolCursorLength + optionValueLength + 1, OMRMEM_CATEGORY_VM);
+
+					if (NULL == optionString) {
+						j9tty_err_printf(PORTLIB, "Storage for mapped tool dump optionString not available, unable to process option\n");
+						return J9VMDLLMAIN_FAILED;
+					} else {
+						strcpy(optionString, toolCursor);
+						strcat(optionString + toolCursorLength, optionValue);
+						/* Record original optionString pointer for freeing mapped tool dump optionString */
+						optionStringOriginal = optionString;
```
pshipton: This code is in a loop. If more than one option is mapped, there is a memory leak.

test/functional/cmdLineTests/cmdLineTest_J9tests/build.xml
```diff
@@ -30,16 +30,42 @@
...
+	<target name="dist" depends="compile" description="generate the distribution">
+		<jar jarfile="${DEST}/cmdLineTest_J9tests.jar" filesonly="true">
+			<fileset dir="${build}" />
+			<fileset dir="${src}" />
+		</jar>
 		<copy todir="${DEST}">
-			<fileset dir="${src}" includes="*.xml"/>
-			<fileset dir="${src}" includes="*.mk"/>
+			<fileset dir="${src}/../" includes="*.xml" />
+			<fileset dir="${src}/../" includes="*.mk" />
```
keithc-ca: Please use a separate property instead of using `..`.

test/functional/cmdLineTests/cmdLineTest_J9tests/src/OnOutOfMemoryErrorTest.java
```diff
@@ -0,0 +1,33 @@
...
+ * @file OnOutOfMemoryErrorTest.java
+ * @brief Generates a java.lang.OutOfMemoryError to test the command line option
+ *        -XX:OnOutOfMemoryError=
+ */
+
+class OnOutOfMemoryErrorTest {
+    public static void main(String[] args) {
+        Object test = new double[Integer.MAX_VALUE];
+    }
+}
```
keithc-ca: All text files should end with a new-line.

runtime/rasdump/dmpsup.c
```diff
@@ -714,6 +740,11 @@ configureDumpAgents(J9JavaVM *vm)
 	}
 	j9mem_free_memory(agentOpts);
 
+	/* Free optionString memory at original address for a mappped tool dump */
+	if (NULL != optionStringOriginal) {
+		j9mem_free_memory(optionStringOriginal);
```
keithc-ca: Arrange to mark `agentOpts.flags` with `J9RAS_DUMP_OPT_ARGS_ALLOC` so any number of the new options will be properly freed above.

runtime/rasdump/dmpsup.c
```diff
@@ -577,7 +578,32 @@ configureDumpAgents(J9JavaVM *vm)
 		}
 		if ( IS_CONSUMABLE(j9vm_args, xdumpIndex) && !IS_CONSUMED(j9vm_args, xdumpIndex) )
 		{
-			GET_OPTION_VALUE(xdumpIndex, ':', &optionString);
+			/* Handle mapped tool dump options */
+			if (HAS_MAPPING(j9vm_args, xdumpIndex)) {
+				char *toolCursor = strstr(MAPPING_J9NAME(j9vm_args, xdumpIndex), ":tool") + 1;
```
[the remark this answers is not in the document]
sharon-wang: For this line, `+ 1` moves the pointer past the colon before `"tool"`, so that it points to `"tool..."` instead of `":tool..."`. This is needed due to how group dump settings are parsed from `optionString`.

However, it prevents the null check from being false. I believe that moving the `+ 1` into the `if` clause should fix this.

runtime/rasdump/dmpsup.c
sharon-wang: @pshipton Would `optionStringCopy` work a bit better? I'm going for something that captures that it is meant to keep track of the original pointer.

pshipton: The mechanism doesn't work, as all mapped options need to be freed and not just the last one. But optionStringAllocated seems a better name.

sharon-wang: Will update name to `optionStringAllocated` and work on a mechanism that frees each mapped option.

pshipton: Note, if you go with what Keith suggested, you won't need this variable any more.

runtime/rasdump/dmpsup.c
```diff
@@ -603,7 +637,7 @@ configureDumpAgents(J9JavaVM *vm)
 						return J9VMDLLMAIN_SILENT_EXIT_VM;
 					}
 					agentOpts[agentNum].kind = kind;
-					agentOpts[agentNum].flags = J9RAS_DUMP_OPT_ARGS_STATIC;
+					agentOpts[agentNum].flags = isMappedToolDump ? J9RAS_DUMP_OPT_ARGS_ALLOC_TOOLMAP : J9RAS_DUMP_OPT_ARGS_STATIC;
```
sharon-wang: @keithc-ca @pshipton 
I have pushed a new approach to free the allocated memory for `optionString`. I'm not sure if this is the best approach as it introduces a new flag: https://github.com/eclipse/openj9/blob/22fe5f4ff24be7e1bd5a9f32d4d612609f2ccdc9/runtime/rasdump/rasdump_internal.h#L77

The reason for this flag is to mark each mapped tool option such that each can be freed accordingly. The original memory allocation for `optionString` is for `tool:events=systhrow,filter=java/lang/OutOfMemoryError,exec=java -version`: https://github.com/eclipse/openj9/blob/22fe5f4ff24be7e1bd5a9f32d4d612609f2ccdc9/runtime/rasdump/dmpsup.c#L593-L599

However, `optionString` is modified to point to `events=systhrow,filter=java/lang/OutOfMemoryError,exec=java -version` in the existing code to only capture the option section for the tool dump: https://github.com/eclipse/openj9/blob/22fe5f4ff24be7e1bd5a9f32d4d612609f2ccdc9/runtime/rasdump/dmpsup.c#L627-L631

As such, the pointer is incremented by 5 to skip over "tool:". Thus, to free the originally allocated memory, I have created the flag for this case so that we use the address originally used to allocate `optionString`, which is at`agentOpts[i].args - strlen("tool:")`: https://github.com/eclipse/openj9/blob/22fe5f4ff24be7e1bd5a9f32d4d612609f2ccdc9/runtime/rasdump/dmpsup.c#L743-L750

Your feedback would be greatly appreciated!

keithc-ca: Perhaps I'm missing something: if the string in `agentOpts[i].args` is allocated, why don't we simply put the right value there (after `tool:`) so we don't need a new flag?

sharon-wang: I think the string in `agentOpts[i].args` is not allocated in the general case (the flag is usually set to `J9RAS_DUMP_OPT_ARGS_STATIC`). 

The reason that the allocation in my implementation is made at `tool:` is because `typeString` is expected to hold all of `tool:events=systhrow,filter=java/lang/OutOfMemoryError,exec=java -version` and `optionString` is only `events=systhrow,filter=java/lang/OutOfMemoryError,exec=java -version`

keithc-ca: Actually, I think the answer is simple: `optionString` is the value returned by `j9mem_allocate_memory()` which is the same value you must pass to `j9mem_free_memory()`.

You don't *want* to backup by the length of "tool:" and so no new flag is required.

runtime/rasdump/dmpsup.c
```diff
@@ -589,6 +624,13 @@ configureDumpAgents(J9JavaVM *vm)
 					agentOpts[agentNum].pass = J9RAS_DUMP_OPTS_PASS_ONE;
 					agentNum++;
 				}
+			} else if (isMappedToolDump) {
+				char * toolString = "tool";
+				agentOpts[agentNum].kind = scanDumpType(&toolString);
+				agentOpts[agentNum].flags = J9RAS_DUMP_OPT_ARGS_ALLOC;
```
sharon-wang: @keithc-ca 
Here's the modified implementation based on our discussion, using the existing flag `J9RAS_DUMP_OPT_ARGS_ALLOC`. The memory allocation for `optionString` no longer includes "tool:".

runtime/rasdump/dmpsup.c
```diff
@@ -577,7 +577,42 @@ configureDumpAgents(J9JavaVM *vm)
 		}
 		if ( IS_CONSUMABLE(j9vm_args, xdumpIndex) && !IS_CONSUMED(j9vm_args, xdumpIndex) )
 		{
-			GET_OPTION_VALUE(xdumpIndex, ':', &optionString);
+			BOOLEAN isMappedToolDump = FALSE;
+			/* Handle mapped tool dump options */
+			if (HAS_MAPPING(j9vm_args, xdumpIndex)) {
+				char *mappingJ9Name = MAPPING_J9NAME(j9vm_args, xdumpIndex);
```
pshipton: Does it work to set VMOPT_XDUMP_TOOL_OUTOFMEMORYERROR_EXEC_EQUALS  to `"events=systhrow,filter=java/lang/OutOfMemoryError,exec="` and just omit the `-Xdump:tool:`?

runtime/rasdump/dmpsup.c
```diff
@@ -577,7 +577,42 @@ configureDumpAgents(J9JavaVM *vm)
...
+				char *toolString = ":tool";
```
pshipton: Since the trailing colon is required (for the +1 later), it should be added here.

runtime/rasdump/dmpsup.c
pshipton: We'd still need to know it's a tool mapping, but perhaps this can be added to the mapping definition.

sharon-wang: > Does it work to set VMOPT_XDUMP_TOOL_OUTOFMEMORYERROR_EXEC_EQUALS to `"events=systhrow,filter=java/lang/OutOfMemoryError,exec="` and just omit the `-Xdump:tool:`?

I think that I might need to keep `-Xdump:tool:` in the string, otherwise the option may not be picked up as a dump option in `dmpsup.c`, since `configureDumpAgents()` looks for arguments in `vmArgsArray` that contain `VMOPT_XDUMP` ("-Xdump") and consumes them accordingly.

runtime/rasdump/dmpsup.c
```diff
@@ -577,7 +577,42 @@ configureDumpAgents(J9JavaVM *vm)
...
+				char *mappingJ9Name = MAPPING_J9NAME(j9vm_args, xdumpIndex);
+				char *toolString = ":tool:";
+				char *toolCursor = strstr(mappingJ9Name, toolString);
+				if (NULL != toolCursor) {
+					char *optionValue = NULL;
+					/* The mapped option specifies the tool command to run after the equals */
+					GET_OPTION_VALUE(xdumpIndex, '=', &optionValue);
+					
+					/* Move toolCursor past ":tool:" */
+					toolCursor += strlen(toolString) + 1;
```
keithc-ca: I don't think you want the `+ 1` here: it would have the effect of pointing at `vents=systhrow...` in `-Xdump:tool:events=systhrow...`.

runtime/rasdump/dmpsup.c
```diff
@@ -577,7 +577,42 @@ configureDumpAgents(J9JavaVM *vm)
...
+
+					if (NULL != optionValue) {
+						int toolCursorLength = strlen(toolCursor);
+						int optionValueLength = strlen(optionValue);
+						int optionStringMemAlloc = toolCursorLength + optionValueLength + 1;
```
keithc-ca: These should use `size_t` instead of `int`; in the printf below, `%zu` should be used instead of `%d`.

test/functional/cmdLineTests/cmdLineTest_J9tests/build.xml
```diff
@@ -30,16 +30,43 @@
...
+		<mkdir dir="${DEST}" />
+		<mkdir dir="${build}" />
+	</target>
+
+	<target name="compile" depends="init" description="Using java ${JDK_VERSION} to compile the source ">
+		<echo>Ant version is ${ant.version}</echo>
+		<echo>============COMPILER SETTINGS============</echo>
+		<echo>===fork:				yes</echo>
+		<echo>===executable:			${compiler.javac}</echo>
+		<echo>===debug:				on</echo>
```
keithc-ca: Please don't use internal tabs here (only spaces after `<echo>`).

sharon-wang:
@keithc-ca Thanks for the review! I have made the requested changes and tested the option to verify that it is still working as expected.

test/functional/cmdLineTests/cmdLineTest_J9tests/build.xml
```diff
@@ -30,16 +30,43 @@
...
+		<mkdir dir="${build}" />
+	</target>
+
+	<target name="compile" depends="init" description="Using java ${JDK_VERSION} to compile the source ">
+		<echo>Ant version is ${ant.version}</echo>
+		<echo>============COMPILER SETTINGS============</echo>
+		<echo>===fork:                         yes</echo>
+		<echo>===executable:                   ${compiler.javac}</echo>
+		<echo>===debug:                        on</echo>
+		<echo>===destdir:                      ${DEST}</echo>
```
sharon-wang: @keithc-ca Tabs for the echo text have been switched to spaces. This whitespace change was added to each `build.xml` file in the `test/functional` directory with similar "COMPILER SETTINGS" echo messages.

sharon-wang: Whoops, need to fix the copyrights for each of the updated files.

keithc-ca: The spacing changes to files that were otherwise not modified for this PR should be made in a separate PR.

sharon-wang:
@keithc-ca Should be good to go - moved extra whitespace changes to a separate PR and tested the option again to confirm it's still working.

test/functional/cmdLineTests/cmdLineTest_J9tests/j9tests.xml
```diff
@@ -59,5 +64,12 @@ SPDX-License-Identifier: EPL-2.0 OR Apache-2.0 OR GPL-2.0 WITH Classpath-excepti
         <output type="success" caseSensitive="yes" regex="no">$NOREADIPINFOFORRAS_MESSAGE$</output>
         <output type="failure" caseSensitive="yes" regex="no">$READIPINFOFORRAS_MESSAGE$</output>
     </test>
+
+    <test id="test -XX:OnOutOfMemoryError=">
+        <command>$EXE$ $ONOUTOFMEMORYERROR_EQUALS$"java -version" $ONOUTOFMEMORYERROR_JAR$</command>
+        <output type="success" caseSensitive="yes" regex="no">$JAVALANGOUTOFMEMORYERROR$</output>
+        <output type="success" caseSensitive="yes" regex="no">$JVMREQUESTINGTOOLDUMP$</output>
```
pshipton: A cmdlinetester test passes if any of the success conditions occur, but in this case we only want the test to pass if all conditions occur. Use "success" for only one condition and "required" for the others.

keithc-ca:
Jenkins test extended plinux jdk8,jdk11

sharon-wang:
@pshipton @keithc-ca 
I have updated the test to use one "success" condition and "required" for the rest.

Testing for this option failed - will look into this before we run the test again.

test/functional/cmdLineTests/cmdLineTest_J9tests/src/OnOutOfMemoryErrorTest.java
```diff
@@ -0,0 +1,33 @@
...
+
+/**
+ * @file OnOutOfMemoryErrorTest.java
+ * @brief Generates a java.lang.OutOfMemoryError to test the command line option
+ *        -XX:OnOutOfMemoryError=
+ */
+
+class OnOutOfMemoryErrorTest {
+    public static void main(String[] args) {
+        Object test = new double[Integer.MAX_VALUE];
```
keithc-ca: I think this needs to be improved: with a large enough heap (>32GB) this might not throw OutOfMemoryError.
Either the command-line arguments must limit the heap to something smaller or this should allocate large objects in a loop (putting them in some collection).

sharon-wang: Would adding `-Xmx8G` be the correct command-line option to limit the heap to 8G?

keithc-ca: That should do it, but the heap doesn't need to be even that big - 1GB is probably overkill.

test/functional/cmdLineTests/cmdLineTest_J9tests/j9tests.xml
```diff
@@ -24,12 +24,18 @@ SPDX-License-Identifier: EPL-2.0 OR Apache-2.0 OR GPL-2.0 WITH Classpath-excepti
...
 <!DOCTYPE suite SYSTEM "cmdlinetester.dtd">
 
-<suite id="J9 Command-Line Option Tests" timeout="180">
+<suite id="J9 Command-Line Option Tests" timeout="600">
     <variable name="READIPINFOFORRAS" value="-XX:+ReadIPInfoForRAS"/>
     <variable name="NOREADIPINFOFORRAS" value="-XX:-ReadIPInfoForRAS"/>
     <variable name="READIPINFOFORRAS_MESSAGE" value="enabled network query to determine host name and IP address for RAS."/>
     <variable name="NOREADIPINFOFORRAS_MESSAGE" value="disabled network query to determine host name and IP address for RAS."/>
 
+    <variable name="JVM_HEAP_LIMIT" value="-Xmx8G"/>
```
pshipton: Please use a smaller limit. Some machines may not have this amount of memory. -Xmx512k should be fine.

pshipton: Typo in my previous comment, which I edited. -Xmx512m

sharon-wang:
I've set the heap limit to `-Xmx512m` and increased the test timeout to "600" - I believe the timeout is what caused the test to fail, since the previous limit was "160" and the test ran for about "260". Running personal tests to confirm that everything is still working as intended.

sharon-wang:
@keithc-ca Test passed in my personal runs. Should be good to go!

keithc-ca:
Jenkins test extended plinux jdk8,jdk11

pshipton:
jenkins test extended win jdk8

PR #5245, opened 2019-03-26
Remove TR_EnableJITaaSRemoteAOT Env Flag

a7ehuo:
[skip ci]
`TR_EnableJITaaSRemoteAOT` env flag is removed from
both client and server. This change requires SVM support
in JITaaS #5158.

### DIFF
--- a/runtime/compiler/control/CompilationThread.cpp
+++ b/runtime/compiler/control/CompilationThread.cpp
@@ -6749,6 +6749,20 @@ bool isMethodIneligibleForAot(J9Method *method)
    return isMethodIneligibleForAot(J9_ROM_METHOD_FROM_RAM_METHOD(method), J9_CLASS_FROM_METHOD(method)->romClass);
    }
 
+bool
+TR::CompilationInfoPerThreadBase::shouldPerformLocalComp(const TR_MethodToBeCompiled *entry)
+   {
+      bool doLocalComp = false;
+      static char *localColdCompilations = feGetEnv("TR_LocalColdCompilations");
+      // As a heuristic, cold compilations should be performed locally because
+      // they are supposed to be cheap with respect to memory and CPU.
+      //
+      if (entry->_optimizationPlan->getOptLevel() <= cold &&
+         (TR::Options::getCmdLineOptions()->getOption(TR_EnableJITaaSHeuristics) || localColdCompilations))
+         doLocalComp = true;
+
+      return doLocalComp;
+   }
 
 /**
  * @brief TR::CompilationInfoPerThreadBase::preCompilationTasks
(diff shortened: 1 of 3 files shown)


runtime/compiler/control/CompilationThread.cpp
```diff
@@ -6867,8 +6880,16 @@ TR::CompilationInfoPerThreadBase::preCompilationTasks(J9VMThread * vmThread,
       //
       TR::IlGeneratorMethodDetails & details = entry->getMethodDetails();
 
-      eligibleForRelocatableCompile =
-            (enableJITaaSRemoteAOT || !entry->isOutOfProcessCompReq()) && // disable remote JITaaS AOT until it's working
+      if (_compInfo.getPersistentInfo()->getJITaaSMode() == SERVER_MODE)
```
mpirvu: rather than testing `_compInfo.getPersistentInfo()->getJITaaSMode() == SERVER_MODE` it's nicer to test entry->isOutOfProcessCompReq()

runtime/compiler/control/CompilationThread.cpp
```diff
@@ -6867,8 +6880,16 @@ TR::CompilationInfoPerThreadBase::preCompilationTasks(J9VMThread * vmThread,
...
+         {
+         // Since all of the preliminary checks have already been done at the client,
+         // eligibleForRelocatableCompile should be true at the server side if the client
+         // requests AOT.
+         eligibleForRelocatableCompile = (entry->_useAotCompilation) ? true : false;
```
mpirvu: `eligibleForRelocatableCompile = entry->_useAotCompilation`  since `_useAotCompilation` is already a bool

runtime/compiler/control/CompilationThread.cpp
```diff
@@ -6918,10 +6940,15 @@ TR::CompilationInfoPerThreadBase::preCompilationTasks(J9VMThread * vmThread,
    // Note that currently we don't support out-of-process relocatable compilations
    if (canDoRelocatableCompile)
       {
-      _vm = TR_J9VMBase::get(_jitConfig, vmThread, TR_J9VMBase::AOT_VM);
       entry->_useAotCompilation = true;
 
-      if (_compInfo.getPersistentInfo()->getJITaaSMode() == CLIENT_MODE && enableJITaaSRemoteAOT)
+      if (_compInfo.getPersistentInfo()->getJITaaSMode() == SERVER_MODE)
```
mpirvu: The comment `// Note that currently we don't support out-of-process relocatable compilations` needs to be deleted

a7ehuo: Should testing `_compInfo.getPersistentInfo()->getJITaaSMode() == SERVER_MODE` be changed to `entry->isOutOfProcessCompReq()` here as well?

mpirvu: yes, since we have a valid entry we can issue the test based on entry

a7ehuo:
@mpirvu Updated. Please check. Thanks.

PR #5253, opened 2019-03-26
Invoke Read Barrier during monitor cache lookup check

amicic:
During the check if a cached monitor's object matches the input object,
when accessing monitor table object slot, use Read Barrier, so that
object address is properly updated, if need.

No stability issue was observed due to the lack of this barrier. This
update should just marginally improve cache hit ratio (if Concurrent
Scavenger cycle is active and if monitor's object slots wasn't already
updated during the cycle).

### DIFF
--- a/runtime/vm/montable.c
+++ b/runtime/vm/montable.c
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 1991, 2017 IBM Corp. and others
+ * Copyright (c) 1991, 2019 IBM Corp. and others
  *
  * This program and the accompanying materials are made available under
  * the terms of the Eclipse Public License 2.0 which accompanies this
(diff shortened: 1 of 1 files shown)


amicic:
@DanHeidinga ping

runtime/vm/montable.c
```diff
@@ -265,7 +265,7 @@ monitorTableAt(J9VMThread* vmStruct, j9object_t object)
 #else
 	objectMonitor = vmStruct->cachedMonitor;
 #endif
-	if ((objectMonitor != NULL) && (((J9ThreadAbstractMonitor*)objectMonitor->monitor)->userData == (UDATA) object)) {
+	if ((objectMonitor != NULL) && (J9MONITORTABLE_OBJECT_LOAD_VM(vm, &((J9ThreadAbstractMonitor*)objectMonitor->monitor)->userData) == object)) {
```
DanHeidinga: I see a couple of possible cases here:

 Object | Monitor->userData | hit or miss
--------|--------------------|--------------
 Not Forwarded | Not Forwarded | hit
 Not Forwarded | Forwarded | miss but found in hashtable
 Forwarded | Not Forwarded | miss but found in hashtable
 Forwarded | Forwarded | hit

Is there any evidence to show the cache hit rate is better with this change?

Given the low cost to do the check, is it worthwhile to check both:
```c
if ((objectMonitor != NULL) 
&& ( (((J9ThreadAbstractMonitor*)objectMonitor->monitor)->userData == (UDATA) object)
|| (J9MONITORTABLE_OBJECT_LOAD_VM(vm, &((J9ThreadAbstractMonitor*)objectMonitor->monitor)->userData) == object) )
) {
```

amicic: First two cases are not possible (assuming an object was originally in Evacuate set, as opposed to Tenured object). Object (in context of a mutator thread) is guarantied to be already forwarded. The value is obtained either from a root slot (fixed by GC at the cycle start) or from a heap object which has  been fixed by either this thread via read barrier or by a  GC background thread or some other thread during the GC cycle (there are even some more indirections possible combining recursively already mentioned scenarios with transferring the value form other already fixed slots).

Currently, last 2 cases are possible. but with this change, only the last one will be possible, what is the only valid one.

I did some simple counting with temp stats and there was a few more hits with the fix, on a non-trivial workload, but the workload did not have performance metric, so I could not tell if there is any real life benefit.

DanHeidinga: If you can add some of this context as a comment to the code, I'm OK with merging this

amicic: Added a comment

DanHeidinga:
Jenkins test sanity zlinux,xlinux jdk8

PR #5260, opened 2019-03-26
Use spaces instead of tabs in module-info files

keithc-ca:
* tabs are not supported by the dependency discovery mechanism in openjdk (since ibmruntimes/openj9-openjdk-jdk10#17).


pshipton:
jenkins compile zlinux jdk11

pshipton:
@keithc-ca Please create a PR for this on the 0.14 branch.

pdbain-ibm:
Would it be helpful to add a pull request check for tabs in module-info files, since we are unlikely to find it them in normal testing?

PR #5263, opened 2019-03-26
Fix the duplicated NLS messages when SH_OSCache::getCacheDir() failed.

hangshao0:
1. Add a new input parameter "allowVerbose" to SH_OSCache::getCacheDir()
2. Always call SH_OSCache::getCacheDir() with allowVerbose = false in
j9shr_getCacheDir()

related to #5134 and #5151 

### DIFF
--- a/runtime/shared_common/shrinit.cpp
+++ b/runtime/shared_common/shrinit.cpp
@@ -4908,7 +4908,8 @@ isFreeDiskSpaceLow(J9JavaVM *vm, U_64* maxsize)
 	bool ret = true;
 	PORT_ACCESS_FROM_JAVAVM(vm);
 
-	if (-1 == SH_OSCache::getCacheDir(vm, vm->sharedCacheAPI->ctrlDirName, cacheDirName, J9SH_MAXPATH, J9PORT_SHR_CACHE_TYPE_PERSISTENT)) {
+	if (-1 == j9shr_getCacheDir(vm, vm->sharedCacheAPI->ctrlDirName, cacheDirName, J9SH_MAXPATH, J9PORT_SHR_CACHE_TYPE_PERSISTENT)) {
+		/* use j9shr_getCacheDir() instead of SH_OSCache::getCacheDir() to avoid dupliated NLS message if SH_OSCache::getCacheDir() failed */
 		Trc_SHR_INIT_isFreeDiskSpaceLow_getDirFailed();
 		goto done;
 	}
@@ -5140,7 +5141,7 @@ j9shr_findGCHints(J9VMThread* currentThread, UDATA *heapSize1, UDATA *heapSize2)
 IDATA
 j9shr_getCacheDir(J9JavaVM* vm, const char* ctrlDirName, char* buffer, UDATA bufferSize, U_32 cacheType)
 {
-	return SH_OSCache::getCacheDir(vm, ctrlDirName, buffer, bufferSize, cacheType);
+	return SH_OSCache::getCacheDir(vm, ctrlDirName, buffer, bufferSize, cacheType, false);
 }
 
 } /* extern "C" */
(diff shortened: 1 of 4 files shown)


hangshao0:
Manually tested:

```
java -Xshareclasses:name=Cache1 -version
JVMSHRC831E Failed to use user's home as the default shared cache directory. It is on a Network File System. Cannot get home directory. Please set another directory via environment variable "HOME" or command line option "cacheDir=", or fix the home directory in the password file entry.
JVMJ9VM015W Initialization error for library j9shr29(11): JVMJ9VM009E J9VMDllMain failed
Error: Could not create the Java Virtual Machine.
Error: A fatal exception has occurred. Program will exit.

java -Xshareclasses:name=Cache1,silent -version
JVMJ9VM015W Initialization error for library j9shr29(11): JVMJ9VM009E J9VMDllMain failed
Error: Could not create the Java Virtual Machine.
Error: A fatal exception has occurred. Program will exit.

java -Xshareclasses:name=Cache1,printStats
JVMSHRC831E Failed to use user's home as the default shared cache directory. It is on a Network File System. Cannot get home directory. Please set another directory via environment variable "HOME" or command line option "cacheDir=", or fix the home directory in the password file entry.

java -Xshareclasses:name=Cache1,nonfatal HelloWorld
JVMSHRC831E Failed to use user's home as the default shared cache directory. It is on a Network File System. Cannot get home directory. Please set another directory via environment variable "HOME" or command line option "cacheDir=", or fix the home directory in the password file entry.
JVMSHRC686I Failed to startup shared class cache. Continue without using it as -Xshareclasses:nonfatal is specified
Hello World !
```

runtime/shared_common/shrinit.cpp
```diff
@@ -3157,6 +3159,21 @@ j9shr_init(J9JavaVM *vm, UDATA loadFlags, UDATA* nonfatal)
 		*nonfatal = 1;
 	}
 
+	if (RESULT_DO_HELP != parseResult && RESULT_DO_MORE_HELP != parseResult) {
+		if (-1 == SH_OSCache::getCacheDir(vm, ctrlDirName, cacheDirName, J9SH_MAXPATH, cacheType)) {
```
pshipton: I'd prefer not to add another call to getCacheDir() in the startup path. The code should be optimal to provide the fastest startup possible.

hangshao0:
Marking this as WIP for now.

hangshao0:
This PR was closed by accident, re-opening it.

hangshao0:
Removing WIP from the title. 

I didn't add a new variable to cache the result of getCacheDir(). We already have a field in SH_OSCache that saves the directory: 
https://github.com/eclipse/openj9/blob/127ba34f1b6b2338beeed7ca1cf2e393640fae66/runtime/shared_common/OSCache.hpp#L304

The variable is set in SH_OSCache::commonStartup(). We are using _cacheDirName to get the cache directory after that. There are cases where we called getCacheDir() before SH_OSCache::commonStartup().
 
When using non-default directory, SH_OSCache::getCacheDir() is called at most 3 times before commonStartup(). But getCacheDir() is quite cheap in this case, it essentially uses omrstr_printf() to return the directory. When using default directory,  SH_OSCache::getCacheDir() is called at most twice. I think there is no need increasing the footprint to add another variable to cache the result just for this one extra call.

All the other callers are from share classes utilities.

runtime/shared_common/shrinit.cpp
hangshao0: This code is removed.

pshipton:

pshipton:

pshipton:
@hangshao0 I'd like to add this fix to the 0.14.2 release. Without this fix an "error" can occur even when a cacheDir is specified, when the default home directory is not suitable. Can you please create a PR against the v0.14.2-release branch.

hangshao0:
PR against v0.14.2: https://github.com/eclipse/openj9/pull/5770

Issue #4837, opened 2019-02-22
jsr292 performance investigation 


cathyzhyi:
OpenJ9 does not perform very well with programs heavily rely on jsr292. This issue is for tracking investigation into the performance issue.

1.  Swapping targeted inlining with normal inlining 
       Targeted inlining was done before normal inlining and won't do any thing if the jsr292 methdos were called from a inlined callee. Swapping to do normal inlining first so that jsr292 methods called from callees can be inlined as well.

2.  add small recongnized jsr292 methods as inlineable target for both targeted inlining and normal inlining:
`java_lang_invoke_MethodHandle_type`  
`java_lang_invoke_ArgumentMoverHandle_extra_L`
`java_lang_invoke_BruteArgumentMoverHandle_extra_L`

3. MutableCallSite target inlining. There are several problems with mutable callsite
        a. `MutableCallSiteTargetGuard` exists already for inlining mutable callsite and if a method handle is got from `MutableCallSite.getTarget()` or `MutableCallSite.target` the guard is created and the methodhandle chain can be inlined. HCR guard breaks the pattern by storing the result into a temp when inlining `MutableCallSite.getTarget`. Need to find a way to keep the pattern so that we can inline mutable callsite target. One cleaner way would be changing the `MutableCallSite.target` field to package access so that we don't need to access the field through function call.
        b. once the target is inlined, propagate known obj info down the call chain.

cathyzhyi:
Tried out prototype changes for item 1-4  in description on JRuby benchmarks from https://github.com/headius/bench2018. Benchmark names are bold and some benchmarks have subbenchmarks. `withchange` includes item 2-4. `withchangeinlineathot` includes item 1-3. `withchangeswap` includes item 1-4  especially adding swapping normal inlining and targeted inlining compared to `withchange`. Swapping shows improvement in most benchmarks but hurts red_black and mandelbrot because leaf methods are not inlined after targeted inlining. `hash` has 5 subbenchmarks and seems to have big fluctuation with the new changes and need further investigation.

| **red_black**    	|         	|          	|               	|                       	|                	|
|------------------	|---------	|----------	|---------------	|-----------------------	|----------------	|
| LOWER is better  	|         	|          	|               	|                       	|                	|
| config           	| hotspot 	| baseline 	| withchange    	| withchangeinlineathot 	| withchangeswap 	|
[...]

PR #5279, opened 2019-03-27
Set prexArg for method handle thunks

cathyzhyi:
The receiver of a method handle thunk is always a known object. Set
prexArg for methodhandle thunks in all cases to propagate the info
without peeking in inlining or global VP.

In the special case of a method handle thunk being  mutable callsite
target, the known object info is only available in the inlined body
rather than taken side of the guard.

issue: #4837

### DIFF
--- a/runtime/compiler/optimizer/InlinerTempForJ9.cpp
+++ b/runtime/compiler/optimizer/InlinerTempForJ9.cpp
@@ -4837,7 +4837,9 @@ TR_J9InlinerUtil::computePrexInfo(TR_CallTarget *target)
                }
             continue;
             }
-         else if (target->_guard->_kind == TR_MutableCallSiteTargetGuard)
+         else if (target->_calleeSymbol->getResolvedMethod()->convertToMethod()->isArchetypeSpecimen()
+                  && target->_calleeSymbol->getResolvedMethod()->getMethodHandleLocation()
+                  && comp()->getOrCreateKnownObjectTable())
             {
             // Here's a situation where inliner is taking it upon itself to draw
             // conclusions about known objects.  VP won't get a chance to figure this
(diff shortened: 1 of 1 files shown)


andrewcraik:
@cathyzhyi is something missing? The CI build failed with
```../../compiler/optimizer/InlinerTempForJ9.cpp: In member function ‘void TR_J9TransformInlinedFunction::createKnownObjReceiverTemp(TR_VirtualGuardSelection*)’:
../../compiler/optimizer/InlinerTempForJ9.cpp:5481:7: error: ‘_knownObjReceiverSymRef’ was not declared in this scope
       _knownObjReceiverSymRef = knownObjSymRef;
       ^
```

cathyzhyi:
This change is dependent on https://github.com/eclipse/omr/pull/3700

cathyzhyi:
This change is dependent on https://github.com/eclipse/omr/pull/3700

cathyzhyi:
@vijaysun-omr this one is ready to be merged

vijaysun-omr:
I understand the first part of the rationale for the change. Could you please elaborate on :
"In the special case of a method handle thunk being mutable callsite target, the known object info is only available in the inlined body rather than taken side of the guard." ?

cathyzhyi:
> I understand the first part of the rationale for the change. Could you please elaborate on :
> "In the special case of a method handle thunk being mutable callsite target, the known object info is only available in the inlined body rather than taken side of the guard." ?

For mutable callsite, the inlined method handle thunk is guarded by `MutableCallSiteTargetGuard` in case the call site target, which is also the method handle object, changed. The known object info here only applies to the inlined body and needs special handling so that when inliner replaces the arguments with priv arg temps it only use known object temps on the inlined body but not the taken side.

cathyzhyi:
@vijaysun-omr could you help to launch `Jenkins test sanity xlinux,win,plinux jdk8,jdk11` tests? Thanks!!

vijaysun-omr:

PR #5282, opened 2019-03-27
Cmdline option to disable CopyForwardMarkCompactHybrid GC

LinHu2016:
- add XXgc:tarokDisableCopyForwardMarkCompactHybrid option for
	disabling Hybrid mode.


amicic:
Change the subject (this change is not really enabling the hybrid mode).

LinHu2016:
@amicic the change for enabling by default is in omr code, https://github.com/eclipse/omr/pull/3701

amicic:
I understand. I'm just saying that the subject (title) of this PR should be something like: "Cmdline option to disable hybrid copy/mark GC."

PR #5284, opened 2019-03-27
Return NULL instead of throwing when the class name is invalid

keithc-ca:
Prior to #4919, ClassLoader.defineClassImpl() would check if the given
class name and immediately throw NoClassDefFoundError if it was invalid.
Beginning with Java 12, JavaLangAccess.defineClass() is used to define
accessor classes (with invalid names). The name validation must still
be done, but throwing the exception must be delayed until it can be
determined to be the appropriate response. The invalid name is signalled
with a bit in the options parameter to defineClassCommon() which must
not set a pending exception in that case.

### DIFF
--- a/runtime/jcl/common/jcldefine.c
+++ b/runtime/jcl/common/jcldefine.c
@@ -56,14 +56,18 @@ defineClassCommon(JNIEnv *env, jobject classLoaderObject,
 	J9TranslationLocalBuffer localBuffer = {J9_CP_INDEX_NONE, LOAD_LOCATION_UNKNOWN, NULL};
 
 	if (vm->dynamicLoadBuffers == NULL) {
-		throwNewInternalError(env, "Dynamic loader is unavailable");
-		return NULL;
+		if (J9_ARE_NO_BITS_SET(options, J9_FINDCLASS_FLAG_NAME_IS_INVALID)) {
+			throwNewInternalError(env, "Dynamic loader is unavailable");
+		}
+		goto done;
 	}
 	dynFuncs = vm->dynamicLoadBuffers;
 
 	if (classRep == NULL) {
-		throwNewNullPointerException(env, NULL);
-		return NULL;
+		if (J9_ARE_NO_BITS_SET(options, J9_FINDCLASS_FLAG_NAME_IS_INVALID)) {
+			throwNewNullPointerException(env, NULL);
+		}
+		goto done;
 	}
 
 	vmFuncs->internalEnterVMFromJNI(currentThread);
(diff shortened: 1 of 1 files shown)


pshipton:
jenkins test sanity plinux jdk8,jdk11,jdk12

runtime/jcl/common/jcldefine.c
```diff
@@ -211,18 +212,25 @@ defineClassCommon(JNIEnv *env, jobject classLoaderObject,
...
-	}
-			
-	result = vmFuncs->j9jni_createLocalRef(env, J9VM_J9CLASS_TO_HEAPCLASS(clazz));
+	if (clazz == NULL) {
+		if (J9_ARE_ANY_BITS_SET(options, J9_FINDCLASS_FLAG_NAME_IS_INVALID)) {
+			/*
+			 * The caller signalled that the name is invalid. Leave the result NULL and
+			 * clear any pending exception; the caller will throw NoClassDefFoundError.
+			 */
+			currentThread->currentException = NULL;
```
gacholio: This should clear the privateFlags bit as well:

https://github.com/eclipse/openj9/blob/dde0f75a7bf23f7203499db658c1561c26b87e4c/runtime/oti/VMHelpers.hpp#L637

runtime/jcl/common/jcldefine.c
```diff
@@ -211,18 +212,25 @@ defineClassCommon(JNIEnv *env, jobject classLoaderObject,
 	}
 
 done:
-	if ((clazz == NULL) && (currentThread->currentException == NULL)) {
-		/* should not get here -- throw the default exception just in case */
-		vmFuncs->setCurrentException(currentThread, J9VMCONSTANTPOOL_JAVALANGCLASSFORMATERROR, NULL);
-	}
-			
-	result = vmFuncs->j9jni_createLocalRef(env, J9VM_J9CLASS_TO_HEAPCLASS(clazz));
+	if (clazz == NULL) {
```
gacholio: Since I've request a real change below, it would be nice to fix this to `NULL ==`, and the one below as well.

runtime/jcl/common/jcldefine.c
keithc-ca: Thanks; fixed.

runtime/jcl/common/jcldefine.c
keithc-ca: Done.

keithc-ca:
Updated based on review feedback.

keithc-ca:
Jenkins test sanity plinux jdk8,jdk11,jdk12

pshipton:
@keithc-ca please create a PR for the 0.14 branch.

Issue #5264, opened 2019-03-26
Implement `TR::SymbolValidationManager::skipFieldRefClassRecord` for JITaaS


dmitry-ten:
This method indexes into ROM constant pool, so need to create a new remote call for that.
Didn't include it in my original SVM PR #5158, to avoid making the change even larger.

dmitry-ten:
@mpirvu fyi

PR #5292, opened 2019-03-28
Enable SymbolValidationManager Part 2

dmitry-ten:
Created a new J9ClassEnv method to fetch ROM class ref name
from the client.

Fixes: #5264


dmitry-ten:
Hm, we could return `uint8_t *` - a pointer to `J9UTF8` data, instead of string. That would avoid copying for baseline. But for JITaaS we'd have to heap-allocate the space.

dmitry-ten:
On a second thought, that would mean we'd have to manually free memory every time we call this method on the server, but not on the client, which is less than ideal.
Maybe instead should separate `getROMClassRefName` into
`getROMClassRefNameLength` and `getROMClassRefNameChars`. Then we'd know how big the name is before fetching it. Would require 2 remote calls though.

dmitry-ten:
Updated with what I said above.
@ymanton this is ready for another review.

PR #5200, opened 2019-03-21
Missing Symref for SIMD duplicated Node

R-Santhir:
In cases where a loop is a candidate for vectorization and an induction
variable is accessed in the loop, any nodes that access the induction variable
must be duplicated. There are situations where this duplicated node will not
have a correct symbol reference, and must be created after the node is
duplicated.

Signed-off-by: Ryan Santhirarajan <rsanth@ca.ibm.com>

PR #5309, opened 2019-03-31
Missing Symref for SIMD duplicated Node

R-Santhir:
In cases where a loop is a candidate for vectorization and an induction
variable is accessed in the loop, any nodes that access the induction variable
must be duplicated. There are situations where this duplicated node will not
have a correct symbol reference, and must be created after the node is
duplicated.

Port of #5200


R-Santhir:
@gita-omr @zl-wang

gita-omr:
@pshipton for some reason ci build says v0.14.0-release does not exist and we really want to merge this fix into it.

pshipton:
This is the same change as #5200, but for the 0.14 branch.

PR #5314, opened 2019-04-01
Remove stale allocation fence

liqunl:
StringPeepHoles deletes `new` node of StringBuilder/StringBuffer but
leaves allocation fence in place. The stale allocation fence will cause
failure in later attempt to remove pending push temps. This change fixes
it.


liqunl:
https://github.com/eclipse/openj9/pull/5306
@andrewcraik @pshipton

PR #5324, opened 2019-04-02
Fix issue relate with CollectionUsageThreshold of memory pool

LinHu2016:
- we should only check if the notification for
	CollectionUsageThreshold need to be fired at the end of
	the collection, which try to recycle the memory pool.
	- for BackCompatible mode scavenge does not try to reclaim
	memory from the whole heap, so we do not fire the notification of
	CollectionUsageThreshold at the end scavenge.
	- record the last GC ID for retrieving more detail gcInfo of
	last GC from J9GarbageCollectorDataMemoryPoolMXBean
	.getCollectionUsage() getPreCollectionUsage() and
	isCollectionUsageThresholdExceeded() also match the same rule.
       - fix potential IllegalArgumentException issue.

relate:https://github.com/eclipse/openj9/pull/2993

### DIFF
--- a/runtime/jcl/common/mgmtinit.c
+++ b/runtime/jcl/common/mgmtinit.c
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 1998, 2019 IBM Corp. and others
+ * Copyright (c) 1998, 2020 IBM Corp. and others
  *
  * This program and the accompanying materials are made available under
  * the terms of the Eclipse Public License 2.0 which accompanies this
@@ -481,6 +481,7 @@ gcStartEvent(J9JavaVM *vm, UDATA heapSize, UDATA heapUsed, UDATA *totals, UDATA 
 	J9MemoryPoolData *memoryPools = mgmt->memoryPools;
 	J9MemoryNotification *notification = NULL;
 	J9MemoryNotification *last = NULL;
+	J9GarbageCollectionInfo* gcInfo = NULL;
 
 	UDATA total = 0;
 	UDATA used = 0;
@@ -492,12 +493,13 @@ gcStartEvent(J9JavaVM *vm, UDATA heapSize, UDATA heapUsed, UDATA *totals, UDATA 
 
 	/* check the wall clock */
 	/* the start time in milliseconds since the Java virtual machine was started */
-	gcData->lastGcInfo.startTime = j9time_current_time_millis();
-	if (gcData->lastGcInfo.startTime < (U_64)mgmt->vmStartTime) {
+	gcInfo = &gcData->lastGcInfo;
+	gcInfo->startTime = j9time_current_time_millis();
+	if (gcInfo->startTime < (U_64)mgmt->vmStartTime) {
 		/* startTime is earlier than vmStartTime in case of wall clock correction while interval is measuring */
-		gcData->lastGcInfo.startTime = 0;
+		gcInfo->startTime = 0;
 	} else {
-		gcData->lastGcInfo.startTime -= mgmt->vmStartTime;
+		gcInfo->startTime -= mgmt->vmStartTime;
 	}
 
 	mgmt->preCollectionHeapSize = heapSize;
@@ -508,15 +510,25 @@ gcStartEvent(J9JavaVM *vm, UDATA heapSize, UDATA heapUsed, UDATA *totals, UDATA 
 		total = totals[idx];
 		used = totals[idx] - frees[idx];
 
-		memoryPool->preCollectionSize = total;
-		memoryPool->preCollectionUsed = used;
-		memoryPool->preCollectionMaxSize = memoryPool->postCollectionMaxSize;
+		/* update pre Memory Usage of last GcInfo for the collector */
+		gcInfo->preUsed[idx] = used;
+		gcInfo->preCommitted[idx] = total;
+
+		gcInfo->preMax[idx] = 0;
+		if (0 != mgmt->lastGCID) {
+			J9GarbageCollectorData *lastGcData = &mgmt->garbageCollectors[getArrayIndexFromManagerID(mgmt, mgmt->lastGCID)];
+			J9GarbageCollectionInfo* lastGcInfo = &lastGcData->lastGcInfo;
+			gcInfo->preMax[idx] = lastGcInfo->postMax[idx];
+		}
+		if (0 == gcInfo->preMax[idx]) {
+			gcInfo->preMax[idx] = memoryPool->postCollectionMaxSize;
+		}
 
 		/* check the peak usage and update */
 		if (memoryPool->peakUsed < used) {
 			memoryPool->peakUsed = used;
 			memoryPool->peakSize = total;
-			memoryPool->peakMax = memoryPool->preCollectionMaxSize;
+			memoryPool->peakMax = gcInfo->preMax[idx];
 		}
 
 		/* if a heap usage threshold is set, check whether we are above or below */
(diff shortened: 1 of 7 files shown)


pshipton:

pshipton:

pshipton:
jenkins compile win jdk8

runtime/jcl/common/mgmtinit.c
```diff
@@ -502,21 +503,32 @@ gcStartEvent(J9JavaVM *vm, UDATA heapSize, UDATA heapUsed, UDATA *totals, UDATA
 
 	mgmt->preCollectionHeapSize = heapSize;
 	mgmt->preCollectionHeapUsed = heapUsed;
+	gcInfo = &gcData->lastGcInfo;
```
DanHeidinga: Should this be done immediate after acquiring the lock so the code above consistently uses the local `gcInfo` pointer rather than sometimes using `gcData->lastGcInfo` still?

I found it confusing to have the two references to the same thing intermixed

runtime/jcl/common/mgmtinit.c
```diff
@@ -502,21 +503,32 @@ gcStartEvent(J9JavaVM *vm, UDATA heapSize, UDATA heapUsed, UDATA *totals, UDATA
...
 	for (idx = 0; idx < mgmt->supportedMemoryPools; ++idx) {
 		J9MemoryPoolData *memoryPool = &memoryPools[idx];
 		total = totals[idx];
 		used = totals[idx] - frees[idx];
 
-		memoryPool->preCollectionSize = total;
-		memoryPool->preCollectionUsed = used;
-		memoryPool->preCollectionMaxSize = memoryPool->postCollectionMaxSize;
+		/* update pre Memory Usage of last GcInfo for the collector */
+		gcInfo->preUsed[idx] = used;
```
DanHeidinga: why doesn't the `memorypool` `pre*` values get updated here anymore?  The peak values till get updated below

LinHu2016:
> I'm having some trouble understanding this change as there's 1 gcInfo and potentially many memoryPools and this change seems to be collecting the data in the gcInfo and then applying it to the memoryPools.
> 
> I don't understand why its being changed this way rather than keeping the data in the memory pools and agregating it in the gcInfo.
> 
> Once I understand that, I'll be in a better position to review this.

Hi @DanHeidinga,  according to JAVA Spec, preCollection(IBM Java only), postCollection Memory Usage in memoryPool only keep information for the GC, which recycle the memory pool. For the case gencon, Tenure memorypool would only update after global gc, scavenge does not reclaim tenure memorypool( actually it consumes tenure space). 

for the details, please check java doc for MemoryPoolMXBean.getCollectionUsage()
https://docs.oracle.com/javase/7/docs/api/java/lang/management/MemoryPoolMXBean.html#getCollectionUsage()

runtime/jcl/common/mgmtmempool.c
```diff
@@ -352,7 +365,13 @@ Java_com_ibm_java_lang_management_internal_MemoryPoolMXBeanImpl_resetPeakUsageIm
 		omrthread_rwmutex_enter_write(mgmt->managementDataLock);
 		pool->peakUsed = used;
 		pool->peakSize = committed;
-		pool->peakMax = pool->postCollectionMaxSize;
+		if (0 != mgmt->lastGCID) {
```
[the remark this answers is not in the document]
LinHu2016: Thanks Dan, plan to create new vm/gc api to return both used,committed and max value of memorypool memory usage.

pshipton:
@DanHeidinga I note there are new changes pushed.

pshipton:
@LinHu2016 is this ready for re-review?

LinHu2016:
@pshipton yes, the new update is ready for reviewing.

pshipton:
@DanHeidinga fyi

pshipton:
I don't expect this is going to be ready for the 0.15 release.

pshipton:
Getting late to put into the 0.17 release, moving forward again.

pshipton:
@DanHeidinga any plans to review this before the 0.18 release, or should it be moved to the next?

DanHeidinga:
Will try and get through it this week (and apologies on the delay here)

runtime/jcl/common/mgmtmempool.c
```diff
@@ -173,14 +178,21 @@ Java_com_ibm_java_lang_management_internal_MemoryPoolMXBeanImpl_getUsageImpl(JNI
...
 
-		javaVM->memoryManagerFunctions->j9gc_pools_memory(javaVM, (id&J9VM_MANAGEMENT_POOL_HEAP_ID_MASK), &totals[0], &frees[0], FALSE);
-		committed = totals[idx];
-		used = committed - frees[idx];
+		UDATA total = 0;
+		UDATA free = 0;
+		UDATA maximum = 0;
+
+		/* acquire vmAccess for retrieving memoryUsage for the pool in order to guarantee values of memoryUsage are consistent. */
+		javaVM->internalVMFunctions->internalAcquireVMAccess(vmThread);
```
DanHeidinga: I think this should be the internalEnterVMFromJNIcall instead

runtime/jcl/common/mgmtmempool.c
```diff
@@ -333,26 +345,32 @@ Java_com_ibm_java_lang_management_internal_MemoryPoolMXBeanImpl_isUsageThreshold
...
 
-		javaVM->memoryManagerFunctions->j9gc_pools_memory(javaVM, (id & J9VM_MANAGEMENT_POOL_HEAP_ID_MASK), &totals[0], &frees[0], FALSE);
-		committed = totals[idx];
-		used = committed - frees[idx];
+		UDATA total = 0;
+		UDATA free = 0;
+		UDATA maximum = 0;
+
+		/* acquire vmAccess for retrieving memoryUsage for the pool in order to guarantee values of memoryUsage are consistent. */
+		javaVM->internalVMFunctions->internalAcquireVMAccess(vmThread);
```
DanHeidinga: I think this should be the internalEnterVMFromJNIcall instead

runtime/jcl/common/mgmtmempool.c
LinHu2016: done

runtime/jcl/common/mgmtmempool.c
```diff
@@ -109,12 +110,17 @@ Java_com_ibm_java_lang_management_internal_MemoryPoolMXBeanImpl_getPeakUsageImpl
...
-		javaVM->memoryManagerFunctions->j9gc_pools_memory(javaVM, (id&J9VM_MANAGEMENT_POOL_HEAP_ID_MASK), &totals[0], &frees[0], FALSE);
-		currentCommitted = totals[idx];
-		currentUsed = totals[idx] - frees[idx];
+		UDATA total = 0;
+		UDATA free = 0;
+		UDATA maximum = 0;
+
+		/* acquire vmAccess for retrieving memoryUsage for the pool in order to guarantee values of memoryUsage are consistent. */
+		javaVM->internalVMFunctions->internalAcquireVMAccess(vmThread);
+		maximum = javaVM->memoryManagerFunctions->j9gc_pool_memoryusage(javaVM, (id&J9VM_MANAGEMENT_POOL_HEAP_ID_MASK), &free, &total);
```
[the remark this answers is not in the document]
LinHu2016: require the access for avoiding GC happens during the call, the GC might cause retrieving inconsistent data

DanHeidinga:
@LinHu2016 Can you rebase this so it only includes your changes?

DanHeidinga:
@LinHu2016 Can you rebase this so it only includes your changes?

LinHu2016:
> @LinHu2016 Can you rebase this so it only includes your changes?

the branch has been rebased with the latest master.

LinHu2016:
update copyright to 2020

runtime/jcl/common/mgmtinit.c
DanHeidinga: @LinHu2016 Can you comment on the above question?  The old code was setting `memoryPool->preCollectionSize` here but the new code isn't

LinHu2016: the pre values has been saved to gcInfo->pre* in stead of updating memoryPool->pre*, because only if current gc manages the memory pool, we can update the memoryPool Collection values (such as the scavenge should not change tenure memorypool collection values, the scavenge does not reclaim memory from tenure pool), we do update collection values from gcInfo in gcEndEvent() (line762-774) for the collector managed pools.

DanHeidinga:

DanHeidinga:
Jenkins test extended zlinux jdk11

DanHeidinga:
Jenkins test extended zlinux jdk11

DanHeidinga:
Relaunched the extended testing as it failed.  @LinHu2016 If it fails again, can you rebase to the latest and we'll rerun the builds?

Issue #5289, opened 2019-03-28
Reduce the number of connections between JITaaS client and server


mpirvu:
In the current implementation the JITaaS client establishes a new connection to the server for each compilation request. This adds unneeded overhead especially when encryption is enabled because SSL handshake is expensive.
One idea is to keep the connection open as long as the compilation thread at the client has something in the compilation queue.

harryyu1994:
Looking into this one.

harryyu1994:
One way to implement this:
`TR::CompilationInfoPerThread::processEntries()` is where each compilation thread on the client processes the compilation queue. We should open a connection here and maintain it as long as the compilation thread is in the  `while (getCompilationThreadState()==COMPTHREAD_ACTIVE)` loop. When we get out of the loop, it means there are no more work on the thread, so we close the connection.

mpirvu:
That's the idea. I believe this will require changes to the infrastructure though.

harryyu1994:
Infrastructure Changes:
Server side:
- `TR::CompilationInfoPerThreadRemote::processEntry` handles each entry/compilation on the server
- In this function, instead of recycling `entry` & deleting `entry->_stream` after the compilation, we should keep the same `entry->_stream` and put `entry` right back in the queue.
Client Side:
- Maintain only 1 connection per compilation thread.

Then when server fails to read from `entry->_stream` in `TR::CompilationInfoPerThreadRemote::processEntry` it means that the connection is gone and that's when we clean up `entry` & `entry->_stream`

harryyu1994:
@mpirvu So after fixing a couple of corner cases, now this "simple" approach can run `-XX:JITaaSClient -version` without any issues. Will perform further testing tomorrow but it seems like it's working. No infrastructure changes required.

mpirvu:
Great! I am eager to know the testing results

harryyu1994:
28% improvement in cpu time for SSL.

PR #5329, opened 2019-04-02
Reduce connections from per comp to per compthread

harryyu1994:
- Currently JITaaS client establishes a new connection to JITaaS server for every compilation. This change reduces the connections to one per compilation thread. This change reduces the number of SSL/TCP handshakes.
- Added two environment variables TR_EnableJITaaSPerCompConn and TR_PrintJITaaSConnStats. When TR_EnableJITaaSPerCompConn is set, we are reverting back to the old way where each client compilation establishes a new connection. When TR_PrintJITaaSConnStats is set, JITaaS connections stats (the number of connections opened and closed on both the server and client) will be printed out.
- Renamed `addRemoteMethodToBeCompiled` to `addOutOfProcessMethodToBeCompiled` as this is a server routine. We use the term `remote` for client routines and `outOfProcess` for server routines.

Issue: #5289
[skip ci]

### DIFF
--- a/runtime/compiler/control/CompilationThread.cpp
+++ b/runtime/compiler/control/CompilationThread.cpp
@@ -850,7 +850,8 @@ TR::CompilationInfoPerThreadBase::CompilationInfoPerThreadBase(TR::CompilationIn
    _compilationThreadState(COMPTHREAD_UNINITIALIZED),
    _compilationShouldBeInterrupted(false),
    _addToJProfilingQueue(false),
-   _cachedClientDataPtr(nullptr)
+   _cachedClientDataPtr(nullptr),
+   _clientStream(NULL)
    {
    // At this point, compilation threads have not been fully started yet. getNumTotalCompilationThreads()
    // would not return a correct value. Need to use TR::Options::_numUsableCompilationThreads
@@ -3268,6 +3269,20 @@ void TR::CompilationInfo::stopCompilationThreads()
          }
       }
 
+   if (feGetEnv("TR_PrintJITaaSConnStats"))
+      {
+      if (getPersistentInfo()->getJITaaSMode() == SERVER_MODE)
+         {
+         fprintf(stderr, "Number of connections opened = %u\n", JITaaS::J9ServerStream::_numConnectionsOpened);
+         fprintf(stderr, "Number of connections closed = %u\n", JITaaS::J9ServerStream::_numConnectionsClosed);
+         }
+      else if (getPersistentInfo()->getJITaaSMode() == CLIENT_MODE)
+         {
+         fprintf(stderr, "Number of connections opened = %u\n", JITaaS::J9ClientStream::_numConnectionsOpened);
+         fprintf(stderr, "Number of connections closed = %u\n", JITaaS::J9ClientStream::_numConnectionsClosed);
+         }
+      }
+
 #ifdef STATS
    if (compBudgetSupport() || dynamicThreadPriority())
       {
@@ -3775,6 +3790,7 @@ TR::CompilationInfoPerThread::processEntries()
          _compInfo.getQueueWeight()
          );
       }
+
    // Possible scenarios here
    // 1. We take a method from queue and compile it
    // 2. We refrain to take a method from queue because that means two simultaneous compilations --> wait to be notified
@@ -3836,7 +3852,7 @@ TR::CompilationInfoPerThread::processEntries()
                   // It's possible that a notification was sent just after the
                   // timeout expired. Hence, if something was added to the compilation
                   // queue we must attempt to process it
-                  if (_compInfo.getMethodQueueSize() > 0)
+                  if (compInfo->getMethodQueueSize() > 0)
                      {
                      setCompilationThreadState(COMPTHREAD_ACTIVE);
                      }
@@ -3946,6 +3962,18 @@ TR::CompilationInfoPerThread::processEntries()
          // Must change the state to prevent an infinite loop
          // Change it to COMPTHREAD_SIGNAL_WAIT because the compilation queue is empty
          setCompilationThreadState(COMPTHREAD_SIGNAL_WAIT);
+         }
+      }
+
+   static bool enableJITaaSPerCompConn = feGetEnv("TR_EnableJITaaSPerCompConn") ? true : false;
+   if (compInfo->getPersistentInfo()->getJITaaSMode() == CLIENT_MODE && !enableJITaaSPerCompConn)
+      {
+      JITaaS::J9ClientStream *client = getClientStream();
+      if (client)
+         {
+         client->~J9ClientStream();
+         TR_Memory::jitPersistentFree(client);
+         setClientStream(NULL);
          }
       }
    }
@@ -12445,7 +12473,7 @@ TR::CompilationInfoPerThread::updateLastLocalGCCounter()
 // entry is queued we do not know any details about the compilation request.
 // The method needs to be executed with compilation monitor in hand
 TR_MethodToBeCompiled *
-TR::CompilationInfo::addRemoteMethodToBeCompiled(JITaaS::J9ServerStream *stream)
+TR::CompilationInfo::addOutOfProcessMethodToBeCompiled(JITaaS::J9ServerStream *stream)
    {
    TR_MethodToBeCompiled *entry = getCompilationQueueEntry(); // allocate a new entry
    if (entry)
@@ -12494,3 +12522,22 @@ TR::CompilationInfo::addRemoteMethodToBeCompiled(JITaaS::J9ServerStream *stream)
       }
    return entry;
    }
+
+void
+TR::CompilationInfo::requeueOutOfProcessEntry(TR_MethodToBeCompiled *entry)
+   {
+   TR_ASSERT(getPersistentInfo()->getJITaaSMode() == SERVER_MODE, "Should be called in JITaaS server mode only");
+
+   recycleCompilationEntry(entry);
+   // we do not requeue outOfProcessEntry for the Per-Compilation-Connection Mode
+   if (feGetEnv("TR_EnableJITaaSPerCompConn"))
+      return;
+
+   if (entry->_stream && addOutOfProcessMethodToBeCompiled(entry->_stream))
+      {
+      // successfully queued the new entry, so notify a thread
+      getCompilationMonitor()->notifyAll();
+      }
+   }
+
+
--- a/runtime/compiler/control/JITaaSCompilationThread.cpp
+++ b/runtime/compiler/control/JITaaSCompilationThread.cpp
@@ -2289,6 +2289,9 @@ remoteCompile(
    )
    {
    TR_ASSERT(vmThread->publicFlags & J9_PUBLIC_FLAGS_VM_ACCESS, "Client must work with VM access");
+   // JITaas: if TR_EnableJITaaSPerCompCon is set, then each remote compilation establishes a new connection 
+   // instead of re-using the connection shared within a compilation thread
+   static bool enableJITaaSPerCompConn = feGetEnv("TR_EnableJITaaSPerCompConn") ? true : false;
 
    // Prepare the parameters for the compilation request
    J9Class *clazz = J9_CLASS_FROM_METHOD(method);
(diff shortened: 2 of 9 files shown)


harryyu1994:
```
Without SSL, baseline:
Client:
Time spent in compilation thread =4856 ms
Time spent in compilation thread =2634 ms
Time spent in compilation thread =2544 ms
Time spent in compilation thread =3610 ms
Time spent in compilation thread =3689 ms
Time spent in compilation thread =4962 ms
Total: 22295

Server:
Time spent in compilation thread =9 ms
Time spent in compilation thread =7619 ms
Time spent in compilation thread =7753 ms
Time spent in compilation thread =17290 ms
Time spent in compilation thread =8425 ms
Time spent in compilation thread =10737 ms
Time spent in compilation thread =14847 ms
Total: 66680

Without SSL, this change:
Client:
Time spent in compilation thread =4454 ms
Time spent in compilation thread =3685 ms
Time spent in compilation thread =3782 ms
Time spent in compilation thread =2489 ms
Time spent in compilation thread =2310 ms
Time spent in compilation thread =4898 ms
Total: 21618

Server:
Time spent in compilation thread =10935 ms
Time spent in compilation thread =7510 ms
Time spent in compilation thread =6 ms
Time spent in compilation thread =439 ms
Time spent in compilation thread =11795 ms
Time spent in compilation thread =8034 ms
Time spent in compilation thread =18804 ms
Time spent in compilation thread =7561 ms
Total: 65084

With SSL, baseline:
Client:
Time spent in compilation thread =6663 ms
Time spent in compilation thread =5186 ms
Time spent in compilation thread =5036 ms
Time spent in compilation thread =3877 ms
Time spent in compilation thread =5131 ms
Time spent in compilation thread =5643 ms
Total: 31536

Server:
Time spent in compilation thread =12077 ms
Time spent in compilation thread =9120 ms
Time spent in compilation thread =9768 ms
Time spent in compilation thread =12253 ms
Time spent in compilation thread =16532 ms
Time spent in compilation thread =15805 ms
Total: 75555

With SSL, this change:

Client:
Time spent in compilation thread =3869 ms
Time spent in compilation thread =3980 ms
Time spent in compilation thread =4092 ms
Time spent in compilation thread =2812 ms
Time spent in compilation thread =5467 ms
Time spent in compilation thread =5838 ms
Total: 26058

Server:
Time spent in compilation thread =8351 ms
Time spent in compilation thread =17726 ms
Time spent in compilation thread =11310 ms
Time spent in compilation thread =534 ms
Time spent in compilation thread =5 ms
Time spent in compilation thread =13536 ms
Time spent in compilation thread =9106 ms
Time spent in compilation thread =10298 ms
Total: 70866
```

harryyu1994:
Summary of above comment:
without SSL: 3% improvement for client, 2.5% improvement for server
with SSL: 21% improvement for client, 6.6% improvement for server

harryyu1994:
@mpirvu Hi Marius, I think this is ready for a review.

harryyu1994:
@ymanton Hi Younes, this one needs a review.

mpirvu:
I will review it, but I want to be extra careful with one, so I am taking my time.

mpirvu:
Another observation: methodToBeCompiled gets a new field which is the clientStream. I think this was needed to pass the stream from processEntries() to remoteCompile(). However, processEntries() is part of CompilationInfoPerThread and we could store the stream there 9actually in CompilationInfoPerThreadBase because that's what remoteCompile() receives as a parameter. This has the advantage of having a smaller methodToBeCompiled because there could be many of them.

runtime/compiler/control/CompilationThread.cpp
```diff
@@ -3948,6 +3970,20 @@ TR::CompilationInfoPerThread::processEntries()
...
       }
+   if (compInfo->getPersistentInfo()->getJITaaSMode() == CLIENT_MODE && !enableJITaaSPerCompConn)
+      {
+      JITaaS::J9ClientStream *client = getClientStream();
+      if (client)
+         {
+         client->~J9ClientStream();
+         TR_Memory::jitPersistentFree(client);
+         setClientStream(NULL);
+         if (printJITaaSConnStats)
```
mpirvu: same as above, remove the 'if' statement

runtime/compiler/control/CompilationThread.cpp
```diff
@@ -12494,3 +12530,24 @@ TR::CompilationInfo::addRemoteMethodToBeCompiled(JITaaS::J9ServerStream *stream)
       }
    return entry;
    }
+
+void
+TR::CompilationInfo::requeueOutOfProcessEntry(TR_MethodToBeCompiled *entry)
+   {
+   bool enableJITaaSPerCompConn = feGetEnv("TR_EnableJITaaSPerCompConn") ? true : false;
```
mpirvu: Maybe move the ASSERT up front and execute `IProfiler_persistIprofileInfo` without any test (it's common on both paths). Then add the test for `enableJITaaSPerCompConn`

runtime/compiler/control/JITaaSCompilationThread.cpp
```diff
@@ -2298,6 +2299,15 @@ remoteCompile(
...
    TR::CompilationInfo *compInfo = compInfoPT->getCompilationInfo();
    bool useAotCompilation = compInfoPT->getMethodBeingCompiled()->_useAotCompilation;
+   JITaaS::J9ClientStream *client;
+   if (enableJITaaSPerCompConn)
+      {
+      client = new (PERSISTENT_NEW) JITaaS::J9ClientStream(compInfo->getPersistentInfo());
+      }
+   else
+      {
+      client = compInfoPT->getClientStream();
```
mpirvu: See my comment above about creating the connection here (but only if it doesn't exist already)

harryyu1994:
For a 10minute AcmeAir benchmark run:
Server with this change:
Number of connections opened = 116
Number of connections closed = 116

Server with this change disabled:
Number of connections opened = 9740
Number of connections closed = 9740

mpirvu:
That's quite an improvement!

runtime/compiler/control/CompilationThread.cpp
```diff
@@ -3406,6 +3416,7 @@ void TR::CompilationInfo::stopCompilationThreads()
          J9Method *ramMethod = NULL;
          uint32_t romMethodOffset = 0;
          JITaaS::J9ClientStream client(getPersistentInfo());
+         incrementJITaaSNumConnectionsOpened();
```
mpirvu: Thinking more about this statistics, we may lose information if some exception is thrown, exception that is not caught at this level, but is caught at an outer level.
The best solution is to include the stats in the J9ClientStream as statics. The constructor needs to increment the number of connections and the destructor needs to decrement it (or have two counters like you have now - that will show the watermark as well)

runtime/compiler/control/JITaaSCompilationThread.cpp
```diff
@@ -2289,6 +2289,8 @@ remoteCompile(
    )
    {
    TR_ASSERT(vmThread->publicFlags & J9_PUBLIC_FLAGS_VM_ACCESS, "Client must work with VM access");
+   // JITaas: if TR_EnableJITaaSPerCompCon is set, then each remote compilation establishes a new connection instead of reusing the connection shared within a compilation thread
```
mpirvu: This comment is way too long; please break it into two lines

runtime/compiler/control/JITaaSCompilationThread.cpp
```diff
@@ -2298,6 +2300,22 @@ remoteCompile(
    std::string detailsStr = std::string((char*) &details, sizeof(TR::IlGeneratorMethodDetails));
    TR::CompilationInfo *compInfo = compInfoPT->getCompilationInfo();
    bool useAotCompilation = compInfoPT->getMethodBeingCompiled()->_useAotCompilation;
+   JITaaS::J9ClientStream *client;
+   if (enableJITaaSPerCompConn)
+      {
+      client = new (PERSISTENT_NEW) JITaaS::J9ClientStream(compInfo->getPersistentInfo());
```
mpirvu: Can the connection establishment throw any exceptions?

harryyu1994: This line of code can throw "JITaaS::StreamFailure" and "std::alloc", will add appropriate handles

harryyu1994:
@mpirvu This change is tested and ready for review.

harryyu1994:
Found a bug, will come back for a review once the fix is ready.

harryyu1994:
```
try
            {
            client = new (PERSISTENT_NEW) JITaaS::J9ClientStream(compInfo->getPersistentInfo());
            compInfoPT->setClientStream(client);
            }
         catch (const JITaaS::StreamFailure &e)
            {
               TR_VerboseLog::writeLineLocked(TR_Vlog_JITaaS, e.what());

            client->~J9ClientStream();
            TR_Memory::jitPersistentFree(client);
            compInfoPT->setClientStream(NULL);

            compiler->failCompilation<JITaaS::StreamFailure>(e.what());
            }
```
When J9ClientStream constructor throws an exception, I tried to call `client->~J9ClientStream()`, should not do this as the object wasn't constructed.

mpirvu:
From Irwin: 
> if `JITaaS::J9ClientStream` only has `TR_ALLOC` in it's class definition, then an exception won't be thrown; it only gets thrown if you specifiy `TR_PERSISTENT_ALLOC_THROW`

So, in this case, if we run out of memory, then `client` could be NULL.

mpirvu:
```
J9ClientStream::J9ClientStream(TR::PersistentInfo *info)
   : J9Stream(),
   _timeout(info->getJITaaSTimeout())
   {
   int connfd = openConnection(info->getJITaaSServerAddress(), info->getJITaaSServerPort(), info->getJITaaSTimeout());
   BIO *ssl = openSSLConnection(_sslCtx, connfd);
   initStream(connfd, ssl);
   }
```
`initStream` does a few dynamic allocations, so those need to be cleaned up
```
  void initStream(int connfd, BIO *ssl)
      {
      _connfd = connfd;
      _ssl = ssl;
      if (_ssl)
         {
         _sslInputStream = new (PERSISTENT_NEW) SSLInputStream(_ssl);
         _sslOutputStream = new (PERSISTENT_NEW) SSLOutputStream(_ssl);
         _inputStream = new (PERSISTENT_NEW) CopyingInputStreamAdaptor(_sslInputStream);
         _outputStream = new (PERSISTENT_NEW) CopyingOutputStreamAdaptor(_sslOutputStream);
         }
      else
         {
         _inputStream = new (PERSISTENT_NEW) FileInputStream(_connfd);
         _outputStream = new (PERSISTENT_NEW) FileOutputStream(_connfd);
         }
      }
```
Typically this is done by catching any exceptions that could be thrown in the constructor, doing the cleanup and then rethrowing.

runtime/compiler/rpc/raw/J9Server.cpp
```diff
@@ -53,6 +56,7 @@ J9ServerStream::J9ServerStream(int connfd, BIO *ssl, uint32_t timeout)
    : J9Stream(),
    _msTimeout(timeout)
    {
+   _numConnectionsOpened++;
```
mpirvu: Let's move this at the very end of this function. If the constructor throws in initStream we don't want to increment the number of connections open.

runtime/compiler/rpc/raw/J9Client.cpp
```diff
@@ -242,6 +245,7 @@ J9ClientStream::J9ClientStream(TR::PersistentInfo *info)
    : J9Stream(),
    _timeout(info->getJITaaSTimeout())
    {
+   _numConnectionsOpened++;
```
mpirvu: Let's move the increment at the end of the constructor, just in case the constructor fails.

runtime/compiler/control/JITaaSCompilationThread.cpp
```diff
@@ -2298,6 +2301,54 @@ remoteCompile(
    std::string detailsStr = std::string((char*) &details, sizeof(TR::IlGeneratorMethodDetails));
    TR::CompilationInfo *compInfo = compInfoPT->getCompilationInfo();
    bool useAotCompilation = compInfoPT->getMethodBeingCompiled()->_useAotCompilation;
+   JITaaS::J9ClientStream *client;
```
mpirvu: Let's initialize with NULL and not worry that some path may leave it un-initialized.

runtime/compiler/rpc/raw/J9Server.cpp
harryyu1994: I intentionally did this so that we could see how many J9ServerStream constructions were attempted. But i guess the purpose of this is to check memory/connection leaks, i should move it to the end of the function.

harryyu1994:
Updated as requested

PR #5330, opened 2019-04-02
Fix MSG directives in StackWalker.java

pdbain-ibm:
Remove leading space which caused the preprocessor to print only the message number, not the message itself.
Also delete a legacy comment.

[ci skip]


DanHeidinga:
> Remove leading space which caused the preprocessor to misbehave.

Can you expand on how the preprocessor misbehaved?  I'm not clear on what this is fixing.

pdbain-ibm:
> > Remove leading space which caused the preprocessor to misbehave.
> 
> Can you expand on how the preprocessor misbehaved? I'm not clear on what this is fixing.

If the metacomment is not well-formed, the message processor prints only the message number.

DanHeidinga:

Issue #5243, opened 2019-03-25
Running out of scratch memory during hot compilations


dmitry-ten:
When running AcmeAir at `initialOptLevel=hot`, some compilations fail with `compilationHeapLimitExceeded` error code. This happens because the compiler runs out of scratch memory, and some failures for large methods should be expected.
However, even after I increased scratch memory space to 2GB (scratchSpaceLimit=2000000), I still see these failures.
Most of the scratch memory is allocated from `J9::SystemSegmentProvider::request`.
I wanted to see which part of the compiler requests so much memory, so
I instrumented code to dump stack traces every time that method is called for a compilation that results in out of memory failure (`java/util/concurrent/ConcurrentLinkedQueue.poll()Ljava/lang/Object;`)

The most frequent stack trace by far (17518 out of 18681) is this:
```
(_ZN2J921SystemSegmentProvider7requestEm+0x92)
[0x7fedefa92920]
(_ZN2TR6Region8allocateEmPv+0xca)
[0x7fedeff875d6]
(_ZN9TR_Memory18allocateHeapMemoryEmN13TR_MemoryBase10ObjectTypeE+0x30)
[0x7fedeff880ea]
(_ZN13TR_HeapMemory8allocateEmN13TR_MemoryBase10ObjectTypeE+0x36)
[0x7fedef9a502c]
(_ZN3OMR11InstructionnwEm13TR_HeapMemoryN13TR_MemoryBase10ObjectTypeE+0x2b)
[0x7fedefd21fcc]
(_ZN3OMR3X867Machine26coerceGPRegisterAssignmentEPN2TR11InstructionEPNS2_8RegisterENS_12RealRegister6RegNumEb+0x161)
[0x7fedf03f5333]
(_ZN29TR_X86RegisterDependencyGroup15assignRegistersEPN2TR11InstructionE16TR_RegisterKinds29TR_X86RegisterDependencyIndexPNS0_13CodeGeneratorE+0xc93)
[0x7fedf0408d61]
(_ZN3OMR3X8628RegisterDependencyConditions28assignPostConditionRegistersEPN2TR11InstructionE16TR_RegisterKindsPNS2_13CodeGeneratorE+0xa0)
[0x7fedf03ee74c]
(_ZN3OMR3X8611Instruction15assignRegistersE16TR_RegisterKinds+0xf7)
[0x7fedf03ed4c9]
[...]

PR #5333, opened 2019-04-03
Optimize JIT WrtBar logic on X86

0dvictor:
1. Removed one unnecessary layer of function call;
2. Reduced register pressure in JIT code;
3. Avoided unnecessary out-of-line branch when alwayscallwritebarrier.

### DIFF
--- a/runtime/codert_vm/xnathelp.m4
+++ b/runtime/codert_vm/xnathelp.m4
@@ -1,4 +1,4 @@
-dnl Copyright (c) 2017, 2018 IBM Corp. and others
+dnl Copyright (c) 2017, 2019 IBM Corp. and others
 dnl
 dnl This program and the accompanying materials are made available under
 dnl the terms of the Eclipse Public License 2.0 which accompanies this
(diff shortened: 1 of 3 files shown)


0dvictor:
Though this PR is not a fix for #5243, it replaces the code containing the cause of #5243.

runtime/codert_vm/xnathelp.m4
```diff
@@ -448,13 +448,35 @@ EXCEPTION_THROW_HELPER(jitThrowIncompatibleReceiver,2)
...
+			mov PARM_REG(3),uword ptr J9TR_VMThread_floatTemp2[_rbp]
+		}, {
+			push uword ptr J9TR_VMThread_floatTemp2[_rbp]
+		})
+	})
+	call FASTCALL_SYMBOL(fast_$1,$2+1)
+	RESTORE_C_VOLATILE_REGS
+	SWITCH_TO_JAVA_STACK
+	push uword ptr J9TR_VMThread_jitReturnAddress[_rbp]
+	END_HELPER($1,0)
```
gacholio: Might be worth a comment here as to why you're passing 0 instead of $2 (parms passed via mem instead of reg).

0dvictor: Good point, comments are added.

andrewcraik:

Issue #4921, opened 2019-02-28
GPF in j9vm29.dll with openj9-0.12.0 on Windows Server 2016


avermeer:
Hello,

We have a GPF (Global Protection Fault) crash of one of our Java utility running with AdoptOpenJDK 11.0.2 with OpenJ9 on Windows Server 2016, with a java dump showing that it occurred in j9vm29.dll.

I found the following page from an IBM site giving some hints about how to analyze the related Java dump:
https://www.ibm.com/support/knowledgecenter/en/SSYKE2_8.0.0/com.ibm.java.vm.80.doc/docs/javadump_tags_info.html

Based on this page, my understanding is that the crash took place in j9vm29.dll library which is part of the OpenJ9 JVM, and we have no JNI in our Java application.

This issue is not systematic, and when it occurred, the machine's memory usage wasn't heavy, and we found no interesting events in Windows events logs.

Here's the java full file:
[javacore.20190227.222307.7284.0002.txt](https://github.com/eclipse/openj9/files/2914289/javacore.20190227.222307.7284.0002.txt)

if this crash report isn't sufficient for OpenJ9 developers to spot where the issue is, then would you please let me know what kind options I could add to 'java' command to get more traces in case this crash takes place again?

Kind regards,
Alexandre Vermeerbergen

pshipton:
@avermeer is there core file created as part of the diagnostic output, and are you able to compress and share it via a private message?

avermeer:
Hello @pshipton,

Unfortunately no, I do not have a core file other than the java core file attached to the incident.

But our crash is on Windows 2016, isn't "core" file a notion specific to UNIX or Linux operating system?

(sorry if the question sounds trivial)

Kind regards,
Alexandre Vermeerbergen

pshipton:
@avermeer np. Windows has the same concept, and regardless of what Windows calls the file, the JVM renames it by default to `core.<date>.<time>.<pid>.<seq>` i.e. like core.20190228.125330.6080.0001.dmp. For a gpf, the default behavior is to create a number of diagnostic files. A javacore, snap (trace output), core, jitdump.

It is a bit surprising you only got the javacore file and nothing else. Usually we need to core file to diagnose a gfp problem. If there is console output, it shows the diagnostic files written, and where they were written. Like so:
```

**Comment by pshipton** [comment]
```
JVMDUMP039I Processing dump event "vmstart", detail "" at 2019/02/28 12:53:30 - please wait.
[...]

PR #5334, opened 2019-04-03
Module hash function uses J9Module* pointer as key

JasonFengJ9:
Module hash function uses `J9Module*` pointer as key

Module name string object (`J9Module->moduleName`) is interned and shared among different modules with same name. Using module name as key within `moduleHashFn` can't differentiate different modules;
Modified `moduleHashFn` to use the `J9Module*` pointer as key instead of the module name;
Updated `moduleHashEqualFn` accordingly;
Updated trace points to record `J9Module*` pointer values.
Also obsoleted redundant trace points.
Note: `moduleExtraInfoHashFn` uses `J9Module*` pointer as key as well.

Verified that this PR fixed the crash in #4921 (tested at Linux AMD platform).

Closes: #4921 

Reviewer: @DanHeidinga 

### DIFF
--- a/runtime/j9vm/java11vmi.c
+++ b/runtime/j9vm/java11vmi.c
@@ -112,6 +112,7 @@ hashTableAtPut(J9HashTable * table, void * value, BOOLEAN collisionIsFailure)
 	} else if (collisionIsFailure) {
 		retval = HASHTABLE_ATPUT_COLLISION_FAILURE;
 	} else {
+		Trc_MODULE_hashTableAtPut(table, value, node);
 		retval = HASHTABLE_ATPUT_SUCCESS;
 	}
 
@@ -320,9 +321,9 @@ trcModulesCreationPackage(J9VMThread * currentThread, J9Module * fromModule, con
 
 	if (NULL != moduleNameUTF) {
 		if (0 == strcmp(moduleNameUTF, JAVA_BASE_MODULE)) {
-			Trc_MODULE_creation_package(currentThread, package, "java.base");
+			Trc_MODULE_createPackage(currentThread, package, "java.base", fromModule);
 		} else {
-			Trc_MODULE_creation_package(currentThread, package, moduleNameUTF);
+			Trc_MODULE_createPackage(currentThread, package, moduleNameUTF, fromModule);
 		}
 		if (moduleNameBuf != moduleNameUTF) {
 			j9mem_free_memory(moduleNameUTF);
@@ -342,13 +343,14 @@ addPackageDefinition(J9VMThread * currentThread, J9Module * fromModule, const ch
 	J9Package * j9package = createPackage(currentThread, fromModule, package);
 
 	if (NULL != j9package) {
+		Trc_MODULE_invokeHashTableAtPut(currentThread, "addPackageDefinition", classLoader, classLoader->packageHashTable, &j9package, j9package, "true");
 		retval = (0 == hashTableAtPut(classLoader->packageHashTable, (void*)&j9package, TRUE));
 	}
 
 	if (!retval) {
 		freePackage(currentThread, j9package);
 	} else {
-		if (TrcEnabled_Trc_MODULE_creation_package) {
+		if (TrcEnabled_Trc_MODULE_createPackage) {
 			trcModulesCreationPackage(currentThread, fromModule, package);
 		}
 	}
@@ -423,6 +425,7 @@ addModuleDefinition(J9VMThread * currentThread, J9Module * fromModule, const cha
 		retval = addMulPackageDefinitions(currentThread, fromModule, packages, numPackages);
 		if (ERRCODE_SUCCESS == retval) {
 			BOOLEAN const success = (0 == hashTableAtPut(classLoader->moduleHashTable, (void*)&fromModule, TRUE));
+			Trc_MODULE_invokeHashTableAtPut(currentThread, "addModuleDefinition", classLoader, classLoader->moduleHashTable, &fromModule, fromModule, "true");
 			if (NULL != version) {
 				fromModule->version = J9_JNI_UNWRAP_REFERENCE(version);
 			}
@@ -481,7 +484,7 @@ trcModulesAddModuleExportsToAll(J9VMThread * currentThread, J9Module * fromModul
 	char *fromModuleNameUTF = vmFuncs->copyStringToUTF8WithMemAlloc(
 		currentThread, fromModule->moduleName, J9_STR_NULL_TERMINATE_RESULT, "", 0, fromModuleNameBuf, J9VM_PACKAGE_NAME_BUFFER_LENGTH, NULL);
 	if (NULL != fromModuleNameUTF) {
-		Trc_MODULE_add_module_exports_to_all(currentThread, package, fromModuleNameUTF);
+		Trc_MODULE_addModuleExportsToAll(currentThread, package, fromModuleNameUTF, fromModule);
 		if (fromModuleNameBuf != fromModuleNameUTF) {
 			j9mem_free_memory(fromModuleNameUTF);
 		}
@@ -495,7 +498,7 @@ exportPackageToAll(J9VMThread * currentThread, J9Module * fromModule, const char
 	J9Package * const j9package = getPackageDefinition(currentThread, fromModule, package, &retval);
 	if (NULL != j9package) {
 		j9package->exportToAll = TRUE;
-		if (TrcEnabled_Trc_MODULE_add_module_exports_to_all) {
+		if (TrcEnabled_Trc_MODULE_addModuleExportsToAll) {
 			trcModulesAddModuleExportsToAll(currentThread, fromModule, package);
 		}
 	}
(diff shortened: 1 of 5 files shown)


runtime/j9vm/java11vmi.c
```diff
@@ -342,13 +343,14 @@ addPackageDefinition(J9VMThread * currentThread, J9Module * fromModule, const ch
 	J9Package * j9package = createPackage(currentThread, fromModule, package);
 
 	if (NULL != j9package) {
+		Trc_MODULE_invokeHashTableAtPut(currentThread, "addPackageDefinition", classLoader->packageHashTable, &j9package, j9package, "true");
```
DanHeidinga: Is the packageHashTable the relevant thing to log here?  I think the classloader itself might be more useful as it's easier to correlate to other debug data and gives the right `!j9classloader 0xaddress` to find the loader in core dump

runtime/j9vm/java11vmi.c
```diff
@@ -423,6 +425,7 @@ addModuleDefinition(J9VMThread * currentThread, J9Module * fromModule, const cha
 		retval = addMulPackageDefinitions(currentThread, fromModule, packages, numPackages);
 		if (ERRCODE_SUCCESS == retval) {
 			BOOLEAN const success = (0 == hashTableAtPut(classLoader->moduleHashTable, (void*)&fromModule, TRUE));
+			Trc_MODULE_invokeHashTableAtPut(currentThread, "addModuleDefinition", classLoader->moduleHashTable, &fromModule, fromModule, "true");
```
DanHeidinga: Similar question here - do we want to log the moduleHashTable or the ClassLoader pointer?

runtime/j9vm/java11vmi.c
```diff
@@ -603,6 +606,7 @@ exportPackageToModule(J9VMThread * currentThread, J9Module * fromModule, const c
 	J9Package * const j9package = getPackageDefinition(currentThread, fromModule, package, &retval);
 	if (NULL != j9package) {
 		if (isModuleDefined(currentThread, toModule)) {
+			Trc_MODULE_invokeHashTableAtPut(currentThread, "exportPackageToModule(exportsHashTable)", j9package->exportsHashTable, &toModule, toModule, "false");
```
DanHeidinga: I'm less sure about what to log here.... though by analogy to the above, it likely should be the `j9package`

runtime/j9vm/java11vmi.c
```diff
@@ -617,6 +621,7 @@ exportPackageToModule(J9VMThread * currentThread, J9Module * fromModule, const c
 					toModule->removeExportsHashTable = vm->internalVMFunctions->hashPackageTableNew(vm, INITIAL_INTERNAL_PACKAGE_HASHTABLE_SIZE);
 				}
 				if (NULL != toModule->removeExportsHashTable) {
+					Trc_MODULE_invokeHashTableAtPut(currentThread, "exportPackageToModule(removeExportsHashTable)", toModule->removeExportsHashTable, &j9package, j9package, "false");
```
DanHeidinga: ...and the toModule here

runtime/j9vm/java11vmi.c
```diff
@@ -650,6 +655,7 @@ allowReadAccessToModule(J9VMThread * currentThread, J9Module * fromModule, J9Mod
 			retval = ERRCODE_SUCCESS;
 		} else if (isModuleDefined(currentThread, toModule)) {
 			BOOLEAN success = FALSE;
+			Trc_MODULE_invokeHashTableAtPut(currentThread, "allowReadAccessToModule(readAccessHashTable)", toModule->readAccessHashTable, &fromModule, fromModule, "false");
```
DanHeidinga: toModule here?

runtime/j9vm/java11vmi.c
```diff
@@ -663,6 +669,7 @@ allowReadAccessToModule(J9VMThread * currentThread, J9Module * fromModule, J9Mod
 					fromModule->removeAccessHashTable = vm->internalVMFunctions->hashModuleTableNew(vm, INITIAL_INTERNAL_MODULE_HASHTABLE_SIZE);
 				}
 				if (NULL != fromModule->removeAccessHashTable) {
+					Trc_MODULE_invokeHashTableAtPut(currentThread, "allowReadAccessToModule(removeAccessHashTable)", fromModule->removeAccessHashTable, &toModule, toModule, "false");
```
DanHeidinga: fromModule here?

runtime/vm/ModularityHashTables.c
```diff
@@ -24,31 +24,19 @@
...
 static UDATA 
 moduleHashFn(void *key, void *userData) 
 {
-	J9JavaVM* javaVM = (J9JavaVM*) userData;
-	j9object_t name = moduleHashGetName(key);
-
-	return javaVM->memoryManagerFunctions->j9gc_stringHashFn(&name, userData);
+	const J9Module** const modulePtr = (const J9Module**)key;
+	
+	return (UDATA)(*modulePtr);
```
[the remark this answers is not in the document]
JasonFengJ9: Thanks @DanHeidinga , `J9Module` objects are allocated with `pool_newElement(vm->modularityPool)` as following:
https://github.com/eclipse/openj9/blob/d12dbf8365974daaa39257ad035cab94000c2a5c/runtime/vm/jvminit.c#L2196
https://github.com/eclipse/openj9/blob/d12dbf8365974daaa39257ad035cab94000c2a5c/runtime/vm/jvminit.c#L2203
https://github.com/eclipse/openj9/blob/d12dbf8365974daaa39257ad035cab94000c2a5c/runtime/j9vm/java11vmi.c#L252
So there is no stack-allocation of `J9Module`. In addition existing `moduleExtraInfoHashFn` & `moduleExtraInfoHashEqualFn` have performed hash on `J9Module` pointer since modularity era.

runtime/j9vm/java11vmi.c
JasonFengJ9: Agreed, will change to logging `classLoader` instead. (This is the part inherited from existing trace points.)

runtime/j9vm/java11vmi.c
JasonFengJ9: Yeah, same thing here.

runtime/j9vm/java11vmi.c
JasonFengJ9: Will log `j9package` here.

This issue investigation didn't involve much about `j9package` or `HashTable` itself, it was mainly to correlate the `Java Module object` with native `J9Module` structure underneath.

The core dump file can indicate the mismatch and identify the bad object however it was difficult to crawl back when the bad object was created and how it becomes invalid. When the failure is intermittent and takes long hours to reproduce, trace points are the friends to record what happened. 

So I would like to put both `j9package` and `j9package->exportsHashTable` into the log since it doesn't require much extra cost.

runtime/j9vm/java11vmi.c
JasonFengJ9: will add.

runtime/j9vm/java11vmi.c
JasonFengJ9: Added.

runtime/j9vm/java11vmi.c
JasonFengJ9: Fixed.

@DanHeidinga the PR has been updated accordingly, please have another look.

DanHeidinga:
Jenkins test sanity plinux,zlinux jdk11

DanHeidinga:
Jenkins test extended win,osx jdk11

DanHeidinga:
@JasonFengJ9 Is there a test we can add that will detect the case from the original bug report?

JasonFengJ9:
@DanHeidinga I attempted to reproduce the failure in a smaller test case with usage of modules `jdk.translet` and `java.xml` but no luck, also tried to build customized modules with similar usages.
The problem is related to certain GC activities to invalidate the objects in question which is not deterministic and takes long time period to reproduce.
Initial crash can be observed in 3/50 runs (1 - 2 hours), with this PR, 70x runs finished without crash. All tests were in Linux AMD platform.

DanHeidinga:
@JasonFengJ9 Can you open a PR for this against the `0.14.0` branch?

DanHeidinga:
@JasonFengJ9 Is there any point were we need to map from module name to the J9Module?  I've done a quick look but am looking for a second set of eyes to confirm we don't need this capability

JasonFengJ9:
@DanHeidinga considering those dynamically generated modules with same names like `jdk.translet`, the module name becomes a literal representation of that module just like class name for a class object instance. So I think it won't be reliable to map a module name to a `J9Module` in such cases.
For those unique modules like `java.base`, it still is a one-to-one map though.

Existing code before this PR uses module name for hashtable usage only (`moduleHashFn` & `moduleHashEqualFn`), there is no actual query method to give a `J9Module` based on `module name`.

DanHeidinga:
Just to note: A name + classloader (module layer?) should be unique as well.

PR #5337, opened 2019-04-03
Remove obsolete JPP flag

pdbain-ibm:
Sidecar19-SE-OpenJ9 was used to distinguish Build 148 of Java 9 from the current build.
Since B148 is dead we need only select Java 9 or later.


pshipton:
jenkins compile zlinux jdk11

PR #5339, opened 2019-04-03
Add utility method to check if we are running on OpenJ9

pdbain-ibm:
@pshipton and @JasonFengJ9 would you kindly review this too?
[ci skip]

### DIFF
--- a/test/functional/TestUtilities/src/org/openj9/test/util/PlatformInfo.java
+++ b/test/functional/TestUtilities/src/org/openj9/test/util/PlatformInfo.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2018, 2018 IBM Corp. and others
+ * Copyright (c) 2018, 2019 IBM Corp. and others
  *
  * This program and the accompanying materials are made available under
  * the terms of the Eclipse Public License 2.0 which accompanies this
@@ -22,8 +22,10 @@
 package org.openj9.test.util;
 
 public class PlatformInfo {
-	private static final String OS_NAME = "os.name";
+	private static final String OS_NAME = "os.name"; //$NON-NLS-1$
 	private static final String osName = System.getProperty(OS_NAME);
+	private	static final boolean isOpenJ9Status = 
+			System.getProperty("java.vm.vendor").contains("OpenJ9"); //$NON-NLS-1$ //$NON-NLS-2$
 
 	public static boolean isWindows() {
 		return ((null != osName) && osName.startsWith("Windows"));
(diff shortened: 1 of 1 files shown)


test/functional/TestUtilities/src/org/openj9/test/util/PlatformInfo.java
```diff
@@ -33,6 +33,10 @@ public static boolean isMacOS() {
 		return ((null != osName) && osName.startsWith("Mac OS"));
 	}
 
+	public static boolean isOpenJ9() {
+		return System.getProperty("java.vm.vendor").contains("OpenJ9");
```
pshipton: Not sure how this is intended to be used, but note that IBM uses OpenJ9 but sets `java.vm.vendor = IBM Corporation`

pdbain-ibm: Then purpose is to check for pure OpenJ9 builds, as opposed to IBM builds which do not contain things like `jps`, `jstack`, etc.

JasonFengJ9: Just echo to Pete's comment, please provide a description about the usage scenario.
From code perspective, `System.getProperty("java.vm.vendor")` can be in class initializer to avoid repeating invocation through method `isOpenJ9()`.
By the way, this is a code change, is it really worth to skip `ci` though I can't eyeball any potential compilation issue?

pshipton: @JasonFengJ9 the ci testing just builds the VM, while this change is to test material

llxia: - Could we get some clarification how this methold will be used?
- We are using `java.vm.name` for AUTO_DETECT
https://github.com/eclipse/openj9/blob/6bf7e0de73f5e1ce5daac1d709d07fca97733900/test/TestConfig/src/org/openj9/envInfo/JavaInfo.java#L102

pdbain-ibm: - Elaborated the comment
- Set the value at class initialization

test/functional/TestUtilities/src/org/openj9/test/util/PlatformInfo.java
```diff
@@ -33,6 +35,16 @@ public static boolean isMacOS() {
 		return ((null != osName) && osName.startsWith("Mac OS"));
 	}
 
+	/**
+	 * Test if this is a pure OpenJ9 JDK, not a derivative or alternate VM.
+	 * This is used by tests which require specific tools or packages with the JVM.
+	 *
+	 * @return true if this is a stock OpenJ9 JDK
+	 */
+	public static boolean isOpenJ9() {
```
pshipton: I'm concerned this test is not going to be robust enough. While it may work as desired today, it may not give the desired results in the future, in particular if OpenJ9 is picked up and used in other ways.

If you want to know if jps and jstack are supported, how about looking for these files?

pdbain-ibm: I do in fact look for these tools elsewhere, but I wanted to check early and explicitly.

pdbain-ibm: Also, I want to check that the tool under test is present when it should be.

PR #5343, opened 2019-04-03
Change staccato flag and library usage to realtime

jason-hall:


runtime/cmake/caches/common.cmake
```diff
@@ -36,7 +36,7 @@ set(OMR_GC_MODRON_SCAVENGER ON CACHE BOOL "")
 set(OMR_GC_REALTIME ON CACHE BOOL "")
 set(OMR_GC_SCAVENGER_DELEGATE ON CACHE BOOL "")
 set(OMR_GC_SEGREGATED_HEAP ON CACHE BOOL "")
-set(OMR_GC_STACCATO ON CACHE BOOL "")
+set(OMR_GC_STACCATO ON CACHE BOOL "");
```
dmitripivkine: is it correct?

runtime/gc_modron_startup/mmparseXXgc.cpp
```diff
@@ -101,11 +100,9 @@ gcParseXXgcArguments(J9JavaVM *vm, char *optArg)
 		}
 		if (try_scan(&scan_start, "stw")) {
 			MM_Scheduler::initializeForVirtualSTW(extensions);
-#if defined(J9VM_GC_STACCATO)
```
dmitripivkine: Change like this required OMR code to be fixed first

dmitripivkine: Also I believe J9VM_GC_STACCATO define check should be replaced to another (modern) flag

runtime/cmake/caches/common.cmake
dmitripivkine: fix this typo

runtime/gc_modron_startup/mmparseXXgc.cpp
dmitripivkine: ok, this definition is under J9VM_GC_REALTIME flag already. So it is ok to just remove J9VM_GC_STACCATO here.
OMR code can be fixed later

dmitripivkine:
With fixed typo this change set is good to go

dmitripivkine:

dmitripivkine:
Jenkins compile win

PR #5346, opened 2019-04-03
Locking Code Revision

babsingh:
**Fix code formatting in monhelpers.c**

1) Remove unneeded code.
2) Initialize the variables during declaration to avoid any potential
issues.

**Fix the comment regarding sched_compat_yield and yield count settings**

If the Completely Fair Scheduler is detected and sched_compat_yield=0,
we do not default to the -Xthr:minimizeUserCPU settings. Instead, we
increase the yield counts to 270 in the three-tier spinlock loops. Refer
to the updated comment for more details.

### DIFF
--- a/runtime/vm/vmthread.c
+++ b/runtime/vm/vmthread.c
@@ -558,7 +558,17 @@ threadParseArguments(J9JavaVM *vm, char *optArg)
 #endif /* defined(OMR_THR_YIELD_ALG) */
 	
 #if defined(LINUX)
-	/* if Completely Fair Scheduler is detected, and sched_compat_yield=0, default to -Xthr:minimizeUserCPU */
+	/* Check the sched_compat_yield setting for the versions of the Completely Fair Scheduler (CFS) which
+	 * have broken the thread_yield behavior. If running in CFS and sched_compat_yield=0, the CPU yielding
+	 * behavior is moderated to act as though CFS is not enabled by increasing the yield count to 270 in
+	 * the three-tier spinlock loops.
+	 *
+	 * sched_compat_yield=1 uses the aggressive CPU yielding behavior of some versions of the O(1)
+	 * scheduler and uses the default yield counts (= 45).
+	 *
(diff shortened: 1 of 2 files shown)


runtime/vm/vmthread.c
```diff
@@ -558,7 +558,14 @@ threadParseArguments(J9JavaVM *vm, char *optArg)
...
 #if defined(LINUX)
-	/* if Completely Fair Scheduler is detected, and sched_compat_yield=0, default to -Xthr:minimizeUserCPU */
+	/* Settings for the three-tier spinlock loops if the Completely Fair Scheduler (CFS) is detected,
+	 * and sched_compat_yield=0. sched_compat_yield=1 enables the aggressive CPU yielding behavior of
+	 * the O(1) scheduler by moving the relinquishing task to the end of the runnable queue. For
+	 * sched_compat_yield=1, the yield count is set to 45 in the three-tier spinlock loops. For the
+	 * CFS and sched_compat_yield=0, the CPU yielding behavior is moderate in comparison to the CPU
+	 * yielding behavior of the O(1) scheduler. So, the yield count is increased to 270 in the
+	 * three-tier spinlock loops.
+	 */
```
DanHeidinga: I had trouble parsing this and its applicability to the code below.  Maybe something like:
```c
/* Check for the sched_compat_yield setting for versions of the the Complete Fair Scheduler (CFS)
 * which have broken thread_yield behaviour.  If running in CFS and sched_compat_yield=0, the CPU
 * yielding behavior is moderated to act as though CFS was not enabled.  Increase the yield count
 * to 270 in the three-tier spinlock loops.
 * 
 * sched_compat_yield=1 uses the aggressive CPU yielding behavior of some versions of the O(1) 
 * scheduler and uses the default yield counts.
 * 
 * Newer versions of linux have removed this compat flag as the thread_yield behaviour has been
 * restored to something sane.
 */
```

babsingh: @DanHeidinga I have updated the comment.

DanHeidinga:

PR #5351, opened 2019-04-03
Minor cleanup with forceOutOfLineVMAccess()

amicic:
As part of preparation for more usage in near future:
- remove the assert it's used only for Concurrent Scavenger (the call
sites will check if CS is enabled or not prior to calling)
- remove #ifdefs, to not force call sites to use it too, hence resulting
in easier to read code. The amount of code behind it is small and there
is no real reason to hide it under ifdef even for non CS configuration.


dmitripivkine:
Jenkins compile win

PR #5360, opened 2019-04-04
Do not generate compressedrefs and awrtbari for static field

liqunl:
Reference type static fields are not compressed and don't live in the
heap, thus compressedrefs anchor and write bar are not needed for load
and store.


liqunl:
https://github.com/eclipse/openj9/pull/5316

liqunl:
@andrewcraik @pshipton

andrewcraik:
good to go once ci finishes

pshipton:
I don't think ci works for branches.

PR #5363, opened 2019-04-04
Fix decomposition of blob version

keithc-ca:
The expression to compute the fourth byte was wrong.


pshipton:

pshipton:
@keithc-ca does this fix anything important, should it be included in the 0.14 release?

pshipton:
jenkins test extended zlinux jdk8

pshipton:
Testing with internal build 413864

keithc-ca:
> does this fix anything important, should it be included in the 0.14 release?

The only use of the broken method that I could find didn't use the fourth byte (that should now be correct), so I don't think think this needs to go into the 0.14 release.

PR #5365, opened 2019-04-04
Fix reading of small types (byte, short) near the end of a segment

keithc-ca:
For example, the final three bytes of a virtual memory segment were previously inaccessible individually.


pshipton:
@keithc-ca does this fix anything important, should it be included in the 0.14 release?

pshipton:

keithc-ca:
> does this fix anything important, should it be included in the 0.14 release?

I think this only affects reading z/OS system dumps and so doesn't need to be in the 0.14 release.

PR #5369, opened 2019-04-04
Remove duplicate definition of mcc_printf

rwy0717:
mcc_printf is defined in OMR. The definition in OMR was changed, so now
this duplicate definition is causing a compilation error. Since this
definition serves no purpose (is not used, has no effect), remove it.


pshipton:
jenkins compile zlinux jdk8

Issue #4476, opened 2019-01-28
Allow multiple hot compilations in parallel in JITaaS mode


mpirvu:
The heuristics in the JIT don't allow multiple expensive compilation to be performed in parallel to avoid increasing the peak footprint too much. This is controlled by an option which defines what an expensive compilation is: `-Xjit:expensiveCompWeight=`
The default value is set such that hot, scorching and JSR292 compilations are considered expensive.
However, in JITaaS world such compilation are performed remotely and therefore there is no risk of blowing the footprint limit on the JVM. Therefore, for the JITaaS client we should set a large value for `expensiveCompWeight` to allow several memory expensive compilations to be performed in parallel.

harryyu1994:
Looking into this one.

PR #5372, opened 2019-04-04
Allow concurrent hot compilations in JITaaS mode

harryyu1994:
The heuristics in the JIT don't allow multiple expensive compilation to be performed in parallel to avoid increasing the peak footprint too much. The default value is set such that hot, scorching and JSR292 compilations are considered expensive. However, in JITaaS world such compilation are performed remotely and therefore there is no risk of blowing the footprint limit on the JVM. By setting a large value for "expensiveCompWeight", we are effectively allowing several memory expensive compilations to be performed concurrently.

Issue: #4476
[skip ci]

### DIFF
--- a/runtime/compiler/control/J9Options.cpp
+++ b/runtime/compiler/control/J9Options.cpp
@@ -2310,6 +2310,11 @@ J9::Options::setupJITaaSOptions()
          // no IProfiler info is collected at the server itself
          self()->setOption(TR_DisableIProfilerThread);
          }
+
+      // In the JITaaS world, expensive compilations are performed remotely so there is no risk of blowing the footprint limit on the JVM
+      // Setting _expensiveCompWeight to a large value so that JSR292/hot/scorching compilation are allowed to be executed concurrently
+      TR::Options::_expensiveCompWeight = TR::CompilationInfo::MAX_WEIGHT;
+
       }
 
    if (TR::Options::getVerboseOption(TR_VerboseJITaaS))


harryyu1994:
@ymanton Hi Younes, this one needs a review.

runtime/compiler/control/J9Options.cpp
```diff
@@ -2310,6 +2310,13 @@ J9::Options::setupJITaaSOptions()
          // no IProfiler info is collected at the server itself
          self()->setOption(TR_DisableIProfilerThread);
          }
+
+      // In the JITaaS world, expensive compilations are performed remotely so there is no risk of blowing the footprint limit on the JVM
+      // Setting _expensiveCompWeight to a large value so that JSR292/hot/scorching compilation are allowed to be executed concurrently
+      if (TR::Options::_expensiveCompWeight < 101)
+         {
+         TR::Options::_expensiveCompWeight = 101; // Scorching weight = 100
+         }
```
ymanton: Can we just set it to `MAX_WEIGHT` without checking anything?

harryyu1994: set to `MAX_WEIGHT`, used 255 instead of `MAX_WEIGHT` to avoid including the `CompilationRuntime.hpp` header in `J9Options.cpp`.

ymanton: Use `MAX_WEIGHT` and don't worry about the header, it's already included in that file.

harryyu1994: Changed as requested

PR #5385, opened 2019-04-05
Rework field/staticAttributes for JITaaS

dmitry-ten:
- Reimplement `TR_ResolvedRelocatableJ9JITaaSServerMethod::field/staticAttributes`
to use the same set of attributes sent from client to the server as
non-AOT version.
This gets rid of a union and makes reusing code for AOT and non-AOT
versions easier.
- Reimplement `TR_J9MethodFieldAttributes`.
First, it's not a union of 2 structs anymore, because of the previous
point.
Second, instead of just being a struct, it's now a class with 2 public
constructors. Constructor corresponds to initializing field or static
attributes. This gets rid of another union.
Third, implement helper methods for setting attributes and override
`operator==`. The operator override will be useful later, when we'll
be comparing cached attributes with newly fetched attributes

### DIFF
--- a/runtime/compiler/env/j9methodServer.hpp
+++ b/runtime/compiler/env/j9methodServer.hpp
@@ -28,43 +28,67 @@
 #include "control/JITaaSCompilationThread.hpp"
 #include "runtime/JITaaSIProfiler.hpp"
 
-struct
-TR_J9MethodResolvedFieldAttributes
+class TR_J9MethodFieldAttributes
    {
-   // for field attributes, use fieldOffset,
-   // for static attributes, use address
-   union
-      {
-      U_32 fieldOffset;
-      void *address;
-      };
-   TR::DataTypes type; 
-   bool volatileP;
-   bool isFinal;
-   bool isPrivate;
-   bool unresolvedInCP;
-   bool result;
-   };
-struct
-TR_J9MethodResolvedRelocatableFieldAttributes
-   {
-   // for field attributes, use fieldOffset,
-   // for static attributes, use address
-   union
-      {
-      IDATA fieldOffset;
-      void *address;
-      };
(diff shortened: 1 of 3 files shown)


mpirvu:
Was this thoroughly tested? Refactoring is always high risk of missing something.

dmitry-ten:
I tested it with the benchmarks we usually run, didn't see any problems.
I'll run sanity tests to make sure.

dmitry-ten:
_sanity testing completed, results look fine.
Will also do _sanity testing with AOT enabled.

dmitry-ten:
For AOT, `JCL_TEST_SE80` failed, but that seems to be unrelated to this change.
Everything else looks fine.

runtime/compiler/env/j9methodServer.hpp
```diff
@@ -28,43 +28,67 @@
...
       {
-      IDATA fieldOffset;
-      void *address;
-      };
-   J9Class *definingClass;
-   UDATA ltype;
-   bool unresolvedInCP;
-   };
+      setMethodFieldAttributesResult(type, volatileP, isFinal, isPrivate, unresolvedInCP, result);
+      if (fieldOffset) *fieldOffset = (uint32_t) *( (uint32_t *) &_fieldOffsetOrAddress);
```
dchopra001: So this won't work on BE architectures.

Also, we should use C++ style casts.

dmitry-ten: Interesting. What is BE architecture?

dchopra001: Big Endian.

dmitry-ten: Ah, I see.

Issue #3634, opened 2018-11-10
MauveSingleThreadLoadTest_0 crashed in JITaaS JDK8 build


llxia:
java -version
```
00:15:57.546 "/home/jenkins/workspace/Grinder/openjdkbinary/j2sdk-image/jre/bin/java" -version
00:15:57.546 java version "1.8.0"
00:15:57.546 Java(TM) SE Runtime Environment (build 1.8.0-foreman_2018_10_22_11_55-b00)
00:15:57.546 IBM J9 VM (build 2.9, JRE 1.8.0 Linux amd64-64-Bit Compressed References 20181103_401303 (JIT enabled, AOT enabled)
00:15:57.546 OpenJ9   - 29bd9e6
00:15:57.546 OMR      - 942d987
00:15:57.546 IBM      - 5c4a9f0)
```
MauveSingleThreadLoadTest_0 test results
```
LT  22:13:50.164 - Starting thread. Suite=0 thread=0
LT  22:14:10.167 - Completed 3.4%. Number of tests started=66127
LT  22:14:30.188 - Completed 10.6%. Number of tests started=205123 (+138996)
LT  stderr Unhandled exception
LT  stderr Type=Segmentation error vmState=0x00000000
LT  stderr J9Generic_Signal_Number=00000004 Signal_Number=0000000b Error_Value=00000000 Signal_Code=00000001
LT  stderr Handler1=00007F9919D21570 Handler2=00007F991961C320 InaccessibleAddress=FFFFFFFFFFFFFFF8
LT  stderr RDI=0000000001E84000 RSI=00007F98FC1BD480 RAX=0000000000000000 RBX=00007F98FC1BD480
LT  stderr RCX=0000000000000001 RDX=00007F991426EDF0 R8=0000000000000006 R9=00007F98FEF7DD51
LT  stderr R10=00007F991426EE40 R11=00007F991426EDF0 R12=0000000001E84000 R13=00007F991426ED78
LT  stderr R14=00007F991426EDC8 R15=00007F991426EE18
LT  stderr RIP=00007F9911016405 GS=0000 FS=0000 RSP=00007F98FC1BD1F0
LT  stderr EFlags=0000000000010202 CS=0033 RBP=0000000001BC7960 ERR=0000000000000005
LT  stderr TRAPNO=000000000000000E OLDMASK=0000000000000000 CR2=FFFFFFFFFFFFFFF8
[...]

PR #5400, opened 2019-04-08
Fix check in isCompilable()

ymanton:
Missing comma causes two versions of doPrivileged()
to bypass the filter and be compiled or inlined,
which causes runtime crashes during stack walking.

Found on the jitaas branch.

Fixes: #3634


ymanton:
Jenkins test sanity all jdk8

mpirvu:
Sanity on win64 failed due to infra. Change is safe, so merging.

mpirvu:
I pressed close by mistake. re-opened and testing starting again

mpirvu:
The two failures are infra.

Issue #4542, opened 2019-02-01
Indy work for jep334


theresa-m:
This is my current understanding of the problem and what needs to happen:

m is the original MethodHandle
mDesc is a nominal descriptor of MethodHandle of type MethodHandleDesc
m' is the MethodHandle resolved mDesc

m and m' are equivalent.

If indy is called on m and then m' the current expected behaviour is that the bsm will be called and MethodHandle will need to be resolved each time. This is unnecessary because the resulting call site entries will contain the same information.

**Steps:**
- [ ] Write a test case to verify that resolved invokedynamic call sites are propagated between retransformed classes (it is assumed that this should fail) https://github.com/eclipse/openj9/pull/5029

- [ ] Discuss next steps including the best way to determine callsite equivalence in `redefineClassesCommon` code 

fyi @DanHeidinga @gacholio

DanHeidinga:
This isn't quite what I was thinking when we talked so apologies for the unclear presentation.

The background:
JVMTI allows classes to be redefined / retransformed - basically modifying method bodies to contain different bytecodes.  

Now, not every retransformation modifies every method.  Many methods are unchanged and are viewed as `equivalent` between the original and the redefined class.  Equivalent in this context means the methods have the same bytecodes that point to equivalent constant pool data (ie: the cp indexes may be different, but the data must be the same).

This task is to confirm whether OpenJ9 propagates the resolved state for invokedynamic bytecodes in equivalent methods from the original class to the new class.  Default assumption - it doesn't.

Step 1 is to write a test that uses invokedynamic in a method.  Run the method and observe the BSM being invoked.  Retransform the class in a way that the method with the invokedynamic is equivalent.  Invoke the method again and see if the BSM is invoked a second time.

If the BSM is invoked a second time, it means the resolution state for those invokedynamics isn't being propagated.  The resolution state is stored in the `J9Class->callsites[]` array and we'll need to find a way to map from the old index into that array to the new index.  Note, indexes are assigned linearly from the start of the class.

theresa-m:
Okay that makes sense. So I'm assuming one of the ways retransformation can be simulated is when a MH's nominal descriptor is resolved? Does retransformation also happen when something like MethodHandles.lookup().findStatic() is called?

PR #5402, opened 2019-04-08
Class redefinition CallSite propagation

theresa-m:
Fixes: #4542
Tests: https://github.com/eclipse/openj9/pull/5029

Method equivalence is used to determine whether CallSites should be propagated.

### DIFF
--- a/runtime/jvmti/jvmtiClass.c
+++ b/runtime/jvmti/jvmtiClass.c
@@ -1164,8 +1164,8 @@ redefineClassesCommon(jvmtiEnv* env,
 			if (!extensionsEnabled) {
 				/* Fast HCR path - where the J9Class is redefined in place. */
 
-				/* Add method equivalences for the methods that were re-defined (reverse of before!). */
-				rc = fixMethodEquivalences(currentThread, classPairs, jitEventDataPtr, TRUE, &methodEquivalences, extensionsUsed);
+				/* Add method equivalences for the methods that were re-defined (reverse of before!). Propagate any equivalent resolved callsites. */
+				rc = fixMethodEquivalencesAndCallSites(currentThread, classPairs, jitEventDataPtr, TRUE, &methodEquivalences, extensionsUsed);
 				if (rc != JVMTI_ERROR_NONE) {
 					goto failed;
 				}
@@ -1230,8 +1230,8 @@ redefineClassesCommon(jvmtiEnv* env,
 				/* Unresolve all classes */
 				unresolveAllClasses(currentThread, classPairs, methodPairs, extensionsUsed);
  
-				/* Update method equivalences */
-				rc = fixMethodEquivalences(currentThread, classPairs, jitEventDataPtr, FALSE, &methodEquivalences, extensionsUsed);
+				/* Update method equivalences. Propagate any equivalent resolved callsites. */
+				rc = fixMethodEquivalencesAndCallSites(currentThread, classPairs, jitEventDataPtr, FALSE, &methodEquivalences, extensionsUsed);
 				if (rc != JVMTI_ERROR_NONE) {
 					goto failed;
 				}
--- a/runtime/util/hshelp.c
+++ b/runtime/util/hshelp.c
@@ -71,6 +71,9 @@ static UDATA areClassRefsIdentical(J9ROMConstantPoolItem * romCP1, U_32 index1, 
 static UDATA areFieldRefsIdentical(J9ROMConstantPoolItem * romCP1, U_32 index1, J9ROMConstantPoolItem * romCP2, U_32 index2);
 static UDATA areNameAndSigsIdentical(J9ROMNameAndSignature * nas1, J9ROMNameAndSignature * nas2);
 static UDATA areSingleSlotConstantRefsIdentical(J9ROMConstantPoolItem * romCP1, U_32 index1, J9ROMConstantPoolItem * romCP2, U_32 index2);
+static UDATA areMethodsEquivalentPropagateCallSites(J9ROMMethod * method1, J9Class * ramClass1, J9ROMMethod * method2, J9Class * ramClass2);
+static UDATA areMethodsEquivalentSub(J9ROMMethod * method1, J9ROMClass * romClass1, J9Class * ramClass1, J9ROMMethod * method2, J9ROMClass * romClass2, J9Class * ramClass2);
+static UDATA areCallSiteDataMethodsEquivalent(J9ROMClass* romClass1, UDATA callSiteIndex1, J9ROMClass* romClass2, UDATA callSiteIndes2);
 static UDATA areDoubleSlotConstantRefsIdentical(J9ROMConstantPoolItem * romCP1, U_32 index1, J9ROMConstantPoolItem * romCP2, U_32 index2);
 static void fixClassSlot(J9VMThread* currentThread, J9Class** classSlot, J9HashTable *classPairs);
 static void fixJNIFieldIDs(J9VMThread * currentThread, J9Class * originalClass, J9Class * replacementClass);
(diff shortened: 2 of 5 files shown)


runtime/util/hshelp.c
```diff
@@ -423,6 +428,32 @@ fixJNIMethodID(J9VMThread *currentThread, J9Method *oldMethod, J9Method *newMeth
...
+	U_32 handleType1 = handleTypeAndCpType1 >> J9DescriptionCpTypeShift;
+	U_32 handleType2 = handleTypeAndCpType2 >> J9DescriptionCpTypeShift;
+	/* get cp types for comparison */
+	U_32 cpTypeIndex1 = handleTypeAndCpType1 & J9DescriptionCpTypeMask;
+	U_32 cpTypeIndex2 = handleTypeAndCpType2 & J9DescriptionCpTypeMask;
+	UDATA cpType1 = J9_CP_TYPE(J9ROMCLASS_CPSHAPEDESCRIPTION(romClass1), cpTypeIndex1);
+	UDATA cpType2 = J9_CP_TYPE(J9ROMCLASS_CPSHAPEDESCRIPTION(romClass2), cpTypeIndex2);
+
+
+	/* This method assumes that all method hadle references refer to methods and not fields */
```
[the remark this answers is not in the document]
theresa-m: What I meant by this is that is for invoke dynamic specifically the MH reference will always be a method at least from my understanding. Is that correct?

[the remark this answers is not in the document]
DanHeidinga: For an indy to succeed, it needs to be a method.  It's still possible to specify a Field MH but the indy resolution will fail as the BSM MH wouldn't have the right type.

runtime/util/hshelp.c
```diff
@@ -482,6 +513,99 @@ areSingleSlotConstantRefsIdentical(J9ROMConstantPoolItem * romCP1, U_32 index1,
...
+	additionalArgCount2 = bsmData2[1];
+	if (additionalArgCount1 != additionalArgCount2) {
+		return FALSE;
+	}
+
+	/* compare additional argument types. At this point we know that both call site entries have the same number of additional arguments.
+	 * each argument represents an index into the constant pool.
+	 */
+	for (i = 0; i < additionalArgCount1; i++) {
+		if (!areSingleSlotConstantRefsIdentical
```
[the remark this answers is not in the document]
theresa-m: I've updated this pr to include this. I've also added additional tests to cover this case in the test pr #5029

runtime/util/hshelp.c
```diff
@@ -482,6 +485,173 @@ areSingleSlotConstantRefsIdentical(J9ROMConstantPoolItem * romCP1, U_32 index1,
...
+/*
+ * Helper methods for areCallSiteDataMethodsEquivalent
+ */
+
+/* ROM call site data header size (16 bits of method handle address + additional argument count) */
+static const U_16 callSiteHeaderOffset = 2;
+
+static UDATA
+isMethodHandleAField(UDATA cpType) {
+	return (cpType == MH_REF_GETFIELD) || (cpType == MH_REF_GETSTATIC) || (cpType == MH_REF_PUTSTATIC) || (cpType == MH_REF_PUTFIELD);
```
DanHeidinga: I tend to use switch statements for this.  Not necessary to change, but something to think about for similar code:

```c
switch(cpType) {
case MH_REF_GETFIELD:
case MH_REF_GETSTATIC:
case MH_REF_PUTSTATIC:
case MH_REF_PUTFIELD:
    return true;
}
return false;
}
```

runtime/util/hshelp.c
```diff
@@ -482,6 +485,173 @@ areSingleSlotConstantRefsIdentical(J9ROMConstantPoolItem * romCP1, U_32 index1,
...
+}
+
+static U_32
+iterateToNextArgument(U_32 sigIndex, U_32 sigLength, U_8* sigData) {
+	if (sigIndex >= sigLength) return sigIndex;
+
+	/* check for object */
+	if (sigData[sigIndex] == 'L') {
+		while((sigIndex < sigLength) && (sigData[sigIndex] != ';')) {
+			sigIndex++;
```
DanHeidinga: We've been trying to switch to `sigIndex += 1` pattern rather than `++` as it tends to be clearer.

runtime/util/hshelp.c
```diff
@@ -482,6 +485,173 @@ areSingleSlotConstantRefsIdentical(J9ROMConstantPoolItem * romCP1, U_32 index1,
...
+	if ((TRUE == isMethodHandleAField(bsmCpType1)) || (TRUE == isMethodHandleAField(bsmCpType2))) {
+		return FALSE;
+	}
+
+	/* verify that bsm MethodHandles are the same:
+	 * step 1) verify that the handle types and cp types match
+	 * step 2) verify that the method references are the same (at this point we know they are
+	 * both method references)
+	 */
+	if (bsmHandleTypeAndCpType1 != bsmHandleTypeAndCpType2) {
```
DanHeidinga: If this is done before the `isMethodHandleAField()` call, then we only need to check if one of them is a field becasue we're sure they are the same type

theresa-m: updated.

runtime/jvmti/jvmtiClass.c
```diff
@@ -1230,8 +1230,8 @@ redefineClassesCommon(jvmtiEnv* env,
 				/* Unresolve all classes */
 				unresolveAllClasses(currentThread, classPairs, methodPairs, extensionsUsed);
  
-				/* Update method equivalences */
-				rc = fixMethodEquivalences(currentThread, classPairs, jitEventDataPtr, FALSE, &methodEquivalences, extensionsUsed);
+				/* Update method equivalences. Propogate any equivalent resolved callsites. */
```
gacholio: spelling - "Propagate"

gacholio: Function name below has similar error.

runtime/util/hshelp.c
```diff
@@ -482,6 +485,177 @@ areSingleSlotConstantRefsIdentical(J9ROMConstantPoolItem * romCP1, U_32 index1,
...
 
+/*
+ * Helper methods for areCallSiteDataMethodsEquivalent
+ */
+
+/* ROM call site data header size (16 bits of method handle address + additional argument count) */
+static const U_16 callSiteHeaderOffset = 2;
+
+static UDATA
+isMethodHandleAField(UDATA cpType) {
```
gacholio: { on its own line,

theresa-m: thanks. I've made your changes.

runtime/util/hshelp.c
```diff
@@ -482,6 +485,180 @@ areSingleSlotConstantRefsIdentical(J9ROMConstantPoolItem * romCP1, U_32 index1,
...
+	 * Structure of CallSiteData structure in ROM class (also see romclasswalk.c)
+	 * NAS:
+	 * 		callSiteCount SRP to J9ROMNameAndSignature structures describing the resolved callsite. Note: SRP is 32 bits
+	 * bsm indices:
+	 * 		callSiteCount 16 bit indices into bsmData. for example 0 is the first bsm, 1 is the second etc.
+	 * bsm data:
+	 * 		(header) The first 16 bits contains the index of the bsm's MethodHandle in the constant pool.
+	 * 		(header) The second 16 bits contains the number of bsm arguments in addition to the required three (MethodHandles.Lookup, String, MethodType).
+	 * 		The additional BSM arguments, these are constant pool indices.
+	 */
```
gacholio: I think it would be good to move the documentation up above all of the code which manipulates the data (and perhaps be a bit more explicit (i.e. describe it more like C structures).

runtime/util/hshelp.c
```diff
@@ -482,6 +485,180 @@ areSingleSlotConstantRefsIdentical(J9ROMConstantPoolItem * romCP1, U_32 index1,
...
+	U_32 bsmCpTypeIndex2 = bsmHandleTypeAndCpType2 & J9DescriptionCpTypeMask;
+	UDATA bsmCpType1 = J9_CP_TYPE(J9ROMCLASS_CPSHAPEDESCRIPTION(romClass1), bsmCpTypeIndex1);
+	UDATA bsmCpType2 = J9_CP_TYPE(J9ROMCLASS_CPSHAPEDESCRIPTION(romClass2), bsmCpTypeIndex2);
+	/* additional helpers */
+	J9ROMMethodRef* bsmMethod = NULL;
+	J9ROMNameAndSignature *bsmNAS = NULL;
+	J9UTF8 *bsmSig = NULL;
+	U_8* sigData = NULL;
+	U_32 sigIndex = 0;
+	U_32 sigLength = -1;
```
gacholio: Why -1? It's not valid for U_32, and the variable appears to never be read before the real value is assigned. 0 is probably a better default.

runtime/util/hshelp.c
```diff
@@ -492,8 +669,55 @@ areDoubleSlotConstantRefsIdentical(J9ROMConstantPoolItem * romCP1, U_32 index1,
...
+ * @param method1 first method to compare
+ * @param romClass1 ROM class associated with method1
+ * @param ramClass1 RAM class associated with method1, must be provided to propagate callsites
+ * @param method2 second method to compare
+ * @param romClass2 ROM class associated with method2
+ * @param ramClass2 RAM class associated with method2, must be provided to propagate callsites
+ * @return true if equivalent, false otherwise
+ */
+static UDATA
+areMethodsEquivalentSub(J9ROMMethod * method1, J9ROMClass * romClass1, J9Class * ramClass1, J9ROMMethod * method2, J9ROMClass * romClass2, J9Class * ramClass2)
```
gacholio: Do we need the ROM and RAM parameters (i.e. can ROM be fetch from RAM, or has there already been some field flipping which would make that invalid)?

gacholio:
Sorry, I accidentally started an real review, so if my comments don't show up, please let me know.

runtime/util/hshelp.c
theresa-m: The main reason for it is having a NULL RAM class input indicates that callsites don't need to be propagated when `areMethodsEquivalent` is called. `areMethodsEquivalentSub` is only supposed to be called by either `areMethodsEquivalent` or `areMethodsEquivalentPropagateCallSites`.

runtime/util/hshelp.c
gacholio: The new doc is great - the tabbing appears inconsistent in the github view, but we haven't come up with a good solution for this - if if looks good in Eclipse, I'm fine.

gacholio: @DanHeidinga Any formatting guidance here?

runtime/util/hshelp.c
```diff
@@ -482,6 +485,185 @@ areSingleSlotConstantRefsIdentical(J9ROMConstantPoolItem * romCP1, U_32 index1,
...
+	return sigIndex;
+}
+
+/**
+ * Verify callsite equivalence. 
+ * 
+ * Structure of CallSiteData structure in ROM class (also see romclasswalk.c):
+ * 
+ * romClass->callSiteData : {
+ *		SRP callSiteNAS[romClass->callSiteCount]; 		// structures describing the resolved callsite. Note: SRP is 32 bits
```
DanHeidinga: The indent here is two tabs - can it be decreased to 1 so the code fits better in this view?

It's hard to tell in this view but the description comments `// .....` look like they are lined up vertically.  A consistent # of spaces or single tab is probably fine

runtime/util/hshelp.c
```diff
@@ -482,6 +485,185 @@ areSingleSlotConstantRefsIdentical(J9ROMConstantPoolItem * romCP1, U_32 index1,
...
+ *		U_16 callSiteBSMIndex[romClass->callSiteCount]; // map from callSiteIndex value to bootStrapMethodData index
+ *		{
+ *			U_16 bootStrapMethodHandleRef; 				// (header) index of the bsm's MethodHandle in the constant pool.
+ *			U_16 argumentCount; 						// (header) number of bsm arguments in addition to the required three (MethodHandles.Lookup, String, MethodType).
+ *			U_16 argument[argumentCount]; 				// The additional BSM arguments, these are constant pool indices.
+ *		} bootStrapMethodData[romClass->bsmCount];
+ * }
```
DanHeidinga: Overall, this is a big improvement for docing this structure

runtime/util/hshelp.c
theresa-m: I've made these changes

runtime/util/hshelp.c
```diff
@@ -482,6 +485,185 @@ areSingleSlotConstantRefsIdentical(J9ROMConstantPoolItem * romCP1, U_32 index1,
 	return utfsAreIdentical(J9ROMSTRINGREF_UTF8DATA((J9ROMStringRef *) ref1), J9ROMSTRINGREF_UTF8DATA((J9ROMStringRef *) ref2));
 }
 
+/*
+ * Helper methods for areCallSiteDataMethodsEquivalent
+ */
+
+/* ROM call site data header size (16 bits of method handle address + additional argument count) */
+static const U_16 callSiteHeaderOffset = 2;
```
gacholio: This might be better as a define (I suspect this will create an actual variable in memory).

runtime/util/hshelp.c
```diff
@@ -482,6 +485,185 @@ areSingleSlotConstantRefsIdentical(J9ROMConstantPoolItem * romCP1, U_32 index1,
...
+	return bsmData;
+}
+
+static U_32
+iterateToNextArgument(U_32 sigIndex, U_32 sigLength, U_8* sigData) 
+{
+	if (sigIndex >= sigLength) return sigIndex;
+
+	/* check for object */
+	if (sigData[sigIndex] == 'L') {
```
gacholio: Constant should be on the left (and on the line below, strictly speaking).

runtime/util/hshelp.c
```diff
@@ -482,6 +485,185 @@ areSingleSlotConstantRefsIdentical(J9ROMConstantPoolItem * romCP1, U_32 index1,
...
+		while((sigIndex < sigLength) && (sigData[sigIndex] != ';')) {
```
gacholio: Space after while, please.

runtime/util/hshelp.c
```diff
@@ -482,6 +485,185 @@ areSingleSlotConstantRefsIdentical(J9ROMConstantPoolItem * romCP1, U_32 index1,
...
+		return FALSE;
+	}
+
+	/* Before continuing verify that the bsm MethodHandle refers to a Method and not a Field. The method
+	 * signature will be used to determine
+	 * whether the additional arguments are double or single slot which is impossible with a field.
+	 * The indy call will fail at a later point. It is only necessary to check one because from the previous
+	 * check both bsm1 and bsm2 are of the same type.
+	 */
+	if (TRUE == isMethodHandleAField(bsmCpType1)) {
```
gacholio: No need for TRUE ==

runtime/util/hshelp.c
```diff
@@ -482,6 +485,185 @@ areSingleSlotConstantRefsIdentical(J9ROMConstantPoolItem * romCP1, U_32 index1,
...
+		sigIndex = iterateToNextArgument(sigIndex, sigLength, sigData);
+	}
+
+	/* compare additional arguments */
+	for (i = 0; i < additionalArgCount1; i++) {
+		if (sigIndex >= sigLength) {
+			return FALSE;
+		}
+
+		if (sigData[sigIndex] == 'D' || sigData[sigIndex] == 'J') {
```
gacholio: Brackets please, and constants on the left.

runtime/util/hshelp.c
```diff
@@ -659,6 +888,25 @@ areMethodsEquivalent(J9ROMMethod * method1, J9ROMClass * romClass1, J9ROMMethod
 					bytecodeSize = (tempIndex + (4 * numEntries)) - index;
 					break;
 				}
+				case JBinvokedynamic: {
+					U_16 callSiteIndex1 = NEXT_U16(bytecodes1);
+					U_16 callSiteIndex2 = NEXT_U16(bytecodes2);
+					if (!areCallSiteDataMethodsEquivalent(romClass1, NEXT_U16(bytecodes1), romClass2, NEXT_U16(bytecodes2))) {
```
gacholio: Some confusion here around the macro - being cleaned up.

runtime/util/hshelp.c
```diff
@@ -659,6 +888,25 @@ areMethodsEquivalent(J9ROMMethod * method1, J9ROMClass * romClass1, J9ROMMethod
...
+						return FALSE;
+					}
+
+					/* if RAM classes are not provided do not attempt to propagate resolved callsite */
+					if ((NULL != ramClass1) && (NULL != ramClass2)) {
+						/* propagate call site data */
+						if (ramClass1->callSites[callSiteIndex1] != NULL) {
```
gacholio: Constant on the left.

theresa-m: updated.

runtime/util/hshelp.c
```diff
@@ -482,6 +485,185 @@ areSingleSlotConstantRefsIdentical(J9ROMConstantPoolItem * romCP1, U_32 index1,
 	return utfsAreIdentical(J9ROMSTRINGREF_UTF8DATA((J9ROMStringRef *) ref1), J9ROMSTRINGREF_UTF8DATA((J9ROMStringRef *) ref2));
 }
 
+/*
+ * Helper methods for areCallSiteDataMethodsEquivalent
+ */
+
+/* ROM call site data header size (16 bits of method handle address + additional argument count) */
+#define callSiteHeaderOffset 2
```
gacholio: Generally, constant defines should be upper case.

theresa-m: updated.

gacholio:

PR #5411, opened 2019-04-08
Fix travis CCACHE warning

keithc-ca:
Setting `CCACHE` in the environment does nothing but elicit this warning from configure:
```
configure: WARNING: Ignoring value of CCACHE from the environment. Use command line variables instead.
```


pshipton:
@dnakamura any concerns with this change?

dnakamura:
LGTM

Issue #5379, opened 2019-04-05
Test-sanity.system Mauve gnu.testlet.java.text.CollationElementIterator


pshipton:
https://ci.eclipse.org/openj9/job/Test-sanity.system-JDK8-linux_390-64_cmprssptrs/305
MauveSingleThreadLoadTest_OpenJ9_0
MauveMultiThreadLoadTest_OpenJ9_0
```
LT  PASS: gnu.testlet.java.text.CollationElementIterator.jdk11: NULLORDER (number 0)
LT  first = {5373952, 82, 0, 0}
LT  PASS: gnu.testlet.java.text.CollationElementIterator.jdk11: first initial test (number 0)
LT  next (2) = {5439488, 83, 0, 0}
LT  PASS: gnu.testlet.java.text.CollationElementIterator.jdk11: next() 2 initial test (number 0)
LT  PASS: gnu.testlet.java.text.CollationElementIterator.jdk11: no primary difference 2 initial test (number 0)
LT  next (3) = {5505024, 84, 0, 0}
LT  FAIL: gnu.testlet.java.text.CollationElementIterator.jdk11: next() 3 initial test (number 0)
LT  FAIL: gnu.testlet.java.text.CollationElementIterator.jdk11: no primary difference 3 initial test (number 0)
LT  next (4) = {5570560, 85, 0, 0}
LT  FAIL: gnu.testlet.java.text.CollationElementIterator.jdk11: next() 4 initial test (number 0)
LT  FAIL: gnu.testlet.java.text.CollationElementIterator.jdk11: no primary difference 4 initial test (number 0)
LT  next (5) = {5701632, 87, 0, 0}
LT  FAIL: gnu.testlet.java.text.CollationElementIterator.jdk11: next() 5 initial test (number 0)
LT  FAIL: gnu.testlet.java.text.CollationElementIterator.jdk11: no primary difference 5 initial test (number 0)
LT  next (6) = {5767168, 88, 0, 0}
LT  FAIL: gnu.testlet.java.text.CollationElementIterator.jdk11: next() 6 initial test (number 0)
LT  FAIL: gnu.testlet.java.text.CollationElementIterator.jdk11: no primary difference 6 initial test (number 0)
LT  next (7) = {5832704, 89, 0, 0}
LT  FAIL: gnu.testlet.java.text.CollationElementIterator.jdk11: next() 7 initial test (number 0)
[...]

PR #5421, opened 2019-04-09
Revert "Look for additional opportunities for stack allocation"

hzongaro:
This reverts commit c2eb8ed7f2ebc2f9d45d6d7c852027f0843747c1.

Revert change that allowed the right-hand side of an assignment to be considered harmless if it was an object that was made to be locally allocated by a previous pass of Escape Analysis.

I will continue to investigate how to fix the problem exposed/caused by this change so it can be re-enabled.

Fixes:  #5379


hzongaro:
@andrewcraik Andrew, may I ask you to review this pull request?

andrewcraik:
I am fine with a revert and then redeliver for this - we'll let travis finish and we should then be good to merge.

Issue #5380, opened 2019-04-05
The shared cache compatibility broken


hangshao0:
New bytecodes were introduced by #1831, which breaks the shared cache compatibility. The shared cache generation was not increased. 

A simple test case is:
```
class Test 
{ 
    public static boolean doPrint() {
		return true;
	}
	public static void main(String args[]) 
    { 
		if (doPrint()) {
			System.out.println("Hello, World"); 
		}
    } 
} 
```

I stored the class to the shared cache using an old JVM before #1831. 
```
java -Xshareclasses:name=Cache1 Test
Hello, World
```
Then startup the shared cache again using the newest JVM (with -Xshareclasses:readonly). 
`java -Xshareclasses:name=Cache1,readonly Test`

I get:
```
Exception in thread "main" java.lang.VerifyError: JVMVRFY012 stack shape inconsistent; class=Test, method=doPrint()Z, pc=1
Exception Details:
  Location:
    Test.doPrint()Z @1: JBireturn
  Reason:
    Type integer (current frame, stack[0]) is not assignable to integer
  Current Frame:
    bci: @1
    flags: { }
    locals: { }
    stack: { integer }
        at java.lang.J9VMInternals.prepareClassImpl(Native Method)
        at java.lang.J9VMInternals.prepare(J9VMInternals.java:304)
        at java.lang.Class.getMethodHelper(Class.java:1243)
        at java.lang.Class.getMethod(Class.java:1187)
        at sun.launcher.LauncherHelper.validateMainClass(LauncherHelper.java:556)
        at sun.launcher.LauncherHelper.checkAndLoadMain(LauncherHelper.java:538)
```

fjeremic:
@hangshao0 can you ensure whoever fixes this that we add a unit test into our sanity builds so this doesn't happen in a future change?

hangshao0:
FYI: @dnakamura

DanHeidinga:
Discussed with @hangshao0:
* OpenJ9 stores the SHA into the cache header and will normally attempt to delete and recreate the cache if the SHAs are different
* Under some circumstances, the cache can't be deleted (group access + readonly) and the VM continues to use it despite the SHA difference
* There are a large number of ways that the compatibility can be broken (romclass changes, new bytecodes, cache structure changes) but they are rare.

Need to find out more of the history to determine why the cache is not ignored when there is a SHA mismatch.

pshipton:
[...]

PR #5422, opened 2019-04-09
Increment the SharedClasses cache generation number

DanHeidinga:
When new bytecodes were added in #1831 for the truncated returns,
the SCC generation number wasn't incremented which can lead to
crashes.

Currently the VM will attempt to destroy and recreate the SCC
if the SHA in the header is different but if can't, it will
continue to use the cache.

The only sure way (today) to prevent this issue is to increment
the cache generation.

Do this now to prevent VerifyErrors due to unknown bytecodes
while we work out the right behaviour.

issue: #5380

Doc issue https://github.com/eclipse/openj9-docs/issues/248

### DIFF
--- a/runtime/shared_common/OSCache.hpp
+++ b/runtime/shared_common/OSCache.hpp
@@ -61,7 +61,7 @@
 #define OSCACHE_LOWEST_ACTIVE_GEN 1
 
 /* Always increment this value by 2. For testing we use the (current generation - 1) and expect the cache contents to be compatible. */
-#define OSCACHE_CURRENT_CACHE_GEN 35
+#define OSCACHE_CURRENT_CACHE_GEN 37
 
 #define J9SH_VERSION(versionMajor, versionMinor) (versionMajor*100 + versionMinor)
 


pshipton:
We should add a note to the release notes, as this will abandon existing caches on disk.

pshipton:
Created a doc issue https://github.com/eclipse/openj9-docs/issues/248

pshipton:
@doveye can we accommodate this change for the 0.14 release?

doveye:
@pshipton Yes we can fit this in. Re this sentence:

> Do this now to prevent VerifyErrors due to unknown bytecodes
while we work out the right behaviour.

It sounds like this is a temporary workaround (perhaps to be fixed in next release?), in which case we should update the release notes only rather than the docs. Please let me know if that assumption is wrong, thx.

pmhayward:
@pshipton : Issue added to 0.14 release notes (using wording in doc issue [https://github.com/eclipse/openj9-docs/issues/248](openj9-docs/issues/248) - thanks!) - see #5433. 
Please confirm whether or not this should also be in the docs 0.14 "What's new" topic. (See @doveye 's comment above.)

pshipton:
> Do this now to prevent VerifyErrors due to unknown bytecodes
while we work out the right behaviour.

I believe @DanHeidinga is considering changes to how the shared cache works, however incrementing the cache generation number is what has always been done when the content changes in an incompatible way. It is not a temporary change, once incremented it will never be reverted.

@DanHeidinga please create a PR for the 0.14 branch.

DanHeidinga:
v0.14.0 PR: https://github.com/eclipse/openj9/pull/5434

PR #5425, opened 2019-04-09
Explicitly link portlibrary against librt on linux

youngar:
With changes to OMR, the port library now requires linking against `librt` for the `shm_*` family of functions. This required change was missed during the OMR review, and currently the OMR acceptance build is failing. 

```
libjvm.so preloadLibrary(/home/jenkins/jenkins-agent/workspace/Build-JDK8-linux_ppc-64_cmprssptrs_le/build/linux-ppc64le-normal-server-release/images/j2re-image/lib/ppc64le/compressedrefs/libj9prt29.so): /home/jenkins/jenkins-agent/workspace/Build-JDK8-linux_ppc-64_cmprssptrs_le/build/linux-ppc64le-normal-server-release/images/j2re-image/lib/ppc64le/compressedrefs/libj9prt29.so: undefined symbol: shm_unlink
```

On most Linux systems, `libj9prt.so` already has a shared object dependency on `librt.so`, although this is not the case on the OSU provided Linux ppc le 64 machine.  We tested `OMR Acceptance (ppc le)` on a couple machines:

| | UNB | OSU |
|--- | --- | ---|
| JDK8 | Pass | Fail |
| JDK11 | Pass | Fail |
| JDK8+this PR | N/A | Pass |

With this change, `libj9prt.so` will always have a proper dependency on `librt.so`.

If possible, could we launch a build against the `eclipse/omr` `master` branch?

### DIFF
--- a/runtime/port/module.xml
+++ b/runtime/port/module.xml
@@ -351,6 +351,9 @@
 				<include-if condition="spec.aix_.*"/>
 				<include-if condition="spec.osx_x86.*"/>
 			</library>
+			<library name="rt" type="system">
+				<include-if condition="spec.linux_.*"/>
+			</library>
 			<library name="perfstat" type="system">
 				<include-if condition="spec.aix_.*"/>
 			</library>


pshipton:
jenkins test sanity plinux jdk8 depends eclipse/omr#master

PR #5426, opened 2019-04-09
Jitaas version checking

satbir-ca:
Client and server needs to be compatible.
Any kind of uncompatiblity will lead to the crash.
To ensure this, there will be a version check
between client and server. Every connection will
require one message exchange between client and
server for their version compatibility check.

If the compatibility check fail the connection
on the server will be closed and client will
see compilation failure.

### DIFF
--- a/runtime/compiler/control/CompilationThread.cpp
+++ b/runtime/compiler/control/CompilationThread.cpp
@@ -2072,6 +2072,7 @@ bool TR::CompilationInfo::shouldRetryCompilation(TR_MethodToBeCompiled *entry, T
             case compilationRecoverableCodeCacheError:
                tryCompilingAgain = true;
                break;
+            case compilationVersionCheckFailed:
             case compilationExcessiveComplexity:
             case compilationHeapLimitExceeded:
             case compilationLowPhysicalMemory:
@@ -3407,6 +3408,7 @@ void TR::CompilationInfo::stopCompilationThreads()
       {
       try
          {
+         JITaaS::J9ClientStream client(getPersistentInfo());
          getSequencingMonitor()->enter();
          uint32_t seqNo = getCompReqSeqNo();
          incCompReqSeqNo();
@@ -3419,11 +3421,10 @@ void TR::CompilationInfo::stopCompilationThreads()
          J9Class *clazz = NULL;
          J9Method *ramMethod = NULL;
          uint32_t romMethodOffset = 0;
-         JITaaS::J9ClientStream client(getPersistentInfo());
          bool useAotCompilation = false;
          client.buildCompileRequest(getPersistentInfo()->getJITaaSId(), romMethodOffset,
-                     ramMethod, clazz, cold, detailsStr, J9::EMPTY, unloadedClasses,
-                     classTuple, optionsStr, recompMethodInfoStr, seqNo, useAotCompilation);
+                  ramMethod, clazz, cold, detailsStr, J9::EMPTY, unloadedClasses,
+                  classTuple, optionsStr, recompMethodInfoStr, seqNo, useAotCompilation);
          }
       catch (const JITaaS::StreamFailure &e)
          {
@@ -6785,6 +6786,10 @@ TR::CompilationInfoPerThreadBase::shouldPerformLocalComp(const TR_MethodToBeComp
       if (entry->_optimizationPlan->getOptLevel() <= cold &&
          (TR::Options::getCmdLineOptions()->getOption(TR_EnableJITaaSHeuristics) || localColdCompilations))
          doLocalComp = true;
+
+      if (getClientStream())
+         if (!(getClientStream()->getVersionStatus()))
+            doLocalComp = true;
 
       return doLocalComp;
    }
--- a/runtime/compiler/control/JITaaSCompilationThread.cpp
+++ b/runtime/compiler/control/JITaaSCompilationThread.cpp
@@ -2445,6 +2445,7 @@ remoteCompile(
    TR_MethodMetaData *metaData = NULL;
    if (status.ok() && (statusCode == compilationOK || statusCode == compilationNotNeeded))
       {
+      client->setVersionStatus(true);
       try
          {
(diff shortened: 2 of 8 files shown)


satbir-ca:
@mpirvu Hi Marius the initial implementation of version check between client and server is ready.
It depend on the Harry's connection reduction issue. So I have created a dummy commit for his changes to create PR, please ignore that commit in review. Once the harry changes are integrated I will force push my single commit. This is WIP.

mpirvu:
The PR has no description of the solution adopted.
It would be nice to detail the design, discuss possible solutions and contrast them in terms of pros and cons.

satbir-ca:
it is there in the commit. I will add to the description also.

runtime/compiler/control/CompilationThread.cpp
```diff
@@ -2216,6 +2216,7 @@ bool TR::CompilationInfo::shouldRetryCompilation(TR_MethodToBeCompiled *entry, T
                entry->_optimizationPlan->setDisableGCR(); // GCR isn't needed
                tryCompilingAgain = true;
                break;
+            case compilationVersionCheckFailed:
```
mpirvu: When the version check fails the compilation needs to be retried locally

runtime/compiler/control/CompilationThread.cpp
```diff
@@ -3760,6 +3764,34 @@ TR::CompilationInfoPerThread::initializeSegmentCache(J9::J9SegmentProvider &segm
...
+   client->read(); //read and discard the J9ServerMessageType::compilationCode.
+   return std::get<0>(client->getRecvData<bool>()); //Read the version check result.
+   }
+
+bool
+TR::CompilationInfo::checkVersion(JITaaS::J9ServerStream *server)
+   {
+   auto recv = (server->read<uint32_t, uint32_t, uint32_t>());
+   bool result = false;
+   uint32_t clientMajorNumber = std::get<0>(recv);
```
mpirvu: I think we can do with just one uint32_t. The patch and the major version are not going to change too often, so we can use only 8 bits, while for the minor version we can use 16 bits.

Or, since the minor version never causes an incompatibility check, we could skip sending it. Thus we could use uint16_t for all 3 version numbers, but send only two of them (the major and the minor)

runtime/compiler/control/JITaaSCompilationThread.hpp
```diff
@@ -28,6 +28,10 @@
 #include "rpc/J9Client.hpp"
 #include "env/PersistentCollections.hpp"
 
+#define MAJOR_NUMBER 0
```
mpirvu: It's better if we attach some size to these for instance `static const uint_8 MAJOR_NUMBER = 0`

runtime/compiler/control/CompilationThread.cpp
```diff
@@ -3760,6 +3764,34 @@ TR::CompilationInfoPerThread::initializeSegmentCache(J9::J9SegmentProvider &segm
...
+   uint32_t ClientPatchNumber = PATCH_NUMBER;
+   client->write(clientMajorNumber , ClientMinorNumber, ClientPatchNumber);
+   client->read(); //read and discard the J9ServerMessageType::compilationCode.
+   return std::get<0>(client->getRecvData<bool>()); //Read the version check result.
+   }
+
+bool
+TR::CompilationInfo::checkVersion(JITaaS::J9ServerStream *server)
+   {
+   auto recv = (server->read<uint32_t, uint32_t, uint32_t>());
```
mpirvu: Why the extra parenthesis?

runtime/compiler/control/CompilationThread.cpp
```diff
@@ -3802,6 +3834,11 @@ TR::CompilationInfoPerThread::processEntries()
             if (compInfo->getPersistentInfo()->getJITaaSMode() == CLIENT_MODE && !client)
                {
                client = new (PERSISTENT_NEW) JITaaS::J9ClientStream(compInfo->getPersistentInfo());
+               if (!compInfo->checkVersion(client))
+                  {
+                     if (TR::Options::getVerboseOption(TR_VerboseJITaaS))
```
mpirvu: Indentation needs to be according to JIT coding convention

runtime/compiler/control/CompilationThread.cpp
```diff
@@ -3802,6 +3834,11 @@ TR::CompilationInfoPerThread::processEntries()
...
+                         TR_VerboseLog::writeLineLocked(TR_Vlog_JITaaS, "version check failed \n");
+                  }
```
mpirvu: What happens now that the version check failed? We need to perform the compilations locally.

satbir-ca:
@ymanton Please review the code.

runtime/compiler/rpc/J9Server.hpp
```diff
@@ -54,9 +54,24 @@ class J9ServerStream : J9Stream
...
       if (!_cMsg.status())
          throw StreamCancel();
+      //Version data will be present only in the first message for a connection.
+      if ((_versionCheckStatus == NOTDONE) &&
+            (version = _cMsg.version()))
+         {
+         if (!checkVersion(version))
+            {
+            _versionCheckStatus = FAILED;
+            throw StreamCancel();
```
mpirvu: The message for StreamCancel is "compilation canceled by client". I don't think this is the best type of exception to be thrown here.

runtime/compiler/control/JITaaSCompilationThread.cpp
```diff
@@ -2554,6 +2555,10 @@ remoteCompile(
          throw;
          }
       }
+   else if (statusCode == compilationVersionCheckFailed)
```
mpirvu: When `enableJITaaSPerCompConn` is false, the client stream will be destroyed. How will we do local compilations and how we will avoid future connection requests?

harryyu1994:
I could look into this as I dealt with a couple of server-client connections related issues.

PR #5431, opened 2019-04-10
Change mechanism to get memory stats in ExtendedOperatingSystemMXBeanImpl to be consistent

ShishirH:
Currently ExtendedOperatingSystemMXBeanImpl.getFreePhysicalMemorySize() has its own native method
getFreePhysicalMemorySizeImpl() which uses port library API j9vmem_get_available_physical_memory
for getting available memory. Other APIs use a method retrieveMemoryUsage() which accounts for
cgroup stats as well. Changed getFreePhysicalMemorySize() to  use the same for consistency.

Removed getFreePhysicalMemorySizeImpl().

Issue:5224

Signed-off-by: Shishir Halaharvi <shishir.neo95@gmail.com>


ashu-mehra:
From [Contributing.md](https://github.com/eclipse/openj9/blob/master/CONTRIBUTING.md):
```
use the keywords described in the following help article to automatically close issues. https://help.github.com/articles/closing-issues-using-keywords/
```

Instead of `Issue: ` you can use one of the keyword listed at https://help.github.com/articles/closing-issues-using-keywords/

runtime/jcl/common/com_ibm_jvm_Stats.c
```diff
@@ -78,11 +79,14 @@ Java_com_ibm_jvm_Stats_getStats(JNIEnv *env, jobject obj)
 		softmx = max;
 	}
 
+
+	j9sysinfo_get_memory_info(&memInfo);
```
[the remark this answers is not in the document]
ShishirH: Checking for the return code before setting the values

runtime/jcl/common/com_ibm_jvm_Stats.c
```diff
@@ -78,11 +79,14 @@ Java_com_ibm_jvm_Stats_getStats(JNIEnv *env, jobject obj)
...
+
 	/* Calculation of free physical memory from existing function in mgmtosext.c*/
-	free = Java_com_ibm_lang_management_internal_ExtendedOperatingSystemMXBeanImpl_getFreePhysicalMemorySizeImpl(env, obj);
+	free = (jlong) memInfo.availPhysical;
 
 	/* Calculation of total physical memory from port lib api */
-	tot = (jlong) j9sysinfo_get_physical_memory();
+	tot = (jlong) memAvail.totalPhysical();
```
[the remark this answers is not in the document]
ShishirH: Fixed

runtime/jcl/common/com_ibm_jvm_Stats.c
```diff
@@ -78,11 +80,16 @@ Java_com_ibm_jvm_Stats_getStats(JNIEnv *env, jobject obj)
...
 
-	/* Calculation of free physical memory from existing function in mgmtosext.c*/
-	free = Java_com_ibm_lang_management_internal_ExtendedOperatingSystemMXBeanImpl_getFreePhysicalMemorySizeImpl(env, obj);
 
-	/* Calculation of total physical memory from port lib api */
-	tot = (jlong) j9sysinfo_get_physical_memory();
+	rc = j9sysinfo_get_memory_info(&memInfo);
+	
+	if (0 == rc) { 
+		/* Calculation of free physical memory from existing function in mgmtosext.c*/
```
ashu-mehra: This comment is not more valid. Please remove it.

runtime/jcl/common/com_ibm_jvm_Stats.c
```diff
@@ -78,11 +80,16 @@ Java_com_ibm_jvm_Stats_getStats(JNIEnv *env, jobject obj)
...
+		free = (jlong) memInfo.availPhysical;
+
+		/* Calculation of total physical memory from port lib api */
```
ashu-mehra: Please remove this comment as well.

ashu-mehra:
Changes look fine to me once the stale comments are removed/updated.

runtime/jcl/common/com_ibm_jvm_Stats.c
ShishirH: Removed

runtime/jcl/common/com_ibm_jvm_Stats.c
ShishirH: Done

ShishirH:
@ashu-mehra Suggested changes made.

ShishirH:
@pshipton Can you review the PR please?

DanHeidinga:
@ShishirH Can you take a pass over the changes and ensure that tabs rather than spaces are used consistently in the change?  I'm seeing a number of places that mix the two.

ShishirH:
@DanHeidinga Should be consistent now.

runtime/jcl/common/com_ibm_jvm_Stats.c
```diff
@@ -78,11 +80,13 @@ Java_com_ibm_jvm_Stats_getStats(JNIEnv *env, jobject obj)
...
 
-	/* Calculation of free physical memory from existing function in mgmtosext.c*/
-	free = Java_com_ibm_lang_management_internal_ExtendedOperatingSystemMXBeanImpl_getFreePhysicalMemorySizeImpl(env, obj);
 
-	/* Calculation of total physical memory from port lib api */
-	tot = (jlong) j9sysinfo_get_physical_memory();
+	rc = j9sysinfo_get_memory_info(&memInfo);
+	
+	if (0 == rc) { 
+		free = (jlong) memInfo.availPhysical;
```
pshipton: This doesn't have the same behavior as the original code, which sets free to -1 if the system call fails.
`(0 == rc)? (jlong) size: (jlong) -1`

runtime/jcl/common/com_ibm_jvm_Stats.c
```diff
@@ -78,11 +80,13 @@ Java_com_ibm_jvm_Stats_getStats(JNIEnv *env, jobject obj)
 		softmx = max;
 	}
 
-	/* Calculation of free physical memory from existing function in mgmtosext.c*/
-	free = Java_com_ibm_lang_management_internal_ExtendedOperatingSystemMXBeanImpl_getFreePhysicalMemorySizeImpl(env, obj);
 
-	/* Calculation of total physical memory from port lib api */
-	tot = (jlong) j9sysinfo_get_physical_memory();
+	rc = j9sysinfo_get_memory_info(&memInfo);
```
pshipton: Looking at the AIX impl of this, there may be a bug. perfstat_memory_total() is documented as returning values in units of 4 KB pages, but the code uses `sysconf(_SC_PAGE_SIZE)` to calculate the availPhysical.

pshipton: omrvmem_get_available_physical_memory() seems implemented for all AIX, while retrieveAIXMemoryStats() is only implemented for !defined(J9OS_I5)

jcl/src/jdk.management/share/classes/com/ibm/lang/management/internal/ExtendedOperatingSystemMXBeanImpl.java
```diff
@@ -154,18 +154,15 @@ private double calculateProcessCpuLoad(long endTs, long endCpuTime, long startTs
 	 */
 	@Override
 	public final long getFreePhysicalMemorySize() {
-		return this.getFreePhysicalMemorySizeImpl();
+		try {
+			memObj = retrieveMemoryUsage(memObj);
```
pshipton: Does this affect performance? How much longer does it take to call retrieveMemoryUsage() than j9vmem_get_available_physical_memory().

pshipton:
For all platforms, I'd like to see a comparison of what retrieveMemoryUsage() returns for free physical compared to j9vmem_get_available_physical_memory(). I already commented about  a possible AIX bug. Looking at Linux, the implementation is quite different. We need to ensure the getFreePhysicalMemorySize() values are reasonable with the updated implementation.

runtime/jcl/common/com_ibm_jvm_Stats.c
```diff
@@ -78,11 +80,13 @@ Java_com_ibm_jvm_Stats_getStats(JNIEnv *env, jobject obj)
...
+	
+	if (0 == rc) { 
+		free = (jlong) memInfo.availPhysical;
+		tot = (jlong) memInfo.totalPhysical;
```
pshipton: Same -1 problem as free.

pshipton:
```
		OperatingSystemMXBean opBean = (OperatingSystemMXBean)ManagementFactory.getOperatingSystemMXBean();
		System.out.println("free physical: " + opBean.getFreePhysicalMemorySize());
		MemoryUsage memUsage = opBean.retrieveMemoryUsage();
		System.out.println("MemoryUsage free: " + memUsage.getFree());
```
AIX is giving different results:
```
free physical: 8208179200
MemoryUsage free: 7671291904
```

Windows is slightly different:
```
free physical: 131979538432
MemoryUsage free: 131979534336
```

Implements the feature on z/OS:
```
free physical: -1
MemoryUsage free: 7921360896
```

Other platforms I tried (xlinux, plinux, zlinux, osx) gave the same results for both.

pshipton:
```
	OperatingSystemMXBean opBean = (OperatingSystemMXBean)ManagementFactory.getOperatingSystemMXBean();
	long start1 = System.nanoTime();
	long res1 = opBean.getFreePhysicalMemorySize();
	long end1 = System.nanoTime();
	System.out.println("free physical: " + res1);
	System.out.println("ntime: " + (end1 - start1));
	System.out.println("utime: " + ((end1 - start1) / 1000));
	long start2 = System.nanoTime();
	MemoryUsage memUsage = opBean.retrieveMemoryUsage();
	long end2 = System.nanoTime();
	System.out.println("MemoryUsage free: " + memUsage.getFree());
	System.out.println("ntime: " + (end2 - start2));
	System.out.println("utime: " + ((end2 - start2) / 1000));
```

Windows is quite a bit slower:
```
free physical: 21675806720
ntime: 14800
utime: 14
MemoryUsage free: 21675806720
ntime: 249706700
utime: 249706
```

ashu-mehra:
fyi - @ShishirH is on vacation for few days.

ashu-mehra:
It appears the implementation of `omrsysinfo_get_memory_info` for AIX and Linux is not consistent in terms of page sizes considered for calculating free memroy. On Linux, `omrsysinfo_get_memory_info` sets `J9MemoryInfo.availPhysical` to a value that includes free memory for `HugePages` as well in addition to regular 4K page size, but AIX implementation does not take into account free memory for large pages (16M) and huge pages (16G). It only returns free memory for 4K and 64K pages.
On the other hand, `omrvmem_get_available_physical_memory` on AIX returns free memory for all page sizes.
We can change AIX implementation of `omrsysinfo_get_memory_info` to match with `omrvmem_get_available_physical_memory`.

But I think there may be another issue with `omrsysinfo_get_memory_info` returning free memory for all page sizes. JIT uses this value for determining memory for scratch space which uses default page size. So, I am guessing JIT would only be interested in free memory for default page size and not for all page sizes. @mpirvu is that correct?

mpirvu:
>  So, I am guessing JIT would only be interested in free memory for default page size and not for all page sizes. 

If those large pages are locked and cannot be converted to regular pages upon request, then yes, the JIT needs to know about free memory for default page size, because that's the only type of memory it ca use.

ashu-mehra:
> On Linux, `omrsysinfo_get_memory_info` sets `J9MemoryInfo.availPhysical` to a value that includes free memory for HugePages as well in addition to regular 4K page size

I was wrong in stating this. @ShishirH did some testing and we observed `J9MemoryInfo.availPhysical` does not include HugePages. `J9MemoryInfo.availPhysical` is set using `MemFree` field in `/proc/meminfo` which excludes free HugePages. So the current mechanism used by `omrsysinfo_get_memory_info` to set `J9MemoryInfo.availPhysical` is fine from JIT perspective.

But this raises the concern that `omrvmem_get_available_physical_memory` is inconsistent in its behavior across platforms. On AIX, it returns available memory for all page sizes, but on Linux it returns available memory only for default page size. 
I guess the correct behavior would depend on what do we expect `ExtendedOperatingSystemMXBean.getFreePhysicalMemorySize` (which uses `omrvmem_get_available_physical_memory`) to return - available memory for all page sizes or only for default page size?
@pshipton - what is your opinion?

pshipton:
I expect the user wants *usable* physical memory. In general I expect the default page size is correct, but perhaps we should extend the api so the caller can pass in the page size(s) desired. For example, if the JIT or object heap is the caller and is allocating large pages, then I expect the query should be about how much large page memory is free.

ashu-mehra:
> I expect the user wants usable physical memory. In general I expect the default page size is correct,

So that means on AIX replacing `omrvmem_get_available_physical_memory` with `omrsysinfo_get_memory_info` won't be a concern.
Only concern left is on windows where `omrsysinfo_get_memory_info` is taking lot more costlier than `omrvmem_get_available_physical_memory`. I believe we can fix that as well if we update `omrsysinfo_get_memory_info` to accept a parameter indicating type of information requested similar to `omrsysinfo_get_number_CPUs_by_type`. `omrsysinfo_get_memory_info` would then fill only the desired fields in `J9MemoryInfo` instead of getting everything from the OS.

>  if the JIT or object heap is the caller and is allocating large pages, then I expect the query should be about how much large page memory is free.

Agree, but I don't think we need it right away, as JIT/object heap are not querying for free memory.

genie-openj9:

PR #5432, opened 2019-04-10
AArch64: Implement _interfaceCallHelper in PicBuilder.spp

knn-k:
This commit implements _interfaceCallHelper in PicBuilder.spp for AArch64.

### DIFF
--- a/runtime/compiler/aarch64/codegen/CallSnippet.cpp
+++ b/runtime/compiler/aarch64/codegen/CallSnippet.cpp
@@ -343,9 +343,9 @@ uint8_t *TR::ARM64UnresolvedCallSnippet::emitSnippetBody()
       }
 
    // CP index and helper offset
-   *(int32_t *)cursor = (helperLookupOffset<<24) | methodSymRef->getCPIndexForVM();
+   *(intptrj_t *)cursor = (helperLookupOffset<<56) | methodSymRef->getCPIndexForVM();
 
-   return cursor + 4;
+   return cursor + 8;
    }
 
 uint32_t TR::ARM64UnresolvedCallSnippet::getLength(int32_t estimatedSnippetStart)
@@ -387,9 +387,9 @@ uint8_t *TR::ARM64VirtualUnresolvedSnippet::emitSnippetBody()
    cursor += 8;
 
    // CP index
-   *(int32_t *)cursor = methodSymRef->getCPIndexForVM();
+   *(intptrj_t *)cursor = methodSymRef->getCPIndexForVM();
 
-   return cursor + 4;
+   return cursor + 8;
    }
 
 uint32_t TR::ARM64VirtualUnresolvedSnippet::getLength(int32_t estimatedSnippetStart)
@@ -429,14 +429,14 @@ uint8_t *TR::ARM64InterfaceCallSnippet::emitSnippetBody()
    cursor += 8;
 
    // CP index
-   *(int32_t *)cursor = methodSymRef->getCPIndexForVM();
-   cursor += 4;
+   *(intptrj_t *)cursor = methodSymRef->getCPIndexForVM();
+   cursor += 8;
 
-   // Add 2 more slots for resolved values
-   *(int32_t *)cursor = 0;
-   cursor += 4;
-   *(int32_t *)cursor = 0;
-   cursor += 4;
+   // Add 2 more slots for resolved values (interface class and iTable offset)
+   *(intptrj_t *)cursor = 0;
+   cursor += 8;
+   *(intptrj_t *)cursor = 0;
+   cursor += 8;
 
    return cursor;
    }
--- a/runtime/compiler/aarch64/runtime/PicBuilder.spp
+++ b/runtime/compiler/aarch64/runtime/PicBuilder.spp
@@ -23,6 +23,7 @@
 #include "j9cfg.h"
 #include "jilconsts.inc"
 
+#define J9VMTHREAD x19
 #define J9SP x20
 
 	.globl	_interpreterUnresolvedStaticGlue
@@ -48,6 +49,7 @@
 	.globl	_interpreterDoubleStaticGlue
 	.globl	_interpreterSyncDoubleStaticGlue
 	.globl	_nativeStaticHelper
+	.globl	_interfaceDispatch
 
 	.extern	jitResolveClass
 	.extern	jitResolveClassFromStaticField
@@ -56,7 +58,10 @@
 	.extern	jitResolveStaticFieldSetter
 	.extern	jitResolveField
 	.extern	jitResolveFieldSetter
+	.extern	jitResolveInterfaceMethod
+	.extern	jitLookupInterfaceMethod
 	.extern	jitCallCFunction
+	.extern	jitThrowException
 	.extern	mcc_reservationAdjustment_unwrapper
 	.extern	mcc_callPointPatching_unwrapper
 	.extern	mcc_lookupHelperTrampoline_unwrapper
@@ -80,10 +85,18 @@
 	.set	J9TR_USCSnippet_CP,		20
 	.set	J9TR_USCSnippet_CPIndex,	28
 
-// Encoding of CPIndex field in USC snippet (helperOffset: 8 bits, cpIndex: 24 bits)
+// Encoding of CPIndex field in USC snippet (helperOffset: 8 bits, cpIndex: 56 bits)
 
-	.set	J9TR_USCSnippet_HelperOffset,	0xFF000000
-	.set	J9TR_USCSnippet_HelperOffsetShift,	24
+	.set	J9TR_USCSnippet_HelperOffset,	0xFF00000000000000
+	.set	J9TR_USCSnippet_HelperOffsetShift,	56
+
+// Interface call snippet
+
+	.set	J9TR_ICSnippet_codeCacheReturnAddress,	0
+	.set	J9TR_UICSnippet_CP,		8
+	.set	J9TR_UICSnippet_CPIndex,	16
+	.set	J9TR_ICSnippet_InterfaceClass,	24
+	.set	J9TR_ICSnippet_MethodIndex,	32
 
 	.text
 	.align 2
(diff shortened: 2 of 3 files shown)


knn-k:
This commit uses `L_refreshHelper` added by #5246.

knn-k:
Rebased to the latest version.

knn-k:
Resolved merge conflict with #5246.

runtime/compiler/aarch64/runtime/PicBuilder.spp
```diff
@@ -246,7 +259,68 @@ _virtualUnresolvedHelper:
...
+	ldp	x4, x5, [J9SP, #32]
+	ldp	x6, x7, [J9SP, #48]
+	add	J9SP, J9SP, #64
+	ldr	x15, [x14, x15]				// jump thru vtable
+	br	x15
+
+L_commonLookupException:
+	add	J9SP, J9SP, #64				// clean up stack but do not restore register values
+	ldr	x0, [J9VMTHREAD, #J9TR_VMThreadCurrentException]	// load pending exception from vmStruct
+	mov	x30, x15					// move correct LR in to get exception throw.
```
0xdaryl: From the three places in this code that can reach this exception throw (e.g., L271, L296), is x15 set up to something?  If so, how?

runtime/compiler/aarch64/runtime/PicBuilder.spp
```diff
@@ -246,7 +259,68 @@ _virtualUnresolvedHelper:
...
+	stp	x4, x5, [J9SP, #32]
+	stp	x6, x7, [J9SP, #48]
+	mov	x3, x30					// preserve LR
+	add	x0, x30, #J9TR_UICSnippet_CP			// get CP/index pair pointer
+	ldr	x1, [x30, #J9TR_ICSnippet_returnAddress]	// get code cache RA
+	mov	x4, x1						// protect RA in x4 (in L_commonLookupException, it is expected)
+	bl	jitResolveInterfaceMethod			// call the helper
+	cbz	x0, L_commonLookupException			// if resolve failed, throw the exception
+	add	x0, x3, #J9TR_Snippet_CallInstruction		// get address of BL instruction in snippet
+	ldr	x1, const_interfaceCompeteSlot2			// get new snippet branch target
```
0xdaryl: Why is this called `interfaceCompeteSlot2` ?  There's no "competition" among threads as far as I can tell.  It's really just an interface dispatch.  If there isn't any competition, can you rename it to something that won't cause confusion?

knn-k: This label name came from the arm32 version.  I am going to rename it.

runtime/compiler/aarch64/runtime/PicBuilder.spp
```diff
@@ -246,7 +259,68 @@ _virtualUnresolvedHelper:
 	hlt	#0	// Not implemented yet
 
 _interfaceCallHelper:
-	hlt	#0	// Not implemented yet
+	stp	x0, x1, [J9SP, #-64]!
+	stp	x2, x3, [J9SP, #16]
+	stp	x4, x5, [J9SP, #32]
+	stp	x6, x7, [J9SP, #48]
+	mov	x3, x30					// preserve LR
```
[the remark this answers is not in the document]
knn-k: Fixed.

runtime/compiler/aarch64/runtime/PicBuilder.spp
knn-k: It looks like I made a mistake in renaming registers.  Changing to use `x28` instead of `x4` and `x15`.

runtime/compiler/aarch64/runtime/PicBuilder.spp
```diff
@@ -85,6 +90,14 @@
 	.set	J9TR_USCSnippet_HelperOffset,	0xFF000000
 	.set	J9TR_USCSnippet_HelperOffsetShift,	24
 
+// Interface call snippet
+
+	.set	J9TR_ICSnippet_returnAddress,	0
+	.set	J9TR_UICSnippet_CP,		8
+	.set	J9TR_UICSnippet_CPIndex,	16
+	.set	J9TR_ICSnippet_InterfaceClass,	20
```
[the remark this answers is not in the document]
knn-k: Correcting `CallSnippet.cpp`.  They have to be 8 bytes (`UDATA` size) each.

https://github.com/eclipse/openj9/blob/c22a975433404b5ea94d1de663f6ee1329f90c21/runtime/codert_vm/cnathelp.cpp#L1049

runtime/compiler/aarch64/codegen/CallSnippet.cpp
```diff
@@ -432,11 +432,11 @@ uint8_t *TR::ARM64InterfaceCallSnippet::emitSnippetBody()
    *(int32_t *)cursor = methodSymRef->getCPIndexForVM();
    cursor += 4;
```
0xdaryl: I checked other platforms, and the cpIndex is expected to be in an 8-byte slot on 64-bits as well.

runtime/compiler/aarch64/runtime/PicBuilder.spp
```diff
@@ -85,6 +91,14 @@
 	.set	J9TR_USCSnippet_HelperOffset,	0xFF000000
 	.set	J9TR_USCSnippet_HelperOffsetShift,	24
 
+// Interface call snippet
+
+	.set	J9TR_ICSnippet_codeCacheReturnAddress,	0
+	.set	J9TR_UICSnippet_CP,		8
+	.set	J9TR_UICSnippet_CPIndex,	16
+	.set	J9TR_ICSnippet_InterfaceClass,	20
```
0xdaryl: If we change the size of cpIndex in the snippet then these offsets will need to be adjusted.

runtime/compiler/runtime/Runtime.cpp
```diff
@@ -1533,6 +1535,8 @@ void initializeCodeRuntimeHelperTable(J9JITConfig *jitConfig, char isSMP)
    SET(TR_ARM64interpreterDoubleStaticGlue,       (void *) _interpreterDoubleStaticGlue,     TR_Helper);
    SET(TR_ARM64interpreterSyncDoubleStaticGlue,   (void *) _interpreterSyncDoubleStaticGlue, TR_Helper);
    SET(TR_ARM64nativeStaticHelper,                (void *) _nativeStaticHelper,              TR_Helper);
+   SET(TR_ARM64interfaceCompeteSlot2,             (void *) _interfaceCompeteSlot2,           TR_Helper);
+   SET(TR_ARM64interfaceSlotsUnavailable,         (void *) _interfaceSlotsUnavailable,       TR_Helper);
```
0xdaryl: Are these two "helpers" required?  They're not used and likely a second-generation leftover from Power.  If we decide to employ a PIC solution on AArch64 then we can design and implement what we need then.

runtime/compiler/runtime/Runtime.cpp
```diff
@@ -585,6 +585,8 @@ JIT_HELPER(_interpreterSyncFloatStaticGlue);
 JIT_HELPER(_interpreterDoubleStaticGlue);
 JIT_HELPER(_interpreterSyncDoubleStaticGlue);
 JIT_HELPER(_nativeStaticHelper);
+JIT_HELPER(_interfaceCompeteSlot2);
+JIT_HELPER(_interfaceSlotsUnavailable);
```
0xdaryl: Not needed.  See below.

runtime/compiler/aarch64/runtime/PicBuilder.spp
0xdaryl: I would suggest just calling the label `interfaceDispatch` to give a bit more clarity.

knn-k: I cannot find a symbol `interfaceDispatch` in the JIT nor in the VM runtime.
What do you mean by this?

runtime/compiler/aarch64/codegen/CallSnippet.cpp
knn-k: Corrected the size not only this line but also in `ARM64UnresolvedCallSnippet` and `ARM64VirtualUnresolvedSnippet`.

runtime/compiler/aarch64/runtime/PicBuilder.spp
knn-k: Adjusted the offsets.

runtime/compiler/runtime/Runtime.cpp
knn-k: Removed these entries.

runtime/compiler/aarch64/runtime/PicBuilder.spp
0xdaryl: I looked into this a bit deeper.  This means:
* renaming the `_interfaceCompeteSlot2` label to `_interfaceDispatch`
* renaming `const_interfaceCompeteSlot2` to `const_interfaceDispatch`, and initializing it with `.dword _interfaceDispatch`
* renaming the helper index `TR_ARM64interfaceCompeteSlot2` to `TR_ARM64interfaceDispatch`
* initializing the `TR_ARM64interfaceDispatch` entry in the helpers table (for example, similar to this entry on ARM): 
 https://github.com/eclipse/openj9/blob/4f639567fd42ca81abab8023687e7942c7d27749/runtime/compiler/runtime/Runtime.cpp#L1431
* deleting `_interfaceSlotsUnavailable` because we don't use it

These are all remnants from a solution copied from Power that isn't used on AArch64 (or ARM for that matter).  If we develop an IPIC scheme for AArch64 then we'll have to architect something then that works best for AArch64.

knn-k: Applied those changes with `_interfaceCompeteSlot2` and `_interfaceSlotsUnavailable`.
It depends on https://github.com/eclipse/omr/pull/4105 for the new symbol `TR_ARM64interfaceDispatch`.

PR #5434, opened 2019-04-10
(v0.14.0) Increment the SharedClasses cache generation number

DanHeidinga:
(Port from `master` branch #5422) 

When new bytecodes were added in #1831 for the truncated returns,
the SCC generation number wasn't incremented which can lead to
crashes.

Currently the VM will attempt to destroy and recreate the SCC
if the SHA in the header is different but if can't, it will
continue to use the cache.

The only sure way (today) to prevent this issue is to increment
the cache generation.

Do this now to prevent VerifyErrors due to unknown bytecodes
while we work out the right behaviour.

issue: #5380

### DIFF
--- a/runtime/shared_common/OSCache.hpp
+++ b/runtime/shared_common/OSCache.hpp
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2001, 2018 IBM Corp. and others
+ * Copyright (c) 2001, 2019 IBM Corp. and others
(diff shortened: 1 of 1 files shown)


pshipton:
The copyright check failed.

pshipton:
@DanHeidinga can you please fix the copyright.

DanHeidinga:
Copyright now fixed

PR #5435, opened 2019-04-10
Always link portlibrary against rt on linux

youngar:
This is needed for the shm_* family of functions

Signed-off-by: Andrew Young <youngar17@gmail.com>

### DIFF
--- a/runtime/port/CMakeLists.txt
+++ b/runtime/port/CMakeLists.txt
@@ -157,7 +157,7 @@ if(OMR_OS_WINDOWS)
 	)
 endif()
 
-if(OMR_NEED_LIBRT)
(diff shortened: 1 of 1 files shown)


DanHeidinga:
@dnakamura Can you give this a quick once over?  It looks right to me

DanHeidinga:
Waiting on the travis build rather than launching a PR build as we don't have cmake pr testing yet

youngar:
`OMR_NEED_LIBRT` is only defined when the standard C library does not have a definition for `clock_gettime`.
https://github.com/eclipse/omr/blob/master/cmake/modules/platform/os/linux.cmake#L53

dnakamura:
LGTM

PR #5441, opened 2019-04-10
Delete jvmti_api.h

DanHeidinga:
Header is empty and unused.  Removing it as a left over from when
we previously enforced a `<dir>_api.h` & `<dir>_internal.h` pattern


pshipton:
jenkins compile win jdk8

Issue #2662, opened 2018-08-20
Refactor MH.resolveConstantDynamic/InvokeDynamic


fengxue-IS:
There are duplicate code between `MethodHandle.resolveConstantDynamic and resolveInvokeDynamic` that could be merged into a common helper method

From #1533

PR #5442, opened 2019-04-10
Create common method for MH.resolveCondy/Indy

theresa-m:
Remove duplicate code by creating common method for MethodHandle.resolveConstantDynamic and resolveInvokeDynamic. Verified indy/condy tests on Java 8 and 11.

Fixes: #2662

### DIFF
--- a/jcl/src/java.base/share/classes/java/lang/invoke/MethodHandle.java
+++ b/jcl/src/java.base/share/classes/java/lang/invoke/MethodHandle.java
@@ -960,6 +960,150 @@ public abstract class MethodHandle
 	private static final int BSM_TYPE_ARGUMENT_INDEX = 2;
 	private static final int BSM_OPTIONAL_ARGUMENTS_START_INDEX = 3;
 	
+	/**
+	 * Gets class object from RAM class address
+	 * @param j9class RAM class address
+	 * @return class object
+	 * @throws Throwable any throwable will be handled by the caller
+	 */
+	private static final Class<?> getClassFromJ9Class(long j9class) throws Throwable {
+		Class<?> classObject = null;
+		if (JITHELPERS.is32Bit()) {
+			classObject = JITHELPERS.getClassFromJ9Class32((int)j9class);
+		} else {
+			classObject = JITHELPERS.getClassFromJ9Class64(j9class);
+		}
+		Objects.requireNonNull(classObject);
+		return classObject;
+	}
+
+	/**
+	 * Retrieve a static argument of the bootstrap method at argIndex from constant pool
+	 * @param access
+	 * @param internalRamClass
+	 * @param classObject RAM class object
+	 * @param bsm bootstrap method
+	 * @param bsmArgs starting address of bootstrap arguments in the RAM class call site
+	 * @param bsmTypeArgCount number of bootstrap arguments
+	 * @param argIndex index of bsm argument starting from zero. Argument should be in addition to the first three 
+	 * 	mandatory arguments (3+): MethodHandles.Lookup, String, MethodType
+	 * @return additional argument from the constant pool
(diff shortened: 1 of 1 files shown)


pshipton:
jenkins copyright check

pshipton:
jenkins line endings check

DanHeidinga:
@fengxue-IS Can you review this?

fengxue-IS:
looking into this

jcl/src/java.base/share/classes/java/lang/invoke/MethodHandle.java
```diff
@@ -1246,57 +1124,66 @@ private static final MethodHandle resolveInvokeDynamic(long j9class, String name
...
-			} catch (NoSuchMethodException nsme) {
-				throw new Error(nsme);
+			if (indyNotCondy) { /* invoke dynamic */
+				if (type == null) {
+					throw new BootstrapMethodError(e);
+				}
+				
+				/* create an exceptionHandle with appropriate drop adapter and install that */
+				try {
+					MethodHandle thrower = MethodHandles.throwException(((MethodType)type).returnType(), BootstrapMethodError.class);
```
theresa-m: Hey @DanHeidinga Jack and I were talking through this error handling section but neither of us were too sure about the purpose of these few lines which ends up returning a MethodHandle that will throw a BootstrapMethodError instead of just throwing the error outright like in constant dynamic. Can you help us understand why this is done the way it is?

DanHeidinga: We need to record the failure so that the same error is thrown everytime.  If we didn't record the error, we'd have to rerun the bootstrap method the next time the invokedynamic instruction was executed.  The spec says we need to rethrow the same error

theresa-m:
note that #5442 touches the same code as this refactor.

DanHeidinga:
Another approach to this is to break the common code into a set of helpers and call them from the indy & condy path.  It might be worth sketching out that solution to see which ends up cleaner.

ie:  create helpers like:
```java
 Class<?> getClassFromJ9Class(long j9class) ;
Object[] getStaticArguments(Object internalRamClass, ConstantPool cp, int bsmArgCount, Object[] staticArgs) ;
....
```

theresa-m:
I played around with it a bit more and I think breaking out methods like you suggest is going to make more sense though there will be slightly less common code. All the type casting and extra versioning comments do make the current iteration hard to read.

theresa-m:
I've updated this pr to a version that breaks the common code into methods and merged the changes from #6029. The old version is here https://github.com/eclipse/openj9/compare/master...theresa-m:indy_refactor_old

jcl/src/java.base/share/classes/java/lang/invoke/MethodHandle.java
```diff
@@ -1144,14 +1185,7 @@ private static final MethodHandle resolveInvokeDynamic(long j9class, String name
...
-			if (JITHELPERS.is32Bit()) {
-				classObject = JITHELPERS.getClassFromJ9Class32((int)j9class);
-			} else {
-				classObject = JITHELPERS.getClassFromJ9Class64(j9class);
-			}
-			
-			Objects.requireNonNull(classObject);
+			Class<?> classObject = getClassFromJ9Class(j9class);
 			
 			type = MethodType.vmResolveFromMethodDescriptorString(methodDescriptor, access.getClassloader(classObject), null);
```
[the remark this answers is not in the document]
theresa-m: `type` is used in the catch block so it should remain outside. I've updated Class<?> typeClass for condy

DanHeidinga:
Jenkins test sanity zlinux jdk8,jdk11

PR #5445, opened 2019-04-10
Ensure consistency during relocation failure

### LINKED ISSUE #5375: JIT Code Cache committed value too high
After testing issue [4674](https://github.com/eclipse/openj9/issues/4674) I have found that the JIT Code Cache committed value as reported by the MemoryPool MBean is reported as the maximum limit (i.e 256M by defafult) rather than what is actually allocated.
This also affects the Java core JIT Code Cache allocation summary data.

With a default 256M JIT Code Cache the value of : 268,435,456 bytes is reported in both cases even though only ~12M was allocated.

dsouzai:
No inlined method relocations can be ignore (with or without the SVM)
because of symref sharing. The way the relocation infrastructure ensured
this was by calling the failCompilation method. However, this is not
consistent with the way error handling is performed. Specifically, this
resulted in the compilation thread failing to unreserve the code cache.

This PR ensures that the case in which the relocation infrastructure
cannot ignore a relocation that it cannot perform conforms to the proper
error handling process.

Fixes #5375

### DIFF
--- a/runtime/compiler/runtime/RelocationRecord.cpp
+++ b/runtime/compiler/runtime/RelocationRecord.cpp
@@ -179,14 +179,30 @@ TR_RelocationRecordGroup::handleRelocation(TR_RelocationRuntime *reloRuntime,
    if (reloRuntime->reloLogger()->logEnabled())
       reloRecord->print(reloRuntime);
 
-   if (reloRecord->ignore(reloRuntime))
+   switch (reloRecord->action(reloRuntime))
       {
-      RELO_LOG(reloRuntime->reloLogger(), 6, "\tignore!\n");
-      return 0;
+      case TR_RelocationRecordAction::apply:
+         {
+         reloRecord->preparePrivateData(reloRuntime, reloTarget);
+         return reloRecord->applyRelocationAtAllOffsets(reloRuntime, reloTarget, reloOrigin);
+         }
+      case TR_RelocationRecordAction::ignore:
+         {
+         RELO_LOG(reloRuntime->reloLogger(), 6, "\tignore!\n");
+         return 0;
+         }
+      case TR_RelocationRecordAction::failCompilation:
+         {
+         RELO_LOG(reloRuntime->reloLogger(), 6, "\tINTERNAL ERROR!\n");
+         return compilationAotClassReloFailure;
+         }
+      default:
+         {
+         TR_ASSERT_FATAL(false, "Unknown relocation action %d\n", reloRecord->action(reloRuntime));
+         }
       }
 
-   reloRecord->preparePrivateData(reloRuntime, reloTarget);
-   return reloRecord->applyRelocationAtAllOffsets(reloRuntime, reloTarget, reloOrigin);
+   return compilationAotClassReloFailure;
    }
 
 #define FLAGS_RELOCATION_WIDE_OFFSETS   0x80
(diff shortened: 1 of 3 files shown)


dsouzai:
@jdmpapin @mstoodle could you please review?

pshipton:
Reminder that after this is reviewed and ready to merge, we'll also need a PR to deliver the same change to the 0.14 branch.

dsouzai:
>Reminder that after this is reviewed and ready to merge, we'll also need a PR to deliver the same change to the 0.14 branch.

Ah ok, thanks for letting me know.

pshipton:
To the committer. The release process to handle release branches:
- review and get the final commit in this PR
- ask the author to cherry pick the commit and create a PR against the release branch, tag it with the milestone (0.14)
- double check the release branch PR
- merge both PRs

fjeremic:
Jenkins test sanity all

runtime/compiler/runtime/RelocationRecord.cpp
```diff
@@ -600,21 +606,15 @@ TR_RelocationRecord::computeHelperAddress(TR_RelocationRuntime *reloRuntime, TR_
 #undef FLAGS_RELOCATION_FLAG_MASK
 
 
-bool
+TR_RelocationRecordIgnoreCode
 TR_RelocationRecord::ignore(TR_RelocationRuntime *reloRuntime)
```
mstoodle: I don't like this being called `ignore()` anymore, given how you're changing it. You're asking more a question like "what should I do with this relocation record?" .  Maybe `action()` ? rather than `ignore()`? then the current set of reasonable answers can be "ignore it", "apply relocation", "fail compilation", or any other variant that makes sense (as opposed to "internal error")...

You're not really asking if you can ignore it, you're asking more a question about what action to perform for this relocation: ignore it, perform it, or "internal error". I feel that would be a bit more clear than asking `ignore()` and is

dsouzai: Heh naming is hard; I was trying to think of what to change `ignore()` to, but I just wasn't creative enough :P. I'll make the requested change.

dsouzai:
@mstoodle made the requested changes, could you please take a look.

dsouzai:

mstoodle:
changes look good, let's be pedantic on testing

mstoodle:
actually before I do that, @mpirvu anything you would like to see changed?

dsouzai:
> Is there value in being more specific on which subtype of relocation failure occured? 

I think there is, but I think it's probably better to have relocation error codes that can be printed out via an option, rather than introduce new types of compilation error codes. That's something that could be done as part of the AOT cleanup work.

dsouzai:
@mstoodle opened https://github.com/eclipse/openj9/pull/5463 (fyi @pshipton)

mpirvu:

mstoodle:

mpirvu:
heh, almost simultaneous testing requests

dsouzai:
The `Sanity - JDK11 - linux_390-64_cmprssptrs` build failed because of a crash in javac; I don't see how it could be related to this change since only the relo files are modified, and javac doesn't run with the SCC.

dsouzai:
`Sanity - JDK8 - linux_390-64_cmprssptrs` is a jenkins error.

dsouzai:
jenkins test sanity zlinux jdk8,jdk11

mpirvu:
The 4 failures are due to infra problems

pshipton:
`Sanity - JDK8 - linux_x86-64_cmprssptrs` passed, but shows failed due to jenkins issues.

Rerunning the other PR testing that failed due to jenkins issues.

jenkins test sanity zlinux,plinux jdk8

pshipton:
jenkins test sanity win jdk11

pshipton:
jenkins test sanity aix,xlinux jdk11

pshipton:

pshipton:
Seems zlinux jdk8 PR testing is broken, opened https://github.com/eclipse/openj9/issues/5471

AdamBrousseau:

pshipton:
Again, note I didn't restart `Sanity - JDK8 - linux_x86-64_cmprssptrs` because it passed, but shows failed due to jenkins issues.
```
19:45:01  TOTAL: 191   EXECUTED: 112   PASSED: 112   FAILED: 0   SKIPPED: 79
19:45:01  ALL TESTS PASSED
```

jdmpapin:
Is it possible to make `failCompilation()` work properly in this scenario instead? There is a pervasive assumption that it's safe to `failCompilation()` everywhere we use `SVM_ASSERT()` and its variants

dsouzai:
@jdmpapin good point; I added a try/catch around the call to `applyRelocations` to ensure that we do run the cleanup code. However, I left the original commit in (even though it is no longer strictly needed anymore) because I think the refactoring is a benefit.

dsouzai:

dsouzai:
@mstoodle for your review :)

dsouzai:
https://openj9.slack.com/archives/C8PQL5N65/p1555094450000400
>Peter Shipton [2:40 PM]
>Jenkins falling over, restarted it. PR testing will need to be restarted afterwards
>
>AdamBrousseau [2:59 PM]
>Opened a critical buzilla


Looks like all the builds are hung right now; is it worth restarting the builds until the bugzilla gets resolved?

pshipton:
We restarted jenkins, we should be fine to run PRs for a while.

dsouzai:

PR #5451, opened 2019-04-11
Remove aarch64-linux-gnu- from native build

knn-k:
The native build on AArch64 Linux uses aarch64-linux-gnu-gcc and other
commands with the aarch64-linux-gnu- prefix, which are used in the cross
build.
This commit removes the use of those unnecessary prefix from the native
build.

### DIFF
--- a/runtime/gc_glue_java/configure_includes/configure_linux_aarch64.mk
+++ b/runtime/gc_glue_java/configure_includes/configure_linux_aarch64.mk
@@ -49,25 +49,24 @@ ifneq (,$(findstring _cross,$(SPEC)))
 	--host=$(OPENJ9_CC_PREFIX) \
 	--build=x86_64-pc-linux-gnu \
 	'OMR_CROSS_CONFIGURE=yes'
-endif
 
-ifeq (default,$(origin AS))
-	AS = $(OPENJ9_CC_PREFIX)-as
-endif
-ifeq (default,$(origin CC))
-	CC = $(OPENJ9_CC_PREFIX)-gcc
-endif
-ifeq (default,$(origin CXX))
-	CXX = $(OPENJ9_CC_PREFIX)-g++
-endif
-ifeq (default,$(origin AR))
-	AR = $(OPENJ9_CC_PREFIX)-ar
+	ifeq (default,$(origin AS))
+		AS = $(OPENJ9_CC_PREFIX)-as
+	endif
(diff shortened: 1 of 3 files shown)


runtime/gc_glue_java/configure_includes/configure_linux_aarch64.mk
```diff
@@ -49,25 +49,24 @@ ifneq (,$(findstring _cross,$(SPEC)))
...
+	endif
+	ifeq (default,$(origin CC))
+		CC = $(OPENJ9_CC_PREFIX)-gcc
+	endif
+	ifeq (default,$(origin CXX))
+		CXX = $(OPENJ9_CC_PREFIX)-g++
+	endif
+	ifeq (default,$(origin AR))
+		AR = $(OPENJ9_CC_PREFIX)-ar
+	endif
```
DanHeidinga:
```
<property name="uma_make_cmd_ranlib" value="$(OPENJ9_CC_PREFIX)-ranlib"/>
```
Should ranlib be in this list to get prefixed as well?  It was previously prefixed in the spec file

DanHeidinga: Same question for `<property name="uma_make_cmd_cpp" value="cpp -E -P"/>`

sxa555: Possibly true although `-ar` is the only one that seemed to be actively blocking the builds from running.

sxa555: Possibly, although `ar` was the only one whose absence was causing the build to fail (Verified by adding a symlink in place on the build machine)

For reference, this could easily be missed by someone building on Ubuntu since it comes with the "long" names by default in the package as well as the short ones, whereas RHEL7 does not.

knn-k: The cross build finishes fine with the change.  As long as I have checked in the build log, they are not used at all.

knn-k:
Are properties `uma_make_cmd_*` used anywhere?
I can find only the definitions of those properties in `buildspecs/*.spec`, regardless of the target platforms.

pshipton:
I find `compiler="{$spec.property.uma_make_cmd_cc$}"` in code internal to IBM.

PR #5461, opened 2019-04-11
Trigger release VM Access hook earlier

amicic:
Trigger the hook before handling for exclusive VM access (possible
notification that exclusive access has been acquired). If called after
the notification, then a party that requested exclusive VM access could
run and race the hook, and the hook could occur after the party
performed the action that could be interested in the hook trigger.

It also makes the code sequence more symmetrical to the one in VM access
acquire:
- in acquire: exclusive VM access handling, invoke hook and then set
thread VM access bit
- in release: reset VM access bit, invoke hook, then exclusive VM access
handling

Fixes https://github.com/eclipse/openj9/issues/5389


gacholio:
jenkins test sanity win jdk8

PR #5463, opened 2019-04-11
(0.14.0) Ensure consistency during relocation failure

dsouzai:
No inlined method relocations can be ignore (with or without the SVM)
because of symref sharing. The way the relocation infrastructure ensured
this was by calling the failCompilation method. However, this is not
consistent with the way error handling is performed. Specifically, this
resulted in the compilation thread failing to unreserve the code cache.

This PR ensures that the case in which the relocation infrastructure
cannot ignore a relocation that it cannot perform conforms to the proper
error handling process.


dsouzai:
0.14 version of https://github.com/eclipse/openj9/pull/5445

Issue #643, opened 2017-11-21
Investigate making compressed references a runtime and not compile time decision


charliegracie:
Currently the OpenJ9 VM can not support both compressed references and not compressed references for 64 bit JVMs at the same time.  This means to create a JDK that supports both you need to have 2 JVMs which significantly increases the footprint of the JDK package, the complexity of creating the JDK and the time it takes to create the JDK.

I am creating this Issue to start the conversation about what needs to be completed to make compressed references a runtime decision such that the JVM supports both at the same time.

charliegracie:
My initial thoughts are as follows:
1. We likely need to compile the interpreter multiple times and pick the right one based on whether or not compressed references is enabled.
2. The JIT already queries the GC for most of it's compressed references decisions so I am hoping it is not that hard to complete the extra work there.
3. The GC will likely have to deal with minor performance regressions for tight loops reading / writing object references
4. Currently there are 3 (maybe more) compile flags that control whether or not compressed references can be used.  I would suggest we move all of this code under a single compiler flag to make the rest of these changes easier.

andrewcraik:
Whatever mechanism we design for the JIT to query the compressed reference configuration - we need to make sure that it is very cheap to run - the JIT should not have to call over the VM every time to get the state otherwise we will see a compile-time penalty that could become prohibitive. Whether we do this by caching the VM result in the JIT to make it easier to inline or something I don't know, but just a point for consideration.

I also assume this item is to allow it to be a runtime decision, but a decision taken once at the start of a given VM so that we don't have to worry about it changing "on the fly"?

charliegracie:
This would be something decided on startup that can not change at runtime after we have initialized the JVM.  I agree this would be something that could be cached in the JIT or any other component for speed purposes.

ymanton:
> 2. The JIT already queries the GC for most of it's compressed references decisions so I am hoping it is not that hard to complete the extra work there.

The assembly routines that use the compile-time flags will be the most problematic to deal with.

JamesKingdon:
How much work would it take to get a prototype running so that we can assess the performance impact?

mstoodle:
Only 51 occurrences of *COMPRESSED* in the JIT that are in assembler files. Anything else I believe could just be taught to use a cached query result so I'm not particularly worried about them.

Of those 51, 10 are && J9VM_OPT_TENANT so presumably that's just dead code...

[...]

PR #5465, opened 2019-04-11
Fix incorrect macros in m4 files

gacholio:
Change from compressed header to compressed pointers caused an issue due
to the m4 files not using the ASM_ version of the build flag macros.

[ci skip]


gacholio:
Merging immediately to fix build break on Z.

gacholio:
Related to #643

Issue #5373, opened 2019-04-04
Performance of ManagementFactory.getRuntimeMXBeans is high on Java 11 and 12


NottyCode:
Java -version output
--------------------
openjdk version "12" 2019-03-19
OpenJDK Runtime Environment AdoptOpenJDK (build 12+33)
Eclipse OpenJ9 VM AdoptOpenJDK (build openj9-0.13.0, JRE 12 Mac OS X amd64-64-Bit Compressed References 20190320_32 (JIT enabled, AOT enabled)
OpenJ9   - caeb51f87
OMR      - 33a33ff2
JCL      - 419f571019 based on )

Summary of problem
------------------

Calls to ManagementFactory.getXXXMXBean are significantly more expensive on Java 11 & 12 than Java 8. Here are some example outputs (each item is `time java Ut`):

Java 8
```26458@alasdaismbp3089.war.can.ibm.com uptime: 113
getMbean:  31
getName:   0
getUptime: 0

real	0m0.179s
user	0m0.175s
sys	0m0.054s
```
Java 11
```
26532@alasdaismbp3089.war.can.ibm.com uptime: 274
getMbean:  138
getName:   1
getUptime: 0

real	0m0.360s
user	0m0.484s
sys	0m0.077s
```

Java 12
```
26495@alasdaismbp3089.war.can.ibm.com uptime: 363
getMbean:  214
getName:   0
getUptime: 0

real	0m0.534s
user	0m0.512s
sys	0m0.155s
```

At first I thought this related to the class libraries, but when I run an adoptopenjdk hotspot build I get this output:

Java 12
```
26538@alasdaismbp3089.war.can.ibm.com uptime: 231
getMbean:  63
getName:   46
getUptime: 1

real	0m0.305s
user	0m0.198s
sys	0m0.081s
```

It is possible that on hotspot the classes are being initialized sooner, but uptime (returned from the RuntimeMXBean.getUptime is higher in the OpenJ9 case and while that could be due to implementation details the time output from the shell suggests that OpenJ9 is taking much longer.

Diagnostic files
----------------

The test code

```
import java.lang.management.*;

public class Ut {

  public static void main(String[] args) {
    long st = System.currentTimeMillis();
    RuntimeMXBean mbean = ManagementFactory.getRuntimeMXBean();
    long st2 = System.currentTimeMillis();
    String pid = mbean.getName();
    long st3 = System.currentTimeMillis();
    long ut = mbean.getUptime();
[...]

PR #5468, opened 2019-04-12
Remove ManagementUtils.convertToOpenType()

keithc-ca:
* it was only used in test code and there's public API to achieve the same goals
* this allows for removal of all the registrations of composite type converters which had a significant impact on startup performance (see #5373)
* also use a nested class rather than a lambda (via a method handle) to register available beans
* update management tests


pshipton:
jenkins test sanity plinux jdk8,jdk11

pshipton:
jenkins test extended plinux jdk8,jdk11

pshipton:
@keithc-ca please create a PR for the 0.14 branch as well.

PR #5470, opened 2019-04-12
Revert "Add VM function to query if the JVM supports VTs"

pshipton:
Reverts eclipse/openj9#4918

Causing some test failures, reverting while we figure it out.


DanHeidinga:
fyi @tajila

PR #5475, opened 2019-04-12
WIP: Cache resolved methods globally

dmitry-ten:
Implement global, per RAM class, cross-compilation caching of
resolved methods on the server.
The main challenge with global caching is that resolved methods
are allocated in heap memory, on both the client and the server,
thus they only exist for the duration of one compilation.
We could change allocation to persistent, but that would increase
memory footprint on the client, which should be avoided at all costs.
I addressed this problem by storing data needed to create a resolved
method persistently on the server, and using "on-demand" mirror creation,
i.e. client mirror is recreated only when it's needed.
This global caching is done in addition to local per-compilation caching
that was already being done. I left it in, because local caching is
cheaper, as it doesn't require global lock on RAM class table, and
it doesn't need to recreate mirror on the client.

This is a brief description of the solution I used:
1. Server creates a resolved method for the first time, by making
a remote call to the client.

2. Server saves attributes used to create resolved method
(i.e. `TR_ResolvedJ9JITaaSServerMethodInfo`) in a persistent map
in `ClassInfo`. `ClassInfo` corresponds to class to which this
resolved method belongs. It also saves resolved method into
a local cache.

3. The next time server needs to create the same resolved method,
it will look into the local cache first.
If local cache doesn't have it, server will look into a global cache.
It will find attributes that will be used
to create resolved method without making a remote call.
If a relocatable version of resolved method is required, appropriate
validation record will be created as well.

4. However, _remoteMirror pointer of a newly created resolved method
might be invalid, because client resolved method only exists for the
duration of one compilation. It is set to NULL to indicate that
remote mirror does not exist.

5.
   a. If a method that requires _remoteMirror to exist is called,
we make a remote call to the client and recreate the mirror.
   b. However, there are a few methods that are called very frequently,
so making a remote call every time we need to recreate a mirror
nullifies most of the message savings of global caching.
For such methods, parameters needed for remote mirror creation are
explicitly added to the main message, so recreating a resolved method
will not incur any extra messages.

### DIFF
--- a/runtime/compiler/control/J9Options.cpp
+++ b/runtime/compiler/control/J9Options.cpp
@@ -2882,9 +2882,6 @@ J9::Options::unpackOptions(char *clientOptions, size_t clientOptionsSize, TR_Mem
    if (rtResolve)
       jitConfig->runtimeFlags |= J9JIT_RUNTIME_RESOLVE; // JITaaS FIXME: don't change global flags
 
-   // disable caching of resolved methods if env variable TR_DisableResolvedMethodsCaching is set
-   TR_ResolvedJ9JITaaSServerMethod::_useCaching = !feGetEnv("TR_DisableResolvedMethodsCaching");
-
    unpackRegex(options->_disabledOptTransformations);
    unpackRegex(options->_disabledInlineSites);
    unpackRegex(options->_disabledOpts);
--- a/runtime/compiler/control/JITaaSCompilationThread.cpp
+++ b/runtime/compiler/control/JITaaSCompilationThread.cpp
@@ -1029,7 +1029,7 @@ bool handleServerMessage(JITaaS::J9ClientStream *client, TR_J9VM *fe)
          break;
       case J9ServerMessageType::ResolvedMethod_staticAttributes:
          {
-         auto recv = client->getRecvData<TR_ResolvedJ9Method *, int32_t, bool, bool>();
+         auto recv = client->getRecvData<TR_ResolvedJ9Method *, int32_t, bool, bool, TR_ResolvedMethodParams>();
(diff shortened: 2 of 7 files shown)


dmitry-ten:
This change reduces the total number of messages by ~10%,
but CPU usage reduction is only ~2.5% percent.
Unclear why improvement is so small.
Maybe should cache unresolved methods locally, for the duration of one compilation.

dmitry-ten:
I re-added a local cache back, this seems to improve CPU consumption.
Now I'm registering 8-9% CPU improvement.
Will test how globally caching only methods of some well-known classes affects the number of messages.

dmitry-ten:
Marking it WIP while I'm comparing different caching strategies.

dmitry-ten:
I realized that there are more methods that require recreating resolved method mirror than I previously thought.
For example `isInterpreted` or `virtualMethodIsOverridden` do not require `_remoteMirror` pointer, but their return values might change in-between compilations, so we still need to know that a valid remote mirror exists.
With these methods recreating remote mirrors, the number of `mirrorResolvedJ9Method` messages rises dramatically, and most of the savings from global caching disappear.
This is very unfortunate, and probably indicates that implementing global caching is not worth it, especially considering how complicated it makes the code.
I can't think of a global caching strategy that would not require recreating remote mirrors. 

@mpirvu any thoughts on this?

mpirvu:
I agree that it may not be worthwhile to implement this work item because the savings are small compared to complexity (complexity leads to subtle bugs and maintenance problems in the future).

dmitry-ten:
Ok, closing this for now. Maybe we'll think of something later.

PR #5482, opened 2019-04-12
Respect GC's alwaysCallReadBarrier request on X86

0dvictor:
JIT's X86 CodeGen now always call GC's read barrier helpers when GC's
read barrier type is "gc_modron_readbar_always", which can be requested
via JVM option "-Xgc:alwaysCallReadBarrier".


cathyzhyi:
LGTM

PR #5488, opened 2019-04-12
Fix for compilation error on X86-32

0dvictor:
Fixed error caused by using deprecated version of 'generateLabelInstruction'.


pshipton:
jenkins compile win32 jdk8 depends eclipse/omr#3723

PR #5438, opened 2019-04-10
Remove uses of INTERP_COMPRESSED_OBJECT_HEADER

gacholio:
Replace with GC_COMPRESSED_POINTERS (J9VM/OMR/DDR versions). In every
build, either both or neither the flags are set. Removal of the flag
definitions will follow later.

[ci skip]

Signed-off-by: Graham Chapman <graham_chapman@ca.ibm.com>

PR #5491, opened 2019-04-13
Undo double define

gacholio:
Blind search/replace duplicated the compressed pointers flag in
jilconsts.

[ci skip]


gacholio:
Related: #5438

PR #5497, opened 2019-04-15
Suppress Artifactory cleanup failure if the build does not exist

samuel-rubin:
When a Test job cleanup it launched, and there is no matching Test folder
on the Artifactory server, a 404 is returned and the build fails.
This will add an error message to the description and log and keep the build green.
Reporting false errors will hide any real errors

[skip ci]

### DIFF
--- a/buildenv/jenkins/jobs/infrastructure/artifactory_cleanup.groovy
+++ b/buildenv/jenkins/jobs/infrastructure/artifactory_cleanup.groovy
@@ -78,12 +78,22 @@ def cleanupBuilds(artifactory_server, artifactory_repo, jobToCheck, artifactory_
         echo "Keeping the latest ${artifactory_max_num_artifacts} builds"
         currentBuild.description = "Keeping ${artifactory_max_num_artifacts} builds of ${jobToCheck}"
 
-        def request = httpRequest authentication: artifactoryCreds, consoleLogResponseBody: true, url: "${artifactory_server}/api/storage/${artifactory_repo}/${jobToCheck}"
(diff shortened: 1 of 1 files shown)


samuel-rubin:
@AdamBrousseau for review

samuel-rubin:
Here is what the job looks like if the build does not exist: https://ci.eclipse.org/openj9/view/Sandbox/job/sam-artifactory-cleanup/9/console

buildenv/jenkins/jobs/infrastructure/artifactory_cleanup.groovy
```diff
@@ -78,12 +78,22 @@ def cleanupBuilds(artifactory_server, artifactory_repo, jobToCheck, artifactory_
...
+            data = readJSON text: request.getContent()
+            numberOfArtifacts = data.children.size()
+            echo "There are ${numberOfArtifacts} builds"
+        } else {
+            currentBuild.description = "Warning: The URL does not exist<br>" + currentBuild.description
+            echo 'Warning: This URL does not exist... EXITING'
+            return
+        }
+    }
+    if (requestStatus == 404){
```
pshipton: What if other http errors occur, should htis be `!= 200`?

samuel-rubin: If a different http error occurs, the httpRequest at L81 would fail. The only http codes that the request will allow are 200 and 404. 
Also, this if statement was added because the return at L91 only ends the pipeline stage. This will end the pipeline job.

PR #5500, opened 2019-04-15
Get special-case invokeinterface receiver early

jdmpapin:
In 7295e2c8 the JIT compiler was modified so that it could generate IL for "improper" `invokeinterface` bytecode instructions, i.e. those targeting private methods or methods of `Object`. For these invocations, the dispatch does not naturally perform the type check required for `invokeinterface`, so an explicit type check is generated ahead of the call. This type check needs the receiver of the call, which until now has been read from the resulting call node. The problem is that it's possible that the node is no longer a call by the time the type check is generated, e.g. in the case of a call to `Object.getClass()`, the call becomes `aloadi <javaLangClassFromClass>` during `genInvokeDirect()`.

The receiver is now read from the stack before calling `genInvokeDirect()` or `genInvokeWithVFTChild()` so that the type check will be unaffected by any such transformations.

### DIFF
--- a/runtime/compiler/ilgen/Walker.cpp
+++ b/runtime/compiler/ilgen/Walker.cpp
@@ -3967,6 +3967,7 @@ TR_J9ByteCodeIlGenerator::genInvokeInterface(int32_t cpIndex)
       _methodSymbol->setMayHaveInlineableCall(true);
       TR::TreeTop *prevLastTree = _block->getExit()->getPrevTreeTop();
       TR::Node *callNode = NULL;
+      TR::Node *receiver = topn(improperMethod->numberOfExplicitParameters());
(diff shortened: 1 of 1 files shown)


DanHeidinga:
@andrewcraik @cathyzhyi Can one of you review this?

cathyzhyi:
LGTM

DanHeidinga:

DanHeidinga:
Jenkins test extended win jdk11

DanHeidinga:
Jenkins test sanity osx jdk12

jdmpapin:
Is this ok to merge now?

DanHeidinga:
@fjeremic Can you give this a second set of eyes and merge it?  @cathyzhyi has already reviewed and approved it.

PR #5502, opened 2019-04-15
Move Realtime BarrierSynchronization into Metronome delegate

jason-hall:


jason-hall:
@dmitripivkine What do you think about moving the functions in BarrierSynchronization.cpp to MetronomeDelegate? To me, it looks like they were all just wrapper functions for java internal VM functions as it is already. Let me know what you think

dmitripivkine:
> @dmitripivkine What do you think about moving the functions in BarrierSynchronization.cpp to MetronomeDelegate? To me, it looks like they were all just wrapper functions for java internal VM functions as it is already. Let me know what you think

I believe this is correct move

dmitripivkine:
Jenkins compile win

Issue #4486, opened 2019-01-29
AArch64: Compile error in trclog.c with strncpy()


knn-k:
I got the following error while I was compiling for linux_aarch64.  It is similar to Issue #3242.
```
aarch64-linux-gnu-gcc -U_FORTIFY_SOURCE -D_FORTIFY_SOURCE=1 -DOPENJ9_BUILD -O -fgnu89-inline -g -DLINUX -D_REENTRANT -D_FILE_OFFSET_BITS=64 -fpic -DIPv6_FUNCTION_SUPPORT -DJ9AARCH64 -Wimplicit -Wreturn-type -Werror -Wall -I. -I../include -I../oti -I../gc_include -I../omr/gc/include -I../nls -I../omr/include_core    -DUT_DIRECT_TRACE_REGISTRATION -DTR_HOST_ARM64 -c -o trclog.o trclog.c
In file included from /root/gcc-linaro-7.3.1-2018.05-x86_64_aarch64-linux-gnu/aarch64-linux-gnu/libc/usr/include/string.h:639:0,
                 from ../omr/include_core/omrstdarg.h:27,
                 from trclog.c:26:
In function 'strncpy',
    inlined from 'getTrcBuf' at trclog.c:1063:3:
/root/gcc-linaro-7.3.1-2018.05-x86_64_aarch64-linux-gnu/aarch64-linux-gnu/libc/usr/include/bits/string3.h:125:10: error: '__builtin_strncpy' writing 512 bytes into a region of size 1 overflows the destination [-Werror=stringop-overflow=]
   return __builtin___strncpy_chk (__dest, __src, __len, __bos (__dest));
          ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
cc1: all warnings being treated as errors
../makelib/targets.mk:272: recipe for target 'trclog.o' failed

# aarch64-linux-gnu-gcc --version
aarch64-linux-gnu-gcc (Linaro GCC 7.3-2018.05) 7.3.1 20180425 [linaro-7.3-2018.05 revision d29120a424ecfbc167ef90065c0eeb7f91977701]
```

It compiles without the error when I remove `-D_FORTIFY_SOURCE=1` from the command-line option.

gacholio:
I suspect a cast of the dest to `void*` in the caller would fix this.

knn-k:
No, casting the destination to `void *` does not change the error.
[...]

PR #5508, opened 2019-04-16
Disable "-D_FORTIFY_SOURCE=1" in runtime/rastrace for AArch64

knn-k:
This commit stops adding "-D_FORTIFY_SOURCE=1" to CFLAGS in
runtime/rastrace when the target platform is AArch64, in order to avoid
compile error with trclog.c.
With this change, "--disable-warnings-as-errors-openj9" can be removed
from the options for configure.

Fixes: #4486

### DIFF
--- a/doc/build-instructions/Build_Instructions_V11.md
+++ b/doc/build-instructions/Build_Instructions_V11.md
@@ -715,8 +715,7 @@ For building on AArch64 Linux:
 ```
 bash configure --with-freemarker-jar=/<path_to>/freemarker.jar \
                --with-boot-jdk=/<path_to_boot_JDK> \
-               --disable-warnings-as-errors \
-               --disable-warnings-as-errors-openj9
+               --disable-warnings-as-errors
 ```
 
 For building in the Docker container:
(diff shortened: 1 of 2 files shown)


knn-k:
It should be much better to disable checks only in the `runtime/rastrace` directory than to configure the build with `--disable-warnings-as-errors-openj9`,  for detecting potential code errors.

It would be even better if it is possible to disable checks using `_FORTIFY_SOURCE` with `trclog.c` only, not affecting other files in the same directory.

DanHeidinga:
@knn-k Is there something different about the compiler / os combo being used on aarch64?

knn-k:
@DanHeidinga Currently Linux+GCC is the only combination used for AArch64 right now.

DanHeidinga:
Sorry @knn-k for the unclear question.  I was wondering which version of Linux and which version of gcc are being used.  Most of our linux platforms are either Ubuntu 16 or centos 6 with gcc 7.3.  Does that match the os / gcc levels being used for aarch64?

knn-k:
The Dockerfile for AArch64 cross build (https://github.com/eclipse/openj9/blob/master/buildenv/docker/jdk11/aarch64_CC/arm-linux-aarch64/Dockerfile) uses the toolchain from Linaro containing GCC 7.3 and sysroot.

I need to ask @sxa555 on what kind of environment he uses for the AArch64 native build.

knn-k:
I changed the `module.xml` a little, following the idea in #4486.
The checks with `_FORTIFY_SOURCE` is disabled only for `trclog.c` in building for linux_aarch64.

DanHeidinga:
Jenkins compile xlinux jdk8

DanHeidinga:
Running a single test compile and will merge once it completes (likely tomorrow morning)

PR #5531, opened 2019-04-17
Add proper JNI / j9object_t error handling

DanHeidinga:

### DIFF
--- a/runtime/jcl/common/sun_reflect_ConstantPool.c
+++ b/runtime/jcl/common/sun_reflect_ConstantPool.c
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 1998, 2018 IBM Corp. and others
+ * Copyright (c) 1998, 2019 IBM Corp. and others
  *
  * This program and the accompanying materials are made available under
  * the terms of the Eclipse Public License 2.0 which accompanies this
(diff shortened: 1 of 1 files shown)


runtime/jcl/common/sun_reflect_ConstantPool.c
```diff
@@ -564,35 +564,62 @@ Java_sun_reflect_ConstantPool_getMemberRefInfoAt0(JNIEnv *env, jobject unusedObj
...
 	}
 
 	if ((NULL != classNameObject) && (NULL != nameObject) && (NULL != signatureObject)) {
 		jclass jlString = JCL_CACHE_GET(env, CLS_java_lang_String);
 		jobject array = (*env)->NewObjectArray(env, 3, jlString, NULL);
+		if (array == NULL) {
+			goto done;
+		}
 		(*env)->SetObjectArrayElement(env, array, 0, classNameObject);
+		if ((*env)->ExceptionCheck(env)) {
```
gacholio: This is a waste - the previous call cannot fail.

runtime/jcl/common/sun_reflect_ConstantPool.c
```diff
@@ -564,35 +564,62 @@ Java_sun_reflect_ConstantPool_getMemberRefInfoAt0(JNIEnv *env, jobject unusedObj
...
+			goto done;
+		}
 		(*env)->SetObjectArrayElement(env, array, 1, nameObject);
+		if ((*env)->ExceptionCheck(env)) {
```
gacholio: Same here.

runtime/jcl/common/sun_reflect_ConstantPool.c
```diff
@@ -564,35 +564,62 @@ Java_sun_reflect_ConstantPool_getMemberRefInfoAt0(JNIEnv *env, jobject unusedObj
...
+			goto done;
+		}
 		(*env)->SetObjectArrayElement(env, array, 2, signatureObject);
-		(*env)->DeleteLocalRef(env, classNameObject);
-		(*env)->DeleteLocalRef(env, nameObject);
-		(*env)->DeleteLocalRef(env, signatureObject);
-		if (!(*env)->ExceptionCheck(env)) {
-			returnValue = array;
+		if ((*env)->ExceptionCheck(env)) {
```
gacholio: And here.

runtime/jcl/common/sun_reflect_ConstantPool.c
DanHeidinga: Theory was that jni check would warn on this code without the exception checks

gacholio: I don't think it will (I suspect we have lots of other places that are not doing the check).

DanHeidinga: Turns out that jni check will warning for a pending exception when calling `SetObjectArrayElement` but it doesn't set the `jniCheckSetPotentialPendingException` flag

gacholio: My mistake - it can throw exceptions (though you have to make a programming error to trigger it).

gacholio:

PR #5538, opened 2019-04-17
Implement jmap

pdbain-ibm:
Get statistics on heap objects: for each class, number of objects and the
aggregate size.  Implement as a management bean.  Use Attach API to invoke the
bean method and return the result.

Add command to get the heap configuration.

Add standalone command for jmap: requires coordinated changes in the extensions
repositories.
https://github.com/ibmruntimes/openj9-openjdk-jdk8/pull/287
https://github.com/ibmruntimes/openj9-openjdk-jdk11/pull/160
https://github.com/ibmruntimes/openj9-openjdk-jdk12/pull/41
https://github.com/ibmruntimes/openj9-openjdk-jdk/pull/95

Add test cases.


jcl/src/jdk.management/share/classes/com/ibm/lang/management/MemoryMXBean.java
```diff
@@ -274,4 +276,10 @@
      * @return number of active GC worker threads
      */
 	public int getCurrentGCThreads();
+	
+	/**
+	 * Returns information on the number and size of objects in the heap.
+	 * @return Array of information objects organized by class
+	 */
+	public HeapClassInformation[] getHeapClassStatistics();
```
pshipton: Ideally this should be implemented to be compatible with sun.tools.attach.HotSpotVirtualMachine.heapHisto(Object... var1) so any app using this will work with OpenJ9.
The acceptable options are "-live" or "-all"

pdbain-ibm:
> The acceptable options are "-live" or "-all"

Could you clarify?  `jmap -help` says:

>     -heap                to print java heap summary
>     -histo[:live]        to print histogram of java object heap; if the "live"
>                          suboption is specified, only count live objects
>     -clstats             to print class loader statistics
>     -finalizerinfo       to print information on objects awaiting finalization

I don't see a `-all`.
I will add the finalizerinfo and clstats options.

pshipton:
For jmap it appers `all` is the default, i.e. if you don't specify `live`. For `heapHisto(Object... var1)` if nothing is specified then it returns only live objects.

pdbain-ibm:
> For jmap it appers `all` is the default, i.e. if you don't specify `live`. For `heapHisto(Object... var1)` if nothing is specified then it returns only live objects.

That was my understanding and is implemented thus.

pshipton:
Don't forget we also need a doc issue opened.

pdbain-ibm:
Per request from Peter Shipton, closing this in favour of a different implementation: create the formatted string in the target's heap walker native and return the string to the attacher.

SueChaplain:
@pdbain-ibm - I've not been following this too closely, but sounds like work might have moved into other PRs. Please can you confirm where you are at with this? We have no docs issue yet, but I don't want to lose sight of it. Thanks!

SueChaplain:
@pdbain-ibm @pshipton - it looks like this has gone in to 0.15.0? If so, plse create a docs issue with the info needed to write this up or advise where the work has moved to. Thanks

PR #5539, opened 2019-04-17
Reduce messages from server

satbir-ca:
1. getClassFromSignature: Make remote call to client
   and validate it. If it is validated then store class
   or store the NULL.
2. getClassFromMethodBlock: Save the class in the J9MethodMap.
   Check the cache before making remote call.

### DIFF
--- a/runtime/compiler/env/VMJ9Server.cpp
+++ b/runtime/compiler/env/VMJ9Server.cpp
@@ -255,9 +255,7 @@ TR_J9ServerVM::getClassLoader(TR_OpaqueClassBlock * classPointer)
 TR_OpaqueClassBlock *
 TR_J9ServerVM::getClassOfMethod(TR_OpaqueMethodBlock *method)
    {
-   JITaaS::J9ServerStream *stream = _compInfoPT->getMethodBeingCompiled()->_stream;
-   stream->write(JITaaS::J9ServerMessageType::VM_getClassOfMethod, method);
-   return std::get<0>(stream->read<TR_OpaqueClassBlock *>());
+   return TR_J9ServerVM::getClassFromMethodBlock(method);
    }
 
 TR_OpaqueClassBlock *
@@ -319,20 +317,12 @@ TR_J9ServerVM::getClassFromSignature(const char *sig, int32_t length, TR_Resolve
       {
       ClassLoaderStringPair key = {cl, std::string(sig, length)};
       auto it = _compInfoPT->getCustomClassByNameMap().find(key);
-
-      // If not found in the cache, ask the client
-      if (it == _compInfoPT->getCustomClassByNameMap().end())
+      if (it != _compInfoPT->getCustomClassByNameMap().end())
+         return it->second;
+      clazz = getClassFromSignature(sig, length, (TR_OpaqueMethodBlock *)method->getPersistentIdentifier(), isVettedForAOT);
+      if (clazz)
          {
-         clazz = getClassFromSignature(sig, length, (TR_OpaqueMethodBlock *)method->getPersistentIdentifier(), isVettedForAOT);
-         if (clazz)
-            {
-            // Put into the HT for next time
-            _compInfoPT->getCustomClassByNameMap()[key] = clazz;
-            }
-         }
-      else
-         {
-         clazz = it->second;
+         _compInfoPT->getCustomClassByNameMap()[key] = clazz;
          }
       }
    return clazz;
@@ -1184,6 +1174,15 @@ TR_J9ServerVM::isOwnableSyncClass(TR_OpaqueClassBlock *clazz)
 TR_OpaqueClassBlock *
 TR_J9ServerVM::getClassFromMethodBlock(TR_OpaqueMethodBlock *method)
    {
+      {
+      OMR::CriticalSection getRemoteROMClass(_compInfoPT->getClientData()->getROMMapMonitor());
+      auto it = _compInfoPT->getClientData()->getJ9MethodMap().find((J9Method*) method);
+      if (it != _compInfoPT->getClientData()->getJ9MethodMap().end())
+         {
+         return it->second._owningClass;
+         }
+      }
+
    JITaaS::J9ServerStream *stream = _compInfoPT->getMethodBeingCompiled()->_stream;
    stream->write(JITaaS::J9ServerMessageType::VM_getClassFromMethodBlock, method);
    return std::get<0>(stream->read<TR_OpaqueClassBlock *>());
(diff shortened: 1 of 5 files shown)


satbir-ca:
@mpirvu Please review the code.

runtime/compiler/env/VMJ9Server.cpp
```diff
@@ -1515,6 +1508,26 @@ TR_J9ServerVM::getROMMethodFromRAMMethod(J9Method *ramMethod)
    return std::get<0>(stream->read<J9ROMMethod *>());
    }
 
+TR_OpaqueClassBlock *
+TR_J9ServerVM::getCacheClassFromSignature(const char * sig, int32_t sigLength, bool isVettedForAOT, TR_OpaqueMethodBlock *methodBlock, J9ClassLoader * cl, bool cacheInvalidatedClass)
+   {
+   ClassLoaderStringPair key = {cl, std::string(sig, sigLength)};
+      {
+         OMR::CriticalSection getCustomClassMap(_compInfoPT->getCustomClassMapMonitor());
```
mpirvu: The code is accessing the 'per compilation' cache which is only accessed by one compilation thread, therefore the monitor is not needed.

runtime/compiler/env/VMJ9Server.cpp
```diff
@@ -1515,6 +1508,26 @@ TR_J9ServerVM::getROMMethodFromRAMMethod(J9Method *ramMethod)
    return std::get<0>(stream->read<J9ROMMethod *>());
    }
 
+TR_OpaqueClassBlock *
+TR_J9ServerVM::getCacheClassFromSignature(const char * sig, int32_t sigLength, bool isVettedForAOT, TR_OpaqueMethodBlock *methodBlock, J9ClassLoader * cl, bool cacheInvalidatedClass)
```
mpirvu: The name of the method should better reflect what the method does. There is another cache for classes, so there could be some confusion.

runtime/compiler/env/VMJ9Server.cpp
```diff
@@ -1704,10 +1719,11 @@ TR_J9SharedCacheServerVM::getClassFromSignature(const char * sig, int32_t sigLen
          }
       }
 
-   if (validated)
-      return j9class;
-   else
-      return NULL;
+   //if Validation fails still save NULL in the cache as it will remain invalidated in the future also
```
mpirvu: The comment seems out of place here because we don't do any caching in the next few lines

runtime/compiler/env/VMJ9Server.cpp
```diff
@@ -1704,10 +1719,11 @@ TR_J9SharedCacheServerVM::getClassFromSignature(const char * sig, int32_t sigLen
...
+   if (!validated)
+      j9class = nullptr;
```
mpirvu: Prefer NULL to nullptr due to compatibility.

runtime/compiler/env/VMJ9Server.cpp
```diff
@@ -1676,13 +1689,15 @@ TR_OpaqueClassBlock *
 TR_J9SharedCacheServerVM::getClassFromSignature(const char * sig, int32_t sigLength, TR_ResolvedMethod * method, bool isVettedForAOT)
    {
    TR_ResolvedRelocatableJ9JITaaSServerMethod* resolvedJITaaSMethod = (TR_ResolvedRelocatableJ9JITaaSServerMethod *)method;
-   return getClassFromSignature(sig, sigLength, (TR_OpaqueMethodBlock *)resolvedJITaaSMethod->getPersistentIdentifier(), isVettedForAOT);
+   J9ClassLoader * cl = resolvedJITaaSMethod->getClassLoader();
```
mpirvu: Why is the change needed here. The newly created getCacheClassFromSignature() only use the 'per-compilation' cache.

runtime/compiler/env/VMJ9Server.cpp
mpirvu: Note that the caching has already happened in `TR_J9ServerVM::getClassFromSignature(sig, sigLength, method, true);`
If you intended to store NULL when the validation didn't succeed, it's not happening.

runtime/compiler/env/VMJ9Server.cpp
```diff
@@ -1676,38 +1677,94 @@ TR_OpaqueClassBlock *
...
+   // then look into the global, 'per client', cache
+   void * systemClassLoader = getSystemClassLoader();
+   if (cl == (J9ClassLoader*)systemClassLoader)
+      {
+      PersistentUnorderedMap<std::string, TR_OpaqueClassBlock*> & systemClassByNameMap = _compInfoPT->getClientData()->getSystemClassByNameMap();
+         {
+         OMR::CriticalSection classFromSigCS(_compInfoPT->getClientData()->getSystemClassMapMonitor());
+         auto it = systemClassByNameMap.find(std::string(sig, sigLength));
+         if (it != systemClassByNameMap.end())
+            return it->second;
```
mpirvu: I am not 100% sure, but I think that we need to validate whatever comes from the cache. The reason is that this cache is global for AOT and non-AOT queries, so it could have been a non-AOT query that populated the cache and the validation was never done. Maybe ask Irwin.

runtime/compiler/env/VMJ9Server.cpp
```diff
@@ -1676,38 +1677,94 @@ TR_OpaqueClassBlock *
...
+         }
+      // classname not found; ask the client and cache the answer  
+      clazz = TR_J9ServerVM::getClassFromSignature(sig, sigLength, (TR_OpaqueMethodBlock *)resolvedJITaaSMethod->getPersistentIdentifier(), isVettedForAOT);
+      if (clazz)
+         {
+         if (!validateClass((TR_OpaqueMethodBlock *)resolvedJITaaSMethod->getPersistentIdentifier(), clazz, isVettedForAOT))
+            clazz = NULL;
+         OMR::CriticalSection classFromSigCS(_compInfoPT->getClientData()->getSystemClassMapMonitor());
```
mpirvu: We should not store NULL in this case, because in the future a non-AOT query will return NULL even though it doesn't have to do a validation.

runtime/compiler/env/VMJ9Server.cpp
```diff
@@ -1676,38 +1677,94 @@ TR_OpaqueClassBlock *
...
+         }
+      }
+   else // look into the 'per compilation' cache
+      {
+      J9ClassLoader * cl = resolvedJITaaSMethod->getClassLoader();
+   
+      ClassLoaderStringPair key = {cl, std::string(sig, sigLength)};
+      auto it = _compInfoPT->getCustomClassByNameMap().find(key);
+      if (it != _compInfoPT->getCustomClassByNameMap().end())
+         return it->second;
```
mpirvu: Same as above, validation is still needed.

runtime/compiler/env/VMJ9Server.cpp
```diff
@@ -1676,38 +1677,94 @@ TR_OpaqueClassBlock *
...
+      clazz = TR_J9ServerVM::getClassFromSignature(sig, sigLength, (TR_OpaqueMethodBlock *)resolvedJITaaSMethod->getPersistentIdentifier(), isVettedForAOT);
+      if (clazz)
+         {
+         if (!validateClass((TR_OpaqueMethodBlock *)resolvedJITaaSMethod->getPersistentIdentifier(), clazz, isVettedForAOT))
```
mpirvu: Same as above, do not store NULL based on the answer form the validation.

runtime/compiler/env/VMJ9Server.cpp
```diff
@@ -1676,38 +1675,114 @@ TR_OpaqueClassBlock *
...
+         auto it = systemClassByNameMap.find(std::string(sig, sigLength));
+         if (it != systemClassByNameMap.end())
+            clazz = it->second;
+         }
+      if (!clazz)
+         {
+         // classname not found; ask the client and cache the answer
+         clazz = TR_J9ServerVM::getClassFromSignature(sig, sigLength, (TR_OpaqueMethodBlock *)resolvedJITaaSMethod->getPersistentIdentifier(), isVettedForAOT);
+         if (clazz)
+            {
```
mpirvu: If the client sent us back a class we must store it in the cache irrespective of the validation result. This is to catch non-AOT requests that do not need a validation.

PR #5540, opened 2019-04-17
Reducing SharedCache_rememberClass messages

satbir-ca:
Reducing the message SharedCache_rememberClass from server to client
by caching the class chain data.

### DIFF
--- a/runtime/compiler/control/JITaaSCompilationThread.cpp
+++ b/runtime/compiler/control/JITaaSCompilationThread.cpp
@@ -2845,6 +2845,7 @@ ClientSessionData::ClientSessionData(uint64_t clientUID, uint32_t seqNo) :
    _romClassMap(decltype(_romClassMap)::allocator_type(TR::Compiler->persistentAllocator())),
    _J9MethodMap(decltype(_J9MethodMap)::allocator_type(TR::Compiler->persistentAllocator())),
    _systemClassByNameMap(decltype(_systemClassByNameMap)::allocator_type(TR::Compiler->persistentAllocator())),
+   _classChainDataMap(decltype(_classChainDataMap)::allocator_type(TR::Compiler->persistentAllocator())),
    _unloadedClassAddresses(NULL),
    _requestUnloadedClasses(true),
    _staticFinalDataMap(decltype(_staticFinalDataMap)::allocator_type(TR::Compiler->persistentAllocator()))
@@ -2855,6 +2856,7 @@ ClientSessionData::ClientSessionData(uint64_t clientUID, uint32_t seqNo) :
    _numActiveThreads = 0;
    _romMapMonitor = TR::Monitor::create("JIT-JITaaSROMMapMonitor");
    _systemClassMapMonitor = TR::Monitor::create("JIT-JITaaSSystemClassMapMonitor");
+   _classChainDataMapMonitor = TR::Monitor::create("JIT-JITaaSClassChainDataMapMonitor");
    _sequencingMonitor = TR::Monitor::create("JIT-JITaaSSequencingMonitor");
    _vmInfo = NULL;
    _staticMapMonitor = TR::Monitor::create("JIT-JITaaSStaticMapMonitor");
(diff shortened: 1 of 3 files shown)


satbir-ca:
@mpirvu Please review the code.

runtime/compiler/control/JITaaSCompilationThread.hpp
```diff
@@ -160,9 +162,10 @@ class ClientSessionData
    // The following hashtable caches <classname> --> <J9Class> mappings
    // All classes in here are loaded by the systemClassLoader so we know they cannot be unloaded
    PersistentUnorderedMap<std::string, TR_OpaqueClassBlock*> _systemClassByNameMap;
- 
+   PersistentUnorderedMap<J9Class *, UDATA *> _classChainDataCacheMap;
```
mpirvu: Name could be better here: delete either `Cache` or `Map` because they mean more or less the same thing.

runtime/compiler/control/JITaaSCompilationThread.cpp
```diff
@@ -2785,6 +2785,7 @@ ClientSessionData::ClientSessionData(uint64_t clientUID, uint32_t seqNo) :
    _romClassMap(decltype(_romClassMap)::allocator_type(TR::Compiler->persistentAllocator())),
    _J9MethodMap(decltype(_J9MethodMap)::allocator_type(TR::Compiler->persistentAllocator())),
    _systemClassByNameMap(decltype(_systemClassByNameMap)::allocator_type(TR::Compiler->persistentAllocator())),
+   _classChainDataCacheMap(decltype(_classChainDataCacheMap)::allocator_type(TR::Compiler->persistentAllocator())),
```
mpirvu: Alignment seems off

runtime/compiler/env/J9SharedCache.cpp
```diff
@@ -810,10 +810,21 @@ UDATA *
 TR_J9JITaaSServerSharedCache::rememberClass(J9Class *clazz, bool create)
    {
    TR_ASSERT(_stream, "stream must be initialized by now");
-   // TODO: might want to keep a per-client cache of (class -> class chain)
-   // that were remembered to reduce the number of remote calls
+   PersistentUnorderedMap<J9Class *, UDATA *> & cache = TR::compInfoPT->getClientData()->getClassClainDataCache();
```
mpirvu: Avoid using TR::compInfoPT repeatedly. This is using thread local storage which is expensive to access, so try to use as little as possible if at all. Here we can just remember the clientDta

runtime/compiler/env/J9SharedCache.cpp
```diff
@@ -810,10 +810,21 @@ UDATA *
...
+      if (it != cache.end())
+         return it->second;
+      }
    _stream->write(JITaaS::J9ServerMessageType::SharedCache_rememberClass, clazz, create);
-   return std::get<0>(_stream->read<UDATA *>());
+   UDATA * chainData = std::get<0>(_stream->read<UDATA *>());
+   if (chainData)
+      {
+      OMR::CriticalSection classChainDataMapMonitor(TR::compInfoPT->getClientData()->getClassChainDataMapMonitor());
+      cache.insert(std::make_pair(clazz, chainData));
```
mpirvu: Looking at the implementation, if create is false then rememberClass() returns 1. It would be wrong to cache that value.

Also, the original rememberClass() has some diagnostic messages. Maybe we should add something similar to this implementation.

mpirvu:
Another point is that classes can be unloaded so the new cache needs to be purged like the other one that stores classes.

runtime/compiler/env/J9SharedCache.cpp
```diff
@@ -810,10 +810,31 @@ UDATA *
...
    _stream->write(JITaaS::J9ServerMessageType::SharedCache_rememberClass, clazz, create);
-   return std::get<0>(_stream->read<UDATA *>());
+   UDATA * chainData = std::get<0>(_stream->read<UDATA *>());
+   if (chainData)
+      {
+      if (!create)
+         {
+         if (TR::Options::getVerboseOption(TR_VerboseJITaaS))
+            TR_VerboseLog::writeLineLocked(TR_Vlog_JITaaS, "not asked to create but could create, returning non-null \n");
+         chainData = (UDATA *) 0x1;
```
mpirvu: Delete this line. Whatever the client tells us is the right answer. If the base code changes 0x1 to 0x2 we don't need to change this piece of code as well.

satbir-ca:
updated

PR #5548, opened 2019-04-17
Check client SCC before sending SharedCache_addHint

### LINKED ISSUE #5546: SharedCache_addHint message should only be sent when JITaaS client uses SCC
While running AcmeAir with -Xshareclasses:none I observed a few `SharedCache_addHint` messages which should be sent at all.

a7ehuo:
[skip ci]

Fix #5546


a7ehuo:
@mpirvu please help review. thanks.

PR #5549, opened 2019-04-17
Handle multiple threads in OSX corereader

mikezhang1234567890:
Fixes https://github.com/eclipse/openj9/issues/5230

Signed-off-by: Mike Zhang <mike.h.zhang@ibm.com>

Reviewer: @keithc-ca

### DIFF
--- a/debugtools/DDR_VM/src/com/ibm/j9ddr/corereaders/macho/MachoDumpReader.java
+++ b/debugtools/DDR_VM/src/com/ibm/j9ddr/corereaders/macho/MachoDumpReader.java
@@ -51,6 +51,7 @@ import java.util.LinkedList;
 import java.util.List;
 import java.util.Map;
 import java.util.Properties;
+import java.util.TreeMap;
 
 import javax.imageio.stream.FileImageInputStream;
 import javax.imageio.stream.ImageInputStream;
@@ -176,7 +177,7 @@ public class MachoDumpReader extends AbstractCoreReader implements ILibraryDepen
 		public List<LoadCommand> loadCommands;
 		public List<SegmentCommand64> segments;
 		public List<LoadCommand> otherLoads;
-		public ThreadCommand threads;
+		public List<ThreadCommand> threads;
 		long streamOffset;
 
 		public MachFile64() {}
@@ -208,10 +209,16 @@ public class MachoDumpReader extends AbstractCoreReader implements ILibraryDepen
 		private final List<IMemoryRange> memoryRanges = new LinkedList<>();
 		private final List<IOSStackFrame> stackFrames = new LinkedList<>();
 
-		public OSXThread(long tid, ThreadCommand.ThreadState thread)
+		public OSXThread(long tid, ThreadCommand thread)
 		{
 			threadId = tid;
-			registers = thread.registers;
+			registers = new TreeMap<>();
+			// Since IOSThread doesn't have a sense of the different types of registers for a thread,
+			// we add all the registers to the OSXThread. The registers all have different names, so 
+			// there should be no conflict.
+			for (ThreadState state : thread.states) {
+				registers.putAll(state.registers);
+			}
 			properties = new Properties();
 		}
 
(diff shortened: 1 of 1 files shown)


debugtools/DDR_VM/src/com/ibm/j9ddr/corereaders/macho/MachoDumpReader.java
```diff
@@ -208,10 +209,13 @@ public MachFile64() {}
...
 		private final List<IOSStackFrame> stackFrames = new LinkedList<>();
 
-		public OSXThread(long tid, ThreadCommand.ThreadState thread)
+		public OSXThread(long tid, ThreadCommand thread)
 		{
 			threadId = tid;
-			registers = thread.registers;
+			registers = new TreeMap<>();
+			for (ThreadState state : thread.states) {
+				registers.putAll(state.registers);
```
keithc-ca: Why do we want to merge registers from multiple `ThreadState` objects?

debugtools/DDR_VM/src/com/ibm/j9ddr/corereaders/macho/MachoDumpReader.java
```diff
@@ -357,18 +361,15 @@ public long getProcessId()
...
-				x86ThreadStates.add(state);
-			} else if (_signalNumber == -1 && state.flavor == ThreadCommand.x86_EXCEPTION_STATE) {
-				_signalNumber = state.registers.get("err").intValue();
+		List<IOSThread> threads = new ArrayList<>(dumpFile.threads.size());
+		for (ThreadCommand command : dumpFile.threads) {
+				OSXThread nativeThread = new OSXThread(0, command);
+				threads.add(nativeThread);
+				if (_signalNumber == -1) {
+					// will use the first thread's exception values, which should be the primary thread
+					_signalNumber = nativeThread.registers.get("err").intValue();
```
keithc-ca: What if `nativeThread.registers` has no `err` key?

debugtools/DDR_VM/src/com/ibm/j9ddr/corereaders/macho/MachoDumpReader.java
```diff
@@ -440,13 +436,17 @@ public MachFile64 readMachFile(long fileOffset) throws IOException, InvalidDumpF
...
 		}
 		long currentOffset = fileOffset + header64Size;
 		for (int i = 0; i < machfile.header.numCommands; i++) {
 			LoadCommand command = LoadCommand.readFullCommand(_fileReader, currentOffset, fileOffset);
 			if (command instanceof SegmentCommand64) {
 				SegmentCommand64 segment = (SegmentCommand64) command;
 				machfile.segments.add(segment);
+			} else if (command instanceof ThreadCommand) {
+				machfile.threads.add((ThreadCommand)command);
+				machfile.otherLoads.add(command);
```
keithc-ca: Why does this add the `ThreadCommand` to `otherLoads `?

debugtools/DDR_VM/src/com/ibm/j9ddr/corereaders/macho/MachoDumpReader.java
mikezhang1234567890: They all represent different sets of registers (x86 thread registers, x87 fpu, exception registers, etc). We only handle the x86 thread registers and the exception registers in the core reader, and the registers all have different names so they can be merged. The `IOSThread` interface doesn't have a notion of different sets of registers for a thread so we just merge the different sets.

debugtools/DDR_VM/src/com/ibm/j9ddr/corereaders/macho/MachoDumpReader.java
mikezhang1234567890: I believe every Mac OSX machine has a `x86_EXCEPTION_STATE` so that key will always exist. It should produce an exception otherwise.

debugtools/DDR_VM/src/com/ibm/j9ddr/corereaders/macho/MachoDumpReader.java
keithc-ca: Thanks, @mikezhang1234567890. A comment to that effect would be helpful in the code.

debugtools/DDR_VM/src/com/ibm/j9ddr/corereaders/macho/MachoDumpReader.java
keithc-ca: We should do better than `NullPointerException`, though.

pshipton:
What is the next step here, is this due for another review?

debugtools/DDR_VM/src/com/ibm/j9ddr/corereaders/macho/MachoDumpReader.java
```diff
@@ -355,27 +362,32 @@ public long getProcessId()
...
-				_signalNumber = state.registers.get("err").intValue();
+		List<IOSThread> threads = new ArrayList<>(dumpFile.threads.size());
+		for (ThreadCommand command : dumpFile.threads) {
+				OSXThread nativeThread = new OSXThread(0, command);
+				threads.add(nativeThread);
+				if (_signalNumber == -1) {
+					// will use the first thread's exception values, which should be the primary thread
+					try {
+						_signalNumber = nativeThread.registers.get("err").intValue();
+					} catch (NullPointerException e) {
```
keithc-ca: Catching `NullPointerException` is almost never the best way to deal with `null`. I expect this only needs to put `nativeThread.registers.get("err")` in a local and check that it's not null.

mikezhang1234567890: @keithc-ca changed the code to actually check if the key exists

keithc-ca:
Jenkins test extended osx,plinux jdk11

PR #5551, opened 2019-04-18
Optimized JNI linkage on X86-64

0dvictor:
Reduced code path and potential register usage.

### DIFF
--- a/runtime/compiler/x/amd64/codegen/AMD64JNILinkage.cpp
+++ b/runtime/compiler/x/amd64/codegen/AMD64JNILinkage.cpp
@@ -521,40 +521,24 @@ void TR::AMD64JNILinkage::buildJNICallOutFrame(
    //    1:   frame flags
    //    0:   RAM method
    //
-   TR::X86RegImmInstruction *instr = generateRegImmInstruction(
-      SUBRegImms(),
-      callNode,
-      espReal,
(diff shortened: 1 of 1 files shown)


runtime/compiler/x/amd64/codegen/AMD64JNILinkage.cpp
```diff
@@ -521,40 +521,24 @@ void TR::AMD64JNILinkage::buildJNICallOutFrame(
...
-         scratchReg = cg()->allocateRegister();
-
-      generateRegImm64Instruction(MOV8RegImm64, callNode, scratchReg, tagBits, cg());
-      generateMemRegInstruction(S8MemReg, callNode, tempMR, scratchReg, cg());
-      }
+   generateImmInstruction(PUSHImm4, callNode, tagBits, cg());
 
    // (skip savedPC slot).
    //
+   generateImmInstruction(PUSHImm4, callNode, 0, cg());
```
cathyzhyi: what is the `push 0` for?

0dvictor: The linkage requires a slot for `savedPC`, but its content is ignored; therefore, I put this push instruction with an arbitrary value 0.

cathyzhyi:
Placing stores with faster and shorter pushes. LGTM.
@vijaysun-omr could you merge this one as well?

andrewcraik:

PR #5561, opened 2019-04-18
Replace nullptr with NULL

dmitry-ten:
Since mainline doesn't support nullptr,
replace all of its occurences in JITaaS code with NULL.


dmitry-ten:
@mpirvu could you review?

dmitry-ten:
Fixed copyright.

PR #5564, opened 2019-04-18
Eliminate IProfiler_persistIprofileInfo messages

### LINKED ISSUE #5486: Large number of IProfiler_persistIprofileInfo messages
Doing some JITaaS runs with a warmed shared class cache the most frequent message sent by the server to the client is by far IProfiler_persistIprofileInfo.
Looking at the code I see that for each AOT load we try to store IProfiler info not only for the method we are loading, but also every inlined method. However, the server should not do any AOT loads (they should be performed at the client).
The only place where the server could issue IProfiler_persistIprofileInfo messages is IlGen, but that is only once per compiled AOT body. The number of IProfiler_persistIprofileInfo messages far exceeds the number of AOT compilations.

a7ehuo:
[skip ci]
Each inlined method could trigger an `IProfiler_persistIprofileInfo`
update. For each compilation, it would result in more than one
`IProfiler_persistIprofileInfo` messages being sent from the server.
The optimization is to save `IProfiler_persistIprofileInfo` request
during the compilation and send a list of the methods at the end
of the compilation along with `finishCompilation()` message to the client.

Fix #5486

### DIFF
--- a/runtime/compiler/control/JITaaSCompilationThread.hpp
+++ b/runtime/compiler/control/JITaaSCompilationThread.hpp
@@ -36,6 +36,7 @@ struct TR_RemoteROMStringKey;
 using IPTable_t = PersistentUnorderedMap<uint32_t, TR_IPBytecodeHashTableEntry*>;
 using IPTableHeapEntry = UnorderedMap<uint32_t, TR_IPBytecodeHashTableEntry*>;
 using IPTableHeap_t = UnorderedMap<J9Method *, IPTableHeapEntry *>;
+using ResolvedMirrorMethodsPersistIP_t = Vector<TR_ResolvedJ9Method *>;
 
 class ClientSessionData
    {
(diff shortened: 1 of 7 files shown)


a7ehuo:
@mpirvu Ready for review. Thanks.

a7ehuo:
@mpirvu Updated. Please have a look. Thanks.

runtime/compiler/control/JITaaSCompilationThread.hpp
```diff
@@ -36,6 +36,7 @@ struct TR_RemoteROMStringKey;
 using IPTable_t = PersistentUnorderedMap<uint32_t, TR_IPBytecodeHashTableEntry*>;
 using IPTableHeapEntry = UnorderedMap<uint32_t, TR_IPBytecodeHashTableEntry*>;
 using IPTableHeap_t = UnorderedMap<J9Method *, IPTableHeapEntry *>;
+using resolvedMirrorMethodsPersistIP_t = Vector<TR_ResolvedJ9Method *>;
```
mpirvu: Typically data structures start with a capital letter so let's make that `ResolvedMirrorMethodsPersistIP_t`

mpirvu:
Also, please resolve conflicts.

a7ehuo:
@mpirvu Updated. Please have a look. Thanks.

PR #5567, opened 2019-04-19
AArch64: Implement initARM64RealRegisterLinkage() in ARM64PrivateLinkage

knn-k:
This commit implements
TR::ARM64PrivateLinkage::initARM64RealRegisterLinkage().

### DIFF
--- a/runtime/compiler/aarch64/codegen/ARM64PrivateLinkage.cpp
+++ b/runtime/compiler/aarch64/codegen/ARM64PrivateLinkage.cpp
@@ -47,9 +47,44 @@ void TR::ARM64PrivateLinkage::mapSingleAutomatic(TR::AutomaticSymbol *p, uint32_
    TR_UNIMPLEMENTED();
    }
 
+static void lockRegister(TR::RealRegister *regToAssign)
+   {
+   regToAssign->setState(TR::RealRegister::Locked);
+   regToAssign->setAssignedRegister(regToAssign);
+   }
+
 void TR::ARM64PrivateLinkage::initARM64RealRegisterLinkage()
    {
-   TR_UNIMPLEMENTED();
+   TR::Machine *machine = cg()->machine();
+   TR::RealRegister *reg;
+   int icount;
+
+   reg = machine->getRealRegister(TR::RealRegister::RegNum::x16); // IP0
+   lockRegister(reg);
+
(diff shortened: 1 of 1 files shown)


knn-k:
I used the same weights for the registers as those used in `ARM64SystemLinkage`.  It needs to be revisited.

runtime/compiler/aarch64/codegen/ARM64PrivateLinkage.cpp
```diff
@@ -47,9 +47,47 @@ void TR::ARM64PrivateLinkage::mapSingleAutomatic(TR::AutomaticSymbol *p, uint32_
...
+   lockRegister(reg);
+
+   reg = machine->getRealRegister(TR::RealRegister::RegNum::x30); // LR
+   lockRegister(reg);
+
+   reg = machine->getRealRegister(TR::RealRegister::RegNum::sp); // SP
+   lockRegister(reg);
+
+   reg = machine->getRealRegister(TR::RealRegister::RegNum::xzr); // zero
+   lockRegister(reg);
```
0xdaryl: There are situations where `xzr` is used in a register dependency.  For example,

https://github.com/eclipse/omr/blob/02c82646ad68a02067d6d3a349f070cb40b2b537/compiler/aarch64/codegen/GenerateInstructions.cpp#L314

If the register is locked, will the register assigner be able to assign `xzr` to the virtual?

If you're trying to exclude `xzr` from assignment as a general GPR, then I'd say the logic that iterates through the available GPRs looking for register candidates needs to exclude `xzr`.  Either that, or teach the RA to be able to assign a real register directly (i.e, `xzr`) without the use of a virtual.  I think we had that discussion before in the context of that code in OMR.

knn-k: ARM64SystemLinkage in OMR also locks `xzr`.
https://github.com/eclipse/omr/blob/6400500b041af0e3b578c3cc984fd0404ef5b881/compiler/aarch64/codegen/ARM64SystemLinkage.cpp#L156
I need to recall what we discussed before.

knn-k:
Updated the code by removing the lines for locking `xzr`.

0xdaryl:
A Jenkins build is pointless as this is an AArch64-specific change.  Merging...

PR #5570, opened 2019-04-19
WIP: Reduce number of VM_getResolvedVirtualMethod messages

### LINKED ISSUE #5559: Reduce number of ResolvedMethod_getResolvedVirtualMethod messages
In a JITaaS setting `ResolvedMethod_getResolvedVirtualMethod` is by far the most frequent message for hot compilations (~ 450 such messages per compilation).
The implementation takes a class and a virtualCallOffset and return a j9method.
I think we can try to cache it globally. For each j9class we can keep another hash where the key is the virtualCallOffset. We have to be careful at the parameter `ignoreRtResolve`

mpirvu:
To reduce the frequency of VM_getResolvedVirtualMethod
messages this commit adds a per-session cache that will store
<virtualCallSelector --> j9method mappings>. There will be one cache
for each j9class, but these caches are created on demand.

[skip ci]

### DIFF
--- a/runtime/compiler/control/JITaaSCompilationThread.cpp
+++ b/runtime/compiler/control/JITaaSCompilationThread.cpp
@@ -3090,6 +3090,11 @@ ClientSessionData::ClassInfo::freeClassInfo()
       _staticAttributesCacheAOT->~TR_FieldAttributesCache();
       jitPersistentFree(_staticAttributesCacheAOT);
       }
+   if (_virtualMethodCache)
+      {
+      _virtualMethodCache->~PersistentUnorderedMap<int32_t, TR_OpaqueMethodBlock *>();
+      jitPersistentFree(_virtualMethodCache);
+      }
    }
 
 ClientSessionData::VMInfo *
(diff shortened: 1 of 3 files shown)


mpirvu:
Unfortunately, I cached `VM_getResolvedVirtualMethod` message instead of `ResolvedMethod_getResolvedVirtualMethod`

dmitry-ten:
This attempts to do the same thing as #5266, where I added `ResolvedMethod_getResolvedVirtualMethod` message.
With my approach you don't need a second remote message to create a resolved method.
If my change was reverted, and this one applied, we would probably have higher hit rates for RAM methods, but it would send a remote message to create a resolved method every time `TR_ResolvedJ9JITaaSServerMethod::getResolvedVirtualMethod` is called.

mpirvu:
> If my change was reverted, and this one applied, we would probably have higher hit rates for RAM methods, but it would send a remote message to create a resolved method every time TR_ResolvedJ9JITaaSServerMethod::getResolvedVirtualMethod is called.

I do not plan on reverting #5266

genie-openj9:

Issue #5565, opened 2019-04-19
JVMTI agent is unable to get Java version in native code on OpenJ9 JDK 12


AntonKatilin:
We use JVMTI's GetSystemProperty function to get Java version as a number like 8, 9, 10, 11, 12 etc. in native code, in an early phase when no Java code has loaded yet. Since OpenJ9 Java 12 this approach no longer works.

Property "java.vm.version" on HotSpot returns something like "12.0.1+12" which we can parse to get the version. OpenJ9 returns "openj9-0.14.0" which is useless for our purpose because it indicates the entire release with different Java versions like Java 11, Java 12 etc.

Property "java.vm.specification.version" returns "11" on OpenJ9 Java 11, and we can parse it. Unfortunately, OpenJ9 Java 12 returns an empty string.

Neither can we use JVMTI's GetVersionNumber function because for Java 11 and Java 12  it returns the same value JVMTI_VERSION_11 = 300B0000.

Could you please make "java.vm.specification.version" return a meaningful value ("12"), or suggest a workaround.

DanHeidinga:
This was changed in https://github.com/eclipse/openj9/commit/7e3036dc1b7203aaabe07305516766596a73d591

Likely due to VersionProps.init not setting `java.vm.specification.version`.  

Looks like we need to reorganize the property setting code a bit

JasonFengJ9:
Will have a look this.

AntonKatilin:
This bug was in JDK 12, and is labelled as fixed in 0.14.2. However, it seems there is no JDK 12 build 0.14.2 or newer available, there is only 0.14.2 for JDK 8 and 0.14.3 for JDK 11. Should we wait for upcoming 0.14.x release for JDK 12 with this bug fixed, or will its fix in a release (not in a nightly build) be actually postponed until 0.15?

pshipton:
@AntonKatilin while technically the problem is fixed in the OpenJ9 0.14.2 branch (and also for 0.15), we decided against building a Java 12 release for 0.14.x at AdoptOpenJDK, since it is a short term support release, and to save on people resources. We're not planning any new release build at Adopt for Java 12 before 0.15. The nightly builds are available with the fix.

One option if you want to discuss it, there is an OpenJ9 community call at 11am EDT every Wed, details in the [#planning](https://openj9.slack.com/messages/C8RBG6N2W) channel on the OpenJ9 Slack.

PR #5576, opened 2019-04-22
Init system property java.vm.specification.version via native

JasonFengJ9:
Init system property `java.vm.specification.version` via native

There is a need to retrieve the system property `java.vm.specification.version` via `JVMTI API` in an early phase before Java code `java.lang.VersionProps.init(systemProperties)` running.
Removed the Java code (Java 12+) setting the property;
Get the property value via `JAVA_SPEC_VERSION` and set it within `vmprops.c:initializeSystemProperties()`;

Verified that the system property `java.vm.specification.version` has correct value.

closes: #5565 

Reviewer: @DanHeidinga 

### DIFF
--- a/runtime/include/vendor_version.h
+++ b/runtime/include/vendor_version.h
@@ -51,6 +51,7 @@
 #define JAVA_VM_VENDOR "Eclipse OpenJ9"
 #define JAVA_VM_NAME "Eclipse OpenJ9 VM"
 
+#if JAVA_SPEC_VERSION < 12
 /* Pre-JDK12 versions use following defines to set system properties
  * java.vendor and java.vendor.url within vmprop.c:initializeSystemProperties(vm).
  * JDK12 (assuming future versions as well) sets these properties via java.lang.VersionProps.init(systemProperties) 
@@ -58,5 +59,6 @@
  */
 #define JAVA_VENDOR "Eclipse OpenJ9"
 #define JAVA_VENDOR_URL "http://www.eclipse.org/openj9"
+#endif /* JAVA_SPEC_VERSION < 12 */
 
 #endif     /* vendor_version_h */
--- a/runtime/include/vendor_version.h
+++ b/runtime/include/vendor_version.h
@@ -49,8 +49,8 @@
 #define VENDOR_SHORT_NAME "OpenJ9"
 
 #define JAVA_VM_VENDOR "Eclipse OpenJ9"
+#define JAVA_VM_NAME "Eclipse OpenJ9 VM"
 
-#if JAVA_SPEC_VERSION < 12
 /* Pre-JDK12 versions use following defines to set system properties
  * java.vendor and java.vendor.url within vmprop.c:initializeSystemProperties(vm).
  * JDK12 (assuming future versions as well) sets these properties via java.lang.VersionProps.init(systemProperties) 
@@ -58,8 +58,5 @@
  */
 #define JAVA_VENDOR "Eclipse OpenJ9"
 #define JAVA_VENDOR_URL "http://www.eclipse.org/openj9"
-#endif /* JAVA_SPEC_VERSION < 12 */
-
-#define JAVA_VM_NAME "Eclipse OpenJ9 VM"
 
 #endif     /* vendor_version_h */
--- a/runtime/vm/vmprops.c
+++ b/runtime/vm/vmprops.c
@@ -580,20 +580,24 @@ initializeSystemProperties(J9JavaVM * vm)
 #define J9_STR(x) J9_STR_(x)
 		specificationVersion = J9_STR(JAVA_SPEC_VERSION);
 	}
-	if (JAVA_SPEC_VERSION < 12) {
-		/* System property "java.specification.version" is defined via java.lang.VersionProps.init(systemProperties) within System.ensureProperties() */
-		rc = addSystemProperty(vm, "java.specification.version", specificationVersion, 0);
-		if (J9SYSPROP_ERROR_NONE != rc) {
-			goto fail;
-		}
-	}
+
+	/* Some properties (*.vm.*) are owned by the VM and need to be set early for all
+	 * versions so they are available to jvmti agents.
+	 * Other java.* properties are owned by the class libraries and setting them can be
+	 * delayed until VersionProps.init() runs.
+	 */
 	rc = addSystemProperty(vm, "java.vm.specification.version", specificationVersion, 0);
 	if (J9SYSPROP_ERROR_NONE != rc) {
 		goto fail;
 	}
-	
-	if (JAVA_SPEC_VERSION < 12) {
-		/* Following system properties are defined via java.lang.VersionProps.init(systemProperties) within System.ensureProperties() */
+
+#if JAVA_SPEC_VERSION < 12
+	/* For Java 12, the following properties are defined via java.lang.VersionProps.init(systemProperties) within System.ensureProperties() */
+	rc = addSystemProperty(vm, "java.specification.version", specificationVersion, 0);
+	if (J9SYSPROP_ERROR_NONE != rc) {
+		goto fail;
+	}
+	{
 		const char *classVersion = NULL;
 		if (JAVA_SPEC_VERSION == 8) {
 			classVersion = "52.0";
@@ -604,17 +608,18 @@ initializeSystemProperties(J9JavaVM * vm)
 		if (J9SYSPROP_ERROR_NONE != rc) {
 			goto fail;
 		}
-
-		rc = addSystemProperty(vm, "java.vendor", JAVA_VENDOR, 0);
-		if (J9SYSPROP_ERROR_NONE != rc) {
-			goto fail;
-		}
-
-		rc = addSystemProperty(vm, "java.vendor.url", JAVA_VENDOR_URL, 0);
-		if (J9SYSPROP_ERROR_NONE != rc) {
-			goto fail;
-		}
 	}
+
+	rc = addSystemProperty(vm, "java.vendor", JAVA_VENDOR, 0);
+	if (J9SYSPROP_ERROR_NONE != rc) {
+		goto fail;
+	}
+
+	rc = addSystemProperty(vm, "java.vendor.url", JAVA_VENDOR_URL, 0);
+	if (J9SYSPROP_ERROR_NONE != rc) {
+		goto fail;
+	}
+#endif /* JAVA_SPEC_VERSION < 12 */
 
 	rc = addSystemProperty(vm, "java.vm.vendor", JAVA_VM_VENDOR, 0);
 	if (J9SYSPROP_ERROR_NONE != rc) {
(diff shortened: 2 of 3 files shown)


runtime/vm/vmprops.c
```diff
@@ -572,35 +574,45 @@ initializeSystemProperties(J9JavaVM * vm)
...
-			classVersion = "55.0";
-			specificationVersion = "11";
-			break;
-		}
-		rc = addSystemProperty(vm, "java.class.version", classVersion, 0);
-		if (J9SYSPROP_ERROR_NONE != rc) {
+	if (JAVA_SPEC_VERSION == 8) {
+		specificationVersion = "1.8";
+	} else {
+		bufLen = j9str_printf(PORTLIB, NULL, 0, "%d", JAVA_SPEC_VERSION);
```
pshipton: Can you use the preprocessor for this?
https://stackoverflow.com/questions/240353/convert-a-preprocessor-token-to-a-string
https://github.com/eclipse/openj9/blob/abc3bd33fdef3079aa6e3f4a78b8dffec5b09e27/runtime/ddr/j9ddr.h#L33

runtime/vm/vmprops.c
```diff
@@ -572,35 +574,45 @@ initializeSystemProperties(J9JavaVM * vm)
...
+	}
+	
+	if (JAVA_SPEC_VERSION < 12) {
+		/* Following system properties are defined via java.lang.VersionProps.init(systemProperties) within System.ensureProperties() */
+		const char *classVersion = NULL;
+		if (JAVA_SPEC_VERSION == 8) {
+			classVersion = "52.0";
+		} else {
+			/* Java 11 */
+			classVersion = "55.0";
```
pshipton: This should handle Java 12, 13, etc.

runtime/vm/vmprops.c
```diff
@@ -572,35 +574,45 @@ initializeSystemProperties(J9JavaVM * vm)
...
+			}
+		} else {
+			rc = J9SYSPROP_ERROR_EMPTY_JAVA_SPEC_VERSION;
 			goto fail;
 		}
-
+	}
+	if (JAVA_SPEC_VERSION < 12) {
+		/* System property "java.specification.version" is defined via java.lang.VersionProps.init(systemProperties) within System.ensureProperties() */
 		rc = addSystemProperty(vm, "java.specification.version", specificationVersion, 0);
```
pshipton: I think we should be adding this even for Java 12 and up, even though VersionProps.init() sets it later. Then the value is available in JVMTI.

runtime/include/vendor_version.h
```diff
@@ -49,17 +49,14 @@
 #define VENDOR_SHORT_NAME "OpenJ9"
 
 #define JAVA_VM_VENDOR "Eclipse OpenJ9"
+#define JAVA_VM_NAME "Eclipse OpenJ9 VM"
 
-#if JAVA_SPEC_VERSION < 12
```
pshipton: What is the purpose of this change?

runtime/vm/vmprops.c
DanHeidinga: I'd like to avoid the allocation here if we can.  

Can you look at C preprocessor stringification - doing something like:
```C
#define STRINGIFY(x) #x 
specificationVersion = STRINGIFY(JAVA_SPEC_VERSION)

```

JasonFengJ9: Yeah, trying the hints from Pete's comment about `J9_STR(x)`.

JasonFengJ9: Copied the MACRO instead of including `"../ddr/j9ddr.h"`.

runtime/vm/vmprops.c
JasonFengJ9: `JDK 12+` class versions are set by `java.lang.VersionProps.init(systemProperties)` within `System.ensureProperties()`.

runtime/include/vendor_version.h
JasonFengJ9: `JAVA_VENDOR` & `JAVA_VENDOR_URL` was decorated by `#if JAVA_SPEC_VERSION < 12` in `vmprops.c`, now it is a runtime check instead hence these two need to be defined for `JAVA_SPEC_VERSION  12+` cases as well.

runtime/vm/vmprops.c
JasonFengJ9: Done. Please have another look.

runtime/vm/vmprops.c
```diff
@@ -572,39 +573,35 @@ initializeSystemProperties(J9JavaVM * vm)
...
 	}
 
-#if JAVA_SPEC_VERSION < 12
-	/* Following system properties are defined via java.lang.VersionProps.init(systemProperties) and following settings within System.ensureProperties() */
-	{
+	if (JAVA_SPEC_VERSION == 8) {
+		specificationVersion = "1.8";
+	} else {
+#define J9_STR_(x) #x
+#define J9_STR(x) J9_STR_(x)
```
DanHeidinga: We don't need two levels of this macro here as we're only working with a constant.
```c
#define J9_STR(x) #x 
```

runtime/vm/vmprops.c
```diff
@@ -572,39 +573,35 @@ initializeSystemProperties(J9JavaVM * vm)
...
+		specificationVersion = J9_STR(JAVA_SPEC_VERSION);
+	}
+	rc = addSystemProperty(vm, "java.specification.version", specificationVersion, 0);
```
DanHeidinga: This is set by `VersionProps.init()`which the `java.vm.specification.version` is not.  To me, it makes sense for the VM to own setting the vm specific one, and the class libraries to own the other.

runtime/vm/vmprops.c
pshipton: Since it used to be set in the VM, I think we should be adding this for Java 12 and up, even though VersionProps.init() sets it later. Then the value is available in JVMTI.

pshipton: Note it can be done automatically as 44 + JAVA_SPEC_VERSION

runtime/vm/vmprops.c
pshipton: Since it used to be set in the VM prior to Java 12, I suggesting we should be adding this for Java 12 and up, even though VersionProps.init() sets it later. Then the value is available in JVMTI.

runtime/vm/vmprops.c
JasonFengJ9: It appears it still requires two levels of macro [1] otherwise the system property value would be `JAVA_SPEC_VERSION`. 

[1] https://gcc.gnu.org/onlinedocs/gcc-4.8.5/cpp/Stringification.html

runtime/vm/vmprops.c
JasonFengJ9: Agreed, updated PR to set `java.vm.specification.version` but not `java.specification.version` in native for `JDK12+`.

JasonFengJ9: Oops, missed Pete's comments above (no refresh).

runtime/vm/vmprops.c
JasonFengJ9: Class version technically is owned by VM, set it in native for all Java versions. Please review again.

runtime/vm/vmprops.c
```diff
@@ -572,39 +576,41 @@ initializeSystemProperties(J9JavaVM * vm)
...
+		goto fail;
+	}
+	rc = addSystemProperty(vm, "java.vm.specification.version", specificationVersion, 0);
+	if (J9SYSPROP_ERROR_NONE != rc) {
+		goto fail;
+	}
 
-		rc = addSystemProperty(vm, "java.vm.specification.version", specificationVersion, 0);
-		if (J9SYSPROP_ERROR_NONE != rc) {
+	bufLen = j9str_printf(PORTLIB, NULL, 0, "%d", classVersionValue);
```
pshipton: The class version is XX.0 not XX

runtime/vm/vmprops.c
DanHeidinga: Doesn't this present a consistency issue?  The VM sets something early, its read in jvmti, and then VersionProps sets something subtly different at runtime. 

I would prefer to have the VM own the `*.vm.*` properties and let the class library own the others.  This seems to be inline with the design of VersionProps

runtime/vm/vmprops.c
DanHeidinga: You're right.  I had misread the doc

DanHeidinga:
Pete and I are providing opposite directions and I think that's making this a difficult PR to land.  Let's get the approach laid out and agreed to first.

My view is that properties set by `VersionProps.init()` [1] should NOT be pre-set by the VM as it may lead to consistency issues with subtle differences between values read early by jvmti and those read later.

I would prefer the VM own the `*.vm.*` properties and defer setting the ones from VersionProps to the java code.

This simplifies the VM code and removes the allocations currently being added from the startup paths.

@pshipton Can you comment on what you think the right approach here is?

[1] https://github.com/ibmruntimes/openj9-openjdk-jdk12/blob/master/src/java.base/share/classes/java/lang/VersionProps.java.template

pshipton:
My concern is that apps could see the change in behavior as a regression. Something that used to work up to jdk11, getting certain system properties via JVMTI, no longer works in jdk12 and later. Behavior is allowed to change on major version boundaries, and I do agree there is a risk of being inconsistent in the future if these properties stop following the expected pattern. It would be interesting to know what system properties Hotspot makes available via JVMTI.

JasonFengJ9:
Built a `JVMTI` native and `jvmtiEnv->GetSystemProperty` within `Agent_OnLoad()` can return the system property `java.vm.specification.version` but not `java.specification.version` or `java.class.version`. 
So it appears `RI` native doesn't provide those properties set by `VersionProps.init()` at early phase as Dan expected.
Will update PR accordingly.

runtime/vm/vmprops.c
JasonFengJ9: @DanHeidinga @pshipton the PR has been updated accordingly and ready for another reviewing.

DanHeidinga:

DanHeidinga:

DanHeidinga:
Jenkins test sanity osx jdk12

DanHeidinga:
@JasonFengJ9 I pushed some refactoring changes to your branch and kicked testing off on it.  Let me know if you have any concerns with the modifications.

> Built a JVMTI native and jvmtiEnv->GetSystemProperty within Agent_OnLoad() can return the system property java.vm.specification.version but not java.specification.version or java.class.version. 

Can you add a test for this under a separate PR?

JasonFengJ9:
Thanks @DanHeidinga, the comments are much clearer.
In addition, I see the compile time check is preferred than run time check. It looks good to me.

>Can you add a test for this under a separate PR?

Will do.

pshipton:
@DanHeidinga please cherry pick the commit(s) and create a PR for it against the v0.14.2-release branch.

JasonFengJ9:
Probably I am going to do the cherry pick PR instead.

Issue #5573, opened 2019-04-22
Confusing VerifyError when trying to load class that inherits from final class


Barteks2x:
When trying to load a class that extends a final class, for example with `X extends Y`, the VerifyError points to Y (the final class) being wrong instead of to X, and there is no information as to what is wrong.

OpenJDK and oracle JVM point to the class that attempts to inherit from final class, with message "Cannot inherit from final class".

DanHeidinga:
@ChengJin01 Can you take a look at this?

ChengJin01:
Will check to see what happened to this issue.

ChengJin01:
I managed to reproduce the problem via ASM code (as such problem will be captured as error during compilation) as follows:

the test case:
```
public class X extends Y
{
    public static void main(String args[]) 
    { 
       X x = new X();
    }
}

public final class Y {
}
```
The output on Hotspot  & OpenJ9
```
jdk_11.0.3_7_hotspot/bin/java  -version
openjdk version "11.0.3" 2019-04-16
OpenJDK Runtime Environment AdoptOpenJDK (build 11.0.3+7)
OpenJDK 64-Bit Server VM AdoptOpenJDK (build 11.0.3+7, mixed mode)

jdk_11.0.3_7_hotspot/bin/java  -cp . X
Error: LinkageError occurred while loading main class X
        java.lang.VerifyError: Cannot inherit from final class  <-------------- 

 jdk_11.0.3_7_openj9/bin/java  -version
openjdk version "11.0.3" 2019-04-16
OpenJDK Runtime Environment AdoptOpenJDK (build 11.0.3+7)
Eclipse OpenJ9 VM AdoptOpenJDK (build openj9-0.14.0, JRE 11 Linux amd64-64-Bit Compressed References 20190417_202 (JIT enabled, AOT enabled)
OpenJ9   - bad1d4d06
OMR      - 4a4278e6
JCL      - 5cc996a803 based on jdk-11.0.3+7)

jdk_11.0.3_7_openj9/bin/java  -cp .  X
Error: LinkageError occurred while loading main class X
        java.lang.VerifyError: Y  <------- point to the final class with misleading message
```
As compared to the Hotspot, the verification check for this error on OpenJ9 is correct but the error message offered in such case is pretty simple which leads to confusion.  I will double-check the code related to the message to see how to make it more descriptive.

ChengJin01:
[...]

PR #5586, opened 2019-04-23
Add the detailed error message for class loading

ChengJin01:
The changes are to make the error message more descriptive
for errors thrown in loading classes.

Fix: #5573

### DIFF
--- a/runtime/vm/createramclass.cpp
+++ b/runtime/vm/createramclass.cpp
@@ -146,7 +146,7 @@ static UDATA processVTableMethod(J9VMThread *vmThread, J9ClassLoader *classLoade
 static VMINLINE UDATA growNewVTableSlot(UDATA *vTableAddress, UDATA vTableMethodCount, void *storeValue);
 static UDATA getVTableIndexForNameAndSigStartingAt(UDATA *vTable, J9UTF8 *name, J9UTF8 *signature, UDATA vTableIndex);
 static UDATA checkPackageAccess(J9VMThread *vmThread, J9Class *foundClass, UDATA classPreloadFlags);
-static void setCurrentExceptionForBadClass(J9VMThread *vmThread, J9UTF8 *badClassName, UDATA exceptionIndex);
+static void setCurrentExceptionForBadClass(J9VMThread *vmThread, J9UTF8 *badClassName, UDATA exceptionIndex, U_32 nlsModuleName, U_32 nlsMessageID);
 static BOOLEAN verifyClassLoadingStack(J9VMThread *vmThread, J9ClassLoader *classLoader, J9ROMClass *romClass);
 static void popFromClassLoadingStack(J9VMThread *vmThread);
 static VMINLINE BOOLEAN loadSuperClassAndInterfaces(J9VMThread *vmThread, J9ClassLoader *classLoader, J9ROMClass *romClass, UDATA options, J9Class *elementClass, UDATA packageID, BOOLEAN hotswapping, UDATA classPreloadFlags, J9Class **superclassOut, J9Module *module);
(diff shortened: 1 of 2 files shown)


ChengJin01:
Code reviewer: @DanHeidinga 
FYI:  @pshipton

runtime/vm/createramclass.cpp
```diff
@@ -1586,14 +1599,14 @@ loadSuperClassAndInterfaces(J9VMThread *vmThread, J9ClassLoader *classLoader, J9
 			}
 
 			/* ensure that the superclass isn't an interface or final */
-			if ((superclass->romClass->modifiers & J9AccFinal) != 0) {
+			if (J9_ARE_ANY_BITS_SET(superclass->romClass->modifiers, J9AccFinal)) {
 				Trc_VM_CreateRAMClassFromROMClass_superclassIsFinal(vmThread, superclass);
-				setCurrentExceptionForBadClass(vmThread, superclassName, J9VMCONSTANTPOOL_JAVALANGVERIFYERROR);
+				setCurrentExceptionForBadClass(vmThread, superclassName, J9VMCONSTANTPOOL_JAVALANGVERIFYERROR, "The final superclass can't be inherited");
```
DanHeidinga: These should be NLS messages

runtime/vm/createramclass.cpp
```diff
@@ -1439,15 +1439,28 @@ checkPackageAccess(J9VMThread *vmThread, J9Class *foundClass, UDATA classPreload
...
 {
-	J9MemoryManagerFunctions *gcFuncs = vmThread->javaVM->memoryManagerFunctions;
-	j9object_t detailMessage = gcFuncs->j9gc_createJavaLangString(vmThread, J9UTF8_DATA(badClassName), J9UTF8_LENGTH(badClassName), J9_STR_XLAT);
+	UDATA bufSize = J9UTF8_LENGTH(badClassName) + (UDATA)strlen(utfMessage) + 2;
+	char *errorBuf = NULL;
+	PORT_ACCESS_FROM_VMC(vmThread);
+
+	errorBuf = (char*)j9mem_allocate_memory(bufSize + 1, J9MEM_CATEGORY_CLASSES);
+	if (NULL == errorBuf) {
+		setNativeOutOfMemoryError(vmThread, 0, 0);
```
DanHeidinga: It's better to throw the original exception (even without the message) then to change it to OOM.

runtime/vm/createramclass.cpp
DanHeidinga: "The final superclass can't be inherited" -> "Final superclass can't be extended"

runtime/vm/createramclass.cpp
```diff
@@ -1439,15 +1439,28 @@ checkPackageAccess(J9VMThread *vmThread, J9Class *foundClass, UDATA classPreload
...
+		return;
+	}
+
+	strcpy(errorBuf, (char*)J9UTF8_DATA(badClassName));
+	strcat(errorBuf, ": ");
+	strcat(errorBuf, utfMessage);
+	errorBuf[bufSize] = '\0';
```
DanHeidinga: Consider using j9str_printf() to size the buffer and then also to fill it in.  See the pattern used https://github.com/eclipse/openj9/blob/d87fb6740f293b952e17f12584acdf62444e3e50/runtime/vm/exceptionsupport.c#L853-L859

runtime/vm/createramclass.cpp
ChengJin01: Previously I was considering whether it is necessary to go with NLS messages (withdrew the changes due to too many changes introduced) but I agreed to modify in this case as it makes more sense and clear to us. Already updated the whole solution with the suggested pattern below.

ChengJin01: The output from the new changes as follows:
```
jdk_openj9/bin/java -cp . X
Error: LinkageError occurred while loading main class X
        java.lang.VerifyError: Final superclass Y can't be extended
```

runtime/vm/createramclass.cpp
ChengJin01: Already updated.

runtime/vm/createramclass.cpp
ChengJin01: The whole solution already modified against the pattern above.

runtime/vm/createramclass.cpp
```diff
@@ -1439,15 +1439,37 @@ checkPackageAccess(J9VMThread *vmThread, J9Class *foundClass, UDATA classPreload
...
+		U_16 badClassNameLength = J9UTF8_LENGTH(badClassName);
+		U_8 * badClassNameStr = J9UTF8_DATA(badClassName);
+
+		UDATA errorMsgLen = j9str_printf(PORTLIB, NULL, 0, nlsMessage, badClassNameLength, badClassNameStr);
+		errorMsg = (char*)j9mem_allocate_memory(errorMsgLen, OMRMEM_CATEGORY_VM);
+		if (NULL == errorMsg) {
+			J9MemoryManagerFunctions *gcFuncs = vmThread->javaVM->memoryManagerFunctions;
+			j9object_t detailMessage = gcFuncs->j9gc_createJavaLangString(vmThread, badClassNameStr, badClassNameLength, J9_STR_XLAT);
+			setCurrentException(vmThread, exceptionIndex, (UDATA *)detailMessage);
+			setNativeOutOfMemoryError(vmThread, 0, 0);
```
DanHeidinga: We can't set both exceptions here as the second call will overwrite the first.  Delete the `setNativeOutOfMemoryError` call and only do `setCurrentException`

ChengJin01: Already removed `setNativeOutOfMemoryError`.

DanHeidinga:

DanHeidinga:

PR #5587, opened 2019-04-23
Fix TestJps and update String test utilities

pdbain-ibm:
Rename search() to searchSubstring() and move it to a common utility class.
Add a method to do exact string comparison in  a list.  Update TestJps to use
exact comparison where appropriate.

Minor cleanup: replace ad-hoc setup code with common utilities.

Fixes https://github.com/eclipse/openj9/issues/5569.

[ci skip]


pshipton:
jenkins test extended win32 jdk8

PR #5588, opened 2019-04-23
Cache resolved methods at the compilation level

dmitry-ten:
Previously, resolved methods were cached
per resolved method, i.e. each resolved method
stored a map of resolved methods it called.
A slightly more efficient strategy is to store
resolved methods in one per compilation cache,
which is what this commit implements.

The biggest challenge is that a resolved
method, belonging to one RAM class, might have
different owning methods (callers). Thus, we
cannot simply store a pointer to a resolved method.
Instead, we now cache method info struct, from
which a resolved method with an aprropriate owning
method is recreated at cache access.

### DIFF
--- a/runtime/compiler/control/JITaaSCompilationThread.cpp
+++ b/runtime/compiler/control/JITaaSCompilationThread.cpp
@@ -4264,3 +4264,97 @@ TR::CompilationInfoPerThreadRemote::getCachedIProfilerInfo(TR_OpaqueMethodBlock 
       }
    return ipEntry;
    }
+
+void
+TR::CompilationInfoPerThreadRemote::cacheResolvedMethod(TR_ResolvedMethodKey key, TR_OpaqueMethodBlock *method, uint32_t vTableSlot, TR_ResolvedJ9JITaaSServerMethodInfo &methodInfo)
+   {
+   static bool useCaching = !feGetEnv("TR_DisableResolvedMethodsCaching");
+   if (!useCaching)
+      return;
+
+   TR::Compilation *comp = getCompilation();
+   if (comp->compileRelocatableCode() && comp->getOption(TR_UseSymbolValidationManager) && comp->getSymbolValidationManager()->inHeuristicRegion())
+      {
+      // Problem: If inside of heuristic region, will not create SVM record, so if the same resolved method is queried again,
+      // but outside of heuristic region, we would return cached pointer and not create the validation record, which is wrong.
+      // Temporary solution: don't cache resolved methods if in heuristic region
+      // In a later change should fix it properly by creating SVM record even if retrieved it from cache.
+      return;
+      }
+
+   if (!_resolvedMethodInfoMap)
+      {
+      // Cache not initialized yet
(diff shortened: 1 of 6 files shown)


dmitry-ten:
@mpirvu ready for review.

mpirvu:
> Instead, we now cache method info struct, from which a resolved method with an appropriate owning method is recreated at cache access.

I am guessing that recreating the resolved method at each cache access is not sending additional messages. Also, would be good to measure CPU consumption at the server. I am worried that we might increase the CPU too much there.

runtime/compiler/control/JITaaSCompilationThread.cpp
```diff
@@ -4264,3 +4264,97 @@ TR::CompilationInfoPerThreadRemote::getCachedIProfilerInfo(TR_OpaqueMethodBlock
       }
    return ipEntry;
    }
+
+void
+TR::CompilationInfoPerThreadRemote::cacheResolvedMethod(TR_ResolvedMethodKey key, TR_OpaqueMethodBlock *method, uint32_t vTableSlot, TR_ResolvedMethod *resolvedMethod, TR_ResolvedJ9JITaaSServerMethodInfo &methodInfo)
```
mpirvu: I am seeing `resolvedMethod` being used anywhere in this body

runtime/compiler/control/JITaaSCompilationThread.cpp
```diff
@@ -4264,3 +4264,97 @@ TR::CompilationInfoPerThreadRemote::getCachedIProfilerInfo(TR_OpaqueMethodBlock
...
+   {
+   static bool useCaching = !feGetEnv("TR_DisableResolvedMethodsCaching");
+   if (!useCaching)
+      return;
+
+   TR::Compilation *comp = getCompilation();
+   if (comp->compileRelocatableCode() && comp->getOption(TR_UseSymbolValidationManager) && comp->getSymbolValidationManager()->inHeuristicRegion())
+      {
+      // Problem: If inside of heuristic region, will not create SVM record, so if the same resolved method is queried again,
```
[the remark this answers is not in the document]
dmitry-ten: Not sure how much we lose, but I actually have a fix for this.
Just didn't want to add it to this PR.
Will open another one once this is merged

runtime/compiler/control/JITaaSCompilationThread.cpp
dmitry-ten: Right, it was used in global cache implementation from where I took most of the code. Will remove.

mpirvu:
Also, `TR_ResolvedJ9JITaaSServerMethod::_useCaching` doesn't seem to be used, is it?

dmitry-ten:
Yes, but I deleted it from the code.

dmitry-ten:
Fixed the issue pointed out.
Will post CPU results on the server, once I get the results back.

dmitry-ten:
@mpirvu measured CPU on the server. With this change measured 3% increase, but I did only 3 runs, so this might be just a fluke.

mpirvu:
3% is not a big deal. If we save on messages it should be ok.
If testing was done I am going to merge this item.

dmitry-ten:
sanity passed.

PR #5589, opened 2019-04-23
Restore Escape Analysis refinement

hzongaro:
This pull request restores a refinement to Escape Analysis (EA) that Andrew Craik @andrewcraik had prototyped, and adds a fix for problem it encountered.  That refinement had been reverted under pull request #5421 as a fix for issue #5379.

The problem described in issue #5379 can be reproduced with the following test, which will end up throwing a `RuntimeException` after EA runs, if https://github.com/eclipse/openj9/commit/e6991e539d1437dbfe734abe46826a43f2383434 is applied.  That's because EA allocates the array that is assigned to `z` below on the stack, though it shouldn't be stack allocated, as an instance of `z` allocated in one iteration of the `do-while` loop escapes to the next iteration - that is, two instances of the array assigned to `z` might be defined and referenced during a single iteration of the `do-while` loop.

```
public class Experiment {
    public static final void main(String[] args) {
        for (long i = 0; i < 10000000; i++) {
            tryIt();
        }
    }

    public static int tryIt() {
        int[] y = new int[] {1,2,3,4,5};
        int sum = 0;
        int prevsum = -1;

        do {
            if (prevsum == sum) throw new RuntimeException("Wrong!");

            prevsum = sum;
            int[] z = new int[5];
            for (int i = 0; i < z.length; i++) z[i] = y[i % y.length] * 2;
            for (int i = 0; i < y.length; i++) sum = sum + y[i];
            y = z;
        } while (sum < 10000000);

        return prevsum;
    }
}
```

The problem lies with [this condition in `EscapeAnalysis.cpp`](https://github.com/eclipse/openj9/blob/c2eb8ed7f2ebc2f9d45d6d7c852027f0843747c1/runtime/compiler/optimizer/EscapeAnalysis.cpp#L2375).  In the process of looking at all reaching definitions for a particular use of a candidate for stack allocation, the code is deciding whether those reaching definitions are "harmless."  One case that's considered harmless is a reaching definition that is itself a candidate for stack allocation.

Previously, the check for whether a definition was a candidate checked whether the definition was a reference to the candidate's allocation node.  Commit c2eb8ed7f2ebc2f9d45d6d7c852027f0843747c1 expanded that to allow the value number of the definition to match the candidate's allocation node.

The problem is that, if the current candidate is the allocation that appears in the declaration of `z`, that allows the definition of `y` from the assignment `y = z` that reaches references to `y` to be considered harmless.  That's because the value number of the RHS in that assignment is equal to the value number of the current candidate's allocation node.

The fix in commit https://github.com/eclipse/openj9/commit/eb5edefe6e2600655bf56164f64e8ccb48aaae18 avoids that by only considering value numbers of candidate allocation nodes other than the candidate that's currently under consideration.

### DIFF
--- a/runtime/compiler/optimizer/EscapeAnalysis.cpp
+++ b/runtime/compiler/optimizer/EscapeAnalysis.cpp
@@ -2371,8 +2371,16 @@ bool TR_EscapeAnalysis::checkAllNewsOnRHSInLoopWithAliasing(int32_t defIndex, TR
 
          for (Candidate *otherAllocNode = _candidates.getFirst(); otherAllocNode; otherAllocNode = otherAllocNode->getNext())
             {
+            // A reaching definition that is an allocation node for a candidate
+            // for stack allocation is harmless.  Also, a reaching definition
+            // that has the value number of a candidate allocation, other than the
+            // current candidate, is harmless.  The added restriction in the
+            // second case avoids allowing the current candidate through from a
+            // a previous loop iteration. 
             if (otherAllocNode->_node == firstChild
-                || _valueNumberInfo->getValueNumber(otherAllocNode->_node) == _valueNumberInfo->getValueNumber(firstChild))
+                   || (candidate->_node != otherAllocNode->_node
+                       && _valueNumberInfo->getValueNumber(otherAllocNode->_node)
+                            == _valueNumberInfo->getValueNumber(firstChild)))
                {
                rhsIsHarmless = true;
                break;
--- a/runtime/compiler/optimizer/EscapeAnalysis.cpp
+++ b/runtime/compiler/optimizer/EscapeAnalysis.cpp
@@ -543,7 +543,8 @@ int32_t TR_EscapeAnalysis::performAnalysisOnce()
    _fixedVirtualCallSites.setFirst(NULL);
 
    _parms = NULL;
-   _localObjectsValueNumbers = NULL;
+   _nonColdLocalObjectsValueNumbers = NULL;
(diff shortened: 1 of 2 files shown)


hzongaro:
@vijaysun-omr Vijay, may I ask you to review this pull request?  The change in commit https://github.com/eclipse/openj9/pull/5589/commits/e6991e539d1437dbfe734abe46826a43f2383434 is unchanged from the change that it restores, which you previously reviewed - the new revision lies in commit https://github.com/eclipse/openj9/pull/5589/commits/eb5edefe6e2600655bf56164f64e8ccb48aaae18

vijaysun-omr:
commit eb5edef looks pretty safe to me

vijaysun-omr:
@hzongaro could you please give some more info on the testing done with this change ?

hzongaro:
> @hzongaro could you please give some more info on the testing done with this change ?
> I am just checking so that we reduce the likelihood of needing to revert this change again.

@vijaysun-omr Vijay, thanks for checking.  I ran with all our internal "gold" level tests, on all platforms.  I also retested against the system test that had failed under issue #5379, and with the benchmark snippet described by Charlie Gracie in https://github.com/eclipse/openj9/issues/2072#issue-329487388, as an additional sanity test.

Would you recommend that I initiate a more complete system test run or any other testing?

vijaysun-omr:
Jenkins test extended all JDK8

vijaysun-omr:
Checks have passed and the changes look to have fixed the functional issue that they were being implemented for (while still being safe enough to not cause new issues in my opinion). Merging.

Issue #5578, opened 2019-04-22
JITaaS assert in JITaaS::J9ServerStream::read()


a7ehuo:
Ran into the assert "`Arity mismatch: Expected 13 args to unpack but got 2-tuple`" with debug JITaaS when running some sanity test, such as `_JCL_Test_OutOfMemoryError_SE80_0` or `_TestFlushReflectionCache_SE80_0`. 

Although `J9ServerStream` complains about expecting to read 13 arguments which is a comp request,  the upper layer`CompilationInfoPerThreadRemote::processEntry()` is not actually trying to do so. It tries to read 2x string objects for CHTable update reply from the client. 

Thanks for the suggestion from @dmitry-ten, the assert seems to go away after reverting the change PR #5329 (SHA 38f3172). 

```
(gdb) frame 20
#20 0x00007f7edb0736ee in TR::CompilationInfoPerThreadRemote::processEntry (this=0x7f7ed8cef0a0, entry=..., scratchSegmentProvider=...)
    at /root/home/ahuo/src/openj9-openjdk-jdk8/openj9/runtime/compiler/control/JITaaSCompilationThread.cpp:3886
3886                auto recv = stream->read<std::string, std::string>();
```
```
#16 0x00007f7edb3e49d6 in TR::assertion (file=0x7f7edb927c00 "/root/home/ahuo/src/openj9-openjdk-jdk8/openj9/runtime/compiler/rpc/StreamTypes.hpp", line=68, 
    condition=0x7f7edb927917 "false", format=0x7f7edb927c54 "Arity mismatch: %s")
    at /root/home/ahuo/src/openj9-openjdk-jdk8/openj9/runtime/omr/compiler/infra/Assert.cpp:161
#17 0x00007f7edadfeecc in JITaaS::StreamArityMismatch::StreamArityMismatch (this=0x7f7ebc021210, message="", __in_chrg=<optimized out>, __vtt_parm=<optimized out>)
    at /root/home/ahuo/src/openj9-openjdk-jdk8/openj9/runtime/compiler/rpc/StreamTypes.hpp:68
[...]

PR #5593, opened 2019-04-23
Fix Arity Mismatch Assertion

harryyu1994:
- Server needs to be informed that client has aborted the remote compilation request midway into the compilation.
- This issue was never handled and was exposed when we switched from using per-compilation connections to using per-compilation-thread connections

[skip ci]
Issue: #5578


harryyu1994:
```
void
J9ClientStream::writeError()
   {
   _cMsg.set_status(false);
   _cMsg.mutable_data()->clear_data();
   writeBlocking(_cMsg);
   }
```
```
 template <typename ...T>
   std::tuple<T...> read()
      {
      readBlocking(_cMsg);
      if (!_cMsg.status())
         throw StreamCancel();
      return getArgs<T...>(_cMsg.mutable_data());
      }
```
We actually do have an existing solution to inform the server to abort, but apparently it didn't look, looking into it.

harryyu1994:
The existing solution is not working as intended because messages could arrive out of order. The assertion still gets triggered when the next compilation request arrive before the abort message. We need some sort of an ack.

harryyu1994:
Ready for review

PR #5594, opened 2019-04-23
Fix assumption that object header size is one slot

dmitripivkine:
In GC_MixedObjectScanner constructor object header size is hard coded to
one slot. Technically it is not correct because object header size can
be changed. This code should request MixedObjectModel for value.

### DIFF
--- a/runtime/gc_glue_java/MixedObjectModel.hpp
+++ b/runtime/gc_glue_java/MixedObjectModel.hpp
@@ -1,6 +1,6 @@
 
 /*******************************************************************************
- * Copyright (c) 1991, 2014 IBM Corp. and others
+ * Copyright (c) 1991, 2019 IBM Corp. and others
  *
  * This program and the accompanying materials are made available under
  * the terms of the Eclipse Public License 2.0 which accompanies this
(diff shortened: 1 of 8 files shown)


dmitripivkine:
Jenkins compile win

dmitripivkine:

amicic:
This pattern shows up in a few other spots in the code (both in OpenJ9 and OMR @charliegracie ). Perhaps we should fix them all?

Should we introduce a helper MixedObjectModel::getHeadlessObject(objectPtr) (or some other simpler name like getData, getPayload, etc.) and replace 

(fomrobject_t *)((uint8_t*)objectPtr + env->getExtensions()->mixedObjectModel.getHeaderSize(objectPtr))

with it (and all other spots too)?

charliegracie:
I think an API that returns the data section of an object makes sense.  Also an API to provide the size of that data section would be great.

charliegracie:
fixing all places makes sense

dmitripivkine:
I am suggesting to merge this one now. I will address modifications of similar pattern in another PR.

dmitripivkine:
I see Aleks point... I will add new helper MixedObjectModel::getHeadlessObject(objectPtr)

dmitripivkine:
Jenkins compile win

dmitripivkine:

PR #5595, opened 2019-04-23
Make DecimalFormatHelper optional

JamesKingdon:
By checking if the decimal format helper class is available before
running the decimalFormatPeephole optimisation we can make the JIT
tolerate the absence of the class, effectively making the opt switchable
by adding or removing the helper class. This improves flexibility for
deployment and serviceability.


vijaysun-omr:
Change looks safe to me.

DanHeidinga:

PR #5596, opened 2019-04-23
Change tabs to spaces in echo messages for build.xml files

sharon-wang:
"echo" messages in the compile target for build.xml should contain spaces instead of tabs. 
These messages in build.xml files in the `test` directory have been updated to contain only spaces.

### DIFF
--- a/test/functional/cmdline_options_tester/build.xml
+++ b/test/functional/cmdline_options_tester/build.xml
@@ -1,7 +1,7 @@
 <?xml version="1.0"?>
 
 <!--
-  Copyright (c) 2016, 2018 IBM Corp. and others
+  Copyright (c) 2016, 2019 IBM Corp. and others
 
   This program and the accompanying materials are made available under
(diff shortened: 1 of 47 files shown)


sharon-wang:
@keithc-ca PR created for switching tabs to spaces in build.xml files

test/functional/cmdline_options_tester/build.xml
```diff
@@ -45,11 +45,10 @@
 
 	<target name="compile" depends="init" description="compile the source ">
 		<echo>Ant version is ${ant.version}</echo>
-		<echo>============COMPILER SETTINGS============</echo>
```
keithc-ca: Why was this line removed?

sharon-wang: Missed this deletion in my local diff - fixing!

sharon-wang:
@keithc-ca Reverted the accidental line deletion - this should be good to go now.

keithc-ca:

PR #5603, opened 2019-04-24
Initialize interface class method ordering

gacholio:
The method ordering re-uses the finalizeLinkOffset field, which is
usually 0 for interface classes, but not always. Explicitly initialize
the ordering to NULL when an interface class is created.

[ci skip]


gacholio:

DanHeidinga:
@gacholio how did we find this?  Is there a test case that we can use to validate it?

gacholio:
Issue was discovered by an IBM-internal run of some JRuby tests.

PR #5604, opened 2019-04-24
 Cache resolved methods inside heuristic regions 

dmitry-ten:
If a resolved method is created and cached inside heuristic
region, validation record will not be created.
However, when resolved method is retrieved from the cache
outisde of heuristic region, we need to add validation record,
which is what this change implements


dmitry-ten:
@mpirvu ready for review. Sanity testing already passed.

ymanton:
How does the in-process JIT deal with this? Are resolved methods inside heuristic regions recreated if they're needed outside of heuristic regions rather than being reused?

dmitry-ten:
Yes, I think the code makes sure that if resolved method is created inside heuristic region, it is not reused outside of it.

ymanton:
Jenkins copyright check

ymanton:
Jenkins line endings check

PR #5606, opened 2019-04-24
cache optimization for classByName

satbir-ca:
Currently there are two cache _systemClassByNameMap and
_customClassByNameMap. Both the cache are merged into a
single cache _classByNameMap. Also the _classByNameMap
is a global cache which give a better cache hit.

For unloaded classes purge the global cache(_classByNameMap).

### DIFF
--- a/runtime/compiler/control/JITaaSCompilationThread.cpp
+++ b/runtime/compiler/control/JITaaSCompilationThread.cpp
@@ -2844,7 +2844,7 @@ ClientSessionData::ClientSessionData(uint64_t clientUID, uint32_t seqNo) :
    _chTableClassMap(decltype(_chTableClassMap)::allocator_type(TR::Compiler->persistentAllocator())),
    _romClassMap(decltype(_romClassMap)::allocator_type(TR::Compiler->persistentAllocator())),
    _J9MethodMap(decltype(_J9MethodMap)::allocator_type(TR::Compiler->persistentAllocator())),
-   _systemClassByNameMap(decltype(_systemClassByNameMap)::allocator_type(TR::Compiler->persistentAllocator())),
+   _classByNameMap(decltype(_classByNameMap)::allocator_type(TR::Compiler->persistentAllocator())),
    _classChainDataMap(decltype(_classChainDataMap)::allocator_type(TR::Compiler->persistentAllocator())),
    _unloadedClassAddresses(NULL),
    _requestUnloadedClasses(true),
@@ -2855,7 +2855,7 @@ ClientSessionData::ClientSessionData(uint64_t clientUID, uint32_t seqNo) :
    _inUse = 1;
    _numActiveThreads = 0;
    _romMapMonitor = TR::Monitor::create("JIT-JITaaSROMMapMonitor");
-   _systemClassMapMonitor = TR::Monitor::create("JIT-JITaaSSystemClassMapMonitor");
+   _classMapMonitor = TR::Monitor::create("JIT-JITaaSClassMapMonitor");
    _classChainDataMapMonitor = TR::Monitor::create("JIT-JITaaSClassChainDataMapMonitor");
    _sequencingMonitor = TR::Monitor::create("JIT-JITaaSSequencingMonitor");
    _vmInfo = NULL;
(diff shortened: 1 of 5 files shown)


satbir-ca:

satbir-ca:
resolved

runtime/compiler/control/JITaaSCompilationThread.cpp
```diff
@@ -2919,6 +2919,16 @@ ClientSessionData::processUnloadedClasses(JITaaS::J9ServerStream *stream, const
             continue; // unloaded class was never cached
 
          J9ROMClass *romClass = it->second.romClass;
+
+         J9UTF8 *clazzName = NNSRP_GET(romClass->className, J9UTF8 *);
+         std::string className((const char *)clazzName->data, (int32_t)clazzName->length);
+         J9ClassLoader * cl = (J9ClassLoader *)(it->second.classLoader);
+         ClassLoaderStringPair key = {cl, className};
+            {
+            OMR::CriticalSection classMapCS(getClassMapMonitor());
```
mpirvu: Is there a guarantee that, if a class is present (as a value) in the classByNameMap, it is also present in the `romClassMap`? If, yes, I would like a comment explaining that. If not, the code is wrong because of the `continue` statement before.

mpirvu: Also, you have a nested critical section which can lead to deadlocks if in another part of the code the order of the monitors is different. It's probably better to iterate twice over the list of classes and avoid the nested critical section.

satbir-ca: we cannot reiterate class list as the classRommap entry is delete by that time. I check the code and i not find any where we are entering the critical section in reverse order. For being in the reverse critical section lock we need to call the method trying to take  ROMmapMonitor from method holding the ClassMapMonitor. These monitor are strictly being used to access the cache only.

mpirvu: The new code does a linear search over the `classByNameMap`. This is slow, but how often that situation is encountered is debatable.
Another improvement could be to acquire the romClassMapMonitor only once because locking and unlocking very frequently creates too much overhead. In either case if an ordering is imposed on the two monitors, I would like that order clearly being specified with a comment where monitors are declared/defined because nobody will remember this requirement coming from this piece of code.

Yet another possible improvement (maybe as a separate item) is to ask for information on the j9class when a j9class is inserted into classByNameMap. That way we guarantee that the j9class inside the classByNameMap is also present in the romClassMap

runtime/compiler/control/JITaaSCompilationThread.cpp
```diff
@@ -2947,12 +2958,39 @@ ClientSessionData::processUnloadedClasses(JITaaS::J9ServerStream *stream, const
...
+            {
+            auto itClass = getClassByNameMap().begin();
+            while (itClass != getClassByNameMap().end())
+               {
+               if (itClass->second == it->first)
+                  {
+                  getClassByNameMap().erase(itClass);
+                  break;
+                  }
+               itClass++;
```
mpirvu: typically iterators are incremented with ++it

runtime/compiler/control/JITaaSCompilationThread.cpp
```diff
@@ -2885,6 +2885,7 @@ ClientSessionData::processUnloadedClasses(JITaaS::J9ServerStream *stream, const
    {
    if (TR::Options::getVerboseOption(TR_VerboseJITaaS))
       TR_VerboseLog::writeLineLocked(TR_Vlog_JITaaS, "Server will process a list of %u unloaded classes for clientUID %llu", (unsigned)classes.size(), (unsigned long long)_clientUID);
+   std::unordered_map<TR_OpaqueClassBlock*, ClassUnloadedData> unloadedClassMap;
```
mpirvu: Does this have to be a hashtable? Wouldn't a vector be a better choice?

mpirvu:
Another point is that the code could benefit from more comments, like the point where a linear scan needs to be done.

runtime/compiler/control/JITaaSCompilationThread.cpp
mpirvu: Also, when using a vector, consider the usage of `reserve()` to avoid repeated allocations. You know the final size which the same as `classes.size()`

mpirvu:
In the future (another PR) we may want to consider using ROMMapMonitor to protect all these hashtables that deal with j9classes. Of course we may have to look at performance implications due to using the same lock more often.

satbir-ca:
updated

PR #5609, opened 2019-04-24
Implement z/OS XPLINK system linkage

fjeremic:
To get JitBuilder working on z/OS we need a proper z/OS system linkage in place, that means both frame shape and callout support. At the moment we have some half-baked solution in OMR which does not cut it, since we cannot even run the simplest JitBuilder test cases (see eclipse/omr#3693).

This PR introduces minimal, but proper XPLINK support to OMR and lays out the groundwork for pushing Java-isms out of OMR and consolidating much of the system linkage support across Linux and z/OS. There are many changes in the PR, each of which deserves it's own commit. Each commit comment outlines the specific change made. The work is not done yet, but to avoid feature creep as this is a *deep* rabbit hole I've chosen this as the point where the code is good enough to be reviewed and contributed to the project while I continue to work on it in the background.

The following is a summary of some of the goals of this PR:

- Implement proper XPLINK z/OS system linkage to allow for JitBuilder test cases in eclipse/omr#3693 to run
- Create necessary snippets for XPLINK
- Partition Linux and z/OS system linkages so they're as identical as possible to make the code simple to understand
- Sink OpenJ9 code into OMR and vice-versa
- Document existing and newly added APIs


fjeremic:
Depends on eclipse/omr#3779.

fjeremic:
Jenkins test sanity all depends eclipse/omr#3779

fjeremic:
Jenkins test sanity zlinux depends eclipse/omr#3779

fjeremic:
Jenkins test sanity zlinux depends eclipse/omr#3779

fjeremic:
Jenkins test sanity all depends eclipse/omr#3779

PR #5610, opened 2019-04-24
Reducing the message getClassFromCP

satbir-ca:
Created a constant pool to class cache and check this cache
before making the remote call.

On unloading the classes purge cache. If the class is not
cached iterate the cache for the classes from the unload
class list and remove the entries.

### DIFF
--- a/runtime/compiler/control/JITaaSCompilationThread.cpp
+++ b/runtime/compiler/control/JITaaSCompilationThread.cpp
@@ -2867,6 +2867,7 @@ ClientSessionData::ClientSessionData(uint64_t clientUID, uint32_t seqNo) :
    _J9MethodMap(decltype(_J9MethodMap)::allocator_type(TR::Compiler->persistentAllocator())),
    _classByNameMap(decltype(_classByNameMap)::allocator_type(TR::Compiler->persistentAllocator())),
    _classChainDataMap(decltype(_classChainDataMap)::allocator_type(TR::Compiler->persistentAllocator())),
+   _constantPoolToClassMap(decltype(_constantPoolToClassMap)::allocator_type(TR::Compiler->persistentAllocator())),
    _unloadedClassAddresses(NULL),
    _requestUnloadedClasses(true),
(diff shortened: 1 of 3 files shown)


satbir-ca:

satbir-ca:

runtime/compiler/control/JITaaSCompilationThread.cpp
```diff
@@ -3451,6 +3440,37 @@ JITaaSHelpers::getRemoteROMClassIfCached(ClientSessionData *clientSessionData, J
...
+               }
+            ++itClass;
+            }
+         }
+      //DO NOT remove the entry from the unloadedClasses as it will be needed to purge other caches.
+      ++it;
+      ++data;
+      }
+   }
+
```
ymanton: Since this is a templated method and it's implementation is only in this cpp file it can only be used in this file. Is that your intention? I'm OK with that, it seems like `purgeCache()` doesn't need to be called from anywhere else so there's no point in making it accessible to others, but if someone decides to call it from somewhere else they'll probably get a confusing linker error. Maybe you should add a comment to `purgeCache()` in the header explaining this.

runtime/compiler/control/JITaaSCompilationThread.hpp
```diff
@@ -344,6 +350,8 @@ class JITaaSHelpers
    static bool getAndCacheRAMClassInfo(J9Class *clazz, ClientSessionData *clientSessionData, JITaaS::J9ServerStream *stream, ClassInfoDataType dataType,  void *data);
    static bool getAndCacheRAMClassInfo(J9Class *clazz, ClientSessionData *clientSessionData, JITaaS::J9ServerStream *stream, ClassInfoDataType dataType1, void *data1, ClassInfoDataType dataType2, void *data2);
    static void getROMClassData(const ClientSessionData::ClassInfo &classInfo, ClassInfoDataType dataType, void *data);
+template <typename map, typename key>
+static void purgeCache (std::vector<ClassUnloadedData> *unloadedClasses, map m, key ClassUnloadedData::*k);
```
ymanton: Can you indent these lines to match the others?

satbir-ca:
updated

PR #5613, opened 2019-04-25
Use global name space for shared semaphores on Windows

pdbain-ibm:
Fixes https://github.com/eclipse/openj9/issues/5533

### DIFF
--- a/jcl/src/java.base/share/classes/com/ibm/tools/attach/target/Advertisement.java
+++ b/jcl/src/java.base/share/classes/com/ibm/tools/attach/target/Advertisement.java
@@ -1,7 +1,7 @@
 /*[INCLUDE-IF Sidecar16]*/
 package com.ibm.tools.attach.target;
 /*******************************************************************************
- * Copyright (c) 2009, 2018 IBM Corp. and others
+ * Copyright (c) 2009, 2019 IBM Corp. and others
  *
  * This program and the accompanying materials are made available under
  * the terms of the Eclipse Public License 2.0 which accompanies this
@@ -41,6 +41,7 @@ public final class Advertisement {
 	private static final String KEY_VERSION = "version"; //$NON-NLS-1$
 	private static final String KEY_PROCESS_ID = "processId"; //$NON-NLS-1$
 	private static final String ADVERT_FILENAME = "attachInfo"; //$NON-NLS-1$
+	private static final String GLOBAL_SEMAPHORE = "globalSemaphore"; //$NON-NLS-1$
 	private Properties props;
 	private final long pid, uid;
 
@@ -81,7 +82,7 @@ public final class Advertisement {
 	 * Factory method to read the data from an advertisement file
 	 * @param advertStream InputStream to an open file
 	 * @return Advertisement object containing the data from the file.
-	 * @throws IOException 
+	 * @throws IOException if file cannot be read
 	 */
 	public static Advertisement readAdvertisementFile(InputStream advertStream) throws IOException {
 		Advertisement advert = null;
@@ -109,7 +110,7 @@ public final class Advertisement {
 		addKeyValue(contentBuffer, KEY_VM_ID, vmId);
 		addKeyValue(contentBuffer, KEY_DISPLAY_NAME, (((null == displayName) || (displayName.length() == 0))? vmId: displayName));
 		addKeyValue(contentBuffer, KEY_NOTIFIER, CommonDirectory.MASTER_NOTIFIER);
-		
+		addKeyValue(contentBuffer, GLOBAL_SEMAPHORE, Boolean.TRUE.toString());
 		File tmpTargetDirectoryFileObject = TargetDirectory.getTargetDirectoryFileObject();
 		File tmpSyncFileObject = TargetDirectory.getSyncFileObject();
 		
@@ -270,6 +271,14 @@ public final class Advertisement {
 	 */
 	public long getUid() {
 		return uid;
+	}
+
+	/**
+	 * 
+	 * @return true if the target uses the global semaphore (Windows only)
+	 */
+	public boolean isGlobalSemaphore() {
+		return Boolean.parseBoolean(props.getProperty(GLOBAL_SEMAPHORE));
 	}
 
 	/**
--- a/jcl/src/java.base/share/classes/com/ibm/tools/attach/target/CommonDirectory.java
+++ b/jcl/src/java.base/share/classes/com/ibm/tools/attach/target/CommonDirectory.java
@@ -1,7 +1,7 @@
 /*[INCLUDE-IF Sidecar16]*/
 package com.ibm.tools.attach.target;
 /*******************************************************************************
- * Copyright (c) 2009, 2018 IBM Corp. and others
+ * Copyright (c) 2009, 2019 IBM Corp. and others
  *
  * This program and the accompanying materials are made available under
  * the terms of the Eclipse Public License 2.0 which accompanies this
@@ -230,7 +230,7 @@ public abstract class CommonDirectory {
 	 */
 	static String openSemaphore() throws IOException {
 		String semName = MASTER_NOTIFIER; /*[PR Jazz 48044] semaphore name is a constant */
-		int status = IPC.openSemaphore(getCommonDirFileObject().getAbsolutePath(), semName);
+		int status = IPC.openSemaphore(getCommonDirFileObject().getAbsolutePath(), semName, true);
 		/*[MSG "K0538", "semaphore {0} status= {1}"]*/
 		if (SEMAPHORE_OKAY != status) {
 			throw new IOException(com.ibm.oti.util.Msg.getString("K0538" , semName, Integer.valueOf(status)));  //$NON-NLS-1$
@@ -246,7 +246,7 @@ public abstract class CommonDirectory {
 		int status = 0;
 		IPC.logMessage("reopenSemaphore"); //$NON-NLS-1$
 		closeSemaphore();
-		status = IPC.openSemaphore(getCommonDirFileObject().getAbsolutePath(), MASTER_NOTIFIER);		
+		status = IPC.openSemaphore(getCommonDirFileObject().getAbsolutePath(), MASTER_NOTIFIER, true);		
 		return status;
 	}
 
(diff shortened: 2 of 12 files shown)


runtime/port/win32/j9shsem.c
```diff
@@ -60,6 +62,33 @@ j9shsem_params_init(struct J9PortLibrary *portLibrary, struct J9PortShSemParamet
...
+ * This allows a user with multiple sessions to share the semaphore.
+ * @param portLibrary port library
+ * @param source original UTF-8 string
+ * @param destination buffer to receive the UTF-16 string
+ * @param destinationSize size of destination buffer in UTF-16 characters
+ * @return the destination buffer or NULL on failure.
+ */
+static wchar_t*
+convertToWindowsFormat(OMRPortLibrary *portLibrary, const char *source, wchar_t *destination, uintptr_t destinationSize) {
+	wchar_t *startPosition = destination;
```
pshipton: Please use a better name than `startPosition`

pshipton:
jenkins test sanity win jdk8

pshipton:

runtime/port/win32/j9shsem.c
pdbain-ibm: Okay.  Changed to ```destinationCursor```.

pshipton:
pipeline test sanity,extended win jdk8

pshipton:
Will this introduce a compatibility problem between older JVMs and newer JVMs, as they will be using different semaphores?

pdbain-ibm:
> Will this introduce a compatibility problem between older JVMs and newer JVMs, as they will be using different semaphores?

Yes, they will be incompatible.

pshipton:
> Yes, they will be incompatible.

Then we need a compatibility story and/or migration plan.

pshipton:
One possibility is to add an option for using the global namespace, so user's have to opt in, and understand the consequences.

pshipton:
Or use the global namespace by default, but provide an option to use the original namespace when trying to communicate with older JVMs.

pdbain-ibm:
> One possibility is to add an option for using the global namespace, so user's have to opt in, and understand the consequences.

I like that: it's a more secure.  I will implement same.

pshipton:
I like the idea of using the global namespace by default for OpenJ9 (but not IBM).

pdbain-ibm:
I will add a line in the advertisement file to indicate if the VM is using a global or session semaphore to make it backward compatible.

pdbain-ibm:
I have reworked it so:
- new VMs can talk to old or new VMs _in the same session_ (as before)
- new VMs can talk to new VMs in different sessions (OpenJ9 only).

runtime/port/win32/j9shsem.c
```diff
@@ -58,8 +60,42 @@ j9shsem_params_init(struct J9PortLibrary *portLibrary, struct J9PortShSemParamet
...
+ * @param destination buffer to receive the UTF-16 string
+ * @param destinationSize size of destination buffer in UTF-16 characters
+ * @return the destination buffer or NULL on failure.
+ */
+static wchar_t*
+convertToWindowsFormat(OMRPortLibrary *portLibrary, const char *source, wchar_t *destination, uintptr_t destinationSize, BOOLEAN global) {
+	wchar_t *destinationCursor = destination;
+	wchar_t *result  = destination;
+
+#if defined(OPENJ9_BUILD)
```
pshipton: Why the limitation?

jcl/src/java.base/share/classes/com/ibm/tools/attach/target/Advertisement.java
```diff
@@ -109,7 +110,9 @@ private static StringBuilder createAdvertContent(String vmId, String displayName
 		addKeyValue(contentBuffer, KEY_VM_ID, vmId);
 		addKeyValue(contentBuffer, KEY_DISPLAY_NAME, (((null == displayName) || (displayName.length() == 0))? vmId: displayName));
 		addKeyValue(contentBuffer, KEY_NOTIFIER, CommonDirectory.MASTER_NOTIFIER);
-		
+		/*[IF Sidecar19-SE | Sidecar18-SE-OpenJ9]*/
```
pshipton: Why the limitation?
Also Sidecar19-SE is redundant as Java11 and later include Sidecar18-SE-OpenJ9

runtime/port/win32/j9shsem.c
```diff
@@ -176,10 +214,10 @@ j9shsem_open(struct J9PortLibrary *portLibrary, struct j9shsem_handle **handle,
   
 		if (shsem_handle->mainLock == NULL) {
 			Trc_PRT_shsem_j9shsem_open_Event2(semaphoreName);
-			rc = createSemaphore(portLibrary, shsem_handle);
+			rc = createSemaphore(portLibrary, shsem_handle, global);
```
pshipton: If the semaphore is being created, shouldn't it always be the global semaphore? I'm thinking there should be an assert to validate this.

jcl/src/java.base/share/classes/com/ibm/tools/attach/target/Advertisement.java
pdbain-ibm: Will fix.

jcl/src/java.base/share/classes/com/ibm/tools/attach/target/CommonDirectory.java
```diff
@@ -276,6 +276,29 @@ public static int notifyVm(int numberOfTargets) {
 		return IPC.notifyVm(getCommonDirFileObject().getAbsolutePath(), MASTER_NOTIFIER, numberOfTargets);
 	}
 
+	/**
+	 * Open the semaphore, post to it, and close it
+	 * @param numberOfTargets number of times to post to the semaphore
+	 * @param global Use the global semaphore (Windows only)
+	 * @return 0 on success
+	 */
+	public static int notifyVm(int numberOfTargets, boolean global) {
```
pshipton: It seems unnecessary to add new API and keep the old API, just to avoid changing a few callers. By my count there are only 2 places that need to change.

runtime/port/win32/j9shsem.c
pdbain-ibm: > If the semaphore is being created, shouldn't it always be the global semaphore?

Not if it is an IBM JVM.  See https://github.com/eclipse/openj9/pull/5613#issuecomment-492289205.

jcl/src/java.base/share/classes/com/ibm/tools/attach/target/IPC.java
```diff
@@ -185,19 +183,44 @@ public static void checkOwnerAccessOnly(String filePath) throws IOException {
...
+	/**
+	 * Open the semaphore, post to it, and close it
+	 * @param ctrlDir directory containing the semaphore
+	 * @param semaphoreName name of the semaphore
 	 * @param numberOfTargets number of times to post to the semaphore
 	 * @return 0 on success
 	 */
-	native static int notifyVm(String ctrlDir, String SemaphoreName,
-			int numberOfTargets);
+	static int notifyVm(String ctrlDir, String semaphoreName, int numberOfTargets) {
```
pshipton: It seems unnecessary to add new API and keep the old API, just to avoid changing a few callers. By my count there are only 1 place that need to change.

runtime/port/win32/j9shsem.c
pshipton: That comment was made in a different context. As long as there is no backwards compatibility concern, it doesn't need to be restricted to OpenJ9.

jcl/src/java.base/share/classes/com/ibm/tools/attach/target/IPC.java
pdbain-ibm: Fixed.

jcl/src/java.base/share/classes/com/ibm/tools/attach/target/CommonDirectory.java
pdbain-ibm: Fixed.

runtime/port/win32/j9shsem.c
pdbain-ibm: Sorry, I misinterpreted the comment.

jcl/src/java.base/share/classes/com/ibm/tools/attach/target/Advertisement.java
```diff
@@ -109,7 +110,9 @@ private static StringBuilder createAdvertContent(String vmId, String displayName
 		addKeyValue(contentBuffer, KEY_VM_ID, vmId);
 		addKeyValue(contentBuffer, KEY_DISPLAY_NAME, (((null == displayName) || (displayName.length() == 0))? vmId: displayName));
 		addKeyValue(contentBuffer, KEY_NOTIFIER, CommonDirectory.MASTER_NOTIFIER);
-		
+		/*[IF Sidecar18-SE-OpenJ9]*/
```
pshipton: I thought this was going to be removed, since it is not necessary.

jcl/src/java.base/share/classes/com/ibm/tools/attach/target/CommonDirectory.java
```diff
@@ -282,7 +293,7 @@ public static int notifyVm(int numberOfTargets) {
 	 * @return 0 on success
 	 */
 	public static int cancelNotify(int numberOfTargets) {
```
pshipton: Similarly to notifyVm(), I don't think we need this old API to avoid changing the one caller.

pdbain-ibm: True.

jcl/src/java.base/share/classes/com/ibm/tools/attach/target/Advertisement.java
pdbain-ibm: Fixed.

pdbain-ibm:
Updated per reviewer's comments.

jcl/src/java.base/share/classes/com/ibm/tools/attach/target/IPC.java
```diff
@@ -185,19 +183,23 @@ public static void checkOwnerAccessOnly(String filePath) throws IOException {
...
 	 * @return 0 on success
 	 */
-	native static int notifyVm(String ctrlDir, String SemaphoreName,
-			int numberOfTargets);
+	native static int notifyVm(String ctrlDir, String semaphoreName,
+			int numberOfTargets, boolean global);
 
 	/**
 	 * Open the semaphore, decrement it without blocking to it, and close it
 	 * @param numberOfTargets number of times to decrement to the semaphore
```
pshipton: Might as well add the remaining arguments to the doc, since you fixed notifyVM()

pdbain-ibm: Done.

pshipton:
jenkins test sanity,extended win,win32,plinux jdk8

pshipton:
@AdamBrousseau it would be nice to have a restart option for these cases.

AdamBrousseau:
Should be there for failed tests
https://ci.eclipse.org/openj9/job/Pipeline_Build_Test_JDK8_ppc64le_linux/158/input/
Just need to do it via Jenkins, not PR comment.

pshipton:
Ok sorry. I looked in jenkins, but not in the right spot.

AdamBrousseau:
I suppose we could add a status line that would link straight to the restart page.

pshipton:
> I suppose we could add a status line that would link straight to the restart page.

Good idea. Most people would have no clue where to look.

pshipton:
jenkins test sanity,extended win32 jdk8

pshipton:
jenkins test extended win32 jdk8

pshipton:
Is any documentation update required?

pdbain-ibm:
> Is any documentation update required?

I don't believe so.

pshipton:
I think this does need to be documented, at least for Java 8. Since tools.jar isn't part of every Java runtime, I think sometimes a (partial) copy is used to do attach. These copies need to be updated.

https://github.com/eclipse/openj9-docs/issues/275

PR #5616, opened 2019-04-25
Fix for JIT starvation detection logic

mpirvu:
The JIT has some logic that attempts to detect when compilation threads
are starved due to competition from (potantially many) application
threads. This logic looks at compilation backlog, at CPU utilization
of the JVM and at CPU utilization of the compilation threads.
In some cases though, the CPU utilization of the compilation threads
is low because of other factors, the most common example being
when compilations are performed out-of-process and there is a large
waiting time for compilation results.
In this PR we amend the starvation detection logic such that, if
the JVM does not consume its entire CPU allowance and there are idle
cycles on the machine as well, the JIT cannot be starved (it should be
able to consume those remaining CPU cycles)

### DIFF
--- a/runtime/compiler/control/CompilationThread.cpp
+++ b/runtime/compiler/control/CompilationThread.cpp
@@ -994,7 +994,12 @@ TR_YesNoMaybe TR::CompilationInfo::detectCompThreadStarvation()
 
    // If there are idle cycles on the CPU set where this JVM can run
    // then the compilation threads should be able to use those cycles
-   // (if (idle > 10%) return 0
+   // The following is just an approximation because we look at idle
+   // cyles on the entire machine
+   if (getCpuUtil()->isFunctional() &&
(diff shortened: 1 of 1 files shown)


mpirvu:
jenkins test sanity all jdk8

mpirvu:
> Just so I understand correctly, the effect of this bug was that application threads were being throttled because we thought the compiler threads were starved, when really the compilation threads were just blocked?

correct

ymanton:
OSX failure is

```
12:38:04      [javac] Type=Segmentation error vmState=0x0005ffff
...
12:38:04      [javac] Module=/usr/lib/system/libsystem_c.dylib
12:38:04      [javac] Module_base_address=00007FFF8632E000 Symbol=__findenv
...
12:38:04      [javac] Method_being_compiled=java/lang/Thread.startImpl()V
```

but it looks completely unrelated to this change. Merging.

PR #5617, opened 2019-04-25
Check for NULL before using target_method

dsouzai:
When the inliner finds a profiled receiver, under AOT, the receiver
class and, if the method is inherited, the base class need to be
validated. However, because this involves validation, it is possible for
getResolvedMethod to return NULL. Therefore, a NULL check is required.


dsouzai:
@jdmpapin could you please review.

dsouzai:
jenkins test sanity all jdk8

PR #5623, opened 2019-04-25
Modify RealtimeMarkingScheme for moving to OMR

jason-hall:

### DIFF
--- a/runtime/gc_glue_java/MetronomeDelegate.cpp
+++ b/runtime/gc_glue_java/MetronomeDelegate.cpp
@@ -33,18 +33,28 @@
 #include "ClassLoaderSegmentIterator.hpp"
 #include "EnvironmentRealtime.hpp"
 #include "FinalizableClassLoaderBuffer.hpp"
+#include "FinalizableObjectBuffer.hpp"
+#include "FinalizableReferenceBuffer.hpp"
+#include "FinalizeListManager.hpp"
 #include "FinalizerSupport.hpp"
 #include "GCExtensionsBase.hpp"
 #include "Heap.hpp"
 #include "HeapRegionDescriptorRealtime.hpp"
 #include "MetronomeAlarmThread.hpp"
 #include "JNICriticalRegion.hpp"
+#include "OwnableSynchronizerObjectBufferRealtime.hpp"
 #include "OwnableSynchronizerObjectList.hpp"
 #include "RealtimeAccessBarrier.hpp"
 #include "RealtimeGC.hpp"
 #include "RealtimeMarkingScheme.hpp"
+#include "RealtimeMarkingSchemeRootMarker.hpp"
+#include "RealtimeMarkingSchemeRootClearer.hpp"
+#include "RealtimeMarkTask.hpp"
+#include "RealtimeRootScanner.hpp"
+#include "ReferenceObjectBufferRealtime.hpp"
 #include "ReferenceObjectList.hpp"
 #include "Scheduler.hpp"
+#include "UnfinalizedObjectBufferRealtime.hpp"
 #include "UnfinalizedObjectList.hpp"
 
 void
@@ -142,6 +152,9 @@ MM_MetronomeDelegate::getSplitArraysProcessed(MM_EnvironmentRealtime *env)
 bool
 MM_MetronomeDelegate::initialize(MM_EnvironmentBase *env)
 {
+	_scheduler = _realtimeGC->_sched;
+	_markingScheme = _realtimeGC->getMarkingScheme();
+
 #if defined(J9VM_GC_DYNAMIC_CLASS_UNLOADING)
 	_unmarkedImpliesClasses = false;
 #endif /* defined(J9VM_GC_DYNAMIC_CLASS_UNLOADING) */
@@ -181,6 +194,9 @@ MM_MetronomeDelegate::initialize(MM_EnvironmentBase *env)
 		return false;
 	}
 	MM_GCExtensions::getExtensions(_javaVM)->accessBarrier = (MM_ObjectAccessBarrier *)accessBarrier;
+
+	_javaVM->realtimeHeapMapBasePageRounded = _markingScheme->_markMap->getHeapMapBaseRegionRounded();
+	_javaVM->realtimeHeapMapBits = _markingScheme->_markMap->getHeapMapBits();
 
 	return true;
 }
@@ -270,6 +286,8 @@ MM_MetronomeDelegate::tearDown(MM_EnvironmentBase *env)
 		_extensions->accessBarrier->kill(env);
 		_extensions->accessBarrier = NULL;
 	}
+
+	_javaVM->realtimeHeapMapBits = NULL;
 }
 
 void
@@ -414,7 +432,7 @@ MM_MetronomeDelegate::processDyingClasses(MM_EnvironmentRealtime *env, UDATA* cl
 	while(NULL != (classLoader = (J9ClassLoader *)classLoaderIterator.nextSlot())) {
 		if (0 == (classLoader->gcFlags & J9_GC_CLASS_LOADER_DEAD)) {
 			Assert_MM_true(NULL == classLoader->unloadLink);
-			if(_realtimeGC->_markingScheme->isMarked(classLoader->classLoaderObject) ) {
+			if(_markingScheme->isMarked(classLoader->classLoaderObject) ) {
 				classLoader->gcFlags &= ~J9_GC_CLASS_LOADER_SCANNED;
 			} else {
 				/* Anonymous classloader should not be unloaded */
@@ -478,10 +496,10 @@ MM_MetronomeDelegate::addDyingClassesToList(MM_EnvironmentRealtime *env, J9Class
 				J9CLASS_EXTENDED_FLAGS_CLEAR(clazz, J9ClassGCScanned);
 
 				J9Object *classObject = clazz->classObject;
-				if (setAll || !_realtimeGC->_markingScheme->isMarked(classObject)) {
+				if (setAll || !_markingScheme->isMarked(classObject)) {
 
 					/* with setAll all classes must be unmarked */
-					Assert_MM_true(!_realtimeGC->_markingScheme->isMarked(classObject));
+					Assert_MM_true(!_markingScheme->isMarked(classObject));
 
 					classUnloadCount += 1;
 
@@ -888,20 +906,20 @@ MM_MetronomeDelegate::doClassTracing(MM_EnvironmentRealtime *env)
 					GC_ClassHeapIterator classHeapIterator(_javaVM, segment);
 					J9Class *clazz = NULL;
 					while(NULL != (clazz = classHeapIterator.nextClass())) {
-						if((0 == (J9CLASS_EXTENDED_FLAGS(clazz) & J9ClassGCScanned)) && _realtimeGC->getMarkingScheme()->isMarked(clazz->classObject)) {
+						if((0 == (J9CLASS_EXTENDED_FLAGS(clazz) & J9ClassGCScanned)) && _markingScheme->isMarked(clazz->classObject)) {
 							J9CLASS_EXTENDED_FLAGS_SET(clazz, J9ClassGCScanned);
 
 							/* Scan class */
 							GC_ClassIterator objectSlotIterator(env, clazz);
 							volatile j9object_t *objectSlotPtr = NULL;
 							while((objectSlotPtr = objectSlotIterator.nextSlot()) != NULL) {
-								didWork |= _realtimeGC->getMarkingScheme()->markObject(env, *objectSlotPtr);
+								didWork |= _markingScheme->markObject(env, *objectSlotPtr);
 							}
 
 							GC_ClassIteratorClassSlots classSlotIterator(clazz);
 							J9Class **classSlotPtr;
 							while((classSlotPtr = classSlotIterator.nextSlot()) != NULL) {
-								didWork |= _realtimeGC->getMarkingScheme()->markClass(env, *classSlotPtr);
+								didWork |= markClass(env, *classSlotPtr);
 							}
 						}
 					}
@@ -909,7 +927,7 @@ MM_MetronomeDelegate::doClassTracing(MM_EnvironmentRealtime *env)
 				}
 			} else {
 				/* Check if the class loader has not been scanned but the class loader is live */
-				if( !(classLoader->gcFlags & J9_GC_CLASS_LOADER_SCANNED) && _realtimeGC->getMarkingScheme()->isMarked((J9Object *)classLoader->classLoaderObject)) {
+				if( !(classLoader->gcFlags & J9_GC_CLASS_LOADER_SCANNED) && _markingScheme->isMarked((J9Object *)classLoader->classLoaderObject)) {
 					/* Flag the class loader as being scanned */
 					classLoader->gcFlags |= J9_GC_CLASS_LOADER_SCANNED;
 
@@ -924,13 +942,13 @@ MM_MetronomeDelegate::doClassTracing(MM_EnvironmentRealtime *env)
 							GC_ClassIterator objectSlotIterator(env, clazz);
 							volatile j9object_t *objectSlotPtr = NULL;
 							while((objectSlotPtr = objectSlotIterator.nextSlot()) != NULL) {
-								didWork |= _realtimeGC->getMarkingScheme()->markObject(env, *objectSlotPtr);
+								didWork |= _markingScheme->markObject(env, *objectSlotPtr);
 							}
 
 							GC_ClassIteratorClassSlots classSlotIterator(clazz);
 							J9Class **classSlotPtr;
 							while((classSlotPtr = classSlotIterator.nextSlot()) != NULL) {
-								didWork |= _realtimeGC->getMarkingScheme()->markClass(env, *classSlotPtr);
+								didWork |= markClass(env, *classSlotPtr);
 							}
 						}
 						_realtimeGC->condYield(env, 0);
@@ -950,7 +968,7 @@ MM_MetronomeDelegate::doClassTracing(MM_EnvironmentRealtime *env)
 					hashTableSetFlag(classLoader->classHashTable, J9HASH_TABLE_DO_NOT_REHASH);
 					clazz = _javaVM->internalVMFunctions->hashClassTableStartDo(classLoader, &walkState);
 					while (NULL != clazz) {
-						didWork |= _realtimeGC->getMarkingScheme()->markClass(env, clazz);
+						didWork |= markClass(env, clazz);
 						clazz = _javaVM->internalVMFunctions->hashClassTableNextDo(&walkState);
 
 						/**
@@ -973,9 +991,9 @@ MM_MetronomeDelegate::doClassTracing(MM_EnvironmentRealtime *env)
 					while (NULL != modulePtr) {
 						J9Module * const module = *modulePtr;
 
-						didWork |= _realtimeGC->getMarkingScheme()->markObject(env, module->moduleObject);
-						didWork |= _realtimeGC->getMarkingScheme()->markObject(env, module->moduleName);
-						didWork |= _realtimeGC->getMarkingScheme()->markObject(env, module->version);
+						didWork |= _markingScheme->markObject(env, module->moduleObject);
+						didWork |= _markingScheme->markObject(env, module->moduleName);
+						didWork |= _markingScheme->markObject(env, module->version);
 						modulePtr = (J9Module**)hashTableNextDo(&walkState);
 					}
 				}
@@ -993,7 +1011,7 @@ MM_MetronomeDelegate::doTracing(MM_EnvironmentRealtime* env)
 {
 	/* TODO CRGTMP make class tracing concurrent */
 #if defined(J9VM_GC_DYNAMIC_CLASS_UNLOADING)
-	if(_realtimeGC->getRealtimeDelegate()->isDynamicClassUnloadingEnabled()) {	
+	if(isDynamicClassUnloadingEnabled()) {	
 		return doClassTracing(env);
 	}
 #endif /* J9VM_GC_DYNAMIC_CLASS_UNLOADING */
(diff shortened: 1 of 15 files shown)


jason-hall:
I moved the classes inheriting from RealtimeRootScanner from RealtimeMarkingScheme into RealtimeRootScanner. The intention here is to put RealtimeRootScanner into the glue folder where the other such Scanners live. I also added delegate functions for the java-specific marking code in RealtimeMarkingScheme; OMR needs much simpler implementations

dmitripivkine:
```
../gc_realtime/libj9realtime.a(RealtimeMarkingScheme.cpp.o): In function `MM_RealtimeMarkingScheme::incrementalConsumeQueue(MM_EnvironmentRealtime*, unsigned long)':
/home/travis/build/eclipse/openj9-openjdk-jdk11/build/linux-x86_64-normal-server-release/vm/runtime/gc_realtime/../../../../../openj9/runtime/gc_realtime/RealtimeMarkingScheme.cpp:115: undefined reference to `MM_MetronomeDelegate::scanPointerArraylet(MM_EnvironmentRealtime*, unsigned int*)'
/home/travis/build/eclipse/openj9-openjdk-jdk11/build/linux-x86_64-normal-server-release/vm/runtime/gc_realtime/../../../../../openj9/runtime/gc_realtime/RealtimeMarkingScheme.cpp:118: undefined reference to `MM_MetronomeDelegate::scanObject(MM_EnvironmentRealtime*, J9Object*)'
collect2: error: ld returned 1 exit status
make[6]: *** [runtime/libj9gc29.so] Error 1
make[5]: *** [runtime/gc/CMakeFiles/j9gc.dir/all] Error 2
```

jason-hall:
@dmitripivkine I don't understand it. It compiles fine for me and that error doesn't make any sense. The prototype for the function it says is undefined isn't even in the code anywhere. Its a different function prototype. I can't see how that error is possible. Could it be trying to compile an older version of one of the files somehow?

dmitripivkine:
Jenkins compile win

dmitripivkine:
Failed to compile on Windows as well:
```
21:22:09     Creating library ../lib/j9gc.lib and object ../lib/j9gc.exp
21:22:09  ../makelib/targets.mk:459: UMA_SINGLE_REBASE specified, suppressing per-directory rebase.
21:22:09  j9realtime.lib(RealtimeMarkingScheme.obj) : error LNK2019: unresolved external symbol "public: unsigned __int64 __cdecl MM_MetronomeDelegate::scanObject(class MM_EnvironmentRealtime *,struct J9Object *)" (?scanObject@MM_MetronomeDelegate@@QEAA_KPEAVMM_EnvironmentRealtime@@PEAUJ9Object@@@Z) referenced in function "public: bool __cdecl MM_RealtimeMarkingScheme::incrementalConsumeQueue(class MM_EnvironmentRealtime *,unsigned __int64)" (?incrementalConsumeQueue@MM_RealtimeMarkingScheme@@QEAA_NPEAVMM_EnvironmentRealtime@@_K@Z)
21:22:09  j9realtime.lib(RealtimeMarkingScheme.obj) : error LNK2019: unresolved external symbol "public: unsigned __int64 __cdecl MM_MetronomeDelegate::scanPointerArraylet(class MM_EnvironmentRealtime *,unsigned int *)" (?scanPointerArraylet@MM_MetronomeDelegate@@QEAA_KPEAVMM_EnvironmentRealtime@@PEAI@Z) referenced in function "public: bool __cdecl MM_RealtimeMarkingScheme::incrementalConsumeQueue(class MM_EnvironmentRealtime *,unsigned __int64)" (?incrementalConsumeQueue@MM_RealtimeMarkingScheme@@QEAA_NPEAVMM_EnvironmentRealtime@@_K@Z)
21:22:09  ../j9gc29.dll : fatal error LNK1120: 2 unresolved externals
21:22:09  make[3]: *** [../makelib/targets.mk:458: ../j9gc29.dll] Error 96
21:22:09  make[2]: *** [makefile:1028: j9gc] Error 2
```

runtime/gc_realtime/RealtimeMarkingScheme.cpp
```diff
@@ -20,405 +20,12 @@
...
 #include "RealtimeGC.hpp"
-#include "RealtimeRootScanner.hpp"
 #include "RealtimeMarkingScheme.hpp"
-#include "ReferenceObjectBufferRealtime.hpp"
-#include "ReferenceObjectList.hpp"
-#include "SlotObject.hpp"
-#include "StackSlotValidator.hpp"
-#include "UnfinalizedObjectBufferRealtime.hpp"
-#include "UnfinalizedObjectList.hpp"
-#include "WorkPacketsRealtime.hpp"
```
dmitripivkine: this include still required

dmitripivkine: ...or not. Probably this comment is not correct. Let me double check

dmitripivkine:
Jenkins compile win

runtime/gc_realtime/RealtimeMarkingScheme.hpp
```diff
@@ -51,8 +50,7 @@ class MM_RealtimeMarkingScheme : public MM_SegregatedMarkingScheme
 private:
 	MM_RealtimeGC *_realtimeGC;  /**< The GC that this markingScheme is associated*/
 	MM_Scheduler *_scheduler;
-	J9JavaVM *_javaVM;  /**< The current JVM*/
-	MM_GCExtensions *_gcExtensions;
+	MM_GCExtensionsBase *_gcExtensions;
```
dmitripivkine: looks like this variable is not necessary. The hierarchy is:
`MM_RealtimeMarkingScheme` <- `MM_SegregatedMarkingScheme` <- `MM_MarkingScheme`
and `MM_MarkingScheme` has `MM_GCExtensionsBase *_extensions;` already. It can be used instead

runtime/gc_glue_java/MetronomeDelegate.cpp
```diff
@@ -1115,5 +1130,712 @@ MM_MetronomeDelegate::releaseExclusiveVMAccess(MM_EnvironmentBase *env, bool rel
 	}
 }
 
+UDATA
+MM_MetronomeDelegate::scanPointerRange(MM_EnvironmentRealtime *env, fj9object_t *startScanPtr, fj9object_t *endScanPtr)
+{
```
dmitripivkine: This method use to be inlined too. Can it be moved to .hpp as well?

runtime/gc_glue_java/MetronomeDelegate.cpp
```diff
@@ -1115,5 +1130,712 @@ MM_MetronomeDelegate::releaseExclusiveVMAccess(MM_EnvironmentBase *env, bool rel
...
+	}
+
+	env->addScannedBytes(pointerFieldBytes);
+	env->addScannedPointerFields(pointerField);
+
+	return pointerField;
+}
+
+UDATA
+MM_MetronomeDelegate::scanMixedObject(MM_EnvironmentRealtime *env, J9Object *objectPtr)
```
dmitripivkine: This method use to be inlined too. Can it be moved to .hpp as well?

runtime/gc_glue_java/MetronomeDelegate.cpp
```diff
@@ -1115,5 +1130,712 @@ MM_MetronomeDelegate::releaseExclusiveVMAccess(MM_EnvironmentBase *env, bool rel
...
+
+	env->addScannedBytes(objectSize);
+	env->addScannedPointerFields(pointerFields);
+	env->incScannedObjects();
+
+	return pointerFields;
+}
+
+UDATA
+MM_MetronomeDelegate::scanReferenceMixedObject(MM_EnvironmentRealtime *env, J9Object *objectPtr)
```
dmitripivkine: This method use to be inlined too. Can it be moved to .hpp as well?

runtime/gc_glue_java/MetronomeDelegate.cpp
```diff
@@ -1115,5 +1130,712 @@ MM_MetronomeDelegate::releaseExclusiveVMAccess(MM_EnvironmentBase *env, bool rel
...
+	
+	env->addScannedBytes(objectSize);
+	env->addScannedPointerFields(pointerFields);
+	env->incScannedObjects();
+	
+	return pointerFields;
+}
+
+UDATA
+MM_MetronomeDelegate::scanPointerArrayObject(MM_EnvironmentRealtime *env, J9IndexableObject *objectPtr)
```
dmitripivkine: This method use to be inlined too. Can it be moved to .hpp as well?

jason-hall:
I was able to move those functions into the MetronomeDelegate header by making the following changes.
- The functions all use functions from MM_EnvironmentRealtime. EnvironmentRealtime.hpp includes Scheduler.hpp which includes MetronomeDelegate.hpp. I removed the Scheduler include from EnvironmentRealtime.hpp
- The functions use MM_RealtimeGC->MM_RealtimeMarkingScheme. I added a MM_RealtimeMarkingScheme* field to MetronomeDelegate to fix this
- MM_RealtimeMarkingScheme header used a macro defined in RealtimeGC.hpp, which I moved to RealtimeMarkingScheme.hpp
- Moved MetronomeDelegate include from Scheduler.hpp to Scheduler.cpp. It was unnecessary and now MetronomeDelegate header can include Scheduler
- Moved the one line of code from RealtimeGC::enqueuePointerArraylet (used nowhere else) to where it is used in the delegate

jason-hall:
I rushed these in so you can look at them if you want. I'll cone back to this after my appointment in the afternoon

dmitripivkine:
Jenkins compile win

runtime/gc_glue_java/MetronomeDelegate.hpp
```diff
@@ -157,10 +166,374 @@ class MM_MetronomeDelegate : public MM_BaseNonVirtual
...
+				fj9object_t *startScanPtr = (fj9object_t*) (slotObject.readReferenceFromSlot());
+				if (NULL != startScanPtr) {
+					fj9object_t *endScanPtr = startScanPtr + arrayletSize / sizeof(fj9object_t);
+					if (i == (numArraylets - 1)) {
+						pointerFields += scanPointerRange(env, startScanPtr, endScanPtr);
+						if (canSetAsScanned) {
+							_markingScheme->setScanAtomic((J9Object *)objectPtr);
+						}
+					} else {
+						env->getWorkStack()->push(env, startScanPtr);
```
dmitripivkine: There was call to `MM_RealtimeGC::enqueuePointerArraylet()` here. This function not only push object pointer to Work Stack but use `ARRAYLET_TO_ITEM()` macro to add `ITEM_IS_ARRAYLET` flag to it. Am I correct that this logic is missed here? Looks like it should be
```
env->getWorkStack()->push(env, ARRAYLET_TO_ITEM(startScanPtr));
```

runtime/gc_realtime/RealtimeRootScanner.hpp
```diff
@@ -117,5 +118,377 @@ class MM_RealtimeRootScanner : public MM_RootScanner
...
+	 * Wraps the scanning of one thread to only happen if it hasn't already occured in this phase of this GC,
+	 * also sets the thread up for the upcoming forwarding phase.
+	 * @return true if the thread was scanned (the caller should offer to yield), false otherwise.
+	 * @see MM_RootScanner::scanOneThread()
+	 */
+	virtual void
+	scanOneThreadImpl(MM_EnvironmentRealtime *env, J9VMThread* walkThread, void* localData)
+	{
+		MM_EnvironmentRealtime* walkThreadEnv = MM_EnvironmentRealtime::getEnvironment(walkThread->omrVMThread);
+		MM_RealtimeGC* realtimeGC = (MM_RealtimeGC*)_realtimeGC;
```
dmitripivkine: The cast is not necessary because `_realtimeGC` has type `(MM_RealtimeGC*)` already. I would not declare `realtimeGC` at all but use `_realtimeGC` instead.

runtime/gc_glue_java/MetronomeDelegate.cpp
```diff
@@ -1115,5 +1131,430 @@ MM_MetronomeDelegate::releaseExclusiveVMAccess(MM_EnvironmentBase *env, bool rel
...
+	/* Mark root set classes */
+#if defined(J9VM_GC_DYNAMIC_CLASS_UNLOADING)
+	if(env->isMasterThread()) {
+		/* TODO: This code belongs somewhere else? */
+		/* Setting the permanent class loaders to scanned without a locked operation is safe
+		 * Class loaders will not be rescanned until a thread synchronize is executed
+		 */
+		if(isDynamicClassUnloadingEnabled()) {
+			((J9ClassLoader *)_javaVM->systemClassLoader)->gcFlags |= J9_GC_CLASS_LOADER_SCANNED;
+			_realtimeGC->getMarkingScheme()->markObject(env, (J9Object *)((J9ClassLoader *)_javaVM->systemClassLoader)->classLoaderObject);
```
dmitripivkine: `_realtimeGC->getMarkingScheme()` can be replaced to `_markingScheme` everywhere in `MM_MetronomeDelegate`

runtime/gc_glue_java/MetronomeDelegate.hpp
```diff
@@ -157,10 +166,374 @@ class MM_MetronomeDelegate : public MM_BaseNonVirtual
...
+		return result;
+	}
+	bool markClass(MM_EnvironmentRealtime *env, J9Class *objectPtr);
+#endif /* defined(J9VM_GC_DYNAMIC_CLASS_UNLOADING) */
+
+	MMINLINE uintptr_t scanPointerArraylet(MM_EnvironmentRealtime *env, fomrobject_t *arraylet)
+	{
+		fj9object_t *startScanPtr = arraylet;
+		fj9object_t *endScanPtr = startScanPtr + (_javaVM->arrayletLeafSize / sizeof(fj9object_t));
+		return scanPointerRange(env, startScanPtr, endScanPtr);
```
dmitripivkine: Style: mixing of `fomrobject_t` and `fj9object_t` in the same function looks inconsistent. Considering transition to OMR types I would replace  `fj9object_t` to `fomrobject_t` in this function

runtime/gc_glue_java/MetronomeDelegate.hpp
```diff
@@ -157,10 +166,374 @@ class MM_MetronomeDelegate : public MM_BaseNonVirtual
 	void acquireExclusiveVMAccess(MM_EnvironmentBase *env, bool waitRequired);
 	void releaseExclusiveVMAccess(MM_EnvironmentBase *env, bool releaseRequired);
 
+	void markLiveObjects(MM_EnvironmentRealtime *env);
+#if defined(J9VM_GC_DYNAMIC_CLASS_UNLOADING)
+	MMINLINE void markClassOfObject(MM_EnvironmentRealtime *env, J9Object *objectPtr)
```
dmitripivkine: Style - new look: for functions declared in .hpp files we should use the same style as in .cpp - have declarated types in separate line (at least for new code):
```
MMINLINE void 
markClassOfObject(MM_EnvironmentRealtime *env, J9Object *objectPtr)
```
If it is convenient for you, would you please use this format for this file?

runtime/gc_glue_java/MetronomeDelegate.cpp
```diff
@@ -1115,5 +1131,430 @@ MM_MetronomeDelegate::releaseExclusiveVMAccess(MM_EnvironmentBase *env, bool rel
...
+#if defined(J9VM_GC_FINALIZATION)
+	GC_FinalizableReferenceBuffer buffer(_extensions);
+#endif /* J9VM_GC_FINALIZATION */
+	J9Object* referenceObj = headOfList;
+
+	while (NULL != referenceObj) {
+		objectsVisited += 1;
+		referenceStats->_candidates += 1;
+
+		Assert_MM_true(_realtimeGC->_markingScheme->isMarked(referenceObj));
```
dmitripivkine: also can be replaced to _markingScheme

dmitripivkine:
Jenkins compile win

amicic:
There is a lot of code that goes into Delegate side that should stay in non-delegate side (and eventually move to OMR). 

The prime example is the high level method markLiveObjects(). The whole method should actually reside in non-delegate side, and only two calls to root scanner (hard roots before and weak roots processing after the main tracing loop) should rely on delegate. There are more spots/methods, where a finer line between delegate and non-delegate code could be drawn. 

MarkingScheme in OMR side (used for Standard group of collectors) is good model to follow.

jason-hall:
@amicic I moved all of the markLiveObjects() logic back into RealtimeMarkingScheme as suggested. I agree its a good idea. All other changes here are just the java-specific scanning code (markClass(), scanReferenceMixedObject(), etc).

OMR's MarkingScheme markLiveObjects functions seem to have no logic in them and just directly call the delegate. Do you have any other specific suggestions on other spots/methods?

amicic:
I was under impression that some of indexable code (splitting logic, for example) could move too, but I can see now that even for Standard collectors it resides in Delegate side (or Scanner to be more specific). I guess, in general, i'd like to see that some of it move to OMR, but it's out of scope of this PR.

A suggestion for a few methods renaming to match with Standard GCs:
- doTracing in Realtime very resembles to completeMarking in MarkingScheme
- incrementalConsumeQueue resembles to completeScan, but we can call it incrementalCompleteScan
- markLiveObject could be broken down into: markLiveObjectsInit, markLiveObjectsRoots, markLiveObjectsScan, markLiveObjectsComplete

amicic:
Another minor suggestion, which does not fit into delegate discussion, but it's complying with my suggestion to align code with OMR MarkingScheme as much as possible.

Move some of code from MM_RealtimeMarkTask::setup/cleanup to markLiveObjectsInit/markLiveObjectsComplete.

Maybe there is a chance that MM_RealtimeMarkTask can completely go away and use MM_ParallelMarkTask instead

jason-hall:
I updated the function names to your better suggestions, but I don't think MM_RealtimeMarkTask can be easily replaced with MM_ParallelMarkTask. MM_RealtimeMarkTask is one of two classes inheriting from MM_IncrementalParallelTask which adds a whole bunch of things.

runtime/gc_realtime/RealtimeRootScanner.hpp
```diff
@@ -117,5 +118,376 @@ class MM_RealtimeRootScanner : public MM_RootScanner
 	}
 };
 
+/**
+ * This scanner will mark objects that pass through its doSlot.
+ */
+class MM_RealtimeMarkingSchemeRootMarker : public MM_RealtimeRootScanner
```
dmitripivkine: Would you please create RealtimeMarkingSchemeRootMarker.hpp (and RealtimeMarkingSchemeRootMarker.cpp if necessary) and move this code there?

runtime/gc_realtime/RealtimeRootScanner.hpp
```diff
@@ -117,5 +118,376 @@ class MM_RealtimeRootScanner : public MM_RootScanner
...
+		}
+	}
+protected:
+private:
+};
+
+/**
+ * This scanner will mark objects that pass through its doSlot.
+ */
+class MM_RealtimeMarkingSchemeRootClearer : public MM_RealtimeRootScanner
```
dmitripivkine: Would you please create MM_RealtimeMarkingSchemeRootClearer.hpp (and MM_RealtimeMarkingSchemeRootClearer.cpp if necessary) and move this code there?

jason-hall:
@dmitripivkine @amicic Is the latest update an improvement? The diff looks messy for some reason, but I divided the existing markLiveObjects() into those functions without changing the content

runtime/gc_realtime/RealtimeMarkingScheme.cpp
```diff
@@ -480,502 +83,88 @@ void
...
-	
-#if defined(J9VM_GC_DYNAMIC_CLASS_UNLOADING)
-	if(_realtimeGC->getRealtimeDelegate()->isDynamicClassUnloadingEnabled()) {
-		markClassOfObject(env, objectPtr);
-	}
-#endif /* J9VM_GC_DYNAMIC_CLASS_UNLOADING */
+	metronomeDelegate->markLiveObjectsRoots(realtimeEnv);
+	_scheduler->condYieldFromGC(realtimeEnv);
+	/* Heap Marking and barrier processing. Cannot delay barrier processing until the end.*/	
+	_realtimeGC->completeMarking(realtimeEnv);
```
amicic: This is where markLiveObjectsScan should start (including this line).

amicic: Again the model is base MarkingScheme (non-realtime). Over there markLiveObjectsScan, does completeMarking, what is the minimum core work. RT does a bit more.

jason-hall: Okay, that's a simple fix. Done

runtime/gc_glue_java/MetronomeDelegate.cpp
```diff
@@ -1115,5 +1133,396 @@ MM_MetronomeDelegate::releaseExclusiveVMAccess(MM_EnvironmentBase *env, bool rel
...
+{
+	MM_RealtimeMarkingSchemeRootClearer rootScanner(env, _realtimeGC);
+	env->setRootScanner(&rootScanner);
+	rootScanner.scanClearable(env);
+	env->setRootScanner(NULL);
+	Assert_MM_true(env->getGCEnvironment()->_referenceObjectBuffer->isEmpty());
+}
+
+void
+MM_MetronomeDelegate::flushRememberedSet(MM_EnvironmentRealtime *env)
```
amicic: This is not remember set. 
This should be renamed to flushReferenceObjects (and possibly called from delegate's markLiveObjectsScan - not sure where exactly markLiveObjectsScan is called from)

runtime/gc_glue_java/MetronomeDelegate.cpp
```diff
@@ -1115,5 +1133,396 @@ MM_MetronomeDelegate::releaseExclusiveVMAccess(MM_EnvironmentBase *env, bool rel
...
+		env->_currentTask->releaseSynchronizedGCThreads(env);
+	}
+
+	env->setRootScanner(NULL);
+}
+
+void
+MM_MetronomeDelegate::markLiveObjectsScan(MM_EnvironmentRealtime *env)
+{
+	MM_RealtimeMarkingSchemeRootClearer rootScanner(env, _realtimeGC);
```
amicic: Does not look right. Clearable roots scanning belongs to complete phase. Again, look at the base MarkingScheme.

runtime/gc_realtime/RealtimeMarkingScheme.cpp
```diff
@@ -480,502 +83,89 @@ void
...
-		Assert_MM_unreachable();
+	if (realtimeEnv->_currentTask->synchronizeGCThreadsAndReleaseMaster(realtimeEnv, UNIQUE_ID)) {
+		metronomeDelegate->setUnmarkedImpliesCleared();
+		realtimeEnv->_currentTask->releaseSynchronizedGCThreads(realtimeEnv);
 	}
 
-	GC_SlotObject referentPtr(_javaVM->omrVM, &J9GC_J9VMJAVALANGREFERENCE_REFERENT(env, objectPtr));
+	MM_EnvironmentRealtime *realtimeEnv = MM_EnvironmentRealtime::getEnvironment(env);
+	/* Process reference objects and finalizable objects. */
+	_realtimeGC->getRealtimeDelegate()->markLiveObjectsScan(realtimeEnv);
```
amicic: This should be called earlier. Before the sync point. And the only thing it should call is flushReferenceObjects.

runtime/gc_glue_java/MetronomeDelegate.cpp
jason-hall: Moving it to the delegate function would change the order of the code from its original implementation. possibly breaking it. markLiveObjectsScan() is called from the spot you showed me the other day in RealtimeMarkTask.

runtime/gc_glue_java/MetronomeDelegate.cpp
jason-hall: Again, if you look at the base class, it has no such code in the complete phase. There are no clearable roots in any phase. I'm not sure why you think its an example I can follow. Perhaps you are thinking of another class somewhere else entirely.

void MM_MarkingScheme::markLiveObjectsComplete(MM_EnvironmentBase *env)
{
	_delegate.workerCompleteGC(env);
}

MMINLINE void MM_MarkingDelegate::workerCompleteGC(MM_EnvironmentBase *env)
{
		/* All threads flush buffers before this point, and complete any remaining language-specific marking tasks */
		if (env->_currentTask->synchronizeGCThreadsAndReleaseSingleThread(env, UNIQUE_ID)) {
			/* Perform single-threaded tasks here */
			env->_currentTask->releaseSynchronizedGCThreads(env);
		}
}

runtime/gc_realtime/RealtimeMarkingScheme.cpp
jason-hall: Again, this would change the original order of the code and break it.

runtime/gc_glue_java/MetronomeDelegate.cpp
jason-hall: Moving the root clearer code to markLiveObjectsRoots, as well, would change the original order of the code. I'm worried that will break it

amicic: It does have it - not sure why you did not see it: https://github.com/eclipse/openj9/blob/master/runtime/gc_glue_java/MarkingDelegate.cpp#L158

runtime/gc_realtime/RealtimeMarkingScheme.cpp
amicic: it will not the break the order if you change what it does, it should be exactly at the spot what now flushReferenceObjects is called

jason-hall:
@amicic It seems your suggestions on how to break down the markLiveObjects() into phases means significantly rewriting the entire order of of the steps of the original code. I'm concerned that may break something? Also, the base class is only stub functions with no implementations. It doesn't really serve as an example.

runtime/gc_realtime/RealtimeMarkingScheme.cpp
amicic: as I explained curently markLiveObjectsScan in delegate does the wrong thing. This is the main scan work not mean for any of root passes

amicic:
when I talk about base class I mean MarkingScheme in runtime/gc_glue_java. It's a parrent calss of RealtimeMarkingScheme. It's not a class with empty stub functions

amicic:
And similarly for the delegate you should look into MarkingDelegate in runtime/gc_glue/java, Don't look into example delegate in OMR - they don't have any language specific, that Realtime or standard Java Marking has.

amicic:
There is no significant rewrite -it's just an intorduction of 4 wrappers (well 8, for Scheme and Delegate) and moving the code from markLiveObjects to those wrappers . Order must not and will not be changed - it's not hard

jason-hall:
@amicic I didn't find it perfectly clear or maybe I misread something. I think I understand it correctly now with the additional details. Thank you.

I'm rewriting my phases to be more like the logic in the gc_glue_java/MarkingDelegate functions called by the stub parent base class MarkingScheme at omr/gc/MarkingScheme as suggested.

amicic:
looks about right after the last changes, but I'll have another fresh look

runtime/gc_realtime/RealtimeMarkingSchemeRootClearer.hpp
```diff
@@ -0,0 +1,291 @@
...
+	MM_RealtimeMarkingSchemeRootClearer(MM_EnvironmentRealtime *env, MM_RealtimeGC *realtimeGC) :
+		MM_RealtimeRootScanner(env, realtimeGC)
+	{
+		_typeId = __FUNCTION__;
+	}
+	
+	/**
+	 * This should not be called
+	 */
+	MMINLINE virtual void
```
dmitripivkine: Declaration `MMINLINE` is ignored with `virtual` and can be removed... here and everywhere

runtime/gc_realtime/RealtimeMarkingSchemeRootClearer.hpp
```diff
@@ -0,0 +1,291 @@
...
+	doSlot(J9Object** slot)
+	{
+		PORT_ACCESS_FROM_ENVIRONMENT(_env);
+		j9tty_printf(PORTLIB, "MM_RealtimeMarkingSchemeRootClearer::doSlot should not be called\n");
+		assert(false);
+	}
```
dmitripivkine: body of this function should be replaced to
```
{
    Assert_MM_unreachable();
}
```

jason-hall:
@dmitripivkine @amicic Is this done or did I miss anything?

dmitripivkine:
Jenkins compile win

dmitripivkine:

dmitripivkine:
Jenkins compile win jdk8

dmitripivkine:
Jenkins test sanity AIX jdk11

dmitripivkine:
Jenkins test sanity AIX jdk11

PR #5634, opened 2019-04-29
Implement JIT fields cache

dmitry-ten:
`TR_J9ServerVM::jitFieldsAreSame` compares 2 fields from classes
at given CP indices, and returns true if they are the same,
false otherwise.
We cache the declaring class and value of a field per RAM class,
which enables us to significantly reduce the number of remote calls
made to the client.

### DIFF
--- a/runtime/compiler/control/JITaaSCompilationThread.cpp
+++ b/runtime/compiler/control/JITaaSCompilationThread.cpp
@@ -48,6 +48,62 @@ extern TR::Monitor *assumptionTableMutex;
 
 uint32_t serverMsgTypeCount[JITaaS::J9ServerMessageType_ARRAYSIZE] = {};
 
+// TODO: this method is copied from runtime/jit_vm/ctsupport.c,
+// in the future it's probably better to make that method publicly accessible
+static UDATA
+findField(J9VMThread *vmStruct, J9ConstantPool *constantPool, UDATA index, BOOLEAN isStatic, J9Class **declaringClass)
+   {
+   J9JavaVM *javaVM = vmStruct->javaVM;
+   J9ROMFieldRef *romRef; 
+   J9ROMClassRef *classRef; /* ROM class of the field */
+   U_32 classRefCPIndex;
+   J9UTF8 *classNameUTF;
+   J9Class *clazz;
+   UDATA result = 0;
+
+   *declaringClass = NULL;
+   romRef = (J9ROMFieldRef*) &(((J9ROMConstantPoolItem*) constantPool->romConstantPool)[index]);
+   classRefCPIndex = romRef->classRefCPIndex;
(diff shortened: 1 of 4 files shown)


runtime/compiler/control/JITaaSCompilationThread.cpp
```diff
@@ -313,7 +362,66 @@ bool handleServerMessage(JITaaS::J9ClientStream *client, TR_J9VM *fe)
...
+               }
+            }      
+         if (compareFields)
+            {
+             f1 = findField(fe->vmThread(), cp1, cpIndex1, isStatic, &declaringClass1);
+             if (f1)
+                {
+                f2 = findField(fe->vmThread(), cp2, cpIndex2, isStatic, &declaringClass2);
+                if (f2)
+                   identical = ((f1 == f2) && (declaringClass1 == declaringClass2));
```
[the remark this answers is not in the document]
dmitry-ten: No, I just forgot to delete this after copy-pasting

dmitry-ten:
@mpirvu could you look at this again?

runtime/compiler/control/JITaaSCompilationThread.cpp
```diff
@@ -48,6 +48,54 @@ extern TR::Monitor *assumptionTableMutex;
 
 uint32_t serverMsgTypeCount[JITaaS::J9ServerMessageType_ARRAYSIZE] = {};
 
+static UDATA
+findField(J9VMThread *vmStruct, J9ConstantPool *constantPool, UDATA index, BOOLEAN isStatic, J9Class **declaringClass)
```
mpirvu: In the future we may want to export this from the VM and call it directly.
Please add a comment where you took this method from.

runtime/compiler/control/JITaaSCompilationThread.cpp
```diff
@@ -313,7 +361,62 @@ bool handleServerMessage(JITaaS::J9ClientStream *client, TR_J9VM *fe)
...
          int32_t isStatic = std::get<4>(recv);
-         client->write(fe->jitFieldsAreSame(method1, cpIndex1, method2, cpIndex2, isStatic));
+         bool identical = false;
+         bool compareFields = false;
+         UDATA f1 = 0, f2 = 0;
+         J9Class *declaringClass1 = NULL, *declaringClass2 = NULL;
+         J9ConstantPool *cp1 = (J9ConstantPool *) method1->ramConstantPool();
+         J9ConstantPool *cp2 = (J9ConstantPool *) method2->ramConstantPool();
+
+         J9RAMFieldRef *ramRef1 = (J9RAMFieldRef*) &(((J9RAMConstantPoolItem *)cp1)[cpIndex1]);
```
mpirvu: Please add a comment with information where this code comes from. If the VM code ever changes and our caching validation catches the error, we need to remember where to look for possible discrepancies.

dmitry-ten:
Addressed comments.
Hit rate for the cache is higher than 50%. 
CPU improvement is around 5%.

PR #5638, opened 2019-04-29
Creating a Jenkins job that will create wrapper jobs

AdamBrousseau:
There are three parts to this:

Wrapper_job.groovy
Wrapper_variables.yml
Wrapper_template
When this is properly configured, Jenkins will have a pipeline with
the different configurations as Boolean values
The pipeline will then get the configuration values from the yaml file
It will then call the template with those values
The template will then create the jobs to the specification

[skip ci]

Signed-off-by: Samuel Rubin samuel.rubin@ibm.com


AdamBrousseau:
@pshipton 
Fixes #4846

AdamBrousseau:
Surprised to see the Eclipse check pass. I thought I had done something like this before but it failed. Maybe an enhancement to the new check?

PR #5639, opened 2019-04-30
Don't check classfiles >= 51 for JSRs with -Xverify:none

DanHeidinga:
Classes of version 51 or greater are not allowed to continue JSR/RET
bytecodes.  If running with -Xverify:none, then we don't need to do
the check for them.  Any issues would be reported by enabling the
verifier.

Currently, verification isn't enabled for the bootclasspath so this
change should improve startup for bootclasspath classes

### DIFF
--- a/runtime/bcutil/cfreader.c
+++ b/runtime/bcutil/cfreader.c
@@ -2668,9 +2668,13 @@ j9bcutil_readClassFileBytes(J9PortLibrary *portLib,
 			return result;
 		}
 	} else {
-		/* special checking to look for jsr's if -noverify - the verifyFunction normally scans and tags 
-			for inlining methods and classes that contain jsr's */
-		hasRET = checkForJsrs(classfile);
+		/* Special checking to look for jsr's if -noverify - the verifyFunction normally scans and tags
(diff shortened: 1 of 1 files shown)


DanHeidinga:
@ChengJin01 Can you review this?

FYI @vijaysun-omr this is one of the changes we discussed

DanHeidinga:

pshipton:
Note there is a typo in the commit comment. continue -> contain

DanHeidinga:
Fixed the typo

DanHeidinga:

DanHeidinga:
The build failed due to bad boot sdk:
```
14:16:18  configure: Found potential Boot JDK using configure arguments
14:16:18  configure: Potential Boot JDK found at /usr/lib/jvm/jdk-7 did not contain an rt.jar; ignoring
14:16:18  configure: error: The path given by --with-boot-jdk does not contain a valid Boot JDK
```

pshipton:

AdamBrousseau:
>The build failed due to bad boot sdk:

This machine (cent6-1) had a system update which somehow killed the java7 openjdk install. We should standardize using bootjdk8 for 8 with OpenJ9 from Adopt.

PR #5640, opened 2019-04-30
Read new field value before initializing the class

gacholio:
When reporting a static field write from the JIT, the new value pointer
passed in is often a pointer into the java stack. If the report helper
causes the class to be initialized, the stack may grow, invalidating the
pointer.

Read the value from the pointer into a local before initializing the
class.

[ci skip]

### DIFF
--- a/runtime/codert_vm/cnathelp.cpp
+++ b/runtime/codert_vm/cnathelp.cpp
@@ -3122,6 +3122,12 @@ old_slow_jitReportStaticFieldWrite(J9VMThread *currentThread)
 	if (J9_EVENT_IS_HOOKED(vm->hookInterface, J9HOOK_VM_PUT_STATIC_FIELD)) {
 		J9Class *fieldClass = dataBlock->fieldClass;
 		if (J9_ARE_ANY_BITS_SET(fieldClass->classFlags, J9ClassHasWatchedFields)) {
+			/* Read the data now, as the incoming pointer is potentially into the
+			 * java stack, which would be invalidated by the class initialization below.
+			 * Ensure the read does not get reordered below by the compiler.
+			 */
+			U_64 value = *(U_64 volatile *)valuePointer;
+			VM_AtomicSupport::compilerReorderingBarrier();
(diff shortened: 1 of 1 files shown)


gacholio:
jenkins test sanity win jdk11

runtime/codert_vm/cnathelp.cpp
```diff
@@ -3122,6 +3122,7 @@ old_slow_jitReportStaticFieldWrite(J9VMThread *currentThread)
 	if (J9_EVENT_IS_HOOKED(vm->hookInterface, J9HOOK_VM_PUT_STATIC_FIELD)) {
 		J9Class *fieldClass = dataBlock->fieldClass;
 		if (J9_ARE_ANY_BITS_SET(fieldClass->classFlags, J9ClassHasWatchedFields)) {
+			U_64 value = *(U_64*)valuePointer;
```
DanHeidinga: Does this need to be marked `volatile` to ensure the compiler doesn't move the read?

gacholio: What do you think needs to be volatile? I strongly suspect that the fact that there's a call between the read and the use would likely prevent the motion.

Given the existing checks before the read, I would have no issue marking valuePointer and it's uses as volatile. I strongly dislike marking locals volatile.

gacholio: Another (not necessarily better) option would be the explciit compiler reordering barrier which could be inserted just before the hook dispatch.

DanHeidinga: I agree the reorder barrier would be sufficient but doesn't it need to be before the call to `initializeClass()` which is where the stack grow may occur

gacholio: Yes, that would make more sense - I'll add one there for the sake of it (still pretty sure it's not necessary as we don't do it in many other places).

gacholio: Added a volatile and a barrier.

gacholio:
jenkins test sanity win jdk11

DanHeidinga:
Jenkins compile osx,aix jdk8

PR #5642, opened 2019-04-30
Miscellaneous updates specific to z/TPF

jjohnst-us:
z/TPF had a build path change which needs to be propagated to the
build (redhat subpath is now gnu).  Additionally, a sleep needed
to be added as part of the z/TPF specific ddrgen and codegen to
allow for some build services to complete.

### DIFF
--- a/runtime/port/sysvipc/j9sharedhelper.c
+++ b/runtime/port/sysvipc/j9sharedhelper.c
@@ -137,11 +137,11 @@ createDirectory(struct J9PortLibrary *portLibrary, char *pathname, uintptr_t per
 		return J9SH_SUCCESS;
 	} else if (J9PORT_ERROR_FILE_EXIST == omrerror_last_error_number()) {
 		Trc_PRT_shared_createDirectory_Exit1();
-		return J9SH_SUCCESS; 
+		return J9SH_SUCCESS;
 	}
-		
+
 	omrstr_printf(tempPath, J9SH_MAXPATH, "%s", pathname);
-		
+
 	current = strchr(tempPath+1, DIR_SEPARATOR); /* skip the first '/' */
 
 	if ((J9SH_DIRPERM_ABSENT == permission)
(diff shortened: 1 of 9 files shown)


jjohnst-us:
Is there a way I can tell which files have the bad copyright?

DanHeidinga:
> Is there a way I can tell which files have the bad copyright?

Normally, I'd say click the details link and then look at the console output in Jenkins but this build failed with an OOM so I'll relaunch

DanHeidinga:
Jenkins line endings check

DanHeidinga:
Jenkins copyright check

runtime/port/sysvipc/j9sharedhelper.c
```diff
@@ -293,7 +293,11 @@ cleanSharedMemorySegments(struct J9PortLibrary* portLibrary)
 			return;
 		}
 	}
-#else
+#elif defined(J9ZTPF) /* defined(J9ZOS390) */
+		/* The z/TPF platform does not support the ipcs command */
```
DanHeidinga: This looks to be indented one level too deep

DanHeidinga:
Jenkins copyright check

DanHeidinga:
Jenkins line endings check

jjohnst-us:
@DanHeidinga With respects to the sleep... it is a can of worms, but keep in mind it would only be a pain point for z/TPF if it does break in the future.  One layer of this complexity is because both our ddrgen and constgen are run as services  on our z/TPF system (on-line) whereas our build is offline on a Linux system.  The build pauses until these steps complete.  As part of currently internal continuous integration we launch a java program which runs the original build script and we have re-directed the standard output and error to a file.  The file contents is actually used as part of the build, because we need to 'find' the link statements and modify them to create z/TPF shared objects (part of the original build script).  So added the sleep because there is a lag as the file is being written by the time we actually go to do the link.  (In other we go to 'find' the link for ddr and constgen, but it wasn't written out yet).  20 seconds is plenty of a delay for the file to be completely written out when the link statement goes to execute.

Hope that helps clarify.

runtime/port/sysvipc/j9sharedhelper.c
jjohnst-us: Update pushed.

DanHeidinga:

PR #5646, opened 2019-04-30
Fix logic that determines when to do load without rdbar

yanluo7:
Use `gc_modron_readbar_none` as condition to check when rdbar needs
to be involved or a simple regular load is needed. Fixes assert on IA32 where 
we erroneously get into rdbar code.


andrewcraik:
@yanluo7 could you explain why the rdbar requirement check checks if a wrtbar or a rdbar is required? Surely the right check is just if the rdbar is required? Some explanation/discussion would be warranted since it seems counter intuitive looking at the code...

yanluo7:
> @yanluo7 could you explain why the rdbar requirement check checks if a wrtbar or a rdbar is required? Surely the right check is just if the rdbar is required? Some explanation/discussion would be warranted since it seems counter intuitive looking at the code...

The existing wrtbar type check appears to be a leftover mistake -- GC folks was also puzzled by it. Removed and replaced with just the rdbar type check.

andrewcraik:

yanluo7:
@andrewcraik can be merged.

PR #5647, opened 2019-04-30
Fix queries for HW and SW read barrier on z

NigelYiboYu:
Fix queries used to emit HW and SW read barrier instructions so that
correct instructions are emitted for HW and SW CS. This ensures that
z13 and below gets software-based CS and z14 gets hardware CS whenever
-Xgc:concurrentScavenge is enabled, and that HW CS can be turned into SW CS 
when -XXgc:softwareRangeCheckReadBarrier is supplied.

### DIFF
--- a/runtime/compiler/z/codegen/J9TreeEvaluator.cpp
+++ b/runtime/compiler/z/codegen/J9TreeEvaluator.cpp
@@ -10126,11 +10126,7 @@ VMinlineCompareAndSwap(TR::Node *node, TR::CodeGenerator *cg, TR::InstOpCode::Mn
    if (TR::Compiler->om.readBarrierType() != gc_modron_readbar_none && isObj)
       {
       TR::Register* tempReadBarrier = cg->allocateRegister();
-      if (TR::Compiler->om.shouldReplaceGuardedLoadWithSoftwareReadBarrier())
-         {
-         TR::TreeEvaluator::generateSoftwareReadBarrier(node, cg, tempReadBarrier, generateS390MemoryReference(*casMemRef, 0, cg));
-         }
-      else
+      if (TR::Compiler->target.cpu.getSupportsGuardedStorageFacility())
          {
          auto guardedLoadMnemonic = comp->useCompressedPointers() ? TR::InstOpCode::LLGFSG : TR::InstOpCode::LGG;
 
@@ -10139,6 +10135,10 @@ VMinlineCompareAndSwap(TR::Node *node, TR::CodeGenerator *cg, TR::InstOpCode::Mn
          // a read barrier
          generateS390IEInstruction(cg, TR::InstOpCode::NIAI, 1, 0, node);
          generateRXInstruction(cg, guardedLoadMnemonic, node, tempReadBarrier, generateS390MemoryReference(*casMemRef, 0, cg));
+         }
+      else
+         {
+         TR::TreeEvaluator::generateSoftwareReadBarrier(node, cg, tempReadBarrier, generateS390MemoryReference(*casMemRef, 0, cg));
          }
       cg->stopUsingRegister(tempReadBarrier);
       }
@@ -12081,15 +12081,15 @@ J9::Z::TreeEvaluator::ardbarEvaluator(TR::Node * node, TR::CodeGenerator * cg)
       // MemRef can generate BRCL to unresolved data snippet if needed.
       TR::MemoryReference* loadMemRef = generateS390MemoryReference(node, cg);
 
-      if (TR::Compiler->om.shouldReplaceGuardedLoadWithSoftwareReadBarrier())
-         {
-         TR::TreeEvaluator::generateSoftwareReadBarrier(node, cg, resultReg, loadMemRef);
-         }
-      else
+      if (TR::Compiler->target.cpu.getSupportsGuardedStorageFacility())
          {
          TR::TreeEvaluator::checkAndSetMemRefDataSnippetRelocationType(node, cg, loadMemRef);
          TR::InstOpCode::Mnemonic loadOp = cg->comp()->useCompressedPointers() ? TR::InstOpCode::LLGFSG : TR::InstOpCode::LGG;
          generateRXInstruction(cg, loadOp, node, resultReg, loadMemRef);
+         }
+      else
+         {
+         TR::TreeEvaluator::generateSoftwareReadBarrier(node, cg, resultReg, loadMemRef);
          }
       node->setRegister(resultReg);
       }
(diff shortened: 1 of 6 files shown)


NigelYiboYu:
@amicic @fjeremic FYI, this is to fix concurrent scavenge parameter handling and query so that different z CPUs can operate in the right HW/SW modes.

runtime/compiler/z/codegen/J9TreeEvaluator.cpp
```diff
@@ -10126,7 +10126,8 @@ VMinlineCompareAndSwap(TR::Node *node, TR::CodeGenerator *cg, TR::InstOpCode::Mn
    if (TR::Compiler->om.readBarrierType() != gc_modron_readbar_none && isObj)
       {
       TR::Register* tempReadBarrier = cg->allocateRegister();
-      if (TR::Compiler->om.shouldReplaceGuardedLoadWithSoftwareReadBarrier())
+      if (TR::Compiler->om.shouldReplaceGuardedLoadWithSoftwareReadBarrier()
+              || !TR::Compiler->target.cpu.getS390SupportsGuardedStorageFacility())
```
fjeremic: This is not the right way to solve this IMO. `shouldReplaceGuardedLoadWithSoftwareReadBarrier` should return `true` if `getS390SupportsGuardedStorageFacility` returns `false`. In fact why does this API even exist? I see that we don't use it on x86 or Power. Shouldn't we just deprecate it?

Is the reason it exists is because we want to support being able to run software CS even on z14+? Thinking about this again can't we simply use `-Xjit:disableZ14` to achieve the same thing?

NigelYiboYu: > Is the reason it exists is because we want to support being able to run software CS even on z14+? 

Yes. This is for z14+. 

`shouldReplaceGuardedLoadWithSoftwareReadBarrier()` is true whenever `-XXgc:softwareRangeCheckReadBarrirer` is set on z. (This is not true for x86/PPC).

Removing this query renders the `-XXgc:softwareRangeCheckReadBarrirer` option useless. Now the question becomes do we want to keep an option to turn HW CS into SW CS on z14+.

fjeremic: > Removing this query renders the `-XXgc:softwareRangeCheckReadBarrirer` option useless. Now the question becomes do we want to keep an option to turn HW CS into SW CS on z14+.

Can't this be done via `-Xjit:disableZ14`? If so I fail to see the need for `-XXgc:softwareRangeCheckReadBarrirer`

amicic: I personally would like to maintain ability to enforce s/w CS on h/w supported platforms.

Does -Xjit:disableZ14 already exist?

GC side needs to know if we run s/w or h/w various. For example, there are no heap alignment requirements for s/w variant and we do not need to call j9gs calls.

Since heap is created very early, I suspect at that point of time we would not know if -Xjit:disableZ14 is specified or not, which is an argument against driving it with an JIT option.

amicic: We could lift the 'condition' that softwareRangeCheckReadBarrirer  option controls whether heap is aligned for not (for GS perposes) and just align it in s/w variant too. Then, we could probably based it on a JIT flag/option and remove GC option.

Other minor arguments for having GC option are, that's self documenting, and could more generically apply in a case if P ever introduces h/w support.

amicic: I'm guessing -Xjit:disableZ14 would disable more things than just guarded storage, which makes s/w vs h/w CS comparison less meaningful.

fjeremic: Sorry I just re-read what I posted. I meant to say do we need `shouldReplaceGuardedLoadWithSoftwareReadBarrier` to exist? We should keep `-XXgc:softwareRangeCheckReadBarrirer` as that is what forces software CS on the GC side. For the JIT however `-Xjit:disableZ14` is available to force the JIT side to use software read barriers even on z14 for example.

NigelYiboYu: `J9ObjectModel::shouldReplaceGuardedLoadWithSoftwareReadBarrier()` ultimately returns a value obtained form this GC query: `_shouldReplaceGuardedLoadWithSoftwareReadBarrier = mmf->j9gc_software_read_barrier_enabled(vm);`  which comes from `-XXgc:softwareRangeCheckReadBarrirer`

On z14+, the JIT needs to know if we should emit HW CS or SW-CS instructions. How can we do this without the GC query?

NigelYiboYu:
This commit passed our tests.

runtime/compiler/z/codegen/J9TreeEvaluator.cpp
fjeremic: The query does not make sense to me though. "Should replace guarded load with software read barrier". There is no replacement happening. We're also deviating from x86 and Power. They don't use this query. IMO I don't see the value of it.

Sure `-Xjit:disableZ14` does more. Are we able to disable just the GS facility? If not Does it not make more sense to have a `-Xjit:disableZ14GSFacility`. What the GC does vs. what the JIT does are two totally separate things.

fjeremic: By the way that query was originally introduced for the debugging peephole hacked up software CS which was deprecated as part of the recent work.

amicic: I dont really object, although I slighly prefer to control it with GC option/flag (would potentially apply to another platform if they end up having similar h/w support).

I presume -Xjit:disableZ14GSFacility (which I prefer over -Xjit:disableZ14) would make tje port library calls that we rely on in GC on whether GS is supported or not, would return 'not supported' on Z14+, with the option disabled (hence no need for any new api to introduce)?

amicic: All it all, I leave it up to you (JIT). If you end up not using j9gc_software_read_barrier_enabled() and we can rely on port lib calls whether GS is disabled or not, we (GC) will remove API and the option.

NigelYiboYu: How about renaming `shouldReplaceGuardedLoadWithSoftwareReadBarrier()` to `useSoftwareReadBarrier()` so that the name is more generic should other platforms get HW read barrier and is not just used by z14 GS?

> What the GC does vs. what the JIT does are two totally separate things.

True, GC and JIT are separated in terms of SW CS in terms of SW-CS handling. They don't have to be in sync. But from users/developers' point of view, would it make more sense to have just one single option to control both the GC and JIT behavior? When I set `-XXgc:softwareRangeCheckReadBarrier`, I expect GC and JIT to both perform SW-CS.

fjeremic: > When I set -XXgc:softwareRangeCheckReadBarrier, I expect GC and JIT to both perform SW-CS.

I don't disagree. My quarrel is with the use of `shouldReplaceGuardedLoadWithSoftwareReadBarrier` which is only used on Z, meanwhile x86 and Power do not use this, or whatever the newly envisioned `useSoftwareReadBarrier` API.

From the JIT side we can just as easily check if `-XXgc:softwareRangeCheckReadBarrier` is active (see tons of examples in J9Options.cpp) and disable GS facility at the compiler level. It will achieve the same effect and get rid of that aforementioned API.

fjeremic: I'm trying to limit the number of APIs we have to support. The less the better as it incurs additional cost maintaining them in the future, especially when we are starting to diverge between different architectures supporting the same feature.

fjeremic:
Pipeline test sanity.functional all JDK8

fjeremic:
Jenkins test sanity all JDK8

PR #5655, opened 2019-05-01
Fix java executable path in OnOutOfMemoryError test

sharon-wang:
Internal test revealed bug in `test -XX:OnOutOfMemoryError=` in `cmdLineTest_J9test_extended`, where the `java` command cannot be found when running the test.

This likely occurs if `JAVA_HOME`/`JAVA_BIN` isn't set on the test machine. By using the full path to the `java` command in the test command, we can avoid this failure.

### DIFF
--- a/test/functional/cmdLineTests/cmdLineTest_J9tests/j9tests.xml
+++ b/test/functional/cmdLineTests/cmdLineTest_J9tests/j9tests.xml
@@ -34,7 +34,6 @@ SPDX-License-Identifier: EPL-2.0 OR Apache-2.0 OR GPL-2.0 WITH Classpath-excepti
     <variable name="ONOUTOFMEMORYERROR_EQUALS" value="-XX:OnOutOfMemoryError="/>
     <variable name="ONOUTOFMEMORYERROR_JAR" value="-cp $Q$$JARPATH$$Q$ OnOutOfMemoryErrorTest"/>
     <variable name="JAVALANGOUTOFMEMORYERROR" value="java.lang.OutOfMemoryError:"/>
-    <variable name="JVMREQUESTINGTOOLDUMP" value="JVM Requesting Tool dump using 'java -version'"/>
 
     <test id="Verify Generate a javacore to STDOUT">
         <command>$EXE$ -Xdump:java:events=vmstart,file=/STDOUT/</command>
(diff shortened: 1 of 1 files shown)


test/functional/cmdLineTests/cmdLineTest_J9tests/j9tests.xml
```diff
@@ -67,9 +67,9 @@ SPDX-License-Identifier: EPL-2.0 OR Apache-2.0 OR GPL-2.0 WITH Classpath-excepti
     </test>
 
     <test id="test -XX:OnOutOfMemoryError=">
-        <command>$EXE$ $JVM_HEAP_LIMIT$ $ONOUTOFMEMORYERROR_EQUALS$"java -version" $ONOUTOFMEMORYERROR_JAR$</command>
+        <command>$EXE$ $JVM_HEAP_LIMIT$ $ONOUTOFMEMORYERROR_EQUALS$"$JAVACOMMAND$ -version" $ONOUTOFMEMORYERROR_JAR$</command>
```
JasonFengJ9: Can't `$EXE$` be used here instead of introducing another `$JAVACOMMAND$`?

sharon-wang: `$EXE$` contains `/path/to/java -Xcompressedrefs  -Xdump` and `$JAVACOMMAND$` contains only `/path/to/java`. Are there any disadvantages to using the options `-Xcompressedrefs  -Xdump` if they are not needed? I wasn't sure if it would be better to use the existing `$EXE$` or introducing `$JAVACOMMAND$` so that only `/path/to/java` is used.

JasonFengJ9: `-Xcompressedrefs` is from `$(JVM_OPTIONS)` and won't present for 32-bit `JVM`. `-Xdump` along with potential `-Xcompressedrefs` have no impact on '-version` output.
Using exiting `$EXE$` requires less change and can still verify that `-version` runs after an `OOM`.

sharon-wang: Thanks for the clarification! The test now uses the existing `$EXE$` - I have updated the `required` test output to match the appropriate regex.

JasonFengJ9:
@llxia pls review.

llxia:
Jenkins test extended xlinux,win jdk8,jdk11

PR #5657, opened 2019-05-02
Initialize per-compilation caches to NULL

dmitry-ten:
Caches in `CompilationInfoPerThreadRemote`
were not initialized to NULL.
Didn't seem to cause any problems, but should
still set them to NULL in the constructor.


dmitry-ten:
@ymanton could you review?

Issue #5658, opened 2019-05-03
 Add support for musl libc


bharathappali:
As OpenJ9 is built on top of OMR technology, I'm raising this issue to have OpenJ9 to work with musl libc as OMR also started to implement [musl support](https://github.com/eclipse/omr/issues/3774)

FYI - @dinogun @ashu-mehra @charliegracie

sxa:
@bharathappali @dinogun What's the current state of this? Will it build on Alpine? Is there a plan to e.g. make musl builds of OpenJ9 available?

avermeer:
Hello,

[JEP 386: Alpine Linux Port](https://openjdk.java.net/jeps/386) is part of Java 16 specifications, so I'm wondering why OpenJ9 Java 16 implementation seems to be missing that target.

OpenJ9's home site claim it to be a JVM with smaller footprint than HotSpot and with faster startup time, but missing official Alpine Linux support means that OpenJ9 usage in Docker world is limited to "big operating systems", unless you use some un-official OpenJ9 ports on Alpine, like https://gitlab.alpinelinux.org/alpine/aports/-/issues/12118  (but this site has up to Java 15 version of OpenJ9, Java 16 version is not available)

Are OpenJ9 developers not interested in Alpine Linux official support, leaving some other community to patch/fork to provide Alpine support?

Kind regards,
Alexandre

pshipton:
While the OpenJ9 (and OMR) projects are open to reviewing contributions that add support for musl/Alpine, I'm not aware of any recent efforts by committers or non-committers to make such contributions. Afaik there are no active plans for OpenJ9 to support JEP 386 in Java 16 or future releases. It appears there was an effort a couple years back to start porting OMR to musl but it was abandoned by the contributor, https://github.com/eclipse/omr/issues/3774.

If there was such an effort, the OpenJ9/OMR project would be looking for machine contributions in order to run builds and tests on the platform.

The links to builds in https://gitlab.alpinelinux.org/alpine/aports/-/issues/12118 appear just the regular builds which afaik don't include Alpine Linux. You can substitute 16 for 15 in the link and find the regular 16 builds as well.

I wouldn't say that Alpine Linux is part of the Java 16 specification, there is no requirement to support particular platforms in the Java specifications. Rather the OpenJDK project did a port to Alpine Linux and it was ready for Java 16.

[...]

PR #5659, opened 2019-05-03
Fix pthread_spinlock_t error in musl

bharathappali:
This PR is a part of the group of PR's that were/will be raised as a fix for issue #5658

In musl environment the OpenJ9 build fails with : 

```
../../../oti/j9port.h:292:9: error: 'pthread_spinlock_t' does not name a type; did you mean 'pthread_sigmask'?
 typedef pthread_spinlock_t spinlock_t;
         ^~~~~~~~~~~~~~~~~~
         pthread_sigmask
```
**Note:**
I have added the code in a view that `-DMUSL` is added to the complier flags.

**Build System Changes**
Flags to be added : `-DMUSL`

This item is still **Work In Progress** as only compile error are being resolved as of now, Will be ready to be reviewed after checking if no runtime errors are found as well after completing musl support

Signed-off-by: bharathappali <bharath.appali@gmail.com>

### DIFF
--- a/runtime/oti/j9port.h
+++ b/runtime/oti/j9port.h
@@ -43,6 +43,10 @@
 #include <unistd.h>
 #endif
 
+#if defined(LINUX) && !defined(J9ZTPF)
+#include <pthread.h>
+#endif /* defined(LINUX) && !defined(J9ZTPF) */
+
 /**
  * @name Port library access
  * @anchor PortAccess


ashu-mehra:
@bharathappali You don't really need to protect `#include <pthread.h>` with `OMR_MUSL_CLIB`. You can remove that as well and unconditionally include `pthread.h`.

dnakamura:
you won't want to unconditionally include `pthread.h`. Since this will break platforms where it doesnt exist/ we dont want to be using it (ie zos, and windows)

runtime/oti/j9port.h
```diff
@@ -43,6 +43,10 @@
 #include <unistd.h>
 #endif
 
+#if defined(OMR_MUSL_CLIB)
+#include <pthread.h>
```
DanHeidinga: Where are other platforms getting the definition of `typedef pthread_spinlock_t spinlock_t;` from?  I'd expect MUSL to pick up pthread in the same way other platforms are.

ashu-mehra:
> you won't want to unconditionally include pthread.h. Since this will break platforms where it doesnt exist/ we dont want to be using it (ie zos, and windows)

You are right, I shouldn't have specified _unconditionally_. I wanted to avoid using `OMR_MUSL_CLIB`. More appropriate suggestion would have beeen to include `pthread.h` for the platforms where it is available i.e. `defined(LINUX) && !defined(J9ZTPF)`.

runtime/oti/j9port.h
bharathappali: @DanHeidinga  for other platforms its getting from `sys/types.h` which includes `bits/pthreadtypes.h` which has `pthread_spinlock_t`. Any of the standard header file included by this file might have included `sys/types.h`.

In case of alpine `pthread_spinlock_t` is an optional definition which is only specified if defined `__NEED_pthread_spinlock_t`. `pthread_spinlock_t` is defined in `bits/alltypes.h` guarded with `__NEED_pthread_spinlock_t` macro

**code from musl - bits/alltypes.h:**
```
#if defined(__NEED_pthread_spinlock_t) && !defined(__DEFINED_pthread_spinlock_t)
typedef int pthread_spinlock_t;
#define __DEFINED_pthread_spinlock_t
#endif
```

`pthread.h` in MUSL defines `__NEED_pthread_spinlock_t` and includes `bits/alltypes.h` so we could get the `pthread_spinlock_t` definition.

**code from musl - pthread.h:**
```
#define __NEED_time_t
#define __NEED_clockid_t
#define __NEED_struct_timespec
#define __NEED_sigset_t
#define __NEED_pthread_t
#define __NEED_pthread_attr_t
#define __NEED_pthread_mutexattr_t
#define __NEED_pthread_condattr_t
#define __NEED_pthread_rwlockattr_t
#define __NEED_pthread_barrierattr_t
#define __NEED_pthread_mutex_t
#define __NEED_pthread_cond_t
#define __NEED_pthread_rwlock_t
#define __NEED_pthread_barrier_t
+-----------------------------------+
| #define __NEED_pthread_spinlock_t |
+-----------------------------------+
#define __NEED_pthread_key_t
#define __NEED_pthread_once_t
#define __NEED_size_t
+----------------------------+
| #include <bits/alltypes.h> |
+----------------------------+
```

DanHeidinga: Thanks for the clarification @bharathappali 

@ashu-mehra's suggestion to include pthread based on `defined(LINUX) && !defined(J9ZTPF)` is a more explicit change and is clear for all platforms.  I think that's the way we should proceed.

bharathappali: Sure @DanHeidinga Will update the PR accordingly

DanHeidinga:
Jenkins test sanity plinux,zlinux,win jdk8

bharathappali:
@DanHeidinga Can i please know if there are any changes need to be done ?

DanHeidinga:
Thanks for the ping on this.  The builds have passed so merging now.

PR #5667, opened 2019-05-03
Remove implementation dependencies from J9S390PrivateLinkage.hpp

0xdaryl:
* Remove CodeGenerator implementation dependence in J9S390PrivateLinkage.hpp
* Add forward declaration for S390JNICallDataSnippet


0xdaryl:
Jenkins test sanity zlinux,zos jdk11

0xdaryl:
@fjeremic , I'd appreciate a review and merge please. 

This came about as part of the work to break the CodeGenerator dependence in OMRLinkage.  I don't think `getSystemStackPointerRegister` is used in a particularly hot path, so outlining it probably won't have much of an impact on performance.  But correct me if I'm wrong.

PR #5668, opened 2019-05-03
Make findField accessible outside of VM

dmitry-ten:
Rename `findField` to `jitFindField` and
add it to `j9protos.h`, so that it can be accessed
from the compiler.
The reason is because a downstream project (JITaaS)
would benefit from being able to directly call this function.

Required for: #5634

### DIFF
--- a/runtime/jit_vm/ctsupport.c
+++ b/runtime/jit_vm/ctsupport.c
@@ -404,7 +404,7 @@ jitResetAllMethodsAtStartupHelper(J9VMThread *vmStruct, J9Class *root)
  * @return pointer cast to UDATA 
  */
 static UDATA
-findField(J9VMThread *vmStruct, J9ConstantPool *constantPool, UDATA index, BOOLEAN isStatic, J9Class **declaringClass)
+jitFindField(J9VMThread *vmStruct, J9ConstantPool *constantPool, UDATA index, BOOLEAN isStatic, J9Class **declaringClass)
 {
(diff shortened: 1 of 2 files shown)


DanHeidinga:
@dmitry-ten there are already several ways to do this kind of operation.  Is there a reason that the existing APIs aren't sufficient?

ie:
https://github.com/eclipse/openj9/blob/30c494cb566d9ba698e2d43753c23508157e7fac/runtime/oti/j9nonbuilder.h#L4405-L4406

https://github.com/eclipse/openj9/blob/30c494cb566d9ba698e2d43753c23508157e7fac/runtime/oti/j9nonbuilder.h#L4435-L4436

gacholio:
Also, I can't imagine any of the VM-supplied code would work in the JITaas environment.

dmitry-ten:
Hm, I think `resolveInstanceFieldRefInto` might work for my purposes.
The original goal was to reduce the number of remote calls for `TR_J9Server ::jitFieldsAreSame`. I want the client to do the same thing as `jitFieldsAreIdentical`, but instead of simply returning whether 2 fields are identical, return their declaring classes and `J9ROMFieldShape *`. 
So, client would retrieve this information and send it to the server. Server would cache it, and if 
`TR_J9Server::jitFieldsAreSame` needs to compare the same field again, it would access it from cache, instead of making a remote call to the client.

gacholio:
Can this be closed?

PR #5670, opened 2019-05-03
Remove all references to `installJitBytecodes()`

DanHeidinga:
installJitBytecodes() is left over from the old builder based
interpreter using a bytecode table rather than the current
switch / computed goto based interpreter.

Removing all references to the function saves an indirect
cross-dll call from startup.


pshipton:
jenkins test sanity win jdk8

PR #5671, opened 2019-05-03
Combine `ResolvedMethod_isUnresolvedString` message into `ResolvedMethod_stringConstant`

### LINKED ISSUE #5649: Reduce number of ResolvedMethod_isUnresolvedString messages
Most of the `ResolvedMethod_isUnresolvedString` messages come from the following piece of code
```
J9::SymbolReferenceTable::findOrCreateStringSymbol(TR::ResolvedMethodSymbol * owningMethodSymbol, int32_t cpIndex)
   {
   TR_ResolvedMethod * owningMethod = owningMethodSymbol->getResolvedMethod();
   void * stringConst = owningMethod->stringConstant(cpIndex);
   TR::SymbolReference * symRef;
   if (owningMethod->isUnresolvedString(cpIndex))
```
As seen above, before calling `isUnresolvedString` we call `stringConstant` which requires another message. We could create a combined frontend function that returns the answer to both queries, so that we need one message instead of two.

a7ehuo:
When sending `stringConstant` info, the client will also send
`isUnresolvedString` info for both `optimizeForAOT` and
not `optimizeForAOT cases.
The server caches the values in a map using cpIndex as the key.

Fixes #5649

### DIFF
--- a/runtime/compiler/control/JITaaSCompilationThread.cpp
+++ b/runtime/compiler/control/JITaaSCompilationThread.cpp
@@ -1560,6 +1560,7 @@ bool handleServerMessage(JITaaS::J9ClientStream *client, TR_J9VM *fe)
          break;
       case J9ServerMessageType::ResolvedMethod_isUnresolvedString:
          {
+         TR_ASSERT(false, "ResolvedMethod_isUnresolvedString is combined into ResolvedMethod_stringConstant");
          auto recv = client->getRecvData<TR_ResolvedJ9Method *, int32_t, bool>();
          auto mirror = std::get<0>(recv);
          auto cpIndex = std::get<1>(recv);
@@ -1572,7 +1573,9 @@ bool handleServerMessage(JITaaS::J9ClientStream *client, TR_J9VM *fe)
          auto recv = client->getRecvData<TR_ResolvedJ9Method *, int32_t>();
          auto mirror = std::get<0>(recv);
          auto cpIndex = std::get<1>(recv);
-         client->write(mirror->stringConstant(cpIndex));
+         client->write(mirror->stringConstant(cpIndex),
+                       mirror->isUnresolvedString(cpIndex, true),
+                       mirror->isUnresolvedString(cpIndex, false));
          }
          break;
       case J9ServerMessageType::ResolvedRelocatableMethod_createResolvedRelocatableJ9Method:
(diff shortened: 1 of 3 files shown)


a7ehuo:
pipeline test sanity.functional+jitaas xlinux jdk8

ymanton:
pipeline test sanity.functional+jitaas xlinux jdk8

AdamBrousseau:
pipeline test sanity.functional+jitaas xlinux jdk8 depends eclipse/openj9-omr#jitaas

runtime/compiler/control/JITaaSCompilationThread.cpp
```diff
@@ -1560,6 +1560,7 @@ bool handleServerMessage(JITaaS::J9ClientStream *client, TR_J9VM *fe)
          break;
       case J9ServerMessageType::ResolvedMethod_isUnresolvedString:
          {
+         TR_ASSERT(false, "ResolvedMethod_isUnresolvedString is combined into ResolvedMethod_stringConstant");
```
ymanton: Can we simply remove the message or are you concerned that we can still reach this line and the corresponding one in `TR_ResolvedJ9JITaaSServerMethod::isUnresolvedString()`, hence the asserts? (If so, an issue tagged with beginner to clean it up later might be a good way to take care of it. :grin:)

runtime/compiler/env/j9methodServer.hpp
```diff
@@ -305,6 +305,14 @@ class TR_ResolvedJ9JITaaSServerMethod : public TR_ResolvedJ9Method
    UnorderedMap<TR_ResolvedMethodKey, TR_ResolvedMethodCacheEntry> _resolvedMethodsCache;
    TR_IPMethodHashTableEntry *_iProfilerMethodEntry;
 
+   struct TR_IsUnresolvedString
+      {
+      TR_IsUnresolvedString(bool optForAOT, bool notOptForAOT): _optForAOT(optForAOT), _notOptForAOT(notOptForAOT) {}
+      bool _optForAOT;
+      bool _notOptForAOT;
```
ymanton: Can you change these to something like `_optimizeForAOTTrueResult` and `_optimizeForAOTFalseResult`? Having "optimizeForAOT" as part of the name helps readers relate it to the original parameter, find it when searching, etc.

runtime/compiler/control/JITaaSCompilationThread.cpp
a7ehuo: At the moment, `isUnresolvedString()` is always called right after `stringConstant()` in `J9::SymbolReferenceTable::findOrCreateStringSymbol()`. The concern is that when these two methods are decoupled in the non-jitaas code in the future for any reason, and if the `ResolvedMethod_isUnresolvedString` message is removed, it will break the code. It'll be hard to track the bug down. I will open a beginner issue to evaluate if this message should be removed or at least the related assert should be removed.

PR #5672, opened 2019-05-03
Fix JNI check to properly handle atomic-free

gacholio:
When atomic-free was introduced, JNI check was modified to enter/exit
the VM instead of acquire/release VM access. This is incorrect in some
cases (particularly, the method exit hook which is called from inside
the VM with VM access).

The end result is that JNI check might incorrectly mark a thread as
being in a native when it is in fact running in the VM, resulting in
assertions or crashes.

Fix JNI check to differentiate between being called from JNI code, or
called internally from the VM.

[ci skip]

### DIFF
--- a/runtime/jnichk/jnichk_internal.h
+++ b/runtime/jnichk/jnichk_internal.h
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 1991, 2018 IBM Corp. and others
+ * Copyright (c) 1991, 2019 IBM Corp. and others
  *
  * This program and the accompanying materials are made available under
  * the terms of the Eclipse Public License 2.0 which accompanies this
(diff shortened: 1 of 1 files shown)


gacholio:
jenkins test all xlinux jdk8

gacholio:
I'll unWIP this once the originally-reported issue has been verified by @manqingl

gacholio:
@DanHeidinga Issue confirmed fixed, ready for review.

runtime/jnichk/jnichk_internal.h
```diff
@@ -102,9 +102,27 @@ typedef struct JNICHK_GREF_HASHENTRY {
...
+		} else { \
+			if (hasNoVMAccess) { \
+				acquireVMAccess(currentThread); \
+			} \
+		} \
+	} while(0)
+		
+#define exitVM(currentThread) \
+	do { \
+		if (inNative) { \
```
DanHeidinga: Will `inNative` always be in-scope?  I'm assuming enterVM/exitVM are always used as a pair?

gacholio: Yes, the assumption is already there wrt the VM access variable.

DanHeidinga:

DanHeidinga:
I see Sanity and Extended had already passed.  Merging without waiting for (unnecessary) build I kicked

Issue #5503, opened 2019-04-15
JITaaS client performance should not drop when server crashes


mpirvu:
If the JITaaS server happens to crash, the client still has the option to perform JIT compilations locally and (provided there is enough memory) there should be no observable drop in throughput.

harryyu1994:
Taking this work on the side

harryyu1994:
First plan I had was to put the client-server connection establish process inside `preCompilationTask` since that is where the decision of whether to do remote compilation or local compilation is done. If the decision is to do a remote compilation but the connection was failed due to server not available, we can change the decision to do a local compilation. We would do this check for every single compilation in case the server become available again. I relied on the connection throwing exceptions when it fails, but it turns out that exceptions are too expensive when the server doesn't come back on for a long time.

harryyu1994:
Second plan is to use this check instead of relying on the `open_connection()` call to throw/catch `StreamFailure` exceptions.

PR #5673, opened 2019-05-03
Allow client to compile locally when server is down

harryyu1994:
Client pings the server every 10 seconds, when server is available, client does remote compilations. When server is unavailable, client does local compilations. (When connection fails the first time, client pings after 1 second, just in case that the connection failure was not caused by server unavailability).

Issue: #5503
[skip ci]

### DIFF
--- a/runtime/compiler/control/CompilationThread.cpp
+++ b/runtime/compiler/control/CompilationThread.cpp
@@ -3418,6 +3418,7 @@ void TR::CompilationInfo::stopCompilationThreads()
          }
       catch (const JITServer::StreamFailure &e)
          {
+         JITaaSHelpers::postStreamFailure(OMRPORT_FROM_J9PORT(_jitConfig->javaVM->portLibrary));
          // catch the stream failure exception if the server dies before the dummy message is send for termination.
          if (TR::Options::getVerboseOption(TR_VerboseJITaaS))
             TR_VerboseLog::writeLineLocked(TR_Vlog_JITaaS, "JITaaS StreamFailure (server unreachable before the termination message was sent): %s", e.what());
@@ -6775,7 +6776,8 @@ TR::CompilationInfoPerThreadBase::shouldPerformLocalComp(const TR_MethodToBeComp
    //
    if (entry->_optimizationPlan->getOptLevel() <= cold &&
       (TR::Options::getCmdLineOptions()->getOption(TR_EnableJITaaSHeuristics) || localColdCompilations) || 
-      !JITServer::ClientStream::isServerCompatible(OMRPORT_FROM_J9PORT(_jitConfig->javaVM->portLibrary)))
+      !JITServer::ClientStream::isServerCompatible(OMRPORT_FROM_J9PORT(_jitConfig->javaVM->portLibrary)) ||
+      (!JITaaSHelpers::isServerAvailable() && !JITaaSHelpers::shouldRetryConnection(OMRPORT_FROM_J9PORT(_jitConfig->javaVM->portLibrary))))
       doLocalComp = true;
 
    return doLocalComp;
--- a/runtime/compiler/rpc/J9Client.cpp
+++ b/runtime/compiler/rpc/J9Client.cpp
@@ -41,9 +41,10 @@
 
 namespace JITServer
 {
-
 int ClientStream::_numConnectionsOpened = 0;
 int ClientStream::_numConnectionsClosed = 0;
+
+// used for checking server compatibility
 int ClientStream::_incompatibilityCount = 0;
 uint64_t ClientStream::_incompatibleStartTime = 0;
 const uint64_t ClientStream::RETRY_COMPATIBILITY_INTERVAL = 10000; //ms
--- a/runtime/compiler/control/JITaaSCompilationThread.cpp
+++ b/runtime/compiler/control/JITaaSCompilationThread.cpp
@@ -48,6 +48,11 @@ extern TR::Monitor *assumptionTableMutex;
 
 uint32_t serverMsgTypeCount[JITServer::MessageType_ARRAYSIZE] = {};
 
+uint64_t JITaaSHelpers::_waitTime = 1000;
+bool JITaaSHelpers::_serverAvailable = true;
+uint64_t JITaaSHelpers::_nextConnectionRetryTime = 0;
+TR::Monitor * JITaaSHelpers::_clientStreamMonitor = NULL;
+
 // TODO: this method is copied from runtime/jit_vm/ctsupport.c,
 // in the future it's probably better to make that method publicly accessible
 static UDATA
@@ -2476,10 +2481,23 @@ remoteCompile(
       {
       try 
          {
-         client = new (PERSISTENT_NEW) JITServer::ClientStream(compInfo->getPersistentInfo());
+         if (JITaaSHelpers::isServerAvailable())
+            {
+            client = new (PERSISTENT_NEW) JITServer::ClientStream(compInfo->getPersistentInfo());
+            }
+         else if (JITaaSHelpers::shouldRetryConnection(OMRPORT_FROM_J9PORT(compInfoPT->getJitConfig()->javaVM->portLibrary)))
+            {
+            client = new (PERSISTENT_NEW) JITServer::ClientStream(compInfo->getPersistentInfo());
+            JITaaSHelpers::postStreamConnectionSuccess();
+            }
+         else
+            {
+            compiler->failCompilation<JITServer::StreamFailure>("Server is not available, should retry with local compilation.");
+            }
          }
       catch (const JITServer::StreamFailure &e)
          {
+         JITaaSHelpers::postStreamFailure(OMRPORT_FROM_J9PORT(compInfoPT->getJitConfig()->javaVM->portLibrary));
          compiler->failCompilation<JITServer::StreamFailure>(e.what());
          }
       catch (const std::bad_alloc &e)
@@ -2494,11 +2512,25 @@ remoteCompile(
          {
          try
             {
-            client = new (PERSISTENT_NEW) JITServer::ClientStream(compInfo->getPersistentInfo());
-            compInfoPT->setClientStream(client);
+            if (JITaaSHelpers::isServerAvailable())
+               {
+               client = new (PERSISTENT_NEW) JITServer::ClientStream(compInfo->getPersistentInfo());
+               compInfoPT->setClientStream(client);
+               }
+            else if (JITaaSHelpers::shouldRetryConnection(OMRPORT_FROM_J9PORT(compInfoPT->getJitConfig()->javaVM->portLibrary)))
+               {
+               client = new (PERSISTENT_NEW) JITServer::ClientStream(compInfo->getPersistentInfo());
+               compInfoPT->setClientStream(client);
+               JITaaSHelpers::postStreamConnectionSuccess();
+               }
+            else
+               {
+               compiler->failCompilation<JITServer::StreamFailure>("Server is not available, should retry with local compilation.");
+               }
             }
          catch (const JITServer::StreamFailure &e)
             {
+            JITaaSHelpers::postStreamFailure(OMRPORT_FROM_J9PORT(compInfoPT->getJitConfig()->javaVM->portLibrary));
             compiler->failCompilation<JITServer::StreamFailure>(e.what());
             }
          catch (const std::bad_alloc &e)
@@ -2597,6 +2629,8 @@ remoteCompile(
       }
    catch (const JITServer::StreamFailure &e)
       {
+      JITaaSHelpers::postStreamFailure(OMRPORT_FROM_J9PORT(compInfoPT->getJitConfig()->javaVM->portLibrary));
+
       client->~ClientStream();
       TR_Memory::jitPersistentFree(client);
       compInfoPT->setClientStream(NULL);
@@ -3888,6 +3922,34 @@ JITaaSHelpers::getROMClassData(const ClientSessionData::ClassInfo &classInfo, Cl
       }
    }
 
+void
+JITaaSHelpers::postStreamFailure(OMRPortLibrary *portLibrary)
+   {
+   OMR::CriticalSection handleStreamFailure(getClientStreamMonitor());
+
+   OMRPORT_ACCESS_FROM_OMRPORT(portLibrary);
+   uint64_t current_time = omrtime_current_time_millis();
+   if (current_time >= _nextConnectionRetryTime)
+      {
+      _waitTime *= 2; // exponential backoff
+      }
+   _nextConnectionRetryTime = current_time + _waitTime;
+   _serverAvailable = false;
+   }
+
+void
+JITaaSHelpers::postStreamConnectionSuccess()
+   {
+   _serverAvailable = true;
+   _waitTime = 1000; // ms
+   }
+
+bool
+JITaaSHelpers::shouldRetryConnection(OMRPortLibrary *portLibrary)
+   {
+   OMRPORT_ACCESS_FROM_OMRPORT(portLibrary);
+   return omrtime_current_time_millis() > _nextConnectionRetryTime;
+   }
 
 // insertIntoOOSequenceEntryList needs to be executed with sequencingMonitor in hand
 void 
--- a/runtime/compiler/control/JITaaSCompilationThread.hpp
+++ b/runtime/compiler/control/JITaaSCompilationThread.hpp
@@ -396,57 +396,81 @@ class CompilationInfoPerThreadRemote : public TR::CompilationInfoPerThread
 class JITaaSHelpers
    {
    public:
-   enum ClassInfoDataType
-      {
-      CLASSINFO_ROMCLASS_MODIFIERS,
-      CLASSINFO_ROMCLASS_EXTRAMODIFIERS,
-      CLASSINFO_BASE_COMPONENT_CLASS,
-      CLASSINFO_NUMBER_DIMENSIONS,
-      CLASSINFO_PARENT_CLASS,
-      CLASSINFO_INTERFACE_CLASS,
-      CLASSINFO_CLASS_HAS_FINAL_FIELDS,
-      CLASSINFO_CLASS_DEPTH_AND_FLAGS,
-      CLASSINFO_CLASS_INITIALIZED,
-      CLASSINFO_BYTE_OFFSET_TO_LOCKWORD,
-      CLASSINFO_LEAF_COMPONENT_CLASS,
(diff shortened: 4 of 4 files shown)


runtime/compiler/control/CompilationThread.cpp
```diff
@@ -7036,7 +7077,48 @@ TR::CompilationInfoPerThreadBase::preCompilationTasks(J9VMThread * vmThread,
...
+               else
+                  {
+                  JITaaS::J9ClientStream::_numLocalCompilations++;
+                  }
+               }
+            else
+               {
+               entry->setRemoteCompReq();
+               }
+            }
```
ymanton: This chunk of code looks identical to the chunk above, is one for AOT and the other for JIT? Can you try to encapsulate this into a function rather than repeating it? It will save us headaches maintaining it in the future.

runtime/compiler/control/CompilationThread.cpp
```diff
@@ -6977,7 +6979,46 @@ TR::CompilationInfoPerThreadBase::preCompilationTasks(J9VMThread * vmThread,
       if ((_compInfo.getPersistentInfo()->getJITaaSMode() == CLIENT_MODE) &&
          !shouldPerformLocalComp(entry))
          {
-         entry->setRemoteCompReq();
+         JITaaS::J9ClientStream *client = getClientStream();
+         if (!client)
+            {
+            if (JITaaS::J9ClientStream::_serverReachable || JITaaS::J9ClientStream::_numLocalCompilations >= 50)
```
ymanton: Using the number of compilations as a counter/timer is convenient, but it might cause a lot of variation and fluctuation in the long run because "50 compilations" might take 1 second or 100 seconds, depending on a lot of variables. Is it possible to use wall clock time instead? End users who need to tweak this sort of thing will probably have an easier time with wall clocks than the number of compilations.

harryyu1994: yea i agree, I will switch to a timer

harryyu1994:
Tested (including bring up/down the server multiple times to see if throughput suffers, working as expected). Ready for a review @ymanton

ymanton:
jenkins test sanity.functional+jitaas xlinux jdk8 depends eclipse/openj9-omr#jitaas

harryyu1994:
```
16:06:26  /home/jenkins/workspace/Build_JDK8_x86-64_linux_Personal/build/linux-x86_64-normal-server-release/vm/compiler/../compiler/rpc/J9Server.cpp:111:8: error: ‘SSL_CTX_set_ecdh_auto’ was not declared in this scope
16:06:26      if (SSL_CTX_set_ecdh_auto(ctx, 1) != 1)
16:06:26          ^~~~~~~~~~~~~~~~~~~~~
```
Looks like this API no longer exists in the newer openssl versions, raising an issue

runtime/compiler/rpc/J9Client.cpp
```diff
@@ -270,4 +273,44 @@ J9ClientStream::shutdown()
...
+   // check server availability 1 second after the first StreamFailure
+   // if it fails again, the _pingServerInterval will be set to 10 seconds
+   // meaning the server will be pinged every 10 seconds to see if it's back up again
+   // without the 1 second delay after the first StreamFailure would cause the performance of client
+   // to suffer as client could fire off a lot of remote compilations that fails
+   PORT_ACCESS_FROM_JITCONFIG(jitConfig);
+   uint64_t current_time = j9time_current_time_millis();
+   if ((current_time - _localCompStartTime) > _pingServerInterval)
+      {
+      if (gethostbyname(info->getJITaaSServerAddress().c_str()))
```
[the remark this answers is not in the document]
mpirvu: Is gethostbyname actually sending a ping?

runtime/compiler/rpc/J9Client.cpp
```diff
@@ -270,4 +273,44 @@ J9ClientStream::shutdown()
...
+   if (!_serverUnreachable)
+      return true;
+
+   // check server availability 1 second after the first StreamFailure
+   // if it fails again, the _pingServerInterval will be set to 10 seconds
+   // meaning the server will be pinged every 10 seconds to see if it's back up again
+   // without the 1 second delay after the first StreamFailure would cause the performance of client
+   // to suffer as client could fire off a lot of remote compilations that fails
+   PORT_ACCESS_FROM_JITCONFIG(jitConfig);
+   uint64_t current_time = j9time_current_time_millis();
```
mpirvu: I would prefer to use some time routine that exists in the OMR level just in case OMR team want the communication library moved over to them

harryyu1994:
I couldn't find any cases where an alive and well server tries to close down a connection, it's usually the clients that close the connections. If we can find cases where server closes connections, one solution would be to make the server inform the client it's closing down a connection. That way we can distinguish an unavailable server from a server that tries to kill a connection. Also I have the 1-second first attempt mechanism in place just in case the server is only unavailable for a split second.

harryyu1994:
Okay so when https://github.com/eclipse/openj9/pull/6045 the version check goes in, the server would reject wrong/ill-formed client messages and close the connection.

mpirvu:
> I guess for wrong/ill-formed client messages, server can send a special message informing the client before closing down the connection.
> Make StreamFailure for unexpected disconnection only.

yes

An alternative: a client that tries to connect to a dead server would fail when the connection is opened. Only at that point we would mark the server unreachable. So, let the client process the StreamFailure which would break the existing connection, then let it try to establish the connection again and if it fails declare server unreachable.

runtime/compiler/rpc/J9Client.cpp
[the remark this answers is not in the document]
harryyu1994: `gethostbyname` is not actually sending a ping, it just queries to see if the hostname exists. I think I need to implement an actual ping routine.

harryyu1994:
We might have a problem with the alternative approach. I have experimented with running AcmeAir where the client reconnect immediately after StreamFailure, and it caused the client to "slow down"/"freeze" for well over 10 seconds when I kill the server. I suspect the reason is that between the time of the first StreamFailure and the reconnection failure, the client has already fired off a bunch of remote compilation requests, which ends up failing again and causes the "freeze". (This is also why I have the reattempt-after-one-second mechanism, we need to make sure we don't reattempt a lot of remote compilation requests if the server is dead, they will slow down the client significantly.)

This problem goes aways when you mark server unreachable immediately after StreamFailure. The new compilations will immediately switch to local compilations. So in my opinion, server informing client about disconnection and making StreamFailure for unexpected disconnection only would be the better approach.

mpirvu:
Some sequence of messages at the server while running sanity:
```
#JITaaS: compThreadID=2 found clientSessionData=00007EFE401E1430 for clientUID=16546041498833903697 seqNo=5644
#JITaaS: Server will process a list of 0 unloaded classes for clientUID 16546041498833903697
#JITaaS: compThreadID=2 has successfully compiled sun/util/resources/TimeZoneNames.getContents()[[Ljava/lang/Object;
#JITaaS: Stream failed while compThreadID=2 was reading the compilation request: JITaaS I/O error: reading message size
#JITaaS: compThreadID=2 did an early abort
#JITaaS: Server received request for stream 00007EFE401E1260
#JITaaS: compThreadID=2 found clientSessionData=00007EFE401E1430 for clientUID=16546041498833903697 seqNo=5645
#JITaaS: Server will process a list of 0 unloaded classes for clientUID 16546041498833903697
#JITaaS: compThreadID=2 has successfully compiled java/util/ResourceBundle.getStringArray(Ljava/lang/String;)[Ljava/lang/String;
```

I don't know why we get that "JITaaS I/O error", but I am guessing that the connection is broken at that point while the client is able to continue. I have yet to determine what happens at the client. It's possible that the client experiences a hiccup; it's possible that the client compiled that method locally.

All in all I agree that each party informing the other about its intentions/status is more resilient.

mpirvu:
Sometimes compilations fail in succession:
```
#JITaaS: compThreadID=3 has successfully compiled java/lang/StringBuilder.append(Ljava/lang/String;)Ljava/lang/StringBuilder;
#JITaaS: compThreadID=1 has successfully compiled java/math/BitLevel.shiftLeftOneBit([I[II)V
#JITaaS: Stream failed while compThreadID=1 was reading the compilation request: JITaaS I/O error: reading message size
#JITaaS: compThreadID=1 did an early abort
#JITaaS: compThreadID=3 found clientSessionData=00007EFE401E1430 for clientUID=14578906776852225384 seqNo=133
#JITaaS: Server will process a list of 0 unloaded classes for clientUID 14578906776852225384
#JITaaS: compThreadID=2 has successfully compiled java/math/Multiplication.multPAP([I[I[III)V
#JITaaS: Stream failed while compThreadID=2 was reading the compilation request: JITaaS I/O error: reading message size
#JITaaS: compThreadID=2 did an early abort
#JITaaS: compThreadID=0 has successfully compiled java/math/Multiplication.square([II[I)[I
#JITaaS: Stream failed while compThreadID=0 was reading the compilation request: JITaaS I/O error: reading message size
#JITaaS: compThreadID=0 did an early abort
#JITaaS: compThreadID=3 has successfully compiled com/ibm/security/bootstrap/DigestBase.b2iBig([BI[III)V
```
compThread1, 2 and 0 all experience an error while reading the size.

runtime/compiler/rpc/J9Client.cpp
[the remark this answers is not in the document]
harryyu1994: https://github.com/eclipse/openj9/issues/6120
I opened an issue, going to change the use of `gethostbyname` to `getaddrinfo` for client openConnection soon, in a separate PR.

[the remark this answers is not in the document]
harryyu1994: A better solution, instead of pinging the server, we can try re-establishing the connection right there, we will get the same effect if it fails and the connection will be conveniently re-established if it goes through.

mpirvu:
I doubt that the client is exiting because I see many compilations from the same client coming afterwards
```
#JITaaS: Stream failed while compThreadID=0 was reading the compilation request: JITaaS I/O error: reading message size
#JITaaS: compThreadID=0 did an early abort
#JITaaS: compThreadID=3 has successfully compiled com/ibm/security/bootstrap/DigestBase.b2iBig([BI[III)V
#JITaaS: compThreadID=3 found clientSessionData=00007EFE401E1430 for clientUID=14578906776852225384 seqNo=134
#JITaaS: Server will process a list of 0 unloaded classes for clientUID 14578906776852225384
#JITaaS: compThreadID=3 has successfully compiled java/lang/String.<init>([CIIZ)V
#JITaaS: compThreadID=3 found clientSessionData=00007EFE401E1430 for clientUID=14578906776852225384 seqNo=135
#JITaaS: Server will process a list of 0 unloaded classes for clientUID 14578906776852225384
#JITaaS: compThreadID=3 has successfully compiled com/ibm/security/bootstrap/SHA.implCompress([BI)V
#JITaaS: compThreadID=3 found clientSessionData=00007EFE401E1430 for clientUID=14578906776852225384 seqNo=136
#JITaaS: Server will process a list of 0 unloaded classes for clientUID 14578906776852225384
#JITaaS: compThreadID=3 has successfully compiled sun/misc/URLClassPath$JarLoader.getResource(Ljava/lang/String;Z)Lsun/misc/Resource;
#JITaaS: compThreadID=3 found clientSessionData=00007EFE401E1430 for clientUID=14578906776852225384 seqNo=137
#JITaaS: Server will process a list of 0 unloaded classes for clientUID 14578906776852225384
#JITaaS: compThreadID=3 has successfully compiled java/io/UnixFileSystem.checkAccess(Ljava/io/File;I)Z
#JITaaS: compThreadID=3 found clientSessionData=00007EFE401E1430 for clientUID=14578906776852225384 seqNo=138
#JITaaS: Server will process a list of 0 unloaded classes for clientUID 14578906776852225384
#JITaaS: compThreadID=3 has successfully compiled java/lang/ClassLoader.getClassLoadingLock(Ljava/lang/String;)Ljava/lang/Object;
#JITaaS: Server received request for stream 00007EFD6E174998
#JITaaS: Server received request for stream 00007EFD6D938168
#JITaaS: compThreadID=1 found clientSessionData=00007EFE401E1430 for clientUID=14578906776852225384 seqNo=139
#JITaaS: Server will process a list of 0 unloaded classes for clientUID 14578906776852225384
#JITaaS: compThreadID=0 found clientSessionData=00007EFE401E1430 for clientUID=14578906776852225384 seqNo=140
```

harryyu1994:
Revisiting this PR. Trying to figure out why we get "JITaaS I/O error". Let's see if we still get these error messages now that server and client both do not close connections without notifying each other.

mpirvu:
> Trying to figure out why we get "JITaaS I/O error"

@dmitry-ten has an explanation here: https://github.com/eclipse/openj9/issues/6133

harryyu1994:
> Just figured out the true reason for these messages: this error appears whenever client destroys a stream that is still connected to a server stream. This happens either when all of the entries in a compilation queue get processed, or when JVM exits. Which accounts for all of the occurrences of this message.

If this is the reason, then it means we won't get the case where server is available but it just randomly throws StreamFailure and terminates the connection. StreamFailure exception at client is only triggered when server is unavailable.

harryyu1994:
I am going to run an experiment where we look at how many streamfailure exception we get at client when the server is fully available. If we don't get any, then we can rely on streamfailure exception at client for the availability check.

harryyu1994:
@mpirvu This one is ready for review.

harryyu1994:
Used AcmeAir benchmark to test, killed the server a few times.
Working as expected and no throughput drops.

runtime/compiler/control/CompilationThread.cpp
```diff
@@ -6953,6 +6955,12 @@ TR::CompilationInfoPerThreadBase::preCompilationTasks(J9VMThread * vmThread,
          }
       }
 
+   if (_compInfo.getPersistentInfo()->getJITaaSMode() == CLIENT_MODE && !JITaaS::J9ClientStream::isServerAvailable())
+      {
+      TR_ASSERT(getClientStream() == NULL, "We should not have an open connection when server is unavailable.");
+      setClientStream(JITaaS::J9ClientStream::retryConnectionToServer(OMRPORT_FROM_J9PORT(_jitConfig->javaVM->portLibrary), _compInfo.getPersistentInfo()));
```
mpirvu: One disadvantage of this approach is that the client will connect right here even though it may not intend to send a compilation request to the server (maybe because the compilation is cheap and heuristics decide to do it locally).
Also, I would place the check about `JITaaS::J9ClientStream::isServerAvailable()` inside `retryConnectionToServer` (including the assert).

runtime/compiler/rpc/J9Client.cpp
```diff
@@ -48,6 +49,9 @@ int J9ClientStream::_incompatibilityCount = 0;
 uint64_t J9ClientStream::_incompatibleStartTime = 0;
 const uint64_t J9ClientStream::RETRY_COMPATIBILITY_INTERVAL = 10000; //ms
 const int J9ClientStream::INCOMPATIBILITY_COUNT_LIMIT = 5;
+uint64_t J9ClientStream::_unavailableStartTime = 0;
+bool J9ClientStream::_serverAvailable = true;
+uint64_t J9ClientStream::RETRY_AVAILABILITY_INTERVAL = 1000; // ms
```
mpirvu: Since this is not a constant we should not use capitals for its name

runtime/compiler/rpc/J9Client.cpp
```diff
@@ -284,4 +288,31 @@ J9ClientStream::shutdown()
...
+J9ClientStream * 
+J9ClientStream::retryConnectionToServer(OMRPortLibrary *portLibrary, TR::PersistentInfo *info)
+   {
+   // retry connection to server 1 second after the first StreamFailure
+   // if it fails again, RETRY_AVAILABILITY_INTERVAL will be set to 10 seconds
+   // return a J9ClientStream if connection is successful and return NULL if it fails or if it's not time to retry yet
+   OMRPORT_ACCESS_FROM_OMRPORT(portLibrary);
+
+   J9ClientStream *client = NULL;
+   uint64_t current_time = omrtime_current_time_millis();
```
mpirvu: I would place a check for `_serverAvailable` at the beginning of this function.

runtime/compiler/rpc/J9Client.hpp
```diff
@@ -138,17 +138,39 @@ class J9ClientStream : J9Stream
       return _incompatibilityCount < INCOMPATIBILITY_COUNT_LIMIT;
       }
 
+   static void setServerUnavailable(OMRPortLibrary *portLibrary)
+      {
+      OMRPORT_ACCESS_FROM_OMRPORT(portLibrary);
+      if (_serverAvailable)
+         {
+         _serverAvailable = false;
+         _unavailableStartTime = omrtime_current_time_millis();
```
mpirvu: Let's set `_unavailableStartTime` every time we call this function. Basically, we should be looking at the most recent event when we found the server unavailable and start the clock from that moment.

runtime/compiler/control/CompilationThread.cpp
mpirvu: It occurred to me that when trying to connect at this point the compilation thread is holding VMaccess. This means that GC cannot happen which in turn mean that the application is going to be blocked.

mpirvu:
I am thinking about the following implementation:
Keep 3 static variables somewhere:
serverAvailable
nextConnectionRetryTime
waitTime

Whenever there is a stream failure, call this method:
```
void postStreamFailure()
{
  lock()
  // Do not increase waiTime if multiple failures happen in parallel from different threads
  if (crtTime >= nextConnectionRetryTime)
      waitTime *= 2; // exponential backoff
  nextConnectionRetryTime = crtTime + waitTime;
  serverAvailable = false;
}
```
When we are about to establish a new connection:
```
try {
   if (serverAvailable) 
      conn = new J9ClientStream()
   else if (crtTime > nextConnectionRetryTime)
   {
      conn = new J9ClientStream()
      postStreamConnectionSuccess(); // This sets serverAvailable = true and waitTime = 1000
   }
   else
   {
   // fail the compilation with an error code that would lead us to retry locally
   }
} catch (StreamFailure) {
   postStreamFailure();
   other stuff
}
```
When doing the test of whether we need a local compilation we need to test
```
if (serverAvailable || crtTime > nextCoonnectionRetryTime)
```
if none of these conditions are met, we are going to issue a local compilation.

harryyu1994:
Tested, ready for review.

harryyu1994:
New plan: move this feature out of `ClientStream`, as these `*Stream` classes will be part of OMR. We don't want to enforce this feature to all users.

harryyu1994:
Tested, ready for review.

PR #5684, opened 2019-05-07
Create implementations of getNextSlotMap with new signature

dmitripivkine:
This is step 1 to remove parameters by reference from signature of
getNextSlotMap: create implementations with new signature.
Step 2 would be modify signature in OMR and step 3 remove
implementations with old signature.
New implementations are mostly copy/paste with minor changes:
- make functions to have single return;
- zero output parameter when missed;
- extend function description comment.


dmitripivkine:
Jenkins compile win

PR #5692, opened 2019-05-08
Create resolved methods in batches

dmitry-ten:
Most of the remote messages creating resolved methods
come from `TR_J9EstimateCodeSize::realEstimateCodeSize`.
The first step of size estimation is to walk through
method bytecodes and create a resolved method for every
invocation bytecode. So this loop will cause many remote
queries.
I merged these queries into one, by iterating through the
bytecodes right before estimator does it, identifying all
bytecodes that require a resolved method, and asking the client
to create them all in one batch, i.e. send only 1 message pair.

### DIFF
--- a/runtime/compiler/env/j9methodServer.cpp
+++ b/runtime/compiler/env/j9methodServer.cpp
@@ -27,6 +27,7 @@
 #include "control/MethodToBeCompiled.hpp"
 #include "control/JITaaSCompilationThread.hpp"
 #include "exceptions/DataCacheError.hpp"
+#include "ilgen/J9ByteCodeIterator.hpp"
 
 ClientSessionData::ClassInfo &
 getJ9ClassInfo(TR::CompilationInfoPerThread *threadCompInfo, J9Class *clazz)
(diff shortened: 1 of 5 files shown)


dmitry-ten:
@ymanton could you review it when you have time?

dmitry-ten:
Ok, ready for review now.

dmitry-ten:
Fixed a merge conflict. This still needs a review.

dmitry-ten:
All of the sanity tests that normally pass passed.

runtime/compiler/env/j9methodServer.cpp
```diff
@@ -1651,6 +1652,102 @@ TR_ResolvedJ9JITaaSServerMethod::addValidationRecordForCachedResolvedMethod(cons
...
+      TR_ResolvedMethodType type = methodTypes[i];
+      TR_ResolvedMethod *resolvedMethod;
+      if (std::get<0>(methodInfos[i]).remoteMirror &&
+         !compInfoPT->getCachedResolvedMethod(
+             compInfoPT->getResolvedMethodKey(type, (TR_OpaqueClassBlock *) _ramClass, cpIndices[i]),
+             this,
+             &resolvedMethod))
+         {
+         compInfoPT->cacheResolvedMethod(
+            compInfoPT->getResolvedMethodKey(type, (TR_OpaqueClassBlock *) _ramClass, cpIndices[i]),
```
ymanton: Can we create a local and reuse it here and above? It's not an expensive operation, but it also can't be inlined and probably can't be optimized away.

dmitry-ten: Done.

PR #5699, opened 2019-05-09
Make Escape Analysis sensitive to aternary operations

hzongaro:
Escape Analysis needs to take into account the fact that uses of objects might occur through references to addresses that are used through `aternary` operations.

Make a pass over trees to check which trees are used as operands whose values an `aternary` operation might yield.  Then in `checkUsesAndDefs`, check for uses for nodes for a particular candidate that might occur through those `aternary` operations.

### DIFF
--- a/runtime/compiler/optimizer/EscapeAnalysis.cpp
+++ b/runtime/compiler/optimizer/EscapeAnalysis.cpp
@@ -1951,26 +1951,36 @@ void TR_EscapeAnalysis::checkDefsAndUses()
 
 void TR_EscapeAnalysis::printUsesThroughAternary(void)
    {
-   traceMsg(comp(), "\nNodes used through aternary operations\n");
-
-   for (auto mi = _nodeUsesThroughAternary->begin(); mi != _nodeUsesThroughAternary->end(); mi++)
+   if (trace())
       {
-      TR::Node *key = mi->first;
-      int32_t nodeIdx = key->getGlobalIndex();
-
-      traceMsg(comp(), "   node [%p] n%dn is used by {", key, nodeIdx);
-
-      TR_ArrayIterator<TR::Node> ai(mi->second);
-      bool first = true;
-
-      for (TR::Node* aternaryNode = ai.getFirst(); aternaryNode; aternaryNode = ai.getNext())
+      if (_nodeUsesThroughAternary)
          {
-         traceMsg(comp(), "%s[%p] n%dn", (first ? "" : ", "), aternaryNode,
-                  aternaryNode->getGlobalIndex());
-         first = false;
-         }
+         traceMsg(comp(), "\nNodes used through aternary operations\n");
 
-      traceMsg(comp(), "}\n");
+         for (auto mi = _nodeUsesThroughAternary->begin(); mi != _nodeUsesThroughAternary->end(); mi++)
+            {
+            TR::Node *key = mi->first;
+            int32_t nodeIdx = key->getGlobalIndex();
+
+            traceMsg(comp(), "   node [%p] n%dn is used by {", key, nodeIdx);
+
+            bool first = true;
+
+            for (auto di = mi->second->begin(), end = mi->second->end(); di != end; di++)
+               {
+               TR::Node *aternaryNode = *di;
+               traceMsg(comp(), "%s[%p] n%dn", (first ? "" : ", "), aternaryNode,
+                        aternaryNode->getGlobalIndex());
+               first = false;
+               }
+
+            traceMsg(comp(), "}\n");
+            }
+         }
+      else
+         {
+         traceMsg(comp(), "\nNo nodes used through aternary operations\n");
+         }
       }
    }
 
@@ -1985,7 +1995,7 @@ void TR_EscapeAnalysis::gatherUsesThroughAternary(void)
       gatherUsesThroughAternaryImpl(node, visited);
       }
 
-   if (trace() && _nodeUsesThroughAternary)
+   if (trace())
       {
       printUsesThroughAternary();
       }
@@ -2018,31 +2028,36 @@ void TR_EscapeAnalysis::associateAternaryWithChild(TR::Node *aternaryNode, int32
    TR::Region &stackMemoryRegion = trMemory()->currentStackRegion();
    TR::Node *child = aternaryNode->getChild(idx);
 
-   TR_Array<TR::Node*> *currChildUses;
-   TR_HashId hashValue;
+   NodeDeque *currChildUses;
 
    if (NULL == _nodeUsesThroughAternary)
       {
       _nodeUsesThroughAternary =
-            new (trStackMemory()) NodeToNodeArrayMap((NodeComparator()),
-                                                     NodeToNodeArrayMapAllocator(stackMemoryRegion));
+            new (trStackMemory()) NodeToNodeDequeMap((NodeComparator()),
+                                                     NodeToNodeDequeMapAllocator(stackMemoryRegion));
       }
 
    auto search = _nodeUsesThroughAternary->find(child);
+   bool nodeAlreadyMapsToAternary = false;
 
    if (_nodeUsesThroughAternary->end() != search)
       {
       currChildUses = search->second;
+
+      // Does NodeDeque already contain this aternary node?
+      nodeAlreadyMapsToAternary =
+            (std::find(search->second->begin(), search->second->end(),
+                       aternaryNode) != search->second->end());
       }
    else
       {
-      currChildUses = new (trStackMemory()) TR_Array<TR::Node*>(trMemory(), 8, false, stackAlloc);
+      currChildUses = new (trStackMemory()) NodeDeque(stackMemoryRegion);
       (*_nodeUsesThroughAternary)[child] = currChildUses;
       }
 
-   if (!currChildUses->contains(aternaryNode))
+   if (!nodeAlreadyMapsToAternary)
       {
-      currChildUses->add(aternaryNode);
+      currChildUses->push_back(aternaryNode);
       }
    }
 
@@ -2206,7 +2221,6 @@ bool TR_EscapeAnalysis::collectValueNumbersOfIndirectAccessesToObject(TR::Node *
 bool TR_EscapeAnalysis::checkUsesThroughAternary(TR::Node *node, Candidate *candidate)
    {
    bool returnValue = true;
-   TR_HashId hashValue;
 
    if (_nodeUsesThroughAternary)
       {
@@ -2215,10 +2229,9 @@ bool TR_EscapeAnalysis::checkUsesThroughAternary(TR::Node *node, Candidate *cand
       // Is this node referenced directly by any aternary nodes?
       if (_nodeUsesThroughAternary->end() != search)
          {
-         TR_ArrayIterator<TR::Node> ai(search->second);
-
-         for (TR::Node* aternaryNode = ai.getFirst(); aternaryNode; aternaryNode = ai.getNext())
+         for (auto di = search->second->begin(), end = search->second->end(); di != end; di++)
             {
+            TR::Node* aternaryNode = *di;
             int32_t aternaryVN = _valueNumberInfo->getValueNumber(aternaryNode);
             int32_t i;
 
--- a/runtime/compiler/optimizer/EscapeAnalysis.cpp
+++ b/runtime/compiler/optimizer/EscapeAnalysis.cpp
@@ -1949,37 +1949,32 @@ void TR_EscapeAnalysis::checkDefsAndUses()
       }
    }
 
-void printUsesThroughAternary(TR::Compilation *comp, TR_HashTab *usesTable)
+void TR_EscapeAnalysis::printUsesThroughAternary(void)
    {
-   traceMsg(comp, "\nNodes used through aternary operations\n");
-   TR_HashTabIterator hti(usesTable);
-   TR_HashId nextID;
-   TR_Array<TR::Node*> *aternaryUses = (TR_Array<TR::Node*>*) hti.getFirst(nextID);
+   traceMsg(comp(), "\nNodes used through aternary operations\n");
 
-   while (aternaryUses)
+   for (auto mi = _nodeUsesThroughAternary->begin(); mi != _nodeUsesThroughAternary->end(); mi++)
       {
-      TR::Node *key = (TR::Node*) usesTable->getKey(nextID);
+      TR::Node *key = mi->first;
       int32_t nodeIdx = key->getGlobalIndex();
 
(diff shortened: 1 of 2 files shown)


hzongaro:
Vijay @vijaysun-omr, Andrew @andrewcraik, may I ask you to review this pull request?  In particular, is this the best approach to looking through aternary to underlying object references.

Also, this pull request is dependent on [OMR pull request 3825](https://github.com/eclipse/omr/pull/3825) which adds new methods to `TR_HashTab` and `TR_HashTabIterator`.  I wasn't sure whether it's preferable to use those classes or if newer code ought to use `std::map`, say.

andrewcraik:
Unless there is a performance reason to use `TR_HashTab` new could should use the `std::map` - generally the STL is less surprising and there is no ongoing maintenance.

runtime/compiler/optimizer/EscapeAnalysis.cpp
```diff
@@ -1945,7 +1949,101 @@ void TR_EscapeAnalysis::checkDefsAndUses()
...
+
+   if (_nodeUsesThroughAternary == NULL
+          && node->getOpCode().isTernary()
+          && node->getDataType() == TR::Address)
+      {
+      _nodeUsesThroughAternary = new (trStackMemory()) TR_HashTab(trMemory(), stackAlloc);
+      }
+
+   int32_t i;
+   for (i=0;i<node->getNumChildren(); i++)
```
andrewcraik: why the larger scope for I than the loop?

andrewcraik: would this make more sense unrolled so as to process each child explicitly since ternaries have a fixed number of children - the loop makes this seem more general than it needs to be...

runtime/compiler/optimizer/EscapeAnalysis.hpp
```diff
@@ -458,6 +459,9 @@ class TR_EscapeAnalysis : public TR::Optimization
    bool     collectValueNumbersOfIndirectAccessesToObject(TR::Node *node, Candidate *candidate, TR::Node *indirectStore, TR::NodeChecklist& visited, int32_t baseChildVN = -1);
    void     checkDefsAndUses();
    bool     checkDefsAndUses(TR::Node *node, Candidate *candidate);
+   void     gatherUsesThroughAternary();
+   void     gatherUsesThroughAternary(TR::Node *node, TR::NodeChecklist& visited);
```
andrewcraik: If this is the recursive implementation body convention would say add something like Impl to its suffix and if it isn't meant to be part of the opt API we could consider hiding it as private / protected.

runtime/compiler/optimizer/EscapeAnalysis.cpp
hzongaro: > why the larger scope for I than the loop?

Hmmm.  I might have started off planning to break out of the loop early or something.  Thanks for catching that!

> would this make more sense unrolled so as to process each child explicitly since ternaries have a fixed number of children

The recursive `gatherUsesThroughAternary` applies more generally, so that call needs to stay in a loop over `[0..getNumChildren())`.  But I will pull out the portion dealing with the second and third operands of `aternary`.

runtime/compiler/optimizer/EscapeAnalysis.hpp
hzongaro: > we could consider hiding it as private / protected

For some reason everything in `TR_EscapeAnalysis` is public. . . .  I'll see about hiding more of it.

runtime/compiler/optimizer/EscapeAnalysis.cpp
hzongaro: Andrew @andrewcraik, both of these are fixed in https://github.com/eclipse/openj9/pull/5699/commits/d10e03a13121fc002836f50c90327408cdfd4f63

runtime/compiler/optimizer/EscapeAnalysis.hpp
hzongaro: Andrew @andrewcraik, I've renamed the recursive implementation of `gatherUsesThroughAternary` to `gatherUsesThroughAternaryImpl` in https://github.com/eclipse/openj9/pull/5699/commits/d10e03a13121fc002836f50c90327408cdfd4f63

In https://github.com/eclipse/openj9/pull/5699/commits/48afe7397e38009460abb6dce0baf9cabb16d479, I made most of the method and fields `protected`, but needed to add two `friend` classes -- `TR_FlowSensitiveEscapeAnalysis` and `TR_LocalFlushElimination` -- that pick through the implementation of `TR_EscapeAnalysis`.  But I'm not really happy with it.  I'm inclined to do this more properly in a follow up issue.  Thoughts?

hzongaro: > do this more properly

Just to clarify, I mean adjust the API so that `TR_FlowSensitiveEscapeAnalysis` and `TR_LocalFlushElimination` gather information from `TR_EscapeAnalysis` in a way that's a bit tidier.

runtime/compiler/optimizer/EscapeAnalysis.cpp
```diff
@@ -1945,7 +1949,102 @@ void TR_EscapeAnalysis::checkDefsAndUses()
...
+      int32_t nodeIdx = key->getGlobalIndex();
+
+      traceMsg(comp(), "   node [%p] n%dn is used by {", key, nodeIdx);
+
+      TR_ArrayIterator<TR::Node> ai(mi->second);
+      bool first = true;
+
+      for (TR::Node* aternaryNode = ai.getFirst(); aternaryNode; aternaryNode = ai.getNext())
+         {
+         traceMsg(comp(), "%s[%p] n%dn", (first ? "" : ", "), aternaryNode,
```
andrewcraik: should be guarded for trace

runtime/compiler/optimizer/EscapeAnalysis.cpp
```diff
@@ -1945,7 +1949,102 @@ void TR_EscapeAnalysis::checkDefsAndUses()
...
+                  aternaryNode->getGlobalIndex());
+         first = false;
+         }
+
+      traceMsg(comp(), "}\n");
```
andrewcraik: guard for trace

runtime/compiler/optimizer/EscapeAnalysis.cpp
```diff
@@ -1945,7 +1949,102 @@ void TR_EscapeAnalysis::checkDefsAndUses()
...
+void TR_EscapeAnalysis::printUsesThroughAternary(void)
+   {
+   traceMsg(comp(), "\nNodes used through aternary operations\n");
+
+   for (auto mi = _nodeUsesThroughAternary->begin(); mi != _nodeUsesThroughAternary->end(); mi++)
+      {
+      TR::Node *key = mi->first;
+      int32_t nodeIdx = key->getGlobalIndex();
+
+      traceMsg(comp(), "   node [%p] n%dn is used by {", key, nodeIdx);
```
andrewcraik: guard for trace

runtime/compiler/optimizer/EscapeAnalysis.cpp
```diff
@@ -1945,7 +1949,102 @@ void TR_EscapeAnalysis::checkDefsAndUses()
...
+
+   for (; tt; tt = tt->getNextTreeTop())
+      {
+      TR::Node *node = tt->getNode();
+      gatherUsesThroughAternaryImpl(node, visited);
+      }
+
+   if (trace() && _nodeUsesThroughAternary)
+      {
+      printUsesThroughAternary();
```
andrewcraik: probably a good idea to say that there were no aternary opcodes found or something rather than being silent.

andrewcraik:
Doxygen for the new APIs would be helpful to make it easier to follow the code.

runtime/compiler/optimizer/EscapeAnalysis.hpp
```diff
@@ -645,6 +650,15 @@ class TR_EscapeAnalysis : public TR::Optimization
...
    TR_BitVector *             _vnTemp;
    TR_BitVector *             _vnTemp2;
+
+   typedef TR::typed_allocator<std::pair<TR::Node* const, TR_Array<TR::Node*>*>, TR::Region&> NodeToNodeArrayMapAllocator;
+   typedef std::less<TR::Node*> NodeComparator;
+   typedef std::map<TR::Node*, TR_Array<TR::Node*>*, NodeComparator, NodeToNodeArrayMapAllocator> NodeToNodeArrayMap;
+
+   NodeToNodeArrayMap *       _nodeUsesThroughAternary;
+
+   friend class TR_FlowSensitiveEscapeAnalysis;
```
andrewcraik: are the declarations of these brought in from the header includes? if not a forward declaration in this header is a must to avoid undefined behaviour on whether a friend class declaration is a forward declaration as well or not.

runtime/compiler/optimizer/EscapeAnalysis.hpp
andrewcraik: The extra encapsulation from protected is reasonable and fixing up the need for the friend class relationship can wait for the time being IMO.

andrewcraik:
change otherwise seems reasonable - an opinion from @vijaysun-omr is warranted since he is the real EA expert. TR_Array could be changed to std::deque to further minimize use of non-standard data structures.

andrewcraik:
@vijaysun-omr review ping?

vijaysun-omr:
I assume the aternary behaviour is like the behaviour we would have got if we had control flow equivalent to the ternary opcode in the sense that the ternary probably has a unique value number and it will be the one that gets added to the related value numbers set. This is (I think) similar to adding the value number of the load that might have been seen after the control flow merge in the original control flow.

hzongaro:
Vijay @vijaysun-omr wrote,
> I assume the aternary behaviour is like the behaviour we would have got if we had control flow equivalent to the ternary opcode in the sense that the ternary probably has a unique value number and it will be the one that gets added to the related value numbers set.

Yes, that's what I was attempting to achieve with this change.  Thank you for reviewing!

Andrew @andrewcraik, FYI - I'm still addressing your most recent review comments.  I'm just working through some build problems unrelated to those changes.

andrewcraik:
@hzongaro Is your last push actually on the branch that is being requested to be merged because the unified diff does not seem to be updated for some reason...

runtime/compiler/optimizer/EscapeAnalysis.cpp
hzongaro: Added a message in the case where no `aternary` opcodes were found in commit https://github.com/eclipse/openj9/pull/5699/commits/f822f8ae1f3604b2a5c3a855d3476f2390a9c4e0

runtime/compiler/optimizer/EscapeAnalysis.hpp
hzongaro: Fixed in commit https://github.com/eclipse/openj9/pull/5699/commits/f822f8ae1f3604b2a5c3a855d3476f2390a9c4e0

runtime/compiler/optimizer/EscapeAnalysis.cpp
```diff
@@ -1945,7 +1949,102 @@ void TR_EscapeAnalysis::checkDefsAndUses()
...
+      }
+   }
+
+void TR_EscapeAnalysis::associateAternaryWithChild(TR::Node *aternaryNode, int32_t idx)
+   {
+   TR::Region &stackMemoryRegion = trMemory()->currentStackRegion();
+   TR::Node *child = aternaryNode->getChild(idx);
+
+   TR_Array<TR::Node*> *currChildUses;
+   TR_HashId hashValue;
```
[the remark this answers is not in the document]
hzongaro: Thanks for catching that!  Removed in commit https://github.com/eclipse/openj9/pull/5699/commits/f822f8ae1f3604b2a5c3a855d3476f2390a9c4e0

hzongaro:
@andrewcraik Andrew, you asked

> @hzongaro Is your last push actually on the branch that is being requested to be merged because the unified diff does not seem to be updated for some reason...

No, it was on a separate branch that I was using for testing.  I didn't want to put it on this branch until I was ready.  I've just delivered that commit onto the `ea-aternary-fix` branch.

hzongaro:
@andrewcraik Andrew, you wrote

> TR_Array could be changed to std::deque to further minimize use of non-standard data structures.

Yes, you're right - I made that change in commit https://github.com/eclipse/openj9/commit/f822f8ae1f3604b2a5c3a855d3476f2390a9c4e0

hzongaro:
@andrewcraik Andrew, you recommended that Doxygen comments be added.  I've added those in commit https://github.com/eclipse/openj9/commit/f822f8ae1f3604b2a5c3a855d3476f2390a9c4e0

hzongaro:
@andrewcraik Andrew, sorry for all the @ mentions!  I mentioned in person that several z/OS tests were failing in a personal build I ran with this change.  I tried rerunning one of those tests with `disableEscapeAnalysis` and reproduced that same failure.  Thus, I believe those failures are unrelated to this EA change and can be investigated independently of this pull request.

andrewcraik:
@hzongaro thanks for all the work on this - I think it looks good - launching sanity builds.

andrewcraik:

andrewcraik:
@vijaysun-omr back to you for another review before we merge once the builds are clean.

vijaysun-omr:
I'm fine with this change.

ymanton:
@hzongaro FYI debug builds are broken:

```
/home/ymanton/jaas/openj9-openjdk-jdk8/build/linux-x86_64-normal-server-release/vm/compiler/../compiler/optimizer/EscapeAnalysis.cpp:3538:19: error: ‘TR_Array<TR::Node*>* TR_EscapeAnalysis::_parms’ is protected within this context
    TR_ASSERT(!ea->_parms, "rememberFieldSymRef: cannot remember peeked field SymRefs");
```

hzongaro:
@ymanton wrote:

> @hzongaro FYI debug builds are broken:
> 
> ```
> /home/ymanton/jaas/openj9-openjdk-jdk8/build/linux-x86_64-normal-server-release/vm/compiler/../compiler/optimizer/EscapeAnalysis.cpp:3538:19: error: ‘TR_Array<TR::Node*>* TR_EscapeAnalysis::_parms’ is protected within this context
>     TR_ASSERT(!ea->_parms, "rememberFieldSymRef: cannot remember peeked field SymRefs");
> ```

Argh!  Sorry about that.  Looking at it.

hzongaro:
@ymanton Younes, the problem I introduced for debug builds with this pull request should be fixed with pull request #6014.  Sorry for the bother!

ymanton:
No bother, thanks.

PR #5704, opened 2019-05-09
Fix bug in profiled inlined method relocation

dsouzai:
For the TR_RelocationRecordProfiledInlinedMethod and subclass relocation
records, what's stored in the relo header is:

1. The class chain of class specified in the TR_VirtualGuard
2. The class chain of the first class loaded by the classloader that
loaded the class in 1.
3. the VFT offset

The idea was to validate that the receiver was the same, and then to add
the appropriate inlined J9Method into the J9JitExceptionTable's inlined
method table during relocation time. The VFT offset was determined by
using the symref stored in the guard (which is the symref of the
*caller*).

A problem occurs when there is an invokeVirtual called on a method via a
class that implements an interface AND the method being invoked is a
default method of that interface. In other words, the defining class of
the method is an interface class, and the implementing class inherits
this method.

When the compiler creates the profiled guard, the TR_VirtualGuard
contains the J9Class of the defining class; in this case it would be the
interface class. However, the symref contains the VFT offset of a
concrete class. Thus, at load time, the relo infrastructure will
materialize the interface class based on the class chain info in the
relo header, and then try to pull out a J9Method from that class' VFT
using the VFT offset. However, interface classes do not have a VFT. As a
result, this can cause undefined behaviour.

This PR addresses this issue by modifying the relo header to store
the method index from the defining class rather than the VFT offset from
the caller. This involves also now changing the the class chains stored
to:

1. The class chain of the defining class of the inlined method
2. The class chain of the first class loaded by the classloader that
loaded the class in 1.

Because this is a profiled inlined method, whether the guard
is a Method Test, VFT Test, or Removed Guard, this will yield the
correct method to populate in the inlined table of the
J9JITExceptionTable.

One important subtlety though, is that in the case of the VFT Test, the
inlined code might have certain assumptions based on the type. However,the
TR_RelocationRecordProfiledInlinedMethod (and subtypes) primarily
deal with populating the inlined table in the J9JITExceptionTable. The
VFT test will result in a TR_ClassPointer relocation record that will
relocate the class being tested by the VFT test guard.

### DIFF
--- a/runtime/compiler/arm/codegen/J9AheadOfTimeCompile.cpp
+++ b/runtime/compiler/arm/codegen/J9AheadOfTimeCompile.cpp
@@ -456,16 +456,16 @@ uint8_t *J9::ARM::AheadOfTimeCompile::initializeAOTRelocationHeader(TR::Iterated
 
          TR::SymbolReference *callSymRef = guard->getSymbolReference();
          TR_ResolvedMethod *owningMethod = callSymRef->getOwningMethod(comp);
+
+         TR_InlinedCallSite & ics = comp->getInlinedCallSite(inlinedSiteIndex);
+         TR_ResolvedMethod *inlinedMethod = ((TR_AOTMethodInfo *)ics._methodInfo)->resolvedMethod;
+         TR_OpaqueClassBlock *inlinedCodeClass = reinterpret_cast<TR_OpaqueClassBlock *>(inlinedMethod->classOfMethod());
+
          *(uintptrj_t *) cursor = (uintptrj_t) owningMethod->constantPool(); // record constant pool
          cursor += SIZEPOINTER;
 
          *(uintptrj_t*) cursor = (uintptrj_t) callSymRef->getCPIndex(); // record cpIndex
          cursor += SIZEPOINTER;
-
-         TR_OpaqueClassBlock *inlinedCodeClass = guard->getThisClass();
-         int32_t len;
-         char *className = fej9->getClassNameChars(inlinedCodeClass, len);
-         //traceMsg(comp(), "inlined site index is %d, inlined code class is %.*s\n", inlinedSiteIndex, len, className);
 
          void *romClass = (void *) fej9->getPersistentClassPointerFromClassPointer(inlinedCodeClass);
          void *romClassOffsetInSharedCache = sharedCache->offsetInSharedCacheFromPointer(romClass);
@@ -479,27 +479,8 @@ uint8_t *J9::ARM::AheadOfTimeCompile::initializeAOTRelocationHeader(TR::Iterated
 
          cursor = self()->emitClassChainOffset(cursor, inlinedCodeClass);
 
-         uintptrj_t offset;
-         TR::MethodSymbol *calleeSymbol = callSymRef->getSymbol()->castToMethodSymbol();
-         if (!TR::Compiler->cls.isInterfaceClass(comp, inlinedCodeClass) && calleeSymbol->isInterface())
-            {
-            int32_t index = owningMethod->getResolvedInterfaceMethodOffset(inlinedCodeClass, callSymRef->getCPIndex());
-            //traceMsg(comp(),"ProfiledGuard offset for interface is %d\n", index);
-            offset = (uintptrj_t) index;
-            if (index < 0)
-               offset = (uintptrj_t) fej9->virtualCallOffsetToVTableSlot(index);
-            //traceMsg(comp(),"ProfiledGuard offset for interface is %d\n", index);
-            }
-         else
-            {
-            offset = (int32_t) callSymRef->getOffset();
-            //traceMsg(comp(),"ProfiledGuard offset for non-interface is %d\n", offset);
-            offset = (uintptrj_t) fej9->virtualCallOffsetToVTableSlot((int32_t) callSymRef->getOffset());
-            //traceMsg(comp(),"ProfiledGuard offset for non-interface is now %d\n", offset);
-            }
-
-         //traceMsg(comp(),"ProfiledGuard offset using %d\n", offset);
-         *(uintptrj_t *) cursor = offset;
+         uintptrj_t methodIndex = fej9->getMethodIndexInClass(inlinedCodeClass, inlinedMethod->getNonPersistentIdentifier());
+         *(uintptrj_t *)cursor = methodIndex;
          cursor += SIZEPOINTER;
          }
          break;
--- a/runtime/compiler/env/VMJ9.cpp
+++ b/runtime/compiler/env/VMJ9.cpp
@@ -6451,6 +6451,28 @@ TR_J9VMBase::getNumMethods(TR_OpaqueClassBlock * classPointer)
    return TR::Compiler->cls.romClassOf(classPointer)->romMethodCount;
    }
 
+/* This API takes a J9Class (TR_OpaqueClassBlock) and a J9Method (TR_OpaqueMethodBlock) and returns the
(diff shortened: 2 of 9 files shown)


dsouzai:
Didn't realize I had to make the PR a draft at creation time, so WIP it'll have to be.

@jdmpapin @mstoodle could you guys please review? I want to do more tests on this, but I figure it's worth getting your eyes on the code.

runtime/compiler/arm/codegen/J9AheadOfTimeCompile.cpp
```diff
@@ -479,27 +479,22 @@ uint8_t *J9::ARM::AheadOfTimeCompile::initializeAOTRelocationHeader(TR::Iterated
...
 
-         uintptrj_t offset;
-         TR::MethodSymbol *calleeSymbol = callSymRef->getSymbol()->castToMethodSymbol();
-         if (!TR::Compiler->cls.isInterfaceClass(comp, inlinedCodeClass) && calleeSymbol->isInterface())
+         J9Method *inlinedJ9Method = reinterpret_cast<J9Method *>(inlinedMethod->getNonPersistentIdentifier());
+         J9Method *resolvedMethods = static_cast<J9Method *>(fej9->getMethods(inlinedCodeClass));
+
+         uint32_t numMethods = fej9->getNumMethods(inlinedCodeClass);
+         uint32_t methodIndex = 0;
+         for (; methodIndex < numMethods ; methodIndex++)
```
jdmpapin: It bothers me a little bit that this is looping when a pointer subtraction would do. Unfortunately, pointer subtraction is UB in case some bug causes `inlinedJ9Method` to be missing from `resolvedMethods`. I think it's possible to do the subtraction using integers and check the range though, e.g.
```
uintptr_t methodOffset = (uintptr_t)inlinedJ9Method - (uintptr_t)resolvedMethods;
TR_ASSERT_FATAL(methodOffset % sizeof (J9Method) == 0, "...");
uintptr_t methodIndex = methodOffset / sizeof (J9Method);
TR_ASSERT_FATAL(methodIndex < numMethods, "...");
// Now it's safe to truncate methodIndex to 32-bit
```
But I understand if you're more comfortable with the loop :man_shrugging:

dsouzai: I don't have a concern with not using a loop as long as there's i add a comment describing what the code's doing.

dsouzai:
> The duplication of logic newly added to the implementations of initializeAOTRelocationHeader() sticks out to me, but that's the case for everything in those methods and deduplicating it is a separate issue

Yeah I'm taking care of that as part of my consolidation effort (which, amusingly enough, will now have a merge conflict because of this change :slightly_smiling_face:)

dsouzai:
@jdmpapin @mstoodle made the requested changes. Could you guys please re-review?

runtime/compiler/env/VMJ9.cpp
```diff
@@ -6451,6 +6451,25 @@ TR_J9VMBase::getNumMethods(TR_OpaqueClassBlock * classPointer)
    return TR::Compiler->cls.romClassOf(classPointer)->romMethodCount;
    }
 
+/* This API takes a J9Class (TR_OpaqueClassBlock) and a J9Method (TR_OpaqueMethodBlock) and returns the
+ * index of the J9Method in the J9Class' array of J9Methods.
+ */
+uintptr_t
+TR_J9VMBase::getMethodIndexInClass(TR_OpaqueClassBlock *classPointer, TR_OpaqueMethodBlock *methodPointer)
```
mstoodle: better name than the one I quickly proposed. I like this a lot better now.

runtime/compiler/env/VMJ9.cpp
```diff
@@ -6451,6 +6451,25 @@ TR_J9VMBase::getNumMethods(TR_OpaqueClassBlock * classPointer)
...
+   {
+   void *methodsInClass = getMethods(classPointer);
+   uint32_t numMethods = getNumMethods(classPointer);
+
+   uintptr_t methodOffset = reinterpret_cast<uintptr_t>(methodPointer) - reinterpret_cast<uintptr_t>(methodsInClass);
+   TR_ASSERT_FATAL((methodOffset % sizeof (J9Method)) == 0, "methodOffset %p isn't a multiple of sizeof(J9Method)\n", methodOffset);
+
+   uintptr_t methodIndex = methodOffset / sizeof (J9Method);
+   TR_ASSERT_FATAL(methodIndex < numMethods, "methodIndex %p greater than numMethods %d for method %p in class %p\n",
```
jdmpapin: Should this format string (and the one just above) be using `%p` for integers?

I'd like to recommend `PRIxPTR`, but the format string is used for both `vfprintf()` and `omrstr_vprintf()`, which don't agree on the interpretation of all format specifiers. In particular, `omrstr_vprintf()` reads a 32-bit integer for `%lx` even on LP64 systems, where `PRIxPTR` could be `"lx"`.

I think the safest way to format a pointer-sized integer here is to cast it to `long long` (which should be 64-bit on all of our platforms) and use `%llx`.

runtime/compiler/arm/codegen/J9AheadOfTimeCompile.cpp
```diff
@@ -479,27 +479,8 @@ uint8_t *J9::ARM::AheadOfTimeCompile::initializeAOTRelocationHeader(TR::Iterated
...
-            offset = (int32_t) callSymRef->getOffset();
-            //traceMsg(comp(),"ProfiledGuard offset for non-interface is %d\n", offset);
-            offset = (uintptrj_t) fej9->virtualCallOffsetToVTableSlot((int32_t) callSymRef->getOffset());
-            //traceMsg(comp(),"ProfiledGuard offset for non-interface is now %d\n", offset);
-            }
-
-         //traceMsg(comp(),"ProfiledGuard offset using %d\n", offset);
-         *(uintptrj_t *) cursor = offset;
+         uintptrj_t methodIndex = fej9->getMethodIndexInClass(inlinedCodeClass, inlinedMethod->getNonPersistentIdentifier());
+         *(uintptrj_t *)cursor =  static_cast<uintptrj_t>(methodIndex);
```
jdmpapin: This line (along with the other copies) is now casting from `uintptrj_t` to `uintptrj_t`.

runtime/compiler/env/VMJ9.cpp
dsouzai: I"ll cast it to `uint64_t` (just in case) and use `%llx` as you suggest.

dsouzai:
@jdmpapin made the requested changes (see https://github.com/eclipse/openj9/compare/1aa0eefd5f423f23dbe2b2bcb0aa1f572f6fa437..5d9d9a4925adbe2bddeec3f11e5006306c29ecb4)

dsouzai:
Jenkins test sanity.functional+aot all jdk8

dsouzai:
I've opened https://github.com/eclipse/openj9/issues/4875 to remove existing tests that are not valid for AOT, but I figure it's worth seeing if the AOT tests work :)

dsouzai:
Jenkins test sanity.functional+aot xlinux,plinux,zlinux jdk8,jdk11

dsouzai:
I took a look at the failing jobs (there were only a handful, which was nice to see). None were AOT issues, but rather, test and configuration issues (of which I've opened https://github.com/eclipse/openj9/issues/5767 and https://github.com/eclipse/openj9/issues/5768). For now I'll just kick off regular sanity builds.

dsouzai:
@mstoodle all tests passed; could you merge if everything looks good to you.

PR #5708, opened 2019-05-09
Make PassThrough x86 NULLCHKs implicit

andrewcraik:
Currently a NULLCHK with a PassThrough will evaluate into test and
branch sequences. This change switches the sequence to simply touch
the memory of the value being NULLCHK'd. A trap will result in the
appropriate null pointer exception and saves a branch at the cost of
some potential cache pollution.

Signed-off-by: Andrew Craik <ajcraik@ca.ibm.com>


andrewcraik:

fjeremic:
Linux failed due to infrastructure. Relaunching.

andrewcraik:
@0xdaryl could you review the x codegen change for me? Ordinarily I do the x reviews, but approving my own work is not something I want to do...

PR #5712, opened 2019-05-09
Purge cached data of clients that suddenly disconnected

dmitry-ten:
When client suddenly terminates, it will not send a final
message to the server that indicates the end of session,
thus the server doesn't know that ClientSessionData should
be purged, and leaks persistent memory.
To fix that, periodically check if client data should be
purged from statistics thread.

### DIFF
--- a/runtime/compiler/runtime/StatisticsThread.cpp
+++ b/runtime/compiler/runtime/StatisticsThread.cpp
@@ -81,6 +81,11 @@ static int32_t J9THREAD_PROC statisticsThreadProc(void * entryarg)
          crtTime = j9time_current_time_millis();
          persistentInfo->setElapsedTime(crtTime - persistentInfo->getStartTime());
 
+            {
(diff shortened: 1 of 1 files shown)


dmitry-ten:
@ymanton could you review?
One problem I have with this is that `purgeOldDataIfNeeded` will only purge client sessions that are at least 1000 minutes old. That seems to be too large of a timeout.

ymanton:
Copyright check failed.

> One problem I have with this is that `purgeOldDataIfNeeded` will only purge client sessions that are at least 1000 minutes old. That seems to be too large of a timeout.

It should be settable via a command line option so users can modify it, and I agree it seems a bit large, but besides that I don't know what a good default value should be at this point. If we have a good reason to lower it in the future we can always consider it then.

dmitry-ten:
Copyright failed due to some jenkins error. @ymanton could you restart?

ymanton:
Jenkins copyright check

PR #5713, opened 2019-05-09
Move max/minAccessedShrCacheMetadata into CompositeCache

hangshao0:
1. CompositeCache is the correct level to put all the APIs/fields that
manage each cache. Move max/minAccessedShrCacheMetadata there.

2. There is no need to call updateAccessedShrCacheMetadataBounds() after
the metadata are released. So update max/minAccessedShrCacheMetadata
only when metadata are not released in the cold run.

### DIFF
--- a/runtime/shared_common/CacheMap.cpp
+++ b/runtime/shared_common/CacheMap.cpp
@@ -216,14 +216,17 @@ SH_CacheMap::newInstance(J9JavaVM* vm, J9SharedClassConfig* sharedClassConfig, S
 void
 SH_CacheMap::dontNeedMetadata(J9VMThread* currentThread) 
 {
-	/* Local copies to avoid race condition */
 	Trc_SHR_CM_j9shr_dontNeedMetadata(currentThread);
+	SH_CompositeCacheImpl* ccToUse = _ccHead;
 
+	if (_metadataReleased) {
+		return;
+	}
 	_metadataReleased = true;
-	uintptr_t  min = _minimumAccessedShrCacheMetadata;
-	uintptr_t  max = _maximumAccessedShrCacheMetadata;
-	size_t length = (size_t) (max - min);
-	_ccHead->dontNeedMetadata(currentThread, (const void *) min, length);
+	do {
+		ccToUse->dontNeedMetadata(currentThread);
+		ccToUse = ccToUse->getNext();
+	} while (NULL != ccToUse);
 }
 
 /**
@@ -274,8 +277,6 @@ SH_CacheMap::initialize(J9JavaVM* vm, J9SharedClassConfig* sharedClassConfig, Bl
 	_growEnabled = false;
 	_isSerialized = false;
 	_isAssertEnabled = true;
-	_minimumAccessedShrCacheMetadata = 0;
-	_maximumAccessedShrCacheMetadata = 0;
 	_metadataReleased = false;
 	
 	/* TODO: Need this function to be able to return pass/fail */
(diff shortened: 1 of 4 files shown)


runtime/shared_common/CompositeCache.cpp
```diff
@@ -6886,3 +6913,95 @@ SH_CompositeCacheImpl::isNewCache(void)
...
+				updateAddress,
+				currentValue,
+				newValue
+		);
+		currentValue = (uintptr_t) _minimumAccessedShrCacheMetadata;
+	}
+
+	updateAddress = (uintptr_t *) &_maximumAccessedShrCacheMetadata;
+
+	if (0 == currentValue) { /* set initial value.  Don't care if someone beats us to this  */
```
pshipton: This isn't necessary because the `newValue > _maximumAccessedShrCacheMetadata` will always catch this case. Plus it's weird because I don't think currentValue will ever be zero at this point.

hangshao0: This line is removed.

runtime/shared_common/CacheMap.cpp
```diff
@@ -216,14 +216,17 @@ SH_CacheMap::newInstance(J9JavaVM* vm, J9SharedClassConfig* sharedClassConfig, S
 void
 SH_CacheMap::dontNeedMetadata(J9VMThread* currentThread) 
 {
-	/* Local copies to avoid race condition */
 	Trc_SHR_CM_j9shr_dontNeedMetadata(currentThread);
+	SH_CompositeCacheImpl* ccToUse = _ccHead;
 
+	if (_metadataReleased) {
```
pshipton: Is there any guarantee this function will only be called once? I'm a little concerned about changing the behavior and releasing less memory after these changes than were released before these changes. If this is called multiple times, the original code tracked updated bounds in some cases and released the larger area.

pshipton:
jenkins test sanity zlinux,plinux,win jdk8

pshipton:
pipeline test sanity zlinux,plinux,win jdk8

runtime/shared_common/CacheMap.cpp
hangshao0: This function can be called more than once by multiple threads that detect vm phase change. The behaviors of original code is:
1. warm run: updateAccessedShrCacheMetadataBounds() is called only when _metadataReleased is false. So even though dontNeedMetadata() can be call more than once, max/minAccessedShrCacheMetadata won' be updated after the first dontNeedMetadata() call. So we were releasing the same region multiple times.

2. cold run: the cold run code was added by https://github.com/eclipse/openj9/pull/3454. The check of _metadataReleased was not added before calling updateAccessedShrCacheMetadataBounds() there. The behavior is inconsistent with warm run. 

This commit adds check of _metadataReleased before calling updateAccessedShrCacheMetadataBounds() in the cold run, so that it is consistent with warm run. It might be possible in a cold run that a smaller region get released after this change, but I guess the difference is negligible and within noise.

pshipton: > but I guess the difference is negligible and within noise
 
We should measure it rather than guessing. Cheng did measurements when the cold run changes were added to validate the results.

hangshao0: Run liberty startup 10 times with cold cache. The RSS (Mean (±Std Dev)) of the metadata is: 
Without this change: 2401 (±335) kbytes. 
With this change: 2298 (±292) kbytes.

pshipton: If I understand the numbers correctly, this is a 100K regression or about 4%. I wonder if we can avoid this regression.

hangshao0: The footprint is even lower with this change. But considering the standard deviation is about 300k, I think the 100k is just some noise.

pshipton: ok, I had it backwards.

PR #5622, opened 2019-04-25
Test JVMTI getSystemProperty at early phase Agent_OnLoad()

JasonFengJ9:
Test `JVMTI getSystemProperty` at early phase `Agent_OnLoad()`

Added test `gsp001` to get following system properties at `Agent_OnLoad()`:
```
  java.vm.vendor
  java.vm.version
  java.vm.name
  java.vm.info
  java.library.path
  java.class.path
  java.vm.specification.version
```
Verified that these system property values are not `null`.

Related: #5565 

Reviewers: @DanHeidinga @llxia 

Signed-off-by: Jason Feng <fengj@ca.ibm.com>

PR #5719, opened 2019-05-09
(v0.14.2-release) Init system property java.vm.specification.version via native

JasonFengJ9:
Init system property `java.vm.specification.version` via native

There is a need to retrieve the system property `java.vm.specification.version` via `JVMTI API` in an early phase before Java code `java.lang.VersionProps.init(systemProperties)` running.
Removed the Java code (Java 12+) setting the property;
Get the property value via `JAVA_SPEC_VERSION` and set it within `vmprops.c:initializeSystemProperties()`;

Verified that the system property `java.vm.specification.version` has correct value.

Cherry-Pick from https://github.com/eclipse/openj9/pull/5576

Reviewer: @pshipton 

### DIFF
--- a/jcl/src/java.base/share/classes/java/lang/System.java
+++ b/jcl/src/java.base/share/classes/java/lang/System.java
@@ -398,8 +398,6 @@ private static void ensureProperties(boolean isInitialization) {
 
 /*[IF Java12]*/
 	java.lang.VersionProps.init(initializedProperties);
-	/* VersionProps.init(systemProperties) above sets java.specification.version value which is used to set java.vm.specification.version. */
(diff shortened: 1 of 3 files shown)


pshipton:
Please include Dan's commit as well so we have the same code as the master branch.

JasonFengJ9:
Sorry, was looking at `System.java` and missed Dan's commit.
Assuming there is no need for the test PR https://github.com/eclipse/openj9/pull/5622.

pshipton:
> Assuming there is no need for the test PR #5622.

Agreed. It is tested in master to avoid regressions, and this is the same change.

Issue #3842, opened 2018-11-26
TestSunAttachClasses_0 failed with missing Agent_OnAttach string on osx JDK11


llxia:
https://ci.eclipse.org/openj9/job/PullRequest-Extended-JDK11-osx_x86-64_cmprssptrs-OpenJ9/9/tapResults/

```
00:21:33.637 + build/macosx-x86_64-normal-server-release/images/jdk/bin/java -version
00:21:34.925 openjdk version "11.0.1-internal" 2018-10-16
00:21:34.925 OpenJDK Runtime Environment (build 11.0.1-internal+0-adhoc.jenkins.PullRequest-Extended-JDK11-osxx86-64cmprssptrs-OpenJ9)
00:21:34.925 Eclipse OpenJ9 VM (build HEAD-6c12d1731, JRE 11 Mac OS X amd64-64-Bit Compressed References 20181126_9 (JIT enabled, AOT enabled)
00:21:34.925 OpenJ9   - 6c12d1731
00:21:34.925 OMR      - 8001cf22
00:21:34.925 JCL      - d42cd0d94f7 based on jdk-11.0.1+13)
```

```
01:42:48.025 ===============================================
01:42:48.025 Running test TestSunAttachClasses_0 ...
01:42:48.025 ===============================================
01:42:48.025 TestSunAttachClasses_0 Start Time: Mon Nov 26 09:44:36 2018 Epoch Time (ms): 1543243476087
01:42:48.025 variation: NoOptions
01:42:48.025 JVM_OPTIONS: -Xcompressedrefs 
01:42:48.025 { mkdir -p "/Users/jenkins/workspace/PullRequest-Extended-JDK11-osx_x86-64_cmprssptrs-OpenJ9/openj9/test/TestConfig/scripts/testKitGen/../../../TestConfig/test_output_15432393657299/TestSunAttachClasses_0"; \
01:42:48.025 	cd "/Users/jenkins/workspace/PullRequest-Extended-JDK11-osx_x86-64_cmprssptrs-OpenJ9/openj9/test/TestConfig/scripts/testKitGen/../../../TestConfig/test_output_15432393657299/TestSunAttachClasses_0"; \
[...]

PR #5737, opened 2019-05-11
Improve Escape Analysis under OSR

hzongaro:
This change defines new optimization passes, `preEscapeAnalysis` and `postEscapeAnalysis`, that help Escape Analysis (EA) do better under voluntary OSR.

The new `preEscapeAnalysis` optimization looks for calls to `OSRInductionHelper` and adds a fake `prepareForOSR` call that references all live auto symrefs and pending pushes.  Any object that ends up as a candidate for stack allocation that appears to be used by such a fake `prepareForOSR` call can be heapified at that point.

During EA itself, those references on `prepareForOSR` calls are marked as ignoreable for the purposes of determining whether an object can be stack allocated.

After EA, the `postEscapeAnalysis` pass will remove the fake calls to `prepareForOSR`.

Submitted on behalf of Andrew Craik <ajcraik@ca.ibm.com>

### DIFF
--- a/runtime/compiler/optimizer/EscapeAnalysis.cpp
+++ b/runtime/compiler/optimizer/EscapeAnalysis.cpp
@@ -544,6 +544,7 @@ int32_t TR_EscapeAnalysis::performAnalysisOnce()
    _fixedVirtualCallSites.setFirst(NULL);
 
    _parms = NULL;
+   _ignoreableUses = NULL;
    _nonColdLocalObjectsValueNumbers = NULL;
    _allLocalObjectsValueNumbers = NULL;
    _visitedNodes = NULL;
@@ -587,6 +588,7 @@ int32_t TR_EscapeAnalysis::performAnalysisOnce()
          }
       else
          {
+         _ignoreableUses = new (trStackMemory()) TR_BitVector(0, trMemory(), stackAlloc);
          _nonColdLocalObjectsValueNumbers = new (trStackMemory()) TR_BitVector(_valueNumberInfo->getNumberOfValues(), trMemory(), stackAlloc);
          _allLocalObjectsValueNumbers = new (trStackMemory()) TR_BitVector(_valueNumberInfo->getNumberOfValues(), trMemory(), stackAlloc);
          _notOptimizableLocalObjectsValueNumbers = new (trStackMemory()) TR_BitVector(_valueNumberInfo->getNumberOfValues(), trMemory(), stackAlloc);
@@ -597,6 +599,7 @@ int32_t TR_EscapeAnalysis::performAnalysisOnce()
    if ( !_candidates.isEmpty())
       {
       findLocalObjectsValueNumbers();
+      findIgnoreableUses();
       }
 
    // Complete the candidate info by finding all uses and defs that are reached
@@ -1261,6 +1264,48 @@ int32_t TR_EscapeAnalysis::performAnalysisOnce()
       }
 
    return cost; // actual cost
+   }
+
+void TR_EscapeAnalysis::findIgnoreableUses()
+   {
+   if (comp()->getOSRMode() != TR::voluntaryOSR)
+      return;
+
+   TR::NodeChecklist visited(comp());
+   bool inOSRCodeBlock = false;
+
+   // Gather all uses under fake prepareForOSR calls - they will be tracked as ignoreable
+   for (TR::TreeTop *treeTop = comp()->getStartTree(); treeTop; treeTop = treeTop->getNextTreeTop())
+       {
+       if (treeTop->getNode()->getOpCodeValue() == TR::BBStart)
+           inOSRCodeBlock = treeTop->getNode()->getBlock()->isOSRCodeBlock();
+       else if (inOSRCodeBlock
+                && treeTop->getNode()->getNumChildren() > 0
+                && treeTop->getNode()->getFirstChild()->getOpCodeValue() == TR::call
+                && treeTop->getNode()->getFirstChild()->getSymbolReference()->getReferenceNumber() == TR_prepareForOSR)
+           {
+           TR::Node *callNode = treeTop->getNode()->getFirstChild();
+           for (int i = 0; i < callNode->getNumChildren(); ++i)
+              findIgnoreableUses(callNode->getChild(i), visited);
+           }
+       }
+   }
+
+void TR_EscapeAnalysis::findIgnoreableUses(TR::Node *node, TR::NodeChecklist &visited)
+   {
+   if (visited.contains(node))
+      return;
+   visited.add(node);
+   if (trace())
+      traceMsg(comp(), "Marking n%dn as an ignoreable use\n", node->getGlobalIndex());
+   _ignoreableUses->set(node->getGlobalIndex());
+
+   int32_t i;
+   for (i = 0; i < node->getNumChildren(); i++)
+      {
+      TR::Node *child = node->getChild(i);
+      findIgnoreableUses(child, visited);
+      }
    }
 
 void TR_EscapeAnalysis::findLocalObjectsValueNumbers()
@@ -2307,15 +2352,28 @@ bool TR_EscapeAnalysis::checkDefsAndUses(TR::Node *node, Candidate *candidate)
                   TR::Node *useNode = _useDefInfo->getNode(useIndex+_useDefInfo->getFirstUseIndex());
                   int32_t useNodeVN = _valueNumberInfo->getValueNumber(useNode);
                   int32_t i;
+                  bool valueNumberSeen = false;
+
                   for (i = candidate->_valueNumbers->size()-1; i >= 0; i--)
                      {
                      if (candidate->_valueNumbers->element(i) == useNodeVN)
-                        break;
+                        {
+                        valueNumberSeen = true;
+
+                        // Only add this value number if it's not to be ignored
+                        if (!_ignoreableUses->get(candidate->_valueNumbers->element(i)))
+                           {
+                           break;
+                           }
+                        }
                      }
 
                   if (i < 0)
                      {
-                     candidate->_valueNumbers->add(useNodeVN);
+                     if (!valueNumberSeen)
+                        {
+                        candidate->_valueNumbers->add(useNodeVN);
+                        }
 
                      if (candidate->isInsideALoop())
                         {
@@ -2348,7 +2406,7 @@ bool TR_EscapeAnalysis::checkDefsAndUses(TR::Node *node, Candidate *candidate)
                         else
                            returnValue = false;
                         }
-                     if (!checkDefsAndUses(useNode, candidate))
+                     if (!valueNumberSeen && !checkDefsAndUses(useNode, candidate))
                         returnValue = false;
                      }
                   }
@@ -3911,7 +3969,12 @@ void TR_EscapeAnalysis::checkEscape(TR::TreeTop *firstTree, bool isCold, bool & 
       if (node->getOpCode().isCheck() || node->getOpCodeValue() == TR::treetop)
          node = node->getFirstChild();
       if (node->getOpCode().isCall() && !vnsCall.contains(node))
-         checkEscapeViaCall(node, vnsCall, ignoreRecursion);
+         {
+         if (node->getSymbolReference()->getReferenceNumber() != TR_prepareForOSR
+             || comp()->getOSRMode() != TR::voluntaryOSR
+             ||  _curBlock->isOSRInduceBlock())
+            checkEscapeViaCall(node, vnsCall, ignoreRecursion);
+         }
       }
    }
 
--- a/runtime/compiler/optimizer/EscapeAnalysis.cpp
+++ b/runtime/compiler/optimizer/EscapeAnalysis.cpp
@@ -2350,30 +2350,27 @@ bool TR_EscapeAnalysis::checkDefsAndUses(TR::Node *node, Candidate *candidate)
                   {
                   int32_t useIndex = cursor;
                   TR::Node *useNode = _useDefInfo->getNode(useIndex+_useDefInfo->getFirstUseIndex());
+
+                  // Only add this value number if it's not to be ignored
+                  if (_ignoreableUses->get(useNode->getGlobalIndex()))
+                     {
+                     continue;
+                     }
+
                   int32_t useNodeVN = _valueNumberInfo->getValueNumber(useNode);
                   int32_t i;
-                  bool valueNumberSeen = false;
 
                   for (i = candidate->_valueNumbers->size()-1; i >= 0; i--)
                      {
                      if (candidate->_valueNumbers->element(i) == useNodeVN)
                         {
-                        valueNumberSeen = true;
-
-                        // Only add this value number if it's not to be ignored
-                        if (!_ignoreableUses->get(candidate->_valueNumbers->element(i)))
-                           {
-                           break;
-                           }
+                        break;
                         }
                      }
 
                   if (i < 0)
                      {
-                     if (!valueNumberSeen)
-                        {
-                        candidate->_valueNumbers->add(useNodeVN);
-                        }
+                     candidate->_valueNumbers->add(useNodeVN);
 
                      if (candidate->isInsideALoop())
                         {
@@ -2406,7 +2403,7 @@ bool TR_EscapeAnalysis::checkDefsAndUses(TR::Node *node, Candidate *candidate)
                         else
                            returnValue = false;
                         }
-                     if (!valueNumberSeen && !checkDefsAndUses(useNode, candidate))
+                     if (!checkDefsAndUses(useNode, candidate))
                         returnValue = false;
                      }
                   }
--- a/runtime/compiler/optimizer/EscapeAnalysis.hpp
+++ b/runtime/compiler/optimizer/EscapeAnalysis.hpp
@@ -452,6 +452,8 @@ class TR_EscapeAnalysis : public TR::Optimization
 
    int32_t  performAnalysisOnce();
    void     findCandidates();
+   void     findIgnoreableUses();
+   void     findIgnoreableUses(TR::Node *node, TR::NodeChecklist& visited);
    void     findLocalObjectsValueNumbers();
    void     findLocalObjectsValueNumbers(TR::Node *node, TR::NodeChecklist& visited);
 
@@ -631,6 +633,7 @@ class TR_EscapeAnalysis : public TR::Optimization
    TR_UseDefInfo             *_useDefInfo;
    bool                      _invalidateUseDefInfo;
    TR_BitVector              *_otherDefsForLoopAllocation;
+   TR_BitVector              *_ignoreableUses;
    TR_BitVector              *_nonColdLocalObjectsValueNumbers;
    TR_BitVector              *_allLocalObjectsValueNumbers;
    TR_BitVector              *_notOptimizableLocalObjectsValueNumbers;


hzongaro:
This change is dependent upon [OMR pull request 3841](https://github.com/eclipse/omr/pull/3841) 

Vijay @vijaysun-omr, may I ask you to review this change?  Andrew @andrewcraik, may I ask you to verify my descriptions of your improvements?

hzongaro:
Force pushed to https://github.com/eclipse/openj9/commit/c31e4739a4de3b84942a7fbe1bdd8b5e39b48ffb in order to rectify copyright problem

runtime/compiler/optimizer/PreEscapeAnalysis.hpp
```diff
@@ -0,0 +1,60 @@
...
+
+typedef TR::typed_allocator<TR::Node *, TR::Region &> NodeDequeAllocator;
+typedef std::deque<TR::Node *, NodeDequeAllocator> NodeDeque;
+
+/**
+ * Pass over trees before Escape Analysis if Voluntary OSR is
+ * enabled.
+ * @see TR_EscapeAnalysis
+ */
+class TR_PreEscapeAnalysis : public TR::Optimization
```
fjeremic: The comment made in the PR would be an excellent Doxygen addition to describe what exactly this optimization does. At the moment it is pretty opaque from a code reader point of view.

hzongaro: Thanks, Filip.  I fixed this in https://github.com/eclipse/openj9/pull/5737/commits/5d03d44b02338ffaab2ab55dde3b898b38a68f14

hzongaro:
CI builds were failing because I had failed to include `PreEscapeAnalysis.cpp` and `PostEscapeAnalysis.cpp` in the list of files to be compiled with `cmake`.  Fixed in https://github.com/eclipse/openj9/pull/5737/commits/1e7477782bde1dea6682066a8ba1e0b37882d6d1

andrewcraik:
description is correct - review from @vijaysun-omr is required since I was involved with the development of this change.

runtime/compiler/optimizer/PostEscapeAnalysis.hpp
```diff
@@ -0,0 +1,55 @@
...
+#define VIRTUALGUARDHEADMERGER_INCL
+
+#include <stdint.h>                           // for int32_t
+#include "optimizer/Optimization.hpp"         // for Optimization
+#include "optimizer/OptimizationManager.hpp"  // for OptimizationManager
+
+namespace TR { class Block; }
+
+/**
+ * The new \c TR_PostEscapeAnalysis optimization performs a simple
```
fjeremic: Minor nitpick but calling this "The new \c TR_PostEscapeAnalysis optimization" will quickly go stale. When does it become not new? We should likely omit that word not to confuse people and think this is a new optimization 5 years down the road.

runtime/compiler/optimizer/EscapeAnalysis.cpp
```diff
@@ -3736,7 +3794,12 @@ void TR_EscapeAnalysis::checkEscape(TR::TreeTop *firstTree, bool isCold, bool &
       if (node->getOpCode().isCheck() || node->getOpCodeValue() == TR::treetop)
          node = node->getFirstChild();
       if (node->getOpCode().isCall() && !vnsCall.contains(node))
-         checkEscapeViaCall(node, vnsCall, ignoreRecursion);
+         {
+         if (node->getSymbolReference()->getReferenceNumber() != TR_prepareForOSR
+             || comp()->getOSRMode() != TR::voluntaryOSR
+             ||  _curBlock->isOSRInduceBlock())
```
vijaysun-omr: Does this last condition mean that we might ignore legitimate escape points in the "induce OSR" block, i.e. ones that are seen before the "induce OSR" call ?

andrewcraik: so I believe the first check says if the symref isn't prepareForOSR we will checkEscapeViaCall regardless so legitimate escapes ahead of the "induce OSR" should be fine.

vijaysun-omr: I'm still a bit unclear about this condition as a whole. a prepareForOSR call won't be a voluntaryOSR call and vice versa. So what's the point exactly of checking that it's neither one of those two calls, i.e. won't one of those conditions be true for every call ?

vijaysun-omr: Ok so the above condition basically says that we should avoid the prepare for OSR call in the OSR catch block but not in the OSR induce block (latter been added by the escape analysis pre pass) in voluntary OSR mode.

vijaysun-omr:
Looks reasonable to me. What kind of testing has been done on this change ?

runtime/compiler/optimizer/EscapeAnalysis.cpp
andrewcraik: yup that is what I believe to be what this is saying and the intent behind it.

andrewcraik:
@hzongaro can you comment on the testing?

hzongaro:
@vijaysun-omr Vijay asked,

> What kind of testing has been done on this change?

Thanks for reviewing, Vijay.  I ran internal "gold" testing (level.sanity, level.regression, level.promotion) on all platforms.

I also tested this change with [OMR pull request #3842](https://github.com/eclipse/omr/pull/3842) against the benchmark referenced in issue #2072.  Unfortunately, even after this change, there still seems to be lingering heap allocation of `Double` objects in the loop.  I'm tracking down the reason for that, but its resolution might have to be the subject of a subsequent pull request.

hzongaro:
Squashed commits down to https://github.com/eclipse/openj9/pull/5737/commits/09d669297ca33ec3f2c53c287814089e99572f4e and https://github.com/eclipse/openj9/pull/5737/commits/5c9af1b3ff8d95b02a0e8e6da00c7008e94609d1

hzongaro:
I wrote,

> Unfortunately, even after this change, there still seems to be lingering heap allocation of Double objects in the loop.  I'm tracking down the reason for that, but its resolution might have to be the subject of a subsequent pull request.

I think the lingering object allocation that I mentioned was due to a "fix" I made for an infinite recursion.  I will mark this pull request as WIP while I fix the problem, so the intended improvement to the Escape Analysis optimization isn't hobbled.

charliegracie:
Is this PR functionally correct? Does it improve the EA and the amount of objets that get stack allocated? If the answer is yes to those questions then I do not see why we do not merge this PR and the result of the investigation into further refinements can not happen in another PR.

vijaysun-omr:
@hzongaro could you please answer ?

hzongaro:
@charliegracie Charlie, I believe the PR is functionally correct, improves EA and reduces the amount of objects that get stack allocated.  My concern is that Andrew's primary motivation for these changes was to improve on the number of objects stack allocated for issue #2072.

In testing the change with the benchmark snippet that appeared in the [description from issue #2072](https://github.com/eclipse/openj9/issues/2072#issue-329487388) -- which I've reproduced below -- I verified that the objects in the loop were stack allocated with Andrew's changes.

```
public void run() {
      Double result = 1d;
      Double bound = Main.until;
      for (Double f = 0d; f < bound; f++) {
        result += f / (result + 1);
      }
      if (log) System.out.println(result);
    }
```

However, last week I went back to the original benchmark link (which now seems to be broken) in the issue.  There, the relevant code in the benchmark was slightly different:

```
    public void run() {
      final Long startTime = System.currentTimeMillis();
            
      Double result = 1d;
      Double bound = Main.until;

      for (Double f = 0d; f < bound; f++) {
        result += f / (result + 1);
      }
      
      final Long endTime = System.currentTimeMillis();
      if (log) System.out.println(result);
      if (log) System.out.println(Thread.currentThread().getName() + " finished in " + (endTime - startTime) + "ms");    
    }
```

With the addition of those boxed `Long` values, the object allocations within the loop remain there.  One reason is a fix that I made in commit eb5edefe6e2600655bf56164f64e8ccb48aaae18, which turns out to be too conservative, and ends up interfering with performing the stack allocations in this seemingly very similar method.

So, although the stack allocations now happen with the version of the code that appeared in issue #2072, the fact that there was no improvement in stack allocation using the original version of the benchmark led me to move this pull request back to _WIP_.

If you feel it's still worthwhile merging this change while the work on ensuring the objects in the original version of the benchmark can be stack allocated continues, I would be happy to remove the _WIP_ from the title.

Please accept my apologies for failing to notice this difference earlier.

andrewcraik:
@hzongaro I think it is worth putting this in. Getting the `Double` to stack allocate requires loosening a conservatively correct check to handle some additional situations and is not likely to refactor all of the other changes made here. There may be an additional enhancement to get this other version working, but what we have so far is an improvement and won't be a regression so we should get it and work on the further enhancement.

hzongaro:
OK.  Removing WIP from the title.

andrewcraik:

andrewcraik:

runtime/compiler/optimizer/EscapeAnalysis.cpp
```diff
@@ -2140,15 +2185,28 @@ bool TR_EscapeAnalysis::checkDefsAndUses(TR::Node *node, Candidate *candidate)
...
+
                   for (i = candidate->_valueNumbers->size()-1; i >= 0; i--)
                      {
                      if (candidate->_valueNumbers->element(i) == useNodeVN)
-                        break;
+                        {
+                        valueNumberSeen = true;
+
+                        // Only add this value number if it's not to be ignored
+                        if (!_ignoreableUses->get(candidate->_valueNumbers->element(i)))
```
andrewcraik: Henry caught this _ignoreableUses is currently working in terms of node global indices so this check needs to be fixed before this is merged.

vijaysun-omr: Please let me know when all changes you want to merge with this PR are done and I will run fresh testing on it.

hzongaro: This is now fixed in commit https://github.com/eclipse/openj9/pull/5737/commits/bfef953a7de444ebe4a9204198142a316d393f26

hzongaro:
With commit https://github.com/eclipse/openj9/pull/5737/commits/bfef953a7de444ebe4a9204198142a316d393f26, there is only one heap allocation remaining in the loop for the original version of the benchmark - the one for the loop index variable `f`.

However, in rerunning testing, I encountered an intermittent failure running `80 JIT_Sanity.abbs.5mins.Mode140.2` that I hadn't seen before - it occurs about 3/100 times.  It's a segmentation fault from

```
TR_OSRMethodData::getLiveRangeInfo(int) () from jre/lib/ppc64le/default/libj9jit29.so
TR_PreEscapeAnalysis::insertFakePrepareForOSR(TR::Block*, TR::Node*) () ffrom jre/lib/ppc64le/default/libj9jit29.so
TR_PreEscapeAnalysis::perform() () from jre/lib/ppc64le/default/libj9jit29.so
```

@andrewcraik Andrew, I am investigating.

andrewcraik:
@hzongaro any update on this one?

hzongaro:
@andrewcraik All I've determined thus far is that [this call to `TR_OSRCompilationData::findOSRMethodData`](https://github.com/eclipse/openj9/blob/bfef953a7de444ebe4a9204198142a316d393f26/runtime/compiler/optimizer/PreEscapeAnalysis.cpp#L95) is returning `NULL` intermittently.  Right now I'm trying to inject some debugging code and run with `traceOSR` to see whether I can spot what's going wrong.

The failure only occurs on Linux PowerPC-LE, and I've also learned that there is already an intermittent failure on the same platform running the same test.  The stack trace (see below) in that case isn't as informative, so it's difficult to tell whether the problem I'm looking at is the same problem.

```
p> 20190604-19:08:05 <gdb> Program terminated with signal 11, Segmentation fault.
p> 20190604-19:08:05 <gdb> #0  0x00003fffa025f120 in pthread_kill ()
p> 20190604-19:08:05 <gdb>    from /usr/lib64/power8/libpthread.so.0
p> 20190604-19:08:05 <gdb> 
p> 20190604-19:08:05 <gdb> backtrace:
p> 20190604-19:08:05 <gdb> #0  0x00003fffa025f120 in pthread_kill () from /usr/lib64/power8/libpthread.so.0
p> 20190604-19:08:05 <gdb> #1  0x00003fff9f25160c in omrdump_create () from /bluebird/builds/bld_418264/sdk/xl6480/jre/lib/ppc64le/default/libj9prt29.so
p> 20190604-19:08:05 <gdb> #2  0x00003fff9ef5bad0 in doSystemDump () from /bluebird/builds/bld_418264/sdk/xl6480/jre/lib/ppc64le/default/libj9dmp29.so
p> 20190604-19:08:05 <gdb> #3  0x00003fff9ef62ab0 in protectedDumpFunction () from /bluebird/builds/bld_418264/sdk/xl6480/jre/lib/ppc64le/default/libj9dmp29.so
p> 20190604-19:08:05 <gdb> #4  0x00003fff9f1fe49c in omrsig_protect () from /bluebird/builds/bld_418264/sdk/xl6480/jre/lib/ppc64le/default/libj9prt29.so
p> 20190604-19:08:05 <gdb> #5  0x00003fff9ef6278c in runDumpAgent () from /bluebird/builds/bld_418264/sdk/xl6480/jre/lib/ppc64le/default/libj9dmp29.so
p> 20190604-19:08:05 <gdb> #6  0x00003fff9ef8d8e8 in triggerDumpAgents () from /bluebird/builds/bld_418264/sdk/xl6480/jre/lib/ppc64le/default/libj9dmp29.so
p> 20190604-19:08:05 <gdb> #7  0x00003fff9f3ef7b0 in generateDiagnosticFiles () from /bluebird/builds/bld_418264/sdk/xl6480/jre/lib/ppc64le/default/libj9vm29.so
p> 20190604-19:08:05 <gdb> #8  0x00003fff9f1fe49c in omrsig_protect () from /bluebird/builds/bld_418264/sdk/xl6480/jre/lib/ppc64le/default/libj9prt29.so
p> 20190604-19:08:05 <gdb> #9  0x00003fff9f3eee48 in structuredSignalHandler () from /bluebird/builds/bld_418264/sdk/xl6480/jre/lib/ppc64le/default/libj9vm29.so
p> 20190604-19:08:05 <gdb> #10 0x00003fff9f20229c in masterSynchSignalHandler$AF34_16 () from /bluebird/builds/bld_418264/sdk/xl6480/jre/lib/ppc64le/default/libj9prt29.so
p> 20190604-19:08:05 <gdb> #11 <signal handler called>
p> 20190604-19:08:05 <gdb> #12 0x00003ffe8f5668d8 in ?? ()
p> 20190604-19:08:05 <gdb> #13 0x00003ffe8f565ee4 in ?? ()
p> 20190604-19:08:05 <gdb> #14 0x00003fff9f3d383c in runJavaThread () from /bluebird/builds/bld_418264/sdk/xl6480/jre/lib/ppc64le/default/libj9vm29.so
p> 20190604-19:08:05 <gdb> #15 0x00003fff9f44b398 in javaProtectedThreadProc () from /bluebird/builds/bld_418264/sdk/xl6480/jre/lib/ppc64le/default/libj9vm29.so
p> 20190604-19:08:05 <gdb> #16 0x00003fff9f1fe49c in omrsig_protect () from /bluebird/builds/bld_418264/sdk/xl6480/jre/lib/ppc64le/default/libj9prt29.so
p> 20190604-19:08:05 <gdb> #17 0x00003fff9f44b190 in javaThreadProc () from /bluebird/builds/bld_418264/sdk/xl6480/jre/lib/ppc64le/default/libj9vm29.so
p> 20190604-19:08:05 <gdb> #18 0x00003fff9f2e9914 in thread_wrapper () from /bluebird/builds/bld_418264/sdk/xl6480/jre/lib/ppc64le/default/libj9thr29.so
p> 20190604-19:08:05 <gdb> #19 0x00003fffa0258728 in start_thread () from /usr/lib64/power8/libpthread.so.0
p> 20190604-19:08:05 <gdb> #20 0x00003fffa0109240 in clone () from /usr/lib64/power8/libc.so.6
```

hzongaro:
@andrewcraik Andrew, in looking at a failure log with `traceOSR` for the problem mentioned in [my comment above](#issuecomment-497373266), I notice many messages in the log for the problematic method indicating `OSR induction cannot be performed after OSR infrastructure has been removed`.  Would that ultimately result in `TR_OSRCompilationData::findOSRMethodData` returning NULL?

I notice that some places that check `comp()->getOSRMode() == TR::voluntaryOSR` check it in conjunction with `comp()->supportsInduceOSR()`, which in turn checks, in part, whether OSR infrastructure has been removed.  Would the correct thing to do be to add a check of `supportsInduceOSR` to [`TR_PreEscapeAnalysis::perform`](https://github.com/eclipse/openj9/blob/bfef953a7de444ebe4a9204198142a316d393f26/runtime/compiler/optimizer/PreEscapeAnalysis.cpp#L33-L35), and other places in these changes that just check the OSR Mode?

andrewcraik:
The note about induction is likely not the cause of the null, but I'll need to look at the log to be sure. No the correct thing to do is not to check supportsInduceOSR - we are just trying to move an existing one and something is going wrong doing that.

I'll discuss with you offline

vijaysun-omr:
@hzongaro is this close to delivery now ?

hzongaro:
Hi, Vijay @vijaysun-omr - I'm still working on tracking down the source of the intermittent crash from `TR_PreEscapeAnalysis::perform`.  It's difficult to estimate how quickly I can find the source of that problem, but it may well be a preexisting problem that this pull request happens to trip over.

hzongaro:
Sorry for the long delay on updating this.  I've marked this change as a work in progress.  As was mentioned in https://github.com/eclipse/openj9/pull/5737#issuecomment-497373266, there is intermittent failure that occurs with this change.  The problem is in unrelated code, but happens to exposed by this change.

OpenJ9 pull request #6626 and [OMR pull request #4177](https://github.com/eclipse/omr/pull/4177) address that problem.  Once they are delivered, I will remove the **WIP** prefix from this pull request.

hzongaro:
Just force pushed rebasing the branch on the latest master.  No changes otherwise.

andrewcraik:

andrewcraik:
have started sanity since the sequence of changes needed to fix the OSR metadata bug exposed by this work has now landed.

hzongaro:
Removed WIP prefix, because as Andrew mentioned, the bug this change exposes has been fixed.

andrewcraik:
@hzongaro I don't think we should merge this until the new Pre/Post are added to the opt strategy in this commit - the danger is that we would keep running regular EA and ignore the 'real' prepares thinking the fake prepares will save us, but we won't have added them

hzongaro:
> @hzongaro I don't think we should merge this until the new Pre/Post are added to the opt strategy in this commit - the danger is that we would keep running regular EA and ignore the 'real' prepares thinking the fake prepares will save us, but we won't have added them

Thanks, Andrew @andrewcraik.  So I think that means that I need to separate out the part of this change that adds PreEscapeAnalysis and PostEscapeAnalysis.  Then after that's merged, [OMR pull request #3842](https://github.com/eclipse/omr/pull/3842), which adds them to the optimization strategy can be merged, and finally, the remainder of this change, with the actual EA changes, can be delivered.  Do I have that right?

andrewcraik:
Yes I think so - we can't change main EA until we have pre/post running, but it is safe to run pre/post with existing EA - you just see more escapes (escapes that would have happened anyway)

hzongaro:
Reorganized changes so that PreEscapeAnaysis and PostEscapeAnalysis are available and running when the Escape Analysis changes are delivered, as Andrew recommends above.  This makes this change now dependent upon pull request #6654 and [OMR pull request #3842](https://github.com/eclipse/omr/pull/3842).

hzongaro:
Just force pushed to bring the `ea-staging-branch-step2` up to date with changes from pull request #6654 - no substantive changes here.

andrewcraik:
@hzongaro now that #6654 has merged can you rebase so we can sanity this one?

hzongaro:
Andrew @andrewcraik, I've brought this branch up to date with the latest changes in OpenJ9.  Before this change is sanity tested, we'll need [OMR pull request 3842](https://github.com/eclipse/omr/pull/3842) to make it through the OpenJ9 OMR acceptance testing.  I will leave this marked as a work in progress until then.

andrewcraik:
Jenkins test sanity all jdk8,jdk11

PR #5743, opened 2019-05-13
Update VT test default values

tajila:
Update VT test default values

Update test default values to make it easier to see where fields 
start and end when reading raw memory.

Add helper to generate a core dump

### DIFF
--- a/test/functional/Valhalla/src/org/openj9/test/lworld/ValueTypeTests.java
+++ b/test/functional/Valhalla/src/org/openj9/test/lworld/ValueTypeTests.java
@@ -107,21 +107,25 @@ public class ValueTypeTests {
 	static MethodHandle[] getObjects = null;
 	
 	/* default values */
-	static int[] defaultPointPositions = {0xFFEEFFEE, 0xAABBAABB};
-	static int[][] defaultLinePositions = {defaultPointPositions, defaultPointPositions};
-	static int[][][] defaultTrianglePositions = {defaultLinePositions, defaultLinePositions, defaultLinePositions};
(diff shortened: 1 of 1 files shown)


tajila:
@llxia Please review these changes

test/functional/Valhalla/src/org/openj9/test/lworld/ValueTypeTests.java
```diff
@@ -1513,4 +1523,8 @@ static void checkEqualLargeObject(Object largeObject, Object contentObject) thro
 			assertEquals(getObject.invoke(getObjects[i].invoke(largeObject)), contentObject);
 		}
 	}
+	
+	static void checkObject(Object ...objects) {
```
llxia: - Why is the method called `checkObject`, but it calls `SystemDump()`? And the parameter `objects ` is not used.
- Will we use this method in the future?

tajila: @llxia 

The purpose of this method is to make it easy to find the objects in DDR when the system core is produced. This method lets one put an object (or many objects) in the argument section of the method frame. When the SystemDump is produced, the object will be in the stack frame and it will be easy to find it in DDR since it will be in the arguments section.  

The alternative approach is to inspect the bytecodes and determine which temp/local slot the object of interest is located in. This approach is more difficult and time consuming then having the objects in known argument slots.

This method is only used for debugging purposes.

llxia:
Will forgot the PR builds as they will not test TV test.

PR #5753, opened 2019-05-14
Create new protobuf message for resolved methods

dmitry-ten:
One problem with how we process regular protobuf messages
is that we cannot send non-trivially copyable structs/classes
over protobuf, i.e. structs containing strings, vectors, maps.
This leads to awkward workarounds, such as putting all primitive
attributes in a struct, and then wrapping it in a tuple
with variable length attributes.
One way to overcome this limitation is by defining a new protobuf message
for a non-trivially copyable struct, then writing
serialization/deserialization code from/to JIT internal struct to/from class generated by protobuf.
This is what I've done for `TR_ResolvedJ9JITaaSServerMethodInfo`.

Another solution would be to not define a new protobuf message, and
instead write code to directly serialize/deserialize JIT internal struct
to/from an array of bytes.

### DIFF
--- a/runtime/compiler/control/JITaaSCompilationThread.cpp
+++ b/runtime/compiler/control/JITaaSCompilationThread.cpp
@@ -1595,7 +1595,7 @@ bool handleServerMessage(JITaaS::J9ClientStream *client, TR_J9VM *fe)
          TR_ResolvedJ9JITaaSServerMethod::createResolvedMethodFromJ9MethodMirror(methodInfo, (TR_OpaqueMethodBlock *) j9method, vTableSlot, mirror, fe, trMemory);
 
          // Collect AOT stats
-         TR_ResolvedJ9Method *resolvedMethod = std::get<0>(methodInfo).remoteMirror;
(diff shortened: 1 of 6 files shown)


dmitry-ten:
@ymanton could you review?
I do not really like that for serialization/deserialization every attribute needs to be set and cast manually, but can't think of a better way to do it, other than the other approach I mentioned in the commit message.

ymanton:
I don't have a strong opinion on this either way. On the one hand specifying messages and their payloads in the protocol is better than making it be implicit in the code, it serves as documentation, we can build other tools that can read/write the same data, etc. On the other hand doing this work halfway is going to leave us with two implementations for a while since I don't think there is any motivation to convert every message in the code base.

Can you remind me what the measured CPU and memory consumption differences were before/after this change?

dmitry-ten:
There was no difference in CPU or memory consumption.

ymanton:
Then I think we should close this and revisit it if/when we decide to implement all messages this way.

dmitry-ten:
Alright.

Issue #5746, opened 2019-05-13
JITaaS compilation failures in JITaaS PR testing


harryyu1994:
```
16:06:26  /home/jenkins/workspace/Build_JDK8_x86-64_linux_Personal/build/linux-x86_64-normal-server-release/vm/compiler/../compiler/rpc/J9Server.cpp:111:8: error: ‘SSL_CTX_set_ecdh_auto’ was not declared in this scope
16:06:26      if (SSL_CTX_set_ecdh_auto(ctx, 1) != 1)
16:06:26 
```
```
16:06:26  /usr/include/openssl/bio.h:668:6: note:   initializing argument 1 of ‘BIO* BIO_new_mem_buf(void*, int)’
16:06:26   BIO *BIO_new_mem_buf(void *buf, int len);
16:06:26        ^~~~~~~~~~~~~~~
16:06:26  /home/jenkins/workspace/Build_JDK8_x86-64_linux_Personal/build/linux-x86_64-normal-server-release/vm/compiler/../compiler/rpc/J9Server.cpp:148:35: error: invalid conversion from ‘const void*’ to ‘void*’ [-fpermissive]
16:06:26      BIO *certMem = BIO_new_mem_buf(&sslCerts[0][0], sslCerts[0].size());
```

harryyu1994:
On it

a7ehuo:
Likely the build still searches the default `/usr/include/` path instead of `/usr/local/include` which is where openssl-1.1.1 is installed, or search where the openssl is fetched and installed. 
```16:08:36  In file included from /usr/include/openssl/err.h:124:0```

This issue is related to issue #5720

Looks like the header versions that are picked up are prior to 1.0.2 
```
SSL_CTX_set_ecdh_auto
- Does not exist in 1.0.0, 1.0.1
- Declared in 1.0.2

BIO_new_mem_buf
- In 1.0.0 and 1.0.1
    - crypto/bio/bio.h:BIO *BIO_new_mem_buf(void *buf, int len);
- 1.0.2
    - include/openssl/bio.h:BIO *BIO_new_mem_buf(const void *buf, int len);
```

harryyu1994:
Since SSL_CTX_set_ecdh_auto is deprecated, we should remove it. The rest of the issue should be fixed by #5720

PR #5755, opened 2019-05-14
Remove deprecated SSL_CTX_set_ecdh_auto calls

harryyu1994:
JITaaS clients and servers use this openssl api which has been deprecated

Issue: #5746
[skip ci]

### DIFF
--- a/runtime/compiler/rpc/J9Client.cpp
+++ b/runtime/compiler/rpc/J9Client.cpp
@@ -62,12 +62,14 @@ void J9ClientStream::static_init(TR::PersistentInfo *info)
       exit(1);
       }
 
+#if OPENSSL_VERSION_NUMBER < 0x010100000L
    if (SSL_CTX_set_ecdh_auto(ctx, 1) != 1)
       {
       perror("failed to configure SSL ecdh");
       ERR_print_errors_fp(stderr);
       exit(1);
       }
+#endif
 
    auto &sslKeys = info->getJITaaSSslKeys();
    auto &sslCerts = info->getJITaaSSslCerts();
--- a/runtime/compiler/rpc/J9Server.cpp
+++ b/runtime/compiler/rpc/J9Server.cpp
@@ -108,12 +108,14 @@ SSL_CTX *createSSLContext(TR::PersistentInfo *info)
 
    SSL_CTX_set_session_id_context(ctx, (const unsigned char*) "JITaaS", 6);
 
+#if OPENSSL_VERSION_NUMBER < 0x010100000L
    if (SSL_CTX_set_ecdh_auto(ctx, 1) != 1)
       {
       perror("failed to configure SSL ecdh");
       ERR_print_errors_fp(stderr);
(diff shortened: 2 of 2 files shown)


a7ehuo:
SSL_CTX_set_ecdh_auto  is not deprecated, it's redeclared in 1.0.2 and later.

harryyu1994:
```
In order to use ECDHE_* cipher suites, OpenSSL < 1.1.0 require apps to call SSL_CTX_set_ecdh_auto(), while OpenSSL >= 1.1.0 deprecated this [1] in favour of automatic ECDH context setting with said function implemented as just noop macro [2].
```
I guess if we want to support 1.0.2 then we can't remove it. It's deprecated after 1.1.0

ymanton:
OK, so our difficulty here is the time period between when the function was removed and when the compat macro was added.

Ideally we would check at configure time if a `SSL_CTX_set_ecdh_auto` symbol was available or not (regardless of whether it's a function or a macro, we don't care) and call it if it is detected, otherwise not. However I don't think our build system is capable of that yet. The next best thing might be to check `OPENSSL_VERSION_NUMBER`[1] at compile time instead and call it for versions older than 1.1.0.

So I propose something like this to guard the call:

```
#if OPENSSL_VERSION_NUMBER < 0x010100000L
SSL_CTX_set_ecdh_auto(...)
#endif
```

[1] https://linux.die.net/man/3/openssl_version_number

harryyu1994:
Since they have noop macro for version 1.1.0 and later, do we even need this change? I've put the change in anyways. Ready for review.

ymanton:
> Since they have noop macro for version 1.1.0 and later, do we even need this change? I've put the change in anyways. Ready for review.

Yes, because there are versions of openssl that remove the function but don't have the macro. That's why we saw the original build failure, some version of openssl on a RHEL machine didn't have the function and didn't have the compat macro.

This sort of thing happens because distributions can modify their builds of openssl. This is especially common on RHEL in my experience because they ship older versions of some libraries but backport some changes.

ymanton:
I think I've managed to reproduce the build failure on RHEL 6.9, where the distro version of OpenSSL is 1.0.1e and `OPENSSL_VERSION_NUMBER` is defined as:

```
#define OPENSSL_VERSION_NUMBER  0x1000105fL
```

`SSL_CTX_set_ecdh_auto` however is not defined, despite what the OpenSSL documentation says, so I don't think this PR would fix the problem. It seems like RedHat has backported some changes to their OpenSSL but not others, so it doesn't follow exactly what's in upstream OpenSSL. The sensible solution is to have the build system detect whether or not the symbol is defined, but we can't do that at the moment.

Once https://github.com/eclipse/openj9/issues/5720 is merged we'll hopefully be able to build against a later version of OpenSSL that comes directly from upstream rather than the distro, so I think we should just close this and not bother supporting this particular build of OpenSSL.

harryyu1994:
Okay closing

PR #5758, opened 2019-05-14
Fix illegal variable shadow

gacholio:
Various functions redeclare "result" in a sub-scope, resulting in an invalid value being returned.

This hasn't been an issue as this code is only used in the always call
access barrier case, which is never tested.

Also fix some local declarations that don't follow the coding standard.

[ci skip]

### DIFF
--- a/runtime/gc_base/ObjectAccessBarrier.cpp
+++ b/runtime/gc_base/ObjectAccessBarrier.cpp
@@ -585,9 +585,8 @@ void *
 MM_ObjectAccessBarrier::mixedObjectReadAddress(J9VMThread *vmThread, J9Object *srcObject, UDATA srcOffset, bool isVolatile)
 {
 	void **actualAddress = J9OAB_MIXEDOBJECT_EA(srcObject, srcOffset,void *);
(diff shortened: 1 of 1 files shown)


amicic:
same problem line 1754?

youngar:
[Looks like it](https://github.com/eclipse/openj9/pull/5758/files#diff-15dd2807c747f452e8dc9a91db02566eL1754)

gacholio:
I was only looked for ifdefs for compressed - I'll scan the whole file.

gacholio:
Fixed that one, didn't see any more,

amicic:
Nothing incorrect, but not following coding standards (the declaration could be dragged down to the only assignment statement). A few of them, here is one: https://github.com/eclipse/openj9/blob/b385c20bf51b9cd1138a7b5ea42118251163b6d9/runtime/gc_base/ObjectAccessBarrier.cpp#L624

gacholio:
I have fixed the uninitialized variables.

amicic:
Jenkins compile win32 jdk8

Issue #5582, opened 2019-04-23
Cache resolved methods from TR_ResolvedJ9JITaaSServerMethod::getResolvedImproperInterfaceMethod


dmitry-ten:
**Problem:**
JITaaS server creates resolved methods, as described [here](https://github.com/eclipse/openj9/blob/jitaas/doc/compiler/jitaas/ResolvedMethod.md).
Creating a resolved method without any caching strategies requires server to make a remote call to the client. During compilation, lots of resolved methods are usually created. For example, inliner creates tons of resolved methods to determine whether a function call should be inlined.
We want to minimize the number of remote calls, thus we currently cache most types of resolved methods.
Examples of resolved methods currently cached:
```
TR_ResolvedJ9JITaaSServerMethod::getResolvedPossiblyPrivateVirtualMethod
TR_ResolvedJ9JITaaSServerMethod::getResolvedStaticMethod
TR_ResolvedJ9JITaaSServerMethod::getResolvedSpecialMethod
...
etc.
```
**Relevant Files:**
```
env/j9methodServer.cpp
env/j9methodServer.hpp
control/JITaaSCompilationThread.cpp
control/JITaaSCompilationThread.hpp
rpc/protos/compile.proto
```
**Proposed solution:**
We should use the same caching strategy for `getResolvedImproperInterfaceMethod`.
Look at the code for resolved methods mentioned above for reference.

dmitry-ten:
This could be a beginner issue for new interns.
@mpirvu could you add `jitaas` tag to this? I'll update the description, once any proposed changes to resolved methods caching are done.

dmitry-ten:
Edited description to be more detailed.

PR #5760, opened 2019-05-14
Add caching for improper resolved methods

chrisc66:
Add caching for improper resolved methods for
getResolvedImproperInterfaceMethod function.

Resolves: #5582

Signed-off-by: Chris <zichun.chong@mail.utoronto.ca>

### DIFF
--- a/runtime/compiler/control/JITaaSCompilationThread.cpp
+++ b/runtime/compiler/control/JITaaSCompilationThread.cpp
@@ -4407,8 +4407,13 @@ TR::CompilationInfoPerThreadRemote::getCachedResolvedMethod(TR_ResolvedMethodKey
       {
       auto comp = getCompilation();
       auto methodCacheEntry = it->second;
-      auto methodInfo = methodCacheEntry.methodInfo;
       TR_OpaqueMethodBlock *method = methodCacheEntry.method;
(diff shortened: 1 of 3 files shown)


chrisc66:
@ymanton could you please review this commit

runtime/compiler/control/JITaaSCompilationThread.cpp
```diff
@@ -4410,6 +4410,11 @@ TR::CompilationInfoPerThreadRemote::getCachedResolvedMethod(TR_ResolvedMethodKey
       auto methodInfo = methodCacheEntry.methodInfo;
       TR_OpaqueMethodBlock *method = methodCacheEntry.method;
       uint32_t vTableSlot = methodCacheEntry.vTableSlot;
+      if (!method)
+         {
+         *resolvedMethod = NULL;
+         return true;
+         }
```
ymanton: I'd suggest re-ordering this so that it's a little easier to read. `methodInfo` and `vTableSlot` are only used below the code you're adding, so it's better to define them after the `if (!method) { ... }` block.

chrisc66:
@ymanton Thanks for the comment, I have updated this commit. Could you review it again?

PR #5772, opened 2019-05-15
Ensure Jenkins reconnects slaves after their sanitation

vsebe:
Add waiting times on slave disconnect and reconnect. On error retry to
reconnect.
Attempt to reconnect nodes disconnected by system.
Check slaves status after sanitation only.

[ci skip]

### DIFF
--- a/buildenv/jenkins/jobs/infrastructure/Cleanup-Nodes
+++ b/buildenv/jenkins/jobs/infrastructure/Cleanup-Nodes
@@ -19,6 +19,7 @@
  *
  * SPDX-License-Identifier: EPL-2.0 OR Apache-2.0 OR GPL-2.0 WITH Classpath-exception-2.0 OR LicenseRef-GPL-2.0 WITH Assembly-exception
  *******************************************************************************/
+import hudson.slaves.OfflineCause
 
 defaultSetupLabel = 'worker'
 defaultLabel = 'ci.role.build || ci.role.test'
@@ -123,8 +124,14 @@ timeout(time: TIMEOUT_TIME.toInteger(), unit: TIMEOUT_UNITS) {
 
                     if (aNode.toComputer().isOffline()) {
                         // skip offline slave
-                        def offlineCause = aNode.toComputer().getOfflineCause().toString()
-                        offlineSlaves.put(nodeName, offlineCause)
+                        def offlineCause = aNode.toComputer().getOfflineCause()
+                        if (offlineCause instanceof OfflineCause.UserCause) {
+                            // skip offline node disconnected by users
+                            offlineSlaves.put(nodeName, offlineCause.toString())
+                        } else {
+                            // cache nodes, will attempt to reconnect nodes disconnected by system later
+                            buildNodes.add(nodeName)
+                        }
                         continue
                     }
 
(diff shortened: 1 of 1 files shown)


vsebe:
[personal build](https://ci.eclipse.org/openj9/view/Sandbox/job/vs_Cleanup-Nodes/94/console)

@AdamBrousseau please review, thank you!

buildenv/jenkins/jobs/infrastructure/Cleanup-Nodes
```diff
@@ -281,13 +297,33 @@ def sanitize_node(nodeName) {
         println(e.getMessage())
     }
 
+    //reconnect slave
     timeout(time: RECONNECT_TIMEOUT.toInteger(), unit: 'MINUTES') {
-        println("\t ${nodeName}: Reconnecting...")
+        println("\t ${nodeName}: Disconnecting...")
+        waitUntil {
+            // wait for Jenkins to disconnect the slave after kill -9 or reboot
+            return (workingComputer.getNode().getRootPath() == null)
```
AdamBrousseau: Technically if the node is back online before we get to this point we would get stuck in this loop. I very highly doubt that would happen though.

buildenv/jenkins/jobs/infrastructure/Cleanup-Nodes
```diff
@@ -281,13 +297,33 @@ def sanitize_node(nodeName) {
...
+            }
+            reconnected = true
+
+            workingComputer.connect(false)
+            waitUntil {
+                // wait for slave agent to relaunch
+                return (workingComputer.getChannel() != null)
+            }
+            workingComputer.setTemporarilyOffline(false, null)
+            workingComputer.waitUntilOnline()
```
AdamBrousseau: What is the difference between `waitUntilOnline()` and `waitUntil { getChannel() != null }` ?

buildenv/jenkins/jobs/infrastructure/Cleanup-Nodes
```diff
@@ -281,13 +297,33 @@ def sanitize_node(nodeName) {
...
         workingComputer.disconnect(null)
         workingComputer.waitUntilOffline()
-        workingComputer.connect(false)
-        workingComputer.setTemporarilyOffline(false, null)
-        workingComputer.waitUntilOnline()
-        println("${nodeName} is back online: ${workingComputer.isOnline()}")
+
+        println("\t ${nodeName}: Connecting...")
+        def reconnected = false
+        retry(2) {
```
AdamBrousseau: Under what condition would we hit the second retry? I'm not sure if `reconnected` is a good name as it appears to be only used for "If this is the second try, then we should wait 30 seconds".

buildenv/jenkins/jobs/infrastructure/Cleanup-Nodes
vsebe: thus the timeout

buildenv/jenkins/jobs/infrastructure/Cleanup-Nodes
vsebe: [waitUntilOnline()](https://javadoc.jenkins-ci.org/hudson/model/Computer.html#waitUntilOnline--) Blocks until the node becomes online. 
`connect()` returns immediately ( `isOnline()` returns true), but the relaunch operation happens asynchronously. `waitUntil` waits on the relaunch operation.

buildenv/jenkins/jobs/infrastructure/Cleanup-Nodes
vsebe: intermittent connection issues, jenkins bugs, I don't expect to happen often, might be overkill

buildenv/jenkins/jobs/infrastructure/Cleanup-Nodes
AdamBrousseau: I'm still confused. It looks to me that these do the same thing
```
waitUntil {
    // wait for slave agent to relaunch
    return (workingComputer.getChannel() != null)
}
```
```
workingComputer.waitUntilOnline()
```

buildenv/jenkins/jobs/infrastructure/Cleanup-Nodes
AdamBrousseau: I'm torn on adding it as insurance or removing it as unnecessary code.

vsebe: I added a reconnect in the `finally` clause, we can dispense of retry

vsebe:
@pshipton please merge, thx

PR #5783, opened 2019-05-16
More runtime checks for compressed refs

gacholio:
Update some more places in the VM with runtime checks for compressed
mode.

Add a GC configuration query for compressed refs. Add a runtime query
for the JIT in J9ObjectModel.

[ci skip]

### DIFF
--- a/runtime/compiler/env/J9ObjectModel.hpp
+++ b/runtime/compiler/env/J9ObjectModel.hpp
@@ -49,6 +49,7 @@ public:
 
    ObjectModel() :
       OMR::ObjectModelConnector(),
+      _compressObjectReferences(false),
       _usesDiscontiguousArraylets(false),
       _arrayLetLeafSize(0),
       _arrayLetLeafLogSize(0),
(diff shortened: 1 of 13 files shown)


gacholio:
jenkins test sanity xlinux,xlinuxlargeheap jdk8

gacholio:
Related: #643 

I have previously verified with @fjeremic that the JIT change is the way we want to go.

runtime/compiler/env/J9ObjectModel.hpp
```diff
@@ -121,8 +122,14 @@ class ObjectModel : public OMR::ObjectModelConnector
    */
    MM_GCWriteBarrierType writeBarrierType() { return _writeBarrierType; }
 
+    /**
+   * @brief: Returns whether or not object references are compressed
+   */
```
fjeremic: Indentation, also I don't think Doxygen will like the colon after the brief tag.

gacholio: I believe I copied this from a few lines above...

gacholio: Looks like I did get an extra space, but the colon is definitely there elsewhere.

gacholio: I'll fix all of the colons after testing completes.

gacholio:
jenkins test sanity xlinux,xlinuxlargeheap jdk8

DanHeidinga:
Jenkins compile win32 jdk8

DanHeidinga:
Launched a win32 compile to ensure all symbols are visible for 32 bit builds

Issue #5741, opened 2019-05-13
extended.functional testDDRExt_SharedClasses_openj9_0 output does not contain success key : java/util/HashMap


pshipton:
https://ci.eclipse.org/openj9/job/Test_openjdk11_j9_extended.functional_x86-64_linux_Nightly/29/tapResults/
```
     [java] [AutoRun] [INFO] Starting testShrcFindJITHext(j9vm.test.ddrext.junit.TestSharedClassesExt)
     [java] [DDRExtTesterBase] [ERROR] '!shrc  jithstats' output does not contain success key : java/util/HashMap
     [java] [DDRExtTesterBase] [ERROR] '!shrc  jithstats' output :!j9sharedclassconfig 0x00007EFF480E3B30
     [java] 
     [java] Meta data region to be used: 0x00007EFF067E5F60..0x00007EFF06800FFC
     [java] !j9sharedcacheheader 0x00007EFEF54000F0
     [java] cache size       : 314544128
     [java] free bytes       : 63963064
     [java] softmx bytes     : 67108864
     [java] read write area  : 0x00007EFEF5400260 - 0x00007EFEF5400260 size 28432 used 0
     [java] segment area     : 0x00007EFEF5407000 - 0x00007EFEF56602E8 size 2462440
     [java] metadata area    : 0x00007EFF067E5F60 - 0x00007EFF06801000 size 110752
     [java] class debug area : 0x00007EFF06801000 - 0x00007EFF08000000 size 25161728 used 543936 (% used 2)
     [java] 1: 0x00007EFF067FE340 JITHINT data !j9x 0x00007EFF067FE358,0x0000000000000004  type 2 updates 0 corrupt -1
     [java] 	getCharFromArrayByIndex(Ljava/lang/Object;I)C !j9rommethod 0x00007EFEF54252B8
     [java] 	com/ibm/jit/JITHelpers !j9romclass 0x00007EFEF5423E88
     [java] 1: 0x00007EFF067FC028 JITHINT data !j9x 0x00007EFF067FC040,0x0000000000000004  type 2 updates 0 corrupt -1
     [java] 	putCharInArrayByIndex(Ljava/lang/Object;IC)V !j9rommethod 0x00007EFEF54251D8
     [java] 	com/ibm/jit/JITHelpers !j9romclass 0x00007EFEF5423E88
[...]

PR #5788, opened 2019-05-16
Remove modified=mod1 in DDR tests

llxia:
Fixes: #5741

### DIFF
--- a/test/functional/DDR_Test/tck_ddrext.xml
+++ b/test/functional/DDR_Test/tck_ddrext.xml
@@ -114,11 +114,11 @@ SPDX-License-Identifier: EPL-2.0 OR Apache-2.0 OR GPL-2.0 WITH Classpath-excepti
 		<echo>Using JVM : ${sdk.executable}${sdk.executable.ext}</echo>
 		<echo>classname = "j9vm.test.corehelper.StackMapCoreGenerator"</echo>
 		<echo>Java VM Args:</echo>
-		<echo>	jvmarg = -Xshareclasses:name=ddrextjunitSCC,reset,modified=mod1</echo>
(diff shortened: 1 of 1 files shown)


llxia:
jenkins test extended xlinux jdk8,jdk11

test/functional/DDR_Test/tck_ddrext.xml
```diff
@@ -114,11 +114,11 @@ SPDX-License-Identifier: EPL-2.0 OR Apache-2.0 OR GPL-2.0 WITH Classpath-excepti
 		<echo>Using JVM : ${sdk.executable}${sdk.executable.ext}</echo>
 		<echo>classname = "j9vm.test.corehelper.StackMapCoreGenerator"</echo>
 		<echo>Java VM Args:</echo>
-		<echo>	jvmarg = -Xshareclasses:name=ddrextjunitSCC,reset,modified=mod1</echo>
+		<echo>	jvmarg = -Xshareclasses:name=ddrextjunitSCC,reset -Xitsn1000</echo>
```
pshipton: Why is `-Xitsn1000` added?

pshipton:
It would be nice to understand why the `modified=mod1` was there, but I can't find anything that depends on it.

test/functional/DDR_Test/tck_ddrext.xml
llxia: we are running `-Xitsn1000` (see line 122). 

I added it into the `echo` so it prints out the exact jvm options that we use in the console output.

llxia:
> It would be nice to understand why the `modified=mod1` was there, but I can't find anything that depends on it.

Yes, we could not find anything that uses it. And the whole test suite passed without `modified=mod1` option.

Issue #5720, opened 2019-05-09
Build JITaaS with the same OpenSSL as the rest of OpenJ9


ymanton:
OpenJ9 can now be built with OpenSSL support. When running `configure` you can specify `--with-openssl=...` to choose which set of OpenSSL headers and libraries to build against. The options are system, fetched, and a path to an OpenSSL source tree. See here for details: https://github.com/ibmruntimes/openj9-openjdk-jdk8/blob/6186c40a2eb7bf12dc0139e349229f703e5f4194/jdk/make/closed/autoconf/custom-hook.m4#L507

When building the JIT with JITaaS support we are unaware of that option and look for OpenSSL headers and libs in the default places, which is equivalent to `--with-openssl=system`. The JIT should instead become aware of that option and build against the same OpenSSL that the rest of the VM is building against.

a7ehuo:
I will look at this one.

a7ehuo:
We can add `OPENSSL_CFLAGS`, `OPENSSL_LIBS`, and `OPENSSL_BUNDLE_LIB_PATH` to `CUSTOM_COMPILER_ENV_VARS` which can be picked up when building JIT. However. `--with-openssl=fetched` will not set `OPENSSL_LIBS`. `OPENSSL_LIBS` will only be set if -`-with-openssl=system` is specified. More details as below:
```
# --with-openssl=fetched <= Jenkins PR and nightly builds use this option
OPENSSL_DIR is /root/home/ahuo/src/openj9-openjdk-jdk8/openssl
OPENSSL_CFLAGS is -I/root/home/ahuo/src/openj9-openjdk-jdk8/openssl/include
OPENSSL_BUNDLE_LIB_PATH is 
OPENSSL_LIBS is 
```
```
# --with-openssl=fetched + --enable-openssl-bundling
OPENSSL_DIR is /root/home/ahuo/src/openj9-openjdk-jdk8/openssl
OPENSSL_CFLAGS is -I/root/home/ahuo/src/openj9-openjdk-jdk8/openssl/include
OPENSSL_BUNDLE_LIB_PATH is /root/home/ahuo/src/openj9-openjdk-jdk8/openssl
OPENSSL_LIBS is 
```
```
# --with-openssl=system
# The option --enable_openssl_bundling is not available with --with-openssl=system. Use option fetched or openssl path to bundle crypto library

OPENSSL_DIR is 
OPENSSL_CFLAGS is -I/usr/local/include
[...]

Issue #303, opened 2017-10-12
OpenJ9 sanity tests fail on OpenJ9 build with OpenJDK10 (18.3)


groeges:
When running the Sanity tests against a build of OpenJ9 with  OpenJDK10 (18..3) the following test suites fail.
```
FAILED test targets:
	JCL_Test_SE90_0
	JCL_Test_SE90_1
```
These fail on these tests:
```
test_getProperties
test_getProperty
test_getProperty2
```
because the tests retrieve the system properties and check the java.version. 
The tests are looking for java.version starting with 9 where the java.version returned from this jdk is '10-internal'. 

Either the test frameworks needs to detect this new jdk version and run new tests or these tests need to be updated to work with both jdk9 and jdk10.

Exact test failure for one of the tests is:
```
test_getProperty2
java.lang.AssertionError: Failed to return correct property value: 10-internal
	at org.testng.AssertJUnit.fail(AssertJUnit.java:59) from jdk.internal.loader.ClassLoaders$AppClassLoader@e62d135f(file:/root/openj9-openjdk-jdk9/openj9/jvmtest/TestConfig/lib/testng.jar)
	at org.testng.AssertJUnit.assertTrue(AssertJUnit.java:24) from jdk.internal.loader.ClassLoaders$AppClassLoader@e62d135f(file:/root/openj9-openjdk-jdk9/openj9/jvmtest/TestConfig/lib/testng.jar)
	at org.openj9.test.java.lang.Test_System.test_getProperty2(Test_System.java:220) from jdk.internal.loader.ClassLoaders$AppClassLoader@e62d135f(file:/root/openj9-openjdk-jdk9/openj9/jvmtest/Java8andUp/GeneralTest.jar)
	at jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(java.base@10-internal/Native Method) from jrt:/java.base
	at jdk.internal.reflect.NativeMethodAccessorImpl.invoke(java.base@10-internal/NativeMethodAccessorImpl.java:62) from jrt:/java.base
	at jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(java.base@10-internal/DelegatingMethodAccessorImpl.java:43) from jrt:/java.base
	at java.lang.reflect.Method.invoke(java.base@10-internal/Method.java:564) from jrt:/java.base
[...]

PR #5789, opened 2019-05-16
Check and use openssl flags set in JVM build

a7ehuo:
[skip ci]
OpenJ9 build supports `--with-openssl=` flag which sets up
openssl include path and library. JITaaS is built by searching
the default places. This change checks and uses
`OPENSSL_DIR`, `OPENSSL_CFLAGS`, `OPENSSL_LIBS`, and
`OPENSSL_BUNDLE_LIB_PATH`, if they are exported in
`CUSTOM_COMPILER_ENV_VARS`.

   - If `--with-openssl=system` is set, JIT will link to OpenSSL lib
     specified by `OPENSSL_LIBS`
   - If `--with-openssl=fetched --enable-openssl-bundling` is set,
     JIT will link to OpenSSL lib specified by `OPENSSL_BUNDLE_LIB_PATH`
   - If `--with-openssl=fetched` is set,  JIT will link to OpenSSL lib
     specified by `OPENSSL_DIR`

Issue #5720

### DIFF
--- a/runtime/compiler/build/toolcfg/gnu/common.mk
+++ b/runtime/compiler/build/toolcfg/gnu/common.mk
@@ -489,3 +489,30 @@ PROTO_CMD?=protoc
 SOLINK_SLINK+=protobuf ssl
 
 CXX_DEFINES+=GOOGLE_PROTOBUF_NO_RTTI
+
+ifneq ($(OPENSSL_CFLAGS),)
+    ADD_OPENSSL_CFLAGS=false
+    # --with-openssl=system, OPENSSL_LIBS looks like "-L/path/to/system/openssllib -lssl -lcrypto"
+    ifneq ($(OPENSSL_LIBS),)
+        SOLINK_LIBPATH+=$(subst -L,,$(firstword $(OPENSSL_LIBS)))
+        ADD_OPENSSL_CFLAGS=true
+    else
+        # --with-openssl=fetched --enable-openssl-bundling
+        ifneq ($(OPENSSL_BUNDLE_LIB_PATH),)
+            SOLINK_LIBPATH+=$(OPENSSL_BUNDLE_LIB_PATH)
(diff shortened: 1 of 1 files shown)


a7ehuo:
Depends on #303 to make it work.

runtime/compiler/build/toolcfg/gnu/common.mk
```diff
@@ -489,3 +489,30 @@ PROTO_CMD?=protoc
 SOLINK_SLINK+=protobuf ssl
 
 CXX_DEFINES+=GOOGLE_PROTOBUF_NO_RTTI
+
+ifneq ($(OPENSSL_CFLAGS),)
+    SET_FLAG=false
+    # --with-openssl=system, OPENSSL_LIBS is -L/usr/local/lib -lssl -lcrypto
```
ymanton: Can we change the `-L/usr/local/lib` to `-L/path/to/system/openssl` or something like that? Depending on the distro it can be `/usr/lib`, `/usr/local/lib`, `/usr/lib64`, `/lib`, etc or a completely different path, so mentioning a specific location can mislead.

runtime/compiler/build/toolcfg/gnu/common.mk
```diff
@@ -489,3 +489,30 @@ PROTO_CMD?=protoc
 SOLINK_SLINK+=protobuf ssl
 
 CXX_DEFINES+=GOOGLE_PROTOBUF_NO_RTTI
+
+ifneq ($(OPENSSL_CFLAGS),)
+    SET_FLAG=false
```
ymanton: Can we name this variable something like `ADD_OPENSSL_CFLAGS ` to make it more clear what it's for?

a7ehuo:
@ymanton Thank you very much for trying this one out. Updated the code as suggested.

PR #5792, opened 2019-05-16
Switch JCL field ID cache setting order to avoid reading NULL

JasonFengJ9:
Switch `JCL` field ID cache setting order to avoid reading `NULL`

There is a time hole between setting `JCL` field IDs `CLS_sun_reflect_ConstantPool` and
`FID_sun_reflect_ConstantPool_constantPoolOop`. One thread could set `CLS_sun_reflect_ConstantPool` but not `FID_sun_reflect_ConstantPool_constantPoolOop` yet; another thread reads the `non-null CLS_sun_reflect_ConstantPool`, skip `initializeSunReflectConstantPoolIDCache()`, and de-reference `NULL
FID_sun_reflect_ConstantPool_constantPoolOop` which caused `gpf`.
Switching the setting order and add issueWriteBarrier().
Also refactoring for code format and comments.

This is discovered in an internal test. Verified that this PR can make the test pass consistently.

Reviewer: @DanHeidinga 

### DIFF
--- a/runtime/jcl/common/jniidcacheinit.c
+++ b/runtime/jcl/common/jniidcacheinit.c
@@ -105,7 +105,8 @@ initializeJavaLangStringIDCache(JNIEnv* env)
 
 
 /**
- * Initializes sun/reflect/ConstantPool class ID and constantPoolOop field ID in JniIDCache struct 
+ * Initializes sun/reflect/ConstantPool (Java 8) or jdk/internal/reflect/ConstantPool (Java 11+) class ID 
+ * and constantPoolOop field ID in JniIDCache struct 
  *
  * @param[in] env The JNI env
  *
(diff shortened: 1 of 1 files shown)


runtime/jcl/common/jniidcacheinit.c
```diff
@@ -114,69 +115,65 @@ initializeJavaLangStringIDCache(JNIEnv* env)
...
+			
+					omrthread_monitor_enter(javaVM->jclCacheMutex);
+					if (NULL == JCL_CACHE_GET(env,CLS_sun_reflect_ConstantPool)) {
+						/* Set FID_sun_reflect_ConstantPool_constantPoolOop first and CLS_sun_reflect_ConstantPool afterwards,
+						 * read CLS_sun_reflect_ConstantPool first in java_lang_Access.c:Java_java_lang_Access_getConstantPool(),
+						 * if NULL is returned, initializeSunReflectConstantPoolIDCache() is invoked to 
+						 * set FID_sun_reflect_ConstantPool_constantPoolOop, next read it and de-reference.
+						 * Otherwise, a NULL FID_sun_reflect_ConstantPool_constantPoolOop could be returned and cause gpf later.
+						 */
+						JCL_CACHE_SET(env, FID_sun_reflect_ConstantPool_constantPoolOop, fieldID);
```
pshipton: It's not really a proper fix. I suggest checking both values for NULL in Java_java_lang_Access_getConstantPool(), and calling initializeSunReflectConstantPoolIDCache() if either is NULL (which still isn't a proper fix, but a little bit better).

pshipton: Or just fix it properly by adding the write barrier after setting the fieldID.

JasonFengJ9: Added `issueWriteBarrier();` between setting `FID_sun_reflect_ConstantPool_constantPoolOop` and `CLS_sun_reflect_ConstantPool`.
This should not affect performance since the code specified is invoked only once.
Please have another look.

pshipton:
jenkins test sanity win,plinux jdk8

PR #5794, opened 2019-05-17
Pass reportByteCodeInfoAtCatchBlock to the server

dmitry-ten:
This option is a static variable and thus it is not
serialized in `J9::Options::packOptions`.
Since it is VM global, can add it to `VMInfo` and fetch
it with the rest of that struct.
This fixes a crash in `cmdLineTester_decompilationTests_1`.


dmitry-ten:
@ymanton ready for review.

Issue #5707, opened 2019-05-09
OpenJ9 cannot load/find/link .jnilib files on macOS; when renamed to .dylib, it works


rubin55:
When running some applications with OpenJ9 that depend on some Java Native Interface stuff on macOS, OpenJ9 seems to not be able to load those dependencies when they're called `<something>.jnilib`. It will load them when one renames or links to the files with a `<something>.dylib` name.

Here's some error output from Tanuki wrapper as used within Anypoint Studio (an Eclipse based IDE):

```
Java Service Wrapper Standard Edition 64-bit 3.5.37
  Copyright (C) 1999-2018 Tanuki Software, Ltd. All Rights Reserved.
    http://wrapper.tanukisoftware.com
  Licensed to MuleSoft Inc. for Mule Runtime Enterprise Edition

Launching a JVM...
Starting the Mule Container...
WrapperManager: Initializing...
WrapperManager: 
WrapperManager: ERROR - Unable to load the Wrapper's native library because none of the
WrapperManager:         following files:
WrapperManager:           libwrapper-macosx-universal-64.dylib
WrapperManager:           libwrapper.dylib
WrapperManager:         could be located on the following java.library.path:
WrapperManager:           /Applications/Anypoint Studio 7.app/Contents/Eclipse/configuration/org.eclipse.osgi/506/data/.runtimes/tooling-em0uavkj/lib/boot
WrapperManager:         Please see the documentation for the wrapper.java.library.path
WrapperManager:         configuration property.
WrapperManager: 
Writing to the backend pipe failed (35): Resource temporarily unavailable
Pipe send failed.  Incomplete.  Sent 8051 of 15226 bytes.
Writing to the backend pipe failed (35): Resource temporarily unavailable
Pipe send failed.  Incomplete.  Sent 0 of 2 bytes.
```

When I enter the above mentioned directory (`[..]/lib/boot`) I don't see a file named `libwrapper-macosx-universal-64.dylib`, but I do see a file named `libwrapper-macosx-universal-64.jnilib`. Note the extension. Renaming or linking to the file makes the wrapper work. This problem does not occur on Hotspot.

DanHeidinga:
After a bit of digging, it sounds like the order is:
- attempt to load library with `.dylib` suffix
- if failed, attempt to load with `.jnilib` suffix.

Likely something we need to change in ClassLoader `loadLibraryWithPath` or maybe `loadLibraryWithClassLoader`

DanHeidinga:
[...]

PR #5795, opened 2019-05-17
Add support for .jnilib extension on Mac OS

fengxue-IS:
- attempt to load native library with .jnilib extension
  after failed to load using .dylib extension

Close: #5707 

### DIFF
--- a/jcl/src/java.base/share/classes/java/lang/ClassLoader.java
+++ b/jcl/src/java.base/share/classes/java/lang/ClassLoader.java
@@ -1896,6 +1896,24 @@ static void loadLibraryWithPath(String libName, ClassLoader loader, String libra
 		}
 	}
 	byte[] message = ClassLoader.loadLibraryWithPath(com.ibm.oti.util.Util.getBytes(libName), loader, libraryPath == null ? null : com.ibm.oti.util.Util.getBytes(libraryPath));
+
+	/* As Mac OS X used to be using lib<name>.jnilib Java native library format,
+	 * VM will attempt loading native library using the legacy library extension
+	 * on Mac OS X to support older applications
+	 */
+	if (System.internalGetProperties().getProperty("os.name").equals("Mac OS X")) { //$NON-NLS-1$ //$NON-NLS-2$
+		if ((message != null) && (libraryPath != null)) {
+			String legacyPath = libraryPath.replaceAll("/$", "") + "/lib" + libName + ".jnilib"; //$NON-NLS-1$ //$NON-NLS-2$ //$NON-NLS-3$ //$NON-NLS-4$
+
+			try {
+				byte[] jnilibLoadMessage = ClassLoader.loadLibraryWithPath(com.ibm.oti.util.Util.getBytes(legacyPath), loader, null);
+				if (jnilibLoadMessage == null) {
+					message = null;
+				}
+			} catch (Exception e) { /* Ignore Exception */ }
+		}
+	}
+
 	if (message != null) {
 		String error;
 		try {


DanHeidinga:
@fengxue-IS Can you test this behaviour on JDK11, 12, & 13?

DanHeidinga:
Jenkins test sanity osx jdk8

DanHeidinga:
Jenkins test extended osx jdk11

fengxue-IS:
I have tried to rename `jdk/lib/libzip.dylib` to `jdk/lib/libzip.jnilib` on Java 11 builds and the library was found and loaded correctly by the JVM.

will test this for Java 12 & 13

DanHeidinga:
Probably worth testing Hotspot's behaviour here as well in case they've dropped this support in newer versions.

fengxue-IS:
Interesting results from testing Hotspot's behaviour:

Looks like Hotspot only allow `.jnilib` library to be loaded using Java System API (`System.loadLibrary()`)
libraries automatically loaded by the JVM (ie. java lib, jvmti agent lib) can only be loaded with `.dylib`

@DanHeidinga do we want to match the RI behaviour?
marking PR WIP while deciding this

fengxue-IS:
looking at the native library loading code, seems like OpenJ9 also permits load library without platform format (`<name>` instead of `lib<name>.ext`)

DanHeidinga:
> @DanHeidinga do we want to match the RI behaviour?

Yes, it doesn't make sense to propagate `.jnilib` into more places than the RI allows it unless there's a compelling reason such as a much cleaner solution by not limiting it.  
`.jnilib` is really a legacy feature that should be supported to ease migration.

fengxue-IS:
Update code to perform the `.jnilib` extension attempt in java level to avoid modifying shared native api for library loading.

@DanHeidinga can you take a look please?

jcl/src/java.base/share/classes/java/lang/ClassLoader.java
```diff
@@ -1896,6 +1896,20 @@ static void loadLibraryWithPath(String libName, ClassLoader loader, String libra
 		}
 	}
 	byte[] message = ClassLoader.loadLibraryWithPath(com.ibm.oti.util.Util.getBytes(libName), loader, libraryPath == null ? null : com.ibm.oti.util.Util.getBytes(libraryPath));
+
+	if (System.internalGetProperties().getProperty("os.name").equals("Mac OS X")) { //$NON-NLS-1$ //$NON-NLS-2$
+		if ((message != null) && (libraryPath != null)) {
+			String legacyPath = libraryPath.replaceAll("/$", "") + "/lib" + libName + ".jnilib"; //$NON-NLS-1$ //$NON-NLS-2$ //$NON-NLS-3$ //$NON-NLS-4$
+
+			try {
+				byte[] message2 = ClassLoader.loadLibraryWithPath(com.ibm.oti.util.Util.getBytes(legacyPath), loader, null);
```
DanHeidinga: Can you pick a more descriptive name than "message2"?

jcl/src/java.base/share/classes/java/lang/ClassLoader.java
```diff
@@ -1896,6 +1896,20 @@ static void loadLibraryWithPath(String libName, ClassLoader loader, String libra
 		}
 	}
 	byte[] message = ClassLoader.loadLibraryWithPath(com.ibm.oti.util.Util.getBytes(libName), loader, libraryPath == null ? null : com.ibm.oti.util.Util.getBytes(libraryPath));
+
+	if (System.internalGetProperties().getProperty("os.name").equals("Mac OS X")) { //$NON-NLS-1$ //$NON-NLS-2$
```
DanHeidinga: Please add a comment describing why this code is added and is OSX specific

jcl/src/java.base/share/classes/java/lang/ClassLoader.java
fengxue-IS: fixed

jcl/src/java.base/share/classes/java/lang/ClassLoader.java
fengxue-IS: fixed

DanHeidinga:
@fengxue-IS change looks good.  Can you add a test for this as well?  It can be done in a separate PR.

DanHeidinga:
Jenkins test sanity osx jdk8

DanHeidinga:

fengxue-IS:
Will open separate PR for test cases

PR #5799, opened 2019-05-17
JDK12+ doesn't need to set the user.timezone property to empty string

### LINKED ISSUE #5796: user.timezone system property can be null in JDK12+
There was a change to JDK12 to allow the user.timezone system property to be null, however OpenJ9 is still setting the value to empty string.

https://github.com/eclipse/openj9/blob/c3018fa8cc0d3514a3cc28109b15724251296849/runtime/jcl/unix/syshelp.c#L219
https://github.com/eclipse/openj9/blob/c3018fa8cc0d3514a3cc28109b15724251296849/runtime/jcl/win32/syshelp.c#L165

https://www.oracle.com/technetwork/java/javase/12-relnote-issues-5211422.html
```
core-libs/java.lang
➜ Initial Value of user.timezone System Property Changed 

The initial value of the user.timezone system property is undefined unless set using a command
line argument, for example, -Duser.timezone="America/New_York". The first time the default
timezone is needed, if user.timezone is undefined or empty the timezone provided by the
operating system is used. Previously, the initial value was the empty string. In JDK 12, 
System.getProperty("user.timezone") may return null.
```

pshipton:
Fixes #5796

### DIFF
--- a/runtime/jcl/unix/syshelp.c
+++ b/runtime/jcl/unix/syshelp.c
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 1998, 2018 IBM Corp. and others
+ * Copyright (c) 1998, 2019 IBM Corp. and others
  *
  * This program and the accompanying materials are made available under
  * the terms of the Eclipse Public License 2.0 which accompanies this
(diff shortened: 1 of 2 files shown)


pshipton:
jenkins test sanity,extended plinux jdk12

